Issue #997, opened 2024-09-06
[CRASH] Assertion failed: pthread_mutex_unlock


towest-nr:
### Crash report

```shell
...
1:S 05 Sep 2024 23:26:20.992 * RDB memory usage when created 2198.31 Mb
1:S 05 Sep 2024 23:26:27.858 * Done loading RDB, keys loaded: 1022610, keys expired: 0.
1:S 05 Sep 2024 23:26:27.858 * PRIMARY <-> REPLICA sync: Finished with success

=== VALKEY BUG REPORT START: Cut & paste starting from here ===
1:S 05 Sep 2024 23:29:26.128 # === ASSERTION FAILED ===
1:S 05 Sep 2024 23:29:26.128 # ==> ae.c:205 'pthread_mutex_unlock(&(eventLoop)->poll_mutex) == 0' is not true

------ STACK TRACE ------

1 redis-server
redis-server *redacted*:6379 [cluster](afterSleep+0x2e)[0x556ad98eff5e]
redis-server *redacted*:6379 [cluster](aeMain+0x64)[0x556ad98e86b4]
redis-server *redacted*:6379 [cluster](main+0x52f)[0x556ad98dc79f]
/lib/x86_64-linux-gnu/libc.so.6(+0x29d90)[0x7f1b00996d90]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x80)[0x7f1b00996e40]
redis-server *redacted*:6379 [cluster](_start+0x25)[0x556ad98dce95]

15 bio_close_file
/lib/x86_64-linux-gnu/libc.so.6(+0x91117)[0x7f1b009fe117]
/lib/x86_64-linux-gnu/libc.so.6(pthread_cond_wait+0x211)[0x7f1b00a00a41]
redis-server *redacted*:6379 [cluster](bioProcessBackgroundJobs+0x1ff)[0x556ad99bf82f]
/lib/x86_64-linux-gnu/libc.so.6(+0x94ac3)[0x7f1b00a01ac3]
/lib/x86_64-linux-gnu/libc.so.6(+0x126850)[0x7f1b00a93850]

18 io_thd_1 *
redis-server *redacted*:6379 [cluster](aeCreateFileEvent+0x146)[0x556ad98e1d76]
redis-server *redacted*:6379 [cluster](+0x1f5d56)[0x556ad9a48d56]
redis-server *redacted*:6379 [cluster](+0x1f5ee4)[0x556ad9a48ee4]
[...]

PR #1011, opened 2024-09-10
Fix crash in async IO threads with TLS

uriyage:
Fix for https://github.com/valkey-io/valkey/issues/997

Root Cause Analysis:
1. Two different jobs (READ and WRITE) may be sent to the same IO thread.
2. When processing the read job in `processIOThreadsReadDone`, the IO thread may find that the write job has also been completed.
3. In this case, the IO thread calls `processClientIOWriteDone` to first process the completed write job and free the COBs https://github.com/valkey-io/valkey/blob/affbea5dc1ae1a0d80019c4f313d2bf1c3fcb7f9/src/networking.c#L4666
4. If there are pending writes (resulting from pipeline commands), a new async IO write job is sent before processing the completed read job https://github.com/valkey-io/valkey/blob/affbea5dc1ae1a0d80019c4f313d2bf1c3fcb7f9/src/networking.c#L2417 When sending the write job, the `TLS_CONN_FLAG_POSTPONE_UPDATE_STATE` flag is set to prevent the IO thread from updating the event loop, which is not thread-safe.
5. Upon resuming the read job processing, the flag is cleared, https://github.com/valkey-io/valkey/blob/affbea5dc1ae1a0d80019c4f313d2bf1c3fcb7f9/src/networking.c#L4685 causing the IO thread to update the event loop.

Fix:
Prevent sending async write job for pending writes when a read job is about to be processed.

Testing:
The issue could not be reproduced due to its rare occurrence, which requires multiple specific conditions to align simultaneously.

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -2384,8 +2384,9 @@ parseResult handleParseResults(client *c) {
 
 /* Process the completion of an IO write operation for a client.
  * This function handles various post-write tasks, including updating client state,
+ * allow_async_writes - A flag indicating whether I/O threads can handle pending writes for this client.
  * returns 1 if processing completed successfully, 0 if processing is skipped. */
-int processClientIOWriteDone(client *c) {
+int processClientIOWriteDone(client *c, int allow_async_writes) {
     /* memory barrier acquire to get the latest client state */
     atomic_thread_fence(memory_order_acquire);
     /* If a client is protected, don't proceed to check the write results as it may trigger conn close. */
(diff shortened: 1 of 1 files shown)


src/networking.c
```diff
@@ -2384,8 +2384,9 @@ parseResult handleParseResults(client *c) {
 
 /* Process the completion of an IO write operation for a client.
  * This function handles various post-write tasks, including updating client state,
+ * allow_async_writes - A flag indicating whether I/O threads can handle pending writes for this client.
```
ranshid, line 2387: the flag basically state if we should offload pending writes to iothreads right?
maybe we should simply state that.
I thought of this flag as "try_send_pending_writes" but TBH it is not that important.

ranshid (commented):
Overall LGTM. minor comment about naming

ranshid (approved):

madolson (approved):

PR #1013, opened 2024-09-10
Fix wrong count for replica's tot-net-out

uriyage:
Fix duplicate calculation of replica's `net_output_bytes`

- Remove redundant calculation leftover from previous refactor
- Add test to prevent regression

### DIFF
--- a/tests/unit/introspection.tcl
+++ b/tests/unit/introspection.tcl
@@ -850,25 +850,42 @@ start_server {tags {"introspection"}} {
     
                 # Wait for replica to be connected before proceeding.
                 wait_for_ofs_sync $primary $replica
+                
+                # Avoid PINGs to make sure tot-net-out is stable.
+                $primary config set repl-ping-replica-period 3600
+
+                # Increase repl timeout to avoid replica disconnecting
+                $primary config set repl-timeout 3600
+                $replica config set repl-timeout 3600
+                
+                # Wait for the replica to receive the command.
+                wait_for_ofs_sync $primary $replica
     
+                # Get the tot-net-out of the replica before sending the command.
+                set info_list [$primary client list]
+                foreach info [split $info_list "\r\n"] {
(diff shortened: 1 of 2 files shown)


zuiderkwast (approved):
Looks correct to me. Thanks for the test.

You say the calculation was counted twice. Does that mean that without this fix, the test reports more than 20000 bytes for tot-net-out after replicating a 10000 bytes key?

enjoy-binbin (approved):
ok, so the old net_output_bytes is in line 2197 (or line 2190). LGTM.

uriyage:
> Looks correct to me. Thanks for the test.
> 
> You say the calculation was counted twice. Does that mean that without this fix, the test reports more than 20000 bytes for tot-net-out after replicating a 10000 bytes key?

Yes.

tests/unit/introspection.tcl
```diff
@@ -836,6 +836,43 @@ start_server {tags {"introspection"}} {
...
+                set info_list [$primary client list]
+                foreach info [split $info_list "\r\n"] {
+                    if {[string match "* flags=S *" $info]} {
+                        set out [get_field_in_client_info $info "tot-net-out"]
+                        break
+                    }
+                }
+
+                assert {$out > 0}
+                assert {$out < [expr $value_size + 1000]}
```
enjoy-binbin:
```suggestion
                assert_morethan $out 0
                assert_lessthan $out [expr $value_size + 1000]
```

ranshid (approved):
LGTM.

I think maybe we can straighten the test by disabling primary+replica KA and increase the repl-timeout so that this will not theoretically becomes flaky?

uriyage:
> LGTM.
> 
> I think maybe we can straighten the test by disabling primary+replica KA and increase the repl-timeout so that this will not theoretically becomes flaky?

Thanks, Changed per your suggestion.

PR #1016, opened 2024-09-11
Improve stability of hostnames test

### LINKED ISSUE #952: [Test failure] Slow runners show issue with hostnames
```
*** [err]: Verify the nodes configured with prefer hostname only show hostname for new nodes in tests/unit/cluster/hostnames.tcl
```

https://github.com/valkey-io/valkey/actions/runs/10540936135/job/29259761374?pr=923#step:5:7524

madolson:
Maybe partially resolves https://github.com/valkey-io/valkey/issues/952.

The hostnames test relies on an assumption that node zero and node six don't communicate with each other to test a bunch of behavior in the handshake stake. This was done by previously dropping all meet packets, however it seems like there was some case where node zero was sending a single pong message to node 6, which was partially initializing the state.

I couldn't track down why this happened, but I adjusted the test to simply pause node zero which also correctly emulates the state we want to be in since we're just testing state on node 6, and removes the chance of errant messages. The test was failing about 5% of the time locally, and I wasn't able to reproduce a failure with this new configuration.

### DIFF
--- a/tests/unit/cluster/hostnames.tcl
+++ b/tests/unit/cluster/hostnames.tcl
@@ -177,8 +177,7 @@ test "Verify the nodes configured with prefer hostname only show hostname for ne
 
     # Also make sure we know about the isolated master, we 
     # just can't reach it.
-    set master_id [R 0 CLUSTER MYID]
-    assert_match "*$master_id*" [R 6 CLUSTER NODES]
+    assert_match "*$primary_id*" [R 6 CLUSTER NODES]
 
     # Stop dropping cluster packets, and make sure everything
     # stabilizes
--- a/tests/unit/cluster/hostnames.tcl
+++ b/tests/unit/cluster/hostnames.tcl
@@ -132,7 +132,7 @@ test "Verify the nodes configured with prefer hostname only show hostname for ne
         R $j config set cluster-announce-hostname "shard-$j.com"
     }
 
-    # Grab the ID so we have it for later for validation
+    # Grab the ID so we have it later for validation
     set primary_id [R 0 CLUSTER MYID]
 
     # Prevent Node 0 and Node 6 from properly meeting,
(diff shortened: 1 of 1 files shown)


enjoy-binbin (approved):
nice!

tests/unit/cluster/hostnames.tcl
```diff
@@ -174,12 +174,12 @@ test "Verify the nodes configured with prefer hostname only show hostname for ne
...
     # Also make sure we know about the isolated master, we 
     # just can't reach it.
+    resume_process [srv 0 pid]
     set master_id [R 0 CLUSTER MYID]
     assert_match "*$master_id*" [R 6 CLUSTER NODES]
 
     # Stop dropping cluster packets, and make sure everything
     # stabilizes
-    R 0 DEBUG DROP-CLUSTER-PACKET-FILTER -1
     R 6 DEBUG DROP-CLUSTER-PACKET-FILTER -1
```
zuiderkwast: We shoudn't unpause node 0 before this check: "Also make sure we know about the isolated master, we just can't reach it."

You unpaused it here just to get MYID of node 0, right? We can fetch `R 0 CLUSTER MYID` early in the test case, before we pause node 0, instead.

zuiderkwast (commented):
Yeah, pause process seems more robust. Good idea!

tests/unit/cluster/hostnames.tcl
madolson: Sure

tests/unit/cluster/hostnames.tcl
```diff
@@ -132,11 +132,14 @@ test "Verify the nodes configured with prefer hostname only show hostname for ne
         R $j config set cluster-announce-hostname "shard-$j.com"
     }
 
+    # Grab the ID so we have it for later for validation
```
madolson:
```suggestion
    # Grab the ID so we have it later for validation
```

zuiderkwast (approved):

PR #1018, opened 2024-09-11
Do election in order based on failed primary rank to avoid voting conflicts

enjoy-binbin:
When multiple primary nodes fail simultaneously, the cluster can not recover
within the default effective time (data_age limit). The main reason is that
the vote is without ranking among multiple replica nodes, which case too many
epoch conflicts.

Therefore, we introduced into ranking based on the failed primary shard-id.
Introduced a new failed_primary_rank var, this var means the rank of this
myself instance in the context of all failed primary list. This var will be
used in failover and we will do the failover election packets in order based
on the rank, this can effectively avoid the voting conflicts.

If a single primary is down, the behavior is the same as before. If multiple
primaries are down, their replica election initiation time will be delayed
by 500ms according to the ranking.


tests/unit/cluster/failover2.tcl
```diff
@@ -64,3 +64,34 @@ start_cluster 3 4 {tags {external:skip cluster} overrides {cluster-ping-interval
     }
 
 } ;# start_cluster
+
```
enjoy-binbin, line 66: This test may be time-consuming. It basically cannot pass before the patch, but can pass locally after the patch.

src/cluster_legacy.c
```diff
@@ -4412,6 +4427,45 @@ int clusterGetReplicaRank(void) {
...
+        /* Skip nodes that do not need to participate in the rank. */
+        if (!nodeFailed(node) || !clusterNodeIsVotingPrimary(node) || node->num_replicas == 0) continue;
+
+        /* If cluster-replica-validity-factor is enabled, skip the invalid nodes. */
+        if (server.cluster_replica_validity_factor) {
+            if ((now - node->fail_time) > (server.cluster_node_timeout * server.cluster_replica_validity_factor))
+                continue;
+        }
+
+        if (memcmp(node->name, myself->replicaof->name, CLUSTER_NAMELEN) < 0) rank++;
```
PingXie: does it make sense to sort by `shard_id`? `replicaof` is not as reliable/up-to-date as `shard_id`. There is the chain replication and there is still the `replicaof` cycle.

src/cluster_legacy.c
```diff
@@ -4582,6 +4636,11 @@ void clusterHandleReplicaFailover(void) {
          * Specifically 1 second * rank. This way replicas that have a probably
          * less updated replication offset, are penalized. */
         server.cluster->failover_auth_time += server.cluster->failover_auth_rank * 1000;
+        /* We add another delay that is proportional to the failed primary rank.
+         * Specifically 0.5 second * rank. This way those failed primaries will be
+         * elected in rank to avoid the vote conflicts. */
+        server.cluster->failover_failed_primary_rank = clusterGetFailedPrimaryRank();
+        server.cluster->failover_auth_time += server.cluster->failover_failed_primary_rank * 500;
```
PingXie, line 4738: Curious - how did you arrive at 500?  Given that `CLUSTERMSG_TYPE_FAILOVER_AUTH_REQUEST` is broadcast and answered pretty much right away, unless the voter is busy, I would think the network round trip time between any two nodes should be significantly less than 50 ms for all deployments. I wonder if we could tighten it up a bit to like 250 or 200?

PingXie (commented):
LGTM overall. I like this idea. Thanks @enjoy-binbin!

src/cluster_legacy.c
enjoy-binbin: Seems to make sense, I'll think about it later.

src/cluster_legacy.c
enjoy-binbin: This 500 is just the experience points gained from here. I usually think that one election round can be completed between 500ms - 1s. Yes, i think the numbers may be adjustable, but I haven't experimented with it.
```
        server.cluster->failover_auth_time = mstime() +
                                             500 + /* Fixed delay of 500 milliseconds, let FAIL msg propagate. */
                                             random() % 500; /* Random delay between 0 and 500 milliseconds. */
```

tests/unit/cluster/failover2.tcl
madolson: We might want to run it solo as well given how much CPU it will consume.

src/cluster_legacy.c
madolson: Yeah, 500 seems high but it seems likely Ok.

src/cluster_legacy.c
madolson: Should it be `myself->shard_id`?

madolson:
Seems like a good idea to me as well.

src/cluster_legacy.c
enjoy-binbin: yes, it should be myself->shard_id, updated.

tests/unit/cluster/failover2.tcl
```diff
@@ -102,3 +100,36 @@ start_cluster 7 3 {tags {external:skip cluster} overrides {cluster-ping-interval
         resume_process [srv -2 pid]
     }
 } ;# start_cluster
+
+run_solo {cluster} {
+start_cluster 32 15 {tags {external:skip cluster} overrides {cluster-ping-interval 1000 cluster-node-timeout 15000}} {
```
madolson: You can indent this.

madolson (approved):
Directionally good with this change, the remaining tests still look good.

Issue #640, opened 2024-06-12
Support field level expire/TTL for hash, set and sorted set


xingbowang:
**The problem/use-case that the feature addresses**
My customer has a use case where each element within a Hash/Set/Sorted Set data type would have different TTL. Right now, it is done by manually write client side logic or lua script. It would be great if Valkey could support element level TTL, so the element get expired independently.

**Description of the feature**

Valkey would allow client to specific different TTL value at element level.

**Alternatives you've considered**

None.

**Additional information**

None.

hpatro:
Historically this was never added due to the amount of complexity. If that still holds true, I think we should evaluate if this fits more as a separate command/modules to incorporate element level TTL for a key. 

Note: Tair* has implemented some of these features in a module.

zvi-code:
sounds like ideally the expiry logic should be embedded into the dict itself [and than can be easily reused for all dict users], have we evaluated such an option? [instead of the additional expiry dict]
we could add expiry `existence bitmap` on the dict buckets [bit for bucket] to have optimized iterator over volatile keys, at some small cost at rehashing, or even ttl range bitmap to group similar ttl keys

madolson:
@zvi-code I documented a bit of that here: https://github.com/valkey-io/valkey/issues/169#issuecomment-2106368024. Embedding the TTL into the dict itself is OK and allows you to do lazy expiration very fast. Active expiration is a lot more painful though, since you need some way to actively go find those values which only somewhat works.

zvi-code:
Thanks @madolson for sharing, I'll take a look. I think with resonable assumptions it could be easier. The assumptions are about how tight do we want to be on eviction time, at least intuitively I feel, it's perfectly ok to be accurate at a multi second granularity[have grouping of some kind at a seconds granularity]. But I'll post my thoughts on the other thread, it looks very interesting!

ranshid:
I have been working on investigation of the problem for the last few weeks.

I would like to provide some initial thoughts I came up with for which I plan to PoC and decide on the best approuch for implementation.

First I would like to list the important tenets for the solution:

## Tenets

* **memory efficiency** - At the highest priority we target a minimal metadata overhead used in order to store and manage items TTL. While the optimal overhead to maintain item TTL is 8 bytes (could be less if we allow keeping offsets from the existing epoch diff time), we understand that maintaining active expiry logic will require use of more bytes for metadata. We will make our top priority effort to minimize this overhead.
[...]

Issue #169, opened 2024-04-03
Re-thinking the main hash table 


zuiderkwast:
This is an idea of the hash table approach I'd like to use for the main key space.

![image](https://github.com/valkey-io/valkey/assets/273886/58f9939e-9d71-4c2c-bf2c-9512cc930bb6)

The picture is taken from https://www.youtube.com/watch?v=ncHmEUmJZf4 at 48:57.

It's an open addressing hash table, consisting of blocks of 64 bytes (one cache line). Each block has space for 7 keys (no values, so it's semantically a set), as shown in the picture. For each key there are 8 hash bits, to quickly eliminate mismatching keys. In addition to that, there's one "presence" bit to indicate whether each of the elements exists or is an empty slot. Finally one "ever full" bit to indicate if the block has ever been full. This is like a shared tombstone and it is the condition to whether probing the next block is necessary. There is no internal order among these 7 keys. A new element can be inserted in the first free slot within the block.

SCAN can work by scanning block-by-block until a block with the "ever full" bit zero.

Incremental rehashing can work the same way as the classic dict, i.e. have two tables allocated during rehash.

The element I want to put in the hash table is a pointer to an object (`robj`) which is should contain all fields like key, value, TTL, LRU, in a compact way, with or without an embedded key.

Splitting this into tasks:

* [x] #991
* [x] #992
* [x] #1094
* [x] #1095
* [x] #1096

Regarding changing how TTLs are stored, it's not necessary to do in the scope of repacing the hash table implementation. I'm moving it to a separate issue: #990.

zuiderkwast:
@hpatro in another issue:

> @madolson and I were discussing internally to explore open addressing further with embedding to get rid of the next pointers.
> @zuiderkwast The part which we were concerned about was how to support `SCAN` command appropriately. Could you share some more thoughts on **"scanning block-by-block until a block with the "ever full" bit zero"**?

Yes I can.

When we scan, we scan one bucket at a time and the cursor corresponds to the bucket number.

The "ever full" bit is like a shared tombstone and it is the condition to whether probing the next block is necessary. So if it's zero, it means that no objects that hash to bucket N are instead inserted in bucket N+1 or later, due to probing. In this case, we can stop if we want. But if the "ever full" bit is one, some keys that should belong to N may have been inserted in N+1, N+2, etc. so to cover these when we scan bucket N, we must to continue scanning N+1, N+2, etc. until we reach a bucket with "ever full" zero.

[...]

PR #1020, opened 2024-09-11
Add valkeyCreate() and valkeyGetKey() functions

zuiderkwast:
@SoftlyRaining PTAL

This is a first draft of the robj hack.

There's a lot of duplicated code with createEmbeddedStringObject, but it's a PoC.

The EMBSTR is compact, but other robj are created with an expire field set to -1 (meaning no expire) and an embedded key pointer set to NULL.


ranshid:
@zuiderkwast / @SoftlyRaining  Sorry for crashing into your party :) 

I wanted to share that I was thinking of embedding the TTL inside the sds instead of robj.
This is mainly in order to serve a solution for #640 and also fit as general infra for [TTL rethinking](https://github.com/valkey-io/valkey/issues/169#issuecomment-2106368024).
I think that since general items are usually sds we can benefit from keeping an optional TTL embedded inside sds.
Not to dive into too many details we do have 5 spare bits in sds flags so we can use 1 in order to indicate an sds has a TTL. 
This way we can make sure a TTL is allocated as part of the sds.
I am working to write up a suggestion but maybe we can sync in order to prevent diverging suggestions about where TTL will be placed?

Issue #1006, opened 2024-09-10
[NEW] Output logs as JSON 


4mig4:
Would it be possible to get Redis logs output in JSON format with a timestamp that has timezone information as the first field.
This is useful when integrating and using redis in large kubernetes based systems which ingest and demand JSON logs to stdout for all apps.

madolson:
There was a previous PR in Redis where we discussed some log format:

> Copied from a copy of Redis #12934, explicitly didn't include a reference to avoid having folks reference Redis issue. 
> Add ability to configure Redis to output logs in logfmt (See https://brandur.org/logfmt) as well as configure timestamp format options to more standard ISO 8601 or unix timestamp.
> 
> This change is implemented by two configs: `log-format`: Either default or logfmt. `log-timestamp-format`: default, iso8601, or unix.
> 
> iso8601 includes time zone. Closes #12918 Closes #12932
> 
> ```
> LOG_FORMAT_DEFAULT + LOG_TIMESTAMP_DEFAULT
> 42888:M 12 Jan 2024 18:26:34.377 * Server initialized
> 42888:M 12 Jan 2024 18:26:34.377 * Ready to accept connections tcp
> 
> LOG_FORMAT_LOGFMT + LOG_TIMESTAMP_DEFAULT
> pid=43765 role=master   timestamp="12 Jan 2024 18:29:14.441" level=notice  message="Server initialized"
> pid=43765 role=master   timestamp="12 Jan 2024 18:29:14.441" level=notice  message="Ready to accept connections tcp"
> 
> LOG_FORMAT_LOGFMT + LOG_TIMESTAMP_ISO8601
> pid=44325 role=master   timestamp="2024-01-12T18:31:24.697+05:00" level=notice  message="Server initialized"
> pid=44325 role=master   timestamp="2024-01-12T18:31:24.697+05:00" level=notice  message="Ready to accept connections tcp"
> 
> LOG_FORMAT_LOGFMT + LOG_TIMESTAMP_UNIX
> pid=44790 role=master   timestamp="1705102366" level=notice  message="Server initialized"
> pid=44790 role=master   timestamp="1705102366" level=notice  message="Ready to accept connections tcp"
> ```


More information as well
> **The problem/use-case that the feature addresses**
> 
> The redis logs include datetime / timestamps in local timezone like the one below:
> 
> ```
[...]

PR #1022, opened 2024-09-12
Configurable log and timestamp formats (logfmt, ISO8601)

azuredream:
Add ability to configure log output format and timestamp format in the logs.

This change adds two new configs:

* `log-format`: Either legacy or logfmt (See https://brandur.org/logfmt)
* `log-timestamp-format`: legacy, iso8601 or milliseconds (since the epoch).

Related to #1006.

Example:

```text
$ ./valkey-server  /home/zhaoz12/git/valkey/valkey/valkey.conf
pid=109463 role=RDB/AOF timestamp="2024-09-10T20:37:25.738-04:00" level=warning message="WARNING Memory overcommit must be enabled! Without it, a background save or replication may fail under low memory condition. Being disabled, it can also cause failures without low memory condition, see https://github.com/jemalloc/jemalloc/issues/1328. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect."
pid=109463 role=RDB/AOF timestamp="2024-09-10T20:37:25.738-04:00" level=notice message="oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo"
pid=109463 role=RDB/AOF timestamp="2024-09-10T20:37:25.738-04:00" level=notice message="Valkey version=255.255.255, bits=64, commit=affbea5d, modified=1, pid=109463, just started"
pid=109463 role=RDB/AOF timestamp="2024-09-10T20:37:25.738-04:00" level=notice message="Configuration loaded"
pid=109463 role=master timestamp="2024-09-10T20:37:25.738-04:00" level=notice message="monotonic clock: POSIX clock_gettime"
pid=109463 role=master timestamp="2024-09-10T20:37:25.739-04:00" level=warning message="Failed to write PID file: Permission denied"
```


madolson:
😀 thanks for opening this, we're working on launching the next version but will take a look shortly.

PingXie:
I like this change. Free-form log lines can be tricky to parse and search, and this feels like a good step toward adopting more structured logging. We could even extend it further to include properties like "area" (e.g., "cluster," "normal replication," "dual channel replication," etc.). Thanks for your work on this, @azuredream!

src/server.c
```diff
@@ -109,11 +109,35 @@ const char *replstateToString(int replstate);
...
 void nolocks_localtime(struct tm *tmp, time_t t, time_t tz, int dst);
 
+/* Formats the timezone offset into a string.
+ *
+ * Assumes:
+ * - size of `buf` is 7 or larger
+ * - `timezone` is valid, within the range of [-50400, +43200].
+ * - `daylight_active` indicates whether dst is active (1) or not (0). */
+
+void format_timezone(char *buf, int timezone, int daylight_active) {
```
```suggestion
/* Formats the timezone offset into a string. daylight_active indicates whether dst is active (1)
 * or not (0). */

void formatTimezone(char *buf, size_t buflen, int timezone, int daylight_active) {
    serverAssert(buflen >= 7);
    serverAssert(timezone >= -50400 && timezone <= 43200);
```

src/server.c
```diff
@@ -133,23 +157,48 @@ void serverLogRaw(int level, const char *msg) {
...
+
+        case LOG_TIMESTAMP_ISO8601:
+            off = strftime(buf, sizeof(buf), "%Y-%m-%dT%H:%M:%S.", &tm);
+            char tzbuf[7];
+            format_timezone(tzbuf, server.timezone, server.daylight_active);
+            snprintf(buf + off, sizeof(buf) - off, "%03d%s", (int)tv.tv_usec / 1000, tzbuf);
+            break;
+
+        case LOG_TIMESTAMP_UNIX: snprintf(buf, sizeof(buf), "%ld", tv.tv_sec); break;
+        default: break;
```
madolson: Remove the default. If someone adds a new type it will throw a compilation warning that it's missing. With the `default` it will silently accept.

src/server.c
```diff
@@ -133,23 +157,48 @@ void serverLogRaw(int level, const char *msg) {
...
+            role_index = (server.primary_host ? 2 : 3); /* Slave or Master. */
+        }
+        switch (server.log_format) {
+        case LOG_FORMAT_LOGFMT: {
+            fprintf(fp, "pid=%d role=%s timestamp=\"%s\" level=%s message=\"%s\"\n", (int)getpid(), roles[role_index],
+                    buf, verbose_level[level], msg);
+            break;
+        }
+
+        default: {
```
```suggestion
        case LOG_TIMESTAMP_ISO8601:
        case LOG_TIMESTAMP_UNIX: {
```
I think this is a bit clearer from a wording perspective.

src/server.c
```diff
@@ -7054,4 +7103,4 @@ int main(int argc, char **argv) {
     return 0;
 }
 
-/* The End */
+/* The End */
```
```suggestion
/* The End */

```
I see a couple of other places you remove the newline at the end of the file, let's leave them there.

src/server.h
```diff
@@ -581,6 +581,15 @@ typedef enum {
...
 #define PAUSE_ACTION_REPLICA (1 << 4) /* pause replica traffic */
 
+/* Sets log format */
+typedef enum { LOG_FORMAT_DEFAULT = 0, LOG_FORMAT_LOGFMT } log_format_type;
+
+/* Sets log timestamp format */
+/* Also update LOG_TIMESTAMP_FORMATS */
+#define LOG_TIMESTAMP_DEFAULT 0
+#define LOG_TIMESTAMP_ISO8601 1
+#define LOG_TIMESTAMP_UNIX 2
```
```suggestion
/* Also update LOG_TIMESTAMP_FORMATS */
typedef enum { LOG_TIMESTAMP_DEFAULT = 0, LOG_TIMESTAMP_ISO8601, LOG_TIMESTAMP_UNIX}
```

src/server.c
```diff
@@ -133,23 +157,48 @@ void serverLogRaw(int level, const char *msg) {
...
+        switch (server.log_timestamp_format) {
+        case LOG_TIMESTAMP_DEFAULT:
+            off = strftime(buf, sizeof(buf), "%d %b %Y %H:%M:%S.", &tm);
+            snprintf(buf + off, sizeof(buf) - off, "%03d", (int)tv.tv_usec / 1000);
+            break;
+
+        case LOG_TIMESTAMP_ISO8601:
+            off = strftime(buf, sizeof(buf), "%Y-%m-%dT%H:%M:%S.", &tm);
+            char tzbuf[7];
+            format_timezone(tzbuf, server.timezone, server.daylight_active);
```
```suggestion
            char tzbuf[7];
            format_timezone(tzbuf, sizeof(tzbuf), server.timezone, server.daylight_active);
```
Based of my other comment.

madolson:
@valkey-io/core-team Please indicate your vote on this comment. The open questions I still have are if we want to also do the larger investigation before 8.1 launches about also including the component the log was generated from. We also have the JSON format to add on.

src/server.c
zuiderkwast: Is it already possible to configure the timezone? To log everything in UTC, the user can just set server timezone to UTC?

src/server.h
```diff
@@ -581,6 +581,15 @@ typedef enum {
 #define PAUSE_ACTION_EVICT (1 << 3)
 #define PAUSE_ACTION_REPLICA (1 << 4) /* pause replica traffic */
 
+/* Sets log format */
+typedef enum { LOG_FORMAT_DEFAULT = 0, LOG_FORMAT_LOGFMT } log_format_type;
```
zuiderkwast: Can we come up with another name for "default"? The name default implies that it will always be the default, which isn't necessarily true forever.

"compact", " traditional"...?

src/server.h
```diff
@@ -581,6 +581,15 @@ typedef enum {
...
+
+/* Sets log timestamp format */
+/* Also update LOG_TIMESTAMP_FORMATS */
+#define LOG_TIMESTAMP_DEFAULT 0
```
zuiderkwast: The date format "default" should have a more descriptive name (and one that doesn't imply that it will always be the default). Is this format used anywhere else, like in email headers?

It contains the name of the moth in English, as in "6 Oct 2024", so can we call it "English" or "British" or something? (Americans would put month first, right?)

zuiderkwast (commented):
I've reviewed this before in the open source redis time. I like it, just a few comments.

It seems that JSON is more popular. We can add that too in another PR. 

Configurable time format is good.

src/server.c
```diff
@@ -109,11 +109,35 @@ const char *replstateToString(int replstate);
...
+    buf[6] = '\0';
+}
+
 /* Low level logging. To use only for very big messages, otherwise
  * serverLog() is to prefer. */
 void serverLogRaw(int level, const char *msg) {
     const int syslogLevelMap[] = {LOG_DEBUG, LOG_INFO, LOG_NOTICE, LOG_WARNING};
     const char *c = ".-*#";
+    const char *verbose_level[] = {"debug", "info", "notice", "warning"};
+    const char *roles[] = {"sentinel", "RDB/AOF", "replica", "master"};
```
```suggestion
    const char *roles[] = {"sentinel", "RDB/AOF", "replica", "primary"};
```

src/config.c
```diff
@@ -122,6 +122,13 @@ configEnum propagation_error_behavior_enum[] = {{"ignore", PROPAGATION_ERR_BEHAV
                                                 {"panic-on-replicas", PROPAGATION_ERR_BEHAVIOR_PANIC_ON_REPLICAS},
                                                 {NULL, 0}};
 
+configEnum log_format_enum[] = {{"default", LOG_FORMAT_DEFAULT}, {"logfmt", LOG_FORMAT_LOGFMT}, {NULL, 0}};
+
+configEnum log_timestamp_format_enum[] = {{"default", LOG_TIMESTAMP_DEFAULT},
+                                          {"iso8601", LOG_TIMESTAMP_ISO8601},
+                                          {"unix", LOG_TIMESTAMP_UNIX},
```
madolson:
```suggestion
configEnum log_timestamp_format_enum[] = {{"legacy", LOG_TIMESTAMP_LEGACY},
                                          {"iso8601", LOG_TIMESTAMP_ISO8601},
                                          {"milliseconds", LOG_TIMESTAMP_MILLISECONDS},
```
@azuredream Was there a reason we added the unix milliseconds timestamp option? While discussing it, we weren't sure if it was needed.

zuiderkwast: (@madolson The PR currently emits it as seconds, not milliseconds, since the epoch.)

madolson: If we want the epoch, I think millisecond timestamp makes a lot more sense. A lot of important events happen sub-second.

src/server.c
```diff
@@ -133,23 +157,48 @@ void serverLogRaw(int level, const char *msg) {
...
+            break;
+
+        case LOG_TIMESTAMP_ISO8601:
+            off = strftime(buf, sizeof(buf), "%Y-%m-%dT%H:%M:%S.", &tm);
+            char tzbuf[7];
+            format_timezone(tzbuf, server.timezone, server.daylight_active);
+            snprintf(buf + off, sizeof(buf) - off, "%03d%s", (int)tv.tv_usec / 1000, tzbuf);
+            break;
+
+        case LOG_TIMESTAMP_UNIX: snprintf(buf, sizeof(buf), "%ld", tv.tv_sec); break;
```
madolson: I think this should be in milliseconds, not seconds.

src/server.c
azuredream: I use `case LOG_FORMAT_DEFAULT ` instead. 

'''
typedef enum { LOG_FORMAT_DEFAULT = 0, LOG_FORMAT_LOGFMT } log_format_type;
'''

src/server.h
azuredream: How about `LOG_TIMESTAMP_DMY`

src/server.c
madolson: Yeah, it is possible, but I think some people prefer to leave the timezone enabled on their machine?

src/server.h
madolson: The decision on this was to use "legacy".

src/server.c
madolson: Just to add an explanation here, I prefer having assertions in the codebase documenting assumptions then comments.

src/server.h
zuiderkwast: More important is what's visible to users, the config. Just "dmy"? Hm...

zuiderkwast: The decision was to use "legacy" for the legacy timestamp format too.

azuredream:
@madolson @zuiderkwast
PR updated as you suggested.
Thanks for reviewing.

azuredream:
Demo logs:

log-format logfmt
log-timestamp-format milliseconds
```text
pid=557888 role=RDB/AOF timestamp="1728757076.562" level=notice message="oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo"
```

log-format legacy
log-timestamp-format milliseconds
```text
558253:C 1728757310.034 * oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo
```

log-format logfmt
log-timestamp-format iso8601
```text
pid=558427 role=RDB/AOF timestamp="2024-10-12T14:24:28.141-04:00" level=notice message="oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo"
```

log-format legacy
log-timestamp-format iso8601
```text
558523:C 2024-10-12T14:24:57.923-04:00 * oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo
```

log-format logfmt
log-timestamp-format legacy
```text
pid=558599 role=RDB/AOF timestamp="12 Oct 2024 14:25:59.813" level=notice message="oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo"
```

log-format legacy
log-timestamp-format legacy
```text
558689:C 12 Oct 2024 14:26:30.872 * oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo
```

src/server.c
```diff
@@ -133,23 +154,47 @@ void serverLogRaw(int level, const char *msg) {
...
+            off = strftime(buf, sizeof(buf), "%Y-%m-%dT%H:%M:%S.", &tm);
+            char tzbuf[7];
+            formatTimezone(tzbuf, sizeof(tzbuf), server.timezone, server.daylight_active);
+            snprintf(buf + off, sizeof(buf) - off, "%03d%s", (int)tv.tv_usec / 1000, tzbuf);
+            break;
+
+        case LOG_TIMESTAMP_MILLISECONDS:
+            snprintf(buf, sizeof(buf), "%ld.%03d", tv.tv_sec, (int)tv.tv_usec / 1000);
+            break;
+        }
```
zuiderkwast: This is seconds with fractions. I believe what @madolson had in mind is just milliseconds since the epoch as an integer, as on this page: https://currentmillis.com/

@madolson Is that right?

zuiderkwast: `tv.tv_sec * 1000 + tv.tv_usec`

It needs more than 32 bits. Note that `long` can be 32-bit on 32-bit platforms, so I think we need `long long` (`%lld`) for this.

azuredream: Code updated as you suggested.
```text
569917:C 1728914477183 * oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo
```

src/server.c
```diff
@@ -133,23 +154,47 @@ void serverLogRaw(int level, const char *msg) {
...
-            role_char = 'C'; /* RDB / AOF writing child. */
+            role_index = 1; /* RDB / AOF writing child. */
         } else {
-            role_char = (server.primary_host ? 'S' : 'M'); /* replica or Primary. */
+            role_index = (server.primary_host ? 2 : 3); /* Slave or Master. */
+        }
+        switch (server.log_format) {
+        case LOG_FORMAT_LOGFMT:
+            fprintf(fp, "pid=%d role=%s timestamp=\"%s\" level=%s message=\"%s\"\n", (int)getpid(), roles[role_index],
+                    buf, verbose_level[level], msg);
```
zuiderkwast: I think this PR looks good now, almost ready to merge.

I'm just thinking about the corner case that log messages containing special characters like `"`, backslash or linebreaks. I believe modules can log any text they want. (Is that true? I didn't check.) I couldn't find if logfmt supports escaping. Maybe we can just replace some forbidden chars with other chars or something.

Ideally we should have a test case with a module that tries to log something like that. Maybe it's difficult. Do we have anything in our test framework to check what we to get in the logs?

If we add `LOG_FORMAT_JSON` in the future, it will need a similar but maybe slightly different check that it's correct UTF-8 and maybe do some escaping of characters like backslash.

azuredream: Not sure if that's a question for me. 

Modules are logging messages through this routine currently, and following the c string rule, such as backslash+n means newline? Why do we need new rules for logfmt?
One reason I can see is that double quotes is too commonly used in users messages. We can define something more complicated to brace message and timestamp. Such as !?+message+?!
Plus, anything we do to the message content will not be transparent to users. And they'll have to change their message parser.

src/server.c
madolson: Yes, the final result is what I expected.

src/server.c
zuiderkwast: It's a question for you and others. :)

> Modules are logging messages through this routine currently, and following the c string rule, such as backslash+n means newline? Why do we need new rules for logfmt?

Modules can log stuff using the function below, which is just a wrapper for serverLog. It's implemented as `VM_Log` in `module.c`.

              void ValkeyModule_Log(ValkeyModuleCtx *ctx,
                                    const char *levelstr,
                                    const char *fmt,
                                    ...);

Let's say a module logs something that contains a key name or something that we can't control. It would make it possible to mess up the log if we don't do anything special to the message. Let's say the message content is the following two-lines string:

    x"
    pid=9999 timestamp=never level=banana message="y

If we just include this as is, it will appear as message "x" and then an injected log line with dummy information and message "y" logged below the first line. (If the module is written in C, then line breaks can be written as `\n` in string literals, but it can also come from e.g. a key name or a command argument that comes from a client and is combined using `snprintf` or in some other way.)

> One reason I can see is that double quotes is too commonly used in users messages. We can define something more complicated to brace message and timestamp. Such as !?+message+?!

Is that possible in logfmt? I haven't found a formal syntax of logfmt but I would guess that all logfmt parsers expect double quotes `"message"`?

> Plus, anything we do to the message content will not be transparent to users. And they'll have to change their message parser.

Yes, that can be a problem. The log message created from within Valkey can be designed to avoid using `"`. For log messages from modules, it is probably fine to forbid it (change `"` to `'` and remove all line breaks) just to make sure the log isn't broken.

src/server.c
```diff
@@ -109,11 +109,53 @@ const char *replstateToString(int replstate);
...
+        }
+        src++;
+    }
+    msg[dst] = '\0';
+}
+
 /* Low level logging. To use only for very big messages, otherwise
  * serverLog() is to prefer. */
-void serverLogRaw(int level, const char *msg) {
+void serverLogRaw(int level, char *msg) {
```
azuredream: Mutability modified.

src/server.c
azuredream: Code updated as your suggestion. 
Please note that I've modified the mutability of `*msg`.

src/server.c
zuiderkwast: Hm... I hope this works. I see many places in the code where this function is called with a string literal, like

    serverLogRaw(LL_WARNING, "Failed applying new module configuration. Restoring previous settings.");

This can't be mutable, I think...

src/server.c
```diff
@@ -127,9 +127,30 @@ void formatTimezone(char *buf, size_t buflen, int timezone, int daylight_active)
...
+    if (msg == NULL) return;
+
+    int src = 0, dst = 0;
+    while (msg[src] != '\0') {
+        if (msg[src] == '"') {
+            msg[dst++] = '\'';
+        } else if (msg[src] != '\n' && msg[src] != '\r') {
+            msg[dst++] = msg[src];
+        }
+        src++;
```
zuiderkwast: Just an idea: If we replace `\r` and `\n` with a space instead, then we don't need to copy/move the rest of the string. We don't need the `dst` index.

```suggestion
    while (msg[src] != '\0') {
        if (msg[src] == '"') {
            msg[src] = '\'';
        } else if (msg[src] == '\n' && msg[src] == '\r') {
            msg[src] = ' ';
        }
        src++;
```

src/server.c
```diff
@@ -133,23 +175,48 @@ void serverLogRaw(int level, const char *msg) {
...
+            role_index = 1; /* RDB / AOF writing child. */
         } else {
-            role_char = (server.primary_host ? 'S' : 'M'); /* replica or Primary. */
+            role_index = (server.primary_host ? 2 : 3); /* Slave or Master. */
+        }
+        switch (server.log_format) {
+        case LOG_FORMAT_LOGFMT:
+            filterInvalidLogfmtChar(msg);
+            fprintf(fp, "pid=%d role=%s timestamp=\"%s\" level=%s message=\"%s\"\n", (int)getpid(), roles[role_index],
+                    buf, verbose_level[level], msg);
```
zuiderkwast: If we can't make `msg` mutable, then, what we need to do is to copy to a temporary buffer. But I think we should avoid that and only do it if there are some forbidden characters.

We can check if there are any forbidden characters first, and then copy to a temp buffer if there are any. It's no problem if this case is less optimal because we expect it to be almost never used, so a heap allocated string is OK.

Something like this:

```suggestion
            char *safemsg = msg;
            if (hasInvalidLogfmtChar(msg)) {
                safemsg = filterInvalidLogfmtChar(msg);
            }
            fprintf(fp, "pid=%d role=%s timestamp=\"%s\" level=%s message=\"%s\"\n", (int)getpid(), roles[role_index],
                    buf, verbose_level[level], safemsg);
            if (safemsg != msg) {
                zfree(safemsg);
            }
```

azuredream: Code updated. All tests passed.

I couldn't do `char *safemsg = msg;` without compiler warning, so I had to do something circuitous.

azuredream:
Code updateded.

src/server.c
```diff
@@ -133,23 +192,55 @@ void serverLogRaw(int level, const char *msg) {
...
-            role_char = 'C'; /* RDB / AOF writing child. */
+            role_index = 1; /* RDB / AOF writing child. */
         } else {
-            role_char = (server.primary_host ? 'S' : 'M'); /* replica or Primary. */
+            role_index = (server.primary_host ? 2 : 3); /* Slave or Master. */
+        }
+        switch (server.log_format) {
+        case LOG_FORMAT_LOGFMT:
+            if (hasInvalidLogfmtChar(msg)) {
+                safemsg = zmalloc(LOG_MAX_LEN);
```
zuiderkwast: OK, this solution looks fine.

Just two small ideas:

* We could declare this variable here, because it's only used within this local scope.

* We could even easily make it stack allocated and skip the zfree.

```suggestion
                char safemsg[LOG_MAX_LEN];
```

src/server.c
zuiderkwast: OK, great. My example was untested. :)

src/server.c
azuredream: Code updated. All tests passed.

I got a compile warning when I tried to declare it within switch body.  
The warning disappeared today...

src/server.c
```diff
@@ -133,23 +191,54 @@ void serverLogRaw(int level, const char *msg) {
...
+        int role_index;
         if (server.sentinel_mode) {
-            role_char = 'X'; /* Sentinel. */
+            role_index = 0; /* Sentinel. */
         } else if (pid != server.pid) {
-            role_char = 'C'; /* RDB / AOF writing child. */
+            role_index = 1; /* RDB / AOF writing child. */
         } else {
-            role_char = (server.primary_host ? 'S' : 'M'); /* replica or Primary. */
+            role_index = (server.primary_host ? 2 : 3); /* Slave or Master. */
```
zuiderkwast: Nit: We should avoid slave and master even in comments.

```suggestion
            role_index = (server.primary_host ? 2 : 3); /* replica or primary. */
```

zuiderkwast (approved):
Looks good to me now.

One of the commits is missing sign-off. See the Details link on the DCO CI job. Can you fix it? You can squash and force-push as you want.

See also the clang-format CI job. It's a trailing whitespace on one line.

Is the Valkey logo logged on startup? It must look very weird with logfmt?

azuredream:
> Is the Valkey logo logged on startup? It must look very weird with logfmt?

Actually it looks fine
```text
pid=716151 role=RDB/AOF timestamp="1729404137284" level=notice message="oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo"
pid=716151 role=RDB/AOF timestamp="1729404137284" level=notice message="Valkey version=255.255.255, bits=64, commit=ca0cedf5, modified=1, pid=716151, just started"
pid=716151 role=RDB/AOF timestamp="1729404137284" level=notice message="Configuration loaded"
pid=716151 role=primary timestamp="1729404137285" level=notice message="monotonic clock: POSIX clock_gettime"
pid=716151 role=primary timestamp="1729404137286" level=warning message="Failed to write PID file: Permission denied"
                .+^+.                                                
            .+#########+.                                            
        .+########+########+.           Valkey 255.255.255 (ca0cedf5/1) 64 bit
    .+########+'     '+########+.                                    
 .########+'     .+.     '+########.    Running in standalone mode
 |####+'     .+#######+.     '+####|    Port: 6379
 |###|   .+###############+.   |###|    PID: 716151                     
 |###|   |#####*'' ''*#####|   |###|                                 
 |###|   |####'  .-.  '####|   |###|                                 
 |###|   |###(  (@@@)  )###|   |###|          https://valkey.io      
 |###|   |####.  '-'  .####|   |###|                                 
 |###|   |#####*.   .*#####|   |###|                                 
 |###|   '+#####|   |#####+'   |###|                                 
 |####+.     +##|   |#+'     .+####|                                 
 '#######+   |##|        .+########'                                 
    '+###|   |##|    .+########+'                                    
        '|   |####+########+'                                        
             +#########+'                                            
                '+v+' 

```

azuredream:
Updated comment, fixed format, fixed unsigned commit.

I appreciate your continuous effort to review my contributions and provide valuable suggestions for improvements. 
Thanks to all of you.
@zuiderkwast @madolson @PingXie

zuiderkwast:
> I appreciate your continuous effort to review my contributions and provide valuable suggestions for improvements.

Thanks, that's our job. :) Thanks for your patience!

It's a bit weird that we print `role=RDB/AOF` for the messages during startup, like `"oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo"`. I guess it was always a "C" for these messages, but I never noticed it.

azuredream:
> It's a bit weird that we print `role=RDB/AOF` for the messages during startup, like `"oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo"`. I guess it was always a "C" for these messages, but I never noticed it.

Please refer to the discussion around Jan 11th. We introduced reader friendly role string Intentionally.
https://github.com/redis/redis/pull/12934
```text
    I suggest role=%c (instead of role_char=%c). Is one char clear enough? If not, we can consider using role=master, role=sentinel, etc. but it makes the log longer. Maybe one char is clear enough?
```

zuiderkwast:
I know we want reader-friendly role. What I think is weird is that we print RDB/AOF during startup. This is not a forked child process performing RDB dump or AOF rewrite.

Now I see why: `server.pid` has not been initialized yet when these line are logged. (I think we can initialize server.pid earlier, but that is for another PR. It's a separate discussion.)

azuredream:
> Now I see why: `server.pid` has not been initialized yet when these line are logged. (I think we can initialize server.pid earlier, but that is for another PR. It's a separate discussion.)

I think that means I don't need to update this PR.
I'll be glad to help if there is any further requirement.

zuiderkwast:
> > Now I see why: `server.pid` has not been initialized yet when these line are logged. (I think we can initialize server.pid earlier, but that is for another PR. It's a separate discussion.)
> 
> I think that means I don't need to update this PR. I'll be glad to help if there is any further requirement.

That's right, I hope this PR is ready to merge now. @madolson do you have any more comments before we merge this?

@azuredream If you want to make a follow-up PR for the role during startup, please continue reading...

1. I'm looking at where `server.pid = getpid()` is done. Currently, it's done in `initServer()` but it would be better to do it before the first logging happens, in `main()` before we load config files. This is a simple fix. With this change, we will print "primary" instead of "RDB/AOF" during startup, which I think makes more sense. The server starts as a primary. Do you want to do it in a follow-up PR?

2. Another potential problem. Actually, I don't think this one is a problem we can solve, but maybe something we can mention in the documentation later.

   I see that loading config is done early during startup. It can happen that some messages are logged before we set the logging format config, so maybe we get some log lines with the legacy format before we switch to logfmt. That's also not good if we want to parse the file using logfmt tools.

   If we fail to load a config file, we log `"Fatal error, can't open config file '%s': %s"` and exit. We always load a config file before we read any config provided on the command line (in `valkey-server dummy.conf --log-format logfmt`, the file is loaded first) so there is no way to get this logged in logfmt format. We probably don't want to change the behaviour to load configs on the command line before the file. That would be a breaking change, I think.

    Another message is when we load config from stdin.

        serverLog(LL_NOTICE, "Reading config from stdin");

   To read config from stdin, you pass "-" on the command line. I tried this and I used `echo` to pipe some config to it. I got one line of "legacy" log before switching to logfmt:

   ```
   $ echo "save ''" | ./valkey-server --log-format logfmt -
   514332:C 21 Oct 2024 10:53:37.887 * Reading config from stdin
   pid=514332 role=RDB/AOF timestamp="21 Oct 2024 10:53:37.887" level=warning message="WARNING Memory overcommit must be enabled! Without it, a background save or replication may fail under low memory condition. Being disabled, it can also cause failures without low memory condition, see https://github.com/jemalloc/jemalloc/issues/1328. To fix this issue add 'vm.overcommit_memory = 1' to /etc/sysctl.conf and then reboot or run the command 'sysctl vm.overcommit_memory=1' for this to take effect."
   pid=514332 role=RDB/AOF timestamp="21 Oct 2024 10:53:37.888" level=notice message="oO0OoO0OoO0Oo Valkey is starting oO0OoO0OoO0Oo"
   ...
   ```

   One line of legacy log format, before we switch to logfmt. I don't know if we can solve this. Just like when loading from a file, we load the stdin before any command line config flags.

azuredream:
> 1. I'm looking at where `server.pid = getpid()` is done. Currently, it's done in `initServer()` but it would be better to do it before the first logging happens, in `main()` before we load config files. This is a simple fix. With this change, we will print "primary" instead of "RDB/AOF" during startup, which I think makes more sense. The server starts as a primary. Do you want to do it in a follow-up PR?

Sure. Please feel free to create an issue and assign to me.

zuiderkwast:
> Please feel free to create an issue and assign to me.

We don't need issues for everything, but if you insist...

zuiderkwast:
Document log format configs in valkey.conf

Configs are documented there. We missed that. Let's do it as a follow-up.

madolson:
> That's right, I hope this PR is ready to merge now. @madolson do you have any more comments before we merge this?

@zuiderkwast Apparently I had some pending comments which I never published. The main was just I realized there was no tests here, which would probably be useful to add. If you agree I can make an issue for it.

PR #1025, opened 2024-09-12
Add the missing help output for new command: client capa redirect

hwware:
Update client help output message for new command: client capa redirect


src/networking.c
```diff
@@ -3485,6 +3485,10 @@ void clientCommand(client *c) {
         const char *help[] = {
 "CACHING (YES|NO)",
 "    Enable/disable tracking of the keys for next command in OPTIN/OPTOUT modes.",
+"CAPA <option> [options...]",
+"    The client claims its some capability options, Options are:",
```
enjoy-binbin:
```suggestion
"    The client claims its some capability options. Options are:",
```

src/networking.c
```diff
@@ -3485,6 +3485,10 @@ void clientCommand(client *c) {
...
+"    * REDIRECT",
+"      The client can handle redirection during primary and replica failover",
```
enjoy-binbin: do we need to mention that the failover is non-cluster mode?

madolson: +1, I think we should clarify that it's for standalone.

```suggestion
"      The client can handle redirection during primary and replica failover in standalone mode.",
```

enjoy-binbin (approved):

madolson (approved):

PR #1029, opened 2024-09-13
Migrate the redis name to server in benchmark code

Shivshankar-Reddy:
I was trying to run valkey-benchmark and missed to start server prior to the benchmark and oberved that it reports it is not able to connect to Redis as below
image

So changed the ouput log to server as below.
image

### DIFF
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -238,7 +238,7 @@ static redisContext *getRedisContext(const char *ip, int port, const char *hosts
     else
         ctx = redisConnectUnix(hostsocket);
     if (ctx == NULL || ctx->err) {
-        fprintf(stderr, "Could not connect to Redis at ");
+        fprintf(stderr, "Could not connect to server at ");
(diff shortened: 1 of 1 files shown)


hwware:
@enjoy-binbin Do we need to update the words to server or Valkey? any idea?

enjoy-binbin:
I would lean towards Valkey. But at the same time, benchmark should be able to connect to any Redis-like software, so maybe server is the bese, i'm not sure which one is better.

@zuiderkwast

enjoy-binbin (approved):

zuiderkwast (approved):
> I would lean towards Valkey. But at the same time, benchmark should be able to connect to any Redis-like software, so maybe server is the bese, i'm not sure which one is better.
> 
> @zuiderkwast

I like "server". Yes it can try to connect to anything. If we get a connection error, it's maybe not even a server. :-)

"Could not connect to 127.0.0.1" ?

PingXie:
`server` works for me. I think we should take this for 8.0.

PingXie (approved):

PR #1030, opened 2024-09-13
Final batch of changes for 8.0 GA release

PingXie:

PR #981, opened 2024-09-02
Fix replica unable trigger migration when it received CLUSTER SETSLOT in advance

enjoy-binbin:
When the primary and replica cluster-allow-replica-migration configuration items are
different, the replica migration does not meet expectations.

Originally: primary not allow, replica allow. In this case, the code will migrate the
replica away, and the primary will keep as an empty primary. However, in #970, we found
a timing issue. For example, if the replica receives CLUSTER SETSLOT first and then receives
the gossip change, it will not trigger the replica-migration.

We perform replica migration explicitly in CLUSTER SETSLOT in this case.

Closes #970.

PR #1031, opened 2024-09-14
8.0 GA Release

PingXie:


enjoy-binbin (approved):

PingXie:
@valkey-io/core-team PTAL

enjoy-binbin:
[Make KEYS to be an exact match if there is no pattern](https://github.com/valkey-io/valkey/pull/1031/commits/3bcd6735f40b3fdd1cc37bb667a8f2428ff5e9e6)
[Revert make KEYS to be an exact match if there is no pattern](https://github.com/valkey-io/valkey/pull/1031/commits/d8a305c33302b58634e576972a042a5327f86518)
[Fix reconfiguring sub-replica causing data loss when myself change shard_id](https://github.com/valkey-io/valkey/pull/1031/commits/e458f9cd268826e85e749ae593fba233c7c25edd)

these there (the top 3) is dup, these was merged when RC2

and this one is missing i guess https://github.com/valkey-io/valkey/pull/981

enjoy-binbin:
https://github.com/valkey-io/valkey/pull/981
https://github.com/valkey-io/valkey/pull/1025

these two are missing, other LGTM.

PingXie:
all fixed.

#981 was not missing but it was lumped with another PR #1027 for some reason. I have rebased/fixed this commit.

zuiderkwast:
This PR appears to have thousands of lines changed. Is it rebased onto current latest 8.0 branch or how is it done?

What's the difference between this PR and #1030? It has much fewer changed lines.

Release notes update for GA?

PingXie:
> This PR appears to have thousands of lines changed. Is it rebased onto current latest 8.0 branch or how is it done?
> 
> What's the difference between this PR and #1030? It has much fewer changed lines.


#1030 is everything we marked "to be backported" in the 8.0 project and this is everything "unstable". We need to make a call between 1030 and 1031 (this one) but we don't need both. I am leaning towards going with 1031.

> 
> Release notes update for GA?

The release notes will have to come after us deciding which pr to take.

PingXie:
> #981 #1025
> 
> these two are missing, other LGTM.

btw, #1025 got lumped into the commit for #981. I don't think it is easier to split at this point so I am going to leave it as-is (@hwware any concerns?). 

Plus, with us taking everything in unstable, my understanding is that we prefer a squash merge so all that matters is the payload. thoughts?

zuiderkwast:
> Plus, with us taking everything in unstable, my understanding is that we prefer a squash merge so all that matters is the payload. thoughts?

You think nobody will use `git log` on the 8.0.0 tag or branch? I think it'd be good to include all commits. When we squash-backported to 7.2 we decided to revert that and take separate commits instead.

Another idea: We could start off a new 8.0 branch from unstable and just add the release notes on top. Would that work?

zuiderkwast (approved):

PingXie:
>You think nobody will use git log on the 8.0.0 tag or branch? I think it'd be good to include all commits. When we squash-backported to 7.2 we decided to revert that and take separate commits instead.

Yeah, sure will keep all the commits.

>Another idea: We could start off a new 8.0 branch from unstable and just add the release notes on top. Would that work?

Will it look odd in the commit history if the GA release isn't a descendant of the RC releases?

PR #1038, opened 2024-09-16
Add test cases for valkey-benchmark negative path

Shivshankar-Reddy:
Added the test case to check the wrong host case for valkey-benchmark.

### DIFF
--- a/tests/integration/valkey-benchmark.tcl
+++ b/tests/integration/valkey-benchmark.tcl
@@ -37,6 +37,20 @@ tags {"benchmark network external:skip logreqres:skip"} {
             default_set_get_checks
         }
 
+        test {benchmark: negative path} {
+            set wrong_host 127.0.0.0
+            set wrong_port 6500
+            foreach e {valkeybenchmark valkeybenchmarkuri valkeybenchmarkuriuserpass} {
+                if {$e eq {valkeybenchmarkuriuserpass}} {
+                    set cmd [valkeybenchmarkuriuserpass $wrong_host $wrong_port "default" pass "-c 5 -n 10 -t set"] 
+                } else {
+                    set cmd [$e $wrong_host $wrong_port "-c 5 -n 10 -t set"]
+                }
+                catch { exec {*}$cmd } res
+                assert_match  "*Could not connect to server at*" $res
+            }
+        }
+
         test {benchmark: connecting using URI set,get} {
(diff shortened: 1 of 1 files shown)


tests/integration/valkey-benchmark.tcl
```diff
@@ -37,6 +37,20 @@ tags {"benchmark network external:skip logreqres:skip"} {
             default_set_get_checks
         }
 
+        test {benchmark: negative path} {
+            set wrong_host 127.0.0.0
```
xbasel, line 41: Hostnames/IPs and ports seem pretty basic; was there a specific reason to test these, or was it due to an edge case encountered or a broader fix?

Shivshankar-Reddy: Hi @xbasel, Yes you are right, this is actually motivated by this PR https://github.com/valkey-io/valkey/pull/1029.

tests/integration/valkey-benchmark.tcl
```diff
@@ -37,6 +37,20 @@ tags {"benchmark network external:skip logreqres:skip"} {
...
+            set wrong_port 6500
+            foreach e {valkeybenchmark valkeybenchmarkuri valkeybenchmarkuriuserpass} {
+                if {$e eq {valkeybenchmarkuriuserpass}} {
+                    set cmd [valkeybenchmarkuriuserpass $wrong_host $wrong_port "default" pass "-c 5 -n 10 -t set"]
```
ranshid, line 45: no nead for set test and multiple clients IMO. just iuse a simple -I

tests/integration/valkey-benchmark.tcl
```diff
@@ -37,6 +37,20 @@ tags {"benchmark network external:skip logreqres:skip"} {
             default_set_get_checks
         }
 
+        test {benchmark: negative path} {
```
ranshid, line 40: negative_path suggested a very high level test.  I am just not sure why we decided to focus on this negative test while there are potentially many more negative testing we can add (eg bad arguments, TLS authentication, cluster discovery etc...). 
Anyway I would suggest to change the name to more explicitly reflect the test.
```suggestion
        test {benchmark: unable to connect to bad host} {
```

ranshid (commented):
I think there is no big blocker here. Only the motivation is a little bit vague IMO.
For example why did we decide to focus only on this scenario?

PR #1041, opened 2024-09-17
Add API UpdateRuntimeArgs for updating the module arguments during runtime

### LINKED ISSUE #1177: [NEW]Update the module args during runtime and save the new args value into the conf file
Before Redis OSS 7, if we load a module with some args during runtime,
and run the command "config rewrite", the module information will not be saved into the 
conf file.

Since Redis OSS 7 and Valkey 7.2,  if we load a module with some args during runtime,
the module information (path, args number, and args value) can be saved into the conf file after config rewrite command is called.
Thus the module will be loaded automatically when the server startup next time.

Following is one example:

bind 172.25.0.58
port 7000
protected-mode no
enable-module-command yes

Generated by CONFIG REWRITE
latency-tracking-info-percentiles 50 99 99.9
dir "/home/ubuntu/valkey"
save 3600 1 300 100 60 10000
user default on nopass sanitize-payload ~* &* +@all
loadmodule tests/modules/datatype.so 10 20

However there is one problem. 
If developers write a module, and update the running args by someway, the updated args can not be saved into the conf file even "config rewrite" is called.
The reason comes from the following function  
rewriteConfigLoadmoduleOption     (src/config.c)

void rewriteConfigLoadmoduleOption(struct rewriteConfigState *state) {
    ..........    
    struct ValkeyModule *module = dictGetVal(de);
    line = sdsnew("loadmodule ");
    line = sdscatsds(line, module->loadmod->path);
    for (int i = 0; i < module->loadmod->argc; i++) {
        line = sdscatlen(line, " ", 1);
        line = sdscatsds(line, module->loadmod->argv[i]->ptr);
    }
    rewriteConfigRewriteLine(state, "loadmodule", line, 1);
    .......
}

The function only save the initial args information (module->loadmod) into the conf file.

There are 2 ways to solve the problem: 

Solution 1:   It needs to take 2 steps.   (This option is not what we choose)

step 1: client calls command:  MODULE SET-ARGUMENT <module_name> [args...] to update the args (module->loadmod) during runtime
step 2: client gets the updated args (module->loadmod) in the module through the new API  ValkeyModule_GetRunTimeArgs(ctx);
Reference:   https://github.com/valkey-io/valkey/pull/1041

Solution 2:  it only takes 1 step

Call int VM_UpdateRunTimeArgs(ValkeyModuleCtx *ctx, int index, char *value) API in the module and update the specific args according to the index (module->loadmod).
Reference:   https://github.com/hwware/valkey/pull/8

Thus, I think the best way is to add a Module API to implement the above 2 features.

Administrators or developer could update the module args by the following way:

MODULE UNLOAD 
Update module args in the conf file
MODULE LOAD 

But I think there are following disadvantages:

1.  some modules can not be unloaded. Such as the example module datatype.so, which is tests/modules/datatype.so
2.  it is not atomic operation for MODULE UNLOAD + MODULE LOAD
3.  sometimes, if we just run the module unload, the client business could be interrupted

**Additional information**

hwware:
Before Redis OSS 7, if we load a module with some arguments during runtime,
and run the command "config rewrite", the module information will not be saved into the
config file.

Since Redis OSS 7 and Valkey 7.2, if we load a module with some arguments during runtime,
the module information (path, arguments number, and arguments value) can be saved into the config file after config rewrite command is called.
Thus, the module will be loaded automatically when the server startup next time.

Following is one example:

bind 172.25.0.58
port 7000
protected-mode no
enable-module-command yes

Generated by CONFIG REWRITE
latency-tracking-info-percentiles 50 99 99.9
dir "/home/ubuntu/valkey"
save 3600 1 300 100 60 10000
user default on nopass sanitize-payload ~* &* +https://github.com/ALL
loadmodule tests/modules/datatype.so 10 20

However, there is one problem.
If developers write a module, and update the running arguments by someway, the updated arguments can not be saved into the config file even "config rewrite" is called.
The reason comes from the following function
rewriteConfigLoadmoduleOption (src/config.c)

void rewriteConfigLoadmoduleOption(struct rewriteConfigState *state) {
..........
struct ValkeyModule *module = dictGetVal(de);
line = sdsnew("loadmodule ");
line = sdscatsds(line, module->loadmod->path);
for (int i = 0; i < module->loadmod->argc; i++) {
line = sdscatlen(line, " ", 1);
line = sdscatsds(line, module->loadmod->argv[i]->ptr);
}
rewriteConfigRewriteLine(state, "loadmodule", line, 1);
.......
}

The function only save the initial arguments information (module->loadmod) into the configfile.

After core members discuss, ref https://github.com/valkey-io/valkey/issues/1177

We decide add the following API to implement this feature:

Original proposal:

int VM_UpdateRunTimeArgs(ValkeyModuleCtx *ctx, int index, char *value);

Updated proposal:

ValkeyModuleString **values VM_GetRuntimeArgs(ValkeyModuleCtx *ctx);
**int VM_UpdateRuntimeArgs(ValkeyModuleCtx *ctx, int argc, ValkeyModuleString **values);

Why we do not recommend the following way: 

MODULE UNLOAD
Update module args in the conf file
MODULE LOAD

I think there are the following disadvantages:

1. Some modules can not be unloaded. Such as the example module datatype.so, which is tests/modules/datatype.so
2. it is not atomic operation for MODULE UNLOAD + MODULE LOAD
3. sometimes, if we just run the module unload, the client business could be interrupted


zuiderkwast:
Looks simple, but what's the use case? Why does a module need to update args in runtime?

A module can already have config that is updated using CONFIG SET, right?

hwware:
@zuiderkwast I update the top description for this PR, i think this is a good way to access the updated module arguments. Pls take a look when you have time, Thanks a lot

zuiderkwast:
You added a module API to access the argv. Interesting.

But, I am not convinced. :)

In a program written in C, the arguments are passed to main as `int main(int argc, char **argv)`. It's not possible to change these args later, from outside the program, while the program is running.

The `argv` for MODULE LOAD is the same. If you set the arguments with MODULE SET-ARGUMENTS later, the module was still loaded with the old arguments and most modules (or new modules like valkey-bloom) will not notice the arguments have changed. But there will be a problem. In the log, it looks like the module is using the new arguments, even if it is using the old arguments.

MODULE UNLOAD + MODULE LOAD doesn't have this problem.

zuiderkwast:
> You added a module API to access the argv. Interesting.
> 
> But, I am not convinced. :)
> 
> In a program written in C, the arguments are passed to main as `int main(int argc, char **argv)`. It's not possible to change these args later, from the outside, while the program is running.
> 
> The `argv` for MODULE LOAD is the same. If you set the arguments with MODULE SET-ARGUMENTS later, the module was still loaded with the old arguments and most modules (or new modules like valkey-bloom) will not notice the arguments have changed. But there will be a problem. In the log, it looks like the module is using the new arguments, even if it is using the old arguments.
> 
> MODULE UNLOAD + MODULE LOAD doesn't have this problem.

hwware:
> You added a module API to access the argv. Interesting.
> 
> But, I am not convinced. :)
> 
> In a program written in C, the arguments are passed to main as `int main(int argc, char **argv)`. It's not possible to change these args later, from outside the program, while the program is running.
> 
> The `argv` for MODULE LOAD is the same. If you set the arguments with MODULE SET-ARGUMENTS later, the module was still loaded with the old arguments and most modules (or new modules like valkey-bloom) will not notice the arguments have changed. But there will be a problem. In the log, it looks like the module is using the new arguments, even if it is using the old arguments.
> 
> MODULE UNLOAD + MODULE LOAD doesn't have this problem.

Thanks for your comment, but some cons for operation MODULE UNLOAD + MODULE LOAD

1.  some modules maybe not unloaded. Such as the example module datatype.so, which is  tests/modules/datatype.so
2.  it is not atomic operation for MODULE UNLOAD + MODULE LOAD
3.  sometimes, if we just run the module unload, the client business could be interrupted
4.  through the way MODULE UNLOAD + MODULE LOAD to update module runtime args, the management plan and data plan software need to be updated as well.

I would like to create an issue to let community to discuss this problem. Thanks

hwware:
@madolson As we discussed, I desgin the API as int VM_UpdateRuntimeArgs(ValkeyModuleCtx *ctx, int argc, ValkeyModuleString **argv), please take a look, Thanks

src/module.c
```diff
@@ -3042,6 +3042,23 @@ client *moduleGetReplyClient(ValkeyModuleCtx *ctx) {
     }
 }
 
+int VM_UpdateRunTimeArgs(ValkeyModuleCtx *ctx, int argc, ValkeyModuleString **argv) {
```
zuiderkwast: Implementation looks good.

The name here is "RunTime" but in the PR title it says "Runtime". Question to native English speakers: Is it one word (runtime) or two words (run time)?

hwware: There is one exception named RuntimeException  in Java, thus the correct function should be VM_UpdateRuntimeArgs(...) ?

hwware: Addressed, update the function name to VM_UpdateRuntimeArgs(...)

src/module.c
```diff
@@ -3042,6 +3042,23 @@ client *moduleGetReplyClient(ValkeyModuleCtx *ctx) {
     }
 }
 
+int VM_UpdateRuntimeArgs(ValkeyModuleCtx *ctx, int argc, ValkeyModuleString **argv) {
```
zuiderkwast: Please add a comment. It is used for generated documentation. Without a comment, the function will not be included in the API documentation. The comment needs to use correct markdown syntax.

You can try the ruby script to generate the documentation markdown: `utils/generate-module-api-doc.rb`. (Later, we run the script and save the output in the valkey-doc repo in the file `topics/module-api-ref.md`.)

tests/unit/moduleapi/moduleconfigs.tcl
```diff
@@ -243,5 +245,17 @@ start_server {tags {"modules"}} {
...
+       set modulename [lmap x [r module list] {dict get $x name}]
+       assert_not_equal [lsearch $modulename myhello] -1
+       string match "10 20 30" [lmap x [r module list] {dict get $x args}]
+       r hello.hi 40 50 60 70
+       string match "40 50 60 70" [lmap x [r module list] {dict get $x args}]
+    }
 }
 
+
+
```
zuiderkwast: We don't need extra space in the end of the file.

zuiderkwast (commented):
Looks good to me. Only documentation comment missing.

What about VM_GetRuntimeArgs, is it not needed?

src/module.c
hwware: Addressed, add comments already

Comandante318 (approved):

src/module.c
```diff
@@ -2254,6 +2254,33 @@ int moduleIsModuleCommand(void *module_handle, struct serverCommand *cmd) {
     return (cp->module == module_handle);
 }
 
+/* ValkeyModule_UpdateRuntimeArgs can be used to update the values of module->loadmod
+ * so that the updated values can be saved into conf file once 'config rewrite' command
+ * is called
+ * One example can be found in file modules/moduleparameter.c
```
zuiderkwast: Good, but I have some small comments:

`module->loadmod` is something internal in Valkey. Module authors are not aware of this. Let's assume module authors are know about the API documentation and `valkeymodule.h`.

Usually we write commands in uppercase, like CONFIG REWRITE.

Dot is missing after "is called".

Is the example in src/modules/ or tests/modules/ ? We have both, so it's better to be specific. Also it can be good mention that that it is in the Valkey source code, because the documentation is published in other places and module authors are maybe not even looking at valkey source code.

src/module.c
```diff
@@ -2254,6 +2254,33 @@ int moduleIsModuleCommand(void *module_handle, struct serverCommand *cmd) {
...
+ *
+ * Returns:
+ * - VALKEYMODULE_OK on successfully updating.
+ * - VALKEYMODULE_ERR on failure.
+ */
+int VM_UpdateRuntimeArgs(ValkeyModuleCtx *ctx, int argc, ValkeyModuleString **argv) {
```
zuiderkwast: All the other API functions that have argv/argc have argv before argc. We should do the same here.

For example OnLoad:

    int ValkeyModule_OnLoad(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc);

hwware: Addressed, I just found the  order of parameters argv and argc is reversed to C main function

zuiderkwast: > Addressed, I just found the order of parameters argv and argc is reversed to C main function

Yes, :joy: probably it was an early mistake when the module API was created. But now it's better to follow this style.

tests/modules/moduleparameter.c
```diff
@@ -0,0 +1,25 @@
...
+  ValkeyModule_UpdateRuntimeArgs(ctx, argv, argc);
+  return ValkeyModule_ReplyWithSimpleString(ctx, "Module runtime args test");
+}
+
+int ValkeyModule_OnLoad(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
+    VALKEYMODULE_NOT_USED(argv);
+    VALKEYMODULE_NOT_USED(argc);
+
+    if (ValkeyModule_Init(ctx,"myhello",1,VALKEYMODULE_APIVER_1)
+        == VALKEYMODULE_ERR) return VALKEYMODULE_ERR;
```
zuiderkwast: "myhello"?

Maybe "moduleparameter" is more descriptive?

tests/modules/moduleparameter.c
```diff
@@ -0,0 +1,25 @@
...
+
+
+    if (ValkeyModule_CreateCommand(ctx,"hello.hi",
+        GET_HELLO,"fast",0,0,0) == VALKEYMODULE_ERR)
```
zuiderkwast: Command name "hello.hi" and function name GET_HELLO are not perfect names. :)

Something like "updateargs", updateArgsCommand?

zuiderkwast (commented):
Sorry for many small reviews. I was going to click Approve, but then I read the test case. more carefully and found some strange function and command names. :)

tests/modules/moduleparameter.c
```diff
@@ -0,0 +1,25 @@
+#include "valkeymodule.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <ctype.h>
+#include <string.h>
+
+int GET_HELLO(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
+  ValkeyModule_UpdateRuntimeArgs(ctx, argv, argc);
+  return ValkeyModule_ReplyWithSimpleString(ctx, "Module runtime args test");
```
madolson:
```suggestion
    ValkeyModule_UpdateRuntimeArgs(ctx, argv, argc);
    return ValkeyModule_ReplyWithSimpleString(ctx, "Module runtime args test");
```
I guess we should probably apply the formatter on these files as well.

src/module.c
```diff
@@ -2254,6 +2254,34 @@ int moduleIsModuleCommand(void *module_handle, struct serverCommand *cmd) {
     return (cp->module == module_handle);
 }
 
+/* ValkeyModule_UpdateRuntimeArgs can be used to update the module argument values.
+ * The function parameter 'argc' indicates the number of updated arguments, and 'argv'
+ * represents the values of the updated arguments.
+ * Once 'CONFIG REWRITE' command is called, the updated argument values can be saved into conf file.
+ * One example can be found in file tests/modules/moduleparameter.c.
```
madolson: We don't normally reference the test modules since these are rendered here: https://valkey.io/topics/modules-api-ref/, which doesn't clearly map to the test files anymore. I'm not sure it's really needed to show an example.

src/module.c
```diff
@@ -2254,6 +2254,34 @@ int moduleIsModuleCommand(void *module_handle, struct serverCommand *cmd) {
...
+ *
+ * Returns:
+ * - VALKEYMODULE_OK on successfully updating.
+ * - VALKEYMODULE_ERR on failure.
+ */
+int VM_UpdateRuntimeArgs(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
+    if (!ctx->module->onload) {
+        return VALKEYMODULE_ERR;
+    }
```
madolson: Why is this only allowed at startup? I thought the whole point was they were able to update these at any point in time.

src/redismodule.h
```diff
@@ -357,6 +357,7 @@
 #define RedisModule_SetModuleAttribs ValkeyModule_SetModuleAttribs
 #define RedisModule_IsModuleNameBusy ValkeyModule_IsModuleNameBusy
 #define RedisModule_WrongArity ValkeyModule_WrongArity
+#define RedisModule_UpdateRuntimeArgs ValkeyModule_UpdateRuntimeArgs
```
zuiderkwast: The Redis module API is just for backward compatibilty. We don't add functions to that API.

```suggestion
```

src/module.c
zuiderkwast: Good point. I'm surprised the test case passes.

hwware: The condition "!ctx->module->onload" is checked here is not like what you thought it can be only allowed to update at startup, the goal is to check if ctx->module->onload is NULL to prevent NULL pointer exception. This API can be called at any time during runtime.

src/module.c
hwware: It looks like no specific test file is mentioned in reference, I will remove it.

tests/modules/moduleparameter.c
hwware: Thanks for your comments, it is updated

src/module.c
zuiderkwast: I think it means that onload is ongoing. If it doesn't, then there is something wrong.

I tried to follow the function calls that happen when a module is loaded. This is what I could find:

* MODULE LOAD -> moduleCommand -> moduleLoad.
* moduleLoad loads the module file using dlopen.
* moduleLoad finds the ValkeyModule_OnLoad function in the module using dlsym.
* moduleLoad calls the ValkeyModule_OnLoad in the module.
* The module's ValkeyModule_OnLoad function calls ValkeyModule_Init.
* ValkeyModule_Init (implemented in valkeymodule.h) calls ValkeyModule_SetModuleAttribs.
* VM_SetModuleAttribs creates the struct ctx->module and sets ctx->module->onload = 1.
* The ValkeyModule_OnLoad function calls other functions, such as ValkeyModule_CreateCommand
* The ValkeyModule_OnLoad function returns. 
* moduleLoad logs serverLog(LL_NOTICE, "Module '%s' loaded from %s", ctx.module->name, path);
* moduleLoad sets ctx.module->onload = 0;
* This ctx is freed, but the module object is added to a dict.

There is no other place that onload is set. That means onload = 1 only when ValkeyModule_OnLoad is running.

zuiderkwast: The onload field in ValkeyModule is not a pointer. It is an int.

    int onload;                           /* Flag to identify if the call is being made from Onload (0 or 1) */

tests/modules/moduleparameter.c
```diff
@@ -0,0 +1,27 @@
+#include "valkeymodule.h"
+#include <stdio.h>
+#include <stdlib.h>
+#include <ctype.h>
+#include <string.h>
+
+int test_module_update_parameter(ValkeyModuleCtx *ctx,
+                                 ValkeyModuleString **argv, int argc) {
+  ValkeyModule_UpdateRuntimeArgs(ctx, argv, argc);
```
zuiderkwast, line 10: Check that ValkeyModule_UpdateRuntimeArgs returns VALKEYMODULE_OK here.

tests/unit/moduleapi/moduleconfigs.tcl
```diff
@@ -243,5 +245,14 @@ start_server {tags {"modules"}} {
...
-}
+    test {Module Update Args } {
+       r module load $testmoduleparameter 10 20 30
 
+       set t [r module list]
+       set modulename [lmap x [r module list] {dict get $x name}]
+       assert_not_equal [lsearch $modulename moduleparameter] -1
+       string match "10 20 30" [lmap x [r module list] {dict get $x args}]
+       r testmoduleparameter.update.parameter 40 50 60 70
+       string match "40 50 60 70" [lmap x [r module list] {dict get $x args}]
```
zuiderkwast: This is not an assert. It will silently fail. It just returns 0 and the return value is ignored here.

tests/modules/moduleparameter.c
hwware: Update to:  return ValkeyModule_UpdateRuntimeArgs(ctx, argv, argc);

tests/unit/moduleapi/moduleconfigs.tcl
hwware: Sorry, I use incorrect grammar, Now it is addressed by using assert_equal

tests/modules/moduleparameter.c
zuiderkwast: Yes, or

```suggestion
  int res = ValkeyModule_UpdateRuntimeArgs(ctx, argv, argc);
  assert(res == VALKEYMODULE_OK);
```

zuiderkwast: The command still needs to send a reply to the client, right?

src/module.c
hwware: Sorry, it looks like on that time, my brain is offline.  Yes, you are right, it is int, not pointer. The condition here is useless, I should remove it.

tests/unit/moduleapi/moduleconfigs.tcl
```diff
@@ -243,5 +245,14 @@ start_server {tags {"modules"}} {
...
     }
-}
+    test {Module Update Args } {
+       r module load $testmoduleparameter 10 20 30
 
+       set t [r module list]
+       set modulename [lmap x [r module list] {dict get $x name}]
+       assert_not_equal [lsearch $modulename moduleparameter] -1
+       assert_equal "{10 20 30}" [lmap x [r module list] {dict get $x args}]
+       r testmoduleparameter.update.parameter 40 50 60 70
```
zuiderkwast: It should return OK on success, right?

```suggestion
       assert_equal OK [r testmoduleparameter.update.parameter 40 50 60 70]
```

zuiderkwast: Or maybe it returns the string "Module runtime args test", the reply sent by this command in the previous commits.

```suggestion
       assert_equal "Module runtime args test" [r testmoduleparameter.update.parameter 40 50 60 70]
```

... but OK seems more normal I think :grin:

hwware: I update this part as assert_equal OK [r testmoduleparameter.update.parameter 40 50 60 70]

tests/modules/moduleparameter.c
hwware: I just have one statement as return ValkeyModule_ReplyWithSimpleString(ctx, "OK");

hwware:
I update some logic in module.c, now it works well. @zuiderkwast

src/module.c
```diff
@@ -2254,6 +2254,29 @@ int moduleIsModuleCommand(void *module_handle, struct serverCommand *cmd) {
...
 }
 
+/* ValkeyModule_UpdateRuntimeArgs can be used to update the module argument values.
+ * The function parameter 'argc' indicates the number of updated arguments, and 'argv'
+ * represents the values of the updated arguments.
+ * Once 'CONFIG REWRITE' command is called, the updated argument values can be saved into conf file.
+ *
+ * Returns:
+ * - VALKEYMODULE_ERR on failure.
+ */
```
zuiderkwast: I think this function can't return VALKEYMODULE_ERR. I can't see any return like this.

But it always returns VALKEYMODULE_OK. It was not mentioned.

tests/unit/moduleapi/moduleconfigs.tcl
```diff
@@ -243,5 +245,14 @@ start_server {tags {"modules"}} {
             assert_equal [r config get moduleconfigs.memory_numeric] "moduleconfigs.memory_numeric 1024"
         }
     }
-}
+    test {Module Update Args } {
```
zuiderkwast: A space

```suggestion
    test {Module Update Args} {
```

zuiderkwast (commented):
Yes, now it looks better. Before, it couldn't work. :D

Just some small comment now, then ready to merge I think.

tests/unit/moduleapi/moduleconfigs.tcl
hwware: The comment is addressed

src/module.c
hwware: The comment is addressed

zuiderkwast (approved):
LGTM

PR #1045, opened 2024-09-18
clang-format: set ColumnLimit to 0 and reformat

mkmkme:
This commit hopefully improves the formatting of the codebase by setting ColumnLimit to 0 and hence stopping clang-format from trying to put as much stuff in one line as possible.

This change enabled us to remove most of `clang-format off` directives and fixed a bunch of lines that looked like this:

```c
#define KEY \
    VALUE /* comment */
```

Additionally, one pair of `clang-format off` / `clang-format on` had `clang-format off` as the second comment and hence didn't enable the formatting for the rest of the file. This commit addresses this issue as well.

Please tell me if anything in the changes seem off. If everything is fine, I will add this commit to `.git-blame-ignore-revs` later.


bjosv:
Nice.
Its a bit annoying that clang-format had some formatting changes somewhere between 18.1.3 and 18.1.6. Maybe that the reason for the failing CI check?
The check seem to use clang-format 18.1.8.

src/ae.h
```diff
@@ -42,11 +42,11 @@
...
-    4 /* With WRITABLE, never fire the event if the                                                                    \
-         READABLE event already fired in the same event                                                                \
-         loop iteration. Useful when you want to persist                                                               \
-         things to disk before sending replies, and want                                                               \
+#define AE_BARRIER                                       \
+    4 /* With WRITABLE, never fire the event if the      \
+         READABLE event already fired in the same event  \
+         loop iteration. Useful when you want to persist \
+         things to disk before sending replies, and want \
          to do that in a group fashion. */
```
zuiderkwast: Would it reject it if we would do like this?

```suggestion
#define AE_BARRIER 4 /* With WRITABLE, never fire the event if the      \
                        READABLE event already fired in the same event  \
                        loop iteration. Useful when you want to persist \
                        things to disk before sending replies, and want \
                        to do that in a group fashion. */
```

I was most likely like this originally before clang-format changed it to the worse...

src/atomicvar.h
zuiderkwast, line 1: Hasn't this file been deleted?

It should be.

It's not used anywhere, is it?

zuiderkwast (commented):
IMO this makes clang-format much less annoying. It lets us remove the `/* clang-format off */` which is a good indication that it respects our style better.

`ColumnLimit: 0` lets the programmer decide where to make line breaks, right? This seems more practical. Clang-format can help with other things.

@PingXie @madolson WDYT?

src/ae.h
mkmkme: Oops, that's my mistake, I've overlooked it. It will not reject it, I'll fix it :)

mkmkme:
> Nice. Its a bit annoying that clang-format had some formatting changes somewhere between 18.1.3 and 18.1.6. Maybe that the reason for the failing CI check? The check seem to use clang-format 18.1.8.

Yeah, good point. I forgot to pin the `clang-format` version when making this change.

Basically I would suggest pinning the `clang-format` version both in CI and in the documentation (and for developers) because it changes within the patch versions (there will be a difference between 18.1.3 and there is a difference between 18.1.6 and 18.1.8)

src/atomicvar.h
mkmkme: I think it was accidentally left in the project. I will remove it with a separate commit within this PR.

mkmkme:
Actually now I realized that the large diff in the CI was caused not by the version mismatch, but by the typo in my formatting script that caused only `.h` files to be reformatted automatically.

My proposal regarding the `clang-format` version is following: keep the things in CI the way they are right now (so, clang-format will be installed from apt.llvm.org) and observe if there are any issues about the formatting in the future.

If there are more backwards-incompatible changes with patch version of `clang-format` (like it was between 18.1.3 and 18.1.6), I will create a new pull request with pinning `clang-format` version.

PingXie:
Yeah I like this change too. Now thinking back the column limit is actually net negative as it forced us to spread `clang-format off` all over the codebase. 

re - version compat. I did get into a lot of issues when switching between version 18 and 19 but I would be surprised if a patch version upgrade breaks things. that said, I see no harm pinning to a specific patch version. 

Thanks for making this change, @mkmkme! Please let me know when this PR is ready for the final review.

mkmkme:
@PingXie 

>but I would be surprised if a patch version upgrade breaks things

It did. I had incompatibility issues between 18.1.3 and 18.1.6 in [libvalkey PR](https://github.com/valkey-io/libvalkey/pull/53#issuecomment-2304796040)

mkmkme:
Also I can see that "Clang Format Check" got its green mark. So @PingXie @zuiderkwast @bjosv this is now ready for a proper review :)

src/t_string.c
```diff
@@ -299,7 +299,7 @@ int parseExtendedStringArgumentsOrReply(client *c, int *flags, int *unit, robj *
             addReplyErrorObject(c,shared.syntaxerr);
             return C_ERR;
         }
-        /* clang-format off */
+        /* clang-format on */
```
PingXie, line 302: delete?
```suggestion
```

PingXie (approved):
LGTM. Thanks @mkmkme!

src/t_string.c
mkmkme: No, if we delete  this one, it will significantly change the "if-else" chain starting at line 229. Hence this is one of two things I decided to keep unformatted.

zuiderkwast (approved):
Looks great to me!

PR #156, opened 2024-04-02
Rename redisDb to serverDb

9bany:
A task of https://github.com/valkey-io/valkey/issues/144

PR #1046, opened 2024-09-18
fix memory allocation for server databases

ranshid:
Fix a bug in the way we allocate memory for the server databases


enjoy-binbin (approved):
WTF. Introduced in #156.

PR #1048, opened 2024-09-18
Fix the typo in the valkey.conf

Shivshankar-Reddy:


enjoy-binbin (commented):
This doesn't look like typo, it's written this way on purpose i guess

Shivshankar-Reddy:
Hi @enjoy-binbin ,
Thank you for the review!
Do you mean they want to emphasize something? or does it have any acronym?

zuiderkwast (approved):
I don't know why BE was capitalized.

Mistake? Maybe.

Acronym? I don't think so.

Emphasis? Maybe, but why emphasis this word? Very strange.

We can change to "be".

enjoy-binbin:
> Emphasis? Maybe, but why emphasis this word? Very strange.

i guess it is just a habit from oran. Something like a modal word. Ok, i am good with merging it.

PR #1052, opened 2024-09-19
Add --cluster option to runtest to run only cluster tests

enjoy-binbin:
Currently cluster tests in unit/cluster are run as part of
the ./runtest. Sometims we change the cluster code and only
want to run cluster tests. This PR added a --cluster option
to runtest so that we can run only cluster tests.

### DIFF
--- a/tests/test_helper.tcl
+++ b/tests/test_helper.tcl
@@ -563,7 +563,7 @@ proc send_data_packet {fd status data {elapsed 0}} {
 
 proc print_help_screen {} {
     puts [join {
-        "--cluster          Run the cluster tests, by default cluster tests are run along with all tests."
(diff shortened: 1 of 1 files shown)


PingXie (approved):

tests/test_helper.tcl
```diff
@@ -557,6 +563,7 @@ proc send_data_packet {fd status data {elapsed 0}} {
 
 proc print_help_screen {} {
     puts [join {
+        "--cluster          Run the cluster tests, by default cluster tests are run along with all tests."
```
hwware:
```suggestion
        "--cluster          Run the cluster tests, by default cluster tests run along with all tests."
```

hwware (approved):
a minor grammar fix

Issue #1054, opened 2024-09-19
[CRASH] When loading module-oss (RediSearch with Coordinator), Valkey Node crashes


wernermorgenstern:
### Crash report

```shell
=== VALKEY BUG REPORT START: Cut & paste starting from here ===
 1:M 19 Sep 2024 17:36:10.666 # valkey 8.0.0 crashed by signal: 11, si_code: 1
 1:M 19 Sep 2024 17:36:10.666 # Accessing address: (nil)
 1:M 19 Sep 2024 17:36:10.666 # Crashed running the instruction at: 0xaaaad4294510
 
 ------ STACK TRACE ------
 EIP:
 valkey-server 0.0.0.0:6379 [cluster](clusterCommandSlots+0x20)[0xaaaad4294510]
 
 136 bio_lazy_free
 /lib/aarch64-linux-gnu/libc.so.6(+0x7b698)[0xffffb14cb698]
 /lib/aarch64-linux-gnu/libc.so.6(pthread_cond_wait+0x200)[0xffffb14ce1d0]
 valkey-server 0.0.0.0:6379 [cluster](bioProcessBackgroundJobs+0x17c)[0xaaaad42b5060]
 /lib/aarch64-linux-gnu/libc.so.6(+0x7ee30)[0xffffb14cee30]
 /lib/aarch64-linux-gnu/libc.so.6(+0xe7adc)[0xffffb1537adc]
 
 135 bio_aof
 /lib/aarch64-linux-gnu/libc.so.6(+0x7b698)[0xffffb14cb698]
 /lib/aarch64-linux-gnu/libc.so.6(pthread_cond_wait+0x200)[0xffffb14ce1d0]
 valkey-server 0.0.0.0:6379 [cluster](bioProcessBackgroundJobs+0x17c)[0xaaaad42b5060]
 /lib/aarch64-linux-gnu/libc.so.6(+0x7ee30)[0xffffb14cee30]
 /lib/aarch64-linux-gnu/libc.so.6(+0xe7adc)[0xffffb1537adc]
 
 134 bio_close_file
 /lib/aarch64-linux-gnu/libc.so.6(+0x7b698)[0xffffb14cb698]
 /lib/aarch64-linux-gnu/libc.so.6(pthread_cond_wait+0x200)[0xffffb14ce1d0]
 valkey-server 0.0.0.0:6379 [cluster](bioProcessBackgroundJobs+0x17c)[0xaaaad42b5060]
[...]

PR #1057, opened 2024-09-19
Change return value of aeTimeProc callback function to long long.

Shivshankar-Reddy:
moduleTimerHandler is aeTimeProc handler and event loop gets created with this. However, found that the function return type is int but actually returns "long long" value(i.e., next_period). and return value being assigned to int variable in processTimeEvents(where time events are processed), this might cause an overflow of the timer values. So changed the return type of the function to long long. And also updated other callback function return type to be consistent.

I found this when I was checking functions reported in https://github.com/valkey-io/valkey/issues/1054 issue stacktrace. (FYI, this is just to update the return type to be consistent and it will not the fix for the issue reported)

### DIFF
--- a/src/ae.c
+++ b/src/ae.c
@@ -363,7 +363,7 @@ static int processTimeEvents(aeEventLoop *eventLoop) {
         }
 
         if (te->when <= now) {
-            int retval;
+            long long retval;
 
             id = te->id;
             te->refcount++;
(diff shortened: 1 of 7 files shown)


madolson (approved):
Seems not particularly worth mentioning in release notes, but let me know if you think otherwise.

PR #1058, opened 2024-09-20
Add log about old primary after myself failover

enjoy-binbin:
Sometims it is hard to see the old primary during a
multi primaries failover, adding this log can help
use to find the old primary node.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -4476,7 +4476,7 @@ void clusterFailoverReplaceYourPrimary(void) {
 
     if (clusterNodeIsPrimary(myself) || old_primary == NULL) return;
 
-    serverLog(LL_NOTICE, "Setting myself to primary after failover, my old primary is %.40s", old_primary->name);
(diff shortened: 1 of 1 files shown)


src/cluster_legacy.c
```diff
@@ -4476,7 +4476,9 @@ void clusterFailoverReplaceYourPrimary(void) {
 
     if (clusterNodeIsPrimary(myself) || old_primary == NULL) return;
 
-    /* 1) Turn this node into a primary . */
+    serverLog(LL_NOTICE, "Setting myself to primary after failover, my old primary is %.40s", old_primary->name);
```
PingXie: We now always log `human_nodename`. I feel that logging the shard_id could be helpful too. I am also wondering if it helps to log `myself`'s node_id. Doing so would make the log statement self-sufficient so we know everything about this event without checking out other parts of the log.

Thoughts?

```suggestion
    serverLog(LL_NOTICE, "Setting myself to primary in shard %.40s after failover; my old primary is %.40s (%s)", myself->shard_id, old_primary->name, old_primary->human_nodename);
```

enjoy-binbin: SGTM.

PingXie (approved):

PR #1060, opened 2024-09-20
Fix timing issue in the new tot-net-out replica test

enjoy-binbin:
Apparently there is a timing issue when using wait_for_ofs_sync:
```
[exception]: Executing test client: can't read "out_before": no such variable.
can't read "out_before": no such variable
```

The reason is that if the connection between the primary
and the replica is not established yet, the master_repl_offset
of the primary and replica in wait_for_ofs_sync is 0, and
the check fails, resulting in no replica client in the
client list below.

In this case, we need to make sure the replica is online
before proceeding.

### DIFF
--- a/tests/unit/introspection.tcl
+++ b/tests/unit/introspection.tcl
@@ -25,7 +25,7 @@ start_server {tags {"introspection"}} {
             set kv [split $item "="]
             set k [lindex $kv 0]
             if {[string match $field $k]} {
-                return [lindex $kv 1]   
+                return [lindex $kv 1]
             }
         }
         return ""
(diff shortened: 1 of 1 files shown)


enjoy-binbin:
the failed ci link: https://github.com/valkey-io/valkey/actions/runs/10950803684/job/30406733305#step:6:723

tests/unit/introspection.tcl
```diff
@@ -845,22 +844,19 @@ start_server {tags {"introspection"}} {
...
                 set replica_pid [srv 0 pid]
-    
+
                 $replica replicaof $primary_host $primary_port
-    
+
                 # Wait for replica to be connected before proceeding.
-                wait_for_ofs_sync $primary $replica
-                
+                wait_replica_online $primary
```
enjoy-binbin, line 851: the real change is in here, the other part just a space format by Clion.

tests/unit/introspection.tcl
```diff
@@ -845,22 +844,19 @@ start_server {tags {"introspection"}} {
...
+
                 # Avoid PINGs to make sure tot-net-out is stable.
                 $primary config set repl-ping-replica-period 3600
 
                 # Increase repl timeout to avoid replica disconnecting
                 $primary config set repl-timeout 3600
                 $replica config set repl-timeout 3600
-                
-                # Wait for the replica to receive the command.
-                wait_for_ofs_sync $primary $replica
```
enjoy-binbin, line 862: and here of course, this one is no needed anymore.

PingXie (approved):

PR #1061, opened 2024-09-20
Check whether to switch to fail when setting the node to pfail in cron

enjoy-binbin:
This may speed up the transition to the fail state a bit.
Previously we would only check when we received a pfail/fail
report from others in gossip. If myself is the last vote,
we can directly switch to fail in here without waiting for
the next gossip packet.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -5066,7 +5066,7 @@ void clusterCron(void) {
             if (!(node->flags & (CLUSTER_NODE_PFAIL | CLUSTER_NODE_FAIL))) {
                 node->flags |= CLUSTER_NODE_PFAIL;
                 update_state = 1;
-                if (server.cluster->size == 1 && clusterNodeIsVotingPrimary(myself)) {
+                if (clusterNodeIsVotingPrimary(myself)) {
                     markNodeAsFailingIfNeeded(node);
                 } else {
                     serverLog(LL_NOTICE, "NODE %.40s (%s) possibly failing.", node->name, node->human_nodename);


src/cluster_legacy.c
```diff
@@ -5066,7 +5066,7 @@ void clusterCron(void) {
             if (!(node->flags & (CLUSTER_NODE_PFAIL | CLUSTER_NODE_FAIL))) {
                 node->flags |= CLUSTER_NODE_PFAIL;
                 update_state = 1;
-                if (server.cluster->size == 1 && clusterNodeIsVotingPrimary(myself)) {
+                if (clusterNodeIsVotingPrimary(myself)) {
```
PingXie, line 5069: This change makes a lot of sense to me. I actually don't fully understand why we special-cased single-shard clusters in https://github.com/valkey-io/valkey/commit/b3aaa0a1362d229ba1ecd44629655f76c77304ec

enjoy-binbin: Maybe it was just to reduce the impact (to solve a specific problem (single shard) at the time). @bentotten would you be able to take a review?

bentotten: For single-shard clusters, there is no need to wait for a quorum of 1, so we can proceed to directly fail the node here and not wait for the original gossip-logic [(for reference)](https://github.com/valkey-io/valkey/blame/unstable/src/cluster_legacy.c#L2256). Unfortunately prior to that, a single-shard cluster would never properly mark a replica as failed.

I believe we still need to wait to achieve a quorum from gossip to mark this node as truly failed, no? If the node uses the information it has from "old gossip" during each cluster cron run, wont this increase the number of calls to this function without increasing the accuracy of the failure report?

bentotten: Though the benefit of the node being able to mark the replica as failed as soon as it knows it cant reach said replica in the last-needed-vote scenario also makes sense - @madolson I know you did the investigation into the original bug, if you have any insights. This was not a scenario I had considered for the original commit, and dont see any issue

enjoy-binbin: right, thanks for the input, that is new to me. clusterNodeFailureReportsCount will cleanup the old failure report, so i think (hope) it can avoid the old gossip information risk. But you do make a good point, let see if others have a new thought.

zuiderkwast: Sorry, I don't have any useful input about this scenario. You know it better than me.

enjoy-binbin: no worry, it is always good to have an extra eyes on it.

PingXie: > If the node uses the information it has from "old gossip" during each cluster cron run, wont this increase the number of calls to this function without increasing the accuracy of the failure report?

Yes, discounting "expired" failure reports is part of the `markNodeAsFailingIfNeeded`'s contract.

PingXie (approved):

PR #1064, opened 2024-09-23
Print an empty primary log when primary lost its last slot

enjoy-binbin:
The one in CLUSTER SETSLOT help us keep track of state better,
of course it also can make the test case happy.

The one in gossip process fixes a problem that a replica can
print a log saying it is an empty primary.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -6443,8 +6443,8 @@ void clusterCommandSetSlot(client *c) {
             clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_UPDATE_STATE | CLUSTER_TODO_FSYNC_CONFIG);
         }
 
-        /* If replica migration is not allow, check if the primary of this shard
-         * loses its last slot and the shard become empty. In this case, we will
+        /* If replica migration is not allowed, check if the primary of this shard
+         * loses its last slot and the shard becomes empty. In this case, we will
          * print the exact same log as during the gossip process. */
         if (!server.cluster_allow_replica_migration && nodeIsPrimary(myself) && slot_was_mine && shard_is_empty) {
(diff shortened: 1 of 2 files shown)


tests/unit/cluster/replica-migration.tcl
```diff
@@ -372,7 +372,7 @@ proc test_cluster_setslot {type} {
             fail "valkey-cli --cluster rebalance returns non-zero exit code, output below:\n$result"
         }
 
-        # Wait for R 3 to report that it is an empty replica (cluster-allow-replica-migration no)
+        # Wait for R 3 to report that it is an empty primary (cluster-allow-replica-migration no)
         wait_for_log_messages -3 {"*I am now an empty primary*"} 0 1000 50
```
enjoy-binbin, line 376: https://github.com/valkey-io/valkey/actions/runs/10985498040/job/30497439660#step:8:5822

Add a log because this, or we can remove this check in the test case.

src/cluster_legacy.c
```diff
@@ -6442,6 +6443,16 @@ void clusterCommandSetSlot(client *c) {
             clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_UPDATE_STATE | CLUSTER_TODO_FSYNC_CONFIG);
         }
 
+        /* If replica migration is not allow, check if the primary of this shard
```
```suggestion
        /* If replica migration is not allowed, check if the primary of this shard
```

src/cluster_legacy.c
```diff
@@ -6442,6 +6443,16 @@ void clusterCommandSetSlot(client *c) {
...
+         * loses its last slot and the shard become empty. In this case, we will
```
```suggestion
         * loses its last slot and the shard becomes empty. In this case, we will
```

PingXie (approved):

PR #1065, opened 2024-09-23
Replace dict with hashset in command tables

rainsupreme:
This changes the type of command tables from dict to hashset. Command table lookup takes ~3% of overall CPU time in benchmarks, so it is a good candidate for optimization.

My initial SET benchmark comparison suggests that hashset is about 4.5 times faster than dict and this replacement reduced overall CPU time by 2.79% 🥳

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -300,8 +300,7 @@ size_t dictSdsEmbedKey(unsigned char *buf, size_t buf_len, const void *key, uint
     return sdscopytobuffer(buf, buf_len, (sds)key, key_offset);
 }
 
-/* A case insensitive version used for the command lookup table and other
- * places where case insensitive non binary-safe comparison is needed. */
+/* Case insensitive non binary-safe comparison */
 int dictSdsKeyCaseCompare(dict *d, const void *key1, const void *key2) {
     UNUSED(d);
     return strcasecmp(key1, key2) == 0;
@@ -436,13 +435,13 @@ int dictResizeAllowed(size_t moreMem, double usedRatio) {
     }
 }
 
-const void* hashsetCommandGetKey(const void *element) {
-    struct serverCommand* command = (struct serverCommand*) element;
+const void *hashsetCommandGetKey(const void *element) {
+    struct serverCommand *command = (struct serverCommand *)element;
     return command->fullname;
 }
 
-const void* hashsetSubcommandGetKey(const void *element) {
-    struct serverCommand* command = (struct serverCommand*) element;
+const void *hashsetSubcommandGetKey(const void *element) {
+    struct serverCommand *command = (struct serverCommand *)element;
     return command->declared_name;
 }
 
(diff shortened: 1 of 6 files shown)


src/server.h
```diff
@@ -67,6 +67,7 @@ typedef long long ustime_t; /* microsecond time type. */
 #include "ae.h"         /* Event driven programming library */
 #include "sds.h"        /* Dynamic safe strings */
 #include "dict.h"       /* Hash tables */
+#include "hashset.h"    /* Hash set */
```
hpatro, line 70: Is the hashset.c/.h file in a separate commit/PR ? I don't see the file in unstable. Or am I missing something ?

rainsupreme: This is targeting the hashset branch which has a lot of work done by Viktor. :)

hpatro: Ah, my bad.

src/server.c
```diff
@@ -307,6 +307,12 @@ int dictSdsKeyCaseCompare(dict *d, const void *key1, const void *key2) {
     return strcasecmp(key1, key2) == 0;
 }
 
+/* Case insensitive key comparison */
+int hashsetStringKeyCaseCompare(hashset *hs, const void *key1, const void *key2) {
```
zuiderkwast, line 310: The function dictSdsKeyCaseCompare above has this comment which is now outdated:

```
/* A case insensitive version used for the command lookup table and other
 * places where case insensitive non binary-safe comparison is needed. */
```

It's not used for the command lookup table anymore.

src/server.c
```diff
@@ -502,15 +518,32 @@ dictType kvstoreExpiresDictType = {
...
+    dictSdsCaseHash,            // hash function
+    hashsetStringKeyCaseCompare,// case insensitive char* key compare
+    NULL,                       // item destructor
+    NULL,                       // resize allowed callback
+    NULL,                       // rehash start callback
+    NULL,                       // rehash end callback
+    NULL,                       // metadata size callback
+    .instant_rehashing = 1,
+    .userdata = NULL
+};
```
zuiderkwast: It's better to use the `.name = value` style everywhere in these types. Then we don't need the comments, we can skip all the nulls and they don't need to be in a fixed order.

```suggestion
hashsetType commandSetType = {
    .elementGetKey = hashsetCommandGetKey,
    .hashFunction = dictSdsCaseHash,
    .keyCompare = hashsetStringKeyCaseCompare,
    .instant_rehashing = 1,
};
```

src/server.h
```diff
@@ -2558,7 +2559,7 @@ struct serverCommand {
                                     * still maintained (if applicable) so that
                                     * we can still support the reply format of
                                     * COMMAND INFO and COMMAND GETKEYS */
-    dict *subcommands_dict;        /* A dictionary that holds the subcommands, the key is the subcommand sds name
+    hashset *subcommands_dict;        /* A set that holds the subcommands, the key is the subcommand sds name
```
zuiderkwast: The name suggests that it's a dict. We can call it `subcommands_set` or just `subcommands`.

zuiderkwast (commented):
Looks awesome! Just a few minor comments. I think clang-format has a few comments too.

The other CI failures are about the unit tests. We should fix them in the hashset branch.

zuiderkwast:
Oops.

```
server.h:2543:14: error: duplicate member ‘subcommands’
 2543 |     hashset *subcommands;          /* A set that holds the subcommands, the key is the subcommand sds name
      |              ^~~~~~~~~~~
```

The other one with the same name is

```C
    /* Array of subcommands (may be NULL) */
    struct serverCommand *subcommands;
```

I don't know why we need this array when we also have the hashset. Anyway, the easiest solution is to avoid the same name. :)

zuiderkwast (approved):
Great. There's some failure in a unit test, but it's not from this PR so I'm going to merge this. Thanks!

PR #1067, opened 2024-09-24
Trigger the election as soon as possible when doing a forced manual failover

enjoy-binbin:
In CLUSTER FAILOVER FORCE case, we will set mf_can_start to
1 and wait for a cron to trigger the election. We can also set
a CLUSTER_TODO_HANDLE_MANUALFAILOVER flag so that we can start
the election as soon as possible instead of waiting for the cron,
so that we won't have a 100ms delay (clusterCron).


PingXie (approved):
LGTM

PR #1072, opened 2024-09-25
Fix bug for CLUSTER SLOTS from EVAL over TLS

zuiderkwast:
For fake clients like the ones used for Lua and modules, we don't determine TLS in the right way, causing CLUSTER SLOTS from EVAL over TLS to fail a debug-assert.

This error was introduced when the caching of CLUSTER SLOTS was introduced, i.e. in 8.0.0.


enjoy-binbin (approved):
OK, i see. LGTM

hwware (approved):

PR #1074, opened 2024-09-25
Fix RDMA build dependence

pizhenwei:
RDMA module has dependence on '$(SERVER_NAME)' rather than the old style '$(REDIS_SERVER_NAME)'.


zuiderkwast:
So this causes a problem when building with -j10. Let's include it in the next patch release.

zuiderkwast (approved):

PR #1077, opened 2024-09-26
Avoid timing issue in diskless-load-swapdb test

enjoy-binbin:
Since we paused the primary node earlier, the replica may enter
cluster down due to primary node pfail. Here set allow read to
prevent subsequent read errors.

### DIFF
--- a/tests/unit/cluster/diskless-load-swapdb.tcl
+++ b/tests/unit/cluster/diskless-load-swapdb.tcl
@@ -78,6 +78,11 @@ test "Main db not affected when fail to diskless load" {
         fail "Fail to stop the full sync"
     }
 
+    # Since we paused the primary node earlier, the replica may enter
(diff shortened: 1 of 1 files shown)


enjoy-binbin:
failed CI:
https://github.com/valkey-io/valkey/actions/runs/10987837289/job/30503299419#step:5:4849
```
[exception]: Executing test client: CLUSTERDOWN The cluster is down.
CLUSTERDOWN The cluster is down
    while executing
"$replica get $slot0_key"
    ("uplevel" body line 70)
    invoked from within
"uplevel 1 $code"
    (procedure "test" line 58)
    invoked from within
"test "Main db not affected when fail to diskless load" {
```

madolson (approved):

zuiderkwast:
7.2 doesn't have this test suite, so we can skip it.

PR #1078, opened 2024-09-26
Try to stabilize the failover call in the slot migration test

enjoy-binbin:
The CI report replica will return the error when performing CLUSTER FAILOVER:
-ERR Master is down or failed, please use CLUSTER FAILOVER FORCE

This may because the primary state is fail or the cluster connection
is disconnected during the primary pause. In this PR, we added some
waits in wait_for_role, if the role is replica, we will wait for the
replication link and the cluster link to be ok.

### DIFF
--- a/tests/support/cluster_util.tcl
+++ b/tests/support/cluster_util.tcl
@@ -277,6 +277,14 @@ proc cluster_get_myself id {
     return {}
 }
 
+# Returns the parsed "myself's primary" node entry as a dictionary.
+proc cluster_get_myself_primary id {
+    set myself [cluster_get_myself $id]
+    set replicaof [dict get $myself slaveof]
+    set node [cluster_get_node_by_id $id $replicaof]
+    return $node
+}
+
 # Get a specific node by ID by parsing the CLUSTER NODES output
 # of the instance Number 'instance_id'
 proc cluster_get_node_by_id {instance_id node_id} {
--- a/tests/support/cluster_util.tcl
+++ b/tests/support/cluster_util.tcl
@@ -277,7 +277,7 @@ proc cluster_get_myself id {
     return {}
 }
 
-# Returns the parsed "myself's primary" node entry as a dictionary.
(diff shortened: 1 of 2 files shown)


enjoy-binbin:
CI link: https://github.com/valkey-io/valkey/actions/runs/11028380661/job/30628495083?pr=1073#step:5:3785

```
[exception]: Executing test client: ERR Master is down or failed, please use CLUSTER FAILOVER FORCE.
ERR Master is down or failed, please use CLUSTER FAILOVER FORCE
    while executing
"[Rn $n] {*}$args"
    (procedure "R" line 2)
    invoked from within
"R 1 cluster failover"
    ("uplevel" body line 12)
    invoked from within
"uplevel 1 $code"
    (procedure "test" line 58)
    invoked from within
"test "Migration target is auto-updated after failover in target shard" {
        # Trigger an auto-failover from R1 to R4
        fail_server 1
```

enjoy-binbin:
@PingXie Do you have time to check this out? (Or just simply approve it since it is not easy to figure out the test) We used to have frequent test issues in this one but they seem to be gone recently. I wonder if we have some other PRs that has implictly fix it? And in addition, this content should be harmless and can stablize the test, don't want to waste the efforts.

tests/support/cluster_util.tcl
```diff
@@ -277,6 +277,14 @@ proc cluster_get_myself id {
     return {}
 }
 
+# Returns the parsed "myself's primary" node entry as a dictionary.
```
zuiderkwast: I guess all of these functions are about CLUSTER NODES? It can be useful to mention it.

```suggestion
# Returns the parsed "myself's primary" CLUSTER NODES entry as a dictionary.
```

zuiderkwast (approved):
Looks like good protection for flaky tests.

tests/support/cluster_util.tcl
enjoy-binbin: yes

PR #1081, opened 2024-09-28
Trigger the election immediately when doing a manual failover

enjoy-binbin:
Currently when a manual failover is triggeded, we will set a
CLUSTER_TODO_HANDLE_FAILOVER to start the election as soon as
possible in the next beforeSleep. But in fact, we won't delay
the election in manual failover, waitting for the next beforeSleep
to kick in will delay the election a some milliseconds.

We can trigger the election immediately in this case in the
same function call, without waitting for beforeSleep, which
can save us some milliseconds.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -4515,8 +4515,9 @@ void clusterFailoverReplaceYourPrimary(void) {
  * 3) Perform the failover informing all the other nodes.
  */
 void clusterHandleReplicaFailover(void) {
+    mstime_t now = mstime();
     mstime_t data_age;
-    mstime_t auth_age = mstime() - server.cluster->failover_auth_time;
+    mstime_t auth_age = now - server.cluster->failover_auth_time;
     int needed_quorum = (server.cluster->size / 2) + 1;
     int manual_failover = server.cluster->mf_end != 0 && server.cluster->mf_can_start;
     mstime_t auth_timeout, auth_retry_time;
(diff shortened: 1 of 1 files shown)


src/cluster_legacy.c
```diff
@@ -4590,20 +4591,26 @@ void clusterHandleReplicaFailover(void) {
...
         /* Now that we have a scheduled election, broadcast our offset
          * to all the other replicas so that they'll updated their offsets
          * if our offset is better. */
         clusterBroadcastPong(CLUSTER_BROADCAST_LOCAL_REPLICAS);
-        return;
+
+        /* Return ASAP if we can't start the election. Doing this allow us to, for example,
+         * a manual failover, we can get it to the next state ASAP instead of waiting for
+         * the next beforeSleep to kick in. */
+        if (now < server.cluster->failover_auth_time) return;
```
zuiderkwast: "Doing this" refers to "return ASAP". But I think you mean "not returning ASAP" allows manual failover to not wait for next beforeSleep.

Idea:

```suggestion
        /* Return ASAP if we can't start the election now. In a manual failover,
         * we can start the election immediately, so in this case we continue to
         * the next state without waiting for the next beforeSleep. */
        if (now < server.cluster->failover_auth_time) return;
```

zuiderkwast (approved):

src/cluster_legacy.c
enjoy-binbin: good, it is better, thanks

PR #1084, opened 2024-09-29
Add server log when module load fails with busy name

enjoy-binbin:
Currently when module loading fails due to busy name, we
don't have a clean way to assist to troubleshooting.

Case 1: when loading the same module multiple times, we can
not detemine the cause of its failure without referring to
the module list or the earliest module load log. The log
may not exist and sometimes it is difficult for people
to associate module list.

Case 2: when multiple modules use the same module name,
we can not quickly associate the busy name without referring
to the module list and the earliest module load log.
Different people wrote modules with the same module name,
they don't easily associate module name.

So in this PR, when doing module onload, we will try to
print a busy name log if this happen. Currently we check
ctx.module since if it is NULL it means the Init call
failed, and Init currently only fails with busy name.

It's kind of ugly. It would have been nice if we could have
had a better way for onload to signal why the load failed.

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -12204,7 +12204,7 @@ int moduleLoad(const char *path, void **module_argv, int module_argc, int is_loa
         } else {
             /* If there is no ctx.module, this means that our ValkeyModule_Init call failed,
              * and currently init will only fail on busy name. */
-            serverLog(LL_WARNING, "Module %s initialization failed. Module name is busy", path);
+            serverLog(LL_WARNING, "Module %s initialization failed. Module name is busy.", path);
         }
         moduleFreeContext(&ctx);
         dlclose(handle);
--- a/src/module.c
+++ b/src/module.c
@@ -12196,11 +12196,15 @@ int moduleLoad(const char *path, void **module_argv, int module_argc, int is_loa
     ValkeyModuleCtx ctx;
     moduleCreateContext(&ctx, NULL, VALKEYMODULE_CTX_TEMP_CLIENT); /* We pass NULL since we don't have a module yet. */
     if (onload((void *)&ctx, module_argv, module_argc) == VALKEYMODULE_ERR) {
-        serverLog(LL_WARNING, "Module %s initialization failed. Module not loaded", path);
         if (ctx.module) {
+            serverLog(LL_WARNING, "Module %s initialization failed. Module not loaded", path);
             moduleUnregisterCleanup(ctx.module);
             moduleRemoveCateogires(ctx.module);
             moduleFreeModuleStructure(ctx.module);
+        } else {
+            /* If there is no ctx.module, this means that our ValkeyModule_Init call failed,
+             * and currently init will only fail on busy name. */
+            serverLog(LL_WARNING, "Module %s initialization failed. Module name is busy", path);
         }
         moduleFreeContext(&ctx);
         dlclose(handle);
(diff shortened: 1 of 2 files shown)


src/module.c
```diff
@@ -12196,11 +12196,15 @@ int moduleLoad(const char *path, void **module_argv, int module_argc, int is_loa
     ValkeyModuleCtx ctx;
     moduleCreateContext(&ctx, NULL, VALKEYMODULE_CTX_TEMP_CLIENT); /* We pass NULL since we don't have a module yet. */
     if (onload((void *)&ctx, module_argv, module_argc) == VALKEYMODULE_ERR) {
-        serverLog(LL_WARNING, "Module %s initialization failed. Module not loaded", path);
         if (ctx.module) {
+            serverLog(LL_WARNING, "Module %s initialization failed. Module not loaded", path);
```
zuiderkwast: Good but it is missing a "." in the end of the second sentence. Perhaps we can swap the sentences too? It seems more clear in the order: 1 what, 2 why.

```suggestion
            serverLog(LL_WARNING, "Module %s not loaded. Initialization failed.", path);
```

src/module.c
```diff
@@ -12196,11 +12196,15 @@ int moduleLoad(const char *path, void **module_argv, int module_argc, int is_loa
...
             moduleUnregisterCleanup(ctx.module);
             moduleRemoveCateogires(ctx.module);
             moduleFreeModuleStructure(ctx.module);
+        } else {
+            /* If there is no ctx.module, this means that our ValkeyModule_Init call failed,
+             * and currently init will only fail on busy name. */
+            serverLog(LL_WARNING, "Module %s initialization failed. Module name is busy", path);
```
zuiderkwast: To match the suggestion above, if you like it.

```suggestion
            serverLog(LL_WARNING, "Module %s not loaded. Module name is busy.", path);
```

zuiderkwast (approved):
Great! Just a suggestion about how we write the log messages. Not sure if it's a breaking change to change a log message, is it?

hwware (approved):

hwware:
Yes, this is a great help for identifying the error reason when loading a module failed

src/module.c
enjoy-binbin: yes, it is more clear in this way. But we have other places like this...

```
    if (onload == NULL) {
        dlclose(handle);
        serverLog(LL_WARNING,
                  "Module %s does not export ValkeyModule_OnLoad() or RedisModule_OnLoad() "
                  "symbol. Module not loaded.",
                  path);
        return C_ERR;
    }

        if (unload_status == VALKEYMODULE_ERR) {
            serverLog(LL_WARNING, "Module %s OnUnload failed. Unload canceled.", name);
            errno = ECANCELED;
            return C_ERR;
        }
```

zuiderkwast: Ah, ok. Then just add the missing dot?

```suggestion
            serverLog(LL_WARNING, "Module %s initialization failed. Module not loaded.", path);
```

src/module.c
```diff
@@ -12196,11 +12196,15 @@ int moduleLoad(const char *path, void **module_argv, int module_argc, int is_loa
...
-        serverLog(LL_WARNING, "Module %s initialization failed. Module not loaded", path);
         if (ctx.module) {
+            serverLog(LL_WARNING, "Module %s initialization failed. Module not loaded.", path);
             moduleUnregisterCleanup(ctx.module);
             moduleRemoveCateogires(ctx.module);
             moduleFreeModuleStructure(ctx.module);
+        } else {
+            /* If there is no ctx.module, this means that our ValkeyModule_Init call failed,
+             * and currently init will only fail on busy name. */
+            serverLog(LL_WARNING, "Module %s initialization failed. Module name is busy", path);
```
enjoy-binbin:
```suggestion
            serverLog(LL_WARNING, "Module %s initialization failed. Module name is busy.", path);
```

zuiderkwast (approved):

madolson (approved):
It's kind of ugly. It would have been nice if we could have had a better way for `onload` to signal why the load failed. I don't see another way to handle it given how the initialization code works though.

PR #1089, opened 2024-09-30
RDMA: Support .is_local method

pizhenwei:
There is no ethernet style virtual device (like lo 127.0.0.1) for RDMA, however a connection with the same local address and peer address are considered as local.

### DIFF
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -1501,6 +1501,26 @@ end:
     return ret;
 }
 
+static int connRdmaIsLocal(connection *conn) {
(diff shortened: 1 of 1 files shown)


enjoy-binbin (approved):

zuiderkwast (approved):

zuiderkwast:
It affects protected mode? Anything else?

pizhenwei:
> It affects protected mode? Anything else?

As far as I know, it affects protected mode and protected config.

PR #1091, opened 2024-09-30
Trigger manual failover on SIGTERM / shutdown to cluster primary

### LINKED ISSUE #939: [NEW] Trigger manual failover on SIGTERM to primary (cluster)

When a primary disappears, its slots are not served until an automatic failover happens. It takes about 3 seconds (node timeout plus some second). It's too much time for us to not accept writes.

If the host machine is about to shutdown for any reason, the processes typically get a sigterm and have some time to shutdown gracefully. In Kubernetes, this is 30 seconds by default.

When a primary receives a SIGTERM, let it trigger a failover to one of the replicas as part of the graceful shutdown.

Our current solution is to have a wrapper process, a small script, starting valkey. Then, this process receives the SIGTERM and can handle this. It's like a workaround though. It's better to have this built in.

**Additional information**

We use cluster.

enjoy-binbin:
## update

The `auto-failover-on-shutdown` was removed in #2292, please see #2292 for more details.

========================================

When a primary disappears, its slots are not served until an automatic
failover happens. It takes about n seconds (node timeout plus some seconds).
It's too much time for us to not accept writes.

If the host machine is about to shutdown for any reason, the processes
typically get a sigterm and have some time to shutdown gracefully. In
Kubernetes, this is 30 seconds by default.

When a primary receives a SIGTERM or a SHUTDOWN, let it trigger a failover
to one of the replicas as part of the graceful shutdown. This can reduce
some unavailability time. For example the replica needs to sense the
primary failure within the node-timeout before initating an election,
and now it can initiate an election quickly and win and gossip it.

The primary does this by sending a CLUSTER FAILOVER command to the replica.
We added a replicaid arg to CLUSTER FAILOVER, after receiving the command,
the replica will check whether the node-id is itself, if not, the command
will be ignored. The node-id is set by the replica through client setname
during the replication handshake.

### New argument for CLUSTER FAILOVER

So the format now become CLUSTER FAILOVER [FORCE TAKEOVER] [REPLICAID node-id],
this arg does not intented for user use, so it will not be added to the JSON
file.

### Replica sends REPLCONF SET-CLUSTER-NODE-ID to inform its node-id

During the replication handshake, replica now will use REPLCONF SET-CLUSTER-NODE-ID
to inform the primary of replica node-id.

### Primary issue CLUSTER FAILOVER

Primary sends CLUSTER FAILOVER FORCE REPLICAID node-id to all replicas because
it is a shared replication buffer but only the replica with the mathching id
will execute it.

### Add a new auto-failover-on-shutdown config

People can disable this feature if they don't like it, the default is 0.

This closes #939.


src/server.c
```diff
@@ -4419,6 +4428,21 @@ int finishShutdown(void) {
      * send them pending writes. */
     flushReplicasOutputBuffers();
 
+    if (server.auto_failover_on_shutdown && server.cluster_enabled && best_replica) {
+        /* Sending a CLUSTER FAILOVER FORCE to the best replica. */
+        const char *buf = "*3\r\n$7\r\nCLUSTER\r\n$8\r\nFAILOVER\r\n$5\r\nFORCE\r\n";
+        if (connWrite(best_replica->conn, buf, strlen(buf)) == (int)strlen(buf)) {
```
zuiderkwast: This is in `finishShutdown()`, just before the primary does `exit()`.

Is there a risk that the written command has not yet been fully sent to the replica when the primary exits? If not, can we do it in an earlier shutdown stage and make sure the replica has received the command? The replica doesn't send any OK reply to the primary on the replication stream, but maybe the primary could wait for the failover auth request from the replica on the cluster bus before the primary shuts down?

Btw, maybe this code should be in `clusterHandleServerShutdown()` which is called a few lines below:

    /* Handle cluster-related matters when shutdown. */
    if (server.cluster_enabled) clusterHandleServerShutdown();

src/server.c
```diff
@@ -4419,6 +4428,21 @@ int finishShutdown(void) {
...
+            serverLog(LL_NOTICE, "Sending CLUSTER FAILOVER FORCE to replica %s succeeded.",
+                      replicationGetReplicaName(best_replica));
+        } else {
+            serverLog(LL_WARNING, "Failed to send CLUSTER FAILOVER FORCE to replica: %s", strerror(errno));
+        }
+    }
+
+    if (server.auto_failover_on_shutdown && server.cluster_enabled && !best_replica) {
+        serverLog(LL_WARNING, "Unable to find a replica to perform an auto failover on shutdown.");
+    }
```
zuiderkwast: Probably this should be an `else` to the above, so we avoid repeating the same logic.

tests/unit/cluster/auto-failover-on-shutdown.tcl
```diff
@@ -0,0 +1,68 @@
...
+        catch {R $srv_id shutdown nosave}
+    } elseif {$how == "sigterm"} {
+        exec kill -SIGTERM [s -$srv_id process_id]
+    }
+}
+
+proc test_main {how} {
+    test "auto-failover-on-shutdown will always pick a best replica and send CLUSTER FAILOVER - $how" {
+        set primary [srv 0 client]
+        set replica1 [srv -3 client]
```
zuiderkwast: This function requires some nodes to be primary and some to be replica. Please add some comment about how the cluster needs to be initiated for this function to work.

tests/unit/cluster/auto-failover-on-shutdown.tcl
```diff
@@ -0,0 +1,68 @@
...
+
+        # Wait the replica2 catch up with the offset
+        wait_replica_acked_ofs $primary $replica2_ip $replica2_port
+
+        # Shutdown the primary.
+        shutdown_on_how 0 $how
+
+        # Wait for the replica2 to become a primary.
+        wait_for_condition 1000 50 {
+            [s -6 role] eq {master}
```
zuiderkwast, line 50: Can we make sure we wait shorter time than the node timeout, so we actually test that this is faster than an automatic failover? This is 50 seconds, right? We can maybe reduce it to 10 seconds and we can set node timeout of the nodes to for example 20 seconds when we start them?

src/cluster_legacy.c
```diff
@@ -4592,6 +4596,11 @@ void clusterHandleReplicaFailover(void) {
         if (server.cluster->mf_end) {
             server.cluster->failover_auth_time = mstime();
             server.cluster->failover_auth_rank = 0;
+            /* todo: see if this is needed. */
+            /* This is a failover triggered by my primary, let's counts its vote. */
+            if (server.cluster->mf_is_primary_failover) {
+                server.cluster->failover_auth_count++;
+            }
```
zuiderkwast: It's not very good that we add some conditional logic about how we are counting the votes. It makes the voting algorithm more complex to analyze.

I understand we need this special case if the primary exits immediately after sending CLUSTER FAILOVER FORCE. Maybe we can avoid it if the primary waits for the failover auth requests and actually votes, before the primary shuts down..?

zuiderkwast (commented):
Nice! Thanks for doing this.

The PR description can be updated to explain the solution. Now it is just copy-pasted from the issue. :)

I'm thinking that doing failover in finishShutdown() is maybe too late. `finishShutdown` is only called when all replicas already have replication offset equal to the primary (checked by `isReadyToShutdown()`), or after timeout (10 seconds). If one replica is very slow, it will delay the failover. I think we can do the manual failover earlier.

This is the sequence:

1. SHUTDOWN or SIGTERM calls `prepareForShutdown()`. Here, pause clients for writing and start waiting for replicas offset.
2. In `serverCron()`, we check `isReadyToShutdown()` which checks if all replicas have `repl_ack_off == primary_repl_offset`. If yes, `finishShutdown()` is called, otherwise wait more.
3. finishShutdown.

I think we can send CLUSTER FAILOVER FORCE to the first replica which has `repl_ack_off == primary_repl_offset`. We can do it in `isReadyToShutdown()` I think. (We can rename to indicated it does more then check if ready.) Then, we also wait for it to send failover auth request and the primary votes before `isReadyToShutdown()` returns true.

What do you think?

enjoy-binbin:
> The PR description can be updated to explain the solution. Now it is just copy-pasted from the issue. :)

The issue desc is good and very detailed so i copied it, i will update it later.

> I'm thinking that doing failover in finishShutdown() is maybe too late. finishShutdown is only called when all replicas already have replication offset equal to the primary (checked by isReadyToShutdown()), or after timeout (10 seconds). If one replica is very slow, it will delay the failover. I think we can do the manual failover earlier.

yean, a failover as soon as possible is good, but itn't true that the primary is down only after it actually exit? so in this case, if a replica is slow and it does not have the chance to catch up the primary, and then the other replica trigger the failover, so the slow replica will need a full sync when it doing the reconfiguration. 

> I think we can send CLUSTER FAILOVER FORCE to the first replica which has repl_ack_off == primary_repl_offset. We can do it in isReadyToShutdown() I think. (We can rename to indicated it does more then check if ready.) Then, we also wait for it to send failover auth request and the primary votes before isReadyToShutdown() returns true.

so let me sort it out again, you are suggesting that if one replica has already caught up the offset, we should trigger a failover immediately?

I guess it is also make sense in this case.

src/server.c
enjoy-binbin: There may be this risk, i am not sure of the details, i thought connWrite would have this guarantee. Waiting the failover auth seems like a good idea, the waiting time is also determined by shutdown-timeout? In this case, i am worried that during the waitting period, the primary got a reconfigurting a become a replica, can this happen?

clusterHandleServerShutdown. right, i forget it, it is a good place to hold this logic.

tests/unit/cluster/auto-failover-on-shutdown.tcl
enjoy-binbin: We actually have verify_log_message later to check the logs

src/cluster_legacy.c
enjoy-binbin: yes, it is not very good so i add a todo. I am also not sure about this logic, and in fact we can skip this logic and the failover will also work since i think we will still have the enouth votes. I add this logic just trying open the discusstion so that we can discuss it and see if should we handle this case.

zuiderkwast:
> if a replica is slow and it does not have the chance to catch up the primary, and then the other replica trigger the failover, so the slow replica will need a full sync when it doing the reconfiguration.

I didn't think about this. The replica can't do psync to the new primary after failover? If it can't, then maybe you're right that the primary should wait for all replicas, at least for some time, to avoid full sync.

So, wait for all, then trigger manual failover. If you want, we can add another wait after that (after "finish shutdown"), so the primary can vote for the replica before exit. Wdyt?

enjoy-binbin:
Sorry for the late reply, i somehow missed this thread.

> I didn't think about this. The replica can't do psync to the new primary after failover? If it can't, then maybe you're right that the primary should wait for all replicas, at least for some time, to avoid full sync.

Yes, i think this may happen, like if the primary does not flush its output buffer to the slow replica, like primary does not write the buffer to the slow replica, when doing the reconfiguration, the slow replica may use an old offset to psync with the new primary, which will cause a full sync. This may happen, but the probability should be small since the primary will call flushReplicasOutputBuffers to write as much as possible before shutdown.

> So, wait for all, then trigger manual failover. If you want, we can add another wait after that (after "finish shutdown"), so the primary can vote for the replica before exit. Wdyt?

wait for the vote, i think both are OK. Even if we don't wait, I think the replica will have enough votes. If we really want to, we can even wait until the replica successfully becomes primary before exiting... Do you have a final decision? I will do whatever you think is right.

zuiderkwast:
> wait for the vote, i think both are OK. Even if we don't wait, I think the replica will have enough votes. If we really want to, we can even wait until the replica successfully becomes primary before exiting... Do you have a final decision? I will do whatever you think is right.

I'm thinking if there are any corner cases, like if the cluster is too small to have quorum without the shutting down primary...

If it is simple, I prefer to let the primary wait and vote. Then we can avoid the `server.cluster->mf_is_primary_failover` variable. I don't like this variable and special case. :)

But if this implementation to wait for the vote will be too complex, then let's just skip the vote. I think it's also fine. Without this feature, we wait for automatic failover, which will also not have the vote from the already shutdown primary.

enjoy-binbin:
> But if this implementation to wait for the vote will be too complex, then let's just skip the vote. I think it's also fine. Without this feature, we wait for automatic failover, which will also not have the vote from the already shutdown primary.

i am going to skip the vote for now, i tried a bit which seemed not easy and not good looking to finish it. Maybe I'll have a better idea later, i will keep it in mind.

tests/unit/cluster/auto-failover-on-shutdown.tcl
```diff
@@ -0,0 +1,72 @@
+proc shutdown_on_how {srv_id how} {
```
zuiderkwast: Why "on"?

Maybe just shutdown_how?

tests/unit/cluster/auto-failover-on-shutdown.tcl
```diff
@@ -0,0 +1,72 @@
...
+        # Pause a replica so it has no chance to catch up with the offset.
+        pause_process $replica1_pid
+
+        # Primary write some data to increase the offset.
+        for {set i 0} {$i < 10} {incr i} {
+            $primary incr key_991803
+        }
+
+        # Wait the replica2 catch up with the offset
+        wait_for_ofs_sync $primary $replica2
```
zuiderkwast: I wish the test can pass even without this wait. If the primary has many writes and receives SIGTERM, we want failover without data loss.

The primary does pause write in prepare shutdown and it waits for replicas offset before finish shutdown, but if we failover immediately, the replica can miss some writes before the failover?

I think we need to postpone failover and do it when the offset is synced, some time between prepare shutdown and finish shutdown, or maybe in finish shutdown as in your first implementation, but don't wait for all replicas when this feature is enabled?

Every implementation has pros and cons?

zuiderkwast (commented):
> i am going to skip the vote for now, i tried a bit which seemed not easy and not good looking to finish it. Maybe I'll have a better idea later, i will keep it in mind.

I understand. Simple is better.

But possible data loss is not good. See comments below.

tests/unit/cluster/auto-failover-on-shutdown.tcl
enjoy-binbin: The test does not enable shutdown-timeout (it is 0 in the test), so the primary won't pause write and wait for the replica in this shutdown-timeout part.

Maybe we can, if the offset is equal to primary_repl_offset, we use a FORCE failover, if it is not, we use a plain failover? Noted that we will call flushReplicasOutputBuffers before doing this failover thing and exit thing, so i guess the replicas may still have a chance to catch up with the offset.

i am thinking it we enable shutdown-timeout, we have a good chance that can find a best replica, if we are not, i don't know if we should wait?

i will try to stable the test and also add a test with shutdown-timeout enabled.

PingXie:
This is an interesting idea. I like the direction we are going in but I agree with @zuiderkwast that potential data loss is not appealing. 

We can do both though IMO triggering a (graceful) failover as part of `CLUSTER FORGET` is more valuable than making it part of shutdown, because it is cleaner to forget a node prior to shutting it down in any production environment. 

Today, we can't forget "myself" nor "my primary" (with the latter being a dynamic state). This adds operational complexity. Imagine that the admin could just send `CLUSTER FORGET` to any node in the cluster and then the server will do the right thing, failing over the primaryship to one of its replicas, if applicable, and then broadcast the forget message to the cluster.

tests/unit/cluster/auto-failover-on-shutdown.tcl
```diff
@@ -26,12 +29,17 @@ proc test_main {how} {
...
-        wait_for_ofs_sync $primary $replica2
-        wait_replica_acked_ofs $primary $replica2_ip $replica2_port
+        if {$shutdown_timeout != 0} {
+            # Wait the replica2 catch up with the offset
+            wait_for_ofs_sync $primary $replica2
+            wait_replica_acked_ofs $primary $replica2 $replica2_ip $replica2_port
+        } else {
+            # If shutdown-timeout is enable, we expect the primary to pause writing
+            # and wait for the replica to catch up with the offset.
+        }
```
zuiderkwast: Did you mean `if {$shutdown_timeout == 0} {`?

tests/unit/cluster/auto-failover-on-shutdown.tcl
zuiderkwast: > The test does not enable shutdown-timeout (it is 0 in the test), so the primary won't pause write and wait for the replica in this shutdown-timeout part.

I think this feature makes more sense with shutdown-timeout. If you care about data loss, then you don't want to disable shutdown-timeout.

> Maybe we can, if the offset is equal to primary_repl_offset, we use a FORCE failover, if it is not, we use a plain failover?

Interesting idea, but more complex? We get more different cases.

> Noted that we will call flushReplicasOutputBuffers before doing this failover thing and exit thing, so i guess the replicas may still have a chance to catch up with the offset.

yes, but now in your code we send CLUSTER FAILOVER only if the replica is already in sync. Otherwise, we will not send it?

```C
            /* This is done only when the replica offset is caught up, to avoid data loss */
            if (replica->repl_state == REPLICA_STATE_ONLINE && replica->repl_ack_off == server.primary_repl_offset) {
                best_replica = replica;
                break;
            }
```

> i am thinking it we enable shutdown-timeout, we have a good chance that can find a best replica, if we are not, i don't know if we should wait?

I'm not sure what you mean. If we enable shutdown-timeout, we can wait, right? If shutdown-timeout is 0, we don't wait. But let's design it for shutdown-timeout > 0, because it makes more sense?

I don't see the problem with waiting. We have the `isReadyToShutdown` that we check in cron while we're waiting. When we have one replica up to date, we can send the CLUSTER FAILOVER FORCE to it. What's the problem with this idea? We need to set a flag like `server.shutdown_failover_sent = 1` to make sure we don't do it twice? Sure...

zuiderkwast:
> We can do both though IMO triggering a (graceful) failover as part of `CLUSTER FORGET` is more valuable than making it part of shutdown, because it is cleaner to forget a node prior to shutting it down in any production environment.

@PingXie Yes, it's a good idea, but this PR is about the scenario that the machine is taken down without the control of the Valkey admin. For example, in Kubernetes when a worker is shutdown, SIGTERM is sent to all processes and it waits for 30 seconds by default. When you shutdown your laptop, I believe it's similar, each application gets SIGTERM and has some time to be able to do a graceful shutdown.

tests/unit/cluster/auto-failover-on-shutdown.tcl
enjoy-binbin: Yes, i think we can make shutdown-failover only take effect when shutdown-timeout is enabled. 

I did try shutdown_failover_sent before, what should we do if the old primary received a UPDATE, should we turn myself into a replica of the new primary? or should we ignore it and just finish the shutdown. Since the waitting time is enough, currently the old primary will become a replica if it see an UPDATE, and then continue the finishShutdown thing. (i am not happy with this become replica thing... sure we can somehow check the shutdown_failover_sent thing in cluster.c to skip some replica thing, but it is odd)

zuiderkwast: > I did try shutdown_failover_sent before, what should we do if the old primary received a UPDATE, should we turn myself into a replica of the new primary?

If we receive an update, we can go immediately to finishShutdown? We don't need to wait for other replicas after that, right?

If we turn into a replica or not is not important as long as there is no data loss. :laughing:

src/cluster_legacy.c
```diff
@@ -1192,6 +1192,35 @@ void clusterInitLast(void) {
...
+            if (replica->repl_state == REPLICA_STATE_ONLINE && replica->repl_ack_off == server.primary_repl_offset) {
+                best_replica = replica;
+                break;
+            }
+        }
+
+        if (best_replica) {
+            /* Send a CLUSTER FAILOVER FORCE to the best replica. */
+            const char *buf = "*3\r\n$7\r\nCLUSTER\r\n$8\r\nFAILOVER\r\n$5\r\nFORCE\r\n";
+            if (connWrite(best_replica->conn, buf, strlen(buf)) == (int)strlen(buf)) {
```
enjoy-binbin: ok, i think here we have a problem, the primary is sending a CLUSTER FAILOVER to replica, replica will feed it to the replication stream.
```
    /* If the client is a primary we need to compute the difference
     * between the applied offset before and after processing the buffer,
     * to understand how much of the replication stream was actually
     * applied to the primary state: this quantity, and its corresponding
     * part of the replication stream, will be propagated to the
     * sub-replicas and to the replication backlog. */
    if (c->flag.primary) {
        long long applied = c->reploff - prev_offset;
        if (applied) {
            replicationFeedStreamFromPrimaryStream(c->querybuf + c->repl_applied, applied);
            c->repl_applied += applied;
        }
    }
```

I always though a ADMIN command won't do this thing, look like i get it wrong all the time. Do you guys think there is an another way to do this?

tests/unit/cluster/auto-failover-on-shutdown.tcl
enjoy-binbin: yes, thinking it again, it seem that becoming a replica is not a big deal, since it can always do a psync so it is not a big cost. I will handle it later.

src/cluster_legacy.c
zuiderkwast: Hm... A new connection from primary to replica?

zuiderkwast: What if it needs auth? A replica uses `primaryuser` and `primaryauth` to connect to the primary. At any time, failover can happen to change the roles of primary and replica. Therefore, I guess it's OK that the primary can use `primaryuser` and `primaryauth` to connect to the replica too...?

enjoy-binbin: Maybe we can implement a FAILOVER command in cluster mode? see https://github.com/valkey-io/valkey/commit/0d18a1e85fdb9967419dc9f7a5905452d065c84a
```
int clusterAllowFailoverCmd(client *c) {
    if (!server.cluster_enabled) {
        return 1;
    }
    addReplyError(c, "FAILOVER not allowed in cluster mode. "
                     "Use CLUSTER FAILOVER command instead.");
    return 0;
}
```

After the FAILOVER command is used, the primary node will psync with the replica, it is not a primary client, so it will not be put into the replication stream. Maybe we can implement it like this. However, replication.c and cluster.c are separate, which is a bit troublesome.

zuiderkwast: You mean the primary sends PSYNC FAILOVER to the replica? 🤔 We add some code for cluster, like `if (cluster) start failover vote`?

If we send this to an old version node, it can be very confused? If it becomes a primary but it does not do a cluster vote, we have a very confused cluster?

enjoy-binbin: Another approach is to add the replica_id parameter to cluster failover, so it become cluster failover replica_id, so that the corresponding replica can execute it after receiving it. Even if the replica writes it to the repl backlog, other replicas will not execute it. The disadvantage is that the old version replica cannot recognize it. Doesn't seem to be a good way either.

zuiderkwast: As long as the old version replica ignores it and doesn't crash, it's safe, right? But there is a risk that a replica crashes if the config `propagation-error-behavior` is set to `panic` or `panic-on-replicas`.

An alternative is to do it only if all replicas are known to understand this command. We can check `r->replica_version` (from `REPLCONF VERSION`) and only do this automatic failover if all replicas support this. It seems safe, wdyt?

enjoy-binbin: yes, it seems safe, i will give it a try.

enjoy-binbin: i gave it a try, it is bit odd i guess, see the top comment, let me know if you don't like it.

src/cluster_legacy.c
```diff
@@ -1233,25 +1233,30 @@ void clusterInitLast(void) {
...
         listNode *replicas_list_node;
         listRewind(server.replicas, &replicas_iter);
         while ((replicas_list_node = listNext(&replicas_iter)) != NULL) {
             client *replica = listNodeValue(replicas_list_node);
-            /* This is done only when the replica offset is caught up, to avoid data loss */
-            if (replica->repl_state == REPLICA_STATE_ONLINE && replica->repl_ack_off == server.primary_repl_offset) {
+            /* This is done only when the replica offset is caught up, to avoid data loss.
+             * And 0x800ff is 8.0.255, we only support new versions for this feature. */
+            if (replica->repl_data->repl_state == REPLICA_STATE_ONLINE &&
+                // replica->repl_data->replica_version > 0x800ff &&
```
zuiderkwast: Forgot to uncomment the check?

In the comment, maybe we shall not say "new" because in a few years, this will not be new anymore.

Why not check `>= 0x80100` instead `> 0x800ff`? It's the same but maybe easier to read?

src/cluster_legacy.c
```diff
@@ -1233,25 +1233,30 @@ void clusterInitLast(void) {
...
                 break;
             }
         }
 
         if (best_replica) {
             /* Send a CLUSTER FAILOVER FORCE to the best replica. */
-            const char *buf = "*3\r\n$7\r\nCLUSTER\r\n$8\r\nFAILOVER\r\n$5\r\nFORCE\r\n";
-            if (connWrite(best_replica->conn, buf, strlen(buf)) == (int)strlen(buf)) {
+            char buf[128];
+            size_t buflen = snprintf(buf, sizeof(buf), "*5\r\n$7\r\nCLUSTER\r\n$8\r\nFAILOVER\r\n$5\r\nFORCE\r\n$9\r\nreplicaid\r\n$%d\r\n%s\r\n", CLUSTER_NAMELEN, (char *)best_replica->name->ptr);
```
zuiderkwast: Very long line. Maybe break it?

Why is "replicaid" in lowercase? Other keywords are uppercase, like "FORCE".

```suggestion
            size_t buflen = snprintf(buf, sizeof(buf),
                                     "*5\r\n$7\r\nCLUSTER\r\n"
                                     "$8\r\nFAILOVER\r\n"
                                     "$5\r\nFORCE\r\n"
                                     "$9\r\nREPLICAID\r\n"
                                     "$%d\r\n%s\r\n",
                                     CLUSTER_NAMELEN,
                                     (char *)best_replica->name->ptr);
```

src/cluster_legacy.c
```diff
@@ -7026,32 +7032,46 @@ int clusterCommandSpecial(client *c) {
         } else {
             addReplyLongLong(c, clusterNodeFailureReportsCount(n));
         }
-    } else if (!strcasecmp(c->argv[1]->ptr, "failover") && (c->argc == 2 || c->argc == 3)) {
-        /* CLUSTER FAILOVER [FORCE|TAKEOVER] */
+    } else if (!strcasecmp(c->argv[1]->ptr, "failover") && (c->argc >= 2)) {
+        /* CLUSTER FAILOVER [FORCE|TAKEOVER] [replicaid <node id>] */
```
```suggestion
        /* CLUSTER FAILOVER [FORCE|TAKEOVER] [REPLICAID <node id>] */
```

src/cluster_legacy.c
```diff
@@ -7026,32 +7032,46 @@ int clusterCommandSpecial(client *c) {
...
             } else {
                 addReplyErrorObject(c, shared.syntaxerr);
                 return 1;
             }
         }
 
+        /* Check if it should be executed by myself. */
+        if (replicaid != NULL) {
+            clusterNode *n = clusterLookupNode(replicaid->ptr, sdslen(replicaid->ptr));
+            if (n != myself) {
```
zuiderkwast: We can just check `myself->name`?

```suggestion
            if (memcmp(replicaid, myself->name, CLUSTER_NAMELEN) != 0) {
```

src/cluster_legacy.c
```diff
@@ -7026,32 +7032,46 @@ int clusterCommandSpecial(client *c) {
...
+        }
+
         /* Check preconditions. */
         if (clusterNodeIsPrimary(myself)) {
-            addReplyError(c, "You should send CLUSTER FAILOVER to a replica");
+            if (replicaid == NULL) addReplyError(c, "You should send CLUSTER FAILOVER to a replica");
             return 1;
         } else if (myself->replicaof == NULL) {
-            addReplyError(c, "I'm a replica but my master is unknown to me");
+            if (replicaid == NULL) addReplyError(c, "I'm a replica but my master is unknown to me");
```
zuiderkwast: Why add this? If the primary is unknown, the failover can't succeed, so I think we need to return an error even if REPLICAID is sent.

src/cluster_legacy.c
```diff
@@ -7026,32 +7032,46 @@ int clusterCommandSpecial(client *c) {
...
             return 1;
         } else if (!force && (nodeFailed(myself->replicaof) || myself->replicaof->link == NULL)) {
-            addReplyError(c, "Master is down or failed, "
-                             "please use CLUSTER FAILOVER FORCE");
+            if (replicaid == NULL) addReplyError(c, "Master is down or failed, please use CLUSTER FAILOVER FORCE");
```
zuiderkwast: Why add this? I think it should be an error even with REPLICAID.

src/config.c
```diff
@@ -3187,6 +3187,7 @@ standardConfig static_configs[] = {
     createBoolConfig("enable-debug-assert", NULL, IMMUTABLE_CONFIG | HIDDEN_CONFIG, server.enable_debug_assert, 0, NULL, NULL),
     createBoolConfig("cluster-slot-stats-enabled", NULL, MODIFIABLE_CONFIG, server.cluster_slot_stats_enabled, 0, NULL, NULL),
     createBoolConfig("hide-user-data-from-log", NULL, MODIFIABLE_CONFIG, server.hide_user_data_from_log, 1, NULL, NULL),
+    createBoolConfig("import-mode", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.import_mode, 0, NULL, NULL),
```
zuiderkwast, line 3195: Wrong feature?

src/replication.c
```diff
@@ -3684,6 +3692,24 @@ void syncWithPrimary(connection *conn) {
...
+        } else {
+            server.repl_state = REPL_STATE_SEND_PSYNC;
+        }
+    }
+
+    /* Receive CLIENT SETNAME reply. */
+    if (server.repl_state == REPL_STATE_RECEIVE_SETNAME_REPLY) {
+        err = receiveSynchronousResponse(conn);
+        if (err == NULL) goto no_response_error;
+        /* Ignore the error if any. 8.1 introduced this logic and we don't care if it failed. */
```
zuiderkwast: But CLIENT SETNAME is older. Introduced in 7.0? 6.0?

Anyway, it's true that we don't care if it failed. It is best effort.

zuiderkwast (commented):
Very nice! I see you updated the top comment, but I think it should explain the new commands and behaviour in some more detail:

* New argument for `CLUSTER FAILOVER`
* Replica sends `CLIENT SETNAME` using its node-id as the name
* Primary sends `CLUSTER FAILOVER FORCE REPLICAID node-id` to all replicas, because it's a shared replication buffer, but only the replica with the matching id will execute it
* `CLUSTER FAILOVER REPLICAID node-id` can also be sent to a primary now? It replicates it and the replica will do the failover, right?
* New config? Maybe someone wants to disable it because they reboot faster than a failover and the node is still a primary when it comes back?

I don't understand why you wrote "To avoid polluting the replication stream" in the top comment. What are we avoiding here?

We can add test with an old replica in the cluster. I have merged #1371 so now it is easy to add a test like that. WDYT?

src/cluster_legacy.c
zuiderkwast: If we want this to be documented, we should update the JSON file too: `src/commands/cluster-failover.json`.

src/cluster_legacy.c
enjoy-binbin: i take this from
```
             * And 0x702ff is 7.2.255, we only support new versions in this case. */
            if (r->repl_data->repl_state == REPLICA_STATE_ONLINE && r->repl_data->replica_version > 0x702ff) {
                num_eligible_replicas++;
            }
```

> Forgot to uncomment the check?

just a easy way that i can test in local.

src/cluster_legacy.c
enjoy-binbin: It should be internal. I am not sure whether we should document it.

src/cluster_legacy.c
enjoy-binbin: I was wondering if there would be some races, that a replica return an error to the priamry. Or maybe we should always return OK if replicaid is passed.

src/config.c
enjoy-binbin: that is a merge commit, it is not in the latest diff

src/replication.c
enjoy-binbin: it is very old, i guess it was introduced in 2.x

src/cluster_legacy.c
zuiderkwast: Replicas don't send the replies to primaries. Only problem is the confic to panic on repöocation errors. But i can't see any races. Can you?

We can return ok if you want. Maybe we should check that `c` is the primary fake client and forbid this option for normal clients?

enjoy-binbin: since we will write it to the backlog, i am worry that after some down and up, the psync will get the command and return an error. Though I haven't verified it specifically.

enjoy-binbin:
> CLUSTER FAILOVER REPLICAID node-id can also be sent to a primary now? It replicates it and the replica will do the failover, right?

no, we won't replicate this command.

> New config? Maybe someone wants to disable it because they reboot faster than a failover and the node is still a primary when it comes back?

yes, i think a new config may allow people to make a better transition, if there are any problems

> I don't understand why you wrote "To avoid polluting the replication stream" in the top comment. What are we avoiding here?

i will remove it, it is a old stuff i guess, in the old version, the code has the issue

> We can add test with an old replica in the cluster. I have merged https://github.com/valkey-io/valkey/pull/1371 so now it is easy to add a test like that. WDYT?

Sound good, i will take a look.

valkey.conf
```diff
@@ -1692,6 +1692,10 @@ aof-timestamp-enabled no
 # shutdown-on-sigint default
 # shutdown-on-sigterm default
 
+# TODO
+#
+# auto-failover-on-shutdown no
```
zuiderkwast: TODO

zuiderkwast:
```suggestion
# If the node receives a SIGTERM or is shut down in another way when it is
# running in cluster mode and is a primary with replicas, it can trigger a
# manual failover to one of its replicas before shutting down. This is faster
# and safer than waiting for an automatic failover to happen. Default is 'no'.
#
# auto-failover-on-shutdown no
```

(Edited to mention cluster mode.)

enjoy-binbin: thanks a lot!

zuiderkwast (approved):
LGTM.

@valkey-io/core-team Let's see if we can include this in 8.1. The only blocker now is to find time to discuss it and approve it.

src/cluster_legacy.c
```diff
@@ -1252,6 +1252,57 @@ void clusterInitLast(void) {
 
 /* Called when a cluster node receives SHUTDOWN. */
 void clusterHandleServerShutdown(void) {
+    if (nodeIsPrimary(myself) && server.auto_failover_on_shutdown) {
```
PingXie: I would suggest moving this non-trivial block of code into its own function for readability.

src/cluster_legacy.c
```diff
@@ -1252,6 +1252,57 @@ void clusterInitLast(void) {
...
+        /* Find the first best replica, that is, the replica with the largest offset. */
+        int old_replica = 0;
```
PingXie: nit. legacy? "old" struck me as being an "out of sync" replica at first.

```suggestion
        int legacy_replica = 0;
```

src/cluster_legacy.c
```diff
@@ -1252,6 +1252,57 @@ void clusterInitLast(void) {
...
+                                     CLUSTER_NAMELEN,
+                                     (char *)best_replica->name->ptr);
+            /* Must install write handler for all replicas first before feeding
+             * replication stream. */
+            prepareReplicasToWrite();
+            feedReplicationBuffer(buf, buflen);
+            serverLog(LL_NOTICE, "Perform auto failover to replica %s on shutdown.", (char *)best_replica->name->ptr);
+        } else {
+            if (old_replica) {
+                serverLog(LL_NOTICE, "Unable to perform auto failover on shutdown since there are old replicas.");
```
```suggestion
                serverLog(LL_NOTICE, "Unable to perform auto failover on shutdown since there are legacy replicas.");
```

src/cluster_legacy.c
```diff
@@ -1252,6 +1252,57 @@ void clusterInitLast(void) {
...
+                                     "$9\r\nREPLICAID\r\n"
+                                     "$%d\r\n%s\r\n",
+                                     CLUSTER_NAMELEN,
+                                     (char *)best_replica->name->ptr);
+            /* Must install write handler for all replicas first before feeding
+             * replication stream. */
+            prepareReplicasToWrite();
+            feedReplicationBuffer(buf, buflen);
+            serverLog(LL_NOTICE, "Perform auto failover to replica %s on shutdown.", (char *)best_replica->name->ptr);
+        } else {
```
PingXie, line 1286: I would suggest testing the negative cases first and return immediately, generally speaking. this would help reduce code indentation.

src/cluster_legacy.c
```diff
@@ -1252,6 +1252,57 @@ void clusterInitLast(void) {
...
+            client *replica = listNodeValue(replicas_list_node);
+            /* This is done only when the replica offset is caught up, to avoid data loss.
+             * And 0x80100 is 8.1.0, we only support this feature in this version. */
+            if (replica->repl_data->replica_version < 0x80100) {
+                old_replica = 1;
+                best_replica = NULL;
+                break;
+            }
+            if (replica->repl_data->repl_state == REPLICA_STATE_ONLINE &&
+                replica->name && sdslen(replica->name->ptr) == CLUSTER_NAMELEN &&
```
PingXie: not sure I understand the need for this name check? also isn't `client setname` a best effort operation?

src/cluster_legacy.c
```diff
@@ -1252,6 +1252,57 @@ void clusterInitLast(void) {
...
+                replica->repl_data->repl_ack_off == server.primary_repl_offset) {
```
PingXie: I feel that the equal test is too strict. would it make sense to pick the replica with the highest replication offset? whether or not we yield the primaryship to it, it is going to become the new primary anyways. the only difference is "how long it takes" I think.

src/cluster_legacy.c
```diff
@@ -1252,6 +1252,57 @@ void clusterInitLast(void) {
...
+        if (best_replica) {
+            /* Send the CLUSTER FAILOVER FORCE REPLICAID node-id to all replicas since
+             * it is a shared replication buffer, but only the replica with the matching
+             * node-id will execute it. The caller will call flushReplicasOutputBuffers,
+             * so in here it is a best effort. */
+            char buf[128];
+            size_t buflen = snprintf(buf, sizeof(buf),
+                                     "*5\r\n$7\r\nCLUSTER\r\n"
+                                     "$8\r\nFAILOVER\r\n"
+                                     "$5\r\nFORCE\r\n"
```
PingXie: can we get by with `TAKEOVER`? the primary is going down so it shouldn't be involved in any slot migration activity I think?

src/cluster_legacy.c
enjoy-binbin: we will need the name to send the failover command, it rely a replicaid arg.

src/cluster_legacy.c
enjoy-binbin: good point, @zuiderkwast thoughts?

src/cluster_legacy.c
enjoy-binbin: I think we discussed it, but I can't find the comment, @zuiderkwast do you have one?

src/cluster_legacy.c
zuiderkwast: It's not too strict. The primary has already paused writes and waited for the replica. It can wait to shutdown for up to 10 seconds by default to make sure at least one of the replica has matched the offset.

If we do manual failover when the offset doesn't match, there is a risk that another replica has a better offset. The best one will usually become the primary in an automatic failover when they calculate the rank.

src/cluster_legacy.c
zuiderkwast: We discussed it long ago. :)

TAKEOVER bumps the epoch without consensus. It is very risky. If there is another failover in progress for another primary, I think we can get some epoch problems and the other failover can fail or something like that.

FORCE is not risky. What it does is just skip waiting for the offset. This is fine because we have already waited for the offset to match.

zuiderkwast: Also note that this shutdown may be unexpected. In a controlled shutdown, the admin should already do a manual failover before shutting down a primary in a controlled way.

The main purpose of this feature is to handle unexpected SIGTERM from the host system, for example the OS is shutting down or rebooting, or a kubernetes worker shutting down, for some reason outside of the Valkey admin's control.

So we don't know for sure that no slot migration is ongoing or some other manual failover is ongoing maybe. With this feature, I think we should just do what automatic failover would do but a bit faster. If all the other primaries are alive and can vote, it will work. We only need a majority.

I guess we could use TAKEOVER if there is no quorum, like if we have a cluster with less than 3 shards...? It could be nice, but we don't have high availability in 1-2 shard clusters anyway where automatic failover doesn't work.

src/cluster_legacy.c
```diff
@@ -1250,8 +1250,66 @@ void clusterInitLast(void) {
...
+    char buf[128];
+    size_t buflen = snprintf(buf, sizeof(buf),
+                             "*5\r\n$7\r\nCLUSTER\r\n"
+                             "$8\r\nFAILOVER\r\n"
+                             "$5\r\nFORCE\r\n"
+                             "$9\r\nREPLICAID\r\n"
+                             "$%d\r\n%s\r\n",
+                             CLUSTER_NAMELEN,
+                             (char *)best_replica->name->ptr);
+    /* Must install write handler for all replicas first before feeding
```
PingXie, line 1307: let's assert on `buflen <= 128`?

src/cluster_legacy.c
PingXie: Yeah I'd expect that in the vast majority cases we will have an exact match, which is great. In the rare or very rare case where we couldn't get a matching replica, are we saying we are ok with giving up on this fast failover path but letting the best replica to win after a full `cluster_node_timeout`? which I believe is the same as the primary picking the best replica?

src/cluster_legacy.c
PingXie: Got it. I am agree `FORCE` is a better option.

That said, `TAKEOVER` is risky only when a slot ownership change is also involved. I think it is *not incorrect* for two primaries to have the same node config epoch - so long as there is no conflict between their slot ownership. The collision will be resolved automatically. But yeah there is no need to open this can of worms.

valkey.conf
```diff
@@ -1692,6 +1692,13 @@ aof-timestamp-enabled no
 # shutdown-on-sigint default
 # shutdown-on-sigterm default
 
+# If the node receives a SIGTERM or is shut down in another way when it is
+# running in cluster mode and is a primary with replicas, it can trigger a
+# manual failover to one of its replicas before shutting down. This is faster
+# and safer than waiting for an automatic failover to happen. Default is 'no'.
+#
```
PingXie: nit
```suggestion
# In cluster mode, if a primary node with replicas receives a SIGTERM or is shut down,
# it can proactively initiate a manual failover. This promotes one of its replicas to
 # primary before shutdown, resulting in a quicker and safer transition than relying on
# an automatic failover. Default is 'no'.
#
```

PingXie (approved):
LGTM overall

src/replication.c
```diff
@@ -3626,6 +3626,14 @@ void syncWithPrimary(connection *conn) {
         err = sendCommand(conn, "REPLCONF", "version", VALKEY_VERSION, NULL);
         if (err) goto write_error;
 
+        /* Inform the primary of our (replica) node name. */
+        if (server.cluster_enabled) {
+            char *argv[] = {"CLIENT", "SETNAME", server.cluster->myself->name};
```
madolson: I don't understand why this isn't a replconf. It seems like this should be more specific information than just setting it as the name.

src/cluster_legacy.c
```diff
@@ -7083,22 +7141,37 @@ int clusterCommandSpecial(client *c) {
         } else {
             addReplyLongLong(c, clusterNodeFailureReportsCount(n));
         }
-    } else if (!strcasecmp(c->argv[1]->ptr, "failover") && (c->argc == 2 || c->argc == 3)) {
-        /* CLUSTER FAILOVER [FORCE|TAKEOVER] */
+    } else if (!strcasecmp(c->argv[1]->ptr, "failover") && (c->argc >= 2)) {
+        /* CLUSTER FAILOVER [FORCE|TAKEOVER] [REPLICAID <NODE ID>]
+         * REPLICAID is currently available only for internal so we won't
+         * put it into the JSON file. */
```
madolson, line 7169: I don't think we should just not document it. Someone will find it and use it, so I think removing it becomes a breaking change. All other arguments, including other internal commands, are documented in the json files.

src/cluster_legacy.c
zuiderkwast: Almost the same. The only difference is if the replica didn't ack the offset in time, or it did but the ack didn't reach the primary yet, so the primary doesn't know.

I'm not completely against this idea, but I don't think it's necessary either. It's a corner case. It's easier to explain the feature if the primary selects the best replica only when it is certain.

src/replication.c
zuiderkwast: Replconf needs some more work and a new field in the client struct, etc. Why is it bad to reuse the setname?

Psync and replconf are internal undocumented commands so I don't see any risk. Is it about code complexity?

src/cluster_legacy.c
PingXie: I am ok with the equal test. I was trying to validate my understanding. I am aligned.

madolson (commented):
The high level idea still seems really good to me. Sorry for not commenting sooner, since I think some of what I'm suggesting reverts earlier discussions, but the conversation wasn't that clear so I don't really understand why this design won out.

I don't really like the primary explicitly picks a replica and then sends it a failover request for all replicas into the replica buffer. It seems odd we'll commit more out of band data into the replication stream and if for any reason that replica doesn't take over none of the other ones will do it quickly.

Since this isn't backwards compatible anyways, can we not use a ping extension message to just broadcast to all the replicas that one of them should try to take over? I had an informal conversation with another person from kubernetes that it would also be nice to mark a node as failing itself and let whichever replica take over.

Alternatively, it also diverges from our failover command design, which relies on the primary sending a PSYNC command with a special flag indicating that it has stepped down which can get rejected.

src/replication.c
PingXie: I thought about reusing `REPLCONF` as well but since what we need a replica's identity, I feel that `CLIENT SETNAME` works too. Ideally though I would love to see us deduce the name automatically on the server side. One idea could be looking up the remote address against all the replica nodes after detecting a replica client.

madolson: Neither https://valkey.io/commands/replconf/ nor https://valkey.io/commands/psync/ are technically undocumented, I agree they aren't very well documented but they are part of the protocol and people build tooling around it. 

My other concern is now show up with a weird name that wasn't there before, so people might take dependencies on it. Do we want to forever behind the replica name (we could instead put something more human readable there)? It's a bit of a binding decision, and would rather have flexibility.

EDIT: I originally missed ping's comment, I think that ideally inferring the client would also be ideal, I'm just not sure that's always possible.

src/cluster_legacy.c
zuiderkwast: Now that i think of it, the cluster failover command is added last on the replication stream so the replica will see it only after replicating everything first. Thus it's very safe to do it. :-)

PingXie:
> It seems odd we'll commit more out of band data into the replication stream 

Yeah, that is indeed not 100% clean. We are reusing the replication stream for non-replication purposes. 

> if for any reason that replica doesn't take over none of the other ones will do it quickly.

I think this is fine and it fall back to the current behavior - a new primary will be elected after `cluster-node-timeout`.

> can we not use a ping extension message to just broadcast to all the replicas that one of them should try to take over?

And mark itself failed? I think it is a good idea and it helps the cluster converge faster on this node's status too.

> which relies on the primary sending a PSYNC command with a special flag indicating that it has stepped down which can get rejected.

Where do we do this today?

zuiderkwast:
> I don't really like the primary explicitly picks a replica and then sends it a failover request for all replicas into the replica buffer. It seems odd we'll commit more out of band data into the replication stream and if for any reason that replica doesn't take over none of the other ones will do it quickly.

Wdym none of the others will do it quickly? Only one can do it quickly and only the primary can select it. If all replicas try to get elected quickly, none of them gets a majority. That's why there's a random delay in automatic failover, which is what we want to avoid here.

If we use the cluster bus, the primary sends it to the chosen replica only? We could use a new message and a cluster bus capability flag instead then.

IIRC the idea of using the replication buffer was that the replica will get to it only when the replication is complete. Maybe we were overthinking it?

zuiderkwast:
> > It seems odd we'll commit more out of band data into the replication stream
> 
> Yeah, that is indeed not 100% clean. We are reusing the replication stream for non-replication purposes.

@PingXie Like we do with SETSLOT? It was a similar discussion. :)

PingXie:
>@PingXie Like we do with SETSLOT? It was a similar discussion. :) 

Not exactly. We do want all replicas to execute SETSLOT. The thing in common though is that there is no user data involved in both cases. I can give you that :)

madolson:
Btw, my opinion isn't that strong. I don't really mind another mechanism, it just seems weird that we are doing something yet another way. 

> Where do we do this today?

https://github.com/valkey-io/valkey/blob/f85c93301ca63f070164b307c6a7371a9ab5ac7b/src/replication.c#L1047

> Wdym none of the others will do it quickly? Only one can do it quickly and only the primary can select it. If all replicas try to get elected quickly, none of them gets a majority. That's why there's a random delay in automatic failover, which is what we want to avoid here.

One of them will have the lowest delay, and yes it will have some random component, but the benefit is that there will be less complexity and only a single failure path. There is a belief at AWS that it's more important to have fewer failover pathways, and they should all reinforce each other since there is a lower chance of bugs there.

> Not exactly. We do want all replicas to execute SETSLOT. The thing in common though is that there is no user data involved in both cases. I can give you that :)

Maybe it's just me, but I think it kind of is data related. It's whether or not the given node is able to serve data related to a given slot. I guess you could say this is giving ownership over all of the slots to a given replica. I think our handling of slot state is a bit of a mess, although that's not really related to this PR.

> IIRC the idea of using the replication buffer was that the replica will get to it only when the replication is complete. Maybe we were overthinking it?

Maybe I'm overthinking it :D

src/replication.c
zuiderkwast: REPLCONF has no documented arguments. The page just says "Usage: REPLCONF".

zuiderkwast: If we document it, I think it will mostly be confusing to users.

Users might think they can send it to the primary  like with standalone FAILOVER TO, but it's not replicated so it doesn't work.

I agree though that even if we don't document it, we probably need to keep it around.

zuiderkwast: > My other concern is now show up with a weird name that wasn't there before, so people might take dependencies on it.

Good point. The replicas show up in CLIENT LIST.

src/cluster_legacy.c
zuiderkwast: If we document it, I think it will mostly be confusing to users.

Users might think they can send it to the primary like with standalone FAILOVER TO, but it's not replicated so it doesn't work.

I agree though that even if we don't document it, we probably need to keep it around.

> All other arguments, including other internal commands, are documented in the json files.

REPLCONF has no arguments in the JSON file. The page just says "Usage: REPLCONF". So this will not be the only undocumented argument.

zuiderkwast:
> One of them will have the lowest delay, and yes it will have some random component, but the benefit is that there will be less complexity and only a single failure path.

@madolson So with this solution, we just skip the time it takes for the primary to be marked as FAIL. That's about half of the time it takes to get a new primary. If we do that, I think this feature misses the point.

Internally, we have a small script that starts valkey-server inside the same container, so valkey-server is a child process of the script. This script catches the SIGTERM from the OS (or kubernetes) and performs the failover by sending commands. It is very fast.

The purpose is to catch a worker reboot (machine, infra level) that is out of control of the valkey admin (or kubernetes operator). By any means necessary we want to avoid this 3 seconds downtime. (I think we use a node timeout of 2 or something.)

> There is a belief at AWS that it's more important to have fewer failover pathways, and they should all reinforce each other since there is a lower chance of bugs there.

It's simplicity to the price of worse availability. We can (and we already do) handle this outside of valkey-server but that's more complex and non-standard solution that an official valkey-operator probably wouldn't use. It's more clean to just start valkey-server as it is.

src/replication.c
enjoy-binbin: > The replicas show up in CLIENT LIST.

I use SETNAME on purpose. I would like to show it in the client list so that we can quickly locate the node and its associated node id in the replication part (not the cluster part). But it’s not that important, sorry i did not mention this impact at the first time.

enjoy-binbin: changed to use replconf

src/server.h
```diff
@@ -1154,6 +1155,7 @@ typedef struct ClientReplicationData {
                                            see the definition of replBufBlock. */
     size_t ref_block_pos;                /* Access position of referenced buffer block,
                                            i.e. the next offset to send. */
+    char nodeid[CLUSTER_NAMELEN];        /* Node id in cluster mode. */
```
zuiderkwast: This is 40 bytes, unused in non-cluster mode. Should we use an allocated string instead?

If we do this, then we don't need the CLUSTER_NAMELEN definition from cluster.h.

valkey.conf
```diff
@@ -1692,6 +1692,13 @@ aof-timestamp-enabled no
 # shutdown-on-sigint default
 # shutdown-on-sigterm default
 
+# In cluster mode, if a primary node with replicas receives a SIGTERM or is shut down,
+# it can proactively initiate a manual failover. This promotes one of its replicas to
+ # primary before shutdown, resulting in a quicker and safer transition than relying on
```
zuiderkwast: There is a space here

```suggestion
# primary before shutdown, resulting in a quicker and safer transition than relying on
```

src/server.h
```diff
@@ -151,6 +151,7 @@ struct hdr_histogram;
 #else
 #define CONFIG_ACTIVE_DEFRAG_DEFAULT 1
 #endif
+#define CLUSTER_NAMELEN 40
```
zuiderkwast: If we need this, it's a sign of some bad design... I hope we can avoid it. This file doesn't include cluster.h and I think there is a reason to keep things separated.

zuiderkwast (commented):
I like the update with REPLCONF. Just some small comments.

src/replication.c
```diff
@@ -1494,6 +1502,21 @@ void replconfCommand(client *c) {
...
             }
             c->repl_data->associated_rdb_client_id = (uint64_t)client_id;
+        } else if (!strcasecmp(c->argv[j]->ptr, "set-cluster-node-id")) {
+            /* REPLCONF SET-CLUSTER-NODE-ID <node-id> */
+            if (!server.cluster_enabled) {
+                addReplyError(c, "This instance has cluster support disabled");
+                return;
+            }
+
+            clusterNode *n = clusterLookupNode(c->argv[j + 1]->ptr, sdslen(c->argv[j + 1]->ptr));
```
enjoy-binbin, line 1518: I'm not sure if we need to do more checks here, this will use clusterNode in replication.c. We can also just sdsdup without any checks

tests/assets/minimal-cluster.conf
```diff
@@ -1,4 +1,5 @@
 # Minimal configuration for testing.
+databases 1
```
enjoy-binbin: This can be removed, in the future we will support multi dbs.
```suggestion
```

I added this because it will print the warning in local.
```
    /* in case cluster mode is enabled dbnum must be 1 */
    if (server.cluster_enabled && server.dbnum > 1) {
        serverLog(LL_WARNING, "WARNING: Changing databases number from %d to 1 since we are in cluster mode",
                  server.dbnum);
        server.dbnum = 1;
    }
```

src/replication.c
zuiderkwast: sdsdup without check seems OK to me. We do the check when we lookup the node on shutdown...

Another possibility it to store the clusterNode pointer in the client struct instead of storing the node-id. Is it safe?

src/cluster_legacy.c
zuiderkwast: For the record, this was mentioned an online meeting: TAKEOVER is risky in a netsplit situation. It can allow failover in a minority partition, while another node wins an election in the majority partition. We end up with two primaries in the same shard.

src/replication.c
enjoy-binbin: on the other hand, like in set-rdb-client-id, we did check if the client_id is valid
```
        } else if (!strcasecmp(c->argv[j]->ptr, "set-rdb-client-id")) {
            /* REPLCONF identify <client-id> is used to identify the current replica main channel with existing
             * rdb-connection with the given id. */
            long long client_id = 0;
            if (getLongLongFromObjectOrReply(c, c->argv[j + 1], &client_id, NULL) != C_OK) {
                return;
            }
            if (!lookupRdbClientByID(client_id)) {
                addReplyErrorFormat(c, "Unrecognized RDB client id %lld", client_id);
                return;
            }
            c->repl_data->associated_rdb_client_id = (uint64_t)client_id;
```

It bother me... should we check? or should we check more like check if the nodeid is a replica of myself.

> Another possibility it to store the clusterNode pointer in the client struct instead of storing the node-id. Is it safe?

This seems safe, but we get stuck in the include header issue again.

zuiderkwast: > on the other hand, like in set-rdb-client-id, we did check if the client_id is valid

OK, so let's keep sdsdup with simple validation then? No need to change it.

>> Another possibility it to store the clusterNode pointer in the client struct instead of storing the node-id. Is it safe?
>
> This seems safe, but we get stuck in the include header issue again.

`replication.c` includes `cluster.h`.

The mess is just `server.h` that has prototypes for many C files. Recently @rjd15372 moved module protypes from `server.h` to `module.h`. I think that was a good idea. We could move replication related stuff from `server.h` to a new file `replication.h`. It would still need `cluster.h` though. It's not trivial to reduce the coupling between units.

zuiderkwast:
Bump @valkey-io/core-team Can we have this in 8.1?

src/cluster_legacy.c
```diff
@@ -1250,8 +1250,67 @@ void clusterInitLast(void) {
...
+    listRewind(server.replicas, &replicas_iter);
+    while ((replicas_list_node = listNext(&replicas_iter)) != NULL) {
+        client *replica = listNodeValue(replicas_list_node);
+        /* This is done only when the replica offset is caught up, to avoid data loss.
+         * And 0x80100 is 8.1.0, we only support this feature in this version. */
+        if (replica->repl_data->replica_version < 0x80100) {
+            legacy_replica = 1;
+            best_replica = NULL;
+            break;
+        }
```
hpatro: Shouldn't we loop over all the replicas and make the decision ?

```suggestion
        if (replica->repl_data->replica_version < 0x80100) {
            legacy_replica = 1;
            continue;
        }
```

best_replica is already NULL. So, we can remove that statement as well.

src/cluster_legacy.c
```diff
@@ -1250,8 +1250,67 @@ void clusterInitLast(void) {
...
+     * so in here it is a best effort. */
+    char buf[128];
+    size_t buflen = snprintf(buf, sizeof(buf),
+                             "*5\r\n$7\r\nCLUSTER\r\n"
+                             "$8\r\nFAILOVER\r\n"
+                             "$5\r\nFORCE\r\n"
+                             "$9\r\nREPLICAID\r\n"
+                             "$%d\r\n%s\r\n",
+                             CLUSTER_NAMELEN,
+                             best_replica->repl_data->replica_nodeid);
```
hpatro: Safer to limit the node id

```suggestion
    size_t buflen = snprintf(buf, sizeof(buf),
                             "*5\r\n$7\r\nCLUSTER\r\n"
                             "$8\r\nFAILOVER\r\n"
                             "$5\r\nFORCE\r\n"
                             "$9\r\nREPLICAID\r\n"
                             "$%d\r\n%.40s\r\n",
                             CLUSTER_NAMELEN,
                             best_replica->repl_data->replica_nodeid);
```

src/replication.c
```diff
@@ -3626,6 +3649,14 @@ void syncWithPrimary(connection *conn) {
         err = sendCommand(conn, "REPLCONF", "version", VALKEY_VERSION, NULL);
         if (err) goto write_error;
 
+        /* Inform the primary of our (replica) node name. */
+        if (server.cluster_enabled) {
+            char *argv[] = {"REPLCONF", "SET-CLUSTER-NODE-ID", server.cluster->myself->name};
+            size_t lens[] = {8, 19, CLUSTER_NAMELEN};
```
hpatro:
```suggestion
            size_t lens[] = {strlen(argv[0]), strlen(argv[1]), CLUSTER_NAMELEN};
```

src/cluster_legacy.c
zuiderkwast: The reason we break in this case is that we have a shared replication buffer and we want to send something new over the replication stream that legacy replicas would not understand. They might panic, depending on a config. That's why we only enable the feature if *all* replicas support it.

zuiderkwast: .. and no, `best_replica` is not always NULL here. We may have already set it to a replica. We don't break when we find a good replica because we also want to check that all replicas support the feature.

hpatro: Gotcha! My bad!

src/cluster_legacy.c
```diff
@@ -1254,8 +1254,67 @@ void clusterInitLast(void) {
...
+     * node-id will execute it. The caller will call flushReplicasOutputBuffers,
+     * so in here it is a best effort. */
+    char buf[128];
+    size_t buflen = snprintf(buf, sizeof(buf),
+                             "*5\r\n$7\r\nCLUSTER\r\n"
+                             "$8\r\nFAILOVER\r\n"
+                             "$5\r\nFORCE\r\n"
+                             "$9\r\nREPLICAID\r\n"
+                             "$%d\r\n%.40s\r\n",
+                             CLUSTER_NAMELEN,
```
zuiderkwast: Both of these length are the same, but one is `CLUSTER_NAMELEN` and the other one is hard-coded to 40? It seems like a fake abstraction. We can use `CLUSTER_NAMELEN` for both lengths ...

```suggestion
                             "$%d\r\n%.*s\r\n",
                             CLUSTER_NAMELEN,
                             CLUSTER_NAMELEN,
```

... or just hard-code 40 in both places ...

```suggestion
                             "$40\r\n%.40s\r\n",
```

commands.xlsx
PingXie, line 1: why is an excel file added?

src/cluster_legacy.c
```diff
@@ -1254,8 +1254,67 @@ void clusterInitLast(void) {
...
+    int legacy_replica = 0;
+    client *best_replica = NULL;
+    listIter replicas_iter;
+    listNode *replicas_list_node;
+    listRewind(server.replicas, &replicas_iter);
+    while ((replicas_list_node = listNext(&replicas_iter)) != NULL) {
+        client *replica = listNodeValue(replicas_list_node);
+        /* This is done only when the replica offset is caught up, to avoid data loss.
+         * And 0x80100 is 8.1.0, we only support this feature in this version. */
+        if (replica->repl_data->replica_version < 0x80100) {
```
```suggestion
        if (replica->repl_data->replica_version < 0x90000) {
```

src/cluster_legacy.c
```diff
@@ -1254,8 +1254,67 @@ void clusterInitLast(void) {
...
+    /* Find the first best replica, that is, the replica with the largest offset. */
+    int legacy_replica = 0;
+    client *best_replica = NULL;
+    listIter replicas_iter;
+    listNode *replicas_list_node;
+    listRewind(server.replicas, &replicas_iter);
+    while ((replicas_list_node = listNext(&replicas_iter)) != NULL) {
+        client *replica = listNodeValue(replicas_list_node);
+        /* This is done only when the replica offset is caught up, to avoid data loss.
+         * And 0x80100 is 8.1.0, we only support this feature in this version. */
```
```suggestion
         * And 0x90000 is 9.0.0, we only support this feature in this version. */
```

src/commands_table.txt
PingXie, line 1: this is an empty file?

utils/create-cluster/create-cluster
```diff
@@ -28,7 +28,7 @@ then
     while [ $((PORT < ENDPORT)) != "0" ]; do
         PORT=$((PORT+1))
         echo "Starting $PORT"
-        $BIN_PATH/valkey-server --port $PORT --protected-mode $PROTECTED_MODE --cluster-enabled yes --cluster-config-file nodes-${PORT}.conf --cluster-node-timeout $TIMEOUT --appendonly yes --appendfilename appendonly-${PORT}.aof --appenddirname appendonlydir-${PORT} --dbfilename dump-${PORT}.rdb --logfile ${PORT}.log --daemonize yes --enable-protected-configs yes --enable-debug-command yes --enable-module-command yes ${ADDITIONAL_OPTIONS}
+        $BIN_PATH/valkey-server --port $PORT --protected-mode $PROTECTED_MODE --databases 1 --cluster-enabled yes --cluster-config-file nodes-${PORT}.conf --cluster-node-timeout $TIMEOUT --appendonly yes --appendfilename appendonly-${PORT}.aof --appenddirname appendonlydir-${PORT} --dbfilename dump-${PORT}.rdb --logfile ${PORT}.log --daemonize yes --enable-protected-configs yes --enable-debug-command yes --enable-module-command yes --auto-failover-on-shutdown yes ${ADDITIONAL_OPTIONS}
```
PingXie: what is the reason for this explicit `database` config?

utils/generate-command-history-table.py
PingXie, line 1: is this meant to replace other codegen tools? like generate-command-code.py? can this be done in a separate pr?

utils/generate-command-history-table.py
```diff
@@ -0,0 +1,699 @@
...
+df = pd.DataFrame(all_commands.values(), columns=all_titles)
+df["since"] = df["since"].str.strip('"')
+df["complexity"] = df["complexity"].str.strip('"')
+df["group_order"] = df["group"].map({v:i for i,v in enumerate(all_groups)})
+sorted_df = df.sort_values(
+    by=["group_order", "since"],
+    key=lambda x: x if x.name != "since" else x.str.split('.').apply(lambda v: tuple(map(int, v))),
+    ascending=[True, True]
+).drop("group_order", axis=1)
+sorted_df.to_excel("commands.xlsx", index=False, engine="openpyxl")
```
PingXie: can we use a more text format?

tests/unit/cluster/auto-failover-on-shutdown.tcl
```diff
@@ -0,0 +1,92 @@
...
+    } elseif {$how == "sigterm"} {
+        exec kill -SIGTERM [s -$srv_id process_id]
+    }
+}
+
+# We will start a cluster with 3 primary nodes and 4 replicas, the primary 1 will have 2 replicas.
+# We will pause the replica 1, and then shutdown the primary 1, and making replica 2 to become
+# the new primary.
+proc test_auto_failover {how shutdown_timeout} {
+    test "auto-failover-on-shutdown will always pick a best replica and send CLUSTER FAILOVER - $how - shutdown-timeout: $shutdown_timeout" {
```
PingXie: nit
```suggestion
    test "auto-failover-on-shutdown hands over primaryship to a fully sync'd replica - $how - shutdown-timeout: $shutdown_timeout" {
```

PingXie (commented):
The main flow and tests LGTM. I just have a few questions on the new script.

I think we should consider a further enhancement after this pr to keep the old primary running a bit longer after the failover and redirect requests to the new primary with `-MOVED`.

commands.xlsx
enjoy-binbin: ooh sorry, this is tmp file (local files), i must add it wrongly after the last merge.

utils/generate-command-history-table.py
enjoy-binbin: this is a tmp local file, i removed it now.

utils/generate-command-history-table.py
enjoy-binbin: this is a tmp local file, i removed it now.

src/commands_table.txt
enjoy-binbin: this is a tmp local file, i removed it now.

utils/create-cluster/create-cluster
enjoy-binbin: The reason is that our previous configuration item will output a warning by default, that is, a warning about changing dbnum to 1.

This is a local change i used for test, i removed it now.

src/cluster_legacy.c
enjoy-binbin: apply your suggestion.

madolson:
Discussed briefly in core team meeting. No more open questions, so once low level details are done we can merge it for 9.0.

zuiderkwast:
@hpatro @madolson Do want to complete your reviews or can we merge it?

PingXie:
>  can we merge it?

I have a few new open questions. Will also reset my approval next

valkey.conf
```diff
@@ -1692,6 +1692,13 @@ aof-timestamp-enabled no
 # shutdown-on-sigint default
 # shutdown-on-sigterm default
 
+# In cluster mode, if a primary node with replicas receives a SIGTERM or is shut down,
+# it can proactively initiate a manual failover. This promotes one of its replicas to
+# primary before shutdown, resulting in a quicker and safer transition than relying on
+# an automatic failover. Default is 'no'.
```
PingXie:
```suggestion
# an automatic failover. For a replica to be eligible for this promotion, it must be fully 
# synchronized with the primary node at the time of the shutdown signal. If no such 
# replica is found, this proactive failover will not occur. Default is 'no'.
```

PingXie (approved):
LGTM.

madolson (approved):
I never reviewed all the details, but the high-level is good to me now.

zuiderkwast:
When reviewing #2195, I'm thinking back on this feature again. Here, we added a new config `auto-failover-on-shutdown`. Why didn't we instead add a new option `failover` to the existing configs for `shutdown-on-sigterm` and `shutdown-on-sigint` and an argument to the SHUTDOWN command (`SHUTDOWN [FAILOVER]`) to trigger the failover on shutdown based on how the shutdown is triggered? This is how other shutdown behavior is selected.

This is not released yet so we can still change this.

enjoy-binbin:
right, i totally forgot it. I guess it might be the difference between active and passive? auto-failover-on-shutdown (and shutdown-on-sig failover for sure) is like a passive way, and SHUTDOWN FAILOVER is a active way.

And you are right, this anyway is not released and we can change it for good.

PR #1097, opened 2024-09-30
avoid double close on replica main channel

### LINKED ISSUE #1088: [BUG] `server.repl__transfer_s` is double-closed/freed in `fullSyncWithPrimary`
This looks like a bug to me. It seems the two connection objects are getting mixed up. We should zero out the RDB channel connection here, i.e., `server.repl_rdb_transfer_s = NULL`. Then on L2726/L2747, we should be testing the normal channel connection instead, i.e., `if (server.repl_transfer_s)`

_Originally posted by @PingXie in https://github.com/valkey-io/valkey/pull/945#discussion_r1774245923_

ranshid:
fixes #1088

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -2738,8 +2738,10 @@ static void fullSyncWithPrimary(connection *conn) {
 
 error:
     sdsfree(err);
-    connClose(conn);
-    server.repl_transfer_s = NULL;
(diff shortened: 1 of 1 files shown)


PingXie (approved):
Thanks @ranshid!

naglera (approved):

PR #288, opened 2024-04-10
Removing Redis from internal lua function names and comments

parthpatel:
* Fixed a bug where 'server *' invocation from lua debugger was not working. Manually tested it as follows - 
```
Before 

lua debugger> redis ping
<redis> ping
<reply> "+PONG"
lua debugger> server ping
<error> Unknown Redis Lua debugger command or wrong number of arguments.

After

% ./src/valkey-cli --ldb --eval /tmp/1
* Stopped at 1, stop reason = step over
-> 1   return redis.call('ping');
lua debugger> server ping
<redis> ping
<reply> "+PONG"
```
* Functions in lua c files lack of documentation. Adding some documentation in this commit.
* I removed Redis from function names in eval.c and script_lua.c files. I did not touch log messages in this commit. I also did not touch references to "RedisProtocol".

PR #1102, opened 2024-10-01
Removing Redis from internal lua function names and comments

parthpatel:
I couldn't figure out how to rebase on top of recent changes, so creating this new PR. This has changes to handle comments on previous PR #288.

### DIFF
--- a/src/eval.c
+++ b/src/eval.c
@@ -1369,7 +1369,7 @@ char *ldbRespToHuman_Double(sds *o, char *reply) {
 /* Log a RESP reply as debugger output, in a human readable format.
  * If the resulting string is longer than 'len' plus a few more chars
  * used as prefix, it gets truncated. */
-void ldbLogServerReply(char *reply) {
+void ldbLogRespReply(char *reply) {
     sds log = sdsnew("<reply> ");
     ldbRespToHuman(&log, reply);
     ldbLogWithMaxLen(log);
@@ -1522,7 +1522,7 @@ void ldbServer(lua_State *lua, sds *argv, int argc) {
          * given by the user (without the first argument) and we also push the 'server' global table and
          * 'server.call' function so:
          * (1 (server table)) + (1 (server.call function)) + (argc - 1 (all arguments without the first)) = argc + 1*/
-        ldbLogServerReply("max lua stack reached");
+        ldbLogRespReply("max lua stack reached");
         return;
     }
 
--- a/src/eval.c
+++ b/src/eval.c
@@ -28,12 +28,12 @@
  */
 
 /*
- * This file initializes the global LUA object and registers functions to call Valkey API from within the LUA language. 
+ * This file initializes the global LUA object and registers functions to call Valkey API from within the LUA language.
  * It heavily invokes LUA's C API documented at https://www.lua.org/pil/24.html. There are 2 entrypoint functions in
  * this file:
- * 1. evalCommand() - Gets invoked every time a user runs LUA script via eval command on Valkey. 
- * 2. scriptingInit() - initServer() function from server.c invokes this to initialize LUA at startup. 
- *                      It is also invoked between 2 eval invocations to reset Lua. 
+ * 1. evalCommand() - Gets invoked every time a user runs LUA script via eval command on Valkey.
+ * 2. scriptingInit() - initServer() function from server.c invokes this to initialize LUA at startup.
+ *                      It is also invoked between 2 eval invocations to reset Lua.
  */
 #include "server.h"
 #include "sha1.h"
@@ -136,7 +136,7 @@ void sha1hex(char *digest, char *script, size_t len) {
     digest[40] = '\0';
 }
 
-/* Adds server.breakpoint() function used by lua debugger.  
+/* Adds server.breakpoint() function used by lua debugger.
  *
  * Allows to stop execution during a debugging session from within
  * the Lua code implementation, like if a breakpoint was set in the code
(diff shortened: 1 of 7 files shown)


parthpatel:
Fixed all problems with the build. Next build will succeed. This is ready for review.

src/script_lua.c
```diff
@@ -1432,52 +1434,69 @@ void luaRegisterLogFunction(lua_State *lua) {
     lua_settable(lua, -3);
 }
 
+/*
+ * Adds server.* functions/fields to lua such as server.call etc.
+ * This function only handles fields common between Functions and LUA scripting.
+ * scriptingInit() and functionsInit() may add additional fields specific to each.
+ */
 void luaRegisterServerAPI(lua_State *lua) {
+    /* In addition to registering server.call/pcall API, we will throw a customer message when a script accesses
```
hpatro: `customer message` sounds weird.

hpatro (commented):
IIRC, there were few tests in your previous PR to verify the availability of both the APIs.

src/script_lua.c
parthpatel: good catch, it is custom message - fixed.

parthpatel:
> IIRC, there were few tests in your previous PR to verify the availability of both the APIs.

@hpatro yeah. scripting.tcl has this block which will replace redis with server and test the same scripts. 

```
foreach script_compatibility_api {server redis} {

# We run the tests using both the server APIs, e.g. server.call(), and valkey APIs, e.g. redis.call(),
# in order to ensure compatibility.
if {$script_compatibility_api eq "server"} {
    proc replace_script_redis_api_with_server {args} {
        set new_string [regsub -all {redis\.} [lindex $args 0] {server.}]
        return lreplace $args 0 0 $new_string
    }

    proc get_script_api_name {} {
        return "server"
    }
} else {
    proc replace_script_redis_api_with_server {args} {
        return {*}$args
    }

    proc get_script_api_name {} {
        return "redis"
    }
}

....
```

src/eval.c
```diff
@@ -1365,7 +1369,7 @@ char *ldbRespToHuman_Double(sds *o, char *reply) {
 /* Log a RESP reply as debugger output, in a human readable format.
  * If the resulting string is longer than 'len' plus a few more chars
  * used as prefix, it gets truncated. */
-void ldbLogRespReply(char *reply) {
```
hwware: Why we need update Reps to Server?

zuiderkwast: We didn't change RESP before. It is the name of the protocol.

We don't know what it means, It's just a name. Maybe now we can say it means Response Protocol. 😀

parthpatel: There was a conflict at this line when I rebased. I didn't know which way to resolve, so resolved it to Server :) 
I have no preference either way. 

Do we want to refer to RESP in code base long-term or just remove references to it? eval.c still has many functions with Resp in it. 

```
char *ldbRespToHuman_Int(sds *o, char *reply);
char *ldbRespToHuman_Bulk(sds *o, char *reply);
char *ldbRespToHuman_Status(sds *o, char *reply);
char *ldbRespToHuman_MultiBulk(sds *o, char *reply);
char *ldbRespToHuman_Set(sds *o, char *reply);
char *ldbRespToHuman_Map(sds *o, char *reply);
char *ldbRespToHuman_Null(sds *o, char *reply);
char *ldbRespToHuman_Bool(sds *o, char *reply);
char *ldbRespToHuman_Double(sds *o, char *reply);
char *ldbRespToHuman(sds *o, char *reply) {
    case ':': p = ldbRespToHuman_Int(o, reply); break;
    case '$': p = ldbRespToHuman_Bulk(o, reply); break;
    case '+': p = ldbRespToHuman_Status(o, reply); break;
    case '-': p = ldbRespToHuman_Status(o, reply); break;
    case '*': p = ldbRespToHuman_MultiBulk(o, reply); break;
    case '~': p = ldbRespToHuman_Set(o, reply); break;
    case '%': p = ldbRespToHuman_Map(o, reply); break;
    case '_': p = ldbRespToHuman_Null(o, reply); break;
    case '#': p = ldbRespToHuman_Bool(o, reply); break;
    case ',': p = ldbRespToHuman_Double(o, reply); break;
/* The following functions are helpers for ldbRespToHuman(), each
char *ldbRespToHuman_Int(sds *o, char *reply) {
char *ldbRespToHuman_Bulk(sds *o, char *reply) {
char *ldbRespToHuman_Status(sds *o, char *reply) {
char *ldbRespToHuman_MultiBulk(sds *o, char *reply) {
        p = ldbRespToHuman(o, p);
char *ldbRespToHuman_Set(sds *o, char *reply) {
        p = ldbRespToHuman(o, p);
char *ldbRespToHuman_Map(sds *o, char *reply) {
        p = ldbRespToHuman(o, p);
        p = ldbRespToHuman(o, p);
char *ldbRespToHuman_Null(sds *o, char *reply) {
char *ldbRespToHuman_Bool(sds *o, char *reply) {
char *ldbRespToHuman_Double(sds *o, char *reply) {
    ldbRespToHuman(&log, reply);
```

parthpatel: @zuiderkwast nevermind, I did not read your comment properly. I have my answer :)

hpatro (approved):
Avoid force-push on Github (if possible), one needs to go through the entire diff again😄.

parthpatel:
> Avoid force-push on Github (if possible), one needs to go through the entire diff again😄.

noted :) I keep forgetting to add signed-off-by... 
Although I did not squash the commit, last revision is in the last commit.

src/script_lua.c
```diff
@@ -56,10 +56,10 @@ static char *libraries_allow_list[] = {
 };
 
 /* Lua API globals */
-static char *redis_api_allow_list[] = {
+static char *server_api_allow_list[] = {
     SERVER_API_NAME,
     REDIS_API_NAME,
-    "__redis__err__handler", /* error handler for eval, currently located on globals.
+    "__server__err__handler", /* error handler for eval, currently located on globals.
```
zuiderkwast: The existence of this global is undocumented, but I guess this change can break user scripts if they access this global, like calling it manually `EVAL "__redis__err__handler('foo')" 0` or something.

@madolson You have seen many strange things from users. Do you think this would be a risk?

zuiderkwast (approved):
LGTM

src/script_lua.c
madolson: Yeah. I guess I would expect this list to include both the err handlers. I can't think of a reason a user would touch this, but a user could theoretically be manipulating this since it's in the accessible globals list.

madolson, line 63:
```suggestion
    "__redis__err__handler",   /* Backwards compatible error handler */
    "__server__err__handler", /* error handler for eval, currently located on globals.
```

parthpatel: Alright, added the suggested change since it is simple change. 
I generally don't prefer to keep backwards compatibility for internal variables. That just makes it too complicated.

madolson (approved):

src/script_lua.c
zuiderkwast: It's not an internal variable though. It's accessible by users, though undocumented.

Both is safe, but we could also just have kept only the redis error handler here instead of adding one more. :-)

PR #1106, opened 2024-10-02
Avoid .c, .d and .o files from being copied to the binary tar.gz releases

melroy89:
As discussed here: https://github.com/orgs/valkey-io/discussions/1103#discussioncomment-10814006

`cp` can't be used anymore, `rsync` is more powerful and allow to exclude files.

Alternatively:

1. Remove the c, d and o files. Which isn't ideal either.
2. Improve the build. Eg. by building inside a `build` directory instead of in the src folder.

Ps. I know these workflows aren't triggered in this PR. Only via "Build Release Packages" workflow action: https://github.com/valkey-io/valkey/actions/workflows/build-release-packages.yml.. So I can't fully test in this PR. But it should work ^^

Ps. ps. I did test `rsync -av --exclude='*.c' --exclude='*.d' --exclude='*.o' src/valkey-*` command in isolation and that works as expected!

### DIFF
--- a/.github/workflows/call-build-linux-arm-packages.yml
+++ b/.github/workflows/call-build-linux-arm-packages.yml
@@ -57,7 +57,7 @@ jobs:
       - name: Create Tarball and SHA256sums
         run: |
           TAR_FILE_NAME=valkey-${{inputs.version}}-${{matrix.distro.platform}}-${{ matrix.distro.arch}}
-          mkdir -p "$TAR_FILE_NAME/bin" "$TAR_FILE_NAME/share"          
+          mkdir -p "$TAR_FILE_NAME/bin" "$TAR_FILE_NAME/share"
           rsync -av --exclude='*.c' --exclude='*.d' --exclude='*.o' src/valkey-* "$TAR_FILE_NAME/bin/"
           cp -v /home/runner/work/valkey/valkey/COPYING "$TAR_FILE_NAME/share/LICENSE"
           tar -czvf $TAR_FILE_NAME.tar.gz $TAR_FILE_NAME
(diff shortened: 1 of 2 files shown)


roshkhatri (approved):
LGTM

.github/workflows/call-build-linux-arm-packages.yml
```diff
@@ -57,9 +57,9 @@ jobs:
       - name: Create Tarball and SHA256sums
         run: |
           TAR_FILE_NAME=valkey-${{inputs.version}}-${{matrix.distro.platform}}-${{ matrix.distro.arch}}
-          mkdir -p $TAR_FILE_NAME/bin $TAR_FILE_NAME/share
-          cp -rfv src/valkey-* $TAR_FILE_NAME/bin
-          cp -v /home/runner/work/valkey/valkey/COPYING $TAR_FILE_NAME/share/LICENSE
+          mkdir -p "$TAR_FILE_NAME/bin" "$TAR_FILE_NAME/share"
+          rsync -av --exclude='*.c' --exclude='*.d' --exclude='*.o' src/valkey-* "$TAR_FILE_NAME/bin/"
+          cp -v /home/runner/work/valkey/valkey/COPYING "$TAR_FILE_NAME/share/LICENSE"
```
zuiderkwast, line 62: Interesting (weird?) that we include COPYING as LICENSE in the container file.

zuiderkwast (approved):

roshkhatri:
Need to cherry pick this to 7.2 and 8 branch

.github/workflows/call-build-linux-arm-packages.yml
madolson: Isn't that a requirement of the 2nd clause?

roshkhatri: Yes, we need to include the LICENSE when distributing the binaries.

zuiderkwast: Yes, but the funny thing is that we rename the file.

PR #1107, opened 2024-10-02
Build binary releases with systemd support

melroy89:
- Add systemd support to the build artifact tarballs, so people can use it under systemd compatible distros. As discussed here: https://github.com/orgs/valkey-io/discussions/1103#discussioncomment-10815549. Adding `libsystemd-dev` to install and add `USE_SYSTEMD=yes` to the build.
 -  Cleanup & bring the arm & x86 workflow files in-sync. It was a bit of a mess ;) (removing `jq wget awscli` from the 'Tarball' step)

### DIFF
--- a/.github/workflows/call-build-linux-x86-packages.yml
+++ b/.github/workflows/call-build-linux-x86-packages.yml
@@ -36,7 +36,7 @@ jobs:
   build-valkey:
     # Capture source tarball and generate checksum for it
     name: Build package ${{ matrix.distro.target }} ${{ matrix.distro.arch }}
-    runs-on: 'ubuntu-latest'
+    runs-on: "ubuntu-latest"
     strategy:
       fail-fast: false
       matrix: ${{ fromJSON(inputs.build_matrix) }}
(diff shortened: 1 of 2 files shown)


.github/workflows/call-build-linux-x86-packages.yml
```diff
@@ -47,10 +47,10 @@ jobs:
           ref: ${{ inputs.version }}
 
       - name: Install dependencies
-        run: sudo apt-get update && sudo apt-get install -y build-essential libssl-dev jq wget awscli
```
hpatro, line 50: jq was unused ?

hpatro (approved):

.github/workflows/call-build-linux-x86-packages.yml
zuiderkwast: Yeah, jq isn't used for building.

zuiderkwast (approved):

melroy89:
@roshkhatri You also might want to cherry pick this PR as well in 8.0.0 or 8.0??

PR #860, opened 2024-08-01
Move prepareClientToWrite out of loop for lrange command to reduce the redundant call.

zhulipeng:
## Description 
When I explore the cycles distributions for `lrange` test ( `valkey-benchmark -p 9001 -t lrange -d 100 -r 1000000 -n 1000000 -c 50 --threads 4`). I found the `prepareClientToWrite` and `clientHasPendingReplies` could be reduced to single call outside instead of called in a loop, ideally we can gain 3% performance. The corresponding `LRANG_100`, `LRANG_300`, `LRANGE_500`, `LRANGE_600` have ~2% - 3% performance boost, the benchmark test prove it helps. 

This patch try to move the `prepareClientToWrite` and its child `clientHasPendingReplies` out of the loop to reduce the function overhead.

image

## Test Environment
OPERATING SYSTEM: Ubuntu 22.04.4 LTS
Kernel: 5.15.0-116-generic
PROCESSOR: Intel Xeon Platinum 8380
Server and Client in same socket.

### Server Configuration
```
taskset -c 0-3 ~/valkey/src/valkey-server /tmp/valkey.conf

port 9001
bind * -::*
daemonize no
protected-mode no
save ""
```

### Benchmark Results 

| Test Name| Perf Boost|
|-|-|
|[memtier_benchmark-1key-list-10-elements-lrange-all-elements](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-list-10-elements-lrange-all-elements.yml)|2%|
|[memtier_benchmark-1key-list-100-elements-lrange-all-elements](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-list-100-elements-lrange-all-elements.yml)|3%|
|[memtier_benchmark-1key-list-1K-elements-lrange-all-elements](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-list-1K-elements-lrange-all-elements.yml)|2%|

PR #1119, opened 2024-10-03
Move prepareClientToWrite out of loop for HGETALL command

imasahiro:
Similar to #860 but this is for HGETALL families (HGETALL/HKEYS/HVALS). This patch moves `prepareClientToWrite` out of the loop to reduce the function overhead.

* Test Environment
    * OS: Rocky Linux release 9.4
    * Kernel: 5.14.0-427.22.1.el9_4.x86_64
    * CPU: Intel(R) Xeon(R) Silver 4210
    * Server and Client at different host/vm
* Test script
```shell
# Load generator side
for i in `seq 1 500`; do
    ./valkey-cli -h $H -p $P hset h500 "k$i" "v$i" > /dev/null
done
./valkey-benchmark -h $H -p $P --csv -d 100 -r 1000000 -n 1000000 -c 50 --threads 4 HGETALL h500
```

```
$ cat tmp.conf
bind * -::*
daemonize no
protected-mode no
save ""

./src/valkey-server ./tmp.conf
```

| test          | unstable(150c197bdd5f) rpc | this patch (rpc) | improvements |
| ------------- | -------------------------- | ---------------- | ------------ |
| HGETALL h1    | 85084.66                   | 86926.29         |  2.16446772  |
| HGETALL h10   | 78400.62                   | 76893.5          | -1.922331737 |
| HGETALL h25   | 64487.01                   | 58802.77         | -8.814550403 |
| HGETALL h50   | 47587.32                   | 49360.78         |  3.726749058 |
| HGETALL h100  | 33028.37                   | 34454.25         |  4.317137055 |
| HGETALL h300  | 14628.22                   | 15540.98         |  6.239720212 |
| HGETALL h500  | 9593.52                    | 10395.44         |  8.358975642 |


src/t_hash.c
```diff
@@ -818,23 +818,24 @@ void genericHgetallCommand(client *c, int flags) {
     robj *emptyResp = (flags & OBJ_HASH_KEY && flags & OBJ_HASH_VALUE) ? shared.emptymap[c->resp] : shared.emptyarray;
     if ((o = lookupKeyReadOrReply(c, c->argv[1], emptyResp)) == NULL || checkType(c, o, OBJ_HASH)) return;
 
+    writePreparedClient *wpc = prepareClientForFutureWrites(c);
```
rjd15372, line 818: You should handle the case when `wpc == NULL`. Since in that case no writes should be done.

src/networking.c
```diff
@@ -1138,6 +1144,13 @@ void addReplyBulkSds(client *c, sds s) {
     _addReplyToBufferOrList(c, "\r\n", 2);
 }
 
+void addWritePreparedReplyBulkSds(writePreparedClient *c, sds s) {
+    _addReplyLongLongWithPrefix(c, sdslen(s), '$');
+    _addReplyToBufferOrList(c, s, sdslen(s));
+    sdsfree(s);
+    _addReplyToBufferOrList(c, "\r\n", 2);
+}
```
madolson: I just realized while reading this code we aren't actually getting the type protections we thought we were getting. Since we need to do the cast from writePreparedClient to normal client. 

This code should look something like this:
```suggestion
void addWritePreparedReplyBulkSds(writePreparedClient *wpc, sds s) {
    client *c = (client *)wpc;
    _addReplyLongLongWithPrefix(c, sdslen(s), '$');
    _addReplyToBufferOrList(c, s, sdslen(s));
    sdsfree(s);
    _addReplyToBufferOrList(c, "\r\n", 2);
}
```

However, this isn't required because the definition of writePreparedClient should have been:
```
typedef struct writePreparedClient writePreparedClient;
```

Which is just an opaque pointer. Would you mind updating it and making the minor changes? Sorry to ask you to do it.

imasahiro: okay, make sense to me. applied!

madolson (approved):

PR #1126, opened 2024-10-04
Fix some unitialized fields in `client` struct

rjd15372:
This commit adds initialization code for the fields `io_last_reply_block` and `io_last_bufpos` of the `client` struct.

While in the current code flow, these fields are only accessed after being written in the `trySendWriteToIOThreads`, I discovered that they were not being initialized while doing some changes to the code flow of IO threads.

I believe it's good pratice to initialize all fields of a struct upon creation, and will avoid future bugs which are usually hard to debug.


madolson (approved):
Yeah, I agree it's better to initialize everything. It's likely free from a performance perspective as well since these operations are fast.

PR #1128, opened 2024-10-04
Adding the "-j" option in ci make commands to parallelize CI builds

### LINKED ISSUE #1123: Parallelize build process during CI
When building the process, we can use multiple cores with `-j`. See https://github.com/valkey-io/valkey/blob/dcac3e14993047845b85ad503cfa371d529256da/.github/workflows/ci.yml#L20.

We can hardcode it based off the known number of cores based off of the runner, we could also look into dynamically discovering the number of cores.

Shivshankar-Reddy:
fixes: https://github.com/valkey-io/valkey/issues/1123 

As per github documentation below is core information on runners.

**Linux:**
public repositories: 4 cores 
private repositories: 2 cores

**Macos:**
its 3 or 4 based on both and its depends on the Processor.

**Reference details for more information:**  Discussion in https://github.com/valkey-io/valkey/issues/1123

- Public repo: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

- Private repo: https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for--private-repositories

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       - name: make
         # Fail build if there are warnings
         # build with TLS just for compilation coverage
-        run: make all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes
+        run: make -j4 all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes
       - name: test
         run: |
           sudo apt-get install tcl8.6 tclx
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):

pizhenwei:
@Shivshankar-Reddy 
It would be thankful if you add "Suggested-by: zhenwei pi <pizhenwei@bytedance.com>" in commit message.

madolson (approved):

PR #1131, opened 2024-10-06
Fix primary crash when processing dirty slots during shutdown wait / failover wait / client pause

enjoy-binbin:
We have an assert in propagateNow. If the primary node receives a
CLUSTER UPDATE such as dirty slots during SIGTERM waitting or during
a manual failover pausing or during a client pause, the delKeysInSlot
call will trigger this assert and cause primary crash.

In this case, we added a new server_del_keys_in_slot state just like
client_pause_in_transaction to track the state to avoid the assert
in propagateNow, the dirty slots will be deleted in the end without
affecting the data consistency.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -6040,6 +6040,9 @@ void removeChannelsInSlot(unsigned int slot) {
 unsigned int delKeysInSlot(unsigned int hashslot) {
     if (!kvstoreDictSize(server.db->keys, hashslot)) return 0;
 
+    /* We may lose a slot during the pause. We need to track this
+     * state so that we don't assert in propagateNow(). */
+    server.server_del_keys_in_slot = 1;
     unsigned int j = 0;
 
     kvstoreDictIterator *kvs_di = NULL;
@@ -6064,6 +6067,7 @@ unsigned int delKeysInSlot(unsigned int hashslot) {
     }
     kvstoreReleaseDictIterator(kvs_di);
 
+    server.server_del_keys_in_slot = 0;
     return j;
 }
 
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -6068,6 +6068,7 @@ unsigned int delKeysInSlot(unsigned int hashslot) {
     kvstoreReleaseDictIterator(kvs_di);
 
     server.server_del_keys_in_slot = 0;
+    serverAssert(server.execution_nesting == 0);
     return j;
 }
 
--- a/src/networking.c
+++ b/src/networking.c
@@ -4573,11 +4573,6 @@ void unpauseActions(pause_purpose purpose) {
     updatePausedActions();
 }
 
-/* Returns 1 if clients are currently paused for the specified purpose. */
-int isPausedPurpose(pause_purpose purpose) {
-    return server.client_pause_per_purpose[purpose].end != 0;
-}
-
 /* Returns bitmask of paused actions */
 uint32_t isPausedActions(uint32_t actions_bitmask) {
     return (server.paused_actions & actions_bitmask);
--- a/src/networking.c
+++ b/src/networking.c
@@ -4548,9 +4548,9 @@ static void pauseClientsByClient(mstime_t endTime, int isPauseClientAll) {
  * The function always succeed, even if there is already a pause in progress.
  * The new paused_actions of a given 'purpose' will override the old ones and
  * end time will be updated if new end time is bigger than currently configured */
-void pauseActions(pause_purpose purpose, mstime_t end, uint32_t actions_bitmask) {
+void pauseActions(pause_purpose purpose, mstime_t end, uint32_t actions) {
     /* Manage pause type and end time per pause purpose. */
-    server.client_pause_per_purpose[purpose].paused_actions = actions_bitmask;
+    server.client_pause_per_purpose[purpose].paused_actions = actions;
 
     /* If currently configured end time bigger than new one, then keep it */
     if (server.client_pause_per_purpose[purpose].end < end) server.client_pause_per_purpose[purpose].end = end;
--- a/src/networking.c
+++ b/src/networking.c
@@ -4536,7 +4536,7 @@ static void pauseClientsByClient(mstime_t endTime, int isPauseClientAll) {
 }
 
 /* Pause actions up to the specified unixtime (in ms) for a given type of
- * commands.
+ * purpose.
  *
  * A main use case of this function is to allow pausing replication traffic
  * so that a failover without data loss to occur. Replicas will continue to receive
@@ -4548,9 +4548,9 @@ static void pauseClientsByClient(mstime_t endTime, int isPauseClientAll) {
  * The function always succeed, even if there is already a pause in progress.
  * The new paused_actions of a given 'purpose' will override the old ones and
  * end time will be updated if new end time is bigger than currently configured */
-void pauseActions(pause_purpose purpose, mstime_t end, uint32_t actions) {
+void pauseActions(pause_purpose purpose, mstime_t end, uint32_t actions_bitmask) {
     /* Manage pause type and end time per pause purpose. */
-    server.client_pause_per_purpose[purpose].paused_actions = actions;
+    server.client_pause_per_purpose[purpose].paused_actions = actions_bitmask;
 
     /* If currently configured end time bigger than new one, then keep it */
     if (server.client_pause_per_purpose[purpose].end < end) server.client_pause_per_purpose[purpose].end = end;
@@ -4571,6 +4571,11 @@ void unpauseActions(pause_purpose purpose) {
     server.client_pause_per_purpose[purpose].end = 0;
     server.client_pause_per_purpose[purpose].paused_actions = 0;
     updatePausedActions();
+}
+
+/* Returns 1 if the specified purpose is currently pausing. */
+int isPausedPurpose(pause_purpose purpose) {
+    return server.client_pause_per_purpose[purpose].end != 0;
 }
 
 /* Returns bitmask of paused actions */
--- a/src/networking.c
+++ b/src/networking.c
@@ -4573,7 +4573,7 @@ void unpauseActions(pause_purpose purpose) {
     updatePausedActions();
 }
 
-/* Returns 1 if the specified purpose is currently pausing. */
+/* Returns 1 if clients are currently paused for the specified purpose. */
 int isPausedPurpose(pause_purpose purpose) {
     return server.client_pause_per_purpose[purpose].end != 0;
 }
--- a/src/server.c
+++ b/src/server.c
@@ -3243,7 +3243,7 @@ static void propagateNow(int dbid, robj **argv, int argc, int target) {
      * replica pause (otherwise data may be lost during a failover) */
     int is_pause_replica_action = isPausedActions(PAUSE_ACTION_REPLICA);
     int is_pause_shutdown_purpose = isPausedPurpose(PAUSE_DURING_SHUTDOWN);
-    serverAssert(!(is_pause_replica_action && !is_pause_shutdown_purpose && !server.client_pause_in_transaction));
+    serverAssert(!is_pause_replica_action || is_pause_shutdown_purpose || server.client_pause_in_transaction);
 
     if (server.aof_state != AOF_OFF && target & PROPAGATE_AOF) feedAppendOnlyFile(dbid, argv, argc);
     if (target & PROPAGATE_REPL) replicationFeedReplicas(dbid, argv, argc);
--- a/src/server.c
+++ b/src/server.c
@@ -3252,10 +3252,16 @@ static void propagateNow(int dbid, robj **argv, int argc, int target) {
      * 2. The primary waits for the replica to replicate before exiting, see
      * `shutdown-timeout` in conf for more details. In this case, if the primary
      * has lost a slot during the SIGTERM waiting, it will delete all the keys
-     * and replicates DEL to its replica. */
+     * and replicates DEL to its replica.
+     *
+     * 3. The primary waits for the replica to replicate during a manual failover.
+     * In this case, if the primary has lost a slot during the pausing, it will
+     * delete all the keys and replicates DEL to its replica. */
     int is_pause_replica_action = isPausedActions(PAUSE_ACTION_REPLICA);
     int is_pause_shutdown_purpose = isPausedPurpose(PAUSE_DURING_SHUTDOWN);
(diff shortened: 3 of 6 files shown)


src/networking.c
```diff
@@ -4573,6 +4573,11 @@ void unpauseActions(pause_purpose purpose) {
     updatePausedActions();
 }
 
+/* Returns 1 if the specified purpose is currently pausing. */
```
zuiderkwast:
```suggestion
/* Returns 1 if clients are currently paused for the specified purpose. */
```

src/networking.c
```diff
@@ -4548,9 +4548,9 @@ static void pauseClientsByClient(mstime_t endTime, int isPauseClientAll) {
  * The function always succeed, even if there is already a pause in progress.
  * The new paused_actions of a given 'purpose' will override the old ones and
  * end time will be updated if new end time is bigger than currently configured */
-void pauseActions(pause_purpose purpose, mstime_t end, uint32_t actions) {
+void pauseActions(pause_purpose purpose, mstime_t end, uint32_t actions_bitmask) {
```
zuiderkwast: Is it a bitmask?

I think it's a bitmap. A mask is used for masking, setting some bits to 1 or 0.

Maybe this variable rename is not necessary?

tests/unit/cluster/slot-ownership.tcl
```diff
@@ -59,3 +59,40 @@ start_cluster 2 2 {tags {external:skip cluster}} {
...
 }
+
+start_cluster 3 1 {tags {external:skip cluster} overrides {shutdown-timeout 100}} {
+    test "Primary will not crash when processing dirty slots during shutdown waiting" {
+        R 0 set FOO 0
+
+        # Pause the replica.
+        pause_process [srv -3 pid]
+
+        # Incr the key and immediately shutdown the primary.
```
zuiderkwast, line 70: Some explanation to help the readers if they are not familiar with the shutdown procedure.

```suggestion
        # Incr the key and immediately shutdown the primary.
        # The primary waits for the replica to replicate before exiting.
```

src/server.c
```diff
@@ -3241,7 +3241,9 @@ static void propagateNow(int dbid, robj **argv, int argc, int target) {
 
     /* This needs to be unreachable since the dataset should be fixed during
      * replica pause (otherwise data may be lost during a failover) */
-    serverAssert(!(isPausedActions(PAUSE_ACTION_REPLICA) && (!server.client_pause_in_transaction)));
+    int is_pause_replica_action = isPausedActions(PAUSE_ACTION_REPLICA);
+    int is_pause_shutdown_purpose = isPausedPurpose(PAUSE_DURING_SHUTDOWN);
+    serverAssert(!(is_pause_replica_action && !is_pause_shutdown_purpose && !server.client_pause_in_transaction));
```
zuiderkwast: This was hard for me to read. This simplification is equivalent but easier to read, or wdyt?

```suggestion
    serverAssert(!is_pause_replica_action || is_pause_shutdown_purpose || server.client_pause_in_transaction);
```

Please update the comment about this assert to, to make it easier to understand the extra cases about shutdown and in-transaction.

zuiderkwast (commented):
Great fix! It was hard to understand first, but the test case helped me. I have some comments about code comments to make it easier for readers to understand.

zuiderkwast:
To be clear, is this what happens:

When a primary has lost a slot, it deletes all the keys and replicates DEL to its replicas? That's why there is some commands to replica even when write commands have been paused.

I think it can also happen (lost a slot) while a primary is waiting for manual failover, right?

src/networking.c
enjoy-binbin: I changed it because it is named like this in server.h
```
void pauseActions(pause_purpose purpose, mstime_t end, uint32_t actions_bitmask);
```

enjoy-binbin: i rename it back to actions.

enjoy-binbin:
> When a primary has lost a slot, it deletes all the keys and replicates DEL to its replicas? That's why there is some commands to replica even when write commands have been paused.

yes, it's the case.

> I think it can also happen (lost a slot) while a primary is waiting for manual failover, right?

right, yes, it can also happen i guess, will try to add a test case to see

src/server.c
enjoy-binbin: yes, the new code is better to understand, i updated the code and the comment.

src/networking.c
zuiderkwast: Ah, ok, maybe bitmask can be ok too... Not important..

src/server.c
```diff
@@ -3243,7 +3243,7 @@ static void propagateNow(int dbid, robj **argv, int argc, int target) {
      * replica pause (otherwise data may be lost during a failover) */
```
zuiderkwast: This comment doesn't mention that we can replicate DEL even when clients writes are paused. I think we should mention when and why it can happen.

enjoy-binbin:
@zuiderkwast thanks, good catch with the manual failover one, i added a new test that can confirm it.

enjoy-binbin:
Thinking of backporting it to 7.2? I added the label, let me know if we don't want to backport it.

src/server.c
```diff
@@ -3252,10 +3252,16 @@ static void propagateNow(int dbid, robj **argv, int argc, int target) {
...
+     *
+     * 3. The primary waits for the replica to replicate during a manual failover.
+     * In this case, if the primary has lost a slot during the pausing, it will
+     * delete all the keys and replicates DEL to its replica. */
     int is_pause_replica_action = isPausedActions(PAUSE_ACTION_REPLICA);
     int is_pause_shutdown_purpose = isPausedPurpose(PAUSE_DURING_SHUTDOWN);
-    serverAssert(!is_pause_replica_action || server.client_pause_in_transaction || is_pause_shutdown_purpose);
+    int is_pause_failover_purpose = isPausedPurpose(PAUSE_DURING_FAILOVER);
+    serverAssert(!is_pause_replica_action || server.client_pause_in_transaction || is_pause_shutdown_purpose ||
+                 is_pause_failover_purpose);
```
zuiderkwast: Now, there is only one more purpose: Paused by the CLIENT PAUSE command.

I guess a slot can be lost even when paused by CLIENT PAUSE, right? It can happen for any pause purpose, so just delete the assert?

How do we handle active expire? Is it prevented during pause? If not, this too can replicate DEL...

enjoy-binbin: we will also pause the expire, all the pause call like WRITE_SET and ALL_SET contains it.
```
/* Actions pause types */
#define PAUSE_ACTION_CLIENT_WRITE (1 << 0)
#define PAUSE_ACTION_CLIENT_ALL (1 << 1) /* must be bigger than PAUSE_ACTION_CLIENT_WRITE */
#define PAUSE_ACTION_EXPIRE (1 << 2)
#define PAUSE_ACTION_EVICT (1 << 3)
#define PAUSE_ACTION_REPLICA (1 << 4) /* pause replica traffic */

/* common sets of actions to pause/unpause */
#define PAUSE_ACTIONS_CLIENT_WRITE_SET \
    (PAUSE_ACTION_CLIENT_WRITE | PAUSE_ACTION_EXPIRE | PAUSE_ACTION_EVICT | PAUSE_ACTION_REPLICA)
#define PAUSE_ACTIONS_CLIENT_ALL_SET \
    (PAUSE_ACTION_CLIENT_ALL | PAUSE_ACTION_EXPIRE | PAUSE_ACTION_EVICT | PAUSE_ACTION_REPLICA)
```

yes, the client pause can also trigger this, i also thought of this last night after  turned off my mac (it was the bedtime). I guess we need a new var just like client_pause_in_transaction to track the state.

enjoy-binbin:
We don't seem to expose in info that the server may be in pause state and its purpose, do we need it?

src/cluster_legacy.c
```diff
@@ -6064,6 +6067,7 @@ unsigned int delKeysInSlot(unsigned int hashslot) {
     }
     kvstoreReleaseDictIterator(kvs_di);
 
+    server.server_del_keys_in_slot = 0;
```
zuiderkwast, line 6066: The `propagateNow()` is triggered by the call to `postExecutionUnitOperations()` above, but only if there are no nested execution units. If `delKeysInSlot` is called from inside another execution unit, this will not work. Maybe we should assert here `assert(server.execution_nesting == 0)`?

src/server.c
```diff
@@ -3245,23 +3245,23 @@ static void propagateNow(int dbid, robj **argv, int argc, int target) {
...
+     * Note that case 2 can happen in one of the following scenarios:
+     * 1) The primary waits for the replica to replicate before exiting, see
+     *    shutdown-timeout in conf for more details. In this case, primary lost
+     *    a slot during the SIGTERM waiting.
+     * 2) The primary waits for the replica to replicate during a manual failover.
+     *    In this case, primary lost a slot during the pausing.
+     * 3) The primary was paused by CLIENT PAUSE, and lost a slot during the
+     *    pausing. */
+    serverAssert(!isPausedActions(PAUSE_ACTION_REPLICA) || server.client_pause_in_transaction ||
+                 server.server_del_keys_in_slot);
```
zuiderkwast, line 3264: You're right, it's good to introduce this variable if we want to have a correct assert. It is very similar to `client_pause_in_transaction`. Actually they have the same purpose: to bypass this assert.

I don't really like that propagateNow has to be aware of special features like cluster. Maybe we can use just one variable `allow_writes_during_client_pause` for both of these?

I want to avoid complexity, but I'm not sure what is the best way to do it. Can we just remove the assert and use test cases instead to check that eviction and expire don't replicate during client pause?

src/cluster_legacy.c
enjoy-binbin: adding a assert here seem not harm, i can add it later.

src/server.c
enjoy-binbin: yes, agree, a `allow_writes_during_client_pause` for both cases should be enough, i can update it later.

> Can we just remove the assert and use test cases instead to check that eviction and expire don't replicate during client pause?

remove the assert? i think i can add some tests to cover the expire and eviction, but this assertion seems to make sense when added (for protection?), it was added in https://github.com/valkey-io/valkey/commit/47579bdf5cccfbac1cf2b6e7e52218c5b9faed86

zuiderkwast: OK, let's keep the assert. :)

Maybe `allow_writes_during_client_pause` has other problems? In `call()` we have this:

```C
    /* Client pause takes effect after a transaction has finished. This needs
     * to be located after everything is propagated. */
    if (!server.in_exec && server.client_pause_in_transaction) {
        server.client_pause_in_transaction = 0;
    }
```

If we call this variable `allow_writes_during_client_pause`, then we can't be sure that it was set because of client pause in multi-exec, or because of some other purpose (like delKeysInSlot). So the code doesn't know if it's safe to set it to 0 here.

So maybe one variable for each purpose is safer. Better separation. (We can use only one bit instead of one int for each. That's a future optimization.)

zuiderkwast (approved):
Let's merge it.

src/server.c
```diff
@@ -3240,8 +3240,28 @@ static void propagateNow(int dbid, robj **argv, int argc, int target) {
     if (!shouldPropagate(target)) return;
 
     /* This needs to be unreachable since the dataset should be fixed during
-     * replica pause (otherwise data may be lost during a failover) */
-    serverAssert(!(isPausedActions(PAUSE_ACTION_REPLICA) && (!server.client_pause_in_transaction)));
+     * replica pause (otherwise data may be lost during a failover).
+     *
+     * Thought, there are exceptions:
```
```suggestion
     * Though, there are exceptions:
```

src/server.c
```diff
@@ -3240,8 +3240,28 @@ static void propagateNow(int dbid, robj **argv, int argc, int target) {
...
+     *    slots will be deleted in the end without affecting the data consistency.
+     *
+     * Note that case 2 can happen in one of the following scenarios:
+     * 1) The primary waits for the replica to replicate before exiting, see
+     *    shutdown-timeout in conf for more details. In this case, primary lost
+     *    a slot during the SIGTERM waiting.
+     * 2) The primary waits for the replica to replicate during a manual failover.
+     *    In this case, primary lost a slot during the pausing.
+     * 3) The primary was paused by CLIENT PAUSE, and lost a slot during the
+     *    pausing. */
```
madolson, line 3262: These two feel like the same case, it's the same underlying pause mechanism. Also, will this orphan data on the replica? It would make more sense to me to merge them.

src/server.c
```diff
@@ -3240,8 +3240,28 @@ static void propagateNow(int dbid, robj **argv, int argc, int target) {
...
+     *
+     * Thought, there are exceptions:
+     *
+     * 1. We allow write commands that were queued up before and after to
+     *    execute, if a CLIENT PAUSE executed during a transaction, we will
+     *    track the state, the CLIENT PAUSE takes effect only after a transaction
+     *    has finished.
+     * 2. Primary loses a slot during the pause, deletes all keys and replicates
+     *    DEL to its replicas. In this case, we will track the state, the dirty
+     *    slots will be deleted in the end without affecting the data consistency.
```
madolson, line 3253: I think the replica will hold on to the orphaned data though. I think the replica will take over, not take over the slot mentioned (since someone else has it) and not trigger either of the paths that has extra data in it. 

You're right it's not a correctness issue though.

madolson (approved):
The code looks fine, I just looked at the code and comments though. I did not look at the tests.

src/server.c
zuiderkwast: Even `1)` is the same underlying mechanism then? They're just different triggers of the client pause.

PingXie (approved):
LGTM

PingXie:
> Thinking of backporting it to 7.2? I added the label, let me know if we don't want to backport it.

Has there been a report that this issue occurred on 7.2?

enjoy-binbin:
> Has there been a report that this issue occurred on 7.2?

Nope

src/server.c
enjoy-binbin: Can you elaborate this a bit more here? I feel like i got the point, but i am afraid i will get it wrong, so before we merge it, i want to make sure i get it right.

src/server.c
enjoy-binbin: so should we need to update the comment here?

hpatro:
> > Has there been a report that this issue occurred on 7.2?
> 
> Nope

Should we remove it from backport? It's a bit painful to cherry pick into 7.2 as it doesn't have the cluster split code.

enjoy-binbin:
> Should we remove it from backport? It's a bit painful to cherry pick into 7.2 as it doesn't have the cluster split code.

ohh, i guess, if there is too much conflict, we can drop the backport.

PR #1132, opened 2024-10-06
Removing incorrect comment about a warning

madolson:
There is a lot of bad legacy usage of `default:` with enums, which is an anti-pattern. If you omit the default, the compiler will tell you if a new enum value was added and that it is missing from a switch statement.

Someone mentioned on another PR they used `default:` because of this warning, so just removing it, but might create an issue to do a wider cleanup.


zuiderkwast (approved):
It was added in light publish iirc. I wasn't sure why there would be a warning without it, but i didn't look into it.

PingXie (approved):
Agreed that this was an anti-pattern and we should switch to `serverPanic()`.

enjoy-binbin (approved):

PR #758, opened 2024-07-08
Async IO threads

uriyage:
This PR is 1 of 3 PRs intended to achieve the goal of 1 million requests per second, as detailed by [dan touitou](https://github.com/touitou-dan) in https://github.com/valkey-io/valkey/issues/22. This PR modifies the IO threads to be fully asynchronous, which is a first and necessary step to allow more work offloading and better utilization of the IO threads.

### Current IO threads state:

Valkey IO threads were introduced in Redis 6.0 to allow better utilization of multi-core machines. Before this, Redis was single-threaded and could only use one CPU core for network and command processing. The introduction of IO threads helps in offloading the IO operations to multiple threads.

**Current IO Threads flow:**

1. Initialization: When Redis starts, it initializes a specified number of IO threads. These threads are in addition to the main thread, each thread starts with an empty list, the main thread will populate that list in each event-loop with pending-read-clients or pending-write-clients.
2. Read Phase: The main thread accepts incoming connections and reads requests from clients. The reading of requests are offloaded to IO threads. The main thread puts the clients ready-to-read in a list and set the global io_threads_op to  IO_THREADS_OP_READ, the IO threads pick the clients up, perform the read operation and parse the first incoming command.
3. Command Processing: After reading the requests, command processing is still single-threaded and handled by the main thread.
4. Write Phase: Similar to the read phase, the write phase is also be offloaded to IO threads. The main thread prepares the response in the clients’ output buffer then the main thread puts the client in the list, and sets the global io_threads_op to  the IO_THREADS_OP_WRITE. The IO threads then pick the clients up and perform the write operation to send the responses back to clients.
5. Synchronization: The main-thread communicate with the threads on how many jobs left per each thread with atomic counter. The main-thread doesn’t access the clients while being handled by the IO threads.

**Issues with current implementation:**

* Underutilized Cores: The current implementation of IO-threads leads to the underutilization of CPU cores. 
    * The main thread remains responsible for a significant portion of IO-related tasks that could be offloaded to IO-threads.
    * When the main-thread is processing client’s commands, the IO threads are idle for a considerable amount of time.
    * Notably, the main thread's performance during the IO-related tasks is constrained by the speed of the slowest IO-thread.
* Limited Offloading: Currently, Since the Main-threads waits synchronously for the IO threads, the Threads perform only read-parse, and write operations, with parsing done only for the first command. If the threads can do work asynchronously we may offload more work to the threads reducing the load from the main-thread.
[...]

PR #1138, opened 2024-10-08
Add io-threads-do-reads config to deprecated config table to have no effect.

### LINKED ISSUE #1116: Make io-threads-do-reads harmless
This config is undocumented since #758. The default was changed to "yes" and it is quite useless to set it to "no". Yet, it can happen that some user has an old config file where it is explicitly set to "no". The result will be bad performace, since I/O threads will not do all the I/O.

----

It's indeed confusing.

1. Either remove the whole option from the code. And thus no need for documentation. _OR:_
2. Introduce the option back in the configuration,  just as a comment is fine. And showing the default value "yes": `# io-threads-do-reads yes` with additional text.

_Originally posted by @melroy89 in https://github.com/valkey-io/valkey/discussions/1019#discussioncomment-10824778_

Shivshankar-Reddy:
this fixes: https://github.com/valkey-io/valkey/issues/1116

_Issue details from #1116 by @zuiderkwast_ 

> This config is undocumented since #758. The default was changed to "yes" and it is quite useless to set it to "no". Yet, it can happen that some user has an old config file where it is explicitly set to "no". The result will be bad performace, since I/O threads will not do all the I/O.
> 
> It's indeed confusing.
> 
> 1. Either remove the whole option from the code. And thus no need for documentation. _OR:_
> 2. Introduce the option back in the configuration,  just as a comment is fine. And showing the default value "yes": `# io-threads-do-reads yes` with additional text.
> 
> _Originally posted by @melroy89 in [#1019 (reply in thread)](https://github.com/orgs/valkey-io/discussions/1019#discussioncomment-10824778)_

### DIFF
--- a/valkey.conf
+++ b/valkey.conf
@@ -1355,7 +1355,7 @@ lazyfree-lazy-user-flush yes
 # prefetch-batch-max-size 16
 #
 # NOTE:
-# 1. 'io-threads-do-reads' config is depricated and will be removed in the future,
+# 1. 'io-threads-do-reads' config is deprecated and will be removed in the future,
 # please avoid using this option if possible.
 #
 # 2. If you want to test the server speedup using valkey-benchmark, make
--- a/valkey.conf
+++ b/valkey.conf
@@ -1355,8 +1355,8 @@ lazyfree-lazy-user-flush yes
 # prefetch-batch-max-size 16
 #
 # NOTE:
-# 1. 'io-threads-do-reads' config is deprecated and will be removed in the future,
-# please avoid using this option if possible.
+# 1. The 'io-threads-do-reads' config is deprecated and has no effect.
+# It will be removed in the future. Please avoid using this option if possible.
 #
 # 2. If you want to test the server speedup using valkey-benchmark, make
 # sure you also run the benchmark itself in threaded mode, using the
--- a/valkey.conf
+++ b/valkey.conf
@@ -1354,7 +1354,11 @@ lazyfree-lazy-user-flush yes
 #
 # prefetch-batch-max-size 16
 #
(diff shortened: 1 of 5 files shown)


melroy89 (commented):
Again please update documentation and example config file explaining this option is deprecated and doesn't do anything anymore.

Shivshankar-Reddy:
Added a note in the valkey.conf file as it is deprecated and avoid using it in future. Thanks for remainding it.

melroy89:
Small typo in title: depricated -> deprecated

zuiderkwast (approved):

valkey.conf
```diff
@@ -1354,7 +1354,11 @@ lazyfree-lazy-user-flush yes
 #
 # prefetch-batch-max-size 16
 #
-# NOTE: If you want to test the server speedup using valkey-benchmark, make
+# NOTE:
+# 1. 'io-threads-do-reads' config is deprecated and will be removed in the future,
+# please avoid using this option if possible.
```
zuiderkwast: Mention also that it has no effect? (Some deprecated features still work, but this one doesn't.)

```suggestion
# 1. The 'io-threads-do-reads' config is deprecated and has no effect. It 
will be removed in the future. Please avoid using this option if possible.
```

madolson: We won't necessarily remove it in the feature, we shouldn't use such language.

zuiderkwast: Sorry for merging too fast. Shall we just delete the sentence "will be removed in the future"? A follow-up PR is welcome.

Shivshankar-Reddy: Sure @madolson, I have noted your point and created https://github.com/valkey-io/valkey/pull/1150 to revoke the statement from valkey.conf.

PR #1144, opened 2024-10-10
Add a key update notification

### LINKED ISSUE #1143: [NEW] Add a dirty key notification
In some cases, the user want to know if some key has been changed.

Added a new keyspace notification type, such as `dirty key notification`.  When a key is modified,  notify a keyspace notification.

jjz921024:
this pr want to close #1143

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -2906,7 +2906,7 @@ static int setConfigNotifyKeyspaceEventsOption(standardConfig *config, sds *argv
     }
     int flags = keyspaceEventsStringToFlags(argv[0]);
     if (flags == -1) {
-        *err = "Invalid event class character. Use 'Ag$lshzxeKEtmdn'.";
+        *err = "Invalid event class character. Use 'Ag$lshzxeKEtmdnu'.";
         return 0;
     }
     server.notify_keyspace_events = flags;
--- a/src/notify.c
+++ b/src/notify.c
@@ -58,6 +58,7 @@ int keyspaceEventsStringToFlags(char *classes) {
         case 'm': flags |= NOTIFY_KEY_MISS; break;
         case 'd': flags |= NOTIFY_MODULE; break;
         case 'n': flags |= NOTIFY_NEW; break;
+        case 'u': flags |= NOTIFY_UPDATE; break;
         default: return -1;
         }
     }
@@ -86,6 +87,7 @@ sds keyspaceEventsFlagsToString(int flags) {
         if (flags & NOTIFY_STREAM) res = sdscatlen(res, "t", 1);
         if (flags & NOTIFY_MODULE) res = sdscatlen(res, "d", 1);
         if (flags & NOTIFY_NEW) res = sdscatlen(res, "n", 1);
+        if (flags & NOTIFY_UPDATE) res = sdscatlen(res,"u",1);
     }
     if (flags & NOTIFY_KEYSPACE) res = sdscatlen(res, "K", 1);
     if (flags & NOTIFY_KEYEVENT) res = sdscatlen(res, "E", 1);
--- a/src/db.c
+++ b/src/db.c
@@ -636,6 +636,7 @@ long long dbTotalServerKeyCount(void) {
 void signalModifiedKey(client *c, serverDb *db, robj *key) {
     touchWatchedKey(db, key);
     trackingInvalidateKey(c, key, 1);
+    notifyKeyspaceEvent(NOTIFY_UPDATE, "update", key, db->id);
 }
 
 void signalFlushedDb(int dbid, int async) {
--- a/src/server.h
+++ b/src/server.h
@@ -615,7 +615,7 @@ typedef enum {
 #define NOTIFY_LOADED (1 << 12)   /* module only key space notification, indicate a key loaded from rdb */
 #define NOTIFY_MODULE (1 << 13)   /* d, module key space notification */
 #define NOTIFY_NEW (1 << 14)      /* n, new key notification */
-#define NOTIFY_UPDATE (1<<15)      /* c, dirty key notification */
+#define NOTIFY_UPDATE (1<<15)      /* u, update key notification */
 #define NOTIFY_ALL                                                                                            \
     (NOTIFY_GENERIC | NOTIFY_STRING | NOTIFY_LIST | NOTIFY_SET | NOTIFY_HASH | NOTIFY_ZSET | NOTIFY_EXPIRED | \
      NOTIFY_EVICTED | NOTIFY_STREAM | NOTIFY_MODULE) /* A flag */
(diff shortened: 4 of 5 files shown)


src/server.c
```diff
@@ -3537,6 +3537,10 @@ void call(client *c, int flags) {
     dirty = server.dirty - dirty;
     if (dirty < 0) dirty = 0;
 
+    if (dirty > 0) {
+        notifyKeyspaceEvent(NOTIFY_DIRTY, "dirty", c->argv[1], c->db->id);
```
ranshid: This will not target specific key, but only indicate the dataset is dirty, there are many multi key commands (like mset, eval etc...) which will make many keys dirty. I think maybe a better place is to place it in `signalModifiedKey`

src/notify.c
```diff
@@ -58,6 +58,7 @@ int keyspaceEventsStringToFlags(char *classes) {
         case 'm': flags |= NOTIFY_KEY_MISS; break;
         case 'd': flags |= NOTIFY_MODULE; break;
         case 'n': flags |= NOTIFY_NEW; break;
+        case 'c': flags |= NOTIFY_DIRTY; break;
```
ranshid: Personally I do not find the 'dirty' as the correct state here. IMO dirty means that the key was modified and was not yet consisted (like for example successfully written to the replica).
Although funny name but maybe: NOTIFY_MODIFY is a better match for what really happened? 
Also maybe NOTIFY_UPDATE?

src/config.c
```diff
@@ -2906,7 +2906,7 @@ static int setConfigNotifyKeyspaceEventsOption(standardConfig *config, sds *argv
     }
     int flags = keyspaceEventsStringToFlags(argv[0]);
     if (flags == -1) {
-        *err = "Invalid event class character. Use 'Ag$lshzxeKEtmdn'.";
+        *err = "Invalid event class character. Use 'Ag$lshzxeKEtmdnc'.";
```
ranshid: why 'c'? maybe 'u' (stands for updated)?

ranshid:
@jjz921024 you will have to sign your commit in order for the DCO to pass.

src/server.c
jjz921024: Thank you for your guidance. I had put it in `signalModifiedKey`.

src/notify.c
jjz921024: I perfer NOTIFY_UPDATE, because we can use 'u' to stand it. (The letter 'm' has been used)

ranshid:
@valkey-io/core-team since this is an interface/config change, tagged it as major decision. can you please take a look?

PingXie:
> @valkey-io/core-team since this is an interface/config change, tagged it as major decision. can you please take a look?

How is this new setting different from `g$lshzxet` (or just `g$lshzxe` if stream events are not interesting)? Are there any specific updates that are not covered by the existing event types?  The way it is implemented now seems to be a super set of the existing event types.

ranshid:
> > @valkey-io/core-team since this is an interface/config change, tagged it as major decision. can you please take a look?
> 
> How is this new setting different from `g$lshzxet` (or just `g$lshzxe` if stream events are not interesting)? Are there any specific updates that are not covered by the existing event types? The way it is implemented now seems to be a super set of the existing event types.

I agree. this will basically aggregate different events under the same event name. mainly to simplify the client side logic and can be achieved without this. However that got me thinking if we see any value to provide some "update" notification indicating an existing value was changed (as apposed to created/deleted/expired etc...). however that would require a different change introduced.

jjz921024:
> > @valkey-io/core-team since this is an interface/config change, tagged it as major decision. can you please take a look?
> 
> How is this new setting different from `g$lshzxet` (or just `g$lshzxe` if stream events are not interesting)? Are there any specific updates that are not covered by the existing event types? The way it is implemented now seems to be a super set of the existing event types.

@PingXie Hi, It might be different with `g$lshzxet`. The existing event type does not contain events for which the key is modified. Please let me know if I'm wrong.

jjz921024:
> > > @valkey-io/core-team since this is an interface/config change, tagged it as major decision. can you please take a look?
> > 
> > 
> > How is this new setting different from `g$lshzxet` (or just `g$lshzxe` if stream events are not interesting)? Are there any specific updates that are not covered by the existing event types? The way it is implemented now seems to be a super set of the existing event types.
> 
> I agree. this will basically aggregate different events under the same event name. mainly to simplify the client side logic and can be achieved without this. However that got me thinking if we see any value to provide some "update" notification indicating an existing value was changed (as apposed to created/deleted/expired etc...). however that would require a different change introduced.

@ranshid Hi, It will aggregate different events under the same event name. But, if we want to provide update notification when a key has been modify. Maybe we should introduce it.

PingXie:
>The existing event type does not contain events for which the key is modified. Please let me know if I'm wrong

I assume you meant "value modification" rather than "key modification" here? The current key notification mechanism breaks events down by data/command types like SET, HSET, etc. From my quick look at the source code, it appears to cover all relevant operations. Are you seeing any "update" operations being missed? I'm cautious about introducing another "superset" event, as it might add unnecessary complexity/overhead. 

BTW, we do have "rename" (or "key modification") events too.

ranshid:
> BTW, we do have "rename" (or "key modification") events too.

Yes we do. we have rename_from and rename_to events

ranshid:
> > The existing event type does not contain events for which the key is modified. Please let me know if I'm wrong
> 
> I assume you meant "value modification" rather than "key modification" here? The current key notification mechanism breaks events down by data/command types like SET, HSET, etc. From my quick look at the source code, it appears to cover all relevant operations. Are you seeing any "update" operations being missed? I'm cautious about introducing another "superset" event, as it might add unnecessary complexity/overhead.

I think, as @PingXie correctly stated, that the current events ARE reflecting most of the mutative data commands. However currently there is not easy way for the client to understand a VALUE was changed other then combining the "new" notification and all other different events which are per operation/datatype. I think it might simplify the client logic to be able to have event notification for "modified" value but as I said this IS currently possible to achieve by clients IMO.

hwware:
I have the similar concern as what Ping's: for existing commands, such as saddCommand function, we already have NOTIFY_SET notification, if we introduce Key-update, do we have redundancy behavior for this command?

soloestoy:
I think the current notifications already covers all "update" operations, so there's no need to add a redundant event.

madolson:
@jjz921024 Can you add more information about what specific problem you are trying to solve. The core team discussed the issue offline, and it feels unnecessary, but we might not be understanding the use case. It seems like you could achieve the same result by just subscribing to all the events and then just ignoring what type of event was being sent.

jjz921024:
Hi, I tried the `AKE` notification types and it did cover all `update` events，I'm sorry I didn't look into it. This pr may be clsoed.

PingXie:
Thanks @jjz921024!

PR #1146, opened 2024-10-10
Fix scan skipping elements while hashset is shrink rehashing

rainsupreme:
There was an optimization to skip unnecessary buckets based on the scan's position relative to rehash progress, but it assumed that we're rehashing to a larger size. I fixed it for shrink-rehashes, and also expanded on the idea to only check the before-rehash table when we're ahead of the rehash cursor, and only check the after-rehash table when we're behind the rehash cursor. If we're on top of the rehash cursor then we need to emit from both tables.

I also fixed the HASHSET_SCAN_SINGLE_STEP flag.


rainsupreme:
oops, had the wrong branch targeted

PR #1147, opened 2024-10-10
[hashset feature] Iterator bugfixes - handle empty hashset and change iteration order so rehashed buckets can be skipped

rainsupreme:
Fixes 2 bugs in hashsetNext():

- null dereference when iterating over new unused hashset
- change order of iteration so skipping past already-rehashed buckets works correctly and won't miss elements

Minor optimization in hashsetScan and some renamed variables.

### DIFF
--- a/src/hashset.c
+++ b/src/hashset.c
@@ -1296,27 +1296,28 @@ size_t hashsetScan(hashset *t, size_t cursor, hashsetScanFunction fn, void *priv
             /* Advance cursor. */
             cursor = nextCursor(cursor, mask);
         } else {
-            /* Let table0 be the the smaller table and table1 the bigger one. */
-            int table0, table1;
+            int table_small, table_large;
             if (t->bucketExp[0] <= t->bucketExp[1]) {
-                table0 = 0;
-                table1 = 1;
+                table_small = 0;
+                table_large = 1;
             } else {
-                table0 = 1;
-                table1 = 0;
+                table_small = 1;
+                table_large = 0;
             }
 
-            size_t mask0 = expToMask(t->bucketExp[table0]);
-            size_t mask1 = expToMask(t->bucketExp[table1]);
+            size_t mask_small = expToMask(t->bucketExp[table_small]);
+            size_t mask_large = expToMask(t->bucketExp[table_large]);
 
             /* Emit elements in the smaller table, if this bucket hasn't already
              * been rehashed. */
-            if (table0 == 0 && !cursorIsLessThan(cursor, t->rehashIdx)) {
-                bucket *b = &t->tables[table0][cursor & mask0];
-                for (int pos = 0; pos < ELEMENTS_PER_BUCKET; pos++) {
-                    if (b->presence & (1 << pos)) {
-                        void *emit = emit_ref ? &b->elements[pos] : b->elements[pos];
-                        fn(privdata, emit);
+            if (table_small == 0 && !cursorIsLessThan(cursor, t->rehashIdx)) {
+                bucket *b = &t->tables[table_small][cursor & mask_small];
+                if (b->presence) {
+                    for (int pos = 0; pos < ELEMENTS_PER_BUCKET; pos++) {
+                        if (b->presence & (1 << pos)) {
+                            void *emit = emit_ref ? &b->elements[pos] : b->elements[pos];
+                            fn(privdata, emit);
+                        }
                     }
                 }
                 in_probe_sequence |= b->everfull;
@@ -1325,21 +1326,23 @@ size_t hashsetScan(hashset *t, size_t cursor, hashsetScanFunction fn, void *priv
             /* Iterate over indices in larger table that are the expansion of
              * the index pointed to by the cursor in the smaller table. */
             do {
-                /* Emit elements in table 1. */
-                bucket *b = &t->tables[table1][cursor & mask1];
-                for (int pos = 0; pos < ELEMENTS_PER_BUCKET; pos++) {
-                    if (b->presence & (1 << pos)) {
-                        void *emit = emit_ref ? &b->elements[pos] : b->elements[pos];
-                        fn(privdata, emit);
+                /* Emit elements in bigger table. */
+                bucket *b = &t->tables[table_large][cursor & mask_large];
+                if (b->presence) {
+                    for (int pos = 0; pos < ELEMENTS_PER_BUCKET; pos++) {
+                        if (b->presence & (1 << pos)) {
+                            void *emit = emit_ref ? &b->elements[pos] : b->elements[pos];
+                            fn(privdata, emit);
+                        }
                     }
                 }
                 in_probe_sequence |= b->everfull;
 
                 /* Increment the reverse cursor not covered by the smaller mask.*/
-                cursor = nextCursor(cursor, mask1);
+                cursor = nextCursor(cursor, mask_large);
 
                 /* Continue while bits covered by mask difference is non-zero */
-            } while ((cursor & (mask0 ^ mask1)) && cursor != start_cursor);
+            } while (cursor & (mask_small ^ mask_large) && cursor != start_cursor);
         }
         if (cursor == 0) {
             cursor_passed_zero = 1;
@@ -1431,8 +1434,11 @@ int hashsetNext(hashsetIterator *iter, void **elemptr) {
             } else {
                 iter->fingerprint = hashsetFingerprint(iter->hashset);
             }
+            if (iter->hashset->tables[0] == NULL) {
+                /* empty hashset, we're done */
+                break;
+            }
             iter->index = 0;
-            /* skip the rehashed slots in table[0] */
             if (hashsetIsRehashing(iter->hashset)) {
                 iter->index = iter->hashset->rehashIdx;
             }
@@ -1442,9 +1448,8 @@ int hashsetNext(hashsetIterator *iter, void **elemptr) {
             iter->posInBucket++;
             if (iter->posInBucket >= ELEMENTS_PER_BUCKET) {
                 iter->posInBucket = 0;
-                iter->index++;
-                if (iter->index >= (long)numBuckets(iter->hashset->bucketExp[iter->table])) {
-                    iter->index = 0;
+                iter->index = nextCursor(iter->index, expToMask(iter->hashset->bucketExp[iter->table]));
+                if (iter->index == 0) {
                     if (hashsetIsRehashing(iter->hashset) && iter->table == 0) {
                         iter->table++;
                     } else {


src/hashset.c
```diff
@@ -1268,30 +1271,35 @@ size_t hashsetScan(hashset *t, size_t cursor, hashsetScanFunction fn, void *priv
...
-                /* Continue while bits covered by mask difference is non-zero */
-            } while (cursor & (mask0 ^ mask1));
+            cursor = nextCursor(cursor, mask_small);
         }
         if (cursor == 0) {
             cursor_passed_zero = 1;
         }
+        if (single_step) break;
     }
-    while (in_probe_sequence && !single_step);
```
zuiderkwast: Here, I though I had written a `do { .... } while (in_probe_sequence && !single_step);` but it was a `while { ... }` and then a dummy `while (...);`.

I have already pushed a fix that changes it to do-while.

src/hashset.c
```diff
@@ -1317,24 +1319,29 @@ size_t hashsetScan(hashset *t, size_t cursor, hashsetScanFunction fn, void *priv
...
-                in_probe_sequence |= b->everfull;
+                    in_probe_sequence |= b->everfull;
 
-                /* Increment the reverse cursor not covered by the smaller mask.*/
-                cursor = nextCursor(cursor, mask1);
+                    /* Increment the reverse cursor not covered by the smaller mask.*/
+                    big_table_cursor = nextCursor(big_table_cursor, mask_big);
+
+                    /* Continue while bits covered by mask difference is non-zero */
+                } while (big_table_cursor & (mask_small ^ mask_big));
```
zuiderkwast: Rebase missed the condition `&& cursor != start_cursor` here?

I added this condition in a recent commit to avoid looping forever if probing chains fill the whole table(s).

src/hashset.c
```diff
@@ -1317,24 +1319,29 @@ size_t hashsetScan(hashset *t, size_t cursor, hashsetScanFunction fn, void *priv
...
+            }
 
-                /* Continue while bits covered by mask difference is non-zero */
-            } while ((cursor & (mask0 ^ mask1)) && cursor != start_cursor);
+            cursor = nextCursor(cursor, mask_small);
```
zuiderkwast: Why increment cursor again? It was done above already.

I guess it's not done if we skip the big table (due to already rehashed, etc.), so it needs to be in an `else` clause?

```suggestion
        } else {
            cursor = nextCursor(cursor, mask_small);
```

src/hashset.c
```diff
@@ -1291,23 +1291,25 @@ size_t hashsetScan(hashset *t, size_t cursor, hashsetScanFunction fn, void *priv
...
 
-            /* Emit elements in the smaller table, if this bucket hasn't already
-             * been rehashed. */
-            if (table0 == 0 && !cursorIsLessThan(cursor, t->rehashIdx)) {
-                bucket *b = &t->tables[table0][cursor & mask0];
+            int behind_rehash = cursorIsLessThan(cursor & mask_small, t->rehashIdx & mask_small);
+            int ahead_rehash = cursorIsLessThan(t->rehashIdx & mask_small, cursor & mask_small);
+
+            if ((growing && !behind_rehash) || (!growing && !ahead_rehash)) {
+                /* Emit elements in smaller table */
```
zuiderkwast: How about some more explicit comment to explain this condition?

```suggestion
            /* Emit elements in smaller table, except if this table is the
             * rehashing source table and the current index has already been
             * rehashed, or the rehashing destination table and the current
             * index has not yet been populated by rehashing. */
            if ((growing && !behind_rehash) || (!growing && !ahead_rehash)) {
```

Can't it happen that we're shrinking and we're behind rehashing, but rehash put some elements before or at the current index?

If the big (rehashing source) table had a lot of empty buckets with 'everfull' set, I guess it can happen, because rehash doesn't follow probing chains like scan does.

If we make rehashing follow probing chains, are we safe here?

src/hashset.c
```diff
@@ -1317,24 +1319,29 @@ size_t hashsetScan(hashset *t, size_t cursor, hashsetScanFunction fn, void *priv
...
-            /* Iterate over indices in larger table that are the expansion of
-             * the index pointed to by the cursor in the smaller table. */
-            do {
-                /* Emit elements in table 1. */
-                bucket *b = &t->tables[table1][cursor & mask1];
-                for (int pos = 0; pos < ELEMENTS_PER_BUCKET; pos++) {
-                    if (b->presence & (1 << pos)) {
-                        void *emit = emit_ref ? &b->elements[pos] : b->elements[pos];
-                        fn(privdata, emit);
+            if ((growing && !ahead_rehash) || (!growing && !behind_rehash)) {
```
zuiderkwast: A similar comment please, to explain: already rehashed source or not yet rehashed destination.

zuiderkwast (commented):
Didn't review the tests yet.

src/unit/test_hashset.c
```diff
@@ -371,6 +371,67 @@ int test_scan(int argc, char **argv, int flags) {
     return 0;
 }
 
+void scanrehashfn(void *privdata, void *element) {
+    scandata *data = (scandata *)privdata;
+    unsigned long j = (unsigned long)element;
+    if (j < (unsigned long)data->count) {
+        data->element_seen[j]++;
+    }
+}
```
zuiderkwast: Is this identical to the old scan function? If yes, then we can just use that.

src/unit/test_hashset.c
```diff
@@ -371,6 +371,67 @@ int test_scan(int argc, char **argv, int flags) {
...
+
+    /* A set of longs, i.e. pointer-sized values. */
+    hashsetType type = {0};
+    long j;
+
+    for (int round = 0; round < num_rounds; round++) {
+        const long count = num_elements_inspected + 1 + (rand() % 1000);
+
+        /* Seed, to make sure each round is different. */
+        test_set_hash_function_seed(argc, argv, flags);
```
zuiderkwast: I created a helper for this.

```suggestion
        randomSeed();
```

src/hashset.c
rainsupreme: Actually I removed it because this loop doesn't follow probe chains so it won't loop forever :)

zuiderkwast (commented):
Cool test case, rigging a race between scan and rehash.

src/hashset.c
rainsupreme: The above sub-loop increments big_table_cursor though, not cursor

src/hashset.c
zuiderkwast: No but the outer loop can loop forever if the inner skips start cursor.

zuiderkwast: Ah, big table cursor! I see.

Then we need to mask the start cursor with small mask only, when we start, right? Otherwise, if start cursor is large, we may miss `cursor == start_cursor` and loop forever.

src/hashset.c
zuiderkwast: Ah, there is a big cursor. Ok.

src/hashset.c
zuiderkwast: Being able to always skip one of the tables is a very good thing, I believe, especially in things like the expiry scanning. Eviction does sampling which is using scan too. Therefore, to make sure it's safe to skip in the destination table, I think we need to make rehash follow probing chains. Then we can ensure that a key is not rehashed from the not-yet-scanned part of the source table to the already-scanned part of the destination table. Do you want to do this change to rehashStep in this PR?

The names like `table_small` are better than `table0`. I'm not sure that all the extra variables actually make it simpler though. Each variable adds a mental indirection, so if the names are not extremely clear and if they are just a trivial abstraction, then it might not be worth it.

I'm thinking that maybe instead of `growing` we can just use `table_small == 0`.

The names `behind_rehash` and `ahead_rehash` confuse me a little, because it can mean two things: Is the rehash ahead of cursor or is cursor ahead of rehash? I think `already_rehashed` and `not_yet_rehashed` are better names and I think it's enough to have one of them: `already_rehashed = cursorIsLessThan(t->rehashIdx, cursor)`.

We don't need to mask the cursor and `rehashIdx` here, because cursor should already be masked with the small mask and, if we make rehashing follows probe chains, the rehashIdx will already be masked with the small mask in too `rehashStep`. Maybe we don't strictly have to rehash from both locations in a larger table to the same location in the smaller destination table at the same time, but if we do this, we can guarantee that a cursor is either rehashed or not rehashed, never half-rehashed. Is this right?

src/hashset.c
```diff
@@ -1291,23 +1291,25 @@ size_t hashsetScan(hashset *t, size_t cursor, hashsetScanFunction fn, void *priv
...
-                table0 = 1;
-                table1 = 0;
+                table_small = 1;
+                table_big = 0;
             }
 
-            size_t mask0 = expToMask(t->bucketExp[table0]);
-            size_t mask1 = expToMask(t->bucketExp[table1]);
+            size_t mask_small = expToMask(t->bucketExp[table_small]);
+            size_t mask_big = expToMask(t->bucketExp[table_big]);
```
zuiderkwast: If we use the 5-letter word "large" instead of "big", it aligns better with "small". :-)

```suggestion
            size_t mask_small = expToMask(t->bucketExp[table_small]);
            size_t mask_large = expToMask(t->bucketExp[table_large]);
```

src/hashset.c
rainsupreme: Hmm. Do you mean that rehashStep would continue until there can be no probe chains ahead of its cursor in either table? It would need to track the furthest bucket touched (due to probe chains in destination table) and continue until the rehash cursor has caught up and covered that bucket, otherwise we can't guarantee that the destination table is completely empty ahead of the rehash cursor.

I feel like that would make rehash steps too large and random, and we'd lose a lot of the benefit of incremental rehashing. I think it'd be better to not skip the destination bucket when scanning.

rainsupreme: Because scan already follows probe chains, I don't think we ever have to worry about rehash moving an unscanned item behind the scan cursor because of its probe chain becoming shorter like you described. If there was an element at the end of a probe chain like that, scan would have either followed the whole potential probe chain and covered the element, or it would be before the earliest bucket it could land in without a probe chain. :)

SoftlyRaining: Oh wait... because rehash _doesn't_ follow probe chains, it could be scanning an item that had been moved forward due to a probe chain, but hashes to a bucket before the rehash cursor... so as it is, rehashing can move items to buckets both before and after the rehash cursor 😵

zuiderkwast: Interesting. I am almost convinced. I will think about it during the night. Do you want to draw an ASCII picture or something to make it look more obvious?

src/hashset.c
rainsupreme: start_cursor is "rounded down" to the mask_small increment when it's assigned, I think, which should make it ok
`cursor = cursor & (expToMask(t->bucketExp[0]) | expToMask(t->bucketExp[1]));`

src/hashset.c
rainsupreme: I've thought about it a while, and I think we can't/shouldn't shortcut anything in scan while rehashing. There are 2 shortcuts to consider:

First, we skip buckets in the old table that have already been rehashed. Because rehash processes one bucket at a time, it may have stopped in the middle of a probe chain and there are elements that haven't been rehashed that move backwards when rehashed. We have to at least read the `everfull` flags and continue scanning to guarantee we cover any elements that might potentially move backwards in a rehash. Here's the hazard illustrated:

Say there's an element Sam. Sam's been here the whole time. When we begin our first a scan step, Sam had been placed forward two buckets in the old table, and rehash is currently ahead of the bucket Sam hashes to, but hasn't reached Sam yet. Scan cursor starts at the bucket Sam hashes to.
```
      Sam (☺) hashes here
      |
old:  E--E--☺--.
new:  .--.--.--.
      |  |
      |  RehashIdx
      |
      cursor
```

If we skip accessing those old buckets with `E`verfull flags, the new table has no `E`verfull flags and we stop here:
```
      Sam (☺) hashes here
      |
old:  E--E--☺--.
new:  .--.--.--.
         |
         RehashIdx
         |
         cursor
```

Then the next time we scan we find that Sam has been rehashed "behind" the scan cursor, he's never emitted, and we've failed our scan guarantee.
```
      Sam (☺) hashes here
      |
old:  E--E--.--.
new:  ☺--.--.--.
         |     |
         |     RehashIdx
         |
         cursor
```


The second possible shortcut: We skip buckets in the new table in the area that hasn't been rehashed yet. This is possible but tricky because rehashing can place elements any distance ahead of rehashIdx due to probe chaining. If the new table is smaller then there's only one bucket on this step and we might as well just check `presence == 0`. If the new table is larger there are more buckets to potentially skip at once, but the condition is something like "large table == new table AND cursor > rehashIdx AND first bucket is empty AND first bucket `everfull` is false", which I believe is too much added complexity. We're probably not more than doubling size so we're likely only skipping 1 bucket anyway (since the condition accesses the first bucket regardless).

src/hashset.c
rainsupreme: I determined that this shortcut is invalid - see my other long comment with the ascii pictures 😅

src/hashset.c
zuiderkwast: I agree that skipping in the new table is not worth the effort.

I still want to argue that it's safe to skip already rehashed indices in the old table though, because those have been rehashed = moved to the new table. (Sam can't be behind the rehash index in the old table, but he can be behind the (index corresponding to the) rehash index in the new table.)

Regarding skipping not-yet-rehashed indices in the new table, I agree it's possible but not really worth the complexity.

My take is that it's not as complex as you say though. Assume the new table is twice as large as the old table. Then, each index x in the old table corresponds to two indices y1 and y2 in the new table. Note that these two indices are neighours though. We have `y1 = x & mask_new` and `y2 = nextCursor(y1, mask_new)`. Therefore, any probe chain staring before these indices in the new table need to involve `y1`. Thus, we only need to check `y1 > (rehashIdx & mask_new) && new_table[y1]->everfull == 0` to know if we can skip the remaining buckets in the new table in this scan call. (This allows us to skip remaining buckets in the new table when we're following in a probe chain in the old table.)

src/hashset.c
```diff
@@ -1429,16 +1423,16 @@ int hashsetNext(hashsetIterator *iter, void **elemptr) {
     while (1) {
         if (iter->index == -1 && iter->table == 0) {
             /* It's the first call to next. */
+            if (iter->hashset->tables[0] == NULL) {
+                /* empty hashset, we're done */
+                break;
+            }
```
zuiderkwast: Doesn't this break the fingerprint check in `hashsetResetIterator` for safe iterators? The fingerprint has not been initialized yet here.

Do we have a test case for iterating over an empty table using safe iterator?

src/hashset.c
```diff
@@ -1429,16 +1423,16 @@ int hashsetNext(hashsetIterator *iter, void **elemptr) {
...
             if (iter->safe) {
                 hashsetPauseRehashing(iter->hashset);
             } else {
                 iter->fingerprint = hashsetFingerprint(iter->hashset);
             }
             iter->index = 0;
-            /* skip the rehashed slots in table[0] */
-            if (hashsetIsRehashing(iter->hashset)) {
-                iter->index = iter->hashset->rehashIdx;
-            }
```
zuiderkwast: Why can't we skip already rehashed indices in table 0 (the old table)? They have been moved to table 1. Probing chains don't matter here.

src/hashset.c
```diff
@@ -1291,50 +1291,51 @@ size_t hashsetScan(hashset *t, size_t cursor, hashsetScanFunction fn, void *priv
...
             cursor = nextCursor(cursor, mask);
         } else {
-            /* Let table0 be the the smaller table and table1 the bigger one. */
-            int table0, table1;
-            if (t->bucketExp[0] <= t->bucketExp[1]) {
-                table0 = 0;
-                table1 = 1;
+            int growing = t->bucketExp[0] <= t->bucketExp[1];
+            int table_small, table_large;
+            if (growing) {
```
zuiderkwast: The `growing` variable is only used in one place. It doesn't add any clarity IMO.

```suggestion
            int table_small, table_large;
            if (t->bucketExp[0] <= t->bucketExp[1]) {
```

src/hashset.c
rainsupreme: The iterator increments normally, but rehash increments the reverse index, so they aren't interchangeable. This is just the quick fix - we could convert iterators to use the same order as rehash and scan - then we'd be able to take shortcuts like this.

src/hashset.c
rainsupreme: That's true! I'll add a UT.

src/hashset.c
rainsupreme: > I still want to argue that it's safe to skip already rehashed indices in the old table though, because those have been rehashed = moved to the new table. (Sam can't be behind the rehash index in the old table, but he can be behind the (index corresponding to the) rehash index in the new table.)

Maybe I'm misunderstanding you, but I think I disagree. Sam can hash to a bucket behind the current rehash index while simultaneously being stored in a bucket ahead of the rehash index due to probe chaining. Suppose when Sam gets rehashed Sam gets placed directly in the bucket he hashes to in the new table, which is behind the rehash index? Remember that rehashing _does not_ follow probe chains and _can_ stop between where Sam is stored and the bucket Sam hashes to!

I've changed it to match the new table being twice as large as the old table:
```
      Sam (☺) hashes to this bucket in the old table
      |
old:  E E ☺ .
new:  .......
        |
        RehashIdx
        |
        cursor

       Sam (☺) hashes here in new table (one of the two new buckets that correspond to the old bucket he hashed to)
       |
old:  E E . .
new:  .☺.....
        |   |
        |   RehashIdx
        |
        cursor
```

src/hashset.c
rainsupreme: I took a look and changing the iteration order is actually easy, so I did that (which means we can skip the buckets now)

zuiderkwast: Oh, right, the iterator just does index++. I forgot that part. +_+

src/hashset.c
zuiderkwast: In your last picture, Sam's face is moved to the new table. In the new table we're not going to skip him. (Especially if we're not skipping anything in the new table.)

src/hashset.c
```diff
@@ -1291,50 +1291,50 @@ size_t hashsetScan(hashset *t, size_t cursor, hashsetScanFunction fn, void *priv
...
                 }
                 in_probe_sequence |= b->everfull;
 
                 /* Increment the reverse cursor not covered by the smaller mask.*/
-                cursor = nextCursor(cursor, mask1);
+                cursor = nextCursor(cursor, mask_large);
 
                 /* Continue while bits covered by mask difference is non-zero */
-            } while ((cursor & (mask0 ^ mask1)) && cursor != start_cursor);
+            } while (cursor & (mask_small ^ mask_large));
```
zuiderkwast: Now that we're just using one cursor again, we need the condition `&& cursor != start_cursor` back.

zuiderkwast (commented):
I want to merge the fix in the iterator and the tests.

We don't seem to agree we about what we can skip in scan. Can you revert those changes for now? (We can keep the renamed variables table0 to table_small, etc.)

Can you also update the PR title and description to match what's actually implemented?

src/hashset.c
rainsupreme: But why not? Sam is behind `cursor` now. Scan assumes it has already covered all buckets behind `cursor` including the bucket Sam moved to, and won't revisit them.

rainsupreme:
I've done the rebasing onto the new squashed hashset branch, and I've removed the hashsetScan() stuff we're still discussing. (I did keep a small simple optimization though.)

src/hashset.c
zuiderkwast: Sorry, I didn't reconnect to your pictures properly.

> But why not? Sam is behind `cursor` now.

OK, if Sam was rehashed to behind the cursor, then he is not covered in this scan call. Agreed.

But then he must have been covered in the previous scan call. In the old table, he was ahead of the cursor, but scan follows the probe chain and covered him. If he hashes to a bucket behind the cursor but he is located in a bucket ahead of the cursor, there must be a probe chain between those two buckets, because probing is what placed him there.

So, in this scan call, we cover him in the old table:

```
      Sam (☺) hashes to this bucket in the old table
      |
old:  E E ☺ .
new:  .......
        |
        RehashIdx
        |
        cursor
```

The only way we could miss him is that rehashing was called from the scan callback, but we have prevented that by pausing rehashing during the scan call.

zuiderkwast (approved):

PR #1151, opened 2024-10-11
Adding KEYINFO command to find out keys that have large number of elements

otheng03:
ISSUE LINK: https://github.com/valkey-io/valkey/issues/1827

- Add two new parameters to valkey.conf:
  - `keyinfo-num-elements-larger-than`
  - `keyinfo-large-num-elements-max-len`
- Implement three new commands:
  - `KEYINFO GET <count> MANY_ELEMENTS` 
  - `KEYINFO LEN MANY_ELEMENTS`
  - `KEYINFO RESET MANY_ELEMENTS`


src/keyinfo.c
```diff
@@ -0,0 +1,133 @@
+#include "keyinfo.h"
+
+void keyinfoFreeEntry(keyinfoEntry *entry) {
+    decrRefCount(entry->key);
+    entry->key = NULL;
+}
+
+/* Initialize the bigkey log. This function should be called a single time at server startup. */
+void keyinfoInit(void) {
```
otheng03, line 9: TODO : 
~~dynamically recreate the bucket when `keyinfo-large-num-elements-max-len` is changed~~
Make tests

src/keyinfo.h
```diff
@@ -0,0 +1,17 @@
+#ifndef __KEYINFO_H__
+#define __KEYINFO_H__
+
+#include "server.h"
+
+/* This structure defines an entry inside the bigkey log bucket */
+typedef struct keyinfoEntry {
+    long long id;
```
otheng03: TODO : id generation rule

hwware:
I have not begun to review your codes, I just take a look the new command names: 
KEYINFO GET <count> MANY_ELEMENTS
KEYINFO LEN MANY_ELEMENTS
KEYINFO RESET MANY_ELEMENTS

The MANY_ELEMENTS is a little bit weird, i am not sure if you are willing to attend Valkey meeting next Monday (Link here https://github.com/valkey-io/valkey/wiki/Weekly-meeting#weekly-meeting-schedule ) and introduce this PR background to Valkey maintainer?  Thanks

otheng03:
Hi, thank you for your message.
I’m not sure if I’ll be able to attend the meeting, as May 3rd to 6th is one of the biggest holiday in my country.
Also, since I’m not fluent in English, I’m a bit concerned about communicating effectively during the meeting.
Would it be possible for me to introduce the feature in a document instead?
If that’s not sufficient, I’d be happy to join the meeting on Monday, May 12, if that works for you.

hwware:
In today meeting, tsc member discussed this PR and express 2 concerns:
1. The new KEYINFO list could cost some memory due to the "key name" in the entry. Because in Valkey, the maximum allowed key size is 512 MB, thus if multiply such these kinds of key is saved to KEYINFO list,  it causes huge memory consumption.  For the slowlog, no key name is saved (https://valkey.io/commands/slowlog-get/)

2. MANY_ELEMENTS name is a little bit weird, I suggest we have a better name

Could you wrote down a specific example how this feature in used in production environment,  (I know the valkey-cli --bigkeys is too expensive in real production) so that TSC member can understand your idea or join next Monday meeting, Thanks

otheng03:
Thanks for the review.
1. Oh, I see. ~~I'll try to modify it to reference the existing key in order to reduce memory usage.~~ I realized that referencing the existing key in hashtabase makes code complex since when the hashtable is resized, we may need to verify whether it's safe to keep using the pointer. Also, having an informational feature affect the system’s core functionality could make maintenance more difficult, so I don't think it's a good approach.
Instead, could we consider adding an option to limit the key length in KEYINFO? (ex: `keyinfo-max-key-len`)

2. I've listed a few candidates. Could you please pick one?
  2.1. keyinfo get -1 bigkeys
  2.2. keyinfo get -1 HIGH_CARDINALITY
  2.3. info bigkeys
  2.4. info highcardinalitykeys
IMHO, `keyinfo get -1 HIGH_CARDINALITY` looks fine, but any suggestions are welcome.

otheng03:
| Could you wrote down a specific example how this feature in used in production environment,

The use case I have in mind is to identify keys with a large number of elements easily and early—before they cause problems in the system.
To achieve this, the monitoring tool periodically checks which keys are large.

I realized that using this feature to retrieve big keys can block other commands if a key is extremely large, such as one that is 512MB in size. It might be better to change the default behavior to return only the beginning part of the key name, and provide an option to return the full key name if needed.

otheng03:
This is a specific usecase I have in mind.
- Regularly checking for big keys and raising an alert when they are detected.
  - I want to save time getting dump files from remote servers and running `valkey-cli --bigkeys`.
  - Due to security restrictions, I sometimes can’t access a remote host, which makes it difficult to get dump.rdb
    - Sometimes, access to remote hosts is granted through a complex verification process.
  - Although rare, there are cases where retrieving the full dataset is restricted for security reasons.

{3C7043BE-333D-48BF-A6A1-F4628CB715D0}

otheng03:
Hi, @hwware.
Thank you, It was a great experience joining the meeting — I really liked how the open discussion moved quickly but still went deep. 
I didn’t get a chance to bring up the topic I had prepared. But I left them by comments in this PR and I needed some time to work on the feedback I got from the May 5 meeting anyway. 
If it's okay, I'd like to attend the meeting often.

hwware:
> Hi, @hwware. Thank you, It was a great experience joining the meeting — I really liked how the open discussion moved quickly but still went deep. I didn’t get a chance to bring up the topic I had prepared. But I left them by comments in this PR and I needed some time to work on the feedback I got from the May 5 meeting anyway. If it's okay, I'd like to attend the meeting often.

I add your PR in our meeting agenda (https://github.com/valkey-io/valkey/wiki/Weekly-meeting#core-team-sync-2025-05-19-1400-utc), you should have chance to discuss your requirement with TSC member in the meeting.

otheng03:
| I add your PR in our meeting agenda (https://github.com/valkey-io/valkey/wiki/Weekly-meeting#core-team-sync-2025-05-19-1400-utc), you should have chance to discuss your requirement with TSC member in the meeting.

Thanks for adding this PR to the agend.
However, I noticed that you've drawn a consensus between `INFO KEYSIZES` and this PR. [#1 of this comment](https://github.com/valkey-io/valkey/issues/1969#issuecomment-2851297560)
I'm a bit concerned that this might lead to redundant discussions.
Would it be better to discuss this further? Anyway, I'm not sure it's acceptable to develop two similar features.

otheng03:
Some alternative architectures came to mind. I'll write them up soon. It might take a few days or weeks. Sorry to keep you waiting.

madolson:
Thanks for investigating this implementation. We discussed it in our weekly meeting, and have some open questions. 
1. Performance: How much CPU does it take to perform the crc16 or some similar hash (xxhash or siphash). We probably can't enable this feature by default if the hashing is taking a lot of time.
2. Module interaction: The module API needs to also be included here. JSON should be able to be supported here as well as modules that edit hash objects.
3. This isn't really the 10 largest keys or the 10 most recently edited large keys. It's really just a bin of "N" large keys. There was an ask to either make it the 10 largest or the 10 most recently edited. That may not be possible though.
4. Do we really need this functionality. Can an end user solve all of their problems with either a `valkey-cli --bigkeys` or using second order information like slowlog or commandlong large-reply. 
5. The deletion code path can lead to dangling keys. If something is async deleted, it looks like it's not getting tracked. If we do a flushall that is asynchronous, we don't do any mainthread work to clean up the data so the item in the table will not get cleaned up.

otheng03:
Hi, I sincerely apologize for the delay in responding. I'm truly sorry for any inconvenience caused by the delayed reply.
At the moment, I need to focus on other areas, so I may not be able to response this issue right away.
However, I would love to revisit it when the opportunity arises.

otheng03:
Thanks for your review. It made me take a step back and rethink the design of this PR.

Based on that, I’d like to propose an updated design below.
I’m not claiming this is a perfect solution, but I believe it could be a more meaningful direction than continuing to iterate on the current approach.

# Proposal: Bucketized Cardinality Tracking with Lazy Evaluation

## Goal

Provide a low-overhead way to identify "large cardinality keys" without maintaining a full global ranking.

- O(1) write-path (=increase and decrease in cardinality)
- Approximate but useful Top-N results
- Avoid full keyspace scan
- Keep memory usage as low as possible

---

## Key Idea

Instead of maintaining an exact global Top-N:

- Track only a **bounded set of candidate keys**
- Group them into **coarse cardinality buckets**
- Update bucket positions **eagerly on writes**
- Refine results **lazily on reads**

---

## Data Structure

### Buckets (coarse cardinality ranges)

```
# Bucket No (cardinality ranges) -> list of pointer for keys
Bucket 1 (1024=Minimum cardinality for tracking) → [K5]
Bucket 2 (1025~2048) → [K4] [K6]
Bucket 3 (2049~4096) → [K2]
Bucket 4 (4097~9182) → [K3] [K7] [K8]
...
```
- Each bucket holds a linked list of keys
- Buckets represent **ranges**, not exact values
- If the top largest buckets already contain more than `M` items, insertions into lower buckets are ignored.

### Candidate Entry

| Property                  |
|---------------------------|
| pointer to the data entry |
| estimated_size            |

- Only candidate entry are tracked (bounded size)

## Write Path (O(1))

On mutation:

1. Compute new cardinality
2. Update `estimated_size`
3. Recompute `bucket_index`
4. Move entry if bucket changed

### Example

```
Before:
Bucket 3 → [K1(size=20)]

LPUSH K1 5 items

After:
Bucket 4 → [K1(size=25)]
```
This avoids stale over-estimation and keeps candidates fresh.

## Admission Policy

Not all keys become candidates.

Example:

- Only admit keys if `size >= threshold`
- If candidate set is full:
    - evict from lowest bucket

---

## Read Path (Lazy Evaluation)

### Step 1: Collect candidates

Scan from highest bucket:
Bucket 6 → Bucket 5 → Bucket 4 ...
Collect more than N

### Step 2: Optional refinement

Within collected candidates:

- Re-check cardinality (cheap metadata or partial recompute)
- Sort locally

---

### Step 3: Return Top-N

K1: ~2400
K9: ~1800
K3: ~1200

## Summary

This design:

- avoids maintaining an exact ranking
- keeps write-path minimal
- still provides useful Top-N insights

> **Bucketized estimation + lazy read-time refinement**

otheng03:
Thanks for your review. It made me take a step back and rethink the design of this PR.

Based on that, I’d like to propose an updated design below.
I’m not claiming this is a perfect solution, but I believe it could be a more meaningful direction than continuing to iterate on the current approach. 

There may be gaps in the design. Feedback is welcome.

# Proposal: Bucketized Cardinality Tracking with Lazy Evaluation

## Goal

Provide a low-overhead way to identify "large cardinality keys" without maintaining a full global ranking.

- O(1) write-path (=increase and decrease in cardinality)
- Approximate but useful Top-N results
- Avoid full keyspace scan
- Keep memory usage as low as possible

---

## Key Idea

Instead of maintaining an exact global Top-N:

- Track only a **bounded set of candidate keys**
- Group them into **coarse cardinality buckets**
- Update bucket positions **eagerly on writes**
- Refine results **lazily on reads**

---

## Data Structure

### Buckets (coarse cardinality ranges)

```
# Bucket No (cardinality ranges) -> list of pointer for keys
Bucket 1 (1024=Minimum cardinality for tracking) → [K5]
Bucket 2 (1025~2048) → [K4] [K6]
Bucket 3 (2049~4096) → [K2]
Bucket 4 (4097~9182) → [K3] [K7] [K8]
...
```
- Each bucket holds a linked list of keys
- Buckets represent **ranges**, not exact values
- If the top largest buckets already contain more than `M` items, insertions into lower buckets are ignored.
- Only candidate entry are tracked (bounded size)

### dictEntry
Add a pointer in dictEntry to reference its node in the bucket list.
This allows us to update a key’s cardinality without scanning the bucket.

## Write Path (O(1))

On mutation:

0. Record the current cardinality
1. Compute new cardinality
2. Update `estimated_size`
3. Recompute `bucket_index`
4. Move entry if bucket changed

### Example

```
Before:
[K1.bucketNode = node of Bucket 3]
Bucket 3 → [K1(size=20)]

LPUSH K1 5 items

After:
Remove K1 from Bucket 3 utlize [K1.bucketNode = node of Bucket 3]
Bucket 4 → [K1(size=25)]
[K1.bucketNode = node of Bucket 4]
```
This avoids stale over-estimation and keeps candidates fresh.

## Admission Policy

Not all keys become candidates.

Example:

- Only admit keys if `size >= threshold`
- If candidate set is full:
    - evict from lowest bucket

---

## Read Path (Lazy Evaluation)

### Step 1: Collect candidates

Scan from highest bucket:
Bucket 6 → Bucket 5 → Bucket 4 ...
Collect more than N

### Step 2: Optional refinement

Within collected candidates:

- Re-check cardinality (cheap metadata or partial recompute)
- Sort locally

---

### Step 3: Return Top-N

K1: ~2400
K9: ~1800
K3: ~1200

## Summary

This design:

> **Bucketized estimation + lazy read-time refinement**

- avoids maintaining an exact ranking
- keeps write-path minimal
- still provides useful Top-N insights
- Adds ~8 bytes per key
  - plus additional overhead proportional to the number of bucketed entries. (Up to M bucket entries, depending on the configuration.) 
- If the top N buckets grow too large, it can increase memory usage and degrade read-path performance.

Next step: I’m planning to prototype this approach and evaluate its performance.

otheng03:
> 1. Performance: How much CPU does it take to perform the crc16 or some similar hash (xxhash or siphash). We probably can't enable this feature by default if the hashing is taking a lot of time.

That’s a good point. It doesn’t seem like this will be used in the new design I proposed, but I can revisit it when it becomes necessary.

> 2. Module interaction: The module API needs to also be included here. JSON should be able to be supported here as well as modules that edit hash objects.

Thanks, got it. I’ll check the module as soon as possible.

> 3. This isn't really the 10 largest keys or the 10 most recently edited large keys. It's really just a bin of "N" large keys. There was an ask to either make it the 10 largest or the 10 most recently edited. That may not be possible though.
> 5. The deletion code path can lead to dangling keys. If something is async deleted, it looks like it's not getting tracked. If we do a flushall that is asynchronous, we don't do any mainthread work to clean up the data so the item in the table will not get cleaned up.

That’s true — the new design might be a better solution.

> 4. Do we really need this functionality. Can an end user solve all of their problems with either a valkey-cli --bigkeys or using second order information like slowlog or commandlong large-reply.
t seems to be a trade-off between reacting after an incident caused by big keys and proactively detecting them to improve service quality.

Using tools like valkey-cli can still help resolve issues after they occur, but we aim to provide a higher level of service by detecting them in advance.

PR #1159, opened 2024-10-13
Make global configs as static

pizhenwei:
Don't expose static configs symbol, and make configEnumGetValue as static function.


hpatro (approved):
LGTM

PR #901, opened 2024-08-12
Add support for setting the group on a unix domain socket

ayush933:
Add new optional, immutable string config called `unixsocketgroup`. 
Change the group of the unix socket to `unixsocketgroup` after `bind()` if specified.

Manually verified that group is modified as expected and fails on illegal arguments (like non existent group name).

Fixes #873.

PR #1160, opened 2024-10-13
Introduce connection context for Unix socket

pizhenwei:
Hide 'unixsocketgroup' and 'unixsocketperm' into a Unix socket specific data structure. A single opaque pointer 'void *priv' is enough for a listener. Once any new config is added, we don't need 'void *priv2', 'void *priv3' and so on.

Fixes: b48596a914 ('Add support for setting the group on a unix domain socket (#901)')
Cc: Ayush Sharma <mrayushs933@gmail.com>


enjoy-binbin (commented):
This was a decision made at the time (for simplicity), https://github.com/valkey-io/valkey/pull/901#pullrequestreview-2237090821

i am ok with both, i will let @madolson do the call.

madolson (approved):
This seems reasonable to me. I didn't think of this solution at the time.

PR #1163, opened 2024-10-14
Use listLast to replace listIndex -1

enjoy-binbin:
Minor cleanup, listLast do the same thing and is widely used
and easier to understand (less code).


zuiderkwast (approved):
Very minor change. :)

PR #1166, opened 2024-10-14
Minor cleanups in acl-v2 tests

enjoy-binbin:
1. Make sure to assert the ERR prefix.
2. Match "Syntax error*" in case of the message change.

### DIFF
--- a/tests/unit/acl-v2.tcl
+++ b/tests/unit/acl-v2.tcl
@@ -42,19 +42,19 @@ start_server {tags {"acl external:skip"}} {
 
     test {Test selector syntax error reports the error in the selector context} {
         catch {r ACL SETUSER selector-syntax on (this-is-invalid)} e
-        assert_match "*ERR Error in ACL SETUSER modifier '(*)*Syntax*" $e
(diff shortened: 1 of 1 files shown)


tests/unit/acl-v2.tcl
```diff
@@ -75,11 +75,11 @@ start_server {tags {"acl external:skip"}} {
 
         # Test invalid selector syntax
         catch {r ACL SETUSER invalid-selector " () "} err
-        assert_match "*ERR*Syntax error*" $err
+        assert_match "ERR*Syntax error*" $err
         catch {r ACL SETUSER invalid-selector (} err
-        assert_match "*Unmatched parenthesis*" $err
+        assert_match "ERR*Unmatched parenthesis*" $err
         catch {r ACL SETUSER invalid-selector )} err
-        assert_match "*ERR*Syntax error" $err
```
enjoy-binbin, line 82: What I actually want to change is this line:
1. this line work as the same as line 78 but without the `*`
2. without the `*`, test fails when extending message to expose more info (internal fork)

zuiderkwast (approved):

PR #1169, opened 2024-10-14
Rename z{malloc,calloc,realloc,free} into valkey_{malloc,calloc,realloc,free}

### LINKED ISSUE #1157: [NEW] Rename zmalloc (and all z-alloc functions) valkey_zmalloc

This is a re-opening of an issue we raised on redis years ago. It was https://github.com/redis/redis/pull/9986

I tried the latest redis server and it still suffer the same issue with redis 7.4.1. I didn't try valkey, but given that the issue was there years ago, and that the valkey fork is rather recent, you shall suffer the same issue too. As soon as you try to create a binary with a static openssl and a static zlib, you end up with some symbol collision, at least for zcalloc.

A suggested back then, it would be cool of these z-alloc function would be namespaced. In the valkey project, this would typically be named valkey_zmalloc.

Currently when we build redis (it would be the same with valkey), we patch redis sources with an ugly `#define zcalloc redis_zcalloc` and it works. However it's no more than ugly workaround, for me this shall be fixed directly in the source code via a massive renaming. The redis team told us back then they are against massive renaming, but given you have massively replaced redis by valkey, I doubt you have the same ideological barriers ;)

Romain-Geissler-1A:
The zcalloc symbol is a symbol name already used by zlib, which is defining other names using the "z" prefix specific to zlib. In practice, linking valkey with a static openssl, which itself might depend on a static libz will result in link time error rejecting multiple symbol definitions.

Fixes: #1157

### DIFF
--- a/src/zmalloc.h
+++ b/src/zmalloc.h
@@ -107,6 +107,15 @@
 #define HAVE_DEFRAG
 #endif
 
+/* The zcalloc symbol is a symbol name already used by zlib, which is defining
+ * other names using the "z" prefix specific to zlib. In practice, linking
+ * valkey with a static openssl, which itself might depend on a static libz
+ * will result in link time error rejecting multiple symbol definitions. */
+#define zmalloc valkey_malloc
+#define zcalloc valkey_calloc
+#define zrealloc valkey_realloc
+#define zfree valkey_free
+
 /* 'noinline' attribute is intended to prevent the `-Wstringop-overread` warning
  * when using gcc-12 later with LTO enabled. It may be removed once the
  * bug[https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96503] is fixed. */
--- a/deps/hdr_histogram/hdr_redis_malloc.h
+++ b/deps/hdr_histogram/hdr_redis_malloc.h
@@ -1,13 +1,13 @@
 #ifndef HDR_MALLOC_H__
 #define HDR_MALLOC_H__
 
-void *zmalloc(size_t size);
+void *valkey_malloc(size_t size);
 void *zcalloc_num(size_t num, size_t size);
-void *zrealloc(void *ptr, size_t size);
-void zfree(void *ptr);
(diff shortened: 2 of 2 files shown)


Romain-Geissler-1A:
Updated version which I hope will work better (usually I do expect CI to start automatically in project i contribute to, so that I can quickly see if it works or not).

src/zmalloc.h
```diff
@@ -107,6 +107,21 @@
 #define HAVE_DEFRAG
 #endif
 
+/* The zcalloc symbol is a symbol name already used by zlib, which is defining
+ * other names using the "z" prefix specific to zlib. Valkey has historically
```
zuiderkwast: This big comment is a little excessive.

Whether valkey should prefix all functions with `valkey_` is debatable. Valkey is not a lib. It's not statically linked to arbitrary 3rd party code.

What we historically did and what we ideally want in the long term don't belong in code comments IMO. Such discussions are done in the issues and PRs.

I think we can keep the first sentence about zlib and the explanation about linking with openssl that may have been statically linked with zlib is a problem.

zuiderkwast (commented):
Code looks good to me. Just commenting on the comment.

zuiderkwast:
CI starts automatically except for first-time contributors. I don't think we can change that.

src/zmalloc.h
Romain-Geissler-1A: Ok. The code comment, commit message, and pull request message has been shortened accordingly with the latest update.

zuiderkwast (approved):
Thanks!

src/zmalloc.h
zuiderkwast: Perfect.

zuiderkwast:
@madolson you commented on the original redis issue. Do you have any comments on this before I merge it?

madolson (approved):
No, I don't have any concerns. I see that I was against this at the time, but the define seems like a rather clean way to handle it.

Issue #1069, opened 2024-09-24
Investigate integrating with fast_float


madolson:
I've seen a couple of posts recently about how both dragonflyDB and Redis integrated with https://github.com/fastfloat/fast_float to improve parsing of doubles. We use strtod (exposed as getDoubleFromObject) in sorted set computation. The library was implemented in C++, but we should be able to extract out the needed code that solves our specific use case.

swaingotnochill:
Hi, I would like to work on this one. Additionally, this will be my first time contributing to Valkey so it would be great if someone could review my work alongside.

PingXie:
go for it, @swaingotnochill. You can tag us in your PR when it is ready for review.

swaingotnochill:
Thanks. I will start working on it and add you guys when I have a draft PR ready.

swaingotnochill:
@madolson  @PingXie  I looked at the fast_float, they have scripts to get it in a single header file to use. Then it's possible to create an interface to expose specific functions just like other dependencies in "deps" folder. And looking at Redis code, they have done something similar. Is the plan to replace the `strtod` in the sorted_set implementation Or all of the occurrences of strtod? (From the benchmarks of fast_float, its 5 times faster than strtod for parsing random floating numbers)

parthpatel:
How would you expose an interface for a function with templating? Are you proposing simply initializing a version of that function for a type and exposing an interface for it? 

```
template <typename T, typename UC = char,
          typename = FASTFLOAT_ENABLE_IF(is_supported_float_type<T>())>
FASTFLOAT_CONSTEXPR20 from_chars_result_t<UC>
from_chars(UC const *first, UC const *last, T &value,
           chars_format fmt = chars_format::general) noexcept;
```

swaingotnochill:
> How would you expose an interface for a function with templating? Are you proposing simply initializing a version of that function for a type and exposing an interface for it?
> 
> ```
> template <typename T, typename UC = char,
>           typename = FASTFLOAT_ENABLE_IF(is_supported_float_type<T>())>
> FASTFLOAT_CONSTEXPR20 from_chars_result_t<UC>
> from_chars(UC const *first, UC const *last, T &value,
>            chars_format fmt = chars_format::general) noexcept;
> ```

Something like this: https://stackoverflow.com/questions/2744181/how-to-call-c-function-from-c (look at the top answer).
Basically, create a wrapper over fast_float to use it in C.

madolson:
[...]

PR #1170, opened 2024-10-15
feat: Integration with fast_float

swaingotnochill:
See #1069 

This is a draft pull request and the work is in progress.

The goal of this is to get timely feedback and reviews from the reviewers.

TODO:
- [x] Create the wrapper and archive it.
- [] Replace it with strtod.
- [] Benchmark it.
- [] Run it on different platforms.

cc: @madolson @PingXie

### DIFF
--- a/deps/fast_float/fast_float.h
+++ b/deps/fast_float/fast_float.h
@@ -1,0 +1,3869 @@
+// fast_float by Daniel Lemire
+// fast_float by João Paulo Magalhaes
+//
+//
+// with contributions from Eugene Golushkov
+// with contributions from Maksim Kita
+// with contributions from Marcin Wojdyr
+// with contributions from Neal Richardson
+// with contributions from Tim Paine
+// with contributions from Fabio Pellacini
+// with contributions from Lénárd Szolnoki
+// with contributions from Jan Pharago
+// with contributions from Maya Warrier
+// with contributions from Taha Khokhar
+//
+//
+// MIT License Notice
+//
+//    MIT License
+//    
+//    Copyright (c) 2021 The fast_float authors
+//    
+//    Permission is hereby granted, free of charge, to any
+//    person obtaining a copy of this software and associated
+//    documentation files (the "Software"), to deal in the
+//    Software without restriction, including without
+//    limitation the rights to use, copy, modify, merge,
+//    publish, distribute, sublicense, and/or sell copies of
+//    the Software, and to permit persons to whom the Software
+//    is furnished to do so, subject to the following
+//    conditions:
+//    
+//    The above copyright notice and this permission notice
+//    shall be included in all copies or substantial portions
+//    of the Software.
+//    
+//    THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF
+//    ANY KIND, EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED
+//    TO THE WARRANTIES OF MERCHANTABILITY, FITNESS FOR A
+//    PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT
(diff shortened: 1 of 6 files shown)


madolson:
@parthpatel FYI.

If possible we would rather not take the C++ dependency, although that might be doable. Do you know how much effort it would be to port the code into C. The templating code seems very generic, but I don't think we need all the flexibility it is providing.

swaingotnochill:
@madolson do you suggest to port the cpp code to c rather than use fast_float?

deps/fast_float/fast_float.h
```diff
@@ -0,0 +1,3869 @@
+// fast_float by Daniel Lemire
```
parthpatel, line 1: if you embed this file, and in future there are bugfixes to this library, how do we track it? is it possible to use submodules for this https://git-scm.com/book/en/v2/Git-Tools-Submodules ?

deps/fast_float/fast_float_strtod.cpp
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+extern "C"
+{
+    double fast_float_strtod(const char *nptr, char **endptr)
+    {
+        double result;
+        auto answer = fast_float::from_chars(nptr, nptr + strlen(nptr), result);
+
+        if (answer.ec == std::errc())
```
parthpatel, line 33: what is this check doing? shouldn't you just check for null?

deps/fast_float/fast_float_strtod.cpp
```diff
@@ -0,0 +1,51 @@
...
+        {
+            if (endptr)
```
parthpatel, line 35: you can move this outside of if block, it is common.

deps/fast_float/fast_float_strtod.cpp
```diff
@@ -0,0 +1,51 @@
...
+ *
+ * @note This function uses the fast_float library (https://github.com/fastfloat/fast_float)
+ *       for the actual conversion, which can be significantly faster than standard library functions.
+ *
+ * @see https://github.com/fastfloat/fast_float for more information on the underlying library.
+ */
+
+extern "C"
+{
+    double fast_float_strtod(const char *nptr, char **endptr)
```
parthpatel, line 28: @madolson do we always use double type in Redis? Would we never want to use float?

parthpatel:
> @parthpatel FYI.
> 
> If possible we would rather not take the C++ dependency, although that might be doable. Do you know how much effort it would be to port the code into C. The templating code seems very generic, but I don't think we need all the flexibility it is providing.

Even if we remove the templating for character type, we still have "class" and "std::" library references. With little more effort, we should be able to remove it. But we won't be able to merge fixes or changes from upstream. In my opinion, we should start with fast_float library as a sub-module (instead of copying the code) and if we run into C++ related issues, then we can still internally change the implementation.

deps/fast_float/fast_float.h
swaingotnochill: The proper scenario is to update the header file with the bug fix version. The diff will be large, but we will be kind of updating the third party dep. However, if we plan to make bug fix, it's not a good idea. Best would be to create a fork and make sure its parallel to the upstream, and we can make custom fixes as well. But it then makes additional overhead.

deps/fast_float/fast_float_strtod.cpp
swaingotnochill: fast_float is written in a way where it returns error and we have to handle them explicitly. That's present in their docs.

parthpatel:
@swaingotnochill I couldn't push changes to your PR, so I created what I was proposing as a draft PR here https://github.com/valkey-io/valkey/pull/1260

zuiderkwast:
Replaced by #1260

PR #1171, opened 2024-10-15
Fix FUNCTION KILL error message being displayed as SCRIPT KILL

enjoy-binbin:
The client that was killed by FUNCTION KILL received a reply of
SCRIPT KILL and the server log also showed SCRIPT KILL.

### DIFF
--- a/src/script_lua.c
+++ b/src/script_lua.c
@@ -1609,7 +1609,13 @@ static void luaMaskCountHook(lua_State *lua, lua_Debug *ar) {
     scriptRunCtx *rctx = luaGetFromRegistry(lua, REGISTRY_RUN_CTX_NAME);
     serverAssert(rctx); /* Only supported inside script invocation */
     if (scriptInterrupt(rctx) == SCRIPT_KILL) {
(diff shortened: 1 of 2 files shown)


src/script_lua.c
```diff
@@ -1609,7 +1609,13 @@ static void luaMaskCountHook(lua_State *lua, lua_Debug *ar) {
...
     serverAssert(rctx); /* Only supported inside script invocation */
     if (scriptInterrupt(rctx) == SCRIPT_KILL) {
-        serverLog(LL_NOTICE, "Lua script killed by user with SCRIPT KILL.");
+        char *err = NULL;
+        if (rctx->flags & SCRIPT_EVAL_MODE) {
+            err = "Script killed by user with SCRIPT KILL.";
+        } else {
+            err = "Script killed by user with FUNCTION KILL.";
+        }
+        serverLog(LL_NOTICE, "%s", err);
```
zuiderkwast, line 1618: ~~`err` = error = -ERR reply to clients.~~

~~Better use another word for log messages. `msg` or `notice`?~~

I see now it is also used below for the error reply. Sorry.

zuiderkwast (approved):

PR #1173, opened 2024-10-15
Add scoped RDB loading context and immediate abort flag

### LINKED ISSUE #1152: [Test Failure] Engine crash during TLS test with dual channel replication
https://github.com/valkey-io/valkey/actions/runs/11283922852/job/31417233387#step:6:6008

It seems to occur on the replica side. 

```
43813 valkey-server *
/lib/x86_64-linux-gnu/libc.so.6(+0x42520)[0x7fd198042520]
/lib/x86_64-linux-gnu/libssl.so.3(+0x32aca)[0x7fd1989dfaca]
/lib/x86_64-linux-gnu/libssl.so.3(SSL_read+0x27)[0x7fd1989dfcb7]
src/valkey-server 127.0.0.1:23382(+0x1fa987)[0x5566f92b0987]
src/valkey-server 127.0.0.1:23382(+0x1705f8)[0x5566f92265f8]
src/valkey-server 127.0.0.1:23382(rdbLoadRioWithLoadingCtx+0x2f5)[0x5566f91b54f5]
src/valkey-server 127.0.0.1:23382(readSyncBulkPayload+0x315)[0x5566f91a6075]
src/valkey-server 127.0.0.1:23382(+0x1f310d)[0x5566f92a910d]
src/valkey-server 127.0.0.1:23382(+0x1fb9a3)[0x5566f92b19a3]
src/valkey-server 127.0.0.1:23382(+0x1fbc34)[0x5566f92b1c34]
src/valkey-server 127.0.0.1:23382(beforeSleep+0x87)[0x5566f9152d17]
src/valkey-server 127.0.0.1:23382(aeMain+0x3e)[0x5566f914a73e]
src/valkey-server 127.0.0.1:23382(main+0x52f)[0x5566f913f7af]
/lib/x86_64-linux-gnu/libc.so.6(+0x29d90)[0x7fd198029d90]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x80)[0x7fd198029e40]
src/valkey-server 127.0.0.1:23382(_start+0x25)[0x5566f913fea5]
```

naglera:
This PR introduces a new mechanism for temporarily changing the
server's loading_rio context during RDB loading operations. The new
`RDB_SCOPED_LOADING_RIO` macro allows for a scoped change of the
`server.loading_rio` value, ensuring that it's automatically restored
to its original value when the scope ends.

Introduces a dedicated flag to `rio` to signal immediate abort, preventing 
potential use-after-free scenarios during replication disconnection in 
dual-channel load. This ensures proper termination of `rdbLoadRioWithLoadingCtx`
when replication is cancelled due to connection loss on main connection.

Fixes https://github.com/valkey-io/valkey/issues/1152


ranshid:
General comment:
Although I agree this fix will work and at first glance I see no issue with it, I would like to suggest maybe to tackle the problem from a more holistic POV:

Basically we would like to have a method to tell the current load process to stop ASAP. This can also be achieved by adding an RIO flag (eg #define RIO_FLAG_STOP_ASAP (1 << 2)) and have rio check for this flag when it is performing different io operations. the only issue is that the rdb RIO is local to the rdbLoadRioWithCtx. we can, however keep a pointer in the server to the current active loading rio so that in any point during load we can set the flag RIO_FLAG_STOP_ASAP on the current loading rio. IMO this would be cleaner.

src/server.h
```diff
@@ -2038,6 +2038,7 @@ struct valkeyServer {
         long long reploff;
         long long read_reploff;
         int dbid;
+        uint64_t close_asap : 1;
```
xbasel: Can we piggyback on the existing state variables to detect when the sync has been aborted / primary connection dropped? Since `cancelReplicationHandshake` is called when the connection is dropped and updates:

`server.repl_rdb_channel_state = REPL_DUAL_CHANNEL_STATE_NONE;`

Can't we simply use `server.repl_rdb_channel_state` in `rdbLoadProgressCallback`?

xbasel: This probably won't work  and will create an issue when RDB dual channel isn't used.

src/rdb.c
```diff
@@ -2933,6 +2933,11 @@ void rdbLoadProgressCallback(rio *r, const void *buf, size_t len) {
     if (server.repl_state == REPL_STATE_TRANSFER && rioCheckType(r) == RIO_TYPE_CONN) {
         server.stat_net_repl_input_bytes += len;
     }
+    if (server.repl_provisional_primary.close_asap == 1) {
+        serverLog(LL_WARNING, "Primary main connection dropped during RDB load callback");
+        return -1;
+    }
+    return 0;
```
madolson: I guess I'm not following why we can't null out the connections here and use that instead of a new close_asap flag.

src/rio.h
```diff
@@ -57,8 +57,10 @@ struct _rio {
      * all the data that was read or written so far. The method should be
      * designed so that can be called with the current checksum, and the buf
      * and len fields pointing to the new block of data to add to the checksum
-     * computation. */
-    void (*update_cksum)(struct _rio *, const void *buf, size_t len);
+     * computation.
```
madolson:
```suggestion
     * computation.
```

src/rdb.c
ranshid: It would still require to add logic to the rioConnRead/Write right? We could flag the rdb with RIO_FLAG_READ_ERROR.

naglera: Null out the fields will look the same as running sync with dual channel disabled.
> It would still require to add logic to the rioConnRead/Write right?

right

src/server.h
naglera: `server.repl_rdb_channel_state` will be equal to `REPL_DUAL_CHANNEL_STATE_NONE` also when dual channel is disabled

src/rdb.c
naglera: > We could flag the rdb with RIO_FLAG_READ_ERROR

While using RIO flags could potentially offer a cleaner solution, it does present its own challenges. As you mentioned, we would need to maintain a pointer in the server to the current active loading RIO. This would require adding and maintaining a `currently_loading_rdb` field in the server struct.  
This approach, while potentially more flexible for future use cases, introduces additional complexity and state management. It would require changes in multiple parts of the RDB load process to properly set, use, and clear this pointer.  
The current proposed solution, while more specific to this use case, has the advantage of being more localized and doesn't require global state management.

madolson: I suppose a third option is to remove the event handler for the replication connection before calling process events and then reinstalling it after the fact?

ranshid: @madolson I am not sure I understand this proposal. We need to process the events while we are loading in order to keep feeding the local replication buffer AFAIK. We could (as a third option) do nothing when we identify the replication link was broken and complete the load (or let it disconnect as well), however I do feel that having the ability to bail out from a load is something we might find handy in the future.

madolson: This idea made sense to me when I posted it but reading back it doesn't make sense, I might have just been missing  something. More generally I want to move away from doing recursive calls for handling processing events, and it that world we can just skip it, but that is likely a much larger change than what we want to do here.

src/replication.c
```diff
@@ -2833,6 +2833,7 @@ int readIntoReplDataBlock(connection *conn, replDataBufBlock *data_block, size_t
     }
     if (nread == 0) {
         serverLog(LL_VERBOSE, "Provisional primary closed connection");
+        if (server.loading_rio) server.loading_rio->flags |= RIO_FLAG_CLOSE_ASAP;
```
ranshid: nit: maybe add a short comment explaining why we are marking it here?

src/rio.h
```diff
@@ -168,8 +169,13 @@ static inline int rioGetWriteError(rio *r) {
     return (r->flags & RIO_FLAG_WRITE_ERROR) != 0;
 }
 
+/* Like rioGetReadError() but for async close errors. */
+static inline int rioGetAsyncCloseError(rio *r) {
```
ranshid, line 177: this is not used anywhere. suggest to drop it for now.

src/replication.c
ranshid: maybe for the sake of encapsulation we can have a function to encapsulate the flagging logic? like rioCloseASAP(rio *rdb)

src/rdb.h
```diff
@@ -139,6 +139,12 @@
 #define RDB_LOAD_ERR_EMPTY_KEY 1 /* Error of empty key */
 #define RDB_LOAD_ERR_OTHER 2     /* Any other errors */
 
+
+/* Macro to temporarily set server.loading_rio within a scope. */
+#define RDB_SCOPED_LOADING_RIO(new_rio)                                                                        \
+    __attribute__((cleanup(_restore_loading_rio))) rio *_old_rio __attribute__((unused)) = server.loading_rio; \
```
ranshid: very nice IMO. only thing is that I do not recall our compiler support scope in Valkey (@zuiderkwast DYK?)... for example is MSVC supported? I guess if we do  we can just place a cleanup logic on the only 2 places where we return from the function?

src/server.h
```diff
@@ -2058,6 +2058,7 @@ struct valkeyServer {
         int dbid;
     } repl_provisional_primary;
     client *cached_primary;             /* Cached primary to be reused for PSYNC. */
+    rio *loading_rio;                   /* Pointer to the Rio object currently used for loading data. */
```
ranshid:
```suggestion
    rio *loading_rio;                   /* Pointer to the rio object currently used for loading data. */
```

tests/integration/dual-channel-replication.tcl
```diff
@@ -1197,3 +1197,67 @@ start_server {tags {"dual-channel-replication external:skip"}} {
...
+                [string match {*replicas_waiting_psync:0*} [$primary info replication]]
+            } else {
+                fail "Primary did not free repl buf block after sync failure"
+            }
+            wait_for_log_messages 0 {"*Failed trying to load the PRIMARY synchronization DB from socket*"} $loglines 1000 10
+            # Replica should retry
+            wait_for_condition 500 1000 {
+                [string match "*slave*,state=wait_bgsave*,type=rdb-channel*" [$primary info replication]] &&
+                [string match "*slave*,state=bg_transfer*,type=main-channel*" [$primary info replication]] &&
+                [s -1 rdb_bgsave_in_progress] eq 1
```
ranshid: maybe we should wait to see that the sync was successful?

ranshid (commented):
Thank you @naglera looks promising! I like scoped actions, but I only want to make sure about the compiler support is not compromised. 
BTW if it is not, we can consider having a generic ScopeGuard macro in Valkey

src/replication.c
ranshid: also see my comment later about encapsulate the flagging in a dedicated function

src/rio.h
naglera: `rioGetWriteError` is not used anywhere either. This would maintain symmetry in our error handling capabilities for read, write, and async close operations, even if they're not currently utilized.

src/rdb.h
naglera: You raise a valid point about compiler support. Regarding the return points, currently `rdbLoadWithLoadingCtx` has about 5 return points, and this number may increase in the future as the function evolves. If we don't enforce a single return path, and not use scope based variable, we risk introducing bugs in the future where the cleanup logic might be missed on new return paths.

tests/integration/dual-channel-replication.tcl
naglera: We use `rdb-key-save-delay` in this test, which intentionally slows down the RDB saving process. Due to potential context switches, the sync time can be unpredictable and might take longer than expected. This unpredictability could lead to test flakyness.

src/replication.c
```diff
@@ -2833,6 +2833,8 @@ int readIntoReplDataBlock(connection *conn, replDataBufBlock *data_block, size_t
     }
     if (nread == 0) {
         serverLog(LL_VERBOSE, "Provisional primary closed connection");
+        /* Signal ongoing RDB load to terminate gracefully */
+        if (server.loading_rio) rioCloseASAP(server.loading_rio);
```
xbasel: Shouldn't this be invoked on line 2832 as well?

naglera: right, incase of connection state changes

tests/integration/dual-channel-replication.tcl
ranshid: we could reduce it to 0 and wait

ranshid (approved):
I approve in order to indicate this generally looks fine to me. We do need to decide on the cleanup attribute use which I think is mostly supported with some exceptions. (At least we would probably get compilation error IMO)

tests/integration/dual-channel-replication.tcl
```diff
@@ -1197,3 +1197,69 @@ start_server {tags {"dual-channel-replication external:skip"}} {
...
+    set primary_host [srv 0 host]
+    set primary_port [srv 0 port]
+    set loglines [count_log_lines 0]
+
+    $primary config set repl-diskless-sync yes
+    $primary config set dual-channel-replication-enabled yes
+    $primary config set loglevel debug
+    $primary config set repl-diskless-sync-delay 5; # allow catch failed sync before retry
+
+    # Generating RDB will cost 100 sec to generate
```
madolson:
```suggestion
    # Generating RDB will take 100 sec to generate
```
Mostly just making a change so I can trigger the tests

src/rdb.h
madolson: The other way to handle this is just to split the function, so we move the rest of the logic into another function that we call and can make sure it's correctly restored on return.

madolson: My ask would be that we split the discussion, let's do it manually for now and then open a separate issue about supporting this? I think generally we should have a general macro for this as well, not just one specific for the server.loading_rio.

tests/integration/dual-channel-replication.tcl
```diff
@@ -1197,3 +1197,69 @@ start_server {tags {"dual-channel-replication external:skip"}} {
...
+
+start_server {tags {"dual-channel-replication external:skip"}} {
+    set primary [srv 0 client]
+    set primary_host [srv 0 host]
+    set primary_port [srv 0 port]
+    set loglines [count_log_lines 0]
+
+    $primary config set repl-diskless-sync yes
+    $primary config set dual-channel-replication-enabled yes
+    $primary config set loglevel debug
```
madolson: Does this need to be set to debug?

tests/integration/dual-channel-replication.tcl
```diff
@@ -1197,3 +1197,69 @@ start_server {tags {"dual-channel-replication external:skip"}} {
         stop_write_load $load_handle
     }
 }
+
+
+start_server {tags {"dual-channel-replication external:skip"}} {
+    set primary [srv 0 client]
+    set primary_host [srv 0 host]
+    set primary_port [srv 0 port]
+    set loglines [count_log_lines 0]
```
madolson: You set this value later, so it has no impact here.

madolson (commented):
Mostly looks good to me.

tests/integration/dual-channel-replication.tcl
naglera: It helped while writing the test. We don't need it anymore

tests/integration/dual-channel-replication.tcl
naglera: I use it between the two sets for 
```
wait_for_log_messages 0 {"*Loading RDB produced by Valkey version*"} $loglines 1000 10
```

madolson: Yes, but you set it again on line 1228. You actually set it three times.

src/server.h
```diff
@@ -2058,6 +2058,7 @@ struct valkeyServer {
         int dbid;
     } repl_provisional_primary;
     client *cached_primary;             /* Cached primary to be reused for PSYNC. */
+    rio *loading_rio;                   /* Pointer to the rio object currently used for loading data. */
```
ranshid, line 2080: Shouldn't we make sure to first nullify it in initServerConfig?

naglera: right

ranshid:
@madolson I reviewed and approved. However since you were also reviewing and had some comments would wait for you to approve as well before we merge.

ranshid:
@naglera we need to rebase and resolve the conflicts

tests/integration/dual-channel-replication.tcl
```diff
@@ -1243,3 +1243,59 @@ start_server {tags {"dual-channel-replication external:skip"}} {
...
+    set primary_host [srv 0 host]
+    set primary_port [srv 0 port]
+
+    $primary config set repl-diskless-sync yes
+    $primary config set dual-channel-replication-enabled yes
+    $primary config set repl-diskless-sync-delay 5; # allow catch failed sync before retry
+
+    # Generating RDB will take 100 sec to generate
+    $primary debug populate 1000000 primary 1
+    $primary config set rdb-key-save-delay -10
```
madolson, line 1259: Does this test need to take 100+ seconds? The dual channel tests are some of the worst offenders today for taking a lot of time to complete.

ranshid: @naglera can you please check if we can reduce that?

naglera: To clarify, while the comment suggests it could take 100 seconds to generate the RDB, the actual test duration is likely much shorter. The master abort the RDB generation process when the replica's main connection is killed, rather than waiting for the full RDB generation to complete.
In theory, we could set this to an even larger number (e.g., 1,000,000 seconds) without significantly impacting the test duration, as the process should be interrupted well before completion.

If you're seeing that the dual channel tests are taking an unusually long time, it would be very helpful if you could share some logs or specific timing information.

madolson (approved):

ranshid:
Daily test failures do not seem related to this feature

PR #1176, opened 2024-10-16
Replace dict with new hashtable for sets datatype

### LINKED ISSUE #1094: Use new hashtable for sets

rainsupreme:
The new `hashtable` provides faster lookups and uses less memory than `dict`.

A TCL test case "SRANDMEMBER with a dict containing long chain" is deleted because it's covered by a hashtable unit test "test_random_entry_with_long_chain", which is already present.

This change also moves some logic from dismissMemory (object.c) to zmadvise_dontneed (zmalloc.c), so the hashtable implementation which needs the dismiss functionality doesn't need to depend on object.c and server.h.

This PR follows #1186.

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -978,7 +978,7 @@ void keysScanCallback(void *privdata, void *entry) {
 
 /* This callback is used by scanGenericCommand in order to collect elements
  * returned by the dictionary iterator into a list. */
-void scanCallback(void *privdata, const dictEntry *de) {
+void dictScanCallback(void *privdata, const dictEntry *de) {
     scanData *data = (scanData *)privdata;
     list *keys = data->keys;
     robj *o = data->o;
@@ -998,9 +998,7 @@ void scanCallback(void *privdata, const dictEntry *de) {
         }
     }
 
-    if (o->type == OBJ_SET) {
-        key = keysds;
-    } else if (o->type == OBJ_HASH) {
+    if (o->type == OBJ_HASH) {
         key = keysds;
         if (!data->only_keys) {
             val = dictGetVal(de);
@@ -1013,11 +1011,31 @@ void scanCallback(void *privdata, const dictEntry *de) {
             val = sdsnewlen(buf, len);
         }
     } else {
-        serverPanic("Type not handled in SCAN callback.");
+        serverPanic("Type not handled in dict SCAN callback.");
     }
 
     listAddNodeTail(keys, key);
     if (val) listAddNodeTail(keys, val);
+}
+
+void hashtableScanCallback(void *privdata, void *entry) {
+    scanData *data = (scanData *)privdata;
+    robj *o = data->o;
+    list *keys = data->keys;
+    data->sampled++;
+
+    /* currently only implemented for SET scan */
+    serverAssert(o && o->type == OBJ_SET && o->encoding == OBJ_ENCODING_HASHTABLE);
+    sds key = (sds)entry; /* Specific for OBJ_SET */
+
+    /* Filter element if it does not match the pattern. */
+    if (data->pattern) {
+        if (!stringmatchlen(data->pattern, sdslen(data->pattern), key, sdslen(key), 0)) {
+            return;
+        }
+    }
+
+    listAddNodeTail(keys, key);
 }
 
 /* Try to parse a SCAN cursor stored at object 'o':
@@ -1083,7 +1101,6 @@ void scanGenericCommand(client *c, robj *o, unsigned long long cursor) {
     sds typename = NULL;
     long long type = LLONG_MAX;
     int patlen = 0, use_pattern = 0, only_keys = 0;
-    dict *ht;
 
     /* Object must be NULL (to iterate keys names), or the type of the object
      * must be Set, Sorted Set, or Hash. */
@@ -1152,34 +1169,35 @@ void scanGenericCommand(client *c, robj *o, unsigned long long cursor) {
      * just return everything inside the object in a single call, setting the
      * cursor to zero to signal the end of the iteration. */
 
-    /* Handle the case of a hash table. */
-    ht = NULL;
+    /* Handle the case of kvstore, dict or hashtable. */
+    dict *dict_table = NULL;
+    hashtable *hashtable_table = NULL;
+    int shallow_copied_list_items = 0;
     if (o == NULL) {
-        ht = NULL;
-    } else if (o->type == OBJ_SET && o->encoding == OBJ_ENCODING_HT) {
-        ht = o->ptr;
+        shallow_copied_list_items = 1;
+    } else if (o->type == OBJ_SET && o->encoding == OBJ_ENCODING_HASHTABLE) {
+        hashtable_table = o->ptr;
+        shallow_copied_list_items = 1;
     } else if (o->type == OBJ_HASH && o->encoding == OBJ_ENCODING_HT) {
-        ht = o->ptr;
+        dict_table = o->ptr;
+        shallow_copied_list_items = 1;
     } else if (o->type == OBJ_ZSET && o->encoding == OBJ_ENCODING_SKIPLIST) {
         zset *zs = o->ptr;
-        ht = zs->dict;
+        dict_table = zs->dict;
+        /* scanning ZSET allocates temporary strings even though it's a dict */
+        shallow_copied_list_items = 0;
     }
 
     list *keys = listCreate();
-    /* Set a free callback for the contents of the collected keys list.
-     * For the main keyspace dict, and when we scan a key that's dict encoded
-     * (we have 'ht'), we don't need to define free method because the strings
-     * in the list are just a shallow copy from the pointer in the dictEntry.
-     * When scanning a key with other encodings (e.g. listpack), we need to
-     * free the temporary strings we add to that list.
-     * The exception to the above is ZSET, where we do allocate temporary
-     * strings even when scanning a dict. */
-    if (o && (!ht || o->type == OBJ_ZSET)) {
+    /* Set a free callback for the contents of the collected keys list if they
+     * are deep copied temporary strings. We must not free them if they are just
+     * a shallow copy - a pointer to the actual data in the data structure */
+    if (!shallow_copied_list_items) {
         listSetFreeMethod(keys, (void (*)(void *))sdsfree);
     }
 
-    /* For main dictionary scan or data structure using hashtable. */
-    if (!o || ht) {
+    /* For main hash table scan or scannable data structure. */
+    if (!o || dict_table || hashtable_table) {
         /* We set the max number of iterations to ten times the specified
          * COUNT, so if the hash table is in a pathological state (very
          * sparsely populated) we avoid to block too much time at the cost
@@ -1188,7 +1206,7 @@ void scanGenericCommand(client *c, robj *o, unsigned long long cursor) {
 
         /* We pass scanData which have three pointers to the callback:
          * 1. data.keys: the list to which it will add new elements;
-         * 2. data.o: the object containing the dictionary so that
+         * 2. data.o: the object containing the hash table so that
          * it is possible to fetch more data in a type-dependent way;
          * 3. data.type: the specified type scan in the db, LLONG_MAX means
          * type matching is no needed;
@@ -1219,8 +1237,10 @@ void scanGenericCommand(client *c, robj *o, unsigned long long cursor) {
              * If cursor is empty, we should try exploring next non-empty slot. */
             if (o == NULL) {
                 cursor = kvstoreScan(c->db->keys, cursor, onlydidx, keysScanCallback, NULL, &data);
+            } else if (dict_table) {
+                cursor = dictScan(dict_table, cursor, dictScanCallback, &data);
             } else {
-                cursor = dictScan(ht, cursor, scanCallback, &data);
+                cursor = hashtableScan(hashtable_table, cursor, hashtableScanCallback, &data);
             }
         } while (cursor && maxiterations-- && data.sampled < count);
     } else if (o->type == OBJ_SET) {
--- a/src/debug.c
+++ b/src/debug.c
@@ -916,30 +916,35 @@ void debugCommand(client *c) {
         addReplyVerbatim(c, stats, sdslen(stats), "txt");
         sdsfree(stats);
     } else if (!strcasecmp(c->argv[1]->ptr, "htstats-key") && c->argc >= 3) {
-        robj *o;
-        dict *ht = NULL;
         int full = 0;
-
         if (c->argc >= 4 && !strcasecmp(c->argv[3]->ptr, "full")) full = 1;
 
-        if ((o = objectCommandLookupOrReply(c, c->argv[2], shared.nokeyerr)) == NULL) return;
+        robj *o = objectCommandLookupOrReply(c, c->argv[2], shared.nokeyerr);
+        if (o == NULL) return;
 
-        /* Get the hash table reference from the object, if possible. */
+        /* Get the dict reference from the object, if possible. */
+        dict *d = NULL;
+        hashtable *ht = NULL;
         switch (o->encoding) {
         case OBJ_ENCODING_SKIPLIST: {
             zset *zs = o->ptr;
-            ht = zs->dict;
+            d = zs->dict;
         } break;
-        case OBJ_ENCODING_HT: ht = o->ptr; break;
+        case OBJ_ENCODING_HT: d = o->ptr; break;
+        case OBJ_ENCODING_HASHTABLE: ht = o->ptr; break;
         }
 
-        if (ht == NULL) {
+        if (d != NULL) {
+            char buf[4096];
+            dictGetStats(buf, sizeof(buf), d, full);
+            addReplyVerbatim(c, buf, strlen(buf), "txt");
+        } else if (ht != NULL) {
+            char buf[4096];
+            hashtableGetStats(buf, sizeof(buf), ht, full);
+            addReplyVerbatim(c, buf, strlen(buf), "txt");
+        } else {
             addReplyError(c, "The value stored at the specified key is not "
                              "represented using an hash table");
-        } else {
-            char buf[4096];
-            dictGetStats(buf, sizeof(buf), ht, full);
-            addReplyVerbatim(c, buf, strlen(buf), "txt");
         }
     } else if (!strcasecmp(c->argv[1]->ptr, "change-repl-id") && c->argc == 2) {
         serverLog(LL_NOTICE, "Changing replication IDs after receiving DEBUG change-repl-id");
--- a/src/defrag.c
+++ b/src/defrag.c
@@ -34,6 +34,7 @@
  */
 
 #include "server.h"
+#include "hashtable.h"
 #include "script.h"
 #include <stddef.h>
 
@@ -379,6 +380,20 @@ static void activeDefragSdsDict(dict *d, int val_type) {
     } while (cursor != 0);
 }
 
+void activeDefragSdsHashtableCallback(void *privdata, void *entry_ref) {
+    UNUSED(privdata);
+    sds *sds_ref = (sds *)entry_ref;
+    sds new_sds = activeDefragSds(*sds_ref);
+    if (new_sds != NULL) *sds_ref = new_sds;
+}
+
+void activeDefragSdsHashtable(hashtable *ht) {
+    unsigned long cursor = 0;
+    do {
+        cursor = hashtableScanDefrag(ht, cursor, activeDefragSdsHashtableCallback, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
+    } while (cursor != 0);
+}
+
 /* Defrag a list of ptr, sds or robj string values */
 static void activeDefragQuickListNode(quicklist *ql, quicklistNode **node_ref) {
     quicklistNode *newnode, *node = *node_ref;
@@ -497,11 +512,9 @@ static void scanCallbackCountScanned(void *privdata, const dictEntry *de) {
 }
 
 static void scanLaterSet(robj *ob, unsigned long *cursor) {
-    if (ob->type != OBJ_SET || ob->encoding != OBJ_ENCODING_HT) return;
-    dict *d = ob->ptr;
-    dictDefragFunctions defragfns = {.defragAlloc = activeDefragAlloc,
-                                     .defragKey = (dictDefragAllocFunction *)activeDefragSds};
-    *cursor = dictScanDefrag(d, *cursor, scanCallbackCountScanned, &defragfns, NULL);
+    if (ob->type != OBJ_SET || ob->encoding != OBJ_ENCODING_HASHTABLE) return;
+    hashtable *ht = ob->ptr;
+    *cursor = hashtableScanDefrag(ht, *cursor, activeDefragSdsHashtableCallback, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
 }
 
 static void scanLaterHash(robj *ob, unsigned long *cursor) {
@@ -560,15 +573,16 @@ static void defragHash(robj *ob) {
 }
 
 static void defragSet(robj *ob) {
-    dict *d, *newd;
-    serverAssert(ob->type == OBJ_SET && ob->encoding == OBJ_ENCODING_HT);
-    d = ob->ptr;
-    if (dictSize(d) > server.active_defrag_max_scan_fields)
+    serverAssert(ob->type == OBJ_SET && ob->encoding == OBJ_ENCODING_HASHTABLE);
+    hashtable *ht = ob->ptr;
+    if (hashtableSize(ht) > server.active_defrag_max_scan_fields) {
         defragLater(ob);
-    else
-        activeDefragSdsDict(d, DEFRAG_SDS_DICT_NO_VAL);
-    /* defrag the dict struct and tables */
-    if ((newd = dictDefragTables(ob->ptr))) ob->ptr = newd;
+    } else {
+        activeDefragSdsHashtable(ht);
+    }
+    /* defrag the hashtable struct and tables */
+    hashtable *newHashtable = hashtableDefragTables(ht, activeDefragAlloc);
+    if (newHashtable) ob->ptr = newHashtable;
 }
 
 /* Defrag callback for radix tree iterator, called for each node,
@@ -766,7 +780,7 @@ static void defragKey(defragKeysCtx *ctx, robj **elemref) {
             serverPanic("Unknown list encoding");
         }
     } else if (ob->type == OBJ_SET) {
-        if (ob->encoding == OBJ_ENCODING_HT) {
+        if (ob->encoding == OBJ_ENCODING_HASHTABLE) {
             defragSet(ob);
         } else if (ob->encoding == OBJ_ENCODING_INTSET || ob->encoding == OBJ_ENCODING_LISTPACK) {
             void *newptr, *ptr = ob->ptr;
--- a/src/object.c
+++ b/src/object.c
@@ -429,9 +429,9 @@ robj *createListListpackObject(void) {
 }
 
 robj *createSetObject(void) {
-    dict *d = dictCreate(&setDictType);
-    robj *o = createObject(OBJ_SET, d);
-    o->encoding = OBJ_ENCODING_HT;
+    hashtable *ht = hashtableCreate(&setHashtableType);
+    robj *o = createObject(OBJ_SET, ht);
+    o->encoding = OBJ_ENCODING_HASHTABLE;
     return o;
 }
 
@@ -506,7 +506,7 @@ void freeListObject(robj *o) {
 
 void freeSetObject(robj *o) {
     switch (o->encoding) {
-    case OBJ_ENCODING_HT: dictRelease((dict *)o->ptr); break;
+    case OBJ_ENCODING_HASHTABLE: hashtableRelease((hashtable *)o->ptr); break;
     case OBJ_ENCODING_INTSET:
     case OBJ_ENCODING_LISTPACK: zfree(o->ptr); break;
     default: serverPanic("Unknown set encoding type");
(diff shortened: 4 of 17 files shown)


tests/unit/type/set.tcl
```diff
@@ -1029,111 +1029,6 @@ foreach type {single multiple single_multiple} {
         r srem $myset {*}$members
     }
 
-    proc verify_rehashing_completed_key {myset table_size keys} {
-        set htstats [r debug HTSTATS-KEY $myset]
-        assert {![string match {*rehashing target*} $htstats]}
-        return {[string match {*table size: $table_size*number of elements: $keys*} $htstats]}
-    }
-
-    test "SRANDMEMBER with a dict containing long chain" {
```
rainsupreme, line 1038: I deleted this test because hashset does not have linked list chains the way that dict does, so the aspect this is attempting to test no longer exists.

zuiderkwast: For the keys and expires using hashet, I updated `DEBUG HTSTATS` to count probing chain lengths instead of linked list lengths. Maybe it makes sense here too...?

src/hashset.c
```diff
@@ -71,6 +71,7 @@
  *   addressing scheme, including the use of linear probing by scan cursor
  *   increment, by Viktor Söderqvist. */
 #include "hashset.h"
+#include "server.h"
```
zuiderkwast: I tried hard to avoid having hashset depend on the whole valkey server.h. It's better to have dependencies in only one direction and allow hashset to be a relatively independent component only depending on low level stuff like zmalloc.h.

You added this for dismissMemory, right?

I think we can move the logic of dismissMemory from `object.c` to `zmalloc.c`. `dismissMemory` basically just calls `zmadvise_dontneed` which already knows the page size without using `server.page_size`. It doesn't accept a size parameter though, but we can change `zmadvise_dontneed` since it's actually only called from dismissMemory. We can add a size parameter and make it do all what dismissMemory does. In server.h we can add a dismissMemory as an alias (define) of `zmadvise_dontneed`.

```C
/* server.h */
#define dismissMemory zmadvise_dontneed

/* zmalloc.c */
void zmadvise_dontneed(void *ptr, size_t size_hint) {
    /* Code moved from dismissMemory */
    ...
    /* Code that was already in zmadvise_dontneed since before */
    ...
}
```

src/hashset.c
```diff
@@ -1511,7 +1523,6 @@ unsigned hashsetSampleElements(hashset *t, void **dst, unsigned count) {
     while (samples.count < count) {
         cursor = hashsetScan(t, cursor, sampleElementsScanFn, &samples, HASHSET_SCAN_SINGLE_STEP);
     }
-    rehashStepOnReadIfNeeded(t);
```
zuiderkwast: You're moving this back to scan? It's been moving back and forth, but not doing it in scan seems to be more similar to what dict does.

zuiderkwast (commented):
Partial review. I'll look more later.

src/hashset.c
rainsupreme: just a mistake - I think I overwrote changes at some point when I shouldn't have 😅

zuiderkwast:
Sorry for force-pushing the hashset branch again, to fix a DCO issue.

Can you rebase and force-push again? (I guess it's better than merge when we have a DCO issue following us.)

tests/unit/type/set.tcl
rainsupreme: Hmm, I think this should be a hashset UT instead. Assuming that our random sampling doesn't follow probe chains we should be unaffected, but we want to guard against regressions in the future. My UT would make a hashset with one long chain of similar elements, then ensure those elements aren't under or over represented.

zuiderkwast: Sounds good. Can you write that UT in another PR towards the hashset branch?

src/db.c
```diff
@@ -1065,34 +1086,37 @@ void scanGenericCommand(client *c, robj *o, unsigned long long cursor) {
      * just return everything inside the object in a single call, setting the
      * cursor to zero to signal the end of the iteration. */
 
-    /* Handle the case of a hash table. */
-    ht = NULL;
+    /* Handle the case of a dict or hashset. */
+    dict *dictTable = NULL;
+    hashset *hashsetTable = NULL;
```
zuiderkwast: We don't use camel case for local variables.

The "Table" suffix seems redundant. You don't like short names like `d` and `hs`, like you did in debug htstats-key?

...

Thinking about this more, maybe we don't even need these variables. `ht` is only used in a few places in this function and we can just check `o->encoding` directly in those places. In the calls to dictScan and hashsetScan, we can pass `o->ptr`. WDYT?

src/db.c
```diff
@@ -1065,34 +1086,37 @@ void scanGenericCommand(client *c, robj *o, unsigned long long cursor) {
...
+     * (we have 'dictTable'), or when we scan a key that's hashset encoded
+     * (we have 'hashsetTable') we don't need to define free method because the
+     * strings in the list are just a shallow copy from the pointer in the
+     * dictEntry.
      * When scanning a key with other encodings (e.g. listpack), we need to
      * free the temporary strings we add to that list.
      * The exception to the above is ZSET, where we do allocate temporary
      * strings even when scanning a dict. */
-    if (o && (!ht || o->type == OBJ_ZSET)) {
+    if (o && ((!dictTable && !hashsetTable) || o->type == OBJ_ZSET)) {
```
zuiderkwast: I wonder if this is the clearest way to express this. It seems to me that just checking `o->encoding` directly would be more clear.

```suggestion
    if (o && ((o->encoding != OBJ_ENCODING_HT && o->encoding != OBJ_ENCODING_HASHSET) || o->type == OBJ_ZSET)) {
```

src/hashset.c
```diff
@@ -1299,18 +1308,16 @@ size_t hashsetScan(hashset *s, size_t cursor, hashsetScanFunction fn, void *priv
...
              * been rehashed. */
-            if (table_small == 0 && !cursorIsLessThan(cursor, s->rehash_idx)) {
-                bucket *b = &s->tables[table_small][cursor & mask_small];
-                if (b->presence) {
-                    for (int pos = 0; pos < ELEMENTS_PER_BUCKET; pos++) {
-                        if (isPositionFilled(b, pos)) {
-                            void *emit = emit_ref ? &b->elements[pos] : b->elements[pos];
-                            fn(privdata, emit);
-                        }
+            bucket *b = &s->tables[table_small][cursor & mask_small];
```
zuiderkwast: Here you give yourself merge conflicts with your other PR. Do this at your own risk. :)

IMO it's better avoid changes to `hashset.{c,h}` in this PR. I did that so far: When I needed something for the robj implementation, I did it in the hashset branch and rebased my robj stuff on top of that afterwards.

src/t_set.c
```diff
@@ -242,9 +243,9 @@ int setTypeRemoveAux(robj *setobj, char *str, size_t len, int64_t llval, int str
         str_is_sds = 0;
     }
 
-    if (setobj->encoding == OBJ_ENCODING_HT) {
+    if (setobj->encoding == OBJ_ENCODING_HASHSET) {
         sds sdsval = str_is_sds ? (sds)str : sdsnewlen(str, len);
-        int deleted = (dictDelete(setobj->ptr, sdsval) == DICT_OK);
+        int deleted = (hashsetDelete(setobj->ptr, sdsval));
```
zuiderkwast:
```suggestion
        int deleted = hashsetDelete(setobj->ptr, sdsval);
```

src/t_zset.c
```diff
@@ -2173,13 +2169,9 @@ int zuiNext(zsetopsrc *op, zsetopval *val) {
 
             /* Move to next element. */
             it->is.ii++;
-        } else if (op->encoding == OBJ_ENCODING_HT) {
-            if (it->ht.de == NULL) return 0;
-            val->ele = dictGetKey(it->ht.de);
+        } else if (op->encoding == OBJ_ENCODING_HASHSET) {
+            if (!hashsetNext(it->ht.iter, (void **)&(val->ele))) return 0;
```
```suggestion
            if (!hashsetNext(it->ht.iter, (void **)&val->ele)) return 0;
```

tests/unit/type/set.tcl
```diff
@@ -33,6 +33,7 @@ start_server {
         assert_equal {0 1} [r smismember myset bla foo]
         assert_equal {0} [r smismember myset bla]
         assert_equal "bar $initelems($type)" [lsort [r smembers myset]]
+        r memory usage myset
```
zuiderkwast, line 36: Added for debugging?

zuiderkwast (commented):
This looks really good. Just a few nits.

src/hashset.c
rainsupreme: That is true - I intend to discard all hashset.* changes in this PR once my bugfix PRs are merged upstream, but some minimal fixes were needed here for tests to succeed while I wait for that. :)

tests/unit/type/set.tcl
rainsupreme: To ensure the command runs without crashing. idk, what do you think? 🤷🏻‍♀️

src/db.c
rainsupreme: You're right, this could be clearer. I think setting a list_items_shallow_copied int in the if/else above will be most clear.

fwiw, once all the data types are ported to hashset the parallel dict/hashset stuff like this will be gone

src/db.c
rainsupreme: I dunno, there would have to be a special case for ZSET though. Good point about the naming at any rate!

I think this code needs an overall refactor. It's supposed to be shared scan code but then it does 6 different kinds of scanning depending on type and encoding. Gross. Not super quick to untangle though, so it'd have to be a different PR.

tests/unit/type/set.tcl
zuiderkwast: To increase coverage? Ok, yes, good.

src/db.c
zuiderkwast: Yeah good point.

rainsupreme:
Thanks Viktor for taking care of the rebase - your changes look good to me. :)

zuiderkwast:
You're welcome. :) The main thing I don't like right now is OBJ_ENCODING_HT for dict vs OBJ_ENCODING_HASHTABLE for hashtable. Maybe we should rename the former to OBJ_ENCODING_DICT and the latter to OBJ_ENCODING_HT instead?

rainsupreme:
That's a fair point, but OBJECT_ENCODING_HT will be removed completely when all data types have been converted to hashset, so it is only an issue short term, I think. I don't want to touch extra files just to rename it temporarily - I feel like that might make the PR more confusing and harder to review. :-/

zuiderkwast:
Are you updating this? It becomes very hard to rebase it if you wait this long between each rebase.

rainsupreme:
I did rebase about a week ago - might have forgotten to push. Ice just
finished my oncall shift and rebasing is at the top of my list. 🙂 really
looking forward to getting back to this

zuiderkwast:
I rebased it (started from unstable and applied commit with cherry-pick) but then I forgot `git cherry-pick --continue` after solving the merge conflict. When I pushed, I just pushed unstable, causing the diff to become empty and then the PR closed. Now I can't push to your branch to fix it. :cry:

rainsupreme:
I got the PR fixed and back up - ready to review and merge! Thanks @zuiderkwast for your work rebasing this again. 😊

src/db.c
```diff
@@ -1013,13 +1011,36 @@ void scanCallback(void *privdata, const dictEntry *de) {
...
+    list *keys = data->keys;
+    data->sampled++;
+
+    /* currently only implemented for SET scan */
+    serverAssert(o && o->type == OBJ_SET && o->encoding == OBJ_ENCODING_HASHTABLE);
+    sds key = (sds)entry; /* Specific for OBJ_SET */
+
+    /* o and typename can not have values at the same time. */
+    serverAssert(!((data->type != LLONG_MAX) && o));
+
```
zuiderkwast: This scan callback is not used for keys, since Jim's changes. Keys use a separate scan callback.

Here, o is always set and data->type is never used. We can delete this assert.

```suggestion
```

src/db.c
```diff
@@ -1152,34 +1172,35 @@ void scanGenericCommand(client *c, robj *o, unsigned long long cursor) {
...
+    /* Set a free callback for the contents of the collected keys list if they
+     * are deep copied temporary strings. We must not free them if they are just
+     * a shallow copy - a pointer to the actual data in the data structure */
+    if (!shallow_copied_list_items) {
         listSetFreeMethod(keys, (void (*)(void *))sdsfree);
     }
 
-    /* For main dictionary scan or data structure using hashtable. */
-    if (!o || ht) {
+    /* For main dictionary scan or scannable data structure. */
```
zuiderkwast: I don't like the word "dictionary" for something that's not a dict.

```suggestion
    /* For main hash table scan or scannable data structure. */
```

src/debug.c
```diff
@@ -916,30 +916,35 @@ void debugCommand(client *c) {
...
         if (c->argc >= 4 && !strcasecmp(c->argv[3]->ptr, "full")) full = 1;
 
-        if ((o = objectCommandLookupOrReply(c, c->argv[2], shared.nokeyerr)) == NULL) return;
+        robj *o = objectCommandLookupOrReply(c, c->argv[2], shared.nokeyerr);
+        if (o == NULL) return;
 
-        /* Get the hash table reference from the object, if possible. */
+        /* Get the dict reference from the object, if possible. */
+        dict *d = NULL;
+        hashtable *hs = NULL;
```
zuiderkwast:
```suggestion
        hashtable *ht = NULL;
```

src/debug.c
```diff
@@ -916,30 +916,35 @@ void debugCommand(client *c) {
...
         switch (o->encoding) {
         case OBJ_ENCODING_SKIPLIST: {
             zset *zs = o->ptr;
-            ht = zs->dict;
+            d = zs->dict;
         } break;
-        case OBJ_ENCODING_HT: ht = o->ptr; break;
+        case OBJ_ENCODING_HT: d = o->ptr; break;
+        case OBJ_ENCODING_HASHTABLE: hs = o->ptr; break;
```
zuiderkwast:
```suggestion
        case OBJ_ENCODING_HT: d = o->ptr; break;
        case OBJ_ENCODING_HASHTABLE: ht = o->ptr; break;
```

src/debug.c
```diff
@@ -916,30 +916,35 @@ void debugCommand(client *c) {
...
-                             "represented using an hash table");
-        } else {
+        if (d != NULL) {
             char buf[4096];
-            dictGetStats(buf, sizeof(buf), ht, full);
+            dictGetStats(buf, sizeof(buf), d, full);
             addReplyVerbatim(c, buf, strlen(buf), "txt");
+        } else if (hs != NULL) {
+            char buf[4096];
+            hashtableGetStats(buf, sizeof(buf), hs, full);
```
zuiderkwast:
```suggestion
        } else if (ht != NULL) {
            char buf[4096];
            hashtableGetStats(buf, sizeof(buf), ht, full);
```

src/defrag.c
```diff
@@ -378,6 +379,20 @@ static void activeDefragSdsDict(dict *d, int val_type) {
...
+    UNUSED(privdata);
+    sds *sds_ref = (sds *)entry_ref;
+    sds new_sds = activeDefragSds(*sds_ref);
+    if (new_sds != NULL) *sds_ref = new_sds;
+}
+
+void activeDefragSdsHashtable(hashtable *hs) {
+    unsigned long cursor = 0;
+    do {
+        cursor = hashtableScanDefrag(hs, cursor, activeDefragSdsHashtableCallback, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
```
zuiderkwast:
```suggestion
void activeDefragSdsHashtable(hashtable *ht) {
    unsigned long cursor = 0;
    do {
        cursor = hashtableScanDefrag(ht, cursor, activeDefragSdsHashtableCallback, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
```

src/defrag.c
```diff
@@ -496,11 +511,9 @@ static void scanCallbackCountScanned(void *privdata, const dictEntry *de) {
...
 
 static void scanLaterSet(robj *ob, unsigned long *cursor) {
-    if (ob->type != OBJ_SET || ob->encoding != OBJ_ENCODING_HT) return;
-    dict *d = ob->ptr;
-    dictDefragFunctions defragfns = {.defragAlloc = activeDefragAlloc,
-                                     .defragKey = (dictDefragAllocFunction *)activeDefragSds};
-    *cursor = dictScanDefrag(d, *cursor, scanCallbackCountScanned, &defragfns, NULL);
+    if (ob->type != OBJ_SET || ob->encoding != OBJ_ENCODING_HASHTABLE) return;
+    hashtable *hs = ob->ptr;
+    *cursor = hashtableScanDefrag(hs, *cursor, activeDefragSdsHashtableCallback, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
```
zuiderkwast:
```suggestion
    hashtable *ht = ob->ptr;
    *cursor = hashtableScanDefrag(ht, *cursor, activeDefragSdsHashtableCallback, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
```

src/defrag.c
```diff
@@ -559,15 +572,16 @@ static void defragHash(robj *ob) {
...
+    if (hashtableSize(ht) > server.active_defrag_max_scan_fields) {
         defragLater(ob);
-    else
-        activeDefragSdsDict(d, DEFRAG_SDS_DICT_NO_VAL);
-    /* defrag the dict struct and tables */
-    if ((newd = dictDefragTables(ob->ptr))) ob->ptr = newd;
+    } else {
+        activeDefragSdsHashtable(ht);
+    }
+    /* defrag the hashtable struct and members */
```
zuiderkwast: Members can mean elements, so let's not use this word.

```suggestion
    /* defrag the hashtable struct and tables */
```

src/object.c
```diff
@@ -622,23 +622,22 @@ void dismissListObject(robj *o, size_t size_hint) {
...
-            dictEntry *de;
-            dictIterator *di = dictGetIterator(set);
-            while ((de = dictNext(di)) != NULL) {
-                dismissSds(dictGetKey(de));
+        if (size_hint / hashtableSize(ht) >= server.page_size) {
+            hashtableIterator iter;
+            hashtableInitIterator(&iter, ht);
+            sds item;
+            while (hashtableNext(&iter, (void **)(&item))) {
+                dismissSds(item);
```
zuiderkwast: I have changed all iterations to avoid the cast to `(void **)`. Let's do it here too.

```suggestion
            void *next;
            while (hashtableNext(&iter, &next)) {
                sds item = next;
                dismissSds(item);
```

src/object.c
```diff
@@ -1160,17 +1160,19 @@ size_t objectComputeSize(robj *key, robj *o, size_t sample_size, int dbid) {
...
-                elesize += dictEntryMemUsage(de) + sdsAllocSize(ele);
+        if (o->encoding == OBJ_ENCODING_HASHTABLE) {
+            hashtable *ht = o->ptr;
+            asize = sizeof(*o) + hashtableMemUsage(ht);
+
+            hashtableIterator iter;
+            hashtableInitIterator(&iter, ht);
+            sds element;
+            while (hashtableNext(&iter, (void **)&element) && samples < sample_size) {
+                elesize += sdsAllocSize(element);
```
zuiderkwast:
```suggestion
            void *next;
            while (hashtableNext(&iter, &next) && samples < sample_size) {
                sds element = next;
                elesize += sdsAllocSize(element);
```

src/rdb.c
```diff
@@ -876,26 +876,25 @@ ssize_t rdbSaveObject(rio *rdb, robj *o, robj *key, int dbid) {
...
                 return -1;
             }
             nwritten += n;
 
-            while ((de = dictNext(di)) != NULL) {
-                sds ele = dictGetKey(de);
+            hashtableIterator iterator;
+            hashtableInitIterator(&iterator, set);
+            sds ele;
+            while (hashtableNext(&iterator, (void **)&ele)) {
```
zuiderkwast:
```suggestion
            void *next;
            while (hashtableNext(&iterator, &next)) {
                sds ele = next;
```

Alternatively, we could just use `void *ele` and `&ele`. Things like `sdslen(ele)` still works.

src/t_set.c
```diff
@@ -134,20 +135,21 @@ int setTypeAddAux(robj *set, char *str, size_t len, int64_t llval, int str_is_sd
...
-    if (set->encoding == OBJ_ENCODING_HT) {
+    if (set->encoding == OBJ_ENCODING_HASHTABLE) {
         /* Avoid duping the string if it is an sds string. */
         sds sdsval = str_is_sds ? (sds)str : sdsnewlen(str, len);
-        dict *ht = set->ptr;
-        void *position = dictFindPositionForInsert(ht, sdsval, NULL);
-        if (position) {
+        hashtable *hs = set->ptr;
+        hashtablePosition position;
+        if (hashtableFindPositionForInsert(hs, sdsval, &position, NULL)) {
```
zuiderkwast:
```suggestion
        hashtable *ht = set->ptr;
        hashtablePosition position;
        if (hashtableFindPositionForInsert(ht, sdsval, &position, NULL)) {
```

src/t_set.c
```diff
@@ -134,20 +135,21 @@ int setTypeAddAux(robj *set, char *str, size_t len, int64_t llval, int str_is_sd
...
             /* Key doesn't already exist in the set. Add it but dup the key. */
             if (sdsval == str) sdsval = sdsdup(sdsval);
-            dictInsertAtPosition(ht, sdsval, position);
+            hashtableInsertAtPosition(hs, sdsval, &position);
```
zuiderkwast:
```suggestion
            hashtableInsertAtPosition(ht, sdsval, &position);
```

src/t_set.c
```diff
@@ -353,10 +355,8 @@ void setTypeReleaseIterator(setTypeIterator *si) {
  *
  * When there are no more elements -1 is returned. */
 int setTypeNext(setTypeIterator *si, char **str, size_t *len, int64_t *llele) {
-    if (si->encoding == OBJ_ENCODING_HT) {
-        dictEntry *de = dictNext(si->di);
-        if (de == NULL) return -1;
-        *str = dictGetKey(de);
+    if (si->encoding == OBJ_ENCODING_HASHTABLE) {
+        if (!hashtableNext(si->hashtable_iterator, (void **)str)) return -1;
         *len = sdslen(*str);
```
zuiderkwast: How can we avoid the cast here? Like this?

```suggestion
        void *next;
        if (!hashtableNext(si->hashtable_iterator, &next)) return -1;
        *str = next;
        *len = sdslen(next);
```

src/t_set.c
```diff
@@ -406,15 +406,15 @@ sds setTypeNextObject(setTypeIterator *si) {
...
  * Note that both the str, len and llele pointers should be passed and cannot
  * be NULL. If str is set to NULL, the value is an integer stored in llele. */
 int setTypeRandomElement(robj *setobj, char **str, size_t *len, int64_t *llele) {
-    if (setobj->encoding == OBJ_ENCODING_HT) {
-        dictEntry *de = dictGetFairRandomKey(setobj->ptr);
-        *str = dictGetKey(de);
+    if (setobj->encoding == OBJ_ENCODING_HASHTABLE) {
+        *str = NULL;
+        hashtableFairRandomEntry(setobj->ptr, (void **)str);
         *len = sdslen(*str);
```
zuiderkwast: Like this? (untested)

```suggestion
        void *entry = NULL;
        hashtableFairRandomEntry(setobj->ptr, &entry);
        *str = entry;
        *len = sdslen(entry);
```

src/t_set.c
```diff
@@ -489,28 +489,28 @@ int setTypeConvertAndExpand(robj *setobj, int enc, unsigned long cap, int panic)
     setTypeIterator *si;
     serverAssertWithInfo(NULL, setobj, setobj->type == OBJ_SET && setobj->encoding != enc);
 
-    if (enc == OBJ_ENCODING_HT) {
-        dict *d = dictCreate(&setDictType);
+    if (enc == OBJ_ENCODING_HASHTABLE) {
+        hashtable *hs = hashtableCreate(&setHashtableType);
```
zuiderkwast: Change to ht.

Try `git grep -F 'hashtable *hs'` to find all.

src/t_set.c
```diff
@@ -1126,29 +1124,25 @@ void srandmemberWithCountCommand(client *c) {
...
 
         /* Remove random elements to reach the right count. */
         while (size > count) {
-            dictEntry *de;
-            de = dictGetFairRandomKey(d);
-            dictUnlink(d, dictGetKey(de));
-            sdsfree(dictGetKey(de));
-            dictFreeUnlinkedEntry(d, de);
+            sds element;
+            hashtableFairRandomEntry(ht, (void **)&element);
```
zuiderkwast: Avoid the cast to `(void **)`.

src/t_set.c
```diff
@@ -1181,14 +1175,15 @@ void srandmemberWithCountCommand(client *c) {
...
+        hashtableInitIterator(&iter, ht);
 
         addReplyArrayLen(c, count);
-        di = dictGetIterator(d);
-        while ((de = dictNext(di)) != NULL) addReplyBulkSds(c, dictGetKey(de));
-        dictReleaseIterator(di);
-        dictRelease(d);
+        serverAssert(count == hashtableSize(ht));
+        sds element;
+        while (hashtableNext(&iter, (void **)&element)) addReplyBulkSds(c, element);
```
zuiderkwast: Avoid this cast

src/t_zset.c
```diff
@@ -2235,13 +2231,9 @@ int zuiNext(zsetopsrc *op, zsetopval *val) {
 
             /* Move to next element. */
             it->is.ii++;
-        } else if (op->encoding == OBJ_ENCODING_HT) {
-            if (it->ht.de == NULL) return 0;
-            val->ele = dictGetKey(it->ht.de);
+        } else if (op->encoding == OBJ_ENCODING_HASHTABLE) {
+            if (!hashtableNext(it->ht.iter, (void **)&val->ele)) return 0;
```
zuiderkwast: Avoid this cast

zuiderkwast (commented):
Nits.

Changes to the merged code after code review include avoiding the cast to `void **`. (Related to the strict aliasing rule or just nicer code?) We should do the corresponding changes here.

When I rebased, I just made it compile. I didn't update anything else, like `hs` variable names.

rainsupreme:
It seems like most of these check failures are also happening in unstable? :(

zuiderkwast:
> It seems like most of these check failures are also happening in unstable? :(

Yes, it's not a blocker for this PR. We should fix them in unstable.

tests/unit/type/set.tcl
```diff
@@ -177,23 +179,23 @@ foreach type {single multiple single_multiple} {
...
         for {set i 0} {$i <  100} {incr i} { r sadd myintset $i }
         for {set i 0} {$i < 1280} {incr i} { r sadd mylargeintset $i }
         for {set i 0} {$i <   50} {incr i} { r sadd mysmallset [format "i%03d" $i] }
-        for {set i 0} {$i <  256} {incr i} { r sadd myhashset [format "i%03d" $i] }
+        for {set i 0} {$i <  256} {incr i} { r sadd myhashtable [format "i%03d" $i] }
         assert_encoding intset myintset
         assert_encoding hashtable mylargeintset
         assert_encoding listpack mysmallset
-        assert_encoding hashtable myhashset
+        assert_encoding hashtable myhashtable
```
zuiderkwast: This test case used "myhashset" even with a dict? All sets here end in "set". I think it just means a set backed by a hash table, so it's not wrong.

tests/unit/type/set.tcl
```diff
@@ -177,23 +179,23 @@ foreach type {single multiple single_multiple} {
 
     test "Set encoding after DEBUG RELOAD" {
         r del myintset
-        r del myhashset
+        r del myhashtable
```
zuiderkwast: Here, hashset just means a set backed by a hashtable. It's not new. This name was used also for the dict-backed set.

```suggestion
        r del myhashset
```

tests/unit/type/set.tcl
```diff
@@ -177,23 +179,23 @@ foreach type {single multiple single_multiple} {
...
         r del mylargeintset
         r del mysmallset
         for {set i 0} {$i <  100} {incr i} { r sadd myintset $i }
         for {set i 0} {$i < 1280} {incr i} { r sadd mylargeintset $i }
         for {set i 0} {$i <   50} {incr i} { r sadd mysmallset [format "i%03d" $i] }
-        for {set i 0} {$i <  256} {incr i} { r sadd myhashset [format "i%03d" $i] }
+        for {set i 0} {$i <  256} {incr i} { r sadd myhashtable [format "i%03d" $i] }
```
zuiderkwast:
```suggestion
        for {set i 0} {$i <  256} {incr i} { r sadd myhashset [format "i%03d" $i] }
```

tests/unit/type/set.tcl
zuiderkwast:
```suggestion
        assert_encoding hashtable myhashset
```

tests/unit/type/set.tcl
```diff
@@ -177,23 +179,23 @@ foreach type {single multiple single_multiple} {
...
         assert_encoding listpack mysmallset
-        assert_encoding hashtable myhashset
+        assert_encoding hashtable myhashtable
 
         r debug reload
         assert_encoding intset myintset
         assert_encoding hashtable mylargeintset
         assert_encoding listpack mysmallset
-        assert_encoding hashtable myhashset
+        assert_encoding hashtable myhashtable
```
zuiderkwast:
```suggestion
        assert_encoding hashtable myhashset
```

zuiderkwast (approved):

PR #1178, opened 2024-10-16
[hashset feature] Replace dict with hashset for keys and expire

### LINKED ISSUE #992: Embed key and TTL in robj
This is to prepare for storing `robj` in a new hash table implementation.

This includes replace dict with hashset in kvstore, db, expire, ...

The new hash table implementation doesn't have a "dict entry", nor does it have keys and values. It just has "elements". An element can effectively be anything that contains the key. The user provides a "get key" callback function in the hashsetType (same idea as the callbacks in dictType).

```
 Hash table
+-----------+        +----------------+
| 1         |        | Object with    |
| 2       ---------> | embedded key,  |
| 3         |        | value, LRU and |
+-----------+        | more...        |
                     +----------------+
```

To use an `robj` as a container for both key and value, we can embed the key into the `robj`. The `robj` (typedef'ed to `struct serverObject`) is accessed in very many places, so we can't easily make it opaque and change all the accesses to it. That would be very much work. Instead, we can add some extra fields (for optional embedded key and embedded TTL) and keep the existing fields. Wherever the robj is passed around in legacy code, an embedded key will just be ignored.

```C
struct serverObject {
    unsigned type : 4;
    unsigned encoding : 4;
    unsigned lru : LRU_BITS; /* LRU_BITS = 24 */
    int refcount;
};
```

```
           robj
+-----------------------+
| type | encoding | LRU |
|-----------------------|
| refcount              |
|-----------------------|      +--------+
| ptr             ------------>| Value  |
+-----------------------+      +--------+
```

There's is already something called embedded *string* (`obj->encoding = OBJ_ENCODING_EMBSTR`, created by `createEmbeddedStringObject`) which allocates extra space after the struct, copies the string contents (an sds string) into the same allocation and points `ptr` to it. We can call this "embedded value" and it's not the same as an embedded key, which doesn't exist yet.

```
     robj with embedded value (legacy)
    +-----------------------+
    | type | encoding | LRU |
    |-----------------------|
    | refcount              |
    |-----------------------|
  ,---ptr                   |
 |  |-----------------------+
 |  | sds header            |
  `-> sds contents          |
    |                       |
    +-----------------------+
```

The sds strings are somewhat weird since they are pointers to the string contents, and they also have a header which is before the content (so internally, the sds library does `s[-1]` to look at the bytes before the content). That's why `ptr` points to the sds content, rather than to the sds header.

The `ptr` pointer takes up some space (8 bytes) in the struct and it would be possible to avoid it (by computing it instead), but it's accessed in thousands of places just like `obj->ptr` so let's leave it like that for now. We can still embed the key. We can avoid adding another pointer to the embedded key though, if we use the same method that was used in #541 to embed a key in dictEntry.

To embed also a *key* (another sds string) into the same robj, let's put the key before the value, because the key never changes. It can just be deleted, and then the whole robj is deleted. The value can change though (shrink or grows, possibly be moved out to its own allocation).

Let's use a bit from `refcount` to flag that there's an embedded key. We can also embed TTL and add a "has ttl" flag.

```diff
 struct serverObject {
     unsigned type : 4;
     unsigned encoding : 4;
     unsigned lru : LRU_BITS; /* LRU_BITS = 24 */
-    int refcount;
+    unsigned hasexpire : 1;
+    unsigned hasembkey : 1;
+    int refcount : 29;
     void *ptr;
+    /* Embedded stuff */
 };
```

The `data[]` field isn't strictly needed, since it's zero size by default and it's possible to allocate more space and use it anyway, but with a data field, we can access the end of the struct like `obj->data` and `&obj->data[offset]`.

```
     Future layout of robj
    +-----------------------+
    | type | encoding | LRU |
    |-----------------------|
    | embkey | hasttl       |
    |-----------------------|
    | refcount              |
    |-----------------------|
  ,---ptr                   |
 |  |-----------------------|
 |  | expire                | Embedded expire (TTL)
 |  |-----------------------| (optional)
 |  | size of sds header    |
 |  | sds header            | Embedded key
 |  | sds contents          | (optional)
 |  |-----------------------|
 |  | sds header            |
  `-> sds contents          | Embedded value
    |                       | (optional)
    +-----------------------+
```

Note the added "size of sds header" in the drawing above. There are different variants of the sds header, with different header sizes. (See `src/sds.h`.) Currently, `createEmbeddedStringObject` always embeds the header `struct sdshdr8` (which is a header of 3 bytes and the max length of strings of this sds variant 256 bytes, since it uses one byte to stores the length of the string). Embedded strings are (currently) only used for short strings anyway.

When we embed keys, though, we want to embed them regardless of their size. Thus we want to able to use different sds headers. We add one byte to store the size of the sds header, so we can compute where the sds content is, and from that we can call `sdslen` and other sds functions to figure out the length of the sds string and where it ends.

Now, what we need is a function to combine two `robj` into one, with an embedded key and a (embedded or not) value.

~~If we are smart, we can convert an object with an embedded value into an object with an embedded key and no value (ptr = NULL), without reallocating or moving the actual value. That's why I'm suggesting (not sure if it's a good idea) that we add a "sds header size" byte also for embedded value. When we convert an robj with embedded string into an robj with embedded key and no value, we just set the `obj->embkey = 1; obj->ptr = NULL;`.~~

Update: The strings from a client are parsed and stored as `robj` objects in `c->argv` and this argument vector needs to be valid even after the command has been executed, for replication and other purposes, so when we're handling a command like `SET k v`, `k` and `v` are both robj with a valid string value "k" and "v" respectively. We need to embed "k" as embedded key in the `v` object, without destroying `k` or turning the `k` into something else. Currently (as of valkey 8.0-rc1) the keys are embedded in a dictEntry and they're copied there (which seems fine, performance wise). Similarly we need to *copy* the string content of `k` into the 'embedded-key' part of `v` and then add `v` to the hash table.

zuiderkwast:
This PR is based on #1186. When it is merged, this PR will change target to unstable.

Keys and expire are embedded in `robj`.

Dict is replaced by hashset for the keys and expire tables. Hashset is a hash table that doesn't have a dict entry. Pointers point directly to robj, where also the expire is stored, if any. Also pubsub channels are changed, since they use kvstore. Kvstore is changed from dict to hashset.

Some db.c functions like `dbAdd`, `setKey` and `setExpire` are changed to return the reallocated object after embedding the key and optional expire. `setKey` does not increment the reference counter, since it would require duplicating the object.

An `robj` value with an embedded key can be called valkey (val + key). A valkey typedef is added for robj and it's used when the object has an embedded key, to distinguish it from an object from other usages of robj (such as a string object from the command parser in c->argv).

```
keys                                      expire
hashset            robj                   hashset
+---+         +-------------------+        +---+
| 1 --------->| type, enc, lru    |<-------- 1 |
| 2 |         | refcount, flags   |        | 2 |
| 3 |         | ptr               |        | 3 |
| . |         | expire?           |        | . |
| . |         | embedded key      |        | . |
| . |         | embedded value?   |        | . |
+---+         +-------------------+        +---+

? = optional
```

A quick benchmark using `valkey-benchmark` showed small improvements for all types, but more for get and set than complex types.

Rebarding memory usage, `DEBUG POPULATE 10000000` (10M) resulted in memory usage (`used_memory_human`) of 669M, compared to 872M on unstable. That's a **23% drop in memory usage** and **21.28 bytes per key**. (DEBUG POPULATE created keys of length 11 and string values of length 13. Both of these fit in an robj allocation of size 32 bytes.)

Fixes #992.

TODO:

* Implement memory prefetching (for io threading, to be ported from the dict implementation).


src/kvstore.h
```diff
@@ -1,85 +1,90 @@
-#ifndef DICTARRAY_H_
-#define DICTARRAY_H_
+#ifndef KVSTORE_H
+#define KVSTORE_H
 
-#include "dict.h"
+#include "hashset.h"
```
madolson, line 4: I think you forgot to include this file

src/server.h
```diff
@@ -871,16 +869,19 @@ struct ValkeyModuleDigest {
...
     unsigned type : 4;
     unsigned encoding : 4;
     unsigned lru : LRU_BITS; /* LRU time (relative to global lru_clock) or
                               * LFU data (least significant 8 bits frequency
                               * and most significant 16 bits access time). */
-    int refcount;
+    unsigned hasexpire : 1;
+    unsigned hasembkey : 1;
+    unsigned hasembkeyptr : 1;
+    unsigned refcount : 30;
```
rainsupreme: adds up to 33? should refcount be 29 bits?

src/server.h
```diff
@@ -871,16 +869,19 @@ struct ValkeyModuleDigest {
 #define LRU_CLOCK_MAX ((1 << LRU_BITS) - 1) /* Max value of obj->lru */
 #define LRU_CLOCK_RESOLUTION 1000           /* LRU clock resolution in ms */
 
-#define OBJ_SHARED_REFCOUNT INT_MAX       /* Global object never destroyed. */
-#define OBJ_STATIC_REFCOUNT (INT_MAX - 1) /* Object allocated in the stack. */
+#define OBJ_SHARED_REFCOUNT ((1 << 30) - 1) /* Global object never destroyed. */
```
rainsupreme: Maybe add a REFCOUNT_BITS constant similar to LRU_BITS?

src/evict.c
```diff
@@ -143,26 +143,14 @@ void evictionPoolAlloc(void) {
...
-        de = samples[j];
-        key = dictGetKey(de);
-
-        /* If the dictionary we are sampling from is not the main
-         * dictionary (but the expires one) we need to lookup the key
-         * again in the key dictionary to obtain the value object. */
-        if (server.maxmemory_policy != MAXMEMORY_VOLATILE_TTL) {
-            if (samplekvs != db->keys) de = kvstoreDictFind(db->keys, slot, key);
-            o = dictGetVal(de);
-        }
```
rainsupreme, line 165: Why can we get rid of this?

src/object.c
```diff
@@ -135,6 +150,148 @@ robj *createStringObject(const char *ptr, size_t len) {
...
+    if (val->hasexpire) {
+        /* Skip expire field */
+        data += sizeof(long long);
+    }
+    if (val->hasembkeyptr) {
+        return *(sds *)data;
+    }
+    if (val->hasembkey) {
+        uint8_t hdr_size = *(uint8_t *)data;
+        data += 1 + hdr_size;
```
rainsupreme, line 251: what if we made `hdr_size` into `key_offset` and have the +1 included in the value?

src/object.c
```diff
@@ -135,6 +150,148 @@ robj *createStringObject(const char *ptr, size_t len) {
...
+    assert(val->hasembkeyptr && !val->hasembkey && valkeyGetKey(val) == NULL);
+
+    /* Find the correct location in val's data field. */
+    unsigned char *data = (void *)(val + 1);
+    if (val->hasexpire) {
+        /* Skip expire field */
+        data += sizeof(long long);
+    }
+    sds oldkey = *(sds *)data;
+    if (oldkey != NULL) sdsfree(oldkey);
```
rainsupreme: This doesn't quite make sense to me with the assert at the top of the function. If `valkeyGetKey(val) == NULL` then there should be nothing here - not embedded or a pointer.

src/object.c
```diff
@@ -135,6 +150,148 @@ robj *createStringObject(const char *ptr, size_t len) {
...
+        /* Calculate sizes */
+        size_t key_sds_size = sdscopytobuffer(NULL, 0, key, NULL);
+        size_t val_len = sdslen(val->ptr);
+
+        size_t min_size = sizeof(robj);
+        min_size += sizeof(long long); /* expire */
+        /* Size of embedded key, incl. 1 byte for prefixed sds hdr size. */
+        min_size += 1 + key_sds_size;
+        /* Size of embedded value (EMBSTR) including \0 term. */
+        min_size += sizeof(struct sdshdr8) + val_len + 1;
```
rainsupreme: Why not call sdscopytobuffer() again? How do we know it always uses sdshdr8?

src/object.c
```diff
@@ -135,6 +150,148 @@ robj *createStringObject(const char *ptr, size_t len) {
...
+
+        /* Set the expire field. */
+        long long expire = -1;
+        *(long long *)data = expire;
+        //memcpy(data, &expire, sizeof(long long));
+        data += sizeof(long long);
+
+        /* Copy embedded string. */
+        sdscopytobuffer(data + 1, key_sds_size, key, data);
+        data += 1 + key_sds_size;
```
rainsupreme: What's the +1? Is that for the sds header size? Is there a reason it's not assigned?

src/kvstore.h
zuiderkwast: This PR currently targets the `hashset` feature branch, where this file exists.

Do you think I should target the unstable branch instead? I'm trying to find a way to split it into smaller PRs...

src/object.c
```diff
@@ -135,6 +150,148 @@ robj *createStringObject(const char *ptr, size_t len) {
...
+
+        /* Copy embedded value (EMBSTR). */
+        struct sdshdr8 *sh = (void *)data;
+        sh->flags = SDS_TYPE_8;
+        sh->len = val_len;
+        size_t capacity = bufsize - (min_size - val_len);
```
rainsupreme: This is correct but I found it rather confusing.

src/kvstore.h
zuiderkwast: How about we first create a PR to unstable with just the hashset implementation and replace dict with hashet in the command lookup table?

madolson: Oh, I missed this, I thought this was the PR for the hashset and I was interested. You can ignore me :D

zuiderkwast: No, seriously, I think it can be very hard to get a huge feature branch merged. We should split it up in some way.

The implementation of the hashset is in the hashset branch in this repo. PTAL!

src/memory_prefetch.c
rainsupreme, line 1: Think we'll implement something, or will we end up completely removing prefetch?

src/defrag.c
```diff
@@ -1054,12 +1036,12 @@ void activeDefragCycle(void) {
...
         typedef struct defragStage {
             kvstore *kvs;
-            dictScanFunction *scanfn;
+            hashsetScanFunction scanfn;
             void *privdata;
         } defragStage;
         defragStage defrag_stages[] = {
             {db->keys, defragScanCallback, db},
-            {db->expires, scanCallbackCountScanned, NULL},
+            //{db->expires, scanCallbackCountScanned, NULL},
```
rainsupreme: ?

rainsupreme (commented):
Didn't look at db.c yet

src/server.h
zuiderkwast: Haha, yes. It explains why the memory footprint is not as good as it should be.

I believe we can get rid of `hasembkeyptr` though and always embed the actual key. Working on that.

src/evict.c
zuiderkwast: The expire hashset points to the actual robj, the same as the keys hashset. There's no need to look it up again.

src/object.c
zuiderkwast: EMBSTR was always hard-coded to using sdshdr8. This is old code, refactored. I wanted to avoid rewriting it.

In the future, a good thing would be to get rid of the `ptr` field in robj and instead compute it using a function. Then we can save another 8 bytes for embedded strings.

src/object.c
zuiderkwast: So the value we store in `key_offset` byte includes its own size?

Yeah, that would work. I copied this code from how keys are currently embedded in dict entry.

src/object.c
zuiderkwast: The +1 is the byte where we store the sds header size.

Assigned? Not sure what you mean.

src/object.c
zuiderkwast: Yes. It's old code, just moved/copied around. You're welcome to suggest simplification, or do it as a follow-up. I wanted to focus on getting things to work first and to avoid changing more than necessary to keep the diff less huge.

src/memory_prefetch.c
zuiderkwast: Yes... Help wanted. :)

src/defrag.c
zuiderkwast: DELETEME

src/object.c
zuiderkwast: You're right. This code isn't ready yet. I want to rewrite it to avoid code duplication and improve it in general.

src/defrag.c
zuiderkwast: When `db->keys` is defragged using `defragKey`, the pointer to the robj in expire is updated too. There's no reason to scan the defrag table. We defrag the hashset allocations though, using `kvstoreHashsetDefragInternals`.

zuiderkwast: I'm deleting this commented out line. Not committed yet though, but I'll push it later.

src/object.c
rainsupreme: I mean, in this function where we allocate and initialize a brand new valkey object, when do we initialize this embedded header size value? What value is it going to have when this function ends?

src/object.c
zuiderkwast: Do you have a suggestion for a code comment or different variable names?

src/object.c
zuiderkwast: The byte at `data` is assigned by `sdscopytobuffer` by the last parameter we pass to it.

src/hashset.c
```diff
@@ -0,0 +1,1705 @@
...
+    samples.size = count;
+    samples.count = 0;
+    samples.elements = dst;
+    size_t cursor = randomSizeT();
+    while (samples.count < count) {
+        cursor = hashsetScan(t, cursor, sampleElementsScanFn, &samples, HASHSET_SCAN_SINGLE_STEP);
+    }
+    rehashStepOnReadIfNeeded(t);
+    return count;
+}
```
rainsupreme: Optimization: Select the random index into `samples` before sampling instead of after. Then just count up elements without copying in `sampleElementsScanFn()` and end the loop early when the random index is reached, and return that element. Space complexity goes from O(count) -> O(1) and can be stack allocated, and average time complexity is 2x faster.

I can write a PR for this. :) The logic is equivalent but maybe the fairness would be less intuitive.

zuiderkwast: Good idea. PR targeting the hashset branch please.

zuiderkwast: But I believe sample elements is used separately for eviction, via kvstoreHashsetSampleElements, so we still need the sample elements function too.

It's a balance between complexity and efficiency. For hashset random, the array is stack allocated and copying just the pointers isn't much overhead since the stack memory is already in CPU cache. So I'm not sure if it's worth it or not. See also Madelyn's comment on the hashset PR for hashset non-fair random.

src/object.c
zuiderkwast: The way sdscopytobuffer is implemented, returning the size of the actual sds representation and not caring about where the byte storing its size is located, I prefer to keep this as is. Possibly we could change the usage of sdscopytobuffer if you can come up with a better idea.

hpatro:
This is great progress. Will take a look this week.

tests/unit/aofrw.tcl
```diff
@@ -17,10 +17,12 @@ start_server {tags {"aofrw external:skip logreqres:skip"} overrides {save {}}} {
...
-            set load_handle1 [start_write_load $master_host $master_port 10]
-            set load_handle2 [start_write_load $master_host $master_port 10]
-            set load_handle3 [start_write_load $master_host $master_port 10]
-            set load_handle4 [start_write_load $master_host $master_port 10]
+            # FIXME: Temporary disabling some load to prevent this test case
+            # from hanging indefinitely. This needs to be investivaged.
+            #set load_handle1 [start_write_load $master_host $master_port 10]
+            #set load_handle2 [start_write_load $master_host $master_port 10]
+            #set load_handle3 [start_write_load $master_host $master_port 10]
+            #set load_handle4 [start_write_load $master_host $master_port 10]
```
zuiderkwast: Without commenting out these lines, this test case is hundreds of times slower than on unstable.

The test case sends random SET commands using multiple write load workers, creating millions of keys and performs AOF rewrite in the background. Then it reloads the AOF using `DEBUG LOADAOF`. It's the loading of AOF without preamble that takes time, i.e. loading by calling the `setCommand()` function. Each SET command takes ~300 microseconds (100 for lookupKey and 200 for setKey), compared to ~1 microsecond before the DEBUG RELOAD and on unstable.

When debugging, I discovered very long probing chains, like length 6000 or so here, possibly spanning all non-empty buckets in the table.

AOF rewrite creates the AOF by iterating the keys hashset. The iterator walks in cursor/probing order. When these keys are loaded in that same order, the first keys all hash to the same bucket and create long probing chains. I tried changing the iterator to just increment normally (index++). It gets slightly better but I still see long probing chains. (Not exactly sure why yet.)

The reason we don't see this slowdown in the preamble case is RDB contains a size and we expand the hash table in advance when loading an RDB.

Anyway, I believe @madolson's idea from https://github.com/valkey-io/valkey/pull/1186#pullrequestreview-2378600726 can solve this:

> instead of using the `everfull` bit to start probing, we could use it to indicate the last position is a pointer to another bucket and do linked list resolution. This does a couple of things:
>
> 1. Prevents significant chain lengths, since collisions won't cause issues.

zuiderkwast: Solved by using bucket-chaining instead of open addressing.

src/memory_prefetch.c
zuiderkwast: @touitou-dan

The prefetch implementation for dict is accessing dict internals to do incremental lookup, one memory access at a time. It's breaking the abstraction and adds duplicated logic. I don't like that part.

For hashset, I have two ideas about how to fetch multiple keys in parallel:

1. A function that looks up multiple keys in parallel. It can return a bitmap with a bit for each key, whether it was found or not.

   ```C
   unsigned hashsetBatchFind(int count, hashset **sets, const void **keys, void **found);
   ```

2. A function that incrementally looks up one key. It needs to be called repeatedly until it's done. The caller provides an (opaque) state that the implementation can use.

   ```C
   int hashsetIncrementalFind(hashset *s, const void *key, void **found, hashsetIncrementalFindState *state);
   ```

   The prefetching code can run multiple such lookups in round-robin.

zuiderkwast:
Included in #1186.

PR #1179, opened 2024-10-16
Fix false sharing issue between main thread and io-threads when access `used_memory_thread`.

zhulipeng:
## Description
When profiling some workloads with `io-threads` enabled. We found the false sharing issue is heavy(**Fig 2**). 

This patch try to split the the elements accessed by main thread and io-threads into different cache line by padding the elements in the head of `used_memory_thread_padded` array. The implementation should be like **Fig 1** demoed. 

Design like below will help mitigate the false sharing between main thread and io-threads, because the main thread has been the bottleneck with io-threads enabled.  We didn't put each element in an individual cache line is that we don't want to bring the additional cache line fetch operation (3 vs 16 cache line) when call function like `zmalloc_used_memory()`.

image
Fig 1. `used_memory_thread_padded` element distribution in cache line.
 
image

Fig 2. perf c2c report with io-threads enabled.

## Perf Boost
With benchmark of **`XADD`**
 `taskset -c 40-55 ~/valkey/src/valkey-benchmark -h 127.0.0.1 -p 9001 -n 10000000 -c 100 --threads 16 -t xadd`.
We can observe around **2%-4%** perf boost.

**Notes:** This optimization should benefit more if the high percentage of cycles of `zmalloc_usable` and `zfree`. 

#### Test Env
- OS: Ubuntu 22.04.4 LTS
- Platform: Intel Xeon Platinum 8380

```
taskset -c 0-15 ~/valkey/src/valkey-server ~/tmp_valkey.conf
port 9001
bind * -::*
daemonize no
protected-mode no
save
io-threads 16
```

### DIFF
--- a/src/zmalloc.c
+++ b/src/zmalloc.c
@@ -102,7 +102,7 @@ static thread_local int thread_index = -1;
  * For the other architecture, lets fall back to the atomic operation to keep safe. */
 #if defined(__i386__) || defined(__x86_64__) || defined(__amd64__) || defined(__POWERPC__) || defined(__arm__) || \
     defined(__arm64__)
-static __attribute__((aligned(CACHE_LINE_SIZE))) size_t __used_memory_thread[MAX_THREADS_NUM + PADDING_ELEMENT_NUM];
+static __attribute__((aligned(CACHE_LINE_SIZE))) size_t used_memory_thread_padded[MAX_THREADS_NUM + PADDING_ELEMENT_NUM];
 #else
 static __attribute__((aligned(CACHE_LINE_SIZE))) _Atomic size_t __used_memory_thread[MAX_THREADS_NUM + PADDING_ELEMENT_NUM];
 #endif
--- a/src/zmalloc.c
+++ b/src/zmalloc.c
@@ -90,6 +90,7 @@ void zlibc_free(void *ptr) {
 
 #define thread_local _Thread_local
 
+#define PADDING_ELEMENT_NUM CACHE_LINE_SIZE / sizeof(size_t) - 1
 #define MAX_THREADS_NUM (IO_THREADS_MAX_NUM + 3 + 1)
 /* A thread-local storage which keep the current thread's index in the used_memory_thread array. */
 static thread_local int thread_index = -1;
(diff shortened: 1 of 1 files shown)


src/zmalloc.c
```diff
@@ -101,10 +102,11 @@ static thread_local int thread_index = -1;
  * For the other architecture, lets fall back to the atomic operation to keep safe. */
 #if defined(__i386__) || defined(__x86_64__) || defined(__amd64__) || defined(__POWERPC__) || defined(__arm__) || \
     defined(__arm64__)
-static __attribute__((aligned(sizeof(size_t)))) size_t used_memory_thread[MAX_THREADS_NUM];
+static __attribute__((aligned(CACHE_LINE_SIZE))) size_t __used_memory_thread[MAX_THREADS_NUM + PADDING_ELEMENT_NUM];
```
PingXie: This padding seems wrong to me. Shouldn't we be padding the element instead? something like the below

```c
typedef struct {
    size_t value;
    char padding[CACHE_LINE_SIZE - sizeof(size_t)];  // Padding to ensure each element starts on a new cache line
} PaddedElement;

static __attribute__((aligned(CACHE_LINE_SIZE))) PaddedElement used_memory_thread[MAX_THREADS_NUM];
```

zuiderkwast: Yeah, or align each element. Can we add the aligned attribute to the element struct?

Or does the aligned attribute on the array declaration actually apply to each element in the array?

zhulipeng: @PingXie @zuiderkwast Thanks for you proposal, I updated the top comment to add more explanations. 
The reason we didn't take the **struct padding** way is the performance degradation. Because when call function like `zmalloc_used_memory()` (called in main thread mostly), the main thread will cost more cycles to fetch more cache lines (3 vs 16 cache lines), per our benchmarking, it will have a **~3%** performance degradation depends on the specified benchmark. 
Now we did some tradeoff, just fix the false sharing between main thread and io-threads, keep the false sharing in the io-threads for memory aggregation, because resource of main thread is bottleneck but resource of io-threads can be scaled.

PingXie: Thanks for the diagram, @lipzhu. That helps a lot. I agree this is a very reasonable trade-off. 

On a side note, I notice that the name of the new "private" array is prefixed with "__". I am fine with it but not sure if @zuiderkwast has any opinions :)

PingXie (approved):
LGTM. Thanks @lipzhu

src/zmalloc.c
zuiderkwast: @lipzhu thank for the explanation.

@PingXie Hehe, yeah, if we want to follow the C standard, any identifier starting with double underscore is reserved. The underscores are not descriptive either. I suggest add some descriptive word to the variable name, like `used_memory_thread_padded`.

zuiderkwast:
```suggestion
static __attribute__((aligned(CACHE_LINE_SIZE))) size_t used_memory_thread_padded[MAX_THREADS_NUM + PADDING_ELEMENT_NUM];
```

zhulipeng: Thanks, changed.

src/zmalloc.c
```diff
@@ -90,6 +90,7 @@ void zlibc_free(void *ptr) {
 
 #define thread_local _Thread_local
 
+#define PADDING_ELEMENT_NUM CACHE_LINE_SIZE / sizeof(size_t) - 1
```
zuiderkwast: Nit: Parentheses is common practice for macro.

```suggestion
#define PADDING_ELEMENT_NUM (CACHE_LINE_SIZE / sizeof(size_t) - 1)
```

zuiderkwast (approved):

PR #1180, opened 2024-10-17
Skip checksum for diskless replication

### LINKED ISSUE #1129: Skip CRC64 checksumming when doing diskless replication

Shivshankar-Reddy:
fixes: #1129 
Below is the background information from #1129 by @madolson 

> When we are not writing to disk (either on the primary or replica) there isn't a lot of sense in doing CRC64 checksumming during fullsync. We do checking summing at the TCP layer already. At AWS we've seen it can take up to 15% overhead for not much of a benefit. We should look into a mechanism so that a replica can indicate that it will load something directly into memory, since today we only indicate we support the EOF marker. The replica might still choose to save the RDB file to disk unless repl-diskless-load is enabled. With this mechanism in place, we should be able to skip the checksum for replication, while still using it for RDB.

Here is my understanding and proposal as a fix for the issue. Please correct me if my understanding is wrong or missing something!

image

- Added 'replication-diskless-load' to sendCommand replica details in syncWithPrimary and FullSyncWithPrimary.
- Added rdbFlag to skip checksum based on 'replica-diskless-load' configuaration from the replica.

### DIFF
--- a/src/rdb.c
+++ b/src/rdb.c
@@ -1407,7 +1407,7 @@ int rdbSaveRio(int req, rio *rdb, int *error, int rdbflags, rdbSaveInfo *rsi) {
     long key_counter = 0;
     int j;
 
-    if (server.rdb_checksum) rdb->update_cksum = rioGenericUpdateChecksum;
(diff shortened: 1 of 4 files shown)


Shivshankar-Reddy:
Abandoning this due to cyclic commit conflict and created https://github.com/valkey-io/valkey/pull/1181 for the same.

PR #1181, opened 2024-10-17
[WIP] Skip the checksum for diskless replication

### LINKED ISSUE #1129: Skip CRC64 checksumming when doing diskless replication

Shivshankar-Reddy:
fixes: #1129 
Below is the background information from #1129 by @madolson 

> When we are not writing to disk (either on the primary or replica) there isn't a lot of sense in doing CRC64 checksumming during fullsync. We do checking summing at the TCP layer already. At AWS we've seen it can take up to 15% overhead for not much of a benefit. We should look into a mechanism so that a replica can indicate that it will load something directly into memory, since today we only indicate we support the EOF marker. The replica might still choose to save the RDB file to disk unless repl-diskless-load is enabled. With this mechanism in place, we should be able to skip the checksum for replication, while still using it for RDB.

Here is my understanding and proposal as a fix for the issue. Please correct me if my understanding is wrong or missing something!

image

- Added 'replication-diskless-load' to sendCommand replica details in syncWithPrimary and FullSyncWithPrimary.
- Added rdbFlag to skip checksum based on 'replica-diskless-load' configuaration from the replica.

### DIFF
--- a/src/rdb.c
+++ b/src/rdb.c
@@ -1407,7 +1407,8 @@ int rdbSaveRio(int req, rio *rdb, int *error, int rdbflags, rdbSaveInfo *rsi) {
     long key_counter = 0;
     int j;
 
-    if (server.rdb_checksum) rdb->update_cksum = rioGenericUpdateChecksum;
+    if (server.rdb_checksum && !(rdbflags & RDBFLAGS_CKSUM_SKIP))
+        rdb->update_cksum = rioGenericUpdateChecksum;
     snprintf(magic, sizeof(magic), "REDIS%04d", RDB_VERSION);
     if (rdbWriteRaw(rdb, magic, 9) == -1) goto werr;
     if (rdbSaveInfoAuxFields(rdb, rdbflags, rsi) == -1) goto werr;
@@ -1451,13 +1452,16 @@ werr:
  * without doing any processing of the content. */
 int rdbSaveRioWithEOFMark(int req, rio *rdb, int *error, rdbSaveInfo *rsi) {
     char eofmark[RDB_EOF_MARK_SIZE];
-
+    int skip_cksum_repl = RDBFLAGS_REPLICATION;
     startSaving(RDBFLAGS_REPLICATION);
     getRandomHexChars(eofmark, RDB_EOF_MARK_SIZE);
     if (error) *error = 0;
     if (rioWrite(rdb, "$EOF:", 5) == 0) goto werr;
     if (rioWrite(rdb, eofmark, RDB_EOF_MARK_SIZE) == 0) goto werr;
     if (rioWrite(rdb, "\r\n", 2) == 0) goto werr;
+    if (server.repl_diskless_sync && req & REPLICA_REQ_CHKSUM_SKIP)
+        skip_cksum_repl |= RDBFLAGS_CKSUM_SKIP;
+    if (rdbSaveRio(req, rdb, error, skip_cksum_repl, rsi) == C_ERR) goto werr;
     if (rdbSaveRio(req, rdb, error, RDBFLAGS_REPLICATION, rsi) == C_ERR) goto werr;
     if (rioWrite(rdb, eofmark, RDB_EOF_MARK_SIZE) == 0) goto werr;
     stopSaving(1);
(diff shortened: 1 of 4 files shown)


hpatro (commented):
Do we avoid any kind of memory corruption by performing a checksum match? If CRC64 is expensive, we could look at other alternatives.

src/rdb.c
```diff
@@ -1451,13 +1452,16 @@ int rdbSaveRio(int req, rio *rdb, int *error, int rdbflags, rdbSaveInfo *rsi) {
...
+    int skip_cksum_repl = RDBFLAGS_REPLICATION;
     startSaving(RDBFLAGS_REPLICATION);
     getRandomHexChars(eofmark, RDB_EOF_MARK_SIZE);
     if (error) *error = 0;
     if (rioWrite(rdb, "$EOF:", 5) == 0) goto werr;
     if (rioWrite(rdb, eofmark, RDB_EOF_MARK_SIZE) == 0) goto werr;
     if (rioWrite(rdb, "\r\n", 2) == 0) goto werr;
+    if (server.repl_diskless_sync && req & REPLICA_REQ_CHKSUM_SKIP)
+        skip_cksum_repl |= RDBFLAGS_CKSUM_SKIP;
+    if (rdbSaveRio(req, rdb, error, skip_cksum_repl, rsi) == C_ERR) goto werr;
```
ranshid, line 1464: we should probably keep only one rdbSaveRio :)

src/replication.c
```diff
@@ -1302,6 +1305,13 @@ void replconfCommand(client *c) {
                                     sdslen(addr));
                 return;
             }
+        } else if (!strcasecmp(c->argv[j]->ptr, "repl-diskless-load")) {
+            /* REPLCONF repl-diskless-load is used to identify the client is capable of
+             * load directly without creating rdb file */
+            long rdb_diskless_load = 0;
+            if (getRangeLongFromObjectOrReply(c, c->argv[j + 1], 0, 1, &rdb_diskless_load, NULL) != C_OK) return;
+            if (rdb_diskless_load == 1)
+                c->replica_req |= REPLICA_REQ_CHKSUM_SKIP;
```
ranshid, line 1314: I think at this point the skip CRC is more "capability" than "requirement" on the replica side. So I would suggest adding a replconf capa indication instead.

ranshid (commented):
@Shivshankar-Reddy Sorry had some urgent issues and holidays so was unable to look at it yet.
Overall I like this added ability!

General question I have: I noticed that you only disabled CRC calculation on the primary side? Basically both primary and replica are performing the CRC calculation. When the replica is completely aware that the data is sent and loaded diskless it can just skip CRC check even when the primary did calculate and report it (for example a replica syncing with an older version primary). the replica can probably know that the Primary is streaming via socket by the fact that it send the EOF marker.

src/rdb.c
Shivshankar-Reddy: Thank you for noticing this. I am aware of this and removed in my local branch, this has had happened because I had merge conflict in the branch so created this by copying the code from original branch. it was missed during the copy paste

src/replication.c
Shivshankar-Reddy: Initially my plan also was to add in capabilities, However the capabilies were not passed many function as requirements being passed. Since requirements are so flexible in the available functions I added in the requirements. But I agree with you this should be part of the capa. PR is still in-progress I will try to make this in Capabilities instead of requirements. Thank you!

Shivshankar-Reddy:
Thank you for the review and the input!

> General question I have: I noticed that you only disabled CRC calculation on the primary side? Basically both primary and replica are performing the CRC calculation. When the replica is completely aware that the data is sent and loaded diskless it can just skip CRC check even when the primary did calculate and report it (for example a replica syncing with an older version primary). the replica can probably know that the Primary is streaming via socket by the fact that it send the EOF marker.

In my initial testing, found that replica was able to deteremine CRC is enabled/disabled so I thought it would be enough to disable at primary end. and I think you have the valid point to consider replica side aswell. I will look into this and will update the diff soon, thanks again!

ranshid:
@Shivshankar-Reddy thank you so much for taking the time to handle this. I think that since it was in conflict with https://github.com/valkey-io/valkey/pull/1479 (which is almost completed) we will close this one.

PR #1182, opened 2024-10-17
Fix SORT GET to ignore special pattern # in cluster slot check

enjoy-binbin:
This special pattern '#' is used to get the element itself,
it does not actually participate in the slot check.

In this case, passing `GET #` will cause '#' to participate
in the slot check, causing the command to get an
`pattern may be in different slots` error.

### DIFF
--- a/src/sort.c
+++ b/src/sort.c
@@ -44,7 +44,7 @@ serverSortOperation *createSortOperation(int type, robj *pattern) {
 }
 
 /* Return 1 if pattern is the special pattern '#'. */
-int isReturnSubstPattern(sds pattern) {
+static int isReturnSubstPattern(sds pattern) {
     return pattern[0] == '#' && pattern[1] == '\0';
 }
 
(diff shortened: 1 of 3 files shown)


src/sort.c
```diff
@@ -43,6 +43,11 @@ serverSortOperation *createSortOperation(int type, robj *pattern) {
     return so;
 }
 
+/* Return 1 if pattern is the special pattern '#'. */
+int isReturnSubstPattern(sds pattern) {
```
zuiderkwast:
```suggestion
static int isReturnSubstPattern(sds pattern) {
```

zuiderkwast (approved):
Was this a bug (wrong result) or just an optimization problem?

We should include in release notes and backport if it was a bugfix.

enjoy-binbin:
it was a bug, it (see the test example) will get an error since we will do a slot check with '#', sorry i forget to mention it, i updated the top comment
```
                addReplyError(c, "GET option of SORT denied in Cluster mode when "
                                 "keys formed by the pattern may be in different slots.");
                syntax_error++;
```

madolson (approved):

PR #1183, opened 2024-10-17
Adds wasmtime-c-api dependency to support WASM scripts/functions

rjd15372:
This commit introduces a new dependency on the `wasmtime-c-api` library.

https://github.com/bytecodealliance/wasmtime

Wasmtime is a standalone runtime for WebAssembly (aka WASM), and we will use it to run WASM binaries inside the Valkey engine.

To enable WASM support in Valkey we need to build Valkey with:

```
make BUILD_WITH_WASM=yes
```

If you have your own `wasmtime-c-api` local build, you can specify its location so that our build system can pick it up:

```
make BUILD_WITH_WASM=yes WASMTIME_ROOT=path
```

If you just want to use the upstream `wasmtime-c-api` binary, you can run:

```
make download-wasm-binary-deps
```

before building valkey to download the upstream binary.

### DIFF
--- a/deps/Makefile
+++ b/deps/Makefile
@@ -74,6 +74,12 @@ fpconv: .make-prerequisites
 
 .PHONY: fpconv
 
+download-wasm-binary-deps: .make-prerequisites
+	@printf '%b %b\n' $(MAKECOLOR)MAKE$(ENDCOLOR) $(BINCOLOR)$@$(ENDCOLOR)
+	cd wasmtime-c-api && $(MAKE)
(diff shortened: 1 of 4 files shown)


zuiderkwast (commented):
This looks like a good start. :+1: Let's review more when there's is a working implementation to review.

rjd15372:
We don't need to add this dependency anymore, since we are almost at the point where we can implement a new scripting engine as an external Valkey Module.

PR #1184, opened 2024-10-17
Clear outdated failure reports more accurately

enjoy-binbin:
There are two changes here:

1. The one in clusterNodeCleanupFailureReports, only primary with slots can
report the failure report, if the primary became a replica its failure report
should be cleared. This may lead to inaccurate node fail judgment in some network
partition cases i guess, it will also affect the CLUSTER COUNT-FAILURE-REPORTS
command.

2. The one in clusterProcessGossipSection, it is not that important, but it can
print a "node is back online" log helps us troubleshoot the problem, although
it may conflict with 1 at some points.


src/cluster_legacy.c
```diff
@@ -2246,9 +2257,9 @@ void clusterProcessGossipSection(clusterMsg *hdr, clusterLink *link) {
         if (node && node != myself) {
             /* We already know this node.
                Handle failure reports, only when the sender is a voting primary. */
```
zarkash-aws: This comment needs to be updated - "only when the sender is a voting primary" is not longer accurate since recovery reports can be processed from any sender.

enjoy-binbin: yes, the comment is outdated, i updated it, thanks

zuiderkwast (approved):
Looks correct to me.

src/cluster_legacy.c
```diff
@@ -1575,6 +1584,8 @@ void clusterNodeCleanupFailureReports(clusterNode *node) {
  * Otherwise 0 is returned. */
 int clusterNodeDelFailureReport(clusterNode *node, clusterNode *sender) {
     list *l = node->fail_reports;
+    if (!listLength(l)) return 0;
```
PingXie, line 1587: nit - this is optional actually but I am fine with having this explicit check - maybe help with cache hit rate since this is the super common case.

src/cluster_legacy.c
```diff
@@ -1558,7 +1563,11 @@ void clusterNodeCleanupFailureReports(clusterNode *node) {
     listRewind(l, &li);
     while ((ln = listNext(&li)) != NULL) {
         fr = ln->value;
-        if (now - fr->time > maxtime) listDelNode(l, ln);
+        if (now - fr->time > maxtime) {
+            listDelNode(l, ln);
+        } else if (!clusterNodeIsVotingPrimary(fr->node)) {
+            listDelNode(l, ln);
```
PingXie, line 1569: the change in `clusterProcessGossipSection` makes sense but I wonder if this change will delay the failover in the zonal outage case. We'd lose 1/3 of failure reports and it would take at least `server.cluster_node_timeout` to regain these reports. thoughts?

enjoy-binbin: yes, maybe i guess, i fail to imagine this, could you give an example?

now we only call clusterNodeCleanupFailureReports in two places, clusterNodeDelFailureReport and clusterNodeFailureReportsCount, so you are saying that in some cases, we might clear the reports before the new reports jump in, so a node that is in a pfail state will wrongly marked with ok?

PingXie: hmm, you are right. actually the delay won't happen. the change here is to remove failure reports that belong to replicas. The new primaries would have taken over by now. this node should've either already known the new primaries or know them very soon.

PingXie (approved):

PR #1185, opened 2024-10-17
Import-mode: Avoid expiration and eviction during data syncing

lyq2333:
New config: `import-mode (yes|no)`

New command: `CLIENT IMPORT-SOURCE (ON|OFF)`

The config, when set to `yes`, disables eviction and deletion of expired keys, except for commands coming from a client which has marked itself as an import-source, the data source when importing data from another node, using the CLIENT IMPORT-SOURCE command.

When we sync data from the source Valkey to the destination Valkey using some sync tools like [redis-shake](https://github.com/tair-opensource/RedisShake), the destination Valkey can perform expiration and eviction, which may cause data corruption. This problem has been discussed in https://github.com/redis/redis/discussions/9760#discussioncomment-1681041 and Redis already have a solution. But in Valkey we haven't fixed it by now.

E.g. we call `set key 1 ex 1` on the source server and transfer this command to the destination server. Then we call `incr key` on the source server before the key expired, we will have a key on the source server with a value of 2. But when the command arrived at the destination server, the key may be expired and has deleted. So we will have a key on the destination server with a value of 1, which is inconsistent with the source server.

In standalone mode, we can use writable replica to simplify the sync process. However, in cluster mode, we still need a sync tool to help us transfer the source data to the destination. The sync tool usually work as a normal client and the destination works as a primary which keep expiration and eviction.

In this PR, we add a new mode named 'import-mode'. In this mode, server stop expiration and eviction just like a replica. Notice that this mode exists only in sync state to avoid data inconsistency caused by expiration and eviction. Import mode only takes effect on the primary. Sync tools can mark their clients as an import source by `CLIENT IMPORT-SOURCE`, which work like a client from primary and can visit expired keys in `lookupkey`.

**Notice: during the migration, other clients, apart from the import source, should not access the data imported by import source.**


soloestoy:
good work! @valkey-io/core-team please take a look

src/replication.c
```diff
@@ -1272,6 +1272,10 @@ void syncCommand(client *c) {
  * - rdb-channel <1|0>
  * Used to identify the client as a replica's rdb connection in an dual channel
  * sync session.
+ *
+ * - pseudo-master <0|1>
+ * Set this connection behaving like a master if server.pseudo_replica is true.
+ * Sync tools can set their connections into 'pseudo-master' state to visit expired keys.
  * */
 void replconfCommand(client *c) {
```
zuiderkwast: The sync tool sends `REPLCONF PSEUDO-MASTER` to say "I'm a pseudo-master, so ignore expire for this connection"?

Almost all other REPLCONF commands are sent by the replica to the primary before doing the PSYNC. This one is different. Can we add a more explicit comment about this difference, similar to `REPLCONF GETACK` which is also different:

```
 * - getack <dummy>
 * Unlike other subcommands, this is used by primary to get the replication
 * offset from a replica.
```

valkey.conf
```diff
@@ -801,6 +801,13 @@ replica-priority 100
 #
 # replica-ignore-disk-write-errors no
 
+# Make the master behave like a replica, which forbids expiration and evcition.
+# This is useful for sync tools, because expiration and evcition may cause the data corruption.
+# Sync tools can set their connections into 'pseudo-master' state by REPLCONF PSEUDO-MASTER to
+# behave like a master(i.e. visit expired keys).
+#
+# pseudo-replica no
+
```
zuiderkwast: I think "pseudo-master" is a little bit confusing. :) I understand how it works now, but only after I read the test cases. Let's try to improve this documentation later.

In Valkey we are no longer not using "master", so new commands and configs should use "primary".

Maybe we can use master if Redis has exactly the same REPLCONF or config, but otherwise let's use primary.

zuiderkwast (commented):
I like it. I think it's a little bit strange, but with some better documentation it's not too strange.

We need to have good support for migration tools, especially for users who want to migrate from proprietary software to open source software. :smile:

> Redis already have a solution.

I remember this discussion from the Redis times. Do you know what Redis solution is (the same `REPLCONF PSEUDO-MASTER`?) or is it secret?

> Any better idea?

Have you considered the idea to let RediShake act as a primary and let the target database replicate from RediShake? It can act as a replication proxy?

```
+-----------+   PSYNC  +-----------+   PSYNC  +--------+
| Source DB |<---------| RediShake |<---------| Valkey |
+-----------+          +-----------+          +--------+
```

valkey.conf
hwware: I agree with you, the new mode server is master node, but it is named as pseudo-replica, it is confused by most people. Let us first give it a better name first.

zuiderkwast: `import-mode`?

Should this node forbid writes from all other clients? It makes it behave even more like a replica.

Btw, I'm thinking now that if we want a better implementation of slot migration, maybe it can use the same or a similar feature. The slot replication is also similar to replication but initiated from the source node. @enjoy-binbin how is the implementation you want to upstream?

soloestoy:
```
+-----------+   PSYNC  +-----------+   PSYNC  +--------+
| Source DB |<---------| RediShake |<---------| Valkey |
+-----------+          +-----------+          +--------+
```

interesting, and we have also considered such a method, but there are several issues.

First, as a cloud provider, we do not allow an instance to become a replica for external instances. This is a very risky operation, especially since the primary connection uses a super user, which has excessive permissions. Additionally, the replica needs to establish an outbound connection, this is also not allowed. These restrictions, I believe, are not unique to cloud providers, many users' security control policies also prohibit such actions.

Another point is that, in a cluster mode, the source and target instances for migration typically have different slot distributions, and redisShake can help with correctly routing the data.

src/evict.c
```diff
@@ -546,8 +546,8 @@ int performEvictions(void) {
         goto update_metrics;
     }
 
-    if (server.maxmemory_policy == MAXMEMORY_NO_EVICTION) {
-        result = EVICT_FAIL; /* We need to free memory, but policy forbids. */
+    if (server.maxmemory_policy == MAXMEMORY_NO_EVICTION || server.pseudo_replica) {
```
soloestoy: it's better to place it in `isSafeToPerformEvictions` together with the `server.primary_host` check.

src/replication.c
```diff
@@ -1418,6 +1422,16 @@ void replconfCommand(client *c) {
                 return;
             }
             c->associated_rdb_client_id = (uint64_t)client_id;
+        } else if (!strcasecmp(c->argv[j]->ptr, "pseudo-master")) {
```
soloestoy: agree with @zuiderkwast , maybe a more explicit command is better, like `CLIENT PSEUDO-MASTER <on|off>`.

BTW, we should also refuse the command if server is not in `pseudo-replica` mode?

soloestoy (commented):
Moreover, I think we should only allow write commands for `pseudo-master` client when server is in `pseudo-replica` mode.

lyq2333:
> > Redis already have a solution.
> 
> I remember this discussion from the Redis times. Do you know what Redis solution is (the same `REPLCONF PSEUDO-MASTER`?) or is it secret?

It's secret. I used to push this PR to Redis Community(#13077) and they say they will PR their implementation. But no news after that.

> > Any better idea?
> 
> Have you considered the idea to let RediShake act as a primary and let the target database replicate from RediShake? It can act as a replication proxy?
> 
> ```
> +-----------+   PSYNC  +-----------+   PSYNC  +--------+
> | Source DB |<---------| RediShake |<---------| Valkey |
> +-----------+          +-----------+          +--------+
> ```

What @soloestoy said is a major limitation for us. BTW, sometimes the destination already has some data,  `PSYNC` will delete them.

valkey.conf
lyq2333: Because clients may sync data to a working server, I think this is not very friendly to the 24/7 businesses if we simply forbid writes from all other clients. But if we allow writes from other clients, we may face the same dual-write problem as writable replica, I'll document it in the `valkey.conf`.

enjoy-binbin:
i did not read it carefully, internally we simply pause the expiration on both side i guess.
Redis has an issue that seems to be discussing this: https://github.com/redis/redis/issues/13478

valkey.conf
enjoy-binbin: Yes, the slot replication is also similar the replication in the implementation, it is something like slot RDB + slot replication propagate + slot failover, something like that

src/evict.c
lyq2333: If place it in `isSafeToPerformEvictions `, the server will ignore maxmemory. I think it's better to return an OOM error to stop the syncing process.

src/replication.c
lyq2333: > The sync tool sends `REPLCONF PSEUDO-MASTER` to say "I'm a pseudo-master, so ignore expire for this connection"?

Yes. It does look a bit strange. I add a new command `CLIENT IMPORT-SOURCE <on|off>` like what @soloestoy say. Maybe it'll look better this way?

lyq2333:
> i did not read it carefully, internally we simply pause the expiration on both side i guess. Redis has an issue that seems to be discussing this: [redis/redis#13478](https://github.com/redis/redis/issues/13478)

Interesting. I think this PR can solve that problem if the server has no other read. If the server has other read, one possible solution is to disable all expiration in `lookupKey` for all clients. But I don't want this PR to affect read/write from normal client.

src/networking.c
```diff
@@ -3567,6 +3567,9 @@ void clientCommand(client *c) {
             "    Protect current client connection from eviction.",
             "NO-TOUCH (ON|OFF)",
             "    Will not touch LRU/LFU stats when this mode is on.",
+            "PSEUDO-PRIMARY (ON|OFF)",
+            "    Set this connection behaving like a primary if server.import_mode is true.",
+            "    Sync tools can set their connections into 'pseudo-primary' state to visit expired keys.",
```
zuiderkwast:
```suggestion
            "IMPORT-SOURCE (ON|OFF)",
            "    Mark this connection as an import source if server.import_mode is true.",
            "    Sync tools can set their connections into 'import-source' state to visit",
            "    expired keys.",
```

src/server.h
```diff
@@ -2071,8 +2071,8 @@ struct valkeyServer {
     char primary_replid[CONFIG_RUN_ID_SIZE + 1]; /* Primary PSYNC runid. */
     long long primary_initial_offset;            /* Primary PSYNC offset. */
     int repl_replica_lazy_flush;                 /* Lazy FLUSHALL before loading DB? */
-    /* Pseudo Replica */
-    int pseudo_replica; /* If true, server is a pseudo replica. */
+    /* Import Mode */
+    int import_mode; /* If true, server is in import mode and forbid expiration and evcition. */
```
zuiderkwast:
```suggestion
    int import_mode; /* If true, server is in import mode and forbid expiration and eviction. */
```

valkey.conf
```diff
@@ -801,6 +801,13 @@ replica-priority 100
 #
 # replica-ignore-disk-write-errors no
 
+# Make the primary forbid expiration and evcition.
```
zuiderkwast:
```suggestion
# Make the primary forbid expiration and eviction.
```

valkey.conf
```diff
@@ -801,6 +801,13 @@ replica-priority 100
...
+# This is useful for sync tools, because expiration and evcition may cause the data corruption.
```
zuiderkwast:
```suggestion
# This is useful for sync tools, because expiration and eviction may cause the data corruption.
```

zuiderkwast:
@lyq2333 Can you commit the changes to `commands.def`?

When you run make locally and you have python3 installed, make updates commands.def if there are any new or changed commands.

lyq2333:
> @lyq2333 Can you commit the changes to `commands.def`?
> 
> When you run make locally and you have python3 installed, make updates commands.def if there are any new or changed commands.

@zuiderkwast Thanks. Sorry forgot it before. I also forgot to sign off and have to force push. No code is modified.

madolson:
Weekly core meeting. No specific comments other than we should review this offline and make progress. Directionally seems like a good idea.

lyq2333:
The scenario we encountered is that some users want to migrate data from Server A to Server B. Both A and B work as primary and B may have some data before the migration. We find expiration and eviction may lead to data inconsistency. So we came up with a simple method, but it still has a few places to discuss.

We plan to introduce a new config(import-mode in this PR) to mark that this server is importing data. As for expiration, in import mode, active expiration in cron is prohibited and passive expiration in commands is limited depend on the client state. Clients marked as import source work like `server.primary`, which will not trigger passive expiration for read and write. But there are many ways to handle other normal clients. We think of some ways as follow.
1. Normal clients work as usual, which means they can perform passive expiration whether for read or write. The drawback is that normal clients accessing the migrating data will also trigger passive expiration and may affect the migration.
2. Normal clients are prohibited from write commands, meanwhile don't trigger passive expiration when read and don't read expired keys. The server works like a read-only replica. The drawback is that clients need to stop writing during the data migration and may affect users business.
3. Normal clients don't trigger passive expiration when read and don't read expired keys, but still trigger passive expiration when write. We can't prohibit passive expiration in write commands, because server will crash when we call incr/decr commands on expired keys.  The reason is that these commands must delete expired keys first, otherwise it will hit the assert in `dbAdd`. The drawback is that normal clients calling write commands on the migrating data will trigger passive expiration. This is a trade-off between the above two ways and I think it's better because it won't affect the normal clients and have less impact on the migration process.

As for eviction, should server disable eviction automatically to ensure data consistency in import mode？Or let users choose the maxmeory-policy, I guess no one will choose an option other than `noeviction`.

@valkey-io/core-team WDYT? Any suggestions would be greatly appreciated.

soloestoy:
> The scenario we encountered is that some users want to migrate data from Server A to Server B. Both A and B work as primary and B may have some data before the migration. We find expiration and eviction may lead to data inconsistency. So we came up with a simple method, but it still has a few places to discuss.
> 
> We plan to introduce a new config(import-mode in this PR) to mark that this server is importing data. As for expiration, in import mode, active expiration in cron is prohibited and passive expiration in commands is limited depend on the client state. Clients marked as import source work like `server.primary`, which will not trigger passive expiration for read and write. But there are many ways to handle other normal clients. We think of some ways as follow.
> 
> 1. Normal clients work as usual, which means they can perform passive expiration whether for read or write. The drawback is that normal clients accessing the migrating data will also trigger passive expiration and may affect the migration.
> 2. Normal clients are prohibited from write commands, meanwhile don't trigger passive expiration when read and don't read expired keys. The server works like a read-only replica. The drawback is that clients need to stop writing during the data migration and may affect users business.
> 3. Normal clients don't trigger passive expiration when read and don't read expired keys, but still trigger passive expiration when write. We can't prohibit passive expiration in write commands, because server will crash when we call incr/decr commands on expired keys.  The reason is that these commands must delete expired keys first, otherwise it will hit the assert in `dbAdd`. The drawback is that normal clients calling write commands on the migrating data will trigger passive expiration. This is a trade-off between the above two ways and I think it's better because it won't affect the normal clients and have less impact on the migration process.
> 
> As for eviction, should server disable eviction automatically to ensure data consistency in import mode？Or let users choose the maxmeory-policy, I guess no one will choose an option other than `noeviction`.
> 
> @valkey-io/core-team WDYT? Any suggestions would be greatly appreciated.

The core issue is that for the target node of data migration that is active, we need to provide normal read and write services. However, we don't want to affect the data being migrated and can't identify which data is being migrated.

Method 3 seems to be a relatively suitable option, but of course, it requires users to ensure they do not perform write operations on the data being migrated.

zuiderkwast:
I agree method 3 seems to be the most suitable. The node behaves like when reading from a replica, a writable replica if write commands are used.

PingXie:
What I have in mind, on a very high level, is a server side mutable config that forces a server (the target) to unconditionally accept all writes from any client. This to me is a "passthrough" mode and I think it is quite clean.

The recent discussion on allowing normal clients to perform writes on the target while the data on the source is being migrated to the target is a form of active-active, which is a broader topic.

I am not sure I understand why the target needs to accept user traffic during the migration? Any reasons why the traffic can't be handled by the source until the target catches up and then we cut over the clients to the target? An analogy would be the primary failover process.

lyq2333:
> What I have in mind, on a very high level, is a server side mutable config that forces a server (the target) to unconditionally accept all writes from any client. This to me is a "passthrough" mode and I think it is quite clean.

@PingXie  Sorry, I'm not sure I understand what you mean. Currently, the target accept all writes from any client unconditionally, but this may cause the data inconsistency  like the example in the top comment. So I think we may need to add a config to address this problem.

The target works like a writable replica. So I think in some situations, before the data migration, the target may have already provided services and some clients have connected to it. It's no need to move these clients to the source then move them back to the target. It's very similar to active-active. Even we have no traffic on the target, we also need to stop the active expiration and eviction on the target.

soloestoy:
Another straightforward method is to add a configuration to globally ignore expiration times, such as "ignore-expire", treating all keys as if no expiration time are set. This way, expired keys will not be deleted. This means that any client can read expired data and modify it as well.

PingXie:
> @PingXie Sorry, I'm not sure I understand what you mean. Currently, the target accept all writes from any client unconditionally, but this may cause the data inconsistency like the example in the top comment. So I think we may need to add a config to address this problem.


I understand the rationale and it makes perfect sense to  me.

What I am concerned about is the implementation. Writable replicas are essentially active-active. I do think active-active has value but I think it needs to be introduced consciously/explicitly for clusters. Some of the suggestions above (such as option 2 and 3) are talking, subtly, about a much narrowed form of active-active hence my callout.

PingXie:
> Another straightforward method is to add a configuration to globally ignore expiration times, such as "ignore-expire", treating all keys as if no expiration time are set. This way, expired keys will not be deleted. This means that any client can read expired data and modify it as well.

Maybe a global "expire-delete-enabled" and start with "no", so that we don't break the existing clients; then add a new client capa "ignore-expire", which would skip expire altogether. Finally, after the migration is done, reenable "expire-delete" to start reclaiming the space again?

soloestoy:
We now have a `debug set-active-expire [0|1]` command to enable or disable active expiration, which can be added to the configuration options.

lyq2333:
> Another straightforward method is to add a configuration to globally ignore expiration times, such as "ignore-expire", treating all keys as if no expiration time are set. This way, expired keys will not be deleted. This means that any client can read expired data and modify it as well.

This is a new option. It's simple and the data migration won't be affected by normal clients, but the drawback is that the clients can read expired data.

lyq2333:
> > Another straightforward method is to add a configuration to globally ignore expiration times, such as "ignore-expire", treating all keys as if no expiration time are set. This way, expired keys will not be deleted. This means that any client can read expired data and modify it as well.
> 
> Maybe a global "expire-delete-enabled" and start with "no", so that we don't break the existing clients; then add a new client capa "ignore-expire", which would skip expire altogether. Finally, after the migration is done, reenable "expire-delete" to start reclaiming the space again?

I think this is the same as option 3?

src/commands/client-import-source.json
```diff
@@ -0,0 +1,40 @@
+{
+    "IMPORT-SOURCE": {
+        "summary": "Mark this client as an import source when server is in import mode.",
+        "complexity": "O(1)",
+        "group": "connection",
+        "since": "8.2.0",
```
madolson:
```suggestion
        "since": "8.1.0",
```
?

zuiderkwast (approved):
Looks good to me.

It would be nice to have a documentation page for data migration. We have https://valkey.io/topics/migration/ but I think we should explain how to move data using this feature as well.

Question: What happens if a cluster node is a primary in import mode and there is a failover so another node becomes the primary? The node becomes a replica. Do we automatically exit import-mode?

lyq2333:
> Question: What happens if a cluster node is a primary in import mode and there is a failover so another node becomes the primary? The node becomes a replica. Do we automatically exit import-mode?

@zuiderkwast Thanks. Nice question. You remind me. How about using `(!server.primary_host && server.import_mode)` as the condition? i.e. `if (server.maxmemory_policy == MAXMEMORY_NO_EVICTION || server.import_mode)` change to `if (server.maxmemory_policy == MAXMEMORY_NO_EVICTION || (!server.primary_host && server.import_mode))`.

lyq2333:
Force push because I forgot to sign off again. Just update the commands.def.

lyq2333:
> > Question: What happens if a cluster node is a primary in import mode and there is a failover so another node becomes the primary? The node becomes a replica. Do we automatically exit import-mode?
> 
> @zuiderkwast Thanks. Nice question. You remind me. How about using `(!server.primary_host && server.import_mode)` as the condition? i.e. `if (server.maxmemory_policy == MAXMEMORY_NO_EVICTION || server.import_mode)` change to `if (server.maxmemory_policy == MAXMEMORY_NO_EVICTION || (!server.primary_host && server.import_mode))`.

DONE.

src/db.c
```diff
@@ -1826,6 +1826,9 @@ keyStatus expireIfNeededWithDictIndex(serverDb *db, robj *key, int flags, int di
     if (server.primary_host != NULL) {
         if (server.current_client && (server.current_client->flag.primary)) return KEY_VALID;
         if (!(flags & EXPIRE_FORCE_DELETE_EXPIRED)) return KEY_EXPIRED;
+    } else if (server.import_mode) {
+        if (server.current_client && (server.current_client->flag.import_source)) return KEY_VALID;
+        if (!(flags & EXPIRE_FORCE_DELETE_EXPIRED)) return KEY_EXPIRED;
```
soloestoy, line 1847: It's better to add comments to explain the various behaviors under `import-mode`.

src/expire.c
```diff
@@ -521,7 +521,7 @@ int checkAlreadyExpired(long long when) {
      *
      * Instead we add the already expired key to the database with expire time
      * (possibly in the past) and wait for an explicit DEL from the primary. */
-    return (when <= commandTimeSnapshot() && !server.loading && !server.primary_host);
+    return (when <= commandTimeSnapshot() && !server.loading && !server.primary_host && !server.import_mode);
```
soloestoy, line 527: adding comments would be better : )

soloestoy (approved):

src/db.c
```diff
@@ -1827,6 +1827,22 @@ keyStatus expireIfNeededWithDictIndex(serverDb *db, robj *key, int flags, int di
...
+         * 
+         * Notice: other clients, apart from the import source, should not access 
+         * the data imported by import source.
+         *
+         * Still we try to return the right information to the caller,
+         * that is, KEY_VALID if we think the key should still be valid,
+         * KEY_EXPIRED if we think the key is expired but don't want to delete it at this time.
+         *
+         * When receiving commands from the import source, keys are never considered
+         * expired. */
```
zuiderkwast: These are good comments! I wish that you add this documentation to the doc repo as well, later.

lyq2333: OK, shall I add a section in this doc(https://github.com/valkey-io/valkey-doc/blob/main/topics/migration.md) ?

zuiderkwast: Yes, I think that's a good place to add it. We can change the title of the document to be more broad, if people want to migrate from other redis-like software and services too.

src/server.c
```diff
@@ -1056,7 +1056,9 @@ void databasesCron(void) {
      * as primary will synthesize DELs for us. */
     if (server.active_expire_enabled) {
         if (iAmPrimary()) {
-            activeExpireCycle(ACTIVE_EXPIRE_CYCLE_SLOW);
+            if (!server.import_mode) {
```
ranshid: any reason for a nested if here?

ranshid (commented):
Still did not take a very close look. Overall it seems aligned with the discussions.
I would like to ask: what is the migration strategy regarding ACLs? I mean replica is basically permitted to all actions, but in case of a migration the Source would have to adjust it's user permissions?

src/server.c
zuiderkwast: I guess the reason is the `else` below.

zuiderkwast: It could be rewritten as this though, if that's better?

```C
if (!iAmPrimary()) {
    expireReplicaKeys();
} else if (!server.import_mode) {
    activeExpireCycle(ACTIVE_EXPIRE_CYCLE_SLOW);
}

**Comment by lyq2333** [comment]
> Still did not take a very close look. Overall it seems aligned with the discussions. I would like to ask: what is the migration strategy regarding ACLs? I mean replica is basically permitted to all actions, but in case of a migration the Source would have to adjust it's user permissions?

Yes, I suggest creating a new user for data migration.

File: src/server.c (hunk 13 above)
Comment (lyq2333) [reply src/server.c]: The reason is just as @zuiderkwast said. it's a bit strange and I have fixed it. Thanks @zuiderkwast.

**Comment by ranshid** [comment]
> > Still did not take a very close look. Overall it seems aligned with the discussions. I would like to ask: what is the migration strategy regarding ACLs? I mean replica is basically permitted to all actions, but in case of a migration the Source would have to adjust it's user permissions?
> 
> Yes, I suggest creating a new user for data migration.

So just to circle back and clarify this: the client which migrates the data will NOT automatically granted with superuser permissions but wil have to count on it's authenticated use to support all needed permissions?

**Comment by zuiderkwast** [comment]
> So just to circle back and clarify this: the client which migrates the data will NOT automatically granted with superuser permissions but wil have to count on it's authenticated use to support all needed permissions?

That's right. It would be good to mention this in the docs.

It wouldn't be very secure if a client could get superuser permissions just by sending the command `CLIENT IMPORT-SOURCE ON`.

**Review by ranshid — APPROVED** [review]
LGTM

**Comment by zuiderkwast** [comment]
Looking forward to a doc PR for

* commands/client-import-source.md
* topics/migration.md

In the migration guide, I hope we can have some guide about how to migrate data with import-mode, for example using redishake.

**Comment by lyq2333** [comment]
> Looking forward to a doc PR for
> 
> * commands/client-import-source.md
> * topics/migration.md
> 
> In the migration guide, I hope we can have some guide about how to migrate data with import-mode, for example using redishake.

@suxb201 hi, I think this pr is helpful for RedisShake. PTAL.

File: src/networking.c [hunk 14]
```diff
@@ -4040,6 +4044,22 @@ void clientCommand(client *c) {
...
         addReply(c, shared.ok);
+    } else if (!strcasecmp(c->argv[1]->ptr, "import-source")) {
+        /* CLIENT IMPORT-SOURCE ON|OFF */
+        if (!server.import_mode) {
+            addReplyError(c, "Server is not in import mode");
+            return;
+        }
+        if (!strcasecmp(c->argv[2]->ptr, "on")) {
+            c->flag.import_source = 1;
+            addReply(c, shared.ok);
```
hpatro, line 4055: What happens to client `import_source` flag which were turned on explicitly when server `import_mode` is turned off?

tests/unit/expire.tcl
```diff
@@ -832,6 +832,80 @@ start_server {tags {"expire"}} {
         close_replication_stream $repl
         assert_equal [r debug set-active-expire 1] {OK}
     } {} {needs:debug}
+
+    test {Import mode should forbid active expiration} {
```
hpatro, line 836: It might be worthwhile to run these tests in cluster-enabled mode as well.

src/db.c
```diff
@@ -1826,6 +1826,25 @@ keyStatus expireIfNeededWithDictIndex(serverDb *db, robj *key, int flags, int di
...
+    } else if (server.import_mode) {
+        /* If we are running in the import mode on a primary, instead of
+         * evicting the expired key from the database, we return ASAP:
+         * the key expiration is controlled by the import source that will
+         * send us synthesized DEL operations for expired keys. The
+         * exception is when write operations are performed on this server
+         * because it's a primary.
+         *
+         * Notice: other clients, apart from the import source, should not access
+         * the data imported by import source.
```
hpatro, line 1838: Shouldn't this be enforced by server instead of providing a warning in the code comment/config file? The client's not marked with import_source should receive `LOADING` error if I'm understanding the use case correctly.

src/server.c
```diff
@@ -1651,7 +1651,7 @@ void beforeSleep(struct aeEventLoop *eventLoop) {
 
     /* Run a fast expire cycle (the called function will return
      * ASAP if a fast cycle is not needed). */
-    if (server.active_expire_enabled && iAmPrimary()) activeExpireCycle(ACTIVE_EXPIRE_CYCLE_FAST);
+    if (server.active_expire_enabled && !server.import_mode && iAmPrimary()) activeExpireCycle(ACTIVE_EXPIRE_CYCLE_FAST);
```
hpatro, line 1654: nit: we could have had similar conditional statement above at line 1058

```
!server.import_mode && iAmPrimary()
```

hpatro (commented):
my 2 cents: easy to introduce bug in the future where a dev can forget to check the import mode flag. It would be nice to introduce a macro for `serverInImportMode` / `serverInDataServingMode` and use it everywhere instead of checking if it's primary and in/not-in import mode independently.

src/networking.c
zuiderkwast: Ah, the client can't use CLIENT IMPORT-SOURCE OFF if import-mode has already been turned off. That's a minor annoyance I think. Import-source has no effect when import-mode is off.

src/db.c
soloestoy: We have discussed this issue before. In real-world scenarios, there are not only cases where data is migrated to an empty instance, but also cases where data is migrated to an instance that already contains data (requiring the user to ensure that the data being migrated is not disrupted). We hope to continue providing normal services in this scenario.

zuiderkwast:
@hpatro @enjoy-binbin Sorry for merging too fast. I didn't know you were still reviewing and testing this.

src/db.c
zuiderkwast: Yes, the feature can cause some inconsistencies if misused, but we decided that it's better to let the user have the power than to limit them.

enjoy-binbin:
>  Sorry for merging too fast. I didn't know you were still reviewing and testing this.

no worry. i mute the notification at the very first beginning, and only review the code today (review and play with it while dealing the conflicts), the code LGTM overall.

src/server.c
zuiderkwast: If you comment on line 1058, it's easier to see what you mean. It is for keys created in a writable replica, if I'm not mistaken(?).

hpatro: My bad. while reviewing the code, it was next to each other.

src/networking.c
hpatro: I think we should clean it up though. Also, shall we update `CLIENT LIST` output to reflect the import source mode?

src/db.c
hpatro: Ack.

src/networking.c
lyq2333: Just print the flag? or add a new type so `client kill` can also use the type?

PingXie: > I think we should clean it up though. Also, shall we update `CLIENT LIST` output to reflect the import source mode?

shouldn't we fail the command unless the server `import_mode` is enabled? also great point on tagging this special client in `CLIENT LIST`. 


> Just print the flag? or add a new type so `client kill` can also use the type?

A new flag would suffice IMO.

PR #1188, opened 2024-10-18
Print logs when the cluster state changes to fail or the fail reason changes

enjoy-binbin:
This log allows us to easily distinguish between full coverage and
minority partition when the cluster fails. Sometimes it is not easy
to see the minority partition in a healthy shards (both primary and
replicas).

And we decided not to add a cluster_fail_reason field to cluster info.
Given that there are only two reasons and both are well-known and if
we ended up adding more down the road we can add it in the furture.


src/cluster_legacy.c
```diff
@@ -5282,6 +5283,20 @@ void clusterCloseAllSlots(void) {
...
 
+void clusterLogWhyFail(int reason) {
+    char *msg;
+    switch (reason) {
+    case CLUSTER_FAIL_NOT_FULL_COVERAGE:
+        msg = "Detect there is at least a hash slot uncovered (no available node is serving it). "
+              "Please check the 'cluster-require-full-coverage' configuration option.";
+        break;
+    case CLUSTER_FAIL_MINORITY_PARTITION: msg = "In a minority partition."; break;
+    default: msg = "Unknown reason code."; break;
```
PingXie: We shouldn't hit this case, right?
```suggestion
    default: serverPanic("Unknown reason code.");
```

src/cluster_legacy.c
```diff
@@ -5282,6 +5283,20 @@ void clusterCloseAllSlots(void) {
  * Cluster state evaluation function
  * -------------------------------------------------------------------------- */
 
+void clusterLogWhyFail(int reason) {
+    char *msg;
+    switch (reason) {
+    case CLUSTER_FAIL_NOT_FULL_COVERAGE:
+        msg = "Detect there is at least a hash slot uncovered (no available node is serving it). "
```
```suggestion
        msg = "At least one hash slot is not served by any available node.  "
```

src/cluster_legacy.c
```diff
@@ -5282,6 +5283,20 @@ void clusterCloseAllSlots(void) {
...
+              "Please check the 'cluster-require-full-coverage' configuration option.";
```
```suggestion
              "Please check the 'cluster-require-full-coverage' configuration.";
```

src/cluster_legacy.c
```diff
@@ -5282,6 +5283,20 @@ void clusterCloseAllSlots(void) {
...
+        break;
+    case CLUSTER_FAIL_MINORITY_PARTITION: msg = "In a minority partition."; break;
```
```suggestion
    case CLUSTER_FAIL_MINORITY_PARTITION: msg = "I am part of a minority partition."; break;
```

src/cluster_legacy.c
```diff
@@ -5282,6 +5283,20 @@ void clusterCloseAllSlots(void) {
...
+    default: msg = "Unknown reason code."; break;
+    }
+    serverLog(LL_NOTICE, "Currently cluster unable to work: %s", msg);
```
PingXie: Let's log errors at `WARNING`
```suggestion
    serverLog(LL_WARNING, "Cluster is currently down: %s", msg);
```

src/cluster_legacy.h
```diff
@@ -368,6 +368,7 @@ struct clusterState {
     clusterNode *myself; /* This node */
     uint64_t currentEpoch;
     int state;              /* CLUSTER_OK, CLUSTER_FAIL, ... */
+    int fail_reason;        /* Why the cluster state changes to fail. */
```
PingXie, line 373: can we include `fail_reason` in the `CLUSTER INFO` output as well?

src/cluster_legacy.c
```diff
@@ -5282,6 +5283,20 @@ void clusterCloseAllSlots(void) {
  * Cluster state evaluation function
  * -------------------------------------------------------------------------- */
 
+void clusterLogWhyFail(int reason) {
+    char *msg;
```
PingXie, line 5369: maybe be short-circuit the call here if `reason == CLUSTER_FAIL_NONE`?

src/cluster_legacy.c
enjoy-binbin: yes, we should panic actually. I copy this logic from clusterLogCantFailover and keep the default, i am changing it.

src/cluster_legacy.h
enjoy-binbin: that is a good idea, i add it to CLUSTER INFO, but it also made it a major decision.

src/cluster_legacy.c
```diff
@@ -5986,6 +6023,7 @@ sds genClusterInfoString(void) {
 
     info = sdscatprintf(info,
                         "cluster_state:%s\r\n"
+                        "cluster_fail_reason:%s\r\n"
```
madolson: I honestly would rather not add an info field. This is one of those things were if the cluster is down, you get more useful information from the other fields (cluster_slots_assigned and cluster_size) and I don't want to add *yet another field* to look at. It's also not necessarily one or the other, it could be both.

enjoy-binbin: yean, this does seem useless, @PingXie thoughts?

PingXie: I am fine with not adding this field now. 

That said, my thinking was that this new field would serve as the high level summary for the failure and then based on the information here, one could then zoom in the details like the cluster slot assignment states. it is probably not that big a deal to include this field right now given that there are only two reasons and both are well-known but if we ended up adding more down the road I think this field would help tie various pieces of states together.

enjoy-binbin: agree, i removed it, please take a new review

src/cluster_legacy.c
```diff
@@ -5362,6 +5363,22 @@ void clusterCloseAllSlots(void) {
...
+    char *msg;
+    switch (reason) {
+    case CLUSTER_FAIL_NOT_FULL_COVERAGE:
+        msg = "At least one hash slot is not served by any available node. "
+              "Please check the 'cluster-require-full-coverage' configuration.";
+        break;
+    case CLUSTER_FAIL_MINORITY_PARTITION: msg = "I am part of a minority partition."; break;
+    default: serverPanic("Unknown fail reason code.");
+    }
+    server.cluster->fail_reason = reason;
```
PingXie: maybe move "server.cluster->fail_reason = reason;" out of this function and in its caller. The name `clusterLogFailReason` suggests that this is a stateless logging function so it is a bit odd to have it update the server state. It does mean that we would need to "duplicate" this line but I personally feel that the code is more readable that way.

PingXie (approved):
LGTM with just one minor comment

src/cluster_legacy.c
enjoy-binbin: yean, i can move it. Noted clusterLogCantFailover also do the same thing, it change the server state in the clusterLogCantFailover function itself.

PingXie: yeah - there was an apparently different bar back then :)

src/cluster_legacy.c
```diff
@@ -5362,6 +5363,21 @@ void clusterCloseAllSlots(void) {
...
+void clusterLogFailReason(int reason) {
+    if (reason == CLUSTER_FAIL_NONE) return;
+
+    char *msg;
+    switch (reason) {
+    case CLUSTER_FAIL_NOT_FULL_COVERAGE:
+        msg = "At least one hash slot is not served by any available node. "
+              "Please check the 'cluster-require-full-coverage' configuration.";
+        break;
+    case CLUSTER_FAIL_MINORITY_PARTITION: msg = "I am part of a minority partition."; break;
```
zuiderkwast: Different case, different style?

We can use same style in both cases.

```suggestion
    case CLUSTER_FAIL_MINORITY_PARTITION:
        msg = "I am part of a minority partition.";
        break;
```

src/cluster.h
```diff
@@ -12,6 +12,12 @@
 #define CLUSTER_FAIL 1                                              /* The cluster can't work */
 #define CLUSTER_NAMELEN 40                                          /* sha1 hex length */
 
+/* Reason why the cluster state changes to fail. When adding new reasons,
+ * make sure to update clusterLogFailReason. */
+#define CLUSTER_FAIL_NONE 0
+#define CLUSTER_FAIL_NOT_FULL_COVERAGE 1
+#define CLUSTER_FAIL_MINORITY_PARTITION 2
```
zuiderkwast, line 19: If we use `enum` and don't add a `default:` case in the switch, there will be a compile warning for the switch if we add another value and forget to add a case for it in the switch. Just an idea. (I learned it from Madelyn.)

OTOH, the style in this file is to use integer and define, so we can follow this style, as you prefer.

zuiderkwast (approved):

PR #780, opened 2024-07-13
Improve cluster cant failover log conditions

enjoy-binbin:
This PR adjusts the logging conditions of clusterLogCantFailover
in this two ways.

1. For the same cant_failover_reason, we will print the log once
in CLUSTER_CANT_FAILOVER_RELOG_PERIOD, but its value is 10s, which
is a bit long, shorten it to 1s, so we can better track its state.
We get to see the system making progress by watching the message.
Using 1s also covers pretty much all cases as i don't see a reason
for using a <1s node timeout, test or prod.

2. We will not print logs before the nolog_fail_time, its value
is cluster-node-timeout+5000. This may casue us to lose some logs,
for example, if cluster-node-timeout is small, auth_timeout will
be 2000, and auth_retry_time will be 4000. In this case, we will
lose all the reasons during the election if the failover is timedout.
So remove the nolog_fail_time logic, since we still do have the
CLUSTER_CANT_FAILOVER_RELOG_PERIOD logic, we won't print too many
logs.

PR #1189, opened 2024-10-18
Limit CLUSTER_CANT_FAILOVER_DATA_AGE log to 10 times period

enjoy-binbin:
If a replica is step into data_age too old stage, it can not
trigger the failover and currently it can not be automatically
recovered and we will print a log every CLUSTER_CANT_FAILOVER_RELOG_PERIOD,
which is every second. If the primary has not recovered or there is
no manual failover, this log will flood the log file.

In this case, limit its frequency to 10 times period, which is
10 seconds in our code. Also in this data_age too old stage,
the repeated logs also can stand for the progress of the failover.

See also #780 for more details about it.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -4439,6 +4439,12 @@ void clusterLogCantFailover(int reason) {
         time(NULL) - lastlog_time < CLUSTER_CANT_FAILOVER_RELOG_PERIOD)
         return;
 
+    /* If data age is too old, this log may be printed repeatedly since it
+     * can not be automatically recovered. In this case, limit its frequency. */
+    if (reason == server.cluster->cant_failover_reason && reason == CLUSTER_CANT_FAILOVER_DATA_AGE &&
+        time(NULL) - lastlog_time < 10 * CLUSTER_CANT_FAILOVER_RELOG_PERIOD)
+        return;
+
     server.cluster->cant_failover_reason = reason;
 
     switch (reason) {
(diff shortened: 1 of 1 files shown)


src/cluster_legacy.c
```diff
@@ -4439,6 +4439,12 @@ void clusterLogCantFailover(int reason) {
         time(NULL) - lastlog_time < CLUSTER_CANT_FAILOVER_RELOG_PERIOD)
         return;
 
+    /* If data age is too old, this log may be printed repeatedly since it
+     * can not be automatically recovered. In this case, limit its frequency. */
+    if (reason == server.cluster->cant_failover_reason && reason == CLUSTER_CANT_FAILOVER_DATA_AGE &&
+        time(NULL) - lastlog_time < 10 * CLUSTER_CANT_FAILOVER_RELOG_PERIOD)
+        return;
+
```
PingXie: This seems a bit more readable? What do you think?

```suggestion
    time_t now = time(NULL);
    
    /* General logging suppression if the same reason has occurred recently. */
    if (now - lastlog_time < CLUSTER_CANT_FAILOVER_RELOG_PERIOD) {
        return;
    }

    /* Special case: If the failure reason is due to data age, log 10 times less frequently. */
    if (reason == CLUSTER_CANT_FAILOVER_DATA_AGE &&
        now - lastlog_time < 10 * CLUSTER_CANT_FAILOVER_RELOG_PERIOD) {
        return;
    }
}
```

PingXie (approved):

src/cluster_legacy.c
enjoy-binbin: It seems like there is a problem. We still need to check `reason == server.cluster->cant_failover_reason`, otherwise we may miss some logs (during conversion)

PingXie: yeah I assumed there is a check of `reason == server.cluster->cant_failover_reason` before my suggestions :). github doesn't allow me to include that line in the suggested change ...

PR #1190, opened 2024-10-18
Make replica CLUSTER RESET flush async based on lazyfree-lazy-user-flush

enjoy-binbin:
Currently, if the replica has a lot of data, CLUSTER RESET
will block for a while and report the slowlog, and it seems
that there is no harm in making it async so external components
can be easier when monitoring it.

### DIFF
--- a/valkey.conf
+++ b/valkey.conf
@@ -1309,9 +1309,9 @@ lazyfree-lazy-user-del yes
 # commands. When neither flag is passed, this directive will be used to determine
 # if the data should be deleted asynchronously.
 #
-# During a resetting, when a replica performs a CLUSTER RESET, the content
-# of the whole database is removed in order to become an empty primary, this
-# directive will be used to determine f the data should be deleted asynchronously.
+# When a replica performs a node reset via CLUSTER RESET, the entire
+# database content is removed to allow the node to become an empty primary.
+# This directive also determines whether the data should be deleted asynchronously.
 #
 # There are many problems with running flush synchronously. Even in single CPU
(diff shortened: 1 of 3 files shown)


zuiderkwast (approved):

madolson (commented):
This feels more like `lazyfree-lazy-user-flush` to me since it is the result of a user action. The lazy-free replica always seemed more random since it could happen arbitrarily.

zuiderkwast (approved):

valkey.conf
```diff
@@ -1308,7 +1308,11 @@ lazyfree-lazy-user-del yes
 # deletion, which can be controlled by passing the [SYNC|ASYNC] flags into the
 # commands. When neither flag is passed, this directive will be used to determine
 # if the data should be deleted asynchronously.
-
+#
+# During a resetting, when a replica performs a CLUSTER RESET, the content
+# of the whole database is removed in order to become an empty primary, this
+# directive will be used to determine f the data should be deleted asynchronously.
+#
```
PingXie: maybe move this section below the original comments so we start with generic information and follow up with specific cases?

```suggestion
#
# When a replica performs a node reset via CLUSTER RESET, the entire
# database content is removed to allow the node to become an empty primary.
# This directive also determines whether the data should be deleted asynchronously.
#
```

PingXie (approved):
some wording recommendation

valkey.conf
enjoy-binbin: nice polish, thanks

madolson (approved):
Shipit

PR #1191, opened 2024-10-18
Mark the node as FAIL when the node is marked as NOADDR and broadcast the FAIL

enjoy-binbin:
Imagine we have a cluster, for example a three-shard cluster,
if shard 1 doing a CLUSTER RESET HARD, it will change the node
name, and then other nodes will mark it as NOADR since the node
name received by PONG has changed.

In the eyes of other nodes, there is one working primary node
left but with no address, and in this case, the address report
in MOVED will be invalid and will confuse the clients. And in
the same time, the replica will not failover since its primary
is not in the FAIL state. And the cluster looks OK to everyone.

This leaves a cluster that appears OK, but with no coverage for
shard 1, obviously we should do something like CLUSTER FORGET
to remove the node and fix the cluster before using it.

But the point in here, we can mark the NOADDR node as FAIL to
advance the cluster state. If a node is NOADDR means it does
not have a valid address, so we won't reconnect it, we won't
send PING, we won't gossip it, it seems reasonable to mark it
as FAIL.


src/cluster_legacy.c
```diff
@@ -3250,13 +3250,20 @@ int clusterProcessPacket(clusterLink *link) {
...
                           link->node->name, link->node->human_nodename, link->node->shard_id,
                           (int)(now - (link->node->ctime)), link->node->flags);
-                link->node->flags |= CLUSTER_NODE_NOADDR;
+                /* We also mark the node as fail because we have disconnected from it,
+                 * and will not reconnect, and obviously we will not gossip NOADDR nodes.
+                 * Marking it as FAIL can help us advance the state, such as the cluster
+                 * state becomes FAIL or the replica can do the failover. Otherwise, the
+                 * NOADDR node will provide an invalid address in redirection and confuse
+                 * the client, and the replica will not initiate a failover since the node
+                 * is not actually in FAIL state. */
```
zuiderkwast: Let's combine this comment with the comment above?

```diff
             } else if (memcmp(link->node->name, hdr->sender, CLUSTER_NAMELEN) != 0) {
                 /* If the reply has a non matching node ID we
                  * disconnect this node and set it as not having an associated
-                 * address. */
+                 * address. This can happen if the node did CLUSTER RESET and changed
+                 * its node ID. In this case, the old node ID will not come back.
+                 * 
+                 * We also mark the node as fail because we have disconnected from it,
+                 * and will not reconnect, and obviously we will not gossip NOADDR nodes.
+                 * Marking it as FAIL can help us advance the state, such as the cluster
+                 * state becomes FAIL or the replica can do the failover. Otherwise, the
+                 * NOADDR node will provide an invalid address in redirection and confuse
+                 * the client, and the replica will not initiate a failover since the node
+                 * is not actually in FAIL state. */
```

zuiderkwast (commented):
It seems to be correct, but I want someone else to take a look. @PingXie?

So when a node is in NOADDR state, it will never be PFAIL and later FAIL? We never get any updates from the old node ID so it should automatically become PFAIL at some point?

Are there any other case where a node can be marked as NOADDR and come back again? Changed IP address of the server but still running?

enjoy-binbin:
> So when a node is in NOADDR state, it will never be PFAIL and later FAIL? We never get any updates from the old node ID so it should automatically become PFAIL at some point?

yes, it won't never be PFAIL or FAIL. It also won't automatically become PFAIL since in clusterCron, we will skip the NOADDR node for the timeout check.

> Are there any other case where a node can be marked as NOADDR and come back again? Changed IP address of the server but still running?

Maybe, but i am not aware of it. If the IP changed, we will call nodeUpdateAddressIfNeeded to update it, so it won't be NOADDR state.

zuiderkwast:
Just an idea: Can we set it to PFAIL? It is unreachable in myself's view but maybe another node can reach it somehow? To mark a node as FAIL is usually a majority decision. We can wait for a majority of nodes to mark it as FAIL, but it takes more time. Is that a problem?

enjoy-binbin:
we won't include the noaddr node in the gossip section. That is the problem so we will never get the majority

zuiderkwast (approved):

zuiderkwast:
> we won't include the noaddr node in the gossip section. That is the problem so we will never get the majority

I guess another option is to start including it in the gossip section then.

PingXie:
>Imagine we have a cluster, for example a three-shard cluster,
if shard 1 doing a CLUSTER RESET HARD, it will change the node
name, and then other nodes will mark it as NOADR since the node
name received by PONG has changed.

This sounds like a human error to me to begin with? If proper steps, such as failing over the primary, cluster forget, etc, were followed, we wouldn't enter this state, IMO. Did this actually happen due to runtime errors?

>obviously we should do something like CLUSTER FORGET
to remove the node and fix the cluster before using it.

Generally speaking, I am more in favor of operational mitigations for non-runtime errors like this. That said, the proposal fix makes sense to me too.

src/cluster_legacy.c
```diff
@@ -3244,19 +3244,28 @@ int clusterProcessPacket(clusterLink *link) {
...
                           "having flags %d",
                           link->node->name, link->node->human_nodename, link->node->shard_id,
                           (int)(now - (link->node->ctime)), link->node->flags);
-                link->node->flags |= CLUSTER_NODE_NOADDR;
+                link->node->flags |= (CLUSTER_NODE_NOADDR | CLUSTER_NODE_FAIL);
                 link->node->ip[0] = '\0';
                 link->node->tcp_port = 0;
                 link->node->tls_port = 0;
                 link->node->cport = 0;
                 freeClusterLink(link);
```
PingXie, line 3361: I think we should record the fail_time for the node and broadcast the FAIL message. Other replicas with higher offsets might not have received this message.

zuiderkwast:
> > obviously we should do something like CLUSTER FORGET
> > to remove the node and fix the cluster before using it.
> 
> Generally speaking, I am more in favor of operational mitigations for non-runtime errors like this. That said, the proposal fix makes sense to me too.

Good point. Then the problem is that the admin API is not safe. In a good API, it should not be possible to mess up the runtime state like this.

When CLUSTER RESET HARD is used, can we handle it as if CLUSTER FORGET of the old node id was called? Add the old node id to forgotten nodes blacklist, etc.

enjoy-binbin:
> When CLUSTER RESET HARD is used, can we handle it as if CLUSTER FORGET of the old node id was called? Add the old node id to forgotten nodes blacklist, etc.

I have also thought about this, for example, gossip the CLUSTER RESET, but finally gave up, not sure if there are other issues.

zuiderkwast:
Gossip the CLUSTER RESET seems like a new message = more work. We already gossip the forgotten nodes, so if we use this, it's not too difficult I think. Also not sure if there are other issues though.

src/cluster_legacy.c
enjoy-binbin: i haven't thought about it, broadcast the FAIL. So there may be a chance that a lower offsets rank replica get the message first and mark it as FAIL and trigger a failover. yean, that is a problem. The broadcast seems like a good idea, i will take a try.

enjoy-binbin:
If we don't want to set the NOADDR node to FAIL, a another way is to add a check for NOADDR node in clusterCron / cluster info so that the cluster can report FAIL. And add a check for NOADDR primary so that a replica has a way that can trigger a failover.

zuiderkwast:
> If we don't want to set the NOADDR node to FAIL, a another way is to add a check for NOADDR node in clusterCron / cluster info so that the cluster can report FAIL. And add a check for NOADDR primary so that a replica has a way that can trigger a failover.

This sounds like a workaround. I think we can set it to FAIL.

But I agree with Ping, we should probably broadcast the FAIL, because FAIL is normally a cluster majority agreed state. The cluster should have a consistent view of which nodes are FAIL and which are not.

zuiderkwast (approved):

PR #1194, opened 2024-10-19
RDMA: Fix listener priv opaque pointer

pizhenwei:
struct connListener.priv should be used by connection type specific data, static local listener data should not use this.

A RDMA config structure is going to be introduced in the next step:

```
typedef struct serverRdmaContextConfig {
    char *bindaddr;
    int bindaddr_count;
    int port;
    int rx_size;
    int comp_vector;
    ...
} serverRdmaContextConfig;
```

Then a builtin RDMA will be supported.

### DIFF
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -127,6 +127,8 @@ typedef struct rdma_listener {
  * handler into pending list */
 static list *pending_list;
 
+static rdma_listener *rdma_listeners;
+
 static ConnectionType CT_RDMA;
 
 static int valkey_rdma_rx_size = VALKEY_RDMA_DEFAULT_RX_SIZE;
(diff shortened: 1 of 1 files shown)


src/rdma.c
```diff
@@ -1537,7 +1539,7 @@ int connRdmaListen(connListener *listener) {
         bindaddr = default_bindaddr;
     }
 
-    listener->priv = rdma_listener = zcalloc_num(bindaddr_count, sizeof(*rdma_listener));
+    rdma_listeners = rdma_listener = zcalloc_num(bindaddr_count, sizeof(*rdma_listener));
```
zuiderkwast, line 1542: So priv is not used anymore?

zuiderkwast (approved):

src/rdma.c
pizhenwei: It will be used for `serverRdmaContextConfig` in the next step (for RDMA builtin).

PR #1196, opened 2024-10-20
Add CMake build system for valkey

eifrah-aws:
Replacing this PR https://github.com/valkey-io/valkey/pull/1082

**NOTE:**

This PR was tested on:
- Amazon Linux 2
- Ubuntu 22.04
- AlmaLinux 9
- macOS Sonoma 14.7 (ARM)
- FreeBSD 9 - **note** on FreeBSD, we require `gmake` for building `valkey`

---

With this PR, users are able to build valkey using `CMake`.

## Example usage:

Build `valkey-server` in Release mode with TLS enabled and using `jemalloc` as the allocator:

```bash
mkdir build-release
cd $_
cmake .. -DCMAKE_BUILD_TYPE=Release \
         -DCMAKE_INSTALL_PREFIX=/tmp/valkey-install \
         -DBUILD_MALLOC=jemalloc -DBUILD_TLS=1
make -j$(nproc) install

# start valkey
/tmp/valkey-install/bin/valkey-server
```

Build `valkey-unit-tests`:

```bash
mkdir build-release-ut
cd $_
cmake .. -DCMAKE_BUILD_TYPE=Release \
         -DBUILD_MALLOC=jemalloc -DBUILD_UNIT_TESTS=1
make -j$(nproc)

# Run the tests
./bin/valkey-unit-tests 
```

Current features supported by this PR:

- Building against different allocators: (`jemalloc`, `tcmalloc`, `tcmalloc_minimal` and `libc`), e.g. to enable `jemalloc` pass `-DBUILD_MALLOC=jemalloc` to `cmake`
- OpenSSL builds (to enable TLS, pass `-DBUILD_TLS=1` to `cmake`)
- Sanitizier: pass `-DBUILD_SANITIZER=<address|thread|undefined>` to `cmake`
- Install target + redis symbolic links
- Build `valkey-unit-tests` executable
- Standard CMake variables are supported. e.g. to install `valkey` under `/home/you/root` pass `-DCMAKE_INSTALL_PREFIX=/home/you/root`

Why using `CMake`? To list *some* of the advantages of using `CMake`:

- Superior IDE integrations: cmake generates the file `compile_commands.json` which is required by `clangd` to get a compiler accuracy code completion (in other words: your VScode will thank you)
- Out of the source build tree: with the current build system, object files are created all over the place polluting the build source tree, the best practice is to build the project on a separate folder
- Multiple build types co-existing: with the current build system, it is often hard to have multiple build configurations. With cmake you can do it easily:
- It is the de-facto standard for C/C++ project these days

More build examples: 

ASAN build:

```bash
mkdir build-asan
cd $_
cmake .. -DBUILD_SANITIZER=address -DBUILD_MALLOC=libc
make -j$(nproc)
```

ASAN with jemalloc:

```bash
mkdir build-asan-jemalloc
cd $_
cmake .. -DBUILD_SANITIZER=address -DBUILD_MALLOC=jemalloc 
make -j$(nproc)
```

As seen by the previous examples, any combination is allowed and co-exist on the same source tree.

## Valkey installation

With this new `CMake`, it is possible to install the binary by running `make install` or creating a package `make package` (currently supported on Debian like distros)

### Example 1: build & install using `make install`:

```bash
mkdir build-release
cd $_
cmake .. -DCMAKE_INSTALL_PREFIX=$HOME/valkey-install -DCMAKE_BUILD_TYPE=Release
make -j$(nproc) install
# valkey is now installed under $HOME/valkey-install
```

### Example 2: create a `.deb` installer:

```bash
mkdir build-release
cd $_
cmake .. -DCMAKE_BUILD_TYPE=Release
make -j$(nproc) package
# ... CPack deb generation output
sudo gdebi -n ./valkey_8.1.0_amd64.deb
# valkey is now installed under /opt/valkey
```

### Example 3: create installer for non Debian systems (e.g. FreeBSD or macOS):

```bash
mkdir build-release
cd $_
cmake .. -DCMAKE_BUILD_TYPE=Release
make -j$(nproc) package
mkdir -p /opt/valkey && ./valkey-8.1.0-Darwin.sh --prefix=/opt/valkey  --exclude-subdir
# valkey-server is now installed under /opt/valkey

```
----

Current `CMake` PR status:

- [x] Targets:
  - [x] valkey-server
  - [x] valkey-cli
  - [x] valkey-benchmark
  - [x] valkey-sentinel
  - [x] RDMA module
  - [x] TLS as module
  - [x] valkey-unit-tests
  - [x] valkey-check-aof
  - [x] valkey-check-rdb
- [x] Features
  - [x] SSL
  - [x] Sanitizer
  - [x] Different allocators (`jemalloc`, `libc`, `tcmalloc` and `tcmalloc_minimal`)
  - [x] Support `.deb` packaging
  - [x] Support for script based installer for non debian systems
- [x] Python code generation
  - [x] generate `release.h`
  - [x] generate `commands.def`
  - [x] generate `fmtargs.h`
  - [x] generate `test_files.h`


eifrah-aws:
@PingXie see this PR instead.

.cmake-format.yaml
```diff
@@ -0,0 +1,76 @@
...
+  max_lines_hwrap: 2
+  _help_line_ending:
+    - What style line endings to use in the output.
+  line_ending: unix
+  _help_command_case:
+    - Format command names consistently as 'lower' or 'upper' case
+  command_case: canonical
+  _help_keyword_case:
+    - Format keywords consistently as 'lower' or 'upper' case
+  keyword_case: unchanged
```
PingXie, line 41: can we align on 'lower' casing too?
```suggestion
  keyword_case: lower
```

.cmake-format.yaml
```diff
@@ -0,0 +1,76 @@
...
+  min_prefix_chars: 4
+  _help_max_prefix_chars:
+    - If the statement spelling length (including space and
+    - parenthesis) is larger than the tab width by more than this
+    - amount, then force reject un-nested layouts.
+  max_prefix_chars: 10
+  _help_max_lines_hwrap:
+    - If a candidate layout is wrapped horizontally but it exceeds
+    - this many lines, then reject the layout.
+  max_lines_hwrap: 2
```
PingXie, line 32: Will this rule (`max_lines_hwrap: 2`)  reformat the directives below into two lines, assuming each line is still below 120 characters?

```
target_link_libraries(
    valkey-unit-tests
    valkeylib
    fpconv
    lualib
    hdr_histogram
    hiredis
    ${VALKEY_SERVER_LDFLAGS})
```

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+cmake_minimum_required(VERSION 3.20)
+
+# Must be done first
+if (APPLE)
+    # Force clang compiler on macOS
+    set(CMAKE_CXX_COMPILER "/usr/bin/clang++")
+    set(CMAKE_C_COMPILER "/usr/bin/clang")
```
PingXie: should we use `find_program` instead to avoid hard-coding the compiler paths?

pizhenwei (commented):
Is it possible to use `WITH_TLS` instead of `USE_TLS`? Then we have a single TLS related building variable.

cmake/Modules/ValkeySetup.cmake
```diff
@@ -0,0 +1,349 @@
...
+
+# TLS support
+if (WITH_TLS)
+    find_package(OpenSSL REQUIRED)
+    message(STATUS "OpenSSL include dir: ${OPENSSL_INCLUDE_DIR}")
+    message(STATUS "OpenSSL libraries: ${OPENSSL_LIBRARIES}")
+
+    include_directories(${OPENSSL_INCLUDE_DIR})
+    add_valkey_server_compiler_options("-DUSE_OPENSSL=1")
+    add_valkey_server_compiler_options("-DBUILD_TLS_MODULE=0")
```
pizhenwei: TLS module support gets lost.

cmake/Modules/Packaging.cmake
```diff
@@ -0,0 +1,44 @@
+set(CPACK_PACKAGE_NAME "valkey")
+set(CPACK_PACKAGE_VERSION_MAJOR 8)
+set(CPACK_PACKAGE_VERSION_MINOR 1)
+set(CPACK_PACKAGE_VERSION_PATCH 0)
```
PingXie: 1. this is duplicated in `version.h`. can we single-source the version numbers in this file instead?
2. the version number for the unstable branch is 255.255.255. Does it make sense to instruct CPACK to not encode the versions in the packaging file name in this case?

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+option(WITH_TLS "Build valkey-server with TLS support" OFF)
+option(BUILD_TESTS "Build valkey-unit-tests" OFF)
+option(WITH_RDMA "Build valkey-rdma module" OFF)
+
+set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${CMAKE_SOURCE_DIR}/cmake/Modules/")
+project("valkey")
+
+set(CMAKE_C_STANDARD 11)
+set(CMAKE_C_STANDARD_REQUIRED ON)
+set(CMAKE_C_EXTENSIONS OFF)
```
PingXie: This is in theory more restrictive than what we have today: `-std=gnu11` is C11 plus gnu extensions I believe. I am surprised that this doesn't break the build? can you double check?

cmake/Modules/ValkeySetup.cmake
```diff
@@ -0,0 +1,349 @@
...
+    set(USE_TLS 1)
+endif ()
+
+if (WITH_RDMA)
+    # RDMA support (Linux only)
+    if (LINUX AND NOT APPLE)
+        add_valkey_server_compiler_options("-DUSE_RDMA=2")
```
pizhenwei: Is it possible to add a helper function to parse `module/yes/no` for both TLS and RDMA? even for QUIC in the future.
Then we have `-DWITH_TLS=yes`, `-DWITH_TLS=module` and `-DWITH_TLS=no` rather than `-DWITH_TLS=1`.

cmake/Modules/SourceFiles.cmake
```diff
@@ -0,0 +1,149 @@
...
+    ${CMAKE_SOURCE_DIR}/src/crccombine.c
+    ${CMAKE_SOURCE_DIR}/src/crc64.c
+    ${CMAKE_SOURCE_DIR}/src/siphash.c
+    ${CMAKE_SOURCE_DIR}/src/crc16.c
+    ${CMAKE_SOURCE_DIR}/src/monotonic.c
+    ${CMAKE_SOURCE_DIR}/src/cli_common.c
+    ${CMAKE_SOURCE_DIR}/src/mt19937-64.c
+    ${CMAKE_SOURCE_DIR}/src/strl.c)
+
+# valkey-rdmo module
```
PingXie:
```suggestion
# valkey-rdma module
```

cmake/Modules/ValkeySetup.cmake
PingXie: so we will no longer build TLS as a module? It makes sense to me but I just wanted to make sure this is indeed intentional.

PingXie (commented):
Thanks @eifrah-aws

cmake/Modules/ValkeySetup.cmake
PingXie: We discussed this a few times. My impression so far is that no one builds TLS as a module. Is this flexibility still needed?

cmake/Modules/ValkeySetup.cmake
PingXie: Same question here - do we get real benefit from making these components true modules? The build time optionality is there. The run-time optionality is lost but I am curious to know if this is indeed needed in a production environment?

CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+
+# Must be done first
+if (APPLE)
+    # Force clang compiler on macOS
+    set(CMAKE_CXX_COMPILER "/usr/bin/clang++")
+    set(CMAKE_C_COMPILER "/usr/bin/clang")
+endif ()
+
+# Options
+option(WITH_TLS "Build valkey-server with TLS support" OFF)
```
PingXie: Would `USE` be a more conventional word? 

```suggestion
option(USE_TLS "Build valkey-server with TLS support" OFF)
```

pizhenwei:
Is it possible to generate `config.h` by cmake?
I guess this may be the next step.

cmake/Modules/ValkeySetup.cmake
pizhenwei: The module style connection driver proves the well abstracted framework. Haha!
Sure, builtin connection driver is fine to me, too! I agree that there is no obvious benefit from module components.

.cmake-format.yaml
eifrah-aws: No. (all files were reformatted before committed)

.cmake-format.yaml
eifrah-aws: Done

eifrah-aws: No. This will change all keywords to lowercase. So, for example this:

```cmake
unset(WITH_TLS CACHE)
```
becomes:

```cmake
unset(WITH_TLS cache)
```

which yields an error (`cache` must be using UPPERCASE)

I did changed the `command_case` to `lower` (so `FIND_PROGRAM` becomes `find_program` which is OK)

cmake/Modules/Packaging.cmake
eifrah-aws: We have 2 options here:
1. change version.h into version.h.in and then call `configure_file` on it to replace the placeholders with the actual version number
2. Keep things as is and parse these values from `VALKEY_VERSION` from `version.h`

For Debian systems, this is the standard way: include the version as part of the `.deb` file.

For example, if you are running ubuntu/debian/mint etc you can run:

```
dpkg -l
...
ii  irqbalance                                       1.9.3-2ubuntu5                            amd64        Daemon to ba>
ii  iso-codes                                        4.16.0-1                                  all          ISO language>
ii  isort                                            5.6.4-1                                   all          utility for >
ii  iucode-tool                                      2.3.1-3build1                             amd64        Intel proces>
ii  java-common                                      0.75+exp1                                 all          Base package>
ii  javascript-common                                11+nmu1                                   all          Base support>
ii  jq                                               1.7.1-3build1                             amd64        lightweight >
ii  kbd                                              2.6.4-2ubuntu2                            amd64        Linux consol>
```

You will notice that all packages include the version number.

eifrah-aws: For now, and in order to keep `Makefile` functioning, I will read the version from `version.h` file

cmake/Modules/ValkeySetup.cmake
eifrah-aws: If its really important, I can restore this functionality (i.e. building TLS as a module). tbh, I was not aware that TLS can be built as module until I read the `Makefile`... 😛

cmake/Modules/ValkeySetup.cmake
eifrah-aws: From what I see in the Makefile, RDMA is either supported as module or not at all

```make
$(error "RDMA is only supported as module (BUILD_RDMA=module), or disabled (BUILD_RDMA=no)")
```

pizhenwei: I'm working on `BUILD_RDMA=yes`, I plan to support it this week.

eifrah-aws: Added support for:

```bash
cmake .. -DWITH_TLS=module # optionally you can pass here: yes|no|1|0
```

The same can be done with RDMA:

```bash
cmake .. -DWITH_RDMA=module # optionally you can pass here: yes|no|1|0
```

While you can still pass `1` or `yes` to`-DWITH_RDMA` options, internally, it will be disabled as we only support module mode for RDMA

CMakeLists.txt
eifrah-aws: I double checked on macOS & Ubuntu with various build combinations (TLS/RDMA etc) -works

The Makefile has this:

```make
 40 # Use -Wno-c11-extensions on clang, either where explicitly used or on
 41 # platforms we can assume it's being used.
 42 ifeq (clang,$(CLANG))
 43   STD+=-Wno-c11-extensions
 44 else
 45 ifneq (,$(findstring FreeBSD,$(uname_S)))
 46   STD+=-Wno-c11-extensions
 47 endif
 48 endif
```

while `-std=c99` and `-std=gnu11` is set by the detection of `stdatomic.h` as seen here:

```make
 52 # Detect if the compiler supports C11 _Atomic.
 53 # NUMBER_SIGN_CHAR is a workaround to support both GNU Make 4.3 and older versions.
 54 NUMBER_SIGN_CHAR := \#
 55 C11_ATOMIC := $(shell sh -c 'echo "$(NUMBER_SIGN_CHAR)include <stdatomic.h>" > foo.c; \
 56     $(CC) -std=gnu11 -c foo.c -o foo.o > /dev/null 2>&1; \
 57     if [ -f foo.o ]; then echo "yes"; rm foo.o; fi; rm foo.c')
 58 ifeq ($(C11_ATOMIC),yes)
 59     STD+=-std=gnu11
 60 else
 61     STD+=-std=c99
 62 endif
```

eifrah-aws: The above check is implemented in `ValkeySetup.cmake` here:
https://github.com/eifrah-aws/valkey/blob/cmake/cmake/Modules/ValkeySetup.cmake#L233

CMakeLists.txt
eifrah-aws: From the `configure` world, I mostly seen:

For boolean flags:

`--enable-XXX`, `--disable-XXX` (e.g. --enable-shared) etc

For flags with values, it is conventional to use `--with-XXXX` for example:

```
./configure --with-jemalloc-prefix=<prefix> ...
```

In any case, I removed this line completely as it is no longer a boolean option (it can accept: `module`, `yes` or `no`)

cmake/Modules/ValkeySetup.cmake
eifrah-aws: Fixed. TLS module functionality has been restored

eifrah-aws: Fixed. TLS can now be built as module as well.

eifrah-aws:
> Is it possible to generate config.h by cmake?
I guess this may be the next step.

Yes, but this will break Makefile users. So until CMake is agreed as the main build tool, I wouldn't go this path

cmake/Modules/Utils.cmake
```diff
@@ -0,0 +1,102 @@
...
+
+# Given input argument `OPTION_VALUE`, check that the `OPTION_VALUE` is from the allowed values (one of:
+# module/yes/no/1/0/true/false)
+#
+# Return value:
+#
+# If ARG is valid, return its number where:
+#
+# ~~~
+# - `no` | `0`  => return `0`
```
pizhenwei: Missing `on` and `off` in comment

cmake/Modules/ValkeySetup.cmake
```diff
@@ -0,0 +1,364 @@
...
+
+if (WITH_RDMA)
+    # RDMA support (Linux only)
+    if (LINUX AND NOT APPLE)
+        valkey_parse_build_option(${WITH_RDMA} USE_RDMA)
+        if (USE_RDMA EQUAL 2) # Module
+            message(STATUS "Building RDMA as module")
+            add_valkey_server_compiler_options("-DUSE_RDMA=2")
+            find_package(PkgConfig REQUIRED)
+            pkg_check_modules(VALKEY_RDMA REQUIRED librdmacm)
```
pizhenwei: The two RDMA related libraries are used by Valkey Over RDMA: `librdmacm`(for connection management) and `libibverbs`(for payload transmission).
So `VALKEY_RDMA` should be `VALKEY_RDMACM`.

pizhenwei (commented):
Hi, I also find the missing CMakeLists.txt for tests:
```
$ find tests -name Makefile
tests/rdma/Makefile
tests/modules/Makefile
```

And cmake related help information in `README.md` is needed?

PingXie:
I wonder if we should drop building TLS as a module now in CMake. Looks like RDMA can only be built as a module now, which is fine, as long as we don't support two build systems. 

cmake/Modules/ValkeySetup.cmake
eifrah-aws: Fixed

eifrah-aws:
> Hi, I also find the missing CMakeLists.txt for tests:

Fixed with recent commit

eifrah-aws:
> And cmake related help information in README.md is needed?

README updated

README.md
```diff
@@ -289,16 +279,93 @@ system reboots.
 You'll be able to stop and start Valkey using the script named
 `/etc/init.d/valkey_<portnumber>`, for instance `/etc/init.d/valkey_6379`.
 
-Code contributions
------------------
+# Building using `CMake`
+
+In addition to the traditional `Makefile` build, Valkey supports an alternative, **experimental** build system using `CMake`
```
madolson:
```suggestion
In addition to the traditional `Makefile` build, Valkey supports an alternative, **experimental**, build system using `CMake`.
```

README.md
```diff
@@ -289,16 +279,93 @@ system reboots.
...
+
+To build and install `Valkey`, in `Release` mode (an optimized build), type this into your terminal:
+
+For Linux:
```
madolson: General preference for a single oneshot that works on all systems, even if it omits `-j`.

src/modules/CMakeLists.txt
madolson, line 1: We don't want to build these by default, they're mostly leftovers from when we didn't have official tests and are more of "Examples". We probably want to remove these from the official repo at some point.

src/debug.c
```diff
@@ -1023,6 +1023,7 @@ void debugCommand(client *c) {
 
 /* =========================== Crash handling  ============================== */
 
+#ifndef VALKEY_USE_TEST_SERVER_ASSERT
```
madolson: Didn't see another discussion, but why is this here?

src/server.c
```diff
@@ -7063,5 +7064,5 @@ int main(int argc, char **argv) {
     aeDeleteEventLoop(server.el);
     return 0;
 }
-
+#endif // VALKEY_USE_TEST_MAIN
```
```suggestion
#endif /* VALKEY_USE_TEST_MAIN */
```

README.md
```diff
@@ -289,16 +279,93 @@ system reboots.
...
+
+Other options supported by Valkey's `CMake` build system:
+
+## Special build flags
+
+- `-DWITH_TLS=<on|off|module>` enable TLS build for Valkey
+- `-DWITH_RDMA=<off|module>` enable RDMA module build (only module mode supported)
+- `-DWITH_MALLOC=<libc|jemalloc|tcmalloc|tcmalloc_minimal>` choose the allocator to use. Default on Linux: `jemalloc`, for other OS: `libc`
+- `-DWITH_SANITIZER=<address|thread|undefined>` build with address sanitizer enabled
+- `-DBUILD_UNIT_TESTS=[1|0]`  when enabled, the build will produce the executable `valkey-unit-tests`
```
madolson: BUILD_UNIT_TESTS seems unused, I wasn't able to get the Unit tests to get built.

madolson (commented):
I am fine removing building TLS as a module. No longer really seems necessary.

Before merging, we should all add it to the CI to make sure it runs as expected.

README.md
eifrah-aws: Unified macOS & Linux. Removed the `-j`

src/modules/CMakeLists.txt
eifrah-aws: Ack. Will make these optional until they are removed

src/debug.c
eifrah-aws: on macOS / ARM, the linker can't handle the duplicate symbol of `serverAssert` + `main` (both redefined in `test_main.c`). CMake now exposes these macros to skip these functions when building the unit tests as these functions can be resolved from `test_main.c`

README.md
eifrah-aws: Fixed

cmake/Modules/SourceFiles.cmake
```diff
@@ -0,0 +1,156 @@
...
+    ${CMAKE_SOURCE_DIR}/src/strl.c)
+
+# valkey-rdma module
+set(VALKEY_RDMA_MODULE_SRCS ${CMAKE_SOURCE_DIR}/src/rdma.c)
+
+# valkey-tls module
+set(VALKEY_TLS_MODULE_SRCS ${CMAKE_SOURCE_DIR}/src/tls.c)
+
+# rdma-test source files
+set(VALKEY_RDMA_MODULE_TEST_SRCS ${CMAKE_SOURCE_DIR}/tests/rdma/rdma-test.c)
```
pizhenwei: What about moving this into `tests/CmakeLists.txt` and `tests/rdma/CmakeLists.txt`(test WITH_RDMA)?

eifrah-aws: > What about moving this into `tests/CmakeLists.txt` and `tests/rdma/CmakeLists.txt`(test WITH_RDMA)?

I can see the logic in your suggestion and also in keeping it here under "SourceFiles" file. I have no strong opinion about it... I will move it

pizhenwei (commented):
LGTM. Also test this PR on my local platform `Ubuntu 22.04.4 LTS`, it works fine.

It would be appreciated if adding these message into commit message.

Tested-by: zhenwei pi <pizhenwei@bytedance.com>
Reviewed-by: zhenwei pi <pizhenwei@bytedance.com>

eifrah-aws:
@madolson per your suggestion, I made the symbols `main` & `serverAssert` "weak" to allow them to be overridden by symbols from `test_main.c` file. Tested on Ubuntu. Amzn Linux 2, FreeBSD & macOS

eifrah-aws:
@PingXie @madolson can one of you re-start the failing CI job?

eifrah-aws:
@PingXie / @madolson can one of you please re-run the the flaky test? (I am pretty sure its flaky as this PR does not change any source file, nor it tests the artifacts produced by `CMake` 😉 )

Also, are there any outstanding issues that preventing us from merging this PR?

PingXie:
Yeah that is a flaky test.

>Also, are there any outstanding issues that preventing us from merging this PR?

Can we drop the option to build TLS as a module? I think now it is a good time to make this change.

@pizhenwei do you have any concerns?

pizhenwei:
> Yeah that is a flaky test.
> 
> > Also, are there any outstanding issues that preventing us from merging this PR?
> 
> Can we drop the option to build TLS as a module? I think now it is a good time to make this change.
> 
> @pizhenwei do you have any concerns?

Dropping tls module option is OK to me!

eifrah-aws:
> Can we drop the option to build TLS as a module? I think now it is a good time to make this change.

Great, will drop it

madolson (commented):
Looks good to me, was able to test and run everything locally on my x86 mac. The only pending item I have now is I would like us to add it to CI so that we don't silently break it. Let's add a separate test here: https://github.com/valkey-io/valkey/blob/unstable/.github/workflows/ci.yml#L13 that does cmake + make. I think for now just doing it on linux is fine.

@valkey-io/core-team I added the approved tag since we did discuss this in our weekly meeting. But tagging again incase anyone else would like to review before we merge.

enjoy-binbin (approved):
Look ok to me. (i don't know much about it so did not do the review)

eifrah-aws:
@madolson / @PingXie 

- TLS module build was dropped
- [CI updated with CMake](https://github.com/valkey-io/valkey/actions/runs/11694153505/job/32567212616?pr=1196) (running tcl tests + `runtests` suite)

madolson (approved):
Didn't see anyone else complain, so I guess we can merge it and work to maintain it now.

README.md
```diff
@@ -289,16 +279,82 @@ system reboots.
...
+mkdir build-release
+cd $_
+cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=/opt/valkey
+sudo make install
+# Valkey is now installed under /opt/valkey
+```
+
+Other options supported by Valkey's `CMake` build system:
+
+## Special build flags
```
PingXie, line 298: not blocking this PR but can we also document the default settings for these flags?

cmake/Modules/Packaging.cmake
```diff
@@ -0,0 +1,44 @@
+set(CPACK_PACKAGE_NAME "valkey")
+
+valkey_parse_version(CPACK_PACKAGE_VERSION_MAJOR CPACK_PACKAGE_VERSION_MINOR CPACK_PACKAGE_VERSION_PATCH)
```
PingXie, line 3: can we consider a custom versioning scheme for the unstable branch such as `${CPACK_PACKAGE_NAME}-unstable-${CMAKE_SYSTEM_NAME}`? It will look like `${CPACK_PACKAGE_NAME}-255.255.255-${CMAKE_SYSTEM_NAME}` otherwise. I am a bit more in favor of explicitness. Separate PR is totally fine.

PingXie (approved):
Great work, [eifrah-aws](https://github.com/eifrah-aws)! Appreciate it

cmake/Modules/Packaging.cmake
eifrah-aws: +1 for `${CPACK_PACKAGE_NAME}-unstable-${CMAKE_SYSTEM_NAME}` for unstable branch

README.md
eifrah-aws: Will be fixed as part of this small PR: https://github.com/valkey-io/valkey/pull/1276

PR #1063, opened 2024-09-23
Fix module / script call CLUSTER SLOTS / SHARDS fake client check crash

enjoy-binbin:
The reason is VM_Call will use a fake client without connection,
so we also need to check if c->conn is NULL.

This also affects scripts. If they are called in the script, the
server will crash. Injecting commands into AOF will also cause
startup failure.

Fixes #1054.

PR #1198, opened 2024-10-21
Use fake client flag to replace not conn check

enjoy-binbin:
The fake client flag was introduced in #1063, 
we want this to replace all !conn fake client checks.

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -4440,6 +4440,7 @@ int checkClientOutputBufferLimits(client *c) {
  * Returns 1 if client was (flagged) closed. */
 int closeClientOnOutputBufferLimitReached(client *c, int async) {
     if (c->flag.fake) return 0; /* It is unsafe to free fake clients. */
+    serverAssert(c->conn);
     serverAssert(c->reply_bytes < SIZE_MAX - (1024 * 64));
     /* Note that c->reply_bytes is irrelevant for replica clients
      * (they use the global repl buffers). */
--- a/src/networking.c
+++ b/src/networking.c
@@ -314,7 +314,9 @@ int prepareClientToWrite(client *c) {
      * is set. */
     if (c->flag.primary && !c->flag.primary_force_reply) return C_ERR;
 
-    if (!c->conn) return C_ERR; /* Fake client for AOF loading. */
+    /* Skip the fake client, such as the fake client for AOF loading. */
+    if (c->flag.fake) return C_ERR;
+    serverAssert(c->conn);
 
     /* Schedule the client to write the output buffers to the socket, unless
      * it should already be setup to do so (it has already pending data). */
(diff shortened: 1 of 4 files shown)


src/networking.c
```diff
@@ -4437,7 +4439,7 @@ int checkClientOutputBufferLimits(client *c) {
  *
  * Returns 1 if client was (flagged) closed. */
 int closeClientOnOutputBufferLimitReached(client *c, int async) {
-    if (!c->conn) return 0; /* It is unsafe to free fake clients. */
+    if (c->flag.fake) return 0; /* It is unsafe to free fake clients. */
```
PingXie, line 4447: serverAssert on `c->conn`?

src/server.c
```diff
@@ -894,7 +894,7 @@ void updateClientMemoryUsage(client *c) {
 }
 
 int clientEvictionAllowed(client *c) {
-    if (server.maxmemory_clients == 0 || c->flag.no_evict || !c->conn) {
+    if (server.maxmemory_clients == 0 || c->flag.no_evict || c->flag.fake) {
         return 0;
     }
     int type = getClientType(c);
```
PingXie, line 901: serverAssert on `c->conn`?

PingXie (commented):
LGTM

src/networking.c
enjoy-binbin: updated.

src/server.c
hpatro: This is good practice. But how do we ensure a new dev/reviewer ensure(s) this? Wouldn't this be better asserted while setting/unsetting the flag in one place? WDYT?

PingXie: > Wouldn't this be better asserted while setting/unsetting the flag in one place?

I think we would need to create a wrapper like the below and make `flag.fake` a `client` internal state.

```c
connection *clientGetConnection(client *c) {
    serverAssert((c->flag.fake && !c->conn) || (!c->flag.fake && c->conn));
    return c->conn;
}
```

PingXie (approved):

src/server.c
enjoy-binbin: the wrapper changed will require a lot of code changes, there a LOT OF code are using c->conn

enjoy-binbin:
i think the change somehow break the CI, i will take a deep look later.

src/server.c
hpatro: @enjoy-binbin We merge this and file a new issue with `help-wanted`/`good-first-issue` tag?

enjoy-binbin: I'm not worried about not doing it myself, i just wonder if it's worth doing (diff causes conflicts).

But looking at the code again, i guess doing this wrapper is indeed a good way to avoid fake client connection issue in the future. I will do it in a later commit.

enjoy-binbin:
ok, i see the test fail because we have a createCachedResponseClient, it is a fake client but with a connection

src/server.c
enjoy-binbin: I'm starting to feel like we might still need to keep c->conn...

zuiderkwast:
@enjoy-binbin Anything missing here?

It's approved, so someone can merge it at any time. If you still want to change something, change it to draft or something else. Or merge it.

enjoy-binbin:
OK let's merge it. We can always do a followe up if needed

PR #757, opened 2024-07-08
Introduce bgsave cancel

ranshid:
In some cases bgsave child process can run for a long time exhausting system resources. Although it is possible to kill the bgsave child process from the system shell, sometimes it is not possible allowing OS level access.

This PR adds a new subcommand to the BGSAVE command.
When user will issue `BGSAVE CANCEL`
It will do one of the 2:
1. In case a bgsave child process is currently running, the child process would be immediately killed thus terminating any save/replication full sync process.
2. In case a bgsave child process is SCHEDULED to run, the scheduled execution will be cancelled.

PR #1200, opened 2024-10-21
Fix BGSAVE CANCEL since and history fields

zuiderkwast:
Fixes wrong "since" and "history" introduced in #757.

### DIFF
--- a/src/commands/bgsave.json
+++ b/src/commands/bgsave.json
@@ -12,7 +12,7 @@
                 "Added the `SCHEDULE` option."
             ],
             [
-                "8.0.0",
+                "8.1.0",
                 "Added the `CANCEL` option."
             ]
         ],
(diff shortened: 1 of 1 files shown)


src/commands/bgsave.json
```diff
@@ -37,7 +37,7 @@
                         "name": "cancel",
                         "token": "CANCEL",
                         "type": "pure-token",
-                        "since": "8.0.0"
+                        "since": "8.1.0"
```
ranshid, line 40: we should fix the history as well right?

enjoy-binbin, line 40:
```suggestion
                        "since": "8.1.0"
```

enjoy-binbin (approved):

ranshid (approved):

PR #1205, opened 2024-10-21
Remove redundant dict_index calculations

nadav-levanoni:
We need to start making use of the new `WithDictIndex` APIs which allow us to reuse the dict_index calculation (avoid over-calling `getKeySlot` for no good reason).

In this PR I optimized `lookupKey` so it now calls `getKeySlot` to reuse the dict_index two additional times.

The benefit here is reduced if the slot is cached in the client, which is generally the case.


nadav-levanoni:
Will add a follow-up PR for keys and scan commands. The impact will be greater because they do not cache the slot in the client.

xbasel (approved):
lgtm

nadav-levanoni:
I added the keysCommand optimization here. I had a cursory look at the scan command, and it might be trickier to find an optimization there.

nadav-levanoni:
Commenting for attention

madolson:
> Commenting for attention

I'm sorry I'm slow :P

madolson (approved):

PR #1208, opened 2024-10-22
Output Module config value of module list command

hwware:
Since Redis OSS 7.0, module loadex command was introduced, and we could set config for the module.  However, there is no way to display these configs when client run the module list or hello command.

This PR solves above the problem and provides module config name and value in the output as below:

image

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -2218,6 +2218,21 @@ static void numericConfigRewrite(standardConfig *config, const char *name, struc
     }
 }
 
+sds getConfigValue(sds config_name) {
+    standardConfig *config = lookupConfig(config_name);
+    switch (config->type) {
+    case BOOL_CONFIG:
+        return boolConfigGet(config);
+    case SDS_CONFIG:
+        return sdsConfigGet(config);
+    case NUMERIC_CONFIG:
+        return numericConfigGet(config);
+    case ENUM_CONFIG:
+        return enumConfigGet(config);
+    default: serverPanic("Config type of module config is not allowed.");
+    }
+}
+
 #define embedCommonNumericalConfig(name, alias, _flags, lower, upper, config_addr, default, num_conf_flags, is_valid, \
                                    apply)                                                                             \
     {                                                                                                                 \
--- a/src/module.c
+++ b/src/module.c
@@ -12356,7 +12356,7 @@ void addReplyLoadedModules(client *c) {
         sds name = dictGetKey(de);
         struct ValkeyModule *module = dictGetVal(de);
         sds path = module->loadmod->path;
-        addReplyMapLen(c, 4);
+        addReplyMapLen(c, 5);
         addReplyBulkCString(c, "name");
         addReplyBulkCBuffer(c, name, sdslen(name));
         addReplyBulkCString(c, "ver");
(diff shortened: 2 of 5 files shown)


hpatro (approved):
Do we also add history for new field in the command response?

hwware:
> Do we also add history for new field in the command response?

Looks like no this behavior

src/module.c
```diff
@@ -12368,6 +12368,20 @@ void addReplyLoadedModules(client *c) {
...
+        listNode *ln;
+        listRewind(module->module_configs, &li);
+        while ((ln = listNext(&li))) {
+            ModuleConfig *module_config = listNodeValue(ln);
+            sds config_name = sdscatfmt(sdsempty(), "%s.%s", module->name, module_config->name);
+            addReplyBulkCBuffer(c, config_name, sdslen(config_name));
+            sds config_value = getConfigValue(config_name);
+            addReplyBulkCBuffer(c, config_value, sdslen(config_value));
+            sdsfree(config_name);
+            sdsfree(config_value);
```
enjoy-binbin, line 12383: we can use `addReplySds` so that we can save this sdsfree lines

src/config.c
```diff
@@ -2218,6 +2218,21 @@ static void numericConfigRewrite(standardConfig *config, const char *name, struc
...
+    switch (config->type) {
+    case BOOL_CONFIG:
+        return boolConfigGet(config);
+    case SDS_CONFIG:
+        return sdsConfigGet(config);
+    case NUMERIC_CONFIG:
+        return numericConfigGet(config);
+    case ENUM_CONFIG:
+        return enumConfigGet(config);
+    default: serverPanic("Config type of module config is not allowed.");
```
enjoy-binbin, line 2232: we don't support the other type? like STRING_CONFIG or SPECIAL_CONFIG, i guess SPECIAL_CONFIG is a special one, but STRING_CONFIG?

enjoy-binbin (commented):
i don't know the details very much, it needs other one to take a look too.

src/config.c
hwware: STRING_CONFIG is not supported so far by origin design, I have no idea the reason.

madolson:
Can't you do the same thing as `config get "moduleconfig.*"` ?

hwware:
> Can't you do the same thing as `config get "moduleconfig.*"` ?

I tested with "config get moduleconfig.*", and got the same result as below
image

But this information belongs to a module (module list command and hello command), 
Do you want to show above information in above 2 commands?

madolson:
> Do you want to show above information in above 2 commands?

What is the usecase for showing the config information in the other two commands? Showing the argv/argc makes sense since there is no other way to get that information. For configs, I feel like using the `CONFIG` commands make more sense.

hwware:
> > Do you want to show above information in above 2 commands?
> 
> What is the usecase for showing the config information in the other two commands? Showing the argv/argc makes sense since there is no other way to get that information. For configs, I feel like using the `CONFIG` commands make more sense.

In fact, there is no usecase for showing the config information for the module. I just think both the args and config information belong to the module properties.  If you think existing behavior works properly, I will close this pr.  Thanks

hwware:
We could get the same result through command config get "moduleconfig.*", let me close it.

PR #1209, opened 2024-10-22
RDMA builtin support

pizhenwei:
Hi,

There are several patches in this PR:
* Abstract set/rewrite config bind option: `bind` option is a special config, `socket` and `tls` are using the same one. However RDMA uses the similar style but different one. Use a bit abstract work to make it flexible for both `socket` and `RDMA`. Even for QUIC in the future.
* Introduce closeListener for connection type: closing socket by a simple syscall would be fine, RDMA has complex logic. Introduce connection type specific close listener method.
* RDMA: Use valkey.conf style instead of module parameters: use `--rdma-bind` and `--rdma-port` style instead of module parameters. The module style config `rdma.bind` and `rdma.port` are removed.
* RDMA: Support builtin: support `make BUILD_RDMA=yes`. module style is still kept for now. Once module style is decided to drop, I volunteer to do it for TLS and RDMA together.

Each patch has independent functionality, also been tested by CI and local commands, so request no-squash merge for this PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,8 +37,13 @@ To build TLS as Valkey module:
 Note that sentinel mode does not support TLS module.
 
 To build with experimental RDMA support you'll need RDMA development libraries
-(e.g. librdmacm-dev and libibverbs-dev on Debian/Ubuntu). For now, Valkey only
-supports RDMA as connection module mode. Run:
+(e.g. librdmacm-dev and libibverbs-dev on Debian/Ubuntu).
+
+To build RDMA support as Valkey built-in:
+
+    % make BUILD_RDMA=yes
+
+To build RDMA as Valkey module:
 
     % make BUILD_RDMA=module
 
@@ -203,20 +208,27 @@ Note that Valkey Over RDMA is an experimental feature.
 It may be changed or removed in any minor or major version.
 Currently, it is only supported on Linux.
 
-To manually run a Valkey server with RDMA mode:
+* RDMA built-in mode:
+    ```
+    ./src/valkey-server --protected-mode no \
+         --rdma-bind 192.168.122.100 --rdma-port 6379
+    ```
 
-    % ./src/valkey-server --protected-mode no \
-         --loadmodule src/valkey-rdma.so bind=192.168.122.100 port=6379
+* RDMA module mode:
+    ```
+    ./src/valkey-server --protected-mode no \
+         --loadmodule src/valkey-rdma.so --rdma-bind 192.168.122.100 --rdma-port 6379
+    ```
 
 It's possible to change bind address/port of RDMA by runtime command:
 
-    192.168.122.100:6379> CONFIG SET rdma.port 6380
+    192.168.122.100:6379> CONFIG SET rdma-port 6380
 
 It's also possible to have both RDMA and TCP available, and there is no
 conflict of TCP(6379) and RDMA(6379), Ex:
 
     % ./src/valkey-server --protected-mode no \
-         --loadmodule src/valkey-rdma.so bind=192.168.122.100 port=6379 \
+         --loadmodule src/valkey-rdma.so --rdma-bind 192.168.122.100 --rdma-port 6379 \
          --port 6379
 
 Note that the network card (192.168.122.100 of this example) should support
--- a/src/config.c
+++ b/src/config.c
@@ -1536,10 +1536,27 @@ void rewriteConfigOOMScoreAdjValuesOption(standardConfig *config, const char *na
 }
 
 /* Rewrite the bind option. */
-void rewriteConfigBindOption(standardConfig *config, const char *name, struct rewriteConfigState *state) {
+static void rewriteConfigBindOption(standardConfig *config, const char *name, struct rewriteConfigState *state, char **bindaddr, int bindaddr_count) {
     UNUSED(config);
     int force = 1;
     sds line, addresses;
+
+    /* Rewrite as bind <addr1> <addr2> ... <addrN> */
+    if (bindaddr_count > 0)
+        addresses = sdsjoin(bindaddr, bindaddr_count, " ");
+    else
+        addresses = sdsnew("\"\"");
+    line = sdsnew(name);
+    line = sdscatlen(line, " ", 1);
+    line = sdscatsds(line, addresses);
+    sdsfree(addresses);
+
+    rewriteConfigRewriteLine(state, name, line, force);
+}
+
+/* Rewrite the bind option. */
+static void rewriteConfigSocketBindOption(standardConfig *config, const char *name, struct rewriteConfigState *state) {
+    UNUSED(config);
     int is_default = 0;
 
     /* Compare server.bindaddr with CONFIG_DEFAULT_BINDADDR */
@@ -1559,17 +1576,7 @@ void rewriteConfigBindOption(standardConfig *config, const char *name, struct re
         return;
     }
 
-    /* Rewrite as bind <addr1> <addr2> ... <addrN> */
-    if (server.bindaddr_count > 0)
-        addresses = sdsjoin(server.bindaddr, server.bindaddr_count, " ");
-    else
-        addresses = sdsnew("\"\"");
-    line = sdsnew(name);
-    line = sdscatlen(line, " ", 1);
-    line = sdscatsds(line, addresses);
-    sdsfree(addresses);
-
-    rewriteConfigRewriteLine(state, name, line, force);
+    rewriteConfigBindOption(config, name, state, server.bindaddr, server.bindaddr_count);
 }
 
 /* Rewrite the loadmodule option. */
@@ -2637,7 +2644,7 @@ static int applyBind(const char **err) {
     tcp_listener->ct = connectionByType(CONN_TYPE_SOCKET);
     if (changeListener(tcp_listener) == C_ERR) {
         *err = "Failed to bind to specified addresses.";
-        if (tls_listener) closeListener(tls_listener); /* failed with TLS together */
+        if (tls_listener) connCloseListener(tls_listener); /* failed with TLS together */
         return 0;
     }
 
@@ -2649,7 +2656,7 @@ static int applyBind(const char **err) {
         tls_listener->ct = connectionByType(CONN_TYPE_TLS);
         if (changeListener(tls_listener) == C_ERR) {
             *err = "Failed to bind to specified addresses.";
-            closeListener(tcp_listener); /* failed with TCP together */
+            connCloseListener(tcp_listener); /* failed with TCP together */
             return 0;
         }
     }
@@ -2922,8 +2929,9 @@ static sds getConfigNotifyKeyspaceEventsOption(standardConfig *config) {
     return keyspaceEventsFlagsToString(server.notify_keyspace_events);
 }
 
-static int setConfigBindOption(standardConfig *config, sds *argv, int argc, const char **err) {
+static int setConfigBindOption(standardConfig *config, sds *argv, int argc, const char **err, char **bindaddr, int *bindaddr_count) {
     UNUSED(config);
+    int orig_bindaddr_count = *bindaddr_count;
     int j;
 
     if (argc > CONFIG_BINDADDR_MAX) {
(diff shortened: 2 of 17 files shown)


zuiderkwast:
@valkey-io/core-team Please approve.

This adds `make BUILD_RDMA=yes`, so just like TLS it can be built as a module or as builtin. Just like for TLS, the default is `no`.

It is a step towards official (non-experimantal) RDMA support.

I already merged the RDMA fork-safety here: #1244. It makes sure the server doesn't crash if we try to talk to an RDMA client from a forked child process.

pizhenwei:
> @valkey-io/core-team Please approve.
> 
> This adds `make BUILD_RDMA=yes`, so just like TLS it can be built as a module or as builtin. Just like for TLS, the default is `no`.
> 
> It is a step towards official (non-experimantal) RDMA support.
> 
> I already merged the RDMA fork-safety here: #1244. It makes sure the server doesn't crash if we try to talk to an RDMA client from a forked child process.

Hi @valkey-io/core-team
The [CMake building support](https://github.com/valkey-io/valkey/pull/1196) has been merged, then we can drop module build option for TLS and RDMA together.

PingXie:
> The CMake building support has been merged, then we can drop module build option for TLS and RDMA together.

I vote for dropping the module packaging. Modularity is what we need and we have it here for RDMA.

zuiderkwast:
I believe it's safe to drop the RDMA module because it's experimental, but I'm not sure about the TLS module. Maybe someone is using it? Maybe we should drop it in a major version (9.0?).

The potential benefit of module is that you don't need OpenSSL installed to start Valkey, as long as you don't load the module. If we ship binary releases or containers with bundled modules, they can be used also by those who don't have or want TLS. Examples are embedded system, IoT devices, etc. We've already talked about packaging binaries with buldled modules like (bloom, json).

valkey.conf
```diff
@@ -300,6 +300,26 @@ tcp-keepalive 300
...
+# By default, RDMA is disabled. To enable it, the "rdma-port" configuration
+# directive can be used to define RDMA-listening ports. Note that RDMA is able
+# to build as builtin or module. To enable RDMA on the port, use:
+#
+# rdma-port 6379
+# rdma-bind 192.168.1.100
+
+# The RDMA receive transfer buffer is 1M, it can be set between 64K and 16M.
+#
+# rdma-rx-size 262144
```
zuiderkwast: Mention the word "default" and use the default value in the example.

```suggestion
# The RDMA receive transfer buffer is 1M by default. It can be set between 64K and 16M.
#
# rdma-rx-size 1048576
```

valkey.conf
```diff
@@ -300,6 +300,26 @@ tcp-keepalive 300
...
+
+# The RDMA completion queue will use the completion vector for signaling completion
+# events. Note that it must be at least zero and less than the maximum number of
+# completion vectors from the RDMA device. Once a negative number (-1) is specified,
+# the server will use a random vector for a new connection. The default vector is -1.
+#
+# rdma-comp-vector 0
```
zuiderkwast: "it must be at least zero" and "The default vector is -1" are contradicting. Very confusing. :smile: Can you fix the text so it's not contradicting itself?

When/why would you want to use a random/non-random vector?

valkey.conf
```diff
@@ -300,6 +300,26 @@ tcp-keepalive 300
 #
 # tls-session-cache-timeout 60
 
+################################### RDMA ######################################
+
+# By default, RDMA is disabled. To enable it, the "rdma-port" configuration
+# directive can be used to define RDMA-listening ports. Note that RDMA is able
+# to build as builtin or module. To enable RDMA on the port, use:
```
zuiderkwast: It's not important how RDMA is built. It's only important that it *is* built.

But I don't think we need to mention it. TLS build is optional too and we don't mention that for the TLS configs.

```suggestion
# directive can be used to define RDMA-listening ports. To enable RDMA on the
# port, use:
```

zuiderkwast (commented):
I didn't review fully. Just the interface changes.

valkey.conf
pizhenwei: Sure, add more information about system interrupt pinning, RDMA completion vector and valkey-server cpu affinity to tune performance. Please take a look at the latest version. Thanks!

pizhenwei:
Hi @zuiderkwast 
Please take a loot at the latest version, modify as your suggestion, and support builtin for cmake.

valkey.conf
```diff
@@ -300,6 +300,51 @@ tcp-keepalive 300
...
+# Example 1. a) Pin hardware interrupt vectors [0, 3] to CPU [0, 3].
+#            b) Set CPU affinity for valkey to CPU [4, X].
+#            c) Any valkey server uses a random RDMA completion vector.
+# All valkey servers will not affect each other and will be isolated from kernel interrupts.
+#
+#   SYS    SYS    SYS    SYS  VALKEY VALKEY     VALKEY
+#    |      |      |      |      |      |          |
+#  CPU0   CPU1   CPU2   CPU3   CPU4   CPU5   ... CPUX
+#    |      |      |      |
+#  INTR0  INTR1  INTR2  INTR3
```
zuiderkwast: Fascinating.

We should have this information also on the RDMA.md documentation page.

zuiderkwast (approved):
LGTM.

Let's wait for a core team majority decision before  merge since it's an interface change.

Next time, can you avoid force-push and instead just push more commits? It's easier to see what's changed since last review. We squash-merge in the end anyway, normally.

valkey.conf
pizhenwei: Sure, glad to do it.

pizhenwei:
> Next time, can you avoid force-push and instead just push more commits? It's easier to see what's changed since last review. We squash-merge in the end anyway, normally.

OK.

zuiderkwast:
@valkey-io/core-team This adds `make BUILD_RDMA=yes` and some configs. Please approve or vote :+1: or :-1:.

The module style configs `rdma.port` and `rdma.bind` are removed. Even if BUILD_RDMA=module, the module is using the new, normal configs. That's why it is a breaking change, but we allow it because RDMA is experimental.

README.md
```diff
@@ -203,20 +208,27 @@ Note that Valkey Over RDMA is an experimental feature.
...
-         --loadmodule src/valkey-rdma.so bind=192.168.122.100 port=6379
+* RDMA module mode:
+    ```
+    ./src/valkey-server --protected-mode no \
+         --loadmodule src/valkey-rdma.so --rdma-bind 192.168.122.100 --rdma-port 6379
+    ```
 
 It's possible to change bind address/port of RDMA by runtime command:
 
-    192.168.122.100:6379> CONFIG SET rdma.port 6380
```
enjoy-binbin, line 213: so this just a typo?

src/config.c
```diff
@@ -2935,11 +2943,77 @@ static int setConfigBindOption(standardConfig *config, sds *argv, int argc, cons
...
+    }
+    for (j = 0; j < argc; j++) bindaddr[j] = zstrdup(argv[j]);
+    *bindaddr_count = argc;
+
+    return 1;
+}
+
+static int setConfigSocketBindOption(standardConfig *config, sds *argv, int argc, const char **err) {
+    UNUSED(config);
+
```
enjoy-binbin:
```suggestion
```

src/config.c
```diff
@@ -2935,11 +2943,77 @@ static int setConfigBindOption(standardConfig *config, sds *argv, int argc, cons
...
+    return setConfigBindOption(config, argv, argc, err, server.bindaddr, &server.bindaddr_count);
+}
+
+static int setConfigRdmaBindOption(standardConfig *config, sds *argv, int argc, const char **err) {
+    UNUSED(config);
+
```
enjoy-binbin:
```suggestion
```

src/config.c
```diff
@@ -2935,11 +2943,77 @@ static int setConfigBindOption(standardConfig *config, sds *argv, int argc, cons
     if (argc == 1 && sdslen(argv[0]) == 0) argc = 0;
 
     /* Free old bind addresses */
-    for (j = 0; j < server.bindaddr_count; j++) {
-        zfree(server.bindaddr[j]);
+    for (j = 0; j < orig_bindaddr_count; j++) {
+        zfree(bindaddr[j]);
+    }
```
enjoy-binbin: not that important, just one is three lines and one is one line.
```suggestion
    for (j = 0; j < orig_bindaddr_count; j++) zfree(bindaddr[j]);
```

enjoy-binbin (approved):
Not a full review, but LGTM.

README.md
pizhenwei: Previously, this parameter is generated automatically by valkey. valkey server combines module name 'rdma' and module parameter 'port' into 'rdma.port'. Now, we use valkey server parameter 'rdma-port' instead.

zuiderkwast: `rdma.port` was the module-defined config, which is removed now because there are normal configs instead.

Before this PR, the RDMA module created module configs in OnLoad. Now, even if RDMA is built as a module, it is using the real configs instead of module-defined config.

The removed module configs is a breaking change, but RDMA is experimental, so it's no problem.

enjoy-binbin: so, now if we are using module mode, the old rdma.port is not valid anymore? Yes, it is a breaking change and we should mention it in the release notes (thought it is experimental)

pizhenwei: > so, now if we are using module mode, the old rdma.port is not valid anymore? Yes, it is a breaking change and we should mention it in the release notes (thought it is experimental)

Right. `rdma.port` is totally removed.

zuiderkwast: > Yes, it is a breaking change and we should mention it in the release notes (thought it is experimental)

It doesn't look good that we have a section of breaking changes in the minor version. We have to write it in a smart way so that it doesn't look bad.

madolson: I would just not list it as a breaking change. I would just call it out as a change in the RDMA code, which is still experimental.

madolson: I guess alternatively, can we make rdma-port an alias for rdma.port?

valkey.conf
```diff
@@ -300,6 +300,51 @@ tcp-keepalive 300
 #
 # tls-session-cache-timeout 60
 
+################################### RDMA ######################################
```
madolson, line 303: We should list this as experimental, like we used to do for cluster mode.

valkey.conf
```diff
@@ -300,6 +300,51 @@ tcp-keepalive 300
...
+# VALKEY VALKEY VALKEY VALKEY VALKEY VALKEY     VALKEY
+#    |      |      |      |      |      |          |
+#  CPU0   CPU1   CPU2   CPU3   CPU4   CPU5  ...  CPUX
+#    |      |      |      |      |      |          |
+#  INTR0  INTR1  INTR2  INTR3  INTR4  INTR5      INTRX
+#
+# Use 0 and positive numbers to specify the RDMA completion vector, or specify -1 to allow
+# the server to use a random vector for a new connection. The default vector is -1.
+#
+# rdma-comp-vector 0
```
```suggestion
# rdma-completion-vector 0
```
Any reason not to be more verbose here? I don't think comp -> completion is a common abbreviation.

I also don't really understand what value you are setting here. If I set 3, am I telling it to only using CPU 3 (or cpus 0 and 1 because that is the bit mask.)

madolson (commented):
Directionally I'm inclined, I mostly just had some questions about the configs and documentation.

README.md
pizhenwei: We have already use `--rdma-bind 192.168.122.100 --rdma-port 6379` instead of `--loadmodule src/valkey-rdma.so port=6379 bind=192.168.122.1` since this change, should we (and other few users in use) really care this dynamic config...

valkey.conf
pizhenwei: `comp-vector` is same as [`struct ibv_cq *ibv_create_cq(struct ibv_context *context, int cqe, void *cq_context,  struct ibv_comp_channel *channel,  int comp_vector);`](https://man7.org/linux/man-pages/man3/ibv_create_cq.3.html), this may be friendly to guys  who are familiar with RDMA programming. This is the reason I use `rdma-comp-vector`.

From valkey user side, which one do you prefer? Should I link this man page here?

README.md
madolson: Yeah, that's true. I'm okay with this as is, I just thought I would bring it up.

valkey.conf
madolson:
```
The CQ will use the completion vector comp_vector for signaling completion events
```

Leads me to believe we should use completion-vector, since it seems like the human friendly name. We don't normally abbreviate config names. comp_vector seems like the prototype is just trying to compress it (which might make sense in the context of rdma developers since they will write it a lot).

madolson (approved):

README.md
zuiderkwast: If RDMA would not be considered experimental, then we would have had to care about this. It's great that we called it experimental. :smile:

zuiderkwast: > I would just not list it as a breaking change. I would just call it out as a change in the RDMA code, which is still experimental.

I removed the breaking-change label again.

pizhenwei:
Hi, @madolson @zuiderkwast 
Append a new commit to fix conflict against 33f42d7fb597ce28040f184ee57ed86d6f6ffbd8. Please take a look at the latest version. Thanks!

zuiderkwast:
@pizhenwei The DCO is failing because some commits are written by some email and signed off with another email. Please fix it. Feel free to squash all commits to one if you want. See DCO job Details: https://github.com/valkey-io/valkey/pull/1209/checks?check_run_id=33460190673

@PingXie @soloestoy @hwware Can we have one more ACK to merge this one? (RDMA is still experimental so it's unclear if it even needs to be a major decision, but it's marked as one now.)

hwware:
> @valkey-io/core-team Please approve.
> 
> This adds `make BUILD_RDMA=yes`, so just like TLS it can be built as a module or as builtin. Just like for TLS, the default is `no`.
> 
> It is a step towards official (non-experimantal) RDMA support.
> 
> I already merged the RDMA fork-safety here: #1244. It makes sure the server doesn't crash if we try to talk to an RDMA client from a forked child process.

Yes, it is the way I recommend. Give user one more option.  Thanks

pizhenwei:
Because of DCO check error, rebase against the latest code and squash into a single commit.

PR #1211, opened 2024-10-23
Cleaned up getSlotOrReply is return -1 instead of C_ERR

enjoy-binbin:
Minor cleanup since getSlotOrReply return -1 on error, not return C_ERR.


zuiderkwast (approved):
Thanks. C_ERR is defined to -1, but it has a different meaning.

PR #1212, opened 2024-10-23
Do security attack check only when command not found to reduce the critical path.

zhulipeng:
### Description

When explored the cycles distribution for main thread with io-threads enabled. We found this security attack check takes significant time in main thread, **~3%** cycles were used to do the commands security check in main thread.

This patch try to completely avoid doing it in the hot path. We can do it only after we looked up the command and it wasn't found, just before we call commandCheckExistence.

### Profiling Data
For example, **Fig 1** demonstrate the cycles distribution of main thread of  benchmark [memtier_benchmark-1Mkeys-load-string-with-1KiB-values.yml]( https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-load-string-with-1KiB-values.yml): `taskset -c 40-55 memtier_benchmark -s 127.0.0.1 -p 9001 "--data-size" "1000" --ratio 1:0 --key-pattern P:P --key-minimum=1 --key-maximum 1000000 --test-time 180 -c 50 -t 16 --hide-histogram`. 

image
**Fig 1. Cycles distribution of main thread.**

image
**Fig 2. Annotate `processCommand`**

### Performance Boost

Didn't run a full list, just pick some benchmarks to validate the optimization works. 

|Benchmark|Performance Boost|
|-|-|
|[memtier_benchmark-1Mkeys-string-get-10B](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-string-get-10B.yml)|**2%**|
|[memtier_benchmark-1Mkeys-hash-hget-hgetall-hkeys-hvals-with-100B-values](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-hash-hget-hgetall-hkeys-hvals-with-100B-values.yml)|**5%**|
|[memtier_benchmark-1Mkeys-load-string-with-1KiB-values](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-load-string-with-1KiB-values.yml)|**3%**|

#### Test Env
- OS: Ubuntu 22.04.4 LTS
- Platform: Intel Xeon Platinum 8380

```
taskset -c 0-15 ~/valkey/src/valkey-server ~/tmp_valkey.conf
port 9001
bind * -::*
daemonize no
protected-mode no
save
io-threads 16
```

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -167,6 +167,7 @@ client *createClient(connection *conn) {
     c->nread = 0;
     c->read_flags = 0;
     c->write_flags = 0;
+    c->io_checked = 0;
     c->cmd = c->lastcmd = c->realcmd = c->io_parsed_cmd = NULL;
     c->cur_script = NULL;
     c->multibulklen = 0;
@@ -4830,6 +4831,13 @@ void ioThreadReadQueryFromClient(void *data) {
     /* Empty command - Multibulk processing could see a <= 0 length. */
     if (c->argc == 0) {
         goto done;
+    }
+
+    /* Handle possible security attacks. */
+    if (!strcasecmp(c->argv[0]->ptr, "host:") || !strcasecmp(c->argv[0]->ptr, "post")) {
+        c->io_checked = IO_CHECKED_NOT_SECURITY;
+    } else {
+        c->io_checked = IO_CHECKED_SECURITY;
     }
 
     /* Lookup command offload */
--- a/src/networking.c
+++ b/src/networking.c
@@ -167,7 +167,6 @@ client *createClient(connection *conn) {
     c->nread = 0;
     c->read_flags = 0;
     c->write_flags = 0;
-    c->io_checked = 0;
     c->cmd = c->lastcmd = c->realcmd = c->io_parsed_cmd = NULL;
     c->cur_script = NULL;
     c->multibulklen = 0;
@@ -4831,13 +4830,6 @@ void ioThreadReadQueryFromClient(void *data) {
     /* Empty command - Multibulk processing could see a <= 0 length. */
     if (c->argc == 0) {
         goto done;
-    }
-
-    /* Handle possible security attacks. */
-    if (!strcasecmp(c->argv[0]->ptr, "host:") || !strcasecmp(c->argv[0]->ptr, "post")) {
-        c->io_checked = IO_CHECKED_NOT_SECURITY;
-    } else {
-        c->io_checked = IO_CHECKED_SECURITY;
     }
 
     /* Lookup command offload */
--- a/src/server.c
+++ b/src/server.c
@@ -3829,7 +3829,8 @@ int processCommand(client *c) {
     }
 
     /* Handle possible security attacks. */
-    if (!strcasecmp(c->argv[0]->ptr, "host:") || !strcasecmp(c->argv[0]->ptr, "post")) {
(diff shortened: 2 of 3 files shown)


zuiderkwast:
Good finding that this check takes significant time.

I wonder if we can completely avoid doing it in the hot path. Can we do it only after we looked up the command and it wasn't found, just before we call commandCheckExistence?

Then we avoid it completely for normal traffic also without io threads.

zhulipeng:
> Good finding that this check takes significant time.
> 
> I wonder if we can completely avoid doing it in the hot path. Can we do it only after we looked up the command and it wasn't found, just before we call commandCheckExistence?
> 
> Then we avoid it completely for normal traffic also without io threads.

@zuiderkwast Got your point, check the potential security attacks only when command not found. 

```diff
diff --git a/src/server.c b/src/server.c
index ab95f8434..d62c3a09f 100644
--- a/src/server.c
+++ b/src/server.c
@@ -3828,12 +3828,6 @@ int processCommand(client *c) {
         reqresAppendRequest(c);
     }

-    /* Handle possible security attacks. */
-    if (!strcasecmp(c->argv[0]->ptr, "host:") || !strcasecmp(c->argv[0]->ptr, "post")) {
-        securityWarningCommand(c);
-        return C_ERR;
-    }
-
     /* If we're inside a module blocked context yielding that wants to avoid
      * processing clients, postpone the command. */
     if (server.busy_module_yield_flags != BUSY_MODULE_YIELD_NONE &&
@@ -3848,6 +3842,13 @@ int processCommand(client *c) {
      * we do not have to repeat the same checks */
     if (!client_reprocessing_command) {
         struct serverCommand *cmd = c->io_parsed_cmd ? c->io_parsed_cmd : lookupCommand(c->argv, c->argc);
+        if (!cmd) {
+            /* Handle possible security attacks. */
+            if (!strcasecmp(c->argv[0]->ptr, "host:") || !strcasecmp(c->argv[0]->ptr, "post")) {
+                securityWarningCommand(c);
+                return C_ERR;
+            }
+        }
         c->cmd = c->lastcmd = c->realcmd = cmd;
         sds err;
         if (!commandCheckExistence(c, &err)) {

```

zuiderkwast (approved):
Yes, exactly!

Can you update the PR title and description?

zhulipeng:
> Yes, exactly!
> 
> Can you update the PR title and description?

Changed.

zuiderkwast:
Thanks. And the description = the first comment? We normally use that as the commit message when we merge. It still mentions "We found some logic in main thread can be offloaded to io-threads to mitigate the bottleneck of main thread to gain the performance." 
Do you get the same benchmark result with the new change?

zhulipeng:
> Thanks. And the description = the first comment? We normally use that as the commit message when we merge. It still mentions "We found some logic in main thread can be offloaded to io-threads to mitigate the bottleneck of main thread to gain the performance." 

Updated, please review.

> Do you get the same benchmark result with the new change?

Yes, performance is better, I just update the perf boost data for the listed 3 benchmarks.

zuiderkwast:
Perfect, thanks!

PR #1186, opened 2024-10-17
New hash table

zuiderkwast:
This PR implements a new hash table and uses it for keys, command lookup and more. There are multiple commits. Do not squash-merge.

1. Hash table implementation and unit tests
2. Use new hash table for command lookup
3. Use new hash table keys and expire

Hash table design
-----------------

The hash table is a cache line optimized and implemented as outlined in #169, but changed to chaining instead of probing after an idea by Madelyn. The key-value entry is user-defined, which allows the user to embed key and value within a single allocation. The hash table supports incremental rehashing, scan, random key, etc. just like the dict but faster and using less memory. Each bucket contains a few bits of metadata per entry. For details, see the comments in `src/hashtable.{c,h}`.

If a bucket is full, the last entry pointer in a bucket can be replaced by a child-bucket pointer and we get a bucket chain.

```
      Bucket          Bucket          Bucket               
-----+---------------+---------------+---------------+-----
 ... | x x x x x x p | x x x x x x x | x x x x x x x | ... 
-----+-------------|-+---------------+---------------+-----
                   |                                       
                   v  Child bucket                         
                 +---------------+                         
                 | x x x x x x p |                         
                 +-------------|-+                         
                               |                           
                               v  Child bucket             
                             +---------------+             
                             | x x x x x x x |             
                             +---------------+             
```

Command lookup
--------------

The 2nd commit relaces dict with hashtable for command lookup. This was implemented by @SoftlyRaining.

Keys and expire
---------------

The 3rd commit replaces dict with the new hash table in kvstore.c and all code that uses it, such as db.c.

The hashtable entry in this case is the robj struct. The key and optionally an expire timestamp are embedded in the `robj` struct, i.e. the key is embedded in the value. Therefore, we can call this a *valkey* object, val + key. This design saves roughly 20 bytes per key for short string keys.

```
 keys                                      expire
hashtable             robj                hashtable
 +---+         +-------------------+        +---+
 | 1 --------->| type, enc, lru    |<-------- 1 |
 | 2 |         | refcount, flags   |        | 2 |
 | 3 |         | ptr               |        | 3 |
 | . |         | [expire]          |        | . |
 | . |         | embedded key      |        | . |
 | . |         | [embedded value]  |        | . |
 +---+         +-------------------+        +---+

[.] = optional
```

Some db.c functions like `dbAdd`, `setKey` and `setExpire` now reallocate the value object to embed the key and optional expire in it. `setKey` does not increment the reference counter, since it would require duplicating the object.

[...]

PR #1217, opened 2024-10-23
[hashset feature] Scan shortcut fix

rainsupreme:
This is targeting the hashset branch and will update the hashset PR when merged. This fixes the issue I saw where scan can miss elements when we're rehashing: https://github.com/valkey-io/valkey/pull/1186#discussion_r1809764417

The UT fails without my fix, and passes with my fix 😁


src/hashset.c
```diff
@@ -1651,24 +1673,32 @@ void hashsetDump(hashset *s) {
...
+        size_t mask = expToMask(s->bucket_exp[table]);
+        size_t cursor = 0;
+        do {
+            bucket *b = &s->tables[table][cursor];
             char c = b->everfull ? 'X' : 'o';
             printf("%c", c);
-        }
+
+            cursor = nextCursor(cursor, mask);
+        } while (cursor != 0);
```
rainsupreme, line 1717: Only debug print functions for human eyes. I prefer to see it in rehash/scan/iteration/probe order.

src/hashset.c
```diff
@@ -438,48 +438,61 @@ static inline int cursorIsLessThan(size_t a, size_t b) {
     return rev(a) < rev(b);
 }
 
-/* Rehashes one bucket. */
+/* Rehashes buckets, following potential probe chains. This means that between
+ * rehash steps, any element that hashes to an already-rehashed bucket has been
+ * rehashed, even if it was moved to a different bucket due to open addressing.
+ * Knowing this allows scan to skip already-rehashed buckets. */
 static void rehashStep(hashset *s) {
```
zuiderkwast: With an added `while` and `if (b->presence)`, this function now gets 5 levels of indentation. It's too much. :) Nested blocks increase the mental complexity.

I have an idea:

```C
static void rehashStep(hashset *s) {
    while (rehashBucket(s));
}

static int rehashBucket(hashset *s) {
    /* All of the old content of rehashStep */
    ...
    /* In the end, return 1 if we're in a probe chain, 0 if we can stop. */
    if (s->rehash_idx == 0) {
        rehashingCompleted(s);
        return 0;
    } else {
        return b->everfull;
    }
}
```

src/hashset.c
```diff
@@ -438,48 +438,61 @@ static inline int cursorIsLessThan(size_t a, size_t b) {
...
-        /* When shrinking, it's possible to avoid computing the hash. We can
-         * just use idx has the hash, but only if we know that probing didn't
-         * push this element away from its primary bucket, so only if the
-         * bucket before the current one hasn't ever been full. */
-        if (s->bucket_exp[1] <= s->bucket_exp[0] &&
-            !s->tables[0][prevCursor(idx, expToMask(s->bucket_exp[0]))].everfull) {
-            hash = idx;
-        } else {
-            hash = hashElement(s, element);
+    uint8_t prev_bucket_everfull = s->tables[0][prevCursor(idx, expToMask(s->bucket_exp[0]))].everfull;
```
zuiderkwast: Computing this requires a memory access that's not useful in the growing case. Growing is normally the most performance sensitive situation.

My guess is that it's cheaper to check it only when we need it...

    if (s->bucket_exp[1] <= s->bucket_exp[0] && ...here...)

On the other hand, the previous bucket has most likely been rehashed just before, so it's in the L1 cache already. Anyway, I'm not sure if one is better than the other.

src/hashset.c
```diff
@@ -1260,10 +1273,13 @@ size_t hashsetScan(hashset *s, size_t cursor, hashsetScanFunction fn, void *priv
...
 
-    /* Mask the start cursor to the bigger of the tables, so we can detect if we
+    /* Mask the start cursor to the smaller of the tables, so we can detect if we
      * come back to the start cursor and break the loop. It can happen if enough
      * tombstones (in both tables while rehashing) make us continue scanning. */
-    cursor = cursor & (expToMask(s->bucket_exp[0]) | expToMask(s->bucket_exp[1]));
+    cursor &= expToMask(s->bucket_exp[0]);
+    if (hashsetIsRehashing(s)) {
+        cursor &= expToMask(s->bucket_exp[1]);
+    }
```
zuiderkwast: With this change, I guess there's a small increased risk of returning duplicates.

Time 1:

* Table size: 16 buckets.
* Scan order: 0 -> 8 -> 4 -> 12 -> 2 -> 10 -> 6 -> 14 -> 1 -> ... -> 15. 
* `SCAN 0` returned new cursor 8.

Time 2:
* Table sizes 16 and 2. Shrinking started. Scan order: 0 -> 1.
* User continues previous scan with `SCAN 8`.
* Old implementation: `SCAN 8` returns elements from small table bucket 0 (masked with small table mask) and large table bucket 8 (masked with large table mask). Large table bucket 0 can safely be skipped.
* New implementation: `SCAN 8` returns elements from small table bucket 0 (masked with small table mask) and large table bucket 0 and 8 (expansion of cursor masked with small table mask).
* Both implementaions: Return new cursor 1.

src/hashset.c
```diff
@@ -1297,9 +1313,9 @@ size_t hashsetScan(hashset *s, size_t cursor, hashsetScanFunction fn, void *priv
             size_t mask_small = expToMask(s->bucket_exp[table_small]);
             size_t mask_large = expToMask(s->bucket_exp[table_large]);
 
-            /* Emit elements in the smaller table, if this bucket hasn't already
-             * been rehashed. */
-            if (table_small == 0 && !cursorIsLessThan(cursor, s->rehash_idx)) {
+            /* Emit elements in the smaller table if it's the new table or if
+             * this bucket hasn't been rehashed yet. */
+            if (table_small == 1 || !cursorIsLessThan(cursor, s->rehash_idx)) {
```
zuiderkwast: The comment is repeating the same logic if the `if` condition. Logic is often easier to understand if it's explained in multiple different ways.

Maybe focus the comment on when it's safe to skip the bucket?

```suggestion
            /* Emit elements in the smaller table. Skip it if it's the old
             * table and this bucket has already been rehashed. */
            if (table_small == 1 || !cursorIsLessThan(cursor, s->rehash_idx)) {
```

src/hashset.c
```diff
@@ -438,48 +438,61 @@ static inline int cursorIsLessThan(size_t a, size_t b) {
...
-         * bucket before the current one hasn't ever been full. */
-        if (s->bucket_exp[1] <= s->bucket_exp[0] &&
-            !s->tables[0][prevCursor(idx, expToMask(s->bucket_exp[0]))].everfull) {
-            hash = idx;
-        } else {
-            hash = hashElement(s, element);
+    uint8_t prev_bucket_everfull = s->tables[0][prevCursor(idx, expToMask(s->bucket_exp[0]))].everfull;
+    while (1) {
+        bucket *b = &s->tables[0][idx];
+        if (b->presence) {
```
zuiderkwast: To avoid adding a big indented block, we could use goto.

```suggestion
        if (!b->presence) goto bucket_done;
```

zuiderkwast (commented):
Finally you convinced me that the current logic has a problem!

This PR solves the issue indeed. I'm inclined to merge it. I just posted some suggestions.

Letting rehash follow probing chains can add a lot of latency to individual commands though. I'm not sure that's worse than adding very little latency to each scan call.

Some context: When debugging a test case of AOF rewrite, I observed probing chains of length 6000. This is when reading an AOF file. Actually I believe all non-empty buckets of the table were in the same probing chain.

Reading an AOF file means reading commands and executing them. The AOF file create by AOF rewrite has the commands stored per key, and the keys are in iterator order = probing order. Do you see the problem?

All commands that hash to the same bucket in a large table come together in the beginning of the AOF file, creating huge probing chains in small tables, before growing kicks in. A SET command takes 300 times longer than normal. I tried to let rehashing follow probing chains, but this got even worse spikes, with long probe chains + rehashing a very long chains.

Here's the test case: https://github.com/valkey-io/valkey/pull/1178/files#r1814492725

If we could let the iterator walk in some other order (such as index++) we could avoid this, but it doesn't completely solve the problem either. I guess there's simply a chance of getting very large probe chains and this cases a severe degradation.

So what can we do? Maybe keep track of not only the number of 'everfulls', but the max lengths of probe chains, and *grow* the table if it's too long? This might help.

But I think the best solution is Madelyn's idea about chaining instead of probing.

I think we need to implement it. It has other benefits too, like
* we can have higher fill factor and we can avoid rehashing indefinitely when there's a fork running;
* scan doesn't need to follow chains wrapping around zero, so it won't return duplicates.

src/hashset.c
rainsupreme: Hmm, good point. I realized we only actually need to check when we're first starting a rehash - in all other cases we know that we ended rehashStep with a bucket without `everfull` set :)

src/hashset.c
rainsupreme: With your helper function idea I was able to make this `if (!b->presence) return;`

src/hashset.c
rainsupreme: Hmm, I see! Yes, that makes sense. I think it'll be fixed if I don't modify `cursor` here and instead apply the masks to `start_cursor`

zuiderkwast: Yes, maybe. But it doesn't matter much, especially if we want to change probing to chaining anyway... Let's not spend too much effort on this.

zuiderkwast:
With chaining instead of probing, I guess we don't need this fix anymore. Thanks anyway!

PR #1218, opened 2024-10-24
[hashset feature] add UT to test random fairness with long probe chains

rainsupreme:
This is targeting the hashset branch and will update that PR if merged.

Here's the UT I promised to replace a tcl test I deleted in another PR 😅


zuiderkwast (approved):
OK, thanks!

When we change to bucket chaining, is this easy to adapt to that? I want to do that change ASAP so it's probably a good idea to think about this already.

rainsupreme:
Yes, this should directly apply to bucket chaining. If it fails it'll be because it got less fair.

PR #763, opened 2024-07-09
Io thread work offload

uriyage:
### IO-Threads Work Offloading 

This PR is the 2nd of 3 PRs intended to achieve the goal of 1M requests per second.
(1st PR: https://github.com/valkey-io/valkey/pull/758)

This PR offloads additional work to the I/O threads, beyond the current read-parse/write operations, to better utilize the I/O threads and reduce the load on the main thread.

It contains the following 3 commits:

### Poll Offload

Currently, the main thread is responsible for executing the poll-wait system call, while the IO threads wait for tasks from the main thread. The poll-wait operation is expensive and can consume up to 30% of the main thread's time. We could have let the IO threads do the poll-wait by themselves, with each thread listening to some of the clients and notifying the main thread when a client's command is ready to execute.

However, the current approach, where the main thread listens for events from the network, has several benefits. The main thread remains in charge, allowing it to know the state of each client (idle/read/write/close) at any given time. Additionally, it makes the threads flexible, enabling us to drain an IO thread's job queue and stop a thread when the load is light without modifying the event loop and moving its clients to a different IO thread. Furthermore, with this approach, the IO threads don't need to wait for both messages from the network and from the main thread instead, the threads wait only for tasks from the main thread.

To enjoy the benefits of both the main thread remaining in charge and the poll being offloaded, we propose offloading the poll-wait as a single-time, non-blocking job to one of the IO threads. The IO thread will perform a poll-wait non-blocking call while the main thread processes the client commands. Later, in `aeProcessEvents`, instead of sleeping on the poll, we check for the IO thread's poll-wait results.

The poll-wait will be offloaded in `beforeSleep` only when there are ready events for the main thread to process. If no events are pending, the main thread will revert to the current behavior and sleep on the poll by itself.

**Implementation Details**

A new call back `custompoll` was added to the `aeEventLoop` when not set to `NULL` the ae will call the `custompoll` callback instead of the `aeApiPoll`.

When the poll is offloaded we will set the `custompoll` to `getIOThreadPollResults` and send a poll-job to the thread. the thread will take a mutex, call a non-blocking (with timeout 0)  to `aePoll` which will populate the fired events array. the IO thread  will set the `server.io_fired_events` to the number of the returning `numevents`, later the main-thread in `custompoll` will return the  `server.io_fired_events` and will set the `customPoll` back to `NULL`.

To ensure thread safety when accessing server.el, all functions that modify the eventloop events were wrapped with a mutex to ensure mutual exclusion when modifying the events.

[...]

Issue #761, opened 2024-07-09
Followup items from #758 Async IO threads


madolson:
List of pending items from #758, all of these are future enhancements, and not pending issues that need to be merged before Valkey 8 (but they could be).
- [x] #1271, remove err_clean in tls 
- [x] #1449, replication traffic
- [x] #1485, writes to replicas
- [x] #1405
- [x] #2044
- [ ] Extend IO free offload to dbDelete - handle defrag
- [ ] lazy free to use IO threads?
- [x] #1338
- [ ] Move all client’s fields accessed by the IO threads to a sub-struct.
- [x] #2033 Allow dynamic modification of IO threads num.

uriyage:
- [ ] #2077

zuiderkwast:
* ~~Offload PING command~~

Edit: Let's not do it. See discussion below.

rjd15372:
Hi all, I'll start working on the "Offload PING command".

rjd15372:
I've been working on the offload of the PING command and I came up with two approaches that I would like your opinion on.

**Approach A)**

The IO thread that is choose to handle the read of the PING command, read, parses, and executes the command, and then also writes the reply, before processing a new IO job,

**Approach B)**

After the PING command has been read and parsed, the main thread schedules the execution of the PING command (enqueues a new job in the IO queue). The IO thread that picks this "execution" job, executes the job and enqueues the write job in the IO queue.

Approach **A** is more simple, but it might block the IO thread if the socket is not available to write, while approach **B** is more complex but more asynchronous.

I'm in favor of approach **B**, even though it will require a generalization of the IO threads component to support a different kind (non-IO job), and also support the concurrent insertion of jobs in the shared queue.

@zuiderkwast @madolson thoughts?

zuiderkwast:
@uriyage, you know IO threading better than anyone. Can you help?

@rjd15372 Approach B sounds like it will not be worth the overhead. Executing PING is very fast, maybe even faster executing in the main thread than delegating it to some thread to execute it.

I like approach A, which should add zero work to the main thread if the socket is writable, which is probably the normal case. If it's not writable, we need to register a write handler in the event loop and add the client to the global list `server.clients_pending_io_write`, which is not thread safe, so this probably needs to be done by the main thread (unless we refactor this in some way).

When the main thread delegates a read-and-parse-command job to an IO thread, it does among other things..

```c
    c->io_read_state = CLIENT_PENDING_IO;
[...]

PR #1220, opened 2024-10-24
Increase the IO_THREADS_MAX_NUM.

zhulipeng:
### Description

This patch try to increase the max number of io-threads from 16(128) to 256 for below reasons:

1. The core number increases a lot in the modern server processors, for example, the [Sierra Forest](https://en.wikipedia.org/wiki/Sierra_Forest) processors are targeted towards with up to **288** cores.
  Due to limitation of **_io-threads_** number (16 and 128 ),  benchmark like https://openbenchmarking.org/test/pts/valkey even cannot run on a high core count server. 

2. For some workloads, the bottleneck could be main thread, but for the other workloads, big key/value which caused heavy io, the bottleneck could be the io-threads, for example benchmark `memtier_benchmark -s 127.0.0.1 -p 9001 "--data-size" "20000" --ratio 1:0 --key-pattern P:P --key-minimum=1 --key-maximum 1000000 --test-time 180 -c 50 -t 16 --hide-histogram`. The QPS is still scalable after 16 io-threads. 
image
**Fig 1. QPS Scale factor with io-threads number grows.**

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -338,7 +338,7 @@ void setcpuaffinity(const char *cpulist);
 #define HAVE_FADVISE
 #endif
 
-#define IO_THREADS_MAX_NUM 16
+#define IO_THREADS_MAX_NUM 256
 
 #ifndef CACHE_LINE_SIZE
 #if defined(__aarch64__) && defined(__APPLE__)
--- a/src/config.c
+++ b/src/config.c
@@ -622,9 +622,6 @@ void loadServerConfigFromString(char *config) {
     if (server.config_hz < CONFIG_MIN_HZ) server.config_hz = CONFIG_MIN_HZ;
     if (server.config_hz > CONFIG_MAX_HZ) server.config_hz = CONFIG_MAX_HZ;
 
-    /* To ensure backward compatibility when io_threads_num is according to the previous maximum of 128. */
-    if (server.io_threads_num > IO_THREADS_MAX_NUM) server.io_threads_num = IO_THREADS_MAX_NUM;
-
     sdsfreesplitres(lines, totlines);
     reading_config_file = 0;
     return;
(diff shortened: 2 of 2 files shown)


PingXie:
This change makes sense to me. I have seen use cases involving even larger payloads (in MBs).

@valkey-io/core-team @uriyage what do you think?

src/config.h
```diff
@@ -338,7 +338,7 @@ void setcpuaffinity(const char *cpulist);
 #define HAVE_FADVISE
 #endif
 
-#define IO_THREADS_MAX_NUM 16
```
hpatro, line 341: Was the number `16` decided earlier for users to avoid shooting themselves in the foot i.e. performance will start deteriorating beyond that count?

zuiderkwast: IIRC, we use 3 bits in an robj to store which thread allocated an robj, so it can be passed back to be freed by the same thread.

hpatro: So, we can't blindly increase this?

zuiderkwast: Maybe not. :) Can you look it up?

There was some discussion related to the PR #763.

zuiderkwast: Seems i was wrong.

zhulipeng: > Was the number `16` decided earlier for users to avoid shooting themselves in the foot i.e. performance will start deteriorating beyond that count?

@hpatro Per my testing before, we will observe a significant performance degradation when the io-threads number greater than online CPUs, this issue still exists with the new valkey async IO, the spinning io-threads preempt CPU from main thread. It should have no relationship with the `16` number. I remember @madolson suggested to trust the operators :). https://github.com/redis/redis/issues/13003

The number `16` @uriyage responded in https://github.com/valkey-io/valkey-private/pull/6#discussion_r1597652783 before, he couldn't observe any improvement beyond 8-10 threads anyway.  Not sure if he tested the larger payloads (in MBs).

zuiderkwast (approved):
If we make it 128, it's no change of the config's upper limit, so it would be easy to merge. If we go beyond that, I guess it's an interface change = core team majority decision.

zhulipeng:
> If we make it 128, it's no change of the config's upper limit, so it would be easy to merge. If we go beyond that, I guess it's an interface change = core team majority decision.

Or we can be more aggressive, increase it to `512` or larger for multiple sockets ? :)

src/config.h
madolson: The work to limit the number of IO threads was so that we could send free requests to the same thread that allocated as was mentioned. For various reasons we decided to not commit to that as part of the V1, but I believe it's on one of the follow items: https://github.com/valkey-io/valkey/issues/761. 

I do believe in trusting operators/admins. (Never trust application developers though, they shoot themselves in the foot). We should definitely be "sane" by default, but give power users the operational tools to do advanced things.

However, it goes from 4 bits to 8 bits, which I think is probably OK.

madolson:
> Or we can be more aggressive, increase it to 512 or larger for multiple sockets ? :)

Valkey's performance on multiple sockets is awful because of NUMA access time. (Although I guess the pre-fetching helps a bit with that?)

madolson (approved):
I'll ship this. I'm assuming @lipzhu didn't test with TLS either, which will also benefit from more IO threads.

enjoy-binbin (approved):

PingXie (approved):

PR #1222, opened 2024-10-24
Maintain return value of rdbSaveDb after writing slot-info aux

enjoy-binbin:
All other places written in this function are maintained it,
although the caller of rdbSaveDb does not reply on it, it is
maintained to be consistent with other places, is its duty.


zuiderkwast (approved):
Good catch

PR #1226, opened 2024-10-25
Fix minor repldbfd leak in updateReplicasWaitingBgsave if fstat fails

enjoy-binbin:
In the old code, if fstat fails, replica->repldbfd will hold the
fd and we are doing a free client. And in freeClient, we check and
close only if repl_state == REPLICA_STATE_SEND_BULK. So if fstat
fails, we will leak the fd.

We can also extend freeClient to handle REPLICA_STATE_WAIT_BGSAVE_END
as well, but here seems to be a more friendly (and safer) way.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -1741,6 +1741,7 @@ void updateReplicasWaitingBgsave(int bgsaveerr, int type) {
         client *replica = ln->value;
 
         if (replica->repl_state == REPLICA_STATE_WAIT_BGSAVE_END) {
+            int repldbfd;
             struct valkey_stat buf;
 
             if (bgsaveerr != C_OK) {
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
OK, this fix works.

Another idea is we remove the dependency between `c->repl_state` and the fd and other things, change this part in freeClient():

```diff
-        if (c->repl_state == REPLICA_STATE_SEND_BULK) {
             if (c->repldbfd != -1) close(c->repldbfd);
             if (c->replpreamble) sdsfree(c->replpreamble);
-        }
```

If there is a `c->repldbfd`, we free it. It's more simple logic.

enjoy-binbin:
yes. I want to keep the dependency at first since it is a indeed a special state we need to handle. Both are ok to me which one do you prefer?

zuiderkwast:
> Both are ok to me which one do you prefer?

Your code is good. Let's merge it. :)

PR #1227, opened 2024-10-25
Kill diskless fork child asap when the last replica drop

enjoy-binbin:
We originally checked the replica connection to whether to kill the
diskless child only when rdbPipeReadHandler is triggered. Actually
we can check it when the replica is disconnected, so that we don't
have to wait for rdbPipeReadHandler to be triggered and can kill
the forkless child as soon as possible.

In this way, when the child or rdbPipeReadHandler is stuck for some
reason, we can kill the child faster and release the fork resources.

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -1552,12 +1552,17 @@ void unlinkClient(client *c) {
          * in which case it needs to be cleaned from that list */
         if (c->flag.replica && c->repl_state == REPLICA_STATE_WAIT_BGSAVE_END && server.rdb_pipe_conns) {
             int i;
+            int still_alive = 0;
             for (i = 0; i < server.rdb_pipe_numconns; i++) {
                 if (server.rdb_pipe_conns[i] == c->conn) {
                     rdbPipeWriteHandlerConnRemoved(c->conn);
                     server.rdb_pipe_conns[i] = NULL;
-                    break;
                 }
+                if (server.rdb_pipe_conns[i]) still_alive++;
+            }
+            if (still_alive == 0) {
+                serverLog(LL_NOTICE, "Diskless rdb transfer, last replica dropped, killing fork child.");
+                killRDBChild();
             }
         }
         /* Only use shutdown when the fork is active and we are the parent. */
@@ -1771,6 +1776,7 @@ void freeClient(client *c) {
         if (server.saveparamslen == 0 && c->repl_state == REPLICA_STATE_WAIT_BGSAVE_END &&
             server.child_type == CHILD_TYPE_RDB && server.rdb_child_type == RDB_CHILD_TYPE_DISK &&
             anyOtherReplicaWaitRdb(c) == 0) {
+            serverLog(LL_NOTICE, "Background saving, persistence disabled, last replica dropped, killing fork child.");
             killRDBChild();
         }
         if (c->repl_state == REPLICA_STATE_SEND_BULK) {
--- a/src/replication.c
+++ b/src/replication.c
@@ -1669,7 +1669,9 @@ void rdbPipeReadHandler(struct aeEventLoop *eventLoop, int fd, void *clientData,
                 if (!conn) continue;
                 stillUp++;
             }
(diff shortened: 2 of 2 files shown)


src/networking.c
```diff
@@ -1552,12 +1552,17 @@ void unlinkClient(client *c) {
          * in which case it needs to be cleaned from that list */
         if (c->flag.replica && c->repl_state == REPLICA_STATE_WAIT_BGSAVE_END && server.rdb_pipe_conns) {
             int i;
+            int still_alive = 0;
             for (i = 0; i < server.rdb_pipe_numconns; i++) {
                 if (server.rdb_pipe_conns[i] == c->conn) {
                     rdbPipeWriteHandlerConnRemoved(c->conn);
                     server.rdb_pipe_conns[i] = NULL;
```
ranshid, line 1559: I wonder if we could also decrement the server.rdb_pipe_numconns here and swap the entry with the last one?
not that important but will reduce the need to iterate over null connections later on...

src/replication.c
```diff
@@ -1708,15 +1709,10 @@ void rdbPipeReadHandler(struct aeEventLoop *eventLoop, int fd, void *clientData,
...
-            stillAlive++;
         }
 
-        if (stillAlive == 0) {
-            serverLog(LL_WARNING, "Diskless rdb transfer, last replica dropped, killing fork child.");
-            killRDBChild();
-        }
         /*  Remove the pipe read handler if at least one write handler was set. */
-        if (server.rdb_pipe_numconns_writing || stillAlive == 0) {
+        if (server.rdb_pipe_numconns_writing) {
```
ranshid, line 1715: Just so that I understand: in case we kill the RDB child when the last replica unlinks, we will either read '0' before in rdbPipeReadHandler or we will read data from the closed pipe but  since we will not write the data to any replica, we will just remove the read handler right? However this will not call  rdbPipeReadHandler again which will close the read pipe end. Maybe I got it all wrong :) @enjoy-binbin did I?

ranshid (commented):
@enjoy-binbin generally looks good to me. I think there are no tests checking the logs for 0 replicas left during bgsave. Maybe worth  adding a test to read and validate the correct logs are issued in that case?

src/replication.c
ranshid: Oh forgot we are reading in a loop here :) sorry then it looks fine to me.

src/networking.c
xbasel: What would be the benefit of incrementally cleaning up rdb_pipe_numconns ?

ranshid: small benefit. just avoid iterating on them later in the read handler

enjoy-binbin: yes, the decrement thing should work i guess, but i may not want to modify it, fear of that i will cause some troubles LOL.

xbasel: IMO - For such a small array, the overhead of managing swapping/shifting logic outweighs the benefit. Simplicity wins here.

xbasel (approved):

src/networking.c
ranshid: Fine, there goes away my wish to connect 1M replicas... :)

enjoy-binbin:
> generally looks good to me. I think there are no tests checking the logs for 0 replicas left during bgsave. Maybe worth adding a test to read and validate the correct logs are issued in that case?

sorry i forgot to reply. `Diskless rdb transfer, last replica dropped, killing fork child` you are saying we don't have test to cover this case?

i see we do have a test `diskless $all_drop replicas drop during rdb pipe` that are checking this log

ranshid:
> > generally looks good to me. I think there are no tests checking the logs for 0 replicas left during bgsave. Maybe worth adding a test to read and validate the correct logs are issued in that case?
> 
> sorry i forgot to reply. `Diskless rdb transfer, last replica dropped, killing fork child` you are saying we don't have test to cover this case?
> 
> i see we do have a test `diskless $all_drop replicas drop during rdb pipe` that are checking this log

Don't know how I missed it :) good to go

ranshid (approved):

zuiderkwast (approved):

PR #1230, opened 2024-10-25
Improve performance of sdssplitargs

madolson:
The current implementation of `sdssplitargs` does repeated `sdscatlen` to build the parsed arguments, which isn't very efficient because it does a lot of extra reallocations and moves through the sds code a lot. It also typically results in memory overhead, because `sdscatlen` over-allocates, which is usually not needed since args are usually not modified after being created.

The new implementation of sdssplitargs does two passes, the first to parse the argument to figure out the final length and the second to actually copy the string. It's generally about 2x faster for larger strings (~100 bytes), and about 20% faster for small strings (~10 bytes). This is generally faster since as long as everything is in the CPU cache, it's going to be fast.

There are a couple of sanity tests, none existed before, as well as some fuzzying which was used to find some bugs and also to do the benchmarking. The original benchmarking code can be seen https://github.com/valkey-io/valkey/pull/1230/commits/6576aeb86adfb5afa74aefb5bc2d2becde95ce4c.

```
test_sdssplitargs_benchmark - unit/test_sds.c:530] Using random seed: 1729883235
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 56.44%, new:13039us, old:29930us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 56.58%, new:12057us, old:27771us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 59.18%, new:9048us, old:22165us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 54.61%, new:12381us, old:27278us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 51.17%, new:16012us, old:32793us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 49.18%, new:16041us, old:31563us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 58.40%, new:12450us, old:29930us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 56.49%, new:13066us, old:30031us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 58.75%, new:12744us, old:30894us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 52.44%, new:16885us, old:35504us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 62.57%, new:8107us, old:21659us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 62.12%, new:8320us, old:21966us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 45.23%, new:13960us, old:25487us
[test_sdssplitargs_benchmark - unit/test_sds.c:577] Improvement: 57.95%, new:9188us, old:21849us
```


src/sds.c
```diff
@@ -1032,6 +1032,86 @@ int hex_digit_to_int(char c) {
     }
 }
 
+/* Helper function for sdssplitargs that parses a single argument. It
+ * populates either the number characters needed to store the parsed argument
```
zuiderkwast: Drop "either"? It seems to populate both if both are provided.

src/unit/test_sds.c
```diff
@@ -328,3 +330,261 @@ int test_sdsHeaderSizes(int argc, char **argv, int flags) {
...
+    case 'D': return 13;
+    case 'e':
+    case 'E': return 14;
+    case 'f':
+    case 'F': return 15;
+    default: return 0;
+    }
+}
+
+static sds *sdssplitargs_legacy(const char *line, int *argc) {
```
zuiderkwast: I would drop the legacy implementation before we merge this PR. The point is to compare them and publish the benchmark results in the PR description, right?

zuiderkwast (commented):
Nice!

Is sdssplitargs used in any hot code path?

madolson:
> Is sdssplitargs used in any hot code path?

It's used in the inline protocol command parsing, which isn't super commonly used but I've seen it in production. It's also in some command parsing like configs.

src/unit/test_sds.c
madolson: Yeah, that's mostly true. I wasn't sure if I wanted to keep it, so I left it here. I'll drop it.

zuiderkwast (approved):

PR #1231, opened 2024-10-25
[hashset feature] Fixed a UT failure that only happened with O3 optimization

rainsupreme:
Original test violated strict aliasing rules. I first wrote verbose comments about how to get away with storing non-pointers, but have now updated it to simply avoid storing non-pointers. Much cleaner and safer.

Technically we could work around this and still store longs directly in a hashset, but currently we don't have a need for that.


zuiderkwast (approved):
Great fix, great comment.

I think the comment should also be either at the hashsetType struct in hashset.h or at hashsetNext in hashset.c, which I hope is where developers will look for documentation.

src/unit/test_hashset.c
```diff
@@ -438,7 +437,13 @@ int test_safe_iterator(int argc, char **argv, int flags) {
     long num_returned = 0;
     hashsetIterator iter;
     hashsetInitSafeIterator(&iter, s);
-    while (hashsetNext(&iter, (void **)&j)) {
+
+    /* Using void* variable to avoid a cast that violates strict aliasing */
+    void *element;
+    while (hashsetNext(&iter, &element)) {
+        static_assert(sizeof(long) == sizeof(void *));
```
zuiderkwast: Instead of `long`, we could use `intptr_t` throughout this test suite. It explicitly means pointer-sized integer.

zuiderkwast: The only place where `sizeof(long) != sizeof(void *)` is windows. See https://en.wikipedia.org/wiki/64-bit_computing#64-bit_data_models

src/hashset.c
```diff
@@ -32,6 +32,33 @@
  * Functions and types are prefixed by "hashset", macros by "HASHSET". Internal
  * names don't use the prefix. Internal functions are 'static'.
  *
+ * Caution: directly storing non-pointer elements
+ * ----------------------------------------------
+ *
```
zuiderkwast: Some intro about what this is about could be useful:

```suggestion
 * Storing non-pointer values, such as pointer-sized integers (long), in a
 * hashset is possible. Be careful when casting to and from `void *` though.
 *
```

src/unit/test_hashset.c
```diff
@@ -404,7 +408,7 @@ int test_iterator(int argc, char **argv, int flags) {
 
     /* Check that all elements were returned exactly once. */
     TEST_ASSERT(num_returned == count);
-    for (j = 0; j < count; j++) {
+    for (long j = 0; j < count; j++) {
```
zuiderkwast: intptr_t here too?

```suggestion
    for (intptr_t j = 0; j < count; j++) {
```

src/unit/test_hashset.c
```diff
@@ -438,7 +441,12 @@ int test_safe_iterator(int argc, char **argv, int flags) {
     long num_returned = 0;
     hashsetIterator iter;
     hashsetInitSafeIterator(&iter, s);
-    while (hashsetNext(&iter, (void **)&j)) {
+
+    /* Using void* variable to avoid a cast that violates strict aliasing */
+    void *element;
+    while (hashsetNext(&iter, &element)) {
+        intptr_t j = (long)element;
```
zuiderkwast:
```suggestion
        intptr_t j = (intptr_t)element;
```

src/unit/test_hashset.c
```diff
@@ -462,7 +470,7 @@ int test_safe_iterator(int argc, char **argv, int flags) {
     /* Check that all elements present during the whole iteration were returned
      * exactly once. (Some are deleted after being returned.) */
     TEST_ASSERT(num_returned >= count);
-    for (j = 0; j < count; j++) {
+    for (long j = 0; j < count; j++) {
```
zuiderkwast:
```suggestion
    for (intptr_t j = 0; j < count; j++) {
```

src/unit/test_hashset.c
```diff
@@ -471,7 +479,7 @@ int test_safe_iterator(int argc, char **argv, int flags) {
     /* Check that elements inserted during the iteration were returned at most
      * once. */
     unsigned long num_optional_returned;
-    for (j = count; j < count * 2; j++) {
+    for (long j = count; j < count * 2; j++) {
```
zuiderkwast:
```suggestion
    for (intptr_t j = count; j < count * 2; j++) {
```

src/unit/test_hashset.c
```diff
@@ -495,18 +503,21 @@ int test_random_element(int argc, char **argv, int flags) {
...
     memset(times_picked, 0, sizeof(times_picked));
     for (long i = 0; i < num_rounds; i++) {
-        long element;
-        assert(hashsetFairRandomElement(s, (void **)&element));
-        assert(element >= 0 && element < count);
-        times_picked[element]++;
+        /* Using void* variable to avoid a cast that violates strict aliasing */
+        void *element;
+        assert(hashsetFairRandomElement(s, &element));
+        intptr_t j = (long)element;
```
zuiderkwast:
```suggestion
        intptr_t j = (intptr_t)element;
```

zuiderkwast (commented):
Great. Seems you missed some occurrences of `long` though.

The reason I went with long originally is that it's more common in the code base and that for `printf` we have `%ld` for long. For `intptr_t` there may be some code but it's less simple. Anyway, we can use `%ld` for `intptr_t` as long as we're not on windows.

zuiderkwast:
Ah, this is the printf thing:

```
unit/test_hashset.c:452:31: error: format ‘%ld’ expects argument of type ‘long int’, but argument 2 has type ‘intptr_t’ {aka ‘int’} [-Werror=format=]
  452 |             printf("Element %ld returned, max == %ld. Num returned: %ld\n", j, count * 2 - 1, num_returned);
      |                             ~~^                                             ~
      |                               |                                             |
      |                               long int                                      intptr_t {aka int}
      |                             %d
```

Even if `long` and `int` are both 32-bit on 32-bit platforms, they're still different in some way... It seems intptr_t is an alias of int and not long in this case.

Maybe we should stick to long after all? That's what we do in the rest of this code base.

src/hashset.c
```diff
@@ -32,6 +32,36 @@
...
+ * Storing non-pointer values, such as pointer-sized integers (long), in a
+ * hashset is possible. Be careful when casting to and from `void *` though.
```
madolson: This seems like we should use a union instead like we do with the dict.c to avoid having to cast between (void *) and primitive types.

madolson: Like we do with https://github.com/valkey-io/valkey/blob/unstable/src/dict.c#L76.

zuiderkwast: In practice, we only store pointers though. In dict, we store integer values in the expire dict and double values in sorted sets(?) because we have a key as well in the dict entry, but in hashset we need the key and value together so there is no use case for storing only a set of integers or a set of floats, apart from in the unit tests.

We don't want a union that includes double and int64_t since it would make it 64 bits even on 32-bit platforms. We should only allow pointer-sized types in that case, such as `intptr_t i` and maybe `float` or `double` depending on platform. On top of that, using a union would make the usage a bit more complex.

Since the only use case is to unit tests, I'd rather skip using non-pointers in the unit tests instead.

madolson: Then let's remove the large comment that rain added and just say it's not supported.

zuiderkwast: Yeah, @SoftlyRaining WDYT?

We'd need to do some extra work to use some allocated objects in the unit tests then, or we can cheat and use longs and add comments about aliasing only in the unit test suite.

rainsupreme: I actually like this best.

src/unit/test_hashset.c
```diff
@@ -377,36 +377,36 @@ int test_iterator(int argc, char **argv, int flags) {
     UNUSED(argv);
     UNUSED(flags);
 
-    long count = 2000000;
+    size_t count = 2000000;
+    uint8_t element_array[count];
+    memset(element_array, 0, sizeof element_array);
```
zuiderkwast, line 401: ~~Equivalent?~~ Ah, it doesn't work. We get "error: variable-sized object may not be initialized".

```suggestion
    uint8_t element_array[count] = {0};
```

src/unit/test_hashset.c
```diff
@@ -377,36 +377,36 @@ int test_iterator(int argc, char **argv, int flags) {
...
     hashsetIterator iter;
     hashsetInitIterator(&iter, s);
-    while (hashsetNext(&iter, (void **)&j)) {
+    uint8_t *item;
+    while (hashsetNext(&iter, (void **)&item)) {
         num_returned++;
-        assert(j >= 0 && j < count);
-        element_returned[j]++;
+        assert(item >= element_array && item < element_array + count);
+        (*item)++;
```
zuiderkwast: Now, the array name `element_returned` is lost, so it can be good to explain in another way what we're doing here. A comment like this?

```suggestion
        /* Increment the element at this position, as a counter
         * of how many times this element has been returned. */
        (*item)++;
```

Another question: Why mix the words "item" and "element"? They mean the same. Let's use element?

src/unit/test_hashset.c
```diff
@@ -420,62 +420,63 @@ int test_safe_iterator(int argc, char **argv, int flags) {
...
         num_returned++;
-        if (j < 0 || j >= count * 2) {
-            printf("Element %ld returned, max == %ld. Num returned: %ld\n", j, count * 2 - 1, num_returned);
+        if (item < element_counts || item >= element_counts + count * 2) {
+            printf("Element %zu returned, max == %zu. Num returned: %zu\n", index, count * 2 - 1, num_returned);
             printf("Safe %d, table %d, index %lu, pos in bucket %d, rehashing? %d\n", iter.safe, iter.table, iter.index,
                    iter.pos_in_bucket, !hashsetIsRehashing(s));
             hashsetHistogram(s);
             exit(1);
         }
```
zuiderkwast: This was debugging code that I used while implementing. I don't think we need to keep it. Do you?

I think we can delete it. There's already an assert below:

    assert(item >= element_counts || item < element_counts + count * 2);

zuiderkwast (commented):
Yeah, this is better. Just some nits.

zuiderkwast (approved):

PR #1232, opened 2024-10-25
Mark main and serverAssert as weak symbols to be overridden

madolson:
At some point unit tests stopped building on MacOS because of duplicate symbols. I had originally solved this problem by using a flag that overrides symbols, but the much better solution is to mark the duplicate symbols as weak and they can be overridden during linking. (Symbols by default are strong, strong symbols override weak symbols)

I also added macos unit build to the CI, so that this doesn't silently break in the future again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -83,7 +83,7 @@ jobs:
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - name: make
-        run: make -j3 SERVER_CFLAGS='-Werror'
+        run: make all-with-unit-tests -j3 SERVER_CFLAGS='-Werror'
 
   build-32bit:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


enjoy-binbin (approved):

.github/workflows/ci.yml
```diff
@@ -83,7 +83,7 @@ jobs:
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - name: make
-        run: make -j3 SERVER_CFLAGS='-Werror'
+        run: make all-with-unit-tests -j3 SERVER_CFLAGS='-Werror'
```
zuiderkwast: make flags before the target seems more common style. We do that in other places.

```suggestion
        run: make -j3 all-with-unit-tests SERVER_CFLAGS='-Werror'
```

src/server.c
```diff
@@ -6810,7 +6810,7 @@ serverTestProc *getTestProcByName(const char *name) {
 }
 #endif
 
-int main(int argc, char **argv) {
+__attribute__((weak)) int main(int argc, char **argv) {
```
zuiderkwast, line 6814: Add a comment about why we add this attribute on main?

zuiderkwast (approved):
Good idea!

src/server.c
madolson, line 6814:
```suggestion
/* Main is marked as weak so that unit tests can use their own main function. */
__attribute__((weak)) int main(int argc, char **argv) {
```

PR #1233, opened 2024-10-25
Document log format configs in valkey.conf

### LINKED ISSUE #1225: Document log format configs in valkey.conf
Document log format configs in valkey.conf

Configs are documented there. We missed that. Let's do it as a follow-up.

_Originally posted by @zuiderkwast in https://github.com/valkey-io/valkey/issues/1022#issuecomment-2436464364_

azuredream:
Add config options for log format and timestamp format introduced by #1022 
Related to #1225

This change adds two new configs into valkey.conf:
log-format
log-timestamp-format

### DIFF
--- a/valkey.conf
+++ b/valkey.conf
@@ -348,18 +348,22 @@ pidfile /var/run/valkey_6379.pid
 # nothing (nothing is logged)
 loglevel notice
 
-# Specifies the format used for logging Redis server messages.
+# Specify the logging format.
 # This can be one of:
-# legacy (traditional Redis log format)
-# logfmt (Structured log format "logfmt")
-log-format legacy
+#
+# - legacy: the default, traditional log format
+# - logfmt: a structured log format; see https://www.brandur.org/logfmt
+#
+# log-format legacy
 
-# log_timestamp_format - Specifies the timestamp format used in Redis logs.
-# This can be one of:
-# legacy       (traditional Redis timestamp format)
-# iso8601      (ISO 8601)
-# milliseconds (milliseconds since the epoch)
-log-timestamp-format legacy
+# Specify the timestamp format used in logs using 'log-timestamp-format'.
+#
+# - legacy: default format
+# - iso8601: ISO 8601 extended date and time with time zone, on the form
+#   yyyy-mm-ddThh:mm:ss.sss±hh:mm
+# - milliseconds: milliseconds since the epoch
+#
+# log-timestamp-format legacy
 
 # Specify the log file name. Also the empty string can be used to force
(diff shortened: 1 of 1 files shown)


enjoy-binbin (approved):

valkey.conf
```diff
@@ -348,6 +348,19 @@ pidfile /var/run/valkey_6379.pid
 # nothing (nothing is logged)
 loglevel notice
 
+# Specifies the format used for logging Redis server messages.
```
zuiderkwast: This is not Redis. :)

All the configs are about the server. No need to mention it, I think. Short is ok, like

```suggestion
# Specify the logging format.
```

valkey.conf
```diff
@@ -348,6 +348,19 @@ pidfile /var/run/valkey_6379.pid
...
+# This can be one of:
+# legacy (traditional Redis log format)
+# logfmt (Structured log format "logfmt")
+log-format legacy
```
zuiderkwast: I think it's better to just add comments, no actual configured value. Users copy this file and they will get this value. Later if we change the default, users have the old value in their copied config file and they will not get the new default.

For the "pidfile" config, we have a strange situation. The value in this template config file is "/var/run/valkey_6379.pid" but if users delete this line, they will get another default "/var/run/valkey.pid".

So, for many other configs, we just say that the default is and show it as a commented example config.

```suggestion
# This can be one of:
#
# - legacy: the default, traditional log format
# - logfmt: a structured log format; see https://www.brandur.org/logfmt
#
# log-format legacy
```

valkey.conf
```diff
@@ -348,6 +348,19 @@ pidfile /var/run/valkey_6379.pid
...
+
+# log_timestamp_format - Specifies the timestamp format used in Redis logs.
+# This can be one of:
+# legacy       (traditional Redis timestamp format)
+# iso8601      (ISO 8601)
+# milliseconds (milliseconds since the epoch)
+log-timestamp-format legacy
```
zuiderkwast: Similar comment as the log-format. I'll try a suggestion:

```suggestion
# Specify the timestamp format used in logs using 'log-timestamp-format'.
#
# - legacy: default format
# - iso8601: ISO 8601 extended date and time with time zone, on the form
#   yyyy-mm-ddThh:mm:ss.sss±hh:mm
# - milliseconds: milliseconds since the epoch
#
# log-timestamp-format legacy
```

azuredream: Code updated. Thanks for reviewing!

zuiderkwast (approved):
Thanks! Sorry for suggesting changes to all of this text. It's still very good help that you did the PR.

PR #1234, opened 2024-10-28
Eliminate snprintf usage at setDeferredAggregateLen

imasahiro:
to align with how we encode the length at `_addReplyLongLongWithPrefix`


madolson (approved):
Makes sense, and should be a little bit faster.

PR #1235, opened 2024-10-28
Minor fix to debug logging in replicationFeedStreamFromPrimaryStream

enjoy-binbin:
We should only print logs when hide-user-data-from-log is off.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -651,7 +651,7 @@ void replicationFeedStreamFromPrimaryStream(char *buf, size_t buflen) {
     /* Debugging: this is handy to see the stream sent from primary
      * to replicas. Disabled with if(0). */
     if (0) {
-        if (server.hide_user_data_from_log) {
+        if (!server.hide_user_data_from_log) {
             printf("%zu:", buflen);
             for (size_t j = 0; j < buflen; j++) {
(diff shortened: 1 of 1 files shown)


src/replication.c
```diff
@@ -651,7 +651,7 @@ void replicationFeedStreamFromPrimaryStream(char *buf, size_t buflen) {
     /* Debugging: this is handy to see the stream sent from primary
      * to replicas. Disabled with if(0). */
     if (0) {
-        if (server.hide_user_data_from_log) {
+        if (!server.hide_user_data_from_log) {
```
enjoy-binbin, line 654: I just happened to be debugging it and found that. Let me know if you think we should make it become a serverLog.

We can also maybe, add a DEBUG subcommand to eliminate this IF 0.

Let me know and i can update this PR.

hwware: I think your code is correct. We should avoid to print these information if hide_user_data_from_log flag is 1.

hwware (approved):

src/replication.c
madolson: IMO, I would just merge the PR as is.

madolson (approved):

PR #871, opened 2024-08-05
To avoid bouncing -REDIRECT during FAILOVER

soloestoy:
Fix #821

During the `FAILOVER` process, when conditions are met (such as when the force time is reached or the primary and replica offsets are consistent), the primary actively becomes the replica and transitions to the `FAILOVER_IN_PROGRESS` state. After the primary becomes the replica, and after handshaking and other operations, it will eventually send the `PSYNC FAILOVER` command to the replica, after which the replica will become the primary. This means that the upgrade of the replica to the primary is an asynchronous operation, which implies that during the `FAILOVER_IN_PROGRESS` state, there may be a period of time where both nodes are replicas. In this scenario, if a `-REDIRECT` is returned, the request will be redirected to the replica and then redirected back, causing back and forth redirection. To avoid this situation, during the `FAILOVER_IN_PROGRESS state`, we temporarily suspend the clients that need to be redirected until the replica truly becomes the primary, and then resume the execution.

PR #1238, opened 2024-10-29
Enhance SENTINEL FAILOVER to use the FAILOVER command to avoid data loss

enjoy-binbin:
Currently, SENTINEL FAILOVER selects a replica to send REPLICAOF NO ONE,
then waits for PROMOTION, and then sends REPLICAOF new_ip new_port to the
old primary and other replicas to complete the failover. The problem here
is that if the old primary has written to it before the role change during
the failover, these writes will be lost.

We can use the FAILOVER command to avoid this data loss. FAILOVER was added
in 0d18a1e85fdb9967419dc9f7a5905452d065c84a, it can coordinates the failover
between the primary and the replica.

Before the original step REPLICAOF NO ONE, we try to send FAILOVER TO ip port
to the primary to go through the FAILOVER process. If the primary does not
support FAILOVER, for example, returns an error, it will fallback to the old
logic, that means fallback to REPLICAOF NO ONE.


enjoy-binbin:
This is a PR that was written very quickly (I wrote it at night and i want to push it as soon as possible). There may be problems with the details or it can be optimized.

Also see the https://github.com/redis/redis/issues/6062 for more data loss details. The test case also proves data loss locally.

src/sentinel.c
```diff
@@ -3837,11 +3840,31 @@ void sentinelCommand(client *c) {
             addReplyBulkLongLong(c, addr->port);
         }
     } else if (!strcasecmp(c->argv[1]->ptr, "failover")) {
-        /* SENTINEL FAILOVER <primary-name> */
+        /* SENTINEL FAILOVER <primary-name> [type <legacy | pause>] */
```
zuiderkwast: Do we really need to let the user select the legacy behaviour? Who wants data loss?

Is it a breaking change if we just change this to always use FAILOVER? (a fallback to the old behavior if FAILOVER is not supported is OK)

enjoy-binbin: I am worried that the primary support FAILOVER, but the replica does not. In this case, the primary seems to return OK, and sentinel has no way of knowing whether the replica support FAILOVER. And then the failover timeout and always timeout in every retry. Not sure about this case. Or we can default to new way and provide option to use the old way if user find out the failover is always timeout?

If we provide options, maybe we should not have this fallback logic seems it may cause trouble if the user really want to use the new way and find out it is using old way at the end. (As long as it is well documented, i guess it should be fine.)

zuiderkwast: In valkey 8, the replica sends `REPLCONF version`. If it's valkey 8 or later we can use failover. If it's older we can use replica of noone?

enjoy-binbin: The REPLCONF version is only sent to primary from a replica, right? You are saying that in FAILOVER, we check the replica version? This require a change in primary side, like we need to do some coding around it and make FAILOVER reject with an old replica.

What happend if a new sentinel send a FAILOVER TO to a primary (without the replica version check)? And this will also break FAILOVER command in some area, i don't think we should limit it.

When FAILOVER is added, we extend PSYNC to take a FAILOVER option in the replica, but we don't extend REPLCONF capa to support it. So a primary don't have a way to know if replica support PSYNC FAILOVER. In this case, primary will abort the FAILOVER.

Added these options somehow i pass the burden of choice to the user (to avoid trouble on myself..), who can choose the right option based on there version, and can use the old way when the new way fails.

zuiderkwast: Ahhh, I mixed up with the shutdown-failover feature. :laughing:

I prefer to avoid a new config if we can. Sorry for the trouble.

If the primary supports FAILOVER but the replica doesn't support FAILOVER, it means the primary is newer than the replica. We don't support this configuration, I think. In a rolling upgrade, the new version should always be added as a replica first. At least that's how to rolling upgrade in a cluster.

The FAILOVER command was added in 6.2. Even if we want to support downgrade, IMO we don't need to support downgrade to < 6.2, because 6.0 is already end-of-life.

So, I think it's acceptable that sentinel tries FAILOVER and if the primary doesn't support it, it fallbacks to REPLICAOF NO ONE. You don't think so?

Another alternative I guess is that sentinel can coordinate the failover:

1. Sentinel sends CLIENT PAUSE WRITE to the primary
2. Sentinel checks offset of the replica
3. Sentinel sends REPLICAOF

enjoy-binbin: > So, I think it's acceptable that sentinel tries FAILOVER and if the primary doesn't support it, it fallbacks to REPLICAOF NO ONE. You don't think so?

i think it is ok, i am ok with the fallback logic. So i just need to keep this PR and drop the legacy type, right? You are right, i thought too much about this in this version thing..

Yes, the client pause is another option, it require more code, i think FAILOVER is somehow easy to use so i just use it, it also has some INFO fields we can see unless CLIENT PAUSE. The benefit of Client PASUE may be that it can work on old Redis (maybe 3.0?). Which one do you think is better?

zuiderkwast: Less work is better. :)

> So i just need to keep this PR and drop the legacy type, right? 

Yes.

src/networking.c
```diff
@@ -2387,7 +2387,7 @@ parseResult handleParseResults(client *c) {
     }
 
     if (c->read_flags & READ_FLAGS_PARSING_NEGATIVE_MBULK_LEN) {
-        /* Multibulk processing could see a <= 0 length. */
+        /* Multibulk rocessing could see a <= 0 length. */
```
zuiderkwast: Mistake

src/sentinel.c
```diff
@@ -110,6 +110,7 @@ static mstime_t sentinel_default_failover_timeout = 60 * 3 * 1000;
 #define SENTINEL_FAILOVER_STATE_NONE 0                 /* No failover in progress. */
 #define SENTINEL_FAILOVER_STATE_WAIT_START 1           /* Wait for failover_start_time*/
 #define SENTINEL_FAILOVER_STATE_SELECT_REPLICA 2       /* Select replica to promote */
+#define SENTINEL_FAILOVER_STATE_SEND_FAILOVER 7        /* Send FAILOVER Command to primary. */
 #define SENTINEL_FAILOVER_STATE_SEND_REPLICAOF_NOONE 3 /* Replica -> Primary */
 #define SENTINEL_FAILOVER_STATE_WAIT_PROMOTION 4       /* Wait replica to change role */
 #define SENTINEL_FAILOVER_STATE_RECONF_REPLICAS 5      /* REPLICAOF newprimary */
```
zuiderkwast: These numbers are only internal in the code? We can change them?

```suggestion
#define SENTINEL_FAILOVER_STATE_SEND_FAILOVER 3        /* Send FAILOVER Command to primary. */
#define SENTINEL_FAILOVER_STATE_SEND_REPLICAOF_NOONE 4 /* Replica -> Primary */
#define SENTINEL_FAILOVER_STATE_WAIT_PROMOTION 5       /* Wait replica to change role */
#define SENTINEL_FAILOVER_STATE_RECONF_REPLICAS 6      /* REPLICAOF newprimary */
```

... and the last one is outside the part where I can suggest in GitHub.

tests/sentinel/tests/05-manual.tcl
```diff
@@ -54,14 +72,43 @@ test "All the other slaves now point to the new primary" {
...
+    }
+}
+
+} ;# end proc test_sentinel_failover
+
+source "../tests/includes/init-tests.tcl"
+test_sentinel_failover "legacy" $master_id
+
+source "../tests/includes/init-tests.tcl"
+test_sentinel_failover "pause" $master_id
```
zuiderkwast: Call them "failover" and "replicaof" instead of legacy and pause?

tests/sentinel/tests/05-manual.tcl
```diff
@@ -1,18 +1,26 @@
...
+
+test "Manual failover works - $type" {
+    R $master_id del foo
     set old_port [RPort $master_id]
     set addr [S 0 SENTINEL GET-PRIMARY-ADDR-BY-NAME mymaster]
     assert {[lindex $addr 1] == $old_port}
 
+    # Rename the FAILOVER command so that we can fallback to REPLICAOF NO ONE.
+    if {$type == "legacy"} {
+        S 0 SENTINEL SET mymaster rename-command FAILOVER FAILOVER_ERR
```
zuiderkwast: Rename to ERR? it's not an error.. Maybe FAILOVER_OLD, FAILOVER_ORIGINAL or FAILOVER_RENAMED?

zuiderkwast (commented):
Yes, this looks good.

I'm not enough familiar with Sentinel. @hwware you are a sentinel expert. Do you want to take a look?

tests/sentinel/tests/05-manual.tcl
enjoy-binbin: This is just to rename to a non-existent name, so that when sentinel sends it to primary, it will return an Unknown command error, something like that.

src/sentinel.c
enjoy-binbin: yes, we can change them, i just wanted to reduce diff and avoid conflicts, but the code is indeed ugly... i will change it

tests/sentinel/tests/05-manual.tcl
enjoy-binbin: i will rename it to other thing to mention the non-existent thing.

zuiderkwast: rename to FAILOVER_NONEXISTENT?

tests/sentinel/tests/05-manual.tcl
```diff
@@ -1,18 +1,26 @@
...
+
+test "Manual failover works - $type" {
+    R $master_id del foo
     set old_port [RPort $master_id]
     set addr [S 0 SENTINEL GET-PRIMARY-ADDR-BY-NAME mymaster]
     assert {[lindex $addr 1] == $old_port}
 
+    # Rename the FAILOVER command so that we can fallback to REPLICAOF NO ONE.
+    if {$type == "legacy"} {
+        S 0 SENTINEL SET mymaster rename-command FAILOVER NON-EXISTENT
```
hwware, line 22: In real production, this step: call rename-command explicitly is must?

enjoy-binbin: No, it is just a test step that try to cover the fallback logic.

hwware:
Overall, LTGM, just one concern about the test case.

tests/sentinel/tests/05-manual.tcl
zuiderkwast: Let's explain in the comment

    # We simulate a server that doesn't have the FAILOVER command.

gmbnomis:
Some time ago, I proposed to use FAILOVER with SENTINEL FAILOVER in redis, see https://github.com/redis/redis/issues/13118. I implemented this proposal and also ported my first implementation to Valkey some time ago, see https://github.com/gmbnomis/valkey/commit/146850ac13c7638f3cf2c08447756ca34094d3a5.

The main issues I learned about when doing that were: (I think that this PR does not address these)
1. We should follow the Sentinel protocol. Clients can expect to be disconnected when the role of the primary changes (see https://valkey.io/topics/sentinel-clients/). A simple FAILOVER command does not achieve that. Thus, clients may find themselves unable to write suddenly.
2. FAILOVER pauses all client writes during the failover (in Valkey, it even pauses some client reads during the FAILOVER_IN_PROGRESS phase, see https://github.com/valkey-io/valkey/pull/871#discussion_r1703980072). This means that Sentinels (remember that SENTINEL FAILOVER is performed without an election) will regard the former primary as down during the failover procedure and may initiate a failover. Also, if I remember correctly, the sentinel instance in charge of the failover may block its own connection to the primary. 
3. The FAILOVER command may hang indefinitely in the FAILOVER_IN_PROGRESS phase even if TIMEOUT is given. We should probably have a mechanism to detect such a condition and do a FAILOVER ABORT.

For reference: my proposal addresses 1. It addresses 2 somewhat (probably a solution seeking a leader election would be safer here) and does not address 3 (I remember that I started working on this but I did not finish it. The idea was to monitor the INFO field "master_failover_state" and send an FAILOVER ABORT if a node is in the state for too long. )

enjoy-binbin:
thanks for the detail comments.

> 1. We should follow the Sentinel protocol. Clients can expect to be disconnected when the role of the primary changes (see https://valkey.io/topics/sentinel-clients/). A simple FAILOVER command does not achieve that. Thus, clients may find themselves unable to write suddenly.

Yes, we should. I remember this thing but nerver be able to found the documentation on it. thanks for the link.

> 2. FAILOVER pauses all client writes during the failover (in Valkey, it even pauses some client reads during the FAILOVER_IN_PROGRESS phase, see https://github.com/valkey-io/valkey/pull/871#discussion_r1703980072). This means that Sentinels (remember that SENTINEL FAILOVER is performed without an election) will regard the former primary as down during the failover procedure and may initiate a failover. Also, if I remember correctly, the sentinel instance in charge of the failover may block its own connection to the primary.

I did not read the whole story, i will find some times to catch up. Yes it seems you are right, all my thinking is in happy path. Maybe we should use FORCE option?

> 3. The FAILOVER command may hang indefinitely in the FAILOVER_IN_PROGRESS phase even if TIMEOUT is given. We should probably have a mechanism to detect such a condition and do a FAILOVER ABORT.

I support in this, a FORCE option is needed, in this case, the worst case is the same as before.

> For reference: my proposal addresses 1. It addresses 2 somewhat (probably a solution seeking a leader election would be safer here) and does not address 3 (I remember that I started working on this but I did not finish it. The idea was to monitor the INFO field "master_failover_state" and send an FAILOVER ABORT if a node is in the state for too long. )

@gmbnomis i did not read your proposal and implementation before, i will read it later, do you want to pick it up? You have done a lot of work here, it would be great that if you could pick it up, or we can slowly push it forward in the future and make all of this happen.

gmbnomis:
> 
> I did not read the whole story, i will find some times to catch up. Yes it seems you are right, all my thinking is in happy path. Maybe we should use FORCE option?
> 

Using FAILOVER with the FORCE option would be more similar to the current SENTINEL FAILOVER in that we don't care about the replication state and just fail over at some point in time. 

> > 3. The FAILOVER command may hang indefinitely in the FAILOVER_IN_PROGRESS phase even if TIMEOUT is given. We should probably have a mechanism to detect such a condition and do a FAILOVER ABORT.
> 
> I support in this, a FORCE option is needed, in this case, the worst case is the same as before.

The FORCE option of FAILOVER does not apply to the FAILOVER_IN_PROGRESS state, so we might get stuck as well and need some mechanism to get out of it. (My current thinking is that this does not need to be very efficient because the probability of this happening seems rather low. But the system should be able to recover from such a failure) 
 
> > For reference: my proposal addresses 1. It addresses 2 somewhat (probably a solution seeking a leader election would be safer here) and does not address 3 (I remember that I started working on this but I did not finish it. The idea was to monitor the INFO field "master_failover_state" and send an FAILOVER ABORT if a node is in the state for too long. )
> 
> @gmbnomis i did not read your proposal and implementation before, i will read it later, do you want to pick it up? You have done a lot of work here, it would be great that if you could pick it up, or we can slowly push it forward in the future and make all of this happen.

@enjoy-binbin I am not completely sure, but you mean to pick up my old proposal and re-start working on it? As there seems to be interest in this feature I'd be happy to do that.  (If so, how should I start? Create a new issue with an update of https://github.com/redis/redis/issues/13118 to flesh out the implementation options further?) 

I am also happy to support getting another implementation into shape.

gmbnomis:
@enjoy-binbin I have posted  an updated description #1291 and implementation #1292 now. I think it addresses all three topics I mentioned above.

enjoy-binbin:
thanks for the work, i am closing this one, let move the discusstion to there

Issue #1141, opened 2024-10-09
[NEW] Active Defrag - duty cycle and high level refactoring


JimB123:
# Overview
In Valkey, it's possible to accumulate fragmentation in blocks of memory managed by jemalloc.  In order to recover this fragmented memory, we have a process (Active Defrag) that scans over all of the data structures in Valkey and decides if any of the individual allocations needs to be relocated.  Jemalloc helps us to identify allocations which need to be relocated.  After relocating allocations on lesser-used memory pages, jemalloc is able to reuse the pages or return them to the OS.

# Issue/Problems

**Duty Cycle:** Active Defrag has configuration values which determine the intended percentage of CPU to be used based on a gradient of the  fragmentation percentage.  However, Active Defrag performs its work on the 100ms `serverCron` timer.  It then computes a duty cycle and performs a single long cycle.  For example, if the intended CPU is computed to be 10%, Active Defrag will perform 10ms of work on this 100ms timer cron.
* This type of cycle introduces large latencies on the client (up to 25ms with default configurations)
* This mechanism is subject to starvation when slow commands delay the `serverCron`

**Maintainability:**  The current Active Defrag code is difficult to read & maintain.  Refactoring of the high level control mechanisms and functions will allow us to more seamlessly adapt to new defragmentation needs.  Specific examples include:
* A single function (`activeDefragCycle`) includes the logic to start/stop/modify the defragmentation as well as performing one "step" of the defragmentation.  This should be separated out, so that the actual defrag activity can be performed on an independent timer (see duty cycle above).
* The code is focused on kvstores, with other actions just thrown in at the end (`defragOtherGlobals`).  There's no mechanism to break this up to reduce latencies.
* For the main dictionary (only), there is a mechanism to set aside large keys to be processed in a later step.  However this code creates a separate list in each kvstore (main dict or not), bleeding/exposing internal defrag logic.  We only need 1 list - inside defrag.  This logic should be more contained for the main key store.
* The structure is not well suited towards other non-main-dictionary items.  For example, pub-sub and pub-sub-shard was added, but it's added in such a way that in CMD mode, with multiple DBs, we will defrag pub-sub repeatedly after each DB.
[...]

PR #977, opened 2024-09-02
Take hz into account in activerehashing to avoid CPU spikes

enjoy-binbin:
Currently in conf we describe activerehashing as: Active rehashing
uses 1 millisecond every 100 milliseconds of CPU time. This is the
case for hz = 10.

If we change hz, the description in conf will be inaccurate. Users
may notice that the server spends some CPU (used in activerehashing)
at high hz but don't know why, since our cron calls are fixed to 1ms.

This PR takes hz into account and fixed the CPU usage at 1% (this may
not be accurate in some cases because we do 100 step rehashing in
dictRehashMicroseconds but it can avoid CPU spikes in this case).

This PR also improves the description of the activerehashing
configuration item to explain this change.

PR #1242, opened 2024-10-31
Refactor of ActiveDefrag to reduce latencies

JimB123:
Refer to:  https://github.com/valkey-io/valkey/issues/1141

This update refactors the defrag code to:
* Make the overall code more readable and maintainable
* Reduce latencies incurred during defrag processing

With this update, the defrag cycle time is reduced to 500us, with more frequent cycles.  This results in much more predictable latencies, with a dramatic reduction in tail latencies.

(See https://github.com/valkey-io/valkey/issues/1141 for more complete details.)

This update is focused mostly on the high-level processing, and does NOT address lower level functions which aren't currently timebound (e.g. `activeDefragSdsDict()`, and `moduleDefragGlobals()`).  These are out of scope for this update and left for a future update.

I fixed `kvstoreDictLUTDefrag` because it was using up to 7ms on a CME single shard.

During unit tests, the following max latencies were measured (in verbose mode):
<table><thead>
  <tr>
    <th>Unit test name</th>
    <th>Old Latency ms</th>
    <th>New Latency ms</th>
  </tr></thead>
<tbody>
  <tr>
    <td>Active defrag main dictionary: cluster</td>
    <td>8</td>
    <td>0</td>
  </tr>
  <tr>
    <td>Active defrag big keys: cluster</td>
    <td>8</td>
    <td>0</td>
  </tr>
  <tr>
    <td>Active defrag main dictionary: standalone</td>
    <td>8</td>
    <td>0</td>
  </tr>
  <tr>
    <td>Active defrag - AOF loading</td>
    <td>20</td>
    <td>0</td>
  </tr>
  <tr>
    <td>Active defrag big keys: standalone</td>
    <td>8</td>
    <td>0</td>
  </tr>
  <tr>
    <td>Active defrag big list: standalone</td>
    <td>8</td>
    <td>0</td>
  </tr>
</tbody>
</table>

In addition, the following test was run on both old and new versions of the software:
```
// Create fragmented host
./src/valkey-benchmark -r 10000000 -n 10000000 -d 3 -t set 
./src/valkey-benchmark -r 9000000 -n 10000000 -d 11 -t set 
./src/valkey-benchmark -r 8000000 -n 10000000 -d 19 -t set 
./src/valkey-benchmark -r 7000000 -n 10000000 -d 27 -t set

// Enable defrag while running some traffic
./src/valkey-cli config set activedefrag yes; ./src/valkey-benchmark -r 7000000 -c 1 -n 1000000 -l -t get
```
Configuration was set so that both old/new clusters would use 10% defrag CPU.
Defrag time OLD:  120 sec
Defrag time NEW:  105 sec
Times were based on a 5-second poll.  (I didn't have logs running.)

The improvement in run time is believed to be due to the unskewed nature of the new code which provides a more accurate 10% of the CPU.

This is the OLD distribution for the GET benchmark while defragging:
```
0.000% <= 0.023 milliseconds (cumulative count 5)
50.000% <= 0.031 milliseconds (cumulative count 953602)
96.875% <= 0.039 milliseconds (cumulative count 980927)
98.438% <= 0.047 milliseconds (cumulative count 990566)
99.219% <= 0.055 milliseconds (cumulative count 993349)
99.609% <= 0.351 milliseconds (cumulative count 996097)
99.805% <= 0.535 milliseconds (cumulative count 998320)
99.902% <= 0.559 milliseconds (cumulative count 999088)
99.951% <= 0.759 milliseconds (cumulative count 999516)
99.976% <= 10.191 milliseconds (cumulative count 999765)
99.988% <= 10.231 milliseconds (cumulative count 999889)
99.994% <= 10.263 milliseconds (cumulative count 999941)
99.997% <= 10.335 milliseconds (cumulative count 999971)
99.998% <= 10.687 milliseconds (cumulative count 999985)
99.999% <= 10.759 milliseconds (cumulative count 999994)
100.000% <= 10.823 milliseconds (cumulative count 999997)
100.000% <= 11.167 milliseconds (cumulative count 999999)
100.000% <= 11.175 milliseconds (cumulative count 1000000)
100.000% <= 11.175 milliseconds (cumulative count 1000000)
```
This is the equivalent NEW distribution:
```
0.000% <= 0.023 milliseconds (cumulative count 8)
50.000% <= 0.031 milliseconds (cumulative count 934278)
93.750% <= 0.039 milliseconds (cumulative count 969396)
98.438% <= 0.055 milliseconds (cumulative count 985215)
99.219% <= 0.535 milliseconds (cumulative count 993656)
99.609% <= 0.543 milliseconds (cumulative count 997369)
99.805% <= 0.551 milliseconds (cumulative count 998365)
99.902% <= 0.583 milliseconds (cumulative count 999070)
99.951% <= 0.783 milliseconds (cumulative count 999515)
99.976% <= 1.023 milliseconds (cumulative count 999775)
99.988% <= 1.047 milliseconds (cumulative count 999903)
99.994% <= 1.071 milliseconds (cumulative count 999945)
99.997% <= 1.175 milliseconds (cumulative count 999970)
99.998% <= 1.295 milliseconds (cumulative count 999985)
99.999% <= 1.527 milliseconds (cumulative count 999994)
100.000% <= 1.543 milliseconds (cumulative count 999997)
100.000% <= 1.551 milliseconds (cumulative count 999999)
100.000% <= 1.735 milliseconds (cumulative count 1000000)
100.000% <= 1.735 milliseconds (cumulative count 1000000)
```
You can see in the new distribution that there is a very slight increase in latencies <= 99.951%, in exchange for a huge reduction in tail latencies.

A CONTROL test WITHOUT Active Defrag running shows a very similar distribution with a variation of roughly 500us in the tail latencies (as expected):
```
0.000% <= 0.023 milliseconds (cumulative count 7)
50.000% <= 0.031 milliseconds (cumulative count 958519)
96.875% <= 0.039 milliseconds (cumulative count 983380)
98.438% <= 0.047 milliseconds (cumulative count 991683)
99.219% <= 0.055 milliseconds (cumulative count 993548)
99.609% <= 0.319 milliseconds (cumulative count 996120)
99.805% <= 0.527 milliseconds (cumulative count 998137)
99.902% <= 0.543 milliseconds (cumulative count 999134)
99.951% <= 0.567 milliseconds (cumulative count 999575)
99.976% <= 0.615 milliseconds (cumulative count 999759)
99.988% <= 0.767 milliseconds (cumulative count 999879)
99.994% <= 0.911 milliseconds (cumulative count 999940)
99.997% <= 1.015 milliseconds (cumulative count 999975)
99.998% <= 1.031 milliseconds (cumulative count 999991)
99.999% <= 1.039 milliseconds (cumulative count 999995)
100.000% <= 1.055 milliseconds (cumulative count 999998)
100.000% <= 1.063 milliseconds (cumulative count 999999)
100.000% <= 1.071 milliseconds (cumulative count 1000000)
100.000% <= 1.071 milliseconds (cumulative count 1000000)
```

### DIFF
--- a/src/ae.c
+++ b/src/ae.c
@@ -85,7 +85,7 @@ aeEventLoop *aeCreateEventLoop(int setsize) {
     if (eventLoop->events == NULL || eventLoop->fired == NULL) goto err;
     eventLoop->setsize = setsize;
     eventLoop->timeEventHead = NULL;
-    eventLoop->timeEventNextId = 0;
+    eventLoop->timeEventNextId = 1;
     eventLoop->stop = 0;
     eventLoop->maxfd = -1;
     eventLoop->beforesleep = NULL;
--- a/src/config.c
+++ b/src/config.c
@@ -3208,10 +3208,11 @@ standardConfig static_configs[] = {
     createIntConfig("list-max-listpack-size", "list-max-ziplist-size", MODIFIABLE_CONFIG, INT_MIN, INT_MAX, server.list_max_listpack_size, -2, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("tcp-keepalive", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.tcpkeepalive, 300, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("cluster-migration-barrier", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.cluster_migration_barrier, 1, INTEGER_CONFIG, NULL, NULL),
-    createIntConfig("active-defrag-cycle-min", NULL, MODIFIABLE_CONFIG, 1, 99, server.active_defrag_cycle_min, 1, INTEGER_CONFIG, NULL, updateDefragConfiguration),                 /* Default: 1% CPU min (at lower threshold) */
-    createIntConfig("active-defrag-cycle-max", NULL, MODIFIABLE_CONFIG, 1, 99, server.active_defrag_cycle_max, 25, INTEGER_CONFIG, NULL, updateDefragConfiguration),                /* Default: 25% CPU max (at upper threshold) */
+    createIntConfig("active-defrag-cycle-min", NULL, MODIFIABLE_CONFIG, 1, 99, server.active_defrag_cpu_min, 1, INTEGER_CONFIG, NULL, updateDefragConfiguration),                   /* Default: 1% CPU min (at lower threshold) */
+    createIntConfig("active-defrag-cycle-max", NULL, MODIFIABLE_CONFIG, 1, 99, server.active_defrag_cpu_max, 25, INTEGER_CONFIG, NULL, updateDefragConfiguration),                  /* Default: 25% CPU max (at upper threshold) */
     createIntConfig("active-defrag-threshold-lower", NULL, MODIFIABLE_CONFIG, 0, 1000, server.active_defrag_threshold_lower, 10, INTEGER_CONFIG, NULL, NULL),                       /* Default: don't defrag when fragmentation is below 10% */
     createIntConfig("active-defrag-threshold-upper", NULL, MODIFIABLE_CONFIG, 0, 1000, server.active_defrag_threshold_upper, 100, INTEGER_CONFIG, NULL, updateDefragConfiguration), /* Default: maximum defrag force at 100% fragmentation */
+    createIntConfig("active-defrag-cycle-us", NULL, MODIFIABLE_CONFIG, 0, 100000, server.active_defrag_cycle_us, 500, INTEGER_CONFIG, NULL, updateDefragConfiguration),
     createIntConfig("lfu-log-factor", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.lfu_log_factor, 10, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("lfu-decay-time", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.lfu_decay_time, 1, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("replica-priority", "slave-priority", MODIFIABLE_CONFIG, 0, INT_MAX, server.replica_priority, 100, INTEGER_CONFIG, NULL, NULL),
--- a/src/defrag.c
+++ b/src/defrag.c
@@ -38,23 +38,126 @@
 
 #ifdef HAVE_DEFRAG
 
-typedef struct defragCtx {
-    void *privdata;
-    int slot;
-    void *aux;
-} defragCtx;
+typedef enum { DEFRAG_NOT_DONE = 0,
+               DEFRAG_DONE = 1 } doneStatus;
 
-typedef struct defragPubSubCtx {
-    kvstore *pubsub_channels;
-    dict *(*clientPubSubChannels)(client *);
-} defragPubSubCtx;
 
-/* Defrag helper for generic allocations.
+/*
+ * Defragmentation is performed in stages.  Each stage is serviced by a stage function
+ * (defragStageFn).  The stage function is passed a target (void*) to defrag.  The contents of that
+ * target are unique to the particular stage - and may even be NULL for some stage functions.  The
+ * same stage function can be used multiple times (for different stages) each having a different
+ * target.
  *
- * returns NULL in case the allocation wasn't moved.
- * when it returns a non-null value, the old pointer was already released
- * and should NOT be accessed. */
-void *activeDefragAlloc(void *ptr) {
+ * The stage function is required to maintain an internal static state.  This allows the stage
+ * function to continue when invoked in an iterative manner.  When invoked with a 0 endtime, the
+ * stage function is required to clear it's internal state and prepare to begin a new stage.  It
+ * should return false (more work to do) as it should NOT perform any real "work" during init.
+ *
+ * Parameters:
+ *  endtime     - This is the monotonic time that the function should end and return.  This ensures
+ *                a bounded latency due to defrag.  When endtime is 0, the internal state should be
+ *                cleared, preparing to begin the stage with a new target.
+ *  target      - This is the "thing" that should be defragged.  It's type is dependent on the
+ *                type of the stage function.  This might be a dict, a kvstore, a DB, or other.
+ *  privdata    - A pointer to arbitrary private data which is unique to the stage function.
+ *
+ * Returns:
+ *  - DEFRAG_DONE if the stage is complete
+ *  - DEFRAG_NOT_DONE if there is more work to do
+ */
+typedef doneStatus (*defragStageFn)(monotime endtime, void *target, void *privdata);
+
+typedef struct {
+    defragStageFn stage_fn; // The function to be invoked for the stage
+    void *target;           // The target that the function will defrag
+    void *privdata;         // Private data, unique to the stage function
+} StageDescriptor;
+
+/* Globals needed for the main defrag processing logic.
+ * Doesn't include variables specific to a stage or type of data. */
+struct DefragContext {
+    monotime start_cycle;           // Time of beginning of defrag cycle
+    long long start_defrag_hits;    // server.stat_active_defrag_hits captured at beginning of cycle
+    list *remaining_stages;         // List of stages which remain to be processed
+    StageDescriptor *current_stage; // The stage that's currently being processed
+
+    long long timeproc_id;      // Eventloop ID of the timerproc (or AE_DELETED_EVENT_ID)
+    monotime timeproc_end_time; // Ending time of previous timerproc execution
+    long timeproc_overage_us;   // A correction value if over/under target CPU percent
+};
+static struct DefragContext defrag;
+
+
+/* There are a number of stages which process a kvstore.  To simplify this, a stage helper function
+ * `defragStageKvstoreHelper()` is defined.  This function aids in iterating over the kvstore.  It
+ * uses these definitions.
+ */
+/* State of the kvstore helper.  The private data (privdata) passed to the kvstore helper MUST BEGIN
+ *  with a kvstoreIterState (or be passed as NULL). */
+#define KVS_SLOT_DEFRAG_LUT -2
+#define KVS_SLOT_UNASSIGNED -1
+typedef struct {
+    kvstore *kvs;
+    int slot;
+    unsigned long cursor;
+} kvstoreIterState;
+/* The kvstore helper uses this function to perform tasks before continuing the iteration.  For the
+ * main dictionary, large items are set aside and processed by this function before continuing with
+ * iteration over the kvstore.
+ *  endtime     - This is the monotonic time that the function should end and return.
+ *  privdata    - Private data for functions invoked by the helper.  If provided in the call to
+ *                `defragStageKvstoreHelper()`, the `kvstoreIterState` portion (at the beginning)
+ *                will be updated with the current kvstore iteration status.
+ *
+ * Returns:
+ *  - DEFRAG_DONE if the pre-continue work is complete
+ *  - DEFRAG_NOT_DONE if there is more work to do
+ */
+typedef doneStatus (*kvstoreHelperPreContinueFn)(monotime endtime, void *privdata);
+
+
+// Private data for main dictionary keys
+typedef struct {
+    kvstoreIterState kvstate;
+    serverDb *db;
+    dictEntry *saved_expire_de;
+} defragKeysCtx;
+static_assert(offsetof(defragKeysCtx, kvstate) == 0, "defragStageKvstoreHelper requires this");
+
+// Private data for pubsub kvstores
+typedef dict *(*getClientChannelsFn)(client *);
+typedef struct {
+    getClientChannelsFn fn;
+} getClientChannelsFnWrapper;
+
+typedef struct {
+    kvstoreIterState kvstate;
+    getClientChannelsFn getPubSubChannels;
+} defragPubSubCtx;
+static_assert(offsetof(defragPubSubCtx, kvstate) == 0, "defragStageKvstoreHelper requires this");
+
+
+/* When scanning a main kvstore, large elements are queued for later handling rather than
+ * causing a large latency spike while processing a hash table bucket.  This list is only used
+ * for stage: "defragStageDbKeys".  It will only contain values for the current kvstore being
+ * defragged.
+ * Note that this is a list of key names.  It's possible that the key may be deleted or modified
+ * before "later" and we will search by key name to find the entry when we defrag the item later.
+ */
+static list *defrag_later;
+static unsigned long defrag_later_cursor;
+
+
+/* this method was added to jemalloc in order to help us understand which
+ * pointers are worthwhile moving and which aren't */
+int je_get_defrag_hint(void *ptr);
+
+/* Defrag function which allocates and copies memory if needed, but DOESN'T free the old block.
+ * It is the responsibility of the caller to free the old block if a non-NULL value (new block)
+ * is returned.  (Returns NULL if no relocation was needed.)
+ */
+static void *activeDefragAllocWithoutFree(void *ptr, size_t *allocation_size) {
     size_t size;
     void *newptr;
     if (!allocatorShouldDefrag(ptr)) {
@@ -67,28 +170,43 @@ void *activeDefragAlloc(void *ptr) {
     size = zmalloc_size(ptr);
     newptr = allocatorDefragAlloc(size);
     memcpy(newptr, ptr, size);
-    allocatorDefragFree(ptr, size);
+    if (allocation_size) *allocation_size = size;
+
     server.stat_active_defrag_hits++;
     return newptr;
 }
 
+/* Defrag helper for generic allocations.
+ *
+ * Returns NULL in case the allocation wasn't moved.
+ * When it returns a non-null value, the old pointer was already released
+ * and should NOT be accessed. */
+void *activeDefragAlloc(void *ptr) {
+    size_t allocation_size;
+    void *newptr = activeDefragAllocWithoutFree(ptr, &allocation_size);
+    if (newptr) allocatorDefragFree(ptr, allocation_size);
+    return newptr;
+}
+
 /* This method captures the expiry db dict entry which refers to data stored in keys db dict entry. */
-void defragEntryStartCbForKeys(void *ctx, void *oldptr) {
-    defragCtx *defragctx = (defragCtx *)ctx;
-    serverDb *db = defragctx->privdata;
+static void defragEntryStartCbForKeys(void *ctx, void *oldptr) {
+    defragKeysCtx *defragctx = (defragKeysCtx *)ctx;
+    serverDb *db = defragctx->db;
     sds oldsds = (sds)dictGetKey((dictEntry *)oldptr);
-    int slot = defragctx->slot;
+    int slot = defragctx->kvstate.slot;
     if (kvstoreDictSize(db->expires, slot)) {
         dictEntry *expire_de = kvstoreDictFind(db->expires, slot, oldsds);
-        defragctx->aux = expire_de;
+        defragctx->saved_expire_de = expire_de;
+    } else {
+        defragctx->saved_expire_de = NULL;
     }
 }
 
 /* This method updates the key of expiry db dict entry. The key might be no longer valid
  * as it could have been cleaned up during the defrag-realloc of the main dictionary. */
-void defragEntryFinishCbForKeys(void *ctx, void *newptr) {
-    defragCtx *defragctx = (defragCtx *)ctx;
-    dictEntry *expire_de = (dictEntry *)defragctx->aux;
+static void defragEntryFinishCbForKeys(void *ctx, void *newptr) {
+    defragKeysCtx *defragctx = (defragKeysCtx *)ctx;
+    dictEntry *expire_de = defragctx->saved_expire_de;
     /* Item doesn't have TTL associated to it. */
     if (!expire_de) return;
     /* No reallocation happened. */
@@ -96,18 +214,18 @@ void defragEntryFinishCbForKeys(void *ctx, void *newptr) {
         expire_de = NULL;
         return;
     }
-    serverDb *db = defragctx->privdata;
+    serverDb *db = defragctx->db;
     sds newsds = (sds)dictGetKey((dictEntry *)newptr);
-    int slot = defragctx->slot;
+    int slot = defragctx->kvstate.slot;
     kvstoreDictSetKey(db->expires, slot, expire_de, newsds);
 }
 
-/*Defrag helper for sds strings
+/* Defrag helper for sds strings
  *
- * returns NULL in case the allocation wasn't moved.
- * when it returns a non-null value, the old pointer was already released
+ * Returns NULL in case the allocation wasn't moved.
+ * When it returns a non-null value, the old pointer was already released
  * and should NOT be accessed. */
-sds activeDefragSds(sds sdsptr) {
+static sds activeDefragSds(sds sdsptr) {
     void *ptr = sdsAllocPtr(sdsptr);
     void *newptr = activeDefragAlloc(ptr);
     if (newptr) {
@@ -118,60 +236,48 @@ sds activeDefragSds(sds sdsptr) {
     return NULL;
 }
 
-/* Defrag helper for robj and/or string objects with expected refcount.
- *
- * Like activeDefragStringOb, but it requires the caller to pass in the expected
- * reference count. In some cases, the caller needs to update a robj whose
- * reference count is not 1, in these cases, the caller must explicitly pass
- * in the reference count, otherwise defragmentation will not be performed.
- * Note that the caller is responsible for updating any other references to the robj. */
-robj *activeDefragStringObEx(robj *ob, int expected_refcount) {
-    robj *ret = NULL;
-    if (ob->refcount != expected_refcount) return NULL;
-
-    /* try to defrag robj (only if not an EMBSTR type (handled below). */
-    if (ob->type != OBJ_STRING || ob->encoding != OBJ_ENCODING_EMBSTR) {
-        if ((ret = activeDefragAlloc(ob))) {
-            ob = ret;
-        }
+/* Performs defrag on a string-type (or generic) robj, but does not free the old robj.  This is the
+ * caller's responsibility.  This is necessary for string objects with multiple references.  In this
+ * case the caller can fix the references before freeing the original object.
+ */
+static robj *activeDefragStringObWithoutFree(robj *ob, size_t *allocation_size) {
+    if (ob->type == OBJ_STRING && ob->encoding == OBJ_ENCODING_RAW) {
+        // Try to defrag the linked sds, regardless of if robj will be moved
+        sds newsds = activeDefragSds((sds)ob->ptr);
+        if (newsds) ob->ptr = newsds;
     }
 
-    /* try to defrag string object */
-    if (ob->type == OBJ_STRING) {
-        if (ob->encoding == OBJ_ENCODING_RAW) {
-            sds newsds = activeDefragSds((sds)ob->ptr);
-            if (newsds) {
-                ob->ptr = newsds;
-            }
-        } else if (ob->encoding == OBJ_ENCODING_EMBSTR) {
-            /* The sds is embedded in the object allocation, calculate the
-             * offset and update the pointer in the new allocation. */
-            long ofs = (intptr_t)ob->ptr - (intptr_t)ob;
-            if ((ret = activeDefragAlloc(ob))) {
-                ret->ptr = (void *)((intptr_t)ret + ofs);
-            }
-        } else if (ob->encoding != OBJ_ENCODING_INT) {
-            serverPanic("Unknown string encoding");
-        }
+    robj *new_robj = activeDefragAllocWithoutFree(ob, allocation_size);
+
+    if (new_robj && ob->type == OBJ_STRING && ob->encoding == OBJ_ENCODING_EMBSTR) {
+        // If the robj is moved, correct the internal pointer
+        long embstr_offset = (intptr_t)ob->ptr - (intptr_t)ob;
+        new_robj->ptr = (void *)((intptr_t)new_robj + embstr_offset);
     }
-    return ret;
+    return new_robj;
 }
+
 
 /* Defrag helper for robj and/or string objects
  *
- * returns NULL in case the allocation wasn't moved.
- * when it returns a non-null value, the old pointer was already released
+ * Returns NULL in case the allocation wasn't moved.
+ * When it returns a non-null value, the old pointer was already released
  * and should NOT be accessed. */
 robj *activeDefragStringOb(robj *ob) {
-    return activeDefragStringObEx(ob, 1);
+    size_t allocation_size;
+    if (ob->refcount != 1) return NULL; // Unsafe to defrag if multiple refs
+    robj *new_robj = activeDefragStringObWithoutFree(ob, &allocation_size);
+    if (new_robj) allocatorDefragFree(ob, allocation_size);
+    return new_robj;
 }
+
 
 /* Defrag helper for lua scripts
  *
- * returns NULL in case the allocation wasn't moved.
- * when it returns a non-null value, the old pointer was already released
+ * Returns NULL in case the allocation wasn't moved.
+ * When it returns a non-null value, the old pointer was already released
  * and should NOT be accessed. */
-luaScript *activeDefragLuaScript(luaScript *script) {
+static luaScript *activeDefragLuaScript(luaScript *script) {
     luaScript *ret = NULL;
 
     /* try to defrag script struct */
@@ -193,7 +299,7 @@ luaScript *activeDefragLuaScript(luaScript *script) {
  * Returns NULL in case the allocation wasn't moved.
  * When it returns a non-null value, the old pointer was already released
  * and should NOT be accessed. */
-dict *dictDefragTables(dict *d) {
+static dict *dictDefragTables(dict *d) {
     dict *ret = NULL;
     dictEntry **newtable;
     /* handle the dict struct */
@@ -211,7 +317,7 @@ dict *dictDefragTables(dict *d) {
 }
 
 /* Internal function used by zslDefrag */
-void zslUpdateNode(zskiplist *zsl, zskiplistNode *oldnode, zskiplistNode *newnode, zskiplistNode **update) {
+static void zslUpdateNode(zskiplist *zsl, zskiplistNode *oldnode, zskiplistNode *newnode, zskiplistNode **update) {
     int i;
     for (i = 0; i < zsl->level; i++) {
         if (update[i]->level[i].forward == oldnode) update[i]->level[i].forward = newnode;
@@ -233,7 +339,7 @@ void zslUpdateNode(zskiplist *zsl, zskiplistNode *oldnode, zskiplistNode *newnod
  * only need to defrag the skiplist, but not update the obj pointer.
  * When return value is non-NULL, it is the score reference that must be updated
  * in the dict record. */
-double *zslDefrag(zskiplist *zsl, double score, sds oldele, sds newele) {
+static double *zslDefrag(zskiplist *zsl, double score, sds oldele, sds newele) {
     zskiplistNode *update[ZSKIPLIST_MAXLEVEL], *x, *newx;
     int i;
     sds ele = newele ? newele : oldele;
@@ -267,7 +373,7 @@ double *zslDefrag(zskiplist *zsl, double score, sds oldele, sds newele) {
 
 /* Defrag helper for sorted set.
  * Defrag a single dict entry key name, and corresponding skiplist struct */
-void activeDefragZsetEntry(zset *zs, dictEntry *de) {
+static void activeDefragZsetEntry(zset *zs, dictEntry *de) {
     sds newsds;
     double *newscore;
     sds sdsele = dictGetKey(de);
@@ -284,13 +390,13 @@ void activeDefragZsetEntry(zset *zs, dictEntry *de) {
 #define DEFRAG_SDS_DICT_VAL_VOID_PTR 3
 #define DEFRAG_SDS_DICT_VAL_LUA_SCRIPT 4
 
-void activeDefragSdsDictCallback(void *privdata, const dictEntry *de) {
+static void activeDefragSdsDictCallback(void *privdata, const dictEntry *de) {
     UNUSED(privdata);
     UNUSED(de);
 }
 
 /* Defrag a dict with sds key and optional value (either ptr, sds or robj string) */
-void activeDefragSdsDict(dict *d, int val_type) {
+static void activeDefragSdsDict(dict *d, int val_type) {
     unsigned long cursor = 0;
     dictDefragFunctions defragfns = {
         .defragAlloc = activeDefragAlloc,
@@ -306,34 +412,7 @@ void activeDefragSdsDict(dict *d, int val_type) {
 }
 
 /* Defrag a list of ptr, sds or robj string values */
-void activeDefragList(list *l, int val_type) {
-    listNode *ln, *newln;
-    for (ln = l->head; ln; ln = ln->next) {
-        if ((newln = activeDefragAlloc(ln))) {
-            if (newln->prev)
-                newln->prev->next = newln;
-            else
-                l->head = newln;
-            if (newln->next)
-                newln->next->prev = newln;
-            else
-                l->tail = newln;
-            ln = newln;
-        }
-        if (val_type == DEFRAG_SDS_DICT_VAL_IS_SDS) {
-            sds newsds, sdsele = ln->value;
-            if ((newsds = activeDefragSds(sdsele))) ln->value = newsds;
-        } else if (val_type == DEFRAG_SDS_DICT_VAL_IS_STROB) {
-            robj *newele, *ele = ln->value;
-            if ((newele = activeDefragStringOb(ele))) ln->value = newele;
-        } else if (val_type == DEFRAG_SDS_DICT_VAL_VOID_PTR) {
-            void *newptr, *ptr = ln->value;
-            if ((newptr = activeDefragAlloc(ptr))) ln->value = newptr;
-        }
-    }
-}
-
-void activeDefragQuickListNode(quicklist *ql, quicklistNode **node_ref) {
+static void activeDefragQuickListNode(quicklist *ql, quicklistNode **node_ref) {
     quicklistNode *newnode, *node = *node_ref;
     unsigned char *newzl;
     if ((newnode = activeDefragAlloc(node))) {
@@ -350,7 +429,7 @@ void activeDefragQuickListNode(quicklist *ql, quicklistNode **node_ref) {
     if ((newzl = activeDefragAlloc(node->entry))) node->entry = newzl;
 }
 
-void activeDefragQuickListNodes(quicklist *ql) {
+static void activeDefragQuickListNodes(quicklist *ql) {
     quicklistNode *node = ql->head;
     while (node) {
         activeDefragQuickListNode(ql, &node);
@@ -361,13 +440,18 @@ void activeDefragQuickListNodes(quicklist *ql) {
 /* when the value has lots of elements, we want to handle it later and not as
  * part of the main dictionary scan. this is needed in order to prevent latency
  * spikes when handling large items */
-void defragLater(serverDb *db, dictEntry *kde) {
+static void defragLater(dictEntry *kde) {
+    if (!defrag_later) {
+        defrag_later = listCreate();
+        listSetFreeMethod(defrag_later, (void (*)(void *))sdsfree);
+        defrag_later_cursor = 0;
+    }
     sds key = sdsdup(dictGetKey(kde));
-    listAddNodeTail(db->defrag_later, key);
+    listAddNodeTail(defrag_later, key);
 }
 
 /* returns 0 if no more work needs to be been done, and 1 if time is up and more work is needed. */
-long scanLaterList(robj *ob, unsigned long *cursor, long long endtime) {
+static long scanLaterList(robj *ob, unsigned long *cursor, monotime endtime) {
     quicklist *ql = ob->ptr;
     quicklistNode *node;
     long iterations = 0;
@@ -392,7 +476,7 @@ long scanLaterList(robj *ob, unsigned long *cursor, long long endtime) {
         activeDefragQuickListNode(ql, &node);
         server.stat_active_defrag_scanned++;
         if (++iterations > 128 && !bookmark_failed) {
-            if (ustime() > endtime) {
+            if (getMonotonicUs() > endtime) {
                 if (!quicklistBookmarkCreate(&ql, "_AD", node)) {
                     bookmark_failed = 1;
                 } else {
@@ -413,14 +497,14 @@ typedef struct {
     zset *zs;
 } scanLaterZsetData;
 
-void scanLaterZsetCallback(void *privdata, const dictEntry *_de) {
+static void scanLaterZsetCallback(void *privdata, const dictEntry *_de) {
     dictEntry *de = (dictEntry *)_de;
     scanLaterZsetData *data = privdata;
     activeDefragZsetEntry(data->zs, de);
     server.stat_active_defrag_scanned++;
 }
 
-void scanLaterZset(robj *ob, unsigned long *cursor) {
+static void scanLaterZset(robj *ob, unsigned long *cursor) {
     if (ob->type != OBJ_ZSET || ob->encoding != OBJ_ENCODING_SKIPLIST) return;
     zset *zs = (zset *)ob->ptr;
     dict *d = zs->dict;
@@ -430,13 +514,13 @@ void scanLaterZset(robj *ob, unsigned long *cursor) {
 }
 
 /* Used as scan callback when all the work is done in the dictDefragFunctions. */
-void scanCallbackCountScanned(void *privdata, const dictEntry *de) {
+static void scanCallbackCountScanned(void *privdata, const dictEntry *de) {
     UNUSED(privdata);
     UNUSED(de);
     server.stat_active_defrag_scanned++;
 }
 
-void scanLaterSet(robj *ob, unsigned long *cursor) {
+static void scanLaterSet(robj *ob, unsigned long *cursor) {
     if (ob->type != OBJ_SET || ob->encoding != OBJ_ENCODING_HT) return;
     dict *d = ob->ptr;
     dictDefragFunctions defragfns = {.defragAlloc = activeDefragAlloc,
@@ -444,7 +528,7 @@ void scanLaterSet(robj *ob, unsigned long *cursor) {
     *cursor = dictScanDefrag(d, *cursor, scanCallbackCountScanned, &defragfns, NULL);
 }
 
-void scanLaterHash(robj *ob, unsigned long *cursor) {
+static void scanLaterHash(robj *ob, unsigned long *cursor) {
     if (ob->type != OBJ_HASH || ob->encoding != OBJ_ENCODING_HT) return;
     dict *d = ob->ptr;
     dictDefragFunctions defragfns = {.defragAlloc = activeDefragAlloc,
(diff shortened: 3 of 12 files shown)


madolson:
@zvi-code Would you mind also helping taking a look on this PR.

src/defrag.c
```diff
@@ -792,35 +838,26 @@ void defragPubsubScanCallback(void *privdata, const dictEntry *de) {
...
-            dictEntry *pubsub_channel = dictFind(pubsub_ctx->clientPubSubChannels(c), newchannel);
+            dict *client_channels = ctx->getPubSubChannels(c);
+            dictEntry *pubsub_channel = dictFind(client_channels, newchannel);
             serverAssert(pubsub_channel);
-            dictSetKey(pubsub_ctx->clientPubSubChannels(c), pubsub_channel, newchannel);
+            dictSetKey(ctx->getPubSubChannels(c), pubsub_channel, newchannel);
         }
         dictReleaseIterator(di);
+        // Now that we're done correcting the references, we can safely free the old channel robj
+        zfree_no_tcache(channel);
```
zvi-code, line 848: why is this needed? `channel` is just the sds key and it is replaced in the `pubsub_channels` dictionary few lines above (L:832). What usage do you see after the  `activeDefragStringObEx` call?

zvi-code:
@JimB123 , I skimmed through the code, at a very high level, it looks good (and the results you posted are very promising!). I will followup with a more detailed review of the code. 
Can you, please, provide an overview of the change in how the new running time is calculated? Specifically, you removed the `server.hz` from the _timelimit_ calculation, are you saying we should stop using it in other places (like in expiry) ? is there a tradeoff ?, could there be imbalance between cron jobs that are scheduled based on different mechanisms ? Can you benchmark your change with large number of clients? [I think `dynamic_hz` is set to `true` by default]

JimB123:
@zvi-code The Valkey event loop supports creation of various timer events.  `server.hz` is used to control the rate for 1 specific timer event, the `serverCron` timer.  This timer handles various periodic maintenance events and is, by default, executed every 100ms (10 times/sec as indicated by the `server.hz` value).  It makes sense to have a single timer for these periodic maintenance event as we don't need a separate timer for every little periodic action.

But this `serverCron` is not the only timer.  Evictions, for example, use a timer with a 0ms delay - it fires continuously (every event loop cycle) if we are over maxmemory and critical eviction processing is required.  Modules have the ability to create various timers.  (At Amazon, we have other internal timers.)

For defrag, there are 2 things that need to happen.  First, we need to make a decision if we should begin to defrag.  It makes sense to decide this using the main `serverCron`.  We don't need a special timer just to check if we should perform defrag.  However, once that decision is made, and we begin the defrag process, it does not make sense to perform the defragmentation using the 100ms server cron.  A dedicated timer is warranted - just like is done with eviction.

In the old code, if we target 10% of the CPU, that was done by letting defrag run for (a continuous) 10ms every time the 100ms `serverCron` fired.  That's just a bad idea!

In the new code, with the same 10% CPU target, we run the defrag job for only 500us, but schedule it on it's own dedicated timer so that it can run more often.  The code modulates the frequency rather than the duration.  (In extreme cases, anti-starvation must still modulate the duration as starvation impacts the frequency.)

>are you saying we should stop using [`server.hz`] in other places (like in expiry) ?

I haven't looked at the lazy expiration before, but looking at it now, that's exactly what I'm saying.  We shouldn't be doing ANYTHING in `serverCron` that attempts to use a measured percentage of the CPU.  The server cron should remain a fast and predictable timer.

>I think `dynamic_hz` is set to `true` by default

OK, wow, this is another thing I haven't looked at before.  IMO, `dynamic_hz` is misguided.  We have this nice periodic job (serverCron) that runs every 100ms.  But one of the things it does is to perform certain actions that iterate over a variable list of clients.  It doesn't want to create latency - so it only iterates over a few clients each cycle.  BUT - now it might take a very long time to iterate over the list of clients, given the 100ms frequency.  So we've decided to perform `serverCron` more often.  This means that we are performing a lot of miscellaneous tasks (like checking if we need to defrag) more often - just because there are more clients?  That doesn't make sense.  A better solution would be to have a separate timer for a clients-cron.  And THAT timer could be run at a higher rate (and dynamically adjusted) based on the number of clients.

This `dynamic_hz` creates the very weird behavior that with the OLD defrag, if there are more clients, latency caused by defrag would be reduced.  That's a super strange behavior.  In the new defrag code, the number of clients doesn't affect defrag latencies.  It's now consistently low.  In fact, the default 500us setting will result in latencies that are half of the lowest possible (1%) setting on the old defrag (at the standard HZ rate).

src/ae.c
```diff
@@ -85,7 +85,7 @@ aeEventLoop *aeCreateEventLoop(int setsize) {
     if (eventLoop->events == NULL || eventLoop->fired == NULL) goto err;
     eventLoop->setsize = setsize;
     eventLoop->timeEventHead = NULL;
-    eventLoop->timeEventNextId = 0;
+    eventLoop->timeEventNextId = 1; // nicer for callers if we don't use 0
```
madolson: I don't understand what this comment means.

valkey.conf
```diff
@@ -2300,9 +2300,8 @@ rdb-save-incremental-fsync yes
 # Fragmentation is a natural process that happens with every allocator (but
 # less so with Jemalloc, fortunately) and certain workloads. Normally a server
 # restart is needed in order to lower the fragmentation, or at least to flush
-# away all the data and create it again. However thanks to this feature
-# implemented by Oran Agra, this process can happen at runtime
```
madolson, line 2337: lol, I never even noticed this was here.

valkey.conf
```diff
@@ -2341,17 +2340,24 @@ rdb-save-incremental-fsync yes
 # active-defrag-threshold-upper 100
 
 # Minimal effort for defrag in CPU percentage, to be used when the lower
-# threshold is reached
-# active-defrag-cycle-min 1
+# threshold is reached.
+# Note: this is a synonym for the deprecated name "active-defrag-cycle-min"
```
madolson: I don't see that compelling of a reason to rename these configs, they still feel okay to keep in the current context. I don't really like the originals, but it's worth mentioning that deprecating will force rewrite them in the config if they do `config rewrite`, so there is some risk to it.

src/defrag.c
```diff
@@ -924,219 +1320,48 @@ void computeDefragCycles(void) {
...
 
-            /* Check if all defragmentation stages have been processed.
-             * If so, mark as finished and reset the stage counter to move on to next database. */
-            if (defrag_stage == num_stages) {
-                all_stages_finished = 1;
-                defrag_stage = 0;
-            }
+    // Historically, server.active_defrag_running is both a logical boolean (indicating that a cycle
+    //  is running), and also an integer representation of the CPU percentage that should be applied.
+    //  A non-zero value here means that a cycle should be started, if it's not already running.
```
madolson:
```suggestion
    /* server.active_defrag_running is both a logical boolean (indicating that a cycle
    /* is running), and also an integer representation of the CPU percentage that should be applied.
    /* A non-zero value here means that a cycle should be started, if it's not already running. */
```

Historically is an awkward word choice here since it implies it's no longer true. 

Second more wider point, the general preference is to use c style comments `/* comment */` for everything,  especially multi-line comments but including single line comments.

src/defrag.c
```diff
@@ -38,27 +38,97 @@
...
+ *                cleared, preparing to begin the stage with a new target.
+ *  target      - This is the "thing" that should be defragged.  It's type is dependent on the
+ *                type of the stage function.  This might be a dict, a kvstore, a DB, or other.
+ *  privdata    - A pointer to arbitrary private data which is unique to the stage function.
+ *
+ * Returns:
+ *  - true if the stage is complete
+ *  - false if there is more work to do
+ */
+typedef bool (*defragStageFn)(monotime endtime, void *target, void *privdata);
```
madolson: Still don't use bool in the project, use int.

src/defrag.c
```diff
@@ -844,76 +882,434 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+                                     kvstore *kvs,
+                                     dictScanFunction scan_fn,
+                                     kvstoreHelperPreContinueFn precontinue_fn,
+                                     const dictDefragFunctions *defragfns,
+                                     void *privdata) {
+    static kvstoreIterState state; // STATIC - this persists
+    if (endtime == 0) {
+        // Starting the stage, set up the state information for this stage
+        state.kvs = kvs;
+        state.slot = -2;
```
madolson:
```suggestion
        state.slot = KVSTORE_ITERATOR_SLOT_INIT;
```

Or something to make this clearer instead of a magic number.

src/defrag.c
```diff
@@ -844,76 +882,434 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+        otherStep = 0;
+        return false;
+    }
+
+    // TODO: These functions aren't timebound
+
+    /* there are many more pointers to defrag (e.g. client argv, output / aof buffers, etc.
+     * but we assume most of these are short lived, we only need to defrag allocations
+     * that remain static for a long time */
+    switch (otherStep++) {
```
madolson: Why aren't these two separate stages? I don't really get why we did the work to generalize the stages then stuff two unrelated stages here. I can understand if it's just not really worth it.

src/defrag.c
```diff
@@ -844,76 +882,434 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
-                        server.stat_active_defrag_key_misses++;
-                    return 1;
-                }
-                iterations = 0;
-                prev_defragged = server.stat_active_defrag_hits;
-                prev_scanned = server.stat_active_defrag_scanned;
+            state.cursor = kvstoreDictLUTDefrag(kvs, state.cursor, dictDefragTables);
+            if (getMonotonicUs() >= endtime) return false;
+        } while (state.cursor != 0);
+        state.slot = -1; // Done with defrag of main structures
```
madolson:
```suggestion
        state.slot = KVSTORE_ITERATOR_SLOT_DEFRAG_SLOTS_INIT;
```

src/defrag.c
```diff
@@ -844,76 +882,434 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+
+    static defragKeysCtx ctx; // STATIC - this persists
+    if (endtime == 0) {
+        ctx.db = db;
+        // Don't return yet.  Call the helper with endtime==0 below.
+    }
+    serverAssert(ctx.db == db);
+
+    // Note: for DB keys, we use the start/finish callback to fix an expires table entry if
+    //       after the main DB entry is moved.
```
madolson:
```suggestion
    // Note: for DB keys, we use the start/finish callback to fix an expires table entry if
    //       the main DB entry is moved.
```

Feels like an extra word, after could also work too, but if after doesn't make sense.

src/defrag.c
```diff
@@ -844,76 +882,434 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+}
+
+
+static bool defragStagePubsubKvstore(monotime endtime, void *target, void *privdata) {
+    // target is server.pubsub_channels or server.pubsubshard_channels
+    getClientChannelsFnWrapper *fnWrapper = privdata;
+
+    static const dictDefragFunctions defragfns = {
+        .defragAlloc = activeDefragAlloc,
+        .defragKey = NULL, // Handled by
```
madolson:
```suggestion
        .defragKey = NULL, // Handled by defragPubsubScanCallback
```

src/defrag.c
```diff
@@ -844,76 +882,434 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+
+typedef struct {
+    defragStageFn stage_fn; // The function to be invoked for the stage
+    void *target;           // The target that the function will defrag
+    void *privdata;         // Private data, unique to the stage function
+} StageDescriptor;
+
+struct DefragContext {
+    monotime start_scan;
+    long long start_stat;
```
madolson:
```suggestion
    long long start_defrag_hits;
```

src/defrag.c
```diff
@@ -844,76 +882,434 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+        // Defrag is being terminated abnormally
+        aeDeleteTimeEvent(server.el, defrag.timeproc_id);
+        defrag.timeproc_id = AE_DELETED_EVENT_ID;
+
+        if (defrag.current_stage) {
+            zfree(defrag.current_stage);
+            defrag.current_stage = NULL;
+        }
+
+        listSetFreeMethod(defrag.remaining_stages, zfree);
```
madolson, line 1129: This is much more commonly done set the list is created than at the end. You can use unlink instead of delete to remove an item from a list without calling free.

src/defrag.c
```diff
@@ -924,219 +1320,48 @@ void computeDefragCycles(void) {
...
-                if (listLength(db->defrag_later) > 0) {
-                    defrag_later_item_in_progress = 1;
-                    continue;
-                }
-
-                /* Move to the next slot in the current stage. If we've reached the end, move to the next stage. */
-                if ((slot = kvstoreGetNextNonEmptyDictIndex(current_stage->kvs, slot)) == -1) defrag_stage++;
-                defrag_later_item_in_progress = 0;
-            }
+    computeDefragCpuPercent(); // updates server.active_defrag_running (historical)
```
madolson: Do we want to update the name of server.active_defrag_running to server.active_defrag_cpu_percent? A value of zero makes sense to be disabled.

src/defrag.c
```diff
@@ -844,76 +882,434 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+    serverAssert(defrag.current_stage || listLength(defrag.remaining_stages) > 0);
+
+    if (hasActiveChildProcess()) {
+        // If there's a child process, pause the defrag, polling until the child completes.
+        defrag.timeproc_end_time = 0; // prevent starvation recovery
+        return 100;
+    }
+
+    monotime starttime = getMonotonicUs();
+    monotime endtime = starttime + computeDutyCycleUs();
```
madolson: My general understanding is a duty cycle is a ratio, so this function is a bit confusing at first read. Maybe something like `computeDefragCylceRunTime`.

src/defrag.c
```diff
@@ -924,219 +1320,48 @@ void computeDefragCycles(void) {
...
-            defrag_later_cursor = 0;
-            current_db = -1;
-            defrag_stage = 0;
-            defrag_cursor = 0;
-            slot = -1;
-            defrag_later_item_in_progress = 0;
-            db = NULL;
-            goto update_metrics;
-        }
+        if (defragIsRunning()) endDefragCycle();
```
madolson: Does it make sense to just endDefragCycle at the start of the main defrag timer proc so there is only one exit path for defrag? Then we don't have to track the timer ID and active defrag will stop immediately when disabled.

src/defrag.c
```diff
@@ -844,76 +882,434 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+              (int)elapsedMs(defrag.start_scan), (int)(server.stat_active_defrag_hits - defrag.start_stat),
+              frag_pct, frag_bytes);
+
+    server.stat_total_active_defrag_time += elapsedUs(server.stat_last_active_defrag_time);
+    server.stat_last_active_defrag_time = 0;
+    server.active_defrag_running = 0;
+}
+
+
+// Called at start of timeProc!
```
madolson: I would prefer us not to have comments describing how a function is currently used (although it's done quite often), since they often get out of sync.

madolson (commented):
Mostly looks good, much more readable than before! Some minor comments.

src/ae.c
JimB123: I can get rid of the comment.

If a caller wants to keep track if a timer is registered (or not), 0 can't currently be used to indicate no timer.  We can use a more arbitrary -1, but since, by default, static memory is initialized to 0, it just makes sense to reserve this for no timer.

And since this is just an incrementing counter, there's no downside to starting at 1 (vs starting at 0).

src/defrag.c
JimB123: It's time to end the hate for bool.  https://github.com/orgs/valkey-io/discussions/1278

Alternatively, I can create a enum (done/not-done), but I pretty strongly oppose using an arbitrary integer.

src/defrag.c
JimB123: I confess, I didn't like the arbitrary number either.  But we were already using an arbitrary -1, so I figured what the heck.  I'll clean it up.

src/defrag.c
JimB123: Essentially historical.  The old code lumps these together.  Originally, the code did them both as a single step, I just added the step construct.  Thought this might be a simpler way to add small misc stuff rather than creating a new stage.  But you're right, it could/should(?) be 2 separate stages.

src/defrag.c
JimB123: The name was carried over from the old code, but you're right, it's not very expressive.

src/defrag.c
JimB123: With the normal processing flow, the list is empty (as asserted below on 1127).  It's only in the unusual case of an abort that there's anything remaining.  This seemed to be the quickest way to allow the list to flush.

You're right that I could set this initially on the list, but that would complicate the normal flow.  I'd either need to null out the reference in the list node before calling `listDelNode` or I could use `listUnlinkNode` and then follow that with an explicit free of the node (not freeing the referenced value).  Not sure either of those options are inherently better than the current code.

src/defrag.c
JimB123: I agree, it's not clear.  I'll update the comment to indicate:
```
// Must be called at the start of the timeProc as this computes the time from the end of the previous timeProc invocation.
```

src/defrag.c
JimB123: Yes, I think that's better locality.

src/defrag.c
JimB123: I was thinking about this also.  I only hesitated because I had to stop somewhere.  Will update.

src/defrag.c
JimB123: I'll fix the variable name.

Re: comments, I'll address all the multi-line comments.  However we ARE using `//` in a number of places throughout the code.  Is this something like the "bool hate"?

valkey.conf
JimB123: ok - I'll revert this and just update the comments to indicate that it is not related to the "cycle" time but rather indicates CPU.

JimB123 (commented):
Replied to some of @madolson comments.  Will publish update.

src/ae.c
madolson: Yeah. I see what you are going for now, I would have assumed -1 to be used for "invalid ID", it's some amount of undocumented behavior, but I'm okay with it for now.

src/defrag.c
madolson: You're right, I guess I won't block on changing it.

src/defrag.c
madolson: OK, again not really a blocker for me.

madolson (approved):
Lgtm. @zvi-code would still like you to take a look if you have time.

madolson:
@valkey-io/core-team there is a small major change here in the form of a config that determines how much time we spend on active Defrag. Please take a look and provide 👍  or 👎 . You can see the definition here:
```
# The time spent (in microseconds) of the periodic active defrag process.  This
# affects the latency impact of active defrag on client commands.  Smaller numbers
# will result in less latency impact at the cost of increased defrag overhead.
# active-defrag-cycle-us 500
```

Other than that, this change just better amortizes the active-defrag work to reduce latency spikes.

src/defrag.c
```diff
@@ -844,299 +887,474 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+    int overage = defrag.timeproc_end_time - intendedEndtime;
+    defrag.timeproc_overage_us += overage; // track over/under desired CPU
+
+    int targetCpuPercent = server.active_defrag_cpu_percent;
+    serverAssert(targetCpuPercent > 0 && targetCpuPercent < 100);
+
+    // Given the desired duty cycle, what inter-cycle delay do we need to achieve that?
+    long totalCycleTimeUs = server.active_defrag_cycle_us * 100 / targetCpuPercent;
+    long delayUs = totalCycleTimeUs - server.active_defrag_cycle_us;
+    // Only increase delay by the fraction of the overage that would be non-duty-cycle
```
enjoy-binbin, line 1224: delayUs or delay_us and // or /**/

I guess we need to somehow unify our style.

JimB123: AFAIK, there is no published set of style requirements, and the code is consistent with other similar examples throughout the Valkey code base.

src/defrag.c
```diff
@@ -844,299 +887,474 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+    if (cpu_pct > server.active_defrag_cpu_percent || server.active_defrag_configuration_changed) {
+        server.active_defrag_configuration_changed = 0;
+        if (defragIsRunning()) {
+            serverLog(LL_VERBOSE, "Changing active defrag CPU, frag=%.0f%%, frag_bytes=%zu, cpu=%d%%",
+                      frag_pct, frag_bytes, cpu_pct);
+        } else {
+            serverLog(LL_VERBOSE, "Starting active defrag, frag=%.0f%%, frag_bytes=%zu, cpu=%d%%",
+                      frag_pct, frag_bytes, cpu_pct);
+        }
+        return cpu_pct;
```
zvi-code: This code is very weird code, I suggest to just set the value of `server.active_defrag_cpu_percent` internally as it was

src/defrag.c
```diff
@@ -844,299 +887,474 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+    UNUSED(id);
+    UNUSED(clientData);
 
-            /* Check if all defragmentation stages have been processed.
-             * If so, mark as finished and reset the stage counter to move on to next database. */
-            if (defrag_stage == num_stages) {
-                all_stages_finished = 1;
-                defrag_stage = 0;
-            }
+    // This function shouldn't be called unless there's work to do.
```
zvi-code: I find this confusing, this function is called from event loop, so it's not that someone could call it when in wrong state... why is it needed?

src/defrag.c
```diff
@@ -844,299 +887,474 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+// Note: target is a DB, (not a KVS like most stages)
+static doneStatus defragStageDbKeys(monotime endtime, void *target, void *privdata) {
+    UNUSED(privdata);
+    serverDb *db = (serverDb *)target;
 
-    /* Once a second, check if the fragmentation justfies starting a scan
-     * or making it more aggressive. */
-    run_with_period(1000) {
-        computeDefragCycles();
+    static defragKeysCtx ctx; // STATIC - this persists
```
zvi-code, line 1032: why we need all these static? can't we avoid them and carry the context with the already existing context in privatdata (or similar)?

src/defrag.c
```diff
@@ -844,299 +887,474 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+    monotime endtime = starttime + computeDefragCycleUs();
+
+    mstime_t latency;
+    latencyStartMonitor(latency);
+
+    if (!defrag.current_stage) {
+        defrag.current_stage = listNodeValue(listFirst(defrag.remaining_stages));
+        listDelNode(defrag.remaining_stages, listFirst(defrag.remaining_stages));
+        // Initialize the stage with endtime==0
+        doneStatus status = defrag.current_stage->stage_fn(0, defrag.current_stage->target, defrag.current_stage->privdata);
```
zvi-code, line 1264: can't we just have explicit init function that will be called when state is removed from the list and set to start? It makes the code less clear and mixes 2 totally separate logic paths

src/defrag.c
```diff
@@ -844,299 +887,474 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
-        if (server.stat_last_active_defrag_time == 0) elapsedStart(&server.stat_last_active_defrag_time);
-    } else if (server.stat_last_active_defrag_time != 0) {
-        server.stat_total_active_defrag_time += elapsedUs(server.stat_last_active_defrag_time);
-        server.stat_last_active_defrag_time = 0;
+    if (defrag.current_stage || listLength(defrag.remaining_stages) > 0) {
+        return computeDelayMs(endtime);
+    } else {
+        /* The timeproc will go away when we return AE_NOMORE.
+         * Mark this now, indicating a normal termination. */
+        defrag.timeproc_id = AE_DELETED_EVENT_ID;
```
zvi-code: why is it needed? don't you override it anyway in `endDefragCycle`?

src/defrag.c
```diff
@@ -844,299 +887,474 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+
+
+void monitorActiveDefrag(void) {
+    if (!server.active_defrag_enabled) return;
+
+    /* Defrag gets paused while a child process is active.  So there's no point in starting a new
+     *  cycle or adjusting the CPU percentage for an existing cycle. */
+    if (hasActiveChildProcess()) return;
+
+    server.active_defrag_cpu_percent = computeDefragCpuPercent();
```
zvi-code: This will make us query jemalloc [in `zmalloc_get_allocator_info`] every time the `databasesCron` is called, previous code tried to avoid this frequent call, is this change intentional?

src/defrag.c
```diff
@@ -844,299 +887,474 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+struct DefragContext {
+    monotime start_scan;
+    long long start_defrag_hits;
+    list *remaining_stages;
+    StageDescriptor *current_stage;
+
+    long long timeproc_id;
+    monotime timeproc_end_time;
+    long timeproc_overage_us;
+};
```
zvi-code: can we have explanation on the structs and their members?
can we have all structs defined at the beginning together with the documentation you already have there?

src/defrag.c
```diff
@@ -844,299 +887,474 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+static struct DefragContext defrag;
```
zvi-code: would also move to the begining

zvi-code:
@JimB123 , added several comments, will continue to complete the review tomorrow

src/defrag.c
JimB123: An attempt to reduce side-effects/globals.  But you're right, the log is already a side-effect.  I renamed the function and updated the global as you suggested.

src/defrag.c
JimB123: Since the timerproc should only be registered with the event loop if there's work to be done, it would indicate a coding error (not properly registering or deregistering the timerproc) if there's nothing to be done.  I updated the comment to clarify.

src/defrag.c
JimB123: This is the equivalent to each stage having global variables.  Each stage is unique and may have it's own globals. 
 Placing these directly within the stage function better scopes the variable, making the code more modular and easier to maintain.

src/defrag.c
JimB123: Since each stage is independent, each stage would need it's own initialization.  That would make each stage 2 functions (rather than 1).  It would also require making the stage specific states global (less scoped).  While a bit strange, I think this is the most modular solution.  It's not too different from `strtok()` which passes a string on the first call (for initialization) and null on subsequent calls.

src/defrag.c
JimB123: You're right, it was a bit weird.  I was clearing `timeproc_id` so that `endDefragCycle` would recognize defrag as not-running and would consider it a normal termination.  I fixed this by adding an explicit parameter for `normal_termination`, and then clear `timeproc_id` in one place.

src/defrag.c
JimB123: It is intentional.  `databasesCron` is called on a 100ms timer.  There's little benefit to trying to defer that call.  And it makes the code clearer.  This eliminates the old code that checked for modification of config values.

Note:  After reviewing your PR, we should update the zmalloc code to use your fancy MIB mechanism.  (Separate PR.)

src/defrag.c
```diff
@@ -38,27 +38,124 @@
 
 #ifdef HAVE_DEFRAG
 
-typedef struct defragCtx {
-    void *privdata;
+typedef enum { DEFRAG_NOT_DONE = 0,
+               DEFRAG_DONE = 1 } doneStatus; // Clearly not a "bool"
```
madolson:
```suggestion
               DEFRAG_DONE = 1 } doneStatus;
```

src/defrag.c
```diff
@@ -118,60 +233,47 @@ sds activeDefragSds(sds sdsptr) {
...
-            /* The sds is embedded in the object allocation, calculate the
-             * offset and update the pointer in the new allocation. */
-            long ofs = (intptr_t)ob->ptr - (intptr_t)ob;
-            if ((ret = activeDefragAlloc(ob))) {
-                ret->ptr = (void *)((intptr_t)ret + ofs);
-            }
-        } else if (ob->encoding != OBJ_ENCODING_INT) {
-            serverPanic("Unknown string encoding");
-        }
+    robj *new_robj = activeDefragAllocWithoutFree(ob);
```
madolson:
```suggestion
    robj *new_robj = activeDefragAllocWithoutFree(ob, NULL);
```

src/defrag.c
```diff
@@ -118,60 +233,47 @@ sds activeDefragSds(sds sdsptr) {
...
- * returns NULL in case the allocation wasn't moved.
- * when it returns a non-null value, the old pointer was already released
+ * Returns NULL in case the allocation wasn't moved.
+ * When it returns a non-null value, the old pointer was already released
  * and should NOT be accessed. */
 robj *activeDefragStringOb(robj *ob) {
-    return activeDefragStringObEx(ob, 1);
+    if (ob->refcount != 1) return NULL; // Unsafe to defrag if multiple refs
+    robj *new_robj = activeDefragStringObWithoutFree(ob);
+    if (new_robj) zfree_no_tcache(ob);
```
madolson:
```suggestion
    size_t size;
    if (ob->refcount != 1) return NULL; // Unsafe to defrag if multiple refs
    robj *new_robj = activeDefragStringObWithoutFree(ob, &size);
    if (new_robj) allocatorDefragFree(ob, size);
```

madolson (commented):
Just cleaning up merge conflicts.

src/defrag.c
```diff
@@ -38,23 +38,124 @@
...
+
+
+/* this method was added to jemalloc in order to help us understand which
+ * pointers are worthwhile moving and which aren't */
+int je_get_defrag_hint(void *ptr);
+
+/* Defrag function which allocates and copies memory if needed, but DOESN'T free the old block.
+ * It is the responsibility of the caller to free the old block if a non-NULL value (new block)
+ * is returned. (Returns NULL if no relocation was needed.) allocation_size will be set if a new
+ * memory block is resturned to allow for optimizations in freeing of the memory.
```
madolson:
```suggestion
 * is returned. (Returns NULL if no relocation was needed.) allocation_size will be set to the
 * size of the pointer if a new memory block is returned to allow for optimizations in freeing
 * of the memory.
```

src/defrag.c
```diff
@@ -768,18 +850,17 @@ float getAllocatorFragmentation(size_t *out_frag_bytes) {
...
+static void defragPubsubScanCallback(void *privdata, const dictEntry *de) {
+    defragPubSubCtx *ctx = privdata;
+    kvstore *pubsub_channels = ctx->kvstate.kvs;
     robj *newchannel, *channel = dictGetKey(de);
     dict *newclients, *clients = dictGetVal(de);
 
     /* Try to defrag the channel name. */
     serverAssert(channel->refcount == (int)dictSize(clients) + 1);
-    newchannel = activeDefragStringObEx(channel, dictSize(clients) + 1);
+    newchannel = activeDefragStringObWithoutFree(channel);
```
madolson:
```suggestion
    size_t size;
    newchannel = activeDefragStringObWithoutFree(channel, &size);
```

src/defrag.c
```diff
@@ -788,35 +869,26 @@ void defragPubsubScanCallback(void *privdata, const dictEntry *de) {
...
-            dictEntry *pubsub_channel = dictFind(pubsub_ctx->clientPubSubChannels(c), newchannel);
+            dict *client_channels = ctx->getPubSubChannels(c);
+            dictEntry *pubsub_channel = dictFind(client_channels, newchannel);
             serverAssert(pubsub_channel);
-            dictSetKey(pubsub_ctx->clientPubSubChannels(c), pubsub_channel, newchannel);
+            dictSetKey(ctx->getPubSubChannels(c), pubsub_channel, newchannel);
         }
         dictReleaseIterator(di);
+        // Now that we're done correcting the references, we can safely free the old channel robj
+        zfree_no_tcache(channel);
```
madolson:
```suggestion
        allocatorDefragFree(channel, size);
```

src/defrag.c
```diff
@@ -118,60 +234,48 @@ sds activeDefragSds(sds sdsptr) {
...
-    /* try to defrag robj (only if not an EMBSTR type (handled below). */
-    if (ob->type != OBJ_STRING || ob->encoding != OBJ_ENCODING_EMBSTR) {
-        if ((ret = activeDefragAlloc(ob))) {
-            ob = ret;
-        }
+/* Performs defrag on a string-type (or generic) robj, but does not free the old robj.  This is the
+ * caller's responsibility.  This is necessary for string objects with multiple references.  In this
+ * case the caller can fix the references before freeing the original object.
+ */
+static robj *activeDefragStringObWithoutFree(robj *ob) {
```
madolson:
```suggestion
static robj *activeDefragStringObWithoutFree(robj *ob, size_t *allocation_size) {
```

src/defrag.c
```diff
@@ -118,60 +234,48 @@ sds activeDefragSds(sds sdsptr) {
...
-            /* The sds is embedded in the object allocation, calculate the
-             * offset and update the pointer in the new allocation. */
-            long ofs = (intptr_t)ob->ptr - (intptr_t)ob;
-            if ((ret = activeDefragAlloc(ob))) {
-                ret->ptr = (void *)((intptr_t)ret + ofs);
-            }
-        } else if (ob->encoding != OBJ_ENCODING_INT) {
-            serverPanic("Unknown string encoding");
-        }
+    robj *new_robj = activeDefragAllocWithoutFree(ob, NULL);
```
madolson:
```suggestion
    robj *new_robj = activeDefragAllocWithoutFree(ob, allocation_size);
```

madolson (commented):
A few more merge conflict resolutions.

madolson (commented):
I think this looks good, just waiting for Zvi review and getting TSC consensus. I pinged them on the slack as well to take a look here.

zuiderkwast:
> @valkey-io/core-team there is a small major change here in the form of a config that determines how much time we spend on active Defrag. Please take a look and provide 👍 or 👎 . You can see the definition here:
> 
> ```
> # The time spent (in microseconds) of the periodic active defrag process.  This
> # affects the latency impact of active defrag on client commands.  Smaller numbers
> # will result in less latency impact at the cost of increased defrag overhead.
> # active-defrag-cycle-us 500
> ```
> 
> Other than that, this change just better amortizes the active-defrag work to reduce latency spikes.

This config is similar to how we control active rehashing and active expire. I just want to put it into context to confirm this is the config we want.

In general, we have `hz` to control how often cron runs. Smaller hz means less overhead but longer latency spikes.

Binbin recently adjusted active rehashing to do work based on `hz` in #977. The higher the `hz`, the shorter time active rehashing runs each time. Active rehashing only has a yes/no config, and the time spent is fixed to 1% of the CPU time after that PR.

For active defrag, we already have `active-defrag-cycle-max` which controls the max percentage of CPU time spent on active defrag. It seem to me that time spent on each cron cycle can be computed from `active-defrag-cycle-max` and `hz`. Then, what's the purpose of `active-defrag-cycle-us`?

madolson:
> For active defrag, we already have active-defrag-cycle-max which controls the max percentage of CPU time spent on active defrag. It seem to me that time spent on each cron cycle can be computed from active-defrag-cycle-max and hz. Then, what's the purpose of active-defrag-cycle-us?

The hz mechanism can get very out of sync under high load, and it seems more straightforward to decouple the active-defrag work from the main server cron and control it separately and let it precisely control it's usage. Active defrag does a lot more work than the rehashing cron, so was causing higher latency spikes. Anything else you want to add @JimB123 ?

zuiderkwast:
> > For active defrag, we already have active-defrag-cycle-max which controls the max percentage of CPU time spent on active defrag. It seem to me that time spent on each cron cycle can be computed from active-defrag-cycle-max and hz. Then, what's the purpose of active-defrag-cycle-us?
> 
> The hz mechanism can get very out of sync under high load, and it seems more straightforward to decouple the active-defrag work from the main server cron and control it separately and let it precisely control it's usage. Active defrag does a lot more work than the rehashing cron, so was causing higher latency spikes. Anything else you want to add @JimB123 ?

What's the relation between `active-defrag-cycle-us` and `active-defrag-cycle-max` then? Are both upper limits of different measures?

madolson:
> What's the relation between active-defrag-cycle-us and active-defrag-cycle-max then? Are both upper limits of different measures?

`active-defrag-cycle-us` is the maximum time we will spend on active defrag before going back to processing clients. `active-defrag-cycle-max` is the maximum amount of CPU time, measured by percent, we want to spend on active defrag. Jim early proposed renaming `active-defrag-cycle-max` to `active-defrag-cpu-max` in an earlier revision which I preferred to avoid, but it would clarify the situation.

zuiderkwast:
The configs limit each other then.

If you set `active-defrag-cycle-us` to 500 and `hz` to 10, you effectively don't let defrag run more than 5ms per second, which is 0.5% of the CPU time.

Even if you set `active-defrag-cycle-max` to 10%, CPU time is limited to 0.5% by the other configs, so cycle-max becomes meaningless, doesn't it?

Can't we just use `active-defrag-cycle-max / hz` as the time limit per cron cycle? (The unit is % of a second, so multiply by some factor to get it in microseconds.)

madolson:
> If you set active-defrag-cycle-us to 500 and hz to 10, you effectively don't let defrag run more than 5ms per second, which is 0.5% of the CPU time.

This PR decouples the cycle time from serverCron so not hit by the hz limits. It independently calculates how much time it should be spending doing defrag and schedules its own timer when active defrag is needed.

src/defrag.c
```diff
@@ -840,299 +914,450 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+         *  latency, because clients are already being impacted by the long cycle time which caused
+         *  the starvation of the timer. */
+        dutyCycleUs = targetCpuPercent * waitedUs / (100 - targetCpuPercent);
+
+        // Also adjust for any accumulated overage(underage).
+        dutyCycleUs -= defrag.timeproc_overage_us;
+        defrag.timeproc_overage_us = 0;
+
+        if (dutyCycleUs < server.active_defrag_cycle_us) {
+            /* We never reduce our cycle time, that would increase overhead.  Instead, we track this
```
zvi-code, line 1198: This is a little not clear. `server.active_defrag_cycle_us` is actually a min, right? when you say "we never reduce" you mean with respect to `server.active_defrag_cycle_us` and not with respect to previous cycle, is that correct?

src/defrag.c
```diff
@@ -840,299 +914,450 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+static int computeDelayMs(monotime intendedEndtime) {
+    defrag.timeproc_end_time = getMonotonicUs();
+    int overage = defrag.timeproc_end_time - intendedEndtime;
+    defrag.timeproc_overage_us += overage; // track over/under desired CPU
+
+    int targetCpuPercent = server.active_defrag_cpu_percent;
+    serverAssert(targetCpuPercent > 0 && targetCpuPercent < 100);
+
+    // Given the desired duty cycle, what inter-cycle delay do we need to achieve that?
+    long totalCycleTimeUs = server.active_defrag_cycle_us * 100 / targetCpuPercent;
```
zvi-code, line 1222: can you explain the calculation here? why are you looking at `server.active_defrag_cycle_us` and not the actual `dutyCycleUs`? An explanation of the overall computation here would be helpful [like you did in `computeDefragCucleUs`]

src/defrag.c
```diff
@@ -38,23 +38,125 @@
...
+// Private data for pubsub kvstores
+typedef dict *(*getClientChannelsFn)(client *);
+typedef struct {
+    getClientChannelsFn fn;
+} getClientChannelsFnWrapper;
+
+typedef struct {
+    kvstoreIterState kvstate;
+    getClientChannelsFn getPubSubChannels;
 } defragPubSubCtx;
```
zvi-code, line 140: please add static assert to enforce the assumption of `kvstoreIterState kvstate;` at the begining of `defragPubSubCtx` and `defragKeysCtx`

src/defrag.c
```diff
@@ -840,299 +914,450 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+
+/* This helper function handles most of the work for iterating over a kvstore.  'privdata', if
+ * provided, MUST begin with 'kvstoreIterState' and this part is automatically updated by this
+ * function during the iteration. */
+static doneStatus defragStageKvstoreHelper(monotime endtime,
+                                           kvstore *kvs,
+                                           dictScanFunction scan_fn,
+                                           kvstoreHelperPreContinueFn precontinue_fn,
+                                           const dictDefragFunctions *defragfns,
+                                           void *privdata) {
```
zvi-code, line 969: I see that `privdata` internal value is set in this function but it's not really used cross iterations or by caller in the upper level, so why are you setting the state in privdata? why not just pass the state to `precontinue_fn`? and not set it on privdata?

src/defrag.c
```diff
@@ -840,299 +914,450 @@ int defragLaterItem(dictEntry *de, unsigned long *cursor, long long endtime, int
...
+         *  Solve:    D          P
+         *          -----   =  -----
+         *          D + W       100
+         *  Solving for D:
+         *     D = P * W / (100 - P)
+         *
+         * Note that dutyCycleUs addresses starvation.  If the wait time was long, we will compensate
+         *  with a proportionately long duty-cycle.  This won't significantly affect perceived
+         *  latency, because clients are already being impacted by the long cycle time which caused
+         *  the starvation of the timer. */
```
zvi-code, line 1190: While it is correct, this is not a good behaviour, especially from tail latency perspective. A better behaviour would increase the "measure" time interval to keep smooth run time of defrag even when there exists a periodic spike. Now a sporadic long running command\cron-job will cause defrag to run in sporadic long running spikes as well, isn't it so?

zvi-code (commented):
I still have some questions about risk that this change will actually increase tail latencies in some scenarios. Added specific comments\questions in relevant code

src/defrag.c
madolson: dutyCycleUs is the upper bound of the time we can spend doing defrag at a given time, not the lower bound. We bias against latency spikes, especially when the system is busy.

zuiderkwast:
> > If you set active-defrag-cycle-us to 500 and hz to 10, you effectively don't let defrag run more than 5ms per second, which is 0.5% of the CPU time.
> 
> This PR decouples the cycle time from serverCron so not hit by the hz limits. It independently calculates how much time it should be spending doing defrag and schedules its own timer when active defrag is needed.

That explains it. Thanks!

tests/unit/memefficiency.tcl
zvi-code, line 1: I saw some tests use `            r config set hz 100` to in various phases of the test, I expect these configs be removed and adjusted to the new implementation

src/defrag.c
JimB123: The inter-cycle delay is computed in milliseconds, and is not accurate as it may be delayed.  At this point in the code, we are computing the number of micro-seconds needed to achieve the desired CPU percentage since the last cycle.  The actual computed number may be a little above/below the configured duty-cycle-us.  If we require less CPU than the duty-cycle, we don't reduce the duty-cycle.  Instead, we add to the "overage" so that we can later compensate with an extra millisecond delay when we get enough overage.

On the flip size, if the computed dutyCycleUs is greater than the configured/intended `server.active_defrag_cycle_us`, that means that we are being starved, and that the timer didn't fire when it was supposed to.  In this case, we allow the duty-cycle to run longer, to compensate for the starvation.

src/defrag.c
JimB123: `server.active_defrag_cycle_us` is the duty-cycle time.  The only time this is altered is when there has been starvation.  In the duty-cycle computation, we are keeping track of slight overages/underages so that we can compensate in the next cycle - achieving a perfect target CPU percentage.

If, for example, the `active_defrag_cycle_us` is 250, and the target CPU percentage is 10%, the TOTAL cycle time would be 2500us (and our 250us is 10% of that).  We would need to delay 2250us - which we round down to 2ms.

Of course then, if the timer pops after 2ms, we only need 200us to achieve 10% - so we will will still run for 250us (the config and 50us will get added to "overage".  Eventually, the overage will allow us to wait 3ms rather than 2ms, maintaining the perfect desired CPU percentage, using the configured duty cycle time.

(I'll add some additional comments to clarify.)

valkey.conf
```diff
@@ -2374,17 +2373,24 @@ rdb-save-incremental-fsync yes
 # active-defrag-threshold-upper 100
 
 # Minimal effort for defrag in CPU percentage, to be used when the lower
-# threshold is reached
+# threshold is reached.
+# Note: this is not actually a cycle time, but is an overall CPU percentage
 # active-defrag-cycle-min 1
```
zuiderkwast: The first sentence already says it's in CPU percentage. No need to mention the same thing twice. Edit the first sentence to make it clearer instead, if needed.

valkey.conf
```diff
@@ -2374,17 +2373,24 @@ rdb-save-incremental-fsync yes
...
 
 # Maximal effort for defrag in CPU percentage, to be used when the upper
-# threshold is reached
+# threshold is reached.
+# Note: this is not actually a cycle time, but is an overall CPU percentage
 # active-defrag-cycle-max 25
```
zuiderkwast: Ditto

src/defrag.c
JimB123: @madolson that's not quite right.  If the timer is getting starved, we will increase the duty-cycle to combat the starvation.  This won't occur unless there is a delay to the timer.

@zvi, I don't agree with this being bad behavior though.  And the tail latency is unimportant in this case.  Defrag is possibly the most critical maintenance task in the system.  With today's defrag, `serverCron` is subject to starvation.  A customer generating a series of slow commands will severely impede defrag.  There is an attempt at anti-starvation code  (which I removed) in `whileBlockedCron`.  Reference: https://github.com/valkey-io/valkey/pull/1242/files#diff-1abc5651133d108c0c420d9411925373c711133e7748d9e4f4c97d5fb543fdd9L1617-L1626
(Note - in re-reviewing that code, I think we still need a hook there - as timers aren't run while blocked.  I'll add a separate comment on that.)

Consider a command that takes 2 seconds to execute - and delay's all of the timers.  In order to get 10% of the CPU (for example), we need to run for 200ms to compensate for the starvation.  HOWEVER, that 200ms is small as compared to the 2-seconds that the original command took.  Tail latencies are dominated by the slow command, and not by active defrag.  Likewise, if the event loop is slowed due to overload, the tail latencies are dominated by the overload, not the additional fraction for defrag.

tests/unit/memefficiency.tcl
JimB123: I was trying to leave the tests as untouched as possible.  But you're correct that the intent of this is no longer meaningful.  I will remove.

src/config.c
```diff
@@ -3208,10 +3208,11 @@ standardConfig static_configs[] = {
...
     createIntConfig("tcp-keepalive", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.tcpkeepalive, 300, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("cluster-migration-barrier", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.cluster_migration_barrier, 1, INTEGER_CONFIG, NULL, NULL),
-    createIntConfig("active-defrag-cycle-min", NULL, MODIFIABLE_CONFIG, 1, 99, server.active_defrag_cycle_min, 1, INTEGER_CONFIG, NULL, updateDefragConfiguration),                 /* Default: 1% CPU min (at lower threshold) */
-    createIntConfig("active-defrag-cycle-max", NULL, MODIFIABLE_CONFIG, 1, 99, server.active_defrag_cycle_max, 25, INTEGER_CONFIG, NULL, updateDefragConfiguration),                /* Default: 25% CPU max (at upper threshold) */
+    createIntConfig("active-defrag-cycle-min", NULL, MODIFIABLE_CONFIG, 1, 99, server.active_defrag_cpu_min, 1, INTEGER_CONFIG, NULL, updateDefragConfiguration),                   /* Default: 1% CPU min (at lower threshold) */
+    createIntConfig("active-defrag-cycle-max", NULL, MODIFIABLE_CONFIG, 1, 99, server.active_defrag_cpu_max, 25, INTEGER_CONFIG, NULL, updateDefragConfiguration),                  /* Default: 25% CPU max (at upper threshold) */
```
zuiderkwast, line 3212: I know these config names are not good, but making the server variables and the config names mismatch doesn't solve the issue.

I suggest keeping the old variable names and using code comments where needed, where these variables are used in computations.

zuiderkwast (commented):
Thanks for this feature. I only reviewed the configs.

src/server.c
```diff
@@ -1609,22 +1609,6 @@ void whileBlockedCron(void) {
...
 
-    /* In some cases we may be called with big intervals, so we may need to do
-     * extra work here. This is because some of the functions in serverCron rely
-     * on the fact that it is performed every 10 ms or so. For instance, if
-     * activeDefragCycle needs to utilize 25% cpu, it will utilize 2.5ms, so we
-     * need to call it multiple times. */
-    long hz_ms = 1000 / server.hz;
-    while (server.blocked_last_cron < server.mstime) {
-        /* Defrag keys gradually. */
-        activeDefragCycle();
```
JimB123, line 1620: We should still have a hook here to ensure that we don't have defrag starvation.  Timers are not fired while blocked.  In a degenerative case - let's say an hour long script - we don't want to become completely unresponsive for 6 minutes after the script ends to achieve 10% CPU.

src/defrag.c
JimB123: done

src/defrag.c
JimB123: 1. the caller might use this in the future
2. the caller needs to send privdata to the precontinue function also

So, we could pass both the `kvstoreIterState` and the `privdata` to the precontinue function separately, but that wouldn't allow potential future access by the caller.

src/config.c
madolson: I suggested changing the variable to make the code clearer since I saw it before. I liked the change, even it's a bit confusing when you have to do the lookup between the config name and the internal variable.

zuiderkwast: So, pros and cons. Ok, I don't mind. I didn't review the code where it's used.

JimB123:
At this point, have rebased and either addressed or responded to all comments.

valkey.conf
JimB123: If we want to keep the old names "cycle-min" and "cycle-max" (as requested by Maddy), I think the note is a necessary clarification.

zuiderkwast: Keep the note, but rewrite the previous sentence and/or the note so it doesn't repeat itself.

src/defrag.c
zvi-code: @JimB123 , I think we should implement a simple mechanism that would smooth usage peaks when possible, but it's not a blocker for this PR and can be suggested in followup improvement.

zvi-code (approved):
LGTM

hwware (approved):

valkey.conf
madolson: I don't know how to reconcile these two comments. Do people like something like this?
```
# Minimal effort for defrag in CPU percentage, not cycle time as the name might
# suggest, to be used when the lower threshold is reached.
```

zuiderkwast: Yes! LGTM

zuiderkwast (approved):
Not a full review. I trust the other reviewers.

PR #1243, opened 2024-11-01
Fix timing issue in cluster-shards tests

enjoy-binbin:
The cluster-node-timeout is 3000 in our tests, the timing test wasn't succeeding, so extending the wait_for made them much more reliable.

### DIFF
--- a/tests/cluster/tests/28-cluster-shards.tcl
+++ b/tests/cluster/tests/28-cluster-shards.tcl
@@ -117,7 +117,7 @@ test "Kill a node and tell the replica to immediately takeover" {
 
 # Primary 0 node should report as fail, wait until the new primary acknowledges it.
 test "Verify health as fail for killed node" {
-    wait_for_condition 50 100 {
+    wait_for_condition 1000 50 {
         "fail" eq [dict get [get_node_info_from_shard $node_0_id 4 "node"] "health"]
     } else {
         fail "New primary never detected the node failed"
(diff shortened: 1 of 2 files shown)


enjoy-binbin:
https://github.com/valkey-io/valkey/actions/runs/11623231198/job/32369948860?pr=1238#step:9:321

```
02:47:37> Verify health as fail for killed node: FAILED: New primary never detected the node failed
(Jumping to next unit after error)
FAILED: caught an error in the test 
assertion:New primary never detected the node failed
```

tests/cluster/tests/28-cluster-shards.tcl
```diff
@@ -117,7 +117,7 @@ test "Kill a node and tell the replica to immediately takeover" {
 
 # Primary 0 node should report as fail, wait until the new primary acknowledges it.
 test "Verify health as fail for killed node" {
-    wait_for_condition 50 100 {
+    wait_for_condition 1000 50 {
```
zuiderkwast, line 120: Every 50 milliseconds, 1000 times?

Sending INFO command to valkey every 50 ms is not too often? Just wondering why you didn't keep 100 and extended the `maxtries` instead, like `wait_for_condition 500 100`.

zuiderkwast (approved):

madolson (approved):

tests/cluster/tests/28-cluster-shards.tcl
enjoy-binbin: I always prefer shorter n ms, it maybe faster. The other reason is that we have a lot of place are doing this 50ms thing.

PR #1244, opened 2024-11-01
RDMA: Prevent IO for child process

pizhenwei:
RDMA MR (memory region) is not forkable, the VMA (virtual memory area) of a MR gets empty in a child process. Prevent IO for child process to avoid server crash.

In the check for whether read and write is allowed in an RDMA connection, a check that if we're in a child process is added. If we are, the function returns an error, which will cause the RDMA client to be disconnected.

Suggested-by: Viktor Söderqvist <viktor.soderqvist@est.tech>

### DIFF
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -143,11 +143,33 @@ static void serverRdmaError(char *err, const char *fmt, ...) {
     va_end(ap);
 }
 
+static inline int connRdmaAllowCommand(void) {
+    /* RDMA MR is not accessible in a child process, avoid segment fault due to
+     * invalid MR access, close it rather than server random crash */
+    if (server.in_fork_child != CHILD_TYPE_NONE) {
+        return C_ERR;
+    }
+
+    return C_OK;
+}
+
+static inline int connRdmaAllowRW(connection *conn) {
+    if (conn->state == CONN_STATE_ERROR || conn->state == CONN_STATE_CLOSED) {
+        return C_ERR;
+    }
+
+    return connRdmaAllowCommand();
+}
+
 static int rdmaPostRecv(RdmaContext *ctx, struct rdma_cm_id *cm_id, ValkeyRdmaCmd *cmd) {
     struct ibv_sge sge;
     size_t length = sizeof(ValkeyRdmaCmd);
     struct ibv_recv_wr recv_wr, *bad_wr;
     int ret;
+
+    if (connRdmaAllowCommand()) {
+        return C_ERR;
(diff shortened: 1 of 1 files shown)


src/rdma.c
```diff
@@ -143,12 +143,34 @@ static void serverRdmaError(char *err, const char *fmt, ...) {
     va_end(ap);
 }
 
+static inline int connRdmaAllowCommand(void) {
+    /* RDMA MR is not accessible in a child process, avoid segment fault due to
+     * invalid MR access, close it rather than server random crash */
+    if (server.in_fork_child != CHILD_TYPE_NONE) {
+        return C_ERR;
+    }
```
zuiderkwast, line 151: This is a great protection! It means the IO will fail if any code tries to read or write to the RDMA client from a child process. The client will be disconnected and the server does not crash.

tests/rdma/rdma-test.c
```diff
@@ -952,6 +953,39 @@ static void *test_routine(void *arg) {
     }
     printf("Valkey Over RDMA test thread[%d] GET %d KVs [OK]\n", tid, keys);
 
+    /* # round 5, test REPLICAOF, also run REPLICAOF only once */
```
zuiderkwast: TBH, I don't understand the purpose of this test case. REPLICAOF does not communicate with the client from a child process. The command just tells the server to become a replica of another valkey server. The replication between the valkey servers is done in another connection which is not RDMA.

I think we can skip this test case and functionality in the python script to start a replica.

----

If we really want to test a command that communicates with the RDMA client from a child process, I believe the only one is PSYNC with dual channel replication and diskless replication enabled. It means that the RDMA client pretends to be a Valkey replica and sends the commands a replica would do when it starts replicating. There is no use case for this, except only to try a crash the server. It sends REPLCONF and PSYNC with various arguments.

Unfortunately, the PSYNC and REPLCONF commands are not documented, so if we want to write this test, we need to be smart and figure out exactly what a replica does in this scenario. Maybe these comments are useful, but not enough I think: https://github.com/valkey-io/valkey/blob/8.0.1/src/replication.c#L3286

So, it's not very easy to test, and with the protection that prevents the server crash, it will be no crash if a client tries to do this. Therefore, I think we don't have to do this test now. (Maybe in the future.)

pizhenwei: Sure, remove `REPLICAOF` test in the new version. Thanks!

zuiderkwast (approved):

Issue #387, opened 2024-04-26
[NEW] Support for wasm besides LUA as a scripting engine


andyxning:
**The problem/use-case that the feature addresses**
WASM is becoming more and more popular and many programming languages support WASM as an target, such as rust, golang etc. These programming languages are more used by programmers than LUA.
A description of the problem that the feature will solve, or the use-case with which the feature will be used.

**Description of the feature**

A description of what you want to happen.

**Alternatives you've considered**

Any alternative solutions or features you've considered, including references to existing open and closed feature requests in this repository.

**Additional information**

Any additional information that is relevant to the feature request.

rjd15372:
@madolson @zuiderkwast I can start doing some research for this issue if you think it's worth to pursue.

madolson:
@rjd15372 Go for it! I think there is some potential for a more performant alternative to lua.

rjd15372:
Cool, can someone assign me to this issue?

andyxning:
I am also interested in this. Ping me if you have any idea

ahmadbelb:
I am also interested in this, maybe we can collaborate !

rjd15372:
I'm doing the ground work to allow running WASM code in Valkey. I'm pretty sure that there will be a lot of things to do on top of it after the basic functionality is merged into the unstable branch.

rjd15372:
The WASM runtime engine that I'll be using is https://github.com/bytecodealliance/wasmtime

zuiderkwast:
> The WASM runtime engine that I'll be using is https://github.com/bytecodealliance/wasmtime

OK! Will you make it an optional build dependency? It's a different language (Rust) and a different license (Apache 2).

andyxning:
wasm engine does not have much influence on this. we can just run through the functionality first and benchmark it later.

andyxning:
also, any proposal for implementing or architecture this support, i think we need first also make a review on it.

rjd15372:
> 
> OK! Will you make it an optional build dependency? It's a different language (Rust) and a different license (Apache 2).

Yes, the build dependency will be optional. And the idea is to download the pre-built wasm library at valkey build time, so no need to have rust toolchain to build the wasm library.

We can link Valkey against the wasmtime static library, and therefore there's no need to package the wasmtime library along with the Valkey package.

rjd15372:
[...]

PR #1245, opened 2024-11-01
WIP: Build without Lua scripting via USE_LUA (#1204)

### LINKED ISSUE #1204: [NEW] Disable LUA integration during build
Based on my observation, there are multiple CVEs occurring due to bugs/issues found in Lua engine/debugger. If a user doesn't have any use case of using Lua scripts and if they can remove it from the Valkey binary, they don't need to address the patching of security vulnerabilities in their system. 
I found a similar request on [Stack Overflow](https://stackoverflow.com/questions/69555952/disable-lua-debugger-in-redis). Currently, a user could disable LUA usage via ACL rules by restricting EVAL/SCRIPT but I think they would still need to patch their servers within a certain period of time. 

Could we make LUA optional ? This could also help us building right set of API(s) for users who rely on LUA to address functionalities not supported by server like CAS. 

CVE(s): https://docs.aws.amazon.com/AmazonElastiCache/latest/dg/cve.html https://cve.mitre.org/cgi-bin/cvekey.cgi?keyword=Lua+Redis

neomantra:
Adds the ability to disable Lua scripting using the build
configuration `USE_LUA`.  This will prevent the building
of the Lua static library and will keep Lua and most Lua
scripting internals out of the build.

This approach strived to minimize the change surface, stubbing out
keys requried functions and otherwise `ifdef`ing out large swathes of code.

The base Lua scripting commands like `EVAL` remain intact, but reply
with an error message `Lua scripting disabled`.  `INFO` commands
still include scripting statistics to prevent breaking any DevOps scripts, etc.

There are decisions to make:

 * What do with `EVAL` commands, etc?  Currently, the command still exists but replies with an Error.

 * What to do with testing infrastructure? `EVAL` is sprinkled throughout. `grep -or EVAL tests/unit  | wc -l  ==> 125`.  Do we break tests into files with `EVAL` and ones without?

### DIFF
--- a/src/defrag.c
+++ b/src/defrag.c
@@ -301,9 +301,9 @@ void activeDefragSdsDict(dict *d, int val_type) {
     dictDefragFunctions defragfns = {
         .defragAlloc = activeDefragAlloc,
         .defragKey = (dictDefragAllocFunction *)activeDefragSds,
-        .defragVal = (val_type == DEFRAG_SDS_DICT_VAL_IS_SDS       ? (dictDefragAllocFunction *)activeDefragSds
-                      : val_type == DEFRAG_SDS_DICT_VAL_IS_STROB   ? (dictDefragAllocFunction *)activeDefragStringOb
-                      : val_type == DEFRAG_SDS_DICT_VAL_VOID_PTR   ? (dictDefragAllocFunction *)activeDefragAlloc
+        .defragVal = (val_type == DEFRAG_SDS_DICT_VAL_IS_SDS     ? (dictDefragAllocFunction *)activeDefragSds
+                      : val_type == DEFRAG_SDS_DICT_VAL_IS_STROB ? (dictDefragAllocFunction *)activeDefragStringOb
+                      : val_type == DEFRAG_SDS_DICT_VAL_VOID_PTR ? (dictDefragAllocFunction *)activeDefragAlloc
 #ifdef USE_LUA
                       : val_type == DEFRAG_SDS_DICT_VAL_LUA_SCRIPT ? (dictDefragAllocFunction *)activeDefragLuaScript
 #endif
--- a/src/defrag.c
+++ b/src/defrag.c
@@ -170,6 +170,7 @@ robj *activeDefragStringOb(robj *ob) {
     return activeDefragStringObEx(ob, 1);
 }
 
+#ifdef USE_LUA
 /* Defrag helper for lua scripts
  *
  * returns NULL in case the allocation wasn't moved.
@@ -189,6 +190,7 @@ luaScript *activeDefragLuaScript(luaScript *script) {
 
     return ret;
 }
+#endif
 
 /* Defrag helper for dict main allocations (dict struct, and hash tables).
  * Receives a pointer to the dict* and return a new dict* when the dict
@@ -302,7 +304,9 @@ void activeDefragSdsDict(dict *d, int val_type) {
         .defragVal = (val_type == DEFRAG_SDS_DICT_VAL_IS_SDS       ? (dictDefragAllocFunction *)activeDefragSds
                       : val_type == DEFRAG_SDS_DICT_VAL_IS_STROB   ? (dictDefragAllocFunction *)activeDefragStringOb
                       : val_type == DEFRAG_SDS_DICT_VAL_VOID_PTR   ? (dictDefragAllocFunction *)activeDefragAlloc
+#ifdef USE_LUA
                       : val_type == DEFRAG_SDS_DICT_VAL_LUA_SCRIPT ? (dictDefragAllocFunction *)activeDefragLuaScript
+#endif
                                                                    : NULL)};
     do {
         cursor = dictScanDefrag(d, cursor, activeDefragSdsDictCallback, &defragfns, NULL);
@@ -812,7 +816,9 @@ void defragOtherGlobals(void) {
     /* there are many more pointers to defrag (e.g. client argv, output / aof buffers, etc.
      * but we assume most of these are short lived, we only need to defrag allocations
      * that remain static for a long time */
+#ifdef USE_LUA
     activeDefragSdsDict(evalScriptsDict(), DEFRAG_SDS_DICT_VAL_LUA_SCRIPT);
+#endif
     moduleDefragGlobals();
     kvstoreDictLUTDefrag(server.pubsub_channels, dictDefragTables);
     kvstoreDictLUTDefrag(server.pubsubshard_channels, dictDefragTables);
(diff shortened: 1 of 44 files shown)


neomantra:
I see the Test Tags system now and am using that to constrain tests which utilize Lua.

zuiderkwast:
> * What do with `EVAL` commands, etc?  Currently, the command still exists but replies with an Error.

This sounds right to me.

> * What to do with testing infrastructure? `EVAL` is sprinkled throughout. `grep -or EVAL tests/unit  | wc -l  ==> 125`.  Do we break tests into files with `EVAL` and ones without?

> I see the Test Tags system now and am using that to constrain tests which utilize Lua.

Test tag sounds like the best solution. Is it feasible? (If not, maybe we can just run a few selected test suites in the CI for the build without Lua.)

neomantra:
Although I perhaps wanted `needs:lua`, I stuck with the tag `scripting`.     I went through and found all the spots where `eval` or `call` were used and where the tests were failing from needing Lua in any way.

It passes tests with both `USE_LUA=on` and `USE_LUA=off` on my Mac, but I was having problems with full `make check` on my Linux boxes... but I had the same problem on `unstable` : spewing `Port <num> was already busy, trying another port...`.   Maybe related to my IPv4-only network, will look deeper.

See [this commit](https://github.com/valkey-io/valkey/pull/1245/commits/5b6d96f0b2ad79458a3e964b6a1b1079cc8d9864) for the tag changes.

zuiderkwast:
> Although I perhaps wanted `needs:lua`, I stuck with the tag `scripting`.

I guess `needs:lua` is better. Scripting is equivalent to Lua now, but we're considering more languages for scripting (such as WASM, #387) so scripting may not always imply Lua in the future.

neomantra:
I hadn't re-run Lua-enabled after fixing some Lua-disabled tests.... this last commit corrects that with the proper empty results.

neomantra:
Sorry for the repeated build failures -- glad the CI catches it :)  

I was missing some targets and hence some Lua coverage.  I still cannot do a full test on Linux with `unstable`.  [I hoped #1243 would help, but it didn't.  Looked a little at the `find_available_port` etc  and it does feel like a port race.]

madolson:
We discussed this in an online meeting. The core team was OK with being able to remove LUA from the server, but we would like to make it more pluggable and modular. Instead of just not installing LUA, we would like to make it so that LUA is not added as an engine (similar to how functions allows you to specify the engine type) and so when you try to run EVAL you get an error like "no default engine loaded". In the future folks can use either Lua 5.4 or LuaJIT.

neomantra:
I made that error message change per your suggestion.  Some of that work (modular engine system) is beyond the scope of this PR, but I think it is a sensible approach. At least the `USE_LUA` injections and `scripting` test tagging here helps show the surface and some cross-cutting concerns.    

If you have specifics to add that are low-hanging, I could explore it.

zuiderkwast:
There is partial logic for a modular engine system. EVAL currently accepts a *shebang* that is intended to be able to select the scripting engine in the future.


```
eval.c-365-        /* Verify lua interpreter was specified */
eval.c:366:        if (strcmp(parts[0], "#!lua")) {
eval.c-367-            if (err) *err = sdscatfmt(sdsempty(), "Unexpected engine in script shebang: %s", parts[0]);
eval.c-368-            sdsfreesplitres(parts, numparts);
eval.c-369-            return C_ERR;
eval.c-370-        }
```

Here we want the module API to provide a way for modules to provide a scripting engine. I created this issue: #1261.

src/eval.c
```diff
@@ -1756,3 +1764,64 @@ void luaLdbLineHook(lua_State *lua, lua_Debug *ar) {
...
+    addReplyError(c, "No default engine loaded");
+}
+
+void evalShaRoCommand(client *c) {
+    addReplyError(c, "No default engine loaded");
+}
+
+void scriptCommand(client *c) {
+    addReplyError(c, "No default engine loaded");
+}
```
hpatro, line 1825: We would need test to cover these scenarios. So, we might need separate test suite/tag for this particular flow.

src/defrag.c
```diff
@@ -299,10 +301,12 @@ void activeDefragSdsDict(dict *d, int val_type) {
...
-        .defragVal = (val_type == DEFRAG_SDS_DICT_VAL_IS_SDS       ? (dictDefragAllocFunction *)activeDefragSds
-                      : val_type == DEFRAG_SDS_DICT_VAL_IS_STROB   ? (dictDefragAllocFunction *)activeDefragStringOb
-                      : val_type == DEFRAG_SDS_DICT_VAL_VOID_PTR   ? (dictDefragAllocFunction *)activeDefragAlloc
+        .defragVal = (val_type == DEFRAG_SDS_DICT_VAL_IS_SDS     ? (dictDefragAllocFunction *)activeDefragSds
+                      : val_type == DEFRAG_SDS_DICT_VAL_IS_STROB ? (dictDefragAllocFunction *)activeDefragStringOb
+                      : val_type == DEFRAG_SDS_DICT_VAL_VOID_PTR ? (dictDefragAllocFunction *)activeDefragAlloc
+#ifdef USE_LUA
                       : val_type == DEFRAG_SDS_DICT_VAL_LUA_SCRIPT ? (dictDefragAllocFunction *)activeDefragLuaScript
+#endif
                                                                    : NULL)};
```
hpatro, line 310: Might be just me. But `ifdef` within the struct declaration gives a hit to the readability. We can consider refactoring this.

hpatro (commented):
Looks like we allow loading of functions in case the engine doesn't support any engine. Would that cause any confusion? Same thing applies for functions arriving via replication stream.

src/defrag.c
neomantra: Maybe like?
```
    dictDefragFunctions defragfns = {
        .defragAlloc = activeDefragAlloc,
        .defragKey = (dictDefragAllocFunction *)activeDefragSds,
        .defragVal = (val_type == DEFRAG_SDS_DICT_VAL_IS_SDS     ? (dictDefragAllocFunction *)activeDefragSds
                      : val_type == DEFRAG_SDS_DICT_VAL_IS_STROB ? (dictDefragAllocFunction *)activeDefragStringOb
                      : val_type == DEFRAG_SDS_DICT_VAL_VOID_PTR ? (dictDefragAllocFunction *)activeDefragAlloc
                                                                 : NULL)};
#ifdef USE_LUA
    if (val_type == DEFRAG_SDS_DICT_VAL_LUA_SCRIPT)
        dictDefragFunctions.defragVal = (dictDefragAllocFunction *)activeDefragLuaScript;
#endif

```

zuiderkwast:
Fixed by https://github.com/valkey-io/valkey/pull/2858 and https://github.com/valkey-io/valkey/pull/3392. (The commit messages don't mention this clearly, but there are build options for building without Lua.)

PR #1247, opened 2024-11-01
Log clusterbus handshake timeout failures

bentotten:
This adds a log when a handshake fails for a timeout. This can help troubleshoot cluster asymmetry issues caused by failed MEETs

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -4872,6 +4872,8 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t handshake_
     /* A Node in HANDSHAKE state has a limited lifespan equal to the
      * configured node timeout. */
     if (nodeInHandshake(node) && now - node->ctime > handshake_timeout) {
+        serverLog(LL_WARNING, "Clusterbus handshake timeout [%s]:%d after a timeout of %lld", node->ip,
+                  node->cport, handshake_timeout);
         clusterDelNode(node);
         return 1;
     }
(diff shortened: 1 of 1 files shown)


bentotten:
Odd - will check failures shortly, passed in local fork

zuiderkwast:
> Odd - will check failures shortly, passed in local fork

I've seen those test cases fail before. I restarted the failed jobs.

enjoy-binbin (approved):
This seems ok to me. Could it be that there is a bad node that causes this log to be printed continuously?

bentotten:
> This seems ok to me. Could it be that there is a bad node that causes this log to be printed continuously?

Do you mean in situations where the MEET is resent?

src/cluster_legacy.c
```diff
@@ -4872,6 +4872,8 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t handshake_
     /* A Node in HANDSHAKE state has a limited lifespan equal to the
      * configured node timeout. */
     if (nodeInHandshake(node) && now - node->ctime > handshake_timeout) {
+        serverLog(LL_WARNING, "Clusterbus handshake timeout [%s]:%d after a timeout of %lld", node->ip,
+                  node->cport, handshake_timeout);
```
```suggestion
        serverLog(LL_WARNING, "Clusterbus handshake timeout %s:%d after %lldms", node->ip,
                  node->cport, handshake_timeout);
```
Minor suggestion to indicate clearly the timeout is in ms and remove the square bracket. There is some usage of both with and without the square bracket.

bentotten: Complete

enjoy-binbin:
> > This seems ok to me. Could it be that there is a bad node that causes this log to be printed continuously?

> Do you mean in situations where the MEET is resent?

Yes, there should only be one print in a handshake_timeout, so it should be fine i guess.

bentotten:
> > > This seems ok to me. Could it be that there is a bad node that causes this log to be printed continuously?
> 
> > Do you mean in situations where the MEET is resent?
> 
> Yes, there should only be one print in a handshake_timeout, so it should be fine i guess.

Sounds great - is this able to be merged?

madolson (approved):

PR #1249, opened 2024-11-02
Add short client info log to CLUSTER MEET / FORGET / RESET commands

enjoy-binbin:
These commands are all administrator commands. If they are operated
incorrectly, serious consequences may occur. Print the full client
info by using catClientInfoString, the info is useful when we want
to identify the source of request.

Since the origin client info is very large and might complicate the
output, we added a catClientInfoShortString function, it will only
print some basic fields, we want these fields that are useful to
identify the client. These fields are:
- id
- addr
- laddr
- connection info
- name
- user
- lib-name
- lib-ver

And also used it to replace the origin client info where it has the
same purpose. Some logging is changed from full client info to short
client info:
- CLUSTER FAILOVER
- FAILOVER / PSYNC
- REPLICAOF NO ONE
- SHUTDOWN


src/cluster_legacy.c
```diff
@@ -6518,6 +6518,10 @@ int clusterCommandSpecial(client *c) {
             addReplyErrorFormat(c, "Invalid node address specified: %s:%s", (char *)c->argv[2]->ptr,
                                 (char *)c->argv[3]->ptr);
         } else {
+            sds client = catClientInfoString(sdsempty(), c, server.hide_user_data_from_log);
```
ranshid: The client info is very large and might complicate this output, maybe we can just print the client name and address? (we can also include the lib and version if that seems important)

enjoy-binbin: yean, adding a simple version seems to be a good idea.

zuiderkwast: Will you add the short client info in this PR?

enjoy-binbin: yes, i will add it later, this indeed is a good idea.

enjoy-binbin: so, what fields should we put in the short client info? just client name and the address (and lib-name lib-version)?

zuiderkwast: Is client name sensitive info? It is redacted in some cases, so maybe skip this one.

Lib name and version, addr and laddr? (including port)

In networking.c, we just log addr and laddr in some places, like `"Error accepting a client connection: %s (addr=%s laddr=%s)"`

zuiderkwast: It was added here: a2e75a78b424a6faca4bba4b1bf8270b98407284

enjoy-binbin: i somehow think we can keep client name since these are mostly admin commands, a connection executing admin command, it is a admin user data. I am not sure about it, let's also use hide_user_data to control it.

src/networking.c
```diff
@@ -3372,6 +3372,30 @@ sds catClientInfoString(sds s, client *client, int hide_user_data) {
...
+
+    sds ret = sdscatfmt(
+        s,
+        FMTARGS(
+            "id=%U", (unsigned long long)client->id,
+            " addr=%s", getClientPeerId(client),
+            " laddr=%s", getClientSockname(client),
+            " %s", connGetInfo(client->conn, conninfo, sizeof(conninfo)),
+            " name=%s", hide_user_data ? "*redacted*" : (client->name ? (char *)client->name->ptr : ""),
+            " cmd=%s", client->lastcmd ? client->lastcmd->fullname : "NULL",
```
ranshid: maybe the cmd is not required in this specific case as we already include that data in the log msg?

zuiderkwast: Yes, if we add so many fields, we can even log the full client info?

We already have a short version as `"Error accepting a client connection: %s (addr=%s laddr=%s)"`. Is it not enough?

If we add another semi-short version, we have 3 variants of client info. Maybe it's too many variants?

enjoy-binbin: I want this short client info to be self-consistent, like it contain these fields but not the other stats fields. These fields are what i think are helpful to view, i can drop it (name / cmd / user) if you guys insist.

`"Error accepting a client connection: %s (addr=%s laddr=%s)"`. I somehow don't think it is a short client info, it is somehow like a connection info, but i guess it can be treated as client info as well.

zuiderkwast: OK, I understand. We want the fields that are useful to identify the client. This is a good definition.

I don't insist on exactly which fields we include. We can include the ones you think are useful.

ranshid: @enjoy-binbin I agree, but the cmd is not related to client identification right?

enjoy-binbin: right, i am removing it.

ranshid (approved):
LGTM - Thank you!

zuiderkwast (approved):
The PR message should also mention that some logging is changed from full client info to short client info.

PR #1250, opened 2024-11-04
RDMA: Use conn ref counter to prevent double close

pizhenwei:
RDMA: Use connection reference counter style
    
The reference counter of connection is used to protect re-entry of close method,
use this style instead the unsafe one.

### DIFF
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -1199,6 +1199,14 @@ static void connRdmaClose(connection *conn) {
         conn->fd = -1;
     }
 
+    /* If called from within a handler, schedule the close but
+     * keep the connection until the handler returns.
+     */
+    if (connHasRefs(conn)) {
+        conn->flags |= CONN_FLAG_CLOSE_SCHEDULED;
+        return;
+    }
+
     if (!cm_id) {
         return;
     }
@@ -1689,7 +1697,6 @@ static int rdmaProcessPendingData(void) {
     listNode *ln;
     rdma_connection *rdma_conn;
     connection *conn;
-    listNode *node;
     int processed;
 
     processed = listLength(pending_list);
@@ -1697,17 +1704,17 @@ static int rdmaProcessPendingData(void) {
     while ((ln = listNext(&li))) {
         rdma_conn = listNodeValue(ln);
         conn = &rdma_conn->c;
-        node = rdma_conn->pending_list_node;
 
         /* a connection can be disconnected by remote peer, CM event mark state as CONN_STATE_CLOSED, kick connection
          * read/write handler to close connection */
         if (conn->state == CONN_STATE_ERROR || conn->state == CONN_STATE_CLOSED) {
-            listDelNode(pending_list, node);
-            /* do NOT call callHandler(conn, conn->read_handler) here, conn is freed in handler! */
-            if (conn->read_handler) {
-                conn->read_handler(conn);
-            } else if (conn->write_handler) {
-                conn->write_handler(conn);
+            listDelNode(pending_list, rdma_conn->pending_list_node);
+            rdma_conn->pending_list_node = NULL;
+            /* Invoke both read_handler and write_handler, unless read_handler
+               returns 0, indicating the connection has closed, in which case
+               write_handler will be skipped. */
+            if (callHandler(conn, conn->read_handler)) {
+                callHandler(conn, conn->write_handler);
             }
 
             continue;


src/rdma.c
```diff
@@ -1689,25 +1697,21 @@ static int rdmaProcessPendingData(void) {
...
     rdma_connection *rdma_conn;
     connection *conn;
-    listNode *node;
     int processed;
 
     processed = listLength(pending_list);
     listRewind(pending_list, &li);
     while ((ln = listNext(&li))) {
         rdma_conn = listNodeValue(ln);
         conn = &rdma_conn->c;
```
xbasel, line 1706: How did this work before when  conn was closed and freed? was `rdma_conn->c` NULL ?

xbasel (commented):
-

src/rdma.c
pizhenwei: A conn will be linked into pending list on:
* a write handler is set. because there is no `POLLOUT` event for RDMA. Try to call write handler before event loop.
* RDMA disconnected CM event. This is async event from RDMA CM (not from CQ completion), use read/write handler to handle this by networking common logic.

Before this change, use func pointer directly like comment in code:
```
            /* do NOT call callHandler(conn, conn->read_handler) here, conn is freed in handler! */
            if (conn->read_handler) {
                conn->read_handler(conn);
            } else if (conn->write_handler) {
                conn->write_handler(conn);
```

xbasel (approved):

src/rdma.c
```diff
@@ -1689,25 +1697,21 @@ static int rdmaProcessPendingData(void) {
...
-            /* do NOT call callHandler(conn, conn->read_handler) here, conn is freed in handler! */
-            if (conn->read_handler) {
-                conn->read_handler(conn);
-            } else if (conn->write_handler) {
-                conn->write_handler(conn);
+            listDelNode(pending_list, rdma_conn->pending_list_node);
+            rdma_conn->pending_list_node = NULL;
+            if (callHandler(conn, conn->read_handler)) {
+                callHandler(conn, conn->write_handler);
             }
```
PingXie, line 1718: I am not familiar with the `process_pending_data` contract but I noticed a behavior change here. We are now calling both the `read_handler` and the `write_handler` while previously we invoked just one and if both are set we give the preference to `read_handler`. Is this change intentional?

pizhenwei: Yes, it's intentional. Please take a look at `callHandler`, if connection is closed successfully, this function returns 0, it would not invoke the second handler.
```
static inline int callHandler(connection *conn, ConnectionCallbackFunc handler) {
    connIncrRefs(conn);
    if (handler) handler(conn);
    connDecrRefs(conn);
    if (conn->flags & CONN_FLAG_CLOSE_SCHEDULED) {
        if (!connHasRefs(conn)) connClose(conn);
        return 0;
    } 
    return 1;
}
```
Oh, I think I need to add the necessary comment in code:
```
            /* in case read_handler is NULL, we have a chance to invoke write_handler.
               if the connection is closed successfully in read_handler, the returned 0
               will not invoke write_handler any more. */
```

zuiderkwast (approved):
This makes sense.

The PR title and description could be more clear. Let's write "reference counter" instead of just "refs".

@PingXie any more comments?

pizhenwei:
> This makes sense.
> 
> The PR title and description could be more clear. Let's write "reference counter" instead of just "refs".
> 
> @PingXie any more comments?

clang format CI error with log
```
===> Unable to contact the store, trying every minute for the next 30 minutes
===> Still unable to contact the store, trying for another 20 minutes
===> Still unable to contact the store, trying for another 10 minutes
===> Still unable to contact the store, aborting
```

It seems something wrong on apt packages.

src/rdma.c
```diff
@@ -1689,25 +1697,24 @@ static int rdmaProcessPendingData(void) {
...
-            /* do NOT call callHandler(conn, conn->read_handler) here, conn is freed in handler! */
-            if (conn->read_handler) {
-                conn->read_handler(conn);
-            } else if (conn->write_handler) {
-                conn->write_handler(conn);
+            listDelNode(pending_list, rdma_conn->pending_list_node);
+            rdma_conn->pending_list_node = NULL;
+            /* in case read_handler is NULL, we have a chance to invoke write_handler.
+               if the connection is closed successfully in read_handler, the returned 0
+               will not invoke write_handler any more. */
```
PingXie: The comment doesn't seem aligned with the logic. Whether we invoke `write_handler` is not predicated on `read_hander` being NULL or not but whether invoking `read_handler` causes the connection to be closed. How about the following?

```suggestion
            /* Invoke both read_handler and write_handler, unless read_handler 
               returns 0, indicating the connection has closed, in which case 
               write_handler will be skipped. */
```

PingXie (approved):
Thanks @pizhenwei

pizhenwei:
@zuiderkwast @PingXie @xbasel 
Thanks for your suggestions and time!
Apply Ping's suggestion and please squash it into one change!

pizhenwei:
> @zuiderkwast @PingXie @xbasel Thanks for your suggestions and time! Apply Ping's suggestion and please squash it into one change!

Because DCO failed, squash the two changes into one, and force pushed.

Issue #1251, opened 2024-11-04
[BUG] Clusters can become inconsistent if one side times out the handshake during cluster meets


madolson:
The following TCL test will reproduce the issue: 
```
start_cluster 2 0 {tags {external:skip cluster} overrides {cluster-replica-no-failover yes}} {
    test "Partial node timeout with two shards" {
        set cluster_port [find_available_port $::baseport $::portcount]
        start_server [list overrides [list cluster-enabled yes cluster-node-timeout 1 cluster-port $cluster_port]] {
            # Node 1 - Meet -> Node 2
            ## Node 1 and Node 2 will have eachother in the handshake state
            ## Node 2 will respond with a pong message. Which will put Node 2 online
            ## Node 2 should respond 'eventually' with a ping message, but we drop it.
            ## Node 2 will timeout its handshake with 1.
            R 1 DEBUG DROP-CLUSTER-PACKET-FILTER 0
            R 2 DEBUG DROP-CLUSTER-PACKET-FILTER 0

            R 1 CLUSTER MEET [srv 0 host] [srv 0 port] $cluster_port

            # Note, the delay is very tight since we want to react fast
            # to the node recieving the meet message.
            wait_for_condition 1000 1 {
                [llength [R 0 CLUSTER NODES]] eq 16
            } else {
                fail "Node 0 never learned about node 1"
            }

            # Pause the process to prevent it from getting or sending messages
            pause_process [srv -1 pid]

            # Now wait for the node 1 to give up the handshake
            wait_for_condition 50 100 {
                [llength [R 0 CLUSTER NODES]] eq 8
            } else {
                fail "Node 1 never learned about node 1"
            }

            resume_process [srv -1 pid]
            R 0 DEBUG DROP-CLUSTER-PACKET-FILTER -1
            R 1 DEBUG DROP-CLUSTER-PACKET-FILTER -1
            R 2 DEBUG DROP-CLUSTER-PACKET-FILTER -1

            after 1000
            puts "Nodes after 1 second"
            puts [R 0 CLUSTER NODES]
            puts [R 1 CLUSTER NODES]
            puts [R 2 CLUSTER NODES]

            after 30000
            puts "Nodes after 30 seconds"
            puts [R 0 CLUSTER NODES]
            puts [R 1 CLUSTER NODES]
            puts [R 2 CLUSTER NODES]
        }
    }
}
```

You'll get something like the following output:
```
Nodes after 1 second
[...]

PR #1252, opened 2024-11-04
Add a filter option to drop all cluster packets

madolson:
A minor debugging change that helped in the investigation of https://github.com/valkey-io/valkey/issues/1251. Basically there are some edge cases where we want to fully isolate a note from receiving packets, but can't suspend the process because we need it to continue sending outbound traffic. So, added a filter for that.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -2981,7 +2981,7 @@ int clusterIsValidPacket(clusterLink *link) {
         return 0;
     }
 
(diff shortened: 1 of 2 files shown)


ranshid:
can't this be done by placing an iptable rule?

enjoy-binbin (approved):

madolson:
> can't this be done by placing an iptable rule?

iptables isn't on macos, so the test isn't as portable.

PR #1253, opened 2024-11-04
Fix argument types of formatting functions

ArtSin:
`cluster_legacy.c`: `slot_info_pairs` has `uint16_t` values, but they were cast to `unsigned long` and `%i` was used.

`valkey-cli.c`: `node->replicas_count` is `int`, not `unsigned long`.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -5544,12 +5544,12 @@ sds representClusterNodeFlags(sds ci, uint16_t flags) {
  * else each slot is added separately. */
 sds representSlotInfo(sds ci, uint16_t *slot_info_pairs, int slot_info_pairs_count) {
     for (int i = 0; i < slot_info_pairs_count; i += 2) {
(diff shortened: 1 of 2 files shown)


deps/lua/src/lua_cjson.c
```diff
@@ -1131,7 +1131,7 @@ static void json_decode_descend(lua_State *l, json_parse_t *json, int slots)
 
     strbuf_free(json->tmp);
     luaL_error(l, "Found too many nested data structures (%d) at character %d",
-        json->current_depth, json->ptr - json->data);
+        json->current_depth, (int)(json->ptr - json->data));
```
madolson: I would rather leave lua alone if we can.

madolson: yeah, I know, but it's a vendored dependency and we would prefer it stay as it is upstream.

madolson (approved):

PR #1254, opened 2024-11-04
Make FUNCTION RESTORE FLUSH flush async based on lazyfree-lazy-user-flush

enjoy-binbin:
FUNCTION RESTORE have a FLUSH option, it will delete all the existing
libraries before restoring the payload. If for some reasons, there are
a lot of libraries, we will block a while in here.

Adding a new functionsLibCtxFreeGeneric, it accepts a async parameter
to determine how to free the function ctx.


ranshid (approved):

src/replication.c
```diff
@@ -2268,7 +2268,7 @@ void readSyncBulkPayload(connection *conn) {
                                       NULL);
 
                 disklessLoadDiscardTempDb(diskless_load_tempDb);
-                functionsLibCtxFree(temp_functions_lib_ctx);
+                functionsLibCtxFree(temp_functions_lib_ctx, 0);
```
enjoy-binbin: i will handle this diskless thing in other PR, we are doing async in temp db, and here with the temp function ctx, we should use async too

```
/* Discard tempDb, this can be slow (similar to FLUSHALL), but it's always async. */
void discardTempDb(serverDb *tempDb, void(callback)(dict *)) {
    int async = 1;

    /* Release temp DBs. */
    emptyDbStructure(tempDb, -1, async, callback);
    for (int i = 0; i < server.dbnum; i++) {
        kvstoreRelease(tempDb[i].keys);
        kvstoreRelease(tempDb[i].expires);
    }

    zfree(tempDb);
}
```

enjoy-binbin:
This doesn't look like a major change, i am going to merge it. @zuiderkwast do you want to take a quick look before i merge it?

src/functions.c
```diff
@@ -189,18 +189,22 @@ void functionsLibCtxClearCurrent(int async) {
...
+void functionsLibCtxFree(functionsLibCtx *functions_lib_ctx, int async) {
+    if (async) {
+        freeFunctionsAsync(functions_lib_ctx);
+    } else {
+        functionsLibCtxClear(functions_lib_ctx);
+        dictRelease(functions_lib_ctx->functions);
+        dictRelease(functions_lib_ctx->libraries);
+        dictRelease(functions_lib_ctx->engines_stats);
+        zfree(functions_lib_ctx);
+    }
```
zuiderkwast: freeFunctionsAsync calls functionsLibCtxFree again, so we get a circular call graph.

```
    functionsLibCtxFree --> freeFunctionsAsync
           ^                        /
            \                      /
             `--------------------'
```

Even if freeFunctionsAsync always calls it with async=0, it's not a perfect design IMO. I think it's better something like this:

```
            functionsLibCtxFreeSyncOrAsync
                   /           \
                  /             \
                 v               v
    functionsLibCtxFree <-- freeFunctionsAsync
```

Here, functionsLibCtxFree is only sync, as before. If we do this, then functionsLibCtxFreeSyncOrAsync can be static and internal in functions.c.

zuiderkwast (commented):
Good idea! Just a comment about the structure of functions.

src/functions.c
enjoy-binbin: yean, that is a bad call, i added a new freeFunctionsSync just like scriptingRelease, does it better?

zuiderkwast: It's better but I was also suggesting that we don't need both of them in functions.h. All the calls from files outside functions.c only need to call sync function.

Also, to minimize the diff, I suggested to keep functionsLibCtxFree as the sync function, so existing calls to this one can remain identical. We can add a new one, which is internal in functions.c, that has the async parameter.

Do you think it's better that the sync function has "sync" in the name?

src/lazyfree.c
```diff
@@ -64,7 +64,7 @@ void lazyFreeLuaScripts(void *args[]) {
 void lazyFreeFunctionsCtx(void *args[]) {
     functionsLibCtx *functions_lib_ctx = args[0];
     size_t len = functionsLibCtxFunctionsLen(functions_lib_ctx);
-    functionsLibCtxFree(functions_lib_ctx);
+    functionsLibCtxFree(functions_lib_ctx, 0);
```
zuiderkwast:
```suggestion
    freeFunctionsSync(functions_lib_ctx);
```

Btw, all function from function.h should start with "function"? It is the style so far.

If we keep this as `functionsLibCtxFree(functions_lib_ctx)` = sync, we can add a helper in functions.c, maybe call it `functionsLibCtxFreeGeneric(functions_lib_ctx, int async)`? or another name. Anyway, I think this structure has benefit that we have fewer public functions in functions.h.

src/functions.c
enjoy-binbin: sorry, it was so long ago that i forgot the origin diff and only looked at the most recent one. 

> Do you think it's better that the sync function has "sync" in the name?

i don't mind actually... scriptingRelease has the same style (this is actully written by me...)

```
void scriptingRelease(int async) {
    if (async)
        freeLuaScriptsAsync(lctx.lua_scripts, lctx.lua_scripts_lru_list, lctx.lua);
    else
        freeLuaScriptsSync(lctx.lua_scripts, lctx.lua_scripts_lru_list, lctx.lua);
}
```

src/functions.c
```diff
@@ -186,18 +186,28 @@ void functionsLibCtxClearCurrent(int async) {
...
     dictRelease(functions_lib_ctx->libraries);
     dictRelease(functions_lib_ctx->engines_stats);
     zfree(functions_lib_ctx);
 }
 
+/* Free the given functions ctx */
+void functionsLibCtxFree(functionsLibCtx *functions_lib_ctx) {
+    functionsLibCtxFreeGeneric(functions_lib_ctx, 0);
+}
+
```
zuiderkwast: There is still some circular call graph here, if we let functionsLibCtxFree call functionsLibCtxFreeGeneric. I suggest this:

```C
/* Free the given functions ctx */
static void functionsLibCtxFreeGeneric(functionsLibCtx *functions_lib_ctx, int async) {
    if (async) {
        freeFunctionsAsync(functions_lib_ctx);
    } else {
        functionsLibCtxFree(functions_lib_ctx);
    }
}

/* Free the given functions ctx */
void functionsLibCtxFree(functionsLibCtx *functions_lib_ctx) {
    functionsLibCtxClear(functions_lib_ctx);
    dictRelease(functions_lib_ctx->functions);
    dictRelease(functions_lib_ctx->libraries);
    dictRelease(functions_lib_ctx->engines_stats);
    zfree(functions_lib_ctx);
}
```

enjoy-binbin: yes, thanks! it is better, i cant circular back my brain, it somehow stuck...

zuiderkwast (approved):
Yes, this is better structure. I hope you agree. :smile:

src/functions.c
zuiderkwast: I think you're too fast. :fast_forward: Easy to make mistake when you're very fast.

PR #1255, opened 2024-11-04
Fix incorrect cache_memory reset in functionsLibCtxClear

enjoy-binbin:
functionsLibCtxClear should clear the provided lib_ctx parameter,
not the static variable curr_functions_lib_ctx, as this contradicts
the function's intended purpose.

The impact i guess is minor, like in some unhappy paths (diskless load
fails, function restore fails?), we will mess up the functions_caches
field, which is used in used_memory_functions / used_memory_scripts
fileds in INFO.

### DIFF
--- a/src/functions.c
+++ b/src/functions.c
@@ -175,7 +175,7 @@ void functionsLibCtxClear(functionsLibCtx *lib_ctx) {
         stats->n_lib = 0;
     }
     dictReleaseIterator(iter);
-    curr_functions_lib_ctx->cache_memory = 0;
+    lib_ctx->cache_memory = 0;
 }
 
 void functionsLibCtxClearCurrent(int async) {


ranshid (approved):

src/functions.c
```diff
@@ -175,7 +175,7 @@ void functionsLibCtxClear(functionsLibCtx *lib_ctx) {
         stats->n_lib = 0;
     }
     dictReleaseIterator(iter);
-    curr_functions_lib_ctx->cache_memory = 0;
+    lib_ctx->cache_memory = 0;
```
xbasel, line 178: The PR description states:

```
We call functionsLibCtxClear in several scenarios, and we don't
necessarily want to reset curr_functions_lib_ctx.
```
It appears to me that the previous code contained a potential bug rather than un-optimized code. The function functionsLibCtxClear was clearing the cache_memory of curr_functions_lib_ctx instead of lib_ctx. Is my understanding correct?

enjoy-binbin: yes, that is the case, your understanding is right

xbasel: I'd suggest changing the commit message to something like:
```
functionsLibCtxClear should clear the provided lib_ctx parameter, not the static variable curr_functions_lib_ctx, as this contradicts the function's intended purpose.
```

What is the impact of the current implementation?

enjoy-binbin: great, i am taking your message, thanks. The impact i guess is minor, like in some unhappy paths (diskless fails, function restore fails?), we will mess up the functions_caches field, which is used in used_memory_functions / used_memory_scripts fileds in INFO.

zuiderkwast: We can describe this impact in the commit message, PR top comment.

zuiderkwast (approved):

src/functions.c
enjoy-binbin: added, please take a look

zuiderkwast: Looks good 👍

PR #1258, opened 2024-11-04
Change errno from EEXIST to EALREADY in serverFork if child process exists

enjoy-binbin:
We set this to EEXIST in 568c2e039bac388003068cd8debb2f93619dd462,
it prints "File exists" which is not quite accurate,
change it to EALREADY, it will print "Operation already in progress".

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -6396,7 +6396,7 @@ void closeChildUnusedResourceAfterFork(void) {
 int serverFork(int purpose) {
     if (isMutuallyExclusiveChildType(purpose)) {
         if (hasActiveChildProcess()) {
-            errno = EEXIST;
+            errno = EALREADY;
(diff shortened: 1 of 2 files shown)


xbasel (approved):

zuiderkwast (commented):
`ValkeyModule_Fork` returns -1 on failure. It also sets `errno`, but it's not documented.

If a module checks `errno` after a call to `ValkeyModule_Fork`, it is now changed. But I think it's OK, since it's not documented.

enjoy-binbin:
yean, should we document it?

zuiderkwast (approved):
> yean, should we document it?

No, if we document it, changes like this in the future can be breaking changes. Better leave it undocumented. :)

PR #1259, opened 2024-11-04
Remove empty DB check branch in KEYS command

enjoy-binbin:
We don't think we really care about optimizing for the empty DB case,
which should be uncommon. Adding branches hurts branch prediction.

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -837,6 +837,11 @@ void keysCommand(client *c) {
         }
         kvs_di = kvstoreGetDictSafeIterator(c->db->keys, pslot);
     } else {
+        if (!kvstoreSize(c->db->keys)) {
+            /* Requested db is empty. */
+            setDeferredArrayLen(c, replylen, 0);
+            return;
+        }
         kvs_it = kvstoreIteratorInit(c->db->keys);
     }
     robj keyobj;
--- a/src/db.c
+++ b/src/db.c
@@ -830,18 +830,8 @@ void keysCommand(client *c) {
     kvstoreDictIterator *kvs_di = NULL;
     kvstoreIterator *kvs_it = NULL;
     if (pslot != -1) {
-        if (!kvstoreDictSize(c->db->keys, pslot)) {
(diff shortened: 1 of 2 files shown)


src/db.c
```diff
@@ -837,6 +837,11 @@ void keysCommand(client *c) {
         }
         kvs_di = kvstoreGetDictSafeIterator(c->db->keys, pslot);
     } else {
+        if (!kvstoreSize(c->db->keys)) {
```
enjoy-binbin: it is unimportant, just an OCD

src/db.c
```diff
@@ -837,6 +837,11 @@ void keysCommand(client *c) {
...
+            /* Requested db is empty. */
+            setDeferredArrayLen(c, replylen, 0);
```
xbasel: Would it make sense to refactor the code like this for clarity ?

```
if ((pslot != -1 && !kvstoreDictSize(c->db->keys, pslot)) || 
    (pslot == -1 && !kvstoreSize(c->db->keys))) {
    // Either the requested slot or db is empty.
    setDeferredArrayLen(c, replylen, 0);
    return;
}

if (pslot != -1) {
    kvs_di = kvstoreGetDictSafeIterator(c->db->keys, pslot);
} else {
    kvs_it = kvstoreIteratorInit(c->db->keys);
}
```

madolson: Honestly I would rather refactor this to just:
```
if (pslot != -1) {
    kvs_di = kvstoreGetDictSafeIterator(c->db->keys, pslot);
} else {
    kvs_it = kvstoreIteratorInit(c->db->keys);
}
```

and just let it fall through the code. I don't think we rally care about optimizing for the empty DB case, which should be uncommon. Adding branches hurts branch prediction.

enjoy-binbin: ok, good point about the branch prediction, i am not sure about the impact, but let's do it, i removed the branches.

xbasel (approved):

madolson (approved):

PR #1260, opened 2024-11-04
Integrating fast_float to optionally replace strtod

### LINKED ISSUE #1069: Investigate integrating with fast_float

parthpatel:
Fast_float is a C++ header-only library to parse doubles using SIMD instructions. The purpose is to speed up sorted sets and other commands that use doubles. A single-file copy of fast_float is included in this repo. This introduces an optional dependency on a C++ compiler.

The use of fast_float is enabled at compile time using the make variable `USE_FAST_FLOAT=yes`. It is disabled by default.

Fixes #1069.


madolson:
I suppose I don't really understand the benefit of a submodule vs inlining. Since we aren't tightly controlling the versioning between the main repo and releases, it becomes much harder to know if there is a security issue impacting a specific release. At the very least we should be pinning a specific version of fast_float that we are pulling.

parthpatel:
> Since we aren't tightly controlling the versioning between the main repo and releases, it becomes much harder to know if there is a security issue impacting a specific release. At the very least we should be pinning a specific version of fast_float that we are pulling.

We track fast_float commit id in the valkey repository with this change - see the copy pasted change from my commit below. For every valkey commit, we can look up exact version of fast_float at all times using this method. Pulling new fast_float version is as simple as git pull on the submodule. 

---
 Submodule [fast_float](https://github.com/fastfloat/fast_float) added at [e800ca](https://github.com/fastfloat/fast_float/tree/e800cabe48831775b401816cf1f72fdc432f40ae)
----

madolson:
Does it get pulled automatically as part of the release into the release artifacts?

parthpatel:
> Does it get pulled automatically as part of the release into the release artifacts?

There is a "git submodule update --init" command in Makefile to initialize it automatically. So yes, It will automatically checkout the same commit every time during build.

madolson:
> There is a "git submodule update --init" command in Makefile to initialize it automatically. So yes, It will automatically checkout the same commit every time during build.

So we are adding a new dependency to the release process, since you need to be able to fetch the code from github. I think we should consider figuring out how to pull the code in when we do a release so that folks don't need to do `git submodule update --init`.

parthpatel:
> > There is a "git submodule update --init" command in Makefile to initialize it automatically. So yes, It will automatically checkout the same commit every time during build.
> 
> So we are adding a new dependency to the release process, since you need to be able to fetch the code from github. I think we should consider figuring out how to pull the code in when we do a release so that folks don't need to do `git submodule update --init`.

The other approach would be to just check-in the whole git repo as a folder under valkey, which should work. What is the issue with git dependency in github release workflows? I can put a post-checkout hook that always initializes modules on checkout, as long as checkout happens outside of the release process.

parthpatel:
I am dropping the submodule idea for now as it requires a larger discussion about release process. I don't have access to @swaingotnochill's repository or PR. Therefore, I pulled his commit into this CR to maintain his author-ship on the code he wrote. I integrated it with Redis and fixed the Makefile. It should build now and will be ready to push.

I will work on benchmarking this separately. I can also turn off fast_float by default if folks have concerns and test it on my branch.

swaingotnochill (commented):
Feel free to modify. I am on a vacation so it will be difficult for me to work until I am back. Cheers

parthpatel:
@zuiderkwast  or @madolson any pointers on how to solve almalinux issue with missing g++? 

```
cd fast_float && make
make[3]: Entering directory '/__w/valkey/valkey/deps/fast_float'
g++ -std=c++11 -O3 -fPIC -c fast_float_strtod.cpp -o fast_float_strtod.o
make[3]: g++: Command not found
make[3]: *** [Makefile:9: fast_float_strtod.o] Error 127
make[3]: Leaving directory '/__w/valkey/valkey/deps/fast_float'
make[2]: *** [Makefile:80: fast_float] Error 2
make[2]: *** Waiting for unfinished jobs....
```

zuiderkwast:
I'm skeptical to submodules too. Offline builds get complicated. Source releases get complicated. So far, we've used either vendored dependencies or system installed ones (like OpenSSL).

I prefer that we vendor this one. We could copy only one or a few files, as we've done with crc64 and some other small libraries?

parthpatel:
All checks are passing now including the 32-bit builds. This is ready for review. @madolson or @zuiderkwast. 

I copied the entire fast_float include directory into the deps/fast_float directory to include all headers as they are. In future, we can simply copy again to upgrade or introduce bugfixes.

src/resp_parser.c
```diff
@@ -150,13 +151,11 @@ static int parseDouble(ReplyParser *parser, void *p_ctx) {
     parser->curr_location = p + 2; /* for \r\n */
     char buf[MAX_LONG_DOUBLE_CHARS + 1];
     size_t len = p - proto - 1;
-    double d;
+    double d = 0;
```
madolson, line 155: We also want to make the change for https://github.com/valkey-io/valkey/blob/unstable/src/util.c#L595. The current code is just for lua parsing.

parthpatel: > We also want to make the change for https://github.com/valkey-io/valkey/blob/unstable/src/util.c#L595. 

Let me create a separate PR for it. I saw that strtod invocation.

madolson: That is like the main reason to use it, I think it should be in this PR.

deps/fast_float/fast_float_strtod.cpp
```diff
@@ -0,0 +1,38 @@
+#include "fast_float/fast_float.h"
```
madolson: This file needs copyright information.

src/Makefile
```diff
@@ -383,6 +383,21 @@ else
 	MAYBE_UNINSTALL_REDIS_SYMLINK=
 endif
 
+# Determine whether to use fast_float as a library or not. 
+BUILD_FAST_FLOAT:=yes
+ifeq ($(BUILD_FAST_FLOAT),yes)
+	DEPENDENCY_TARGETS+= fast_float
+        FINAL_LIBS += ../deps/fast_float/fast_float_strtod.o
+	# Mac uses Clang and the standard c++ library is named libc++. For others, it is named libstdc++.
```
madolson:
```suggestion
	# Prefer to use libc++ when compiling with Clang. 
```
MacOS does use clang, but clang is also available on Linux.

src/Makefile
```diff
@@ -383,6 +383,21 @@ else
 	MAYBE_UNINSTALL_REDIS_SYMLINK=
 endif
 
+# Determine whether to use fast_float as a library or not. 
+BUILD_FAST_FLOAT:=yes
```
madolson: You can't disable fast_float in the current implementation, since it will fail to build. I think we should probably provide an alternative implementation for fast_float when it's unavailable to maximize compatibility. Either that, or we just require all this stuff.

You probably should also do something like add a function like `valkey_strtod` in util.c that can be called more widely that determines which one to build at compile time.

deps/fast_float/Makefile
```diff
@@ -0,0 +1,16 @@
+# Author: Roshan Swain <swainroshan001@gmail.com>
+
```
madolson:
```suggestion
```
His contribution will still be included in the commit history.

deps/fast_float/README.md
```diff
@@ -0,0 +1,30 @@
...
+## INTEGRATION
+
+1. `fast_float_strtod` is a wrapper around the `from_chars` function.
+
+## RESOURCES
+
+1. fast_float: https://github.com/fastfloat/fast_float
+
+
+...REMOVE THIS LATER...
```
madolson: Should we remove this now?

src/Makefile
```diff
@@ -477,7 +492,7 @@ endif
 
 # valkey-server
 $(SERVER_NAME): $(ENGINE_SERVER_OBJ)
-	$(SERVER_LD) -o $@ $^ ../deps/hiredis/libhiredis.a ../deps/lua/src/liblua.a ../deps/hdr_histogram/libhdrhistogram.a ../deps/fpconv/libfpconv.a $(FINAL_LIBS)
+	$(SERVER_LD) -v -o $@ $^ ../deps/hiredis/libhiredis.a ../deps/lua/src/liblua.a ../deps/hdr_histogram/libhdrhistogram.a ../deps/fpconv/libfpconv.a $(FINAL_LIBS)
```
madolson:
```suggestion
	$(SERVER_LD) -o $@ $^ ../deps/hiredis/libhiredis.a ../deps/lua/src/liblua.a ../deps/hdr_histogram/libhdrhistogram.a ../deps/fpconv/libfpconv.a $(FINAL_LIBS)
```
What is the -v for?

parthpatel: I enabled verbosity to debug compilation issues forgot to remove it.

src/Makefile
zuiderkwast: And Linux is also available on Mac.

Mac != MacOS.

src/Makefile
zuiderkwast: If we vendor it, it's always available? Or is there some platform where it isn't?

deps/fast_float/fast_float_strtod.cpp
zuiderkwast: Is this file from the `fast_float` project or is it our glue?

If it's our own glue code, I think we should put it under `src/` rather than under `deps/`.

The deps files should be the files from the dependency itself. I hope we don't have to modify them all to insert a copyright notice. If they don't have one, then I hope it's enough to include their LICENSE file.

deps/fast_float/README.md
zuiderkwast: I think we shall move this text to deps/README.md and skip the two-level directory `deps/fast_float/fast_float/`. The directory `deps/fast_float/` should contain only files from the fast_float project. Our glue code can live under src/ IMO.

deps/fast_float/fast_float/ascii_number.h
zuiderkwast, line 1: These files seems to be from fast_float's repo `include/fast_float/`.

Can we keep the original directory structure from the fast_float repo? We don't have to copy all files, but the files we copy can be in the exact same path to avoid confusion.

In `deps/README.md`, we should describe exactly which files we copied and how to update them when there's a new fast_float version.

deps/fast_float/fast_float_strtod.cpp
madolson: This one looks not from the original repo.

src/Makefile
madolson: I think the concern is about if C++ is available. I assume it's everywhere, but people have complained about this type of incompatibility in the past.

zuiderkwast: OK, then it's safer to allow building without fast_float and then `valkey_strtod` can provide a callback using the old implementation.

Regarding the make variable prefix, some use `BUILD` (`BUILD_TLS`) and some use `USE` (`USE_SYSTEMD`, `USE_JEMALLOC`, ...). It seems that `USE` is more common actually.

deps/fast_float/README.md
parthpatel: I cleaned up the README files. 

> Our glue code can live under src/ IMO.

Not without using CXX to compile it though.

src/Makefile
parthpatel: I made fast_float optional and off by default.

src/Makefile
parthpatel: Fixed the comment, this check is intended to support Clang not just Mac.

deps/fast_float/fast_float_strtod.cpp
parthpatel: Added license to glue to code. The original files don't have license embedded, so we will have to manually add it every time. These kind of issues are one of the reason why I think submodule is superior approach and we should just add git as dependency to build workflows.

src/resp_parser.c
parthpatel: Converted all invocations.

deps/fast_float/fast_float.h
```diff
@@ -1,3869 +0,0 @@
...
-// with contributions from Neal Richardson
-// with contributions from Tim Paine
-// with contributions from Fabio Pellacini
-// with contributions from Lénárd Szolnoki
-// with contributions from Jan Pharago
-// with contributions from Maya Warrier
-// with contributions from Taha Khokhar
-//
-//
-// MIT License Notice
```
zuiderkwast: Actually, I quite like this amalgamated version. It contains the license information too, which is important.

I believe they created this script just for projects like us, so that we can include only one file in our project.

deps/fast_float/fast_float/ascii_number.h
parthpatel: no preference either way...moved it under include directory.

deps/fast_float/fast_float.h
parthpatel: > I believe they created this script just for projects like us, so that we can include only one file in our project.

It makes it very difficult to compare or check for changes in future. I reverted it to single file though. If we are not going submodule route, I don't have preferences between single vs multi file :)

zuiderkwast: Every solution has pros and cons. :)

zuiderkwast:
This is the structure I would like:

* deps/fast_float/fast_float.h: The amalgamated fast float code, created using the script in the fast float repo. We shouldn't put anything else under deps/fast_float/ IMO.

* deps/README.md: How to update the fast float dependency. Looks good already.

* src/fast_float_strtod.cpp: Our glue code.

* src/unit/test_fast_float.c: Unit test.

* src/Makefile: Conditional compilation of src/fast_float_strtod.cpp. Only compile it if USE_FAST_FLOAT=yes.

* src/util.c: valkey_strtod() or maybe can we let string2d() be our wrapper function and use it everywhere instead of strtod()?

I think we can enable fast float by default. It's vendored and statically linked, the code is always there and it needs no runtime dependency. If the c++ compiler is missing, we can print a useful error message from the Makefile: "c++ not found. Either install c++ or compile with USE_FAST_FLOAT=no."

deps/fast_float/README.md
zuiderkwast: I mean that we only compile it if USE_FAST_FLOAT is enabled.

deps/fast_float/Makefile
```diff
@@ -0,0 +1,49 @@
...
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
+
+
+CXX?=g++
```
zuiderkwast: I think `g++` is GCC and `clang++` is Clang, and there are others.

Is there no generic binary name for a c++ compiler? For C, the standard compiler binary name is `cc`.

zuiderkwast:
> * src/Makefile: Conditional compilation of src/fast_float_strtod.cpp. Only compile it if USE_FAST_FLOAT=yes.

Does this make sense?

I can help with Makefile if you need.

madolson:
We will do it with opt-in, and then at a later point decide if we want to make it opt-out. At least for now we are okay with taking the C++ dependency. @soloestoy @PingXie Please comment if you don't want to take the C++ dependency.

parthpatel:
> This is the structure I would like:
> 
> * deps/fast_float/fast_float.h: The amalgamated fast float code, created using the script in the fast float repo. We shouldn't put anything else under deps/fast_float/ IMO.
> * deps/README.md: How to update the fast float dependency. Looks good already.
> * src/fast_float_strtod.cpp: Our glue code.
> * src/unit/test_fast_float.c: Unit test.
> * src/Makefile: Conditional compilation of src/fast_float_strtod.cpp. Only compile it if USE_FAST_FLOAT=yes.
> * src/util.c: valkey_strtod() or maybe can we let string2d() be our wrapper function and use it everywhere instead of strtod()?
> 
> I think we can enable fast float by default. It's vendored and statically linked, the code is always there and it needs no runtime dependency. If the c++ compiler is missing, we can print a useful error message from the Makefile: "c++ not found. Either install c++ or compile with USE_FAST_FLOAT=no."

* None of these choices have impact on performance, so I have no preference. If everyone agrees to this, then I will switch to it. @madolson @PingXie @soloestoy 
* I did think about making the switch between fast_float and strtod in fast_float_strtod function, but then it adds an additional function call, so did not go that route. I don't think compiler can optimize it away if it is included as separate library. we can't compile it using C compiler along with Redis binary so if we are ok with one more function call, I can make it a wrapper. I will hold off on further changes until everyone agrees.
* Same with OFF by default or ON by default. Does everyone agree to ON by default?

deps/fast_float/Makefile
parthpatel: This line only sets CXX variable if it is not already set. It will be set all the times on the known platforms.

> Is there no generic binary name for a c++ compiler? 

gcc sets c++ binary under /usr/bin, but clang install on RHEL did not. I don't think there is a standard. We should just rely on Makefile to set this variable correctly, rather than rely on this line.

Edit: Actually even with g++, Mac compilation is using Clang, so not sure what is happening there. Investigating.

zuiderkwast:
> * I did think about making the switch between fast_float and strtod in fast_float_strtod function, but then it adds an additional function call, so did not go that route. I don't think compiler can optimize it away if it is included as separate library. we can't compile it using C compiler along with Redis binary so if we are ok with one more function call, I can make it a wrapper. I will hold off on further changes until everyone agrees.

I wouldn't want an extra function call if it affects performance. Performance is the point of this PR after all. :)

This is statically linked and we use link-time-optimization by default (-flto and -O3), which can inline across compilation units, so I guess it can be inlined? Another option is to make a `static inline` function in util.h that contains the ifdefs. This is always inlined and we can isolate the ifdefs to one place. Or I'd even be fine with the ifdefs if the other alternatives have some problem.

> * Same with OFF by default or ON by default. Does everyone agree to ON by default?

We discussed this in a core team meeting today and made a decision to use OFF by default, so that's already decided by a majority of the core team. Off by default is because we'll release it in a minor versions. (Later, we might decide change the default when we bumb the major version, like Valkey 9.0.)

deps/fast_float/Makefile
parthpatel: Apple just symlinks g++ and c++ to clang in their binaries. Same is happening on RHEL5, so I am just going to move this to c++, and let it error if the symlink is not present.

parthpatel:
> This is statically linked and we use link-time-optimization by default (-flto and -O3), which can inline across compilation units, so I guess it can be inlined? Another option is to make a static inline function in util.h that contains the ifdefs. This is always inlined and we can isolate the ifdefs to one place. Or I'd even be fine with the ifdefs if the other alternatives have some problem.

Evaluating this further, I can't make fast_float_strtod a wrapper coz then it will require C++ compilation always. if we create another wrapper on top of fast_float_strtod, that's just too many wrappers. I recommend making fast_float_strtod the default wrapper once USE_FAST_FLOAT is defaulted to yes and c++ is required for compilation always. I am keeping existing code and enabling -flto, if compiler sees it fit then it will optimize the code to be inline. 


> src/fast_float_strtod.cpp: Our glue code.

On the topic of moving the fast_float_strtod.cpp to src/, the src/Makefile has become a monolith. From my experience so far with this Makefile, I am now strongly against this. We should move towards smaller build units. 

If the intention is to separate fast_float library from integration code, then we can create two separate folders under deps - one for library and one for integration with its Makefiles. We can also move integration code folder outside of deps, if it does not fit there. But let's not merge it into src/Makefile. 

> src/util.c: valkey_strtod() or maybe can we let string2d() be our wrapper function and use it everywhere instead of strtod()?

that said, this can still happen if we can inline string2d. I am still inclined to just keep the code as it is without relying too much on compile inlining things.

deps/fast_float/Makefile
```diff
@@ -0,0 +1,47 @@
...
+# AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
+# IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+# ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+# LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+# CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+# SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+# INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+# CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+# ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+# POSSIBILITY OF SUCH DAMAGE.
```
madolson: You only need the copyright information in the actual code files, so you can drop it here and in the header.

src/util.c
```diff
@@ -51,6 +51,10 @@
 #include "sha256.h"
 #include "config.h"
 
+#ifdef USE_FAST_FLOAT
+#include "../deps/fast_float/fast_float_strtod.h"
+#endif
```
madolson:
```suggestion
#ifdef USE_FAST_FLOAT
#include "../deps/fast_float/fast_float_strtod.h"
static inline double valkey_strtod(const char *str, char **endptr) {
    errno = 0;
    return fast_float_strtod(str, endptr);
}
#else
static inline double valkey_strtod(const char *str, char **endptr) {
    return strtod(str, endptr);
}
#endif
```
This cuts down the usage of ifdefs to a single location in this file. This can even be just in the header if we really want.

deps/fast_float/fast_float_strtod.cpp
```diff
@@ -0,0 +1,46 @@
...
+    const char* fast_float_strtod(const char *str, double *value)
+    {
+        double temp = 0; 
+        auto answer = fast_float::from_chars(str, str + strlen(str), temp);
+        *value = temp; 
+        if (answer.ec != std::errc()) { 
+            errno = (answer.ec == std::errc::result_out_of_range) ? ERANGE : EINVAL;
+        }
+        return answer.ptr;
+    }
```
madolson:
```suggestion
    double fast_float_strtod(const char *str, const char *endptr)
    {
        double temp = 0; 
        auto answer = fast_float::from_chars(str, str + strlen(str), temp);
        *endptr = answer.ptr; 
        if (answer.ec != std::errc()) { 
            errno = (answer.ec == std::errc::result_out_of_range) ? ERANGE : EINVAL;
        }
        return temp;
    }
```
Is there a reason we aren't emulating strtod here? It seems odd we are just returning the parameters slightly differently.

deps/fast_float/test_fast_float.c
```diff
@@ -0,0 +1,61 @@
...
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+#include "fast_float_strtod.h"
```
madolson: This test should go in src/unit, where the other unit tests are going.

madolson:
> I wouldn't want an extra function call if it affects performance. Performance is the point of this PR after all. :)

You won't be able to see the performance of a single branchless jump instruction since it will be prefetched into the instruction cache.

deps/fast_float/Makefile
```diff
@@ -0,0 +1,47 @@
...
+
+.PHONY: all clean test
+
+# We don't need explicit target for fast_float_strtod.o as Makefile will use 
+# CXXFLAGS and CXX variables to auto-generate it for us. 
+all: fast_float_strtod.o
+
+clean:
+	rm -f *.o || true;
+	rm test || true;
```
madolson: Why do we always succeed here?

deps/fast_float/Makefile
```diff
@@ -0,0 +1,47 @@
...
+
+CXX?=c++
+# We want -O3 and -flto, so that compiler can inline the function if possible. 
+CXXFLAGS+=-std=c++11 -O3 -fPIC -flto -D FASTFLOAT_ALLOWS_LEADING_PLUS
+
+.PHONY: all clean test
+
+# We don't need explicit target for fast_float_strtod.o as Makefile will use 
+# CXXFLAGS and CXX variables to auto-generate it for us. 
+all: fast_float_strtod.o
```
madolson: I think we want to build an archive here instead of pulling the C++ toolchain into the main makefile. You can see deps/hdr_histogram as an example. That will produce a bunch of symbols that we can that compile against. As we talked about, linker time optimizations still can happen with an archive so it won't affect performance but it'll keep the C++ code more self-contained here.

deps/fast_float/fast_float_strtod.h
```diff
@@ -0,0 +1,64 @@
...
+    *              0.0 and stores error code in errno to ERANGE or EINVAL.
+    *
+    * @return On success, returns char pointer pointing to '\0' at the end of the string.
+    *         On failure, returns char pointer pointing to first invalid character in the string.
+    *
+    * @note This function uses the fast_float library (https://github.com/fastfloat/fast_float)
+    *       for the actual conversion, which can be significantly faster than standard library functions.
+    *
+    * @see https://github.com/fastfloat/fast_float for more information on the underlying library.
+    */
```
madolson:
```suggestion
   /**
    * Converts a null-terminated byte string to a double using the fast_float library.
    *
    * This function provides a C-compatible wrapper around the fast_float library's string-to-double
    * conversion functionality. It aims to offer a faster alternative to the standard strtod function.
    *
    * nptr A pointer to the null-terminated byte string to be converted.
    * value A pointer to the double variable where the function stores converted double value.
    *              On success, the function stores the converted double value. On failure, it stores
    *              0.0 and stores error code in errno to ERANGE or EINVAL.
    *
    * On success, returns char pointer pointing to '\0' at the end of the string.
    *         On failure, returns char pointer pointing to first invalid character in the string.
    *
    * This function uses the fast_float library (https://github.com/fastfloat/fast_float)
    *       for the actual conversion, which can be significantly faster than standard library functions.
    */
```
We don't use doxygen, so I think the tags don't make a lot of sense.

parthpatel: It still makes the documentation readable. I generally prefer it.

deps/fast_float/Makefile
parthpatel: Check the new structure. fast_float integration is now moved under src/fast_float. Main makefile remains unpolluted except it will need libc++ to compile against even for gcc, because the c++ compiled library uses it.

deps/fast_float/Makefile
parthpatel: No point failing clean especially if the build had failed previously and "test" file was never generated.

deps/fast_float/test_fast_float.c
parthpatel: src/unit is broken - https://github.com/valkey-io/valkey/issues/1290

deps/fast_float/fast_float_strtod.cpp
parthpatel: There is one less if/else, as endptr can be null. This approach removes that if branch. It also removes the requirement to pass NULL since endptr is optional parameter.

src/util.c
parthpatel: Incorporated this into the new structure. Changed the interface though.

parthpatel:
This build should succeed now. Here is a summary of changes from today: 
1. Moved fast_float integration code to src/fast_float. Unit test still resides in this same directory because src/unit build is broken. 
2. Created a wrapper named valkey_strtod in fast_float_strtod.h header file. It is "static inline" function. This removes ifdefs from individual code files. 
3. USE_FAST_FLOAT is off by default and now disables entire fast float compilation. 
4. Switching back to using single fast_float.h header file. deps/fast_float only contains 1 file now. README is merged into deps/README.md

I verified that this compiles for Mac and Rhel for both USE_FAST_FLOAT=yes and USE_FAST_FLOAT=no. Used "make test" to also include tests.

deps/fast_float/fast_float_strtod.h
zuiderkwast: Don't introduce new style in this 15 years old code base please. Consistency is more important.

src/fast_float/Makefile
```diff
@@ -0,0 +1,49 @@
...
+#
+# Redistribution and use in source and binary forms, with or without
+# modification, are permitted provided that the following conditions are met:
+#
+#   * Redistributions of source code must retain the above copyright notice,
+#     this list of conditions and the following disclaimer.
+#   * Redistributions in binary form must reproduce the above copyright
+#     notice, this list of conditions and the following disclaimer in the
+#     documentation and/or other materials provided with the distribution.
+#   * Neither the name of Redis nor the names of its contributors may be used
```
zuiderkwast: This is not Redis. For new files, we don't use this license text with "Redis" in the 3rd clause.

Legally, we don't need the full text in each file. Let's use this very short version:

```
# Copyright (c) 2024, Valkey contributors
# All rights reserved.
# SPDX-License-Identifier: BSD 3-Clause
```

src/Makefile
```diff
@@ -383,6 +383,21 @@ else
 	MAYBE_UNINSTALL_REDIS_SYMLINK=
 endif
 
+fast_float/fast_float_strtod.o: 
+	cd fast_float && $(MAKE) CFLAGS=" $(CFLAGS) " LDFLAGS=" $(LDFLAGS) "
```
zuiderkwast: The default make target is the first target in the file. Here, you're adding this target above the `all` target, so when I type `make`, it only builds this file.

```
$ make
    CC Makefile.dep
cd fast_float && make CFLAGS=" -funwind-tables " LDFLAGS="  "
make[1]: Entering directory '/home/viktor/repos/valkey/src/fast_float'
g++ -std=c++11 -O3 -fPIC -D FASTFLOAT_ALLOWS_LEADING_PLUS -funwind-tables   -c fast_float_strtod.cpp -o fast_float_strtod.o
make[1]: Leaving directory '/home/viktor/repos/valkey/src/fast_float'
```

This needs to be somewhere below `all:`.

----

I tried changing this to get rid of the separate Makefile under src/fast_float/ and to use the same optimization and debug flags as for the rest of Valkey, because it can be configured for debugging, etc.

I just added this code below the rules for `%.o:` and `unit/%.o:`

```
ifeq ($(USE_FAST_FLOAT),yes)
CXX ?= c++
CXXFLAGS += -Wall -W $(OPT) $(DEBUG)
fast_float/fast_float_strtod.o: fast_float/fast_float_strtod.cpp ../deps/fast_float/fast_float.h
	$(QUIET_CXX)$(CXX) $(CXXFLAGS) $(CFLAGS) $(SERVER_CFLAGS) -D FASTFLOAT_ALLOWS_LEADING_PLUS -c $< -o $@
endif
```

The `QUIET_CXX` variable is just for nice output, because we have something similar for CC. The definition next to QUIET_CC:

```diff
 ifndef V
 QUIET_CC = @printf '    %b %b\n' $(CCCOLOR)CC$(ENDCOLOR) $(SRCCOLOR)$@$(ENDCOLOR) 1>&2;
+QUIET_CXX = @printf '    %b %b\n' $(CCCOLOR)C++$(ENDCOLOR) $(SRCCOLOR)$@$(ENDCOLOR) 1>&2;
 QUIET_GEN = @printf '    %b %b\n' $(CCCOLOR)GEN$(ENDCOLOR) $(SRCCOLOR)$@$(ENDCOLOR) 1>&2;
 QUIET_LINK = @printf '    %b %b\n' $(LINKCOLOR)LINK$(ENDCOLOR) $(BINCOLOR)$@$(ENDCOLOR) 1>&2;
```

----

With these changes, I don't see the need for the separate sub directory `src/fast_float/`.

src/Makefile
```diff
@@ -427,6 +442,13 @@ ENGINE_TEST_OBJ:=$(sort $(patsubst unit/%.c,unit/%.o,$(ENGINE_TEST_FILES)))
 ENGINE_UNIT_TESTS:=$(ENGINE_NAME)-unit-tests$(PROG_SUFFIX)
 ALL_SOURCES=$(sort $(patsubst %.o,%.c,$(ENGINE_SERVER_OBJ) $(ENGINE_CLI_OBJ) $(ENGINE_BENCHMARK_OBJ)))
 
+# If USE_FAST_FLOAT is yes, then add fast_float libraries as requirements to build server and cli object.
+ifeq ($(USE_FAST_FLOAT),yes)
+	ENGINE_SERVER_OBJ+= fast_float/fast_float_strtod.o
+	ENGINE_CLI_OBJ+= fast_float/fast_float_strtod.o
+endif
```
zuiderkwast: It's possible to move the other definitions to here, so we don't need multiple if for fast float.

The comment doesn't add much. It just repeats what the code says, in a different lanugage. So let's drop it. (Comments are useful if they explain why, and things that are not obvious from the code itself.)

Please use space on both sides of `+=`.

```suggestion
ifeq ($(USE_FAST_FLOAT),yes)
	ENGINE_SERVER_OBJ += fast_float/fast_float_strtod.o
	ENGINE_CLI_OBJ += fast_float/fast_float_strtod.o
	FINAL_CFLAGS += -D USE_FAST_FLOAT
	ifeq (clang,$(CLANG))
		FINAL_LIBS += -lc++
	else
		FINAL_LIBS += -lstdc++
	endif
endif
```

src/fast_float/fast_float_strtod.h
zuiderkwast, line 1: Let's call this file `src/valkey_strtod.h`, named after the function it provides, rather than what it uses behind the scenes. Fast float is just one of the two implementations it can use.

src/fast_float/fast_float_strtod.h
```diff
@@ -0,0 +1,80 @@
...
+ */
+const char *fast_float_strtod(const char *str, double *value);
+
+static inline const char *valkey_strtod(const char *str, double *value) {
+    errno = 0;
+    return fast_float_strtod(str, value);
+}
+
+#else
+
```
zuiderkwast: The non-fast-float case needs the strtod from stdlib, right?

```suggestion

#include <stdlib.h>

```

zuiderkwast (commented):
Great progress.

> On the topic of moving the fast_float_strtod.cpp to src/, the src/Makefile has become a monolith. From my experience so far with this Makefile, I am now strongly against this. We should move towards smaller build units.

This is another topic. We appear to be moving towards CMake in the long term. Let's not "solve" this now. Keep the monolith and flat directory structure.

The Makefile monolith provides a common ground for optimization flags, and debug flags and other stuff. I'm posting make code in the comments below that you can use. (Personally, I don't think this file is messy at all, given all it does. I guess you need to understand make well to like it.)

> Unit test still resides in this same directory because src/unit build is broken.

What do you mean broken? They're fairly new and they work well in the CI. Please explain what problems you encountered so we can help you.

src/fast_float/fast_float_strtod.h
madolson: Also, strtod has the signature `double strtod(const char *str, char **endptr);`, I think we should ideally conform so that it's more normal to follow.

src/fast_float/Makefile
madolson: I also don't think we should add it to makefiles, I think just the source files is enough.

src/Makefile
parthpatel: > The default make target is the first target in the file. Here, you're adding this target above the all target, so when I type make, it only builds this file.

Good catch. I was too busy testing "make test" across all platforms to focus on normal make. Fixed.

Working on the Makefile changes, but I am against a single monolith Makefile. Experienced valkey developers may be able to make out what's happening there but new developers will have significant friction. If everyone else is in agreement, I will push that change.

src/fast_float/fast_float_strtod.h
parthpatel: > Also, strtod has the signature double strtod(const char *str, char **endptr);, I think we should ideally conform so that it's more normal to follow.

I changed the signature to remove some if checks. endptr can be null. Some of the invocations pass null and others don't. I think this is a minor change. 

> Let's call this file src/valkey_strtod.h,

Done.

parthpatel:
> What do you mean broken? They're fairly new and they work well in the CI. Please explain what problems you encountered so we can help you.

I pasted this earlier somewhere, but this issue has the build log for broken "make test-unit" invocation.
src/unit is broken - https://github.com/valkey-io/valkey/issues/1290

parthpatel:
For other following this PR: I, @madolson  and @zuiderkwast discussed the right folder structure for this change on slack. We decided that the very first version(minus the git submodule part) is the preferred folder/file structure, i.e. deps/fast_float contains the vendor library and deps/fast_float_integration contains the integration code. I will revert this PR to that structure and then we can discuss CMake related changes.

deps/README.md
```diff
@@ -105,3 +106,16 @@ We use a customized version based on master branch commit e4448cf6d1cd08fff51981
 2. Copy updated files from newer version onto files in /hdr_histogram.
 3. Apply the changes from 1 above to the updated files.
 
+fast_float 
+---
+The fast_float library provides fast header-only implementations for the C++ from_chars
+functions for `float` and `double` types as well as integer types.  These functions convert ASCII strings representing decimal values (e.g., `1.3e10`) into binary types. We provide exact rounding (including
+round to even). In our experience, these `fast_float` functions many times faster than comparable number-parsing functions from existing C++ standard libraries.
```
zuiderkwast: "We" and "In our experience" are not our words. I think it's better to just refer to fast float as "it". Also, I'm not sure if all of this is relevant.

```suggestion
functions for `float` and `double` types as well as integer types.  These functions convert ASCII strings representing decimal values (e.g., `1.3e10`) into binary types.
The functions are much faster than comparable number-parsing functions from existing C++ standard libraries.
```

deps/README.md
```diff
@@ -105,3 +106,16 @@ We use a customized version based on master branch commit e4448cf6d1cd08fff51981
...
+
+Specifically, `fast_float` provides the following two functions to parse floating-point numbers with a C++17-like syntax (the library itself only requires C++11):
```
zuiderkwast: This doesn't seem relevant.

But what I think is very important to mention is the distinction between the external dependency and our integration code. That was the source of this long debate.

```suggestion
The `fast_float` directory contains the C++ header file from the fast float project.
The `fast_float_integration` directory contains code to be able to use fast float from C. The integration code is written for Valkey. It's not from the fast float project.
```

deps/fast_float_c_interface/Makefile
```diff
@@ -0,0 +1,15 @@
+CXX?=c++
+# FASTFLOAT_ALLOWS_LEADING_PLUS allows +inf to be parsed as inf, instead of error. 
+CXXFLAGS=-std=c++11 -fPIC -D FASTFLOAT_ALLOWS_LEADING_PLUS
+
+.PHONY: all clean
+all: valkey_strtod.o
+clean:
+	rm -f *.o || true;
```
zuiderkwast, line 36: With `-f`, rm never fails.

```suggestion
	rm -f *.o
```

deps/fast_float_c_interface/valkey_strtod.cpp
```diff
@@ -0,0 +1,46 @@
...
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
+
+#include "../fast_float/fast_float.h"
+#include "valkey_strtod.h"
+#include <cerrno>
+
+extern "C"
+{
+    const char* fast_float_strtod(const char *str, double *value)
```
zuiderkwast: @madolson you mentioned you don't want a file called `fast_float_strtod.cpp` so here we instead have a file called `valkey_strtod.cpp` which provides a function called `fast_float_strtod`. Is this what you had in mind? To me, it seems like some weird compromise that just makes it all more confusing.

deps/fast_float_c_interface/valkey_strtod.h
```diff
@@ -0,0 +1,82 @@
...
+ * the actual conversion, which can be significantly faster than standard library functions.
+ *
+ * Refer to https://github.com/fastfloat/fast_float for more information on the underlying library.
+ */
+const char *fast_float_strtod(const char *str, double *value);
+
+static inline const char *valkey_strtod(const char *str, double *value) {
+    errno = 0;
+    return fast_float_strtod(str, value);
+}
```
zuiderkwast: 1. I'm fine with the directory name `fast_float_c_interface`, although it's not what we agreed on in the long chat. We agreed on `fast_float_inteface`. If we keep changing things after making decisions, we have to make new decisions over and over.

2. This file provides a static inline `valkey_strtod` function and it's located under `deps/fast_float_c_interface/`, but it doesn't always use fast float. It only conditionally uses it.

   The purpose of naming the wrapper `valkey_strtod` rather than `fast_float_strtod` was that it should be implementation agnostic to the caller. When it's placed in this directory, you are led to believe it's using fast float even when it isn't. I see that files needing strtod do `#include "../deps/fast_float_c_interface/valkey_strtod.h` to get the `valkey_strtod` function.

   We have three pieces here:

   1. the external fast float C++ header (`deps/fast_float/fast_float.h`)
   2. interface code to be able to call C++ from C. IMO this should be specific to fast float and be called something like `deps/fast_float_c_interface/fast_float_strtod.{cpp,h}`, but @madolson you don't like this name?
   3. conditional code to choose between the fast float and the stdlib implementations. IMO this should be placed in `src/valkey_strtod.h`.

   For iii, compare this to jemalloc. The conditional code for which allocator we use is abstracted by zmalloc and is placed in `src/zmalloc.{c,h}`.

   If you think differently, please convince me. I may be wrong again. :)

.github/workflows/ci.yml
```diff
@@ -116,7 +116,7 @@ jobs:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - name: make
         run: |
-          sudo apt-get update && sudo apt-get install libc6-dev-i386
+          sudo apt-get update && sudo apt-get install libc6-dev-i386 libstdc++-11-dev-i386-cross gcc-multilib g++-multilib
```
zuiderkwast: Why do we need to add this C++ stuff in jobs where we don't use `USE_FAST_FLOAT=yes`? It looks like a mistake.

I don't see that we use `USE_FAST_FLOAT=yes` anywhere in the CI job. I think we should use it in at least one of the CI test jobs. I suggest we use it in `test-ubuntu-latest`, or maybe in most of the jobs but not all.

deps/Makefile
```diff
@@ -116,3 +117,9 @@ jemalloc: .make-prerequisites
 	cd jemalloc && $(MAKE) lib/libjemalloc.a
 
 .PHONY: jemalloc
+
+fast_float_c_interface: .make-prerequisites
+	@printf '%b %b\n' $(MAKECOLOR)MAKE$(ENDCOLOR) $(BINCOLOR)$@$(ENDCOLOR)
+	cd fast_float_c_interface && $(MAKE) CFLAGS="$(OPT) $(STD) $(WARN) $(DEBUG)" LDFLAGS="$(LDFLAGS)"
```
zuiderkwast: Here, we pass the variable CFLAGS to the next Makefile. We don't pass the individual OTP, STD, WARN and DEBUG variables.

deps/fast_float_c_interface/Makefile
```diff
@@ -0,0 +1,15 @@
...
+
+# src/Makefile passes values for OPT, STD and WARN variables. These are core CFLAGS used by executable targets.
+# We don't need to pass USE_FAST_FLOAT value to compiler here. The code in src/ includes the header files as is,
+# and at that point disables fast_float code based on USE_FAST_FLOAT value. Therefore, only src/Makefile passes 
+# USE_FAST_FLOAT value to compiler.
+valkey_strtod.o: valkey_strtod.cpp valkey_strtod.h
+	$(CXX) $(OPT) $(STD) $(WARN) $(CXXFLAGS) -o $@ -c $<
```
zuiderkwast: The `deps/Makefile` passes the CFLAGS variable, but not the OPT, STD and WARN flags, so these are not defined here. That was probably a typo in deps/Makefile.

The STD passed from src/Makefile contains `-std=c11`, which is not appropriate for C++. The fast float code is written for C++11 so we should use that for fast float.

The WARN we use in src/Makefile contains some warning that is invalid in C++. I got a warning locally when I tried to use the same flags for C++ that we use for C. Each project uses differnt code style and needs a different set of warnings, so let's use whatever fast float recommends, or something generic like `-Wall -Wextra`.

Maybe we can pass OPT and DEBUG from src/Makefile.

deps/fast_float_c_interface/valkey_strtod.cpp
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *   * Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *   * Neither the name of Redis nor the names of its contributors may be used
```
zuiderkwast: Redis is back. :D

src/unit/test_valkey_strtod.c
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * Redistribution and use in source and binary forms, with or without
+ * modification, are permitted provided that the following conditions are met:
+ *
+ *   * Redistributions of source code must retain the above copyright notice,
+ *     this list of conditions and the following disclaimer.
+ *   * Redistributions in binary form must reproduce the above copyright
+ *     notice, this list of conditions and the following disclaimer in the
+ *     documentation and/or other materials provided with the distribution.
+ *   * Neither the name of Redis nor the names of its contributors may be used
```
zuiderkwast: Redis is back.

deps/fast_float_c_interface/valkey_strtod.h
madolson: @zuiderkwast I was expecting this file to end up under src/*, since as you mentioned its our code. This is not CPP code.

deps/fast_float_c_interface/valkey_strtod.cpp
madolson: If it's a generic file that includes both implementations, `valkey_strtod` makes sense. If it's just a fast float implementation, `fast_float_strtod.cpp` makes sense.

deps/fast_float/Makefile
madolson: You can use `-f` like you do the first case

deps/fast_float/fast_float_strtod.cpp
madolson: Then we shouldn't call it strtod, since it's not a strtod replacement.

deps/fast_float_c_interface/valkey_strtod.h
zuiderkwast: It seems we agree to use `deps/fast_float_c_interface/fast_float_strtod.cpp` for the interfacing code and `src/valkey_strtod.h` for our conditional code then?

deps/fast_float/fast_float_strtod.cpp
zuiderkwast: @madolson I agree.

Why can't we return the double though? And is checking for NULL noticeable performance-wise?

In C, `strtod` is the standard way to convert a string to a double, so I think we should prefer emulating it. In C++ there are other standard functions that fast float is emulating.

@parthpatel can you show in a comment how properly emulating strtod would look like? One more if-else you say...

deps/Makefile
```diff
@@ -116,3 +117,9 @@ jemalloc: .make-prerequisites
...
+
+.PHONY: jemalloc
```
madolson:
```suggestion
.PHONY: fast_float_c_interface
```

deps/Makefile
parthpatel: yeah I forgot to update this. good catch. fixing it.

deps/fast_float_c_interface/valkey_strtod.h
madolson: Yes, I agree with that.

deps/fast_float_c_interface/valkey_strtod.cpp
parthpatel: yeah, we hadn't fixed that yet in that old commit. I will fix it in next commit.

deps/fast_float_c_interface/Makefile
parthpatel: It does on mac. 
```
/tmp % rm -f *.0
zsh: no matches found: *.0
```

.github/workflows/ci.yml
parthpatel: This is old commit. I will add a separate job.

src/Makefile
```diff
@@ -455,7 +472,7 @@ persist-settings: distclean
 	echo SERVER_LDFLAGS=$(SERVER_LDFLAGS) >> .make-settings
 	echo PREV_FINAL_CFLAGS=$(FINAL_CFLAGS) >> .make-settings
 	echo PREV_FINAL_LDFLAGS=$(FINAL_LDFLAGS) >> .make-settings
-	-(cd ../deps && $(MAKE) $(DEPENDENCY_TARGETS))
+	-(cd ../deps && $(MAKE) $(DEPENDENCY_TARGETS) OPT="$(OPT)" STD="$(STD)" WARN="$(WARN)")
```
madolson: Removing this line fixes the freebsd build. I guess we are now changing all of the flags in deps during build. It's probably the flto flag, but not sure. It seems like only fpconv, hdr_histogram and linenoise also use there variables.

deps/fast_float_c_interface/Makefile
madolson: `-(rm -f *.o)` I guess would be the normal makefile way, to ignore the error.

src/Makefile
parthpatel: Yeah, I did not realize that OPT, STD etc are standard variables used by many Makefiles. I will remove them and explicitly specify what to use in fast_float. I won't pass anything from main Makefile. This is unnecessary - now that I think about it.

deps/fast_float_c_interface/valkey_strtod.h
parthpatel: Moving valkey_strtod.h to src/.

deps/fast_float_c_interface/Makefile
parthpatel: I cleaned up all of this - dropped the idea of passing anything from src/Makefile for now. Explicitly specifying what we need in fast_float_c_interface/Makefile.

deps/fast_float_c_interface/Makefile
zuiderkwast: MacOS have ridiculously outdated coreutils, because they refused to lift to GPLv3. I guess it's one of the reasons I'm running fedora on my macbook instead of macOS, hehe.

.github/workflows/ci.yml
```diff
@@ -110,6 +128,13 @@ jobs:
       - name: make
         run: make -j3 all-with-unit-tests SERVER_CFLAGS='-Werror'
 
+  build-macos-latest-fast-float:
```
madolson: Forward looking this will be the default, I would just update all of the existing builds to include it. Our MacOS runners are often in demand and we don't need 4 more tests to execute a couple of lines of code.

Either that we have default build + extras build.

.github/workflows/ci.yml
```diff
@@ -119,6 +144,13 @@ jobs:
           sudo apt-get update && sudo apt-get install libc6-dev-i386
           make -j4 SERVER_CFLAGS='-Werror' 32bit
 
+  build-32bit-fast-float:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: make
+        run: "# Fast float requires C++ 32-bit libraries to compile on 64-bit ubuntu machine i.e. \"-cross\" suffixed version. \n# Cross-compiling c++ to 32-bit also requires multilib support for g++ compiler i.e. \"-multilib\" suffixed version of g++.\n# g++-multilib should generally include libstdc++.*cross version as well, but also added explicitly just in case. \nsudo apt-get update && sudo apt-get install libc6-dev-i386 libstdc++-11-dev-i386-cross gcc-multilib g++-multilib\nmake -j4 SERVER_CFLAGS='-Werror' 32bit USE_FAST_FLOAT=yes\n#magic___^_^___line\n"
```
madolson: Can we move this comment outside of the run step?

deps/fast_float_c_interface/fast_float_strtod.cpp
```diff
@@ -0,0 +1,28 @@
...
+ *    * Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+ *    * Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+*/
```
madolson:
```suggestion
/*
 * Copyright Valkey Contributors.
 * All rights reserved.
 * SPDX-License-Identifier: BSD 3-Clause
 */
```

src/valkey_strtod.h
```diff
@@ -0,0 +1,66 @@
...
+ *    * Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+ *    * Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
```
madolson:
```suggestion
```

src/unit/test_valkey_strtod.c
```diff
@@ -0,0 +1,50 @@
...
+ *   * Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the following disclaimer in the documentation and/or other materials provided with the distribution.
+ *   * Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
```
madolson:
```suggestion
/*
 * Copyright Valkey Contributors.
 * All rights reserved.
 * SPDX-License-Identifier: BSD 3-Clause
 */
```

src/valkey_strtod.h
```diff
@@ -0,0 +1,66 @@
...
+#ifndef FAST_FLOAT_STRTOD_H
+#define FAST_FLOAT_STRTOD_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
```
madolson:
```suggestion
```
This is never imported by C++ code anymore.

src/valkey_strtod.h
```diff
@@ -0,0 +1,66 @@
...
+    char *endptr;
+    *value = strtod(str, &endptr);
+    return endptr;
+}
+
+#endif
+
+#ifdef __cplusplus
+}
+#endif
```
madolson:
```suggestion
```

src/unit/test_valkey_strtod.c
```diff
@@ -0,0 +1,50 @@
...
+    value = 0;
+    valkey_strtod("-inf", &value);
+    assert(isinf(value));
+    assert(errno == 0);
+
+    value = 0;
+    valkey_strtod("inf", &value);
+    assert(isinf(value));
+    assert(errno == 0);
+#endif
```
madolson:
```suggestion
    double value = 0;
    errno = 0;
    valkey_strtod("231.2341234", &value);
    TEST_ASSERT(value == 231.2341234);
    TEST_ASSERT(errno == 0);

    value = 0;
    valkey_strtod("+inf", &value);
    TEST_ASSERT(isinf(value));
    TEST_ASSERT(errno == 0);

    value = 0;
    valkey_strtod("-inf", &value);
    TEST_ASSERT(isinf(value));
    TEST_ASSERT(errno == 0);

    value = 0;
    valkey_strtod("inf", &value);
    TEST_ASSERT(isinf(value));
    TEST_ASSERT(errno == 0);
```

src/unit/test_valkey_strtod.c
```diff
@@ -0,0 +1,50 @@
...
+ *   * Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote products derived from this software without specific prior written permission.
+ *
+ * THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+ */
+
+
+#include "../valkey_strtod.h"
+#include "assert.h"
```
madolson:
```suggestion
```

madolson (commented):
Looks really close now, I'm going to just apply some of these comments right away.

.github/workflows/ci.yml
zuiderkwast: I like update existing jobs to use fast_float. Don't add more jobs.

deps/fast_float_c_interface/Makefile
zuiderkwast:
```suggestion
	-(rm -f *.o)
```

src/valkey_strtod.h
```diff
@@ -0,0 +1,46 @@
...
+ *               On failure, returns char pointer pointing to first invalid character in the string.
+ *
+ * note: This function uses the fast_float library (https://github.com/fastfloat/fast_float) for
+ * the actual conversion, which can be significantly faster than standard library functions.
+ *
+ * Refer to https://github.com/fastfloat/fast_float for more information on the underlying library.
+ */
+const char *fast_float_strtod(const char *str, double *value);
+
+static inline const char *valkey_strtod(const char *str, double *value) {
```
zuiderkwast: * Refer to deps/fast_float_c_interface/ instead.
* Change to the strtod return value and behaviour.

.github/workflows/ci.yml
parthpatel: yamlfmt complained, so I just applied whatever its suggestion was :)

.github/workflows/ci.yml
parthpatel: > Looks really close now, I'm going to just apply some of these comments right away.

Go for it.

.github/workflows/ci.yml
zuiderkwast: Suggestions from clang-format and similar are not always good.

I don't like this very long oneline string. Let's find something better. You already have this in another place:

```

  test-ubuntu-latest-fast-float:
    runs-on: ubuntu-latest
    steps:
      - name: make
        # build with TLS and FAST_FLOAT just for compilation coverage
        run: make -j4 all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes USE_FAST_FLOAT=yes
      - name: test
```

.github/workflows/ci.yml
```diff
@@ -34,6 +34,25 @@ jobs:
         run: |
           ./src/valkey-unit-tests
 
+  # Similar to the main build, but include additional feature coverage
+  test-ubuntu-latest-extra:
```
madolson: I applied what I think makes sense for CI. @zuiderkwast if you have a second, please take a look.

madolson:
Running ASAN, https://github.com/valkey-io/valkey/actions/runs/11966743688.

.github/workflows/ci.yml
zuiderkwast: Yes, thanks. I see you modified 32-bit and macos jobs to enable fast float and not add new jobs for these. That's good. Adding 1 new job is better than adding 3, although I suggested adding zero.

I was suggesting enabling it `test-ubuntu-latest` too. We have for example the `test-sanitizer-address` and `test-ubuntu-latest-cmake ` that run without fast float. But it's fine to add this one job. We can maybe clean the jobs once a year or so.

madolson: Yeah, I ended up with just one new test in the CI for now. I suppose you're saying the CI could do fast float and by default, and daily could hit both since we have like 8 different tests. I guess that makes sense, let me adapt that.

src/Makefile
```diff
@@ -433,13 +433,6 @@ ifeq ($(USE_FAST_FLOAT),yes)
 	DEPENDENCY_TARGETS += fast_float_c_interface
 	FAST_FLOAT_STRTOD_OBJECT := ../deps/fast_float_c_interface/fast_float_strtod.o
 	FINAL_LIBS += $(FAST_FLOAT_STRTOD_OBJECT)
-	# fast_float_strtod uses functions from c++ std library, so let's add it for linking as well.
-	# Clang uses libc++ for standard c++ library vs libstdc++ used by gcc.
-	ifeq (clang,$(CLANG))
-		FINAL_LIBS += -lc++
-	else
-		FINAL_LIBS += -lstdc++
-	endif
```
zuiderkwast: @madolson Why don't we need these?

madolson: The only thing I saw being used from the standardlib was exceptions, which we don't need and don't want. All the builds seems to be succeeding as well.

madolson (approved):
LGTM, @eifrah-aws can you help followup with the CMake changes like we discussed offline?

deps/fast_float_c_interface/fast_float_strtod.cpp
```diff
@@ -0,0 +1,26 @@
...
+extern "C"
+{
+    double fast_float_strtod(const char *str, const char** endptr)
+    {
+        double temp = 0;
+        auto answer = fast_float::from_chars(str, str + strlen(str), temp);
+        if (answer.ec != std::errc()) {
+            errno = (answer.ec == std::errc::result_out_of_range) ? ERANGE : EINVAL;
+        } else {
+            errno = 0;
```
zuiderkwast: Library functions (like stdlib's strtod and others) don't set errno to zero on success. That's something the caller can do, and we do it in the calling code.

```suggestion
```

zuiderkwast (approved):
Thanks for pushing this through! Just one nit.

swaingotnochill (approved):

swaingotnochill:
should we do a benchmark on different architectures for this change?

zuiderkwast:
> should we do a benchmark on different architectures for this change?

@swaingotnochill That never harms! I think it only affects commands with floats, like INCRBYFLOAT and sorted sets, so I would only benchmark things like that.

Issue #1228, opened 2024-10-25
valkey erroring in 8.0.x but not in 7.2.7 - NOSUB 'unsubscribe' command executed not in subscribed mode


amirreza-sf80:
**Describe the bug**

we are trying to add a valkey backend to the celery project
[this is the PR](https://github.com/celery/celery/pull/9300).
as mentioned in the pr, testing celery against `valkey7.2.7` completes
but  when using `latest` this error happens:
` valkey.exceptions.ResponseError: NOSUB 'unsubscribe' command executed not in subscribed mode`

also this is not a consistent error, meaning sometimes a test fails, sometime that test passes but another test fails

**To reproduce**

clone the PR
install dependencies:
```
python -m venv .venv
source .venv/bin/activate

cd requirements
pip install -r default.txt
pip install -r test.txt
pip install -r test-ci-default.txt
pip install -r test-integration.txt
```

`docker run -d -p 5672:5672 -p 15672:15672 --name rabbitmq rabbitmq:management`
`docker run -d -p 6380:6379 --name valkey valkey/valkey:latest` (and 7.2.7)
`docker run -d -p 6379:6379 --name redis redis`
`docker start rabbitmq valkey redis && tox -e 3.12-integration-rabbitmq_valkey && docker stop rabbitmq valkey redis`

**Expected behavior**

tests should pass, they pass with 7.2.7, but fail with 8.0.x

enjoy-binbin:
This is a breaking change introduced in #759, i guess you need to adjust your test to handle this csae? FYI @hwware @zuiderkwast

amirreza-sf80:
hi again

well there are two points:
1. if it is expected, it should be consistent, but as i mentioned this is not consistent

2. i've failed to find where a `unsubscribe` command is called in the code
the second one could be my mistake, but the first one remains

zuiderkwast:
> 1. if it is expected, it should be consistent, but as i mentioned this is not consistent

It's a behaviour change between Valkey 7.2 and 8.0.

On Valkey 8.0, UNSUBSCRIBE returns NOSUB if there are no subscribed channels. Do we have a bug here? Or maybe you have a race in your test case that sometimes there is a subscribed channel and other times there are no subscribed channels?

I think it's safe to change your test case to except either a NOSUB error or a successful unsubscribe.

amirreza-sf80:
hi
by `not consistent` i mean when using version 8, sometimes a test failes with this error, sometimes it doesn't.

[...]

PR #759, opened 2024-07-08
Decline unsubscribe related command in non-subscribed mode

hwware:
Now, when clients run the unsubscribe, sunsubscribe and punsubscribe commands in the non-subscribed mode, it returns 0.
Indeed this is a bug, we should not allow client run these kind of commands here.

Thus, this PR fixes this bug, but it is a break change for existing clients

## Update
This was introduced in Valkey8, and it is a breaking change, this will be revert in #1265

PR #1265, opened 2024-11-05
Revert "Decline unsubscribe related command in non-subscribed mode"

hwware:
This PR goal is to revert the changes on PR https://github.com/valkey-io/valkey/pull/759

Recently, we got some reports that in Valkey 8.0 the PR  https://github.com/valkey-io/valkey/pull/759 (Decline unsubscribe related command in non-subscribed mode) causes break change.   (https://github.com/valkey-io/valkey/issues/1228)

Although from my thought, call commands "unsubscribeCommand", "sunsubscribeCommand", "punsubscribeCommand" in request-response mode make no sense.  This is why I created PR https://github.com/valkey-io/valkey/pull/759

But breaking change is always no good,  @valkey-io/core-team How do you think we revert this PR code changes?

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -4165,12 +4165,6 @@ int processCommand(client *c) {
         return C_OK;
     }
 
-    /* Not allow several UNSUBSCRIBE commands executed under non-pubsub mode */
-    if (!c->flag.pubsub && (c->cmd->proc == unsubscribeCommand || c->cmd->proc == sunsubscribeCommand ||
-                            c->cmd->proc == punsubscribeCommand)) {
-        rejectCommandFormat(c, "-NOSUB '%s' command executed not in subscribed mode", c->cmd->fullname);
-        return C_OK;
-    }
     /* Only allow commands with flag "t", such as INFO, REPLICAOF and so on,
      * when replica-serve-stale-data is no and we are a replica with a broken
      * link with primary. */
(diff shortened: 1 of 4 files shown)


enjoy-binbin (approved):
LGTM, i agree we should revert it for now, and we can discuss it in the future.

zuiderkwast (approved):
I agree, let's revert it.

@PingXie mentioned another point in a chat: The NOSUB error makes UNSUBSCRIBE no longer idempotent.

madolson (approved):

zuiderkwast:
@PingXie has mentioned in a chat:

> btw, I agree with reverting the "nosub" (PR 759) change

We have a majority.

src/server.c
```diff
@@ -4165,12 +4165,6 @@ int processCommand(client *c) {
         return C_OK;
     }
 
-    /* Not allow several UNSUBSCRIBE commands executed under non-pubsub mode */
```
enjoy-binbin, line 4168: Before we merge it, can we find some places that can document this behavior?

hwware: In Valkey 7.2, these 3 commands are same as the other commands, client can call them without any restriction and there is no any special document. Thus, I think we do not need to any document for them. Let me merge this PR.

zuiderkwast: If we want to document it, a note on the `[S|P]UNSUBSCRIBE` command pages would be a good place.

hwware: Which kind of content we should write there? Just curious

zuiderkwast: Example:

*In 8.0.0 and 8.0.1, this command can return an error if unsubscribe is called when not in subscribed mode, when no channels are subscribed. In 8.0.2, the behaviour from 7.2 was restored.*

PR #1271, opened 2024-11-07
Improvements for TLS with I/O threads

uriyage:
Main thread profiling revealed significant overhead in TLS operations, even with read/write offloaded to I/O threads:

Perf results:

**10.82%** 8.82% `valkey-server libssl.so.3 [.] SSL_pending` # Called by main thread after I/O completion 

**10.16%** 5.06% `valkey-server libcrypto.so.3 [.] ERR_clear_error` # Called for every event regardless of thread handling

This commit further optimizes TLS operations by moving more work from the main thread to I/O threads:

Improve TLS offloading to I/O threads with two main changes:

1. Move `ERR_clear_error()` calls closer to SSL operations
   - Currently, error queue is cleared for every TLS event
   - Now only clear before actual SSL function calls
   - This prevents unnecessary clearing in main thread when operations
     are handled by I/O threads

2. Optimize `SSL_pending()` checks
   - Add `TLS_CONN_FLAG_HAS_PENDING` flag to track pending data
   - Move pending check to follow read operations immediately
   - I/O thread sets flag when pending data exists
   - Main thread uses flag to update pending list

Performance improvements:
Testing setup based on https://valkey.io/blog/unlock-one-million-rps-part2/

Before:
- SET: 896,047 ops/sec
- GET: 875,794 ops/sec

After:
- SET: 985,784 ops/sec (+10% improvement)
- GET: 1,066,171 ops/sec (+22% improvement)

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -446,6 +446,7 @@ typedef enum {
 #define TLS_CONN_FLAG_WRITE_WANT_READ (1 << 1)
 #define TLS_CONN_FLAG_FD_SET (1 << 2)
 #define TLS_CONN_FLAG_POSTPONE_UPDATE_STATE (1 << 3)
+#define TLS_CONN_FLAG_HAS_PENDING (1 << 4)
 
 typedef struct tls_connection {
     connection c;
@@ -614,7 +615,7 @@ static void updatePendingData(tls_connection *conn) {
 
     /* If SSL has pending data, already read from the socket, we're at risk of not calling the read handler again, make
      * sure to add it to a list of pending connection that should be handled anyway. */
-    if (SSL_pending(conn->ssl) > 0) {
+    if (conn->flags & TLS_CONN_FLAG_HAS_PENDING) {
         if (!conn->pending_list_node) {
             listAddNodeTail(pending_list, conn);
             conn->pending_list_node = listLast(pending_list);
@@ -622,6 +623,14 @@ static void updatePendingData(tls_connection *conn) {
     } else if (conn->pending_list_node) {
         listDelNode(pending_list, conn->pending_list_node);
         conn->pending_list_node = NULL;
+    }
+}
+
+void updateSSLPendingFlag(tls_connection *conn) {
+    if (SSL_pending(conn->ssl) > 0) {
+        conn->flags |= TLS_CONN_FLAG_HAS_PENDING;
+    } else {
+        conn->flags &= ~TLS_CONN_FLAG_HAS_PENDING;
     }
 }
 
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
Looks good to me, but I'd like @madolson to take a look too.

Impressing performance result, and this is with more lines deleted than added!

Does the change have any performance impact without I/O threads enabled?

madolson:
Will check tomorrow, but @uriyage please fix the DCO when you get a chance.

ranshid:
> Looks good to me, but I'd like @madolson to take a look too.
> 
> Impressing performance result, and this is with more lines deleted than added!
> 
> Does the change have any performance impact without I/O threads enabled?

I would be surprised if there was any performance improvement when io-threads are not used. This will actually cause SSL_pending and  ERR_clear_error to be called multiple times in some scenarios where we are reading multiple times on the read handler. BUT I think those cases are rare and AFAIK are mostly limited to cases like the connSyncReadLine.

src/tls.c
```diff
@@ -1021,9 +1024,7 @@ static ssize_t connTLSSyncRead(connection *conn_, char *ptr, ssize_t size, long
     tls_connection *conn = (tls_connection *)conn_;
 
     setBlockingTimeout(conn, timeout);
-    ERR_clear_error();
-    int ret = SSL_read(conn->ssl, ptr, size);
-    ret = updateStateAfterSSLIO(conn, ret, 0);
+    int ret = connTLSRead(conn_, ptr, size);
```
ranshid: I think this will cause us to potentially keep updating the event in updateStateAfterSSLIO. Same for the  connTLSSyncReadLine. Maybe you want to create an internal function which will also get the update_event indication?

madolson (commented):
Other than Ran's comment, this looks good to me. Nice perf results!

src/tls.c
uriyage: Thanks, I reverted this change. Instead, we will call `updatePendingFlags` explicitly.

uriyage:
> Looks good to me, but I'd like @madolson to take a look too.
> 
> Impressing performance result, and this is with more lines deleted than added!
> 
> Does the change have any performance impact without I/O threads enabled?

As expected, there is no performance difference without I/O threads as the behavior remains the same.


> I would be surprised if there was any performance improvement when io-threads are not used. This will actually cause SSL_pending and ERR_clear_error to be called multiple times in some scenarios where we are reading multiple times on the read handler. BUT I think those cases are rare and AFAIK are mostly limited to cases like the connSyncReadLine.

The only change is indeed for `connSyncReadLine`, but as we use it to call the read syscall one byte at a time, I believe performance is not our concern in this case.

zuiderkwast (approved):

enjoy-binbin (approved):

madolson (approved):

PR #1272, opened 2024-11-07
Make sure to copy null terminator byte in dual channel code

murphyjacob4:
As @madolson pointed out, these do have proper null terminators. This cleans them up to follow the rest of the code which copies the last byte explicitly, which should help reduce cognitive load and make it more resilient should code refactors occur (e.g. non-static allocation of memory, changes to other functions).

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -2698,6 +2698,7 @@ static int dualChannelReplHandleEndOffsetResponse(connection *conn, sds *err) {
     /* Initiate repl_provisional_primary to act as this replica temp primary until RDB is loaded */
     server.repl_provisional_primary.conn = server.repl_transfer_s;
     memcpy(server.repl_provisional_primary.replid, primary_replid, CONFIG_RUN_ID_SIZE);
+    server.repl_provisional_primary.replid[CONFIG_RUN_ID_SIZE] = '\0';
     server.repl_provisional_primary.reploff = reploffset;
     server.repl_provisional_primary.read_reploff = reploffset;
     server.repl_provisional_primary.dbid = dbid;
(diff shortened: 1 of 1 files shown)


src/replication.c
```diff
@@ -2698,6 +2698,7 @@ static int dualChannelReplHandleEndOffsetResponse(connection *conn, sds *err) {
     /* Initiate repl_provisional_primary to act as this replica temp primary until RDB is loaded */
     server.repl_provisional_primary.conn = server.repl_transfer_s;
     memcpy(server.repl_provisional_primary.replid, primary_replid, CONFIG_RUN_ID_SIZE);
+    server.repl_provisional_primary.replid[CONFIG_RUN_ID_SIZE] = '\0';
```
madolson:
```suggestion
    memcpy(server.repl_provisional_primary.replid, primary_replid, sizeof(server.repl_provisional_primary.replid));
```

I'm partial to this, which seems to be the convention used elsewhere, to make sure we're copying the entire buffer. Then it's less likely for these two lines to get split.

This one is static storage class, so it'll always be '\0' so this one should be fine.

src/replication.c
```diff
@@ -4270,6 +4271,7 @@ void replicationResurrectProvisionalPrimary(void) {
      * drain. */
     replicationCreatePrimaryClientWithHandler(server.repl_transfer_s, server.repl_provisional_primary.dbid, NULL);
     memcpy(server.primary->replid, server.repl_provisional_primary.replid, CONFIG_RUN_ID_SIZE);
+    server.primary->replid[CONFIG_RUN_ID_SIZE] = '\0';
```
madolson: I think we're protected here as well because replicationCreatePrimaryClientWithHandler will initialize it with a '\0' (although we set it with a wrong value initially, and then we have to fix it here, which is a bit weird).

src/replication.c
murphyjacob4: I think the codebase is a bit split, e.g. see https://github.com/valkey-io/valkey/blob/5ea9b52e7395db48aa0bae9410b3e2363964bee2/src/replication.c#L3134

But I agree with your logic here with using sizeof.

And you're right, this one wouldn't be a bug so long as it zeroed out on initialization

src/replication.c
murphyjacob4: You're right, I missed that there is a copy of the null terminator happening inside the function above. So it should not have a bug.

I do think copying the extra byte reduces the cognitive load for this code, so I think we should proceed with the change.

I'll rename the PR to make it clear that this is a code cleanup.

madolson (approved):

Issue #1261, opened 2024-11-04
[NEW] Add scripting languages (for EVAL, etc.) using module API


zuiderkwast:
**The problem/use-case that the feature addresses**

1. Allow adding more scripting languages
2. Allow replacing the vendored Lua implementation with another one (e.g. LuaJIT or Lua 5.4)

**Description of the feature**

Looking at `function.c`, there is work started that should allows different "engines" for functions (FUNCTION CREATE, FCALL, etc.). For example, there is a function to register an engine. Currenly, Lua is the only engine, implemented in `functions_lua.c`. There is some separation here.

Extend the existing modularity to the module API: `ValkeyModule_RegisterScriptingEngine` or similar.

The module registers a callback that is invoked for executing code in commands like the EVAL, EVALSHA and FCALL.

In the beginning of an EVAL script, users can add a shebang, a line like `#!lua` and some optional flags or parameters, to select the scripting engine. This mechanism already exists, but currently, only "lua" exists. A module should be able to provide their own languages.

To add Lua engine in parallel to the built-in Lua implementation, the module can register with a different name like "lua5.4", "luajit". For a module to be able to replace the default "lua" engine, the built-in Lua support needs to be disabled. For that, see #1204.

**Alternatives you've considered**

...

**Additional information**

Related discussions:

* #1204
* #1229
* #387

hpatro:
This will be a good addition to Valkey to provide the underlying abstraction to support new engines easily. 

Few questions which comes to my mind and some discussed in the weekly meeting

1. Do we support multiple engines at a given point in time?
2. Do we plan to host other first party engines in the near future like V8 engine? KeyDB in the past created this: https://github.com/Snapchat/ModJS
3. If yes, do we route them based on the shebang (do we fail the script without shebang ?)
4. or introduce new command/sub-command for each engine ?
5. We would need certain API to describe the supported engine and it's version.

neomantra:
> Allow replacing the vendored Lua implementation with another one (e.g. LuaJIT or Lua 5.4)

[...]

Issue #1229, opened 2024-10-25
[NEW] Allow building with LuaJIT


secwall:
**The problem/use-case that the feature addresses**

In some cases Valkey users put some not-trivial logic in lua scripts. LuaJIT is a bit more performant than Lua (but has some limitations). We could add ability to build Valkey with LuaJIT instead of Lua.

**Description of the feature**

Add a configuration option `USE_LUAJIT`, that would build Valkey with LuaJIT instead of Lua. Proof-of-concept is here: https://github.com/secwall/valkey/tree/luajit. Most tests are passing (large strings are not possible in LuaJIT, so 3 GB cjson test fails). For now only x86_64 arch is sandboxed, but arm64 seem to work (at least on my Macbook M1).

**Alternatives you've considered**

I've tried to take a look at https://luau.org. But it's not a drop-in replacement of Lua and requires some hacks to implement it properly.

**Additional information**

I've tried to run a benchmark from #1168.
Unstable:
```
  throughput summary: 9683.35 requests per second
  latency summary (msec):
          avg       min       p50       p95       p99       max
        5.156     0.112     5.207     6.423     7.967    10.815
```
LuaJIT:
```
  throughput summary: 19650.22 requests per second
  latency summary (msec):
          avg       min       p50       p95       p99       max
        2.536     0.088     2.535     3.055     3.647     6.295
```
CPU: `Intel(R) Xeon(R) CPU E5-2660 v4 @ 2.00GHz`, no save, io-threads=9.

efimkin:
wow!

zuiderkwast:
Great! Are the benchmarks the same as in #1168 where you work with values of size 100KB? Can you test it with a more common workload? I believe LuaJIT is faster, but it would be interesting to see some more benchmark comparisons:

* Small or large values
* More values, such as multiple keys or multiple fields from hash or set type keys
* A few lines or many lines of Lua code

LuaJIT is compatible with Lua 5.1, which is what we need. It has been discussed in Redis a few years ago but it was not accepted.

If we add some option to build Valkey without Lua (#1204), we can also make it possible to choose a different Lua implementation.

We have some patches to our vendored Lua verions, to make it more sandboxed and safe(?). Do we need to vendor LuaJIT and add similar patches for that? It would be nice if we can avoid patching it and vendoring it.

secwall:
> Are the benchmarks the same as in https://github.com/valkey-io/valkey/pull/1168 where you work with values of size 100KB?

[...]

PR #1277, opened 2024-11-08
Adds support for scripting engines as Valkey modules

rjd15372:
This PR extends the module API to support the addition of different scripting engines to execute user defined functions.

The scripting engine can be implemented as a Valkey module, and can be dynamically loaded with the `loadmodule` config directive, or with the `MODULE LOAD` command.

This PR also adds an example of a dummy scripting engine module, to show how to use the new module API. The dummy module is implemented in `tests/modules/helloscripting.c`.

The current module API support, only allows to load scripting engines to run functions using `FCALL` command.

The additions to the module API are the following:

```c
/* This struct represents a scripting engine function that results from the
 * compilation of a script by the engine implementation. */
struct ValkeyModuleScriptingEngineCompiledFunction

typedef ValkeyModuleScriptingEngineCompiledFunction **(*ValkeyModuleScriptingEngineCreateFunctionsLibraryFunc)(
    const char *code,
    size_t timeout,
    char **err);

typedef void (*ValkeyModuleScriptingEngineCallFunctionFunc)(
    ValkeyModuleCtx *module_ctx,
    ValkeyModuleScriptingEngineFunctionCtx *func_ctx,
    void *compiled_function,
    ValkeyModuleString **keys,
    size_t nkeys,
    ValkeyModuleString **args,
    size_t nargs);

typedef size_t (*ValkeyModuleScriptingEngineGetUsedMemoryFunc)(
    ValkeyModuleScriptingEngineCtx *engine_ctx);

typedef size_t (*ValkeyModuleScriptingEngineGetFunctionMemoryOverheadFunc)(
    void *compiled_function);

typedef size_t (*ValkeyModuleScriptingEngineGetEngineMemoryOverheadFunc)(
    ValkeyModuleScriptingEngineCtx *engine_ctx);

typedef void (*ValkeyModuleScriptingEngineFreeFunctionFunc)(
    void *compiled_function);

/* This struct stores the callback functions implemented by the scripting
 * engine to provide the functionality for the `FUNCTION *` commands. */
typedef struct ValkeyModuleScriptingEngineMethodsV1 {
    uint64_t version; /* Version of this structure for ABI compat. */

    /* Library create function callback. When a new script is loaded, this
     * callback will be called with the script code, and returns a list of
     * ValkeyModuleScriptingEngineCompiledFunc objects. */
    ValkeyModuleScriptingEngineCreateFunctionsLibraryFunc create_functions_library;

    /* The callback function called when `FCALL` command is called on a function
     * registered in this engine. */
    ValkeyModuleScriptingEngineCallFunctionFunc call_function;

    /* Function callback to get current used memory by the engine. */
    ValkeyModuleScriptingEngineGetUsedMemoryFunc get_used_memory;

    /* Function callback to return memory overhead for a given function. */
    ValkeyModuleScriptingEngineGetFunctionMemoryOverheadFunc get_function_memory_overhead;

    /* Function callback to return memory overhead of the engine. */
    ValkeyModuleScriptingEngineGetEngineMemoryOverheadFunc get_engine_memory_overhead;

    /* Function callback to free the memory of a registered engine function. */
    ValkeyModuleScriptingEngineFreeFunctionFunc free_function;
} ValkeyModuleScriptingEngineMethodsV1;

/* Registers a new scripting engine in the server.
 *
 * - `engine_name`: the name of the scripting engine. This name will match
 *   against the engine name specified in the script header using a shebang.
 *
 * - `engine_ctx`: engine specific context pointer.
 *
 * - `engine_methods`: the struct with the scripting engine callback functions
 * pointers.
 */
int ValkeyModule_RegisterScriptingEngine(ValkeyModuleCtx *ctx,
                                         const char *engine_name,
                                         void *engine_ctx,
                                         ValkeyModuleScriptingEngineMethods engine_methods);

/* Removes the scripting engine from the server.
 *
 * `engine_name` is the name of the scripting engine.
 *
 */
int ValkeyModule_UnregisterScriptingEngine(ValkeyModuleCtx *ctx, const char *engine_name);
```


rjd15372:
I'm opening this PR in draft mode because I still need to fix test failures, and add new tests as well.

rjd15372:
I just realized I still need to implement the `MODULE UNLOAD` support. The existing code was not prepared to allow the deletion of a scripting engine.

tests/modules/helloscripting.c
```diff
@@ -0,0 +1 @@
+../../src/modules/helloscripting.c
```
zuiderkwast: Is this a symlink?

It's not portable to include symlinks in a git repo and we haven't done this so far. Just move the file to tests instead. I don't think we need it as an example module under src/modules/.

madolson:
Core team said we are directionally aligned with, we still want someone to take a deeper look to make sure the details makes sense.

src/functions.c
```diff
@@ -400,34 +409,96 @@ libraryJoin(functionsLibCtx *functions_lib_ctx_dst, functionsLibCtx *functions_l
...
  *
  * - engine_name - name of the engine to register
- * - engine_ctx - the engine ctx that should be used by the server to interact with the engine */
-int functionsRegisterEngine(const char *engine_name, engine *engine) {
+ * - engine_module - the valkey module that implements this engine
+ * - engine_ctx - the engine ctx that should be used by the server to interact with the engine
+ * - create_func - the callback function that is called when a new script is loaded, which adds
+ *                 adds new functions to the engine function library.
+ * - call_func - the callback function that is called when a function is called e.g., using the
+ *               'FCAL' command.
```
zuiderkwast:
```suggestion
 *               'FCALL' command.
```

tests/unit/moduleapi/scriptingengine.tcl
```diff
@@ -0,0 +1,38 @@
...
+    test {Call scripting engine function: calling foo works} {
+        r fcall foo 0 134
+    } {134}
+
+    test {Call scripting engine function: calling bar works} {
+        r fcall bar 0
+    } {432}
+
+    test {Replace function library and call functions} {
+        set result [r function load replace "#!hello name=mylib\nFUNCTION foo\nARGS 0\nRETURN\nFUNCTION bar\nCONSTI 500\nRETURN"]
```
zuiderkwast, line 63: It's a funny language. I don't quite understand the syntax though. Is it a stack machine?

src/valkeymodule.h
```diff
@@ -794,6 +794,16 @@ typedef void (*ValkeyModuleInfoFunc)(ValkeyModuleInfoCtx *ctx, int for_crash_rep
 typedef void (*ValkeyModuleDefragFunc)(ValkeyModuleDefragCtx *ctx);
 typedef void (*ValkeyModuleUserChangedFunc)(uint64_t client_id, void *privdata);
 
+/* Type definitions for implementing scripting engines modules. */
+typedef struct ValkeyModuleScriptingEngineFunctionLibrary ValkeyModuleScriptingEngineFunctionLibrary;
+typedef struct ValkeyModuleScriptingEngineFunctionCallCtx ValkeyModuleScriptingEngineFunctionCallCtx;
+typedef int (*ValkeyModuleCreateScriptingEngineFunc)(void *engine_ctx, ValkeyModuleScriptingEngineFunctionLibrary *li, const char *code, size_t timeout, char **err);
+typedef void (*ValkeyModuleCallScriptingEngineFunctionFunc)(ValkeyModuleScriptingEngineFunctionCallCtx *func_ctx, void *engine_ctx, void *compiled_function, ValkeyModuleString **keys, size_t nkeys, ValkeyModuleString **args, size_t nargs);
```
zuiderkwast: Can we make all of these follow the same pattern with a ValkeyModuleScriptingEngine prefix? Not sure the suggestion is good, but you get the idea:

```suggestion
typedef int (*ValkeyModuleScriptingEngineCreateFunc)(void *engine_ctx, ValkeyModuleScriptingEngineFunctionLibrary *li, const char *code, size_t timeout, char **err);
typedef void (*ValkeyModuleScriptingEngineFunctionCallFunc)(ValkeyModuleScriptingEngineFunctionCallCtx *func_ctx, void *engine_ctx, void *compiled_function, ValkeyModuleString **keys, size_t nkeys, ValkeyModuleString **args, size_t nargs);
```

rjd15372:
The PR is ready for review.

zuiderkwast (commented):
This looks like great progress. Is it ready for a proper review or are you going to change much more?

Shall we do EVAL as a follow-up PR?

Each engine has only one context which is used for all calls, right? Would it be possible for an engine to use a context (or sub-context, etc.) for each client? This idea came up in the LuaJIT discussion by @secwall in #1229. A separate Lua context per client would isolate each client from each other, which compensates for the fact that Lua isn't completely sandboxed by design. Maybe the engine can organize this by itself, if the engine just has access to the current client at the time of a FCALL or EVAL?

src/valkeymodule.h
rjd15372: Yes, that was my intention, but for some reason I typed the names in this weird form.

tests/unit/moduleapi/scriptingengine.tcl
rjd15372: Yes, it's a stack machine. Although not very useful at the moment :smiley:

tests/modules/helloscripting.c
rjd15372: Ok, I'll remove the file from the `src/modules` directory. That way I'll not need a symlink anymore.

rjd15372:
> This looks like great progress. Is it ready for a proper review or are you going to change much more?

It's ready. I don't think I'll add more changes apart from the changes asked by the reviewers.

> 
> Shall we do EVAL as a follow-up PR?

Yes, that's the plan.

> 
> Each engine has only one context which is used for all calls, right? Would it be possible for an engine to use a context (or sub-context, etc.) for each client? This idea came up in the LuaJIT discussion by @secwall in #1229. A separate Lua context per client would isolate each client from each other, which compensates for the fact that Lua isn't completely sandboxed by design. Maybe the engine can organize this by itself, if the engine just has access to the current client at the time of a FCALL or EVAL?

I'll take a look if it's possible, and will open a follow-up PR with the changes.

tests/unit/moduleapi/scriptingengine.tcl
zuiderkwast: Don't make it more useful than it already is! :D

Some code comments explaining this in the test case and/or the test module would be helpful, even if it's just a dummy language.

I didn't realize we need a real language for this, but I understand we do, to be able to register functions.

zuiderkwast:
Can you avoid force-pushing? It's easier to review what's changed since last time if you just push small commits. You can even merge unstable to it, rather than rebase. We squash-merge PRs anyway in the end and we use the PR title and description as the commit message.

rjd15372:
> Can you avoid force-pushing? It's easier to review what's changed since last time if you just push small commits. You can even merge unstable to it, rather than rebase. We squash-merge PRs anyway in the end and we use the PR title and description as the commit message.

Ah sorry. I wasn't aware that we were squashing the commits upon merge. I'll stop force-pushing.

rjd15372:
Today I added more test cases to test the error code paths.

tests/modules/helloscripting.c
```diff
@@ -0,0 +1,354 @@
...
+ * The language only supports 32-bit integer, and it only allows to return an
+ * integer constant, or return the value passed as the first argument to the
+ * function.
+ *
+ * Example of a program:
+ *
+ * ```
+ * FUNCTION foo
+ * ARGS 0        # pushes the value in the first argument to the top of the
+ * stack RETURN        # returns the current value on the top of the stack
```
zuiderkwast:
```suggestion
 * ARGS 0        # pushes the value in the first argument to the top of the
 *               # stack
 * RETURN        # returns the current value on the top of the stack
```

Not mentioned:

* FUNCTION starts a function?
* RETURN also marks the end of the function?

src/function_lua.c
```diff
@@ -114,8 +114,8 @@ static int luaEngineCreate(void *engine_ctx, functionLibInfo *li, sds blob, size
     lua_pop(lua, 1);                                   /* pop the metatable */
 
     /* compile the code */
-    if (luaL_loadbuffer(lua, blob, sdslen(blob), "@user_function")) {
-        *err = sdscatprintf(sdsempty(), "Error compiling function: %s", lua_tostring(lua, -1));
+    if (luaL_loadbuffer(lua, blob, strlen(blob), "@user_function")) {
+        *err = valkey_asprintf("Error compiling function: %s", lua_tostring(lua, -1));
```
PingXie, line 142: What is the benefit of `valkey_asprintf` over `sdscatprintf`? `err` has been traditionally an sds string. It is not interchangeable with the normal pointer. This change would add cognitive load so I would like to better understand the rationale behind this change.

rjd15372: I had to stop using `sds` strings because Valkey modules can't use that type since it's an internal type of the server.

rjd15372:
@zuiderkwast I've pushed a small update that refactors a bit the code. I've already have branch with the implementation of the EVAL using the scripting engine module API almost done. 

It would be good to finish the review of this PR and merge it, so we can focus on the next PRs.

rjd15372:
Updated the PR description with the new API functions added by this PR.

src/module.c
```diff
@@ -67,6 +67,7 @@
 #include <sys/wait.h>
 #include <fcntl.h>
 #include <string.h>
+#include "functions.h"
```
zuiderkwast: Please put this include among our other includes with `"..."`, above the system includes with `<...>`.

src/strl.c
```diff
@@ -81,6 +84,24 @@ valkey_strlcat(char *dst, const char *src, size_t dsize)
 }
 
 
+/*
+ * This function is similar to asprintf function, but it uses zmalloc for
+ * allocating the string buffer.
+ *
+ * IMPORTANT: don't forget to free the buffer when no longer needed.
+ */
+char *valkey_asprintf(char const *fmt, ...) {
```
zuiderkwast: I wonder why this file `strl.c` is a separate file and not part of `util.c`. A guess is that it is copied from another project, since it has a different license (MIT). If that's the reason, then maybe we should keep it that way and add `valkey_asprintf` in `util.c` instead. WDYT?

src/valkeymodule.h
```diff
@@ -794,6 +795,48 @@ typedef void (*ValkeyModuleInfoFunc)(ValkeyModuleInfoCtx *ctx, int for_crash_rep
...
+typedef void ValkeyModuleScriptingEngineCtx;
+typedef void ValkeyModuleScriptingEngineFunctionCtx;
+
+typedef struct ValkeyModuleScriptingEngineCompiledFunction {
+    char *name;       /* Function name */
+    void *function;   /* Opaque object representing a function, usually it'
+                         the function compiled code. */
+    char *desc;       /* Function description */
+    uint64_t f_flags; /* Function flags */
+} ValkeyModuleScriptingEngineCompiledFunction;
```
zuiderkwast: Can these structs be opaque and defined in module.c? If not, why?

For the other non-opaque structs, we've often included a version field, so we can update it in the future by adding more fields at the end. The `valkeymodule.h` is copied in each module and initialized in a way that permits us to add more fields in the struct. Here is an example:

```C
#define VALKEYMODULE_CLIENTINFO_VERSION 1
typedef struct ValkeyModuleClientInfo {
    uint64_t version; /* Version of this structure for ABI compat. */
    uint64_t flags;   /* VALKEYMODULE_CLIENTINFO_FLAG_* */
    uint64_t id;      /* Client ID. */
    char addr[46];    /* IPv4 or IPv6 address. */
    uint16_t port;    /* TCP port. */
    uint16_t db;      /* Selected DB. */
} ValkeyModuleClientInfoV1;
#define ValkeyModuleClientInfo ValkeyModuleClientInfoV1
#define VALKEYMODULE_CLIENTINFO_INITIALIZER_V1 {.version = 1}
```

Should we do this here too? If not, why not?

src/valkeymodule.h
```diff
@@ -1649,6 +1691,19 @@ VALKEYMODULE_API int (*ValkeyModule_RdbSave)(ValkeyModuleCtx *ctx,
                                              ValkeyModuleRdbStream *stream,
                                              int flags) VALKEYMODULE_ATTR;
 
+VALKEYMODULE_API int (*ValkeyModule_RegisterScriptingEngine)(ValkeyModuleCtx *ctx,
+                                                             const char *engine_name,
+                                                             void *engine_ctx,
```
zuiderkwast: There's a type for engine context. Is this one of those?

```suggestion
                                                             ValkeyModuleScriptingEngineCtx *engine_ctx,
```

src/valkeymodule.h
```diff
@@ -1649,6 +1691,19 @@ VALKEYMODULE_API int (*ValkeyModule_RdbSave)(ValkeyModuleCtx *ctx,
...
+                                                             ValkeyModuleScriptingEngineCreateFunctionsLibraryFunc create_functions_library_func,
+                                                             ValkeyModuleScriptingEngineCallFunctionFunc call_function_func,
+                                                             ValkeyModuleScriptingEngineGetUsedMemoryFunc get_used_memory_func,
+                                                             ValkeyModuleScriptingEngineGetFunctionMemoryOverheadFunc get_function_memory_overhead_func,
+                                                             ValkeyModuleScriptingEngineGetEngineMemoryOverheadFunc get_engine_memory_overhead_func,
+                                                             ValkeyModuleScriptingEngineFreeFunctionFunc free_function_func) VALKEYMODULE_ATTR;
```
zuiderkwast: Do we want to add an optional callback to kill a script or function? It can be invoked by SCRIPT KILL or FUNCTION KILL.

Thinking about extenisility: If we want to add more callbacks in the future, we will have to add another API function. Is it likely that we will want to do that? It's easier if don't need to do that; if we manage to get it right from the beginning. :)

zuiderkwast (commented):
This one looks close now. I think we can merge it soon and add EVAL in the other PR, to keep each PR small.

Regarding the `sds` strings that can't be used in the module API: Did you settle with `char *` for the strings or are you still considering ValkeyModuleString AKA `robj *` for the strings? Just curious.

src/valkeymodule.h
zuiderkwast: Another style of the API could be to use a struct with all the callbacks and then just register the struct.

The struct could have a version field for future extensibility.

Pros and cons? What do you think?

```C
int VM_RegisterScriptingEngine(ValkeyModuleCtx *ctx,
                               const char *engine_name,
                               void *engine_ctx,
                               ValkeyModuleScriptingEngineCallbacks *callbacks);

typedef struct {
    int version; /* version field for ABI compatibility */
    ... callbacks ...
    ValkeyModuleScriptingEngineGetUsedMemoryFunc get_used_memory_func;
    ... etc ...
} ValkeyModuleScriptingEngineCallbacksV1;
typedef ValkeyModuleScriptingEngineCallbacksV1 ValkeyModuleScriptingEngineCallbacks;

/* Modules use this to initialize version field. */
#define VALKEYMODULE_SCRIPTING_ENGINE_CALLBACKS_VERSION 1
```

src/strl.c
rjd15372: Yes, I think that's a good idea. I'll move it to `util.c`.

src/valkeymodule.h
rjd15372: Regarding the `ValkeyModuleScriptingEngineCompiledFunction` struct, this struct is must be populated by the implementation of the scripting engine, and then its fields are read individually by the core implementation. We can make it opaque by providing a new VM_* function that accepts the fields as parameters and returns a pointer to the structure, but I'm not sure we gain much by doing that, specially if in the future we want to add/remove fields from this struct.

Regarding the structure of non-opaque structs to support versioning, I think it's a good idea. I was not aware of this pattern. I'll do the same thing with the non-opaque structs created in this PR and in the following ones.

zuiderkwast: OK, I agree we don't need a VM function just to populate a struct.

Let it be non-opaque and use version field. :+1:

> I'll do the same thing with the non-opaque structs created in this PR and in the following ones.

Good, at least whenever it makes sense. Maybe there are exceptions, where we don't need it to be extensible. :)

src/valkeymodule.h
rjd15372: I think it makes sense to use a struct to group the callbacks because of the extensibility argument. I will change it.

rjd15372:
> Regarding the `sds` strings that can't be used in the module API: Did you settle with `char *` for the strings or are you still considering RedisModuleString AKA `robj *` for the strings? Just curious.

We will use `char*` for strings.

rjd15372:
@zuiderkwast I'm now looking at the structure of `ValkeyModuleTypeMethods`, and how is used, and the function pointers definitions are duplicated in `valkeymodule.h` and `server.h`. Probably I will refactor this PR to follow the same pattern, and avoid using `ValkeyModule*` types in the core code.

zuiderkwast:
> @zuiderkwast I'm now looking at the structure of `ValkeyModuleTypeMethods`, and how is used, and the function pointers definitions are duplicated in `valkeymodule.h` and `server.h`. Probably I will refactor this PR to follow the same pattern, and avoid using `ValkeyModule*` types in the core code.

Good finding! I'm not that familiar with ValkeyModuleTypeMethods, but yes it looks like a similar thing. The module struct just adds versioning, right? Do this if you think it's reasonable.

src/valkeymodule.h
rjd15372: Since I introduced a "versioned struct" to store the engine callbacks, the other structs used by the callback functions don't need to be versioned because they will all be under the umbrella of the engine callback structs version value.

src/functions.h
```diff
@@ -54,9 +54,15 @@
 
 typedef struct functionLibInfo functionLibInfo;
 
+/* ValkeyModule type aliases for scripting engine structs and types. */
+typedef ValkeyModuleScriptingEngineCtx EngineCtx;
+typedef ValkeyModuleScriptingEngineFunctionCtx FunctionCtx;
+typedef ValkeyModuleScriptingEngineCompiledFunction CompiledFunction;
+typedef ValkeyModuleScriptingEngineMethods EngineMethods;
```
zuiderkwast: Nice.

Just thinking, apart from in the module API, we use a lowercase first letter for types and structs, at least in most cases (or always? I'm not sure). For consistency, it's probably best to follow the existing style.

rjd15372: Yeah, you're right. I'll change it to be consistent in the rest of the code.

src/valkeymodule.h
rjd15372: > Do we want to add an optional callback to kill a script or function? It can be invoked by SCRIPT KILL or FUNCTION KILL.

I'll address this in the followup PR that implements EVAL.

rjd15372:
Hi @zuiderkwast, I resolved all comments.

src/function_lua.c
```diff
@@ -93,15 +89,30 @@ static void luaEngineLoadHook(lua_State *lua, lua_Debug *ar) {
...
-static int luaEngineCreate(void *engine_ctx, functionLibInfo *li, sds blob, size_t timeout, sds *err) {
-    int ret = C_ERR;
+static compiledFunction **luaEngineCreate(
+    engineCtx *engine_ctx,
+    const char *blob,
+    size_t timeout,
+    size_t *out_num_compiled_functions,
+    char **err) {
+    compiledFunction **compiled_functions = NULL;
     luaEngineCtx *lua_engine_ctx = engine_ctx;
```
zuiderkwast: It looks like this comment was wrong before this change and it becomes a better match with your code change. :)

It can improved though. The function returns a pointer to an array of compiled functions and the number of elements in this array is set in `out_num_compiled_functions`, right?

Another thing: The indentation of the parameters is the same as the body of the function which makes it slightly harder to see where the body begins. Most other functions are indented like this:

```C
static compiledFunction **luaEngineCreate(engineCtx *engine_ctx,
                                          const char *blob,
                                          ...) {
   ...
}
```

src/functions.c
```diff
@@ -403,37 +423,103 @@ libraryJoin(functionsLibCtx *functions_lib_ctx_dst, functionsLibCtx *functions_l
...
+                            engineMethods *engine_methods) {
     sds engine_name_sds = sdsnew(engine_name);
     if (dictFetchValue(engines, engine_name_sds)) {
         serverLog(LL_WARNING, "Same engine was registered twice");
         sdsfree(engine_name_sds);
         return C_ERR;
     }
 
+    engine *engine = zmalloc(sizeof(struct engine));
+    *engine = (struct engine){
```
zuiderkwast: Normally we use the typedef rather than the struct name.

```suggestion
    engine *engine = zmalloc(sizeof(engine));
    *engine = (engine){
```

src/function_lua.c
```diff
@@ -177,11 +209,12 @@ static void luaEngineCall(scriptRunCtx *run_ctx,
 
     serverAssert(lua_isfunction(lua, -1));
 
+    scriptRunCtx *run_ctx = (scriptRunCtx *)func_ctx;
```
zuiderkwast, line 226: With these changes, is the `scriptRunCtx` only used in `function_lua.c` now? I see you removed it from many functions.

If it's only used in this file, let's move the struct defintion from `script.h` to `function_lua.c`.

src/functions.h
```diff
@@ -65,42 +71,50 @@ typedef struct engine {
...
+    /* Invoking a function, func_ctx is an opaque object (from engine POV).
+     * The func_ctx should be used by the engine to interaction with the server,
      * such interaction could be running commands, set resp, or set
      * replication mode
      */
-    void (*call)(scriptRunCtx *r_ctx,
+    void (*call)(ValkeyModuleCtx *module_ctx,
                  void *engine_ctx,
+                 void *func_ctx,
                  void *compiled_function,
```
zuiderkwast: Why are these `void *` rather than the real types?

src/functions.h
```diff
@@ -65,42 +71,50 @@ typedef struct engine {
...
-                 robj **keys,
+                 ValkeyModuleString **keys,
                  size_t nkeys,
-                 robj **args,
+                 ValkeyModuleString **args,
```
zuiderkwast: ValkeyModuleString is already a typedef of `robj` so why change this? Don't we want to avoid the module-API names in non-module-API code?

src/module.c
```diff
@@ -13032,6 +13052,51 @@ int VM_RdbSave(ValkeyModuleCtx *ctx, ValkeyModuleRdbStream *stream, int flags) {
...
 
+/* Registers a new scripting engine in the server.
+ *
+ * - `engine_name`: the name of the scripting engine. This name will match
+ *   against the engine name specified in the script header using a shebang.
+ *
+ * - `engine_ctx`: engine specific context pointer.
+ *
+ * - `engine_methods`: the struct with the scripting engine callback functions
+ * pointers.
```
zuiderkwast: Please mention that it returns OK/ERR on success/error and that something is logged when it fails.

Nit: Alignment typo.

```suggestion
 *   pointers.
```

src/module.c
```diff
@@ -13032,6 +13052,51 @@ int VM_RdbSave(ValkeyModuleCtx *ctx, ValkeyModuleRdbStream *stream, int flags) {
...
+                  VALKEYMODULE_SCRIPTING_ENGINE_ABI_VERSION,
+                  (unsigned long)engine_methods->version);
+        return VALKEYMODULE_ERR;
+    }
+
+    if (functionsRegisterEngine(engine_name,
+                                ctx->module,
+                                engine_ctx,
+                                engine_methods) != C_OK) {
+        return VALKEYMODULE_ERR;
```
zuiderkwast: Does functionsRegisterEngine log something when it fails, so this is consistent with the other failures of this module API function?

src/module.c
```diff
@@ -13032,6 +13052,51 @@ int VM_RdbSave(ValkeyModuleCtx *ctx, ValkeyModuleRdbStream *stream, int flags) {
...
+    }
+
+    return VALKEYMODULE_OK;
+}
+
+/* Removes the scripting engine from the server.
+ *
+ * `engine_name` is the name of the scripting engine.
+ *
+ */
```
zuiderkwast, line 13144:
```suggestion
 * Returns VALKEYMODULE_OK.
 */
```

src/module.h
```diff
@@ -0,0 +1,14 @@
+#ifndef _MODULE_H_
+#define _MODULE_H_
+
+typedef struct ValkeyModuleCtx ValkeyModuleCtx;
```
zuiderkwast, line 8: Can you add a short comment about this file?

The point seems to be to expose some functions defined in module.c that are not part of the module API, for the core to interact with modules.

src/script_lua.c
```diff
@@ -1269,6 +1269,37 @@ sds luaGetStringSds(lua_State *lua, int index) {
...
+            ret_str[n] = str[i];
+        } else {
+            ret_str[n++] = '\\';
+            ret_str[n] = '0';
+        }
+        n++;
+    }
+
+    return ret_str;
+}
```
zuiderkwast: This function escapes all nulls, i.e replaces them with a backslash and the digit zero. What's this function used for and what's the reasoning for escaping like this? We're not generating C source code, are we?

Can't we just return the string up to the first null?

The function needs a top comment to document it too.

src/util.c
```diff
@@ -1380,3 +1381,25 @@ int snprintf_async_signal_safe(char *to, size_t n, const char *fmt, ...) {
     va_end(args);
     return result;
 }
+
+/*
+ * This function is similar to asprintf function, but it uses zmalloc for
+ * allocating the string buffer.
+ *
+ * IMPORTANT: don't forget to free the buffer when no longer needed.
```
zuiderkwast: I'm not a huge fan of all-caps warnings to the reader. :)

If we just say "returns a freshly allocated string", it should be enough for the reader to know it needs to be free'd.

```suggestion
/* A printf-like function that returns a freshly allocated string.
 *
 * This function is similar to asprintf function, but it uses zmalloc for
 * allocating the string buffer. */
```

zuiderkwast (commented):
Great, it looks really good now. Just a few minor comments now.

src/function_lua.c
rjd15372: It's also used in other places like `eval.c`, `script_lua.c`, and `script.c`

src/functions.h
rjd15372: Because I forgot to change them :smile:

src/functions.h
rjd15372: I forgot to reverse this change. Thanks for noticing it.

src/function_lua.c
zuiderkwast: OK, so it's the function context used by the lua engine? You could add a comment about it at the definition of scriptRunCtx if you want.

rjd15372: It's used by both the Lua engine, and the server infrastructure to manage the scripts execution. I'll add a comment.

src/script_lua.c
rjd15372: The string values stored in the Lua stack may contain null characters, and there is a test case that is testing an invalid shebang line in a lua script by adding null characters in the middle of the shebang line. If we return the string just until the first null, the validation will fail.

To solve this problem I decided to escape the null characters in this function. But now that I'm looking again at it, what I should have done was, instead of perform the escaping of null chars, return the string pointer along with the length of the string, and let the callers deal with the null chars problem.

I'm going to change this code to return the string length as well.

src/module.c
rjd15372: Yes.

rjd15372:
@zuiderkwast resolved all comments.

src/module.h
```diff
@@ -0,0 +1,18 @@
+#ifndef _MODULE_H_
+#define _MODULE_H_
+
+/* This header file exposes a set of functions defined in module.c that are
+ * not part of the module API, but are used by the core to interact with modules
+ */
```
zuiderkwast, line 6: I noticed (while reviewing another PR) that all the module structs and prototypes are in `server.h`. We should keep them in one place, so either move these to `server.h` or move everything module related from `server.h` to `module.h`. (The former seems easier.)

zuiderkwast:
@valkey-io/core-team This implements part of #1261 (EVAL will come in a follow-up PR). We seemed directionally aligned in earlier discussions.

Please approve the API described in the top comment ( :+1: or :-1: ).

src/module.h
rjd15372: I wonder why `server.h` has all these structs and prototypes from all components. Having huge header file makes it harder to find things and to isolate components.
Is there any plans to break up server.h in smaller parts?

rjd15372:
Updated top comment with callback definitions

src/module.h
zuiderkwast: It's historic. Once upon a time, all the C code was all in one file too, redis.c or something. It was broken up into multiple files, but not the header file.

We can break it into pieces too, but let's do it in a separate PR or issue. There are probably a spaghetti dependencies to sort out.

rjd15372: Right, that split can be done in a different PR. I would like to move all module related declarations to `module.h` in a different PR.

madolson:
Madelyn and Wen were OK with moving forward with this PR.

src/valkeymodule.h
```diff
@@ -794,6 +795,87 @@ typedef void (*ValkeyModuleInfoFunc)(ValkeyModuleInfoCtx *ctx, int for_crash_rep
...
+ */
+typedef struct ValkeyModuleScriptingEngineCompiledFunction {
+    char *name;       /* Function name */
+    size_t name_len;  /* The length of the function name string */
+    void *function;   /* Opaque object representing a function, usually it'
+                         the function compiled code. */
+    char *desc;       /* Function description */
+    size_t desc_len;  /* The length of the description string */
+    uint64_t f_flags; /* Function flags */
+} ValkeyModuleScriptingEngineCompiledFunction;
```
zuiderkwast: @PingXie These `name` and `desc` are the strings that were sds before. We can't use sds in the module API, so we have `char *` and `ValkeyModuleString` to choose between.

@rjd15372 you mentioned that these can potentially contain NUL bytes, which is why you needed to add `name_len` and `desc_len`. This seems like a reason to use `robj` (AKA ValkeyModuleString) for these instead of `char *`. They're not in any hot path so the overhead of an robj isn't a problem in this case. Shall we close on this?

rjd15372: I'm fine with using robj here. I'll update the PR soon.

rjd15372: done

src/function_lua.c
```diff
@@ -241,17 +241,13 @@ static void luaEngineFreeFunction(engineCtx *engine_ctx,
...
+                                              sds name,
+                                              sds desc,
                                               luaFunctionCtx *lua_f_ctx,
                                               uint64_t flags) {
     *func = (compiledFunction){
-        .name = name,
-        .name_len = name_len,
-        .desc = desc,
-        .desc_len = desc_len,
+        .name = createStringObject(name, sdslen(name)),
```
zuiderkwast: `luaGetStringSds` duplicates the string and `createStringObject` duplicates it again here.

We can rename/change `luaGetStringSds` to `luaGetStringObject` and avoid copying it twice. Then it's also more consistently passed around as an robj.

`luaGetStringSds` is actually only used for these name and desc, just three calls in this file, nowhere else in the code base.

?

src/functions.h
```diff
@@ -65,42 +71,50 @@ typedef struct engine {
      * timeout - timeout for the library creation (0 for no timeout)
      * err - description of error (if occurred)
      * returns C_ERR on error and set err to be the error message */
```
PingXie: update comments too? or just remove it, the parameters are pretty self-explanatory to me

src/function_lua.c
PingXie: Got it. I think it will be less intrusive and more isolated to convert the char * point to sds at the leaf of the call stack as opposed to propagating it all the way up the stack, i.e, converting char * right after `engine->create` returns. The error case would be rare so overhead is not a concern.

src/functions.c
```diff
@@ -990,12 +1116,42 @@ sds functionsCreateWithLibraryCtx(sds code, int replace, sds *err, functionsLibC
...
-    if (engine->create(engine->engine_ctx, new_li, md.code, timeout, err) != C_OK) {
+    size_t num_compiled_functions = 0;
+    compiledFunction **compiled_functions =
+        engine->create(engine->engine_ctx,
+                       md.code,
+                       timeout,
+                       &num_compiled_functions,
+                       err);
+    if (compiled_functions == NULL) {
+        serverAssert(num_compiled_functions == 0);
```
PingXie, line 1154: this is where I think we should convert char * to sds for `err`. I think we will save a lot of code churn this way and also avoid learning a new concept, i.e., `err` being `char *`

src/valkeymodule.h
PingXie: What about using `const char *` for both `name` and `description`? The life time of these two strings seems to be the same as that of the module (or the server). If so, we don't have to worry about how they are allocated. `const char *` would work just fine. Thoughts?

src/functions.c
zuiderkwast: Ah, you mean only that. Because we return errors like this in many other places where we also use addReplyErrorSds to send the error to the client?

zuiderkwast:
> What about using `const char *` for both `name` and `description`? The life time of these two strings seems to be the same as that of the module (or the server). If so, we don't have to worry about how they are allocated. `const char *` would work just fine. Thoughts?

@PingXie These are the name and description of a *function*, not the engine itself. The lifetime is not that of the module, as functions and libraries can be deleted and redefined using the FUNCTION subcommands. The function names are strings that come from the Lua interpreter (in the Lua case). They can contain NUL characters. There's even a test for that. That's why we needed a `size_t length` along with the char pointers, but now it's changed to an `robj` object instead.

The following Lua code passed to FUNCTION CREATE results in a function with the name "my_hset" and the lib name "mylib":

```lua
#!lua name=mylib

local function my_hset(keys, args)
  local hash = keys[1]
  local time = server.call('TIME')[1]
  return server.call('HSET', hash, '_last_modified_', time, unpack(args))
end

server.register_function('my_hset', my_hset)
```

src/valkeymodule.h
zuiderkwast: @PingXie These are the names and descriptions of a *function*, not the engine itself. The lifetime is not that of the module, as functions and libraries can be deleted and redefined using the FUNCTION subcommands. The function names are strings that come from the Lua interpreter (in the Lua case). They can contain NUL characters. There's even a test for that. That's why we needed a `size_t length` along with the char pointers, but now it's changed to an `robj` object instead.

The following Lua code passed to FUNCTION CREATE results in a function with the name "my_hset" and the lib name "mylib":

```lua
#!lua name=mylib

local function my_hset(keys, args)
  local hash = keys[1]
  local time = server.call('TIME')[1]
  return server.call('HSET', hash, '_last_modified_', time, unpack(args))
end

server.register_function('my_hset', my_hset)
```

PingXie: I see. Would it be simpler to just duplicate these C strings as opposed to refcounting them via robj?

src/valkeymodule.h
```diff
@@ -794,6 +795,85 @@ typedef void (*ValkeyModuleInfoFunc)(ValkeyModuleInfoCtx *ctx, int for_crash_rep
...
+typedef struct ValkeyModuleScriptingEngineCompiledFunction {
+    ValkeyModuleString *name; /* Function name */
+    void *function;           /* Opaque object representing a function, usually it'
+                                 the function compiled code. */
+    ValkeyModuleString *desc; /* Function description */
+    uint64_t f_flags;         /* Function flags */
+} ValkeyModuleScriptingEngineCompiledFunction;
+
+typedef ValkeyModuleScriptingEngineCompiledFunction **(*ValkeyModuleScriptingEngineCreateFunctionsLibraryFunc)(
+    ValkeyModuleScriptingEngineCtx *engine_ctx,
```
PingXie, line 831: since this is a module API, would it make sense to future proof it with a `ValkeyModuleCtx *module_ctx` argument as well even if it is not used today? The analogy I am thinking about is the `this` pointer. Similar comments for the remaining new module scripting engine API.

src/valkeymodule.h
```diff
@@ -1650,6 +1729,14 @@ VALKEYMODULE_API int (*ValkeyModule_RdbSave)(ValkeyModuleCtx *ctx,
                                              ValkeyModuleRdbStream *stream,
                                              int flags) VALKEYMODULE_ATTR;
 
+VALKEYMODULE_API int (*ValkeyModule_RegisterScriptingEngine)(ValkeyModuleCtx *ctx,
+                                                             const char *engine_name,
+                                                             ValkeyModuleScriptingEngineCtx *engine_ctx,
```
PingXie: align the names and order with other new module scripting APIs?
```suggestion
VALKEYMODULE_API int (*ValkeyModule_RegisterScriptingEngine)(ValkeyModuleCtx *module_ctx,
                                                             ValkeyModuleScriptingEngineCtx *engine_ctx,
                                                             const char *engine_name,
```

src/valkeymodule.h
```diff
@@ -794,6 +795,85 @@ typedef void (*ValkeyModuleInfoFunc)(ValkeyModuleInfoCtx *ctx, int for_crash_rep
...
+    ValkeyModuleScriptingEngineGetUsedMemoryFunc get_used_memory;
+
+    /* Function callback to return memory overhead for a given function. */
+    ValkeyModuleScriptingEngineGetFunctionMemoryOverheadFunc get_function_memory_overhead;
+
+    /* Function callback to return memory overhead of the engine. */
+    ValkeyModuleScriptingEngineGetEngineMemoryOverheadFunc get_engine_memory_overhead;
+
+    /* Function callback to free the memory of a registered engine function. */
+    ValkeyModuleScriptingEngineFreeFunctionFunc free_function;
```
PingXie: super nit - moving this next to the "constructor" API (`create_functions_library`)?

src/valkeymodule.h
rjd15372: @PingXie that was the initial implementation before changing it in commit 4a06209e . I don't mind to revert back to the original implementation where I use a char* + size_t to store these names.

src/valkeymodule.h
rjd15372: I agree with you and actually I was already adding `ValkeyModuleCtx*` yesterday. I'll update the PR.

src/function_lua.c
rjd15372: done

src/functions.h
rjd15372: done

src/functions.c
rjd15372: done

src/valkeymodule.h
rjd15372: done

src/valkeymodule.h
rjd15372: done

src/valkeymodule.h
rjd15372: done

rjd15372:
@zuiderkwast @PingXie I've addressed all comments and pushed the commits with the changes.

~~Besides the comments, I've also add a new module API function in commit 3c60980373fc6c98b37b13e07b6c23b4897a4165~~

~~The new module API function is used by the scripting engine implementation to return the result of the execution of a compiled function. This allows to have a clear separation between the scripting engine implementation and the client object that called the `FCALL` or `EVAL` in the future.~~

zuiderkwast:
> I've also add a new module API function in commit [3c60980](https://github.com/valkey-io/valkey/commit/3c60980373fc6c98b37b13e07b6c23b4897a4165)
> 
> The new module API function is used by the scripting engine implementation to return the result of the execution of a compiled function. This allows to have a clear separation between the scripting engine implementation and the client object that called the `FCALL` or `EVAL` in the future.

This is basically a preparation for calling scripts from modules, right?

It's a good idea, but it exposes even more API. If it's not usable from the module API yet, then it's perhaps better to hide it from valkeymodule.h so we don't commit to this exact API just yet...?

For example, I see you named a field `long double d128` but I'm not sure a long double is always 128 bits. I think it can be 80 or 96 bits too. All these field names are stuck in the module API once we add it.

rjd15372:
> This is basically a preparation for calling scripts from modules, right?
> 
> It's a good idea, but it exposes even more API. If it's not usable from the module API yet, then it's perhaps better to hide it from valkeymodule.h so we don't commit to this exact API just yet...?
> 
> For example, I see you named a field `long double d128` but I'm not sure a long double is always 128 bits. I think it can be 80 or 96 bits too. All these field names are stuck in the module API once we add it.

Actually, I think I misunderstood how the module API is implemented. I was under the impression that if the scripting engine wanted to call another command using the `ValkeyModule_Call`, it would have to pass a `ValkeyModuleCtx` object with a fake client connection already. But that is not the case, in all API functions that require to make an internal connection to the cluster using a fake client, they'll create a temporary connection, or use an existing connection from a pool.
Then there are also other API functions that don't even use the `client` pointer from the context object.

Therefore, I'm going to revert this change.

rjd15372:
Revert is done. I think now everything is addressed.

src/valkeymodule.h
```diff
@@ -794,6 +795,91 @@ typedef void (*ValkeyModuleInfoFunc)(ValkeyModuleInfoCtx *ctx, int for_crash_rep
...
+    uint64_t f_flags;         /* Function flags */
+} ValkeyModuleScriptingEngineCompiledFunction;
+
+typedef ValkeyModuleScriptingEngineCompiledFunction **(*ValkeyModuleScriptingEngineCreateFunctionsLibraryFunc)(
+    ValkeyModuleCtx *module_ctx,
+    ValkeyModuleScriptingEngineCtx *engine_ctx,
+    const char *code,
+    size_t timeout,
+    size_t *out_num_compiled_functions,
+    char **err);
```
PingXie, line 835: Thinking about this some more, I'd like to propose returning an integer error code instead:


```suggestion
typedef int (*ValkeyModuleScriptingEngineCreateFunctionsLibraryFunc)(
    ValkeyModuleCtx *module_ctx,
    ValkeyModuleScriptingEngineCtx *engine_ctx,
    const char *code,
    size_t timeout,
    size_t *out_num_compiled_functions,
    ValkeyModuleScriptingEngineCompiledFunction **compiled_functions);
```

This would

1. avoid the assumption that a scripting module would allocate these strings using zmalloc, which can't be enforced by compiler
2. eliminate the "char * to sds" conversion
3. help ensure consistency over the error strings across different scripting engines.

src/valkeymodule.h
```diff
@@ -794,6 +795,91 @@ typedef void (*ValkeyModuleInfoFunc)(ValkeyModuleInfoCtx *ctx, int for_crash_rep
...
+    ValkeyModuleCtx *module_ctx,
+    ValkeyModuleScriptingEngineCtx *engine_ctx);
+
+typedef size_t (*ValkeyModuleScriptingEngineGetFunctionMemoryOverheadFunc)(
+    ValkeyModuleCtx *module_ctx,
+    void *compiled_function);
+
+typedef size_t (*ValkeyModuleScriptingEngineGetEngineMemoryOverheadFunc)(
+    ValkeyModuleCtx *module_ctx,
+    ValkeyModuleScriptingEngineCtx *engine_ctx);
```
PingXie: Can we have a single "get_memory" or "get_info" function that returns all 3 fields in a versioned struct as opposed to having 3 module APIs?

src/valkeymodule.h
PingXie: Yeah I am good with `robj`. No need to revert.

src/valkeymodule.h
rjd15372: I  like the suggestion but I see one problem with 3. How the scripting engine communicates the kind of error that made this function to fail back to the client? 

For instance, most of the times this function might fail because of a syntax error in the script, and the scripting engine can give a nice message specifying where the syntax error is located, which in this case will not be sent to the client.

We might log the error in the server log, but I assume the client of valkey might not have access to the server logs.

src/valkeymodule.h
rjd15372: Yep, I think we can

src/valkeymodule.h
PingXie: Interesting. Do you have an example? How does the built-in Lua interpreter handle syntax errors? I wonder if we could generalize an outgoing line number and an outgoing column number? Or we'd like to provide more detailed context like a real compiler.

rjd15372: Currently, when there's some syntax, or runtime error we send the error message returned by lua:
```
127.0.0.1:6379> function load replace  "#!lua name=mylib\nlocal function foo()\nretrn 1\nend\nserver.register_function('foo', bar)\n"
(error) ERR Error compiling function: user_function:3: '=' expected near '1'

127.0.0.1:6379> function load replace  "#!lua name=mylib\nlocal function foo()\nreturn 1\nend\nserver.register_function('foo', bar)\n"
(error) ERR Error registering functions: ERR user_function:5: Script attempted to access nonexistent global variable 'bar'
```

I think the more generic approach is to allow the scripting engine to return a C string with the error message.

src/valkeymodule.h
rjd15372: I've merged `ValkeyModuleScriptingEngineGetUsedMemoryFunc` and `ValkeyModuleScriptingEngineGetEngineMemoryOverheadFunc`.

I had to keep `ValkeyModuleScriptingEngineGetFunctionMemoryOverheadFunc` because that is the memory overhead of a specific function.

src/valkeymodule.h
zuiderkwast: @PingXie How about we get this PR merged and we can still change the API a bit more before 8.1 is released? Ricardo has the follow-up PR for EVAL waiting to be added. It will slightly modify the API already.

zuiderkwast: @rjd15372 We should document this in some way for the module API docs, for example that the module needs to allocate the `err` using `ValkeyModule_Alloc` (a wrapper for zmalloc).

The comments before each `VM_` function definition gets extracted to produce this page: https://valkey.io/topics/modules-api-ref/

We also put comments for structs in valkeymodule.h, but it's not perfect for documentation. For example, Rust module authors don't directly use this file.

We can consider creating a new documentation topic page for engines later.

PingXie: > How about we get this PR merged and we can still change the API a bit more before 8.1 is released?

Agreed.

PingXie (approved):
Great job 🎉 Thanks @rjd15372!

zuiderkwast (approved):

src/valkeymodule.h
```diff
@@ -794,6 +795,93 @@ typedef void (*ValkeyModuleInfoFunc)(ValkeyModuleInfoCtx *ctx, int for_crash_rep
...
+    ValkeyModuleString *name; /* Function name */
+    void *function;           /* Opaque object representing a function, usually it'
+                                 the function compiled code. */
+    ValkeyModuleString *desc; /* Function description */
+    uint64_t f_flags;         /* Function flags */
+} ValkeyModuleScriptingEngineCompiledFunction;
+
+/* This struct is used to return the memory information of the scripting
+ * engine. */
+typedef struct ValkeyModuleScriptingEngineMemoryInfo {
```
PingXie, line 822: this info struct needs to be versioned. see `ValkeyModuleClientInfo`.

src/valkeymodule.h
```diff
@@ -794,6 +795,93 @@ typedef void (*ValkeyModuleInfoFunc)(ValkeyModuleInfoCtx *ctx, int for_crash_rep
...
+typedef size_t (*ValkeyModuleScriptingEngineGetFunctionMemoryOverheadFunc)(
+    ValkeyModuleCtx *module_ctx,
+    void *compiled_function);
+
+typedef void (*ValkeyModuleScriptingEngineFreeFunctionFunc)(
+    ValkeyModuleCtx *module_ctx,
+    ValkeyModuleScriptingEngineCtx *engine_ctx,
+    void *compiled_function);
+
+typedef ValkeyModuleScriptingEngineMemoryInfo (*ValkeyModuleScriptingEngineGetMemoryInfoFunc)(
```
PingXie, line 856: the info buffer needs be allocated by the caller with size provided for future-proofing. see `ValkeyModule_GetClientInfoById`

src/valkeymodule.h
PingXie: > I think the more generic approach is to allow the scripting engine to return a C string with the error message.

Yeah if we would like to provide detailed compilation errors then we need to keep this capability but I would suggest `robj` too over C strings.

src/valkeymodule.h
rjd15372: I don't think it requires versioning. The version number in `ValkeyModuleScriptingEngineMethodsV1` should be enough. If we need to change the structure of `ValkeyModuleClientInfo`, we can bump the version of `VALKEYMODULE_SCRIPTING_ENGINE_ABI_VERSION`.

We're doing the same for the `ValkeyModuleScriptingEngineCompiledFunction` struct.

PingXie: sometimes we might just need a new field to the struct. bumping the ABI version would be more disruptive and might be breaking. If we version the struct, we can just add the new field to the end of the struct. The module who only understands V1 struct would still be able to parse the return info (all the way to the end of the V1 field list).

zuiderkwast: > sometimes we might just need a new field to the struct. bumping the ABI version would be more disruptive and might be breaking. If we version the struct, we can just add the new field to the end of the struct. The module who only understands V1 struct would still be able to parse the return info (all the way to the end of the V1 field list).

Yes, and we already version the struct ValkeyModuleScriptingEngineMethodsV1, which is the entry point for all scripting engine structs. If we add a field in any of these structs, we bump the version in the main struct ValkeyModuleScriptingEngineMethods.


@PingXie is there any benefit is versioning every single struct? And what do you mean by bumping the ABI version would be more disruptive? All we're talking about here is just versioned structs...

PR #1279, opened 2024-11-08
Set fields to NULL after free in freeClient()

QuChen88:
Null out several references after freeing the object in `freeClient()`.

This is just to make the code more safe, to protect against use-after-free for future changes.

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -1684,6 +1684,7 @@ void freeClient(client *c) {
     /* UNWATCH all the keys */
     unwatchAllKeys(c);
     listRelease(c->watched_keys);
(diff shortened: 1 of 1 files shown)


hpatro (approved):
Did it cause a double free or use-after-free somewhere we would need to fix?

QuChen88:
I didn't spot an issue in OSS. This change is just to enforce some best practices of setting the pointer references to NULL after freeing the object to avoid future mistakes like such. 

We did this change internally in our code base, so I figured it is better to open source it.

enjoy-binbin (approved):

zuiderkwast (approved):

PR #1281, opened 2024-11-08
Simplify dictType callbacks and move some macros from dict.h to dict.c

QuChen88:
Remove the dict pointer argument to the `dictType` callbacks `keyDup`, `keyCompare`, `keyDestructor` and `valDestructor`. This argument was unused in all of the callback implementations.

The macros `dictFreeKey()` and `dictFreeVal()` are made internal to dict and moved from dict.h to dict.c. They're also changed from macros to static inline functions.


src/config.c
```diff
@@ -1016,7 +1016,6 @@ void configGetCommand(client *c) {
 /* We use the following dictionary type to store where a configuration
  * option is mentioned in the old configuration file, so it's
  * like "maxmemory" -> list of line numbers (first line is zero). */
```
JimB123: Somehow this comment got moved from it's original target.  I think it belongs with `optionToLineDictType` (below)

src/dict.h
```diff
@@ -55,8 +55,8 @@ typedef struct dictType {
     uint64_t (*hashFunction)(const void *key);
     void *(*keyDup)(dict *d, const void *key);
     int (*keyCompare)(dict *d, const void *key1, const void *key2);
-    void (*keyDestructor)(dict *d, void *key);
-    void (*valDestructor)(dict *d, void *obj);
+    void (*keyDestructor)(struct dictType *type, void *key);
+    void (*valDestructor)(struct dictType *type, void *obj);
```
JimB123: I'm not sure why we even need a `dictType` passed to these routines?  The purpose of the callback is to dispose of the key or value.  So the callback function will be unique to the key or value - not related to the `dict` or `dictType`.

If there needs to be a difference in how the callback functions based on the `dictType`, then it should be a different callback - which is configured into the `dictType`.

This callback interface is C-style polymorphism.  With polymorphism, you don't usually have IF logic inside a function.  Instead, you have 2 functions.

QuChen88: You are right. `dictType` is not required in keyDestructor and valDestructor functions.

src/dict.h
```diff
@@ -55,8 +55,8 @@ typedef struct dictType {
     uint64_t (*hashFunction)(const void *key);
     void *(*keyDup)(dict *d, const void *key);
     int (*keyCompare)(dict *d, const void *key1, const void *key2);
-    void (*keyDestructor)(dict *d, void *key);
```
zuiderkwast, line 58: It's somewhat assymetric that keyDup is passed the dict and keyDestructor isn't.

I don't think any keyDup in our code base uses it and neither does keyCompare.

src/dict.c
```diff
@@ -742,12 +742,24 @@ dictEntry *dictUnlink(dict *d, const void *key) {
...
+}
+
 /* You need to call this function to really free the entry after a call
  * to dictUnlink(). It's safe to call this function with 'he' = NULL. */
 void dictFreeUnlinkedEntry(dict *d, dictEntry *he) {
     if (he == NULL) return;
-    dictFreeKey(d, he);
-    dictFreeVal(d, he);
+    dictFreeKey(d->type, he);
+    dictFreeVal(d->type, he);
```
zuiderkwast: These functions are called in three places. Taking `d->type` in all three places, before calling the function, looks like code duplication to me. It's better to pass `d` and let the receiving function do it.

zuiderkwast (commented):
Hi Q!

Removing unused code: good.

Moving these macros to inline functions inside the .c file: good.

Regarding:

> changes the function signature of those to not require dict and instead pass in dictType for better code maintenance.

Unimportant changes like this actually add some maintenance: Reviewing, and later handling the merge conflicts this change probably causes to other PRs, etc. Leaving old code as it is usually means less maintenance. It is an improvement of the code though, so let's merge it. :)

src/dict.h
zuiderkwast: Some callbacks are passed the dict `d` because they want to access the dict metadata inside the dict. Examples are dictRehashingStarted and dictRehashingComplete used by kvstore, which stores a pointer to the kvstore instance in the dict metadata.

So I can see why a destructor potentially could want to be passed a dict `d`, but I can't see why any of these callback would need the dictType. But I agree, the destructors in our code base don't need any of these, so let's just pass the object to be freed.

src/dict.h
QuChen88: Interesting. I can take a look into `keyDup()` and `keyCompare` to see if it can be simplified as well or not.

src/dict.c
QuChen88: This has led to some downstream complications in our internal code. Namely, we have use-cases when we unlink an item from the main dictionary and free it afterwards. When the actual free happens we can't assume the `dict` itself still exists as now we allocate dict on demand, so the `dict` itself can be freed if user performs flush operation. 

Conceptually it doesn't really make sense to require the `dict` itself to be present when we free the key and value.

QuChen88:
Addressed the comments. New revision is updated

src/dict.c
zuiderkwast: That's true, but we don't want to structure the code only to account for downstream patches. So far, none of the dict functions take a dictType, apart from dictCreate. We are making the API a bit messy by changing this.

We are using two-phase unlink in Valkey, for example in kvstore.c:

```C
dictEntry *
kvstoreDictTwoPhaseUnlinkFind(kvstore *kvs, int didx, const void *key, dictEntry ***plink, int *table_index) {
    dict *d = kvstoreGetDict(kvs, didx);
    if (!d) return NULL;
    return dictTwoPhaseUnlinkFind(kvstoreGetDict(kvs, didx), key, plink, table_index);
}

void kvstoreDictTwoPhaseUnlinkFree(kvstore *kvs, int didx, dictEntry *he, dictEntry **plink, int table_index) {
    dict *d = kvstoreGetDict(kvs, didx);
    dictTwoPhaseUnlinkFree(d, he, plink, table_index);
    cumulativeKeyCountAdd(kvs, didx, -1);
    freeDictIfNeeded(kvs, didx);
}
```

Here we delete dict only after it has been unlinked and also freed.

Can you explain what you do differently in your downstream patches and why? Is it for data tiering or something like that?

QuChen88: It is true that currently in OSS there is no use-case when you hold onto an entry that is unlinked from the main dict and later frees it. Internally we have such a use-case.

`dictFreeKey()` and `dictFreeVal()` are no longer API methods as defined in `dict.h` so this change should not be making the API messy. 

However, this reminds me that the API `dictFreeUnlinkedEntry()` in `dict.h` is still taking in `dict` as argument. It has no good reason to as the entry is already unlinked from the dict itself. I think maybe I should change it to 

```
void dictFreeUnlinkedEntry(dictType *type, dictEntry *he) {
    if (he == NULL) return;
    dictFreeKey(type, he); 
    dictFreeVal(type, he);
    /* Clear the dictEntry */
    if (!entryIsKey(he)) zfree(decodeMaskedPtr(he));
}
```

As you said, `dictCreate()` already takes in `dictType` as argument so I won't feel guilty for making this change. :)

zuiderkwast: > `dictFreeKey()` and `dictFreeVal()` are no longer API methods as defined in `dict.h` so this change should not be making the API messy.

True. We make these internal in dict.c now, so it's no disaster to change them to whatever we want. But there's also no reason to change them. Why would we change 1 to 2, and not all the way to 3 or 4?

1. `dictFreeVal(d, de);`
2. `dictFreeVal(d->type, de);`
3. `dictFreeVal(d->type->valDestructor, de);`
4. `if (d->type->valDestructor) d->type->valDestructor(dictGetVal(de))`;

Your downstream patches can actually use 4 directly.

`dictCompareKeys()` still is in the public API. It's used from memory prefetch only, right? I wouldn't mind we make `dictCompareKeys()` internal in dict.c and the memory prefetch can just use method 4 above, i.e. just call the `d->type->keyCompare` function directly. Or don't change it at all. :) Not changing is safe if you want to avoid a debate.

QuChen88: For now I will leave `dictCompareKeys()` in `dict.h` to avoid a debate. Don't want to keep blowing up the scope of this PR. 

I think it is a good idea to just pass in the `keyDestructor` into `dictFreeKey()`, and `valDestructor` into `dictFreeVal()`. Will explore that along with `void dictFreeUnlinkedEntry(dictType *type, dictEntry *he)`

zuiderkwast: To avoid the debate about dictCompareKeys, you'd need to change it back to what it was though.

src/dict.c
```diff
@@ -640,7 +640,7 @@ int dictReplace(dict *d, void *key, void *val) {
      * reverse. */
     void *oldval = dictGetVal(existing);
     dictSetVal(d, existing, val);
-    if (d->type->valDestructor) d->type->valDestructor(d, oldval);
+    if (d->type->valDestructor) d->type->valDestructor(oldval);
```
zuiderkwast, line 643: Here and in some other places, the callbacks are already called directly without using the helper functions. We could do that everywhere actually. No need for dictFreeVal.

src/dict.c
QuChen88: I feel we should not be invoking the low level `dict->type->valDestructor` directly in our internal higher level logic to free an unlinked dictEntry. Rather it should go through the higher level API `void dictFreeUnlinkedEntry(dictType *type, dictEntry *he)`. 

To avoid code duplications, I think I will keep `dictCompareKeys()` to be in `dict.h`. I am thinking of changing it from a macro into a `static inline int dictCompareKeys(dictType *type, const void *key1, const void *key2)`.

src/dict.c
QuChen88: This is the only place where it is directly invoking `valDestructor()` which is for the case when the old value needs to be freed. Also from a high level code API POV, this code along with `dictFreeVal()`, are purely internal implementations inside `dict.c` and has no effect to the API definitions that are exposed to the outside. Don't think this is a big issue.

zuiderkwast (approved):
OK, I'm happy. :)

We need to update the PR title and description before merging.

Issue #869, opened 2024-08-04
[BUG] Cluster: Inconsistent MOVED/READONLY errors for read-only client


gmbnomis:
**Describe the bug**

Issuing a write command in Valkey cluster (7.2.6 and 8.0) while in read-only mode (i.e. after a "READONLY" command) returns
different errors depending on whether the command is part of a script or not.

(I think this issue is not very significant, but it is an inconsistency)

**To reproduce**

Use the following commands on a cluster replica node. In this particular example, key "a" is served by a different node, key "b" is served by the primary of the current node:

```
127.0.0.1:30006> READONLY
OK
127.0.0.1:30006> SET a a
(error) MOVED 15495 127.0.0.1:30003
127.0.0.1:30006> SET b a
(error) MOVED 3300 127.0.0.1:30001
127.0.0.1:30006> EVAL "return redis.call('SET', KEYS[1], ARGV[1])" 1 a a
(error) MOVED 15495 127.0.0.1:30003
127.0.0.1:30006> EVAL "return redis.call('SET', KEYS[1], ARGV[1])" 1 b a
(error) READONLY You can't write against a read only replica. script: d8f2fad9f8e86a53d2a6ebd960b33c4972cacc37, on @user_script:1.
```

**Expected behavior**

I think the description of the "READONLY" command does not make fully clear what is supposed to happen if a client uses a writing command, but I suppose we should see "MOVED" in all four cases here.

zuiderkwast:
Makes sense. MOVED if all cases would be better.

I don't think this is a very severe bug, because clients are not expected to use READONLY with write commands. It is not clearly documented what will happen if they do. I don't think it needs to be a blocker for 8.0.0.

proost:
In my case, it works differently. I use Valkey 8.0.1.

`127.0.0.1:7004` has slots from 10923 to 16383

```
127.0.0.1:7004> READONLY
OK
127.0.0.1:7004> CLUSTER KEYSLOT a
(integer) 15495
127.0.0.1:7004> CLUSTER KEYSLOT b
(integer) 3300
127.0.0.1:7004> SET a a
-> Redirected to slot [15495] located at 127.0.0.1:7003
OK
127.0.0.1:7004> EVAL "return redis.call('SET', KEYS[1], ARGV[1])" 1 a a
(error) READONLY You can't write against a read only replica. script: d8f2fad9f8e86a53d2a6ebd960b33c4972cacc37, on @user_script:1.
[...]

PR #1284, opened 2024-11-09
fix: readonly client moved inconsistency

proost:
issue: https://github.com/valkey-io/valkey/issues/869

When client is read-only and client send lua script with write command to replica, return redirect response


src/script.c
```diff
@@ -373,6 +373,33 @@ static int scriptVerifyWriteCommandAllow(scriptRunCtx *run_ctx, char **err) {
...
+        int error_code;
+        int hashslot = -1;
+        clusterNode *n = getNodeByQuery(c, c->cmd, c->argv, c->argc, &hashslot, &error_code);
+        if (n == NULL || !clusterNodeIsMyself(n)) {
+            if (error_code == CLUSTER_REDIR_MOVED || error_code == CLUSTER_REDIR_ASK) {
+                int port = clusterNodeClientPort(n, connIsTLS(original_c->conn));
+                *err = sdscatprintf(sdsempty(), "-%s %d %s:%d", (error_code == CLUSTER_REDIR_ASK) ? "ASK" : "MOVED",
+                                    hashslot, clusterNodePreferredEndpoint(n, c), port);
+                return C_ERR;
+            }
```
hpatro: Could we use `clusterRedirectClient` method instead of handling the error manually?

proost: unforunately, we can't because after returned `scriptVerifyWriteCommandAllow` we should go to "error" label.
in the `error` label, handling reply.

gmbnomis:
@proost I just realized that we should not return `MOVED` in this case (sorry for the confusion), see https://github.com/valkey-io/valkey/issues/869#issuecomment-2555865260 for an explanation.

sarthakaggarwal97:
From the discussion it looks like it may not be needed anymore. I am closing the PR. Please feel free to open it if this is work in progress. Thank you.

PR #1285, opened 2024-11-09
Fix empty primary may have dirty slots data due to bad migration

enjoy-binbin:
If we become an empty primary for some reason, we still need to
check if we need to delete dirty slots, because we may have dirty
slots data left over from a bad migration. Like the target node
forcibly executes CLUSTER SETSLOT NODE to take over the slot without
performing key migration.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -2451,6 +2451,7 @@ void clusterUpdateSlotsConfigWith(clusterNode *sender, uint64_t senderConfigEpoc
      * need to delete all the keys in the slots we lost ownership. */
     uint16_t dirty_slots[CLUSTER_SLOTS];
     int dirty_slots_count = 0;
+    int delete_dirty_slots = 0;
 
     /* We should detect if sender is new primary of our shard.
      * We will know it if all our slots were migrated to sender, and sender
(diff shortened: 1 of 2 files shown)


src/cluster_legacy.c
```diff
@@ -2686,7 +2693,11 @@ void clusterUpdateSlotsConfigWith(clusterNode *sender, uint64_t senderConfigEpoc
          *
          * In order to maintain a consistent state between keys and slots
          * we need to remove all the keys from the slots we lost. */
-        for (int j = 0; j < dirty_slots_count; j++) {
+        delete_dirty_slots = 1;
+    }
+
+    if (delete_dirty_slots) {
+        for (j = 0; j < dirty_slots_count; j++) {
```
```suggestion
        for (int j = 0; j < dirty_slots_count; j++) {
```

src/cluster_legacy.c
```diff
@@ -6044,7 +6055,7 @@ void removeChannelsInSlot(unsigned int slot) {
 /* Remove all the keys in the specified hash slot.
  * The number of removed items is returned. */
 unsigned int delKeysInSlot(unsigned int hashslot) {
-    if (!kvstoreDictSize(server.db->keys, hashslot)) return 0;
+    if (!countKeysInSlot(hashslot)) return 0;
```
enjoy-binbin, line 6058: this is just a cleanup to use the helper function, we do use it in everywhere

zuiderkwast (approved):
Another corner case covered. Nice!

PR #1009, opened 2024-09-10
Fix replica not able to initate election in time when epoch fails

enjoy-binbin:
If multiple primary nodes go down at the same time, their replica nodes will
initiate the elections at the same time. There is a certain probability that
the replicas will initate the elections in the same epoch.

And obviously, in our current election mechanism, only one replica node can
eventually get the enough votes, and the other replica node will fail to win
due the the insufficient majority, and then its election will time out and
we will wait for the retry, which result in a long failure time.

If another node has been won the election in the failover epoch, we can assume
that my election has failed and we can retry as soom as possible.

PR #1288, opened 2024-11-11
Stabilize dual replication test to avoid getting LOADING error

enjoy-binbin:
When doing `$replica replicaof no one`, we may get a LOADING
error, this is because during the test execution, the replica
may reconnect very quickly, and the full sync is initiated,
and the replica has entered the LOADING state.

In this commit, we make sure the primary is pasued after the
fork, so the replica won't enter the LOADING state, and with
this fix, this test seems more natural and predictable.

### DIFF
--- a/tests/integration/dual-channel-replication.tcl
+++ b/tests/integration/dual-channel-replication.tcl
@@ -23,14 +23,20 @@ proc get_client_id_by_last_cmd {r cmd} {
     return $client_id
 }
 
-# Wait until the process enters a paused state, then resume the process.
-proc wait_and_resume_process idx {
+# Wait until the process enters a paused state.
+proc wait_process_pause idx {
     set pid [srv $idx pid]
     wait_for_condition 50 1000 {
         [string match "T*" [exec ps -o state= -p $pid]]
     } else {
         fail "Process $pid didn't stop, current state is [exec ps -o state= -p $pid]"
     }
+}
+
+# Wait until the process enters a paused state, then resume the process.
+proc wait_and_resume_process idx {
+    set pid [srv $idx pid]
+    wait_process_pause $idx
     resume_process $pid
 }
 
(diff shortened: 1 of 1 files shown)


enjoy-binbin:
Failed CI link: https://github.com/valkey-io/valkey/actions/runs/11756610760/job/32752932086?pr=1009#step:5:5820. Mentioned in here: https://github.com/valkey-io/valkey/pull/1009#issuecomment-2466458043

naglera:
LGTM. 
I previously believed that the `replicaof` command would be processed even during RDB loading. However, upon examining the code more closely, I see that it is actually disallowed when the loading process is asynchronous, such as in the case of dual-channel synchronization.

naglera (approved):

xbasel (approved):

tests/integration/dual-channel-replication.tcl
```diff
@@ -23,14 +23,20 @@ proc get_client_id_by_last_cmd {r cmd} {
     return $client_id
 }
 
-# Wait until the process enters a paused state, then resume the process.
-proc wait_and_resume_process idx {
+# Wait until the process enters a paused state.
+proc wait_process_pause idx {
```
zuiderkwast: Is "paused" more clear in this function name?

```suggestion
proc wait_process_paused idx {
```

zuiderkwast (commented):
Makes sense. It's a complicated test case, but it seems we need this pause to stabilize it.

tests/integration/dual-channel-replication.tcl
enjoy-binbin: yes, it is clearer

zuiderkwast (approved):

enjoy-binbin:
@naglera https://github.com/valkey-io/valkey/actions/runs/11788187416/job/32834861441#step:6:6097

Today we have a CI failure, it look like i introduced it.

The reason is that, since i added a pause, the replica will reconnect, and the rdb channel is connected. And then the rdb child is drop, and we will postpone the rdb client. 

And then in here, we will wait a client_output_buffer_limit_disconnections, and we got an unexpected rdb client, the previous one, so the test fail.
```
            wait_for_condition 500 1000 {
                [s -1 client_output_buffer_limit_disconnections] > $cur_client_closed_count
            } else {
                fail "Primary should disconnect replica due to COB overrun"
            }
```

i was wondering in here, in this case, should we still want to postpone the rdb client? In this case, it seems like we no longer need the rdb client since the rdb child is dead.
```diff
void updateReplicasWaitingBgsave(int bgsaveerr, int type) {
    listNode *ln;
    listIter li;

    /* Note: there's a chance we got here from within the REPLCONF ACK command
     * so we must avoid using freeClient, otherwise we'll crash on our way up. */

    listRewind(server.replicas, &li);
    while ((ln = listNext(&li))) {
        client *replica = ln->value;

        if (replica->repl_state == REPLICA_STATE_WAIT_BGSAVE_END) {
            int repldbfd;
            struct valkey_stat buf;

            if (bgsaveerr != C_OK) {
+                if (replica->flag.protected_rdb_channel) {
+                    replica->flag.protected_rdb_channel = 0;
+                    // removeReplicaFromPsyncWait(replica);
+                    serverLog(xxx)
+                }
                freeClientAsync(replica);
                serverLog(LL_WARNING, "SYNC failed. BGSAVE child returned an error");
                continue;
            }
```

i ran a daily test with --single and --loops 10, the test passed, so the fix seems solid. daily test: https://github.com/enjoy-binbin/valkey/actions/runs/11810068690

artikell:
@enjoy-binbin This problem occurs again, maybe you can take a look. 

https://github.com/artikell/valkey/actions/runs/14535230702/job/4078222933

PR #1292, opened 2024-11-12
Improve "SENTINEL FAILOVER" by using the "FAILOVER" command

### LINKED ISSUE #1291: [NEW] Improve "SENTINEL FAILOVER" by using the "FAILOVER" command

The "SENTINEL FAILOVER" command does not offer to failover in a coordinated way 
e.g. for maintenance. It just assumes that the current primary isn't reachable anymore,
easily leading to situations with multiple primaries and loss of writes.

Introduce a "coordinated" variant of the "SENTINEL FAILOVER" command that actually
takes the current primary into account during the failover.

Since version 6.2, Redis supports the "FAILOVER" command to
switch primary and replica roles in a coordinated fashion. However, one cannot use
this command in a sentinel setup (sentinel will usually failover again).

Additionally, the semantics of "FAILOVER" is different from the semantics defined
in the sentinel client protocol. In the latter, a client will be disconnected
when nodes switch roles. "FAILOVER" only disconnects connections with blocking commands.
(Interestingly, there are clients that disconnect if they find that a connection turns
read only (e.g. redis-py))

We could try to make "FAILOVER" work nevertheless: Below is a proposal for a 
`SENTINEL FAILOVER <primary name> COORDINATED` command that uses "FAILOVER" 
in a modified forced failover procedure. 

Adapt the "FAILOVER" command for this use case (e.g. by killing client connections 
after failover. But how do we keep up the connection that is used to control
the failover?)

**Additional information**

The proposed implementation is at #1292. We can keep the current failover state machine with the following
changes for a coordinated failover:

**SENTINEL_FAILOVER_STATE_NONE**

No change required

**SENTINEL_FAILOVER_STATE_WAIT_START**

The coordinated case asks for a leader election in contrast to the standard "SENTINEL FAILOVER"
case. We don't need to change the command used for voting, as it works irrespective of the
subjective up/down state of the primary (all sentinels except the one in charge of the failover
will typically regard the primaty as up).

Once the quorum is reached, the state machine will proceed to the next state.

The benefit of this approach is that other sentinels won't start a failover for the failover timeout period.
As a primary may seem unreachable during the failover procedure (because clients are blocked), other sentinels might initiate a failover during the coordinated failover otherwise. 

**SENTINEL_FAILOVER_STATE_SELECT_SLAVE**

No change required

**SENTINEL_FAILOVER_STATE_SEND_SLAVEOF_NOONE**

Instead of using "SLAVEOF NOONE", send (to the current primary):

```
MULTI
CLIENT PAUSE <timeout> WRITE
FAILOVER TO <promoted replica host> <promoted replica port> TIMEOUT <timeout>
EXEC
```

and progress to `SENTINEL_FAILOVER_STATE_WAIT_PROMOTION`.

Rationale for the CLIENT PAUSE:

As said above, according to the sentinel client protocol, the client shall
be disconnected when the primary role changes to replica. 

FAILOVER does not do that, it just pauses writing clients and unpauses them
once it gets the acknowledgement for the PSYNC FAILOVER.

Fortunately, a CLIENT PAUSE takes precedence over the client pause that is part
of FAILOVER. When the FAILOVER succeeds (in time), the clients remain paused and we 
can disconnect them in the next state.

**NB:** The CLIENT PAUSE is tricky, because we must avoid any write commands in
order not to become paused ourselves. INFO and PING are fine, but the PUBLISH for the
sentinel hello blocks. Additionally, other sentinels will begin to regard the primary as
not responding once they issue a PUBLISH.Thus, sentinels must not send PUBLISHEs while clients are 
blocked (i.e. to a primary in failover state).

**Note:** If the primary does not understand or does not accept the "FAILOVER" command (which
should not happen as the command exists since Redis 6.2), the failover will simply timeout,
because no role switch will happen.

**SENTINEL_FAILOVER_STATE_WAIT_PROMOTION**

No need to change the waiting part: Wait for the role switch in `sentinelRefreshInstanceInfo`.

However, when we detect the change, we need to take care of the clients now. Send:

```
MULTI
CONFIG REWRITE
CLIENT KILL TYPE normal
CLIENT KILL TYPE pubsub
CLIENT UNPAUSE
EXEC
```

To _both_ the former primary (now already replica) and the new primary (before we
call the client reconf script)

If no switch happens until the failover_timeout is reached, the sentinel failover
will be aborted. The FAILOVER command should already have timed out by this time.

If the sentinel crashes before we can issue the commands above, clients will remain
connected and become unpaused at some point in time. However, once the remaining sentinels
will notice that the former primary is not a primary anymore they will initiate a
failover to a replica. This will disconnect the clients from the former primary.

**SENTINEL_FAILOVER_STATE_RECONF_SLAVES**

No change (We can continue to ignore the former primary, as it became a replica
of the new primary by FAILOVER. In contrast to the current "SENTINEL FAILOVER", we
don't need to reconfigure this node later)

**SENTINEL_FAILOVER_STATE_UPDATE_CONFIG**

No change

**Handling of Valkey instances stuck in Failover**

The FAILOVER command may ["encounter some scenarios it can not automatically remediate from and may get stuck."](https://valkey.io/commands/failover/#failover-abort). 
The supervision of the sentinels can be used to handle these situations:
1. "REPLICAOF" is not accepted during a failover. Thus, send a "FAILOVER ABORT" before sending a "REPLICAOF"
   to a node in `sentinelSendReplicaOf()`. (If there is no ongoing failover, the error will just be ignored)
2. Sentinel monitors nodes for deviations from the expected state (wrong role or replication). Add a check
   for replicas that are in a failover state for too long and reconfigure them.

gmbnomis:
This PR implements #1291 and consists of the following commits:

Two commits to fix problems with the sentinel tests:

Commit: **Wait for all Sentinels to be connected before starting tests**

Up to now the sentinel test initialization verified that all sentinels detect each other.
However, detection does not imply connection, which led to intermittent failures in the
coordinated failover tests (no leader elected since disconnected sentinels do not take part in a
vote).

Fix this by waiting until no sentinel reports being "disconnected".

Commit: **sentinel-tests: Clean up config after config set tests**

Three commits with the actual implementation:

Commit: **Add option for coordinated failover to Sentinel** (coordinated failover without leader election)
Commit: **Allow Sentinel to recover from a stuck FAILOVER**
Commit: **SENTINEL FAILOVER COORDINATED actually does a leader election** (adds leader election before failover starts)


gmbnomis:
@enjoy-binbin I don't want to be pushy, but this PR is over a month old now. Could you have a look at the PR and/or #1291 when you find some time?

enjoy-binbin:
sorry, i dont have enough time to review this, i guess i will be free in the next month. @hwware wen, do you think you have the time to review this in a short time? since you know the sentinel better than me.

hwware:
Sure, let me review it. @enjoy-binbin

hwware:
> Reference i

@gmbnomis Sorry, let me review it ASAP.

hwware:
Could you please provide a CI result which include sentinel test suites, because there is no sentinel test in the commit CI? Thanks

src/commands/sentinel-failover.json
```diff
@@ -3,22 +3,35 @@
         "summary": "Forces a Sentinel failover.",
         "group": "sentinel",
         "since": "2.8.4",
-        "arity": 3,
+        "arity": -3,
         "container": "SENTINEL",
         "function": "sentinelCommand",
+        "history": [
+            [
+                "8.0.0",
```
hwware:
```suggestion
                "8.1.0",
```

src/commands/sentinel-failover.json
```diff
@@ -3,22 +3,35 @@
...
             {
                 "name": "primary-name",
                 "type": "string"
+            },
+            {
+                "token": "COORDINATED",
+                "name": "coordinated",
+                "type": "pure-token",
+                "optional": true,
+                "since": "8.0.0"
```
hwware:
```suggestion
                "since": "8.1.0"
```

src/sentinel.c
```diff
@@ -138,6 +139,14 @@ static mstime_t sentinel_default_failover_timeout = 60 * 3 * 1000;
 #define SENTINEL_SIMFAILURE_CRASH_AFTER_ELECTION (1 << 0)
 #define SENTINEL_SIMFAILURE_CRASH_AFTER_PROMOTION (1 << 1)
 
+/* The failover status of a monitored Valkey instance */
+#define SENTINEL_VALKEY_FAILOVER_NS 0 /* Not supported*/
+#define SENTINEL_VALKEY_NO_FAILOVER 1 /* Failover supported, no failover ongoing */
+#define SENTINEL_VALKEY_FAILOVER 2    /* Failover supported, failover ongoing */
```
hwware: The name with VALKEY is weird, it could have better name. Example is SENTINEL_MONITORED_INSTANCE_FAILOVER etc

src/sentinel.c
```diff
@@ -212,6 +221,9 @@ typedef struct sentinelValkeyInstance {
     mstime_t role_reported_time;
     mstime_t replica_conf_change_time; /* Last time replica primary addr changed. */
 
+    int valkey_failover_state;            /* Whether the Valkey instance supports/reports a failover in "master_failover_state" */
```
hwware: Comment as above

tests/sentinel/tests/18-stuck-failover.tcl
```diff
@@ -0,0 +1,53 @@
...
+    # Pick a replica
+    foreach_valkey_id id {
+        if {$id != $master_id} {
+            set failover_to_port [RPort $id]
+            set failover_to_pid [get_instance_attrib valkey $id pid]
+        }
+    }
+
+    wait_for_condition 1000 50 {
+        [string match "*flags master link-pending-commands*" [S 0 SENTINEL MASTER mymaster] ]
```
hwware:
```suggestion
        [string match "*flags master link-pending-commands*" [S 0 SENTINEL PRIMARY mymaster] ]
```
See if the updated command works

tests/sentinel/tests/16-config-set-config-get.tcl
```diff
@@ -56,3 +56,11 @@ test "SENTINEL CONFIG SET, wrong number of arguments" {
         fail "Expected to return Missing argument error"
     }
 }
+
+test "Undo SENTINEL CONFIG SET changes" {
+    foreach_sentinel_id id {
+        assert_equal {OK} [S $id SENTINEL CONFIG SET resolve-hostnames no announce-port 0]
+    }
+    assert_match {*no*0*} [S 1 SENTINEL CONFIG GET resolve-hostnames announce-port]
+    assert_match {announce-port 0} [S 1 SENTINEL CONFIG GET announce-port]
```
hwware: Why we need this test? and why we need Line 65 again to get announce-port second time?

gmbnomis:
> Could you please provide a CI result which include sentinel test suites, because there is no sentinel test in the commit CI? Thanks

[Here](https://github.com/gmbnomis/valkey/actions/runs/12439701547/job/34733941478#step:8:1) is a manual run of the Daily job test-ubuntu-jemalloc with sentinel tests enabled. Is this what you are looking for? If not, please let me know what I should run.

tests/sentinel/tests/16-config-set-config-get.tcl
gmbnomis: It turns out that the sentinel test framework does not run tests completely independent. Configuration changes like the ones tested in this test are persisted and used in the following tests. 

I considered three options:
1. Change the test framework: While this might be a good idea, I did not want to do this in this PR
2. Undo the config changes at the end of the "config-set-config-get" test.
3. Keep "config-set-config-get" at the end and insert the new tests "manual-coordinated" and "stuck-failover" in the middle, e.g. after "05-manual"

I took option 2 and that's why this code is here. Code lines 64 and 65 just reflect what's done in lines 7 and  8 of the file. You are right that line 65 is redundant in this case and can be deleted.

Alternatively, I could try option 3 and rename the test files.

src/commands/sentinel-failover.json
gmbnomis: changed

src/sentinel.c
gmbnomis: Yes, I did not like it much either. I tried to make it different from the SENTINEL_FAILOVER_... constants this way.

I changed it to the names you suggested.

src/sentinel.c
gmbnomis: changed accordingly

tests/sentinel/tests/18-stuck-failover.tcl
gmbnomis: Thanks, I missed that one. Changed.

tests/sentinel/tests/16-config-set-config-get.tcl
gmbnomis: I deleted line 65 for the time being.

src/commands/sentinel-failover.json
gmbnomis: changed

hwware (approved):

gmbnomis:
@hwware Thank you for your review! Given that this PR has been inactive for an additional month, could you please provide guidance on how we can move forward with getting it merged?

hwware:
Sorry, I am busying on other stuffs recently. Let me check this PR and https://github.com/valkey-io/valkey/pull/1964 together. I think they are great helpful for sentinel enhancement.

hwware:
Please update the json file (version to 9.0) then the maintainers have chance to merge this PR.  I think it already passes too long time, Sorry

gmbnomis:
> Please update the json file (version to 9.0) then the maintainers have chance to merge this PR

@hwware That's great news. I updated the version number and rebased onto the current `unstable`.

jdheyburn:
I've used for my sentinel failovers, and the speed of which failovers happen now is amazing!

One side effect I have found is that sentinel is often lagging behind in reconciling the new primary, and reports the incorrect primary until it has reconciled.

I raised this change to try to help lower it - do you think it makes sense @gmbnomis? 

- https://github.com/valkey-io/valkey/compare/unstable...jdheyburn:jdheyburn%2Fenhancement%2Flower-coordinated-failover-poll

PR #1293, opened 2024-11-12
Optimize PFCOUNT, PFMERGE command by SIMD acceleration

Nugine:
This PR optimizes the performance of HyperLogLog commands (PFCOUNT, PFMERGE) by adding AVX2 fast paths.

Two AVX2 functions are added for conversion between raw representation and dense representation. They are 15 ~ 30 times faster than scalar implementaion. Note that sparse representation is not accelerated.

AVX2 fast paths are enabled when the CPU supports AVX2 (checked at runtime) and the hyperloglog configuration is default (HLL_REGISTERS == 16384 && HLL_BITS == 6).

`PFDEBUG SIMD (ON|OFF)` subcommand is added for unit tests. A new TCL unit test checks that the results produced by non-AVX2 and AVX2 implementations are exactly equal.

When merging 3 dense hll structures, the benchmark shows a 12x speedup compared to the scalar version.

```
pfcount key1 key2 key3
pfmerge keyall key1 key2 key3
```

```
======================================================================================================
Type             Ops/sec    Avg. Latency     p50 Latency     p99 Latency   p99.9 Latency       KB/sec 
------------------------------------------------------------------------------------------------------
PFCOUNT-scalar    5665.56        35.29839        32.25500        63.99900        67.58300       608.60
PFCOUNT-avx2     72377.83         2.75834         2.67100         5.34300         6.81500      7774.96
------------------------------------------------------------------------------------------------------
PFMERGE-scalar    9851.29        20.28806        20.09500        36.86300        39.16700       615.71
PFMERGE-avx2    125621.89         1.59126         1.55100         3.11900         4.70300     15702.74
------------------------------------------------------------------------------------------------------

scalar: valkey:unstable  2df56d87c0ebe802f38e8922bb2ea1e4ca9cfa76
avx2:   Nugine:hll-simd  8f9adc34021080d96e60bd0abe06b043f3ed0275

CPU:    13th Gen Intel® Core™ i9-13900H × 20
Memory: 32.0 GiB
OS:     Ubuntu 22.04.5 LTS
```

Experiment repo: https://github.com/Nugine/redis-hyperloglog
Benchmark script: https://github.com/Nugine/redis-hyperloglog/blob/main/scripts/memtier.sh
Algorithm: https://github.com/Nugine/redis-hyperloglog/blob/main/cpp/bench.cpp

### DIFF
--- a/src/hyperloglog.c
+++ b/src/hyperloglog.c
@@ -1080,6 +1080,10 @@ int hllAdd(robj *o, unsigned char *ele, size_t elesize) {
  */
 ATTRIBUTE_TARGET_AVX2
 void hllMergeDenseAVX2(uint8_t *reg_raw, const uint8_t *reg_dense) {
+    /* Shuffle indices for unpacking bytes of dense registers
+     * From: {XXXX|AAAB|BBCC|CDDD|EEEF|FFGG|GHHH|XXXX}
+     * To:   {AAA0|BBB0|CCC0|DDD0|EEE0|FFF0|GGG0|HHH0}
+     */
     const __m256i shuffle = _mm256_setr_epi8( //
         4, 5, 6, -1,                          //
         7, 8, 9, -1,                          //
@@ -1251,6 +1255,10 @@ int hllMerge(uint8_t *max, robj *hll) {
  */
 ATTRIBUTE_TARGET_AVX2
 void hllDenseCompressAVX2(uint8_t *reg_dense, const uint8_t *reg_raw) {
+    /* Shuffle indices for packing bytes of dense registers
+     * From: {AAA0|BBB0|CCC0|DDD0|EEE0|FFF0|GGG0|HHH0}
+     * To:   {AAAB|BBCC|CDDD|0000|EEEF|FFGG|GHHH|0000}
+     */
     const __m256i shuffle = _mm256_setr_epi8( //
         0, 1, 2,                              //
         4, 5, 6,                              //
--- a/src/hyperloglog.c
+++ b/src/hyperloglog.c
@@ -212,8 +212,10 @@ struct hllhdr {
 
 static char *invalid_hll_err = "-INVALIDOBJ Corrupted HLL object detected";
 
+#ifdef HAVE_AVX2
 static int simd_enabled = 1;
 #define HLL_USE_AVX2 (simd_enabled && __builtin_cpu_supports("avx2"))
+#endif
 
 /* =========================== Low level bit macros ========================= */
 
@@ -1758,9 +1760,13 @@ void pfdebugCommand(client *c) {
         if (c->argc != 3) goto arityerr;
 
         if (!strcasecmp(c->argv[2]->ptr, "on")) {
+#ifdef HAVE_AVX2
             simd_enabled = 1;
+#endif
         } else if (!strcasecmp(c->argv[2]->ptr, "off")) {
+#ifdef HAVE_AVX2
             simd_enabled = 0;
+#endif
         } else {
             addReplyError(c, "Argument must be ON or OFF");
         }
--- a/src/hyperloglog.c
+++ b/src/hyperloglog.c
@@ -35,6 +35,10 @@
 #include <stdint.h>
 #include <math.h>
 
+#ifdef HAVE_AVX2
+#include <immintrin.h>
+#endif
+
 /* The HyperLogLog implementation is based on the following ideas:
  *
  * * The use of a 64 bit hash function as proposed in [1], in order to estimate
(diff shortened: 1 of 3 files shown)


src/hyperloglog.c
```diff
@@ -1064,6 +1068,132 @@ int hllAdd(robj *o, unsigned char *ele, size_t elesize) {
...
+ * 1) HLL_REGISTERS == 16384 && HLL_BITS == 6
+ * 2) The CPU supports AVX2 (checked at runtime in hllMergeDense)
+ *
+ * reg_raw: pointer to the raw representation array (16384 bytes, one byte per register)
+ * reg_dense: pointer to the dense representation array (12288 bytes, 6 bits per register)
+ */
+ATTRIBUTE_TARGET_AVX2
+void hllMergeDenseAVX2(uint8_t *reg_raw, const uint8_t *reg_dense) {
+    const __m256i shuffle = _mm256_setr_epi8( //
+        4, 5, 6, -1,                          //
```
xbasel, line 1095: Not sure I understand why the shuffle starts with bytes 4,5,6 ?

Nugine:
```
Before shuffle: {XXXX|AAAB|BBCC|CDDD|EEEF|FFGG|GHHH|XXXX}
After  shuffle: {AAA0|BBB0|CCC0|DDD0|EEE0|FFF0|GGG0|HHH0}
```

The `AAA` bytes is moved from position `4,5,6` to position `0,1,2`.
The shuffle indices of position `0,1,2,3` is `4,5,6,-1`. (`-1` indicates zero byte).

src/hyperloglog.c
```diff
@@ -1064,6 +1068,132 @@ int hllAdd(robj *o, unsigned char *ele, size_t elesize) {
...
+        a4 = _mm256_and_si256(x, _mm256_set1_epi32(0x00fc0000));
+
+        a2 = _mm256_slli_epi32(a2, 2);
+        a3 = _mm256_slli_epi32(a3, 4);
+        a4 = _mm256_slli_epi32(a4, 6);
+
+        __m256i y1, y2, y;
+        y1 = _mm256_or_si256(a1, a2);
+        y2 = _mm256_or_si256(a3, a4);
+        y = _mm256_or_si256(y1, y2);
```
xbasel, line 1165: would this yield the same result?
```
_m256i y;
y=_mm256_or_si256(a1, a2);
y=_mm256_or_si256(y, a3);
y=_mm256_or_si256(y, a4);

```
Do y1 and y2 serve a functional purpose, or are they used just for clarity in managing the bitwise operations?

Nugine: > would this yield the same result?

Yes. `y = a1 | a2 | a3 | a4;`

> Do y1 and y2 serve a functional purpose, or are they used just for clarity in managing the bitwise operations?

It's a personal style convenient for counting number of instructions and dependencies between operations.

src/hyperloglog.c
xbasel: Suggest adding a comment above `shuffle` variable  to clarify its role in packing and unpacking data between dense and raw formats.

xbasel:
How was this change tested, particularly to confirm that the non-AVX2 and AVX2 implementations produce the same results?

Nugine:
> How was this change tested, particularly to confirm that the non-AVX2 and AVX2 implementations produce the same results?

The algorithms are verified by comparing the results between scalar code and simd code with random input.
https://github.com/Nugine/redis-hyperloglog/blob/main/cpp/bench.cpp

src/hyperloglog.c
xbasel: Just to confirm my understanding: is the 4-byte padding at the beginning and end used to accommodate the "lost" 2 bits for each 6-bit register?

Nugine: > Just to confirm my understanding: is the 4-byte padding at the beginning and end used to accommodate the "lost" 2 bits for each 6-bit register?

Not exactly.  AVX2 must load 32 bytes at once. To process 32 dense registers (24 bytes) at once, there are 8 bytes that are useless. And due to the [16-bytes limit](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=mm256_shuffle&ig_expand=6006) of `vpshufb`, the best layout is 4+12+12+4.

Nugine: > Suggest adding a comment above `shuffle` variable to clarify its role in packing and unpacking data between dense and raw formats.

Changed in https://github.com/valkey-io/valkey/pull/1293/commits/a6136ca1dad60ae6779cb6bbfefc7fa4681894ed

xbasel (approved):

zuiderkwast:
This is so cool! :sunglasses:

> > How was this change tested, particularly to confirm that the non-AVX2 and AVX2 implementations produce the same results?
> 
> The algorithms are verified by comparing the results between scalar code and simd code with random input. https://github.com/Nugine/redis-hyperloglog/blob/main/cpp/bench.cpp

I think we need to test this in our repo in some way. The binary representation can't change, because things like replicas will not understand it, so we should verify the binary representation.

A hyperloglog key is actually a string so we can use the GET command to get the binary representation. In a TCL test case, we can use GET and compare the reply to the binary data we have stored earlier. Alternatively we can use DUMP. Can you add it?

zuiderkwast:
@lipzhu You're the performance expert. Do you want to review this?

src/hyperloglog.c
```diff
@@ -1363,12 +1607,16 @@ void pfmergeCommand(client *c) {
...
-        case HLL_SPARSE: hllSparseSet(o, j, max[j]); break;
+    if (use_dense) {
+        hllDenseCompress(hdr->registers, max);
+    } else {
+        for (j = 0; j < HLL_REGISTERS; j++) {
+            if (max[j] == 0) continue;
+            hdr = o->ptr;
+            switch (hdr->encoding) {
+            case HLL_DENSE: hllDenseSet(hdr->registers, j, max[j]); break;
+            case HLL_SPARSE: hllSparseSet(o, j, max[j]); break;
```
zhulipeng, line 1626: Nit: Missing indentation.

src/hyperloglog.c
```diff
@@ -1114,6 +1243,121 @@ int hllMerge(uint8_t *max, robj *hll) {
...
+     *
+     * SHUFFLE to:
+     * {AAAB|BBCC|CDDD|0000|EEEF|FFGG|GHHH|0000}
+     *
+     * STORE the lower half and higher half respectively:
+     * AAABBBCCCDDD0000
+     *             EEEFFFGGGHHH0000
+     * AAABBBCCCDDDEEEFFFGGGHHH0000
+     *
+     * Note that the last 4 bytes are padding bytes.
```
zhulipeng, line 1308: Will the last 4 bytes padding bytes impact the binary dump files which caused the inconsistent with origin one?

zhulipeng (commented):
@Nugine Great job, the performance is impressive. 

Echo @zuiderkwast @xbasel .
Maybe a unit test needed to verify the results is totally same W/ AVX2 instructions and make sure binary file will not changed when RDB presentence is enabled?

src/hyperloglog.c
Nugine: clang-format-18 removes the indentation 😂

src/hyperloglog.c
Nugine: The 4 padding bytes produced by AVX2 STORE are overwritten by the last scalar computation (32 registers).

The two AVX2 functions keep the memory format unchanged. They just accelerate the computation.

Nugine:
WIP: unit tests for verifying the results produced by non-AVX2 and AVX2 implementations.

I have other things to do these days. So it may take a while.

---

Edit: Done!

src/hyperloglog.c
zuiderkwast: Yes, `case` and labels for goto are indented one step less than other code. This is correct indentation, a common style in C.

tests/unit/hyperloglog.tcl
```diff
@@ -222,6 +222,46 @@ start_server {tags {"hll"}} {
...
+        r pfmerge hllsimd{t} hll1{t} hll2{t} hll3{t}
+
+        set scalar [r pfcount hllscalar{t}]
+        set simd [r pfcount hllsimd{t}]
+        assert {$scalar > 0}
+        assert {$simd > 0}
+        assert_equal $scalar $simd
+
+        set scalar [r pfdebug getreg hllscalar{t}]
+        set simd [r pfdebug getreg hllsimd{t}]
```
xbasel: is this as reliable as getting hllscalar/hllsimd as keys ( binary representation) ?

Nugine: Not very reliable. Changed to getting binary representation in https://github.com/valkey-io/valkey/pull/1293/commits/454dbfc03f2578a960d18d22dfa5718bc08d22a5

zhulipeng (approved):

zuiderkwast (approved):
Thanks @Nugine and also thanks for helping review @xbasel @lipzhu!

I don't fully understand the AVX code, but the test case is great and with the additional reviews I'm OK with merging this.

PR #336, opened 2024-04-19
Introduce `commandlog heavytraffic` to record big response packet.

CharlesChen888:
#### Why fat log?

We already have slowlog, which keeps a record of recent queries that take relatively long time to execute. However sometimes it is not enough for debugging.

For example, we often receive questions from users like: "Why is my QPS not high, but the network flow is substantial?" or "Why did the network flow suddenly increase?" The situation is network flow suddenly increases without significant higher QPS. 

The most common reason is that users are accessing big keys, like getting a long string, or using commands like keys, scan, hget, zrange to request too many elements at once. This is common among our users.

Here a new log is introduced to keep a record when a query triggers a large response (and we may call this log "fatlog"), in which the command, response size, time stamp, client name and client network address are logged. This can help to monitor traffic flow spikes and big keys. And it is a quick way to pinpoint the problem.

#### Alternative solution?

- Slowlog: the command with big response packet may not be slow enough to trigger slowlog.
- Big key: we can detect big keys, but we are not sure whether they are accessed at a certain time, and accessing a big key does not necessarily generate large responses.
- Client traffic statistics: we may need to locate the exact commands.

#### Detail introduction

Just like slowlog, fatlog is not actually logged in a file, and is accessible through a series of commands:
```
FATLOG GET [count]
FATLOG LEN
FATLOG RESET
```
```
127.0.0.1:6379> config set fatlog-log-bigger-than 10
OK
127.0.0.1:6379> set a 1234567890
OK
127.0.0.1:6379> get a
"1234567890"
127.0.0.1:6379> fatlog get
1) 1) (integer) 3
   2) (integer) 1713516667
   3) (integer) 17
   4) 1) "get"
      2) "a"
   5) "127.0.0.1:62969"
   6) ""
```
also a help command is provided:
```
FATLOG HELP
```
And just like slowlog, two config items are provided to set how many log items are preserved and the response size threshold of being logged.
```
fatlog-max-len (default value: 128)
fatlog-log-bigger-than (default value: 1MB)
```

#### About implementation

You will find that a lot changes have been appllied to slowlog.c. That is because the implementation of fatlog is basically reusing the code of slowlog, since they have similar format and are controled by similar commands. So what used to be slowlog.c now contains code of both slowlog and fatlog.

PR #1294, opened 2024-11-12
Feature COMMANDLOG to record slow execution and large request/reply

soloestoy:
As discussed in PR #336.

We have different types of resources like CPU, memory, network, etc. The `slowlog` can only record commands eat lots of CPU during the processing phase (doesn't include read/write network time), but can not record commands eat too many memory and network. For example:

1. run "SET key value(10 megabytes)" command would not be recored in slowlog, since when processing it the SET command only insert the value's pointer into db dict. But that command eats huge memory in query buffer and bandwidth from network. In this case, just 1000 tps can cause 10GB/s network flow.
2. run "GET key" command and the key's value length is 10 megabytes. The get command can eat huge memory in output buffer and bandwidth to network.

This PR introduces a new command `COMMANDLOG`, to log commands that consume significant network bandwidth, including both input and output. Users can retrieve the results using `COMMANDLOG get <count> large-request` and `COMMANDLOG get <count> large-reply`, all subcommands for `COMMANDLOG` are:

* `COMMANDLOG HELP`
* `COMMANDLOG GET <count> <slow|large-request|large-reply>`
* `COMMANDLOG LEN <slow|large-request|large-reply>`
* `COMMANDLOG RESET <slow|large-request|large-reply>`

And the slowlog is also incorporated into the commandlog.

For each of these three types, additional configs have been added for control:

* `commandlog-request-larger-than` and `commandlog-large-request-max-len` represent the threshold for large requests(the unit is Bytes) and the maximum number of commands that can be recorded.
* `commandlog-reply-larger-than` and `commandlog-large-reply-max-len` represent the threshold for large replies(the unit is Bytes) and the maximum number of commands that can be recorded.
* `commandlog-execution-slower-than` and `commandlog-slow-execution-max-len` represent the threshold for slow executions(the unit is microseconds) and the maximum number of commands that can be recorded.
    * Additionally, `slowlog-log-slower-than` and `slowlog-max-len` are now set as aliases for these two new configs.


hwware:
Just for Note: Let's first refresh the memory in pr https://github.com/valkey-io/valkey/pull/336, the last comment conclusion is:

After a core team meeting, we decided adding a new command COMMANDLOG with subcommands HEAVYTRAFFIC and SLOW, and then slowlog.c can be renamed to a common commandlog.c.  https://github.com/valkey-io/valkey/pull/336#issuecomment-2246809112

hwware:
From my understanding, for the command:  commandlog get should be:
COMMANDLOG get count slow | heavytraffic-input | heavytraffic-output.

commandlog len should be: 
COMMANDLOG len slow | heavytraffic-input | heavytraffic-output.

commandlog reset should be:
COMMANDLOG reset slow | heavytraffic-input | heavytraffic-output.

Can you describe the terms heavytraffic-input and heavytraffic-output in the json file (argument part) because I can only know
them from the source codes and valkey.conf so far? 


And I think the existing slowlog commands should be deprecated? If yes, I think you should update the related json files as well.

valkey.conf
```diff
@@ -1908,30 +1908,58 @@ aof-timestamp-enabled no
...
 # a value of zero forces the logging of every command.
 slowlog-log-slower-than 10000
-
 # There is no limit to this length. Just be aware that it will consume memory.
-# You can reclaim memory used by the slow log with SLOWLOG RESET.
+# You can reclaim memory used by the slow log with SLOWLOG RESET or COMMANDLOG RESET SLOW.
 slowlog-max-len 128
 
+# The second type is HEAVYTRAFFIC-INPUT, which is used to log commands with large inputs.
+# These commands can consume network bandwidth and the client's query buffer. The input
```
hwware:
```suggestion
# which can consume network bandwidth and the client's query buffer. The input
```

valkey.conf
```diff
@@ -1908,30 +1908,58 @@ aof-timestamp-enabled no
...
+# the overall command size to exceed the set threshold, it will be recorded.
+#
+# The threshold for command input size can be adjusted using the config below, with the
+# unit being bytes.
+heavytraffic-input-larger-than 1024
+# Record the number of commands.
+heavytraffic-input-max-len 128
+
+# The third type is HEAVYTRAFFIC-OUTPUT, which is used to log commands with large
+# execution results. This type consumes network bandwidth and the client's output buffer.
```
hwware:
```suggestion
# execution results. This kind of commands consume network bandwidth and the client's output buffer.
```

src/commandlog.c
```diff
@@ -0,0 +1,265 @@
...
+        commandlogGetReply(c, COMMANDLOG_TYPE_SLOW, count);
+    } else {
+        addReplySubcommandSyntaxError(c);
+    }
+}
+
+int commandlogGetTypeOrReply(client *c, robj *o) {
+    if (!strcasecmp(o->ptr, "slow")) return COMMANDLOG_TYPE_SLOW;
+    if (!strcasecmp(o->ptr, "heavytraffic-input")) return COMMANDLOG_TYPE_HEAVYTRAFFIC_INPUT;
+    if (!strcasecmp(o->ptr, "heavytraffic-output")) return COMMANDLOG_TYPE_HEAVYTRAFFIC_OUTPUT;
```
madolson:
```suggestion
    if (!strcasecmp(o->ptr, "large-request")) return COMMANDLOG_TYPE_HEAVYTRAFFIC_INPUT;
    if (!strcasecmp(o->ptr, "large-reply")) return COMMANDLOG_TYPE_HEAVYTRAFFIC_OUTPUT;
```

Alternative based on network in and out:
`large-input` and `large-output`.

We also discussed `large-response` as well, but reply was preferred since it seems more widely used.

hwware (approved):
LGTM, approved

src/config.c
```diff
@@ -3250,15 +3250,19 @@ standardConfig static_configs[] = {
 
     /* Unsigned Long configs */
     createULongConfig("active-defrag-max-scan-fields", NULL, MODIFIABLE_CONFIG, 1, LONG_MAX, server.active_defrag_max_scan_fields, 1000, INTEGER_CONFIG, NULL, NULL), /* Default: keys with more than 1000 fields will be processed separately */
-    createULongConfig("slowlog-max-len", NULL, MODIFIABLE_CONFIG, 0, LONG_MAX, server.slowlog_max_len, 128, INTEGER_CONFIG, NULL, NULL),
+    createULongConfig("slowlog-max-len", NULL, MODIFIABLE_CONFIG, 0, LONG_MAX, server.commandlog[COMMANDLOG_TYPE_SLOW].max_len, 128, INTEGER_CONFIG, NULL, NULL),
```
hpatro: As we are deprecating slowlog command, shall we create an alias for `slowlog-*` configs with corresponding `commandlog-*` ?

src/config.c
```diff
@@ -3250,15 +3250,19 @@ standardConfig static_configs[] = {
...
     createLongLongConfig("cluster-node-timeout", NULL, MODIFIABLE_CONFIG, 0, LLONG_MAX, server.cluster_node_timeout, 15000, INTEGER_CONFIG, NULL, NULL),
     createLongLongConfig("cluster-ping-interval", NULL, MODIFIABLE_CONFIG | HIDDEN_CONFIG, 0, LLONG_MAX, server.cluster_ping_interval, 0, INTEGER_CONFIG, NULL, NULL),
-    createLongLongConfig("slowlog-log-slower-than", NULL, MODIFIABLE_CONFIG, -1, LLONG_MAX, server.slowlog_log_slower_than, 10000, INTEGER_CONFIG, NULL, NULL),
+    createLongLongConfig("slowlog-log-slower-than", NULL, MODIFIABLE_CONFIG, -1, LLONG_MAX, server.commandlog[COMMANDLOG_TYPE_SLOW].threshold, 10000, INTEGER_CONFIG, NULL, NULL),
+    createLongLongConfig("large-request-larger-than", NULL, MODIFIABLE_CONFIG, -1, LLONG_MAX, server.commandlog[COMMANDLOG_TYPE_LARGE_REQUEST].threshold, 1024, INTEGER_CONFIG, NULL, NULL),
+    createLongLongConfig("large-reply-larger-than", NULL, MODIFIABLE_CONFIG, -1, LLONG_MAX, server.commandlog[COMMANDLOG_TYPE_LARGE_REPLY].threshold, 1024, INTEGER_CONFIG, NULL, NULL),
```
hpatro: Shall we add a prefix of `commandlog-` to these? Will be easier for user to find and update these together.

valkey.conf
```diff
@@ -1908,30 +1908,58 @@ aof-timestamp-enabled no
...
+# Which can consume network bandwidth and the client's query buffer. The request
+# size of a command includes the command itself and all its arguments. For example,
+# in the command SET KEY VALUE, if the key or value is relatively large and causes
+# the overall command size to exceed the set threshold, it will be recorded.
+#
+# The threshold for command request size can be adjusted using the config below, with the
+# unit being bytes.
+large-request-larger-than 1024
+# Record the number of commands.
+large-request-max-len 128
```
hpatro: Are we sure about the default value? Isn't it quite low?

src/server.c
```diff
@@ -3402,7 +3403,9 @@ void slowlogPushCurrentCommand(client *c, struct serverCommand *cmd, ustime_t du
      * use the original client to get the client information. */
     c = scriptIsRunning() ? scriptGetCaller() : c;
 
-    slowlogPushEntryIfNeeded(c, argv, argc, duration);
+    commandlogPushEntryIfNeeded(c, argv, argc, c->duration, COMMANDLOG_TYPE_SLOW);
+    commandlogPushEntryIfNeeded(c, argv, argc, c->net_input_bytes_curr_cmd, COMMANDLOG_TYPE_LARGE_REQUEST);
+    commandlogPushEntryIfNeeded(c, argv, argc, c->net_output_bytes_curr_cmd, COMMANDLOG_TYPE_LARGE_REPLY);
```
hpatro: Should we be maintaining three different list with unique log entries? Or should we have a single list/entry with multiple type associated with it.

src/commands/slowlog.json
```diff
@@ -4,6 +4,11 @@
         "complexity": "Depends on subcommand.",
         "group": "server",
         "since": "2.2.12",
-        "arity": -2
+        "arity": -2,
+        "deprecated_since": "8.1.0",
```
zuiderkwast, line 8: For compatibility with other databases, I think we should be extremely careful about deprecating commands. Some users are very careful to not use deprecated commands. I remember when some product team were working on upgrading to Valkey and reading all the release notes of the 7.x versions, they noticed that CLUSTER SLOTS is deprecated, it almost caused them a lot of extra work, until I explained to them that it is no longer deprecated.

We can document that it is an alias of COMMANDLOG, but we don't have to mark it as deprecated.

src/config.c
hwware: I suggest to not changing the existing config name. Leave it is fine

valkey.conf
hwware: you mean large-request-larger-than?

src/commands/slowlog.json
soloestoy: seems we need a "not-recommended" field.

src/server.c
soloestoy: Are you saying to put all three types of records into one list? I think they should be independent, otherwise, you would need to traverse the list to get the count of each specific type.

src/config.c
soloestoy: This is a good idea, the names could be like:
1. `commandlog-execution-slower-than` and `commandlog-slow-execution-max-len`
2. `commandlog-request-larger-than` and `commandlog-large-request-max-len`
3. `commandlog-reply-larger-than` and `commandlog-large-reply-max-len`

src/commands/slowlog.json
zuiderkwast: We can just describe it in the docs in text?

Just curious, why mark as not recommended? Is there a reason to make users move away from the slowlog commands?

src/server.c
hwware: I think 3 different lists are better because client maybe not care about all 3 entries.

src/commands/slowlog.json
hwware: From my understanding, the new command 'COMMANDLOG get count slow | heavytraffic-input | heavytraffic-output.' is encouraged in new version, thus the command 'Slowlog' is not suggested to be used anymore. 

I think it makes no sense that 2 commands with the same weight implement the same function in a system, they are duplicated.

zuiderkwast: OK, so there are pros and cons:

+ Avoid duplicate/alias commands. Of course this is good to avoid.
+ Keep compatibility with other systems (forks and managed databases) and tools written for them. This is a good thing too.

You say keep compatibility with other forks and tools makes no sense? Or you say we keep SLOWLOG for backward compatibility but it's recommended to use COMMANDLOG instead?

This is fine. I'm just concerned that if we say "deprecated", users will think that we want to remove it and diverge from redis and other forks. This is a bad communication to the community IMO.

I have seen some comments that people are worried that Valkey will diverge from open source Redis versions. If Redis and Valkey diverges, I hope we can safely say it is Redis fault, not our fault. Did you see the recent discussion about the `redis-rs` Rust library?

hwware: > OK, so there are pros and cons:
> 
> * Avoid duplicate/alias commands. Of course this is good to avoid.
> * Keep compatibility with other systems (forks and managed databases) and tools written for them. This is a good thing too.
> 
> You say keep compatibility with other forks and tools makes no sense? Or you say we keep SLOWLOG for backward compatibility but it's recommended to use COMMANDLOG instead?

  First all, I do not express making no sense to compatible with other forks and tools. I think Valkey should try the best compatible with others. And I would suggest in the new version, user had better choose the COMMANDLOG command instead of SLOWLOG command. 
> 
> This is fine. I'm just concerned that if we say "deprecated", users will think that we want to remove it and diverge from redis and other forks. This is a bad communication to the community IMO.

   Of course, from my personal experience, if I see API noted deprecated, I would like to avoid to use them. Maybe we can find the other words to encourage to use COMMANDLOG command
> 
> I have seen some comments that people are worried that Valkey will diverge from open source Redis versions. If Redis and Valkey diverges, I hope we can safely say it is Redis fault, not our fault. Did you see the recent discussion about the `redis-rs` Rust library?

  Yes, I just saw the lots of comments from redis-rs discussion yesterday. But for deprecating an existing command in Redis OSS, I think we could find a better solution to let our user know to deal with it.

zuiderkwast: > First all, I do not express making no sense to compatible with other forks and tools.

Sorry! I know you don't.

I think we want the same thing. I just wanted to highlight that there are pros and cons. Maybe I'm overrreacting. Maybe deprecated is right here.

Whatever we call it, deprecated or not recommended, we can describe in the text that there is no plan to remove the slowlog command. It can make worried people less worried maybe.

hwware: > > First all, I do not express making no sense to compatible with other forks and tools.
> 
> Sorry! I know you don't.  

Do not worry, I know you ^_^
> 
> I think we want the same thing. I just wanted to highlight that there are pros and cons. Maybe I'm overrreacting. Maybe deprecated is right here.
> 
> Whatever we call it, deprecated or not recommended, we can describe in the text that there is no plan to remove the slowlog command. It can make worried people less worried maybe.

In fact, 'not recommended' is much properly than deprecated if we can add it in json file. @soloestoy  can we?  haha, 
I never did it before

zuiderkwast: Ok, "not-recommended" or "not-recommended-since" field is not bad, haha. We can keep replaced-by too.

It has no effect in any tools, but we can use it for the website. We just need to implement some code for it in website repo.

soloestoy: I'm ok with "not-recommended", I think we can open another PR to implement and discuss which commands should be changed from "deprecated" to "no-recommended".

valkey.conf
hpatro: Is 1KB a good default for both input/output? Lot's of command response can easily exceed 1KB response I believe.

src/server.c
hpatro: @soloestoy If it's already implemented in Alibaba, do you have any anecdote from production where the same command showing up in slowlog/large-request/large-reply? My intuition is there will be quite a bit of overlap. Hence, I was thinking if we should avoid creating unique log entries for each commandlog type.

src/config.c
hpatro: My suggestion was to create alias which I see @soloestoy has added now.

valkey.conf
soloestoy: we can discuss it further, my thought is to work backward from the results. In my view, 100MB/s (nearly 1Gbps) is a large amount of traffic. With this level of traffic, I believe the ops that valkey can provide are on the order of 100,000. Therefore, calculating it backward, a default value of 10KB seems more reasonable.

src/server.c
soloestoy: In my opinion, with the current implementation of valkey, there is no direct relationship between large requests/responses and slow execution, because network transmission is not included in the execution time. They need to be measured and viewed independently.

For example, the execution time of command `set key value(1MB)` might be only 1 microsecond, because it only needs to insert the pointer of this 1MB value into the database. The network read, memory allocation, and data copying for this 1MB are not counted in the execution time.

soloestoy: Command `get key` is different, since the execution time includes copying value to output buffer.

hpatro: That's correct. I was thinking about the scan type operation. They would ideally take longer cpu cycles as well as the reply size would be large as well.

valkey.conf
```diff
@@ -1915,29 +1915,57 @@ aof-timestamp-enabled no
...
 # slow log. When a new command is logged the oldest one is removed from the
 # queue of logged commands.
-
+#
 # The following time is expressed in microseconds, so 1000000 is equivalent
 # to one second. Note that a negative number disables the slow log, while
 # a value of zero forces the logging of every command.
-slowlog-log-slower-than 10000
-
+commandlog-execution-slower-than 10000
```
zuiderkwast, line 2011: @soloestoy Let's mention the configs in the top comment.

I see here not only slowlog command is now an alias but also the configs. Shall we mention the slowlog configs in valkey.conf and describe they are now aliases for the new configs?

When I see also configs are changed into aliases, it makes me think deprecated is maybe the right term. But we can mention that the slowlog configs still work and we can list them here too. We can also say that there is no plan to remove them?

zuiderkwast:
@valkey-io/core-team Let's make a TSC decision on the command syntax and configs as described in the top comment.

Approve or vote 👍 or 👎.

src/commandlog.c
```diff
@@ -0,0 +1,265 @@
...
+ * IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE
+ * ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT OWNER OR CONTRIBUTORS BE
+ * LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR
+ * CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF
+ * SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS
+ * INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN
+ * CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE)
+ * ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
+ * POSSIBILITY OF SUCH DAMAGE.
+ */
```
madolson:
```suggestion
/*
 * Copyright Valkey Contributors.
 * All rights reserved.
 * SPDX-License-Identifier: BSD 3-Clause
 */
```
Use SPDX as well, I think this file has been materially changed.

src/server.c
madolson: Independent lists make the most sense to me.

valkey.conf
```diff
@@ -1915,29 +1915,57 @@ aof-timestamp-enabled no
...
+commandlog-large-request-max-len 128
+
+# The third type is LARGE_REPLY, which is used to log commands with large replies.
+# This kind of commands consumes network bandwidth and the client's output buffer.
+# Examples include commands that retrieve a large amount of data, such as KEYS or HGETALL.
+# Even the GET command can produce a large reply if the value is substantial.
+#
+# The threshold for the size of command reply (in bytes) and the number of commands
+# that can be logged.
+commandlog-reply-larger-than 1024
```
madolson: This feels small to me. We did some analysis, and this is within the P99 of string objects we saw, which we expect people to often need to fetch entire values out of. I was expecting this to be more like 1MB.

src/server.h
```diff
@@ -375,6 +375,19 @@ typedef enum blocking_type {
                                     buffer configuration. Just the first  \
                                     three: normal, replica, pubsub. */
 
+#define COMMANDLOG_TYPE_SLOW 0
+#define COMMANDLOG_TYPE_LARGE_REQUEST 1
+#define COMMANDLOG_TYPE_LARGE_REPLY 2
+#define COMMANDLOG_TYPE_MAX 3
```
madolson: This would be better as an enum, so we get type checking.

tests/unit/commandlog.tcl
```diff
@@ -0,0 +1,361 @@
...
+        set slowlog_resp [r commandlog get -1 slow]
+
+        # Make sure all 3 commands were logged, but the sensitive fields are omitted
+        assert_equal 4 [llength $slowlog_resp]
+        assert_match {* key 9 5000} [lindex [lindex $slowlog_resp 2] 3]
+        assert_match {* key 9 5000 AUTH (redacted)} [lindex [lindex $slowlog_resp 1] 3]
+        assert_match {* key 9 5000 AUTH2 (redacted) (redacted)} [lindex [lindex $slowlog_resp 0] 3]
+    } {} {needs:repl}
+
+    test {COMMANDLOG slow - Rewritten commands are logged as their original command} {
```
madolson, line 183: This is effectively a duplicate of the version in slowlog.tcl, can we delete that other file and just make this test for both command log and slow log?

src/commandlog.c
```diff
@@ -0,0 +1,265 @@
...
+ * queries that took more than M microseconds to execute, or consumed
+ * too much network bandwidth and memory for input/output buffers.
+ *
+ * The execution time to reach to be logged in the slow log is set
+ * using the 'slowlog-log-slower-than' config directive, that is also
+ * readable and writable using the CONFIG SET/GET command.
+ *
+ * Other configurations such as `large-request-larger-than` and
+ * `large-reply-larger-than` can be found with more detailed
+ * explanations in the config file.
```
madolson:
```suggestion
 * Other configurations such as `commandlog-request-larger-than` and
 * `commandlog-reply-larger-than` can be found with more detailed
 * explanations in the config file.
```

valkey.conf
```diff
@@ -1915,29 +1915,57 @@ aof-timestamp-enabled no
...
+# in the command SET KEY VALUE, if the key or value is relatively large and causes
+# the overall command size to exceed the set threshold, it will be recorded.
+#
+# The threshold for command request size can be adjusted using the config below, with the
+# unit being bytes.
+commandlog-request-larger-than 1024
+# Record the number of commands.
+commandlog-large-request-max-len 128
+
+# The third type is LARGE_REPLY, which is used to log commands with large replies.
```
```suggestion
# The third type is LARGE-REPLY, which is used to log commands with large replies.
```
For consistency with the types used in the commands.

madolson (commented):
Mostly LGTM, still agree with high level direction.

valkey.conf
soloestoy: for both request and reply? or just reply

tests/unit/commandlog.tcl
soloestoy: in theory we can delete slowlog.tcl, but i'd like to keep it in case something different between slowlog and commandlog.

valkey.conf
madolson: Just reply. I don't have any data about requests. I would be okay having them both be at 1MB for parity, but I really don't know what a sensible value is for large requests.

enjoy-binbin (approved):
overall LGTM, didn't review the code carefully in depth though.

valkey.conf
zuiderkwast: Request 1KB and reply 1MB seems like very big difference. :) Maybe use 1MB for both? It seems less random.

A use case is to store a large string like a binary image or some serialized data like JSON or XML. In this case, GET and SET use the same size...

soloestoy: >A use case is to store a large string like a binary image or some serialized data like JSON or XML. In this case, GET and SET use the same size...

This is a compelling case.

Another scenario I also thought of is when the reply of a command like `HGETALL key` is 1MB, which is a large reply that we are concerned with, but we don't care about 1,000 `HSET key field value` commands that are each 1KB. From this perspective, it is also reasonable for the request to default to 1MB.

src/commandlog.c
```diff
@@ -0,0 +1,242 @@
...
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD 3-Clause
+ */
+
+#include "commandlog.h"
+
+/* Create a new commandlog entry.
+ * Incrementing the ref count of all the objects retained is up to
+ * this function. */
+commandlogEntry *commandlogCreateEntry(client *c, robj **argv, int argc, long long value, int type) {
```
PingXie: nit - use the `static` marker for internal functions.
```suggestion
static commandlogEntry *commandlogCreateEntry(client *c, robj **argv, int argc, long long value, int type) {
```

src/server.c
```diff
@@ -3476,10 +3476,11 @@ void preventCommandReplication(client *c) {
     c->flag.prevent_repl_prop = 1;
 }
 
-/* Log the last command a client executed into the slowlog. */
-void slowlogPushCurrentCommand(client *c, struct serverCommand *cmd, ustime_t duration) {
-    /* Some commands may contain sensitive data that should not be available in the slowlog. */
-    if (cmd->flags & CMD_SKIP_SLOWLOG) return;
+/* Log the last command a client executed into the commandlog. */
+void commandlogPushCurrentCommand(client *c, struct serverCommand *cmd) {
```
PingXie: nit - this looks like belonging to commandlog.c to me

src/commandlog.c
```diff
@@ -0,0 +1,242 @@
...
+    commandlogEntry *ce = zmalloc(sizeof(*ce));
+    int j, slargc = argc;
```
PingXie: nit - `sl` --> `cl`?
```suggestion
    int j, clargc = argc;
```

src/commandlog.c
```diff
@@ -0,0 +1,242 @@
...
+        server.commandlog[i].entries = listCreate();
+        server.commandlog[i].entry_id = 0;
+        listSetFreeMethod(server.commandlog[i].entries, commandlogFreeEntry);
+    }
+}
+
+/* Push a new entry into the command log.
+ * This function will make sure to trim the command log accordingly to the
+ * configured max length. */
+void commandlogPushEntryIfNeeded(client *c, robj **argv, int argc, long long value, int type) {
```
PingXie: once `commandlogPushCurrentCommand` is moved into this file, I think it would be beneficial to inline `commandlogPushEntryIfNeeded` to help reduce the likelihood of branching since we are now testing 3 different thresholds.

src/server.h
```diff
@@ -1688,10 +1704,7 @@ struct valkeyServer {
     long long stat_sync_full;                      /* Number of full resyncs with replicas. */
     long long stat_sync_partial_ok;                /* Number of accepted PSYNC requests. */
     long long stat_sync_partial_err;               /* Number of unaccepted PSYNC requests. */
-    list *slowlog;                                 /* SLOWLOG list of commands */
-    long long slowlog_entry_id;                    /* SLOWLOG current entry ID */
-    long long slowlog_log_slower_than;             /* SLOWLOG time limit (to get logged) */
-    unsigned long slowlog_max_len;                 /* SLOWLOG max number of items logged */
+    commandlog commandlog[COMMANDLOG_TYPE_NUM];    /* Logs of commands. */
```
PingXie, line 1707: nit - we could also consider moving `commandlog` out of this gigantic `valkeyServer` struct and into a top level global in commandlog.h. I feel it is time to start trimming`valkeyServer`.

valkey.conf
```diff
@@ -1969,29 +1969,57 @@ aof-timestamp-enabled no
...
+# Currently, there are three types of commandlogs supported: those used to record
+# commands with long execution times, those with large requests, and those with
+# large query replies.
+
+# The first type is SLOW, log queries that exceeded a specified execution time.
+# The execution time does not include the I/O operations like talking with the client,
+# sending the reply and so forth, but just the time needed to actually execute the
+# command (this is the only stage of command execution where the thread is blocked
+# and can not serve other requests in the meantime).
+#
```
PingXie: nit - I don't think there is an intrinsic order among these categories so how about the following?

Also I assume we will support `RESET` of all 3 types.

```suggestion
#
# Currently, three types of command logs are supported:
#
# SLOW: Logs commands that exceed a specified execution time. This excludes time spent 
# on I/O operations like client communication and focuses solely on the command's 
# processing time, where the main thread is blocked.
#
# LARGE-REQUEST: Logs commands with requests exceeding a defined size. This helps 
# identify potentially problematic commands that send excessive data to the server.
#
# LARGE-REPLY: Logs commands that generate replies exceeding a defined size. This 
# helps identify commands that return unusually large amounts of data, which may 
# impact network performance or client processing.
#
# Each log type has two key parameters:
# 1. A threshold value that determines when a command is logged. This threshold is specific 
#    to the type of log (e.g., execution time, request size, or reply size). A negative value disables
#    logging. A value of 0 logs all commands.
# 2. A maximum length that specifies the number of entries to retain in the log. Increasing 
#    the length allows more entries to be stored but consumes additional memory. To clear all
#    entries for a specific log type and reclaim memory, use the `COMMANDLOG RESET` 
#    subcommand followed by the log type.
#
```

valkey.conf
```diff
@@ -1969,29 +1969,57 @@ aof-timestamp-enabled no
...
+# You can configure the slow threshold with two parameters: one tells the server
 # what is the execution time, in microseconds, to exceed in order for the
 # command to get logged, and the other parameter is the length of the
 # slow log. When a new command is logged the oldest one is removed from the
 # queue of logged commands.
-
+#
 # The following time is expressed in microseconds, so 1000000 is equivalent
 # to one second. Note that a negative number disables the slow log, while
 # a value of zero forces the logging of every command.
```
PingXie:
```suggestion
# SLOW Command Logs
# The SLOW log records commands that exceed a specified execution time. The execution time 
# does not include I/O operations, such as client communication or sending responses. 
# It only measures the time spent executing the command, during which the thread is blocked 
# and cannot handle other requests.
#
# The threshold is measured in microseconds.
#
# Backward Compatibility: The parameter `slowlog-log-slower-than` is still supported but 
# deprecated in favor of `commandlog-slow-execution`.
```

valkey.conf
```diff
@@ -1969,29 +1969,57 @@ aof-timestamp-enabled no
...
+commandlog-slow-execution-max-len 128
+
+# The second type is LARGE-REQUEST, which is used to log commands with large requests.
+# Which can consume network bandwidth and the client's query buffer. The request
+# size of a command includes the command itself and all its arguments. For example,
+# in the command SET KEY VALUE, if the key or value is relatively large and causes
+# the overall command size to exceed the set threshold, it will be recorded.
+#
+# The threshold for command request size can be adjusted using the config below, with the
+# unit being bytes.
```
PingXie:
```suggestion
# LARGE_REQUEST Command Logs
# The LARGE_REQUEST log tracks commands with requests exceeding a specified size. The request size 
# includes the command itself and all its arguments. For example, in `SET KEY VALUE`, the size is 
# determined by the combined size of the key and value. Commands that consume excessive network 
# bandwidth or query buffer space are recorded here.

# The threshold is measured in bytes.
```

valkey.conf
```diff
@@ -1969,29 +1969,57 @@ aof-timestamp-enabled no
...
+# Record the number of commands.
+commandlog-large-request-max-len 128
+
+# The third type is LARGE-REPLY, which is used to log commands with large replies.
+# This kind of commands consumes network bandwidth and the client's output buffer.
+# Examples include commands that retrieve a large amount of data, such as KEYS or HGETALL.
+# Even the GET command can produce a large reply if the value is substantial.
+#
+# The threshold for the size of command reply (in bytes) and the number of commands
+# that can be logged.
```
PingXie:
```suggestion
# LARGE_REPLY Command Logs
# The LARGE_REPLY log records commands that produce replies exceeding a specified size. These replies 
# may consume significant network bandwidth or client output buffer space. Examples include commands 
# like `KEYS` or `HGETALL` that return large datasets. Even a `GET` command may qualify if the value 
# is substantial.

# The threshold is measured in bytes. 
```

src/commandlog.c
```diff
@@ -0,0 +1,242 @@
...
+            "    client name",
+            "LEN",
+            "    Return the length of the slowlog.",
+            "RESET",
+            "    Reset the slowlog.",
+            NULL,
+        };
+        addReplyHelp(c, help);
+    } else if (c->argc == 2 && !strcasecmp(c->argv[1]->ptr, "reset")) {
+        commandlogReset(COMMANDLOG_TYPE_SLOW);
```
PingXie, line 182: only the `SLOW` list can be reset? Why are the others not supported?

src/commandlog.c
madolson: I think we should trust the compiler to do the inlining. As long we have o3 enabled it may choose to inline functions, we don't need to mark it explicitly.

src/commandlog.c
soloestoy: because this is only for the old `slowlog`

src/server.h
soloestoy: i'd like to keep it in `valkeyServer`, it's easy to let other system to use it, and the prefix `server.` can tell us it is global

valkey.conf
soloestoy: github says: "Applying suggestions on deleted lines is currently not supported."

src/commandlog.c
PingXie: ah, thanks!

src/commands/slowlog.json
PingXie: I think we should just leave `slowlog` alone. I don't see any downside. The pros mentioned above (such as avoiding alias) is only relevant to the Valkey developers. I would vote for just documenting that `slowlog` is `commandlog slow`. `cluster slots` before the caching fix would've been a better candidate for "not-recommended".

Separate PR to address this issue is fine.

PingXie (approved):
Thanks @soloestoy!

src/server.h
PingXie: not a blocker for me but I think there are options to achieve both: keeping the `valkeyServer` size manageable while at the same time making the global variables/states easily identifiable, such as prefixing them with `g_` for global globals and `s_` for static/local globals.

soloestoy:
Merged! Thanks every one, the next work is adding "not-recommended" for command's spec.

ranshid:
@soloestoy do you plan to add documentation rtelated to this commandlog?

PR #1295, opened 2024-11-12
Brocast a PONG to all node in cluster when role changed

enjoy-binbin:
When a node role changes, we should brocast the change to notify other nodes.
For example, one primary and one replica, after a failover, the replica
became a new primary, the primary became a new replica.

And then we trigger a second cluster failover for the new replica, the
new replica will send a MFSTART to its primary, ie, the new primary.

But the new primary may reject the MFSTART due to this logic:
```
    } else if (type == CLUSTERMSG_TYPE_MFSTART) {
        /* This message is acceptable only if I'm a primary and the sender
         * is one of my replicas. */
        if (!sender || sender->replicaof != myself) return 1;
```

In the new primary views, sender is still a primary, and sender->replicaof
is NULL, so we will return. Then the manual failover timedout.

Another possibility is that other primaries refuse to vote after receiving
the FAILOVER_AUTH_REQUEST, since in their's views, sender is still a primary,
so it refuse to vote, and then manual failover timedout.
```
void clusterSendFailoverAuthIfNeeded(clusterNode *node, clusterMsg *request) {
    ...
        if (clusterNodeIsPrimary(node)) {
            serverLog(LL_WARNING, "Failover auth denied to %.40s (%s) for epoch %llu: it is a primary node", node->name,
                      node->human_nodename, (unsigned long long)requestCurrentEpoch);
```

The reason is that, currently, we only update the node->replicaof information
when we receive a PING/PONG from the sender. For details, see clusterProcessPacket.
Therefore, in some scenarios, such as clusters with many nodes and a large
cluster-ping-interval (that is, cluster-node-timeout), the role change of the node
will be very delayed.

Added a DEBUG DISABLE-CLUSTER-RANDOM-PING command, send cluster ping to a random 
node every second (see clusterCron).


src/debug.c
```diff
@@ -436,6 +436,8 @@ void debugCommand(client *c) {
             "CLOSE-CLUSTER-LINK-ON-PACKET-DROP <0|1>",
             "    This is valid only when DROP-CLUSTER-PACKET-FILTER is set to a valid packet type.",
             "    When set to 1, the cluster link is closed after dropping a packet based on the filter.",
+            "CLUSTER-RANDOM-MIN-PING <0|1>",
+            "    Enable or disable the cluster random min ping.",
```
hwware: Enable is 0 and disable is 1? Or reverse?

src/server.h
```diff
@@ -2189,6 +2189,7 @@ struct valkeyServer {
     int cluster_slot_stats_enabled;                        /* Cluster slot usage statistics tracking enabled. */
     /* Debug config that goes along with cluster_drop_packet_filter. When set, the link is closed on packet drop. */
     uint32_t debug_cluster_close_link_on_packet_drop : 1;
+    uint32_t debug_cluster_random_min_ping : 1;
```
hwware: What does this mean of min-ping?

src/debug.c
enjoy-binbin: 1 is enable and 0 is disable. My wording is a bit bad.

src/server.h
enjoy-binbin: it is for this random ping logic, i can not think of a better word.
```
    /* Ping some random node 1 time every 10 iterations, so that we usually ping
     * one random node every second. */
    if (server.debug_cluster_random_min_ping && !(iteration % 10)) {
        int j;

        /* Check a few random nodes and ping the one with the oldest
         * pong_received time. */
        for (j = 0; j < 5; j++) {
            de = dictGetRandomKey(server.cluster->nodes);
            clusterNode *this = dictGetVal(de);

            /* Don't ping nodes disconnected or with a ping currently active. */
            if (this->link == NULL || this->ping_sent != 0) continue;
            if (this->flags & (CLUSTER_NODE_MYSELF | CLUSTER_NODE_HANDSHAKE)) continue;
            if (min_pong_node == NULL || min_pong > this->pong_received) {
                min_pong_node = this;
                min_pong = this->pong_received;
            }
        }
        if (min_pong_node) {
            serverLog(LL_DEBUG, "Pinging node %.40s", min_pong_node->name);
            clusterSendPing(min_pong_node->link, CLUSTERMSG_TYPE_PING);
        }
    }
```

zuiderkwast: So it means *ping one random node every second*?

enjoy-binbin: yes.

zuiderkwast: So why "min"? Is it a minimum?

> it is for this random ping logic, i can not think of a better word.

I think you said a better word yourself: "random ping". Or another variant: "ping random node".

```suggestion
    uint32_t debug_cluster_random_ping : 1;
```

src/debug.c
zuiderkwast: I suggest a more little bit more detailed explanation. It also helps to say that it is enabled by default. All the other ones are off by default.

```suggestion
            "    Send cluster ping to a random node every second. Enabled by default.",
```

src/server.h
enjoy-binbin: ok, good, i mentioned min earlier probably because this min_pong_node naming.

```
            if (min_pong_node == NULL || min_pong > this->pong_received) {
                min_pong_node = this;
                min_pong = this->pong_received;
            }
```

zuiderkwast: Ah, min_pong_node means minimum pong_received timestamp, the node we didn't get a pong from for a long time. I understand now, but I think "min" is not important for the name of the debug flag.

src/debug.c
```diff
@@ -607,6 +609,9 @@ void debugCommand(client *c) {
     } else if (!strcasecmp(c->argv[1]->ptr, "close-cluster-link-on-packet-drop") && c->argc == 3) {
         server.debug_cluster_close_link_on_packet_drop = atoi(c->argv[2]->ptr);
         addReply(c, shared.ok);
+    } else if (!strcasecmp(c->argv[1]->ptr, "cluster-random-min-ping") && c->argc == 3) {
```
zuiderkwast: Let's remove "min" from the name?


```suggestion
    } else if (!strcasecmp(c->argv[1]->ptr, "cluster-random-ping") && c->argc == 3) {
```

enjoy-binbin: yes, i forgot this one.

zuiderkwast (approved):

src/cluster_legacy.c
```diff
@@ -5001,7 +5002,7 @@ void clusterCron(void) {
 
     /* Ping some random node 1 time every 10 iterations, so that we usually ping
      * one random node every second. */
-    if (!(iteration % 10)) {
+    if (server.debug_cluster_random_ping && !(iteration % 10)) {
```
```suggestion
    if (!server.debug_disable_cluster_ping && !(iteration % 10)) {
```
I think this will come across as a debug feature that isn't normally enabled because of the name. I think we should invert the flag so it's explicitly disabling some functionality.

madolson (commented):
Overall I think sending out the broadcast of messages is fine and should be safe.

src/debug.c
```diff
@@ -436,8 +436,8 @@ void debugCommand(client *c) {
             "CLOSE-CLUSTER-LINK-ON-PACKET-DROP <0|1>",
             "    This is valid only when DROP-CLUSTER-PACKET-FILTER is set to a valid packet type.",
             "    When set to 1, the cluster link is closed after dropping a packet based on the filter.",
-            "CLUSTER-RANDOM-PING <0|1>",
-            "    Send cluster ping to a random node every second. Enabled by default.",
+            "DISABLE-CLUSTER-RANDOM-PING <0|1>",
+            "    Send cluster ping to a random node every second.",
```
zuiderkwast:
```suggestion
            "    Disable sending cluster ping to a random node every second.",
```

madolson (approved):

PR #1297, opened 2024-11-13
Unprotect rdb channel when bgsave child fails in dual channel replication

enjoy-binbin:
If bgsaveerr is error, there is no need to protect the rdb channel.
The impact of this may be that when bgsave fails, we will protect
the rdb channel for 60s. It may occupy the reference of the repl
buf block, making it impossible to recycle it until we free the
client due to COB or free the client after 60s.

We kept the RDB channel open as long as the replica hadn't established
a main connection, even if the snapshot process failed. There is no
value in keeping the RDB client in this case.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -1745,6 +1745,10 @@ void updateReplicasWaitingBgsave(int bgsaveerr, int type) {
             struct valkey_stat buf;
 
             if (bgsaveerr != C_OK) {
+                if (replica->flag.protected_rdb_channel) {
+                    /* If bgsaveerr is error, there is no need to protect the rdb channel. */
+                    replica->flag.protected_rdb_channel = 0;
+                }
                 freeClientAsync(replica);
                 serverLog(LL_WARNING, "SYNC failed. BGSAVE child returned an error");
                 continue;
--- a/src/replication.c
+++ b/src/replication.c
@@ -1745,10 +1745,8 @@ void updateReplicasWaitingBgsave(int bgsaveerr, int type) {
             struct valkey_stat buf;
 
             if (bgsaveerr != C_OK) {
-                if (replica->flag.protected_rdb_channel) {
-                    /* If bgsaveerr is error, there is no need to protect the rdb channel. */
-                    replica->flag.protected_rdb_channel = 0;
-                }
+                /* If bgsaveerr is error, there is no need to protect the rdb channel. */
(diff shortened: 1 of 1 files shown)


enjoy-binbin:
The issue was found in https://github.com/valkey-io/valkey/pull/1288#issuecomment-2469771257

src/replication.c
```diff
@@ -1745,6 +1745,10 @@ void updateReplicasWaitingBgsave(int bgsaveerr, int type) {
             struct valkey_stat buf;
 
             if (bgsaveerr != C_OK) {
+                if (replica->flag.protected_rdb_channel) {
+                    /* If bgsaveerr is error, there is no need to protect the rdb channel. */
+                    replica->flag.protected_rdb_channel = 0;
+                }
```
naglera:
```suggestion
                /* If bgsaveerr is error, there is no need to protect the rdb channel. */
                replica->flag.protected_rdb_channel = 0;
```

naglera: We can reset `replica->flag.protected_rdb_channel` either way

naglera (approved):
Thank you, @enjoy-binbin, for spotting that. We kept the RDB channel open as long as the replica hadn't established a main connection, even if the snapshot process failed. 
There is no value in keeping the RDB client in this case.

src/replication.c
enjoy-binbin: yes, i originally kept it with the idea of ​​maybe adding some logs

src/replication.c
```diff
@@ -1745,6 +1745,8 @@ void updateReplicasWaitingBgsave(int bgsaveerr, int type) {
             struct valkey_stat buf;
 
             if (bgsaveerr != C_OK) {
+                /* If bgsaveerr is error, there is no need to protect the rdb channel. */
+                replica->flag.protected_rdb_channel = 0;
```
xbasel, line 1749: Is there a cleanup function or handler where this can be done?

enjoy-binbin: It seems that there is no, @naglera do you have one?

zuiderkwast (approved):

src/replication.c
enjoy-binbin: i am going to merge this one so that the CI can happy

naglera: We don't have cleanup function for this stage

PR #837, opened 2024-07-29
Fix bugs related to synchronous TLS connections blocking

naglera:
- Fix TLS bug where connection were shutdown by primary's main process while the child process was still writing- causing main process to be blocked.
- TLS connection fix -file descriptors are set to blocking mode in the main thread, followed by a blocking write. This sets the file descriptors to non-blocking if TLS is used (see `connTLSSyncWrite()`) (@xbasel). 
- Improve the reliability of dual-channel tests. Modify the pause mechanism to verify process status directly, rather than relying on log.
- Ensure that `server.repl_offset` and `server.replid` are updated correctly when dual channel synchronization completes successfully. Thist led to failures in replication tests that validate replication IDs or compare replication offsets.

PR #1298, opened 2024-11-13
Preserve original fd blocking state in TLS I/O operations

xbasel:
This change prevents unintended side effects on connection state and improves consistency with non-TLS sync operations.

For example, when invoking `connTLSSyncRead` with a blocking file descriptor, the mode is switched to non-blocking upon `connTLSSyncRead` exit. If the code assumes the file descriptor remains blocking and calls the normal `read` expecting it to block, it may result in a short read.

This caused a crash in dual-channel, which was fixed in this PR by relocating `connBlock()`:
https://github.com/valkey-io/valkey/pull/837

### DIFF
--- a/src/anet.c
+++ b/src/anet.c
@@ -70,14 +70,21 @@ int anetGetError(int fd) {
     return sockerr;
 }
 
-int anetSetBlock(char *err, int fd, int non_block) {
+static int anetGetSocketFlags(char *err, int fd) {
     int flags;
 
-    /* Set the socket blocking (if non_block is zero) or non-blocking.
-     * Note that fcntl(2) for F_GETFL and F_SETFL can't be
-     * interrupted by a signal. */
     if ((flags = fcntl(fd, F_GETFL)) == -1) {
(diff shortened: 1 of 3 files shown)


xbasel:
Please note that timeouts are still not restored. Not sure we should worry about timeouts.

ranshid (commented):
I agree that the timeout is not the thing we should keep track of, however I wonder why we cannot just flag the connection with indication that it is blocked instead of going through the fcntl all the time?

xbasel:
> I agree that the timeout is not the thing we should keep track of, however I wonder why we cannot just flag the connection with indication that it is blocked instead of going through the fcntl all the time?

Using an app-level flag adds complexity and risks desync with the kernel. fcntl is lightweight, reliable, and avoids duplication.

ranshid:
@xbasel can you please signoff your commit?

Also please elaborate more on the top comment. provide some example etc...

ranshid (approved):

enjoy-binbin (commented):
It seems we didn't backport this to 8.0.

PR #1301, opened 2024-11-13
Fix log printing always shows the role as child under daemonize

enjoy-binbin:
In #1282, we init server.pid earlier to keep log message role
consistent, but we forgot to consider daemonize. In daemonize
mode, we will always print the child role.

We need to reset server.pid after daemonize(), otherwise the
log printing role will always be the child. It also causes a
incorrect server.pid value, affecting the concatenation of
some pid names.


zuiderkwast (approved):
Great, thanks!

@azuredream FYI

xbasel (approved):

PR #1303, opened 2024-11-14
Introduce FORCE_DEFRAG compilation option to allow activedefrag run when allocator is not jemalloc

### LINKED ISSUE #1241: [NEW]Allow activedefrag to run when jemalloc is not used

Currently we only allow activedefrag mechanism to run when jemalloc is defined as the memory allocator.
JEmalloc is not compatible with memory instrumentation tools like Valgrind and AddressSanitizer to be used together with activedefrag. Aside from the fact that this limitation prevents the use of activedefrag feature when one wants to use a different allocator, When enabled, activedefrag constantly replaces memory allocations for different allocations like data objects, dictionaries etc... the problem is that in many situations the data object might have several references which might not be be updated and considered when the code was introduced. This might lead to different write-after-free scenarios which are extremly hard to debug and our test coverage will not cover them (as activedefrag will not be enabled when memory instrumentation is used). 

Although by default valkey is used together with JEmalloc, having the ability to  run activedefrag with other allocators can help identify defrag related issues during tests.
 
This proposal is to enable the ability to explicitly define the HAVE_DEFRAG flag during Valkey compilation. In case it was not explicitly enabled the system will behave the same as today. 
The active defrag code is very independent and the main dependencies are:

je_mallctl - used in order to get current statistics about the memory allocation state. there is really no good parallel functionality with other allocators like tcmalloc and glibc malloc. it is true that mallinfo2 (supported from glibc 2.33) can provide hints about potential external fragmentation, but AFAIK there is no real way to deduct the fragmentation ratio using it.
I think we can provide a NEW DEBUG subcommand or internal config that will force the return value of getAllocatorFragmentation when compiled in that mode (no jemalloc + HAVE_DEFRAG).

je_get_defrag_hint - this function was added to jemalloc in order to support decision to defrag memory buffer allocation. It compares the utilization ratio of the slab hiolding the allocation in question to the total average among non-full slabs.
I think that since this mode is meant for testing purpose, we can just provide a naive implementation of it which always return true so that we can maximize the propability of locating issues related to defrag.    

JEmalloc also has a [debug mode](https://github.com/jemalloc/jemalloc/wiki/Use-Case:-Find-a-memory-corruption-bug) and some other configurations which helps locate memory corruption issues. However these can mainly track down double free and non clean memory allocations which will require to add many type of checks on buffers before reading/writing to them. For example it will not catch case of write-after-free. I think that testing with instrumentation tools (eg AddressSanitizer) would yield much better coverage.

ranshid:
Introduce compile time option to force activedefrag to run even when
jemalloc is not used as the allocator.
This is in order to be able to run tests with defrag enabled
while using memory instrumentation tools.

fixes: https://github.com/valkey-io/valkey/issues/1241


madolson:
I think we'll want to merge this PR first btw, https://github.com/valkey-io/valkey/pull/1242, since it already exists. I think there will be some merge conflicts.

CMakeLists.txt
```diff
@@ -41,3 +41,4 @@ unset(BUILD_UNIT_TESTS CACHE)
 unset(BUILD_TEST_MODULES CACHE)
 unset(BUILD_EXAMPLE_MODULES CACHE)
 unset(USE_TLS CACHE)
+unset(FORCE_DEFRAG CACHE)
```
madolson: We should add this to some CI presumably. The daily ASAN probably?

src/CMakeLists.txt
```diff
@@ -22,6 +22,12 @@ if (VALKEY_RELEASE_BUILD)
     set_property(TARGET valkey-server PROPERTY INTERPROCEDURAL_OPTIMIZATION TRUE)
 endif ()
 
+if (FORCE_DEFRAG)
+    message(STATUS "Forcing Active Defrag run on valkey-server")
+    target_compile_definitions(valkey-server PRIVATE FORCE_DEFRAG)
```
madolson: I would prefer a slightly more verbose and descriptive name, like, "DEBUG_FORCE_DEFRAG" or something, to indicate this isn't really used for production.

src/defrag.c
```diff
@@ -755,6 +755,15 @@ void defragScanCallback(void *privdata, const dictEntry *de) {
  * or not, a false detection can cause the defragmenter to waste a lot of CPU
  * without the possibility of getting any results. */
 float getAllocatorFragmentation(size_t *out_frag_bytes) {
```
madolson: Kind of feels like you should override computeDefragCycles instead, and have that should always set `active_defrag_running` to like 100%.

src/defrag.c
```diff
@@ -956,7 +966,7 @@ void activeDefragCycle(void) {
     mstime_t latency;
     int all_stages_finished = 0;
     int quit = 0;
-
+#if !defined(FORCE_DEFRAG)
```
madolson: I still kind of think you should have to turn defrag on. I'm okay with overriding it so it always runs if you do, but it it's sort of counter intuitive to comment out this block of code.

madolson (commented):
Mostly makes sense to me. Some ideas for clarity but nothing to really change the direction.

CMakeLists.txt
ranshid: Yes I agree. maybe it can also be run as part of the CI, depending on the extra time it would take.

src/defrag.c
ranshid: That was my initial thought as well but I went all the way :) 
an alternative would be to make the active defrag config on by default when this compilation flag is on. WDYT?

src/defrag.c
ranshid: I am fine with it (makes sense) I do not, however want to break tests which are checking the active_defrag_running, so maybe I can just use the active-defrag-cycle-max as the return value?
Alternatively we can have a tag to skip tests in case of DEBUG_FORCE_DEFRAG

src/defrag.c
madolson: > an alternative would be to make the active defrag config on by default when this compilation flag is on. WDYT?

I'm okay with this, it makes sense to me.

src/defrag.c
madolson: > Alternatively we can have a tag to skip tests in case of DEBUG_FORCE_DEFRAG

This intuitively makes a bit more sense to me, since defrag shouldn't really work correctly since we are completely breaking defrag here.

madolson: All of the tests already are checking for jemalloc memory to enable the defragmentation, so I think that the tests should all still be skipped anyways?

madolson:
@ranshid Zvis change is now merged. Jim's is basically also ready to go so you should be able to resume updating this if you want.

src/defrag.c
xbasel: Can the linker feature `--wrap` (used via gcc with `-Wl,--wrap`) be leveraged to mock `getAllocatorFragmentation` and `get_defrag_hint`, providing a less invasive and cleaner solution for testing purposes?

ranshid: nice idea. let me take a look and see if this can be used

ranshid:
> @ranshid Zvis change is now merged. Jim's is basically also ready to go so you should be able to resume updating this if you want.

Thank you. just got time to circle back on this one. merging and continue where I left off

CMakeLists.txt
ranshid: Added another workflow to the daily

src/CMakeLists.txt
ranshid: done

src/defrag.c
ranshid: I think my last change keeps the same logic and the code is cleaner. I find having the defrag reporting the fragmentation based on the config limits is fine but take a look and let me know.

src/defrag.c
ranshid: done

src/defrag.c
ranshid: @xbasel I think that my last change keeps all overriden functions in the allocator_defrag.h so not much value should come from wrapping them and reference by the linker

ranshid:
we will probably hit issues from day 1. eg:

```
Logged sanitizer errors (pid 53103):
==53103==ERROR: AddressSanitizer: heap-use-after-free on address 0x6080000026b8 at pc 0x0001006130d0 bp 0x00016fa72ec0 sp 0x00016fa72eb8
READ of size 4 at 0x6080000026b8 thread T0
    #0 0x1006130cc in defragStageKvstoreHelper+0x650 (valkey-server:arm64+0x1002870cc)
    #1 0x100610d64 in activeDefragTimeProc+0x2e0 (valkey-server:arm64+0x100284d64)
    #2 0x1003c128c in aeProcessEvents+0x630 (valkey-server:arm64+0x10003528c)
    #3 0x10040334c in main+0x6d7c (valkey-server:arm64+0x10007734c)
    #4 0x19d1ff150 in start+0x9a8 (dyld:arm64e+0xfffffffffff4d150)

0x6080000026b8 is located 24 bytes inside of 96-byte region [0x6080000026a0,0x608000002700)
freed by thread T3 here:
    #0 0x101338d40 in free+0x98 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x54d40)
    #1 0x1005bed10 in lazyfreeFreeDatabase+0x54 (valkey-server:arm64+0x100232d10)
    #2 0x100578750 in bioProcessBackgroundJobs+0x4e8 (valkey-server:arm64+0x1001ec750)
    #3 0x101335858 in asan_thread_start(void*)+0x40 (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x51858)
    #4 0x19d589f90 in _pthread_start+0x84 (libsystem_pthread.dylib:arm64e+0x6f90)
    #5 0x19d584d30 in thread_start+0x4 (libsystem_pthread.dylib:arm64e+0x1d30)

previously allocated by thread T0 here:
    #0 0x101338fd0 in calloc+0x9c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x54fd0)
    #1 0x1004120a0 in valkey_calloc+0x3c (valkey-server:arm64+0x1000860a0)
    #2 0x1003cd844 in kvstoreCreate+0xa8 (valkey-server:arm64+0x100041844)
    #3 0x1005bf778 in emptyDbAsync+0xbc (valkey-server:arm64+0x100233778)
    #4 0x100458ce8 in emptyData+0x354 (valkey-server:arm64+0x1000ccce8)
    #5 0x100459cd4 in flushallCommand+0x1c0 (valkey-server:arm64+0x1000cdcd4)
    #6 0x1003e9500 in call+0x3cc (valkey-server:arm64+0x10005d500)
    #7 0x1003ec7a4 in processCommand+0x17b4 (valkey-server:arm64+0x1000607a4)
    #8 0x100424358 in processInputBuffer+0x498 (valkey-server:arm64+0x100098358)
    #9 0x100422204 in readQueryFromClient+0xd4 (valkey-server:arm64+0x100096204)
    #10 0x100668688 in connSocketEventHandler+0x160 (valkey-server:arm64+0x1002dc688)
    #11 0x1003c0fb8 in aeProcessEvents+0x35c (valkey-server:arm64+0x100034fb8)
    #12 0x10040334c in main+0x6d7c (valkey-server:arm64+0x10007734c)
    #13 0x19d1ff150 in start+0x9a8 (dyld:arm64e+0xfffffffffff4d150)

Thread T3 created by T0 here:
    #0 0x1013301c8 in pthread_create+0x5c (libclang_rt.asan_osx_dynamic.dylib:arm64e+0x4c1c8)
    #1 0x100578104 in bioInit+0x28c (valkey-server:arm64+0x1001ec104)
    #2 0x1003ff6b4 in main+0x30e4 (valkey-server:arm64+0x1000736b4)
    #3 0x19d1ff150 in start+0x9a8 (dyld:arm64e+0xfffffffffff4d150)
```
I think it is after the changes introduced in https://github.com/valkey-io/valkey/pull/1242 since the db expires can be freed in the background but is passed as target in beginDefragCycle.

I wonder if to invest time to fix it now or address them later

src/zmalloc.c
```diff
@@ -205,6 +205,7 @@ void *zmalloc_usable(size_t size, size_t *usable) {
     return ptr;
 }
 
+
```
madolson:
```suggestion
```

.github/workflows/daily.yml
```diff
@@ -689,6 +689,56 @@ jobs:
...
+        (github.event_name == 'schedule' && github.repository == 'valkey-io/valkey') ||
+        (github.event_name == 'pull_request' && github.event.pull_request.base.ref != 'unstable')) &&
+      !contains(github.event.inputs.skipjobs, 'sanitizer')
+    timeout-minutes: 14400
+    strategy:
+      fail-fast: false
+      matrix:
+        compiler: [gcc, clang]
+    env:
+      CC: ${{ matrix.compiler }}
```
madolson:
```suggestion
```
Do we really need two compilers? I think just using GCC should be enough.

madolson (approved):
Direction looks fine to me, but just a minor comment about removing one of the test permutations.

src/allocator_defrag.h
```diff
@@ -5,10 +5,11 @@
 #include <jemalloc/jemalloc.h>
 /* We can enable the server defrag capabilities only if we are using Jemalloc
  * and the version that has the experimental.utilization namespace in mallctl . */
-#if defined(JEMALLOC_VERSION_MAJOR) &&                              \
-    (JEMALLOC_VERSION_MAJOR > 5 ||                                  \
-     (JEMALLOC_VERSION_MAJOR == 5 && JEMALLOC_VERSION_MINOR > 2) || \
-     (JEMALLOC_VERSION_MAJOR == 5 && JEMALLOC_VERSION_MINOR == 2 && JEMALLOC_VERSION_BUGFIX >= 1))
+#if (defined(JEMALLOC_VERSION_MAJOR) &&                                                                 \
+     (JEMALLOC_VERSION_MAJOR > 5 ||                                                                      \
```
madolson:
```suggestion
     (JEMALLOC_VERSION_MAJOR > 5 ||                                                                     \
```

PR #1305, opened 2024-11-14
Manual failover vote is not limited by two times the node timeout

enjoy-binbin:
This limit should not restrict manual failover, otherwise in some
scenarios, manual failover will time out.

For example, if some FAILOVER_AUTH_REQUESTs or some FAILOVER_AUTH_ACKs
are lost during a manual failover, it cannot vote in the second manual
failover. Or in a mixed scenario of plain failover and manual failover,
it cannot vote for the subsequent manual failover.

The problem with the manual failover retry is that the mf will pause
the client 5s in the primary side. So every retry every manual failover
timed out is a bad move.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -4371,8 +4371,9 @@ void clusterSendFailoverAuthIfNeeded(clusterNode *node, clusterMsg *request) {
     if (!force_ack && mstime() - node->replicaof->voted_time < server.cluster_node_timeout * 2) {
         serverLog(LL_WARNING,
                   "Failover auth denied to %.40s (%s): "
-                  "can't vote about this replica before %lld milliseconds",
+                  "can't vote for any replica of %.40s (%s) within %lld milliseconds",
                   node->name, node->human_nodename,
+                  node->replicaof->name, node->replicaof->human_nodename,
                   (long long)((server.cluster_node_timeout * 2) - (mstime() - node->replicaof->voted_time)));
         return;
     }
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -4363,11 +4363,15 @@ void clusterSendFailoverAuthIfNeeded(clusterNode *node, clusterMsg *request) {
 
     /* We did not voted for a replica about this primary for two
      * times the node timeout. This is not strictly needed for correctness
-     * of the algorithm but makes the base case more linear. */
-    if (mstime() - node->replicaof->voted_time < server.cluster_node_timeout * 2) {
+     * of the algorithm but makes the base case more linear.
+     *
+     * This limitation does not restrict manual failover. If a user initiates
+     * a manual failover, we need to allow it to vote, otherwise the manual
+     * failover may time out. */
+    if (!force_ack && mstime() - node->replicaof->voted_time < server.cluster_node_timeout * 2) {
         serverLog(LL_WARNING,
-                  "Failover auth denied to %.40s %s: "
-                  "can't vote about this primary before %lld milliseconds",
+                  "Failover auth denied to %.40s (%s): "
+                  "can't vote about this replica before %lld milliseconds",
                   node->name, node->human_nodename,
                   (long long)((server.cluster_node_timeout * 2) - (mstime() - node->replicaof->voted_time)));
         return;
@@ -4394,7 +4398,7 @@ void clusterSendFailoverAuthIfNeeded(clusterNode *node, clusterMsg *request) {
 
     /* We can vote for this replica. */
     server.cluster->lastVoteEpoch = server.cluster->currentEpoch;
-    node->replicaof->voted_time = mstime();
+    if (!force_ack) node->replicaof->voted_time = mstime();
     clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_FSYNC_CONFIG);
     clusterSendFailoverAuth(node);
     serverLog(LL_NOTICE, "Failover auth granted to %.40s (%s) for epoch %llu", node->name, node->human_nodename,
--- a/tests/unit/cluster/manual-failover.tcl
+++ b/tests/unit/cluster/manual-failover.tcl
@@ -212,7 +212,7 @@ start_cluster 3 1 {tags {external:skip cluster} overrides {cluster-ping-interval
             [s 0 role] eq {slave} &&
             [s -3 role] eq {master}
         } else {
-            fail "The third falover does not happened"
+            fail "The second failover does not happen"
         }
         wait_for_cluster_propagation
     }
@@ -233,12 +233,12 @@ start_cluster 3 1 {tags {external:skip cluster} overrides {cluster-ping-interval
         R 1 config set cluster-node-timeout 150000
         R 2 config set cluster-node-timeout 150000
 
-        # R 3 performs a plain failover and it will work.
+        # R 3 performs an automatic failover and it will work.
         R 3 config set cluster-replica-no-failover no
         wait_for_condition 1000 50 {
             [s -3 role] eq {master}
         } else {
-            fail "The first failover does not happened"
+            fail "The first failover does not happen"
         }
 
         # Resume the primary and wait for it to become a replica.
@@ -256,7 +256,7 @@ start_cluster 3 1 {tags {external:skip cluster} overrides {cluster-ping-interval
             [s 0 role] eq {master} &&
             [s -3 role] eq {slave}
         } else {
-            fail "The second failover does not happened"
+            fail "The second failover does not happen"
         }
         wait_for_cluster_propagation
 
@@ -266,7 +266,7 @@ start_cluster 3 1 {tags {external:skip cluster} overrides {cluster-ping-interval
             [s 0 role] eq {slave} &&
             [s -3 role] eq {master}
         } else {
-            fail "The third falover does not happened"
+            fail "The third falover does not happen"
         }
         wait_for_cluster_propagation
     }
--- a/tests/unit/cluster/manual-failover.tcl
+++ b/tests/unit/cluster/manual-failover.tcl
@@ -183,3 +183,91 @@ test "Wait for instance #0 to return back alive" {
 }
 
 } ;# start_cluster
+
+start_cluster 3 1 {tags {external:skip cluster} overrides {cluster-ping-interval 1000 cluster-node-timeout 2000}} {
+    test "Manual failover vote is not limited by two times the node timeout - drop the auth ack" {
+        set CLUSTER_PACKET_TYPE_FAILOVER_AUTH_ACK 6
+        set CLUSTER_PACKET_TYPE_NONE -1
+
+        # Setting a large timeout to make sure we hit the voted_time limit.
+        R 0 config set cluster-node-timeout 150000
+        R 1 config set cluster-node-timeout 150000
+        R 2 config set cluster-node-timeout 150000
+
+        # Let replica drop FAILOVER_AUTH_ACK so that the election won't
+        # get the enough votes and the election will time out.
+        R 3 debug drop-cluster-packet-filter $CLUSTER_PACKET_TYPE_FAILOVER_AUTH_ACK
+
+        # The first manual failover will time out.
+        R 3 cluster failover
+        wait_for_log_messages 0 {"*Manual failover timed out*"} 0 1000 50
+        wait_for_log_messages -3 {"*Manual failover timed out*"} 0 1000 50
+
+        # Undo packet drop, so that replica can win the next election.
+        R 3 debug drop-cluster-packet-filter $CLUSTER_PACKET_TYPE_NONE
+
+        # Make sure the second manual failover will work.
+        R 3 cluster failover
+        wait_for_condition 1000 50 {
+            [s 0 role] eq {slave} &&
+            [s -3 role] eq {master}
+        } else {
+            fail "The third falover does not happened"
+        }
+        wait_for_cluster_propagation
+    }
+} ;# start_cluster
+
+start_cluster 3 1 {tags {external:skip cluster} overrides {cluster-ping-interval 1000 cluster-node-timeout 2000}} {
+    test "Manual failover vote is not limited by two times the node timeout - mixed failover" {
+        # Make sure the failover is triggered by us.
+        R 1 config set cluster-replica-validity-factor 0
+        R 3 config set cluster-replica-no-failover yes
+        R 3 config set cluster-replica-validity-factor 0
+
+        # Pause the primary.
+        pause_process [srv 0 pid]
+        wait_for_cluster_state fail
+
+        # Setting a large timeout to make sure we hit the voted_time limit.
+        R 1 config set cluster-node-timeout 150000
+        R 2 config set cluster-node-timeout 150000
+
+        # R 3 performs a plain failover and it will work.
+        R 3 config set cluster-replica-no-failover no
+        wait_for_condition 1000 50 {
+            [s -3 role] eq {master}
+        } else {
+            fail "The first failover does not happened"
+        }
+
+        # Resume the primary and wait for it to become a replica.
+        resume_process [srv 0 pid]
+        wait_for_condition 1000 50 {
+            [s 0 role] eq {slave}
+        } else {
+            fail "Old primary not converted into replica"
+        }
+        wait_for_cluster_propagation
+
+        # The old primary doing a manual failover and wait for it.
+        R 0 cluster failover
+        wait_for_condition 1000 50 {
+            [s 0 role] eq {master} &&
+            [s -3 role] eq {slave}
+        } else {
+            fail "The second failover does not happened"
+        }
+        wait_for_cluster_propagation
+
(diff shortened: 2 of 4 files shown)


madolson:
> For example, if some FAILOVER_AUTH_REQUESTs or some FAILOVER_AUTH_ACKs
are lost during a manual failover, it cannot vote in the second manual
failover. Or in a mixed scenario of plain failover and manual failover,
it cannot vote for the subsequent manual failover.

I'm not sure I agree with this. I think there should be some built in timeout into the system and you should retry.

enjoy-binbin:
The problem with the manual failover retry is that the mf will pause the client 5s in the primary side. So every retry every manual failover timed out is a bad move

tests/unit/cluster/manual-failover.tcl
```diff
@@ -183,3 +183,91 @@ test "Wait for instance #0 to return back alive" {
...
+        # Undo packet drop, so that replica can win the next election.
+        R 3 debug drop-cluster-packet-filter $CLUSTER_PACKET_TYPE_NONE
+
+        # Make sure the second manual failover will work.
+        R 3 cluster failover
+        wait_for_condition 1000 50 {
+            [s 0 role] eq {slave} &&
+            [s -3 role] eq {master}
+        } else {
+            fail "The third falover does not happened"
```
zuiderkwast: The third failover or the second?

```suggestion
            fail "The second failover does not happen"
```

tests/unit/cluster/manual-failover.tcl
```diff
@@ -183,3 +183,91 @@ test "Wait for instance #0 to return back alive" {
...
+
+        # Pause the primary.
+        pause_process [srv 0 pid]
+        wait_for_cluster_state fail
+
+        # Setting a large timeout to make sure we hit the voted_time limit.
+        R 1 config set cluster-node-timeout 150000
+        R 2 config set cluster-node-timeout 150000
+
+        # R 3 performs a plain failover and it will work.
```
zuiderkwast:
```suggestion
        # R 3 performs an automatic failover and it will work.
```

tests/unit/cluster/manual-failover.tcl
```diff
@@ -183,3 +183,91 @@ test "Wait for instance #0 to return back alive" {
...
+        R 3 config set cluster-replica-no-failover no
+        wait_for_condition 1000 50 {
+            [s -3 role] eq {master}
+        } else {
+            fail "The first failover does not happened"
```
zuiderkwast:
```suggestion
            fail "The first failover does not happen"
```

tests/unit/cluster/manual-failover.tcl
```diff
@@ -183,3 +183,91 @@ test "Wait for instance #0 to return back alive" {
...
+        }
+        wait_for_cluster_propagation
+
+        # The old primary doing a manual failover and wait for it.
+        R 0 cluster failover
+        wait_for_condition 1000 50 {
+            [s 0 role] eq {master} &&
+            [s -3 role] eq {slave}
+        } else {
+            fail "The second failover does not happened"
```
zuiderkwast:
```suggestion
            fail "The second failover does not happen"
```

tests/unit/cluster/manual-failover.tcl
```diff
@@ -183,3 +183,91 @@ test "Wait for instance #0 to return back alive" {
...
+        }
+        wait_for_cluster_propagation
+
+        # R 3 performs a manual failover and it will work.
+        R 3 cluster failover
+        wait_for_condition 1000 50 {
+            [s 0 role] eq {slave} &&
+            [s -3 role] eq {master}
+        } else {
+            fail "The third falover does not happened"
```
zuiderkwast:
```suggestion
            fail "The third falover does not happen"
```

zuiderkwast (commented):
> The problem with the manual failover retry is that the mf will pause the client 5s in the primary side. So every retry every manual failover timed out is a bad move

Yes, client pause for a long time is a bad move. :chess_pawn: :x:

We already had this comment:

    /* We did not voted for a replica about this primary for two
     * times the node timeout. This is not strictly needed for correctness
     * of the algorithm but makes the base case more linear. */

Hm, not strictly needed for correctness means that it's OK to change it. It doesn't affect correctness. You added this to the same comment:

     * This limitation does not restrict manual failover. If a user initiates
     * a manual failover, we need to allow it to vote, otherwise the manual
     * failover may time out. */

I think it's safe. I like the fix. The test cases look good too. Just some nits.

tests/unit/cluster/manual-failover.tcl
enjoy-binbin: it is the second, you are right

src/cluster_legacy.c
```diff
@@ -4363,11 +4363,15 @@ void clusterSendFailoverAuthIfNeeded(clusterNode *node, clusterMsg *request) {
...
+     *
+     * This limitation does not restrict manual failover. If a user initiates
+     * a manual failover, we need to allow it to vote, otherwise the manual
+     * failover may time out. */
+    if (!force_ack && mstime() - node->replicaof->voted_time < server.cluster_node_timeout * 2) {
         serverLog(LL_WARNING,
-                  "Failover auth denied to %.40s %s: "
-                  "can't vote about this primary before %lld milliseconds",
+                  "Failover auth denied to %.40s (%s): "
+                  "can't vote about this replica before %lld milliseconds",
```
zuiderkwast: I think the text "can't vote about this primary" means to select a new primary for the shard. It is bad English though.

If we want to talk about the replica, we should say "vote for this primary". Also "before" is some wrong grammar.

```suggestion
                  "Failover auth denied to %.40s (%s): "
                  "can't vote for this replica within %lld milliseconds",
```

But the timestamp `node->replicaof->voted_time` is actually for all replicas of the same primary, so we can't vote for any of them within this time.

I don't know if there is a simle way to write this. It is actually about the primary, so shall we mention the primary node name in the log? Something like this?

```suggestion
                  "Failover auth denied to %.40s (%s): "
                  "can't vote for any replica of %.40s (%s) within %lld milliseconds",
```

zuiderkwast (approved):
Looks good. I added a comment about a log message, because you touched it. :) We don't have to fix it though.

src/cluster_legacy.c
enjoy-binbin: right, you are right about this.

enjoy-binbin:
i was hoping that @PingXie and @madolson can take another look at this.

madolson:
> The problem with the manual failover retry is that the mf will pause the client 5s in the primary side. So every retry every manual failover timed out is a bad move

Yeah, I suppose that makes sense. I guess the concern is that this increases the chances of two replicas getting promoted, which can result in some data loss. If replica A takes over, after winning the election, it will start serving some traffic, then independently replica C was sent a takeover, it will likely not have time to reconfigure to A, and will take over and drop all of the writes that A received.

I suppose I'm not that worried about the new behavior since manual failovers already can cause data loss.

> Hm, not strictly needed for correctness means that it's OK to change it.

You can materially degrade the end user experience but still be "correct".

src/cluster_legacy.c
```diff
@@ -4394,7 +4399,7 @@ void clusterSendFailoverAuthIfNeeded(clusterNode *node, clusterMsg *request) {
 
     /* We can vote for this replica. */
     server.cluster->lastVoteEpoch = server.cluster->currentEpoch;
-    node->replicaof->voted_time = mstime();
+    if (!force_ack) node->replicaof->voted_time = mstime();
```
madolson, line 4402: I don't get this, don't we still want to fence off normal failovers if we are triggering a manual failover?

zuiderkwast:
> I suppose I'm not that worried about the new behavior since manual failovers already can cause data loss.

Can they? Can you elaborate? The documentation says "without any window for data loss". We need to fix the docs in that case. I was under the impression that it's safe to use manual failover. https://valkey.io/commands/cluster-failover/

> A  manual  failover  is  a special kind of failover that is usually executed when there are no actual failures, but we wish to swap the current master with one of its replicas (which is the node we send the command to), in a safe way, without any window for data loss.

madolson:
> Can they? Can you elaborate?

Well, generally we can lose tail data because we do async replication. Force and takeover can also cause dataloss, since they are just forcing their way into the primaries without being the leader.

There is also the case basically outlined here, which is that a a node could have already won the leader election for a shard but the rest of the shard doesn't know and start accepting traffic. An operator could force the replica to takeover from what i thinks is the primary, but just doesn't know there is another primary. The bound was previously the nodetimeout * 2, now we removed that guard. We could guard against this by having a failover include who it thinks own the slots it is taking over.

src/cluster_legacy.c
enjoy-binbin: Do you mean both normal failover and the manual failover happen in the same time? This must happen in a force or takeover path? I think it might be rare, adding this is because i am worried that after a manual failover, if the primary is down quickly, then the normal failover will not be able to finish. I want all the manual failover bypass this logic.

enjoy-binbin:
> Well, generally we can lose tail data because we do async replication. Force and takeover can also cause dataloss, since they are just forcing their way into the primaries without being the leader.

So a plain manual failover can lose tail data? Don't we have client pause and offset wait, even though we do async replication, the wait should ensure that all traffic is consumed? I know foce and takeover can cause data loss, it is expected.

> There is also the case basically outlined here, which is that a a node could have already won the leader election for a shard but the rest of the shard doesn't know and start accepting traffic. An operator could force the replica to takeover from what i thinks is the primary, but just doesn't know there is another primary.

ok i got this point, there is a timing window is that, the admin don't know there is winner, in this case, i think if the admin is doing a "force" or a "takeover", this data loss should be expected. It is race i guess.

> We could guard against this by having a failover include who it thinks own the slots it is taking over.

Sorry, bad english, can you elaborate this a bit more? So you are saying, FAILOVER_AUTH_REQUESTs carry the old primary info, and we only vote for it if the old primary info is match with myself's local view. And we do the same thing in FAILOVER_AUTH_ACK, we only accept it when the ack include the old primary info and match myself's local view. I think this can solve some problem, but still we have a timing window (but rare) that not all the nodes knows the new winner.

src/cluster_legacy.c
madolson: If you send a manual failover it won't reset the voted_time, so an automatic failover request from another replica might come in and then this primary will also send a vote for the new replica that has requested it? This might cause the manual failover to get reverted and cause churn in the system.

The force_ack path is already ignores the voted_time, so why wouldn't we update the voted_time here to prevent a future automatic failover from starting within the timeout?

madolson:
> even though we do async replication, the wait should ensure that all traffic is consumed

I thought it was possible for another replica to start and win an election while the normal failover was ongoing, start serving traffic, and then have the manual failover complete and takeover, which loses the data from the intermediary. I suppose we could say as long as there are no node failures during the failover, it doesn't lose data. 

> ok i got this point, there is a timing window is that, the admin don't know there is winner, in this case, i think if the admin is doing a "force" or a "takeover", this data loss should be expected. It is race i guess.

So my main concern is an automatic failover followed by a manual failover, where the manual failover takes over from the old primary and not the one from the automatic failover. I suppose this case is unlikely anyways, since the primary is both healthy (able to progress in the failover) while still being unhealthy in a way that allowed the automatic failover to happen. The new primary from the automatic failover might have started serving traffic, which might get lost from the manual failover.

src/cluster_legacy.c
enjoy-binbin: ok. i think i got your point, i will remove this change, i will try to think of a test case to cover it.

enjoy-binbin: This is the problem I want to solve, this is why i add this force_ack check. It is odd that i lost the context and did not put it in the TCL tests. The test case will do two manual failover, and the third one is the automatic failover, in here, if we remove the force_ack check, the automatic failover won't get the votes, since other primary has vote for the old primary in the first manual failover. 
```
start_cluster 3 1 {tags {external:skip cluster} overrides {cluster-ping-interval 1000 cluster-node-timeout 5000}} {
    test "xxxx" {
        R 3 cluster failover
        wait_for_condition 1000 50 {
            [s 0 role] eq {slave} &&
            [s -3 role] eq {master}
        } else {
            fail "The first failover does not happen"
        }
        wait_for_cluster_propagation

        R 0 cluster failover
        wait_for_condition 1000 50 {
            [s 0 role] eq {master} &&
            [s -3 role] eq {slave}
        } else {
            fail "The second failover does not happen"
        }
        wait_for_cluster_propagation

        # Let R 3 trigger the automatic failover
        pause_process [srv 0 pid]
        wait_for_cluster_state fail
        after 2000
    }
} ;# start_cluster
```

> If you send a manual failover it won't reset the voted_time, so an automatic failover request from another replica might come in and then this primary will also send a vote for the new replica that has requested it? This might cause the manual failover to get reverted and cause churn in the system.

i try this one, it is hard to write the test case, if they are happen in the same time, it is likely they do the election in the same epoch, so only one can win. If the automatic failover is the after one, it use a old epoch and it cant win the election. Or the manual failover win the election and broadcast the new config, the automatic failover also dont have the chance to win the election. So this look safe for me.

zuiderkwast:
These are interesting cases. They seem very unlikely but they're not theoretically impossible.

As a though experiment, how hard would it be reproduce these scenarios in test cases?

src/cluster_legacy.c
PingXie: I actually don’t see why this wait is necessary for automatic failovers, let alone manual ones. Does anyone have concrete context on why [this commit](https://github.com/valkey-io/valkey/commit/a445aa30a0fb72bc54da083998b3458fad92820f) introduced it? It doesn’t seem tied to any real-world use case or operational challenge.

This enforced wait significantly slows down recovery in scenarios where multiple primaries fail close together—such as during an AZ outage or a major network partition. 

Note that we already debounce transient primary failures with `cluster-node-timeout`, ensuring failover is only triggered after sustained outages. Election timing is naturally staggered by replica spacing, making the likelihood of simultaneous elections from replicas of the same shard very low. The one-vote-per-epoch rule further throttles retries and ensures orderly elections. On top of that, quorum-based primary failure confirmation, cluster-state convergence, and slot ownership validation are all built into the process.

If the concern is about stability during failovers, these existing mechanisms already handle it effectively IMO. Having this global wait not only fails to provide additional protection but actively hurts recovery speed, particularly in high-availability setups where responsiveness is critical.

Please let me know if I miss some specific edge cases that this wait was meant to address. Otherwise, I think we should move forward with removing it for both automatic and manual failovers.

PingXie:
> > ok i got this point, there is a timing window is that, the admin don't know there is winner, in this case, i think if the admin is doing a "force" or a "takeover", this data loss should be expected. It is race i guess.
> 
> So my main concern is an automatic failover followed by a manual failover, where the manual failover takes over from the old primary and not the one from the automatic failover. I suppose this case is unlikely anyways, since the primary is both healthy (able to progress in the failover) while still being unhealthy in a way that allowed the automatic failover to happen. The new primary from the automatic failover might have started serving traffic, which might get lost from the manual failover.

I think `takeover` and `force` are in a completely different category from automatic failover and **normal** manual failover. These are essentially back doors designed for last-resort scenarios, where admins/operators must be fully aware of the risks involved, including potential data loss or inconsistencies, which btw is an existing risk. 

My recommendation would be for admins/operators to ensure they have a clear view of the cluster state before performing `takeover` or `force`. These commands are meant for extraordinary situations and should be used with caution and full knowledge of their implications.

src/cluster_legacy.c
enjoy-binbin: I'd also prefer to remove it completely, if there is other specific edge cases, we should use other way to fix it instead of this limitation (the slow down thing is suck i think and has never been documented).

I thought it was trying to solve this problem before, but it also doesn't make sense. i don't think we will change cluster-node-timeout during it.
```
M, R1, R2

t1: M is donw, R1 do the election, M->voted_time = t1. R1 became the new primry, M online, and then R1 down

after a node timeout

t2: R1 down, M do the election, R1->voted_time = t2, M became the new primary, R1 online, and then M down

after a node timeout:

t3: M down, R1 do the election, M->voted_time = t3 > 2 timeout, R1 become the new primary.
```

madolson: > If the concern is about stability during failovers, these existing mechanisms already handle it effectively IMO. Having this global wait not only fails to provide additional protection but actively hurts recovery speed, particularly in high-availability setups where responsiveness is critical.

This isn't a global wait, it's only against the specific primary. It's to prevent double failovers.

My general understanding, is if a primary just took over we don't want someone else to try to take over and give the new primary some amount of time to take over. I have not seen this issue though, it might have been over optimizing? The double failure mode, where a node fails and then another node fails within the nodetimeout also doesn't seem that common either though.

I guess I agree with removing it. It seems like it will make the code a lot simpler.

enjoy-binbin:
So the conclusion is that we all agreed to remove it completely,
it will make the code a lot simpler. And if there is other specific
edge cases we are missing, we will fix it in other way. see https://github.com/valkey-io/valkey/pull/1305#pullrequestreview-2446273648

The new PR is #1356

PR #461, opened 2024-05-08
Make cluster meet reliable under link failures

srgsanky:
When there is a link failure while an ongoing MEET request is sent the sending node stops sending anymore MEET and starts sending PINGs. Since every node responds to PINGs from unknown nodes with a PONG, the receiving node never adds the sending node. But the sending node adds the receiving node when it sees a PONG. This can lead to asymmetry in cluster membership. This changes makes the sender keep sending MEET until it sees a PONG, avoiding the asymmetry.

PR #1307, opened 2024-11-14
Send MEET packet to node if there is no inbound link to fix inconsistency when handshake timedout

### LINKED ISSUE #1251: [BUG] Clusters can become inconsistent if one side times out the handshake during cluster meets

pieturin:
In some cases, when meeting a new node, if the handshake times out, we can end up with an inconsistent view of the cluster where the new node knows about all the nodes in the cluster, but the cluster does not know about this new node (or vice versa).
To detect this inconsistency, we now check if a node has an outbound link but no inbound link, in this case it probably means this node does not know us. In this case we (re-)send a MEET packet to this node to do a new handshake with it.
If we receive a MEET packet from a known node, we disconnect the outbound link to force a reconnect and sending of a PING packet so that the other node recognizes the link as belonging to us. This prevents cases where a node could send MEET packets in a loop because it thinks the other node does not have an inbound link.

This fixes the bug described in #1251.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3240,10 +3240,12 @@ int clusterProcessPacket(clusterLink *link) {
                  * the gossip section here since we have to trust the sender because
                  * of the message type. */
                 clusterProcessGossipSection(hdr, link);
-            } else if (sender->link) {
+            } else if (sender->link && now - sender->ctime > server.cluster_node_timeout) {
                 /* The MEET packet is from a known node, so the sender thinks that I do not know it.
                  * Freeing my outbound link to that node, to force a reconnect and sending a PING.
-                 * Once that node receives our PING, it should recognize the new connection as an inbound link from me. */
+                 * Once that node receives our PING, it should recognize the new connection as an inbound link from me.
+                 * We should only free the outbound link if the node is known for more time than the handshake timeout,
+                 * since during this time, the other side might still be trying to complete the handshake. */
                 serverAssert(link != sender->link); // we should always receive a MEET packet on an inbound link
                 serverLog(LL_NOTICE, "Freeing outbound link to node %.40s after receiving a MEET packet from this known node", sender->name);
                 freeClusterLink(sender->link);
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3237,8 +3237,8 @@ int clusterProcessPacket(clusterLink *link) {
                 clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG);
 
                 /* If this is a MEET packet from an unknown node, we still process
-                * the gossip section here since we have to trust the sender because
-                * of the message type. */
+                 * the gossip section here since we have to trust the sender because
+                 * of the message type. */
                 clusterProcessGossipSection(hdr, link);
             } else if (sender->link) {
                 /* The MEET packet is from a known node, so the sender thinks that I do not know it.
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3241,13 +3241,17 @@ int clusterProcessPacket(clusterLink *link) {
                  * of the message type. */
                 clusterProcessGossipSection(hdr, link);
             } else if (sender->link && now - sender->ctime > server.cluster_node_timeout) {
-                /* The MEET packet is from a known node, so the sender thinks that I do not know it.
+                /* The MEET packet is from a known node, after the handshake timeout, so the sender thinks that I do not
+                 * know it.
                  * Freeing my outbound link to that node, to force a reconnect and sending a PING.
                  * Once that node receives our PING, it should recognize the new connection as an inbound link from me.
                  * We should only free the outbound link if the node is known for more time than the handshake timeout,
                  * since during this time, the other side might still be trying to complete the handshake. */
-                serverAssert(link != sender->link); // we should always receive a MEET packet on an inbound link
-                serverLog(LL_NOTICE, "Freeing outbound link to node %.40s after receiving a MEET packet from this known node", sender->name);
+
+                /* We should always receive a MEET packet on an inbound link. */
+                serverAssert(link != sender->link);
+                serverLog(LL_NOTICE, "Freeing outbound link to node %.40s after receiving a MEET packet from this known node",
+                          sender->name);
                 freeClusterLink(sender->link);
             }
         }
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -1335,6 +1335,10 @@ clusterLink *createClusterLink(clusterNode *node) {
  * with this link will have the 'link' field set to NULL. */
 void freeClusterLink(clusterLink *link) {
     serverAssert(link != NULL);
+    serverLog(LL_DEBUG, "Freeing cluster link for node: %.40s:%s",
+              link->node ? link->node->name : "<unknown>",
+              link->inbound ? "inbound" : "outbound");
+
     if (link->conn) {
         connClose(link->conn);
         link->conn = NULL;
@@ -1350,6 +1354,7 @@ void freeClusterLink(clusterLink *link) {
         } else if (link->node->inbound_link == link) {
             serverAssert(link->inbound);
             link->node->inbound_link = NULL;
+            link->node->inbound_link_freed_time = mstime();
         }
     }
     zfree(link);
@@ -1489,6 +1494,7 @@ clusterNode *createClusterNode(char *nodename, int flags) {
     node->fail_time = 0;
     node->link = NULL;
     node->inbound_link = NULL;
+    node->inbound_link_freed_time = node->ctime;
     memset(node->ip, 0, sizeof(node->ip));
     node->announce_client_ipv4 = sdsempty();
     node->announce_client_ipv6 = sdsempty();
@@ -1695,6 +1701,9 @@ void clusterAddNode(clusterNode *node) {
  *    it is a replica node.
  */
 void clusterDelNode(clusterNode *delnode) {
+    serverAssert(delnode != NULL);
+    serverLog(LL_DEBUG, "Deleting node %s from cluster view", delnode->name);
+
     int j;
     dictIterator *di;
     dictEntry *de;
@@ -2077,7 +2086,7 @@ void clearNodeFailureIfNeeded(clusterNode *node) {
 /* Return 1 if we already have a node in HANDSHAKE state matching the
  * specified ip address and port number. This function is used in order to
  * avoid adding a new handshake node for the same address multiple times. */
-int clusterHandshakeInProgress(char *ip, int port, int cport) {
+static int clusterHandshakeInProgress(char *ip, int port, int cport) {
     dictIterator *di;
     dictEntry *de;
 
@@ -2099,7 +2108,7 @@ int clusterHandshakeInProgress(char *ip, int port, int cport) {
  *
  * EAGAIN - There is already a handshake in progress for this address.
  * EINVAL - IP or port are not valid. */
-int clusterStartHandshake(char *ip, int port, int cport) {
+static int clusterStartHandshake(char *ip, int port, int cport) {
     clusterNode *n;
     char norm_ip[NET_IP_STR_LEN];
     struct sockaddr_storage sa;
@@ -3225,12 +3234,12 @@ int clusterProcessPacket(clusterLink *link) {
             setClusterNodeToInboundClusterLink(node, link);
             clusterAddNode(node);
             clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG);
-        }
 
-        /* If this is a MEET packet from an unknown node, we still process
-         * the gossip section here since we have to trust the sender because
-         * of the message type. */
-        if (!sender && type == CLUSTERMSG_TYPE_MEET) clusterProcessGossipSection(hdr, link);
+            /* If this is a MEET packet from an unknown node, we still process
+             * the gossip section here since we have to trust the sender because
+             * of the message type. */
+            clusterProcessGossipSection(hdr, link);
+        }
 
         /* Anyway reply with a PONG */
         clusterSendPing(link, CLUSTERMSG_TYPE_PONG);
@@ -3241,7 +3250,7 @@ int clusterProcessPacket(clusterLink *link) {
         serverLog(LL_DEBUG, "%s packet received: %.40s", clusterGetMessageTypeString(type),
                   link->node ? link->node->name : "NULL");
 
-        if (sender && (sender->flags & CLUSTER_NODE_MEET)) {
+        if (sender && nodeIsMeeting(sender)) {
             /* Once we get a response for MEET from the sender, we can stop sending more MEET. */
             sender->flags &= ~CLUSTER_NODE_MEET;
             serverLog(LL_NOTICE, "Successfully completed handshake with %.40s (%s)", sender->name,
@@ -3666,7 +3675,7 @@ void clusterLinkConnectHandler(connection *conn) {
      * of a PING one, to force the receiver to add us in its node
      * table. */
     mstime_t old_ping_sent = node->ping_sent;
-    clusterSendPing(link, node->flags & CLUSTER_NODE_MEET ? CLUSTERMSG_TYPE_MEET : CLUSTERMSG_TYPE_PING);
+    clusterSendPing(link, nodeIsMeeting(node) ? CLUSTERMSG_TYPE_MEET : CLUSTERMSG_TYPE_PING);
     if (old_ping_sent) {
         /* If there was an active ping before the link was
          * disconnected, we want to restore the ping time, otherwise
@@ -3686,6 +3695,9 @@ void clusterLinkConnectHandler(connection *conn) {
      */
 
     serverLog(LL_DEBUG, "Connecting with Node %.40s at %s:%d", node->name, node->ip, node->cport);
+    if (nodeIsMeeting(node)) {
+        serverLog(LL_DEBUG, "Sending MEET packet on connection to node %.40s", node->name);
+    }
 }
 
 /* Performs sanity check on the message signature and length depending on the type. */
@@ -3745,7 +3757,9 @@ void clusterReadHandler(connection *conn) {
 
         if (nread <= 0) {
             /* I/O error... */
-            serverLog(LL_DEBUG, "I/O error reading from node link: %s",
+            serverLog(LL_DEBUG, "I/O error reading from node link (%.40s:%s): %s",
+                      link->node ? link->node->name : "<unknown>",
+                      link->inbound ? "inbound" : "outbound",
                       (nread == 0) ? "connection closed" : connGetLastError(conn));
             handleLinkIOError(link);
             return;
@@ -4941,6 +4955,16 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t handshake_
         clusterDelNode(node);
         return 1;
     }
+    if (node->link != NULL && node->inbound_link == NULL &&
+        !nodeInHandshake(node) && !nodeIsMeeting(node) && !nodeTimedOut(node) &&
+        now - node->inbound_link_freed_time > handshake_timeout) {
+        /* Node has an outbound link, but no inbound link for more than the handshake timeout.
+         * This probably means this node does not know us yet, wherease we know it.
+         * So we send it a MEET packet to do a handshake with it and correct the inconsistent cluster view. */
+        node->flags |= CLUSTER_NODE_MEET;
+        serverLog(LL_NOTICE, "Sending MEET packet to node %.40s because there is no inbound link for it", node->name);
+        clusterSendPing(node->link, CLUSTERMSG_TYPE_MEET);
+    }
 
     if (node->link == NULL) {
         clusterLink *link = createClusterLink(node);
@@ -4963,6 +4987,7 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t handshake_
             return 0;
         }
     }
+
     return 0;
 }
 
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -1702,7 +1702,7 @@ void clusterAddNode(clusterNode *node) {
  */
 void clusterDelNode(clusterNode *delnode) {
     serverAssert(delnode != NULL);
-    serverLog(LL_DEBUG, "Deleting node %s from cluster view", delnode->name);
+    serverLog(LL_DEBUG, "Deleting node %.40s from cluster view", delnode->name);
 
     int j;
     dictIterator *di;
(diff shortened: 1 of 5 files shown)


src/cluster_legacy.c
```diff
@@ -3224,12 +3233,12 @@ int clusterProcessPacket(clusterLink *link) {
...
 
-        /* If this is a MEET packet from an unknown node, we still process
-         * the gossip section here since we have to trust the sender because
-         * of the message type. */
-        if (!sender && type == CLUSTERMSG_TYPE_MEET) clusterProcessGossipSection(hdr, link);
+            /* If this is a MEET packet from an unknown node, we still process
+             * the gossip section here since we have to trust the sender because
+             * of the message type. */
+            clusterProcessGossipSection(hdr, link);
+        }
```
hpatro: We don't need this change.

src/cluster_legacy.h
```diff
@@ -60,6 +60,7 @@ typedef struct clusterLink {
 #define nodeIsPrimary(n) ((n)->flags & CLUSTER_NODE_PRIMARY)
 #define nodeIsReplica(n) ((n)->flags & CLUSTER_NODE_REPLICA)
 #define nodeInHandshake(n) ((n)->flags & CLUSTER_NODE_HANDSHAKE)
+#define nodeIsMeeting(n) ((n)->flags & CLUSTER_NODE_MEET)
```
hpatro: How about `nodeInMeetProcess` / `nodeInMeetState` ?

src/cluster_legacy.c
```diff
@@ -4909,9 +4923,21 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t handshake_
     /* A Node in HANDSHAKE state has a limited lifespan equal to the
      * configured node timeout. */
     if (nodeInHandshake(node) && now - node->ctime > handshake_timeout) {
+        serverLog(LL_NOTICE, "Clusterbus handshake timeout [%s]:%d after a timeout of %lld",
+                  node->ip, node->cport, handshake_timeout);
         clusterDelNode(node);
         return 1;
     }
+    if (node->link != NULL && node->inbound_link == NULL &&
+        !nodeInHandshake(node) && !nodeIsMeeting(node) && !nodeTimedOut(node) &&
```
hpatro: Should we create a macro for this node state check? Not readable at this point.

src/cluster_legacy.c
```diff
@@ -4934,6 +4960,7 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t handshake_
             return 0;
         }
     }
+
```
hpatro: Not required.

tests/unit/cluster/cluster-reliable-meet.tcl
```diff
@@ -75,3 +78,162 @@ tags {tls:skip external:skip cluster} {
...
+            }
+            assert {[llength [R 1 CLUSTER NODES]] == 18}
+            # At this point, from node 1 point of view, the handshake with node 0 timed out.
+
+            # Allow all messages
+            R 1 DEBUG DROP-CLUSTER-PACKET-FILTER $CLUSTER_PACKET_TYPE_NONE
+            R 2 DEBUG DROP-CLUSTER-PACKET-FILTER $CLUSTER_PACKET_TYPE_NONE
+
+            # Now Node 0 will send a MEET packet to Node 1 & 2 since it has an outbound link to these nodes but no inblound link.
+            # Handshake should now complete successfully.
```
hpatro, line 271:
```suggestion
            # Now Node 0 will send a MEET packet to Node 1 & 2 since it has an outbound link to these nodes but no inbound link.
            # Handshake should now complete successfully.
```

tests/unit/cluster/cluster-reliable-meet.tcl
```diff
@@ -75,3 +78,162 @@ tags {tls:skip external:skip cluster} {
...
+            wait_for_condition 50 200 {
+                [llength [R 0 CLUSTER NODES]] == 26 &&
+                [llength [R 1 CLUSTER NODES]] == 26 &&
+                [llength [R 2 CLUSTER NODES]] == 26
```
hpatro: Could we match certain value in the string output? I don't like this magic number comparison which can change in the near future.

tests/unit/cluster/cluster-reliable-meet.tcl
```diff
@@ -75,3 +78,162 @@ tags {tls:skip external:skip cluster} {
 
 set ::singledb $old_singledb
 
+proc cluster_get_first_node_in_handshake id {
```
hpatro, line 81: Would recommend to move away from all the length comparison being made to discover the state of a node. Either utilize `CLUSTER LINKS` or do string comparison on `CLUSTER NODES` output.

hpatro (commented):
What if we disconnect the outbound link if inbound link is not available? I think it will lead to the same reconnection flow. Would it help with having simpler code and one unified flow. I'm not sure if it will perform the `MEET` operation though.

src/cluster_legacy.h
pieturin: Ah yes, I do prefer `nodeInMeetState()`.

src/cluster_legacy.c
pieturin: True, but this double if with the same condition was driving me crazy.

pieturin:
> What if we disconnect the outbound link if inbound link is not available?

In this case we would just re-open an outbound connection, which the other node will accept, but it won't force the other node to recognize us as being part of the cluster if it doesn't trust us yet. The only way to force the other node to add us to its cluster view is for us to send a MEET packet.

src/cluster_legacy.c
pieturin: `nodeInNormalState()`?

hpatro:
> > What if we disconnect the outbound link if inbound link is not available?
> 
> 
> 
> In this case we would just re-opened an outbound connection, which the other node will accept, but it won't force the other node to recognize us as being part of the cluster if it doesn't trust us yet. The only way to force the other node to add us to its cluster view is for us to send a MEET packet.

CLUSTER MEET is an admin operation but I guess we are fine with the case of reinitiating it if the operation wasn't successful in first place and retry it.

src/cluster_legacy.c
hpatro: I don't mind this. But in general we avoid changes to the lines of code not related to the PR.

tests/unit/cluster/cluster-reliable-meet.tcl
```diff
@@ -75,3 +78,199 @@ tags {tls:skip external:skip cluster} {
...
+        [cluster_get_node_by_id 2 $node2_id] != {} &&
+        [llength [R 0 CLUSTER LINKS]] == 4 &&
+        [llength [R 1 CLUSTER LINKS]] == 4 &&
+        [llength [R 2 CLUSTER LINKS]] == 4
+    } {
+        return 1
+    } else {
+        return 0
+    }
+}
```
hpatro: From ChatGPT:

```suggestion
proc cluster_nodes_all_know_each_other {num_nodes} {
    # Collect node IDs dynamically
    set node_ids {}
    for {set i 0} {$i < $num_nodes} {incr i} {
        lappend node_ids [dict get [get_myself $i] id]
    }

    # Check if all nodes know each other
    foreach node_id $node_ids {
        foreach check_node_id $node_ids {
            for {set node_index 0} {$node_index < $num_nodes} {incr node_index} {
                if {[cluster_get_node_by_id $node_index $check_node_id] == {}} {
                    return 0
                }
            }
        }
    }

    # Verify cluster link counts for each node
    set expected_links [expr {2 * ($num_nodes - 1)}]
    for {set i 0} {$i < $num_nodes} {incr i} {
        if {[llength [R $i CLUSTER LINKS]] != $expected_links} {
            return 0
        }
    }
    return 1
}
```

src/cluster_legacy.c
hpatro: `nodeInHealthyState()` ?

enjoy-binbin:
> In some cases, when meeting a new node, if the handshake times out, we can end up with an inconsistent view of the cluster where the new node knows about all the nodes in the cluster, but the cluster does not know about this new node (or vice versa).

can you elaborate where the handshake timeout happens? Is this a different issue from #461? sorry i don't read the code.

enjoy-binbin (commented):
can we do something like #461? only clear the CLUSTER_NODE_MEET flag when myself receive a "ack" (not the plain PONG but something with a strong ack, ack that sender has already meet myself?) I haven't thought about it carefully, but i feel it is more reliable?

madolson:
> only clear the CLUSTER_NODE_MEET flag when myself receive a "ack" (not the plain PONG but something with a strong ack, ack that sender has already meet myself?

Do you mean by like adding a new flag? I think the concern is we could still end up in the inverse state, where the the node that received the "strong" ack will put the other node online but then might go offline. 

My original thought was that as long as one node believes the other is part of the cluster, is should try to have the other node join. It's sort of like an "enhanced" version of how we built up the mesh when two disjoin clusters meet each other.

pieturin:
> can we do something like https://github.com/valkey-io/valkey/pull/461? only clear the CLUSTER_NODE_MEET flag when myself receive a "ack" (not the plain PONG but something with a strong ack, ack that sender has already meet myself?) I haven't thought about it carefully, but i feel it is more reliable?

We could do a 3-way handshake to strengthen the handshake reliability, instead of the current -> MEET/PONG, <- PING/PONG. We could add an extra back and forth between the two nodes before clearing the flags. But we can always end up in a situation where one node thinks the handshake is done and the other node times out the handshake because the last packet got delayed or dropped.

With this solution, if the handshake has succeeded on one side and not the other, we ensure both sides will eventually know each other.

pieturin:
With this fix, we can sometime get in a (potentially) infinite loop where a node keeps sending a MEET packet to the other node, but both nodes know each other. This sometimes (although rarely) happens in the second test `Handshake eventually succeeds after node handshake timeout on one side with inconsistent view of the cluster`.

The following sequence of events can trigger this issue:

1. Nodes 1 & 2 know each other, but don't know node 0.
2. We make node 0 meet node 1. But the handshake times-out on node 1's side, but succeed on node 0's side.
3. When node 1 marks the handshake as timed out, it will close both connections with node 0.
4. From node 0 perspective, both connections to node 1 are closed. But since it knows node 1, it will re-open an outbound connection to it.
5. Node 1 will accept the inbound connection from node 0, but it doesn't know this node, so it doesn't register this connection as belonging to any known node (ie: `link->node` stays NULL).
6. With the change from this PR, node 0 will detect that node 2 doesn't have any inbound connection to it. So node 0 will send a MEET packet to node 2. (for this bug to happen, node 2 should be met first)
7. Handshake with node 0 and 2 succeeds.
8. Now node 2 gossips about node 0 to node 1. So node 1 will add node 0 to its list of known nodes.
9. Node 1 opens a connection to node 0. At this point node 0 has both inbound and outbound connections to node 1. But from node 1's perspective, it only has an outbound connection to node 0. The inbound connection is not attached to any node (still has `link->node` set to NULL).
10. So node 1 sends a MEET packet to node 0, since it doesn't think it has an inbound connection for it.
11. The handshake completes successfully since node 0 responds to the MEET packet. But still no inbound connection.
12. So node 1 keeps sending MEET packets to node 0 until node 0 sends a PING packet to node 1. When node 1 receives a PING packet from the node 0, it will set the node's inbound connection ([here](https://github.com/valkey-io/valkey/blob/33f42d7fb597ce28040f184ee57ed86d6f6ffbd8/src/cluster_legacy.c#L2928)).

Node 0 should *eventually* send a PING packet to node 1, but there is no guarantee as to when that can happen. When I reproduce the issue, node 0 never gets a chance to send a PING to node 1 because node 0 overrides `node->pong_received` for node 1 when node 2 gossips about node 1 with a higher `pong_received` value. And node 2 always has a lower `pong_received` value compared to node 1 when trying to select a node to send a PING to here:
https://github.com/valkey-io/valkey/blob/33f42d7fb597ce28040f184ee57ed86d6f6ffbd8/src/cluster_legacy.c#L5045

I think there are various ways to mitigate this issue:

1. Do nothing, the MEET packet loop should eventually stop. (but I have to update the test so that it's not flaky)
2. Change the pinging decision logic to force a PING to every nodes at least every X amount of time, even if we know that another node was able to ping it recently. X can be set to something like `2 * (server.cluster_ping_interval ? server.cluster_ping_interval : server.cluster_node_timeout / 2)`. This will incur an increase in cluster bus traffic on large clusters.
3. If I receive a MEET packet and I already have an outbound link for that node, then I should free my existing outbound link to it, and re-open a new one.

I think option 3 should work best without making too many changes to the current logic. But I'm open to suggestions.

pieturin:
`tests/unit/expire.tcl` is currently failing in unstable too. I'll rebase once there is a fix.

enjoy-binbin:
https://github.com/valkey-io/valkey/pull/1368 the expire test is fixed

src/cluster_legacy.c
```diff
@@ -3205,33 +3214,44 @@ int clusterProcessPacket(clusterLink *link) {
...
+                 * the gossip section here since we have to trust the sender because
+                 * of the message type. */
+                clusterProcessGossipSection(hdr, link);
+            } else if (sender->link && now - sender->ctime > server.cluster_node_timeout) {
+                /* The MEET packet is from a known node, so the sender thinks that I do not know it.
+                 * Freeing my outbound link to that node, to force a reconnect and sending a PING.
+                 * Once that node receives our PING, it should recognize the new connection as an inbound link from me.
+                 * We should only free the outbound link if the node is known for more time than the handshake timeout,
+                 * since during this time, the other side might still be trying to complete the handshake. */
+                serverAssert(link != sender->link); // we should always receive a MEET packet on an inbound link
```
hpatro: We generally avoid single line comment.

src/cluster_legacy.c
```diff
@@ -3205,33 +3214,44 @@ int clusterProcessPacket(clusterLink *link) {
...
+                }
+                setClusterNodeToInboundClusterLink(node, link);
+                clusterAddNode(node);
+                clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG);
+
+                /* If this is a MEET packet from an unknown node, we still process
+                 * the gossip section here since we have to trust the sender because
+                 * of the message type. */
+                clusterProcessGossipSection(hdr, link);
+            } else if (sender->link && now - sender->ctime > server.cluster_node_timeout) {
```
hpatro, line 3243: Shouldn't we still process the gossip section of the new message to get the latest information?

hpatro:
@pieturin Could you update the top comment as well about the exact behavior change?

src/cluster_legacy.c
pieturin: We do that below for all packets sent from a known sender: https://github.com/valkey-io/valkey/blob/349bc7547bf28ae304537bd6888d575e2409f25c/src/cluster_legacy.c#L3505

hpatro (approved):
LGTM. But I would like someone else as well to look at it before merging this in.

pieturin:
Updated the PR description.

pieturin:
@PingXie, @madolson, @enjoy-binbin, any comments on this PR?

enjoy-binbin (approved):
Top comment and code LGTM, did not review the tests.

src/cluster_legacy.c
```diff
@@ -3926,6 +3952,11 @@ void clusterSetGossipEntry(clusterMsg *hdr, int i, clusterNode *n) {
 /* Send a PING or PONG packet to the specified node, making sure to add enough
  * gossip information. */
 void clusterSendPing(clusterLink *link, int type) {
+    serverLog(LL_DEBUG, "Sending %s packet to node %.40s (%s)",
+              clusterGetMessageTypeString(type),
+              link->node ? link->node->name : "<unknown>",
+              link->inbound ? "inbound" : "outbound");
```
madolson: This is hundreds of messages per second being written into the log for large clusters. I'm not sure this is necessarily a good idea, as we sometimes enable this on production systems for debugging.

madolson: `%.40s (%s)` is also usually the convention for the human readable nodename in (), which isn't true here, you are using inbound/outbound, I would prefer  to include human nodename as well as the inbound outbound information.

madolson (approved):
I suppose the comment isn't really a blocker for me, it's just about documentation.

src/cluster_legacy.c
pieturin: We also log on every received cluster packet at the debug level, so I thought it wouldn't be much of a difference. And I added it because it was useful to me to debug this issue. But I can remove it. If I leave it, I can update to add the human readable nodename.

src/cluster_legacy.c
```diff
@@ -3205,33 +3214,48 @@ int clusterProcessPacket(clusterLink *link) {
...
+                 * know it.
+                 * Freeing my outbound link to that node, to force a reconnect and sending a PING.
+                 * Once that node receives our PING, it should recognize the new connection as an inbound link from me.
+                 * We should only free the outbound link if the node is known for more time than the handshake timeout,
+                 * since during this time, the other side might still be trying to complete the handshake. */
+
+                /* We should always receive a MEET packet on an inbound link. */
+                serverAssert(link != sender->link);
+                serverLog(LL_NOTICE, "Freeing outbound link to node %.40s after receiving a MEET packet from this known node",
+                          sender->name);
```
PingXie, line 3254: we now always log both `name` and `human_nodename`.

```suggestion
                serverLog(LL_NOTICE, "Freeing outbound link to node %.40s (%s) after receiving a MEET packet from this known node",
                          sender->name.
                          sender->human_nodename);
```

src/cluster_legacy.c
```diff
@@ -3205,33 +3214,48 @@ int clusterProcessPacket(clusterLink *link) {
...
+                }
+                setClusterNodeToInboundClusterLink(node, link);
+                clusterAddNode(node);
+                clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG);
+
+                /* If this is a MEET packet from an unknown node, we still process
+                 * the gossip section here since we have to trust the sender because
+                 * of the message type. */
+                clusterProcessGossipSection(hdr, link);
+            } else if (sender->link && now - sender->ctime > server.cluster_node_timeout) {
```
PingXie, line 3243: should we align with the handshake timeout value computed in `clusterCron`, which is the max of `cluster_node_timeout` and 1s.

Also consider abstracting this check to a higher level construct? `nodeHandshakeTimedout` perhaps? This handshake timeout primitive is used in `clusterNodeCronHandleReconnect` too.

src/cluster_legacy.c
```diff
@@ -3205,33 +3214,48 @@ int clusterProcessPacket(clusterLink *link) {
...
+                /* The MEET packet is from a known node, after the handshake timeout, so the sender thinks that I do not
+                 * know it.
+                 * Freeing my outbound link to that node, to force a reconnect and sending a PING.
```
PingXie, line 3246: clarifying the PING will be sent over the new outbound link?
```suggestion
                 * Free my outbound link to that node, triggering a reconnect and a PING over the new link.
```

src/cluster_legacy.c
PingXie: +1 on logging both `name` and `human_nodename` using `%.40s (%s)`.  Can you please give all the new log lines a once-over?

src/cluster_legacy.c
```diff
@@ -4941,6 +4972,15 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t handshake_
         clusterDelNode(node);
         return 1;
     }
+    if (node->link != NULL && node->inbound_link == NULL && nodeInNormalState(node) &&
```
PingXie, line 4976: nit
```suggestion

    if (nodeInNormalState(node) && node->link != NULL && node->inbound_link == NULL && 
```

madolson (approved):

PingXie (commented):
Forgot to send out my (partial) review :) 

Also I don't think I fully understand the problem that we are trying to solve and why the fix works.

PingXie:
> Also I don't think I fully understand the problem that we are trying to solve 

Do I understand correctly that we are trying to solve two problems here?

1. inconsistent cluster topology within the existing node. 

More specifically, a new node that hasn't gone through the full handshake process can still get gossiped around the cluster and then picked up by existing nodes that are not involved in the handshake process? 

* existing nodes in a cluster: `a`, `b`, `c`
* new node: `A`
* `a` is instructed to meet `A`.
*  the handshake process never completes but `A`'s `PONG` response to `a`'s `MEET` should make it to `a`
* `a` then gossips `A`'s id to `b` and `c`, which then add `A` to their cluster view - because the server doesn't check `A`'s MEET state (`CLUSTER_STATE_MEET`) today.
* now `a` thinks `A` is in the MEET state but `b` and `c` think `A` is normal and `A` sees no other node.

2. `cluster meet` reliability 

If `A` fails to reply `PONG` to `a`'s `MEET`, both `a` and `A` should eventually remove each other from its cluster view respectively and `A` will never make to `b` and `c`. In this sense, the existing cluster, made up of `a`, `b`, and `c`, is still consistent. 

> and why the fix works.

The fix for the first problem makes sense to me: resending `MEET` after the handshake timeout.

I am not sure how/if the second problem is addressed by this PR? if both `a` and `A` remove each other from their cluster view respectively, how can the handshake process be restarted?

pieturin:
> I am not sure how/if the second problem is addressed by this PR? if both a and A remove each other from their cluster view respectively, how can the handshake process be restarted?

If the handshake times out on both side of the meet, then, this fix won't do anything. This fix is only meant to make sure we never end up with an inconsistent view of the cluster (where one side knows the other nodes, but not the other side). This makes the meet handshake more reliable but not fool proof.
We could revisit the proposal to make MEET synchronous or sticky (see: https://github.com/redis/redis/issues/11095), which would solve the second problem.

The inconsistent view of the cluster can happen in two different ways (that I could see):
- The handshake succeeds on one node but times out on the other node. This is tested by the test `Handshake eventually succeeds after node handshake timeout on one side with inconsistent view of the cluster`.
```
            # Node 0 -- MEET -> Node 1
            # Node 0 <- PONG -- Node 1
            # Node 0 <- PING -- Node 1 [Node 0 will mark the handshake as successful]
            # Node 0 -- PONG -> Node 1 [we drop this message, so node 1 will eventually mark the handshake as timed out]
```
- The handshake times out on both sides, but the new node learns about other nodes of the existing cluster through the gossip section in the MEET packet. In this case, even if the handshake times out, the new node will know some of the nodes from the existing cluster, but they will not know this new node. This is tested by the test `Handshake eventually succeeds after node handshake timeout on both sides with inconsistent view of the cluster`.
```
            # Node 1 -- MEET -> Node 0 [Node 0 might learn about Node 2 from the gossip section of the msg]
            # Node 1 <- PONG -- Node 0 [we drop this message, so Node 1 will eventually mark the handshake as timed out]
            # Node 1 <- PING -- Node 0 [we drop this message, so Node 1 will never send a PONG and Node 0 will eventually mark the handshake as timed out]
```

> - a then gossips A's id to b and c, which then add A to their cluster view - because the server doesn't check A's MEET state (CLUSTER_STATE_MEET) today.

We don't gossip a node if it's in HANDSHAKE state. See: https://github.com/valkey-io/valkey/blob/5f7fe9ef21f1feae42257ac93ab33d8f8c06e97f/src/cluster_legacy.c#L4039-L4048

I'll prepare a follow-up PR to address your comments @PingXie, thanks for the review!

src/cluster_legacy.c
pieturin: I didn't think it was worth doing, but I can create a function for it, I think that's better.

PingXie:
>If the handshake times out on both side of the meet, then, this fix won't do anything.

make sense

> We don't gossip a node if it's in HANDSHAKE state. 

Right but in this case `A` is in the `MEET` state from `a`'s perspective so it gets picked/gossiped.

> I'll prepare a follow-up PR to address your comments 

Thanks. Will TAL.

pieturin:
> Right but in this case A is in the MEET state from a's perspective so it gets picked/gossiped.

Ah yes, you're right, this is what should happen:

```
            # Node a -- MEET -> Node A [Node A has flags HANDSHAKE|MEET]
            # Node a <- PONG -- Node A [After receiving this packet, Node A has flag MEET]
            # Node a <- PING -- Node A [After receiving this packer, we clear the MEET flag for Node A]
            # Node a -- PONG -> Node A
```

> now a thinks A is in the MEET state but b and c think A is normal and A sees no other node.

This is correct that `a` thinks `A` is in MEET state, and `b` and `c` can get gossip information about it. But `A` might know about `b` and `c`, from the gossip section of the MEET packet from `a`.

PR #1312, opened 2024-11-15
Extract the scripting engine code from the functions unit

### LINKED ISSUE #1470: Support Detailed Compilation Errors by Allowing Scripting Engine to Return `robj` Instead of C Strings
> I think the more generic approach is to allow the scripting engine to return a C string with the error message.

Yeah if we would like to provide detailed compilation errors then we need to keep this capability but I would suggest `robj` too over C strings.

_Originally posted by @PingXie in https://github.com/valkey-io/valkey/pull/1277#discussion_r1894844063_

rjd15372:
This commit creates a new compilation unit for the scripting engine code by extracting the existing code from the functions unit.
We're doing this refactor to prepare the code for running the `EVAL` command using different scripting engines.

This PR has a module API change: we changed the type of error messages returned by the callback  `ValkeyModuleScriptingEngineCreateFunctionsLibraryFunc` to be a `ValkeyModuleString` (aka `robj`);

This PR also fixes #1470.


hpatro:
@rjd15372 You could point the merge branch as your initial branch. It will show only the diff then.

rjd15372:
@hpatro 
> @rjd15372 You could point the merge branch as your initial branch. It will show only the diff then.

I don't know if that is possible because the branch that I used to open #1277 is from my own fork of the valkey repo.

rjd15372:
@zuiderkwast @PingXie this refactor is ready for review.

PingXie (commented):
I took a quick look. I'd suggest renaming "engine" to "scripting engine" globally first. We have been using "engine", "core engine", and "Valkey server" interchangeably in our daily conversations, documentation, and the code. We should qualify this new "engine" as "scripting engine" to avoid confusion.

rjd15372:
> I took a quick look. I'd suggest renaming "engine" to "scripting engine" globally first. We have been using "engine", "core engine", and "Valkey server" interchangeably in our daily conversations, documentation, and the code. We should qualify this new "engine" as "scripting engine" to avoid confusion.

Sounds good :+1:

cmake/Modules/SourceFiles.cmake
```diff
@@ -106,7 +106,8 @@ set(VALKEY_SERVER_SRCS
     ${CMAKE_SOURCE_DIR}/src/connection.c
     ${CMAKE_SOURCE_DIR}/src/unix.c
     ${CMAKE_SOURCE_DIR}/src/server.c
-    ${CMAKE_SOURCE_DIR}/src/logreqres.c)
+    ${CMAKE_SOURCE_DIR}/src/logreqres.c
+    ${CMAKE_SOURCE_DIR}/src/scripting_engine.c)
```
zuiderkwast: Add it next to the related script.c and functions.c lines?

src/function_lua.c
```diff
@@ -121,7 +122,7 @@ static compiledFunction **luaEngineCreate(ValkeyModuleCtx *module_ctx,
                                           const char *code,
                                           size_t timeout,
                                           size_t *out_num_compiled_functions,
-                                          char **err) {
+                                          robj **err) {
```
zuiderkwast, line 125: Is this #1470?

src/scripting_engine.h
```diff
@@ -0,0 +1,72 @@
+#ifndef _ENGINE_H_
+#define _ENGINE_H_
```
zuiderkwast: SCRIPTING_ENGINE_H

src/scripting_engine.h
```diff
@@ -0,0 +1,72 @@
...
+typedef int (*engineIterCallback)(scriptingEngine *engine, void *context);
+
+/*
+ * Engine manager API functions.
+ */
+int engineManagerInit(void);
+size_t engineManagerGetCacheMemory(void);
+size_t engineManagerGetNumEngines(void);
+size_t engineManagerGetMemoryUsage(void);
+int engineManagerRegisterEngine(const char *engine_name,
```
zuiderkwast: I suggest prefixing the names with `scriptingEngine`.

Maybe we can avoid some repetition of "engine" and make the "manager" part be implicit, at least in some cases? Isn't it clear enough from the parameters that a function is for one engine or for the entire subsystem (manager)?

Idea:

```suggestion
int scriptingEngineRegister(const char *engine_name,
```

zuiderkwast (commented):
This looks pretty good. Can you list the changes in the PR top comment, in particular the module API changes?

src/function_lua.c
rjd15372: Yes

src/scripting_engine.c
```diff
@@ -0,0 +1,284 @@
...
+ */
+scriptingEngine *scriptingEngineManagerFind(sds engine_name) {
+    dictEntry *entry = dictFind(engineMgr.engines, engine_name);
+    if (entry) {
+        return dictGetVal(entry);
+    }
+    return NULL;
+}
+
+sds scriptingEngineGetName(scriptingEngine *engine) {
```
PingXie, line 174: it would be ideal if we express the "constness" of the returned sds string explicitly. Similar callouts to other helper functions returning plain pointers. That said, since we have never enforced "constness" in the past in our code base I suspect this could turn into quite a code churn. So definitely not a blocker for this PR (and it is primarily for code hygiene).

```
typedef const char* sds_const; // in src/sds.h
```
```suggestion
sds_const scriptingEngineGetName(scriptingEngine *engine) {
```

src/functions.c
```diff
@@ -1076,29 +949,23 @@ void functionFreeLibMetaData(functionsLibMetaData *md) {
     if (md->engine) sdsfree(md->engine);
 }
 
-static void freeCompiledFunctions(engineInfo *ei,
+static void freeCompiledFunctions(scriptingEngine *engine,
                                   compiledFunction **compiled_functions,
                                   size_t num_compiled_functions,
                                   size_t free_function_from_idx) {
-    setupEngineModuleCtx(ei, NULL);
```
PingXie, line 1083: We now call `setupEngineModuleCtx` for every function freed as opposed to one per engine. Maybe this is a non-issue, if we don't expect a large number of functions in a library?

src/scripting_engine.c
```diff
@@ -0,0 +1,284 @@
...
+ *
+ * The `context` pointer is also passed in each callback call.
+ */
+void scriptingEngineManagerForEachEngine(engineIterCallback callback,
+                                         void *context) {
+    dictIterator *iter = dictGetIterator(engineMgr.engines);
+    dictEntry *entry = NULL;
+    while ((entry = dictNext(iter))) {
+        scriptingEngine *e = dictGetVal(entry);
+        if (!callback(e, context)) {
```
PingXie: it is a bit odd to short-circuit the loop when one callback failed. also I don't see any callback needing to return any value. I wonder if we should make the callback return `void` for now until there is a true requirement. this way, we can skip the error handling strategy discussion.

src/server.c
```diff
@@ -2875,6 +2876,11 @@ void initServer(void) {
         server.maxmemory_policy = MAXMEMORY_NO_EVICTION;
     }
 
+    if (scriptingEngineManagerInit() == C_ERR) {
+        serverPanic("Scripting engine manager initialization failed, check the server logs.");
+        exit(1);
```
PingXie: nit - we don't need to `exit(1)` after `serverPanic`.

PingXie (commented):
Overall LGTM. Will give it once-over once the active comments are resolved. Thanks @rjd15372!

src/scripting_engine.c
zuiderkwast: I like this idea! The const-sds type would be useful in some places where we now have to cast away the const modifier. Even if we don't introduce it here, we can do it in another PR. @SoftlyRaining I know you have run into this in another PR.

rainsupreme: I do like this! There have been various places I would like to have used it :)

src/functions.c
rjd15372: I don't expect a large number of functions, and also this code is in the `FUNCTION LOAD` code path, which is not perf sensitive.

src/scripting_engine.c
rjd15372: Sure, sounds good.

src/scripting_engine.c
rjd15372: I agree that it would be nice have this constness property for sds strings. 
I suggest to open an issue for doing this improvement, because from a quick look at the code, it will require changes to code that is not related to this feature.

src/functions.c
zuiderkwast: Either way, this is something we can change later, right? It's not something that modules see, so it doesn't need backward compatibility?

src/scripting_engine.c
```diff
@@ -0,0 +1,282 @@
...
+    engineCtx *ctx;
+
+    /* Callback functions implemented by the scripting engine module */
+    engineMethods methods;
+} scriptingEngineImpl;
+
+typedef struct scriptingEngine {
+    sds name;                    /* Name of the engine */
+    ValkeyModule *module;        /* the module that implements the scripting engine */
+    scriptingEngineImpl *impl;   /* engine callbacks that allows to interact with the engine */
```
zuiderkwast: Why is `impl` a separate allocation? It seems to be just a grouping of engineCtx and engineMethods. Why not do it without a separate allocation?

```suggestion
    scriptingEngineImpl impl;    /* engine context and callbacks to interact with the engine */
```

src/scripting_engine.c
```diff
@@ -0,0 +1,282 @@
...
+ * that are loaded in the server and implemented by Valkey Modules.
+ *
+ * Returns C_ERR if some error occurs during the initialization.
+ */
+int scriptingEngineManagerInit(void) {
+    engineMgr.engines = dictCreate(&engineDictType);
+    return C_OK;
+}
+
+size_t scriptingEngineManagerGetCacheMemory(void) {
```
zuiderkwast: What's the cache memory of an engine? I don't think it's common knowledge. A comment would be useful.

src/scripting_engine.c
```diff
@@ -0,0 +1,282 @@
...
+ * Returns C_ERR in case of an error during registration.
+ */
+int scriptingEngineManagerRegister(const char *engine_name,
+                                   ValkeyModule *engine_module,
+                                   engineCtx *engine_ctx,
+                                   engineMethods *engine_methods) {
+    sds engine_name_sds = sdsnew(engine_name);
+
+    if (dictFetchValue(engineMgr.engines, engine_name_sds)) {
+        serverLog(LL_WARNING, "Same engine was registered twice");
```
zuiderkwast: Log the engine name too?

src/functions.c
PingXie: agreed. we don't need to optimize proactively.

src/scripting_engine.c
PingXie: separate PR dedicated to "constness" is better IMO.

PingXie: filed #1542.

zuiderkwast (commented):
I didn't finish reviewing but I have some pending comments since last week. Posting them now.

I think this is close to ready to merge.

src/scripting_engine.c
rjd15372: Yeah, you're right.

src/scripting_engine.c
rjd15372: I was using the same name as the original field (before the refactoring), but actually this should be called something else because it's storing the sum of the memory overhead of each registered scripting engine.

zuiderkwast (approved):
Let's get this merged so we can move forward with EVAL PR.

We need to finalize the API in time before 8.1.

This PR has a module API change: the error message as ValkeyModuleString in the callback types. I guess it's worth mentioning in the PR top comment.

rjd15372:
@zuiderkwast thanks. I updated the PR description.

src/scripting_engine.c
PingXie: now that the `const_sds` pr is merged. let's make sure we make good use of it. separate pr is also fine.

PingXie (approved):
Thanks @rjd15372!

PR #502, opened 2024-05-15
Adjust sds types

poiuj:
sds type should be determined based on the size of the underlying buffer, not the logical length of the sds. Currently we truncate the alloc field in case buffer is larger than we can handle. It leads to a mismatch between alloc field and the actual size of the buffer. Even considering that alloc doesn't include header size and the null terminator.

It also leads to a waste of memory with jemalloc. For example, let's consider creation of sds of length 253. According to the length, the appropriate type is SDS_TYPE_8. But we allocate `253 + sizeof(struct sdshdr8) + 1` bytes, which sums to 257 bytes. In this case jemalloc allocates buffer from the next size bucket. With current configuration on Linux it's 320 bytes. So we end up with 320 bytes buffer, while we can't address more than 255.

The same happens with other types and length close enough to the appropriate powers of 2.

The downside of the adjustment is that with allocators that do not allocate larger than requested chunks (like GNU allocator), we switch to a larger type "too early". It leads to small waste of memory. Specifically: sds of length 31 takes 35 bytes instead of 33 (2 bytes wasted) sds of length 255 takes 261 bytes instead of 259 (2 bytes wasted) sds of length 65,535 takes 65,545 bytes instead of 65,541 (4 bytes wasted) sds of length 4,294,967,295 takes 4,294,967,313 bytes instead of 4,294,967,305 (8 bytes wasted)

PR #1314, opened 2024-11-16
Fix sds unittest tests to check for zmalloc_usable_size

enjoy-binbin:
s_malloc_size == zmalloc_size, currently sdsAllocSize does not
calculate PREFIX_SIZE when no malloc_size available, this casue
test_typesAndAllocSize fail in the new unittest, what we want to
check is actually zmalloc_usable_size.

### DIFF
--- a/src/unit/test_sds.c
+++ b/src/unit/test_sds.c
@@ -252,10 +252,6 @@ int test_sds(int argc, char **argv, int flags) {
     return 0;
 }
 
-size_t sdsRealAllocSize(sds s) {
-    return sdsAllocSize(s) + PREFIX_SIZE;
-}
-
 int test_typesAndAllocSize(int argc, char **argv, int flags) {
     UNUSED(argc);
     UNUSED(argv);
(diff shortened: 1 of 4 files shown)


enjoy-binbin:
I guess we don't want to modify sdsAllocSize to include it? so i just modify the test.
Also see #502 and #694 for more details.

The failure test CI: https://github.com/valkey-io/valkey/actions/runs/11865335400/job/33070348096#step:7:254

src/unit/test_sds.c
```diff
@@ -252,50 +252,55 @@ int test_sds(int argc, char **argv, int flags) {
     return 0;
 }
 
+size_t sdsRealAllocSize(sds s) {
+    return sdsAllocSize(s) + PREFIX_SIZE;
+}
+
```
madolson: From a modularization perspective, we could add this as a function in zmalloc.c and forward declare it here to access it for tests but not expose it in the header file. I don't know if I prefer that approach though.

madolson (approved):
I'm okay with this to get the tests passing regardless, I feel like there is probably a better way to expose the API to the testing though.

src/unit/test_sds.c
```diff
@@ -252,50 +252,55 @@ int test_sds(int argc, char **argv, int flags) {
...
 
     sds x = sdsnewlen(NULL, 31);
     TEST_ASSERT_MESSAGE("len 31 type", (x[-1] & SDS_TYPE_MASK) == SDS_TYPE_5);
+    TEST_ASSERT_MESSAGE("len 31 sdsAllocSize", sdsRealAllocSize(x) == s_malloc_size(sdsAllocPtr(x)));
     sdsfree(x);
 
     x = sdsnewlen(NULL, 32);
     TEST_ASSERT_MESSAGE("len 32 type", (x[-1] & SDS_TYPE_MASK) >= SDS_TYPE_8);
-    TEST_ASSERT_MESSAGE("len 32 sdsAllocSize", sdsAllocSize(x) == s_malloc_size(sdsAllocPtr(x)));
+    TEST_ASSERT_MESSAGE("len 32 sdsAllocSize", sdsRealAllocSize(x) == s_malloc_size(sdsAllocPtr(x)));
```
zuiderkwast: s_malloc_size = zmalloc_size, but what we want to check is zmalloc_usable_size, right?

We can add a s_malloc_usable_size in sdsalloc.h?

```suggestion
    TEST_ASSERT_MESSAGE("len 32 sdsAllocSize", sdsAllocSize(x) == s_malloc_usable_size(sdsAllocPtr(x)));
```

src/zmalloc.h
```diff
@@ -168,6 +168,17 @@ size_t zmalloc_usable_size(void *ptr);
  * obtained from z[*]_usable() family functions, there is no need for this step. */
 #define zmalloc_usable_size(p) zmalloc_size(p)
 
+#ifdef HAVE_MALLOC_SIZE
+#define PREFIX_SIZE 0
```
zuiderkwast: If we expose this in the zmalloc api, then we should prefix it like ZMALLOC_PREFIX_SIZE, but I don't think we nees to expose it.

We already have zmalloc_size vs zmalloc_usable_size.

src/unit/test_sds.c
enjoy-binbin: so damn right, what we want to check is actually zmalloc_usable_size.

zuiderkwast (approved):

PR #1317, opened 2024-11-18
Allow `MEMORY MALLOC-STATS` and `MEMORY PURGE` during loading phase

### LINKED ISSUE #1299: MEMORY DOCTOR/MALLOC-STATS in loading phase

Sometimes I wish to investigate memory issues during the loading phase. However I get:
```
socket_redis> memory malloc-stats
LOADING Redis is loading the dataset in memory
```

I would like to see the memory details provided by these commands during loading

**Additional information**

The check seems to be here https://github.com/valkey-io/valkey/blob/unstable/src/server.c#L4176-L4181:
```
    /* Loading DB? Return an error if the command has not the
     * CMD_LOADING flag. */
    if (server.loading && !server.async_loading && is_denyloading_command) {
        rejectCommand(c, shared.loadingerr);
        return C_OK;
    }
```

The question: Is there specific reasons those commands are disallowed during loading?

If not, is it as simple as adding CMD_LOADING flag to all the MEMORY commands, or there are considerations to take into account for making this work?

knggk:
- Enable investigation of memory issues during loading
- Previously, all memory commands were rejected with LOADING error (except memory help)
- `MEMORY MALLOC-STATS` and `MEMORTY PURGE` are now allowed as they don't depend on the dataset
- `MEMORY STATS` and `MEMORY USAGE KEY` remain disallowed

Fixes #1299


hwware:
I would like to suggest only 3 commands except "memory stats" to be allowed during loading process because it costs a lot of resources and involves key issues. (key count, key used memory)

tests/unit/loading.tcl
```diff
@@ -0,0 +1,38 @@
...
+        $rd memory malloc-stats
+        $rd memory stats
+        $rd memory help
+        # Memory usage on key while loading is not well defined -> keep error
+        $rd memory usage key:1
+        $rd memory purge
+        $rd ping
+
+        assert_match {Hi Sam, *} [$rd read]
+        assert_match {*} [$rd read]
```
hpatro: Could we do some partial match of string value here?

tests/unit/loading.tcl
```diff
@@ -0,0 +1,38 @@
...
+        assert_match {peak.allocated *} [$rd read]
+        assert_match {{MEMORY <subcommand> *}} [$rd read]
+        # Memory usage keeps getting rejected in loading because the dataset is not visible
+        assert_error {*LOADING*} {$rd read}
+        assert_match OK [$rd read]
+        assert_error {*LOADING*} {$rd read}
```
hpatro: Could we add comment of the actual command above the response, would be easier to read.

tests/unit/loading.tcl
knggk: Sounds good. I am now matching on `*alloc*`. It matches both `Stats not supported for the current allocator` when on Mac, and the actually jemalloc output of interest when on Linux.

tests/unit/loading.tcl
knggk: Agree, have updated as per your recommendation.

tests/unit/loading.tcl
```diff
@@ -0,0 +1,42 @@
+start_server [list overrides [list "key-load-delay" 50 loading-process-events-interval-bytes 1024]] {
```
enjoy-binbin: The test is failing, i guess we may need a external skip flag.
```suggestion
start_server [list overrides [list "key-load-delay" 50 loading-process-events-interval-bytes 1024] tags [list "external:skip"]] {
```

enjoy-binbin (changes requested):
LGTM, somehow i think we can put the test in introspection.tcl

enjoy-binbin:
please also update the top comment, we can copy the issue content instead of just linking it.

tests/unit/loading.tcl
enjoy-binbin: i commited it so that we can see if the CI will be happy or not.

knggk:
> please also update the top comment, we can copy the issue content instead of just linking it.

I updated the title and top comment.

tests/unit/loading.tcl
knggk: It looks like CI is still unhappy but on unrelated tests?

knggk:
> LGTM, somehow i think we can put the test in introspection.tcl

Moved the test into introspection.tcl. Let's see if it passes on the CI.

enjoy-binbin:
https://github.com/valkey-io/valkey/pull/1368 The expire test is fixed

hpatro (approved):

hwware (approved):
Everything looks good, @enjoy-binbin If you agree, pls merge it.

knggk:
> #1368 The expire test is fixed

I merged unstable, and the tests now pass. Thank you!

enjoy-binbin (approved):
LGTM. @valkey-io/core-team please approval

zuiderkwast (approved):
Change seems good. Didn't review the diff. Is it a major decision?

madolson (approved):
@zuiderkwast I think its technically a major decision based on past decisions, since it's altering command behavior. Moving forward maybe it shouldn't be though.

PR #1320, opened 2024-11-19
Receipt of REPLCONF VERSION reply should be triggered by event

enjoy-binbin:
This add the missing return when repl_state change to RECEIVE_VERSION_REPLY,
this way we won’t be blocked if the primary doesn’t reply with REPLCONF
VERSION.

In practice i guess this is no likely to block in this context, reading
small responses are are likely to be received in one packet, so this
is just a cleanup (consistent with the previous state machine processing).

Also update the state machine diagram to mention the VERSION reply.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -3379,15 +3379,15 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
  * ┌────────▼──────────┐        │ │  │DUAL_CHANNEL_RECEIVE_ENDOFF│     │    │by the primary      │
  * │RECEIVE_IP_REPLY   │        │ │  └───────┬───────────────────┘     │ ┌──▼────────────────┐   │
  * └────────┬──────────┘        │ │          │$ENDOFF                  │ │RECEIVE_PSYNC_REPLY│   │
- *          │                   │ │          ├─────────────────────────┘ └──┬────────────────┘   │
- *          │                   │ │          │                              │+CONTINUE           │
- *          │                   │ │  ┌───────▼───────────────┐           ┌──▼────────────────┐   │
- *          │                   │ │  │DUAL_CHANNEL_RDB_LOAD  │           │TRANSFER           │   │
+ *          │+OK                │ │          ├─────────────────────────┘ └──┬────────────────┘   │
+ * ┌────────▼──────────┐        │ │          │                              │+CONTINUE           │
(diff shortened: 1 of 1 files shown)


zuiderkwast (commented):
So we have to wait for the read handler to run again for each of the replies?

zuiderkwast:
There's a state machine diagram in a comment in this file too. Maybe we should update it too.

```
 * Replica state machine * 
 ┌───────────────────┐     
 │RECEIVE_PING_REPLY │     
 └────────┬──────────┘     
          │+PONG           
 ┌────────▼──────────┐     
 │SEND_HANDSHAKE     │     
 └────────┬──────────┘     
          │                
 ┌────────▼──────────┐     
 │RECEIVE_AUTH_REPLY │     
 └────────┬──────────┘     
          │+OK             
 ┌────────▼──────────┐     
 │RECEIVE_PORT_REPLY │     
 └────────┬──────────┘     
          │+OK             
 ┌────────▼──────────┐     
 │RECEIVE_IP_REPLY   │     
 └────────┬──────────┘     
          │                
          │                
          │                
          │                
          │+OK             
 ┌────────▼──────────┐     
 │RECEIVE_CAPA_REPLY │     
 └────────┬──────────┘     
          │                
 ┌────────▼───┐            
 │SEND_PSYNC  │            
 └─┬──────────┘            
```

zuiderkwast (approved):

enjoy-binbin:
> So we have to wait for the read handler to run again for each of the replies?

yean, look like the answer is yes. (in happy path we don't have to since we pipeline the write, and we can read the reply without any block)

zuiderkwast:
> > So we have to wait for the read handler to run again for each of the replies?
>
> yean, look like the answer is yes. (in happy path we don't have to since we pipeline the write, and we can read the reply without any block)

But we `return` after each read state. To read the next reply, we wait for read handler.

zuiderkwast (approved):
Beautiful

enjoy-binbin:
> But we return after each read state. To read the next reply, we wait for read handler.

yes, what i want to say is the "ideally path". In current code, we do need to wait for the next read handler.

zuiderkwast:
> > But we return after each read state. To read the next reply, we wait for read handler.
> 
> yes, what i want to say is the "ideally path". In current code, we do need to wait for the next read handler.

OK, if sync is slow to start, we can optimize it later. I don't think it's too slow.

PR #1324, opened 2024-11-19
Set Command with IFEQ Support

### LINKED ISSUE #1215: [NEW] Support Check and Set functionality

Allow Valkey user to perform conditional update to a key based on certain condition like matches old value/checksum. This should ideally be supported for string data type via SET command.

API changes in SET command:

```
SET <key> <value> IFEQ <comparison-value>
```

Behavior change:

Where the command is executed, IFF the comparison-value equals the current value in the key the rest of the code will be executed. If the values aren't the same, it returns a new error -NOTEQUAL. Alternative: Return a (nil) instead. We'll decide between these two cases in the PR. https://github.com/valkey-io/valkey/issues/1215#issuecomment-2458100191

Memcached CAS: 

https://docs.memcached.org/protocols/basic/#cas 

Past discussions:

https://groups.google.com/g/redis-db/c/a4zK2k1Lefo
https://github.com/redis/redis/pull/8361
https://github.com/redis/redis/issues/12485

**Alternatives considered**

Use LUA scripting

sarthakaggarwal97:
This PR allows the Valkey users to perform conditional updates where the SET command is completed if the given comparison-value matches the key’s current value.

Syntax:

```
SET key value IFEQ comparison-value
```

Behavior:

If the values match, the SET completes as expected. If they do not match, the command returns a (nil), except if the GET argument is also given (see below).

Behavior with Additional Flags:

1.  ```SET key value IFEQ comparison-value GET``` returns the existing value, regardless of whether it matches comparison-value or not. The conditional set operation is performed if the given comparison value matches the existing value. To check if the SET succeeded, the caller needs to check if the returned string matches the comparison-value.
2. ```SET key value IFEQ comparison-value XX``` is a syntax error.
3.  ```SET key value IFEQ comparison-value NX``` is a syntax error.

Closes: #1215


tests/unit/type/string.tcl
```diff
@@ -582,6 +582,19 @@ if {[string match {*jemalloc*} [s mem_allocator]]} {
       set err1
     } {*WRONGTYPE*}
 
+    test "SET with IFEQ conditional" {
```
hpatro, line 585: Could you also add test case for interaction with other parameters?

src/t_string.c
```diff
@@ -102,6 +104,15 @@ void setGenericCommand(client *c,
 
     found = (lookupKeyWrite(c->db, key) != NULL);
 
+    /* Handle the IFEQ conditional check */
+    if ((flags & OBJ_SET_IFEQ) && found) {
+        robj *current_value = lookupKeyRead(c->db, key);
+        if (current_value == NULL || compareStringObjects(current_value, comparison) != 0) {
```
hpatro: What happens if the value currently stored isn't a STRING data type? I also think we should return an error message in that case.

hpatro: Add a test case as well.

hpatro (commented):
Would be nice to think through the behavior with other flags/parameters of SET command https://valkey.io/commands/set/

`NX -- Only set the key if it does not already exist` - should we error in this case if CAS is provided?
`XX -- Only set the key if it already exists` - This becomes redundant to use I believe with CAS.
`GET -- Return the old string stored at key, or nil if key did not exist. An error is returned and SET aborted if the value stored at key is not a string.` - I think we need to support this parameter in some form. The scenario which comes to my mind is when CAS fails and a user doesn't need to send a GET command again to find out the value stored in the engine. However, we need to think about how to differentiate between success scenario vs failure scenario.

Also, we need to document them.

hpatro:
@sarthakaggarwal97 Please avoid force pushing. force push removes the reviewer's history in Github and one needs to look at the entire change again.

sarthakaggarwal97:
noted @hpatro, will avoid it.

tests/unit/type/string.tcl
```diff
@@ -582,6 +582,59 @@ if {[string match {*jemalloc*} [s mem_allocator]]} {
...
+        r del foo
+
+        r set foo "initial_value"
+
+        assert_equal {OK} [r set foo "new_value" ifeq "initial_value"]
+        assert_equal "new_value" [r get foo]
+
+        assert_equal {} [r set foo "should_not_set" ifeq "wrong_value"]
```
hpatro, line 593: @valkey-io/core-team On failure of compare/set instead of nil value we should return an error with old value in it. Otherwise, a client would need to perform another GET operation.

src/t_string.c
```diff
@@ -102,6 +104,31 @@ void setGenericCommand(client *c,
...
+    /* Handle the IFEQ conditional check */
+    if (flags & OBJ_SET_IFEQ && found) {
+        robj *current_value = lookupKeyRead(c->db, key);
+
+        if (current_value == NULL || current_value->type != OBJ_STRING || checkType(c, comparison, OBJ_STRING) != 0) {
+            if (!(flags & OBJ_SET_GET)) {
+                addReplyError(c, "value(s) must be present or string");
+            }
+            return;
+        }
```
hpatro: In this block `found` is true, so I believe we don't need to check `current_value` as NULL. We don't support key with `NULL` value in Valkey.
Also, `comparison` object is from the client input, so I think we can be assured it would be of string type. We should use the checkType method call for `current_value`. We can simplify it to:

```suggestion
        if (checkType(c, current_value, OBJ_STRING)) {
            return;
        }
```

hpatro (commented):
Thanks @sarthakaggarwal97 for the PR! Looks pretty good. Could you also document the behavior in the top comment. Will be easier for others to review and we can finalize it.

src/t_string.c
sarthakaggarwal97: Used the type to verify the current_value's data type

tests/unit/type/string.tcl
sarthakaggarwal97: added tests with xx, nx and get

src/t_string.c
sarthakaggarwal97: We return an error now!

src/t_string.c
```diff
@@ -100,7 +102,32 @@ void setGenericCommand(client *c,
         if (getGenericCommand(c) == C_ERR) return;
     }
 
-    found = (lookupKeyWrite(c->db, key) != NULL);
+    const robj *value = lookupKeyWrite(c->db, key);
```
hpatro: Would prefer 
```suggestion
    const robj *existing_value = lookupKeyWrite(c->db, key);
```

src/t_string.c
```diff
@@ -100,7 +102,32 @@ void setGenericCommand(client *c,
...
+    found = value != NULL;
+
+    /* Handle the IFEQ conditional check */
+    if (flags & OBJ_SET_IFEQ && found) {
+
```
hpatro: Remove.

src/t_string.c
```diff
@@ -100,7 +102,32 @@ void setGenericCommand(client *c,
...
+        if (value->type != OBJ_STRING) {
```
hpatro: Using the `checkType` method adds an error message to the client buffer and the error message will remain the same.

tests/unit/type/string.tcl
```diff
@@ -582,6 +582,59 @@ if {[string match {*jemalloc*} [s mem_allocator]]} {
...
+        assert_equal "new_value" [r get foo]
+    }
+
+    test "SET with IFEQ conditional - non-string current value" {
+        r del foo
+
+        r sadd foo "some_set_value"
+        assert_error {ERR value(s) must be present or string} {r set foo "new_value" ifeq "some_set_value"}
+    }
```
hpatro: Can we also add non-string value with GET parameter?

src/t_string.c
hpatro:
```
if (!(flags & OBJ_SET_GET) && checkType(c, o, OBJ_STRING)) {
    return;
}
```

src/t_string.c
sarthakaggarwal97: changed!

src/t_string.c
sarthakaggarwal97: good suggestion, thanks!

tests/unit/type/string.tcl
sarthakaggarwal97: added the test, thanks

tests/unit/type/string.tcl
enjoy-binbin: return an error with old value seems odd to me. i think in normal cases, a client is unlikely get an error in set IFEQ, they hold the old value in somewhere, and if the old value is unvalid, this mean client should abort the set, this usually means the client should abort the entire business logic. In this case, client should GET the new value as needed and usually they don't really need the new value, they juse want the result of whether the SET succeeded or not.

hpatro: @enjoy-binbin I understand if the value is updated from a DB to Valkey as a cache, the client won't benefit much from the value present. 

However, if I think Valkey being used as a datastore and value is updated by multiple clients based on the value stored (let say increment old value by 1). for this case, they would need to know the previous value.

Let's hear other devs opinion on this and resolve this.

hpatro (approved):
LGTM overall. The behavior currently is:

1. if value matches, set the new value.
2. If existing value is of different type, return error `WRONGTYPE Operation against a key holding the wrong kind of value`
3. if existing value is a mismatch, return nil. (want us to finalize on this).

src/commands/set.json
```diff
@@ -85,6 +85,20 @@
                 "name": "value",
                 "type": "string"
             },
+            {
+                "name": "comparison-value",
```
enjoy-binbin: i suppose we also need to add a since field, and update the history field.

src/t_string.c
```diff
@@ -100,7 +102,27 @@ void setGenericCommand(client *c,
         if (getGenericCommand(c) == C_ERR) return;
     }
 
-    found = (lookupKeyWrite(c->db, key) != NULL);
+    robj *existing_value = lookupKeyWrite(c->db, key);
+    found = existing_value != NULL;
+
+    /* Handle the IFEQ conditional check */
+    if (flags & OBJ_SET_IFEQ && found) {
+        if (!(flags & OBJ_SET_GET) && checkType(c, existing_value, OBJ_STRING)) {
```
enjoy-binbin, line 110: can this happen? i see in the above, we already do this getGenericCommand thing, this will handle all these error path i guess?
```
    if (flags & OBJ_SET_GET) {
        if (getGenericCommand(c) == C_ERR) return;
    }
```

supporting GET and IFEQ look odd to me. GET will get the old value, isn't the value passed in by IFEQ is the same old value?

src/t_string.c
```diff
@@ -304,34 +336,35 @@ int parseExtendedStringArgumentsOrReply(client *c, int *flags, int *unit, robj *
     return C_OK;
 }
 
-/* SET key value [NX] [XX] [KEEPTTL] [GET] [EX <seconds>] [PX <milliseconds>]
- *     [EXAT <seconds-timestamp>][PXAT <milliseconds-timestamp>] */
+/* SET key value [IFEQ <comparison-value>] [NX] [XX] [KEEPTTL] [GET] [EX <seconds>]
```
enjoy-binbin: let add this comment in the last? so we don't have to touch the first line.

tests/unit/type/string.tcl
madolson: I like the idea of returning value if the IFEQ fails, not as an error, just as a string. That way you can also check to see if someone else did the same work you did (and silently succeed) or retry you work again.

src/t_string.c
sarthakaggarwal97: thanks @enjoy-binbin for the review. 

>can this happen?

We currently just want to support Check and Set if the existing value is string. The check for `OBJ_SET_GET` flag prevents us to add additional line of response, when get is supposed to already add response before. 

>supporting GET and IFEQ look odd to me

I feel it's okay to support `GET` with `IFEQ`, in case the user gives an incorrect existing value, and `GET` can help retrieve it. It would also solve the discussion around if `IFEQ` should return the value incase the input doesn't matches the existing value (adding GET flag can always do that now).

@madolson @hpatro what do you guys think on this?

src/commands/set.json
sarthakaggarwal97: I missed pushing this earlier, thanks for pointing out! Updated

src/t_string.c
sarthakaggarwal97: updated!

zuiderkwast:
Let's make IFEQ mutually exclusive with NX and XX. It doesn't make sense to combine them so let's make it a syntax error now. (If we don't, we can't fix it in the future without a breaking change.)

    SET key value [ NX | XX | IFEQ comparison-value ] [ other args... ]

In the JSON file, putting them in the same "oneOf" block, this is also used for the website and man pages where the syntax is rendered.

Regarding the combining IFEQ with the GET parameter will be difficult to use, but not impossible with the behaviour you described. If SET replied with the 'comparison-value', it means the SET has succeeded. I think it's logical, even if it will probably be very rarely used.

src/commands/set.json
```diff
@@ -23,6 +23,10 @@
             [
                 "7.0.0",
                 "Allowed the `NX` and `GET` options to be used together."
+            ],
+            [
+                "8.2.0",
```
zuiderkwast: Our next version is `8.1.0`.

```suggestion
                "8.1.0",
```

src/commands/set.json
```diff
@@ -85,6 +89,21 @@
                 "name": "value",
                 "type": "string"
             },
+            {
+                "name": "comparison-value",
+                "type": "string",
+                "token": "IFEQ",
+                "since": "8.2.0",
```
zuiderkwast:
```suggestion
                "since": "8.1.0",
```

and move this block to within the "oneOf" for NX | XX below. The `"since": "2.6.12"` can be moved from the oneOf container to the NX and XX args individually.

src/t_string.c
```diff
@@ -305,33 +337,34 @@ int parseExtendedStringArgumentsOrReply(client *c, int *flags, int *unit, robj *
 }
 
 /* SET key value [NX] [XX] [KEEPTTL] [GET] [EX <seconds>] [PX <milliseconds>]
- *     [EXAT <seconds-timestamp>][PXAT <milliseconds-timestamp>] */
+ *     [EXAT <seconds-timestamp>][PXAT <milliseconds-timestamp>] [IFEQ <comparison-value>] */
```
zuiderkwast: SET key value [NX | XX | IFEQ comparison-value] [GET]
        [EX seconds | PX milliseconds | EXAT seconds-timestamp |
         PXAT milliseconds-timestamp | KEEPTTL]

Something like that is more in line with the actual syntax. On website and man pages we use lowercase for syntax variables. No `< >` brackets.

src/t_string.c
```diff
@@ -374,10 +407,11 @@ void getCommand(client *c) {
  */
 void getexCommand(client *c) {
     robj *expire = NULL;
+    robj *comparison = NULL;
     int unit = UNIT_SECONDS;
     int flags = OBJ_NO_FLAGS;
 
-    if (parseExtendedStringArgumentsOrReply(c, &flags, &unit, &expire, COMMAND_GET) != C_OK) {
+    if (parseExtendedStringArgumentsOrReply(c, &flags, &unit, &expire, &comparison, COMMAND_GET) != C_OK) {
```
zuiderkwast: IFEQ is not allowed with COMMAND_GET, so I think we can just pass NULL for the comparison argument. It will not be used. Makes sense?

```suggestion
    if (parseExtendedStringArgumentsOrReply(c, &flags, &unit, &expire, NULL, COMMAND_GET) != C_OK) {
```

tests/assets/test_cli_hint_suite.txt
```diff
@@ -68,18 +68,18 @@
...
-"SET key value EXAT" "[NX|XX] [GET] [EX seconds|PX milliseconds|EXAT unix-time-seconds|PXAT unix-time-milliseconds|KEEPTTL]"
-"SET key value EXAT " "unix-time-seconds [NX|XX] [GET]"
-"SET key value PX" "[NX|XX] [GET] [EX seconds|PX milliseconds|EXAT unix-time-seconds|PXAT unix-time-milliseconds|KEEPTTL]"
-"SET key value PX " "milliseconds [NX|XX] [GET]"
-"SET key value PXAT" "[NX|XX] [GET] [EX seconds|PX milliseconds|EXAT unix-time-seconds|PXAT unix-time-milliseconds|KEEPTTL]"
-"SET key value PXAT " "unix-time-milliseconds [NX|XX] [GET]"
-"SET key value KEEPTTL " "[NX|XX] [GET]"
-"SET key value XX " "[GET] [EX seconds|PX milliseconds|EXAT unix-time-seconds|PXAT unix-time-milliseconds|KEEPTTL]"
+"SET key value " "[IFEQ comparison-value] [NX|XX] [GET] [EX seconds|PX milliseconds|EXAT unix-time-seconds|PXAT unix-time-milliseconds|KEEPTTL]"
+"SET key value EX" "[IFEQ comparison-value] [NX|XX] [GET] [EX seconds|PX milliseconds|EXAT unix-time-seconds|PXAT unix-time-milliseconds|KEEPTTL]"
```
zuiderkwast: All of these...

```suggestion
"SET key value EX" "[NX|XX|IFEQ comparison-value]] [GET] [EX seconds|PX milliseconds|EXAT unix-time-seconds|PXAT unix-time-milliseconds|KEEPTTL]"
```

zuiderkwast (commented):
This looks pretty good. Just don't want to allow IFEQ combined with NX and XX.

src/commands/set.json
hpatro: I also like this approach. Doesn't make a lot of sense to combine IFEQ with NX | XX.

src/t_string.c
sarthakaggarwal97: @zuiderkwast so we are disallowing IFEQ with GET as well. Makes sense.

zuiderkwast: `COMMAND_GET` here means the GET command. It's not the GET argument for SET. If I'm not mistaken.

src/commands/set.json
```diff
@@ -89,17 +93,33 @@
...
-                        "token": "XX"
+                        "token": "XX",
+                        "since": "2.6.12"
+                    },
+                    {
+                        "name": "comparison-value",
+                        "type": "string",
+                        "token": "IFEQ",
+                        "since": "8.1.0",
+                        "optional": true,
```
zuiderkwast: It doesn't need optional, because the whole "oneof" block is optional.

We want the syntax `[NX | XX | IFEQ comparison-value]`, not `[NX | XX | [IFEQ comparison-value]]`.

tests/assets/test_cli_hint_suite.txt
```diff
@@ -68,17 +68,17 @@
...
-"SET key value EX " "seconds [NX|XX] [GET]"
-"SET key value EX 23 " "[NX|XX] [GET]"
-"SET key value EXAT" "[NX|XX] [GET] [EX seconds|PX milliseconds|EXAT unix-time-seconds|PXAT unix-time-milliseconds|KEEPTTL]"
-"SET key value EXAT " "unix-time-seconds [NX|XX] [GET]"
-"SET key value PX" "[NX|XX] [GET] [EX seconds|PX milliseconds|EXAT unix-time-seconds|PXAT unix-time-milliseconds|KEEPTTL]"
-"SET key value PX " "milliseconds [NX|XX] [GET]"
-"SET key value PXAT" "[NX|XX] [GET] [EX seconds|PX milliseconds|EXAT unix-time-seconds|PXAT unix-time-milliseconds|KEEPTTL]"
-"SET key value PXAT " "unix-time-milliseconds [NX|XX] [GET]"
-"SET key value KEEPTTL " "[NX|XX] [GET]"
+"SET key value " "[NX|XX|[IFEQ comparison-value]] [GET] [EX seconds|PX milliseconds|EXAT unix-time-seconds|PXAT unix-time-milliseconds|KEEPTTL]"
```
zuiderkwast: I guess if you remove "optional" for IFEQ in the JSON file, this will become like this:

```suggestion
"SET key value " "[NX|XX|IFEQ comparison-value] [GET] [EX seconds|PX milliseconds|EXAT unix-time-seconds|PXAT unix-time-milliseconds|KEEPTTL]"
```

src/t_string.c
```diff
@@ -305,33 +337,34 @@ int parseExtendedStringArgumentsOrReply(client *c, int *flags, int *unit, robj *
 }
 
 /* SET key value [NX] [XX] [KEEPTTL] [GET] [EX <seconds>] [PX <milliseconds>]
- *     [EXAT <seconds-timestamp>][PXAT <milliseconds-timestamp>] */
+ *     [EXAT <seconds-timestamp>][PXAT <milliseconds-timestamp>] [IFEQ comparison-value] */
```
zuiderkwast: We should correct this comment to the correct syntax:

    SET key value [NX | XX | IFEQ comparison-value] [GET]
        [EX seconds | PX milliseconds |
         EXAT unix-time-seconds | PXAT unix-time-milliseconds |
         KEEPTTL]

tests/unit/type/string.tcl
```diff
@@ -582,6 +582,55 @@ if {[string match {*jemalloc*} [s mem_allocator]]} {
...
+        assert_error {WRONGTYPE Operation against a key holding the wrong kind of value} {r set foo "new_value" ifeq "some_set_value"}
+    }
+
+
+    test "SET with IFEQ conditional - with get" {
+        r del foo
+
+        assert_equal {} [r set foo "new_value" ifeq "initial_value" get]
+
+        r set foo "initial_value"
```
zuiderkwast: This checks that the value before SET was empty. We should check also the value after SET, to check that SET didn't set the key to the value.

```suggestion
        assert_equal {} [r set foo "new_value" ifeq "initial_value" get]
        assert_equal {} [r get foo]

        r set foo "initial_value"
```

sarthakaggarwal97: done!

src/commands/set.json
sarthakaggarwal97: noted! misinterpreted it before. Have made the change.

src/commands/set.json
sarthakaggarwal97: addressed!

src/t_string.c
```diff
@@ -304,34 +336,35 @@ int parseExtendedStringArgumentsOrReply(client *c, int *flags, int *unit, robj *
     return C_OK;
 }
 
-/* SET key value [NX] [XX] [KEEPTTL] [GET] [EX <seconds>] [PX <milliseconds>]
- *     [EXAT <seconds-timestamp>][PXAT <milliseconds-timestamp>] */
+/* SET key value [NX | XX | IFEQ comparison-value] [KEEPTTL] [GET] [EX <seconds>]
+ *     [PX <milliseconds>] [EXAT <seconds-timestamp>][PXAT <milliseconds-timestamp>] */
```
zuiderkwast: I see you corrected `NX | XX | IFEQ` here, but let's correct the other arguments too, to indicate which ones are mutually exclusive. (For example, we don't allow combining EX and PX with each other or with KEEPTTL.) It's not introduced in this PR, but still, I don't like comments that are wrong.

```suggestion
/* SET key value [NX | XX | IFEQ comparison-value] [GET]
 *     [EX seconds | PX milliseconds |
 *      EXAT seconds-timestamp | PXAT milliseconds-timestamp | KEEPTTL] */
```

src/t_string.c
```diff
@@ -295,7 +317,17 @@ int parseExtendedStringArgumentsOrReply(client *c, int *flags, int *unit, robj *
             *unit = UNIT_MILLISECONDS;
             *expire = next;
             j++;
-        } else {
+        } else if ((opt[0] == 'i' || opt[0] == 'I') &&
+            (opt[1] == 'f' || opt[1] == 'F') &&
+            (opt[2] == 'e' || opt[2] == 'E') &&
+            (opt[3] == 'q' || opt[3] == 'Q') && opt[4] == '\0' &&
+            next && !(*flags & OBJ_SET_NX || *flags & OBJ_SET_XX) && (command_type == COMMAND_SET))
```
zuiderkwast: Not important, but it seems more logical to put this `else if` after the NX and XX checks, because they're related and they are in this order in JSON file and in other places.

zuiderkwast (approved):
Great, LGTM now.

Just a few nits. Not a blocker for merging.

tests/unit/type/string.tcl
zuiderkwast: With the current implementation of `SET key value IFEQ compare GET` in this PR, the old value is returned regardless of whether it's matching the `compare` string or not. This can be used for the use case @madolson described.

Without GET, I don't think we can return the old value on mismatch, because of this corner case: When the old value is the string "OK", if the client library returns "OK" as a string (i.e. if it doesn't distinguish between bulk string and simple string), the user can't tell if "OK" means that the SET succeeded or if just returned the old value which was the string "OK".

zuiderkwast:
@madolson @hpatro @enjoy-binbin Let's finalize the exact behavior with the various flags. This is the currently implemented logic:

* IFEQ is mutually exclusive with NX and XX. They can't be combined.
* IFEQ without GET: On success `+OK` and on mismatch `(nil)` is returned. This matches the behavior of XX and NX.
* IFEQ with GET: In this case, the old value is returned regardless of whether the comparison succeeded.

Please confirm ( :+1: ) or protest ( :scream: ).

hpatro:
> * IFEQ without GET: On success +OK and on mismatch (nil) is returned. This matches the behavior of XX and NX.
> * IFEQ with GET: In this case, the old value is returned regardless of whether the comparison succeeded.

What I like about this is, it's opt in. If the user actually wants the old value only then they get it. And, if the stored value is large, they could choose to not get it back.

zuiderkwast:
@sarthakaggarwal97 can you prepare a doc PR? The valkey-doc repo, commands/set.md

enjoy-binbin (approved):

sarthakaggarwal97:
@zuiderkwast I had raised https://github.com/valkey-io/valkey-doc/pull/189. Please review this and let me know if I missed something here.

sarthakaggarwal97:
Hi @valkey-io/core-team, are we good to merge this?

madolson:
> Please confirm ( 👍 ) or protest ( 😱 ).

You can't use 😱 as a reaction :)

sarthakaggarwal97:
> You can't use 😱 as a reaction :)

@zuiderkwast didn't want anyone to protest 😂

src/t_string.c
```diff
@@ -228,14 +250,23 @@ int parseExtendedStringArgumentsOrReply(client *c, int *flags, int *unit, robj *
...
                    (opt[1] == 'x' || opt[1] == 'X') && opt[2] == '\0' &&
-                   !(*flags & OBJ_SET_NX) && (command_type == COMMAND_SET))
+                   !(*flags & OBJ_SET_NX || *flags & OBJ_SET_IFEQ) && (command_type == COMMAND_SET))
         {
             *flags |= OBJ_SET_XX;
+        } else if ((opt[0] == 'i' || opt[0] == 'I') &&
+            (opt[1] == 'f' || opt[1] == 'F') &&
+            (opt[2] == 'e' || opt[2] == 'E') &&
+            (opt[3] == 'q' || opt[3] == 'Q') && opt[4] == '\0' &&
+            next && !(*flags & OBJ_SET_NX || *flags & OBJ_SET_XX) && (command_type == COMMAND_SET))
```
madolson: Let's also make sure the OBJ_SET_IFEQ is not set here, so that `SET FOO BAR IFEQ FOO IFEQ BAR` is not accepted.

madolson:
Just a minor comment, otherwise it looks good to me. We have a weekly meeting later today, I'll ping folks just to close the loop.

src/t_string.c
sarthakaggarwal97: thank you, added this check as well

madolson (approved):

madolson:
We discussed this a bit in the weekly meeting, a callout was that we want to make sure that clients add support for this.

The most significant callout was that there is some uncertainty now if this is the right API. The ask was that we should spend more time to figure out if this is the best way to solve the original use cases, so if we could create some examples in the documentation page about what problem this solves and why it's better than LUA or multi, we will follow up. @hpatro  and @sarthakaggarwal97 let me know if you have any questions.

zuiderkwast:
Client changes?

That's only if the client does some detailed wrapping of each and every command argument, right? Is that really considered good practice? I would like to call it an anti-pattern.

No changes are needed for any of the clients I've been involved in: libvalkey (hiredis-fork), valkey-cli and ered (an Erlang client). A command line is just represented as a list of strings and called like `client.call("SET", "foo", "bar", "IFEQ", "baz")`. Why do some clients do it differently? It just seems very unpractical.

soloestoy:
>* IFEQ with GET: In this case, the old value is returned regardless of whether the comparison succeeded.

When using `GET` parameters, the result of `SET IFEQ` is ambiguous, and we cannot intuitively determine whether the command successfully modified the data. There is a similar issue with `NX` and `XX`. I have submitted an issue #1415 to discuss this problem.

zuiderkwast:
> why it's better than LUA or multi

For simple use cases, it's simpler than Lua and WATCH + MULTI. It's also faster. (With WATCH + MULTI you need two round trips to the server.)

madolson:
> That's only if the client does some detailed wrapping of each and every command argument, right? Is that really considered good practice? I would like to call it an anti-pattern.

I mean, I'll pin it for a future discussion, but all of the GLIDE clients have fully typed inputs and outputs so that it's easier for developers to see what the output will look like instead of having to infer it.

PingXie:
> but all of the GLIDE clients have fully typed inputs

I think @soloestoy's question in #1415 is about semantics, i.e., `nil` could mean both failure and success. The type of the return value for `set` however is still string with this change.

PR #1326, opened 2024-11-20
Make KEYS can visit expired key in import-source state

enjoy-binbin:
After #1185, a client in import-source state can visit expired
key both in read commands and write commands, this commit handle
keyIsExpired function to handle import-source state as well, so
KEYS can visit the expired key.

This is not particularly important, but it ensures the definition,
also doing some cleanup around the test, verified that the client
can indeed visit the expired key.

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -388,7 +388,7 @@ robj *dbRandomKey(serverDb *db) {
             if (allvolatile && (server.primary_host || server.import_mode) && --maxtries == 0) {
                 /* If the DB is composed only of keys with an expire set,
                  * it could happen that all the keys are already logically
-                 * expired in the repilca, so the function cannot stop because
+                 * expired in the replica, so the function cannot stop because
                  * expireIfNeeded() is false, nor it can stop because
                  * dictGetFairRandomKey() returns NULL (there are keys to return).
                  * To prevent the infinite loop we do some tries, but if there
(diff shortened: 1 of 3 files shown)


zuiderkwast:
I guess an importing client will not use KEYS and SCAN, but I agree, it's good to ensure the contract that this client can read expired keys.

src/db.c
```diff
@@ -1798,7 +1798,13 @@ int keyIsExpiredWithDictIndex(serverDb *db, robj *key, int dict_index) {
 /* Check if the key is expired. */
 int keyIsExpired(serverDb *db, robj *key) {
     int dict_index = getKVStoreIndexForKey(key->ptr);
-    return keyIsExpiredWithDictIndex(db, key, dict_index);
+    if (!keyIsExpiredWithDictIndex(db, key, dict_index)) return 0;
+
+    /* See expireIfNeededWithDictIndex for more details. */
+    if (server.primary_host == NULL && server.import_mode) {
```
lyq2333, line 1804: Shall we use `iAmPrimary()` here? I am not very clear about the difference between these two.

lyq2333 (commented):
Agree with @zuiderkwast.  LGTM.

lyq2333 (approved):

src/db.c
enjoy-binbin: yes, i think we should, we probably need to use iAmMaster to replace all the server. primary_host == NULL. The difference is that i guess, when restarting a cluster replica node, its primary_host may be NULL but it is a replica node in cluster role.

enjoy-binbin: i guess it is a small change (but there are too many references that need to look with), i am going to open a new PR for a better review.

enjoy-binbin: i remember @soloestoy had dealt with it before, do you have the old fix (backups) locally that you want to pick up?

zuiderkwast (approved):

PR #1327, opened 2024-11-20
Skip reclaim file page cache test in valgrind

enjoy-binbin:
The test is incompatible with valgrind. Added a new `--valgrind`
argument to test suite, which will cause that test to be skipped.

We skipped it in the past, see https://github.com/valkey-io/valkey/commit/5b61b0dc6d2579ee484fa6cf29bfac59513f84ab.

### DIFF
--- a/src/unit/test_util.c
+++ b/src/unit/test_util.c
@@ -286,7 +286,9 @@ static int cache_exist(int fd) {
 int test_reclaimFilePageCache(int argc, char **argv, int flags) {
     UNUSED(argc);
     UNUSED(argv);
-    UNUSED(flags);
+
+    /* The test is incompatible with valgrind, skip it. */
+    if (flags & UNIT_TEST_VALGRIND) return 0;
 
 #if defined(__linux__)
     char *tmpfile = "/tmp/redis-reclaim-cache-test";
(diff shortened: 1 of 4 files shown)


enjoy-binbin:
daily: https://github.com/enjoy-binbin/valkey/actions/runs/11928806942

src/unit/test_util.c
```diff
@@ -286,7 +286,9 @@ static int cache_exist(int fd) {
 int test_reclaimFilePageCache(int argc, char **argv, int flags) {
     UNUSED(argc);
     UNUSED(argv);
-    UNUSED(flags);
+
+    /* The test is incompatible with valgrind, skip it. */
```
zuiderkwast, line 290: I'm just curious why it's incompatible. If you know, feel free to mention it in this code comment.

zuiderkwast (approved):

src/unit/test_util.c
enjoy-binbin: i dont know the real reason, i am guessing valgrind may do some trick magic like mock the mmap to track the memory (and it did not handle this check), i am not an expert on it, just a guessing.

zuiderkwast:
FYI. This test case also fails for me locally on linux on macbook M2, without valgrind.

enjoy-binbin:
> FYI. This test case also fails for me locally on linux on macbook M2, without valgrind.

ok... so a fake linux will also fail, i don't have this env so i can not try.

zuiderkwast:
It's not "fake linux" :rofl:

It is normal Linux compiled for Aarch64 with 16KB page size and 128B cache line size and

madolson (approved):
Cool

PR #1332, opened 2024-11-21
Use goto to cleanup error handling in readSyncBulkPayload

enjoy-binbin:
The goto error label is the same as the error return, use goto
to reduce the references.
```
error:
    cancelReplicationHandshake(1);
    return;
```

Also this can make the log printing more continuous under the
error, that is, we print the error log first, and then print
the reconnecting log at the last (in cancelReplicationHandshake).

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -2091,8 +2091,7 @@ void readSyncBulkPayload(connection *conn) {
             }
             serverLog(LL_WARNING, "I/O error trying to sync with PRIMARY: %s",
                       (nread == -1) ? connGetLastError(conn) : "connection lost");
-            cancelReplicationHandshake(1);
-            return;
+            goto error;
         }
         server.stat_net_repl_input_bytes += nread;
 
@@ -2257,7 +2256,6 @@ void readSyncBulkPayload(connection *conn) {
 
         if (loadingFailed) {
             stopLoading(0);
-            cancelReplicationHandshake(1);
             rioFreeConn(&rdb, NULL);
 
             if (server.repl_diskless_load == REPL_DISKLESS_LOAD_SWAPDB) {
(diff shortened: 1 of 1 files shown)


enjoy-binbin:
a test log example:
```
67697:S 21 Nov 2024 16:01:03.309 * Loading RDB produced by Valkey version 255.255.255
67697:S 21 Nov 2024 16:01:03.309 * RDB age 0 seconds
67697:S 21 Nov 2024 16:01:03.309 * RDB memory usage when created 1.23 Mb
67697:S 21 Nov 2024 16:01:03.309 # binbintest load error
67697:S 21 Nov 2024 16:01:03.309 # Failed trying to load the PRIMARY synchronization DB from disk, check server logs.
67697:S 21 Nov 2024 16:01:03.310 * Reconnecting to PRIMARY 127.0.0.1:6379 after failure
67697:S 21 Nov 2024 16:01:03.310 * PRIMARY <-> REPLICA sync started
67697:S 21 Nov 2024 16:01:03.310 * PRIMARY <-> REPLICA sync: Discarding the half-loaded data
```

the reconnecting and sync started should be after discarding one.

zuiderkwast (approved):
Makes sense.

The "Reconnecting" is logged by cancelReplicationHandshake.

cancelReplicationHandshake is a no-op in some cases where we just did return before, right?

enjoy-binbin:
> cancelReplicationHandshake is a no-op in some cases where we just did return before, right?

no, cancelReplicationHandshake is not a no-op, and we are not just did return before, we did cancelReplicationHandshake and return together before, this commit just like, adjust the code location (or order).

zuiderkwast:
> > cancelReplicationHandshake is a no-op in some cases where we just did return before, right?
> 
> no, cancelReplicationHandshake is not a no-op, and we are not just did return before, we did cancelReplicationHandshake and return together before, this commit just like, adjust the code location (or order).

Ah, I see now. Every `goto error` replaces cancelReplicationHandshake + return.

PR #1333, opened 2024-11-21
Adjust memory prefetch to the new hash table

uriyage:
Adjust memory prefetch for new hash table implementation in IO-threading.

This change adapts the existing prefetch mechanism to work with the new 
hash table structure in IO-threading scenarios.

Changes:
- Use new `hashtableIncrementalFind` API for improved prefetching
- Object free to be done in main-thread while keeping obj->ptr offloading 
  in IO-thread since the DB object is now allocated by the main-thread 
  and not by the io-thread as it used to be.

Performance impact vs new hashtable **_without_** prefetch:
- GET: 950K TPS → 1100K TPS (15.8% improvement)
- SET: 904K TPS → 1040K TPS (15% improvement)

Performance impact vs unstable (with the old dictionary):
- GET: 1080K TPS → 1100K TPS (1.85% improvement)
- SET: 1190K TPS → 1040K TPS (12.6% regression)

Note:  SET performance regression vs unstable needs to be investigated and addressed.

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -912,6 +912,7 @@ void keysCommand(client *c) {
 typedef struct {
     list *keys;     /* elements that collect from dict */
     robj *o;        /* o must be a hash/set/zset object, NULL means current db */
+    serverDb *db;   /* database currently being scanned */
     long long type; /* the particular type when scan the db */
     sds pattern;    /* pattern string, NULL means no pattern */
     long sampled;   /* cumulative number of keys sampled */
@@ -950,6 +951,15 @@ void keysScanCallback(void *privdata, void *entry) {
     /* Filter object if its key does not match the pattern. */
     if (data->pattern) {
         if (!stringmatchlen(data->pattern, sdslen(data->pattern), key, sdslen(key), 0)) {
+            return;
+        }
+    }
+
+    /* Handle and skip expired key. */
+    if (objectIsExpired(obj)) {
+        robj kobj;
+        initStaticStringObject(kobj, key);
+        if (expireIfNeeded(data->db, &kobj, obj, 0) != KEY_VALID) {
             return;
         }
     }
@@ -1184,6 +1194,7 @@ void scanGenericCommand(client *c, robj *o, unsigned long long cursor) {
          * only keys are returned. */
         scanData data = {
             .keys = keys,
+            .db = c->db,
             .o = o,
             .type = type,
             .pattern = use_pattern ? pat : NULL,
@@ -1252,25 +1263,7 @@ void scanGenericCommand(client *c, robj *o, unsigned long long cursor) {
         serverPanic("Not handled encoding in SCAN.");
     }
 
-    /* Step 3: Filter the expired keys */
-    /* TODO: Do this in the keysScanCallback where we have the valkey objects
-     * that contain the TTL (or add valkey object to the list instead of just
-     * the keys). Then we don't need to look them up again here. */
-    if (o == NULL && listLength(keys)) {
-        robj kobj;
-        listIter li;
-        listNode *ln;
-        listRewind(keys, &li);
-        while ((ln = listNext(&li))) {
-            sds key = listNodeValue(ln);
-            initStaticStringObject(kobj, key);
-            if (expireIfNeeded(c->db, &kobj, NULL, 0) != KEY_VALID) {
-                listDelNode(keys, ln);
-            }
-        }
-    }
-
-    /* Step 4: Reply to the client. */
+    /* Step 3: Reply to the client. */
     addReplyArrayLen(c, 2);
     addReplyBulkLongLong(c, cursor);
 
--- a/src/db.c
+++ b/src/db.c
@@ -52,13 +52,13 @@ typedef enum {
     KEY_DELETED    /* The key was deleted now. */
 } keyStatus;
 
-keyStatus expireIfNeededWithDictIndex(serverDb *db, robj *key, int flags, int dict_index);
-keyStatus expireIfNeeded(serverDb *db, robj *key, int flags);
-int keyIsExpiredWithDictIndex(serverDb *db, robj *key, int dict_index);
-int keyIsExpired(serverDb *db, robj *key);
+static keyStatus expireIfNeededWithDictIndex(serverDb *db, robj *key, robj *val, int flags, int dict_index);
+static keyStatus expireIfNeeded(serverDb *db, robj *key, robj *val, int flags);
+static int keyIsExpiredWithDictIndex(serverDb *db, robj *key, int dict_index);
+static int objectIsExpired(robj *val);
 static valkey *dbSetValue(serverDb *db, robj *key, robj *val, int overwrite, void **oldref);
 static int getKVStoreIndexForKey(sds key);
-valkey *dbFindExpiresWithDictIndex(serverDb *db, sds key, int dict_index);
+static valkey *dbFindExpiresWithDictIndex(serverDb *db, sds key, int dict_index);
 
 /* Update LFU when an object is accessed.
  * Firstly, decrement the counter if the decrement time is reached.
@@ -111,9 +111,7 @@ robj *lookupKey(serverDb *db, robj *key, int flags) {
         int expire_flags = 0;
         if (flags & LOOKUP_WRITE && !is_ro_replica) expire_flags |= EXPIRE_FORCE_DELETE_EXPIRED;
         if (flags & LOOKUP_NOEXPIRE) expire_flags |= EXPIRE_AVOID_DELETE_EXPIRED;
-        /* TODO: Use a better design for expireIfNeeded, now that the expire is
-         * stored in the value, for example passing val to expireIfNeeded. */
-        if (objectGetExpire(val) != -1 && expireIfNeeded(db, key, expire_flags) != KEY_VALID) {
+        if (expireIfNeeded(db, key, val, expire_flags) != KEY_VALID) {
             /* The key is no longer valid. */
             val = NULL;
         }
@@ -435,7 +433,7 @@ robj *dbRandomKey(serverDb *db) {
 
         key = objectGetKey(valkey);
         keyobj = createStringObject(key, sdslen(key));
-        if (dbFindExpiresWithDictIndex(db, key, randomDictIndex)) {
+        if (objectIsExpired(valkey)) {
             if (allvolatile && (server.primary_host || server.import_mode) && --maxtries == 0) {
                 /* If the DB is composed only of keys with an expire set,
                  * it could happen that all the keys are already logically
@@ -447,7 +445,7 @@ robj *dbRandomKey(serverDb *db) {
                  * return a key name that may be already expired. */
                 return keyobj;
             }
-            if (expireIfNeededWithDictIndex(db, keyobj, 0, randomDictIndex) != KEY_VALID) {
+            if (expireIfNeededWithDictIndex(db, keyobj, valkey, 0, randomDictIndex) != KEY_VALID) {
                 decrRefCount(keyobj);
                 continue; /* search for another key. This expired. */
             }
@@ -815,7 +813,7 @@ void delGenericCommand(client *c, int lazy) {
     int numdel = 0, j;
 
     for (j = 1; j < c->argc; j++) {
-        if (expireIfNeeded(c->db, c->argv[j], 0) == KEY_DELETED) continue;
+        if (expireIfNeeded(c->db, c->argv[j], NULL, 0) == KEY_DELETED) continue;
         int deleted = lazy ? dbAsyncDelete(c->db, c->argv[j]) : dbSyncDelete(c->db, c->argv[j]);
         if (deleted) {
             signalModifiedKey(c, c->db, c->argv[j]);
@@ -1266,7 +1264,7 @@ void scanGenericCommand(client *c, robj *o, unsigned long long cursor) {
         while ((ln = listNext(&li))) {
             sds key = listNodeValue(ln);
             initStaticStringObject(kobj, key);
-            if (expireIfNeeded(c->db, &kobj, 0) != KEY_VALID) {
+            if (expireIfNeeded(c->db, &kobj, NULL, 0) != KEY_VALID) {
                 listDelNode(keys, ln);
             }
         }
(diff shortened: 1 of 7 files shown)


src/io_threads.c
```diff
@@ -501,7 +503,12 @@ int tryOffloadFreeObjToIOThreads(robj *obj) {
         return C_ERR;
     }
 
-    IOJobQueue_push(jq, decrRefCountVoid, obj);
+    /* We offload only the free of the ptr that may be allocated by the I/O thread.
+     * The object itself was allocated by the main thread and will be freed by the main thread. */
+    IOJobQueue_push(jq, zfree, obj->ptr);
```
zuiderkwast: In OBJ_ENCODING_RAW, the pointer is an sds string, right?

```suggestion
    IOJobQueue_push(jq, sdsfree, obj->ptr);
```

ranshid:
@uriyage can you explain in the top comment that this is meant to improve the io-threading case?
Also please state the numbers here compared to unstable and compared to hashset before your change

src/memory_prefetch.c
```diff
@@ -8,15 +8,282 @@
...
+    size_t key_count;               /* Number of keys in the current batch */
+    size_t client_count;            /* Number of clients in the current batch */
+    size_t max_prefetch_size;       /* Maximum number of keys to prefetch in a batch */
+    size_t executed_commands;       /* Number of commands executed in the current batch */
+    int *slots;                     /* Array of slots for each key */
+    void **keys;                    /* Array of keys to prefetch in the current batch */
+    client **clients;               /* Array of clients in the current batch */
+    hashtable **keys_tables;        /* Main table for each key */
+    hashtable **expire_tables;      /* Expire table for each key */
+    KeyPrefetchInfo *prefetch_info; /* Prefetch info for each key */
```
zuiderkwast: All of these that are for each key, why not put them in a struct and put an array of such structs here? Then we just need one allocation.

```C
struct {
    int slot;
    void *key;
    client *client;
    ...
} batch_items[];
```

Just an idea, for the future. Right now I think we want minimal changes compared to unstable, because this hashtable feature is already huge.

src/memory_prefetch.c
```diff
@@ -8,15 +8,282 @@
...
+    if (max_prefetch_size == 0) {
+        return;
+    }
+
+    batch = zcalloc(sizeof(PrefetchCommandsBatch));
+    batch->max_prefetch_size = max_prefetch_size;
+    batch->clients = zcalloc(max_prefetch_size * sizeof(client *));
+    batch->keys = zcalloc(max_prefetch_size * sizeof(void *));
+    batch->keys_tables = zcalloc(max_prefetch_size * sizeof(hashtable *));
+    batch->expire_tables = zcalloc(max_prefetch_size * sizeof(hashtable *));
```
zuiderkwast: Can all of these be done in a single zcalloc? I guess it can if we change as suggested in the comment above.

Regarding expire table, it doesn't need to be accessed for non-expired keys, so maybe we don't need to prefetch it? The expire is stored in the robj which can be accessed without looking up the expire table.

zuiderkwast (commented):
Thanks! Mainly one comment about zfree vs sdsfree.

Did you try to minimize the diff of memory_prefetch.c compared to unstable?

src/io_threads.c
uriyage: Yes this is an error, I fixed it but forgot to commit, fixed. Thanks!

src/memory_prefetch.c
uriyage: What are the scenarios/conditions that trigger access to the expire table?

src/memory_prefetch.c
uriyage: Good idea. I remember there was a reason why we did it this way, but I can't recall what it was. I'll check it out.

src/memory_prefetch.c
zuiderkwast: The expire table needs to be accessed only to insert, delete or replace an object.

For example in GET, if a key is expired, the command needs to access the expire table to delete it. So maybe it's good to do it, but OTOH, if this unlikely to happen, then we evict other stuff from the L1 cache when we add this.

I'm not sure, so let's leave it for now?

uriyage:
> @uriyage can you explain in the top comment that this is meant to improve the io-threading case? Also please state the numbers here compared to unstable and compared to hashset before your change

Done, Thanks.

src/memory_prefetch.c
zuiderkwast: Another possibility is that we can do it when we have the robj AKA valkey object. Here we can actually check the expire value in this prefetch code ... and prefetch the expire table only if it's expired, or at least only if it has an expire at all.

src/io_threads.c
```diff
@@ -501,7 +503,12 @@ int tryOffloadFreeObjToIOThreads(robj *obj) {
         return C_ERR;
     }
 
-    IOJobQueue_push(jq, decrRefCountVoid, obj);
+    /* We offload only the free of the ptr that may be allocated by the I/O thread.
+     * The object itself was allocated by the main thread and will be freed by the main thread. */
+    IOJobQueue_push(jq, sdsfree, obj->ptr);
```
zuiderkwast: Hm... The compiler doesn't like sdsfree here because it expects `void (*)(void *)`.

I guess we have two options:
1. Create a wrapper that takes a void *.
2. Cast it to `void (*)(void *)`, via `(void *)` to avoid the warnings.

src/memory_prefetch.c
zuiderkwast: For each client, there is one argv, but this argv can contain multiple keys. I suppose that's why we have `key_count != clientcount`.

src/memory_prefetch.c
```diff
@@ -8,15 +8,282 @@
...
+    size_t executed_commands;       /* Number of commands executed in the current batch */
+    int *slots;                     /* Array of slots for each key */
+    void **keys;                    /* Array of keys to prefetch in the current batch */
+    client **clients;               /* Array of clients in the current batch */
+    hashtable **keys_tables;        /* Main table for each key */
+    hashtable **expire_tables;      /* Expire table for each key */
+    KeyPrefetchInfo *prefetch_info; /* Prefetch info for each key */
+} PrefetchCommandsBatch;
+
+static PrefetchCommandsBatch *batch = NULL;
```
ranshid, line 39: high level comment:
I think this is fine that we have a single global batch for reading jobs from io-threads. I do think that we could have all the related functions accept a batch as input so it might be used for other future cases were we would maybe like to start a batch on the engine side for some other purpose.
So we could have a  
static PrefetchCommandsBatch *g_batch = NULL;
and pass it to the relevant functions.

src/io_threads.c
```diff
@@ -493,6 +493,8 @@ int tryOffloadFreeObjToIOThreads(robj *obj) {
 
     if (obj->refcount > 1) return C_ERR;
 
+    if (obj->type != OBJ_ENCODING_RAW || obj->type != OBJ_STRING) return C_ERR;
```
zuiderkwast: In unstable, we can offload freeing of a nested object, like hashes or zsets.

This function is called only from dbSetValue, when a value is replaced, but there we actually want to keep the robj and only replace the robj->ptr if encoding is RAW. So there are not many scenarios where this will be used, I think.

It would be good if we can do that, without freeing the robj itself. We need the ptr and the encoding I think.

I have this draft code to add to dbSetValue:

```C
    /* Can we keep the object in the database and just replace the ptr? */
    if (old->refc == 1 && old->encoding == OBJ_ENCODING_RAW && val->encoding != OBJ_ENCODING_EMBSTR) {
        sdsfree(old->ptr); /* TODO: If it's very large, offload to io thread of lazy free. */
        old->ptr = val->ptr;
        old->type = val->type;
        old->encoding = val->encoding;
        decrRefCount(val);
        return old;
    }
```

I would like to keep the `old` robj for any encoding except EMBSTR. That is, if the key was for example a zset but it is replaced with a string value, we want to lazyfree the zset ptr but keep the robj.

src/io_threads.c
zuiderkwast: :bulb: Better idea:

```suggestion
    IOJobQueue_push(jq, zfree, sdsAllocPtr(obj->ptr));
```

src/memory_prefetch.c
uriyage: Thanks, I will run some tests with expired keys to check what give the best result.

src/io_threads.c
uriyage: This will require memory access to the ptr which we try to avoid.

uriyage: Maybe we can add sdsfreeVoid version as we did with decrRefCound for obj.
```
/* This variant of decrRefCount() gets its argument as void, and is useful
 * as free method in data structures that expect a 'void free_object(void*)'
 * prototype for the free method. */
void decrRefCountVoid(void *o) {
    decrRefCount(o);
}
```

zuiderkwast: > This will require memory access to the ptr

Good point.

> sdsfreeVoid

Yes, this is option 1 in my first comment. You don't like `IOJobQueue_push(jq, (void (*)(void *))(void *)sdsfree, obj->ptr);`?  It's too ugly, right?

OK, probably sdsfreeVoid is better.

src/memory_prefetch.c
ranshid: I think another option would be to interleave the expire table scan with the hashtable scan in case the command is a mutational command (as @zuiderkwast suggested). although we do not know the key has expire or not, waiting for the key and only then start the expiry scan seems like something we can optimize via parallelization.
In most cases I have seen either the whole dataset has TTL or only a minor fraction. maybe we can use that assumption and optimize since most workloads do have expiry on objects (for volatile eviction). Maybe we can test if tit improves later on @uriyage

src/io_threads.c
ranshid: can't we just typedef the (void (*)(void *))(void *) and cast to the type?

zuiderkwast: Yes, that would the same, but maybe more readable. The `(void *)` in end is an extra cast just to avoid warnings for casting between incompatible type.

src/memory_prefetch.c
zuiderkwast: Agree this can be improved, but let's consider the size of the hashstable feature diff. It's already quite huge. This code is the same as in unstable. I checked the diff between this file in unstable locally and it really keeps what can be kept from unstable.

src/io_threads.c
uriyage: Added `sdsfreeVoid` and removed `decrRefCountVoid` which is no longer in use.

src/io_threads.c
zuiderkwast: See #1337

zuiderkwast: I've added the optimization from #1337 to this PR.

src/memory_prefetch.c
uriyage: Current code performance tests results:

| Scenario  | With Expire Prefetch (K ops/sec) | Without Expire Prefetch (K ops/sec) | Change | Unstable(979f4c1ce) (K ops/sec) |
|----------|---------------------------|-----------------------------| ---------|-------------------|
| All Keys Expired  SET | 747 | 697 | +7% | 639 |
| All Keys Expired  GET | 985 | 930 | +5.9% | 1008 |
| 50% Expired GET | 983 | 985 | ~0% | 1018 |
| 1% Expired GET | 1049 | 1102 | -4.8% | 1009 |

The results show that prefetch from expired table has different impacts depending on the percentage of expired keys. While it improves performance when all keys have expiry (+7% for SET, +5.9% for GET), it actually hurts performance by 4.8% when only 1% of keys are expired.

> For example in GET, if a key is expired, the command needs to access the expire table to delete it. So maybe it's good to do it

The optimization to access the expire table only when an object has actually expired is still in `TODO`

After modifying the code to check the expiration time before accessing the expire table
```c
-         /* TODO: Use a better design for expireIfNeeded, now that the expire is
-          * stored in the value, for example passing val to expireIfNeeded. */
-        if (objectGetExpire(val) != -1 && expireIfNeeded(db, key, expire_flags) != KEY_VALID) {
-
+        int expired = 0;
+        mstime_t when = objectGetExpire(val)
+
+        if (when > 0) {
+            mstime_t now = commandTimeSnapshot();
+            
+            if (now > when) {
+                expired = 1;
+            }
+        }
+
+        if (expired && expireIfNeeded(db, key, expire_flags) != KEY_VALID) {
             /* The key is no longer valid. */
             val = NULL;
         }
```
here are the new results:

| Operation | With Expire Prefetch (K ops/sec) | Without Expire Prefetch (K ops/sec) | Change |
|-----------|---------------------------|-----------------------------| -------|
| SET | 765 | 800 | -4.4% |
| GET | 1020 | 1115 | -8.5% |

Given these results, I'm convinced now we should remove the expire table prefetch. With the optimization to check the expiration time before accessing the table, we'll only need to access it when an object has actually expired. This change shows significant performance improvements, especially for GET.

Note that SET operations also show better performance as currently we check for `expireIfNeeded` despite the key being about to be overwritten.
@ranshid @zuiderkwast

src/db.c
```diff
@@ -331,16 +330,36 @@ static valkey *dbSetValue(serverDb *db, robj *key, robj *val, int overwrite, voi
...
+        long long expire = objectGetExpire(old);
+        new = objectSetKeyAndExpire(val, key->ptr, expire);
+        *oldref = new;
+        /* Replace the old value at its location in the expire space. */
+        if (expire >= 0) {
+            int dict_index = getKVStoreIndexForKey(key->ptr);
+            void **expireref = kvstoreHashtableFindRef(db->expires, dict_index, key->ptr);
+            serverAssert(expireref != NULL);
+            *expireref = new;
+        }
```
uriyage, line 360: @zuiderkwast 
Perf results for this change (+8.4% for SET commands)

| Operation | Current (K ops/sec) | Previous (K ops/sec) | Change | Notes |
|-----------|----------|-----------|---------|--------|
| SET | 1,120 | 1,033 | +8.4% | Below unstable |
| GET | 1,115 | 1,115 | No change | As expected |
| SET (Unstable) | 1,170 | - | +4.3% | |


SET commands are still below Unstable results of 1170K

zuiderkwast (approved):

PR #1334, opened 2024-11-21
Support empty callback on function and free temp function in async way

enjoy-binbin:
We have a replicationEmptyDbCallback, it is a callback used by emptyData
while flushing away old data. Previously, we did not add this callback
logic for function, in case of abuse, there may be a lot of functions,
and also to make the code consistent, we add the same callback logic
for function.

Changes around this commit:
1. Extend emptyData / functionsLibCtxClear to support passing callback
   when flushing functions.
2. Added disklessLoad function create and discard helper function, just
   like disklessLoadInitTempDb and disklessLoadDiscardTempDb), we wll
   always flush the temp function in a async way to avoid any block.
3. Cleanup around discardTempDb, remove the callback pointer since in
   async way we don't need the callback.
4. Remove functionsLibCtxClear call in readSyncBulkPayload, because we
   called emptyData in the previous lines, which also empty functions.

We are doing this callback in replication is because during the flush,
replica may block a while if the flush is doing in the sync way, to avoid
the primary to detect the replica is timing out, replica will use this
callback to notify the primary (we also do this callback when loading
a RDB). And in the async way, we empty the data in the bio and there is
no slw operation, so it will ignores the callback.

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -572,7 +572,7 @@ long long emptyData(int dbnum, int flags, void(callback)(dict *)) {
 
     if (with_functions) {
         serverAssert(dbnum == -1);
-        functionsLibCtxClearCurrent(async);
+        functionsLibCtxClearCurrent(async, callback);
     }
 
     /* Also fire the end event. Note that this event will fire almost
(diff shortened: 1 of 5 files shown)


zuiderkwast (approved):
Make sense, but I had to look up some things that are not explained in the PR. It can be good to mention in the PR comment:

* Why do we have this callback and what is it used for. I think it's only to notify the primary that the replica is still alive when it is doing a slow operation, right?

* emptyDbStructure ignores the callback if the async parameter is set. Why? Because replica doesn't need to send alive messages to the primary in this case?

enjoy-binbin:
> Why do we have this callback and what is it used for. I think it's only to notify the primary that the replica is still alive when it is doing a slow operation, right?

yes, when doing the flush in a sync way, replica will notify the primary to keep the line online.

> emptyDbStructure ignores the callback if the async parameter is set. Why? Because replica doesn't need to send alive messages to the primary in this case?

yes, in the async way, we empty the data in the bio (it also don't have a way to call the callback), there is no slow operation so no need to notify the primary druing the flush.

zuiderkwast:
Yes, i see you mentioned this in the pr description. looks good, thanks. Let's merge.

PR #1337, opened 2024-11-21
Hashtable feature: Optimize replace large string in db

zuiderkwast:
Avoid allocations when replacing a large string using SET.

Keep the old object in the database and new value are both large strings, just update its pointer to the new string, rather than replacing the whole object.

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -331,6 +331,21 @@ static valkey *dbSetValue(serverDb *db, robj *key, robj *val, int overwrite, voi
         /* Because of VM_StringDMA, old may be changed, so we need get old again */
         old = *oldref;
     }
+
+    /* Can we keep the old object in the database and just replace the ptr? */
+    if ((old->refcount == 1 && old->type == OBJ_STRING && old->encoding != OBJ_ENCODING_EMBSTR) &&
+        (val->refcount == 1 && val->type == OBJ_STRING && val->encoding != OBJ_ENCODING_EMBSTR)) {
+        if (old->encoding == OBJ_ENCODING_RAW) {
+            /* TODO: Offload sdsfree to io thread. */
+            sdsfree(old->ptr);
+        }
+        old->encoding = val->encoding;
+        old->ptr = val->ptr;
+        val->ptr = NULL;
+        decrRefCount(val);
+        return old;
+    }
+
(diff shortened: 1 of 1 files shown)


zuiderkwast:
@uriyage @ranshid

src/db.c
```diff
@@ -331,6 +331,21 @@ static valkey *dbSetValue(serverDb *db, robj *key, robj *val, int overwrite, voi
...
+    if ((old->refcount == 1 && old->type == OBJ_STRING && old->encoding != OBJ_ENCODING_EMBSTR) &&
+        (val->refcount == 1 && val->type == OBJ_STRING && val->encoding != OBJ_ENCODING_EMBSTR)) {
+        if (old->encoding == OBJ_ENCODING_RAW) {
+            /* TODO: Offload sdsfree to io thread. */
+            sdsfree(old->ptr);
+        }
+        old->encoding = val->encoding;
+        old->ptr = val->ptr;
+        val->ptr = NULL;
+        decrRefCount(val);
```
zuiderkwast: Lazy free and io threads offload free are based on robj.

I see that this code is used in some case like SPOP with count. An old set is replaced with a new set. To offload free of the old set, I have a simple idea: We can just swap the ptr and store the old ptr in the val->ptr. Then we can offload free of val.

WDYT?

```suggestion
    if ((old->refcount == 1 && old->encoding != OBJ_ENCODING_EMBSTR) &&
        (val->refcount == 1 && val->encoding != OBJ_ENCODING_EMBSTR)) {
        int tmp_type = old->type;
        int tmp_encoding = old->encoding;
        void *tmp_ptr = old->ptr;
        old->type = val->type;
        old->encoding = val->encoding;
        old->ptr = val->ptr;
        val->type = tmp_type;
        val->encoding = tmp_encoding;
        val->ptr = tmp_ptr;
        if (tryOffloadFreeObjToIOThreads(val) == C_OK) {
            /* OK */
        } else if (server.lazyfree_lazy_server_del) {
            freeObjAsync(key, val, db->id);
        } else {
            decrRefCount(val);
        }
```

zuiderkwast: I pushed a refined variant of this idea in a new commit.

zuiderkwast:
I included this change in #1333

PR #1338, opened 2024-11-21
Offload TLS negotiation to I/O threads

uriyage:
### Overview
This PR introduces the ability to offload TLS handshake negotiations to I/O threads, significantly improving performance under high TLS connection loads.

### Key Changes
- Added infrastructure to offload TLS negotiations to I/O threads
- Refactored SSL event handling to allow I/O threads modify conn flags.
- Introduced new connection flag to identify client connections

### Performance Impact
Testing with 650 clients with SET commands and 160 new TLS connections per second in the background:

#### Throughput Impact of new TLS connections
- **With Offloading**: Minimal impact (1050K → 990K ops/sec)
- **Without Offloading**: Significant drop (1050K → 670K ops/sec)

#### New Connection Rate
- **With Offloading**: 
  - 1,757 conn/sec
- **Without Offloading**: 
  - 477 conn/sec

### Implementation Details
1. **Main Thread**:
   - Initiates negotiation-offload jobs to I/O threads
   - Adds connections to pending-read clients list (using existing read offload mechanism)
   - Post-negotiation handling:
     - Creates read/write events if needed for incomplete negotiations
     - Calls accept handler for completed negotiations

2. **I/O Thread**:
   - Performs TLS negotiation
   - Updates connection flags based on negotiation result

Related issue:https://github.com/valkey-io/valkey/issues/761

### DIFF
--- a/src/connection.h
+++ b/src/connection.h
@@ -54,9 +54,9 @@ typedef enum {
     CONN_STATE_ERROR
 } ConnectionState;
 
-#define CONN_FLAG_CLOSE_SCHEDULED (1 << 0) /* Closed scheduled by a handler */
-#define CONN_FLAG_WRITE_BARRIER (1 << 1)   /* Write barrier requested */
-#define CONN_FLAG_NO_OFFLOAD (1 << 2)      /* Connection should not be offloaded to IO threads. */
+#define CONN_FLAG_CLOSE_SCHEDULED (1 << 0)      /* Closed scheduled by a handler */
+#define CONN_FLAG_WRITE_BARRIER (1 << 1)        /* Write barrier requested */
+#define CONN_FLAG_ALLOW_ACCEPT_OFFLOAD (1 << 2) /* Connection accept can be offloaded to IO threads. */
 
 #define CONN_TYPE_SOCKET "tcp"
 #define CONN_TYPE_UNIX "unix"
--- a/src/connection.h
+++ b/src/connection.h
@@ -56,7 +56,7 @@ typedef enum {
 
 #define CONN_FLAG_CLOSE_SCHEDULED (1 << 0) /* Closed scheduled by a handler */
 #define CONN_FLAG_WRITE_BARRIER (1 << 1)   /* Write barrier requested */
-#define CONN_FLAG_CLIENT (1 << 2)          /* Connection is of a client - not a cluster link. */
+#define CONN_FLAG_NO_OFFLOAD (1 << 2)      /* Connection should not be offloaded to IO threads. */
 
 #define CONN_TYPE_SOCKET "tcp"
 #define CONN_TYPE_UNIX "unix"
--- a/src/connection.h
+++ b/src/connection.h
@@ -56,6 +56,7 @@ typedef enum {
 
 #define CONN_FLAG_CLOSE_SCHEDULED (1 << 0) /* Closed scheduled by a handler */
 #define CONN_FLAG_WRITE_BARRIER (1 << 1)   /* Write barrier requested */
+#define CONN_FLAG_CLIENT (1 << 2)          /* Connection is of a client - not a cluster link. */
 
 #define CONN_TYPE_SOCKET "tcp"
 #define CONN_TYPE_UNIX "unix"
--- a/src/io_threads.c
+++ b/src/io_threads.c
@@ -9,8 +9,6 @@
 static __thread int thread_id = 0; /* Thread local var */
 static pthread_t io_threads[IO_THREADS_MAX_NUM] = {0};
 static pthread_mutex_t io_threads_mutex[IO_THREADS_MAX_NUM];
-typedef void (*tls_negotiation_callback)(void *);
-tls_negotiation_callback tls_negotiation_cb;
 
 /* IO jobs queue functions - Used to send jobs from the main-thread to the IO thread. */
 typedef void (*job_handler)(void *);
@@ -557,29 +555,29 @@ void trySendPollJobToIOThreads(void) {
     IOJobQueue_push(jq, IOThreadPoll, server.el);
 }
 
-void setTLSNegotiationCallback(tls_negotiation_callback cb) {
-    tls_negotiation_cb = cb;
-}
-
-static void ioThreadTLSNegotiation(void *data) {
+static void ioThreadAccept(void *data) {
     client *c = (client *)data;
-    tls_negotiation_cb(c->conn);
+    connAccept(c->conn, NULL);
     c->io_read_state = CLIENT_COMPLETED_IO;
 }
 
 /*
- * This function attempts to offload TLS negotiation for a client connection to an I/O thread.
- * Returns C_OK if the TLS negotiation was successfully queued for processing by an I/O thread,
- * or C_ERR if the client is not eligible for offloading.
+ * Attempts to offload an Accept operation (currently used for TLS accept) for a client
+ * connection to I/O threads.
+ *
+ * Returns:
+ *   C_OK  - If the accept operation was successfully queued for processing
+ *   C_ERR - If the connection is not eligible for offloading
+ *
  * Parameters:
- *   conn: The connection object for which TLS negotiation should be performed
+ *   conn - The connection object to perform the accept operation on
  */
-int trySendTLSNegotiationToIOThreads(connection *conn) {
+int trySendAcceptToIOThreads(connection *conn) {
     if (server.io_threads_num <= 1) {
         return C_ERR;
     }
 
-    if (!(conn->flags & CONN_FLAG_NO_OFFLOAD)) {
+    if (!(conn->flags & CONN_FLAG_ALLOW_ACCEPT_OFFLOAD)) {
         return C_ERR;
     }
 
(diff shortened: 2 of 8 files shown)


src/io_threads.c
```diff
@@ -9,6 +9,7 @@
 static __thread int thread_id = 0; /* Thread local var */
 static pthread_t io_threads[IO_THREADS_MAX_NUM] = {0};
 static pthread_mutex_t io_threads_mutex[IO_THREADS_MAX_NUM];
+void (*tls_negotiation_cb)(void *);
```
madolson: Can we make a typdef for this, it'll be a bit more readable.

src/connection.h
```diff
@@ -56,6 +56,7 @@ typedef enum {
 
 #define CONN_FLAG_CLOSE_SCHEDULED (1 << 0) /* Closed scheduled by a handler */
 #define CONN_FLAG_WRITE_BARRIER (1 << 1)   /* Write barrier requested */
+#define CONN_FLAG_CLIENT (1 << 2)          /* Connection is of a client - not a cluster link. */
```
madolson:
```suggestion
#define CONN_FLAG_NO_OFFLOAD (1 << 2)          /* Connection should not be offloaded to IO threads. */
```
The goal was to keep the connection layer independent of the the server layer.

src/io_threads.c
```diff
@@ -554,3 +555,56 @@ void trySendPollJobToIOThreads(void) {
...
+    }
+
+    client *c = connGetPrivateData(conn);
+    if (c->io_read_state != CLIENT_IDLE) {
+        return C_OK;
+    }
+
+    if (server.active_io_threads_num <= 1) {
+        return C_ERR;
+    }
```
```suggestion
    if (server.active_io_threads_num <= 1) {
        return C_ERR;
    }

    if (!(conn->flags & CONN_FLAG_CLIENT)) {
        return C_ERR;
    }

    client *c = connGetPrivateData(conn);
    if (c->io_read_state != CLIENT_IDLE) {
        return C_OK;
    }
```

madolson: It seems everywhere else we just check for active threads.

madolson (commented):
This implementation basically breaks the connection abstraction, since it has the TLS implementation calling functions related to IO threading (which is supposed to be agnostic to the connection type). I was sort of expecting we would offload the event handler.

uriyage:
> This implementation basically breaks the connection abstraction, since it has the TLS implementation calling functions related to IO threading (which is supposed to be agnostic to the connection type). I was sort of expecting we would offload the event handler.

We still don't check in any point for the connection type, since the TLS code calls the IO threads to offload the negotiation with a supplied callback, not the other way around. Maybe we can rename it to 'accept' instead of 'tls_negotiate' to be more abstract. 

>  I was sort of expecting we would offload the event handler.

Not sure I get this, would you please elaborate.

src/io_threads.c
uriyage: Thanks, Done.

src/connection.h
uriyage: Changed.

src/io_threads.c
uriyage: Consider a scenario where the main thread sends a job to the IO thread. After the IO thread completes the job, the main thread deactivates all threads. In this case, we want to ensure the main thread processes the returned job from the IO thread before performing an accept operation even if the io threads are not active.

To achieve this:

First, check if IO threads are enabled at all (less expensive check)
Then, verify the read_state is not idle
Finally, check the active state

Unlike read/write where we anyway check for the read/write states, the accept flow operates at the connection layer rather than the client layer, so we can't check the read state there.

src/tls.c
```diff
@@ -116,6 +119,8 @@ static list *pending_list = NULL;
 
 static pthread_mutex_t *openssl_locks;
 
+static void TLSNegotiate(void *conn);
```
ranshid: what did I miss? why is this defined twice? (first in line 65)?

src/io_threads.c
```diff
@@ -9,6 +9,8 @@
 static __thread int thread_id = 0; /* Thread local var */
 static pthread_t io_threads[IO_THREADS_MAX_NUM] = {0};
 static pthread_mutex_t io_threads_mutex[IO_THREADS_MAX_NUM];
+typedef void (*tls_negotiation_callback)(void *);
+tls_negotiation_callback tls_negotiation_cb;
```
ranshid: IMO this whole callback addition is over complicated. I understand the motivation was to avoid exposing iothreads jobqueue internals to TLS (non-module) and allow compiling io-threads without TLS (non-module). But I think it would leave the code much cleaner when we just implememt the trySendTLSNegotiationToIOThreads under USE_OPENSSL. Alternative TLS implementation would probably not be able to use that API in any case, and I think in the future we might want to offer modules with API to coordinate actions to io-threads.

src/server.h
```diff
@@ -2767,6 +2768,7 @@ void dictVanillaFree(void *val);
 #define READ_FLAGS_PRIMARY (1 << 14)
 #define READ_FLAGS_DONT_PARSE (1 << 15)
 #define READ_FLAGS_AUTH_REQUIRED (1 << 16)
+#define READ_FLAGS_TLS_NEGOTIATION (1 << 17)
```
ranshid: why can't this be a connection level flag?

src/networking.c
```diff
@@ -4725,6 +4726,9 @@ int processIOThreadsReadDone(void) {
         connSetPostponeUpdateState(c->conn, 0);
         connUpdateState(c->conn);
 
+        /* No client's data was read only TLS handshake. */
+        if (c->read_flags & READ_FLAGS_TLS_NEGOTIATION) continue;
+
```
ranshid: I do not like the fact that we mix io-threads and openssl specific logic with generic io-threads logic. maybe we can place a dedicated function which will default to always return C_OK when USE_OPENSSL is not defined?

ranshid (commented):
Uri. We discussed offline about several architectual changes to reduce the code changes. Also placed some comments I think we can improve.

src/tls.c
uriyage: Thanks, I missed that, removed.

src/server.h
uriyage: Removed, we will check for the conn accepting state.

src/networking.c
uriyage: Changed to check for the conn state.

src/io_threads.c
uriyage: Thanks,
I removed the callback mechanism.
The I/O thread now directly calls connAccept instead of using callbacks.

src/tls.c
```diff
@@ -698,28 +730,8 @@ static void tlsHandleEvent(tls_connection *conn, int mask) {
...
-            /* If not handled, it's an error */
-            conn->c.state = CONN_STATE_ERROR;
-        } else {
-            conn->c.state = CONN_STATE_CONNECTED;
-        }
-
-        if (!callHandler((connection *)conn, conn->c.conn_handler)) return;
-        conn->c.conn_handler = NULL;
-        break;
+        connTLSAccept((connection *)conn, NULL);
```
ranshid: I have some concerns about ignoring the return value and ALWAYS skip break and run then updateSLL. maybe in some cases we complete immidiately and need the update to run?

src/tls.c
```diff
@@ -698,28 +730,8 @@ static void tlsHandleEvent(tls_connection *conn, int mask) {
...
+        return;
```
ranshid: can we add comment explaining why we decide to return here like was done in line 715?

src/tls.c
uriyage: Restored the same logic as before.

src/tls.c
uriyage: Changed per your previous command

ranshid: "comment". I never **command** anyone (only my wife does :) )

src/tls.c
```diff
@@ -650,12 +649,46 @@ static void updateSSLEvent(tls_connection *conn) {
     if (!need_write && (mask & AE_WRITABLE)) aeDeleteFileEvent(server.el, conn->c.fd, AE_WRITABLE);
 }
 
+static int TLSHandleAcceptResult(tls_connection *conn, int call_handler_on_error) {
```
ranshid, line 652: can we assert here if we are NOT in CONN_STATE_ACCEPTING ?

uriyage: You mean we are IN CONN_STATE_ACCEPTING, right?

ranshid:
@uriyage it looks much better now IMO. I still want @madolson to also take a better look since these parts are very fragile and hard to miss some cases. Let's also wait for her to provide feedback

ranshid (approved):

src/server.c
```diff
@@ -5862,6 +5863,7 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "io_threaded_reads_processed:%lld\r\n", server.stat_io_reads_processed,
                 "io_threaded_writes_processed:%lld\r\n", server.stat_io_writes_processed,
                 "io_threaded_freed_objects:%lld\r\n", server.stat_io_freed_objects,
+                "io_threaded_accept:%lld\r\n", server.stat_io_accept_offloaded,
```
```suggestion
                "io_threaded_accept_processed:%lld\r\n", server.stat_io_accept_offloaded,
```
For consistency?

madolson (approved):

PR #1340, opened 2024-11-22
Add cmake-build-debug and cmake-build-release to gitignore

enjoy-binbin:


madolson (approved):
Yeah, I originally talked with @eifrah-aws about this and he thought it would be fine to skip adding these, but it's probably for the best to include these.

PR #1342, opened 2024-11-22
Optimize sdscatrepr by batch processing printable characters

RayaCoo:
[#11725](https://github.com/redis/redis/pull/11725 ) optimize sdscatrepr by reducing realloc calls, furthermore, we can reduce memcpy calls by batch processing of consecutive printable characters.

### DIFF
--- a/src/sds.c
+++ b/src/sds.c
@@ -954,23 +954,30 @@ void sdsfreesplitres(sds *tokens, int count) {
 sds sdscatrepr(sds s, const char *p, size_t len) {
     s = sdsMakeRoomFor(s, len + 2);
     s = sdscatlen(s, "\"", 1);
-    while (len--) {
-        switch (*p) {
-        case '\\':
-        case '"': s = sdscatprintf(s, "\\%c", *p); break;
-        case '\n': s = sdscatlen(s, "\\n", 2); break;
-        case '\r': s = sdscatlen(s, "\\r", 2); break;
-        case '\t': s = sdscatlen(s, "\\t", 2); break;
-        case '\a': s = sdscatlen(s, "\\a", 2); break;
-        case '\b': s = sdscatlen(s, "\\b", 2); break;
-        default:
-            if (isprint(*p))
-                s = sdscatlen(s, p, 1);
-            else
+    while (len) {
+        if (isprint(*p)) {
+            const char *start = p;
+            while (len && isprint(*p)) {
+                len--;
+                p++;
+            }
+            s = sdscatlen(s, start, p - start);
+        } else {
+            switch (*p) {
+            case '\\':
+            case '"': s = sdscatprintf(s, "\\%c", *p); break;
+            case '\n': s = sdscatlen(s, "\\n", 2); break;
+            case '\r': s = sdscatlen(s, "\\r", 2); break;
+            case '\t': s = sdscatlen(s, "\\t", 2); break;
+            case '\a': s = sdscatlen(s, "\\a", 2); break;
+            case '\b': s = sdscatlen(s, "\\b", 2); break;
+            default:
                 s = sdscatprintf(s, "\\x%02x", (unsigned char)*p);
-            break;
+                break;
+            }
+            p++;
+            len--;
         }
-        p++;
     }
     return sdscatlen(s, "\"", 1);
 }


madolson:
I wonder if we should do something like https://github.com/valkey-io/valkey/pull/1230, which removes all the reallocation and just does a single initial allocation. Checking through the code, this doesn't seem to be anywhere on the hot path though, so that is probably over optimizing.

RayaCoo:
As @artikell said in [#11725](https://github.com/redis/redis/pull/11725), If there is a monitor, `replicationFeedMonitors` will 
 call `sdscatrepr` frequently. So I use `valkey-benchmark --threads 2 -P 10 -r 100 -n 10000000 getset __rand_int__ __rand_int__` to finish the test as @artikell did before.
#### unstable (no monitor)
```
Summary:
  throughput summary: 714081.69 requests per second
  latency summary (msec):
        0.654     0.112     0.687     0.847     0.879     3.247
```
#### unstable (1 monitor)
```
Summary:
  throughput summary: 332967.06 requests per second
  latency summary (msec):
        1.449     0.384     1.647     1.855     1.927     4.263
```
#### PR (no monitor)
```
Summary:
  throughput summary: 714030.69 requests per second
  latency summary (msec):
        0.645     0.128     0.671     0.839     0.871     2.839
```
#### PR (1 monitor)
```
Summary:
  throughput summary: 395116.38 requests per second
  latency summary (msec):
        1.221     0.248     1.367     1.575     1.631     3.735
```

- The test results show about an **18%** improvement with one monitor.

RayaCoo:
> I wonder if we should do something like #1230, which removes all the reallocation and just does a single initial allocation. Checking through the code, this doesn't seem to be anywhere on the hot path though, so that is probably over optimizing.

@madolson 
First, I don't think `sdscatrepr` can "just does a single initial allocation" , as we don't know the final length of the SDS in advance unless we first iterate through `p` to calculate it. what [#11725](https://github.com/redis/redis/pull/11725) does is pre-allocate memory to avoid frequent reallocations.
Secondly, `sdscatrepr` is frequently called in our product for auditing purposes. I have also seen some community discussions about audit-related issues, like [Improved Redis auditing support](https://github.com/redis/redis/issues/10957), If we consider extending the `monitor` command in the future, `sdscatrepr` might become a hot path. 
Additionally, I believe the implementation of sdscatrepr is not efficient enough. Printable characters typically appear consecutively, and processing one printable character at a time with frequent calls to `memcpy` is unnecessary.

madolson:
> Additionally, I believe the implementation of sdscatrepr is not efficient enough. Printable characters typically appear consecutively, and processing one printable character at a time with frequent calls to memcpy is unnecessary.

I agree with optimizing the code, I was wondering we should spend more effort to optimize is than just what you are proposing in this PR. 

> First, I don't think sdscatrepr can "just does a single initial allocation" , as we don't know the final length of the SDS in advance unless we first iterate through p to calculate it. what https://github.com/redis/redis/pull/11725 does is pre-allocate memory to avoid frequent reallocations.

Yeah, I was suggesting that we could do two passes. Two passes seems like it would be slower, but the work I did with sdscat showed that it was still much faster both because sdscat* functions are slow and because if we have to realloc that is also really slow. If you aren't interesting in trying any alternatives, I think this is worth merging as is though. (I'll take a look at the actual code shortly)

soloestoy:
Currently, in `sdscatrepr`, we first use `sdsMakeRoomFor` to pre-allocate memory. `sdsMakeRoomFor` is greedy by default, I think that in most scenarios, there is already enough redundant space, and there's no need to use realloc again.

And in my understanding, the use cases for `sdscatrepr` are usually to generate temporary data that won't persist long-term in the server. Or do you think it's necessary for us to use `sdsMakeRoomForNonGreedy` instead of `sdsMakeRoomFor`? At that point, getting the exact length is essential.

madolson (approved):

madolson:
> And in my understanding, the use cases for sdscatrepr are usually to generate temporary data that won't persist long-term in the server. Or do you think it's necessary for us to use sdsMakeRoomForNonGreedy instead of sdsMakeRoomFor? At that point, getting the exact length is essential.

I'm not as concerned about extra memory usage. I was more interested if we could improve the performance if we avoided sds functions, which tend to be a bit slow compared to direct character manipulation. I'm fine with the PR as is though.

PR #1347, opened 2024-11-25
Save open's errno when opening temp rdb fails to prevent it from being modified

enjoy-binbin:
Apparently on Mac, sleep will modify errno to ETIMEDOUT, and then it
prints the misleading message: Operation timed out.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -3684,13 +3684,18 @@ void syncWithPrimary(connection *conn) {
 
     /* Prepare a suitable temp file for bulk transfer */
     if (!useDisklessLoad()) {
+        /* We save the errno of open to prevent some systems from modifying it after
+         * the sleep call. For example, sleep in Mac will change errno to ETIMEDOUT. */
+        int saved_errno;
         while (maxtries--) {
             snprintf(tmpfile, 256, "temp-%d.%ld.rdb", (int)server.unixtime, (long int)getpid());
             dfd = open(tmpfile, O_CREAT | O_WRONLY | O_EXCL, 0644);
             if (dfd != -1) break;
+            saved_errno = errno;
             sleep(1);
         }
         if (dfd == -1) {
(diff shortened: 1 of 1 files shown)


src/replication.c
```diff
@@ -3684,13 +3684,18 @@ void syncWithPrimary(connection *conn) {
 
     /* Prepare a suitable temp file for bulk transfer */
     if (!useDisklessLoad()) {
+        /* We save the errno of open to prevent some systems from modifying it after
+         * the sleep call. For example, sleep in Mac will change errno to ETIMEDOUT. */
+        int saved_errno;
         while (maxtries--) {
             snprintf(tmpfile, 256, "temp-%d.%ld.rdb", (int)server.unixtime, (long int)getpid());
             dfd = open(tmpfile, O_CREAT | O_WRONLY | O_EXCL, 0644);
             if (dfd != -1) break;
```
enjoy-binbin, line 3690: another choice is that we can print the error message in each try, so we don't need this saved_errno. The cost is that we may print 5 times of the log.

src/replication.c
```diff
@@ -3684,13 +3684,18 @@ void syncWithPrimary(connection *conn) {
...
+            saved_errno = errno;
             sleep(1);
```
zuiderkwast: Another possibility is to save use the saved errno only during sleep. Smaller scope is simpler code?

```suggestion
            /* Bla bla bla MacOS bla bla ETIMEDOUT. */
            int saved_errno = errno;
            sleep(1);
            errno = save_errno;
```

zuiderkwast (approved):

src/replication.c
zuiderkwast: We could, but it's not needed, right?

enjoy-binbin: yes, it is no needed

PR #1349, opened 2024-11-25
Avoid double close on repl_transfer_fd

enjoy-binbin:
The code is ok before 2de544cfcc6d1aa7cf6d0c75a6116f7fc27b6fd6,
but now we will set server.repl_transfer_fd right after dfd was
initiated, and in here we have a double close error since dfd and
server.repl_transfer_fd are the same fd.

Also move the declaration of dfd/maxtries to a small scope to avoid
the confusion since they are only used in this code.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -3740,7 +3740,6 @@ no_response_error: /* Handle receiveSynchronousResponse() error when primary has
     /* Fall through to regular error handling */
 
 error:
-    if (dfd != -1) close(dfd);
     connClose(conn);
     server.repl_transfer_s = NULL;
     if (server.repl_rdb_transfer_s) {
--- a/src/replication.c
+++ b/src/replication.c
@@ -3414,7 +3414,6 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
  * establish a connection with the primary. */
 void syncWithPrimary(connection *conn) {
     char tmpfile[256], *err = NULL;
-    int dfd = -1, maxtries = 5;
     int psync_result;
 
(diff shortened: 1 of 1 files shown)


src/replication.c
```diff
@@ -3740,7 +3740,6 @@ void syncWithPrimary(connection *conn) {
     /* Fall through to regular error handling */
 
 error:
-    if (dfd != -1) close(dfd);
```
zuiderkwast, line 3743: `dfd` is only used in this code:

```C
    /* Prepare a suitable temp file for bulk transfer */
    if (!useDisklessLoad()) {
        while (maxtries--) {
            snprintf(tmpfile, 256, "temp-%d.%ld.rdb", (int)server.unixtime, (long int)getpid());
            dfd = open(tmpfile, O_CREAT | O_WRONLY | O_EXCL, 0644);
            if (dfd != -1) break;
            sleep(1);
        }
        if (dfd == -1) {
            serverLog(LL_WARNING, "Opening the temp file needed for PRIMARY <-> REPLICA synchronization: %s",
                      strerror(errno));
            goto error;
        }
        server.repl_transfer_tmpfile = zstrdup(tmpfile);
        server.repl_transfer_fd = dfd;
    }
```

Let's move the declaration of `dfd` to this small scope to avoid the confusion.

zuiderkwast (commented):
Makes sense, but it's difficult to see that in this very long function.

src/replication.c
enjoy-binbin: yes! it also hunting me me!

zuiderkwast (approved):

PR #1351, opened 2024-11-25
allow READONLY&READWRITE only after CLIENT CAPA REDIRECT for standalone mode

### LINKED ISSUE #1348: READONLY not returning an error in standalone mode broke some test case for Valkey-py.
READONLY not returning an error in standalone mode broke some test case for Valkey-py.

Reported by @mkmkme:

> Yeah, the test started failing because the test was as simple as
>
>     async def test_readonly_invalid_cluster_state(self, r: valkey.Valkey):
>         with pytest.raises(exceptions.ValkeyError):
>             await r.readonly()
>
> So it assumed readonly command always returns an error

I guess we should have added a check for CLIENT CAPA REDIRECT in the READONLY and READWRITE commands, to keep the old behaviour for old clients.

_Originally posted by @zuiderkwast in https://github.com/valkey-io/valkey/issues/325#issuecomment-2494305316_

soloestoy:
fixes #1348

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1472,9 +1472,7 @@ void askingCommand(client *c) {
  * with read-only commands to keys that are served by the replica's primary. */
 void readonlyCommand(client *c) {
     if (server.cluster_enabled == 0 && !(c->capa & CLIENT_CAPA_REDIRECT)) {
-        addReplyError(c, "This instance has cluster support disabled,"
-                         " you need to execute the CLIENT CAPA REDIRECT command,"
-                         " before you can use the READONLY command in standalone mode.");
+        addReplyError(c, "Redirects are not enabled");
         return;
     }
     c->flag.readonly = 1;
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1471,12 +1471,24 @@ void askingCommand(client *c) {
  * In this mode replica will not redirect clients as long as clients access
  * with read-only commands to keys that are served by the replica's primary. */
 void readonlyCommand(client *c) {
+    if (server.cluster_enabled == 0 && !(c->capa & CLIENT_CAPA_REDIRECT)) {
+        addReplyError(c, "This instance has cluster support disabled,"
+                         " you need to execute the CLIENT CAPA REDIRECT command,"
+                         " before you can use the READONLY command in standalone mode.");
+        return;
+    }
     c->flag.readonly = 1;
     addReply(c, shared.ok);
 }
 
 /* The READWRITE command just clears the READONLY command state. */
 void readwriteCommand(client *c) {
+    if (server.cluster_enabled == 0 && !(c->capa & CLIENT_CAPA_REDIRECT)) {
+        addReplyError(c, "This instance has cluster support disabled,"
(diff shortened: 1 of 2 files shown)


zuiderkwast:
I'm not sure we should fix this. See the discussion in #1348.

src/cluster.c
```diff
@@ -1471,12 +1471,24 @@ void askingCommand(client *c) {
  * In this mode replica will not redirect clients as long as clients access
  * with read-only commands to keys that are served by the replica's primary. */
 void readonlyCommand(client *c) {
+    if (server.cluster_enabled == 0 && !(c->capa & CLIENT_CAPA_REDIRECT)) {
+        addReplyError(c, "This instance has cluster support disabled,"
+                         " you need to execute the CLIENT CAPA REDIRECT command,"
+                         " before you can use the READONLY command in standalone mode.");
```
zuiderkwast: These very long error messages that include the full documentation are not my favourite style.

If you (a user) get an error, you would normally go to the documentation for the READONLY command and find out what it means.

A more concise way is to say that READONLY is allowed only if redirects are enabled. This includes cluster (always) and standalone (with client capa redirect).

```suggestion
        addReplyError(c, "Redirects are not enabled");
```

zuiderkwast (commented):
I'm convinced we should do this fix, especially for compatibility with other databases (kvrocks, keydb, who knows). Many clients support not only Valkey.

Only one comment is about the error message.

soloestoy:
> especially for compatibility with other databases (kvrocks, keydb, who knows).

I think the order is reversed, they should be compatible with us, not the other way around, haha.

zuiderkwast:
> > especially for compatibility with other databases (kvrocks, keydb, who knows).
> 
> I think the order is reversed, they should be compatible with us, not the other way around, haha.

Yes, we can think that we are the best and greatest, but this is like Redis style thinking. I don't want to think in this way.

zuiderkwast (approved):

src/cluster.c
```diff
@@ -1471,12 +1471,20 @@ void askingCommand(client *c) {
  * In this mode replica will not redirect clients as long as clients access
  * with read-only commands to keys that are served by the replica's primary. */
 void readonlyCommand(client *c) {
+    if (server.cluster_enabled == 0 && !(c->capa & CLIENT_CAPA_REDIRECT)) {
```
madolson, line 1474: It feels weird to me to require sending the `CLIENT CAPA REDIRECT` first before readonly. I feel like clients should be able to send them in either order.

Overall i'm inclined to leave the code as is.

This is also a breaking change from 8.0, which I don't want to do. (I think we would have to postpone this to  9.0)

PingXie: >It feels weird to me to require sending the CLIENT CAPA REDIRECT first before readonly. I feel like clients should be able to send them in either order.

It is not really about the order but more about compatibility, IMO. Without `CLIENT CAPA REDIRECT`, we fall back to the original behavior on standalone. 

 > This is also a breaking change from 8.0, which I don't want to do.
 
If we classify this as a bug, we should back port the fix to 8.0 and then the behavior would be the same across 7.2, 8.0, and future versions.

PR #1356, opened 2024-11-26
Automatic failover vote is not limited by two times the node timeout

enjoy-binbin:
This is a follow of #1305, we now decided to apply the same change
to automatic failover as well, that is, move forward with removing
it for both automatic and manual failovers.

Quote from Ping during the review:
Note that we already debounce transient primary failures with node
timeout, ensuring failover is only triggered after sustained outages.
Election timing is naturally staggered by replica spacing, making the
likelihood of simultaneous elections from replicas of the same shard
very low. The one-vote-per-epoch rule further throttles retries and
ensures orderly elections. On top of that, quorum-based primary failure
confirmation, cluster-state convergence, and slot ownership validation
are all built into the process.

Quote from Madelyn during the review:
It against the specific primary. It's to prevent double failovers.
If a primary just took over we don't want someone else to try to
take over and give the new primary some amount of time to take over.
I have not seen this issue though, it might have been over optimizing?
The double failure mode, where a node fails and then another node fails
within the nodetimeout also doesn't seem that common either though.

So the conclusion is that we all agreed to remove it completely,
it will make the code a lot simpler. And if there is other specific
edge cases we are missing, we will fix it in other way.

See discussion #1305 for more information.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -1498,7 +1498,6 @@ clusterNode *createClusterNode(char *nodename, int flags) {
     node->cport = 0;
     node->tls_port = 0;
     node->fail_reports = listCreate();
-    node->voted_time = 0;
     node->orphaned_time = 0;
     node->repl_offset_time = 0;
     node->repl_offset = 0;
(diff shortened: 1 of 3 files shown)


PingXie (approved):
Thanks @enjoy-binbin!

zuiderkwast:
@enjoy-binbin and @PingXie you seem convinced it shall be deleted. I didn't think about this enough to really understand it, but I trust you. The code looks good.

I hope @madolson will confirm.

enjoy-binbin:
for the record, Madelyn mention in here https://github.com/valkey-io/valkey/pull/1305#discussion_r1857122581 that she agree with removing this
> I guess I agree with removing it. It seems like it will make the code a lot simpler.

PR #1357, opened 2024-11-26
Add new module API flag to bypass command validation

### LINKED ISSUE #1175: lookupCommandByCString in VM_Replicate

Avoid performance hit in VM_Replicate function call. 

My team recently was trying to optimize the performance of our module. We found this function call `lookupCommandByCString` inside `VM_Replicate` to taking a lot of CPU time.
https://github.com/valkey-io/valkey/blob/unstable/src/module.c#L3539-L3540

The main reason is that it allocates memory to convert const char * into sds, then perform a dict lookup. From `VM_Replicate` function, it looks like it is just doing a sanity check without using it for anything else. My guess is the purpose is to protect the engine from running a 3rd party module. However, the protection seems induces too much performance penalty. If we develop our in-house module and don't run other modules, could we bypass this check to avoid performance hit. Is it possible to add a config flag "trust-module-replication" to bypass this check?

We considered VM_ReplicateVerbatim, but we sometimes needs to rewrite the command during replication. So we could not use it.

**Additional information**

sungming2:
Fixes #1175

### Problem
Performance degradation occurs due to the sanity check([lookupCommandByCString](https://github.com/valkey-io/valkey/blob/unstable/src/module.c#L3539)) in the VM_Replicate function, which converts const char* into sds and free them for command dictionary lookup. This check is mainly used for debugging purpose, but it is unnecessary for trusted modules. The user seeks a way to bypass this check for performance gains.

### Solution
Introduce a new SKIP_COMMAND_VALIDATION option which allows individual modules to opt out of command validation

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -2566,9 +2566,25 @@ void VM_Yield(ValkeyModuleCtx *ctx, int flags, const char *busy_reply) {
  * By default, the server will not fire key-space notifications that happened inside
  * a key-space notification callback. This flag allows to change this behavior
  * and fire nested key-space notifications. Notice: if enabled, the module
- * should protected itself from infinite recursion. */
+ * should protected itself from infinite recursion.
+ *
+ * VALKEYMODULE_OPTIONS_SKIP_COMMAND_VALIDATION:
+ * When set, this option allows the module to skip command validation.
+ * This is useful in scenarios where the module needs to bypass
+ * command validation (lookupCommandByCString) for specific operations
+ * to reduce overhead or handle trusted custom command logic. */
 void VM_SetModuleOptions(ValkeyModuleCtx *ctx, int options) {
     ctx->module->options = options;
+}
+
+/* Add specified options to the module's current options. */
+void VM_AddModuleOptions(ValkeyModuleCtx *ctx, int options) {
+    ctx->module->options |= options;
+}
+
+/* Remove specified options from the module's current options. */
+void VM_RemoveModuleOptions(ValkeyModuleCtx *ctx, int options) {
+    ctx->module->options &= ~options;
 }
 
 /* Signals that the key is modified from user's perspective (i.e. invalidate WATCH
@@ -3630,8 +3646,10 @@ int VM_Replicate(ValkeyModuleCtx *ctx, const char *cmdname, const char *fmt, ...
     int argc = 0, flags = 0, j;
     va_list ap;
 
-    cmd = lookupCommandByCString((char *)cmdname);
-    if (!cmd) return VALKEYMODULE_ERR;
+    if (!ctx->module || !(ctx->module->options & VALKEYMODULE_OPTIONS_SKIP_COMMAND_VALIDATION)) {
+        cmd = lookupCommandByCString((char *)cmdname);
+        if (!cmd) return VALKEYMODULE_ERR;
+    }
 
     /* Create the client and dispatch the command. */
     va_start(ap, fmt);
@@ -7535,15 +7553,17 @@ void VM_EmitAOF(ValkeyModuleIO *io, const char *cmdname, const char *fmt, ...) {
     int argc = 0, flags = 0, j;
     va_list ap;
 
-    cmd = lookupCommandByCString((char *)cmdname);
-    if (!cmd) {
-        serverLog(LL_WARNING,
-                  "Fatal: AOF method for module data type '%s' tried to "
-                  "emit unknown command '%s'",
-                  io->type->name, cmdname);
-        io->error = 1;
-        errno = EINVAL;
-        return;
+    if (!io->ctx || !io->ctx->module || !(io->ctx->module->options & VALKEYMODULE_OPTIONS_SKIP_COMMAND_VALIDATION)) {
+        cmd = lookupCommandByCString((char *)cmdname);
+        if (!cmd) {
+            serverLog(LL_WARNING,
+                      "Fatal: AOF method for module data type '%s' tried to "
+                      "emit unknown command '%s'",
+                      io->type->name, cmdname);
+            io->error = 1;
+            errno = EINVAL;
+            return;
+        }
     }
 
     /* Emit the arguments into the AOF in RESP format. */
@@ -13892,6 +13912,8 @@ void moduleRegisterCoreAPI(void) {
     REGISTER_API(ModuleTypeGetValue);
     REGISTER_API(IsIOError);
     REGISTER_API(SetModuleOptions);
+    REGISTER_API(AddModuleOptions);
+    REGISTER_API(RemoveModuleOptions);
     REGISTER_API(SignalModifiedKey);
     REGISTER_API(SaveUnsigned);
     REGISTER_API(LoadUnsigned);
--- a/src/module.c
+++ b/src/module.c
@@ -2571,20 +2571,13 @@ void VM_Yield(ValkeyModuleCtx *ctx, int flags, const char *busy_reply) {
  * VALKEYMODULE_OPTIONS_SKIP_COMMAND_VALIDATION:
  * When set, this option allows the module to skip command validation.
  * This is useful in scenarios where the module needs to bypass
- * command validation (lookupCommandByCString) for specific operations
- * to reduce overhead or handle trusted custom command logic. */
+ * command validation for specific operations
+ * to reduce overhead or handle trusted custom command logic.
+ * ValkeyModule_Replicate and ValkeyModule_EmitAOF
+ * are affected by this option, allowing them to operate without
+ * command validation check. */
 void VM_SetModuleOptions(ValkeyModuleCtx *ctx, int options) {
     ctx->module->options = options;
-}
-
-/* Add specified options to the module's current options. */
-void VM_AddModuleOptions(ValkeyModuleCtx *ctx, int options) {
-    ctx->module->options |= options;
-}
-
-/* Remove specified options from the module's current options. */
-void VM_RemoveModuleOptions(ValkeyModuleCtx *ctx, int options) {
-    ctx->module->options &= ~options;
 }
 
 /* Signals that the key is modified from user's perspective (i.e. invalidate WATCH
@@ -13912,8 +13905,6 @@ void moduleRegisterCoreAPI(void) {
     REGISTER_API(ModuleTypeGetValue);
     REGISTER_API(IsIOError);
     REGISTER_API(SetModuleOptions);
-    REGISTER_API(AddModuleOptions);
-    REGISTER_API(RemoveModuleOptions);
     REGISTER_API(SignalModifiedKey);
     REGISTER_API(SaveUnsigned);
     REGISTER_API(LoadUnsigned);
--- a/src/valkeymodule.h
+++ b/src/valkeymodule.h
@@ -325,10 +325,15 @@ typedef uint64_t ValkeyModuleTimerID;
  * If enabled, the module is responsible to break endless loop. */
 #define VALKEYMODULE_OPTIONS_ALLOW_NESTED_KEYSPACE_NOTIFICATIONS (1 << 3)
 
+/* Skipping command validation can improve performance by reducing the overhead associated
+ * with command checking, especially in high-throughput scenarios where commands
+ * are already pre-validated or trusted. */
+#define VALKEYMODULE_OPTIONS_SKIP_COMMAND_VALIDATION (1 << 4)
+
 /* Next option flag, must be updated when adding new module flags above!
  * This flag should not be used directly by the module.
  * Use ValkeyModule_GetModuleOptionsAll instead. */
-#define _VALKEYMODULE_OPTIONS_FLAGS_NEXT (1 << 4)
+#define _VALKEYMODULE_OPTIONS_FLAGS_NEXT (1 << 5)
 
 /* Definitions for ValkeyModule_SetCommandInfo. */
 
@@ -1424,6 +1429,8 @@ VALKEYMODULE_API ValkeyModuleType *(*ValkeyModule_ModuleTypeGetType)(ValkeyModul
 VALKEYMODULE_API void *(*ValkeyModule_ModuleTypeGetValue)(ValkeyModuleKey *key)VALKEYMODULE_ATTR;
 VALKEYMODULE_API int (*ValkeyModule_IsIOError)(ValkeyModuleIO *io) VALKEYMODULE_ATTR;
 VALKEYMODULE_API void (*ValkeyModule_SetModuleOptions)(ValkeyModuleCtx *ctx, int options) VALKEYMODULE_ATTR;
+VALKEYMODULE_API void (*ValkeyModule_AddModuleOptions)(ValkeyModuleCtx *ctx, int options) VALKEYMODULE_ATTR;
+VALKEYMODULE_API void (*ValkeyModule_RemoveModuleOptions)(ValkeyModuleCtx *ctx, int options) VALKEYMODULE_ATTR;
 VALKEYMODULE_API int (*ValkeyModule_SignalModifiedKey)(ValkeyModuleCtx *ctx,
                                                        ValkeyModuleString *keyname) VALKEYMODULE_ATTR;
 VALKEYMODULE_API void (*ValkeyModule_SaveUnsigned)(ValkeyModuleIO *io, uint64_t value) VALKEYMODULE_ATTR;
@@ -2034,6 +2041,8 @@ static int ValkeyModule_Init(ValkeyModuleCtx *ctx, const char *name, int ver, in
     VALKEYMODULE_GET_API(ModuleTypeGetValue);
     VALKEYMODULE_GET_API(IsIOError);
     VALKEYMODULE_GET_API(SetModuleOptions);
+    VALKEYMODULE_GET_API(RemoveModuleOptions);
+    VALKEYMODULE_GET_API(AddModuleOptions);
     VALKEYMODULE_GET_API(SignalModifiedKey);
     VALKEYMODULE_GET_API(SaveUnsigned);
     VALKEYMODULE_GET_API(LoadUnsigned);
--- a/src/valkeymodule.h
+++ b/src/valkeymodule.h
@@ -1429,8 +1429,6 @@ VALKEYMODULE_API ValkeyModuleType *(*ValkeyModule_ModuleTypeGetType)(ValkeyModul
 VALKEYMODULE_API void *(*ValkeyModule_ModuleTypeGetValue)(ValkeyModuleKey *key)VALKEYMODULE_ATTR;
 VALKEYMODULE_API int (*ValkeyModule_IsIOError)(ValkeyModuleIO *io) VALKEYMODULE_ATTR;
 VALKEYMODULE_API void (*ValkeyModule_SetModuleOptions)(ValkeyModuleCtx *ctx, int options) VALKEYMODULE_ATTR;
-VALKEYMODULE_API void (*ValkeyModule_AddModuleOptions)(ValkeyModuleCtx *ctx, int options) VALKEYMODULE_ATTR;
-VALKEYMODULE_API void (*ValkeyModule_RemoveModuleOptions)(ValkeyModuleCtx *ctx, int options) VALKEYMODULE_ATTR;
 VALKEYMODULE_API int (*ValkeyModule_SignalModifiedKey)(ValkeyModuleCtx *ctx,
                                                        ValkeyModuleString *keyname) VALKEYMODULE_ATTR;
 VALKEYMODULE_API void (*ValkeyModule_SaveUnsigned)(ValkeyModuleIO *io, uint64_t value) VALKEYMODULE_ATTR;
@@ -2041,8 +2039,6 @@ static int ValkeyModule_Init(ValkeyModuleCtx *ctx, const char *name, int ver, in
     VALKEYMODULE_GET_API(ModuleTypeGetValue);
     VALKEYMODULE_GET_API(IsIOError);
     VALKEYMODULE_GET_API(SetModuleOptions);
-    VALKEYMODULE_GET_API(RemoveModuleOptions);
-    VALKEYMODULE_GET_API(AddModuleOptions);
     VALKEYMODULE_GET_API(SignalModifiedKey);
     VALKEYMODULE_GET_API(SaveUnsigned);
     VALKEYMODULE_GET_API(LoadUnsigned);
--- a/tests/modules/propagate.c
+++ b/tests/modules/propagate.c
@@ -250,7 +250,9 @@ int propagateTestSimpleCommand(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, 
 
     /* Replicate two commands to test MULTI/EXEC wrapping. */
     ValkeyModule_Replicate(ctx,"INCR","c","counter-1");
+    ValkeyModule_AddModuleOptions(ctx, VALKEYMODULE_OPTIONS_SKIP_COMMAND_VALIDATION);
     ValkeyModule_Replicate(ctx,"INCR","c","counter-2");
+    ValkeyModule_RemoveModuleOptions(ctx, VALKEYMODULE_OPTIONS_SKIP_COMMAND_VALIDATION);
     ValkeyModule_ReplyWithSimpleString(ctx,"OK");
     return VALKEYMODULE_OK;
 }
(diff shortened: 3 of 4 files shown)


src/valkeymodule.h
```diff
@@ -782,6 +782,11 @@ typedef enum {
     VALKEYMODULE_ACL_LOG_CHANNEL   /* Channel authorization failure */
 } ValkeyModuleACLLogEntryReason;
 
+typedef enum {
+    VALKEYMODULE_FLAG_DEFAULT = 0,     /* Default behavior */
+    VALKEYMODULE_FLAG_SKIP_VALIDATION, /* Skip validation */
+} ValkeyModuleFlag;
```
sungming2: Can I get some suggestions if this naming is too general

sungming2:
Open a pr for documentation of this: https://github.com/valkey-io/valkey-doc/pull/192

src/valkeymodule.h
hwware: Is there benefit set the flag as enum, more flags later? Can we just use int skip (0 for non-skip, 1 for skip) ?

src/valkeymodule.h
```diff
@@ -1092,6 +1097,8 @@ VALKEYMODULE_API int (*ValkeyModule_StringToStreamID)(const ValkeyModuleString *
 VALKEYMODULE_API void (*ValkeyModule_AutoMemory)(ValkeyModuleCtx *ctx) VALKEYMODULE_ATTR;
 VALKEYMODULE_API int (*ValkeyModule_Replicate)(ValkeyModuleCtx *ctx, const char *cmdname, const char *fmt, ...)
     VALKEYMODULE_ATTR;
+VALKEYMODULE_API int (*ValkeyModule_ReplicateWithFlag)(ValkeyModuleCtx *ctx, ValkeyModuleFlag flag, const char *cmdname, const char *fmt, ...)
```
hwware: If ValkeyModuleFlag can be set as int, here we can pass int value;

src/module.c
```diff
@@ -13523,6 +13512,34 @@ void moduleDefragGlobals(void) {
     dictReleaseIterator(di);
 }
 
+/* Helper function for VM_Replicate and VM_ReplicateWithFlag to replicate the specified command
+ * and arguments to replicas and AOF, as effect of execution of the calling command implementation.
+ * Skip command validation if the ValkeyModuleFlag is set to VALKEYMODULE_FLAG_SKIP_VALIDATION. */
+int moduleReplicate(ValkeyModuleCtx *ctx, ValkeyModuleFlag flag, const char *cmdname, const char *fmt, va_list ap) {
```
hwware: Pls move this function to just above the VM_Replicate, reference the function zsetInitLexRange. It is a helper function too.

hwware:
The PR comes from a performance issue about current VM_Replicated, then could you pls provide the result how much performance improvement on this PR?

sungming2:
I don't see notable performance improvement in the multiple tests (1. benchmark 2. CPU utilization, 3. CPU profiling with perf) so I wonder if it is worth introducing this new API. @hwware @hpatro 

## 1. Benchmark
### With validation
```
====== eval return redis.call("propagate-test.validation") 0 ======                                                     
  100000 requests completed in 0.71 seconds
  50 parallel clients
  74 bytes payload
  keep alive: 1
  multi-thread: no

0.000% <= 0.111 milliseconds (cumulative count 1)
50.000% <= 0.295 milliseconds (cumulative count 54440)
75.000% <= 0.335 milliseconds (cumulative count 75261)
87.500% <= 0.383 milliseconds (cumulative count 88850)
93.750% <= 0.407 milliseconds (cumulative count 94702)
96.875% <= 0.423 milliseconds (cumulative count 96981)
98.438% <= 0.479 milliseconds (cumulative count 98634)
99.219% <= 0.503 milliseconds (cumulative count 99282)
99.609% <= 0.599 milliseconds (cumulative count 99611)
99.805% <= 0.791 milliseconds (cumulative count 99811)
99.902% <= 0.999 milliseconds (cumulative count 99903)
99.951% <= 1.367 milliseconds (cumulative count 99952)
99.976% <= 1.615 milliseconds (cumulative count 99976)
99.988% <= 1.679 milliseconds (cumulative count 99989)
99.994% <= 1.687 milliseconds (cumulative count 99995)
99.997% <= 1.799 milliseconds (cumulative count 99997)
99.998% <= 1.831 milliseconds (cumulative count 99999)
99.999% <= 1.895 milliseconds (cumulative count 100000)
100.000% <= 1.895 milliseconds (cumulative count 100000)

3.697% <= 0.207 milliseconds (cumulative count 2697)
60.307% <= 0.303 milliseconds (cumulative count 60307)
94.702% <= 0.407 milliseconds (cumulative count 94702)
99.282% <= 0.503 milliseconds (cumulative count 99282)
99.620% <= 0.607 milliseconds (cumulative count 99620)
99.725% <= 0.703 milliseconds (cumulative count 99725)
99.826% <= 0.807 milliseconds (cumulative count 99826)
99.878% <= 0.903 milliseconds (cumulative count 99878)
99.904% <= 1.007 milliseconds (cumulative count 99904)
99.919% <= 1.103 milliseconds (cumulative count 99919)
99.928% <= 1.207 milliseconds (cumulative count 99928)
99.939% <= 1.303 milliseconds (cumulative count 99939)
99.962% <= 1.407 milliseconds (cumulative count 99962)
99.968% <= 1.503 milliseconds (cumulative count 99968)
99.975% <= 1.607 milliseconds (cumulative count 99975)
99.995% <= 1.703 milliseconds (cumulative count 99995)
99.998% <= 1.807 milliseconds (cumulative count 99998)
100.000% <= 1.903 milliseconds (cumulative count 100000)

Summary:
  throughput summary: 141043.72 requests per second
  latency summary (msec):
        0.305     0.104     0.295     0.415     0.495     1.895
```

### Bypass validation
```
====== eval return redis.call("propagate-test.novalidation") 0 ======                                                     
  100000 requests completed in 0.69 seconds
  50 parallel clients
  76 bytes payload
  keep alive: 1
  multi-thread: no

0.000% <= 0.119 milliseconds (cumulative count 1)
50.000% <= 0.279 milliseconds (cumulative count 52272)
75.000% <= 0.327 milliseconds (cumulative count 76134)
87.500% <= 0.367 milliseconds (cumulative count 87689)
93.750% <= 0.399 milliseconds (cumulative count 95285)
96.875% <= 0.423 milliseconds (cumulative count 97098)
98.438% <= 0.471 milliseconds (cumulative count 98588)
99.219% <= 0.543 milliseconds (cumulative count 99244)
99.609% <= 0.759 milliseconds (cumulative count 99619)
99.805% <= 0.927 milliseconds (cumulative count 99810)
99.902% <= 1.087 milliseconds (cumulative count 99903)
99.951% <= 1.271 milliseconds (cumulative count 99952)
99.976% <= 1.527 milliseconds (cumulative count 99977)
99.988% <= 1.575 milliseconds (cumulative count 99989)
99.994% <= 1.623 milliseconds (cumulative count 99994)
99.997% <= 1.663 milliseconds (cumulative count 99997)
99.998% <= 1.687 milliseconds (cumulative count 99999)
99.999% <= 1.719 milliseconds (cumulative count 100000)
100.000% <= 1.719 milliseconds (cumulative count 100000)

4.562% <= 0.207 milliseconds (cumulative count 4562)
67.223% <= 0.303 milliseconds (cumulative count 67223)
96.219% <= 0.407 milliseconds (cumulative count 96219)
99.021% <= 0.503 milliseconds (cumulative count 99021)
99.438% <= 0.607 milliseconds (cumulative count 99438)
99.562% <= 0.703 milliseconds (cumulative count 99562)
99.678% <= 0.807 milliseconds (cumulative count 99678)
99.787% <= 0.903 milliseconds (cumulative count 99787)
99.858% <= 1.007 milliseconds (cumulative count 99858)
99.910% <= 1.103 milliseconds (cumulative count 99910)
99.944% <= 1.207 milliseconds (cumulative count 99944)
99.954% <= 1.303 milliseconds (cumulative count 99954)
99.956% <= 1.407 milliseconds (cumulative count 99956)
99.973% <= 1.503 milliseconds (cumulative count 99973)
99.993% <= 1.607 milliseconds (cumulative count 99993)
99.999% <= 1.703 milliseconds (cumulative count 99999)
100.000% <= 1.807 milliseconds (cumulative count 100000)

Summary:
  throughput summary: 144300.14 requests per second
  latency summary (msec):
        0.294     0.112     0.279     0.399     0.503     1.719
```


## 2. CPU utilization 
* while continuously calling VM_Replicate/ReplicateWithFlag with benchmark tool (benchmark -n -l)
### With validation 
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                    
  33349 sungming  20   0  144672  25056   5888 R  54.0   0.2  12:51.11 valkey-server    
```
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                  
   1598 sungming  20   0  144672  25084   5700 R  93.3   0.2   0:30.32 valkey-server  
```

### Bypass validation
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                    
  33349 sungming  20   0  144672  25064   5888 S  52.8   0.2  13:29.67 valkey-server      
```
```
    PID USER      PR  NI    VIRT    RES    SHR S  %CPU  %MEM     TIME+ COMMAND                                                  
   1598 sungming  20   0  144672  25116   5700 R  92.7   0.2   1:06.09 valkey-server
```

## 3. Performance profiling with `perf` 

### with validation
```
-    8.10%     0.10%  valkey-server    valkey-server       [.] VM_Replicate                                                    ▒
   - 8.00% VM_Replicate                                                                                                        ▒
      - 7.65% moduleReplicate                                                                                                  ▒
         - 3.08% lookupCommandBySdsLogic                                                                                       ▒
            - 1.21% dictFetchValue                                                                                             ▒
                 0.88% dictFind                                                                                                ▒
            - 1.10% sdssplitlen.constprop.1                                                                                    ▒
                 0.56% _sdsnewlen                                                                                              ▒
         - 1.91% moduleCreateArgvFromUserFormat                                                                                ▒
              1.12% createEmbeddedStringObject                                                                                 ▒
           0.85% _sdsnewlen                                                                                                    ▒
           0.60% alsoPropagate                    
```

### bypass validation
```
-    3.84%     0.14%  valkey-server    valkey-server       [.] VM_ReplicateWithFlag                                            ▒
   - 3.70% VM_ReplicateWithFlag                                                                                                ▒
      - 3.47% moduleReplicate                                                                                                  ▒
         - 2.04% moduleCreateArgvFromUserFormat                                                                                ▒
              1.14% createEmbeddedStringObject                                                                                 ▒
              0.64% valkey_realloc                                                                                             ▒
           0.59% alsoPropagate                                                    
```

hwware:
@valkey-io/core-team Do you think this is a proper solution?

sungming2:
I added another CPU utilization test to push around 90%

src/valkeymodule.h
zuiderkwast: It's always complicated to have types like this in the module API. I believe we support backward and forward compatiblity for modules. I mean, a module built with the valkeymodule.h from Valkey 10 can run on Valkey 9 and vice versa. A module can check if some function exists in the current server by checking if the function is NULL, like `if (ValkeyModule_ReplicateWithFlag != NULL) ...`, but to check which flags are valid in an enum, we don't have a way for modules to check that.

I think Wen's suggestion with an `int` is simpler, but an even simpler API is to just add `ValkeyModule_ReplicateWithoutValidataion`. WDYT?

Internally, we can use an enum or int, but that's doesn't have to be part of the API.

sungming2: I think `ValkeyModule_ReplicateWithoutValidataion` is also good idea

sungming2:
New performance test showing performance increase.
https://github.com/valkey-io/valkey/issues/1175#issuecomment-2588551962

enjoy-binbin:
The new module api, if the command does not exist, the server will crash, right?

Are we considering adding flags to the client to bypass some checks? For example, a trusted client flag?

zuiderkwast:
> The new module api, if the command does not exist, the server will crash, right?

@enjoy-binbin I may be wrong, but I believe the primary can replica anything. The replica receiving it will try to execute it and will fail with an error which is normally just ignored. But there is a config that makes the replica crash:

```
# The propagation error behavior controls how the server will behave when it is
# unable to handle a command being processed in the replication stream from a primary
# or processed while reading from an AOF file. Errors that occur during propagation
# are unexpected, and can cause data inconsistency. 
#
# If an application wants to ensure there is no data divergence, this configuration
# should be set to 'panic' instead. The value can also be set to 'panic-on-replicas'
# to only panic when a replica encounters an error on the replication stream. One of
# these two panic values will become the default value in the future once there are
# sufficient safety mechanisms in place to prevent false positive crashes.
#
# propagation-error-behavior ignore
```

 > Are we considering adding flags to the client to bypass some checks? For example, a trusted client flag?

What do you mean? The module can send stuff to the replication stream even no actually command was sent by any real client. A module is always trusted. A module can do all kinds of dangerous things that can crash the server.

enjoy-binbin:
right, sorry about this confusing wording, i got the context mixed up (reading other discussions at the same time). anyway, what i means is that, can we add flags to the client, or like ctx flags then can bypass some internal checks? i am not sure about it though, i am not that good at module, i am ok with the new module api if needed.

zuiderkwast:
@valkey-io/core-team can we settle on the `ReplicateWithoutValidation` API? No flags added now. TSC please vote 👍 or 👎 .

hpatro:
> @valkey-io/core-team can we settle on the `ReplicateWithoutValidation` API? No flags added now. TSC please vote 👍 or 👎 .

@valkey-io/core-team nudge!

hwware:
> > @valkey-io/core-team can we settle on the `ReplicateWithoutValidation` API? No flags added now. TSC please vote 👍 or 👎 .
> 
> @valkey-io/core-team nudge!

Add it to next meeting agenda

zuiderkwast (dismissed):
Implementation LGTM.

src/module.c
```diff
@@ -3625,34 +3653,23 @@ int VM_ReplyWithLongDouble(ValkeyModuleCtx *ctx, long double ld) {
...
-    alsoPropagate(ctx->client->db->id, argv, argc, target);
+    return result;
+}
 
-    /* Release the argv. */
-    for (j = 0; j < argc; j++) decrRefCount(argv[j]);
-    zfree(argv);
-    server.dirty++;
-    return VALKEYMODULE_OK;
+/* Same as ValkeyModule_Replicate, but performing without command validation.
```
```suggestion
/* Same as ValkeyModule_Replicate, but without performing command validation.
```

src/module.c
```diff
@@ -3583,6 +3583,34 @@ int VM_ReplyWithLongDouble(ValkeyModuleCtx *ctx, long double ld) {
     return VALKEYMODULE_OK;
 }
 
+/* Helper function for VM_Replicate and VM_ReplicateWithoutValidation to replicate the specified command
+ * and arguments to replicas and AOF, as effect of execution of the calling command implementation.
+ * Skip command validation if skip is set to true(non-zero). */
+int moduleReplicate(ValkeyModuleCtx *ctx, int skip, const char *cmdname, const char *fmt, va_list ap) {
```
```suggestion
static int moduleReplicate(ValkeyModuleCtx *ctx, bool skip, const char *cmdname, const char *fmt, va_list ap) {
```

madolson (approved):
Looks good to me as well, just minor suggestions.

src/module.c
```diff
@@ -3583,6 +3583,34 @@ int VM_ReplyWithLongDouble(ValkeyModuleCtx *ctx, long double ld) {
     return VALKEYMODULE_OK;
 }
 
+/* Helper function for VM_Replicate and VM_ReplicateWithoutValidation to replicate the specified command
+ * and arguments to replicas and AOF, as effect of execution of the calling command implementation.
+ * Skip command validation if skip is set to true(non-zero). */
+static int moduleReplicate(ValkeyModuleCtx *ctx, bool skip, const char *cmdname, const char *fmt, va_list ap) {
```
enjoy-binbin: i see here we are using bool but we are passing 0/1

sungming2: thanks, addressed

enjoy-binbin (approved):

tests/modules/propagate.c
```diff
@@ -250,8 +250,9 @@ int propagateTestSimpleCommand(ValkeyModuleCtx *ctx, ValkeyModuleString **argv,
 
     /* Replicate two commands to test MULTI/EXEC wrapping. */
     ValkeyModule_Replicate(ctx,"INCR","c","counter-1");
-    ValkeyModule_Replicate(ctx,"INCR","c","counter-2");
-    ValkeyModule_ReplyWithSimpleString(ctx,"OK");
+    ValkeyModule_ReplicateWithoutValidation(ctx, "INCR",
+                                            "c", "counter-2");
```
hpatro: nit: Can't this in a single line?

tests/modules/propagate.c
```diff
@@ -266,8 +267,8 @@ int propagateTestMixedCommand(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, i
     ValkeyModule_FreeCallReply(reply);
 
     ValkeyModule_Replicate(ctx,"INCR","c","counter-1");
-    ValkeyModule_Replicate(ctx,"INCR","c","counter-2");
-
+    ValkeyModule_ReplicateWithoutValidation(ctx, "INCR",
+                                            "c", "counter-2");
```
hpatro: Same as above.

tests/unit/moduleapi/propagate.tcl
```diff
@@ -681,7 +681,6 @@ tags "modules" {
     }
 }
 
-
```
hpatro: Remove.

hpatro (approved):

src/module.c
```diff
@@ -3625,34 +3653,23 @@ int VM_ReplyWithLongDouble(ValkeyModuleCtx *ctx, long double ld) {
  * The command returns VALKEYMODULE_ERR if the format specifiers are invalid
  * or the command name does not belong to a known command. */
 int VM_Replicate(ValkeyModuleCtx *ctx, const char *cmdname, const char *fmt, ...) {
```
```suggestion
int VM_Replicate(ValkeyModuleCtx *ctx, const char *cmdname, const char *fmt, ...) {
```

After a fairly lengthy conversation, the core team (and jacob) came up with an alternative proposal. We weren't a fan of the config that effects all modules, but there is an API `voidValkeyModule_SetModuleOptions(ValkeyModuleCtx *ctx, int options);` which allows use to set a specific module config. We can add a new option for `SKIP_COMMAND_VALIDATION`, so specific modules can opt in to it and skip validation. This means that module developers can always use the same Replicate API.

If a module would like to have a per-site decision about validation, they can toggle the SKIP_COMMAND_VALIDATION on/off. 

Sorry for the late input.

sungming2: I will make a change based on above proposal

sungming2: Do we want to allow these APIs to use the skip validation too? because I can see these APIs use `lookupCommandByCString` only for validation purpose
[VM_CreateCommand](https://github.com/valkey-io/valkey/blob/unstable/src/module.c#L1348)
[VM_EmitAOF](https://github.com/valkey-io/valkey/blob/unstable/src/module.c#L7531)

madolson: For EmitAOF I think we should skip it.
For CreateCommand, I don't think we can skip it, since I believe it's used to prevent duplicates.

src/module.c
```diff
@@ -2566,11 +2566,27 @@ void VM_Yield(ValkeyModuleCtx *ctx, int flags, const char *busy_reply) {
...
  * a key-space notification callback. This flag allows to change this behavior
  * and fire nested key-space notifications. Notice: if enabled, the module
- * should protected itself from infinite recursion. */
+ * should protected itself from infinite recursion.
+ *
+ * VALKEYMODULE_OPTIONS_SKIP_COMMAND_VALIDATION:
+ * When set, this option allows the module to skip command validation.
+ * This is useful in scenarios where the module needs to bypass
+ * command validation (lookupCommandByCString) for specific operations
+ * to reduce overhead or handle trusted custom command logic. */
```
zuiderkwast: Let's mention which API function are affected by this: `ValkeyModule_Replicate` and `VM_EmitAOF`. Otherwise, it seems that it could affect others like `VM_Call` and who knows what. Let's also mention this option at `ValkeyModule_Replicate` and `VM_EmitAOF`.

`lookupCommandByCString` is internal implementation details that module authors shouldn't need to be aware of. I don't think we should mention it here.

src/module.c
```diff
@@ -2566,11 +2566,27 @@ void VM_Yield(ValkeyModuleCtx *ctx, int flags, const char *busy_reply) {
...
+/* Add specified options to the module's current options. */
+void VM_AddModuleOptions(ValkeyModuleCtx *ctx, int options) {
+    ctx->module->options |= options;
+}
+
+/* Remove specified options from the module's current options. */
+void VM_RemoveModuleOptions(ValkeyModuleCtx *ctx, int options) {
+    ctx->module->options &= ~options;
+}
+
```
zuiderkwast: One of the main arguments against `VM_ReplicateWithoutValidation` was that we wanted to avoid adding API functions.

Before we add these, we should probably consider the alternative of exposing a `VM_GetModuleOptions` instead. A get-and-set API is a slighly more simple or intuitive for the API, but slighly less practical for modules. Modules would do need to do like this:

```c
    int options = ValkeyModule_GetModuleOptions(ctx);
    options |= VALKEYMODULE_OPTIONS_SKIP_COMMAND_VALIDATION;
    ValkeyModule_SetModuleOptions(ctx, options);
```

That's my suggestion. Another alternative is to not add any of these at all. Then the module itself has to keep track of which options it's using.

zuiderkwast (commented):
When we discussed `ValkeyModule_SetModuleOptions`, it seems we forgot that there is no getter for these options.

src/module.c
madolson: Yeah, and if it's not clear, this documentation get's mirrored on valkey.io, so it should be developer facing.

src/module.c
madolson: I would vote to not add these. My assumption is that the module knows their global options and can set them. We aren't sure there is a use case for toggling this, and we can add something like GET later if there is a need.

sungming2: @zuiderkwast 
> we should probably consider the alternative of exposing a VM_GetModuleOptions

Yes, that was one of the options I considered and explored for different use cases. The only concern I had was that the client would need to call the API twice to set/remove the new module option, as you mentioned. If we want to minimize the number of exposed APIs, I agree that having GetModuleOptions would be a better approach and get-set pattern feels more intuitive.

sungming2: @madolson 
If we don't want to consider toggling the option right now, I will remove them

zuiderkwast: Yes I think it's better to remove them now. We can add the GetModuleOptions later, in a separate PR. It's a separate decision to make.

tests/modules/propagate.c
```diff
@@ -347,6 +347,9 @@ int ValkeyModule_OnLoad(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int arg
 
     detached_ctx = ValkeyModule_GetDetachedThreadSafeContext(ctx);
 
+    /* This option tests skip command validation for ValkeyModule_Replicate */
+    ValkeyModule_SetModuleOptions(ctx, VALKEYMODULE_OPTIONS_SKIP_COMMAND_VALIDATION);
```
sungming2, line 351: I'm unsure whether we need an entirely new test suite for this option or if it's fine to test it this way since we don't consider toggling

zuiderkwast: This seems fine to me.

What would you want to test? Scenarios with replication of non-existing commands?

src/module.c
```diff
@@ -7535,15 +7546,17 @@ void VM_EmitAOF(ValkeyModuleIO *io, const char *cmdname, const char *fmt, ...) {
...
-                  "emit unknown command '%s'",
-                  io->type->name, cmdname);
-        io->error = 1;
-        errno = EINVAL;
-        return;
+    if (!io->ctx || !io->ctx->module || !(io->ctx->module->options & VALKEYMODULE_OPTIONS_SKIP_COMMAND_VALIDATION)) {
+        cmd = lookupCommandByCString((char *)cmdname);
+        if (!cmd) {
+            serverLog(LL_WARNING,
+                      "Fatal: AOF method for module data type '%s' tried to "
```
zuiderkwast, line 7553: Interesting that we write "Fatal" although this is not a fatal error. Or is it?

zuiderkwast (approved):

tests/modules/propagate.c
sungming2: No I just wonder if it is ok to enable this option for this test module. thanks for confirming

src/module.c
sungming2: I don't think it's fatal error too as it does not return any error or abort something. We maybe want to remove the wording "Fatal" from this log?

madolson: I mean, honestly it probably should be a fatal. Like, this is breaking the durability system since we failed to write a record to the AOF file.

madolson (approved):
Nice! @valkey-io/core-team please double check the new updates. The surface area is much smaller, just a new option `VALKEYMODULE_OPTIONS_SKIP_COMMAND_VALIDATION`.

hpatro:
I guess it doesn't need to be a major decision anymore.

zuiderkwast:
@madolson We made the decision about this new option in the last core team meeting, right? I think we can consider it approved.

tests/modules/propagate.c
zuiderkwast: Yes, it seems OK to me.

PingXie (approved):

madolson:
> @madolson We made the decision about this new option in the last core team meeting, right? I think we can consider it approved.

We've had apparent consensus in the past but we didn't make it explicit. In the worst case I would like the folks to have the notification and they can choose to ignore it.

PR #1360, opened 2024-11-26
Optimize IO thread offload for modified argv

uriyage:
#### Background
In our IO threads architecture, IO threads allocate client argv's and later when we free it after processCommand we offload its free to the IO threads.
With jemalloc, it's crucial that the same thread that allocates memory also frees it. 

For some commands we modify the client's argv in the main thread during command processing (for example in `SET EX` command we rewrite the command to use absolute time for replication propagation).

#### Current issues
1. When commands are rewritten (e.g., expire commands), we store the original argv 
   in `c->original_argv`. However, we're currently:
   - Freeing new argv (allocated by main thread) in IO threads
   - Freeing original argv (allocated by IO threads) in main thread
2. Currently, `c->original_argv` points to new array with old 
   objects, while `c->argv` has old array with new objects, making memory free management complicated.

#### Changes
1. Refactored argv modification handling code to ensure consistency - both array and objects are now either all new or all old
2. Moved original_argv cleanup to happen in resetClient after argv cleanup
3. Modified IO threads code to properly handle original argv cleanup when argv are modified.

#### Performance Impact
Benchmark with `SET EX` commands (650 clients, 512 byte value, 8 IO threads):
- New implementation: **729,548 ops/sec**
- Old implementation: **633,243 ops/sec**
Representing a **~15%** performance improvement due to more efficient memory handling.


src/networking.c
```diff
@@ -4184,24 +4188,61 @@ void securityWarningCommand(client *c) {
...
+ *
+ * It performs the following operations:
+ * - Stores the initial command vector if not already saved
+ * - Manages memory allocation for command argument modifications
+ *
+ * new_argc - The new number of arguments to allocate space for if necessary.
+ * new_argv - Optional pointer to a new argument vector. If NULL, space will be
+ *                allocated for new_argc arguments, preserving the existing arguments.
+ */
+static void backupAndUpdateClientArgv(client *c, int new_argc, robj **new_argv) {
```
zuiderkwast, line 4267: Just noting that this function became pretty complex now. It would be useful to have unit test for it, but I will not require it. I hope we have some coverage for the edge cases, like a test that does replication of RESTORE with TTL and without ABSTTL. (It does redactClientCommandArgument twice and adds ABSTTL to the end.)

src/io_threads.c
```diff
@@ -439,13 +439,29 @@ void IOThreadFreeArgv(void *data) {
...
 
 /* This function attempts to offload the client's argv to an IO thread.
+ *
+ * If free_original_argv is set to 1, the function will offload the free of the client's original argv.
+ * Otherwise, it will offload the free of the client's current argv.
+ *
  * Returns C_OK if the client's argv were successfully offloaded to an IO thread,
  * C_ERR otherwise. */
-int tryOffloadFreeArgvToIOThreads(client *c) {
+int tryOffloadFreeArgvToIOThreads(client *c, int free_original_argv) {
```
zuiderkwast: Here, it is clear what the arguments mean, but at the call site, it is not immediately clear to the reader what the 2nd argument is, when you see the call like `tryOffloadFreeArgvToIOThreads(c, 0)`.

Trying to think of ideas:

* We could use some `enum {FREE_ARGV, FREE_ORIGINAL_ARGV}` instead of 0 and 1.
* We could pass pointers to the argv and argc fields within the client struct instead of the current parameters, so `tryOffloadFreeArgvToIOThreads(robj ***argv_ptr, int *argc_ptr)` and then at the call site it would look like `tryOffloadFreeArgvToIOThreads(&c->original_argv, &c->original_argc)`. Is this good?

zuiderkwast (commented):
Very nice! Just some points I'm not sure if we can improve, but I'm not sure we have to change anything. Maybe not. Let's think about and discuss first.

src/networking.c
uriyage: Thanks, Added unittest for it.

src/io_threads.c
uriyage: Thanks, good suggestion, changed to `tryOffloadFreeArgvToIOThreads(client *c, int argc, robj **argv)` no need for `***argv_ptr`

src/unit/test_networking.c
```diff
@@ -0,0 +1,131 @@
...
+    }
+    zfree(c->original_argv);
+
+    for (int i = 0; i < c->argc; i++) {
+        if (c->argv[i]) decrRefCount(c->argv[i]);
+    }
+    zfree(c->argv);
+    zfree(c);
+
+    return 0;
```
zuiderkwast, line 72: We could check that `zmalloc_used_memory() == 0` in the end to check we didn't leak anything.

src/networking.c
```diff
@@ -1481,14 +1481,19 @@ void freeClientOriginalArgv(client *c) {
...
     if (!c->original_argv) return;
 
-    for (int j = 0; j < c->original_argc; j++) decrRefCount(c->original_argv[j]);
-    zfree(c->original_argv);
+    if (tryOffloadFreeArgvToIOThreads(c, c->original_argc, c->original_argv) == C_ERR) {
+        for (int j = 0; j < c->original_argc; j++) decrRefCount(c->original_argv[j]);
+        zfree(c->original_argv);
+    }
+
     c->original_argv = NULL;
```
zuiderkwast, line 1496: If we free it in the main thread we set it to NULL, but not if we free it in the IO thread. Shouldn't we set it to NULL in both cases?

src/io_threads.c
zuiderkwast: Good idea. That is even better.

src/networking.c
uriyage: Not sure I understand, `c->original_argv = NULL` will always be executed.

zuiderkwast: Oh! You're right. I didn't see properly. was reviewing on the phone.

zuiderkwast (approved):
Great. Unit tests look good.

src/unit/test_networking.c
ranshid: I agree, although I do not know the Unity framework and if it runs several tests in parallel on different threads?
Anyhow we have the unitests run with sanitize on test-sanitizer-address for daylies 2 right?

zuiderkwast: The tests run one by one. Other tests already check used memory.

Yes, the sanitizer will catch a leak so I'm not worried.

src/server.c
```diff
@@ -3670,10 +3670,6 @@ void call(client *c, int flags) {
         replicationFeedMonitors(c, server.monitors, c->db->id, argv, argc);
     }
 
-    /* Clear the original argv.
-     * If the client is blocked we will handle slowlog when it is unblocked. */
-    if (!c->flag.blocked) freeClientOriginalArgv(c);
-
```
ranshid, line 3684: I am not sure we have an issue here, but you assume this will be cleaned when the client is reset, but what about all the multiple call calls in the multi/exec path?

zuiderkwast: Good catch!

ranshid (approved):
MOSTLY LGTM by now. @zuiderkwast any objections we merghe this?

zuiderkwast:
> MOSTLY LGTM by now. @zuiderkwast any objections we merghe this?

No objections. I've approved it myself. There's just a merge conflict that needs to be resolved. @uriyage can you fix it?

PR #1371, opened 2024-11-28
Add cross-version test framework (and a simple test)

### LINKED ISSUE #76: [NEW] Introduce automated cross version and cross fork testing infrastructure
## DESCRIPTION
Introduce cross version/cross fork integration testing infrastructure. With the compatibility version release and planned new major version release, it will be good to improve the testing/release certification process. This will help Valkey to prepare for release(s) more confidently and avoid pain for user(s) during migration/upgrade(s). 

### Example Scenario:

Issue: https://github.com/redis/redis/issues/12685

Redis 7.2 introduced cluster bus message extensions feature by default and it caused failure of engine upgrade from older version (i.e. Redis 6.2 or lower) due to message broadcasted from engine running 7.2 not being compatible in older versions.

PR to fix the issue during upgrade: https://github.com/valkey-io/valkey/pull/52

zuiderkwast:
This includes a way to run two versions of the server from the TCL test framework. It's a preparation to add more cross-version tests. The runtest script accepts a new parameter

    ./runtest --other-server-path path/to/valkey-server

and a new tag "needs:other-server" for test cases and start_server. Tests with this tag are automatically skipped if `--other-server-path` is not provided.

This PR adds it in a CI job with Valkey 7.2.7 by downloading a binary release.

Fixes #76


tests/test_helper.tcl
```diff
@@ -600,6 +601,9 @@ proc print_help_screen {} {
         "--tls-module       Run tests in TLS mode with Valkey module."
         "--host <addr>      Run tests against an external host."
         "--port <port>      TCP port to use against external host."
+        "--old-server-path <path>"
```
xbasel: IMHO using `--old-server-path` might limit the scope to only older versions, while `--other-server-path` or `--compat-server-path` feels less restrictive and better reflects its purpose for inter-version compatibility testing.

.github/workflows/ci.yml
```diff
@@ -18,10 +18,15 @@ jobs:
         # Fail build if there are warnings
         # build with TLS just for compilation coverage
         run: make -j4 all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes USE_FAST_FLOAT=yes
+      - name: install old server for compatibility testing
+        run: |
+          cd tests/tmp
+          wget https://download.valkey.io/releases/valkey-7.2.7-focal-x86_64.tar.gz
+          tar -xvf valkey-7.2.7-focal-x86_64.tar.gz
```
hpatro: I would prefer this to be in a script which can be invoked by multiple tests and provide engine version as parameter and skip downloading if it already exists.

tests/support/server.tcl
```diff
@@ -436,13 +441,17 @@ proc start_server {options {code undefined}} {
...
 
     # Wait for the server to be ready and check for server liveness/client connectivity before starting the test.
     set wait_ready true
 
     # parse options
     foreach {option value} $options {
         switch $option {
+            "start-old-server" {
+                set start_old_server $value
+            }
```
hpatro: Don't clearly get how does this option pick which old server to start.

tests/test_helper.tcl
zuiderkwast: I thought about words like "other", "alternative" server, etc. This is how I thought when I picked "old":

* If we're testing the compatibility between an older and a newer versions, let's say Valkey 8.1 and Valkey 9.0, either upgrade, downgrade or cluster bus compatibility, those are tests that we would add in the later version, Valkey 9 in this case. We wouldn't normally add tests in an old version to test against a future version.
* A name like "other server" is less clear when reading a test case about compatibility between and old and a new version. If you see "old server", it's very clear which one is the new version and which is the old version. A name like other would be more ambiguous.

Not sure the above are important though and I understand your point that we can also test against something like KeyDB, AWS MemoryDB, Rockskv, etc. Also the same Valkey version compiled in different ways, with/without some modules, without Lua, etc, etc. Yes, there are definitely more use cases than old and new, so I agree something like "other" is more generic.

Other just means *not the system under test*, which is what matters here. "Compat" doesn't seem general enough though. We may (hypothetically) want to test something against some *incompatible* version to test the failure modes (e.g. that we log some error and don't crash).

I'll change to "other" then, unless you have more ideas or pros/cons.

.github/workflows/ci.yml
zuiderkwast: I was thinking keep it minimal and that we can add some abstraction for it when we need it.

What kind of script do you have in mind? A python script under utils/, a yaml file under .github/ or a TCL script under tests/? If we put it under utils/, does that mean we officially support the script?

If we add a script with these few lines, is the version parameter a string like "7.2.7" and we assume the rest of the URLs are always identical? What if we're testing something from another URL, or with a different executable name (an old redis-server, kvdb-server, etc.) Then the parameters are bigger than the script itself.

So far, we don't have any scripts that downloads anything from hard-coded URLs, apart from in the CI jobs, so it think this might be the best place for it. Can we add some github action definition that can be used from multiple github actions?

tests/support/server.tcl
zuiderkwast: It's just a boolean (1 or 0), to start the old server or to start the normal server.

The server to start is a global variable `$::old_server_path`, set by the `test_helper.tcl` script when parsing the `--old-server-path` option.

.github/workflows/ci.yml
hpatro: Sticking to Github actions options seems problematic i.e. how do other vendors/users test backward compatibility locally? 

My thought is to have a python script with well defined old server name+version (enum) and download path (doesn't need to be identical). And yes, we would need to support it officially.

zuiderkwast: > how do other vendors/users test backward compatibility locally?

Like this:

    ./runtest --old-server-path ../my-other-valkey-clone/src/valkey-server

You may want to build it from source, either a source release or a git checkout, even to test a development branch against unstable.

Using a binary release from our website was just a simple way to get the CI running. Not sure it's the best way or a method I want to officially support.

> My thought is to have a python script with well defined old server name+version (enum) and download path (doesn't need to be identical). And yes, we would need to support it officially.

It seems like a commitment to support a bunch of URLs to binary releases. I don't really see the benefit. Is it to have predictable tests? Each compatiblity test uses a fixed version, so it knows what to expect?

Keep in mind that it's also a different binary for a different architecture and per OS. To run it on ARM, you can't use the same binary. There are even differences between ubuntu versions, like OpenSSL version and such things.

.github/workflows/ci.yml
```diff
@@ -18,10 +18,15 @@ jobs:
         # Fail build if there are warnings
         # build with TLS just for compilation coverage
         run: make -j4 all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes USE_FAST_FLOAT=yes
+      - name: install old server for compatibility testing
```
madolson, line 21: Since I guess we might theoretically do other types of compatibility, maybe something like:
```suggestion
      - name: install secondary server for compatibility testing
```

tests/test_helper.tcl
madolson: I proposed secondary in my other comment, I like other too. I don't really like `compat`.

.github/workflows/ci.yml
madolson: For daily are we going to run it against all supported versions? That can also be a followup.

madolson (commented):
Simple and efficient.

.github/workflows/ci.yml
zuiderkwast: Yes, sounds good, and we should add some more meaningful test cases too, like the one we wanted for the light-weight cluster message and the cluster ping-extension issues we've had in the past. I just wanted to create the scaffolding and a simple test case first.

tests/test_helper.tcl
zuiderkwast: I've heard secondary in the context of primary-secondary, meaning primary-replica.

I prefer "other" so far. The essence is that it's a different one than the SUT one. Other = different.

We could use "different", but `--different-server-path` doesn't look that great.

zuiderkwast: Changed to "other"

tests/support/server.tcl
zuiderkwast: Added a comment. (And changed to "other".)

.github/workflows/ci.yml
zuiderkwast: I changed to "other" in the test framework. In this particular case, the CI job is testing against 7.2.7 which is actually an old version, so I kept "old" here.

We can change it to "other" or "secondary" in the future, if we change the CI job to use some matrix of versions or something like that. Sounds good?

zuiderkwast:
> For daily are we going to run it against all supported versions? That can also be a followup.

Yes, a follow up. More importantly, we should add some cross-version tests for the cluster bus with ping extensions, lightweigt pubsub and such things, especially when we're developing new such features to make sure we don't break the rolling upgrade scenarios.

I'd like to see rolling upgrade tests of a whole cluster too.

zuiderkwast:
The `wait_for_sync` time out in the CI job, so apparently it's flaky. Why isn't 5 seconds enough? This function is used in many other test cases.

```
$ cd tests ; git grep -A5 'proc wait_for_sync'
support/util.tcl:proc wait_for_sync r {
support/util.tcl-    wait_for_condition 50 100 {
support/util.tcl-        [status $r master_link_status] eq "up"
support/util.tcl-    } else {
support/util.tcl-        fail "replica didn't sync in time"
support/util.tcl-    }
support/util.tcl-}
```

tests/README.md
```diff
@@ -24,16 +24,17 @@ There are additional runtime options that can further adjust the test suite to
...
+| Option                    | Impact                                                   |
+| ------------------------- | -------------------------------------------------------- |
+| `--singledb`              | Only use database 0, don't assume others are supported. |
+| `--ignore-encoding`       | Skip all checks for specific encoding.  |
+| `--ignore-digest`         | Skip key value digest validations. |
+| `--cluster-mode`          | Run in strict Valkey Cluster compatibility mode. |
+| `--large-memory`          | Enables tests that consume more than 100MB |
+| `--tls`                   | Run tests with TLS. See below. |
+| `--tls-module`            | Run tests with TLS, when TLS support is built as a module. |
+| `--other-server-path PATH | Run compatibility tests with an other server executable. |
```
madolson:
```suggestion
| `--other-server-path PATH` | Run compatibility tests with an other server executable. |
```

tests/README.md
```diff
@@ -88,16 +89,20 @@ The following compatibility and capability tags are currently used:
...
-| `needs:repl`              | Uses replication and needs to be able to `SYNC` from server. |
+| `ipv6`                    | Uses IPv6. |
+| `needs:repl`, `repl`      | Uses replication and needs to be able to `SYNC` from server. |
 | `needs:debug`             | Uses the `DEBUG` command or other debugging focused commands (like `OBJECT REFCOUNT`). |
 | `needs:pfdebug`           | Uses the `PFDEBUG` command. |
 | `needs:config-maxmemory`  | Uses `CONFIG SET` to manipulate memory limit, eviction policies, etc. |
 | `needs:config-resetstat`  | Uses `CONFIG RESETSTAT` to reset statistics. |
 | `needs:reset`             | Uses `RESET` to reset client connections. |
 | `needs:save`              | Uses `SAVE` or `BGSAVE` to create an RDB file. |
+| `needs:other-server         | Requires `--other-server-path`. |
```
madolson:
```suggestion
| `needs:other-server`         | Requires `--other-server-path`. |
```

madolson: It also slightly erks me this is misaligned, but not something I think needs to be fixed.

madolson:
> The wait_for_sync time out in the CI job, so apparently it's flaky. Why isn't 5 seconds enough? This function is used in many other test cases.

I've never seen this fail anywhere else before. I wonder if there is something about how it's being used in this instance that might take extra time.

tests/README.md
zuiderkwast: I'll fix it. I missed this when i did s/old/other/.

.github/workflows/ci.yml
madolson: Let's do this incrementally maybe? I think it's better to have this code in and running and we can adjust it as needed, for now just a wget seems OK.

madolson (approved):

.github/workflows/ci.yml
madolson: No strong opinion from me, the current wording makes sense.

zuiderkwast:
Thx @madolson. I'm just worried about introducing a flaky test here. I'll merge it if everything passes after updating the PR.

hpatro:
@zuiderkwast let us know if you need any help with wrapping up this PR. @sarthakaggarwal97 is happy to help.

zuiderkwast:
> @zuiderkwast let us know if you need any help with wrapping up this PR. @sarthakaggarwal97 is happy to help.

I answered this in my previous comment. Please help ensure we don't have a flaky test. And please clean up the flaky tests in unstable. Let's have at least two weeks without Daily failing.

zuiderkwast:
After merging unstable, the `test-ubuntu-latest` runs on ubuntu-24 "noble". Our "focal" binaries require libssl 1.1.1 so they won't start. I updated the binary name to "noble" assuming we have a binary like that, but to my surprise we have only

* focal (2020)
* bionic (2018)

And they're built with libssl 1.1.1 so they refuse to start on any recent OS....

@hpatro Any suggestions what to do now?

zuiderkwast:
It seems the binary for 7.2.8 for "noble" is just missing. Using 7.2.7 for noble works. However, it might break automatically next time ubuntu-latest is bumbed, in two years or so.

I'm going to merge this now.

Follow-up: If we use binaries built for ubuntu-24, we should probably lock the CI job to run on ubuntu-24 rather than ubuntu-latest. Another option is to run a container instead of a binary download. Help wanted with this follow-up.

hpatro:
> It seems the binary for 7.2.8 for "noble" is just missing. Using 7.2.7 for noble works. However, it might break automatically next time ubuntu-latest is bumbed, in two years or so.

It's odd that we have noble for 7.2.7, unsure if we built it manually and uploaded it to the S3 bucket. I don't see we have it listed on the website as well https://valkey.io/download/releases/v7-2-7/ .

zuiderkwast:
There's an issue for it in the website repo. Many things need help in website and docs repos.

PR #1376, opened 2024-11-28
Use zfree_with_size for client buffer

stav-bentov:
Replace occurrences of 'zfree' with 'zfree_with_size' to improve performance. 
'zfree_with_size' function avoids calling 'zmalloc_size' to retrieve buffer size and 
uses previuos calculation of size for calling  'zfree_with_size'. This results in faster 
memory deallocation and reduces overhead.

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -1753,7 +1753,7 @@ void freeClient(client *c) {
     /* Free data structures. */
     listRelease(c->reply);
     c->reply = NULL;
-    zfree(c->buf);
+    zfree_with_size(c->buf, c->buf_usable_size);
     c->buf = NULL;
     freeReplicaReferencedReplBuffer(c);
     freeClientArgv(c);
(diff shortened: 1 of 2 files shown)


zuiderkwast (approved):
Looks good. Can you see any actual performance improvement in some profiling?

@uriyage @ranshid @poiuj FYI

stav-bentov:
> Looks good. Can you see any actual performance improvement in some profiling?
> 
> @uriyage @ranshid @poiuj FYI

Thank you for the feedback.
The performance improvement is not significant, but I believe it's still worth merging.
From my testing, I observed an average improvement of about **0.44%** in requests per second when running 
`benchmark -r 3000000 -d 500 -t multi -k 0 -n 10000000` with get/set (80% get and 20% set requests) **after** filling the database with 3M keys.

zuiderkwast:
Thanks! Yes, of course it's worth merging. I was just curious. It's always interesting to be able to measure an optimization.

PR #1379, opened 2024-12-01
Fix test_reclaimFilePageCache to avoid tmpfs

### LINKED ISSUE #897: [Test failure] test_reclaimFilePageCache fails on Amazon Linux 2023 AARCH64
`test_reclaimFilePageCache` consistently fails on Amazon Linux 2023:
```
[test_reclaimFilePageCache - unit/test_util.c:302] Failed assertion: !cache_exist(fd)
[fail] - test_util.c:test_reclaimFilePageCache
```
The test successfully creates the tmp file. But looks like `fadvise` doesn't behave as expected. Another option is that `mincore` doesn't behave inside `cache_exists` test utility.

Kernel info:
```
$ uname -srv
Linux 6.1.87-99.174.amzn2023.aarch64 #1 SMP Tue Apr 23 00:06:47 UTC 2024
```

xbasel:
Avoid tmpfs as fadvise(FADV_DONTNEED) has no effect on memory-backed filesystems.

Fixes https://github.com/valkey-io/valkey/issues/897

### DIFF
--- a/src/unit/test_util.c
+++ b/src/unit/test_util.c
@@ -296,11 +296,11 @@ int test_reclaimFilePageCache(int argc, char **argv, int flags) {
     if (flags & UNIT_TEST_VALGRIND) return 0;
 
 #if defined(__linux__)
-    struct statfs buf;
+    struct statfs stats;
 
     /* Check if /tmp is memory-backed (e.g., tmpfs) */
-    if (statfs("/tmp", &buf) == 0) {
-        if (buf.f_type != TMPFS_MAGIC) { // Not tmpfs, use /tmp
+    if (statfs("/tmp", &stats) == 0) {
+        if (stats.f_type != TMPFS_MAGIC) { // Not tmpfs, use /tmp
             return 0;
         }
     }
--- a/src/unit/test_util.c
+++ b/src/unit/test_util.c
@@ -314,7 +314,6 @@ int test_reclaimFilePageCache(int argc, char **argv, int flags) {
     }
     
     char *tmpfile = "/tmp/redis-reclaim-cache-test";
-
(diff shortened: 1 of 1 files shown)


src/unit/test_util.c
```diff
@@ -283,6 +285,24 @@ static int cache_exist(int fd) {
...
+ */
+static const char *get_non_tmpfs_temp_file_path(void) {
+    struct statfs buf;
+
+    /* Check if /tmp is memory-backed (e.g., tmpfs) */
+    if (statfs("/tmp", &buf) == 0) {
+        if (buf.f_type != TMPFS_MAGIC) { // Not tmpfs, use /tmp
+            return "/tmp/redis-reclaim-cache-test";
+        }
+    }
```
ranshid: Can we just make this check part of the test_reclaimFilePageCache and return 0 in case it is not file backed? I think forcing this test to retun 0? Honestly I would also be fine with eliminating this test.

xbasel: done (use tmp and only if its non-tmpfs)

src/unit/test_util.c
```diff
@@ -291,7 +309,12 @@ int test_reclaimFilePageCache(int argc, char **argv, int flags) {
     if (flags & UNIT_TEST_VALGRIND) return 0;
 
 #if defined(__linux__)
+    if (!is_tmp_disk_based()) {
+        return 0; // Can not be tested on a ram drive.
+    }
+    
     char *tmpfile = "/tmp/redis-reclaim-cache-test";
+
```
ranshid, line 307:
```suggestion
```

ranshid (approved):

PR #1384, opened 2024-12-03
Add CLUSTER FLUSHSLOT command

### LINKED ISSUE #1133: [Feature] Implement CLUSTER FLUSHSLOT command
As part of the per-slot dictionary effort, we implemented all of the needed changes so that we could efficiently unlink and free the data in a single slot with lazyfree. This is useful in two cases:
1. When a primary loses ownership of a slot, it will temporarily freeze while it synchronously frees the slot. 
2. When doing resharding, you may prefer to forcibly delete all the data from a slot and assign it to another node. 

The main point where code needs to be updated: https://github.com/valkey-io/valkey/blob/b5eb7930797023f46331d68b964b3871625da79d/src/cluster_legacy.c#L6040

Alternative naming conventions (Naming was finalized, leaving for historical purposes):
1. `SFLUSH` (presumably slot flush). The S prefix is also used for set, so it may not make the most sense to emulate.
2. `FLUSHSLOT`. Since this is mostly be an admin command, it might make sense to attach the cluster.

wuranxx:
Add cluster flushslot command. Closes #1133.

Syntax:

    CLUSTER FLUSHSLOT <slot> [ASYNC|SYNC]

## Background

We want to implement the ability to SYNC or ASYNC free all data from a single slot. This is useful in two cases:

1. When a primary loses ownership of a slot, it will temporarily freeze while it synchronously frees the slot.
2. When doing resharding, you may prefer to forcibly delete all the data from a slot and assign it to another node.

## Internal changes

Modify the function signature of `delKeysInSlot`, adding parameters `lazy`, `propagate_del` and `send_del_event` to control the behavior.


src/cluster_legacy.c
```diff
@@ -6842,6 +6842,23 @@ int clusterCommandSpecial(client *c) {
     } else if (!strcasecmp(c->argv[1]->ptr, "links") && c->argc == 2) {
         /* CLUSTER LINKS */
         addReplyClusterLinksDescription(c);
+    } else if (!strcasecmp(c->argv[1]->ptr, "flushslot") && (c->argc == 3 || c->argc == 4)) {
+        /* CLUSTER FLUSHSLOT <slot> [ASYNC|SYNC] */
+        int slot;
+        int lazy = 1;
```
zuiderkwast: Let's use `server.lazyfree_lazy_user_flush` by default? That's the default for FLUSHALL when the caller doesn't specify sync or async.

wuranxx: fixed.

wuranxx:
Please help review this PR. Thank you. @hwware @madolson @PingXie @zuiderkwast

hwware:
There are too much information in issue https://github.com/valkey-io/valkey/issues/1133. To save some of the reviewer energy, could you pls update the top comment to describe the background why we want to this new command? You could reference PR https://github.com/valkey-io/valkey/pull/1392.   

Thanks a lot

src/cluster_legacy.c
```diff
@@ -7041,3 +7049,21 @@ int clusterDecodeOpenSlotsAuxField(int rdbflags, sds s) {
...
+    if (c->argc == 4) {
+        if (!strcasecmp(c->argv[3]->ptr, "async")) {
+            lazy = 1;
+        } else if (!strcasecmp(c->argv[3]->ptr, "sync")) {
+            lazy = 0;
+        } else {
+            addReplyErrorObject(c, shared.syntaxerr);
+            return;
+        }
+    }
```
PingXie: we should fail the `c->argc > 4` case I think

tests/unit/cluster/cluster-flush-slot.tcl
```diff
@@ -0,0 +1,34 @@
...
+        # set key
+        for {set i 0} {$i < 1000} {incr i} {
+            R 0 set "{FC}-$i" "value"
+        }
+        set after_keys_num [expr {$slot_keys_num + 1000}]
+        assert_equal [R 0 CLUSTER COUNTKEYSINSLOT $key_slot] $after_keys_num
+
+        # flush slot key
+        R 0 CLUSTER FLUSHSLOT $key_slot ASYNC
+        assert_equal [R 0 CLUSTER COUNTKEYSINSLOT $key_slot] 0
```
PingXie, line 31: we should add a wait (`wait_for_condition`). This could be a source of test flakiness.

PingXie (commented):
LGTM overall. Thanks @wuranxx!

src/cluster_legacy.c
wuranxx: In the call site of `clusterCommandFlushslot`, I have already performed a check on the number of parameters. Is it still necessary to validate the number of parameters again within the function?

```c
 else if (!strcasecmp(c->argv[1]->ptr, "flushslot") && (c->argc == 3 || c->argc == 4)) {
        /* CLUSTER FLUSHSLOT <slot> [ASYNC|SYNC] */
        clusterCommandFlushslot(c);
    }
```

wuranxx:
Before I rebased to the latest  unstable (old commit is 32f7541fe34e5e0520a5917d09661756d330bd11) , the Tcl tests and manual command tests ran fine locally. 

After I rebased to the latest unstable 6df376d68a97e9c0da4549f57db96742b5482202, the command failed when executed locally. 

It is possible that changes in the code during this period caused the inconsistent behavior. Until I confirm the cause, I have marked this PR as WIP (Work In Progress).

tests/unit/cluster/cluster-flush-slot.tcl
wuranxx: this command execute failed when I rebased to the latest unstable https://github.com/valkey-io/valkey/commit/6df376d68a97e9c0da4549f57db96742b5482202, I try to confirm the cause.

wuranxx:
> Before I rebased to the latest unstable (old commit is [32f7541](https://github.com/valkey-io/valkey/commit/32f7541fe34e5e0520a5917d09661756d330bd11)) , the Tcl tests and manual command tests ran fine locally.
> 
> After I rebased to the latest unstable [6df376d](https://github.com/valkey-io/valkey/commit/6df376d68a97e9c0da4549f57db96742b5482202), the command failed when executed locally.
> 
> It is possible that changes in the code during this period caused the inconsistent behavior. Until I confirm the cause, I have marked this PR as WIP (Work In Progress).

Fixed. For details, see #1399.

zuiderkwast:
@enjoy-binbin @murphyjacob4 Do you want this command for atomic slot migration, for example when a slot migration is cancelled, the importing node can send CLUSTER FLUSHSLOT to its replicas to clean up the slot?

murphyjacob4:
I think we will have a protocol for cancelling the slot migration, and this will trigger the drop for all slot ranges being migrated locally. But we should be able to use the async `delKeysInSlot` functionality to implement this

zuiderkwast:
> I think we will have a protocol for cancelling the slot migration, and this will trigger the drop for all slot ranges being migrated locally. But we should be able to use the async `delKeysInSlot` functionality to implement this

@murphyjacob4 The replicas are not involved in the slot migration protocol IIUC. So when a primary cleans up a cancelled slot import, do you just replicate a lot of DEL commands to the replicas? Replicating a single FLUSHSLOT may save some work and bandwidth...

murphyjacob4:
> do you just replicate a lot of DEL commands to the replicas? Replicating a single FLUSHSLOT may save some work and bandwidth...

Yeah that's a good idea. Currently, this is what `delKeysInSlot` does, and it could result in a lot of bandwidth for big slots. Disabling the replication of the individual keys in `delKeysInSlot` and replacing with a single `FLUSHSLOT` makes sense to me. This path is also executed in the dirty slot case, so it won't just be an improvement for atomic slot migration (although I anticipate it will be executed more frequently there).

src/cluster_legacy.c
zuiderkwast: It's a bit weird that the arity check is split into multiple places. But I don't mind too much because it's like that for many commands already.

But I think this should be in cluster.c instead of cluster_legacy.c. Cluster_legacy is about the cluster bus, and we would replace this file if we replace the cluster bus for cluster V2. At least, that's the idea. This command doesn't use the cluster bus at all, so I think it should be in cluster.c and be called directly from `clusterCommand()`.

src/cluster_legacy.c
```diff
@@ -6353,8 +6355,13 @@ unsigned int delKeysInSlot(unsigned int hashslot) {
...
+            dbAsyncDelete(&server.db[0], key);
+        } else {
+            dbSyncDelete(&server.db[0], key);
+        }
         dbDelete(&server.db[0], key);
-        propagateDeletion(&server.db[0], key, server.lazyfree_lazy_server_del);
+        propagateDeletion(&server.db[0], key, lazy);
         signalModifiedKey(NULL, &server.db[0], key);
         /* The keys are not actually logically deleted from the database, just moved to another node.
          * The modules needs to know that these keys are no longer available locally, so just send the
```
zuiderkwast: This code comment about keyspace notifications:

```
        /* The keys are not actually logically deleted from the database, just moved to another node.
         * The modules needs to know that these keys are no longer available locally, so just send the
         * keyspace notification to the modules, but not to clients. */
```

`delKeysInSlot` is used for slot migrations, which are considered not deleting the keys but just migrating them to another node.

This is not necessarily the case for CLUSTER FLUSHSLOT. I think we should treat CLUSTER FLUSHSLOT more like FLUSHDB. For FLUSHDB, what are the keyspace notifications sent? A lot of DEL events or one FLUSHDB event?

Maybe we shouldn't use `delKeysInSlot` here, or use some extra argument or flag to indicate if it's a migration or if we're just deleting the keys.

src/cluster_legacy.c
```diff
@@ -6353,8 +6355,13 @@ unsigned int delKeysInSlot(unsigned int hashslot) {
         enterExecutionUnit(1, 0);
         sds sdskey = objectGetKey(valkey);
         robj *key = createStringObject(sdskey, sdslen(sdskey));
+        if (lazy) {
+            dbAsyncDelete(&server.db[0], key);
+        } else {
+            dbSyncDelete(&server.db[0], key);
```
zuiderkwast: Here, the deleted keys are replicated just by multiple DEL commands. For comparison, FLUSHDB is replicated as a single FLUSHDB command, which is much less traffic than many DEL commands. If we want to replicate FLUSHSLOT as a single FLUSHSLOT command, then we should probably check that the replicas supports this though. We already have the replica's version stored in `c->repl_data->replica_version` (REPLCONF VERSION is sent by the replica since 8.0) so we could check that it's >= 8.1 for all connected replicas. It's also fine to skip this. We can do it later as a separate optimization.

zuiderkwast (commented):
Hi @wuranxx and sorry for the delay.

It looks strait-forward, but there are some details we should think about a bit more. I'm not actually sure what's best to do in all cases, but see my comments below.

src/cluster_legacy.c
wuranxx: 1.  For FLUSHDB, what are the keyspace notifications sent? A lot of DEL events or one FLUSHDB event?
    - In my test, flushdb not trigger any event. 

```
192.168.0.1:2774> set k2 1
OK
192.168.0.1:2774> flushdb
OK
```

```
192.168.0.1:2774> PSUBSCRIBE __key*@0__:*
Reading messages... (press Ctrl-C to quit)
1) "psubscribe"
2) "__key*@0__:*"
3) (integer) 1
1) "pmessage"
2) "__key*@0__:*"
3) "__keyspace@0__:k2"
4) "set"
1) "pmessage"
2) "__key*@0__:*"
3) "__keyevent@0__:set"
4) "k2"
```

2.  Maybe we shouldn't use delKeysInSlot here, or use some extra argument or flag to indicate if it's a migration or if we're just deleting the keys.
    - yes, we should add some flag to mark is migration or del.


If we want to like flushdb, we should write a new method to implement flushslot logic.

zuiderkwast: > In my test, flushdb not trigger any event.

That's surprising, but OK. I guess flushslot should behave the same as flushdb. I'm open to other suggestions though. If this was a bad design for flushdb, then we can do something else for flushslot.

> If we want to like flushdb, we should write a new method to implement flushslot logic.

Let's to it?

src/cluster_legacy.c
zuiderkwast: Actually I think we don't have to worry about old replicas. This is a new feature and users should use new features only after all nodes have been upgraded.

src/cluster_legacy.c
wuranxx: Okay, I'll try to rewrite the flushslot as per the flushdb implementation.

wuranxx:
I add command scene to delKeysInSlot. In command sence, delKeysInSlot will not propagate del command, but will notify del event.

flushslot now just propagate `flushslot` command to aof and replica.

please help review this PR, thank you! @zuiderkwast @PingXie

src/cluster_legacy.c
```diff
@@ -6449,13 +6450,24 @@ unsigned int delKeysInSlot(unsigned int hashslot) {
...
+        }
+        // if is command, skip del propagate
+        if (!is_cmd) propagateDeletion(&server.db[0], key, lazy);
         signalModifiedKey(NULL, &server.db[0], key);
-        /* The keys are not actually logically deleted from the database, just moved to another node.
-         * The modules needs to know that these keys are no longer available locally, so just send the
-         * keyspace notification to the modules, but not to clients. */
-        moduleNotifyKeyspaceEvent(NOTIFY_GENERIC, "del", key, server.db[0].id);
+        if (is_cmd) {
+            /* In cluster flushslot scene, the keys are actually deleted. */
```
```suggestion
            /* In the `cluster flushslot` scenario, the keys are actually deleted so notify everyone. */
```

src/cluster_legacy.c
```diff
@@ -6449,13 +6450,24 @@ unsigned int delKeysInSlot(unsigned int hashslot) {
...
+            notifyKeyspaceEvent(NOTIFY_GENERIC, "del", key, server.db[0].id);
+        } else {
+            /* The keys are not actually logically deleted from the database, just moved to another node.
+            * The modules needs to know that these keys are no longer available locally, so just send the
+            * keyspace notification to the modules, but not to clients. */
+            /* In cluster flushslot scene, the is actually deleted, fire del event. */
```
```suggestion
```

src/commands/cluster-flushslot.json
```diff
@@ -0,0 +1,46 @@
+{
+    "FLUSHSLOT": {
+        "summary": "Remove all keys from the target slot.",
+        "complexity": "O(N) where N is the number of keys in the target slot",
+        "group": "cluster",
+        "since": "8.1.0",
```
PingXie:
```suggestion
        "since": "9.0.0",
```

src/commands/cluster-flushslot.json
```diff
@@ -0,0 +1,46 @@
...
+        "arity": -3,
+        "container": "CLUSTER",
+        "function": "clusterCommand",
+        "command_flags": [
+            "WRITE"
+        ],
+        "acl_categories": [
+            "KEYSPACE",
+            "DANGEROUS"
+        ],
```
PingXie: I think we should block this command on async loading. also, `ADMIN` is a more appropriate flag, which gets `DANGEROUS` automatically.
```suggestion
        "command_flags": [
            "WRITE",
            "NO_ASYNC_LOADING",
            "ADMIN"
        ],
        "acl_categories": [
            "KEYSPACE"
        ],
```

src/commands/cluster-flushslot.json
```diff
@@ -0,0 +1,46 @@
...
+            {
+                "name": "flush-type",
+                "type": "oneof",
+                "optional": true,
+                "arguments": [
+                    {
+                        "name": "async",
+                        "type": "pure-token",
+                        "token": "ASYNC",
+                        "since": "8.1.0"
```
PingXie:
```suggestion
                        "since": "9.0.0"
```

src/commands/cluster-flushslot.json
```diff
@@ -0,0 +1,46 @@
...
+                    },
+                    {
+                        "name": "sync",
+                        "type": "pure-token",
+                        "token": "SYNC",
+                        "since": "8.1.0"
```
PingXie:
```suggestion
                        "since": "9.0.0"
```

tests/unit/cluster/cluster-flush-slot.tcl
```diff
@@ -0,0 +1,34 @@
...
+    }
+}
```
PingXie, line 33: Can you add a test case to validate the command is indeed replicated as-is, as opposed to the key deletions getting replicated?

PingXie (commented):
Overall LGTM. We should also add official documentation for this new command.

src/cluster_legacy.c
murphyjacob4: For atomic slot migration - we will need send some FLUSHSLOT commands implicitly:

1. If an import doesn't succeed (is failed or cancelled) we need to flush the contents of the staged slots to replicas
2. If there is an import to a target primary and FLUSHDB is run on that target, the target will need to break this out into FLUSHSLOT commands to avoid the importing slots being flushed
3. If there is an export happening from some source primary and FLUSHDB is run, we need to break this out into FLUSHSLOT commands (so that the target only flushes the impacted importing slots)

Plus - in the case that a dirty slot is found - we currently delete it by sending individual key deletions. It would make sense to create the FLUSHSLOT command implicitly to improve replication traffic.

Due to the cases of implicit FLUSHSLOT propagation - I would suggest that the "delKeysInSlot" helper method include logic for deciding about if we should propagate as individual deletions or as FLUSHSLOT based on the capabilities of the replicas. I'm sure for many customers, they may be running in a cross-version deployment for some amount of time while they do rolling upgrades

tests/unit/cluster/cluster-flush-slot.tcl
```diff
@@ -32,3 +32,33 @@ start_cluster 2 2 {tags {external:skip cluster}} {
...
+            puts "R $l info:"
+            puts [R $l info replication]
+            puts [R $l info commandSTATS]
+        }
+
+        set info [R 2 info commandSTATS]
+        # not del cmd
+        assert { ![string match "*cmdstat_del*" $info] }
+        # has flushslot cmd
+        assert { [string match "*cmdstat_cluster|flushslot:calls=1*" $info] }
```
zuiderkwast: Use `assert_no_match` and `assert_match` instead.

wuranxx: fixed.

tests/unit/cluster/cluster-flush-slot.tcl
wuranxx: Of course. Added.

src/cluster.h
```diff
@@ -1,3 +1,5 @@
+#include <stdbool.h>
+
 #ifndef __CLUSTER_H
 #define __CLUSTER_H
```
zuiderkwast: `#ifndef __CLUSTER_H` should be the first line in the file, I think.

src/cluster_legacy.c
zuiderkwast: @murphyjacob4

> we will need send some FLUSHSLOT commands implicitly

Yes, so do you need any changes here? You can call this function and then you explicitly replicate a FLUSHSLOT command. Or do you want this function to do it?

I think you may want to skip keyspace events and skip replication using DEL commands. With the `is_cmd` argument you can't disable both. See other comment.

> include logic for deciding about if we should propagate as individual deletions or as FLUSHSLOT based on the capabilities of the replicas. I'm sure for many customers, they may be running in a cross-version deployment for some amount of time while they do rolling upgrades

I agree that we should support cross-version deployments, but in this case I'm not worried and I don't think we need to decide based on replica capabilities.

We require that all nodes are upgraded before you start using new features. For example, an old replica wouldn't understand a new datatype, or information like hash field expiration, etc. The FLUSHSLOT command and the atomic slot migration are both new features. If a node has ASM, it also has FLUSHSLOT. If you have an 8.1 replica in your cluster, you shouldn't start using these features just yet. Makes sense?

src/cluster_legacy.c
```diff
@@ -6535,13 +6535,14 @@ void removeChannelsInSlot(unsigned int slot) {
 
 /* Remove all the keys in the specified hash slot.
  * The number of removed items is returned. */
-unsigned int delKeysInSlot(unsigned int hashslot) {
+unsigned int delKeysInSlot(unsigned int hashslot, int lazy, bool is_cmd) {
```
zuiderkwast: I like the functionality here but I'm not sure about the name of the `is_cmd` argument. It controls two separate things:

is_cmd false: 

* Propagate the deletions using DEL commands
* Don't send keyspace events

is_cmd true:

* Don't propagate deletions to replicas
* Send keyspace events

To send an implicit FLUSHSLOT command from atomic slot migration when a slot migration is aborted, I guess we want to propagate it as a FLUSHSLOT and not send keyspace events. Replicating FLUSHSLOT can be done using a separate call.

Maybe we should have two separate options instead of `is_cmd`.

zuiderkwast (commented):
I think this is very close to being merged.

There is a merge conflict that was caused by #1671 that needs to be resolved.

zuiderkwast:
I noticed there is unfortunately already a command with a very similar name: [CLUSTER FLUSHSLOTS](https://valkey.io/commands/cluster-flushslots/):

> The CLUSTER FLUSHSLOTS deletes all information about slots from the connected node. It can only be called when the database is empty.

The new command in this PR `CLUSTER FLUSHSLOT` is maybe too similar to that command?

Maybe we should actually make it a top-level command `FLUSHSLOT` without CLUSTER? It will match `FLUSHDB` and `FLUSHALL`. We need to discuss it again in the core team meeting. No need to change the PR before we have discussed it again.

madolson:
Weekly meeting notes. We will keep the command as is, but we will document this as primarily an internal and debugging command.

PingXie:
>Weekly meeting notes. We will keep the command as is, but we will document this as primarily an internal and debugging command.

I think we discussed two decisions

1. whether this should be a top level command?

based on the use case described in the issue, this is clearly not a "data access" command meant for the application developers, hence the guidance of keeping it under `cluster`.

2. whether it is a good idea to introduce a new command whose name differs from that of an existing command by just one character? `cluster flushshot` vs `cluster flushshots`.

I don't remember if we reached a conclusion?

zuiderkwast:
> I don't remember if we reached a conclusion?

Thanks Ping for summarizing the discussion. We did reach the conclusion to stick to `cluster flushslot`, as Madelyn wrote.

Neither of these commands are frequently used so the similarity between the two commands (`cluster flushshot` vs `cluster flushshots`) is not a major concern.

src/cluster_legacy.c
wuranxx: We should provide two parameters, one is `propagate_del` control  the `del` command is propagated, and the other is `send_del_event` control whether  the `del` keyspace event is sent.

When `cluster flushslot`  is executed, it is confirmed that the `propagate_del` is false, and `send_del_event` should be true?

zuiderkwast: These two parameters sound good to me.

> When `cluster flushslot` is executed, it is confirmed that the `propagate_del` is false, and `send_del_event` should be true?

Yes, I confirm.

The command `cluster flushslot` itself is replicated (separate from this function) so we don't need to replicate individual DEL commands. And yes, send events because the keys are really deleted, as opposed to some scenarios in slot migration where the keys are considered to be moved to another node and not really deleted from the cluster. I read this in a code comment.

zuiderkwast: @murphyjacob4 can you confirm, for ASM when replicating an aborted slot migration, you need to set these parameters to (false, false) right?

hwware:
> > Weekly meeting notes. We will keep the command as is, but we will document this as primarily an internal and debugging command.
> 
> I think we discussed two decisions
> 
> 1. whether this should be a top level command?
> 
> based on the use case described in the issue, this is clearly not a "data access" command meant for the application developers, hence the guidance of keeping it under `cluster`.
> 
> 2. whether it is a good idea to introduce a new command whose name differs from that of an existing command by just one character? `cluster flushshot` vs `cluster flushshots`.
> 
> I don't remember if we reached a conclusion?

I forgot if you drop off meeting earlier, we have one conclusion that the new command "cluster flushslot" as an internal command.

hwware:
> > > Weekly meeting notes. We will keep the command as is, but we will document this as primarily an internal and debugging command.
> > 
> > 
> > I think we discussed two decisions
> > 
> > 1. whether this should be a top level command?
> > 
> > based on the use case described in the issue, this is clearly not a "data access" command meant for the application developers, hence the guidance of keeping it under `cluster`.
> > 
> > 2. whether it is a good idea to introduce a new command whose name differs from that of an existing command by just one character? `cluster flushshot` vs `cluster flushshots`.
> > 
> > I don't remember if we reached a conclusion?
> 
> I forgot if you drop off meeting earlier, we have one conclusion that the new command "cluster flushslot" as an internal command, like cluster-SYNCSLOTS (a new internal command too)
image

src/cluster_legacy.c
wuranxx: added.

src/commands/cluster-flushslot.json
```diff
@@ -0,0 +1,47 @@
...
+            {
+                "name": "flush-type",
+                "type": "oneof",
+                "optional": true,
+                "arguments": [
+                    {
+                        "name": "async",
+                        "type": "pure-token",
+                        "token": "ASYNC",
+                        "since": "9.0.0"
```
zuiderkwast: Nit: We don't need "since" for the options that are added command itself is new. The command itself has "since": "9.0.0".

zuiderkwast (approved):
LGTM.

@murphyjacob4 Flush one slot at a time, is that enough for cancelled atomic slot migration? If you have 1000 ongoing slot migrations that are cancelled, will you need to send 1000 flushslot commands to the replicas at the same time?

murphyjacob4:
> @murphyjacob4 Flush one slot at a time, is that enough for cancelled atomic slot migration? If you have 1000 ongoing slot migrations that are cancelled, will you need to send 1000 flushslot commands to the replicas at the same time?

I think it is fine. ASM actually has a few scenarios for FLUSHSLOT:

1. If a slot migration is cancelled, each slot within this migration will be removed by the target primary and FLUSHSLOT will be propagated to the target replica(s)
2. If FLUSHSLOT is run on the source primary, it will be sent to the target primary (via slot migration link) foreach ongoing slot export on that node (if any). Target primary will forward this to the target replicas.
3. If FLUSHSLOT is run on the target primary (by user), it will be denied if the slot is being imported (should be covered by this CL I think, will return -MOVED)
4. If FLUSHDB is run on the source primary, if there is a slot export occurring, it will replicate as a FLUSHSLOT command for each owned slot. For all target primaries and their replicas, the FLUSHSLOT commands that match the ongoing slot migration will be propagated via the path in 2
5. If FLUSHDB is run on the target primary, if there is a slot import occurring, it will replicate as a FLUSHSLOT command for each owned slot. Locally, we will leave the importing data. We will not replicate a FLUSHSLOT for the ongoing slot import (since we want to keep that data

In the worst case, it would be something like ~700KiB of replication data to send 16384 FLUSHSLOT commands (each would be <= 43 bytes, `*3\r\n$7\r\nCLUSTER\r\n$9\r\nFLUSHSLOT\r\n$5\r\n16383`). But this seems like a very rare case. We have a path to optimizing with something like `CLUSTER FLUSHSLOTSRANGE` a la `CLUSTER ADDSLOTSRANGE`, but I don't think it is required, and just FLUSHSLOT should better than propagating DEL for each key (at least for 99% of use cases)

zuiderkwast:
> * If FLUSHDB is run on the source primary, if there is a slot export occurring, it will replicate as a FLUSHSLOT command for each owned slot. For all target primaries and their replicas, the FLUSHSLOT commands that match the ongoing slot migration will be propagated via the path in 2
> * If FLUSHDB is run on the target primary, if there is a slot import occurring, it will replicate as a FLUSHSLOT command for each owned slot. Locally, we will leave the importing data. We will not replicate a FLUSHSLOT for the ongoing slot import (since we want to keep that data

@murphyjacob4 Now that we have multi-db in cluster mode, FLUSHSLOT will affects all databases. FLUSHDB should only flush the currently SELECT'ed database.

hwware:
> > @murphyjacob4 Flush one slot at a time, is that enough for cancelled atomic slot migration? If you have 1000 ongoing slot migrations that are cancelled, will you need to send 1000 flushslot commands to the replicas at the same time?
> 
> I think it is fine. ASM actually has a few scenarios for FLUSHSLOT:
> 
> 1. If a slot migration is cancelled, each slot within this migration will be removed by the target primary and FLUSHSLOT will be propagated to the target replica(s)
> 2. If FLUSHSLOT is run on the source primary, it will be sent to the target primary (via slot migration link) foreach ongoing slot export on that node (if any). Target primary will forward this to the target replicas.
> 3. If FLUSHSLOT is run on the target primary (by user), it will be denied if the slot is being imported (should be covered by this CL I think, will return -MOVED)
> 4. If FLUSHDB is run on the source primary, if there is a slot export occurring, it will replicate as a FLUSHSLOT command for each owned slot. For all target primaries and their replicas, the FLUSHSLOT commands that match the ongoing slot migration will be propagated via the path in 2
> 5. If FLUSHDB is run on the target primary, if there is a slot import occurring, it will replicate as a FLUSHSLOT command for each owned slot. Locally, we will leave the importing data. We will not replicate a FLUSHSLOT for the ongoing slot import (since we want to keep that data
> 
> In the worst case, it would be something like ~700KiB of replication data to send 16384 FLUSHSLOT commands (each would be <= 43 bytes, `*3\r\n$7\r\nCLUSTER\r\n$9\r\nFLUSHSLOT\r\n$5\r\n16383`). But this seems like a very rare case. We have a path to optimizing with something like `CLUSTER FLUSHSLOTSRANGE` a la `CLUSTER ADDSLOTSRANGE`, but I don't think it is required, and just FLUSHSLOT should better than propagating DEL for each key (at least for 99% of use cases)

We have consensus that flushslot is an internal command only, I think case 2 and 3 will not happen. How do you think?

zuiderkwast:
> We have consensus that flushslot is an internal command only, I think case 2 and 3 will not happen. How do you think?

Agree, but I'm not sure about case 4 and 5. Maybe we need `CLUSTER FLUSHSLOT 2333 DB 2` to handle FLUSHDB on one db during atomic slot migration. It can be extended later if we need it. Otherwise, we could change flushdb so it only affects owned slots?

src/commands/cluster-flushslot.json
wuranxx: fixed.

hwware (approved):

hwware:
> > We have consensus that flushslot is an internal command only, I think case 2 and 3 will not happen. How do you think?
> 
> Agree, but I'm not sure about case 4 and 5. Maybe we need `CLUSTER FLUSHSLOT 2333 DB 2` to handle FLUSHDB on one db during atomic slot migration. It can be extended later if we need it. Otherwise, we could change flushdb so it only affects owned slots?

I think case 4 is a valid scenario. However, for case 5, it depends on how the operator does. One worst case is causing data inconsistent. (source primary is exporting data, and FLUSHDB command runs target primary, then some data will lost)

soloestoy:
For the rollback of atomic slot migration, using `FLUSHSLOT` to affect all databases is more appropriate. However, considering the replication of `FLUSHDB` during migration and the possibility that users might need to clear a specific slot in a particular database, I think `CLUSTER FLUSHSLOT <slotid> [DB id]` would be a good option.

hwware:
> For the rollback of atomic slot migration, using `FLUSHSLOT` to affect all databases is more appropriate. However, considering the replication of `FLUSHDB` during migration and the possibility that users might need to clear a specific slot in a particular database, I think `CLUSTER FLUSHSLOT <slotid> [DB id]` would be a good option.

I think the proper command format as below:

CLUSTER FLUSHSLOT <slot id> [[DB id] | all] [ASYNC|SYNC]

if there is no any argument after <slot id>, this command will only apply current db, it is aligned with FLUSHDB
if the command as CLUSTER FLUSHSLOT <slot id> [DB id], it will only remove specific db keys
if command as CLUSTER FLUSHSLOT <slot id> all, it will remove keys from all dbs

I add this PR in our next Monday meeting agenda.

madolson:
@valkey-io/core-team Shouldn't this have had a major decision tag? I remember we decided this was an internal command, but it doesn't look like it was really restricted to just be internal.

zuiderkwast:
The issue has the major decision tag. Seemed enough...(?)

> it doesn't look like it was really restricted to just be internal.

If we don't add a markdown file for it in the docs, it will not show up in the docs, but maybe it will show up somewhere else? How can we make it completely internal?

ranshid:
@valkey-io/core-team did we miss creating a doc for this command in 9.0? it has the needs-doc-pr label, but was unable to locate any PR for it...

PR #1386, opened 2024-12-03
RDMA: Fix dead loop when transfer large data (20KB)

### LINKED ISSUE #1385: [BUG] When Server's io-threads > 1 and large data is transferred over RDMA, the server gets stuck in an endless loop
**Describe the bug**

When Server's io-threads > 1 and large data is transferred over RDMA, the server gets stuck in an endless loop

**To reproduce**

Run Server 
```
./src/valkey-server \
    --port 6379 \
    --protected-mode no \
    --appendonly no \
    --loglevel notice \
    --loadmodule src/valkey-rdma.so bind=<your host> port=6380 \
    --server_cpulist 0-31:2 \
    --bgsave_cpulist 96-127:2 \
    --io-threads 8
```
Run [redis-benchmark with rdma](https://github.com/pizhenwei/redis/tree/feature-rdma-with-cli) (provided by @pizhenwei)
You may need to modify the code before running it
```
diff --git a/src/redis-benchmark.c b/src/redis-benchmark.c
index a77565526..390d33b6b 100644
--- a/src/redis-benchmark.c
+++ b/src/redis-benchmark.c
@@ -636,6 +636,9 @@ static void writeHandler(aeEventLoop *el, int fd, void *privdata, int mask) {
                         fprintf(stderr, "Error writing to the server: %s\n", strerror(errno));
                     freeClient(c);
                     return;
+                } else if (nwritten > 0) {
+                    c->written += nwritten;
+                    return;
                 }
             } else {
                 aeDeleteFileEvent(el,c->context->fd,AE_WRITABLE);
```
and then run

```
./src/redis-benchmark \
    -h <your host> \
    -n 10000000 \
    -r 100 \
    -t set \
    -p 6380 \
    -d 10485760  \
    -c 50 \
    --threads 16 \
    --rdma
```

**Expected behavior**

Get a benchmark report as usual

**Actual behavior**

Main thread of Server run 100% cpu, and redis-benchmark waits

**Additional information**

I am a newcomer, please bear with me for my English level and description ability

fengquyoumo:
Determine the status of the Client when attempting to read data. If state=CLIENT_COMPLETED_IO, no read attempt is made and I/O operations on the Client are rescheduled by the main thread.

> And 20474 Byte = PROTO_IOBUF_LEN(16KB) + SDS_HDR_VAR(16, s)(4090 Byte)

Fixes #1385

### DIFF
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -1816,7 +1816,7 @@ static ConnectionType CT_RDMA = {
     .has_pending_data = rdmaHasPendingData,
     .process_pending_data = rdmaProcessPendingData,
     .postpone_update_state = postPoneUpdateRdmaState,
-    .update_state = updateRdmaState
+    .update_state = updateRdmaState,
 };
 
 ConnectionType *connectionTypeRdma(void) {
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -77,8 +77,11 @@ typedef enum ValkeyRdmaOpcode {
 #define VALKEY_RDMA_INVALID_OPCODE 0xffff
 #define VALKEY_RDMA_KEEPALIVE_MS 3000
 
+#define RDMA_CONN_FLAG_POSTPONE_UPDATE_STATE (1 << 0)
+
 typedef struct rdma_connection {
     connection c;
+    int flags;
     struct rdma_cm_id *cm_id;
     int last_errno;
     listNode *pending_list_node;
@@ -678,7 +681,6 @@ static connection *connCreateAcceptedRdma(int fd, void *priv) {
 static void connRdmaEventHandler(struct aeEventLoop *el, int fd, void *clientData, int mask) {
     rdma_connection *rdma_conn = (rdma_connection *)clientData;
     connection *conn = &rdma_conn->c;
-    client *c = connGetPrivateData(conn);
     struct rdma_cm_id *cm_id = rdma_conn->cm_id;
     RdmaContext *ctx = cm_id->context;
     int ret = 0;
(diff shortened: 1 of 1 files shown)


zuiderkwast:
@pizhenwei RDMA expert, please take a look.

@fengquyoumo Thanks for your contribution. You need to commit with `git commit -s` which adds the Signed-off-by header in the commit message. See CONTRIBUTING.md about why we need this. How to fix it, see the Details link at the "DCO" CI job.

pizhenwei:
Hi @fengquyoumo 
I've got a cold, and will be back soon. please be a little more patient.

fengquyoumo:
@zuiderkwast  :) Thanks for your help!

pizhenwei:
Hi @fengquyoumo 
Thanks for your report and contribution!

There is a lack of multi-threads support from RDMA. In general, you are right on this issue, but it's not valkey connection style. I suggest:
* implement `postpone_update_state` and `update_state` for RDMA
* `CLIENT_PENDING_IO` is only used by `networking.c` and `io_threads.c`, don't break this. Use `RDMA_CONN_FLAG_POSTPONE_UPDATE_STATE` in `rdma.c`
* handle pending data in `rdma.c` correctly

Please see `tls.c` as example:
```
    .postpone_update_state = postPoneUpdateSSLState,
    .update_state = updateSSLState,
```

fengquyoumo:
@pizhenwei Thanks for your help！ 
I implemented `postpone_update_state` and `update_state` for RDMA to solve this issue. 
Also, the first implementation had serious performance issues, but this one worked just fine.

Please review the code again when you are free, Thanks!

src/rdma.c
```diff
@@ -77,8 +77,11 @@ typedef enum ValkeyRdmaOpcode {
 #define VALKEY_RDMA_INVALID_OPCODE 0xffff
 #define VALKEY_RDMA_KEEPALIVE_MS 3000
 
+#define RDMA_CONN_FLAG_POSTPONE_UPDATE_STATE (1 << 0)
+
 typedef struct rdma_connection {
     connection c;
+    int flags;
```
pizhenwei: I guess
```
struct rdma_cm_id *cm_id;
int flags;
int last_errno;
```

would reduce 8 bytes.

Otherwise LGTM. I also test it locally, it works fine.

Hi @zuiderkwast , would you please take a look at this change?

fengquyoumo: :) Thanks a lot for the detailed feedback!

pizhenwei (approved):
LGTM.

zuiderkwast (approved):
Great. Thanks both of you for the fix and the review.

PR #1387, opened 2024-12-03
move clientCron onto a separate timer

JimB123:
The `serverCron()` function contains a variety of maintenance functions and is set up as a timer job, configured to run at a certain rate (hz).  The default rate is 10hz (every 100ms).

One of the things that `serverCron()` does is to perform maintenance functions on connected clients.  Since the number of clients is variable, and can be very large, this could cause latency spikes when the 100ms `serverCron()` task gets invoked.  To combat those latency spikes, a feature called "dynamic-hz" was introduced.  This feature will run `serverCron()` more often, if there are more clients.  The clients get processed up to 200 at a time.  The delay for `serverCron()` is shortened with the goal of processing all of the clients every second.

The result of this is that some of the other (non-client) maintenance functions also get (unnecessarily) run more often.  Like `cronUpdateMemoryStats()` and `databasesCron()`.  Logically, it doesn't make sense to run these functions more often, just because we happen to have more clients attached.

This PR separates client activities onto a separate, variable, timer.  The "dynamic-hz" feature is eliminated.  Now, `serverCron` will run at a standard configured rate.  The separate clients cron will automatically adjust based on the number of clients.  This has the added benefit that often, the 2 crons will fire during separate event loop invocations and will usually avoid the combined latency impact of doing both maintenance activities together.

The new timer follows the same rules which were established with the dynamic HZ feature.
* The goal is to process all of the clients once per second
* We never want to process more than 200 clients in a single invocation (`MAX_CLIENTS_PER_CLOCK_TICK`)
* We always process at least 5 clients at a time (`CLIENTS_CRON_MIN_ITERATIONS`)
* The minimum rate is determined by HZ

The delay (ms) for the new timer is also more precise, computing the number of milliseconds needed to achieve the goal of reaching all of the clients every second.  The old dynamic-hz feature just performs a doubling of the HZ until the clients processing rate is achieved (i.e. delays of 100ms, 50ms, 25ms, 12ms...)

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -913,7 +913,7 @@ int clientsCronResizeOutputBuffer(client *c, mstime_t now_ms) {
 size_t ClientsPeakMemInput[CLIENTS_PEAK_MEM_USAGE_SLOTS] = {0};
 size_t ClientsPeakMemOutput[CLIENTS_PEAK_MEM_USAGE_SLOTS] = {0};
 
-int clientsCronTrackExpansiveClients(client *c, int time_idx) {
+int clientsCronTrackExpensiveClients(client *c, int time_idx) {
     size_t qb_size = c->querybuf ? sdsAllocSize(c->querybuf) : 0;
     size_t argv_size = c->argv ? zmalloc_size(c->argv) : 0;
     size_t in_usage = qb_size + c->argv_len_sum + argv_size;
@@ -1035,8 +1035,8 @@ int updateClientMemUsageAndBucket(client *c) {
 }
 
 /* Return the max samples in the memory usage of clients tracked by
- * the function clientsCronTrackExpansiveClients(). */
-void getExpansiveClientsInfo(size_t *in_usage, size_t *out_usage) {
+ * the function clientsCronTrackExpensiveClients(). */
+void getExpensiveClientsInfo(size_t *in_usage, size_t *out_usage) {
     size_t i = 0, o = 0;
     for (int j = 0; j < CLIENTS_PEAK_MEM_USAGE_SLOTS; j++) {
         if (ClientsPeakMemInput[j] > i) i = ClientsPeakMemInput[j];
@@ -1046,37 +1046,22 @@ void getExpansiveClientsInfo(size_t *in_usage, size_t *out_usage) {
     *out_usage = o;
 }
 
-/* This function is called by serverCron() and is used in order to perform
+/* This function is called by clientsTimerProc() and is used in order to perform
  * operations on clients that are important to perform constantly. For instance
  * we use this function in order to disconnect clients after a timeout, including
  * clients blocked in some blocking command with a non-zero timeout.
  *
  * The function makes some effort to process all the clients every second, even
- * if this cannot be strictly guaranteed, since serverCron() may be called with
- * an actual frequency lower than server.hz in case of latency events like slow
+ * if this cannot be strictly guaranteed, since clientsTimerProc() may be called with
+ * an actual frequency lower than the intended rate in case of latency events like slow
  * commands.
  *
  * It is very important for this function, and the functions it calls, to be
- * very fast: sometimes the server has tens of hundreds of connected clients, and the
- * default server.hz value is 10, so sometimes here we need to process thousands
- * of clients per second, turning this function into a source of latency.
+ * very fast. Sometimes the server has tens of thousands of connected clients, and all
+ * of them need to be processed every second.
  */
-#define CLIENTS_CRON_MIN_ITERATIONS 5
-void clientsCron(void) {
-    /* Try to process at least numclients/server.hz of clients
-     * per call. Since normally (if there are no big latency events) this
-     * function is called server.hz times per second, in the average case we
-     * process all the clients in 1 second. */
-    int numclients = listLength(server.clients);
-    int iterations = numclients / server.hz;
+static void clientsCron(int clients_this_cycle) {
     mstime_t now = mstime();
-
-    /* Process at least a few clients while we are at it, even if we need
-     * to process less than CLIENTS_CRON_MIN_ITERATIONS to meet our contract
-     * of processing each client once per second. */
-    if (iterations < CLIENTS_CRON_MIN_ITERATIONS)
-        iterations = (numclients < CLIENTS_CRON_MIN_ITERATIONS) ? numclients : CLIENTS_CRON_MIN_ITERATIONS;
-
 
     int curr_peak_mem_usage_slot = server.unixtime % CLIENTS_PEAK_MEM_USAGE_SLOTS;
     /* Always zero the next sample, so that when we switch to that second, we'll
@@ -1088,14 +1073,13 @@ void clientsCron(void) {
      * some slow command is called taking multiple seconds to execute. In that
      * case our array may end containing data which is potentially older
      * than CLIENTS_PEAK_MEM_USAGE_SLOTS seconds: however this is not a problem
-     * since here we want just to track if "recently" there were very expansive
+     * since here we want just to track if "recently" there were very expensive
      * clients from the POV of memory usage. */
     int zeroidx = (curr_peak_mem_usage_slot + 1) % CLIENTS_PEAK_MEM_USAGE_SLOTS;
     ClientsPeakMemInput[zeroidx] = 0;
     ClientsPeakMemOutput[zeroidx] = 0;
 
-
-    while (listLength(server.clients) && iterations--) {
+    while (listLength(server.clients) && clients_this_cycle--) {
         client *c;
         listNode *head;
 
@@ -1105,14 +1089,14 @@ void clientsCron(void) {
         c = listNodeValue(head);
         listRotateHeadToTail(server.clients);
         if (c->io_read_state != CLIENT_IDLE || c->io_write_state != CLIENT_IDLE) continue;
+
         /* The following functions do different service checks on the client.
          * The protocol is that they return non-zero if the client was
          * terminated. */
         if (clientsCronHandleTimeout(c, now)) continue;
         if (clientsCronResizeQueryBuffer(c)) continue;
         if (clientsCronResizeOutputBuffer(c, now)) continue;
-
-        if (clientsCronTrackExpansiveClients(c, curr_peak_mem_usage_slot)) continue;
+        if (clientsCronTrackExpensiveClients(c, curr_peak_mem_usage_slot)) continue;
 
         /* Iterating all the clients in getMemoryOverheadData() is too slow and
          * in turn would make the INFO command too slow. So we perform this
@@ -1124,6 +1108,40 @@ void clientsCron(void) {
 
         if (closeClientOnOutputBufferLimitReached(c, 0)) continue;
     }
+}
+
+/* A periodic timer that performs client maintenance.
+ * This cron task follows the following rules:
+ *  - All clients need to be checked (at least) once per second
+ *  - To manage latency, we don't check more than MAX_CLIENTS_PER_CLOCK_TICK at a time
+ *  - The minimum rate will be defined by server.hz
+ *  - At least CLIENTS_CRON_MIN_ITERATIONS will be performed each cycle
+ */
+#define CLIENTS_CRON_MIN_ITERATIONS 5
+long long clientsTimerProc(struct aeEventLoop *eventLoop, long long id, void *clientData) {
+    UNUSED(eventLoop);
+    UNUSED(id);
+    UNUSED(clientData);
+
+    int numclients = listLength(server.clients);
+    int clients_this_cycle = numclients / server.hz; // Initial computation based on standard hz
+    int delayMs;
+
+    if (clients_this_cycle < CLIENTS_CRON_MIN_ITERATIONS) {
+        clients_this_cycle = min(numclients, CLIENTS_CRON_MIN_ITERATIONS);
+    }
+
+    if (clients_this_cycle > MAX_CLIENTS_PER_CLOCK_TICK) {
+        clients_this_cycle = MAX_CLIENTS_PER_CLOCK_TICK;
+        float required_hz = (float)numclients / MAX_CLIENTS_PER_CLOCK_TICK;
+        delayMs = 1000.0 / required_hz;
+    } else {
+        delayMs = 1000 / server.hz;
+    }
+
+    clientsCron(clients_this_cycle);
+
+    return delayMs;
 }
 
 /* This function handles 'background' operations we are required to do
@@ -1346,19 +1364,6 @@ long long serverCron(struct aeEventLoop *eventLoop, long long id, void *clientDa
      * handler if we don't return here fast enough. */
     if (server.watchdog_period) watchdogScheduleSignal(server.watchdog_period);
 
-    server.hz = server.config_hz;
-    /* Adapt the server.hz value to the number of configured clients. If we have
-     * many clients, we want to call serverCron() with an higher frequency. */
-    if (server.dynamic_hz) {
-        while (listLength(server.clients) / server.hz > MAX_CLIENTS_PER_CLOCK_TICK) {
-            server.hz *= 2;
-            if (server.hz > CONFIG_MAX_HZ) {
-                server.hz = CONFIG_MAX_HZ;
-                break;
-            }
-        }
-    }
-
     /* for debug purposes: skip actual cron work if pause_cron is on */
     if (server.pause_cron) return 1000 / server.hz;
 
@@ -1443,9 +1448,6 @@ long long serverCron(struct aeEventLoop *eventLoop, long long id, void *clientDa
                       zmalloc_used, hmem);
         }
     }
-
-    /* We need to do a few operations on clients asynchronously. */
-    clientsCron();
 
     /* Handle background operations on databases. */
     databasesCron();
@@ -2629,7 +2631,6 @@ void initServer(void) {
     /* Initialization after setting defaults from the config system. */
     server.aof_state = server.aof_enabled ? AOF_ON : AOF_OFF;
     server.fsynced_reploff = server.aof_enabled ? 0 : -1;
-    server.hz = server.config_hz;
     server.in_fork_child = CHILD_TYPE_NONE;
     server.rdb_pipe_read = -1;
     server.rdb_child_exit_pipe = -1;
@@ -2775,10 +2776,15 @@ void initServer(void) {
     server.acl_info.invalid_channel_accesses = 0;
 
     /* Create the timer callback, this is our way to process many background
-     * operations incrementally, like clients timeout, eviction of unaccessed
-     * expired keys and so forth. */
+     * operations incrementally, like eviction of unaccessed expired keys, etc. */
     if (aeCreateTimeEvent(server.el, 1, serverCron, NULL, NULL) == AE_ERR) {
-        serverPanic("Can't create event loop timers.");
+        serverPanic("Can't create serverCron timer.");
+        exit(1);
+    }
+    /* A separate timer for client maintenance.  Runs at a variable speed depending
+     * on the client count. */
+    if (aeCreateTimeEvent(server.el, 1, clientsTimerProc, NULL, NULL) == AE_ERR) {
+        serverPanic("Can't create event clientsTimerProc timer.");
         exit(1);
     }
 
@@ -5549,7 +5555,7 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "uptime_in_seconds:%I\r\n", (int64_t)uptime,
                 "uptime_in_days:%I\r\n", (int64_t)(uptime / (3600 * 24)),
                 "hz:%i\r\n", server.hz,
-                "configured_hz:%i\r\n", server.config_hz,
+                "configured_hz:%i\r\n", server.hz,
                 "lru_clock:%u\r\n", server.lruclock,
                 "executable:%s\r\n", server.executable ? server.executable : "",
                 "config_file:%s\r\n", server.configfile ? server.configfile : "",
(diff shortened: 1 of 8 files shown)


src/server.c
```diff
@@ -1126,6 +1110,40 @@ void clientsCron(void) {
...
+ */
+#define CLIENTS_CRON_MIN_ITERATIONS 5
+long long clientsTimerProc(struct aeEventLoop *eventLoop, long long id, void *clientData) {
+    UNUSED(eventLoop);
+    UNUSED(id);
+    UNUSED(clientData);
+
+    int numclients = listLength(server.clients);
+    int clients_this_cycle = numclients / server.hz; // Initial computation based on standard hz
+    int delayMs;
```
hpatro: nit: `delay_ms` or `delayms`.

hpatro (commented):
Seems right to me. Are there any possible unintended behavior change due to decoupling this from serverCron? Would be good to callout.

src/server.c
```diff
@@ -1126,6 +1110,40 @@ void clientsCron(void) {
...
+ *  - At least CLIENTS_CRON_MIN_ITERATIONS will be performed each cycle
+ */
+#define CLIENTS_CRON_MIN_ITERATIONS 5
+long long clientsTimerProc(struct aeEventLoop *eventLoop, long long id, void *clientData) {
+    UNUSED(eventLoop);
+    UNUSED(id);
+    UNUSED(clientData);
+
+    int numclients = listLength(server.clients);
+    int clients_this_cycle = numclients / server.hz; // Initial computation based on standard hz
```
enjoy-binbin:
```suggestion
    int clients_this_cycle = numclients / server.hz; /* Initial computation based on standard hz */
```

enjoy-binbin (approved):
Make sense to me. The code looks correct to me

src/server.c
```diff
@@ -1126,6 +1110,40 @@ void clientsCron(void) {
...
+        clients_this_cycle = MAX_CLIENTS_PER_CLOCK_TICK;
+        float required_hz = (float)numclients / MAX_CLIENTS_PER_CLOCK_TICK;
+        delayMs = 1000.0 / required_hz;
+    } else {
+        delayMs = 1000 / server.hz;
+    }
+
+    clientsCron(clients_this_cycle);
+
+    return delayMs;
```
JimB123:
```suggestion
    int delay_ms;

    if (clients_this_cycle < CLIENTS_CRON_MIN_ITERATIONS) {
        clients_this_cycle = min(numclients, CLIENTS_CRON_MIN_ITERATIONS);
    }

    if (clients_this_cycle > MAX_CLIENTS_PER_CLOCK_TICK) {
        clients_this_cycle = MAX_CLIENTS_PER_CLOCK_TICK;
        float required_hz = (float)numclients / MAX_CLIENTS_PER_CLOCK_TICK;
        delay_ms = 1000.0 / required_hz;
    } else {
        delay_ms= 1000 / server.hz;
    }

    clientsCron(clients_this_cycle);

    return delay_ms;
```

src/server.c
```diff
@@ -1126,6 +1110,40 @@ void clientsCron(void) {
...
+    if (clients_this_cycle < CLIENTS_CRON_MIN_ITERATIONS) {
+        clients_this_cycle = min(numclients, CLIENTS_CRON_MIN_ITERATIONS);
+    }
+
+    if (clients_this_cycle > MAX_CLIENTS_PER_CLOCK_TICK) {
+        clients_this_cycle = MAX_CLIENTS_PER_CLOCK_TICK;
+        float required_hz = (float)numclients / MAX_CLIENTS_PER_CLOCK_TICK;
+        delay_ms = 1000.0 / required_hz;
+    } else {
+        delay_ms= 1000 / server.hz;
```
JimB123:
```suggestion
        delay_ms = 1000 / server.hz;
```

src/server.c
```diff
@@ -1126,6 +1110,40 @@ void clientsCron(void) {
...
+    int clients_this_cycle = numclients / server.hz; /* Initial computation based on standard hz */
+    int delay_ms;
+
+    if (clients_this_cycle < CLIENTS_CRON_MIN_ITERATIONS) {
+        clients_this_cycle = min(numclients, CLIENTS_CRON_MIN_ITERATIONS);
+    }
+
+    if (clients_this_cycle > MAX_CLIENTS_PER_CLOCK_TICK) {
+        clients_this_cycle = MAX_CLIENTS_PER_CLOCK_TICK;
+        float required_hz = (float)numclients / MAX_CLIENTS_PER_CLOCK_TICK;
```
soloestoy: we need check `CONFIG_MAX_HZ`, ensure the `required_hz` would not exceed `CONFIG_MAX_HZ`.

src/server.c
```diff
@@ -1126,6 +1110,40 @@ void clientsCron(void) {
...
 }
 
+/* A periodic timer that performs client maintenance.
+ * This cron task follows the following rules:
+ *  - All clients need to be checked (at least) once per second
+ *  - To manage latency, we don't check more than MAX_CLIENTS_PER_CLOCK_TICK at a time
+ *  - The minimum rate will be defined by server.hz
+ *  - At least CLIENTS_CRON_MIN_ITERATIONS will be performed each cycle
+ */
+#define CLIENTS_CRON_MIN_ITERATIONS 5
```
soloestoy: it's better to align the name with `MAX_CLIENTS_PER_CLOCK_TICK` and put it to `server.h` together, we have the chance now

src/server.c
```diff
@@ -1126,6 +1110,40 @@ void clientsCron(void) {
     }
 }
 
+/* A periodic timer that performs client maintenance.
+ * This cron task follows the following rules:
+ *  - All clients need to be checked (at least) once per second
```
soloestoy: i thinks it's unnecessary, if we have 1 million clients, check all of them in one second is very expensive.

src/server.c
```diff
@@ -1126,6 +1110,40 @@ void clientsCron(void) {
...
+ *  - To manage latency, we don't check more than MAX_CLIENTS_PER_CLOCK_TICK at a time
+ *  - The minimum rate will be defined by server.hz
```
soloestoy, line 1117: actually the minimum rate is 1, may be less than `server.hz`

src/server.c
```diff
@@ -2775,10 +2776,15 @@ void initServer(void) {
...
-     * expired keys and so forth. */
+     * operations incrementally, like eviction of unaccessed expired keys, etc. */
     if (aeCreateTimeEvent(server.el, 1, serverCron, NULL, NULL) == AE_ERR) {
-        serverPanic("Can't create event loop timers.");
+        serverPanic("Can't create serverCron timer.");
+        exit(1);
+    }
+    /* A separate timer for client maintenance.  Runs at a variable speed depending
+     * on the client count. */
+    if (aeCreateTimeEvent(server.el, 1, clientsTimerProc, NULL, NULL) == AE_ERR) {
```
soloestoy: i'd love to reuse the `clientsCron`, just like `serverCron`.

src/server.c
```diff
@@ -5549,7 +5555,7 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "uptime_in_seconds:%I\r\n", (int64_t)uptime,
                 "uptime_in_days:%I\r\n", (int64_t)(uptime / (3600 * 24)),
                 "hz:%i\r\n", server.hz,
```
soloestoy, line 5566: i think it's better to show `clients_hz` here, since it is out of the hz control

src/server.c
```diff
@@ -1126,6 +1110,40 @@ void clientsCron(void) {
...
 
+/* A periodic timer that performs client maintenance.
+ * This cron task follows the following rules:
+ *  - All clients need to be checked (at least) once per second
+ *  - To manage latency, we don't check more than MAX_CLIENTS_PER_CLOCK_TICK at a time
+ *  - The minimum rate will be defined by server.hz
+ *  - At least CLIENTS_CRON_MIN_ITERATIONS will be performed each cycle
+ */
+#define CLIENTS_CRON_MIN_ITERATIONS 5
+long long clientsTimerProc(struct aeEventLoop *eventLoop, long long id, void *clientData) {
```
soloestoy: we should also count the time into `server.el_cron_duration`

src/server.c
JimB123:
```suggestion
        float required_hz = (float)numclients / MAX_CLIENTS_PER_CLOCK_TICK;
        if (required_hz > CONFIG_MAX_HZ) required_hz = CONFIG_MAX_HZ
```

src/server.c
```diff
@@ -1126,6 +1110,41 @@ void clientsCron(void) {
     }
 }
 
+/* A periodic timer that performs client maintenance.
+ * This cron task follows the following rules:
+ *  - All clients need to be checked (at least) once per second
+ *  - To manage latency, we don't check more than MAX_CLIENTS_PER_CLOCK_TICK at a time
+ *  - The minimum rate will be defined by server.hz
+ *  - At least CLIENTS_CRON_MIN_ITERATIONS will be performed each cycle
```
JimB123:
```suggestion
 * This cron task follows the following rules:
 *  - To manage latency, we don't check more than MAX_CLIENTS_PER_CLOCK_TICK at a time
 *  - The minimum rate will be defined by server.hz
 *  - The maxmum rate will be defined by CONFIG_MAX_HZ
 *  - At least CLIENTS_CRON_MIN_ITERATIONS will be performed each cycle
 *  - All clients need to be checked (at least) once per second (if possible given other constraints)
```

src/server.c
JimB123: `server.h` is a big dumping ground - and included almost everywhere.  I don't see a good reason to move an existing define which is already localized and move it into a global namespace.

Note: I wanted to leave as much code untouched as possible (including this define).  If that isn't a goal, I'd prefer making this a local `const` inside the function rather than moving the existing define to a more global namespace.

src/server.c
```diff
@@ -1126,6 +1110,42 @@ void clientsCron(void) {
...
+    int delay_ms;
+
+    if (clients_this_cycle < CLIENTS_CRON_MIN_ITERATIONS) {
+        clients_this_cycle = min(numclients, CLIENTS_CRON_MIN_ITERATIONS);
+    }
+
+    if (clients_this_cycle > MAX_CLIENTS_PER_CLOCK_TICK) {
+        clients_this_cycle = MAX_CLIENTS_PER_CLOCK_TICK;
+        float required_hz = (float)numclients / MAX_CLIENTS_PER_CLOCK_TICK;
+        if (required_hz > CONFIG_MAX_HZ) required_hz = CONFIG_MAX_HZ
```
JimB123:
```suggestion
        if (required_hz > CONFIG_MAX_HZ) required_hz = CONFIG_MAX_HZ;
```

src/server.c
JimB123: Given the new check for `CONFIG_MAX_HZ`, the maximum we can achieve is 200 clients * 500hz = 100,000 per sec.

src/server.c
JimB123: I intentionally didn't use "cron" because the name is ambiguous.  This function IS a time proc as defined by AE.
Naming is consistent with the defrag time proc (`activeDefragTimeProc`)

We have other functions called "cron" that are simply functions called by a timer proc.  Like `replicationCron()`, `databasesCron()`.

src/server.c
```diff
@@ -1046,38 +1046,23 @@ void getExpansiveClientsInfo(size_t *in_usage, size_t *out_usage) {
     *out_usage = o;
 }
 
-/* This function is called by serverCron() and is used in order to perform
+/* This function is called by clientsTimerProc() and is used in order to perform
```
JimB123:
```suggestion
/* This function is called by clientsTimeProc() and is used in order to perform
```

src/server.c
```diff
@@ -1046,38 +1046,23 @@ void getExpansiveClientsInfo(size_t *in_usage, size_t *out_usage) {
...
-/* This function is called by serverCron() and is used in order to perform
+/* This function is called by clientsTimeProc() and is used in order to perform
  * operations on clients that are important to perform constantly. For instance
  * we use this function in order to disconnect clients after a timeout, including
  * clients blocked in some blocking command with a non-zero timeout.
  *
  * The function makes some effort to process all the clients every second, even
- * if this cannot be strictly guaranteed, since serverCron() may be called with
- * an actual frequency lower than server.hz in case of latency events like slow
+ * if this cannot be strictly guaranteed, since clientsTimerProc() may be called with
```
JimB123:
```suggestion
 * if this cannot be strictly guaranteed, since clientsTimeProc() may be called with
```

src/server.c
```diff
@@ -1126,6 +1110,42 @@ void clientsCron(void) {
...
+/* A periodic timer that performs client maintenance.
+ * This cron task follows the following rules:
+ *  - To manage latency, we don't check more than MAX_CLIENTS_PER_CLOCK_TICK at a time
+ *  - The minimum rate will be defined by server.hz
+ *  - The maxmum rate will be defined by CONFIG_MAX_HZ
+ *  - At least CLIENTS_CRON_MIN_ITERATIONS will be performed each cycle
+ *  - All clients need to be checked (at least) once per second (if possible given other constraints)
+ */
+#define CLIENTS_CRON_MIN_ITERATIONS 5
+long long clientsTimerProc(struct aeEventLoop *eventLoop, long long id, void *clientData) {
```
JimB123:
```suggestion
long long clientsTimeProc(struct aeEventLoop *eventLoop, long long id, void *clientData) {
```

src/server.c
```diff
@@ -2775,10 +2778,15 @@ void initServer(void) {
...
-     * expired keys and so forth. */
+     * operations incrementally, like eviction of unaccessed expired keys, etc. */
     if (aeCreateTimeEvent(server.el, 1, serverCron, NULL, NULL) == AE_ERR) {
-        serverPanic("Can't create event loop timers.");
+        serverPanic("Can't create serverCron timer.");
+        exit(1);
+    }
+    /* A separate timer for client maintenance.  Runs at a variable speed depending
+     * on the client count. */
+    if (aeCreateTimeEvent(server.el, 1, clientsTimerProc, NULL, NULL) == AE_ERR) {
```
JimB123:
```suggestion
    if (aeCreateTimeEvent(server.el, 1, clientsTimeProc, NULL, NULL) == AE_ERR) {
```

src/server.c
```diff
@@ -2775,10 +2778,15 @@ void initServer(void) {
...
+        serverPanic("Can't create event clientsTimerProc timer.");
```
JimB123:
```suggestion
        serverPanic("Can't create event clientsTimeProc timer.");
```

src/server.c
JimB123: Not sure what you're asking/suggesting here.  I don't think we can just change the definition of the existing metric.

I think maybe you're suggesting a new metric?  Currently the required hz for the `clientsTimeProc` is computed locally to the function, in each iteration.  Are you suggesting adding a new field to `server...` and adding a metric above `hz` with `clients_hz`?

src/server.c
JimB123: Currently `server.el_cron_duration` is set to the duration of the most recent `serverCron`.  It's not cumulative or additive.  Then, in `beforeSleep()`, we add (locals) `duration_before_aof` & `duration_after_write` and record the sum with `durationAddSample` which both accumulates a total duration AND increments a counter of samples taken.

I think you are suggesting that inside `serverCron` we should alter the existing code to do `server.el_cron_duration += ...` and also increment `server.el_cron_duration` in this function also.

If we do this, it will affect the samples taken counter as we will get samples for both `serverCron` and the `clientTimeProc`.  Usually, these won't happen in the same event loop cycle, but at times they might align.  

Thoughts?

*As a side note, I notice that `server.el_cron_duration` is defined as `monotime`.  That is not the intended usage for the type.  `monotime` is intended as a timestamp (a point in time).  Durations should not be `monotime` as they are not timestamps.  This is shown in `monotonic.h` in that all of the duration functions return `uint64_t`, not `monotime`.*

src/server.c
soloestoy: if we wanna minimize the scope, we can put `MAX_CLIENTS_PER_CLOCK_TICK` here and change `CLIENTS_CRON_MIN_ITERATIONS` to `MIN_CLIENTS_PER_CLOCK_TICK`

src/server.c
soloestoy: yes, a new metric, since the `clients_hz` can be changed and affect the poll_wait's sleep time, this will further affect the CPU usage

src/server.c
soloestoy: I'm not a native English speaker : ), what is the specific difference between cron and timeproc? I'm more accustomed to cron, and the statistics related to its timing are also el_cron_duration.

src/server.c
soloestoy: 1. in `serverCron` we get `server.el_cron_duration += serverCron`
2. in `clientsCron` we get `server.el_cron_duration += clientsCron`
3. in `beforeSleep` we get `server.el_cron_duration += duration_before_aof + duration_after_write` and use `durationAddSample` to accumulate the total duration, and reset `server.el_cron_duration` to `0`

src/server.c
JimB123: If you look at `ae.h`, the callback is defined as `aeTimeProc` here:  https://github.com/valkey-io/valkey/blob/unstable/src/ae.h#L70

"cron" seems to be used for anything that's invoked at some interval.  This function is specifically an `aeTimeProc`.

soloestoy: I know `ae.h`, but the underlying logic doesn't necessarily need to be passed through to the upper layers. Logically speaking, cron is also suitable and easy to understand.

src/server.c
```diff
@@ -1126,6 +1110,42 @@ void clientsCron(void) {
...
+        if (required_hz > CONFIG_MAX_HZ) required_hz = CONFIG_MAX_HZ;
+        delay_ms = 1000.0 / required_hz;
+    } else {
+        delay_ms = 1000 / server.hz;
+    }
+
+    clientsCron(clients_this_cycle);
+
+    return delay_ms;
+}
```
JimB123:
```suggestion
long long clientsTimeProc(struct aeEventLoop *eventLoop, long long id, void *clientData) {
    UNUSED(eventLoop);
    UNUSED(id);
    UNUSED(clientData);
    
    const int MIN_CLIENTS_PER_CYCLE = 5;
    const int MAX_CLIENTS_PER_CYCLE = 200;
    
    monotonic start_time;
    elapsedStart(&start_time);

    int numclients = listLength(server.clients);
    int clients_this_cycle = numclients / server.hz; /* Initial computation based on standard hz */
    int delay_ms;

    if (clients_this_cycle < MIN_CLIENTS_PER_CYCLE) {
        clients_this_cycle = min(numclients, MIN_CLIENTS_PER_CYCLE);
    }

    if (clients_this_cycle > MAX_CLIENTS_PER_CYCLE) {
        clients_this_cycle = MAX_CLIENTS_PER_CYCLE;
        float required_hz = (float)numclients / MAX_CLIENTS_PER_CYCLE;
        if (required_hz > CONFIG_MAX_HZ) required_hz = CONFIG_MAX_HZ;
        delay_ms = 1000.0 / required_hz;
    } else {
        delay_ms = 1000 / server.hz;
    }

    clientsCron(clients_this_cycle);

    server.clients_hz = 1000 / delay_ms;
    server.el_cron_duration += elapsedUs(start_time);
    return delay_ms;
}
```

src/server.h
```diff
@@ -1681,10 +1680,6 @@ struct valkeyServer {
     char *configfile;         /* Absolute config file path, or NULL */
     char *executable;         /* Absolute executable file path. */
     char **exec_argv;         /* Executable argv vector (copy). */
-    int dynamic_hz;           /* Change hz value depending on # of clients. */
-    int config_hz;            /* Configured HZ value. May be different than
-                                 the actual 'hz' field value if dynamic-hz
-                                 is enabled. */
     mode_t umask;             /* The umask value of the process on startup */
     int hz;                   /* serverCron() calls frequency in hertz */
```
JimB123, line 1684:
```suggestion
    int hz;                   /* serverCron() calls frequency in hertz */
    int clients_hz;           /* clientsTimeProc() frequency in hertz */
```

src/server.c
JimB123: done.

src/server.c
JimB123: done

src/server.c
JimB123: Maybe we'll have to agree to disagree on this one.

This function carries the specific signature defined in `ae.h`.  It has the parameters of an `aeTimeProc`.  I want to highlight that THIS is the function directly invoked by AE as a timer callback.  This function performs the management part of the timer (how often it should fire, some metrics, etc.).

If you scroll down about 20 lines, you'll see that this function actually calls another function called `clientsCron()` which is largely the same as the original `clientsCron()` which existed before this function was created.  This is the actual business logic (separate from timer management) that needs to get invoked.

So, consider it this way.  `clientsCron()` really hasn't changed.  It used to be called by `serverCron()` and now it's called by `clientsTimeProc()`.

We could, for consistency, create a `serverTimeProc` to call `serverCron`.  If so, like I've done here, that would separate the management of the timer from the work that's done when the timer is fired.

src/server.c
JimB123: I've aligned the names as you've suggested and localized them both to the function where they are used.

src/server.c
```diff
@@ -1126,6 +1110,49 @@ void clientsCron(void) {
...
+ */
+long long clientsTimeProc(struct aeEventLoop *eventLoop, long long id, void *clientData) {
+    UNUSED(eventLoop);
+    UNUSED(id);
+    UNUSED(clientData);
+    
+    const int MIN_CLIENTS_PER_CYCLE = 5;
+    const int MAX_CLIENTS_PER_CYCLE = 200;
+    
+    monotonic start_time;
```
JimB123:
```suggestion
    monotime start_time;
```

src/server.c
```diff
@@ -1126,6 +1110,49 @@ void clientsCron(void) {
...
+ */
+long long clientsTimeProc(struct aeEventLoop *eventLoop, long long id, void *clientData) {
+    UNUSED(eventLoop);
+    UNUSED(id);
+    UNUSED(clientData);
+    
+    const int MIN_CLIENTS_PER_CYCLE = 5;
+    const int MAX_CLIENTS_PER_CYCLE = 200;
+    
+    monotime start_time;
```
JimB123:
```suggestion
    UNUSED(clientData);

    const int MIN_CLIENTS_PER_CYCLE = 5;
    const int MAX_CLIENTS_PER_CYCLE = 200;

    monotime start_time;
```

madolson (approved):
I suppose I'm dubious about the new info field, but not sure I feel that strongly about it.

PR #1388, opened 2024-12-03
defrag: use monotime in module interface

JimB123:
The recent PR (https://github.com/valkey-io/valkey/pull/1242) converted Active Defrag to use `monotime`.  In that change, a conversion was performed to continue to use `ustime()` as part of the module interface.  Since this time is only used internally, and never actually exposed to the module, we can convert this to use `monotime` directly.


hpatro (approved):

hpatro:
The test failure seems unrelated.

PR #1389, opened 2024-12-04
Optimize ZRANK to avoid path comparisons

ranshid:
ZRANK is a widly used command for workloads using sorted-sets. For example, in leaderboards It enables query the specific rank of a player. 
The way ZRANK is currently implemented is:
1. locate the element in the SortedSet hashtable.
2. take the score of the element and use it in order to locate the element in the SkipList (when listpack encoding is not used) 
3. During the SkipLis scan for the elemnt we keep the path and use it in order to sum the span in each path node in order to calculate the elemnt rank

One problem with this approach is that it involves multiple compare operations in order to locate the element. Specifically string comparison can be expensive since it will require access multiple memory locations for the items the element string is compared against.
Perf analysis showed this can take up to 20% of the rank scan time
[
perf
](url)

We can improve the rank search by taking advantage of the fact that the element node in the skiplist is pointed by the hashtable value! 
Our Skiplist implementation is using FatKeys, where each added node is assigned a randomly chosen height. Say we keep a height record for every skiplist element. In order to get an element rank we simply:
1. locate the element in the SortedSet hashtable.
2. we go directly to the node in the skiplist.
3. we jump to the full height of the node and take the span value.
4. we continue going foreward and always jump to the heighst point in each node we get to, making sure to sum all the spans.
5. we take off the summed spans from the SkipList length and we now have the specific node rank :) 

In order to test this method I created several benchmarks. All benchmarks used the same seeds and the lists contained 1M elements. Since a very important factor is the number of scores compared to the number of elements (since small ratio means more string compares during searches) each benchmark test used different number of scores (1, 10K, 100K, 1M)
some results:

**TPS**

Scores range | non-optimized | optimized | gain
-- | -- | -- | --
1 | 416042 | 605363 | 45.51%
10K | 359776 | 459200 | 27.63%
100K | 380387 | 459157 | 20.71%
1M | 416059 | 450853 | 8.36%

**Latency**

Scores range | non-optimized | optimized | gain
-- | -- | -- | --
1 | 1.191000 | 0.831000 | -30.23%
10K | 1.383000 | 1.095000 | -20.82%
100K | 1.311000 | 1.087000 | -17.09%
1M | 1.191000 | 1.119000 | -6.05%

###  Memory efficiency

adding another field to each skiplist node can cause degredation in memory efficiency for large sortedsets. We use the fact that level 0 recorded span of ALL nodes can either be 1 or zero (for the last node). So we use wrappers in order to get a node span and override the span for level 0 to hold the node height.

### DIFF
--- a/src/t_zset.c
+++ b/src/t_zset.c
@@ -112,11 +112,11 @@ static inline void zslSetNodeHeight(zskiplistNode *x, int height) {
 
 /* Create a skiplist node with the specified number of levels.
  * The SDS string 'ele' is referenced by the node after the call. */
-zskiplistNode *zslCreateNode(int level, double score, sds ele) {
-    zskiplistNode *zn = zmalloc(sizeof(*zn) + level * sizeof(struct zskiplistLevel));
+zskiplistNode *zslCreateNode(int height, double score, sds ele) {
+    zskiplistNode *zn = zmalloc(sizeof(*zn) + height * sizeof(struct zskiplistLevel));
     zn->score = score;
     zn->ele = ele;
-    zslSetNodeHeight(zn, level);
+    zslSetNodeHeight(zn, height);
     return zn;
 }
 
@@ -551,10 +551,7 @@ unsigned long zslGetRank(zskiplist *zsl, double score, sds ele) {
     return 0;
 }
 
-/* Find the rank for an element by both score and key.
- * Returns 0 when the element cannot be found, rank otherwise.
- * Note that the rank is 1-based due to the span of zsl->header to the
- * first element. */
+/* Find the rank for a specific skipllist node. */
 unsigned long zslGetRankByNode(zskiplist *zsl, zskiplistNode *x) {
     int i = zslGetNodeHeight(x) - 1;
     unsigned long rank = zslGetNodeSpanAtLevel(x, i);
--- a/src/t_zset.c
+++ b/src/t_zset.c
@@ -72,12 +72,45 @@ void zsetConvertAndExpand(robj *zobj, int encoding, unsigned long cap);
 zskiplistNode *zslGetElementByRankFromNode(zskiplistNode *start_node, int start_level, unsigned long rank);
 zskiplistNode *zslGetElementByRank(zskiplist *zsl, unsigned long rank);
 
+static inline zskiplistNode *zslGetNodeByScoreRef(double *ref) {
+    return (zskiplistNode *)(ref - 1);
+}
+
+static inline unsigned long zslGetNodeSpanAtHeight(zskiplistNode *x, int height) {
+    return height > 0 ? x->level[height].span : x->level[height].forward ? 1
+                                                                         : 0;
+}
+
+static inline void zslSetNodeSpanAtHeight(zskiplistNode *x, int height, unsigned long span) {
+    if (height > 0)
+        x->level[height].span = span;
+}
+
+static inline void zslIncrNodeSpanAtHeight(zskiplistNode *x, int height, unsigned long incr) {
+    if (height > 0)
+        x->level[height].span += incr;
+}
+
+static inline void zslDecrNodeSpanAtHeight(zskiplistNode *x, int height, unsigned long decr) {
+    if (height > 0)
+        x->level[height].span -= decr;
+}
+
+static inline unsigned long zslGetNodeHeight(zskiplistNode *x) {
+    return x->level[0].span;
+}
+
+static inline void zslSetNodeHeight(zskiplistNode *x, int height) {
+    x->level[0].span = height;
+}
+
 /* Create a skiplist node with the specified number of levels.
  * The SDS string 'ele' is referenced by the node after the call. */
 zskiplistNode *zslCreateNode(int level, double score, sds ele) {
     zskiplistNode *zn = zmalloc(sizeof(*zn) + level * sizeof(struct zskiplistLevel));
     zn->score = score;
     zn->ele = ele;
+    zslSetNodeHeight(zn, level - 1);
     return zn;
 }
 
@@ -147,7 +180,7 @@ zskiplistNode *zslInsert(zskiplist *zsl, double score, sds ele) {
         while (x->level[i].forward &&
                (x->level[i].forward->score < score ||
                 (x->level[i].forward->score == score && sdscmp(x->level[i].forward->ele, ele) < 0))) {
-            rank[i] += x->level[i].span;
+            rank[i] += zslGetNodeSpanAtHeight(x, i);
             x = x->level[i].forward;
         }
         update[i] = x;
@@ -161,9 +194,10 @@ zskiplistNode *zslInsert(zskiplist *zsl, double score, sds ele) {
         for (i = zsl->level; i < level; i++) {
             rank[i] = 0;
             update[i] = zsl->header;
-            update[i]->level[i].span = zsl->length;
+            zslSetNodeSpanAtHeight(update[i], i, zsl->length);
         }
         zsl->level = level;
+        zslSetNodeHeight(zsl->header, level - 1);
     }
     x = zslCreateNode(level, score, ele);
     for (i = 0; i < level; i++) {
@@ -171,13 +205,13 @@ zskiplistNode *zslInsert(zskiplist *zsl, double score, sds ele) {
         update[i]->level[i].forward = x;
 
         /* update span covered by update[i] as x is inserted here */
-        x->level[i].span = update[i]->level[i].span - (rank[0] - rank[i]);
-        update[i]->level[i].span = (rank[0] - rank[i]) + 1;
+        zslSetNodeSpanAtHeight(x, i, zslGetNodeSpanAtHeight(update[i], i) - (rank[0] - rank[i]));
+        zslSetNodeSpanAtHeight(update[i], i, (rank[0] - rank[i]) + 1);
     }
 
     /* increment span for untouched levels */
     for (i = level; i < zsl->level; i++) {
-        update[i]->level[i].span++;
+        zslIncrNodeSpanAtHeight(update[i], i, 1);
     }
 
     x->backward = (update[0] == zsl->header) ? NULL : update[0];
@@ -195,10 +229,10 @@ void zslDeleteNode(zskiplist *zsl, zskiplistNode *x, zskiplistNode **update) {
     int i;
     for (i = 0; i < zsl->level; i++) {
         if (update[i]->level[i].forward == x) {
-            update[i]->level[i].span += x->level[i].span - 1;
+            zslIncrNodeSpanAtHeight(update[i], i, zslGetNodeSpanAtHeight(x, i) - 1);
             update[i]->level[i].forward = x->level[i].forward;
         } else {
-            update[i]->level[i].span -= 1;
+            zslDecrNodeSpanAtHeight(update[i], i, 1);
         }
     }
     if (x->level[0].forward) {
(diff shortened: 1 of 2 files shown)


src/t_zset.c
```diff
@@ -72,12 +72,41 @@ void zsetConvertAndExpand(robj *zobj, int encoding, unsigned long cap);
 zskiplistNode *zslGetElementByRankFromNode(zskiplistNode *start_node, int start_level, unsigned long rank);
 zskiplistNode *zslGetElementByRank(zskiplist *zsl, unsigned long rank);
 
+static inline unsigned long zslGetNodeSpanAtHeight(zskiplistNode *x, int height) {
+    return height > 0 ? x->level[height].span : x->level[height].forward ? 1
+                                                                         : 0;
```
zuiderkwast: Nested `?:` is easier to read if it's indented similar to `else if` ...

```C
    return height > 0 ?
               x->level[height].span :
           x->level[height].forward ?
               1 :
           0;
```

... but I suppose clang-format doesn't like this style. Let's avoid the nested ?: and use `if` in one of the levels?

```suggestion
    if (height > 0) return x->level[height].span;
    return x->level[height].forward ? 1 : 0;
```

src/t_zset.c
```diff
@@ -72,12 +72,41 @@ void zsetConvertAndExpand(robj *zobj, int encoding, unsigned long cap);
...
+}
+
+static inline void zslSetNodeSpanAtHeight(zskiplistNode *x, int height, unsigned long span) {
+    if (height > 0)
+        x->level[height].span = span;
+}
```
zuiderkwast: From the code that was refactored, it appears that this height should always exist. It looks wrong if this is a no-op for invalid heights. Can we delete the `if` or change it to an assert?

src/t_zset.c
```diff
@@ -72,12 +72,41 @@ void zsetConvertAndExpand(robj *zobj, int encoding, unsigned long cap);
...
+
+static inline void zslIncrNodeSpanAtHeight(zskiplistNode *x, int height, unsigned long incr) {
+    if (height > 0)
+        x->level[height].span += incr;
```
zuiderkwast: Change to assert?

src/t_zset.c
```diff
@@ -72,12 +72,41 @@ void zsetConvertAndExpand(robj *zobj, int encoding, unsigned long cap);
...
+}
+
+static inline void zslDecrNodeSpanAtHeight(zskiplistNode *x, int height, unsigned long decr) {
+    if (height > 0)
+        x->level[height].span -= decr;
```
zuiderkwast: Change to assert?

src/t_zset.c
```diff
@@ -72,12 +72,41 @@ void zsetConvertAndExpand(robj *zobj, int encoding, unsigned long cap);
...
+}
+
+static inline unsigned long zslGetNodeHeight(zskiplistNode *x) {
+    return x->level[0].span;
```
zuiderkwast, line 104: The span of level 0 used for storing the height? This is unintuitive so let's add a comment for this, here and there. This code looks like a typo if you don't know this encoding.

```suggestion
    /* Since the span at level 0 is always 1 (or 0 for the last node), this
     * field is instead used for storing the height of the node. */
    return x->level[0].span;
```

src/t_zset.c
```diff
@@ -72,12 +72,41 @@ void zsetConvertAndExpand(robj *zobj, int encoding, unsigned long cap);
...
+}
+
+static inline void zslSetNodeHeight(zskiplistNode *x, int height) {
+    x->level[0].span = height;
```
zuiderkwast, line 110: Comment here too.

src/t_zset.c
```diff
@@ -72,12 +72,41 @@ void zsetConvertAndExpand(robj *zobj, int encoding, unsigned long cap);
 zskiplistNode *zslGetElementByRankFromNode(zskiplistNode *start_node, int start_level, unsigned long rank);
 zskiplistNode *zslGetElementByRank(zskiplist *zsl, unsigned long rank);
 
+static inline unsigned long zslGetNodeSpanAtHeight(zskiplistNode *x, int height) {
```
zuiderkwast: There is another word "level", used in the structs. Let's distinguish between level and height?

0 ≤ *level* < *height* ?

```suggestion
static inline unsigned long zslGetNodeSpanAtLevel(zskiplistNode *x, int level) {
```

src/t_zset.c
```diff
@@ -72,12 +72,41 @@ void zsetConvertAndExpand(robj *zobj, int encoding, unsigned long cap);
...
+    x->level[0].span = height;
+}
+
 /* Create a skiplist node with the specified number of levels.
  * The SDS string 'ele' is referenced by the node after the call. */
 zskiplistNode *zslCreateNode(int level, double score, sds ele) {
     zskiplistNode *zn = zmalloc(sizeof(*zn) + level * sizeof(struct zskiplistLevel));
     zn->score = score;
     zn->ele = ele;
+    zslSetNodeHeight(zn, level - 1);
```
zuiderkwast: It looks odd that we set height to `level - 1`. Isn't the height the number of levels in a node?

Is `level` here the number of levels here? Maybe we should rename the level parameter to height?

src/t_zset.c
```diff
@@ -511,6 +541,21 @@ unsigned long zslGetRank(zskiplist *zsl, double score, sds ele) {
...
 }
 
+/* Find the rank for an element by both score and key.
+ * Returns 0 when the element cannot be found, rank otherwise.
+ * Note that the rank is 1-based due to the span of zsl->header to the
+ * first element. */
+unsigned long zslGetRankByNode(zskiplist *zsl, zskiplistNode *x) {
+    int i = zslGetNodeHeight(x);
+    unsigned long rank = zslGetNodeSpanAtHeight(x, i);
+    while (x->level[zslGetNodeHeight(x)].forward) {
```
zuiderkwast: The height is the top level index? If level is 0-based, I would think the top level is `height - 1`.

Maybe we should either change this to `zslGetNodeHeight(x) - 1` or rename height to top-level as `zslGetNodeTopLevelIndex(x)`?

zuiderkwast (commented):
Very smart idea.

Instead of looking up a skiplist node, we start at a node find the way out of the skip list.

What can we call this the inverse lookup? It deserves a cool name. :sunglasses:

src/t_zset.c
ranshid: So the span at loevel 0 can change and either be 1 or 0. In case (for example) a node is deleted from the end of the list the node preceeding it will be set to span 0.

src/t_zset.c
ranshid: I am not sure if that would work for all code flows. let me check

src/t_zset.c
ranshid: So yes. height is the number of levels in the node. I can make this change

src/t_zset.c
ranshid: Yes I keep the height as the top level index. I think it might be more readable if I keep the height as the number of levels.

src/t_zset.c
ranshid: > What can we call this the inverse lookup? It deserves a cool name.

RRC (Revers Rank Check)?

src/t_zset.c
ranshid: this will not work. assert can happen on the good path

src/t_zset.c
ranshid: this will not work. assert can happen on the good path

src/t_zset.c
ranshid: I added different comments to explain the change

src/t_zset.c
ranshid: done

src/t_zset.c
ranshid: done

src/t_zset.c
zuiderkwast: OK. And it's not a good idea to check `if (height > 0)` at the call site instead of inside the function?

src/t_zset.c
```diff
@@ -511,6 +551,21 @@ unsigned long zslGetRank(zskiplist *zsl, double score, sds ele) {
     return 0;
 }
 
+/* Find the rank for an element by both score and key.
+ * Returns 0 when the element cannot be found, rank otherwise.
+ * Note that the rank is 1-based due to the span of zsl->header to the
+ * first element. */
+unsigned long zslGetRankByNode(zskiplist *zsl, zskiplistNode *x) {
```
zuiderkwast: What do you mean by "Returns 0 when the element cannot be found"? The element is provided by the caller in `x`, isn't it?

It returns 0 only if `x` is the last element in the skiplist?

src/t_zset.c
```diff
@@ -1173,6 +1228,13 @@ unsigned char *zzlDeleteRangeByRank(unsigned char *zl, unsigned int start, unsig
  * Common sorted set API
  *----------------------------------------------------------------------------*/
 
+/* Utility function used for mapping the hashtable entry to the matching skiplist node.
+ * For example, this is used in case of ZRANK query. */
+static inline zskiplistNode *zsetGetSLNodeByEntry(dictEntry *de) {
+    double *score_ref = ((double *)dictGetVal(de));
+    return (zskiplistNode *)(score_ref - 1);
+}
```
zuiderkwast: Oooh, exciting stuff... So the dict value is a pointer to the double score within the skiplist node?

This becomes quite weird here. A better design would be that the dict value points to the skiplist node. Would that be a very big change?

src/t_zset.c
```diff
@@ -72,12 +72,51 @@ void zsetConvertAndExpand(robj *zobj, int encoding, unsigned long cap);
...
+    x->level[0].span = height;
+}
+
 /* Create a skiplist node with the specified number of levels.
  * The SDS string 'ele' is referenced by the node after the call. */
 zskiplistNode *zslCreateNode(int level, double score, sds ele) {
     zskiplistNode *zn = zmalloc(sizeof(*zn) + level * sizeof(struct zskiplistLevel));
     zn->score = score;
     zn->ele = ele;
+    zslSetNodeHeight(zn, level);
```
zuiderkwast: Nit: We could also rename the parameter `level` to `height` in this function, alternatively `num_levels`.

src/t_zset.c
ranshid: Sorry - bad copy-and-paste :) let me fix that

src/t_zset.c
ranshid: I also thought about it but wanted to keep this change modest. Since we will probably change the structure after the hashtable introduction I think making a bigger change is not very important right now.

src/t_zset.c
ranshid: I think since these functions are called in so many places it simplifies the code doing it internally.

src/t_zset.c
zuiderkwast: OK, we can make the larger change later then.

src/t_zset.c
```diff
@@ -511,6 +551,18 @@ unsigned long zslGetRank(zskiplist *zsl, double score, sds ele) {
     return 0;
 }
 
+/* Find the rank for a specific skipllist node. */
```
```suggestion
/* Find the rank for a specific skiplist node. */
```

src/t_zset.c
zuiderkwast: I wonder why this works though and if it works on all possible platforms and compilers. The struct is

```C
typedef struct zskiplistNode {
    sds ele;
    double score;
    ...
} zskiplistNode;
```

We have a pointer to `score`. We treat it as an array of doubles and look at the previous element, which is `ele`, but `ele` is a pointer. On 32-bit, a pointer and a double are not of the same size, but due to alignment, it happens to work anyway.

Should we just trust the tests here then? We do try to support some odd system where we don't have CI. This definitely needs a follow-up...

zuiderkwast (approved):
I guess I'm good with merging this then.

src/t_zset.c
ranshid: Absolutly true. will fix it now

zuiderkwast: `offsetof(zskiplistNode, score)` is perfect. Thanks!

zuiderkwast (approved):
Perfect

Issue #900, opened 2024-08-12
[NEW] Benchmark CLI tool rewrite


vitarb:
Existing benchmark tool is dated. The way it's implemented makes it difficult to introduce new data types or custom workloads into it. Its support for cluster mode is also very limited. I suggest we collect requirements and consider rewriting it from scratch.

Some use cases that I would like to be supported in the future:
- Support for custom data types, such as JSON, vectors, etc.
- Support for mixed workloads, such as read/write/delete in certain order with ability to operate on a given key across multiple operations, potentially with some delay.
- Better cluster mode support (ability to read from replicas), so we can test maximum read throughput of the cluster.
- Support for distributed execution (in order to be able to saturate larger clusters)
- Metrics publishing (support for prometheus), so we can see detailed performance information over time tagged with different workloads, and be able to aggregate the data for distributed runs more easily.

This issue is an attempt to gauge interest in this effort from the valkey community and gather additional use cases that hypothetical rewrite should address.

bbarani:
Some additional requirements to consider:

- Ability to view cluster / node / memory level stats during benchmarking
- Support for distributed workloads for large clusters on multiple machines.
- Ability to generate custom workload from a cluster 
- Ability to compare specific runs (  baseline vs contender )
- Ability to simulate the number of clients

poiuj:
Great idea. Recently I needed to benchmark specific cases. I ended up with ad hoc python script that does what was missing for me in the benchmark CLI tool. Specifically I wanted to fill a node with data, run "warm up" to randomize the memory layout, then introduce fragmentation, and only then run specific benchmark with benchmark CLI tool. 

Other missing features to consider:
1. Using different value sizes. Real life workloads rarely use values of the exact same size. The benchmark tool should support this.
2. Using different distributions over keys range. Today benchmark CLI supports uniform distribution. Which is pretty artificial. To emulate hot keys, we want to use a different distribution.

Bot of the cases could be solved by running several benchmark tools in parallel. But it's rather a workaround.

vitarb:
> could be solved by running several benchmark tools in parallel. But it's rather a workaround

I agree, having more flexibility in ability to define workloads that consist of more than one command configuration would be helpful. This would allow mixed writes, read/write/delete, etc. Combining results from multiple parallel runs can be a little tricky too.

hpatro:
[...]

PR #1392, opened 2024-12-05
Support for reading from replicas in valkey-benchmark

bluayer:
**Background**
When conducting performance tests using `valkey-benchmark`, reading from replicas was not supported. Consequently, even in cluster mode, all reads were directed to the primary nodes. This limitation made it challenging to obtain accurate metrics during workload stress testing for performance measurement or before a version upgrade.

Related issue : https://github.com/valkey-io/valkey/issues/900

**Changes**
1. Replaced the use of `CLUSTER NODES` with `CLUSTER SLOTS` when fetching cluster configuration. This allows for easier identification of replica slots.
2. Support for reading from replicas by executing the client in `READONLY` mode.
3. Support reading from replicas even during slot migrations.
4. Introduced a CLI option `--rfr` to enable reading from replicas only or all cluster nodes. A warning added to indicate that write requests might not be handled correctly when using this option.


bluayer:
I fully understand that the maintainers of this project are extremely busy and prioritize tasks based on the importance. However, I believe this PR could be a good contribution to this project. If you don't mind, can I ask for a PR review or a reviewer assignment? @madolson

madolson:
@ranshid Can you take a look?

src/valkey-benchmark.c
```diff
@@ -1056,156 +1067,67 @@ static clusterNode **addClusterNode(clusterNode *node) {
...
+            assert(nr->element[0]->str != NULL);
+            assert(nr->element[2]->str != NULL);
+            sds ip = sdsnew(nr->element[0]->str);
+            sds name = sdsnew(nr->element[2]->str);
+            int port = nr->element[1]->integer;
+
+            printf("\nNode %s [%d-%d] %s:%d\n", name, from, to, ip, port);
+
+            clusterNode *node = NULL;
             node = createClusterNode(sdsnew(ip), port);
```
ranshid: why are we creating the node again each time we check a new range? IIRC the same node can apear several times for different ranges

src/valkey-benchmark.c
```diff
@@ -1056,156 +1067,67 @@ static clusterNode **addClusterNode(clusterNode *node) {
...
+    assert(reply->type == REDIS_REPLY_ARRAY);
+    for (i = 0; i < reply->elements; i++) {
+        redisReply *r = reply->element[i];
+        assert(r->type == REDIS_REPLY_ARRAY);
+        assert(r->elements >= 3);
+        int from = r->element[0]->integer;
+        int to = r->element[1]->integer;
+        for (j = 2; j < r->elements; j++) {
+            int is_primary = (j == 2);
+            if (config.read_from_replicas == is_primary) continue;
```
ranshid: Not sure why we choose to skip the primary and only read from replicas. we can read from both primary and replicas IMO

src/valkey-benchmark.c
```diff
@@ -1279,30 +1203,41 @@ static int fetchClusterSlotsConfiguration(client c) {
...
-            if (name) sdsfree(name);
-            goto cleanup;
+        size_t start, end;
+        if (config.read_from_replicas) {
+            start = 3;
+            end = r->elements;
+        } else {
+            start = 2;
+            end = 3;
+        }
```
ranshid: same question here. why not include primaries as well?

src/valkey-benchmark.c
```diff
@@ -1557,6 +1494,10 @@ int parseOptions(int argc, char **argv) {
         "                    If the command is supplied on the command line in cluster\n"
         "                    mode, the key must contain \"{tag}\". Otherwise, the\n"
         "                    command will not be sent to the right cluster node.\n"
+        " --replicas         Enable read from replicas in cluster mode.\n"
```
ranshid: Suggest use --read-from-replicas or --rfr which is much more intuitive  IMO

src/valkey-benchmark.c
```diff
@@ -1056,156 +1067,67 @@ static clusterNode **addClusterNode(clusterNode *node) {
...
-                        p += 3;
-                        char *closing_bracket = strchr(p, ']');
-                        if (closing_bracket) *closing_bracket = '\0';
-                        sds slot = sdsnew(slotsdef);
-                        sds src = sdsnew(p);
-                        node->importing_count += 2;
-                        node->importing = zrealloc(node->importing, (node->importing_count * sizeof(sds)));
-                        node->importing[node->importing_count - 2] = slot;
-                        node->importing[node->importing_count - 1] = src;
-                    }
```
ranshid, line 1185: If we are no longer managing slots migration we should probably remove it from te clusterNode struct entierly.
I am not sure what was the original motivation to manage it, but probably reading the slots configuration again when we get ASK error is not very helpfull for the benchmark - maybe weneed to fix that one as well

ranshid (commented):
Mostly LGTM. some comments. specifically I did not understand why we are not supporting reading form the primnaries as well

src/valkey-benchmark.c
bluayer: Thank you for pointing that out. I'll fix it.

src/valkey-benchmark.c
bluayer: Discussion is needed in this regard.
I've seen many clusters that operate with "Primary" handling only writes and "Replica" handling only reads. (Of course, this is despite the fact that Primary can read/write).
Therefore, there were three directions for implementation, and I adopted method 1, which is similar to the real world I've seen.

1/ Read from the replica option only reads from the replica, and if you need to read from the primary as well, run a separate process (current implementation)
2/ Read from the replica option reads from the entire cluster, including the primary (the way you described)
3/ Separate options depending on whether to include the primary

Changing to the method you told me is not difficult, please let me know your thoughts :)

bluayer: In addition, if we include the primary, we need to think about the handling write command.
e.g. `valkey-benchmark -h cluster_host -c 4 -t get,set -n 1000 --cluster --rfr`

I understand that I need to change the client to redirect to primary. It may be possible to simply divide reading into primary, but it seems likely that users want to be able to handle writing itself(valkey-benchmark). I think handling write commands should be implemented at other PR.

src/valkey-benchmark.c
bluayer: I agree 👍

src/valkey-benchmark.c
bluayer: I agree and I'll remove attributes(migrating, importing) from the struct. 
But, IMO it would be good to fetch the cluster slot configuration if the cluster slot is being changed by the benchmark (e.g. auto-scaling).

ranshid: >IMO it would be good to fetch the cluster slot configuration if the cluster slot is being changed by the benchmark (e.g. auto-scaling).

I do not think that is realy bad, specifically  for MOVED it is probably wanted, for asking, since we will not make use of it maybe it will just increase the overhead. but we can keep it for now since this was the state before as well.

src/valkey-benchmark.c
ranshid: > Therefore, there were three directions for implementation, and I adopted method 1, which is similar to the real world I've seen.
> 
> 1/ Read from the replica option only reads from the replica, and if you need to read from the primary as well, run a separate process (current implementation)
> 2/ Read from the replica option reads from the entire cluster, including the primary (the way you described)
> 3/ Separate options depending on whether to include the primary

I agree that most client will mainly target replicas for reads but they usually also provide configuration to control that which is why I support option 3. 

> in addition, if we include the primary, we need to think about the handling write command.
> e.g. valkey-benchmark -h cluster_host -c 4 -t get,set -n 1000 --cluster --rfr

Idealy it would be that the benchmark will also take the command docs of the specific commands and see if it is read or write and act accordingly. not an impossible mission but we can do it at a later phase.

bluayer: Great. I'll implement the following tasks in relation to this discussion:
1/ Provide separate options for reading from replicas only and for reading from the entire cluster. (Option 3)

2/ Add CLI description about the option could not work properly when executing with write requests or commands.

And in the next PR/phase, I'll implement a way to distribute writes and reads well at the cluster level!

Btw, thank you for your advices and opinions ☺️

bluayer:
@ranshid Thanks for your review and suggestion. I fixed the lines based on your review and suggestion. Please take a look the changes.

ranshid:
> @ranshid Thanks for your review and suggestion. I fixed the lines based on your review and suggestion. Please take a look the changes.

Thank you @bluayer I will try to take a look today but might not get to it and I will be on vacation till Monday. So sorry in advance if this will delay to next week.

bluayer:
Of course, I completely understand you because it's the holiday season! Don't worry about being postponed and have a great vacation ☺️

bluayer:
> Thank you @bluayer I will try to take a look today but might not get to it and I will be on vacation till Monday. So sorry in advance if this will delay to next week.

Thank you @ranshid for telling me that. Of course, I completely understand you because it's the holiday season! Don't worry about being postponed and have a great vacation ☺️

src/valkey-benchmark.c
```diff
@@ -1460,6 +1402,10 @@ int parseOptions(int argc, char **argv) {
                 config.num_threads = 0;
         } else if (!strcmp(argv[i], "--cluster")) {
             config.cluster_mode = 1;
+        } else if (!strcmp(argv[i], "--rfa")) {
+            config.read_from_all = 1;
+        } else if (!strcmp(argv[i], "--rfro")) {
+            config.read_from_replicas_only = 1;
```
ranshid: I would suggest unify this as a single config option --rfr with optional values: 'yes', 'no' (default) and 'any'
this might simplify the configuration. Also I would suggest using enum in order to manage the  'config.read_from_replicas' for the same option.

bluayer: Is the method you're suggesting similar to the following code?

```c
typedef enum readFromReplicas {
    FROM_PRIMARY_ONLY = 0, // default
    FROM_REPLICAS_ONLY,
    FRON_ALL
} readFromReplicas;
...

config.read_from_replicas = FROM_PRIMARY_ONLY;

...

else if (!strcmp(argv[i], "--rfr")) {
    config.read_from_replicas = FROM_REPLICAS_ONLY;
    if (argv[i +1] && atoi(argv[++i]) == 2) { config.read_from_replicas = FROM_ALL; }
}```

Comment (ranshid) [reply src/valkey-benchmark.c]: Yes. this is also more aligned with how clients usually configure RFR AFAIK

Comment (bluayer) [reply src/valkey-benchmark.c]: I adopted your feedback :)

File: src/valkey-benchmark.c [hunk 7]
```diff
@@ -1557,6 +1509,12 @@ int parseOptions(int argc, char **argv) {
         "                    If the command is supplied on the command line in cluster\n"
         "                    mode, the key must contain \"{tag}\". Otherwise, the\n"
         "                    command will not be sent to the right cluster node.\n"
+        " --rfr <mode>       Enable read from replicas in cluster mode.\n"
+        "                    This command must be used with the --cluster option.\n"
+        "                    When using this option, it is recommended to use only \n"
+        "                    the commands for read requests.\n"
+        "                    default=read from replica only\n"
+        "                    2=read from all nodes(primary and replica)\n"
```
ranshid: Lets fix this help section:
1. I think the command should accept string values eg 'primaries' (default), 'replicas', 'all' 
2. Lets make sure we socument all possible values here 
3. IIUC the default is read from primaries only right?

src/valkey-benchmark.c
```diff
@@ -1460,6 +1407,11 @@ int parseOptions(int argc, char **argv) {
                 config.num_threads = 0;
         } else if (!strcmp(argv[i], "--cluster")) {
             config.cluster_mode = 1;
+        } else if (!strcmp(argv[i], "--rfr")) {
+            config.read_from_replica = FROM_REPLICA_ONLY;
+            if (argv[i + 1] && atoi(argv[++i]) == 2) {
+                config.read_from_replica = FROM_ALL;
+            }
```
ranshid: Not sure if we should not enable accepting all possible values here. I agree it might be strange to configure rfr as 'no' or 'primaries' but it makes more sense to disable it but still provide the configuration. 
We can alternatively allow integer mode values but we should document them well in the help section.

Also this way the command limne is very error prone. it is better to always request a value tmatching to the --rfr and validate it is in the correct format (in case of integer check the range and in case string validate is matches the allowed values)

ranshid (commented):
Just started to play with the benchmark before approving and I noticed some issues

src/valkey-benchmark.c
```diff
@@ -835,7 +855,15 @@ static void showLatencyReport(void) {
         printf("  %d bytes payload\n", config.datasize);
         printf("  keep alive: %d\n", config.keepalive);
         if (config.cluster_mode) {
-            printf("  cluster mode: yes (%d primaries)\n", config.cluster_node_count);
+            const char *node_prefix = NULL;
+            if (config.read_from_replica == FROM_ALL) {
+                node_prefix = "all";
```
ranshid: The 'all' name here does not reflect the same semntics as was done before. for example when I use all I will get something like:
```
Cluster has 6 all nodes:

all 0: fcc9d096ad7dfa9f365c917941201a5bf037eb20 127.0.0.1:7000
all 1: 7cd62de6bb802699dad0c33961b2542c88afbeca 127.0.0.1:7004
all 2: 7330c7461582a447520745518f2d3a53ea2a5b0c 127.0.0.1:7001
all 3: f8b7e3188b0ec5bc6ea1a52342733dbb3dde2680 127.0.0.1:7005
all 4: 1a70c15fad7fc590092bce033dbc09b0cc057a86 127.0.0.1:7002
all 5: 4b1228f49237178f92583d72ec1c6d06e452eaed 127.0.0.1:7003
```

which does not sound right. we should probably reflect the correct node role taken after parsing the cluster nodes.

src/valkey-benchmark.c
bluayer: Good comments.
1. I'll fix it to string value such as "replicas" and "all". It's weird that “primaries” is the default, as you mentioned below, and I'd rather use `goto invalid`.
2. I agree.
3. If the user run it with `--cluster` only, the default is reading from primaries only. If the user run with `--cluster --rfr`, the default is reading from replicas only.

src/valkey-benchmark.c
bluayer: I can't agree with you more. I'll remove integer mode and implement string mode with validation function.
And If the user enter 'no' or 'primaries' with `--rfr`, it's more appropriate to disable it.

src/valkey-benchmark.c
```diff
@@ -1557,6 +1523,17 @@ int parseOptions(int argc, char **argv) {
...
+        "                    This command must be used with the --cluster option.\n"
+        "                    When using this option, it is recommended to use only \n"
+        "                    the commands for read requests.\n"
+        "                    There are two mode for reading from replicas.\n"
+        "                    If nothing is entered or 'replicas' is entered, this\n"
+        "                    sends read requests to replicas only (default) \n"
+        "                    If 'all' is entered, this sends read requests to all nodes,\n"
+        "                    including primaries and replicas. \n"
+        "                    If you only want to request the primaries, use --cluster option only.\n"
+        "                    2=read from all nodes(primary and replica)\n"
```
ranshid:
```suggestion
        " --rfr <mode>       Enable read from replicas in cluster mode.\n"
        "                    This command must be used with the --cluster option.\n"
        "                    There are three modes for reading from replicas:\n"
        "                    'no' - sends read requests to primaries only (default) \n"
        "                    'yes' - sends read requests to replicas only.\n"
        "                    'all' - sends read requests to all nodes.\n"
        "                    Since write commands will not be accepted by replicas,\n"
        "                    it is recommended to enable read from replicas only for read command tests.\n"
```

src/valkey-benchmark.c
```diff
@@ -1460,6 +1413,19 @@ int parseOptions(int argc, char **argv) {
...
+                if (!strcmp(argv[i+1], "all")) {
+                    config.read_from_replica = FROM_ALL;
+                    i++;
+                } else if (!strcmp(argv[i+1], "replicas")) {
+                    config.read_from_replica = FROM_REPLICA_ONLY;
+                    i++;
+                } else if (strlen(argv[i+1]) > 0 && argv[i+1][0] != '-') {
+                    goto invalid;
+                }
+            }
```
ranshid:
```suggestion
        } else if (!strcmp(argv[i], "--rfr")) {
            if (argv[++i]) {
                if (!strcmp(argv[i], "all")) {
                    config.read_from_replica = FROM_ALL;
                } else if (!strcmp(argv[i], "yes")) {
                    config.read_from_replica = FROM_REPLICA_ONLY;
                } else if (!strcmp(argv[i], "no")) {
                    config.read_from_replica = FROM_PRIMARY_ONLY;
                } else {
                    goto invalid;
                }
            } else
                goto invalid;
```

src/valkey-benchmark.c
```diff
@@ -1742,15 +1720,24 @@ int main(int argc, char **argv) {
...
-        printf("Cluster has %d primary nodes:\n\n", config.cluster_node_count);
+        const char *node_prefix = NULL;
+        if (config.read_from_replica == FROM_ALL) {
+            node_prefix = "primaries and replicas";
+        } else if (config.read_from_replica == FROM_REPLICA_ONLY) {
+            node_prefix = "replica";
+        } else {
+            node_prefix = "primary";
+        }
+        printf("Cluster has %d %s nodes:\n\n", config.cluster_node_count, node_prefix);
```
ranshid:
```suggestion
        const char *node_roles = NULL;
        if (config.read_from_replica == FROM_ALL) {
            node_roles = "cluster";
        } else if (config.read_from_replica == FROM_REPLICA_ONLY) {
            node_roles = "replica";
        } else {
            node_roles = "primary";
        }
        printf("Cluster has %d %s nodes:\n\n", config.cluster_node_count, node_roles);
```

ranshid (commented):
some minor suggestions

src/valkey-benchmark.c
bluayer: In this code, If we have other options after '--rfr', the other option can't execute, right?

e.g. `--rfr -r 10`

ranshid: exactly. I do not think we should support --rfr with no value following it

ranshid: it makes it clearer IMO

bluayer: 1 more question : Do we need the lines at 1427 ~ 1428?
This seems to mean that if no mode is specified after the `--rfr` option, it will fail to execute. Is it intended that --rfr must be used with mode?

e.g.
`--rfr` : Failed
`--rfr no`: Success

ranshid: IMO yes. having the same option which can either accept a value or not is something I rarely seen. I think it is clearer that a value is either expected or not.

src/valkey-benchmark.c
```diff
@@ -1235,17 +1179,12 @@ static int fetchClusterSlotsConfiguration(client c) {
     if (is_fetching_slots) return -1; // TODO: use other codes || errno ?
     atomic_store_explicit(&config.is_fetching_slots, 1, memory_order_relaxed);
     fprintf(stderr, "WARNING: Cluster slots configuration changed, fetching new one...\n");
+    fprintf(stderr, "If you are using the --rfr option and sending write requests (set type commands),\nthe requests could not be processed properly.\n");
+
```
ranshid: I would suggest avoid make this printed for each error. I would think it might be nivce to have it printed once in case the --rfr option is used with something other than pthe default.

```suggestion
```

bluayer: I think It's unnecessary now. When I put this line at first, the description or operation of the options was not clear. So  It was necessary at that time, but it's unnecessary now because everything is clear.

ranshid (approved):
LGTM

ranshid:
@bluayer all looks good IMO. Thank you for taking the time to introduce and refactor!

bluayer:
Hello, @madolson !
My customers are asking for this feature when they test their workload and consider switching from Redis to Valkey in these days.
Could you please discuss including the feature in the next minor version?

PR #1395, opened 2024-12-05
Fix coredump when use hellodict example module

Codebells:
image

 In the ValkeyModule_OnLoad method of the file hellodict.c, the parameter keystep of ValkeyModule_CreateCommand should be 1. Otherwise, execute command will coredump.
```
MODULE LOAD /home/tiger/valkey/src/modules/hellodict.so
COMMAND GETKEYS HELLODICT.SET key value
```


zuiderkwast (approved):
Thanks!

We obviously we don't have any tests for these example modules.

We should probably add tests for them if we want to keep them in the repo.

Codebells:
Agree, I will add tests for them in next MR.

Issue #668, opened 2024-06-19
Extended Valkey client introspection functionality 


madolson:
We want to standardize the way that admin users can describe and interact with the client commands. We are proposing to generalize the filters introduced in CLIENT KILL so that it also works to count and list clients as well.

The following commands could work with the same set of filters.

* CLIENT COUNT [filters...] *(New command suggested)*
* CLIENT KILL [filters...]
* CLIENT LIST [filters...]

The supported filters are to be decided. Suggestions, including already existing filters:

* TYPE [NORMAL|MASTER|PRIMARY|SLAVE|REPLICA|PUBSUB] *(existing)*
* USER username *(#1401)*
* ADDR ip:port *(#1401)*
* LADDR ip:port *(#1401)*
* SKIPME [yes|no] *(#1401)*
* MAXAGE time *(#1401)*
* ID client-id *(existing)*
* ID client-id [client-id ...] *(#1401)* ~~for compatibility reasons with client list, client-id must be specified last~~.
* PATTERN pattern the client must subscribe to *(Needed?)*
* CHANNEL channel the client must subscribe to *(Needed?)*
* SHARD-CHANNEL `<shard channel the client must subscribe to>` *(Needed?)*
* FLAGS Client must include this flag, this option can be repeated, or the argument can be a string of multiple flags. Example: `CLIENT LIST FLAGS "bPt"`.
* NAME client name
* IDLE time (or ~~MIN-IDLE~~) minimum idle time of the client.

More filter suggestions that came up during discussions:

* IP or CLIENT-IP, or allow overloaded `ADDR ip`. (The existing filter syntax `ADDR ip:port` is not very useful, because the client's port is ephemeral and will change for each connection.)
* LIB-NAME
* LIB-VER
* DB number
* TOT-NET-IN bytes
* TOT-NET-OUT bytes

In Redis we also discussed adding a new command, `CLIENT DESCRIBE`, will also introduce the notion of **selectors**. Selectors allow you to scope down the fields you are interested in.

```
CLIENT DESCRIBE SELECT ID ADDR FILTER MIN-AGE 100 
```
- [ ] Extend filters for CLIENT KILL and LIST command
- [ ] Support negative filtering: https://github.com/valkey-io/valkey/issues/1936
- [ ] Support client count  [filters...] *(New command suggested)*

madolson:
Originally reference from https://github.com/redis/redis/issues/12311.

hpatro:
[...]

PR #1398, opened 2024-12-06
Add new flag in `CLIENT LIST` for import-source client

### LINKED ISSUE #1350: Fail `CLIENT IMPORT-SOURCE` When Import Mode is Disabled and Reflect Client Import Source State in `CLIENT LIST`
I think we should clean it up though. Also, shall we update `CLIENT LIST` output to reflect the import source mode?

_Originally posted by @hpatro in https://github.com/valkey-io/valkey/pull/1185#discussion_r1851049362_

lyq2333:
Fixes #1350 and https://github.com/valkey-io/valkey/pull/1185#discussion_r1851049362.

- Add new flag "I" in `CLIENT LIST` for import-source client
- Add `DEBUG_CONFIG` for import-mode
- Allow import-source status to be turned off when import-mode is off

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -3209,7 +3209,7 @@ standardConfig static_configs[] = {
     createBoolConfig("enable-debug-assert", NULL, IMMUTABLE_CONFIG | HIDDEN_CONFIG, server.enable_debug_assert, 0, NULL, NULL),
     createBoolConfig("cluster-slot-stats-enabled", NULL, MODIFIABLE_CONFIG, server.cluster_slot_stats_enabled, 0, NULL, NULL),
     createBoolConfig("hide-user-data-from-log", NULL, MODIFIABLE_CONFIG, server.hide_user_data_from_log, 1, NULL, NULL),
-    createBoolConfig("import-mode", NULL, MODIFIABLE_CONFIG, server.import_mode, 0, NULL, NULL),
+    createBoolConfig("import-mode", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.import_mode, 0, NULL, NULL),
 
     /* String Configs */
     createStringConfig("aclfile", NULL, IMMUTABLE_CONFIG, ALLOW_EMPTY_STRING, server.acl_filename, "", NULL, NULL),
--- a/src/networking.c
+++ b/src/networking.c
@@ -4103,7 +4103,7 @@ void clientCommand(client *c) {
     } else if (!strcasecmp(c->argv[1]->ptr, "import-source")) {
         /* CLIENT IMPORT-SOURCE ON|OFF */
         if (!server.import_mode && strcasecmp(c->argv[2]->ptr, "off")) {
-            addReplyError(c, "Server is not in import mode and we only allow to close import-source status.");
+            addReplyError(c, "Server is not in import mode");
             return;
         }
         if (!strcasecmp(c->argv[2]->ptr, "on")) {
--- a/src/networking.c
+++ b/src/networking.c
@@ -3340,6 +3340,7 @@ sds catClientInfoString(sds s, client *client, int hide_user_data) {
     if (client->flag.readonly) *p++ = 'r';
     if (client->flag.no_evict) *p++ = 'e';
     if (client->flag.no_touch) *p++ = 'T';
+    if (client->flag.import_source) *p++ = 'I';
     if (p == flags) *p++ = 'N';
     *p++ = '\0';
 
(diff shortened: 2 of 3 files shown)


src/networking.c
```diff
@@ -4101,8 +4102,8 @@ void clientCommand(client *c) {
         addReply(c, shared.ok);
     } else if (!strcasecmp(c->argv[1]->ptr, "import-source")) {
         /* CLIENT IMPORT-SOURCE ON|OFF */
-        if (!server.import_mode) {
-            addReplyError(c, "Server is not in import mode");
+        if (!server.import_mode && strcasecmp(c->argv[2]->ptr, "off")) {
+            addReplyError(c, "Server is not in import mode and we only allow to close import-source status.");
```
zuiderkwast: I don't think we need the extra explanation.

```suggestion
            addReplyError(c, "Server is not in import mode");
```

src/config.c
```diff
@@ -3209,7 +3209,7 @@ standardConfig static_configs[] = {
     createBoolConfig("enable-debug-assert", NULL, IMMUTABLE_CONFIG | HIDDEN_CONFIG, server.enable_debug_assert, 0, NULL, NULL),
     createBoolConfig("cluster-slot-stats-enabled", NULL, MODIFIABLE_CONFIG, server.cluster_slot_stats_enabled, 0, NULL, NULL),
     createBoolConfig("hide-user-data-from-log", NULL, MODIFIABLE_CONFIG, server.hide_user_data_from_log, 1, NULL, NULL),
-    createBoolConfig("import-mode", NULL, MODIFIABLE_CONFIG, server.import_mode, 0, NULL, NULL),
+    createBoolConfig("import-mode", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.import_mode, 0, NULL, NULL),
```
zuiderkwast, line 3212: How is this config useful for debugging?

tests/unit/expire.tcl
```diff
@@ -833,6 +833,24 @@ start_server {tags {"expire"}} {
         assert_equal [r debug set-active-expire 1] {OK}
     } {} {needs:debug}
 
+    test {import-source can be closed when import-mode is off} {
+        r flushall
+        set id [r client id]
+
+        catch {r client import-source on} err
+        assert_match {*Server is not in import mode and we only allow to close import-source status*} $err
```
zuiderkwast: We shouldn't match on the exact error messages. Clients are not supposed to do this either, because it can change. The first word is a code that indicates the class of error and this is what clients should look at.

```suggestion
        assert_match {ERR*} $err
```

zuiderkwast (commented):
Looks good in general. Thanks!

zuiderkwast:
> Should I add a new option for `CLIENT KILL`, i.e. `CLIENT KILL FLAG`?

It's a good idea, but let's move this idea to #668.

src/config.c
lyq2333: The reason is in https://github.com/valkey-io/valkey/issues/1350#issuecomment-2496938879.

lyq2333:
> > Should I add a new option for `CLIENT KILL`, i.e. `CLIENT KILL FLAG`?
> 
> It's a good idea, but let's move this idea to #668.

Thanks for your review and suggestion. I like this idea 😄

zuiderkwast (approved):
LGTM, thanks!

@valkey-io/core-team please approve the interface changes mentioned in the top comment.

tests/unit/expire.tcl
```diff
@@ -833,6 +833,24 @@ start_server {tags {"expire"}} {
...
+        r config set import-mode yes
+        assert_equal [r client import-source on] {OK}
+
+        assert_match {*flags=I*} [r client list id $id]
+
+        r config set import-mode no
+        assert_equal [r client import-source off] {OK}
+
+        assert_match {*flags=N*} [r client list id $id]
+    }
```
enjoy-binbin: we can save some lines since it is a very simple test 
```suggestion
    test {import-source can be closed when import-mode is off} {
        r config set import-mode no
        assert_error "ERR Server is not in import mode" {r client import-source on}

        r config set import-mode yes
        assert_equal [r client import-source on] {OK}
        assert_match {*flags=I*} [r client list id [r client id]]

        r config set import-mode no
        assert_equal [r client import-source off] {OK}
        assert_match {*flags=N*} [r client list id [r client id]]
    }
```

enjoy-binbin (approved):

tests/unit/expire.tcl
zuiderkwast: Good point, we can skip `flushall`.

I suggested earlier to match the error only on `ERR*` instead of matching on the full error message. IMO we don't need to match the full message because we can allow it to change.

enjoy-binbin: i think matching it is no harm. We can always change the text and fix the test. Or at least match a few keyword since we do have a syntax error and we can know it is on the test path.

zuiderkwast: OK, good point that we want to check that it's not a syntax error. It's unfortunate that most of the errors are just ERR. We can't change it now, but it would have been better if each error had its own code, like SYNTAX for syntax error, etc.

soloestoy (approved):

madolson (approved):

zuiderkwast:
@lyq2333 Can you open a doc PR to add the flag here: https://github.com/valkey-io/valkey-doc/blob/main/commands/client-list.md

lyq2333:
> @lyq2333 Can you open a doc PR to add the flag here: https://github.com/valkey-io/valkey-doc/blob/main/commands/client-list.md

OK, I will include it in the introduction of import mode.

PR #1401, opened 2024-12-06
Adding Missing filters to CLIENT LIST and Dedup Parsing

sarthakaggarwal97:
Adds filter options to CLIENT LIST:

    * USER <username>
      Return clients authenticated by <username>.
    * ADDR <ip:port>
      Return clients connected from the specified address.
    * LADDR <ip:port>
      Return clients connected to the specified local address.
    * SKIPME (YES|NO)
      Exclude the current client from the list (default: no).
    * MAXAGE <maxage>
      Only list connections older than the specified age.

Modifies the ID filter to CLIENT KILL to allow multiple IDs

    * ID <client-id> [<client-id>...]
      Kill connections by client ids.

This makes CLIENT LIST and CLIENT KILL accept the same options.

For backward compatibility, the default value for SKIPME is NO for CLIENT LIST and YES for CLIENT KILL.

The MAXAGE comes from CLIENT KILL, where it *keeps* clients with the given max age and kills the older ones. This logic becomes weird for CLIENT LIST, but is kept for similary with CLIENT KILL, for the use case of first testing manually using CLIENT LIST, and then running CLIENT KILL with the same filters.

The `ID client-id [client-id ...]` no longer needs to be the last filter. The parsing logic determines if an argument is an ID or not based on whether it can be parsed as an integer or not.

Partly addresses: #668


src/server.h
```diff
@@ -1386,6 +1386,16 @@ typedef struct client {
         net_output_bytes_curr_cmd; /* Total network output bytes sent to this client, by the current command. */
 } client;
 
+typedef struct {
+    uint64_t id;
```
zuiderkwast: A filter should be able to have a list of IDs, as in `CLIENT LIST ID 3 4 5`. We need some list here, or maybe an `intset`? (See intset.c)

We need some comments too. Does 0 and NULL for these fields mean match any client? It's worth mentioning in a comment.

This struct and the new functions can probably be just local in networking.c. No other files need to use these.

src/networking.c
```diff
@@ -3438,6 +3438,28 @@ sds getAllClientsInfoString(int type, int hide_user_data) {
     return o;
 }
 
+sds getAllFilteredClientsInfoString(clientFilter *client_filter, int hide_user_data) {
+    listNode *ln;
+    listIter li;
+    client *client;
+    sds o = sdsnewlen(SDS_NOINIT, 200 * listLength(server.clients));
```
zuiderkwast: The number of connected clients can be thousands, so maybe we shouldn't allocate this huge number. I suggest we initialize it to a fixed 500 bytes or so.

src/networking.c
```diff
@@ -3438,6 +3438,28 @@ sds getAllClientsInfoString(int type, int hide_user_data) {
...
+    listRewind(server.clients, &li);
+    while ((ln = listNext(&li)) != NULL) {
+        client = listNodeValue(ln);
+        if (client_filter->addr && strcmp(getClientPeerId(client), client_filter->addr) != 0) continue;
+        if (client_filter->laddr && strcmp(getClientSockname(client), client_filter->laddr) != 0) continue;
+        if (client_filter->type != -1 && getClientType(client) != client_filter->type) continue;
+        if (client_filter->id != 0 && client->id != client_filter->id) continue;
+        if (client_filter->user && client->user != client_filter->user) continue;
+        if (client_filter->skipme) continue;
+        if (client_filter->max_age != 0 && (long long)(commandTimeSnapshot() / 1000 - client->ctime) < client_filter->max_age) continue;
```
zuiderkwast: We should have a function just for checking if a client matches a filter and returns 1 or 0. Then it can be used in CLIENT LIST and CLIENT KILL.

```C
int isClientMatchingFilter(client *c, clientFilter *client_filter);
```

zuiderkwast (commented):
Thanks.

I just took a quick look and I can see some things that appear to be broken:

* CLIENT LIST with multiple IDs, for example `CLIENT LIST ID 3 4 5`. (It means id is *3 or 4 or 5*.) We a test case for this.
* The default for "skipme" for CLIENT KILL should be YES by default (when filters are used), but for CLIENT LIST, skipme should be NO by default. Otherwise, we have a breaking change. Please add some tests without SKIPME but with other filters matching the calling client to cover the default skipme value for LIST and KILL.


We need some test cases combining multiple filter rules at the same time.

sarthakaggarwal97:
Thanks @zuiderkwast for the comments! I am working on them

src/networking.c
sarthakaggarwal97: yeah, I was considering the same. The difference is KILL is not allowed to close the origin client connection, but can take that condition out separately.

zuiderkwast: Exactly, it is better that the function just checks for a match, instead of creating info or killing the user, so it can be used by both commands.

src/networking.c
```diff
@@ -3780,13 +3790,13 @@ void clientCommand(client *c) {
...
+            if (client_filter.addr && strcmp(getClientPeerId(client), client_filter.addr) != 0) continue;
+            if (client_filter.laddr && strcmp(getClientSockname(client), client_filter.laddr) != 0) continue;
+            if (client_filter.type != -1 && getClientType(client) != client_filter.type) continue;
+            if (client_filter.id != 0 && client->id != client_filter.id) continue;
+            if (client_filter.user && client->user != client_filter.user) continue;
+            if (c == client && client_filter.skipme) continue;
+            if (client_filter.max_age != 0 && (long long)(commandTimeSnapshot() / 1000 - client->ctime) < client_filter.max_age) continue;
 
             /* Kill it. */
             if (c == client) {
```
zuiderkwast: I mean, here, use the shared code for checking if it's a match, something like

```suggestion
            if (!clientMatchesFilter(c, client_filter) continue;

            /* Kill it. */
            if (c == client) {
```

src/networking.c
```diff
@@ -3701,75 +3767,19 @@ void clientCommand(client *c) {
...
         /* CLIENT KILL <ip:port>
          * CLIENT KILL <option> [value] ... <option> [value] */
-        char *addr = NULL;
-        char *laddr = NULL;
-        user *user = NULL;
-        int type = -1;
-        uint64_t id = 0;
-        long long max_age = 0;
-        int skipme = 1;
+        clientFilter client_filter = {0, 0, NULL, NULL, NULL, -1, 1};
```
zuiderkwast: Can you write the struct with the field names, so it's easier to read it?

```suggestion
        clientFilter client_filter = {
            .type = -1,
            .skipme = 1,
        };
```

src/networking.c
zuiderkwast: > The difference is KILL is not allowed to close the origin client connection, but can take that condition out separately.

I'm not sure I understand what you mean, but I added a comment about the CLIENT KILL code just to be clear.

sarthakaggarwal97: Yes, we are on the same page. Thank you!

sarthakaggarwal97:
@zuiderkwast thank you for your comments. Adressed most of them. Working on adding tests meanwhile. Trying to see how can I have tests with multiple clients.

Sharing some responses from local testing

Client Kill Command
```
127.0.0.1:6379> client list
id=3 addr=127.0.0.1:63139 laddr=127.0.0.1:6379 fd=8 name= age=11 idle=0 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=10 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1946 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=53 tot-net-out=209291 tot-cmds=1
id=4 addr=127.0.0.1:63141 laddr=127.0.0.1:6379 fd=9 name= age=8 idle=8 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
id=5 addr=127.0.0.1:63142 laddr=127.0.0.1:6379 fd=10 name= age=5 idle=5 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
127.0.0.1:6379> client info
id=3 addr=127.0.0.1:63139 laddr=127.0.0.1:6379 fd=8 name= age=16 idle=0 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=10 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1946 events=r cmd=client|info user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=79 tot-net-out=210285 tot-cmds=2
127.0.0.1:6379> client kill id 4
(integer) 1
127.0.0.1:6379> client list
id=3 addr=127.0.0.1:63139 laddr=127.0.0.1:6379 fd=8 name= age=39 idle=0 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=10 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1946 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=146 tot-net-out=210626 tot-cmds=4
id=5 addr=127.0.0.1:63142 laddr=127.0.0.1:6379 fd=10 name= age=33 idle=33 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
127.0.0.1:6379> client list
id=3 addr=127.0.0.1:63139 laddr=127.0.0.1:6379 fd=8 name= age=50 idle=0 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=10 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1946 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=172 tot-net-out=211295 tot-cmds=5
id=5 addr=127.0.0.1:63142 laddr=127.0.0.1:6379 fd=10 name= age=44 idle=44 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
id=7 addr=127.0.0.1:63168 laddr=127.0.0.1:6379 fd=9 name= age=3 idle=3 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
127.0.0.1:6379> client kill id 5 7 
(integer) 2
127.0.0.1:6379> client list
id=3 addr=127.0.0.1:63139 laddr=127.0.0.1:6379 fd=8 name= age=72 idle=0 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=10 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1946 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=246 tot-net-out=212296 tot-cmds=7
id=8 addr=127.0.0.1:63183 laddr=127.0.0.1:6379 fd=9 name= age=7 idle=7 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
id=9 addr=127.0.0.1:63184 laddr=127.0.0.1:6379 fd=10 name= age=4 idle=4 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
127.0.0.1:6379> client kill id 3 9 addr 127.0.0.1:63183
(integer) 0
127.0.0.1:6379> client list
id=3 addr=127.0.0.1:63139 laddr=127.0.0.1:6379 fd=8 name= age=97 idle=0 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=10 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1946 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=352 tot-net-out=213295 tot-cmds=9
id=8 addr=127.0.0.1:63183 laddr=127.0.0.1:6379 fd=9 name= age=32 idle=32 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
id=9 addr=127.0.0.1:63184 laddr=127.0.0.1:6379 fd=10 name= age=29 idle=29 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
127.0.0.1:6379> client kill id 3 9 addr 127.0.0.1:63184
(integer) 1
127.0.0.1:6379> client list
id=3 addr=127.0.0.1:63139 laddr=127.0.0.1:6379 fd=8 name= age=107 idle=0 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=10 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1946 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=458 tot-net-out=214298 tot-cmds=11
id=8 addr=127.0.0.1:63183 laddr=127.0.0.1:6379 fd=9 name= age=42 idle=42 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1

```

Client List Command

```
127.0.0.1:6379> client list
id=3 addr=127.0.0.1:63139 laddr=127.0.0.1:6379 fd=8 name= age=1361 idle=8 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=484 tot-net-out=215973 tot-cmds=13
id=8 addr=127.0.0.1:63183 laddr=127.0.0.1:6379 fd=9 name= age=1296 idle=1296 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
id=10 addr=127.0.0.1:55765 laddr=127.0.0.1:6379 fd=10 name= age=10 idle=0 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=10 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1946 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=53 tot-net-out=209291 tot-cmds=1
127.0.0.1:6379> client list id 3 8
id=3 addr=127.0.0.1:63139 laddr=127.0.0.1:6379 fd=8 name= age=1368 idle=15 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=484 tot-net-out=215973 tot-cmds=13
id=8 addr=127.0.0.1:63183 laddr=127.0.0.1:6379 fd=9 name= age=1303 idle=1303 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
127.0.0.1:6379> client list id 3 8 age 1300
ERR syntax error
127.0.0.1:6379> client list id 3 8 maxage 1300
id=3 addr=127.0.0.1:63139 laddr=127.0.0.1:6379 fd=8 name= age=1387 idle=34 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=484 tot-net-out=215973 tot-cmds=13
id=8 addr=127.0.0.1:63183 laddr=127.0.0.1:6379 fd=9 name= age=1322 idle=1322 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
127.0.0.1:6379> client list id 3 8 maxage 1500
127.0.0.1:6379> client list id 3 8 addr 127.0.0.1:63183
id=8 addr=127.0.0.1:63183 laddr=127.0.0.1:6379 fd=9 name= age=1349 idle=1349 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1

```

src/networking.c
```diff
@@ -41,11 +41,40 @@
...
 #include <ctype.h>
 #include <stdatomic.h>
+#include "intset.h"
+#include <stdbool.h>
+
+/**
+ * This struct is used to encapsulate filtering criteria for operations on clients
+ * such as identifying specific clients to kill or retrieve. Each field in the struct
+ * represents a filter that can be applied based on specific attributes of a client.
+ */
```
zuiderkwast: Good, but we don't use doxygen style comments. We put the start and end of comment on the same line. This is just to follow existing style.

This applies to all `/**`. I'll not comment in all places.

```suggestion
/* This struct is used to encapsulate filtering criteria for operations on clients
 * such as identifying specific clients to kill or retrieve. Each field in the struct
 * represents a filter that can be applied based on specific attributes of a client. */
```

src/networking.c
```diff
@@ -41,11 +41,40 @@
...
+typedef struct {
+    /** A set of client IDs to filter. If NULL, no ID filtering is applied. */
+    intset *ids;
+    /** Maximum age (in seconds) of a client connection for filtering.
+     * Connections younger than this value will not match.
+     * A value of 0 means no age filtering. */
```
zuiderkwast: Is this right? Isn't the maximum age something that excludes the *oldest* clients?

src/networking.c
```diff
@@ -3426,7 +3455,7 @@ sds getAllClientsInfoString(int type, int hide_user_data) {
     listNode *ln;
     listIter li;
     client *client;
-    sds o = sdsnewlen(SDS_NOINIT, 200 * listLength(server.clients));
+    sds o = sdsnewlen(SDS_NOINIT, 500);
```
zuiderkwast: Ah, this is where you got 200 * listLength. This function is creates client info for all clients though, without filtering, so here we probably need around 200 bytes per line.

Did you change the wrong line?

src/networking.c
```diff
@@ -3557,6 +3602,96 @@ void quitCommand(client *c) {
...
+            if (filter->ids == NULL) {
+                filter->ids = intsetNew(); // Initialize the intset for IDs
+            }
+            i++; // Move to the first ID after "ID"
+
+            // Process all IDs until a non-numeric argument or end of args
+            while (i < c->argc) {
+                long long id;
+                if (!string2ll(c->argv[i]->ptr, sdslen(c->argv[i]->ptr), &id)) {
+                    break; // Stop processing IDs if a non-numeric argument is encountered
```
zuiderkwast: I guess we can do like this, but if we want to keep the existing behavior, this should probably be a syntax error. Currently, ID must to be the last filter on the command line. That's what the docs say and it was also mentioned in the issue.

@madolson Do you think we should allow ID not being the last filter in the command line? We can actually distinguish a numeric id from a non-numeric new filter token, if we want.

src/networking.c
```diff
@@ -41,11 +41,40 @@
 #include <math.h>
 #include <ctype.h>
 #include <stdatomic.h>
+#include "intset.h"
+#include <stdbool.h>
```
zuiderkwast: Please put all local files includes `#include "..."` above all the standard lib `#include <...>` ones.

So far, we haven't used stdbool (bool, true, false) in this project, so we should at least have a discussion with TSC before we start allowing it. I think we should use it, but @madolson are you still skeptical?

src/networking.c
```diff
@@ -3810,6 +3888,7 @@ void clientCommand(client *c) {
         /* If this client has to be closed, flag it as CLOSE_AFTER_REPLY
          * only after we queued the reply to its output buffers. */
         if (close_this_client) c->flag.close_after_reply = 1;
+        zfree(client_filter.ids);
```
zuiderkwast: I see this zfree on error in multiple places. We should refactor it so we only need it in one place.

One way to do this is to use a goto label like `client_kill_done` just before this last zfree in the end and on error we use goto instead of zfree + return.

```suggestion
client_kill_done:
        zfree(client_filter.ids);
```

We can also consider splitting it into smaller functions, if it helps. This function is very large.

zuiderkwast (commented):
> Trying to see how can I have tests with multiple clients.

You can create multiple clients.

```
        set c1 [valkey_client]
        set c2 [valkey_client]
        set c3 [valkey_client]
        ...
        $c1 close
        $c2 close
        $c3 close
```

src/networking.c
zuiderkwast: Ah, I'm wrong, sorry. I understand "younger than this value will not match" means they will not be killed. The docs for CLIENT KILL says "Closes all the connections that are older than the specified age, in seconds".

So the value is the max age for clients to *not match*, which makes sense for CLIENT KILL but maybe very weird for CLIENT LIST?

Do we want `CLIENT LIST MAXAGE 1000` to mean "list all clients older than 1000"? It sounds illogical, but OTOH, I guess people will use CLIENT LIST first to list some clients and then CLIENT KILL with the same filters to kill them, so for this use case, the logic needs to be the same. :thinking:

Can we keep this behavior but come up with a good formulation for documenting this in CLIENT LIST?

zuiderkwast: The safest solution is maybe to skip MAXAGE for CLIENT LIST. To do that, we can add a flag like `allow_maxage` for `parseClientFilters`. When `allow_maxage` is false (for CLIENT LIST), it will be a syntax error. Makes sense?

src/networking.c
zuiderkwast: Good. Just FYI, it's possible to skip fields that are 0 or NULL, because they are zeroed by default with this syntax.

src/networking.c
sarthakaggarwal97: yeah I did :p 
Thanks, will be fixed in the next commit.

src/networking.c
sarthakaggarwal97: I don't see a strong reason to use bool here, I can make do with 0/1. I'm new to the project, so that's why I used bool I guess. Switching it to int function for now.

src/networking.c
sarthakaggarwal97: TIL!

src/networking.c
sarthakaggarwal97: Ideally, it would work both ways with this implementation. If we want to restrict the IDs towards the end, then I can look into make some changes. 
Happy to let you and @madolson take the call on this.

Currently we have syntax error if non numeric, or unknown filter is passed.
```
127.0.0.1:6379> client list
id=3 addr=127.0.0.1:63139 laddr=127.0.0.1:6379 fd=8 name= age=48081 idle=46728 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=484 tot-net-out=215973 tot-cmds=13
id=8 addr=127.0.0.1:63183 laddr=127.0.0.1:6379 fd=9 name= age=48016 idle=48016 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=209291 tot-cmds=1
id=11 addr=127.0.0.1:64620 laddr=127.0.0.1:6379 fd=10 name= age=3 idle=0 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=10 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1946 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=53 tot-net-out=209291 tot-cmds=1
127.0.0.1:6379> client list id 3 str
ERR syntax error

```

src/networking.c
sarthakaggarwal97: > Do we want CLIENT LIST MAXAGE 1000 to mean "list all clients older than 1000"? It sounds illogical

Yeah, honestly it was a bit strange to me as well because KILL and LIST are kinda opposite in what they do. But currently, we return the list of clients with age greater than the specified maxage.

> The safest solution is maybe to skip MAXAGE for CLIENT LIST

Yes we can always skip it. But would we want to support this in the future? If yes, maybe we can properly document the behavior and keep supporting the maxage filter.

@madolson @hpatro would love to get your thoughts on this as well.

src/networking.c
sarthakaggarwal97: But, I think it's still good to have a conversation around it. It can potentially make code clearer I guess and helpful in few usecases as well.

src/networking.c
sarthakaggarwal97: > We can also consider splitting it into smaller functions, if it helps. This function is very large.

If its okay, I can take it up in a separate PR? Both PRs would be easier to review separately.

src/networking.c
sarthakaggarwal97: finally done :D

src/networking.c
zuiderkwast: Yes, I agree with you.

src/networking.c
zuiderkwast: No, let's not add technical dept with some plan to fix it later. Many times, that kind of follow-up never happens. Review is not too hard.

If you want, you can do the refactoring in separate commits within the same PR. It's possible to look at the diff commit-per-commit.

src/networking.c
```diff
@@ -41,11 +41,56 @@
...
+void clientCommandGetName(client *c);
+void clientCommandUnpause(client *c);
+void clientCommandPause(client *c);
+void clientCommandTracking(client *c);
+void clientCommandCaching(client *c);
+void clientCommandGetredir(client *c);
+void clientCommandTrackingInfo(client *c);
+void clientCommandNoTouch(client *c);
+void clientCommandCapa(client *c);
+void clientCommandImportSource(client *c);
```
zuiderkwast: The split of clientCommand into these functions looks good!

Let's just make these `static void`. Static means private to the current c file. (Need to add `static` in both places, here and at the implementation.)

src/networking.c
```diff
@@ -3557,236 +3618,273 @@ void quitCommand(client *c) {
...
+
+void clientCommandList(client *c) {
+    int type = -1;
+    sds response = NULL;
+
+    if (c->argc > 3) {
+        clientFilter filter = {.ids = NULL, .max_age = 0, .addr = NULL, .laddr = NULL, .user = NULL, .type = -1, .skipme = 0};
+        int i = 2;
+
+        if (parseClientFilters(c, i, &filter) != C_OK) {
```
zuiderkwast: Many other functions that send an error reply when parsing fails are named ending with "OrReply" (example: `getTimeoutFromObjectOrReply`). I think it's a good naming convention. Let's use it for this function too.

```suggestion
        if (parseClientFiltersOrReply(c, i, &filter) != C_OK) {
```

src/networking.c
```diff
@@ -3557,236 +3618,273 @@ void quitCommand(client *c) {
...
          * CLIENT KILL <option> [value] ... <option> [value] */
-        char *addr = NULL;
-        char *laddr = NULL;
-        user *user = NULL;
-        int type = -1;
-        uint64_t id = 0;
-        long long max_age = 0;
-        int skipme = 1;
+
+        listNode *ln;
```
zuiderkwast: This indentation looks wrong. If you run `clang-format -i networking.c` it can fix it for you, or when you fix the merge conflict, the CI job for clang-format will tell you.

zuiderkwast (commented):
The refactoring looks good. I have just a few small comments and there is some merge conflict.

src/networking.c
sarthakaggarwal97: yeah I think it fixed a bunch. Will wait for the clang CI to pass.

src/networking.c
sarthakaggarwal97: thanks, addressed.

src/networking.c
sarthakaggarwal97: thanks, addressed.

src/networking.c
sarthakaggarwal97: It passed!

src/networking.c
zuiderkwast: Yes, I know it's possible. The question is if there is some reason to forbid it, for example to make it less confusing for users or some tools. Personally, I think we can allow it, but API changes are TSC majority decisions.

src/networking.c
```diff
@@ -3559,564 +3620,681 @@ void quitCommand(client *c) {
...
+        "    * ADDR <ip:port>",
+        "      Return clients connected from the specified address.",
+        "    * LADDR <ip:port>",
+        "      Return clients connected to the specified local address.",
+        "    * ID <client-id>",
+        "      Return clients with the specified IDs.",
+        "    * SKIPME (YES|NO)",
+        "      Exclude the current client from the list (default: no).",
+        "    * MAXAGE <maxage>",
+        "      Kill connections older than the specified age.",
```
zuiderkwast: This is help for CLIENT LIST, so "Kill" is not right here.

src/networking.c
```diff
@@ -41,11 +41,56 @@
 #include <math.h>
 #include <ctype.h>
 #include <stdatomic.h>
-
+#include "intset.h"
```
zuiderkwast: Please move this `#include "..."` to the other ones, above all the standard lib `#include <...>` ones.

zuiderkwast (commented):
LGTM. Just two nits.

src/networking.c
```diff
@@ -3559,564 +3620,681 @@ void quitCommand(client *c) {
...
+        "    * TYPE (NORMAL|PRIMARY|REPLICA|PUBSUB)",
+        "      Return clients of specified type.",
+        "    * USER <username>",
+        "      Return clients authenticated by <username>.",
+        "    * ADDR <ip:port>",
+        "      Return clients connected from the specified address.",
+        "    * LADDR <ip:port>",
+        "      Return clients connected to the specified local address.",
+        "    * ID <client-id>",
+        "      Return clients with the specified IDs.",
```
zuiderkwast: We allow multiple client-id

```suggestion
        "    * ID <client-id> [<client-id>...]",
        "      Return clients with the specified IDs.",
```

We should update the same line for CLIENT KILL too.

zuiderkwast:
@valkey-io/core-team Please approve the extended syntax for CLIENT LIST as described in the top comment.

The MAXAGE argument for CLIENT LIST is a little strange, because it's the max age of clients to *exclude*, for symmetry with CLIENT KILL. An alternative is to forbid MAXAGE for CLIENT LIST.

src/networking.c
```diff
@@ -3440,6 +3485,22 @@ sds getAllClientsInfoString(int type, int hide_user_data) {
     return o;
 }
 
+sds getAllFilteredClientsInfoString(clientFilter *client_filter, int hide_user_data) {
```
hpatro: We can make this static

src/networking.c
```diff
@@ -3440,6 +3485,22 @@ sds getAllClientsInfoString(int type, int hide_user_data) {
...
+    listNode *ln;
+    listIter li;
+    client *client;
+    sds o = sdsnewlen(SDS_NOINIT, 500);
```
hpatro: Why did we choose this magic number `500` ?

src/networking.c
```diff
@@ -3440,6 +3485,22 @@ sds getAllClientsInfoString(int type, int hide_user_data) {
...
+    sdsclear(o);
```
hpatro, line 3389: Do we need to clear this at init?

src/networking.c
```diff
@@ -3559,564 +3620,681 @@ void quitCommand(client *c) {
     c->flag.close_after_reply = 1;
 }
 
-void clientCommand(client *c) {
-    listNode *ln;
-    listIter li;
+static int parseClientFiltersOrReply(client *c, int i, clientFilter *filter) {
```
hpatro:
```suggestion
static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter) {
````

File: src/networking.c [hunk 22]
```diff
@@ -3559,564 +3620,681 @@ void quitCommand(client *c) {
...
-            int j;
-            o = sdsempty();
-            for (j = 3; j < c->argc; j++) {
-                long long cid;
-                if (getLongLongFromObjectOrReply(c, c->argv[j], &cid, "Invalid client ID")) {
-                    sdsfree(o);
-                    return;
+            i++; // Move to the first ID after "ID"
+
+            // Process all IDs until a non-numeric argument or end of args
```
hpatro: There used to be a preference for multi line comments. But I've not seen clang format check complaining about it.

src/networking.c
```diff
@@ -3559,564 +3620,681 @@ void quitCommand(client *c) {
...
-        addReply(c, shared.ok);
-    } else if (!strcasecmp(c->argv[1]->ptr, "caching") && c->argc >= 3) {
-        if (!c->flag.tracking) {
-            addReplyError(c, "CLIENT CACHING can be called only when the "
-                             "client is in tracking mode with OPTIN or "
-                             "OPTOUT mode enabled");
+
+        if ((options.tracking_optin && c->flag.tracking_optout) ||
+            (options.tracking_optout && c->flag.tracking_optin)) {
+            addReplyError(c, "You can't switch OPTIN/OPTOUT mode before disabling "
```
hpatro, line 3987: Are we trying to introduce new error message here ?

src/networking.c
```diff
@@ -3559,564 +3620,681 @@ void quitCommand(client *c) {
...
+    } else if (!strcasecmp(c->argv[1]->ptr, "pause") && (c->argc == 3 || c->argc == 4)) {
+        clientCommandPause(c);
+    } else if (!strcasecmp(c->argv[1]->ptr, "tracking") && c->argc >= 3) {
+        clientCommandTracking(c);
+    } else if (!strcasecmp(c->argv[1]->ptr, "caching") && c->argc >= 3) {
+        clientCommandCaching(c);
+    } else if (!strcasecmp(c->argv[1]->ptr, "getredir") && c->argc == 2) {
+        clientCommandGetredir(c);
+    } else if (!strcasecmp(c->argv[1]->ptr, "trackinginfo") && c->argc == 2) {
+        clientCommandTrackingInfo(c);
```
hpatro: Do we need this entire piece of refactoring though? It would be ideal if we can limit the changes to the scope of the PR.

src/networking.c
sarthakaggarwal97: done!

src/networking.c
sarthakaggarwal97: done!

src/networking.c
sarthakaggarwal97: done!

src/networking.c
zuiderkwast: @hpatro I actually suggested to split it into smaller functions in this comment: https://github.com/valkey-io/valkey/pull/1401#discussion_r1880090690

For CLIENT LIST and CLIENT KILL, it makes the code simpler. Technically, we don't need to do it for all the others though if we want to minimize the diff. This refactoring was very cleanly done though, so the body of each subcommand implementation doesn't appear to be changed in the diff. They're still in the same order and unchanged.

src/networking.c
zuiderkwast: Yes, we used to enforce `/*  */` style comments, but I've given up on enforcing it.

src/networking.c
zuiderkwast: It's my fault. :)

In getAllClientsInfoString, there's a fixed `200 * listLength(server.clients)` but that's for listing all clients. Here, we might just list a few client out of many, so I though just some hundred bytes is a better initial size.

src/networking.c
sarthakaggarwal97: This was a part of refactoring. I have not changed the existing implementation.

zuiderkwast: It's not new. Did you review the diff for individual commits? If you look at the diff for the whole PR, you see that this is not changed.

src/networking.c
zuiderkwast: Yes, I belive we do.

`sdsnewlen` with `SDS_NOINIT` creates a string of length 500 with undefined content. `sdsclear` sets the nul terminator and the length.

```c
void sdsclear(sds s) {
    sdssetlen(s, 0);
    s[0] = '\0';
}
```

sarthakaggarwal97: I followed suit from here: https://github.com/valkey-io/valkey/blob/unstable/src/networking.c#L3432
I don't think its needed though. Will check if its noop, then will remove it.

src/networking.c
sarthakaggarwal97: lmk folks if there is a better / right value for it. Not sure how to measure this exactly.

hpatro: Ah okay. never bothered to expand all the resolved conversation. I think we should be fine to let it expand on it's and not define this hardcoded value. WDYT?

src/networking.c
hpatro: It might sound counterintuitive to modern best practices but I got used to not jumping around the IDE with this project and look at these if blocks together in one place. I don't have a strong motivation to block this change though.

src/networking.c
hpatro: @zuiderkwast Do you know why it was enforced earlier though?

sarthakaggarwal97: Have changed few comments!

src/networking.c
zuiderkwast: Yeah it would be fine to start from empty and let it expand dynamically. I don't mind. It will expand to ~300 for the first client, then double the allocation when needed.

sarthakaggarwal97: noted. have updated it to empty.

src/networking.c
zuiderkwast: Yeah, I agree, and we usually want to avoid unnecessary changes.

What I commented on before this refactoring was that in an earlier commit in this PR, there were multiple `zfree` on each error case, something like this:

```C
        if (...) {
            if (parseClientFilters(c, i, &client_filter) != C_OK) {
                zfree(client_filter.ids);
                return;
            }
        } else {
            zfree(client_filter.ids);
            return;
        }

        /* finally, the non-error case */
        ...
        zfree(client_filter.ids);
```

There were different ways to solve it, not necessarily split it up into multple functions. This was solved first by a goto, like this:

```C
        if (...) {
            if (parseClientFilters(c, i, &client_filter) != C_OK) {
                goto client_kill_done;
            }
        } else {
            goto client_kill_done;
        }

        /* finally, the non-error case */
        ...

client_kill_done:
        zfree(client_filter.ids);
```

Perhaps the `client_kill_done` label in the middle of a large function seemed odd and that's a reason to break CLIENT KILL into a separate function? I don't know, but I didn't want to be too picky about exactly how a contributor formulates the code. :)

@hpatro if you prefer to avoid the large chain of else-if, we could revert it, or there is another way: We could change `"function": "clientCommand"` in `src/commands/client-kill.json` to point directly to the function implementing this subcommand, `"function": "clientKillCommand"`, and similar. It would have to take the argv and argc arguments.

What's best?

src/commands/client-list.json
```diff
@@ -35,6 +35,10 @@
             [
                 "8.0.0",
                 "Replaced `master` `TYPE` with `primary`. `master` still supported for backward compatibility."
+            ],
+            [
+                "8.1.0",
+                "Added filters like USER, ADDR, LADDR, SKIPME, and MAXAGE"
```
zuiderkwast: Good, I forgot about this one. The word "like" sounds like it's some examples but it's actually a complete list.

```Suggestion
                "Added filters USER, ADDR, LADDR, SKIPME and MAXAGE"
```

zuiderkwast: I realize now that we have a change to ID of CLIENT KILL that we should update in client-kill.json.

It now accepts multiple IDs. Before, it only accepted one id. We should mention that in the history and add `"multiple": true` for the ID argument.

sarthakaggarwal97: @zuiderkwast missed this comment before. Have addressed it now. thank you

src/commands/client-kill.json
```diff
@@ -35,6 +35,10 @@
             [
                 "8.0.0",
                 "Replaced `master` `TYPE` with `primary`. `master` still supported for backward compatibility."
+            ],
+            [
+                "8.1.0",
+                "Accepts multiple IDs as an option."
```
zuiderkwast: The grammar looks a bit weird. I suggest:

```suggestion
                "`ID` option accepts multiple IDs."
```

src/networking.c
```diff
@@ -3561,564 +3622,684 @@ void quitCommand(client *c) {
...
+        client_filter.addr = c->argv[2]->ptr;
+        client_filter.skipme = 0; /* With the old form, you can kill yourself. */
+    } else if (c->argc > 3) {
+        int i = 2; /* Next option index. */
+
+        /* New style syntax: parse options. */
+        if (parseClientFiltersOrReply(c, i, &client_filter) != C_OK) {
+            /* Free the intset on error */
+            goto client_kill_done;
             return;
```
zuiderkwast: `return` is never reached. We can delete it, right? (After every goto.)

zuiderkwast (commented):
A few nits. You need to commit commands.def too. (If you have python3 installed and run `make`, it updates commands.def and you need to commit those changes.)

src/networking.c
sarthakaggarwal97: yeah I guess so! Have removed it. thanks

src/networking.c
hpatro: > We could change "function": "clientCommand" in src/commands/client-kill.json to point directly to the function implementing this subcommand, "function": "clientKillCommand", and similar. It would have to take the argv and argc arguments.

This would be definitely ideal. @sarthakaggarwal97 do you want to try this refactoring? It should be incremental work with the current change.  We wouldn't need the container command method to redirect to the sub command I believe.

hpatro:
@valkey-io/core-team Could you please have a look at this?

src/networking.c
sarthakaggarwal97: I have some questions. It would be great if you guys can help here:

1. Are we going to this refactoring only for Client List / Kill command or for all the client commands. And if so, then clientCommand function need not exist right? If we do it for Client List / Kill only, then we need to remove the options from the current clientCommand function?
2. Do we still separate out the functions for all the subcommands in the refactoring?

hwware:
> @valkey-io/core-team Could you please have a look at this?

I already check the new parameters for client command, but I have the reserved opinion for this: this pr consumes more cpu 
on client filters in server side, let us discuss it in the meeting (I prefer)

src/networking.c
zuiderkwast: > * Are we going to this refactoring only for Client List / Kill command or for all the client commands.

I think we can do it for all subcommands, but I'm fine with only LIST/KILL too. @hpatro What do you prefer?

> And if so, then clientCommand function need not exist right? If we do it for Client List / Kill only, then we need to remove the options from the current clientCommand function?

That's right. If none of the `src/commands/*.json` refers to clientCommand anymore, then clientCommand doesn't need to exist. There are no other calls to this function in this code base.

> * Do we still separate out the functions for all the subcommands in the refactoring?

IMO, we want to avoid the long else-if with calls to other functions. For the subcommands we separate out, fix the json files for those to point directly to the new functions. If we don't do it for all subcommands, then keep the code in clientCommand for the subcommands we don't want to change.

hpatro: I would prefer all sub commands.

hpatro:
> > @valkey-io/core-team Could you please have a look at this?
> 
> I already check the new parameters for client command, but I have the reserved opinion for this: this pr consumes more cpu on client filters in server side, let us discuss it in the meeting (I prefer)

This is not around the data path. Do you think it's concerning? I think administrator would benefit with the flexibility and limit the response size with the filters. Definitely, we can discuss more in the meeting.

src/networking.c
hpatro: Thanks for making the change. One note about this change, I thought maybe we could remove `src/commands/client.json` file however we would still need to have it in place. With it being there, the primary command gets defined and the server further allows operation on subcommands. Also, ACL rules also needs to support filtering at primary command level.

src/networking.c
```diff
@@ -42,10 +43,56 @@
...
+     cmd == clientCommandGetName ||      \
+     cmd == clientCommandUnpause ||      \
+     cmd == clientCommandPause ||        \
+     cmd == clientCommandTracking ||     \
+     cmd == clientCommandCaching ||      \
+     cmd == clientCommandGetredir ||     \
+     cmd == clientCommandTrackingInfo || \
+     cmd == clientCommandNoTouch ||      \
+     cmd == clientCommandCapa ||         \
+     cmd == clientCommandImportSource)
```
hpatro: The alternative to this is to retrieve the parent (`c->cmd->parent`) command and compare it with `clientCommand`.

src/networking.c
```diff
@@ -3561,566 +3626,640 @@ void quitCommand(client *c) {
     c->flag.close_after_reply = 1;
 }
 
-void clientCommand(client *c) {
```
hpatro, line 3476: Could you check what happens to the `parent` ptr for sub command(s) ? We might still need to have a dummy method of `clientCommand` to avoid invalid access of function not defined.

```suggestion
void clientCommand(client *c) {
        addReplySubcommandSyntaxError(c);
}
```

sarthakaggarwal97: yes, this seems to be working. I have added ClientCommand as a function in client.json

hpatro (approved):
Thanks @sarthakaggarwal97 for working on this! LGTM.

PingXie:
> The MAXAGE argument for CLIENT LIST is a little strange, because it's the max age of clients to _exclude_, for symmetry with CLIENT KILL. An alternative is to forbid MAXAGE for CLIENT LIST.

Yeah, it's a bit odd, but I think it's still pretty easy to understand, so I'm happy to keep it as is.

> > I already check the new parameters for client command, but I have the reserved opinion for this: this pr consumes more cpu on client filters in server side, let us discuss it in the meeting (I prefer)
> 
> This is not around the data path. Do you think it's concerning? I think administrator would benefit with the flexibility and limit the response size with the filters. Definitely, we can discuss more in the meeting.

Haven't looked at the code yet, but generally I'm okay with a bit of a performance hit for admin commands if it adds useful functionality.  Besides, `CLIENT LIST` is already flagged as "admin/slow/dangerous" anyway, so no surprises there. We can certainly discuss it in a meeting.

hpatro:
@valkey-io/core-team Please vote :thumbsup: or :thumbsdown:

src/commands/client-caching.json
```diff
@@ -6,7 +6,7 @@
         "since": "6.0.0",
         "arity": 3,
         "container": "CLIENT",
-        "function": "clientCommand",
+        "function": "clientCommandCaching",
```
enjoy-binbin: i think normally we should use clientCachingCommand? My impression is that i am using this to retrieve the proc, for example like other subcommand, functionListCommand configSetCommand

soloestoy: yes, `clientCachingCommand` is better

soloestoy:
I noticed that the `FALGS` filter hasn't been implemented. I think this would be a very useful filter.

For example, users could use `client list flags b` or `client kill flags b` to query or kill all blocked clients. We could even support a negation operation, like `client list flags !b` or `client kill flags !b`, to query or kill all non-blocked clients.

BTW, IP is also a very useful filter for querying or killing clients from specific IP addresses.

zuiderkwast:
> I noticed that the `FALGS` filter hasn't been implemented. I think this would be a very useful filter.

@soloestoy the PR is only unifying the filters of KILL and LIST. Add more filters in another PR. Full discussion in #668.

src/commands/client-caching.json
sarthakaggarwal97: Makes sense, thank you! Will refactor this in a bit.

sarthakaggarwal97: I have updated.

madolson:
Core team meeting. Can we run a quick performance test to validate that it's not significantly regressing. Let's run `valkey-benchmark` with 10 and 10000 clients connected, and see how much more time it takes to return. Also include `cmdstats` so that we can compare IO vs server side compute time.

hpatro:
> Core team meeting. Can we run a quick performance test to validate that it's not significantly regressing. Let's run `valkey-benchmark` with 10 and 10000 clients connected, and see how much more time it takes to return. Also include `cmdstats` so that we can compare IO vs server side compute time.

Current suggestion:

1. Write a small script to generate idle ‘N’ connections or use the benchmark to generate active clients via ` valkey-benchmark -l -c 10 -t PING`
2. Use the benchmark on separate terminal to gauge the performance of CLIENT LIST (possibly with filters) command. `valkey-benchmark -n 100000 -c 1 CLIENT LIST`

Let’s capture data with modifying client parameter for both steps. 

If there are any other suggestion @hwware / @zuiderkwast, feel free to add. We'll capture that.

zuiderkwast:
> Can we run a quick performance test to validate that it's not significantly regressing. Let's run `valkey-benchmark` with 10 and 10000 clients connected, and see how much more time it takes to return. Also include `cmdstats` so that we can compare IO vs server side compute time.

For the benchmark, we should run `CLIENT LISTS` with some filter, right? `CLIENT LIST TYPE NORMAL`?

hwware:
> > Core team meeting. Can we run a quick performance test to validate that it's not significantly regressing. Let's run `valkey-benchmark` with 10 and 10000 clients connected, and see how much more time it takes to return. Also include `cmdstats` so that we can compare IO vs server side compute time.
> 
> Current suggestion:
> 
> 1. Write a small script to generate idle ‘N’ connections or use the benchmark to generate active clients via ` valkey-benchmark -l -c 10 -t PING`
> 2. Use the benchmark on separate terminal to gauge the performance of CLIENT LIST (possibly with filters) command. `valkey-benchmark -n 100000 -c 1 CLIENT LIST`
> 
> Let’s capture data with modifying client parameter for both steps.
> 
> If there are any other suggestion @hwware / @zuiderkwast, feel free to add. We'll capture that.

I would like to suggest as following:

1. Run valkey-benchmark as 10000 client read/write request (datasize could be 256 bytes or several kb)
2. Run valkey-benchmark as 10 client with CLIENT List "filter", the filter can be some combinations as the following:
     * USER <username>
       Return clients authenticated by <username>.
     * ADDR <ip:port>
       Return clients connected from the specified address.
     * LADDR <ip:port>
       Return clients connected to the specified local address.
     * MAXAGE <maxage>
       Only list connections older than the specified age.

sarthakaggarwal97:
@hwware @zuiderkwast @hpatro Thank you for the suggestions, really helpful. 

I had just one query, since some of the filters were not supported in `Client List` originally, how would we compare the performance through benchmark? Or do we want to compare the performance with just `Client List`?
We can definitely compare with `id/type` filters which were supported before.

sarthakaggarwal97:
Hi folks, sharing some data around the performance. The results do not look very deviated and are possibly within margin. I have tried my best to keep the setup as similar as possible. I was not able to get the data for 10000 clients, as both of my machines kinda couldn't process it (on both branches). If required, I can try it on a beefier machine and get back.

Screenshot 2025-01-13 at 6 54 37 PM

sarthakaggarwal97:
I was able to extract the numbers for 10000 clients as well. Seeing negligible variance across the latency / throughput parameters for with and without filters.

Sharing Command Info Stats for 10000 Clients. Please review.

#### Unstable:

##### Command: `client list`
`cmdstat_client|list:calls=100000,usec=1828098028,usec_per_call=18280.98,rejected_calls=0,failed_calls=0`


##### Command: `client list type normal`
`cmdstat_client|list:calls=100000,usec=1746052452,usec_per_call=17460.53,rejected_calls=0,failed_calls=0`


#### PR:

##### Command: `client list`
`cmdstat_client|list:calls=100000,usec=1736309756,usec_per_call=17363.10,rejected_calls=0,failed_calls=0`


##### Command: `client list type normal`
`cmdstat_client|list:calls=100000,usec=1835707578,usec_per_call=18357.08,rejected_calls=0,failed_calls=0`

### Notes:

**Setup:**
1. Machine: `c5.12xlarge`
2. Running: `valkey-benchmark -l -c 10 -t PING`
3. Running benchmarks for client list (with and without filters)
4. Resetting stats using `config resetstat` after every run.

hwware:
> @hwware @zuiderkwast @hpatro Thank you for the suggestions, really helpful.
> 
> I had just one query, since some of the filters were not supported in `Client List` originally, how would we compare the performance through benchmark? Or do we want to compare the performance with just `Client List`? We can definitely compare with `id/type` filters which were supported before.

My idea is sending read/write operations through 10000 or more clients, and at the same time, simulating operator runs CLIENT LIST filter in another valkey-bench, and you can watch if there are qps drop in the read/write side. 

Because the client list with filter will execute a slow operation, the server qps could drop. If there is no significant drop, we can merge this PR, thanks

zuiderkwast:
I can see in the benchmarks that this PR actually makes CLIENT LIST faster than before with 1000 clients connected. It seems that there is some initial cost to set up the filter struct (so slightly dropped latency when there are very few clients connected) but once the filter struct is set up, it is faster to filter each client.

Because of this, I would assume that QPS for the read/write clients will be affected less by CLIENT LIST with this PR compared to unstable. Testing is of course better, but it looks pretty safe to me.

sarthakaggarwal97:
Thanks folks for the suggestions and inputs. 

Based on @hwware suggestions, I performed the SET/GET operations while running `client list <filter> <filter-value>` for 10000 clients. 
I do not see any major deviation in performance from the current unstable command.

Sharing the results:

Screenshot 2025-01-14 at 3 09 17 PM

Notes:
1. Latency Unit: `msec`
2. RPS: `requests per second`

hpatro:
@sarthakaggarwal97 Thanks for sharing the results and is inline with our expectation. I'm good with the stats. The above benchmarking is performed with one client running the `CLIENT LIST (filters)` and 10k clients running SET/GET operations. The setup is quite similar to a real world scenario where the no. of application clients >>> no. of administrator clients.

hwware:
It looks good to me.  Approved

hwware (approved):

hpatro:
I see atleast 4 core-team member have approved this either through voting or PR approval. @valkey-io/core-team Do you want to review further or is it good to be merged?

zuiderkwast:
Let's merge it.

@sarthakaggarwal97 do you want to update the docs too? The `commands/*.md` files in the valkey-doc repo. We can have one doc PR for all these updates if you want.

sarthakaggarwal97:
Thank you everyone for helping in this one. 

@zuiderkwast let me raise the doc PR as well. Thank you!

PR #1402, opened 2024-12-06
Fix sanitizer builds with clang

### LINKED ISSUE #1394: [test-failure] Sanitizer builds, '_DEFAULT_SOURCE' macro redefined
The unit tests for networking.c were introduced in #1360. @uriyage can you take a look? Can it be solved by just reordering the includes?

* https://github.com/valkey-io/valkey/actions/runs/12170639212/job/33946009664
* https://github.com/valkey-io/valkey/actions/runs/12150400350/job/33883091510

```
In file included from unit/test_networking.c:3:
In file included from unit/../networking.c:30:
In file included from unit/../server.h:33:
unit/../fmacros.h:37:9: error: '_DEFAULT_SOURCE' macro redefined [-Werror,-Wmacro-redefined]
#define _DEFAULT_SOURCE
        ^
/usr/include/features.h:236:10: note: previous definition is here
# define _DEFAULT_SOURCE        1
         ^
In file included from unit/test_networking.c:3:
unit/../networking.c:2049:22: error: use of undeclared identifier 'IOV_MAX'
    int iovmax = min(IOV_MAX, c->conn->iovcnt);
                     ^
unit/../networking.c:2049:22: error: use of undeclared identifier 'IOV_MAX'
```

zuiderkwast:
By including <stdatomic.h> after the other includes in the unit test, we can avoid redefining a macro which led to a build failure.

Fixes #1394

### DIFF
--- a/src/fmacros.h
+++ b/src/fmacros.h
@@ -34,7 +34,7 @@
 
 #if defined(__linux__)
 #define _GNU_SOURCE
-#define _DEFAULT_SOURCE 2
+#define _DEFAULT_SOURCE
 #endif
 
 #if defined(_AIX)
--- a/src/fmacros.h
+++ b/src/fmacros.h
@@ -34,7 +34,7 @@
 
 #if defined(__linux__)
 #define _GNU_SOURCE
-#define _DEFAULT_SOURCE
+#define _DEFAULT_SOURCE 2
 #endif
 
 #if defined(_AIX)
--- a/src/unit/test_networking.c
+++ b/src/unit/test_networking.c
@@ -1,8 +1,8 @@
-#include <stdatomic.h>
-
 #include "../networking.c"
 #include "../server.c"
 #include "test_help.h"
+
+#include <stdatomic.h>
 
 int test_backupAndUpdateClientArgv(int argc, char **argv, int flags) {
     UNUSED(argc);


zuiderkwast:
We have `#include <stdatomic.h>` in several files now, so I wonder why we still have this conditional code in the Makefile:

```make
# Detect if the compiler supports C11 _Atomic.
# NUMBER_SIGN_CHAR is a workaround to support both GNU Make 4.3 and older versions.
NUMBER_SIGN_CHAR := \#
C11_ATOMIC := $(shell sh -c 'echo "$(NUMBER_SIGN_CHAR)include <stdatomic.h>" > foo.c; \
	$(CC) -std=gnu11 -c foo.c -o foo.o > /dev/null 2>&1; \
	if [ -f foo.o ]; then echo "yes"; rm foo.o; fi; rm foo.c')
ifeq ($(C11_ATOMIC),yes)
	STD+=-std=gnu11
else
	STD+=-std=c99
endif
```

If we can't compile the `foo.c` test file here, we can't compile replication.c and other files, so we're effectively requiring it and we're always building with -std=gnu11 now?

zuiderkwast:
@uriyage FYI this fixes it.

enjoy-binbin (approved):

PR #1403, opened 2024-12-06
Fix Module Update Args test when other modules are loaded

### LINKED ISSUE #1400: [test-failure] Module Update Args test fails on TLS-module CI jobs
Daily jobs with TLS built as a module are failing since #1041 was merged.

```
*** [err]: Module Update Args in tests/unit/moduleapi/moduleconfigs.tcl
Expected '{10 20 30}' to be equal to '{10 20 30} {}' (context: type eval line 7 cmd {assert_equal "{10 20 30}" [lmap x [r module list] {dict get $x args}]} proc ::test)
```

The TLS module is another module here. We need to change how we do this assert to only check the args of the module were testing.

@hwware ~~Can you fix it?~~ I fixed it in #1403. PTAL

zuiderkwast:
Fixes #1400


enjoy-binbin (approved):

hwware:
@zuiderkwast Thanks for fixing it.

PR #1405, opened 2024-12-08
client struct: lazy init components and optimize struct layout

uriyage:
# Refactor client structure to use modular data components

## Current State
The client structure allocates memory for replication / pubsub / multi-keys / module / blocked data for every client, despite these features being used by only a small subset of clients. In addition the current field layout in the client struct is suboptimal, with poor alignment and unnecessary padding between fields, leading to a larger than necessary memory footprint of 896 bytes per client. Furthermore, fields that are frequently accessed together during operations are scattered throughout the struct, resulting in poor cache locality.

## This PR's Change

1.  Lazy Initialization 
- **Components are only allocated when first used:**
  - PubSubData: Created on first SUBSCRIBE/PUBLISH operation
  - ReplicationData: Initialized only for replica connections
  - ModuleData: Allocated when module interaction begins
  - BlockingState: Created when first blocking command is issued
  - MultiState: Initialized on MULTI command

2. Memory Layout Optimization:
   - Grouped related fields for better locality
   - Moved rarely accessed fields (e.g., client->name) to struct end
   - Optimized field alignment to eliminate padding

3. Additional changes:
   - Moved watched_keys to be static allocated in the `mstate` struct
   - Relocated replication init logic to replication.c
  

### Key Benefits
- **Efficient Memory Usage:**
- 45% smaller base client structure - Basic clients now use 528 bytes (down from 896).
- Better memory locality for related operations
- Performance improvement in high throughput scenarios. No performance regressions in other cases.

### Performance Impact

Tested with 650 clients and 512 bytes values.

#### Single Thread Performance
| Operation   | Dataset | New (ops/sec) | Old (ops/sec) | Change % |
|------------|---------|---------------|---------------|-----------|
| SET        | 1 key   | 261,799      | 258,261      | +1.37%    |
| SET        | 3M keys | 209,134      | ~209,000     | ~0%       |
| GET        | 1 key   | 281,564      | 277,965      | +1.29%    |
| GET        | 3M keys | 231,158      | 228,410      | +1.20%    |

#### 8 IO Threads Performance
| Operation   | Dataset | New (ops/sec) | Old (ops/sec) | Change % |
|------------|---------|---------------|---------------|-----------|
| SET        | 1 key   | 1,331,578    | 1,331,626    | -0.00%    |
| SET        | 3M keys | 1,254,441    | 1,152,645    | +8.83%    |
| GET        | 1 key   | 1,293,149    | 1,289,503    | +0.28%    |
| GET        | 3M keys | 1,152,898    | 1,101,791    | +4.64%    |

#### Pipeline Performance (3M keys)
| Operation | Pipeline Size | New (ops/sec) | Old (ops/sec) | Change % |
|-----------|--------------|---------------|---------------|-----------|
| SET       | 10          | 548,964      | 538,498      | +1.94%    |
| SET       | 20          | 606,148      | 594,872      | +1.89%    |
| SET       | 30          | 631,122      | 616,606      | +2.35%    |
| GET       | 10          | 628,482      | 624,166      | +0.69%    |
| GET       | 20          | 687,371      | 681,659      | +0.84%    |
| GET       | 30          | 725,855      | 721,102      | +0.66%    |

### Observations:
1. Single-threaded operations show consistent improvements (1-1.4%)
2. Multi-threaded performance shows significant gains for large datasets:
   - SET with 3M keys: +8.83% improvement
   - GET with 3M keys: +4.64% improvement
3. Pipeline operations show consistent improvements:
   - SET operations: +1.89% to +2.35%
   - GET operations: +0.66% to +0.84%
4. No performance regressions observed in any test scenario

Related issue:https://github.com/valkey-io/valkey/issues/761

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -651,14 +651,14 @@ void *VM_PoolAlloc(ValkeyModuleCtx *ctx, size_t bytes) {
  * Helpers for modules API implementation
  * -------------------------------------------------------------------------- */
 
-void initClientModuleData(client *c) {
+static void initClientModuleData(client *c) {
     if (c->module_data) return;
     c->module_data = zcalloc(sizeof(ClientModuleData));
 }
 
 void freeClientModuleData(client *c) {
     if (!c->module_data) return;
-    /* Free the RedisModuleBlockedClient held onto for reprocessing if not already freed. */
+    /* Free the ValkeyModuleBlockedClient held onto for reprocessing if not already freed. */
     zfree(c->module_data->module_blocked_client);
     zfree(c->module_data);
     c->module_data = NULL;
--- a/src/module.c
+++ b/src/module.c
@@ -651,6 +651,19 @@ void *VM_PoolAlloc(ValkeyModuleCtx *ctx, size_t bytes) {
  * Helpers for modules API implementation
  * -------------------------------------------------------------------------- */
 
+void initClientModuleData(client *c) {
+    if (c->module_data) return;
+    c->module_data = zcalloc(sizeof(ClientModuleData));
+}
+
+void freeClientModuleData(client *c) {
+    if (!c->module_data) return;
+    /* Free the RedisModuleBlockedClient held onto for reprocessing if not already freed. */
+    zfree(c->module_data->module_blocked_client);
+    zfree(c->module_data);
+    c->module_data = NULL;
+}
+
 void moduleEnqueueLoadModule(sds path, sds *argv, int argc) {
     int i;
     struct moduleLoadQueueEntry *loadmod;
@@ -721,11 +734,11 @@ void moduleReleaseTempClient(client *c) {
     c->flag.fake = 1;
     c->user = NULL; /* Root user */
     c->cmd = c->lastcmd = c->realcmd = c->io_parsed_cmd = NULL;
-    if (c->bstate.async_rm_call_handle) {
-        ValkeyModuleAsyncRMCallPromise *promise = c->bstate.async_rm_call_handle;
+    if (c->bstate && c->bstate->async_rm_call_handle) {
+        ValkeyModuleAsyncRMCallPromise *promise = c->bstate->async_rm_call_handle;
         promise->c = NULL; /* Remove the client from the promise so it will no longer be possible to abort it. */
         freeValkeyModuleAsyncRMCallPromise(promise);
-        c->bstate.async_rm_call_handle = NULL;
+        c->bstate->async_rm_call_handle = NULL;
     }
     moduleTempClients[moduleTempClientCount++] = c;
 }
@@ -897,7 +910,7 @@ static CallReply *moduleParseReply(client *c, ValkeyModuleCtx *ctx) {
 
 void moduleCallCommandUnblockedHandler(client *c) {
     ValkeyModuleCtx ctx;
-    ValkeyModuleAsyncRMCallPromise *promise = c->bstate.async_rm_call_handle;
+    ValkeyModuleAsyncRMCallPromise *promise = c->bstate->async_rm_call_handle;
     serverAssert(promise);
     ValkeyModule *module = promise->module;
     if (!promise->on_unblocked) {
@@ -6569,7 +6582,7 @@ ValkeyModuleCallReply *VM_Call(ValkeyModuleCtx *ctx, const char *cmdname, const 
             .ctx = (ctx->flags & VALKEYMODULE_CTX_AUTO_MEMORY) ? ctx : NULL,
         };
         reply = callReplyCreatePromise(promise);
-        c->bstate.async_rm_call_handle = promise;
+        c->bstate->async_rm_call_handle = promise;
         if (!(call_flags & CMD_CALL_PROPAGATE_AOF)) {
             /* No need for AOF propagation, set the relevant flags of the client */
             c->flag.module_prevent_aof_prop = 1;
@@ -7679,7 +7692,7 @@ void VM_LatencyAddSample(const char *event, mstime_t latency) {
 
 /* Returns 1 if the client already in the moduleUnblocked list, 0 otherwise. */
 int isModuleClientUnblocked(client *c) {
-    ValkeyModuleBlockedClient *bc = c->bstate.module_blocked_handle;
+    ValkeyModuleBlockedClient *bc = c->bstate->module_blocked_handle;
 
     return bc->unblocked == 1;
 }
@@ -7697,7 +7710,7 @@ int isModuleClientUnblocked(client *c) {
  * The structure ValkeyModuleBlockedClient will be always deallocated when
  * running the list of clients blocked by a module that need to be unblocked. */
 void unblockClientFromModule(client *c) {
-    ValkeyModuleBlockedClient *bc = c->bstate.module_blocked_handle;
+    ValkeyModuleBlockedClient *bc = c->bstate->module_blocked_handle;
 
     /* Call the disconnection callback if any. Note that
      * bc->disconnect_callback is set to NULL if the client gets disconnected
@@ -7765,9 +7778,10 @@ ValkeyModuleBlockedClient *moduleBlockClient(ValkeyModuleCtx *ctx,
     client *c = ctx->client;
     int islua = scriptIsRunning();
     int ismulti = server.in_exec;
+    initClientBlockingState(c);
 
-    c->bstate.module_blocked_handle = zmalloc(sizeof(ValkeyModuleBlockedClient));
-    ValkeyModuleBlockedClient *bc = c->bstate.module_blocked_handle;
+    c->bstate->module_blocked_handle = zmalloc(sizeof(ValkeyModuleBlockedClient));
+    ValkeyModuleBlockedClient *bc = c->bstate->module_blocked_handle;
     ctx->module->blocked_clients++;
 
     /* We need to handle the invalid operation of calling modules blocking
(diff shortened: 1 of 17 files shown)


zuiderkwast:
> * 45% smaller base client structure - Basic clients now use 528 bytes (down from 896).

Amazing!

src/module.c
```diff
@@ -643,6 +643,19 @@ void *VM_PoolAlloc(ValkeyModuleCtx *ctx, size_t bytes) {
  * Helpers for modules API implementation
  * -------------------------------------------------------------------------- */
 
+void initClientModuleData(client *c) {
```
zuiderkwast: Let's use static for internal functions

src/module.c
```diff
@@ -643,6 +643,19 @@ void *VM_PoolAlloc(ValkeyModuleCtx *ctx, size_t bytes) {
...
+    if (c->module_data) return;
+    c->module_data = zcalloc(sizeof(ClientModuleData));
+}
+
+void freeClientModuleData(client *c) {
+    if (!c->module_data) return;
+    /* Free the RedisModuleBlockedClient held onto for reprocessing if not already freed. */
```
zuiderkwast: The what? ValkeyModuleBlockedClient?

src/multi.c
```diff
@@ -233,11 +244,11 @@ void execCommand(client *c) {
             serverAssert(c->flag.blocked == 0);
         }
 
-        /* Commands may alter argc/argv, restore mstate. */
-        c->mstate.commands[j].argc = c->argc;
-        c->mstate.commands[j].argv = c->argv;
-        c->mstate.commands[j].argv_len = c->argv_len;
-        c->mstate.commands[j].cmd = c->cmd;
+        /* Commands may alter argc/argv, restore mstate-> */
```
zuiderkwast: Too much search-replace? Replaced "." with "->" even in the end of this sentence.

src/server.h
```diff
@@ -1340,50 +1300,95 @@ typedef struct client {
...
-    rax *client_tracking_prefixes; /* A dictionary of prefixes we are already
-                                      subscribed to in BCAST mode, in the
-                                      context of client side caching. */
+typedef struct client {
+    uint64_t id; /* Client incremental unique ID. */
+    connection *conn;
+    sds querybuf;                        /* Buffer we use to accumulate client queries. */
+    size_t qb_pos;                       /* The position we have read in querybuf. */
+    robj **argv;                         /* Arguments of current command. */
+    int argc;                            /* Num of arguments of current command. */
```
zuiderkwast: Now that they are sorted and grouped in a better way, can you add some comment for each group of fields to explain how and why each group is grouped like it is? For example, the group of fields for reading and parsing a command accessed by IO threads. Otherwise, in the future, I guess someone will add random new fields anywhere in the middle and mess up again.

src/multi.c
```diff
@@ -33,33 +33,42 @@
...
-        initClientMultiState(c);
-    }
+    if (!c->mstate || !c->mstate->commands) return;
+
+    freeClientMultiStateCmds(c);
+    c->mstate->count = 0;
+    c->mstate->cmd_flags = 0;
+    c->mstate->cmd_inv_flags = 0;
+    c->mstate->argv_len_sums = 0;
+    c->mstate->alloc_count = 0;
```
zuiderkwast, line 71: This is called by discardTransaction() in EXEC and DISCARD.

Why not free the multi state in these cases? If a client has used transactions ones, it is likely to do it again, so therefore we keep these allocations?

zuiderkwast (commented):
Very good! It's a large diff but most of the changed lines are trivial changes. I can't verify everything so I have to trust you and the tests for the correctness.

I noticed in many helper functions, we're using these client sub-structs, for example `c->mstate->something`, without checking that `c->mstate` exists. It should always exist in these cases so it's not an error, but would it make sense to add `serverAssert(c->mstate != NULL)` in all these places?

If we have code coverage by tests, we will find these errors anyway, so IMHO it doens't matter too match if it's a null-pointer crash or an assert. What is your reasoning about this?

src/module.c
uriyage: Thanks, done.

src/module.c
uriyage: This was a copy-paste. fixed.

src/multi.c
uriyage: Thanks :)

src/server.h
uriyage: I added the comments. I will create a follow-up PR to group all input/output-related data under a static sub-struct for safer use with the I/O thread.

src/multi.c
uriyage: Yes, for performance reasons to avoid excessive allocation/deallocation cycles, and also for simplicity, as we fall back to the current logic once the client has used those structs once.

src/networking.c
```diff
@@ -264,7 +238,7 @@ void putClientInPendingWriteQueue(client *c) {
      * if not already done and, for replicas, if the replica can actually receive
      * writes at this stage. */
     if (!c->flag.pending_write &&
-        (c->repl_state == REPL_STATE_NONE || (isReplicaReadyForReplData(c) && !c->repl_start_cmd_stream_on_ack))) {
+        (!c->repl_data || c->repl_data->repl_state == REPL_STATE_NONE || (isReplicaReadyForReplData(c) && !c->repl_data->repl_start_cmd_stream_on_ack))) {
```
zuiderkwast: This line is very long. Break it like this?

```suggestion
        (!c->repl_data ||
         c->repl_data->repl_state == REPL_STATE_NONE ||
         (isReplicaReadyForReplData(c) && !c->repl_data->repl_start_cmd_stream_on_ack))) {
```

src/networking.c
```diff
@@ -1836,10 +1741,9 @@ void freeClient(client *c) {
     if (c->name) decrRefCount(c->name);
     if (c->lib_name) decrRefCount(c->lib_name);
     if (c->lib_ver) decrRefCount(c->lib_ver);
-    freeClientMultiState(c);
+    if (c->mstate) freeClientMultiState(c);
```
zuiderkwast: In the implementation of `freeClientMultiState`, I see code like `if (!c->mstate) return;`. So we check twice if `c->mstate` exists. Is that intentional?

If we only want to check it in one place, which place is better? If we check at the call site, the we can change the check in the function to an assert. Or if we check in the implementation, we don't need the check here and can just call `freeClientMultiState(c);` unconditionally. Or is check twice better for branch prediction or something?

It's a similar situation for freeClientPubSubData, freeClientModuleData, blocking state and some other client sub-structs.

Which style do you prefer?

src/networking.c
```diff
@@ -4417,16 +4324,16 @@ size_t getClientMemoryUsage(client *c, size_t *output_buffer_mem_usage) {
      * i.e. unused sds space and internal fragmentation, just the string length. but this is enough to
      * spot problematic clients. */
     mem += c->argv_len_sum + sizeof(robj *) * c->argc;
-    mem += multiStateMemOverhead(c);
+    mem += c->mstate ? multiStateMemOverhead(c) : 0;
```
zuiderkwast: Doesn't `multiStateMemOverhead` return 0 in this case, so we don't need the conditional code here?

```Suggestion
    mem += multiStateMemOverhead(c);
```

src/networking.c
```diff
@@ -4417,16 +4324,16 @@ size_t getClientMemoryUsage(client *c, size_t *output_buffer_mem_usage) {
...
 
     /* Add memory overhead of pubsub channels and patterns. Note: this is just the overhead of the robj pointers
      * to the strings themselves because they aren't stored per client. */
-    mem += pubsubMemOverhead(c);
+    mem += c->pubsub_data ? pubsubMemOverhead(c) : 0;
```
zuiderkwast: Change not needed because pubsubMemOverhead has this check and returns 0.

```suggestion
    mem += pubsubMemOverhead(c);
```

zuiderkwast (commented):
Just a few nits now.

src/networking.c
uriyage: Since this call occurs only once per connection, branch prediction is not critical. I will remove the external conditions.

zuiderkwast (approved):

src/replication.c
```diff
@@ -1230,6 +1232,72 @@ void syncCommand(client *c) {
...
+        client *replica = ln->value;
+        if (replica != except_me && replica->repl_data->repl_state == REPLICA_STATE_WAIT_BGSAVE_END) {
+            return 1;
+        }
+    }
+    return 0;
+}
+
+void initClientReplicationData(client *c) {
+    if (c->repl_data) return;
```
ranshid, line 1255: Not that I see a real concern here, but we might think this should have been under assert check which would have forced the user to make sure he freeClientReplicationData before init again (if that case ever exists)

src/networking.c
```diff
@@ -1651,11 +1610,7 @@ void clearClientConnectionState(client *c) {
     clientSetDefaultAuth(c);
     moduleNotifyUserChanged(c);
     discardTransaction(c);
-
-    pubsubUnsubscribeAllChannels(c, 0);
-    pubsubUnsubscribeShardAllChannels(c, 0);
-    pubsubUnsubscribeAllPatterns(c, 0);
-    unmarkClientAsPubSub(c);
+    freeClientPubSubData(c);
```
ranshid, line 1615: So in this case we perform the actions of freeClient (eg all the dictReleases) when we unlink the client. Not sure it is problematic, but it is a change in order which is very basic.

ranshid (commented):
I think the overall change looks O.K, but some cases are not 100% trivial IMO and the change impacts almost all logical flows.  I would suggest we run the daily tests to verify this change.

ranshid (approved):

PR #1407, opened 2024-12-08
Add tests for src/module examples

Codebells:
We found there is coredump in hellodict https://github.com/valkey-io/valkey/pull/1395 , and there is no tcl test for these example modules.
1. Add tcl for src/module example. 
    * helloblock tcl
    * hellodict tcl
    * hellotimer tcl
    * hellotype tcl
    * helloworld tcl
2. Add a common tcl function `check_commands_specs` to check command by `command getkeys`. 
    * Use this function, we can test all command by `command getkeys`, except the following two kind of command.
        * cmd flags is `CMD_KEY_NOT_KEY`
        * cmd with special parameters, like `xread`.

I recommand all module's tcl test to add this function to check command.

### DIFF
--- a/tests/unit/moduleapi/src_module_example.tcl
+++ b/tests/unit/moduleapi/src_module_example.tcl
@@ -1,0 +1,107 @@
+source tests/support/cli.tcl
+
+set helloblockmodule [file normalize src/modules/helloblock.so]
+set hellodictmodule [file normalize src/modules/hellodict.so]
+set hellotimermodule [file normalize src/modules/hellotimer.so]
+set hellotypemodule [file normalize src/modules/hellotype.so]
+set helloworldmodule [file normalize src/modules/helloworld.so]
+
+start_server {tags {"modules"}} {
+    r module load $hellodictmodule
+    r module load $helloblockmodule
+    r module load $hellotypemodule
+    r module load $hellotimermodule
+    r module load $helloworldmodule
+    r select 0
+
+    test {test module command specs} {
+        check_commands_specs r
+    }
+    test {test hellodict} {
+        foreach i {a b c d e f g h i j k l m n o p q r s t u v w x y z} {
+            r hellodict.set $i $i
+            assert_match [r hellodict.get $i] $i
+        }
+        assert_equal [llength [r hellodict.keyrange a b 2]] 2
+        assert_equal [llength [r hellodict.keyrange a z 2]] 2
+    }
+
+    test {test helloblock} {
+        assert_match [r hello.block 1 2] "Request timedout"
+        assert {[r hello.block 1 2000] > 0}
+        r set foo bar
+        assert {[llength [r hello.keys]] > 0}
+        r flushall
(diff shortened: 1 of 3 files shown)


tests/unit/moduleapi/src_module_example.tcl
```diff
@@ -0,0 +1,103 @@
...
+        set lexres [r hello.lexrange myzset "\[b" "\[f" 0 0]
+        assert_equal [llength $lexres] 2
+
+        r hset hkey a a
+        r hello.hcopy hkey a b
+        assert_match [r hget hkey a] [r hget hkey b]
+
+        assert_match [r hello.leftpad td 4 0] "00td"
+    }
+}
```
enjoy-binbin, line 107: please also do a module unload in the end

Codebells: When I try to unload hellotype module, I found there is only `VM_CreateDataType`function, why don't have any function to delete it? So that I can't unload hellotype module, because of `ERR Error unloading module: the module exports one or more module-side data types can't unload.`

artikell: i think u can skip unload hello_type module

madolson (commented):
Thanks for the contribution. I'm sort of skeptical of maintaining these src modules at all. They originally existed because we didn't have module API tests. I think now that we have module API tests, we should move all of the references to those test modules.

Codebells:
> Thanks for the contribution. I'm sort of skeptical of maintaining these src modules at all. They originally existed because we didn't have module API tests. I think now that we have module API tests, we should move all of the references to those test modules.

OK，I will move src/modules to test/modules, so we can maintain all these modules in test/modules.

enjoy-binbin:
> OK，I will move src/modules to test/modules, so we can maintain all these modules in test/modules.

i think we should either to keep it as a demo module that can guaid people to write the special type module. Or we should just remove it from the code.

i think all the content in src/modules are cover by test/modules, we don't need to keep src/modules in the test folder.

madolson:
> i think we should either to keep it as a demo module that can guide people to write the special type module. Or we should just remove it from the code.

I think that is more what the module API reference is for https://valkey.io/topics/modules-api-ref. I know Redis had an SDK repo as well for modules, maybe we have a module example repo?

If we make a module example repo, I would also like to raise the quality of the modules, because I think they're a bit hacky today.

Codebells:
> > OK，I will move src/modules to test/modules, so we can maintain all these modules in test/modules.
> 
> i think we should either to keep it as a demo module that can guaid people to write the special type module. Or we should just remove it from the code.
> 
> i think all the content in src/modules are cover by test/modules, we don't need to keep src/modules in the test folder.

We need to make a dicision on whether to delete or keep these files.    
  
No matter how we deal with these files, we should test command to avoid the problem in [hellodict](https://github.com/valkey-io/valkey/pull/1395) happen again.       

I think use the function `check_commands_specs` to check all command is a good way.

PR #1410, opened 2024-12-09
Use binary representation in assert crash log, cleanup in crash log

enjoy-binbin:
Change assert crash log to also use binary representation like 5bdd72bea77d4bb237441c9a671e80edcdc998ad.
And do not print the password in assert crash log like 56eef6fb5ab7a755485c19f358761954ca459472.

In addition, for 5bdd72bea77d4bb237441c9a671e80edcdc998ad, we will print '"argv"',
because originally the code would print a '', and sdscatrepr will add an extra "",
so now removing the extra '' here.

Extract the getArgvReprString method and clean up the code a bit.

Examples:
```
debug assert "\x00abc"

before:
client->argv[0] = "debug" (refcount: 1)
client->argv[1] = "assert" (refcount: 1)
client->argv[2] = "" (refcount: 1)

after:
client->argv[0] = "debug" (refcount: 1)
client->argv[1] = "assert" (refcount: 1)
client->argv[2] = "\x00abc" (refcount: 1)

debug panic "\x00abc"

before:
argc: '3'
argv[0]: '"debug"'
argv[1]: '"panic"'
argv[2]: '"\x00abc"'

after:
argc: 3
argv[0]: "debug"
argv[1]: "panic"
argv[2]: "\x00abc"
```

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -1049,6 +1049,14 @@ __attribute__((noinline, weak)) void _serverAssert(const char *estr, const char 
     bugReportEnd(0, 0);
 }
 
+/* Returns the argv argument in binary representation, limited to length 128. */
+sds getArgvReprString(robj *argv) {
+    robj *decoded = getDecodedObject(argv);
+    sds repr = sdscatrepr(sdsempty(), decoded->ptr, min(sdslen(decoded->ptr), 128));
+    decrRefCount(decoded);
+    return repr;
+}
+
 /* Checks if the argument at the given index should be redacted from logs. */
 int shouldRedactArg(const client *c, int idx) {
     serverAssert(idx < c->argc);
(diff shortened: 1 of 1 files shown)


src/debug.c
```diff
@@ -1049,6 +1049,14 @@ __attribute__((noinline, weak)) void _serverAssert(const char *estr, const char
     bugReportEnd(0, 0);
 }
 
+/* Returns the argv argument in binary representation, limited to length 128. */
+sds getArgvReprString(robj *argv) {
+    robj *decoded = getDecodedObject(argv);
```
zuiderkwast, line 1054: getDecodedObject can allocate memory. We used a buffer on the stack for this before (`char buf[128]`). I think the reason is that we want to avoid memory allocation in the crash handler code, becuase the crash can be because of OOM?

zuiderkwast (commented):
The crash report looks better with this change. Just I'm not sure about stack vs heap allocation in crash handler code.

src/debug.c
enjoy-binbin: It is a good point. In this case  it is a assert crash report. It doesn't seem to have the memory issue. 

And in the other one, it can cause by a oom and it is already doing the alloca though

zuiderkwast: > it is already doing the alloca though

Is it? Where? Stack allocation can still succeed if we run out of heap allocation memory.

enjoy-binbin: line 1900

zuiderkwast: Ah, it was already using getDecodedObject. OK, then, approved. :)

zuiderkwast (approved):

PR #1140, opened 2024-10-09
Fix empty response for ACL CAT category subcommand for module defined categories

roshkhatri:
The module commands which were added to acl categories were getting skipped when `ACL CAT category` command was executed. 

This PR fixes the bug.
Before:
```
127.0.0.1:6379> ACL CAT foocategory
(empty array)
```
After:
```
127.0.0.1:6379> ACL CAT foocategory
aclcheck.module.command.test.add.new.aclcategories
```

PR #1413, opened 2024-12-09
[Backporting]Fix empty response for ACL CAT category subcommand for module defined…

roshkhatri:
[Backporting]Fix empty response for ACL CAT category subcommand for module defined categories (#1140)

The module commands which were added to acl categories were getting skipped when `ACL CAT category` command was executed.

This PR fixes the bug.
Before:
```
127.0.0.1:6379> ACL CAT foocategory
(empty array)
```
After:
```
127.0.0.1:6379> ACL CAT foocategory
aclcheck.module.command.test.add.new.aclcategories
```

### DIFF
--- a/tests/unit/moduleapi/aclcheck.tcl
+++ b/tests/unit/moduleapi/aclcheck.tcl
@@ -131,7 +131,19 @@ start_server {tags {"modules acl"}} {
         assert_equal {User j7 has no permissions to run the 'aclcheck.module.command.aclcategories.write.function.read.category' command} $e
     }
 
-    test "Unload the module - aclcheck" {
+    test {test if foocategory acl categories is added} {
+        r acl SETUSER j8 on >password -@all +@foocategory
+        assert_equal [r acl DRYRUN j8 aclcheck.module.command.test.add.new.aclcategories] OK
+    }
+
+    test {test if ACL CAT output for the new category is correct} {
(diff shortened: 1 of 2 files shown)


tests/unit/moduleapi/aclcheck.tcl
```diff
@@ -131,7 +131,19 @@ start_server {tags {"modules acl"}} {
...
+    }
+
+    test {test if ACL CAT output for the new category is correct} {
+        assert_equal [r ACL CAT foocategory] aclcheck.module.command.test.add.new.aclcategories
+    }
+
+    test {test permission compaction and simplification for categories added by a module} {
+        r acl SETUSER j9 on >password -@all +@foocategory -@foocategory
+        catch {r ACL GETUSER j9} res
+        assert_equal {-@all -@foocategory} [lindex $res 5]
```
hpatro, line 146: All these tests aren't part of the `unstable` merge changes? I don't think we should get change beyond the scope of the commit https://github.com/valkey-io/valkey/commit/9b8a06137c20c6fa112e0ce5830bec51edf39999.

roshkhatri: Thats true.

roshkhatri:
I think this was a mistake to backport - we will not be able to backport this as we do not support modules to declare ACL categories in 7.2.

PR #60, opened 2024-03-28
Dual channel replication

naglera:
In this PR we introduce the main benefit of dual channel replication by continuously steaming the COB (client output buffers) in parallel to the RDB and thus keeping the primary's side COB small AND accelerating the overall sync process. By streaming the replication data to the replica during the full sync, we reduce
1. Memory load from the primary's node.
2. CPU load from the primary's main process. [Latest performance tests](#data)

## Motivation
* Reduce primary memory load. We do that by moving the COB tracking to the replica side. This also decrease the chance for COB overruns. Note that primary's input buffer limits at the replica side are less restricted then primary's COB as the replica plays less critical part in the replication group. While increasing the primary’s COB may end up with primary reaching swap and clients suffering, at replica side we’re more at ease with it. Larger COB means better chance to sync successfully.
* Reduce primary main process CPU load. By opening a new, dedicated connection for the RDB transfer, child processes can have direct access to the new connection. Due to TLS connection restrictions, this was not possible using one main connection. We eliminate the need for the child process to use the primary's child-proc -> main-proc pipeline, thus freeing up the main process to process clients queries.

 ## Dual Channel Replication high level interface design
  - Dual channel replication begins when the replica sends a `REPLCONF CAPA dual-channel` to the primary during initial 
     handshake. This is used to state that the replica is capable of dual channel sync and that this is the replica's main channel, which is not used for snapshot transfer. 
   - When replica lacks sufficient data for PSYNC, the primary will send `+DUALCHANNELSYNC` response instead 
     of RDB data. As a next step, the replica creates a new connection (rdb-channel) and configures it against 
     the primary with the appropriate capabilities and requirements. The replica then requests a sync 
     using the RDB channel. 
   - Prior to forking, the primary sends the replica the snapshot's end repl-offset, and attaches the replica 
     to the replication backlog to keep repl data until the replica requests psync. The replica uses the main 
     channel to request a PSYNC starting at the snapshot end offset. 
   - The primary main threads sends incremental changes via the main channel, while the bgsave process 
     sends the RDB directly to the replica via the rdb-channel. As for the replica, the incremental 
     changes are stored on a local buffer, while the RDB is loaded into memory. 
   - Once the replica completes loading the rdb, it drops the rdb channel connection and streams the accumulated incremental 
     changes into memory. Repl steady state continues normally.

## New replica state machine

image

## Data <a name="data"></a>
image

image

image

## Explanation 
[...]

PR #1422, opened 2024-12-11
replication: fix io-threads possible race by moving waitForClientIO

uriyage:
#### The Problem
In #60 (Dual channel replication) a new `connWrite` call was added before the `waitForClientIO` check. This created a race condition where the main thread may attempt to write to a client that could have pending writes in IO threads.

#### The Fix
Moved the `waitForClientIO()` call earlier in `syncCommand`, before any `connWrite` call. This ensures all pending IO operations are completed before attempting to write to the client.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -1036,6 +1036,9 @@ void syncCommand(client *c) {
     /* ignore SYNC if already replica or in monitor mode */
     if (c->flag.replica) return;
 
+    /* Wait for any IO pending operation to finish before changing the client state to replica */
+    waitForClientIO(c);
+
     /* Check if this is a failover request to a replica with the same replid and
      * become a primary if so. */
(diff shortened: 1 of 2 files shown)


ranshid:
@uriyage I am fine with the fix but let's also introduce assertion or debug assertion in connWrite to catch these things.

uriyage:
> @uriyage I am fine with the fix but let's also introduce assertion or debug assertion in connWrite to catch these things.

We need a flag on the connection to indicate that it is a client struct, which will allow us to assert it in connWrite. We will add this flag in PR #1338, and I will add the assertion after that PR is merged.

uriyage:
> > @uriyage I am fine with the fix but let's also introduce assertion or debug assertion in connWrite to catch these things.
> 
> We need a flag on the connection to indicate that it is a client struct, which will allow us to assert it in connWrite. We will add this flag in PR #1338, and I will add the assertion after that PR is merged.

@ranshid done

ranshid:
@uriyage . Sorry - can you please rebase this so I can merge later?

ranshid (approved):

PR #1424, opened 2024-12-11
Add latency sample for transaction and pipeline

RayCao99:
If users put too many commands in a transaction or pipeline, the execution time might exceed the `latency_monitor_threshold` value. 
So, Maybe it's necessary to record latency for pipeline and transaction, which can help them to find out potential misuse of transaction or pipeline.

### DIFF
--- a/src/multi.c
+++ b/src/multi.c
@@ -152,6 +152,7 @@ void execCommand(client *c) {
     int j;
     robj **orig_argv;
     int orig_argc, orig_argv_len;
+    mstime_t latency;
     struct serverCommand *orig_cmd;
 
     if (!c->flag.multi) {
@@ -188,6 +189,8 @@ void execCommand(client *c) {
 
     /* Exec all the queued commands */
     unwatchAllKeys(c); /* Unwatch ASAP otherwise we'll waste CPU cycles */
+
+    latencyStartMonitor(latency);
 
     server.in_exec = 1;
 
@@ -252,6 +255,11 @@ void execCommand(client *c) {
     c->argc = orig_argc;
     c->cmd = c->realcmd = orig_cmd;
     discardTransaction(c);
+
+    latencyEndMonitor(latency);
+    if (!server.loading) {
+        latencyAddSampleIfNeeded("transaction", latency);
+    }
 
     server.in_exec = 0;
 }
--- a/src/networking.c
+++ b/src/networking.c
@@ -3112,6 +3112,10 @@ int canParseCommand(client *c) {
 }
 
 int processInputBuffer(client *c) {
+    mstime_t latency;
+    latencyStartMonitor(latency);
+    int processed = 0;
+
     /* Parse the query buffer. */
     while (c->querybuf && c->qb_pos < sdslen(c->querybuf)) {
         if (!canParseCommand(c)) {
@@ -3146,8 +3150,12 @@ int processInputBuffer(client *c) {
              * ASAP in that case. */
             return C_ERR;
         }
+        processed++;
     }
-
+    latencyEndMonitor(latency);
+    if (processed > 1) {
+        latencyAddSampleIfNeeded("pipeline", latency);
+    }
     return C_OK;
 }
 


src/networking.c
```diff
@@ -3112,6 +3112,10 @@ int canParseCommand(client *c) {
 }
 
 int processInputBuffer(client *c) {
+    mstime_t latency;
+    latencyStartMonitor(latency);
```
xbasel, line 3116: Has the impact of `gettimeofday` been measured? it will be executed on every single command.

RayCao99: I believe `gettimeofday` is a very lightweight operation because of vDSO.

src/multi.c
```diff
@@ -152,6 +152,7 @@ void execCommand(client *c) {
     int j;
     robj **orig_argv;
     int orig_argc, orig_argv_len;
+    mstime_t latency;
```
madolson, line 155: For exec, I wonder if we should just allow the exec command to be in the slowlog.

madolson (commented):
I'm a bit dubious of the practical benefit of having a latency alarm on `processInputBuffer`. The latency you see from a deep pipeline is captured in some of the event loop metrics.

@enjoy-binbin Would appreciate your thoughts here.

enjoy-binbin:
> I'm a bit dubious of the practical benefit of having a latency alarm on processInputBuffer. The latency you see from a deep pipeline is captured in some of the event loop metrics.

this processInputBuffer does make sense to me, some client or proxy indeed will use a depth pipeline.

In addition, our slowlog currently only has the process time of the command, we dont have the command read / write time. This sometimes bothers me since the server process time is not equal to the command process time seen in the client side.

src/multi.c
soloestoy: This is also an option, but with the current implementation of the `EXEC` command, it is used to indicate the start of a transaction. If the execution time of other commands within the transaction is all accounted for in the `EXEC` command, could this cause ambiguity?

madolson: I know more people check slowlog then check the latency metrics. Someone also mentioned to me once it was weird that scripts are included even though also have slow internal commands. I don't have a strong opinion though.

JimB123:
For pipelining, there is no guarantee of atomicity.  If the main thread is blocked for an excessive period due to pipelining, I would assert that's a bug.  (I'm aware of a fix for this, not yet available in OSS.)  Rather than adding metrics to capture this case, I'd recommend addressing the root problem.

For EXEC, there's this issue https://github.com/valkey-io/valkey/issues/4261 and PR https://github.com/valkey-io/valkey/pull/4267

soloestoy:
@JimB123 Thanks for the feedback.

Agreed that pipelining has no atomicity guarantee — that's well understood, and
this PR isn't trying to monitor atomicity at all. What it targets is the latency
introduced by the batching itself: when a client (or proxy) packs a large number
of commands into a single pipeline, the main thread can be blocked for the
aggregate duration even though every individual command stays well below the
commandlog/latency threshold, so nothing is surfaced today. The intent here is
purely observability — exposing that batch-induced latency so operators can spot
the misuse and size their pipelines reasonably.

On "if the main thread is blocked for an excessive period due to pipelining,
that's a bug": a fix that bounds a single batch's processing time is valuable and
I'd welcome it, but it's complementary rather than a replacement. Even with such a
fix in place, an operator still benefits from knowing which client is sending
oversized pipelines and how much contiguous time those batches cost — that's
exactly the signal that lets them tune the batch size on the client/proxy side.
Bounding the symptom and giving visibility into the cause aren't mutually
exclusive.

For the EXEC side, I agree it's better handled on its own — #4261 / #4267 look
like the right home for transaction logging.

JimB123:
> a fix that bounds a single batch's processing time is valuable and I'd welcome it, but it's complementary rather than a replacement. Even with such a fix in place, an operator still benefits from knowing which client is sending oversized pipelines and how much contiguous time those batches cost — that's exactly the signal that lets them tune the batch size on the client/proxy side. Bounding the symptom and giving visibility into the cause aren't mutually exclusive.

I think that this becomes a non-problem if a client's pipelining time is bounded.  There's a conceptual issue.  From the Valkey server, a pipeline is a never-ending stream of requests - there's no such thing as a "batch".  However, many clients have implemented the interface for pipelining as a batch with a fixed number of commands - sending the batch as a unit, waiting for responses as a unit, before moving to the next batch.

Valkey performs a read, arbitrarily limited to 16k, and then executes commands until the end of the 16k chunk.  If we time-bound the processing (rather than continuing to the end of the 16k) then the latency problem is completely eliminated, and no observability is needed.  What would we even observe?  We don't know how much data is sent.  We don't have a concept of a batch.  We couldn't even add up all the parts that make a "batch".

If too many pipelined commands are sent, the issue eventually becomes one of flow control.  Valkey shouldn't read commands which can't be immediately processed (more than 1 chunk anyway).  When Valkey stops reading, the TCP input buffer fills, this causes the TCP output buffer on the client side to fill, and eventually, the client can't send any more.

Assuming that Valkey arbitrarily breaks large pipelines into tiny chunks (maybe < 1ms?), what would a latency metric expose?  How would we define it?

PR #1425, opened 2024-12-11
add parameter for limiting maximum AOF files size on disk

### LINKED ISSUE #540: [NEW] Limit maximum size on disk of AOF files. Avoid disk full, long load times.

If AOF rewrite fails repeatedly the size of the *.aof files will grow unbounded. Rewrite can fail for a number of reasons, one of the more obivous ones is when memory pressure and write rate on the system is high.

If this continues the following undesirable effects surface:
1 - the time-to-load the larger *.aof files will increase any may exceed some system goal.
2 - (more critical) the disk partition may fill up.  At this point valkey will start failing write commands - which normally would lower memory pressure an allow rewrite to succeed and this would recover the node.  In this scenario there is no free disk space for rewrite to succeed so the node cannot recover

(Note: I plan to submit the pull request, I have a working change in our internal repository).

Valkey already tracks ```server.aof_current_size``` 

I propose to add a new configurable variable aof_max_size which acts like an artificial disk size. At the beginning of aofWrite() we can perform a check & return a ENOSPC error. In this scenario valkey will treat this situation as disk full & begin blocking new write commands. However no check is performed during aof rewrite (or other disk write funcions) so AOF rewrite can suceed & recover the node to usable once the disk space is freed.

We considered putting the *.aof files in their own partition but that would make administering valkey much more complex for just a small behavior change.
We considered having valkey check the free space on disk against some threshold. But that would be a more complex feature to build and execute. While it makes some sense for the avoiding disk full use case it is not so intuitive for the avoiding long load time use case.

**Additional information**

n/a

kronwerk:
one can optionally want to limit AOF disk usage not with disk real size, but with some tailored value. this PR makes it possible.
possible use cases:
1. filling disk up with AOF to the real limit may break something in user's environment (logs, watchdogs, anything) - setting lower value prevents that completely;
2. setting this new limit to some low value may allow automatic solving of filling disk problem for those users who don't care too much about disk cost:
- intense writes may overcome automatic AOF rewrite with some user's settings at some moment;
- AOF reach lower than full disk limit (e.g. 50-60% of disk);
- eventually automatic AOF rewrite comes in once more and finally fixes that.
while on 100% filled disk (like nowadays) that automatic solution is impossible due to insufficient space for tmp AOF rewrite files (0 space, actually) - one should size up disk under running database which makes some people a bit nervous, as I noticed, and in some cases that's not always convenient or even possible.

Fixes #540


src/aof.c
```diff
@@ -1010,16 +1010,22 @@ int startAppendOnly(void) {
...
- * there is an actual error condition we'll get it at the next try. */
-ssize_t aofWrite(int fd, const char *buf, size_t len) {
-    ssize_t nwritten = 0, totwritten = 0;
+ * there is an actual error condition we'll get it at the next try.
+ * We also check for aof-max-size limit here returning "no space" on exceed. */
+ssize_t aofWrite(int fd, const char *buf, size_t len, off_t aof_current_size, unsigned long long aof_max_size) {
+    ssize_t nwritten = 0, totwritten = 0, nonewritten = -1;
+
+    if (aof_max_size && (unsigned long long)aof_current_size >= aof_max_size) {
+        errno = ENOSPC;
```
xbasel: The error message:
`Error writing to the AOF file: <ENOSPC>`
is misleading if the issue is not actual disk space but hitting the configured max file size. Consider logging a more explicit message, and using `EFBIG` instead of `ENOSPC` for clarity.

kronwerk: I thought about smth like that, but haven't found a better error code than ENOSPC. the one you offered is (https://man7.org/linux/man-pages/man2/open.2.html):
`       EOVERFLOW
              pathname refers to a regular file that is too large to be
              opened.  The usual scenario here is that an application
              compiled on a 32-bit platform without
              -D_FILE_OFFSET_BITS=64 tried to open a file whose size
              exceeds (1<<31)-1 bytes; see also O_LARGEFILE above.  This
              is the error specified by POSIX.1; before Linux 2.6.24,
              Linux gave the error EFBIG for this case.`

seems that it's more misleading (imho) - we're not opening a file, we're writing to it, and we're exactly out of space (though the limit is not a real disk limit).

we can try to define our own errno type, a new one, to differ it from the real disk limit and form a more descriptive message based on that upper in code stack - but shouldn't that be an overkill for such an issue?

xbasel: POSIX error codes aren't tied to specific system/library calls and can be applied broadly. In the example above, it was in the context of `open`. Applications can use their own interpretations, this is an example:
https://github.com/valkey-io/valkey/blob/unstable/src/module.c#L5427

I'm fine with using either error codes, but there should be a custom error message to avoid misleading users into thinking it's a real disk space issue if ENOSPC is used.

PR #1426, opened 2024-12-11
RESP3 PUSH support for MONITOR command

### LINKED ISSUE #1428: [NEW] Support monitor command with RESP3 push type

Add support for MONITOR command for the RESP3 push type, so, instead of 
```
+1733939274.692107 [0 127.0.0.1:51014] "get" "a"
```
the connection after sending hello 3 and monitor will receive 
```
>2
$7
monitor
+1733939225.708882 [0 127.0.0.1:51014] "get" "a"
```

Original issue is from here https://github.com/antirez/RESP3/issues/21 - first support for redis was opened here https://github.com/redis/redis/pull/11216 . 
Now I am reopening it here :) I am not sure this implementation is correct (I am not C expert), in case of any comments I am happy to apply suggested changes.

PR with port: https://github.com/valkey-io/valkey/pull/1426/files

KowalczykBartek:
Change MONITOR in RESP3 mode to report monitored commands using the RESP3 push type, instead of as a stream of simple string replies.

This makes it possible for a client to send commands while it's in monitoring mode. It also makes it possible to use MONITOR with clients like hiredis, that handle push replies by delegating them to a push-callback provided by the user.

Old format, a stream of replies on the form

```
+1733939274.692107 [0 127.0.0.1:51014] "get" "a"
```

New format, after sending HELLO 3 and MONITOR, the connection will receive push messages on the form

```
>2
$7
monitor
+1733939225.708882 [0 127.0.0.1:51014] "get" "a"
```

Additional change: If MONITOR is called again when already in monitoring mode, the command replies with an error instead of no reply. A command with no reply confuses clients. They're expecting a reply to each command.

Closes #1428.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -705,10 +705,12 @@ void replicationFeedMonitors(client *c, list *monitors, int dictid, robj **argv,
     while ((ln = listNext(&li))) {
         client *monitor = ln->value;
         if (monitor->resp > 2) {
+            struct ClientFlags old_flags = monitor->flag;
             monitor->flag.pushing = 1;
             addReplyPushLen(monitor, 2);
             addReply(monitor, shared.monitorbulk);
             addReply(monitor, cmdobj);
+            if (!old_flags.pushing) monitor->flag.pushing = 0;
         } else {
             addReply(monitor, cmdobj);
         }
--- a/src/replication.c
+++ b/src/replication.c
@@ -704,7 +704,14 @@ void replicationFeedMonitors(client *c, list *monitors, int dictid, robj **argv,
     listRewind(monitors, &li);
     while ((ln = listNext(&li))) {
         client *monitor = ln->value;
-        addReply(monitor, cmdobj);
+        if(monitor->resp > 2) {
+            monitor->flag.pushing = 1;
+            addReplyPushLen(monitor,2);
+            addReply(monitor,shared.monitorbulk);
+            addReply(monitor,cmdobj);
+        } else {
+            addReply(monitor,cmdobj);
+        }
         updateClientMemUsageAndBucket(monitor);
     }
     decrRefCount(cmdobj);
--- a/src/replication.c
+++ b/src/replication.c
@@ -704,7 +704,7 @@ void replicationFeedMonitors(client *c, list *monitors, int dictid, robj **argv,
     listRewind(monitors, &li);
     while ((ln = listNext(&li))) {
         client *monitor = ln->value;
-        if(monitor->resp > 2) {
+        if (monitor->resp > 2) {
             monitor->flag.pushing = 1;
             addReplyPushLen(monitor,2);
             addReply(monitor,shared.monitorbulk);
--- a/src/replication.c
+++ b/src/replication.c
@@ -706,11 +706,11 @@ void replicationFeedMonitors(client *c, list *monitors, int dictid, robj **argv,
         client *monitor = ln->value;
         if (monitor->resp > 2) {
             monitor->flag.pushing = 1;
-            addReplyPushLen(monitor,2);
-            addReply(monitor,shared.monitorbulk);
-            addReply(monitor,cmdobj);
+            addReplyPushLen(monitor, 2);
+            addReply(monitor, shared.monitorbulk);
+            addReply(monitor, cmdobj);
         } else {
-            addReply(monitor,cmdobj);
+            addReply(monitor, cmdobj);
         }
         updateClientMemUsageAndBucket(monitor);
     }
--- a/src/server.c
+++ b/src/server.c
@@ -4259,10 +4259,10 @@ int processCommand(client *c) {
         return C_OK;
     }
 
-    /* Prevent a replica (but not a monitor client) from sending commands that access the keyspace.
+    /* Prevent a replica from sending commands that access the keyspace.
      * The main objective here is to prevent abuse of client pause check
      * from which replicas are exempt. */
-    if ((c->flag.replica && !c->flag.monitor) && (is_may_replicate_command || is_write_command || is_read_command)) {
+    if (c->flag.replica && (is_may_replicate_command || is_write_command || is_read_command)) {
         rejectCommandFormat(c, "Replica can't interact with the keyspace");
         return C_OK;
     }
@@ -6182,8 +6182,11 @@ void monitorCommand(client *c) {
         return;
     }
 
-    /* ignore MONITOR if already replica or in monitor mode */
-    if (c->flag.replica) return;
+    /* Gently notify the client that the monitor command has already been issued. */
+    if (c->flag.replica) {
+        addReplyError(c, "The connection is already in monitoring mode.");
+        return;
+    }
 
     c->flag.replica = 1;
     c->flag.monitor = 1;
--- a/src/server.c
+++ b/src/server.c
@@ -2035,6 +2035,7 @@ void createSharedObjects(void) {
     shared.ssubscribebulk = createStringObject("$10\r\nssubscribe\r\n", 17);
     shared.sunsubscribebulk = createStringObject("$12\r\nsunsubscribe\r\n", 19);
     shared.smessagebulk = createStringObject("$8\r\nsmessage\r\n", 14);
+    shared.monitorbulk = createStringObject("$7\r\nmonitor\r\n", 13);
     shared.psubscribebulk = createStringObject("$10\r\npsubscribe\r\n", 17);
     shared.punsubscribebulk = createStringObject("$12\r\npunsubscribe\r\n", 19);
 
--- a/src/server.c
+++ b/src/server.c
@@ -4259,10 +4259,10 @@ int processCommand(client *c) {
         return C_OK;
     }
 
-    /* Prevent a replica from sending commands that access the keyspace.
+    /* Prevent a replica (but not a monitor client) from sending commands that access the keyspace.
      * The main objective here is to prevent abuse of client pause check
      * from which replicas are exempt. */
-    if (c->flag.replica && (is_may_replicate_command || is_write_command || is_read_command)) {
+    if ((c->flag.replica && !c->flag.monitor) && (is_may_replicate_command || is_write_command || is_read_command)) {
         rejectCommandFormat(c, "Replica can't interact with the keyspace");
         return C_OK;
     }
--- a/tests/unit/introspection.tcl
+++ b/tests/unit/introspection.tcl
@@ -299,6 +299,20 @@ start_server {tags {"introspection"}} {
         set _ $res
     } {*"set" "foo"*"get" "foo"*}
 
+    test {MONITOR should support RESP3 protocol} {
+        set rd [valkey_deferring_client]
+        $rd HELLO 3
+        $rd read ; # Consume the HELLO reply
+
+        $rd monitor
+        $rd read ; # Consume the MONITOR reply
+
+        r set foo bar
+
+        assert_match {monitor*"set"*"foo"*"bar"*} [$rd read]
+        $rd close
+    }
+
     test {MONITOR can log commands issued by the scripting engine} {
         set rd [valkey_deferring_client]
         $rd monitor
--- a/tests/unit/introspection.tcl
+++ b/tests/unit/introspection.tcl
@@ -317,52 +317,20 @@ start_server {tags {"introspection"}} {
         $rd close
     }
 
-    test {MONITOR and CLIENT TRACKING should work on the same connection with RESP3} {
-        set rd1 [valkey_deferring_client]
-        set rd2 [valkey_deferring_client]
+    test {multiple MONITOR commands should result in ERR} {
+        set rd [valkey_deferring_client]
+        $rd HELLO 3
+        $rd read ; # Consume the HELLO reply
 
-        $rd1 HELLO 3
-        $rd1 read ; # Consume the HELLO reply
+        $rd readraw 1 ;
 
-        $rd1 client tracking on
-        $rd1 read ; # Consume the TRACKING reply
+        $rd monitor
+        assert_equal "+OK" [$rd read]
 
-        $rd1 monitor
-        $rd1 read ; # Consume the MONITOR reply
-
-        $rd1 set foo bar
-        assert_equal "OK" [$rd1 read]
-        assert_match {monitor*"set"*"foo"*"bar"*} [$rd1 read]
-
-        # Because we need to verify exact RESP3 response correctness,
-        # we need to instruct valkey client to return raw, unparsed response.
-        $rd1 readraw 1 ;
-
-        $rd1 get foo
-        assert_equal "\$3" [$rd1 read]
-        assert_equal "bar" [$rd1 read]
-
-        assert_equal ">2" [$rd1 read]
-        assert_equal "\$7" [$rd1 read]
-        assert_equal "monitor" [$rd1 read]
-        assert_match {*"get"*"foo"*} [$rd1 read]
-
-        $rd2 set foo baz
-
-        assert_equal ">2" [$rd1 read]
-        assert_equal "\$10" [$rd1 read]
-        assert_equal "invalidate" [$rd1 read]
-        assert_equal "*1" [$rd1 read]
-        assert_equal "\$3" [$rd1 read]
-        assert_equal "foo" [$rd1 read]
-
-        assert_equal ">2" [$rd1 read]
-        assert_equal "\$7" [$rd1 read]
-        assert_equal "monitor" [$rd1 read]
-        assert_match {*"set"*"foo"*"baz"*} [$rd1 read]
-
-        $rd1 close
-        $rd2 close
+        $rd monitor
+        assert_equal "-ERR The connection is already in monitoring mode." [$rd read]
+        
+        $rd close
     }
 
     test {MONITOR can log commands issued by the scripting engine} {
--- a/tests/unit/introspection.tcl
+++ b/tests/unit/introspection.tcl
@@ -306,7 +306,7 @@ start_server {tags {"introspection"}} {
 
         $rd monitor
         $rd read ; # Consume the MONITOR reply
-        $rd readraw 1 ;
+        $rd readraw 1;
 
         r set foo bar
         assert_equal ">2" [$rd read]
@@ -322,7 +322,7 @@ start_server {tags {"introspection"}} {
         $rd HELLO 3
         $rd read ; # Consume the HELLO reply
 
-        $rd readraw 1 ;
+        $rd readraw 1;
 
         $rd monitor
         assert_equal "+OK" [$rd read]
--- a/tests/unit/introspection.tcl
+++ b/tests/unit/introspection.tcl
@@ -306,11 +306,61 @@ start_server {tags {"introspection"}} {
 
         $rd monitor
         $rd read ; # Consume the MONITOR reply
+        $rd readraw 1 ;
 
         r set foo bar
-
-        assert_match {monitor*"set"*"foo"*"bar"*} [$rd read]
+        assert_equal ">2" [$rd read]
+        assert_equal "\$7" [$rd read]
+        assert_equal "monitor" [$rd read]
+        assert_match {*"set"*"foo"*"bar"*} [$rd read]
+        
         $rd close
+    }
+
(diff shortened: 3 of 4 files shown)


tests/unit/introspection.tcl
```diff
@@ -299,6 +299,20 @@ start_server {tags {"introspection"}} {
...
+        set rd [valkey_deferring_client]
+        $rd HELLO 3
+        $rd read ; # Consume the HELLO reply
+
+        $rd monitor
+        $rd read ; # Consume the MONITOR reply
+
+        r set foo bar
+
+        assert_match {monitor*"set"*"foo"*"bar"*} [$rd read]
```
zuiderkwast: To check that we actually get RESP3 pushes, we could use `$rd readraw 1` and verify that we get the correct RESP3 encoded replies.

We should probably also check that we can send regular commands (for example PING) while monitoring.

zuiderkwast (commented):
Hello! Nice!

Clang-format complains about missing space after `if` and after comma.

tests/unit/introspection.tcl
KowalczykBartek: > We should probably also check that we can send regular commands (for example PING) while monitoring.

that's was good idea, I think I found a bug (I am playing with my toy client [respy](https://github.com/KowalczykBartek/respy) ), let me fix your comment, write some test and describe behaviour with monitor+regular command.

btw, thanks for super fast reply, in redis repo it took 2years ;) 

EDIT: 
there is no bug in valkey but in my poor implementation (client), on PING, I received correct PONG and the monitor push 
```
+--------+-------------------------------------------------+----------------+
|00000000| 2b 50 4f 4e 47 0d 0a 3e 32 0d 0a 24 37 0d 0a 6d |+PONG..>2..$7..m|
|00000010| 6f 6e 69 74 6f 72 0d 0a 2b 31 37 33 33 39 35 38 |onitor..+1733958|
|00000020| 38 31 39 2e 38 39 37 37 36 34 20 5b 30 20 31 32 |819.897764 [0 12|
|00000030| 37 2e 30 2e 30 2e 31 3a 35 39 32 33 33 5d 20 22 |7.0.0.1:59233] "|
|00000040| 50 49 4e 47 22 0d 0a                            |PING"..         |
+--------+-------------------------------------------------+----------------+
```


Philosophical question: 
should this be allowed ?
```
HELLO 3 //resp3 on
CLIENT TRACKING on
MONITOR
GET somekey
```
so, two push operations on the single connection. I am asking because right now, I receive 
```
         +-------------------------------------------------+
         |  0  1  2  3  4  5  6  7  8  9  a  b  c  d  e  f |
+--------+-------------------------------------------------+----------------+
|00000000| 2d 45 52 52 20 52 65 70 6c 69 63 61 20 63 61 6e |-ERR Replica can|
|00000010| 27 74 20 69 6e 74 65 72 61 63 74 20 77 69 74 68 |'t interact with|
|00000020| 20 74 68 65 20 6b 65 79 73 70 61 63 65 0d 0a    | the keyspace.. |
+--------+-------------------------------------------------+----------------+
```
and tracking is silently closed

zuiderkwast: The answer to the philosophical question is just boring, not very philosophical. Monitor clients are flagged as replicas internally. I don't know why it's like that, but that's why you can't use GET. My guess is that in some old implementation it was the same code. We can change it if nothing breaks.

Can you try to find out by reading source code? If we don't flag monitor clients as replicas, do any test cases break?

KowalczykBartek:
```
Can you try to find out by reading source code? 
```
I will play around this and get back :)

KowalczykBartek: @zuiderkwast do you know maybe, if I can run some subset of tests, single unit tests file against my standalone redis ? Lets say I started my redis and want to run against it a tests/unit/introspection.tcl , is that possible ?

zuiderkwast: Yes it's possible.

`./runtest --help` is useful.

`--single unit/introspection` to run one test suite.

zuiderkwast: The test suite normally starts the server. If you want to run against an already started server, I think it's possible, but I don't know exactly how.

KowalczykBartek:
```
./runtest --port 6379 --host localhost --single unit/introspection
./src/valkey-server --enable-debug-command yes  
```
gave me what I wanted :)

KowalczykBartek: - I fixed formatting 
- used a readraw in existing test and added new for monitor+tracking combination

I also played around
```c
c->flag.replica = 0
```
but this resulted in failing tests because of 
```
------ STACK TRACE ------
EIP:
0   valkey-server                       0x000000010458ce40 clientHasPendingReplies + 200

Backtrace:
0   libsystem_platform.dylib            0x000000018ec60184 _sigtramp + 56
1   valkey-server                       0x000000010458ce2c clientHasPendingReplies + 180
2   valkey-server                       0x000000010458c51c prepareClientToWrite + 276
3   valkey-server                       0x000000010458ac14 addReply + 44
4   valkey-server                       0x00000001045a471c replicationFeedMonitors + 932
5   valkey-server                       0x000000010456b230 call + 1928
6   valkey-server                       0x000000010456cc20 processCommand + 3768
7   valkey-server                       0x0000000104585438 processInputBuffer + 516
8   valkey-server                       0x000000010458486c readQueryFromClient + 92
9   valkey-server                       0x000000010469c3a8 connSocketEventHandler + 180
10  valkey-server                       0x0000000104551ac0 aeProcessEvents + 372
11  valkey-server                       0x0000000104579fbc main + 19392
12  dyld                                0x000000018e8a8274 start + 2840
```
And I was unable to figure out what is wrong ... I used to Java's stacktraces :/ But I tried this approach https://github.com/valkey-io/valkey/pull/1426/commits/2c6304cee7693b1d6ad75a66cb6c8eb3bae108c9#diff-1abc5651133d108c0c420d9411925373c711133e7748d9e4f4c97d5fb543fdd9R4265 and it kinda works :D Unfortunately I am unable to say its correct or no

zuiderkwast:
You need to sign off the commits using `git commit -s`. See the description in CONTRIBUTING.md why we need this. See also the Details link at the "DCO" failing CI job for some help how to fix it.

tests/unit/introspection.tcl
```diff
@@ -299,6 +299,70 @@ start_server {tags {"introspection"}} {
...
+        $rd1 read ; # Consume the TRACKING reply
+
+        $rd1 monitor
+        $rd1 read ; # Consume the MONITOR reply
+
+        $rd1 set foo bar
+        assert_equal "OK" [$rd1 read]
+        assert_match {monitor*"set"*"foo"*"bar"*} [$rd1 read]
+
+        $rd1 readraw 1 ;
```
hpatro: Could you add a comment to this to indicate what it does.

src/server.c
```diff
@@ -4258,10 +4259,10 @@ int processCommand(client *c) {
         return C_OK;
     }
 
-    /* Prevent a replica from sending commands that access the keyspace.
+    /* Prevent a replica (but not a monitor client) from sending commands that access the keyspace.
      * The main objective here is to prevent abuse of client pause check
      * from which replicas are exempt. */
-    if (c->flag.replica && (is_may_replicate_command || is_write_command || is_read_command)) {
+    if ((c->flag.replica && !c->flag.monitor) && (is_may_replicate_command || is_write_command || is_read_command)) {
```
hpatro: I'm not sure about the implication from this change. But if we are planning to address this, shouldn't we not mark the client as replica and handle changes for monitor clients to get the replication stream?

tests/unit/introspection.tcl
KowalczykBartek: I pushed this, hope its good :) 
```
# Because we need to verify exact RESP3 response correctness,
# we need to instruct valkey client to return raw, unprased response.
```

src/server.c
KowalczykBartek:
```
shouldn't we not mark the client as replica and handle changes for monitor clients to get the replication stream?
```
I think that's too internal for me to answer :/

zuiderkwast: I agree with @hpatro. A replica can't send commands but a monitor client is not a replica. So, it's more logical if it's not flagged as a replica. Flagging it as areplica was probably an old hack to make it work.

I don't know how big changes we have to make...

hpatro: I would suggest us to stick with the existing behavior then and tackle this separately.

tests/unit/introspection.tcl
hpatro: *unparsed

KowalczykBartek: sorry :/ fixed

src/server.c
KowalczykBartek: so I will revert this change with if-statement, so just a monitor without tracking will be possible for now, is that ok?

hpatro: I'm curious. Could you explain how/why do you need tracking with monitor functionality?

In past I've used MONITOR just for debugging purposes.

KowalczykBartek: I don't honestly (I found this by accident) :D I am just following @zuiderkwast https://github.com/valkey-io/valkey/pull/1426#discussion_r1881776807 :) For me its perfectly fine (the only implication is that, you can connection which cannot be reused for ordinary queries)

zuiderkwast: Yes, let's skip this change for now. We can start with enabling monitor clients to use commands like PING, INFO and others.

I don't think there's a use case for calling commands while monitoring, but it's just an arbitrary limitation.

The main point of this PR is to allow RESP3 clients to use monitor and delegate the push messages to the user via some callback. It can already be used in hiredis, for example, where you can specify a callback that gets all pushes.

Another thing that confuses client libraries is that when calling MONITOR when already in monitor mode, you don't get a reply at all, so clients like hiredis would hang waiting for a reply. I would like an error like "already monitoring".

tests/unit/introspection.tcl
zuiderkwast: I think you can use `addr2line` to get the line numbers.

src/server.c
KowalczykBartek: So, I fixed comments (I hope), right now multiple MONITOR will result in 
```
-ERR The connection is already in monitoring mode
```

zuiderkwast (approved):
LGTM

@valkey-io/core-team Please approve (or not) the changed behaviour of MONITOR described in the top comment.

It's a potentially breaking change (unless we consider it a bugfix?) so we should release it in the next major release, Valkey 9, right?

hpatro (approved):
LGTM

tests/unit/introspection.tcl
```diff
@@ -299,6 +299,40 @@ start_server {tags {"introspection"}} {
...
+        
+        $rd close
+    }
+
+    test {multiple MONITOR commands should result in ERR} {
+        set rd [valkey_deferring_client]
+        $rd HELLO 3
+        $rd read ; # Consume the HELLO reply
+
+        $rd readraw 1 ;
```
enjoy-binbin:
```suggestion
        $rd readraw 1
```

tests/unit/introspection.tcl
```diff
@@ -299,6 +299,40 @@ start_server {tags {"introspection"}} {
...
     } {*"set" "foo"*"get" "foo"*}
 
+    test {MONITOR should support RESP3 protocol} {
+        set rd [valkey_deferring_client]
+        $rd HELLO 3
+        $rd read ; # Consume the HELLO reply
+
+        $rd monitor
+        $rd read ; # Consume the MONITOR reply
+        $rd readraw 1 ;
```
enjoy-binbin:
```suggestion
        $rd readraw 1
```

src/replication.c
```diff
@@ -704,7 +704,14 @@ void replicationFeedMonitors(client *c, list *monitors, int dictid, robj **argv,
     listRewind(monitors, &li);
     while ((ln = listNext(&li))) {
         client *monitor = ln->value;
-        addReply(monitor, cmdobj);
+        if (monitor->resp > 2) {
+            monitor->flag.pushing = 1;
```
enjoy-binbin, line 709: we should undo the pushing flag in the last?

enjoy-binbin (approved):

src/replication.c
KowalczykBartek:
```
Logged crash report (pid 67673):
67673:M 16 Dec 2024 12:11:36.111 # === ASSERTION FAILED CLIENT CONTEXT ===
67673:M 16 Dec 2024 12:11:36.111 # client->flags = 18014398509481990
67673:M 16 Dec 2024 12:11:36.111 # client->conn = fd=13
67673:M 16 Dec 2024 12:11:36.111 # client->argc = 0
67673:M 16 Dec 2024 12:11:36.111 # === RECURSIVE ASSERTION FAILED ===
67673:M 16 Dec 2024 12:11:36.111 # ==> networking.c:1087 'c->flag.pushing' is not true
```
without this flag set instance is crashing - I just saw its used everywhere when push is used 
```
void addReplyPubsubMessage(client *c, robj *channel, robj *msg, robj *message_bulk) {
    struct ClientFlags old_flags = c->flag;
    c->flag.pushing = 1;
    if (c->resp == 2)
        addReply(c, shared.mbulkhdr[3]);
```
I did the same.

tests/unit/introspection.tcl
KowalczykBartek: fixed.

tests/unit/introspection.tcl
KowalczykBartek: fixed.

src/replication.c
zuiderkwast: @enjoy-binbin Well spotted! Yes, we must reset it, right? Otherwise there will be some problem if the client sends commands. If the pushing flag is set, the replies are queued and not sent to the client until later.

Btw, @KowalczykBartek you didn't add a test case where the client sends a PING? We should add it and the test case needs to check that the pong-reply and monitored-ping push come in the right order. The pong reply should come first.

This is the code in networking.c that cares about this flag:

```C
    /* If we're processing a push message into the current client (i.e. executing PUBLISH
     * to a channel which we are subscribed to, then we wanna postpone that message to be added
     * after the command's reply (specifically important during multi-exec). the exception is
     * the SUBSCRIBE command family, which (currently) have a push message instead of a proper reply.
     * The check for executing_client also avoids affecting push messages that are part of eviction.
     * Check CLIENT_PUSHING first to avoid race conditions, as it's absent in module's fake client. */
    if (c->flag.pushing && c == server.current_client && server.executing_client &&
        !cmdHasPushAsReply(server.executing_client->cmd)) {
        _addReplyProtoToList(c, server.pending_push_messages, s, len);
        return;
    }
```

zuiderkwast: @KowalczykBartek It seem you misunderstood the comment. We should *reset the flag to its previous value after pushing*.

```
int old_pushing = c->flags.pushing;
c->flags.pushing = 1;
push stuff...
c->flags.pushing = old_pushing;
```

KowalczykBartek: I pushed both if statement for flag reset and test but I faced some issue with tcl, that's what valkey returns (I see this with my toy client) for combination of monitor + ping
```
+PONG
>2
$7
monitor
+1734353183.716803 [0 127.0.0.1:50339] "PING"
```
but client in test doesn't see this PONG :/ 

```tlc

    test {MONITOR should came after PONG reply} {
        $rd HELLO 3
        $rd read ; # Consume the HELLO reply

        $rd monitor
        $rd read ; # Consume the MONITOR reply
        $rd readraw 1;

        r ping

        assert_equal "+pong" [$rd read]
        assert_equal ">2" [$rd read]
        assert_equal "\$7" [$rd read]
        assert_equal "monitor" [$rd read]
        assert_match {*"ping"*} [$rd read]

        $rd close
    }
```
in this line 
```
 assert_equal "+pong" [$rd read]
```
it see `>2`, so, ping is skipped because of some reason(test is failing, but should not), do you have any idea what is going on ? I even run 
```
./src/valkey-server --enable-debug-command yes
```
against standalone server with another client listening on MONITOR, but tests are not "monitorable", I am unable to see executed commands :/

zuiderkwast: `r` is one client and `$rd` is another. You can see the PONG as the return value of `r ping`:

    assert_equal pong [r ping]

... but what we need to test here is to send the command from the monitoring client, i.e. using `$rd ping`.

tests/unit/introspection.tcl
```diff
@@ -333,6 +333,26 @@ start_server {tags {"introspection"}} {
...
+    test {MONITOR should came after PONG reply} {
+        set rd [valkey_deferring_client]
+        $rd HELLO 3
+        $rd read ; # Consume the HELLO reply
+
+        $rd monitor
+        $rd read ; # Consume the MONITOR reply
+        $rd readraw 1;
+
+        r ping
```
zuiderkwast:
```suggestion
        $rd ping
```

src/replication.c
KowalczykBartek: yeah.. I am blind, sorry for this :/ I pushed fixed test case

tests/unit/introspection.tcl
KowalczykBartek: fixed

madolson:
> It's a potentially breaking change (unless we consider it a bugfix?) so we should release it in the next major release, Valkey 9, right?

It's definitely a breaking change and definitely not a bug fix. Can we implement this not as a breaking change, maybe with an optional argument? I am very much against breaking changes when possible.

zuiderkwast:
> > It's a potentially breaking change (unless we consider it a bugfix?) so we should release it in the next major release, Valkey 9, right?
> 
> It's definitely a breaking change and definitely not a bug fix. Can we implement this not as a breaking change, maybe with an optional argument? I am very much against breaking changes when possible.

It was meant as a rhetorical question. :) I did add the label.

> I am very much against breaking changes when possible.

I am too. Why do I keep forgetting?

I guess there are two reasons why I thought this is OK:
1. I assumed nobody uses HELLO 3 and MONITOR together, because it doesn't work the way it should. (Here, I'm probably wrong. Clients do all sorts of weird things.)
2. I think this is an error in how RESP3 is used, yes a bug. RESP3 was supposed to solve the request-response model by adding the push type to be used for any reply that is not a one-to-one match to a command. As long as there are commands that don't follow this (MONITOR and all the SUBSCRIBE commands), then clients can't rely on this without special logic for each command. Maybe the opt-in we should use is `hello 4`, a "RESP bug fix release", which would also fix the subscribe commands to return an in-band +OK too. ...

KowalczykBartek:
Reading the opinions and feelings I think its more a bug than a feature :D 
https://github.com/antirez/RESP3/issues/21
https://github.com/redis/redis/issues/6426#issuecomment-1485059034
:) 

But that's not a problem for me to push this flag's support (of course if I will be able to implement this :)), please just write me how it should looks like

PingXie:
This is a good change but it is indeed breaking. Also I am curious about the rationale behind tying this behavior to REPS 3? Can we consider opt'ing in via a new `CLIENT CAPA` instead?

zuiderkwast:
> This is a good change but it is indeed breaking. Also I am curious about the rationale behind tying this behavior to REPS 3? Can we consider opt'ing in via a new `CLIENT CAPA` instead?

RESP2 doesn't have pushes.... It only makes sense for RESP3 and RESP3 is already opt-in.

I can see a few ways to proceed:

1. Opt-in using client capa (e.g. `HELLO 3, CLIENT CAPA monitor-pushes, MONITOR`)
2. Opt-in using hello (e.g. `HELLO 4, MONITOR`)
3. Opt-in using monitor parameter (e.g. `HELLO 3, MONITOR PUSH`)
4. Just do the breaking change (`HELLO 3, MONITOR`)

A CLIENT CAPA just for modifying the behavior of one command seems odd. Then, an optional parameter to MONITOR seems more clean.

PingXie:
Thanks @zuiderkwast. Yeah, "client capa" is over kill. I like option 3.

madolson:
Just to avoid leaving this dangling. We chatted about this briefly today and we are concerned about the breaking change. If someone really wants to have something like monitor that works well with clients, we would like to build some new system just for that. @KowalczykBartek If you are still interested, we can chat about it, otherwise I'm just going to close it.

PR #1427, opened 2024-12-11
Replace dict with new hashtable: sorted set datatype

### LINKED ISSUE #1096: Use new hashtable for sorted sets

rainsupreme:
This PR replaces dict with hashtable in the ZSET datatype. Instead of mapping key to score as dict did, the hashtable maps key to a node in the skiplist, which contains the score. This takes advantage of hashtable performance improvements and saves 15 bytes per set item - 24 bytes overhead before, 9 bytes after.

Closes #1096

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -1026,7 +1026,7 @@ void hashtableScanCallback(void *privdata, void *entry) {
     serverAssert(o != NULL);
 
     /* get key */
-    if (o->type == OBJ_SET && o->encoding == OBJ_ENCODING_HASHTABLE) {
+    if (o->type == OBJ_SET) {
         key = (sds)entry;
     } else if (o->type == OBJ_ZSET) {
         zskiplistNode *node = (zskiplistNode *)entry;
@@ -1042,7 +1042,7 @@ void hashtableScanCallback(void *privdata, void *entry) {
         }
     }
 
-    if (o->type == OBJ_SET && o->encoding == OBJ_ENCODING_HASHTABLE) {
+    if (o->type == OBJ_SET) {
         /* no value, key used by reference */
     } else if (o->type == OBJ_ZSET) {
         /* zset data is copied */
--- a/src/db.c
+++ b/src/db.c
@@ -1004,13 +1004,6 @@ void dictScanCallback(void *privdata, const dictEntry *de) {
         if (!data->only_keys) {
             val = dictGetVal(de);
         }
-    } else if (o->type == OBJ_ZSET) {
-        key = sdsdup(keysds);
-        if (!data->only_keys) {
-            char buf[MAX_LONG_DOUBLE_CHARS];
-            int len = ld2string(buf, sizeof(buf), *(double *)dictGetVal(de), LD_STR_AUTO);
-            val = sdsnewlen(buf, len);
-        }
     } else {
         serverPanic("Type not handled in dict SCAN callback.");
     }
@@ -1021,13 +1014,26 @@ void dictScanCallback(void *privdata, const dictEntry *de) {
 
 void hashtableScanCallback(void *privdata, void *entry) {
     scanData *data = (scanData *)privdata;
+    sds val = NULL;
+    sds key = NULL;
+
     robj *o = data->o;
     list *keys = data->keys;
     data->sampled++;
 
-    /* currently only implemented for SET scan */
-    serverAssert(o && o->type == OBJ_SET && o->encoding == OBJ_ENCODING_HASHTABLE);
-    sds key = (sds)entry; /* Specific for OBJ_SET */
+    /* This callback is only used for scanning elements within a key (hash
+     * fields, set elements, etc.) so o must be set here. */
+    serverAssert(o != NULL);
+
+    /* get key */
+    if (o->type == OBJ_SET && o->encoding == OBJ_ENCODING_HASHTABLE) {
+        key = (sds)entry;
+    } else if (o->type == OBJ_ZSET) {
+        zskiplistNode *node = (zskiplistNode *)entry;
+        key = node->ele;
+    } else {
+        serverPanic("Type not handled in hashset SCAN callback.");
+    }
 
     /* Filter element if it does not match the pattern. */
     if (data->pattern) {
@@ -1036,7 +1042,23 @@ void hashtableScanCallback(void *privdata, void *entry) {
         }
     }
 
+    if (o->type == OBJ_SET && o->encoding == OBJ_ENCODING_HASHTABLE) {
+        /* no value, key used by reference */
+    } else if (o->type == OBJ_ZSET) {
+        /* zset data is copied */
+        zskiplistNode *node = (zskiplistNode *)entry;
+        key = sdsdup(node->ele);
+        if (!data->only_keys) {
+            char buf[MAX_LONG_DOUBLE_CHARS];
+            int len = ld2string(buf, sizeof(buf), node->score, LD_STR_AUTO);
+            val = sdsnewlen(buf, len);
+        }
+    } else {
+        serverPanic("Type not handled in hashset SCAN callback.");
+    }
+
     listAddNodeTail(keys, key);
+    if (val) listAddNodeTail(keys, val);
 }
 
 /* Try to parse a SCAN cursor stored at object 'o':
@@ -1184,7 +1206,7 @@ void scanGenericCommand(client *c, robj *o, unsigned long long cursor) {
         shallow_copied_list_items = 1;
     } else if (o->type == OBJ_ZSET && o->encoding == OBJ_ENCODING_SKIPLIST) {
         zset *zs = o->ptr;
-        dict_table = zs->dict;
+        hashtable_table = zs->ht;
         /* scanning ZSET allocates temporary strings even though it's a dict */
         shallow_copied_list_items = 0;
     }
--- a/src/defrag.c
+++ b/src/defrag.c
@@ -298,54 +298,46 @@ static void zslUpdateNode(zskiplist *zsl, zskiplistNode *oldnode, zskiplistNode 
 }
 
 /* Defrag helper for sorted set.
- * Update the robj pointer, defrag the skiplist struct and return the new score
- * reference. We may not access oldele pointer (not even the pointer stored in
- * the skiplist), as it was already freed. Newele may be null, in which case we
- * only need to defrag the skiplist, but not update the obj pointer.
- * When return value is non-NULL, it is the score reference that must be updated
- * in the dict record. */
-static double *zslDefrag(zskiplist *zsl, double score, sds oldele, sds newele) {
-    zskiplistNode *update[ZSKIPLIST_MAXLEVEL], *x, *newx;
-    int i;
-    sds ele = newele ? newele : oldele;
+ * Defragment a single skiplist node, update skiplist pointers, and update the
+ * hashtable pointer to the node */
+static void activeDefragZsetNode(void *privdata, void *entry_ref) {
+    zskiplist *zsl = privdata;
+    zskiplistNode **node_ref = (zskiplistNode **)entry_ref;
 
-    /* find the skiplist node referring to the object that was moved,
-     * and all pointers that need to be updated if we'll end up moving the skiplist node. */
-    x = zsl->header;
-    for (i = zsl->level - 1; i >= 0; i--) {
-        while (x->level[i].forward && x->level[i].forward->ele != oldele && /* make sure not to access the
-                                                                               ->obj pointer if it matches
-                                                                               oldele */
-               (x->level[i].forward->score < score ||
-                (x->level[i].forward->score == score && sdscmp(x->level[i].forward->ele, ele) < 0)))
+    /* defragment node internals */
+    sds newsds = activeDefragSds((*node_ref)->ele);
+    if (newsds) (*node_ref)->ele = newsds;
+
+    const double score = (*node_ref)->score;
+    const sds ele = (*node_ref)->ele;
+
+    /* find skiplist pointers that need to be updated if we end up moving the
+     * skiplist node. */
+    zskiplistNode *update[ZSKIPLIST_MAXLEVEL];
+    zskiplistNode *x = zsl->header;
+    for (int i = zsl->level - 1; i >= 0; i--) {
+        while (1) {
+            /* stop when we've reached the end of this level or the next node
+             * comes after our target in sorted order */
+            zskiplistNode *next = x->level[i].forward;
+            if (!next) break;
+            if (next->score > score) break;
+            if (next->score == score && sdscmp(next->ele, ele) >= 0) {
+                break;
+            }
             x = x->level[i].forward;
+        }
         update[i] = x;
     }
-
-    /* update the robj pointer inside the skip list record. */
     x = x->level[0].forward;
-    serverAssert(x && score == x->score && x->ele == oldele);
-    if (newele) x->ele = newele;
+    /* should have arrived at intended node */
+    serverAssert(x == *node_ref);
 
     /* try to defrag the skiplist record itself */
-    newx = activeDefragAlloc(x);
+    zskiplistNode *newx = activeDefragAlloc(x);
     if (newx) {
         zslUpdateNode(zsl, x, newx, update);
-        return &newx->score;
-    }
-    return NULL;
-}
-
-/* Defrag helper for sorted set.
- * Defrag a single dict entry key name, and corresponding skiplist struct */
-static void activeDefragZsetEntry(zset *zs, dictEntry *de) {
-    sds newsds;
-    double *newscore;
-    sds sdsele = dictGetKey(de);
-    if ((newsds = activeDefragSds(sdsele))) dictSetKey(zs->dict, de, newsds);
-    newscore = zslDefrag(zs->zsl, *(double *)dictGetVal(de), sdsele, newsds);
-    if (newscore) {
-        dictSetVal(zs->dict, de, newscore);
+        *node_ref = newx; /* update hashtable pointer */
     }
 }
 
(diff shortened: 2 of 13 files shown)


zuiderkwast:
> This PR is ready for review! Needed changes have been merged and I've rebased

Great! You could mark it as not draft then. :) We use the top comment and PR title for the final commit message when it gets merged, so you could update those to concisely describe the change (i.e. like a commit message).

It's late in my time zone so I'll look tomorrow.

src/db.c
```diff
@@ -1020,13 +1013,26 @@ void dictScanCallback(void *privdata, const dictEntry *de) {
...
-    serverAssert(o && o->type == OBJ_SET && o->encoding == OBJ_ENCODING_HASHTABLE);
-    sds key = (sds)entry; /* Specific for OBJ_SET */
+    /* This callback is only used for scanning elements within a key (hash
+     * fields, set elements, etc.) so o must be set here. */
+    serverAssert(o != NULL);
+
+    /* get key */
+    if (o->type == OBJ_SET && o->encoding == OBJ_ENCODING_HASHTABLE) {
+        key = (sds)entry;
+    } else if (o->type == OBJ_ZSET) {
```
zuiderkwast: Why check the encoding for sets but not for sorted-sets?

I think we can skip it in both cases, right? Also the similar `if` a few lines below.

src/t_zset.c
```diff
@@ -1566,29 +1556,17 @@ int zsetAdd(robj *zobj, double score, sds ele, int in_flags, int *out_flags, dou
  * element was not there). It does not resize the dict after deleting the
  * element. */
```
zuiderkwast: This comment is for zsetRemoveFromSkiplist. It should be updated to not say dict.

The comment also says "It does not resize the dict after deleting the element". Is this correct? I think hashtablePop can shrink the hashtable, but I don't remember exactly.

src/t_zset.c
```diff
@@ -2585,6 +2553,22 @@ static void zdiff(zsetopsrc *src, long setnum, zset *dstzset, size_t *maxelelen,
...
+const void *sdsDoubleMapGetKey(const void *entry) {
+    struct sdsDoublePair *keyval = (struct sdsDoublePair *)entry;
+    return keyval->key;
+}
+
+hashtableType sds_double_map_hashtable = {
+    .hashFunction = dictSdsHash,
+    .entryGetKey = sdsDoubleMapGetKey,
+    .keyCompare = hashtableSdsKeyCompare,
+    .entryDestructor = zfree};
```
zuiderkwast: Can you add a few lines of comments about what these are used for?

temporary stuff for union/intersection/diff bla bla bla...

src/t_zset.c
```diff
@@ -2585,6 +2553,22 @@ static void zdiff(zsetopsrc *src, long setnum, zset *dstzset, size_t *maxelelen,
     }
 }
 
+struct sdsDoublePair {
+    sds key;
+    double score;
+};
+
+const void *sdsDoubleMapGetKey(const void *entry) {
+    struct sdsDoublePair *keyval = (struct sdsDoublePair *)entry;
```
zuiderkwast: Does this work with an implicit cast? Maybe add `const` to the struct to avoid warning when casting from const to non-const?

```suggestion
    const struct sdsDoublePair *keyval = entry;
```

src/defrag.c
```diff
@@ -297,54 +297,46 @@ static void zslUpdateNode(zskiplist *zsl, zskiplistNode *oldnode, zskiplistNode
...
-    x = zsl->header;
-    for (i = zsl->level - 1; i >= 0; i--) {
-        while (x->level[i].forward && x->level[i].forward->ele != oldele && /* make sure not to access the
-                                                                               ->obj pointer if it matches
-                                                                               oldele */
-               (x->level[i].forward->score < score ||
-                (x->level[i].forward->score == score && sdscmp(x->level[i].forward->ele, ele) < 0)))
+ * Defragment a single skiplist node, update skiplist pointers, and update the
+ * hashtable pointer to the node */
+static void activeDefragZsetNode(void *privdata, void *entry_ref) {
```
zuiderkwast, line 303: Can you mention in the comment that this is the hashtableScan callback?

src/server.c
```diff
@@ -555,14 +555,16 @@ hashtableType setHashtableType = {
     .keyCompare = hashtableSdsKeyCompare,
     .entryDestructor = dictSdsDestructor};
 
+const void *zsetHashtableGetKey(const void *element) {
+    zskiplistNode *node = (zskiplistNode *)element;
```
zuiderkwast: Implicit cast should work if we keep `const`?

```suggestion
    const zskiplistNode *node = element;
```

src/t_zset.c
```diff
@@ -1504,18 +1498,16 @@ int zsetAdd(robj *zobj, double score, sds ele, int in_flags, int *out_flags, dou
      * converted the key to skiplist. */
     if (zobj->encoding == OBJ_ENCODING_SKIPLIST) {
         zset *zs = zobj->ptr;
-        zskiplistNode *znode;
-        dictEntry *de;
 
-        de = dictFind(zs->dict, ele);
-        if (de != NULL) {
+        zskiplistNode **existing_node = (zskiplistNode **)hashtableFindRef(zs->ht, ele);
```
zuiderkwast: `existing_node` is not really node. Should we call it `node_ptr` or `node_ref` or something?

I'm wondering if we can also avoid the cast between `void **` and `zskiplistNode **`.

How about this style?

```C
    void **node_ref_in_hashtable = hashtableFindRef(zs->ht, ele);
    zskiplistNode *existing_node = *node_ref_in_hashtable;

    ...

    /* Note that this assignment updates the node pointer stored in
     * the hashtable */
    *node_ref_in_hashtable = zslUpdateScore(zs->zsl, curscore, ele, score);
```

zuiderkwast (commented):
This looks awesome. Just a few nits.

src/db.c
ranshid: So basically the condition which was removed on line 1028 should only include check for the encoding.

serverAssert(o && o->encoding == OBJ_ENCODING_HASHTABLE);

src/defrag.c
```diff
@@ -297,54 +297,46 @@ static void zslUpdateNode(zskiplist *zsl, zskiplistNode *oldnode, zskiplistNode
...
+
+    const double score = (*node_ref)->score;
+    const sds ele = (*node_ref)->ele;
+
+    /* find skiplist pointers that need to be updated if we end up moving the
+     * skiplist node. */
+    zskiplistNode *update[ZSKIPLIST_MAXLEVEL];
+    zskiplistNode *x = zsl->header;
+    for (int i = zsl->level - 1; i >= 0; i--) {
+        while (1) {
```
ranshid: In general I prefer avoid while(1) loops. is there a very urgent need to do it here? and not keep the old condition (minus the oldelem thing)?

src/t_zset.c
```diff
@@ -2762,73 +2745,73 @@ void zunionInterDiffGenericCommand(client *c, robj *dstkey, int numkeysIndex, in
...
+        hashtableExpand(dstzset->ht, hashtableSize(temp_ht));
+        hashtableIterator iter;
+        hashtableInitIterator(&iter, temp_ht);
+
+        void *next;
+        while (hashtableNext(&iter, &next)) {
+            struct sdsDoublePair *element = next;
+            zskiplistNode *znode = zslInsert(dstzset->zsl, element->score, element->key);
+            element->key = NULL; /* passed ownership to skiplist node */
+            hashtableAdd(dstzset->ht, znode);
```
ranshid: I feel uncomfortable with this part. Did we check the perf impact of this copy operation? unions can sometimes introduce sets with millions of elements. and we are also using twice the hashtable memory for maintaining this. 
I am not sure why we can't just create the skiplist nodes (based on random heights matching the current expansion) and insert them to the skiplist after we complete and scan the hashtable.
Or maybe I missed something?

ranshid (commented):
great work @SoftlyRaining !

Some minor comments and I feel somewhat uncomfortable about the way we implemented  the union.

src/t_zset.c
zuiderkwast: Good point Ran. The dict solution could reuse the dict for the final sorted set, so this temporary hashtable can be a problem.

Why is this union algorithm even split in to two steps (1) create the hash table and (2) build the sorted set skiplist? Why don't we combine both in one step and build the skiplist + hashtable together?

zuiderkwast: I missed that updating the scores is much cheaper if we haven't yet built the skiplist. Ran, your idea sounds good, if I understand it correctly. In step 1, should we create skiplist nodes without initializing the pointers? Then in step 2 we link the nodes together? We'd have to refactor zslInsert to separate the node creation and the insertion into the skiplist?

ranshid: yes that is basically what I meant

src/db.c
rainsupreme: Ah, yes. Rather it should only check the type - we already know it uses hashtable encoding because we're in a hashtable callback, but we use the callback with a couple different types now.

src/t_zset.c
rainsupreme: You're right, it does shrink if needed (unless shrinking is paused presumably).

src/t_zset.c
rainsupreme: Yes, I think that's much better

src/defrag.c
rainsupreme: I thought the exit conditions formatted this way were more readable. I'll see if I can rearrange it and keep it readable.

src/t_zset.c
rainsupreme: I guess I'm not sure what you mean when you say based on random heights, but I think I get the idea. Yes, that would be more efficient. I'll take a look at making this change.

zuiderkwast: > I guess I'm not sure what you mean when you say based on random heights

When a skiplist node is created, it always gets a random height, i.e. a random number of levels with pointers to other nodes. So, just "create a skiplist node" implies it has some random height.

src/t_zset.c
```diff
@@ -296,17 +314,15 @@ int zslDelete(zskiplist *zsl, double score, sds ele, zskiplistNode **node) {
...
     int i;
 
     /* We need to seek to element to update to start: this is useful anyway,
      * we'll have to update or remove it. */
     x = zsl->header;
     for (i = zsl->level - 1; i >= 0; i--) {
-        while (x->level[i].forward &&
-               (x->level[i].forward->score < curscore ||
-                (x->level[i].forward->score == curscore && sdscmp(x->level[i].forward->ele, ele) < 0))) {
+        while (compareNodeScoreEle(x->level[i].forward, curscore, ele) < 0) {
```
ranshid: I think it would make much more sense to pass the node to this function. the only case we are calling this is when we use zadd in order to change an element's score. when we have a direct pointer to the node we can just skip this search.

src/t_zset.c
```diff
@@ -163,75 +163,95 @@ void zslFree(zskiplist *zsl) {
...
-/* Insert a new node in the skiplist. Assumes the element does not already
- * exist (up to the caller to enforce that). The skiplist takes ownership
- * of the passed SDS string 'ele'. */
-zskiplistNode *zslInsert(zskiplist *zsl, double score, sds ele) {
-    zskiplistNode *update[ZSKIPLIST_MAXLEVEL], *x;
+/* Compares node and score/ele; defines zset ordering. Return value:
+ *     positive if node comes after score/ele.
+ *     negative if node comes before score/ele.
+ *     0 if node's score and ele are both equal to those provided. */
+static int compareNodeScoreEle(const zskiplistNode *node, const double score, const sds ele) {
```
ranshid: I would suggest:
1. to stick to the current function naming convention eg: zslCompareNode...
2. I think in all cases we are using this API we can change the calling function to get a node instead of a score and sds pointer. this  might simplify the API.
I would suggest making this: int zslCompareNodes(const zskiplistNode *n1, const zskiplistNode *n2)

src/t_zset.c
ranshid: Oh. Sorry not true, ignore. Forgot we must scan to take the update path in order to delete the node.
However we could optimize by searching the update path only in case a node removal/addition is really needed

src/t_zset.c
rainsupreme: Good idea! I took a look and was able to delete zslGetRank :)

src/t_zset.c
zuiderkwast: If it's separate and non-trivial, let's do it as follow-up?

src/t_zset.c
ranshid: Yeh, although I am not sure in this case it would be optimal :) 
The reason I kept zslGetRank is that I had  (a non proven) suspicion that using the zslGetRank in the zcount would be more "cache friendly" since we are taking the route to the node in zslNthInRange while zslGetRankByNode will take the continued route to the end of the skiplist which is probably NOT in cache. lets make sure if we are not impacting performance here. we can probably change the zslGetRankByNode to something else like zslGetRankFromNode or something IDK

src/defrag.c
```diff
@@ -296,55 +296,45 @@ static void zslUpdateNode(zskiplist *zsl, zskiplistNode *oldnode, zskiplistNode
...
+static void activeDefragZsetNode(void *privdata, void *entry_ref) {
+    zskiplist *zsl = privdata;
+    zskiplistNode **node_ref = (zskiplistNode **)entry_ref;
+
+    /* defragment node internals */
+    sds newsds = activeDefragSds((*node_ref)->ele);
+    if (newsds) (*node_ref)->ele = newsds;
+
+    const double score = (*node_ref)->score;
+    const sds ele = (*node_ref)->ele;
```
zuiderkwast: It looks as if we can use a `node` variable.
```suggestion
    zskiplistNode **node_ref = (zskiplistNode **)entry_ref;
    zskiplistNode *node = *node_ref;

    /* defragment node internals */
    sds newsds = activeDefragSds(node->ele);
    if (newsds) node->ele = newsds;

    const double score = node->score;
    const sds ele = node->ele;
```

zuiderkwast (commented):
Just fix the naming converion (compareNodeScoreEle -> zslCompareNodeScoreEle, etc.). Then I think it's good to merge.

I don't really get why you want to delete `zslGetRank`. If it's an optimization beyond the replacement of dict, then it can as well be a follow up, right? As you want.

rainsupreme:
I poked around and found that all uses of zslGetRank could more efficiently use zslGetRankByNode, and it became dead code. I already made the revision at any rate but you're right, it might've been a separate small cleanup PR 😅

ranshid:
> I poked around and found that all uses of zslGetRank could more efficiently use zslGetRankByNode, and it became dead code. I already made the revision at any rate but you're right, it might've been a separate small cleanup PR 😅

@SoftlyRaining My only concern is a potential small degradation in performance of zcount and zlexcount. can we just make sure to verify we have no impact on these operations?

rainsupreme:
> > I poked around and found that all uses of zslGetRank could more efficiently use zslGetRankByNode, and it became dead code. I already made the revision at any rate but you're right, it might've been a separate small cleanup PR 😅
> 
> @SoftlyRaining My only concern is a potential small degradation in performance of zcount and zlexcount. can we just make sure to verify we have no impact on these operations?

I will remove that aspect of the change so it can be more thoroughly investigated and benchmarked as a separate PR. I'd prefer to avoid delaying the core hashtable work. :)

zuiderkwast (approved):
LGTM

@ranshid The zslGetRank changes have been removed, so I guess it's safe to merge. WDYT?

src/t_zset.c
```diff
@@ -250,91 +273,62 @@ void zslDeleteNode(zskiplist *zsl, zskiplistNode *x, zskiplistNode **update) {
...
+ * element, which is more costly. A pointer to the new node is returned. */
+static zskiplistNode *zslUpdateScore(zskiplist *zsl, zskiplistNode *node, double newscore) {
+    /* If the node, after the score update, would be still exactly
+     * at the same position, we can just update the score without
+     * actually removing and re-inserting the element in the skiplist. */
+    if ((node->backward == NULL || node->backward->score < newscore) &&
+        (node->level[0].forward == NULL || node->level[0].forward->score > newscore)) {
+        node->score = newscore;
+        return NULL;
+    }
```
ranshid, line 312: Small mark: it is still possible that a node position change will NOT take place after this check. For example in case we update the score to something that exactly matches the score of the prev or next node. 
The check can be extended to check also equality of the score (but will also need to compare the key order).
I guess this is fine for now, but maybe extend the comment above to explain that?

ranshid (approved):
Small comment which I think we could skip handling for now and maybe only extend the comment. 
I did not rescan the entire change so LGTM

src/t_zset.c
```diff
@@ -250,91 +273,62 @@ void zslDeleteNode(zskiplist *zsl, zskiplistNode *x, zskiplistNode **update) {
...
- *
- * The function returns the updated element skiplist node pointer. */
-zskiplistNode *zslUpdateScore(zskiplist *zsl, double curscore, sds ele, double newscore) {
-    zskiplistNode *update[ZSKIPLIST_MAXLEVEL], *x;
-    int i;
+ * element, which is more costly. A pointer to the new node is returned. */
+static zskiplistNode *zslUpdateScore(zskiplist *zsl, zskiplistNode *node, double newscore) {
+    /* If the node, after the score update, would be still exactly
+     * at the same position, we can just update the score without
+     * actually removing and re-inserting the element in the skiplist. */
```
```suggestion
     * actually removing and re-inserting the element in the skiplist.
     * (TODO: The check can be extended to check also equality of the
     * score, but then we'll also need to compare the key order). */
```

src/t_zset.c
zuiderkwast: Right, the edge cases. Now, we remove and re-insert the node at the same position in this case, which is OK. It's not introduced in this PR anyway. We can convert the comment to a follow-up issue.

We could use zslCompareNodes here, but then we'd need to set the new score before we compare and then revert it if the check fails. Or use a stack-allocated temporary node for comparing, just to be able to use zslCompareNodes.

```suggestion
    double oldscore = node->score;
    node->score = newscore;
    if ((node->backward == NULL || zslCompareNodes(node->backward, node) <= 0) &&
        zslCompareNodes(node->level[0].forward, node) >= 0) {
        return NULL;
    } else {
        /* Restore score to restore skiplist order. */
        node->score = oldscore;
    }
```

PR #1429, opened 2024-12-11
Fix Valkey binary build workflow, version support changes.

### LINKED ISSUE #1269: [BUG] Focal OS Valkey binaries in download site seems like wrongly compiled .
**Describe the bug**

Valkey binaries in download site seems like wrongly compiled . When we tried to run binaries from https://download.valkey.io/releases/valkey-8.0.1-focal-x86_64.tar.gz  . They are erroring with 
valkey-server: error while loading shared libraries: libssl.so.3: cannot open shared object file: No such file or directory . Where focul does  not have have libssl3 .

**To reproduce**

Download https://download.valkey.io/releases/valkey-8.0.1-focal-x86_64.tar.gz in Ubuntu 20.04 os and run 
`valkey-8.0.1-focal-x86_64/bin/valkey-server  /etc/valkey/valkey.conf`
**Expected behavior**

Start valkey server normally. 

**Additional information**

I tried multiple versions and same behavior so seems like where ever it is getting compiled it is not correct.

### LINKED ISSUE #1343: Deprecate support for ubuntu 18 and Add support for Ubuntu 22
We should deprecate support for Ubuntu18 binaries and after the recent addition of support of Ubuntu24, I think we should also add support for Ubuntu 22.

This would be in effort to maintain release binaries for all the Ubuntu Releases with Standard Support which can be seen here https://ubuntu.com/about/release-cycle.

roshkhatri:
Fixes https://github.com/valkey-io/valkey/issues/1269 and https://github.com/valkey-io/valkey/issues/1343
This change makes the binary build on the target ubuntu version.

This PR also deprecated ubuntu18 and valkey will not support:

- X86:
  - Ubuntu 20
  - Ubuntu 22
  - Ubuntu 24
 - ARM:
   - Ubuntu 20
   - Ubuntu 22
   
Removed ARM ubuntu 24 as the action we are using for ARM builds does not support Ubuntu 24.

### DIFF
--- a/.github/actions/generate-package-build-matrix/action.yml
+++ b/.github/actions/generate-package-build-matrix/action.yml
@@ -24,7 +24,7 @@ runs:
 
     - name: Get targets
       run: |
-        x86_arch=$(jq -c '[.linux_targets[] | select(.arch=="x86_64") | .target |= sub("ubuntu([0-9]+\\.[0-9]+)"; "ubuntu-\1")]' utils/releasetools/build-config.json)
+        x86_arch=$(jq -c '[.linux_targets[] | select(.arch=="x86_64")]' utils/releasetools/build-config.json)
         x86_matrix=$(echo "{ \"distro\" : $x86_arch }" | jq -c .)
         echo "X86_MATRIX=$x86_matrix" >> $GITHUB_ENV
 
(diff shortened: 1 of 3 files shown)


enjoy-binbin (approved):

madolson:
For future references, future binary releases will have libc issues fixed. (This was mostly reported with TLS)

madolson (approved):

PR #1430, opened 2024-12-12
defrag: eliminate persistent kvstore pointer and edge case fixes

JimB123:
This update addresses several issues in defrag:
1. In the defrag redesign (https://github.com/valkey-io/valkey/pull/1242), a bug was introduced where `server.cronloops` was no longer being incremented in the `whileBlockedCron()`.  This resulted in some memory statistics not being updated while blocked.
2. In the test case for AOF loading, we were seeing errors due to defrag latencies.  However, running the math, the latencies are justified given the extremely high CPU target of the testcase.  Adjusted the expected latency check to allow longer latencies for this case where defrag is undergoing starvation while AOF loading is in progress.
3. A "stage" is passed a "target".  For the main dictionary and expires, we were passing in a `kvstore*`.  However, on flushall or swapdb, the pointer may change.  It's safer and more stable to use an index for the DB (a DBID).  Then if the pointer changes, we can detect the change, and simply abort the stage.  (If there's still fragmentation to deal with, we'll pick it up again on the next cycle.)
4. We always start a new stage on a new defrag cycle.  This gives the new stage time to run, and prevents latency issues for certain stages which don't operate incrementally.  However, often several stages will require almost no work, and this will leave a chunk of our CPU allotment unused.  This is mainly an issue in starvation situations (like AOF loading or LUA script) - where defrag is running infrequently, with a large duty-cycle.  This change allows a new stage to be initiated if we still have a standard duty-cycle remaining.  (This can happen during starvation situations where the planned duty cycle is larger than the standard cycle.  Most likely this isn't a concern for real scenarios, but it was observed in testing.)
5. Minor comment correction in `server.h`

### DIFF
--- a/src/defrag.c
+++ b/src/defrag.c
@@ -120,7 +120,7 @@ typedef doneStatus (*kvstoreHelperPreContinueFn)(monotime endtime, void *privdat
 // Private data for main dictionary keys
 typedef struct {
     kvstoreIterState kvstate;
-    serverDb *db;
+    int dbid;
 } defragKeysCtx;
 static_assert(offsetof(defragKeysCtx, kvstate) == 0, "defragStageKvstoreHelper requires this");
 
@@ -735,7 +735,7 @@ static void defragModule(serverDb *db, robj *obj) {
 /* for each key we scan in the main dict, this function will attempt to defrag
  * all the various pointers it has. */
 static void defragKey(defragKeysCtx *ctx, robj **elemref) {
-    serverDb *db = ctx->db;
+    serverDb *db = &server.db[ctx->dbid];
     int slot = ctx->kvstate.slot;
     robj *newob, *ob;
     unsigned char *newzl;
@@ -919,7 +919,7 @@ static doneStatus defragLaterStep(monotime endtime, void *privdata) {
         robj *ob = found;
 
         long long key_defragged = server.stat_active_defrag_hits;
-        bool timeout = (defragLaterItem(ob, &defrag_later_cursor, endtime, ctx->db->id) == 1);
+        bool timeout = (defragLaterItem(ob, &defrag_later_cursor, endtime, ctx->dbid) == 1);
         if (key_defragged != server.stat_active_defrag_hits) {
             server.stat_active_defrag_key_hits++;
         } else {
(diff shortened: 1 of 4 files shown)


src/defrag.c
```diff
@@ -1283,9 +1299,8 @@ static void beginDefragCycle(void) {
     defrag.remaining_stages = listCreate();
 
     for (int dbid = 0; dbid < server.dbnum; dbid++) {
-        serverDb *db = &server.db[dbid];
-        addDefragStage(defragStageDbKeys, db, NULL);
-        addDefragStage(defragStageExpiresKvstore, db->expires, NULL);
+        addDefragStage(defragStageDbKeys, (void*)(uintptr_t)dbid, NULL);
+        addDefragStage(defragStageExpiresKvstore, (void*)(uintptr_t)dbid, NULL);
```
JimB123:
```suggestion
        addDefragStage(defragStageDbKeys, (void *)(uintptr_t)dbid, NULL);
        addDefragStage(defragStageExpiresKvstore, (void *)(uintptr_t)dbid, NULL);
```

src/defrag.c
```diff
@@ -1222,29 +1229,38 @@ static long long activeDefragTimeProc(struct aeEventLoop *eventLoop, long long i
     }
 
     monotime starttime = getMonotonicUs();
-    monotime endtime = starttime + computeDefragCycleUs();
+    int dutyCycleUs = computeDefragCycleUs();
+    monotime endtime = starttime + dutyCycleUs;
+    bool haveMoreWork = true;
```
hpatro, line 1234: I think we were avoiding camel cased variables in methods. But the clang format check seems to be not complaining.

hpatro (commented):
The change looks good to me. But if I understand correctly there isn't enough test coverage to discover such issues. Should we add more test with defrag to avoid things like this in future?

hpatro:
Talked to @JimB123, he mentioned with this change https://github.com/valkey-io/valkey/pull/1303, it should help us discover some of the above issues. Currently the testing with defrag is miniscule.

hpatro (approved):

PR #1431, opened 2024-12-12
Check length before reading in `stringmatchlen`

thaliaarchi:
Fixes four cases where `stringmatchlen` could overrun the pattern if it is not terminated with NUL.

These commits are cherry-picked from my [fork](https://github.com/thaliaarchi/antirez-stringmatch) which extracts `stringmatch` as a library and compares it to other projects by antirez which use the same matcher.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -104,22 +104,22 @@ static int stringmatchlen_impl(const char *pattern,
 
             pattern++;
             patternLen--;
-            not_op = pattern[0] == '^';
+            not_op = patternLen && pattern[0] == '^';
             if (not_op) {
                 pattern++;
                 patternLen--;
             }
             match = 0;
             while (1) {
-                if (pattern[0] == '\\' && patternLen >= 2) {
+                if (patternLen >= 2 && pattern[0] == '\\') {
                     pattern++;
                     patternLen--;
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
Thanks!

This is not a security issue, right? We always call this with null-terminated strings, I believe.

It's still good to remove the assumption about null-terminated strings and never look beyond the last char of the strings.

thaliaarchi:
I count 20 calls in Valkey to `stringmatch`/`stringmatchlen`. I presume calls to `stringmatch` are safe, since callers know it expects a NUL-terminated string. Every call to `stringmatchlen` passes a pattern from an sds type using length `sdslen(pattern)`. The sds API seems to guarantee that its strings are NUL-terminated. Thus, unless there's a bug in sds or callers break the `stringmatch` expectation, this bug is a non-issue. It's just defensive, not a security issue. However, I haven't traced the pattern initializations to be sure, because many are stored in more complex data structures.

enjoy-binbin (approved):

zuiderkwast:
Thanks! I edited the top comment because we use it as the commit message when squash-merging.

thaliaarchi:
Thanks for the edit. I would have put the extra context in another comment if I knew it would be in the message. Oh well.

PR #1432, opened 2024-12-12
valkey-cli auto-exit from subscribed mode

Nikhil-Manglore:
Resolves issue with valkey-cli not auto exiting from subscribed mode on reaching zero pub/sub subscription (previously filed on Redis) https://github.com/redis/redis/issues/12592


src/valkey-cli.c
```diff
@@ -218,6 +218,10 @@ static struct config {
     int shutdown;
     int monitor_mode;
     int pubsub_mode;
+    int pubsub_channel_count;
+    int pubsub_pattern_count;
+    int pubsub_shard_count;
```
zuiderkwast: There is something strange with how the subscriptions are counted. It seems like channels and patterns are counted together with one counter while only shard-channels are counted with a separate counter.

I tried this just now:

```
127.0.0.1:6379> subscribe ch
1) "subscribe"
2) "ch"
3) (integer) 1
127.0.0.1:6379(subscribed mode)> psubscribe p*
1) "psubscribe"
2) "p*"
3) (integer) 2
127.0.0.1:6379(subscribed mode)> ssubscribe s
1) "ssubscribe"
2) "s"
3) (integer) 1
127.0.0.1:6379(subscribed mode)> unsubscribe
1) "unsubscribe"
2) "ch"
3) (integer) 1
127.0.0.1:6379(subscribed mode)> punsubscribe
1) "punsubscribe"
2) "p*"
3) (integer) 0
```

If you keep three counters internally, I think you can fail to detect when unsubscribe is finished. If you use all three in the same test case, I think you can see this behaviour.

src/valkey-cli.c
```diff
@@ -2224,6 +2228,31 @@ static int cliReadReply(int output_raw_strings) {
...
+                if (strcmp(cmd, "subscribe") == 0 || strcmp(cmd, "unsubscribe") == 0) {
+                    config.pubsub_channel_count = count;
+                } else if (strcmp(cmd, "psubscribe") == 0 || strcmp(cmd, "punsubscribe") == 0) {
+                    config.pubsub_pattern_count = count;
+                } else if (strcmp(cmd, "ssubscribe") == 0 || strcmp(cmd, "sunsubscribe") == 0) {
+                    config.pubsub_shard_count = count;
+                }
+                config.pubsub_total_count = config.pubsub_channel_count + config.pubsub_pattern_count + config.pubsub_shard_count;
+
+                if (config.pubsub_total_count == 0) {
```
zuiderkwast: Why do we store the total count in config? Can't it just be a local variable?

zuiderkwast (commented):
Hello! Welcome to Valkey!

I remember this issue. :)

src/valkey-cli.c
Nikhil-Manglore: Yeah I had that initially but changed it when I was debugging, I'll change it back to a local variable.

src/valkey-cli.c
Nikhil-Manglore: I found this issue as well when I was manually testing the code, currently looking in this issue.

zuiderkwast: Isn't the solution to use two counters instead of three?

    int pubsub_unsharded_count; /* channels and patterns */
    int pubsub_sharded_count;   /* shard channels */

Nikhil-Manglore: Yes I think you're correct, i've pushed those changes. I have another question. Are we allowed to use punsubscribe to unsubscribe from channels we have used the subscribe command on?

for example:
```subscribe ch1 ch2 ch3```
```punsubscribe ch*```

Comment (zuiderkwast) [reply src/valkey-cli.c]: I believe patterns and channels are stored separately on the server. If you `subscribe ch1` and `psubscribe ch*`, then you will get messages on `ch` until you do both `unsubscribe` and `punsubscribe`. That's what I think, but I haven't tested this.

Comment (Nikhil-Manglore) [reply src/valkey-cli.c]: When I tested it that's what I got as well. We had to do unsubscribe and punsubscribe to fully unsubscribe from the channels and get out of subscribe mode. I assume this is the desired functionality.

File: src/valkey-cli.c [hunk 3]
```diff
@@ -2224,6 +2226,29 @@ static int cliReadReply(int output_raw_strings) {
...
+                    config.pubsub_sharded_count = count;
+                }
+                
+                if (config.pubsub_unsharded_count == 0 && config.pubsub_sharded_count == 0) {
+                    config.pubsub_mode = 0;
+                    cliRefreshPrompt();
+                }
+            }
+        }
+    }
```
zuiderkwast: The logic looks right to me. I think we should move it though.

You have put this logic in `cliReadReply`, but I think we should put it where `cliReadReply` is called, or just a few lines below that. That's where most of the client state is handled per command and it's also where the logic is that sets `config.pubsub_mode = 1;`, so it seems more logical to put this in the same place. The reply can be accessed as `config.last_reply`.

Somewhere here:

```C

        /* Read response, possibly skipping pubsub/push messages. */
        while (1) {
            if (cliReadReply(output_raw) != REDIS_OK) {
                zfree(argvlen);
                return REDIS_ERR;
            }
            fflush(stdout);
            if (config.pubsub_mode || num_expected_pubsub_push > 0) {
                if (isPubsubPush(config.last_reply)) {
                    if (num_expected_pubsub_push > 0 && !strcasecmp(config.last_reply->element[0]->str, command)) {
                        /* This pushed message confirms the
                         * [p|s][un]subscribe command. */
                        if (is_subscribe && !config.pubsub_mode) {
                            config.pubsub_mode = 1;
                            cliRefreshPrompt();
                        }
```

tests/unit/pubsub.tcl
```diff
@@ -514,4 +514,79 @@ start_server {tags {"pubsub network"}} {
         assert_equal [r read] {message foo vaz}
     } {} {resp3}
 
+    test "SUBSCRIBE and UNSUBSCRIBE with multiple channels" {
+        # Note: this is testing whether the client exits pubsub mode when subscribed to 0 channels.
+        set rd1 [valkey_deferring_client]
+        
+        assert_equal {1 2 3} [subscribe $rd1 {chan1 chan2 chan3}]
```
zuiderkwast: There is a problem with this test: It doesn't test `valkey-cli`. The client used here is the TCL client which is used internally in the test framework.

The tests for `valkey-cli` are in the file `tests/integration/valkey-cli.tcl`.

zuiderkwast (commented):
Logic looks correct, but I have two comments.

I think it would be good to add a test for valkey-cli that uses SUBSCRIBE, PSUBSCRIBE, SSUBSCRIBE in the same session, an exits pubsub mode when all three are unsubscribed.

tests/unit/pubsub.tcl
Nikhil-Manglore: I deleted the three test cases in the pubsub.tcl file since they didn't test the client and then added one test case to the valkey-cli.tcl file that tests all three commands at once.

src/valkey-cli.c
```diff
@@ -2427,6 +2405,25 @@ static int cliSendCommand(int argc, char **argv, long repeat) {
                             config.pubsub_mode = 1;
                             cliRefreshPrompt();
                         }
+
+                        /* Handle pubsub mode */
+                        if (config.last_reply->elements >= 3) {
```
zuiderkwast: This code now is now inside an `if (num_expected_pubsub_push > 0 ...) { ... }` block.

For a command like UNSUBSCRIBE without channels specified, num_expected_pubsub_push is 1 here, so this code doesn't run if UNSUBSCRIBE returns multiple messages. UNSUBSCRIBE without channels actually unsubscribe all, so it can return multiple replies and the last count can be `0`.

So we need to move this code outside of `if (num_expected_pubsub_push > 0 ...) { ... }`. Put it before that `if`?

tests/integration/valkey-cli.tcl
```diff
@@ -608,6 +608,47 @@ if {!$::tls} { ;# fake_redis_node doesn't support TLS
...
+        write_cli $fd "PUNSUBSCRIBE pattern*"
+        assert_match "*punsubscribe*pattern**" [read_cli $fd]
+
+         # Verify still in pubsub mode
+        catch {run_command $fd "GET key"} err
+        assert_match "*ERR*only*SUBSCRIBE*UNSUBSCRIBE*allowed*" $err
+
+        # Unsubscribe from shard channel
+        write_cli $fd "SUNSUBSCRIBE schannel1"
+        assert_match "*sunsubscribe*schannel1*" [read_cli $fd]
```
zuiderkwast: For the last unsubscribe (or sunsubsribe), we should test the special scenario that we are still subscribed to two or more channels and the command is called without channels, meaning unsubscribe all.

tests/integration/valkey-cli.tcl
```diff
@@ -608,6 +608,47 @@ if {!$::tls} { ;# fake_redis_node doesn't support TLS
...
+
+        # Verify that we've exited pubsub mode
+        assert_equal "PONG" [run_command $fd "PING"]
```
zuiderkwast: This only verifies that the server believes that the client is not in pubsub mode. It would work even without these cli fixes.

What we need to test is that the client know that we have exited pubsub mode. Maybe we can check that the prompt is back to normal `127.0.0.1:6379>` instead of `127.0.0.1:6379(subscribed mode)>`? I don't know if this test framework can do this. I hope so.

zuiderkwast (commented):
Good, thanks. I have some new comments. :)

tests/integration/valkey-cli.tcl
Nikhil-Manglore: Thanks for the suggestions, I pushed the updated code. Let me know if you want this check handled another way.

src/valkey-cli.c
```diff
@@ -2398,6 +2398,34 @@ static int cliSendCommand(int argc, char **argv, long repeat) {
...
+                        } else if (strcmp(cmd, "sunsubscribe") == 0) {
+                            config.pubsub_sharded_count = count;
+                        }
+
+                        if (config.pubsub_unsharded_count == 0 && config.pubsub_sharded_count == 0) {
+                            config.pubsub_mode = 0;
+                            cliRefreshPrompt();
+                        } else {
+                            config.pubsub_mode = 1;
+                        }
```
zuiderkwast: There is already a check that sets `pubsub_mode = 1` below. It becomes redundant with this change. We should only do this in one place.

I think it's better to do it here, but we also need to call cliRefreshPrompt and we should do it only if we're not already in pubsub mode. I suggest like this:

```suggestion
                        if (config.pubsub_unsharded_count == 0 && config.pubsub_sharded_count == 0) {
                            if (config.pubsub_mode) {
                                config.pubsub_mode = 0;
                                cliRefreshPrompt();
                            }
                        } else if (!config.pubsub_mode) {
                            config.pubsub_mode = 1;
                            cliRefreshPrompt();
                        }
```

And this code below can be deleted:

                        if (is_subscribe && !config.pubsub_mode) {
                            config.pubsub_mode = 1;
                            cliRefreshPrompt();
                        }

tests/integration/valkey-cli.tcl
```diff
@@ -611,41 +611,65 @@ if {!$::tls} { ;# fake_redis_node doesn't support TLS
...
+                assert_match "*unsubscribe*channel3*" $line
+            }
+        }
+
+        # Verify that we've exited pubsub mode
+        set response [run_command $fd "PING"]
+        
+        if {[string first "(subscribed mode)" $response] >= 0} {
+            return -code error "Client is still in pubsub mode"
+        }
```
zuiderkwast: Interesting! Does this work though? Does the response really contain the prompt?

We should add the opposite check when we enter subscribed mode, to see that the prompt does include "(subscribed mode)" when we are in subscribed mode. In this way, we also know that this kind of check works.

zuiderkwast (commented):
It's getting better! I'm not sure if check for "(subscribed mode)" in the prompt works as intended though.

Maybe we need to test this in some other way. If you can't solve it, then I'll try to think of something.

tests/integration/valkey-cli.tcl
zuiderkwast: I've done some source code reading. The prompt is only printed when STDIN is a TTY. When we run the cli from the tests, the prompt is not printed.

(The code for this is in the command line editor linenoise (`deps/linenoise/linenoise.c`), function `linenoise()`, which calls linenoiseNoTTY which doesn't print the prompt.)

So, to test if valkey-cli thinks that it's in pubsub mode or not, we need to do something else. Either we can modify linenoise to add some way to make it behave like an interactive terminal even it it's not running in a terminal. That seems complicated though. I have another idea.

There is some prepared work in `valkey-cli` for "internal commands", starting with ":".

```
not connected> :banana
unknown valkey-cli internal command ':banana'
```

The only such commands right now are `:set nohints` and `:set hints` to disable or enable syntax hints.

We could add an internal command to be able to check if we're in pubsub mode. I think for example `:get prompt` (to print the `config.prompt`) would be simple enough and useful for testing. WDYT?

src/valkey-cli.c
Nikhil-Manglore: I was manually testing the code and it seems that when we just use a ```unsubscribe``` command to unsubscribe to all channels the count isn't updated properly. So although we are able to unsubscribe to all channels we won't be able to automatically exit subscribe mode. I will continue to look into it.

zuiderkwast: When unsubscribing all channels, the server sends multiple `unsubscribe` messages and the last one should have count `0`. So we should be able to detect this.

I think now that this code only detects the first unsubscribe message. The rest are received in `cliWaitForMessagesOrStdin`.

This code:

```C
        } else if (FD_ISSET(context->fd, &readfds)) {
            /* Message from the server */
            if (cliReadReply(0) != REDIS_OK) {
                cliPrintContextError();
                exit(1);
            }
            fflush(stdout);
```

It also calls `cliReadReply`. Maybe your first solution where the code was inside `cliReadReply` could actually detect the unsubscribe all channels case. We can put it back there if you want. But can we also do the check to set subscribed node to 1 here? It's good to have the code for pubsub_mode = 1 and pubsub_mode = 0 in the same place.

Nikhil-Manglore: Sorry for the delay. I reverted to the previous commit where the code was in `cliReadReply` and I still seem to get the same issue even though I'm sure it was working last week when I was manually testing. I put a print statement to debug and here is the output

```
127.0.0.1:6379> subscribe ch1 ch2 ch3
1) "subscribe"
2) "ch1"
3) (integer) 1
1) "subscribe"
2) "ch2"
3) (integer) 2
1) "subscribe"
2) "ch3"
3) (integer) 3
127.0.0.1:6379(subscribed mode)> unsubscribe
1) "unsubscribe"
2) "ch3"
3) (integer) 2
Command: unsubscribe, Count: 2
1) "unsubscribe"
2) "ch2"
3) (integer) 1
1) "unsubscribe"
2) "ch1"
3) (integer) 0
```

So it seems to only get the reply of the first unsubscribe instead of all of them

I will continue to look into this issue but am currently unsure why this is occurring

hpatro: @Nikhil-Manglore How is the deep dive going here?

Nikhil-Manglore: Currently still going through the source code to see why this is occurring. Seems that it's not reading all the unsubscribe responses at once. After the above output if we do another "unsubscribe" we will be pulled out of subscribe mode. So the flags to check the number of subscribers seem to be working correctly i'm just unsure why all the messages sent from server to client aren't being read at once.

Nikhil-Manglore: I just fixed the issue by adding code to check for multiple unsubscribe messages in the method `cliWaitForMessagesOrStdin()`. I will push that code now and begin fixing the test cases.

tests/integration/valkey-cli.tcl
Nikhil-Manglore: I have pushed the updated code. I was able to implement all functionality while being able to refactor the code as we had initially discussed. I also was able to implement your suggestion into the test cases and they work.

Also I don't think we can delete this code because otherwise the client never enters subscribe mode

```
if (is_subscribe && !config.pubsub_mode) {
        config.pubsub_mode = 1;
        cliRefreshPrompt();
}
                    ```

File: src/valkey-cli.c [hunk 10]
```diff
@@ -2397,6 +2425,37 @@ static int cliSendCommand(int argc, char **argv, long repeat) {
...
+                                    
+                                if (config.pubsub_unsharded_count == 0 && config.pubsub_sharded_count == 0) {
+                                    if (config.pubsub_mode) {
+                                        config.pubsub_mode = 0;
+                                        cliRefreshPrompt();
+                                    }
+                                } else {
+                                    config.pubsub_mode = 1;
+                                    cliRefreshPrompt();
+                                }
```
hpatro: If we are already in pubsub_mode, do we need to perform this ?

src/valkey-cli.c
```diff
@@ -2246,7 +2249,32 @@ static void cliWaitForMessagesOrStdin(void) {
...
+                    
+                    if (strcmp(cmd, "unsubscribe") == 0 || 
+                        strcmp(cmd, "punsubscribe") == 0 ||
+                        strcmp(cmd, "sunsubscribe") == 0) {
+                        
+                        if (strcmp(cmd, "unsubscribe") == 0) {
+                            config.pubsub_unsharded_count = count;
+                        } else if (strcmp(cmd, "sunsubscribe") == 0) {
+                            config.pubsub_sharded_count = count;
+                        }
```
hpatro: How are we dealing with `punsubscribe` here?

tests/integration/valkey-cli.tcl
```diff
@@ -608,6 +608,67 @@ if {!$::tls} { ;# fake_redis_node doesn't support TLS
         assert_equal "a\n1\nb\n2\nc\n3" [exec {*}$cmdline ZRANGE new_zset 0 -1 WITHSCORES]
     }
 
+    test "valkey-cli pubsub mode with multiple subscription types" {
```
hpatro, line 776: Could we split them up into smaller test cases?

src/valkey-cli.c
```diff
@@ -2246,7 +2249,32 @@ static void cliWaitForMessagesOrStdin(void) {
                 sds out = cliFormatReply(reply, config.output, 0);
                 fwrite(out, sdslen(out), 1, stdout);
                 fflush(stdout);
+
+                // Handle unsubscribe responses
+                if (isPubsubPush(reply) && reply->elements >= 3) {
```
hpatro: This code block seems like duplication of the code below. Could you check if we can place it in one place and handle all scenarios.

tests/integration/valkey-cli.tcl
Nikhil-Manglore: I added three new test cases to cover all the possibilities that we had talked about on Friday. I kept this test case though because it combines all three subscription types. Let me know if I should organize the tests in a different way.

src/valkey-cli.c
Nikhil-Manglore: It doesn't seem like we can put the code block in just one place. From my understanding `cliWaitForMessagesOrStdin()` handles interactive commands while `clientSendCommands` handles non-interactive commands. 

So initially i just had all the code in `clientSendCommands`. When we attempted to use an `unsubscribe` command to unsubscribe from all remaining channels this method would only receive the first response. This is because we were in interactive mode so the system would process the first unsubscribe response (ch1). By adding the same code to the `cliWaitForMessagesOrStdin()` method we are then able to handle commands that are in non-interactive mode. So we'll get responses back from the last two channels we unsubscribed to (ch2, ch3). I tested putting all the code in just the  `cliWaitForMessagesOrStdin()` and we were only able to get responses that we were unsubscribing from the last two channels and not the first one. So I believe we do need the code in both of these methods to handle both interactive and non-interactive mode responses.

What I did instead was create a helper method with the unsubscribe code so that it isn't redundant.

tests/integration/valkey-cli.tcl
zuiderkwast: Hm... I wanted to test the combination of all kinds, to make sure we exit the pubsub mode at the right time, when both counters reach zero.

I guess what @hpatro meant is to just split the test case into smaller parts without really changing anything else. It would make the test cases depend on each other, but that's probably fine. Many test are designed in that way already, so you can't really run them one by one.

src/valkey-cli.c
```diff
@@ -2397,6 +2425,25 @@ static int cliSendCommand(int argc, char **argv, long repeat) {
...
+                        /* If it's an unsubscribe command, call the helper */
+                        if (strcmp(cmd, "unsubscribe") == 0 || strcmp(cmd, "punsubscribe") == 0 || strcmp(cmd, "sunsubscribe") == 0) {
+                            if (config.pubsub_mode && isPubsubPush(config.last_reply)) {
+                                processUnsubscribeReply(config.last_reply);
+                            }
+                        } 
+
+                        /* Handle subscribe commands */
+                        else if (strcmp(cmd, "subscribe") == 0 || strcmp(cmd, "psubscribe") == 0) {
+                            config.pubsub_unsharded_count++;
```
zuiderkwast: Just noticed: In the unsubscribe case, you just set `config.pubsub_unsharded_count = reply->element[2]->integer`. We should do the same here.

The corner case where this matters is when the user subscribes to the same channel twice:

```
127.0.0.1:6379> subscribe x
1) "subscribe"
2) "x"
3) (integer) 1
127.0.0.1:6379(subscribed mode)>  subscribe x
1) "subscribe"
2) "x"
3) (integer) 1
```

tests/integration/valkey-cli.tcl
zuiderkwast: One more corner case to add to the tests: Subscribe to the same channel more than once.

Btw, I think the implementation looks pretty good now. We're close. :)

src/valkey-cli.c
```diff
@@ -2226,9 +2228,29 @@ static int cliReadReply(int output_raw_strings) {
...
+        if (strcmp(cmd, "unsubscribe") == 0 || strcmp(cmd, "punsubscribe") == 0) {
+            config.pubsub_unsharded_count = count;
+        } else if (strcmp(cmd, "sunsubscribe") == 0) {
+            config.pubsub_sharded_count = count;
+        }
+
+        if (config.pubsub_unsharded_count == 0 && config.pubsub_sharded_count == 0) {
+            config.pubsub_mode = 0;
+            cliRefreshPrompt();
+        }
```
zuiderkwast: Why not let this helper handle subscribe and unsubscribe messages together here? They have almost the same logic. (We can name it handlePubsubMode or something.)

It seems to me that we can call this after just checking `isPubsubPush` and then we can remove any other code for setting the pubsub mode, right?

```suggestion
        if (strcmp(cmd, "subscribe") == 0 || strcmp(cmd, "psubscribe") == 0 ||
            strcmp(cmd, "unsubscribe") == 0 || strcmp(cmd, "punsubscribe") == 0) {
            config.pubsub_unsharded_count = count;
        } else if (strcmp(cmd, "ssubscribe") == 0 || strcmp(cmd, "sunsubscribe") == 0) {
            config.pubsub_sharded_count = count;
        }

        if (config.pubsub_unsharded_count == 0 && config.pubsub_sharded_count == 0 && config.pubsub_mode) {
            config.pubsub_mode = 0;
            cliRefreshPrompt();
        } else if (config.pubsub_unsharded_count + config.pubsub_sharded_count > 0 && !config.pubsub_mode) {
            config.pubsub_mode = 1;
            cliRefreshPrompt();
        }
```

src/valkey-cli.c
```diff
@@ -2428,22 +2434,14 @@ static int cliSendCommand(int argc, char **argv, long repeat) {
...
-                            config.pubsub_unsharded_count = count;
-                        } else if (strcmp(cmd, "ssubscribe") == 0) {
-                            config.pubsub_sharded_count = count;
+                        
+                        /* Call the helper to handle both subscribe and unsubscribe commands */
+                        if (strcmp(cmd, "subscribe") == 0 || strcmp(cmd, "psubscribe") == 0 ||
+                            strcmp(cmd, "ssubscribe") == 0 || strcmp(cmd, "unsubscribe") == 0 ||
+                            strcmp(cmd, "punsubscribe") == 0 || strcmp(cmd, "sunsubscribe") == 0) {
+                            handlePubSubMode(config.last_reply);
                         }
```
zuiderkwast: We can delete more. We don't need to check the command because we check it inside handlePubSubMode. If we call handlePubSubMode with some other reply, it does nothing.

```suggestion
                        handlePubSubMode(config.last_reply);
```

Also, we can delete the `if (config.last_reply->elements >= 3)` check above because that's also checked inside the helper.

Finally, can we delete other code to set `config.pubsub_mode = 1` a few lines below now?

```diff
                     if (num_expected_pubsub_push > 0 && !strcasecmp(config.last_reply->element[0]->str, command)) {
                         /* This pushed message confirms the
                          * [p|s][un]subscribe command. */
-                        if (is_subscribe && !config.pubsub_mode) {
-                            config.pubsub_mode = 1;
-                            cliRefreshPrompt();
-                        }
                         if (--num_expected_pubsub_push > 0) {
                             continue; /* We need more of these. */
                         }
```

Nikhil-Manglore: Oh I realized this earlier and had pushed those changes a couple minutes before you sent these comments. The most recent commits should contain both these changes

src/valkey-cli.c
```diff
@@ -2226,9 +2228,35 @@ static int cliReadReply(int output_raw_strings) {
         fflush(stdout);
         sdsfree(out);
     }
+    
     return REDIS_OK;
 }
 
+/* Helper method to process unsubscribe responses */
```
hpatro:
```suggestion
/* Helper method to handle pubsub subscription/unsubscription. */
```

src/valkey-cli.c
```diff
@@ -2226,9 +2228,35 @@ static int cliReadReply(int output_raw_strings) {
...
+static void handlePubSubMode(redisReply *reply) {
+    if (reply->elements >= 3) {
```
hpatro: Do we need this if condition? the couple of places from where you invoke this method have already verified it's a pubsub related response through (`isPubsubPush`).

src/valkey-cli.c
```diff
@@ -2226,9 +2228,35 @@ static int cliReadReply(int output_raw_strings) {
...
+        /* Update counts based on the command type */
+        if (strcmp(cmd, "subscribe") == 0 || strcmp(cmd, "psubscribe") == 0 || 
+            strcmp(cmd, "unsubscribe") == 0 || strcmp(cmd, "punsubscribe") == 0) {
+            config.pubsub_unsharded_count = count;
+        } else if (strcmp(cmd, "ssubscribe") == 0 || strcmp(cmd, "sunsubscribe") == 0) {
+            config.pubsub_sharded_count = count;
+        }
+
+        /* Update pubsub mode based on the current counts */
+        if (config.pubsub_unsharded_count == 0 && config.pubsub_sharded_count == 0 && config.pubsub_mode) {
```
hpatro: nit: we could make this condition symmetrical to your approach in the below condition 
```suggestion
        if (config.pubsub_unsharded_count + config.pubsub_sharded_count == 0 && config.pubsub_mode) {
```

src/valkey-cli.c
```diff
@@ -2397,13 +2431,12 @@ static int cliSendCommand(int argc, char **argv, long repeat) {
             fflush(stdout);
             if (config.pubsub_mode || num_expected_pubsub_push > 0) {
                 if (isPubsubPush(config.last_reply)) {
+                    /* Handle pubsub mode */
```
hpatro: nit: not really helpful.

hpatro (approved):
LGTM overall. minor nitpicks.

zuiderkwast:
The clang-format CI job is complaining about trailing whitespace. See https://github.com/valkey-io/valkey/actions/runs/12676568798/job/35330205216?pr=1432

To fix it, you can run `clang-format -i src/valkey-cli.c` or just manually read the diff suggested by clang-format CI job.

Nikhil-Manglore:
I just realized one of the external tests seems to fail here in the checks but always passes locally. Do you know why this could be occurring?

zuiderkwast (approved):
Thanks for your patience!

zuiderkwast:
> I just realized one of the external tests seems to fail here in the checks but always passes locally. Do you know why this could be occurring?

I don't know. I think some github runners are slow or there are some other problems. The failures are not related to this PR.

hpatro:
Thanks @Nikhil-Manglore for working on this!

PR #778, opened 2024-07-12
For MEETs, save the extensions support flag immediately during MEET processing

bentotten:
For backwards compatibility reasons, a node will wait until it receives a cluster message with the extensions flag before sending its own extensions. This leads to a delay in shard ID propagation that can corrupt nodes.conf with inaccurate shard IDs if a node is restarted before this can stabilize.

This fixes much of that delay by immediately triggering the extensions-supported flag during the MEET processing and attaching the node to the link, allowing the PONG reply to contain OSS extensions.

Partially fixes #774

PR #1436, opened 2024-12-13
Drop the MEET packet if the link node is in handshake state

enjoy-binbin:
After #1307 got merged, we notice there is a assert happen in setClusterNodeToInboundClusterLink:
```
=== ASSERTION FAILED ===
==> '!link->node' is not true
```

In #778, we will call setClusterNodeToInboundClusterLink to attach the node to the link
during the MEET processing, so if we receive a another MEET packet in a short time, the
node is still in handshake state, we will meet this assert and crash the server.

If the link is bound to a node and the node is in the handshake state, and we receive
a MEET packet, it may be that the sender sent multiple MEET packets so in here we are
dropping the MEET to avoid the assert in setClusterNodeToInboundClusterLink. The assert
will happen if the other sends a MEET packet because it detects that there is no inbound
link, this node creates a new node in HANDSHAKE state (with a random node name), and
respond with a PONG. The other node receives the PONG and removes the CLUSTER_NODE_MEET
flag. This node is supposed to open an outbound connection to the other node in the next
cron cycle, but before this happens, the other node re-sends a MEET on the same link
because it still detects no inbound connection.

Note that in getNodeFromLinkAndMsg, the node in the handshake state has a random name
and not truly "known", so we don't know the sender. Dropping the MEET packet can prevent
us from creating a random node, avoid incorrect link binding, and avoid duplicate MEET
packet eliminate the handshake state.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3004,7 +3004,8 @@ int clusterIsValidPacket(clusterLink *link) {
     }
 
     if (type == server.cluster_drop_packet_filter || server.cluster_drop_packet_filter == -2) {
-        serverLog(LL_WARNING, "Dropping packet that matches debug drop filter");
+        serverLog(LL_WARNING, "Dropping packet of type %s that matches debug drop filter",
+                  clusterGetMessageTypeString(type));
         return 0;
     }
 
@@ -3095,7 +3096,7 @@ int clusterProcessPacket(clusterLink *link) {
         if (server.debug_cluster_close_link_on_packet_drop &&
             (type == server.cluster_drop_packet_filter || server.cluster_drop_packet_filter == -2)) {
             freeClusterLink(link);
-            serverLog(LL_WARNING, "Closing link for matching packet type %hu", type);
+            serverLog(LL_WARNING, "Closing link for matching packet type %s", clusterGetMessageTypeString(type));
             return 0;
         }
         return 1;
@@ -3111,13 +3112,25 @@ int clusterProcessPacket(clusterLink *link) {
             freeClusterLink(link);
             serverLog(
                 LL_NOTICE,
-                "Closing link for node that sent a lightweight message of type %hu as its first message on the link",
-                type);
+                "Closing link for node that sent a lightweight message of type %s as its first message on the link",
+                clusterGetMessageTypeString(type));
             return 0;
         }
         clusterNode *sender = link->node;
         sender->data_received = now;
         clusterProcessLightPacket(link, type);
+        return 1;
+    }
+
+    if (type == CLUSTERMSG_TYPE_MEET && link->node && nodeInHandshake(link->node)) {
+        /* If the link is bound to a node and the node is in the handshake state, and we receive
+         * a MEET packet, it may be that the sender sent multiple MEET packets when reconnecting,
+         * and in here we are dropping the MEET. Note that in getNodeFromLinkAndMsg, the node in
+         * the handshake state has a random name and not truly "known", so we don't know the sender.
+         * Dropping the MEET packet can prevent us from creating a random node, avoid incorrect
+         * link binding, and avoid duplicate MEET packet eliminate the handshake state. */
+        serverLog(LL_NOTICE, "Dropping MEET packet from node %.40s because the node is already in handshake state",
+                  link->node->name);
         return 1;
     }
 
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3124,11 +3124,20 @@ int clusterProcessPacket(clusterLink *link) {
 
     if (type == CLUSTERMSG_TYPE_MEET && link->node && nodeInHandshake(link->node)) {
         /* If the link is bound to a node and the node is in the handshake state, and we receive
-         * a MEET packet, it may be that the sender sent multiple MEET packets when reconnecting,
-         * and in here we are dropping the MEET. Note that in getNodeFromLinkAndMsg, the node in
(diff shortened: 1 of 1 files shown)


enjoy-binbin:
The ci fail link: https://github.com/valkey-io/valkey/actions/runs/12306207562/job/34347454716#step:3:4364

Runing in a loop can trigger the failiure
`./runtest --single unit/cluster/cluster-reliable-meet --dont-clean --dont-pre-clean --loops 10`


@pieturin could you take a look and see if i do it right?

hpatro (approved):
This code flow has become too brittle and we are always in catchup mode. I think we should step back and reconsider the `MEET` flow. 

@madolson / @PingXie / @enjoy-binbin

src/cluster_legacy.c
```diff
@@ -3121,6 +3122,18 @@ int clusterProcessPacket(clusterLink *link) {
         return 1;
     }
 
+    if (type == CLUSTERMSG_TYPE_MEET && link->node && nodeInHandshake(link->node)) {
```
madolson, line 3125: Instead of dropping it, should we should skip creating the new node down where we are normally handling new meet messages?

madolson: It feels like if we are in handshake with the other node, and we get a meet message, we should be OK to put the node online? My understanding is that the bug is more around the double setup around the node.

hpatro: What are the risks of dropping it (if it's in a short span of time) ?

src/cluster_legacy.c
```diff
@@ -3121,6 +3122,18 @@ int clusterProcessPacket(clusterLink *link) {
...
+        /* If the link is bound to a node and the node is in the handshake state, and we receive
+         * a MEET packet, it may be that the sender sent multiple MEET packets when reconnecting,
```
pieturin: I don't think this is happening on a reconnect. Since on a reconnect, the link would be new, and therefore `link->node` would be NULL.
From my understanding the crash is happening if the other node sends a MEET packet because it detects that there is no inbound link, this node creates a new node in HANDSHAKE state (with a random node name), and respond with a PONG. The other node receives the PONG and removes the CLUSTER_NODE_MEET flag. This node is supposed to open an outbound connection to the other node in the next cron cycle, but before this happens, the other node re-sends a MEET on the same link because it still detects no inbound connection. And that's when we crash.

pieturin:
Something else we could do, is to avoid sending MEET packets repeatedly. We can introduce a node field like `last_meet_packet_sent` and only send a MEET packet if the last packet was sent more than the handshake timeout ago here:
https://github.com/valkey-io/valkey/blob/5f7fe9ef21f1feae42257ac93ab33d8f8c06e97f/src/cluster_legacy.c#L4978-L4986

src/cluster_legacy.c
pieturin: Does't this condition mean that we would never end up entering this other condition now?
https://github.com/valkey-io/valkey/blob/5f7fe9ef21f1feae42257ac93ab33d8f8c06e97f/src/cluster_legacy.c#L3276-L3292

madolson: > What are the risks of dropping it (if it's in a short span of time) ?

If an end user sends a `cluster meet` from both nodes at the same time, wouldn't this trigger it to drop the meet packet unnecessarily? I think that might be fine, it just seems not correct to drop the packet.

madolson (approved):
I'm okay merging this and following up as well, given how consistently it causes issues in the CI.

enjoy-binbin:
i forget to mention if we double this line, we will get the crash in #778.
```diff
    clusterSendPing(link, nodeInMeetState(node) ? CLUSTERMSG_TYPE_MEET : CLUSTERMSG_TYPE_PING);
+    clusterSendPing(link, nodeInMeetState(node) ? CLUSTERMSG_TYPE_MEET : CLUSTERMSG_TYPE_PING);
```

src/cluster_legacy.c
enjoy-binbin: i guess i want to mention is that it is the other node doing this reconnect, and send two MEET in a short time to myself node.

yes, your understanding is right, my wording is a bit wrong.

src/cluster_legacy.c
enjoy-binbin: i think we should somehow to drop the MEET or go through the code and find out if we want to keep the MEET, what else code should we do. Droping it seem a safer way i guess, but if we can keep the MEET and make it work, i guess it is a better way.

src/cluster_legacy.c
enjoy-binbin: could you suggest a comment that i can use to put it into the comment? i am a bit suck with the long wording.

enjoy-binbin:
i am merging it to unblock the CI, please feel free to update the comment or code if needed in #1441.

PingXie:
> This code flow has become too brittle and we are always in catchup mode. I think we should step back and reconsider the `MEET` flow.

100% agreed. The entropy in the `MEET` flow has become increasingly unmanageable. Was the `MEET` protocol ever formally spec'd? I suspect not. It might be worthwhile to start by drafting a formal spec of the current behavior as a baseline.

PingXie:
Dropping newer MEET messages feels counterintuitive. While it may help with the assert, it seems to compromise resilience. In this situation, the node associated with the first MEET is likely stale. Could this lead to the MEET ultimately failing, effectively nullifying the retries on the sender side?

src/cluster_legacy.c
pieturin: The new comment looks good. Thank you.

PR #1363, opened 2024-11-26
Use `configure-aws-credentials` workflow instead of passing `secret_access_key`

vudiep411:
## Summary
This PR fixes #1346 where we can get rid of the long term credentials by using OpenID Connect. OpenID Connect (OIDC) allows your GitHub Actions workflows to access resources in Amazon Web Services (AWS), without needing to store the AWS credentials as long-lived GitHub secrets.

## Changes
We can remove these secrets that were passed in previously:
```
token: ${{ secrets.GITHUB_TOKEN }}
      bucket: ${{ secrets.AWS_S3_BUCKET }}
      access_key_id: ${{ secrets.AWS_S3_ACCESS_KEY_ID }}
      secret_access_key: ${{ secrets.AWS_S3_ACCESS_KEY }}
```

Instead we only need the `role-to-assume` arn. For more information [OIDC](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/about-security-hardening-with-openid-connect).

## Prerequisites
Before merging this PR, we need to make sure to set up the proper Identity providers on the production AWS account. Follow this [guides](https://docs.github.com/en/actions/security-for-github-actions/security-hardening-your-deployments/configuring-openid-connect-in-amazon-web-services). 
 **Quick guide**:
1. Create an Identity provider with OpenID Connect.
Provider url: `https://token.actions.githubusercontent.com` <br/>
Audience: `sts.amazonaws.com`
Screenshot 2024-11-26 at 1 02 23 PM

2. Assign the role into this new provider with this trust policy like below
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Effect": "Allow",
            "Principal": {
                "Federated": "arn:aws:iam::<aws_account_id>:oidc-provider/token.actions.githubusercontent.com"
            },
            "Action": "sts:AssumeRoleWithWebIdentity",
            "Condition": {
                "StringLike": {
                    "token.actions.githubusercontent.com:sub": [
                        "repo:valkey-io/valkey:ref:refs/heads/unstable",
                        "repo:valkey-io/valkey:ref:refs/tags/*"
                    ],
                    "token.actions.githubusercontent.com:aud": "sts.amazonaws.com"
                }
            }
        }
    ]
}
```
3. In Github secrets, Add these secrets:
Screenshot 2024-11-26 at 2 11 02 PM

## Results
**Github action run:**
Screenshot 2024-11-26 at 2 06 17 PM

**Files in S3 Dev env:**
[...]

PR #1438, opened 2024-12-13
Skip build-release-packages CI job in forks

enjoy-binbin:
The CI job was introduced in #1363, we should skip it in forks.


roshkhatri (approved):
Now that you opened a PR, I realized we do not want everyone's workflow to fail who pull these changes to their fork. The author can manually trigger the workflow too and in valkey we can run it on push.

hwware (approved):

PR #1439, opened 2024-12-13
Change msgBlock struck to not explicitly include message

madolson:
We are getting a number of errors like:
```
array subscript ‘clusterMsg[0]’ is partly outside array bounds of ‘unsigned char[2272]’
```

Which is basically GCC telling us that we have an object which is longer than the underlying storage of the allocation. We actually do this a lot, but GCC is generally not aware of how big the underlying allocation is it doesn't throw this error. We are specifically getting this error because the msgBlock can be of variable length depending on the type of message, but GCC assumes it's the longest one possible. The solution I went with here was make the message type optional, so that it wasn't included in the size. I think this also makes some sense, since it's really just a helper for us to easily cast the object around.

I considered disabling this error, but it is generally pretty useful since it can catch real issues. Another solution would be to over-allocate to the largest possible object. 

Results: https://github.com/valkey-io/valkey/actions/runs/12323895097/job/34400428111

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -424,7 +424,7 @@ typedef struct {
     union {
         clusterMsg msg;
         clusterMsgLight msg_light;
-    };
+    } data[];
 } clusterMsgSendBlock;
 
 /* -----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


hpatro (commented):
Seems to be crashing around cluster code flow in the CI https://github.com/valkey-io/valkey/actions/runs/12324020161/job/34400761034 .

madolson:
Yeah, I think I actually found a simpler method, so going to open another PR.

zuiderkwast:
Curious what the simpler method is.

madolson:
Refactoring the allocation to a seperate function made the errors go away on most of the calls, but not all, but I didn't realize that when I closed it.

PR #1441, opened 2024-12-13
Only (re-)send MEET packet once every handshake timeout period

pieturin:
Add `meet_sent` field in `clusterNode` indicating the last time we sent a MEET packet. Use this field to only (re-)send a MEET packet once every handshake timeout period when detecting a node without an inbound link.

When receiving multiple MEET packets on the same link while the node is in handshake state, instead of dropping the packet, we now simply prevent the creation of a new node. This way we still process the MEET packet's gossip and reply with a PONG as any other packets.

Improve some logging messages to include `human_nodename`. Add `nodeExceedsHandshakeTimeout()` function.

This is a follow-up to this previous PR: https://github.com/valkey-io/valkey/pull/1307
And a partial fix to the crash described in: https://github.com/valkey-io/valkey/pull/1436

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -4044,8 +4044,10 @@ void clusterSendPing(clusterLink *link, int type) {
     clusterMsg *hdr = &msgblock->msg;
 
     if (!link->inbound) {
-        if (type == CLUSTERMSG_TYPE_PING)       link->node->ping_sent = mstime();
-        else if (type == CLUSTERMSG_TYPE_MEET)  link->node->meet_sent = mstime();
+        if (type == CLUSTERMSG_TYPE_PING)
+            link->node->ping_sent = mstime();
+        else if (type == CLUSTERMSG_TYPE_MEET)
+            link->node->meet_sent = mstime();
     }
 
     /* Populate the gossip fields */
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3229,10 +3229,15 @@ int clusterProcessPacket(clusterLink *link) {
                      * to this is the flag that indicates extensions are supported, as
                      * we want to send extensions right away in the return PONG in order
                      * to reduce the amount of time needed to stabilize the shard ID. */
-                    clusterNode *node;
-
-                    node = createClusterNode(NULL, CLUSTER_NODE_HANDSHAKE);
-                    serverAssert(nodeIp2String(node->ip, link, hdr->myip) == C_OK);
+                    clusterNode *node = createClusterNode(NULL, CLUSTER_NODE_HANDSHAKE);
+                    if (nodeIp2String(node->ip, link, hdr->myip) != C_OK) {
+                        /* We cannot get the IP info from the link, it probably means the connection is closed. */
+                        serverLog(LL_NOTICE, "Closing link even though we received a MEET packet on it, "
+                                  "because the connection has an error");
+                        freeClusterLink(link);
+                        freeClusterNode(node);
+                        return 0;
+                    }
                     getClientPortFromClusterMsg(hdr, &node->tls_port, &node->tcp_port);
                     node->cport = ntohs(hdr->cport);
                     if (hdr->mflags[0] & CLUSTERMSG_FLAG0_EXT_DATA) {
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3223,12 +3223,12 @@ int clusterProcessPacket(clusterLink *link) {
             if (!sender) {
                 if (!link->node) {
                     /* Add this node if it is new for us and the msg type is MEET.
-                    * In this stage we don't try to add the node with the right
-                    * flags, replicaof pointer, and so forth, as this details will be
-                    * resolved when we'll receive PONGs from the node. The exception
-                    * to this is the flag that indicates extensions are supported, as
-                    * we want to send extensions right away in the return PONG in order
-                    * to reduce the amount of time needed to stabilize the shard ID. */
+                     * In this stage we don't try to add the node with the right
+                     * flags, replicaof pointer, and so forth, as this details will be
+                     * resolved when we'll receive PONGs from the node. The exception
+                     * to this is the flag that indicates extensions are supported, as
+                     * we want to send extensions right away in the return PONG in order
+                     * to reduce the amount of time needed to stabilize the shard ID. */
                     clusterNode *node;
 
                     node = createClusterNode(NULL, CLUSTER_NODE_HANDSHAKE);
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -121,6 +121,7 @@ void freeClusterLink(clusterLink *link);
 int verifyClusterNodeId(const char *name, int length);
 sds clusterEncodeOpenSlotsAuxField(int rdbflags);
 int clusterDecodeOpenSlotsAuxField(int rdbflags, sds s);
+static int nodeExceedsHandshakeTimeout(clusterNode *node, mstime_t now);
 
 /* Only primaries that own slots have voting rights.
  * Returns 1 if the node has voting rights, otherwise returns 0. */
@@ -1336,9 +1337,10 @@ clusterLink *createClusterLink(clusterNode *node) {
  * with this link will have the 'link' field set to NULL. */
 void freeClusterLink(clusterLink *link) {
     serverAssert(link != NULL);
-    serverLog(LL_DEBUG, "Freeing cluster link for node: %.40s:%s",
+    serverLog(LL_DEBUG, "Freeing cluster link for node: %.40s:%s (%s)",
               link->node ? link->node->name : "<unknown>",
-              link->inbound ? "inbound" : "outbound");
+              link->inbound ? "inbound" : "outbound",
+              link->node ? link->node->human_nodename : "<unknown>");
 
     if (link->conn) {
         connClose(link->conn);
@@ -1492,6 +1494,7 @@ clusterNode *createClusterNode(char *nodename, int flags) {
     node->last_in_ping_gossip = 0;
     node->ping_sent = node->pong_received = 0;
     node->data_received = 0;
+    node->meet_sent = 0;
     node->fail_time = 0;
     node->link = NULL;
     node->inbound_link = NULL;
@@ -1702,7 +1705,7 @@ void clusterAddNode(clusterNode *node) {
  */
 void clusterDelNode(clusterNode *delnode) {
     serverAssert(delnode != NULL);
-    serverLog(LL_DEBUG, "Deleting node %.40s from cluster view", delnode->name);
+    serverLog(LL_DEBUG, "Deleting node %.40s (%s) from cluster view", delnode->name, delnode->human_nodename);
 
     int j;
     dictIterator *di;
@@ -3263,18 +3266,18 @@ int clusterProcessPacket(clusterLink *link) {
                  * the gossip section here since we have to trust the sender because
                  * of the message type. */
                 clusterProcessGossipSection(hdr, link);
-            } else if (sender->link && now - sender->ctime > server.cluster_node_timeout) {
+            } else if (sender->link && nodeExceedsHandshakeTimeout(sender, now)) {
                 /* The MEET packet is from a known node, after the handshake timeout, so the sender thinks that I do not
                  * know it.
-                 * Freeing my outbound link to that node, to force a reconnect and sending a PING.
+                 * Free my outbound link to that node, triggering a reconnect and a PING over the new link.
                  * Once that node receives our PING, it should recognize the new connection as an inbound link from me.
                  * We should only free the outbound link if the node is known for more time than the handshake timeout,
                  * since during this time, the other side might still be trying to complete the handshake. */
 
                 /* We should always receive a MEET packet on an inbound link. */
                 serverAssert(link != sender->link);
-                serverLog(LL_NOTICE, "Freeing outbound link to node %.40s after receiving a MEET packet from this known node",
-                          sender->name);
+                serverLog(LL_NOTICE, "Freeing outbound link to node %.40s (%s) after receiving a MEET packet from this known node",
+                          sender->name, sender->human_nodename);
                 freeClusterLink(sender->link);
             }
         }
@@ -4040,7 +4043,10 @@ void clusterSendPing(clusterLink *link, int type) {
     clusterMsgSendBlock *msgblock = createClusterMsgSendBlock(type, estlen);
     clusterMsg *hdr = &msgblock->msg;
 
-    if (!link->inbound && type == CLUSTERMSG_TYPE_PING) link->node->ping_sent = mstime();
+    if (!link->inbound) {
+        if (type == CLUSTERMSG_TYPE_PING)       link->node->ping_sent = mstime();
+        else if (type == CLUSTERMSG_TYPE_MEET)  link->node->meet_sent = mstime();
+    }
 
     /* Populate the gossip fields */
     int maxiterations = wanted * 3;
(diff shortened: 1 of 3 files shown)


pieturin:
@enjoy-binbin, @PingXie, please take a look at this PR when you have the chance.

enjoy-binbin (approved):
LGTM, I also thought about limiting it, but I thought we should be more robust to double packets.

src/cluster_legacy.c
```diff
@@ -4957,10 +4965,23 @@ void clusterHandleManualFailover(void) {
...
+static mstime_t getHandshakeTimeout(void) {
+    /* The handshake timeout is the time after which a handshake node that was
+     * not turned into a normal node is removed from the nodes. Usually it is
+     * just the cluster_node_timeout value, but when cluster_node_timeout is
+     * too small we use the value of 1 second. */
+    return max(server.cluster_node_timeout, 1000);
+}
+
+static int nodeExceedsHandshakeTimeout(clusterNode *node, mstime_t now) {
+    serverAssert(node != NULL);
```
hpatro: I don't think these kind of assertion help a lot.

hpatro (approved):

pieturin:
In the last revision of this PR, I removed the code to drop the MEET packet from https://github.com/valkey-io/valkey/pull/1436 and replaced it with a check to prevent from creating a new node when receiving several MEET packets on the same link while the node is in handshake.

@enjoy-binbin, @madolson, @PingXie, please tell me if you prefer this over dropping the packet entirely.

src/cluster_legacy.c
```diff
@@ -3238,7 +3220,26 @@ int clusterProcessPacket(clusterLink *link) {
...
+                 * The other sends a MEET packet because it detects that there is no inbound link,
+                 * this node creates a new node in HANDSHAKE state (with a random node name), and
+                 * respond with a PONG. The other node receives the PONG and removes the
+                 * CLUSTER_NODE_MEET flag.
+                 * This node is supposed to open an outbound connection to the other node in the
+                 * next cron cycle, but before this happens, the other node might re-send a MEET on
+                 * the same link because it still detects no inbound connection.
+                 *
+                 * Note that in getNodeFromLinkAndMsg, the node in the handshake state has a random name
+                 * and not truly "known", so we don't know the sender. */
```
PingXie: Maybe we could simplify the comments a bit?

```suggestion
                /* A second MEET packet was received on an existing link during the handshake process.
                 * This happens when the other node detects no inbound link, and re-sends a MEET packet
                 * before this node can respond with a PING. This MEET is a no-op.
                 *
                 * Note: Nodes in HANDSHAKE state are not fully "known" (random names), so the sender 
                 * remains unidentified at this point. The MEET packet might be re-sent if the inbound 
                 * connection is still unestablished by the next cron cycle.
                 */
```

src/cluster_legacy.c
```diff
@@ -3238,7 +3220,26 @@ int clusterProcessPacket(clusterLink *link) {
...
+                debugServerAssert(link->inbound && nodeInHandshake(link->node));
```
PingXie: I wonder if we should process the gossip sections here too?

src/cluster_legacy.c
```diff
@@ -4960,10 +4966,22 @@ void clusterHandleManualFailover(void) {
...
+static mstime_t getHandshakeTimeout(void) {
+    /* The handshake timeout is the time after which a handshake node that was
+     * not turned into a normal node is removed from the nodes. Usually it is
+     * just the cluster_node_timeout value, but when cluster_node_timeout is
+     * too small we use the value of 1 second. */
+    return max(server.cluster_node_timeout, 1000);
+}
+
+static int nodeExceedsHandshakeTimeout(clusterNode *node, mstime_t now) {
+    return now - node->ctime > getHandshakeTimeout() ? 1 : 0;
```
PingXie, line 4976: Should we use `inbound_link_freed_time` instead? also the field name isn't super clear IMO. Can we consider reusing`meet_sent`? We should also move the reset of this time out of `freeClusterLink`.

```suggestion
    return now - node-> meet_sent > getHandshakeTimeout() ? 1 : 0;
```

src/cluster_legacy.c
```diff
@@ -4972,19 +4990,22 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t handshake_
...
+    if (nodeInHandshake(node) && nodeExceedsHandshakeTimeout(node, now)) {
+        serverLog(LL_WARNING, "Clusterbus handshake timeout %s:%d", node->ip, node->cport);
         clusterDelNode(node);
         return 1;
     }
-    if (node->link != NULL && node->inbound_link == NULL && nodeInNormalState(node) &&
-        now - node->inbound_link_freed_time > handshake_timeout) {
+    if (nodeInNormalState(node) && node->link != NULL && node->inbound_link == NULL &&
+        now - node->inbound_link_freed_time > getHandshakeTimeout() &&
+        now - node->meet_sent > getHandshakeTimeout()) {
```
PingXie, line 4998: If we could reuse `meet_sent`, this check would be simplified a lot.

```suggestion
        nodeExceedsHandshakeTimeout) {
```

PingXie (commented):
LGTM overall. 

I'd prefer not dropping the second MEET. Not because the dropping itself is a concern but the fix is a bit ad-hoc. I feel that it is just a valid case and we should just handle it.

Another comment is on the new `inbound_link_freed_time` field. I feel that it is not created at the right level (of abstraction). It should not be a concern of `freeClusterLink`. I actually like the new `meet_sent` more. What do you think?

src/cluster_legacy.c
pieturin: We can't use `meet_sent` here as this function is also used by the receiver of the MEET packet to check if the handshake timed out.

src/cluster_legacy.c
pieturin: I think we still need `inbound_link_freed_time` as we don't want to send a MEET packet just because the inbound connection was temporarily closed because of a fluke in the network.

src/cluster_legacy.c
pieturin: Yes, I think we should.

enjoy-binbin:
> When receiving multiple MEET packets on the same link while the node is in handshake state, instead of dropping the packet, we now simply prevent the creation of a new node. This way we still process the MEET packet's gossip and reply with a PONG as any other packets.

ok, this LGTM and it was actually my first attempt, back then i mistakenly thought that repeated MEET would eliminate the handshake state, but now that I look at the code I think it is safe.

pieturin:
> Another comment is on the new inbound_link_freed_time field. I feel that it is not created at the right level (of abstraction). It should not be a concern of freeClusterLink. I actually like the new meet_sent more. What do you think?

I don't think we can get rid of `inbound_link_freed_time` and only use `meet_sent`. We don't want to send a MEET packet every time the inbound connection is dropped. We should wait for the timeout period to first make sure the other side hasn't tried to re-connect with us.

Why do you think it should not be a concern of `freeClusterLink()`? This function already updates the node's `link` or `inbound_link`, so it seems natural for it to also update `inbound_link_freed_time`.

src/cluster_legacy.c
pieturin: > also the field name isn't super clear IMO.

Would you prefer something like `inbound_link_last_freed_time`?

madolson:
> When receiving multiple MEET packets on the same link while the node is in handshake state, instead of dropping the packet, we now simply prevent the creation of a new node. This way we still process the MEET packet's gossip and reply with a PONG as any other packets.

Btw, I also agree with this. I mentioned on the other thread I think this is the better behavior, but people were seeing a lot of CI failures so I was okay with a temporary solution.

pieturin:
The code coverage reports one missing line:
https://github.com/valkey-io/valkey/blob/0ccc5cc9533b6a56fdb8a8dd4d5a1af4f23d2b0f/src/cluster_legacy.c#L3253

This line should not be reached anymore since we are waiting for the handshake timeout before re-sending a MEET packet.

PingXie:
> > Another comment is on the new inbound_link_freed_time field. I feel that it is not created at the right level (of abstraction). It should not be a concern of freeClusterLink. I actually like the new meet_sent more. What do you think?
> 
> I don't think we can get rid of `inbound_link_freed_time` and only use `meet_sent`. We don't want to send a MEET packet every time the inbound connection is dropped. We should wait for the timeout period to first make sure the other side hasn't tried to re-connect with us.
> 
> Why do you think it should not be a concern of `freeClusterLink()`? This function already updates the node's `link` or `inbound_link`, so it seems natural for it to also update `inbound_link_freed_time`.

`freeClusterLink()` is invoked in various scenarios, but the new state we're introducing is only relevant in the handshake path. This suggests that the scope of the new state is broader than necessary.

On a second thought, I think a better name would be `handshake_start_time`. This name better reflects its purpose.

* On the initiator side, `handshake_start_time` would be set whenever a `MEET` command is sent.
* On the recipient side, it would be set when the first `PING` message is sent on the outgoing link for a node in the HANDSHAKE state.

pieturin:
> On a second thought, I think a better name would be handshake_start_time. This name better reflects its purpose.

I still think having this is not enough. We should still have `inbound_link_freed_time` to know if the link was freed recently in case it's just a temporary reconnect between the two nodes, we don't want to send a MEET packet because the handshake happened days ago, and now the inbound link is NULL for a fraction of a second. Does this make sense?
And I can't use ping_sent/pong_received/data_received since those are also updated for the outbound link.

> freeClusterLink() is invoked in various scenarios, but the new state we're introducing is only relevant in the handshake path. This suggests that the scope of the new state is broader than necessary.

But we need to update `inbound_link_freed_time` every time the inbound link is freed in order to make sure we don't send the MEET packet on a short disconnect.

I guess a way to get rid of `inbound_link_freed_time` would be to replace it with another field that would track if the `inbound_link` was ever opened on a node, something like `was_inbound_link_set`.

With this, we could replace the condition with something like:
```
    if (nodeInNormalState(node) && node->link != NULL && node->inbound_link == NULL &&
        nodeExceedsHandshakeTimeout(node) && now - node->meet_sent > getHandshakeTimeout() &&
        !node->was_inbound_link_set) {
```

What do you think @PingXie?

pieturin:
I had a talk offline with @PingXie, and we think we can change the condition to send a MEET packet to something like:
```
    if (nodeInNormalState(node) && node->link != NULL && node->inbound_link == NULL &&
        nodeExceedsHandshakeTimeout(node) &&
        now - node->handshake_start_time > getHandshakeTimeout() && now - node->handshake_start_time < getHandshakeTimeout() * 2) {
```

And with that we should be able to get rid of `inbound_link_freed_time`.

pieturin:
A crash was detected recently during the daily runs: https://github.com/valkey-io/valkey/actions/runs/12403569355/job/34627291048#step:8:3854

```
17847:M 19 Dec 2024 00:15:58.007 - Accepting cluster node connection from 127.0.0.1:45768
17847:M 19 Dec 2024 00:15:58.021 * Error converting peer IP to string: (null)


17847:M 19 Dec 2024 00:15:58.021 # === ASSERTION FAILED ===
17847:M 19 Dec 2024 00:15:58.021 # ==> cluster_legacy.c:3252 'nodeIp2String(node->ip, link, hdr->myip) == C_OK' is not true

------ STACK TRACE ------

17847 valkey-server *
src/valkey-server 127.0.0.1:27131 [cluster](clusterProcessPacket+0x1304) [0x4e5634]
src/valkey-server 127.0.0.1:27131 [cluster](clusterReadHandler+0x11e) [0x4e59de]
/__w/valkey/valkey/src/valkey-tls.so(+0x2f1e) [0x7f083983ff1e]
src/valkey-server 127.0.0.1:27131 [cluster](aeMain+0x8a) [0x41afea]
src/valkey-server 127.0.0.1:27131 [cluster](main+0x4d7) [0x40f547]
/lib64/libc.so.6(+0x40c8) [0x7f083985a0c8]
/lib64/libc.so.6(__libc_start_main+0x8b) [0x7f083985a18b]
src/valkey-server 127.0.0.1:27131 [cluster](_start+0x25) [0x410ef5]

...

# Cluster info
cluster_state:fail
cluster_slots_assigned:0
cluster_slots_ok:0
cluster_slots_pfail:0
cluster_slots_fail:0
cluster_known_nodes:2
cluster_size:0
cluster_current_epoch:0
cluster_my_epoch:0
cluster_stats_messages_ping_sent:1
cluster_stats_messages_pong_sent:1
cluster_stats_messages_sent:2
cluster_stats_messages_meet_received:2
cluster_stats_messages_received:2
total_cluster_links_buffer_limit_exceeded:0

------ CLUSTER NODES OUTPUT ------
de11af5df0d8b71a5df786c7a1e7ab0a0beab7fc 127.0.0.1:27128@37127,,tls-port=27127,shard-id=2ea5410be1f7dcac4755a6c9bcc60558e3f4f14b handshake - 1734567357566 0 0 connected
26d986e4e07f0d2c768d62dfdaf9f0ae03d18e49 127.0.0.1:27131@27129,,tls-port=27130,shard-id=236191ce869481919bb106a49cb235538b419178 myself,master - 0 0 0 connected
```

My proposal after talking with @madolson is to remove this assert and instead just remove the newly created node and close the link, since if we cannot get the IP from the link it probably means the connection was closed.
https://github.com/valkey-io/valkey/blob/1c97317518e40efb4c271ae3b0656cc6e43f0110/src/cluster_legacy.c#L3274

pieturin:
Doing the solution with `handshake_start_time` doesn't work with the existing tests. I see that a node is sending a MEET packet just because of a temporary disconnect during one of the `cluster-reliable-meet` tests. Because the other nodes frees the link and reconnects because it didn't receive any PONG for more than `server.cluster_node_timeout / 2`. And then the `cluster-reliable-meet` test fails because the node with the inconsistent view never sends the MEET (not sure why yet though). So I don't think it's worth trying to make this work. It seems more finicky than relying on `inbound_link_freed_time`.

src/cluster_legacy.c
```diff
@@ -3239,42 +3221,59 @@ int clusterProcessPacket(clusterLink *link) {
...
+                    clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG);
+                } else {
+                    /* A second MEET packet was received on an existing link during the handshake process.
+                     * This happens when the other node detects no inbound link, and re-sends a MEET packet
+                     * before this node can respond with a PING. This MEET is a no-op.
+                     *
+                     * Note: Nodes in HANDSHAKE state are not fully "known" (random names), so the sender
+                     * remains unidentified at this point. The MEET packet might be re-sent if the inbound
+                     * connection is still unestablished by the next cron cycle.
+                     */
```
PingXie, line 3257:
```suggestion
                     * connection is still unestablished by the next cron cycle. */
```

src/cluster_legacy.c
```diff
@@ -3239,42 +3221,59 @@ int clusterProcessPacket(clusterLink *link) {
...
+            } else if (sender->link && nodeExceedsHandshakeTimeout(sender, now)) {
                 /* The MEET packet is from a known node, after the handshake timeout, so the sender thinks that I do not
                  * know it.
-                 * Freeing my outbound link to that node, to force a reconnect and sending a PING.
+                 * Free my outbound link to that node, triggering a reconnect and a PING over the new link.
                  * Once that node receives our PING, it should recognize the new connection as an inbound link from me.
                  * We should only free the outbound link if the node is known for more time than the handshake timeout,
                  * since during this time, the other side might still be trying to complete the handshake. */
 
                 /* We should always receive a MEET packet on an inbound link. */
```
PingXie, line 3273: super nit - maybe wrap around 100 chars? these lines are within the recommended 120 column width but they seem to stand out quite a bit

src/cluster_legacy.c
```diff
@@ -3239,42 +3221,59 @@ int clusterProcessPacket(clusterLink *link) {
...
+                    /* Add this node if it is new for us and the msg type is MEET.
+                     * In this stage we don't try to add the node with the right
+                     * flags, replicaof pointer, and so forth, as this details will be
+                     * resolved when we'll receive PONGs from the node. The exception
+                     * to this is the flag that indicates extensions are supported, as
+                     * we want to send extensions right away in the return PONG in order
+                     * to reduce the amount of time needed to stabilize the shard ID. */
+                    clusterNode *node = createClusterNode(NULL, CLUSTER_NODE_HANDSHAKE);
+                    if (nodeIp2String(node->ip, link, hdr->myip) != C_OK) {
+                        /* We cannot get the IP info from the link, it probably means the connection is closed. */
```
PingXie, line 3234:
```suggestion
                        /* Unable to retrieve the node's IP address from the connection. Without a valid IP, the node 
                         * becomes unusable in the cluster. This failure might be due to the connection being closed. 
                         * To avoid leaving the cluster in an inconsistent state, we close the link and remove the node. */
```

src/cluster_legacy.c
```diff
@@ -3239,42 +3221,59 @@ int clusterProcessPacket(clusterLink *link) {
...
+                        serverLog(LL_NOTICE, "Closing link even though we received a MEET packet on it, "
+                                             "because the connection has an error");
```
PingXie, line 3236:
```suggestion
                        serverLog(LL_NOTICE, "Closing link and removing node due to failure to retrieve IP from the connection, "
                                             "possibly caused by a closed connection.");
```

PingXie (approved):
Some minor wording changes in the code comments and logs. Thanks @pieturin!

PingXie:
> And then the `cluster-reliable-meet` test fails because the node with the inconsistent view never sends the MEET (not sure why yet though)

I think I know why. The `handshake_start_time` is lost in this case along with the node that is created as a response to the first `MEET`. The node is then recreated via the gossip route, but this new node never initiates a handshake, so its `handshake_start_time` is never set. Here is the possible sequence:

1. A cluster with two primaries, `P1` and `P2`.
2. A new node `N` is to be added.
3. `P1` is instructed to meet `N`, but `P1` is also told to drop all cluster traffic from `N`. Therefore, from `N`'s perspective, `P1` remains in the handshake state.
4. Eventually, `P1`'s representation on `N`, a `clusterNode` instance, times out and gets deleted.
5. However, when `P1` sends out its first `MEET`, it also gossips about `P2` to `N`, which shows up as a normal node on `N`.
6. `P2` then gossips `P1` back to `N`, now as a normal node! This is why `handshake_start_time` is never initialized for this new version of `P1`'s representation on `N`.

I think this is just another argument for #1471.

PR #1442, opened 2024-12-14
Avoid importing memory aligned malloc

### LINKED ISSUE #1434: Santitizer builds "error: 'malloc' is deprecated"
In some builds, we get this build error:

```
In file included from hyperloglog.c:39:
In file included from /usr/lib/llvm-18/lib/clang/18/include/immintrin.h:26:
In file included from /usr/lib/llvm-18/lib/clang/18/include/xmmintrin.h:31:
/usr/lib/llvm-18/lib/clang/18/include/mm_malloc.h:35:12: error: 'malloc' is deprecated [-Werror,-Wdeprecated-declarations]
   35 |     return malloc(__size);
      |            ^
./server.h:4023:42: note: 'malloc' has been explicitly marked deprecated here
 4023 | void *malloc(size_t size) __attribute__((deprecated));
      |                                          ^
```

https://github.com/valkey-io/valkey/actions/runs/12287187355/job/34288772967

We mark malloc as deprecated to be sure that zmalloc is used everywhere? Can't we override malloc, so malloc automatically calls zmalloc? I think there are various ways to do that.

madolson:
We deprecate the usage of classic malloc and free, but under certain circumstances they might get imported from intrinsics. The original thought is we should just override malloc and free to use zmalloc and zfree, but I think we should continue to deprecate it to avoid accidental imports of allocations.

Closes https://github.com/valkey-io/valkey/issues/1434.

### DIFF
--- a/src/hyperloglog.c
+++ b/src/hyperloglog.c
@@ -36,6 +36,8 @@
 #include <math.h>
 
 #ifdef HAVE_AVX2
+/* Define __MM_MALLOC_H to prevent importing the memory aligned
+ * allocation functions, which we don't use. */
 #define __MM_MALLOC_H
 #include <immintrin.h>
 #endif
--- a/src/hyperloglog.c
+++ b/src/hyperloglog.c
@@ -36,6 +36,7 @@
 #include <math.h>
 
 #ifdef HAVE_AVX2
+#define __MM_MALLOC_H
 #include <immintrin.h>
 #endif
 


enjoy-binbin (approved):

src/hyperloglog.c
```diff
@@ -36,6 +36,9 @@
 #include <math.h>
 
 #ifdef HAVE_AVX2
+/* Define __MM_MALLOC_H to prevent importing the memory aligned
+ * allocation functions, which we don't use. */
+#define __MM_MALLOC_H
```
zuiderkwast, line 41: Do all compilers use the same name for this macro or is it just clang?

zuiderkwast: If it's only for clang, then it feels like a workaround just for that, but let's merge this for now to silence the CI.

zuiderkwast (approved):
malloc/free are used in deps like Lua though, aren't they?

Issue #68, opened 2024-03-28
[NEW] Support for Empty SET


nmvk:
**The problem/use-case that the feature addresses**
Currently, a key is deleted when the set is empty. Distinguishing between an empty set and a key that does not exist offers many advantages. For example, if we store SQL results in a set, an empty set can represent that a query was executed but the result was empty, whereas null might imply that we need to make a database call to store in the set.

**Description of the feature**
Without breaking any backward compatibilities we scan support the Empty set with set of new commands and arguments

**Write commands**
`SADD key`
Existing SADD would be modified to make members as optional, invoking SADD with only key will create an empty set.

`SPOP key [count] MT` → _Additional argument_
Add new optional argument to retain the SET if no element exist.

`SREMMT key member [member ...]` → _[New Command]_
Same as SREM would keep the empty set once all members are removed.

`SMOVEMT` , `SINTERSTOREMT`, `SDIFFSTOREMT`, and `SUNIONSTOREMT` would be implemented as followups if needed. One can delete the empty set using `DEL` command.

**Read commands**
`SCARD` would still return 0 for empty set. To distinguish empty set from a non existing key one can use `EXISTS` followed by `SCARD`. Same can be approach for the `SISMEMBER` and `SMISMEMBER`. 

`SMEMBERSMT key` → _[New Command]_
Similar to `SMEMBERS` but will return null when key does not exist. This command can be achieved by using `EXISTS` + `SMEMBERS` but would be supported for ease of use. 

**Alternatives you've considered**
Provide a new param to change the default behavior.

hpatro:
This is another feature which had quite a strong demand https://github.com/redis/redis/issues/6048 . I think we should make a decision to distinguish between empty vs NULL for data type like set.

@valkey-io/core-team

sungming2:
I am looking into this issue

zuiderkwast:
I'm skeptical. If it had been like that from the beginning that commands like SPOP leaves an empty set, then that would have been much better. Then SPOP, HDEL, ZREM and other commands would just not delete a key. Only DEL would delete a key.

But for historical reasons, we have what we have. Adding empty sets now and at the same keeping the default behavior of deleting the key when it becomes empty has problems:

* Complexity. Duplicated commands and new arguments to existing commands.
* Risk of confusion. Although it may be useful to some, this can cause confusions to users. I believe the simplicity is a key success factor.

[...]

PR #1444, opened 2024-12-16
Support empty set

sungming2:
### Summary
This pull request introduces support for empty sets in Valkey, controlled by a new configuration option, allow-empty-set which is a strong demand in the Redis community (https://github.com/redis/redis/issues/6048). With this change, Valkey can preserve keys for sets even after all their elements have been removed, rather than deleting the key immediately. This feature provides additional flexibility for use cases where an empty set has semantic meaning

### Why This Change is Needed
In Valkey, when a set becomes empty (all its elements are removed), the key representing that set is automatically deleted from the database. While this behavior works for most use cases, it introduces limitations for certain scenarios where the distinction between an empty set (existent key) and a non-existent key is important.
For example:
Key: `active_users`
```
SADD active_users sungming 
SREM active_users sungming
EXISTS active_users 
> (integer) 0
```
Now in this case, the absence of the key (active_users) could mean either:
* The query hasn't been executed yet.
* The query was executed but returned no users.

The application would need to query its persistent database (e.g., SQL) to determine whether the key represents an empty set or has not been created yet. Alternatively, they could add a dummy value, such as an empty string (""), to the set to ensure the key persists. However, this approach is not ideal and could lead to unintended behavior or safety issues.

### Key changes
1. New Configuration Option (allow-empty-set):
Added the allow-empty-set configuration to valkey.conf.
Default: ```no``` (empty sets are deleted by default, preserving backward compatibility).
When enabled, keys for empty sets are retained in the database until explicitly deleted.

3. Database Behavior:
Empty sets are represented as valid keys with zero cardinality (SCARD returns 0).
These keys are preserved in memory and are only removed with the DEL command.

4. Persistence:
Empty sets are serialized and saved in the RDB file when allow-empty-set is enabled.
During RDB/AOF loading, empty sets are restored only if allow-empty-set is enabled. Otherwise, they are skipped.

### Test
Added unit tests and integration tests (TCL) to verify:
1. Persistence and RDB reloading of empty sets.
2. Correct behavior of set commands when allow-empty-set is enabled/disabled.
3. Compatibility with default behavior when the option is disabled.

### Backward Compatibility
The default behavior (allow-empty-set = no) ensures backward compatibility with existing applications.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -3206,6 +3206,7 @@ standardConfig static_configs[] = {
     createBoolConfig("cluster-slot-stats-enabled", NULL, MODIFIABLE_CONFIG, server.cluster_slot_stats_enabled, 0, NULL, NULL),
     createBoolConfig("hide-user-data-from-log", NULL, MODIFIABLE_CONFIG, server.hide_user_data_from_log, 1, NULL, NULL),
     createBoolConfig("import-mode", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.import_mode, 0, NULL, NULL),
+    createBoolConfig("allow-empty-set", NULL, MODIFIABLE_CONFIG, server.allow_empty_set, 0, NULL, NULL),
 
     /* String Configs */
     createStringConfig("aclfile", NULL, IMMUTABLE_CONFIG, ALLOW_EMPTY_STRING, server.acl_filename, "", NULL, NULL),
--- a/tests/unit/type/set.tcl
+++ b/tests/unit/type/set.tcl
@@ -1168,6 +1168,111 @@ foreach type {single multiple single_multiple} {
     }
 }
 
+start_server {
+    tags {"set external:skip"}
+    overrides {
+        "set-max-intset-entries" 1
+        "allow-empty-set" "yes"
+    }
+} {
+    proc save_load_rdb {{config_override {}}} {
+        r save
+        restart_server 0 true false true now $config_override
+        wait_done_loading r
+    }
+    proc create_emptyset {key members} {
+        r del $key
+        foreach member $members { r sadd $key $member }
+        foreach member $members { r srem $key $member }
+    }
+    proc verify_emptyset {key} {
+        assert_equal 0 [r scard $key]
+        assert_equal 1 [r exists $key]
+    }
+
+    test {EMTPYSET save and reload empty listset from RDB} {
+        r flushdb
+        create_emptyset listset {a}
+        verify_emptyset listset
+        assert_equal listpack [r object encoding listset]
+        save_load_rdb
+        verify_emptyset listset
+    }
+
+    test {EMTPYSET save and reload empty intset from RDB} {
+        r flushdb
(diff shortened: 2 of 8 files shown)


src/config.c
```diff
@@ -3206,6 +3206,7 @@ standardConfig static_configs[] = {
     createBoolConfig("cluster-slot-stats-enabled", NULL, MODIFIABLE_CONFIG, server.cluster_slot_stats_enabled, 0, NULL, NULL),
     createBoolConfig("hide-user-data-from-log", NULL, MODIFIABLE_CONFIG, server.hide_user_data_from_log, 1, NULL, NULL),
     createBoolConfig("import-mode", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.import_mode, 0, NULL, NULL),
+    createBoolConfig("allow-empty-set", NULL, MODIFIABLE_CONFIG, server.allow_empty_set, 0, NULL, NULL),
```
madolson, line 3209: I don't think we want this as a configuration. Application developers should be the ones that get to decided what behavior they want related to empty sets.

sungming2: Thanks for the comment.
I considered introducing new commands, but IMO, there are potential downsides:
1. New commands like SREMMT, SMOVEMT, and SPOPMT could increase the complexity of command management. This might lead to further requests to implement similar "MT" commands for other data types, such as ```lists``` and ```sorted sets```, adding more maintenance overhead.
2. Using a configuration setting is more flexible and could be extended to support other data types in the future like ```allow-empty-object```, making it a more scalable solution. 

Could I get thoughts on this?

tests/unit/type/set.tcl
```diff
@@ -1168,6 +1168,111 @@ foreach type {single multiple single_multiple} {
...
+    }
+} {
+    proc save_load_rdb {{config_override {}}} {
+        r save
+        restart_server 0 true false true now $config_override
+        wait_done_loading r
+    }
+    proc create_emptyset {key members} {
+        r del $key
+        foreach member $members { r sadd $key $member }
```
murphyjacob4, line 1185: I don't personally like that in order to create an empty set, you first need to add an arbitrary element then delete that element. The primary use case for this feature seems to be for query caching, where really your goal is to empty the set (or create a new empty set)

I left a comment in the issue proposing a new API - `SEMPTY` or something similar (maybe `SCREATE`). Even if we decide on a different approach than what I was suggesting there - I think an API like this will be needed to make it user friendly

PingXie:
I marked this as "major-decision-needed" and left some comments on #68.

@valkey-io/core-team FYI

hwware:
> I marked this as "major-decision-needed" and left some comments on #68.
> 
> @valkey-io/core-team FYI

I just left comment there as well.

ranshid:
@valkey-io/client-maintainers -  this is currently on the "needs review" for 9.2 release. I think we decided to drop it for now, correct? I will remove the 9.2 project for now, bit if I made a mistake please feel free to correct me

PR #1449, opened 2024-12-17
Offload reading the replication stream to IO threads

uriyage:
Support Primary client IO offload.

Related issue: https://github.com/valkey-io/valkey/issues/761

### DIFF
--- a/src/io_threads.c
+++ b/src/io_threads.c
@@ -322,7 +322,7 @@ int trySendReadToIOThreads(client *c) {
     /* If IO thread is already reading, return C_OK to make sure the main thread will not handle it. */
     if (c->io_read_state != CLIENT_IDLE) return C_OK;
     /* Currently, replica/master writes are not offloaded and are processed synchronously. */
-    if (c->flag.primary || getClientType(c) == CLIENT_TYPE_REPLICA) return C_ERR;
+    if (getClientType(c) == CLIENT_TYPE_REPLICA) return C_ERR;
     /* With Lua debug client we may call connWrite directly in the main thread */
     if (c->flag.lua_debug) return C_ERR;
     /* For simplicity let the main-thread handle the blocked clients */
@@ -345,6 +345,7 @@ int trySendReadToIOThreads(client *c) {
     c->cur_tid = tid;
     c->read_flags = canParseCommand(c) ? 0 : READ_FLAGS_DONT_PARSE;
     c->read_flags |= authRequired(c) ? READ_FLAGS_AUTH_REQUIRED : 0;
+    c->read_flags |= c->flag.primary ? READ_FLAGS_PRIMARY : 0;
 
     c->io_read_state = CLIENT_PENDING_IO;
     connSetPostponeUpdateState(c->conn, 1);
@@ -364,7 +365,7 @@ int trySendWriteToIOThreads(client *c) {
     /* Nothing to write */
     if (!clientHasPendingReplies(c)) return C_ERR;
     /* Currently, replica/master writes are not offloaded and are processed synchronously. */
-    if (c->flag.primary || getClientType(c) == CLIENT_TYPE_REPLICA) return C_ERR;
+    if (getClientType(c) == CLIENT_TYPE_REPLICA) return C_ERR;
     /* We can't offload debugged clients as the main-thread may read at the same time  */
     if (c->flag.lua_debug) return C_ERR;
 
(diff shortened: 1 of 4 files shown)


src/io_threads.c
```diff
@@ -364,7 +365,7 @@ int trySendWriteToIOThreads(client *c) {
     /* Nothing to write */
     if (!clientHasPendingReplies(c)) return C_ERR;
     /* Currently, replica/master writes are not offloaded and are processed synchronously. */
```
zuiderkwast: The comment needs an update. Write to primary is now offloaded.

To replicas, the writes are still non-blocking IO, not synchronous IO as a reader might think. Perhaps we just don't need to mention that part.

```suggestion
    /* Currently, replica writes are not offloaded to IO threads. */
```

zuiderkwast (approved):
That was a small diff! Nice.

zuiderkwast:
The PR title isn't very clear for someone who doesn't know the internals.

I believe a user could better understand a title like "Offload reading the replication stream to IO threads".

src/networking.c
```diff
@@ -4954,7 +4954,9 @@ void ioThreadReadQueryFromClient(void *data) {
     }
 
 done:
-    trimClientQueryBuffer(c);
+    if (!(c->read_flags & READ_FLAGS_PRIMARY)) {
```
madolson, line 4969: I think this is worth a comment explaining why we don't want to trim the client query buffer. The special copying between buffers is not obvious.

src/replication.c
```diff
@@ -4154,6 +4156,13 @@ void replicationCachePrimary(client *c) {
     c->bufpos = 0;
     resetClient(c);
 
+    /* Reset the primary IO state. */
+    c->nwritten = 0;
+    c->nread = 0;
+    c->io_read_state = c->io_write_state = CLIENT_IDLE;
+    c->io_parsed_cmd = NULL;
+    c->flag.pending_command = 0;
+
```
ranshid: I know this is the only case we are CURRENTLY using it but might help wrapping it in a dedicated resetClientIOState IMO

uriyage:
> The PR title isn't very clear for someone who doesn't know the internals.
> 
> I believe a user could better understand a title like "Offload reading the replication stream to IO threads".

Thanks, changed.

src/networking.c
uriyage: Done

src/io_threads.c
uriyage: Thanks, changed.

src/replication.c
uriyage: Done.

zuiderkwast (approved):
LGTM

I will not merge it yet in case others want to review it a bit more.

jdheyburn:
To the layman such as myself, what benefit does this introduce WRT replication?

zuiderkwast:
> To the layman such as myself, what benefit does this introduce WRT replication?

Replicas can serve more readonly traffic when reading the replication stream can be offloaded to an IO thread instead of being handled by the bottleneck main thread. It's useful in setups where replicas are used for readonly traffic.

It seems more beneficial to let primaries offload the writes to replicas to IO threads. Hopefully, that will come in another PR. @uriyage WDYT?

madolson (approved):

uriyage:
> > To the layman such as myself, what benefit does this introduce WRT replication?
> 
> Replicas can serve more readonly traffic when reading the replication stream can be offloaded to an IO thread instead of being handled by the bottleneck main thread. It's useful in setups where replicas are used for readonly traffic.
> 
> It seems more beneficial to let primaries offload the writes to replicas to IO threads. Hopefully, that will come in another PR. @uriyage WDYT?

@zuiderkwast Yes, I will open a new PR in the coming days for offloading the writes to replicas. It is a bit more complicated, which is why I started with the primary writes offload.

enjoy-binbin (approved):

ranshid (approved):

ranshid:
@uriyage I think we need to check the failed test in TLS which I can see:
```
Error writing to client: error:0A00010F:SSL routines::bad length
```
Maybe it is related (is it the use of writev?)

ranshid:
I rerun the test to see if this is a stable fail

uriyage:
> I rerun the test to see if this is a stable fail

The test runs without IO threads, in this case, the PR changes have no effect, meaning the failure is unrelated to this PR changes

zuiderkwast:
I don't want to block PRs that are ready to merge. We should really focus on the failing tests instead of starting too many new features though.

PR #1451, opened 2024-12-17
Fix undefined behavior defined by ASAN

madolson:
Asan now supports making sure you are passing in the correct pointer type, which seems useful but we can't support it since we pass in an incorrect pointer in several places. This is most commonly done with generic free functions, where we simply cast it to the correct type.

Example: https://github.com/valkey-io/valkey/actions/runs/12334182647/job/34424037283#step:6:318.

It's not a lot of code to clean up, so it seems appropriate to cleanup instead of disabling the check.

Validating run: https://github.com/valkey-io/valkey/actions/runs/12382893783/job/34564554891


src/adlist.c
```diff
@@ -77,6 +77,12 @@ void listRelease(list *list) {
     zfree(list);
 }
 
+/* Just like listRelease, but takes in a into a generic free
+ * function that expects (void *). */
```
zuiderkwast: Some "in a into a" typo here.

```suggestion
/* Just like listRelease, but takes the list as a (void *).
 * Useful as generic free callback. */
```

zuiderkwast (approved):
Great!

PR #1452, opened 2024-12-17
Attempt to read secondary error from info test

madolson:
The test attempts to write 1MB of data in order to trigger a disconnect. Normally, the data is fully flushed and we get the error on the read (I/O error). However, it's possible we might fail the write, which leaves the client in an inconsistent state. On the next command, we finally process the I/O error on the FD. So, the simple fix is to consume any secondary errors.

https://github.com/valkey-io/valkey/actions/runs/12363680454/job/34505497852


tests/unit/info.tcl
```diff
@@ -392,6 +392,7 @@ start_server {tags {"info" "external:skip" "debug_defrag:skip"}} {
             set org_qbuf_limit [lindex [r config get client-query-buffer-limit] 1]
             r config set client-query-buffer-limit 1048576
             catch {r set key [string repeat a 1048576]}
+            catch {r read} ; # We sometimes get a broken pipe and I/O error, so make sure there are no more errors
```
zuiderkwast: Nit. The PR top comment is much more comprehensive than this comment. Can you explain both failure modes in a code comments? (write succeeds and read fails, or write fails and the next read fails too as a second error)

zuiderkwast (approved):
Great! Let's merge this so we can make daily green.

Is there a potentially better solution where `r` handles this internally?

madolson:
> Is there a potentially better solution where r handles this internally?

Probably? I wasn't really able to follow the error handling logic in the client, which had special handling for I/O errors but not for other failure modes. I'll spend a few more minutes to see if I can better squash it there.

EDIT: I think the client is just not well thought out. Re-opening the connection after an error feels fragile today and attempting to catch the I/O error in other ways is really difficult since we don't know the state (we can get stuck reads for example).

tests/unit/info.tcl
madolson: Sure

enjoy-binbin (approved):

zuiderkwast (approved):

tests/unit/info.tcl
```diff
@@ -391,7 +391,13 @@ start_server {tags {"info" "external:skip" "debug_defrag:skip"}} {
             # set qbuf limit to minimum to test stat
             set org_qbuf_limit [lindex [r config get client-query-buffer-limit] 1]
             r config set client-query-buffer-limit 1048576
-            catch {r set key [string repeat a 1048576]}
+            catch {r set key [string repeat a 2048576]} e
```
hpatro, line 394: Why did we almost double the amount of data to be written? And what's the reason for such errors (is it well understood)?

madolson: Yeah, this just makes the failure mode I discuss happen more consistently since we have lots more to write after th server disconnects. So if my change doesn't work somewhere it'll be obvious.

PR #1455, opened 2024-12-18
Add GETPXT, MGETPXT (Get with millisecond expiration) commands

arcivanov:
Distributed cache synchronization is an eternal problem. Luckily, currently (2024), cloud deployed services provide high quality time synchronization for deployed resources, allowing precision timing coordination with sub-millisecond precision. This command allows to take advantage of this fact to implement simple distributed cache synchronization.

Using absolute Unix millisecond timestamps for synchronization allows machine-local caches that feed from a shared Valkey cache in the cluster to expire the local entries within 1ms of each other, assuming the sub-ms clock synchronization, providing no-cost distributed cache synchronization.

Unfortunately, Valkey does not have commands that allow to conveniently (and atomically!) retrieve a value and the accompanying expiration timestamp in a single command. The workaround exists allowing to use GET followed by PEXPIRETIME, pipelined. This naturally involves additional processing including checking for the key's existence twice.

This patch introduces a new command GETPXT (GET + PeXpireTime), which behaves as documented below and elsewhere in the commit.

-----
Returns `null` if the value not found or expired.
Returns an array of length 2 as `[<string key value>, <integer expiration>]`. If expiration is not set on the key, expiration returned is `-1`.

Added tests to cover


madolson:
Is your application bottlenecked on throughput? The use case generally makes sense to me, but I would rather avoiding adding a command if we can optimize the existing codepath. I will say that fetching the expire time is getting a lot more efficient with the new cache efficient dictionary we are releasing with 8.1 (hopefully), so maybe it's not as critical.

Another note, there is some overlap in naming with `GETEX`, which is GET + Set expiry.

arcivanov:
Two commands, apart from the overhead of processing two commands and potentially non-atomic nature of the retrieval (I'm not sure here as far as architecture - are pipelined commands atomic?), will end up calling `lookupKeyReadOrReply` (https://github.com/valkey-io/valkey/blob/unstable/src/t_string.c#L381C14-L381C34) and `lookupKeyReadWithFlags` (https://github.com/valkey-io/valkey/blob/unstable/src/expire.c#L726C9-L726C31) needlessly. There is only one lookup in the proposed patch.

I am not married to a command name and considered adding an option to the GET command, although this being the most frequently used command it would introduce parsing overhead (however small). Looking for further guidance.

madolson:
>are pipelined commands atomic?

Pipeline commands are not atomic, but you can send them in a multi-exec which does make them a transaction.

> There is only one lookup in the proposed patch.

Understood. However, on modern hardware the bottleneck is mostly on DRAM latency if you have a miss in the CPU cache, so practically the double miss won't matter much. Give me one sec, I'll come up with a procedure you can use to quickly test the performance.

arcivanov:
> Pipeline commands are not atomic, but you can send them in a multi-exec which does make them a transaction.

Yep, already using pipeline'd MULTI/GET/PEXPIRETIME/EXEC. The EXEC is marked "slow" in documentation. 
Additionally, now that I think of it, there should be a WATCH for the key in question in case the value OR the expiration time are modified while it's being retrieved.

The above scheme, however, is entirely avoided if an atomic command is used :smile:

zuiderkwast:
> The EXEC is marked "slow" in documentation.

EXEC itself is not slow. It's basically a no-op.

MULTI-EXEC works in a very simple way. The commands are queued up and when EXEC is called, all the commands are executed together and this makes it atomic.

So, you could say EXEC is slow, because it executes all the commands that have been queued up. If you have actually slow commands in the transaction, they will make EXEC slow.

The "slow" flag in the documentation is mainly confusing and I wish we could just delete this in the documentation. Currently, every command that isn't explicitly marked as "fast" are marked as "slow".

> Additionally, now that I think of it, there should be a WATCH for the key in question in case the value OR the expiration time are modified while it's being retrieved.

I don't get this. MULTI/GET/PEXPIRETIME/EXEC is atomic, just like your proposed GETPXT. No need for WATCH, or am I missing something?

arcivanov:
> I don't get this. MULTI/GET/PEXPIRETIME/EXEC is atomic, just like your proposed GETPXT. No need for WATCH, or am I missing something?

Ah, nevermind, I read into the documentation - WATCH is only needed when used before the transaction is started with MULTI. Withdrawn.

zuiderkwast:
Since MULTI-EXEC can do this atomically, we will most likely not accept this new command without another good reason, something that MULTI-EXEC doesn't solve. Is the performance of MULTI-EXEC not good enough?

arcivanov:
@zuiderkwast well, there are atomic sets (`SET key value PXAT tsms` and `GETEX`), all of which could be accomplished with MULTI-EXEC, and yet these commands are in the system. It seems logical that there are atomic counterparts on the `GET` side. Additionally, `MGET` can be accomplished by pipelining `GET`s inside `MULTI-EXEC`, yet the command is still there along with numerous others that could be accomplished via pipelined transactions or scripting.

 > Is the performance of MULTI-EXEC not good enough?
 
I will conduct the formal performance tests but given that a MULTI-EXEC will result in more parsing, more wire traffic and two lookups instead of one, I doubt the performance will be identical. Our goal is to fit as much performance as possible into the smallest machine possible to save costs.

zuiderkwast:
@arcivanov Thanks, so I can summarize you points like this:

1. Performance
2. Symmetry with other commands (SET PXAT, GETEX)
3. Convenience

In the past when there was a BDFL governance, commands got added based on personal taste. In the last years, we have been more restrictive in adding new commands, especially when the same effect can already be achieved. There's no absolute rule about this though.

I'm looking forward to see your performance results. To me, it seems like the most convincing reason to add this. My guess is that a transaction is slower, but not many times slower.

zuiderkwast:
Are you querying multiple keys at the same time?

If yes, then there's an alternative to consider. MGET can be used to get multiple keys. If we can allow PEXPIRETIME to return the timestamp for multiple keys, then you could fetch this info for multiple keys like

    MULTI
    MGET key1 key2 key3
    PEXPIRETIME key1 key2 key3
    EXEC

arcivanov:
These are very preliminary and I'll push the code once I clean it up a little bit but here is the data:

Average latency for `MULTI-GET-PEXPIRE-EXEC` is 20% slower than `GETPXT`.

Again, this is very preliminary and I need to exclude a few potential sources of error in the bench (namely `GET key:__rand_int__\r\nPEXPIRETIME key:__rand_int__` accessing different records).

```
$ src/valkey-benchmark -t set_pxat,getpxt,getpxt_simulated -n 3000000 -r 3000000 -c 100
====== SET w/ PXAT ======                                                     
  3000000 requests completed in 29.06 seconds
  100 parallel clients
  3 bytes payload
  keep alive: 1
  multi-thread: no

0.000% <= 0.111 milliseconds (cumulative count 2)
50.000% <= 0.471 milliseconds (cumulative count 1557625)
75.000% <= 0.511 milliseconds (cumulative count 2260843)
87.500% <= 0.655 milliseconds (cumulative count 2634115)
93.750% <= 0.887 milliseconds (cumulative count 2813469)
96.875% <= 1.127 milliseconds (cumulative count 2907931)
98.438% <= 1.391 milliseconds (cumulative count 2953950)
99.219% <= 1.711 milliseconds (cumulative count 2976925)
99.609% <= 2.055 milliseconds (cumulative count 2988424)
99.805% <= 2.247 milliseconds (cumulative count 2994159)
99.902% <= 2.527 milliseconds (cumulative count 2997104)
99.951% <= 2.711 milliseconds (cumulative count 2998582)
99.976% <= 2.879 milliseconds (cumulative count 2999278)
99.988% <= 3.079 milliseconds (cumulative count 2999641)
99.994% <= 3.343 milliseconds (cumulative count 2999818)
99.997% <= 15.079 milliseconds (cumulative count 2999909)
99.998% <= 16.055 milliseconds (cumulative count 2999955)
99.999% <= 16.655 milliseconds (cumulative count 2999978)
100.000% <= 16.959 milliseconds (cumulative count 2999989)
100.000% <= 17.135 milliseconds (cumulative count 2999995)
100.000% <= 17.215 milliseconds (cumulative count 2999998)
100.000% <= 17.263 milliseconds (cumulative count 2999999)
100.000% <= 17.327 milliseconds (cumulative count 3000000)
100.000% <= 17.327 milliseconds (cumulative count 3000000)

0.001% <= 0.207 milliseconds (cumulative count 18)
0.035% <= 0.303 milliseconds (cumulative count 1062)
0.711% <= 0.407 milliseconds (cumulative count 21343)
73.339% <= 0.503 milliseconds (cumulative count 2200176)
85.490% <= 0.607 milliseconds (cumulative count 2564694)
89.396% <= 0.703 milliseconds (cumulative count 2681874)
92.088% <= 0.807 milliseconds (cumulative count 2762641)
94.091% <= 0.903 milliseconds (cumulative count 2822718)
95.691% <= 1.007 milliseconds (cumulative count 2870722)
96.718% <= 1.103 milliseconds (cumulative count 2901555)
97.539% <= 1.207 milliseconds (cumulative count 2926177)
98.088% <= 1.303 milliseconds (cumulative count 2942649)
98.521% <= 1.407 milliseconds (cumulative count 2955644)
98.806% <= 1.503 milliseconds (cumulative count 2964186)
99.043% <= 1.607 milliseconds (cumulative count 2971305)
99.218% <= 1.703 milliseconds (cumulative count 2976552)
99.364% <= 1.807 milliseconds (cumulative count 2980910)
99.465% <= 1.903 milliseconds (cumulative count 2983938)
99.549% <= 2.007 milliseconds (cumulative count 2986477)
99.694% <= 2.103 milliseconds (cumulative count 2990817)
99.989% <= 3.103 milliseconds (cumulative count 2999658)
99.997% <= 4.103 milliseconds (cumulative count 2999899)
99.997% <= 5.103 milliseconds (cumulative count 2999900)
99.997% <= 15.103 milliseconds (cumulative count 2999909)
99.999% <= 16.103 milliseconds (cumulative count 2999957)
100.000% <= 17.103 milliseconds (cumulative count 2999993)
100.000% <= 18.111 milliseconds (cumulative count 3000000)

Summary:
  throughput summary: 103234.69 requests per second
  latency summary (msec):
        0.544     0.104     0.471     0.959     1.591    17.327
====== GETPXT ======                                                     
  3000000 requests completed in 28.71 seconds
  100 parallel clients
  3 bytes payload
  keep alive: 1
  multi-thread: no

0.000% <= 0.111 milliseconds (cumulative count 2)
50.000% <= 0.479 milliseconds (cumulative count 1815098)
75.000% <= 0.495 milliseconds (cumulative count 2323502)
87.500% <= 0.527 milliseconds (cumulative count 2629031)
93.750% <= 0.639 milliseconds (cumulative count 2812974)
96.875% <= 0.807 milliseconds (cumulative count 2906398)
98.438% <= 0.983 milliseconds (cumulative count 2954330)
99.219% <= 1.271 milliseconds (cumulative count 2976796)
99.609% <= 1.927 milliseconds (cumulative count 2988347)
99.805% <= 2.151 milliseconds (cumulative count 2994232)
99.902% <= 2.479 milliseconds (cumulative count 2997082)
99.951% <= 2.599 milliseconds (cumulative count 2998588)
99.976% <= 2.711 milliseconds (cumulative count 2999283)
99.988% <= 2.871 milliseconds (cumulative count 2999641)
99.994% <= 3.103 milliseconds (cumulative count 2999817)
99.997% <= 3.311 milliseconds (cumulative count 2999909)
99.998% <= 3.703 milliseconds (cumulative count 2999955)
99.999% <= 3.959 milliseconds (cumulative count 2999979)
100.000% <= 4.047 milliseconds (cumulative count 2999989)
100.000% <= 4.255 milliseconds (cumulative count 2999995)
100.000% <= 4.303 milliseconds (cumulative count 2999998)
100.000% <= 4.607 milliseconds (cumulative count 2999999)
100.000% <= 4.631 milliseconds (cumulative count 3000000)
100.000% <= 4.631 milliseconds (cumulative count 3000000)

0.001% <= 0.207 milliseconds (cumulative count 21)
0.031% <= 0.303 milliseconds (cumulative count 941)
0.619% <= 0.407 milliseconds (cumulative count 18574)
81.712% <= 0.503 milliseconds (cumulative count 2451346)
92.731% <= 0.607 milliseconds (cumulative count 2781945)
95.222% <= 0.703 milliseconds (cumulative count 2856660)
96.880% <= 0.807 milliseconds (cumulative count 2906398)
97.931% <= 0.903 milliseconds (cumulative count 2937935)
98.582% <= 1.007 milliseconds (cumulative count 2957449)
98.891% <= 1.103 milliseconds (cumulative count 2966733)
99.128% <= 1.207 milliseconds (cumulative count 2973827)
99.265% <= 1.303 milliseconds (cumulative count 2977947)
99.370% <= 1.407 milliseconds (cumulative count 2981106)
99.438% <= 1.503 milliseconds (cumulative count 2983153)
99.492% <= 1.607 milliseconds (cumulative count 2984764)
99.534% <= 1.703 milliseconds (cumulative count 2986029)
99.572% <= 1.807 milliseconds (cumulative count 2987145)
99.604% <= 1.903 milliseconds (cumulative count 2988106)
99.640% <= 2.007 milliseconds (cumulative count 2989204)
99.768% <= 2.103 milliseconds (cumulative count 2993028)
99.994% <= 3.103 milliseconds (cumulative count 2999817)
100.000% <= 4.103 milliseconds (cumulative count 2999992)
100.000% <= 5.103 milliseconds (cumulative count 3000000)

Summary:
  throughput summary: 104493.21 requests per second
  latency summary (msec):
        0.507     0.104     0.479     0.695     1.151     4.631
MULTI
GET key:__rand_int__
PEXPIRETIME key:__rand_int__
EXEC
====== GETPXT Simulated ======                                                     
  3000000 requests completed in 30.63 seconds
  100 parallel clients
  3 bytes payload
  keep alive: 1
  multi-thread: no

0.000% <= 0.231 milliseconds (cumulative count 2)
50.000% <= 0.535 milliseconds (cumulative count 1501247)
75.000% <= 0.655 milliseconds (cumulative count 2254474)
87.500% <= 0.807 milliseconds (cumulative count 2633915)
93.750% <= 0.943 milliseconds (cumulative count 2814682)
96.875% <= 1.095 milliseconds (cumulative count 2906403)
98.438% <= 1.295 milliseconds (cumulative count 2953727)
99.219% <= 1.607 milliseconds (cumulative count 2976665)
99.609% <= 2.111 milliseconds (cumulative count 2988500)
99.805% <= 2.487 milliseconds (cumulative count 2994164)
99.902% <= 2.831 milliseconds (cumulative count 2997092)
99.951% <= 3.103 milliseconds (cumulative count 2998553)
99.976% <= 3.311 milliseconds (cumulative count 2999278)
99.988% <= 3.495 milliseconds (cumulative count 2999647)
99.994% <= 3.671 milliseconds (cumulative count 2999819)
99.997% <= 16.879 milliseconds (cumulative count 2999909)
99.998% <= 17.583 milliseconds (cumulative count 2999955)
99.999% <= 17.887 milliseconds (cumulative count 2999978)
100.000% <= 17.999 milliseconds (cumulative count 2999989)
100.000% <= 18.079 milliseconds (cumulative count 2999996)
100.000% <= 18.111 milliseconds (cumulative count 2999998)
100.000% <= 18.127 milliseconds (cumulative count 2999999)
100.000% <= 18.143 milliseconds (cumulative count 3000000)
100.000% <= 18.143 milliseconds (cumulative count 3000000)

0.016% <= 0.303 milliseconds (cumulative count 481)
0.573% <= 0.407 milliseconds (cumulative count 17186)
35.339% <= 0.503 milliseconds (cumulative count 1060175)
68.272% <= 0.607 milliseconds (cumulative count 2048150)
80.013% <= 0.703 milliseconds (cumulative count 2400388)
87.797% <= 0.807 milliseconds (cumulative count 2633915)
92.577% <= 0.903 milliseconds (cumulative count 2777308)
95.385% <= 1.007 milliseconds (cumulative count 2861545)
96.987% <= 1.103 milliseconds (cumulative count 2909623)
97.999% <= 1.207 milliseconds (cumulative count 2939961)
98.491% <= 1.303 milliseconds (cumulative count 2954722)
98.836% <= 1.407 milliseconds (cumulative count 2965070)
99.061% <= 1.503 milliseconds (cumulative count 2971835)
99.222% <= 1.607 milliseconds (cumulative count 2976665)
99.328% <= 1.703 milliseconds (cumulative count 2979852)
99.405% <= 1.807 milliseconds (cumulative count 2982162)
99.461% <= 1.903 milliseconds (cumulative count 2983820)
99.509% <= 2.007 milliseconds (cumulative count 2985264)
99.607% <= 2.103 milliseconds (cumulative count 2988213)
99.952% <= 3.103 milliseconds (cumulative count 2998553)
99.997% <= 4.103 milliseconds (cumulative count 2999896)
99.997% <= 5.103 milliseconds (cumulative count 2999900)
99.997% <= 17.103 milliseconds (cumulative count 2999917)
100.000% <= 18.111 milliseconds (cumulative count 2999998)
100.000% <= 19.103 milliseconds (cumulative count 3000000)

Summary:
  throughput summary: 97930.40 requests per second
  latency summary (msec):
        0.612     0.224     0.535     0.991     1.479    18.143
```

arcivanov:
I have pushed the changes to the benchmark and here are the results done over the 1GbE network of Framework 16 client over LAN with `valkey-server` running on HP DL360G9 2xE5-2695v4 32-core 256GB. Built with regular `make` call with no march optimizations.

```
$ src/valkey-benchmark -t set_pxat,getpxt,getpxt_simulated -n 10000000 -r 3000000 -c 600 -h gh-x86-64-0002.local
====== SET w/ PXAT ======                                                     
  10000000 requests completed in 93.39 seconds
  600 parallel clients
  3 bytes payload
  keep alive: 1
  multi-thread: no

0.000% <= 1.535 milliseconds (cumulative count 1)
50.000% <= 4.767 milliseconds (cumulative count 5009571)
75.000% <= 5.415 milliseconds (cumulative count 7517549)
87.500% <= 6.071 milliseconds (cumulative count 8760775)
93.750% <= 6.615 milliseconds (cumulative count 9379676)
96.875% <= 7.399 milliseconds (cumulative count 9688551)
98.438% <= 8.287 milliseconds (cumulative count 9844120)
99.219% <= 8.943 milliseconds (cumulative count 9921979)
99.609% <= 9.823 milliseconds (cumulative count 9960975)
99.805% <= 10.903 milliseconds (cumulative count 9980493)
99.902% <= 12.199 milliseconds (cumulative count 9990247)
99.951% <= 13.655 milliseconds (cumulative count 9995136)
99.976% <= 14.839 milliseconds (cumulative count 9997562)
99.988% <= 16.415 milliseconds (cumulative count 9998785)
99.994% <= 18.543 milliseconds (cumulative count 9999390)
99.997% <= 20.959 milliseconds (cumulative count 9999698)
99.998% <= 23.695 milliseconds (cumulative count 9999849)
99.999% <= 25.359 milliseconds (cumulative count 9999924)
100.000% <= 31.023 milliseconds (cumulative count 9999962)
100.000% <= 31.471 milliseconds (cumulative count 9999981)
100.000% <= 33.215 milliseconds (cumulative count 9999991)
100.000% <= 33.407 milliseconds (cumulative count 9999997)
100.000% <= 33.471 milliseconds (cumulative count 9999998)
100.000% <= 212.479 milliseconds (cumulative count 10000000)
100.000% <= 212.479 milliseconds (cumulative count 10000000)

0.000% <= 1.607 milliseconds (cumulative count 10)
0.000% <= 1.703 milliseconds (cumulative count 30)
0.001% <= 1.807 milliseconds (cumulative count 122)
0.005% <= 1.903 milliseconds (cumulative count 530)
0.023% <= 2.007 milliseconds (cumulative count 2304)
0.073% <= 2.103 milliseconds (cumulative count 7280)
6.311% <= 3.103 milliseconds (cumulative count 631111)
20.516% <= 4.103 milliseconds (cumulative count 2051603)
66.060% <= 5.103 milliseconds (cumulative count 6605969)
88.080% <= 6.103 milliseconds (cumulative count 8808027)
96.145% <= 7.103 milliseconds (cumulative count 9614532)
98.132% <= 8.103 milliseconds (cumulative count 9813161)
99.327% <= 9.103 milliseconds (cumulative count 9932714)
99.678% <= 10.103 milliseconds (cumulative count 9967775)
99.827% <= 11.103 milliseconds (cumulative count 9982710)
99.898% <= 12.103 milliseconds (cumulative count 9989794)
99.936% <= 13.103 milliseconds (cumulative count 9993604)
99.962% <= 14.103 milliseconds (cumulative count 9996241)
99.977% <= 15.103 milliseconds (cumulative count 9997700)
99.986% <= 16.103 milliseconds (cumulative count 9998594)
99.990% <= 17.103 milliseconds (cumulative count 9998969)
99.992% <= 18.111 milliseconds (cumulative count 9999244)
99.995% <= 19.103 milliseconds (cumulative count 9999469)
99.996% <= 20.111 milliseconds (cumulative count 9999582)
99.997% <= 21.103 milliseconds (cumulative count 9999712)
99.998% <= 22.111 milliseconds (cumulative count 9999770)
99.998% <= 23.103 milliseconds (cumulative count 9999820)
99.999% <= 24.111 milliseconds (cumulative count 9999867)
99.999% <= 25.103 milliseconds (cumulative count 9999906)
99.999% <= 26.111 milliseconds (cumulative count 9999947)
100.000% <= 28.111 milliseconds (cumulative count 9999952)
100.000% <= 29.103 milliseconds (cumulative count 9999955)
100.000% <= 31.103 milliseconds (cumulative count 9999964)
100.000% <= 32.111 milliseconds (cumulative count 9999984)
100.000% <= 33.119 milliseconds (cumulative count 9999987)
100.000% <= 34.111 milliseconds (cumulative count 9999998)
100.000% <= 213.119 milliseconds (cumulative count 10000000)

Summary:
  throughput summary: 107083.58 requests per second
  latency summary (msec):
        4.869     1.528     4.767     6.815     8.703   212.479
====== GETPXT ======                                                     
  10000000 requests completed in 87.80 seconds
  600 parallel clients
  3 bytes payload
  keep alive: 1
  multi-thread: no

0.000% <= 0.743 milliseconds (cumulative count 2)
50.000% <= 3.551 milliseconds (cumulative count 5020352)
75.000% <= 4.439 milliseconds (cumulative count 7510858)
87.500% <= 4.975 milliseconds (cumulative count 8755434)
93.750% <= 5.583 milliseconds (cumulative count 9376140)
96.875% <= 6.055 milliseconds (cumulative count 9690462)
98.438% <= 6.519 milliseconds (cumulative count 9844744)
99.219% <= 7.247 milliseconds (cumulative count 9922143)
99.609% <= 7.855 milliseconds (cumulative count 9961079)
99.805% <= 8.303 milliseconds (cumulative count 9980558)
99.902% <= 8.863 milliseconds (cumulative count 9990261)
99.951% <= 9.783 milliseconds (cumulative count 9995134)
99.976% <= 11.071 milliseconds (cumulative count 9997563)
99.988% <= 11.671 milliseconds (cumulative count 9998781)
99.994% <= 14.071 milliseconds (cumulative count 9999390)
99.997% <= 16.207 milliseconds (cumulative count 9999695)
99.998% <= 17.743 milliseconds (cumulative count 9999848)
99.999% <= 18.671 milliseconds (cumulative count 9999925)
100.000% <= 19.439 milliseconds (cumulative count 9999963)
100.000% <= 21.711 milliseconds (cumulative count 9999981)
100.000% <= 22.207 milliseconds (cumulative count 9999991)
100.000% <= 22.383 milliseconds (cumulative count 9999996)
100.000% <= 22.447 milliseconds (cumulative count 9999998)
100.000% <= 22.479 milliseconds (cumulative count 10000000)
100.000% <= 22.479 milliseconds (cumulative count 10000000)

0.000% <= 0.807 milliseconds (cumulative count 9)
0.000% <= 0.903 milliseconds (cumulative count 27)
0.000% <= 1.007 milliseconds (cumulative count 38)
0.001% <= 1.103 milliseconds (cumulative count 51)
0.001% <= 1.207 milliseconds (cumulative count 68)
0.001% <= 1.303 milliseconds (cumulative count 85)
0.001% <= 1.407 milliseconds (cumulative count 102)
0.001% <= 1.503 milliseconds (cumulative count 122)
0.001% <= 1.607 milliseconds (cumulative count 142)
0.002% <= 1.703 milliseconds (cumulative count 204)
0.007% <= 1.807 milliseconds (cumulative count 682)
0.028% <= 1.903 milliseconds (cumulative count 2809)
0.103% <= 2.007 milliseconds (cumulative count 10282)
0.295% <= 2.103 milliseconds (cumulative count 29533)
35.943% <= 3.103 milliseconds (cumulative count 3594272)
64.626% <= 4.103 milliseconds (cumulative count 6462603)
89.190% <= 5.103 milliseconds (cumulative count 8919049)
97.138% <= 6.103 milliseconds (cumulative count 9713786)
99.112% <= 7.103 milliseconds (cumulative count 9911210)
99.730% <= 8.103 milliseconds (cumulative count 9973007)
99.921% <= 9.103 milliseconds (cumulative count 9992114)
99.959% <= 10.103 milliseconds (cumulative count 9995875)
99.977% <= 11.103 milliseconds (cumulative count 9997698)
99.989% <= 12.103 milliseconds (cumulative count 9998895)
99.992% <= 13.103 milliseconds (cumulative count 9999171)
99.994% <= 14.103 milliseconds (cumulative count 9999392)
99.996% <= 15.103 milliseconds (cumulative count 9999550)
99.997% <= 16.103 milliseconds (cumulative count 9999682)
99.998% <= 17.103 milliseconds (cumulative count 9999796)
99.999% <= 18.111 milliseconds (cumulative count 9999882)
99.999% <= 19.103 milliseconds (cumulative count 9999946)
100.000% <= 20.111 milliseconds (cumulative count 9999975)
100.000% <= 22.111 milliseconds (cumulative count 9999989)
100.000% <= 23.103 milliseconds (cumulative count 10000000)

Summary:
  throughput summary: 113896.51 requests per second
  latency summary (msec):
        3.758     0.736     3.551     5.751     6.975    22.479
====== GETPXT Simulated ======                                                     
  10000000 requests completed in 97.23 seconds
  600 parallel clients
  3 bytes payload
  keep alive: 1
  multi-thread: no

0.000% <= 1.671 milliseconds (cumulative count 1)
50.000% <= 4.783 milliseconds (cumulative count 5025523)
75.000% <= 5.495 milliseconds (cumulative count 7515973)
87.500% <= 6.255 milliseconds (cumulative count 8751646)
93.750% <= 6.791 milliseconds (cumulative count 9378528)
96.875% <= 7.319 milliseconds (cumulative count 9688869)
98.438% <= 7.943 milliseconds (cumulative count 9844871)
99.219% <= 8.455 milliseconds (cumulative count 9922561)
99.609% <= 8.935 milliseconds (cumulative count 9961099)
99.805% <= 9.463 milliseconds (cumulative count 9980635)
99.902% <= 10.103 milliseconds (cumulative count 9990304)
99.951% <= 11.087 milliseconds (cumulative count 9995145)
99.976% <= 12.871 milliseconds (cumulative count 9997561)
99.988% <= 14.103 milliseconds (cumulative count 9998786)
99.994% <= 19.279 milliseconds (cumulative count 9999392)
99.997% <= 34.879 milliseconds (cumulative count 9999695)
99.998% <= 37.951 milliseconds (cumulative count 9999849)
99.999% <= 39.935 milliseconds (cumulative count 9999926)
100.000% <= 42.015 milliseconds (cumulative count 9999963)
100.000% <= 43.231 milliseconds (cumulative count 9999981)
100.000% <= 43.391 milliseconds (cumulative count 9999991)
100.000% <= 43.519 milliseconds (cumulative count 9999997)
100.000% <= 43.551 milliseconds (cumulative count 9999999)
100.000% <= 43.647 milliseconds (cumulative count 10000000)
100.000% <= 43.647 milliseconds (cumulative count 10000000)

0.000% <= 1.703 milliseconds (cumulative count 2)
0.000% <= 1.807 milliseconds (cumulative count 5)
0.001% <= 1.903 milliseconds (cumulative count 55)
0.003% <= 2.007 milliseconds (cumulative count 315)
0.014% <= 2.103 milliseconds (cumulative count 1439)
6.588% <= 3.103 milliseconds (cumulative count 658806)
25.425% <= 4.103 milliseconds (cumulative count 2542508)
64.305% <= 5.103 milliseconds (cumulative count 6430514)
85.388% <= 6.103 milliseconds (cumulative count 8538781)
95.970% <= 7.103 milliseconds (cumulative count 9597041)
98.731% <= 8.103 milliseconds (cumulative count 9873097)
99.690% <= 9.103 milliseconds (cumulative count 9968958)
99.903% <= 10.103 milliseconds (cumulative count 9990304)
99.952% <= 11.103 milliseconds (cumulative count 9995201)
99.971% <= 12.103 milliseconds (cumulative count 9997052)
99.978% <= 13.103 milliseconds (cumulative count 9997752)
99.988% <= 14.103 milliseconds (cumulative count 9998786)
99.992% <= 15.103 milliseconds (cumulative count 9999204)
99.993% <= 16.103 milliseconds (cumulative count 9999346)
99.994% <= 17.103 milliseconds (cumulative count 9999357)
99.994% <= 18.111 milliseconds (cumulative count 9999363)
99.994% <= 19.103 milliseconds (cumulative count 9999388)
99.994% <= 20.111 milliseconds (cumulative count 9999400)
99.994% <= 30.111 milliseconds (cumulative count 9999412)
99.995% <= 31.103 milliseconds (cumulative count 9999504)
99.996% <= 32.111 milliseconds (cumulative count 9999569)
99.996% <= 33.119 milliseconds (cumulative count 9999619)
99.997% <= 34.111 milliseconds (cumulative count 9999657)
99.997% <= 35.103 milliseconds (cumulative count 9999701)
99.998% <= 36.127 milliseconds (cumulative count 9999769)
99.998% <= 37.119 milliseconds (cumulative count 9999813)
99.999% <= 38.111 milliseconds (cumulative count 9999854)
99.999% <= 39.103 milliseconds (cumulative count 9999892)
99.999% <= 40.127 milliseconds (cumulative count 9999933)
99.999% <= 41.119 milliseconds (cumulative count 9999947)
100.000% <= 42.111 milliseconds (cumulative count 9999964)
100.000% <= 43.103 milliseconds (cumulative count 9999978)
100.000% <= 44.127 milliseconds (cumulative count 10000000)

Summary:
  throughput summary: 102846.80 requests per second
  latency summary (msec):
        4.853     1.664     4.783     6.951     8.287    43.647
```

arcivanov:
@madolson @zuiderkwast The tests have been written and the performance results have been provided. 
Please let me know if there is anything else you need to move this along.

zuiderkwast:
Hi @arcivanov! I saw the benchmarks but I didn't look closely at them until now. I can see GETPXT has 10% higher throughput and 35% lower latency compared to MULTI-GET-PEXPIRETIME-EXEC.

In the past, many commands were added and they have accumulated a large maintenance burden. We have become more restrictive to adding new commands.

If your use case (distributed cache synchronization) is common enough, then maybe it's worth improving the area. I guess I don't fully understand the setup though. You can do reads from readonly replicas and get synchronized data including expire timestamps. You're doing something different?

The difference between latency and throughput in you benchmark is interesting. I assume the server interleaves the four commands in the MULTI-EXEC with commands from other clients, so after parsing and executing MULTI for one client, the server handles some other client, then parses and queues the GET for the first client, then some other client, etc. I think we should be able to optimize a pipelined MULTI-EXEC for latency by parsing and executing the full transaction in one go. This can push the latency down to match the difference in throughput, let's say only around 10% slower than your GETPXT latency. MULTI-EXEC will always have some overhead compared to single commands, but if we can optimize it, I think we can improve not only this use case but also others. This could be an optimization worth exploring.

In the diff, I saw the addition of `__rand_lst__` in `valkey-benchmark`, which expands to the last random value substituted for `__rand_int__`. This is a very useful addition for testing transactions and commands that operate on the same key more than once! I would like to merge that on it's own, if you move it to a separate PR.

arcivanov:
> In the past, many commands were added and they have accumulated a large maintenance burden. We have become more restrictive to adding new commands.

I understand. I have also seen the command implementation having dug through them and given the observed complexity of the commands in question I don't agree that the burden is significant but it's in the eye of the beholder, I guess and given the total volume of the commands, you might be right.

If you don't like me adding a new command (GETPXT) would you be open to me adding optional arguments to the GET command similarly how SET command is augmented with optional arguments as follows:

`GET key [EX | PX | EXAT | PXAT] `

This would provide:

1. Symmetry with the existing SET command.
2. An ability to atomically retrieve a value and a timestamp from ValKey that is currently *completely missing* (as opposed to the SET, which is present).

The above two points benefit our cache approach, yes, but irrespective of our needs they go towards completeness of the command API.

Would you be amenable to the above approach? I can drop the MGETPXT and use GET pipelining instead to reduce the maintenance burden.

zuiderkwast:
> > In the past, many commands were added and they have accumulated a large maintenance burden. We have become more restrictive to adding new commands.
> 
> I understand. I have also seen the command implementation having dug through them and given the observed complexity of the commands in question I don't agree that the burden is significant but it's in the eye of the beholder, I guess and given the total volume of the commands, you might be right.

Yeah, each command implementation simple. It's more the total volume, including documentation, test coverage, client support, ...

> If you don't like me adding a new command (GETPXT) would you be open to me adding optional arguments to the GET command similarly how SET command is augmented with optional arguments as follows:
> 
> `GET key [EX | PX | EXAT | PXAT] `

Then I'd rather add GETPXT actually. :)

We don't want to add options that modify the return type of a command. That's bad for some clients in statically typed languages. There may be a `valkey.get` function that's typed to return a string.

> The above two points benefit our cache approach, yes, but irrespective of our needs they go towards completeness of the command API.

You have a point here. We have `SET [EX | PX | EXAT | PXAT]` to set value and expire, and we have GETEX to get the value and set expire. I understand you would guess there's a GET-with-expire too, and your use case is reasonable.

I'm not sure about the name GETPXT and if we should have all the EX/PX/EXAT/PXAT variants or some an option to control the unit. Let's start a name discussion after deciding about adding it or not.

zuiderkwast:
@valkey-io/core-team What's your opinion on adding the commands mentioned in the title?

arcivanov:
Bump!

arcivanov:
@valkey-io/core-team @zuiderkwast bump

hwware:
Test enviroment:

arcivanov:
> Test enviroment:
?

hwware:
> GETPXT Simulated

Sorry, can you repost your latest test result, because I saw  GETPXT Simulated test result twice,  I doubt it should include one test for 
MULTI
GET key:__rand_int__
PEXPIRETIME key:__rand_int__
EXEC

BTW: I also saw it looks like difference performance result between these tests, I just want to confirm with you,  Thanks
I would like to add tag to let maintainer vote after you confirm the test result.

arcivanov:
@hwware Ok, I'll need to rerun the tests. Thanks for your attention to this!

madolson:
Hey! We talked about it in our weekly, https://github.com/valkey-io/valkey/wiki/Weekly-meeting, and we didn't come to a clear consensus. The 20% isn't that compelling number to add a new command for, since we see very slow adoption of new commands. We also discussed doing some profiling to see if there was some other way we could optimize this workload without introducing a new command so everyone could benefit. Sorry for not giving a more concrete answer.

arcivanov:
@madolson Ok, thanks for the update!

arcivanov:
Bump for update?

zuiderkwast:
Updates:

* We have been focusing on optimizing pipelines, especially when multiple different keys are accessed, so maybe not that much for the scenarios like `MULTI; GET foo; PEXPIRETIME foo; EXEC`, but any command/transaction/pipeline that accesses more than one key is likely to be faster in 9.0.0. It was merged after 9.0.0-rc1 though, but it will be in rc2.
  - #2092 
* We have improved the benchmark tool to be able to benchmark the scenarios discussed here in
  - #2057
  - #2102

I believe it should be possible to run valkey-benchmark with commands such as

    valkey-benchmark --threads 2 -r 1000000 -n 10000000 -- \
        multi ';' \
        mget key:__rand_1st__ key:__rand_2nd__ key:__rand_3rd__ ';' \
        pexpiretime key:__rand_1st__ ';' \
        pexpiretime key:__rand_2nd__ ';' \
        pexpiretime key:__rand_3rd__ ';' \
        exec

arcivanov:
Ok, thanks! I'll rebase and rerun the tests.

arcivanov:
## Re-benchmarked on 9.2-dev, with SAR instrumentation

The original numbers in this PR are from 2024. Since the rebase onto current `unstable` lands on a substantially rewritten `valkey-benchmark` (#2057, #2102), I re-ran the comparison from scratch on a quiet 2-socket box with full system instrumentation. Short version: the unpipelined result is in line with the original measurement — roughly 8-12%, so nothing regressed in the rebase — but that regime is round-trip-bound and badly understates the benefit. **Once you pipeline, GETPXT delivers 2.7x the transactions at ~36% of the server CPU per transaction.**

### Test environment

| | |
|---|---|
| **Host** | 2-socket Intel Xeon E5-2699A v4 @ 2.40 GHz (Broadwell-EP), 22 cores / 44 threads per socket, 88 logical CPUs total |
| **Topology** | 2 NUMA nodes — node0 = CPUs 0-21,44-65; node1 = CPUs 22-43,66-87. Node distance 10 / 21 |
| **Cache / memory** | 110 MiB L3 (2 instances), 251 GiB RAM (≈128 GiB per node) |
| **OS / kernel** | Fedora Linux 44, kernel 7.1.10-200.fc44.x86_64 |
| **Compiler** | gcc 16.2.1 (Red Hat 16.2.1-2), stock build, **no** `-march=native` |
| **CPU governor** | `performance`, pinned for the duration of the run |
| **Instrumentation** | sysstat 12.7.9 (`sar`/`sadf`) |
| **Valkey** | `156272db7` on top of `7be7d564f` (unstable, 9.2-dev), `v=255.255.255`, jemalloc 5.3.0, `io-threads=1` |

**Placement.** The server was pinned to NUMA node 0 and every benchmark client to NUMA node 1 with `numactl --cpunodebind/--membind`, so the two never contend for cores or memory bandwidth, and all traffic crossed the socket boundary over TCP on loopback. `io-threads=1` means the server is effectively single-core-bound, which is why its CPU tops out near 100% of one core at high pipeline depth; both paths ran under identical configuration, so this affects the absolute ceiling but not the comparison.

**Server configuration.** `--save '' --appendonly no --protected-mode no`, 3,000,000 keys pre-populated per payload size with `SET ... PXAT`, so every key carries a real expiration and `GETPXT`/`PEXPIRETIME` return a live timestamp rather than the `-1` fast path.

**Isolation.** The box is shared, so isolation was enforced rather than assumed: the harness refuses to start if the port is already bound, refuses to measure if any other benchmark process is alive, and every unrelated workload on the machine was stopped for the duration. The SAR recording lets this be audited after the fact — across all 35 cells, system-wide user CPU outside the two pinned dies never exceeded 3.81%, `%iowait` was **0.00** in every single cell, and memory use stayed between 1.5% and 7.4%.

**Run structure.** 35 cells, 20 s measured per cell after a 5 s warmup that is discarded, with a 4 s idle gap between cells so the SAR windows separate cleanly. Zero cells failed. One continuous `sar -u -r -b -B -w -n DEV 1` recording spans the whole campaign, alongside a 1 Hz sampler of the server's `utime`/`stime` from procfs; each cell's epoch window is written to a ledger and the recording is sliced per cell afterwards with `sadf -dU`, using `[t_start + warmup + 1, t_end - 1]` to exclude ramp edges. Point-sampling CPU around each cell was tried first and abandoned — it silently produces impossible numbers when a run finishes early.

**Two accounting details that matter for reading the numbers.** `valkey-benchmark` increments `requests_finished` once per *reply*, and the simulated path issues a four-command sequence, so its raw rps counts four per transaction; every MULTI figure in this comment has been divided by 4. Latency is *not* divided: `c->latency` is stamped once per read event and reset per write batch, so each sample already measures the full round trip to the arrival of the reply batch, symmetric with GETPXT's single reply. The SAR packet counts confirm the sequence really is one round trip — both paths use 2.00 loopback packets per transaction unpipelined.

**One caveat.** `cswch/s` came out too noisy to interpret — two cells with an identical configuration (`getpxt`, `c=200`, `d=3`, 16 threads, `P=1`) produced 76.7k and 33.3k at within-noise throughput — so no conclusions are drawn from it, and it is included in the table only for completeness.

---

### Results

#### Pipeline depth sweep — the headline result

`c=50`, `d=3`, 8 client threads. `throughput`, `latency` and `CPU/txn` columns are GETPXT relative to the MULTI transaction; positive `CPU/txn` means the MULTI path costs that much more server CPU per transaction.

| pipeline | GETPXT txn/s | MULTI txn/s | throughput | GETPXT avg ms | MULTI avg ms | latency | GETPXT p99 | MULTI p99 | GETPXT µs CPU/txn | MULTI µs CPU/txn | CPU/txn |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| P=1 | 79,621 | 71,230 | **+11.8%** | 0.621 | 0.693 | **-10.4%** | 1.231 | 1.383 | 6.85 | 8.33 | **+21.6%** |
| P=8 | 459,336 | 302,537 | **+51.8%** | 0.856 | 1.277 | **-33.0%** | 1.719 | 2.431 | 1.46 | 2.61 | **+79.4%** |
| P=32 | 987,422 | 455,993 | **+116.5%** | 1.508 | 3.114 | **-51.6%** | 2.543 | 4.095 | 0.84 | 1.91 | **+127.6%** |
| P=128 | 1,356,527 | 503,399 | **+169.5%** | 3.907 | 10.946 | **-64.3%** | 5.599 | 13.967 | 0.69 | 1.94 | **+180.7%** |

#### Concurrency sweep (unpipelined)

`d=3`, 16 client threads, `P=1`. The server never exceeded 60% of one core in any of these cells, so this regime is round-trip-bound rather than server-bound.

| clients | GETPXT txn/s | MULTI txn/s | throughput | GETPXT avg ms | MULTI avg ms | latency | GETPXT p99 | MULTI p99 | GETPXT µs CPU/txn | MULTI µs CPU/txn | CPU/txn |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| c=50 | 80,085 | 71,959 | **+11.3%** | 0.616 | 0.685 | **-10.1%** | 1.207 | 1.367 | 6.83 | 8.24 | **+20.7%** |
| c=200 | 81,400 | 72,424 | **+12.4%** | 2.421 | 2.727 | **-11.2%** | 4.799 | 5.415 | 6.89 | 8.35 | **+21.2%** |
| c=600 | 81,244 | 73,503 | **+10.5%** | 7.375 | 8.151 | **-9.5%** | 8.799 | 9.711 | 6.27 | 7.96 | **+26.9%** |
| c=1000 | 81,259 | 72,252 | **+12.5%** | 12.294 | 13.826 | **-11.1%** | 13.903 | 18.767 | 6.61 | 8.09 | **+22.5%** |

#### Payload sweep (unpipelined)

`c=600`, 16 client threads, `P=1`.

| value size | GETPXT txn/s | MULTI txn/s | throughput | GETPXT avg ms | MULTI avg ms | latency | GETPXT p99 | MULTI p99 | GETPXT µs CPU/txn | MULTI µs CPU/txn | CPU/txn |
|---|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 3B | 81,493 | 73,298 | **+11.2%** | 7.352 | 8.172 | **-10.0%** | 8.655 | 11.095 | 6.63 | 8.00 | **+20.8%** |
| 64B | 80,856 | 72,984 | **+10.8%** | 7.410 | 8.208 | **-9.7%** | 8.599 | 9.863 | 6.69 | 8.06 | **+20.5%** |
| 512B | 77,593 | 70,573 | **+9.9%** | 7.724 | 8.490 | **-9.0%** | 9.303 | 15.919 | 7.15 | 8.43 | **+17.9%** |
| 4096B | 65,273 | 60,686 | **+7.6%** | 9.177 | 9.875 | **-7.1%** | 13.591 | 12.647 | 9.35 | 10.50 | **+12.3%** |

#### SAR-derived server behaviour

Server CPU split and loopback wire cost, from the continuous `sar` recording sliced to each cell's window. Note that both paths use exactly the same number of packets per transaction — the MULTI sequence is genuinely a single round trip, which is what makes the comparison fair.

| cell | user % | system % | bytes/txn | packets/txn | cswch/s |
|---|---:|---:|---:|---:|---:|
| GETPXT `P=1` | 8.1 | 46.5 | 173 | 1.998 | 136,725 |
| MULTI `P=1` | 13.7 | 45.6 | 267 | 2.000 | 121,296 |
| GETPXT `P=8` | 25.7 | 41.2 | 82 | 0.250 | 91,156 |
| MULTI `P=8` | 49.3 | 29.8 | 176 | 0.250 | 47,334 |
| GETPXT `P=32` | 56.3 | 26.6 | 72 | 0.063 | 164,672 |
| MULTI `P=32` | 72.0 | 15.2 | 166 | 0.063 | 64,187 |
| GETPXT `P=128` | 80.5 | 13.2 | 70 | 0.016 | 389,000 |
| MULTI `P=128` | 91.4 | 6.2 | 164 | 0.016 | 138,966 |

#### Repeatability

Three back-to-back repeats of the same cell (`c=600`, `d=3`, 16 threads, `P=1`), to establish the noise floor.

| path | run 1 | run 2 | run 3 | spread |
|---|---:|---:|---:|---:|
| GETPXT | 80,796 | 80,688 | 81,192 | 0.62% |
| MULTI | 73,042 | 73,524 | 73,123 | 0.66% |

#### Client-thread sizing

Run first, to confirm the load generator was not itself the bottleneck. Throughput plateaus at 4 client threads, so the 8 and 16 thread configurations used above have headroom.

| client threads | GETPXT txn/s | server CPU % of one core |
|---|---:|---:|
| 1 | 60,199 | 61.2 |
| 4 | 80,250 | 54.3 |
| 8 | 81,168 | 54.1 |
| 16 | 84,210 | 55.6 |
| 24 | 83,665 | 56.8 |

### Charts

01-pipeline-throughput-cpu

02-sar-cpu-split

03-latency

04-sar-wire

05-unpipelined-sweeps

06-sar-timeline-phaseE

### Reading of the results

Unpipelined, GETPXT is worth 8-12% in throughput (7.6% in the weakest cell, the 4 KiB payload) and 7-11% in latency, holding across concurrency from 50 to 1000 clients and payloads from 3 B to 4 KiB. That is consistent with the original 2024 measurements, which is the main thing I wanted to confirm about the rebase. Repeatability is tight enough to trust it: three back-to-back repeats spread 0.62% for GETPXT and 0.66% for the MULTI path, an order of magnitude below the effect.

But the server sat at 51-64% of one core throughout those cells. The bottleneck there is the round trip, not the server, so throughput is largely pinned by client scheduling and the measured gap is a floor rather than the real saving. The honest metric in that regime is CPU per transaction, which already shows the MULTI path costing 12-27% more.

Pipelining removes the round trip from the equation and the real cost difference appears. At `P=128` GETPXT does 1.36 M transactions/s against 503 K, with p99 latency of 5.6 ms against 14.0 ms, at 0.69 µs of server CPU per transaction against 1.94 µs. The SAR user/system decomposition shows exactly where that goes: at `P=128` the MULTI path runs 91.4% user / 6.2% system while GETPXT runs 80.5% / 13.2%. The syscall volume is identical — the extra cost is four commands to parse, dispatch and queue, plus the transaction bookkeeping, all of it burned in userspace. The wire agrees, at 164 bytes per transaction against 70, a 2.3x protocol overhead at pipeline depth (1.5x unpipelined, where the fixed per-packet framing dominates).

That is the argument for the command, and it is a much stronger one than the 8% this PR currently claims.

### Attachments

`getpxt-bench-data.zip` has the raw driver output, the per-cell ledger with the epoch windows, and the SAR-joined per-cell table, with a README covering the units and the two accounting rules above. `getpxt-bench-scripts.zip` has the whole harness plus the post-processing and chart/table generators, so the run is reproducible end to end. The 435 MB `sar.dat` itself is not attached; re-run the campaign rather than trying to re-slice it.

[getpxt-bench-scripts.zip](https://github.com/user-attachments/files/31672945/getpxt-bench-scripts.zip)
[getpxt-bench-data.zip](https://github.com/user-attachments/files/31672946/getpxt-bench-data.zip)

### Unrelated bug found on the way

The human-readable report path of `valkey-benchmark` hangs in HDR percentile iteration at high throughput; the whole campaign runs with `--csv` to avoid it. Filed as #4583 with a fix in #4585 — independent of this PR.

arcivanov:
@zuiderkwast I have moved to latest unstable and reran and expanded the benches (same machine). I think the wins are pretty good and the argument for new commands is pretty strong.

arcivanov:
@madolson @hwware

PR #1461, opened 2024-12-19
Workflow changes to fix old release binaries.

roshkhatri:
- Moves `build-config.json` to workflow dir to build old versions with new configs.
- Enables contributors to test release Wf on private repo by adding `github.event_name == 'workflow_dispatch' ||`


.github/workflows/build-release-packages.yml
```diff
@@ -23,7 +23,7 @@ jobs:
   # This job provides the version metadata from the tag for the other jobs to use.
   release-build-get-meta:
     name: Get metadata to build
-    if: github.repository == 'valkey-io/valkey'
+    if: github.event_name == 'workflow_dispatch' || github.repository == 'valkey-io/valkey'
```
zuiderkwast, line 26: Is this to be able to test the workflow in a fork?

zuiderkwast (approved):
LGTM, thanks!

.github/workflows/build-release-packages.yml
roshkhatri: Yes, contributors can use `workflow_dispatch` in their fork to trigger the workflow

roshkhatri: Screenshot 2024-12-19 at 12 24 53 PM

zuiderkwast: Thats what you wrote in the top comment, yey...

zuiderkwast:
@enjoy-binbin @madolson FYI: #1429 only fixed the builds for future versions. This PR makes it possible to re-deploy binary releases using the workflow in the unstable branch, so we can fix the binary releases for the old versions.

PR #1463, opened 2024-12-20
Resolve bounds checks on cluster_legacy.c

madolson:
We are getting a number of errors like:
```
array subscript ‘clusterMsg[0]’ is partly outside array bounds of ‘unsigned char[2272]’
```

Which is basically GCC telling us that we have an object which is longer than the underlying storage of the allocation. We actually do this a lot, but GCC is generally not aware of how big the underlying allocation is, so it doesn't throw this error. We are specifically getting this error because the msgBlock can be of variable length depending on the type of message, but GCC assumes it's the longest one possible. The solution I went with here was make the message type optional, so that it wasn't included in the size. I think this also makes some sense, since it's really just a helper for us to easily cast the object around.

I considered disabling this error, but it is generally pretty useful since it can catch real issues. Another solution would be to over-allocate to the largest possible object, which could hurt performance as we initialize it to zero.

Results: https://github.com/madolson/valkey/actions/runs/12423414811/job/34686899884

This is a slightly cleaned up version of https://github.com/valkey-io/valkey/pull/1439. I thought I had another strategy but alas, it didn't work out.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -424,8 +424,18 @@ typedef struct {
     union {
         clusterMsg msg;
         clusterMsgLight msg_light;
-    };
+    } data[];
 } clusterMsgSendBlock;
+
+/* Helper function to extract a normal message from a send block. */
+static clusterMsgLight *getLightMessageFromSendBlock(clusterMsgSendBlock *msgblock) {
+    return &msgblock->data[0].msg_light;
+}
+
+/* Helper function to extract a light message from a send block. */
+static clusterMsg *getMessageFromSendBlock(clusterMsgSendBlock *msgblock) {
+    return &msgblock->data[0].msg;
+}
 
 /* -----------------------------------------------------------------------------
  * Initialization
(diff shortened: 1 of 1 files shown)


enjoy-binbin (approved):

PingXie:
> `array subscript ‘clusterMsg[0]’ is partly outside array bounds of ‘unsigned char[2272]’`

Is the repro step to build Valkey clean with "-Werror=array-bounds "? I don't see any compilation errors though.

madolson:
@PingXie We get them on some newer versions of GCC, like https://github.com/valkey-io/valkey/actions/runs/12422937837/job/34685598163.

PingXie:
I see. I couldn't tell the gcc version used in our CI jobs. I am not getting a repro locally with [gcc 14.2](https://gcc.gnu.org/gcc-14/).

madolson:
> I see. I couldn't tell the gcc version used in our CI jobs. I am not getting a repro locally with [gcc 14.2](https://gcc.gnu.org/gcc-14/).

I believe it's 14, but not sure which minor version. I was only able to reproduce it with the exact CI job version, ubuntu 24 LTS, with address sanitizer. GCC is apparently known for having bad bound checking, so I tried to come up with a solution that didn't hurt readability but removed the warnings.

PingXie:
Yeah this does sound like a GCC bug. There is enough information in the source code to conclude that our memory access is bounded. I suspect that the change might confuse this GCC version enough so it just gave up :)

PingXie (approved):

PR #1466, opened 2024-12-20
[Client Introspection] Client Commands Extended Filtering

sarthakaggarwal97:
In this PR, we introduce a support for few more filters for `CLIENT LIST` and `CLIENT KILL` commands. We introduce these new filters:

1. FLAGS `Client must include this flag. This can be a string with bunch of flags present one after the other.`
2. NAME `client name`
3. IDLE `minimum idle time of the client`
4. LIB-NAME `clients with the specified lib name.`
5. LIB-VER `clients with the specified lib version.`
6. DB `clients currently operating on the specified database ID`
7. IP `client ip address`
8. CAPA `client capabilities` 

Partly Addresses: #668


sarthakaggarwal97:
@zuiderkwast @hpatro would you please help review this. I have added the additional filters requested in #668.
Once the decision is being taken over the `IP` filter, I can add that as well. Thank you!

sarthakaggarwal97:
@valkey-io/core-team requesting you to take a look at the PR. Related to #668! Thank you

src/networking.c
```diff
@@ -64,6 +64,30 @@ typedef struct {
     int type;
     /* Boolean flag to determine if the current client (`me`) should be filtered. 1 means "skip me", 0 means otherwise. */
     int skipme;
+    /* Client name to filter. If NULL, no name filtering is applied. */
+    char *name;
+    /* Minimum idle time (in seconds) of a client connection for filtering.
+     * Connections with idle time more than this value will match.
+     * A value of 0 means no idle time filtering. */
+    long long min_idle;
```
zuiderkwast: Mininum here is the same as maximum in the MAXAGE filter, but we use the opposite word.

MAXAGE is illogical, but it makes some sense if you want to keep only young clients and kill the rest of them. In the same way you probably want to keep only active (non-idle) clients.

To avoid the confusion about min and max, can we just call this filter IDLE?

src/networking.c
```diff
@@ -3585,6 +3614,73 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
...
+            filter->min_idle = tmp;
+            index += 2;
+        } else if (!strcasecmp(c->argv[index]->ptr, "flags") && moreargs) {
+            filter->flags = c->argv[index + 1]->ptr;
+            index += 2;
+        } else if (!strcasecmp(c->argv[index]->ptr, "name") && moreargs) {
+            filter->name = c->argv[index + 1]->ptr;
+            index += 2;
+        } else if (!strcasecmp(c->argv[index]->ptr, "subscribed-pattern") && moreargs) {
+            filter->subscribed_pattern = createObject(OBJ_STRING, sdsnew(c->argv[index + 1]->ptr));
```
zuiderkwast: Can't we just use the objects in argv without copying? No need to copy it and free it afterwards, or am I missing something?

```suggestion
            filter->subscribed_pattern = c->argv[index + 1];
```

src/networking.c
```diff
@@ -3602,11 +3698,130 @@ static int clientMatchesFilter(client *client, clientFilter client_filter) {
...
+                !c->flag.dirty_cas && !c->flag.close_after_reply &&
+                !c->flag.unblocked && !c->flag.close_asap &&
+                !c->flag.unix_socket && !c->flag.readonly &&
+                !c->flag.no_evict && !c->flag.no_touch &&
+                !c->flag.import_source) {
+                return 1; /* Matches 'N' */
+            }
+            break;
+        default:
+            /* Invalid flag, return false */
```
zuiderkwast, line 3952: It's worth considering whether we should reject the command with an error if invalid flags are used. Otherwise some client may try to use flags from a future version and they will be silently ignored on this version.

zuiderkwast (commented):
Looks good to me on a high level. The refactoring is already done in the previous PR so the implementation in this PR strait-forward.

Still, it's a large change to the user-facing interface that needs to be carefully reviewed and needs majority approval. We are near the cutoff for 8.1 and we have many features that seem more important to prioritize. We may need to postpone this feature to 9.0, or maybe we can accept it after RC1.

soloestoy:
does this PR support negative filtering?

sarthakaggarwal97:
@soloestoy it doesn't support negative filtering right now. Would you recommend doing it in this PR itself?

soloestoy:
>@soloestoy it doesn't support negative filtering right now. Would you recommend doing it in this PR itself?

that would be very appreciated :)

sarthakaggarwal97:
@soloestoy do we want negative filtering only for flags or every other filter? Also, how would you recommend the filter command should look like with negation?

src/networking.c
```diff
@@ -64,6 +64,30 @@ typedef struct {
...
+     * A value of 0 means no idle time filtering. */
+    long long min_idle;
+    /* Client flags for filtering. If NULL, no filtering is applied. */
+    char *flags;
+    /* Client subscribed pattern for filtering. If NULL, no filtering is applied. */
+    robj *subscribed_pattern;
+    /* Client subscribed channel for filtering. If NULL, no filtering is applied. */
+    robj *subscribed_channel;
+    /* Client subscribed shard channel for filtering. If NULL, no filtering is applied. */
+    robj *subscribed_shard_channel;
```
hpatro: Doing an exact match might not be user friendly I believe. Would have been nice if we could support pattern matching on different types.

tests/unit/introspection.tcl
```diff
@@ -121,6 +121,284 @@ start_server {tags {"introspection"}} {
...
+        # Cleanup
+        catch {$c1 close}
+    }
+
+    test {CLIENT KILL with SHARDCHANNEL filter} {
+        # Create a client and subscribe to a shard channel
+        set c1 [valkey_client]
+        $c1 client setname mytestclient
+        $c1 ssubscribe myshardchannel
+
```
hpatro: Most of these tests are duplicate. Would be nice if we can loop over the different types and assert the behavior of disconnection.

example: https://github.com/valkey-io/valkey/blob/unstable/tests/unit/client-eviction.tcl#L301

tests/unit/introspection.tcl
```diff
@@ -505,19 +783,19 @@ start_server {tags {"introspection"}} {
         set rd [valkey_deferring_client]
         $rd monitor
         $rd read ; # Discard the OK
-
```
hpatro: Remove unnecessary edits.

hpatro (commented):
Mostly LGTM.

src/networking.c
sarthakaggarwal97: @zuiderkwast these field needs to be robj with sds type in order to be compared with client fields using `dictFind`. It was not working with just assigning the values like `c->argv[index + 1];`

This was the check: https://github.com/valkey-io/valkey/blob/unstable/src/server.c#L496, which comes from https://github.com/valkey-io/valkey/blob/unstable/src/dict.c#L814

src/networking.c
sarthakaggarwal97: Currently, I rely on `dictFind()` method, which performs an exact match. I agree, pattern matching is much more intuitive. If the exact match is not much beneficial, I would rather take these subscribed filters in a separate PR all together to unblock this PR for now.

@zuiderkwast adding you as well to this thread since you have been closely following up on these client enhancements.

src/networking.c
sarthakaggarwal97: Addressed this. Now when we parse the filter, we also check whether the flag is valid or not.

src/networking.c
sarthakaggarwal97: thank you, changed it to idle.

tests/unit/introspection.tcl
sarthakaggarwal97: I have a test which checks for multiple filters together, will this help?

```
    test {CLIENT LIST with multiple filters} {
        r client setname mytestclient
        set client_info [r client info]
        regexp {id=([^ ]+) name=([^ ]+)} $client_info _ myid myname
        set cl [split [r client list id $myid name $myname] "\r\n"]
        regexp {id=([^ ]+) name=([^ ]+) .* cmd=([^ ]+)} [lindex $cl 0] _ actual_id actual_name actual_cmd
        assert_equal $myid $actual_id
        assert_equal $myname $actual_name
        assert_equal "client|list" $actual_cmd
    }
```

Or you are looking for testing combinations of multiple filters as well?

src/networking.c
zuiderkwast: OK, interesting. Makes sense. It's possible to do an optimization though: If it's already an sds-robj, we could increment the reference counter instead of duplicating the object.

zuiderkwast:
> @soloestoy do we want negative filtering only for flags or every other filter? Also, how would you recommend the filter command should look like with negation?

I suggest we can add an optional `[NOT]` before every filter? e.g. `CLIENT LIST NOT TYPE REPLICA NOT ID 7 NOT LIB-NAME foo`.

Maybe it can be done in a separate PR? @soloestoy

src/networking.c
zuiderkwast: If we make them pattern-matches, i think the name should indicate that, like `SUBSCRIBED-CHANNEL-MATCH` or something. But what about PSUBSCRIBE patterns? Do we want to pattern-match on a pattern? This can be quite confusing... Users may need to escape pattern wildcard and such... `SUBSCRIBED-PATTERN-MATCH foo*\*` to match a PSUBSCRIBE pattern starting with `foo` and ending with `*`.

I think exact match is more simple and intuitive, but I'm also totally fine with a scope-cut excluding pubsub in this PR.

sarthakaggarwal97: >  Do we want to pattern-match on a pattern? This can be quite confusing

Thanks @zuiderkwast for your opinion. This is a valid point as well. If exact match is still useful than nothing, I am happy to keep it in this PR. @hpatro thoughts?

hpatro: Yeah, I was of the opinion we can tackle it in a separate PR with better understanding of the API interface which will be beneficial for the user.

sarthakaggarwal97: sounds good! I have removed pubsub filters to unblock this PR for now. We can follow it up with better API interface later.

src/networking.c
sarthakaggarwal97: thanks, made this change as well!

sarthakaggarwal97:
@hpatro @zuiderkwast please review the PR and let me know if you have any major comments / concerns on this PR.

src/networking.c
```diff
@@ -3677,6 +3698,70 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
...
+                addReplyError(c, "idle should be greater than 0");
+                return C_ERR;
+            }
+
+            filter->idle = tmp;
+            index += 2;
+        } else if (!strcasecmp(c->argv[index]->ptr, "flags") && moreargs) {
+            filter->flags = c->argv[index + 1]->ptr;
+            if (validateClientFlagFilter(filter->flags) == C_ERR) {
+                addReplyError(c, "unknown flags found in the filter");
```
hpatro: Should we add the `filter->flags` to the error response for easier logging on the application side ?

```suggestion
                addReplyErrorFormat(c, "unknown flags found in the provided filter: %s", filter->flags);
```

src/networking.c
```diff
@@ -3685,6 +3770,39 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
     return C_OK;
 }
 
+static int validateClientFlagFilter(const char *flag_filter) {
+    for (int i = 0; flag_filter[i] != '\0'; i++) {
```
hpatro: Could we pass it around as a sds ?

```
static int validateClientFlagFilter(sds flag_filter) {
    for (int i = 0; i < sdslen(flag_filter); i++) {
```

tests/unit/introspection.tcl
```diff
@@ -121,6 +121,196 @@ start_server {tags {"introspection"}} {
...
+        set c1 [valkey_client]
+        set c2 [valkey_client]
+        set c3 [valkey_client]
+        $c1 client setname client1
+        $c2 client setname client1
+        $c3 client setname client2
+        $c3 multi
+
+        # Wait 1 second to ensure idle time
+        after 1000  ;# Wait 1 second
```
hpatro: Above comment seems sufficient 

```suggestion
        after 1000
```

src/networking.c
```diff
@@ -3694,11 +3812,106 @@ static int clientMatchesFilter(client *client, clientFilter *client_filter) {
...
+    if (client->net_input_bytes < client_filter->tot_net_in) return 0;
+    if (client->net_output_bytes < client_filter->tot_net_out) return 0;
 
     /* If all conditions are satisfied, the client matches the filter. */
     return 1;
 }
 
+static int clientMatchesFlagFilter(client *c, const char *flag_filter) {
+    /* Iterate through the provided flag filter string */
+    for (int i = 0; flag_filter[i] != '\0'; i++) {
```
hpatro: Same question as `validateClientFlagFilter`.

src/networking.c
```diff
@@ -3694,11 +3812,106 @@ static int clientMatchesFilter(client *client, clientFilter *client_filter) {
...
+        case 'x': /* Multi flag */
+            if (!c->flag.multi) return 0;
+            break;
+        case 'b': /* Blocked flag */
+            if (!c->flag.blocked) return 0;
+            break;
+        case 't': /* Tracking flag */
+            if (!c->flag.tracking) return 0;
+            break;
+        case 'R': /* Invalid Client flag */
```
hpatro: Seems confusing.

src/commands/client-list.json
```diff
@@ -144,6 +148,62 @@
...
+                "type": "integer",
+                "optional": true,
+                "since": "8.1.0"
+            },
+            {
+                "token": "TOTAL-NET-OUT",
+                "name": "total-net-out",
+                "type": "integer",
+                "optional": true,
+                "since": "8.1.0"
```
hpatro: All of these needs to be targeting 9.0.

hpatro (commented):
We will merge this in 9.0. So, need to update the target version for the updated commands accordingly. Need to look through the tests next.

src/commands/client-list.json
sarthakaggarwal97: done

src/networking.c
sarthakaggarwal97: done

src/networking.c
sarthakaggarwal97: done

tests/unit/introspection.tcl
sarthakaggarwal97: done

src/networking.c
sarthakaggarwal97: done

src/networking.c
sarthakaggarwal97: done

src/commands.def
```diff
@@ -1353,7 +1362,8 @@ commandHistory CLIENT_LIST_History[] = {
 {"7.0.0","Added `resp`, `multi-mem`, `rbs` and `rbp` fields."},
 {"7.0.3","Added `ssub` field."},
 {"8.0.0","Replaced `master` `TYPE` with `primary`. `master` still supported for backward compatibility."},
-{"8.1.0","Added filters USER, ADDR, LADDR, SKIPME, and MAXAGE"},
+{"8.1.0","Added filters USER, ADDR, LADDR, SKIPME, and MAXAGE."},
+{"9.0.0","Added filters NAME, MINIDLE, FLAGS, LIB-NAME, LIB-VER, DB, TOTAL-NET-IN and TOTAL-NET-OUT"},
```
hpatro: nit: irks me we don't have a period for this sentence.

src/networking.c
```diff
@@ -3911,7 +4169,20 @@ void clientKillCommand(client *c) {
      * only after we queued the reply to its output buffers. */
     if (close_this_client) c->flag.close_after_reply = 1;
 client_kill_done:
-    zfree(client_filter.ids);
+    freeClientFilter(&client_filter);
+}
+
+static void freeClientFilter(clientFilter *filter) {
+    zfree(filter->ids);
+    sdsfree(filter->flags);
```
hpatro: To be on the safer side, could we perform NULL check for these as well to avoid double free?

tests/unit/introspection.tcl
```diff
@@ -1035,6 +1225,113 @@ start_server {tags {"introspection"}} {
...
+                    set c2 [valkey_client]
+
+                    $c1 select 2
+                    $c2 client kill db 2
+
+                    set result [$c2 client list]
+                    assert {[string match {*db=2*} $result] == 0}
+
+                    catch {$c2 close}
+                }
```
hpatro: why do we start a separate server within this test? And the indentation looks a bit weird. 

Generally we do the inverse, start the server and write tests within it.

hpatro (approved):
Overall LGTM. Some nit picks.

hpatro:
@valkey-io/core-team This PR is ready to be signed off. Could you folks vote for it 👍 / 👎 ?

src/networking.c
```diff
@@ -3720,19 +3933,41 @@ void clientHelpCommand(client *c) {
...
+        "    * NAME <client-name>",
+        "      Kill connections with the specified name.",
+        "    * IDLE <idle>",
+        "      Return clients with idle time greater than or equal to <idle> seconds.",
+        "    * SUBSCRIBED-PATTERN <subscribed-pattern>",
+        "      Kill connections subscribed to a matching subscribed pattern.",
+        "    * SUBSCRIBED-CHANNEL <subscribed channel>",
+        "      Kill connections subscribed to a matching subscribed channel.",
+        "    * SUBSCRIBED-SHARD-CHANNEL <subscribed-shard-channel>",
+        "      Kill connections subscribed to a matching subscribe shard channel.",
```
zuiderkwast: We've skipped these so let's skip them in the HELP text too.

src/networking.c
```diff
@@ -3749,6 +3984,28 @@ void clientHelpCommand(client *c) {
...
+        "    * NAME <client-name>",
+        "      Return clients with the specified name.",
+        "    * IDLE <idle>",
+        "      Return clients with idle time greater than or equal to <idle> seconds.",
+        "    * SUBSCRIBED-PATTERN <subscribed-pattern>",
+        "      Return clients subscribed to a matching subscribed-pattern.",
+        "    * SUBSCRIBED-CHANNEL <subscribed-channel>",
+        "      Return clients subscribed to the specified subscribed-channel.",
+        "    * SUBSCRIBED-SHARD-CHANNEL <shard-subscribed-channel>",
+        "      Return clients subscribed to the specified subscribe shard channel.",
```
zuiderkwast: Skip

src/commands.def
sarthakaggarwal97: haha fair enough, missed it client-list json.

tests/unit/introspection.tcl
sarthakaggarwal97: Since multi db is not supported currently in cluster mode so external server tests fails on this test. In order to skip the test, currently we start a new server and use `{external:skip}` tag. Took inspiration from other tcl tests in the package.

sarthakaggarwal97:
@zuiderkwast @hpatro addressed your comments. Please take a look, thank you!

madolson:
My preference would be to remove `TOT-NET-IN` and `TOT-NET-OUT` because they have a bit less utility, and I would like to keep the API as streamlined as possible. 

@valkey-io/core-team Please approve the new fields that we are adding.

sarthakaggarwal97:
thanks @madolson for the feedback! I have removed the `TOT-NET-IN` and `TOT-NET-OUT` filters.

soloestoy:
> > @soloestoy do we want negative filtering only for flags or every other filter? Also, how would you recommend the filter command should look like with negation?
> 
> I suggest we can add an optional `[NOT]` before every filter? e.g. `CLIENT LIST NOT TYPE REPLICA NOT ID 7 NOT LIB-NAME foo`.
> 
> Maybe it can be done in a separate PR? @soloestoy

I'm ok that we can add `not` or `!` in a separate PR.

But I still strongly recommend adding filtering such as `IP` and `CAPA` etc.

zuiderkwast:
@soloestoy I agree, let's add CAPA (implement just like FLAGS, a string of letters) and IP (just like ADDR but only the client's IP without port).

src/networking.c
```diff
@@ -3680,6 +3697,48 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
                 return C_ERR;
             }
             index += 2;
+        } else if (!strcasecmp(c->argv[index]->ptr, "idle") && moreargs) {
+            long long tmp;
```
hwware: We should name a local variable more readable, I think here it means seconds or idle_time. Please update it to a better name.

src/networking.c
```diff
@@ -3680,6 +3697,48 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
...
+        } else if (!strcasecmp(c->argv[index]->ptr, "lib-name") && moreargs) {
+            filter->lib_name = c->argv[index + 1];
+            incrRefCount(filter->lib_name);
+            index += 2;
+        } else if (!strcasecmp(c->argv[index]->ptr, "lib-ver") && moreargs) {
+            filter->lib_ver = c->argv[index + 1];
+            incrRefCount(filter->lib_ver);
+            index += 2;
+        } else if (!strcasecmp(c->argv[index]->ptr, "db") && moreargs) {
+            int tmp;
```
hwware: I think here it means db_id or db_index. Please update it to a better name.

src/networking.c
```diff
@@ -3680,6 +3697,48 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
...
+                addReplyError(c, "idle should be greater than 0");
+                return C_ERR;
+            }
+
+            filter->idle = tmp;
+            index += 2;
+        } else if (!strcasecmp(c->argv[index]->ptr, "flags") && moreargs) {
+            filter->flags = sdsnew(c->argv[index + 1]->ptr);
+            if (validateClientFlagFilter(filter->flags) == C_ERR) {
+                addReplyErrorFormat(c, "unknown flags found in the provided filter: %s", filter->flags);
```
hwware:
```suggestion
                addReplyErrorFormat(c, "Unknown flag found in the provided filter: %s", filter->flags);
```

src/networking.c
```diff
@@ -3680,6 +3697,48 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
...
+            filter->lib_ver = c->argv[index + 1];
+            incrRefCount(filter->lib_ver);
+            index += 2;
+        } else if (!strcasecmp(c->argv[index]->ptr, "db") && moreargs) {
+            int tmp;
+            if (getIntFromObjectOrReply(c, c->argv[index + 1], &tmp,
+                                        "db is not an integer or out of range") != C_OK)
+                return C_ERR;
+            if (tmp < 0 || tmp >= server.dbnum) {
+                addReplyErrorFormat(c, "db number should be between 0 and %d", server.dbnum - 1);
```
hwware:
```suggestion
                addReplyErrorFormat(c, "DB number should be between 0 and %d", server.dbnum - 1);
```

src/networking.c
sarthakaggarwal97:
```
127.0.0.1:6379> CLIENT INFO
id=2875 addr=127.0.0.1:38610 laddr=127.0.0.1:6379 fd=10 name= age=0 idle=0 flags=N capa= db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=26 qbuf-free=20448 argv-mem=10 multi-mem=0 rbs=16384 rbp=16384 obl=0 oll=0 omem=0 tot-mem=37786 events=r cmd=client|info user=default redir=-1 resp=2 lib-name= lib-ver=

```

Client Info has `db` which I think is better here as well. Anything else could confuse the user.

src/networking.c
sarthakaggarwal97: done!

hwware (approved):

sarthakaggarwal97:
@valkey-io/core-team added the capa and ip filters, and addressed comments. Please take a look, thank you.

sarthakaggarwal97:
@zuiderkwast I want to confirm if we currently only support "redirect" capa for clients. Please let me know if we support more. thank you

src/networking.c
```diff
@@ -3697,11 +3820,144 @@ static int clientMatchesFilter(client *client, clientFilter *client_filter) {
...
+            sdsfree(client_ip);
+            return 0;
+        }
+        sdsfree(client_ip);
+    } else {
+        if (sdscmp(peerid, ip) != 0)
+            return 0;
+    }
+    return 1;
+}
```
zuiderkwast: To split by colon is not complete.

Look at `getClientPeerId()`. It calls `genClientAddrString()`. It has this comment:

```
/* An "Address String" is a colon separated ip:port pair.
 * For IPv4 it's in the form x.y.z.k:port, example: "127.0.0.1:1234".
 * For IPv6 addresses we use [] around the IP part, like in "[::1]:1234".
 * For Unix sockets we use path:0, like in "/tmp/redis:0".
 *
 * An Address String always fits inside a buffer of NET_ADDR_STR_LEN bytes,
 * including the null term.
 *
 * On failure the function still populates 'addr' with the "?:0" string in case
 * you want to relax error checking or need to display something anyway (see
 * anetFdToString implementation for more info). */
```

For addresses with IPv6, we should match the IP without the square brackets. For Unix socket connections and for failed connections, I think we shouldn't match it at all because it's not a client IP.

We could use `connAddrPeerName()` instead of `getClientPeerId()` but involves a syscall for each client which is probably not a good idea. `getClientPeerId()` is cached. So perhaps we should just strip the square brackets. If the port is 0 it either means it's a Unix socket or there is no connection, I think we can skip the client if port is 0.

The sdscmp in the else case is not needed because `peerid` should always contain a colon. We can just return 0 in this case.

We can also try to optimize it since it can take some time if there are 50000 connected clients. We can avoid `strrchr()` which needs to traverse the whole string to find the end. If we start with `strncmp()`, we stop as soon as a mismatching character in the IP is encountered. I guess this code can be quite efficient, although untested:

```suggestion
static int clientMatchesIpFilter(client *c, sds ip) {
    const char *peerid = getClientPeerId(c);
    if (peerid[0] == '[') peerid++; /* IPv6 wrapped in square brackets. */
    size_t len = sdslen(ip);
    if (strncmp(peerid, ip, len) != 0) return 0;
    peerid += len;
    if (peerid[0] == ']') peerid++; /* IPv6 wrapped in square bracket. */
    if (peerid[0] != ':') return 0; /* IP:port colon not matched. */
    peerid++;
    if (peerid[0] == '0') return 0; /* Port is zero. */
    return 1;
}
```

sarthakaggarwal97: thanks @zuiderkwast for the great call out. I have incorporated your changes, they look good.

src/commands/client-kill.json
```diff
@@ -162,6 +166,61 @@
...
+                                "optional": true,
+                                "since": "9.0.0"
+                            },
+                            {
+                                "token": "DB",
+                                "name": "db",
+                                "type": "integer",
+                                "optional": true,
+                                "since": "9.0.0"
+                            },                            {
```
zuiderkwast: Missing line break

```suggestion
                            },
                            {
```

src/commands/client-list.json
```diff
@@ -144,6 +148,61 @@
...
+                "optional": true,
+                "since": "9.0.0"
+            },
+            {
+                "token": "DB",
+                "name": "db",
+                "type": "integer",
+                "optional": true,
+                "since": "9.0.0"
+            },                            {
```
zuiderkwast:
```suggestion
            },
            {
```

tests/unit/introspection.tcl
```diff
@@ -121,6 +121,244 @@ start_server {tags {"introspection"}} {
...
+
+        set client_info [$c1 client info]
+        regexp {addr=([^:]+):} $client_info -> iponly
+
+        # Kill client by IP only
+        r client kill ip $iponly skipme yes
+
+        set err [catch {$c1 ping} error_message]
+        assert {$err == 1}
+        assert {[string match "*I/O error*" $error_message]}
```
zuiderkwast: Use `assert_error` here and all similar places.

```suggestion
        assert_error "*I/O error*" {$c1 ping}
```

tests/unit/introspection.tcl
```diff
@@ -121,6 +121,244 @@ start_server {tags {"introspection"}} {
...
+        assert_match *client-with-r* $output
+    } {}
+
+    test {CLIENT KILL with IP filter} {
+        set c1 [valkey_client]
+        $c1 client setname "killme-ip"
+        r client setname "client-normal"
+
+        set client_info [$c1 client info]
+        regexp {addr=([^:]+):} $client_info -> iponly
```
zuiderkwast, line 257: This is OK for IPv4.

Shall we add a test for IPv6 too, to make sure the filter works correctly when addr has square brackets around the IPv6, like `addr=[::1]:12345` and killing client `::1`? I'm not extremely worried about it so I will not require it. Only if you want.

For inspiration, there are some IPv6 tests in [unit/cluster/announce-client-ip.tcl](https://github.com/valkey-io/valkey/blob/unstable/tests/unit/cluster/announce-client-ip.tcl#L71). In short, start a separate server with tag `ipv6` and bind to `127.0.0.1` and `::1`. Then connecting a client to `::1` makes it connect over IPv6.

tests/unit/introspection.tcl
```diff
@@ -1035,6 +1273,66 @@ start_server {tags {"introspection"}} {
...
+        $c2 client kill lib-ver 1.2.3
+
+        set result [$c2 client list]
+        assert {[string match {*lib-ver=1.2.3*} $result] == 0}
+
+        catch {$c2 close}
+    }
+
+    test {CLIENT KILL can filter by DB} {
+        start_server {config "default.conf" args {--save --loglevel verbose}} {
```
zuiderkwast: Why start another server in this test case?

tests/unit/introspection.tcl
sarthakaggarwal97: thanks, added a couple of tests

tests/unit/introspection.tcl
sarthakaggarwal97: fixed this!

zuiderkwast (approved):
LGTM now :+1:

PingXie:
Thanks @sarthakaggarwal97!

I didn't review closely but the high level command syntax LGTM so please consider this as my vote for the PR.

I do wonder though if we should add regex support in future PRs - there are quite some strings here, `client-name`, `lib-name`, etc.

also, we ran into a use case recently where we would like to know which client(s) contributed most of the data to the server. I would think `tot-net-in` could be a good indicator though ideally if we could find a way to tie the key space usage to the client(s). another thing on my wishlist is the sorting support, based on one of the client fields.

zuiderkwast:
@PingXie

> I do wonder though if we should add regex support in future PRs - there are quite some strings here, client-name, lib-name, etc.

I don't think we use regexp anywhere in the code base. Glob-style pattern maybe?

> I would think `tot-net-in` could be a good indicator

`tot-net-in` and `tot-net-out` filters were part of this PR earlier but were removed after this comment: https://github.com/valkey-io/valkey/pull/1466#issuecomment-2783698299 

> another thing on my wishlist is the sorting support, based on one of the client fields

Let's continue the high-level discussion about these topics in the issue #668.

soloestoy (approved):

madolson (approved):

src/networking.c
```diff
@@ -3688,6 +3764,53 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
     return C_OK;
 }
 
+static int validateClientCapaFilter(sds capa) {
```
hpatro, line 3767: We need some mechanism to ensure if we add a new client capa filter, we need to update this along with it.

src/networking.c
```diff
@@ -3688,6 +3764,53 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
...
+            /* Valid capability, do nothing. */
+            break;
+        default:
+            return C_ERR;
+        }
+    }
+    return C_OK;
+}
+
+static int validateClientFlagFilter(sds flag_filter) {
```
hpatro, line 3781: Same as above.

tests/unit/introspection.tcl
```diff
@@ -1035,6 +1294,65 @@ start_server {tags {"introspection"}} {
...
+        r CLIENT SETINFO lib-ver 1.2.3
+        set result [r client list lib-ver 1.2.3]
+        assert_match {*lib-ver=1.2.3*} $result
+    }
+
+    test {CLIENT LIST can filter by DB number} {
+        r select 2
+        set result [r client list db 2]
+        assert_match {*db=2*} $result
+    } {} {external:skip}
```
hpatro, line 1314: I presume we skip these to avoid being run on cluster mode or is there some other reasoning. I've added the extra tests label to verify that.

src/networking.c
```diff
@@ -3723,19 +3979,35 @@ void clientHelpCommand(client *c) {
...
-        "      Kill connections by client ids.",
+        "      Kill connections by client IDs.",
         "    * MAXAGE <maxage>",
         "      Kill connections older than the specified age.",
+        "    * FLAGS <flags>",
+        "      Kill connections that include the specified flags.",
+        "    * NAME <client-name>",
+        "      Kill connections with the specified name.",
+        "    * IDLE <idle>",
+        "      Return clients with idle time greater than or equal to <idle> seconds.",
```
```suggestion
        "      Kill connections with idle time greater than or equal to <idle> seconds.",
```

src/networking.c
```diff
@@ -3914,7 +4203,26 @@ void clientKillCommand(client *c) {
...
+
+static void freeClientFilter(clientFilter *filter) {
+    if (filter->ids != NULL)
+        zfree(filter->ids);
+    if (filter->flags != NULL)
+        sdsfree(filter->flags);
+    if (filter->capa != NULL)
+        sdsfree(filter->capa);
+    if (filter->ip != NULL)
+        sdsfree(filter->ip);
```
```suggestion
    zfree(filter->ids);
    sdsfree(filter->flags);
    sdsfree(filter->capa);
    sdsfree(filter->ip);
```

All of these functions have built in null checks. We should probably also null them after freeing them, like you do with the lib_name and lib_ver. At the very least be consistent?

hpatro:
Added a couple of changes to cleanup idle time and fix db test.

src/networking.c
hpatro: I think it's safe to do a NULL check before free-ing. I agree we should `NULL` it out.

src/networking.c
sarthakaggarwal97: I think we can update the documentation which suggest that we currently support only replicate capa filter? If new capa is added, we can take it as a subsequent change?

src/networking.c
hpatro: Added them.

src/networking.c
sarthakaggarwal97: Can we make use of documentation to specify what client flag filter we support? Otherwise currently there is no common method to validate flags I think.

tests/unit/introspection.tcl
sarthakaggarwal97: yes, multiple DB is not supported in cluster mode.

hpatro: I see some test failures with the extra test run. @sarthakaggarwal97 Could you help resolve them? I think we are good to go after that.

hpatro:
Great work @sarthakaggarwal97! It's merged now. 🎉

hpatro:
@sarthakaggarwal97 Please submit the doc PR.

sarthakaggarwal97:
@hpatro Doc [PR](https://github.com/valkey-io/valkey-doc/pull/262) has been merged, thank you

PR #1467, opened 2024-12-20
[Client Introspection] Client Count Command

### LINKED ISSUE #668: Extended Valkey client introspection functionality

sarthakaggarwal97:
With this PR, we would include a new subcommand called `CLIENT COUNT`.

This subcommand will be responsible for counting the number of clients, while supporting the existing filters and the newly added filters in #1466.

Sample Output:

```
127.0.0.1:6379> client list
id=8 addr=127.0.0.1:49924 laddr=127.0.0.1:6379 fd=8 name= age=1413 idle=933 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=client|count user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=380 tot-net-out=217586 tot-cmds=9
id=9 addr=127.0.0.1:49927 laddr=127.0.0.1:6379 fd=9 name= age=1411 idle=0 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=10 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1946 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=79 tot-net-out=216501 tot-cmds=2
id=10 addr=127.0.0.1:49928 laddr=127.0.0.1:6379 fd=10 name= age=1409 idle=1409 flags=N db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1920 events=r cmd=command|docs user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=27 tot-net-out=215501 tot-cmds=1
127.0.0.1:6379> client count
(integer) 3
127.0.0.1:6379> client count id 8 9
(integer) 2
127.0.0.1:6379> client count id 8 9 addr 127.0.0.1:49927
(integer) 1
127.0.0.1:6379> client count id 8 9 flags N
(integer) 2
127.0.0.1:6379> 

```

Built on top of: #1401, #1466 
Resolves: #668


hpatro (commented):
I would like to understand if anyone will benefit out of this much. With the recently introduced filters, it's pretty convenient to reduce the `CLIENT LIST` output already. And without seeing the response, would anyone just filter and would like to see count?

I know it's not a lot of code addition to the server but want us to think through if we want to support this. Users could optionally just perform a `wc -l` on the `CLIENT LIST <filter>` output.

sarthakaggarwal97:
I think if there is an opportunity to provide count of the box, we should explore it. It could be better than depending on clients and users to have additional logic on their end. Also, we can save on additional compute when we convert clients into their readable info strings and save up network bandwidth. wdyt?

sarthakaggarwal97:
@valkey-io/valkey-committers requesting your feedback here! thank you

hwware:
From my point of view, I do not understand the requirement for this command. I think users could implement this function on their side.

hpatro:
We will revisit if there will be any request for it, for now closing it out.

PR #1472, opened 2024-12-22
Fix switch case compilation error in the new helloscripting

enjoy-binbin:
It is missing the curly braces.


zuiderkwast:
If there are local variables in a case, it's good to make it a scope. Is there a warning for that in some CI job?

zuiderkwast (approved):

zuiderkwast:
I see the error in the daily runs: https://github.com/valkey-io/valkey/actions/runs/12449533103/job/34755280012

zuiderkwast:
Introduced in #1277

PR #1473, opened 2024-12-22
Fix JSON description of SET command

gmbnomis:
In the `arguments` section, the `arguments` key is only used for arguments of type `block` or `oneof`.

Consequently, the `arguments` given for `IFEQ` are ignored by the server. However, they lead to strange results when rendering the command's page for the web documentation.

Fix this by removing `arguments` for `IFEQ`.


zuiderkwast (approved):
Thanks! This look right and matches the descriptions in src/commands/README.md.

PR #1475, opened 2024-12-23
Move auth check to the front, before command exist/arity/protected check

enjoy-binbin:
When requirepass is enabled, we want command calls to return NOAUTH
instead of ERR with the error message.

Previously, these checks were all before the auth check:
- command existence exist
- command arity check
- command protected check

This may expose information such as whether the server supports the
command, whether the configuration item is enabled, etc. This is more
of a consistency issue as the same error message is returned when
requirepass is enabled, not a security issue.

This is a behavior change, though perhaps not a breaking one.

We put the auth check into !client_reprocessing_command block, that
means the reprocessing command would skip auth check, it's also a behavior
change but align with the old behavior before the refactor of reprocessing,
so it should be fine.

Before:
```
127.0.0.1:6379> foo bar
(error) ERR unknown command 'foo', with args beginning with: 'bar' 
127.0.0.1:6379> set foo
(error) ERR wrong number of arguments for 'set' command
127.0.0.1:6379> module load foo
(error) ERR MODULE command not allowed...
```

After:
```
127.0.0.1:6379> foo bar
(error) NOAUTH Authentication required.
127.0.0.1:6379> set foo
(error) NOAUTH Authentication required.
127.0.0.1:6379> module load foo
(error) NOAUTH Authentication required.
```


src/server.c
```diff
@@ -4006,21 +4006,6 @@ int processCommand(client *c) {
             rejectCommandSds(c, err);
```
madolson: Wouldn't your argument also extend to these two checks? You can check if the server has passwords enabled by sending both an invalid command or a command that doesn't exist.

```
127.0.0.1:6379> set Foo
(error) ERR wrong number of arguments for 'set' command
127.0.0.1:6379> set FOO BAR
(error) NOAUTH Authentication required.
127.0.0.1:6379> bah
(error) ERR unknown command 'bah', with args beginning with: 
```

Seems like we should cover them all.

enjoy-binbin: ok, I have thought of this too, and think this check can be put in advance, but it makes sense, I will cover them together.

enjoy-binbin: btw, i am going to handle this as well.
```
            /* Handle possible security attacks. */
            if (!strcasecmp(c->argv[0]->ptr, "host:") || !strcasecmp(c->argv[0]->ptr, "post")) {
                securityWarningCommand(c);
                return C_ERR;
            }
```

enjoy-binbin: On the other hand, this can let people check whether the server supports xxx command

PingXie: I believe `AUTH` is primarily intended to protect user data. I understand the consistency concern regarding returning the same error message when AUTH is validated, but I don’t see a security issue with revealing a command's presence, arity, or protected status?

BTW, this is a behavior change, though perhaps not a breaking one.

enjoy-binbin: > I believe AUTH is primarily intended to protect user data

i think it should protect everything according to the existing code.

> I understand the consistency concern regarding returning the same error message when AUTH is validated, but I don’t see a security issue with revealing a command's presence, arity, or protected status?

yes, it is not a security issue, i just think it's better this way

> BTW, this is a behavior change, though perhaps not a breaking one.

yes, it is a behavior change but i guess it is safe.

PingXie (approved):

zuiderkwast (approved):

madolson:
> BTW, this is a behavior change, though perhaps not a breaking one.

I don't think it's a breaking change since it is moving from error -> error case. I would just list it as a release. 

I'm also OK with the change atm, but it looks like there are still some tests that need to get fixed.

madolson:
Maybe we should leave this until 9.0 though, just so that we have the safer boundary of the major version to claim we are making a behavior change. I agree that it makes sense to raise the security bar, but I don't think it's all that important if it's in 8.1 or 9.0.

enjoy-binbin:
ok, let move it to 9.0

soloestoy (approved):

soloestoy:
I see you put the auth check into `!client_reprocessing_command` field, that means the reprocessing command would skip auth check, it's also a behavior change but align with the old behavior before the refactor of reprocessing, I'm OK with that.

PR #945, opened 2024-08-26
Refactor return and goto statements

naglera:
Consolidate the cleanup of local variables to a single point within the method, ensuring proper resource management and p
reventing memory leaks or double-free issues.

Previoslly descused here:
- https://github.com/valkey-io/valkey/pull/60#discussion_r1667872633
- https://github.com/valkey-io/valkey/pull/60#discussion_r1668045666

PR #1476, opened 2024-12-23
Refactor syncWithPrimary and replicaTryPartialResynchronization

nitaicaro:
In continuation to https://github.com/valkey-io/valkey/pull/945

**syncWithPrimary:**
- Refactored all error handling to function `syncWithPrimaryHandleError`
- Refactored repl_state state machine from if-else format to switch-case format
- Besides changing the repl_state, all state machine logic moved to helper functions

**replicaTryPartialResynchronization:**
This function was performing two different jobs based on the value of the read_reply argument - 
- read_reply == 0: Sends the PSYNC command to the primary server.
- read_reply == 1: Reads and processes the reply to the PSYNC command.

This change simplifies the logic by clearly separating the writing and reading stages of the PSYNC process.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -63,6 +63,7 @@ void replicationSteadyStateInit(void);
 void dualChannelSetupMainConnForPsync(connection *conn);
 void dualChannelSyncHandleRdbLoadCompletion(void);
 static void dualChannelFullSyncWithPrimary(connection *conn);
+void syncWithPrimary(connection *conn);
 
 /* We take a global flag to remember if this instance generated an RDB
  * because of replication, so that we can remove the RDB file in case
@@ -3093,54 +3094,17 @@ void dualChannelSyncHandleRdbLoadCompletion(void) {
     return;
 }
 
-/* Try a partial resynchronization with the primary if we are about to reconnect.
- * If there is no cached primary structure, at least try to issue a
- * "PSYNC ? -1" command in order to trigger a full resync using the PSYNC
- * command in order to obtain the primary replid and the primary replication
- * global offset.
+/* Handles the initial step of the partial resynchronization process by
+ * preparing and sending a PSYNC command to the primary server.
+ * This function determines the appropriate replication ID (replid)
+ * and replication offset based on the server's state. If successful,
+ * the function signals readiness for the reply processing phase, which is
+ * processed in replicaProcessPsyncReply().
  *
- * This function is designed to be called from syncWithPrimary(), so the
- * following assumptions are made:
- *
- * 1) We pass the function an already connected socket "fd".
- * 2) This function does not close the file descriptor "fd". However in case
- *    of successful partial resynchronization, the function will reuse
- *    'fd' as file descriptor of the server.primary client structure.
- *
- * The function is split in two halves: if read_reply is 0, the function
- * writes the PSYNC command on the socket, and a new function call is
- * needed, with read_reply set to 1, in order to read the reply of the
- * command. This is useful in order to support non blocking operations, so
- * that we write, return into the event loop, and read when there are data.
- *
- * When read_reply is 0 the function returns PSYNC_WRITE_ERR if there
- * was a write error, or PSYNC_WAIT_REPLY to signal we need another call
- * with read_reply set to 1. However even when read_reply is set to 1
- * the function may return PSYNC_WAIT_REPLY again to signal there were
- * insufficient data to read to complete its work. We should re-enter
- * into the event loop and wait in such a case.
- *
- * The function returns:
- *
- * PSYNC_CONTINUE: If the PSYNC command succeeded and we can continue.
- * PSYNC_FULLRESYNC: If PSYNC is supported but a full resync is needed.
- *                   In this case the primary replid and global replication
- *                   offset is saved.
- * PSYNC_NOT_SUPPORTED: If the server does not understand PSYNC at all and
- *                      the caller should fall back to SYNC.
- * PSYNC_WRITE_ERROR: There was an error writing the command to the socket.
- * PSYNC_WAIT_REPLY: Call again the function with read_reply set to 1.
- * PSYNC_TRY_LATER: Primary is currently in a transient error condition.
- *
- * Notable side effects:
- *
- * 1) As a side effect of the function call the function removes the readable
- *    event handler from "fd", unless the return value is PSYNC_WAIT_REPLY.
- * 2) server.primary_initial_offset is set to the right value according
- *    to the primary reply. This will be used to populate the 'server.primary'
- *    structure replication offset.
- */
-
+ * Return Values:
+ * - PSYNC_WRITE_ERROR: There was an error writing the command to the socket.
+ * - PSYNC_WAIT_REPLY: PSYNC was successfully sent, awaiting a reply. The next
+ *   step is to call replicaProcessPsyncReply(). */
 #define PSYNC_WRITE_ERROR 0
 #define PSYNC_WAIT_REPLY 1
 #define PSYNC_CONTINUE 2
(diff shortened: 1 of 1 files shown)


src/replication.c
```diff
@@ -3051,54 +3052,55 @@ void dualChannelSyncHandleRdbLoadCompletion(void) {
 #define PSYNC_NOT_SUPPORTED 4
 #define PSYNC_TRY_LATER 5
 #define PSYNC_FULLRESYNC_DUAL_CHANNEL 6
-int replicaTryPartialResynchronization(connection *conn, int read_reply) {
+int replicaSendPsyncCommand(connection *conn) {
```
naglera, line 3115: Function documentation is not accurate after this change (should be splitted to two cases)

nitaicaro: Updated

src/replication.c
```diff
@@ -2993,109 +2994,97 @@ void dualChannelSyncHandleRdbLoadCompletion(void) {
     return;
 }
 
-/* Try a partial resynchronization with the primary if we are about to reconnect.
- * If there is no cached primary structure, at least try to issue a
- * "PSYNC ? -1" command in order to trigger a full resync using the PSYNC
- * command in order to obtain the primary replid and the primary replication
- * global offset.
+/**
+ * Handles the initial step of the partial resynchronization process by
```
PingXie: nit - I don't remember we start the comment block with `/**`?
```suggestion
/* Handles the initial step of the partial resynchronization process by
```

src/replication.c
```diff
@@ -2993,109 +2994,97 @@ void dualChannelSyncHandleRdbLoadCompletion(void) {
...
- * 1) As a side effect of the function call the function removes the readable
- *    event handler from "fd", unless the return value is PSYNC_WAIT_REPLY.
- * 2) server.primary_initial_offset is set to the right value according
- *    to the primary reply. This will be used to populate the 'server.primary'
- *    structure replication offset.
+ * Return Values:
+ * - PSYNC_WRITE_ERROR: There was an error writing the command to the socket.
+ * - PSYNC_WAIT_REPLY: PSYNC was successfully sent, awaiting a reply. The next
+ *   step is to call replicaProcessPsyncReply().
  */
```
PingXie: nit - I believe this is more common.
```suggestion
 *   step is to call replicaProcessPsyncReply(). */
```

src/replication.c
```diff
@@ -2993,109 +2994,97 @@ void dualChannelSyncHandleRdbLoadCompletion(void) {
...
+    if (reply != NULL) {
+        serverLog(LL_WARNING, "Unable to send PSYNC to primary: %s", reply);
+        sdsfree(reply);
+        connSetReadHandler(conn, NULL);
+        return PSYNC_WRITE_ERROR;
     }
+    return PSYNC_WAIT_REPLY;
+}
+
+
```
PingXie: I personally like using double empty lines between functions but the convention is single empty link I think

src/replication.c
```diff
@@ -3738,7 +3763,7 @@ void syncWithPrimary(connection *conn) {
     server.repl_transfer_read = 0;
     server.repl_transfer_last_fsync_off = 0;
     server.repl_transfer_lastio = server.unixtime;
-    return;
+    goto ok;
 
 no_response_error: /* Handle receiveSynchronousResponse() error when primary has no reply */
```
PingXie: These `goto` labels (`no_response_error`, `error`, `write_error`, and `ok`) have really gone too far now. Let's consolidate them into a single `syncWithPrimaryHandleError` function. I am hopeful that we can remove all the `goto`s in this function.

PingXie (commented):
This kind of refactoring is really valuable for the long-term health of Valkey.  It makes the codebase easier to understand, maintain, and improve, which ultimately helps us deliver better features faster. Thanks @nitaicaro and @naglera! 

My main focus in this review is on the high level code structure, which looks great to me. That said, the use of goto in `syncWithPrimary` stands out.  Let's aim to eliminate all goto statements from that function. I'll take another look and provide more detailed feedback once that's addressed.

src/replication.c
nitaicaro: Done

nitaicaro:
> This kind of refactoring is really valuable for the long-term health of Valkey. It makes the codebase easier to understand, maintain, and improve, which ultimately helps us deliver better features faster. Thanks @nitaicaro and @naglera!
> 
> My main focus in this review is on the high level code structure, which looks great to me. That said, the use of goto in `syncWithPrimary` stands out. Let's aim to eliminate all goto statements from that function. I'll take another look and provide more detailed feedback once that's addressed.

Done

PingXie:
@nitaicaro thanks for the reminder! I didn't forget about this one :) I am planning to continue my review next week.

PingXie:
btw, there seems to be some merge conflicts. do you have a plan to resolve them soon?

src/replication.c
```diff
@@ -3070,109 +3071,92 @@ void dualChannelSyncHandleRdbLoadCompletion(void) {
...
-        if (server.failover_state == FAILOVER_IN_PROGRESS) {
-            reply = sendCommand(conn, "PSYNC", psync_replid, psync_offset, "FAILOVER", NULL);
-        } else {
-            reply = sendCommand(conn, "PSYNC", psync_replid, psync_offset, NULL);
-        }
+    /* Issue the PSYNC command, if this is a primary with a failover in
+     * progress then send the failover argument to the replica to cause it
+     * to become a primary */
+    if (server.failover_state == FAILOVER_IN_PROGRESS) {
+        reply = sendCommand(conn, "PSYNC", psync_replid, psync_offset, "FAILOVER", NULL);
```
PingXie, line 3148: not related to this PR but this doesn't look right. `PSYNC FAILOVER` is used in the standalone failover case by the *primary* only while `replicaSendPsyncCommand` is invoked by a replica only. In other words, this path will never be hit. this looks like a day-1 issue to me though.

No action is needed for this PR.

src/replication.c
```diff
@@ -3412,6 +3396,229 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
...
+
+    /* We accept only two replies as valid, a positive +PONG reply
+     * (we just check for "+") or an authentication error.
+     * Note that older versions of Redis OSS replied with "operation not
+     * permitted" instead of using a proper error code, so we test
+     * both. */
+    if (*err[0] != '+' && strncmp(*err, "-NOAUTH", 7) != 0 && strncmp(*err, "-NOPERM", 7) != 0 &&
+        strncmp(*err, "-ERR operation not permitted", 28) != 0) {
+        serverLog(LL_WARNING, "Error reply to PING from primary: '%s'", *err);
+        sdsfree(*err);
```
PingXie: shouldn't we retain `*err` here so the caller can get a string? or I think we should null out `*err` if freeing the string is desired.

PingXie: btw, this seems like an existing issue.

src/replication.c
```diff
@@ -3412,6 +3396,229 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
...
+     * EOF: supports EOF-style RDB transfer for diskless replication.
+     * PSYNC2: supports PSYNC v2, so understands +CONTINUE <new repl ID>.
+     * skip-rdb-checksum: supports skipping RDB checksum during full sync.
+     *                    Inform the primary of this capa only during diskless sync
+     *                    using a connection that has integrity checks (such as TLS).
+     *                    In non-diskless sync, or non-integrity-checked connection, there is more
+     *                    concern for data corruprion so we keep this extra layer of detection.
+     *
+     * The primary will ignore capabilities it does not understand. */
+    int send_skip_rdb_checksum_capa = replicationSupportSkipRDBChecksum(conn, useDisklessLoad(), 1); // we can ignore primary's conditions when sending capa (is_primary_stream_verified=1)
```
PingXie: nit  - this line is a bit too long. maybe move the comment to its own line above the code?

src/replication.c
```diff
@@ -3412,6 +3396,229 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
...
+}
+
+int syncWithPrimaryHandleReceiveAuthReplyState(connection *conn, sds *err) {
+    if (server.primary_auth) {
+        *err = receiveSynchronousResponse(conn);
+        if (*err == NULL) return C_ERR;
+        if (*err[0] == '-') {
+            serverLog(LL_WARNING, "Unable to AUTH to PRIMARY: %s", *err);
+            sdsfree(*err);
+            return C_ERR;
```
PingXie, line 3562: similarly I think we should either not free `*err` (preferred) or null out `*err` after freeing the memory

src/replication.c
```diff
@@ -3412,6 +3396,229 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
...
+    /* Ignore the error if any, not all the Redis OSS versions support
+     * REPLCONF listening-port. */
+    if (*err[0] == '-') {
+        serverLog(LL_NOTICE,
+                  "(Non critical) Primary does not understand "
+                  "REPLCONF listening-port: %s",
+                  *err);
+    }
+    sdsfree(*err);
+    return C_OK;
```
PingXie, line 3581: null out `*err`

src/replication.c
```diff
@@ -3412,6 +3396,229 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
...
+        if (*err == NULL) return C_ERR;
+        /* Ignore the error if any, not all the Redis OSS versions support
+         * REPLCONF ip-address. */
+        if (*err[0] == '-') {
+            serverLog(LL_NOTICE,
+                      "(Non critical) Primary does not understand "
+                      "REPLCONF ip-address: %s",
+                      *err);
+        }
+        sdsfree(*err);
```
PingXie: null out `*err`

src/replication.c
```diff
@@ -3412,6 +3396,229 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
...
+                  "(Non critical) Primary does not understand "
+                  "REPLCONF capa: %s",
+                  *err);
+    }
+    sdsfree(*err);
+    *err = NULL;
+    return C_OK;
+}
+
+int syncWithPrimaryHandleReceiveVersionReplyState(connection *conn, sds *err) {
```
PingXie: these functions (`syncWithPrimaryHandleReceiveVersionReplyState`, `syncWithPrimaryHandleReceiveCapaReplyState`, `syncWithPrimaryHandleReceivePortReplyState `, `syncWithPrimaryHandleReceiveAuthReplyState `) look very similar. any chance to reduce some code? the code part is pretty much the same (copied below) with the exception being the log line.

```c
    *err = receiveSynchronousResponse(conn);
    if (*err == NULL) return C_ERR;
    /* Ignore the error if any, not all the Redis OSS versions support
     * REPLCONF capa. */
    if (*err[0] == '-') {
        serverLog(LL_NOTICE,
                  "(Non critical) Primary does not understand "
                  "REPLCONF capa: %s",
                  *err);
    }
    sdsfree(*err);
    *err = NULL;
```

src/replication.c
```diff
@@ -3412,6 +3396,229 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
...
+    }
+    sdsfree(*err);
+    *err = NULL;
+    return C_OK;
+}
+
+int syncWithPrimaryHandleSendPsyncState(connection *conn, sds *err) {
+    if (replicaSendPsyncCommand(conn) == PSYNC_WRITE_ERROR) {
+        *err = sdsnew("Write error sending the PSYNC command.");
+        abortFailover("Write error to failover target");
```
PingXie: how about just reuse the `*err` string? less typing and easier to read
```suggestion
        abortFailover(*err);
```

src/replication.c
```diff
@@ -3412,6 +3396,229 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
...
+        return C_ERR;
+    }
+    return C_OK;
+}
+
+void syncWithPrimaryHandleError(connection *conn) {
```
PingXie: I think it is safer to pass in the pointer of pointer to a connection object since this function is destroying the connection object - it is a good practice to null out the pointer where the object is destroyed. 
```suggestion
void syncWithPrimaryHandleError(connection **conn) {
```

src/replication.c
```diff
@@ -3412,6 +3396,229 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
...
+    connClose(conn);
```
PingXie:
```suggestion
    connClose(*conn);
    *conn = NULL;
```

src/replication.c
```diff
@@ -3412,6 +3396,229 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
...
+        connClose(server.repl_rdb_transfer_s);
+        server.repl_rdb_transfer_s = NULL;
+    }
+    replicationAbortSyncTransfer();
+    server.repl_state = REPL_STATE_CONNECT;
+}
+
+void syncWithPrimaryHandleWriteError(connection *conn, sds *err) {
+    serverLog(LL_WARNING, "Sending command to primary in replication handshake: %s", *err);
+    sdsfree(*err);
```
PingXie: null out `*err` as well

src/replication.c
```diff
@@ -3505,240 +3712,106 @@ void syncWithPrimary(connection *conn) {
...
-        if (err[0] != '+' && strncmp(err, "-NOAUTH", 7) != 0 && strncmp(err, "-NOPERM", 7) != 0 &&
-            strncmp(err, "-ERR operation not permitted", 28) != 0) {
-            serverLog(LL_WARNING, "Error reply to PING from primary: '%s'", err);
-            sdsfree(err);
-            goto error;
-        } else {
-            serverLog(LL_NOTICE, "Primary replied to PING, replication can continue...");
+    case REPL_STATE_RECEIVE_PING_REPLY:
+        if (syncWithPrimaryHandleReceivePingReplyState(conn, &err) == C_ERR) {
+            if (err == NULL)
```
PingXie: nit - I don't have a strong opinion but would it look better if we combined the two and inferred from the log message from the value of `err`?

PingXie (commented):
Thanks for your patience, @nitaicaro. The overall code structure LGTM and it is a lot cleaner!

I have some local comments that I think should be easy to address. We can aim at merging this PR shortly after 8.1 GA.

src/replication.c
nitaicaro: You are right, this seems to affect all paths where `err` is used to differentiate between no-reply error vs other errors. Will fix

src/replication.c
nitaicaro: Not sure I understand this one - did you mean combine the cases of `REPL_STATE_CONNECTING` and `REPL_STATE_RECEIVE_PING_REPLY`? Or something else?

src/replication.c
nitaicaro: I tried but there are small differences between these functions that make this tricky. For example, `syncWithPrimaryHandleReceiveAuthReplyState` returns C_ERR after logging, the rest do not, so this creates the following common handler:

```
int syncWithPrimaryHandleReceiveCommonReplyState(connection *conn, sds *err, const char *log_message, int is_err_critical) {
    *err = receiveSynchronousResponse(conn);
    if (*err == NULL) return C_ERR;

    /* Ignore the error if requested, and log a message */
    if (*err[0] == '-') {
        serverLog(LL_WARNING, "%s: %s", log_message, *err);
        if (is_err_critical) return C_ERR;
    }

    sdsfree(*err);
    *err = NULL;

    return C_OK;
}
```

Which is arguably less elegant than the current implementation.

PingXie: yeah it doesn't look very clean either. thanks for trying out the idea.

src/replication.c
```diff
@@ -3412,6 +3396,233 @@ void dualChannelSetupMainConnForPsync(connection *conn) {
...
+    server.repl_transfer_s = NULL;
+    if (server.repl_rdb_transfer_s) {
+        connClose(server.repl_rdb_transfer_s);
+        server.repl_rdb_transfer_s = NULL;
+    }
+    replicationAbortSyncTransfer();
+    server.repl_state = REPL_STATE_CONNECT;
+}
+
+void syncWithPrimaryHandleWriteError(connection *conn, sds *err) {
```
PingXie: the connection pointer needs to be null'd out as well since the object is destructed by `syncWithPrimaryHandleError` in this function.
```suggestion
void syncWithPrimaryHandleWriteError(connection **conn, sds *err) {
```

src/replication.c
PingXie: I was thinking about combining `syncWithPrimaryHandleNoResponseError` and `syncWithPrimaryHandleError`. not a blocker

PingXie (commented):
LGTM with only one last comment. Will approve once it is addressed. Thanks @nitaicaro

src/replication.c
nitaicaro: That makes sense - we call `syncWithPrimaryHandleNoResponseError` when `receiveSynchronousResponse` returns`NULL`, in which case it already logs `serverLog(LL_WARNING, "Failed to read response from the server: %s", connGetLastError(conn));`.

src/replication.c
nitaicaro: Done

src/replication.c
```diff
@@ -3505,240 +3715,100 @@ void syncWithPrimary(connection *conn) {
...
-        if (err[0] != '+' && strncmp(err, "-NOAUTH", 7) != 0 && strncmp(err, "-NOPERM", 7) != 0 &&
-            strncmp(err, "-ERR operation not permitted", 28) != 0) {
-            serverLog(LL_WARNING, "Error reply to PING from primary: '%s'", err);
-            sdsfree(err);
-            goto error;
-        } else {
-            serverLog(LL_NOTICE, "Primary replied to PING, replication can continue...");
+    case REPL_STATE_RECEIVE_PING_REPLY:
+        if (syncWithPrimaryHandleReceivePingReplyState(conn, &err) == C_ERR) {
+            syncWithPrimaryHandleError(&conn);
```
PingXie, line 3775: are we leaking `err`?

```suggestion
            syncWithPrimaryHandleWriteError(&conn, &err);
```

nitaicaro: No leak - err is now freed inside `syncWithPrimaryHandleReceivePingReplyState`

PingXie: >err is now freed inside syncWithPrimaryHandleReceivePingReplyState

This is an implementation detail. we should go by the function contract. I would say either we free `err` here or we should change `syncWithPrimaryHandleReceivePingReplyState` to not return `err`. either way would work for me.

nitaicaro: Good idea - thinking about this more, we can remove `syncWithPrimaryHandleWriteError` completely, since all it does on top of `syncWithPrimaryHandleError` is log a line, which we can move to inside the handlers. Then none of the `syncWithPrimaryHandler`s will have to take `err` as an argument - they will allocate and free it internally.  What do you think?

PingXie: Yeah removing `syncWithPrimaryHandleWriteError` and consolidating on `syncWithPrimaryHandleError` is good. That being said, I would still like us to fix `syncWithPrimaryHandleReceivePingReplyState` to either return an `err` string that needs to be freed by the caller or not return an `err` at all.

nitaicaro: Done

src/replication.c
```diff
@@ -3528,269 +3757,114 @@ void syncWithPrimary(connection *conn) {
...
+            return;
         }
-        sdsfree(err);
-        err = NULL;
         if (server.cluster_enabled) {
             server.repl_state = REPL_STATE_RECEIVE_NODEID_REPLY;
             return;
         } else {
             server.repl_state = REPL_STATE_SEND_PSYNC;
+            goto case_send_psync;
```
nitaicaro, line 3832: Not the cleanest solution, but could not think of another way to do this

PingXie: that's ok. let's merge now and address this later.

PingXie (approved):
This looks great! Thanks so much for your patience, @nitaicaro! Really appreciate it. Let's merge it!

PingXie:
@valkey-io/core-team anyone else would like to take a look at this pr before it gets merged?

enjoy-binbin (commented):
I am worry this will cause conflict with https://github.com/valkey-io/valkey/pull/1949. @murphyjacob4

murphyjacob4:
Sadly, a merge conflict seems inevitable. I wouldn't block this PR on #1949, I can handle the merge conflicts in that PR

PR #1479, opened 2024-12-23
skip CRC checksumming during diskless full sync with TLS enabled.

### LINKED ISSUE #1129: Skip CRC64 checksumming when doing diskless replication

talxsha:
Implemented a mechanism to eliminate CRC64 checksumming during full sync when not writing to disk (using a connection that has data integrity checks such as TLS), as it adds overhead with minimal benefit.

Nodes can skip CRC calculations when these conditions are met:
1. Running diskless sync on primary.
2. Running diskless load on the replica.
3. Primary-replica connection is integrity checked.

Closes #1129


ranshid:
@talxsha before I look into this, lets put some details in the top comment. linking the issue is not what we susually do.
Please state shortly what is the problem we are solving and what this solution includes.

src/replication.c
```diff
@@ -1244,11 +1244,12 @@ void syncCommand(client *c) {
  * the primary can accurately lists replicas and their listening ports in the
  * INFO output.
  *
- * - capa <eof|psync2|dual-channel>
+ * - capa <eof|psync2|dual-channel|disable_sync_crc>
```
ranshid: I am not super fond of the disable_sync_crc capability name. maybe a better name woulod be bypass_crc?

src/config.c
```diff
@@ -3156,6 +3156,7 @@ static int applyClientMaxMemoryUsage(const char **err) {
 standardConfig static_configs[] = {
     /* Bool configs */
     createBoolConfig("rdbchecksum", NULL, IMMUTABLE_CONFIG, server.rdb_checksum, 1, NULL, NULL),
+    createBoolConfig("disable-sync-crc", NULL, MODIFIABLE_CONFIG, server.disable_sync_crc, 0, NULL, NULL),
```
ranshid: we normally do not like to introduce new configurations. In this case the feature is controlled via capability so no issues with compatibility. Is there a way this would still be required in some cases>

src/server.c
```diff
@@ -2218,6 +2218,7 @@ void initServerConfig(void) {
     server.fsynced_reploff_pending = 0;
     server.rdb_client_id = -1;
     server.loading_process_events_interval_ms = LOADING_PROCESS_EVENTS_INTERVAL_DEFAULT;
+    server.repl_meet_disable_crc_cond = 0;
```
ranshid: I did not follow on why we need to keep this flag on the server? aren't all the checks in readSyncBulkPayload valid at the point of decision to flag the rdb?

src/server.h
```diff
@@ -1838,6 +1842,7 @@ struct valkeyServer {
     double stat_fork_rate;                         /* Fork rate in GB/sec. */
     long long stat_total_forks;                    /* Total count of fork. */
     long long stat_rejected_conn;                  /* Clients rejected because of maxclients */
+    size_t stat_total_crc_disabled_syncs_stated;   /* Total number of full syncs stated with CRC checksum disabled */   // AMZN
```
ranshid: Please remove the AMZN, we are not in Kansas anymore. Also the stat name is weird, maybe we can use another stat eg, sync_bypass_crc? .
Also note that In general I would not find any reason to have this statistic unless it is used for writing tests right? Maybe such stats are better be placed under the debug section of the info, but I guess we already have so many stats so i would let it pass.

src/rdb.c
```diff
@@ -3601,6 +3612,12 @@ int rdbSaveToReplicasSockets(int req, rdbSaveInfo *rsi) {
         }
         serverSetCpuAffinity(server.bgsave_cpulist);
 
+        if (disable_sync_crc_capa == 1) {
```
ranshid:
```suggestion
        if (disable_sync_crc_capa) {
```

src/rdb.c
```diff
@@ -3354,7 +3355,7 @@ int rdbLoadRioWithLoadingCtx(rio *rdb, int rdbflags, rdbSaveInfo *rsi, rdbLoadin
         if (rioRead(rdb, &cksum, 8) == 0) goto eoferr;
         if (server.rdb_checksum && !server.skip_checksum_validation) {
             memrev64ifbe(&cksum);
-            if (cksum == 0) {
+            if (cksum == 0 || (rdb->flags & RIO_FLAG_DISABLE_CRC) != 0) {
```
ranshid:
```suggestion
            if (cksum == 0 || (rdb->flags & RIO_FLAG_DISABLE_CRC)) {
```

src/config.c
madolson: Yes, please don't introduce a config for this.

madolson:
> Added the functionality to disable CRC calculations during diskless full sync with TLS enabled.

Also add justification for why we should do this only when TLS is enabled. Given that the network has built in checksumming, I'm still not convinced about the tradeoff we are making given that the steady state replication is not checksummed.

ranshid:
@madolson should I tag it as a major-decision ? I think it worth discussion.

src/replication.c
```diff
@@ -2084,6 +2086,12 @@ void readSyncBulkPayload(connection *conn) {
             serverLog(LL_NOTICE, "PRIMARY <-> REPLICA sync: receiving %lld bytes from primary %s",
                       (long long)server.repl_transfer_size, use_diskless_load ? "to parser" : "to disk");
         }
+
+        // Set a flag to determin later whether or not the replica will skip CRC calculations for this sync -
+        // Disable CRC on replica if: (1) TLS is enabled; (2) replica disable_sync_crc is enabled; (3) diskelss sync enabled on both replica and primary.
+        // Otherwise, CRC should be enabled/disabled as per server.rdb_checksum
+        if (connIsTLS(conn) && server.disable_sync_crc && use_diskless_load && usemark)
```
ranshid: a. I think we should encapsulate this whole condition in some function to make the code more readable.
b. The connIsTLS part of the decision is somewhat too intrusive IMO. I think maybe we can add an API in the connection abstraction like connIntegrityChecked or something like this. maybe there will be non-TLS connections (eg QUIC) which will provide some integrity mechanism which will not be defined as "TLS"

madolson:
> @madolson should I tag it as a major-decision ? I think it worth discussion.

For now it's not. It's just an internal one. I would probably just ping PingXie directly and core team if anyone else is interested.

src/connection.h
```diff
@@ -121,6 +121,9 @@ typedef struct ConnectionType {
 
     /* TLS specified methods */
     sds (*get_peer_cert)(struct connection *conn);
+
+    /* Miselenious */
+    int (*connIntegrityChecked)(void);  // return 1 iff connection type has built-in integrity checks
```
ranshid:
```suggestion
    int (*connIntegrityChecked)(void);  // return 1 if connection type has built-in integrity checks
```

src/server.h
```diff
@@ -1988,6 +1993,7 @@ struct valkeyServer {
     char *rdb_filename;                   /* Name of RDB file */
     int rdb_compression;                  /* Use compression in RDB? */
     int rdb_checksum;                     /* Use RDB checksum? */
+    int bypass_crc;                       /* Skip RDB checksum? Applicable only for TLS enabled diskless full sync */
```
ranshid: I am not sure why do we need to keep this server flag?

src/replication.c
```diff
@@ -1972,6 +1974,11 @@ static int useDisklessLoad(void) {
     return enabled;
 }
 
+/* Returns 1 if the replica can skip CRC calculations during full sync */
+int replicationBypassCRC(connection *conn, int is_replica_diskless_load, int is_primary_diskless_sync) {
```
ranshid: can we change it to be more explicit name? eg `replicationSupportBypassCRC`

src/replication.c
```diff
@@ -2251,6 +2259,7 @@ void readSyncBulkPayload(connection *conn) {
 
         serverLog(LL_NOTICE, "PRIMARY <-> REPLICA sync: Loading DB in memory");
         startLoading(server.repl_transfer_size, RDBFLAGS_REPLICATION, asyncLoading);
+        if (replicationBypassCRC(conn, use_diskless_load, usemark)) rdb.flags |= RIO_FLAG_BYPASS_CRC;
```
ranshid: lets document some explaining why we think it is O.K to take the decision based on the usemark and connection support.

src/rdb.c
```diff
@@ -3558,8 +3559,13 @@ int rdbSaveToReplicasSockets(int req, rdbSaveInfo *rsi) {
         safe_to_exit_pipe = pipefds[0];          /* read end */
         server.rdb_child_exit_pipe = pipefds[1]; /* write end */
     }
+    /*
+     * For replicas with repl_state == REPLICA_STATE_WAIT_BGSAVE_END and replica_req == req:
+     * Check replica capabilities, if every replica supports bypassing CRC, primary should also bypass CRC, otherwise, use CRC.
+     */
+    int bypass_crc_capa = 1;
```
ranshid:
```suggestion
    int bypass_crc = 1;
```

src/rdb.c
```diff
@@ -3591,6 +3597,11 @@ int rdbSaveToReplicasSockets(int req, rdbSaveInfo *rsi) {
             }
             replicationSetupReplicaForFullResync(replica, getPsyncInitialOffset());
         }
+
+        // do not bypass CRC on the primary if connection doesn't have integrity check or if the replica doesn't support it
+        if (!connIsIntegrityChecked(replica->conn) || !(replica->replica_capa & REPLICA_CAPA_BYPASS_CRC)) 
+            bypass_crc_capa = 0;
```
ranshid:
```suggestion
            bypass_crc = 0;
```

src/rdb.c
```diff
@@ -3614,6 +3625,12 @@ int rdbSaveToReplicasSockets(int req, rdbSaveInfo *rsi) {
         }
         serverSetCpuAffinity(server.bgsave_cpulist);
 
+        if (bypass_crc_capa) {
```
ranshid:
```suggestion
        if (bypass_crc) {
```

src/rdb.c
```diff
@@ -3679,6 +3696,7 @@ int rdbSaveToReplicasSockets(int req, rdbSaveInfo *rsi) {
             }
         }
         if (!dual_channel) close(safe_to_exit_pipe);
+        if (bypass_crc_capa) server.stat_total_sync_bypass_crc++;
```
ranshid:
```suggestion
        if (bypass_crc) server.stat_total_sync_bypass_crc++;
```

ranshid (approved):
@madolson the code seems fine now. would you like to take a quick look as well?

src/tls.c
```diff
@@ -1186,6 +1190,9 @@ static ConnectionType CT_TLS = {
 
     /* TLS specified methods */
     .get_peer_cert = connTLSGetPeerCert,
+
+    /* Miselenious */
```
madolson:
```suggestion
    /* Miscellaneous */
```

src/unix.c
```diff
@@ -207,6 +207,9 @@ static ConnectionType CT_Unix = {
     .process_pending_data = NULL,
     .postpone_update_state = NULL,
     .update_state = NULL,
+
+    /* Miselenious */
```
madolson:
```suggestion
    /* Miscellaneous */
```

src/server.c
```diff
@@ -5879,6 +5880,7 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "instantaneous_input_repl_kbps:%.2f\r\n", (float)getInstantaneousMetric(STATS_METRIC_NET_INPUT_REPLICATION) / 1024,
                 "instantaneous_output_repl_kbps:%.2f\r\n", (float)getInstantaneousMetric(STATS_METRIC_NET_OUTPUT_REPLICATION) / 1024,
                 "rejected_connections:%lld\r\n", server.stat_rejected_conn,
+                "total_sync_bypass_crc:%ld\r\n", server.stat_total_sync_bypass_crc,
```
```suggestion
                "total_sync_bypass_crc:%ld\r\n", server.stat_total_sync_bypass_crc,
```
What is the use of this metric? What are end users supposed to do with this information?

src/socket.c
```diff
@@ -447,6 +447,9 @@ static ConnectionType CT_Socket = {
     .process_pending_data = NULL,
     .postpone_update_state = NULL,
     .update_state = NULL,
+
+    /* Miselenious */
```
madolson:
```suggestion
    /* Miscellaneous */
```

src/replication.c
```diff
@@ -3511,11 +3527,19 @@ void syncWithPrimary(connection *conn) {
          *
          * EOF: supports EOF-style RDB transfer for diskless replication.
          * PSYNC2: supports PSYNC v2, so understands +CONTINUE <new repl ID>.
+         * BYPASS-CRC: supports skipping CRC calculations during full sync.
```
madolson: BYPASS isn't the right word, since it implies we are doing something else instead. I think `SKIP-CRC` is what you want.

src/rio.c
```diff
@@ -425,6 +425,7 @@ void rioFreeFd(rio *r) {
 /* This function can be installed both in memory and file streams when checksum
  * computation is needed. */
 void rioGenericUpdateChecksum(rio *r, const void *buf, size_t len) {
+    if ((r->flags & RIO_FLAG_BYPASS_CRC) != 0) return; // skip CRC64 calculations
```
madolson:
```suggestion
    if ((r->flags & RIO_FLAG_SKIP_CHECKSUM) != 0) return; // skip checksum
```

tests/integration/bypass-crc.tcl
```diff
@@ -0,0 +1,37 @@
+start_server {tags {"repl tls"} overrides {save {}}} {
+    set primary [srv 0 client]
+    set primary_host [srv 0 host]
+    set primary_port [srv 0 port]
+    set primary_bypassed_crc_counter 0
+    foreach mds {no yes} {
+        foreach sdl {disabled on-empty-db swapdb flush-before-load} {
```
madolson: I don't know what either mds or sdl stand for, can we use more normal names?

tests/integration/bypass-crc.tcl
```diff
@@ -0,0 +1,37 @@
...
+            test "Bypass CRC sync - tls:$::tls, repl_diskless_sync:$mds, repl_diskless_load:$sdl" {
+                $primary config set repl-diskless-sync $mds
+                start_server {overrides {save {}}} {
+                    set replica [srv 0 client]
+                    $replica config set repl-diskless-load $sdl
+                    $replica replicaof $primary_host $primary_port
+                    
+                    wait_for_condition 50 100 {
+                        [string match {*master_link_status:up*} [$replica info replication]]
```
madolson:
```suggestion
                        [s 0 master_link_status] eq {up}
```

tests/integration/bypass-crc.tcl
```diff
@@ -0,0 +1,37 @@
...
+                    } else {
+                        fail "Replication not started"
+                    }
+
+                    set replica_bypassing_crc_count [string match {*total_sync_bypass_crc:1*} [$replica info stats]]
+                    set stats [regexp -inline {total_sync_bypass_crc:(\d+)} [$primary info stats]]
+                    set primary_bypass_crc_count [lindex $stats 1]
```
madolson:
```suggestion
                    set replica_bypassing_crc_count [s 0 total_sync_bypass_crc]
                    set primary_bypass_crc_count [s 1 total_sync_bypass_crc]
```

tests/integration/bypass-crc.tcl
```diff
@@ -0,0 +1,37 @@
+start_server {tags {"repl tls"} overrides {save {}}} {
```
madolson: We only need to run these tests with TLS enabled, otherwise this condition is always false.

src/replication.c
```diff
@@ -2252,7 +2260,14 @@ void readSyncBulkPayload(connection *conn) {
 
         serverLog(LL_NOTICE, "PRIMARY <-> REPLICA sync: Loading DB in memory");
         startLoading(server.repl_transfer_size, RDBFLAGS_REPLICATION, asyncLoading);
-
+        if (replicationSupportBypassCRC(conn, use_diskless_load, usemark)) {
+            /* We can bypass CRC checks when data is transmitted through a verified stream.
+             * The usemark flag indicates that the primary is streaming the data directly without
+             * writing it to storage.
+             * Similarly, the use_diskless_load flag indicates that the
+             * replica will load the payload directly into memory without first writing it to disk. */
```
madolson: This comment is documenting the internal behavior of the function, which means if the function changes this comment is unlikely to get updated. I would move the comment to `replicationSupportBypassCRC`.

src/rdb.c
```diff
@@ -3614,6 +3624,12 @@ int rdbSaveToReplicasSockets(int req, rdbSaveInfo *rsi) {
         }
         serverSetCpuAffinity(server.bgsave_cpulist);
 
+        if (bypass_crc) {
+            serverLog(LL_NOTICE, "CRC checksum is disabled for this RDB transfer");
+            // mark rdb object to skip CRC checksum calculations
```
madolson:
```suggestion
```
This comment is just describing the next line, which seems very self explanatory to me.

src/replication.c
```diff
@@ -2494,6 +2509,7 @@ char *sendCommand(connection *conn, ...) {
     while (1) {
         arg = va_arg(ap, char *);
         if (arg == NULL) break;
+        if (strcmp(arg, "") == 0) continue;
```
madolson: A few things things:
1. Empty strings can be valid arguments, this might break some obscure functionality like if someone set a password like an empty string. (I don't think that is allowed today, but would like to be more resistant to weird edge cases. You can introduce an arbitrary static string that points to nothing to have the same behavior.
2. If you continue with this path, you should use strncmp() where possible. You can actually just do `if (arg[0] == '\0')` as well.

madolson (commented):
The code looks okay, mostly some nits to improve clarity.

src/server.c
talxsha: This metric is merely for testing purposes

tests/integration/bypass-crc.tcl
talxsha: I used the naming used in other tests I saw just to stay with format. Will change it

tests/integration/bypass-crc.tcl
talxsha: Correct. Shouldn't we have tests to check that without TLS we do not skip CRC?

src/server.c
madolson: We generally prefer using log lines for testing purposes, since these info fields are part of our public facing contract and we can't change them.

tests/integration/bypass-crc.tcl
madolson: Can't we just have one test for that though? do we really need the 6 that we have?

src/replication.c
ranshid: IMO using some const indicator is not the correct way. Can we just use SendCommandArgv?

madolson (approved):

ranshid:
@talxsha please fix the spellcheck/format issues and also the tests should probably be tagged with cluster:skip

tests/integration/skip-rdb-checksum.tcl
```diff
@@ -0,0 +1,49 @@
...
+    set primary [srv 0 client]
+    set primary_host [srv 0 host]
+    set primary_port [srv 0 port]
+    set primary_skipped_rdb_checksum_counter 0
+    if {$::tls} {
+        foreach primary_diskless_sync {no yes} {
+            foreach replica_diskless_load {disabled on-empty-db swapdb flush-before-load} {
+                test "Skip RDB checksum sync - tls:$::tls, repl_diskless_sync:$primary_diskless_sync, repl_diskless_load:$replica_diskless_load" {
+                    test_skip_rdb_checksum $primary $primary_host $primary_port $primary_skipped_rdb_checksum_counter $primary_diskless_sync $replica_diskless_load
+                } {} {cluster:skip}
```
ranshid: I think the cluster:skip rag can be moved to the start_server. this way we avoid creating cluster for nothing.

PR #1482, opened 2024-12-24
Cleanup aof temp files in sigShutdownHandler before exiting

enjoy-binbin:
I think this code previously only meant to handle save in shutdown,
which means we may a do foreground save in shutdown. It also implicitly
handles the backgroud save, for example the child process will also
get the chance to call getpid() and clean up the temp RDB file.

However, we did not handle AOFRW, so we also leave temp files in here.
Noted that we can not use the same aofRemoveTempFile(getpid()) trick
for AOF. Since we rename the temp file (temp-rewriteaof-bg-pid.aof) to
the base aof file only in backgroundRewriteDoneHandler, and in this time,
there is no child process, so the cleanup job must be done by the parent
process.

The following sequences have been tested (not just signal handlers):
1. Long foreground save + shutdown
2. Long backgroud save + shutdown
3. Long backgroud save + long lua script + shutdown
4. Long bgrewriteaof + shutdown
5. Long bgrewriteaof + long lua script + shutdown
6. Short bgrewriteaof + long done handler + shutdown

All temporary files will be removed in the local tests.

### DIFF
--- a/src/aof.c
+++ b/src/aof.c
@@ -922,7 +922,7 @@ void killAppendOnlyChild(void) {
     if (kill(server.child_pid, SIGUSR1) != -1) {
         while (waitpid(-1, &statloc, 0) != server.child_pid);
     }
-    aofRemoveTempFile(server.child_pid);
+    aofRemoveTempFile(server.child_pid, 0);
     resetChildState();
     server.aof_rewrite_time_start = -1;
 }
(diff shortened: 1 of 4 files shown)


enjoy-binbin:
@zuiderkwast This is a minor cleanup, would you like to review it when you have time?

src/aof.c
```diff
@@ -2470,14 +2470,37 @@ void bgrewriteaofCommand(client *c) {
...
-
-    snprintf(tmpfile, 256, "temp-rewriteaof-bg-%d.aof", (int)childpid);
-    bg_unlink(tmpfile);
-
-    snprintf(tmpfile, 256, "temp-rewriteaof-%d.aof", (int)childpid);
-    bg_unlink(tmpfile);
+    char tmpfile2[256];
+    char pid[32];
+
+    /* Generate temp aof file name using aync-signal safe functions. */
```
zuiderkwast:
```suggestion
    /* Generate temp aof file name using async-signal safe functions. */
```

src/aof.c
```diff
@@ -2470,14 +2470,37 @@ void bgrewriteaofCommand(client *c) {
     }
 }
 
-void aofRemoveTempFile(pid_t childpid) {
+/* Note that we may call this function in signal handle 'sigShutdownHandler',
```
zuiderkwast:
```suggestion
/* Note that we may call this function in signal handler 'sigShutdownHandler',
```

zuiderkwast (approved):
Makes sense to me.

PR #1485, opened 2024-12-24
Offload replication writes to IO threads

uriyage:
This PR offloads the write to replica clients to IO threads.

## Main Changes

* Replica writes will be offloaded but only after the replica is in online mode..
* Replica reads will still be done in the main thread to reduce complexity and because read traffic from replicas is negligible.
### Implementation Details

In order to offload the writes, `writeToReplica` has been split into 2 parts:
1. The write itself made by the IO thread or by the main thread
2. The post write where we update the replication buffers refcount will be done in the main-thread after the write-job is done in the IO thread (similar to what we do with a regular client)

### Additional Changes

* In `writeToReplica` we now use `writev` in case more than 1 buffer exists.
* Changed client `nwritten` field to `ssize_t` since with a replica the `nwritten` can theoretically exceed `int` size (not subject to `NET_MAX_WRITES_PER_EVENT` limit).
* Changed parsing code to use `memchr` instead of `strchr`:
  * During parsing command, ASAN got stuck for unknown reason when called to `strchr` to look for the next `\r`
  * Adding assert for null-terminated querybuf didn't resolve the issue.
  * Switched to `memchr` as it's more secure and resolves the issue

### Testing
* Added integration tests
* Added unit tests

**Related issue:** #761


src/networking.c
```diff
@@ -2007,36 +2007,121 @@ client *lookupClientByID(uint64_t id) {
...
+    /* Determine the last block and buffer position based on thread context */
+    if (inMainThread()) {
+        last_node = listLast(server.repl_buffer_blocks);
+        if (!last_node) return;
+        bufpos = ((replBufBlock *)listNodeValue(last_node))->used;
+    } else {
+        last_node = c->io_last_reply_block;
+        serverAssert(last_node != NULL);
+        bufpos = c->io_last_bufpos;
+    }
```
xbasel: Consider simplifying this code to reduce duplication and improve clarity, eg:
```
    listNode *last_node = inMainThread() ? listLast(server.repl_buffer_blocks) : c->io_last_reply_block;
    if (!last_node) return;

    size_t bufpos = inMainThread() ? 
        ((replBufBlock *)listNodeValue(last_node))->used :  c->io_last_bufpos;
```

src/unit/test_networking.c
```diff
@@ -4,6 +4,305 @@
...
+    }
+
+    /* Test 2: Single block partial write */
+    {
+        replBufBlock *block = zmalloc(sizeof(replBufBlock) + 128);
+        block->size = 128;
+        block->used = 100;
+        block->refcount = 1;
+
+        listAddNodeTail(server.repl_buffer_blocks, block);
```
xbasel, line 221: Looks like this code is duplicated, I suggesting refactoring this into:
`
appendReplBufBlock(size_t size, size_t used) ;
`

src/networking.c
uriyage: I believe the current version is clearer since we check `inMainThread()` only once. Additionally, we handle the `!last_node` case differently depending on whether we are in the main thread or not.

src/unit/test_networking.c
uriyage: The size and used values differ each time

xbasel (commented):
fyi - I tested this code and I got this error. I did not see it with HEAD~1

`
2576783:S 06 Jan 2025 16:06:12.381 # Protocol error (Master using the inline protocol. Desync?) from client: id=6 addr=127.0.0.1:6379 laddr=127.0.0.1:56284 fd=11 name=*redacted* age=43 idle=0 flags=M db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=22870 qbuf-free=18084 argv-mem=0 multi-mem=0 rbs=1024 rbp=42 obl=0 oll=0 omem=0 tot-mem=42880 events=r cmd=set user=*redacted* redir=-1 resp=2 lib-name= lib-ver= tot-net-in=218001645 tot-net-out=1809 tot-cmds=1267450. Query buffer during protocol error: 'SET..$16..key:__rand_int__..$128..VXKeHogKgJ=[5V9_X^b?48OKF2jGA<' (... more 896 bytes ...) 'mcS2^N1J?ELSX@CfKQ7cM5aea\ngY8a3LGgNVa9eRA46XS8>7ABe1>Jl9O\Rm\..'
`

I did run: src/valkey-benchmark -t set -d 128 -n 5000000 --threads 10
valkey with 4 io threads


Looks like data is corrupted, look at the second message:
`
*3\r
$3\r
SET\r
$16\r
key:000000630274\r
$128\r
VXKeHogKgJ=[5V9_X^b?48OKF2jGA<f:iR@50o7dS3JV4Q6L68lC[GTA]0DaMg?_oSmcS2^N1J?ELSX@CfKQ7cM5aea\\ngY8a3LGgNVa9eRA46XS8>7ABe1>Jl9O\\Rm\\\r
*3\r
$3\r
SET\r
$16\r
key:000000420097\r
$128\r
VXKeHogKgJ=[5V9_X^b?48OKF2jGA<f:iR@50o7dS3JV4Q6L68lC[GTA]0DaMg?_oSmcS2^N1J?ELSX@CfKQ7cM5aea\\ngY8a3LGgNVa9eRA46XS8>7ABe1>Jl9O\\Rm\\\r
o7dS3JV4Q6L68lC[GTA]0DaMg?_oSmcS2^N1J?ELSX@CfKQ7cM5aea\\ngY8a3LGgNVa9eRA46XS8>7ABe1>Jl9O\\Rm\\\r
`

The payload `o7dS3JV4Q6L68lC[GTA]0DaMg?_oSmcS2^N1J?ELSX@CfKQ7cM5aea\\ngY8a3LGgNVa9eRA46XS8>7ABe1>Jl9O\\Rm\\\r` is duplicated.

uriyage:
> fyi - I tested this code and I got this error. I did not see it with HEAD~1

Many thanks @xbasel for finding it. 
I fixed the issue, it was a stop condition that was mistakenly removed.

tests/unit/networking.tcl
```diff
@@ -285,6 +285,46 @@ start_server {config "minimal.conf" tags {"external:skip"} overrides {enable-deb
...
+                } else {
+                    fail "Replication not started."
+                }
+                
+                # get the current io_threaded_writes_processed
+                set info [$master info stats]
+                set io_threaded_writes_processed [getInfoProperty $info io_threaded_writes_processed]
+                
+                # Send a write command to the master
+                $master set a 1
```
madolson:
```suggestion
                # Send a write command to the primary
                $primary set a 1
```
We were supposed to move to inclusive language awhile ago, I guess we're not backporting it but it would be good to use it.

src/replication.c
```diff
@@ -4136,6 +4136,8 @@ void replicationCachePrimary(client *c) {
     serverAssert(server.primary != NULL && server.cached_primary == NULL);
     serverLog(LL_NOTICE, "Caching the disconnected primary state.");
 
+    /* Wait for IO operations to be done before proceeding */
```
madolson: Why? I guess I don't immediately understand why we need to wait for IO on these client unlinks, but not the other places we free the client. EDIT: I suppose this is related to the fact the client might be re-used later.

src/networking.c
```diff
@@ -2007,36 +2007,122 @@ client *lookupClientByID(uint64_t id) {
...
+
+    listNode *first_node = c->ref_repl_buf_node;
+
+    /*Handle the single block case */
```
```suggestion
    /* Handle the single block case */
```
nit

src/networking.c
```diff
@@ -2007,36 +2007,122 @@ client *lookupClientByID(uint64_t id) {
...
-            if (nwritten <= 0) {
-                c->write_flags |= WRITE_FLAGS_WRITE_ERROR;
-                return;
-            }
-            c->nwritten += nwritten;
-            c->ref_block_pos += nwritten;
+    /* Determine the last block and buffer position based on thread context */
+    if (inMainThread()) {
+        last_node = listLast(server.repl_buffer_blocks);
+        if (!last_node) return;
```
madolson, line 1981: When can last_node be NULL here? We should only be calling writeToReplica when there is data right?

src/io_threads.c
```diff
@@ -345,6 +345,7 @@ int trySendReadToIOThreads(client *c) {
     c->cur_tid = tid;
     c->read_flags = canParseCommand(c) ? 0 : READ_FLAGS_DONT_PARSE;
     c->read_flags |= authRequired(c) ? READ_FLAGS_AUTH_REQUIRED : 0;
+    c->read_flags |= c->flag.primary ? READ_FLAGS_PRIMARY : 0;
```
madolson: This and the replica flag aren't really read or write flags. Maybe we should just have these act as a separate set of flags used by the IO system that are just set once?

src/unit/test_networking.c
```diff
@@ -4,6 +4,305 @@
...
+/* Mock structures and functions */
+typedef struct mockConnection {
+    connection conn;
+    int error;
+    char *buffer;
+    size_t buf_size;
+    size_t written;
+} mockConnection;
+
+/* Mock connWrite function */
```
madolson: Sort of nitpicky, but this is a fake and not a mock, since it is an alternative simplified implementation. A mock would have configurable return objects.

src/networking.c
```diff
@@ -2778,7 +2876,7 @@ void processMultibulkBuffer(client *c) {
         serverAssertWithInfo(c, NULL, c->argc == 0);
 
         /* Multi bulk length cannot be read without a \r\n */
-        newline = strchr(c->querybuf + c->qb_pos, '\r');
+        newline = memchr(c->querybuf + c->qb_pos, '\r', sdslen(c->querybuf) - c->qb_pos);
```
madolson, line 2810: Was this related to to the stopping condition that was mentioned? strchr should always work, because c->querybuf should always be null terminated.

uriyage: For some reason in dual-channel testswith ASan, we're experiencing timeout issues when the querybuf reaches ~4MB. The problem occurs because ASan takes an unusually long time to execute `strchr` operations, which run four times per command on a buffer filled with short set commands. While ASan doesn't report any errors, it takes approximately 10 minutes to process the querybuf. I know ASan implements a special interceptor for `strchr` and it may be slow see similar issue with `strstr` https://github.com/google/sanitizers/issues/1832, but I haven't yet found why this issue only happens when the I/O thread allocates the buffer. Regardless I believe `memchr` is more safer.

src/unit/test_networking.c
uriyage: Thanks, changed.

src/io_threads.c
uriyage: It would require an additional flag that needs to be turned on/off everywhere we toggle between primary and replica. I'm not entirely sure this would be the better approach.

src/networking.c
uriyage: I believe you're right, but I'm hesitant to deviate from the current implementation which allows it to be null.
```c
static void writeToReplica(client *c) {
    ...
    while (clientHasPendingReplies(c)) {
```

src/networking.c
uriyage: Thanks, fixed.

src/replication.c
uriyage: I'm not sure if this is required, but I thought it would be safer to check that no I/O operations are running before unlinking the client.  After thinking about it, we already wait for client I/O in freeClient, so this check isn't needed here since it's called from freeClient. I moved this check inside unlinkClient since it might be called from other places without calling freeClient first.

tests/unit/networking.tcl
uriyage: Thanks, changed.

madolson (approved):

ranshid (approved):
@uriyage Please just resolve the conlicts. We will merge it right after the tests are passing

uriyage:
> @uriyage Please just resolve the conlicts. We will merge it right after the tests are passing

@ranshid, Done.

zuiderkwast:
Use-after-free in unit test (ASAN & Valgrind), in freeReplicaReferencedReplBuffer.

https://github.com/valkey-io/valkey/actions/runs/13230922385/job/36927928217#step:10:147

uriyage:
> Use-after-free in unit test (ASAN & Valgrind), in freeReplicaReferencedReplBuffer.
> 
> https://github.com/valkey-io/valkey/actions/runs/13230922385/job/36927928217#step:10:147

Fixed in: https://github.com/valkey-io/valkey/pull/1697/files

PR #1486, opened 2024-12-24
Fix: restore replica output bytes stat update

uriyage:
This PR fixes the missing stat update for `total_net_repl_output_bytes` that was removed during the refactoring in PR #758. The metric was not being updated when writing to replica connections.

Changes:
- Restored the stat update in postWriteToClient for replica connections
- Added integration test to verify the metric is properly updated

### DIFF
--- a/tests/integration/replication.tcl
+++ b/tests/integration/replication.tcl
@@ -194,6 +194,30 @@ start_server {tags {"repl external:skip"}} {
             }          
             assert_match {*calls=1,*,rejected_calls=0,failed_calls=1*} [cmdrstat blpop $B]
         }
+        
+        test {Replica output bytes metric} {
+            # reset stats 
+            $A config resetstat
+            
+            set info [$A info stats]
+            set replica_bytes_output [getInfoProperty $info "total_net_repl_output_bytes"]
+            assert_equal $replica_bytes_output 0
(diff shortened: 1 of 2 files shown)


tests/integration/replication.tcl
```diff
@@ -194,6 +194,30 @@ start_server {tags {"repl external:skip"}} {
...
+            wait_for_condition 50 100 {
+                [$B get key] eq {value}
+            } else {
+                fail "Replica did not receive the command"
+            }
+            
+            # get the new stats
+            set info [$A info stats]
+            set replica_bytes_output [getInfoProperty $info "total_net_repl_output_bytes"]
+            assert {$replica_bytes_output > 0}
```
enjoy-binbin:
```suggestion
            assert_morethan $replica_bytes_output 0
```

enjoy-binbin (approved):

tests/integration/replication.tcl
uriyage: @enjoy-binbin We want to assert it is not equal to zero, perhaps you mean `assert_morethan`?

enjoy-binbin: yes, i copy the wrong one, it is assert_morethan

PR #1487, opened 2024-12-25
Add `availability_zone` to the HELLO response

rueian:
It's inconvenient for client implementations to extract the `availability_zone` information from the `INFO` response. The `INFO` response contains a lot of information that a client implementation typically doesn't need.

This PR adds the availability zone to the `HELLO` response. Clients usually already use the `HELLO` command for protocol negotiation and also get the server `version` and `role` from its response. To keep the `HELLO` response small, the field is only added if availability zone is configured.

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -4195,7 +4195,7 @@ void helloCommand(client *c) {
 
     /* Let's switch to the specified RESP mode. */
     if (ver) c->resp = ver;
-    addReplyMapLen(c, 6 + !server.sentinel_mode);
+    addReplyMapLen(c, 6 + !server.sentinel_mode + (sdslen(server.availability_zone) != 0));
 
     addReplyBulkCString(c, "server");
     addReplyBulkCString(c, server.extended_redis_compat ? "redis" : SERVER_NAME);
@@ -4224,6 +4224,11 @@ void helloCommand(client *c) {
 
     addReplyBulkCString(c, "modules");
     addReplyLoadedModules(c);
+
+    if (sdslen(server.availability_zone) != 0) {
+        addReplyBulkCString(c, "availability_zone");
+        addReplyBulkCBuffer(c, server.availability_zone, sdslen(server.availability_zone));
+    }
 }
 
 /* This callback is bound to POST and "Host:" command names. Those are not
--- a/tests/unit/protocol.tcl
+++ b/tests/unit/protocol.tcl
@@ -255,6 +255,14 @@ start_server {tags {"protocol hello"}} {
 
         set reply [r HELLO 2]
         assert_equal [dict get $reply availability_zone] myzone
+
+        r CONFIG SET availability-zone ""
+
+        set reply [r HELLO 3]
+        assert_equal [dict exists $reply availability_zone] 0
+
+        set reply [r HELLO 2]
+        assert_equal [dict exists $reply availability_zone] 0
     }
 }
 
--- a/tests/unit/protocol.tcl
+++ b/tests/unit/protocol.tcl
@@ -232,6 +232,32 @@ start_server {tags {"protocol network"}} {
 
 }
 
+start_server {tags {"protocol hello"}} {
+    test {HELLO without protover} {
+        set reply [r HELLO 3]
+        assert_equal [dict get $reply proto] 3
+
+        set reply [r HELLO]
+        assert_equal [dict get $reply proto] 3
+
+        set reply [r HELLO 2]
+        assert_equal [dict get $reply proto] 2
+
+        set reply [r HELLO]
+        assert_equal [dict get $reply proto] 2
+    }
+
+    test {HELLO and availability-zone} {
+        r CONFIG SET availability-zone myzone
+
+        set reply [r HELLO 3]
+        assert_equal [dict get $reply availability_zone] myzone
+
+        set reply [r HELLO 2]
+        assert_equal [dict get $reply availability_zone] myzone
+    }
+}
+
 start_server {tags {"regression"}} {
     test "Regression for a crash with blocking ops and pipelining" {
         set rd [valkey_deferring_client]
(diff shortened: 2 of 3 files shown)


src/networking.c
```diff
@@ -4224,6 +4224,9 @@ void helloCommand(client *c) {
 
     addReplyBulkCString(c, "modules");
     addReplyLoadedModules(c);
+
+    addReplyBulkCString(c, "availability_zone");
+    addReplyBulkCString(c, server.availability_zone);
```
zuiderkwast: Shall we only include it if it's non-empty? So it doesn't take any space when availability zone is not used. I think we should keep the HELLO response small.

zuiderkwast (commented):
I like the idea but i think we should only include it if it's configured.

src/networking.c
rueian: updated.

zuiderkwast (approved):
@valkey-io/core-team please approve if you agree.

enjoy-binbin (approved):
LGTM, please also add a test, i think it can be fit in protocol.tcl

rueian:
> LGTM, please also add a test, i think it can be fit in protocol.tcl

Thanks for the direction but how about we add the test in the tracking.tcl? That is the only place I found that has tests for the HELLO command.

enjoy-binbin:
tracking.tcl seem odd, the test does not do anything close to tracking, i suggest we also move `HELLO without protover` to protocol.tcl.

rueian:
> tracking.tcl seem odd, the test does not do anything close to tracking, i suggest we also move `HELLO without protover` to protocol.tcl.

Updated.

tests/unit/protocol.tcl
```diff
@@ -232,6 +232,32 @@ start_server {tags {"protocol network"}} {
...
+    }
+
+    test {HELLO and availability-zone} {
+        r CONFIG SET availability-zone myzone
+
+        set reply [r HELLO 3]
+        assert_equal [dict get $reply availability_zone] myzone
+
+        set reply [r HELLO 2]
+        assert_equal [dict get $reply availability_zone] myzone
```
hwware, line 257: Here I think you miss one test case when you set availability-zone as ""

rueian: Indeed. I have added it below. Thanks!

hwware (approved):

src/networking.c
```diff
@@ -4224,6 +4224,11 @@ void helloCommand(client *c) {
 
     addReplyBulkCString(c, "modules");
     addReplyLoadedModules(c);
+
+    if (sdslen(server.availability_zone) != 0) {
+        addReplyBulkCString(c, "availability_zone");
+        addReplyBulkCBuffer(c, server.availability_zone, sdslen(server.availability_zone));
+    }
```
hpatro, line 4231: Should we just reply with empty string if AZ information doesn't exist? The client doesn't need to figure out if the server supports this field or not then.

zuiderkwast: That was the original implementation, but I requested that we only include it if non-empty, i.e. the feature is used. This is to keep the HELLO reply small.

A client who wants to support older versions still need to handle the case that this field is missing.

hpatro: Ack.

madolson (approved):

rueian:
Thanks for all the reviewing. Here is the PR for doc changes: https://github.com/valkey-io/valkey-doc/pull/207.

tests/unit/protocol.tcl
```diff
@@ -232,6 +232,40 @@ start_server {tags {"protocol network"}} {
...
 
+start_server {tags {"protocol hello"}} {
+    test {HELLO without protover} {
+        set reply [r HELLO 3]
+        assert_equal [dict get $reply proto] 3
+
+        set reply [r HELLO]
+        assert_equal [dict get $reply proto] 3
+
+        set reply [r HELLO 2]
```
madolson, line 243: We got a test failure here: https://github.com/valkey-io/valkey/actions/runs/12661473464/job/35284720606#step:6:8973. Looks like the previous test had the `logreqres:skip` tag.

rueian: > We got a test failure here: https://github.com/valkey-io/valkey/actions/runs/12661473464/job/35284720606#step:6:8973. Looks like the previous test had the `logreqres:skip` tag.

Hi @madolson, I have opened a PR to add the tag https://github.com/valkey-io/valkey/pull/1528.

madolson: So fast!

zuiderkwast: The reply-schemas-validator job in Daily runs with the following options:

    ./runtest --log-req-res --no-latency --dont-clean --force-resp3 --tags -slow --verbose --dump-logs

I think the problem is `--force-resp3`. For this flag, there is some code in the test client `tests/support/valkey.tcl` that rewrites "HELLO 2" to "HELLO 3".

```tcl
    # Transform HELLO 2 to HELLO 3 if force_resp3
    # All set the connection var testing_resp3 in case of HELLO 3
    if {[llength $argv] > 0 && [string compare -nocase $method "HELLO"] == 0} {
        if {[lindex $argv 0] == 3} {
            set ::valkey::testing_resp3($id) 1
        } else {
            set ::valkey::testing_resp3($id) 0
            if {$::force_resp3} {
                # If we are in force_resp3 we run HELLO 3 instead of HELLO 2
                lset argv 0 3
            }
        }
    }
```

Adding the tag `logreqres:skip` makes sure the test case is excluded? But it couples `--log-req-res` with `--force-resp3` then...

Another way I've seen some test suites do is just `if {$::force_resp3} { ... }`.

Additionally: All tests using `HELLO 3` seem to have the `resp3` tag. Maybe we should add it too?

zuiderkwast: Hehe, you're both very fast!

PR #1488, opened 2024-12-26
Add build folder to gitignore.

zhulipeng:
Default cmake build folder in vscode is `"cmake.buildDirectory": "${workspaceFolder}/build"`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,7 @@ nodes*.conf
 tests/cluster/tmp/*
 tests/rdma/rdma-test
 tags
+build/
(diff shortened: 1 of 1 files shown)


rjd15372:
For me it also makes sense to use the `build` directory for building valkey using cmake, which is the default directory for using cmake.

madolson (approved):
I feel like this came up at some point related to build-release. I'm OK with this.

rjd15372 (approved):

enjoy-binbin (approved):

PR #1489, opened 2024-12-26
refactor: move all valkey modules related declarations to `module.h`

rjd15372:
In this commit we move all structures and functions declarations related to Valkey modules from `server.h` to the recently added `module.h` file.

This re-organization makes it easier for new contributers to find the valkey modules related code, as well as reducing the compilation times when changes are made to the modules code.

Signed-off-by: Ricardo Dias <ricardo.dias@percona.com>


src/module.h
```diff
@@ -5,13 +5,226 @@
...
+    } while (0)
+
+void moduleEnqueueLoadModule(sds path, sds *argv, int argc);
+sds moduleLoadQueueEntryToLoadmoduleOptionStr(ValkeyModule *module);
+struct ValkeyModuleCtx *moduleAllocateContext(void);
+void moduleScriptingEngineInitContext(struct ValkeyModuleCtx *out_ctx,
                                       ValkeyModule *module,
                                       client *client);
-void moduleFreeContext(ValkeyModuleCtx *ctx);
+void moduleFreeContext(struct ValkeyModuleCtx *ctx);
```
zuiderkwast: Is there a reason for using the `struct` name instead of the typedef, other than copying the existing content from server.h? We normally use the typedef names, but I know we're not consistent and this is just move code, so I will not insist.

src/module.c
```diff
@@ -644,6 +651,33 @@ void *VM_PoolAlloc(ValkeyModuleCtx *ctx, size_t bytes) {
...
+
+    line = sdsnew("loadmodule ");
+    line = sdscatsds(line, module->loadmod->path);
+    for (int i = 0; i < module->loadmod->argc; i++) {
+        line = sdscatlen(line, " ", 1);
+        line = sdscatsds(line, module->loadmod->argv[i]->ptr);
+    }
+
+    return line;
+}
```
zuiderkwast: Please add a short comment above the function.

It returns config file syntax and was in config.c before. By moving it to module.c you managed to make the module load queue struct opaque, which is good, but it splits the "loadmodule" handling between these two files.

The lines `strcasecmp(argv[0], "loadmodule")` and `rewriteConfigRewriteLine(state, "loadmodule", line, 1);` are still in config.c. Perhaps we can let config.c own the config directive name "loadmodule" and pass it to this function?

In config.c:

```C
sds line = moduleLoadQueueEntryToLoadmoduleOptionStr("loadmodule", module);
```

And here:

```C
sds moduleLoadQueueEntryToLoadmoduleOptionStr(const char *prefix, ValkeyModule *module) {
    sds line = sdsnew(prefix);
    line = sdscatlen(line, " ", 1);
    ...
```

src/module.h
```diff
@@ -5,13 +5,226 @@
...
+    struct serverObject *key;    /* Optional name of key processed */
+    int dbid;                    /* The dbid of the key being processed, -1 when unknown. */
+    sds pre_flush_buffer;        /* A buffer that should be flushed before next write operation
+                                  * See rdbSaveSingleModuleAux for more details */
+};
+
+/* Macro to initialize an IO context. Note that the 'ver' field is populated
+ * inside rdb.c according to the version of the value to load. */
+#define moduleInitIOContext(iovar, mtype, rioptr, keyptr, db) \
+    do {                                                      \
```
zuiderkwast: Moved code, right? I'd prefer a static inline function instead of a macro, but since it's just moved code, i will not require it.

zuiderkwast (commented):
Thanks! Just some minor comments.

src/module.h
rjd15372: No reason at all, and I also think we should be consistent across the code base, and use typedef names. I missed this one, thanks for catching it.

src/module.h
rjd15372: Yeah, I just moved the code. I'll change this to an inline function.

src/module.c
```diff
@@ -665,10 +665,11 @@ void moduleEnqueueLoadModule(sds path, sds *argv, int argc) {
     listAddNodeTail(server.loadmodule_queue, loadmod);
 }
 
-sds moduleLoadQueueEntryToLoadmoduleOptionStr(ValkeyModule *module) {
+sds moduleLoadQueueEntryToLoadmoduleOptionStr(ValkeyModule *module,
+                                              const char *config_option_str) {
     sds line;
 
-    line = sdsnew("loadmodule ");
+    line = sdsnew(config_option_str);
```
zuiderkwast, line 672: We need to add space after "loadmodule".

```suggestion
    line = sdsnew(config_option_str);
    line = sdscatlen(line, " ", 1);
```

rjd15372: wow, good catch. I hope that the tests that are currently running catch this bug.

src/module.h
```diff
@@ -117,64 +117,66 @@ struct ValkeyModule {
...
  * a data structure, so that a digest can be created in a way that correctly
  * reflects the values. See the DEBUG DIGEST command implementation for more
  * background. */
-struct ValkeyModuleDigest {
-    unsigned char o[20];      /* Ordered elements. */
-    unsigned char x[20];      /* Xored elements. */
-    struct serverObject *key; /* Optional name of key processed */
-    int dbid;                 /* The dbid of the key being processed */
-};
+typedef struct ValkeyModuleDigest {
```
zuiderkwast, line 159: Nit or maybe just FYI: If it's not a recursive struct definition, it doesn't even need to have a struct name in recent C versions.

rjd15372: I didn't know that. Thanks!

src/module.c
zuiderkwast: Maybe we don't? The tests seem to be passing.

src/module.h
rjd15372: Ah, but now I have a problem. I don't think it's possible to forward declare an "unamed" struct.

zuiderkwast: Yeah, so keep the struct name in this case. :)

src/module.c
rjd15372: phew, tests caught it!

zuiderkwast (approved):

zuiderkwast:
Great, I merged so it doesn't block your next PR. :)

PR #1492, opened 2024-12-28
Immediately restart the defrag cycle if we still need to defrag

### LINKED ISSUE #1454: [Test failure] Active Defrag on large item: cluster
There seems to be one active defrag issue left, where the defragmentation is around 1.1 - 1.2 when it is supposed to be less than 1.1. I am at a bit of a loss for this one, I wasn't able to figure out any local reproduction.

Some example: 
1. https://github.com/valkey-io/valkey/actions/runs/12363680454/job/34505495158#step:8:6531
2. https://github.com/valkey-io/valkey/actions/runs/12363680454/job/34505490704#step:8:6457
3. https://github.com/valkey-io/valkey/actions/runs/12363680454/job/34505492618#step:8:6515

@zvi-code @JimB123 Any thoughts?

madolson:
One very subtle change of https://github.com/valkey-io/valkey/pull/1242, is that we now wait for the server cron to start a new defrag cycle. Whereas previously we started it immediately. All of the current tests rely on the behavior of "Active defrag runs until it is done", which was broken because now there is a short period of active defragmentation reports itself off. Running a bunch of tests while this is in a draft, but opening it incase anyone else was taking a look.

Intended to close https://github.com/valkey-io/valkey/issues/1454.

https://github.com/valkey-io/valkey/actions/runs/12524316525/job/34934755959 [Only failure is active-defrag related, but not related to this issue]
https://github.com/valkey-io/valkey/actions/runs/12524547410

### DIFF
--- a/src/defrag.c
+++ b/src/defrag.c
@@ -1113,8 +1113,7 @@ static void endDefragCycle(bool normal_termination) {
     server.stat_last_active_defrag_time = 0;
     server.active_defrag_cpu_percent = 0;
 
-    /* During the defrag cycle, more fragmentation might have accumulated, so
-     * immediately check if we should start another cycle. */
+    /* Immediately check to see if we should start another defrag cycle. */
     monitorActiveDefrag();
 }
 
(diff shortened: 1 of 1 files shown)


madolson:
First full run looked good, so starting a second one but also making it ready for review since I'm going to sleep.

enjoy-binbin (approved):

JimB123:
A simple and elegant fix.  I was aware of this subtle change, but felt it was cleaner to only start the cycle in one place.  I agree with you though that this is an expedient way to stabilize the tests.

PR #1496, opened 2024-12-30
Remove releasetools folder

hwware:
The release tool utils\releasetools\ does not work anymore in Valkey, in this PR, we remove it.


roshkhatri (approved):
LGTM, Yes we do not use this anymore

zuiderkwast (approved):

PR #1497, opened 2024-12-31
Adds support for running `EVAL` with different scripting engines

### LINKED ISSUE #1261: [NEW] Add scripting languages (for EVAL, etc.) using module API

### LINKED ISSUE #1468: Future-Proof `ValkeyModuleScriptingEngineMemoryInfo` by Adding Version
this info struct needs to be versioned. see `ValkeyModuleClientInfo`.

_Originally posted by @PingXie in https://github.com/valkey-io/valkey/pull/1277#discussion_r1894842945_

rjd15372:
In this PR we re-implement the `EVAL` commands (`EVAL`, `EVALSHA`, `SCRIPT LOAD`, etc...) to use the scripting engine infrastructure introduced in 6adef8e2f97c3cd2dd4620fb9d5af8d426f1b548. This allows `EVAL` to run scripts using different scripting engines.

The Lua scripting engine implementation code was moved into its own subdirectory `src/lua`.

This new implementation generalizes the module API for implementing scripting engines to work with both `FUNCTION` and `EVAL` commands.

Module API changes include:
* Rename of callback `ValkeyModuleScriptingEngineCreateFunctionsLibraryFunc` to `ValkeyModuleScriptingEngineCompileCodeFunc`.
* Addition of a new enum `enum ValkeyModuleScriptingEngineSubsystemType` to specify the scripting engine subsystem (EVAL, or FUNCTION, or both).
* In most callbacks was added a new parameter with the type `ValkeyModuleScriptingEngineSubsystemType`.
* New callback specific for EVAL `ValkeyModuleScriptingEngineResetEvalEnvFunc`.
* New API function `ValkeyModuleScriptingEngineExecutionState (*ValkeyModule_GetFunctionExecutionState)(ValkeyModuleScriptingEngineServerRuntimeCtx *server_ctx)` that is used by scripting engines to query the server about the execution state of the script that is running.

Fixes #1261
Fixes #1468
Follow-up of #1277


rjd15372:
@zuiderkwast this is ready for review.

src/lua/debug_lua.c
```diff
@@ -0,0 +1,992 @@
+#include "debug_lua.h"
```
zuiderkwast, line 7: We should have some copyright/license comment at the top. I believe 3 lines are enough:

```
# Copyright xxxxxxxxxxxx
# All rights reserved.
# SPDX-License-Identifier: BSD-3-Clause
```

We use *Copyright 2025- Valkey contributors* for new files, but for moved code we should probably keep the old names and years too.

src/module.c
```diff
@@ -61,6 +61,7 @@
 #include "hdr_histogram.h"
 #include "crc16_slottable.h"
 #include "valkeymodule.h"
+#include "module.h"
 #include "io_threads.h"
 #include "module.h"
```
zuiderkwast: module.h is included twice here

zuiderkwast (commented):
It's a huge change.

The module API changes look good to me.

It's impossible to proof-read the logic but I guess we have pretty good test coverage. Breaking up eval.c into smaller units is good idea, just hard to review. Can you summarize which code was just moved unchanged and which code was written, so it's possible to tell what to review carefully and what not?

rjd15372:
> [...]. Can you summarize which code was just moved unchanged and which code was written, so it's possible to tell what to review carefully and what not?

@zuiderkwast sure.

* Important changes in `eval.c`:
  * All `ldb*` functions (lua debug support) were just moved to `lua/debug_lua.[hc]` so no need to do a thorough review. 
  * Old `scriptingRelease` is now called `evalRelease` and has significantly changed to support the async release of resources for any kind of scripting engine.
  *  `evalRegisterNewScript` function.
  * `evalGenericCommand` function
  * `eval.c` is the solely responsible for managing the cache for scripts indexed by the SHA1 hash of the script body. The scripting engine is not aware of this cache. So `eval.c` implements the `SCRIPT *` and `EVAL*` commands, by offloading the compilation and execution of scripts to the scripting engine.
  * Lua engine specific code, like initialization, compilation of scripts, and execution of scripts, was moved into `lua/script_lua.c`
* `lua/engine_lua.c` is the entry point of the lua scripting engine for supporting both `EVAL` and `FCALL` commands. It has the scripting engine callbacks implementation.
*  `scripting_engine.c` is the scripting engines core that implements the management of the scripting engines that are registered in the server. It was improved with the support for `EVAL`.
* Module API (`valkeymodule.h`) changes:
  * New enum `ValkeyModuleScriptingEngineSubsystemType` to specify the script execution subsystem (either `EVAL`, or `FUNCTION`, or both). Almost all scripting engine API callbacks have a parameter with this enum type.
  * The callback `ValkeyModuleScriptingEngineCreateFunctionsLibraryFunc` was renamed to `ValkeyModuleScriptingEngineCompileCodeFunc`.
  * New callback `ValkeyModuleScriptingEngineResetEvalEnvFunc` to support the reset of the `EVAL` runtime context.
  * New Module API function  `ValkeyModule_GetFunctionExecutionState` that returns the execution state of a script. This is used by scripting engines to know whether the script was killed. Scripting engines will have to poll this function periodically.

src/eval.c
```diff
@@ -35,39 +35,42 @@
...
+    decrRefCount(sh->body);
     zfree(val);
 }
 
 static uint64_t dictStrCaseHash(const void *key) {
     return dictGenCaseHashFunction((unsigned char *)key, strlen((char *)key));
 }
 
-/* lctx.lua_scripts sha (as sds string) -> scripts (as luaScript) cache. */
+/* evalCtx.scripts sha (as sds string) -> scripts (as evalScript) cache. */
```
zuiderkwast, line 74: I can't see a struct called `evalScript`. Is this the same as `scriptHolder`?

I like the name `evalScript` slightly more. The eval-prefix is good.

src/eval.c
```diff
@@ -35,39 +35,42 @@
...
+    listNode *node; /* list node in scripts_lru_list list. */
+} scriptHolder;
 
 static void dictLuaScriptDestructor(void *val) {
     if (val == NULL) return; /* Lazy freeing will set value to NULL. */
-    decrRefCount(((luaScript *)val)->body);
+    scriptHolder *sh = (scriptHolder *)val;
+    decrRefCount(sh->body);
     zfree(val);
 }
```
zuiderkwast: Shouldn't this free the compiledFunction `sh->script` too in some way?

This is not specific to Lua anymore. Remove Lua from the function name?

src/eval.c
```diff
@@ -77,36 +80,29 @@ dictType shaScriptObjectDictType = {
...
-    sds *src;                    /* Lua script source code split by line. */
-    int lines;                   /* Number of lines in 'src'. */
-    int currentline;             /* Current line number. */
-    sds cbuf;                    /* Debugger client command buffer. */
-    size_t maxlen;               /* Max var dump / reply length. */
-    int maxlen_hint_sent;        /* Did we already hint about "set maxlen"? */
-} ldb;
+/* Eval context */
+struct evalCtx {
+    dict *scripts;                  /* A dictionary of SHA1 -> scripts */
```
zuiderkwast: the values are evalScript objects?

src/eval.c
```diff
@@ -135,198 +131,71 @@ void sha1hex(char *digest, char *script, size_t len) {
...
+    dictReleaseIterator(iter);
+    dictRelease(scripts);
+
+    listRelease(scripts_lru_list);
+    if (engine_callbacks) {
+        listIter *iter = listGetIterator(engine_callbacks, 0);
+        listNode *node = NULL;
+        while ((node = listNext(iter)) != NULL) {
+            callableLazyEvalReset *callback = listNodeValue(node);
+            callback->engineLazyEvalResetCallback(callback->context);
```
zuiderkwast: This function is freeEvalScriptsSync but Lazy here looks lika something async. What's happening?

src/eval.c
```diff
@@ -398,6 +274,12 @@ int evalExtractShebangFlags(sds body, uint64_t *out_flags, ssize_t *out_shebang_
             }
         }
         sdsfreesplitres(parts, numparts);
+    } else {
+        // When no shebang is declared, assume the engine is LUA.
+        if (out_engine) {
+            *out_engine = zcalloc(4);
+            valkey_strlcpy(*out_engine, "lua", 4);
```
zuiderkwast:
```suggestion
            *out_engine = zstrdup("lua");
```

src/valkeymodule.h
```diff
@@ -831,12 +831,46 @@ typedef struct ValkeyModuleScriptingEngineMemoryInfo {
...
+typedef enum ValkeyModuleScriptingEngineSubsystemType {
+    VMSE_EVAL,
+    VMSE_FUNCTION,
+    VMSE_ALL
+} ValkeyModuleScriptingEngineSubsystemType;
+
+typedef enum ValkeyModuleScriptingEngineExecutionState {
+    VMSE_STATE_EXECUTING,
+    VMSE_STATE_KILLED,
+} ValkeyModuleScriptingEngineExecutionState;
```
zuiderkwast, line 842: Can there be any other states in the future? PAUSED?

If it's just about whether it's killed or not, we can consider `bool`. WDYT?

src/valkeymodule.h
```diff
@@ -831,12 +831,46 @@ typedef struct ValkeyModuleScriptingEngineMemoryInfo {
...
+
+typedef struct ValkeyModuleScriptingEngineCallableLazyEvalReset {
+    void *context;
+
+    /*
+     * Callback function used for resetting the EVAL context implemented by an
+     * an engine. This callback will be called by a background thread when it's
```
zuiderkwast: Repeated "an"

tests/modules/helloscripting.c
```diff
@@ -28,6 +29,15 @@
  * CONSTI 432    # pushes the value 432 to the top of the stack
  * RETURN        # returns the current value on the top of the stack and marks
  *               # the end of the function declaration.
+ *
+ * FUNCTION sleep # declaration of function 'sleep'
+ * ARGS 0         # pushes the value in the first argument to the top of the
+ *                # stack
+ * SLEEP          # Pops the current value in the stack and sleeps for `value`
```
zuiderkwast: SLEEP is an instruction and `sleep` is the function name. Maybe use another function name to avoid confusion.

tests/unit/moduleapi/scriptingengine.tcl
```diff
@@ -123,6 +123,22 @@ start_server {tags {"modules"}} {
         assert_equal $result 432
     }
 
+    test {Test function kill} {
+        set rd [valkey_deferring_client]
+        r config set busy-reply-threshold 10
+        r function load REPLACE "#!hello name=mylib\nFUNCTION sleep\nARGS 0\nSLEEP\nARGS 0\nRETURN"
```
zuiderkwast: Add a test with EVAL in the hello language too?

src/eval.c
rjd15372: The compiled function is being freed in `freeEvalScriptsSync` before clearing the hashtable that stores the `evalScript` objects. 
I don't remember why I'm not freeing the compiled function here. I'm going move the code to here, since I don't see a good reason to keep it in `freeEvalScriptsSync`.

src/eval.c
rjd15372: This function is called in both cases. In the async case, it will call this function in the background thread. Maybe I'll remove the `Sync` suffix to avoid confusion.

src/valkeymodule.h
rjd15372: I don't know if we will have other states in the future. In general I prefer to use enums, even if it has only two symbols, because of type checking reasons and because it is more informative when reading the code.

But I don't have a strong opinion in this case. If we decide to not use this enum, then I'll change the function signature of 
```c
ValkeyModuleScriptingEngineExecutionState (*ValkeyModule_GetFunctionExecutionState)(ValkeyModuleScriptingEngineServerRuntimeCtx *server_ctx)
```

to

```c
int (*ValkeyModule_IsScriptingEngineFunctionExecutionKilled)(ValkeyModuleScriptingEngineServerRuntimeCtx *server_ctx)
```

What do you think is best?

zuiderkwast: Keep the enum. I agree, enums are more informative.

tests/modules/helloscripting.c
```diff
@@ -28,6 +29,15 @@
  * CONSTI 432    # pushes the value 432 to the top of the stack
  * RETURN        # returns the current value on the top of the stack and marks
  *               # the end of the function declaration.
+ *
+ * FUNCTION sleep # declaration of function 'sleep'
```
zuiderkwast:
```suggestion
 * FUNCTION baz   # declaration of function 'baz'
```

To avoid overloading the name with the sleep instruction. The other functions here are foo and bar.

zuiderkwast (commented):
Just followed up since the last update. I think it looks very close to merging now.

rjd15372:
Hi @PingXie, I've pushed a commit that adds the version field to the structures used by the scripting engine module API as we discussed yesterday.

src/lua/debug_lua.c
PingXie: I wasn't paying much attention to the license header and was under the impression that it is already covered by the LICENSE file on the project. now that on a second thought, I think you have a good point and I agree we should have a per-file license header consistently. however, I don't think the phrase "All rights reserved." is necessary for BSD-licensed code. BSD 3-Clause explicitly allows redistribution and modification, making "All rights reserved" somewhat contradictory. More specifically. I am thinking of the following:

for files carried over

```c
/*
 * Copyright (c) [YEAR], Redis Ltd.
 * SPDX-License-Identifier: BSD-3-Clause
 */
```

for files carried over but modified in a non-trivial way:

```c
/*
 * Copyright (c) [YEAR], Redis Ltd.
 * Copyright (c) [YEAR], Valkey Contributors
 * SPDX-License-Identifier: BSD-3-Clause
 */
```

for new files:

```c
/*
 * Copyright (c) [YEAR], Valkey Contributors
 * SPDX-License-Identifier: BSD-3-Clause
 */
```

src/lua/debug_lua.h
```diff
@@ -0,0 +1,32 @@
+#ifndef _LUA_DEBUG_H_
+#define _LUA_DEBUG_H_
+
+typedef char *sds;
```
PingXie, line 4: this is not a forward but real definition. any issue with including sds.h?

zuiderkwast (approved):
I'd like to merge this now, before RC1, so we can have some chance to test it as part of RC1. It's good to get the module API changes in before we release RC1 so we can avoid breaking changes between the release candidates too (which is not strictly necessary but it's nice). We can fix any issues or code review comments afterwards, before GA.

@PingXie Are you OK with that?

src/server.c
```diff
@@ -2911,12 +2921,25 @@ void initServer(void) {
         serverPanic("Scripting engine manager initialization failed, check the server logs.");
     }
 
+    // Since we initialized the scripting engine manager, we need to ensure
+    // that commands with `CMD_NOSCRIPT` flag are not allowed to run in scripts.
+    //
```
PingXie: nit - let's keep the c style comments for now. I don't think we have closed on the discussion of using c++ style yet.

src/valkeymodule.h
```diff
@@ -850,28 +883,48 @@ typedef struct ValkeyModuleScriptingEngineMemoryInfo {
...
  *
  * - `module_ctx`: the module runtime context.
  *
- * - `engine_ctx`: the engine specific context pointer.
+ * - `engine_ctx`: the scripting engine runtime context.
  *
- * - `func_ctx`: the context opaque structure that represents the runtime
- *               context for the function.
+ * - `server_ctx`: the context opaque structure that represents the server-side
+ * runtime context for the function.
```
PingXie:
```suggestion
 *   runtime context for the function.
```

src/lua/engine_lua.c
```diff
@@ -0,0 +1,392 @@
...
+    serverAssert(module_ctx == NULL);
+
+    luaEngineCtx *lua_engine_ctx = (luaEngineCtx *)engine_ctx;
+    compiledFunction **functions = NULL;
+
+    if (type == VMSE_EVAL) {
+        lua_State *lua = lua_engine_ctx->eval_lua;
+
+        if (luaL_loadbuffer(
+                lua, code, strlen(code), "@user_script")) {
```
PingXie: this line break looks odd.
```suggestion
        if (luaL_loadbuffer(lua, code, strlen(code), "@user_script")) {
```

PingXie (approved):
I really like the new code structure. Great refactoring work @rjd15372!

I have some random nonblocking comments.

src/lua/debug_lua.c
PingXie: Let's move the discussion to #1692

src/server.c
zuiderkwast: We already have many of them and clang-format doesn't say anything, but sure the C89 style is our main style.

```suggestion
    /* Since we initialized the scripting engine manager, we need to ensure that
     * commands with `CMD_NOSCRIPT` flag are not allowed to run in scripts. */
```

rjd15372: Sorry, I spent the last 8 years coding in C++ and my muscle memory wrote this code :smile:

src/lua/debug_lua.h
rjd15372: When I include "sds.h" I get the following build warning:
```
In file included from lua/../server.h:33,
                 from lua/script_lua.h:51,
                 from lua/debug_lua.c:8:
lua/../fmacros.h:37:9: warning: "_DEFAULT_SOURCE" redefined
   37 | #define _DEFAULT_SOURCE
      |         ^~~~~~~~~~~~~~~
In file included from /usr/include/sys/types.h:25,
                 from lua/../sds.h:38,
                 from lua/debug_lua.h:4,
                 from lua/debug_lua.c:7:
/usr/include/features.h:246:10: note: this is the location of the previous definition
  246 | # define _DEFAULT_SOURCE        1
```

We're defining the symbol `_DEFAULT_SOURCE`  in our code in `src/fmacros.h`, and this symbol is also defined in `/usr/include/features.h`. 
Probably we should fix our own code in `src/fmacros.h` to only define `_DEFAULT_SOURCE` if it's not defined already.

src/server.c
zuiderkwast: This is valid C since 1999. It's just our old style convention.

PR #1501, opened 2025-01-01
Accelerate hash table iterator with prefetching

NadavGigi:
This PR introduces improvements to the hashtable iterator, implementing prefetching technique described in the blog post [Unlock One Million RPS - Part 2](https://valkey.io/blog/unlock-one-million-rps-part2/) . The changes lay the groundwork for further enhancements in use cases involving iterators. Future PRs will build upon this foundation to improve performance and functionality in various iterator-dependent operations.

In the pursuit of maximizing iterator performance, I conducted a comprehensive series of experiments. My tests encompassed a wide range of approaches, including processing multiple bucket indices in parallel, prefetching the next bucket upon completion of the current one, and several other timing and quantity variations. Surprisingly, after rigorous testing and performance analysis, the simplest implementation presented in this PR consistently outperformed all other more complex strategies.

## Implementation
Each time we start iterating over a bucket, we prefetch data for future iterations:
- We prefetch the entries of the next bucket (if it exists).
- We prefetch the structure (but not the entries) of the bucket after the next.
This prefetching is done when we pick up a new bucket, increasing the chance that the data will be in cache by the time we need it.

## Performance

The data below was taken by conducting keys command on 64cores Graviton 3 Amazon EC2 instance with 50 mil keys in size of 100 bytes each. The results regarding the duration of “keys *” command was taken from “info all” command.
```
+--------------------+------------------+-----------------------------+
| prefetching        | Time (seconds)   | Keys Processed per Second   |
+--------------------+------------------+-----------------------------+
| No                 | 11.112279        | 4,499,529                   |
| Yes                | 3.141916         | 15,913,862                  |
+--------------------+------------------+-----------------------------+
Improvement:
Comparing the iterator without prefetching to the one with prefetching, 
we can see a speed improvement of 11.112279 / 3.141916 ≈ 3.54 times faster.
```

#### Setup:
- 64cores Graviton 3 Amazon EC2 instance.
-  50 mil keys in size of 100 bytes each.
-  Running valkey server over RAM file system.
-  crc checksum and comperssion off.

#### Results
```
+--------------------+------------------+-----------------------------+
| prefetching        | Time (seconds)   | Keys Processed per Second   |
+--------------------+------------------+-----------------------------+
| No                 | 28               | 1,785,700                   |
| Yes                | 19.6             | 2,550,000                   |
+--------------------+------------------+-----------------------------+
Improvement:
- Reduced SAVE time by 30% (8.4 seconds faster)
- Increased key processing rate by 42.8% (764,300 more keys/second)
```

### DIFF
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -498,7 +498,7 @@ size_t nextCursor(size_t v, size_t mask) {
 }
 
 /* Returns the next bucket in a bucket chain, or NULL if there's no next. */
-static bucket *bucketNext(bucket *b) {
+static bucket *getChildBucket(bucket *b) {
     return b->chained ? b->entries[ENTRIES_PER_BUCKET - 1] : NULL;
 }
 
@@ -548,12 +548,12 @@ static void rehashStep(hashtable *ht) {
     rehashBucket(ht, b);
     if (b->chained) {
         /* Rehash and free child buckets. */
-        bucket *next = bucketNext(b);
+        bucket *next = getChildBucket(b);
         b->chained = 0;
         b = next;
         while (b != NULL) {
             rehashBucket(ht, b);
-            next = bucketNext(b);
+            next = getChildBucket(b);
             zfree(b);
             if (ht->type->trackMemUsage) ht->type->trackMemUsage(ht, -sizeof(bucket));
             ht->child_buckets[0]--;
@@ -708,7 +708,7 @@ static bucket *findBucket(hashtable *ht, uint64_t hash, const void *key, int *po
                     }
                 }
             }
-            b = bucketNext(b);
+            b = getChildBucket(b);
         } while (b != NULL);
     }
     return NULL;
@@ -753,7 +753,7 @@ static void bucketConvertToUnchained(bucket *b) {
  * This function needs the penultimate 'before_last' bucket in the chain, to be
  * able to update it when the last bucket is freed. */
 static void pruneLastBucket(hashtable *ht, bucket *before_last, bucket *last, int table_index) {
-    assert(before_last->chained && bucketNext(before_last) == last);
+    assert(before_last->chained && getChildBucket(before_last) == last);
     assert(!last->chained);
     assert(last->presence == 0 || __builtin_popcount(last->presence) == 1);
     bucketConvertToUnchained(before_last);
@@ -775,10 +775,10 @@ static void fillBucketHole(hashtable *ht, bucket *b, int pos_in_bucket, int tabl
     assert(b->chained && !isPositionFilled(b, pos_in_bucket));
     /* Find the last bucket */
     bucket *before_last = b;
-    bucket *last = bucketNext(b);
+    bucket *last = getChildBucket(b);
     while (last->chained) {
         before_last = last;
-        last = bucketNext(last);
+        last = getChildBucket(last);
     }
     /* Unless the last bucket is empty, find an entry in the last bucket and
      * move it to the hole in b. */
@@ -800,10 +800,10 @@ static void fillBucketHole(hashtable *ht, bucket *b, int pos_in_bucket, int tabl
 static void compactBucketChain(hashtable *ht, size_t bucket_index, int table_index) {
     bucket *b = &ht->tables[table_index][bucket_index];
     while (b->chained) {
-        bucket *next = bucketNext(b);
+        bucket *next = getChildBucket(b);
         if (next->chained && next->presence == 0) {
             /* Empty bucket in the middle of the chain. Remove it from the chain. */
-            bucket *next_next = bucketNext(next);
+            bucket *next_next = getChildBucket(next);
             b->entries[ENTRIES_PER_BUCKET - 1] = next_next;
             zfree(next);
             if (ht->type->trackMemUsage) ht->type->trackMemUsage(ht, -sizeof(bucket));
@@ -846,7 +846,7 @@ static bucket *findBucketForInsert(hashtable *ht, uint64_t hash, int *pos_in_buc
             bucketConvertToChained(ht, b);
             ht->child_buckets[table]++;
         }
-        b = bucketNext(b);
+        b = getChildBucket(b);
     }
     /* Find a free slot in the bucket. There must be at least one. */
     int pos;
@@ -934,6 +934,51 @@ static inline incrementalFind *incrementalFindFromOpaque(hashtableIncrementalFin
     return (incrementalFind *)(void *)state;
 }
 
+/* Prefetches all filled entries in the given bucket to optimize future memory access. */
+static void prefetchBucketEntries(bucket *b) {
+    for (int pos = 0; pos < numBucketPositions(b); pos++) {
+        if (isPositionFilled(b, pos)) {
+            valkey_prefetch(b->entries[pos]);
+        }
+    }
+}
+
+/* Returns the child bucket if chained, otherwise the next bucket in the table. returns NULL if neither exists. */
+static bucket *getNextBucket(bucket *current_bucket, size_t bucket_index, hashtable *ht, int table_index) {
+    bucket *next_bucket = NULL;
+    if (current_bucket->chained) {
+        next_bucket = getChildBucket(current_bucket);
+    } else {
+        size_t table_size = numBuckets(ht->bucket_exp[table_index]);
+        size_t next_index = bucket_index + 1;
+        if (next_index < table_size) {
+            next_bucket = &ht->tables[table_index][next_index];
+        }
+    }
+    return next_bucket;
+}
+
+/* This function prefetches data that will be needed in subsequent iterations:
+ * - The entries of the next bucket
+ * - The next of the next bucket
+ * It attempts to bring this data closer to the L1 cache to reduce future memory access latency.
+ *
+ * Cache state before this function is called(due to last call for this function):
+ * 1. The current bucket and its entries are likely already in cache.
+ * 2. The next bucket is in cache.
+ */
+static void prefetchNextBucketEntries(iter *iter, bucket *current_bucket) {
+    size_t next_index = iter->index + 1;
+    bucket *next_bucket = getNextBucket(current_bucket, next_index, iter->hashtable, iter->table);
+    if (next_bucket) {
+        prefetchBucketEntries(next_bucket);
+        bucket *next_next_bucket = getNextBucket(next_bucket, next_index + 1, iter->hashtable, iter->table);
+        if (next_next_bucket) {
+            valkey_prefetch(next_next_bucket);
+        }
+    }
+}
+
 /* --- API functions --- */
 
 /* Allocates and initializes a new hashtable specified by the given type. */
@@ -979,7 +1024,7 @@ void hashtableEmpty(hashtable *ht, void(callback)(hashtable *)) {
                             }
                         }
                     }
-                    bucket *next = bucketNext(b);
+                    bucket *next = getChildBucket(b);
 
                     /* Free allocated bucket. */
                     if (b != &ht->tables[table_index][idx]) {
@@ -1375,7 +1420,7 @@ int hashtableReplaceReallocatedEntry(hashtable *ht, const void *old_entry, void 
                     return 1;
                 }
             }
-            b = bucketNext(b);
+            b = getChildBucket(b);
         } while (b != NULL);
     }
     return 0;
@@ -1538,8 +1583,8 @@ int hashtableIncrementalFindStep(hashtableIncrementalFindState *state) {
                     bucket_idx = data->hash & mask;
                 }
                 data->bucket = &ht->tables[data->table][bucket_idx];
-            } else if (bucketNext(data->bucket) != NULL) {
-                data->bucket = bucketNext(data->bucket);
+            } else if (getChildBucket(data->bucket) != NULL) {
+                data->bucket = getChildBucket(data->bucket);
             } else if (data->table == 0 && ht->rehash_idx >= 0) {
                 data->table = 1;
                 size_t mask = expToMask(ht->bucket_exp[1]);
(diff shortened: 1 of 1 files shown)


src/hashtable.c
```diff
@@ -934,6 +946,76 @@
...
+        iter->threads[i].curr_bucket = NULL;
+        iter->threads[i].pos = 0;
+        iter->threads[i].index = 0;
+        iter->threads[i].table = 0;
+        iter->threads[i].last_seen_size = 0;
+    }
+}
+
+static void prefetchBucketEntries(bucket *b) {
+    for (uint8_t pos = 0; pos < numBucketPositions(b); pos++) {
```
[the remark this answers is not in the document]
NadavGigi: Resolved

madolson:
How does this compare to having an iterator that actually returns a batch of items. Something like:
```
void *entries[7];
size_t num_entries;
entries = getBatchEntries(iterator, *num_entries);
if (entries) {
    for (size_t i = 0; i < num_entries; i++) {
        whatever(entries[i]);
    }
}
```

I generally prefer to avoid manually executing prefetching when we can just efficiently process the data, as we then give more hints to the compiler and the processor so it can efficiently do its own re-ordering and prefetching.

madolson:
> It's important to note that while we refer to 'threads' in this implementation, we're not actually using operating system threads.

Then don't name them threads, it makes the implementation much harder to follow.

NadavGigi:
> How does this compare to having an iterator that actually returns a batch of items. Something like:
> 
> ```
> void *entries[7];
> size_t num_entries;
> entries = getBatchEntries(iterator, *num_entries);
> if (entries) {
>     for (size_t i = 0; i < num_entries; i++) {
>         whatever(entries[i]);
>     }
> }
> ```
> 
> I generally prefer to avoid manually executing prefetching when we can just efficiently process the data, as we then give more hints to the compiler and the processor so it can efficiently do its own re-ordering and prefetching.

zuiderkwast:
I agree with Madelyn that the word "threads" is not suitable. Can we call it batches?

This looks quite complex. Was this adapted from an implementation for dict?

For hashtable, there is a natural batch of one hashtable bucket (up to 7 elements). Can't we just prefetch all elements in a hashtable bucket and the child bucket if any? When we return the last element in the bucket, we'd prefetch the next bucket's elements. Then we just at read and return the elements as before without keeping track of anything. It'd be much simpler.

For KEYS, we just want to return the elements so I guess it can be faster with a more complex solution, but for other usages like rdb dump and aof rewrite, we do more things after each call to hashtableNext so I would assume most of the benefit can be seen even with the simpler approach.

Did you try the simpler approach already?

src/hashtable.c
```diff
@@ -1828,77 +1902,114 @@ void hashtableReleaseIterator(hashtableIterator *iterator) {
...
-                iter->bucket = &iter->hashtable->tables[iter->table][iter->index];
+                iter->index = 0;
+                iter->table++;
             }
+            initBucketPrefetchInfo(info, iter);
+            valkey_prefetch(info->curr_bucket);
+            info->state = BUCKET_PREFETCH_ENTRIES;
+            break;
+        default:
+            iter->num_of_active_buckets--;
```
uriyage: why decrement if we are about to assert(0)?

src/hashtable.c
```diff
@@ -1828,77 +1902,114 @@ void hashtableReleaseIterator(hashtableIterator *iterator) {
...
-            *elemptr = b->entries[iter->pos_in_bucket];
+        if (iter->hashtable->tables[0] == NULL) return 0; /* Empty hashtable */
+        if (hashtableIsRehashing(iter->hashtable)) {
+            iter->index = iter->hashtable->rehash_idx - 1;
         }
-        return 1;
+    }
+    /* When a bucket prefetch info attempts to pick up a new bucket
+     * and the iterator index exceeds the hashtable size,
+     * that bucket prefetch info finishes, decrementing num_of_active_buckets. */
```
uriyage: Consider adding 
```c
iter->num_of_active_buckets = min(iter->num_of_active_buckets, iter->num_of_buckets);
```
 to avoid repeatedly calling `handleBucketState` for the same bucket when `iter->num_of_buckets is` < `iter->num_of_active_buckets`.

NadavGigi:
> How does this compare to having an iterator that actually returns a batch of items. Something like:
> 
> ```
> void *entries[7];
> size_t num_entries;
> entries = getBatchEntries(iterator, *num_entries);
> if (entries) {
>     for (size_t i = 0; i < num_entries; i++) {
>         whatever(entries[i]);
>     }
> }
> ```
> 
> I generally prefer to avoid manually executing prefetching when we can just efficiently process the data, as we then give more hints to the compiler and the processor so it can efficiently do its own re-ordering and prefetching.
@madolson 
Thank you for your suggestion. If I understand correctly, this approach would necessitate changes to every piece of code that currently uses the iterator, not just the iterator implementation itself. However, I may have misinterpreted your proposal. Could you please provide further explanation or clarification on how this batch iterator could be implemented without requiring widespread changes to the existing codebase?


> > It's important to note that while we refer to 'threads' in this implementation, we're not actually using operating system threads.
> 
> Then don't name them threads, it makes the implementation much harder to follow.

Done.

NadavGigi:
> I agree with Madelyn that the word "threads" is not suitable. Can we call it batches?
> 
> This looks quite complex. Was this adapted from an implementation for dict?
> 
> For hashtable, there is a natural batch of one hashtable bucket (up to 7 elements). Can't we just prefetch all elements in a hashtable bucket and the child bucket if any? When we return the last element in the bucket, we'd prefetch the next bucket's elements. Then we just at read and return the elements as before without keeping track of anything. It'd be much simpler.
> 
> For KEYS, we just want to return the elements so I guess it can be faster with a more complex solution, but for other usages like rdb dump and aof rewrite, we do more things after each call to hashtableNext so I would assume most of the benefit can be seen even with the simpler approach.
> 
> Did you try the simpler approach already?
@zuiderkwast 
1. I have changed the name "threads" to bucketPrefetchInfo since its the number of buckets we scan in parrallel.
2. Yes, it was adapted from dict implementation.
3. No, I have tried a simpler version that benefits only from prefetching current bucket("batch 1" in performance table).
Following your comment, I have implemented the solution you suggested(indeed, very easy approach) and we can see the results here:
```
By conducting keys command on 64cores Graviton 3 Amazon EC2 instance with 50 mil keys in size of 100 bytes each.
+--------------------+------------------+-----------------------------+
| Implementation     | Time (seconds)   | Keys Processed per Second   |
+--------------------+------------------+-----------------------------+
| Iterator without   | 11.112279        | 4,499,529                   |
|    prefetching     |                  |                             |
| Viktor's approach  | 3.712602         | 13,467,954                  |
| batch iterator     | 3.336432         | 14,985,700                  |
+--------------------+------------------+-----------------------------+
```
***Batch iterator processes keys about 11% faster than Viktor's approach.***
***SAVE command time is the same for batch iterator and Viktor's approach.***

src/hashtable.c
NadavGigi: Its not a possible scenario. `iter->num_of_active_buckets` gets the value of `iter->num_of_bucket` during iterator init and then only decremented while we never changing `iter->num_of_bucket`.

src/hashtable.c
NadavGigi: Done.

zuiderkwast:
> _**Batch iterator processes keys about 11% faster than Viktor's approach.**_
> _**SAVE command time is the same for batch iterator and Viktor's approach.**_

Thanks for implementing and comparing this! From your table we can see that "my" approach is 3.0 times more throughput than without prefetching, compared to 3.33 times for your Nadav's approch.

Given the performance vs complexity, I'm leaning towards "my" simpler approach. Since you've already done the implementation, can you post it in a separate draft PR so we can see it?

Maybe the simple approach can be optimized a bit more, without too much increased complexity. (For example, prefetching the next 2-3 top-level buckets so we don't slow when there's an empty bucket.)

NadavGigi:
> > _**Batch iterator processes keys about 11% faster than Viktor's approach.**_
> > _**SAVE command time is the same for batch iterator and Viktor's approach.**_
> 
> Thanks for implementing and comparing this! From your table we can see that "my" approach is 3.0 times more throughput than without prefetching, compared to 3.33 times for your Nadav's approch.
> 
> Given the performance vs complexity, I'm leaning towards "my" simpler approach. Since you've already done the implementation, can you post it in a separate draft PR so we can see it?
> 
> Maybe the simple approach can be optimized a bit more, without too much increased complexity. (For example, prefetching the next 2-3 top-level buckets so we don't slow when there's an empty bucket.)

@zuiderkwast 
The implementation has been refined and now yields even better results. It's very similar to your approach but with some optimizations.

src/hashtable.c
```diff
@@ -934,6 +934,32 @@ static inline incrementalFind *incrementalFindFromOpaque(hashtableIncrementalFin
...
+        valkey_prefetch(bucketNext(b));
+    }
+    for (int pos = 0; pos < numBucketPositions(b); pos++) {
+        if (isPositionFilled(b, pos)) {
+            valkey_prefetch(b->entries[pos]);
+        }
+    }
+}
+
+static void prefetchNextFilledBucket(iter *iter, bucket *current_bucket) {
```
zuiderkwast: Please add a short comment above this and the other function.

For readers, it's useful to understand which memory is already in the cache (or assumed to be in the cache) and what is being prefetched. It makes it easier for readers to understand the optimization. Here, for example, it's assumed that `current_bucket` is already in the cache, right?

src/hashtable.c
```diff
@@ -934,6 +934,32 @@ static inline incrementalFind *incrementalFindFromOpaque(hashtableIncrementalFin
...
+    if (current_bucket->chained) {
+        prefetchBucketEntries(bucketNext(current_bucket));
```
zuiderkwast: Shouldn't we prefetch the next bucket (e.g. `valkey_prefetch(bucketNext(current_bucket))`) some time before we actually access the next bucket? The `prefetchBucketEntries` function accesses the memory of the bucket so it's a cold memory read if it hasn't been prefetched, isn't it?

src/hashtable.c
```diff
@@ -934,6 +934,32 @@ static inline incrementalFind *incrementalFindFromOpaque(hashtableIncrementalFin
...
+    } else {
+        size_t table_size = numBuckets(iter->hashtable->bucket_exp[iter->table]);
+        for (size_t bucket_index = (size_t)iter->index + 1; bucket_index < table_size; bucket_index++) {
+            bucket *b = &iter->hashtable->tables[iter->table][bucket_index];
+            if (b->presence) {
+                prefetchBucketEntries(b);
```
zuiderkwast: The bucket at `b` has not yet been prefetched, so when we access `b->presence`, it is a cold memory lookup, isn't it? Or am I missing something?

Ideally, we should do `valkey_prefetch(b)` some time before we do `b->presence` and `prefetchBucketEntries(b)`.

zuiderkwast (commented):
Awesome! This implementation is much simpler and it's impressive that it's just as fast as the other one.

It seems that it could be optimized even a bit more to avoid a few more cache misses, or maybe am I missing something?

In general, the basic idea for prefetching is to only follow links one step at a time, with some time between each step, to make sure that we only access memory that has already been prefetched. I guess we can do this where we do `iter->index++`:

* Prefetch the bucket at index + 2 (just `valkey_prefetch(b)`)
* Prefetch the top-level elements and child-bucket pointer at index + 1. (The bucket itself has been prefetched the last time we incremented the index.)
* Prefetch the first level child-bucket elements at the current index. We start returning elements at the current top-level bucket elements that are already prefetched.

This makes sure that when we do index++ the next time, we only access memory that has already been previously prefetched. Right?

To also support long chains of child-buckets, we can prefetch the sub-child-bucket elements when we start returning elements from the child-bucket, and so on.

If this gets too complex and doesn't give any better performance, then ignore it.

src/hashtable.c
NadavGigi: The key improvement here is leveraging the CPU's automatic prefetching capabilities for contiguous memory:

1. For non-chained buckets in the table, we rely on the CPU to prefetch the empty buckets automatically, as they're in a contiguous block of memory.
2. We only explicitly prefetch the 'next of the next' bucket, not its entries.
3. For chained buckets, we still handle prefetching explicitly, as they may not be contiguous in memory.

src/hashtable.c
NadavGigi: Done

src/hashtable.c
NadavGigi: Done.
Now we access only prefetched buckets.

src/hashtable.c
```diff
@@ -934,6 +934,48 @@ static inline incrementalFind *incrementalFindFromOpaque(hashtableIncrementalFin
...
+    for (int pos = 0; pos < numBucketPositions(b); pos++) {
+        if (isPositionFilled(b, pos)) {
+            valkey_prefetch(b->entries[pos]);
+        }
+    }
+}
+
+static bucket *getNextBucket(bucket *current_bucket, size_t bucket_index, hashtable *ht, int table_index) {
+    if (current_bucket->chained) {
+        return bucketNext(current_bucket);
```
zuiderkwast: This looks good. I see the names `getNextBucket` an `bucketNext` are very similar, but they are different. Maybe we can rename `bucketNext` to `getChildBucket` or something, to better distinguish child bucket from next bucket-index?

Also a comment for each function is always good. For this one, just mention that it returns the child bucket if any, otherwise the bucket at the next bucket index.

NadavGigi: Done.

zuiderkwast: Not done. I think you didn't read my comment properly. :)

I suggested to rename the other one, the one that returns the child bucket or null.

zuiderkwast: Or something else, but to give them names that explain what they do and also make it clear that they are different and do different things.

NadavGigi: Im sorry for the misunderstanding.
I have changed the name from `bucketNext` to `getChildBucket`.

zuiderkwast (approved):
This looks great to me now.

Can you run the benchmark again after the last changes?

NadavGigi:
> This looks great to me now.
> 
> Can you run the benchmark again after the last changes?

After running the benchmarks, I can confirm that the results remain the same as those presented in the table from the first comment.

PR #1502, opened 2025-01-02
Replace dict with new hashtable: hash datatype

### LINKED ISSUE #1095: Use new hashtable for hashes
The new hashtable doesn't have a dict entry, so we need to invent one to hold field and value.

Possibly, we can embed field or value or both. The method used earlier is the function `sdscopytobuffer` and a byte to store the size of the sds header.

A field name never changes so it seems safe to always embed the field name, while the value can change and it can be very large so we can allow it to be non-embedded. This is very similar to string keys with the EMBSTR vs RAW encoding.

Idea of layout with both embedded:

    +-----------------------------+-----------------------------+
    | field                       | value                       |
    | hdr-size | hdr | content \0 | hdr-size | hdr | content \0 |
    +-----------------------------+-----------------------------+

Idea of layout with field embedded, value as pointer, stored first to make it aligned. (#1502 simply uses this representation for all field-value pairs.)

    +-------+-----------------------------+
    | value | field                       |
    | ptr   | hdr-size | hdr | content \0 |
    +-------+-----------------------------+

To distinguish between different representations, we could use the LSB bits in the pointer to it, the pointer stored in the hashtable. (This solution was not chosen in #1579.)

     hashtable
    +----------+        +-----------+----------------+
    | ptr  000 -------->| value-ptr | embedded field |
    |          |        +-----------+----------------+
    |          |
    |          |        +----------------+----------------+
    | ptr  001 -------->| embedded field | embedded value |
    |          |        +----------------+----------------+
    | ...      |
    +----------+

Another way to indicate different kinds of entries is to use some bits in the sds header of the field. That's the chosen representation in #1579.

rainsupreme:
This PR replaces dict with the new hashtable data structure in the HASH datatype. There is a new struct for hashtable items which contains a pointer to value sds string and the embedded key sds string. These values were previously stored in dictEntry. This structure is kept opaque so we can easily add small value embedding or other optimizations in the future.

closes #1095

### DIFF
--- a/src/defrag.c
+++ b/src/defrag.c
@@ -373,13 +373,6 @@ void activeDefragSdsHashtableCallback(void *privdata, void *entry_ref) {
     if (new_sds != NULL) *sds_ref = new_sds;
 }
 
-void activeDefragSdsHashtable(hashtable *ht) {
-    unsigned long cursor = 0;
-    do {
-        cursor = hashtableScanDefrag(ht, cursor, activeDefragSdsHashtableCallback, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
-    } while (cursor != 0);
-}
-
 /* Defrag a list of ptr, sds or robj string values */
 static void activeDefragQuickListNode(quicklist *ql, quicklistNode **node_ref) {
     quicklistNode *newnode, *node = *node_ref;
@@ -481,26 +474,29 @@ static void scanHashtableCallbackCountScanned(void *privdata, void *elemref) {
     server.stat_active_defrag_scanned++;
 }
 
-/* Used as dict scan callback when all the work is done in the dictDefragFunctions. */
-static void scanCallbackCountScanned(void *privdata, const dictEntry *de) {
-    UNUSED(privdata);
-    UNUSED(de);
-    server.stat_active_defrag_scanned++;
-}
-
 static void scanLaterSet(robj *ob, unsigned long *cursor) {
     if (ob->type != OBJ_SET || ob->encoding != OBJ_ENCODING_HASHTABLE) return;
     hashtable *ht = ob->ptr;
     *cursor = hashtableScanDefrag(ht, *cursor, activeDefragSdsHashtableCallback, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
 }
 
+static void activeDefragHashField(void *privdata, void *element_ref) {
+    UNUSED(privdata);
+    hashTypeEntry **entry_ref = (hashTypeEntry **)element_ref;
+
+    /* defragment field */
+    hashTypeEntry *new_entry = activeDefragAlloc(*entry_ref);
+    if (new_entry) *entry_ref = new_entry;
+
+    /* defragment value */
+    sds new_value = activeDefragSds((*entry_ref)->value);
+    if (new_value) (*entry_ref)->value = new_value;
+}
+
 static void scanLaterHash(robj *ob, unsigned long *cursor) {
-    if (ob->type != OBJ_HASH || ob->encoding != OBJ_ENCODING_HT) return;
-    dict *d = ob->ptr;
-    dictDefragFunctions defragfns = {.defragAlloc = activeDefragAlloc,
-                                     .defragKey = (dictDefragAllocFunction *)activeDefragSds,
-                                     .defragVal = (dictDefragAllocFunction *)activeDefragSds};
-    *cursor = dictScanDefrag(d, *cursor, scanCallbackCountScanned, &defragfns, NULL);
+    if (ob->type != OBJ_HASH || ob->encoding != OBJ_ENCODING_HASHTABLE) return;
+    hashtable *ht = ob->ptr;
+    *cursor = hashtableScanDefrag(ht, *cursor, activeDefragHashField, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
 }
 
 static void defragQuicklist(robj *ob) {
@@ -538,15 +534,19 @@ static void defragZsetSkiplist(robj *ob) {
 }
 
 static void defragHash(robj *ob) {
-    dict *d, *newd;
-    serverAssert(ob->type == OBJ_HASH && ob->encoding == OBJ_ENCODING_HT);
-    d = ob->ptr;
-    if (dictSize(d) > server.active_defrag_max_scan_fields)
+    serverAssert(ob->type == OBJ_HASH && ob->encoding == OBJ_ENCODING_HASHTABLE);
+    hashtable *ht = ob->ptr;
+    if (hashtableSize(ht) > server.active_defrag_max_scan_fields) {
         defragLater(ob);
-    else
-        activeDefragSdsDict(d, DEFRAG_SDS_DICT_VAL_IS_SDS);
-    /* defrag the dict struct and tables */
-    if ((newd = dictDefragTables(ob->ptr))) ob->ptr = newd;
+    } else {
+        unsigned long cursor = 0;
+        do {
+            cursor = hashtableScanDefrag(ht, cursor, activeDefragHashField, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
+        } while (cursor != 0);
+    }
+    /* defrag the hashtable struct and tables */
+    hashtable *new_hashtable = hashtableDefragTables(ht, activeDefragAlloc);
+    if (new_hashtable) ob->ptr = new_hashtable;
 }
 
 static void defragSet(robj *ob) {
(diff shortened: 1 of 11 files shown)


src/server.h
```diff
@@ -3420,6 +3420,15 @@ robj *setTypeDup(robj *o);
...
 #define HASH_SET_COPY 0
 
+typedef struct {
+    sds value;
+    unsigned char field_offset;
+    char field_data[];
+} hashTypeEntry;
+hashTypeEntry *hashTypeCreateEntry(sds field, sds value);
+const void *hashTypeEntryGetKey(const void *entry);
+void freeHashTypeEntry(void *entry);
```
zuiderkwast: Please make this one opaque and put it in `t_hash.c`.

Any accesses to `value` should use a function. Otherwise, we end up with lots of direct accesses to the value field and it it becomes very hard to implement embedded value in the future. (I believe embedded field+value can save considerable memory and CPU when it fits in one cache line, which should be a common case.)

src/defrag.c
```diff
@@ -480,26 +473,29 @@ static void scanHashtableCallbackCountScanned(void *privdata, void *elemref) {
...
-    server.stat_active_defrag_scanned++;
-}
-
 static void scanLaterSet(robj *ob, unsigned long *cursor) {
     if (ob->type != OBJ_SET || ob->encoding != OBJ_ENCODING_HASHTABLE) return;
     hashtable *ht = ob->ptr;
     *cursor = hashtableScanDefrag(ht, *cursor, activeDefragSdsHashtableCallback, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
 }
 
+static void activeDefragHashField(void *privdata, void *element_ref) {
```
zuiderkwast: If hashTypeEntry is made opaque, we need to move this function (or parts of it) to t_hash.c. We can just pass the activeDefragAlloc an argument to the function, similar to how we're defragging the hashtable internals.

zuiderkwast (commented):
I skimmed through it. I looks strait-forward. The only material thing to focus on seems to be the hashTypeEntry abstraction.

It would be good to have a clear abstraction and separation, even within t_hash.c, so we don't have too many direct accesses into the hashTypeEntry internals. It will help us later when we want to add things like TTL and embedded value.

src/t_hash.c
```diff
@@ -30,6 +30,65 @@
...
 #include <math.h>
 
+struct hashTypeEntry {
+    sds value;
+    unsigned char field_offset;
+    unsigned char field_data[];
+};
+
+/* takes ownership of value, does not take ownership of field */
+hashTypeEntry *hashTypeCreateEntry(const sds field, sds value) {
```
zuiderkwast: * `const sds` is meaningless. Is it time to introduce the `const_sds` typedef in sds.h now?

* Take ownership of field but not value seems asymmetrical, but anything else is suboptimal, right? That's what we do for keys too IIRC. It's OK.

* Shall we add a comment for the entry API, like the one below for the hash type API?

```
/*-----------------------------------------------------------------------------
 * Hash entry API
 *----------------------------------------------------------------------------*/
```

src/t_hash.c
```diff
@@ -547,22 +592,24 @@ void hashReplyFromListpackEntry(client *c, listpackEntry *e) {
...
+        hashtableFairRandomEntry(hashobj->ptr, &entry);
+        sds sds_field = hashTypeEntryGetField(entry);
+        field->sval = (unsigned char *)sds_field;
+        field->slen = sdslen(sds_field);
         if (val) {
-            sds s = dictGetVal(de);
-            val->sval = (unsigned char *)s;
-            val->slen = sdslen(s);
+            hashTypeEntry *hash_entry = entry;
+            sds sds_val = hashTypeEntryGetValue(hash_entry);
```
zuiderkwast: The entry can be cast implicitly. You did that for `sds sds_field = hashTypeEntryGetField(entry);` a few lines above.

```suggestion
            sds sds_val = hashTypeEntryGetValue(entry);
```

src/defrag.c
```diff
@@ -481,26 +474,24 @@ static void scanHashtableCallbackCountScanned(void *privdata, void *elemref) {
...
-    server.stat_active_defrag_scanned++;
-}
-
 static void scanLaterSet(robj *ob, unsigned long *cursor) {
     if (ob->type != OBJ_SET || ob->encoding != OBJ_ENCODING_HASHTABLE) return;
     hashtable *ht = ob->ptr;
     *cursor = hashtableScanDefrag(ht, *cursor, activeDefragSdsHashtableCallback, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
 }
 
+static void activeDefragHashTypeEntry(void *privdata, void *element_ref) {
```
zuiderkwast, line 484: Please add a comment saying it's used as a scan callback.

src/t_hash.c
```diff
@@ -30,6 +30,65 @@
...
+    hashTypeEntry *entry = zmalloc(total_size);
+
+    entry->value = value;
+    sdscopytobuffer(entry->field_data, field_size, field, &entry->field_offset);
+    return entry;
+}
+
+sds hashTypeEntryGetField(const hashTypeEntry *entry) {
+    const unsigned char *field = entry->field_data + entry->field_offset;
+    return (sds)field;
```
zuiderkwast, line 57: This (and the GetValue) should probably return `const_sds` if we add that type.

src/t_hash.c
```diff
@@ -30,6 +30,65 @@
...
+    entry->value = value;
+}
+
+size_t hashTypeEntryAllocSize(hashTypeEntry *entry) {
+    size_t size = zmalloc_usable_size(entry);
+    size += sdsAllocSize(entry->value);
+    return size;
+}
+
+hashTypeEntry *hashTypeEntryDefrag(hashTypeEntry *entry, void *(*defragfn)(void *), sds (*sdsdefragfn)(sds)) {
```
zuiderkwast, line 85: Please add a short comment above each of the functions, at least the non-trivial ones like this one.

src/t_hash.c
```diff
@@ -228,28 +284,27 @@ int hashTypeSet(robj *o, sds field, sds value, int flags) {
...
+        void *existing;
+        if (hashtableFindPositionForInsert(ht, field, &position, &existing)) {
+            /* does not exist yet */
+            hashTypeEntry *entry = hashTypeCreateEntry(field, v);
+            hashtableInsertAtPosition(ht, entry, &position);
         } else {
-            sdsfree(dictGetVal(existing));
-            dictSetVal(ht, existing, v);
+            /* exists: replace value */
+            hashTypeEntryReplaceValue(existing, v);
```
zuiderkwast, line 322: I think `hashTypeEntryReplaceValue` should return the entry, so we don't have to change the signature in the future. With embedded value, this function may need to reallocate the entry.

For now, we can just assert that it returns `existing`.

```suggestion
            serverAssert(hashTypeEntryReplaceValue(existing, v) == existing);
```

src/t_hash.c
```diff
@@ -1024,11 +1071,11 @@ void hrandfieldWithCountCommand(client *c, long l, int withvalues) {
...
         while ((hashTypeNext(&hi)) != C_ERR) {
             int ret = DICT_ERR;
-            sds key, value = NULL;
+            sds field, value = NULL;
 
-            key = hashTypeCurrentObjectNewSds(&hi, OBJ_HASH_KEY);
+            field = hashTypeCurrentObjectNewSds(&hi, OBJ_HASH_FIELD);
             if (withvalues) value = hashTypeCurrentObjectNewSds(&hi, OBJ_HASH_VALUE);
-            ret = dictAdd(d, key, value);
+            ret = dictAdd(d, field, value);
```
zuiderkwast, line 1092: Follow-up: Let's replace this sdsReplyDictType with a hashtable.

src/t_hash.c
```diff
@@ -1069,25 +1116,25 @@ void hrandfieldWithCountCommand(client *c, long l, int withvalues) {
     else {
         /* Hashtable encoding (generic implementation) */
         unsigned long added = 0;
-        listpackEntry key, value;
+        listpackEntry field, value;
         dict *d = dictCreate(&hashDictType);
```
zuiderkwast, line 1134: Follow-up: Let's replace this dict with a hashtable. Then we can delete that dict type.

zuiderkwast (commented):
Awesome! Just a few nits.

Introducing the `sds_const` type can preferably be a follow-up. It might trigger a long discussion.

src/t_hash.c
rainsupreme: I agree with you, let's not introduce const_sds in this change. I'll remove the const here - it does not mean `const char*` as I intended when I wrote it.

I'll add the comment header, that's a good idea :)

src/t_hash.c
rainsupreme: It's only called in one place, and it's a static function - it'll be quite easy to change later. Do you think we'll have some kind of trouble from leaving it simpler like this until embedded values are implemented?

zuiderkwast: OK, leave it. We can refactor it when we need it.

zuiderkwast (approved):
LGTM

PR #1506, opened 2025-01-04
Improve readability and efficiency for ZUNION operation

QuChen88:
Currently zunion operation in `t_zset.c` is using `dictAddRaw()` to search for the element in the accumulating dictionary. If the element was not found and got added, it would then set the key to a new key string via `dictSetKey()`, which is counter intuitive and not very effective.

This PR improves the code readability and effectiveness by using `dictFind()` to search for the element in the dictionary, and then uses `dictAddRaw()` and `dictSetDoubleVal()` to create the element in the set, and avoids the need to call `dictSetKey()`. 

Note - this change is consistent with the original implementation of sorted set https://github.com/valkey-io/valkey/blob/d680eb6dbdf2d2030cb96edfb089be1e2a775ac1/src/t_zset.c#L2286. It was later changed to a "not-so-nice" style in commit https://github.com/valkey-io/valkey/commit/68bf45fa1e77c7e2e3f3ab4a7894f9bc1a1da57b

### DIFF
--- a/src/t_zset.c
+++ b/src/t_zset.c
@@ -2793,7 +2793,7 @@ void zunionInterDiffGenericCommand(client *c, robj *dstkey, int numkeysIndex, in
                 if (isnan(score)) score = 0;
 
                 /* Search for this element in the accumulating dictionary. */
-                de = dictAddRaw(dstzset->dict, zuiSdsFromValue(&zval), &existing);
+                existing = dictFind(dstzset->dict, zuiSdsFromValue(&zval));
                 /* If we don't have it, we need to create a new entry. */
                 if (!existing) {
                     tmp = zuiNewSdsFromValue(&zval);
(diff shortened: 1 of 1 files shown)


ranshid:
Thank you @QuChen88 I think the zset code is about to change a lot following https://github.com/valkey-io/valkey/pull/1427. maybe you want to take a look and propose alternatives there?

JimB123:
I think the reason for the original code is that in a single operation, it searches and adds.  With the new code, it will need to perform a hash table lookup for the find, and then another hash table lookup for the add.

QuChen88:
My biggest issue was that the insertion into the dict happened outside of the `if (existing)` check previously. 

The new code in the proposed PR https://github.com/valkey-io/valkey/pull/1427 looks correct to me. It first finds the element in the hashset and then if it is not present will add the element.

zuiderkwast:
> The new code in the proposed PR #1427 looks correct to me. It first finds the element in the hashset and then if it is not present will add the element.

Great, so let's close this PR? #1427 will be merged very soon.

PR #1511, opened 2025-01-06
[feature] sentinel use info sentinel command to run faster

kukey:
In sentinel cluster, sentinels will send info command to collect valkey-server state;
but in large cluster, the info command response is heavy and slow, so we need some 
simple info response for sentinel.

#### valkey-server:
  * add info-simple-for-sentinel flag,  to distinguish the valkey-server is support simple info;
  * info command add sentinel subcommand, to cover sentinel used info

#### sentinel:
  * sentinelValkeyInstance struct add info_simple flag;
  * sentinelSendPeriodicCommands will check info_simple to send diff comamnd;
  * reconnect valkey instacne, reset info_simple flag;
  * handle info response from server to parse info_simple flag

### performence test:
  * info comamnd:
   ```valkey-benchmark -q -n 1000000 info```
  info: 65841.45 requests per second, p50=0.703 msec
  * info sentinel:
```valkey-benchmark -q -n 1000000 info sentinel```
info sentinel: 236910.67 requests per second, p50=0.111 msec

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -3382,6 +3382,9 @@ standardConfig static_configs[] = {
     createSpecialConfig("replicaof", "slaveof", IMMUTABLE_CONFIG | MULTI_ARG_CONFIG, setConfigReplicaOfOption, getConfigReplicaOfOption, rewriteConfigReplicaOfOption, NULL),
     createSpecialConfig("latency-tracking-info-percentiles", NULL, MODIFIABLE_CONFIG | MULTI_ARG_CONFIG, setConfigLatencyTrackingInfoPercentilesOutputOption, getConfigLatencyTrackingInfoPercentilesOutputOption, rewriteConfigLatencyTrackingInfoPercentilesOutputOption, NULL),
 
+    /* Capabalities */
+    createBoolConfig("info-simple-for-sentinel", NULL, IMMUTABLE_CONFIG, server.info_simple_for_sentinel, 1, NULL, NULL),
+
     /* NULL Terminator, this is dropped when we convert to the runtime array. */
     {NULL},
 };
--- a/src/sentinel.c
+++ b/src/sentinel.c
@@ -248,7 +248,8 @@ typedef struct sentinelValkeyInstance {
      * are set to NULL no script is executed. */
     char *notification_script;
     char *client_reconfig_script;
-    sds info; /* cached INFO output */
+    sds info;        /* cached INFO output */
+    int info_simple; /* Instance support info simple for sentinel. */
 } sentinelValkeyInstance;
 
 /* Main state. */
@@ -1346,6 +1347,9 @@ sentinelValkeyInstance *createSentinelValkeyInstance(char *name,
     ri->role_reported_time = mstime();
     ri->replica_conf_change_time = mstime();
 
+    /* capability */
+    ri->info_simple = 0;
+
     /* Add into the right table. */
     dictAdd(table, ri->name, ri);
     return ri;
@@ -2358,6 +2362,8 @@ void sentinelReconnectInstance(sentinelValkeyInstance *ri) {
 
             /* Send a PING ASAP when reconnecting. */
             sentinelSendPing(ri);
+
+            ri->info_simple = 0;
         }
     }
     /* Pub / Sub */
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Review by li-benson — APPROVED** [review]

File: src/config.c [hunk 1]
```diff
@@ -3373,6 +3373,9 @@ standardConfig static_configs[] = {
     createSpecialConfig("replicaof", "slaveof", IMMUTABLE_CONFIG | MULTI_ARG_CONFIG, setConfigReplicaOfOption, getConfigReplicaOfOption, rewriteConfigReplicaOfOption, NULL),
     createSpecialConfig("latency-tracking-info-percentiles", NULL, MODIFIABLE_CONFIG | MULTI_ARG_CONFIG, setConfigLatencyTrackingInfoPercentilesOutputOption, getConfigLatencyTrackingInfoPercentilesOutputOption, rewriteConfigLatencyTrackingInfoPercentilesOutputOption, NULL),
 
+    /* Capabalities */
+    createBoolConfig("info-simple-for-sentinel", NULL, IMMUTABLE_CONFIG, server.info_simple_for_sentinel, 1, NULL, NULL),
```
hwware, line 3386: Is this config only for sentinel? If yes, I think it should be moved to sentinel.c,  we have some specific config parameters for sentinel node.

kukey:
@hwware yes，this config only for sentinel, but it`s mean valkey has a certain ability that sentinel could send diff command to collect instance stats.
It`s a server capability. I think it can`t remove to sentinel.c.

2025年1月8日 03:47，Wen Hui ***@***.***> 写道：



@hwware commented on this pull request.

________________________________

In src/config.c<https://github.com/valkey-io/valkey/pull/1511#discussion_r1905963404>:

> @@ -3373,6 +3373,9 @@ standardConfig static_configs[] = {
     createSpecialConfig("replicaof", "slaveof", IMMUTABLE_CONFIG | MULTI_ARG_CONFIG, setConfigReplicaOfOption, getConfigReplicaOfOption, rewriteConfigReplicaOfOption, NULL),
     createSpecialConfig("latency-tracking-info-percentiles", NULL, MODIFIABLE_CONFIG | MULTI_ARG_CONFIG, setConfigLatencyTrackingInfoPercentilesOutputOption, getConfigLatencyTrackingInfoPercentilesOutputOption, rewriteConfigLatencyTrackingInfoPercentilesOutputOption, NULL),

+    /* Capabalities */
+    createBoolConfig("info-simple-for-sentinel", NULL, IMMUTABLE_CONFIG, server.info_simple_for_sentinel, 1, NULL, NULL),


Is this config only for sentinel? If yes, I think it should be moved to sentinel.c, we have some specific config parameters for sentinel node.

gmbnomis:
I am wondering why you consider the performance of the INFO command relevant in this case  (or to be more precise what you mean by "large cluster"):

Sentinels do not scale by sharding, all members of the sentinel cluster manage all primaries. The reason to have multiple sentinel instances is to increase robustness. By default, each sentinel instance sends an INFO command every 10 seconds to each monitored Valkey node. (And IIRC, one can change that timing only by using a debug command)

So, even if we assume that we have 9 Sentinels (I assume most deployment use 3 or 5), this means that each Valkey node receives around one INFO command per second. For example:

```
$ valkey-cli -p 30000 info commandstats | grep cmdstat_info ; sleep 60 ; valkey-cli -p 30000 info commandstats | grep cmdstat_info

cmdstat_info:calls=123,usec=25555,usec_per_call=207.76,rejected_calls=0,failed_calls=0
cmdstat_info:calls=178,usec=34934,usec_per_call=196.26,rejected_calls=0,failed_calls=0
```

Which are 54 INFO commands per minute. This means that the performance of the INFO command is completely negligible. 

@kukey  Could you explain in which scenario the performance improvement achieved by this PR becomes relevant?

kukey:
@gmbnomis We use 3-6 sentinel to manage thousands of valley isntance, if sentinel use this simple info, could save more cpu and bandwidth; and to the client and server, it`s more stable and smoother latency

src/config.c
kukey: yes，this config only for sentinel, but it's mean valkey has a certain ability that sentinel could send diff command to collect instance stats.
It's a server capability. I think it can't move to sentinel.c.

gmbnomis:
> @gmbnomis We use 3-6 sentinel to manage thousands of valley isntance, if sentinel use this simple info, could save more cpu and bandwidth; and to the client and server, it`s more stable and smoother latency

I see. So, the performance improvements of the INFO command given in the PR description are not really the point of this PR. It is about sentinel performance.

I noticed that the  `master_link_down_since_seconds` field (usually, it is in the "replication" section) is not part of your proposal. However,  it is expected by sentinel. If this field is present and greater than zero, the info period for the respective node decreases from 10s to 1s, i.e. a tenfold increase in INFO commands.

I am wondering if the improvements you see may be caused (partly) by the omission of this field?

kukey:
@gmbnomis thanks, i miss the `master_link_down_since_seconds` field.

PR #1513, opened 2025-01-06
Fix LUA garbage collector (CVE-2024-46981)

madolson:
Reset GC state before closing the lua VM to prevent user data to be wrongly freed while still might be used on destructor callbacks.

Created and publish by Redis in their OSS branch.

### DIFF
--- a/src/eval.c
+++ b/src/eval.c
@@ -285,6 +285,7 @@ void scriptingInit(int setup) {
 void freeLuaScriptsSync(dict *lua_scripts, list *lua_scripts_lru_list, lua_State *lua) {
     dictRelease(lua_scripts);
     listRelease(lua_scripts_lru_list);
+    lua_gc(lctx.lua, LUA_GCCOLLECT, 0);
     lua_close(lua);
 
 #if !defined(USE_LIBC)


src/eval.c
```diff
@@ -285,6 +285,7 @@ void scriptingInit(int setup) {
 void freeLuaScriptsSync(dict *lua_scripts, list *lua_scripts_lru_list, lua_State *lua) {
     dictRelease(lua_scripts);
     listRelease(lua_scripts_lru_list);
+    lua_gc(lctx.lua, LUA_GCCOLLECT, 0);
```
hpatro, line 288: Do we need to perform a full garbage collection cycle? Should we just stop the GC via `LUA_GCSTOP` ?

madolson: I'll be honest, I do not understand the fix. I have been unable to reproduce the issue, and the CVE was not published yet so I don't know?

hpatro: Doing a `LUA_GCCOLLECT` should also be correct, however there could be a slight latency spike with that.

madolson: > Doing a LUA_GCCOLLECT should also be correct, however there could be a slight latency spike with that.

Yes. We applied this fix internally already because it was disclosed previously, and haven't seen much issue as of yet.

hpatro (approved):

PR #1514, opened 2025-01-06
Fix Read/Write key pattern selector (CVE-2024-51741)

madolson:
The explanation on the original commit was wrong. Key based access must have a `~` in order to correctly configure whey key prefixes to apply the selector to. If this is missing, a server assert will be triggered later.

### DIFF
--- a/tests/unit/acl-v2.tcl
+++ b/tests/unit/acl-v2.tcl
@@ -116,10 +116,31 @@ start_server {tags {"acl external:skip"}} {
         assert_match "*NOPERM*key*" $err
     }
 
-    test {Validate read and write permissions format} {
+    test {Validate read and write permissions format - empty permission} {
         catch {r ACL SETUSER key-permission-RW %~} err
         set err
     } {ERR Error in ACL SETUSER modifier '%~': Syntax error}
+
+    test {Validate read and write permissions format - empty selector} {
+        catch {r ACL SETUSER key-permission-RW %} err
+        set err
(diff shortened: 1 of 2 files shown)


tests/unit/acl-v2.tcl
```diff
@@ -116,11 +116,32 @@ start_server {tags {"acl external:skip"}} {
...
+        set err
+    } {NOPERM No permissions to access a key}
+
+    test {Validate read and write permissions format - no pattern} {
+        # No pattern results with R/W access to no key (currently we accept this syntax error)
+        r ACL SETUSER key-permission-RW on nopass %RW +@all
+        $r2 auth key-permission-RW password
+        catch {$r2 SET x 5} err
+        set err
+    } {NOPERM No permissions to access a key}
```
hpatro, line 143: Did a quick check locally both the ACL rules get simplified to the following i.e. `~` no key. I also think it's too late to error out on the no pattern with a selector. Possibly add more assertion to the test case to verify the simplified ACL rule on the server. (can be done later).
```
user key-permission-RW on nopass sanitize-payload ~ resetchannels +@all
```

hpatro (approved):

tests/unit/acl-v2.tcl
madolson: I'll make a note. I also think the comment is wrong, but was just pulling it verbatim to retain origin from the Redis BSD Stream.

madolson: https://github.com/valkey-io/valkey/pull/1515

PR #1517, opened 2025-01-07
Incr expired_keys if the expiration time is already expired

RayaCoo:
This PR modifies two parts of the `deleteExpiredKeyFromOverwriteAndPropagate` function, which is called when the expiration time has already expired.
* change Keyspace Event notify from `GENERIC-del` to `EXPIRED-expired`. 
Since this delete operation is triggered by EXPIRE, and the flag pass to `dbGenericDeleteWithDictIndex` is `DB_FLAG_KEY_EXPIRED`, which is used to notify module, maybe we should align the behavior and send Keyspace Event notify like `deleteExpiredKeyAndPropagateWithDictIndex` did.
* As i said befoce, this del operation is triggered by EXPIRE, maybe increasing `server.stat_expiredkeys` in this case could more accurately reflect the number of keys deleted due to expiration.

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -1860,7 +1860,8 @@ void deleteExpiredKeyFromOverwriteAndPropagate(client *c, robj *keyobj) {
     robj *aux = server.lazyfree_lazy_expire ? shared.unlink : shared.del;
     rewriteClientCommandVector(c, 2, aux, keyobj);
     signalModifiedKey(c, c->db, keyobj);
-    notifyKeyspaceEvent(NOTIFY_GENERIC, "del", keyobj, c->db->id);
+    notifyKeyspaceEvent(NOTIFY_EXPIRED, "expired", keyobj, c->db->id);
+    server.stat_expiredkeys++;
 }
 
 /* Propagate an implicit key deletion into replicas and the AOF file.
--- a/tests/unit/expire.tcl
+++ b/tests/unit/expire.tcl
@@ -142,6 +142,31 @@ start_server {tags {"expire"}} {
         list $e $f
     } {somevalue {}}
 
+    test {EXPIRE / EXPIREAT / PEXPIRE / PEXPIREAT Expiration time is already expired} {
+        set prev_expired [s expired_keys]
+        r del x
+
+        r set x somevalue
+        r expire x -1
+        assert_equal {0} [r exists x]
+        assert_equal {1} [expr {[s expired_keys] - $prev_expired}]
+
+        r set x somevalue
+        r expireat x [expr [clock seconds] - 1]
+        assert_equal {0} [r exists x]
+        assert_equal {2}  [expr {[s expired_keys] - $prev_expired}]
+
+        r set x somevalue
+        r pexpire x -1
+        assert_equal {0} [r exists x]
+        assert_equal {3}  [expr {[s expired_keys] - $prev_expired}]
+
+        r set x somevalue
+        r pexpireat x [expr [clock milliseconds] - 1]
+        assert_equal {0} [r exists x]
+        assert_equal {4}  [expr {[s expired_keys] - $prev_expired}]
+    }
+
     test {TTL returns time to live in seconds} {
         r del x
         r setex x 10 somevalue
--- a/tests/unit/expire.tcl
+++ b/tests/unit/expire.tcl
@@ -143,28 +143,28 @@ start_server {tags {"expire"}} {
     } {somevalue {}}
 
     test {EXPIRE / EXPIREAT / PEXPIRE / PEXPIREAT Expiration time is already expired} {
-        set prev_expired [s expired_keys]
-        r del x
+        r flushall
+        r config resetstat
 
         r set x somevalue
         r expire x -1
         assert_equal {0} [r exists x]
-        assert_equal {1} [expr {[s expired_keys] - $prev_expired}]
+        assert_equal {1} [s expired_keys]
 
         r set x somevalue
         r expireat x [expr [clock seconds] - 1]
         assert_equal {0} [r exists x]
-        assert_equal {2}  [expr {[s expired_keys] - $prev_expired}]
+        assert_equal {2}  [s expired_keys]
 
         r set x somevalue
-        r pexpire x -1
+        r pexpire x -1000
         assert_equal {0} [r exists x]
-        assert_equal {3}  [expr {[s expired_keys] - $prev_expired}]
+        assert_equal {3} [s expired_keys]
 
         r set x somevalue
-        r pexpireat x [expr [clock milliseconds] - 1]
+        r pexpireat x [expr [clock milliseconds] - 1000]
         assert_equal {0} [r exists x]
-        assert_equal {4}  [expr {[s expired_keys] - $prev_expired}]
+        assert_equal {4} [s expired_keys]
     }
 
     test {TTL returns time to live in seconds} {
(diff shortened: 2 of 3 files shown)


soloestoy:
The core of this issue is whether we should consider setting an expired time for a key as a delete operation or an expiration.

I believe it should be considered as expiration because the user's intended action is to set an expire time for the key. However, due to some reasons (such as operational delays), the time has already passed. Nonetheless, the key is still deleted due to expiration. Therefore, in terms of statistics and notifications, it should be counted as expiration-based deletion.

Additionally, we have already sent notifications regarding the expiration for the module:
```c
void deleteExpiredKeyFromOverwriteAndPropagate(client *c, robj *keyobj) {
    int deleted = dbGenericDelete(c->db, keyobj, server.lazyfree_lazy_expire, DB_FLAG_KEY_EXPIRED);

int dbGenericDelete(serverDb *db, robj *key, int async, int flags) {
    int dict_index = getKVStoreIndexForKey(key->ptr);
    return dbGenericDeleteWithDictIndex(db, key, async, flags, dict_index);
}

int dbGenericDeleteWithDictIndex(serverDb *db, robj *key, int async, int flags, int dict_index) {
    hashtablePosition pos;
    void **ref = kvstoreHashtableTwoPhasePopFindRef(db->keys, dict_index, key->ptr, &pos);
    if (ref != NULL) {
        robj *val = *ref;
        /* VM_StringDMA may call dbUnshareStringValue which may free val, so we
         * need to incr to retain val */
        incrRefCount(val);
        /* Tells the module that the key has been unlinked from the database. */
        moduleNotifyKeyUnlink(key, val, db->id, flags);

```

However, there is a change here: the notification has been changed from `GENERIC-del` to `EXPIRED-expired`. This might be a potential breaking change, but theoretically, the actual impact should be minimal. Typically, if there's a concern about a key being deleted, the generic, expired, and evicted events would all be subscribed to. @valkey-io/core-team PTAL

tests/unit/expire.tcl
```diff
@@ -142,6 +142,31 @@ start_server {tags {"expire"}} {
         list $e $f
     } {somevalue {}}
 
+    test {EXPIRE / EXPIREAT / PEXPIRE / PEXPIREAT Expiration time is already expired} {
+        set prev_expired [s expired_keys]
+        r del x
```
enjoy-binbin: i guess we can just do these, and then when doing the assert_equal, we can `assert_equal {1} [s expired_keys]`, so we dont need the extra prev_expired var. Doing the flushall also make sure there is no chance the other test keys leftover and make it flaky
```suggestion
        r flushall
        r config resetstat
```

tests/unit/expire.tcl
```diff
@@ -142,6 +142,31 @@ start_server {tags {"expire"}} {
...
+        assert_equal {0} [r exists x]
+        assert_equal {2}  [expr {[s expired_keys] - $prev_expired}]
+
+        r set x somevalue
+        r pexpire x -1
+        assert_equal {0} [r exists x]
+        assert_equal {3}  [expr {[s expired_keys] - $prev_expired}]
+
+        r set x somevalue
+        r pexpireat x [expr [clock milliseconds] - 1]
```
enjoy-binbin: lets also incr it to avoid the timing issue.
```suggestion
        r pexpireat x [expr [clock milliseconds] - 1000]
```

enjoy-binbin (approved):
overall make sense to me.

tests/unit/expire.tcl
RayCao99: Done

tests/unit/expire.tcl
RayCao99: Done

madolson:
> This might be a potential breaking change, but theoretically, the actual impact should be minimal.

Yeah, I think at the very least this needs to be marked as a breaking change, since we are changing the event type. Even if the new behavior is more correct, it's such a long standing behavior someone might have coded around it.

soloestoy (approved):

soloestoy:
since it's a breaking change (very small), @valkey-io/core-team please vote

src/db.c
```diff
@@ -1860,7 +1860,8 @@ void deleteExpiredKeyFromOverwriteAndPropagate(client *c, robj *keyobj) {
     robj *aux = server.lazyfree_lazy_expire ? shared.unlink : shared.del;
     rewriteClientCommandVector(c, 2, aux, keyobj);
     signalModifiedKey(c, c->db, keyobj);
-    notifyKeyspaceEvent(NOTIFY_GENERIC, "del", keyobj, c->db->id);
+    notifyKeyspaceEvent(NOTIFY_EXPIRED, "expired", keyobj, c->db->id);
```
zuiderkwast, line 1863: ~~What's the difference between `expire` and `expired` event?~~ I think the difference between "expire" and "expired" is clear in the docs.

In the docs I can find for "expired":

> * Every time a key with a time to live associated is removed from the data set because it expired, an expired event is generated.

and for "expire" event:

> * EXPIRE and all its variants (PEXPIRE, EXPIREAT, PEXPIREAT) generate an expire event when called with a positive timeout (or a future timestamp).  Note that when  these  com‐
         mands are called with a negative timeout value or timestamp in the past, the key is deleted and only a del event is generated instead.

This seems to me like a bug fix. A key expiration is supposed to trigger an `expired` event. A `del` event seems like a bug. Or am I missing anything?

soloestoy: you can find "expired" in active expire path: `activeExpireCycleTryExpire` -> `deleteExpiredKeyAndPropagate` -> `deleteExpiredKeyAndPropagateWithDictIndex`

zuiderkwast: Thanks. Yes I found it.

Shall we postpone the breaking change to 9.0?

soloestoy: i think 8.1 is ok, it's a very very small change maybe even not breaking.

zuiderkwast: If we do it in 8.1, we can describe it as a bug fix. It looks better than a breaking change in a minor version.

But let's be sure we know what we are doing. We have done some mistakes before such as #1537 and #1228. We often think that small things like that will not be a problem.

@madolson said

> Even if the new behavior is more correct, it's such a long standing behavior someone might have coded around it.

soloestoy: I'm very sure what it is : )

ranshid:
Added the client-changes-needed since it MIGHT impact some client code managing client side caching

hwware:
The change totally makes sense to me. Approved

src/db.c
zuiderkwast: > I'm very sure what it is : )

What is it exactly?

If it's a breaking change, we shouldn't release it in a minor version.

* Breaking change → release in 9.0.
* Not breaking change → release in 8.1 and remove the breaking-change label.

soloestoy: This is a bugfix, but it might also introduce a breaking change. These two are not mutually exclusive, and I don't understand why it has to be an either-or situation.

zuiderkwast: Because we use semantic versioning. It's the only reason.

Major version must be incremented if we introduce backward incompatible API changes.

If the public documented API is not changed, then we are just fixing incorrect behaviour. Then a patch release is enough.

https://semver.org

soloestoy: OK, this PR doesn't introduce API changes, I removed the breaking-change flag, and 8.1 is the right way

mcollina:
This should not impact iovalkey at all.

PR #1519, opened 2025-01-07
add paused_actions for INFO Clients

soloestoy:
Add `paused_actions` and `paused_timeout_milliseconds` for INFO Clients to inform users about if clients are paused.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -5941,7 +5941,7 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "instantaneous_eventloop_cycles_per_sec:%llu\r\n", getInstantaneousMetric(STATS_METRIC_EL_CYCLE),
                 "instantaneous_eventloop_duration_usec:%llu\r\n", getInstantaneousMetric(STATS_METRIC_EL_DURATION),
                 "paused_actions:%s\r\n", paused_actions,
-                "paused_timeout:%lld\r\n", paused_timeout));
+                "paused_timeout_milliseconds:%lld\r\n", paused_timeout));
         info = genValkeyInfoStringACLStats(info);
     }
 
--- a/src/server.c
+++ b/src/server.c
@@ -5649,6 +5649,17 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
         unsigned long blocking_keys, blocking_keys_on_nokey, watched_keys;
         getExpansiveClientsInfo(&maxin, &maxout);
         totalNumberOfStatefulKeys(&blocking_keys, &blocking_keys_on_nokey, &watched_keys);
+
+        char *paused_actions = "none";
+        long long paused_timeout = 0;
+        if (server.paused_actions & PAUSE_ACTION_CLIENT_ALL) {
+            paused_actions = "all";
+            paused_timeout = getPausedActionTimeout(PAUSE_ACTION_CLIENT_ALL);
+        } else if (server.paused_actions & PAUSE_ACTION_CLIENT_WRITE) {
+            paused_actions = "write";
+            paused_timeout = getPausedActionTimeout(PAUSE_ACTION_CLIENT_WRITE);
+        }
+
         if (sections++) info = sdscat(info, "\r\n");
         info = sdscatprintf(
             info,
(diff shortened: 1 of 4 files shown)


src/server.c
```diff
@@ -5863,6 +5863,8 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
             server.stat_last_eviction_exceeded_time ? (long long)elapsedUs(server.stat_last_eviction_exceeded_time) : 0;
         long long current_active_defrag_time =
             server.stat_last_active_defrag_time ? (long long)elapsedUs(server.stat_last_active_defrag_time) : 0;
+        char *paused_actions = server.paused_actions & PAUSE_ACTION_CLIENT_ALL ? "all" : server.paused_actions & PAUSE_ACTION_CLIENT_WRITE ? "write"
```
madolson: Paused actions will never return "all" right? If it's all, it won't execute the command running "all". It's fine, but just worth mentioning this will only show write or none.

madolson (commented):
I think this is a good info field to add.

madolson (approved):
Just fix the format error and it's good for me.

tests/unit/pause.tcl
```diff
@@ -1,4 +1,19 @@
 start_server {tags {"pause network"}} {
+    test "Test check paused_actions in info stats" {
+        assert_equal [s paused_actions] "none"
+
+        r client PAUSE 10000 WRITE
+        assert_equal [s paused_actions] "write"
+        r client unpause
+
+        r multi
+        r client PAUSE 1000 All
```
hpatro: nit pick: I find it weird seeing it in mixed case :| . Could we write it in lower or upper case?

```suggestion
        r client PAUSE 1000 ALL
```

src/server.c
```diff
@@ -5930,7 +5932,8 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "eventloop_duration_sum:%llu\r\n", server.duration_stats[EL_DURATION_TYPE_EL].sum,
                 "eventloop_duration_cmd_sum:%llu\r\n", server.duration_stats[EL_DURATION_TYPE_CMD].sum,
                 "instantaneous_eventloop_cycles_per_sec:%llu\r\n", getInstantaneousMetric(STATS_METRIC_EL_CYCLE),
-                "instantaneous_eventloop_duration_usec:%llu\r\n", getInstantaneousMetric(STATS_METRIC_EL_DURATION)));
+                "instantaneous_eventloop_duration_usec:%llu\r\n", getInstantaneousMetric(STATS_METRIC_EL_DURATION),
+                "paused_actions:%s\r\n", paused_actions));
```
artikell: The end time of the paused action is also a key information. Is this considered for inclusion?

madolson: It makes sense to me, I would be okay adding it.

hwware:
I am confused why this pr is tagged as major-decision-pending?

madolson:
@hwware New info fields. I thought I posted a message, but maybe I forgot. @valkey-io/core-team I suppose the major decision is if people are okay with the new info fields listed. (Maybe also including the one @artikell suggested).

zuiderkwast:
This field will be "none" most of the time, right? If we add many fields like this, is the size of the INFO reply a concern?

I remember for the graceful shutdown feature, we added a field that is only present during shutdown:

> * **shutdown_in_milliseconds**: The maximum time remaining for replicas to catch up the replication before completing the shutdown sequence.  This field is only present during shutdown.

In the same way, I wanted to add the availability_zone in HELLO reply only if it's non-empty, to avoid bloating the reply with a field that's always empty for many users.

enjoy-binbin:
this make sense to me, i also mention this in https://github.com/valkey-io/valkey/pull/1131#issuecomment-2401188840. I would also want a purpose field if we add the action.

hwware:
@valkey-io/core-team Although it is approved, status is still major-decision pending, can we just have a vote for it?

zuiderkwast:
The "stats" section is maybe wrong? Isn't it better to put it in another INFO section?

soloestoy:
@zuiderkwast which section you think is better?

zuiderkwast:
`Clients` section is better? Paused is about paused clients.

Stats is counters of various information. `paused_actions` is just the current state, so not really stats.

soloestoy:
`Clients` makes sense, since it is from `CLIENT PAUSE` command

madolson:
I was going to suggest client or server, since it's server side state. I don't feel strongly to change it, I also agreed I thought stats was the wrong place.

PR #1524, opened 2025-01-07
Add `availability_zone` to the HELLO command history

rueian:
This PR is for https://github.com/valkey-io/valkey/pull/1487.

### DIFF
--- a/src/commands.def
+++ b/src/commands.def
@@ -1719,6 +1719,7 @@ struct COMMAND_ARG ECHO_Args[] = {
 #ifndef SKIP_CMD_HISTORY_TABLE
 /* HELLO history */
 commandHistory HELLO_History[] = {
+{"8.1.0","A new `availability_zone` field is added to the response if the `availability_zone` config is set."},
 {"6.2.0","`protover` made optional; when called without arguments the command reports the current connection's context."},
 };
 #endif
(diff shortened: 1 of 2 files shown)


madolson (approved):
Oops, good followup. We should find a way to add some type of failure on the history, since I know we miss this periodically.

madolson:
@rueian You also need to run `make` and update the commands.def file.

rueian:
> @rueian You also need to run `make` and update the commands.def file.

Oh, thanks for reminding me about that! I have updated the commands.def file.

enjoy-binbin (approved):

src/commands.def
```diff
@@ -1719,6 +1719,7 @@ struct COMMAND_ARG ECHO_Args[] = {
 #ifndef SKIP_CMD_HISTORY_TABLE
 /* HELLO history */
 commandHistory HELLO_History[] = {
+{"8.1.0","A new `availability_zone` field is added to the response if the `availability_zone` config is set."},
```
enjoy-binbin:
```suggestion
{"8.1.0","A new `availability_zone` field is added to the response if the `availability-zone` config is set."},
```

src/commands/hello.json
```diff
@@ -7,6 +7,10 @@
         "arity": -1,
         "function": "helloCommand",
         "history": [
+            [
+                "8.1.0",
+                "A new `availability_zone` field is added to the response if the `availability_zone` config is set."
```
enjoy-binbin:
```suggestion
                "A new `availability_zone` field is added to the response if the `availability-zone` config is set."
```

rueian:
Thank you @enjoy-binbin for the correction!

PR #1525, opened 2025-01-08
refactor: use `argp` library for parsing command line options and arguments

rjd15372:
This commit changes the command line parsing code of `valkey-benchmark` to use the `argp` library (https://www.gnu.org/software/libc/manual/html_node/Argp.html), which is included in libc.

This makes the options and arguments parsing more structured and easy to extend.

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -181,6 +181,11 @@ else
 	# Homebrew x86/ppc uses /usr/local as HOMEBREW_PREFIX
 	OPENSSL_PREFIX?=/usr/local/opt/openssl
 endif
+	# GNU libc argp library needs to be installed with homebrew:
+	# `brew install argp-standalone`
+	FINAL_LIBS+= -largp
+	FINAL_LDFLAGS+= -L/opt/homebrew/lib
+	FINAL_CFLAGS+= -I/opt/homebrew/include
 else
 ifeq ($(uname_S),AIX)
         # AIX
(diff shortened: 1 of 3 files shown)


madolson:
`argp` is apparently not on macos.

rjd15372:
> `argp` is apparently not on macos.

Shoot. I should have checked that before implementing this. I guess that's why the parsing was being done manually.

rjd15372:
To fix the problem of `argp` not being available in macosx libc, I've added `arpg-standalone` package as a build dependency for macosx. `argp-standalone` is easily installed using homebrew, as we do in the github action that runs the build in macosx.

src/Makefile
```diff
@@ -181,6 +181,11 @@ else
 	# Homebrew x86/ppc uses /usr/local as HOMEBREW_PREFIX
 	OPENSSL_PREFIX?=/usr/local/opt/openssl
 endif
+	# GNU libc argp library needs to be installed with homebrew:
+	# `brew install argp-standalone`
+	FINAL_LIBS+= -largp
+	FINAL_LDFLAGS+= -L/opt/homebrew/lib
+	FINAL_CFLAGS+= -I/opt/homebrew/include
```
madolson, line 188: This requires all mac users to use homebrew to use the cli doesn't it, where it is working just fine today.

I guess I'm wondering is it really so important to remove our custom parsing logic?

PingXie: FWIW, I can't use homebrew on my company issued mac

rjd15372: I guess it's not worth changing the custom parsing logic if it does not work in all platforms out of the box.

rjd15372:
Closing this PR since `argp` is not available in macosx.

PR #1532, opened 2025-01-08
Fix flakly half-finish importing cluster test

### LINKED ISSUE #1526: [Test Failure] Half-finish importing in tests/unit/cluster/half-migrated-slot.tcl
```
Expected 'xyz' to be equal to '' (context: type source line 88 file /Users/runner/work/valkey/valkey/tests/unit/cluster/half-migrated-slot.tcl cmd {assert_equal "xyz" [$cluster get aga]} proc ::test)
```

I've seen this test failure before, but the most recent occurrence is: https://github.com/valkey-io/valkey/actions/runs/12661473464/job/35284721513.

sarthakaggarwal97:
With this, the source node is also informed that the slot ownership has been changed.
100 iterations passed for this test, which were flakily failing before.

Fixes #1526.

### DIFF
--- a/tests/unit/cluster/half-migrated-slot.tcl
+++ b/tests/unit/cluster/half-migrated-slot.tcl
@@ -84,6 +84,7 @@ test "Move key back" {
 test "Half-finish importing" {
     # Now we half finish 'importing' node
     assert_equal {OK} [$nodeto(link) cluster setslot 609 node $nodeto(id)]
+    assert_equal {OK} [$nodefrom(link) cluster setslot 609 node $nodeto(id)]
     fix_cluster $nodefrom(addr)
     assert_equal "xyz" [$cluster get aga]
 }
--- a/tests/unit/cluster/half-migrated-slot.tcl
+++ b/tests/unit/cluster/half-migrated-slot.tcl
@@ -81,11 +81,11 @@ test "Move key back" {
     assert_equal "xyz" [$cluster get aga]
 }
 
+reset_cluster
+
 test "Half-finish importing" {
     # Now we half finish 'importing' node
     assert_equal {OK} [$nodeto(link) cluster setslot 609 node $nodeto(id)]
-    assert_equal {OK} [$nodefrom(link) cluster setslot 609 node $nodeto(id)]
-    fix_cluster $nodefrom(addr)
     assert_equal "xyz" [$cluster get aga]
 }
 


tests/unit/cluster/half-migrated-slot.tcl
```diff
@@ -84,6 +84,7 @@ test "Move key back" {
 test "Half-finish importing" {
     # Now we half finish 'importing' node
     assert_equal {OK} [$nodeto(link) cluster setslot 609 node $nodeto(id)]
+    assert_equal {OK} [$nodefrom(link) cluster setslot 609 node $nodeto(id)]
```
hpatro: From the name of the test I think it was intentional from the author to perform `setslot` it only on the target side. Similar to `half-finish migration` test above.

zuiderkwast: Isn't the point of "Half-finish importing" that we should only finish the importing on the target node, not on the source node, and check that the cluster can handle this? Or what does "half-finish" mean here?

sarthakaggarwal97: @hpatro @zuiderkwast I was not sure if half-finish importing is related to setslots itself or not. Without this, the key was ```lost``` in more than a few cases. The output for ```keys *``` was empty. 
I tried with setting additional timeouts, retries, ```wait_for_cluster_propagation```, it didn't seem to reliably fix it. I can mark this PR as draft, and look more behind the scenes.

sarthakaggarwal97: @hpatro I replied to you in the comment below.

zuiderkwast: Thanks, yes it would be good to understand better why the keys are lost. `$nodeto(link) cluster setslot 609 node $nodeto(id)` should make this node bump the epoch and the keys are already migrated to this node, so it seems strange that they are lost. Do you know what the `cluster_fix` line below does?

sarthakaggarwal97: I dove deep a bit into this. 

> Do you know what the cluster_fix line below does

I seems to be running the ```valkey-cli --cluster fix``` command. Based on documentation ~ `in order to fix the cluster so that keys will be migrated according to the hash slots each node is authoritative or not.`
Although, it would have run successfully or else the test might have failed with `fail "Cluster could not settle with configuration"`


I was checking this documentation: https://valkey.io/commands/cluster-setslot/

> The order of step 4 and 5 is important. The destination node is responsible for propagating the change to the rest of the cluster.

Looks like one of the nodes was not informed about the slot migration (the test wants to assert during this behaviour only).  With this half-finish importing, we are at step 4, when we assert whether the key `aga` is present or not. When I was locally replicating this, it seems like the request is indeed getting redirected from the `non-slot-owning` node to the `slot-owning` node.

```
╰─ ./src/valkey-cli -c -h 127.0.0.1 -p 7000                                                                                                                                                                                                                      ─╯
127.0.0.1:7000> get aga
-> Redirected to slot [609] located at 127.0.0.1:6379
"xyz"
127.0.0.1:6379> 

```

`$cluster` looks like needs to be refreshed to know which is the master node for the slot 609. I am not sure if there is a way that client redirection fails here. It's a bit hard for me to replicate this manually with commands. So far, I have not been able to even after moving the key few times.

The latest change to refresh the cluster before we inform the nodes about the migrated key seems to reliably fix the test.

tests/unit/cluster/half-migrated-slot.tcl
```diff
@@ -81,10 +81,11 @@ test "Move key back" {
     assert_equal "xyz" [$cluster get aga]
 }
 
+reset_cluster
+
 test "Half-finish importing" {
     # Now we half finish 'importing' node
     assert_equal {OK} [$nodeto(link) cluster setslot 609 node $nodeto(id)]
-    fix_cluster $nodefrom(addr)
```
hpatro, line 87: This was unnecessary ?

zuiderkwast: With `reset_cluster` before this test case, it isn't necessary I guess, but the two test cases seem to be designed to run together. The test case above sets the slot to migrating/importing and migrates a key.

`fix_cluster` runs `valkey-cli --cluster fix` against the $nodefrom. The nodefrom doesn't know that the migration is finished, so it tries to do guess which node should own the slot based on migrating state and which node owns most of the keys in the slot, IIRC.

On my laptop, this test suite passes every time. How did you reproduce it?

zuiderkwast: Btw, this purpose of this test suite seems to be to test `valkey-cli --cluster fix`. The first line of this test suite is the comment

    # Tests for fixing migrating slot at all stages:

sarthakaggarwal97: > so it tries to do guess which node should own the slot based on migrating state and which node owns most of the keys in the slot, IIRC.

Yeah, that seems to be correct.

> On my laptop, this test suite passes every time. How did you reproduce it?

Running on macos ~ not sure if this efficient but just a small shell command to loop the suite. It fails 2-3 times out of 100 I guess.

```
for i in {1..100}; do                                                                                                                                                                                                                                         ─╯
    ./runtest --single unit/cluster/half-migrated-slot
done
```

zuiderkwast: Ah, I did it like this, which seems equivalent:

    ./runtest --clients 1 --loops 100  --single unit/cluster/half-migrated-slot

On linux it passed every time.

tests/unit/cluster/half-migrated-slot.tcl
zuiderkwast: > The order of step 4 and 5 is important. The destination node is responsible for propagating the change to the rest of the cluster.

This is interesting. I wonder what can happen if this is not followed. This test case is obviously not following this rule, so maybe the test case is invalid? Shall we just delete it?

@PingXie Do you remember what happens if you send SETSLOT NODE to the source node only? Do the other nodes gossip back the ownership of the slot to the old owner?

hpatro: I think `$cluster` client shouldn't be refreshed if the key exists either on source / target node, the server should respond back with data or the test is leading to an incorrect cluster state which shouldn't happen if the set of operations are performed correctly by the administrator. I would recommend figuring out which is the case.

tests/unit/cluster/half-migrated-slot.tcl
sarthakaggarwal97: I was trying `loop` and it was not working smh. thanks this command is better.

zuiderkwast: `reset_cluster` updates the client's view using CLUSTER SLOTS or CLUSTER NODES. Without it, the cluster client should think that the old node owns the slot and then it should receive a redirect and follow it...

tests/unit/cluster/half-migrated-slot.tcl
sarthakaggarwal97: > I think $cluster client shouldn't be refreshed if the key exists either on source / target node

yeah agreed! 

So I just fixed the order of all the commands in the test, but still it didn't really fix the problem. `reset_cluster` seems to be helping point to node which is the master for that slot. `fix_cluster` is helping avoid too many redirects I suppose. Based on some experiments, if I remove `fix_cluster` and not put `reset_cluster` as well, few of the tests fail with `error "Too many redirections or failures contacting Valkey Cluster."`

I am looking into it more, trying to see which scenario is leading to wrong redirects. But again, maybe the issue is with `cluster fix` somewhere.

sarthakaggarwal97: This is interesting. This captures that output when `cluster fix` is run post `assert_equal {OK} [$nodeto(link) cluster setslot 609 node $nodeto(id)]`

**The iteration that failed:**

```
Fix Cluster Response: 127.0.0.1:21112 (92bcf30d...) -> 0 keys | 8192 slots | 0 replicas.
127.0.0.1:21111 (4abc9fde...) -> 0 keys | 8192 slots | 0 replicas.
[OK] 0 keys in 2 primaries.
0.00 keys per slot on average.
>>> Performing Cluster Check (using node 127.0.0.1:21112)
M: 92bcf30d8080f4b924fb609f8c69cf91410e6291 127.0.0.1:21112
   slots:[0-8191] (8192 slots) master
M: 4abc9fdefb81fc7f42d54fa202577772d321efb8 127.0.0.1:21111
   slots:[8192-16383] (8192 slots) master
[OK] All nodes agree about slots configuration.
>>> Check for open slots...
>>> Check slots coverage...
[OK] All 16384 slots covered.
```

**The iteration that passed:**
```
Fix Cluster Response: 127.0.0.1:21112 (927fff91...) -> 0 keys | 8191 slots | 0 replicas.
127.0.0.1:21111 (51a2c417...) -> 1 keys | 8193 slots | 0 replicas.
[OK] 1 keys in 2 primaries.
0.00 keys per slot on average.
>>> Performing Cluster Check (using node 127.0.0.1:21112)
M: 927fff912b41e394fb2c29177129701c29f30087 127.0.0.1:21112
   slots:[0-608],[610-8191] (8191 slots) master
M: 51a2c41705d4a7532192004b6604aa6b0cd03d0a 127.0.0.1:21111
   slots:[609],[8192-16383] (8193 slots) master
[OK] All nodes agree about slots configuration.
>>> Check for open slots...
>>> Check slots coverage...
[OK] All 16384 slots covered.
```

zuiderkwast: The output of `valkey-cli --cluster fix` is useful for debugging. Do this to show the output:

```diff
diff --git a/tests/support/cluster_util.tcl b/tests/support/cluster_util.tcl
index 4f641c5e9..3ea9d4a37 100644
--- a/tests/support/cluster_util.tcl
+++ b/tests/support/cluster_util.tcl
@@ -74,6 +74,8 @@ proc fix_cluster {addr} {
     } result]
     if {$code != 0} {
         puts "valkey-cli --cluster fix returns non-zero exit code, output below:\n$result"
+    } else {
+        puts "Output from cluster fix:\n$result"
     }
     # Note: valkey-cli --cluster fix may return a non-zero exit code if nodes don't agree,
     # but we can ignore that and rely on the check below.
```

sarthakaggarwal97: The outputs I pasted above are from --cluster fix. Trying with the changes you shared.

zuiderkwast: I didn't see you had just posted this output. :)

sarthakaggarwal97: Sharing the difference from the test suite. Rest of the output was same for the tests before this one.

The suite that failed:

```
Output from cluster fix:
127.0.0.1:21112 (d15db001...) -> 0 keys | 8192 slots | 0 replicas.
127.0.0.1:21111 (755f34ea...) -> 0 keys | 8192 slots | 0 replicas.
[OK] 0 keys in 2 primaries.
0.00 keys per slot on average.
>>> Performing Cluster Check (using node 127.0.0.1:21112)
M: d15db00128e2846a625dd38659322771c52c2e77 127.0.0.1:21112
   slots:[0-8191] (8192 slots) master
M: 755f34ea54b64ea4e6c278542b3bfe6a5788aade 127.0.0.1:21111
   slots:[8192-16383] (8192 slots) master
[OK] All nodes agree about slots configuration.
>>> Check for open slots...
[WARNING] Node 127.0.0.1:21112 has slots in migrating state 609.
[WARNING] The following slots are open: 609.
>>> Fixing open slot 609
Set as migrating in: 127.0.0.1:21112
>>> Case 4: Closing slot 609 on 127.0.0.1:21112
>>> Check slots coverage...
[OK] All 16384 slots covered.
```

The suite that passed:
```
Output from cluster fix:
127.0.0.1:21112 (f1c73dea...) -> 0 keys | 8191 slots | 0 replicas.
127.0.0.1:21111 (13968322...) -> 1 keys | 8193 slots | 0 replicas.
[OK] 1 keys in 2 primaries.
0.00 keys per slot on average.
>>> Performing Cluster Check (using node 127.0.0.1:21112)
M: f1c73deadd9f0bc2591efda8a48e98968a071de1 127.0.0.1:21112
   slots:[0-608],[610-8191] (8191 slots) master
M: 13968322f7aa35626f2a3bb69cc2c4528567571e 127.0.0.1:21111
   slots:[609],[8192-16383] (8193 slots) master
[OK] All nodes agree about slots configuration.
>>> Check for open slots...
>>> Check slots coverage...
[OK] All 16384 slots covered.
```

zuiderkwast: > `reset_cluster` seems to be helping point to node which is the master for that slot.

It makes the cluster client (`$cluster`) reload the slot mapping. I don't know from which node it loads it. It may matter if the nodes have a different view on slot ownership. This does nothing to the cluster nodes though.

> `fix_cluster` is helping avoid too many redirects I suppose.

What it does is that it tries to fix a broken cluster by finishing aborted slot migrations and making sure every slot has one owner. It modifies the cluster itself.

> Based on some experiments, if I remove `fix_cluster` and not put `reset_cluster` as well, few of the tests fail with `error "Too many redirections or failures contacting Valkey Cluster."`

This makes sense. This test case told the `nodefrom` node that the `nodeto` node is now the owner, so after this, `nodeto` redirects any commands for this slot to `nodeto`. But the `nodeto` node still thinks that it's not yet the owner, that it's just in an ongoing importing state, so it redirects back to `nodefrom`. (It's possible to read from a node in importing state if you follow an ASK redirect, i.e. if you send ASKING before the GET command.)

This is just to understand the situation here. What would be really interesting to understand is what differs between the success and the failure case with the unmodified test case. I hope the output of cluster fix can give a clue.

zuiderkwast: To see what `cluster_fix` does, look at this function in `valkey-cli.c`:

```C
/* Slot 'slot' was found to be in importing or migrating state in one or
 * more nodes. This function fixes this condition by migrating keys where
 * it seems more sensible. */
static int clusterManagerFixOpenSlot(int slot) {
```

sarthakaggarwal97: This is slightly bugging me. Before it even reaches the `clusterManagerFixOpenSlot`

It says via `clusterManagerShowClusterInfo`:

```
Output from cluster fix:
127.0.0.1:21112 (d15db001...) -> 0 keys | 8192 slots | 0 replicas.
127.0.0.1:21111 (755f34ea...) -> 0 keys | 8192 slots | 0 replicas.
```

That means, there are no keys to begin with. 

As a result of that, we end up falling into Case 4 (`>>> Case 4: Closing slot 609 on 127.0.0.1:21112`). 
> Case 4: There are no slots claiming to be in importing state, but  there is a migrating node that actually don't have any key or is the slot owner. We can just close the slot, probably a reshard interrupted in the middle.

In Case 4, if there are no keys, the slot is basically closed.


Also one of the nodes are in migrating state: 
```
[WARNING] Node 127.0.0.1:21112 has slots in migrating state 609.
```

hpatro: @sarthakaggarwal97 Before we hit this, could you confirm the key count prior to this?

PingXie: >@PingXie Do you remember what happens if you send SETSLOT NODE to the source node only? 

I think with the current implementation, the source will assign the slot to the target without bumping up the target's config epoch. The source will then stop claiming this slot in future cluster messages. Therefore, only the source knows the slot belongs to the target, but everyone else in the cluster, including the target, still thinks the source owns it. This effectively loses the slot. Any request targeting it will be redirected to the source by any other node. However, since the source updated its slot mapping and assigned the slot to the target, it will redirect the request to the target, which believes the source is the owner, sending the request back to the source, completing the cycle. Eventually, the client would time out.

> Do the other nodes gossip back the ownership of the slot to the old owner?

I don't think so. After executing `SETSLOT NODE` on the old owner (source), it will no longer claim the slot in its cluster messages. This eliminates any possibility of slot ownership conflicts on all other nodes, so the following "update" logic will not be executed. https://github.com/valkey-io/valkey/blob/e60990e5798a4a9a943aa444dba24855d59674c0/src/cluster_legacy.c#L3557

zuiderkwast: Ok, as I also thought, we get into a ping-pong redirect.

But whoever owns or doesn't own the slot, the key shouldn't be lost so `valkey-cli --cluster fix` should be able to fix this situation. I don't understand why the key is lost sometimes. My guess now is that it's a valkey-cli bug.

sarthakaggarwal97: > Therefore, only the source knows the slot belongs to the target, but everyone else in the cluster, including the target, still thinks the source owns it. This effectively loses the slot.

This has been my hypothesis too. Based on multiple runs and debugs, this looks to be true. If we don't run `--cluster fix`, it's errored out because of too many redirects in the test suite or client timeout eventually.

> My guess now is that it's a valkey-cli bug.

Possibly. The key is accessible even when the `setslot node` command is run on the target node, but once the `fix_cluster` is executed, the key isn't there anymore (in rare cases).

PingXie: >I don't understand why the key is lost sometimes.

Sorry this is a long thread. Can someone summarize the sequence of critical events that led to the key getting lost? For instance, when was the key introduced? on the source or the target? etc

sarthakaggarwal97: There is a test suite `half-migrated-slot.tcl`, where `test "Half-finish importing"` is failing flaky-ly because it is not able to find the key.

This test runs prior to the test that is failing.
```
test "Move key back" {
    # 'aga' key is in 609 slot
    assert_equal {OK} [$nodefrom(link) cluster setslot 609 migrating $nodeto(id)]
    assert_equal {OK} [$nodeto(link) cluster setslot 609 importing $nodefrom(id)]
    assert_equal {OK} [$nodefrom(link) migrate $nodeto(host) $nodeto(port) aga 0 10000]
    wait_for_cluster_propagation
    assert_equal "xyz" [$cluster get aga]
}
```

With this, the slot migration is in progress, where target is importing, source is migrating, and we migrate the only key `aga` present in the slot.

Based on the [documentation](https://valkey.io/commands/cluster-setslot/), it order of migration operations is important, which seems to be not prefect in the test, but isn't the cause of flaky tests here.

Then in the next test, we inform the destination node about the slot migration via this.

```
// key is still present here
assert_equal {OK} [$nodeto(link) cluster setslot 609 node $nodeto(id)]

// key is still present here
fix_cluster $nodeto(addr)

// key is now absent in rare cases, hence the below assertion fails
assert_equal "xyz" [$cluster get aga]
```

This is the output from cluster fix when the test fails, it doesn't seem to even find the key. This is coming from somewhere here: https://github.com/valkey-io/valkey/blob/unstable/src/valkey-cli.c#L7191 and subsequent functions that are called.

```
127.0.0.1:21125 (30447658...) -> 0 keys | 8192 slots | 0 replicas.
127.0.0.1:21126 (a316bb6e...) -> 0 keys | 8192 slots | 0 replicas.
[OK] 0 keys in 2 primaries.
0.00 keys per slot on average.
>>> Performing Cluster Check (using node 127.0.0.1:21125)
M: 30447658b60642f5f14046a28fe129641d1cdbc4 127.0.0.1:21125
   slots:[8192-16383] (8192 slots) master
M: a316bb6e95463d92dbbe9e7687b8cdefbad1cc43 127.0.0.1:21126
   slots:[0-8191] (8192 slots) master
[OK] All nodes agree about slots configuration.
>>> Check for open slots...
[WARNING] Node 127.0.0.1:21126 has slots in migrating state 609.
[WARNING] The following slots are open: 609.
>>> Fixing open slot 609
Set as migrating in: 127.0.0.1:21126
>>> Case 4: Closing slot 609 on 127.0.0.1:21126
>>> Check slots coverage...
```


In the output, it seems to be going into `Case 4`, in which the slot is closed when it doesn't see any key in that slot.

zuiderkwast: @sarthakaggarwal97 can you try @hpatro's suggestion:

> Before we hit this, could you confirm the key count prior to this? 

Just check the number of keys in the slot on the $nodeto, for example like this:

```diff
diff --git a/tests/unit/cluster/half-migrated-slot.tcl b/tests/unit/cluster/half-migrated-slot.tcl
index 5374ba4ba..8565c7aba 100644
--- a/tests/unit/cluster/half-migrated-slot.tcl
+++ b/tests/unit/cluster/half-migrated-slot.tcl
@@ -84,7 +84,9 @@ test "Move key back" {
 test "Half-finish importing" {
     # Now we half finish 'importing' node
     assert_equal {OK} [$nodeto(link) cluster setslot 609 node $nodeto(id)]
+    assert_equal 1 [$nodeto(link) cluster countkeysinslot 609]
     fix_cluster $nodefrom(addr)
+    assert_equal 1 [$nodeto(link) cluster countkeysinslot 609]
     assert_equal "xyz" [$cluster get aga]
 }
 
```

zuiderkwast: You checked this already? I see you wrote

> The key is accessible even when the `setslot node` command is run on the target node, but once the `fix_cluster` is executed, the key isn't there anymore (in rare cases).

So valkey-cli didn't count the keys correctly? I wonder how it does this.

To see which commands valkey-cli sent to the nodes, it should be possible to check command stats on the nodes:

```tcl
set nodeto_command_stats [$nodeto(link) info commandstats]
puts "Nodeto command stats: $nodeto_command_stats"
```

sarthakaggarwal97: > You checked this already? I see you wrote

Yeah, I placed these everywhere. When the test fails, the `keys *` response is empty

```
 puts [$nodefrom(link) keys *]
 puts [$nodeto(link) keys *]
```

> To see which commands valkey-cli sent to the nodes, it should be possible to check command stats on the nodes:

Let me try your method as well. Thanks @hpatro suggestion, I tried something similar with monitor [truncating the output.]

sarthakaggarwal97: @zuiderkwast this is with `command stats` when the test fails.

```
Nodeto command stats: # Commandstats
cmdstat_exec:calls=1,usec=25,usec_per_call=25.00,rejected_calls=0,failed_calls=0
cmdstat_select:calls=1,usec=0,usec_per_call=0.00,rejected_calls=0,failed_calls=0
cmdstat_info:calls=2,usec=72,usec_per_call=36.00,rejected_calls=0,failed_calls=0
cmdstat_get:calls=3,usec=3,usec_per_call=1.00,rejected_calls=2,failed_calls=0
cmdstat_asking:calls=2,usec=0,usec_per_call=0.00,rejected_calls=0,failed_calls=0
cmdstat_cluster|setslot:calls=14,usec=101,usec_per_call=7.21,rejected_calls=0,failed_calls=0
cmdstat_cluster|addslotsrange:calls=1,usec=62,usec_per_call=62.00,rejected_calls=0,failed_calls=0
cmdstat_cluster|nodes:calls=25,usec=595,usec_per_call=23.80,rejected_calls=0,failed_calls=0
cmdstat_cluster|addslots:calls=1,usec=6,usec_per_call=6.00,rejected_calls=0,failed_calls=0
cmdstat_cluster|bumpepoch:calls=1,usec=6,usec_per_call=6.00,rejected_calls=0,failed_calls=0
cmdstat_cluster|countkeysinslot:calls=9,usec=3,usec_per_call=0.33,rejected_calls=0,failed_calls=0
cmdstat_cluster|myid:calls=1,usec=0,usec_per_call=0.00,rejected_calls=0,failed_calls=0
cmdstat_cluster|getkeysinslot:calls=3,usec=2,usec_per_call=0.67,rejected_calls=0,failed_calls=0
cmdstat_cluster|shards:calls=11,usec=370,usec_per_call=33.64,rejected_calls=0,failed_calls=0
cmdstat_cluster|delslots:calls=2,usec=13,usec_per_call=6.50,rejected_calls=0,failed_calls=0
cmdstat_cluster|info:calls=7,usec=141,usec_per_call=20.14,rejected_calls=0,failed_calls=0
cmdstat_cluster|slots:calls=11,usec=323,usec_per_call=29.36,rejected_calls=0,failed_calls=0
cmdstat_config|set:calls=1,usec=5,usec_per_call=5.00,rejected_calls=0,failed_calls=0
cmdstat_debug:calls=1,usec=20,usec_per_call=20.00,rejected_calls=0,failed_calls=0
cmdstat_keys:calls=3,usec=8,usec_per_call=2.67,rejected_calls=0,failed_calls=0
cmdstat_restore-asking:calls=2,usec=17,usec_per_call=8.50,rejected_calls=0,failed_calls=0
cmdstat_ping:calls=1,usec=0,usec_per_call=0.00,rejected_calls=0,failed_calls=0
cmdstat_migrate:calls=1,usec=415,usec_per_call=415.00,rejected_calls=0,failed_calls=0
cmdstat_multi:calls=1,usec=1,usec_per_call=1.00,rejected_calls=0,failed_calls=0
cmdstat_dbsize:calls=10,usec=5,usec_per_call=0.50,rejected_calls=0,failed_calls=0

```

hpatro: From the monitor command output shared above, I don't see a `DEL` operation there but the `keys *` output is empty? That doesn't add up. Could you also check the output of `DEBUG HTSTATS 0` along with `keys *`?

hpatro: Also for commandstats, could you reset after the half-migrating test is completed and then capture it. It's a bit difficult to process the information as it is from the start of the server.

sarthakaggarwal97: `DEBUG HTSTATS 0` is empty when the test fails

Output:
```
[Dictionary HT]
[Expires HT]

```

Command Stats after reset:

```
Nodeto command stats: # Commandstats
cmdstat_restore-asking:calls=1,usec=12,usec_per_call=12.00,rejected_calls=0,failed_calls=0
cmdstat_debug:calls=1,usec=3,usec_per_call=3.00,rejected_calls=0,failed_calls=0
cmdstat_keys:calls=3,usec=10,usec_per_call=3.33,rejected_calls=0,failed_calls=0
cmdstat_dbsize:calls=2,usec=0,usec_per_call=0.00,rejected_calls=0,failed_calls=0
cmdstat_get:calls=1,usec=1,usec_per_call=1.00,rejected_calls=0,failed_calls=0
cmdstat_asking:calls=1,usec=0,usec_per_call=0.00,rejected_calls=0,failed_calls=0
cmdstat_config|resetstat:calls=1,usec=36,usec_per_call=36.00,rejected_calls=0,failed_calls=0
cmdstat_cluster|slots:calls=1,usec=10,usec_per_call=10.00,rejected_calls=0,failed_calls=0
cmdstat_cluster|nodes:calls=4,usec=93,usec_per_call=23.25,rejected_calls=0,failed_calls=0
cmdstat_cluster|setslot:calls=2,usec=12,usec_per_call=6.00,rejected_calls=0,failed_calls=0
cmdstat_cluster|shards:calls=1,usec=11,usec_per_call=11.00,rejected_calls=0,failed_calls=0
cmdstat_cluster|countkeysinslot:calls=2,usec=0,usec_per_call=0.00,rejected_calls=0,failed_calls=0
cmdstat_cluster|info:calls=1,usec=14,usec_per_call=14.00,rejected_calls=0,failed_calls=0

```

I don't see a delete call :(

sarthakaggarwal97: While experimenting with the test, this seemed intriguing.

These lines are executed, after the `fix_cluster $nodefrom(addr)`
```
    puts $nodeto(addr)
    puts $nodefrom(addr)
    puts [$nodeto(link) get aga]
```

I got this response
```
127.0.0.1:21143
127.0.0.1:21144
[exception]: Executing test client: MOVED 609 127.0.0.1:21144.
MOVED 609 127.0.0.1:21144
```

But `HTSTATS` for both the nodes aren't showing the key.

hpatro: I tried running it on mac/linux, it never fails for me. Do we need to build it with any particular flags to reproduce it?

sarthakaggarwal97: No! Just `make` and if I run the test in loop for few times. I am on `Apple M3 Chip`. Are you also checking aarch/arm? (Not sure if this is making a difference.)

sarthakaggarwal97: @PingXie @hpatro @zuiderkwast any ideas where we can look for here. Or should we look for a bug in `--cluster fix` command itself.

zuiderkwast: No new ideas, sorry. I would look at bugs in `valkey-cli --cluster fix` as well, yes, and figure out exactly what's happening when the key gets deleted. Narrow down the problem, use gdb with breakpoints and stacktraces, spray printf all over the code......

madolson:
@sarthakaggarwal97 This test failed against yesterday, are you still working on this? https://github.com/valkey-io/valkey/actions/runs/13424893119/job/37505823362

sarthakaggarwal97:
I haven't seen this test failure in a while, so closing the PR!

PR #1534, opened 2025-01-09
Add latency stats around cluster config file operations

enjoy-binbin:
When the cluster changes, we need to persist the cluster configuration,
and these file IO operations may cause latency.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -816,6 +816,7 @@ int clusterSaveConfig(int do_fsync) {
     ssize_t written_bytes;
     int fd = -1;
     int retval = C_ERR;
+    mstime_t latency;
 
     server.cluster->todo_before_sleep &= ~CLUSTER_TODO_SAVE_CONFIG;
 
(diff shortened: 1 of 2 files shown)


src/cluster_legacy.c
```diff
@@ -844,31 +849,52 @@ int clusterSaveConfig(int do_fsync) {
...
 
     if (do_fsync) {
+        latencyStartMonitor(latency);
         server.cluster->todo_before_sleep &= ~CLUSTER_TODO_FSYNC_CONFIG;
         if (valkey_fsync(fd) == -1) {
             serverLog(LL_WARNING, "Could not sync tmp cluster config file: %s", strerror(errno));
             goto cleanup;
         }
+        latencyEndMonitor(latency);
+        latencyAddSampleIfNeeded("cluster-config-fsync", latency);
```
madolson, line 863: We've only seen significant time taken in the fsync (both here and in the dir). Have you observed the other operations taking significant amounts of time?

enjoy-binbin: The other one I see most is open (10ms - 5s), which seems to try to get the dir lock something like that. And write sometimes take a few ms. I haven't seen the others in the production environment but i think i do see other take a few ms in the test environment (i dont see the unlink one). I see we have a lot latency around AOF, fsync/rename/write/fstats, so i added it all.

madolson: Sounds good. I suppose open does also make sense. I am thinking now that since AWS uses EBS volumes, we might see the same type of latency on IO operations that other distributions see.

madolson (approved):

hpatro (approved):
Could we also add some sanity tests?

enjoy-binbin:
```
start_cluster 1 1 {tags {"latency-monitor cluster external:skip needs:latency"} overrides {latency-monitor-threshold 1}} {
    test "Cluster config file latency" {
        puts [r latency latest]
    }
}
```

i can get this ouput in my local mac, but it is not certain that they will show in CI. Do you have a better idea?
```
{command 1736483355 1 1} {cluster-config-fsync 1736483355 6 6} {cluster-config-rename 1736483355 1 1} {cluster-config-open 1736483355 1 1} {cluster-config-close 1736483355 1 1} {fork 1736483355 1 1} {cluster-config-dir-fsync 1736483355 5 5}
[ok]: Cluster config file latency (1 ms)
```

PingXie (approved):

hpatro:
> i can get this ouput in my local mac, but it is not certain that they will show in CI. Do you have a better idea?
> 

Didn't follow. Could you explain why it won't show up in CI?

enjoy-binbin:
> Didn't follow. Could you explain why it won't show up in CI?

If it is 0ms, it will not show up in the tests.

PR #1540, opened 2025-01-09
Set GH actions job timeout to a day

hpatro:
It's currently set to 14400 which is 10 days. I think we shouldn't leave jobs lingering around for that amount of period.


enjoy-binbin (approved):

enjoy-binbin:
i somehow feel that one day is still quite long.

hpatro:
> i somehow feel that one day is still quite long.

Yes, I also feel so. Wasn't sure how long can jobs be under waiting state for GH action runners.

hpatro:
@enjoy-binbin Any suggestions on the timeout value we should set? I could maybe push this change as is and observe if we hit any regression.

enjoy-binbin:
lets go ahead!

PR #1541, opened 2025-01-10
Fix module LatencyAddSample still work when latency-monitor-threshold is 0

enjoy-binbin:
When latency-monitor-threshold is set to 0, it means the latency monitor
is disabled, and in VM_LatencyAddSample, we wrote the condition incorrectly,
causing us to record latency when latency was turned off.

This bug was introduced in the very first day, see e3b1d6d, was merged in 2019.

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -7680,7 +7680,8 @@ void VM__Assert(const char *estr, const char *file, int line) {
  * command. The call is skipped if the latency is smaller than the configured
  * latency-monitor-threshold. */
 void VM_LatencyAddSample(const char *event, mstime_t latency) {
-    if (latency >= server.latency_monitor_threshold) latencyAddSample(event, latency);
+    if (server.latency_monitor_threshold && latency >= server.latency_monitor_threshold)
+        latencyAddSample(event, latency);
 }
 
 /* --------------------------------------------------------------------------
--- a/src/module.c
+++ b/src/module.c
@@ -7680,8 +7680,7 @@ void VM__Assert(const char *estr, const char *file, int line) {
  * command. The call is skipped if the latency is smaller than the configured
  * latency-monitor-threshold. */
 void VM_LatencyAddSample(const char *event, mstime_t latency) {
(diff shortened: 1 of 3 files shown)


enjoy-binbin:
Currently LATENCY LATEST only has the following fields:
- Event name.
- Unix timestamp of the latest latency spike for the event.
- Latest event latency in millisecond.
- All-time maximum latency for this event.

I would also like to take it opportunity to discuss, if we can add more fields, like:
- min, the all-time minimum latency for this event (maybe not that useful)
- sum, the all-time sum latency for this event (i think it is useful for some events)
- cnt, the event count that trigger the latency, with the sum we can calc the avg

PingXie (approved):

zuiderkwast (approved):
Great.

When was the bug introduced? Backporting, release notes, etc...

PingXie:
> Great.
> 
> When was the bug introduced? Backporting, release notes, etc...

Day 1 bug introduced in 7.2.4 Redis: https://github.com/valkey-io/valkey/commit/e3b1d6d3ad

madolson (approved):

madolson:
The PR was merged in 2019, it's a much older bug than that.

src/module.c
```diff
@@ -7680,7 +7680,8 @@ void VM__Assert(const char *estr, const char *file, int line) {
  * command. The call is skipped if the latency is smaller than the configured
  * latency-monitor-threshold. */
 void VM_LatencyAddSample(const char *event, mstime_t latency) {
-    if (latency >= server.latency_monitor_threshold) latencyAddSample(event, latency);
+    if (server.latency_monitor_threshold && latency >= server.latency_monitor_threshold)
+        latencyAddSample(event, latency);
```
```suggestion
    latencyAddSampleIfNeeded(event, latency);
```
Optional, I just noticed we have this helper.

enjoy-binbin: great.

enjoy-binbin:
I added the backport label, feel free to change it if you think otherwise.

PR #1543, opened 2025-01-11
Use MSG_ZEROCOPY for plaintext replication traffic

### LINKED ISSUE #1335: Psync Zero-Copy Transmission
### Problem Statement
In the current design, the primary maintains a replication buffer to record mutation commands for syncing the replicas. This replication buffer is implemented as a linked list of chunked buffers. The primary periodically transmits these recorded commands to each replica by issuing socket writes on the replica connections, which involve copying data from the user-space buffer to the kernel.
The transmission is performed by the [writeToReplica](https://github.com/valkey-io/valkey/blob/49863109453faa907ce2c8b1158e60a6777d28ab/src/networking.c#L1997) function, which uses _connWrite_ to send data over the socket.

This user-space to kernel buffer copy consumes CPU cycles and increases the memory footprint.  The overhead becomes more noticeable when a replica lags significantly behind the primary, as _pysnc_ triggers a transmission burst. This burst may temporarily reduce the primary's responsiveness, with excessive copying and potential TCP write buffer exhaustion being major contributing factors.

### Proposal
Modern Linux systems support zero-copy transmission, which operates by:
* The user-space provides a buffer directly to the kernel for transmission.
* The kernel uses this buffer without duplicating it.
* Transmission progress is communicated back to the user-space through notifications.  _epoll_ can be used as the notification mechanism.   

The primary downside of zero-copy is the need for userspace to manage the send buffer. However, this limitation is much less applicable for the _psync_ use case as Valkey already manages the _pysnc_ replication buffers.

It’s important to note that using zero-copy for _psync_ requires careful adjustments of the replica client write buffers management logic. Specifically, the logic to ensure that the total accumulated replication write buffer size, across all the replica connections, is limited to the value of _client-output-buffer-limit replica_.

Further reading on zero-copy can be found [here](https://docs.kernel.org/networking/msg_zerocopy.html). 
Note that this article states that zero-copy is most effective for large payloads, and experimentation is necessary to determine the minimum payload size. For Memorystore vector search cluster communication, enabling zero-copy in gRPC improved QPS by approximately 8.6%.

### Zero-Copy Beyond _Psync_
Zero-copy can also optimize transmission to clients. In the current implementation, dictionary entries are first copied into the client object's write buffer and then copied again during transmission to the client socket, resulting in two memory copies. Using zero-copy eliminates the client socket copy.
Similarly to the _psync_ use case, implementing zero-copy for client transmission requires careful adjustments to the client’s write buffer management logic. The following considerations, while not exhaustive, outline key aspects to address:
1. Adjustments to the accumulated client send buffer size upper limit logic may be required, similarly to the _psync_ use case. Note that the upper limit for the accumulated client send buffer is controlled by the _client-output-buffer-limit normal_ parameter.
2. Currently, for large dictionary entries, transmission to the client involves multiple iterations of copying chunks of data from the dictionary entry into the client send buffer, and subsequently from the send buffer to the socket. If the socket's TCP send buffer becomes full, the copying process is suspended until some send buffer is available, preventing excessive memory usage.
Since zero-copy doesn’t consume TCP buffers, excessive memory usage prevention must be handled differently. One approach is to defer copying the next portion of the dictionary entry until a confirmation is received that a significant part of the pending write buffer has been received by the replica.

murphyjacob4:
### Summary

This PR integrates with `MSG_ZEROCOPY` for plaintext replication traffic. [MSG_ZEROCOPY ](https://www.kernel.org/doc/html/v4.15/networking/msg_zerocopy.html) is a Linux kernel functionality that allows writes to avoid copying data from user space to kernel space. Instead, `MSG_ZEROCOPY` allows the kernel to pin the user space data into kernel space for asynchronous writing without any copying. In turn, users have to track the ongoing writes and ensure the data is kept stable, listening on the socket's message queue for a completion notification.

### Design

We track the ongoing writes through a dynamic circular buffer, `zeroCopyTracker`. Each ongoing write is indexed by its sequence number, matching the one assigned by the kernel, with each entry holding a refcount into the replication backlog. We also add a new event type to the event loop APIs (`AE_ERROR_QUEUE`) to register for `SO_EE_ORIGIN_ZEROCOPY` notifications from the kernel. When we choose to use zero-copy for writes, we append a tracking entry to the circular buffer and register for the AE_ERROR_QUEUE event, which later fires and notifies us of a batch of completions, allowing us to trim the tracker which in turn decrements the refcount and trims the replication backlog.

To maintain a net improvement on performance,`MSG_ZEROCOPY` is only used in situations where it should perform better than our typical `write` syscalls:

1. `MSG_ZEROCOPY` is only used for writes that are over 10 KiB (by default, user configurable)
2. `MSG_ZEROCOPY` is only used for remote connections
3. `MSG_ZEROCOPY` is only used for plaintext connections
    1. Adding support for TLS should be feasible, but leaving this as an iterative improvement

This PR enables `MSG_ZEROCOPY` as the new default behavior for writes from the replication backlog that meet the aforementioned criteria on builds that support it. It also introduces a config to disable it altogether via `--tcp-tx-zerocopy no`.

Although this PR only implements it for replication, the core concept should be reusable at various other places where we do writes, so long as reference counting can be used to defer cleanup until a later `epoll` cycle is notified of the write completion. The prospects of combining this with something like IO uring via [IORING_OP_SENDZC](https://lwn.net/Articles/879724/) are especially exciting. We should aim to continue to evolve the core concept of the `zeroCopyTracker` to meet these needs.

Some other notable call outs in the implementation:

1. The connection shutdown flow needs to be modified for connections that have ongoing zero copy writes. We are not allowed to touch this memory until the kernel has explicitly told us to do so. Instead of the normal `close` operation, we instead go into a draining phase where we first call `shutdown`, then await the zero copy tracker to reach zero length before later calling `close` and freeing the client in entirety. 
2. Using zero-copy should be a net improvement for memory, however it does change the semantics of how memory is accounted. Previously, TCP memory was not tracked at all by the Valkey process, but now we will be internalizing this TCP memory into the buffers that we hold for the kernel. Overall, this should reduce memory usage on the machine since each ongoing write for the same memory will not require a copy, but it may lead to situations where replication backlog and Valkey process memory looks larger in comparison to previous versions.
3. For testing purposes, we add two DEBUG commands, one to allow us to simulate a slow replica via pausing completion notification events, and another to allow us to use zero-copy over loopback.
4. Introduced some new stats to help track how many zero-copy writes are happening, how many are currently in flight, and how many connections are in the "draining" state. Additionally, added memory tracking info for the `zeroCopyTracker`.
5. Since the kernel only uses uint32_t sequence numbers, it is important that we handle sequence number wrap around gracefully in `zeroCopyTracker`. A single replication link with 4 billion writes is not unexpected on a long-running instance.

### Performance

Copying and pasting some initial performance comparisons from https://github.com/valkey-io/valkey/issues/1335:

Key size (B) | Primary Throughput Delta | Replica Throughput Delta
-- | -- | -- 
1024 | +2.31% | +2.31%
4096  | +4.30% | +4.30%
10240 | +7.67% | +7.67%
40960 | +17.67% | +17.67%
102400  | +16.02% | +16.02%
409600  | +16.84% | +1.57%

closes https://github.com/valkey-io/valkey/issues/1335


src/ae.c
```diff
@@ -475,34 +477,49 @@ int aeProcessEvents(aeEventLoop *eventLoop, int flags) {
...
-                }
+            if (invert) {
+                event_order[0] = AE_ERROR_QUEUE;
+                event_order[1] = AE_WRITABLE;
+                event_order[2] = AE_READABLE;
+            } else {
+                event_order[0] = AE_ERROR_QUEUE;
+                event_order[1] = AE_READABLE;
+                event_order[2] = AE_WRITABLE;
             }
```
ranshid, line 488:
```suggestion
            event_order[0] = AE_ERROR_QUEUE;
            event_order[1] = invert ? AE_WRITABLE : AE_READABLE;
            event_order[0] = invert ? AE_READABLE : AE_WRITABLE;
```

src/ae.h
```diff
@@ -39,14 +39,17 @@
...
-                         to do that in a group fashion. */
+#define AE_NONE 0             /* No events registered. */
+#define AE_READABLE 1 << 0    /* Fire when descriptor is readable. */
+#define AE_WRITABLE 1 << 1    /* Fire when descriptor is writable. */
+#define AE_BARRIER 1 << 2     /* With WRITABLE, never fire the event if the      \
+                                 READABLE event already fired in the same event  \
+                                 loop iteration. Useful when you want to persist \
+                                 things to disk before sending replies, and want \
+                                 to do that in a group fashion. */
+#define AE_ERROR_QUEUE 1 << 3 /* Fire when descriptor has a message on the \
```
ranshid, line 50:
```suggestion
#define AE_READABLE (1 << 0)    /* Fire when descriptor is readable. */
#define AE_WRITABLE (1 << 1)    /* Fire when descriptor is writable. */
#define AE_BARRIER (1 << 2)     /* With WRITABLE, never fire the event if the      \
                                 READABLE event already fired in the same event  \
                                 loop iteration. Useful when you want to persist \
                                 things to disk before sending replies, and want \
                                 to do that in a group fashion. */
#define AE_ERROR_QUEUE (1 << 3) /* Fire when descriptor has a message on the \
```

src/config.c
```diff
@@ -3289,6 +3303,7 @@ standardConfig static_configs[] = {
     createIntConfig("rdma-port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.rdma_ctx_config.port, 0, INTEGER_CONFIG, NULL, updateRdmaPort),
     createIntConfig("rdma-rx-size", NULL, IMMUTABLE_CONFIG, 64 * 1024, 16 * 1024 * 1024, server.rdma_ctx_config.rx_size, 1024 * 1024, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("rdma-completion-vector", NULL, IMMUTABLE_CONFIG, -1, 1024, server.rdma_ctx_config.completion_vector, -1, INTEGER_CONFIG, NULL, NULL),
+    createIntConfig("tcp-zerocopy-min-write-size", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.tcp_zerocopy_min_write_size, CONFIG_DEFAULT_ZERO_COPY_MIN_WRITE_SIZE, INTEGER_CONFIG, NULL, NULL),
```
ranshid, line 3306: I think this falls under the "configurations we do not want users to mess with".  agree 10K is the recommended sweet spot, but not sure if this needs any kind of tunability option ATM.
Consider dropping this config for now

src/networking.c
```diff
@@ -1911,31 +1936,75 @@ client *lookupClientByID(uint64_t id) {
...
+            if (use_zerocopy) {
+                /* Lazily enable zero copy at the socket level only on first use */
+                if (!c->zero_copy_tracker) {
+                    connSetZeroCopy(c->conn, 1);
+                    c->zero_copy_tracker = createZeroCopyTracker();
+                }
+                nwritten = zeroCopyWriteToConn(c->conn, o->buf + c->repl_data->ref_block_pos, data_len);
+            } else {
+                nwritten = connWrite(c->conn, o->buf + c->repl_data->ref_block_pos, data_len);
+            }
```
ranshid, line 1959: This seems like it could be encapsulated in the connection abstraction maybe? I wonder if the tracker should be a client or a connection owned? I think it makes more sense that it will be part of the connection.

ranshid (commented):
Just some highlevel comments.
I still did not dive deep into the review of this, but it does sound promising. 
I also wonder how this will look like when we do integrate io-uring.
there would be several options to go:
1. no io-uring AND no ZERO-COPY support
2. io-uring support but NO ZERO-COPY support
3. both io-uring AND ZERO-COPY support

I think we would probably want to support all 3 modes, but we could also just invest in io-uring+ZERO copy support which could simplify the tracker code (am I correct in this observation?)

src/networking.c
murphyjacob4: Good question - on my initial pass I was thinking that the tracker fit better as a client-level abstraction simply because the logic is heavily tied to the replication backlog, and having the connection deal with replication  backlog mechanics felt like a bad smell. But perhaps it would be better to start with a tracker that supports a more generic callback-based approach:

```
typedef void (*ZeroCopyWriteCompleteCallback)(void *);
/* Tracking struct used to trigger cleanup once a zero-copy write is finished */
typedef struct zeroCopyRecord {
    int active;
    ZeroCopyWriteCompleteCallback callback;
    void *privdata;
} zeroCopyRecord;
```

Then maybe connection exposes an API like:

```
int connWriteZeroCopy(connections *conn, char *bug, size_t len, ZeroCopyWriteCompleteCallback callback, void *privdata) {
    /* Do write and enqueue zeroCopyRecord */
    /* Later, when zero copy is done, connection will trigger the ZeroCopyWriteCompleteCallback with privdata */
}
```

Is this what you had in mind?

src/config.c
murphyjacob4: I also didn't have it at first - but in tests it was much easier to deterministically trigger zero copy writes if I can play with the minimum write size.

> configurations we do not want users to mess with

I can move it to a DEBUG command if we would prefer, if we have concerns about maintaining new configs?

murphyjacob4:
> I think we would probably want to support all 3 modes, but we could also just invest in io-uring+ZERO copy support which could simplify the tracker code (am I correct in this observation?)

Yeah I believe the polling mechanism will be different:
* MSG_ZEROCOPY - poll the socket for error queue messages with `EPOLLERR`
* io-uring+zero copy -  polling a completion queue for completion queue entries with `EPOLLIN`

So the difference would be in having one vs. two handlers, but ideally once the completion notification is read we can merge into a common format and process it similarly for both.

There are limitations regarding io-uring adoption (https://www.phoronix.com/news/Google-Restricting-IO_uring) so having both modes would be ideal for those who are unable/unwilling to adopt IO uring.

src/config.c
ranshid: I am not sure why it makes the test much easier? I thiknk DEBUG command is fine and also a HIDDEN config which is named prefixed/suffixed with 'debug' is fine IMO.

murphyjacob4: > I am not sure why it makes the test much easier?

Ultimately we have no control over if a call to `send` will return <= the number of bytes we want it to send. So this leads to nondeterministic behavior - we can call `SET` with a key of length 10 KiB, and sometimes see that this doesn't trigger a zero-copy write since it is pieced into two `send` syscalls.

> I think this falls under the "configurations we do not want users to mess with".

At a higher level - can I ask why it would be bad to expose such a config to the user? I can imagine that a combination of network hardware and kernel version may impact what setting makes sense here.  I don't expect many folks to play with it - but it isn't a confusing setting and it isn't like maintenance cost will be high.

PR #1546, opened 2025-01-12
Add unsigned 64-bit numeric config in module API

artikell:
- Supports `VALKEYMODULE_CONFIG_UNSIGNED` and `UNSIGNED_CONFIG` for a wider range of configurations.
- Supported the API for `ValkeyModule_RegisterUnsignedNumericConfig`
- `string2ull` method supports the length parameter, making it more secure


madolson:
Is there a concrete need for this, or are you just looking to make the API more usable by extending the use cases? We had discussed previously in Redis, and had decided to avoid having a large number of APIs, we would only support signed until someone indicated they clearly needed a 64 bit unsigned value. You can always restrict the value to be positive and cast it to unsigned (but just miss some of the high values).

artikell:
> Is there a concrete need for this, or are you just looking to make the API more usable by extending the use cases? We had discussed previously in Redis, and had decided to avoid having a large number of APIs, we would only support signed until someone indicated they clearly needed a 64 bit unsigned value. You can always restrict the value to be positive and cast it to unsigned (but just miss some of the high values).

Yes, we encountered some scenarios.

In some designs, uint64 is used as the ID, and the higher bits are used to store some information. 
Logically, use `creatULongLongConfig` to manage configurations.

However, for the `creatULongLongConfig` method, it can only exceed ULLONG_MAX in the MEMORY_CONFIG configuration item. so, I think we need a configuration to support a larger range.

madolson:
> In some designs, uint64 is used as the ID, and the higher bits are used to store some information.
Logically, use creatULongLongConfig to manage configurations.

Hmm, interesting. OK, that makes sense to me.

artikell:
Do you have any suggestions regarding this PR? Maybe you can help review it? Thank you. @madolson

madolson:
Hey yes! Sorry, it wasn't on my list for 8.1, I'm just now catching up on other items. Will prioritize it tomorrow.

tests/modules/moduleconfigs.c
```diff
@@ -120,6 +145,9 @@ int registerBlockCheck(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc
 
     result = ValkeyModule_RegisterNumericConfig(ctx, "numeric", -1, VALKEYMODULE_CONFIG_DEFAULT, -5, 2000, getNumericConfigCommand, setNumericConfigCommand, longlongApplyFunc, &longval);
     response_ok |= (result == VALKEYMODULE_OK);
+    
+    result = ValkeyModule_RegisterUnsignedNumericConfig(ctx, "unsigned_numeric", 1, VALKEYMODULE_CONFIG_DEFAULT | VALKEYMODULE_CONFIG_UNSIGNED, 0, LLONG_MAX * 1ULL + 1000, getUnsignedNumericConfigCommand, setUnsignedNumericConfigCommand, unsignedlonglongApplyFunc, &mutable_ull_val);
```
madolson, line 149: I'm not sure why we need both the API and the new enum value. Can't RegisterUnsignedNumericConfig imply `VALKEYMODULE_CONFIG_UNSIGNED` ? We need the new API anyways because there are typed differences.

tests/modules/moduleconfigs.c
```diff
@@ -1,9 +1,12 @@
 #include "valkeymodule.h"
 #include <strings.h>
+#include <limits.h>
```
madolson, line 3: Is this used?

src/redismodule.h
```diff
@@ -691,6 +691,7 @@
 #define RedisModule_EventLoopAddOneShot ValkeyModule_EventLoopAddOneShot
 #define RedisModule_RegisterBoolConfig ValkeyModule_RegisterBoolConfig
 #define RedisModule_RegisterNumericConfig ValkeyModule_RegisterNumericConfig
+#define RedisModule_RegisterUnsignedNumericConfig ValkeyModule_RegisterUnsignedNumericConfig
```
madolson: We were planning on leaving these APIs to preserve the documented Redis module APIs, and not include new Valkey only APIs here.

src/config.c
```diff
@@ -2133,10 +2150,20 @@ static int numericParseString(standardConfig *config, sds value, const char **er
...
+        *res = strtoull(value, &endptr, 8);
         if (errno == 0 && *endptr == '\0') return 1; /* No overflow or invalid characters */
     }
 
+    /* Attempt to parse as an unsigned number */
+    if (config->data.numeric.flags & UNSIGNED_CONFIG) {
+        unsigned long long ull;
+        int ok = string2ull(value, sdslen(value), &ull);
+        if (ok) {
+            *res = (long long)ull;
```
madolson, line 2198: This feels weird, we're casting this to a ll and then casting it back. I originally thought this truncated the top bit, but apparently it preserves the bit pattern. I still think we maybe should have a dedicated function for handling unsigned numerics here.

tests/modules/moduleconfigs.c
```diff
@@ -103,6 +118,16 @@ int longlongApplyFunc(ValkeyModuleCtx *ctx, void *privdata, ValkeyModuleString *
     return VALKEYMODULE_OK;
 }
 
+int unsignedlonglongApplyFunc(ValkeyModuleCtx *ctx, void *privdata, ValkeyModuleString **err) {
+    VALKEYMODULE_NOT_USED(ctx);
+    VALKEYMODULE_NOT_USED(privdata);
+    if (mutable_ull_val == immutable_ull_val) {
+        *err = ValkeyModule_CreateString(NULL, "These configs cannot equal each other.", 38);
+        return VALKEYMODULE_ERR;
+    }
```
madolson, line 127: Is this functionality used anywhere?

madolson (commented):
Mostly looks good to me, just some comments about structure and if we can simplify the new API.

src/config.c
artikell: This is a reference to the implementation of memtoull in the MEMORY_CONFIG configuration. 
Do you mean a similar method needs to be implemented to cast the operation? Or add a numericParseString method for unsigned types?

tests/modules/moduleconfigs.c
artikell: Yes, the LLONG_MAX constant was used

tests/modules/moduleconfigs.c
artikell: That can be implemented, but it will introduce some difficult to understand issues because the parameters of the ValkeyModule_RegisterNumericConfig method are all of type long long. For example, min and max values.

What's your opinion?

src/redismodule.h
artikell: i get. I will remove it

tests/modules/moduleconfigs.c
artikell: Yes, this is a testing method used in the following logic

artikell:
> Mostly looks good to me, just some comments about structure and if we can simplify the new API.

In fact, unsigned longlong is special because its numerical range is the largest. For other types of data, it is not necessary to use an unsigned type, just control the data range.

Add an API that can also defend against a large number of type casts in encoding, avoiding various errors caused by them.

src/config.c
zuiderkwast: I think this is fine, but let's add some code comments about it. When parsing an unsigned number, although it's returned as a signed number, the caller needs to interpret it as an unsigned long long.

This is what getModuleUnsignedNumericConfig and other functions do.

zuiderkwast (approved):

PR #1549, opened 2025-01-12
Test coverage for ECHO for reply schema validation

zuiderkwast:
After #1545 disabled some tests for reply schema validation, we now have another issue that ECHO is not covered.

```
WARNING! The following commands were not hit at all:
  echo
ERROR! at least one command was not hit by the tests
```

(https://github.com/valkey-io/valkey/actions/runs/12728730819/job/35479652660)

This patch adds a test case for ECHO in the unit/other test suite. I haven't checked if there are more commands that aren't covered.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1220,7 +1220,7 @@ jobs:
     if: |
       (github.event_name == 'workflow_dispatch' ||
         (github.event_name == 'schedule' && github.repository == 'valkey-io/valkey') ||
-        (github.event_name == 'pull_request' && github.event.pull_request.base.ref != 'unstable')) &&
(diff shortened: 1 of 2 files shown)


enjoy-binbin (approved):

.github/workflows/daily.yml
```diff
@@ -1220,7 +1220,7 @@ jobs:
     if: |
       (github.event_name == 'workflow_dispatch' ||
         (github.event_name == 'schedule' && github.repository == 'valkey-io/valkey') ||
-        (github.event_name == 'pull_request' && github.event.pull_request.base.ref != 'unstable')) &&
+        (github.event_name == 'pull_request' && (contains(github.event.pull_request.labels.*.name, 'run-extra-tests') || github.event.pull_request.base.ref != 'unstable'))) &&
```
zuiderkwast, line 1223: @enjoy-binbin I was going to revert this change before merging. Sorry i didn't mention it.

I guess it's fine to run schema validator in run-extra-tests, but it takes one hour to run.

PR #1553, opened 2025-01-13
Introduce const_sds for const-content sds

### LINKED ISSUE #1542: Enforce `const` Where Applicable
it would be ideal if we express the "constness" of the returned sds string explicitly. Similar callouts to other helper functions returning plain pointers. That said, since we have never enforced "constness" in the past in our code base I suspect this could turn into quite a code churn. So definitely not a blocker for this PR (and it is primarily for code hygiene).

```
typedef const char* sds_const; // in src/sds.h
```
```suggestion
sds_const scriptingEngineGetName(scriptingEngine *engine) {
```

_Originally posted by @PingXie in https://github.com/valkey-io/valkey/pull/1312#discussion_r1902237016_

zuiderkwast:
`sds` is a typedef of `char *`.

`const sds` means `char * const`, i.e. a const-pointer to non-const content.

More often, you would want `const char *`, i.e. a pointer to const-content. Until now, it's not possible to express that. This PR adds `const_sds` which is a pointer to const-content sds.

To get a const-pointer to const-content sds, you can use `const const_sds`.

In this PR, some uses of `const sds` are replaced by `const_sds`. We can use it more later.

Fixes #1542

### DIFF
--- a/src/sds.c
+++ b/src/sds.c
@@ -187,19 +187,19 @@ sds sdsnew(const char *init) {
 }
 
 /* Duplicate an sds string. */
-sds sdsdup(const sds s) {
+sds sdsdup(const_sds s) {
     return sdsnewlen(s, sdslen(s));
 }
 
 /*
  * This method returns the minimum amount of bytes required to store the sds (header + data + NULL terminator).
  */
-static inline size_t sdsminlen(const sds s) {
+static inline size_t sdsminlen(const_sds s) {
     return sdslen(s) + sdsHdrSize(s[-1]) + 1;
 }
 
 /* This method copies the sds `s` into `buf` which is the target character buffer. */
-size_t sdscopytobuffer(unsigned char *buf, size_t buf_len, const sds s, uint8_t *hdr_size) {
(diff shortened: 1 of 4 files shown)


rjd15372 (approved):
LGTM

rainsupreme (approved):
I like this! Looks good to me as-is, but any thoughts on proactively spreading const_sds to other areas of code?

zuiderkwast:
> any thoughts on proactively spreading const_sds to other areas of code

When we have it, it's possible to use it in new code, so we can suggest it in code reviews.

Constness has to be added bottom-up, i.e. start in sds.c with functions like sdslen. A non-const value can always be used as a const value, but not vice versa.

We could add `const` in many other situations too, not related to sds. I don't think it's worth doing a lot of refactoring though just for that.

ranshid (approved):
LGTM even though I still think that the const should have been applied on the pointer as well like I commented in https://github.com/valkey-io/valkey/issues/1542
IMO we should not allow "assignment" to a const_sds. 

We can always evolve to doing this as well I guess

PR #1554, opened 2025-01-13
Escape unix socket group in unit tests

secwall:
In some cases unix groups could have whitespace and/or `\` in them.
One example is my workstation. It's a MacOS in an Active Directory domain. So my user has group `LD\Domain Users`.
Running `make test` on `unstable` and `8.0` branches fails with:
```
*** FATAL CONFIG FILE ERROR (Version 8.0.2) ***
Reading the configuration file, at line 30
>>> 'unixsocketgroup LD\Domain Users'
wrong number of arguments
```

I'm not sure if we need to fix this in 8.0. But it seems that it should be fixed in unstable.

### DIFF
--- a/tests/unit/other.tcl
+++ b/tests/unit/other.tcl
@@ -568,7 +568,8 @@ if {$::verbose} {
 set tempFileId [open $tempFileName w]
 set group [dict get [file attributes $tempFileName] -group]
 if {$group != ""} {
-    start_server [list tags {"repl external:skip"} overrides [list unixsocketgroup $group unixsocketperm 744]] {
+    set escaped_group "\"[string map {"\\" "\\\\"} $group]\""
+    start_server [list tags {"repl external:skip"} overrides [list unixsocketgroup $escaped_group unixsocketperm 744]] {
         test {test unixsocket options are set correctly} {
             set socketpath [lindex [r config get unixsocket] 1]
             set attributes [file attributes $socketpath]


madolson (approved):

madolson:
Agree, don't think we need to backport it.

tests/unit/other.tcl
```diff
@@ -568,7 +568,8 @@ if {$::verbose} {
 set tempFileId [open $tempFileName w]
 set group [dict get [file attributes $tempFileName] -group]
 if {$group != ""} {
-    start_server [list tags {"repl external:skip"} overrides [list unixsocketgroup $group unixsocketperm 744]] {
+    set escaped_group "\"[string map {"\\" "\\\\"} $group]\""
+    start_server [list tags {"repl external:skip"} overrides [list unixsocketgroup $escaped_group unixsocketperm 744]] {
```
zuiderkwast, line 572: Wouldn't it be more correct to do the escaping inside `start_server` so it works everywhere and for all options?

(`proc start_server` is defined in tests/support/server.tcl.)

secwall: Probably yes. Do we have a list of all string options that we should escape?
Could we try to escape any string option if we don't have such list?

zuiderkwast: I don't know. :) I think some investigation is needed to find out the best solution here. Now it's already merged so let's not worry too much about it.

Issue #274, opened 2024-04-09
[NEW] Redis-compatibility mode


zuiderkwast:
**The problem/use-case that the feature addresses**

We want to change the word "Redis" in error replies, log entries, HELLO reply, etc. but there is risk of breaking the Redis compatibility if we change them.

This is what we're talking about:

* #204.
  * "-LOADING Redis is loading the dataset in memory"
  * "-BUSY Redis is busy running a (...)"
  * "-MISCONF Redis is configured to (...)"
* #207. This config affects only two log messages:
  * "=== REDIS BUG REPORT START: Cut & paste starting from here ==="
  * "=== REDIS BUG REPORT END. Make sure to include from START to END. ==="

  We *don't* use this config for any other log entries such as

  * ~~"Redis needs to enable the AOF but can't trigger a background AOF rewrite operation. Check the above logs for more info about the error."~~
  * ~~"Redis forked for debugging eval"~~
  * ~~"Redis aborting for OUT OF MEMORY. Allocating %zu bytes!"~~
* #61
  * "server" => "redis", "version" => "...."
* INFO field (part of #41)
  * "redis_mode:standalone" (Change to "server_mode" or just "mode")

**Description of the feature**

Introduce a TEMPORARY "compatibility switch", so if a user has some scripts or clients hard-coded to look for "Redis", the user can enable the redis-compatibility mode.

I say temporary, because the purpose is to let users fix their applications.

:question: What's the time plan? I suggest we add the switch in Valkey 8. In Valkey 9 we remove it, but we make the config have no effect, so we don't crash if the user has the config. In Valkey 10 we remove the config.

:question: What shall be controlled by it? I suggest the things listed above. For all other error replies and log messages we can just make a breaking change. Other suggestions?

:question: Shall it be a compile time switch (ifdef) or runtime (config). I suggest a config. Not all users compile from source.

**Alternatives you've considered**

We could just do breaking changes. (High risk of problems for users?)

We could just keep Redis in these messages forever. (Safe, but not nice.)

**Additional information**

It has been discussed but no decision was made yet. We need to make decisions about all " :question: " above.

hwware:
Do we need consider TCL test case redis-compatibility in the thread? Or I need open a new one for test case?

zuiderkwast:
@hwware i don't understand. Why do we need to change test cases? Just to make CI pass?

hwware:
> @hwware i don't understand. Why do we need to change test cases? Just to make CI pass?

[...]

PR #1559, opened 2025-01-14
In LOLWUT's reply, change "Redis ver." to "Valkey ver."

sarthakaggarwal97:
Make LOLWUT it print the current Valkey version rather than the Redis version. Retain the old behaviour and print the compatible Redis version if the `extended-redis-compat` config is enabled.

Additionally, change "Redis" to "Valkey" in an error message "This Redis command is not allowed from script", also depending on the `extended-redis-compat` config.

### DIFF
--- a/src/lolwut.c
+++ b/src/lolwut.c
@@ -44,7 +44,7 @@ void lolwut6Command(client *c);
  * This is what unstable versions of the server will display. */
 void lolwutUnstableCommand(client *c) {
     sds rendered = sdscatprintf(sdsempty(), "%s ver.", server.extended_redis_compat ? "Redis" : "Valkey");
-    rendered = sdscat(rendered, VALKEY_VERSION);
+    rendered = sdscat(rendered, server.extended_redis_compat ? REDIS_VERSION : VALKEY_VERSION);
     rendered = sdscatlen(rendered, "\n", 1);
     addReplyVerbatim(c, rendered, sdslen(rendered), "txt");
     sdsfree(rendered);
--- a/src/lolwut.c
+++ b/src/lolwut.c
@@ -43,7 +43,7 @@ void lolwut6Command(client *c);
 /* The default target for LOLWUT if no matching version was found.
  * This is what unstable versions of the server will display. */
 void lolwutUnstableCommand(client *c) {
-    sds rendered = sdsnew("Valkey ver. ");
+    sds rendered = sdscatprintf(sdsempty(), "%s ver.", server.extended_redis_compat ? "Redis" : "Valkey");
     rendered = sdscat(rendered, VALKEY_VERSION);
     rendered = sdscatlen(rendered, "\n", 1);
     addReplyVerbatim(c, rendered, sdslen(rendered), "txt");
--- a/src/lolwut.c
+++ b/src/lolwut.c
@@ -43,7 +43,7 @@ void lolwut6Command(client *c);
 /* The default target for LOLWUT if no matching version was found.
  * This is what unstable versions of the server will display. */
 void lolwutUnstableCommand(client *c) {
-    sds rendered = sdsnew("Redis ver. ");
+    sds rendered = sdsnew("Valkey ver. ");
     rendered = sdscat(rendered, VALKEY_VERSION);
     rendered = sdscatlen(rendered, "\n", 1);
     addReplyVerbatim(c, rendered, sdslen(rendered), "txt");
--- a/src/script.c
+++ b/src/script.c
@@ -548,7 +548,7 @@ void scriptCall(scriptRunCtx *run_ctx, sds *err) {
 
     /* There are commands that are not allowed inside scripts. */
     if (!server.script_disable_deny_script && (cmd->flags & CMD_NOSCRIPT)) {
-        *err = sdsnew("This Redis command is not allowed from script");
+        *err = sdscatprintf(sdsempty(), "This %s command is not allowed from script", server.extended_redis_compat ? "Redis" : "Valkey");
         goto error;
     }
 
--- a/tests/unit/other.tcl
+++ b/tests/unit/other.tcl
@@ -385,6 +385,8 @@ start_server {tags {"other"}} {
             set info [r info server]
             assert_match "*redis_mode:*" $info
             assert_no_match "*server_mode:*" $info
+            set lolwut_output [r lolwut]
+            assert_match {*Redis ver.*} $lolwut_output
             r config set extended-redis-compatibility no
             set hello [r hello 3]
             assert_equal "valkey" [dict get $hello server]
@@ -392,6 +394,8 @@ start_server {tags {"other"}} {
             set info [r info server]
             assert_no_match "*redis_mode:*" $info
             assert_match "*server_mode:*" $info
+            set lolwut_output [r lolwut]
+            assert_match {*Valkey ver.*} $lolwut_output
         }
     }
 }
--- a/tests/unit/other.tcl
+++ b/tests/unit/other.tcl
@@ -385,6 +385,10 @@ start_server {tags {"other"}} {
             set info [r info server]
             assert_match "*redis_mode:*" $info
             assert_no_match "*server_mode:*" $info
+            set lolwut_output [r lolwut 5]
+            assert_match {*Redis ver.*} $lolwut_output
+            set lolwut_output [r lolwut 6]
+            assert_match {*Redis ver.*} $lolwut_output
             set lolwut_output [r lolwut]
             assert_match {*Redis ver.*} $lolwut_output
             r config set extended-redis-compatibility no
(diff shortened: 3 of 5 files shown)


sungming2 (approved):

enjoy-binbin (approved):

src/lolwut.c
```diff
@@ -43,7 +43,7 @@ void lolwut6Command(client *c);
 /* The default target for LOLWUT if no matching version was found.
  * This is what unstable versions of the server will display. */
 void lolwutUnstableCommand(client *c) {
-    sds rendered = sdsnew("Redis ver. ");
+    sds rendered = sdsnew("Valkey ver. ");
```
zuiderkwast:
```suggestion
    sds rendered = sdsnew(SERVER_TITLE " ver. ");
```

zuiderkwast (commented):
Good, but we can avoid hard-coding the name.

(It's weird that we have SERVER_TITLE but we decided to name the version variable VALKEY_VERSION instead of SERVER_VERSION. Maybe there was a reason for that. I don't remember the discussion.)

ranshid:
Trying to understand the issue. AFAIK changing the lolwut can break some clients using it in order to get the current verison.
I agree that providing the unstable version is problematic and we might want to do the same as we did with the info server version (i.e use fixed 7.2.4). 

The other issue is that we also have the "Easter eggs" lolwat (which are somewhat broken, for example you cannot enter lolwut which are also presenting the VALKEY_VERSION version and Redis brand (I wonder if these we can be changed).

In anycase I think this should be marked as a breaking change

src/lolwut.c
madolson: I thought we left this intentionally because it's technically breaking change. Why do we need to change it?

zuiderkwast: I guess because seeing "Redis" anywhere sticks out. :D If we want to keep LOLWUT without updates, then we should probably instead return REDIS_VERSION rather than VALKEY_VERSION?

sarthakaggarwal97: > I guess because seeing "Redis" anywhere sticks out.

Yeah, that's why I raised the change. I wasn't aware that we left this out intentionally.
Do you think that we should take this up in the next major version? 

Also, we kinda still use it here as well: https://github.com/valkey-io/valkey/blob/unstable/src/networking.c#L4277

zuiderkwast: Oh, I think we missed those security warning logs. We decided during the initial rebranding work that we don't need compatibility for log messages, so I guess it's OK to change them.

I did a quick grep and I found one more occurrence in an error message in `script.c`: `"This Redis command is not allowed from script"`. My first thought is that I don't think it's risky to change it, but if we think it is, then we could let it depend on the extended-redis-compat configuration.

sarthakaggarwal97: > We decided during the initial rebranding work that we don't need compatibility for log messages, so I guess it's OK to change them.

I guess I can raise another PR for this, if that's okay. If we decide on changing in the `script.c` as well, I am happy to do it as well.

zuiderkwast: @madolson @ranshid WDYT?

ranshid: For the lolwut command I would suggest to keep the Redis and fixed version (7.2.4). I would also suggest that we deprecate the lolwut and setup a path forward to remove support for it. (BTW we should do the same for the version 5/6 lolwut IMO)

For the security logs, I have no real concern as this is reported in logs. I do wonder if this cross-application-scripting is really a security concern? maybe we should just remove this whole thing?

ranshid:
@sarthakaggarwal97 I just looked. seems like @zuiderkwast suggestion to control this feature via extended-redis-compat might be fine.

sarthakaggarwal97:
@ranshid @zuiderkwast if I understand this correctly, extended-redis-compat is stemming from https://github.com/valkey-io/valkey/issues/274 right?

I see we have a configuration available  in `valkey.conf` as `extended-redis-compatibility`. So we can continue with this change relying on this configuration?

zuiderkwast:
> @ranshid @zuiderkwast if I understand this correctly, extended-redis-compat is stemming from #274 right?
> 
> I see we have a configuration available in `valkey.conf` as `extended-redis-compatibility`. So we can continue with this change relying on this configuration?

That's right. I think it's OK to use this config for lolwut. We should make a TSC decision about it though.

I remember when we added this config, there were some concerns that it will give us a lot of extra maintenance work, so we decided to limit the scope to only a few things and to make it temporary and be phased out in 9.0. See the test "Extended Redis Compatibility config" in tests/unit/other.tcl

hwware (approved):

src/lolwut.c
```diff
@@ -43,7 +43,7 @@ void lolwut6Command(client *c);
 /* The default target for LOLWUT if no matching version was found.
  * This is what unstable versions of the server will display. */
 void lolwutUnstableCommand(client *c) {
-    sds rendered = sdsnew("Valkey ver. ");
+    sds rendered = sdscatprintf(sdsempty(), "%s ver.", server.extended_redis_compat ? "Redis" : "Valkey");
     rendered = sdscat(rendered, VALKEY_VERSION);
```
zuiderkwast: @ranshid wrote

> For the lolwut command I would suggest to keep the Redis and fixed version (7.2.4).

I think this is still a good idea to follow for the compatibility mode, so when we print "Redis ver." use REDIS_VERSION and when we print "Valkey ver." use VALKEY_VERSION.

REDIS_VERSION is the open source redis version we are compatible with.

zuiderkwast: This means in 8.1 we would print either "Redis ver. 7.2.4" or "Valkey ver. 8.1.0".

If we would print "Redis ver. 8.1.0" it would be just wrong.

sarthakaggarwal97: makes sense @zuiderkwast, I have updated it!

ranshid:
Will just ask again:
why not rebrand the `lolwut5Command` and `lolwut6Command`?

sarthakaggarwal97:
@ranshid apologies, but could you please elaborate on the rebranding on commands? What exactly would be required here?

PingXie:
>I think it's OK to use this config for lolwut. We should make a TSC decision about it though.

+1

zuiderkwast:
We discussed it in a meeting today and decided to use the extended redis compat config for this as implemented in this PR, but the same change needs to be done in `lolwut5.c` and `lolwut6.c` too. Waiting for this change before merging.

sarthakaggarwal97:
@zuiderkwast added extended redis compat config for lolwut5, lolwut6 as well. Please check once!

zuiderkwast (approved):
LGTM, thanks.

@ranshid Please take another look.

tests/unit/other.tcl
```diff
@@ -385,13 +385,17 @@ start_server {tags {"other"}} {
             set info [r info server]
             assert_match "*redis_mode:*" $info
             assert_no_match "*server_mode:*" $info
+            set lolwut_output [r lolwut]
+            assert_match {*Redis ver.*} $lolwut_output
```
enjoy-binbin, line 393:
```suggestion
            set lolwut_output [r lolwut 5]
            assert_match {*Redis ver.*} $lolwut_output
            set lolwut_output [r lolwut 6]
            assert_match {*Redis ver.*} $lolwut_output
            set lolwut_output [r lolwut]
            assert_match {*Redis ver.*} $lolwut_output
```

tests/unit/other.tcl
```diff
@@ -385,13 +385,17 @@ start_server {tags {"other"}} {
...
             r config set extended-redis-compatibility no
             set hello [r hello 3]
             assert_equal "valkey" [dict get $hello server]
             assert_equal $version [dict get $hello version]
             set info [r info server]
             assert_no_match "*redis_mode:*" $info
             assert_match "*server_mode:*" $info
+            set lolwut_output [r lolwut]
+            assert_match {*Valkey ver.*} $lolwut_output
```
enjoy-binbin, line 402:
```suggestion
            set lolwut_output [r lolwut 5]
            assert_match {*Valkey ver.*} $lolwut_output
            set lolwut_output [r lolwut 6]
            assert_match {*Valkey ver.*} $lolwut_output
            set lolwut_output [r lolwut]
            assert_match {*Valkey ver.*} $lolwut_output
```

enjoy-binbin (approved):

enjoy-binbin:
please also update the top comment.

hpatro (approved):

sarthakaggarwal97:
@enjoy-binbin added those suggestions!

zuiderkwast:
> please also update the top comment.

Done. Now describing the exact behavior with the config. @sarthakaggarwal97 please note.

sarthakaggarwal97:
@zuiderkwast ah sorry, I got confused. I thought it was just for the suggestion. thank you.

src/script.c
```diff
@@ -548,7 +548,7 @@ void scriptCall(scriptRunCtx *run_ctx, sds *err) {
 
     /* There are commands that are not allowed inside scripts. */
     if (!server.script_disable_deny_script && (cmd->flags & CMD_NOSCRIPT)) {
-        *err = sdsnew("This Redis command is not allowed from script");
+        *err = sdscatprintf(sdsempty(), "This %s command is not allowed from script", server.extended_redis_compat ? "Redis" : "Valkey");
```
zuiderkwast, line 551: I think we should mention this change too in the top comment.

Issue #1558, opened 2025-01-14
Create development guide


madolson:
I'm just starting an issue to dump all of the conversations we've had about getting started as a developer on Valkey. At some point we should commit them to a markdown, but just wanted to start taking notes since I've seen a few examples of best practices/naming conventions/common pitfalls.

### General coding guidelines. 
1. Most of the style comments are enforced by clang format.
2. C style comments `/* comment */` can be used for both single and multi-line comments. C++ comments `//` can only be used for single line comments.

### Naming conventions
Valkey has a long history of inconsistent naming conventions. Generally follow the style of the surrounding code, but you can also always use the following conventions for variable and structure names:
  * Variable names: snake_case or all lower case (valkeyobject or valkey_object)
  * Function names: camelCase or namespace_camelCase (createObjectList or networking_createObjectList).
  * Macros: UPPER_CASE (DICT_CREATE)
  * structures: camelCase (user)

### Best practices
1. Avoid adding configuration when a feature can be fully controlled by heuristics. We want Valkey to work correctly out of the box without much tuning. When the workload characteristics can't be inferred or imply a tradeoff (CPU vs memory), then provide a configuration.

sarthakaggarwal97:
We can also include the minimum test / functional coverage required, and how can we write / run TCL tests (maybe)?

ranshid:
Some things that came to mind:
1. use of static functions when possible and internal to the object
2. architecture specific definitions: where to place them/ which architectures should be supported/ which compiler flavors etc...
3. regarding configurations, maybe we should help programers decide when feature control should be supported by config or debug command (eg tests)
4. makefile/Cmake conventions (?)

pieturin:
We should specify that we want to keep line lengths to max 120 characters.

enjoy-binbin:
> We should specify that we want to keep line lengths to max 120 characters.

90 chars? i think 120 is still quite long and hard to read the long sentence for me (bad at reading)

enjoy-binbin:
also, i think we should mention the backport thing? How do we do the backport now? Do we do the backport when a PR got merged like #1575? Or we just mark a label and when the release date come, someone from the trusted list do the local cherry-pick thing and push they together?

hpatro:
[...]

PR #1561, opened 2025-01-14
Update comments and log message in cluster_legacy.c

pieturin:
Update comments and log message in `cluster_legacy.c`.

Follow-up from #1441.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3278,8 +3278,8 @@ int clusterProcessPacket(clusterLink *link) {
                         /* Unable to retrieve the node's IP address from the connection. Without a
                          * valid IP, the node becomes unusable in the cluster. This failure might be
                          * due to the connection being closed. */
-                        serverLog(LL_NOTICE, "Closing link even though we received a MEET packet on it, "
-                                             "because the connection has an error");
+                        serverLog(LL_NOTICE, "Closing cluster link due to failure to retrieve IP from the connection, "
+                                             "possibly caused by a closed connection.");
                         freeClusterLink(link);
                         return 0;
                     }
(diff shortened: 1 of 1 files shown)


pieturin:
@PingXie, could you take a look please? This should address the last comments you had in my previous PR.

src/cluster_legacy.c
```diff
@@ -3278,8 +3278,8 @@ int clusterProcessPacket(clusterLink *link) {
                         /* Unable to retrieve the node's IP address from the connection. Without a
                          * valid IP, the node becomes unusable in the cluster. This failure might be
                          * due to the connection being closed. */
-                        serverLog(LL_NOTICE, "Closing link even though we received a MEET packet on it, "
-                                             "because the connection has an error");
+                        serverLog(LL_NOTICE, "Closing cluster link due to failure to retrieve IP from the connection, "
```
enjoy-binbin, line 3281: lets just change the log if needed? the comment changes seem unnecessary from the current perspective.

pieturin: Updating the comments was an ask from @PingXie, but I can remove those changes if you prefer.

enjoy-binbin: yes, let drop the comments changes

PingXie: any specific concerns about the comment changes, @enjoy-binbin? my intent was for readability but curious to hear your concerns. note that these comments are newly added too.

enjoy-binbin: we still have many rows out there that does not meet the 100 chars condition. Thought these comments are newly code, it is stll got merged before, this does not make much different compare the old code, the git blame log is still the main concern for me.

PingXie: I see. I think git blame should be a secondary concern now. at least in my setup (vim), navigating through commits is not an issue. I would assume other editors like vscode would be even more friendlier. that said, in this particular case, the change is all about reflowing, I am fine reverting it too.

PingXie (approved):

src/cluster_legacy.c
enjoy-binbin: It will be much comfortable if we do it before the merge, sadly Wen merged it without checking that all open comments were closed, and he didn't even review the code, we should at least wait for the author to become available or for us to take over and insert the suggestion before merging. I must say this somehow bother me

src/cluster_legacy.c
```diff
@@ -3318,12 +3318,14 @@ int clusterProcessPacket(clusterLink *link) {
...
+                 * thinks that I do not know it.
+                 * Free my outbound link to that node, triggering a reconnect and a PING over the
+                 * new link.
+                 * Once that node receives our PING, it should recognize the new connection as an
+                 * inbound link from me. We should only free the outbound link if the node is known
+                 * for more time than the handshake timeout, since during this time, the other side
+                 * might still be trying to complete the handshake. */
 
                 /* We should always receive a MEET packet on an inbound link. */
                 serverAssert(link != sender->link);
```
enjoy-binbin, line 3331: i am ok with this change since the line is indeed quite long, we should add this in #1558.

Also if we are doing it, please also handle line 3332, it is long too.
(i usually keep the line at around 85 in person.)

enjoy-binbin (approved):

PR #1564, opened 2025-01-15
Add paused_reason to INFO CLIENTS

enjoy-binbin:
In #1519, we added paused_actions and paused_timeout_milliseconds,
it would be helpful if we add the paused_purpose since users also
want to know the purpose for the pause.

Currently available options:
- client_pause: trigger by CLIENT PAUSE command.
- shutdown_in_progress: during shutdown, primary waits the replicas to catch up the offset.
- failover_in_progress: during failover, primary waits the replica to catch up the offset.
- none


soloestoy (approved):

soloestoy (approved):

src/server.h
```diff
@@ -2714,7 +2714,8 @@ void pauseActions(pause_purpose purpose, mstime_t end, uint32_t actions);
 void unpauseActions(pause_purpose purpose);
 uint32_t isPausedActions(uint32_t action_bitmask);
 uint32_t isPausedActionsWithUpdate(uint32_t action_bitmask);
-mstime_t getPausedActionTimeout(uint32_t action);
+char *getPausedPurposeString(pause_purpose purpose);
```
hwware: I suggest removing the String from function name getPausedPurposeString(), char* already represents the return type.  I know in server.h, some function names keep the String even the return type is char*, and some do not.  Let us keep simple and normalize it.

enjoy-binbin: updated.

hwware (approved):

hwware:
@valkey-io/core-team Please vote for this, Thanks a lot

src/networking.c
```diff
@@ -4520,12 +4520,30 @@ void flushReplicasOutputBuffers(void) {
...
 
-mstime_t getPausedActionTimeout(uint32_t action) {
+char *getPausedPurpose(pause_purpose purpose) {
+    switch (purpose) {
+    case PAUSE_BY_CLIENT_COMMAND:
+        return "client_command";
+    case PAUSE_DURING_SHUTDOWN:
+        return "during_shutdown";
+    case PAUSE_DURING_FAILOVER:
+        return "during_failover";
```
zuiderkwast: I like the INFO field, but I think the name and values are not perfect. There's something about the English language that doesn't sound right to me.

```
paused_purpose:client_command
paused_purpose:during_shutdown
paused_purpose:during_failover
```

I think the word "purpose" is not correct. Purpose is not what caused something to happen. The purpose is the end goal of some action, not the source.

To describe the source of the pause, I think "cause" or "reason" are better words. "paused_cause" or "paused_reason"?

And "during" is about time. It is the answer to when, not why. Maybe we can find a better word or just skip "during"?

How about "paused_reason:shutdown"? Or "paused_reason:shutdown_in_progress" or something like that?

enjoy-binbin: yean, paused_reason seems better, updated.

zuiderkwast (approved):
Good, thanks.

@soloestoy @hwware You already approved. Do you agree with the changed field name and values in the last update?

hwware:
> Good, thanks.
> 
> @soloestoy @hwware You already approved. Do you agree with the changed field name and values in the last update?

I am fine with them

madolson (approved):

zuiderkwast:
@enjoy-binbin Let's not forget the docs before 8.1.

PR #1566, opened 2025-01-15
Call makeObjectShared to normalize a shared object into a true shared object

enjoy-binbin:
Currently, only int and shared.redacted call makeObjectShared for
shared objects.

Although this breaks the blame log, there is a few points i would
like to mention:
1. For threading, our fork for some reasons calling decrRefCount in
   the threads for shared.command. Although this is not used in OSS
   code now, it may be a issue in the future.
2. For cleanup, all shared objects's refcount should be the same.
3. For CoW, probably can reduce CoW although it is very minor.

We decided in #2189 to keep `extended-redis-compatibility` for now.


src/server.c
```diff
@@ -1964,153 +1964,153 @@ void afterSleep(struct aeEventLoop *eventLoop, int numevents) {
  * initializations can be moved back inside createSharedObjects() below. */
 void createSharedObjectsWithCompat(void) {
     const char *name = server.extended_redis_compat ? "Redis" : SERVER_TITLE;
-    if (shared.loadingerr) decrRefCount(shared.loadingerr);
```
enjoy-binbin, line 2003: I'm struggling with this, not sure how to handle it yet.

zuiderkwast: It's a little sad that we need to store two versions of these strings and need to check the compat flag every time we return these strings. We have already planned to remove this compat config at some time.

If we can make extended-redis-compat a readonly config, we can initialize them only once, but that's not perfect. One of the purposes is that users can change them in runtime if they have some problem.

Another idea: These strings are changed only when the compat config is changed. Is it possible to wait for all threads to stop and then replace the shared objects and free the old ones? We do something similar for MODULE UNLOAD to make sure no module is accessing a command that is going to be removed:

```C
void moduleUnregisterCommands(struct ValkeyModule *module) {
    /* Drain IO queue before modifying commands dictionary to prevent concurrent access while modifying it. */
    drainIOThreadsQueue();
```

WDYT?

enjoy-binbin: lazyfree job does not fit in the I/O threads, unforunately

src/server.c
```diff
@@ -2027,126 +2022,164 @@ void createSharedObjectsWithCompat(void) {
                                            name, name));
 }
 
+void createSharedObjectsWithCompat(void) {
```
enjoy-binbin: It is not perfect, we can delete it together when we remove extended-redis-compatibility.

zuiderkwast: Can we just create all the strings with duplicated code? It is somehow less complicated.

```C
shared.loadingerr_valkey = makeObjectShared(createObject(
    OBJ_STRING, sdscatfmt(sdsempty(), "-LOADING %s is loading the dataset in memory\r\n", SERVER_NAME));
...
shared.loadingerr_redis = makeObjectShared(createObject(
    OBJ_STRING, sdsnew("-LOADING Redis is loading the dataset in memory\r\n"));
...
```

enjoy-binbin: yes, we can, it is indeed less complicated, do you prefer this? 

My thinking is that we can just remove the whole block without changing `shared.loadingerr` again in the future.

I am ok with both way, i don't have much preference.

src/server.c
```diff
@@ -1998,25 +1998,20 @@ void afterSleep(struct aeEventLoop *eventLoop, int numevents) {
 /* These shared strings depend on the extended-redis-compatibility config and is
  * called when the config changes. When the config is phased out, these
  * initializations can be moved back inside createSharedObjects() below. */
-void createSharedObjectsWithCompat(void) {
-    const char *name = server.extended_redis_compat ? "Redis" : SERVER_TITLE;
-    if (shared.loadingerr) decrRefCount(shared.loadingerr);
+void createSharedObjectsForCompat(int compat) {
+    const char *name = compat ? "Redis" : SERVER_TITLE;
     shared.loadingerr =
         createObject(OBJ_STRING, sdscatfmt(sdsempty(), "-LOADING %s is loading the dataset in memory\r\n", name));
```
zuiderkwast: Do `makeObjectShared` on these too?

src/server.c
```diff
@@ -2027,126 +2022,164 @@ void createSharedObjectsWithCompat(void) {
...
-    shared.wrongtypeerr =
-        createObject(OBJ_STRING, sdsnew("-WRONGTYPE Operation against a key holding the wrong kind of value\r\n"));
-    shared.err = createObject(OBJ_STRING, sdsnew("-ERR\r\n"));
-    shared.nokeyerr = createObject(OBJ_STRING, sdsnew("-ERR no such key\r\n"));
-    shared.syntaxerr = createObject(OBJ_STRING, sdsnew("-ERR syntax error\r\n"));
-    shared.sameobjecterr = createObject(OBJ_STRING, sdsnew("-ERR source and destination objects are the same\r\n"));
-    shared.outofrangeerr = createObject(OBJ_STRING, sdsnew("-ERR index out of range\r\n"));
-    shared.noscripterr = createObject(OBJ_STRING, sdsnew("-NOSCRIPT No matching script.\r\n"));
+    shared.wrongtypeerr = makeObjectShared(createObject(
+            OBJ_STRING, sdsnew("-WRONGTYPE Operation against a key holding the wrong kind of value\r\n")));
```
zuiderkwast: If we anyway touch these lines, maybe we can add a small function `createSharedString` to do this. It makes the code smaller:

```suggestion
    shared.wrongtypeerr = createSharedString("-WRONGTYPE Operation against a key holding the wrong kind of value\r\n");
```

The implementation can be static inline, very small:

```c
static inline robj *createSharedString(const char *string) {
    return makeObjectShared(createObject(OBJ_STRING, sdsnew(string));
}

File: src/server.c (hunk 2 above)
Comment (zuiderkwast) [reply src/server.c]: Yeah I guess I prefer it. It's still easy to change it. But I guess we should discuss if we should delete it or not... If we want to delete it, we can do it already now.

File: src/server.c (hunk 4 above)
Comment (enjoy-binbin) [reply src/server.c]: it bother me there are some different usage:
```
    shared.wrongtypeerr = makeObjectShared(createObject(
	            OBJ_STRING, sdsnew("-WRONGTYPE Operation against a key holding the wrong kind of value\r\n")));

shared.messagebulk = makeObjectShared(createStringObject("$7\r\nmessage\r\n", 13));

        shared.mbulkhdr[j] = makeObjectShared(createObject(OBJ_STRING, sdscatprintf(sdsempty(), "*%d\r\n", j)));
```

Use createStringObject seem better since it can handle OBJ_ENCODING_EMBSTR. I want to unify them, but i also don't want to touch them on the other side.

File: src/server.c [hunk 5]
```diff
@@ -1995,170 +1995,210 @@ void afterSleep(struct aeEventLoop *eventLoop, int numevents) {
...
+    shared.slowmoduleerr_valkey =
+        createSharedStringByPtr(sdscatfmt(sdsempty(), "-BUSY %s is busy running a module command.\r\n", name));
+    shared.bgsaveerr_valkey =
+        createSharedStringByPtr(sdscatfmt(sdsempty(),
+                                          "-MISCONF %s is configured to save RDB snapshots, but it's currently"
+                                          " unable to persist to disk. Commands that may modify the data set are"
+                                          " disabled, because this instance is configured to report errors during"
+                                          " writes if RDB snapshotting fails (stop-writes-on-bgsave-error option)."
+                                          " Please check the %s logs for details about the RDB error.\r\n",
+                                          name, name));
```
zuiderkwast: Oh... I didn't realize we would duplicate large texts like this. It's not perfect....

:thinking: Is there any way to do this without duplicating a lot of code?

Maybe we can put these shared variants in arrays indexed by compat mode 0 or 1. Then we can have just one function to initialize these. We call it twice, with `compat=0 name=SERVER_NAME` and with `compat=1 name=Redis`.

```suggestion
    shared.slowmoduleerr_variants[is_compat] =
        createSharedStringByPtr(sdscatfmt(sdsempty(), "-BUSY %s is busy running a module command.\r\n", name));
    shared.bgsaveerr_variants[is_compat] =
        createSharedStringByPtr(sdscatfmt(sdsempty(),
                                          "-MISCONF %s is configured to save RDB snapshots, but it's currently"
                                          " unable to persist to disk. Commands that may modify the data set are"
                                          " disabled, because this instance is configured to report errors during"
                                          " writes if RDB snapshotting fails (stop-writes-on-bgsave-error option)."
                                          " Please check the %s logs for details about the RDB error.\r\n",
                                          name, name));
```

src/server.c
```diff
@@ -1995,170 +1995,210 @@ void afterSleep(struct aeEventLoop *eventLoop, int numevents) {
 
 /* =========================== Server initialization ======================== */
 
+static inline robj *createSharedString(const char *string) {
+    return makeObjectShared(createStringObject(string, strlen(string)));
+}
+
+static inline robj *createSharedStringByPtr(void *ptr) {
+    return makeObjectShared(createObject(OBJ_STRING, ptr));
```
zuiderkwast: ptr is sds, right?

```suggestion
static inline robj *createSharedStringBySds(sds *s) {
    return makeObjectShared(createObject(OBJ_STRING, s));
```

src/server.c
enjoy-binbin: yes, the previous commit is to call the function twice to create the robj. Great, the array thing is a good idea.

src/server.c
```diff
@@ -1999,48 +1999,26 @@ static inline robj *createSharedString(const char *string) {
     return makeObjectShared(createStringObject(string, strlen(string)));
 }
 
-static inline robj *createSharedStringByPtr(void *ptr) {
-    return makeObjectShared(createObject(OBJ_STRING, ptr));
+static inline robj *createSharedStringByPtr(sds s) {
```
zuiderkwast: ByPtr, rename to BySds or FromSds?

It takes ownership of the sds, so it can be seen as a function to convert from sds to robj. Therefore maybe FromSds is better?

enjoy-binbin: I suddenly realized that i used ptr because of this...
```
    for (j = 0; j < OBJ_SHARED_INTEGERS; j++) {
        shared.integers[j] = createSharedStringByPtr((void *)(long)j);
        initObjectLRUOrLFU(shared.integers[j]);
        shared.integers[j]->encoding = OBJ_ENCODING_INT;
    }
```

zuiderkwast: Ah.

Is this the only place we don't pass an sds? If it's only one place, we can keep the old implementation without this new function.

Btw, we does this call initObjectLRUOrLFU? It's a no-op for shared objects.

```c
    for (j = 0; j < OBJ_SHARED_INTEGERS; j++) {
        shared.integers[j] = makeObjectShared(createObject(OBJ_STRING, (void *)(long)j));
        shared.integers[j]->encoding = OBJ_ENCODING_INT;
    }
```

enjoy-binbin: yes, that is true, updated.

src/server.c
```diff
@@ -1995,170 +1995,180 @@ void afterSleep(struct aeEventLoop *eventLoop, int numevents) {
...
+static inline robj *createSharedString(const char *string) {
+    return makeObjectShared(createStringObject(string, strlen(string)));
+}
+
+static inline robj *createSharedStringFromSds(sds s) {
+    return makeObjectShared(createObject(OBJ_STRING, s));
+}
+
 /* These shared strings depend on the extended-redis-compatibility config and is
  * called when the config changes. When the config is phased out, these
```
zuiderkwast: This comment is outdated now. Maybe just delete this sentence that says it's called when the config changes.

zuiderkwast (approved):

zuiderkwast (approved):

PR #1568, opened 2025-01-15
Accelerate hash table iterator with value prefetching

NadavGigi:
This PR builds upon the [previous entry prefetching optimization](https://github.com/valkey-io/valkey/pull/1501) to further enhance performance by implementing value prefetching for hashtable iterators.
## Implementation
Modified `hashtableInitIterator` to accept a new flags parameter, allowing control over iterator behavior.
Implemented conditional value prefetching within `hashtableNext` based on the new `HASHTABLE_ITER_PREFETCH_VALUES` flag.
When the flag is set, hashtableNext now calls `prefetchBucketValues` at the start of each new bucket, preemptively loading the values of filled entries into the CPU cache.
The actual prefetching of values is performed using type-specific callback functions implemented in `server.c`:
- For `robj` the `hashtableObjectPrefetchValue` callback is used to prefetch the value if not embeded.

This implementation is specifically focused on main database iterations at this stage. Applying it to hashtables that hold other object types should not be problematic, but its performance benefits for those cases will need to be proven through testing and benchmarking.

#### Setup:
- 64cores Graviton 3 Amazon EC2 instance.
-  50 mil keys with different value sizes.
-  Running valkey server over RAM file system.
-  crc checksum and comperssion off.

#### Action
- save command.

#### Results
The results regarding the duration of “save” command was taken from “info all” command.
```
+--------------------+------------------+------------------+ 
| Prefetching        | Value size (byte)| Time (seconds)   | 
+--------------------+------------------+------------------+ 
| No                 | 100              | 20.112279        | 
| Yes                | 100              | 12.758519        | 
| No                 | 40               | 16.945366        | 
| Yes                | 40               | 10.902022        |
| No                 | 20               | 9.817000         | 
| Yes                | 20               | 9.626821         |
| No                 | 10               | 9.71510          | 
| Yes                | 10               | 9.510565         |
+--------------------+------------------+------------------+
```
The results largely align with our expectations, showing significant improvements for larger values (100 bytes and 40 bytes) that are stored outside the robj. For smaller values (20 bytes and 10 bytes) that are embedded within the robj, we see almost no improvement, which is as expected.

However, the small improvement observed even for these embedded values is somewhat surprising. Given that we are not actively prefetching these embedded values, this minor performance gain was not anticipated.

perf record on save command **without** value prefetching:
```
                --99.98%--rdbSaveDb
                          |          
                          |--91.38%--rdbSaveKeyValuePair
                          |          |          
                          |          |--42.72%--rdbSaveRawString
                          |          |          |          
                          |          |          |--26.69%--rdbWriteRaw
                          |          |          |          |          
                          |          |          |           --25.75%--rioFileWrite.lto_priv.0
                          |          |          |          
                          |          |           --15.41%--rdbSaveLen
                          |          |                     |          
                          |          |                     |--7.58%--rdbWriteRaw
                          |          |                     |          |          
                          |          |                     |           --7.08%--rioFileWrite.lto_priv.0
                          |          |                     |                     |          
                          |          |                     |                      --6.54%--_IO_fwrite
                          |          |                     |                                         
                          |          |                     |          
                          |          |                      --7.42%--rdbWriteRaw.constprop.1
                          |          |                                |          
                          |          |                                 --7.18%--rioFileWrite.lto_priv.0
                          |          |                                           |          
                          |          |                                            --6.73%--_IO_fwrite
                          |          |                                                            
                          |          |          
                          |          |--40.44%--rdbSaveStringObject
                          |          |          
                          |           --7.62%--rdbSaveObjectType
                          |                     |          
                          |                      --7.39%--rdbWriteRaw.constprop.1
                          |                                |          
                          |                                 --7.04%--rioFileWrite.lto_priv.0
                          |                                           |          
                          |                                            --6.59%--_IO_fwrite
                          |                                                               
                          |          
                           --7.33%--hashtableNext.constprop.1
                                     |          
                                      --6.28%--prefetchNextBucketEntries.lto_priv.0
```
perf record on save command **with** value prefetching:
```
               rdbSaveRio
               |          
                --99.93%--rdbSaveDb
                          |          
                          |--79.81%--rdbSaveKeyValuePair
                          |          |          
                          |          |--66.79%--rdbSaveRawString
                          |          |          |          
                          |          |          |--42.31%--rdbWriteRaw
                          |          |          |          |          
                          |          |          |           --40.74%--rioFileWrite.lto_priv.0
                          |          |          |          
                          |          |           --23.37%--rdbSaveLen
                          |          |                     |          
                          |          |                     |--11.78%--rdbWriteRaw
                          |          |                     |          |          
                          |          |                     |           --11.03%--rioFileWrite.lto_priv.0
                          |          |                     |                     |          
                          |          |                     |                      --10.30%--_IO_fwrite
                          |          |                     |                                |          
                          |          |                     |          
                          |          |                      --10.98%--rdbWriteRaw.constprop.1
                          |          |                                |          
                          |          |                                 --10.44%--rioFileWrite.lto_priv.0
                          |          |                                           |          
                          |          |                                            --9.74%--_IO_fwrite
                          |          |                                                      |          
                          |          |          
                          |          |--11.33%--rdbSaveObjectType
                          |          |          |          
                          |          |           --10.96%--rdbWriteRaw.constprop.1
                          |          |                     |          
                          |          |                      --10.51%--rioFileWrite.lto_priv.0
                          |          |                                |          
                          |          |                                 --9.75%--_IO_fwrite
                          |          |                                           |          
                          |          |          
                          |           --0.77%--rdbSaveStringObject
                          |          
                           --18.39%--hashtableNext
                                     |          
                                     |--10.04%--hashtableObjectPrefetchValue
                                     |
                                      --6.06%--prefetchNextBucketEntries        

```
Conclusions:

The prefetching strategy appears to be working as intended, shifting the performance bottleneck from data access to I/O operations.
The significant reduction in rdbSaveStringObject time suggests that string objects(which are the values) are being accessed more efficiently.

### DIFF
--- a/src/acl.c
+++ b/src/acl.c
@@ -655,7 +655,7 @@ void ACLChangeSelectorPerm(aclSelector *selector, struct serverCommand *cmd, int
     ACLResetFirstArgsForCommand(selector, id);
     if (cmd->subcommands_ht) {
         hashtableIterator iter;
-        hashtableInitSafeIterator(&iter, cmd->subcommands_ht);
+        hashtableInitIterator(&iter, cmd->subcommands_ht, HASHTABLE_ITER_SAFE);
         void *next;
         while (hashtableNext(&iter, &next)) {
             struct serverCommand *sub = next;
@@ -673,7 +673,7 @@ void ACLChangeSelectorPerm(aclSelector *selector, struct serverCommand *cmd, int
  * found and the operation was performed. */
 void ACLSetSelectorCommandBitsForCategory(hashtable *commands, aclSelector *selector, uint64_t cflag, int value) {
     hashtableIterator iter;
-    hashtableInitIterator(&iter, commands);
+    hashtableInitIterator(&iter, commands, 0);
     void *next;
     while (hashtableNext(&iter, &next)) {
         struct serverCommand *cmd = next;
@@ -741,7 +741,7 @@ void ACLCountCategoryBitsForCommands(hashtable *commands,
                                      unsigned long *off,
                                      uint64_t cflag) {
     hashtableIterator iter;
-    hashtableInitIterator(&iter, commands);
+    hashtableInitIterator(&iter, commands, 0);
     void *next;
     while (hashtableNext(&iter, &next)) {
         struct serverCommand *cmd = next;
@@ -2765,7 +2765,7 @@ sds getAclErrorMessage(int acl_res, user *user, struct serverCommand *cmd, sds e
 /* ACL CAT category */
 void aclCatWithFlags(client *c, hashtable *commands, uint64_t cflag, int *arraylen) {
     hashtableIterator iter;
-    hashtableInitIterator(&iter, commands);
+    hashtableInitIterator(&iter, commands, 0);
     void *next;
     while (hashtableNext(&iter, &next)) {
         struct serverCommand *cmd = next;
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -300,7 +300,7 @@ typedef struct {
     long index;
     uint16_t pos_in_bucket;
     uint8_t table;
-    uint8_t safe;
+    uint8_t flags;
     union {
         /* Unsafe iterator fingerprint for misuse detection. */
         uint64_t fingerprint;
@@ -936,6 +936,7 @@ static inline incrementalFind *incrementalFindFromOpaque(hashtableIncrementalFin
 
 /* Prefetches all filled entries in the given bucket to optimize future memory access. */
 static void prefetchBucketEntries(bucket *b) {
+    if (!b->presence) return;
     for (int pos = 0; pos < numBucketPositions(b); pos++) {
         if (isPositionFilled(b, pos)) {
             valkey_prefetch(b->entries[pos]);
@@ -977,6 +978,26 @@ static void prefetchNextBucketEntries(iter *iter, bucket *current_bucket) {
             valkey_prefetch(next_next_bucket);
         }
     }
+}
+
+/* Prefetches the values associated with the entries in the given bucket by
+ * calling the entryPrefetchValue callback in the hashtableType */
+static void prefetchBucketValues(bucket *b, hashtable *ht) {
+    if (!b->presence) return;
+    assert(ht->type->entryPrefetchValue != NULL);
+    for (int pos = 0; pos < numBucketPositions(b); pos++) {
+        if (isPositionFilled(b, pos)) {
+            ht->type->entryPrefetchValue(b->entries[pos]);
+        }
+    }
+}
+
+static inline int isSafe(iter *iter) {
+    return (iter->flags & HASHTABLE_ITER_SAFE);
+}
+
+static inline int shouldPrefetchValues(iter *iter) {
+    return (iter->flags & HASHTABLE_ITER_PREFETCH_VALUES);
 }
 
 /* --- API functions --- */
@@ -1792,31 +1813,32 @@ size_t hashtableScanDefrag(hashtable *ht, size_t cursor, hashtableScanFunction f
 
 /* --- Iterator --- */
 
-/* Initialize a iterator, that is not allowed to insert, delete or even lookup
- * entries in the hashtable, because such operations can trigger incremental
- * rehashing which moves entries around and confuses the iterator. Only
- * hashtableNext is allowed. Each entry is returned exactly once. Call
- * hashtableResetIterator when you are done. See also
- * hashtableInitSafeIterator. */
-void hashtableInitIterator(hashtableIterator *iterator, hashtable *ht) {
-    iter *iter;
-    iter = iteratorFromOpaque(iterator);
-    iter->hashtable = ht;
-    iter->table = 0;
-    iter->index = -1;
-    iter->safe = 0;
-}
-
-/* Initialize a safe iterator, which is allowed to modify the hash table while
- * iterating. It pauses incremental rehashing to prevent entries from moving
- * around. Call hashtableNext to fetch each entry. You must call
- * hashtableResetIterator when you are done with a safe iterator.
+/* Initialize an iterator for a hashtable.
  *
- * It's allowed to insert and replace entries. Deleting entries is only allowed
- * for the entry that was just returned by hashtableNext. Deleting other entries
- * is possible, but doing so can cause internal fragmentation, so don't.
+ * The 'flags' argument can be used to tweak the behaviour. It's a bitwise-or
+ * (zero means no flags) of the following:
  *
- * Guarantees:
+ * - HASHTABLE_ITER_SAFE: Use a safe iterator that can handle
+ *   modifications to the hash table while iterating.
+ * - HASHTABLE_ITER_PREFETCH_VALUES: Enables prefetching of entries values,
+ *   which can improve performance in some scenarios. Because the hashtable is generic and
+ *   doesn't care which object we store, the callback entryPrefetchValue must be set to help
+ *   us prefetch necessary fields of specific object types stored in the hashtable.
+ *
+ * For a non-safe iterator (default, when HASHTABLE_ITER_SAFE is not set):
+ * It is not allowed to insert, delete or even lookup entries in the hashtable,
+ * because such operations can trigger incremental rehashing which moves entries
+ * around and confuses the iterator. Only hashtableNext is allowed. Each entry
+ * is returned exactly once.
+ *
+ * For a safe iterator (when HASHTABLE_ITER_SAFE is set):
+ * It is allowed to modify the hash table while iterating. It pauses incremental
+ * rehashing to prevent entries from moving around. It's allowed to insert and
+ * replace entries. Deleting entries is only allowed for the entry that was just
+ * returned by hashtableNext. Deleting other entries is possible, but doing so
+ * can cause internal fragmentation, so don't.
+ *
+ * Guarantees for safe iterators:
  *
  * - Entries that are in the hash table for the entire iteration are returned
  *   exactly once.
@@ -1829,18 +1851,31 @@ void hashtableInitIterator(hashtableIterator *iterator, hashtable *ht) {
  *
  * - Entries that are inserted during the iteration may or may not be returned
  *   by the iterator.
+ *
+ * Call hashtableNext to fetch each entry. You must call hashtableResetIterator
+ * when you are done with the iterator.
  */
-void hashtableInitSafeIterator(hashtableIterator *iterator, hashtable *ht) {
-    hashtableInitIterator(iterator, ht);
+void hashtableInitIterator(hashtableIterator *iterator, hashtable *ht, uint8_t flags) {
+    iter *iter;
+    iter = iteratorFromOpaque(iterator);
+    iter->hashtable = ht;
+    iter->table = 0;
+    iter->index = -1;
+    iter->flags = flags;
+}
+
+/* Reinitializes the iterator for the provided hashtable while
+ * preserving the flags from its previous initialization. */
+void hashtableReinitIterator(hashtableIterator *iterator, hashtable *ht) {
     iter *iter = iteratorFromOpaque(iterator);
-    iter->safe = 1;
+    hashtableInitIterator(iterator, ht, iter->flags);
 }
 
 /* Resets a stack-allocated iterator. */
 void hashtableResetIterator(hashtableIterator *iterator) {
     iter *iter = iteratorFromOpaque(iterator);
     if (!(iter->index == -1 && iter->table == 0)) {
-        if (iter->safe) {
+        if (isSafe(iter)) {
             hashtableResumeRehashing(iter->hashtable);
             assert(iter->hashtable->pause_rehash >= 0);
         } else {
@@ -1850,19 +1885,11 @@ void hashtableResetIterator(hashtableIterator *iterator) {
 }
 
 /* Allocates and initializes an iterator. */
-hashtableIterator *hashtableCreateIterator(hashtable *ht) {
+hashtableIterator *hashtableCreateIterator(hashtable *ht, uint8_t flags) {
     iter *iter = zmalloc(sizeof(*iter));
     hashtableIterator *opaque = iteratorToOpaque(iter);
-    hashtableInitIterator(opaque, ht);
+    hashtableInitIterator(opaque, ht, flags);
     return opaque;
-}
-
-/* Allocates and initializes a safe iterator. */
-hashtableIterator *hashtableCreateSafeIterator(hashtable *ht) {
-    hashtableIterator *iterator = hashtableCreateIterator(ht);
-    iter *iter = iteratorFromOpaque(iterator);
-    iter->safe = 1;
-    return iterator;
 }
 
 /* Resets and frees the memory of an allocated iterator, i.e. one created using
@@ -1880,7 +1907,7 @@ int hashtableNext(hashtableIterator *iterator, void **elemptr) {
     while (1) {
         if (iter->index == -1 && iter->table == 0) {
             /* It's the first call to next. */
-            if (iter->safe) {
+            if (isSafe(iter)) {
                 hashtablePauseRehashing(iter->hashtable);
                 iter->last_seen_size = iter->hashtable->used[iter->table];
             } else {
@@ -1907,7 +1934,7 @@ int hashtableNext(hashtableIterator *iterator, void **elemptr) {
                 iter->bucket = getChildBucket(iter->bucket);
             } else if (iter->pos_in_bucket >= ENTRIES_PER_BUCKET) {
                 /* Bucket index done. */
-                if (iter->safe) {
+                if (isSafe(iter)) {
                     /* If entries in this bucket chain have been deleted,
                      * they've left empty spaces in the buckets. The chain is
                      * not automatically compacted when rehashing is paused. If
@@ -1936,6 +1963,9 @@ int hashtableNext(hashtableIterator *iterator, void **elemptr) {
         }
         bucket *b = iter->bucket;
         if (iter->pos_in_bucket == 0) {
+            if (shouldPrefetchValues(iter)) {
+                prefetchBucketValues(b, iter->hashtable);
+            }
             prefetchNextBucketEntries(iter, b);
         }
         if (!isPositionFilled(b, iter->pos_in_bucket)) {
--- a/src/hashtable.h
+++ b/src/hashtable.h
@@ -60,6 +60,8 @@ typedef struct {
     /* Callback to free an entry when it's overwritten or deleted.
      * Optional. */
     void (*entryDestructor)(void *entry);
+    /* Callback to prefetch the value associated with a hashtable entry. */
+    void (*entryPrefetchValue)(const void *entry);
     /* Callback to control when resizing should be allowed. */
     int (*resizeAllowed)(size_t moreMem, double usedRatio);
     /* Invoked at the start of rehashing. */
@@ -90,6 +92,10 @@ typedef void (*hashtableScanFunction)(void *privdata, void *entry);
 
 /* Scan flags */
 #define HASHTABLE_SCAN_EMIT_REF (1 << 0)
+
+/* Iterator flags */
+#define HASHTABLE_ITER_SAFE (1 << 0)
+#define HASHTABLE_ITER_PREFETCH_VALUES (1 << 1)
 
 /* --- Prototypes --- */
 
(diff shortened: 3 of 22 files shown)


src/acl.c
```diff
@@ -657,7 +657,7 @@ void ACLChangeSelectorPerm(aclSelector *selector, struct serverCommand *cmd, int
         hashtableIterator iter;
         hashtableInitSafeIterator(&iter, cmd->subcommands_ht);
         void *next;
-        while (hashtableNext(&iter, &next)) {
+        while (hashtableNext(&iter, &next, 0)) {
```
madolson: Why aren't we prefetching in these cases?

NadavGigi: In this pr i have added support only for hashtables that storing robj and hashTypeEntry.
I will add support for more complex data structures in a different pr.

zuiderkwast: This makes the `hashtableNext` call less simple and the change affects all places even where it is not used. I have another suggestions: storing the iterator flags in the iterator:

```C
        hashtableIterator iter;
        hashtableInitSafeIterator(&iter, cmd->subcommands_ht);
        hashtableIteratorSetFlags(HASHTABLE_ITER_PREFETCH_VALUES);
        void *next;
        while (hashtableNext(&iter, &next)) {
```

Alternatively, we could set the prefetch callback directly in the iterator instead of storing it in the hashtableType.

```C
        hashtableIteratorSetPrefetchCallback(&iter, hashTypeEntryPrefetchValue);
```

zuiderkwast (commented):
Interesting idea. I just skimmed though, not a full review.

To summarize, this is a second level of pre-fetching. Theoretically, we could do even more levels, right? It could even be generalized to any number of levels. We could use an array of prefetch callbacks.... (not saying we should)

We should probably also consider the bigger picture where this kind of callbacks can be used not only for the iterator, but also for the hashtableIncrementalFind that's already used with IO threads, and with other operations like Scan.

A hypothetical use case: Imagine we have a sorted set where we want to find the 10 members sorted by score, ZRANGE. We could find the starting element and also prefetch in up to 10 steps the next elements.

Have you been thinking about more prefetching ideas? Can you share your ideas? It's useful to understand the bigger picture better before we add many small optimizations for specific scenarios.

NadavGigi:
> Interesting idea. I just skimmed though, not a full review.
> 
> To summarize, this is a second level of pre-fetching. Theoretically, we could do even more levels, right? It could even be generalized to any number of levels. We could use an array of prefetch callbacks.... (not saying we should)
> 
> We should probably also consider the bigger picture where this kind of callbacks can be used not only for the iterator, but also for the hashtableIncrementalFind that's already used with IO threads, and with other operations like Scan.
> 
> A hypothetical use case: Imagine we have a sorted set where we want to find the 10 members sorted by score, ZRANGE. We could find the starting element and also prefetch in up to 10 steps the next elements.
> 
> Have you been thinking about more prefetching ideas? Can you share your ideas? It's useful to understand the bigger picture better before we add many small optimizations for specific scenarios.


@zuiderkwast Thank you for your thoughtful feedback and for taking the time to review my PR. I apologize for not sharing my broader ideas earlier.

Let me address these and share my current thoughts on the matter:

1. Multi-level Prefetching:
   Your observation about the potential for multiple levels of prefetching is intriguing. While my current implementation focuses on a second level, I'm not entirely sure how well extending this to multiple levels would work in practice. 

   I think that your levels idea will happen anyway with this PR when we have nested hashtables. With other data structures, each one should be considered individually if it can benefit from prefetching or not. The perefetching of deeper levels should be the caller responsible and by the iterator.

   Anyway, notice that with each level of prefetching, we will have to prefetch more buckets and entries ahead of the current bucket from which we're returning entries.
   

2. Broader Applications:
   I agree that we should consider wider applications of this prefetching mechanism including the ZRANGE usecase you suggested. I have already started thinking about it and will share the insights soon.

Current Ideas and Roadmap:
I'm approaching these optimizations by porting techniques that have proven effective in dict implementations, while being mindful that direct application may not always yield the same benefits in hashtable contexts. Here's my current thinking where each of these optimizations builds upon the previous ones :

1. Entries prefetching - Done
2. Value prefetching for db iterations. Prefetching the internal fields of robj and hashTypeEntry - in PR.
3. Expanding the value prefetching to other data stcructures. - Implementation
4. second-level prefetching mechanism. This mechanism operates at the data structure level and is triggered when receiving client commands, before their actual execution. It leverages the existing incremental find API through a callback system. For instance, when processing a command like 'hget myhash f1 v1', a specific callback is invoked. This callback utilizes the incremental find API to initiate prefetching of relevant data structures, potentially improving performance by reducing latency during command execution.
5. Using prefetch (similar to what you suggested in [hashtableIncrementalFindInit](https://github.com/valkey-io/valkey/blob/2d0b8e360847146fa43a8463773dde8125187fa7/src/hashtable.c#L1511)) to speed up sinterGenericCommand and sunionDiffGenericCommand. - Implementation.

src/acl.c
NadavGigi: I think that the first option you suggested is better than the second.
What do you think about this option(which i have already implemented):
```
typedef struct {
    hashtable *hashtable;
    bucket *bucket;
    long index;
    uint16_t pos_in_bucket;
    uint8_t table;
   **int flags;**
    union {
        /* Unsafe iterator fingerprint for misuse detection. */
        uint64_t fingerprint;
        /* Safe iterator temporary storage for bucket chain compaction. */
        uint64_t last_seen_size;
    };
} iter;
```
I have combined the safe flag and the prefetch flag into one.
Then when we want to init safe and value prefetching iterator we will do:
```
hashtableInitIterator(&iter, ht, HASHTABLE_ITER_SAFE | HASHTABLE_ITER_PREFETCH_VALUES);
```
I agree that this code require more changes than your approach but the duplicated code of the safe and unsafe iterator in init and create is removed.
Let me know what you think i will change it accordingly.

zuiderkwast: > hashtableInitIterator(&iter, ht, HASHTABLE_ITER_SAFE | HASHTABLE_ITER_PREFETCH_VALUES);

Great idea! I like it!

>  `int flags;`

Regarding the flags in the iterator struct, let's use `uint8_t`? It's enough and it will make the fields `pos_in_table, table, flags` a nice group of 32 bits.

zuiderkwast:
> I apologize for not sharing my broader ideas earlier.

@NadavGigi No need to apologize! I didn't ask before. :D

> 1\. Multi-level Prefetching:
>        Your observation about the potential for multiple levels of prefetching is intriguing. While my current implementation focuses on a second level, I'm not entirely sure how well extending this to multiple levels would work in practice.

Yes, intriguing, that was the intention of my comment. :smile: We should base the decisions on actual use cases and facts.

The ZRANGE idea I had isn't even in the hashtable code. It is the skiplist code. We could definitely do some prefetching in that kind of code. And in the quicklist implementation for lists we could do it.

> 4\. second-level prefetching mechanism. This mechanism operates at the data structure level and is triggered when receiving client commands, before their actual execution. It leverages the existing incremental find API through a callback system. For instance, when processing a command like 'hget myhash f1 v1', a specific callback is invoked. This callback utilizes the incremental find API to initiate prefetching of relevant data structures, potentially improving performance by reducing latency during command execution.

This is interesting! Should we add this callback information in the commands' JSON files? Feel free to open an issue to explain it, or a PR if you already have some code.

> 5\. Using prefetch (similar to what you suggested in [hashtableIncrementalFindInit](https://github.com/valkey-io/valkey/blob/2d0b8e360847146fa43a8463773dde8125187fa7/src/hashtable.c#L1511)) to speed up sinterGenericCommand and sunionDiffGenericCommand. - Implementation.

Cool!

It seems to me that we should prioritize SCAN next, don't you think? If users find out that KEYS is faster than SCAN, maybe they switch to using KEYS. :scream_cat: And we use scan for expiration and other internal jobs too.

It seems easy to add prefetching to scan. We can just prefetch the next bucket. Maybe it doesn't speed up `SCAN 0 COUNT 10` much, but it might speed up `SCAN 0 COUNT 100`...

src/acl.c
NadavGigi: I have changed the implementation.
flags is now uint8_t

src/server.c
```diff
@@ -577,6 +577,14 @@ const void *hashtableObjectGetKey(const void *entry) {
     return objectGetKey(entry);
 }
 
+/* Prefetch the value if it's not embedded. */
+void hashtableObjectPrefetchValue(const void *entry) {
+    const robj *obj = entry;
+    if (obj->encoding != OBJ_ENCODING_EMBSTR) {
+        valkey_prefetch(obj->ptr);
```
zuiderkwast: OBJ_ENCODING_INT should also not be prefetched. In this case, `obj->ptr` is an integer cast to a pointer, so not a real memory address.

We can also skip ptr == NULL. Not sure if it's ever NULL, but it is not impossible.

src/hashtable.c
```diff
@@ -979,6 +986,24 @@ static void prefetchNextBucketEntries(iter *iter, bucket *current_bucket) {
     }
 }
 
+/* Prefetches the values associated with filled entries in the given bucket. */
```
zuiderkwast: "Filled" is about positions that may contain an entry or be empty. A position is filled. An entry is always in a filled position.

We can refer to the callback in this comment too.

```suggestion
/* Prefetches the values associated with the entries in the given bucket by
 * calling the entryPrefetchValue callback in the hashtableType, if defined. */
```

src/hashtable.c
```diff
@@ -979,6 +986,24 @@ static void prefetchNextBucketEntries(iter *iter, bucket *current_bucket) {
...
+static void prefetchBucketValues(bucket *b, hashtable *ht) {
+    if (!b->presence) return;
+    for (int pos = 0; pos < numBucketPositions(b); pos++) {
+        if (isPositionFilled(b, pos)) {
+            entryPrefetchValue(ht, b->entries[pos]);
+        }
+    }
```
zuiderkwast: I guess it is just as simple, and easier to see that it calls the type callback, without the extra entryPrefetchValue helper function. We can return early if there is no entryPrefetchValue callback. Then we can delete the entryPrefetchValue helper. WDYT?

```suggestion
    if (!b->presence || ht->type->entryPrefetchValue == NULL) return;
    for (int pos = 0; pos < numBucketPositions(b); pos++) {
        if (isPositionFilled(b, pos)) {
            ht->type->entryPrefetchValue(b->entries[pos]);
        }
    }
```

src/hashtable.c
```diff
@@ -1829,18 +1855,31 @@ void hashtableInitIterator(hashtableIterator *iterator, hashtable *ht) {
  *
  * - Entries that are inserted during the iteration may or may not be returned
  *   by the iterator.
+ *
+ * Call hashtableNext to fetch each entry. You must call hashtableResetIterator
+ * when you are done with the iterator.
```
zuiderkwast, line 1856: Reset is only strictly required for safe iterators, to resume rehashing which is paused when we init a safe iterator. For non-safe iterators it's not strictly required, but it does no harm. It's OK to keep this. I guess it's good practice to always reset it.

zuiderkwast (commented):
I like it! Just some nits.

src/server.c
xbasel: prefetch does not generate faults if the address is not valid:
https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html

zuiderkwast: That's true but it can try to prefetch some random address that we don't need and evict other stuff from cache.

xbasel: The likelihood of a random integer being a valid address for __builtin_prefetch is extremely low, especially on 64-bit systems, because the valid address space used by a process is only a tiny fraction of the massive 64-bit virtual address space. I’m not opposed to ensuring it’s not an integer or NULL, but this is more about code/logic correctness than addressing a probable issue.

zuiderkwast: OK, that's probably true, but why do you even want to exclude OBJ_ENCODING_EMBSTR then? It looks like a mistake if you exclude it but don't exclude OBJ_ENCODING_INT.

```suggestion
    if (obj->encoding != OBJ_ENCODING_EMBSTR && obj->encoding != OBJ_ENCODING_INT) {
        valkey_prefetch(obj->ptr);
```

NadavGigi: you are correct.
We should avoid prefetching when obj->encoding == OBJ_ENCODING_INT and i will change it.

xbasel: I was merely saying that even if we prefetch an invalid address, the impact would be minimal.
I think I am in favor of filtering out non-prefetchable pointers.

xbasel: I also opened this issue for second level prefetching, which I already started coding
https://github.com/valkey-io/valkey/issues/1594

NadavGigi: I have made some performance testing with and without the null check and there was no difference between the two.

xbasel: This is kind of expected, its comparing a value in L1 cache to NULL, perhaps no more than 4-5 CPU cycles.

src/hashtable.c
NadavGigi: I agree its better to "overkill" and keep it for both iterators

src/hashtable.c
NadavGigi: I agree.

src/hashtable.h
```diff
@@ -48,6 +48,8 @@ typedef uint64_t hashtableIncrementalFindState[5];
  * optional. With all callbacks omitted, the hashtable is effectively a set of
  * pointer-sized integers. */
 typedef struct {
+    /* Callback to prefetch the value associated with a hashtable entry. */
+    void (*entryPrefetchValue)(const void *entry);
```
zuiderkwast: Nit: This shouldn't be the first callback in the list. It's not the most important or most basic one. Let's move it after the basic ones, for example after entryDestructor.

zuiderkwast (approved):
LGTM.

@madolson @xbasel do you still want to review this, or OK to merge?

NadavGigi:
> > I apologize for not sharing my broader ideas earlier.
> 
> @NadavGigi No need to apologize! I didn't ask before. :D
>
Haha, I guess I was being a bit too formal there!
> > 1. Multi-level Prefetching:
> > Your observation about the potential for multiple levels of prefetching is intriguing. While my current implementation focuses on a second level, I'm not entirely sure how well extending this to multiple levels would work in practice.
> 
> Yes, intriguing, that was the intention of my comment. 😄 We should base the decisions on actual use cases and facts.
> 
> The ZRANGE idea I had isn't even in the hashtable code. It is the skiplist code. We could definitely do some prefetching in that kind of code. And in the quicklist implementation for lists we could do it.
>
Great Idea! it can be added to the prefetching tasks list. When working on sunionDiffGenericCommand i will try to make a general batch prefetching function that will might be good for all.
> > 4. second-level prefetching mechanism. This mechanism operates at the data structure level and is triggered when receiving client commands, before their actual execution. It leverages the existing incremental find API through a callback system. For instance, when processing a command like 'hget myhash f1 v1', a specific callback is invoked. This callback utilizes the incremental find API to initiate prefetching of relevant data structures, potentially improving performance by reducing latency during command execution.
> 
> This is interesting! Should we add this callback information in the commands' JSON files? Feel free to open an issue to explain it, or a PR if you already have some code.
> 
> > 5. Using prefetch (similar to what you suggested in [hashtableIncrementalFindInit](https://github.com/valkey-io/valkey/blob/2d0b8e360847146fa43a8463773dde8125187fa7/src/hashtable.c#L1511)) to speed up sinterGenericCommand and sunionDiffGenericCommand. - Implementation.
> 
> Cool!
> 
> It seems to me that we should prioritize SCAN next, don't you think? If users find out that KEYS is faster than SCAN, maybe they switch to using KEYS. 🙀 And we use scan for expiration and other internal jobs too.
> 
> It seems easy to add prefetching to scan. We can just prefetch the next bucket. Maybe it doesn't speed up `SCAN 0 COUNT 10` much, but it might speed up `SCAN 0 COUNT 100`...

Sounds great!
Myaybe we can come up with something like:
Start by prefetching next bucket and then prefetching the next bucket entries when we reach the middle of the current bucket we iterating.
I dont know yet how it will be prioritized but i will let you know as soon as possible.

src/hashtable.h
NadavGigi: Done.

xbasel:
> xbasel

I'd like to have another look. Should be quick.

src/hashtable.c
```diff
@@ -979,6 +980,25 @@ static void prefetchNextBucketEntries(iter *iter, bucket *current_bucket) {
     }
 }
 
+/* Prefetches the values associated with the entries in the given bucket by
+ * calling the entryPrefetchValue callback in the hashtableType, if defined. */
+static void prefetchBucketValues(bucket *b, hashtable *ht) {
+    if (!b->presence || ht->type->entryPrefetchValue != NULL) return;
```
madolson: Having `ht->type->entryPrefetchValue != NULL` and setting `HASHTABLE_ITER_PREFETCH_VALUES` is a misconfiguration that isn't clearly flagged. Would be nice to have a runtime assert to make sure we are catching that.

src/server.c
```diff
@@ -577,6 +577,16 @@ const void *hashtableObjectGetKey(const void *entry) {
     return objectGetKey(entry);
 }
 
+/* Prefetch the value if it's not embedded. */
+void hashtableObjectPrefetchValue(const void *entry) {
+    const robj *obj = entry;
+    if (obj->encoding != OBJ_ENCODING_EMBSTR &&
+        obj->encoding != OBJ_ENCODING_INT &&
+        obj->ptr != NULL) {
```
madolson: When is obj->ptr NULL? It shouldn't ever be null, would rather this be a server assert since that means something is corrupted on the server.

src/hashtable.c
```diff
@@ -1936,6 +1962,9 @@ int hashtableNext(hashtableIterator *iterator, void **elemptr) {
         }
         bucket *b = iter->bucket;
         if (iter->pos_in_bucket == 0) {
+            if (isPrefetchingValues(iter)) {
```
madolson: nit: The wording here is a little confusing, implies it's already prefetching as opposed to the action it should take.
```suggestion
            if (shouldPrefetchValues(iter)) {
```

Alternatively, `isPrefetchValuesIterator` would also make more sense.

src/hashtable.c
```diff
@@ -1792,31 +1812,32 @@ size_t hashtableScanDefrag(hashtable *ht, size_t cursor, hashtableScanFunction f
...
+/* Initialize an iterator for a hashtable.
  *
- * It's allowed to insert and replace entries. Deleting entries is only allowed
- * for the entry that was just returned by hashtableNext. Deleting other entries
- * is possible, but doing so can cause internal fragmentation, so don't.
+ * The 'flags' argument can be used to tweak the behaviour. It's a bitwise-or
+ * (zero means no flags) of the following:
+ *
+ * - HASHTABLE_ITER_SAFE: Use a safe iterator that can handle concurrent
+ *   modifications to the hash table.
```
madolson:
```suggestion
 * - HASHTABLE_ITER_SAFE: Use a safe iterator that can handle
 *   modifications to the hash table while iterating. 
```

madolson (commented):
Just minor comments about wording.

src/server.c
NadavGigi: I have removed the NULL check, but I'm not sure this is where we want to make this assertion. Prefetch does not generate faults if the address is invalid, as noted in the GCC documentation: https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html

This part of the PR was discussed here: https://github.com/valkey-io/valkey/pull/1568#discussion_r1922847238
Perhaps we should open a separate issue to discuss where and how to best catch this unexpected behavior.

src/hashtable.c
NadavGigi: Agree.

src/server.c
zuiderkwast: > I'm not sure this is where we want to make this assertion

Right, no assertion here.

The assert I think Madelyn wants is another one, to check that the callback is set:

    serverAssert(ht->type->entryPrefetchValue != NULL);

instead of checking it in the `if` above:

    if (!b->presence || ht->type->entryPrefetchValue != NULL) return;

xbasel (approved):

src/hashtable.c
NadavGigi: Done.

src/server.c
NadavGigi: An assertion has been added.

src/hashtable.c
NadavGigi: Done.

src/server.c
zuiderkwast: LGTM.

In your next PR, it'd be good if you can avoid force-push and instead just add many small commits. It makes it easier to do incremental review and see what's changed since the last review. We squash-merge the PR to one commit anyway when we merge it.

madolson:
@NadavGigi Another remind to not force push, just keep pushing little commits, we can squash them at the end.

madolson (approved):

PR #1570, opened 2025-01-16
Extend LATENCY LATEST to add sum / cnt stats

enjoy-binbin:
Currently LATENCY LATEST only has the following fields:
- Event name.
- Unix timestamp of the latest latency spike for the event.
- Latest event latency in millisecond.
- All-time maximum latency for this event.

This PR introduced these fields:
- sum: the all-time sum latency for this event (i think it is useful for some events)
- cnt: the event count that trigger the latency, with the sum we can calc the avg

### DIFF
--- a/src/latency.c
+++ b/src/latency.c
@@ -86,7 +86,6 @@ void latencyAddSample(const char *event, mstime_t latency) {
         ts = zmalloc(sizeof(*ts));
         ts->idx = 0;
         ts->max = 0;
-        ts->min = 0;
         ts->sum = 0;
         ts->cnt = 0;
         memset(ts->samples, 0, sizeof(ts->samples));
@@ -94,7 +93,6 @@ void latencyAddSample(const char *event, mstime_t latency) {
     }
 
     if (latency > ts->max) ts->max = latency;
-    if (latency < ts->min || ts->min == 0) ts->min = latency;
     ts->sum += latency;
     ts->cnt++;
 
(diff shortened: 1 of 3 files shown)


zuiderkwast (approved):
OK, I think we can add this.

We are adding these in the end of an array reply. I remember we did something like that in CLUSTER SLOTS and we said it is a backward-compatible change, but still it broke some clients. For example, hiredis-cluster was checking the exact array length in the handling of CLUSTER SLOTS.

LATENCY LATEST is not as risky to change as CLUSTER SLOTS, but we should still be aware that some monitoring tools maybe can break. We should mention it in the release notes and we should add some note in the documentation that new fields can be added. A note like that makes it more clear that adding more fields in the reply is a backward compatible change. Now we don't have a note like that on https://valkey.io/commands/latency-latest/.

soloestoy (approved):

zuiderkwast:
@valkey-io/core-team Please vote to extend the multi-bulk reply with three more numbers.

src/latency.c
```diff
@@ -612,11 +618,14 @@ void latencyCommandReplyWithLatestEvents(client *c) {
         struct latencyTimeSeries *ts = dictGetVal(de);
         int last = (ts->idx + LATENCY_TS_LEN - 1) % LATENCY_TS_LEN;
 
-        addReplyArrayLen(c, 4);
+        addReplyArrayLen(c, 7);
         addReplyBulkCString(c, event);
         addReplyLongLong(c, ts->samples[last].time);
         addReplyLongLong(c, ts->samples[last].latency);
         addReplyLongLong(c, ts->max);
+        addReplyLongLong(c, ts->min);
```
madolson: I vote to remove min. It's basically the lowest value above min, I think sum and cnt are more useful.

zuiderkwast: I guess you mean min is the lowest value above the configured `latency-monitor-threshold`.

enjoy-binbin: Will it be useful in scenarios where `latency-monitor-threshold` may be frequently adjusted? or the real latency is much larger than `latency-monitor-threshold`?

i am not sure about it actually, i am ok with removing it if you both think it is useless.

zuiderkwast: Only Madelyn thinks it's useless. I don't have an opinion about it. :laughing:

madolson: I'm pretty strongly opinionated about only adding fields we are sure are useful. If we aren't sure something is useful, I would prefer to defer adding it until someone brings it up or has a use case. I prefer to lean towards simplicity as much as possible. The fewer the fields exist, the less the end users have to think about.

> Will it be useful in scenarios where latency-monitor-threshold may be frequently adjusted? or the real latency is much larger than latency-monitor-threshold?

I'm not sure what the min would tell you in that scenario.

madolson (approved):

hwware (approved):

PR #1572, opened 2025-01-16
Enable module API SendClusterMessage to use light message header type

hpatro:
This change uses the light message header for cluster module message type to be sent to a target node or broadcast across the cluster. The light message header was introduced in Valkey 8 to reduce network in/out over clusterbus and have been already used for pub/sub message transfer. It's only used if the target node supports the light message header (~30B) otherwise it falls back to using the regular header (~2KB). The module API `VM_SendClusterMessage` remains as is.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -195,11 +195,13 @@ dictType clusterSdsToListType = {
 typedef struct {
     enum {
         ITER_DICT,
-        ITER_LIST
+        ITER_LIST,
+        ITER_NODE,
     } type;
     union {
         dictIterator di;
         listIter li;
+        clusterNode *node;
     };
 } ClusterNodeIterator;
 
@@ -213,6 +215,11 @@ static void clusterNodeIterInitMyShard(ClusterNodeIterator *iter) {
     serverAssert(nodes != NULL);
     iter->type = ITER_LIST;
     listRewind(nodes, &iter->li);
+}
+
+static void clusterNodeIterNode(ClusterNodeIterator *iter, clusterNode *node) {
+    iter->type = ITER_NODE;
+    iter->node = node;
 }
 
 static clusterNode *clusterNodeIterNext(ClusterNodeIterator *iter) {
@@ -229,6 +236,15 @@ static clusterNode *clusterNodeIterNext(ClusterNodeIterator *iter) {
         /* Return the value associated with the node, or NULL if no more nodes */
         return ln ? listNodeValue(ln) : NULL;
     }
+
+    case ITER_NODE: {
+        if (iter->node) {
+            clusterNode *node = iter->node;
+            iter->node = NULL;
+            return node;
+        }
+        return NULL;
+    }
     }
     serverPanic("Unknown iterator type %d", iter->type);
 }
@@ -236,6 +252,8 @@ static clusterNode *clusterNodeIterNext(ClusterNodeIterator *iter) {
 static void clusterNodeIterReset(ClusterNodeIterator *iter) {
     if (iter->type == ITER_DICT) {
         dictResetIterator(&iter->di);
+    } else if (iter->type == ITER_NODE) {
+        iter->node = NULL;
     }
 }
 
@@ -3019,11 +3037,19 @@ static void clusterProcessPublishPacket(clusterMsgDataPublish *publish_data, uin
     }
 }
 
-static void clusterProcessLightPacket(clusterLink *link, uint16_t type) {
+static void clusterProcessLightPacket(clusterNode *sender, clusterLink *link, uint16_t type) {
     clusterMsgLight *hdr = (clusterMsgLight *)link->rcvbuf;
-
+    serverLog(LL_DEBUG, "Processing light packet of type: %s", clusterGetMessageTypeString(type));
     if (type == CLUSTERMSG_TYPE_PUBLISH || type == CLUSTERMSG_TYPE_PUBLISHSHARD) {
         clusterProcessPublishPacket(&hdr->data.publish.msg, type);
+    } else if (type == CLUSTERMSG_TYPE_MODULE) {
+        uint64_t module_id = hdr->data.module.msg.module_id; /* Endian-safe ID */
+        uint32_t len = ntohl(hdr->data.module.msg.len);
+        uint8_t type = hdr->data.module.msg.type;
+        unsigned char *payload = hdr->data.module.msg.bulk_data;
+        moduleCallClusterReceivers(sender->name, module_id, type, payload, len);
+    } else {
+        serverAssert(0);
     }
 }
 
@@ -3031,6 +3057,7 @@ static inline int messageTypeSupportsLightHdr(uint16_t type) {
     switch (type) {
     case CLUSTERMSG_TYPE_PUBLISH: return 1;
     case CLUSTERMSG_TYPE_PUBLISHSHARD: return 1;
+    case CLUSTERMSG_TYPE_MODULE: return 1;
     }
     return 0;
 }
@@ -3123,8 +3150,14 @@ int clusterIsValidPacket(clusterLink *link) {
         explen = sizeof(clusterMsg) - sizeof(union clusterMsgData);
         explen += sizeof(clusterMsgDataUpdate);
     } else if (type == CLUSTERMSG_TYPE_MODULE) {
-        explen = sizeof(clusterMsg) - sizeof(union clusterMsgData);
-        explen += sizeof(clusterMsgModule) - 3 + ntohl(hdr->data.module.msg.len);
+        if (is_light) {
+            clusterMsgLight *hdr_light = (clusterMsgLight *)link->rcvbuf;
+            explen = sizeof(clusterMsgLight) - sizeof(union clusterMsgData);
+            explen += sizeof(clusterMsgModule) - 3 + ntohl(hdr_light->data.module.msg.len);
+        } else {
+            explen = sizeof(clusterMsg) - sizeof(union clusterMsgData);
+            explen += sizeof(clusterMsgModule) - 3 + ntohl(hdr->data.module.msg.len);
+        }
     } else {
         /* We don't know this type of packet, so we assume it's well formed. */
         explen = totlen;
@@ -3178,7 +3211,7 @@ int clusterProcessPacket(clusterLink *link) {
         }
         clusterNode *sender = link->node;
         sender->data_received = now;
-        clusterProcessLightPacket(link, type);
+        clusterProcessLightPacket(sender, link, type);
         return 1;
     }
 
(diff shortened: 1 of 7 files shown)


src/cluster_legacy.c
```diff
@@ -195,11 +195,13 @@ dictType clusterSdsToListType = {
 typedef struct {
     enum {
         ITER_DICT,
-        ITER_LIST
+        ITER_LIST,
+        ITER_NODE,
```
hpatro, line 199: This helps simplify the code a bit however it's just iterating over one node.

src/cluster_legacy.c
```diff
@@ -4316,22 +4349,50 @@ void clusterSendUpdate(clusterLink *link, clusterNode *node) {
...
+    if (link) {
+        clusterNodeIterNode(&iter, link->node);
+    } else {
+        /* Broadcast to all the nodes. */
+        clusterNodeIterInitAllNodes(&iter);
+    }
+    clusterNode *node;
+    while ((node = clusterNodeIterNext(&iter)) != NULL) {
+        if (node->flags & (CLUSTER_NODE_MYSELF | CLUSTER_NODE_HANDSHAKE)) continue;
+        if (nodeSupportsLightMsgHdr(node)) {
```
zuiderkwast: We can't use this judgement. It returns true for an 8.0 node, but an 8.0 node can't handle a light-weight module message. It only handles light publish. Right?

I think we need to rename that function to `nodeSupportsLightPublish` and introduce a new bit in the cluster for *node supports light module message*.

zuiderkwast: This is a typical example where we need the framework in #1371.

hpatro: Right, I think I got carried away with the method name :D

zuiderkwast (commented):
It looks pretty good, but we need cross-version testing IMO with a mixed cluster to make sure it works during rolling upgrade, i.e. a mixed cluster. Can you take my work from #1371 and add it here? Or help completing it.

sarthakaggarwal97:
> Can you take my work from https://github.com/valkey-io/valkey/pull/1371 and add it here? Or help completing it.

@zuiderkwast just out of curiosity what is pending over here? I see @madolson approved the PR as well. I can help contribute / complete it.

tests/unit/moduleapi/cross-version-cluster.tcl
```diff
@@ -0,0 +1,61 @@
...
+
+tags {external:skip cluster modules} {
+
+# To run this test use the `--other-server-path` parameter and pass in a compatible server path supporting
+# SendClusterMessage module API. 
+#
+# ./runtest-moduleapi --single unit/moduleapi/cross-version-cluster --other-server-path tests/tmp/valkey-8-0/valkey-server
+#
+# Test cross version compatibility of cluster module send message API.
+start_server {config "minimal-cluster.conf" start-other-server 1} {
```
zuiderkwast: Why not specify the tags here?

Add the tag `needs:other-server` to skip it if `--other-server-path` is not given, so we don't break the simple `make test`.

```suggestion
start_server {config "minimal-cluster.conf" start-other-server 1 tags {"needs:other-server external:skip cluster modules"}} {
```

src/cluster_legacy.c
```diff
@@ -3016,18 +3034,27 @@ static void clusterProcessPublishPacket(clusterMsgDataPublish *publish_data, uin
...
-
+    serverLog(LL_DEBUG, "Processing light packet of type: %s", clusterGetMessageTypeString(type));
     if (type == CLUSTERMSG_TYPE_PUBLISH || type == CLUSTERMSG_TYPE_PUBLISHSHARD) {
         clusterProcessPublishPacket(&hdr->data.publish.msg, type);
+    } else if (type == CLUSTERMSG_TYPE_MODULE) {
+        uint64_t module_id = hdr->data.module.msg.module_id; /* Endian-safe ID */
+        uint32_t len = ntohl(hdr->data.module.msg.len);
+        uint8_t type = hdr->data.module.msg.type;
+        unsigned char *payload = hdr->data.module.msg.bulk_data;
+        moduleCallClusterReceivers(sender->name, module_id, type, payload, len);
```
zuiderkwast: This is duplicated code, copied from clusterProcessPacket for "heavy-header" module packets. It's not many lines, but I think it's anyway good to avoid duplicated code. We can move these lines part to a new function:

```suggestion
    } else if (type == CLUSTERMSG_TYPE_MODULE) {
        clusterProcessModulePacket(&hdr->data.module.msg);
```

src/cluster_legacy.c
```diff
@@ -4323,22 +4362,50 @@ void clusterSendUpdate(clusterLink *link, clusterNode *node) {
...
+                int msgtype_light = CLUSTERMSG_TYPE_MODULE;
+                msgtype_light |= CLUSTERMSG_LIGHT;
+                msgblock_light = createClusterMsgSendBlock(msgtype_light, msglen_light);
+                clusterMsgLight *hdr = getLightMessageFromSendBlock(msgblock_light);
+                hdr->data.module.msg.module_id = module_id; /* Already endian adjusted. */
+                hdr->data.module.msg.type = type;
+                hdr->data.module.msg.len = htonl(len);
+                memcpy(hdr->data.module.msg.bulk_data, payload, len);
+            }
+            clusterSendMessage(node->link, msgblock_light);
```
zuiderkwast: There is some duplicated code for light and heavy here, right?

This function corresponds to clusterPropagatePublish (for module instead of publish), but this is more code compared to the publish one. I can see that clusterPropagatePublish calls a helper function clusterCreatePublishMsgBlock with an `is_light` flag. It seems to reduce the duplicated code a bit. Maybe do a similar refactoring here?

src/module.c
```diff
@@ -8982,6 +8982,8 @@ void VM_RegisterClusterMessageReceiver(ValkeyModuleCtx *ctx,
 /* Send a message to all the nodes in the cluster if `target` is NULL, otherwise
  * at the specified target, which is a VALKEYMODULE_NODE_ID_LEN bytes node ID, as
  * returned by the receiver callback or by the nodes iteration functions.
+ * This API  uses lesser overhead if the target server support the light message header type (~30B)
+ * or else it uses the regular message overhead (~2KB).
```
zuiderkwast: After a completed rolling upgrade, all nodes use the same message type. I think this scenario with mixed nodes is details that we don't need to mention, as long as the users understand that it works.

Maybe we can mention something in the style we sometimes have in the docs, like "In Valkey 8.1 and later, bla bla bla"...?

```suggestion
 *
 * In Valkey 8.1 and later, the cluster protocol overhead for this message is
 * ~30B, to compare with earlier versions where it's ~2KB.
```

tests/modules/cluster.c
```diff
@@ -34,13 +34,52 @@ int test_cluster_shards(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int arg
     return VALKEYMODULE_OK;
 }
 
+#define MSGTYPE_PING 1
+#define MSGTYPE_PONG 2
+
+/* test.pingall */
+int PingallCommand_ValkeyCommand(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
```
zuiderkwast: This function doesn't follow the naming style in this module. The others are snake case and called `test_cluster_slots` and similar. How about `pingall_command`?

Is there any risk of confusion with this module's PING/PONG and the cluster bus PING/PONG? Maybe call these something else? DING/DONG?

tests/modules/cluster.c
```diff
@@ -34,13 +34,52 @@ int test_cluster_shards(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int arg
...
+    VALKEYMODULE_NOT_USED(argv);
+    VALKEYMODULE_NOT_USED(argc);
+
+    ValkeyModule_SendClusterMessage(ctx, NULL, MSGTYPE_PING, "Hey", 3);
+    return ValkeyModule_ReplyWithSimpleString(ctx, "OK");
+}
+
+/* Callback for message MSGTYPE_PING */
+void PingReceiver(ValkeyModuleCtx *ctx,
```
zuiderkwast: I don't see where this function is called. I can't see where we call `ValkeyModule_RegisterClusterMessageReceiver`. Am I missing something?

tests/unit/moduleapi/cluster.tcl
```diff
@@ -5,6 +5,25 @@ source tests/support/cli.tcl
...
+start_cluster 3 0 [list config_lines $modules] {
+    set node1 [srv 0 client]
+    set node2 [srv -1 client]
+    set node3 [srv -2 client]
+
+    test "Cluster module send message API - VM_SendClusterMessage" {
+        assert_equal OK [$node1 test.pingall]
+        assert_equal 2 [CI 0 cluster_stats_messages_module_sent]
+        wait_for_condition 50 100 {
+            [CI 1 cluster_stats_messages_module_received] eq 1 &&
```
zuiderkwast, line 19: We don't check that the message receiver callback was called. Should we?

An idea is to let `test.pingall` be a blocking command that waits for the pong from each node and returns the number of replies, but maybe that's more complicated. You can decide.

zuiderkwast (commented):
Very nice! Some minor comments.

tests/unit/moduleapi/cross-version-cluster.tcl
hpatro: I expect we will have more `start_server` proc added to the test file and for that we won't need to add tags to each of it.

zuiderkwast: OK, fair enough.

But can we have correct indentation after `tags {...} {`?

The reason we don't have it in some places is that wrapping code was added after the original code and we (they, in the past) didn't want to touch all the lines.

tests/modules/cluster.c
hpatro: It's just copy/paste from `hellocluster.c`. I didn't care much. Let me do a bit of cleanup.

tests/unit/moduleapi/cross-version-cluster.tcl
hpatro: Sure, I thought we didn't want to have an additional tab/depth per line.

tests/unit/moduleapi/cluster.tcl
hpatro: Yeah, we don't. We lack coverage for module message receiver callback. The stat counter ticks only if the payload is valid. So, we confirm that the target is able to process the packet i.e. if it's valid only.

 I think it's beyond the scope of this PR to verify the callback. I will file an (good first) issue.

tests/unit/moduleapi/cross-version-cluster.tcl
zuiderkwast: Could be... Tbh I'm not entirely sure what the reason is but i think correct is better.

src/cluster_legacy.c
```diff
@@ -4319,26 +4358,72 @@ void clusterSendUpdate(clusterLink *link, clusterNode *node) {
...
+            if (msgblock_light == NULL) {
+                msgblock_light = createModuleMsgBlock(module_id, type, payload, len, 1);
+            }
+            clusterSendMessage(node->link, msgblock_light);
+        } else {
+            if (msgblock == NULL) {
+                msgblock = createModuleMsgBlock(module_id, type, payload, len, 0);
+            }
+            clusterSendMessage(node->link, msgblock);
+        }
```
zuiderkwast: Sorry for diving into these details. It's not very important but I can't help seeing the possibility to de-duplicate some more here. Something like this? (untested)

```suggestion
        int is_light = nodeSupportsLightMsgHdrForModule(node);
        clusterMsgSendBlock **block = is_light ? &msgblock_light : &msgblock;
        if (*block == NULL) {
            *block = createModuleMsgBlock(module_id, type, payload, len, is_light);
        }
        clusterSendMessage(node->link, *block);
```

Even more compact (but maybe more obscure?) is to use an array `clusterMsgSendBlock *msgblock[2]` and use `is_light` as an index into that array.

```c
        int is_light = nodeSupportsLightMsgHdrForModule(node);
        if (msgblock[is_light] == NULL) {
            msgblock[is_light] = createModuleMsgBlock(module_id, type, payload, len, is_light);
        }
        clusterSendMessage(node->link, msgblock[is_light]);
```

Is it better? Idk, you can decide.

zuiderkwast (approved):
LGTM.

I added one suggestion but I'm not sure if it's objectively better or not so you can decide. :)

If you feel like adding some test coverage for the module message receiver callback API in another PR, that would be nice.

src/cluster_legacy.c
hpatro: Looks clean. Incorporated.

src/cluster_legacy.c
```diff
@@ -4397,7 +4397,8 @@ static clusterMsgSendBlock *createModuleMsgBlock(int64_t module_id, uint8_t type
  *
  * If link is NULL, then the message is broadcasted to the whole cluster. */
 void clusterSendModule(clusterLink *link, uint64_t module_id, uint8_t type, const char *payload, uint32_t len) {
-    clusterMsgSendBlock *msgblock = NULL, *msgblock_light = NULL;
+    clusterMsgSendBlock *msgblock[CLUSTERMSG_HDR_NUM];
+    memset(msgblock, 0, sizeof(msgblock));
```
zuiderkwast: Equivalent without memset?

```suggestion
    clusterMsgSendBlock *msgblock[CLUSTERMSG_HDR_NUM] = {0};
```

zuiderkwast:
There are some committs from other PRs here now. Some rebased/merge mistake?

If you reset hard to before the last merge commit and then rebase again, the strange ones should disappear. Try that?

Anyway, seems good to merge.

madolson:
Is it intentional the new test isn't getting run anywhere? Only the main test runs with other server, this one doesn't run anywhere.

hpatro:
> There are some committs from other PRs here now. Some rebased/merge mistake?
> 
> If you reset hard to before the last merge commit and then rebase again, the strange ones should disappear. Try that?
> 
> Anyway, seems good to merge.

Looks like a bad merge. Couldn't fix it via merge. Did a rebase and force pushed it.

PR #1574, opened 2025-01-16
Fix memory leak in forgotten node ping ext code path

pieturin:
When processing a cluster bus PING extension, there is a memory leak when adding a new key to the `nodes_black_list` dict. We now make sure to free the key `sds` if the dict did not take ownership of it.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -2936,6 +2936,10 @@ void clusterProcessPingExtensions(clusterMsg *hdr, clusterLink *link) {
             if (n && n != myself && !(nodeIsReplica(myself) && myself->replicaof == n)) {
                 sds id = sdsnewlen(forgotten_node_ext->name, CLUSTER_NAMELEN);
                 dictEntry *de = dictAddOrFind(server.cluster->nodes_black_list, id);
+                if (dictGetKey(de) != id) {
+                    /* The dict did not take ownership of the id string, so we need to free it. */
+                    sdsfree(id);
+                }
                 uint64_t expire = server.unixtime + ntohu64(forgotten_node_ext->ttl);
                 dictSetUnsignedIntegerVal(de, expire);
                 clusterDelNode(n);


pieturin:
@zuiderkwast, could you please take a look at this PR since you did the original implementation of this feature?

zuiderkwast (approved):
Looks correct, thanks!

This happens if some node gossips that a node is forgotten, but we already know it's forgotten. I guess after some time, all nodes know it's forgotten an they gossip this information many times for some time.

How did you find this? Did you run your own tests with address santitizer, valgrind, etc.?

I wonder why we didn't find this leak with our tests and address santitizer.

src/cluster_legacy.c
```diff
@@ -2936,6 +2936,10 @@ void clusterProcessPingExtensions(clusterMsg *hdr, clusterLink *link) {
             if (n && n != myself && !(nodeIsReplica(myself) && myself->replicaof == n)) {
                 sds id = sdsnewlen(forgotten_node_ext->name, CLUSTER_NAMELEN);
                 dictEntry *de = dictAddOrFind(server.cluster->nodes_black_list, id);
+                if (dictGetKey(de) != id) {
+                    /* The dict did not take ownership of the id string, so we need to free it. */
+                    sdsfree(id);
+                }
```
hpatro, line 2942: I don't remember when we introduced this API `dictAddOrFind`. I have been used to performing `dictFind` first and perform an insertion if dict entry is `NULL`.

hpatro (approved):

hpatro:
I think we should backport it as well.

src/cluster_legacy.c
zuiderkwast: Git blame says 2016:

    09a50d34a2 src/dict.c (antirez           2016-09-14 16:43:38 +0200  654) dictEntry *dictAddOrFind(dict *d, void *key) {

pieturin:
> How did you find this? Did you run your own tests with address santitizer, valgrind, etc.?
I wonder why we didn't find this leak with our tests and address santitizer.

Recently some of our internal testing exercising the forgotten node path started failing with ASAN because of this leak. It started failing after I cherry-picked my earlier CLUSTER MEET reliability improvement PRs. Not sure why, but from what I could see, maybe because of some extra packets being exchanged.

I briefly thought about how to write a test for this, but I don't think it would be easy to reliably hit this condition.

PR #1575, opened 2025-01-17
Backport "Fix memory leak in forgotten node ping ext code path" to 8.0

pieturin:
When processing a cluster bus PING extension, there is a memory leak when adding a new key to the `nodes_black_list` dict. We now make sure to free the key `sds` if the dict did not take ownership of it.

Backport of #1574 from unstable.


pieturin:
Failing tests seem unrelated.

zuiderkwast (approved):

PR #1578, opened 2025-01-17
Add `FUNCTION LOAD` and `FCALL` benchmark

rjd15372:
This commit adds two new benhchmarks to `valkey-benchmark` tool.

The `FUNCTION LOAD` benchmark evaluates the performance of loading new Lua scripts. This benchmark is parameterized by the number of functions that are present in the script. To control the number of functions we can use the CLI option `--num-functions-in-script`.

The `FCALL` benchmarks evaluates the perfomance of calling a Lua function of a previously loaded script. This benchmark is parameterized by the number of keys that are passed in the `FCALL` command. To control the number of keys we can use the CLI option `--num-keys-in-fcall`.

### DIFF
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -129,7 +129,7 @@ static struct config {
     _Atomic int is_updating_slots;
     _Atomic int slots_last_update;
     int enable_tracking;
-    int num_functions_in_script;
+    int num_functions;
     int num_keys_in_fcall;
     pthread_mutex_t liveclients_mutex;
     pthread_mutex_t is_updating_slots_mutex;
@@ -1428,8 +1428,8 @@ int parseOptions(int argc, char **argv) {
                 goto invalid;
         } else if (!strcmp(argv[i], "--enable-tracking")) {
             config.enable_tracking = 1;
-        } else if (!strcmp(argv[i], "--num-functions-in-script")) {
-            config.num_functions_in_script = atoi(argv[++i]);
+        } else if (!strcmp(argv[i], "--num-functions")) {
+            config.num_functions = atoi(argv[++i]);
         } else if (!strcmp(argv[i], "--num-keys-in-fcall")) {
             config.num_keys_in_fcall = atoi(argv[++i]);
         } else if (!strcmp(argv[i], "--help")) {
@@ -1559,11 +1559,10 @@ usage:
         " -I                 Idle mode. Just open N idle connections and wait.\n"
         " -x                 Read last argument from STDIN.\n"
         " --seed <num>       Set the seed for random number generator. Default seed is based on time.\n"
-        " --num-functions-in-script <num> Sets the number of functions present in the Lua lib\n"
-        "                                 that is loaded when running the 'function_load' test.\n"
-        "                                 (default 10).\n"
-        " --num-keys-in-fcall <num>       Sets the number of keys passed to FCALL command when\n"
-        "                                 when running the 'fcall' test. (default 1)\n",
+        " --num-functions <num>     Sets the number of functions present in the Lua lib that is\n"
+        "                           loaded when running the 'function_load' test. (default 10).\n"
+        " --num-keys-in-fcall <num> Sets the number of keys passed to FCALL command when running\n"
+        "                           the 'fcall' test. (default 1)\n",
         tls_usage,
         " --help             Output this help and exit.\n"
         " --version          Output version and exit.\n\n"
@@ -1637,7 +1636,7 @@ char *generateFunctionScript(uint32_t num_functions, int with_keys) {
     int written = snprintf(buffer, buffer_len, "#!lua name=benchlib\n");
     while (num_functions > 0 && (buffer_len - written) > 0) {
         assert(buffer_len - written > 0);
-        size_t n = 0;
+        int n = 0;
         if (with_keys) {
             n = snprintf(buffer + written, buffer_len - written,
                          "local function foo%u(keys, args)\nreturn keys[0]\nend\n",
@@ -1648,7 +1647,7 @@ char *generateFunctionScript(uint32_t num_functions, int with_keys) {
                          num_functions);
         }
 
-        if (n < 0 || n >= buffer_len - written) {
+        if (n < 0 || (size_t)n >= buffer_len - written) {
             break;
         }
         written += n;
@@ -1726,7 +1725,7 @@ int main(int argc, char **argv) {
     config.is_updating_slots = 0;
     config.slots_last_update = 0;
     config.enable_tracking = 0;
-    config.num_functions_in_script = 10;
+    config.num_functions = 10;
     config.num_keys_in_fcall = 1;
     config.resp3 = 0;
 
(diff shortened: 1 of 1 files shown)


rjd15372:
I'll be adding more scripting engine related benchmarks to `valkey-benchmark` in follow up PRs

zuiderkwast:
Do these run by default or only when specified with `-t`?

Maybe we should consider excluding them by default and only run them when explicitly requested?

rjd15372:
> Do these run by default or only when specified with `-t`? 

Yes

> Maybe we should consider excluding them by default and only run them when explicitly requested?

I can do that, but why do you think we should not run them by default?

zuiderkwast:
> > Maybe we should consider excluding them by default and only run them when explicitly requested?
> 
> I can do that, but why do you think we should not run them by default?

We could, but it seems like a bigger change somehow. I guess we should have a core team decision since these tools are also part of the public API.

How slow are the new tests compared to the existing ones for SET and GET, for let's say `-n 10000000`?

zuiderkwast:
@madolson What do you think about adding more tests to valkey-benchmark? Can it break any user scenarios if we run them by default? It makes the total run take longer and more/different output to be printed...

src/valkey-benchmark.c
```diff
@@ -1617,6 +1628,34 @@
...
+    /* 64K buffer to hold script code */
+    const size_t buffer_len = 64 * 1024;
+    char *buffer = zmalloc(buffer_len);
+    memset(buffer, 0, buffer_len);
+
+    int written = snprintf(buffer, buffer_len, "#!lua name=benchlib\n");
+    while (num_functions > 0) {
+        assert(buffer_len - written > 0);
+        if (with_keys) {
+            written += snprintf(buffer + written, buffer_len - written,
```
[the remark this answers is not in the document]
zuiderkwast: Can we rewrite this in a way the checker understands or should I just dismiss it?

src/valkey-benchmark.c
```diff
@@ -1552,7 +1558,12 @@ int parseOptions(int argc, char **argv) {
         "                    on the command line.\n"
         " -I                 Idle mode. Just open N idle connections and wait.\n"
         " -x                 Read last argument from STDIN.\n"
-        " --seed <num>       Set the seed for random number generator. Default seed is based on time.\n",
+        " --seed <num>       Set the seed for random number generator. Default seed is based on time.\n"
+        " --num-functions-in-script <num> Sets the number of functions present in the Lua script\n"
+        "                                 that is loaded when running the 'function_load' test.\n"
+        "                                 (default 10).\n"
```
zuiderkwast: The word "script" is used only for EVAL-related scripts in the docs and command names (like SCRIPT LOAD).

Library (or lib) might be the right word.

zuiderkwast (commented):
Looks good in general and after the last meeting, it's decided that it's not a major change.

I hope we can silence the CodeQL warnings.

src/valkey-benchmark.c
```diff
@@ -1559,7 +1559,7 @@ int parseOptions(int argc, char **argv) {
         " -I                 Idle mode. Just open N idle connections and wait.\n"
         " -x                 Read last argument from STDIN.\n"
         " --seed <num>       Set the seed for random number generator. Default seed is based on time.\n"
-        " --num-functions-in-script <num> Sets the number of functions present in the Lua script\n"
+        " --num-functions-in-script <num> Sets the number of functions present in the Lua lib\n"
```
```suggestion
        " --num-functions-in-lib <num> Sets the number of functions present in the Lua lib\n"
```

zuiderkwast: Or just `--num-functions`?

rjd15372: I've chose to go with `--num-functions`

src/valkey-benchmark.c
```diff
@@ -1617,6 +1627,43 @@ long long showThroughput(struct aeEventLoop *eventLoop, long long id, void *clie
...
+    char *buffer = zmalloc(buffer_len);
+    memset(buffer, 0, buffer_len);
+
+    int written = snprintf(buffer, buffer_len, "#!lua name=benchlib\n");
+    while (num_functions > 0 && (buffer_len - written) > 0) {
+        assert(buffer_len - written > 0);
+        int n = 0;
+        if (with_keys) {
+            n = snprintf(buffer + written, buffer_len - written,
+                         "local function foo%u(keys, args)\nreturn keys[0]\nend\n",
```
zuiderkwast, line 1644: This script is quite trivial. It tests only the overhead of starting the script runtime, etc.

If we want to compare Lua5.x to LuaJIT at some point, we may want more fancy scripts such as the ones in this comment: https://github.com/valkey-io/valkey/issues/1229#issuecomment-2439984030. We can do that later I guess.

src/valkey-benchmark.c
```diff
@@ -1958,6 +2007,51 @@ int main(int argc, char **argv) {
             free(cmd);
         }
 
+        if (test_is_selected("function_load")) {
+            char *script = generateFunctionScript(config.num_functions, 0);
+            len = redisFormatCommand(&cmd, "function load replace %s", script);
+            benchmark("FUNCTION LOAD", cmd, len);
```
zuiderkwast, line 2015: Load the same script many times? We could optimize this case, but that would make the benchmark useless. :laughing:

src/valkey-benchmark.c
```diff
@@ -1958,6 +2007,51 @@ int main(int argc, char **argv) {
...
+            }
+            zfree(cmd_argv);
+
+            benchmark("FCALL", cmd, len);
+            free(cmd);
+        }
+
+        if (test_is_selected("fcall")) {
+        }
+
```
zuiderkwast: Cleanup?

```suggestion
```

src/valkey-benchmark.c
```diff
@@ -1552,7 +1558,11 @@ int parseOptions(int argc, char **argv) {
         "                    on the command line.\n"
         " -I                 Idle mode. Just open N idle connections and wait.\n"
         " -x                 Read last argument from STDIN.\n"
-        " --seed <num>       Set the seed for random number generator. Default seed is based on time.\n",
+        " --seed <num>       Set the seed for random number generator. Default seed is based on time.\n"
+        " --num-functions <num>     Sets the number of functions present in the Lua lib that is\n"
+        "                           loaded when running the 'function_load' test. (default 10).\n"
+        " --num-keys-in-fcall <num> Sets the number of keys passed to FCALL command when running\n"
+        "                           the 'fcall' test. (default 1)\n",
```
zuiderkwast: In man pages, option lists are rendered with a line break when the option doesn't fit to the left of the aligned descriptions, similar to this:

```suggestion
        " --seed <num>       Set the seed for random number generator. Default seed is based on time.\n"
        " --num-functions <num>\n"
        "                    Sets the number of functions present in the Lua lib that is\n"
        "                    loaded when running the 'function_load' test. (default 10).\n"
        " --num-keys-in-fcall <num>\n"
        "                    Sets the number of keys passed to FCALL command when running\n"
        "                    the 'fcall' test. (default 1)\n",
```

I think it looks better. WDYT?

zuiderkwast (approved):
I think this looks fine to merge.

zuiderkwast:
We got this failure in Daily, tests with TLS: https://github.com/valkey-io/valkey/actions/runs/13253032412/job/36994779502#step:6:5044. Not sure if it's related to this PR but it might be. No other PR has anything to do with valkey-benchmark.

```
*** [err]: benchmark: full test suite in tests/integration/valkey-benchmark.tcl
valkey-benchmark non zero code. first line: 
```

zuiderkwast:
Fix here: https://github.com/valkey-io/valkey/pull/1710

PR #1579, opened 2025-01-17
Embed hash value in hash type entry

### LINKED ISSUE #1551: Follow-up of #1502: Embed field and value in hashTypeEntry
When field + value fit within a cache line, embed both of them in the same allocation. We use fewer allocations and save the memory of a pointer we save a memory access when reading the value.

Follow-up of #1502.

zuiderkwast:
For a hash type key represented as a hash table, embed the field and value in one allocation when they fit together in an allocation of up to 128 bytes. For larger fields and values, another layout is used where the value is a separately allocated sds string.

**Implementation**

The hashTypeEntry pointer is changed to be the field sds, i.e. a pointer to the embedded field content, which is located after the sds header in memory. We encode the entry layout in the some unused bits in the sds header.

Entry with embedded field and value, used when they're both small. The value is stored as SDS_TYPE_8. The field can use any SDS type.

    +--------------+---------------+
    | field        | value         |
    | hdr "foo" \0 | hdr8 "bar" \0 |
    +------^-------+---------------+
           |
           |
         entry pointer = field sds

Entry with value-pointer and embedded field, used for larger field-value pairs. The field is SDS type 8 or higher.

    +-------+--------------+
    | value | field        |
    | ptr   | hdr "foo" \0 |
    +-------+------^-------+
                   |
                   |
                entry pointer = field sds

An embedded sds5 field implies the first entry layout.

Fixes #1551.


src/t_hash.c
```diff
@@ -29,50 +29,210 @@
...
+        hashTypeEntryPtrValue *entry = zmalloc(alloc_size);
+        entry->value = value;
+        if (field_is_sds5) {
+            /* We can't use SDS_TYPE_5 to encode extra information in the unused
+             * allocation size, so convert to SDS_TYPE_8. */
+            struct sdshdr8 *sh = (void *)entry->field_data;
+            sh->flags = SDS_TYPE_8;
+            sh->len = sdslen(field);
+            embedded_field_sds = (sds)(entry->field_data + sizeof(struct sdshdr8));
+            memcpy(embedded_field_sds, field, sdslen(field) + 1);
```
ranshid: May I suggest something?
We can add another type of sdscopytobuffer which forces a specific type and can also be used by sdscopytobuffer:

```
/* This method copies the sds `s` into `buf` which is the target character buffer. */
size_t sdscopytobufferwithtype(unsigned char *buf, size_t buf_len, char type, const sds s, uint8_t *hdr_size) {
    assert((s[-1] & SDS_TYPE_MASK) <= type);
    size_t header_len = sdsHdrSize(type);
    size_t str_len = sdslen(s);
    size_t alloc_size = header_len + str_len + 1;
    if (hdr_size) *hdr_size = sdsHdrSize(type);
    if (buf == NULL) {
        return alloc_size;
    }
    assert(buf_len >= alloc_size);
    sds embed_sds = (sds)(buf + header_len);
    embed_sds[-1] = type & SDS_TYPE_MASK;
    sdssetlen(embed_sds, str_len);
    sdssetalloc(embed_sds, alloc_size);
    memcpy(embed_sds, s, str_len + 1);
    return alloc_size;
}
```

src/t_hash.c
```diff
@@ -29,50 +29,210 @@
...
+ * ENTRY_ENC_PTR_VALUE, used for larger fields and values:
+ *
+ *     +-------+--------------+
+ *     | value | field        |
+ *     | ptr   | hdr "foo" \0 |
+ *     +-------+------^-------+
+ *                    |
+ *                    |
+ *                 entry pointer = field sds
+ */
```
ranshid: Sorry for turning into Scrooge McDuck.
I actually thought about this the opposite way around. I think we can FORCE non sds5 fields only when the value IS embedded.
So the question of: canUseEmbeddedValueEntry can take into account the new field sds header length and the value sds size.
This may also allow us to use the alloc of embedded fields to hold the value header size, so when the field sds alloc is 0 I actually know this is a ENTRY_ENC_PTR_VALUE and in case it is not 0 I know it is ENTRY_ENC_EMB_VALUE.

zuiderkwast: Regarding maximum greed, this is how I'm thinking:

For small values, the relative overhead is large. For large values, the relative overhead is small. Therefore, I want to minimize the overhead for small field-value pairs, which are the embedded values. Some use case of hashes is to store a numeric value, which is only a few bytes as a string. That's why we have things like HINCRBY.

Since sds5 can hold up to 32 bytes, many of the field-value pairs can be sds5 + sds5 and fit into a 16, 24 or 32-byte allocation. We lose one byte if we use sds8 instead of sds5 for the field here. (We lose two bytes for using sds8 instead of sds5 and gain one for skipping value-hdr-size.)

Of course the optimal design depends on the size distribution of fields and values used in practice. Maybe it's actually better to optimize for a larger field and value sizes. Do you have any statistics on real world field and value size distributions?

zuiderkwast: Now, I picked sds8 for embedded value. Field can be sds5 in this case. See updated top comment. This choice maximizes the size of values we can store embedded in a cache line.

src/sds.h
```diff
@@ -95,6 +95,22 @@ static inline unsigned char sdsType(const_sds s) {
...
+    return sdsType(s) == SDS_TYPE_5 ? 0 : flags >> SDS_TYPE_BITS;
+}
+
+/* Allow storing user data in some unused bits in the SDS header, except for
+ * SDS_TYPE_5. Aux is a 5-bit unsigned integer, a value between 0 and 31. The
+ * aux value is lost if the SDS is auto-resized. */
+static inline void sdsSetAuxBits(sds s, unsigned char aux) {
+    if (sdsType(s) == SDS_TYPE_5) return;
+    unsigned char flags = s[-1];
+    s[-1] = (char)((flags & SDS_TYPE_MASK) | ((aux << SDS_TYPE_BITS) & ~SDS_TYPE_MASK));
```
ranshid: I would prefer we had a specific API to set AUX bit and override/get the complete AUX value.
I think managing the entire 5 bits as a single value is helpful maintaining multiple distinctive encoding options, but for the case were we might have different optiong which can coexist (eg embedded value = yes/no, ttl = yes/no ... ) it might not give much value. also it would make it harder to only set a single bit since we would have to reag the AUX value turn the bit and write it back. 
I would suggest:
void setAuxBit(sds s, int b, int on)
int getAuxBit(sds s, int b)
void setAuxVal(sds s, char val)
char getAuxVal(sds s)

src/t_hash.c
```diff
@@ -29,52 +29,202 @@
...
+         *     +--------------+---------------+
+         */
+        size_t min_size = field_size + value_size;
+        size_t buf_size;
+        char *buf = zmalloc_usable(min_size, &buf_size);
+        sdswrite(buf, field_size, field_sds_type, field, field_len);
+        sdswrite(buf + field_size, buf_size - field_size, SDS_TYPE_8, value, value_len);
+        embedded_field_sds = (sds)(buf + sdsHdrSize(field_sds_type));
+        /* Field sds aux bits are zero, which we use for this entry encoding. */
+        serverAssert(entryGetEncoding(embedded_field_sds) == ENTRY_ENC_EMB_VALUE);
```
ranshid: maybe for the sake of symmetry it is better to explicitly set the AUX bit and not count on it being zero?

src/t_hash.c
```diff
@@ -29,52 +29,202 @@
...
-};
+} hashTypeEntryPtrValue;
+
+static inline hashTypeEntryEnc entryGetEncoding(const hashTypeEntry *entry) {
+    return sdsGetAuxBits(entry);
+}
+
+/* Returns the containing struct for an entry without embedded value. */
+static hashTypeEntryPtrValue *getEntryStruct(const hashTypeEntry *entry) {
+    serverAssert(entryGetEncoding(entry) == ENTRY_ENC_PTR_VALUE);
```
ranshid: Why not create a single function that can return the entry pointer for any encoding type?

ranshid (commented):
Still not a complete review - I still might have some comments about hashTypeEntryReplaceValue which I am looking into

src/t_hash.c
zuiderkwast: ~~I did an `assert(type != sds5)` in the setter, but we can allow setting them to zero I suppose.~~

I forgot what I did. :) The setter is a no-op for sds5...

src/sds.h
zuiderkwast: OK, in this PR we're only using one bit so adding all four functions seems like over engineering. setAuxBit and getAuxBit is sufficient for now.

src/t_hash.c
zuiderkwast: There is one: hashTypeEntryAllocPtr.

There is no struct for the embedded field + value entry. For that one, only a char array is used. A struct isn't useful in this case.

src/t_hash.c
zuiderkwast: I don't really see the value in calling a noop setter to set the bit to zero though. It's likely just confusing to the reader. We still rely on the fact that sds5 doesn't store any aux bits.

My thought is that the assert here catches any future changes that might break this encoding.

It's theoretically possible to encode two aux bits in sds5 if we change how the SDS type is encoding using 3 bits, but it's more complex. Currently it's just a bit mask but it would have to be different cases. I don't think it's worth the complexity.

src/t_hash.c
```diff
@@ -29,52 +29,206 @@
...
+
+#define FIELD_SDS_AUX_BIT_ENTRY_HAS_PTR_VALUE 0
+
+/* Struct used for ENTRY_ENC_PTR_VALUE. */
+typedef struct {
     sds value;
-    char field_offset;
     char field_data[];
-};
+} hashTypeEntryPtrValue;
```
ranshid: IMO this only leads to some confusion when for fully embedded entry we use some unstructured array of characters and for partially embedded we use some structured structure. suggest to remove this.

src/t_hash.c
```diff
@@ -29,52 +29,206 @@
...
+    return sdsGetAuxBit(entry, FIELD_SDS_AUX_BIT_ENTRY_HAS_PTR_VALUE);
+}
+
+/* Returns the containing struct for an entry without embedded value. */
+static hashTypeEntryPtrValue *getEntryStruct(const hashTypeEntry *entry) {
+    serverAssert(entryGetEncoding(entry) == ENTRY_ENC_PTR_VALUE);
+    unsigned char *buf = sdsAllocPtr(entry);
+    buf -= offsetof(hashTypeEntryPtrValue, field_data);
+    return (void *)buf;
+}
```
ranshid: I understand the motivation but IMO we should use a more explicit name for this:

```suggestion
static inline sds *encodedPtrValueEntryGetValueref(const hashTypeEntry *entry) {
    serverAssert(entryGetEncoding(entry) == ENTRY_ENC_PTR_VALUE);
    return (sds *)(sdsAllocPtr((sds)entry)) - 1;
}
```

src/t_hash.c
```diff
@@ -29,52 +29,206 @@
...
+         *     | field        | value         |
+         *     | hdr "foo" \0 | hdr8 "bar" \0 |
+         *     +--------------+---------------+
+         */
+        size_t min_size = field_size + value_size;
+        size_t buf_size;
+        char *buf = zmalloc_usable(min_size, &buf_size);
+        sdswrite(buf, field_size, field_sds_type, field, field_len);
+        sdswrite(buf + field_size, buf_size - field_size, SDS_TYPE_8, value, value_len);
+        embedded_field_sds = (sds)(buf + sdsHdrSize(field_sds_type));
```
ranshid:
```suggestion
        embedded_field_sds = sdswrite(buf, field_size, field_sds_type, field, field_len);
        sdswrite(buf + field_size, buf_size - field_size, SDS_TYPE_8, value, value_len);
```

src/t_hash.c
```diff
@@ -29,52 +29,206 @@
...
+         *     +-------+---------------+
+         */
+        char field_sds_type = sdsReqType(field_len);
+        if (field_sds_type == SDS_TYPE_5) field_sds_type = SDS_TYPE_8;
+        field_size = sdsReqSize(field_len, field_sds_type);
+        size_t alloc_size = sizeof(void *) + field_size;
+        hashTypeEntryPtrValue *entry = zmalloc(alloc_size);
+        entry->value = value;
+        sdswrite(entry->field_data, field_size, field_sds_type, field, field_len);
+        embedded_field_sds = (sds)(entry->field_data + sdsHdrSize(field_sds_type));
```
ranshid:
```suggestion
        size_t alloc_size = sizeof(value) + field_size;
        char *buf = zmalloc(alloc_size);
        memcpy(buf, &value, sizeof(value));
        embedded_field_sds = sdswrite(buf + sizeof(value), field_size, field_sds_type, field, field_len);
```

src/t_hash.c
```diff
@@ -29,52 +29,206 @@
...
+    switch (entryGetEncoding(entry)) {
+    case ENTRY_ENC_EMB_VALUE: {
+        /* Skip field content, field null terminator and value sds8 hdr. */
+        size_t offset = sdslen(entry) + 1 + sdsHdrSize(SDS_TYPE_8);
+        char *buf = (char *)entry + offset;
+        return buf;
+    }
+    case ENTRY_ENC_PTR_VALUE: {
+        const hashTypeEntryPtrValue *entry_struct = getEntryStruct(entry);
+        return entry_struct->value;
```
ranshid:
```suggestion
        sds *value_ref = encodedPtrValueEntryGetValueref(entry);
        return *value_ref;
```

src/t_hash.c
```diff
@@ -29,52 +29,206 @@
...
-static void hashTypeEntryReplaceValue(hashTypeEntry *entry, sds value) {
-    sdsfree(entry->value);
-    entry->value = value;
+/* Returns the address of the entry allocation. */
+static void *hashTypeEntryAllocPtr(hashTypeEntry *entry) {
+    switch (entryGetEncoding(entry)) {
+    case ENTRY_ENC_EMB_VALUE:
+        return sdsAllocPtr(entry);
+    case ENTRY_ENC_PTR_VALUE:
+        return getEntryStruct(entry);
```
ranshid:
```suggestion
        return encodedPtrValueEntryGetValueref(entry);
```

src/t_hash.c
```diff
@@ -29,52 +29,206 @@
...
+        if (field_size + value_size <= CACHE_LINE_SIZE) {
+            /* Convert to entry with embedded value. */
+            hashTypeEntry *new_entry = hashTypeCreateEntry(field, value);
+            freeHashTypeEntry(entry);
+            return new_entry;
+        } else {
+            /* Not embedded value. */
+            hashTypeEntryPtrValue *entry_struct = getEntryStruct(entry);
+            sdsfree(entry_struct->value);
+            entry_struct->value = value;
```
ranshid:
```suggestion
            sds *value_ref = encodedPtrValueEntryGetValueref(entry);
            sdsfree(*value_ref);
            *value_ref = value;
```

src/t_hash.c
```diff
@@ -29,52 +29,206 @@
...
+
+/* Returns memory usage of a hashTypeEntry, including all allocations owned by
+ * the hashTypeEntry. */
+size_t hashTypeEntryMemUsage(hashTypeEntry *entry) {
+    switch (entryGetEncoding(entry)) {
+    case ENTRY_ENC_EMB_VALUE:
+        return zmalloc_usable_size(sdsAllocPtr(entry));
+    case ENTRY_ENC_PTR_VALUE: {
+        hashTypeEntryPtrValue *entry_struct = getEntryStruct(entry);
+        return zmalloc_usable_size(entry_struct) + sdsAllocSize(entry_struct->value);
```
ranshid:
```suggestion
        return zmalloc_usable_size(entry_struct) + sdsAllocSize(hashTypeEntryGetValue(entry));
```

src/t_hash.c
```diff
@@ -85,25 +239,37 @@ size_t hashTypeEntryAllocSize(hashTypeEntry *entry) {
...
-
-    return new_entry;
+    void *alloc_ptr = hashTypeEntryAllocPtr(entry);
+    switch (entryGetEncoding(entry)) {
+    case ENTRY_ENC_EMB_VALUE:
+        break;
+    case ENTRY_ENC_PTR_VALUE: {
+        hashTypeEntryPtrValue *entry_struct = alloc_ptr;
+        sds new_value = sdsdefragfn(entry_struct->value);
+        if (new_value) entry_struct->value = new_value;
```
ranshid:
```suggestion
        sds *value_ref = encodedPtrValueEntryGetValueref(entry);
        sds new_value = sdsdefragfn(*value_ref);
        if (new_value) *value_ref = new_value;
```

src/t_hash.c
```diff
@@ -85,25 +239,37 @@ size_t hashTypeEntryAllocSize(hashTypeEntry *entry) {
...
+    return defragfn(alloc_ptr);
 }
 
 /* Used for releasing memory to OS to avoid unnecessary CoW. Called when we've
  * forked and memory won't be used again. See zmadvise_dontneed() */
 void dismissHashTypeEntry(hashTypeEntry *entry) {
     /* Only dismiss values memory since the field size usually is small. */
-    dismissSds(entry->value);
+    if (entryGetEncoding(entry) == ENTRY_ENC_PTR_VALUE) {
+        dismissSds(getEntryStruct(entry)->value);
```
ranshid:
```suggestion
        dismissSds(hashTypeEntryGetValue(entry));
```

src/t_hash.c
```diff
@@ -85,25 +239,37 @@ size_t hashTypeEntryAllocSize(hashTypeEntry *entry) {
...
+    }
 }
 
 void freeHashTypeEntry(hashTypeEntry *entry) {
-    sdsfree(entry->value);
-    zfree(entry);
+    void *alloc_ptr = hashTypeEntryAllocPtr(entry);
+    if (entryGetEncoding(entry) == ENTRY_ENC_PTR_VALUE) {
+        sdsfree(((hashTypeEntryPtrValue *)alloc_ptr)->value);
```
ranshid:
```suggestion
        sdsfree(hashTypeEntryGetValue(entry));
```

ranshid (commented):
Sorry for neglecting this (in favor of 8.1 release...)
Overall I think it looks fine.
IMO there were some parts which caused me confusion, so I suggested some minor refactoring.

src/t_hash.c
zuiderkwast: Sure, working with char-ptr is fine when it's only one pointer and an embedded field. I can change to that.

If we'll have more fields in a future variant though, like ttl, value-ptr and embedded field, then at some point a struct is probably worth it... But that's a future problem.

src/t_hash.c
zuiderkwast: Sure, if accessing the value is the only thing it's used for... I'll look tomorrow.

src/t_hash.c
zuiderkwast: No, the function name `encodedPtrValueEntryGetValueref` is for finding the value, not the start of the allocation. It's an abuse of the abstraction, which is more confusing than just repeating the calculation here, like

```suggestion
        return sdsAllocPtr(entry) - sizeof(sds *);
```

src/sds.h
```diff
@@ -95,6 +95,27 @@ static inline unsigned char sdsType(const_sds s) {
...
+/* Returns a user data bit stored in the SDS header by sdsSetAuxBit. The bit
+ * index is 0-4. Returns 0 or 1. Always returns 0 for SDS_TYPE_5. */
+static inline int sdsGetAuxBit(const_sds s, int bit) {
+    unsigned char flags = s[-1];
+    return sdsType(s) == SDS_TYPE_5 ? 0 : flags >> (SDS_TYPE_BITS + bit);
+}
+
+/* Stores a bit in an unused area in the SDS header, except for SDS_TYPE_5. The
+ * bit index is 0-4. The value is 0 or 1. The aux bits are lost if the SDS is
+ * auto-resized. */
```
ranshid: I agree this worth mentioning. the bits are more fit to be used for immutable strings (in that regard creating an "immutable string type" or "field" type would make sense, but maybe we can do this refactor in a separate PR.

src/t_hash.c
```diff
@@ -29,52 +29,183 @@
...
+    return sdsGetAuxBit(entry, FIELD_SDS_AUX_BIT_ENTRY_HAS_VALUE_PTR);
+}
+
+/* Returns the location of a pointer to a separately allocated value. Only for
+ * an entry without an embedded value. */
+static sds *hashTypeEntryGetValueRef(const hashTypeEntry *entry) {
+    serverAssert(entryHasValuePtr(entry));
+    char *field_data = sdsAllocPtr(entry);
+    field_data -= sizeof(sds *);
+    return (sds *)(void *)field_data;
```
ranshid: did not understand this....
```suggestion
    return (sds *)field_data;
```

src/t_hash.c
```diff
@@ -29,52 +29,183 @@
...
 }
 
 sds hashTypeEntryGetValue(const hashTypeEntry *entry) {
-    return entry->value;
+    if (entryHasValuePtr(entry)) {
+        return *hashTypeEntryGetValueRef(entry);
+    } else {
+        /* Skip field content, field null terminator and value sds8 hdr. */
+        size_t offset = sdslen(entry) + 1 + sdsHdrSize(SDS_TYPE_8);
+        char *buf = (char *)entry + offset;
```
ranshid: nit: I cannot explain why it bothers me :) 
```suggestion
        sds buf = (char *)entry + offset;
```

src/t_hash.c
```diff
@@ -29,52 +29,183 @@
...
+    size_t value_len = sdslen(value);
+    size_t value_size = sdsReqSize(value_len, SDS_TYPE_8);
+    if (!entryHasValuePtr(entry)) {
+        /* Reuse the allocation if the new value fits and leaves no more than
+         * 25% unused space after replacing the value. */
+        char *alloc_ptr = sdsAllocPtr(entry);
+        size_t required_size = field_size + value_size;
+        size_t alloc_size;
+        if (required_size <= EMBED_VALUE_MAX_ALLOC_SIZE &&
+            required_size <= (alloc_size = zmalloc_usable_size(alloc_ptr)) &&
```
ranshid: Just a small note. I prefer this implementation because of the simplicity.
I would say that we could only compare the value size diff (the current VS the new value) since the field is immutable and the size is only impacted by the values diff. Only stating this since IIRC zmalloc_usable_size has some notable performance impact

ranshid (approved):
Overall LGTM

Some minor comments so you maybe consider

src/t_hash.c
zuiderkwast: It's forbidden to cast from `char *` to `sds *`. They're not compatible pointer types. Only `void *` is compatible with all pointer types.

src/t_hash.c
zuiderkwast: You need to explain. Otherwise, I don't take your comment as valid. :laughing:

An `sds` variable named `buf` would bother me. `char *buf` is used when working with the underlying bytes. When working with the sds string abstractions, I'd call it `sds value` or something.

I'm fine with `return (char *)entry + offset;` though, or if you insist `return (sds)((char *)entry + offset);`

src/t_hash.c
ranshid: mmm. O.K maybe some compilers will complain here (my clang did not). Go ahead.

ranshid: BTW - I think since field_data is (char * ) than there is no issue here. sds is simply an alias for char* so I think there should be no issue.

src/t_hash.c
ranshid: I think for the same reason you choose to cast char* to sds* via void*. there is no real type change only alias change right? It just seem like we should return the same "type" even though there is not special need to since sds is an alias to char*

ranshid: but as I said. not important...

src/sds.h
zuiderkwast: Yeah, this is already quite low level. I'll add a sentence like *This is only for special uses like immutable SDS embedded in other structures.*

src/t_hash.c
zuiderkwast: There is an issue. We want to return `sds *`, not `sds`.

`sds *` means `char **`.

~~I already got a compile warning here. That's why I know.~~

[Edit] Hm... maybe it was't here. Let me do some research about these casting rules...

zuiderkwast: With just `return field_data;` and hoping for an implicit cast to `sds *` I get this error:

```
error: returning ‘char *’ from a function with incompatible return type ‘char **’ [-Wincompatible-pointer-types]
```

ranshid: well. I did suggest to return (sds*) field_data for that reason.

src/t_hash.c
zuiderkwast: Here you're right that `sds = char *` so there's not really a cast. I think the intuition and the naming is somewhat relevant here though. Your suggestion is good but I'll change `char *buf` to `sds value`. Or possibly to make it easier to follow why we get this offset, it can be done in multiple steps...

```C
        char *buf = entry;
        buf += sdslen(entry) + 1;      /* Move to after the embedded field */
        buf += sdsHdrSize(SDS_TYPE_8); /* Skip value header */
        return buf;
```

Or for brevity I'll stick with a single line, since there are some drawings and comments about this elsewhere.

```C
        return (char *)entry + offset;
```

src/t_hash.c
zuiderkwast: I managed to get rid of zmalloc_usable_size. We encode the unused space in the embedded value SDS header's alloc field. It can be used.

PR #1582, opened 2025-01-18
Add support for MustObeyClient Module API

KarthikSubbarao:
This PR adds support for MustObeyClient Module API.

The purpose of this API is for Modules to handle commands / callbacks knowing whether commands are arriving from the primary client or AOF client and should never be rejected.

A use case is that Modules can have validation logic in command handlers which only should be executed on primary nodes. Replica nodes must obey commands replicated.

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -3719,6 +3719,14 @@ ValkeyModuleString *VM_GetClientUserNameById(ValkeyModuleCtx *ctx, uint64_t id) 
     return str;
 }
 
+/* Returns 1 if commands are arriving from the primary client or AOF client
+ * and should never be rejected.
+ * Returns 0 otherwise (and also if ctx or if the client is NULL). */
+int VM_MustObeyClient(ValkeyModuleCtx *ctx) {
+    if (!ctx || !ctx->client) return 0;
+    return mustObeyClient(ctx->client);
+}
+
 /* This is a helper for VM_GetClientInfoById() and other functions: given
  * a client, it populates the client info structure with the appropriate
  * fields depending on the version provided. If the version is not valid
@@ -13839,6 +13847,7 @@ void moduleRegisterCoreAPI(void) {
     REGISTER_API(ChannelAtPosWithFlags);
     REGISTER_API(GetClientId);
     REGISTER_API(GetClientUserNameById);
+    REGISTER_API(MustObeyClient);
     REGISTER_API(GetContextFlags);
     REGISTER_API(AvoidReplicaTraffic);
     REGISTER_API(PoolAlloc);
--- a/src/module.c
+++ b/src/module.c
@@ -3721,6 +3721,8 @@ ValkeyModuleString *VM_GetClientUserNameById(ValkeyModuleCtx *ctx, uint64_t id) 
 
 /* Returns 1 if commands are arriving from the primary client or AOF client
  * and should never be rejected.
+ * This check can be used in places such as skipping validation of commands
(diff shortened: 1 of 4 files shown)


src/module.c
```diff
@@ -3719,6 +3719,14 @@ ValkeyModuleString *VM_GetClientUserNameById(ValkeyModuleCtx *ctx, uint64_t id)
     return str;
 }
 
+/* Returns 1 if commands are arriving from the primary client or AOF client
+ * and should never be rejected.
+ * Returns 0 otherwise (and also if ctx or if the client is NULL). */
+int VM_MustObeyClient(ValkeyModuleCtx *ctx) {
+    if (!ctx || !ctx->client) return 0;
```
ranshid, line 3728: My only fear is that it will become confusing since many times the context holds a fake/tmp client which will not be obeyed right? I think this is why in most cases were module API was checking the mustObeyClient it was targeting the server.current_client. Not sure this is a big concern but maybe we should allow somehow to also check the current running client? like  `int VM_MustObeyCurrentRunningClient()`?

Given said that. I do agree fake clients and tmp clients are usually used during background/blocking operations which might not be relevant for AOF/replication flows so maybe we can consider that in the future.

src/module.c
```diff
@@ -3719,6 +3719,14 @@ ValkeyModuleString *VM_GetClientUserNameById(ValkeyModuleCtx *ctx, uint64_t id)
     return str;
 }
 
+/* Returns 1 if commands are arriving from the primary client or AOF client
+ * and should never be rejected.
+ * Returns 0 otherwise (and also if ctx or if the client is NULL). */
```
madolson: Since this documentation is used to generate the public facing docs, I think it would also be useful to document *why* one would use this. Specifically it should be used for validation that might cause divergence on primaries and replicas (or AOF files).

src/module.c
madolson: Yeah, I looked through the code and didn't see any specific examples where we would need to find the real client. Replication from the primary should not block or rely on background behavior.

src/module.c
KarthikSubbarao: Yes, that makes sense. I updated the comment

tests/modules/propagate.c
```diff
@@ -390,6 +405,11 @@ int ValkeyModule_OnLoad(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int arg
                 "write",1,1,1) == VALKEYMODULE_ERR)
             return VALKEYMODULE_ERR;
 
+    if (ValkeyModule_CreateCommand(ctx,"propagate-test.obeyed",
+        propagateTestObeyed,
+        "",1,1,1) == VALKEYMODULE_ERR)
+    return VALKEYMODULE_ERR;
```
madolson: Why return error here?

KarthikSubbarao: Thanks - I missed this indentation.

Fixed it here: https://github.com/valkey-io/valkey/pull/1582/commits/93acad9f71e3ed04c656739ca96cfb7f852505cd

madolson (approved):

madolson:
@valkey-io/core-team Please vote. This is a pretty straight forward API to give modules the ability to quickly check if they should apply the command and ignore checks because it's coming from a primary or AOF. Worth mentioning that this is more performant than ValkeyModule_GetContextFlags, since that command also checks for memory bounds.

enjoy-binbin (approved):

PR #1584, opened 2025-01-18
Lower latenct-monitor-threashold in expire-cycle test case

### LINKED ISSUE #1580: [Test failure] LATENCY of expire events are correctly collected (unit/latency-monitor)
This test fails for me locally.

```
*** [err]: LATENCY of expire events are correctly collected in tests/unit/latency-monitor.tcl
Expected '{command 1737141346 674 674}' to match '*expire-cycle*' (context: type eval line 19 cmd {assert_match {*expire-cycle*} [r latency latest]} proc ::test) 
```

I'm running Linux on Macbook M2.

zuiderkwast:
The test case checks for expire-cycle in LATENCY LATEST, but with the new hash table, the expiry-cycle is too fast to be logged by latency monitor. Lower the latency monitor threshold to make it more likely to be logged.

Fixes #1580

### DIFF
--- a/tests/unit/latency-monitor.tcl
+++ b/tests/unit/latency-monitor.tcl
@@ -143,11 +143,7 @@ tags {"needs:debug"} {
 } ;# tag
 
     test {LATENCY of expire events are correctly collected} {
-<<<<<<< Updated upstream
-        r config set latency-monitor-threshold 5
-=======
-        r config set latency-monitor-threshold 2
->>>>>>> Stashed changes
+        r config set latency-monitor-threshold 1
         r config set lazyfree-lazy-expire no
         r flushdb
         if {$::valgrind} {set count 100000} else {set count 1000000}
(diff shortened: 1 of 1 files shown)


tests/unit/latency-monitor.tcl
```diff
@@ -143,7 +143,11 @@ tags {"needs:debug"} {
 } ;# tag
 
     test {LATENCY of expire events are correctly collected} {
-        r config set latency-monitor-threshold 20
+<<<<<<< Updated upstream
+        r config set latency-monitor-threshold 5
+=======
+        r config set latency-monitor-threshold 2
+>>>>>>> Stashed changes
```
ranshid: Why 2/5 and not '1'? I would be happy if we just had a way to report any action to latency monitor (zero will currently disable it).

zuiderkwast: Oops, I added a merge conflict... No reason. Let's make it 1.

```suggestion
        r config set latency-monitor-threshold 1
```

ranshid (approved):

PR #1590, opened 2025-01-20
Fix Protocol desync regression test

### LINKED ISSUE #1583: Test "Protocol desync regression test" in test_protocol.tcl not exactly checking the desync bug
this is the test:
```
if {$::tls} {
                set s [::tls::socket [srv 0 host] [srv 0 port]]
            } else {
                set s [socket [srv 0 host] [srv 0 port]]
            }
            puts -nonewline $s $seq
            set payload [string repeat A 1024]"\n"
            set test_start [clock seconds]
            set test_time_limit 30
            while 1 {
                if {[catch {
                    puts -nonewline $s payload
                    flush $s
                    incr payload_size [string length $payload]
                }]} {
                    set retval [gets $s]
                    close $s
                    break
                } else {
                    set elapsed [expr {[clock seconds]-$test_start}]
                    if {$elapsed > $test_time_limit} {
                        close $s
                        error "assertion:Valkey did not closed connection after protocol desync"
                    }
                }
            }
            set retval
```

the test write null "\x00" to the server and then write a lot of "payload" string.

then get the error "-ERR Protocol error: too big inline request\r\n"

it will also fail the same without sending any of the \x00 character.

we should choose from:
1. fix the test
2. fix the \x00 behaviour as excpected
3. remove the test

ranshid:
The desync regression test was created as a regression test for the following bug:
in case we embed NULL termination inside inline/multi-bulk message we will not be able to perform strchr in order to 
identify the newline(\n)/carriage-return(\r) in the client query buffer. this can influence (for example) replica reading primary stream and keep filling it's query buffer endlessly consuming more and more memory. 

In order to handle the above risk, a check was added to verify the inline bulk and multi-bulk size are not exceeding the 64K bytes in the query-buffer. A test was placed in order to verify this.

This PR introduce the following fixes to the desync regression test:
1. fix the sent payload to flush 1024 bytes block of 'A's  instead of 'payload' which was sent by mistake.
2. Make sure that the connection is correctly terminated on protocol error by the server after exceeding the 64K and not over 64K.
3. add another test intrinsic which will also verify the nested bulk with embedded null termination (was not verified before)

fixes https://github.com/valkey-io/valkey/issues/1583

NOTE: Although it is possible to change the use of strchr to a more "safe" utility (eg memchr) which will not pause scan at first occurrence of '\0', we still like to protect against over excessive usage of the query buffer and also preserve the current behavior(?). We will look into improving this though in a followup issue.

### DIFF
--- a/tests/unit/protocol.tcl
+++ b/tests/unit/protocol.tcl
@@ -69,7 +69,7 @@ start_server {tags {"protocol network"}} {
     }
 
     set c 0
-    foreach seq [list "\x00" "*\x00" "$\x00"] {
+    foreach seq [list "\x00" "*\x00" "$\x00" "*1\r\n$\x00"] {
         incr c
         test "Protocol desync regression test #$c" {
             if {$::tls} {
@@ -88,6 +88,8 @@ start_server {tags {"protocol network"}} {
                     puts -nonewline $s $payload
                     flush $s
                 }]} {
+                    set retval [gets $s]
+                    puts "got error response from client: $retval"
                     puts "exception after writing $payload_size bytes"
                     assert_morethan $payload_size $PROTO_INLINE_MAX_SIZE
                     break
--- a/tests/unit/protocol.tcl
+++ b/tests/unit/protocol.tcl
@@ -79,6 +79,8 @@ start_server {tags {"protocol network"}} {
             }
             fconfigure $s -blocking 0
             puts -nonewline $s $seq
+            # PROTO_INLINE_MAX_SIZE is hardcoded in Valkey code to 64K. doing the same here 
+            # since we would like to validate it is enforced. 
             set PROTO_INLINE_MAX_SIZE [expr 1024 * 64]
             set payload [string repeat A 1024]"\n"
             set payload_size 0
--- a/tests/unit/protocol.tcl
+++ b/tests/unit/protocol.tcl
@@ -88,7 +88,6 @@ start_server {tags {"protocol network"}} {
                     puts -nonewline $s $payload
                     flush $s
                 }]} {
-                    puts "exception after writing $payload_size bytes"
                     assert_morethan $payload_size $PROTO_INLINE_MAX_SIZE
                     break
                 }
--- a/tests/unit/protocol.tcl
+++ b/tests/unit/protocol.tcl
@@ -77,29 +77,30 @@ start_server {tags {"protocol network"}} {
             } else {
                 set s [socket [srv 0 host] [srv 0 port]]
             }
+            fconfigure $s -blocking 0
             puts -nonewline $s $seq
+            set PROTO_INLINE_MAX_SIZE [expr 1024 * 64]
             set payload [string repeat A 1024]"\n"
-            set test_start [clock seconds]
-            set test_time_limit 30
-            while 1 {
+            set payload_size 0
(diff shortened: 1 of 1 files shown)


tests/unit/protocol.tcl
```diff
@@ -77,29 +77,30 @@ start_server {tags {"protocol network"}} {
...
                 if {[catch {
-                    puts -nonewline $s payload
-                    flush $s
                     incr payload_size [string length $payload]
+                    puts -nonewline $s $payload
+                    flush $s
                 }]} {
-                    set retval [gets $s]
-                    close $s
+                    puts "exception after writing $payload_size bytes"
```
ranshid:
```suggestion
```

uriyage (approved):

tests/unit/protocol.tcl
```diff
@@ -69,37 +69,37 @@ start_server {tags {"protocol network"}} {
...
         incr c
         test "Protocol desync regression test #$c" {
             if {$::tls} {
                 set s [::tls::socket [srv 0 host] [srv 0 port]]
             } else {
                 set s [socket [srv 0 host] [srv 0 port]]
             }
+            fconfigure $s -blocking 0
             puts -nonewline $s $seq
+            set PROTO_INLINE_MAX_SIZE [expr 1024 * 64]
```
zuiderkwast, line 84: Is this a hard-coded check in the implementation? It looks like it, but please add a comment about it anyway. For me (the reader) it's not easy to remember if it's a hard-coded or configurable limit.

tests/unit/protocol.tcl
```diff
@@ -69,37 +69,37 @@ start_server {tags {"protocol network"}} {
     }
 
     set c 0
-    foreach seq [list "\x00" "*\x00" "$\x00"] {
+    foreach seq [list "\x00" "*\x00" "$\x00" "*1\r\n$\x00"] {
```
zuiderkwast, line 72: This adds some new thing to test? Not only fixing a failing test it seems.

tests/unit/protocol.tcl
```diff
@@ -69,37 +69,37 @@ start_server {tags {"protocol network"}} {
...
-            set test_time_limit 30
-            while 1 {
+            set payload_size 0
+            while {$payload_size <= $PROTO_INLINE_MAX_SIZE} {
                 if {[catch {
-                    puts -nonewline $s payload
-                    flush $s
                     incr payload_size [string length $payload]
+                    puts -nonewline $s $payload
+                    flush $s
```
zuiderkwast, line 91: Why are these lines reordered? Does it make a difference?

zuiderkwast (approved):
Looks good, with some nit. Main change from time limit to size limit makes sense.

Please just mention in the top comment this and the other changes.

tests/unit/protocol.tcl
ranshid: right - this was not covered by the early test

tests/unit/protocol.tcl
ranshid: only the $ for the payload. not sure why the entire context was recorded

ranshid:
> Looks good, with some nit. Main change from time limit to size limit makes sense.
> 
> Please just mention in the top comment this and the other changes.

done

Issue #23, opened 2024-03-25
[NEW] Atomic slot migration HLD 


PingXie:
See #1949 for the latest updates.

# Problem Statement

Efficient slot migration is critical for maintaining the scalability and resilience of Valkey clusters, but the current client-driven migration approach presents significant challenges.

Operators currently need to perform multiple manual steps to migrate slots, including setting migration states, transferring data, and updating ownership, all of which are prone to errors and inefficiencies.

While existing migration methods ensure availability during the process, they rely on mechanisms like redirection (-MOVED, -ASK) to maintain client access to data. These mechanisms introduce significant client-side complexity, requiring clients to handle redirections, retries, and reconfigurations during migration. Ensuring atomicity and consistency during migrations remains a challenge, especially for operations involving multiple keys or large datasets.

Compounding these challenges, limited observability leaves operators without adequate insight into the migration process, making it difficult to monitor progress or resolve issues efficiently.

# Command Definitions

## CLUSTER IMPORT SLOTS

*   **Syntax:** `CLUSTER IMPORT SLOTS <slot_start> <slot_end> [<slot_start> <slot_end>...]`
*   **Arguments:**
    *   `<slot_start>`: The starting slot number of a range (integer between 0 and 16383).
    *   `<slot_end>`: The ending slot number of a range (integer between 0 and 16383 and equal to or greater than `<slot_start>`), inclusive.
*   **Return Value:**
    *   `<op_id>` on queuing success, where `<op_id>` is a UUID for the long running import operation.
    *   `-ERR <error message>` on failure (e.g., invalid slot range).
*   **Description:** This command asynchronously imports specified slot ranges to the target primary node, allowing background import management. Existing target node slots and slots with undetermined sources are skipped. Note that although the import ID is ephemeral and lost if the target primary is terminated, atomic slot migration ensures the cluster remains consistent, with keys in a slot always residing on a single shard.

## CLUSTER IMPORT STATUS

*   **Syntax:** `CLUSTER IMPORT STATUS OPID <id>`
*   **Arguments:**
    *   `<op_id>`: The unique identifier of the import operation, as returned by `CLUSTER IMPORT SLOTS`.
    *   `-ERR <error message>` on failure (e.g., invalid operation ID).
*   **Return Value (RESP2 and RESP3):** A map with the following information:
    *   Total number of slots requested for import.
    *   Number of slots successfully imported.
    *   Number of slots that failed to import.
    *   Number of slots currently being imported.
[...]

Issue #1355, opened 2024-11-26
[NEW] Introduce First-Class Durability Support in Valkey


PingXie:
# Problem Statement #

Valkey offers several options for users seeking to ensure data durability, but each has notable limitations.

A commonly used approach is the `WAIT` command, which ensures that changes are replicated to a specified number of replicas before confirming success to the client. However, this places the burden of managing durability on the application, introducing complexity and increasing the risk of errors.

Another option is to use Append-Only File (AOF) with `ALWAYS` fsync, while avoiding replicas altogether. This ensures that writes are persisted to disk before being acknowledged to the client. However, AOF is not a true write-ahead log (WAL). For instance, in cases where writes or fsync operations fail, the changes might still be appended to the AOF file even though the client was informed of a failure. This creates a mismatch between the client’s perception and the actual system state, leading to what can be described as durability inconsistency. Moreover, this approach provides durability guarantees only on a single node.

Similarly, the `WAITAOF` command, which ensures data is written to the AOF file before acknowledgment, shares the same challenges of client-side complexity and durability inconsistency. While it provides durability under specific conditions, it does not guarantee correctness in all failure scenarios.

As a result of these limitations, Valkey currently lacks a mechanism to provide a robust durability contract. Specifically, there is no reliable way to guarantee that at least N copies of data are consistently maintained—where N is the user-configured replication factor—in the face of single-point failures such as a node outage or an availability zone (AZ) failure.

# Proposal # 

This proposal introduces first-class support for synchronous replication and deterministic elections in Valkey, enabling users to maintain at least N consistent copies of their data even in the event of single-point failures. Here, N refers to the user-configured replication factor. Additionally, this proposal ensures that only replicas fully synchronized with the primary are eligible to contend for primaryship, guaranteeing predictable and reliable behavior across all failure scenarios.

## Synchronous Replication ## 

When operating in "durability" mode—a new server configuration—the primary node acknowledges a write command only after successfully replicating the change to the expected number of replicas and ensuring that those changes are committed.

To prevent unbounded delays, the primary enforces a configurable timeout for replica acknowledgments. If the number of responding replicas falls below the durability threshold after the timeout expires, the write will fail with a "not enough replicas" error. This ensures a balance between durability guarantees and write latency.

## Deterministic Elections ##

A replica’s eligibility for election is explicitly controlled by the primary. All replicas start in an ineligible state, which prevents them from contending for the primary role.

[...]

PR #1591, opened 2025-01-20
Introduce atomic slot migration

murphyjacob4:
Starting this PR in draft so we can review early and gather alignment.

### Summary

The solution is mostly based on discussion and proposals in #23 :

- Target driven: the user sends `CLUSTER IMPORT` to the target and it syncs down the slot
- Uses AOF format for snapshot, proxying all commands directly from source primary through the target primary to all target replicas 
- Consensus-less epoch bump to broadcast the topology change
 
### Import and Export Workflows

We utilize a new internal `CLUSTER SYNCSLOTS` command with additional sub-commands to transition through the slot migration state machine.

On either end of the migration, we track the ongoing migrations via two job queues: slot_import_jobs and slot_export_jobs. Right now, we only support one concurrent job in both the slot import and slot export job queue. There is no design restriction here (outside of perhaps some protocol additions to `CLUSTER SYNCSLOTS` - but it helps with simplicity for the first iteration.

### Workflow overview

1. User sends `CLUSTER IMPORT SLOTSRANGE <slot start> <slot end> ...` to target node `T`
2. `T` initiates a new connection to source node `S`
3. If required, `T` runs `AUTH` based on replication configuration
4. `T` initiates `CLUSTER SYNCSLOTS START` to `S`
5. `S` begins tracking the client that sent the command as the slot export client. It spawns a child process at the next available time and runs AOF rewrite with just the specified slots. It then begins accumulating a backlog of writes in the slot export client output buffer, without installing the write handler.
6. At the end of the AOF rewrite, `S` also appends `CLUSTER SYNCSLOTS ENDSNAPSHOT`
7. `T` processes the AOF rewrite as it would any other client using `readQueryFromClient`. Once it gets the `CLUSTER SYNCSLOTS ENDSNAPSHOT`, `T` sends back a `CLUSTER SYNCSLOTS PAUSE` to pause `S`.
8. Upon getting the command, `S` unblocks the slot export client to `T` which has been accumulating ongoing writes. `S` then pauses itself, and sends `CLUSTER SYNCSLOTS PAUSEOFFSET <offset>` back to `T` with the current offset. (Note that the offset is not the primary replication offset, it is actually a computed offset based on how much we have been accumulating on `T`'s client.)
9. `T` waits for it's replication offset to catch up to the sent offset, and once caught up executes the consensus-less bump.
10. `S` finds out about the bump via cluster gossip, unpauses itself, and cleans up dirty keys.

If at any point a client is disconnected on either end, or a timeout is reached on the target node, the migration is marked as failed. If a migration fails, we delete all keys in the slots we were migrating.

### Filtering traffic

We filter the traffic to the target node through a filtered AOF rewrite and a filtered replication stream. The filtered AOF rewrite requires some refactoring of the snapshot code for reuse, but utilizes the same overall procedure (piping through the parent process to the target node connection).

The filtered replication stream hooks into the existing replication code and appends the commands directly to the client output buffer. We don't use replication backlog as there is no easy way to filter it once added to the backlog without re-processing it to query for the slot number of each command.

We add a new check in `putClientInPendingWriteQueue` to prevent the two command streams from merging. The parent process will just accumulate the replication stream in the client output buffer until we get the notification that the target is done with the snapshot.

### Other notes

- Since there are many checks that apply to both slot migration and replication - I added a new client flag "replicated" which simply means that a client is being replicated from another node. In many places, we now check the `replicated` flag instead of the `primary` flag.
- To support non-contiguous slot ranges, the APIs and code use a slot bitmap to share information about what slots are being migrated.
- We already delete unowned slots after loading RDB and AOF. Replicas that are promoted to primary now go through the same steps to delete all unowned keys. This prevents the keys from being leaked forever if the previous primary disappears.

### Remaining work items

* Continue cleaning up code, especially around the state machines
* TCL tests
* Metrics
* Improved logging
* Improved pause logic - where we only pause the slots being moved
* Hide the in-progress import better for commands like `KEYS` and `RANDOMKEY`
* Other management APIs (i.e. `CLUSTER IMPORT STATUS`, `CLUSTER IMPORT CANCEL`)
* Incorporate dual-channel-style replication

### DIFF
--- a/src/aof.c
+++ b/src/aof.c
@@ -32,6 +32,7 @@
 #include "rio.h"
 #include "functions.h"
 #include "module.h"
+#include "cluster.h"
 
 #include <signal.h>
 #include <fcntl.h>
@@ -2190,27 +2191,20 @@ werr:
     return 0;
 }
 
-int shouldFilterSlot(int slot, void * slot_ranges) {
-    if (slot_ranges == NULL) return 0;
-    list *ranges = (list *)slot_ranges;
-    listIter li;
-    listNode *ln;
-    listRewind(ranges, &li);
-    while ((ln = listNext(&li))) {
-        slotRange *range = (slotRange *) ln->value;
-        if (slot >= range->start && slot <= range->end) return 0;
-    }
-    return 1;
+int shouldFilterSlot(int slot, void * privdata) {
+    if (privdata == NULL) return 0;
+    unsigned char *slot_bitmap = (unsigned char *)privdata;
+    return !bitmapTestBit(slot_bitmap, slot);
 }
 
-int rewriteAppendOnlyFileRio(rio *aof, list *slot_ranges) {
+int rewriteAppendOnlyFileRio(rio *aof, unsigned char *slot_bitmap) {
     int j;
     long key_count = 0;
     long long updated_time = 0;
     kvstoreIterator *kvs_it = NULL;
 
     /* Record timestamp at the beginning of rewriting AOF. */
-    if (server.aof_timestamp_enabled && slot_ranges == NULL) {
+    if (server.aof_timestamp_enabled && isSlotBitmapAllSlots(slot_bitmap)) {
         sds ts = genAofTimestampAnnotationIfNeeded(1);
         if (rioWrite(aof, ts, sdslen(ts)) == 0) {
             sdsfree(ts);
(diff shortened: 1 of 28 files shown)


hwware:
Do you have plan to implement the following 2 cases ? 

1. CLUSTER IMPORT CANCEL
2.  Source Primary failover during slot migration

murphyjacob4:
> 1. CLUSTER IMPORT CANCEL

All this needs to do is call freeClient on the slot migration source client and delete all keys in the slot bitmap of that migration. The source is tracking the slot migration and when the client close notification comes in - it frees its local tracking information. This is the same process that already occurs if a migration times out.

> 2. Source Primary failover during slot migration

The implementation plan would be:

1. Where we currently check the timeout on the target side, also check that the source is still primary, and if not cancel the operation (as discussed above)
2. On the source side, we would introduce a similar check, if we find out we are demoted, cancel the operation and close the slot migration link. Right now the source will just give up if that happens, however if we add retry in the future, subsequent retries to do CLUSTER SYNCSLOTS will fail with an error due to not being primary, which we would detect and fail the operation.

Note that due to consensus-less epoch bump - if there is a race between failover and slot migration - both may succeed, and later one of those will win deterministically based on the epoch collision protocol - so we will lose some writes on the epoch collision losing side due to a period of two nodes declaring primaryship. This is the same issue that exists in the current consensus-less slot migration implementation - and we are looking to address it as part of https://github.com/valkey-io/valkey/issues/1355 instead

murphyjacob4:
Hi folks, so @enjoy-binbin, @PingXie, and I discussed offline. Tencent has offered a solution that they have developed internally. Moving forward, @enjoy-binbin and I will join efforts and work on bridging gaps in the Tencent solution to meet the requirements we outlined in #23. Hopefully we will have a shared PR for review soon. 

Given this, I am going to go ahead and close this pull request.

PR #1602, opened 2025-01-22
RDMA: Protect RDMA memory regions

pizhenwei:
Use Linux syscall mmap/munmap to manage a RDMA memory region, then we have a guard page protected VMA like (cat /proc/PID/maps):
 785018afe000-785018aff000 ---p 00000000 00:00 0  -> top guard page
 785018aff000-785018bff000 rw-p 00000000 00:00 0  -> RDMA memory region
 785018bff000-785018c00000 ---p 00000000 00:00 0  -> bottom guard page

Once any code accesses memory unexpectedly, segment fault occurs.

### DIFF
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -23,6 +23,7 @@
 #include <sys/types.h>
 #include <sys/socket.h>
 #include <netdb.h>
+#include <sys/mman.h>
 
 #define CONN_TYPE_RDMA "rdma"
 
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
LGTM. I'm don't fully understand the details but I think you know this area.

hwware (approved):

Issue #1108, opened 2024-10-02
[NEW] Rolling downgrade, forward compatibility


zuiderkwast:
**The problem/use-case that the feature addresses**

After a rolling upgrade of a cluster, if some problem is found, such as CPU or memory usage or some bug, the administrator wants to be able to a rolling downgrade again and take some time to investigate the problem.

When adding new nodes to a cluster, the standard procedure is to add new nodes as replicas to existing nodes. For replication to work, we require that a replica's RDB version >= primary's RDB version. This works for rolling upgrade, but it doesn't work for downgrading if the RDB version has changed.

Some system that I've been made aware of is stuck on Redis 6 (RDB 9) because of the requirement to be able to do rolling downgrades.

The cluster bus is already backward and forward compatible AFAIC. It would be good to avoid breaking that. :)

Note that we're only talking about adding new nodes of an old version to a cluster. We don't need to start an old node with a config file from a new version such as nodes.conf, nor do we want to reuse a node-id when replacing nodes in the cluster.

**Description of the feature**

Possibility to generate RDB in an older version than the latest, if no keys require the newer RDB version.

An example is the bump from RDB 10 to 11 (that I'm guilty of by introducing listpack encoding for sets). If we configure the node to generate RDB 10, the node can easily be made to store this listpack as a RDB_TYPE_SET (which is just each element as a string) instead of RDB_TYPE_SET_LISTPACK (a listpack dump). The bump from 9 to 10 is similar.

**Alternatives you've considered**

#59 = Full sync without RDB preamble, although it could be combined with a REPLCONF to check if the replica supports the current RDB version and fallback to AOF without preamble. (RDB is a bit faster and smaller in size.)

**Additional information**

Redis never supported downgrades. This would be a Valkey feature that would make Valkey superior in terms of compatibility.

| Valkey/Redis | RDB  | Why bump |
|--------------|------|-------|
| Redis 5.0    |  9   |       |
| Redis 6.0    |  9   |       |
| Redis 6.2    |  9   |       |
| Redis 7.0    | 10   | Change ziplist to listpack for hashes, sets, etc. |
| Valkey 7.2   | 11   | Add listpack encoding to sets |
| Valkey 8.0   | 11   |       |

madolson:
The major decisions:
1. Do we want to be able to support this downgrade.
2. How far back to we want to support restoring into older versions? (At least one version, maybe much further)
3. How will we document this.
[...]

Issue #59, opened 2024-03-28
Allow using AOF (the collection of commands) instead of RDB for full synchronization


soloestoy:
Whenever we introduce new data structures or new encodings, we need to adjust the RDB's version and encoding format. Internally in the server, this is not an issue. However, changes to RDB create a substantial adaption workload for external tools.

For example, some data migration tools like redis-shake, which needs to parse the RDB file during full synchronization, has to put effort into adapting to new versions whenever there are changes in the RDB (RDB version as well as many changes in storage structures). The tool needs to parse RDB to extract key-value pairs and transform them into restore commands before sending them to the target instance.

Also, there are many detailed issues to be addressed, such as the restore command's parameters not exceeding 500MB (due to `proto-max-bulk-len` limitation), meaning that when dealing with larger payloads, it's necessary to analyze the specific storage format for splitting, and then reverse-engineer it back into commands for replay (for example, a large hash would be converted into multiple HSET commands).

Furthermore, instances on old versions cannot parse RDB from new versions and cannot use the restore command for data migration (some special scenarios may require rolling back to a previous version by using data migration tools).

To solve the problems and to ensure that migration tools are not affected by changes in RDB format, we can add a new method for full synchronization: using the AOF file (where AOF specifically refers to a collection of commands, not using an RDB preamble). While full synchronization between master and replica still uses the RDB file, data migration tools could declare the file format they wish to use during full synchronization via the REPLCONF command. By doing so, data migration tools can simply forward commands without needing to parse RDB, allowing the full synchronization data to be directly passed through to the target instance, thus simplifying the adaptation work.

hpatro:
@soloestoy I've not looked much into AOF, so my question might be very naive.

Wouldn't a user need to pay certain amount of performance penalty (would vary with `appendfsync` config level) to keep AOF enabled and store the collection of commands whereas with the RDB methodology, libraries needs to stay on top of the updates in the RDB versions but there isn't performance penalty for regular command execution.

soloestoy:
@hpatro your question is very good. Perhaps I didn't express it clearly at the beginning, but I can explain it by answering your question.

The AOF file used during full synchronization does not rely on the `appendonly` config. It simply means that we originally used RDB files as snapshots, but now we have replaced it with a new AOF format snapshot. We also need to fork the child process, which converts all data into a set of write commands in AOF format. This way, the migration tool receives a set of commands during the full synchronization phase, eliminating the need to parse RDB encoding and achieving version independence.

[...]

PR #1604, opened 2025-01-22
Relaxed RDB version check

zuiderkwast:
New config `rdb-version-check` with values:

* `strict`: Reject future RDB versions.
* `relaxed`: Try parsing future RDB versions and fail only when an unknown RDB opcode or type is encountered.

This can make it possible for Valkey 8.1 to try read a dump from for example Valkey 9.0 or later on a best-effort basis. The conditions for when this is expected to work can be defined when the future Valkey versions are released. Loading is expected to fail in the following cases:

* If the data set contains any new key types or other data elements not supported by the current version.
* If the RDB contains new representations or encodings of existing key types or other data elements.

This change also prepares for the next RDB version bump. A range of RDB versions (12-79) is reserved, since it's expected to be used by foreign software RDB versions, so Valkey will not accept versions in this range even with the `relaxed` version check. The DUMP/RESTORE format has no magic string; only the RDB version number.

This change also prepares for the magic string to change from REDIS to VALKEY next time we bump the RDB version.

Related to #1108.


hpatro (commented):
> New config `rdb-version-check` with values:
> 
> * `strict`: Reject future RDB versions.
> * `relaxed`: Try parsing future RDB versions and fail only when an unknown RDB opcode or type is encountered.

Should we introduce three modes? last one being load all the keys which are feasible to be done on the server and instead of failing on error, just skip the key. 

For caching scenarios where the application can handle pulling missing key from the database, the third mode might be handy.

soloestoy (changes requested):
One of my concerns is that even if users don't utilize the new features, failures might still occur.

For example, when we upgraded RDB encoding from ziplist to listpack, from the user's perspective, they weren't using any new features and were simply performing regular operations like `LPUSH`, `SADD`, or `HSET`. However, the RDB encoding had changed, making a successful downgrade impossible.

In such cases, users might challenge us and demand a downgrade of the RDB version, which is something I'd very much like to avoid. If we could reach an agreement here—that is, we only allow a "best effort" approach to loading a higher version RDB and won't compromise on performing RDB version downgrades in the future—then this PR would be acceptable. It's important to establish this baseline, and the baseline will not be repeatedly tested or challenged in the future.

Moreover, I am currently working on full AOF synchronization development. I believe this approach is more promising than attempting to load a higher version RDB, so I think we can put this PR on hold.

valkey.conf
```diff
@@ -549,6 +549,15 @@ rdbcompression yes
 # tell the loading code to skip the check.
 rdbchecksum yes
 
+# Valkey can try to load an RDB dump produced by a future version of Valkey.
+# This can only work on a best-effort basis, because future RDB versions may
+# contain information that's not know to the current version. If no new features
```
```suggestion
# contain information that's not known to the current version. If no new features
```

tests/unit/dump.tcl
```diff
@@ -124,6 +124,25 @@ start_server {tags {"dump"}} {
         close_replication_stream $repl
     } {} {needs:repl}
 
+    test {RESTORE key with future RDB version, strict version check} {
+        #              str len    RDB 222 CRC64 checksum
+        #               |   |      |       |
+        set bar_dump "\x00\x03bar\xde\x00\x0fYUza\xd3\xec\xe0"
+        catch {r restore foo 0 $bar_dump replace} e
+        set e
+    } {ERR DUMP payload version or checksum are wrong}
```
madolson:
```suggestion
       assert_error {ERR DUMP payload version or checksum are wrong} {r restore foo 0 $bar_dump replace}
    }
```
Minor thing, it seems silly to have so many lines when we have a helper for this.

madolson (commented):
I think this change makes sense.

> Moreover, I am currently working on full AOF synchronization development. I believe this approach is more promising than attempting to load a higher version RDB, so I think we can put this PR on hold.

Your proposal doesn't solve the problem this change is addressing, since we can still end up with data that the lower version doesn't know how to store. I think we should make progress on this PR independent of your suggested change.

soloestoy:
>Your proposal doesn't solve the problem this change is addressing, since we can still end up with data that the lower version doesn't know how to store.

AOF full-resync is also not meant to completely solve this issue, but I believe it will be better than RDB. At this point, RDB encoding changes are unaffected.

>I think we should make progress on this PR independent of your suggested change.

Perhaps I didn't express clearly. AOF full-resync and this PR are not related, but I am very concerned about the future risks that this PR may bring. I hope that the things we do will stop at only "trying our best" to load higher version RDBs, and that we will not compromise by doing RDB version downgrades in the future. This is the premise that I care about.

madolson:
> Perhaps I didn't express clearly. AOF full-resync and this PR are not related, but I am very concerned about the future risks that this PR may bring. I hope that the things we do will stop at only "trying our best" to load higher version RDBs, and that we will not compromise by doing RDB version downgrades in the future. This is the premise that I care about.

Let's commit to it then! AWS is also interested in supporting forward loading of data from newer versions as long as it is compatible. I think the other managed providers would love this as well. I am fine introducing a much stronger guarantee as long as people are willing to implement the functionality. I have no concerns with saying we are "indefinitely forward" compatible as long as the data is compatible.

soloestoy:
I'm OK with that, @valkey-io/core-team let's vote on it: we support relaxing the RDB version check and will "try our best" to load recognizable data. For higher version data that cannot be recognized, we will not and will not in the future support on downgrading compatibility for RDB versions.

zuiderkwast:
@soloestoy Yes, best-effort and it can fail even if no new features are used. That's what the second bullet in the top comment explains: The loading is expected to fail "if the RDB contains new representations or encodings of existing key types or other data elements". This means that it can fail even if no new features are used.

> It's important to establish this baseline, and the baseline will not be repeatedly tested or challenged in the future.

I agree with you. However, whatever we decide now, I think we will be very careful in the future about changing encodings (listpacks, etc.) because we will not want to break the RDB forward-compatibility. We will probably discuss if changing the encoding is worth it or not.

> Moreover, I am currently working on full AOF synchronization development. 

This is very good news! I really hope we can have it in 8.1.

We already have REPLCONF VERSION, so the primary already knows the replica's version. If the primary can see that the replica can't read the RDB version, then I wish it can send an AOF full sync as a fallback. If we do that, then don't need to worry about doing RDB version bumps in the future. Let's discuss it in #59.

soloestoy (approved):

enjoy-binbin (approved):

hwware (approved):

hwware:
@soloestoy We are not sure if you will join the next meeting on Monday because of the Chinese lunar new year, please vote here if possible.  Thanks

madolson (approved):

PR #1606, opened 2025-01-23
Initialize one variable in struct to avoid risk

hwware:
In C, we had better initialize every variable in struct, this PR fixes one missed variable Initialization.


enjoy-binbin (commented):
let's move it near the failover_auth_rank

ranshid (approved):

hwware:
> let's move it near the failover_auth_rank

Addressed. Thanks

Issue #1319, opened 2024-11-19
[NEW] Multiple DB supports in cluster mode


madolson:
When moving from Standalone to Cluster, there are two API changes that end users need to consider: cross-slot commands and moving from multiple DBs to a single database. Although the cross-slot requirement is a requirement in order to make sure Valkey clusters scale, there is no similar requirement for DBs. The decision to only support one Database was an optimization, mainly to simplify the key to slot mapping.

This feature was not added in Redis since the old core team considered using multiple databases to be an anti-pattern compared to using prefixes. For example, instead of using database 1 and 0, you could have all keys prefixed with `0::` and `1::` and then build ACLs on top of that.

This use case works, but has some drawbacks. One common workload is loading in a fresh dataset into a secondary database and then performing a `SWAPDB` operation and then an async flush on the old data.

We will have some technical difficulty with implementing multiple databases now with the introduction of dict per slot, since we would have to duplicate all of the structure for each dictionary.

wuranxx:
Does this issue expect to support multiple databases in cluster mode?

I think this is a valuable feature. In production environments, many customers are accustomed to using the multi-DB feature. When migrating from standalone mode to a clustered setup, they expect the provider to offer this capability.

madolson:
> Does this issue expect to support multiple databases in cluster mode?

Yes, thanks for you commenting. I made this as a placeholder to follow up about, so haven't fully added the details yet.

hpatro:
> I think this is a valuable feature. In production environments, many customers are accustomed to using the multi-DB feature. When migrating from standalone mode to a clustered setup, they expect the provider to offer this capability.

@wuranxx Do they use it as a multi tenant setup? I think we should also consider supporting first party ACL support for multiple DBs.

murphyjacob4:
> the ability to flush or swap DBs

Although not currently implemented, we could use DBs as an abstraction to allow users to control and monitor individual workloads that are isolated from one another but collocated on the same Valkey cluster. I think it is a common use case to have the same cluster hosting data for many workloads/microservices. Maybe databases are an easier avenue to solving this use case than just prefixes?

- It should be possible for the engine to collect per-DB stats (e.g. number of keys, memory footprint, etc). A command like `DBINFO <db_num>` could make it easy for users to monitor their server when they have many applications/use cases.
[...]

PR #1609, opened 2025-01-24
On-demand database allocation instead of preallocation

### LINKED ISSUE #1597: [BUG] The engine isn't usable when running with a large number of databases.
Valkey supports up to `INT_MAX` databases. While most customers use the default (16 databases), if Valkey is started with a large number of databases, the engine's CPU usage increases significantly, rendering it almost unusable, even when the databases are empty and there is no incoming traffic.

For example, an engine which was started with 
```
valkey-server --databases 10000000
```
exhibits a latency of over 200 milliseconds:
```
✗ time bin/valkey-cli ping
PONG
bin/valkey-cli ping  0.00s user 0.00s system 1% cpu 0.225 total
```

Perf shows:

```
# Children      Self  Command        Shared Object     Symbol
# ........  ........  .............  ................  .......................................
#
    97.78%     0.00%  valkey-server  valkey-server     [.] _start
            |
            ---_start
               __libc_start_main_impl (inlined)
               __libc_start_call_main
               main
               aeMain
               aeProcessEvents
               |
                --97.77%--processTimeEvents
                          serverCron
                          |
                           --97.73%--databasesCron
                                     |
                                     |--95.31%--activeExpireCycle
                                     |          |
                                     |           --76.50%--kvstoreSize
                                     |
                                      --2.42%--kvstoreSize
```

We can see the the CPU is busy (%70 utilization):
```
bin/valkey-cli info all |egrep "uptime|used_cpu_"
uptime_in_seconds:410
uptime_in_days:0
used_cpu_sys:1.842625
used_cpu_user:290.709860
used_cpu_sys_children:0.000000
used_cpu_user_children:0.000000
used_cpu_sys_main_thread:1.842629
used_cpu_user_main_thread:290.709501
```

**Expected Behavior**
The engine should remain responsive with low CPU usage when handling empty databases without any load.

Proposed Solutions

    Lazy Initialization: Avoid allocating memory and data structures for databases that are never accessed. Initialize databases only when they are explicitly used.
    Limit Maximum Databases: If the current architecture makes efficient handling of large numbers of databases infeasible, reduce the maximum number of databases supported or clearly document this limitation.

xbasel:
Allocate database structures lazily to prevent excessive memory usage
when a large number of databases is configured but not actually used.

fixes https://github.com/valkey-io/valkey/issues/1597


src/aof.c
```diff
@@ -2210,7 +2210,8 @@ int rewriteAppendOnlyFileRio(rio *aof) {
 
     for (j = 0; j < server.dbnum; j++) {
         char selectcmd[] = "*2\r\n$6\r\nSELECT\r\n";
-        serverDb *db = server.db + j;
+        if (server.db[j] == NULL || kvstoreSize(server.db[j]->keys) == 0) continue;
+        serverDb *db = server.db[j];
         if (kvstoreSize(db->keys) == 0) continue;
```
ranshid: we can probably remove this check now right?

src/cluster_legacy.c
```diff
@@ -5704,7 +5704,7 @@ int verifyClusterConfigWithData(void) {
 
     /* Make sure we only have keys in DB0. */
     for (j = 1; j < server.dbnum; j++) {
-        if (kvstoreSize(server.db[j].keys)) return C_ERR;
+        if (server.db[j] && kvstoreSize(server.db[j]->keys)) return C_ERR;
```
ranshid: this condition will probably repeat many times. suggest we encapsulate in a function?

src/db.c
```diff
@@ -668,29 +669,36 @@ serverDb *initTempDb(void) {
     return tempDb;
 }
 
+void freeServerdb(serverDb *db) {
+    if (db == NULL) return;
+    kvstoreRelease(db->keys);
+    kvstoreRelease(db->expires);
+    zfree(db);
+}
+
```
ranshid: This seems strange and incomplete. if we are writing a dedicated function to free a DB we need to make sure ALL it's fields are being reclaimed (or assert if they are in use). What about the watched keys? blocked keys? etc...

ranshid (commented):
some immediate comments.

src/cluster_legacy.c
xbasel: introduced new function, `databaseEmpty`

src/db.c
xbasel: This has been removed.

xbasel:
Running valkey with 10 million databases:
```
valkey-server --databases 10000000
```
empty dbs:

Before:
```
# Memory
used_memory:6086209848
used_memory_human:5.67G
used_memory_rss:6063915008
used_memory_rss_human:5.65G
```

After:
```
# Memory
used_memory:84790224
used_memory_human:80.86M
used_memory_rss:7602176
used_memory_rss_human:7.25M
```

db 8 million:
```
# Keyspace
db0:keys=1,expires=0,avg_ttl=0
db8000000:keys=1,expires=0,avg_ttl=0
127.0.0.1:6379[8000000]>
```

src/server.c
```diff
@@ -2716,9 +2721,55 @@ void makeThreadKillable(void) {
...
-void initServer(void) {
-    int j;
+/* Return non-zero if the database is empty  */
+int databaseEmpty(int id) {
+    return id < 0 || id >= server.dbnum || !server.db[id] || kvstoreSize(server.db[id]->keys) == 0;
+}
+
+// /* Initialize temporary db on replica for use during diskless replication. */
+// serverDb *initTempDb(int id) {
+//     int slot_count_bits = 0;
```
zuiderkwast: Not finished?

src/server.c
```diff
@@ -2716,9 +2721,55 @@ void makeThreadKillable(void) {
...
+    db->id = id;
+    db->avg_ttl = 0;
+    return db;
+}
 
+void initDatabase(int id) {
+    if (server.db[id] == NULL) {
+        server.db[id] = createDatabase(id);
+    }
+}
```
zuiderkwast: In `kvstore.c` there is lazy creation of hashtables. It is similar to lazy creation of databases. In kvstore, a function called createHashtableIfNeeded is used in many places. It lazy-creates it and returns it.

I think we can use a similar pattern here. The name with "if needed" is also more clear than "init".

```suggestion
serverDb *createDatabaseIfNeeded(int id) {
    if (server.db[id] == NULL) {
        server.db[id] = createDatabase(id);
    }
    return server.db[id];
}
```

It can be used for example like this for the SELECT command:

```c
    c->db = createDatabaseIfNeeded(id);
```

src/server.c
```diff
@@ -2716,9 +2721,55 @@ void makeThreadKillable(void) {
     pthread_setcanceltype(PTHREAD_CANCEL_ASYNCHRONOUS, NULL);
 }
 
-void initServer(void) {
-    int j;
+/* Return non-zero if the database is empty  */
+int databaseEmpty(int id) {
+    return id < 0 || id >= server.dbnum || !server.db[id] || kvstoreSize(server.db[id]->keys) == 0;
```
zuiderkwast: Include "is" to make the function name more clear, like `databaseIsEmpty`. Without it, it may looks like the function empties the database. We have some other functions like `emptyData` that does that.

It's not forbidded to use `bool` anymore. We made a decision to allow it. Using `int` for booleans is also fine though.

Even if a database has no keys, it can have clients waiting for keys to be created (blocked on BLPOP), watched keys, etc., so the db structure it is not completely empty. Isn't what we need just to check if the database has been created or not, i.e. just check `server.db[id] != NULL`? In that case, we can call it `databaseIsCreated` or `databaseExists` instead.

zuiderkwast (commented):
Great, this looks pretty strait-forward.

The benchmark with 10M databases is great, but not a very common I think. :) The main motivation for this optimization IMO is in combination with #1319. For cluster mode with 16 empty databases, does it go from 10MB to 1MB?

xbasel:
> Great, this looks pretty strait-forward.
> 
> The benchmark with 10M databases is great, but not a very common I think. :) The main motivation for this optimization IMO is in combination with #1319. For cluster mode with 16 empty databases, does it go from 10MB to 1MB?

This has been created artificially (this change hasn't been merged into multi-database code,  i changed the code to create 16k slots without cluster mode)

before (preallocated) with (16 databases, 16k slots):
```
# Memory
used_memory:15887960
used_memory_human:15.15M
used_memory_rss:15204352
used_memory_rss_human:14.50M
used_memory_peak:15887960
used_memory_peak_human:15.15M
```
allocated-on-demand (16 databases, 16k slots), database 0 is always pre-allocated..
```
# Memory
used_memory:7327440
used_memory_human:6.99M
used_memory_rss:14548992
used_memory_rss_human:13.88M
```

src/server.c
xbasel: Should have been removed. `createDatabase` is not used, including for temp databases. I'll remove.

src/server.c
xbasel: The naming is bad here. This function is used to check if a  certain db has no keys. I've replaced with:
```
bool dbHasNoKeys(int dbid)
```
eg. https://github.com/valkey-io/valkey/pull/1609/files#diff-ade042c73ad8dc4e34f1a75b2a9e26a5342d1dee539a6188c2565feea0372a20L2214

xbasel:
@zuiderkwast /  @ranshid reminder :)

zuiderkwast:
It's not forgotten, but the focus now is on stabilizing the 8.1 release and fix the flaky tests before 8.1 GA. This one will have to wait until after 8.1. It will be good to release it together with the related cluster-multidb feature.

soloestoy:
I see many changes from `server.db->keys` to `server.db[0]->keys`, such as in countKeysInSlot. This will conflict with multi-databases. I think this PR should be aligned after the multi-databases merge.

src/server.c
```diff
@@ -6158,13 +6179,15 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
         if (sections++) info = sdscat(info, "\r\n");
         info = sdscatprintf(info, "# Keyspace\r\n");
         for (j = 0; j < server.dbnum; j++) {
+            serverDb *db = server.db[j];
+            if (db == NULL) continue;
```
xbasel, line 6183: This breaks the existing behavior. Now, databases that were never used or initialized will not be logged. We should either pretend the DB exists and print hardcoded zero values for NULL databases, or choose to log only non-empty databases, and document this in the release notes.

src/cluster_legacy.c
```diff
@@ -6540,20 +6540,20 @@ unsigned int delKeysInSlot(unsigned int hashslot) {
     for (int i = 0; i < server.dbnum; i++) {
         kvstoreHashtableIterator *kvs_di = NULL;
         void *next;
-        serverDb db = server.db[i];
-        kvs_di = kvstoreGetHashtableIterator(db.keys, hashslot, HASHTABLE_ITER_SAFE);
+        serverDb *db = server.db[0];
```
zuiderkwast: Shouldn't this be `i`?

And should we skip uninitialized databases?

```suggestion
        serverDb *db = server.db[i];
        if (db == NULL) continue;
```

src/db.c
```diff
@@ -650,47 +651,44 @@ long long emptyData(int dbnum, int flags, void(callback)(hashtable *)) {
...
-        tempDb[i].keys = kvstoreCreate(&kvstoreKeysHashtableType, slot_count_bits, flags);
-        tempDb[i].expires = kvstoreCreate(&kvstoreExpiresHashtableType, slot_count_bits, flags);
-    }
-
-    return tempDb;
-}
-
 /* Discard tempDb, it's always async. */
-void discardTempDb(serverDb *tempDb) {
+void discardTempDb(serverDb **tempDb) {
```
zuiderkwast, line 656: When we change the type here, we should update the comment to make it more clear.

Idea: Empty and release all databases in the temp db array and set them to NULL.

src/defrag.c
```diff
@@ -1227,6 +1227,7 @@ static void beginDefragCycle(void) {
     defrag.remaining_stages = listCreate();
 
     for (int dbid = 0; dbid < server.dbnum; dbid++) {
+        if (dbHasNoKeys(dbid)) continue;
```
zuiderkwast, line 1230: I don't think we should skip defrag of empty databases if they exist. The defrag still defrags the internal allocations of the hashtables structs if they exist.

Call chain: defragStageDbKeys -> defragStageKvstoreHelper -> kvstoreHashtableDefragTables -> hashtableDefragTables.

We can skip it if it's uninitialized though (or is it better to do the NULL check inside defragStageDbKeys and defragStageExpiresKvstore instead of here?)

```Suggestion
        if (server.db[dbid] == NULL) continue;
```

src/evict.c
```diff
@@ -568,7 +568,9 @@ int performEvictions(void) {
                  * so to start populate the eviction pool sampling keys from
                  * every DB. */
                 for (i = 0; i < server.dbnum; i++) {
-                    db = server.db + i;
+                    db = server.db[i];
+                    if (db == NULL) continue;
+                    ;
```
zuiderkwast: :question:

```suggestion
```

src/evict.c
```diff
@@ -634,7 +636,9 @@ int performEvictions(void) {
              * incrementally visit all DBs. */
             for (i = 0; i < server.dbnum; i++) {
                 j = (++next_db) % server.dbnum;
-                db = server.db + j;
+                db = server.db[j];
+                if (db == NULL) continue;
+                ;
```
zuiderkwast:
```suggestion
```

src/db.c
```diff
@@ -2838,9 +2841,9 @@ int bitfieldGetKeys(struct serverCommand *cmd, robj **argv, int argc, getKeysRes
     return 1;
 }
 
-bool dbHasNoKeys(void) {
+bool dbsHaveNoKeys(void) {
```
zuiderkwast: It's odd that this function is in db.c while dbHasNoKeys and createDatabase is in server.c.

I like to think of server.c as the unit handling the global server state and db.c handling the serverDb abstract datatype. This is not true though. The abstractions are mixed and the globals are accessed from all over the place.

Anyway, we can try to avoid making this mess even more messy by keeping these related functions together, or do you have another idea?

xbasel: I've moved it to server.c (dbHasNoKeys was originally in db.c). Now both are in server.c

src/cluster_legacy.c
xbasel: Yes, we should. Bad bug. Fixing.

src/db.c
xbasel: Modified.

src/db.c
zuiderkwast: OK. What's your reasoning to prefer server.c vs db.c for these?

I remember in the multi-db PR, someone suggested to move dbHasNoKeys to db.c. There was no clear explanation of why though, maybe only based on the function name...

Note: I don't really prefer either one of them, only that similar/related ones are together.

src/defrag.c
xbasel: Good catch. Adding the NULL check here is the right approach.

src/db.c
xbasel: IMO, logically `db.c` shouldn't know about server.c. Both `dbHasNoKeys` and `dbsHaveNoKeys` access a global state owned by `server.c`, so conceptually they may belong there. But in practice, it's intertwined... Either is fine.. 

This function for example
```
/* Return non-zero if the database is empty  */
int dbHasNoKeys(int dbid) {
    return dbid < 0 || dbid >= server.dbnum || !server.db[dbid] || kvstoreSize(server.db[dbid]->keys) == 0;
}
```
Knows about server.db, and dbnum, and dbid.. why `db.c` should know these?

zuiderkwast: Thanks. Now we have this explanation spelled out in case someone suggests move it again. :)

Just a though experiment: If we ever want to clean up these abstractions to the point where only server.c has access to the server struct and only db.c has access to the serverDb struct, then we may need to split these in two: `dbHasNoKeys(serverDb *db);` in db.c and `serverDbHasNoKeys(int dbid);` in server.c that calls the other one. That's not going to happen any day soon though.

src/db.c
```diff
@@ -650,47 +651,44 @@ long long emptyData(int dbnum, int flags, void(callback)(hashtable *)) {
...
-        tempDb[i].keys = kvstoreCreate(&kvstoreKeysHashtableType, slot_count_bits, flags);
-        tempDb[i].expires = kvstoreCreate(&kvstoreExpiresHashtableType, slot_count_bits, flags);
-    }
-
-    return tempDb;
-}
-
-/* Discard tempDb, it's always async. */
-void discardTempDb(serverDb *tempDb) {
+/* Empty and release all databases in the temp db array and set them to NULL. */
```
zuiderkwast: Hm... :thinking: I suggested this comment but missed the `zfree(tempDb)` in the bottom of the function. :smile: So we do actually free the whole thing.

I guess I was confused by the fact that we set `tempDb[i] = NULL;` if we're going to free the array anyway. It's unnecessary to set the elements to null here but I guess it's considered good practice...

```suggestion
/* Discard tempDb array. It's always async. */
```

src/replication.c
```diff
@@ -2701,7 +2701,7 @@ int sendCurrentOffsetToReplica(client *replica) {
     char buf[128];
     int buflen;
     buflen = snprintf(buf, sizeof(buf), "$ENDOFF:%lld %s %d %llu\r\n", server.primary_repl_offset, server.replid,
-                      server.db->id, (long long unsigned int)replica->id);
+                      server.db[0]->id, (long long unsigned int)replica->id);
```
zuiderkwast, line 2704: The id is the same as the index, so it's always 0 here?

Why is this code using the id of the first database only? Unknown historical reason? I guess someone thought `server.db` is the "current" db, but actually it's the db array.

zuiderkwast (approved):
Yeah this looks good to merge.

src/replication.c
xbasel: Looks the original code was always accessing db 0.  Used by dual channel. 
@naglera  Can you take a look?

naglera: This is a mistake. The intended behavior was to send the database ID of the last replicated item (i.e. `server.replicas_eldb`). 
It hasn't caused any issues so far, as primary is resetting it in `replicationSetupReplicaForFullResync`: https://github.com/valkey-io/valkey/blob/838ba44cd65b15822f4136a4aed33cc13caa4cae/src/replication.c#L820 for each new replica. 
I will address this in a separate PR.

src/server.h
```diff
@@ -1579,7 +1579,7 @@ struct valkeyServer {
     int hz;                   /* serverCron() calls frequency in hertz */
     int clients_hz;           /* clientsTimeProc() frequency in hertz */
     int in_fork_child;        /* indication that this is a fork child */
-    serverDb *db;
+    serverDb **db;
```
zuiderkwast:
```suggestion
    serverDb **db;            /* each db created when it's first used */
```

PR #1612, opened 2025-01-24
Make sure to disable pause after fork for dual channel test

### LINKED ISSUE #1484: [Test Failure] Sometimes failing valgrind test
We sometimes get a valgrind failure with no error, just printing out valgrind text:

Example: https://github.com/valkey-io/valkey/actions/runs/12474649298/job/34816982513
```

*** [err]: Valgrind error: ==90396== Memcheck, a memory error detector
==90396== Copyright (C) 2002-2022, and GNU GPL'd, by Julian Seward et al.
==90396== Using Valgrind-3.22.0 and LibVEX; rerun with -h for copyright info
==90396== Command: src/valkey-server ./tests/tmp/valkey.conf.3985.1683
==90396== 
```

There might be more of an error that is getting truncated.  Not sure what is causing it.

madolson:
Might close https://github.com/valkey-io/valkey/issues/1484.

I noticed that we don't disable pause after fork on the last test that was getting executed, so it might getting stuck in pause loops after the test ends if it tries another psync for any reason.

### DIFF
--- a/tests/integration/dual-channel-replication.tcl
+++ b/tests/integration/dual-channel-replication.tcl
@@ -850,7 +850,6 @@ start_server {tags {"dual-channel-replication external:skip"}} {
         $replica config set repl-timeout 60
         $primary config set repl-backlog-size 1mb
         
-        $replica debug pause-after-fork 1
         $primary debug populate 1000 primary 100000
         # Set primary with a slow rdb generation, so that we can easily intercept loading
         # 10ms per key, with 1000 keys is 10 seconds
@@ -858,6 +857,7 @@ start_server {tags {"dual-channel-replication external:skip"}} {
 
         test "Test dual-channel-replication primary gets cob overrun during replica rdb load" {
             set cur_client_closed_count [s -1 client_output_buffer_limit_disconnections]
+            $replica debug pause-after-fork 1
             $replica replicaof $primary_host $primary_port
             wait_for_condition 500 1000 {
                 [s -1 client_output_buffer_limit_disconnections] > $cur_client_closed_count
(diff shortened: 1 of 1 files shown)


madolson:
@zuiderkwast FYI, run extra tests doesn't run valgrind since they take like 2 hours.

tests/integration/dual-channel-replication.tcl
```diff
@@ -871,6 +871,7 @@ start_server {tags {"dual-channel-replication external:skip"}} {
                 fail "Primary did not free repl buf block after sync failure"
             }
             wait_and_resume_process 0
+            $replica debug pause-after-fork 0
```
ranshid, line 878: Just to understand: this needs to be disabled before reading the log msgs?

madolson: It can go anywhere after the wait_and_continue, but before the end of the test.

zuiderkwast: If the replica is getting stuck in `sync` loops, we should disable it ASAP, right? Not after the three `stop_write_load` in the end of the test case.

We can disable `pause-after-fork` here, but then I think we should also move the enabling of `pause-after-fork` inside the test case (the `test { ... }` block), rather than before the start of the test case. Symmetry like that makes it easier to follow the logic, and also to skip test cases using `--only` and similar options.

tests/integration/dual-channel-replication.tcl
```diff
@@ -871,6 +871,7 @@ start_server {tags {"dual-channel-replication external:skip"}} {
                 fail "Primary did not free repl buf block after sync failure"
             }
```
naglera, line 872: Replica may still get paused if more than 5 seconds elapse between `wait_and_resume_process 0` and `$replica debug pause-after-fork 0`. While not ideal, we might consider this approach:
```
            $primary config set repl-diskless-sync-delay 100
            wait_and_resume_process 0
            $replica debug pause-after-fork 0
```

madolson: That seems reasonable to me.

tests/integration/dual-channel-replication.tcl
madolson: > If the replica is getting stuck in sync loops, we should disable it ASAP, right? Not after the three stop_write_load in the end of the test case.

The stoping of the load is actually outside of the test, I agree we should disable it before that part of the test. 

Addressed your other comment as well.

ranshid (approved):

naglera (approved):

zuiderkwast (approved):

PR #1613, opened 2025-01-24
Embed keys and hash fields as SDS type 5

### LINKED ISSUE #1567: Extra memory consumed by embedded sds in new HashObject fields
Following https://github.com/valkey-io/valkey/pull/1502 which introduced hashtable instead of dict in hash objects we  started embedding the field sds in the hashTableEntry.
The problem is that the field sds might arrive from different sources:
1. command arguments - in the usual case the sds will be provided from the parsed command arguments.. In such cases the sds is from a stringObject which means it will always have a minimal header size of 3 bytes (sds8).
2. listpack conversion - when we convert from listpack to hashtable the listpack is scanned and the field sds is being created from the listpack string via sdsnewsize which will use the minimal header size (ie small strings will use sds5 which is 1 byte).
3. Modules - for example VM_HashSet will create a RAW string object which will have the sds allocated with a minimal size header (ie sds5 for small strings which is 1 byte long).

When we create the hashTable entry in hashTypeCreateEntry we will embed the field sds according to the provided sds representation, so in case the field originated at a parsed command argument it will use extra 2 bytes for the header. 

While there would probably NOT be any degradation in overall memory utilization (since the new hashtable is more memory efficient) it might cause strange results following listpack conversions.
for example:
say hash1 is created when the hash_max_listpack_entries config is 0 and added with 10 small fields
and hash2 is created when  the  hash_max_listpack_entries config is 0 9 and added with 10 small fields

after all 10 elements were added both tables are expected to show the same memory consumption, but hash1 would show as using extra 18 bytes of memory.

NOTE - I do think the issue is minor and would probably be addressed during the work on https://github.com/valkey-io/valkey/issues/1551 and/or https://github.com/valkey-io/valkey/issues/640 So I mainly opened it in order to have a better tracking of the issue.

zuiderkwast:
Until now, these embedded strings were SDS type 8, because they were copied as-is from EMBSTR encoded string objects, which always use SDS type 8. This change allows them to be embedded as SDS type 5, saving two bytes.

The implementation of embeddeding SDS strings in other structures is refactored. `sdswrite()` is a new function to create an SDS representation into a buffer provided by the caller. `sdscopytobuffer()`, which just copied the layout of an existing sds, is deleted.

The DEBUG SDSLEN command output is changed slightly, to account for correct allocation sizes of embedded strings and not report the same memory twice.

Fixes #1567


src/object.c
```diff
@@ -54,14 +54,15 @@ robj *createObjectWithKeyAndExpire(int type, void *ptr, const sds key, long long
     /* Calculate sizes */
     int has_expire = (expire != -1 ||
                       (key != NULL && sdslen(key) >= KEY_SIZE_TO_INCLUDE_EXPIRE_THRESHOLD));
-    size_t key_sds_size = 0;
+    size_t key_sds_len = key != NULL ? sdslen(key) : 0;
+    char key_sds_type = key != NULL ? sdsReqType(key_sds_len) : 0;
+    size_t key_sds_size = key != NULL ? sdsReqSize(key_sds_len, key_sds_type) : 0;
```
ranshid: small nitpick: personally I find the implementation more clear if we add
in has_embdkey = key != NULL

at start and replace all other checks for key != NULL etc...

src/object.c
```diff
@@ -187,26 +188,13 @@ static robj *createEmbeddedStringObjectWithKeyAndExpire(const char *val_ptr,
...
-    if (val_ptr == SDS_NOINIT) {
-        sh->buf[val_len] = '\0';
-    } else if (val_ptr != NULL) {
-        memcpy(sh->buf, val_ptr, val_len);
-        sh->buf[val_len] = '\0';
-    } else {
-        memset(sh->buf, 0, val_len + 1);
-    }
-    o->ptr = sh->buf;
+    o->ptr = sdswrite(data, val_sds_size, SDS_TYPE_8, val_ptr, val_len);
```
ranshid, line 199: I wonder if we could make this part optional. ie in case the val_ptr is NULL we do not embed the value.
This way we could reuse this function and call it form createObjectWithKeyAndExpire and save this code duplication?

src/debug.c
```diff
@@ -684,12 +684,16 @@ void debugCommand(client *c) {
...
         } else {
+            /* Report the complete robj allocation size as the key's allocation
+             * size. Report 0 as allocation size for embedded values. */
+            size_t obj_alloc = zmalloc_usable_size(val);
+            size_t val_alloc = val->encoding == OBJ_ENCODING_RAW ? sdsAllocSize(val->ptr) : 0;
             addReplyStatusFormat(c,
-                                 "key_sds_len:%lld, key_sds_avail:%lld, key_zmalloc: %lld, "
-                                 "val_sds_len:%lld, val_sds_avail:%lld, val_zmalloc: %lld",
-                                 (long long)sdslen(key), (long long)sdsavail(key), (long long)sdsAllocSize(key),
+                                 "key_sds_len:%lld, key_sds_avail:%lld, key_alloc:%lld, "
```
ranshid: since we are already changing it. maybe we can just make this in order and explicitly output: obj_alloc, instead of communicating key alloc and value alloc which does not really makes sense when both of them are embedded ?

tests/unit/type/string.tcl
```diff
@@ -755,4 +755,31 @@ if {[string match {*jemalloc*} [s mem_allocator]]} {
         assert_encoding "int" bar
         lappend res [r get bar]
     } {12 12}
+
+    test {Memory usage of embedded string value} {
+        # Check that we can fit 9 bytes of key + value into a 32 bit allocation,
```
ranshid:
```suggestion
        # Check that we can fit 9 bytes of key + value into a 32 byte allocation,
```

src/object.c
ranshid: also remind me: why is it we need force the value to be SDS_TYPE_8?

ranshid (commented):
(Sorry it took me time to get to it, was very sick for a long period of time) 

Overall LGTM!

Some small comments and/or questions.

src/object.c
zuiderkwast: If we want to get rid of the `ptr` field (#1658) we need to know the SDS type or header size, so we need to use a fixed SDS type or a hdr-size stored or a bit flag in the object to tell if it's sds5 or sds8.

I think we can leave it as SDS 8 for now. We can change it in the future if it's beneficial.

src/debug.c
zuiderkwast: Yeah and we can delete the commas so it looks more like some other output such as DEBUG OBJECT.

ranshid (approved):

Issue #1611, opened 2025-01-24
[Test Failure] Socket timeout for AOF test


madolson:
Error
```
sock56048fcc87c0 => (SPAWNED SERVER) pid:38170 - tests/integration/aof.tcl
```

We are seeing a fairly consistent failure like https://github.com/valkey-io/valkey/actions/runs/12919448042/job/36029989392#step:8:6146. 

In this case, the client connecting the AOF server is hanging after starting the process but before executing the test. We can see that `Turning appendonly on and off within a transaction` finished, so it must be failing on `Test cluster slots / cluster shards in aof won't crash` which is the test that comes after. I really don't have a great read on this, but there seems to be a few occurrences . Another example is https://github.com/valkey-io/valkey/actions/runs/12940498740/job/36094850711#step:8:6290. 

@enjoy-binbin I see you added the test, https://github.com/valkey-io/valkey/commit/80fcbd3fece6decb6195dc1a4289ebb675b5d45d. Maybe you have some thoughts on it.

enjoy-binbin:
Not sure if it has anything to do with it `Starting server on 127.0.0.1:23206 Port 23206 was already busy, trying another port...`, i don't have other clues right now.

roshkhatri:
most recent: https://github.com/valkey-io/valkey/actions/runs/14805407197/job/41572746840#step:8:6290

roshkhatri:
This seams like a containerized test issue.
I was seeing the same errors while running all the test on local.
Ran the test in a loop like this
```
./runtest --list-tests | while read -r test; do\
 echo "Running test: $test"; \
./runtest --single "$test" --verbose --accurate --dump-logs; \
done && \
echo "All tests completed!"
```

~~Which runs successfully.~~
Which also fails .

We would want to skip these tests 

I also tried running the `integration/aof.tcl` separately. It does not crash/fail/timeout.

roshkhatri:
looks like the problem are getting caused in `unit/introspection`
```
[root@cc0134db2d25 valkey]# ./runtest --single unit/introspection --verbose --accurate --dump-logs
Cleanup: may take some time... OK
Starting test server at port 21079
[ready]: 36925
Testing unit/introspection
[ready]: 36926
[ready]: 36924
[ready]: 36927
[ready]: 36928
[ready]: 36930
[ready]: 36929
[ready]: 36931
[ready]: 36932
[ready]: 36934
[ready]: 36933
[ready]: 36935
[ready]: 36937
[...]

PR #1614, opened 2025-01-25
Do command existence and arity checks when loading AOF to avoid crash

enjoy-binbin:
Do command existence and arity checks when loading AOF to avoid crash

Currently, loading commands such as `cluster` or `cluster slots xxx`
from AOF will cause the server to crash.
1. `cluster` is a container command, and executing proc will cause a
    crash because we do not check subcommand and arity.
2. `cluster slots xxx`, arity check fail, reply with an error from the
    AOF client and trigger a panic.

Of course, there are many other ways for a problematic AOF to cause the
panic, but it is still necessary do some basic checks before executing.
In this way, in these basic cases, we can print useful error messages
instead of crashing directly.

### DIFF
--- a/src/aof.c
+++ b/src/aof.c
@@ -1532,10 +1532,11 @@ int loadSingleAppendOnlyFile(char *filename) {
         }
 
         /* Command lookup */
+        sds err = NULL;
         cmd = lookupCommand(argv, argc);
-        if (!cmd) {
-            serverLog(LL_WARNING, "Unknown command '%s' reading the append only file %s", (char *)argv[0]->ptr,
-                      filename);
+        if ((!cmd && !commandCheckExistence(NULL, argv, argc, &err)) || (cmd && !commandCheckArity(cmd, argc, &err))) {
+            serverLog(LL_WARNING, "Error reading the append only file %s, error: %s", filename, err);
+            sdsfree(err);
             freeClientArgv(fakeClient);
             ret = AOF_FAILED;
             goto cleanup;
--- a/src/aof.c
+++ b/src/aof.c
@@ -1533,8 +1533,8 @@ int loadSingleAppendOnlyFile(char *filename) {
 
         /* Command lookup */
         sds err = NULL;
-        cmd = lookupCommand(argv, argc);
-        if ((!cmd && !commandCheckExistence(NULL, argv, argc, &err)) || (cmd && !commandCheckArity(cmd, argc, &err))) {
+        fakeClient->cmd = fakeClient->lastcmd = cmd = lookupCommand(argv, argc);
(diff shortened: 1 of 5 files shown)


enjoy-binbin:
Found this when looking #1611

src/aof.c
```diff
@@ -1532,10 +1532,11 @@ int loadSingleAppendOnlyFile(char *filename) {
         }
 
         /* Command lookup */
+        sds err = NULL;
         cmd = lookupCommand(argv, argc);
-        if (!cmd) {
-            serverLog(LL_WARNING, "Unknown command '%s' reading the append only file %s", (char *)argv[0]->ptr,
-                      filename);
+        if ((!cmd && !commandCheckExistence(NULL, argv, argc, &err)) || (cmd && !commandCheckArity(cmd, argc, &err))) {
```
zuiderkwast: Why not pass `fakeClient` to this function? It already has `fakeClient->argv` and `fakeClient->argc` so we don't need to add the new parameters.

```suggestion
        if ((!cmd && !commandCheckExistence(fakeClient, &err)) || (cmd && !commandCheckArity(cmd, argc, &err))) {
```

zuiderkwast (commented):
How will there be commands like `CLUSTER SLOTS XXXX` in the AOF file? Is it to be more resilient to future changes, like we already replicate CLUSTER SETSLOT now and maybe we will replicate CLUSTER FAILOVER in #1091?

I'm thinking for a container command like `CLUSTER` without subcommand, and for syntax error in `CLUSTER SLOTS`, it's safe to just ignore the command. It is safe because it's not a write command and is has no admin side effects. WDYT?

enjoy-binbin:
> How will there be commands like CLUSTER SLOTS XXXX in the AOF file? Is it to be more resilient to future changes, like we already replicate CLUSTER SETSLOT now and maybe we will replicate CLUSTER FAILOVER in https://github.com/valkey-io/valkey/pull/1091?

CLUSTER SLOTS will not be in the AOF, this is something like a damaged AOF. Using other commands may also cause the crashes, not just the cluster command, this is an example.

> I'm thinking for a container command like CLUSTER without subcommand, and for syntax error in CLUSTER SLOTS, it's safe to just ignore the command. It is safe because it's not a write command and is has no admin side effects. WDYT?

we shouldn't ignore it in silence, i think we should report this error.

src/aof.c
enjoy-binbin: I also thought about using it, but it requires assigning fakeClient->cmd, Ii haven't thought it through, i'll take a closer look to see if it has any impact.

enjoy-binbin: thanks, this look safe, i forgot freeClientArgv will nullify the cmd.

zuiderkwast (approved):
Looks good.

PR #1616, opened 2025-01-26
Enhancements and Refactor for Maintainability, and Performance

stav-bentov:
**Feature Addition:**
- Added IP address validation in the node address update logic to ensure robust input handling and prevent invalid configurations.
Commit: a7178a383

**Refactoring:**
- Cluster Size Computation: Extracted cluster size computation into a helper function for improved readability and reusability.
Commit: a3dd42bf4
- File Event Processing: Encapsulated loop logic into a dedicated function processSingleFileEvent() to isolate responsibilities and simplify the main code flow.
Commit: 215dcd602
- Key Expiration Check: Optimized the key expiration check logic with branch prediction hints, enhancing performance during runtime.
Commit: 733ef16c6

### DIFF
--- a/src/ae.c
+++ b/src/ae.c
@@ -393,6 +393,57 @@ int aePoll(aeEventLoop *eventLoop, struct timeval *tvp) {
     return ret;
 }
 
+static void processSingleFileEvent(
+        aeEventLoop *eventLoop, aeFiredEvent *firedEvent) {
+    int mask = firedEvent->mask;
+    int fd = firedEvent->fd;
+    aeFileEvent *fe = &eventLoop->events[fd];
+    int fired = 0; /* Number of events fired for current fd. */
+
+    /* Normally we execute the readable event first, and the writable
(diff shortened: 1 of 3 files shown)


enjoy-binbin (commented):
we don't do this refactor thing for this small reason, it breaks the git blame log

zuiderkwast (commented):
This PR contains three changes unrelated to each other:

1. Config validation, makes CONFIG SET return an error in some case? Where are the tests?
2. Branch prediction help. Some benchmark numbers or flame graph to motivate this change would be appreciated.
3. Move code. Reduced nesting is arguably a good idea, but some might say it is moved out of its context. It does not objectively make the code better. IMO the benefit does not outweigh the cost, considering extra effort by reviewers (not easy to see if the code is changed or just moved, it takes time) and the risk of merge conflicts on PRs and downstream forks.

1 and 2 can maybe be merged, but they're unrelated to each other so the should be in separate PRs. For 3 I agree with @enjoy-binbin.

Issue #1619, opened 2025-01-27
[NEW] Approximate percentage execution time per command


bluayer:
**The problem/use-case that the feature addresses**

In Valkey, users need to avoid a lot of time-consuming operations. However, users cannot know easily which operations or commands are consuming the time out of the total execution time by only looking at the execution time. Especially, if you're using various types of commands, it's difficult to check the influence of the command.

**Description of the feature**

The approximate percentage of execution time clearly shows what percentage of the engine that command occupies.
Also, based on this intent, it do not require floating operation for accuracy. If accurate calculation is required for some reason, it is recommended that the user perform the calculation.

PR #1620, opened 2025-01-27
Add approximate usec percentage per command

bluayer:
## Background
https://github.com/valkey-io/valkey/issues/1619

## Changes
- Add a function for calculating total command usec time
- With total usec, caculate approximate usec percentage
  - Don't need to use floating point-level accuracy because it's just hint for identifying occupied commands.

## Sample screenshot

Screenshot 2025-01-27 at 12 27 24 PM

### Concern
A recursive function is used when calculating the total command time, but there may be a better way. Please give me feedback.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -5430,7 +5430,24 @@ const char *getSafeInfoString(const char *s, size_t len, char **tmp) {
     return memmapchars(new, len, unsafe_info_chars, unsafe_info_chars_substs, sizeof(unsafe_info_chars) - 1);
 }
 
-sds genValkeyInfoStringCommandStats(sds info, hashtable *commands) {
+long long getTotalCommandUsec(hashtable *commands) {
+    hashtableIterator iter;
+    void *next;
+    long long total_usec = 0;
+    hashtableInitIterator(&iter, commands, HASHTABLE_ITER_SAFE);
+    while (hashtableNext(&iter, &next)) {
+        struct serverCommand *cmd = next;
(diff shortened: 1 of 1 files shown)


madolson (commented):
Thanks for the contribution, but I don't think we want to accept the feature idea. This data can be pretty easily derived on the client side, and all it does is take up server side CPU computation. (and a division at that, which can be expensive) It will also add more data to the output. So unless someone else thinks it's a good idea I think we can close this.

bluayer:
I understand what direction Valkey is.
I've currently two ideas while creating this PR, and I'd like to hear your opinions.

1/ What if we calculate usec_per_call as decimal(not float)? In most use cases, more precise time than microseconds is not required.

2/ There was an idea that it would be good if stat and info also had a level, like the log level. For example, at the production level, dashboards are usually configured separately, so there is no need to provide statistics in a way that is easy for people to see.

zuiderkwast:
We already have complaints that INFO is a performance problem and we want to avoid making it slower and increase the size of the reply. Computing derived information can be done on the client side, or using EVAL if you want.

madolson:
> 2/ There was an idea that it would be good if stat and info also had a level, like the log level. For example, at the production level, dashboards are usually configured separately, so there is no need to provide statistics in a way that is easy for people to see.

This could be maybe done. We could add more debug sections that aren't shown by default but could be opt'd in to.

bluayer:
I close this PR based on comments.

PR #1624, opened 2025-01-27
IP Address Validation and Key Expiration Optimization #1622 #1623

stav-bentov:
**IP Address Validation in Node Address Update:**
- Ensures that the node address is updated only when a valid IP address is obtained.-
- This prevents erroneous updates with invalid or placeholder values (e.g., an empty string or "?").

**Optimize Key Expiration Check with Branch Prediction Hint:**
- Introduces a condition to leverage branch prediction for improved performance.
- The added check reduces unnecessary evaluations when the database size (dbSize(db)) is 0.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -44,7 +44,7 @@ jobs:
         (github.event_name == 'schedule' && github.repository == 'valkey-io/valkey') ||
         (github.event_name == 'pull_request' && (contains(github.event.pull_request.labels.*.name, 'run-extra-tests') || github.event.pull_request.base.ref != 'unstable'))) &&
       !contains(github.event.inputs.skipjobs, 'ubuntu')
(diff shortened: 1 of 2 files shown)


src/cluster_legacy.c
```diff
@@ -2456,6 +2456,10 @@ int nodeUpdateAddressIfNeeded(clusterNode *node, clusterLink *link, clusterMsg *
     if (node->tcp_port == tcp_port && node->cport == cport && node->tls_port == tls_port && strcmp(ip, node->ip) == 0)
         return 0;
 
+    /* We should not update the node address if we were not able to get a valid
+     * IP address. */
+    if (ip[0] == '\0' || strcmp(ip, "?") == 0) return 0;
```
enjoy-binbin: What situation would cause this? Have you encountered it in a real life?

PR #1628, opened 2025-01-27
deflake tracking-redir-broken test

ranshid:
This address 2 issues:

1. It is possible (somehow) that the inner server client (r) was not working resp 3 when entering this test.
this makes sure it does.

2. in case the test failed it might leave the redirection client closed. there is a cross test assumption it should be open, so moved most of the assert checks to the end of the test.

example fail: https://github.com/valkey-io/valkey/actions/runs/12979601179/job/36195523412

### DIFF
--- a/tests/unit/tracking.tcl
+++ b/tests/unit/tracking.tcl
@@ -236,6 +236,8 @@ start_server {tags {"tracking network logreqres:skip"}} {
     }
 
     test {RESP3 Client gets tracking-redir-broken push message after cached key changed when rediretion client is terminated} {
+        # make sure r is working resp 3
+        r HELLO 3
         r CLIENT TRACKING on REDIRECT $redir_id
         $rd_sg SET key1 1
         r GET key1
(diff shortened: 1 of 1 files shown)


tests/unit/tracking.tcl
```diff
@@ -247,16 +249,17 @@ start_server {tags {"tracking network logreqres:skip"}} {
         for {set i 0} {$i <= $MAX_TRIES && $res < 0} {incr i} {
             set res [lsearch -exact [r PING] "tracking-redir-broken"]
         }
-        assert {$res >= 0}
-        # Consume PING reply
-        assert_equal PONG [r read]
```
enjoy-binbin, line 252: why this fail? Because the r is not in resp3?

ranshid: I suspect that it did somehow. maybe the 'r' is disconnected and reconnected at some point

enjoy-binbin (approved):

tests/unit/tracking.tcl
enjoy-binbin: let's give it a try.

PR #1630, opened 2025-01-27
add parameter for initiating bgrewriteaof on exceeding maximum AOF size

### LINKED ISSUE #540: [NEW] Limit maximum size on disk of AOF files. Avoid disk full, long load times.

kronwerk:
there are some scenarios of filling up the whole disk besides any chosen AOF rewrite strategy - the simplest is this:
- we have auto-aof-rewrite-percentage 100,
- and last automatic rewrite compacted AOF to 51% of disk,
-> we'll never try again.

we can lower auto-percentage, but in some cases it's not what we want to. I suggest that we add soft upper limit in this PR:
- if no AOF rewrite is running currently,
- and limit is set in config,
- and AOF exceed the limit,
- and last rewrite compacted AOF to somewhat smaller than limit (otherwise we might end up with useless attempts of uninterrupted rewrites)
-> bgrewriteaof is initiated.

this doesn't completely solves the problem, but gives some additional possibilities for some users to improve the situation initially described in #540 (noticed that solution from the issue seems to be inappropriate for some cases)

### DIFF
--- a/src/aof.c
+++ b/src/aof.c
@@ -1011,16 +1011,22 @@ int startAppendOnly(void) {
  * the first call is short, there is a end-of-space condition, so the next
  * is likely to fail. However apparently in modern systems this is no longer
  * true, and in general it looks just more resilient to retry the write. If
- * there is an actual error condition we'll get it at the next try. */
-ssize_t aofWrite(int fd, const char *buf, size_t len) {
-    ssize_t nwritten = 0, totwritten = 0;
+ * there is an actual error condition we'll get it at the next try.
+ * We also check for aof-max-size limit here returning "no space" on exceed. */
+ssize_t aofWrite(int fd, const char *buf, size_t len, off_t aof_current_size, unsigned long long aof_max_size) {
+    ssize_t nwritten = 0, totwritten = 0, nonewritten = -1;
+
+    if (aof_max_size && (unsigned long long)aof_current_size >= aof_max_size) {
+        errno = ENOSPC;
+        return nonewritten;
+    }
 
     while (len) {
         nwritten = write(fd, buf, len);
 
         if (nwritten < 0) {
             if (errno == EINTR) continue;
-            return totwritten ? totwritten : -1;
+            return totwritten ? totwritten : nonewritten;
         }
 
         len -= nwritten;
@@ -1120,7 +1126,7 @@ void flushAppendOnlyFile(int force) {
     }
 
     latencyStartMonitor(latency);
-    nwritten = aofWrite(server.aof_fd, server.aof_buf, sdslen(server.aof_buf));
+    nwritten = aofWrite(server.aof_fd, server.aof_buf, sdslen(server.aof_buf), server.aof_current_size, server.aof_max_size);
     latencyEndMonitor(latency);
     /* We want to capture different events for delayed writes:
      * when the delay happens with a pending fsync, or with a saving child
--- a/src/aof.c
+++ b/src/aof.c
@@ -1012,12 +1012,13 @@ int startAppendOnly(void) {
  * is likely to fail. However apparently in modern systems this is no longer
  * true, and in general it looks just more resilient to retry the write. If
  * there is an actual error condition we'll get it at the next try.
- * We also check for aof-max-size limit here returning "no space" on exceed. */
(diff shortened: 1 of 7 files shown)


kronwerk:
@xbasel @soloestoy pls take a look, I've dumped my previous similar PR where you were reviewers

src/aof.c
```diff
@@ -1011,16 +1011,22 @@ int startAppendOnly(void) {
  * the first call is short, there is a end-of-space condition, so the next
  * is likely to fail. However apparently in modern systems this is no longer
  * true, and in general it looks just more resilient to retry the write. If
- * there is an actual error condition we'll get it at the next try. */
-ssize_t aofWrite(int fd, const char *buf, size_t len) {
-    ssize_t nwritten = 0, totwritten = 0;
+ * there is an actual error condition we'll get it at the next try.
+ * We also check for aof-max-size limit here returning custom error on exceed. */
+ssize_t aofWrite(int fd, const char *buf, size_t len, off_t aof_current_size, unsigned long long aof_max_size) {
+    ssize_t nwritten = 0, totwritten = 0, nonewritten = -1;
```
xbasel: Seems like `nonewritten` should be a constant/literal.
Perhaps directly returning -1/`C_ERR` instead of using an extra variable.

src/aof.c
```diff
@@ -1011,16 +1011,22 @@ int startAppendOnly(void) {
  * the first call is short, there is a end-of-space condition, so the next
  * is likely to fail. However apparently in modern systems this is no longer
  * true, and in general it looks just more resilient to retry the write. If
- * there is an actual error condition we'll get it at the next try. */
-ssize_t aofWrite(int fd, const char *buf, size_t len) {
-    ssize_t nwritten = 0, totwritten = 0;
+ * there is an actual error condition we'll get it at the next try.
+ * We also check for aof-max-size limit here returning custom error on exceed. */
+ssize_t aofWrite(int fd, const char *buf, size_t len, off_t aof_current_size, unsigned long long aof_max_size) {
```
xbasel: aof_max_size -> size_t

xbasel: Not related to this change, but do you have any idea why aof_current_size (and some other aof-related variables) are already defined as off_t instead of size_t?

src/server.c
```diff
@@ -4627,13 +4627,17 @@ int writeCommandsDeniedByDiskError(void) {
     return DISK_ERROR_TYPE_NONE;
 }
 
+char *getAofWriteErrStr(int error_code) {
```
xbasel: If this function is strictly for AOF-related and always uses `server.aof_last_write_errno`, passing `error_code` is misleading. I'd remove the parameter and read `server.aof_last_write_errno` directly to avoid confusion and unintended wider use.

src/aof.c
kronwerk: not really, no idea

src/server.c
kronwerk: there is a place in code where we use errno also:
`serverLog(LL_WARNING, "Error writing to the AOF file: %s", getAofWriteErrStr(errno));`
changed to use code in both places inside this function

PR #1631, opened 2025-01-28
Fix timing issue in pause test

madolson:
The wait might execute exactly in 1000ms, which means there are still 9000ms left of the timeout, so we need to do less than or equal to in the comparison.

https://github.com/valkey-io/valkey/actions/runs/13000845302/job/36259020129


zuiderkwast (approved):
Wow, one millisecond.

enjoy-binbin (approved):

PR #1632, opened 2025-01-28
Optimize key expiration check with branch prediction hint

stav-bentov:
Optimized the key expiration check logic with branch prediction hints, enhancing performance during runtime.
* Added a conditional check `if (unlikely(dbSize(db) > 0))` before calling `expireIfNeededWithDictIndex`. This optimization:
  * Reduces unnecessary function calls when the database is empty.
  * Improves branch prediction efficiency by using the unlikely hint.

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -114,9 +114,11 @@ robj *lookupKey(serverDb *db, robj *key, int flags) {
         int expire_flags = 0;
         if (flags & LOOKUP_WRITE && !is_ro_replica) expire_flags |= EXPIRE_FORCE_DELETE_EXPIRED;
         if (flags & LOOKUP_NOEXPIRE) expire_flags |= EXPIRE_AVOID_DELETE_EXPIRED;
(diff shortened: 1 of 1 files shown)


src/db.c
```diff
@@ -114,9 +114,11 @@ robj *lookupKey(serverDb *db, robj *key, int flags) {
         int expire_flags = 0;
         if (flags & LOOKUP_WRITE && !is_ro_replica) expire_flags |= EXPIRE_FORCE_DELETE_EXPIRED;
         if (flags & LOOKUP_NOEXPIRE) expire_flags |= EXPIRE_AVOID_DELETE_EXPIRED;
-        if (expireIfNeededWithDictIndex(db, key, val, expire_flags, dict_index) != KEY_VALID) {
-            /* The key is no longer valid. */
-            val = NULL;
+        if (unlikely(dbSize(db) > 0)) { // This condition check found to improve branch prediction
```
madolson, line 117: Why is it "unlikely" that dbSize > 0? This doesn't make any sense. Can you publish the benchmarking your did to find this?

PR #1635, opened 2025-01-28
Refactor Event Processing Logic into Separate Function

stav-bentov:
Encapsulated loop logic into a dedicated function `processSingleFileEvent` to isolate responsibilities and simplify the main code flow.

### DIFF
--- a/src/ae.c
+++ b/src/ae.c
@@ -393,6 +393,58 @@ int aePoll(aeEventLoop *eventLoop, struct timeval *tvp) {
     return ret;
 }
 
+static void processSingleFileEvent(
+    aeEventLoop *eventLoop,
+    aeFiredEvent *firedEvent) {
+    int mask = firedEvent->mask;
+    int fd = firedEvent->fd;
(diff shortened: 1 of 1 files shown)


src/ae.c
```diff
@@ -393,6 +393,58 @@ int aePoll(aeEventLoop *eventLoop, struct timeval *tvp) {
     return ret;
 }
 
+static void processSingleFileEvent(
+    aeEventLoop *eventLoop,
+    aeFiredEvent *firedEvent) {
```
madolson, line 398:
```suggestion
static void processSingleFileEvent(aeEventLoop *eventLoop, aeFiredEvent *firedEvent) {
```

zuiderkwast (commented):
I don't think this change is necessary. I vote to not merge it.

zuiderkwast:
Btw, hi Stav and thanks for contributing. :)

Sometimes not all maintainers agree. It can be a good idea to wait some time for more opinions and to discuss it, especially if you want to convince us about something important.

PR #1636, opened 2025-01-28
Refactor Cluster Size Calculation Logic into Separate Function

stav-bentov:
Refactored the cluster size calculation logic into a dedicated function, `clusterDetermineClusterSize`, to enhance code organization and readability.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -5546,6 +5546,31 @@ void clusterCloseAllSlots(void) {
     memset(server.cluster->importing_slots_from, 0, sizeof(server.cluster->importing_slots_from));
 }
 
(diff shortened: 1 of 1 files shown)


src/cluster_legacy.c
```diff
@@ -5546,6 +5546,31 @@ void clusterCloseAllSlots(void) {
     memset(server.cluster->importing_slots_from, 0, sizeof(server.cluster->importing_slots_from));
 }
 
+static void clusterDetermineClusterSize(int *reachable_primaries) {
```
madolson, line 5549:
```suggestion
static int clusterDetermineClusterSize(void) {
```
Then return the size at the end.

madolson (commented):
I would vote not to accept this change. I don't think it materially improves the code quality.

PR #1637, opened 2025-01-28
Hex encode the data in dump test

madolson:
Addresses the failure here: https://github.com/valkey-io/valkey/actions/runs/13000845302/job/36259016156#step:5:7272.

This change does three things:
1. For some reason TCL 8.5 (which is used on macos) is handling `\x03ba` as `\0xba`, according to https://www.tcl-lang.org/man/tcl8.5/TclCmd/Tcl.htm#M27 so we encode "bar" using hex escapes too.
2. Fix a spacing issue. 
3. Make it so that if the restore fails, it immediately errors.

### DIFF
--- a/tests/unit/dump.tcl
+++ b/tests/unit/dump.tcl
@@ -125,16 +125,16 @@ start_server {tags {"dump"}} {
     } {} {needs:repl}
 
     test {RESTORE key with future RDB version, strict version check} {
-        #              str len    RDB 222 CRC64 checksum
-        #               |   |      |       |
+        #              str len  "bar"      RDB 222 CRC64 checksum
+        #               |   |     |         |       |
         set bar_dump "\x00\x03\x62\x61\x72\xde\x00\x0fYUza\xd3\xec\xe0"
         assert_error {ERR DUMP payload version or checksum are wrong} {r restore foo 0 $bar_dump replace}
     }
 
     test {RESTORE key with future RDB version, relaxed version check} {
-        #              str len    RDB 222 CRC64 checksum
-        #               |   |      |       |
         r config set rdb-version-check relaxed 
+        #               |type|len|           |  RDB  |      CRC64        |
+        #               |str | 3 |   "bar"   |  222  |     checksum      |
         r restore foo 0 "\x00\x03\x62\x61\x72\xde\x00\x0fYUza\xd3\xec\xe0" replace
         r config set rdb-version-check strict
         assert_equal {bar} [r get foo]
--- a/tests/unit/dump.tcl
+++ b/tests/unit/dump.tcl
@@ -127,20 +127,18 @@ start_server {tags {"dump"}} {
     test {RESTORE key with future RDB version, strict version check} {
         #              str len    RDB 222 CRC64 checksum
         #               |   |      |       |
-        set bar_dump "\x00\x03bar\xde\x00\x0fYUza\xd3\xec\xe0"
-       assert_error {ERR DUMP payload version or checksum are wrong} {r restore foo 0 $bar_dump replace}
+        set bar_dump "\x00\x03\x62\x61\x72\xde\x00\x0fYUza\xd3\xec\xe0"
+        assert_error {ERR DUMP payload version or checksum are wrong} {r restore foo 0 $bar_dump replace}
     }
 
     test {RESTORE key with future RDB version, relaxed version check} {
         #              str len    RDB 222 CRC64 checksum
         #               |   |      |       |
-        set bar_dump "\x00\x03bar\xde\x00\x0fYUza\xd3\xec\xe0"
-        r config set rdb-version-check relaxed
-        catch {r restore foo 0 $bar_dump replace} e
+        r config set rdb-version-check relaxed 
+        r restore foo 0 "\x00\x03\x62\x61\x72\xde\x00\x0fYUza\xd3\xec\xe0" replace
         r config set rdb-version-check strict
         assert_equal {bar} [r get foo]
-        set e
-    } {OK}
+    }
 
     test {DUMP of non existing key returns nil} {
         r dump nonexisting_key


enjoy-binbin (approved):

tests/unit/dump.tcl
```diff
@@ -127,20 +127,18 @@ start_server {tags {"dump"}} {
     test {RESTORE key with future RDB version, strict version check} {
         #              str len    RDB 222 CRC64 checksum
         #               |   |      |       |
-        set bar_dump "\x00\x03bar\xde\x00\x0fYUza\xd3\xec\xe0"
-       assert_error {ERR DUMP payload version or checksum are wrong} {r restore foo 0 $bar_dump replace}
+        set bar_dump "\x00\x03\x62\x61\x72\xde\x00\x0fYUza\xd3\xec\xe0"
```
hwware, line 130: I think the changes break the comment order in the line 128 and 129, please fix it before merged.

tests/unit/dump.tcl
```diff
@@ -127,20 +127,18 @@ start_server {tags {"dump"}} {
...
     }
 
     test {RESTORE key with future RDB version, relaxed version check} {
         #              str len    RDB 222 CRC64 checksum
         #               |   |      |       |
-        set bar_dump "\x00\x03bar\xde\x00\x0fYUza\xd3\xec\xe0"
-        r config set rdb-version-check relaxed
-        catch {r restore foo 0 $bar_dump replace} e
+        r config set rdb-version-check relaxed 
+        r restore foo 0 "\x00\x03\x62\x61\x72\xde\x00\x0fYUza\xd3\xec\xe0" replace
```
hwware, line 138: I suggest swapping the line 137 and 138, and fix the comment order as well,

tests/unit/dump.tcl
```diff
@@ -127,20 +127,18 @@ start_server {tags {"dump"}} {
     test {RESTORE key with future RDB version, strict version check} {
         #              str len    RDB 222 CRC64 checksum
         #               |   |      |       |
```
zuiderkwast: This aligns the comment with the string you changed (if I got this right).

```suggestion
        #              str len  "bar"      RDB 222 CRC64 checksum
        #               |   |     |         |       |
```

zuiderkwast (commented):
Thanks! MacOS is weird...

The comments above the changed strings were supposed to point to the right parts inside the strings, to make it easier to read the strings.

zuiderkwast:
> For some reason macos is not properly handling \x03ba. I don't know why, and couldn't find any existing issue for it. Replacing it with it's actual

I think I found it. MacOS is running TCL 8.5, right? https://www.tcl-lang.org/man/tcl8.5/TclCmd/Tcl.htm#M27

> \xhh
>     The hexadecimal digits hh give an eight-bit hexadecimal value for the Unicode character that will be inserted. Any number of hexadecimal digits may be present; however, all but the last two are ignored (the result is always a one-byte quantity). The upper bits of the Unicode character will be 0.

This seems to have change in TCL 8.6 https://www.tcl-lang.org/man/tcl8.6/TclCmd/Tcl.htm#M27 to 

> \xhh
>     The hexadecimal digits hh (one or two of them) give an eight-bit hexadecimal value for the Unicode character that will be inserted. The upper bits of the Unicode character will be 0 (i.e., the character will be in the range U+000000–U+0000FF).

tests/unit/dump.tcl
```diff
@@ -127,20 +127,18 @@ start_server {tags {"dump"}} {
...
-        set bar_dump "\x00\x03bar\xde\x00\x0fYUza\xd3\xec\xe0"
-        r config set rdb-version-check relaxed
-        catch {r restore foo 0 $bar_dump replace} e
+        r config set rdb-version-check relaxed 
+        r restore foo 0 "\x00\x03\x62\x61\x72\xde\x00\x0fYUza\xd3\xec\xe0" replace
         r config set rdb-version-check strict
         assert_equal {bar} [r get foo]
-        set e
-    } {OK}
+    }
```
zuiderkwast, line 141: The reason I used catch instead of letting it fail immediately was to restore the config afterwards so even if this test case fails, it doesn't affect other test cases. Maybe that's too fancy.

tests/unit/dump.tcl
```diff
@@ -127,20 +127,18 @@ start_server {tags {"dump"}} {
...
         #               |   |      |       |
-        set bar_dump "\x00\x03bar\xde\x00\x0fYUza\xd3\xec\xe0"
-       assert_error {ERR DUMP payload version or checksum are wrong} {r restore foo 0 $bar_dump replace}
+        set bar_dump "\x00\x03\x62\x61\x72\xde\x00\x0fYUza\xd3\xec\xe0"
+        assert_error {ERR DUMP payload version or checksum are wrong} {r restore foo 0 $bar_dump replace}
     }
 
     test {RESTORE key with future RDB version, relaxed version check} {
         #              str len    RDB 222 CRC64 checksum
         #               |   |      |       |
```
zuiderkwast: We can move this comment to just before the string, and fix the alignment.

```suggestion
```

tests/unit/dump.tcl
zuiderkwast: The `config set` needs to be before `restore`, but we can move the comment to here and fix the alignment.

Is this comment style better?

```suggestion
        #               |type|len|           |  RDB  |      CRC64        |
        #               |str | 3 |   "bar"   |  222  |     checksum      |
        r restore foo 0 "\x00\x03\x62\x61\x72\xde\x00\x0fYUza\xd3\xec\xe0" replace
```

zuiderkwast (approved):
Since you're traveling, I'll apply the suggestions and merge it so we can fix daily.

hwware (approved):

Issue #1642, opened 2025-01-29
[BUG] heap-use-after-free in cluster bus


xingbowang:
**Describe the bug**

A heap-use-after-free is found in cluster bus code.

- freeClusterNode free the human_nodename. https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L1725
- Then it calls freeClusterLink to free the links. https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L1730
- freeClusterLink print human_nodename here, which just got freed by the caller freeClusterNode. https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L1383 

**To reproduce**

Steps to reproduce the behavior and/or a minimal code sample.
update file tests/cluster/run.tcl to enable debug log, then run `make test-cluster`.
```
~/Documents/GitHub/valkey$git diff
diff --git a/tests/cluster/run.tcl b/tests/cluster/run.tcl
index d29f17db7..917734f71 100644
--- a/tests/cluster/run.tcl
+++ b/tests/cluster/run.tcl
@@ -17,6 +17,7 @@ proc main {} {
         "appendonly yes"
         "enable-protected-configs yes"
         "enable-debug-command yes"
+        "loglevel debug"
         "save ''"
     }
     run_tests
```

**Expected behavior**

No heap-use-after-free is detected in the bug reported.

**Additional information**

Any additional information that is relevant to the problem.

PR #1643, opened 2025-01-29
Fix a heap-use-after-free bug in cluster bus

xingbowang:
https://github.com/valkey-io/valkey/issues/1642

A heap-use-after-free is found in cluster bus code.

freeClusterNode free the human_nodename. https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L1725
Then it calls freeClusterLink to free the links. https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L1730
freeClusterLink print human_nodename here, which just got freed by the caller freeClusterNode. https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L1383

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -1721,14 +1721,16 @@ void freeClusterNode(clusterNode *n) {
     nodename = sdsnewlen(n->name, CLUSTER_NAMELEN);
     serverAssert(dictDelete(server.cluster->nodes, nodename) == DICT_OK);
     sdsfree(nodename);
-    sdsfree(n->hostname);
-    sdsfree(n->human_nodename);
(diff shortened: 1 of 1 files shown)


hpatro (approved):
LGTM

hpatro:
@xingbowang could you fix the DCO https://github.com/valkey-io/valkey/pull/1643/checks?check_run_id=36369847238 ?

pieturin:
Thank you for the fix. This was due to the new debug log I introduced in `freeClusterLink()` in one of my previous PRs. The fix looks good.

xingbowang:
Will fix the DCO.

PR #1648, opened 2025-01-30
Update daily failure notification job list

hpatro:
Two jobs were missing from the job list for failure notification

* test-ubuntu-tls-io-threads
* test-sanitizer-force-defrag

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1262,7 +1262,7 @@ jobs:
   notify-about-job-results:
     runs-on: ubuntu-latest
     if: always() && github.event_name == 'schedule' && github.repository == 'valkey-io/valkey'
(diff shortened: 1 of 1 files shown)


hpatro:
We should possibly move to list the jobs dynamically rather than depending on the hardcoded list.

zuiderkwast (approved):

zuiderkwast:
> We should possibly move to list the jobs dynamically rather than depending on the hardcoded list.

Yeah... But that seems more difficult. :)

hpatro:
> > We should possibly move to list the jobs dynamically rather than depending on the hardcoded list.
> 
> Yeah... But that seems more difficult. :)

Yep, couldn't find an easier way immediately.

PR #1651, opened 2025-01-31
Fix ACL regression with renamed commands

rainsupreme:
This regression occurred when we converted the command tables from dict to hashtable. For both server.commands and server.orig_commands pointed to the same struct and used the same field as a key. When a command is renamed it changed this key without updating server.orig_commands, causing it to be unable to find any renamed command.

server.orig_commands is used by ACLs which are always configured and manipulated using the original command names. For example, `-config` should always remove permission to use the `config` command, even if its been renamed to `asdfconfig`.

I added a tcl test which passes before the command table conversion, breaks after the conversion, and passes again with this fix.

Changes in acl.c are only for code cleanup and not necessary for the fix:
- deleted some dead code
- added `static` to methods (clarifies scope/use, aids compiler optimization)
- couldn't resist changing 1 `goto` to `break` 🤷🏻‍♀️


tests/unit/acl.tcl
```diff
@@ -1244,3 +1244,60 @@ start_server {overrides {user "default on nopass ~* +@all -flushdb"} tags {acl e
     }
 }
 
+tags {acl external:skip} {
+    set renames [list rename-command "PING RainWasHere" rename-command "RainWasHere VPING" rename-command "INCR VINCR" rename-command "SET VSET" rename-command "SADD VSADD" rename-command "HGET VHGET" rename-command "HSET VHSET" rename-command "CONFIG VCONFIG" ]
```
zuiderkwast: It's possible to line break TCL with escaped line breaks.

```suggestion
    set renames [list rename-command "PING RainWasHere" \
                      rename-command "RainWasHere VPING" \
                      rename-command "INCR VINCR" \
                      rename-command "SET VSET" \
                      rename-command "SADD VSADD" \
                      rename-command "HGET VHGET" \
                      rename-command "HSET VHSET" \
                      rename-command "CONFIG VCONFIG" ]
```

src/acl.c
```diff
@@ -152,17 +152,6 @@ struct ACLUserFlag {
     {NULL, 0} /* Terminator. */
 };
 
-struct ACLSelectorFlags {
```
zuiderkwast, line 155: So this was dead code?

src/acl.c
```diff
@@ -735,50 +724,14 @@ int ACLSetSelectorCategory(aclSelector *selector, const char *category, int allo
     return C_OK;
 }
 
-void ACLCountCategoryBitsForCommands(hashtable *commands,
```
zuiderkwast, line 738: And this, dead code? Fascinating. When you added static, the compiler told you it's unused?

src/server.c
```diff
@@ -519,7 +519,12 @@ int hashtableResizeAllowed(size_t moreMem, double usedRatio) {
...
-const void *hashtableCommandGetKey(const void *element) {
+const void *hashtableCommandGetCurrentName(const void *element) {
+    struct serverCommand *command = (struct serverCommand *)element;
+    return command->current_name;
+}
+
+const void *hashtableCommandGetOriginalName(const void *element) {
     struct serverCommand *command = (struct serverCommand *)element;
     return command->fullname;
 }
```
zuiderkwast, line 530: Have you considered adding a field `command->original_name` instead and keep `command->fullname` as the current name? It matches the other naming better:

| current | original |
| --- | --- |
| server.commands | server.orig_commands   |
| commandSetType  | originalCommandSetType |
| hashtableCommandGet(Current?)Name | hashtableCommandGetOriginalName |

src/config.c
```diff
@@ -542,8 +542,14 @@ void loadServerConfigFromString(char *config) {
...
-                sdsfree(cmd->fullname);
-                cmd->fullname = sdsdup(argv[2]);
+                if (cmd->current_name != cmd->fullname) {
+                    /* cmd->current_name initially points to the same string,
+                     * and we can't leave fullname dangling.
+                     * If its been renamed previously then this string is only
+                     * used here and should be freed. */
+                    sdsfree(cmd->current_name);
+                }
+                cmd->current_name = sdsdup(argv[2]);
```
zuiderkwast: Not sure if this verbose comment is necessary. It's obvious, isn't it?

zuiderkwast (commented):
Great, thanks for finding and fixing this before the release!

src/config.c
rainsupreme: I'm not sure this ownership arrangement is obvious, but it should be easy enough to figure out in a few minutes of reading code. I guess the comment in the command struct should be enough.

src/acl.c
rainsupreme: Yep! Kind of handy that way. I also searched the entire project for any occurrence of these strings.

src/config.c
zuiderkwast: When you see `cmd->current_name` is assigned with `strdup` below, it's quite obvious it needs to be free'd before that, except if the the string is shared with some other variable..

Keep a short comment? One line is enough?

src/server.c
rainsupreme: I initially wrote it that way, but I found that I'd have to change a lot of references to `fullname` which expect it to be the original name. Before the conversion `fullname` was left unchanged even when renamed - only the key in the dictEntry would be different. It seems that by changing `fullname` this regression also affected logging, error messages, command help text, the latency command... Adding `current_name` more closely matches the original logic before conversion and fixes all of that.

zuiderkwast: OK!

src/acl.c
rainsupreme: Yes 🪦

tests/unit/acl.tcl
```diff
@@ -1244,3 +1244,67 @@ start_server {overrides {user "default on nopass ~* +@all -flushdb"} tags {acl e
...
+                      rename-command "HSET VHSET" \
+                      rename-command "CONFIG VCONFIG" ]
+    start_server [list config_lines $renames ] {
+        test {ACL still denies individual renamed commands} {
+            r ACL setuser newuser on nopass ~* +acl -ping +incr
+            r AUTH newuser anypass
+            r VINCR mycounter ; # Should not raise an error
+            catch {r vping} e
+            set e
+        } {*NOPERM*ping*}
```
zuiderkwast: Nit: We have `assert_error` for this.

```suggestion
            assert_error {*NOPERM*ping*} {r vping}
        }
```

tests/unit/acl.tcl
```diff
@@ -1244,3 +1244,67 @@ start_server {overrides {user "default on nopass ~* +@all -flushdb"} tags {acl e
...
+        test {ACL command classes aren't affected by command renaming} {
+            r ACL setuser newuser -@all +@set +acl
+            r VSADD myset a b c; # Should not raise an error
+            r ACL setuser newuser +@all -@string
+            r VSADD myset a b c; # Again should not raise an error
+            # String commands instead should raise an error
+            catch {r vset foo bar} e
+            r ACL setuser newuser allcommands; # Undo commands ACL
+            set e
+        } {*NOPERM*set*}
```
zuiderkwast: Could use assert_error here too.

```suggestion
            assert_error {*NOPERM*set*} {r vset foo bar}
            r ACL setuser newuser allcommands; # Undo commands ACL
        }
```

zuiderkwast (approved):
OK, LGTM! (I looked at the test cases a bit more this time and added two nits.)

PR #1658, opened 2025-02-03
Opaque robj: add get/set methods and remove all robj->ptr refs

rainsupreme:
I wanted to know if I could do this by quickly writing some scripts. It probably took twice as long as I thought, but only 4-5 hours so not terrible, and now I know. 🤷🏻‍♀️ All but the necessary references in object.c have been replaced using objectGetVal and objectSetVal.

I know we want to make robj opaque so it's easier to refactor it or change its internal structure. I thought there was an issue for this but I can't find it offhand.

### DIFF
--- a/src/acl.c
+++ b/src/acl.c
@@ -1450,7 +1450,7 @@ void ACLInit(void) {
  *  ENOENT: if the specified user does not exist at all.
  */
 int ACLCheckUserCredentials(robj *username, robj *password) {
-    user *u = ACLGetUserByName(username->ptr, sdslen(username->ptr));
+    user *u = ACLGetUserByName(objectGetVal(username), sdslen(objectGetVal(username)));
     if (u == NULL) {
         errno = ENOENT;
         return C_ERR;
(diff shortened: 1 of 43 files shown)


zuiderkwast:
Awesome! Tbh, I'd prefer the complete change in one PR so we can get the benefit in memory usage. A separate commit within the same PR for the assignments and then a final commit to actually save memory in the object struct.

Feel free to write down you regexes so we can review those ~~instead of~~ alongside the actual diff. :D

rainsupreme:
Ok, I added `objectSetVal`.

I could post the python scripts I used, but they're pretty bad. I only worked on them to the extent that it'd save me time - They mangle a few instances but did the bulk of the editing for me. Here's the one I used for the last objectSetVal commit. The json I load is derived from VS Code's "find all references" feature.
```python
import json

with open('ptrrefs.json') as f:
    refs = json.load(f)

for file in list(refs.keys()):
    with open(file, 'r') as src:
        lines = src.readlines()
    print('====', file, '====')
    for (lineno, char) in refs[file]:
        lineno -= 1 # line number is 1-based
        char -= 1 # char number also 1-based

        line = lines[lineno]
        count = line.count('->ptr = ')
        if count == 0:
            continue
        assert(count == 1)
        (robj, val_ptr) = line.strip()[:-1].split('->ptr = ')
        whitespace = line[:len(line) - len(line.lstrip())]
        print(lineno, line[:-1])
        lines[lineno] =  f'{whitespace}objectSetVal({robj}, {val_ptr});\n'

    with open(file, 'w') as src:
        src.writelines(lines)
```

zuiderkwast (commented):
Ok, very nice. Next thing to try:

1. Rename the ptr field to something else, update getval and setval, see if it still compiles and works.
2. For EMBSTR, ignore the ptr field and just embed data att the offsetof(ptr) and compute the val dynamically. We may need to keep pointer field so we don't affect the sizeof(robj) because we have arrays of robjs in some places. Or we can rewrite those in some way maybe.

madolson:
@SoftlyRaining We can close this now right? It's replaced by https://github.com/valkey-io/valkey/pull/2516

PR #1659, opened 2025-02-04
Code cleanup in object.c to support future work

rainsupreme:
Summary of changes:
- Define/allocate layout of robj's embedded data in once place instead of two
- More consistent naming: o to refer to robj, val/ptr/val_ptr to refer to contained value
- Add static to helpers to clarify scope/usage and improve compiler optimization

My primary goal is to make the structure and memory layout of robj easier to change and work with in the future. This will support future work like eliminating the 8 byte `ptr` when value is embedded directly, and potentially expand value embedding beyond `OBJ_ENCODING_EMBSTR` and `OBJ_ENCODING_INT`.

I'm a bit unsure about the function signature of `createObjectWithEmbeddedData` - let me know if you have a cleaner idea!

### DIFF
--- a/src/server.h
+++ b/src/server.h
@@ -2894,11 +2894,11 @@ void trimStringObjectIfNeeded(robj *o, int trim_small_values);
 #define sdsEncodedObject(objptr) (objptr->encoding == OBJ_ENCODING_RAW || objptr->encoding == OBJ_ENCODING_EMBSTR)
 
 /* Objects with key attached, AKA valkey (val+key) objects */
-robj *createObjectWithKeyAndExpire(int type, void *ptr, const sds key, long long expire);
-robj *objectSetKeyAndExpire(robj *val, sds key, long long expire);
-robj *objectSetExpire(robj *val, long long expire);
-sds objectGetKey(const robj *val);
-long long objectGetExpire(const robj *val);
+robj *createObjectWithEmbeddedData(int type, void *val_ptr, const sds key, long long expire, size_t val_embedded_size, size_t *val_embed_capacity);
+robj *objectSetKeyAndExpire(robj *o, sds key, long long expire);
+robj *objectSetExpire(robj *o, long long expire);
+sds objectGetKey(const robj *o);
+long long objectGetExpire(const robj *o);
 
 /* Synchronous I/O with timeout */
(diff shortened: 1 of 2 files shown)


rainsupreme:
I'm combining this with #1658 I think it'll be cleaner

src/server.h
```diff
@@ -2894,11 +2894,11 @@ void trimStringObjectIfNeeded(robj *o, int trim_small_values);
 #define sdsEncodedObject(objptr) (objptr->encoding == OBJ_ENCODING_RAW || objptr->encoding == OBJ_ENCODING_EMBSTR)
 
 /* Objects with key attached, AKA valkey (val+key) objects */
-robj *createObjectWithKeyAndExpire(int type, void *ptr, const sds key, long long expire);
-robj *objectSetKeyAndExpire(robj *val, sds key, long long expire);
-robj *objectSetExpire(robj *val, long long expire);
-sds objectGetKey(const robj *val);
-long long objectGetExpire(const robj *val);
+robj *createObjectWithEmbeddedData(int type, void *val_ptr, const sds key, long long expire, size_t val_embedded_size, size_t *val_embed_capacity);
```
zuiderkwast, line 2897: The deduplication is good. Thanks for that.

I don't think we should expose this function. It's more complex than what the caller needs. So far, embedded strings are created using `createStringObject` with size checks inside of object.c to decide to embed or not. We can expose `createStringObjectWithKeyAndExpire` too if you want.

For non-strings, I think we should still expose `createObjectWithKeyAndExpire` and make it a thin wrapper of `createObjectWithEmbeddedData` which can be internal.

WDYT?

rainsupreme: Yeah, good point. It makes sense not to expose that complex function signature. For now I think it makes sense not to expose functions we don't actually use outside of object.c - we can always expose them later when we need them.

It kind of makes sense to have createObjectWithKeyAndExpire, but it isn't used anywhere. Do we have a near-future use for it?

zuiderkwast: Oh, if it's not used, let's not expose it.

I have a local branch where I want to create the object with expire from the start in commands like SETEX, but I don't think I'll prioritize it soon.

PR #1669, opened 2025-02-05
Increase fixed delay with small randomness in failover delay logic to avoid same election

sungming2:
### Problem
We introduced the primary failover rank to delay elections when multiple primaries fail simultaneously, preventing elections in the same epoch and failover timeouts caused by insufficient votes. However, under certain timing conditions, this delay can still result in multiple nodes initiating elections in the same epoch.

e.g., the two replicas started the election at the same time and got failover timeout then restarted
```
654043:S 30 Jan 2025 16:19:37.076 * Starting a failover election for epoch 9, node config epoch is 7
654043:S 30 Jan 2025 16:19:39.110 * Currently unable to failover: Waiting for votes, but majority still not reached.
654043:S 30 Jan 2025 16:19:39.110 * Needed quorum: 4. Number of votes received so far: 2
...
654043:S 30 Jan 2025 16:19:48.034 * Currently unable to failover: Failover attempt expired.
654043:S 30 Jan 2025 16:19:48.034 * Needed quorum: 4. Number of votes received so far: 2
...
654043:S 30 Jan 2025 16:19:57.876 * Starting a failover election for epoch 10, node config epoch is 7
654043:S 30 Jan 2025 16:19:58.440 * Failover election won: I'm the new primary.
654043:S 30 Jan 2025 16:19:58.440 * configEpoch set to 10 after successful failover
```

```
654224:S 30 Jan 2025 16:19:37.074 * Starting a failover election for epoch 9, node config epoch is 4
654224:S 30 Jan 2025 16:19:38.095 * Needed quorum: 4. Number of votes received so far: 3
654224:S 30 Jan 2025 16:19:39.104 * Currently unable to failover: Waiting for votes, but majority still not reached.
...
654224:S 30 Jan 2025 16:19:47.096 * Currently unable to failover: Failover attempt expired.
654224:S 30 Jan 2025 16:19:47.096 * Needed quorum: 4. Number of votes received so far: 3
...
654224:S 30 Jan 2025 16:19:58.472 * Starting a failover election for epoch 11, node config epoch is 4
654224:S 30 Jan 2025 16:19:58.849 * Failover election won: I'm the new primary.
654224:S 30 Jan 2025 16:19:58.849 * configEpoch set to 11 after successful failover
```
### Solution
1. In some cases, replica node may not receive all FAILED propagation messages before the election starts, potentially resulting in an incorrect failover rank or defaulting to rank 0.

2. To prevent simultaneous elections, we need to introduce a sufficient delay before starting the election. I observed that replicas can begin elections for the same epoch within a 500ms difference, and the random delay should be less than the fixed delay to avoid the same epoch election issue in the worst case.
    * worst example of `500 + random() % 500`:
R1 (starts failover handling at 10:00:00.000): 500ms + (499 % 500) = 999ms delay → Election starts at `10:00:00.999`
R2 (starts failover handling at 10:00:00.500): 500ms + (0 % 500) = 500ms delay → Election starts at `10:00:01.000`

3. Once the failover rank is upgraded, it should not be downgraded again because it will likely trigger the same epoch election.
e.g.,
Untitled Diagram drawio (1)

### Test
Ran test over a thousand times

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3260,6 +3260,7 @@ int clusterProcessPacket(clusterLink *link) {
         sender_claimed_config_epoch = ntohu64(hdr->configEpoch);
         if (sender_claimed_current_epoch > server.cluster->currentEpoch)
             server.cluster->currentEpoch = sender_claimed_current_epoch;
+
         /* Update the sender configEpoch if it is a primary publishing a newer one. */
         if (sender_claims_to_be_primary && sender_claimed_config_epoch > sender->configEpoch) {
             sender->configEpoch = sender_claimed_config_epoch;
@@ -4675,9 +4676,12 @@ int clusterGetFailedPrimaryRank(void) {
         if (!nodeFailed(node) || !clusterNodeIsVotingPrimary(node) || node->num_replicas == 0) continue;
 
         /* If cluster-replica-validity-factor is enabled, skip the invalid nodes. */
-        if (server.cluster_replica_validity_factor) {
-            if ((now - node->fail_time) > (server.cluster_node_timeout * server.cluster_replica_validity_factor))
+        if (nodeFailed(node) && server.cluster_replica_validity_factor) {
+            if ((now - node->fail_time) > (server.cluster_node_timeout * server.cluster_replica_validity_factor)) {
+                serverLog(LL_DEBUG, "Skip failed primary rank since validity factor is enabled. node failed time: %llu",
+                          (unsigned long long)node->fail_time);
                 continue;
+            }
         }
 
         if (memcmp(node->shard_id, myself->shard_id, CLUSTER_NAMELEN) < 0) rank++;
@@ -4860,7 +4864,7 @@ void clusterHandleReplicaFailover(void) {
     if (auth_age > auth_retry_time) {
         server.cluster->failover_auth_time = now +
                                              500 +           /* Fixed delay of 500 milliseconds, let FAIL msg propagate. */
-                                             random() % 500; /* Random delay between 0 and 500 milliseconds. */
+                                             random() % 100; /* Random delay between 0 and 100 milliseconds. */
         server.cluster->failover_auth_count = 0;
         server.cluster->failover_auth_sent = 0;
         server.cluster->failover_auth_rank = clusterGetReplicaRank();
@@ -4869,10 +4873,10 @@ void clusterHandleReplicaFailover(void) {
          * less updated replication offset, are penalized. */
         server.cluster->failover_auth_time += server.cluster->failover_auth_rank * 1000;
         /* We add another delay that is proportional to the failed primary rank.
-         * Specifically 0.5 second * rank. This way those failed primaries will be
+         * Specifically (1 second + random % 100) * rank. This way those failed primaries will be
          * elected in rank to avoid the vote conflicts. */
         server.cluster->failover_failed_primary_rank = clusterGetFailedPrimaryRank();
-        server.cluster->failover_auth_time += server.cluster->failover_failed_primary_rank * 500;
+        server.cluster->failover_auth_time += server.cluster->failover_failed_primary_rank * (1000 + random() % 100);
         /* However if this is a manual failover, no delay is needed. */
         if (server.cluster->mf_end) {
             server.cluster->failover_auth_time = now;
@@ -4917,7 +4921,7 @@ void clusterHandleReplicaFailover(void) {
 
         int new_failed_primary_rank = clusterGetFailedPrimaryRank();
         if (new_failed_primary_rank != server.cluster->failover_failed_primary_rank) {
-            long long added_delay = (new_failed_primary_rank - server.cluster->failover_failed_primary_rank) * 500;
+            long long added_delay = (new_failed_primary_rank - server.cluster->failover_failed_primary_rank) * (1000 + random() % 100);
             server.cluster->failover_auth_time += added_delay;
             server.cluster->failover_failed_primary_rank = new_failed_primary_rank;
             serverLog(LL_NOTICE, "Failed primary rank updated to #%d, added %lld milliseconds of delay.",
@@ -4941,8 +4945,8 @@ void clusterHandleReplicaFailover(void) {
     if (server.cluster->failover_auth_sent == 0) {
         server.cluster->currentEpoch++;
         server.cluster->failover_auth_epoch = server.cluster->currentEpoch;
-        serverLog(LL_NOTICE, "Starting a failover election for epoch %llu, node config epoch is %llu",
-                  (unsigned long long)server.cluster->currentEpoch, (unsigned long long)nodeEpoch(myself));
+        serverLog(LL_NOTICE, "Starting a failover election for epoch %llu, node config epoch is %llu, failover primary rank is %llu",
+                  (unsigned long long)server.cluster->currentEpoch, (unsigned long long)nodeEpoch(myself), (unsigned long long)server.cluster->failover_failed_primary_rank);
         clusterRequestFailoverAuth();
         server.cluster->failover_auth_sent = 1;
         clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_UPDATE_STATE | CLUSTER_TODO_FSYNC_CONFIG);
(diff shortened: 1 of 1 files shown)


sungming2:
@hpatro @enjoy-binbin could you take a look at this pr please when you have time

enjoy-binbin:
> Once the failover rank is upgraded, it should not be downgraded again because it will likely trigger the same epoch election.

can you try to keep this only change and test it again? I did have some doubt when i added this line.

src/cluster_legacy.c
```diff
@@ -3260,17 +3260,18 @@ int clusterProcessPacket(clusterLink *link) {
...
+
         if (sender_claims_to_be_primary && sender_claimed_config_epoch > sender->configEpoch) {
+            /* Update the sender configEpoch if it is a primary publishing a newer one. */
             sender->configEpoch = sender_claimed_config_epoch;
             clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_FSYNC_CONFIG);
 
+            /* Another node has claimed an epoch greater than or equal to ours.
+             * If we have an ongoing election, reset it because we cannot win
+             * with an epoch smaller than or equal to the incoming claim. This
+             * allows us to start a new election as soon as possible. */
```
enjoy-binbin: also please remove these non-related changes.

src/cluster_legacy.c
```diff
@@ -4860,9 +4871,10 @@ void clusterHandleReplicaFailover(void) {
     if (auth_age > auth_retry_time) {
         server.cluster->failover_auth_time = now +
                                              500 +           /* Fixed delay of 500 milliseconds, let FAIL msg propagate. */
-                                             random() % 500; /* Random delay between 0 and 500 milliseconds. */
+                                             random() % 100; /* Random delay between 0 and 100 milliseconds. */
         server.cluster->failover_auth_count = 0;
         server.cluster->failover_auth_sent = 0;
+        server.cluster->failover_failed_primary_rank = 0;
```
enjoy-binbin: and this, see line 4886

src/cluster_legacy.c
```diff
@@ -4783,6 +4787,13 @@ void clusterFailoverReplaceYourPrimary(void) {
 
     /* 5) If there was a manual failover in progress, clear the state. */
     resetManualFailover();
+
+    /* 6) Since we have changed to a new primary node, the previously set
+     * failover_auth_time and failover_auth_sent should no longer be used */
+    if (server.cluster->failover_auth_time) {
+        server.cluster->failover_auth_time = 0;
+        server.cluster->failover_auth_sent = 0;
+    }
```
enjoy-binbin: we have became a new primary, does this affect it? we will no longer enter the failover check logic. you are saying the primary became a replica in a short time and the auth_time is wrong? We do clear the auth_time in clusterSetPrimary

sungming2: Yes, that was the reason why I added this cleanup. I missed this point `We do clear the auth_time in clusterSetPrimary`.
I will remove it thanks

src/cluster_legacy.c
```diff
@@ -4872,7 +4876,7 @@ void clusterHandleReplicaFailover(void) {
          * Specifically 0.5 second * rank. This way those failed primaries will be
          * elected in rank to avoid the vote conflicts. */
```
hpatro: This comment is no longer true. Could you update it?

src/cluster_legacy.c
```diff
@@ -4672,12 +4673,15 @@ int clusterGetFailedPrimaryRank(void) {
         clusterNode *node = dictGetVal(de);
 
         /* Skip nodes that do not need to participate in the rank. */
-        if (!nodeFailed(node) || !clusterNodeIsVotingPrimary(node) || node->num_replicas == 0) continue;
+        if (!(nodeFailed(node) || nodeTimedOut(node)) || !clusterNodeIsVotingPrimary(node) || node->num_replicas == 0) continue;
```
hpatro: It doesn't hurt to explain the conditional statement in code comment. Would be easier to understand why we choose to skip the failed/partially failed node.

hpatro (commented):
Could you update the high level comment around the behavior of this PR.

sungming2:
> can you try to keep this only change and test it again? I did have some doubt when i added this line.

I tried it hundreds of times with only this change `>`, I don't see any specific error due to this change but I can see some same epoch elections because it has insufficient delay

R1
```
162615:S 06 Feb 2025 11:41:23.803 * Start of election delayed for 672 milliseconds (rank #0, primary rank #0, offset 14).
162615:S 06 Feb 2025 11:41:24.210 * Failed primary rank updated to #1, added 500 milliseconds of delay.
162615:S 06 Feb 2025 11:41:25.014 * Starting a failover election for epoch 8, node config epoch is 0, rank is 1
```

R2
```
162730:S 06 Feb 2025 11:41:24.209 * Start of election delayed for 735 milliseconds (rank #0, primary rank #0, offset 14).
162730:S 06 Feb 2025 11:41:25.012 * Starting a failover election for epoch 8, node config epoch is 3, rank is 0
```

enjoy-binbin:
I want to know whether the main reason here is delay or because the pfail node is not included in the rank. In the logs i am guessing the main reason is the delay, can this be fixed if we only improve the delay logic? 

From the log, we can see that it is because of the `added 500 milliseconds of delay.` that they initiate the election at the same time. This is a familiar shortcoming and widely accepted. The reason i think is that, the tcl `pause_process` has some delay so these primaries doesn't down in the same time but within 500ms range. Maybe it is a acceptable shortcoming we can try to improve the TCL code? Or we do can improve the delay logic, i haven't figured this out yet (it does happen, no matter how you adjust it, the delay will always randomly conflict)

> Once the failover rank is upgraded, it should not be downgraded again because it will likely trigger the same epoch election.

Is this 3 also caused by the delay design? My original idea was that when a primary is up again, the primaries that were delayed by rank can initiate elections as soon as possible.

sungming2:
I agree with you, but what I am really expecting with this change is it will reduce the chance of elections starting at the same time. (there are cases that pfail node is not included that is why I added this change) 

And yes, we can refine this delay logic or implement an entirely new approach later, such as adding a flag to indicate whether a node is in the election process and by comparing the sender's current epoch with its auth epoch, we detect/assume an election conflict and immediately reset it if necessary (could be wrong, just my instant thought).


> My original idea was that when a primary is up again, the primaries that were delayed by rank can initiate elections as soon as possible.

You are right, but I thought it would be better to add a slight delay rather than risk an election conflict since both replicas must wait until the auth timeout (node timeout * 4) in the worst case.

I can just apply increasing delay with small random only to this pr (1000 + random() %100) if you are ok with it as this is a clear improvement we can do right away. What do you think?

enjoy-binbin:
> I can just apply increasing delay with small random only to this pr (1000 + random() %100) if you are ok with it as this is a clear improvement we can do right away. What do you think?

yes, please do it, we also need other eyes to take a look with this delay logic after the changes.

hpatro:
@enjoy-binbin what are your thoughts on this? Should we target it for 8.1 ?

src/cluster_legacy.c
```diff
@@ -4675,9 +4676,12 @@ int clusterGetFailedPrimaryRank(void) {
         if (!nodeFailed(node) || !clusterNodeIsVotingPrimary(node) || node->num_replicas == 0) continue;
 
         /* If cluster-replica-validity-factor is enabled, skip the invalid nodes. */
-        if (server.cluster_replica_validity_factor) {
-            if ((now - node->fail_time) > (server.cluster_node_timeout * server.cluster_replica_validity_factor))
+        if (nodeFailed(node) && server.cluster_replica_validity_factor) {
```
enjoy-binbin: do we still need this `nodeFailed(node)`, doesn't the if continue above skip it? Let's also remove the debug log?

enjoy-binbin:
> what are your thoughts on this? Should we target it for 8.1 ?

We have been running this delay internally for a long time and have not found any problems (or maybe they have not been exposed), so i guess there is no rush. I am ok with this delay logic, but i think this needs an extra eyes. @madolson @PingXie @zuiderkwast

src/cluster_legacy.c
sungming2: I forgot to remove it. Thanks

madolson:
The update seems OK to me. It feels a bit arbitrary, and it would be nice if it was more dynamic instead of relying on a bunch of internal constants. I'd be okay merging this for 8.1.

sungming2:
Plan to remove fixed delay: https://github.com/valkey-io/valkey/pull/2227

zuiderkwast:
#2227 only removes the delay if it is known to be safe, in a special case. It falls back to the current behavior.

The changed delays in this PR can still be useful, I think. Sorry, I didn't look at this PR before.

If two primaries are down at the same time, it can mean that some physical machine is down that affects multiple Valkey nodes. Then, some replicas might be down too and then #2227 is not effective.

PR #1670, opened 2025-02-05
Enabling TLS Certificate rotation

kkishoreyadav:
This PR fixes an issue where TLS certificate rotation was not working correctly by properly setting last_modified_time in the Valkey code. With this fix, Valkey correctly detects certificate updates, ensuring that the latest TLS certificates are used as expected.

image

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -280,6 +280,18 @@ error:
     return NULL;
 }
 
+/* Given a path to a file, return the last time it was accessed (in seconds) */
+static time_t getLastModifiedTime(const char* path){
+    struct stat path_stat;
+    stat(path, &path_stat);
+
+#ifdef __APPLE__
+    return path_stat.st_mtimespec.tv_sec;
+#else
+    return path_stat.st_mtime;
+#endif
+}
+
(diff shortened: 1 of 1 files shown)


madolson:
Were you able to build and test this locally? It looks like you are adding some new fields but they are never getting checked.

Would you also mind explaining the issue in a little bit more detail?

hpatro:
Would be handy to showcase before and after. Also, would it be possible to add test case for the above. Otherwise, we are still at risk of regression in behavior.

ovaiskhansc:
> Would be handy to showcase before and after

@hpatro what kind of before and after comparison were you thinking?

hpatro:
> > Would be handy to showcase before and after
> 
> @hpatro what kind of before and after comparison were you thinking?

I'm not sure about the flow. Trying to understand, what was failing prior to this?

ovaiskhansc:
@kkishoreyadav this port seems to be missing quiet a bit of files. Let's either close this PR or add the missing commits here.

kkishoreyadav:
I realzied I missed some commits and will re-open the pr with the missed commits and comments from the team

Issue #1336, opened 2024-11-21
[NEW] Support database level ACL


hpatro:
**The problem/use-case that the feature addresses**

Extend ACL to include a way to restrict users to a specific database. This would make it possible to assign individual databases to a user, opening the possibiity for a single Valkey instance to be multi-tenant. 

**Description of the feature**

Enable ACL syntax to define DB level permissions that only allows a user to connect/select a specific database.

**Additional Information**

There is a separate ongoing discussion on support multiple DBs in Valkey cluster mode #1319. If we plan to support multiple DBs feature in cluster mode, we should rather make the guardrails stronger around it.

Past discussion: https://github.com/redis/redis/issues/8099

madolson:
If we ever do this, we need to remember that we need to figure out how to follow select logic in multi-exec. Today we check all the permissions before execution, which doesn't work since you can `SELECT` a different database inside a multi.

dvkashapov:
As I see it, there's two ways of accomplishing database level ACL:

1. Expand aclSelector to include list of allowed databases
That way we can check if there is access to a given database. It would be fairly straightforward implementation, and there shouldn't be a problem with MULTI/EXEC just because logic would be similar to channels (i.e. one of SELECT's access restricted db => NOPERM before exec)

2. Allow different ACLs for different databases
This seems like a better solution from usability standpoint, for example: user would be able to only write to one db and only read from another. In such case there will definitely be problems when implementing checks in MULTI/EXEC because user may have partial access to both db's, allowing user to start an EXEC with one ACL, do SELECT and access another database with other ACL (or even change ACL on the fly).

The second option seems better because it would allow to setup granular access. What do you think?

dmitrypol:
you could sort of hack it by restricting specific user to run `select 1` but I am sure there are holes in this approach.  But I 100% agree that Valkey multi-tenancy would be a very useful feature.

PR #1671, opened 2025-02-05
Add multi-database support to cluster mode

xbasel:
## cluster: add multi-database support in cluster mode

Add multi-database support in cluster mode to align with standalone mode
and facilitate migration. Previously, cluster mode was restricted to a
single database (DB0). This change allows multiple databases while
preserving the existing slot-based key distribution.

### Key Features:
- Database-Agnostic Hashing. The hashing algorithm is unchanged.
  Identical keys always map to the same slot across all databases,
  ensuring consistent key distribution and compatibility with
  existing single-database setups.
- Multi-DB commands support. SELECT, MOVE, and COPY are now supported in
  cluster mode.
- Fully backward compatible with no API changes.
- SWAPDB is not supported in cluster mode. It is unsafe due to inconsistency risks.

### Command-Level Changes:
- SELECT / MOVE / COPY are now supported in cluster mode.
- MOVE / COPY (with db) are rejected (TRYAGAIN error) during slot migration to prevent multi-DB inconsistencies.
- SWAPDB will return an error if used when cluster mode is enabled.
- GETKEYSINSLOT, COUNTKEYSINSLOT and MIGRATE will operate in the context of the selected database. 
  This means, for example, that migrating keys in a slot will require iterating and repeating across all databases. 

### Slot Migration Process:
- Multi-DB support in cluster mode affects slot migration. Operators should now iterate over all the configured databases.
 
### Transaction Handling (MULTI/EXEC):
- getNodeByQuery key lookup behavior changed:
  - No key lookups when queuing commands in MULTI, only cross-slot
    validation.
  - Key lookups happen at EXEC time.
  - SELECT inside MULTI/EXEC is now checked, ensuring key validation
    uses the selected DB at lookup.

### Valkey-cli:
- valkey-cli has been updated to support resharding across all databases.

### Configuration:
- Introduce new configuration `cluster-databases`.
  The new configuration controls the maximal number of databases in cluster mode.

Implements  https://github.com/valkey-io/valkey/issues/1319


src/db.c
```diff
@@ -1728,12 +1714,6 @@ void swapMainDbWithTempDb(serverDb *tempDb) {
 void swapdbCommand(client *c) {
     int id1, id2;
 
-    /* Not allowed in cluster mode: we have just DB 0 there. */
```
dvkashapov: Would that be enough for swapdb to work in cluster mode? What will happen in setup with 2 shards, each responsible for half of slots in db's?

xbasel: With this implementation SWAPDB must be executed in all primary nodes. There are three options:

1.   Allow SWAPDB and shift responsibility to the user – Risky, non-atomic, can cause temporary inconsistency and data corruption. Needs strong warnings.
2.   Keep SWAPDB disabled in cluster mode – Safest, avoids inconsistency.
3.    Make SWAPDB cluster-wide and atomic or  – Complex, unclear feasibility.

I think option 2 is the safest bet. @JoBeR007 wdyt?

zuiderkwast: Is SWAPDB replicated as a single command? Then it's atomic.

If it's risky, it's risky in standslone mode with replicas too, right?

I think we can allow it. Swapping the data can only be done in some non-realtime workloads anyway I think.

dvkashapov: I think _risky because of replication_ and _risky because of the need to execute SWAPDB on all primary nodes_ are unrelated just because as a user you can't control first, but user is the main risk in the second case.
I would keep SWAPDB disabled in cluster mode, if we decide to continue with this implementation

zuiderkwast: In cluster mode, consistency is per slot.

xbasel: > Is SWAPDB replicated as a single command? Then it's atomic.
> 
> If it's risky, it's risky in standslone mode with replicas too, right?
> 
> I think we can allow it. Swapping the data can only be done in some non-realtime workloads anyway I think.

I don’t think it’s very risky with standalone replicas. The only downside is if  SWAPDB propagation to the replica takes time, a client might still access the wrong database. At least the client won’t be able to modify the wrong database, as they can only read.
In cluster mode, the same (logical) DB can be DB0 on one node and DB1 on another, but similar issues already exist today, FLUSHDB on one node doesn’t clear the entire DB since data exists in other slots/nodes. But as you said, consistency is per slot.

zuiderkwast: Yes, FLUSHDB is very similar in this regard. If a failover happens just before this command has been propagated to replicas, it's a big thing, but it's no surprise I think. The client can use WAIT or check replication offset to make sure the FLUSHDB or SWAPDB was successful on the replicas.

soloestoy: Regarding this, I think it is not just an issue of Multi-database but is more related to atomic slot migration. If a shard is in a stable state (not undergoing slot migration), then `flushdb`/`flushall`/`swapdb` are safe. However, if slot migration is in progress, it might lead to data inconsistency.

I think this needs to be considered alongside atomic-slot-migration:
1. During the ATM process, for slots being migrated, if we encounter `flushall`/`flushdb`, we can send a command like `flushslot` or `flushslotall` to the target shard
2. As for `swapdb`, I recommend temporarily prohibiting execution during the ATM process

@PingXie @enjoy-binbin , please also take note of this.

src/cluster.c
```diff
@@ -1102,7 +1110,7 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
              * NODE <node-id>. */
             int flags = LOOKUP_NOTOUCH | LOOKUP_NOSTATS | LOOKUP_NONOTIFY | LOOKUP_NOEXPIRE;
             if ((migrating_slot || importing_slot) && !pubsubshard_included) {
-                if (lookupKeyReadWithFlags(&server.db[0], thiskey, flags) == NULL)
+                if (lookupKeyReadWithFlags(c->db, thiskey, flags) == NULL)
```
soloestoy: Here, I modified it to use `c->db`, so for most commands, the key it wants to access can be correctly located. However, some cross-DB commands, such as `COPY`, still require additional checks. The ultimate solution is atomic-slot-migration I believe. Once ATM is implemented, the `TRYAGAIN` issue will no longer occur.

soloestoy:
I'm happy that we did "Unified db rehash method for both standalone and cluster #12848" when developing kvstore , which made the implementation of multi-database simpler.

src/db.c
PingXie: make sense. @murphyjacob4 FYI

src/cluster_legacy.c
```diff
@@ -6881,6 +6878,15 @@ void clusterCommandSetSlot(client *c) {
...
 }
 
+int dbHasNoKeys(void) {
+    for (int i = 0; i < server.dbnum; i++) {
+        if (kvstoreSize(server.db[i].keys) != 0) {
+            return 0;
+        }
+    }
+    return 1;
+}
```
hpatro: I think we should move this to db.c

tests/support/cluster.tcl
```diff
@@ -196,7 +196,7 @@ proc ::valkey_cluster::__method__masternode_notfor_slot {id slot} {
     error "Slot $slot is everywhere"
 }
 
-proc ::valkey_cluster::__dispatch__ {id method args} {
+proc ::valkey_cluster::__dispatch__ {id method args} {
```
hpatro: Avoid.

hpatro (commented):
We need to add history to `SWAPDB`, `SELECT`, `MOVE` json files to indicate it's supported since 9.0.

tests/cluster/tests/05-cluster-multidatabases.tcl
```diff
@@ -0,0 +1,481 @@
+# Tests multi-databases in cluster mode
```
madolson: This is the legacy clustering system. Ideally this test should be in unit/cluster

tests/cluster/tests/05-cluster-multidatabases.tcl
```diff
@@ -0,0 +1,481 @@
...
+
+proc pause {{message "Hit Enter to continue ==> "}} {
```
madolson: I see you used https://github.com/valkey-io/valkey/blob/f85c93301ca63f070164b307c6a7371a9ab5ac7b/tests/support/util.tcl#L1214 for break points in other places in the code, make sure to clean this all up.

tests/unit/cluster/cli.tcl
```diff
@@ -434,6 +434,105 @@ start_server [list overrides [list cluster-enabled yes cluster-node-timeout 1 cl
...
+                    fail "For DB $db, key $key: expected '$expected', got '$actual'"
+                }
+            }
+        }
+    }
+}
+
 } ;# tags
 
 set ::singledb $old_singledb
```
madolson: Worth noting we no longer need this constraint, so we might want to consider removing this throughout the codebase except when we are running in external mode.

src/cluster.c
madolson: I noticed that getNodeByQuery doesn't follow selects either, so this might not be the right database. If you for example have:
```
SELECT 0
GET FOO
SELECT 1
GET FOO
```
c->db won't be correct here either. COPY and move are also such problems as mentioned. I wonder if there is some way to make this correct without having ATM so we can limit the breakage if you're moving from standalone to cluster.

src/db.c
madolson: I made a comment on the issue about this, but also worth mentioning it's hard to orchestrate `SWAPDB`. Even in steady state, `flushdb` and `flushall` are idempotent (you can send them multiple times) but swapdb isn't. If a command times out on one node, it's hard to reason about if it was successful and how to retry it. I think we should continue to disable `SWAPDB` in cluster mode for now, unless we introduce an idempotent way to do the swap.

tests/unit/lazyfree.tcl
```diff
@@ -1,5 +1,12 @@
 start_server {tags {"lazyfree"}} {
     test "UNLINK can reclaim memory in background" {
+
+        # The test framework invokes "flushall", replacing kvstores even if empty.
```
madolson: I would rather we did a sync flushall then in the test framework, so we don't have these random waits all over the place.

src/cluster.c
soloestoy: Generally, c->db can obtain the correct context information. Are you referring to the scenario where the select command is also used within a transaction (MULTI/EXEC)?

tests/unit/lazyfree.tcl
xbasel: The wait here is only to allow lazy free to complete and for used_memory to update. We don't need to sleep after using `FLUSHALL` in other tests.

Additionally, once https://github.com/valkey-io/valkey/pull/1609 is merged, it's unlikely that this sleep will be necessary

src/db.c
xbasel: Maybe introduce UUID tracking for SWAPDB requests works. 
disabling SWAPDB for now.

src/cluster.c
xbasel: I've noticed that `getNodeByQuery` is being invoked while queuing commands in a `MULTI` context. Is this intentional? It seems unnecessary to check for key existence before execution, as the database state can change and keys might be migrated. I would expect this check to happen when `EXEC` is executed instead. Any thoughts?

xbasel: The only benefit of this early validation, as I see it, is detecting cross-slot keys sooner. I think key existence validation should happen during EXEC execution.

xbasel: @madolson / @soloestoy 
Can you check https://github.com/valkey-io/valkey/pull/1671/commits/da1ee65a596847d192447ed0cea0a8c6cdc62512 ?

tests/unit/lazyfree.tcl
madolson: How are you guaranteeing we never need to wait for the `FLUSHALL` in other tests?

xbasel: I don’t. If `FLUSHALL` is just wiping databases, there’s no need to wait. The wait here is only for observing memory impact. Why do you think we need to wait every time `FLUSHALL` is called?

src/cluster.c
soloestoy: looks good, but COPY and MOVE still have the problem. A simple way is refuse these commands during slot migration, or we can wait atomic slot migration finished that we don't need to check the migrating status.

xbasel: Yes, the commit above does not address `COPY` and `MOVE`.  
I've merged a comment to reject these commands during slot migration.

src/db.c
```diff
@@ -2851,3 +2837,12 @@ int bitfieldGetKeys(struct serverCommand *cmd, robj **argv, int argc, getKeysRes
     }
     return 1;
 }
+
+int dbHasNoKeys(void) {
```
hwware: Valkey already agree to use bool.  Reference: https://github.com/orgs/valkey-io/discussions/1278
Thus for this kind of function, bool is much eaiser to understand than an integer return type.

xbasel: fixed https://github.com/valkey-io/valkey/pull/1671/commits/362b6596f92e06215bfac9dc04d6c02346ae2ac7

src/cluster.c
```diff
@@ -909,11 +917,11 @@ void clusterCommand(client *c) {
             return;
         }
 
-        unsigned int keys_in_slot = countKeysInSlot(slot);
+        unsigned int keys_in_slot = countKeysInSlotForDb(slot, c->db);
         unsigned int numkeys = maxkeys > keys_in_slot ? keys_in_slot : maxkeys;
         addReplyArrayLen(c, numkeys);
         kvstoreHashtableIterator *kvs_di = NULL;
-        kvs_di = kvstoreGetHashtableIterator(server.db->keys, slot, 0);
+        kvs_di = kvstoreGetHashtableIterator(c->db->keys, slot, 0);
```
hwware, line 927: This change I thought is not compatible with current user expectation. 
Now CLUSTER GETKEYSINSLOT <slot> <count> and CLUSTER COUNTKEYSINSLOT <slot> only return the value of db0. The changes return the sum of all db.
I think the better way is to add more parameters (such as db number) on CLUSTER GETKEYSINSLOT <slot> <count> and CLUSTER COUNTKEYSINSLOT <slot> to the specific db. And add one more cluster command to get all db keys.

Maybe need to discuss details

PingXie: @hwware we are discussing two commands at https://github.com/valkey-io/valkey/issues/1319#issuecomment-2665101547

tests/unit/lazyfree.tcl
madolson: I was thinking that other people might not be aware of this constraint, and might encounter similar issues to you where the memory is not behaving the way they expect. We actually recently made the change (`FLUSHALL` used to default to being sync until Valkey 8.0). So maybe we should be explicitly doing the `FLUSHALL SYNC` inside the test framework itself.

Also, now that I read this again, is this true? The cluster test framework invokes flushall, this test doesn't seem to invoke flushall at all. Is this still necessary?

src/cluster.c
xbasel: @hwware Worth noting that existing clients (working exclusively with DB0) would not be impacted. Since their clients already operate on DB0, both COUNTKEYSINSLOT and GETKEYSINSLOT will return the same results whether iterating over all databases or reading only from DB0.

Only clients that choose to use multiple databases may need to make adjustments

I also agree that these commands can be further enhanced by adding optional parameters. Please see my comment here:
https://github.com/valkey-io/valkey/issues/1319#issuecomment-2687727943

hwware: Got it, Thanks

xbasel:
documentation: https://github.com/valkey-io/valkey-doc/pull/242

hwware:
It looks like there are still some test cases failed related to multiply db feature. Please fix them first, Thanks

xbasel:
> It looks like there are still some test cases failed related to multiply db feature. Please fix them first, Thanks

Ready

src/valkey-cli.c
```diff
@@ -4801,8 +4801,10 @@ static redisReply *clusterManagerMigrateKeysInReply(clusterManagerNode *source,
     argv_len = zcalloc(argc * sizeof(size_t));
     char portstr[255];
     char timeoutstr[255];
+    char dbnum[255];
     snprintf(portstr, 10, "%d", target->port);
     snprintf(timeoutstr, 10, "%d", timeout);
+    snprintf(dbnum, 10, "%d", config.dbnum);
```
zuiderkwast: It is quite weird that we use buffers of length 255 but we limit to 10 bytes including nul-terminator in snprintf. We can use `char dbnum[10]` instead of 255 and the same for the others.

src/cluster.c
xbasel: **MOVE**: Since the destination key must not exist for MOVE to succeed, we cannot determine whether it was already migrated (and deleted). Therefore, it is considered a missing key and should be rejected while the slot is being migrated.

**COPY**: In a cross-DB context (i.e., when the db parameter is provided), both the source and destination keys must exist; otherwise, the engine assumes any missing keys were migrated. To keep the logic simple, COPY is always rejected in a cross-DB context during slot migration, regardless of keys existence, but remains allowed within the selected DB.

zuiderkwast: > To keep the logic simple, COPY is always rejected in a cross-DB context, regardless of keys existence, but remains allowed within the selected DB.

Always during slot migration or always always?

xbasel: Only during slot migration. Updated the comment.

Once this PR is merged, I'll try to refactor getNodeByQuery to allow COPY during slot migration (if both keys exist) without making it look like it's parsing commands and partially executing them.

xbasel:
@soloestoy / @zuiderkwast / @madolson Just checking - do you have any remaining comments, or can you approve the PR as is?

hwware:
Can you sign the DCO please?  Here https://github.com/valkey-io/valkey/pull/1671/checks?check_run_id=39005472844

src/valkey-cli.c
zuiderkwast:
```suggestion
    char portstr[10];
    char timeoutstr[10];
    char dbnum[10];
    snprintf(portstr, sizeof(portstr), "%d", target->port);
    snprintf(timeoutstr, sizeof(timeoutstr), "%d", timeout);
    snprintf(dbnum, sizeof(dbnum), "%d", config.dbnum);
```

xbasel:
> Can you sign the DCO please? Here https://github.com/valkey-io/valkey/pull/1671/checks?check_run_id=39005472844

@hwware 
There are multiple commits without a sign-off. If I rebase or change history, existing PR comments will disappear. Since this PR will be squashed on merge, the final commit will include a proper sign-off line. Does that work for DCO compliance?

zuiderkwast:
> If I rebase or change history, existing PR comments will disappear. Since this PR will be squashed on merge, the final commit will include a proper sign-off line. Does that work for DCO compliance?

Maybe it works but we usually want the commits in the PR to be signed off anyway. If you do `git rebase HEAD~35 --signoff` and force push, I don't think the PR comments will disappear.

xbasel:
> > If I rebase or change history, existing PR comments will disappear. Since this PR will be squashed on merge, the final commit will include a proper sign-off line. Does that work for DCO compliance?
> 
> Maybe it works but we usually want the commits in the PR to be signed off anyway. If you do `git rebase HEAD~35 --signoff` and force push, I don't think the PR comments will disappear.

I'd rather do this just before a reviewer is ready to approve in order to not risk losing comment history :), are you ready for approval ? (merge after GA)

zuiderkwast:
> I'd rather do this just before a reviewer is ready to approve in order to not risk losing comment history :)

I recommend doing it ASAP, but at least before `git merge unstable`. If you merge unstable first, the signoff will be more difficult.

> are you ready for approval ? (merge after GA)

No, none of the reviewers have approved it yet.

zuiderkwast:
Now, you just did the exact opposite of what I recommended. Now it will be harder to fix the signoff.

xbasel:
> It looks like there are still some test cases failed related to multiply db feature. Please fix them first, Thanks

All tests pass now.

hwware (approved):

hwware:
@valkey-io/core-team Do we have any concern for this PR? If the only thing we need to do is waiting for merge?

madolson:
I would still like to review it again to make sure it's not breaking any other assumptions. Other might as well.

mcollina:
What changes will be needed in clients?

xbasel:
> What changes will be needed in clients?
@mcollina 
Only slot migration changes if `valkey-cli` is not used.  No API changes. The only change is that now admins need to iterator over databases to migrate keys as `GETKEYSINSLOT` works on the selected database:
```
Iterate through all databases:
    SELECT database
    keys = GETKEYSINSLOT
    MIGRATE source target keys
```

avifenesh:
> > What changes will be needed in clients?
> @mcollina 
> Only slot migration changes if `valkey-cli` is not used.  No API changes. The only change is that now admins need to iterator over databases to migrate keys as `GETKEYSINSLOT` works on the selected database:
> ```
> Iterate through all databases:
>     SELECT database
>     keys = GETKEYSINSLOT
>     MIGRATE source target keys
> ```
> 
> 

Administration is out of clients scope, they are part of operations, not application logic.
Do i miss something?

xbasel:
> > > What changes will be needed in clients?
> > > @mcollina
> > > Only slot migration changes if `valkey-cli` is not used.  No API changes. The only change is that now admins need to iterator over databases to migrate keys as `GETKEYSINSLOT` works on the selected database:
> > 
> > 
> > ```
> > Iterate through all databases:
> >     SELECT database
> >     keys = GETKEYSINSLOT
> >     MIGRATE source target keys
> > ```
> 
> Administration is out of clients scope, they are part of operations, not application logic. Do i miss something?

Admins are still clients in this context - they use Valkey commands too, even if it's for ops, not app logic.

hwware:
> > What changes will be needed in clients?
> > @mcollina
> > Only slot migration changes if `valkey-cli` is not used.  No API changes. The only change is that now admins need to iterator over databases to migrate keys as `GETKEYSINSLOT` works on the selected database:
> 
> ```
> Iterate through all databases:
>     SELECT database
>     keys = GETKEYSINSLOT
>     MIGRATE source target keys
> ```

From my understanding, to the client, there is no any change. Because by default, there is only one database number so far.

xbasel:
> > > What changes will be needed in clients?
> > > @mcollina
> > > Only slot migration changes if `valkey-cli` is not used.  No API changes. The only change is that now admins need to iterator over databases to migrate keys as `GETKEYSINSLOT` works on the selected database:
> > 
> > ```
> > Iterate through all databases:
> >     SELECT database
> >     keys = GETKEYSINSLOT
> >     MIGRATE source target keys
> > ```
> 
> From my understanding, to the client, there is no any change. Because by default, there is only one database number so far. 

This is correct. Only clients using multiple DBs will need to iterate with 'select'.

JimB123:
Have we reasoned about ACL impacts?  Can ACLs protect specific DBs?

In scripts, if keys are declared, there is no DB indication.  If a script is permitted to perform SELECT, there is no mechanism to know which DBs the keys will be in.  (Also, last I checked, scripts aren't required to declare keys.  It's optional.  I'm assuming this is still true?)

Of course this is all the case for non-cluster.  Just making sure we aren't missing any got-chas in this area.

zuiderkwast:
> Have we reasoned about ACL impacts? Can ACLs protect specific DBs?

We have talked about it. No, ACLs can't protect specific DBs currently. We might add that in the future, but currently it's not possible so there is no problem.

hpatro:
> Of course this is all the case for non-cluster. Just making sure we aren't missing any got-chas in this area.

Exactly this is the reasoning to proceed without ACL support as of now. I had created the issue https://github.com/valkey-io/valkey/issues/1336 sometime back and seems like folks want it.

soloestoy (approved):

soloestoy:
>#### Key Features
>Lazy Database Initialization: Databases are created on demand, reducing unnecessary memory usage.

what does that mean?

xbasel:
> >#### Key Features
> >Lazy Database Initialization: Databases are created on demand, reducing unnecessary memory usage.
> 
> what does that mean?

The current code pre-allocates databases, with cluster mode memory is amplified due to the number of slots.
See this PR, which addresses this issue, and will be merged after this pr is merged:


https://github.com/valkey-io/valkey/pull/1609

soloestoy:
> > > #### Key Features
> > > Lazy Database Initialization: Databases are created on demand, reducing unnecessary memory usage.
> > 
> > 
> > what does that mean?
> 
> The current code pre-allocates databases, with cluster mode memory is amplified due to the number of slots. See this PR, which addresses this issue, and will be merged after this pr is merged:
> 
> #1609

OK, we can mention this feature on that correct PR, this PR doesn't contain it.

valkey.conf
```diff
@@ -442,6 +442,7 @@ logfile ""
 # a different one on a per-connection basis using SELECT <dbid> where
 # dbid is a number between 0 and 'databases'-1
 databases 16
+cluster-databases 1
```
zuiderkwast: We should explain that `databases` is ignored in cluster mode and vice versa.

Shall we put `cluster-databases` here or next to the other cluster configs?

For most of the configs, we write the config with default only in a comment. Here it's written without a comment to match `databases 16` on the line above?

I suggest we move `cluster-databases` to the cluster config section and add a note about it here. I don't have a strong opinion though.

```suggestion
# Note that 'databases' does not affect cluster mode. In cluster mode,
# use the 'cluster-databases' directive instead.
databases 16
```

And for the cluster config section, here is a suggestion:

```
# Set the number of databases in cluster mode. This is the cluster counterpart
# of the 'databases' directive. In Valkey 9.0 and later, multiple databases are
# permitted in cluster mode and the SELECT command can be used to switch between
# them. The default number of databases in cluster mode is 1.
#
# cluster-databases 1
```

zuiderkwast (commented):
Looks great. Disclaimer: I didn't review the tests carefully.

I have only one comment about the comments about the new config.

valkey.conf
xbasel: I grouped them for visibility since they’re related, but I agree it fits better under the cluster section. A note solves this. I moved it to cluster section and added a comment to `databases` to make it clear that the config is ignore in cluster. Also commented out the value as the default is 1, it was only uncommented to match databases.

Thanks for the suggestions!
 
The change: https://github.com/valkey-io/valkey/pull/1671/commits/939d059546deaf7e41b5bf7b0b192dbe9e5b7df5

tests/unit/cluster/slot-migration.tcl
```diff
@@ -251,7 +258,7 @@ start_cluster 3 3 {tags {external:skip cluster} overrides {cluster-allow-replica
 proc create_empty_shard {p r} {
     set node_timeout [lindex [R 0 CONFIG GET cluster-node-timeout] 1]
     assert_equal {OK} [R $p CLUSTER RESET]
-    assert_equal {OK} [R $r CLUSTER RESET]
+    assert_equal {OK} [R $r CLUSTER RESET]0
```
zuiderkwast: Typo

```suggestion
    assert_equal {OK} [R $r CLUSTER RESET]
```

zuiderkwast (approved):
LGTM (except a typo)

zuiderkwast:
> What changes will be needed in clients?

@mcollina To support this feature, the client lib should just allow SELECT, MOVE and COPY to be used in cluster mode, if it doesn't already allow this.

> > Administration is out of clients scope, they are part of operations, not application logic. Do i miss something?
> 
> Admins are still clients in this context - they use Valkey commands too, even if it's for ops, not app logic.

@avifenesh @xbasel I think what @mcollina means by "client" here is "client library" such as iovalkey, valkey-go, etc.

avifenesh:
> > What changes will be needed in clients?
> 
> @mcollina To support this feature, the client lib should just allow SELECT, MOVE and COPY to be used in cluster mode, if it doesn't already allow this.
> 
> > > Administration is out of clients scope, they are part of operations, not application logic. Do i miss something?
> > 
> > 
> > Admins are still clients in this context - they use Valkey commands too, even if it's for ops, not app logic.
> 
> @avifenesh @xbasel I think what @mcollina means by "client" here is "client library" such as iovalkey, valkey-go, etc.

If this is the intent, so I don't understand :)
Might miss something, but I've yet to see a client library built in a way to help "manage" the cluster.
There are ops, and there's application, and the logic is separated as far as I saw. It can be a very fascinating idea to mix them, but I don't understand what the client needs to support over what clients already support.

xbasel:
> 
> If this is the intent, so I don't understand :) Might miss something, but I've yet to see a client library built in a way to help "manage" the cluster. There are ops, and there's application, and the logic is separated as far as I saw. It can be a very fascinating idea to mix them, but I don't understand what the client needs to support over what clients already support.

@avifenesh 

I don't think client libraries handle any cluster management logic  like resharding, slot migration, etc. That’s ops territory.

Worth checking if client libraries silently reject commands like `SELECT` or `MOVE` while operating in cluster mode. Maybe some of them block these at the client level without even sending them to the server.

dmitrypol:
this would be a very useful feature

tests/unit/cluster/multidb.tcl
```diff
@@ -0,0 +1,401 @@
+# Tests multi-databases in cluster mode
+
+# source "../tests/includes/init-tests.tcl"
```
madolson:
```suggestion
```
Presumably this is some left over code from when the tests were in the old framework?

tests/support/cluster_util.tcl
```diff
@@ -256,7 +256,8 @@ proc start_cluster {masters replicas options code {slot_allocator continuous_slo
 
     # Configure the starting of multiple servers. Set cluster node timeout
     # aggressively since many tests depend on ping/pong messages. 
-    set cluster_options [list overrides [list cluster-enabled yes cluster-ping-interval 100 cluster-node-timeout 3000 cluster-slot-stats-enabled yes]]
+
+    set cluster_options [list overrides [list cluster-enabled yes cluster-ping-interval 100 cluster-node-timeout 3000 cluster-databases 16 cluster-slot-stats-enabled yes]]
```
madolson, line 260: A few lines below this we set `::singledb` to 1. Normally we test on a random DB, which has helped us catch some random regressions in the past, I think it would be helpful if we can remove that to improve coverage. I'm okay skipping that for now, but would like to follow up and add it.

tests/unit/cluster/multidb.tcl
```diff
@@ -0,0 +1,401 @@
...
+        set remaining_keys [R $primary_id cluster countkeysinslot $slot]
+        assert_equal $remaining_keys 0
+    }
+
+
+
+    test "Replication: Write to multiple databases and verify replica" {
+        set primary_id 0    
+
+        set replica [get_my_replica [R $primary_id cluster nodes]]
```
madolson:
```suggestion
        set replica [get_my_replica [R $primary_id cluster replicas]]    
```
Minor thing, but there is a function that just returns the replicas of a given node.

tests/unit/cluster/multidb.tcl
```diff
@@ -0,0 +1,401 @@
...
+            R $primary_id select $db
+            for {set i 0} {$i < $keys_per_db} {incr i} {
+                set key "key$i"
+                R $primary_id del $key
+            }
+            assert_equal [R $primary_id dbsize] 0
+        }
+
+        # Wait for deletion to replicate
+        after 500
```
madolson:
```suggestion
        # Wait for deletion to replicate
        assert_equal 1 [R $primary_id wait 1 5000]
```
500ms isn't a lot for a timing based wait conditional, there is a few waits here that could just be done with removed with just a wait command.

tests/unit/cluster/multidb.tcl
```diff
@@ -0,0 +1,401 @@
...
+                set key "${db}key$i"
+                set val "value_db${db}_key$i"
+                R $primary_id set $key $val
+                assert_equal [R $primary_id get $key] $val
+            }
+            assert_equal [R $primary_id dbsize] $keys_per_db
+        }
+
+        # Run BGSAVE to save the RDB
+        R $primary_id save
```
madolson: This isn't a bgsave, this is a foreground save that blocks the main process.
```suggestion
        # Run save to save the RDB
        R $primary_id save
```

tests/unit/cluster/multidb.tcl
```diff
@@ -0,0 +1,401 @@
...
+        
+        # Restart instance
+        
+        restart_server $primary_id true false
+        # kill_instance valkey $primary_id
+        # restart_instance valkey $primary_id
```
madolson:
```suggestion
```
More dead code?

src/db.c
```diff
@@ -2851,3 +2837,12 @@ int bitfieldGetKeys(struct serverCommand *cmd, robj **argv, int argc, getKeysRes
...
     return 1;
 }
+
+bool dbHasNoKeys(void) {
+    for (int i = 0; i < server.dbnum; i++) {
+        if (kvstoreSize(server.db[i].keys) != 0) {
+            return 0;
+        }
+    }
+    return 1;
```
madolson:
```suggestion
            return false;
        }
    }
    return true;
```

src/cluster.c
```diff
@@ -1048,6 +1058,15 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
         numkeys = getKeysFromCommand(mcmd, margv, margc, &result);
         keyindex = result.keys;
 
+        if (mcmd->proc == selectCommand) {
+            long long id;
+            if (getLongLongFromObject(margv[i], &id) != C_OK || (id < 0 || id >= server.dbnum)) {
+                if (error_code) *error_code = CLUSTER_REDIR_UNSTABLE;
+                return NULL;
+            }
+            currentDb = server.db + (int)id;
```
madolson:
```suggestion
            currentDb = server.db[(int) id];
```

src/cluster.c
```diff
@@ -1048,6 +1058,15 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
         numkeys = getKeysFromCommand(mcmd, margv, margc, &result);
         keyindex = result.keys;
 
+        if (mcmd->proc == selectCommand) {
+            long long id;
+            if (getLongLongFromObject(margv[i], &id) != C_OK || (id < 0 || id >= server.dbnum)) {
```
madolson: This isn't complete verification the command would succeed, this command could fail for other reasons like ACLs.

src/cluster.c
```diff
@@ -1097,15 +1117,38 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
...
+            /* Block MOVE command as the destination key is not expected to exist, and we don't know if it was migrated */
+            if ((migrating_slot || importing_slot) && mcmd->proc == moveCommand) {
+                if (error_code) *error_code = CLUSTER_REDIR_UNSTABLE;
+                getKeysFreeResult(&result);
+                return NULL;
+            }
+
+            /* Block the COPY command if it's cross-DB to keep the code simple.
+             * Allowing cross-DB COPY is possible, but it would require looking up the second key in the target DB.
+             * The command should only be allowed if the key exists. We may revisit this decision in the future. */
```
madolson, line 1130: Why not just block all cross-DB operations if we are going to block this one? I hadn't really considered that before this moment. The main ask was basically simple multi-tenancy, not doing material data manipulation across DBs. If we disallow selects in multi, move, rename (with a DB) and copy (with a DB), we're left with a smaller surface area for the getNodeByQuery. That removes a lot of this code off the hot path since we are not longer required to do a lot of validation to know if the command can be executed or not.

src/cluster.c
```diff
@@ -1048,6 +1058,15 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
         numkeys = getKeysFromCommand(mcmd, margv, margc, &result);
         keyindex = result.keys;
 
+        if (mcmd->proc == selectCommand) {
+            long long id;
+            if (getLongLongFromObject(margv[i], &id) != C_OK || (id < 0 || id >= server.dbnum)) {
+                if (error_code) *error_code = CLUSTER_REDIR_UNSTABLE;
```
madolson: Why are we returning unstable (aka tryagain), shouldn't we return ourself and less the normal error processing run through.

madolson (commented):
Mostly minor comments, but also one question about if we should simplify the API a little bit by not allowing any crossDB operations to remove the code getNodeByQuery. That codepath is very hot, basically all cluster requests go through it, so maybe we can solve most of the requests by being simple.

src/cluster.c
zuiderkwast: The main benefit is to make it easy to migrate from standalone to cluster. If some operations are missing, then the obstacle still exists.

I agree we should optimize for the hot code path, but I think we should try to support the cross-DB operations too. Otherwise it's another subtle difference between standalone and cluster.

src/cluster.c
xbasel: It's easier if use a pointer here.

src/cluster.c
xbasel: We need the database from the command. I don't know if we should check ACLs here. We use the database to lookup the key, not to actually execute the command. I'll use `selectDb()` to temporarily change the db (which runs the checks, excluding ACL).

src/cluster.c
xbasel: > Why are we returning unstable (aka tryagain), shouldn't we return ourself and less the normal error processing run through.

I don't think we should do this, we can be a in multi/exec, with missing keys in the  previous commands.

src/cluster.c
xbasel: `MOVE` -> blocked
`COPY` -> this code blocks it only if it's cross-DB. Not sure the code around `COPY` can be simplified.
`RENAME` -> is not a cross-DB command.

We allow `SELECT` in multi-exec. The current code handles this case, although it might make sense to block it and keep the code simple for now. What do you think?

tests/support/cluster_util.tcl
xbasel: I'll skip this for now and worry about it separately.

src/cluster.c
```diff
@@ -1048,6 +1058,19 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
         numkeys = getKeysFromCommand(mcmd, margv, margc, &result);
         keyindex = result.keys;
 
+        if (mcmd->proc == selectCommand) {
+            serverDb *origDb = currentDb;
+            long long id;
+            if (getLongLongFromObject(margv[i], &id) != C_OK || selectDb(c, (int)id) != C_OK) {
```
ranshid: I am not sure I understand why it is O.K to access margv[i]? the db number is always the first argument right?

src/cluster.c
ranshid: I also did not understand this. If I read correctly, this is only in case the select command is being run and the provided db num is incorrectly parsed or out of range. Why would retry ever solve such cases?

src/cluster.c
ranshid: I have reduced concern blocking these commands during legacy slot migration, since users migrating from standalone would also (hopefully) make use of the new atomic slot migration API which will not block these commands.
Regarding the code path - I am not sure why @madolson you think it is that complicated?

src/cluster.c
xbasel: > I am not sure I understand why it is O.K to access margv[i]? the db number is always the first argument right?

Fixed..

src/cluster.c
xbasel: On second thought, I think a failed **SELECT** in a **MULTI/EXEC** block should be ignored, since its failure doesn't impact the selected database state. Modified the code and tests.

src/cluster.c
madolson: > Regarding the code path - I am not sure why @madolson you think it is that complicated?

It's just code on the hot path of getNodeByQuery, which we should probably avoid as much as possible. Every check hurts branch prediction. It's not a lot but we should try to avoid it. 

None of this is a deal breaker for me though.

tests/support/cluster_util.tcl
madolson: https://github.com/valkey-io/valkey/issues/2021

madolson (approved):
I'm good with the current iteration.

src/cluster.c
```diff
@@ -1097,15 +1119,37 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
...
              * If it is pubsubshard command, it isn't required to check
              * the channel being present or not in the node during the
              * slot migration, the channel will be served from the source
              * node until the migration completes with CLUSTER SETSLOT <slot>
              * NODE <node-id>. */
             int flags = LOOKUP_NOTOUCH | LOOKUP_NOSTATS | LOOKUP_NONOTIFY | LOOKUP_NOEXPIRE;
-            if ((migrating_slot || importing_slot) && !pubsubshard_included) {
-                if (lookupKeyReadWithFlags(&server.db[0], thiskey, flags) == NULL)
+            if ((!c->flag.multi || (c->flag.multi && cmd->proc == execCommand)) && // Multi/Exec validation happens on exec
+                (migrating_slot || importing_slot) && !pubsubshard_included) {
```
ranshid:
```suggestion
            if ((migrating_slot || importing_slot) && !pubsubshard_included) &&
            (!c->flag.multi || (c->flag.multi && cmd->proc == execCommand))) {
```

src/cluster.c
```diff
@@ -1097,15 +1119,37 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
...
+                    return NULL;
+                }
+            }
+
             /* Migrating / Importing slot? Count keys we don't have.
              * If it is pubsubshard command, it isn't required to check
              * the channel being present or not in the node during the
              * slot migration, the channel will be served from the source
              * node until the migration completes with CLUSTER SETSLOT <slot>
              * NODE <node-id>. */
```
ranshid:
```suggestion
            /* Migrating / Importing slot? During exec we count keys we don't have.
             * If it is pubsubshard command, it isn't required to check
             * the channel being present or not in the node during the
             * slot migration, the channel will be served from the source
             * node until the migration completes with CLUSTER SETSLOT <slot>
             * NODE <node-id>. */
```

src/cluster.c
```diff
@@ -1048,6 +1058,17 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
         numkeys = getKeysFromCommand(mcmd, margv, margc, &result);
         keyindex = result.keys;
 
+        if (mcmd->proc == selectCommand) {
+            /* SELECT inside MULTI/EXEC is detected here. If it fails, we ignore it since it has no real impact at
+             * execution time. Even if a command fails, the remaining commands will still execute. */
```
ranshid: I am not sure the comment is correct. select can still be called via Module calling select via VM_call or select is called via script right? Maybe we should also check the mstate has count larger than 1?

xbasel: You're right the comment isn't accurate. 
Not sure I understand why we need to check if mstate.count>1 ..

madolson: I think Ran was calling out that we also have problems with selects happening from a module command or a script since they will reference the *wrong* keys for the later migrating and importing checks. We don't have a way to truly know which DB the keys will access though. We intend to solve this with atomic slot migration though, since we don't need to know where the keys are.

madolson:
I created issues for everything I noticed was a followup, wanted to merge to allow other PRs like atomic slot migration to rebase.

src/cluster.c
ranshid: >Not sure I understand why we need to check if mstate.count>1

My point here was that we can just avoid this handling for cases where there is no transaction or a transaction containing only `SELECT` (whatever that one is good for). do not think there was a blocker here though.

zuiderkwast:
Almost all tests in Daily failed today. What happened here?

    SELECT 1 failed: ERR DB index is out of range

https://github.com/valkey-io/valkey/actions/runs/14826557992

Issue #1649, opened 2025-01-30
[NEW] Prevent endless replication loop by throttling write commands


lschmidtcavalcante-sc:
**The problem/use-case that the feature addresses**

The problem of endless replication loop is described here https://redis.io/blog/the-endless-redis-replication-loop-what-why-and-how-to-solve-it/, but it can be summarized as: in cases where the replica is slow to consume the replication buffer (for example, busy retrieving the initial snapshot of the database), said buffer can overflow triggering the synchronization process to start from scratch again.

**Description of the feature**

Implement a feature flag (say `write-throttling`) that once enabled evaluates each incoming write request:

- If the incoming write request will cause an overflow of the replication buffer, then drop it and return `THROTTLED` error;
- If the incoming write request will not cause an overflow of the replication buffer, then add to the replication buffer and proceed as usual.

This way the endless replication loop is avoided at the cost of dropping some requests.

**Alternatives you've considered**

There are currently two solutions to endless replication loop:

1. Reduce the traffic so the buffer doesn't overflow (either cut the traffic or wait for non-peak hours)
2. Increase the replication buffer size, giving more time for the replica to catch up.

These solutions will continue to exist and the proposed `write-throttling` is not expected to be enabled by default.

hpatro:
@lschmidtcavalante-sc Thanks for starting this issue. I think operationally it makes sense to introduce a mechanism to avoid the endless replication loop. I've a few questions on this.

* Do you envision this as a boolean feature flag?
* Could it cause complete write outage? (this might be a bad/unexpected experience for unfamiliar users)
* Can't the client be blocked until the full sync is over instead of introducing a new `THROTTLED` response. Clients are generally very slow to catchup with these changes.

lschmidtcavalcante-sc:
1. Yes, it would be a boolean flag
2. I can only envision this scenario happening if there is a slow replica that doesn't consume its replication buffer and it also never disconnects. Nonetheless, I don't expect this feature to be enabled by default.
3. By client you mean the caller of Valkey? If so, then this solution is good because it doesn't rely on their behavior to `THROTTLED` (or another error stating that we are close to being full), we are simply not going to write their requests to the replication buffer regardless of their retries

ovaiskhansc:
@hpatro just wanted to add to point # 2 above that we use this functionality when ValKey is used as a cache and in that context a write failure is preferred instead of a OOM crash.

hpatro:
[...]

PR #1672, opened 2025-02-05
Implementation of write throttling

lschmidtcavalcante-sc:
This feature (config `write-throttling`) aims to give an alternative to client disconnection when the output buffer overflows. Once enabled, it evaluates if the incoming write command will exceed the output buffer limit of any replica: if it will exceed, then discard the write command and return `THROTTLED`; otherwise write it to the output buffer.

The expected main usage of this feature is during replication: while the replica is busy copying the initial snapshot its output buffer may overflow, causing a restart of the replication process. At the cost of dropping some requests (responding as `THROTTLED`) an endless replication loop can be averted.

Issue #1649


zuiderkwast:
We will make the cut-off for 8.1 release candidates in a few days, so this will most likely have to wait until the next release. Other than review, we'll need to make some decision about the feature and about throttling in general.

> The expected main usage of this feature is during replication: while the replica is busy copying the initial snapshot its output buffer may overflow, causing a restart of the replication process.

Regarding this problem, we already have dual-channel replication feature introduced in 8.0 which offloads the buffering to the replica side. I think it can help to some extent.

src/server.c
```diff
@@ -2059,6 +2060,7 @@ void createSharedObjects(void) {
         createObject(OBJ_STRING, sdsnew("-EXECABORT Transaction discarded because of previous errors.\r\n"));
     shared.noreplicaserr = createObject(OBJ_STRING, sdsnew("-NOREPLICAS Not enough good replicas to write.\r\n"));
     shared.busykeyerr = createObject(OBJ_STRING, sdsnew("-BUSYKEY Target key name already exists.\r\n"));
+    shared.throttlederr = createObject(OBJ_STRING, sdsnew("-THROTTLED\r\n"));
```
hpatro, line 2057: Sending a generic throttled error seems risky to let each client build their own retry logic. Either we need to send out an estimated replication duration as retry-after time or maintain a backoff period for each client and reset it once the replication backlog gets cleared.

src/server.c
```diff
@@ -4324,6 +4326,33 @@ int processCommand(client *c) {
...
+     * disconnection.
+     */
+    if (server.write_throttling && is_write_command && server.primary_host == NULL) {
+        /* Size of the command to be stored in the output buffer. */
+        unsigned long long command_size = 3 + digits10(c->argc); /* 3: *[argc]\r\n */
+        for (int iarg = 0; iarg < c->argc; ++iarg) {
+            size_t arg_length = sdslen(c->argv[iarg]->ptr);
+            command_size += 5 + digits10(arg_length) + arg_length; /* 5: $[arglen]\r\n[arg]\r\n */
+        }
+
```
hpatro: Wouldn't `c->net_input_bytes_curr_cmd` suffice to get the command size?

hpatro (commented):
Thanks for the PR. If TSC approves it, we would also need to add CLIENT opt in flag to enable this feature for client, that can handle the feature..

src/server.c
lschmidtcavalcante-sc: That is much better, thanks!

src/server.c
xbasel: Or block the client with a configurable timeout, then either resume writing if COB pressure subsides or return an error.

xbasel:
I'm not sure this approach is ideal for multi-replica setups. A single problematic replica could render the engine unusable, even when losing one replica might be acceptable.

ovaiskhansc:
@xbasel just trying to understand what you mean here. This error is primarily to protect the "primary" node from slowing down when it is under stress for caching scenarios where we care about timely response more than write success.

There is also a related discussion here: https://github.com/valkey-io/valkey/issues/1649

madolson:
> This error is primarily to protect the "primary" node from slowing down when it is under stress for caching scenarios where we care about timely response more than write success.

If the replica is overloaded, it won't be able to consume the output stream generated by the primary and it will build up in the replica's client output buffer. So the primary will start throttling writes even though it's not the one under pressure, the replica is.

src/server.h
```diff
@@ -1799,6 +1799,8 @@ struct valkeyServer {
                                                  invocation of the event loop. */
     unsigned int max_new_conns_per_cycle;     /* The maximum number of tcp connections that will be accepted during each
                                                  invocation of the event loop. */
+    int write_throttling;                     /* 1 if write throttling is enabled: prevents output buffer overflow (and
```
hwware:
```suggestion
    int write_throttling;                     /* 1 if write throttling is enabled: prevents replication buffer overflow (and
```

hwware: I am interested in this topic, so far it looks like the solution is throwing an error code.

lschmidtcavalcante-sc:
Sorry for the long absence.

It is a fair callout @xbasel , and @ovaiskhansc reply needs to take into account the context that we are coming from a KeyDB world that doesn't have a dual-channel equivalent -- in Valkey world the motivation changes a bit. So when @ovaiskhansc mentions to protect the primary, he means the extra work of re-starting the whole process of synchronization.

PR #1673, opened 2025-02-05
Allow TCL 9.0 for tests

zuiderkwast:
Makes our tests possible to run with TCL 9.

The latest Fedora now has TCL 9.0 and it's working now, including the TCL TLS package. (This wasn't working earlier due to some packaging errors for TCL packages in Fedora, which have been fixed now.)

This PR also removes the custom compilation of TCL 8 used in our Daily jobs and uses the system default TCL version instead. The TCL version depends on the OS. For the latest Fedora, you get 9.0, for macOS you get 8.5 and for most other OSes you get 8.6.

The checks for TCL 8.7 are removed, because 8.7 doesn't exist. It was never released.


zuiderkwast:
Fedora now comes with TCL 9.0 and we started getting failures like this: https://github.com/valkey-io/valkey/actions/runs/13147311114/job/36688160501

hwware:
Which version is used in our online CI?

zuiderkwast:
> Which version is used in our online CI?

Most of them have 8.6. Macos has 8.5. Fedora rawhide has now only 9.0. I guess we need to support all of them.

zuiderkwast:
TCL 9.0 works now. The test-fedorarawhide-jemalloc job passed.

The remaining failures are other problems. One of them is solved in #1675.

zuiderkwast:
:scream: Still fails with TLS:

```
version conflict for package "tcl": have 9.0.0, need 8.4
```

Need to solve this too.

zuiderkwast:
:sweat: The TCL TLS package doesn't seem to support TCL 9.0 yet. See here: https://core.tcl-lang.org/tcltls/wiki/Documentation

There's a beta version of this package 2.0 which will support TCL 9.0 according to this page: https://core.tcl-lang.org/tcltls/file?name=doc/tls.html&ci=trunk

Currently, fedora rawhide includes tcltls 1.7 which is incompatible with tcl 9.0 and effectively broken. I guess they will fix it soon.

Alternatives to this PR:

* Download and build TCL 8.6, without using fedora's package repos.
* Remove Fedora rawhide from our daily scope. It is the current development verions, like our unstable branch. Maybe we shouldn't actually run it in daily?

enjoy-binbin:
> Remove Fedora rawhide from our daily scope. It is the current development verions, like our unstable branch. Maybe we shouldn't actually run it in daily?

i think we can go with this option.

tests/instances.tcl
```diff
@@ -294,7 +292,7 @@ proc parse_options {} {
             incr j
             set ::host ${val}
         } elseif {$opt eq {--tls} || $opt eq {--tls-module}} {
-            package require tls 1.6
```
hwware, line 296: I think we should not touch tls lib

zuiderkwast: tls 1.6 requires TCL 8.x and it rejects TCL 9.x.

If we want to use TCL 9.0, we need to use tls 2.0, which is only a beta version now. It's not yet released.

If we want to allow different TCL versions, we also need different TLS lib versions, so I think we can remove the hard-coded version requirements.

madolson:
> Remove Fedora rawhide from our daily scope. It is the current development verions, like our unstable branch. Maybe we shouldn't actually run it in daily?

I think having it here is good as it notifies us about upcoming issues, but I agree for the moment we shouldn't be blocked on it.

zuiderkwast:
@jonathanspw is looking at solving this in Fedora, probably add TCL 8 back.

zuiderkwast:
The CI failures are not related to the TCL version. They are other failures, mostly known ones, not related to this PR.

zuiderkwast:
@valkey-io/valkey-committers Can someone review this?

We should start supporting TCL 9. Distros have it now and people start having trouble running the tests locally.

ranshid (approved):

runtest
```diff
@@ -1,5 +1,5 @@
 #!/bin/sh
-TCL_VERSIONS="8.5 8.6 8.7"
+TCL_VERSIONS="8.5 8.6 9.0"
```
enjoy-binbin, line 2: we are remoing 8.7? If some people are using 8.7, if that be a problem?

enjoy-binbin (approved):

runtest
zuiderkwast: TCL 8.7 was never released. I didn't think it will ever be released.

I see now there has been some development. In 2021 it was in alpha stage according to https://www.tcl-lang.org/software/tcltk/8.7.html but I guess 8.7 will not be developed much anymore after 9.0 is released. If 8.7 is released later, we can add it again.

enjoy-binbin: ok, SGTM.

roshkhatri:
Do we need to backport these changes to release branches?
I see tests are failing due to TCL 8.5

sarthakaggarwal97:
Context: The goal is to make already released branches flaky test free and backport the existing [effort](https://github.com/valkey-io/valkey/issues/2701) from unstable to stabilize the branches. Currently due to the [TCL 8.5 check](https://github.com/valkey-io/valkey/blob/8.0/tests/test_helper.tcl#L5), fedora runners aren't able to run the test because they have 9.0 installed. 

Sample Link of the failure: https://github.com/sarthakaggarwal97/valkey/actions/runs/19113712295/job/54617221974

So I guess we will need to backport these changes to allow for TCL 9.0. @zuiderkwast @madolson any thoughts on this? Is it safe to remove this check (considering there could be breaking changes from TCL 8 to TCL 9?)

Another option is to skip fedora tests for now on the already released branches.

zuiderkwast:
Yes, it's fine to backport this PR. It's a good idea.

> Is it safe to remove this check (considering there could be breaking changes from TCL 8 to TCL 9?)

Yes, this PR fixes the small differences that need to be fixed for the tests to work with TCL 8 and 9. One example is this change:

```diff
-    fconfigure $fd -encoding binary
+    fconfigure $fd -translation binary
```

ranshid:
I see that this: https://github.com/valkey-io/valkey/pull/1673 was not backported to 9.0 (intentionally?)
we have it in Unstable, 8.0 and 8.1. should we backport to 9.0? (@zuiderkwast)

zuiderkwast:
Yeah, let's backport to 9.0.

PR #1674, opened 2025-02-05
Implement: CLUSTER REPLICATE NO ONE

skolosov-snap:
Currently, ValKey doesn't allow to detach replica attached to primary node. So, if you want to change cluster topology the only way to do it is to reset (````CLUSTER RESET```` command) the node. However, this results into removing node from the cluster what affects clients. All clients will keep sending traffic to this node (with getting inaccurate responses) until they refresh their topology.

In this change we implement supporting of new argument for CLUSTER REPLICATE command: ````CLUSTER REPLICATE NO ONE````. When calling this command the node will be converted from replica to empty primary node but still staying in the cluster. Thus, all traffic coming from the clients to this node can be redirected to correct node.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -7049,7 +7049,36 @@ int clusterCommandSpecial(client *c) {
         clusterDelNode(n);
         clusterDoBeforeSleep(CLUSTER_TODO_UPDATE_STATE | CLUSTER_TODO_SAVE_CONFIG);
         addReply(c, shared.ok);
-    } else if (!strcasecmp(c->argv[1]->ptr, "replicate") && c->argc == 3) {
+    } else if (!strcasecmp(c->argv[1]->ptr, "replicate") && (c->argc == 3 || c->argc == 4)) {
+        /* CLUSTER REPLICATE (<NODE ID> | NO ONE)*/
+        if (c->argc == 4) {
+            /* CLUSTER REPLICATE NO ONE */
+            if (strcasecmp(c->argv[2]->ptr, "NO") != 0 || strcasecmp(c->argv[3]->ptr, "ONE") != 0) {
+                addReplySubcommandSyntaxError(c);
+                return 1;
+            }
+            if (nodeIsPrimary(myself)) {
+                addReply(c, shared.ok);
+                return 1;
+            }
+            sds client = catClientInfoShortString(sdsempty(), c, server.hide_user_data_from_log);
+            serverLog(LL_NOTICE, "Stop replication and turning myself into empty primary (request from '%s').", client);
+            sdsfree(client);
+            clusterSetNodeAsPrimary(myself);
+            clusterPromoteSelfToPrimary();
+            flushAllDataAndResetRDB(server.repl_replica_lazy_flush ? EMPTYDB_ASYNC : EMPTYDB_NO_FLAGS);
+            clusterCloseAllSlots();
+            resetManualFailover();
+
+            /* Moving new primary to its own shard. */
+            char new_shard_id[CLUSTER_NAMELEN];
+            getRandomHexChars(new_shard_id, CLUSTER_NAMELEN);
+            updateShardId(myself, new_shard_id);
+
+            clusterDoBeforeSleep(CLUSTER_TODO_UPDATE_STATE | CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_BROADCAST_ALL);
+            addReply(c, shared.ok);
+            return 1;
+        }
         /* CLUSTER REPLICATE <NODE ID> */
         /* Lookup the specified node in our table. */
         clusterNode *n = clusterLookupNode(c->argv[2]->ptr, sdslen(c->argv[2]->ptr));
--- a/src/commands.def
+++ b/src/commands.def
@@ -768,7 +768,9 @@ struct COMMAND_ARG CLUSTER_REPLICAS_Args[] = {
 
 #ifndef SKIP_CMD_HISTORY_TABLE
 /* CLUSTER REPLICATE history */
-#define CLUSTER_REPLICATE_History NULL
+commandHistory CLUSTER_REPLICATE_History[] = {
+{"9.0.0","Added support of 'NO ONE' arg instead of <node-id> resulting into detaching replica from primary node."},
+};
 #endif
 
 #ifndef SKIP_CMD_TIPS_TABLE
@@ -781,9 +783,21 @@ struct COMMAND_ARG CLUSTER_REPLICAS_Args[] = {
 #define CLUSTER_REPLICATE_Keyspecs NULL
 #endif
 
+/* CLUSTER REPLICATE args no_one argument table */
+struct COMMAND_ARG CLUSTER_REPLICATE_args_no_one_Subargs[] = {
+{MAKE_ARG("no",ARG_TYPE_PURE_TOKEN,-1,"NO",NULL,NULL,CMD_ARG_NONE,0,NULL)},
+{MAKE_ARG("one",ARG_TYPE_PURE_TOKEN,-1,"ONE",NULL,NULL,CMD_ARG_NONE,0,NULL)},
+};
+
+/* CLUSTER REPLICATE args argument table */
+struct COMMAND_ARG CLUSTER_REPLICATE_args_Subargs[] = {
+{MAKE_ARG("node-id",ARG_TYPE_STRING,-1,NULL,NULL,NULL,CMD_ARG_NONE,0,NULL)},
+{MAKE_ARG("no-one",ARG_TYPE_BLOCK,-1,NULL,NULL,"9.0.0",CMD_ARG_NONE,2,NULL),.subargs=CLUSTER_REPLICATE_args_no_one_Subargs},
+};
+
 /* CLUSTER REPLICATE argument table */
 struct COMMAND_ARG CLUSTER_REPLICATE_Args[] = {
-{MAKE_ARG("node-id",ARG_TYPE_STRING,-1,NULL,NULL,NULL,CMD_ARG_NONE,0,NULL)},
+{MAKE_ARG("args",ARG_TYPE_ONEOF,-1,NULL,NULL,NULL,CMD_ARG_NONE,2,NULL),.subargs=CLUSTER_REPLICATE_args_Subargs},
 };
 
 /********** CLUSTER RESET ********************/
@@ -1024,7 +1038,7 @@ struct COMMAND_STRUCT CLUSTER_Subcommands[] = {
 {MAKE_CMD("myshardid","Returns the shard ID of a node.","O(1)","7.2.0",CMD_DOC_NONE,NULL,NULL,"cluster",COMMAND_GROUP_CLUSTER,CLUSTER_MYSHARDID_History,0,CLUSTER_MYSHARDID_Tips,1,clusterCommand,2,CMD_LOADING|CMD_STALE,0,CLUSTER_MYSHARDID_Keyspecs,0,NULL,0)},
 {MAKE_CMD("nodes","Returns the cluster configuration for a node.","O(N) where N is the total number of Cluster nodes","3.0.0",CMD_DOC_NONE,NULL,NULL,"cluster",COMMAND_GROUP_CLUSTER,CLUSTER_NODES_History,0,CLUSTER_NODES_Tips,1,clusterCommand,2,CMD_LOADING|CMD_STALE,0,CLUSTER_NODES_Keyspecs,0,NULL,0)},
 {MAKE_CMD("replicas","Lists the replica nodes of a primary node.","O(N) where N is the number of replicas.","5.0.0",CMD_DOC_NONE,NULL,NULL,"cluster",COMMAND_GROUP_CLUSTER,CLUSTER_REPLICAS_History,0,CLUSTER_REPLICAS_Tips,1,clusterCommand,3,CMD_ADMIN|CMD_STALE,0,CLUSTER_REPLICAS_Keyspecs,0,NULL,1),.args=CLUSTER_REPLICAS_Args},
-{MAKE_CMD("replicate","Configure a node as replica of a primary node.","O(1)","3.0.0",CMD_DOC_NONE,NULL,NULL,"cluster",COMMAND_GROUP_CLUSTER,CLUSTER_REPLICATE_History,0,CLUSTER_REPLICATE_Tips,0,clusterCommand,3,CMD_NO_ASYNC_LOADING|CMD_ADMIN|CMD_STALE,0,CLUSTER_REPLICATE_Keyspecs,0,NULL,1),.args=CLUSTER_REPLICATE_Args},
+{MAKE_CMD("replicate","Configure a node as replica of a primary node or detach a replica from its primary.","O(1)","3.0.0",CMD_DOC_NONE,NULL,NULL,"cluster",COMMAND_GROUP_CLUSTER,CLUSTER_REPLICATE_History,1,CLUSTER_REPLICATE_Tips,0,clusterCommand,-3,CMD_NO_ASYNC_LOADING|CMD_ADMIN|CMD_STALE,0,CLUSTER_REPLICATE_Keyspecs,0,NULL,1),.args=CLUSTER_REPLICATE_Args},
 {MAKE_CMD("reset","Resets a node.","O(N) where N is the number of known nodes. The command may execute a FLUSHALL as a side effect.","3.0.0",CMD_DOC_NONE,NULL,NULL,"cluster",COMMAND_GROUP_CLUSTER,CLUSTER_RESET_History,0,CLUSTER_RESET_Tips,0,clusterCommand,-2,CMD_ADMIN|CMD_STALE|CMD_NOSCRIPT,0,CLUSTER_RESET_Keyspecs,0,NULL,1),.args=CLUSTER_RESET_Args},
 {MAKE_CMD("saveconfig","Forces a node to save the cluster configuration to disk.","O(1)","3.0.0",CMD_DOC_NONE,NULL,NULL,"cluster",COMMAND_GROUP_CLUSTER,CLUSTER_SAVECONFIG_History,0,CLUSTER_SAVECONFIG_Tips,0,clusterCommand,2,CMD_NO_ASYNC_LOADING|CMD_ADMIN|CMD_STALE,0,CLUSTER_SAVECONFIG_Keyspecs,0,NULL,0)},
 {MAKE_CMD("set-config-epoch","Sets the configuration epoch for a new node.","O(1)","3.0.0",CMD_DOC_NONE,NULL,NULL,"cluster",COMMAND_GROUP_CLUSTER,CLUSTER_SET_CONFIG_EPOCH_History,0,CLUSTER_SET_CONFIG_EPOCH_Tips,0,clusterCommand,3,CMD_NO_ASYNC_LOADING|CMD_ADMIN|CMD_STALE,0,CLUSTER_SET_CONFIG_EPOCH_Keyspecs,0,NULL,1),.args=CLUSTER_SET_CONFIG_EPOCH_Args},
--- a/src/commands/cluster-replicate.json
+++ b/src/commands/cluster-replicate.json
@@ -1,12 +1,18 @@
 {
     "REPLICATE": {
-        "summary": "Configure a node as replica of a primary node.",
+        "summary": "Configure a node as replica of a primary node or detach a replica from its primary.",
         "complexity": "O(1)",
         "group": "cluster",
         "since": "3.0.0",
-        "arity": 3,
+        "arity": -3,
         "container": "CLUSTER",
         "function": "clusterCommand",
+        "history": [
+            [
+                "9.0.0",
+                "Added support of 'NO ONE' arg instead of <node-id> resulting into detaching replica from primary node."
+            ]
+        ],
         "command_flags": [
             "NO_ASYNC_LOADING",
             "ADMIN",
@@ -14,8 +20,31 @@
         ],
         "arguments": [
             {
-                "name": "node-id",
-                "type": "string"
+                "name": "args",
+                "type": "oneof",
+                "arguments": [
+                    {
+                        "name": "node-id",
+                        "type": "string"
+                    },
+                    {
+                        "name": "no-one",
+                        "type": "block",
+                        "since": "9.0.0",
+                        "arguments": [
+                            {
+                                "name": "no",
+                                "type": "pure-token",
+                                "token": "NO"
+                            },
+                            {
+                                "name": "one",
+                                "type": "pure-token",
+                                "token": "ONE"
+                            }
+                        ]
+                    }
+                ]
             }
         ],
         "reply_schema": {
--- a/tests/unit/cluster/replica-detach.tcl
+++ b/tests/unit/cluster/replica-detach.tcl
@@ -1,0 +1,85 @@
+proc find_slots_for_node {node_idx node_id_to_find} {
+    set shards_cfg [R $node_idx CLUSTER SHARDS]
+    foreach shard_cfg $shards_cfg {
+        set nodes [dict get $shard_cfg nodes]
+        foreach node $nodes {
+            if {[dict get $node id] eq $node_id_to_find} {
+                return [dict get $shard_cfg slots]
+            }
+        }
+    }
+    return [list]
+}
+
+proc find_primary_node {number_of_shards shard_id} {
+    # primary indexes: [0..$number_of_shards-1]
+    for {set i 0}  {$i < $number_of_shards} {incr i} {
+        if {[R $i CLUSTER MYSHARDID] eq $shard_id} {
+            return $i
+        }
+    }
+    fail "Can't find shard $shard_id across $number_of_shards primaries."
+}
+
+set number_of_shards 3
+start_cluster $number_of_shards $number_of_shards {tags {external:skip cluster}} {
+    test "Cluster should start ok" {
+        wait_for_cluster_state ok
+    }
+
+    test "CLUSTER REPLICATE NO ONE should turn node into empty primary" {
+        upvar number_of_shards number_of_shards
+        # primary indexes: [0..$number_of_shards-1]
+        # replica indexes: [$number_of_shards..2*$number_of_shards-1]
+        # We are just taking first replica index for testing.
+        set replica_node $number_of_shards
+        set replica_node_id [R $replica_node CLUSTER MYID]
+        set shard_id [R $replica_node CLUSTER MYSHARDID]
+        # make sure node is replica
+        wait_for_condition 100 100 {
+            [string first "slave" [R $replica_node ROLE]] >= 0
+        } else {
+            puts "R $replica_node ROLE: [R $replica_node ROLE]"
+            fail "R $replica_node didn't assume replica role in time"
+        }
+        # Adding some data to replica's shard
+        set primary_node [find_primary_node $number_of_shards $shard_id]
+        wait_for_condition 100 100 {
+            ([catch {R $primary_node SET [randomKey] "some-value"} resp] == 0) && ($resp eq {OK})
+        } else {
+            puts "R $primary_node SCAN 0: [R $primary_node SCAN 0]"
+            fail "Failed to add some key to primary node $primary_node in shard $shard_id."
+        }
+        # and wait for data is replicated to replica
+        wait_for_condition 100 100 {
+            [R $replica_node DBSIZE] > 0
+        } else {
+            puts "R $replica_node DBSIZE: [R $replica_node DBSIZE]"
+            fail "Replica $replica_node didn't replicate data from primary."
+        }
+
+        assert_equal "OK" [R $replica_node CLUSTER REPLICATE NO ONE]
+
+        # make sure node is turned into primary
+        wait_for_condition 100 100 {
+            [string first "master" [R $replica_node ROLE]] >= 0
+        } else {
+            puts "R $replica_node ROLE: [R $replica_node ROLE]"
+            fail "R $replica_node didn't assume primary role in time"
+        }
+
+        # make sure shard id is changed
+        assert_not_equal $shard_id [R $replica_node CLUSTER MYSHARDID]
+        assert_equal 0 [R $replica_node DBSIZE]
+
+        # checking that new primary has no slots across all nodes
+        foreach node_idx {0 1 2 3 4 5} {
+            wait_for_condition 100 100 {
+                [llength [find_slots_for_node $node_idx $replica_node_id]] == 0
+            } else {
+                puts "R $node_idx still returns node $replica_node_id owning slots: [find_slots_for_node $node_idx $replica_node_id]"
+                fail "R $node_idx didn't refresh topology after detaching replica $replica_node/$replica_node_id"
+            }
+        }
+    }
+}


src/cluster_legacy.c
```diff
@@ -7029,48 +7029,73 @@ int clusterCommandSpecial(client *c) {
...
-    } else if (!strcasecmp(c->argv[1]->ptr, "replicate") && c->argc == 3) {
+    } else if (!strcasecmp(c->argv[1]->ptr, "replicate") && (c->argc == 3 || c->argc == 4)) {
         /* CLUSTER REPLICATE <NODE ID> */
         /* Lookup the specified node in our table. */
-        clusterNode *n = clusterLookupNode(c->argv[2]->ptr, sdslen(c->argv[2]->ptr));
-        if (!n) {
-            addReplyErrorFormat(c, "Unknown node %s", (char *)c->argv[2]->ptr);
-            return 1;
+        clusterNode *n = NULL;
+        if (c->argc == 4) {
```
enjoy-binbin, line 7054: can you try to do a fast path in here? the diff is kind of ugly and hard to read.

skolosov-snap: I'm sorry for naive question, but what is a fast path? Is this some kind of formatting tool?

src/cluster_legacy.c
```diff
@@ -7029,48 +7029,73 @@ int clusterCommandSpecial(client *c) {
...
+        if (n == NULL) {
+            if (nodeIsPrimary(myself)) {
+                addReply(c,shared.ok);
+                return 1;
+            }
+            serverLog(LL_NOTICE,"Stop replication and turning myself into empty primary.");
+            clusterSetNodeAsPrimary(myself);
+            if (server.primary != NULL) {
+                replicationUnsetPrimary();
+            }
```
hpatro: All other invocation of `replicationUnsetPrimary()` don't have this wrapped under the condition. Is it unnecessary to invoke it if server.primary is NULL ?

hpatro: Also unsure why `clusterPromoteSelfToPrimary` was introduced, seems like it's the same behavior at this point but good to call this abstraction if there will be additional steps introduced in the future  for cluster mode.

hpatro (commented):
`cluster-replicate.json` file should be updated and as part of the build `commands.def` will get updated. or if it was accidentally not staged, please add that.

Also, could you run the clang-format on your end to fix some of the formatting issue.

src/cluster_legacy.c
enjoy-binbin: like simply just move this block to here and do a return, so you dont have the change the other lines, so the diff will be smaller
```
if (n == NULL) {
            if (nodeIsPrimary(myself)) {
                addReply(c,shared.ok);
                return 1;
            }
            serverLog(LL_NOTICE,"Stop replication and turning myself into empty primary.");
            clusterSetNodeAsPrimary(myself);
            if (server.primary != NULL) {
                replicationUnsetPrimary();
            }
....
}
```

skolosov-snap: Updated.

src/cluster_legacy.c
skolosov-snap: >Also unsure why clusterPromoteSelfToPrimary was introduced, seems like it's the same behavior at this point but good to call this abstraction if there will be additional steps introduced in the future for cluster mode.

Ok.

skolosov-snap: > All other invocation of replicationUnsetPrimary() don't have this wrapped under the condition. Is it unnecessary to invoke it if server.primary is NULL ?

Probably not needed. It looks like left over from KeyDB's implementation where primary needed to be passed to [replicationUnsetMaster](https://github.com/Snapchat/KeyDB/blob/603ebb27fb82a27fb98b0feb6749b0f7661a1c4b/src/cluster.cpp#L4941-L4945)

skolosov-snap:
> `cluster-replicate.json` file should be updated and as part of the build `commands.def` will get updated. or if it was accidentally not staged, please add that.
> 
> Also, could you run the clang-format on your end to fix some of the formatting issue.

Updated.

src/cluster_legacy.c
```diff
@@ -7029,9 +7029,29 @@ int clusterCommandSpecial(client *c) {
...
+            if (nodeIsPrimary(myself)) {
+                addReply(c, shared.ok);
+                return 1;
+            }
+            serverLog(LL_NOTICE, "Stop replication and turning myself into empty primary.");
+            clusterSetNodeAsPrimary(myself);
+            clusterPromoteSelfToPrimary();
+            int empty_db_flags = server.repl_replica_lazy_flush ? EMPTYDB_ASYNC : EMPTYDB_NO_FLAGS;
+            emptyData(-1, empty_db_flags, NULL);
```
zuiderkwast: To compare with the non-cluster command `REPLICAOF NO ONE`, that one does not empty the data.

In a cluster it's different, but why exactly? Do other cluster commands flush the data like this?

src/commands/cluster-replicate.json
```diff
@@ -14,8 +14,30 @@
...
+                            {
+                                "name": "no",
+                                "type": "pure-token",
+                                "token": "NO"
+                            },
+                            {
+                                "name": "one",
+                                "type": "pure-token",
+                                "token": "ONE"
+                            }
```
zuiderkwast: In `REPLICAOF ip port | REPLICAOF NO ONE`, it is two arguments in both cases, like host=NO port=ONE, so it is the same arity in both cases.

For `CLUSTER REPLICATE (node-id | NO ONE)`, NO ONE replaces the node-id, so the arity is 3 or 4. Need to change `arity: 3` in line 7 of this JSON file. Or maybe should use a single token instead of NO ONE? For example NONE?

zuiderkwast (commented):
This feature makes sense to me.

@valkey-io/core-team New arguments = major decision. Please approve or vote if you agree.

src/cluster_legacy.c
skolosov-snap: In this case we should delete data because after detaching replica it is still in the cluster but has a role of empty primary. Thus, data should never be used. Any request will be redirected to other node in the cluster. In case of ```REPLICAOF NO ONE``` after disabling replication the node still can accept client requests and serve it with existing data.

src/commands/cluster-replicate.json
skolosov-snap: Updated arity field. I'm not sure what this field is for. Since it can be calculated from ```arguments``` field. Also everything works even with ```arity=3```. I think similarity to ```REPLICAOF NO ONE``` is better than trying to keep the same arity for different cases. I think client are not thinking about arity at all but similarity between commands is easier for them remember. In addition, for us it is easier to distinguish ```<node-id>``` from a special case.

zuiderkwast: OK.

I don't know why we have arity here but it's returned by COMMAND and COMMAND INFO, which I think some clients use.

We can use a negative number `-3`. It means *3 or more*.

zuiderkwast:
The CI job "DCO" is failing. You need to use `git commit -s`. See the Details link next to the DCO job.

Why we need it? See here: https://github.com/valkey-io/valkey/blob/unstable/CONTRIBUTING.md#developer-certificate-of-origin thanks!

src/commands/cluster-replicate.json
```diff
@@ -1,10 +1,10 @@
 {
     "REPLICATE": {
-        "summary": "Configure a node as replica of a primary node.",
+        "summary": "Configure a node as replica of a primary node or detaches replica from primary.",
         "complexity": "O(1)",
         "group": "cluster",
         "since": "3.0.0",
-        "arity": 3,
+        "arity": -4,
```
zuiderkwast:
```suggestion
        "arity": -3,
```

-3 = 3 or more. `CLUSTER REPLICATE node-id` is 3.

skolosov-snap: Yes. Already figured it out.

skolosov-snap:
> The CI job "DCO" is failing. You need to use `git commit -s`. See the Details link next to the DCO job.
> 
> Why we need it? See here: https://github.com/valkey-io/valkey/blob/unstable/CONTRIBUTING.md#developer-certificate-of-origin thanks!

Done

src/cluster_legacy.c
```diff
@@ -7029,9 +7029,29 @@ int clusterCommandSpecial(client *c) {
         clusterDelNode(n);
         clusterDoBeforeSleep(CLUSTER_TODO_UPDATE_STATE | CLUSTER_TODO_SAVE_CONFIG);
         addReply(c, shared.ok);
-    } else if (!strcasecmp(c->argv[1]->ptr, "replicate") && c->argc == 3) {
-        /* CLUSTER REPLICATE <NODE ID> */
+    } else if (!strcasecmp(c->argv[1]->ptr, "replicate") && (c->argc == 3 || c->argc == 4)) {
+        /* CLUSTER REPLICATE (<NODE ID> | NO ONE)*/
         /* Lookup the specified node in our table. */
+        if (c->argc == 4) {
+            if (0 != strcasecmp(c->argv[2]->ptr, "NO") || 0 != strcasecmp(c->argv[3]->ptr, "ONE")) {
```
enjoy-binbin: we don't do this 0 != xxx trick, let's keep the same style for the code

src/cluster_legacy.c
```diff
@@ -7029,9 +7029,29 @@ int clusterCommandSpecial(client *c) {
...
+                addReply(c, shared.ok);
+                return 1;
+            }
+            serverLog(LL_NOTICE, "Stop replication and turning myself into empty primary.");
+            clusterSetNodeAsPrimary(myself);
+            clusterPromoteSelfToPrimary();
+            int empty_db_flags = server.repl_replica_lazy_flush ? EMPTYDB_ASYNC : EMPTYDB_NO_FLAGS;
+            emptyData(-1, empty_db_flags, NULL);
+            /* Reset manual failover state. */
+            resetManualFailover();
```
enjoy-binbin: it become a primary, do we need this reset? it has no way to enter the failover check logic. or at least we remove the comment since this line is super easy.
```suggestion
            resetManualFailover();
```

src/cluster_legacy.c
```suggestion
            emptyData(-1, server.repl_replica_lazy_flush ? EMPTYDB_ASYNC : EMPTYDB_NO_FLAGS, NULL);
```

src/cluster_legacy.c
```diff
@@ -7029,9 +7029,29 @@ int clusterCommandSpecial(client *c) {
...
+        if (c->argc == 4) {
+            if (0 != strcasecmp(c->argv[2]->ptr, "NO") || 0 != strcasecmp(c->argv[3]->ptr, "ONE")) {
+                addReplySubcommandSyntaxError(c);
+                return 1;
+            }
+            if (nodeIsPrimary(myself)) {
+                addReply(c, shared.ok);
+                return 1;
+            }
+            serverLog(LL_NOTICE, "Stop replication and turning myself into empty primary.");
```
enjoy-binbin: let's also use `catClientInfoShortString` to get the client info for a better audit.

src/cluster_legacy.c
skolosov-snap: I'm not sure what you mean by "no way to enter the failover check logic". This is to abort any in-progress failover (which can happen, right?).
Comment is removed.

src/cluster_legacy.c
skolosov-snap: Done

src/cluster_legacy.c
skolosov-snap: Done

src/cluster_legacy.c
skolosov-snap: Updated.
I see a tons of ```strcasecmp()==0``` calls. You probably want to refactor it then.

src/cluster_legacy.c
enjoy-binbin: i mean, when the node become a primary, it has no way enter this mf handle logic.
```
        if (nodeIsReplica(myself)) {
            clusterHandleManualFailover();
```

skolosov-snap: Sorry. I'm not sure I got you right. If you mean that there is no way to have manual failover to be in progress by the time we come to https://github.com/valkey-io/valkey/blob/bed392bebf6a0110f0cf3e86cf3f2c6845f5437a/src/cluster_legacy.c#L7050, then I would disagree with that. Manual failover could start before we got ```CLUSTER REPLICATE NO ONE```. Also, ```CLUSTER RESET``` (which also may changes replica to primary) does the same: it resets any in-progress failover: https://github.com/valkey-io/valkey/blob/bed392bebf6a0110f0cf3e86cf3f2c6845f5437a/src/cluster_legacy.c#L1298

enjoy-binbin: i know it could start before the NO ONE, i mean, if mf is in progress, will it be a problem if it is not reset? What will happen? I know CLUSTER RESET do the same thing, it is a old time stuff, i am ok with resetting it, i was just thinking about this, we definitely have other places where we haven't reset it when the role changed.

src/cluster_legacy.c
enjoy-binbin: `strcasecmp()==0` is ok i guess, `0==strcasecmp()` is bad in reading.

src/cluster_legacy.c
zuiderkwast: > we definitely have other places where we haven't reset it when the role changed.

Interesting. Is this a problem?

skolosov-snap:
Any objection to merge it?

zuiderkwast:
We're busy making the 8.1.0 release candidate just now. This one will need to wait and get merged after that.

src/cluster_legacy.c
skolosov-snap: What is the reason of not resetting it?

madolson:
> Any objection to merge it?

We should also have some tests validating this new behavior works as intended. Have a cluster, disconnect the replica, make sure slots/shards and all are still consistent and the rest of the cluster agrees on the state.

src/cluster_legacy.c
enjoy-binbin: i am not against it, this is just a question, after it becomes a primary, will there be any problems with the uncleaned mf failover? As far as i know there is no problem (only a log issue like the replica (empty primary) will print the mf timeout something like that). The original primary has a known issue with pause timeout.

i am ok with resetting it. I would probably prefer to have one that refuses to execute NO ONE if the replica is in failover or is in mf failover. The admin should do it in a stable state, which means the replica should not be in the failover state (or at least should not in mf failover). But there is no harm i guess (i dont see a problem here), so i am also ok with it.

src/commands/cluster-replicate.json
```diff
@@ -14,8 +14,30 @@
...
+                "type": "oneof",
+                "arguments": [
+                    {
+                        "name": "node-id",
+                        "type": "string"
+                    },
+                    {
+                        "name": "no-one",
+                        "type": "block",
+                        "arguments": [
```
enjoy-binbin, line 34: we should also add a since field, and a history field to mention this args.

src/cluster_legacy.c
```diff
@@ -7029,9 +7029,29 @@ int clusterCommandSpecial(client *c) {
         clusterDelNode(n);
         clusterDoBeforeSleep(CLUSTER_TODO_UPDATE_STATE | CLUSTER_TODO_SAVE_CONFIG);
         addReply(c, shared.ok);
-    } else if (!strcasecmp(c->argv[1]->ptr, "replicate") && c->argc == 3) {
-        /* CLUSTER REPLICATE <NODE ID> */
+    } else if (!strcasecmp(c->argv[1]->ptr, "replicate") && (c->argc == 3 || c->argc == 4)) {
+        /* CLUSTER REPLICATE (<NODE ID> | NO ONE)*/
         /* Lookup the specified node in our table. */
```
enjoy-binbin: please also move this comment to below, near the clusterLookupNode one.

PingXie:
>  if you want to change cluster topology the only way to do it is to reset (`CLUSTER RESET` command) the node. However, this results into removing node from the cluster what affects clients. 

Can we introduce a new mode so it doesn't forget all the nodes in the cluster? I think conceptually we are discussing a form of reset still so it seems to me that the solution is too tactical. Maybe `CLUSTER RESET SOFT`?

BTW, I just noticed that the forget path is not always working. The reset node joined back to the cluster quickly.

> In this change we implement supporting of new argument for CLUSTER REPLICATE command: `CLUSTER REPLICATE NO ONE`. When calling this command the node will be converted from replica to empty primary node but still staying in the cluster. Thus, all traffic coming from the clients to this node can be redirected to correct node.

I don't see the implementation moves the node to a new shard. This would leave two primaries (one real and one empty) in the original shard, which will confuse the client.

hpatro:
> I don't see the implementation moves the node to a new shard. This would leave two primaries (one real and one empty) in the original shard, which will confuse the client.

Good catch!

skolosov-snap:
> > if you want to change cluster topology the only way to do it is to reset (`CLUSTER RESET` command) the node. However, this results into removing node from the cluster what affects clients.
> 
> Can we introduce a new mode so it doesn't forget all the nodes in the cluster? I think conceptually we are discussing a form of reset still so it seems to me that the solution is too tactical. Maybe `CLUSTER RESET SOFT`?
> 

IMHO that is just a syntactical question. Whatever command name we would come up with, the behavior of it would be the same: **turn replica into primary with leaving it in the cluster**. If you think the name of `CLUSTER RESET SOFT` is better I can support it in that name. My personal opinion is that `CLUSTER RESET` is not the best command to implement this feature, because the main thing what `CLUSTER RESET` does is excluding node from the cluster and that is exactly what we want to avoid. On the other hand `CLUSTER REPLICATE NO ONE` is consistent with similar non-cluster version of `REPLICAOF NO ONE` what should not confuse client but even give some kind of similarity.

> BTW, I just noticed that the forget path is not always working.

What do you mean by "forget path is not always working"? What forget path? We are not doing any forgetting here.

> The reset node joined back to the cluster quickly.

AFAIU if node is reset it will not be added back to the cluster automatically, but only when somebody does it explicitly. So if you want to remove replica manually (i.e. node is scheduled for maintenance) the only option for you is to reset with affecting all clients (start seeing errors).

> > In this change we implement supporting of new argument for CLUSTER REPLICATE command: `CLUSTER REPLICATE NO ONE`. When calling this command the node will be converted from replica to empty primary node but still staying in the cluster. Thus, all traffic coming from the clients to this node can be redirected to correct node.
> 
> I don't see the implementation moves the node to a new shard. This would leave two primaries (one real and one empty) in the original shard, which will confuse the client.

What is a **shard**? Would you please shed a light on this term? Is it something special for ValKey? AFAIU shard is primary with replicas attached to it. So, when we switched role of the node from replica to empty primary, doesn't it mean that we moved it to separate new shard?

src/cluster_legacy.c
skolosov-snap: `0==strcasecmp()` is standard/best practice in C/C++ for some valid reasons. Not a big deal to me.

src/commands/cluster-replicate.json
skolosov-snap: updated.

src/cluster_legacy.c
skolosov-snap: updated.

skolosov-snap:
I believe I figured it out. There is internal ````shards```` dict that needs to be updated.

zuiderkwast:
@PingXie

> Can we introduce a new mode so it doesn't forget all the nodes in the cluster? I think conceptually we are discussing a form of reset still so it seems to me that the solution is too tactical. Maybe `CLUSTER RESET SOFT`?

I agree with @skolosov-snap that CLUSTER REPLICATE NO ONE is better. The node is not leaving the cluster so RESET seems less intuitive.

> BTW, I just noticed that the forget path is not always working. The reset node joined back to the cluster quickly.

To make the cluster forget a node, you need to send `CLUSTER FORGET` to the other nodes in the cluster. If you only reset a node, it will disconnect but the other nodes will find it again soon. I believe this is the documented behavior.

skolosov-snap:
> To make the cluster forget a node, you need to send `CLUSTER FORGET` to the other nodes in the cluster. If you only reset a node, it will disconnect but the other nodes will find it again soon. I believe this is the documented behavior.

I understand that. I didn't get how it relates to this change?

skolosov-snap:
> To make the cluster forget a node, you need to send `CLUSTER FORGET` to the other nodes in the cluster. If you only reset a node, it will disconnect but the other nodes will find it again soon. I believe this is the documented behavior.

Looking at [the doc](https://valkey.io/commands/cluster-reset/) it should behave differently:
> This command is mainly useful to re-provision a Valkey Cluster node in order to be used in the context of a new, different cluster.

So, I believe potentially it may be fixed in the future.

@PingXie, @zuiderkwast You are right about current `CLUSTER RESET` behavior. However, it takes about 15 secs to re-join the cluster. So, we gonna have 15 seconds of errors which is pretty long for errors in such scenarios:
```
27859:M 19 Feb 2025 17:11:14.472 * Connection with replica 127.0.0.1:17383 lost.
27859:M 19 Feb 2025 17:11:29.523 * Sending MEET packet to node 096e0230bddddb11d32e36497adcc4235bffa7d1 () because there is no inbound link for it
27859:M 19 Feb 2025 17:11:29.593 * Successfully completed handshake with 096e0230bddddb11d32e36497adcc4235bffa7d1 ()
```

PingXie:
@skolosov-snap @zuiderkwast you have a good point. ` CLUSTER REPLICATE NO ONE` works better. 

BTW, a follow up thought, should we consider a single token `NONE` as opposed to `NO ONE`? I understand the rationale of mimicking `REPLICAOF NO ONE` but it using two tokens for one concept seems quite arbitrary. I wonder if it is a lesser evil if we go with `NONE` this time? So `CLUSTER REPLICATE NONE` perhaps?

zuiderkwast:
> a single token `NONE` as opposed to `NO ONE`?

I suggested it earlier in this PR in comment https://github.com/valkey-io/valkey/pull/1674#discussion_r1949147014 and I was already convinced that NO ONE is better. :laughing:

PingXie:
> > a single token `NONE` as opposed to `NO ONE`?
> 
> I suggested it earlier in this PR in comment [#1674 (comment)](https://github.com/valkey-io/valkey/pull/1674#discussion_r1949147014) and I was already convinced that NO ONE is better. 😆

Hmm the comment link seems broken. What convinced you? :)

skolosov-snap:
> BTW, a follow up thought, should we consider a single token `NONE` as opposed to `NO ONE`? I understand the rationale of mimicking `REPLICAOF NO ONE` but it using two tokens for one concept seems quite arbitrary. I wonder if it is a lesser evil if we go with `NONE` this time? So `CLUSTER REPLICATE NONE` perhaps?

Not a big deal to me, but IMHO consistency/similarity between commands is better. What benefit will we get if we replace it with single toke?
Also, I understand that probability of having ````NONE```` node-id is near the zero, but is not zero. I.e. in a future if we let say allow to set ID by user or in some kind of test.
Again, not a big deal to me, but I don't see any benefits from making a single token (except may be shorter ````if-statement```` in the code).

PingXie:
>Also, I understand that probability of having NONE node-id is near the zero, but is not zero. 

The probability will be practically 0. 

> I.e. in a future if we let say allow to set ID by user or in some kind of test.

We have `human_nodename` for this purpose.

```c
    sds human_nodename;                     /* The known human readable nodename for this node */
```

zuiderkwast:
> > > a single token `NONE` as opposed to `NO ONE`?
> > 
> > I suggested it earlier in this PR in comment [#1674 (comment)](https://github.com/valkey-io/valkey/pull/1674#discussion_r1949147014) and I was already convinced that NO ONE is better. 😆
> 
> Hmm the comment link seems broken. What convinced you? :)

Interesting, the link works for me. It's one of the resolved comments above on `src/commands/cluster-replicate.json`.

Quoting @skolosov-snap's comment which made sense to me:

> I think similarity to REPLICAOF NO ONE is better than trying to keep the same arity for different cases. I think client are not thinking about arity at all but similarity between commands is easier for them remember. In addition, for us it is easier to distinguish `<node-id>` from a special case.

skolosov-snap:
@PingXie is command naming still concern for you?

madolson:
> The probability will be practically 0.

No it's zero, the node ID is hex characters, you can't have N or O.

I also think it's worth keeping `NO ONE` because that just makes in mirror replicaof.

tests/cluster/tests/12-replication.tcl
```diff
@@ -0,0 +1,61 @@
+# Replication test.
+#
+# Check CLUSTER REPLICATE commands.
+
+source "../tests/includes/init-tests.tcl"
+
+proc find_slots_for_node {node_idx node_id_to_find} {
```
madolson: This is the old style test, the new style is in tests/unit/cluster. These tests have a lot of missing debugging information and also aren't run as part of the CI since they don't get parallelized. 

The tests should port mostly cleanly, the main difference is that you need to call `create_cluster 3 6` at the start but most of the other helps still exist.

skolosov-snap: moved to ````tests/unit/cluster/replica-detach.tcl````

PingXie:
> @PingXie is command naming still concern for you?

I am still inclined to a single token but `NO ONE` is NOT a blocker for me. Can we take an explicit vote on the command name, for the record? @valkey-io/core-team

👍 for "NO ONE"

src/cluster_legacy.c
```diff
@@ -7047,7 +7047,36 @@ int clusterCommandSpecial(client *c) {
...
+            sds client = catClientInfoShortString(sdsempty(), c, server.hide_user_data_from_log);
+            serverLog(LL_NOTICE, "Stop replication and turning myself into empty primary (request from '%s').", client);
+            sdsfree(client);
+            clusterSetNodeAsPrimary(myself);
+            clusterPromoteSelfToPrimary();
+            emptyData(-1, server.repl_replica_lazy_flush ? EMPTYDB_ASYNC : EMPTYDB_NO_FLAGS, NULL);
+            clusterCloseAllSlots();
+            resetManualFailover();
+
+            // moving new primary to its own shard.
```
enjoy-binbin: @PingXie please also ack this logic.
```suggestion
            /* Moving new primary to its own shard. */
```

enjoy-binbin (approved):
The new code LGTM, thanks.

soloestoy:
> > @PingXie is command naming still concern for you?
> 
> I am still inclined to a single token but `NO ONE` is NOT a blocker for me. Can we take an explicit vote on the command name, for the record? @valkey-io/core-team
> 
> 👍 for "NO ONE"

I prefer `replicate no-one`, align with the original "node-id", "no-one" is just a special node-id

skolosov-snap:
> > @PingXie is command naming still concern for you?
> 
> I am still inclined to a single token but `NO ONE` is NOT a blocker for me. Can we take an explicit vote on the command name, for the record? @valkey-io/core-team
> 
> 👍 for "NO ONE"

I see 3 "Thumb ups" in your comment and I see 4 "Thumb ups" in PR description. Which are counts for voting? What is the quorum for positive decision? What is the outcome?

zuiderkwast:
@skolosov-snap The quorum is a majority of @valkey-io/core-team i.e. at least 4 members.

We have two competing suggestions. The votes, if I count the thumbs-up correctly:

* `NO ONE`: 4 votes (@madolson @PingXie @enjoy-binbin @zuiderkwast @hwware)
* `NO-ONE`: 3 votes (@soloestoy @PingXie @enjoy-binbin)

It looks like the first option already has a majority. Some members have voted for both. I assume this means you're OK with both options, but I'm not sure what you are thinking. Do you want to clarify?

skolosov-snap:
> @skolosov-snap The quorum is a majority of @valkey-io/core-team i.e. at least 4 members.
> 
> We have two competing suggestions. The votes, if I count the thumbs-up correctly:
> 
> * `NO ONE`: 4 votes (@madolson @PingXie @enjoy-binbin @zuiderkwast)
> * `NO-ONE`: 3 votes (@soloestoy @PingXie @enjoy-binbin)
> 
> It looks like the first option already has a majority. Some members have voted for both. I assume this means you're OK with both options, but I'm not sure what you are thinking. Do you want to clarify?

If you are asking my opinion I prefer similarity to existing commands which is "NO ONE". What I want is to get some progress in merging of this PR.

madolson:
> If you are asking my opinion I prefer similarity to existing commands which is "NO ONE". What I want is to get some progress in merging of this PR.

Btw, is this keydb behavior? If it is I would also bias towards keydb compatibility.

skolosov-snap:
> > If you are asking my opinion I prefer similarity to existing commands which is "NO ONE". What I want is to get some progress in merging of this PR.
> 
> Btw, is this keydb behavior? If it is I would also bias towards keydb compatibility.

Yes. This is cherry-pick from KeyDB.

skolosov-snap:
What is the state of this PR? Are you expecting any further changes from me?

madolson:
> What is the state of this PR? Are you expecting any further changes from me?

No, I think it's on the list to sync up with in our weekly engineering sync. https://github.com/valkey-io/valkey/wiki/Weekly-meeting Feel free to join, we'll make a decision and then it should unblock this PR to get merged.

hwware:
@madolson I just vote for “NO ONE”, thus can we have a decision offline?

madolson:
Oh nice, then I guess `NO ONE` wins. Then yeah, I think we can consider the major considerations all resolved now. I'll remove it from our topic.

skolosov-snap:
What is the next step?

hwware:
> What is the next step?

Wait for merge and release on Valkey 9

skolosov-snap:
> > What is the next step?
> 
> Wait for merge and release on Valkey 9

Actually, my question was related to the merging process. How should it be merged? Is it something I should do to make it happen? Merging button is still not available to me.

hwware:
When it is time to merge, Valkey maintainer will merge it for you, Do not worry @valkey-io/core-team

src/cluster_legacy.c
PingXie: LGTM

tests/unit/cluster/replica-detach.tcl
```diff
@@ -0,0 +1,50 @@
...
+
+start_cluster 3 3 {tags {external:skip cluster}} {
+    test "Cluster should start ok" {
+        wait_for_cluster_state ok
+    }
+
+    test "CLUSTER REPLICATE NO ONE should turn node into empty primary" {
+        set replica_node 3
+        set replica_node_id [R $replica_node CLUSTER MYID]
+        # make sure node is replica
```
PingXie, line 38: add some data to the shard and validate that data is purged too.

tests/unit/cluster/replica-detach.tcl
```diff
@@ -0,0 +1,50 @@
...
+
+        # make sure node is turned into primary
+        wait_for_condition 100 100 {
+            [string first "master" [R $replica_node ROLE]] >= 0
+        } else {
+            puts "R $replica_node ROLE: [R $replica_node ROLE]"
+            fail "R $replica_node didn't assume primary role in time"
+        }
+
+        # checking that new primary has no slots across all nodes
```
PingXie, line 75: let's also check that the shard id is changed

tests/unit/cluster/replica-detach.tcl
skolosov-snap: Done

tests/unit/cluster/replica-detach.tcl
skolosov-snap: Done

src/commands/cluster-replicate.json
```diff
@@ -1,21 +1,50 @@
...
         "complexity": "O(1)",
         "group": "cluster",
         "since": "3.0.0",
-        "arity": 3,
+        "arity": -3,
         "container": "CLUSTER",
         "function": "clusterCommand",
+        "history": [
+            [
+                "8.1.0",
```
hpatro: Could we update this to 9.0 ?

hpatro (approved):

src/commands/cluster-replicate.json
skolosov-snap: Done

hpatro: Could you push out the commands.def change as well?

skolosov-snap: Oops. Done.

src/commands/cluster-replicate.json
```diff
@@ -1,21 +1,50 @@
...
+                "type": "oneof",
+                "arguments": [
+                    {
+                        "name": "node-id",
+                        "type": "string"
+                    },
+                    {
+                        "name": "no-one",
+                        "type": "block",
+                        "since": "8.1.0",
```
zuiderkwast:
```suggestion
                        "since": "9.0.0",
```

src/commands/cluster-replicate.json
```diff
@@ -1,21 +1,50 @@
...
         "complexity": "O(1)",
         "group": "cluster",
         "since": "3.0.0",
-        "arity": 3,
+        "arity": -3,
         "container": "CLUSTER",
         "function": "clusterCommand",
+        "history": [
+            [
+                "9.0",
```
zuiderkwast: These files seem to use x.y.z format in all version numbers.

```suggestion
                "9.0.0",
```

zuiderkwast (approved):
LGTM.

We just need to update the version numbers to 9.0.0

zuiderkwast:
@PingXie @madolson Do you still want to review this or can we merge it?

src/commands/cluster-replicate.json
skolosov-snap: Done

src/commands/cluster-replicate.json
skolosov-snap: Done

hpatro:
I think it's safe to merge @zuiderkwast and address any further comment in future (if any).

Shall I go ahead?

zuiderkwast:
@hpatro I've been accused of YOLO-merging PRs before, when people were still reviewing. :laughing:

src/commands.def
```diff
@@ -1024,7 +1038,7 @@ struct COMMAND_STRUCT CLUSTER_Subcommands[] = {
...
 {MAKE_CMD("replicas","Lists the replica nodes of a primary node.","O(N) where N is the number of replicas.","5.0.0",CMD_DOC_NONE,NULL,NULL,"cluster",COMMAND_GROUP_CLUSTER,CLUSTER_REPLICAS_History,0,CLUSTER_REPLICAS_Tips,1,clusterCommand,3,CMD_ADMIN|CMD_STALE,0,CLUSTER_REPLICAS_Keyspecs,0,NULL,1),.args=CLUSTER_REPLICAS_Args},
-{MAKE_CMD("replicate","Configure a node as replica of a primary node.","O(1)","3.0.0",CMD_DOC_NONE,NULL,NULL,"cluster",COMMAND_GROUP_CLUSTER,CLUSTER_REPLICATE_History,0,CLUSTER_REPLICATE_Tips,0,clusterCommand,3,CMD_NO_ASYNC_LOADING|CMD_ADMIN|CMD_STALE,0,CLUSTER_REPLICATE_Keyspecs,0,NULL,1),.args=CLUSTER_REPLICATE_Args},
+{MAKE_CMD("replicate","Configure a node as replica of a primary node or detaches replica from primary.","O(1)","3.0.0",CMD_DOC_NONE,NULL,NULL,"cluster",COMMAND_GROUP_CLUSTER,CLUSTER_REPLICATE_History,1,CLUSTER_REPLICATE_Tips,0,clusterCommand,-3,CMD_NO_ASYNC_LOADING|CMD_ADMIN|CMD_STALE,0,CLUSTER_REPLICATE_Keyspecs,0,NULL,1),.args=CLUSTER_REPLICATE_Args},
```
PingXie:
```suggestion
{MAKE_CMD("replicate","Configure a node as replica of a primary node or detach a replica from its primary.","O(1)","3.0.0",CMD_DOC_NONE,NULL,NULL,"cluster",COMMAND_GROUP_CLUSTER,CLUSTER_REPLICATE_History,1,CLUSTER_REPLICATE_Tips,0,clusterCommand,-3,CMD_NO_ASYNC_LOADING|CMD_ADMIN|CMD_STALE,0,CLUSTER_REPLICATE_Keyspecs,0,NULL,1),.args=CLUSTER_REPLICATE_Args},
```

src/commands/cluster-replicate.json
```diff
@@ -1,21 +1,50 @@
 {
     "REPLICATE": {
-        "summary": "Configure a node as replica of a primary node.",
+        "summary": "Configure a node as replica of a primary node or detaches replica from primary.",
```
PingXie:
```suggestion
        "summary": "Configure a node as replica of a primary node or detach a replica from its primary.",
```

src/cluster_legacy.c
```diff
@@ -7049,7 +7049,36 @@ int clusterCommandSpecial(client *c) {
...
+            if (nodeIsPrimary(myself)) {
+                addReply(c, shared.ok);
+                return 1;
+            }
+            sds client = catClientInfoShortString(sdsempty(), c, server.hide_user_data_from_log);
+            serverLog(LL_NOTICE, "Stop replication and turning myself into empty primary (request from '%s').", client);
+            sdsfree(client);
+            clusterSetNodeAsPrimary(myself);
+            clusterPromoteSelfToPrimary();
+            emptyData(-1, server.repl_replica_lazy_flush ? EMPTYDB_ASYNC : EMPTYDB_NO_FLAGS, NULL);
```
PingXie: I think we should also kill the child process.

```suggestion
            flushAllDataAndResetRDB(server.repl_replica_lazy_flush ? EMPTYDB_ASYNC : EMPTYDB_NO_FLAGS);
```

PingXie (commented):
LGTM. Just a few minor suggestions. Thanks @skolosov-snap!

src/cluster_legacy.c
skolosov-snap: I applied this change but it looks to me like unnecessary. ````"CLUSTER REPLICATE NO ONE"```` can be only called on replica nodes. Thus, there can't be any RDB in process.

BTW, why don't we do the same on ````"CLUSTER RESET"```` command? In opposite to ````"CLUSTER REPLICATE NO ONE"````, ````"CLUSTER RESET"```` can be called on primary.

skolosov-snap: @PingXie 
I applied this change but it looks to me like unnecessary. ````"CLUSTER REPLICATE NO ONE"```` can be only called on replica nodes. Thus, there can't be any RDB in process (or it already should be in terminating phase).

BTW, why don't we do the same on ````"CLUSTER RESET"```` command? In opposite to ````"CLUSTER REPLICATE NO ONE"````, ````"CLUSTER RESET"```` can be called on primary.

madolson (approved):

madolson:
> I've been accused of YOLO-merging PRs before, when people were still reviewing. 😆

TBF, I'm fairly sure we've all done that. When I was at Kube-con I heard so many endless drama stories about dysfunctional projects where everyone was always mad at each other about code quality, sneaking changes without reviews, etc. By comparison people seem to think we are a highly functional group.

src/cluster_legacy.c
PingXie: I think it is a coincidence that we don't use chained replication in the cluster mode but that is an orthogonal decision so it is still preferable imo to not take that dependency.

Yeah I think cluster reset should kill the rdb child process too.

PingXie (approved):

PingXie:
> By comparison ~people seem to think~ we are a highly functional group.

Small correction 😀

hpatro:
@skolosov-snap Thanks for the PR. It's merged now 🥳 . Could you update the docs as well ?

The repository for docs: https://github.com/valkey-io/valkey-doc/

PR #1676, opened 2025-02-06
Fix kvstore overhead_hashtable_lut

zuiderkwast:
Fix calculation of overhead_hashtable_lut in kvstore when the flag `KVSTORE_ALLOCATE_HASHTABLES_ON_DEMAND` is *not* given.

The memory overhead is incremented when a hashtable is created, but in kvstoreCreate, we reset the sum after creating the tables.

Removed a bunch of fields that are initialized to zero by zcalloc already. Moved some other initialization to before creating the hashtables (not strictly necessary but more logical to init stuff before creating stuff that may modify these numbers).


rainsupreme (approved):
LGTM! Hopefully this fixes that UT issue? 😁

enjoy-binbin (approved):

PR #1677, opened 2025-02-06
Save config file after replica migration and brocast a PONG when role changed

enjoy-binbin:
This does some cleanup around clusterSetPrimary:
1. In replica migration, the replica already has a new primary, we
need to update the config file.
2. When the role changed, we need to broadcast a PONG in the cluster
to notify the change. This is a follow up for #1295.


zuiderkwast (approved):
Makes sense to me.

PR #1678, opened 2025-02-06
Add reply schema for LATENCY-LATESTS new fields

enjoy-binbin:
These fields was added in #224.

### DIFF
--- a/src/commands/latency-latest.json
+++ b/src/commands/latency-latest.json
@@ -20,11 +20,11 @@
         ],
         "reply_schema": {
             "type": "array",
-            "description": "An array where each element is a four elements array representing the event's name, timestamp, latest and all-time latency measurements.",
+            "description": "An array where each element is a n elements array representing the event name and information about the event latency.",
             "items": {
                 "type": "array",
-                "minItems": 4,
-                "maxItems": 4,
+                "minItems": 6,
+                "maxItems": 6,
                 "items": [
                     {
                         "type": "string",
--- a/src/commands/latency-latest.json
+++ b/src/commands/latency-latest.json
@@ -20,7 +20,7 @@
         ],
         "reply_schema": {
             "type": "array",
-            "description": "An array where each element is a n elements array representing the event name and information about the event latency.",
+            "description": "An array where each element is an array representing the event name, timestamp, latest and all-time latency measurements.",
             "items": {
                 "type": "array",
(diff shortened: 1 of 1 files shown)


enjoy-binbin:
This also fix the reply daily:
```
JSON schema validation error on /home/runner/work/valkey/valkey/tests/tmp/server.7213.155/stdout.reqres: ['command', 1738801393, 716, 716, 716, 1] is too long

Failed validating 'maxItems' in schema['items']:
    {'items': [{'description': 'Event name.', 'type': 'string'},
               {'description': 'Timestamp.', 'type': 'integer'},
               {'description': 'Latest latency in milliseconds.',
                'type': 'integer'},
               {'description': 'Max latency in milliseconds.',
                'type': 'integer'}],
     'maxItems': 4,
     'minItems': 4,
     'type': 'array'}
```

src/commands/latency-latest.json
```diff
@@ -41,6 +41,14 @@
                     {
                         "type": "integer",
                         "description": "Max latency in milliseconds."
+                    },
+                    {
+                        "type": "integer",
+                        "description": "Sum of the latencies recorded in the time series for this event."
```
enjoy-binbin, line 47: Not sure if we will add history and since for the new fields. I see we did not do it in memory-stats

src/commands/latency-latest.json
```diff
@@ -20,11 +20,11 @@
         ],
         "reply_schema": {
             "type": "array",
-            "description": "An array where each element is a four elements array representing the event's name, timestamp, latest and all-time latency measurements.",
+            "description": "An array where each element is a n elements array representing the event name and information about the event latency.",
```
zuiderkwast: "n elements" doesn't mean much if we don't say what *n* is. We can just say *an array*?

"about the event latency" also doesn't say much. It could be also about EVENT HISTORY. I think "timestamp, latest and all-time latency measurements" is OK. Maximum, sum and count can all be "all-time measurements", right?

```suggestion
            "description": "An array where each element is an array representing the event name, timestamp, latest and all-time latency measurements.",
```

zuiderkwast (approved):

src/commands/latency-latest.json
zuiderkwast: We don't need it IMHO.

PR #1680, opened 2025-02-06
Increase the HZ to make test more reliable

madolson:
After introducing, https://github.com/valkey-io/valkey/pull/1387, we saw a significant increase in other spurious wakeups because of the client cron that was added, which affected the "instantaneous eventloops per second" metric (showing it higher than before". All I did was increase the server hz to get more samples and increase the target value. This seems to work more consistently now. I also removed retries since the instantaneous value isn't dependent on number of retries. 

Additionally, `assert_lessthan $value [expr $retries*22000]` makes no sense to me. The value is usually around 30-100us, since all it's doing is waking up and running a little bit of cron. The retries doesn't make much sense, since the retries don't impact the instantaneous value. I just removed the retries and left the 22k value for now, maybe valgrind is slow.

### DIFF
--- a/tests/unit/info.tcl
+++ b/tests/unit/info.tcl
@@ -337,7 +337,7 @@ start_server {tags {"info" "external:skip" "debug_defrag:skip"}} {
             r config set hz 100
             set retries 0
             for {set retries 1} {$retries < 4} {incr retries} {
-                after 1600 ;# Wait for 16 cron tick so that sample array is fulfilled
+                after 200 ;# Wait for at least 16 cron tick so that sample array is fulfilled
                 set value [s instantaneous_eventloop_cycles_per_sec]
                 if {$value > 0} break
             }
--- a/tests/unit/info.tcl
+++ b/tests/unit/info.tcl
@@ -334,9 +334,10 @@ start_server {tags {"info" "external:skip" "debug_defrag:skip"}} {
 
         test {stats: instantaneous metrics} {
             r config resetstat
+            r config set hz 100
             set retries 0
             for {set retries 1} {$retries < 4} {incr retries} {
-                after 1600 ;# hz is 10, wait for 16 cron tick so that sample array is fulfilled
+                after 1600 ;# Wait for 16 cron tick so that sample array is fulfilled
                 set value [s instantaneous_eventloop_cycles_per_sec]
                 if {$value > 0} break
             }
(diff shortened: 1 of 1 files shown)


tests/unit/info.tcl
```diff
@@ -334,21 +334,25 @@ start_server {tags {"info" "external:skip" "debug_defrag:skip"}} {
 
         test {stats: instantaneous metrics} {
             r config resetstat
+            r config set hz 100
             set retries 0
             for {set retries 1} {$retries < 4} {incr retries} {
-                after 1600 ;# hz is 10, wait for 16 cron tick so that sample array is fulfilled
+                after 1600 ;# Wait for 16 cron tick so that sample array is fulfilled
```
zuiderkwast: With hz 100, it's 160 cron ticks?

```suggestion
                after 1600 ;# Wait for 160 cron tick so that sample array is fulfilled
```

tests/unit/info.tcl
```diff
@@ -334,21 +334,25 @@ start_server {tags {"info" "external:skip" "debug_defrag:skip"}} {
...
                 if {$value > 0} break
             }
 
             assert_lessthan $retries 4
             if {$::verbose} { puts "instantaneous metrics instantaneous_eventloop_cycles_per_sec: $value" }
             assert_morethan $value 0
-            assert_lessthan $value [expr $retries*15] ;# default hz is 10
+            # Hz is configured to 100, so we expect a value of about 100, but in practice it will be lower
+            # because of imprecision in kernel wakeups, but we also have some other wakeups like clients cron.
+            assert_lessthan $value 100
```
zuiderkwast: If we expect *about 100*, shouldn't we accept some range above and below 100?

    assert {$value >= 80 and $value <= 120}

JimB123: I was looking at this test.  It wasn't clear to me why we're expecting `value` to be less than `retries * 15`.

Retries should normally be 1.  It's not clear to me that it would ever go to 2.

So why, when we wait for 1600 ms, do we expect the 100ms cron to fire less than 15 times?

tests/unit/info.tcl
madolson:
```suggestion
                after 200 ;# Wait for at least 16 cron tick so that sample array is fulfilled
```

tests/unit/info.tcl
madolson: I have no idea why there was a retry counter here, or why retry changed the expected value (theyre both instantaneous values). Going back through the history, I'm guessing the original test was flaky when the test was *very slow*.

tests/unit/info.tcl
madolson: Sorry, I accidentally edited your comment on my phone instead of replying to it. Yeah, I actually tested a little bit more and it wasn't actually getting filled, so waiting a bit longer.

zuiderkwast (approved):

tests/unit/info.tcl
JimB123: You want to fix the comment too?

zuiderkwast:
Still flaky. I got this failure right now:

```
*** [err]: stats: instantaneous metrics in tests/unit/info.tcl
Expected '194' to be less than '150' (context: type eval line 11 cmd {assert_lessthan $value 150} proc ::test)
```

https://github.com/valkey-io/valkey/actions/runs/13201529468/job/36854504972?pr=1685#step:5:6269

madolson:
> Expected '194' to be less than '150' (context: type eval line 11 cmd {assert_lessthan $value 150} proc ::test)

Must be missing something else than, because that is way higher than should be possible. EDIT: I'm totally missing something, I misunderstood how the clients cron work. The value should be less than 200.

zuiderkwast:
What's the formula for the clients cron? Let's add a code about it in this test case when you've figured it out.

PR #1682, opened 2025-02-07
Improved failover pause stats test stability

enjoy-binbin:
We don't drop the packets, so the failover may finish in a short time that we fail to check the paused fields.

### DIFF
--- a/tests/unit/pause.tcl
+++ b/tests/unit/pause.tcl
@@ -418,15 +418,35 @@ start_server {tags {"pause network"}} {
 
 start_cluster 1 1 {tags {"external:skip cluster pause network"}} {
     test "Test check paused info during the cluster failover in info clients" {
+        set CLUSTER_PACKET_TYPE_NONE -1
+        set CLUSTER_PACKET_TYPE_FAILOVER_AUTH_ACK 6
+
         assert_equal [s 0 paused_reason] "none"
         assert_equal [s 0 paused_actions] "none"
         assert_equal [s 0 paused_timeout_milliseconds] 0
 
+        # Let replica drop FAILOVER_AUTH_ACK so that the election won't
+        # get the enough votes and the election will time out.
+        R 1 debug drop-cluster-packet-filter $CLUSTER_PACKET_TYPE_FAILOVER_AUTH_ACK
         R 1 cluster failover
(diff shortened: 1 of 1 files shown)


enjoy-binbin:
https://github.com/valkey-io/valkey/actions/runs/13190340371/job/36821941102#step:6:6533
```
*** [err]: Test check paused info during the cluster failover in info clients in tests/unit/pause.tcl
Expected 'none' to be equal to 'failover_in_progress' (context: type eval line 9 cmd {assert_equal [s 0 paused_reason] "failover_in_progress"} proc ::test)
```

tests/unit/pause.tcl
```diff
@@ -418,15 +418,35 @@ start_server {tags {"pause network"}} {
...
         assert_equal [s 0 paused_timeout_milliseconds] 0
 
+        # Let replica drop FAILOVER_AUTH_ACK so that the election won't
+        # get the enough votes and the election will time out.
+        R 1 debug drop-cluster-packet-filter $CLUSTER_PACKET_TYPE_FAILOVER_AUTH_ACK
         R 1 cluster failover
         wait_for_log_messages 0 {"*Manual failover requested by replica*"} 0 10 1000
 
+        # Failover will definitely time out, so on the primary side we will pause for
+        # `CLUSTER_MF_TIMEOUT * CLUSTER_MF_PAUSE_MULT` this long.
```
enjoy-binbin, line 435: I have an idea for CLUSTER_MF_TIMEOUT to be a configurable config, what do you think? @madolson @PingXie @zuiderkwast I think an admin could configure how long a primary would be paused in the worst case scenario, such as a failover timeout (for whatever reason, votes or a bug).

madolson: >I have an idea for CLUSTER_MF_TIMEOUT to be a configurable config

Yeah, I think this is reasonable. We had some discussions internally because the `FAILOVER` command allows passing in a timeout, which I think is preferable to a config, but it's more difficult to do that in cluster mode. I would be ok with just a config.

madolson (approved):

zuiderkwast (approved):

tests/unit/pause.tcl
enjoy-binbin: good to know, i will process it in the next PR.

PR #1686, opened 2025-02-07
Add Cluster Bus Port out of range error message for Cluster Meet command

hwware:
Now for the cluster meet command, we first check the port and cport number, and then the function clusterStartHandshake() is called.
In the function clusterStartHandshake(), ip address is checked first and then port and cport range are checked.
Even port or cport range is out of bound, only error message "Invalid node address specified" is reported.
This is not correct.

In this PR,  I just make the port and cport check together before clusterStartHandshake() function. Thus the port number and range can be checked together.

One example:

Current behavior:

127.0.0.1:7000> cluster meet 10.21.96.98 65000
(error) ERR Invalid node address specified: 10.21.96.98:65000
127.0.0.1:7000> cluster meet 1928.2292.2983.3884.26622 6379
(error) ERR Invalid node address specified: 1928.2292.2983.3884.26622:6379

Whatever the wrong ip address or incorrect bus port number, user get the same error message.

New behavior:

127.0.0.1:7000> cluster meet 10.21.96.98 65000
(error) ERR Cluster bus port number is out of range
127.0.0.1:7000> cluster meet 1928.2292.2983.3884.26622 6379
(error) ERR Invalid node address specified: 1928.2292.2983.3884.26622:6379

User can get much clearer error message.

Related note pr: https://github.com/valkey-io/valkey-doc/pull/240

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -777,6 +777,33 @@ int isValidAuxString(char *s, unsigned int length) {
     return 1;
 }
 
+int verifyPortNumber(client *c, long long *port, long long *cport) {
+    if (getLongLongFromObject(c->argv[3], port) != C_OK) {
+        addReplyErrorFormat(c, "Invalid base port specified: %s", (char *)c->argv[3]->ptr);
+        return C_ERR;
+    }
+    if (*port <= 0 || *port > 65535) {
+        addReplyErrorFormat(c, "Port number is out of range");
+        return C_ERR;
+    }
+
+    if (c->argc == 5) {
+        if (getLongLongFromObject(c->argv[4], cport) != C_OK) {
+            addReplyErrorFormat(c, "Invalid bus port specified: %s", (char *)c->argv[4]->ptr);
+            return C_ERR;
+        }
+    } else {
+        *cport = *port + CLUSTER_PORT_INCR;
+    }
+
+    if (*cport <= 0 || *cport > 65535) {
+        addReplyErrorFormat(c, "Cport number is out of range");
+        return C_ERR;
+    }
+
+    return C_OK;
+}
+
 void clusterCommandMyId(client *c) {
     char *name = clusterNodeGetName(getMyClusterNode());
     if (name) {
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -777,7 +777,7 @@ int isValidAuxString(char *s, unsigned int length) {
     return 1;
 }
 
-int verifyPortNumber(client *c, long long *port, long long *cport) {
+int parseAndValidateConnectionPorts(client *c, long long *port, long long *cport) {
     if (getLongLongFromObject(c->argv[3], port) != C_OK) {
         addReplyErrorFormat(c, "Invalid base port specified: %s", (char *)c->argv[3]->ptr);
         return C_ERR;
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -777,33 +777,6 @@ int isValidAuxString(char *s, unsigned int length) {
     return 1;
 }
 
-int parseAndValidateConnectionPorts(client *c, long long *port, long long *cport) {
-    if (getLongLongFromObject(c->argv[3], port) != C_OK) {
-        addReplyErrorFormat(c, "Invalid base port specified: %s", (char *)c->argv[3]->ptr);
-        return C_ERR;
-    }
-    if (*port <= 0 || *port > 65535) {
-        addReplyErrorFormat(c, "Port number is out of range");
-        return C_ERR;
-    }
-
-    if (c->argc == 5) {
-        if (getLongLongFromObject(c->argv[4], cport) != C_OK) {
-            addReplyErrorFormat(c, "Invalid bus port specified: %s", (char *)c->argv[4]->ptr);
-            return C_ERR;
-        }
-    } else {
-        *cport = *port + CLUSTER_PORT_INCR;
-    }
-
-    if (*cport <= 0 || *cport > 65535) {
-        addReplyErrorFormat(c, "Cport number is out of range");
-        return C_ERR;
-    }
-
-    return C_OK;
-}
-
 void clusterCommandMyId(client *c) {
     char *name = clusterNodeGetName(getMyClusterNode());
     if (name) {
--- a/src/cluster.h
+++ b/src/cluster.h
@@ -133,5 +133,4 @@ int isNodeAvailable(clusterNode *node);
 long long getNodeReplicationOffset(clusterNode *node);
 sds aggregateClientOutputBuffer(client *c);
 void resetClusterStats(void);
-int parseAndValidateConnectionPorts(client *c, long long *port, long long *cport);
 #endif /* __CLUSTER_H */
--- a/src/cluster.h
+++ b/src/cluster.h
@@ -133,4 +133,5 @@ int isNodeAvailable(clusterNode *node);
 long long getNodeReplicationOffset(clusterNode *node);
 sds aggregateClientOutputBuffer(client *c);
 void resetClusterStats(void);
+int verifyPortNumber(client *c, long long *port, long long *cport);
 #endif /* __CLUSTER_H */
--- a/src/cluster.h
+++ b/src/cluster.h
@@ -133,5 +133,5 @@ int isNodeAvailable(clusterNode *node);
 long long getNodeReplicationOffset(clusterNode *node);
 sds aggregateClientOutputBuffer(client *c);
 void resetClusterStats(void);
-int verifyPortNumber(client *c, long long *port, long long *cport);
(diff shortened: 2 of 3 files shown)


src/cluster.c
```diff
@@ -778,6 +778,33 @@ int isValidAuxString(char *s, unsigned int length) {
     return 1;
 }
 
+int verifyPortNumber(client *c, long long *port, long long *cport) {
```
xbasel: Is there a reason this function is in cluster.c/h instead of static in cluster_legacy.c? It seems to be used only in one place.

src/cluster.h
```diff
@@ -134,4 +134,5 @@ int isNodeAvailable(clusterNode *node);
 long long getNodeReplicationOffset(clusterNode *node);
 sds aggregateClientOutputBuffer(client *c);
 void resetClusterStats(void);
+int verifyPortNumber(client *c, long long *port, long long *cport);
```
xbasel: This function does more than just verification, it also parses and assigns port values. Consider renaming it to something like `parseAndValidatePorts`  for better clarity.

src/cluster.c
hwware: The function resetClusterStats() , isNodeAvailable() , clusterNodeIsMyself() are declared in cluster.h, and defined in cluster.c, I just follow the rule.

src/cluster.h
hwware: It makes sense to me. Will update.

src/cluster.c
```diff
@@ -778,6 +778,33 @@ int isValidAuxString(char *s, unsigned int length) {
     return 1;
 }
 
+int parseAndValidateConnectionPorts(client *c, long long *port, long long *cport) {
+    if (getLongLongFromObject(c->argv[3], port) != C_OK) {
+        addReplyErrorFormat(c, "Invalid base port specified: %s", (char *)c->argv[3]->ptr);
```
zuiderkwast: This function has a strong coupling with the CLUSTER MEET command. The `c->argv[3]` and `c->argv[4]` are hard-coded to match exactly the args of that command. It is actually a helper function that's only usable for custer meet.

I think it's better just keep this code directly in clusterCommandSpecial where cluster meet is handled. If we keep it there, it is clear from the surrounding context what argv[3] and argv[4] are.

We can do some other simplification of the validation, like function to parse and validate a port, such as `int getPortFromObject(robj *arg, int *port)` that just parses the number and does a port range check.

src/cluster_legacy.c
```diff
@@ -2197,12 +2197,6 @@ static int clusterStartHandshake(char *ip, int port, int cport) {
         return 0;
     }
 
-    /* Port sanity check */
-    if (port <= 0 || port > 65535 || cport <= 0 || cport > 65535) {
-        errno = EINVAL;
-        return 0;
-    }
-
```
zuiderkwast, line 2205: It's good that you move this to CLUSTER MEET. I agree. :+1:

zuiderkwast:
The behavior change is that CLUSTER MEET can return an error instead of OK when the ports are invalid. This means the user can see the error early and understand what's wrong.

This is the user-visible change, so it should be the title of the PR I think. The rest is just refactoring which is not important for users and release notes.

enjoy-binbin (commented):
why we need this? The old code also seems to be very easy to understand and maintain

hwware:
> why we need this? The old code also seems to be very easy to understand and maintain

Current old cause developers hard to identify the problem.  Existing logic is to check port,, and then in function clusterStartHandshake check the ip first and check the port again.  

The easier way to put all port checking logic together, and then check the ip.

zuiderkwast:
@enjoy-binbin @hwware The first commit is smaller. It just moves the check from clusterStartHandshake to CLUSTER MEET so it can return an error, without extra refactoring.

https://github.com/valkey-io/valkey/pull/1686/commits/ce4c0d4bbd5b528bc4273a41f55c08f0c75a871a

hwware:
> @enjoy-binbin @hwware The first commit is smaller. It just moves the check from clusterStartHandshake to CLUSTER MEET so it can return an error, without extra refactoring.
> 
> [ce4c0d4](https://github.com/valkey-io/valkey/commit/ce4c0d4bbd5b528bc4273a41f55c08f0c75a871a)

Yes, I just make thing easier. It looks like i update codes to the fist commit, and check the port and cport in the Cluster Meet part. 
Sometimes simple is good enough

src/cluster_legacy.c
```diff
@@ -6911,6 +6909,11 @@ int clusterCommandSpecial(client *c) {
             cport = port + CLUSTER_PORT_INCR;
         }
 
+        if (cport <= 0 || cport > 65535) {
+            addReplyErrorFormat(c, "Cport number is out of range");
```
zuiderkwast: Just thinking about this error message. In the docs, we called it cluster-bus-port:

    CLUSTER MEET ip port [cluster-bus-port]

And in the other error for this field uses the word "bus port", when getLongLongFromObject fails:

    addReplyErrorFormat(c, "Invalid bus port specified: %s", (char *)c->argv[4]->ptr);

What do you think about this:

```suggestion
            addReplyErrorFormat(c, "Cluster bus port number is out of range");
```

zuiderkwast (commented):
Yes, this is simpler. I like it more. @enjoy-binbin WDYT?

I have just a minor comment about the error message.

src/cluster_legacy.c
hwware: Agree, we should match with the doc.

zuiderkwast (approved):
LGTM.

Before merge, please update the PR title to clarify the user-visible behavior, that CLUSTER MEET validates the ports and returns error for invalid ports.

That is what we should mention in the release notes.

enjoy-binbin:
> The first commit is smaller. It just moves the check from clusterStartHandshake to CLUSTER MEET so it can return an error, without extra refactoring.

we do return an error in this check.
```
        if (clusterStartHandshake(c->argv[2]->ptr, port, cport) == 0 && errno == EINVAL) {
            addReplyErrorFormat(c, "Invalid node address specified: %s:%s", (char *)c->argv[2]->ptr,
                                (char *)c->argv[3]->ptr);
        } else {
            sds client = catClientInfoShortString(sdsempty(), c, server.hide_user_data_from_log);
            serverLog(LL_NOTICE, "Cluster meet %s:%lld (user request from '%s').", (char *)c->argv[2]->ptr, port,
                      client);
            sdsfree(client);
            addReply(c, shared.ok);
        }
```

> Current old cause developers hard to identify the problem. Existing logic is to check port,, and then in function clusterStartHandshake check the ip first and check the port again.

how so... they are doing different thing, one is extract the number and check it range, the other is check the port range. `clusterStartHandshake` can has it own check i think, the code is pretty simple why it would casue the trouble to the developers.

hwware:
> > The first commit is smaller. It just moves the check from clusterStartHandshake to CLUSTER MEET so it can return an error, without extra refactoring.
> 
> we do return an error in this check.
> 
> ```
>         if (clusterStartHandshake(c->argv[2]->ptr, port, cport) == 0 && errno == EINVAL) {
>             addReplyErrorFormat(c, "Invalid node address specified: %s:%s", (char *)c->argv[2]->ptr,
>                                 (char *)c->argv[3]->ptr);
>         } else {
>             sds client = catClientInfoShortString(sdsempty(), c, server.hide_user_data_from_log);
>             serverLog(LL_NOTICE, "Cluster meet %s:%lld (user request from '%s').", (char *)c->argv[2]->ptr, port,
>                       client);
>             sdsfree(client);
>             addReply(c, shared.ok);
>         }
> ```
> 
> > Current old cause developers hard to identify the problem. Existing logic is to check port,, and then in function clusterStartHandshake check the ip first and check the port again.
> 
> how so... they are doing different thing, one is extract the number and check it range, the other is check the port range. `clusterStartHandshake` can has it own check i think, the code is pretty simple why it would casue the trouble to the developers.

Through this PR, user experience is much clearer and easy to identify where the problem is, for example:

Current behavior:

127.0.0.1:7000> cluster meet 10.21.96.98 65000
**(error) ERR Invalid node address specified: 10.21.96.98:65000**
127.0.0.1:7000> cluster meet 1928.2292.2983.3884.26622 6379
(error) ERR Invalid node address specified: 1928.2292.2983.3884.26622:6379

Whatever the wrong ip address or incorrect bus port number, user get the same error message.


New behavior:

127.0.0.1:7000> cluster meet 10.21.96.98 65000
**(error) ERR Cluster bus port number is out of range**
127.0.0.1:7000> cluster meet 1928.2292.2983.3884.26622 6379
(error) ERR Invalid node address specified: 1928.2292.2983.3884.26622:6379

User can get much clearer error message.

PR #1687, opened 2025-02-07
Increase threshold for info.tcl even higher

madolson:
It's possible for the clientCron and the serverCron to get out of sync, so that even though they are both at 100hz, they could be firing independently. (e.g. server cron fires on 5, 15, 25ms and client cron fires at 0, 10, 20ms) So it's actually possible for 2x the number of crons.


hpatro (approved):

hwware (approved):
This change makes sense.

hpatro:
Bunch of different failures unrelated to this PR. merged it.

madolson:
Thanks, yeah, I ran it some more times on my private branch and I think this is the right threshold.

PR #1690, opened 2025-02-08
Add cluster-manual-failover-timeout to configure the timeout for manual failover

enjoy-binbin:
Allows cluster admins to configure the cluster manual failover timeout as
needed, admins can configure how long a primary would be paused in the
worst case scenario such as a failover timed out due to the insufficient
votes.

The configuration name is cluster-manual-failover-timeout, the unit is
milliseconds, and the range is [1, INT_MAX]ms.


valkey.conf
```diff
@@ -1841,6 +1841,38 @@ aof-timestamp-enabled no
...
+#    sure it processed all the data from the primary before it continues.
+# 4. The replica starts a failover, obtains a new configuration epoch from the
+#    majority of the masters, and broadcasts the new configuration.
+# 5. The old primary receives the configuration update: unpause its clients and
+#    starts replying with redirection messages so that they'll continue the chat
+#    with the new primary.
+#
+# This way clients are moved away from the old primary to the new primary atomically
+# and only when the replica that is turning into the new primary has processed
+# all of the replication stream from the old primary.
```
enjoy-binbin: i took this from CLUSTER FAILOVER command docs, let me know if it is too much, please also feel free to insert a suggestion if you have a good wording.

src/server.h
```diff
@@ -2085,6 +2085,7 @@ struct valkeyServer {
     unsigned long cluster_blacklist_ttl;                   /* Duration in seconds that a node is denied re-entry into
                                                             * the cluster after it is forgotten with CLUSTER FORGET. */
     int cluster_slot_stats_enabled;                        /* Cluster slot usage statistics tracking enabled. */
+    long long cluster_mf_timeout;                          /* Milliseconds to do a manual failover. */
```
hpatro:
```suggestion
    mstime_t cluster_mf_timeout;                          /* Milliseconds to do a manual failover. */
```

valkey.conf
hpatro: Could we also mention what's the behavior if it's set to 0, is it indefinite timeout?

enjoy-binbin: i think we should set a min and max value of it (that is not allow 0), but i am not sure about the value, do you have some ideas?

madolson: I agree, I don't think a value of 0 makes sense since there is no way to abort (afaik, you might be able to trigger another failover). I would vote a minimum time of like 1000ms.

valkey.conf
```diff
@@ -1841,6 +1841,38 @@ aof-timestamp-enabled no
...
+# a safe way, without any window for data loss.
+#
+# It works in the following way:
+# 1. The replica tells the primary to stop processing queries from clients.
+# 2. The primary replies to the replica with the current replication offset.
+# 3. The replica waits for the replication offset to match on its side, to make
+#    sure it processed all the data from the primary before it continues.
+# 4. The replica starts a failover, obtains a new configuration epoch from the
+#    majority of the masters, and broadcasts the new configuration.
+# 5. The old primary receives the configuration update: unpause its clients and
```
```suggestion
# 5. The old primary receives the configuration update: unpauses its clients and
```

valkey.conf
enjoy-binbin: do we also want a max value? The big value also does not makes senese and we will hit the overflow issue when doing this `now + (server.cluster_mf_timeout * CLUSTER_MF_PAUSE_MULT)` if someone test it.

madolson: I don't have a strong opinion, but agree a high value is bad since it's not interruptible. Maybe like a minute? (Seems like such a small range, maybe minimum should be lower, like 250ms)

enjoy-binbin: [250, 60000] SGTM, i changed to it.

madolson (approved):
Nice!

@valkey-io/core-team Ok, another small incremental improvement for 8.1. I would suggest just reading the valkey.conf update. I'll add it the weekly meeting as well.

valkey.conf
```diff
@@ -1841,6 +1841,38 @@ aof-timestamp-enabled no
...
+# of its replicas (which is the node we send the CLUSTER FAILOVER command to) in
+# a safe way, without any window for data loss.
+#
+# It works in the following way:
+# 1. The replica tells the primary to stop processing queries from clients.
+# 2. The primary replies to the replica with the current replication offset.
+# 3. The replica waits for the replication offset to match on its side, to make
+#    sure it processed all the data from the primary before it continues.
+# 4. The replica starts a failover, obtains a new configuration epoch from the
+#    majority of the masters, and broadcasts the new configuration.
```
zuiderkwast: masters -> primaries

Should we document the complete procedure here in the config file? It is documented in the docs already. We could refer to the docs. I don't mind that we include this here if we keep it short and relevant for the config.

enjoy-binbin:
```suggestion
#    majority of the primaries, and broadcasts the new configuration.
```

PingXie: I would vote to keep it simple. valkey.conf is end user facing, so the inner workings aren't that relevant in this file.

PingXie (commented):
`cluster-manual-failover-timeout` works for me. thanks @enjoy-binbin!

madolson:
Core team discussion: We would like to be able to increase the max configuration, but there is no cancel command today. If a user were to set the value too high, it could cause a prolonged failure. Instead of implementing a new `FAILOVER CANCEL` command, we will just allow setting the config to be between 1 and "INT_MAX".

valkey.conf
enjoy-binbin: i kind of like this, the conf itself describe the relevant content, usually people may not jump to the doc web, like viewing the configuration items on a terminal.

But i don't have a strong opinion on this.

zuiderkwast: I think it's a problem to have duplicated text. If we fix some typo in one of them, we still have the other copy somewhere and it can be hard to keep both of them in sync. I have seen it before. For example, until some months ago we had a file called `TLS.md` in this repo and `topics/encryption.md` in the doc repo. These files were almost identical, but some new info was added in one of them and some other info was added in the other one.

The config is just a timeout. The caller doesn't need the full procedure to set the timeout. A short description that the replica and the primary wait for each other for some time is useful, but other than that, refer to the CLUSTER FAILOVER command manual please.

madolson (approved):

valkey.conf
madolson: Longer term, do we want a single source of truth building out both? We could build some JSON files that document what this configs do, then attach it both to the relevant man pages, website pages, and incorporate it into valkey.conf file.

enjoy-binbin:
@madolson I see we added a needs-doc-pr label, please check this to see if it satisfies https://github.com/valkey-io/valkey-doc/pull/238

zuiderkwast (approved):

soloestoy (approved):

hwware (approved):

PR #1691, opened 2025-02-09
Fix log message "DB 9: 100 keys (0 volatile) in 16 slots HT"

zuiderkwast:
With verbose logging, this message is logged every 5 seconds.

Before #1186, there were 128 slots in a database with 100 keys. Slots here means the number of entries that can be stored on the top-level of a hash table, which in the dict was the same as the number of buckets. With the new hash table, there are multiple entries per bucket, so the number of buckets is not a useful metric to show.

This PR multiplies the number of buckets with the number of entry positions per bucket. With this change, the log message is instead

    36236:M 09 Feb 2025 15:47:18.923 - DB 9: 100 keys (0 volatile) in 112 slots HT.

### DIFF
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -1104,6 +1104,11 @@ size_t hashtableChainedBuckets(hashtable *ht, int table) {
     return ht->child_buckets[table];
 }
 
+/* Returns the number of entry positions per bucket. */
+unsigned hashtableEntriesPerBucket(void) {
+    return ENTRIES_PER_BUCKET;
+}
+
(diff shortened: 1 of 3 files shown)


ranshid:
@zuiderkwast , not that I have issues with this fix. but why is this information important in any way??? I get the number of keys and volatile keys is somewhat important, but the hashtable size? it is so implementation dependent even for future changes to the hashtable and the number can change whenever we might decide to change the load factor right?

ranshid (approved):

PR #1694, opened 2025-02-09
valkey-cli: re-select previously selected database after reconnect

### LINKED ISSUE #1693: [BUG] Valkey-cli Incorrect database selection after connection drop
When the connection is dropped, valkey-cli defaults back to `DB0` upon reconnection, but the prompt still displays the previously selected database. This leads to misleading behavior where commands appear to be executed in the wrong database.

```
127.0.0.1:6379> info keyspace
# Keyspace
127.0.0.1:6379> select 9
OK
127.0.0.1:6379[9]> ping
Error: Server closed the connection
not connected> set x 1
OK
127.0.0.1:6379[9]> info keyspace
# Keyspace
db0:keys=1,expires=0,avg_ttl=0
127.0.0.1:6379[9]>
```

**To reproduce**

Open valkey-cli
```
valkey-cli -n 9
```
then restart the server and run in the previously opened `valkey-cli`:
```
set x 1
```
The prompt will show the selected db 9, but it will write to db0. 

**Expected behavior**

Either the previously selected database should be reselected, or the prompt should reflect DB0. The later is less desirable if the client explicitly started `valkey-cli` with the parameter `-n`.

xbasel:
DB0 is selecetd after reconnection regardless of the previously selected database. The prompt show the previously selected databse while working with DB0, which is misleading.
The code has been chanegd to reselect the requested database.

Fixes #1693

### DIFF
--- a/src/valkey-cli.c
+++ b/src/valkey-cli.c
@@ -1610,17 +1610,23 @@ static int cliSwitchProto(void) {
     return result;
 }
 
+static void resetConfig(void) {
+    config.dbnum = 0;
+    config.in_multi = 0;
+    config.pubsub_mode = 0;
+}
+
 /* Connect to the server. It is possible to pass certain flags to the function:
  *      CC_FORCE: The connection is performed even if there is already
  *                a connected socket.
  *      CC_QUIET: Don't print errors if connection fails. */
 static int cliConnect(int flags) {
+    resetConfig();
+
     if (context == NULL || flags & CC_FORCE) {
         if (context != NULL) {
             redisFree(context);
-            config.dbnum = 0;
-            config.in_multi = 0;
-            config.pubsub_mode = 0;
+            resetConfig();
             cliRefreshPrompt();
         }
 
--- a/src/valkey-cli.c
+++ b/src/valkey-cli.c
@@ -1621,9 +1621,8 @@ static void resetConfig(void) {
  *                a connected socket.
  *      CC_QUIET: Don't print errors if connection fails. */
 static int cliConnect(int flags) {
-    resetConfig();
-
     if (context == NULL || flags & CC_FORCE) {
+        resetConfig();
         if (context != NULL) {
             redisFree(context);
             resetConfig();
--- a/tests/integration/valkey-cli.tcl
+++ b/tests/integration/valkey-cli.tcl
@@ -69,16 +69,6 @@ start_server {tags {"cli logreqres:skip"}} {
         set _ [format_output [read_cli $fd]]
     }
 
-    # Extract a specific field's value from a CLIENT INFO string
-    proc get_client_info_field {client_string field} {
-        set pattern [format {%s=([^ ]+)} $field]
-        if {[regexp -nocase $pattern $client_string match value]} {
-            return $value
-        }
-        return "Field '$field' not found"
-    }
-
-
     proc test_interactive_cli {name code} {
         set ::env(FAKETTY) 1
         set fd [open_cli]
@@ -845,14 +835,11 @@ if {!$::tls} { ;# fake_redis_node doesn't support TLS
 
     test "valkey-cli make sure selected db survives connection drops" {    
         set fd [open_cli]
-        
-        # Target database
-        set db 9
-
+                
         # Select the database        
-        assert_equal "OK" [run_command $fd "select $db"]
+        assert_equal "OK" [run_command $fd "select $::dbnum"]
 
-        # Kill all normal clients, to disconnect vlakey-cli client
+        # Kill all normal clients, to disconnect valkey-cli client
         r client kill type normal
         after 10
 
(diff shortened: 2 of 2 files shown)


xbasel:
Fixes issue https://github.com/valkey-io/valkey/issues/1693

enjoy-binbin (commented):
i think we can do a re-select in here? after the reconnect
```diff
static int issueCommandRepeat(int argc, char **argv, long repeat) {
    /* In Lua debugging mode, we want to pass the "help" to the server to get
     * it's own HELP message, rather than handle it by the CLI, see ldbRepl.
     *
     * For the normal server HELP, we can process it without a connection. */
    if (!config.eval_ldb && (!strcasecmp(argv[0], "help") || !strcasecmp(argv[0], "?"))) {
        cliOutputHelp(--argc, ++argv);
        return REDIS_OK;
    }

    while (1) {
        if (config.cluster_reissue_command || context == NULL || context->err == REDIS_ERR_IO ||
            context->err == REDIS_ERR_EOF) {
            if (cliConnect(CC_FORCE) != REDIS_OK) {
                cliPrintContextError();
                config.cluster_reissue_command = 0;
                return REDIS_ERR;
            }
+           reset config
+           cliSelect();
        }
```

xbasel:
> i think we can do a re-select in here? after the reconnect
> 
> ```diff
> static int issueCommandRepeat(int argc, char **argv, long repeat) {
>     /* In Lua debugging mode, we want to pass the "help" to the server to get
>      * it's own HELP message, rather than handle it by the CLI, see ldbRepl.
>      *
>      * For the normal server HELP, we can process it without a connection. */
>     if (!config.eval_ldb && (!strcasecmp(argv[0], "help") || !strcasecmp(argv[0], "?"))) {
>         cliOutputHelp(--argc, ++argv);
>         return REDIS_OK;
>     }
> 
>     while (1) {
>         if (config.cluster_reissue_command || context == NULL || context->err == REDIS_ERR_IO ||
>             context->err == REDIS_ERR_EOF) {
>             if (cliConnect(CC_FORCE) != REDIS_OK) {
>                 cliPrintContextError();
>                 config.cluster_reissue_command = 0;
>                 return REDIS_ERR;
>             }
> +           reset config
> +           cliSelect();
>         }
> ```

`cliSelect()` is already invoked in `cliConnect()`:
https://github.com/valkey-io/valkey/blob/unstable/src/valkey-cli.c#L1670

src/valkey-cli.c
```diff
@@ -1610,17 +1610,23 @@ static int cliSwitchProto(void) {
...
+
 /* Connect to the server. It is possible to pass certain flags to the function:
  *      CC_FORCE: The connection is performed even if there is already
  *                a connected socket.
  *      CC_QUIET: Don't print errors if connection fails. */
 static int cliConnect(int flags) {
+    resetConfig();
+
     if (context == NULL || flags & CC_FORCE) {
         if (context != NULL) {
```
zuiderkwast: If there's already a connection and FORCE is not given, we don't want to reset the config, do we?

```suggestion
static int cliConnect(int flags) {
    if (context == NULL || flags & CC_FORCE) {
        resetConfig();
        if (context != NULL) {
```

tests/integration/valkey-cli.tcl
```diff
@@ -833,6 +843,37 @@ if {!$::tls} { ;# fake_redis_node doesn't support TLS
...
+    test "valkey-cli make sure selected db survives connection drops" {    
+        set fd [open_cli]
+        
+        # Target database
+        set db 9
+
+        # Select the database        
+        assert_equal "OK" [run_command $fd "select $db"]
+
+        # Kill all normal clients, to disconnect vlakey-cli client
```
zuiderkwast: vlakey? typo

tests/integration/valkey-cli.tcl
```diff
@@ -833,6 +843,37 @@ if {!$::tls} { ;# fake_redis_node doesn't support TLS
         close_cli $fd
     }
 
+    test "valkey-cli make sure selected db survives connection drops" {    
+        set fd [open_cli]
+        
+        # Target database
+        set db 9
```
zuiderkwast: We already have `$::dbnum` which depends on `$::singledb`. See the first lines in this file.

tests/integration/valkey-cli.tcl
```diff
@@ -833,6 +843,37 @@ if {!$::tls} { ;# fake_redis_node doesn't support TLS
...
+
+        # Select the database        
+        assert_equal "OK" [run_command $fd "select $db"]
```
zuiderkwast: Fails in cluster mode?

tests/integration/valkey-cli.tcl
```diff
@@ -833,6 +843,37 @@ if {!$::tls} { ;# fake_redis_node doesn't support TLS
...
+        
+        set client_info [run_command $fd "client info"]
+
+        # set actual_db [get_client_info_field $client_info "db"]
+
+        # # Validate that the selected DB matches the reported DB
+        # assert_equal $db $actual_db
+        
+        # r flushall
+        # r select 0
```
zuiderkwast: Unused stuff. Delete it?

Btw, the reason `get_client_info_field` doesn't already exist is that other test cases just do it this way:

```tcl
regexp {db=(\d+)} $client_info _ actual_db
```

zuiderkwast (commented):
Looks good in general.

If you write `Fixes #xxxx` or `Closes #xxxx` in the PR top comment, it links the PR to the issue closes the issue when merging. It's useful.

This PR obsoletes #1382. I prefer this new PR because it includes a test case.

tests/integration/valkey-cli.tcl
xbasel: A typing race condition. :)

tests/integration/valkey-cli.tcl
xbasel: `cluster:skip` tag is set. but now it won't be needed as $::dbnum will be used (`select 0` should work in cluster mode).

zuiderkwast (approved):

PR #1695, opened 2025-02-10
Update availability-zone conf comment to mention hello command

enjoy-binbin:
We add it to HELLO response in #1487.

### DIFF
--- a/valkey.conf
+++ b/valkey.conf
@@ -2513,7 +2513,7 @@ jemalloc-bg-thread yes
 # ignore-warnings ARM64-COW-BUG
 
 # Inform Valkey of the availability zone if running in a cloud environment.  Currently
-# this is only exposed via the info command for clients to use, but in the future we
-# we may also use this when making decisions for replication.
+# this is only exposed via the info/hello commands for clients to use, but in the future we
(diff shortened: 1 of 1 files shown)


valkey.conf
```diff
@@ -2513,7 +2513,7 @@ jemalloc-bg-thread yes
 # ignore-warnings ARM64-COW-BUG
 
 # Inform Valkey of the availability zone if running in a cloud environment.  Currently
-# this is only exposed via the info command for clients to use, but in the future we
-# we may also use this when making decisions for replication.
+# this is only exposed via the info/hello commands for clients to use, but in the future we
+# we may also use this when making decisions for replication. Default is a empty string.
```
zuiderkwast: Let's write INFO and HELLO in uppercase. It's more clear.

I don't really understand "in the future we may also use this when making decisions for replication". I don't like documentation about the future for features that maybe will never happen. Let's delete that?

```suggestion
# this is exposed in the INFO and HELLO commands for clients to use. Default is
# the empty string.
```

enjoy-binbin: SGTM.

zuiderkwast (approved):

PR #1698, opened 2025-02-10
show capa in CLIENT LIST / CLIENT INFO

soloestoy:


src/networking.c
```diff
@@ -3382,6 +3382,10 @@ sds catClientInfoString(sds s, client *client, int hide_user_data) {
     }
     *p = '\0';
 
+    p = capa;
+    if (client->capa & CLIENT_CAPA_REDIRECT) *p++ = 'r';
```
xbasel, line 3386: I think it would be useful to have a map like this next to the CAPA constants to do this automatically rather than adding each CAPA manually:
```
capa_map = {
    {CLIENT_CAPA_REDIRECT, 'R'},
    {CLIENT_CAPA_???, 'Z'},
};
```

xbasel:
What's the use-case?

zuiderkwast (approved):
I accept. :+1:

src/networking.c
madolson: What is the explicit value, this seems somewhat OK to me.

tests/integration/replica-redirect.tcl
```diff
@@ -46,6 +46,10 @@ start_server {tags {needs:repl external:skip}} {
             assert_error "REDIRECT $primary_host:$primary_port" {r get foo}
         }
 
+        test {CLIENT INFO} {
+            r client info
+        } {id=* addr=*:* laddr=*:* fd=* name=* age=* idle=* flags=N capa=r db=* sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=* argv-mem=* multi-mem=0 rbs=* rbp=* obl=0 oll=0 omem=0 tot-mem=* events=r cmd=client|info user=* redir=-1 resp=* lib-name=* lib-ver=* tot-net-in=* tot-net-out=* tot-cmds=*}
```
madolson, line 51:
```suggestion
        } {id=* addr=*:* laddr=*:* fd=* name=* age=* idle=* flags=N db=* sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=* argv-mem=* multi-mem=0 rbs=* rbp=* obl=0 oll=0 omem=0 tot-mem=* events=r cmd=client|info user=* redir=-1 resp=* lib-name=* lib-ver=* tot-net-in=* tot-net-out=* tot-cmds=* capa=r}
```

Zhoa: "I was trying to keep similar fields together"
Madelyn: I think adding field at the end is the safest.

madolson (approved):

enjoy-binbin (approved):

PR #1701, opened 2025-02-10
Scripting Engine Debugger Support

rjd15372:
This PR introduces the support for implementing remote debuggers in scripting engines modules.

The module API is extended with scripting engines callbacks and new functions that can be used by scripting engine modules to implement a remote debugger.

Most of the code that was used to implement the Lua debugger, was refactored and moved to the `scripting_engine.c` file, and only the code specific to the Lua engine, remained in the `debug_lua.c` file.

The `SCRIPT DEBUG (YES|NO|SYNC)` command was extend with an optional parameter that can be used to specify the engine name, where we want to enable the debugger. If no engine name is specified, the Lua engine is used to keep backwards compatibility.

In [src/valkeymodule.h](https://github.com/valkey-io/valkey/pull/1701/files#diff-b91520205c29a3a5a940786e509b2f13a5e73a1ac2016be773e62ea64c7efb28) we see the module API changes. And in the `helloscripting.c` file we can see how to implement a simple debugger for the dummy HELLO scripting engine.


src/eval.c
```diff
@@ -609,19 +609,42 @@ void scriptCommand(client *c) {
...
+                addReplyErrorFormat(c, "No scripting engine found with name '%s' to enable debug", (const char *)engine_name);
+                return;
+            }
+        } else {
+            en = scriptingEngineManagerFind("lua");
+            if (en == NULL) {
+                addReplyError(c, "No scripting engine found with name 'lua' to enable debug");
+                return;
+            }
+        }
```
zuiderkwast: Unify these?

```Suggestion
        const char *engine_name = c->argc == 4 ? c->argv[3]->ptr : "lua";
        scriptingEngine *en = scriptingEngineManagerFind(engine_name);
        if (en == NULL) {
            addReplyErrorFormat(c, "No scripting engine found with name '%s' to enable debug", engine_name);
            return;
        }
```

madolson:
Postponing until 9.0.

zuiderkwast (commented):
I had this pending comment since forever. Posting it now.

We talked about this in a core team meeting. We said that nobody uses the Lua debugger and that we could basically deprecate it. No meed to make this large change to extend the debugger support. The main reason to do it is to avoid breaking it if we move Lua to a module?

rjd15372:
> We talked about this in a core team meeting. We said that nobody uses the Lua debugger and that we could basically deprecate it. No meed to make this large change to extend the debugger support. The main reason to do it is to avoid breaking it if we move Lua to a module?

Correct. This is required if we want to keep the debugger feature for Lua when moving Lua to a module.
I'm also fine to deprecate the debugger feature, but if that is the case we should deprecate it in 9.0 and raise somekind of warning to the user when he tries to use the debugger.

src/module.c
```diff
@@ -13211,6 +13211,27 @@ ValkeyModuleScriptingEngineExecutionState VM_GetFunctionExecutionState(
     return ret == SCRIPT_CONTINUE ? VMSE_STATE_EXECUTING : VMSE_STATE_KILLED;
 }
 
+void VM_ScriptingEngineDebuggerLog(ValkeyModuleString *msg, int truncate) {
```
zuiderkwast, line 13513: These need proper markdown comments. It's extracted and used for the API reference manual.

src/valkey-cli.c
```diff
@@ -3469,12 +3469,27 @@ static int evalMode(int argc, char **argv) {
         }
         fclose(fp);
 
+        char *engine_name = NULL;
+        if (script[0] == '#' && script[1] == '!') {
+            const char *sp = strchr(script, '\n');
+            engine_name = strndup(script + 2, (sp - script) - 2);
```
zuiderkwast, line 3539: Engine name is only the first word in the shebang. There can be some parameters after the engine name, separated by a space. Not sure if we need to accept more characters like \r, \t....

`strpbrk(script, "\r\n ")`

src/scripting_engine.c
```diff
@@ -309,3 +310,749 @@ engineMemoryInfo scriptingEngineCallGetMemoryInfo(scriptingEngine *engine,
     engineTeardownModuleCtx(engine);
     return mem_info;
 }
+
+debuggerEnableRet scriptingEngineCallDebuggerEnable(scriptingEngine *engine,
+                                                    subsystemType type,
+                                                    const debuggerCommand **commands,
+                                                    size_t *commands_len) {
+    if (engine->impl.methods.debugger_enable == NULL ||
```
zuiderkwast, line 412: Need to check ABI version before accessing these fields. Otherwise it may crash when running a module compiled with valkeymodule.h from Valkey 8.1.

    if (engine->impl.methods.version >= 2) ...

src/valkeymodule.h
```diff
@@ -979,6 +979,126 @@ typedef ValkeyModuleScriptingEngineMemoryInfo (*ValkeyModuleScriptingEngineGetMe
...
+    size_t params_len;                                             /* The length of the array of parameters. */
+    const char *desc;                                              /* The description of the command that is shown in the help message. */
+    int invisible;                                                 /* Whether this command should be hidden in the help message. */
+    ValkeyModuleScriptingEngineDebuggerCommandHandlerFunc handler; /* The function pointer that implements this command. */
+    void *context;                                                 /* The pointer to a context structure that is passed when invoking the command handler. */
+} ValkeyModuleScriptingEngineDebuggerCommandV1;
+
+#define ValkeyModuleScriptingEngineDebuggerCommand ValkeyModuleScriptingEngineDebuggerCommandV1
+
+#define VALKEYMODULE_SCRIPTING_ENIGNE_DEBUGGER_COMMAND(NAME, PREFIX, PARAMS, PARAMS_LEN, DESC, INVISIBLE, HANDLER, ...) \
```
zuiderkwast: Can this variadic macro be extended in future versions? What do we expect VA ARGS to be here?

src/commands/script-debug.json
```diff
@@ -34,6 +34,12 @@
                         "token": "NO"
                     }
                 ]
+            },
+            {
+                "name": "engine",
+                "type": "string",
+                "optional": true,
+                "since": "8.1.0"
```
zuiderkwast: 9.0.0

zuiderkwast (commented):
I like to see Lua moved into a module. To keep backward compat behaviour here, we need to make it load by default and add some new config to disable loading the default modules. But that's a future PR.

Regarding deeping or deleting the debugger, I think it's safer to keep the debugger, so I guess we should get this merged.

I'm fine with merging this without too detailed review, as long as it has been tested. Preferably we should test the ABI compatibility too, at least manually. Old Valkey should ignore new module's debugger callbacks. New Valkey should work (not crash on SCRIPT DEBUG, etc.) with module compiled for old valkeymodule.h.

src/valkey-cli.c
rjd15372: Good point. Will use your suggestion.

src/valkeymodule.h
rjd15372: This VA ARGS is only used to pass an optional pointer to populate the `context` field of the `ValkeyModuleScriptingEngineDebuggerCommand` struct.

The alternative is to have two macros, one that does not have the arg to pass the context, and the other that has.

rjd15372: I'm seeing errors in the macos build because of this.  I will change to two macros.

rjd15372:
@zuiderkwast I've updated the PR, and addressed your comments. Added new tests and fixed some bugs that appeared in the tests.

rjd15372:
@zuiderkwast I've updated this PR with the current unstable branch. We should get this PR merged so I can open the PR that moves the lua engine into a module.

src/commands/script-debug.json
```diff
@@ -34,6 +34,12 @@
                         "token": "NO"
                     }
                 ]
+            },
+            {
+                "name": "engine",
+                "type": "string",
+                "optional": true,
+                "since": "9.0.0"
```
zuiderkwast: It's already too late to make it to 9.0.

```suggestion
                "since": "9.1.0"
```

zuiderkwast (commented):
Yes, let's get this merged.

You need to commit the generated file commands.def.

src/valkeymodule.h
```diff
@@ -1087,7 +1224,49 @@ typedef struct ValkeyModuleScriptingEngineMethods {
...
+typedef struct ValkeyModuleScriptingEngineMethodsV2 {
+    uint64_t version; /* Version of this structure for ABI compat. */
+
+    /* Compile code function callback. When a new script is loaded, this
+     * callback will be called with the script code, compiles it, and returns a
+     * list of `ValkeyModuleScriptingEngineCompiledFunc` objects. */
+    union {
+        ValkeyModuleScriptingEngineCompileCodeFuncV1 compile_code_v1;
+        ValkeyModuleScriptingEngineCompileCodeFunc compile_code;
+    };
```
zuiderkwast: Instead of duplicating the fields of V1, can't this V2 version just include ValkeyModuleScriptingEngineMethodsV1 as an anonymous struct?

zuiderkwast (commented):
I would like to get @madolson's eyes on the module API before we merge this.

src/valkeymodule.h
rjd15372: Wouldn't that require to also duplicate the V1 fields declaration?  Maybe I'm not understanding your suggestion, can you clarify?

zuiderkwast: Is the below a syntax error? :thinking:

```C
struct V1 {
   int foo;
   int bar;
}

struct V2 {
   struct V1; // Syntax error?
   int baz;
}
```

If not, there must be some other way to avoid duplication. We could cast between V1 and V2, but that's not very nice. Use a `union { V1; V2 } internally... Haven't we extended a struct like this before in the module API? There must be a nice way....

zuiderkwast: What if we just add fields to the same struct?

```C
struct S {
   int version;
   int foo;
   int bar;
   int baz; // Added in V2
}

// module.c

void f(void *x) {
    struct S s = x;
    printf("foo=%d bar=%d", s.foo, s.bar);
    if (s.version >= 2) {
        printf(" baz=%d", s.baz);
    } else {
        // Don't access baz. It's outside the allocation if the module is compiled with V1.
    }
}
```

rjd15372: I used a macro definition to avoid the duplication. WDYT?

zuiderkwast (approved):
Speed review, but it seems safe enough. I'm not worried about the debugger, but only about breaking the regular scripting with the refactoring. We have good test coverage for scripting though so I think it's safe.

rjd15372:
@madolson do you want to review the module API additions?

zuiderkwast:
I believe we can merge this without waiting for the others to review it, so we can unblock the other work. If there are any concerns regarding the module API, we have time to change it until 9.1.

It needs some rebase now and bump the scripting engine API version to 3?

rjd15372:
@zuiderkwast yes, I've fixed the merge conflicts and updated the code accordingly. Run a few more tests and looks good. Will merge this now.

PR #1703, opened 2025-02-10
Use `DEEPBIND` flag when loading external modules using `dlopen`

rjd15372:
The current flags used in `dlopen` to load external modules don't allow modules to define symbols that are already defined by the valkey server binary because the symbol resolution first looks in the server memory and only if it does not find anything, it looks in the module (shared library) memory.

This might become a problem if, for instance, we try to implement a new scripting engine based on a newer version of Lua. The Lua interpreter library shares many symbol names with the Lua interpreter included in the Valkey server binary.

To fix the above problem, this PR adds the flag `RTLD_DEEPBIND` to the flags used in `dlopen` on systems that support it, which changes the symbol resolution strategy to look for the symbol in the module memory first, if the code executing is from the module.

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -12280,8 +12280,9 @@ int moduleLoad(const char *path, void **module_argv, int module_argc, int is_loa
      * same symbols, does not work with ASAN. Therefore, we exclude
      * RTLD_DEEPBIND when doing test builds with ASAN.
      * See https://github.com/google/sanitizers/issues/611 for more details.
-     * This flag is also not available in OSX, but the default symbol resolution
-     * behavior in OSX is the same as when we use DEEPBIND in Linux. */
+     * This flag is also not available in macos, but the default symbol
+     * resolution behavior in macos is the same as when we use DEEPBIND in
+     * Linux. */
     dlopen_flags |= RTLD_DEEPBIND;
 #endif
 
--- a/src/module.c
+++ b/src/module.c
@@ -12274,7 +12274,16 @@ int moduleLoad(const char *path, void **module_argv, int module_argc, int is_loa
         }
     }
 
-    handle = dlopen(path, RTLD_NOW | RTLD_LOCAL);
+    int dlopen_flags = RTLD_NOW | RTLD_LOCAL;
+#ifndef __SANITIZE_ADDRESS__
+    /* RTLD_DEEPBIND, which is required for loading modules that contains the
+     * same symbols, does not work with ASAN. Therefore, we exclude
+     * RTLD_DEEPBIND when doing test builds with ASAN.
+     * See https://github.com/google/sanitizers/issues/611 for more details.*/
+    dlopen_flags |= RTLD_DEEPBIND;
+#endif
+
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):

src/module.c
```diff
@@ -12274,7 +12274,19 @@ int moduleLoad(const char *path, void **module_argv, int module_argc, int is_loa
         }
     }
 
-    handle = dlopen(path, RTLD_NOW | RTLD_LOCAL);
+    int dlopen_flags = RTLD_NOW | RTLD_LOCAL;
+#if !defined(__SANITIZE_ADDRESS__) && !defined(__APPLE__)
```
zuiderkwast: Just checking some exotic systems I know some people have cared about in the past:

* :x: Illumos (and probably other Solaris derived systems) doesn't have this flag. https://www.illumos.org/man/3C/dlopen
* :heavy_check_mark: FreeBSD has it. https://man.freebsd.org/cgi/man.cgi?query=dlopen
* :x: OpenBSD doesn't have it. https://man.openbsd.org/dlopen.3
* :x: NetBSD doesn't have it. https://man.netbsd.org/dlopen.3
* :x: DragonflyBSD doesn't have it. https://man.dragonflybsd.org/?command=dlopen

Some of these may have a similar flag but I didn't read carefully.

It's probably safer to use DEEPBIND only on explicitly listed systems. Instead of *if not macos* we could write the check as *if linux or freebsd*.

rjd15372: You're right. I'll change the condition

zuiderkwast: @panjf2000 FYI. You're our most active contributor for BSD-related things.

zuiderkwast:
@rjd15372  Daily job failed:

https://github.com/valkey-io/valkey/actions/runs/13253032412/job/36994775025


==89691==You are trying to dlopen a /home/runner/work/valkey/valkey/tests/modules/reply.so shared library with RTLD_DEEPBIND flag which is incompatible with sanitizer runtime (see https://github.com/google/sanitizers/issues/611 for details). If you want to run /home/runner/work/valkey/valkey/tests/modules/reply.so library under sanitizers please remove RTLD_DEEPBIND from dlopen flags.

enjoy-binbin:
I see we did skip it, is there anything missing in the condition?
```
    int dlopen_flags = RTLD_NOW | RTLD_LOCAL;
#if (defined(__linux__) || defined(__FreeBSD__)) && !defined(__SANITIZE_ADDRESS__)
    /* RTLD_DEEPBIND, which is required for loading modules that contains the
     * same symbols, does not work with ASAN. Therefore, we exclude
     * RTLD_DEEPBIND when doing test builds with ASAN.
     * See https://github.com/google/sanitizers/issues/611 for more details.
     *
     * This flag is also currently only available in Linux and FreeBSD. */
    dlopen_flags |= RTLD_DEEPBIND;
#endif
```

zuiderkwast:
@enjoy-binbin see this: https://github.com/valkey-io/valkey/pull/1707

Issue #1702, opened 2025-02-10
[BUG]No need to set keepalive option for the unix domain socket


chzhoo:
**Describe the bug**

Keepalive option is a concept at the TCP layer, but unix domain sockets do not have TCP layers. But we still set keepalive for Unix domain sockets using the function connKeepAlive. No need to set the tcpnodelay options either.

**To reproduce**

Step 1
strace -o strace.log ./src/redis-server --unixsocket ./redis.sock

Step 2
./src/valkey-cli  -s ./valkey.sock info

Step 3
We can see the following error from strace.log file
```
......
setsockopt(11, SOL_TCP, TCP_NODELAY, [1], 4) = -1 EOPNOTSUPP (Operation not supported)
setsockopt(11, SOL_SOCKET, SO_KEEPALIVE, [1], 4) = 0
setsockopt(11, SOL_TCP, TCP_KEEPIDLE, [300], 4) = -1 EOPNOTSUPP (Operation not supported)
......

```
**Expected behavior**

No need to set keepalive option for the Unix domain socket

**Additional information**

xbasel:
My understanding is that these `EOPNOTSUPP`  errors do not break functionality, but they can be misleading. 
should be an easy fix.

hpatro:
@sungming2 Could you take a look into this?

sungming2:
I'm looking

PR #1706, opened 2025-02-11
Move TCP/TLS specific options from generic client to connection type

sungming2:
### Problem
Unix domain socket doesn't operate over the TCP layer, making these options unnecessary and, in some cases, unsupported. This results in errors like EOPNOTSUPP (Operation not supported) when attempting to apply these settings using setsockopt

### Test
* UNIX
```
192903:M 10 Feb 2025 19:51:15.617 * Ready to accept connections unix
accept4(10, {sa_family=AF_UNIX}, [110->2], SOCK_CLOEXEC|SOCK_NONBLOCK) = 11
```

* TCP
```
192043:M 10 Feb 2025 19:48:46.858 * Ready to accept connections tcp
accept4(8, {sa_family=AF_INET, sin_port=htons(47132), sin_addr=inet_addr("127.0.0.1")}, [128->16], SOCK_CLOEXEC|SOCK_NONBLOCK) = 10
setsockopt(10, SOL_TCP, TCP_NODELAY, [1], 4) = 0
setsockopt(10, SOL_SOCKET, SO_KEEPALIVE, [1], 4) = 0
```

* TLS
```
191419:M 10 Feb 2025 19:46:33.504 * Ready to accept connections tls
accept4(8, {sa_family=AF_INET, sin_port=htons(48788), sin_addr=inet_addr("127.0.0.1")}, [128->16], SOCK_CLOEXEC|SOCK_NONBLOCK) = 10
setsockopt(10, SOL_TCP, TCP_NODELAY, [1], 4) = 0
setsockopt(10, SOL_SOCKET, SO_KEEPALIVE, [1], 4) = 0
```

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -157,8 +157,14 @@ client *createClient(connection *conn) {
      * in the context of a client. When commands are executed in other
      * contexts (for instance a Lua script) we need a non connected client. */
     if (conn) {
-        connEnableTcpNoDelay(conn);
-        if (server.tcpkeepalive) connKeepAlive(conn, server.tcpkeepalive);
+        /* These options are only applicable to TCP or TLS connection.*/
+        const char *typename = conn->type->get_type(NULL);
+        if (!strcasecmp(CONN_TYPE_SOCKET, typename) ||
+            !strcasecmp(CONN_TYPE_TLS, typename)) {
+            connEnableTcpNoDelay(conn);
+            if (server.tcpkeepalive) connKeepAlive(conn, server.tcpkeepalive);
+        }
+
         connSetReadHandler(conn, readQueryFromClient);
         connSetPrivateData(conn, c);
         conn->flags |= CONN_FLAG_ALLOW_ACCEPT_OFFLOAD;
--- a/src/networking.c
+++ b/src/networking.c
@@ -157,14 +157,6 @@ client *createClient(connection *conn) {
      * in the context of a client. When commands are executed in other
      * contexts (for instance a Lua script) we need a non connected client. */
     if (conn) {
-        /* These options are only applicable to TCP or TLS connection.*/
-        const char *typename = conn->type->get_type(NULL);
-        if (!strcasecmp(CONN_TYPE_SOCKET, typename) ||
-            !strcasecmp(CONN_TYPE_TLS, typename)) {
-            connEnableTcpNoDelay(conn);
-            if (server.tcpkeepalive) connKeepAlive(conn, server.tcpkeepalive);
-        }
-
         connSetReadHandler(conn, readQueryFromClient);
         connSetPrivateData(conn, c);
         conn->flags |= CONN_FLAG_ALLOW_ACCEPT_OFFLOAD;
--- a/src/socket.c
+++ b/src/socket.c
@@ -326,6 +326,9 @@ static void connSocketAcceptHandler(aeEventLoop *el, int fd, void *privdata, int
             return;
         }
         serverLog(LL_VERBOSE, "Accepted %s:%d", cip, cport);
+
+        anetEnableTcpNoDelay(NULL, cfd);
+        if (server.tcpkeepalive) anetKeepAlive(NULL, cfd, server.tcpkeepalive);
         acceptCommonHandler(connCreateAcceptedSocket(cfd, NULL), flags, cip);
     }
 }
--- a/src/tls.c
+++ b/src/tls.c
@@ -798,6 +798,9 @@ static void tlsAcceptHandler(aeEventLoop *el, int fd, void *privdata, int mask) 
             return;
         }
         serverLog(LL_VERBOSE, "Accepted %s:%d", cip, cport);
+
+        anetEnableTcpNoDelay(NULL, cfd);
+        if (server.tcpkeepalive) anetKeepAlive(NULL, cfd, server.tcpkeepalive);
         acceptCommonHandler(connCreateAcceptedTLS(cfd, &server.tls_auth_clients), flags, cip);
     }
 }


src/networking.c
```diff
@@ -157,8 +157,11 @@ client *createClient(connection *conn) {
      * in the context of a client. When commands are executed in other
      * contexts (for instance a Lua script) we need a non connected client. */
     if (conn) {
-        connEnableTcpNoDelay(conn);
-        if (server.tcpkeepalive) connKeepAlive(conn, server.tcpkeepalive);
+        if (strcasecmp(CONN_TYPE_SOCKET, conn->type->get_type(NULL)) == 0) {
```
madolson: Don't we also need to check for TLS?

I also just wonder if we should add a comment if the behavior isn't detrimental

sungming2: You are right, we need to enable them for TLS too since it operates on top of TCP.
I also verified it with system call trace with TLS mode
```
186289:M 10 Feb 2025 19:26:25.777 * Ready to accept connections tls
accept4(8, {sa_family=AF_INET, sin_port=htons(54748), sin_addr=inet_addr("127.0.0.1")}, [128->16], SOCK_CLOEXEC|SOCK_NONBLOCK) = 10
186289:M 10 Feb 2025 19:26:39.557 * connEnableTcpNoDelay: tls
setsockopt(10, SOL_TCP, TCP_NODELAY, [1], 4) = 0
setsockopt(10, SOL_SOCKET, SO_KEEPALIVE, [1], 4) = 0
setsockopt(10, SOL_TCP, TCP_KEEPIDLE, [300], 4) = 0
setsockopt(10, SOL_TCP, TCP_KEEPINTVL, [100], 4) = 0
setsockopt(10, SOL_TCP, TCP_KEEPCNT, [3], 4) = 0
getpeername(10, {sa_family=AF_INET, sin_port=htons(54748), sin_addr=inet_addr("127.0.0.1")}, [128->16]) = 0
```

src/networking.c
```diff
@@ -157,8 +157,14 @@ client *createClient(connection *conn) {
      * in the context of a client. When commands are executed in other
      * contexts (for instance a Lua script) we need a non connected client. */
     if (conn) {
-        connEnableTcpNoDelay(conn);
```
hpatro, line 160: Why is enabling delay not part of the `ConnectionType` struct? 

Can't we create a method, let's say `connEnableNoDelay` and based on the implementation on each type, it would be invoked? These conditional statement doesn't seem to adhere to the abstraction.

pizhenwei: For now, there are "tcp", "unix", "tls" and "rdma" connection types, but TCP NODELAY is used by TCP and TLS/TCP only, so I don't think setting TCP specific methods in generic client is good choice. Rather than testing the connection type from client generic codes, I prefer to hide this into connection type. For example, enable these options on a new client accepted for TCP connection:

```
static void connSocketAcceptHandler(aeEventLoop *el, int fd, void *privdata, int mask) {
    int cport, cfd;
    int max = server.max_new_conns_per_cycle;
    struct ClientFlags flags = {0};
    char cip[NET_IP_STR_LEN];
    UNUSED(el);
    UNUSED(mask);
    UNUSED(privdata);

    while (max--) {
        cfd = anetTcpAccept(server.neterr, fd, cip, sizeof(cip), &cport);
        if (cfd == ANET_ERR) {
            if (errno != EWOULDBLOCK) serverLog(LL_WARNING, "Accepting client connection: %s", server.neterr);
            return;
        }   
        serverLog(LL_VERBOSE, "Accepted %s:%d", cip, cport);
        anetEnableTcpNoDelay(NULL, cfd);
        if (server.tcpkeepalive) anetKeepAlive(NULL, cfd, server.tcpkeepalive);
        acceptCommonHandler(connCreateAcceptedSocket(cfd, NULL), flags, cip);
    }   
}
```

sungming2: I think that is a good idea. thanks for the suggestion

hpatro (approved):
LGTM

pizhenwei (approved):
LGTM, thanks!

ranshid:
What about connTLSConnect and connSocketConnect?

pizhenwei:
> What about connTLSConnect and connSocketConnect?

This is a TCP client side socket, but not any newly accepted 'client' from server side.

But from the description of `tcp-keepalive`, it's not documented as server only, so I think it's OK to apply on both client/server TCP socket.

```
# TCP keepalive.
#
# If non-zero, use SO_KEEPALIVE to send TCP ACKs to clients in absence
# of communication. This is useful for two reasons:
#
# 1) Detect dead peers.
# 2) Force network equipment in the middle to consider the connection to be
#    alive.
#
# On Linux, the specified value (in seconds) is the period used to send ACKs.
# Note that to close the connection the double of the time is needed.
# On other kernels the period depends on the kernel configuration.
tcp-keepalive 300
```

If this option is consider to apply for both, I suggest moving the two TCP option setting codes into `anetTcpAccept` and `anetTcpGenericConnect` instead.

sungming2:
I think the client is already using the keepalive option with fixed interval. please correct me if I am wrong
https://github.com/valkey-io/valkey/blob/f85c93301ca63f070164b307c6a7371a9ab5ac7b/src/valkey-cli.c#L1664-L1668

ranshid:
> > What about connTLSConnect and connSocketConnect?
> 
> This is a TCP client side socket, but not any newly accepted 'client' from server side.
> 
> But from the description of `tcp-keepalive`, it's not documented as server only, so I think it's OK to apply on both client/server TCP socket.
> 
> ```
> # TCP keepalive.
> #
> # If non-zero, use SO_KEEPALIVE to send TCP ACKs to clients in absence
> # of communication. This is useful for two reasons:
> #
> # 1) Detect dead peers.
> # 2) Force network equipment in the middle to consider the connection to be
> #    alive.
> #
> # On Linux, the specified value (in seconds) is the period used to send ACKs.
> # Note that to close the connection the double of the time is needed.
> # On other kernels the period depends on the kernel configuration.
> tcp-keepalive 300
> ```
> 
> If this option is consider to apply for both, I suggest moving the two TCP option setting codes into `anetTcpAccept` and `anetTcpGenericConnect` instead.

Well changing it on the anetTcpAccept would probably cause us to set and then re-set it again on the cluster flow. probably not major thing but we have different keepalive configuration on the cluster bus 

> > What about connTLSConnect and connSocketConnect?
> 
> This is a TCP client side socket, but not any newly accepted 'client' from server side.
> 
> But from the description of `tcp-keepalive`, it's not documented as server only, so I think it's OK to apply on both client/server TCP socket.
> 
> ```
> # TCP keepalive.
> #
> # If non-zero, use SO_KEEPALIVE to send TCP ACKs to clients in absence
> # of communication. This is useful for two reasons:
> #
> # 1) Detect dead peers.
> # 2) Force network equipment in the middle to consider the connection to be
> #    alive.
> #
> # On Linux, the specified value (in seconds) is the period used to send ACKs.
> # Note that to close the connection the double of the time is needed.
> # On other kernels the period depends on the kernel configuration.
> tcp-keepalive 300
> ```
> 
> If this option is consider to apply for both, I suggest moving the two TCP option setting codes into `anetTcpAccept` and `anetTcpGenericConnect` instead.

Well setting this on the anetTcpAccept would probably mean we will have to keep the setting on the clusterAcceptHandler which will have to override it with server.cluster_node_timeout.

regarding TCP nodelay - IMO this should not relate to accepted/connected connections. on the cluster-bus we initiate connections from nodes and use them to send cluster related traffic and I am not sure why  we do not set the TCP_nodelay there as well (if indeed we do not). I think disabling Nigel on the sender side is important for cluster bus (eg pub/sub).

pizhenwei:
> I think the client is already using the keepalive option with fixed interval. please correct me if I am wrong
> 
> https://github.com/valkey-io/valkey/blob/f85c93301ca63f070164b307c6a7371a9ab5ac7b/src/valkey-cli.c#L1664-L1668

For example, replication node uses `connConnect` as a client.

ranshid:
I personally see no reason why we should not set NODELAY on any type of TCP connection, but we need to better control the timeouts on clusterbus/replication and regular clients.

pizhenwei:
> regarding TCP nodelay - IMO this should not relate to accepted/connected connections. on the cluster-bus we initiate connections from nodes and use them to send cluster related traffic and I am not sure why we do not set the TCP_nodelay there as well (if indeed we do not). I think disabling Nigel on the sender side is important for cluster bus (eg pub/sub).

TCP Nigel would not improve performance on modern CPU(~3.0G Hz) and network interface(10Gb ~ 100Gb), but it may leads unexpected networking latency. Valkey is performance sensitive, disabling it may be a good choice.

pizhenwei:
> I personally see no reason why we should not set NODELAY on any type of TCP connection, 

Agree.

> but we need to better control the timeouts on clusterbus/replication and regular clients.

What about setting server timeout by default on connConnect instead of  the default 2 hours kernel keepalive, clusterbus/replication overwrite it if necessary.

ranshid:
>What about setting server timeout by default on connConnect instead of the default 2 hours kernel keepalive, clusterbus/replication overwrite it if necessary.

I do not see any value in setting some some value and override it later. we just need to make sure we do set it correctly for both the connect and accept flows. previously setting it in the client MOSTLY did the work for the normal clients/replication (even though for the replica the create client with the primary happens after much of the handshake) 
For the clusterbus, the fact that we do not (at least) setup the TCP_NODELAY on the connect path is wrong IMO and should be fixed, unless I missed something and it is set somewhere

sungming2:
> For the clusterbus, the fact that we do not (at least) setup the TCP_NODELAY on the connect path is wrong IMO and should be fixed, unless I missed something and it is set somewhere

Should this be addressed in a separate issue/pr? It seems unrelated to the current issue. I am happy to work on a separate pr for it if you don't mind.

sungming2:
> I do not see any value in setting some some value and override it later.

Could you clarify your concern about it please? I would like to understand the specific downsides of overriding the keepalive option. Does it have a significant impact, or is it more of a preference for cleaner configuration?

pizhenwei:
> > For the clusterbus, the fact that we do not (at least) setup the TCP_NODELAY on the connect path is wrong IMO and should be fixed, unless I missed something and it is set somewhere
> 
> Should this be addressed in a separate issue/pr? It seems unrelated to the current issue. I am happy to work on a separate pr for it if you don't mind.

@madolson @zuiderkwast  @ranshid 
I personally prefer to merge current change to resolve [issue](https://github.com/valkey-io/valkey/issues/1702), then open a new issue to discuss TCP NODELAY or a new PR to enable TCP NODELAY by default.

ranshid:
> > > For the clusterbus, the fact that we do not (at least) setup the TCP_NODELAY on the connect path is wrong IMO and should be fixed, unless I missed something and it is set somewhere
> > 
> > 
> > Should this be addressed in a separate issue/pr? It seems unrelated to the current issue. I am happy to work on a separate pr for it if you don't mind.
> 
> @madolson @zuiderkwast @ranshid I personally prefer to merge current change to resolve [issue](https://github.com/valkey-io/valkey/issues/1702), then open a new issue to discuss TCP NODELAY or a new PR to enable TCP NODELAY by default.

I have no big concerns about it, but lets just make sure we do not degrade in anyway by making sure the replication connection to the primary is also configured correctly. Probably no risk there, since it hardly send anything and the replica is making keepalive checks with pings, but I think to keep the current logic is preferred.  

>Should this be addressed in a separate issue/pr? It seems unrelated to the current issue. I am happy to work on a separate pr for it if you don't mind.

@sungming2 yes. separate PR is totally fine

hpatro:
@valkey-io/core-team What's the current strategy around merging changes into unstable ? Can we merge this in?

zuiderkwast:
> What's the current strategy around merging changes into unstable ? Can we merge this in?

@hpatro You mean if we have a feature freeze? We shouldn't merge large features right now until we've stabilized the flaky tests a bit more. This one is more of a small bug fix though which looks very safe. I think you can merge it. Add it to 8.1 project and add relevant labels if any.

PR #1708, opened 2025-02-11
Pass null-terminated node ID for `VM_RegisterClusterMessageReceiver` and add test coverage

### LINKED ISSUE #1656: Add test coverage for module callback registration for cluster message via `VM_RegisterClusterMessageReceiver`
Adding some test coverage for the module message receiver callback API in another PR, that would be nice.

_Originally posted by @zuiderkwast in https://github.com/valkey-io/valkey/pull/1572#pullrequestreview-2587906183_

Nikhil-Manglore:
* Pass `sender_id` as `NULL` terminated string as part of `ValkeyModuleClusterMessageReceiver` for ease of usage by the module(s).

* Implement test coverage described in #1656 and ensures that nodes in the cluster module properly acknowledge a "DING" message by sending a "DONG" response.

Closes #1656.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3045,7 +3045,13 @@ static void clusterProcessModulePacket(clusterMsgModule *module_data, clusterNod
     uint32_t len = ntohl(module_data->len);
     uint8_t type = module_data->type;
     unsigned char *payload = module_data->bulk_data;
-    moduleCallClusterReceivers(sender->name, module_id, type, payload, len);
+
+    /* Ensure sender name is properly truncated to 40 characters */
+    char truncated_name[41];  /* 40 chars + null terminator */
+    strncpy(truncated_name, sender->name, 40);
+    truncated_name[40] = '\0';
+
+    moduleCallClusterReceivers(truncated_name, module_id, type, payload, len);
 }
 
 static void clusterProcessLightPacket(clusterNode *sender, clusterLink *link, uint16_t type) {
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3047,11 +3047,9 @@ static void clusterProcessModulePacket(clusterMsgModule *module_data, clusterNod
     unsigned char *payload = module_data->bulk_data;
 
     /* Ensure sender name is properly truncated to 40 characters */
-    char truncated_name[41]; /* 40 chars + null terminator */
-    strncpy(truncated_name, sender->name, 40);
-    truncated_name[40] = '\0';
-
+    sds truncated_name = sdsnewlen(sender->name, 40);
     moduleCallClusterReceivers(truncated_name, module_id, type, payload, len);
+    sdsfree(truncated_name);
 }
 
 static void clusterProcessLightPacket(clusterNode *sender, clusterLink *link, uint16_t type) {
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3047,7 +3047,7 @@ static void clusterProcessModulePacket(clusterMsgModule *module_data, clusterNod
     unsigned char *payload = module_data->bulk_data;
 
     /* Ensure sender name is properly truncated to 40 characters */
-    char truncated_name[41];  /* 40 chars + null terminator */
+    char truncated_name[41]; /* 40 chars + null terminator */
     strncpy(truncated_name, sender->name, 40);
     truncated_name[40] = '\0';
 
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3046,10 +3046,9 @@ static void clusterProcessModulePacket(clusterMsgModule *module_data, clusterNod
     uint8_t type = module_data->type;
     unsigned char *payload = module_data->bulk_data;
 
-    /* Ensure sender name is properly truncated to 40 characters */
-    sds truncated_name = sdsnewlen(sender->name, 40);
-    moduleCallClusterReceivers(truncated_name, module_id, type, payload, len);
-    sdsfree(truncated_name);
+    sds sender_name = sdsnewlen(sender->name, 40);
+    moduleCallClusterReceivers(sender_name, module_id, type, payload, len);
+    sdsfree(sender_name);
 }
 
 static void clusterProcessLightPacket(clusterNode *sender, clusterLink *link, uint16_t type) {
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3046,7 +3046,7 @@ static void clusterProcessModulePacket(clusterMsgModule *module_data, clusterNod
     uint8_t type = module_data->type;
     unsigned char *payload = module_data->bulk_data;
 
-    sds sender_name = sdsnewlen(sender->name, 40);
+    sds sender_name = sdsnewlen(sender->name, CLUSTER_NAMELEN);
     moduleCallClusterReceivers(sender_name, module_id, type, payload, len);
     sdsfree(sender_name);
 }
--- a/src/module.c
+++ b/src/module.c
@@ -8940,7 +8940,14 @@ void moduleCallClusterReceivers(const char *sender_id,
  * was already a registered callback, this will replace the callback function
  * with the one provided, otherwise if the callback is set to NULL and there
  * is already a callback for this function, the callback is unregistered
- * (so this API call is also used in order to delete the receiver). */
+ * (so this API call is also used in order to delete the receiver).
+ *
+ * When a message of this type is received, the registered callback function
+ * will be invoked with details, including the node ID.
+ *
+ * Behavior Change:
+ * Before Valkey 8.1: The node ID was not null-terminated.
+ * Currently: The node ID is now null-terminated for easier handling. */
 void VM_RegisterClusterMessageReceiver(ValkeyModuleCtx *ctx,
                                        uint8_t type,
                                        ValkeyModuleClusterMessageReceiver callback) {
--- a/src/module.c
+++ b/src/module.c
@@ -8946,8 +8946,9 @@ void moduleCallClusterReceivers(const char *sender_id,
  * will be invoked with details, including the node ID.
  *
  * Behavior Change:
- * Before Valkey 8.1: The node ID was not null-terminated.
- * Currently: The node ID is now null-terminated for easier handling. */
+ * Before Valkey 8.1: The node ID was not NULL terminated.
+ * Currently: The node ID 'sender_id' in the callback receiver is now a
+ * NULL terminated string for easier handling. */
 void VM_RegisterClusterMessageReceiver(ValkeyModuleCtx *ctx,
                                        uint8_t type,
                                        ValkeyModuleClusterMessageReceiver callback) {
--- a/src/module.c
+++ b/src/module.c
@@ -8943,12 +8943,10 @@ void moduleCallClusterReceivers(const char *sender_id,
  * (so this API call is also used in order to delete the receiver).
  *
  * When a message of this type is received, the registered callback function
- * will be invoked with details, including the node ID.
+ * will be invoked with details, including the 40-byte node ID of the sender.
  *
- * Behavior Change:
- * Before Valkey 8.1: The node ID was not NULL terminated.
- * Currently: The node ID 'sender_id' in the callback receiver is now a
- * NULL terminated string for easier handling. */
+ * In Valkey 8.1 and later, the node ID is null-terminated. Prior to 8.1, it was
+ * not null-terminated */
 void VM_RegisterClusterMessageReceiver(ValkeyModuleCtx *ctx,
                                        uint8_t type,
                                        ValkeyModuleClusterMessageReceiver callback) {
--- a/tests/modules/cluster.c
+++ b/tests/modules/cluster.c
@@ -46,6 +46,19 @@ int PingallCommand(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
     return ValkeyModule_ReplyWithSimpleString(ctx, "OK");
 }
 
+void DingReceiver(ValkeyModuleCtx *ctx, const char *sender_id, uint8_t type, const unsigned char *payload, uint32_t len) {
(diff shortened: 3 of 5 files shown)


Nikhil-Manglore:
Looking into the CI Failure

tests/modules/cluster.c
```diff
@@ -46,6 +46,19 @@ int PingallCommand(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
     return ValkeyModule_ReplyWithSimpleString(ctx, "OK");
 }
 
+void DingReceiver(ValkeyModuleCtx *ctx, const char *sender_id, uint8_t type, const unsigned char *payload, uint32_t len) {
+    ValkeyModule_Log(ctx, "notice", "DING (type %d) RECEIVED from %.*s: '%.*s'", type, VALKEYMODULE_NODE_ID_LEN, sender_id, (int)len, payload);
+    ValkeyModule_SendClusterMessage(ctx, NULL, MSGTYPE_DONG, "Message Received!", 17);
+    ValkeyModuleCallReply *reply = ValkeyModule_Call(ctx, "INCR", "c", "dings_received");
```
hpatro: This will be tricky.`INCR` is a write operation so it can only be applied on a primary. For testing purpose, log verification is sufficient. So, we could remove this operation.

tests/modules/cluster.c
```diff
@@ -46,6 +46,15 @@ int PingallCommand(ValkeyModuleCtx *ctx, ValkeyModuleString **argv, int argc) {
     return ValkeyModule_ReplyWithSimpleString(ctx, "OK");
 }
 
+void DingReceiver(ValkeyModuleCtx *ctx, const char *sender_id, uint8_t type, const unsigned char *payload, uint32_t len) {
+    ValkeyModule_Log(ctx, "notice", "DING (type %d) RECEIVED from %.*s: '%.*s'", type, VALKEYMODULE_NODE_ID_LEN, sender_id, (int)len, payload);
+    ValkeyModule_SendClusterMessage(ctx, NULL, MSGTYPE_DONG, "Message Received!", 17);
```
hpatro: Could you only send the response back to the sender and not broadcast to everyone?

tests/unit/moduleapi/cluster.tcl
```diff
@@ -16,12 +16,26 @@ start_cluster 3 0 [list config_lines $modules] {
         assert_equal OK [$node1 test.pingall]
         assert_equal 2 [CI 0 cluster_stats_messages_module_sent]
         wait_for_condition 50 100 {
-            [CI 1 cluster_stats_messages_module_received] eq 1 &&
-            [CI 2 cluster_stats_messages_module_received] eq 1
+            [CI 1 cluster_stats_messages_module_received] eq 2 &&
+            [CI 2 cluster_stats_messages_module_received] eq 2
```
hpatro: This is because when you receive `DING` you send out another message in the receiver. Check my comment on the module code.

src/cluster_legacy.c
```diff
@@ -3045,7 +3045,13 @@ static void clusterProcessModulePacket(clusterMsgModule *module_data, clusterNod
     uint32_t len = ntohl(module_data->len);
     uint8_t type = module_data->type;
     unsigned char *payload = module_data->bulk_data;
-    moduleCallClusterReceivers(sender->name, module_id, type, payload, len);
+
+    /* Ensure sender name is properly truncated to 40 characters */
+    char truncated_name[41]; /* 40 chars + null terminator */
+    strncpy(truncated_name, sender->name, 40);
+    truncated_name[40] = '\0';
```
hpatro: Could you create SDS instead of manually copying and adding `\0` and send it in the callback and free it after that?

src/cluster_legacy.c
```diff
@@ -3045,7 +3045,11 @@ static void clusterProcessModulePacket(clusterMsgModule *module_data, clusterNod
     uint32_t len = ntohl(module_data->len);
     uint8_t type = module_data->type;
     unsigned char *payload = module_data->bulk_data;
-    moduleCallClusterReceivers(sender->name, module_id, type, payload, len);
+
+    /* Ensure sender name is properly truncated to 40 characters */
+    sds truncated_name = sdsnewlen(sender->name, 40);
+    moduleCallClusterReceivers(truncated_name, module_id, type, payload, len);
+    sdsfree(truncated_name);
```
hpatro: This is not truncating per se. This is more like adding a `\0` at the end (null terminated string). You could just name it as sender_id.

```suggestion
    sds sender_name = sdsnewlen(sender->name, 40);
    moduleCallClusterReceivers(sender_name, module_id, type, payload, len);
    sdsfree(sender_name);
```

tests/unit/moduleapi/cluster.tcl
```diff
@@ -22,6 +22,24 @@ start_cluster 3 0 [list config_lines $modules] {
...
+        assert_equal OK [$node1 test.pingall]
+        wait_for_condition 50 100 {
+            [CI 0 cluster_stats_messages_module_received] eq 2  &&
+            [CI 1 cluster_stats_messages_module_received] eq 1 &&
+            [CI 2 cluster_stats_messages_module_received] eq 1
+        } else {
+            fail "node 2 or node 3 didn't receive DING messages or node 1 didn't receive DONG message"
+        }
+
+        verify_log_message -1 "*DING (type 1) RECEIVED*Hey*" 0
```
hpatro, line 40: We should verify DING received on both the nodes.

tests/unit/moduleapi/cluster.tcl
```diff
@@ -22,6 +22,24 @@ start_cluster 3 0 [list config_lines $modules] {
...
+        verify_log_message 0 "*DONG (type 2) RECEIVED*Message Received*" 0
```
hpatro: As both receiver will send back `DONG`, we should use the utility method `count_log_message` to verify if we received `DONG` twice.

hpatro (commented):
Mostly LGTM.

hpatro (approved):

src/cluster_legacy.c
```diff
@@ -3045,7 +3045,10 @@ static void clusterProcessModulePacket(clusterMsgModule *module_data, clusterNod
     uint32_t len = ntohl(module_data->len);
     uint8_t type = module_data->type;
     unsigned char *payload = module_data->bulk_data;
-    moduleCallClusterReceivers(sender->name, module_id, type, payload, len);
+
+    sds sender_name = sdsnewlen(sender->name, 40);
+    moduleCallClusterReceivers(sender_name, module_id, type, payload, len);
+    sdsfree(sender_name);
```
enjoy-binbin, line 3051: so in the past the module need to an extra CLUSTER_NAMELEN check, do we have somewhere that document this behavior?

zuiderkwast: Why do we need to dup the string? Is this fixing some bug? Is the name not null terminated?

Nikhil-Manglore: > Why do we need to dup the string? Is this fixing some bug? Is the name not null terminated?

The sender name is greater than 40 characters and the expected length that was originally defined in the codebase was 40 characters. So when we call `ValkeyModule_SendClusterMessage` it's only expecting the first 40 characters which causes the response to not be sent back to the sender. After checking the expected sender_id and the sender_id that was passed into the function the first 40 characters matched up but there were a lot of random characters after that.

zuiderkwast: What is the sender name? Is it not a 40 characters node-id?

If it's longer, this can never have worked and can never have been used by modules?

It sounds to me that this is a bug fix, well hidden under a title "Add test coverage". :laughing: If this is as serious as it sounds to me, then we should backport the bugfix to 8.0.x and 7.2.x

zuiderkwast: > The sender name is greater than 40 characters

This is not true. It is 40 characters, but it's not null terminated. This is not documented anywhere in the module API, which is a problem.

We can make it null-terminated by copying it to a new null-terminated string as you do, or we can do a hack and just insert a `\0` on position 40 and restore it afterwards to avoid the copying. But fixing this leaves us with another subtle problem:

If a module works for a Valkey version that has this fix, then someone tries the same module with an older Valkey, it doesn't work. Thus, we need to document this in some way that in the past it wasn't null-terminated and now it is. The doc comment at `VM_RegisterClusterMessageReceiver` is a place where we can put this.

We can also just skip fixing this in this PR. IMO we should just document that it is not null-terminated.

hpatro: It was my suggestion to send a NULL terminated string instead of not NULL terminated. If a module has already been built around handling the current behavior not NULL terminated string, they shouldn't see any difference.

I felt it was not module developer friendly to pass a not NULL terminated string. I think we We should add the documentation that it is a `NULL` terminated string going forward (8.1 or 9.0).

Nikhil-Manglore: Should I add the documentation to this PR?

zuiderkwast: > It was my suggestion to send a NULL terminated string instead of not NULL terminated. If a module has already been built around handling the current behavior not NULL terminated string, they shouldn't see any difference.

@hpatro Yeah I agree null terminated string is better. I was mainly surprised to see a change like that in a PR titled "Add test coverage". Don't hide behavior changes under a title like that. It can raise suspicion. :grin:

If we do change to null-terminated, then that's the main thing of this PR and should be reflected in the title. It's even something to mention in the release notes.

I don't mind that we do it, in this or a separate PR as you wish, but it needs some highlight (the PR title) and if we do it, the doc comments describing what it is and the difference compared to older valkey versions too.

@Nikhil-Manglore It's up to you if you want to include the null-term thing in this PR or not, but if you do, then also fix the docs.

Nikhil-Manglore: That makes sense, I will make the titles more explicit next time. I will include it in this PR because otherwise the test coverage won't work. I will update the docs and push the changes.

zuiderkwast: > because otherwise the test coverage won't work

Well, otherwise you would have to do things differently in Ding/Dong_Receiver functions.

`ValkeyModule_SendClusterMessage` takes the node-id as a null-terminated string? Yeah, this is a mess. It's good that you're fixing it.

hpatro: Yeah, @Nikhil-Manglore started off with writing the tests and then found this behavior. I think we are all in the same page about fixing it and updating the docs.

src/module.c
```diff
@@ -8940,7 +8940,14 @@ void moduleCallClusterReceivers(const char *sender_id,
...
  * is already a callback for this function, the callback is unregistered
- * (so this API call is also used in order to delete the receiver). */
+ * (so this API call is also used in order to delete the receiver).
+ *
+ * When a message of this type is received, the registered callback function
+ * will be invoked with details, including the node ID.
+ *
+ * Behavior Change:
+ * Before Valkey 8.1: The node ID was not null-terminated.
+ * Currently: The node ID is now null-terminated for easier handling. */
```
hpatro: Suggestion:
```
The node ID `sender_id` in the callback receiver is now a NULL terminated string for easier handling.
```

zuiderkwast: Yes, great! Just some minor suggestions:

* Mention that it's 40 bytes long, so it's possible know how to write modules to work with older Valkey versions.
* Write "In Valkey 8.1 and later ..." like how we write in some other places in the docs.

```suggestion
 * will be invoked with details, including the 40 bytes node ID of the sender.
 *
 * In Valkey 8.1 and later, the node ID is null-terminated. Prior to 8.1, it was
 * not null-terminated.
```

src/cluster_legacy.c
```diff
@@ -3045,7 +3045,10 @@ static void clusterProcessModulePacket(clusterMsgModule *module_data, clusterNod
     uint32_t len = ntohl(module_data->len);
     uint8_t type = module_data->type;
     unsigned char *payload = module_data->bulk_data;
-    moduleCallClusterReceivers(sender->name, module_id, type, payload, len);
+
+    sds sender_name = sdsnewlen(sender->name, 40);
```
zuiderkwast: There's a macro for this 40.

```suggestion
    sds sender_name = sdsnewlen(sender->name, CLUSTER_NAMELEN);
```

tests/unit/moduleapi/cluster.tcl
```diff
@@ -22,6 +22,25 @@ start_cluster 3 0 [list config_lines $modules] {
...
     }
+
+    $node1 CONFIG RESETSTAT
+    $node2 CONFIG RESETSTAT
+    $node3 CONFIG RESETSTAT
+
+    test "Cluster module message DING/DONG acknowledgment" {
+        assert_equal OK [$node1 test.pingall]
+        wait_for_condition 50 100 {
+            [CI 0 cluster_stats_messages_module_received] eq 2  &&
```
zuiderkwast: Super-nit: two spaces

```suggestion
            [CI 0 cluster_stats_messages_module_received] eq 2 &&
```

zuiderkwast (commented):
Yeah, just some nits, then let's merge it.

tests/unit/moduleapi/cluster.tcl
Nikhil-Manglore: Sorry just saw this, will make a new commit with these updates.

zuiderkwast: I can commit the suggestions in the github web page if you want

Nikhil-Manglore: It's ok, i'm about to push the updated code.

zuiderkwast (approved):

tests/unit/moduleapi/cluster.tcl
```diff
@@ -22,6 +22,25 @@ start_cluster 3 0 [list config_lines $modules] {
             fail "node 2 or node 3 didn't receive cluster module message"
         }
     }
+
+    $node1 CONFIG RESETSTAT
+    $node2 CONFIG RESETSTAT
+    $node3 CONFIG RESETSTAT
+
+    test "Cluster module message DING/DONG acknowledgment" {
```
madolson, line 30: This test seems to be failing: https://github.com/valkey-io/valkey/actions/runs/13447133659/job/37574838815.

Nikhil-Manglore: Will look into it

PR #1710, opened 2025-02-11
Fix new benchmark test under the TLS

enjoy-binbin:
We need to call cliSecureConnection if we are runing under the TLS, otherwise the connection can not be use.

### DIFF
--- a/tests/integration/valkey-benchmark.tcl
+++ b/tests/integration/valkey-benchmark.tcl
@@ -12,6 +12,9 @@ proc common_bench_setup {cmd} {
     if {[catch { exec {*}$cmd } error]} {
         set first_line [lindex [split $error "\n"] 0]
         puts [colorstr red "valkey-benchmark non zero code. first line: $first_line"]
+        if {$::verbose} {
+            puts [colorstr red "valkey-benchmark non zero code, the output is: $error"]
+        }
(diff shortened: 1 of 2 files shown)


tests/integration/valkey-benchmark.tcl
```diff
@@ -12,6 +12,7 @@ proc common_bench_setup {cmd} {
     if {[catch { exec {*}$cmd } error]} {
         set first_line [lindex [split $error "\n"] 0]
         puts [colorstr red "valkey-benchmark non zero code. first line: $first_line"]
+        puts [colorstr red "valkey-benchmark non zero code, the output is: $error"]
```
zuiderkwast: We can keep only the full output. We don't need only the first line.

```suggestion
        puts [colorstr red "valkey-benchmark non zero code, the output is: $error"]
```

zuiderkwast (approved):

tests/integration/valkey-benchmark.tcl
enjoy-binbin: LGTM, the change will be applied before merging, now waiting for CI to try to finish most of its work.

zuiderkwast:
@ranshid FYI

rjd15372 (approved):

PR #1712, opened 2025-02-11
Correctly detect address sanitizer with Clang

zuiderkwast:
Fixes the failure trying to use dlopen with DEEPBIND with ASAN when compiling with Clang:

==90510==You are trying to dlopen a /home/runner/work/valkey/valkey/tests/modules/keyspecs.so shared library with RTLD_DEEPBIND flag which is incompatible with sanitizer runtime (see https://github.com/google/sanitizers/issues/611 for details). If you want to run /home/runner/work/valkey/valkey/tests/modules/keyspecs.so library under sanitizers please remove RTLD_DEEPBIND from dlopen flags.

https://github.com/valkey-io/valkey/actions/runs/13261241213/job/37018133361

The previous check only covers GCC.

Additionally, don't require GLIBC when FreeBSD is used. FreeBSD has it's own libc which supports DEEPBIND according to its docs.

Follow-up of #1703, #1707

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -172,6 +172,16 @@
 #define VALKEY_NO_SANITIZE(sanitizer)
 #endif
 
+#if defined(__SANITIZE_ADDRESS__)
+/* GCC */
+#define VALKEY_ADDRESS_SANITIZER 1
+#elif defined(__has_feature)
+#if __has_feature(address_sanitizer)
+/* Clang */
+#define VALKEY_ADDRESS_SANITIZER 1
+#endif
+#endif
+
 /* Define rdb_fsync_range to sync_file_range() on Linux, otherwise we use
  * the plain fsync() call. */
 #if (defined(__linux__) && defined(SYNC_FILE_RANGE_WAIT_BEFORE))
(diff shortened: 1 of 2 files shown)


rjd15372 (approved):
@zuiderkwast  Thanks for fixing this!

src/config.h
```diff
@@ -172,6 +172,16 @@
...
 
+#if defined(__SANITIZE_ADDRESS__)
+/* GCC */
+#define VALKEY_ADDRESS_SANITIZER 1
+#elif defined(__has_feature)
+#if __has_feature(address_sanitizer)
+/* Clang */
+#define VALKEY_ADDRESS_SANITIZER 1
+#endif
+#endif
```
zuiderkwast, line 183: Another option is to use the GCC macro for this. I might be cleaner.

```suggestion
#elif defined(__has_feature) && !defined(__SANITIZE_ADDRESS__)
#if __has_feature(address_sanitizer)
/* Clang */
#define __SANITIZE_ADDRESS__
#endif
#endif
```

rjd15372: I  don't have a strong opinion, but I actually prefer the original code, since if someone wants to check how `VALKEY_ADDRESS_SANITIZER` is defined, it will learn that there are differences between GCC and Clang.

enjoy-binbin (approved):

Issue #1650, opened 2025-01-31
Valkey minor versioning for new releases


QuChen88:
I have discussed with various people offline about whether the next GA minor version should be 8.1 or 8.2. There seem to be some ambiguity in this topic so putting it here to facilitate discussions.

So far, all Redis official GA releases since the beginning of time are using even number for all minor versions, i.e. 6.2, 7.2, 7.4, etc. As mentioned in the release [documentation](https://cndoc.github.io/redis-doc-cn/cn/topics/releases.html), `the minor is even in stable versions of Redis. The minor is odd in unstable, development, frozen, release candidates.` Valkey also GA'ed major version 8.0 with an even minor version number. However, there hasn't been a new minor version release for Valkey. How to determine the minor version in releases is not explicitly mentioned in the Valkey release [docs](https://valkey.io/topics/releases/).

Here I think it still makes sense to have the next release candidate be on 8.1 version, and the GA version to be 8.2. The reason being that is how Redis and Valkey versioning worked in the past and it is good to keep a self-consistent versioning scheme going forward. For example - 

- In 7.2, the release candidate for 7.2.0 RC1 was versioned `7.1.240`, see https://github.com/valkey-io/valkey/blob/e26a769d9627ebecb8607375580970a740348956/src/version.h
- In 8.0, the release candidate for 8.0.0 RC1 was versioned `7.9.240`, see https://github.com/valkey-io/valkey/blob/8.0.0-rc1/src/version.h

Given this pattern, I think we should publish the next RC as `8.1.240`, and the GA version should be `8.2.0` to avoid confusion. 

Any thoughts?

madolson:
>  There seem to be some ambiguity in this topic so putting it here to facilitate discussions.

I'm not sure there is any ambiguity on the topic. We had previously decided to go with 8.1, we have advertised it as such in conferences, and it is the more common numbering scheme for semantic versioning.

> In 8.0, the release candidate for 8.0.0 RC1 was versioned 7.9.240, see https://github.com/valkey-io/valkey/blob/8.0.0-rc1/src/version.h

Officially it was versioned as 8.0.0-rc1, which is aligned with semantic versioning. We were using other versions historically since we were constrained by the 3 digit system.

QuChen88:
For major version releases, Redis also named those officially as 5.0 RC1, and 6.0 RC1, etc. 

[...]

PR #1713, opened 2025-02-11
Valkey release 8.1.0-rc1

ranshid:
This is the PR for 8.1.0 release. 
It includes: 
1. Release note changes
2. version changes

### DIFF
--- a/00-RELEASENOTES
+++ b/00-RELEASENOTES
@@ -17,8 +17,8 @@ Upgrade urgency LOW: This is the first release candidate of Valkey 8.1, with
 performance improvements, extended observability and cluster improvements and different bug fixes. 
 It includes a new implementation of the Valkey dictionary which is more memory and cache efficient, 
 better performance for encryption in transit, reduced replication overhead by offloading work to I/O threads,
-faster failover support in cluster mode, major improvements to the active defrag process to reduce the impact on command processing
-and different API changes for improved usability. 
+faster failover support in cluster mode, major improvements to the active defrag process to reduce the impact on command processing,
+different API changes for improved usability and ability to track large requests and replies. 
 Valkey now supports new new check-and-set feature for native STRINGs.
 
 API and Interface changes
--- a/00-RELEASENOTES
+++ b/00-RELEASENOTES
@@ -1,16 +1,142 @@
-Hello! This file is just a placeholder, since this is the "unstable" branch
-of Valkey, the place where all the development happens.
+Valkey 8.1 release notes
+========================
+--------------------------------------------------------------------------------
+Upgrade urgency levels:
+LOW:      No need to upgrade unless there are new features you want to use.
+MODERATE: Program an upgrade of the server, but it's not urgent.
+HIGH:     There is a critical bug that may affect a subset of users. Upgrade!
+CRITICAL: There is a critical bug affecting MOST USERS. Upgrade ASAP.
+SECURITY: There are security fixes in the release.
+--------------------------------------------------------------------------------
 
-There is no release notes for this branch, it gets forked into another branch
-every time there is a partial feature freeze in order to eventually create
-a new stable release.
 
-Usually "unstable" is stable enough for you to use it in development environments
-however you should never use it in production environments. It is possible
-to download the latest stable release here:
+================================================================================
+Valkey 8.1.0 RC1  -  Released Thu 11 Feb 2025
+================================================================================
+Upgrade urgency LOW: This is the first release candidate of Valkey 8.1, with
+performance improvements, extended observability and cluster improvements and different bug fixes. 
+It includes a new implementation of the Valkey dictionary which is more memory and cache efficient, 
+better performance for encryption in transit, reduced replication overhead by offloading work to I/O threads,
+faster failover support in cluster mode, major improvements to the active defrag process to reduce the impact on command processing
+and different API changes for improved usability. 
+Valkey now supports new new check-and-set feature for native STRINGs.
 
-    https://valkey.io/download/
+API and Interface changes
+=========================
+* Introduce cancel argument to bgsave command (#757)
+* Add conditional update support to the `SET` command using `IFEQ` argument (#1324)
+* Add more filters to `CLIENT LIST` (#1401)
+* Add `availability_zone` to the HELLO response (#1487)
 
-More information is available at https://valkey.io
+Observability and Monitoring changes
+====================================
+* Extend `LATENCY LATEST` to add sum / cnt stats (#1570)
+* Add `paused_actions` and `paused_timeout_milliseconds` for `INFO CLIENTS` (#1519)
+* Add paused_reason to `INFO CLIENTS` (#1564)
+* Added `COMMANDLOG` to record slow executions and large requests/replies (#1294)
+* Fix cluster info sent stats for message with light header (#1563)
+* Add latency stats around cluster config file operations (#1534)
+* Add new flag in `CLIENT LIST` for import-source client (#1398)
+* Show client capabilities in `CLIENT LIST` / `CLIENT INFO` (#1698)
 
-Happy hacking!
+Performance/Efficiency Improvements - Core
+==========================================
+* Introduce a new memory efficient hash table to store keys (#1186)
+* Accelerate hash table iterator with prefetching (#1501)
+* Accelerate hash table iterator with value prefetching (#1568)
+* Replace dict with new hashtable: hash datatype (#1502)
+* Replace dict with new hashtable for sets datatype (#1176)
+* Replace dict with new hashtable: sorted set datatype (#1427)
+* Free strings during BGSAVE/BGAOFRW to reduce copy-on-write (#905)
+* Create an empty lua table with specified initial capacity as much as possible (#1092)
+* Move prepareClientToWrite out of loop for HGETALL command (#1119)
+* Improved hashing algorithm for Lua tables (#1168)
+* Replace dict with new hashtable for sets datatype (#1176)
+* Do security attack check only when command not found to reduce the critical path. (#1212)
+* Trim free space from inline command argument strings to avoid excess memory usage (#1213)
+* Increase the max number of io threads to 256. (#1220)
+* Refactor of ActiveDefrag to reduce latencies (#1242)
+* Integrate fast_float to optionally replace strtod (#1260)
+* Improvements for TLS with I/O threads (#1271)
+* Optimize PFCOUNT, PFMERGE command by SIMD acceleration (#1293)
+* Optimize sdscatrepr by batch processing printable characters (#1342)
+* Optimize ZRANK to avoid path comparisons (#1389)
+* Move clientCron onto a separate timer (#1387)
+* Client struct: lazy init components and optimize struct layout (#1405)
+* Offload reading the replication stream to IO threads (#1449)
+* Skip CRC checksumming during diskless full sync with TLS enabled. (#1479)
+
+New/Modified configurations
+===========================
+* Deprecate `io-threads-do-reads`, which has no effect since io threads will now always do reads. (#1138)
+* Introduce `import-mode` config to avoid expiration and eviction during data syncing (#1185)
+* Introduce new `rdb-version-check` config which allows for relaxed RDB version verification (#1604)
+* Deprecate `dynamic-hz`, since server cron jobs are handled dynamically by default (#1387) 
+* Introduce `log-format` and `log-timestamp-format` to control the log format (#1022)
+* Introducing `active-defrag-cycle-us` for more fine-grinned control of memory defragmentation run time (#1242) 
+* Introduce new configurations to control the new `COMMANDLOG` reporting thresholds (#1294)
+
+Build and Packaging changes
+=================================
+* Introduce CMake build system for valkey (#1196)
+* RDMA builtin support (#1209)
+* Fix Valkey binary build workflow, version support changes. (#1429)
+* Remove Valkey specific changes in jemalloc source code (#1266)
+
+Module Improvements
+===================
+* Add API UpdateRuntimeArgs for updating the module arguments during runtime (#1041)
+* Add support for MustObeyClient Module API (#1582)
+* Adds support for scripting engines as Valkey modules (#1277, #1497)
+
+Cluster improvements
+====================
+* Do election in order based on failed primary rank to avoid voting conflicts (#1018)
+* Make replica `CLUSTER RESET` flush async based on `lazyfree-lazy-user-flush` (#1190)
+* Trigger the election as soon as possible when doing a forced manual failover (#1067)
+* Make manual failover reset the on-going election to promote failover (#1274)
+* Brocast a PONG to all node in cluster when role changed (#1295)
+* Manual failover vote is not limited by two times the node timeout (#1305)
+* Automatic failover vote is not limited by two times the node timeout (#1356)
+
+Behavior Changes
+================
+* Streams use an additional 8 bytes to track their internal size (#688)
+* Take hz into account in activerehashing to avoid CPU spikes (#977)
+* Incr `expired_keys` if the expiration time is already expired (#1517)
+* Fix replica not able to initiate election in time when epoch fails (#1009)
+* Make `FUNCTION RESTORE FLUSH` flush async based on `lazyfree-lazy-user-flush` (#1254)
+* Allow `MEMORY MALLOC-STATS` and `MEMORY PURGE` during loading phase (#1317)
+* Use `DEEPBIND` flag when loading external modules in order to avoid symbol conflicts (#1703)
+
+
+Logging and Tooling Improvements
+================================
+* Remove the restriction that cli --cluster create requires at least 3 primary nodes (#1075)
+* Add short client info log to CLUSTER MEET / FORGET / RESET commands (#1249)
+* Support for reading from replicas in valkey-benchmark (#1392)
+* valkey-cli will now re-select previously selected database after reconnect (#1694)
+* valkey-cli will now auto-exit from subscribed mode when there are no more active subscriptions (#1432)
+
+Bug Fixes
+=========
+* Mark the node as FAIL when the node is marked as NOADDR and broadcast the FAIL (#1191)
+* [Bug Fix] Optimize RDB Load Performance and Fix Cluster Mode Resizing (#1199)
+* Log as primary role (M) instead of child process (C) during startup (#1282)
+* Fix empty primary may have dirty slots data due to bad migration (#1285)
+* RDMA: Fix dead loop when transfer large data (20KB) (#1386)
+
+We appreciate the efforts of all who contributed code to this release!
+
+Karthik Subbarao (KarthikSubbarao), Xuyang WANG (Nugine), Yury Fridlyand (Yury-Fridlyand), Ben Totten (bentotten), Danish Mehmood (danish-mehmood),
+Eran Ifrah (eifrah-aws), gmbnomis, kronwerk, Stefan Mueller (muelstefamzn), Ran Shidlansik (ranshid), secwall, Tal Shachar (talxsha),
+Uri Yagelnik (uriyage), Basel Naamna (xbasel), Alan Scherger (flyinprogrammer), Amit Nagler (naglera), Binbin (enjoy-binbin),
+Caiyi Wu (Codebells), Guillaume Koenig (knggk), Harkrishn Patro (hpatro), Jacob Murphy (murphyjacob4), Jim Brunner (JimB123), Josef Šimánek (simi),
+Jungwoo Song (bluayer), Karthick Ariyaratnam (karthyuom), Lipeng Zhu (lipzhu), Madelyn Olson (madolson), Masahiro Ide (imasahiro), Melroy van den Berg (melroy89),
+Mikhail Koviazin (mkmkme), Nadav Gigi (NadavGigi), Nadav Levanoni (nadav-levanoni), Nikhil Manglore (Nikhil-Manglore), Parth Patel (parthpatel), Pierre (pieturin),
+Ping Xie (PingXie), Qu Chen (QuChen88), Rain Valentine (SoftlyRaining), Ray Cao (RayaCoo), Ricardo Dias (rjd15372), Romain Geissler (Romain-Geissler-1A),
+Roman Gershman (romange), Roshan Khatri (roshkhatri), Rueian (rueian), Sarthak Aggarwal (sarthakaggarwal97), Seungmin Lee (sungming2), Shai Zarka (zarkash-aws),
+Shivshankar (Shivshankar-Reddy), Sinkevich Artem (ArtSin), Stav Ben-Tov (stav-bentov), Thalia Archibald (thaliaarchi), Vadym Khoptynets (poiuj),
+Viktor Söderqvist (zuiderkwast), Viktor Szépe (szepeviktor), Vu Diep (vudiep411), Wen Hui (hwware), Yanqi Lv (lyq2333), Zvi Schneider (zvi-code),
+bodong.ybd (yangbodong22011), chx9 (chx9), otheng (otheng03), skyfirelee (artikell), xingbowang (xingbowang), zhaozhao.zz (soloestoy), zhenwei pi(pizhenwei),
+zixuan zhao (azuredream), 烈香 (hengyoush), 风去幽墨 (fengquyoumo)
--- a/src/version.h
+++ b/src/version.h
@@ -4,13 +4,13 @@
  * similar. */
 #define SERVER_NAME "valkey"
 #define SERVER_TITLE "Valkey"
-#define VALKEY_VERSION "255.255.255"
-#define VALKEY_VERSION_NUM 0x00ffffff
+#define VALKEY_VERSION "8.1.0"
+#define VALKEY_VERSION_NUM 0x00080100
 /* The release stage is used in order to provide release status information.
  * In unstable branch the status is always "dev".
  * During release process the status will be set to rc1,rc2...rcN.
  * When the version is released the status will be "ga". */
-#define VALKEY_RELEASE_STAGE "dev"
+#define VALKEY_RELEASE_STAGE "rc1"
 
 /* Redis OSS compatibility version, should never
  * exceed 7.2.x. */


ranshid:
@valkey-io/core-team I hope I did fine with the release notes - please review and I will make the necessary changes.

hwware (commented):
I do not check the details, but I think we miss the **_Thanks part_**, please 
reference this https://github.com/valkey-io/valkey/blob/7.2.8/00-RELEASENOTES

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
-Usually "unstable" is stable enough for you to use it in development environments
-however you should never use it in production environments. It is possible
-to download the latest stable release here:
+================================================================================
+Valkey 8.1.0 RC1  -  Released Thu 11 Feb 2025
+================================================================================
+Upgrade urgency LOW: This is the first release candidate of Valkey 8.1, with
+performance improvements, extended observability and cluster improvements and different bug fixes. 
+It includes a new implementation of the Valkey dictionary which is more memory and cache efficient, 
+better performance for encryption in transit, reduced replication overhead by offloading work to io-threads,
```
zuiderkwast:
```suggestion
better performance for encryption in transit, reduced replication overhead by offloading work to I/O threads,
```

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+faster failover support in cluster mode, major improvements to the active defrag process to reduce the impact on command processing
+and different API changes for improved usability.
+In this release valkey also supports CMake build system for enhanced building options and better IDE integration.
```
zuiderkwast:
```suggestion
In this release, Valkey also supports CMake build system for enhanced building options and better IDE integration. 
```

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
 
-Usually "unstable" is stable enough for you to use it in development environments
-however you should never use it in production environments. It is possible
-to download the latest stable release here:
+================================================================================
+Valkey 8.1.0 RC1  -  Released Thu 11 Feb 2025
+================================================================================
+Upgrade urgency LOW: This is the first release candidate of Valkey 8.1, with
+performance improvements, extended observability and cluster improvements and different bug fixes. 
+It includes a new implementation of the Valkey dictionary which is more memory and cache efficient,
```
zuiderkwast, line 18: Shall we mention the new *check-and-set* feature (IFEQ) here? We can promote it if we want to...

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
 
-    https://valkey.io/download/
+API and Interface changes
+=========================
+* Introduce bgsave cancel (#757)
+* Add paused_reason to INFO CLIENTS (#1564)
+* show capa in CLIENT LIST / CLIENT INFO (#1698)
+* Set Command with IFEQ Support (#1324)
+* Add new flag in `CLIENT LIST` for import-source client (#1398)
+* Adding Missing filters to CLIENT LIST and Dedup Parsing (#1401)
```
zuiderkwast: I think we can remove some irrelevant information about internal refactoring.

```suggestion
* Adding more filters to CLIENT LIST (#1401)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+* Manual failover vote is not limited by two times the node timeout (#1305)
+* Automatic failover vote is not limited by two times the node timeout (#1356)
+
+Behavior Changes
+================
+* Rax size tracking (#688)
+* Take hz into account in activerehashing to avoid CPU spikes (#977)
+* Incr expired_keys if the expiration time is already expired (#1517)
+* Fix replica not able to initiate election in time when epoch fails (#1009)
+* Revert "Decline unsubscribe related command in non-subscribed mode" (#1265)
```
zuiderkwast: This is backported to 8.0.x right? Is it then a new behaviour change in 8.1?

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+=================================
+* RDMA builtin support (#1209)
+* Fix Valkey binary build workflow, version support changes. (#1429)
+
+Code improvements
+=================
+* Remove valkey specific changes in jemalloc source code (#1266)
+
+Module Improvements
+===================
```
hwware, line 87: Please add PR https://github.com/valkey-io/valkey/pull/1041. It is related to module part, Thanks

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+* Optimize ZRANK to avoid path comparisons (#1389)
+* move clientCron onto a separate timer (#1387)
+* client struct: lazy init components and optimize struct layout (#1405)
+* Offload reading the replication stream to IO threads (#1449)
+* skip CRC checksumming during diskless full sync with TLS enabled. (#1479)
+* Accelerate hash table iterator with value prefetching (#1568)
+
+New/Modified configurations
+===========================
+* Add io-threads-do-reads config to deprecated config table to have no effect. (#1138)
```
zuiderkwast: A description like this fits better under New/Modified configurations:

```suggestion
* io-threads-do-reads is deprecated and has no effect. (#1138)
```

But wait, there are more. Shall we mention all?

Removed or deprecated configs:

* dynamic-hz

New configs:

* import-mode
* log-format
* log-timestamp-format
* rdb-version-check
* commandlog-execution-slower-than
* commandlog-slow-execution-max-len
* commandlog-request-larger-than
* commandlog-large-request-max-len
* commandlog-reply-larger-than
* commandlog-large-reply-max-len
* active-defrag-cycle-us

Exiting configs with new options:

* repl-diskless-load has a new value: `flush-before-load`

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+* Make manual failover reset the on-going election to promote failover (#1274)
+* Brocast a PONG to all node in cluster when role changed (#1295)
+* Manual failover vote is not limited by two times the node timeout (#1305)
+* Automatic failover vote is not limited by two times the node timeout (#1356)
+
+Behavior Changes
+================
+* Rax size tracking (#688)
+* Take hz into account in activerehashing to avoid CPU spikes (#977)
+* Incr expired_keys if the expiration time is already expired (#1517)
```
zuiderkwast: Incr in what kind o stats? Keyspace notification?? I don't remember but it might be good to mention here.

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+* RDMA: Fix dead loop when transfer large data (20KB) (#1386)
+* Fix: restore replica output bytes stat update (#1486)
+* Use the correct command proc for the LOOKUP_NOTOUCH exception in lookupKey (#1499)
+* Fix for when active rehashing kvstore dictionaries are replaced by defrag (#1512)
+* Fix module LatencyAddSample still work when latency-monitor-threshold is 0 (#1541)
+* Fix default value of primary-reboot-down-after-period in sentinel.conf (#1040)
+
+Experimental
+============
+* Relaxed RDB version check (#1604)
```
zuiderkwast: Experimental???

Last time we used that term, it meant that the feature can be removed or changed in any minor version.

This one is not experimental IMHO. It's quite important that it's not listed as an experimental feature. It may be a deal-breaker to some very paranoid users that I've heard about.

Can we just list it under new configs?

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+* Allow `MEMORY MALLOC-STATS` and `MEMORY PURGE` during loading phase (#1317)
+
+Logging and Tooling Improvements
+================================
+* Configurable log and timestamp formats (logfmt, ISO8601) (#1022)
+* Remove the restriction that cli --cluster create requires at least 3 primary nodes (#1075)
+* Add CMake build system for valkey (#1196)
+* Add short client info log to CLUSTER MEET / FORGET / RESET commands (#1249)
+* Support for reading from replicas in valkey-benchmark (#1392)
+* valkey-cli: re-select previously selected database after reconnect (#1694)
```
zuiderkwast: Add this one?

* valkey-cli: Auto-exist subscribed mode #1432

It doesn't have the release-notes label, but I just remembered it. We can add it if you think it's useful.

zuiderkwast (commented):
Good job!

Shall we make some noise about IFEQ? It might be a cool feature for some users.

00-RELEASENOTES
ranshid: I also had thoughts on that. If we backport something to 8.0 but it is in a minor release, minor release can also be AFTER we release a new MAJOR release right? So I did a compare against 8.0.0 and did not account for things be introduced to 8.0.1/2

00-RELEASENOTES
ranshid: Yes It made changes to the keyspace notifications. I can elaborate some more

00-RELEASENOTES
ranshid: I know. but @PingXie was somewhat reluctant to make a fuss about it since this is an "incomplete" feature right?

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+* Refactor of ActiveDefrag to reduce latencies (#1242)
+* Integrating fast_float to optionally replace strtod (#1260)
+* Improvements for TLS with I/O threads (#1271)
+* Optimize PFCOUNT, PFMERGE command by SIMD acceleration (#1293)
+* Optimize sdscatrepr by batch processing printable characters (#1342)
+* Optimize ZRANK to avoid path comparisons (#1389)
+* move clientCron onto a separate timer (#1387)
+* client struct: lazy init components and optimize struct layout (#1405)
+* Offload reading the replication stream to IO threads (#1449)
+* skip CRC checksumming during diskless full sync with TLS enabled. (#1479)
```
hwware:
```suggestion
* Skip CRC checksumming during diskless full sync with TLS enabled. (#1479)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+* Trim free space from inline command argument strings to avoid excess memory usage (#1213)
+* Increase the IO_THREADS_MAX_NUM. (#1220)
+* Refactor of ActiveDefrag to reduce latencies (#1242)
+* Integrating fast_float to optionally replace strtod (#1260)
+* Improvements for TLS with I/O threads (#1271)
+* Optimize PFCOUNT, PFMERGE command by SIMD acceleration (#1293)
+* Optimize sdscatrepr by batch processing printable characters (#1342)
+* Optimize ZRANK to avoid path comparisons (#1389)
+* move clientCron onto a separate timer (#1387)
+* client struct: lazy init components and optimize struct layout (#1405)
```
hwware:
```suggestion
* Client struct: lazy init components and optimize struct layout (#1405)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+* Replace dict with new hashtable: sorted set datatype (#1427)
+* Trim free space from inline command argument strings to avoid excess memory usage (#1213)
+* Increase the IO_THREADS_MAX_NUM. (#1220)
+* Refactor of ActiveDefrag to reduce latencies (#1242)
+* Integrating fast_float to optionally replace strtod (#1260)
+* Improvements for TLS with I/O threads (#1271)
+* Optimize PFCOUNT, PFMERGE command by SIMD acceleration (#1293)
+* Optimize sdscatrepr by batch processing printable characters (#1342)
+* Optimize ZRANK to avoid path comparisons (#1389)
+* move clientCron onto a separate timer (#1387)
```
hwware:
```suggestion
* Move clientCron onto a separate timer (#1387)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+* Replace dict with new hashtable for sets datatype (#1176)
+* Do security attack check only when command not found to reduce the critical path. (#1212)
+* Accelerate hash table iterator with prefetching (#1501)
+* Replace dict with new hashtable: hash datatype (#1502)
+* Replace dict with new hashtable for sets datatype (#1176)
+* Replace dict with new hashtable: sorted set datatype (#1427)
+* Trim free space from inline command argument strings to avoid excess memory usage (#1213)
+* Increase the IO_THREADS_MAX_NUM. (#1220)
+* Refactor of ActiveDefrag to reduce latencies (#1242)
+* Integrating fast_float to optionally replace strtod (#1260)
```
hwware:
```suggestion
* Integrate fast_float to optionally replace strtod (#1260)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+* Feature COMMANDLOG to record slow execution and large request/reply (#1294)
+* Fix cluster info sent stats for message with light header (#1563)
+* Add latency stats around cluster config file operations (#1534)
 
-Happy hacking!
+Performance/Efficiency Improvements - Core
+==========================================
+* New hash table (#1186)
+* Free strings during BGSAVE/BGAOFRW to reduce copy-on-write (#905)
+* Creates an empty lua table with specified initial capacity as much as possible (#1092)
```
hwware:
```suggestion
* Create an empty lua table with specified initial capacity as much as possible (#1092)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,146 @@
...
+* Add new flag in `CLIENT LIST` for import-source client (#1398)
+* Adding Missing filters to CLIENT LIST and Dedup Parsing (#1401)
+* Add `availability_zone` to the HELLO response (#1487)
+* Import-mode: Avoid expiration and eviction during data syncing (#1185)
 
-More information is available at https://valkey.io
+Observability and Monitoring changes
+====================================
+* Extend LATENCY LATEST to add sum / cnt stats (#1570)
+* add paused_actions for INFO Clients (#1519)
```
hwware:
```suggestion
* Add paused_actions for INFO Clients (#1519)
```

00-RELEASENOTES
hwware: Update Adding to Add

00-RELEASENOTES
ranshid: I can mention each of them, but it would cause some PR duplicaitons

ranshid: @zuiderkwast - check and see if it is better now?

00-RELEASENOTES
ranshid: this was the most complicated piece of work I have ever did in Valkey - @hwware can you take a look if that seems right?

ranshid:
> I do not check the details, but I think we miss the **_Thanks part_**, please reference this https://github.com/valkey-io/valkey/blob/7.2.8/00-RELEASENOTES

That was the most complicated work I ever did in Valkey. @hwware can you check the result looks good?

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+* Extend LATENCY LATEST to add sum / cnt stats (#1570)
+* Add paused_actions for INFO Clients (#1519)
+* Feature COMMANDLOG to record slow execution and large request/reply (#1294)
+* Fix cluster info sent stats for message with light header (#1563)
+* Add latency stats around cluster config file operations (#1534)
 
-Happy hacking!
+Performance/Efficiency Improvements - Core
+==========================================
+* New hash table (#1186)
```
madolson:
```suggestion
* Introduce a new memory efficient hash table to store keys (#1186)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+* Fix incorrect cache_memory reset in functionsLibCtxClear (#1255)
+* Log as primary role (M) instead of child process (C) during startup (#1282)
+* Fix empty primary may have dirty slots data due to bad migration (#1285)
+* Fix LRU crash when getting too many random lua scripts (#1310)
+* RDMA: Fix dead loop when transfer large data (20KB) (#1386)
+* Fix: restore replica output bytes stat update (#1486)
+* Use the correct command proc for the LOOKUP_NOTOUCH exception in lookupKey (#1499)
+* Fix for when active rehashing kvstore dictionaries are replaced by defrag (#1512)
+* Fix module LatencyAddSample still work when latency-monitor-threshold is 0 (#1541)
+* Fix default value of primary-reboot-down-after-period in sentinel.conf (#1040)
```
madolson:
```suggestion
* Mark the node as FAIL when the node is marked as NOADDR and broadcast the FAIL (#1191)
* [Bug Fix] Optimize RDB Load Performance and Fix Cluster Mode Resizing (#1199)
* Log as primary role (M) instead of child process (C) during startup (#1282)
* Fix empty primary may have dirty slots data due to bad migration (#1285)
* RDMA: Fix dead loop when transfer large data (20KB) (#1386)
```

Removing all the ones that are backported to 8.0, so they aren't bug fixes for this version technically.

00-RELEASENOTES
madolson: This is not backported as far as I can tell to 8.0.

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+* Manual failover vote is not limited by two times the node timeout (#1305)
+* Automatic failover vote is not limited by two times the node timeout (#1356)
+
+Behavior Changes
+================
+* Rax size tracking (#688)
+* Take hz into account in activerehashing to avoid CPU spikes (#977)
+* Incr expired_keys if the expiration time is already expired (#1517)
+* Fix replica not able to initiate election in time when epoch fails (#1009)
+* Revert "Decline unsubscribe related command in non-subscribed mode" (#1265)
```
madolson: This we reverted in 8.0.2, so I don't think we need to call it out here.

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+* Make replica CLUSTER RESET flush async based on lazyfree-lazy-user-flush (#1190)
+* Trigger the election as soon as possible when doing a forced manual failover (#1067)
+* Make manual failover reset the on-going election to promote failover (#1274)
+* Brocast a PONG to all node in cluster when role changed (#1295)
+* Manual failover vote is not limited by two times the node timeout (#1305)
+* Automatic failover vote is not limited by two times the node timeout (#1356)
+
+Behavior Changes
+================
+* Rax size tracking (#688)
```
madolson:
```suggestion
* Streams use an additional 8 bytes to track their internal size (#688)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+* Fix Valkey binary build workflow, version support changes. (#1429)
+
+Code improvements
+=================
+* Remove Valkey specific changes in jemalloc source code (#1266)
+
+Module Improvements
+===================
+* Add API UpdateRuntimeArgs for updating the module arguments during runtime (#1041)
+* Use `DEEPBIND` flag when loading external modules using `dlopen` (#1703)
```
madolson: I would add this to the behavior change section, also maybe clarify the relationship to modules.

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+
+Build and Packaging changes
+=================================
+* RDMA builtin support (#1209)
+* Fix Valkey binary build workflow, version support changes. (#1429)
+
+Code improvements
+=================
+* Remove Valkey specific changes in jemalloc source code (#1266)
+
```
madolson:
```suggestion
```
I would move this to the performance section. The point wasn't to improve the code, but to make jemalloc more efficient.

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+* Relaxed RDB version check (#1604)
+* Deprecate dynamic-hz feature (#1387) 
+* New configuration to control the log format (#1022)
+* Introducing active-defrag-cycle-us for more fine-grinned control of ActiveDefrag run time (#1242) 
+* Introducing new configurations to control the new COMMANDLOG reporting thresholds (#1294)
+
+Build and Packaging changes
+=================================
+* RDMA builtin support (#1209)
+* Fix Valkey binary build workflow, version support changes. (#1429)
```
madolson: Maybe put CMAKE here?

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+* Optimize ZRANK to avoid path comparisons (#1389)
+* Move clientCron onto a separate timer (#1387)
+* Client struct: lazy init components and optimize struct layout (#1405)
+* Offload reading the replication stream to IO threads (#1449)
+* Skip CRC checksumming during diskless full sync with TLS enabled. (#1479)
+* Accelerate hash table iterator with value prefetching (#1568)
+
+New/Modified configurations
+===========================
+* io-threads-do-reads is deprecated and has no effect. (#1138)
```
madolson:
```suggestion
* io-threads-do-reads is deprecated and has no effect since io threads will now always do reads. (#1138)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+Performance/Efficiency Improvements - Core
+==========================================
+* New hash table (#1186)
+* Free strings during BGSAVE/BGAOFRW to reduce copy-on-write (#905)
+* Create an empty lua table with specified initial capacity as much as possible (#1092)
+* Move prepareClientToWrite out of loop for HGETALL command (#1119)
+* Improved hashing algorithm in luaS_newlstr (#1168)
+* Replace dict with new hashtable for sets datatype (#1176)
+* Do security attack check only when command not found to reduce the critical path. (#1212)
+* Accelerate hash table iterator with prefetching (#1501)
```
madolson: Since these aren't ordered, maybe let's put all of the hash table ones together

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+* Replace dict with new hashtable: hash datatype (#1502)
+* Replace dict with new hashtable for sets datatype (#1176)
+* Replace dict with new hashtable: sorted set datatype (#1427)
+* Trim free space from inline command argument strings to avoid excess memory usage (#1213)
+* Increase the IO_THREADS_MAX_NUM. (#1220)
```
madolson:
```suggestion
* Increase the max number of io threads to 256. (#1220)
```
This is more a configuration as well.

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+and different API changes for improved usability. 
+Valkey now supports new new check-and-set feature for native STRINGs.
+In this release Valkey also supports CMake build system for enhanced building options and better IDE integration. 
 
-    https://valkey.io/download/
+API and Interface changes
+=========================
+* Introduce bgsave cancel (#757)
+* Add paused_reason to INFO CLIENTS (#1564)
+* show capa in CLIENT LIST / CLIENT INFO (#1698)
```
madolson:
```suggestion
* Show client capabilities in CLIENT LIST / CLIENT INFO (#1698)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+better performance for encryption in transit, reduced replication overhead by offloading work to I/O threads,
+faster failover support in cluster mode, major improvements to the active defrag process to reduce the impact on command processing
+and different API changes for improved usability. 
+Valkey now supports new new check-and-set feature for native STRINGs.
+In this release Valkey also supports CMake build system for enhanced building options and better IDE integration. 
 
-    https://valkey.io/download/
+API and Interface changes
+=========================
+* Introduce bgsave cancel (#757)
```
madolson:
```suggestion
* Introduce cancel argument to bgsave command (#757)
```

madolson (commented):
Cool. Mostly comment on lines which were most confusing to me.

00-RELEASENOTES
zuiderkwast: Yes, very good!

00-RELEASENOTES
zuiderkwast: Also to improve the code by un-patching the dependency. I've seen some distro patch our Makefile to use system jemalloc instead. They didn't get defrag before but now they do. Found it:

https://gitlab.archlinux.org/archlinux/packaging/packages/valkey

Specifically this patch file:

https://gitlab.archlinux.org/archlinux/packaging/packages/valkey/-/blob/main/valkey-5.0-use-system-jemalloc.patch

00-RELEASENOTES
zuiderkwast: Regarding "Improved hashing algorithm in luaS_newlstr": this title is hard to get. It improves performance of Lua tables, optimizing Lua scripts in general.

Suggestion:

    * Improved hashing algorithm for Lua tables (#1168)

00-RELEASENOTES
```diff
@@ -141,6 +149,17 @@ Bug Fixes
 * Fix module LatencyAddSample still work when latency-monitor-threshold is 0 (#1541)
 * Fix default value of primary-reboot-down-after-period in sentinel.conf (#1040)
 
-Experimental
-============
-* Relaxed RDB version check (#1604)
+We appreciate the efforts of all who contributed code to this release!
+
+Karthik Subbarao(KarthikSubbarao),Xuyang WANG(Nugine),Yury Fridlyand(Yury-Fridlyand),Ben Totten(bentotten),Danish Mehmood(danish-mehmood),
```
zuiderkwast: Add a space after the commas.

00-RELEASENOTES
madolson: That's a good point, we could also put in the build part. I think we are close, like 90% of the way towards allowing someone to bring their own jemalloc version.

00-RELEASENOTES
zuiderkwast: And space before `(`

00-RELEASENOTES
```diff
@@ -1,16 +1,165 @@
...
+faster failover support in cluster mode, major improvements to the active defrag process to reduce the impact on command processing
+and different API changes for improved usability. 
+Valkey now supports new new check-and-set feature for native STRINGs.
+In this release Valkey also supports CMake build system for enhanced building options and better IDE integration. 
 
-    https://valkey.io/download/
+API and Interface changes
+=========================
+* Introduce bgsave cancel (#757)
+* Add paused_reason to INFO CLIENTS (#1564)
```
soloestoy: it's better put it into `Observability and Monitoring changes` with `#1519` together

00-RELEASENOTES
soloestoy: capa and new flag in `CLIENT LIST` are both belong to Observability?

src/version.h
```diff
@@ -4,8 +4,8 @@
  * similar. */
 #define SERVER_NAME "valkey"
 #define SERVER_TITLE "Valkey"
-#define VALKEY_VERSION "255.255.255"
-#define VALKEY_VERSION_NUM 0x00ffffff
+#define VALKEY_VERSION "8.1.0"
+#define VALKEY_VERSION_NUM 0x00080100
```
enjoy-binbin, line 8: i think we have a different version for RC? something like this:
```
// 8.0 RC1
#define VALKEY_VERSION "7.9.240"
#define VALKEY_VERSION_NUM 0x000709f0
```

00-RELEASENOTES
ranshid: I removed it. did not understand the bottom line, but I think it was introduced to 8.0.2 and which is why this can be removed

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
 
-More information is available at https://valkey.io
+Observability and Monitoring changes
+====================================
+* Extend LATENCY LATEST to add sum / cnt stats (#1570)
+* Add paused_actions for INFO Clients (#1519)
+* Feature COMMANDLOG to record slow execution and large request/reply (#1294)
+* Fix cluster info sent stats for message with light header (#1563)
+* Add latency stats around cluster config file operations (#1534)
+* Add paused_reason to INFO CLIENTS (#1564)
```
soloestoy:
```suggestion
* Add paused_actions and paused_timeout_milliseconds for INFO Clients (#1519)
* Add paused_reason to INFO CLIENTS (#1564)
* Feature COMMANDLOG to record slow execution and large request/reply (#1294)
* Fix cluster info sent stats for message with light header (#1563)
* Add latency stats around cluster config file operations (#1534)
* Add new flag in `CLIENT LIST` for import-source client (#1398)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+and different API changes for improved usability. 
+Valkey now supports new new check-and-set feature for native STRINGs.
+In this release Valkey also supports CMake build system for enhanced building options and better IDE integration. 
 
-    https://valkey.io/download/
+API and Interface changes
+=========================
+* Introduce cancel argument to bgsave command (#757)
+* Set Command with IFEQ Support (#1324)
+* Add new flag in `CLIENT LIST` for import-source client (#1398)
```
soloestoy:
```suggestion
```

src/version.h
ranshid: O.K will try using 8.0.240

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Fix empty primary may have dirty slots data due to bad migration (#1285)
+* Fix LRU crash when getting too many random lua scripts (#1310)
+* RDMA: Fix dead loop when transfer large data (20KB) (#1386)
+* Fix: restore replica output bytes stat update (#1486)
+* Use the correct command proc for the LOOKUP_NOTOUCH exception in lookupKey (#1499)
+* Fix for when active rehashing kvstore dictionaries are replaced by defrag (#1512)
+* Fix module LatencyAddSample still work when latency-monitor-threshold is 0 (#1541)
+* Fix default value of primary-reboot-down-after-period in sentinel.conf (#1040)
+
+We appreciate the efforts of all who contributed code to this release!
```
soloestoy, line 129: Is this based on the code contribution ranking of version 8.1? What tool was used?

ranshid: @soloestoy I just used github cli to get all the PR contributors of merged PRs which were merged AFTER 8.0.0 - is there another preferred way?

soloestoy: I'm just curious : )

ranshid: something in the form of:
```
gh pr list --limit 100000 --state merged --json mergedAt,author --jq '.[] | select(.mergedAt > "'"$(git show -s --format=%ci $TAG_COMMIT_HASH)"'") | "\(.author.name)(\(.author.login))"' | sort -u | paste -sd "," -
```

ranshid:
@valkey-io/core-team Thank you all for reviewing this!
I made attempt to address all of your comments.
Can you please approve once you are O.K with this content so that I know when is this good to me merged?

As a side note: For each contribution I took the PR title as the contribution statement. I think all your comments were good and valid, but I suggest for the future we should take extra care about reviewing and fixing the PR titles before merging them. specially when the PR is tagged with release-notes.

src/version.h
ranshid: @enjoy-binbin changing this breaks many tests as the new commands/subcommands (eg IFEQ) are written in the json with 8.1 support...

00-RELEASENOTES
zuiderkwast: :medal_sports: Let's add this to the wiki page about how to make releases.

src/version.h
zuiderkwast: https://github.com/valkey-io/valkey/issues/1650

We discussed this in the meeting last week. I think we were leaning towards using 8.1.0 in release candidates and GA, but maybe we didn't make a decision... Using 8.0.240 was the other alternative. We said that it's unlikely that users depend on this in release candidates though.

@enjoy-binbin for example in #939, do you consider REPLCONF VERSION 8.0.240 an old or a new replica?

enjoy-binbin: Maybe all RCs can be considered an old version, i am not sure about it. I don't know...

I have an old memory is that Redis can make breaking change to change something in the past RC. Maybe something like https://github.com/redis/redis/pull/11058/

zuiderkwast: I think we can use 8.0.240 but we should probably introduce some function to compare versions internally to use in features like #939, such as

```c
bool isVersionGreaterOrEqual(int numeric, int major, int minor, int patch, int rc) {
    int x = (major << 16) | (minor << 8) | patch;
    if (rc) x = x - 17 + rc;
    return numeric >= x;
}
```

*EDIT*: or better, a function to convert to numeric. Then in #939 we can check like this:

```c
    if (replica_version >= numericVersion(8, 1, 0, 2)) {
        /* Replica supports CLUSTER FAILOVER REPLICAID */
    }
```

enjoy-binbin: emmm, i did not realize we are doing some version checking in the tests, it is odd i must say.

zuiderkwast: > Maybe all RCs can be considered an old version, i am not sure about it. I don't know...

If we consider RCs old versions, it means features like #939 don't work in RCs. :/

> I have an old memory is that Redis can make breaking change to change something in the past RC

Yes, I think you're right. There is no compatibility guarantee between RC. Th semantic versioning definition says the same for pre-releases: https://semver.org/#spec-item-9 *"A pre-release version indicates that the version is unstable and might not satisfy the intended compatibility requirements as denoted by its associated normal version."*

ranshid:
So we hit some issue withRC versioning: 
Currently, the way it was done in past releases, we used to follow the last major with PATCH number starting from 240.
This works for most cases but there are few caveats:
1. For external clients (eg valkey-cli) tests the support for new commands might break since new API are tagged with version support >= than the released version. for example:  https://github.com/valkey-io/valkey/actions/runs/13281314141/job/37080127772?pr=1713
2. Some new implementations/tests rely on specific version checking (eg https://github.com/valkey-io/valkey/pull/1091/files)

There are several options we can choose:
1. keep the version formatting we decided on. This would also require us to fix the current tests as well as implement some version check logic in the code.
For example having this code:
```
if (replica->repl_data->replica_version < 0x80100) 
```

might have to be changed to support version like 8.0.240 is treated the same as 8.1.0.
For example we could have functions providing version data and know how to treat the 240 and higher patch version logic, so it can later be used like:
```
if (valkeyGetMajorVersion(replica->repl_data->replica_version) < 8 || 
     valkeyGetMinorVersion(replica->repl_data->replica_version) < 1) 
```

2. giveup on the current minor version thing and still use 0,2,4,6,8 for GA versions and 1,3,5,7,9 for RCs

Other alternatives are welcome (We discussed some alternatives in the last maintainers meeting, but I was unable to recover all the options)

zuiderkwast:
We were discussing some alternatives in a chat today. Some options are listed below. Whatever we decide, we can fix the failing tests that depend on version.

We can also consider the numeric VALKEY_VERSION_NUM. It's visible in the Lua API and module API.

* Use `"8.0.240"`.
  * Pros: Most similar to old versioning scheme
  * Cons: Looks like a patch version of 8.0.
* Use `"8.1.0"`.
  * Pros:
    * No broken tests.
    * COMMAND INFO "since" fields are correct for clients.
  * Cons:
    * No difference between release candidates.
    * May lead to compatibility issues between release candidates if we merge #1091 in RC2.
* Use `"8.1.0-rc1"` as version in INFO and HELLO and internally use numeric version 0x800f0 (8.0.240).
  * Pros:
    * The most correctly reported version. 
    * RC visible in INFO and HELLO.
    * one-to-one conversion between numeric and string version.
    * Can prevent compatibility problems between release candidates.
  * Cons:
    * Can break some clients, but only in the release candidates.

ranshid:
> We were discussing some alternatives in a chat today. Some options are listed below. Whatever we decide, we can fix the failing tests that depend on version.
> 
> We can also consider the numeric VALKEY_VERSION_NUM. It's visible in the Lua API and module API.
> 
> * Use `"8.0.240"`.
>   
>   * Pros: Most similar to old versioning scheme
>   * Cons: Looks like a patch version of 8.0.

I think there are more cons here:
 * It breaks some valkey-cli compatibility tests which we will have to fix and track for future testing
 * code which aims to take version considerations will have to consider RC versions. For example 
 https://github.com/valkey-io/valkey/pull/1091 is doing:
```
if (replica->repl_data->replica_version < 0x80100)
```
which will have to change to:
```
if (replica->repl_data->replica_version < 0x800f1)
```
which is strange given no one remembers RCs later on.
> * Use `"8.1.0"`.
>   
>   * Pros:
>     
>     * No broken tests.
>     * COMMAND INFO "since" fields are correct for clients.
>   * Cons:
>     
>     * No difference between release candidates.
>     * May lead to compatibility issues between release candidates if we merge [Trigger manual failover on SIGTERM / shutdown to cluster primary #1091](https://github.com/valkey-io/valkey/pull/1091) in RC2.

I actually like this option because it is the simplest one. I think we could improve the observability by introducing a "release-status" indication somewhere in the info section which will be rc1/2/3/...n for release candidates and 'ga' once we release the final candidate.

regarding the rc breakage - I am not sure how much effort we need to take for keeping RCs compatible? IMO we could offer no guarantees regarding RCs compatibility since they are released for a very limited time window.
  
> * Use `"8.1.0-rc1"` as version in INFO and HELLO and internally use numeric version 0x800f0 (8.0.240).
>   
>   * Pros:
>     
>     * The most correctly reported version.
>     * RC visible in INFO and HELLO.
>     * one-to-one conversion between numeric and string version.
>     * Can prevent compatibility problems between release candidates.
>   * Cons:
>     
>     * Can break some clients, but only in the release candidates.

I think this option is good because of the clarity in the reporting. I think the cons are missing some maintenance effort and confusion that can happen when developing future "version aware" features and the need to understand which version to use (the VALKEY_VERSION or the VALKEY_VERSION_NUM). but I suppose these will not happen very often.

00-RELEASENOTES
ranshid: >Let's add this to the wiki page about how to make releases. 

Oh after I am done with this release I will make sure to document all the steps. For example it was not clear that we need to create a branch (8.1) first, and the steps to create it are not trivial. also the way I gather the commits was:

```
gh pr list --limit 100000 --state merged --label "release-notes" --json number,mergedAt,title   --jq '.[] | select(.mergedAt > "'"$(git show -s --format=%ci $TAG_COMMIT_HASH)"'") | "\(.title) (#\(.number))"'
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Introduce CMake build system for valkey (#1196)
+* RDMA builtin support (#1209)
+* Fix Valkey binary build workflow, version support changes. (#1429)
+* Remove Valkey specific changes in jemalloc source code (#1266)
+
+Module Improvements
+===================
+* Add API UpdateRuntimeArgs for updating the module arguments during runtime (#1041)
+* Add support for MustObeyClient Module API (#1582)
+* Adds support for scripting engines as Valkey modules (#1277)
```
rjd15372:
```suggestion
* Adds support for scripting engines as Valkey modules (#1277, #1497)
```

We should also add the second PR reference to this line.

src/version.h
zuiderkwast: > emmm, i did not realize we are doing some version checking in the tests, it is odd i must say.

It's not only the tests. `valkey-cli` checks the version of the connected server to decide which arguments to show in the command hints. For example to show or hide IFEQ. I have a fix for it in #1717 if we want to use 8.0.240 style version for RCs.

madolson: In the core meeting we had decided to store the "real" version. Which would be 8.1.0 in this case, so that clients can actually validate the intended behavior.

Not quite following what is the concern here?

madolson:
> Use "8.1.0".

Would be my preference. The risk of breaking clients in the release candidates seems like the biggest thing that would prevent testing. We also talked about adding a separate field that "only" exists during the release candidate.

src/version.h
zuiderkwast: That's what we decided I think, but we didn't write it down. :) https://github.com/valkey-io/valkey/wiki/Weekly-meeting and the Q's issue have no mentions of this... Let's write down what we still remember from the meeting.

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+faster failover support in cluster mode, major improvements to the active defrag process to reduce the impact on command processing
+and different API changes for improved usability. 
+Valkey now supports new new check-and-set feature for native STRINGs.
+In this release Valkey also supports CMake build system for enhanced building options and better IDE integration. 
 
-    https://valkey.io/download/
+API and Interface changes
+=========================
+* Introduce cancel argument to bgsave command (#757)
+* Set Command with IFEQ Support (#1324)
```
madolson:
```suggestion
* Add conditional update support to the `SET` command using `IFEQ` argument (#1324)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Add more filters to CLIENT LIST (#1401)
```
madolson:
```suggestion
* Add more filters to `CLIENT LIST` (#1401)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Add `availability_zone` to the HELLO response (#1487)
 
-More information is available at https://valkey.io
+Observability and Monitoring changes
+====================================
+* Extend LATENCY LATEST to add sum / cnt stats (#1570)
```
madolson:
```suggestion
* Extend `LATENCY LATEST` to add sum / cnt stats (#1570)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Add paused_actions and paused_timeout_milliseconds for INFO Clients (#1519)
```
madolson:
```suggestion
* Add `paused_actions` and `paused_timeout_milliseconds` for `INFO CLIENTS` (#1519)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Add paused_reason to INFO CLIENTS (#1564)
```
madolson:
```suggestion
* Add paused_reason to `INFO CLIENTS` (#1564)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Feature COMMANDLOG to record slow execution and large request/reply (#1294)
```
madolson:
```suggestion
* Added `COMMANDLOG` to record slow executions and large requests/replies (#1294)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Fix cluster info sent stats for message with light header (#1563)
+* Add latency stats around cluster config file operations (#1534)
+* Add new flag in `CLIENT LIST` for import-source client (#1398)
+* Show client capabilities in CLIENT LIST / CLIENT INFO (#1698)
```
madolson:
```suggestion
* Show client capabilities in `CLIENT LIST` / `CLIENT INFO` (#1698)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Optimize sdscatrepr by batch processing printable characters (#1342)
+* Optimize ZRANK to avoid path comparisons (#1389)
+* Move clientCron onto a separate timer (#1387)
+* Client struct: lazy init components and optimize struct layout (#1405)
+* Offload reading the replication stream to IO threads (#1449)
+* Skip CRC checksumming during diskless full sync with TLS enabled. (#1479)
+
+New/Modified configurations
+===========================
+* io-threads-do-reads is deprecated and has no effect since io threads will now always do reads. (#1138)
```
madolson:
```suggestion
* Deprecate `io-threads-do-reads`, which has no effect since io threads will now always do reads. (#1138)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Import-mode: Avoid expiration and eviction during data syncing (#1185)
```
madolson:
```suggestion
* Introduce `import-mode` config to avoid expiration and eviction during data syncing (#1185)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Relaxed RDB version check (#1604)
```
madolson:
```suggestion
* Introduce new `rdb-version-check` config which allows for relaxed RDB version verification (#1604)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Deprecate dynamic-hz feature (#1387)
```
madolson:
```suggestion
* Deprecate `dynamic-hz`, since server cron jobs are handled dynamically by default (#1387) 
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Skip CRC checksumming during diskless full sync with TLS enabled. (#1479)
+
+New/Modified configurations
+===========================
+* io-threads-do-reads is deprecated and has no effect since io threads will now always do reads. (#1138)
+* Import-mode: Avoid expiration and eviction during data syncing (#1185)
+* Relaxed RDB version check (#1604)
+* Deprecate dynamic-hz feature (#1387) 
+* New configuration to control the log format (#1022)
+* Introducing active-defrag-cycle-us for more fine-grinned control of ActiveDefrag run time (#1242)
```
madolson:
```suggestion
* Introducing `active-defrag-cycle-us` for more fine-grinned control of memory defragmentation run time (#1242) 
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Offload reading the replication stream to IO threads (#1449)
+* Skip CRC checksumming during diskless full sync with TLS enabled. (#1479)
+
+New/Modified configurations
+===========================
+* io-threads-do-reads is deprecated and has no effect since io threads will now always do reads. (#1138)
+* Import-mode: Avoid expiration and eviction during data syncing (#1185)
+* Relaxed RDB version check (#1604)
+* Deprecate dynamic-hz feature (#1387) 
+* New configuration to control the log format (#1022)
```
madolson:
```suggestion
* Introduce `log-format` and `log-timestamp-format` to control the log format (#1022)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Introducing active-defrag-cycle-us for more fine-grinned control of ActiveDefrag run time (#1242) 
+* Introducing new configurations to control the new COMMANDLOG reporting thresholds (#1294)
```
madolson:
```suggestion
* Introduce new configurations to control the new `COMMANDLOG` reporting thresholds (#1294)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+Module Improvements
+===================
+* Add API UpdateRuntimeArgs for updating the module arguments during runtime (#1041)
+* Add support for MustObeyClient Module API (#1582)
+* Adds support for scripting engines as Valkey modules (#1277, #1497)
+
+Cluster improvements
+====================
+* Do election in order based on failed primary rank to avoid voting conflicts (#1018)
+* Make replica CLUSTER RESET flush async based on lazyfree-lazy-user-flush (#1190)
```
madolson:
```suggestion
* Make replica `CLUSTER RESET` flush async based on `lazyfree-lazy-user-flush` (#1190)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Manual failover vote is not limited by two times the node timeout (#1305)
+* Automatic failover vote is not limited by two times the node timeout (#1356)
+
+Behavior Changes
+================
+* Streams use an additional 8 bytes to track their internal size (#688)
+* Take hz into account in activerehashing to avoid CPU spikes (#977)
+* Incr expired_keys if the expiration time is already expired (#1517)
+* Fix replica not able to initiate election in time when epoch fails (#1009)
+* Make FUNCTION RESTORE FLUSH flush async based on lazyfree-lazy-user-flush (#1254)
```
madolson:
```suggestion
* Make `FUNCTION RESTORE FLUSH` flush async based on `lazyfree-lazy-user-flush` (#1254)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Use `DEEPBIND` flag when loading external modules in order to avoid symbol conflicts (#1703)
+
+
+Logging and Tooling Improvements
+================================
+* Configurable log and timestamp formats (logfmt, ISO8601) (#1022)
+* Remove the restriction that cli --cluster create requires at least 3 primary nodes (#1075)
+* Add short client info log to CLUSTER MEET / FORGET / RESET commands (#1249)
+* Support for reading from replicas in valkey-benchmark (#1392)
+* valkey-cli: re-select previously selected database after reconnect (#1694)
```
madolson:
```suggestion
* valkey-cli will now re-select previously selected database after reconnect (#1694)
```

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* valkey-cli auto-exit from subscribed mode (#1432)
```
madolson:
```suggestion
* valkey-cli will now auto-exit from subscribed mode when there are no more subscribers (#1432)
```
Otherwise this sounds like such a funny feature :D

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Incr expired_keys if the expiration time is already expired (#1517)
+* Fix replica not able to initiate election in time when epoch fails (#1009)
+* Make FUNCTION RESTORE FLUSH flush async based on lazyfree-lazy-user-flush (#1254)
+* Allow `MEMORY MALLOC-STATS` and `MEMORY PURGE` during loading phase (#1317)
+* Use `DEEPBIND` flag when loading external modules in order to avoid symbol conflicts (#1703)
+
+
+Logging and Tooling Improvements
+================================
+* Configurable log and timestamp formats (logfmt, ISO8601) (#1022)
```
madolson:
```suggestion
```
This is a duplicate.

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+Valkey 8.1.0 RC1  -  Released Thu 11 Feb 2025
+================================================================================
+Upgrade urgency LOW: This is the first release candidate of Valkey 8.1, with
+performance improvements, extended observability and cluster improvements and different bug fixes. 
+It includes a new implementation of the Valkey dictionary which is more memory and cache efficient, 
+better performance for encryption in transit, reduced replication overhead by offloading work to I/O threads,
+faster failover support in cluster mode, major improvements to the active defrag process to reduce the impact on command processing
+and different API changes for improved usability. 
+Valkey now supports new new check-and-set feature for native STRINGs.
+In this release Valkey also supports CMake build system for enhanced building options and better IDE integration.
```
madolson:
```suggestion
```

I think this is important, but as a one liner end users don't really care about this.

00-RELEASENOTES
```diff
@@ -1,16 +1,162 @@
...
+* Make manual failover reset the on-going election to promote failover (#1274)
+* Brocast a PONG to all node in cluster when role changed (#1295)
+* Manual failover vote is not limited by two times the node timeout (#1305)
+* Automatic failover vote is not limited by two times the node timeout (#1356)
+
+Behavior Changes
+================
+* Streams use an additional 8 bytes to track their internal size (#688)
+* Take hz into account in activerehashing to avoid CPU spikes (#977)
+* Incr expired_keys if the expiration time is already expired (#1517)
```
madolson:
```suggestion
* Incr `expired_keys` if the expiration time is already expired (#1517)
```

madolson (commented):
I feel more nit picky, but just generally trying to use ` consistently as well as trying to make the main paragraph a little bit more relevant and succinct.

00-RELEASENOTES
madolson: This wasn't fully addressed, some of these were left, was there a reason for this?

00-RELEASENOTES
zuiderkwast: Conditional update/insert, more famously known as check-and-set?

```suggestion
* Add check-and-set support to the `SET` command using `IFEQ` argument (#1324)
```

00-RELEASENOTES
zuiderkwast: This file isn't a markdown file though. Should we make it actually render as markdown?

00-RELEASENOTES
zuiderkwast: subscribers -> subscriptions? The client is one subscriber, to multiple channels/patterns.

madolson: Subscribers must have at least one subscription, so I suppose either is fine.

00-RELEASENOTES
madolson: It does render as markdown in the release page.

PingXie (approved):
LGTM. Great work! Thanks @ranshid

00-RELEASENOTES
madolson: Is it more famously known? I'm indifferent in either case.

madolson (approved):
Ok, I'm happy with it now. The only thing I'm just now noticing is commandlog is missing from top level comment, but other than that I think it's pretty good.

zuiderkwast (approved):

ranshid:
> Ok, I'm happy with it now. The only thing I'm just now noticing is commandlog is missing from top level comment, but other than that I think it's pretty good.

what do you mean by commandlog? you mean the slow commands log?

zuiderkwast:
> what do you mean by commandlog? you mean the slow commands log?

Yes, slow and fat. Large request and large response log.

It's an admin feature. Maybe not that exciting for users? They only care when they have a problem... WDYT?

madolson:
> Yes, slow and fat. Large request and large response log.

Yeah, there is a new command called commandlog for observability. Which has been a pretty commonly requested features.

hwware (approved):

enjoy-binbin (approved):

ranshid:
> > Yes, slow and fat. Large request and large response log.
> 
> Yeah, there is a new command called commandlog for observability. Which has been a pretty commonly requested features.

I know the new feature, just did not understand it's relation to the top-comment. you mean the top comment in the notes. I understand now. I can add something. thank you!

soloestoy (approved):
LGTM, and agree it's better to mention commandlog in top level

PR #1714, opened 2025-02-11
Fix valkey-benchmark FCALL failure with TLS

xbasel:
The recently added `FCALL` support in valkey-benchmark opens a connection to load a script but always uses a non-TLS connection, even when TLS is enabled. This causes protocol errors.

The test `integration/valkey-benchmark` fails with: 
```
valkey-benchmark: valkey-benchmark.c:2028: main: Assertion `reply != NULL' failed.
```

This fix ensures the correct connection wrapper is used.

Fixes a bug introduced by https://github.com/valkey-io/valkey/pull/1578

### DIFF
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -2020,22 +2020,17 @@ int main(int argc, char **argv) {
         if (test_is_selected("fcall")) {
             char *script = generateFunctionScript(1, config.num_keys_in_fcall > 0);
 
-            char *ip = config.conn_info.hostip;
-            int port = config.conn_info.hostport;
-            redisContext *conn = redisConnect(ip, port);
-            assert(conn != NULL);
-            if (config.tls == 1) {
-                const char *err = NULL;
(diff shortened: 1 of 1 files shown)


src/valkey-benchmark.c
```diff
@@ -2020,22 +2020,17 @@ int main(int argc, char **argv) {
...
-            int port = config.conn_info.hostport;
-            redisContext *conn = redisConnect(ip, port);
-            assert(conn != NULL);
-            if (config.tls == 1) {
-                const char *err = NULL;
-                if (cliSecureConnection(conn, config.sslconfig, &err) == REDIS_ERR && err) {
-                    fprintf(stderr, "Could not negotiate a TLS connection: %s\n", err);
-                }
-                assert(err == NULL);
+            redisContext *ctx = getRedisContext(config.conn_info.hostip, config.conn_info.hostport, NULL);
```
madolson, line 2023: This also won't work in cluster mode right?

madolson (approved):
Looks right, you'll need to force push to fix the DSO though

PR #1715, opened 2025-02-11
Prevent double writing out of responses

madolson:
Fix issue https://github.com/valkey-io/valkey/actions/runs/13268563078/job/37042222814.

We were writing out a second response on the error.


enjoy-binbin (approved):

rjd15372:
@madolson thanks for the fix

Issue #1647, opened 2025-01-30
[test-failure] Test timeout for RESP3 client redirection for tracked key


hpatro:
```
[TIMEOUT]: clients state report follows.
sock55dda5d70500 => (IN PROGRESS) RESP3 Client gets tracking-redir-broken push message after cached key changed when rediretion client is terminated
Killing still running Valkey server 7266
```

https://github.com/valkey-io/valkey/actions/runs/13042887120/job/36388342710

hpatro:
Only failure from the daily run of yesterday https://github.com/valkey-io/valkey/actions/runs/13042887120/attempts/1

hpatro:
@Nikhil-Manglore Could you please take a look at this?

Nikhil-Manglore:
Sure, looking into it

zuiderkwast:
Has anyone reproduced this locally?

@xbasel recommends running github actions locally using https://github.com/nektos/act. Not sure if it makes a difference...

Nikhil-Manglore:
I was not able to reproduce it locally, I will try to set up those github actions and try again.

zuiderkwast:
If you can reproduce it, you can try my PR [#1719](https://github.com/valkey-io/valkey/pull/1719) and see if it fixes it. I have a hunch it might fix it.

xbasel:
Managed to reproduce, still not sure how:
```
[ok]: Tracking only occurs for scripts when a command calls a read-only command (2 ms)




c[TIMEOUT]: clients state report follows.
sock6532e6236960 => (IN PROGRESS) RESP3 Client gets tracking-redir-broken push message after cached key changed when rediretion client is terminated
Killing still running Valkey server 12839

                   The End

Execution time of different units:

!!! WARNING The following tests failed:

*** [TIMEOUT]: clients state report follows.
Cleanup: may take some time... OK
➜  valkey git:(unstable) ✗
```

But I did overload the CPU while the test was executing.

rjd15372:
@madolson opened a PR to fix this issue https://github.com/valkey-io/valkey/pull/1823

PR #1719, opened 2025-02-12
Don't leak TCL connections in unit/tracking

zuiderkwast:
While investigating #1647, I realized we repeatedly start new test clients without closing the old ones. This is leaking file descriptors in TCL, IIUC.

This patch does `close` before opening a new client. Without this patch, running the test locally 1000 times, with 100 parallel test workers ...

    ./runtest --clients 100 --single unit/tracking --loops 200

... it fails with what looks like running out of local ports for sockets:

    [exception]: Executing test client: couldn't open socket: address already in use. 
    ...
    Port 22488 was already busy, trying another port...
    Port 24490 was already busy, trying another port...

With this fix, I can successfully run with `--clients 100` and even increase it to `--clients 1000` and `--loops 2000`. With higher numbers I still get the "address already in use" error.

I don't know if this is related to the timeout in #1647. Maybe the timeout is related to running out of sockets?

### DIFF
--- a/tests/unit/tracking.tcl
+++ b/tests/unit/tracking.tcl
@@ -23,7 +23,9 @@ start_server {tags {"tracking network logreqres:skip"}} {
             # info which will not be consumed.
             r CLIENT TRACKING off
             $rd QUIT
+            $rd close
             $rd_redirection QUIT
+            $rd_redirection close
             set rd [valkey_deferring_client]
             set rd_redirection [valkey_deferring_client]
             $rd_redirection client id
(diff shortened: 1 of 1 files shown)


xbasel (approved):

xbasel (approved):
lgtm

hpatro (approved):
Nice fix. I presume there might be multiple tests in the repo where we don't close the tcl connections.

sarthakaggarwal97 (approved):

PR #1721, opened 2025-02-13
Fix temp file leak druing replication error handling

enjoy-binbin:
Before actually entering REPL_STATE_TRANSFER, we usually have
some other things to do, such as registering the ae handler, etc.
If an error occurs at this time, we may leak the previously opened
temp file.

This commit adds a new cleanupTransferResources function to do the
cleanup, avoiding code duplication.

### DIFF
--- a/src/replication.c
+++ b/src/replication.c
@@ -3887,7 +3887,7 @@ void cleanupTransferResources(void) {
         return;
     }
 
-    serverAssert(server.repl_transfer_fd != -1 && server.repl_transfer_tmpfile != NULL);
+    serverAssert(server.repl_transfer_tmpfile != NULL);
     close(server.repl_transfer_fd);
     bg_unlink(server.repl_transfer_tmpfile);
     zfree(server.repl_transfer_tmpfile);
--- a/src/replication.c
+++ b/src/replication.c
@@ -3842,6 +3842,8 @@ no_response_error: /* Handle receiveSynchronousResponse() error when primary has
     /* Fall through to regular error handling */
 
 error:
+    connClose(conn);
+    server.repl_transfer_s = NULL;
     if (server.repl_rdb_transfer_s) {
         connClose(server.repl_rdb_transfer_s);
         server.repl_rdb_transfer_s = NULL;
@@ -3898,6 +3900,8 @@ void cleanupTransferResources(void) {
  * Never call this function directly, use cancelReplicationHandshake() instead.
  */
 void undoConnectWithPrimary(void) {
+    if (server.repl_transfer_s == NULL) return;
+
     connClose(server.repl_transfer_s);
     server.repl_transfer_s = NULL;
 }
--- a/src/replication.c
+++ b/src/replication.c
@@ -47,6 +47,7 @@
 #include <ctype.h>
 
 void cleanupTransferResources(void);
+void replicationAbortSyncTransfer(void);
 void replicationDiscardCachedPrimary(void);
 void replicationResurrectCachedPrimary(connection *conn);
 void replicationResurrectProvisionalPrimary(void);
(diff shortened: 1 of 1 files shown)


src/replication.c
```diff
@@ -2871,8 +2867,7 @@ static void dualChannelFullSyncWithPrimary(connection *conn) {
         connClose(server.repl_rdb_transfer_s);
         server.repl_rdb_transfer_s = NULL;
     }
-    if (server.repl_transfer_fd != -1) close(server.repl_transfer_fd);
-    server.repl_transfer_fd = -1;
+    cleanupTransferResources();
     server.repl_state = REPL_STATE_CONNECT;
     replicationAbortDualChannelSyncTransfer();
```
enjoy-binbin, line 2867: replicationAbortDualChannelSyncTransfer also some cleanups, like replicationAbortDualChannelSyncTransfer will close repl_rdb_transfer_s and repl_transfer_fd. We can remove it in here. Please let me know if you want it removed.

src/replication.c
```diff
@@ -2871,8 +2867,7 @@ static void dualChannelFullSyncWithPrimary(connection *conn) {
         connClose(server.repl_rdb_transfer_s);
         server.repl_rdb_transfer_s = NULL;
```
enjoy-binbin: the leak may happen in here, we should also unlink the tmp file

src/replication.c
```diff
@@ -3856,10 +3852,7 @@ void syncWithPrimary(connection *conn) {
...
     }
-    if (server.repl_transfer_fd != -1) close(server.repl_transfer_fd);
-    if (server.repl_transfer_tmpfile) zfree(server.repl_transfer_tmpfile);
```
enjoy-binbin, line 3860: the leak may happen in here, we should also unlink the tmp file.

src/replication.c
naglera: I think we should call `cleanupTransferResources` from `replicationAbortDualChannelSyncTransfer` to maintain consistency in our error handling. 

We should also consider modifying `replicationAbortSyncTransfer` to handle the cleanup of temporary RDB file resources even when the replication state is not `REPL_STATE_TRANSFER`.

src/replication.c
```diff
@@ -3850,16 +3842,11 @@ void syncWithPrimary(connection *conn) {
     /* Fall through to regular error handling */
 
 error:
-    connClose(conn);
-    server.repl_transfer_s = NULL;
```
enjoy-binbin: remove this since replicationAbortSyncTransfer will handle this as well

src/replication.c
```diff
@@ -2867,14 +2864,8 @@ static void dualChannelFullSyncWithPrimary(connection *conn) {
         connClose(server.repl_transfer_s);
         server.repl_transfer_s = NULL;
     }
-    if (server.repl_rdb_transfer_s) {
```
enjoy-binbin, line 2870: remove this since replicationAbortDualChannelSyncTransfer will handle it as well

src/replication.c
```diff
@@ -3886,11 +3875,33 @@ int connectWithPrimary(void) {
...
+ * file fd and deleting the temp file.
+ *
+ * Noted that repl_transfer_fd and repl_transfer_tmpfile should be set/unset together. */
+void cleanupTransferResources(void) {
+    if (server.repl_transfer_fd == -1) {
+        serverAssert(server.repl_transfer_tmpfile == NULL);
+        return;
+    }
+
+    serverAssert(server.repl_transfer_fd != -1 && server.repl_transfer_tmpfile != NULL);
```
naglera: We already do the first part of the compression in the if statment above
```suggestion
    serverAssert(server.repl_transfer_tmpfile != NULL);
```

naglera (approved):

zuiderkwast (approved):

murphyjacob4:
@zuiderkwast, I see you marked it as "To be backported" in 7.2, but I don't think this needs backporting since it seems to be fixing issues introduced during dual channel (introduced in 8.0)

Let me know if this isn't the case

enjoy-binbin:
This also affect normal replication flow, but it is small and exist forever So i think we can skip the backport.

Fix temp file leak druing replication error handling
That is also the reason why the title did not mention dual channel

PR #1722, opened 2025-02-13
Fix raxRemove crash at memcpy() due to key size exceeds max Rax size

VoletiRam:
Fix raxRemove crash at memcpy() (line 1181#) due to key size exceeds `RAX_NODE_MAX_SIZE`. Note that this could happen when key size was more than 512MB if we allow it by increasing the default `proto-max-bulk-len`. The crash could happen when we recompress the rax after removing a key due to expiry or DEL while memcpy() merge the key that exceed 512MB limit. While the counting phase has the size check, the actual compress logic is missing it which lead to this crash.

Crash explanation with example:
Screenshot 2025-02-12 at 8 52 16 PM
Screenshot 2025-02-12 at 9 48 25 PM
`


zuiderkwast:
Great! Can you add a test case that would fail without this fix?

madolson:
You also need to fix the DCO.

VoletiRam:
Thank you for taking a look. @zuiderkwast @madolson 

1. Fixed the missing DCO.  
2. Added a regression unit test that reproduces the crash scenario:
   - The test is in test_rax.c.
   - It's disabled by default as it is a large memory test.

Please let me know if you need any adjustments to the change. Thank you.

src/unit/test_rax.c
```diff
@@ -1023,3 +1023,55 @@ int test_raxFuzz(int argc, char **argv, int flags) {
...
 }
+
+/* Test for crash in raxRemove when two keys share a common prefix in the same rax
+ * structure. Upon deletion of one key, rax attempts to recompress the structure
+ * back to its original form for other key. The crash occurs when the
+ * recompression exceeds 512MB size limit due to a missing size check in the
+ * compression logic.
+ *
+ * This test is disabled by default because it uses a lot of memory. */
+int test_raxRemoveCrash(int argc, char **argv, int flags) {
```
zuiderkwast: "Remove crash" is not a very descriptive title. :D

The comment above says things like "The crash occurs" but it should instead describe that the test checks that a crash doesn't occur. It can say that in the past, there was a crash.

```suggestion
int test_raxRecompressHugeKey(int argc, char **argv, int flags) {
```

zuiderkwast (commented):
Thanks for the test!

I have just a minor comment about the name and comment for the test case.

I have tried this test case without the fix and it indeed crashes. This is the stacktrace when running in GDB (with optimization so we can't see the args; sorry for that):

```
Program received signal SIGSEGV, Segmentation fault.
0x0000fffff7c94e80 in __memcpy_generic () from /lib64/libc.so.6
(gdb) bt
#0  0x0000fffff7c94e80 in __memcpy_generic () from /lib64/libc.so.6
#1  0x00000000004bc0fc in raxRemove (rax=0xfffff200c000, s=<optimized out>, len=<optimized out>, old=<optimized out>) at unit/../rax.c:1181
#2  0x00000000004c3540 in test_raxRemoveCrash (argc=<optimized out>, argv=<optimized out>, flags=<optimized out>) at unit/test_rax.c:1071
#3  0x00000000004762b4 in runTestSuite (test=test@entry=0x771a18 <unitTestSuite+176>, argc=argc@entry=4, argv=argv@entry=0xffffffffee08, flags=flags@entry=6) at unit/test_main.c:25
#4  0x000000000045a4cc in main (argc=4, argv=0xffffffffee08) at unit/test_main.c:61
```

src/unit/test_rax.c
VoletiRam: Thank you for the quick response and the review. I have addressed your suggestions.

zuiderkwast (approved):

madolson (approved):

enjoy-binbin (approved):

zuiderkwast:
The PR title isn't very good to use in the release notes. Users wouldn't understand what scenario you could get this crash. We could write something that doesn't require understanding of the internals.

Any ideas?

AFAIU, you won't get this in streams because streams only use short rax keys. Rax is used in ACL too but that too is unlikely to use very long keys. Client tracking with very long keys > 512M? Possibly. Module API "Dict" uses rax. Here I guess it'd be possible to reproduce the crash.

PR #1723, opened 2025-02-13
Introduce valkey_version_full info field which includes the release stage

ranshid:
This will introduce a new info field indicating the Valkey version and the release status.
The release status is used in order to provide release metadata information about the current version
release state. 
- In unstable branch the status is always "dev". 
- during release process the status will be set to rc1,rc2...rcN. 
- When the version is released the status will be "ga".

The reflected full version will be introduced to the info command output in valkey_version_full info field.
example:

in unstable:
```
# Server
redis_version:7.2.4
server_name:valkey
valkey_version:255.255.255
valkey_version_full:255.255.255-dev
redis_git_sha1:2f2b8d15
redis_git_dirty:1
redis_build_id:2f18af70ce63295b
```

during RC release:

```
# Server
redis_version:7.2.4
server_name:valkey
valkey_version:8.1.0
valkey_version_full:8.1.0-rc1
redis_git_sha1:2f2b8d15
redis_git_dirty:1
redis_build_id:2f18af70ce63295b
```

during Major/Minor/Patch release:

```
# Server
redis_version:7.2.4
server_name:valkey
valkey_version:8.1.3
valkey_version_full:8.1.3-ga
redis_git_sha1:2f2b8d15
redis_git_dirty:1
redis_build_id:2f18af70ce63295b
```

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -5668,6 +5668,7 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "redis_version:%s\r\n", REDIS_VERSION,
                 "server_name:%s\r\n", SERVER_NAME,
                 "valkey_version:%s\r\n", VALKEY_VERSION,
+                "valkey_version_full:%s\r\n", VALKEY_VERSION "-" VALKEY_RELEASE_EXT,
                 "redis_git_sha1:%s\r\n", serverGitSHA1(),
                 "redis_git_dirty:%i\r\n", strtol(serverGitDirty(), NULL, 10) > 0,
                 "redis_build_id:%s\r\n", serverBuildIdString(),
--- a/src/version.h
+++ b/src/version.h
@@ -6,6 +6,11 @@
 #define SERVER_TITLE "Valkey"
 #define VALKEY_VERSION "255.255.255"
 #define VALKEY_VERSION_NUM 0x00ffffff
+/* The release extension is used in order to provide release metadata information
+ * about the version release status. In unstable branch the status is always "dev".
+ * during release process the status will be set to rc1,rc2...rcN. When the version is released
+ * the status will be "ga". */
+#define VALKEY_RELEASE_EXT "dev"
 
 /* Redis OSS compatibility version, should never
  * exceed 7.2.x. */


enjoy-binbin (approved):

zuiderkwast:
It's a major decision to settle on the name and content of INFO fields. Let's do it as follow-up.

zuiderkwast:
What I don't like in this PR is the full version will be displayed as `8.1.0-ga`. This is not the name of the version. It's not semantic versioning and it's not the name we use, except in relation to release candidates. I think this field should match the git tag.

* The generally available version (the git tag) will be `8.1.0`, not `8.1.0-ga`. :x: 

* For release candidates, the git tag and full version is `8.1.0-rc1`. OK :heavy_check_mark:

* For the unstable branch, it will show as `255.255.255-dev`. I don't think it makes much sense, mostly because the numbers are already chosen to be a non-version number. OK :heavy_check_mark:

----

I have another idea: Use the output of `git describe --tags --always --dirty`.

* It shows the tag if we're on a tag (`8.1.0` or `8.1.0-rc1`)
* It shows the number of commits added on top of a tag, like now in the 8.0 branch it shows `8.0.2-1-g1a34a4ff7`, meaning 1 commit added after the tag 8.0.2.
* It adds the `-dirty` suffix if there are uncommitted changes.
* It shows just the commit id in unstable where there is not tag.

If we use this, then we can drop the `redis_git_sha` and `redis_git_dirty` fields in 9.0. I hope we can clean up because we have too many garbage fields. :)

ranshid:
@zuiderkwast Sorry - maybe I was rushing to merge it and should have waited for the discussion. 
We can and should definitely discuss then.

> What I don't like in this PR is the full version will be displayed as 8.1.0-ga. This is not the name of the version. It's not semantic versioning and it's not the name we use, except in relation to release candidates. I think this field should match the git tag.

regarding semantic versioning. I must say that the SemVer does allow setting metadata information following the version as described [here](https://semver.org/#spec-item-10):
```
Build metadata MAY be denoted by appending a plus sign and a series of dot separated identifiers immediately following the patch or pre-release version. Identifiers MUST comprise only ASCII alphanumerics and hyphens [0-9A-Za-z-]. Identifiers MUST NOT be empty. Build metadata MUST be ignored when determining version precedence. Thus two versions that differ only in the build metadata, have the same precedence. Examples: 1.0.0-alpha+001, 1.0.0+20130313144700, 1.0.0-beta+exp.sha.5114f85, 1.0.0+21AF26D3----117B344092BD.
``` 
maybe we can change this to use the '+' sign instead?

> I have another idea: Use the output of git describe --tags --always --dirty

This actually sound like a nice idea. maybe we can reflect the tag. but a question:
this is usually something done dynamically during the build (when we create the release.h)
so you suggest that for each build we would just select the latest tag on the branch we are on and reflect that right?

rjd15372:
I agree with @zuiderkwast idea of using `git describe --tags --always --dirty`. For using this approach the version field needs to be set during the build process.

ranshid:
@zuiderkwast I am fine with adding this, or at least place a PR to revert this and place your suggestion.

zuiderkwast:
@ranshid Intersting that semver allows + for metadata. So `+ga` suffix? OK.... Maybe.

> using git describe --tags --always --dirty. For using this approach the version field needs to be set during the build process.

@rjd15372 Yes, this is a problem for source releases. Just like the existing `redis_git_sha` we should have some fallback here if we decide to use git. I don't have a good answer to this.

I guess we need to think this from different angles a bit more. :)

rjd15372:
> @rjd15372 Yes, this is a problem for source releases. Just like the existing `redis_git_sha` we should have some fallback here if we decide to use git. I don't have a good answer to this.
> 
> I guess we need to think this from different angles a bit more. :)

The process for preparing the source release can include a step to write the correct version in `version.h` based on the git command.
How are source releases generated right now? Is it something that github does automatically?

zuiderkwast:
> The process for preparing the source release can include a step to write the correct version in `version.h` based on the git command.

Yes, maybe, or at least we could have some check that we don't have the completely wrong version in version.h

> How are source releases generated right now? Is it something that github does automatically?

Yes, or anyone can make one. It's just the content of the repo. I think it's good to keep it this simple.

PR #1724, opened 2025-02-13
Introduce Valkey release stage indication

ranshid:
In #1723 we introduced the release status as a full version info field (valkey_version_full)
However this does not follow the [SemVer version specifications](https://semver.org/).
We considered some alternatives:
1. using `git describe --tags --always --dirty` during the build process in order to place the current tag information.
   The problem with this laternative is that sometimes builds are done on untrackes sources (eg tar.gz downloads) and lack 
   the git info in order to correctly manage this data.
2. Include a [pre-release indication](https://semver.org/#spec-item-9) as part of the [VALKEY_VERSION](https://github.com/valkey-io/valkey/blob/unstable/src/version.h#L7) string - however this might break some clients testing with RC versions and would require to support/chop the prerelease indicator on REPLCONF other places.

In this PR we introduce reporting the release stage as a dedicated info field.
The release status value will reflect the release stage:
 * In unstable branch the status is always "dev".
 * During release process the status will be set to rc1,rc2...rcN. 
 * When the version is released the status will be "ga".

### DIFF
--- a/src/version.h
+++ b/src/version.h
@@ -6,6 +6,11 @@
 #define SERVER_TITLE "Valkey"
 #define VALKEY_VERSION "255.255.255"
 #define VALKEY_VERSION_NUM 0x00ffffff
+/* The release extension is used in order to provide release metadata information
+ * about the version release status. In unstable branch the status is always "dev".
+ * during release process the status will be set to rc1,rc2...rcN. When the version is released
+ * the status will be "ga". */
+#define VALKEY_RELEASE_EXT "dev"
 
 /* Redis OSS compatibility version, should never
  * exceed 7.2.x. */
--- a/src/version.h
+++ b/src/version.h
@@ -8,7 +8,7 @@
 #define VALKEY_VERSION_NUM 0x00ffffff
 /* The release stage is used in order to provide release status information.
  * In unstable branch the status is always "dev".
- * During release process the status will be set to rc1,rc2...rcN. 
+ * During release process the status will be set to rc1,rc2...rcN.
  * When the version is released the status will be "ga". */
 #define VALKEY_RELEASE_STAGE "dev"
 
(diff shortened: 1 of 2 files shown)


zuiderkwast (commented):
I'm OK with this change.

@valkey-io/core-team The background is the current plan to use `valkey_version:8.1.0` identical in all release candidates and the GA release. We need some other field to see the difference between release candidates and GA.

Let's finalize about this ASAP, before we release RC1.

src/version.h
```diff
@@ -6,11 +6,11 @@
...
 #define VALKEY_VERSION "255.255.255"
 #define VALKEY_VERSION_NUM 0x00ffffff
-/* The release extension is used in order to provide release metadata information
- * about the version release status. In unstable branch the status is always "dev".
- * during release process the status will be set to rc1,rc2...rcN. When the version is released
- * the status will be "ga". */
-#define VALKEY_RELEASE_EXT "dev"
+/* The release stage is used in order to provide release status information.
+ * In unstable branch the status is always "dev".
+ * During release process the status will be set to rc1,rc2...rcN.
```
enjoy-binbin:
```suggestion
 * During release process the status will be set to rc1,rc2...rcN.
```

enjoy-binbin (approved):
Ok with this.

src/version.h
enjoy-binbin: format ci

PingXie (approved):
Thanks, Ran!

madolson (approved):

madolson:
@zuiderkwast You happy with this? I see you didn't approve, otherwise it seems like we have enough folks to merge it.

zuiderkwast (approved):

enjoy-binbin:
@ranshid i belive we will need a doc PR to document it in the INFO

PR #1725, opened 2025-02-13
Respect process umask when creating data files

kronwerk:
Create data files (both AOF and RDB) using file mode as specified by the umask, which is inherited from the parent process, instead of hard-coded to 0644. The latter can be inconvenient in some cases where Valkey runs as a user and the files need to be accessed by a different user.

By passing file mode 0666 to `open()`, which is masked by the umask, the user can affect the mode of the created files, which will be `(0666 & ~umask)`. If the umask is 0022 (the default in most systems), the files will be created with mode 0644. If the user wants 0664, they can set umask to 0002 in the parent (e.g. in the shell) before starting Valkey.

### DIFF
--- a/src/aof.c
+++ b/src/aof.c
@@ -95,6 +95,11 @@ void aof_background_fsync_and_close(int fd);
 #define AOF_MANIFEST_KEY_FILE_SEQ "seq"
 #define AOF_MANIFEST_KEY_FILE_TYPE "type"
 
+/* Get AOF file mode from server. */
+static inline mode_t getAofFileMode(void) {
+    return ~server.umask & 0666;
+}
+
 /* Create an empty aofInfo. */
 aofInfo *aofInfoCreate(void) {
     return zcalloc(sizeof(aofInfo));
@@ -530,7 +535,7 @@ int writeAofManifestFile(sds buf) {
     sds tmp_am_name = getTempAofManifestFileName();
     sds tmp_am_filepath = makePath(server.aof_dirname, tmp_am_name);
 
-    int fd = open(tmp_am_filepath, O_WRONLY | O_TRUNC | O_CREAT, 0644);
+    int fd = open(tmp_am_filepath, O_WRONLY | O_TRUNC | O_CREAT, getAofFileMode());
     if (fd == -1) {
         serverLog(LL_WARNING, "Can't open the AOF manifest file %s: %s", tmp_am_name, strerror(errno));
 
@@ -728,7 +733,7 @@ void aofOpenIfNeededOnServerStart(void) {
 
     /* Here we should use 'O_APPEND' flag. */
     sds aof_filepath = makePath(server.aof_dirname, aof_name);
-    server.aof_fd = open(aof_filepath, O_WRONLY | O_APPEND | O_CREAT, 0644);
+    server.aof_fd = open(aof_filepath, O_WRONLY | O_APPEND | O_CREAT, getAofFileMode());
     sdsfree(aof_filepath);
     if (server.aof_fd == -1) {
         serverLog(LL_WARNING, "Can't open the append-only file %s: %s", aof_name, strerror(errno));
@@ -790,7 +795,7 @@ int openNewIncrAofForAppend(void) {
         new_aof_name = sdsdup(getNewIncrAofName(temp_am));
     }
     sds new_aof_filepath = makePath(server.aof_dirname, new_aof_name);
-    newfd = open(new_aof_filepath, O_WRONLY | O_TRUNC | O_CREAT, 0644);
+    newfd = open(new_aof_filepath, O_WRONLY | O_TRUNC | O_CREAT, getAofFileMode());
     sdsfree(new_aof_filepath);
     if (newfd == -1) {
         serverLog(LL_WARNING, "Can't open the append-only file %s: %s", new_aof_name, strerror(errno));
--- a/src/aof.c
+++ b/src/aof.c
@@ -95,11 +95,6 @@ void aof_background_fsync_and_close(int fd);
 #define AOF_MANIFEST_KEY_FILE_SEQ "seq"
 #define AOF_MANIFEST_KEY_FILE_TYPE "type"
 
-/* Get AOF file mode from server. */
-static inline mode_t getAofFileMode(void) {
-    return ~server.umask & 0666;
-}
-
 /* Create an empty aofInfo. */
 aofInfo *aofInfoCreate(void) {
     return zcalloc(sizeof(aofInfo));
@@ -535,7 +530,7 @@ int writeAofManifestFile(sds buf) {
     sds tmp_am_name = getTempAofManifestFileName();
     sds tmp_am_filepath = makePath(server.aof_dirname, tmp_am_name);
 
-    int fd = open(tmp_am_filepath, O_WRONLY | O_TRUNC | O_CREAT, getAofFileMode());
+    int fd = open(tmp_am_filepath, O_WRONLY | O_TRUNC | O_CREAT, getDataFileMode());
     if (fd == -1) {
         serverLog(LL_WARNING, "Can't open the AOF manifest file %s: %s", tmp_am_name, strerror(errno));
 
@@ -733,7 +728,7 @@ void aofOpenIfNeededOnServerStart(void) {
 
     /* Here we should use 'O_APPEND' flag. */
     sds aof_filepath = makePath(server.aof_dirname, aof_name);
-    server.aof_fd = open(aof_filepath, O_WRONLY | O_APPEND | O_CREAT, getAofFileMode());
+    server.aof_fd = open(aof_filepath, O_WRONLY | O_APPEND | O_CREAT, getDataFileMode());
     sdsfree(aof_filepath);
     if (server.aof_fd == -1) {
         serverLog(LL_WARNING, "Can't open the append-only file %s: %s", aof_name, strerror(errno));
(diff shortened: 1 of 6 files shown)


rjd15372:
I don't have much experience in administrating valkey deployments, but maybe using the server binary mask isn't also what we're expecting.
If 644 is not flexible, then maybe we need a config option to override the default file permissions for both the rdb and aof files.

kronwerk:
> I don't have much experience in administrating valkey deployments, but maybe using the server binary mask isn't also what we're expecting. If 644 is not flexible, then maybe we need a config option to override the default file permissions for both the rdb and aof files.

yeah, that was another option - but it appears to be an overkill to create 1 (or may be even 2, for AOF and RDB separately) another config setting to achieve this flexibility. I (maybe I'm wrong?) can't find a situation there we gonna need different permissions for AOF and RDB - and Valkey binary and AOF|RDB also

kronwerk:
@madolson hi, could you please take a look at this PR? thx

src/aof.c
```diff
@@ -530,7 +530,7 @@ int writeAofManifestFile(sds buf) {
     sds tmp_am_name = getTempAofManifestFileName();
     sds tmp_am_filepath = makePath(server.aof_dirname, tmp_am_name);
 
-    int fd = open(tmp_am_filepath, O_WRONLY | O_TRUNC | O_CREAT, 0644);
+    int fd = open(tmp_am_filepath, O_WRONLY | O_TRUNC | O_CREAT, getDataFileMode());
```
zuiderkwast: I believe we can just use 0666 directly. It gets masked down to the process mask anyway so the if umask is 0022, the file is created with 0644, 0002 => 0664, and so on.

zuiderkwast: I see you changed to this. Did you verify that this is actually how it works? I'm only 80% sure that this is how it works. :laughing:

kronwerk: planning to add test for this, hope this is not too hard in current test framework, heh

zuiderkwast: https://wiki.tcl-lang.org/page/umask

tests/integration/rdb.tcl
```diff
@@ -114,13 +114,27 @@ start_server [list overrides [list "dir" $server_path] keep_persistence true] {
     r del stream
 }
 
+set dump_path [file join $server_path dump.rdb]
+assert_equal [file attributes $dump_path -permissions] 00644
+
+package require Tclx
+umask 0002
```
zuiderkwast: Nice test case.

But we should restore it to the previous value, which is not necessarily 0022. Save the old umask here and restore that afterwards.

```suggestion
set old_umask [umask]
umask 0002
```

tests/integration/rdb.tcl
```diff
@@ -114,13 +114,27 @@ start_server [list overrides [list "dir" $server_path] keep_persistence true] {
...
+
+start_server [list overrides [list "dir" $server_path] keep_persistence true] {
+    test {Test nondefault umask applied} {
+        r save
+        assert_equal [file attributes $dump_path -permissions] 00664
+    }
+}
+umask 0022
```
zuiderkwast:
```suggestion
umask $old_umask
```

kronwerk: agreed, changed test for any old umask, I guess

tests/integration/rdb.tcl
kronwerk: agreed, changed test for any old umask, I guess

kronwerk:
@valkey-io/core-team PTAL
I guess this may be useful for those who use persistent data for backups - and the change itself looks rather safe

kronwerk:
Let me explain this just a bit more in detail.

Someone could possibly need to use persistent data on disk outside of Valkey, but trying to avoid using valkey user - e.g. with a separate user for just backups only with it's separate rights and accesses. But with default UMask it may be not possible.

In this PR we make it possible - for example, this utility can do both RDB and AOF backup/restore: https://github.com/wal-g/wal-g

But without this patch it won't be possible to run it from another user because 644 default is too tight for AOF folder to interact with it during restore, for example - and you should workaround it in a rather strange ways to mitigate that.

And I'm sure, this is not the only case, it's just an example - and the fix, as I've mentioned earlier, looks rather safe itself.

@valkey-io/core-team PTAL

tests/integration/rdb.tcl
```diff
@@ -114,13 +114,37 @@ start_server [list overrides [list "dir" $server_path] keep_persistence true] {
     r del stream
 }
 
+package require Tclx
```
zuiderkwast, line 118: Let's add some comment as a heading for this umask test code block, since it's outside of a regular test block.

tests/integration/rdb.tcl
```diff
@@ -114,13 +114,37 @@ start_server [list overrides [list "dir" $server_path] keep_persistence true] {
...
+set old_umask [umask]
+set old_perm 666
```
zuiderkwast: Doing this calculation in one step looks cleaner?

```suggestion
set old_perm [expr {666 - $old_umask}]
```

tests/integration/rdb.tcl
```diff
@@ -114,13 +114,37 @@ start_server [list overrides [list "dir" $server_path] keep_persistence true] {
...
+set dump_path [file join $server_path dump.rdb]
+assert_equal [file attributes $dump_path -permissions] 00$old_perm
+
+if {$old_umask == 22} {
+    set new_umask 2
+} else {
+    set new_umask 22
+}
+set new_perm 666
+incr new_perm -$new_umask
```
zuiderkwast: Can be done in one step using `expr`.

tests/integration/rdb.tcl
```diff
@@ -114,13 +114,37 @@ start_server [list overrides [list "dir" $server_path] keep_persistence true] {
...
+
+umask $new_umask
+start_server [list overrides [list "dir" $server_path] keep_persistence true] {
+    test {Test nondefault umask applied} {
+        r save
+        assert_equal [file attributes $dump_path -permissions] 00$new_perm
```
zuiderkwast: I tested this command from the Tcl shell in Tcl 9.0. It has a slightly different string output with a lowercase "o".

```
$ tclsh
% file attributes dump.rdb -permissions
0o644
```

Apparently, this is some new way to write octal numbers. It's not string-equal to 0064 but it's numerically equal. `assert_equal` uses string equality though. To accept both variants we could change to numeric comparison. Something like...

```suggestion
        assert {[file attributes $dump_path -permissions] == 00$new_perm}
```

zuiderkwast (commented):
Implementation LGTM. Just minor comments on the test code.

tests/integration/rdb.tcl
zuiderkwast: ... or assert_range, which uses >= and <=

```suggestion
        # Use numeric comparison for compatibility with Tcl 8 and 9.
        assert_range [file attributes $dump_path -permissions] 00$new_perm 00$new_perm
```

kronwerk:
@zuiderkwast thx for review - could u pls take a look at fixes?

zuiderkwast (approved):
@valkey-io/core-team FYI, this allows users to control the file mode of RDB and AOF files created by Valkey. It keeps existing behavior by default in systems where the sysadmin didn't change the default umask. (If they did change it, it's typically set to 0002, then it normally means they want new files to be created with mode 0664, so with this PR we respect the sysadmin's configuration.) Please comment if you have any concerns.

enjoy-binbin (approved):

zuiderkwast:
Grepping for 0644 in the source code, there are more places where we create files with 644: Log file, config file (on config rewrite), cluster config file, cli history file.

Should these be updated too?

```
$ git grep '0644' -- src/
src/acl.c:    if ((fd = open(tmpfilename, O_WRONLY | O_CREAT, 0644)) == -1) {
src/aof.c:    int fd = open(tmp_am_filepath, O_WRONLY | O_TRUNC | O_CREAT, 0644);
src/aof.c:    server.aof_fd = open(aof_filepath, O_WRONLY | O_APPEND | O_CREAT, 0644);
src/aof.c:    newfd = open(new_aof_filepath, O_WRONLY | O_TRUNC | O_CREAT, 0644);
src/cluster_legacy.c:    if ((fd = open(tmpfilename, O_WRONLY | O_CREAT, 0644)) == -1) {
src/cluster_legacy.c:    int fd = open(filename, O_WRONLY | O_CREAT | O_CLOEXEC, 0644);
src/config.c:    else if (fchmod(fd, 0644 & ~server.umask) == -1)
src/debug.c:    int fd = log_to_stdout ? STDOUT_FILENO : open(server.logfile, O_APPEND | O_CREAT | O_WRONLY, 0644);
src/replication.c:            dfd = open(tmpfile, O_CREAT | O_WRONLY | O_EXCL, 0644);
src/server.c:    fd = log_to_stdout ? STDOUT_FILENO : open(server.logfile, O_APPEND | O_CREAT | O_WRONLY, 0644);
src/unit/test_util.c:    int fd = open(tmpfile, O_RDWR | O_CREAT, 0644);
src/valkey-cli.c:        fd = open(filename, O_CREAT | O_WRONLY, 0644);
```

zuiderkwast:
@kronwerk This PR broke some of the Daily jobs. Tclx is not installed in these jobs.

Example: https://github.com/valkey-io/valkey/actions/runs/15150454620/job/42595495697#step:8:4750

Can you look into it?

Fusl:
> @kronwerk This PR broke some of the Daily jobs. Tclx is not installed in these jobs.

This unfortunately also broke my development workflow. `tclx` is not available as an official package on Arch Linux, only as an AUR package and I can't use those.

zuiderkwast:
One test suite already uses Tclx since before but it seems to be optional: `tests/integration/corrupt-dump-fuzzer.tcl`. If Tclx is not available, the test works anyway:

```tcl
# catch sigterm so that in case one of the random command hangs the test,
# usually due to the server not putting a response in the output buffers,
# we'll know which command it was
if { ! [ catch {
    package require Tclx
} err ] } {
    signal error SIGTERM
}
```

We could skip these umask tests if Tclx is not available. Or put some tag on it and only run it if some flag is passed to the runtest script.

zuiderkwast:
Here is a PR to make Tclx optional and skip the test if it's not installed:

* https://github.com/valkey-io/valkey/pull/2123

Issue #41, opened 2024-03-26
The output of INFO server


madolson:
The command `INFO` includes the following lines:

```
# Server
redis_version:255.255.255
redis_git_sha1:daadbca2
redis_git_dirty:1
redis_build_id:187c43196ce2ead2
redis_mode:standalone
```

Clients use this information to find out which version the server is running and from this they can conclude which features it supports.

The question is whether API compatibility justifies using a trademark name in the output of the command.

We have some alternatives:

1. Change "redis_version" to "server_version" (etc.) and also add a "server_name:newname"
2. Change the prefix, e.g. "newname_version", etc.
3. Keep the existing output.

romange:
Consider checking with Sidekiq - https://github.com/search?q=repo%3Asidekiq%2Fsidekiq%20redis_version&type=code

The maintainer of the framework may even agree to change the code to adapt to both standards but it will still be a problem for older versions of the library.

romange:
Also https://github.com/StackExchange/StackExchange.Redis/blob/2f69707d87488febb7376739ed4dc51ad9a422c6/src/StackExchange.Redis/ResultProcessor.cs#L864

zuiderkwast:
@romange Thx. We'll keep the `redis_version` field and it will indicate the last Redis version we're compatible with.

In addition to that, we'll add `server_name:valkey` and `server_version:x.y.z` for the Valkey version. I don't think any clients break because of adding more fields, but it's theoretically possible so we might even wait until our next major release before adding them.

yangbodong22011:
As far as I know, `redis_mode` is used in many places:

1. StackExchange.Redis: Determine the startup mode through `redis_mode` as standalone, cluster or sentinel. https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/ResultProcessor.cs#L872

2. Prometheus Exporter, use `redis_mode` to display instance_info. https://github.com/oliver006/redis_exporter/blob/master/exporter/info.go#L168

I think we should start a separate thread to let sdk/tool developers report their commands or fields that depend on Redis. (If possible, the best approach is to not change these old fields)

zuiderkwast:
I'm been thinking a bit about these new "server_version" and "server_name" fields in INFO. Ideally, services and forks based on valkey should be able to show their own names, e.g. "server_name:elastcache" or "server_name:foo" and then it may be useful to say which valkey version and which redis version it's compatible with.

[...]

PR #1736, opened 2025-02-14
Change redis to server in INFO fields redis_{git_sha1,git_dirty,build_id}

zuiderkwast:
These fields were mentioned in #41 but we didn't change them. I don't remember why.

I hope we can changing them in 9.0 when we're phasing out the redis compatibility fields.


src/server.c
```diff
@@ -5669,9 +5669,9 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "server_name:%s\r\n", SERVER_NAME,
                 "valkey_version:%s\r\n", VALKEY_VERSION,
                 "valkey_release_stage:%s\r\n", VALKEY_RELEASE_STAGE,
-                "redis_git_sha1:%s\r\n", serverGitSHA1(),
-                "redis_git_dirty:%i\r\n", strtol(serverGitDirty(), NULL, 10) > 0,
-                "redis_build_id:%s\r\n", serverBuildIdString(),
+                "server_git_sha1:%s\r\n", serverGitSHA1(),
```
madolson, line 5672: Some part of me says let's just leave them. One more thing to not list as a breaking change.

hwware: I remember we should keep this info message as is, otherwise is a break change. For the other update in this PR, everything is fine.

zuiderkwast: Yes, it's a breaking change.

So I think we will not change it in 9.0 then. But I'm wondering: How long are we going to keep everything 'redis'? Forever? Five years? If we keep it forever, then users will always remember that Redis was the original and Valkey is just a fork.

If we want to accept the risk, we are allowed to make breaking changes in a major version. I understand "redis_version" is used by many clients, but the git fields are likely not used by many. If we decide to change it some time, I hope users will understand that we want to phase out 'redis' named fields.

PR #1737, opened 2025-02-14
Fix error "SSL routines::bad length" when connTLSWrite is called second time with smaller buffer

### LINKED ISSUE #1135: [BUG] 0A00010F:SSL routines::bad length causing high latency and reconnects
**Describe the bug**

When running a high throughput Redis cluster (3 shards, 1 replica per shard) the read replicas start to hit a strange issue where we see very frequent reconnects and an inability to maintain a stable connection. Our cluster is set up with TLS (self-signed cert) and after enabling `verbose` logging I correlated these log lines to the reconnects that we're seeing in our application:

```
1:S 08 Oct 2024 00:41:39.521 - Error writing to client: error:0A00010F:SSL routines::bad length
```

**To reproduce**

I have attached two simple Kotlin apps that can reproduce this behavior, even against a local 3 shard cluster. This does not appear to be an io-threads issue because I set `io-threads: 1` to rule that out as being part of the cause.

[valkey-push.tar.gz](https://github.com/user-attachments/files/17285929/valkey-push.tar.gz)
[valkey-range.tar.gz](https://github.com/user-attachments/files/17285930/valkey-range.tar.gz)

**Expected behavior**

When running the `LPUSH` on the primary nodes and the `LRANGE + UNLINK` on the read replica it should be able to handle the throughput without frequent reconnects and `bad length` errors in the `verbose` logs.

**Additional information**

I spent some time adding additional logging, trying to figure out the source of the bug and so far I haven't found the conclusive cause but I do have some hints. The error above happens when `SSL_write` returns back error code `SSL_ERROR_WANT_WRITE` but the next write that occurs contains a different length than it expected (since `SSL_MODE_ACCEPT_MOVING_WRITE_BUFFER` appears to be enabled it doesn't need to be the exact same buffer).

What I see happening locally is we have some data to write (say, `81856` bytes). In the `_writeToClient` method the replyList is not empty so we jump into `writevToClient()`. In this method the 81856 bytes get split up into 4 arrays and put into `iov` (with an `iovcount` of `4`. Each array has `20464` bytes of data. 

These arrays get passed into `connTLSWritev()` in `tls.c`. The total size of the array is larger than `NET_MAX_WRITES_PER_EVENT` (64k) so it's going to invoke `connTLSWrite` one time for each array. The first call happens and it tries to write 20464 bytes of data but it's actually only able to write 16k (`16384`) bytes. We hit the following statement and break out of the loop:

```
if ((size_t)sent != iov[i].iov_len) break;
```

That's going to send us back to `writevToClient()` in `networking.c`. We increment the `totwritten` and modify the underlying `iov` arrays to move past the 16k of data that we just wrote. After this, we call back into `connTLSWritev()` in `tls.c`. This time we have `81856 - 16384 = 65472` bytes to write. This is _just_ under the 64k limit so we copy the arrays into a single buffer and try to write it out. Again, the `SSL_write` only writes 16kb of data (I think this is the max allowed per write?) and we repeat the process above until we've written all the data out.

This seems to work fine most of the time but I'm frequently seeing cases where it appears that the `c->reply` list on the connection has a different length between the time that we write out the first `16kb` of data and the time that we re-enter `writevToClient()` and attempt to send the un-sent data (e.g. - the exact payload that the `SSL_write` call is expecting to see). It's not clear to me whether or not it's expected that the `c->reply` list would change in the midst of writing a payload, but I assume it's just continually appended to? And perhaps the issue here is just that there's no handling of the case where an exact payload needs to be re-written?

I still have a pretty weak grasp on a lot of this code so I'm really hoping that someone who knows a lot more than me would be able to tell what's going on. I'm more than happy to try out any code changes locally to validate a fix.

zori-janea:
Issue described in #1135 

When call to `connTLSWrite` fails on next attempt `connTLSWritev` should provide buffer with at least same amount of bytes,
but if first call was with buffer smaller than `NET_MAX_WRITES_PER_EVENT` then buffer gets larger and exceed `NET_MAX_WRITES_PER_EVENT`, `connTLSWritev` will not combine `iov` into one buffer resulting into chance that first element of `iov` is smaller than last failed call to `connTLSWrite` and causing `SSL routines::bad length`.

This change force combining `iov`s into one buffer if first element of `iov` is smaller than last failed write by `connTLSWrite`

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -914,8 +914,9 @@ static int connTLSWrite(connection *conn_, const void *data, size_t data_len) {
 
     if (conn->c.state != CONN_STATE_CONNECTED) return -1;
     ERR_clear_error();
+    serverAssert(data_len >= conn->last_failed_write_data_len);
     ret = SSL_write(conn->ssl, data, data_len);
-    conn->last_failed_write_data_len = ret < 0 ? data_len : 0;
+    conn->last_failed_write_data_len = ret <= 0 ? data_len : 0;
     return updateStateAfterSSLIO(conn, ret, 1);
 }
 
@@ -950,7 +951,7 @@ static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt)
      * and send it away by one call to connTLSWrite(). */
     char buf[iov_bytes_len];
     size_t offset = 0;
-    for (int i = 0; i < iovcnt; i++) {
+    for (int i = 0; i < iovcnt && offset < iov_bytes_len; i++) {
         memcpy(buf + offset, iov[i].iov_base, iov[i].iov_len);
         offset += iov[i].iov_len;
     }
--- a/src/tls.c
+++ b/src/tls.c
@@ -452,6 +452,9 @@ typedef struct tls_connection {
     SSL *ssl;
     char *ssl_error;
     listNode *pending_list_node;
+    /* Per https://docs.openssl.org/master/man3/SSL_write, after a write call with partially written data,
+     * we must make subsequent write calls with the same length. We use this field to keep track of 
+     * the previous write length. */
     size_t last_failed_write_data_len;
 } tls_connection;
 
--- a/src/tls.c
+++ b/src/tls.c
@@ -453,7 +453,7 @@ typedef struct tls_connection {
     char *ssl_error;
     listNode *pending_list_node;
     /* Per https://docs.openssl.org/master/man3/SSL_write, after a write call with partially written data,
-     * we must make subsequent write calls with the same length. We use this field to keep track of 
+     * we must make subsequent write calls with the same length. We use this field to keep track of
      * the previous write length. */
     size_t last_failed_write_data_len;
 } tls_connection;
--- a/src/tls.c
+++ b/src/tls.c
@@ -452,6 +452,7 @@ typedef struct tls_connection {
     SSL *ssl;
     char *ssl_error;
     listNode *pending_list_node;
+    size_t last_failed_write_data_len;
 } tls_connection;
 
 static connection *createTLSConnection(int client_side) {
@@ -911,10 +912,12 @@ static int connTLSWrite(connection *conn_, const void *data, size_t data_len) {
     if (conn->c.state != CONN_STATE_CONNECTED) return -1;
     ERR_clear_error();
     ret = SSL_write(conn->ssl, data, data_len);
+    conn->last_failed_write_data_len = ret < 0 ? data_len : 0;
     return updateStateAfterSSLIO(conn, ret, 1);
 }
 
 static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt) {
+    tls_connection *conn = (tls_connection *)conn_;
     if (iovcnt == 1) return connTLSWrite(conn_, iov[0].iov_base, iov[0].iov_len);
 
     /* Accumulate the amount of bytes of each buffer and check if it exceeds NET_MAX_WRITES_PER_EVENT. */
@@ -927,7 +930,7 @@ static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt)
     /* The amount of all buffers is greater than NET_MAX_WRITES_PER_EVENT,
      * which is not worth doing so much memory copying to reduce system calls,
      * therefore, invoke connTLSWrite() multiple times to avoid memory copies. */
-    if (iov_bytes_len > NET_MAX_WRITES_PER_EVENT) {
+    if (iov_bytes_len > NET_MAX_WRITES_PER_EVENT && iovcnt > 0 && iov[0].iov_len >= conn->last_failed_write_data_len) {
         ssize_t tot_sent = 0;
         for (int i = 0; i < iovcnt; i++) {
             ssize_t sent = connTLSWrite(conn_, iov[i].iov_base, iov[i].iov_len);
--- a/src/tls.c
+++ b/src/tls.c
@@ -917,10 +917,11 @@ static int connTLSWrite(connection *conn_, const void *data, size_t data_len) {
     /* In case when last write failed due to some internal reason, retry has to provide
      * at least the same amount of bytes (https://docs.openssl.org/master/man3/SSL_write).
      * If that condition is not met, OpenSSL will return "SSL routines::bad length".
-     * When this assertion gets hit, caller has to be fixed to provide required bytes. */
+     * Currently we only suspect this can happen during primary cron sending '\n' 
+     * indication to the replica, so we silently return from this function without 
+     * impacting the connection state. */
     if (data_len < conn->last_failed_write_data_len) {
-        /* Keep assertion when debug is enabled, otherwise just return -1 */
-        debugServerAssert(0);
+        // TODO: place debugAssert for this case once the known issue described is resolved
         return -1;
     }
     ret = SSL_write(conn->ssl, data, data_len);
--- a/src/tls.c
+++ b/src/tls.c
@@ -918,7 +918,11 @@ static int connTLSWrite(connection *conn_, const void *data, size_t data_len) {
      * at least the same amount of bytes (https://docs.openssl.org/master/man3/SSL_write).
      * If that condition is not met, OpenSSL will return "SSL routines::bad length".
      * When this assertion gets hit, caller has to be fixed to provide required bytes. */
-    debugServerAssert(data_len >= conn->last_failed_write_data_len);
+    if (data_len < conn->last_failed_write_data_len) {
+        /* Keep assertion when debug is enabled, otherwise just return -1 */
+        debugServerAssert(0);
+        return -1;
+    }
     ret = SSL_write(conn->ssl, data, data_len);
     conn->last_failed_write_data_len = ret <= 0 ? data_len : 0;
     return updateStateAfterSSLIO(conn, ret, 1);
--- a/src/tls.c
+++ b/src/tls.c
@@ -917,8 +917,8 @@ static int connTLSWrite(connection *conn_, const void *data, size_t data_len) {
     /* In case when last write failed due to some internal reason, retry has to provide
      * at least the same amount of bytes (https://docs.openssl.org/master/man3/SSL_write).
      * If that condition is not met, OpenSSL will return "SSL routines::bad length".
-     * Currently we only suspect this can happen during primary cron sending '\n' 
-     * indication to the replica, so we silently return from this function without 
+     * Currently we only suspect this can happen during primary cron sending '\n'
+     * indication to the replica, so we silently return from this function without
      * impacting the connection state. */
     if (data_len < conn->last_failed_write_data_len) {
         // TODO: place debugAssert for this case once the known issue described is resolved
--- a/src/tls.c
+++ b/src/tls.c
@@ -914,10 +914,11 @@ static int connTLSWrite(connection *conn_, const void *data, size_t data_len) {
 
     if (conn->c.state != CONN_STATE_CONNECTED) return -1;
     ERR_clear_error();
-    if (data_len < conn->last_failed_write_data_len) {
-        conn->c.last_errno = EBUSY;
-        return -1;
-    }
+    /* In case when last write failed due to some internal reason, retry has to provide
+     * at least the same amount of bytes (https://docs.openssl.org/master/man3/SSL_write).
+     * If that condition is not met, OpenSSL will return "SSL routines::bad length".
+     * When this assertion gets hit, caller has to be fixed to provide required bytes. */
+    debugServerAssert(data_len >= conn->last_failed_write_data_len);
     ret = SSL_write(conn->ssl, data, data_len);
     conn->last_failed_write_data_len = ret <= 0 ? data_len : 0;
     return updateStateAfterSSLIO(conn, ret, 1);
--- a/src/tls.c
+++ b/src/tls.c
@@ -914,7 +914,10 @@ static int connTLSWrite(connection *conn_, const void *data, size_t data_len) {
 
     if (conn->c.state != CONN_STATE_CONNECTED) return -1;
     ERR_clear_error();
-    serverAssert(data_len >= conn->last_failed_write_data_len);
+    if (data_len < conn->last_failed_write_data_len) {
+        conn->c.last_errno = EBUSY;
+        return -1;
+    }
     ret = SSL_write(conn->ssl, data, data_len);
     conn->last_failed_write_data_len = ret <= 0 ? data_len : 0;
     return updateStateAfterSSLIO(conn, ret, 1);
@@ -931,9 +934,14 @@ static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt)
         if (iov_bytes_len > NET_MAX_WRITES_PER_EVENT) break;
     }
 
-    /* The amount of all buffers is greater than NET_MAX_WRITES_PER_EVENT,
-     * which is not worth doing so much memory copying to reduce system calls,
-     * therefore, invoke connTLSWrite() multiple times to avoid memory copies. */
+    /* In case the amount of all buffers is greater than NET_MAX_WRITES_PER_EVENT,
+     * it might not worth doing so much memory copying to reduce system calls,
+     * therefore, invoke connTLSWrite() multiple times to avoid memory copies.
+     * However, in case when last write failed we still have to repeat sending last_failed_write_data_len
+     * bytes. Because of openssl implementation we cannot repeat sending writes with length smaller than
+     * the last failed write (https://docs.openssl.org/master/man3/SSL_write) so in case the first io buffer
+     * does not provide at least the same amount of bytes as previous failed write, we will have to fallback to
+     * memory copy to a static buffer before calling SSL_write. */
     if (iov_bytes_len > NET_MAX_WRITES_PER_EVENT && iovcnt > 0 && iov[0].iov_len >= conn->last_failed_write_data_len) {
         ssize_t tot_sent = 0;
         for (int i = 0; i < iovcnt; i++) {
@@ -948,7 +956,11 @@ static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt)
     /* The amount of all buffers is less than NET_MAX_WRITES_PER_EVENT,
      * which is worth doing more memory copies in exchange for fewer system calls,
      * so concatenate these scattered buffers into a contiguous piece of memory
-     * and send it away by one call to connTLSWrite(). */
+     * and send it away by one call to connTLSWrite().
+     * However, code can fallback here in case when last write failed and first
+     * element of io is buffer not big enough to provide required amount of bytes
+     * to retry, so iov_bytes_len may exceed NET_MAX_WRITES_PER_EVENT by the amount
+     * of remaining bytes from last taken io. */
     char buf[iov_bytes_len];
     size_t offset = 0;
     for (int i = 0; i < iovcnt && offset < iov_bytes_len; i++) {


src/tls.c
```diff
@@ -452,6 +452,7 @@ typedef struct tls_connection {
     SSL *ssl;
     char *ssl_error;
     listNode *pending_list_node;
+    size_t last_failed_write_data_len;
```
madolson, line 458:
```suggestion
    /* Per https://docs.openssl.org/master/man3/SSL_write, after a write call with partially written data,
     * we must make subsequent write calls with the same length. We use this field to keep track of 
     * the previous write length. */
    size_t last_failed_write_data_len;
```

src/tls.c
```diff
@@ -927,7 +933,7 @@ static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt)
     /* The amount of all buffers is greater than NET_MAX_WRITES_PER_EVENT,
      * which is not worth doing so much memory copying to reduce system calls,
      * therefore, invoke connTLSWrite() multiple times to avoid memory copies. */
-    if (iov_bytes_len > NET_MAX_WRITES_PER_EVENT) {
+    if (iov_bytes_len > NET_MAX_WRITES_PER_EVENT && iovcnt > 0 && iov[0].iov_len >= conn->last_failed_write_data_len) {
```
ranshid, line 951: Please correct me if I am wrong. in case we used the static buffer on the first time, falling back to use it in case the first buffer len is smaller than the conn->last_failed_write_data_len might lead to stack overflow on the following write to the the static buffer. I suggest that we also modify the write to the static buffer like this:

```
/* The amount of all buffers is less than NET_MAX_WRITES_PER_EVENT,
     * which is worth doing more memory copies in exchange for fewer system calls,
     * so concatenate these scattered buffers into a contiguous piece of memory
     * and send it away by one call to connTLSWrite(). */
    char buf[iov_bytes_len];
    size_t offset = 0;
    for (int i = 0; i < iovcnt && offset < iov_bytes_len; i++) {
        memcpy(buf + offset, iov[i].iov_base, iov[i].iov_len);
        offset += iov[i].iov_len;
    }
    return connTLSWrite(conn_, buf, iov_bytes_len);
```

zori-janea: You are right, I tried to check why it didn't failed when I run test locally many times, and using `unit/other` test to fail it. Then looking into `writevToClient` I got https://github.com/valkey-io/valkey/blob/4e0149a381ac5e030537712f046c235110168b60/src/networking.c#L2094 and the part `iov_bytes_len < NET_MAX_WRITES_PER_EVENT` prevent this error, as it provides only `iov`s to match `NET_MAX_WRITES_PER_EVENT` + additional bytes from last `iov` that are also included when `iov_bytes_len` is calulcated, but the issue may happen in `writeToReplica` as there is no such check (should be added?).
Anyway I think we shouldn't rely on parent call for such things.

ranshid (approved):

ranshid:
@madolson the fix LGTM. do you approve as well?

madolson:
Looking now.

madolson:
At the very least no, there are tests failing:
https://github.com/valkey-io/valkey/actions/runs/13368943644/job/37355772561?pr=1737#step:6:5752

src/tls.c
```diff
@@ -910,11 +914,14 @@ static int connTLSWrite(connection *conn_, const void *data, size_t data_len) {
 
     if (conn->c.state != CONN_STATE_CONNECTED) return -1;
     ERR_clear_error();
+    serverAssert(data_len >= conn->last_failed_write_data_len);
```
madolson: This is triggering in the test because of the periodic write pings that we do. Maybe instead of crashing we return EAGAIN here?

madolson: Link to the location, https://github.com/valkey-io/valkey/blob/f85c93301ca63f070164b307c6a7371a9ab5ac7b/src/replication.c#L1957. @ranshid and I also discussed a new connection function, "connBestEffortWrite" which could drop the state here.

ranshid: you are right I completely forgot about that one. let me take a look

zori-janea: I think it just expose another way to trigger this issue, when some data wasn't written correctly and retry it with something else that may cause invalid protocol data (is sending new line in the middle of the other command safe?).
Returning `EAGAIN` should fix it, but I think not sending this ping when any data is awaiting to be sent is more accurate.
I added assert to expose anything that may cause invalid write to openssl, maybe also change it for now to `debugServerAssert`?

Edit: about first point it may also cause issue in non tls if newline in the middle of write isn't ok

src/tls.c
```diff
@@ -927,7 +934,7 @@ static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt)
     /* The amount of all buffers is greater than NET_MAX_WRITES_PER_EVENT,
```
xbasel: Worth updating comments

zori-janea: Good point and it forced me to rethink this solution as adding `&& iovcnt > 0 && iov[0].iov_len >= conn->last_failed_write_data_len` completely changed how this function works also `The amount of all buffers is less than NET_MAX_WRITES_PER_EVENT` from the next comment became invalid and it may exceed `NET_MAX_WRITES_PER_EVENT` by additional bytes from last used `iov`.

So maybe let handle failed write separately. @madolson, @ranshid and @xbasel let me know what do you think.
```diff
diff --git a/src/tls.c b/src/tls.c
index a080a9e2b..5679e2d8a 100644
--- a/src/tls.c
+++ b/src/tls.c
@@ -924,6 +924,34 @@ static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt)
     tls_connection *conn = (tls_connection *)conn_;
     if (iovcnt == 1) return connTLSWrite(conn_, iov[0].iov_base, iov[0].iov_len);

+    /* Last write failed so handle it first, as write may fail multiple times in a row,
+     * just provide least amount of needed bytes. */
+    if (conn->last_failed_write_data_len > 0)
+    {
+        /* Amount of bytes in first iov element is enough to procced last failed write, so use only this one. */
+        if (iovcnt > 0 && iov[0].iov_len >= conn->last_failed_write_data_len)
+            return connTLSWrite(conn_, iov[0].iov_base, iov[0].iov_len);
+
+        /* First element of iov isn't big enough so copy least amount of needed bytes. */
+        size_t bytes_to_copy = conn->last_failed_write_data_len;
+        size_t offset = 0;
+        char buf[bytes_to_copy];
+
+        for (int i = 0; i < iovcnt && bytes_to_copy > 0; i++) {
+            if (iov[i].iov_len <= bytes_to_copy) {
+                memcpy(buf + offset, iov[i].iov_base, iov[i].iov_len);
+                offset += iov[i].iov_len;
+                bytes_to_copy -= iov[i].iov_len;
+            } else {
+                memcpy(buf + offset, iov[i].iov_base, bytes_to_copy);
+                offset += bytes_to_copy;
+                bytes_to_copy = 0;
+            }
+        }
+
+        return connTLSWrite(conn_, buf, offset);
+    }
+
     /* Accumulate the amount of bytes of each buffer and check if it exceeds NET_MAX_WRITES_PER_EVENT. */
     size_t iov_bytes_len = 0;
     for (int i = 0; i < iovcnt; i++) {
@@ -934,7 +962,7 @@ static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt)
     /* The amount of all buffers is greater than NET_MAX_WRITES_PER_EVENT,
      * which is not worth doing so much memory copying to reduce system calls,
      * therefore, invoke connTLSWrite() multiple times to avoid memory copies. */
-    if (iov_bytes_len > NET_MAX_WRITES_PER_EVENT && iovcnt > 0 && iov[0].iov_len >= conn->last_failed_write_data_len) {
+    if (iov_bytes_len > NET_MAX_WRITES_PER_EVENT) {
         ssize_t tot_sent = 0;
         for (int i = 0; i < iovcnt; i++) {
             ssize_t sent = connTLSWrite(conn_, iov[i].iov_base, iov[i].iov_len);
@@ -951,7 +979,7 @@ static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt)
      * and send it away by one call to connTLSWrite(). */
     char buf[iov_bytes_len];
     size_t offset = 0;
-    for (int i = 0; i < iovcnt && offset < iov_bytes_len; i++) {
+    for (int i = 0; i < iovcnt; i++) {
         memcpy(buf + offset, iov[i].iov_base, iov[i].iov_len);
         offset += iov[i].iov_len;
     }
```

xbasel: If the previous write failed, shouldn't we expect iov[0] to contain at least last_failed_write_data_len bytes?

zori-janea: We can't expect, let assume you have `iov` of 2 elements ok 10KB, so they will be combined into one buffer and that failed by some reason. On second try we got more bytes to send and now has for instance `iov` has 7 elements of 10KB what will exceed NET_MAX_WRITES_PER_EVENT, so it will try call `connTLSWrite` on each `iov` separately and in this case first `iov` is only 10KB instead of 20KB from previous call

xbasel: You're right. On second thought, we can't and shouldn't assume how the data is structured.

Unrelated -- if `last_failed_write_data_len` is positive, you invoke `connTLSWrite` and don't proceed to the remaining buffers? (So it results in a partial write?)

xbasel: (And generally, I think the new approach is clearer.)

zori-janea: Yes this code may result into partial write and I don't think it's worth adding here logic of reducing buffer and calling `connTLSWrite` again as this logic already inside parent calls, for instance: https://github.com/valkey-io/valkey/blob/4e0149a381ac5e030537712f046c235110168b60/src/networking.c#L2040-L2052

I'm open for suggestion and ideas

src/tls.c
ranshid: >@ranshid and I also discussed a new connection function, "connBestEffortWrite" which could drop the state here.

@madolson I am not sure how we can solve it like this.
For example it is possible that we consumed all the write buffer. in this case it is possible that a followup ping send would be O.K to go (the connection has no last_failed_write_data_len > 0) however it MIGHT be blocked on the ping (return SSL_ERROR_WANT_WRITE). in this case even if we ignore the fact that the ping failed, the next send data we will try to send will miss the first bytes (since the ping is cached inside the openssl). I tested this in a test program and indeed it showed that this would be the case.

src/tls.c
ranshid: >We can't expect, let assume you have iov of 2 elements ok 10KB, so they will be combined into one buffer and that failed by some reason. On second try we got more bytes to send and now has for instance iov has 7 elements of 10KB what will exceed NET_MAX_WRITES_PER_EVENT, so it will try call connTLSWrite on each iov separately and in this case first iov is only 10KB instead of 20KB from previous call

@zori-janea I don't follow. in your current solution (in the PR) the casex you describe would still attempt to fillup the buffer since the iov[0] is less than 20KB... so what is the problem in the current implementation you provided (aside from the issue with the pings)

zori-janea: Yes current solution will also cover it correctly, what I mean I also think this version is easier to understand. When @xbasel pointed about correcting comments I got that from last comment in this function `The amount of all buffers is less than NET_MAX_WRITES_PER_EVENT` is no longer true as when `iov_bytes_len > NET_MAX_WRITES_PER_EVENT` is true but `iov[0].iov_len >= conn->last_failed_write_data_len` is false, it means that `iov_bytes_len` will exceed `NET_MAX_WRITES_PER_EVENT` and the amount of bytes exceed depends how big can be single `iov`, for instance as I've tested one iov was around 20KB so, 3 of them is below NET_MAX_WRITES_PER_EVENT, but in case of failed it will take 4.

Both version are fine and working.

src/tls.c
zori-janea: Yes, you are right, but this seems to be an issue inside ping and probably ping should go via `client` not `connection` directly in https://github.com/valkey-io/valkey/blob/f85c93301ca63f070164b307c6a7371a9ab5ac7b/src/replication.c#L1957
but I don't know is it possible use `server.primary` inside ping method, if I didn't miss anything it should be `client` for that connection

ranshid: This code is on the replica side. I do not think this is the problem. more possible the issue observed in the tests is related to sending pings

ranshid: Actually I am more confused. forgot that for pings (real ping commands we send to the replica) we are feeding the replication backlog so that should not meet this assert. the case of sending OutOfBand keepalive (\n) we do it from the primary only in cases were we should not have anything "pending" on the socket/SSL to the replica since we are either post exchange or we are writing RDB to disk (in which case we should not be streaming to the replica even with RDB second channel). I am probably missing a case were this is not true, but I cannot think of which.

src/tls.c
```diff
@@ -944,7 +951,7 @@ static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt)
      * and send it away by one call to connTLSWrite(). */
     char buf[iov_bytes_len];
     size_t offset = 0;
-    for (int i = 0; i < iovcnt; i++) {
+    for (int i = 0; i < iovcnt && offset < iov_bytes_len; i++) {
```
sungming2, line 972: nit: This condition may be unnecessary since offset is expected to match `iov_bytes_len` in the final iteration of the loop.
and out of curiosity, is it possible that accumulated buffer could be less than `last_failed_write_data_len`
e.g., failed 60KB and 20KB in buffers

sungming2: and out of curiosity, is it possible that accumulated buffer size could be less than `last_failed_write_data_len` here?
e.g., 60KB failed, and 20KB in buffers

ranshid: >offset is expected to match iov_bytes_len in the final iteration of the loop.

@sungming2 why is that? if we take the suggestion in this PR (the change in line 937) we can now fallback here even if the accumulated size of all vector buffers is larger than iov_bytes_len since that is calculated up to the first sum which exceeds NET_MAX_WRITES_PER_EVENT

src/tls.c
ranshid: @zori-janea I actually think the original proposal was simpler and I would suggest to stick with it. I agree the comment should be fixed to reflect the additional logic you added.

src/tls.c
ranshid: @zori-janea I think we need to understand how these ping (if indeed this is the case) are getting mixed in here.
I suggest we do return with the EAGAIN now and investigate the potential issue with the '\n' later.

ranshid: Maybe EAGAIN is not the correct thing to do? I think we should return an error since returning EAGIN is usually done with updateSSLEvent. I think we can return an error like -1 and set the errno to EBUSY or something. In the good path the error should not be taken into any consideration IMO

zori-janea: @ranshid is it ok?
```suggestion
    if (data_len < conn->last_failed_write_data_len) {
        conn->c.last_errno = EBUSY;
        return -1;
    }
```

Whatever we put here, it won't be correct until we find what is trying to send some smaller buffer then previous call. We can also just remove this assert and merge only changes in `connTLSWritev`

ranshid: Maybe you are right. removing the assert would probably still leave the same error as we have today for the small writes. I would be fine with that

src/tls.c
```diff
@@ -927,7 +934,7 @@ static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt)
     /* The amount of all buffers is greater than NET_MAX_WRITES_PER_EVENT,
      * which is not worth doing so much memory copying to reduce system calls,
      * therefore, invoke connTLSWrite() multiple times to avoid memory copies. */
```
zori-janea: @xbasel @ranshid is it ok?
```suggestion
    /* The amount of all buffers is greater than NET_MAX_WRITES_PER_EVENT,
     * which is not worth doing so much memory copying to reduce system calls,
     * therefore, invoke connTLSWrite() multiple times to avoid memory copies,
     * but in case when last write failed, first invoke of connTLSWrite() has to
     * provide at least the same amount of bytes as previous failed write, so
     * first element of iov has to be greater or equal to last failed write,
     * if not code will fall back to create buffer that provide enough bytes to
     * a single invoke of connTLSWrite(). */
```

src/tls.c
ranshid: @zori-janea can you please just fix the comment as @xbasel suggested?

src/tls.c
```diff
@@ -944,7 +951,7 @@ static int connTLSWritev(connection *conn_, const struct iovec *iov, int iovcnt)
      * and send it away by one call to connTLSWrite(). */
```
zori-janea: @xbasel @ranshid is it ok? I'm not very happy of this but I don't have better idea, as this code now is handling 2 cases
1. combines a much bytes as possible in one syscall
2. provide enough buffer to retry last failed write
```suggestion
     * and send it away by one call to connTLSWrite().
     * IMPORTATNT: Code can fall back here in case when last write failed and first
     * element of iov is not big enough to provide required amount of bytes to retry,
     * so iov_bytes_len may exceed NET_MAX_WRITES_PER_EVENT by the amount
     * of remaining bytes from last taken iov. */
```

src/tls.c
zori-janea: @ranshid I added it as suggestion, as I'm not very sure how to phrase it correctly
take a look at https://github.com/valkey-io/valkey/pull/1737#discussion_r1961469388 and https://github.com/valkey-io/valkey/pull/1737#discussion_r1961469388

src/tls.c
ranshid: Maybe something like this:

```
/* In case the amount of all buffers is greater than NET_MAX_WRITES_PER_EVENT,
     * it might not worth doing so much memory copying to reduce system calls,
     * therefore, invoke connTLSWrite() multiple times to avoid memory copies.
     * However, in case when last write failed we still have to repeat sending last_failed_write_data_len
     * bytes. Because of openssl implementation we cannot repeat sending writes with length smaller than 
     * the last failed write (https://docs.openssl.org/master/man3/SSL_write) so in case the first io buffer
     * does not provide at least the same amount of bytes as previous failed write, we will have to fallback to 
     * memory copy to a static buffer before calling SSL_write  */
```

src/tls.c
ranshid: @zori-janea I rethought - we cannot just remove the assert since it would cause the connections state to be CONN_STATE_ERROR. lets just return the made-up error.

zori-janea: I pushed changes, take a look

src/tls.c
```diff
@@ -910,11 +914,17 @@ static int connTLSWrite(connection *conn_, const void *data, size_t data_len) {
 
     if (conn->c.state != CONN_STATE_CONNECTED) return -1;
     ERR_clear_error();
+    if (data_len < conn->last_failed_write_data_len) {
+        conn->c.last_errno = EBUSY;
+        return -1;
+    }
```
ranshid: lets place a comment explaining this logic and why we think it is O.K to do so.

madolson: @ranshid I think we should also try to re-introduce the debugServerAssert. We can do it at the calling sites though, https://github.com/valkey-io/valkey/blob/cf19deb61d8ea0ba01a61f57edc1e9440367d98b/src/networking.c#L2255 I suppose might be a good place to check. Not a blocker for me on this PR.

madolson (commented):
Soft approval from me. I agree with Ran about improving the docs and also made a suggestion but don't feel like it needs to be in this PR, we can add it later.

src/tls.c
ranshid: @madolson placing a debugAssert would still cause the tests to fail right?

ranshid: also I am not sure about your proposal. In the current suggestion the error is not modifying the connection state (which is indeed strange) I am starting to think that even modifying the conn->c.last_errno is not needed since we would just want to return -1.

zori-janea: I think we should have this
```suggestion
    /* In case when last write failed due to some internal reason, retry has to provide
     * at least the same amount of bytes (https://docs.openssl.org/master/man3/SSL_write).
     * If that condition is not met, OpenSSL will return "SSL routines::bad length".
     * When this assertion gets hit, caller has to be fixed to provide required bytes. */
    debugServerAssert(data_len >= conn->last_failed_write_data_len);
```

Default test config has `enable-debug-assert yes` and this should be enough during tests(do I miss anything?), but prevent Valkey crash on production, in worst case it should lead to previous behaviour `SSL routines::bad length`, that should mean to us that some other places in code are not providing the same buffer size (or greater).

Let me know is it looking good to you

src/tls.c
```diff
@@ -910,11 +914,18 @@ static int connTLSWrite(connection *conn_, const void *data, size_t data_len) {
 
     if (conn->c.state != CONN_STATE_CONNECTED) return -1;
     ERR_clear_error();
+    /* In case when last write failed due to some internal reason, retry has to provide
+     * at least the same amount of bytes (https://docs.openssl.org/master/man3/SSL_write).
+     * If that condition is not met, OpenSSL will return "SSL routines::bad length".
+     * When this assertion gets hit, caller has to be fixed to provide required bytes. */
+    debugServerAssert(data_len >= conn->last_failed_write_data_len);
```
ranshid: So this debug assert would still leave us with the current issue (causing connection to be disconnected) in production I think we should still return -1 after the debug assert

zori-janea: Let take one step back, my intention for this PR was to resolve issue `bad length` when it comes from `connTLSWritev` and trying to test via that path issue didn't happen again. Adding later this assertion (my mistake that I used `serverAssert` instead of `debugServerAssert`) was to be sure that everything providing enough data length and this exposed other path like ping via newline, but this is out of scope of this PR (at least not intended to change by me at the begging).

So I think it can be merged as is, as it resolves part of the issue that comes from `connTLSWritev` and `debugServerAssert` will not break production servers and help debugging it in the future.
If you think change to returning `-1` will make it even better (I didn't tested it) we can change this to:
```suggestion
    if (data_len < conn->last_failed_write_data_len) {
        /* Keep assertion when debug is enabled, otherwise just return -1 */
        debugServerAssert(data_len >= conn->last_failed_write_data_len);
        return -1;
    }
```
@ranshid if it looks good to you feel free to just apply it.
Note: I know `if` and `assert` conditions looks repeated(one inverted), but it's intended to show exact failed condition in logs.

ranshid:
```suggestion
    if (data_len < conn->last_failed_write_data_len) {
        /* Keep assertion when debug is enabled, otherwise just return -1 */
        debugServerAssert(0);
        return -1;
    }
```

ranshid: >Let take one step back, my intention for this PR was to resolve issue bad length when it comes from connTLSWritev and trying to test via that path issue didn't happen again. Adding later this assertion (my mistake that I used serverAssert instead of debugServerAssert) was to be sure that everything providing enough data length and this exposed other path like ping via newline, but this is out of scope of this PR (at least not intended to change by me at the begging).

@zori-janea I agree this PR was addressing an issue which was more urgent to solve. the other issue is something I think we can handle differently. given said that, we already know we have a potential issue which will still cause the connection to be closed if we do not return -1 (since the connection would go into error state) I think I am O.K with that and mainly waiting to see if @madolson would like to give a "hard" approval.

ranshid: >So I think it can be merged as is, as it resolves part of the issue that comes from connTLSWritev and debugServerAssert will not break production servers and help debugging it in the future.

I am +1 for merging it this way. we can probably find a way to make sure we flush the replication output buffer before we start the BGsave process.

src/tls.c
madolson: Yeah, we want this to be a debug assert since it will crash during tests BUT will just result in disconnects in production workloads. This looks right to me.

ranshid:
@zori-janea I will wait for the test result and merge it. would like to take the opportunity to thank you for tackling this issue and for the patience during the review!

src/tls.c
madolson: We can't have steady state test failures since people won't trust the tests. I'm okay with omitting the debug assertions, merging the code, then following up with a fix for both though. I agree, let's not keep cycling on this PR though.

src/tls.c
```diff
@@ -910,11 +914,22 @@ static int connTLSWrite(connection *conn_, const void *data, size_t data_len) {
...
     ERR_clear_error();
+    /* In case when last write failed due to some internal reason, retry has to provide
+     * at least the same amount of bytes (https://docs.openssl.org/master/man3/SSL_write).
+     * If that condition is not met, OpenSSL will return "SSL routines::bad length".
+     * When this assertion gets hit, caller has to be fixed to provide required bytes. */
+    if (data_len < conn->last_failed_write_data_len) {
+        /* Keep assertion when debug is enabled, otherwise just return -1 */
+        debugServerAssert(0);
+        return -1;
+    }
```
ranshid:
```suggestion
    /* In case when last write failed due to some internal reason, retry has to provide
     * at least the same amount of bytes (https://docs.openssl.org/master/man3/SSL_write).
     * If that condition is not met, OpenSSL will return "SSL routines::bad length".
     * Currently we only suspect this can happen during primary cron sending '\n' 
     * indication to the replica, so we silently return from this function without 
     * impacting the connection state. */
    if (data_len < conn->last_failed_write_data_len) {
        // TODO: place debugAssert for this case once the known issue described is resolved
        return -1;
    }
```

src/tls.c
```diff
@@ -910,11 +914,23 @@ static int connTLSWrite(connection *conn_, const void *data, size_t data_len) {
 
     if (conn->c.state != CONN_STATE_CONNECTED) return -1;
     ERR_clear_error();
+    /* In case when last write failed due to some internal reason, retry has to provide
+     * at least the same amount of bytes (https://docs.openssl.org/master/man3/SSL_write).
+     * If that condition is not met, OpenSSL will return "SSL routines::bad length".
+     * Currently we only suspect this can happen during primary cron sending '\n' 
+     * indication to the replica, so we silently return from this function without
```
madolson:
```suggestion
     * Currently we only suspect this can happen during primary cron sending '\n'
     * indication to the replica, so we silently return from this function without
```

zori-janea:
@ranshid I'm glad that we got to the point to move forward and thank you for reviewing it 🎉
I will try to look into (when possible) into the ping part issue, as doc says
`We send a single newline character that is valid protocol but is guaranteed to either be sent entirely or not, since the byte is indivisible.`
It's fine for non-TLS but as we got openssl has internal state so even a single byte isn't correct here
https://github.com/valkey-io/valkey/blob/f85c93301ca63f070164b307c6a7371a9ab5ac7b/src/replication.c#L1944-L1952

src/tls.c
murphyjacob4: Hello from the future! 👋 

So the check iov_bytes_len < NET_MAX_WRITES_PER_EVENT is evaluated before we decide to add the next block, not after:

Iteration 1: iov_bytes_len is 0 (<64KB). We add a small reply block (e.g., 10KB). iov_bytes_len becomes 10KB.
Iteration 2: iov_bytes_len is 10KB (<64KB). The loop continues. We add the next reply block, which happens to be a huge 10MB payload. iov_bytes_len becomes 10.01MB
Iteration 3: iov_bytes_len is 10.01MB (not <64KB). The loop terminates.

So it can still cause a stack overflow. I sent out #4307 to address the issue.

PR #1742, opened 2025-02-16
Explicitly use github arm runners for ARM release

ranshid:
In the last 8.1.0-rc1 release attempt we identified that the cross compilation on x86 for ARM is broken
(example: https://github.com/valkey-io/valkey/actions/runs/13324096504/job/37225952044)

This PR sets use of arm runners during release binaries.

### DIFF
--- a/.github/workflows/call-build-linux-arm-packages.yml
+++ b/.github/workflows/call-build-linux-arm-packages.yml
@@ -35,7 +35,7 @@ jobs:
   build-valkey:
     # Capture source tarball and generate checksum for it
     name: Build package ${{ matrix.distro.target }} ${{ matrix.distro.arch }}
-    runs-on: "ubuntu-latest"
+    runs-on: "ubuntu-24.04-arm"
     strategy:
       fail-fast: false
       matrix: ${{ fromJSON(inputs.build_matrix) }}


trim21:
there are some bug on GitHub side， currently jobs on ubuntu-24.04-arm may randomly failed with segment error

https://github.com/actions/partner-runner-images/issues/47
prefer ubuntu-22.04-arm maybe

enjoy-binbin (approved):

ranshid:
> there are some bug on GitHub side， currently jobs on ubuntu-24.04-arm may randomly failed with segment error
> 
> [actions/partner-runner-images#47](https://github.com/actions/partner-runner-images/issues/47) prefer ubuntu-22.04-arm maybe

Interesting. I would still prefer to be aligned with the rest of our daily-runs on ARM which are using ubuntu-24.04. In case we hit this bug we can issue another PR in order to temporary revert to 22.04.

Thank you @trim21 !

ranshid:
> there are some bug on GitHub side， currently jobs on ubuntu-24.04-arm may randomly failed with segment error
> 
> [actions/partner-runner-images#47](https://github.com/actions/partner-runner-images/issues/47) prefer ubuntu-22.04-arm maybe

Interesting. I would still prefer to be aligned with the rest of our daily-runs on ARM which are using ubuntu-24.04. In case we hit this bug we can issue another PR in order to temporary revert to 22.04.

Thank you @trim21 !

PR #1743, opened 2025-02-16
Support for RDB analysis reports

artikell:
The PR add analysis capabilities to the valkey-check-rdb tool. Now the tool can statistically analyze various types of keys, expired keys, and key elements.

- Added the --stats parameter to the valkey-check-rdb tool's rdb analysis functionality, which enables opening reports. 
- Supported the --format parameter with three options, table/csv/info, to provide more flexible output formats. 
- Add function & lua script number info

The content format is as follows：
```
... ...
[offset 8999] \o/ RDB looks OK! \o/
[info] 60 keys read
[info] 0 expires
[info] 0 already expired
db.0.type.name                  string  list    set     zset    hash    module  stream
db.0.keys.total                 10      0       0       0       0       0       0    
db.0.exipre_keys.total          0       0       0       0       0       0       0    
db.0.already_expired.total      0       0       0       0       0       0       0    
db.0.keys.size                  200     0       0       0       0       0       0    
db.0.keys.value_size            2000    0       0       0       0       0       0    
db.0.elements.total             10      0       0       0       0       0       0    
db.0.elements.size              2000    0       0       0       0       0       0    
db.0.elements.num.max           1       0       0       0       0       0       0    
db.0.elements.num.avg           1.00    0.00    0.00    0.00    0.00    0.00    0.00 
db.0.elements.num.p99           1.00    0.00    0.00    0.00    0.00    0.00    0.00 
db.0.elements.num.p90           1.00    0.00    0.00    0.00    0.00    0.00    0.00 
db.0.elements.num.p50           1.00    0.00    0.00    0.00    0.00    0.00    0.00 
db.0.elements.size.max          200     0       0       0       0       0       0    
db.0.elements.size.avg          200.00  0.00    0.00    0.00    0.00    0.00    0.00 
db.0.elements.size.p99          200.00  0.00    0.00    0.00    0.00    0.00    0.00 
db.0.elements.size.p90          200.00  0.00    0.00    0.00    0.00    0.00    0.00 
db.0.elements.size.p50          200.00  0.00    0.00    0.00    0.00    0.00    0.00
```

### DIFF
--- a/src/valkey-check-rdb.c
+++ b/src/valkey-check-rdb.c
@@ -31,6 +31,8 @@
 #include "server.h"
 #include "rdb.h"
 #include "module.h"
+#include "hdr_histogram.h"
+#include "fpconv_dtoa.h"
 
 #include <stdarg.h>
 #include <sys/time.h>
@@ -39,7 +41,33 @@
 
 void createSharedObjects(void);
 void rdbLoadProgressCallback(rio *r, const void *buf, size_t len);
+void computeDatasetProfile(int dbid, robj *keyobj, robj *o);
+
 int rdbCheckMode = 0;
+int rdbCheckProfiler = 0;
+
+#define LOW_TRACKE_VALUE 1
+#define MAX_ELEMENTS_TRACKE 200 * 1024
+#define MAX_ELEMENTS_SIZE_TRACKE 1024 * 1024
+
+typedef struct rdbProfiler {
+    size_t type;
+    unsigned long keys;
+    unsigned long expires;
+    unsigned long already_expired;
+
+    unsigned long all_key_size;
+    unsigned long all_value_size;
+
+    unsigned long elements;
+    unsigned long all_elements_size;
+
+    unsigned long elements_max;
+    unsigned long elements_size_max;
+
+    struct hdr_histogram *element_count_histogram;
+    struct hdr_histogram *element_size_histogram;
+} rdbProfiler;
 
 struct {
     rio *rio;
@@ -51,6 +79,12 @@ struct {
     int doing;                     /* The state while reading the RDB. */
     int error_set;                 /* True if error is populated. */
     char error[1024];
+    int databases;
+    int format;
+
+    /* profiler */
+    rdbProfiler **profiler; /* profiler group by datatype,encoding,isexpired */
+    int profiler_num;
 } rdbstate;
 
 /* At every loading step try to remember what we were about to do, so that
@@ -65,6 +99,9 @@ struct {
 #define RDB_CHECK_DOING_READ_AUX 7
 #define RDB_CHECK_DOING_READ_MODULE_AUX 8
 #define RDB_CHECK_DOING_READ_FUNCTIONS 9
+
+#define OUTPUT_FORMAT_INFO 0
+#define OUTPUT_FORMAT_FORM 1
 
 char *rdb_check_doing_string[] = {
     "start",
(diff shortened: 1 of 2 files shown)


zuiderkwast (commented):
Cool. Not sure about the argument name `--profiler`. Profiler is more often used for things like CPU time.

How about `--stats` or `--verbose`?

An RDB file can also contain Functions IIRC. Shall we print the count too?

@enjoy-binbin I know you wanted to print the eval scripts in old RDB files. Maybe we can print the number of scripts as part of stats here too?

artikell:
`--stats` seems more suitable for this scenario, let me add the data of the function.

enjoy-binbin:
> @enjoy-binbin I know you wanted to print the eval scripts in old RDB files. Maybe we can print the number of scripts as part of stats here too?

You have a wrong memory, i dont want to print the lua scripts, i want to print the number of scripts. See #1448

zuiderkwast:
> > @enjoy-binbin I know you wanted to print the eval scripts in old RDB files. Maybe we can print the number of scripts as part of stats here too?
> 
> You have a wrong memory, i dont want to print the lua scripts, i want to print the number of scripts. See #1448

Sorry, I mean number of scripts. This PR prints the number of many things, such as the of functions, number of keys of each type, etc. To print the number of lua scripts if any would make sense here? Maybe only print it if it's non-zero, because it's only for quite old RDB files. WDYT?

enjoy-binbin:
> This PR prints the number of many things, such as the of functions, number of keys of each type, etc. To print the number of lua scripts if any would make sense here? Maybe only print it if it's non-zero, because it's only for quite old RDB files. WDYT?

SGTM, we can print it (that is what i was looking for at that time). I can re-open the old PR if needed if you want to keep the history. Or we can do it in this PR.

zuiderkwast:
> Or we can do it in this PR.

I guess in this PR is better in that case. This PR prints the info in another way. Or maybe reopen it after this PR is merged.

@artikell do you want to add the number of lua scripts from old RDB files as in #1448? They are just aux-fields with the name "lua". I suggest print them only if the number is non-zero since it's old stuff that will always be zero except in very old RDB files.

zuiderkwast:
@artikell What is the "form" formats? Can you mention it in the PR top comment?

What's the use case of the "form" format? If it is to consume by scripts, I guess CSV or JSON formats is more common. We already use those formats in valkey-cli and valkey-benchmark.

artikell:
> @artikell What is the "form" formats? Can you mention it in the PR top comment?
> 
> What's the use case of the "form" format? If it is to consume by scripts, I guess CSV or JSON formats is more common. We already use those formats in valkey-cli and valkey-benchmark.

- The `form` is a format aligned by tabs.(better to call it "table"?)  I think it is more conducive to people's reading. Just like the format in the commit, the result is that it is aligned. In my opinion, CSV can also be supported.

- I have considered using JSON, but implementing the YAML format is a bit simpler (at least, all it takes is adding some spaces). 

Regarding the number of Lua scripts, I will refer to Binbin's MR to make the addition in this MR.

zuiderkwast:
> * The `form` is a format aligned by tabs.(better to call it "table"?)  I think it is more conducive to people's reading.

Yes maybe "table" is a better name. The example in the PR comment is the "info" format? It is also readable for users. Not a big difference to "form"? I'm thinking maybe we don't need both....

For CSV, JSON, YAML, I don't have a strong opinion but it seems nice to align the output format from our different tools. CSV seems to be supported by most of them so add it here too?

artikell:
> > * The `form` is a format aligned by tabs.(better to call it "table"?)  I think it is more conducive to people's reading.
> 
> Yes maybe "table" is a better name. The example in the PR comment is the "info" format? It is also readable for users. Not a big difference to "form"? I'm thinking maybe we don't need both....
> 
> For CSV, JSON, YAML, I don't have a strong opinion but it seems nice to align the output format from our different tools. CSV seems to be supported by most of them so add it here too?

@zuiderkwast I've understood your thoughts. Now I've changed it to three formats:
- Table: It is the default structure in the PR comment, which is conducive to human reading.
- CSV: A formatted output that is convenient for other scripts to parse.
- Info: Referring to the original format, it is used for the TCL script to make correctness judgments.

This is CSV output：
image

This is Info output：
image

src/valkey-check-rdb.c
```diff
@@ -374,6 +760,50 @@ int redis_check_rdb(char *rdbfilename, FILE *fp) {
...
+            i++;
+        } else {
+            goto checkRdbUsage;
+        }
+    }
+
+    return;
+
+checkRdbUsage:
+    fprintf(stderr, "Usage: %s <rdb-file-name> [--format form|info|yaml] [--stats]\n", argv[0]);
```
zuiderkwast: This doesn't match the implemented formats: table, info and csv.

zuiderkwast (commented):
The CSV output is prefixed with some non-CSV content like

```
[offset 8999] \o/ RDB looks OK! \o/
[info] 60 keys read
[info] 0 expires
[info] 0 already expired
```

.... so I guess it doesn't work to direct the output to file and load it as a CSV file. 

Can we skip the prefix information for the CSV output mode? An option could be to print the non-CSV prefix info to stderr and the CSV content to stdout (?).

artikell:
> The CSV output is prefixed with some non-CSV content like
> 
> ```
> [offset 8999] \o/ RDB looks OK! \o/
> [info] 60 keys read
> [info] 0 expires
> [info] 0 already expired
> ```
> 
> .... so I guess it doesn't work to direct the output to file and load it as a CSV file.
> 
> Can we skip the prefix information for the CSV output mode? An option could be to print the non-CSV prefix info to stderr and the CSV content to stdout (?).

I have added an `--output` parameter, which enables the output of statistical information to a file by modifying the target of `stdout`. It seems to meet the requirements.

src/valkey-check-rdb.c
```diff
@@ -374,6 +784,54 @@ int redis_check_rdb(char *rdbfilename, FILE *fp) {
...
+        lastarg = (i == (argc - 1));
+        if (!strcmp(argv[1], "-v") || !strcmp(argv[1], "--version")) {
+            sds version = getVersion();
+            printf("valkey-check-rdb %s\n", version);
+            sdsfree(version);
+            exit(0);
+        } else if (!strcmp(argv[i], "--stats")) {
+            rdbCheckStats = 1;
+        } else if (!strcmp(argv[i], "--output")) {
+            valkey_strlcpy(rdbstate.stats_output, argv[i + 1], strlen(argv[i + 1]));
```
zuiderkwast: Why not heap allocate it with zstrdup instead and avoid the size limitation?

src/valkey-check-rdb.c
```diff
@@ -374,6 +784,54 @@ int redis_check_rdb(char *rdbfilename, FILE *fp) {
...
+            i++;
+        } else {
+            goto checkRdbUsage;
+        }
+    }
+
+    return;
+
+checkRdbUsage:
+    fprintf(stderr, "Usage: %s <rdb-file-name> [--format table|info|csv] [--stats]\n", argv[0]);
```
zuiderkwast: Add the `--output FILE` argument too.

zuiderkwast (commented):
Yeah the `--output` argument will work with `--csv`. I hope users will find it, instead of using shell redirect or pipe `valkey-check-rdb --csv > file`.

src/valkey-check-rdb.c
artikell: Make sence. done.

src/valkey-check-rdb.c
```diff
@@ -280,7 +674,13 @@ int redis_check_rdb(char *rdbfilename, FILE *fp) {
                 decrRefCount(auxkey);
                 goto eoferr;
             }
-
+            if (!strcasecmp(auxkey->ptr, "lua")) {
+                /* In older version before 7.0, we may save lua scripts in a replication RDB,
+                 * although it is not an actually aux field, we will still print it in here since
+                 * it's easy to filter using external grep. Use a counter so that at the end we
+                 * can print its number, if any. */
```
zuiderkwast: Some parts of this comment is from #1448, not very relevant here?

```suggestion
                /* In older version before 7.0, we may save lua scripts in a replication RDB. */
```

zuiderkwast (approved):
LGTM, thanks and sorry for the delay.

Please update the PR top comment to mention the parameters that were actually implemented.

@enjoy-binbin or @madolson do you want to take a look? If not, I will merge it.

src/valkey-check-rdb.c
```diff
@@ -104,11 +147,357 @@ char *rdb_type_string[] = {
...
+
+    hdr_record_value(stats->element_size_histogram, (int64_t)eleSize);
+}
+
+void stats_record_simple(size_t eleSize, size_t eleCount, rdbStats *stats) {
+    stats_record_count(eleCount, stats);
+    stats_record_element_size(eleSize, eleCount, stats);
+}
+
+void stats_record_element_size_add(rdbStats *to, rdbStats *from) {
```
enjoy-binbin: I saw some function names using `_` and some are using camelCase.

roshkhatri:
This commit causes a build failure in alpine-linux
```
valkey-check-rdb.c: In function 'rdbShowGenericInfo':
valkey-check-rdb.c:498:20: error: assignment of read-only variable 'stdout'
  498 |             stdout = fdopen(1, "w");
      |                    ^
make[1]: *** [Makefile:546: valkey-check-rdb.o] Error 1
make[1]: *** Waiting for unfinished jobs....
make[1]: Leaving directory '/usr/src/valkey/src'
make: *** [Makefile:6: all] Error 2
make: Leaving directory '/usr/src/valkey'
```
https://github.com/valkey-io/valkey-container/actions/runs/15694960906/job/44219225483#step:5:739

PR #1745, opened 2025-02-17
Comment out assert in kvstore for overhead lut

### LINKED ISSUE #1657: [test-failure] Unit test failure in test_kvstore.c
https://github.com/valkey-io/valkey/actions/runs/13120508567/job/36605063835

```
[serverAssert - unit/../kvstore.c:329] - kvs->overhead_hashtable_lut == 0
```

zuiderkwast:
Until we found out why the assert is flaky, let's comment it out. We can live with this overhead lut being slightly wrong, but we don't want the assert to abort the program.

Fixes #1657.

### DIFF
--- a/src/kvstore.c
+++ b/src/kvstore.c
@@ -319,7 +319,8 @@ void kvstoreRelease(kvstore *kvs) {
         if (metadata->rehashing_node) metadata->rehashing_node = NULL;
         hashtableRelease(ht);
     }
-    assert(kvs->overhead_hashtable_lut == 0);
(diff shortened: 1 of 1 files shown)


ranshid (approved):
LGTM but is the motivation to help dailys? or to prevent production assert?
In case the later, maybe just turn to debug assert?

zuiderkwast:
> LGTM but is the motivation to help dailys? or to prevent production assert? In case the later, maybe just turn to debug assert?

Both. I considered debug assert but kvstore doesn't include server.h so we don't have it.

PR #1750, opened 2025-02-18
Fix hashtable memory leak and kvstore `overhead_hashtable_lut` assert

### LINKED ISSUE #1657: [test-failure] Unit test failure in test_kvstore.c

artikell:
When releasing a kvstore and its hashtables, if it is still rehashed and table 0 has no data, it may still have allocated buckets that were not released. This lead to allocated hashtable buckets being leaked. This looked like missed statistics but it was actually a memory leak.

fix: https://github.com/valkey-io/valkey/issues/1657

### DIFF
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -1032,31 +1032,29 @@ void hashtableEmpty(hashtable *ht, void(callback)(hashtable *)) {
         if (ht->bucket_exp[table_index] < 0) {
             continue;
         }
-        if (ht->used[table_index] > 0) {
-            for (size_t idx = 0; idx < numBuckets(ht->bucket_exp[table_index]); idx++) {
-                if (callback && (idx & 65535) == 0) callback(ht);
-                bucket *b = &ht->tables[table_index][idx];
-                do {
-                    /* Call the destructor with each entry. */
-                    if (ht->type->entryDestructor != NULL && b->presence != 0) {
-                        for (int pos = 0; pos < ENTRIES_PER_BUCKET; pos++) {
-                            if (isPositionFilled(b, pos)) {
-                                ht->type->entryDestructor(b->entries[pos]);
-                            }
+        for (size_t idx = 0; idx < numBuckets(ht->bucket_exp[table_index]); idx++) {
+            if (callback && (idx & 65535) == 0) callback(ht);
+            bucket *b = &ht->tables[table_index][idx];
+            do {
+                /* Call the destructor with each entry. */
+                if (ht->type->entryDestructor != NULL && b->presence != 0) {
+                    for (int pos = 0; pos < ENTRIES_PER_BUCKET; pos++) {
+                        if (isPositionFilled(b, pos)) {
+                            ht->type->entryDestructor(b->entries[pos]);
                         }
                     }
-                    bucket *next = getChildBucket(b);
+                }
+                bucket *next = getChildBucket(b);
 
-                    /* Free allocated bucket. */
-                    if (b != &ht->tables[table_index][idx]) {
-                        zfree(b);
-                        if (ht->type->trackMemUsage) {
-                            ht->type->trackMemUsage(ht, -sizeof(bucket));
-                        }
+                /* Free allocated bucket. */
+                if (b != &ht->tables[table_index][idx]) {
+                    zfree(b);
+                    if (ht->type->trackMemUsage) {
+                        ht->type->trackMemUsage(ht, -sizeof(bucket));
                     }
-                    b = next;
-                } while (b != NULL);
-            }
+                }
+                b = next;
+            } while (b != NULL);
         }
         zfree(ht->tables[table_index]);
         if (ht->type->trackMemUsage) {
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -1058,7 +1058,7 @@ void hashtableEmpty(hashtable *ht, void(callback)(hashtable *)) {
                 } while (b != NULL);
             }
         }
-        
+
         zfree(ht->tables[table_index]);
         if (ht->type->trackMemUsage) {
             ht->type->trackMemUsage(ht, -sizeof(bucket) * numBuckets(ht->bucket_exp[table_index]));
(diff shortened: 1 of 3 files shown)


artikell:
@zuiderkwast Please help me review it, maybe you have a better way to fix it.

src/unit/test_kvstore.c
```diff
@@ -70,6 +70,13 @@ int test_kvstoreIteratorRemoveAllKeysNoDeleteEmptyHashtable(int argc, char **arg
     UNUSED(argv);
     UNUSED(flags);
 
+    uint8_t seed[16];
+    int cycle = 17;
+    for (unsigned int i = 0; i < sizeof(seed); i++) {
+        seed[i] = (i + cycle) % 0xFF;
+    }
+    hashtableSetHashFunctionSeed(seed);
+
```
zuiderkwast: It's a good idea to seed the hash function, but why not use `getRandomBytes` as we do in test_hashtable.c?

src/hashtable.c
```diff
@@ -1032,31 +1032,29 @@ void hashtableEmpty(hashtable *ht, void(callback)(hashtable *)) {
         if (ht->bucket_exp[table_index] < 0) {
             continue;
         }
-        if (ht->used[table_index] > 0) {
```
zuiderkwast, line 1035: Can we use `ht->used[table_index] > 0 || ht->child_buckets[table_index] > 0`?

`ht->child_buckets[table_index]` is the number of allocated child buckets in the table that we need to free.

zuiderkwast (commented):
Awesome! You found it!

I realized the diff is small if I review it with [whitespace changes hidden](https://github.com/valkey-io/valkey/pull/1750/files?diff=unified&w=1).

It means we also have a memory leak, right? We don't free the allocated child buckets?

rjd15372:
I'm pretty sure this will also fix the assert errors that I was investigating in today's daily tests:
https://github.com/valkey-io/valkey/actions/runs/13380552486/job/37368256686
https://github.com/valkey-io/valkey/actions/runs/13380552486/job/37368260953

src/hashtable.c
zuiderkwast: It's still weird that there are empty buckets in a table without any elements. Those should be deleted when the elements are deleted....

zuiderkwast: I've spotted something!

`kvstoreIteratorNext` calls `hashtableReinitIterator` to init the same iterator with the next hash table, but it doesn't reset the iterator for the previous hash table, so it leaves rehashing paused.

I think we need this:

```diff
 /* Reinitializes the iterator for the provided hashtable while
  * preserving the flags from its previous initialization. */
 void hashtableReinitIterator(hashtableIterator *iterator, hashtable *ht) {
     iter *iter = iteratorFromOpaque(iterator);
+    hashtableResetIterator(iterator);
     hashtableInitIterator(iterator, ht, iter->flags);
 }
```

src/unit/test_kvstore.c
artikell: In fact, the original logic is a kind of random, so it is difficult to find.

I think this requires a separate test, and we can output the seed

src/hashtable.c
artikell: Yes, I didn't notice child_buckets, I'll add these two logics

src/unit/test_kvstore.c
artikell: I added a KvstoreConflictHashtableTestType to ensure a high conflict rate

src/hashtable.c
artikell: There seems to be some problem with adding hashtableResetIterator. Maybe we can check it separately. @zuiderkwast

zuiderkwast: Sure, we can do it separately, but I think this might be a big problem. Once kvstore iterates over the hash tables, it leaves rehashing paused on them forever...

What was the problem?

artikell: I haven't had time to locate it today, but here are two test results:
https://github.com/valkey-io/valkey/actions/runs/13403931169/job/37440255634
https://github.com/valkey-io/valkey/actions/runs/13403899712/job/37440162349

It causes a crash

zuiderkwast: Ok, I see the assert:

```
10180:M 19 Feb 2025 02:08:05.703 # === ASSERTION FAILED ===
10180:M 19 Feb 2025 02:08:05.703 # ==> hashtable.c:1887 'iter->hashtable->pause_rehash >= 0' is not true
```

This means it was reset twice, which unpaused rehashing twice.

Looking at kvstoreIteratorNext again, it calls kvstoreIteratorNextHashtable which resets the iterator, but it seems to miss this for the last hashtable it iterates over. Notice the `return NULL` on the first line:

```c
/* Returns next hash table from the iterator, or NULL if iteration is complete. */
static hashtable *kvstoreIteratorNextHashtable(kvstoreIterator *kvs_it) {
    if (kvs_it->next_didx == -1) return NULL;

    /* The hashtable may be deleted during the iteration process, so here need to check for NULL. */
    if (kvs_it->didx != -1 && kvstoreGetHashtable(kvs_it->kvs, kvs_it->didx)) {
        /* Before we move to the next hashtable, reset the iter of the previous hashtable. */
        hashtableIterator *iter = &kvs_it->di;
        hashtableResetIterator(iter);
        /* In the safe iterator context, we may delete entries. */
        freeHashtableIfNeeded(kvs_it->kvs, kvs_it->didx);
    }

    kvs_it->didx = kvs_it->next_didx;
    kvs_it->next_didx = kvstoreGetNextNonEmptyHashtableIndex(kvs_it->kvs, kvs_it->didx);
    return kvs_it->kvs->hashtables[kvs_it->didx];
}
```

zuiderkwast (approved):
Nice, this fixes the memory leak.

There seems to be a problem with iterator not being reset properly which I think leads to rehashing being paused forever, which is quite serious if this is true. We can try to think about it but otherwise I'll merge this later today.

PR #1751, opened 2025-02-19
Allow to specify the random number generator seed in unit tests

rjd15372:
In this commit, we add a new parameter to the unit tests binary, called `--seed`, that allows to specify the seed used by the several random numbers generators used in the Valkey server code.

We also now print the seed information right before starting the unit tests execution, so that someone that needs to reproduce an error, can use the same seed to re-run the failed tests.

### DIFF
--- a/src/util.c
+++ b/src/util.c
@@ -51,6 +51,7 @@
 #include "sha256.h"
 #include "config.h"
 #include "zmalloc.h"
+#include "serverassert.h"
 
 #include "valkey_strtod.h"
 
@@ -905,16 +906,38 @@ int version2num(const char *version) {
     return v;
 }
 
+/* Global state. */
+static int seed_initialized = 0;
+static unsigned char seed[64]; /* 512 bit internal block size. */
+
+void setRandomSeedCString(char *seed_str, size_t len) {
+    assert(seed_initialized == 0);
+    assert(len == sizeof(seed) * 2);
+    for (size_t i = 0; i < sizeof(seed); i++) {
+        sscanf(seed_str + (i * 2), "%02hhX", &seed[i]);
+    }
+    seed_initialized = 1;
+}
+
+char *getRandomSeedCString(size_t *len) {
+    char *buff = zmalloc(sizeof(seed) * 2 + 1);
+    for (size_t i = 0; i < sizeof(seed); i++) {
+        snprintf(buff + (i * 2), 3, "%02hhX", seed[i]);
+    }
+    buff[sizeof(seed) * 2] = 0;
+    if (len) {
+        *len = sizeof(seed) * 2;
+    }
+    return buff;
+}
+
 /* Get random bytes, attempts to get an initial seed from /dev/urandom and
  * the uses a one way hash function in counter mode to generate a random
  * stream. However if /dev/urandom is not available, a weaker seed is used.
  *
  * This function is not thread safe, since the state is global. */
 void getRandomBytes(unsigned char *p, size_t len) {
-    /* Global state. */
-    static int seed_initialized = 0;
-    static unsigned char seed[64]; /* 512 bit internal block size. */
-    static uint64_t counter = 0;   /* The counter we hash with the seed. */
+    static uint64_t counter = 0; /* The counter we hash with the seed. */
 
     if (!seed_initialized) {
         /* Initialize a seed and use SHA1 in counter mode, where we hash
--- a/src/util.c
+++ b/src/util.c
@@ -943,7 +943,7 @@ void setRandomSeedCString(char *seed_str, size_t len) {
 }
 
 /* This function populates a char buffer with 129 bytes with the 64 bytes
- * random seed encoded as 128 hexadecimal digits. */
+ * random seed encoded as 128 hexadecimal digits and a null terminator. */
 void getRandomSeedCString(char *buff, size_t len) {
     assert(len == (sizeof(seed) * 2 + 1));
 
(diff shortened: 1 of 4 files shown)


zuiderkwast (commented):
Should we remove the random seed we do in test_hashtable.c and maybe some other test suites? Otherwise, it's still not reproducible...

artikell:
Maybe we can support a hook point for beforeTest and afterTest, which will help us restore the scene without affecting other cases.

rjd15372:
> Should we remove the random seed we do in test_hashtable.c and maybe some other test suites? Otherwise, it's still not reproducible...

I think it does not matter. We are setting up the seed used by `getRandomBytes` in the beginning of the test. Therefore, the bytes generated in `randomSeed` of the `test_hashtable.c` test, will always be the same if we set the same seed.

zuiderkwast:
> We are setting up the seed used by `getRandomBytes` in the beginning of the test. Therefore, the bytes generated in `randomSeed` of the `test_hashtable.c` test, will always be the same if we set the same seed.

Are you sure? `test_hashtable.c` uses `randomSeed` uses `getRandomBytes` which takes from `/dev/random`, so i would guess it does matter.

rjd15372:
> Are you sure? `test_hashtable.c` uses `randomSeed` uses `getRandomBytes` which takes from `/dev/random`, so i would guess it does matter.

But `getRandomBytes` only uses `/dev/random` to initialize the seed in the case that the seed has not been initialized before.
Since we are initializing the seed in the beginning of the test, the following calls to `getRandomBytes` will use the existing seed.

zuiderkwast:
> Maybe we can support a hook point for beforeTest and afterTest, which will help us restore the scene without affecting other cases.

Do we need that? It seem like overkill for this purpose. I'd suggest we just remove all random seeding in each test suite and only do it once by the test framework. And print it in the beginning.

It's nice that this test framework is extremely simple, without scaffolding like beforeSuite, beforeTestCase.

zuiderkwast:
@rjd15372 I see now that getRandomBytes only reads `/dev/random` once, so I think you're right that it's reproducible, but only if we run all tests together so getRandomBytes is called the same number of times. If we want to run a single test suite with `--single` and `--seed`, there can be a difference, right?

The test framework could can re-seed using the same seed before each test suite, maybe....

rjd15372:
Correct. It's only reproducible if you run the test with the same CLI arguments plus the seed arg. If you use the `--single` arg, it will not be the same.

Yeah, maybe re-seeding before each test run would allow us to run a single test. I'll update the PR.

src/util.c
```diff
@@ -905,36 +906,70 @@ int version2num(const char *version) {
...
+}
+
+/* This function returns a string with 64 bytes encoded as 128 hexadecimal
+ * digits, that represents the 64 bytes random seed. */
+char *getRandomSeedCString(size_t *len) {
+    if (!seed_initialized) {
+        initializeRandomSeed();
+    }
+
+    char *buff = zmalloc(sizeof(seed) * 2 + 1);
```
zuiderkwast: Why not move the allocation out of this function? Then the caller can allocate it using zmalloc, malloc or stack allocated.

If this is fixed to 64 bytes seed, then add the `assert(len == sizeof(seed) * 2);` here too?

```suggestion
/* This function returns a string with 64 bytes encoded as 128 hexadecimal
 * digits, that represents the 64 bytes random seed. */
char *getRandomSeedCString(char *buff, size_t *len) {
    if (!seed_initialized) {
        initializeRandomSeed();
    }
    assert(len == sizeof(seed) * 2);
```

rjd15372:
@zuiderkwast what do you think of doing something similar to this PR for TCL tests?

zuiderkwast:
> @zuiderkwast what do you think of doing something similar to this PR for TCL tests?

That's be great but it seems much harder. It would need to involve some server config or debug command to set the seed, right? And Lua's and TCL's internal seeds too?

A fixed hashtable seed per slot that's transferred on failover and slot migrations is something we've been thinking about in #4. It's a different purpose but it seems to be related and somewhat overlapping in parts of implementation I guess.

rjd15372:
> > @zuiderkwast what do you think of doing something similar to this PR for TCL tests?
> 
> That's be great but it seems much harder. It would need to involve some server config or debug command to set the seed, right? And Lua's and TCL's internal seeds too?

I was thinking about something simpler. Just set the seed on the TCL side. The server would still be random but it would be better than nothing.

src/util.c
```diff
@@ -905,36 +906,67 @@ int version2num(const char *version) {
...
+void setRandomSeedCString(char *seed_str, size_t len) {
+    assert(len == sizeof(seed) * 2);
+    for (size_t i = 0; i < sizeof(seed); i++) {
+        sscanf(seed_str + (i * 2), "%02hhX", &seed[i]);
+    }
+    seed_initialized = 1;
+}
+
+/* This function populates a char buffer with 129 bytes with the 64 bytes
+ * random seed encoded as 128 hexadecimal digits. */
```
zuiderkwast:
```suggestion
/* This function populates a char buffer with 129 bytes with the 64 bytes
 * random seed encoded as 128 hexadecimal digits and a null terminator. */
```

src/util.c
```diff
@@ -905,36 +906,67 @@ int version2num(const char *version) {
...
+    assert(len == (sizeof(seed) * 2 + 1));
+
+    if (!seed_initialized) {
+        initializeRandomSeed();
+    }
+
+    for (size_t i = 0; i < sizeof(seed); i++) {
+        snprintf(buff + (i * 2), 3, "%02hhX", seed[i]);
+    }
+    buff[sizeof(seed) * 2] = 0;
```
zuiderkwast: `\0` is a more common way to express the ASCII nul character.

```suggestion
    buff[sizeof(seed) * 2] = '\0';
```

zuiderkwast (approved):

PR #1756, opened 2025-02-20
make net_input_bytes_curr_cmd more readable

soloestoy:
```
        /* Per-slot network bytes-in calculation, 3rd and 4th components.
         * Here, the deferred c->argc from 2nd component is added, resulting in c->argc * 5 instead of * 4. */
        c->net_input_bytes_curr_cmd += (c->argv_len_sum + (c->argc * 5 + 2));
```

When I was reading the code, this line confused me a bit. Why is it necessary to replace multiplying by 4 with multiplying by 5 here? Then I realized it was because the '$' character is included. So, I made some changes to make this calculation consistent with the description in the comments. Additionally, the counting for "\r\n" has also been moved to each part separately. I think this makes it more intuitive. WDYT @kyle-yh-kim ?

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -2854,30 +2854,30 @@ void processMultibulkBuffer(client *c) {
          *
          * Calculation: For multi bulk buffer, we accumulate four factors, namely;
          *
-         * 1) multibulklen_slen + 1
+         * 1) multibulklen_slen + 3
          *    Cumulative string length (and not the value of) of multibulklen,
-         *    including +1 from RESP first byte.
-         * 2) bulklen_slen + c->argc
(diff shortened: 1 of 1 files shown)


soloestoy:
@enjoy-binbin @PingXie would you be interested in reviewing this PR? Given your experience in code maintainability, your insights would be particularly valuable.

enjoy-binbin (approved):
Seems better to me. @kyle-yh-kim please ack the logic.

kyle-yh-kim:
Thanks for looking into this. 

Agreed, looks better this way - having the calculations correctly reflect each component, rather than having them scattered across adjacent components. Approved!

PR #1762, opened 2025-02-21
Update and stabilize defrag tests

### LINKED ISSUE #1746: Active defrag pubsub, defrag didn't stop
These two failures often come together:

```
 *** [err]: Active defrag pubsub: standalone in tests/unit/memefficiency.tcl
defrag didn't stop.
*** [err]: Active defrag big list: standalone in tests/unit/memefficiency.tcl
Expected 1.54 >= 1.7 (context: type eval line 39 cmd {assert {$frag >= $expected_frag}} proc ::test)
```

https://github.com/valkey-io/valkey/actions/runs/13360144721/job/37308505262#step:7:9123

JimB123:
Resolves: https://github.com/valkey-io/valkey/issues/1746
A number of tests related to defrag have had stability problems.

One reason for stability issues is that the tests are run sequentially on the same server, and it takes jemalloc some time to release freed pages.  This is especially noticed immediately after a flushall.  Over a period of 10 seconds, it is observable that the "fragmentation bytes" can decrease by several MB.  Another reason is that there's no standardization between tests.  For each test, people have been independently hacking/tweaking the success criteria, without addressing underlying issues.

This update revamps all of the defrag tests:
* A fresh server is started for each test.  Running each test in isolation improves stability.
* A uniform function `log_frag` is now used for debug logging
* A uniform function `perform_defrag_test` ensures that each test is written and executed in a uniform fashion.  Limits are imposed to ensure that the defrag results are consistent/reproducible.  The intent is to eliminate failures do to various tweaks to values in individual tests.
* Latency is tested much more strictly for most tests, reflecting the recent improvements to defrag latency.
* The test `defrag edge case` has been removed.  This test attempted to create N pages with EXACTLY equal fragmentation in an attempt to confuse the defrag logic.  It's unlikely that this test was performing correctly, and had questionable value.
* Tests for hash/list/set/zset/stream have been separated and standardized.  It was unlikely that the old test was performing properly as none of the actual data structures were fragmented!

It's noted that pubsub doesn't appear to be defragging correctly.  The old test was based on deletion of strings (only) which doesn't actually reflect what happens when a pubsub channel is removed.  The test has been reduced to only check that pubsub is not damaged during defrag - but doesn't test for defrag efficacy.  This isn't likely a significant issue as it would be unlikely to create many thousands of pubsub channels and then have associated fragmentation issues.  https://github.com/valkey-io/valkey/issues/1774

### DIFF
--- a/tests/support/test.tcl
+++ b/tests/support/test.tcl
@@ -124,7 +124,8 @@ proc assert_refcount_morethan {key ref} {
 
 # Wait for the specified condition to be true, with the specified number of
 # max retries and delay between retries. Otherwise the 'elsescript' is
-# executed.
+# executed. If 'debugscript' is provided, it is executed after failure of
+# the confition (before the retry delay).
 proc wait_for_condition {maxtries delay e _else_ elsescript {_debug_ ""} {debugscript ""}} {
     while {[incr maxtries -1] >= 0} {
         set errcode [catch {uplevel 1 [list expr $e]} result]
--- a/tests/support/test.tcl
+++ b/tests/support/test.tcl
@@ -125,13 +125,16 @@ proc assert_refcount_morethan {key ref} {
 # Wait for the specified condition to be true, with the specified number of
 # max retries and delay between retries. Otherwise the 'elsescript' is
 # executed.
-proc wait_for_condition {maxtries delay e _else_ elsescript} {
+proc wait_for_condition {maxtries delay e _else_ elsescript {_debug_ ""} {debugscript ""}} {
     while {[incr maxtries -1] >= 0} {
         set errcode [catch {uplevel 1 [list expr $e]} result]
         if {$errcode == 0} {
             if {$result} break
         } else {
             return -code $errcode $result
+        }
+        if {$_debug_ == "debug"} {
+            uplevel 1 $debugscript
         }
         after $delay
     }
--- a/tests/unit/memefficiency.tcl
+++ b/tests/unit/memefficiency.tcl
@@ -189,7 +189,7 @@ run_solo {defrag} {
 
         # populate the DB with data
         set initial_allocated [s allocator_allocated]
-        uplevel 1 $pop_code
+        uplevel 1 $opts(populate)
 
         log_frag "after adding data"
         # check that enough data has been populated
--- a/tests/unit/memefficiency.tcl
+++ b/tests/unit/memefficiency.tcl
@@ -567,6 +567,7 @@ run_solo {defrag} {
     set tests {}
     lappend tests [list test_main_dictionary standalone $aof_overrides] ;# only need AOF for main dict, standalone
     lappend tests [list test_eval_scripts standalone $std_overrides]
+    # https://github.com/valkey-io/valkey/issues/1795
     # lappend tests [list test_big_hash standalone $std_overrides]
     lappend tests [list test_big_list standalone $std_overrides]
     lappend tests [list test_big_set standalone $std_overrides]
@@ -576,6 +577,7 @@ run_solo {defrag} {
 
     lappend tests [list test_main_dictionary cluster $std_overrides]
     lappend tests [list test_eval_scripts cluster $std_overrides]
+    # https://github.com/valkey-io/valkey/issues/1795
     # lappend tests [list test_big_hash cluster $std_overrides]
     lappend tests [list test_big_list cluster $std_overrides]
     lappend tests [list test_big_set cluster $std_overrides]
--- a/tests/unit/memefficiency.tcl
+++ b/tests/unit/memefficiency.tcl
@@ -489,9 +489,10 @@ run_solo {defrag} {
             set chan [string repeat A 50]
 
             # https://github.com/valkey-io/valkey/issues/1774
-            # NOTE - pubsub defrag isn't working properly.  This wasn't caught before the test refactor.
-            #  This commented out code should work.  Instead, is substituted code which verifies that
-            #  defrag doesn't break the pubsub, but defrag efficacy is not verified.
+            # NOTE - pubsub defrag isn't working properly.  This wasn't caught before the
+            #  test refactor.  This commented out code should work.  Instead, is substituted
+            # code which verifies that defrag doesn't break the pubsub, but defrag efficacy
+            # is not verified.
             # perform_defrag_test $title {
             #     for {set j 0} {$j < $n} {incr j} {
             #         set channel $chan$j
--- a/tests/unit/memefficiency.tcl
+++ b/tests/unit/memefficiency.tcl
@@ -153,7 +153,7 @@ run_solo {defrag} {
         } elseif {$op == ">"} {
             assert_morethan $frag $value
         } else {
-            error "Operator value($op) must be 'LT' or 'GT'"
+            error "Operator value($op) must be '<' or '>'"
         }
     }
 
--- a/tests/unit/memefficiency.tcl
+++ b/tests/unit/memefficiency.tcl
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
     }
 }
 
+
 run_solo {defrag} {
-    proc test_active_defrag {type} {
-    if {[string match {*jemalloc*} [s mem_allocator]] && [r debug mallctl arenas.page] <= 8192} {
-        test "Active defrag main dictionary: $type" {
-            r config set activedefrag no
-            r config set active-defrag-threshold-lower 5
-            r config set active-defrag-cycle-min 65
-            r config set active-defrag-cycle-max 75
-            r config set active-defrag-ignore-bytes 2mb
-            r config set maxmemory 100mb
-            r config set maxmemory-policy allkeys-lru
-            r config set lazyfree-lazy-user-del no
-            r config set lazyfree-lazy-user-flush no
+    set ::one_mb [expr 1024 * 1024]
 
-            populate 700000 asdf1 150
-            populate 100 asdf1 150 0 false 1000
-            populate 170000 asdf2 300
-            populate 100 asdf2 300 0 false 1000
+    # make logging a bit more readable
+    proc to_mb {bytes} {
+        return [format "%6.2f MB" [expr $bytes / 1024.0 / 1024.0]]
+    }
 
-            assert {[scan [regexp -inline {expires\=([\d]*)} [r info keyspace]] expires=%d] > 0}
-            after 120 ;# serverCron only updates the info once in 100ms
-            set frag [s allocator_frag_ratio]
-            if {$::verbose} {
-                puts "frag $frag"
+    # used to create keys which increase in size by 2x
+    # key 0 will have 2 fields (0,1), key 1 will have 4 (0,1,2,3), key 2 will have 8 ...
+    # incby allows incrementing by 1 or 2 (but no other values will work properly)
+    proc next_exp_kf {key field {incby 1}} {
+        incr field $incby
+        if {$field == 2 ** ($key+1)} {
+            incr key
+            set field 0
+        }
+        return [list $key $field]
+    }
+
+    # Logs defragging state if ::verbose is true
+    proc log_frag {title} {
+        # Note, this delay is outside of the "if" so that behavior is the same, with and
+        #  without verbose logging.  Also, this allows for assertions to follow without
+        #  adding additional delay statements
+        after 120  ;# serverCron only updates the info once in 100ms
+        if {$::verbose} {
+            puts "Frag info: $title"
+            puts "    used: [to_mb [s allocator_allocated]]"
+            puts "    frag: [to_mb [s allocator_frag_bytes]]"
+            puts "    frag ratio: [s allocator_frag_ratio]"
+            puts "    hits: [s active_defrag_hits]"
+            puts "    misses: [s active_defrag_misses]"
+        }
+    }
+
+    # Start defrag and wait for it to stop
+    # The optional code block is executed after defrag has started
+    proc perform_defrag {{code_block {}}} {
+        r config set latency-monitor-threshold 5
+        r latency reset
+
+        set old_defrag_time [s total_active_defrag_time]
+        r config set activedefrag yes
+
+        # Wait for defrag to start
+        wait_for_condition 10 50 {
+            [s total_active_defrag_time] ne $old_defrag_time
+        } else {
+            log_frag "defrag not started"
+            puts [r info memory]
+            puts [r info stats]
+            puts [r memory malloc-stats]
+            fail "defrag not started."
+        }
+
+        # Execute the passed in code block
+        uplevel 1 $code_block
+
+        # Wait for the active defrag to stop working.
+        wait_for_condition 150 200 {
+            [s active_defrag_running] eq 0
+        } else {
+            log_frag "defrag didn't stop"
+            puts [r info memory]
+            puts [r memory malloc-stats]
+            fail "defrag didn't stop."
+        } debug {
+            log_frag "defragging..."
+        }
+
+        r config set activedefrag no
+        after 120 ;# ensure memory stats are current when function exits
+    }
+
+    # Checks for any significant latency events
+    proc validate_latency {limit_ms} {
+        if {!$::no_latency} {
+            set max_latency 0
+            foreach event [r latency latest] {
+                lassign $event eventname time latency max
+                if {$eventname == "active-defrag-cycle" || $eventname == "while-blocked-cron"} {
+                    set max_latency $max
+                }
             }
-            assert {$frag >= 1.4}
-
-            r config set latency-monitor-threshold 5
-            r latency reset
-            r config set maxmemory 110mb ;# prevent further eviction (not to fail the digest test)
-            set digest [debug_digest]
-            catch {r config set activedefrag yes} e
-            if {[r config get activedefrag] eq "activedefrag yes"} {
-                # Wait for the active defrag to start working (decision once a
-                # second).
-                wait_for_condition 50 100 {
-                    [s total_active_defrag_time] ne 0
-                } else {
-                    after 120 ;# serverCron only updates the info once in 100ms
-                    puts [r info memory]
-                    puts [r info stats]
-                    puts [r memory malloc-stats]
-                    fail "defrag not started."
-                }
-
-                # This test usually runs for a while, during this interval, we test the range.
-                assert_range [s active_defrag_running] 65 75
-                r config set active-defrag-cycle-min 1
-                r config set active-defrag-cycle-max 1
-                after 120 ;# serverCron only updates the info once in 100ms
-                assert_range [s active_defrag_running] 1 1
-                r config set active-defrag-cycle-min 65
-                r config set active-defrag-cycle-max 75
-
-                after 1000 ;# Give defrag time to work (might be multiple cycles)
-
-                # Wait for the active defrag to stop working.
-                wait_for_condition 2000 100 {
-                    [s active_defrag_running] eq 0
-                } else {
-                    after 120 ;# serverCron only updates the info once in 100ms
-                    puts [r info memory]
-                    puts [r memory malloc-stats]
-                    fail "defrag didn't stop."
-                }
-
-                # Test the fragmentation is lower.
-                after 120 ;# serverCron only updates the info once in 100ms
-                set frag [s allocator_frag_ratio]
-                set max_latency 0
-                foreach event [r latency latest] {
-                    lassign $event eventname time latency max
-                    if {$eventname == "active-defrag-cycle"} {
-                        set max_latency $max
-                    }
-                }
-                if {$::verbose} {
-                    puts "frag $frag"
-                    set misses [s active_defrag_misses]
-                    set hits [s active_defrag_hits]
-                    puts "hits: $hits"
-                    puts "misses: $misses"
-                    puts "max latency $max_latency"
+            if {$::verbose} {
+                puts "Validating max latency ($max_latency) is LT $limit_ms"
+                if {$max_latency > 0} {
                     puts [r latency latest]
                     puts [r latency history active-defrag-cycle]
-                }
-                assert {$frag < 1.1}
-                # due to high fragmentation, 100hz, and active-defrag-cycle-max set to 75,
-                # we expect max latency to be not much higher than 7.5ms but due to rare slowness threshold is set higher
-                if {!$::no_latency} {
-                    assert {$max_latency <= 30}
+                    puts [r latency history while-blocked-cron]
                 }
             }
-            # verify the data isn't corrupted or changed
-            set newdigest [debug_digest]
-            assert {$digest eq $newdigest}
-            r save ;# saving an rdb iterates over all the data / pointers
+            assert {$max_latency <= $limit_ms}
+        }
+    }
 
-            # if defrag is supported, test AOF loading too
-            if {[r config get activedefrag] eq "activedefrag yes" && $type eq "standalone"} {
+    # Validate expected fragmentation ratio
+    # op = "LT" (less than) or "GT" (greater than)
+    proc validate_frag_ratio {op value} {
+        set frag [s allocator_frag_ratio]
+        if {$::verbose} {
+            puts "Validating frag ($frag) is $op $value"
+        }
+        set allocated_bytes [s allocator_allocated]
+        if {$allocated_bytes < 20 * $::one_mb} {
+            # If allocated bytes is too small, the ratios get wonky.  Since we use 2MB for
+            # active-defrag-ignore-bytes, let's make sure that we have at least 10x that amount
+            # allocated before trying to verify any fragmentation ratios.  Otherwise the tests
+            # are likely to get flaky.
+            error "test error: trying to validate frag ratio with only $allocated_bytes allocated"
+        }
+        if {$op == "LT"} {
+            assert_lessthan $frag $value
+        } elseif {$op == "GT"} {
+            assert_morethan $frag $value
+        } else {
+            error "Operator value($op) must be 'LT' or 'GT'"
+        }
+    }
+
+    # Performs a standardized defrag test.  The first code block generates data.  The second code block
+    # fragments that data (usually be deleting half).
+    #  - caller must generate at least 40 MB of data
+    #  - after fragmentation, at least 20 MB of data must remain, and fragmentation ratio must exceed 1.4
+    # Positional parameters:
+    #    name - name of the test
+    #    pop_code - code which populates (unfragemented) data
+    # Named parameters:
+    #    fragment {code} - required, fragments the populated data
+    #    while_defragging {code} - optional, code executed after defrag has started
+    #    latency <ms> - optional, verifies the latency to a ms target (default 5)
+    proc perform_defrag_test {name pop_code args} {
+        set opts(latency) 5
+        set opts(while_defragging) {}
+        array set opts $args
+        assert {[info exists opts(fragment)]}
+
+        r config set active-defrag-threshold-lower 5
+        r config set active-defrag-cycle-min 40
+        r config set active-defrag-cycle-max 60
+        r config set active-defrag-ignore-bytes 2mb
+        r config set maxmemory 0
+
+        log_frag "test start ($name)"
+
+        # at this point, we expect fragmentation BYTES to be below a threshold
+        #  (percentage is wildly variable with low memory usage)
+        assert {[s allocator_frag_bytes] < 2 * $::one_mb}
+
+        # populate the DB with data
+        set initial_allocated [s allocator_allocated]
+        uplevel 1 $pop_code
+
+        log_frag "after adding data"
+        # check that enough data has been populated
+        set required [expr 40 * $::one_mb]
+        set allocated [expr [s allocator_allocated] - $initial_allocated]
+        if {$allocated < $required} {
+            fail "Tests are required to create at least $required bytes of data before fragmentation - only $allocated bytes created"
+        }
+        # check that at this point, fragmentation is very low
+        validate_frag_ratio LT 1.05
+
+        # fragment the data
+        uplevel 1 $opts(fragment)
+
+        log_frag "after fragmenting data"
+        # we want some fragmentation, but we still want a minimum allocation
+        validate_frag_ratio GT 1.4
+        set required [expr 20 * $::one_mb]
+        set allocated [expr [s allocator_allocated] - $initial_allocated]
+        if {$allocated < $required} {
+            fail "Tests are required to retain at least $required bytes of data after fragmentation - only $allocated bytes retained"
+        }
+
+        set digest [debug_digest]
+
+        perform_defrag $opts(while_defragging)
+
+        log_frag "after defragging"
+        validate_frag_ratio LT 1.1
+        validate_latency $opts(latency)
+
+        # verify the data isn't corrupted or changed
+        set newdigest [debug_digest]
+        assert {$digest eq $newdigest}
+        r save ;# saving an rdb iterates over all the data / pointers
+    }
+
+    proc test_main_dictionary {type} {
+        set title "Active Defrag main dictionary: $type"
+        test $title {
+            set n 200000
+
+            perform_defrag_test $title {
+                # add a mass of string keys
+                set rd [valkey_deferring_client]
+                for {set j 0} {$j < $n} {incr j} {
+                    $rd setrange $j 200 a
+                    if {$j % 3 == 0} {
+                        $rd expire $j 1000 ;# put expiration on some
+                    }
+                }
+                for {set j 0} {$j < $n} {incr j} {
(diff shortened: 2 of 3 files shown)


tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
-                }
+    # Validate expected fragmentation ratio
+    # op = "LT" (less than) or "GT" (greater than)
+    proc validate_frag_ratio {op value} {
+        set frag [s allocator_frag_ratio]
+        if {$::verbose} {
+            puts "Validating frag ($frag) is $op $value"
+        }
+        set allocated_bytes [s allocator_allocated]
+        if {$allocated_bytes < 20 * $::one_mb} {
```
zuiderkwast: If the point of `$::one_mb` was to avoid the multiplication in various places, then it seems we need a `$::twenty_mb` global too (not). Better just skip the global I think.

```suggestion
        if {$allocated_bytes < 20 * 1024 * 1024} {
```

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
-            populate 100 asdf1 150 0 false 1000
-            populate 170000 asdf2 300
-            populate 100 asdf2 300 0 false 1000
-
-            assert {[scan [regexp -inline {expires\=([\d]*)} [r info keyspace]] expires=%d] > 0}
-            after 120 ;# serverCron only updates the info once in 100ms
-            set frag [s allocator_frag_ratio]
-            if {$::verbose} {
-                puts "frag $frag"
+    set ::one_mb [expr 1024 * 1024]
```
zuiderkwast: I don't think we need this variable.

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
+
+    # make logging a bit more readable
+    proc to_mb {bytes} {
+        return [format "%6.2f MB" [expr $bytes / 1024.0 / 1024.0]]
+    }
```
zuiderkwast, line 44: Let's put these functions on top-level, i.e. outside of the `run_solo` block.

Possibly we can put some of these functions in some utility library, if you find a good place, but that's not necessary.

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
+
+    # used to create keys which increase in size by 2x
+    # key 0 will have 2 fields (0,1), key 1 will have 4 (0,1,2,3), key 2 will have 8 ...
+    # incby allows incrementing by 1 or 2 (but no other values will work properly)
+    proc next_exp_kf {key field {incby 1}} {
```
zuiderkwast, line 52: I don't really understand what this function returns and these comments don't seem to explain it in a way that I can understand. I can see it returns a list of two elements.

Can we name this function in a more generic way? It doesn't seem to be be specific to keys and fields, but rather to a sequence of numbers.

tests/support/test.tcl
```diff
@@ -125,14 +125,17 @@ proc assert_refcount_morethan {key ref} {
 # Wait for the specified condition to be true, with the specified number of
 # max retries and delay between retries. Otherwise the 'elsescript' is
 # executed.
-proc wait_for_condition {maxtries delay e _else_ elsescript} {
+proc wait_for_condition {maxtries delay e _else_ elsescript {_debug_ ""} {debugscript ""}} {
```
zuiderkwast, line 129: The word `debug` describes what we use it for rather than what it does. It's evaluated between each retry and it's possible to do anything here, not only debug stuff. I think it should be called something that better describes what it does, like `on-retry`?

Or do you have a good explanation of why it should be called `debug`?

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
+            # If allocated bytes is too small, the ratios get wonky.  Since we use 2MB for
+            # active-defrag-ignore-bytes, let's make sure that we have at least 10x that amount
+            # allocated before trying to verify any fragmentation ratios.  Otherwise the tests
+            # are likely to get flaky.
+            error "test error: trying to validate frag ratio with only $allocated_bytes allocated"
+        }
+        if {$op == "LT"} {
+            assert_lessthan $frag $value
+        } elseif {$op == "GT"} {
+            assert_morethan $frag $value
```
zuiderkwast: I believe it'd be possible to use `<` and `>` instead of LT and GT. There's nothing special about these symbols in TCL.

How does this look?

```TCL
validate_frag_ratio < 1.05
```

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
+        } else {
+            error "Operator value($op) must be 'LT' or 'GT'"
+        }
+    }
+
+    # Performs a standardized defrag test.  The first code block generates data.  The second code block
+    # fragments that data (usually be deleting half).
```
zuiderkwast:
```suggestion
    # fragments that data (usually by deleting half).
```

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
+        r save ;# saving an rdb iterates over all the data / pointers
+    }
+
+    proc test_main_dictionary {type} {
+        set title "Active Defrag main dictionary: $type"
+        test $title {
+            set n 200000
+
+            perform_defrag_test $title {
+                # add a mass of string keys
```
zuiderkwast: Reading this, it's not immediately obvious that this first code block is just to populate data. If there would be a block with a `populate` keyword, it would be more obvious for the readers.

What do you think about a syntax like this?

```TCL
perform_defrag_test $title populate {
    ...
} fragment {
    ...
} while_defragging {
    ...
}
```

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +287,323 @@ run_solo {defrag} {
...
+                for {set j 0} {$j < $n} {incr j} {
+                    $rd read ; # Discard script load replies
+                    $rd read ; # Discard set replies
+                }
+            } fragment {
+                # Delete all the keys to create fragmentation
+                for {set j 0} {$j < $n} {incr j} { $rd del k$j }
+                for {set j 0} {$j < $n} {incr j} { $rd read } ; # Discard del replies
+                $rd close
+            } latency 100 ;# scripts aren't defragged incrementally, expect big latency
```
zuiderkwast: Put this latency arg after the title but before all the code blocks? It'd be easier to spot it.

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +287,323 @@ run_solo {defrag} {
...
-            set n 50000
-            set dummy_channel "[string repeat x 400]"
             set rd [valkey_deferring_client]
-            set rd_pubsub [valkey_deferring_client]
+            set chan [string repeat A 50]
+
+            # https://github.com/valkey-io/valkey/issues/1774
+            # NOTE - pubsub defrag isn't working properly.  This wasn't caught before the test refactor.
+            #  This commented out code should work.  Instead, is substituted code which verifies that
+            #  defrag doesn't break the pubsub, but defrag efficacy is not verified.
```
zuiderkwast: These lines are too long IMHO. Even if we allow longer lines of code, it's easier to read comments that are not that wide (especially on a phone, or laptop with e.g. side-by-side editor and terminal). Are you using emacs and automatic reflow using M-q (fill-paragraph)? I suggest setting fill-column to 80.

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +287,323 @@ run_solo {defrag} {
...
+            #     for {set j 0} {$j < $n} {incr j 2} {
+            #         set channel $chan$j
+            #         $rd unsubscribe $channel
+            #         $rd read
+            #         $rd sunsubscribe $channel
+            #         $rd read
+            #     }
+            # }
+            # --- BEGIN TEMPORARY CODE ---
+            log_frag "test start (pubsub)"
```
zuiderkwast, line 518: Will you finish this?

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +287,323 @@ run_solo {defrag} {
...
-            set expected_frag 1.7
-            # add a mass of list nodes to two lists (allocations are interlaced)
-            set val [string repeat A 100] ;# 5 items of 100 bytes puts us in the 640 bytes bin, which has 32 regs, so high potential for fragmentation
-            set elements 500000
-            for {set j 0} {$j < $elements} {incr j} {
-                $rd lpush biglist1 $val
-                $rd lpush biglist2 $val
+            log_frag "after 30 seconds of defrag"
+            after 120
+            # --- END TEMPORARY CODE ---
```
zuiderkwast, line 541: Notice

zuiderkwast (commented):
Very nice! I didn't follow what all the tests are doing, so I'll just have to trust that they're testing what they're supposed to test. I just added some minor comments about naming and such.

We can merge it soon and see if it solves the problems. There seems to be some temporary and commented-out code though that you may want to clean up.

tests/unit/memefficiency.tcl
JimB123:
```suggestion
```

tests/unit/memefficiency.tcl
JimB123: I kind-of consider the `run_solo` block to be the top level block for these tests.  In general, I like to limit things to the smallest possible scope.  So unless we're moving this to some general place, I think it's located ok.  But I guess I don't feel strongly.  If we move this though, maybe all of the other functions should move also?  I'd rather just leave it.

tests/unit/memefficiency.tcl
JimB123: In the old code, we were testing 1 big hash.  However we have a different data structure for a small hash (which was untested).

The new code creates hashes (and sets, and sorted sets, and lists) in exponentially increasing size.  Fragments them all, and then defrags.

As stated in the comment, 
Hash 0 will contain 2 fields/values.  Hash 1 will contain 4, Hash 2 will contain 8...  Same with sets, lists, etc.

This function just helps mathematically by finding the next key/field numbered pair.  It's a little mathematically simpler than passing in "n", but could be done that way also.  (`k` would be the highest value where `2^(k+1)` is `<=n` and `f` would be `n - 2^(k+1)`)

It generates a sequence like this
(0 0), (0 1),
(1 0), (1 1), (1 2), (1 3), 
(2 0), (2 1), (2 2), (2 3), (2 4), (2 5), (2 6), (2 7), 
...

I'm open to naming suggestions, but it seems specific enough that I couldn't think of a good simple name.  I figured it was better to just try to comment it and it would be obvious by usage.

Mostly this just helps make the loops simpler we don't need a nested loop:  for key =... { for val = ...

tests/support/test.tcl
JimB123: Well, obviously we've never had a use case for this before.  And debugging is the only use case I can think of for this.

We're waiting for something to become true.  Usually we just wait - but if we're debugging, we might want to collect some information while we're waiting.

It could be called `before_retry` (or maybe `before_delay_before_retry` ! Doh!)  Or `after_false`?

But none of those names sound much better than `debug` to me.  If you feel strongly about one, let me know.

tests/unit/memefficiency.tcl
JimB123:
```suggestion
            # NOTE - pubsub defrag isn't working properly.  This wasn't caught before the
            #  test refactor.  This commented out code should work.  Instead, is substituted
            # code which verifies that defrag doesn't break the pubsub, but defrag efficacy
            # is not verified.
```

tests/unit/memefficiency.tcl
JimB123: No.  I created: https://github.com/valkey-io/valkey/issues/1774
I redesigned all of the defrag structure.  I rewrote all of the tests.  If I took responsibility for everything I found that doesn't work properly, I'd never stop.  I'll let some pubsub expert deep-dive to determine why it's not defragging.

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,247 @@ start_server {tags {"memefficiency external:skip"}} {
...
+    #    fragment {code} - required, fragments the populated data
+    #    while_defragging {code} - optional, code executed after defrag has started
+    #    latency <ms> - optional, verifies the latency to a ms target (default 5)
+    proc perform_defrag_test {name pop_code args} {
+        set opts(latency) 5
+        set opts(while_defragging) {}
+        array set opts $args
+        assert {[info exists opts(fragment)]}
+
+        r config set active-defrag-threshold-lower 5
```
JimB123:
```suggestion
    # Performs a standardized defrag test.  The "populate" block generates data.  The "fragment" block
    # fragments that data (usually by deleting half).
    #  - caller must generate at least 40 MB of data
    #  - after fragmentation, at least 20 MB of data must remain, and fragmentation ratio must exceed 1.4
    # Positional parameters:
    #    name - name of the test
    # Named parameters:
    #    populate {code} - required, populates initial unfragmented data
    #    fragment {code} - required, fragments the populated data
    #    while_defragging {code} - optional, code executed after defrag has started
    #    latency <ms> - optional, verifies the latency to a ms target (default 5)
    proc perform_defrag_test {name args} {
        set opts(latency) 5
        set opts(while_defragging) {}
        array set opts $args
        assert {[info exists opts(populate)]}
        assert {[info exists opts(fragment)]}

        r config set active-defrag-threshold-lower 5
```

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,247 @@ start_server {tags {"memefficiency external:skip"}} {
...
+
+        log_frag "test start ($name)"
+
+        # at this point, we expect fragmentation BYTES to be below a threshold
+        #  (percentage is wildly variable with low memory usage)
+        assert {[s allocator_frag_bytes] < 2 * $::one_mb}
+
+        # populate the DB with data
+        set initial_allocated [s allocator_allocated]
+        uplevel 1 $pop_code
```
JimB123:
```suggestion
        uplevel 1 $opts(populate)
```

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,247 @@ start_server {tags {"memefficiency external:skip"}} {
...
+        assert {$digest eq $newdigest}
+        r save ;# saving an rdb iterates over all the data / pointers
+    }
+
+    proc test_main_dictionary {type} {
+        set title "Active Defrag main dictionary: $type"
+        test $title {
+            set n 200000
+
+            perform_defrag_test $title {
```
JimB123:
```suggestion
            perform_defrag_test $title populate {
```

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +286,323 @@ run_solo {defrag} {
...
-                } else {
-                    after 120 ;# serverCron only updates the info once in 100ms
-                    puts [r info memory]
-                    puts [r info stats]
-                    puts [r memory malloc-stats]
-                    fail "defrag not started."
-                }
-
-                after 1000 ;# Give defrag time to work (might be multiple cycles)
+            perform_defrag_test $title {
```
JimB123:
```suggestion
            # scripts aren't defragged incrementally, expect big latency
            perform_defrag_test $title latency 100 populate {
```

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +286,323 @@ run_solo {defrag} {
...
-            for {set j 0} {$j < 10000} {incr j} {
-                $rd hset bighash $j [concat "asdfasdfasdf" $j]
-                $rd lpush biglist [concat "asdfasdfasdf" $j]
-                $rd zadd bigzset $j [concat "asdfasdfasdf" $j]
-                $rd sadd bigset [concat "asdfasdfasdf" $j]
-                $rd xadd bigstream * item 1 value a
-            }
-            for {set j 0} {$j < 50000} {incr j} {
-                $rd read ; # Discard replies
+            perform_defrag_test $title {
```
JimB123:
```suggestion
            perform_defrag_test $title populate {
```

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +286,323 @@ run_solo {defrag} {
...
-                # scale the hash to 1m fields in order to have a measurable the latency
-                for {set j 10000} {$j < 1000000} {incr j} {
-                    $rd hset bighash $j [concat "asdfasdfasdf" $j]
-                }
-                for {set j 10000} {$j < 1000000} {incr j} {
-                    $rd read ; # Discard replies
-                }
-                # creating that big hash, increased used_memory, so the relative frag goes down
-                set expected_frag 1.3
+            perform_defrag_test $title {
```
JimB123:
```suggestion
            perform_defrag_test $title populate {
```

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +286,323 @@ run_solo {defrag} {
...
+            # number of total fields.  sets are progressively increasing sizes.
+            set n 200000
 
-            # create some fragmentation
-            for {set j 0} {$j < 500000} {incr j 2} {
-                $rd del $j
-            }
-            for {set j 0} {$j < 500000} {incr j 2} {
-                $rd read ; # Discard replies
+            perform_defrag_test $title {
```
JimB123:
```suggestion
            perform_defrag_test $title populate {
```

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +286,323 @@ run_solo {defrag} {
...
-                    fail "defrag not started."
-                }
+    proc test_big_zset {type} {
+        set title "Active Defrag big zset: $type"
+        test $title {
+            # number of total fields.  zsets are progressively increasing sizes.
+            set n 200000
 
-                after 1000 ;# Give defrag some time to work (it may run several cycles)
+            perform_defrag_test $title {
```
JimB123:
```suggestion
            perform_defrag_test $title populate {
```

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +286,323 @@ run_solo {defrag} {
...
-                if {$::verbose} {
-                    puts "frag $frag"
-                    set misses [s active_defrag_misses]
-                    set hits [s active_defrag_hits]
-                    puts "hits: $hits"
-                    puts "misses: $misses"
-                    puts "max latency $max_latency"
-                    puts [r latency latest]
-                    puts [r latency history active-defrag-cycle]
+            perform_defrag_test $title {
```
JimB123:
```suggestion
            perform_defrag_test $title populate {
```

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +286,323 @@ run_solo {defrag} {
...
+                for {set j 0} {$j < $n} {incr j} {
+                    $rd read ; # Discard script load replies
+                    $rd read ; # Discard set replies
+                }
+            } fragment {
+                # Delete all the keys to create fragmentation
+                for {set j 0} {$j < $n} {incr j} { $rd del k$j }
+                for {set j 0} {$j < $n} {incr j} { $rd read } ; # Discard del replies
+                $rd close
+            } latency 100 ;# scripts aren't defragged incrementally, expect big latency
```
JimB123:
```suggestion
            }
```

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
+            # If allocated bytes is too small, the ratios get wonky.  Since we use 2MB for
+            # active-defrag-ignore-bytes, let's make sure that we have at least 10x that amount
+            # allocated before trying to verify any fragmentation ratios.  Otherwise the tests
+            # are likely to get flaky.
+            error "test error: trying to validate frag ratio with only $allocated_bytes allocated"
+        }
+        if {$op == "LT"} {
+            assert_lessthan $frag $value
+        } elseif {$op == "GT"} {
+            assert_morethan $frag $value
```
JimB123:
```suggestion
    # Validate expected fragmentation ratio
    # op = "<" (less than) or ">" (greater than)
    proc validate_frag_ratio {op value} {
        set frag [s allocator_frag_ratio]
        if {$::verbose} {
            puts "Validating frag ($frag) $op $value"
        }
        set allocated_bytes [s allocator_allocated]
        if {$allocated_bytes < 20 * 1024 * 1024} {
            # If allocated bytes is too small, the ratios get wonky.  Since we use 2MB for
            # active-defrag-ignore-bytes, let's make sure that we have at least 10x that amount
            # allocated before trying to verify any fragmentation ratios.  Otherwise the tests
            # are likely to get flaky.
            error "test error: trying to validate frag ratio with only $allocated_bytes allocated"
        }
        if {$op == "<"} {
            assert_lessthan $frag $value
        } elseif {$op == ">"} {
            assert_morethan $frag $value
```

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
+        } else {
+            error "Operator value($op) must be 'LT' or 'GT'"
```
JimB123:
```suggestion
            error "Operator value($op) must be '<' or '>'"
```

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
+
+        log_frag "after adding data"
+        # check that enough data has been populated
+        set required [expr 40 * $::one_mb]
+        set allocated [expr [s allocator_allocated] - $initial_allocated]
+        if {$allocated < $required} {
+            fail "Tests are required to create at least $required bytes of data before fragmentation - only $allocated bytes created"
+        }
+        # check that at this point, fragmentation is very low
+        validate_frag_ratio LT 1.05
```
JimB123:
```suggestion
        validate_frag_ratio < 1.05
```

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
+
+        # fragment the data
+        uplevel 1 $opts(fragment)
+
+        log_frag "after fragmenting data"
+        # we want some fragmentation, but we still want a minimum allocation
+        validate_frag_ratio GT 1.4
```
JimB123:
```suggestion
        validate_frag_ratio > 1.4
```

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
-                foreach event [r latency latest] {
-                    lassign $event eventname time latency max
-                    if {$eventname == "active-defrag-cycle"} {
-                        set max_latency $max
+        set digest [debug_digest]
+
+        perform_defrag $opts(while_defragging)
+
+        log_frag "after defragging"
+        validate_frag_ratio LT 1.1
```
JimB123:
```suggestion
        validate_frag_ratio < 1.1
```

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +287,325 @@ run_solo {defrag} {
...
-                    puts [r latency history "while-blocked-cron"]
-                }
-                # make sure we had defrag hits during AOF loading
-                assert {$hits > 100000}
+                log_frag "after AOF loading"
                 # make sure the defragger did enough work to keep the fragmentation low during loading.
-                # we cannot check that it went all the way down, since we don't wait for full defrag cycle to complete.
-                assert {$frag < 1.4}
+                # we cannot check that it went all the way down, as we don't continue to defrag after loading complete.
+                validate_frag_ratio LT 1.4
```
JimB123:
```suggestion
                validate_frag_ratio < 1.4
```

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
+        r config set active-defrag-cycle-min 40
+        r config set active-defrag-cycle-max 60
+        r config set active-defrag-ignore-bytes 2mb
+        r config set maxmemory 0
+
+        log_frag "test start ($name)"
+
+        # at this point, we expect fragmentation BYTES to be below a threshold
+        #  (percentage is wildly variable with low memory usage)
+        assert {[s allocator_frag_bytes] < 2 * $::one_mb}
```
JimB123:
```suggestion
        assert {[s allocator_frag_bytes] < 2 * 1024 * 1024}
```

tests/unit/memefficiency.tcl
```diff
@@ -36,106 +36,248 @@ start_server {tags {"memefficiency external:skip"}} {
...
+        # check that at this point, fragmentation is very low
+        validate_frag_ratio < 1.05
+
+        # fragment the data
+        uplevel 1 $opts(fragment)
+
+        log_frag "after fragmenting data"
+        # we want some fragmentation, but we still want a minimum allocation
+        validate_frag_ratio > 1.4
+        set required [expr 20 * $::one_mb]
```
JimB123:
```suggestion
        set required [expr 40 * 1024 * 1024]
        set allocated [expr [s allocator_allocated] - $initial_allocated]
        if {$allocated < $required} {
            fail "Tests are required to create at least $required bytes of data before fragmentation - only $allocated bytes created"
        }
        # check that at this point, fragmentation is very low
        validate_frag_ratio < 1.05

        # fragment the data
        uplevel 1 $opts(fragment)

        log_frag "after fragmenting data"
        # we want some fragmentation, but we still want a minimum allocation
        validate_frag_ratio > 1.4
        set required [expr 20 * 1024 * 1024]
```

tests/unit/memefficiency.tcl
zuiderkwast: Limiting things to a small scope is usually a good idea. Less nesting is usually also a good idea. :)

The nesting gives an illusion that these functions are within a namespace or something, but they aren't. The `run_solo` block doesn't define a local scope. The functions defined here are still global. The only effect of defining them inside `run_solo` is that their definition is deferred until `run_solo` does `eval` on it's block. I don't think this is a good feature of TCL.

If we ever want to split this into two separate `run_solo` blocks or something, or skip `run_solo` then it's not a benefit that they're defined inside one of them.

tests/unit/memefficiency.tcl
zuiderkwast: Your reply here is better than the code comment. I can understand this. :)

> As stated in the comment,
> Hash 0 will contain 2 fields/values. Hash 1 will contain 4, Hash 2 will contain 8... Same with sets, lists, etc.

Yes, but the comment doesn't say that the function returns just one field-value pair as a list.

Your reply here says it: "This function just helps mathematically by finding the next key/field numbered pair". That's a missing piece.

> This function just helps mathematically by finding the next key/field numbered pair. It's a little mathematically simpler than passing in "n", but could be done that way also. (k would be the highest power of 2^k that's less than n and f would be n - 2^k)
>
> It generates a sequence like this
(0 0), (0 1),
(1 0), (1 1), (1 2), (1 3),
(2 0), (2 1), (2 2), (2 3), (2 4), (2 5), (2 6), (2 7),

This is good. I still don't fully understand what you do pass in to the function. You just say what you don't pass in: *simpler than passing in "n"*. :smile:

So, do I understand correctly:

* `key` is the row, the sequence number, the first element in the returned two-element list
* `field` is column, the index within a sequence, the second element in the returned two-element list

Is this right? No, wait, it increments key sometimes. Honestly I don't understand what this function really does return. It's just very confusing.

zuiderkwast: Eureka, it's just a single sequence! And you pass in an element in the sequence and it gives you the next one!

Now I got it. I needed to understood that it's just one sequence, not one per key.

I think this example can help:

```suggestion
    #
    # Given an pair (key, field), returns the next pair in the following
    # sequence:
    #
    # (0 0), (0 1),
    # (1 0), (1 1), (1 2), (1 3),
    # (2 0), (2 1), (2 2), (2 3), (2 4), (2 5), (2 6), (2 7),
    proc next_exp_kf {key field {incby 1}} {
```

tests/support/test.tcl
zuiderkwast: No, I don't feel strongly. I just wanted a good motivation for picking `debug`. Debug is fine.

But please add some sentence in comment above to document it.

tests/unit/memefficiency.tcl
zuiderkwast: Ok, makes sense. Thanks for opening the issue.

Shall we link to this issue in the comment above then? We usually don't refer to issues in code comments, but this is a TODO that we want to track. The issue can provide more information and discussion. If someone fixes it, we want to close the issue.

tests/unit/memefficiency.tcl
JimB123: ok.  I agree.  I was mostly inclined to keep the functions in the run_solo block because that's the way it was done before my change.

zuiderkwast: Oh, in that case I shouldn't have commented on it in the first place. Sorry about that.

tests/unit/memefficiency.tcl
```diff
@@ -38,16 +38,19 @@ start_server {tags {"memefficiency external:skip"}} {
...
 
-    # used to create keys which increase in size by 2x
-    # key 0 will have 2 fields (0,1), key 1 will have 4 (0,1,2,3), key 2 will have 8 ...
-    # incby allows incrementing by 1 or 2 (but no other values will work properly)
+    # Utility function used to create a sequence where each key has an increasing
+    # number of fields/values.  For each key, the number for fields/values will
+    # increase by 2x.  Key 0 will have 2 values, key 1 will have 4, key 2 will
+    # have 8, etc.  Pass in the previous key/field pair to get the next.
+    # The sequence is: (0,0), (0,1), (1,0), (1,1), (1,2), (1,3), (2,0)...
+    # The optional 'incby' parameter allows incrementing by 2 across the sequence.
```
zuiderkwast, line 51: Great! This looks very comprehensive to me.

tests/unit/memefficiency.tcl
```diff
@@ -145,601 +290,326 @@ run_solo {defrag} {
...
-                for {set j 0} {$j < $keys} {incr j} {
-                    $rd setrange $j 600 x
-                }
-                for {set j 0} {$j < $keys} {incr j} {
-                    $rd read ; # Discard replies
+            if {$type == "cluster"} {
+                start_cluster 1 0 [list tags $cluster_tags overrides $overrides] {
+                    # Note: `start_cluster` passes the code through to another level which requires us
+                    #  to do an uplevel here.  Otherwise `test_proc` isn't recognized.
+                    uplevel 1 {$test_proc $type}
```
zuiderkwast, line 606: This looks like a workaround for an error in `start_cluster`.

A correct solution could be to pass the number of uplevels upward. I mean `start_cluster` would pass the number 2 to `start_multiple_servers` which passes this on to `start_server`, so it can use `uplevel 2 $code`. I guess we could add an optional argument in the end, after the 'code' argument, in `start_multiple_servers` and `start_server`. Does it make sense?

zuiderkwast (approved):
I'm OK with merging this although I didn't take the time to understand the tests fully. Thanks!

tests/unit/memefficiency.tcl
JimB123: I don't plan to address existing design issues with `start_cluster` - that's outside the scope of defrag testing.  Not sure where that will lead.  It's disappointing that it doesn't work the same as `start_server`.

JimB123:
@zuiderkwast Viktor - BTW I tried to move functions outside of the `run_solo` block as you had suggested.  Interestingly, it worked fine locally.  Unfortunately it fails across many builds on github.  I've reverted it as it was - which is the same as it has historically been.  Example of failures: https://github.com/valkey-io/valkey/actions/runs/13572466450/job/37940705455?pr=1762

Also, I created a bug report for the HASH finding.  As of now, we can't run the HASH test.  @SoftlyRaining is looking at it.  Seems like maybe something in the new HashSet modifications is corrupting the HASH?  We repeatedly crash generating the digest after defrag.  https://github.com/valkey-io/valkey/issues/1795    For now, I'm commenting out the hash test.  This is reasonably equivalent to the old test as the hash structures weren't actually being defragged!

tests/unit/memefficiency.tcl
zuiderkwast: Your workaround relies on this wrong behavior though, making this fail if we fix start_cluster later... Nah.. Let's make it a follow-up.

I almost have a solution in mind so I might do it for breakfast one of these days.

rainsupreme (approved):
I like how the tests are organized and the coverage here! I don't know tcl well enough to know the syntactic minutia but otherwise this looks good to me :)

Issue #1758, opened 2025-02-20
TCP_NODELAY missing in outgoing connections


sungming2:
This is a follow up issue from https://github.com/valkey-io/valkey/pull/1706#issuecomment-2669567875

The cluster bus initiates connections for node-to-node communication (e.g., Pub/Sub, PING/PONG..).
The absence of TCP_NODELAY can cause unnecessary latency due to [Nagle’s algorithm](https://en.wikipedia.org/wiki/Nagle%27s_algorithm) (buffers small TCP packets), which should be disabled for traffic.

We need to consider enabling TCP_NODELAY by default

xbasel:
I'm not sure TCP_NODELAY should be enabled for replication. It could increase bandwidth and CPU usage. Did you run any performance tests to evaluate the impact?

PR #1763, opened 2025-02-21
Enable TCP_NODELAY by default in incoming and outgoing connections

sungming2:
### Issue: https://github.com/valkey-io/valkey/issues/1758
This is a follow up issue from https://github.com/valkey-io/valkey/pull/1706#issuecomment-2669567875

### Problem: 
The absence of TCP_NODELAY can cause unnecessary latency due to [Nagle’s algorithm](https://en.wikipedia.org/wiki/Nagle%27s_algorithm) (buffers small TCP packets), which could be disabled for traffic.
We need to consider enabling TCP_NODELAY by default in both incoming and outgoing connections

@pizhenwei mentioned
* modern computers have high bandwidth networking interface, a few small packets will not affect the bandwidth any more.
* a lot of instances run on cloud virtual machine. Intel VT-x and AMD-v uses software emulated LAPIC to support TSC deadline timer, it uses ~10x CPU cycles (~3000 cycle VS ~300 cycles on physical machine) to setup CPU timer during TCP Nagle's algorithm. Disabling it will reduce CPU cycles for VM and cloud virtual machine.

### DIFF
--- a/src/anet.c
+++ b/src/anet.c
@@ -496,6 +496,9 @@ static int anetTcpGenericConnect(char *err, const char *addr, int port, const ch
             continue;
         }
 
+        /* Enable TCP_NODELAY by default */
+        anetEnableTcpNoDelay(NULL, s);
+
         /* If we ended an iteration of the for loop without errors, we
          * have a connected socket. Let's return to the caller. */
         goto end;
@@ -684,6 +687,10 @@ int anetTcpAccept(char *err, int serversock, char *ip, size_t ip_len, int *port)
         if (ip) inet_ntop(AF_INET6, (void *)&(s->sin6_addr), ip, ip_len);
         if (port) *port = ntohs(s->sin6_port);
     }
+
+    /* Enable TCP_NODELAY by default */
+    anetEnableTcpNoDelay(NULL, fd);
+
     return fd;
 }
 
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -357,7 +357,6 @@ migrateCachedSocket *migrateGetSocket(client *c, robj *host, robj *port, long ti
         sdsfree(name);
         return NULL;
     }
-    connEnableTcpNoDelay(conn);
 
     /* Add to the cache and return it to the caller. */
     cs = zmalloc(sizeof(*cs));
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -1493,7 +1493,7 @@ void clusterAcceptHandler(aeEventLoop *el, int fd, void *privdata, int mask) {
             connClose(conn);
             return;
         }
-        connEnableTcpNoDelay(conn);
+
         connKeepAlive(conn, server.cluster_node_timeout / 1000 * 2);
 
         /* Use non-blocking I/O for cluster messages. */
--- a/src/connection.h
+++ b/src/connection.h
@@ -380,8 +380,6 @@ static inline const char *connGetInfo(connection *conn, char *buf, size_t buf_le
 /* anet-style wrappers to conns */
 int connBlock(connection *conn);
 int connNonBlock(connection *conn);
-int connEnableTcpNoDelay(connection *conn);
-int connDisableTcpNoDelay(connection *conn);
 int connKeepAlive(connection *conn, int interval);
 int connSendTimeout(connection *conn, long long ms);
 int connRecvTimeout(connection *conn, long long ms);
--- a/src/replication.c
+++ b/src/replication.c
@@ -1152,7 +1152,7 @@ void syncCommand(client *c) {
     /* Setup the replica as one waiting for BGSAVE to start. The following code
      * paths will change the state if we handle the replica differently. */
     c->repl_data->repl_state = REPLICA_STATE_WAIT_BGSAVE_START;
-    if (server.repl_disable_tcp_nodelay) connDisableTcpNoDelay(c->conn); /* Non critical if it fails. */
+    if (server.repl_disable_tcp_nodelay) anetDisableTcpNoDelay(NULL, c->conn); /* Non critical if it fails. */
     c->repl_data->repldbfd = -1;
     c->flag.replica = 1;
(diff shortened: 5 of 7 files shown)


xbasel:
I'm not sure TCP_NODELAY should be enabled for replication. It could increase bandwidth and CPU usage. Did you run any performance tests to evaluate the impact?

xbasel:
> I'm not sure TCP_NODELAY should be enabled for replication. It could increase bandwidth and CPU usage. Did you run any performance tests to evaluate the impact?

Actually, I was wrong. It looks like TCP_NODELAY is already enabled for replication, but I'm still not sure if it's the right thing to do.

sungming2:
Is this not the reason that we want to keep low latency on replication?
https://github.com/valkey-io/valkey/blob/df5caa637a1d4c833fb9fa994e8a7c7fad1ef022/valkey.conf#L808-L821

sungming2:
@ranshid @pizhenwei could you take a look at this PR please? thanks

pizhenwei (commented):
From the point of my view, auto enabling TCP NODELAY is reasonable:
* modern computers have high bandwidth networking interface, a few small packets will not affect the bandwidth any more.
* a lot of instances run on cloud virtual machine. Intel VT-x and AMD-v uses software emulated LAPIC to support TSC deadline timer, it uses ~10x CPU cycles (~3000 cycle VS ~300 cycles on physical machine) to setup CPU timer during TCP Nagle's algorithm. Disabling it will reduce CPU cycles for VM and cloud virtual machine.
* there are more than 2500 lines in valkey.conf, what about removing several unnecessary options?

By the way, `connEnableTcpNoDelay` could be remove together. TCP NODELAY is TCP specific option only, it should works inside TCP connection type, no need to expose.

ranshid:
>From the point of my view, auto enabling TCP NODELAY is reasonable

I also agree on that. In most cases I have seen enabling NODELAY only helped mitigating issues related to tail latency. I do recall (However), that we had some claims that a better performance was observed on some cloud infrastructures when Nagle's algorithm was enabled. I am fine with making this change. In general we do not like placing configurations and prefer having the correct default, so I think we can maybe consider if we would like to place some Internal control configuration for this default.

>By the way, connEnableTcpNoDelay could be remove together. TCP NODELAY is TCP specific option only, it should works inside TCP connection type, no need to expose.

agree it is kinda dead code now, but maybe we can can keep it for symmetry and potential future usecases?

pizhenwei:
> > From the point of my view, auto enabling TCP NODELAY is reasonable
> 
> I also agree on that. In most cases I have seen enabling NODELAY only helped mitigating issues related to tail latency. I do recall (However), that we had some claims that a better performance was observed on some cloud infrastructures when Nagle's algorithm was enabled. I am fine with making this change. In general we do not like placing configurations and prefer having the correct default, so I think we can maybe consider if we would like to place some Internal control configuration for this default.
> 
> > By the way, connEnableTcpNoDelay could be remove together. TCP NODELAY is TCP specific option only, it should works inside TCP connection type, no need to expose.
> 
> agree it is kinda dead code now, but maybe we can can keep it for symmetry and potential future usecases?

@zuiderkwast @hpatro Do you want to share any suggestion?

ranshid:
> I think we can maybe consider if we would like to place some Internal control configuration for this default.

Just to clarify I meant adding a configuration in the future if and when we find the need

zuiderkwast (approved):
I'm OK with this change in its current form. It's small and simple. I'm not familiar with various cloud infrastructure behavior, but the explanation by @pizhenwei sounds reasonable.

No new config for this please.

We can possibly delete or deprecate some configs but that should be done in a major version.

If the code is dead, then we can delete it, but it's not dead code as long as that config exists, right?

The PR title and top comment should explain the user-visible changes, i.e. which kind of connections are affected by this change. Outgoing cluster bus connections only?

ranshid (approved):
Overall LGTM. Since there is a slight behavioral change lets tag this in the release-notes.

hpatro (approved):
Quickly checked there is no further usage of `connEnableTcpNoDelay` and doesn't have much significance to provide the abstraction at connection level. Let's clean that up.

hpatro:
~~I believe there is no behavior change with this PR, it's just refactoring the code to enable nodelay always for TCP/TLS connection accept method itself.~~

sungming2:
> I believe there is no behavior change with this PR, it's just refactoring the code to enable nodelay always for TCP/TLS connection accept method itself.

doesn't this PR also enable nodelay for outgoing connections, which we previously didn't?

hpatro:
> From the point of my view, auto enabling TCP NODELAY is reasonable:
> 
> * modern computers have high bandwidth networking interface, a few small packets will not affect the bandwidth any more.
> * a lot of instances run on cloud virtual machine. Intel VT-x and AMD-v uses software emulated LAPIC to support TSC deadline timer, it uses ~10x CPU cycles (~3000 cycle VS ~300 cycles on physical machine) to setup CPU timer during TCP Nagle's algorithm. Disabling it will reduce CPU cycles for VM and cloud virtual machine.

Makes lot of sense to enable it by default. 

> By the way, `connEnableTcpNoDelay` could be remove together. TCP NODELAY is TCP specific option only, it should works inside TCP connection type, no need to expose.

Did a cleanup as part of this PR itself.

> * there are more than 2500 lines in valkey.conf, what about removing several unnecessary options?

Removing the configs completely would be difficult for users migrating from an older version to newer version. I think we could deprecate some of it. Good item for 9.0.

hpatro:
> > I believe there is no behavior change with this PR, it's just refactoring the code to enable nodelay always for TCP/TLS connection accept method itself.
> 
> doesn't this PR also enable nodelay for outgoing connections, which we previously didn't?

My bad, I see it was done only for migrate command outgoing connection.

hpatro:
Thanks @sungming2 for the PR. Will close it out after the build succeeds.

PR #1765, opened 2025-02-22
feat: add tcp keepalive probes for managed service Valkey by cloud service providers

proost:
I want to use valkey computing resource more efficiently. This is tiny improvement, but it can reduce unnecessary resource usage. Because it can remove dead connections more earlier. 

Even there is lots of transient error in networks can get benefit from it.

out-of-scope:

I'm currently using Amazon ElastiCache for Valkey. After this PR is merged, I wish many CSPs provide "tcp-keepalive-probes" in parameter groups including AWS.

### DIFF
--- a/src/anet.c
+++ b/src/anet.c
@@ -151,7 +151,7 @@ int anetCloexec(int fd) {
 
 /* Enable TCP keep-alive mechanism to detect dead peers,
  * TCP_KEEPIDLE, TCP_KEEPINTVL and TCP_KEEPCNT will be set accordingly. */
-int anetKeepAlive(char *err, int fd, int interval, const int probes) {
+int anetKeepAlive(char *err, int fd, int interval, int probes) {
     int enabled = 1;
     if (setsockopt(fd, SOL_SOCKET, SO_KEEPALIVE, &enabled, sizeof(enabled))) {
         anetSetError(err, "setsockopt SO_KEEPALIVE: %s", strerror(errno));
--- a/src/anet.c
+++ b/src/anet.c
@@ -151,7 +151,7 @@ int anetCloexec(int fd) {
 
 /* Enable TCP keep-alive mechanism to detect dead peers,
  * TCP_KEEPIDLE, TCP_KEEPINTVL and TCP_KEEPCNT will be set accordingly. */
-int anetKeepAlive(char *err, int fd, int interval, const int *probes) {
+int anetKeepAlive(char *err, int fd, int interval, const int probes) {
     int enabled = 1;
     if (setsockopt(fd, SOL_SOCKET, SO_KEEPALIVE, &enabled, sizeof(enabled))) {
         anetSetError(err, "setsockopt SO_KEEPALIVE: %s", strerror(errno));
@@ -160,7 +160,6 @@ int anetKeepAlive(char *err, int fd, int interval, const int *probes) {
 
     int idle;
     int intvl;
-    int cnt;
 
     /* There are platforms that are expected to support the full mechanism of TCP keep-alive,
      * we want the compiler to emit warnings of unused variables if the preprocessor directives
@@ -171,7 +170,7 @@ int anetKeepAlive(char *err, int fd, int interval, const int *probes) {
     UNUSED(interval);
     UNUSED(idle);
     UNUSED(intvl);
-    UNUSED(cnt);
+    UNUSED(probes);
 #endif
 
 #ifdef __sun
@@ -216,9 +215,6 @@ int anetKeepAlive(char *err, int fd, int interval, const int *probes) {
         anetSetError(err, "setsockopt TCP_KEEPINTVL: %s\n", strerror(errno));
         return ANET_ERR;
     }
-
-    cnt = 3;
-    if (probes != NULL) cnt = *probes;
 
     if (setsockopt(fd, IPPROTO_TCP, TCP_KEEPCNT, &probes, sizeof(probes))) {
         anetSetError(err, "setsockopt TCP_KEEPCNT: %s\n", strerror(errno));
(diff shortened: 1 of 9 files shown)


src/socket.c
```diff
@@ -475,9 +475,9 @@ int connDisableTcpNoDelay(connection *conn) {
     return anetDisableTcpNoDelay(NULL, conn->fd);
 }
 
-int connKeepAlive(connection *conn, int interval) {
+int connKeepAlive(connection *conn, int interval, const int *probes) {
```
xbasel: why `probes` is passed as a pointer and not by value?

proost: [d29f931](https://github.com/valkey-io/valkey/pull/1765/commits/d29f9316019dcbb45e12b6c2cbaabbe2dad60234)

OK, I Changed it. Use "CLI_KEEPALIVE_PROBES".

xbasel:
Not sure why this change is needed. The default of 3 is already quite aggressive, and reducing it further may cause premature disconnects. Exposing it as a config makes sense, but what's the specific use case for this change?

proost:
@xbasel 

Currently Valkey cluster i used has many connections, So I want to try remove unnecessary connections more eagerly.

And there is plan to open service to another region which is physically far from current region and maybe more unstable env. Later, bigger probes may help to reduce high churning connections. (Best option is making a new valkey cluster to close to user's region and stable env. But there is reason internally can't do)

src/anet.c
```diff
@@ -151,7 +151,7 @@ int anetCloexec(int fd) {
 
 /* Enable TCP keep-alive mechanism to detect dead peers,
  * TCP_KEEPIDLE, TCP_KEEPINTVL and TCP_KEEPCNT will be set accordingly. */
-int anetKeepAlive(char *err, int fd, int interval) {
+int anetKeepAlive(char *err, int fd, int interval, const int probes) {
```
xbasel: `const` isn’t commonly used for by-value parameters in Valkey. For consistency with the codebase, I'd remove it.

proost: OK.
[2fef5e8](https://github.com/valkey-io/valkey/pull/1765/commits/2fef5e8f4d3218c51b4082e40164ac2085f3f8e0)

valkey.conf
```diff
@@ -174,6 +174,16 @@ timeout 0
...
+# TCP keepalive probes.
+#
+# The number of TCP keepalive probes to send before considering the connection dead.
+# Only valid if tcp-keepalive is non-zero.
+# This is useful below situations:
+#
+# 1) Faster detection of dead peers.
+# 2) Avoiding premature disconnection due to network equipment in the middle.
+tcp-keepalive-probes 3
+
```
xbasel: IMO tt's a bit verbose. You can simplify it while keeping the essential details:
`# Number of TCP keepalive probes before closing a dead connection. `

proost: [e24f3f9](https://github.com/valkey-io/valkey/pull/1765/commits/e24f3f9715c22ef960a26f14e2fa22d86e6016b3)

pizhenwei:
> Not sure why this change is needed. The default of 3 is already quite aggressive, and reducing it further may cause premature disconnects.

Agree. The default 3 is quite common.

> Exposing it as a config makes sense, but what's the specific use case for this change?

Ahhh, we will have a ~3000 lines config file?

pizhenwei:
> Currently Valkey cluster i used has many connections, So I want to try remove unnecessary connections more eagerly.
> 
> And there is plan to open service to another region which is physically far from current region and maybe more unstable env. Later, bigger probes may help to reduce high churning connections. (Best option is making a new valkey cluster to close to user's region and stable env. But there is reason internally can't do)

I have concern on the real usage of valkey over an unstable networking, valkey is always consider as 'fast memory KV database', but it gets slow/unstable due to networking...

xbasel:
> > Not sure why this change is needed. The default of 3 is already quite aggressive, and reducing it further may cause premature disconnects.
> 
> Agree. The default 3 is quite common.
> 
> > Exposing it as a config makes sense, but what's the specific use case for this change?
> 
> Ahhh, we will have a ~3000 lines config file?


@pizhenwei There are already 80 different configs in the file, and most of it is actually documentation. The key question is whether this new config adds useful flexibility. What do you think?

sungming2:
> it can remove dead connections more earlier.

IMO, if we really want to define the connections lost earlier/later with config, I think we need to introduce  another config for `KEEPALIVE_INTERVAL` (I guess we don't have it in config) because it will send probes with the intervals
e.g.,
```
tcp_keepalive_intvl=5   # interval (seconds)
tcp_keepalive_probes=5  # probes (TCP_KEEPCNT)
```
will detect a dead connection in about 5*5 = (keepalive_time + 25 seconds).
but introducing more configs may increase complexity which we probably don't like.
Please correct me if I am wrong.

pizhenwei:
> > > Not sure why this change is needed. The default of 3 is already quite aggressive, and reducing it further may cause premature disconnects.
> > 
> > 
> > Agree. The default 3 is quite common.
> > > Exposing it as a config makes sense, but what's the specific use case for this change?
> > 
> > 
> > Ahhh, we will have a ~3000 lines config file?
> 
> @pizhenwei There are already 80 different configs in the file, and most of it is actually documentation. The key question is whether this new config adds useful flexibility. What do you think?

I don't know which scenario valkey will get real benefit from this change. So I prefer the default probe rather than  increasing complexity.

proost:
@pizhenwei 
then is exposing keepalive internval to config ok?

sungming2:
I think we prefer not to introduce configs if I read the context correctly

proost:
there are lots of opposite opinions. So Close it and I will find another ways.

PR #1768, opened 2025-02-24
Fixed issue with `CONFIG RESETSTAT` in cluster module message callback test

### LINKED ISSUE #1764: [test-failure] Cluster module message callback test
```
*** [err]: Cluster module message DING/DONG acknowledgment in tests/unit/moduleapi/cluster.tcl
node 2 or node 3 didn't receive DING messages or node 1 didn't receive DONG message
```
This test seems to be failing: https://github.com/valkey-io/valkey/actions/runs/13447133659/job/37574838815.

_Originally posted by @madolson in https://github.com/valkey-io/valkey/pull/1708#discussion_r1966250292_

Nikhil-Manglore:
**Problem**
The cluster module test "Cluster module message DING/DONG acknowledgment" is failing despite successful message exchanges between nodes.  While these messages are being properly exchanged between the nodes, running multiple tests on the same cluster instance is causing issues with message statistics tracking. `CONFIG RESETSTAT` doesn't seem to reset the stats under certain cases which causes the error as the stats are adding onto each other from previous tests.

**Solution**
Remove some overlap with the previous test case so we don't need `CONFIG RESETSTAT` between the test cases.

Resolves #1764

### DIFF
--- a/tests/unit/moduleapi/cluster.tcl
+++ b/tests/unit/moduleapi/cluster.tcl
@@ -22,10 +22,12 @@ start_cluster 3 0 [list config_lines $modules] {
             fail "node 2 or node 3 didn't receive cluster module message"
         }
     }
+}
 
-    $node1 CONFIG RESETSTAT
-    $node2 CONFIG RESETSTAT
-    $node3 CONFIG RESETSTAT
+start_cluster 3 0 [list config_lines $modules] {
+    set node1 [srv 0 client]
+    set node2 [srv -1 client]
+    set node3 [srv -2 client]
 
     test "Cluster module message DING/DONG acknowledgment" {
         assert_equal OK [$node1 test.pingall]
--- a/tests/unit/moduleapi/cluster.tcl
+++ b/tests/unit/moduleapi/cluster.tcl
@@ -21,26 +21,16 @@ start_cluster 3 0 [list config_lines $modules] {
         } else {
             fail "node 2 or node 3 didn't receive cluster module message"
         }
-    }
-}
-
-start_cluster 3 0 [list config_lines $modules] {
-    set node1 [srv 0 client]
-    set node2 [srv -1 client]
-    set node3 [srv -2 client]
-
-    test "Cluster module message DING/DONG acknowledgment" {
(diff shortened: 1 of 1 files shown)


enjoy-binbin (approved):

zuiderkwast (commented):
It great that you found it! I was suspecting the RESETSTAT too but I wasn't sure why.

This fix solves the problem so it's good.

When I look at these two test cases together, I see that the first one tests VM_SendClusterMessage and the second one tests VM_RegisterClusterMessageReceiver. But the second one also repeats what the first one tests, so there is redundancy here.

We could remove some redundant messages and checks here if we combine them and let the first test check the DING message and the second one only check for the DONG message.

Something like this. What do you think?

```TCL
    test "Cluster module send message API - VM_SendClusterMessage" {
        assert_equal OK [$node1 test.pingall]
        assert_equal 2 [CI 0 cluster_stats_messages_module_sent]
        wait_for_condition 50 100 {
            [CI 1 cluster_stats_messages_module_received] eq 1 &&
            [CI 2 cluster_stats_messages_module_received] eq 1
        } else {
            fail "node 2 or node 3 didn't receive cluster module message"
        }
        verify_log_message -1 "*DING (type 1) RECEIVED*Hey*" 0
        verify_log_message -2 "*DING (type 1) RECEIVED*Hey*" 0
    }

    test "Cluster module receive message API - VM_RegisterClusterMessageReceiver" {
        wait_for_condition 50 100 {
            [CI 0 cluster_stats_messages_module_received] eq 2
        } else {
            fail "node 1 didn't receive DONG messages"
        }
        assert_equal 2 [count_log_message 0 "* <cluster> DONG (type 2) RECEIVED*"]
    }
```

The only drawback is that you can't skip the first and run only the second tests using `--only` but that is broken in very many test suites anyway.

Nikhil-Manglore:
Yes I think that's a great idea! I had noticed the redundancy but had proceeded to write my own test case. I like your way a lot better though. Just pushed the updated code.

zuiderkwast (approved):

PR #1776, opened 2025-02-25
Add CLIENT INFO filter

sungming2:
### Problem 
The CLIENT INFO command currently returns all client details, which can be excessive. This proposal suggests adding an optional parameter to request specific fields, reducing unnecessary output and improving usability.

CLIENT INFO output:
```
id=5 addr=127.0.0.1:33448 laddr=127.0.0.1:6379 fd=11 name= age=118 idle=0 flags=N db=8 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=10 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=1562 events=r cmd=client|info user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=131 tot-net-out=214470 tot-cmds=4
```

### Proposal

Enhance CLIENT INFO command by adding an optional parameter to specify the fields users want to retrieve.
```
CLIENT INFO [field1] [field2] [...]
```

Example
----
#### 1. Default behavior (no parameters) remains unchanged:
```
client info
```
Output:
```
id=2 addr=127.0.0.1:12345 laddr=127.0.0.1:6379 fd=8 name= age=3000 idle=10 ...
```
----
#### 2. Retrieve specific fields:
```
client info id addr
```
Output:
```
id=2 addr=127.0.0.1:12345
```
----
#### 3. Retrieve a single field (No parsing required):
```
client info id
```
Output:
```
2
```

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -3340,98 +3340,120 @@ int isClientConnIpV6(client *c) {
     return getClientPeerId(c)[0] == '[';
 }
 
+sds getClientInfoAttribute(client *client, const char *attr, int hide_user_data) {
+    char conninfo[CONN_INFO_LEN];
+    if (!strcasecmp(attr, "id")) return sdscatfmt(sdsempty(), "%U", (unsigned long long)client->id);
+    if (!strcasecmp(attr, "addr")) return sdscat(sdsempty(), getClientPeerId(client));
+    if (!strcasecmp(attr, "laddr")) return sdscat(sdsempty(), getClientSockname(client));
+    if (!strcasecmp(attr, "fd")) return sdscat(sdsempty(), connGetInfo(client->conn, conninfo, sizeof(conninfo)));
+    if (!strcasecmp(attr, "name")) return sdscat(sdsempty(), hide_user_data ? "*redacted*" : (client->name ? (char *)client->name->ptr : ""));
+    if (!strcasecmp(attr, "age")) return sdscatfmt(sdsempty(), "%I", (long long)(commandTimeSnapshot() / 1000 - client->ctime));
+    if (!strcasecmp(attr, "idle")) return sdscatfmt(sdsempty(), "%I", (long long)(server.unixtime - client->last_interaction));
+    if (!strcasecmp(attr, "flags")) {
+        char flags[17], *p = flags;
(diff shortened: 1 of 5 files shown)


src/networking.c
```diff
@@ -3340,98 +3340,120 @@ int isClientConnIpV6(client *c) {
     return getClientPeerId(c)[0] == '[';
 }
 
+sds getClientInfoAttribute(client *client, const char *attr, int hide_user_data) {
```
sungming2, line 3343: I wanted to avoid these repetitive conditions but it seems C doesn't have a graceful way to handle it such as
```
attribute_map attribute_handlers[] = {
    {"id", () => { return client->id }},
    {"addr", () => { return getClientPeerId() }},
    {"laddr", () => { return getClientSockname() }},
    // Add other attributes...
};
```
I am open to getting any other suggestions if it doesn't look good

hpatro (commented):
Just discussed about this in our standup.

Some thought on this. 

* Do we really need it? 
* I found the regex in the tcl helper to do the job just fine.
* I think @madolson mentioned this, it might be more beneficial for `CLIENT LIST` rather  than `CLIENT INFO`. `CLIENT INFO` isn't that heavily used. 

Would be good to hear others perspective as well.

sungming2:
If users don’t actively use CLIENT INFO and if this filter is primarily for testing purposes, I’m fine with closing it.
What do you think? @xbasel

xbasel:
> If users don’t actively use CLIENT INFO and if this filter is primarily for testing purposes, I’m fine with closing it. What do you think? @xbasel

[sungming2](https://github.com/sungming2)
I'm not sure how many users actively use this command. I sometimes use CLIENT INFO (and CLIENT LIST) via valkey-cli and find it tedious to visually search for fields. I assume clients especially administrators who run commands interactively, might have the same experience. 
No strong opinions either way.

Worth checking whether this is beneficial for `CLIENT LIST`.

sungming2:
After offline discussion with @madolson, we probably don't need it right now so I will close it and get back to it when we get some demand from users

PR #325, opened 2024-04-16
replica redirect read&write to primary in standalone mode

soloestoy:
To implement #319 

1. replica is able to redirect read and write commands to it's primary in standalone mode
    * reply with "-REDIRECT primary-ip:port"
2. add a subcommand `CLIENT CAPA redirect`, a client can announce the capability to handle redirection
    * if a client can handle redirection, the data access commands (read and write) will be redirected
3. allow `readonly` and `readwrite` command in standalone mode, may be a breaking change
    * a client with redirect capability cannot process read commands on a replica by default
    * use READONLY command can allow read commands on a replica

PR #1781, opened 2025-02-25
standalone REDIRECT: Fix scripting and further MULTI/EXEC scenarios

### LINKED ISSUE #868: [BUG] "CLIENT CAPA redirect" has no effect in scripts
**Describe the bug**

A read or write command should be redirected when issued by a client that understands "REDIRECT". This does not work when the command is part of a script.

**To reproduce**

Connect the CLI client to a replica and  do:

```
localhost:6380> client capa redirect
OK
localhost:6380> set foo bar
(error) REDIRECT redis:6379
localhost:6380> EVAL "return redis.call('SET', KEYS[1], ARGV[1])" 1 foo bar
(error) READONLY You can't write against a read only replica. script: d8f2fad9f8e86a53d2a6ebd960b33c4972cacc37, on @user_script:1.
localhost:6380> get foo
(error) REDIRECT redis:6379
localhost:6380> EVAL "return redis.call('GET', KEYS[1])" 1 foo
"bar"
```

(Note: I did a `set foo bar` on the primary to test replication, that's why the last command returns "bar".)

**Expected behavior**

REDIRECT should be issued regardless of the context if the client understands redirect. (for reference, Valkey cluster issues the same "MOVED" in both cases:

```
127.0.0.1:30004> readwrite
OK
127.0.0.1:30004> set b b
(error) MOVED 3300 127.0.0.1:30001
127.0.0.1:30004> EVAL "return redis.call('SET', KEYS[1], ARGV[1])" 1 b b
(error) MOVED 3300 127.0.0.1:30001
```
)

gmbnomis:
In cluster mode, the necessity of a "MOVED" response is mainly determined based on the keys of a command whereas in standalone redirect mode, the necessity of a "REDIRECT" response is determined based on the command's flags.

It turns out that looking at keys is necessary. Moreover, the currently used command flags do not encompass all situations in which a REDIRECT is necessary:

- WATCH command (redirect in READWRITE mode is necessary)
- transactions with no keys must not be redirected
- scripting scenarios: scripts (with keys) must be redirected in some situations (issue #868).

The new logic is a heavily "distilled" version of `getNodeByQuery` used in cluster mode. As we don't need to look at slots and their consistency, the information necessary is the flags of the command and whether it accesses at least one key.

Fixes #868

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -4163,6 +4163,51 @@ void unprepareCommand(client *c) {
     c->slot = -1;
 }
 
+/* Check if the command will access at least one key (or has a "key-like"
+   argument like a sharded pub/sub command) */
+int commandHasKey(client *c, struct serverCommand *cmd, robj **argv, int argc) {
+    multiState *ms, _ms;
+    multiCmd mc;
+    int i;
+
+    if (cmd->proc == execCommand) {
+        /* If CLIENT_MULTI flag is not set EXEC is just going to return an
+         * error later. */
+        if (!c->flag.multi) return 0;
+        ms = c->mstate;
+    } else {
+        /* In order to have a single codepath create a fake Multi State
+         * structure if the client is not in MULTI/EXEC state, this way
+         * we have a single codepath below. */
+        ms = &_ms;
+        _ms.commands = &mc;
+        _ms.count = 1;
+        mc.argv = argv;
+        mc.argc = argc;
+        mc.cmd = cmd;
+    }
+
+    /* Check if the command accesses a key */
+    for (i = 0; i < ms->count; i++) {
+        struct serverCommand *mcmd;
+        robj **margv;
+        int margc, numkeys;
+
+        mcmd = ms->commands[i].cmd;
+        margc = ms->commands[i].argc;
+        margv = ms->commands[i].argv;
+
+        getKeysResult result;
+        initGetKeysResult(&result);
+        numkeys = getKeysFromCommand(mcmd, margv, margc, &result);
+        getKeysFreeResult(&result);
+        if (numkeys > 0) {
+            return 1;
+        }
+    }
+    return 0;
+}
+
 /* If this function gets called we already read a whole
  * command, arguments are in the client argv/argc fields.
  * processCommand() execute the command or prepare the
@@ -4270,6 +4315,7 @@ int processCommand(client *c) {
     int is_denyloading_command = (combined_inv_flags & CMD_LOADING);
     int is_may_replicate_command = (combined_flags & (CMD_WRITE | CMD_MAY_REPLICATE));
     int is_deny_async_loading_command = (combined_flags & CMD_NO_ASYNC_LOADING);
+    int is_pubsub_command = (combined_flags & CMD_PUBSUB);
 
     const int obey_client = mustObeyClient(c);
 
@@ -4313,7 +4359,7 @@ int processCommand(client *c) {
     }
 
     if (clientSupportStandAloneRedirect(c) && !obey_client &&
-        (is_write_command || (is_read_command && !c->flag.readonly))) {
+        commandHasKey(c, c->cmd, c->argv, c->argc) && (is_write_command || (!c->flag.readonly && !is_pubsub_command))) {
         if (server.failover_state == FAILOVER_IN_PROGRESS) {
             /* During the FAILOVER process, when conditions are met (such as
              * when the force time is reached or the primary and replica offsets
--- a/tests/integration/replica-redirect.tcl
+++ b/tests/integration/replica-redirect.tcl
@@ -37,7 +37,7 @@ start_server {tags {needs:repl external:skip}} {
             $rr close
         }
 
-        test {replica allow read command by default} {
+        test {replica allows read command by default} {
             r get foo
         } {}
 
@@ -45,10 +45,15 @@ start_server {tags {needs:repl external:skip}} {
             assert_error {READONLY*} {r set foo bar}
         }
 
-        test {replica redirect read and write command after CLIENT CAPA REDIRECT} {
+        test {replica redirects read and write command after CLIENT CAPA REDIRECT} {
             r client capa redirect
             assert_error "REDIRECT $primary_host:$primary_port" {r set foo bar}
             assert_error "REDIRECT $primary_host:$primary_port" {r get foo}
+        }
+
+        test {replica redirects (read-only) script with keys} {
+            assert_error "REDIRECT $primary_host:$primary_port" {r eval {return redis.error_reply("ERR script must not be executed")} 1 foo}
+            assert_error "REDIRECT $primary_host:$primary_port" {r eval_ro {return redis.error_reply("ERR script must not be executed")} 1 foo}
         }
 
         test {CLIENT INFO} {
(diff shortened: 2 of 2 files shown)


src/server.c
```diff
@@ -4154,40 +4200,41 @@ int processCommand(client *c) {
...
-             * redirected to the replica and then redirected back, causing back
-             * and forth redirection. To avoid this situation, during the
-             * FAILOVER_IN_PROGRESS state, we temporarily suspend the clients
-             * that need to be redirected until the replica truly becomes the primary,
-             * and then resume the execution. */
-            blockPostponeClient(c);
-        } else {
-            if (c->cmd->proc == execCommand) {
-                discardTransaction(c);
+    if (!server.cluster_enabled && c->capa & CLIENT_CAPA_REDIRECT && server.primary_host && !obey_client) {
```
soloestoy:
```suggestion
    if (!server.cluster_enabled && c->capa & CLIENT_CAPA_REDIRECT && server.primary_host && !obey_client &&
        commandHasKey(c, c->cmd, c->argv, c->argc) && (is_write_command || (!c->flag.readonly && !is_pubsub_command))) {
        if (server.failover_state == FAILOVER_IN_PROGRESS) {
            /* During the FAILOVER process, when conditions are met (such as
             * when the force time is reached or the primary and replica offsets
             * are consistent), the primary actively becomes the replica and
             * transitions to the FAILOVER_IN_PROGRESS state.
             *
             * After the primary becomes the replica, and after handshaking
             * and other operations, it will eventually send the PSYNC FAILOVER
             * command to the replica, then the replica will become the primary.
             * This means that the upgrade of the replica to the primary is an
             * asynchronous operation, which implies that during the
             * FAILOVER_IN_PROGRESS state, there may be a period of time where
             * both nodes are replicas.
             *
             * In this scenario, if a -REDIRECT is returned, the request will be
             * redirected to the replica and then redirected back, causing back
             * and forth redirection. To avoid this situation, during the
             * FAILOVER_IN_PROGRESS state, we temporarily suspend the clients
             * that need to be redirected until the replica truly becomes the primary,
             * and then resume the execution. */
            blockPostponeClient(c);
        } else {
            if (c->cmd->proc == execCommand) {
                discardTransaction(c);
            } else {
                flagTransaction(c);
            }
            c->duration = 0;
            c->cmd->rejected_calls++;
            addReplyErrorSds(c, sdscatprintf(sdsempty(), "-REDIRECT %s:%d", server.primary_host, server.primary_port));
        }
        return C_OK;
    }
```

soloestoy:
>WATCH command (redirect in READWRITE mode is necessary)

This makes sense.

>transactions with no keys must not be redirected

I don't understand, this already aligns with the current behavior.

>scripting scenarios: scripts (with keys) must be redirected in some situations

Can you elaborate it? `EVAL` and `EVAL_RO` would or not be redirected in different scenarios?

gmbnomis:
> > transactions with no keys must not be redirected
> 
> I don't understand, this already aligns with the current behavior.

Mostly yes, I suppose, except when a script with no keys is used, see test case "replica allows MULTI that accesses no key". So, I think we need to look at the keys the commands within a transaction use, not only at the flags of those commands.
 
> > scripting scenarios: scripts (with keys) must be redirected in some situations
> 
> Can you elaborate it? `EVAL` and `EVAL_RO` would or not be redirected in different scenarios?

The proposal is to behave like in cluster mode, which basically means that EVAL and EVAL_RO do not behave differently when it comes to redirects (I think this is because `EVAL_RO` is rather new. If one does not have `EVAL_RO`, it makes sense for `EVAL` to assume that keys are used for reading in READONLY mode.) Thus:

In READWRITE mode on replica:

- Script without key is not redirected
- Script with keys is redirected (both `EVAL` and `EVAL_RO`)

In READONLY mode on replica:

- Script without keys is not redirected
- Script with keys is not redirected (both `EVAL` and `EVAL_RO`)
 
   There is a difference in the reported error, though,  if the script actually tries to write. For `EVAL_RO` this is `ERR Write commands are not allowed from read-only scripts`. And for `EVAL` this is `READONLY You can't write against a read only replica`. See also the corresponding discussion for cluster mode at https://github.com/valkey-io/valkey/issues/869#issuecomment-2563203949.

src/server.c
gmbnomis: I applied the suggestion and squashed it into a single commit.

ranshid:
> > WATCH command (redirect in READWRITE mode is necessary)
> 
> This makes sense.
> 

@soloestoy Why do you think it makes sense? I think using watch on the replicas also makes perfect sense as the watch also protects mutations arriving on the replication stream right? I think it does not make sense to redirect watch command, but maybe I misunderstood the intension here? @gmbnomis

gmbnomis:
> > > WATCH command (redirect in READWRITE mode is necessary)
> 
> @soloestoy Why do you think it makes sense? I think using watch on the replicas also makes perfect sense as the watch also protects mutations arriving on the replication stream right? I think it does not make sense to redirect watch command, but maybe I misunderstood the intension here? @gmbnomis

Yes, `WATCH` on a replica must be possible, but then one needs to enable `READONLY` for the connection in redirect mode  (exactly like in cluster mode).

ranshid:
> > > > WATCH command (redirect in READWRITE mode is necessary)
> > 
> > 
> > @soloestoy Why do you think it makes sense? I think using watch on the replicas also makes perfect sense as the watch also protects mutations arriving on the replication stream right? I think it does not make sense to redirect watch command, but maybe I misunderstood the intension here? @gmbnomis
> 
> Yes, `WATCH` on a replica must be possible, but then one needs to enable `READONLY` for the connection in redirect mode (exactly like in cluster mode).

Not sure why I need to be READONLY for that. although rare, writable replicas are still supported AFAIK. Also although I agree most known clients will issue READONLY on all replica in cluster mode disabled when rfr is used, AFAIK the READONLY was mainly introduced as a cluster command and less intended for serving cluster mode disabled. At the very least, this should be considered as a breaking change for clients.
Also what is the expected result of responding with REDIRECT to `WATCH`? should the client redirect it to the primary node and continue processing the transaction on the replica? I am not sure what is the proposal in this case.

gmbnomis:
> > Yes, `WATCH` on a replica must be possible, but then one needs to enable `READONLY` for the connection in redirect mode (exactly like in cluster mode).
> 
> Not sure why I need to be READONLY for that. although rare, writable replicas are still supported AFAIK. 

There is no change in the behavior of `WATCH` (or any other command accessing keys) unless the client enables
redirect mode on the connection:

```
127.0.0.1:6380> readwrite
OK
127.0.0.1:6380> watch a
OK
127.0.0.1:6380> readonly
OK
127.0.0.1:6380> watch a
OK
127.0.0.1:6380> client capa redirect
OK
127.0.0.1:6380> watch a
OK
127.0.0.1:6380> readwrite
OK
127.0.0.1:6380> watch a
(error) REDIRECT 127.0.0.1:6379
```
I think that a client states it intent that it wants to be redirected to the primary when it enables redirect mode.

> Also although I agree most known clients will issue READONLY on all replica in cluster mode disabled when rfr is used, AFAIK the READONLY was mainly introduced as a cluster command and less intended for serving cluster mode disabled. At the very least, this should be considered as a breaking change for clients. Also what is the expected result of responding with REDIRECT to `WATCH`? should the client redirect it to the primary node and continue processing the transaction on the replica? I am not sure what is the proposal in this case.

The use of readonly and readwrite has been part of redirect mode from the beginning, see  https://github.com/valkey-io/valkey/pull/325. So, this PR does not propose any new handling of readonly/readwrite in that mode (or any change on the client supporting redirect mode). It just fixes a couple of scenarios in which we should redirect (taking cluster mode as the reference)

As for a redirected `WATCH`, the entire transaction should take place on the primary, then.

ranshid:
> > > Yes, `WATCH` on a replica must be possible, but then one needs to enable `READONLY` for the connection in redirect mode (exactly like in cluster mode).
> > 
> > 
> > Not sure why I need to be READONLY for that. although rare, writable replicas are still supported AFAIK.
> 
> There is no change in the behavior of `WATCH` (or any other command accessing keys) unless the client enables redirect mode on the connection:
> 
> ```
> 127.0.0.1:6380> readwrite
> OK
> 127.0.0.1:6380> watch a
> OK
> 127.0.0.1:6380> readonly
> OK
> 127.0.0.1:6380> watch a
> OK
> 127.0.0.1:6380> client capa redirect
> OK
> 127.0.0.1:6380> watch a
> OK
> 127.0.0.1:6380> readwrite
> OK
> 127.0.0.1:6380> watch a
> (error) REDIRECT 127.0.0.1:6379
> ```
> 
> I think that a client states it intent that it wants to be redirected to the primary when it enables redirect mode.
> 
> > Also although I agree most known clients will issue READONLY on all replica in cluster mode disabled when rfr is used, AFAIK the READONLY was mainly introduced as a cluster command and less intended for serving cluster mode disabled. At the very least, this should be considered as a breaking change for clients. Also what is the expected result of responding with REDIRECT to `WATCH`? should the client redirect it to the primary node and continue processing the transaction on the replica? I am not sure what is the proposal in this case.
> 
> The use of readonly and readwrite has been part of redirect mode from the beginning, see #325. So, this PR does not propose any new handling of readonly/readwrite in that mode (or any change on the client supporting redirect mode). It just fixes a couple of scenarios in which we should redirect (taking cluster mode as the reference)
> 
> As for a redirected `WATCH`, the entire transaction should take place on the primary, then.

I understand your point (forgot that -REDIRECT will not work for writable replicas per a decision we took in #325) however maintaining the watch migrations not trivial IMO. For example say we redirect the watch and later there will be a failover switch between the primary and replica and the transaction will then be executed on the other node? 
I think it would be difficult to control how transactions are being handled by clients and maybe we should discuss it with client maintainers in order to understand what is the preferred way to operate transactions redirections.

ranshid:
@gmbnomis Also I agree that this is not an issue only with your current proposal. it is also an issue for cluster mode redirecting watch. I think there was an issue that point was discussed and we thought of introducing a new way to start a transaction with watches or something

ranshid:
> @gmbnomis Also I agree that this is not an issue only with your current proposal. it is also an issue for cluster mode redirecting watch. I think there was an issue that point was discussed and we thought of introducing a new way to start a transaction with watches or something

Anyhow, now that I understand better the suggestion I am fine with us only keeping better alignment with cluster mode MOVED. 
Thank you for taking the time to explain!

gmbnomis:
@soloestoy @ranshid I rebased the PR on the current unstable. This bug is now open since 8.0: I really would like to make some progress here (this bug is the reason why we do not even consider looking at using REDIRECT mode for our applications, as we use a lot of Lua scripts).

PR #1784, opened 2025-02-26
Save to Disk in Bio Thread and refactor readSyncBulkPayload

nitaicaro:
## **Introduction**

This PR introduces a new feature that enables replicas to perform disk-based synchronization on a dedicated background thread (Bio thread). Benchmarking results demonstrate significant improvements in synchronization duration. In extreme cases, this optimization allows syncs that would have previously failed to succeed. 

## **Problem Statement**

Some administrators prefer the disk-based full synchronization mode for replicas. This mode allows replicas to continue serving clients with data while downloading the RDB file.

Valkey's predominantly single-threaded nature creates a challenge: serving client read requests and saving data from the socket to disk are not truly concurrent operations. In practice, the replica alternates between processing client requests and replication data, leading to inefficient behavior and prolonged sync durations, especially under high load.

## **Proposed Solution**

To address this, the solution offloads the task of downloading the RDB file from the socket to a background thread. This allows the main thread to focus exclusively on handling client read requests while the background thread handles communication with the primary.

### **Potential for Improvement**
In theory, this optimization can lead to unbounded improvement in sync duration. By eliminating competition between client read events and socket communication (i.e., events related to handling RDB download with the primary), sync times become independent on load - the main thread handles only client reads, while the background thread focuses on primary RDB download events, allowing the system to perform consistently even under high load. 

The full valkey-benchmark commands can be found in the appendix below.

### Sync Duration with Feature Disabled (times in seconds)

16 threads, 64 clients: 172 seconds
32 threads, 128 clients: 436 seconds
48 threads, 192 clients: 710 seconds

### Sync Duration with Feature Enabled (times in seconds)

16 threads, 64 clients: 33 seconds (80.8% improvement)
32 threads, 128 clients: 33 seconds (92.4% improvement)
48 threads, 192 clients: 33 seconds (95.3% improvement)

image

## **Alternative Solutions Considered**

**IO Threads**  
IO threads to not have an advantage over Bio in this case: The save-to-disk job is rare (most likely no more than several executions in a replica's lifetime), and there is never more than one simultaneous execution. Bio threads make more sense for a single, slow long running operation.

**io_uring**  
For a single connection, io_uring doesn't provide as much of a performance boost because the primary advantage comes from batching many I/O operations together to reduce syscall overhead. With just one connection, we won't have enough operations to benefit significantly from these optimizations.

**Prioritizing primary's socket in the event loop**
This approach would help, but less effectively than using a Bio thread. We would still need to allocate attention to handling read requests, which could limit its benefit. It could be more useful on smaller instance types with limited CPU cores. Edit: In practice, this feature actually does this naturally when there is a single core - when both threads are running on the same core, the Bio thread can get up to 50% of CPU time now to handle RDB download, whereas before RDB download events were queued after a bunch of read events (and didn't get even close to 50%).

## **Code Design**

This PR introduces both the new BIO-based RDB sync flow described above and a significant (and much-needed) refactoring of the readSyncBulkPayload function. Previously, this single monolithic function handled the entire replication flow, covering both disk-based and socket-based syncs, and was responsible for everything from sync preparation and RDB reception to loading and finalization. The refactoring splits these responsibilities into clearer, more modular components, improving readability, reusability, and maintainability.

### Disk-Based RDB Sync (Bio thread)
The old disk-based RDB save logic has been removed from the main thread. It is now exclusively handled by a dedicated Bio thread, following the team’s decision to remove the need for a config.

New Flow Overview:

After the replication handshake completes in `syncWithPrimary()` (or at the end of the dual-channel handshake), the replica determines whether to use disk-based sync via `useDisklessLoad()` (unchanged).

If disk sync is chosen, a read handler is set on the primary’s connection:

`connSetReadHandler(conn, receiveRDBinBioThread);`
When the primary begins sending the RDB payload, this read handler is triggered. It creates a dedicated Bio thread to perform the sync.

The Bio thread executes `replicaReceiveRDBFromPrimaryToDisk`, which is now the main handler for receiving and saving RDB to disk.
This function replaces the disk logic previously found in `readSyncBulkPayload`, but executes in a busy loop (as opposed to being event-driven like `readSyncBulkPayload`) since the Bio thread is single-purpose and not event-based.

The logic inside `replicaReceiveRDBFromPrimaryToDisk` is a direct port of the previous flow, with some refactoring for clarity and reusability across sync modes.

Upon completion or failure, the Bio thread signals the main thread using a shared variable (`replica_bio_disk_save_state`).

The main thread detects Bio completion in a new function called `handleBioThreadFinishedRDBDownload()` (triggered via replication cron). This function then initiates the RDB load and resets relevant stats.

Note: The actual RDB loading remains on the main thread, as it already halts I/O during load and doesn’t benefit from thread offloading, simplifying thread-safety concerns.

### Diskless Sync (Socket)
For socket-based sync, the logic remains the same: asynchronous and event-driven, but has been refactored for modularity and readability.

The original implementation in `readSyncBulkPayload` mixed socket and disk sync logic with heavy branching.

Now, memory-based sync is isolated in a new function:

`replicaReceiveRDBFromPrimaryToMemory`
This function retains the same execution pattern as `readSyncBulkPayload` (i.e., reacting to available data on the socket), but is now cleaner and easier to follow.

Common logic that was previously duplicated across both sync modes is now factored out into shared helpers:

`replicaBeforeLoadPrimaryRDB()`
`replicaAfterLoadPrimaryRDB()`

The result is a more maintainable and modular flow. Reviewers are encouraged to compare this function side-by-side with the previous `readSyncBulkPayload` in order to understand the change easily.

### Metrics
Previously, replication metrics such as `repl_transfer_size` and `repl_transfer_read` were tracked on the main thread inside `readSyncBulkPayload`. With the move to Bio threads, those variables could not be safely reused due to thread safety concerns.

To address this:

New Bio-specific metrics are introduced:

`server.bio_repl_transfer_size`
`server.bio_repl_transfer_read`

These are updated by the Bio thread during disk-based syncs.

At the end of the sync, the values are merged back into the standard metrics (server.repl_transfer_*) for consistency.

If INFO is queried during an ongoing Bio-based sync, the reported values reflect either the Bio-specific metrics or a combined view when needed, ensuring accurate observability.

### **Benchmarking Setup**

- **Client machine:** AWS c5a.16xlarge  
- **Server machines:** AWS c5a.2xlarge

```bash
# Step 1: Fill the primary and replica DBs with 6GB of data:

./valkey-benchmark -h <host> -p <port> -l -d 128 -t set -r 30000000 --threads 16 -c 64

# Step 2: Initiate heavy read load on the replica:

./valkey-benchmark -h <host> -p <port> -t get -r 30000000 --threads <t> -c <t> -n 1000000000 -P <P>

# Step 3: Enable/disable the config controlling the new feature:

./valkey-cli -h <host> -p <port> config set replica-save-to-disk-in-bio-thread <yes/no>

# Step 4: Initiate sync:

./valkey-cli -h <replica host> -p <replica port> replicaof <primary host> <primary port>

### CODE REVIEW

File: src/bio.c [hunk 1]
```diff
@@ -203,10 +217,54 @@ void bioCreateFsyncJob(int fd, long long offset, int need_reclaim_cache) {
...
+    bio_job *job = zmalloc(sizeof(*job));
+    job->save_to_disk_args.conn = conn;
+    job->save_to_disk_args.sync_io_timeout = sync_io_timeout;
+    // job->save_to_disk_args.dest = zmalloc(sizeof(rio));
+    // memcpy(job->save_to_disk_args.dest, dest, sizeof(rio));
+    job->save_to_disk_args.dest_fd = dest_fd;
+    bioSubmitJob(BIO_SAVE_TO_DISK, job);
+}
+
+long long cancelReplicationHandshakeTimeEvent(
```
nitaicaro: I'm going to replace these with "done handlers" in the replication cron.

src/bio.c
```diff
@@ -77,6 +79,7 @@ static unsigned int bio_job_to_worker[] = {
     [BIO_AOF_FSYNC] = 1,
     [BIO_CLOSE_AOF] = 1,
     [BIO_LAZY_FREE] = 2,
+    [BIO_SAVE_TO_DISK] = 3,
```
xbasel: `BIO_SAVE_TO_DISK` is too generic. Something like `BIO_RDB_SAVE` is probably clearer.

src/bio.c
```diff
@@ -108,6 +111,11 @@ typedef union bio_job {
         lazy_free_fn *free_fn; /* Function that will free the provided arguments */
         void *free_args[];     /* List of arguments to be passed to the free function */
     } free_args;
+    struct {
```
xbasel, line 116: space above

src/bio.c
```diff
@@ -278,6 +293,158 @@ void *bioProcessBackgroundJobs(void *arg) {
             if (job_type == BIO_CLOSE_AOF) close(job->fd_args.fd);
         } else if (job_type == BIO_LAZY_FREE) {
             job->free_args.free_fn(job->free_args.free_args);
+        } else if (job_type == BIO_SAVE_TO_DISK) {
+            int usemark;
+            char lastbytes[RDB_EOF_MARK_SIZE], buf[PROTO_IOBUF_LEN], eofmark[RDB_EOF_MARK_SIZE];
```
xbasel: Declaring multiple arrays in one line reduces readability. Split them into separate lines for clarity.

src/replication.c
```diff
@@ -2649,6 +2678,12 @@ void freePendingReplDataBuf(void) {
     server.pending_repl_data.len = 0;
 }
 
+void receiveRDBinBioThread(connection *conn) {
+    serverLog(LL_NOTICE, "Replica main thread creating Bio thread to save RDB to disk");
+    connSetReadHandler(conn, NULL);
```
xbasel, line 2795: What happens to the write handler? Is the main thread supposed to do any writes in the meantime?

src/replication.c
```diff
@@ -3918,6 +3963,10 @@ void replicationAbortSyncTransfer(void) {
     cleanupTransferResources();
 }
 
+void waitForDiskSaveBioThreadComplete(void) {
+    while (bioPendingJobsOfType(BIO_SAVE_TO_DISK));
```
xbasel: This is a blocking operation and should be avoided in the main thread.  Although this is already being done in the main thread I think, via: `bioDrainWorker`

src/replication.c
```diff
@@ -2100,9 +2100,60 @@ void replicationAttachToNewPrimary(void) {
...
+int shouldAbortSave(void) {
+    return atomic_load_explicit(&server.replica_bio_abort_save, memory_order_relaxed);
+}
+
+int inspectBulkPayloadHeaderForErrors(char *buf) {
+    if (buf[0] == '-')
+        return INSPECT_BULK_PAYLOAD_PRIMARY_ABORT;
+    else if (buf[0] == '\0')
+        return INSPECT_BULK_PAYLOAD_PRIMARY_PING;
+    else if (buf[0] != '$')
```
xbasel: no point for the "else" keyword, reduces readability.

src/bio.c
```diff
@@ -278,6 +293,158 @@ void *bioProcessBackgroundJobs(void *arg) {
...
+                    replicaBioSaveServerLog(LL_WARNING, "PRIMARY aborted replication with an error: %s", buf + 1);
+                    error = 1;
+                    goto done;
+                } else if (ret == INSPECT_BULK_PAYLOAD_PRIMARY_PING) {
+                    atomic_store_explicit(&server.repl_transfer_lastio, atomic_load_explicit(&server.unixtime, memory_order_relaxed), memory_order_relaxed);
+                    memset(buf, 0, PROTO_IOBUF_LEN);
+                    received_ping = 1;
+                } else if (ret == INSPECT_BULK_PAYLOAD_PRIMARY_BAD_PROTO) {
+                    replicaBioSaveServerLog(LL_WARNING,
+                                            "Bad protocol from PRIMARY, the first byte is not '$' (we received '%s'), are you sure the host "
```
xbasel: This log (and the one above) is repeated in multiple places. Put it in a shared constant to avoid duplication and inconsistency.

src/bio.c
```diff
@@ -278,6 +293,158 @@ void *bioProcessBackgroundJobs(void *arg) {
...
+                if (nread <= 0) {
+                    if (connGetState(conn) == CONN_STATE_CONNECTED) {
+                        /* equivalent to EAGAIN */
+                        memset(buf, 0, PROTO_IOBUF_LEN);
+                        continue;
+                    }
+                    replicaBioSaveServerLog(LL_WARNING, "I/O error trying to sync with PRIMARY: %s",
+                                            (nread == -1) ? connGetLastError(conn) : "connection lost");
+                    error = 1;
+                    goto done;
```
xbasel: why not to create an `error` label instead of using a variable?

src/bio.c
```diff
@@ -278,6 +293,158 @@ void *bioProcessBackgroundJobs(void *arg) {
...
+                if (shouldAbortSave()) {
+                    main_thread_abort = 1;
+                    goto done;
+                }
+                nread = connSyncReadLine(conn, buf, 1024, server.repl_syncio_timeout * 1000);
+                if (nread == -1) {
+                    replicaBioSaveServerLog(LL_WARNING, "I/O error reading bulk count from PRIMARY: %s", connGetLastError(conn));
+                    error = 1;
+                    goto done;
+                } else {
```
xbasel: This `else` isn't necessary. remove for better readability.  This applies to other places in the code.

src/bio.c
```diff
@@ -278,6 +293,158 @@ void *bioProcessBackgroundJobs(void *arg) {
...
+                    server.bio_stat_net_repl_input_bytes += nread + 1;
+                }
+                int ret = inspectBulkPayloadHeaderForErrors(buf);
+                if (ret == INSPECT_BULK_PAYLOAD_PRIMARY_ABORT) {
+                    replicaBioSaveServerLog(LL_WARNING, "PRIMARY aborted replication with an error: %s", buf + 1);
+                    error = 1;
+                    goto done;
+                } else if (ret == INSPECT_BULK_PAYLOAD_PRIMARY_PING) {
+                    atomic_store_explicit(&server.repl_transfer_lastio, atomic_load_explicit(&server.unixtime, memory_order_relaxed), memory_order_relaxed);
+                    memset(buf, 0, PROTO_IOBUF_LEN);
```
xbasel: why this is being done?

src/bio.c
```diff
@@ -278,6 +293,158 @@ void *bioProcessBackgroundJobs(void *arg) {
...
+            off_t left = 0, repl_transfer_last_fsync_off = 0;
+            int error = 0, main_thread_abort = 0, received_ping = 0;
+            connection *conn = job->save_to_disk_args.conn;
+            int dest_fd = job->save_to_disk_args.dest_fd;
+
+            /* Put the socket in blocking mode to simplify RDB transfer.
+             * We'll restore it when the RDB is received. */
+            connBlock(conn);
+            connRecvTimeout(conn, server.repl_syncio_timeout * 1000);
+            do {
```
xbasel: why do we need two loops in the code?

src/bio.c
```diff
@@ -278,6 +293,158 @@ void *bioProcessBackgroundJobs(void *arg) {
...
+                received_ping = 0;
```
xbasel: Can you please explain the role of `received_ping` ?

xbasel (commented):
Initial comments.

src/replication.c
nitaicaro: We have to do this to prevent a race condition where a new sync starts before the previous Bio thread finishes.
Busy-waiting is also how it's done for io-threads, see `waitForClient()`

src/bio.c
nitaicaro: This is the purpose of a ping from the primary - to refresh `last_io` field in order to avoid timeout

src/bio.c
nitaicaro: The primary sometimes sends a ping in order to keep the connection alive. I'll bring back the comment that explains this

src/bio.c
nitaicaro: We are basically imitating what `readSyncBulkPayload` does for normal replication. We try to read the bulk payload length from the primary. Ideally one pass would be enough (no need for loop), but the primary is sometimes not fast enough in sending the length, so it periodically sends pings (newlines) until it's ready. We have to keep looping while we receive these pings.

The second loop goes on until the primary finishes sending all the data (we see an EOF or the amount we read is equal to the previously passed payload length).

So the first loop is conditioned on receiving pings, the second one on sync completion.

src/bio.c
nitaicaro: I think we cannot have more than one label inside the `else if (job_type == BIO_RDB_SAVE)`.
If we have 
```
done:
...
error:
...
```

then `error` would always get executed after `done` (we cannot return after done since we need to reach the end of the iteration:
```
zfree(job);

/* Lock again before reiterating the loop, if there are no longer
 * jobs to process we'll block again in pthread_cond_wait(). */
pthread_mutex_lock(&bio_mutex[worker]);
listDelNode(bio_jobs[worker], ln);
bio_jobs_counter[job_type]--;
pthread_cond_signal(&bio_newjob_cond[worker]);
```
)

I guess we can do
```
done:
...
goto really_done;
error:
...
really_done:

```
But I'm not sure it makes things better

src/replication.c
nitaicaro: Good question, it was never set to anything on the replica's side, see:
https://github.com/valkey-io/valkey/blob/089b830479f4984e77e570560f85e37212879740/src/replication.c#L3517

^ this is executed during sync handshake. After the sync is done and we enter steady-state we initialize it: https://github.com/valkey-io/valkey/blob/089b830479f4984e77e570560f85e37212879740/src/replication.c#L4380

src/replication.c
xbasel: (1) can `bioDrainWorker` be used?
(2) Did you consider timing out the operation and shutting down ?

nitaicaro: Good point - since we switched to blocking mode on the bio thread we can have a deadlock here. I'll remove this, and then the timeouts of `read()` will guarantee that we eventually reach `shouldAbortSave()` which will free the main thread.

Edit:
Actually we set a timeout so we cannot block forever:
`connRecvTimeout(conn, server.repl_syncio_timeout * 1000);`

So we are guaranteed to eventually reach `shouldAbortSave()` which will free the main thread from busy-waiting.

xbasel: Blocking the main thread forever is bad practice, even if the child thread is logically guaranteed to finish. 
Maybe timing it out and crashing is the right approach.

nitaicaro: Your suggestion to switch to `bioDrain` seems to be the right solution. Updated.
About blocking forever, this seems to be a risk in other places in the code too, whether they busy-wait with `bioDrain `or with `bioPendingJobsOfType`. I suggest we open an issue to address this more comprehensively.

src/bio.c
```diff
@@ -278,6 +294,154 @@ void *bioProcessBackgroundJobs(void *arg) {
             if (job_type == BIO_CLOSE_AOF) close(job->fd_args.fd);
         } else if (job_type == BIO_LAZY_FREE) {
             job->free_args.free_fn(job->free_args.free_args);
+        } else if (job_type == BIO_RDB_SAVE) {
```
xbasel, line 300: Can this logic be refactored into a function, and ideally moved to replication.c ?

ranshid:
adding the major-decision-pending even though there is not a breaking change but I think another thread work should be carefully considered

madolson (commented):
Major decision is approved. No config.

madolson:
@Fusl In the weekly meeting we thought about your workload when reviewing this feature and deciding if we should enable it by default. Do you think you would want to have this feature enabled by default?

Fusl:
> Do you think you would want to have this feature enabled by default?

I currently use `repl-diskless-load swapdb` everywhere where I have clients reading from replicas and performance matters, I don't think this feature would be relevant to my current workload.
I'm wondering what the overhead/performance penalty of this feature is for environments where there is only a single CPU core available (such as in resource-limited containerized environments) with no, with light, and with heavy reads being done from replicas. Do we see an overall decrease in performance due to two threads contesting for the same CPU core?

I think more testing for this feature needs to be done on real cluster setups before we turn it on by default.

nitaicaro:
> > Do you think you would want to have this feature enabled by default?
> 
> I currently use `repl-diskless-load swapdb` everywhere where I have clients reading from replicas and performance matters, I don't think this feature would be relevant to my current workload. I'm wondering what the overhead/performance penalty of this feature is for environments where there is only a single CPU core available (such as in resource-limited containerized environments) with no, with light, and with heavy reads being done from replicas. Do we see an overall decrease in performance due to two threads contesting for the same CPU core?
> 
> I think more testing for this feature needs to be done on real cluster setups before we turn it on by default.

Thanks @Fusl , good points.

This change is unrelated to repl-diskless-load swapdb.

For disk-based sync, I tested this on single core, using a smaller version of the benchmark from the PR. So we see a drop in throughput during sync, but it completes successfully in a few seconds, on a scenario where without the Bio thread, it just hangs. 

Overall, even on a single core, this looks like a clear win for stability.

We can explain this easily - the Bio thread can get up to 50% of CPU time now to handle RDB download, whereas before it was queued after a bunch of read events (and doesn't get even close to 50%).

P.s., without Bio, writing to disk in the main thread can block in uninterruptible sleep due to kernel I/O waits, stalling the engine. So as a bonus, the main thread is less likely to enter this mode now.

xbasel:
Please fix DCO errors.

src/bio.c
```diff
@@ -108,6 +111,12 @@ typedef union bio_job {
         lazy_free_fn *free_fn; /* Function that will free the provided arguments */
         void *free_args[];     /* List of arguments to be passed to the free function */
     } free_args;
+
+    struct {
+        int type;
+        connection *conn; /* Connection to download the RDB from */
+        int dest_fd;      /* fd to save the RDB to */
```
ranshid: nit - just to avoid unnecessary alignment paddings
```suggestion
        int type;
        int dest_fd;      /* fd to save the RDB to */
        connection *conn; /* Connection to download the RDB from */
```

src/bio.c
ranshid: I must say I agree with @xbasel. this seems like a lot od code duplication and although I sometimes pro duplicating the code if the alternative is very bad, in this case I am not convinced it does not worth investing more time to make readSyncBulkPayload used from here, or separate the shared code in a way both flows can use it.

src/bio.c
```diff
@@ -278,6 +294,154 @@ void *bioProcessBackgroundJobs(void *arg) {
...
+                }
+                nread = connSyncReadLine(conn, buf, 1024, server.repl_syncio_timeout * 1000);
+                if (nread == -1) {
+                    replicaBioSaveServerLog(LL_WARNING, "I/O error reading bulk count from PRIMARY: %s", connGetLastError(conn));
+                    error = 1;
+                    goto done;
+                }
+                /* nread here is returned by connSyncReadLine(), which calls syncReadLine() and
+                 * convert "\r\n" to '\0' so 1 byte is lost. */
+                server.bio_stat_net_repl_input_bytes += nread + 1;
```
ranshid: so we will increment bio_stat_net_repl_input_bytes what about  stat_net_repl_input_bytes nread?

ranshid: Oh I see it will be updated in handleBioThreadFinishedRDBDownload which will happen after the save will complete right? Not sure this is so great. maybe some applications are tracking it during rdb transfer. Suggest to merge the code and also dynamically update the repl stats.

ranshid (commented):
just some top level comments. 

I also think that the top comment does great providing justification in terms of benchmark results. However it is missing the details of the design part. at first I thought it would be straight forward (simply operating readSyncBulkPayload in bio thread) but as we now added multiple synchronization and indications to interact with the replication cron. it would be great to document it

src/bio.c
nitaicaro: It was done for thread safety reasons. We wanted to completely separate the code accessed by the main vs bio thread. Let's discuss in our meeting

src/replication.c
```diff
@@ -2278,9 +2284,12 @@ void readSyncBulkPayload(connection *conn) {
 
         /* If the transfer is yet not complete, we need to read more, so
          * return ASAP and wait for the handler to be called again. */
-        if (!eof_reached) return;
+        if (!eof_reached) return C_RETRY;
+
+        return C_OK;
```
nitaicaro: Before this change, we would not return anything, and just allow the main thread to fall through to the load_rdb part. Now we want to return, terminate the bio thread, signal completion, and have the main thread do load_rdb. This is why we return C_OK.

src/replication.c
```diff
@@ -2124,173 +2124,97 @@ void replicationAttachToNewPrimary(void) {
...
+ * transfer. In the latter case, the following format is used:
+ *
+ * $EOF:<40 bytes delimiter>
+ *
+ * At the end of the file the announced delimiter is transmitted. The
+ * delimiter is long and random enough that the probability of a
+ * collision with the actual file content can be ignored.
+ * For the latter case, where a EOF mark is used, we return 1,
+ * and otherwise 0. */
+int inspectBulkPayloadHeaderForEOF(char *buf, char *eofmark, char *lastbytes) {
```
ranshid: can we use bool instead of int for the return type?

src/replication.c
```diff
@@ -2124,173 +2124,97 @@ void replicationAttachToNewPrimary(void) {
...
-
-        /* Sync data on disk from time to time, otherwise at the end of the
-         * transfer we may suffer a big delay as the memory buffers are copied
-         * into the actual disk. */
-        if (server.repl_transfer_read >= server.repl_transfer_last_fsync_off + REPL_MAX_WRITTEN_BEFORE_FSYNC) {
-            off_t sync_size = server.repl_transfer_read - server.repl_transfer_last_fsync_off;
-            rdb_fsync_range(server.repl_transfer_fd, server.repl_transfer_last_fsync_off, sync_size);
-            server.repl_transfer_last_fsync_off += sync_size;
-        }
+int inspectBulkPayloadForEOF(char *buf, ssize_t nread, char *eofmark, char *lastbytes) {
```
ranshid: lets also document this function 
AND
Can this return bool instead as well?

src/replication.c
```diff
@@ -2124,173 +2124,97 @@ void replicationAttachToNewPrimary(void) {
...
+        memcpy(lastbytes + rem, buf, nread);
+    }
+    return (memcmp(lastbytes, eofmark, RDB_EOF_MARK_SIZE) == 0);
+}
 
-        /* Check if the transfer is now complete */
-        if (!usemark) {
-            if (server.repl_transfer_read == server.repl_transfer_size) eof_reached = 1;
-        }
+int tryReadBulkPayloadMetadata(connection *conn, char *buf, char *eofmark, char *lastbytes, int *usemark, off_t *repl_transfer_size) {
```
ranshid, line 2137: this function holds lot's of logic. can you please document the main logical steps it takes and the existing return values and their meaning

src/replication.c
```diff
@@ -2673,6 +2803,32 @@ void freePendingReplDataBuf(void) {
     server.pending_repl_data.len = 0;
 }
 
+void receiveRDBinBioThread(int is_dual_channel) {
```
ranshid: can we change it to receive `bool is_dual_channel`

src/server.c
```diff
@@ -1315,7 +1315,7 @@ void databasesCron(void) {
 static inline void updateCachedTimeWithUs(int update_daylight_info, const long long ustime) {
     server.ustime = ustime;
     server.mstime = server.ustime / 1000;
-    server.unixtime = server.mstime / 1000;
+    atomic_store_explicit(&server.unixtime, server.mstime / 1000, memory_order_relaxed);
```
ranshid: TBH I am not sure all this relaxed access should be really handled. 
AFAIK all the modern processors and compilers will probably be fine doing the simple load store correctly given the memory is aligned. I am not referring to a performance concern, since relaxed atomic store will probably cause nothing on most platform (but that is exactly why we might not need it :) ) 
My problem is that simply the fact we now need to start consider for each global variable "mmm is it possible this will be accessed by another thread right now or not?" is the main problem.
For example, you did not change ALL places were we store the server.repl_transfer_lastio (which is probably fine, since some of them are storing a value when there are CURRENTLY no other threads loading it which are running) however this will complicate the future maintenance IMO.
I think just adding the _Atomic for these should be enough. 
For the synchronization atomics (like server.replica_bio_disk_save_state) is a totally different story and there we MUST place the correct barriers and/or use explicit acquire/release semantics.

would love to hear what @lipzhu @xbasel and others think?

src/replication.c
ranshid: @nitaicaro so what about it?

How do we handle the KA notification to the primary during load? 
The way I know it is done in replicationSendNewlineToPrimary, but now it has no connection? 
also if it did, I do not think it is good to have different threads do the read and write. probably will work in most implementations due to the full duplex implementation, but some L5 parts might break (like openssl)

src/server.h
```diff
@@ -1975,6 +1989,18 @@ struct valkeyServer {
                                                 * delay (start sooner if they all connect). */
     int dual_channel_replication;              /* Config used to determine if the replica should
                                                 * use dual channel replication for full syncs. */
+    _Atomic int replica_bio_disk_save_state;   /* Flag set by the bio thread to indicate that the
+                                                * RDB save to disk has completed, or failed */
+    _Atomic int replica_bio_abort_save;        /* Flag set by main thread, used to signal to replica's
```
ranshid: why not bool?

src/replication.c
```diff
@@ -2309,258 +2233,464 @@ void readSyncBulkPayload(connection *conn) {
...
+        server.repl_transfer_s = conn;
+    }
+
+    /* Handle completion */
+    if (error) {
+        replicaBioSaveServerLog(LL_WARNING, "Error downloading RDB");
+        atomic_store_explicit(&server.replica_bio_disk_save_state, REPL_BIO_DISK_SAVE_STATE_FAIL, memory_order_release);
+    } else {
+        replicaBioSaveServerLog(LL_NOTICE, "Done downloading RDB");
+        server.replica_bio_disk_save_conn = conn;
```
ranshid: can you explain why do we need this? we already restored the repl_transfer_s/repl_rdb_transfer_s so what is the point of remembering another connection?

ranshid (commented):
left some comments. lets address them and then I will continue

src/server.h
nitaicaro: `replica_bio_disk_save_state` can hold many values, but for `replica_bio_abort_save` you are correct

src/replication.c
nitaicaro: We don't have to do it when _receiving_ the RDB - the Bio thread has the connection and it continuously reads from it, so we don't time out.

When _loading_ the RDB, we don't read from the primary's connection, and the primary could think that the replica timed out, so it uses `replicationSendNewlineToPrimary` like you said. The `connSetWriteHandler(conn, NULL); ` that I linked above does not delete the connection itself, only the write handler used for async communication. But `replicationSendNewlineToPrimary` uses `connWrite()` which is a direct write attempt that doesn't require an event handler.

src/server.c
nitaicaro: As discussed, leaving this as-is and waiting for further feedback

src/replication.c
nitaicaro: The RDB load part of the sync is handled by the main thread in `handleBioThreadFinishedRDBDownload` which is called asynchronously in the replication cron. That function needs to use that connection but it does not "know" whether we had a single or dual channel sync.

In other words - in the beginning of the sync we took away the connection from the main thread and gave it to the Bio thread. This is our way to give it back to the main thread.

src/bio.c
nitaicaro: Done

src/replication.c
ranshid: I understand the second part (even though I do not understand why we keep it this way). I still do not understand why we need to return it.

is it to perform this logic?:

```
/* Final setup of the connected replica <- primary link */
    if (conn == server.repl_rdb_transfer_s) {
        dualChannelSyncHandleRdbLoadCompletion();
    } else {
        replicationCreatePrimaryClient(server.repl_transfer_s, rsi->repl_stream_db);
        server.repl_state = REPL_STATE_CONNECTED;
        server.repl_down_since = 0;
        /* Send the initial ACK immediately to put this replica in online state. */
        replicationSendAck();
    }
```

src/replication.c
ranshid: So, I am not referring to the write handler. I am referring to the write in 2 threads. this could break some implementations which can have issues supporting read and write being done in 2 different threads. also I am not sure this is the loading phase (maybe I was wrong about referring to replicationSendNewlineToPrimary). also the replicationSendNewlineToPrimary will NOT send '\n' when the repl_transfer_s is NULL AFAIK so the replica will NOT KA when loading (which again is not the case right?) 
I do think that in the RDB_TRANSFER pahse the replica will send occasional ACK using replicationSendAck in the replicationCron, but again I have some concerns about it breaking external TLS or other transport implementations

ranshid: O.K after we discussed and I looked, the replica does not need to write anything to the primary during transfer (loading is somewhat different). I guess it is fine then.

src/replication.c
```diff
@@ -4040,6 +4233,12 @@ void replicationAbortSyncTransfer(void) {
  *
  * Otherwise zero is returned and no operation is performed at all. */
 int cancelReplicationHandshake(int reconnect) {
+    if (bioPendingJobsOfType(BIO_RDB_SAVE)) {
+        atomic_store_explicit(&server.replica_bio_abort_save, 1, memory_order_relaxed);
```
ranshid: no sure if the relaxed access is needed here. we have the pthread providing a barrier and the server.replica_bio_abort_save is _Atomic...

src/replication.c
```diff
@@ -4817,14 +5062,14 @@ void replicationCron(void) {
 
     /* Non blocking connection timeout? */
     if (server.primary_host && (server.repl_state == REPL_STATE_CONNECTING || replicaIsInHandshakeState()) &&
-        (time(NULL) - server.repl_transfer_lastio) > server.repl_timeout) {
+        (time(NULL) - atomic_load_explicit(&server.repl_transfer_lastio, memory_order_relaxed)) > server.repl_timeout) {
```
ranshid: Still I think the relaxed access here is not needed

src/replication.c
```diff
@@ -4040,6 +4233,12 @@ void replicationAbortSyncTransfer(void) {
  *
  * Otherwise zero is returned and no operation is performed at all. */
 int cancelReplicationHandshake(int reconnect) {
+    if (bioPendingJobsOfType(BIO_RDB_SAVE)) {
+        atomic_store_explicit(&server.replica_bio_abort_save, 1, memory_order_relaxed);
+        bioDrainWorker(BIO_RDB_SAVE);
+        atomic_store_explicit(&server.replica_bio_abort_save, 0, memory_order_relaxed);
```
ranshid: can we add an assertion here checking the bio state is indeed terminated?

src/replication.c
```diff
@@ -4040,6 +4233,12 @@ void replicationAbortSyncTransfer(void) {
...
+    }
+    resetBioRDBSaveState();
```
ranshid, line 4207: why this is not placed under the prev `if (bioPendingJobsOfType(BIO_RDB_SAVE))`?

nitaicaro: If the Bio thread had terminated, we still have the part where we load the RDB - In this case the save state will be `REPL_BIO_DISK_SAVE_STATE_FINISHED`, not `REPL_BIO_DISK_SAVE_STATE_NONE`

ranshid: NVM. discussed this offline and we can keep it like this for now. The main reason is that it might be helpful to avoid different race condition cases

src/bio.c
```diff
@@ -333,3 +354,7 @@ void bioKillThreads(void) {
         }
     }
 }
+
+int inBioThread(void) {
```
xbasel, line 358: Why is this needed? I see it's only used with `debugServerAssert`.

nitaicaro: It's also used in `tryReadBulkPayloadMetadata`

src/bio.c
```diff
@@ -77,13 +79,15 @@ static unsigned int bio_job_to_worker[] = {
...
     [BIO_LAZY_FREE] = 2,
+    [BIO_RDB_SAVE] = 3,
 };
 
 static pthread_t bio_threads[BIO_WORKER_NUM];
 static pthread_mutex_t bio_mutex[BIO_WORKER_NUM];
 static pthread_cond_t bio_newjob_cond[BIO_WORKER_NUM];
 static list *bio_jobs[BIO_WORKER_NUM];
 static unsigned long bio_jobs_counter[BIO_NUM_OPS] = {0};
+static __thread int bio_thread_id = 0; /* Thread local var */
```
xbasel: The comment is redundant. __thread already implies it's thread-local.

src/bio.c
```diff
@@ -225,6 +242,8 @@ void *bioProcessBackgroundJobs(void *arg) {
     if (pthread_sigmask(SIG_BLOCK, &sigset, NULL))
         serverLog(LL_WARNING, "Warning: can't mask SIGALRM in bio.c thread: %s", strerror(errno));
 
+    bio_thread_id = (int)worker;
```
xbasel: why `bio_thread_id` type doesn't match `worker`s type?

src/replication.c
```diff
@@ -2124,173 +2124,103 @@ void replicationAttachToNewPrimary(void) {
...
-#define REPL_MAX_WRITTEN_BEFORE_FSYNC (1024 * 1024 * 8) /* 8 MB */
-void readSyncBulkPayload(connection *conn) {
-    char buf[PROTO_IOBUF_LEN];
-    ssize_t nread, readlen, nwritten;
-    int use_diskless_load = useDisklessLoad();
-    serverDb **diskless_load_tempDb = NULL;
-    functionsLibCtx *temp_functions_lib_ctx = NULL;
-    int empty_db_flags = server.repl_replica_lazy_flush ? EMPTYDB_ASYNC : EMPTYDB_NO_FLAGS;
-    off_t left;
+int inspectBulkPayloadHeaderForErrors(char *buf) {
```
xbasel: This function adds unnecessary indirection for very simple logic that's only used once. Inlining the check directly into `tryReadBulkPayloadMetadata()` would improve readability and make the control flow easier to follow without jumping between locations.

also if this function is really needed, it should be static.

src/server.h
```diff
@@ -3015,6 +3039,13 @@ int sendCurrentOffsetToReplica(client *replica);
 void addRdbReplicaToPsyncWait(client *replica);
 void initClientReplicationData(client *c);
 void freeClientReplicationData(client *c);
+void replicaReceiveRDBFromPrimaryToMemory(connection *conn);
+int inspectBulkPayloadHeaderForErrors(char *buf);
+bool inspectBulkPayloadHeaderForEOF(char *buf, char *eofmark, char *lastbytes);
```
xbasel: why this function is exposed as an API ?

src/replication.c
```diff
@@ -2124,173 +2124,103 @@ void replicationAttachToNewPrimary(void) {
...
+ * transfer. In the latter case, the following format is used:
+ *
+ * $EOF:<40 bytes delimiter>
+ *
+ * At the end of the file the announced delimiter is transmitted. The
+ * delimiter is long and random enough that the probability of a
+ * collision with the actual file content can be ignored.
+ * For the latter case, where a EOF mark is used, we return 1,
+ * and otherwise 0. */
+bool inspectBulkPayloadHeaderForEOF(char *buf, char *eofmark, char *lastbytes) {
```
xbasel: 1. Not sure I understand from the comment what this function is supposed to return, neither from the name. Maybe rename to `usermark()`. 
2. Can be inlined, or made static. Its used from a single location.

src/replication.c
```diff
@@ -2124,173 +2124,103 @@ void replicationAttachToNewPrimary(void) {
...
-        nread = connSyncReadLine(conn, buf, 1024, server.repl_syncio_timeout * 1000);
-        if (nread == -1) {
-            serverLog(LL_WARNING, "I/O error reading bulk count from PRIMARY: %s", connGetLastError(conn));
-            goto error;
-        } else {
-            /* nread here is returned by connSyncReadLine(), which calls syncReadLine() and
-             * convert "\r\n" to '\0' so 1 byte is lost. */
+/* Check if the bulk payload contains the EOF delimiter.
+ * Updates lastbytes with the most recent bytes and returns 1 if EOF is found. */
+bool inspectBulkPayloadForEOF(char *buf, ssize_t nread, char *eofmark, char *lastbytes) {
```
xbasel: `inspectBulkPayloadForEOF` is a bit misleading, it's updating state, not just inspecting. Maybe rename to `updateEofTracker()` or something similar  for clarity. Also not sure why this function isn't inlined (or at least static)

src/replication.c
```diff
@@ -2124,173 +2124,103 @@ void replicationAttachToNewPrimary(void) {
...
+        memcpy(lastbytes, buf + nread - RDB_EOF_MARK_SIZE, RDB_EOF_MARK_SIZE);
+    } else {
+        int rem = RDB_EOF_MARK_SIZE - nread;
+        memmove(lastbytes, lastbytes + nread, rem);
+        memcpy(lastbytes + rem, buf, nread);
+    }
+    return (memcmp(lastbytes, eofmark, RDB_EOF_MARK_SIZE) == 0);
+}
+
+/* During replication, the primary sends sync metadata as the first line
```
xbasel, line 2128: btw, there's an overlap in comment here and in `inspectBulkPayloadHeaderForEOF`

src/server.h
```diff
@@ -3015,6 +3039,13 @@ int sendCurrentOffsetToReplica(client *replica);
...
+bool inspectBulkPayloadForEOF(char *buf, ssize_t nread, char *eofmark, char *lastbytes);
+int tryReadBulkPayloadMetadata(connection *conn, char *buf, char *eofmark, char *lastbytes, int *usemark, off_t *repl_transfer_size);
```
xbasel: this this function is declared here?

src/replication.c
nitaicaro: Done

src/bio.c
nitaicaro: Done

src/bio.c
nitaicaro: Done

src/server.h
nitaicaro: Done

src/server.h
nitaicaro: Done

tests/integration/replica-redirect.tcl
```diff
@@ -19,6 +19,11 @@ start_server {tags {needs:repl external:skip}} {
             r replicaof $primary_host $primary_port
             wait_replica_online $primary
 
+            # In case of bio thread RDB download, there can be up to 1000ms 
+            # (1 replication cron loop) delay until the rdb starts loading
+            after 1000
```
xbasel, line 24: Why is this necessary? the next `wait_done_loading` isn't enough?

nitaicaro: The way `wait_done_loading` works is by polling PING to detect when Valkey becomes responsive, which only happens after the RDB loading process has started. However, there can be up to a 1000ms delay before the loading even begins - because the replication cron that triggers the RDB load runs asynchronously, and we can't predict exactly when its next iteration will happen. Without the after 1000, `wait_done_loading` might start too early and incorrectly conclude that the server is already ready when it hasn't started loading yet.

ranshid (approved):
LGTM thank you @nitaicaro - good work!

PR #1786, opened 2025-02-26
Fix undefined behaviour in bitops unit test

zuiderkwast:
The unit test was added in #1741. It fails when compiled with UBSAN. Using a local array like `char buf[size]` is undefined behaviour when `size == 0`. This fix just makes it size 1 in this case.

The failure I get locally:

    unit/test_bitops.c:28:13: runtime error: variable length array bound evaluates to non-positive value 0


rjd15372 (approved):
LGTM!

ranshid (approved):

PR #1787, opened 2025-02-26
Initial draft of contributing guide

### LINKED ISSUE #1558: Create development guide

madolson:
Took the suggestions from the development guide issue and tried to write them up. The one thing still missing is the backporting, since it seems we aligned on creating a bot, but we don't have the bot yet.

Fixes: https://github.com/valkey-io/valkey/issues/1558

### DIFF
--- a/DEVELOPMENT_GUIDE.md
+++ b/DEVELOPMENT_GUIDE.md
@@ -34,11 +34,12 @@ There isn't a well defined test for what is considered a material change, but a 
 Valkey uses two types of tests: unit and integration tests.
 All contributions should include a test of some form. 
 
-Unit tests are present in the src/unit directory, and are intended to test individual structures or files.
+Unit tests are present in the `src/unit` directory, and are intended to test individual structures or files.
 For example, most changes to datastructures should include corresponding unit tests.
 
-Integration tests are located in the tst/ directory, and are intended to test end-to-end functionality.
+Integration tests are located in the `tests/` directory, and are intended to test end-to-end functionality.
 Adding new commands should come with corresponding integration tests.
+When writing cluster mode tests, do not use the legacy `tests/cluster` framework, which has been deprecated.
 
 ## Best practices
 1. Avoid adding configuration when a feature can be fully controlled by heuristics. 
--- a/DEVELOPMENT_GUIDE.md
+++ b/DEVELOPMENT_GUIDE.md
@@ -1,0 +1,49 @@
+# Valkey development guidelines
+This document provides an general overview for writing and designing code for Valkey.
+During our long development history, we've made a lot of inconsistent decisions, but we strive to get incrementally better.
+
+## General style guidelines
+Most of the style guidelines are enforced by clang format, but some additional comments are included here.
+
+1. C style comments `/* comment */` can be used for both single and multi-line comments. C++ comments `//` can only be used for single line comments.
+1. Generally keep line lengths below 90 characters, however there is no explicit line length enforcement.
+1. Use static functions when a function is only intended to be accessed from the same file.
+
+## Naming conventions
+Valkey has a long history of inconsistent naming conventions. Generally follow the style of the surrounding code, but you can also always use the following conventions for variable and structure names:
+
+- Variable names: `snake_case` or all lower case (e.g. `valkeyobject` or `valkey_object`)
+- Function names: `camelCase` or `namespace_camelCase` (e.g. `createObjectList` or `networking_createObjectList`).
+- Macros: `UPPER_CASE` (e.g. `DICT_CREATE`)
+- Structures: `camelCase` (e.g. `user`)
+
+## Licensing information
+When creating new source code files, use the following snippet to indicate the license:
+```
+/*
+ * Copyright (c) Valkey Contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+```
+
+If you are making material changes to a file that has a different license at the top, also add the above license snippet.
+There isn't a well defined test for what is considered a material change, but a good rule of thumb is if it's more than 100 lines of code.
+
+## Test coverage
+Valkey uses two types of tests: unit and integration tests.
+All contributions should include a test of some form. 
+
+Unit tests are present in the src/unit directory, and are intended to test individual structures or files.
+For example, most changes to datastructures should include corresponding unit tests.
+
+Integration tests are located in the tst/ directory, and are intended to test end-to-end functionality.
+Adding new commands should come with corresponding integration tests.
+
+## Best practices
+1. Avoid adding configuration when a feature can be fully controlled by heuristics. 
+We want Valkey to work correctly out of the box without much tuning.
+Configurations can be added to provide additional tuning of features. 
+When the workload characteristics can't be inferred or imply a tradeoff (CPU vs memory), then provide a configuration.
+2. Try to limit the number of lines changed in a PR when possible.
+We do a lot of backporting as a project, and the more lines changed, the higher the chance of having to resolve merge conflicts.
--- a/DEVELOPMENT_GUIDE.md
+++ b/DEVELOPMENT_GUIDE.md
@@ -12,7 +12,7 @@ Most of the style guidelines are enforced by clang format, but some additional c
 ## Naming conventions
 Valkey has a long history of inconsistent naming conventions. Generally follow the style of the surrounding code, but you can also always use the following conventions for variable and structure names:
 
-- Variable names: `snake_case` or all lower case (e.g. `valkeyobject` or `valkey_object`)
+- Variable names: `snake_case` or all lower case for short names  (e.g. `valkey_object` or `keylen`)
 - Function names: `camelCase` or `namespace_camelCase` (e.g. `createObjectList` or `networking_createObjectList`).
 - Macros: `UPPER_CASE` (e.g. `DICT_CREATE`)
 - Structures: `camelCase` (e.g. `user`)
--- a/DEVELOPMENT_GUIDE.md
+++ b/DEVELOPMENT_GUIDE.md
@@ -7,7 +7,7 @@ Most of the style guidelines are enforced by clang format, but some additional c
 
 1. C style comments `/* comment */` can be used for both single and multi-line comments. C++ comments `//` can only be used for single line comments.
 1. Generally keep line lengths below 90 characters when reasonable, however there is no explicit line length enforcement.
-   Use your best judgement for readabilty.
+   Use your best judgement for readability.
 1. Use static functions when a function is only intended to be accessed from the same file.
    For historical reasons, some private functions are prefixed by `_`, and they are kept as is to make it easier to backport changes.
 
@@ -31,7 +31,7 @@ When creating new source code files, use the following snippet to indicate the l
 ```
 
 If you are making material changes to a file that has a different license at the top, also add the above license snippet.
-There isn't a well defined test for what is considered a material change, but a good rule of thumb is that material changes are are more than 100 lines of code.
+There isn't a well defined test for what is considered a material change, but a good rule of thumb is that material changes are more than 100 lines of code.
 
 ## Test coverage
 Valkey uses two types of tests: unit and integration tests.
@@ -51,7 +51,7 @@ Valkey keeps most of the user documentation in the [valkey-doc](https://github.c
    Command history is also documented in the [command json file](https://github.com/valkey-io/valkey/tree/unstable/src/commands). 
 3. Server info fields are documented in the [info](https://github.com/valkey-io/valkey-doc/blob/main/commands/info.md) command.
 
-When a PR is opened that requires documention to be updated, the `needs-doc-pr` should be added until the corresponding documentation PR is open.
+When a PR is opened that requires documentation to be updated, the `needs-doc-pr` should be added until the corresponding documentation PR is open.
 
 ## Best practices
 1. Avoid adding configuration when a feature can be fully controlled by heuristics. 
@@ -60,4 +60,4 @@ Configurations can be added to provide additional tuning of features.
 When the workload characteristics can't be inferred or imply a tradeoff (CPU vs memory), then provide a configuration.
 2. Try to limit the number of lines changed in a PR when possible.
 We do a lot of backporting as a project, and the more lines changed, the higher the chance of having to resolve merge conflicts.
-Please seperate refactoring and functional changes into separate PRs, to make it easier to handle backporting.
+Please separate refactoring and functional changes into separate PRs, to make it easier to handle backporting.
--- a/DEVELOPMENT_GUIDE.md
+++ b/DEVELOPMENT_GUIDE.md
@@ -16,7 +16,7 @@ Most of the style guidelines are enforced by clang format, but some additional c
 
 1. C style comments `/* comment */` can be used for both single and multi-line comments.
    C++ comments `//` can only be used for single line comments.
-   Multi line comments should have the leading `*` in aligned and the final `*/` should be on the same line as the last line of text.
+   Multi line comments should have the leading `*` align and the final `*/` should be on the same line as the last line of text.
    e.g.
 ```c
 /* Blah Blah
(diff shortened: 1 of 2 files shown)


hwware (approved):

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,50 @@
+# Valkey development guidelines
+This document provides an general overview for writing and designing code for Valkey.
+During our long development history, we've made a lot of inconsistent decisions, but we strive to get incrementally better.
+
+## General style guidelines
+Most of the style guidelines are enforced by clang format, but some additional comments are included here.
+
+1. C style comments `/* comment */` can be used for both single and multi-line comments. C++ comments `//` can only be used for single line comments.
+1. Generally keep line lengths below 90 characters, however there is no explicit line length enforcement.
```
PingXie: I believe we landed on 120'ish previously but I am personally more in favor of 90 .

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,50 @@
...
+1. Use static functions when a function is only intended to be accessed from the same file.
```
PingXie, line 30: should we also outlaw the "_" prefix explicitly now? We can fix the existing ones gradually. 

@zuiderkwast

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,50 @@
...
+
+## Naming conventions
+Valkey has a long history of inconsistent naming conventions. Generally follow the style of the surrounding code, but you can also always use the following conventions for variable and structure names:
+
+- Variable names: `snake_case` or all lower case (e.g. `valkeyobject` or `valkey_object`)
```
PingXie: can we converge on `snake_case` exclusively, for the new code?

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,50 @@
...
+
+If you are making material changes to a file that has a different license at the top, also add the above license snippet.
+There isn't a well defined test for what is considered a material change, but a good rule of thumb is if it's more than 100 lines of code.
+
+## Test coverage
+Valkey uses two types of tests: unit and integration tests.
+All contributions should include a test of some form. 
+
+Unit tests are present in the `src/unit` directory, and are intended to test individual structures or files.
+For example, most changes to datastructures should include corresponding unit tests.
```
PingXie:
```suggestion
For example, most changes to data structures should include corresponding unit tests.
```

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,50 @@
...
+
+Integration tests are located in the `tests/` directory, and are intended to test end-to-end functionality.
+Adding new commands should come with corresponding integration tests.
+When writing cluster mode tests, do not use the legacy `tests/cluster` framework, which has been deprecated.
+
```
PingXie: can we add a documentation section as well? new commands, configs, non-trivial features should come with documentation? blogs are also encouraged for major features, such as the 1M QPS, the hash table rewrite

PingXie (commented):
LGTM overall.

DEVELOPMENT_GUIDE.md
madolson: The original decision was prefer less than 90, but have a strong reason to have less than 120.

DEVELOPMENT_GUIDE.md
madolson: Sure

DEVELOPMENT_GUIDE.md
madolson: I generally don't like, "For new code use X", because than it's confusing if someone is trying to conform to the existing standard. Either we should enforce it everywhere I think we should be clear about A or B.

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,50 @@
+# Valkey development guidelines
+This document provides an general overview for writing and designing code for Valkey.
```
madolson:
```suggestion
This document provides a general overview for writing and designing code for Valkey.
```

DEVELOPMENT_GUIDE.md
madolson: Yeah, I'll add a blurb here.

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,50 @@
...
+ * Copyright (c) Valkey Contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
+```
+
+If you are making material changes to a file that has a different license at the top, also add the above license snippet.
+There isn't a well defined test for what is considered a material change, but a good rule of thumb is if it's more than 100 lines of code.
+
+## Test coverage
```
sarthakaggarwal97, line 57: is there a documentation that mentions how can a developer run some tests? single tests? loop through tests etc

sarthakaggarwal97: If not, should we mention this here?

roshkhatri: It would be in the `tests/README` IIRC.
And I think the options for `runtest` like loop, should be added in `tests/README`

sarthakaggarwal97: yeah, makes sense! I can probably spend some time and improve it separately.

sarthakaggarwal97 (approved):

DEVELOPMENT_GUIDE.md
zuiderkwast: I like the "follow surrounding code" guideline.

The all-lower is common in standard C function names like strlen, memcpy, etc. I think it's OK to use this style for short names like keylen, objptr if surrounding code uses this style, but it becomes hard to read in longer names like valkeyobject.

```suggestion
- Variable names: `snake_case` or all lower case for short names  (e.g. `valkey_object` or `keylen`)
```

madolson:
@valkey-io/core-team Do we have further thoughts on this, I think it's mostly up to date. I want to add some more things, but would rather do those all incrementally.

enjoy-binbin (approved):

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,63 @@
...
+This document provides a general overview for writing and designing code for Valkey.
+During our long development history, we've made a lot of inconsistent decisions, but we strive to get incrementally better.
+
+## General style guidelines
+Most of the style guidelines are enforced by clang format, but some additional comments are included here.
+
+1. C style comments `/* comment */` can be used for both single and multi-line comments. C++ comments `//` can only be used for single line comments.
+1. Generally keep line lengths below 90 characters when reasonable, however there is no explicit line length enforcement.
+   Use your best judgement for readability.
+1. Use static functions when a function is only intended to be accessed from the same file.
```
enjoy-binbin:
```suggestion
1. C style comments `/* comment */` can be used for both single and multi-line comments. C++ comments `//` can only be used for single line comments.
2. Generally keep line lengths below 90 characters when reasonable, however there is no explicit line length enforcement.
   Use your best judgement for readability.
3. Use static functions when a function is only intended to be accessed from the same file.
```

enjoy-binbin (approved):
LGTM.

hpatro (commented):
Good starting point.

Maybe we can put up our own AGENTS.md or rename this new file as it would encompass most of the DEVELOPMENT_GUIDE.md. Humans could also scroll through that if they need some guidance. Seems like the way forward atm.

madolson:
AGENTS.md are typically supposed to have a more terse format (to limit token usage). "DO X, DON'T do Y", right now this file is trying to also explain why we are doing something. We could have an agent re-write this file into an AGENTS.md

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,63 @@
+# Valkey development guidelines
+This document provides a general overview for writing and designing code for Valkey.
+During our long development history, we've made a lot of inconsistent decisions, but we strive to get incrementally better.
+
+## General style guidelines
+Most of the style guidelines are enforced by clang format, but some additional comments are included here.
+
+1. C style comments `/* comment */` can be used for both single and multi-line comments. C++ comments `//` can only be used for single line comments.
```
ranshid: We also start multilines with a * (and clang formatter will kill you in case you leave a space after the closing '*/' IIRC) also I recall we once kept that a multiline comment is always ending with the '*/' on the same line and not on a new line like:
```
/* blh blh blh blh
 * blh blh blh blh */
```

and not:

```
/* blh blh blh blh
 * blh blh blh blh 
 */
```

DEVELOPMENT_GUIDE.md
ranshid: >Use static functions when a function is only intended to be accessed from the same file.

when the function is not declared in any header file, should it not be made static? this is somewhat what we require today...

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,63 @@
+# Valkey development guidelines
+This document provides a general overview for writing and designing code for Valkey.
+During our long development history, we've made a lot of inconsistent decisions, but we strive to get incrementally better.
+
+## General style guidelines
```
ranshid, line 14: lets place a comment about prefer use of `bool` over `int` in new added code.

ranshid (approved):
It is a good start IMO.
I can think of many things to extend this like:
how to choose the correct configuration type
how to correctly modify/write json files for commands
integration tests best practices (eg use wait_fo instead of after...)
writing architecture specific defines where, when and how...
writing os/compiler specific defines where, when and how... (ie config.h)
and probably more we can add in the future.

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,63 @@
...
+   For historical reasons, some private functions are prefixed by `_`, and they are kept as is to make it easier to backport changes.
+
+## Naming conventions
+Valkey has a long history of inconsistent naming conventions.
+Generally follow the style of the surrounding code, but you can also always use the following conventions for variable and structure names:
+
+- Variable names: `snake_case` or all lower case for short names  (e.g. `valkey_object` or `keylen`)
+- Function names: `camelCase` or `namespace_camelCase` (e.g. `createObjectList` or `networking_createObjectList`).
+- Macros: `UPPER_CASE` (e.g. `DICT_CREATE`)
+- Structures: `camelCase` (e.g. `user`)
```
zuiderkwast: We only use the namespace form in certain areas where it's used concistenly. The `networking` namespace is not one of these, so I don't think it's a good example.

I don't think we should use "valkey" as an example as we've been trying to avoid trademarks in code. We've used "server" instead for many server things.

Let's use examples of existing functions/variables/macros instead of made up ones.

```suggestion
- Variable names: `snake_case` or all lower case for short names  (e.g. `cached_reply` or `keylen`).
- Function names: `camelCase` or `namespace_camelCase` (e.g. `createStringObject` or `IOJobQueue_isFull`).
- Macros: `UPPER_CASE` (e.g. `MAKE_CMD`).
- Structures: `camelCase` (e.g. `user`).
```

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,63 @@
...
+3. Server info fields are documented in the [info](https://github.com/valkey-io/valkey-doc/blob/main/commands/info.md) command.
+
+When a PR is opened that requires documentation to be updated, the `needs-doc-pr` should be added until the corresponding documentation PR is open.
+
+## Best practices
+1. Avoid adding configuration when a feature can be fully controlled by heuristics. 
+We want Valkey to work correctly out of the box without much tuning.
+Configurations can be added to provide additional tuning of features. 
+When the workload characteristics can't be inferred or imply a tradeoff (CPU vs memory), then provide a configuration.
+2. Try to limit the number of lines changed in a PR when possible.
```
zuiderkwast: This is the most important guideline of all. 😄 We could place near the top of the page rather than at the bottom.

zuiderkwast (commented):
There are probably many more unwritten conventions in our subconcious minds but let's start with these and get it merged.

DEVELOPMENT_GUIDE.md
zuiderkwast: Yeah, and some information about when to use comments and when not. For example:

Use commends to explain why something is done the way it is, when it's not clear otherwise.

Don't use comments to describe what the code is doing. That should be clear from the code itself.

DEVELOPMENT_GUIDE.md
madolson: I'll move best practices and other stuff up to the top!

DEVELOPMENT_GUIDE.md
madolson: Maybe a preference of mine, but I normally use numbered lists with all 1s (since markdown renders them correctly). That way a new item can be added without having to modify subsequent lines if it wasn't added to the end.

DEVELOPMENT_GUIDE.md
ranshid: and?

madolson: lol, I was still thinking about it. I don't know if I care if you use bool or not.

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,73 @@
...
+ * Blah Blah. */
+   ```
+2. Comments should generally be used to describe behavior that is not obvious from reading the code itself.
+   This includes complex behavior, why code was written the way it was and describing non-obvious behavior.
+   Additionally, functions should be documented to explain all of the function's behavior without having to read the code.
+1. Generally keep line lengths below 90 characters when reasonable, however there is no explicit line length enforcement.
+   Use your best judgement for readability.
+1. Use static functions when a function is only intended to be accessed from the same file.
+   For historical reasons, some private functions are prefixed by `_`, and they are kept as is to make it easier to backport changes.
+
```
madolson:
```suggestion
   For historical reasons, some private functions are prefixed by `_`, and they are kept as is to make it easier to backport changes.
1. Use the boolean type for true/false values.
   For historical reasons, some functions used the integer type, and they are kept as is to make it easier to backport changes.
```

DEVELOPMENT_GUIDE.md
```diff
@@ -0,0 +1,74 @@
...
+2. Comments should generally be used to describe behavior that is not obvious from reading the code itself.
+   This includes complex behavior, why code was written the way it was and describing non-obvious behavior.
+   Additionally, functions should be documented to explain all of the function's behavior without having to read the code.
+1. Generally keep line lengths below 90 characters when reasonable, however there is no explicit line length enforcement.
+   Use your best judgement for readability.
+1. Use static functions when a function is only intended to be accessed from the same file.
+   For historical reasons, some private functions are prefixed by `_`, and they are kept as is to make it easier to backport changes.
+1. Use the boolean type for true/false values.
+   For historical reasons, some functions used the integer type, and they are kept as is to make it easier to backport changes.
+## Naming conventions
```
madolson, line 35:
```suggestion
   For historical reasons, some functions used the integer type, and they are kept as is to make it easier to backport changes.

## Naming conventions
```

DEVELOPMENT_GUIDE.md
madolson: Ok, wrote some guidance.

PR #1791, opened 2025-02-27
Adding json support for log-format config

### LINKED ISSUE #1006: [NEW] Output logs as JSON

jbergstroem:
Introduce a json log format that can be invoked by passing `log-format json` in your config.

We also refactor `escapeJsonString` to `util` since it is now a shared function.

Closes: https://github.com/valkey-io/valkey/issues/1006

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -50,6 +50,7 @@
 #include "sds.h"
 #include "module.h"
 #include "scripting_engine.h"
+#include "util.h"
 
 #include "eval.h"
 
--- a/src/server.c
+++ b/src/server.c
@@ -182,7 +182,6 @@ void serverLogRaw(int level, const char *msg) {
     const char *verbose_level[] = {"debug", "info", "notice", "warning"};
     const char *roles[] = {"sentinel", "RDB/AOF", "replica", "primary"};
     const char *role_chars = "XCSM";
-    const char *json_format = "{\"pid\":%d,\"role\":\"%s\",\"timestamp\":\"%s\",\"level\":\"%s\",\"message\":\"%s\"}\n";
     const char *logfmt_format = "pid=%d role=%s timestamp=\"%s\" level=%s message=\"%s\"\n";
     FILE *fp;
     char buf[64];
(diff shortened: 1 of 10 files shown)


jbergstroem:
~The DCO action is confusing:~
> #### Summary
>
> Commit sha: [d8edd52](https://github.com/valkey-io/valkey/pull/1791/commits/d8edd5220a3fcfecf7a0590f40c4693664b6e3a0), Author: Johan Bergström, Committer: Johan Bergström; Expected "Johan Bergström bugs@bergstroem.nu", but got "Johan Bergström [bugs@bergstroem.nu](mailto:bugs@bergstroem.nu)".

Edit: Fixed. I think it was complaining about using the wrong representation of `ö` in my last name.

zuiderkwast:
Nice to see this has been started.

We need to escape the strings. There's some code for that in cli_common.c:

    sds escapeJsonString(sds s, const char *p, size_t len);

We could move it to another file like util.c, since it will no longer be only for the CLI.

jbergstroem:
> Nice to see this has been started.
> 
> We need to escape the strings. There's some code for that in cli_common.c:
> 
> ```
> sds escapeJsonString(sds s, const char *p, size_t len);
> ```
> 
> We could move it to another file like util.c, since it will no longer be only for the CLI.

Makes sense. Would be much more comfortable if I could test this as we go. Should we start a unit test for basic functions?

zuiderkwast:
> Would be much more comfortable if I could test this as we go. Should we start a unit test for basic functions?

Yes, we can do that. Another option is to add an integration test where do something that causes something to be logged and then verify what's logged in the log file.

Grep for `verify_log_message` to find tests that already do this.

We can use the command `DEBUG LOG message` to write to the log. The server needs to be started with `--enable-debug-command yes` but that's enabled in the tests already.

jbergstroem:
@zuiderkwast long time since last update -- apologies; life came in-between. I've done the changes you requested (refactor, add tests). Will make it ready for review. Finally – if there is places that lacks documentation, feel free to point me to them. Thanks for any feedback!

src/server.c
```diff
@@ -237,14 +238,20 @@ void serverLogRaw(int level, const char *msg) {
...
-                        buf, verbose_level[level], msg);
+                fprintf(fp, logfmt_format, (int)getpid(), roles[role_index], buf, verbose_level[level], msg);
             }
             break;
 
+        case LOG_FORMAT_JSON: {
+            sds jsonmsg = escapeJsonString(sdsempty(), msg, strlen(msg));
+            fprintf(fp, "{\"pid\":%d,\"role\":\"%s\",\"timestamp\":\"%s\",\"level\":\"%s\",\"message\":%s}\n",
+                    (int)getpid(), roles[role_index], buf, verbose_level[level], jsonmsg);
+            sdsfree(jsonmsg);
```
dvkashapov, line 252: There's allocation and freeing of sds for every log line, should we do something about this? Maybe it is a good idea to have some thread local buffer to be reused every time?

src/server.c
dvkashapov, line 1: There's no include of utils in server.c, let's explicitly `#include "util.h"`

src/server.c
zuiderkwast: I think the temporary allocation is OK, because

1. Logging is not in the very hot code path. It's not like command execution.
2. The allocator will most likely give you memory that was recently freed and is alreaydy in CPU cache.

tests/integration/logging.tcl
```diff
@@ -144,4 +144,14 @@ if {$backtrace_supported} {
     }
 }
 
+# test JSON log format with special characters
+start_server {overrides {log-format json}} {
+    test {JSON log format properly escapes special characters} {
+        set log_lines [count_log_lines 0]
+
+        r debug log "Test \"quotes\" and \\backslash\\ and newline"
```
zuiderkwast: Newline? There's no newline within the logged message here. (If you want to test it, use "\n" or we can just skip it.)

```suggestion
        r debug log "Test \"quotes\" and \\backslash\\"
```

valkey.conf
```diff
@@ -425,6 +425,7 @@ loglevel notice
 #
 # - legacy: the default, traditional log format
 # - logfmt: a structured log format; see https://www.brandur.org/logfmt
+# - json: a structured log format; see https://json.org/
```
zuiderkwast: I believe everyone is familiar with JSON. No need for the link.

```suggestion
# - json: a structured log format
```

zuiderkwast (approved):
LGTM

@valkey-io/core-team When we decided to add the *logfmt* format in https://github.com/valkey-io/valkey/pull/1022#issuecomment-2395532968, it appears we also already agreed to add the JSON log format. Please comment if you don't agree.

enjoy-binbin (approved):
LGTM.

zuiderkwast:
Thanks Johan!

PR #1793, opened 2025-02-27
Replaced the use of sdsReplyDictType and hashDictType.

### LINKED ISSUE #1550: Follow-up of #1502: Replace temp dicts with hashtable
Let's replace these temporary dicts with a hashtables:

* sdsReplyDictType
  * (_Originally posted by @zuiderkwast in https://github.com/valkey-io/valkey/pull/1502#discussion_r1908915484_)
* hashDictType
  * (_Originally posted by @zuiderkwast in https://github.com/valkey-io/valkey/pull/1502#discussion_r1908916604_)

zarkash-aws:
This PR resolves #1550 by replacing the use of temporary dict structures with hashtable for random elements selection. 

hrandfieldWithCountCommand:

CASE 3: Replaced sdsReplyDictType with sdsReplyHashtableType to store a pointer to the entry. 
CASE 4: Replaced hashDictType with setHashtableType, as only the field needs to be stored.

zrandmemberWithCountCommand:
used similar logic to replace the use of hashDictType and sdsReplyDictType in CASE 3.

RDB load skiplist field duplication validation logic:
Replace use of dict for storing listpack/hash loaded fields when detecting duplications.

### DIFF
--- a/src/t_hash.c
+++ b/src/t_hash.c
@@ -1272,7 +1272,7 @@ void hrandfieldWithCountCommand(client *c, long l, int withvalues) {
             addWritePreparedReplyBulkSds(wpc, sdsdup(field));
             if (withvalues) addWritePreparedReplyBulkSds(wpc, sdsdup(value));
         }
-        
+
         hashtableResetIterator(&iter);
         hashtableRelease(ht);
     }
--- a/src/t_hash.c
+++ b/src/t_hash.c
@@ -1239,50 +1239,50 @@ void hrandfieldWithCountCommand(client *c, long l, int withvalues) {
      * used into CASE 4 is highly inefficient. */
     if (count * HRANDFIELD_SUB_STRATEGY_MUL > size) {
         /* Hashtable encoding (generic implementation) */
-        dict *d = dictCreate(&sdsReplyDictType);
-        dictExpand(d, size);
+        hashtable *ht = hashtableCreate(&hashHashtableType);
+        hashtableExpand(ht, size);
         hashTypeIterator hi;
         hashTypeInitIterator(hash, &hi);
 
-        /* Add all the elements into the temporary dictionary. */
-        while ((hashTypeNext(&hi)) != C_ERR) {
-            int ret = DICT_ERR;
+        /* Add all the elements into the temporary hashtable. */
+        while (hashTypeNext(&hi) != C_ERR) {
+            int ret = 0;
             sds field, value = NULL;
 
             field = hashTypeCurrentObjectNewSds(&hi, OBJ_HASH_FIELD);
             if (withvalues) value = hashTypeCurrentObjectNewSds(&hi, OBJ_HASH_VALUE);
-            ret = dictAdd(d, field, value);
-
-            serverAssert(ret == DICT_OK);
+            hashTypeEntry *entry = hashTypeCreateEntry(field, value);
+            sdsfree(field);
+            ret = hashtableAdd(ht, entry);
+            serverAssert(ret);
         }
-        serverAssert(dictSize(d) == size);
+        serverAssert(hashtableSize(ht) == size);
         hashTypeResetIterator(&hi);
 
         /* Remove random elements to reach the right count. */
         while (size > count) {
-            dictEntry *de;
-            de = dictGetFairRandomKey(d);
-            dictUnlink(d, dictGetKey(de));
-            sdsfree(dictGetKey(de));
-            sdsfree(dictGetVal(de));
-            dictFreeUnlinkedEntry(d, de);
+            void *element;
+            hashtableFairRandomEntry(ht, &element);
+            sds field = hashTypeEntryGetField((hashTypeEntry*)element);
+            hashtableDelete(ht, field);
             size--;
         }
 
-        /* Reply with what's in the dict and release memory */
-        dictIterator *di;
-        dictEntry *de;
-        di = dictGetIterator(d);
-        while ((de = dictNext(di)) != NULL) {
-            sds field = dictGetKey(de);
-            sds value = dictGetVal(de);
+        /* Reply with what's in the temporary hashtable and release memory */
+        hashtableIterator iter;
+        hashtableInitIterator(&iter, ht, 0);
+        void *next;
+        while (hashtableNext(&iter, &next)) {
+            hashTypeEntry* entry = (hashTypeEntry*)next;
+            sds field = hashTypeEntryGetField(entry);
+            sds value = hashTypeEntryGetValue(entry);
             if (withvalues && c->resp > 2) addWritePreparedReplyArrayLen(wpc, 2);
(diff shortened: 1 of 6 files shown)


zarkash-aws:
update - 
In CASE 3: Changed hashtable type from `hashHashtableType` to `sdsReplyHashtableType`. Updated hashtable to store a pointer to the hashtype entries instead of storing entries directly. This simplifies handling of the 'withvalue' case, as it no longer requires explicitly providing a value to create an entry.

src/t_hash.c
```diff
@@ -1239,50 +1239,42 @@ void hrandfieldWithCountCommand(client *c, long l, int withvalues) {
...
-
-            field = hashTypeCurrentObjectNewSds(&hi, OBJ_HASH_FIELD);
-            if (withvalues) value = hashTypeCurrentObjectNewSds(&hi, OBJ_HASH_VALUE);
-            ret = dictAdd(d, field, value);
-
-            serverAssert(ret == DICT_OK);
+        /* Add all the elements into the temporary hashtable. */
+        while (hashTypeNext(&hi) != C_ERR) {
+            int ret = 0;
+            ret = hashtableAdd(ht, (&hi)->next);
```
xbasel: why not 
`
ret = hashtableAdd(ht, hi.entry);
`
?

This code can be reduce to:
```
        while (hashTypeNext(&hi) != C_ERR) {
            serverAssert(hashtableAdd(ht, hi.next));
        }
```

src/t_hash.c
```diff
@@ -1239,50 +1239,42 @@ void hrandfieldWithCountCommand(client *c, long l, int withvalues) {
...
+        hashtableIterator iter;
+        hashtableInitIterator(&iter, ht, 0);
+        void *next;
+        while (hashtableNext(&iter, &next)) {
+            sds field = hashTypeEntryGetField(next);
+            sds value = hashTypeEntryGetValue(next);
             if (withvalues && c->resp > 2) addWritePreparedReplyArrayLen(wpc, 2);
-            addWritePreparedReplyBulkSds(wpc, field);
-            if (withvalues) addWritePreparedReplyBulkSds(wpc, value);
+            addWritePreparedReplyBulkSds(wpc, sdsdup(field));
```
xbasel: Consider removing addWritePreparedReplyBulkSds and use addReplyBulkCBuffer instead, no sdsdup is needed

zarkash-aws:
removed the function addWritePreparedReplyBulkSds, as it is no longer in use.

src/t_hash.c
```diff
@@ -1239,50 +1239,40 @@ void hrandfieldWithCountCommand(client *c, long l, int withvalues) {
...
-            sds field, value = NULL;
-
-            field = hashTypeCurrentObjectNewSds(&hi, OBJ_HASH_FIELD);
-            if (withvalues) value = hashTypeCurrentObjectNewSds(&hi, OBJ_HASH_VALUE);
-            ret = dictAdd(d, field, value);
-
-            serverAssert(ret == DICT_OK);
+        /* Add all the elements into the temporary hashtable. */
+        while (hashTypeNext(&hi) != C_ERR) {
+            serverAssert(hashtableAdd(ht, (&hi)->next));
```
zuiderkwast: `hi.next` looks like it's the next element but it's actually the current element. Maybe we should rename the `void *next` iterator field to `hashTypeEntry *currentHashtableEntry`. It might make the code a bit more clear.

Or maybe better: we could treat it as internals of the iterator and not access it directly here. The hashTypeIterator is supposed to work regardless of encoding and here rely on it being a hashtable. But since we know it's a hashtable, we can use a hashtableIterator directly instead of a hashTypeIterator:

```C
        hashtableIterator iter;
        hashtableInitIterator(&iter, hash->ptr, 0);
        void *entry;
        while (hashtableNext(&iter, &entry)) {
            hashtableAdd(ht, entry);
        }
        hashtableResetIterator(&iter);
```

Regarding the assert: In general, it's not good practice to do work inside asserts, because if we ever want to compile the code with assert disabled, the entry will not be added. I know we don't follow this very well, but still. We can just check the result in the assert, like

```C
            int added = hashtableAdd(ht, entry);
            serverAssert(added);
```

src/t_hash.c
```diff
@@ -1239,50 +1239,40 @@ void hrandfieldWithCountCommand(client *c, long l, int withvalues) {
...
+        hashtableInitIterator(&iter, ht, 0);
+        void *next;
+        while (hashtableNext(&iter, &next)) {
+            sds field = hashTypeEntryGetField(next);
+            sds value = hashTypeEntryGetValue(next);
             if (withvalues && c->resp > 2) addWritePreparedReplyArrayLen(wpc, 2);
-            addWritePreparedReplyBulkSds(wpc, field);
-            if (withvalues) addWritePreparedReplyBulkSds(wpc, value);
+            addReplyBulkCBuffer(c, field, sdslen(field));
+            if (withvalues) addReplyBulkCBuffer(c, value, sdslen(value));
```
zuiderkwast: The "WritePrepared" part of `addWritePreparedReplyBulkSds` is an optimization that was made so we don't have to check if the client is prepared every time. There is a CBuffer variant of this function as well. Let's use it so we don't drop this optimization.

```suggestion
            addWritePreparedReplyBulkCBuffer(wpc, field, sdslen(field));
            if (withvalues) addWritePreparedReplyBulkCBuffer(wpc, value, sdslen(value));
```

zuiderkwast (commented):
Looks good! Just some minor comments.

zuiderkwast:
One more thing: I don't see that sdsReplyDictType and hashDictType are deleted. One of the points of this refactoring is to be able to delete some code.

ranshid:
> One more thing: I don't see that sdsReplyDictType and hashDictType are deleted. One of the points of this refactoring is to be able to delete some code.

I think @zarkash-aws only handled the hash type case. we can also apply the logic for the zset case and remove these functions. @zarkash-aws please update the PR once you have this implemented

zuiderkwast:
> I think @zarkash-aws only handled the hash type case. we can also apply the logic for the zset case and remove these functions.

Ok, I didn't realize. We can merge this PR first to keep the PR small? Or add the zset stuff too, as you want. I don't mind.

zuiderkwast (approved):
Looks good to me.

Ran, any comments? Let's merge it?

ranshid:
> Looks good to me.
> 
> Ran, any comments? Let's merge it?

Lets have @zarkash-aws complete the entire implementation as supporting zset is not that complicated IMO.

zarkash-aws:
completed the implementation for zset & rdb.c, and removed hashDictType and sdsReplyDictType.

zuiderkwast (approved):
Approved again. :)

Please update the PR's top comment to mention not only hrandfieldWithCountCommand. Also mention sorted sets commands and rdb. It doesn't need to explain any details, just mention.

ranshid (approved):
LGTM

PR #1798, opened 2025-02-27
Adds a memory leak check after running a unit test

rjd15372:
This commit adds check after each test function execution of a unit test, that checks if there is still memory allocated.

This check prevents situations where tests, which do these kind of checks, fail due to memory leaks caused by previous tests.

### DIFF
--- a/src/unit/test_main.c
+++ b/src/unit/test_main.c
@@ -29,19 +29,18 @@ int runTestSuite(struct unitTestSuite *test, int argc, char **argv, int flags) {
     for (int id = 0; test->tests[id].proc != NULL; id++) {
         test_num++;
         int test_result = (test->tests[id].proc(argc, argv, flags) != 0);
-
-        /* Check if the test has cleaned up all the memory used. */
-        if (zmalloc_used_memory() > 0) {
-            printf("[" KRED "%s - %s" KRESET "] Memory leak detected of %zu bytes\n", test->tests[id].name, test->filename, zmalloc_used_memory());
-            test_result = 1;
-        }
-
         if (!test_result) {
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):

madolson (approved):

madolson:
Seems to be a lot of memory leaks though :/

zuiderkwast:
> Seems to be a lot of memory leaks though :/

Hm... Are the test cases not independent? If not, then maybe we should just detect leaks between each suite...?

madolson:
Yeah, it looks like the listpack tests are cumulative, which is interesting. We should probably make each one self contained, so we can run it without the others.

zuiderkwast:
> Yeah, it looks like the listpack tests are cumulative, which is interesting. We should probably make each one self contained, so we can run it without the others.

There's currently no way to run a single test case though. It'd be more work to refactor these tests so I suggest we just check mem leaks per suite. It's much easier.

madolson:
> There's currently no way to run a single test case though. It'd be more work to refactor these tests so I suggest we just check mem leaks per suite. It's much easier.

I'm okay with that for now.

enjoy-binbin (approved):

rjd15372:
@zuiderkwast @madolson I changed the implementation to just check for memory leak after running all tests of a suite.

If you think it's worth I can refactor the tests in a different PR to make each test run in isolation without depending on previous tests.

enjoy-binbin:
> If you think it's worth I can refactor the tests in a different PR to make each test run in isolation without depending on previous tests.

this seem a good idea, please go ahead, thanks.

But we also need to look at the time investment, as some of our use cases may indeed depend on each other.

PR #1800, opened 2025-02-27
Fix incorrect assertion in client list operations

uriyage:
The current assertion introduced in https://github.com/redis/redis/pull/11220/

```c
serverAssert(&c->clients_pending_write_node.next != NULL || &c->clients_pending_write_node.prev != NULL);
```

is incorrect for two reasons:

1. Using &pointer.next would always be non-NULL since it's the address of the field.
2. The check is incorrect even without the & because in a single-node list, both pointers can be NULL.

Fix:
1. Remove the always-true assertion
2. Add proper assertions in listUnlinkNode to ensure the node membership in the list to cover all list cases.

### DIFF
--- a/src/adlist.c
+++ b/src/adlist.c
@@ -32,7 +32,7 @@
 #include <stdlib.h>
 #include "adlist.h"
 
-#include "server.h"
+#include "serverassert.h"
 #include "zmalloc.h"
 
 /* Create a new list. The created list can be freed with
--- a/src/adlist.c
+++ b/src/adlist.c
@@ -31,6 +31,8 @@
 
 #include <stdlib.h>
 #include "adlist.h"
+
+#include "server.h"
 #include "zmalloc.h"
 
 /* Create a new list. The created list can be freed with
@@ -187,6 +189,7 @@ void listDelNode(list *list, listNode *node) {
  * Remove the specified node from the list without freeing it.
  */
 void listUnlinkNode(list *list, listNode *node) {
+    serverAssert(list->len > 0);
     if (node->prev)
         node->prev->next = node->next;
     else
--- a/src/adlist.c
+++ b/src/adlist.c
@@ -190,14 +190,22 @@ void listDelNode(list *list, listNode *node) {
  */
 void listUnlinkNode(list *list, listNode *node) {
     assert(list->len > 0);
-    if (node->prev)
+
+    if (node->prev) {
+        assert(node->prev->next == node);
         node->prev->next = node->next;
-    else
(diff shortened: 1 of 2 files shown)


src/adlist.c
```diff
@@ -31,6 +31,8 @@
 
 #include <stdlib.h>
 #include "adlist.h"
+
+#include "server.h"
```
zuiderkwast: I don't want this dependency. The datastructure is a small building block. It should be independend of server. The server can depend on it but not vice versa.

Additionally, we also use adlist in non-server things like valkey-cli and valkey-benchmark.

Let's include serverassert.h instead. It's minimal. (The name is unfortunate though because it's not about valkey-server specifically.)

```Suggestion
#include "serverassert.h"
```

src/adlist.c
```diff
@@ -187,6 +189,7 @@ void listDelNode(list *list, listNode *node) {
  * Remove the specified node from the list without freeing it.
  */
 void listUnlinkNode(list *list, listNode *node) {
+    serverAssert(list->len > 0);
```
zuiderkwast: It's just `assert` when we use `serverassert.h`.

```suggestion
    assert(list->len > 0);
```

zuiderkwast (commented):
Good! Small comment though.

src/adlist.c
uriyage: Thanks!

src/networking.c
```diff
@@ -1583,10 +1583,11 @@ void unlinkClient(client *c) {
 
     /* Remove from the list of pending writes if needed. */
     if (c->flag.pending_write) {
-        serverAssert(&c->clients_pending_write_node.next != NULL || &c->clients_pending_write_node.prev != NULL);
         if (c->io_write_state == CLIENT_IDLE) {
+            serverAssert(listSearchKey(server.clients_pending_write, c) != NULL);
```
enjoy-binbin: do we have some other ways to assert it? I am worried that this search operation will be slow

zuiderkwast: We can use debugServerAssert. It only enabled by the config `enable-debug-assert` and does nothing otherwise.

```suggestion
            debugServerAssert(listSearchKey(server.clients_pending_write, c) != NULL);
```

uriyage: Thanks, changed to debug assert.

uriyage: Found a better general way to assert node membership in the list using O(1) operation for all lists.

enjoy-binbin (approved):

PR #1801, opened 2025-02-27
Fix defrag when type/encoding changes during scan

rainsupreme:
I was hunting for defrag bugs with Jim and found a couple improvements to make. Jim pointed out that in several of the callbacks, if the encoding were to change it simply returns without doing anything to `cursor` to make it reach 0, meaning that it would continue no-op working on that item without making any progress. Type and encoding can change while the defrag scan is in progress if the value is mutated or replaced by something else with the same key.

### DIFF
--- a/src/defrag.c
+++ b/src/defrag.c
@@ -403,7 +403,6 @@ static long scanLaterList(robj *ob, unsigned long *cursor, monotime endtime) {
     quicklistNode *node;
     long iterations = 0;
     int bookmark_failed = 0;
-    if (ob->type != OBJ_LIST || ob->encoding != OBJ_ENCODING_QUICKLIST) return 0;
 
     if (*cursor == 0) {
         /* if cursor is 0, we start new iteration */
@@ -446,7 +445,6 @@ static void scanLaterZsetCallback(void *privdata, void *element_ref) {
 }
 
 static void scanLaterZset(robj *ob, unsigned long *cursor) {
-    if (ob->type != OBJ_ZSET || ob->encoding != OBJ_ENCODING_SKIPLIST) return;
     zset *zs = (zset *)ob->ptr;
     *cursor = hashtableScanDefrag(zs->ht, *cursor, scanLaterZsetCallback, zs->zsl, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
 }
@@ -460,7 +458,6 @@ static void scanHashtableCallbackCountScanned(void *privdata, void *elemref) {
 }
 
 static void scanLaterSet(robj *ob, unsigned long *cursor) {
-    if (ob->type != OBJ_SET || ob->encoding != OBJ_ENCODING_HASHTABLE) return;
     hashtable *ht = ob->ptr;
     *cursor = hashtableScanDefrag(ht, *cursor, activeDefragSdsHashtableCallback, NULL, activeDefragAlloc, HASHTABLE_SCAN_EMIT_REF);
 }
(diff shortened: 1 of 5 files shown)


src/defrag.c
```diff
@@ -398,7 +398,6 @@ static long scanLaterList(robj *ob, unsigned long *cursor, monotime endtime) {
     quicklistNode *node;
     long iterations = 0;
     int bookmark_failed = 0;
-    if (ob->type != OBJ_LIST || ob->encoding != OBJ_ENCODING_QUICKLIST) return 0;
 
     if (*cursor == 0) {
         /* if cursor is 0, we start new iteration */
```
zuiderkwast: I see you added a check before this function is called. Let's change these checks to asserts instead of simply deleting them.

If `ob->type` or `ob->encoding` isn't what we expect, then the `quicklist *ql = ob->ptr` variable assigned above isn't a quicklist and that can be very bad below...

src/rdb.c
```diff
@@ -2077,7 +2077,7 @@ robj *rdbLoadObject(int rdbtype, rio *rdb, sds key, int dbid, int *error) {
 
         /* Too many entries? Use a hash table right from the start. */
         if (len > server.hash_max_listpack_entries)
-            hashTypeConvert(o, OBJ_ENCODING_HASHTABLE);
+            hashTypeEnsureHashtableEncoded(o);
```
zuiderkwast: (This and many similar lines.)

I don't think we should do this refactoring for two reasons:

1. We want to avoid unnecessary changes. More touched lines means unnecessary risk of merge conflicts and harder to follow changes using git history and git blame.
2. Just because we currently only convert from listpack to hashtable, it doesn't mean we never will do any other conversion. We might want to convert back to listpack if the hashtable is small and we detect it during some iteration over it.

src/defrag.c
```diff
@@ -842,7 +834,7 @@ static int defragLaterItem(robj *ob, unsigned long *cursor, monotime endtime, in
             long long endtimeWallClock = ustime() + (endtime - getMonotonicUs());
             return moduleLateDefrag(sds_key_passed_as_robj, ob, cursor, endtimeWallClock, dbid);
         } else {
-            *cursor = 0; /* object type may have changed since we schedule it for later */
+            *cursor = 0; /* object type/encoding may have changed since we schedule it for later */
         }
     } else {
```
JimB123, line 850: This, and all the changes in `defrag.c` look good to me.  I'll defer to Viktor's comments re: the other refactoring.

src/defrag.c
JimB123: I like assertions.  If, for no other reason than they clearly document the designer's intent.

I agree that the code is now correct as written.  However, I also agree with Viktor.  We can make it perfectly clear that the function is only being called with the specific types by adding an assertion.

src/defrag.c
```diff
@@ -398,7 +398,6 @@ static long scanLaterList(robj *ob, unsigned long *cursor, monotime endtime) {
     quicklistNode *node;
     long iterations = 0;
     int bookmark_failed = 0;
-    if (ob->type != OBJ_LIST || ob->encoding != OBJ_ENCODING_QUICKLIST) return 0;
```
JimB123, line 406: While you're working on this file, please also fix `activeDefragSdsHashtableCallback()`.  It should be `static`.

src/defrag.c
rainsupreme: Fair enough! I like it as "self documenting code" but I don't know what kind of performance impact it incurs. I considered whether I could change the argument type to be more specific than `robj *` but set and hash don't have their own unique type.

JimB123 (approved):

src/defrag.c
zuiderkwast: The asserts have have "likely" hits to help branch prediction so I don't think it has any performance impact at all. If it does, it is not a very hot code path anyway.

zuiderkwast (approved):

PR #1802, opened 2025-02-28
Use the wrapper from cli_common instead of hiredis

pizhenwei:
There are mixed usage around `redisConnect*`, we should use
`redisConnectWrapper*` wrapper from cli_common instead of
`redisConnect` from hiredis. This is a cleanup.

### DIFF
--- a/src/cli_common.c
+++ b/src/cli_common.c
@@ -428,19 +428,33 @@ sds cliVersion(void) {
 }
 
 /* This is a wrapper to call redisConnect or redisConnectWithTimeout. */
-redisContext *redisConnectWrapper(const char *ip, int port, const struct timeval tv) {
-    if (tv.tv_sec == 0 && tv.tv_usec == 0) {
-        return redisConnect(ip, port);
-    } else {
-        return redisConnectWithTimeout(ip, port, tv);
+redisContext *redisConnectWrapper(const char *ip, int port, const struct timeval tv, int nonblock) {
+    redisOptions options = {0};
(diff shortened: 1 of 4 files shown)


hwware (approved):
LGTM.

enjoy-binbin (commented):
`There are mixed usage of ` I failed to see that, could you elaborate it a little more?

I guess you are saying we should use the wrapper instead of the  `redisConnect` from hiredis? Ok, this is a cleanup and LGTM

src/cli_common.c
```diff
@@ -428,19 +428,33 @@ sds cliVersion(void) {
 }
 
 /* This is a wrapper to call redisConnect or redisConnectWithTimeout. */
-redisContext *redisConnectWrapper(const char *ip, int port, const struct timeval tv) {
-    if (tv.tv_sec == 0 && tv.tv_usec == 0) {
-        return redisConnect(ip, port);
-    } else {
-        return redisConnectWithTimeout(ip, port, tv);
+redisContext *redisConnectWrapper(const char *ip, int port, const struct timeval tv, int nonblock) {
```
enjoy-binbin, line 431: Maybe we could rename it to valkeyConnectWrapper or cliConnectWrapper?

pizhenwei: I heard that valkey will use libvalkey instead of hiredis in the future, replacing symbol at that time together?

enjoy-binbin: SGTM.

enjoy-binbin (approved):

Issue #1755, opened 2025-02-20
[NEW] Support BYPOLYGON option for GEOSEARCH


KarthikSubbarao:
**The problem/use-case that the feature addresses**

Currently, Valkey users can only search through their GEO Data with `GEOSEARCH` using `BYRADIUS` (specified distance as radius) or using `BYBOX` (specified width and height) from a specific starting coordinate (`FROMMEMBER` or `FROMLONLAT`).

This can result in Valkey returning a result set of members from a much larger area (CIRCLE/BOX) beyond what some applications / use cases desire.

There are workloads which require finding locations (members) within a complex and irregularly shaped area, such as a specific neighborhood, city district, protected area, tracking movement patterns, etc.

It would be great if Valkey can support a custom and narrowed down search for GEO coordinates.

**Description of the feature**

The feature proposed here is extending the existing `GEOSEARCH` command to support a new search mechanism: `BYPOLYGON`.

`BYPOLYGON` will accept LONLAT coordinates as vertices of the polygon (should be closed) and will return all the members that exist within this area.

`GEOSEARCH` with `BYPOLYGON` does **not** need any specific starting coordinate (`FROMMEMBER` or `FROMLONLAT`). 
* Note: This is different from the current behavior today where `GEOSEARCH` requires a starting coordinate.

##### High Level Approach:
* Using the provided vertices, compute the centroid and the bounding box.
* Using the centroid and the bounding box, calculate the max radius (via haversine formula) to use for the existing and currently used `GEOHASH` based nearest neighbor search.
* Using the neighbors, filter out points which are not within the polygon using a simple ray casting algorithm ([link](https://en.wikipedia.org/wiki/Point_in_polygon)).

I have a functional implementation of this here (https://github.com/KarthikSubbarao/valkey/tree/geo), and I can work towards a PR (with cleaning up/optimizing/testing) based on feedback here if this idea sounds good.

##### Example usage:
Here, we can see it is possible to search for the specific members with a provided area using `BYPOLYGON` to avoid getting back a result set with other members: Agrigento and Catania.
```
127.0.0.1:7000> GEOADD Sicily 13.361389 38.115556 "Palermo" 14.733333 36.933333 "Ragusa" 15.087269 37.502669 "Catania" 15.287602 37.070104 "Siracusa" 13.588412 37.311348 "Agrigento"
(integer) 5
[...]

PR #1809, opened 2025-03-03
Support BYPOLYGON option for GEOSEARCH

KarthikSubbarao:
This PR implements the `BYPOLYGON` search (described here: https://github.com/valkey-io/valkey/issues/1755) for the GEOSEARCH and GEOSEARCHSTORE valkey commands

##### Example usage:
Here, we can see it is possible to search for the specific members with a provided area using `BYPOLYGON` to avoid getting back a result set with other members: Agrigento and Catania.
```
127.0.0.1:7000> GEOADD Sicily 13.361389 38.115556 "Palermo" 14.733333 36.933333 "Ragusa" 15.087269 37.502669 "Catania" 15.287602 37.070104 "Siracusa" 13.588412 37.311348 "Agrigento"
(integer) 5
127.0.0.1:7000> GEOSEARCH Sicily BYPOLYGON 6 13.268550 38.354626 15.379554 37.143270 14.7 36.6 14.193438 37.334376 12.972021 38.138380 13.3 38.2
1) "Palermo"
2) "Ragusa"
3) "Siracusa"
127.0.0.1:7000>
```

Illustrating by using a python flask server to read from a local Valkey server and plot GEO data from Valkey using folium and render this:

Image


src/geohash_helper.c
```diff
@@ -280,3 +365,21 @@ int geohashGetDistanceIfInRectangle(double width_m,
     *distance = geohashGetDistance(x1, y1, x2, y2);
     return 1;
 }
+
+/* Check if a point is in a polygon using ray casting. */
+int geohashGetDistanceIfInPolygon(double x1, double y1, double *xy, double (*vertices)[2], int num_vertices, double *distance) {
```
yangbodong22011: Do we need to distinguish whether the point is on the edge?

KarthikSubbarao: Thank you , this is a good question. I thought about this during implementation.

It is unlikely that a point will be on the an edge of a polygon to the exact precision (except for the very clear cases) - so the `BYPOLYGON` search might not return any results even if we handle the "is a point on the edge" check.  


You can let me know what you think based on your experience with GIS applications

yangbodong22011: Refer ray cast algorithm by W. Randolph Franklin https://github.com/tair-opensource/TairGis/blob/main/src/spatial/polyraycast.c#L157

KarthikSubbarao: I will look into the linked algorithm and will get back to you here in the next day or two

KarthikSubbarao: The current code in this PR is based on PNPOLY - Point Inclusion in Polygon Test by W. Randolph Franklin (WRF). See: https://wrfranklin.org/Research/Short_Notes/pnpoly.html


I read through the link you suggested (https://github.com/tair-opensource/TairGis/blob/main/src/spatial/polyraycast.c#L157).
* From reading the code, the algorithm looks exactly the same as this PR with two additional cases handled: (1) it uses cross product to check if a point is on the line & segment (2) checks if a point is on the vertex.  
* It seems more computationally expensive to handle these two edge cases which will not be caught because it compares to the exact floating point precision since it using `==` (instead of eplison)

Specifically the linked code has edge case handling using `==` which will very rarely be true unless the point is exactly on the vertex or segment/line.
This seems unlikely to occur unless we are using test data or a geographically very small polygon and set of coordinates
```
	if (a.x == b.x && a.y == b.y) {
		if (p.x == a.x && p.y == a.y) {
			return RAY_ON;
		}
		return RAY_OUT;
	}
	// try to check if p is on line ab and p is on segment ab.
	if (((a.x <= p.x && p.x <= b.x) || (b.x <= p.x && p.x <= a.x)) && ((a.y <= p.y && p.y <= b.y) || (b.y <= p.y && p.y <= a.y)) && (p.y - a.y) * (b.x - a.x) == (p.x - a.x) * (b.y - a.y)) {
		return RAY_ON;
	} 
```

src/geohash_helper.c
```diff
@@ -101,11 +101,72 @@ uint8_t geohashEstimateStepsByRadius(double range_meters, double lat) {
...
+    double height = 0.0, width = 0.0;
+    if (shape->type == CIRCULAR_TYPE) {
+        // For circular shapes, use the given radius directly.
+        height = shape->conversion * shape->t.radius;
+        width = shape->conversion * shape->t.radius;
+    } else if (shape->type == RECTANGLE_TYPE) {
+        // For rectangles, calculate the diagonal as the radius.
+        height = shape->conversion * shape->t.r.height / 2;
+        width = shape->conversion * shape->t.r.width / 2;
+    } else if (shape->type == POLYGON_TYPE) {
```
KarthikSubbarao, line 111: PR changes at time of comment:
This block of code converts Lon & Lat to Cartesian Coordinates to find the Bounding Box **and** Centroid of the polygon.

Proposed PR changes:
* Use Lon and Lat directly to find the Bounding Box
* Convert to Cartesian Coordinates to compute the Centroid

PR changes at time of comment:
```
        double x = 0.0, y = 0.0, z = 0.0;
        double min_x = EARTH_RADIUS_IN_METERS, max_x = -EARTH_RADIUS_IN_METERS;
        double min_y = EARTH_RADIUS_IN_METERS, max_y = -EARTH_RADIUS_IN_METERS;
        double min_z = EARTH_RADIUS_IN_METERS, max_z = -EARTH_RADIUS_IN_METERS;
        // For polygons, calculate the bounding box & centroid based on the vertices.
        for (int i = 0; i < shape->t.polygon.num_vertices; i++) {
            double cur_x = 0, cur_y = 0, cur_z = 0;
            double longitude = shape->t.polygon.points[i][0];
            double latitude = shape->t.polygon.points[i][1];
            // Code below in this loop is for calculating centroid.
            // Convert degree to radians.
            double lon_rad = deg_rad(longitude);
            double lat_rad = deg_rad(latitude);
            // Convert to Cartesian coordinates
            cur_x = EARTH_RADIUS_IN_METERS * cos(lat_rad) * cos(lon_rad);
            cur_y = EARTH_RADIUS_IN_METERS * cos(lat_rad) * sin(lon_rad);
            cur_z = EARTH_RADIUS_IN_METERS * sin(lat_rad);
            x += cur_x;
            y += cur_y;
            z += cur_z;
            if (cur_x < min_x) min_x = cur_x;
            if (cur_x > max_x) max_x = cur_x;
            if (cur_y < min_y) min_y = cur_y;
            if (cur_y > max_y) max_y = cur_y;
            if (cur_z < min_z) min_z = cur_z;
            if (cur_z > max_z) max_z = cur_z;
        }
        // Code block below is for setting the final bounding box values.
        double min_lon = atan2(min_y, min_x);
        double min_hyp = sqrt(min_x * min_x + min_y * min_y);
        double min_lat = atan2(min_z, min_hyp);
        double max_lon = atan2(max_y, max_x);
        double max_hyp = sqrt(max_x * max_x + max_y * max_y);
        double max_lat = atan2(max_z, max_hyp);
        bounds[0] = rad_deg(min_lon);
        bounds[1] = rad_deg(min_lat);
        bounds[2] = rad_deg(max_lon);
        bounds[3] = rad_deg(max_lat);
        // Code block below is for calculating centroid.
        // Average the Cartesian coordinates
        x /= num_vertices;
        y /= num_vertices;
        z /= num_vertices;
        // Convert back to latitude and longitude
        double central_lon = atan2(y, x);           // Longitude in radians
        double central_hyp = sqrt(x * x + y * y);   // Hypotenuse
        double central_lat = atan2(z, central_hyp); // Latitude in radians
        // Convert back to degrees
        shape->xy[0] = rad_deg(central_lon);
        shape->xy[1] = rad_deg(central_lat);
        printf("Geo centroid coordinates lon lat: %f, %f\r\n", shape->xy[0], shape->xy[1]);
        return 1;
    }
```

Proposed: See the PR now

madolson:
Major decision approved as long as there are no additional API, config, or info changes. Once the PR is closed we can merge it.

KarthikSubbarao:
@yangbodong22011 / @valkey-io/core-team  - You can take a look when you have time

src/geohash_helper.c
zuiderkwast: Regarding points on the edge, doubles use different precision on ARM vs x86 so we get different rounding errors on different platforms (normal for floating point). We recently introduced an epsilon in one case here:

* #2063

src/commands/geosearch.json
```diff
@@ -45,108 +45,147 @@
...
-                    }
-                ]
-            },
-            {
-                "name": "by",
-                "type": "oneof",
-                "arguments": [
-                    {
-                        "name": "circle",
+                        "name": "radius-or-box",
```
zuiderkwast: This complexity is to exactly define in the syntax that FROM is only for radius and box and not for polygon, right?

The main use case for the syntax definition is the documentation. If this gets very complex, it will be hard to read for users.

IIUC, this will render as a nested thing like

`GEOSEARCH  key  <<FROMMEMBER  member  |  FROMLONLAT longitude latitude> <BYRADIUS radius <M | KM | FT | MI> | BYBOX width height <M | KM | FT | MI>> | BYPOLYGON ...> [ASC | DESC] [COUNT count [ANY]] [WITHCOORD] [WITHDIST] [WITHHASH]`

If we make it less exact and just change FROMMEMBER | FROMLONLAT to be optional and mention in the doc text that it's mandatory for circle and box, but not used for polygons, the syntax will not be nested and it will be easier to read, I guess...

`GEOSEARCH  key  [FROMMEMBER  member  |  FROMLONLAT longitude latitude] <BYRADIUS radius <M | KM | FT | MI> | BYBOX width height <M | KM | FT | MI> | BYPOLYGON ...> [ASC | DESC] [COUNT count [ANY]] [WITHCOORD] [WITHDIST] [WITHHASH]`

What do you think?

KarthikSubbarao: Yes, the complexity was just for correctness in commands docs (which is used as prompts in the valkey-cli).

But as you said, we can make FROMMEMBER and FROMLONLAT optional and clarify in the documentation that either one is mandatory for byradius and bybox

KarthikSubbarao: I will make the changes to keep it simple

src/geohash_helper.c
KarthikSubbarao: Thanks for sharing, @zuiderkwast .

If we really wanted to support including "points on the edge", we would have to use fabs / epsilon here as well.

I had posted [here](https://github.com/valkey-io/valkey/issues/1755#issuecomment-2819852982) that the BYRADIUS search does not do this today when it checks if `distance > radius`:
```
int geohashGetDistanceIfInRadius(double x1, double y1, double x2, double y2, double radius, double *distance) {
    *distance = geohashGetDistance(x1, y1, x2, y2);
    if (*distance > radius) return 0;
    return 1;
}
```

So, I don't think we need to handle this for the BYPOLYGON search

src/commands/geosearch.json
KarthikSubbarao: I can make the changes to keep it simple.


Just wanted to note on aspect I observe when using `valkey-cli`. 

Even with making `FROMMEMBER | FROMLONLAT` optional, I do not see complete prompting unless I provide `FROMMEMBER | FROMLONLAT` in the command.

Prompt can be seen when I use `FROMMEMBER | FROMLONLAT`:
Screenshot 2025-05-19 at 12 57 29 PM


Prompt cannot be seen when I **do not** use `FROMMEMBER | FROMLONLAT`: 
Screenshot 2025-05-19 at 12 57 47 PM

However, this is a minor issue

zuiderkwast: Looks like a bug in the cli hints?...

KarthikSubbarao: Yes, I will look into it and update back here

src/commands/geosearchstore.json
```diff
@@ -75,12 +75,14 @@
                     {
                         "token": "FROMMEMBER",
                         "name": "member",
-                        "type": "string"
+                        "type": "string",
+                        "optional": true
```
zuiderkwast: Try moving "optional" to the "oneof" container wrapping this one.

KarthikSubbarao: Yes, that fixes this issue. Thank you: https://github.com/valkey-io/valkey/pull/1809#discussion_r2096401973

src/commands/geosearch.json
KarthikSubbarao: Moving the optional flag up one level fixed the issue.

The prompt is working fine, and we can update the documentation that FROMMEMBER and FROMLONLAT are optional for BYPOLYGON, but either one is mandatory for BYRADIUS and BYBOX

src/geohash_helper.c
zuiderkwast: Agree, let's not do it now.

In #2063 we did it only for the special case of finding a point with zero radius from itself.

src/commands/geosearch.json
zuiderkwast: > Moving the optional flag up one level fixed the issue

Great!

> FROMMEMBER and FROMLONLAT are optional for BYPOLYGON

Optional but ignored? We return an error if it's provided?

src/geo.c
```diff
@@ -520,6 +531,7 @@ void georadiusGeneric(client *c, int srcKeyIndex, int flags) {
     /* Find long/lat to use for radius or box search based on inquiry type */
     int base_args;
     GeoShape shape = {0};
+    shape.t.polygon.points = NULL;
```
zuiderkwast: `shape = {0}` zeroes out the whole struct. No need to initialize `shape.t.polygon.points`.

src/geo.c
```diff
@@ -635,22 +672,26 @@ void georadiusGeneric(client *c, int srcKeyIndex, int flags) {
     if (storekey && (withdist || withhash || withcoords)) {
         addReplyErrorFormat(c, "%s is not compatible with WITHDIST, WITHHASH and WITHCOORD options",
                             flags & GEOSEARCHSTORE ? "GEOSEARCHSTORE" : "STORE option in GEORADIUS");
+        geoPolygonPointsFree(&shape);
         return;
     }
 
-    if ((flags & GEOSEARCH) && !(frommember || fromloc)) {
+    if ((flags & GEOSEARCH) && !(frommember || fromloc) && !bypolygon) {
         addReplyErrorFormat(c, "exactly one of FROMMEMBER or FROMLONLAT can be specified for %s",
```
zuiderkwast, line 682: What error message do we get for `(frommember || fromloc) && bypolygon`? I'd expect something like "FROMMEMBER and FROMLONLAT can't be used with BYPOLYGON".

Please add a test case to cover this too.

src/commands/geosearch.json
KarthikSubbarao: Sorry, I meant to say FROMMEMBER and FROMLONLAT are not supported for the  BYPOLYGON search in the current state of PR. However, in the documentation (GEOSEARCH.json), we have made FROMMEMBER and FROMLONLAT optional for keeping the json files readable and simple.


In the BYPOLYGON search, we use the centroid of the polygon as the point to search from (As opposed to RADIUS and BOX search which uses FROMMEMBER or FROMLONLAT).


Let me know your thoughts on actual command validation logic: Currently, if FROMMEMBER or FROMLONLAT are provided along with the BYPOLYGON search, the command is rejected.

src/geohash_helper.c
```diff
@@ -280,3 +341,25 @@ int geohashGetDistanceIfInRectangle(double width_m,
...
+ * The algorithm is based on PNPOLY - Point Inclusion in Polygon Test by W. Randolph Franklin (WRF).
+ * See: https://wrfranklin.org/Research/Short_Notes/pnpoly.html
+ * Returns 1 if inside the polyon and returns 0 otherwise. */
+int geohashGetDistanceIfInPolygon(double centroidLon, double centroidLat, double *point, double (*vertices)[2], int num_vertices, double *distance) {
+    int i, j;
+    int inside = 0;
+    for (i = 0, j = num_vertices - 1; i < num_vertices; j = i++) {
+        double *vertexA = vertices[i];
+        double *vertexB = vertices[j];
+        if (((vertexA[1] > point[1]) != (vertexB[1] > point[1])) && (point[0] < (vertexB[0] - vertexA[0]) * (point[1] - vertexA[1]) / (vertexB[1] - vertexA[1]) + vertexA[0])) {
```
zuiderkwast: This is the core of this algorithm. It's a very long line that's difficult to read. Let's line break it to make it easier to read.

In the original algorithm on the web page, it's split into two lines like below. Let's do at least this.

```suggestion
        if (((vertexA[1] > point[1]) != (vertexB[1] > point[1])) &&
            (point[0] < (vertexB[0] - vertexA[0]) * (point[1] - vertexA[1]) / (vertexB[1] - vertexA[1]) + vertexA[0])) {
```

src/geohash_helper.c
```diff
@@ -280,3 +341,25 @@ int geohashGetDistanceIfInRectangle(double width_m,
...
+            inside = !inside;
+        }
+    }
+    if (inside) {
+        *distance = geohashGetDistance(centroidLon, centroidLat, point[0], point[1]);
```
zuiderkwast, line 365: What's the distance used for? GEOSEARCH WITHDIST?

src/geohash_helper.c
```diff
@@ -99,11 +99,55 @@ uint8_t geohashEstimateStepsByRadius(double range_meters, double lat) {
  */
 int geohashBoundingBox(GeoShape *shape, double *bounds) {
```
zuiderkwast, line 102: Mention in this comment that it sets the centroid coordinates in the shape if it's a polygon.

tests/assets/test_cli_hint_suite.txt
```diff
@@ -99,7 +99,7 @@
 "GEORADIUS key 1 2 3 M COUNT 12 ASC " "[WITHCOORD] [WITHDIST] [WITHHASH] [STORE key|STOREDIST key]"
 
 # Incorrect argument type for double-valued token parameter.
-"GEOSEARCH k FROMLONLAT " "longitude latitude BYRADIUS radius M|KM|FT|MI|BYBOX width height M|KM|FT|MI [ASC|DESC] [COUNT count [ANY]] [WITHCOORD] [WITHDIST] [WITHHASH]"
+"GEOSEARCH k FROMLONLAT " "longitude latitude BYRADIUS radius M|KM|FT|MI|BYBOX width height M|KM|FT|MI|BYPOLYGON num-vertices longitude latitude [longitude latitude ...] [ASC|DESC] [COUNT count [ANY]] [WITHCOORD] [WITHDIST] [WITHHASH]"
```
zuiderkwast, line 102: Not about this PR specifically. I just noticed the hints don't show the groupings. It should group them with <> or () or something, e.g.

`<<BYRADIUS radius M|KM|FT|MI> | <BYBOX width height M|KM|FT|MI> | <BYPOLYGON num-vertices longitude latitude [longitude latitude ...]>>`'

We could create an issue for that.

src/geohash_helper.c
```diff
@@ -99,11 +99,55 @@ uint8_t geohashEstimateStepsByRadius(double range_meters, double lat) {
...
+            if (latitude > max_lat) max_lat = latitude;
+            /* Convert to Cartesian coordinates and accumulate for centroid. */
+            double lon_rad = deg_rad(longitude);
+            double lat_rad = deg_rad(latitude);
+            double cur_x = EARTH_RADIUS_IN_METERS * cos(lat_rad) * cos(lon_rad);
+            double cur_y = EARTH_RADIUS_IN_METERS * cos(lat_rad) * sin(lon_rad);
+            double cur_z = EARTH_RADIUS_IN_METERS * sin(lat_rad);
+            x += cur_x;
+            y += cur_y;
+            z += cur_z;
```
zuiderkwast: I don't really understand this. If I don't then I guess code readers also don't so some code comment to explain this to the readers would be good.

Do I understand correctly:

1. Long and lat are in degrees.
2. Convert to radians so we can use cos and sin on them.
3. Multiply with the radius of the earth...

I don't fully understand this Cartesian coordinate system. It starts from the center of the earth. The Z axis is north-south, the X axis is towards somewhere south of Greenwich (London) and the Y axis is the Equador-Bangladesh axis, 90 degrees from the Greenwich?

The point (x,y,z) is on the surface, right? Is the average of all such points is still on the surface? Or that doesn't matter?

zuiderkwast (commented):
I got through this now. It seems to make sense. I'm not very familiar with the Geo code so far.

src/geo.c
zuiderkwast: Or should we accept an optional FROM for polygons? We could use it for WITHDIST instead of the centroid if it's provided... WDYT?

src/geohash_helper.c
KarthikSubbarao: Yes, distance is used for the `WITHDIST` argument. Currently, the distance is calculated from the centroid to the geo point / member

zuiderkwast: Got it.

src/geo.c
KarthikSubbarao: > What error message do we get for (frommember || fromloc) && bypolygon? I'd expect something like "FROMMEMBER and FROMLONLAT can't be used with BYPOLYGON".
Please add a test case to cover this too.

Currently, `-ERR syntax error` gets returned to the client because of this catch all block here in `void georadiusGeneric(`:
```
            } else {
                geoPolygonPointsFree(&shape);
                return;
            }
```

I have a test case for this already in geo.tcl over here:
```
        # Error Case - Users cannot use BYPOLYGON along with FROMLONLAT or FROMMEMBER.
        catch {r GEOSEARCH points BYPOLYGON 3 151.2039101016274 -33.87446472104431 151.2132606898065 -33.882942295615855 151.2132606898065 -33.882942295615855 FROMLONLAT 151.2229523062706 -33.883973760201364} e
        assert_match {*ERR syntax error*} $e
        catch {r GEOSEARCH points BYPOLYGON 3 151.2039101016274 -33.87446472104431 151.2132606898065 -33.882942295615855 151.2132606898065 -33.882942295615855 FROMMEMBER memberA} e
        assert_match {*ERR syntax error*} $e
```

KarthikSubbarao: > Or should we accept an optional FROM for polygons? We could use it for WITHDIST instead of the centroid if it's provided... WDYT?

IMO, using `FROMLONLAT` or `FROMMEMBER` to calculate the distance might be slightly confusing since they are used as the starting point for BYRADIUS and BYBOX.

In the case of BYPOLYGON, the withdist currently uses the centroid which was not directly provided by the user, but is calculated as a result of polygon vertices. 

Since the search results are all based on the centroid, using different distance calculation (starting from a different member) might be confusing I feel.


If we want to use an external member for the distance calculation, we can probably consider adding a new `DISTFROM` parameter. The benefit of this new param is that it will have a consistent behavior for all 3 searches - BOX, RADIUS, BYPOLYGON.
* However, if this is not something users are asking for, we can keep this `DISTFROM` parameter as a separate geo improvement for later

src/geohash_helper.c
KarthikSubbarao: Sure, I updated it in the latest commit

src/geohash_helper.c
KarthikSubbarao: Thanks, I updated it in the latest commit

src/geo.c
KarthikSubbarao: Sure, I updated it in the latest commit

tests/assets/test_cli_hint_suite.txt
KarthikSubbarao: Yes, that would be nice. Encapsulating it makes it visually clearer

src/geohash_helper.c
KarthikSubbarao: > Do I understand correctly:
Long and lat are in degrees.
Convert to radians so we can use cos and sin on them.

This is correct

> Multiply with the radius of the earth...

If I think about this more, we do not need to multiply with the radius here.
The centroid calculation can be done just by summing the Cartesian coordinates. The radius is needed only when calculating the distance between points on earth (this is why we use it in `geohashGetDistance`).

src/geo.c
zuiderkwast: Use the centroid. 👍 

Regarding the syntax error, i guess a specific error can be nicer, since the syntax in the json/doc/hints now appears to allow FROM with BYPOLYGON. I don't feel strongly though.

tests/assets/test_cli_hint_suite.txt
zuiderkwast: Yeah, without any marker it's ambiguous.

src/geohash_helper.c
zuiderkwast: So no need to divide the sum by the number of points either then?

KarthikSubbarao: We will have to sum up the Cartesian coordinates, divide by num vertices to get the average and then convert it back to LON/LAT since other operations (e.g. Geo Hash based KNN search) that are performed after this function depend on the `shape->xy` being the LON / LAT. 

> divide by num vertices to get the average and then convert back to LON/LAT

Without dividing & converting back to LON/LAT, the result will not represent the actual centroid's position on the sphere, but instead it will be a vector sum

zuiderkwast: If we don't divide by the number of verices, the sum will be a point high up in space, but the same angle (long, lat) as if we divide, isn't it? Muliplying and dividing just moves the point higher or lower (towards or away from earth center).

So we can skip multiply by earth radius and divide by num vertices and still have the same angle. We just need the trigonometry to go back to long-lat. No?

KarthikSubbarao: > If we don't divide by the number of verices, the sum will be a point high up in space, but the same angle (long, lat) as if we divide, isn't it? 

Yes, I see what you mean. You are definitely right. We don't need to worry about the magnitude here. The division is just for more correctness in the "average"

But the angle is all that matters, so we can skip the division to save on some computations

zuiderkwast (approved):
LGTM now.

There's a merge conflict caused by #2116. It's easy to solve. (Edit) I solved it.

zuiderkwast:
Will you make a doc PR?

KarthikSubbarao:
> Will you make a doc PR?

Yes, I will work on updating the GEOSEARCH AND GEOSEARCHSTORE command documentation pages

PR #1767, opened 2025-02-24
Add large memory flag for asan tests

ranshid:
fixes: https://github.com/valkey-io/valkey/issues/1766

PR #1810, opened 2025-03-03
Stop running large memory test for address santizer

madolson:
After this change, https://github.com/valkey-io/valkey/pull/1767/files, the address sanitizer test seems to be causing the test process to just die, possibly from OOM: https://github.com/valkey-io/valkey/actions/runs/13620692597/job/38069694391#step:6:4755. 

If we switch the large memory tests to be sequential on ASAN, it might resolve it, but I wanted to open this to propose reverting it to get the tests passing again.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -705,10 +705,10 @@ jobs:
           sudo apt-get install tcl8.6 tclx -y
       - name: test
         if: true && !contains(github.event.inputs.skiptests, 'valkey')
-        run: ./runtest --accurate --verbose --dump-logs --large-memory ${{github.event.inputs.test_args}}
+        run: ./runtest --accurate --verbose --dump-logs ${{github.event.inputs.test_args}}
       - name: module api test
         if: true && !contains(github.event.inputs.skiptests, 'modules')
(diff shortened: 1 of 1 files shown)


enjoy-binbin (approved):

ranshid:
@madolson should we revert test-sanitizer-undefined as well? I do not recall I see these failed, but maybe we should completely revert so that it would be cleaner and comeup with an alternative way to run large-memory tests in dailys?

zuiderkwast (approved):

zuiderkwast:
Yes, let's merge this as a temporary solution.

More people report similar issues with exit code 143 for many different projects: https://github.com/actions/runner-images/issues/6680.

A suggestion mentioned is to use large runners. Do we have large runners already?

* [Standard runner specs](https://docs.github.com/en/actions/using-github-hosted-runners/using-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)
* [Large runner specs](https://docs.github.com/en/actions/using-github-hosted-runners/using-larger-runners/about-larger-runners#specifications-for-general-larger-runners)

madolson:
> A suggestion mentioned is to use large runners. Do we have large runners already?

Unfortunately large runners aren't free, so we would have to pay for them. We get small runners for free though.

PR #1811, opened 2025-03-03
Introduce MPTCP

pizhenwei:
Multipath TCP (MPTCP) is an extension of the standard TCP protocol that allows a single transport connection to use multiple network interfaces or paths. MPTCP is useful for applications like bandwidth aggregation, failover, and more resilient connections.

Linux kernel starts to support MPTCP since v5.6, it's time to support it.

The test report shows that MPTCP reduces latency by ~25% in a 1% networking packet drop environment.

Thanks to Matthieu Baerts <matttbe@kernel.org> for lots of review suggestions.

Proposed-by: Geliang Tang <geliang@kernel.org>
Tested-by: Gang Yan <yangang@kylinos.cn>
Signed-off-by: zhenwei pi <zhenwei.pi@linux.dev>
Signed-off-by: zhenwei pi <pizhenwei@bytedance.com>

Cc Linux kernel MPTCP maintainer @matttbe

### DIFF
--- a/src/anet.c
+++ b/src/anet.c
@@ -573,7 +573,26 @@ static int anetV6Only(char *err, int s) {
     return ANET_OK;
 }
 
-static int _anetTcpServer(char *err, int port, char *bindaddr, int af, int backlog) {
+/* Until glibc 2.39, getaddrinfo with hints.ai_protocol of IPPROTO_MPTCP leads error.
+ * Use hints.ai_protocol IPPROTO_IP (0) or IPPROTO_TCP (6) to resolve address and overwrite
+ * it when MPTCP is enabled.
+ * Ref: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/net/mptcp/mptcp_connect.c
+ */
+static int anetTcpSetMptcp(char *err, int ai_protocol, int mptcp) {
+    if (mptcp) {
+#ifdef IPPROTO_MPTCP
+        UNUSED(err);
+        return IPPROTO_MPTCP;
+    }
+#else
+        anetSetError(err, "MPTCP is not supported on this platform");
+        return ANET_ERR;
+    }
+#endif
+    return ai_protocol;
+}
+
+static int _anetTcpServer(char *err, int port, char *bindaddr, int af, int backlog, int mptcp) {
     int s = -1, rv;
     char _port[6]; /* strlen("65535") */
     struct addrinfo hints, *servinfo, *p;
@@ -591,7 +610,10 @@ static int _anetTcpServer(char *err, int port, char *bindaddr, int af, int backl
         return ANET_ERR;
     }
     for (p = servinfo; p != NULL; p = p->ai_next) {
-        if ((s = socket(p->ai_family, p->ai_socktype, p->ai_protocol)) == -1) continue;
+        rv = anetTcpSetMptcp(err, p->ai_protocol, mptcp);
+        if (rv == ANET_ERR) goto error;
+
+        if ((s = socket(p->ai_family, p->ai_socktype, rv)) == -1) continue;
 
         if (af == AF_INET6 && anetV6Only(err, s) == ANET_ERR) goto error;
         if (anetSetReuseAddr(err, s) == ANET_ERR) goto error;
@@ -611,12 +633,12 @@ end:
     return s;
 }
 
-int anetTcpServer(char *err, int port, char *bindaddr, int backlog) {
-    return _anetTcpServer(err, port, bindaddr, AF_INET, backlog);
+int anetTcpServer(char *err, int port, char *bindaddr, int backlog, int mptcp) {
+    return _anetTcpServer(err, port, bindaddr, AF_INET, backlog, mptcp);
 }
 
-int anetTcp6Server(char *err, int port, char *bindaddr, int backlog) {
-    return _anetTcpServer(err, port, bindaddr, AF_INET6, backlog);
+int anetTcp6Server(char *err, int port, char *bindaddr, int backlog, int mptcp) {
+    return _anetTcpServer(err, port, bindaddr, AF_INET6, backlog, mptcp);
 }
 
 int anetUnixServer(char *err, char *path, mode_t perm, int backlog, char *group) {
--- a/src/anet.c
+++ b/src/anet.c
@@ -578,7 +578,7 @@ static int anetV6Only(char *err, int s) {
  * it when MPTCP is enabled.
  * Ref: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/net/mptcp/mptcp_connect.c
  */
-static int anetTcpSetMptcp(char *err, int ai_protocol, int mptcp) {
+static int anetTcpGetProtocol(char *err, int ai_protocol, int mptcp) {
     if (mptcp) {
 #ifdef IPPROTO_MPTCP
         UNUSED(err);
@@ -610,7 +610,7 @@ static int _anetTcpServer(char *err, int port, char *bindaddr, int af, int backl
         return ANET_ERR;
     }
     for (p = servinfo; p != NULL; p = p->ai_next) {
-        rv = anetTcpSetMptcp(err, p->ai_protocol, mptcp);
+        rv = anetTcpGetProtocol(err, p->ai_protocol, mptcp);
         if (rv == ANET_ERR) goto error;
 
         if ((s = socket(p->ai_family, p->ai_socktype, rv)) == -1) continue;
--- a/src/anet.c
+++ b/src/anet.c
@@ -607,10 +607,7 @@ static int _anetTcpServer(char *err, int port, char *bindaddr, int af, int backl
         return ANET_ERR;
     }
     for (p = servinfo; p != NULL; p = p->ai_next) {
-        rv = anetTcpGetProtocol(mptcp);
-        if (rv == ANET_ERR) goto error;
-
-        if ((s = socket(p->ai_family, p->ai_socktype, rv)) == -1) continue;
+        if ((s = socket(p->ai_family, p->ai_socktype, anetTcpGetProtocol(mptcp))) == -1) continue;
 
         if (af == AF_INET6 && anetV6Only(err, s) == ANET_ERR) goto error;
         if (anetSetReuseAddr(err, s) == ANET_ERR) goto error;
--- a/src/anet.c
+++ b/src/anet.c
@@ -50,7 +50,9 @@
 #include "anet.h"
 #include "config.h"
 #include "util.h"
-#include "server.h"
+#include "serverassert.h"
+
+#define UNUSED(x) (void)(x)
 
 static void anetSetError(char *err, const char *fmt, ...) {
     va_list ap;
@@ -582,7 +584,7 @@ static int anetTcpGetProtocol(int is_mptcp_enabled) {
 #ifdef IPPROTO_MPTCP
     return is_mptcp_enabled ? IPPROTO_MPTCP : IPPROTO_TCP;
 #else
-    serverAssert(!is_mptcp_enabled);
+    assert(!is_mptcp_enabled);
     return IPPROTO_TCP;
 #endif
 }
--- a/src/anet.c
+++ b/src/anet.c
@@ -50,8 +50,7 @@
 #include "anet.h"
 #include "config.h"
 #include "util.h"
-
-#define UNUSED(x) (void)(x)
+#include "server.h"
 
 static void anetSetError(char *err, const char *fmt, ...) {
     va_list ap;
@@ -573,22 +572,19 @@ static int anetV6Only(char *err, int s) {
     return ANET_OK;
 }
 
-/* Until glibc 2.39, getaddrinfo with hints.ai_protocol of IPPROTO_MPTCP leads error.
+/* XXX: Until glibc 2.41, getaddrinfo with hints.ai_protocol of IPPROTO_MPTCP leads error.
  * Use hints.ai_protocol IPPROTO_IP (0) or IPPROTO_TCP (6) to resolve address and overwrite
  * it when MPTCP is enabled.
  * Ref: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/net/mptcp/mptcp_connect.c
+ *      https://sourceware.org/git/?p=glibc.git;a=commit;h=a8e9022e0f829d44a818c642fc85b3bfbd26a514
  */
-static int anetTcpGetProtocol(char *err, int ai_protocol, int mptcp) {
-    if (mptcp) {
+static int anetTcpGetProtocol(int is_mptcp_enabled) {
 #ifdef IPPROTO_MPTCP
-        UNUSED(err);
-        return IPPROTO_MPTCP;
+    return is_mptcp_enabled ? IPPROTO_MPTCP : IPPROTO_TCP;
 #else
-        anetSetError(err, "MPTCP is not supported on this platform");
-        return ANET_ERR;
+    serverAssert(!is_mptcp_enabled);
+    return IPPROTO_TCP;
 #endif
-    }
-    return ai_protocol;
 }
 
 static int _anetTcpServer(char *err, int port, char *bindaddr, int af, int backlog, int mptcp) {
@@ -609,7 +605,7 @@ static int _anetTcpServer(char *err, int port, char *bindaddr, int af, int backl
         return ANET_ERR;
     }
     for (p = servinfo; p != NULL; p = p->ai_next) {
-        rv = anetTcpGetProtocol(err, p->ai_protocol, mptcp);
+        rv = anetTcpGetProtocol(mptcp);
         if (rv == ANET_ERR) goto error;
 
         if ((s = socket(p->ai_family, p->ai_socktype, rv)) == -1) continue;
--- a/src/anet.c
+++ b/src/anet.c
@@ -583,12 +583,11 @@ static int anetTcpGetProtocol(char *err, int ai_protocol, int mptcp) {
 #ifdef IPPROTO_MPTCP
         UNUSED(err);
         return IPPROTO_MPTCP;
-    }
 #else
         anetSetError(err, "MPTCP is not supported on this platform");
         return ANET_ERR;
-    }
 #endif
+    }
     return ai_protocol;
 }
 
--- a/src/config.c
+++ b/src/config.c
@@ -3276,7 +3276,7 @@ standardConfig static_configs[] = {
     createIntConfig("timeout", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.maxidletime, 0, INTEGER_CONFIG, NULL, NULL), /* Default client timeout: infinite */
     createIntConfig("replica-announce-port", "slave-announce-port", MODIFIABLE_CONFIG, 0, 65535, server.replica_announce_port, 0, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("tcp-backlog", NULL, IMMUTABLE_CONFIG, 0, INT_MAX, server.tcp_backlog, 511, INTEGER_CONFIG, NULL, NULL), /* TCP listen backlog. */
-    createIntConfig("mptcp", NULL, IMMUTABLE_CONFIG, 0, INT_MAX, server.mptcp, 0, INTEGER_CONFIG, NULL, NULL),               /* Multipath TCP. */
+    createBoolConfig("mptcp", NULL, IMMUTABLE_CONFIG, server.mptcp, 0, NULL, NULL), /* Multipath TCP. */
     createIntConfig("cluster-port", NULL, IMMUTABLE_CONFIG, 0, 65535, server.cluster_port, 0, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("cluster-announce-bus-port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.cluster_announce_bus_port, 0, INTEGER_CONFIG, NULL, updateClusterAnnouncedPort), /* Default: Use +10000 offset. */
     createIntConfig("cluster-announce-port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.cluster_announce_port, 0, INTEGER_CONFIG, NULL, updateClusterAnnouncedPort),         /* Use server.port */
--- a/src/config.c
+++ b/src/config.c
@@ -3276,6 +3276,7 @@ standardConfig static_configs[] = {
     createIntConfig("timeout", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.maxidletime, 0, INTEGER_CONFIG, NULL, NULL), /* Default client timeout: infinite */
     createIntConfig("replica-announce-port", "slave-announce-port", MODIFIABLE_CONFIG, 0, 65535, server.replica_announce_port, 0, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("tcp-backlog", NULL, IMMUTABLE_CONFIG, 0, INT_MAX, server.tcp_backlog, 511, INTEGER_CONFIG, NULL, NULL), /* TCP listen backlog. */
+    createIntConfig("mptcp", NULL, IMMUTABLE_CONFIG, 0, INT_MAX, server.mptcp, 0, INTEGER_CONFIG, NULL, NULL),               /* Multipath TCP. */
     createIntConfig("cluster-port", NULL, IMMUTABLE_CONFIG, 0, 65535, server.cluster_port, 0, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("cluster-announce-bus-port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.cluster_announce_bus_port, 0, INTEGER_CONFIG, NULL, updateClusterAnnouncedPort), /* Default: Use +10000 offset. */
     createIntConfig("cluster-announce-port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.cluster_announce_port, 0, INTEGER_CONFIG, NULL, updateClusterAnnouncedPort),         /* Use server.port */
--- a/src/config.c
+++ b/src/config.c
@@ -2428,15 +2428,11 @@ static int isValidIpV6(char *val, const char **err) {
 }
 
 static int isValidMptcp(int val, const char **err) {
-#ifndef IPPROTO_MPTCP
-    if (val) {
+    if (val && !anetHasMptcp()) {
         *err = "MPTCP is not supported on this platform";
         return 0;
     }
-#else
-    UNUSED(val);
-    UNUSED(err);
-#endif
+
     return 1;
 }
 
--- a/src/config.c
+++ b/src/config.c
@@ -3276,7 +3276,7 @@ standardConfig static_configs[] = {
     createIntConfig("timeout", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.maxidletime, 0, INTEGER_CONFIG, NULL, NULL), /* Default client timeout: infinite */
     createIntConfig("replica-announce-port", "slave-announce-port", MODIFIABLE_CONFIG, 0, 65535, server.replica_announce_port, 0, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("tcp-backlog", NULL, IMMUTABLE_CONFIG, 0, INT_MAX, server.tcp_backlog, 511, INTEGER_CONFIG, NULL, NULL), /* TCP listen backlog. */
-    createBoolConfig("mptcp", NULL, IMMUTABLE_CONFIG, server.mptcp, 0, NULL, NULL), /* Multipath TCP. */
+    createBoolConfig("mptcp", NULL, IMMUTABLE_CONFIG, server.mptcp, 0, NULL, NULL),                                          /* Multipath TCP. */
     createIntConfig("cluster-port", NULL, IMMUTABLE_CONFIG, 0, 65535, server.cluster_port, 0, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("cluster-announce-bus-port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.cluster_announce_bus_port, 0, INTEGER_CONFIG, NULL, updateClusterAnnouncedPort), /* Default: Use +10000 offset. */
     createIntConfig("cluster-announce-port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.cluster_announce_port, 0, INTEGER_CONFIG, NULL, updateClusterAnnouncedPort),         /* Use server.port */
--- a/src/config.c
+++ b/src/config.c
@@ -2427,6 +2427,19 @@ static int isValidIpV6(char *val, const char **err) {
     return 1;
 }
 
+static int isValidMptcp(int val, const char **err) {
+#ifndef IPPROTO_MPTCP
+    if (val) {
+        *err = "MPTCP is not supported on this platform";
+        return 0;
+    }
+#else
+    UNUSED(val);
+    UNUSED(err);
+#endif
+    return 1;
+}
+
 /* Validate specified string is a valid proc-title-template */
 static int isValidProcTitleTemplate(char *val, const char **err) {
     if (!validateProcTitleTemplate(val)) {
@@ -3276,7 +3289,7 @@ standardConfig static_configs[] = {
     createIntConfig("timeout", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.maxidletime, 0, INTEGER_CONFIG, NULL, NULL), /* Default client timeout: infinite */
     createIntConfig("replica-announce-port", "slave-announce-port", MODIFIABLE_CONFIG, 0, 65535, server.replica_announce_port, 0, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("tcp-backlog", NULL, IMMUTABLE_CONFIG, 0, INT_MAX, server.tcp_backlog, 511, INTEGER_CONFIG, NULL, NULL), /* TCP listen backlog. */
-    createBoolConfig("mptcp", NULL, IMMUTABLE_CONFIG, server.mptcp, 0, NULL, NULL),                                          /* Multipath TCP. */
+    createBoolConfig("mptcp", NULL, IMMUTABLE_CONFIG, server.mptcp, 0, isValidMptcp, NULL),                                  /* Multipath TCP. */
     createIntConfig("cluster-port", NULL, IMMUTABLE_CONFIG, 0, 65535, server.cluster_port, 0, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("cluster-announce-bus-port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.cluster_announce_bus_port, 0, INTEGER_CONFIG, NULL, updateClusterAnnouncedPort), /* Default: Use +10000 offset. */
     createIntConfig("cluster-announce-port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.cluster_announce_port, 0, INTEGER_CONFIG, NULL, updateClusterAnnouncedPort),         /* Use server.port */
--- a/valkey.conf
+++ b/valkey.conf
@@ -147,6 +147,15 @@ port 6379
 # in order to get the desired effect.
 tcp-backlog 511
 
+# Multipath TCP
+#
+# Multipath TCP (MPTCP) allows a single transport connection to use multiple network
+# interfaces or paths. MPTCP is useful for applications like bandwidth aggregation,
+# failover, and more resilient connections.
+# Note that MPTCP is supported in the official Linux kernel starting with version 5.6.
+#
+# mptcp 1
+
 # Unix socket.
 #
 # Specify the path for the Unix socket that will be used to listen for
--- a/valkey.conf
+++ b/valkey.conf
@@ -147,13 +147,13 @@ port 6379
 # in order to get the desired effect.
 tcp-backlog 511
 
-# Multipath TCP
+# Multipath TCP (MPTCP)
 #
-# Multipath TCP (MPTCP) allows a single transport connection to use multiple network
-# interfaces or paths. It's useful for applications like bandwidth aggregation,
-# failover, and more resilient connections. Set this to yes to allow clients to
-# use MPTCP. Clients that don't support it will use "plain" TCP.
-# Note that MPTCP is supported in the official Linux kernel starting with version 5.6.
+# MPTCP splits a single TCP connection into subflows over multiple interfaces or paths.
+# It enables bandwidth aggregation, failover, and improved reliability.
+# When set to 'yes', clients will be able to use MPTCP if requested. When not
+# requested, regular TCP can be used like before.
+# Note: MPTCP is supported in the mainline Linux kernel starting from version 5.6.
 #
 # mptcp yes
 
--- a/valkey.conf
+++ b/valkey.conf
@@ -154,7 +154,7 @@ tcp-backlog 511
 # failover, and more resilient connections.
 # Note that MPTCP is supported in the official Linux kernel starting with version 5.6.
 #
-# mptcp 1
+# mptcp yes
 
 # Unix socket.
 #
--- a/valkey.conf
+++ b/valkey.conf
@@ -150,8 +150,9 @@ tcp-backlog 511
 # Multipath TCP
 #
 # Multipath TCP (MPTCP) allows a single transport connection to use multiple network
-# interfaces or paths. MPTCP is useful for applications like bandwidth aggregation,
-# failover, and more resilient connections.
+# interfaces or paths. It's useful for applications like bandwidth aggregation,
+# failover, and more resilient connections. Set this to yes to allow clients to
+# use MPTCP. Clients that don't support it will use "plain" TCP.
 # Note that MPTCP is supported in the official Linux kernel starting with version 5.6.
 #
 # mptcp yes
--- a/src/anet.h
+++ b/src/anet.h
@@ -54,8 +54,8 @@
 int anetTcpNonBlockConnect(char *err, const char *addr, int port);
 int anetTcpNonBlockBestEffortBindConnect(char *err, const char *addr, int port, const char *source_addr);
 int anetResolve(char *err, char *host, char *ipbuf, size_t ipbuf_len, int flags);
-int anetTcpServer(char *err, int port, char *bindaddr, int backlog);
-int anetTcp6Server(char *err, int port, char *bindaddr, int backlog);
+int anetTcpServer(char *err, int port, char *bindaddr, int backlog, int mptcp);
+int anetTcp6Server(char *err, int port, char *bindaddr, int backlog, int mptcp);
 int anetUnixServer(char *err, char *path, mode_t perm, int backlog, char *group);
 int anetTcpAccept(char *err, int serversock, char *ip, size_t ip_len, int *port);
 int anetUnixAccept(char *err, int serversock);
--- a/src/anet.h
+++ b/src/anet.h
@@ -75,4 +75,12 @@ int anetSetSockMarkId(char *err, int fd, uint32_t id);
 int anetGetError(int fd);
 int anetIsFifo(char *filepath);
 
+static inline int anetHasMptcp(void) {
+#ifdef IPPROTO_MPTCP
+    return 1;
+#else
+    return 0;
+#endif
+}
+
 #endif
--- a/src/server.c
+++ b/src/server.c
@@ -2651,10 +2651,10 @@ int listenToPort(connListener *sfd) {
         if (optional) addr++;
         if (strchr(addr, ':')) {
             /* Bind IPv6 address. */
-            sfd->fd[sfd->count] = anetTcp6Server(server.neterr, port, addr, server.tcp_backlog);
+            sfd->fd[sfd->count] = anetTcp6Server(server.neterr, port, addr, server.tcp_backlog, server.mptcp);
         } else {
             /* Bind IPv4 address. */
-            sfd->fd[sfd->count] = anetTcpServer(server.neterr, port, addr, server.tcp_backlog);
+            sfd->fd[sfd->count] = anetTcpServer(server.neterr, port, addr, server.tcp_backlog, server.mptcp);
         }
         if (sfd->fd[sfd->count] == ANET_ERR) {
             int net_errno = errno;
--- a/src/server.h
+++ b/src/server.h
@@ -1612,6 +1612,7 @@ struct valkeyServer {
     int port;                              /* TCP listening port */
     int tls_port;                          /* TLS listening port */
     int tcp_backlog;                       /* TCP listen() backlog */
+    int mptcp;                             /* Use Multipath TCP */
     char *bindaddr[CONFIG_BINDADDR_MAX];   /* Addresses we should bind to */
     int bindaddr_count;                    /* Number of addresses in server.bindaddr[] */
     char *bind_source_addr;                /* Source address to bind on for outgoing connections */
--- a/tests/unit/introspection.tcl
+++ b/tests/unit/introspection.tcl
@@ -661,6 +661,7 @@ start_server {tags {"introspection"}} {
             rdbchecksum
             daemonize
             tcp-backlog
+            mptcp
             always-show-logo
             syslog-enabled
             cluster-enabled


pizhenwei:
This PR could be tested by https://github.com/pizhenwei/valkey/tree/mptcp-with-hiredis

xbasel:
Not sure how useful MPTCP is for low-latency data center inner networks. It was built for mobile devices to switch between WiFi and cellular, handling packet loss.
How does MPTCP perform when there is no packet drop? Does it introduce any overhead or impact latency in such stable network conditions?

I do see the benefit of bandwidth aggregation though (which can be fixed externally, eg. openmptcprouter)

src/config.c
```diff
@@ -3276,6 +3276,7 @@ standardConfig static_configs[] = {
     createIntConfig("timeout", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.maxidletime, 0, INTEGER_CONFIG, NULL, NULL), /* Default client timeout: infinite */
     createIntConfig("replica-announce-port", "slave-announce-port", MODIFIABLE_CONFIG, 0, 65535, server.replica_announce_port, 0, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("tcp-backlog", NULL, IMMUTABLE_CONFIG, 0, INT_MAX, server.tcp_backlog, 511, INTEGER_CONFIG, NULL, NULL), /* TCP listen backlog. */
+    createIntConfig("mptcp", NULL, IMMUTABLE_CONFIG, 0, INT_MAX, server.mptcp, 0, INTEGER_CONFIG, NULL, NULL),               /* Multipath TCP. */
```
xbasel: Maybe it makes sense to have a dedicated config for replication.

pizhenwei: Yes, I have the same opinion in the next step.

Because server and client benchmark MPTCP support could be tested easily, we can know the test result in the commit message.

pizhenwei:
> Not sure how useful MPTCP is for low-latency data center inner networks. It was built for mobile devices to switch between WiFi and cellular, handling packet loss. How does MPTCP perform when there is no packet drop? Does it introduce any overhead or impact latency in such stable network conditions?
> 
> I do see the benefit of bandwidth aggregation though (which can be fixed externally, eg. openmptcprouter)

MPTCP leads additional ~10% latency on Linux-6.11 from test over no packet drop stable network, so it is NOT suitable for any client/replication. If server listens on MPTCP, and the client side uses TCP, the connection will be TCP only. This means the client/replication side has a chance to decide TCP/MPTCP..

Hi @matttbe, is it possible to improve the performance of MPTCP as fast as TCP?

pizhenwei:
> Not sure how useful MPTCP is for low-latency data center inner networks. It was built for mobile devices to switch between WiFi and cellular, handling packet loss. How does MPTCP perform when there is no packet drop? Does it introduce any overhead or impact latency in such stable network conditions?
> 
> I do see the benefit of bandwidth aggregation though (which can be fixed externally, eg. openmptcprouter)

MPTCP leads additional ~10% latency on Linux-6.11 from test over no packet drop stable network, so it is NOT suitable for any client/replication. If server listens on MPTCP, and the client side uses TCP, the connection will be TCP only. This means the client/replication side has a chance to decide TCP/MPTCP..

Hi @matttbe, is it possible to improve the performance of MPTCP as fast as TCP?

matttbe:
Hello,

> Not sure how useful MPTCP is for low-latency data center inner networks. It was built for mobile devices to switch between WiFi and cellular, handling packet loss. How does MPTCP perform when there is no packet drop? Does it introduce any overhead or impact latency in such stable network conditions?

@xbasel: MPTCP can also be useful in data centre for fast recoveries when some links fail, to use the path with the lowest latency, or even to combine multiple paths. There were a few scientific papers on the subject (using MPTCP in data center). I guess with Valkey, there will be a higher focused on the latency aspect.

Please note that on the server side, supporting MPTCP is "[cheap](https://www.mptcp.dev/faq.html#why--when-should-mptcp-be-enabled-by-default)": when clients don't request to use MPTCP, server applications will receive a "plain" TCP sockets from the kernel when connections are accepted, making the performance impact minimal.

> is it possible to improve the performance of MPTCP as fast as TCP?

To be able to use multiple paths for the same connection, a few bytes need to be added to the header of each TCP packet. More details [here](https://www.mptcp.dev/faq.html#are-there-any-performance-impacts-when-using-mptcp). It means that if only one path is used per connection, MPTCP cannot be as fast as TCP. But that's normal, that's the small cost to pay to leverage multiple network paths, potentially increasing throughput and reliability. Of course, if only one path is used per MPTCP connection, no need to use **Multipath** TCP ;)

Still, we are working on reducing the gap between TCP and MPTCP single path, e.g. this recent modification [here](https://lore.kernel.org/20250218-net-next-mptcp-rx-path-refactor-v1-0-4a47d90d7998@kernel.org).

xbasel:
Unrelated to this PR, any idea if its possible create subflows on the same interface ?

matttbe:
> Unrelated to this PR, any idea if its possible create subflows on the same interface ?

Yes it is. This page should explain what is possible with the default path-manager: https://www.mptcp.dev/pm.html

Some ideas:
- the server can announce multiple addresses v4/v6, even with a port. So it can announce the same IP with different ports
- the client can use multiple addresses, v4/v6. They can also be assigned to the same interface or not. The `fullmesh` mode can help to create even more paths than "supposedly" needed.
- technically, the client could establish multiples subflows from the same source address and to the same destination using different source ports, but that seems very particular, and only the userspace PM will let you do that.

Don't hesitate to share the use-case if something is not supported ;-)

xbasel:
Some cloud providers and some  internet routers enforce per-flow bandwidth limits. 
MPTCP can "fix" this issue... Basically to use MPTCP as "download accelerator".

src/config.c
pizhenwei: @xbasel 

This is the first step - server side supports MPTCP. If this is merged into valkey, then I'll support:
- explicit MPTCP config for replication. This will be helpful for replication data synchronization over a long distance.
- support libvalkey. After valkey uses libvalkey instead of hiredis, the client tools will support MPTCP easily. https://github.com/valkey-io/valkey/pull/1802 is ready to review, this will help MPTCP to support MPTCP with a few code changes.

src/anet.c
```diff
@@ -573,7 +573,26 @@ static int anetV6Only(char *err, int s) {
     return ANET_OK;
 }
 
-static int _anetTcpServer(char *err, int port, char *bindaddr, int af, int backlog) {
+/* Until glibc 2.39, getaddrinfo with hints.ai_protocol of IPPROTO_MPTCP leads error.
+ * Use hints.ai_protocol IPPROTO_IP (0) or IPPROTO_TCP (6) to resolve address and overwrite
+ * it when MPTCP is enabled.
+ * Ref: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/net/mptcp/mptcp_connect.c
+ */
+static int anetTcpSetMptcp(char *err, int ai_protocol, int mptcp) {
```
xbasel: The 'mptcp' parameter is redundant; the function should assume  MPTCP was requested (the function name suggests this).

or alternatively change the function name (anetTcpSet...) and keep the param

pizhenwei: This would be better:
```
static int _anetTcpServer(char *err, int port, char *bindaddr, int af, int backlog, int mptcp) {
    ...
    memset(&hints, 0, sizeof(hints));
    hints.ai_family = af; 
    hints.ai_socktype = SOCK_STREAM;
    hints.ai_flags = AI_PASSIVE; /* No effect if bindaddr != NULL */
    hints.ai_protocol = mptcp ? IPPROTO_MPTCP : IPPROTO_IP;
    ...
    getaddrinfo(bindaddr, _port, &hints, &servinfo);
```

However, there is a lack of getaddrinfo MPTCP support, `getaddrinfo` will fail.

As the comment, currently we have to adapt this scenario, then we can use a single function `static int anetTcpSetMptcp(char *err, int ai_protocol, int mptcp)` to handle this rather than test mptcp flag in a address resolving loop:

```
    for (p = servinfo; p != NULL; p = p->ai_next) {
        if (mptcp) {
              ...
        } else {
              ...
        }

        if ((s = socket(p->ai_family, p->ai_socktype, rv)) == -1) continue;

        if (af == AF_INET6 && anetV6Only(err, s) == ANET_ERR) goto error;
        if (anetSetReuseAddr(err, s) == ANET_ERR) goto error;
        if (anetListen(err, s, p->ai_addr, p->ai_addrlen, backlog, 0, NULL) == ANET_ERR) s = ANET_ERR;
        goto end;
    }
```

matttbe: I don't know the code but for `hints.ai_protocol`, can you not always set 0 or modulo 256.

pizhenwei: I'm not sure it's the reason or not, from [glibc source](https://sourceware.org/git/?p=glibc.git;a=blob;f=nss/getaddrinfo.c;h=21e899ff1d62e2ad6738d250c32f9c7262a59a92;hb=4cf2d869367e3813c6c8f662915dedb1f3830c53#l132):

```
static const struct gaih_typeproto gaih_inet_typeproto[] =
{
  { 0, 0, 0, false, "" },
  { SOCK_STREAM, IPPROTO_TCP, 0, true, "tcp" },
  { SOCK_DGRAM, IPPROTO_UDP, 0, true, "udp" },
#if defined SOCK_DCCP && defined IPPROTO_DCCP
  { SOCK_DCCP, IPPROTO_DCCP, 0, false, "dccp" },
#endif
#ifdef IPPROTO_UDPLITE
  { SOCK_DGRAM, IPPROTO_UDPLITE, 0, false, "udplite" },
#endif
#ifdef IPPROTO_SCTP
  { SOCK_STREAM, IPPROTO_SCTP, 0, false, "sctp" },
  { SOCK_SEQPACKET, IPPROTO_SCTP, 0, false, "sctp" },
#endif
  { SOCK_RAW, 0, GAI_PROTO_PROTOANY|GAI_PROTO_NOSERVICE, true, "raw" },
  { 0, 0, 0, false, "" }
};
```

there is a lack of MPTCP support...

pizhenwei: I send a [patch](https://patchwork.sourceware.org/project/glibc/patch/20250304095415.1917979-1-pizhenwei@bytedance.com/) to glibc reviewing list.
Even this patch is right to handle getaddrinfo MPTCP issue, it will be a long time to wait glibc upgrade. So I guess current style (static int anetTcpSetMptcp(char *err, int ai_protocol, int mptcp)) is still needed.

matttbe: @pizhenwei thank you for the patch for glibc!

> I don't know the code but for `hints.ai_protocol`, can you not always set 0 or modulo 256.

I just looked at the code, I better understand what you were doing there. So maybe it is just the name function name that is misleading?

```c
protocol = anetTcpGetProtocol(err, p->ai_protocol, mptcp);
if (protocol == ANET_ERR) goto error;
if ((s = socket(p->ai_family, p->ai_socktype, protocol)) == -1) continue;
```

src/anet.c
```diff
@@ -573,7 +573,26 @@ static int anetV6Only(char *err, int s) {
...
+    if (mptcp) {
+#ifdef IPPROTO_MPTCP
+        UNUSED(err);
+        return IPPROTO_MPTCP;
+    }
+#else
+        anetSetError(err, "MPTCP is not supported on this platform");
+        return ANET_ERR;
+    }
+#endif
```
matttbe: Maybe clearer with:

```suggestion
#ifdef IPPROTO_MPTCP
        UNUSED(err);
        return IPPROTO_MPTCP;
#else
        anetSetError(err, "MPTCP is not supported on this platform");
        return ANET_ERR;
#endif
    }
```

pizhenwei:
After renaming `anetTcpSetMptcp` to `anetTcpGetProtocol`, this function gets clearer.

Thanks to @matttbe for this suggestion!

Hi @xbasel 
Do you have any suggestion about this change?

xbasel:
> Hi @xbasel Do you have any suggestion about this change?

No. LGTM now.

pizhenwei:
Hi @PingXie @zuiderkwast 
This change is reviewed by @xbasel and Linux kernel MPTCP maintainer @matttbe , also tested by another Linux kernel MPTCP maintainer @geliangtang 's team.

Would you please take a look?

zuiderkwast:
I like this because it can be done without additional libraries.

Currently we're busy with the 8.1 release and I will be away for the next week. This one will have to wait to the next release. I have some questions though:

1. Is there any harm in enabling MPTCP by default? I.e. can we skip the `mptcp` on/off config and just allow it to be used on supported platforms?

   As @matttbe mentioned, it falls back to plain TCP when the client doesn't explicitly request MPTCP.

   > Please note that on the server side, supporting MPTCP is "[cheap](https://www.mptcp.dev/faq.html#why--when-should-mptcp-be-enabled-by-default)": when clients don't request to use MPTCP, server applications will receive a "plain" TCP sockets from the kernel when connections are accepted, making the performance impact minimal.

3. I'm wondering how clients and servers can announce and find the peer's additional IPs and/or ports. Do we need to add more configs for this?

4. If we need to expose multiple IPs and ports, how can we report this in commands like CLUSTER SLOTS?

matttbe:
> 1. Is there any harm in enabling MPTCP by default? I.e. can we skip the `mptcp` on/off config and just allow it to be used on supported platforms?
> As @matttbe mentioned, it falls back to plain TCP when the client doesn't explicitly request MPTCP.

We do recommend this, that's what GoLang is doing since v1.24 for example. But then, it is required to change the way the socket creation is handled in this PR: for the moment, there is an error if MPTCP is not supported by the kernel. Instead, if it is not possible to create an MPTCP socket, I suggest to simply retry with a plain TCP socket, then continue like before, e.g.

```c
#ifdef IPPROTO_MPTCP
    if ((s = socket(p->ai_family, p->ai_socktype, IPPROTO_MPTCP)) == -1)
#endif
    {
        if ((s = socket(p->ai_family, p->ai_socktype, p->ai_protocol)) == -1) continue;
    }
```

Note that I think it would still make sense to have an option to easily disable MPTCP support, just in case. Disabling MPTCP globally is not difficult, simply with `sysctl net.mptcp.enabled=0`, but it might require different permissions.

matttbe:
@zuiderkwast : Oops, I forgot about the two other items

> 2. I'm wondering how clients and servers can announce and find the peer's additional IPs and/or ports. Do we need to add more configs for this?

No need to change anything on the application side for that, the kernel will do that automatically once it has been configured, see [here](https://www.mptcp.dev/pm.html).

There are tools that can automatically configure the MPTCP endpoints for the kernel, e.g. NetworkManager >=1.40. If these tools are not available, this can be easily done manually thanks to the `ip mptcp endpoint add <IP address> dev <interface> <type>` command.

> 3. If we need to expose multiple IPs and ports, how can we report this in commands like CLUSTER SLOTS?

This should not change anything compared to before. If the `bind()` commands restricts to a specific IP, the restriction will still be the same. If there is no such restrictions, it means the service can be reached from different IPs, like before.

If an MPTCP endpoint has been configured to accept connections to a different port, then this port will only be used for additional subflows (paths).

Of course, there are ways for the application to get info about the connection and which subflows are being used, see [here](https://www.mptcp.dev/mptcp-info.html).

pizhenwei:
> > 1. Is there any harm in enabling MPTCP by default? I.e. can we skip the `mptcp` on/off config and just allow it to be used on supported platforms?
> >    As @matttbe mentioned, it falls back to plain TCP when the client doesn't explicitly request MPTCP.
> 
> We do recommend this, that's what GoLang is doing since v1.24 for example. But then, it is required to change the way the socket creation is handled in this PR: for the moment, there is an error if MPTCP is not supported by the kernel. Instead, if it is not possible to create an MPTCP socket, I suggest to simply retry with a plain TCP socket, then continue like before, e.g.
> 
> ```c
> #ifdef IPPROTO_MPTCP
>     if ((s = socket(p->ai_family, p->ai_socktype, IPPROTO_MPTCP)) == -1)
> #endif
>     {
>         if ((s = socket(p->ai_family, p->ai_socktype, p->ai_protocol)) == -1) continue;
>     }
> ```
> 
> Note that I think it would still make sense to have an option to easily disable MPTCP support, just in case. Disabling MPTCP globally is not difficult, simply with `sysctl net.mptcp.enabled=0`, but it might require different permissions.

Hi @matttbe 

If we change valkey as this style, valkey will listen on MPTCP by default on a higher Linux kernel implicitly, and client compiled by golang  v1.24 (or higher) will get a few higher latency. Maybe lots of golang users need additional work for trouble shoot to find the root cause. So I prefer explicit `mptcp` config for the server side, the client side also uses explicit option. Valkey clients usually access server in a good networking environment, however replica may access it across data center for data backup, enabling/disabling global MPTCP kernel config is not the key idea of this PR.

```
+------+                     +-------+ 
|valkey| --MPTCP across DC-- |replica| 
+------+                     +-------+
   |
  TCP
   |
 client
```

matttbe:
> > > 1. Is there any harm in enabling MPTCP by default? I.e. can we skip the `mptcp` on/off config and just allow it to be used on supported platforms?
> > >    As @matttbe mentioned, it falls back to plain TCP when the client doesn't explicitly request MPTCP.
> > 
> > 
> > We do recommend this, that's what GoLang is doing since v1.24 for example. But then, it is required to change the way the socket creation is handled in this PR: for the moment, there is an error if MPTCP is not supported by the kernel. Instead, if it is not possible to create an MPTCP socket, I suggest to simply retry with a plain TCP socket, then continue like before, e.g.
> > ```c
> > #ifdef IPPROTO_MPTCP
> >     if ((s = socket(p->ai_family, p->ai_socktype, IPPROTO_MPTCP)) == -1)
> > #endif
> >     {
> >         if ((s = socket(p->ai_family, p->ai_socktype, p->ai_protocol)) == -1) continue;
> >     }
> > ```
> >
> > Note that I think it would still make sense to have an option to easily disable MPTCP support, just in case. Disabling MPTCP globally is not difficult, simply with `sysctl net.mptcp.enabled=0`, but it might require different permissions.
> 
> Hi @matttbe
> 
> If we change valkey as this style, valkey will listen on MPTCP by default on a higher Linux kernel implicitly, and client compiled by golang v1.24 (or higher) will get a few higher latency. Maybe lots of golang users need additional work for trouble shoot to find the root cause. So I prefer explicit `mptcp` config for the server side, the client side also uses explicit option. Valkey clients usually access server in a good networking environment, however replica may access it across data center for data backup, enabling/disabling global MPTCP kernel config is not the key idea of this PR.

With GoLang 1.24, only the server side is using MPTCP by default, not the client side. That's good to have servers supporting MPTCP by default, so the clients, the ones who usually benefits the most from MPTCP, can use it if they want to.

The latency is not supposed to increase with MPTCP, it should even decrease when it is possible to pick a path with a lower latency or less loaded. If the latency increases significantly with MPTCP single path compared to TCP, that's not normal. In this case, please report a new issue on our side.

Dwyane-Yan:
Based on the test results, if MPTCP connection is used by default, it may incur some performance loss.
Below are some simple test results using valkey-benchmark:
                Origin                  MPTCP
             rps  |  avg_msec        rps  |  avg_msec
ping      132363.6|  1.494        110492.4|  1.793
set       11013.0 |  18.152       10747.1 |  18.592
get       11371.9 |  17.568       11090.4 |  18.021

The above test results were obtained from tests conducted in a virtual environment set up using a script.
Link: https://github.com/Dwyane-Yan/mptcp_net-next/blob/export/tools/testing/selftests/net/mptcp/mptcp_valkey.sh.
The configuration of this script is a little complex:
        1、Requires the Linux tools/testing/selftests environment.
        2、cd to .tools/testing/selftests/net/mptcp
        3、make
        4、cp [valkey-server/valkey-benchmark/valkey-cli(with mptcp)] ./
        valkey-server and valkey-benchmark is compiled in https://github.com/pizhenwei/valkey/tree/mptcp-with-hiredis
        5、sudo ./mptcp_valkey.sh
This script has parameters:
        -l represents a lossy network environment, and -m represents MPTCP subflows.

Usage example:
        sudo ./mptcp-valkey.sh -m 2 # Represents an environment with 2 network interfaces.

The above test results were obtained using:
sudo ./mptcp-valkey.sh
sudo ./mptcp-valkey.sh -m 1, which simulates a scenario where only one network interface exists using MPTCP.

By the way,  there is a simpler testing method here: using 'mptcpize run' for testing.
The mptcpize is a tool that can force transfer the TCP socket into MPTCP.
This eliminates the need to set up additional environments. You can easily obtain test results by simply
running 'mptcpize run ./valkey-server' and 'mptcpize run ./valkey-benchmark -h XXX -d 1024 -c 200 -n 10000000 -t ping'.
Similarly, based on this testing in a real-world environment, the results also prove that there is a slight
performance decrease when using MPTCP with only one network interface.

Thanks the community for attention to MPTCP. Hope we can work together to make this happen.

Dwyane-Yan:
Based on the test results, if MPTCP connection is used by default, it may incur some performance loss.
Below are some simple test results using valkey-benchmark:

```
                Origin                  MPTCP
             rps  |  avg_msec        rps  |  avg_msec
ping      132363.6|  1.494        110492.4|  1.793
set       11013.0 |  18.152       10747.1 |  18.592
get       11371.9 |  17.568       11090.4 |  18.021

```

The above test results were obtained from tests conducted in a virtual environment set up using a script.
Link: https://github.com/Dwyane-Yan/mptcp_net-next/blob/export/tools/testing/selftests/net/mptcp/mptcp_valkey.sh.
The configuration of this script is a little complex:
```
        1、Requires the Linux tools/testing/selftests environment.
        2、cd to .tools/testing/selftests/net/mptcp
        3、make
        4、cp [valkey-server/valkey-benchmark/valkey-cli(with mptcp)] ./
        5、sudo ./mptcp_valkey.sh
```
valkey-server 、valkey-benchmark、valkey-cli is compiled in https://github.com/pizhenwei/valkey/tree/mptcp-with-hiredis

This script has parameters:
        -l represents a lossy network environment, and -m represents MPTCP subflows.

Usage example:
       ` sudo ./mptcp-valkey.sh -m 2 # Represents an environment with 2 network interfaces.
`
The above test results were obtained using:
```
sudo ./mptcp-valkey.sh
sudo ./mptcp-valkey.sh -m 1,
``` 
which simulates a scenario where only one network interface exists using MPTCP.

By the way,  there is a simpler testing method here: using `'mptcpize run' `for testing.
The mptcpize is a tool that can force transfer the TCP socket into MPTCP.
This eliminates the need to set up additional environments. You can easily obtain test results by simply
running` 'mptcpize run ./valkey-server'` and` 'mptcpize run ./valkey-benchmark -h XXX -d 1024 -c 200 -n 10000000 -t ping`'.
Similarly, based on this testing in a real-world environment, the results also prove that there is a slight
performance decrease when using MPTCP with only one network interface.

Thanks the community for attention to MPTCP. Hope we can work together to make this happen.

src/config.c
zuiderkwast: The config should be a bool config, createBoolConfig. Values specified as yes/no.

matttbe:
> Similarly, based on this testing in a real-world environment, the results also prove that there is a slight performance decrease when using MPTCP with only one network interface.

@Dwyane-Yan Thank you for the tests and sharing the results! Do you have more details about the benchmark tool? It looks like it will create many "small" requests (1KB), but will it create one connection per request, or one (or a few) connections doing many requests?

Because you already contributed to MPTCP in the Linux kernel, maybe would like to analyse how the kernel is behaving with TCP and MPTCP? It would be really good to analyse the performances in this use-case -- e.g. with `perf`, maybe with some flamegraphes? -- and identify where are the major differences between TCP and MPTCP single path.

If that's OK, can we move this discussion to a new issue on [MPTCP side](https://github.com/multipath-tcp/mptcp_net-next/issues/)?

src/config.c
pizhenwei: Right, fix it and append a commit. Thanks!

src/anet.c
matttbe: @pizhenwei it is not the most important, but I think you missed this comment :arrow_up:

src/anet.c
```diff
@@ -591,7 +610,10 @@ static int _anetTcpServer(char *err, int port, char *bindaddr, int af, int backl
         return ANET_ERR;
     }
     for (p = servinfo; p != NULL; p = p->ai_next) {
-        if ((s = socket(p->ai_family, p->ai_socktype, p->ai_protocol)) == -1) continue;
+        rv = anetTcpGetProtocol(err, p->ai_protocol, mptcp);
+        if (rv == ANET_ERR) goto error;
+
+        if ((s = socket(p->ai_family, p->ai_socktype, rv)) == -1) continue;
```
matttbe: @pizhenwei if Valkey maintainers are still OK with that, I think it would make sense to create an MPTCP socket by default, if available. The option can still be used to turn it off if really needed. Then, it will be up to the client to decide to use MPTCP or not: that will ease the deployments. If MPTCP is only needed for some specific clients, then they can use it, and the others can continue to use "plain" TCP like before.

So here, `anetTcpGetProtocol` can be dropped, and something like this can be done:

```suggestion
#ifdef IPPROTO_MPTCP
        if (!mptcp || (s = socket(p->ai_family, p->ai_socktype, IPPROTO_MPTCP)) == -1)
#endif
        {
                if ((s = socket(p->ai_family, p->ai_socktype, p->ai_protocol)) == -1) continue;
        }
```

WDYT?

src/anet.c
pizhenwei: @matttbe Sorry, I missed this. Thank you very much!

src/config.c
```diff
@@ -3276,6 +3276,7 @@ standardConfig static_configs[] = {
     createIntConfig("timeout", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.maxidletime, 0, INTEGER_CONFIG, NULL, NULL), /* Default client timeout: infinite */
     createIntConfig("replica-announce-port", "slave-announce-port", MODIFIABLE_CONFIG, 0, 65535, server.replica_announce_port, 0, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("tcp-backlog", NULL, IMMUTABLE_CONFIG, 0, INT_MAX, server.tcp_backlog, 511, INTEGER_CONFIG, NULL, NULL), /* TCP listen backlog. */
+    createBoolConfig("mptcp", NULL, IMMUTABLE_CONFIG, server.mptcp, 0, NULL, NULL),                                          /* Multipath TCP. */
```
matttbe: Linked to my previous comment:

```suggestion
    createBoolConfig("mptcp", NULL, IMMUTABLE_CONFIG, server.mptcp, 1, NULL, NULL),                                          /* Multipath TCP. */
```

valkey.conf
```diff
@@ -147,6 +147,15 @@ port 6379
...
 tcp-backlog 511
 
+# Multipath TCP
+#
+# Multipath TCP (MPTCP) allows a single transport connection to use multiple network
+# interfaces or paths. MPTCP is useful for applications like bandwidth aggregation,
+# failover, and more resilient connections.
+# Note that MPTCP is supported in the official Linux kernel starting with version 5.6.
+#
+# mptcp yes
```
matttbe, line 158: Also linked to my previous comment

```suggestion
# MPTCP will be used if available. If not, "plain" TCP connections will be used.
mptcp yes
```

src/anet.c
pizhenwei: Hi @matttbe 

* explicit mptcp (not strong opinion): I still prefer explicit mptcp config for the server side, because TCP and MPTCP are different transmission protocol theoretically. The maintainers (who deploy valkey-server) of valkey services should know the transmission protocol of their service and necessary operations for trouble shooting.
* default no (not strong opinion): coping the config from 8.0 for the new version, but it listens on different socket protocol, it will make user confused.
* `anetTcpGetProtocol` is needed (strong opinion): if user specifies `mptcp yes` on a lower kernel or other OS, it should be an error, because the valkey-server could NOT satisfy the user's request.

Let's leave this to the maintainer to decide...

matttbe: @pizhenwei thank you for your reply.

> explicit mptcp (not strong opinion): I still prefer explicit mptcp config for the server side, because TCP and MPTCP are different transmission protocol theoretically. The maintainers (who deploy valkey-server) of valkey services should know the transmission protocol of their service and necessary operations for trouble shooting.

MPTCP is an extension to TCP, it is not a radical change. On the wire, TCP is still used, but with extra options in the header. But indeed, sysadmins should be told about the behaviour change, but at the end, they should not worry about that. Please note that when the client doesn't request to use MPTCP, the server application will get a "plain" TCP socket after the `accept()`. So no modification on that side.

> * `anetTcpGetProtocol` is needed (strong opinion): if user specifies `mptcp yes` on a lower kernel or other OS, it should be an error, because the valkey-server could NOT satisfy the user's request.

If MPTCP is used by default, the application should not fail if it is not available. Or you need tree choices: auto (fallback to TCP if MPTCP is not available), on (error if not available), off (don't try to use MPTCP).

> Let's leave this to the maintainer to decide...

Indeed, my proposal was just a suggestion to ease MPTCP deployments without extra steps on the server side, only on the client side.

pizhenwei:
Force push the last commit to fix DCO error.

src/anet.c
pizhenwei: Hi @matttbe @Dwyane-Yan @zuiderkwast 

getaddrinfo MPTCP issue is fixed in glibc, please see [commit](https://sourceware.org/git/?p=glibc.git;a=commit;h=a8e9022e0f829d44a818c642fc85b3bfbd26a514).
But I think this work around logic is still needed, because Linux distributions may upgrade glibc in several years. maybe use the correct logic instead someday ...

src/anet.c
zuiderkwast: I understand it would be nice to enable it by default so the client can make the decisions. I also read https://www.mptcp.dev/faq.html#why--when-should-mptcp-be-enabled-by-default.

But I think we should have it disabled by default, for the following reasons:

* Uniform experience. On other platforms, like BSD and on older Linux, don't have this MPTCP. If a user tries to enable it where it's not supported, I like an error instead of a silent fallback.
* It's a new feature. We often add new features disabled by default so users can opt in to try them. In a future version, we can consider making it enabled (or "auto") by default.

pizhenwei: > I understand it would be nice to enable it by default so the client can make the decisions. I also read https://www.mptcp.dev/faq.html#why--when-should-mptcp-be-enabled-by-default.
> 
> But I think we should have it disabled by default, for the following reasons:
> 
> * Uniform experience. On other platforms, like BSD and on older Linux, don't have this MPTCP. If a user tries to enable it where it's not supported, I like an error instead of a silent fallback.

Right, if a user specifies the MPTCP config on an unsupported platform, an explicit error is better than a silent fallback.

> * It's a new feature. We often add new features disabled by default so users can opt in to try them. In a future version, we can consider making it enabled (or "auto") by default.

Yes, MPTCP is a recent feature.

Fully agree!

src/anet.c
```diff
@@ -573,7 +573,25 @@ static int anetV6Only(char *err, int s) {
...
+#ifdef IPPROTO_MPTCP
+        UNUSED(err);
+        return IPPROTO_MPTCP;
+#else
+        anetSetError(err, "MPTCP is not supported on this platform");
+        return ANET_ERR;
+#endif
+    }
+    return ai_protocol;
+}
```
zuiderkwast: We actually call this with `ai_protocol` set to `AF_INET` or `AF_INET6`.

I think the name and the behavior of this function is a little unintuitive.

I think we can check if MPTCP is supported only when the user tries to enable the config, i.e. in the config's `is_valid` callback, and refuse to enable it if it isn't supported, i.e. Valkey refuses to start if it's not supported. Then we don't need to check when we create a socket and we can just do like this:

```C
rv = mptcp ? IPPROTO_MPTCP : p->ai_protocol;
```

src/config.c
zuiderkwast: Let's use a validation function:

```suggestion
    createBoolConfig("mptcp", NULL, IMMUTABLE_CONFIG, server.mptcp, 0, isValidMptcp, NULL),                                  /* Multipath TCP. */
```

This function can be defined like this:

```C
static int isValidMptcp(int val, const char **err) {
#ifndef IPPROTO_MPTCP
    if (val) {
        *err = "MPTCP is not supported on this platform");
        return 0;
    }
#else
    UNUSED(val);
    UNUSED(err);
#endif
    return 1;
}
```

valkey.conf
```diff
@@ -147,6 +147,15 @@ port 6379
 # in order to get the desired effect.
 tcp-backlog 511
 
+# Multipath TCP
+#
+# Multipath TCP (MPTCP) allows a single transport connection to use multiple network
+# interfaces or paths. MPTCP is useful for applications like bandwidth aggregation,
+# failover, and more resilient connections.
```
zuiderkwast: Minor edit to mention how to enable it and what it means for clients.

```suggestion
# interfaces or paths. It's useful for applications like bandwidth aggregation,
# failover, and more resilient connections. Set this to yes to allow clients to
# use MPTCP. Clients that don't support it will use "plain" TCP.
```

src/anet.c
zuiderkwast: I mean we don't need this function. We can do the simple logic instead of calling a function.

pizhenwei: There is no such definition `IPPROTO_MPTCP` on a lower version Linux or BSD, it will fail to build. Rather than placing `#ifdef IPPROTO_MPTCP` in the loop, I use a small function (`anetTcpGetProtocol`) to do this job.

If we have to test `IPPROTO_MPTCP` here, should we also test it in the validation function again?

zuiderkwast: Hehe, yes, you're right. :laughing:

So I suppose we need this function. I think the name `anetTcpGetProtocol` is a strange if it doesn't always return a protocol, but if we make it always return a TCP protocol (IPROTO_TCP or IPROTO_MPTCP), then I think it's clear.

> If we have to test `IPPROTO_MPTCP` here, should we also test it in the validation function again?

It's seems to me more correct to do the validation in the config validation function, so we don't depend on how the error reporting works in the startup of the server. We can keep the ifdef check in both places, but we don't need to return `ANET_ERR` here. We can just do an assert. The error case can't happen here if we validate the config.

How about this?

```C
static int anetTcpGetProtocol(bool is_mptcp_enabled) {
#ifdef IPPROTO_MPTCP
    return is_mptcp_enabled ? IPPROTO_MPTCP : IPROTO_TCP;
#else
    serverAssert(!is_mptcp_enabled);
    return IPROTO_TCP;
#endif
```

pizhenwei: @zuiderkwast apply your suggestions on the top version, thanks a lot!

src/anet.c
```diff
@@ -50,8 +50,7 @@
 #include "anet.h"
 #include "config.h"
 #include "util.h"
-
-#define UNUSED(x) (void)(x)
+#include "server.h"
```
zuiderkwast: Oh.. I didn't realize this implication. To keep some parts less coupled with the huge server.h which essentially makes everything depend on everything, we usually try to avoid including it when possible. There is a smaller header file `serverassert.h` that provides an `assert` macro that we can use instead. It has the same effect, but it avoids the coupling.

src/anet.c
zuiderkwast: I'm glad you like it. :)

src/anet.c
zuiderkwast:
```suggestion
#include "serverassert.h"

#define UNUSED(x) (void)(x)
```

src/anet.c
```diff
@@ -573,7 +572,22 @@ static int anetV6Only(char *err, int s) {
...
+ * Use hints.ai_protocol IPPROTO_IP (0) or IPPROTO_TCP (6) to resolve address and overwrite
+ * it when MPTCP is enabled.
+ * Ref: https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/tools/testing/selftests/net/mptcp/mptcp_connect.c
+ *      https://sourceware.org/git/?p=glibc.git;a=commit;h=a8e9022e0f829d44a818c642fc85b3bfbd26a514
+ */
+static int anetTcpGetProtocol(int is_mptcp_enabled) {
+#ifdef IPPROTO_MPTCP
+    return is_mptcp_enabled ? IPPROTO_MPTCP : IPPROTO_TCP;
+#else
+    serverAssert(!is_mptcp_enabled);
```
zuiderkwast: If we include `serverassert.h` rather than `server.h`:

```suggestion
    assert(!is_mptcp_enabled);
```

zuiderkwast (approved):
Code LGTM.

@valkey-io/core-team majority, please ack the new config.

```
# Multipath TCP
#
# Multipath TCP (MPTCP) allows a single transport connection to use multiple network
# interfaces or paths. It's useful for applications like bandwidth aggregation,
# failover, and more resilient connections. Set this to yes to allow clients to
# use MPTCP. Clients that don't support it will use "plain" TCP.
# Note that MPTCP is supported in the official Linux kernel starting with version 5.6.
#
# mptcp yes
```

:+1: = ACK

src/anet.c
```diff
@@ -591,7 +607,10 @@ static int _anetTcpServer(char *err, int port, char *bindaddr, int af, int backl
         return ANET_ERR;
     }
     for (p = servinfo; p != NULL; p = p->ai_next) {
-        if ((s = socket(p->ai_family, p->ai_socktype, p->ai_protocol)) == -1) continue;
+        rv = anetTcpGetProtocol(mptcp);
+        if (rv == ANET_ERR) goto error;
+
+        if ((s = socket(p->ai_family, p->ai_socktype, rv)) == -1) continue;
```
matttbe: detail: `anetTcpGetProtocol()` no longer returns an error. You could even use `anetTcpGetProtocol()` directly if prefer:

```suggestion
        if ((s = socket(p->ai_family, p->ai_socktype, anetTcpGetProtocol(mptcp))) == -1) continue;
```

zuiderkwast: Good point!

pizhenwei: @matttbe Thanks! Apply this suggestion in the latest version.

valkey.conf
```diff
@@ -147,6 +147,16 @@ port 6379
...
 
+# Multipath TCP
+#
+# Multipath TCP (MPTCP) allows a single transport connection to use multiple network
+# interfaces or paths. It's useful for applications like bandwidth aggregation,
+# failover, and more resilient connections. Set this to yes to allow clients to
+# use MPTCP. Clients that don't support it will use "plain" TCP.
+# Note that MPTCP is supported in the official Linux kernel starting with version 5.6.
+#
+# mptcp yes
```
xbasel: I suggest this rewording for better clarity and technical accuracy:
```
# Multipath TCP (MPTCP)
#
# MPTCP splits a single TCP connection into subflows over multiple interfaces or paths.
# It enables bandwidth aggregation, failover, and improved reliability.
# When set to 'yes', MPTCP is enabled for clients.  Clients that don't support it
# will fall back to regular TCP.
# Note: MPTCP is supported in the mainline Linux kernel starting from version 5.6.
```

xbasel:
Just a late thought  - since we're not controlling subflows or using MPTCP specific features, and Valkey uses standard sockets, could `mptcpize` be enough here?

The advantage I see with explicit MPTCP support is allowing administrators to easily enable/disable MPTCP via config.

matttbe:
> Just a late thought - since we're not controlling subflows or using MPTCP specific features, and Valkey uses standard sockets, could `mptcpize` be enough here?

Yes it could be enough.

> The advantage I see with explicit MPTCP support is allowing administrators to easily enable/disable MPTCP via config.

Indeed, `mptcpize` is more a workaround, and some sysadmins don't want to use `LD_PRELOAD` to change the behaviour of some apps. Plus it will force all TCP sockets to use MPTCP, maybe not all need to use MPTCP. When there is an option, at least it says "someone else tested this app with MPTCP" :)

Also, later, it will be easier to enable MPTCP on the server side by default if needed!

valkey.conf
matttbe: Good idea!

Instead of:

```
# When set to 'yes', MPTCP is enabled for clients.  Clients that don't support it
# will fall back to regular TCP.
```

What about:

```
# When set to 'yes', clients will be able to use MPTCP if requested. When not
# requested, regular TCP can be used like before.
```

(Not sure the second part is really needed)

xbasel: Yea, I believe the second part may be unnecessary. Clients opting for MPTCP are likely familiar with the basics

zuiderkwast:
We discussed this in a meeting today.

Regarding future plans, do we want to support MPTCP for replication? If yes, we will need to add another config such as `repl-mptcp` for a replica's outgoing connections?

xbasel:
> We discussed this in a meeting today.
> 
> Regarding future plans, do we want to support MPTCP for replication? If yes, we will need to add another config such as `repl-mptcp` for a replica's outgoing connections?

I think replication should support MPTCP. Why not use server.mptcp? (Though a dedicated config could be useful to disable MPTCP for replication.)

pizhenwei:
> We discussed this in a meeting today.
> 
> Regarding future plans, do we want to support MPTCP for replication? If yes, we will need to add another config such as `repl-mptcp` for a replica's outgoing connections?

Yes, repl-mptcp is in the next step.

pizhenwei:
> > We discussed this in a meeting today.
> > Regarding future plans, do we want to support MPTCP for replication? If yes, we will need to add another config such as `repl-mptcp` for a replica's outgoing connections?
> 
> I think replication should support MPTCP. Why not use server.mptcp? (Though a dedicated config could be useful to disable MPTCP for replication.)

Because of the historic reason, `bind` and `port` are used for server TCP config, so I also use a short `mptcp` for it, also place the example config near them. I'm also fine with `server.mptcp`.

pizhenwei:
> > Just a late thought - since we're not controlling subflows or using MPTCP specific features, and Valkey uses standard sockets, could `mptcpize` be enough here?
> 
> Yes it could be enough.
> 
> > The advantage I see with explicit MPTCP support is allowing administrators to easily enable/disable MPTCP via config.
> 
> Indeed, `mptcpize` is more a workaround, and some sysadmins don't want to use `LD_PRELOAD` to change the behaviour of some apps. Plus it will force all TCP sockets to use MPTCP, maybe not all need to use MPTCP. When there is an option, at least it says "someone else tested this app with MPTCP" :)
> 
> Also, later, it will be easier to enable MPTCP on the server side by default if needed!

I have the same opinion, `mptcpize` forces all the incoming and outgoing connections to use MPTCP, it may be not expected by administrators.

zuiderkwast:
> I'm also fine with `server.mptcp`.

@xbasel When you said "Why not use server.mptcp?", you were referring to the variable in the code, not the config name? Configs with a dots in their names are for modules, i.e. `module-name.config-name`. I guess you meant to reuse the config `mptcp` for incoming client connections and for outgoing replication connections.

I believe it should be two separate configs. I expect users want to use replication without MPTCP for replicas within a data center with very reliable networking, while still accepting incoming MPTCP connections. You'd enable `repl-mptcp` on a replicas that are located in a different data center, a different continent, etc.

We had a high level consensus to add the feature. Only the details about config name and the default value were postponed. @madolson @PingXie do you think we can settle on the current state of this PR, `mptcp` with `no` by default?

Future follow-ups:

* `repl-mptcp` to enable MPTCP for outgoing replication connections
* Support MPTCP in valkey-cli and valkey-benchmark, when we have replaced hiredis with libvalkey

xbasel:
> I believe it should be two separate configs. I expect users want to use replication without MPTCP for replicas within a data center with very reliable networking, while still accepting incoming MPTCP connections. You'd enable `repl-mptcp` on a replicas that are located in a different data center, a different continent, etc.

@zuiderkwast I meant using the same config. Unless Valkey is running on hosts with multi-homing, MPTCP behaves just like regular TCP - only one subflow gets used. So in many setups, there's no downside.
I'm not entirely against adding a separate config, the only downside is just one more knob in the config file :)

zuiderkwast:
> Unless Valkey is running on hosts with multi-homing, MPTCP behaves just like regular TCP - only one subflow gets used. So in many setups, there's no downside.

Wrong. Each TCP header gets some extra bytes. Please read the comments in this issue such as this one: https://github.com/valkey-io/valkey/pull/1811#issuecomment-2703674908

pizhenwei:
> > I'm also fine with `server.mptcp`.
> 
> @xbasel When you said "Why not use server.mptcp?", you were referring to the variable in the code, not the config name? Configs with a dots in their names are for modules, i.e. `module-name.config-name`. I guess you meant to reuse the config `mptcp` for incoming client connections and for outgoing replication connections.
> 
> I believe it should be two separate configs. I expect users want to use replication without MPTCP for replicas within a data center with very reliable networking, while still accepting incoming MPTCP connections. You'd enable `repl-mptcp` on a replicas that are located in a different data center, a different continent, etc.
> 
> We had a high level consensus to add the feature. Only the details about config name and the default value were postponed. @madolson @PingXie do you think we can settle on the current state of this PR, `mptcp` with `no` by default?
> 
> Future follow-ups:
> 
> * `repl-mptcp` to enable MPTCP for outgoing replication connections
> * Support MPTCP in valkey-cli and valkey-benchmark, when we have replaced hiredis with libvalkey

I create [MPTCP support PR](https://github.com/valkey-io/libvalkey/pull/189) for libvalkey, please take a look.

xbasel:
> > Unless Valkey is running on hosts with multi-homing, MPTCP behaves just like regular TCP - only one subflow gets used. So in many setups, there's no downside.
> 
> Wrong. Each TCP header gets some extra bytes. Please read the comments in this issue such as this one: [#1811 (comment)](https://github.com/valkey-io/valkey/pull/1811#issuecomment-2703674908)

Yeah, I was wrong - I overlooked the protocol level overhead (and for some reason assumed MPTCP would magically disable itself if only one path/interface exists).
I ran tests and saw non-negligible latency and higher sys CPU even with a single flow...

I think it should be disabled by default.

matttbe:
> > > Unless Valkey is running on hosts with multi-homing, MPTCP behaves just like regular TCP - only one subflow gets used. So in many setups, there's no downside.
> > 
> > 
> > Wrong. Each TCP header gets some extra bytes. Please read the comments in this issue such as this one: [#1811 (comment)](https://github.com/valkey-io/valkey/pull/1811#issuecomment-2703674908)
> 
> Yeah, I was wrong - I overlooked the protocol level overhead (and for some reason assumed MPTCP would magically disable itself if only one path/interface exists). I ran tests and saw non-negligible latency and higher sys CPU even with a single flow...

Out of curiosity, what kind of test was it? Only small connections? With which kernel version?

madolson:
> We had a high level consensus to add the feature. Only the details about config name and the default value were postponed. @madolson @PingXie do you think we can settle on the current state of this PR, mptcp with no by default?

Yes, all my followup questions have so far been answered in this thread. The default of no does seem correct.

hwware:
The performance issue https://github.com/valkey-io/valkey/pull/1811#issuecomment-2703674908 is solved？ @Dwyane-Yan @matttbe I have a little bit worry about it or I missed some threads?

matttbe:
Just to make sure we don't mix up things here:
- It can be interesting to enable MPTCP by default on the server side, to let the clients decide to use MPTCP or not. In terms of performances, the impact should not be noticeable: if the clients don't request to use MPTCP, the application on the server side will receive a "plain" TCP socket like before. More details [here](https://www.mptcp.dev/faq.html#why--when-should-mptcp-be-enabled-by-default).
  - If MPTCP is enabled by default, it makes sense not to fail if it is not possible to create an MPTCP listening socket, and fallback to TCP automatically.
  - Of course, it might make sense to first introduce the option, then change the default later to play it safer. That's what they did in Go where MPTCP is now enabled on the server side by default.
- On the client side (or for replications in a DC), it might be better not to enable MPTCP by default, but have an option to let users enabling it when needed. It makes sense to use MPTCP when ... multiple paths are available :) (and the kernel has been informed of which IP addresses can be used with MPTCP)
- Regarding the performances, using multiple paths typically makes less sense with short connections. Otherwise, the overhead is probably not worth it. From what I understood, the benchmarks were done with small connections. Also, performances are getting better in newer kernel versions, and something that can be improved if someone can take the time to look at that (I don't know if @Dwyane-Yan will look at that). I don't think it should block this PR. In some cases, a small performance impact can be justified by an improvement of the resilience.

xbasel:
> Out of curiosity, what kind of test was it? Only small connections? With which kernel version?

@matttbe 

I used sockperf (tcp). default size 65507.

Summary:
| Metric             | TCP       | MPTCP (1 flow) | % Slower (MPTCP) |
|--------------------|-----------|----------------|------------------|
| Avg Latency (µs)   | 50.896    | 57.124         | +12.25%          |
| Max Latency (µs)   | 134.017   | 187.902        | +40.17%          |
| p50 (Median)       | 50.491    | 56.836         | +12.56%          |
| p99                | 57.549    | 64.794         | +12.59%          |
| Messages/sec       | 9819.3    | 8749.1         | -10.91%          |

Perf suggests these may contribute to higher latency:
```
__lock_sock_fast
 lock_sock_nested
mptcp_sendmsg
mptcp_recvmsg
```

Runs on kernel `6.5.0-1018`, on ARM.


Do these numbers make sense ?

Output:
```
**TCP**
sockperf: == version #3.7-no.git ==
sockperf[CLIENT] send on:sockperf: using recvfrom() to block on socket(s)

[ 0] IP = 172.31.22.130   PORT = 11111 # TCP
sockperf: Warmup stage (sending a few dummy messages)...
sockperf: Starting test...
sockperf: Test end (interrupted by timer)
sockperf: Test ended
sockperf: [Total Run] RunTime=10.000 sec; Warm up time=400 msec; SentMessages=98194; ReceivedMessages=98193
sockperf: ========= Printing statistics for Server No: 0
sockperf: [Valid Duration] RunTime=10.000 sec; SentMessages=98193; ReceivedMessages=98193
sockperf: ====> avg-latency=50.896 (std-dev=2.003)
sockperf: # dropped messages = 0; # duplicated messages = 0; # out-of-order messages = 0
sockperf: Summary: Latency is 50.896 usec
sockperf: Total 98193 observations; each percentile contains 981.93 observations
sockperf: ---> <MAX> observation =  134.017
sockperf: ---> percentile 99.999 =  128.781
sockperf: ---> percentile 99.990 =  105.104
sockperf: ---> percentile 99.900 =   71.896
sockperf: ---> percentile 99.000 =   57.549
sockperf: ---> percentile 90.000 =   52.452
sockperf: ---> percentile 75.000 =   51.316
sockperf: ---> percentile 50.000 =   50.491
sockperf: ---> percentile 25.000 =   49.925
sockperf: ---> <MIN> observation =   48.137


**MPTCP with single flow**
sockperf: == version #3.7-no.git ==
sockperf[CLIENT] send on:sockperf: using recvfrom() to block on socket(s)

[ 0] IP = 172.31.22.130   PORT = 11111 # TCP
sockperf: Warmup stage (sending a few dummy messages)...
sockperf: Starting test...
sockperf: Test end (interrupted by timer)
sockperf: Test ended
sockperf: [Total Run] RunTime=10.000 sec; Warm up time=400 msec; SentMessages=87492; ReceivedMessages=87491
sockperf: ========= Printing statistics for Server No: 0
sockperf: [Valid Duration] RunTime=10.000 sec; SentMessages=87491; ReceivedMessages=87491
sockperf: ====> avg-latency=57.124 (std-dev=2.674)
sockperf: # dropped messages = 0; # duplicated messages = 0; # out-of-order messages = 0
sockperf: Summary: Latency is 57.124 usec
sockperf: Total 87491 observations; each percentile contains 874.91 observations
sockperf: ---> <MAX> observation =  187.902
sockperf: ---> percentile 99.999 =  167.123
sockperf: ---> percentile 99.990 =  138.633
sockperf: ---> percentile 99.900 =   88.516
sockperf: ---> percentile 99.000 =   64.794
sockperf: ---> percentile 90.000 =   58.810
sockperf: ---> percentile 75.000 =   57.703
sockperf: ---> percentile 50.000 =   56.836
sockperf: ---> percentile 25.000 =   56.025
sockperf: ---> <MIN> observation =   53.679

```

matttbe:
@xbasel thank you for sharing this. I quickly tried in a VM using a kernel v6.15-rc0 with localhost and I don't see such differences when running the `ping-pong` test a few time, with and without MPTCP (single path).

According to what perf told you, the improvement might be due to [some recent improvements](https://lore.kernel.org/netdev/20250218-net-next-mptcp-rx-path-refactor-v1-0-4a47d90d7998@kernel.org/) on MPTCP side trying to improve the performances with a single path. That's getting better :)

```
root@mptcpdev# mptcpize run sockperf ping-pong -i 127.0.0.1 -p 5001 --tcp -m 65507 -t 10
sockperf: == version #3.7-no.git ==
sockperf[CLIENT] send on:sockperf: using recvfrom() to block on socket(s)

[ 0] IP = 127.0.0.1       PORT =  5001 # TCP
sockperf: Warmup stage (sending a few dummy messages)...
sockperf: Starting test...
sockperf: Test end (interrupted by timer)
sockperf: Test ended
sockperf: [Total Run] RunTime=10.000 sec; Warm up time=400 msec; SentMessages=145841; ReceivedMessages=145840
sockperf: ========= Printing statistics for Server No: 0
sockperf: [Valid Duration] RunTime=9.550 sec; SentMessages=138507; ReceivedMessages=138507
sockperf: ====> avg-latency=34.443 (std-dev=4.053)
sockperf: # dropped messages = 0; # duplicated messages = 0; # out-of-order messages = 0
sockperf: Summary: Latency is 34.443 usec
sockperf: Total 138507 observations; each percentile contains 1385.07 observations
sockperf: ---> <MAX> observation =   71.974
sockperf: ---> percentile 99.999 =   71.212
sockperf: ---> percentile 99.990 =   59.751
sockperf: ---> percentile 99.900 =   46.406
sockperf: ---> percentile 99.000 =   41.442
sockperf: ---> percentile 90.000 =   37.610
sockperf: ---> percentile 75.000 =   35.265
sockperf: ---> percentile 50.000 =   34.659
sockperf: ---> percentile 25.000 =   34.108
sockperf: ---> <MIN> observation =   14.016
root@mptcpdev# mptcpize run sockperf ping-pong -i 127.0.0.1 -p 5001 --tcp -m 65507 -t 10
sockperf: == version #3.7-no.git ==
sockperf[CLIENT] send on:sockperf: using recvfrom() to block on socket(s)

[ 0] IP = 127.0.0.1       PORT =  5001 # TCP
sockperf: Warmup stage (sending a few dummy messages)...
sockperf: Starting test...
sockperf: Test end (interrupted by timer)
sockperf: Test ended
sockperf: [Total Run] RunTime=10.000 sec; Warm up time=400 msec; SentMessages=156938; ReceivedMessages=156937
sockperf: ========= Printing statistics for Server No: 0
sockperf: [Valid Duration] RunTime=9.550 sec; SentMessages=148166; ReceivedMessages=148166
sockperf: ====> avg-latency=32.180 (std-dev=5.986)
sockperf: # dropped messages = 0; # duplicated messages = 0; # out-of-order messages = 0
sockperf: Summary: Latency is 32.180 usec
sockperf: Total 148166 observations; each percentile contains 1481.66 observations
sockperf: ---> <MAX> observation =  131.405
sockperf: ---> percentile 99.999 =   77.499
sockperf: ---> percentile 99.990 =   62.962
sockperf: ---> percentile 99.900 =   48.165
sockperf: ---> percentile 99.000 =   43.806
sockperf: ---> percentile 90.000 =   36.322
sockperf: ---> percentile 75.000 =   34.925
sockperf: ---> percentile 50.000 =   33.667
sockperf: ---> percentile 25.000 =   31.428
sockperf: ---> <MIN> observation =   14.071
```

```
root@mptcpdev# sockperf ping-pong -i 127.0.0.1 -p 5001 --tcp -m 65507 -t 10
sockperf: == version #3.7-no.git ==
sockperf[CLIENT] send on:sockperf: using recvfrom() to block on socket(s)

[ 0] IP = 127.0.0.1       PORT =  5001 # TCP
sockperf: Warmup stage (sending a few dummy messages)...
sockperf: Starting test...
sockperf: Test end (interrupted by timer)
sockperf: Test ended
sockperf: [Total Run] RunTime=10.000 sec; Warm up time=400 msec; SentMessages=143223; ReceivedMessages=143222
sockperf: ========= Printing statistics for Server No: 0
sockperf: [Valid Duration] RunTime=9.550 sec; SentMessages=133718; ReceivedMessages=133718
sockperf: ====> avg-latency=35.676 (std-dev=1.916)
sockperf: # dropped messages = 0; # duplicated messages = 0; # out-of-order messages = 0
sockperf: Summary: Latency is 35.676 usec
sockperf: Total 133718 observations; each percentile contains 1337.18 observations
sockperf: ---> <MAX> observation =   80.125
sockperf: ---> percentile 99.999 =   79.564
sockperf: ---> percentile 99.990 =   57.668
sockperf: ---> percentile 99.900 =   45.900
sockperf: ---> percentile 99.000 =   41.863
sockperf: ---> percentile 90.000 =   38.276
sockperf: ---> percentile 75.000 =   35.772
sockperf: ---> percentile 50.000 =   35.246
sockperf: ---> percentile 25.000 =   34.810
sockperf: ---> <MIN> observation =   29.435
root@mptcpdev# sockperf ping-pong -i 127.0.0.1 -p 5001 --tcp -m 65507 -t 10
sockperf: == version #3.7-no.git ==
sockperf[CLIENT] send on:sockperf: using recvfrom() to block on socket(s)

[ 0] IP = 127.0.0.1       PORT =  5001 # TCP
sockperf: Warmup stage (sending a few dummy messages)...
sockperf: Starting test...
sockperf: Test end (interrupted by timer)
sockperf: Test ended
sockperf: [Total Run] RunTime=10.000 sec; Warm up time=400 msec; SentMessages=148342; ReceivedMessages=148341
sockperf: ========= Printing statistics for Server No: 0
sockperf: [Valid Duration] RunTime=9.550 sec; SentMessages=139163; ReceivedMessages=139163
sockperf: ====> avg-latency=34.278 (std-dev=5.977)
sockperf: # dropped messages = 0; # duplicated messages = 0; # out-of-order messages = 0
sockperf: Summary: Latency is 34.278 usec
sockperf: Total 139163 observations; each percentile contains 1391.63 observations
sockperf: ---> <MAX> observation =  227.885
sockperf: ---> percentile 99.999 =  103.828
sockperf: ---> percentile 99.990 =   60.973
sockperf: ---> percentile 99.900 =   46.296
sockperf: ---> percentile 99.000 =   42.253
sockperf: ---> percentile 90.000 =   38.456
sockperf: ---> percentile 75.000 =   36.052
sockperf: ---> percentile 50.000 =   35.491
sockperf: ---> percentile 25.000 =   34.890
sockperf: ---> <MIN> observation =   12.513
```

PingXie:
nice to see the single subflow optimization, @matttbe! 

1. does the payload size still matter with the single subflow optimization? in other words, will we see higher overhead with a smaller payload size? 64KB is quite big in the Valkey world.

2. given that the patch was submitted very recently and @xbasel's [test results](https://github.com/valkey-io/valkey/pull/1811#issuecomment-2794298076), I think it is still reasonable to disable MPTCP by default for now. we can always revisit this decision in the future.

thoughts?

PR #1816, opened 2025-03-04
Enable large-memory tests solo runs in daily workflow

ranshid:
This change re-introduce large-memory tests to be run as solo tests on daily runs.

the change includes:
1. separate the large-memory tests into separate test block (mainly helpful in workflow manual dispatch cases)
2. place all current large memory tests in run solo blocks in order to prevent potential runner OOM failures.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -15,7 +15,7 @@ on:
         default: "valgrind,sanitizer,tls,freebsd,macos,alpine,32bit,iothreads,ubuntu,rpm-distros,malloc,specific,fortify,reply-schema,arm"
       skiptests:
         description: "tests to skip (delete the ones you wanna keep, do not leave empty)"
-        default: "valkey,modules,sentinel,cluster,unittest"
+        default: "valkey,modules,sentinel,cluster,unittest,large-memory"
       test_args:
         description: "extra test arguments"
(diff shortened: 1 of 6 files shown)


madolson (approved):
No functional concerns, but did you test this?

ranshid:
> No functional concerns, but did you test this?

well, mostly pass, there was one timeout fail which I am re-running to see if this was flaky:
https://github.com/ranshid/valkey/actions/runs/13650915478

ranshid:
> > No functional concerns, but did you test this?
> 
> well, mostly pass, there was one timeout fail which I am re-running to see if this was flaky: https://github.com/ranshid/valkey/actions/runs/13650915478

@madolson / @zuiderkwast the tests are all green now. I think it is safe to merge. @madolson already approved. @zuiderkwast any concerns on your side?

zuiderkwast (approved):
Looks good! No concern.

Just curious: Were all large-memory tests but one already running in solo mode? Only `Module stream XADD big fields doesn't create empty key` was the problem?

ranshid:
> Looks good! No concern.
> 
> Just curious: Were all large-memory tests but one already running in solo mode? Only `Module stream XADD big fields doesn't create empty key` was the problem?

That is what I found

PR #1819, opened 2025-03-05
Fix engine crash on module client blocking during keyspace events

yairgott:
This change enhances user experience and consistency by allowing a module to block a client on keyspace event notifications. Consistency is improved by allowing that reads after writes on the same connection yield expected results. For example, in ValkeySearch, mutations processed earlier on the same connection will be available for search.

The implementation extends `VM_BlockClient` to support blocking clients on keyspace event notifications. Internal clients, LUA clients, clients issueing multi exec and those with the `deny_blocking` flag set are not blocked. Once blocked, a client’s reply is withheld until it is explicitly unblocked.

### DIFF
--- a/src/blocked.c
+++ b/src/blocked.c
@@ -98,6 +98,7 @@ void freeClientBlockingState(client *c) {
     dictRelease(c->bstate->keys);
     zfree(c->bstate);
     c->bstate = NULL;
+    if (c->deferred_reply) commitDeferredReplyBuffer(c);
 }
 
 /* Block a client for the specific operation type. Once the CLIENT_BLOCKED
--- a/src/blocked.c
+++ b/src/blocked.c
@@ -99,7 +99,7 @@ void freeClientBlockingState(client *c) {
     zfree(c->bstate);
     c->bstate = NULL;
 
-    debugServerAssert(!c->deferred_reply);
+    debugServerAssert(!isDeferredReplyEnabled(c));
     commitDeferredReplyBuffer(c, 0);
 }
 
--- a/src/blocked.c
+++ b/src/blocked.c
@@ -98,7 +98,9 @@ void freeClientBlockingState(client *c) {
     dictRelease(c->bstate->keys);
     zfree(c->bstate);
     c->bstate = NULL;
-    if (c->deferred_reply) commitDeferredReplyBuffer(c);
+
+    serverAssert(!c->deferred_reply);
+    commitDeferredReplyBuffer(c, 0);
 }
 
 /* Block a client for the specific operation type. Once the CLIENT_BLOCKED
--- a/src/blocked.c
+++ b/src/blocked.c
@@ -98,9 +98,6 @@ void freeClientBlockingState(client *c) {
     dictRelease(c->bstate->keys);
     zfree(c->bstate);
     c->bstate = NULL;
-
-    debugServerAssert(!isDeferredReplyEnabled(c));
-    commitDeferredReplyBuffer(c, 0);
 }
 
 /* Block a client for the specific operation type. Once the CLIENT_BLOCKED
--- a/src/blocked.c
+++ b/src/blocked.c
@@ -99,7 +99,7 @@ void freeClientBlockingState(client *c) {
     zfree(c->bstate);
     c->bstate = NULL;
 
-    serverAssert(!c->deferred_reply);
+    debugServerAssert(!c->deferred_reply);
     commitDeferredReplyBuffer(c, 0);
 }
 
--- a/src/module.c
+++ b/src/module.c
@@ -198,8 +198,8 @@ typedef struct ValkeyModuleCtx ValkeyModuleCtx;
 #define VALKEYMODULE_CTX_NEW_CLIENT (1 << 7)  /* Free client object when the \
                                                  context is destroyed */
 #define VALKEYMODULE_CTX_CHANNELS_POS_REQUEST (1 << 8)
-#define VALKEYMODULE_CTX_COMMAND (1 << 9) /* Context created to serve a command from call() or AOF (which calls cmd->proc directly) */
-
+#define VALKEYMODULE_CTX_COMMAND (1 << 9)                /* Context created to serve a command from call() or AOF (which calls cmd->proc directly) */
+#define VALKEYMODULE_CTX_KEYSPACE_NOTIFICATION (1 << 10) /* Context created a keyspace notification event */
 
 /* This represents a key opened with VM_OpenKey(). */
 struct ValkeyModuleKey {
@@ -7795,6 +7795,8 @@ void unblockClientFromModule(client *c) {
  * in that case the privdata argument is disregarded, because we pass the
  * reply callback the privdata that is set here while blocking.
  *
+ * For details on return values and error codes, see the comment block for
+ * VM_BlockClient.
  */
 ValkeyModuleBlockedClient *moduleBlockClient(ValkeyModuleCtx *ctx,
                                              ValkeyModuleCmdFunc reply_callback,
@@ -7807,8 +7809,27 @@ ValkeyModuleBlockedClient *moduleBlockClient(ValkeyModuleCtx *ctx,
                                              void *privdata,
                                              int flags) {
     client *c = ctx->client;
+    if (c->flag.blocked || getClientType(c) != CLIENT_TYPE_NORMAL || c->flag.deny_blocking) {
+        /* Early return if duplicate block attempt or client is not normal or
+         * client is set to deny blocking. */
+        errno = ENOTSUP;
+        return NULL;
+    }
+
+    if (ctx->flags & (VALKEYMODULE_CTX_TEMP_CLIENT | VALKEYMODULE_CTX_NEW_CLIENT)) {
+        /* Temporary clients can't be blocked */
+        errno = EINVAL;
+        return NULL;
+    }
+    int is_keyspace_notification = ctx->flags & (VALKEYMODULE_CTX_KEYSPACE_NOTIFICATION);
     int islua = scriptIsRunning();
     int ismulti = server.in_exec;
+    if ((islua || ismulti) && is_keyspace_notification) {
+        /* Avoid blocking within transactions when context initiated by
+         * keyspace notification. */
+        errno = EINVAL;
+        return NULL;
+    }
     initClientBlockingState(c);
 
     c->bstate->module_blocked_handle = zmalloc(sizeof(ValkeyModuleBlockedClient));
@@ -7863,6 +7884,11 @@ ValkeyModuleBlockedClient *moduleBlockClient(ValkeyModuleCtx *ctx,
         } else {
             c->bstate->timeout = timeout;
             blockClient(c, BLOCKED_MODULE);
+        }
+        /* Defer response until after being unblocked for a context originated from
+         * keyspace notification events */
+        if (is_keyspace_notification) {
+            initDeferredReplyBuffer(c);
         }
     }
     return bc;
@@ -8091,14 +8117,27 @@ int moduleTryServeClientBlockedOnKey(client *c, robj *key) {
  *     free_privdata:    called in order to free the private data that is passed
  *                       by ValkeyModule_UnblockClient() call.
  *
- * Note: ValkeyModule_UnblockClient should be called for every blocked client,
- *       even if client was killed, timed-out or disconnected. Failing to do so
- *       will result in memory leaks.
+ * Notes:
+ * 1. ValkeyModule_UnblockClient should be called for every blocked client,
+ * even if client was killed, timed-out or disconnected. Failing to do so
+ * will result in memory leaks.
+ * 2. Attempting to block the client on keyspace event notification in versions
+ *    prior to 8.1.1 leads to a crash.
  *
  * There are some cases where ValkeyModule_BlockClient() cannot be used:
  *
  * 1. If the client is a Lua script.
  * 2. If the client is executing a MULTI block.
+ * 3. If the client is a temporary module client.
+ * 4. If the client is already blocked.
+ *
+ * In cases 1 and 2, a call to ValkeyModule_BlockClient() will **not** block the
+ * client, but instead produce a specific error reply. Note that if the
+ * BlockClient call originated from within a keyspace notification, no error
+ * reply is generated but nullptr is returned while the errno is set to EINVAL.
+ *
+ * In case 3 and 4, a call to ValkeyModule_BlockClient() are no-op, returning
+ * nullptr. errno is set to EINVAL for case 3 while ENOTSUP for case 4.
  *
  * In these cases, a call to ValkeyModule_BlockClient() will **not** block the
  * client, but instead produce a specific error reply.
@@ -8290,6 +8329,12 @@ int moduleClientIsBlockedOnKeys(client *c) {
  * needs to be passed to the client, included but not limited some slow
  * to compute reply or some reply obtained via networking.
  *
+ * Returns VALKEYMODULE_OK on success. On failure, VALKEYMODULE_ERR is returned
+ * and `errno` is set as follows:
+ *
+ * - EINVAL if bc is NULL.
+ * - ENOTSUP if bc contains `blocked on keys` but its timeout callback is NULL.
+ *
  * Note 1: this function can be called from threads spawned by the module.
  *
  * Note 2: when we unblock a client that is blocked for keys using the API
@@ -8300,10 +8345,17 @@ int moduleClientIsBlockedOnKeys(client *c) {
  * ValkeyModule_BlockClientOnKeys() is accessible from the timeout
  * callback via VM_GetBlockedClientPrivateData). */
 int VM_UnblockClient(ValkeyModuleBlockedClient *bc, void *privdata) {
+    if (!bc) {
+        errno = EINVAL;
+        return VALKEYMODULE_ERR;
+    }
     if (bc->blocked_on_keys) {
         /* In theory the user should always pass the timeout handler as an
          * argument, but better to be safe than sorry. */
-        if (bc->timeout_callback == NULL) return VALKEYMODULE_ERR;
+        if (bc->timeout_callback == NULL) {
+            errno = ENOTSUP;
+            return VALKEYMODULE_ERR;
+        }
         if (bc->unblocked) return VALKEYMODULE_OK;
         if (bc->client) moduleBlockedClientTimedOut(bc->client, 1);
     }
@@ -8392,11 +8444,17 @@ void moduleHandleBlockedClients(void) {
             moduleInvokeFreePrivDataCallback(c, bc);
         }
 
-        /* It is possible that this blocked client object accumulated
-         * replies to send to the client in a thread safe context.
-         * We need to glue such replies to the client output buffer and
-         * free the temporary client we just used for the replies. */
-        if (c) AddReplyFromClient(c, bc->reply_client);
+        if (c) {
+            /* Replies which were added after the client is blocked by a module
+             * are accumulated separately. We need to transmit those replies
+             * to the client. */
+            commitDeferredReplyBuffer(c);
+            /* It is possible that this blocked client object accumulated
+             * replies to send to the client in a thread safe context.
+             * We need to glue such replies to the client output buffer and
+             * free the temporary client we just used for the replies. */
+            AddReplyFromClient(c, bc->reply_client);
+        }
         moduleReleaseTempClient(bc->reply_client);
         moduleReleaseTempClient(bc->thread_safe_ctx_client);
 
@@ -8492,9 +8550,10 @@ void moduleBlockedClientTimedOut(client *c, int from_module) {
 
     moduleFreeContext(&ctx);
 
-    if (!from_module)
+    if (!from_module) {
         updateStatsOnUnblock(c, bc->background_duration, 0,
                              ((server.stat_total_error_replies != prev_error_replies) ? ERROR_COMMAND_FAILED : 0));
+    }
 
     /* For timeout events, we do not want to call the disconnect callback,
      * because the blocked client will be automatically disconnected in
@@ -8880,8 +8939,14 @@ void moduleNotifyKeyspaceEvent(int type, const char *event, robj *key, int dbid)
         if ((sub->event_mask & type) &&
             (sub->active == 0 || (sub->module->options & VALKEYMODULE_OPTIONS_ALLOW_NESTED_KEYSPACE_NOTIFICATIONS))) {
             ValkeyModuleCtx ctx;
-            moduleCreateContext(&ctx, sub->module, VALKEYMODULE_CTX_TEMP_CLIENT);
+            if (server.executing_client == NULL) {
+                moduleCreateContext(&ctx, sub->module, VALKEYMODULE_CTX_TEMP_CLIENT);
+            } else {
+                moduleCreateContext(&ctx, sub->module, VALKEYMODULE_CTX_NONE);
+                ctx.client = server.executing_client;
+            }
             selectDb(ctx.client, dbid);
+            ctx.flags |= VALKEYMODULE_CTX_KEYSPACE_NOTIFICATION;
 
             /* mark the handler as active to avoid reentrant loops.
              * If the subscriber performs an action triggering itself,
--- a/src/module.c
+++ b/src/module.c
@@ -8899,12 +8899,16 @@ int VM_NotifyKeyspaceEvent(ValkeyModuleCtx *ctx, int type, const char *event, Va
     return VALKEYMODULE_OK;
 }
 
+unsigned long moduleNotifyKeyspaceSubscribersCnt(void) {
+    return listLength(moduleKeyspaceSubscribers);
+}
+
 /* Dispatcher for keyspace notifications to module subscriber functions.
  * This gets called  only if at least one module requested to be notified on
  * keyspace notifications */
 void moduleNotifyKeyspaceEvent(int type, const char *event, robj *key, int dbid) {
     /* Don't do anything if there aren't any subscribers */
-    if (listLength(moduleKeyspaceSubscribers) == 0) return;
+    if (moduleNotifyKeyspaceSubscribersCnt() == 0) return;
 
     /* Ugly hack to handle modules which use write commands from within
      * notify_callback, which they should NOT do!
--- a/src/module.c
+++ b/src/module.c
@@ -8448,7 +8448,7 @@ void moduleHandleBlockedClients(void) {
             /* Replies which were added after the client is blocked by a module
              * are accumulated separately. We need to transmit those replies
              * to the client. */
-            commitDeferredReplyBuffer(c);
+            commitDeferredReplyBuffer(c, 0);
             /* It is possible that this blocked client object accumulated
              * replies to send to the client in a thread safe context.
              * We need to glue such replies to the client output buffer and
--- a/src/networking.c
+++ b/src/networking.c
@@ -149,6 +149,12 @@ static inline int isReplicaReadyForReplData(client *replica) {
            !(replica->flag.close_asap);
 }
 
+static void resetDeferredReplyBuffer(client *c) {
+    listRelease(c->deferred_reply);
+    c->deferred_reply = NULL;
+    c->deferred_reply_bytes = 0;
+}
+
 client *createClient(connection *conn) {
     client *c = zmalloc(sizeof(client));
 
@@ -203,8 +209,10 @@ client *createClient(connection *conn) {
     c->duration = 0;
     clientSetDefaultAuth(c);
     c->reply = listCreate();
+    c->deferred_reply = NULL;
     c->deferred_reply_errors = NULL;
     c->reply_bytes = 0;
+    c->deferred_reply_bytes = 0;
     c->obuf_soft_limit_reached_time = 0;
     listSetFreeMethod(c->reply, freeClientReplyValue);
     listSetDupMethod(c->reply, dupClientReplyValue);
@@ -277,7 +285,6 @@ void putClientInPendingWriteQueue(client *c) {
         listLinkNodeHead(server.clients_pending_write, &c->clients_pending_write_node);
     }
 }
-
 /* This function is called every time we are going to transmit new data
  * to the client. The behavior is the following:
  *
@@ -326,6 +333,7 @@ int prepareClientToWrite(client *c) {
      * it should already be setup to do so (it has already pending data). */
     if (!clientHasPendingReplies(c)) putClientInPendingWriteQueue(c);
 
+    if (c->deferred_reply == NULL) c->flag.buffered_reply = 1;
     /* Authorize the caller to queue in the output buffer of this client. */
     return C_OK;
 }
@@ -425,13 +433,15 @@ void _addReplyProtoToList(client *c, list *reply_list, const char *s, size_t len
          * least PROTO_REPLY_CHUNK_BYTES */
         size_t usable_size;
         size_t size = len < PROTO_REPLY_CHUNK_BYTES ? PROTO_REPLY_CHUNK_BYTES : len;
+        if (c->deferred_reply) size = len;
         tail = zmalloc_usable(size + sizeof(clientReplyBlock), &usable_size);
         /* take over the allocation's internal fragmentation */
         tail->size = usable_size - sizeof(clientReplyBlock);
         tail->used = len;
         memcpy(tail->buf, s, len);
         listAddNodeTail(reply_list, tail);
-        c->reply_bytes += tail->size;
+        unsigned long long *reply_bytes = (c->deferred_reply) ? &c->deferred_reply_bytes : &c->reply_bytes;
+        *reply_bytes += tail->size;
 
         closeClientOnOutputBufferLimitReached(c, 1);
     }
@@ -462,6 +472,11 @@ void _addReplyToBufferOrList(client *c, const char *s, size_t len) {
 
     c->net_output_bytes_curr_cmd += len;
 
+    if (c->deferred_reply) {
+        _addReplyProtoToList(c, c->deferred_reply, s, len);
+        return;
+    }
+
     /* We call it here because this function may affect the reply
      * buffer offset (see function comment) */
     reqresSaveClientReplyOffset(c);
@@ -477,7 +492,6 @@ void _addReplyToBufferOrList(client *c, const char *s, size_t len) {
         _addReplyProtoToList(c, server.pending_push_messages, s, len);
         return;
     }
-
     size_t reply_len = _addReplyToBuffer(c, s, len);
     if (len > reply_len) _addReplyProtoToList(c, c->reply, s + reply_len, len - reply_len);
 }
@@ -564,6 +578,7 @@ void afterErrorReply(client *c, const char *s, size_t len, int flags) {
         return;
     }
 
+    commitDeferredReplyBuffer(c);
     if (!(flags & ERR_REPLY_FLAG_NO_STATS_UPDATE)) {
         /* Increment the global error counter */
         server.stat_total_error_replies++;
@@ -1283,6 +1298,32 @@ void addReplySubcommandSyntaxError(client *c) {
     sdsfree(cmd);
 }
 
+void initDeferredReplyBuffer(client *c) {
+    if (c->deferred_reply == NULL) {
+        c->deferred_reply = listCreate();
+    }
+}
+
+/* Move the client deferred reply buffer into the client reply buffer and put the client
+ * in the pending write queue. */
+void commitDeferredReplyBuffer(client *c) {
+    if (c->deferred_reply == NULL || listLength(c->deferred_reply) == 0) {
+        resetDeferredReplyBuffer(c);
+        return;
+    }
+
+    listJoin(c->reply, c->deferred_reply);
+    c->reply_bytes += c->deferred_reply_bytes;
+
+    resetDeferredReplyBuffer(c);
+    if (prepareClientToWrite(c) != C_OK) {
+        return;
+    }
+    /* We call it here because this function may affect the reply
+     * buffer offset (see function comment) */
+    reqresSaveClientReplyOffset(c);
+}
+
 /* Append 'src' client output buffers into 'dst' client output buffers.
  * This function clears the output buffers of 'src' */
 void AddReplyFromClient(client *dst, client *src) {
@@ -1299,6 +1340,7 @@ void AddReplyFromClient(client *dst, client *src) {
         return;
     }
 
+
     /* First add the static buffer (either into the static buffer or reply list) */
     addReplyProto(dst, src->buf, src->bufpos);
 
@@ -1310,6 +1352,7 @@ void AddReplyFromClient(client *dst, client *src) {
      * checks in it. */
     if (dst->flag.close_after_reply) return;
 
+    commitDeferredReplyBuffer(src);
     /* Concatenate the reply list into the dest */
     if (listLength(src->reply)) listJoin(dst->reply, src->reply);
     dst->reply_bytes += src->reply_bytes;
@@ -1737,6 +1780,7 @@ void freeClient(client *c) {
     c->reply = NULL;
     zfree_with_size(c->buf, c->buf_usable_size);
     c->buf = NULL;
+    resetDeferredReplyBuffer(c);
 
     freeClientArgv(c);
     freeClientOriginalArgv(c);
@@ -2579,6 +2623,8 @@ void resetClient(client *c) {
     c->slot = -1;
     c->flag.executing_command = 0;
     c->flag.replication_done = 0;
+    c->flag.buffered_reply = 0;
+    c->flag.keyspace_notified = 0;
     c->net_output_bytes_curr_cmd = 0;
 
     /* Make sure the duration has been recorded to some command. */
@@ -2589,6 +2635,7 @@ void resetClient(client *c) {
 
     if (c->deferred_reply_errors) listRelease(c->deferred_reply_errors);
     c->deferred_reply_errors = NULL;
+    commitDeferredReplyBuffer(c);
 
     /* We clear the ASKING flag as well if we are not inside a MULTI, and
      * if what we just executed is not the ASKING command itself. */
@@ -4478,10 +4525,10 @@ void rewriteClientCommandVector(client *c, int argc, ...) {
 
 /* Completely replace the client command vector with the provided one. */
 void replaceClientCommandVector(client *c, int argc, robj **argv) {
-    int j;
     backupAndUpdateClientArgv(c, argc, argv);
     c->argv_len_sum = 0;
-    for (j = 0; j < c->argc; j++)
+    c->flag.buffered_reply = 0;
+    for (int j = 0; j < c->argc; j++)
         if (c->argv[j]) c->argv_len_sum += getStringObjectLen(c->argv[j]);
     c->cmd = lookupCommandOrOriginal(c->argv, c->argc);
     serverAssertWithInfo(c, NULL, c->cmd != NULL);
@@ -4512,6 +4559,7 @@ void rewriteClientCommandArgument(client *c, int i, robj *newval) {
 
     /* If this is the command name make sure to fix c->cmd. */
     if (i == 0) {
+        c->flag.buffered_reply = 0;
         c->cmd = lookupCommandOrOriginal(c->argv, c->argc);
         serverAssertWithInfo(c, NULL, c->cmd != NULL);
     }
@@ -4535,10 +4583,15 @@ size_t getClientOutputBufferMemoryUsage(client *c) {
             repl_node_num = last->id - cur->id + 1;
         }
         return repl_buf_size + (repl_node_size * repl_node_num);
-    } else {
-        size_t list_item_size = sizeof(listNode) + sizeof(clientReplyBlock);
-        return c->reply_bytes + (list_item_size * listLength(c->reply));
     }
+
+    size_t list_item_size = sizeof(listNode) + sizeof(clientReplyBlock);
+    size_t usage = c->reply_bytes + (list_item_size * listLength(c->reply));
+    if (c->deferred_reply) {
+        usage += c->deferred_reply_bytes +
+                 (list_item_size * listLength(c->deferred_reply));
+    }
+    return usage;
 }
 
 /* Returns the total client's memory usage.
--- a/src/networking.c
+++ b/src/networking.c
@@ -1782,7 +1782,7 @@ void freeClient(client *c) {
     c->reply = NULL;
     zfree_with_size(c->buf, c->buf_usable_size);
     c->buf = NULL;
-    resetDeferredReplyBuffer(c);
+    listRelease(c->deferred_reply);
 
     freeClientArgv(c);
     freeClientOriginalArgv(c);
--- a/src/networking.c
+++ b/src/networking.c
@@ -1342,7 +1342,6 @@ void AddReplyFromClient(client *dst, client *src) {
         return;
     }
 
-
     /* First add the static buffer (either into the static buffer or reply list) */
     addReplyProto(dst, src->buf, src->bufpos);
 
--- a/src/networking.c
+++ b/src/networking.c
@@ -473,7 +473,7 @@ void _addReplyToBufferOrList(client *c, const char *s, size_t len) {
      * The check for executing_client also avoids affecting push messages that are part of eviction.
      * Check CLIENT_PUSHING first to avoid race conditions, as it's absent in module's fake client. */
     int defer_push_message = c->flag.pushing && c == server.current_client && server.executing_client &&
-        !cmdHasPushAsReply(server.executing_client->cmd);
+                             !cmdHasPushAsReply(server.executing_client->cmd);
     if (defer_push_message == 0 && isDeferredReplyEnabled(c)) {
         _addReplyProtoToList(c, c->deferred_reply, s, len);
         return;
--- a/src/networking.c
+++ b/src/networking.c
@@ -466,23 +466,23 @@ void _addReplyToBufferOrList(client *c, const char *s, size_t len) {
 
     c->net_output_bytes_curr_cmd += len;
 
-    if (isDeferredReplyEnabled(c)) {
-        _addReplyProtoToList(c, c->deferred_reply, s, len);
-        return;
-    }
-
-    /* We call it here because this function may affect the reply
-     * buffer offset (see function comment) */
-    reqresSaveClientReplyOffset(c);
-
     /* If we're processing a push message into the current client (i.e. executing PUBLISH
      * to a channel which we are subscribed to, then we wanna postpone that message to be added
      * after the command's reply (specifically important during multi-exec). the exception is
      * the SUBSCRIBE command family, which (currently) have a push message instead of a proper reply.
      * The check for executing_client also avoids affecting push messages that are part of eviction.
      * Check CLIENT_PUSHING first to avoid race conditions, as it's absent in module's fake client. */
-    if (c->flag.pushing && c == server.current_client && server.executing_client &&
-        !cmdHasPushAsReply(server.executing_client->cmd)) {
+    int defer_push_message = c->flag.pushing && c == server.current_client && server.executing_client &&
+        !cmdHasPushAsReply(server.executing_client->cmd);
+    if (defer_push_message == 0 && isDeferredReplyEnabled(c)) {
+        _addReplyProtoToList(c, c->deferred_reply, s, len);
+        return;
+    }
+    /* We call it here because this function may affect the reply
+     * buffer offset (see function comment) */
+    reqresSaveClientReplyOffset(c);
+
+    if (defer_push_message) {
         _addReplyProtoToList(c, server.pending_push_messages, s, len);
         return;
     }
--- a/src/networking.c
+++ b/src/networking.c
@@ -578,7 +578,7 @@ void afterErrorReply(client *c, const char *s, size_t len, int flags) {
         return;
     }
 
-    commitDeferredReplyBuffer(c);
+    commitDeferredReplyBuffer(c, 1);
     if (!(flags & ERR_REPLY_FLAG_NO_STATS_UPDATE)) {
         /* Increment the global error counter */
         server.stat_total_error_replies++;
@@ -1306,7 +1306,9 @@ void initDeferredReplyBuffer(client *c) {
 
 /* Move the client deferred reply buffer into the client reply buffer and put the client
  * in the pending write queue. */
-void commitDeferredReplyBuffer(client *c) {
+void commitDeferredReplyBuffer(client *c, int skip_if_blocked) {
+    if (skip_if_blocked && c->flag.blocked) return;
+
     if (c->deferred_reply == NULL || listLength(c->deferred_reply) == 0) {
         resetDeferredReplyBuffer(c);
         return;
@@ -1352,7 +1354,6 @@ void AddReplyFromClient(client *dst, client *src) {
      * checks in it. */
     if (dst->flag.close_after_reply) return;
 
-    commitDeferredReplyBuffer(src);
     /* Concatenate the reply list into the dest */
     if (listLength(src->reply)) listJoin(dst->reply, src->reply);
     dst->reply_bytes += src->reply_bytes;
@@ -2635,7 +2636,7 @@ void resetClient(client *c) {
 
     if (c->deferred_reply_errors) listRelease(c->deferred_reply_errors);
     c->deferred_reply_errors = NULL;
-    commitDeferredReplyBuffer(c);
+    commitDeferredReplyBuffer(c, 1);
 
     /* We clear the ASKING flag as well if we are not inside a MULTI, and
      * if what we just executed is not the ASKING command itself. */
--- a/src/networking.c
+++ b/src/networking.c
@@ -1296,6 +1296,12 @@ inline int isDeferredReplyEnabled(client *c) {
     return c->deferred_reply_bytes != ULLONG_MAX;
 }
 
+/* Commands that generate replies before triggering keyspace notifications must
+ * use a deferred reply buffer. This allows postponing the actual transmission
+ * of the reply until after the client is unblocked, in case it was blocked by
+ * a keyspace notification. This is necessary because modules subscribed to
+ * keyspace notifications can block the client from within the notification
+ * callback. */
 void initDeferredReplyBuffer(client *c) {
     if (moduleNotifyKeyspaceSubscribersCnt() == 0) return;
     if (c->deferred_reply == NULL) c->deferred_reply = listCreate();
--- a/src/networking.c
+++ b/src/networking.c
@@ -149,12 +149,6 @@ static inline int isReplicaReadyForReplData(client *replica) {
            !(replica->flag.close_asap);
 }
 
-static void resetDeferredReplyBuffer(client *c) {
-    listRelease(c->deferred_reply);
-    c->deferred_reply = NULL;
-    c->deferred_reply_bytes = 0;
-}
-
 client *createClient(connection *conn) {
     client *c = zmalloc(sizeof(client));
 
@@ -212,7 +206,7 @@ client *createClient(connection *conn) {
     c->deferred_reply = NULL;
     c->deferred_reply_errors = NULL;
     c->reply_bytes = 0;
-    c->deferred_reply_bytes = 0;
+    c->deferred_reply_bytes = ULLONG_MAX;
     c->obuf_soft_limit_reached_time = 0;
     listSetFreeMethod(c->reply, freeClientReplyValue);
     listSetDupMethod(c->reply, dupClientReplyValue);
@@ -333,7 +327,7 @@ int prepareClientToWrite(client *c) {
      * it should already be setup to do so (it has already pending data). */
     if (!clientHasPendingReplies(c)) putClientInPendingWriteQueue(c);
 
-    if (c->deferred_reply == NULL) c->flag.buffered_reply = 1;
+    if (!isDeferredReplyEnabled(c)) c->flag.buffered_reply = 1;
     /* Authorize the caller to queue in the output buffer of this client. */
     return C_OK;
 }
@@ -433,14 +427,14 @@ void _addReplyProtoToList(client *c, list *reply_list, const char *s, size_t len
          * least PROTO_REPLY_CHUNK_BYTES */
         size_t usable_size;
         size_t size = len < PROTO_REPLY_CHUNK_BYTES ? PROTO_REPLY_CHUNK_BYTES : len;
-        if (c->deferred_reply) size = len;
+        if (isDeferredReplyEnabled(c)) size = len;
         tail = zmalloc_usable(size + sizeof(clientReplyBlock), &usable_size);
         /* take over the allocation's internal fragmentation */
         tail->size = usable_size - sizeof(clientReplyBlock);
         tail->used = len;
         memcpy(tail->buf, s, len);
         listAddNodeTail(reply_list, tail);
-        unsigned long long *reply_bytes = (c->deferred_reply) ? &c->deferred_reply_bytes : &c->reply_bytes;
+        unsigned long long *reply_bytes = (isDeferredReplyEnabled(c)) ? &c->deferred_reply_bytes : &c->reply_bytes;
         *reply_bytes += tail->size;
 
         closeClientOnOutputBufferLimitReached(c, 1);
@@ -472,7 +466,7 @@ void _addReplyToBufferOrList(client *c, const char *s, size_t len) {
 
     c->net_output_bytes_curr_cmd += len;
 
-    if (c->deferred_reply) {
+    if (isDeferredReplyEnabled(c)) {
         _addReplyProtoToList(c, c->deferred_reply, s, len);
         return;
     }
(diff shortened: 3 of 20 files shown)


tests/modules/block_keyspace_notification.c
```diff
@@ -0,0 +1,148 @@
...
+#define _DEFAULT_SOURCE /* For usleep */
+#include "valkeymodule.h"
+#include <stdio.h>
+#include <string.h>
+#include <strings.h>
+#include <pthread.h>
+#include <unistd.h>
+#define EVENT_LOG_MAX_SIZE 1024
+static pthread_mutex_t event_log_mutex = PTHREAD_MUTEX_INITIALIZER;
+typedef struct KeyspaceEventData {
```
enjoy-binbin, line 15: please fix the format first, thest test codes (.c / .tcl) need some spaces between them

tests/unit/moduleapi/block_keyspace_notification.tcl
```diff
@@ -0,0 +1,80 @@
...
+        }
+    	assert_match "OK" [r b_keyspace.clear]
+        # Expiration should happen in the background
+        wait_for_condition 1000 50 {
+            [r b_keyspace.events] eq "{event expired key h}"
+        } else {
+            fail "Expired event not propagated within 5 seconds"
+        }
+    }
+    test "Unload the module - blockedclient" {
```
enjoy-binbin:
```suggestion
    test "Unload the module - testblockingkeyspacenotif" {
```

src/t_zset.c
```diff
@@ -1851,32 +1853,35 @@ static void zaddGenericCommand(client *c, int flags) {
...
         if (!(retflags & ZADD_OUT_NOP)) processed++;
         score = newscore;
     }
-    server.dirty += (added + updated);
+    if (!reply_err) {
+        server.dirty += (added + updated);
+    }
+    if (added || updated) {
+        signalModifiedKey(c, c->db, key);
+        notifyKeyspaceEvent(c, NOTIFY_ZSET, incr ? "zincr" : "zadd", key, c->db->id);
```
enjoy-binbin: So now we need to notify event before addReply, right? We may need to add an assert in notifyKeyspaceEvent to check that the client has not replied.

And this event block cannot return any errors? we need to document it.

madolson: I also like the idea of an assertion. I would only add it should probably be a debug assertion, since I don't really want to crash the server in production because of it.

src/module.c
```diff
@@ -7807,8 +7807,24 @@ ValkeyModuleBlockedClient *moduleBlockClient(ValkeyModuleCtx *ctx,
...
+    }
+
+    if (ctx->flags & (VALKEYMODULE_CTX_TEMP_CLIENT | VALKEYMODULE_CTX_NEW_CLIENT)) {
+        /* Temporary clients can't be blocked */
+        return NULL;
+    }
+    int iskeyspacenotification = ctx->flags & (VALKEYMODULE_CTX_KEYSPACE_NOTIFICATION);
     int islua = scriptIsRunning();
     int ismulti = server.in_exec;
+    if ((islua || ismulti) && iskeyspacenotification) {
```
madolson: Don't you need to support multi (and lua too I guess) for your workload? The client could multi exec two writes and it won't block subsequent reads. We could feasibly block the entire multi-block's response.

src/networking.c
```diff
@@ -303,7 +303,7 @@ void putClientInPendingWriteQueue(client *c) {
 int prepareClientToWrite(client *c) {
     /* If it's the Lua client we always return ok without installing any
      * handler since there is no socket at all. */
-    if (c->flag.script || c->flag.module) return C_OK;
+    if (c->flag.script || c->flag.module || c->flag.blocked_by_module) return C_OK;
```
madolson: I was expecting more changes in networking.c. If a client was already prepared to write (either it already was installed on an earlier event loop or their was a previous pipelined request in the same event loop) won't the reply still get written out?

tests/unit/moduleapi/block_keyspace_notification.tcl
```diff
@@ -0,0 +1,80 @@
+
+set testmodule [file normalize tests/modules/block_keyspace_notification.so]
+start_server {tags {"modules"}} {
+    r module load $testmodule
+    test {Blocking keyspace notification one} {
+    	assert_match "OK" [r b_keyspace.clear]
+        assert_match "1" [r hset a b c]
+        assert_equal "{event hset key a}" [r b_keyspace.events]
+    }
+    test {Blocking keyspace notification two} {
```
madolson, line 12: We should probably also introduce a test that does pipelining to make sure we only block the ingested writes and not other reads.

src/db.c
```diff
@@ -1850,7 +1850,7 @@ void deleteExpiredKeyFromOverwriteAndPropagate(client *c, robj *keyobj) {
     robj *aux = server.lazyfree_lazy_expire ? shared.unlink : shared.del;
     rewriteClientCommandVector(c, 2, aux, keyobj);
     signalModifiedKey(c, c->db, keyobj);
-    notifyKeyspaceEvent(NOTIFY_EXPIRED, "expired", keyobj, c->db->id);
+    notifyKeyspaceEvent(NULL, NOTIFY_EXPIRED, "expired", keyobj, c->db->id);
```
QuChen88: Should this notify client `c`?

src/module.c
```diff
@@ -7807,8 +7807,24 @@ ValkeyModuleBlockedClient *moduleBlockClient(ValkeyModuleCtx *ctx,
...
+        /* Early return if duplicate block attempt or client is not normal or
+         * client is set to deny blocking. */
+        return NULL;
+    }
+
+    if (ctx->flags & (VALKEYMODULE_CTX_TEMP_CLIENT | VALKEYMODULE_CTX_NEW_CLIENT)) {
+        /* Temporary clients can't be blocked */
+        return NULL;
+    }
+    int iskeyspacenotification = ctx->flags & (VALKEYMODULE_CTX_KEYSPACE_NOTIFICATION);
```
QuChen88: nitpick: `is_keyspace_notification`

src/module.c
yairgott: multi: ValkeySearch has a special mechanism to honor multi semantics which does not rely on blocking the client. I can elaborate on this but it seems off topic but I would just say that it's an expensive code path. I don't see an issue with allowing such for a keyspace notification context. That being said, it might be safer to just return NULL. I'm also open to change this. 

Lua: ValkeySearch doesn't has special handling for Lua but it should and it would probably be very similar to multi . At least in the version that we initially developed this feature, blocking a Lua context resulted in segfault. I can re-evaluate this but, similarly to multi, it might be the safer option for now to just return NULL.

Noting that the existing code responds with an error message if the block was called as part of a multi or lua. Responding with an error seems to me a bad behavior as it's a logical internal error. We might want to revisit this. There are some additional flaws to the current implementation that we may need to address in the future.

src/module.c
```diff
@@ -8099,6 +8120,16 @@ int moduleTryServeClientBlockedOnKey(client *c, robj *key) {
  *
  * 1. If the client is a Lua script.
  * 2. If the client is executing a MULTI block.
+ * 3. If the client is a temporary module client.
+ * 4. If the client is already blocked.
+ *
+ * In cases 1 and 2, a call to VlakeyModule_BlockClient() will **not** block the
```
QuChen88: nitpick: `ValkeyModule_BlockClient()`

src/module.c
```diff
@@ -7807,8 +7807,24 @@ ValkeyModuleBlockedClient *moduleBlockClient(ValkeyModuleCtx *ctx,
                                              void *privdata,
                                              int flags) {
     client *c = ctx->client;
+    if (c->flag.blocked || getClientType(c) != CLIENT_TYPE_NORMAL || c->flag.deny_blocking) {
```
QuChen88, line 7812: Can replication client (i.e. primary or replica) get blocked?

src/networking.c
yairgott: `prepareClientToWrite` is called by `addReply*` so it's important to ensure that `addReply*` is called after the client is blocked.

src/t_zset.c
yairgott: Adding debug mode assertion sounds good! Can you elaborate which signal can I use for this? Noting that the client may have pending writes from previous command processing.

Noting that in the existing code, sometimes `addReply` is called before `notifyKeyspaceEvent` and sometimes it's vise-versa. This PR makes `addReply` be called after `notifyKeyspaceEvent` consistently. 

> And this event block cannot return any errors? we need to document it.

Not sure that I fully understand the comment. Noting that the event handling can't reply to the user as the client context is not available to the module.

tests/modules/block_keyspace_notification.c
yairgott: done

tests/unit/moduleapi/block_keyspace_notification.tcl
yairgott: hmm, I'm not sure that I fully understand. Noting that blocking a client wouldn't hold back processing sub-sequential commands but their responses wouldn't be transmitted back until the client is unblocked.

src/db.c
yairgott: Good catch!

src/module.c
yairgott: done

src/module.c
yairgott: done

tests/unit/moduleapi/block_keyspace_notification.tcl
yairgott: done

src/module.c
yairgott: Yes, check that the client type is CLIENT_TYPE_NORMAL ensures this: https://github.com/valkey-io/valkey/blob/unstable/src/server.h#L374

src/networking.c
madolson: Maybe I'm misunderstanding this check, but I thought it was to bypass installing the write handler. In which case, the client could already have a write handler installed from an earlier read.

I also assume this check should be lower (like I'm not sure why we're placing the data in the CoB when close_asap or reply_off are configured).

tests/unit/moduleapi/block_keyspace_notification.tcl
yairgott: Anyway, added a pipeline test case

tests/unit/moduleapi/block_keyspace_notification.tcl
```diff
@@ -0,0 +1,99 @@
...
+            $rd1 hset key_1 field_1 value_1
+            $rd1 hget key_1 field_1
+            wait_for_blocked_clients_count 1
+            assert_equal "" [r b_keyspace.events]
+            wait_for_blocked_clients_count 0
+            assert_equal "{event hset key key_1}" [r b_keyspace.events]
+            assert_equal "1" [$rd1 read]
+            assert_equal "value_1" [$rd1 read]
+        }
+    }
```
madolson:
```suggestion
    test {Blocking keyspace notification with pipelining (skip if Valkey is not 7.2.7)} {
        set i [r info]
        regexp {valkey_version:(.*?)\r\n} $i - version
        if { $version eq "7.2.7" } {
            puts "Skipping test: Valkey version is 7.2.7"
        } else {
            assert_match "OK" [r b_keyspace.clear]
            set rd1 [valkey_deferring_client]
            pause_process [srv 0 pid]
            # Queue up two commands
            $rd1 hget key_1 field_1
            $rd1 hset key_1 field_1 value_1
            set start [clock milliseconds]; # Record time before unblocking
            resume_process [srv 0 pid]
            puts [$rd1 read]; read HGET response
            puts [$rd1 read]; read HSET RESPONSE
            puts [expr [clock milliseconds] - $start]; # Print time after both reads are done, so this is after unblocking
            assert_equal "" [r b_keyspace.events]            
            wait_for_blocked_clients_count 0
            assert_equal "{event hset key key_1}" [r b_keyspace.events]
            assert_equal "1" [$rd1 read]
            assert_equal "value_1" [$rd1 read]
        }
    }
```

I think if you make this change highlights the nuance I'm trying to get after. If you have a read and a write in the same EPOLL, the HSET will have it's content written out immediately without blocking. (you'll see about 20ms of time) If you swap the hget and hset, you see the intended result of about 1 second.

src/networking.c
```diff
@@ -324,7 +326,9 @@ int prepareClientToWrite(client *c) {
 
     /* Schedule the client to write the output buffers to the socket, unless
      * it should already be setup to do so (it has already pending data). */
-    if (!clientHasPendingReplies(c)) putClientInPendingWriteQueue(c);
+    if (!clientHasPendingReplies(c)) {
+        putClientInPendingWriteQueue(c);
+    }
```
```suggestion
    if (!clientHasPendingReplies(c)) putClientInPendingWriteQueue(c);
```
Just to keep the code the same.

src/networking.c
```diff
@@ -303,7 +305,7 @@ void putClientInPendingWriteQueue(client *c) {
 int prepareClientToWrite(client *c) {
     /* If it's the Lua client we always return ok without installing any
      * handler since there is no socket at all. */
-    if (c->flag.script || c->flag.module) return C_OK;
+    if (c->flag.script || c->flag.module || c->blocked_client_pending_reply != NULL) return C_OK;
```
madolson: I still think this check is too early in this function. I think we should be respecting `CLIENT REPLY OFF` and the client being closed. I don't really see a reason not to put it right above `putClientInPendingWriteQueue`.

src/networking.c
```diff
@@ -373,6 +377,15 @@ void deleteCachedResponseClient(client *recording_client) {
     freeClient(recording_client);
 }
 
+static void resetBlockedClientPendingReply(client *c) {
+    c->blocked_client_pending_reply_bytes = 0;
+    if (c->blocked_client_pending_reply == NULL) {
+        return;
+    }
+    listRelease(c->blocked_client_pending_reply);
+    c->blocked_client_pending_reply = NULL;
```
```suggestion
    listRelease(c->blocked_client_pending_reply);
    c->blocked_client_pending_reply = NULL;
```
Most Valkey functions gracefully handle freeing NULL pointers.

tests/unit/moduleapi/block_keyspace_notification.tcl
```diff
@@ -0,0 +1,136 @@
...
+        assert_equal "{event hset key b}" [r b_keyspace.events]
+        wait_for_blocked_clients_count 0
+        assert_equal "{event hset key b} {event hset key c}" [r b_keyspace.events]
+    }
+    test {Blocking keyspace notification with pipelining hset after hget (skip if Valkey is not 7.2.7)} {
+        set i [r info]
+        regexp {valkey_version:(.*?)\r\n} $i - version
+        if { $version eq "7.2.7" } {
+            puts "Skipping test: Valkey version is 7.2.7"
+        } else {
```
madolson: Ideally we would build a generic mechanism for requiring a minimum server version. For now I think you can just do the check once at the top of the file and return immediately if the version fails.

tests/modules/block_keyspace_notification.c
```diff
@@ -0,0 +1,159 @@
...
+        return VALKEYMODULE_ERR;
+    }
+    event_log = ValkeyModule_Alloc(sizeof(KeyspaceEventLog));
+    event_log->next_index = 0;
+    int keySpaceAll = ValkeyModule_GetKeyspaceNotificationFlagsAll();
+    if (!(keySpaceAll & VALKEYMODULE_NOTIFY_LOADED)) {
+        // VALKEYMODULE_NOTIFY_LOADED event are not supported we can not start
+        return VALKEYMODULE_ERR;
+    }
+    if(ValkeyModule_SubscribeToKeyspaceEvents(ctx, VALKEYMODULE_NOTIFY_LOADED, KeySpace_NotificationGeneric) != VALKEYMODULE_OK){
```
madolson: These can all be OR'd together, should make this code block a bit more readable.

src/module.c
```diff
@@ -8099,6 +8120,16 @@ int moduleTryServeClientBlockedOnKey(client *c, robj *key) {
  *
  * 1. If the client is a Lua script.
  * 2. If the client is executing a MULTI block.
+ * 3. If the client is a temporary module client.
+ * 4. If the client is already blocked.
+ *
+ * In cases 1 and 2, a call to ValkeyModule_BlockClient() will **not** block the
```
madolson, line 8134: We need to document somewhere that you can call ValkeyModule_BlockClient() inside a keyspace notification. The comments on these functions become the documentation on the website: https://valkey.io/topics/modules-api-ref/.

tests/unit/moduleapi/block_keyspace_notification.tcl
```diff
@@ -0,0 +1,124 @@
...
+        test {Blocking keyspace notification one} {
+            assert_match "OK" [r b_keyspace.clear]
+            assert_match "1" [r hset a b c]
+            assert_equal "{event hset key a}" [r b_keyspace.events]
+        }
+        test {Blocking keyspace notification two} {
+            assert_match "OK" [r b_keyspace.clear]
+            set rd1 [valkey_deferring_client]
+            $rd1 hset b c d
+            after 500
```
madolson: I don't like using after because it often leads to long tests and flakiness. I would prefer we change the test module to add an API that unblocks all pending clients, so you can very carefully sequence events. This is not a blocker for me right now though.

madolson (commented):
Ok, I'm aligned with the design now. Just some other misc comments.

src/networking.c
yairgott: This is actually not required now with the new design. I removed it.

tests/unit/moduleapi/block_keyspace_notification.tcl
yairgott: done

tests/modules/block_keyspace_notification.c
yairgott: done

src/module.c
yairgott: done

tests/unit/moduleapi/block_keyspace_notification.tcl
yairgott: ack

madolson:
@valkey-io/core-team This API is a bit late to be added. The feature adds a new functionality to using keyspace notifications, allowing them to block clients. The goal is to provide read and after write consistency for the VSS module.

There are two different ways to provide this guarantee, either block the write until it's visible in the index or block the read until all pending writes were ingested. The search module preferred to use the blocking of the write approach, but need an API to block the writes.

@murphyjacob4 Will follow up to see if there is a away to avoid this new functionality.

src/notify.c
```diff
@@ -112,7 +113,7 @@ void notifyKeyspaceEvent(int type, char *event, robj *key, int dbid) {
      * This bypasses the notifications configuration, but the module engine
      * will only call event subscribers if the event type matches the types
      * they are interested in. */
-    moduleNotifyKeyspaceEvent(type, event, key, dbid);
+    moduleNotifyKeyspaceEvent(c, type, event, key, dbid);
```
QuChen88: Can you simply use `server.current_client` when you invoke this method? That way you don't need to pass in `client*` into `notifyKeyspaceEvent()`.

src/module.c
```diff
@@ -8880,8 +8924,14 @@ void moduleNotifyKeyspaceEvent(int type, const char *event, robj *key, int dbid)
         if ((sub->event_mask & type) &&
             (sub->active == 0 || (sub->module->options & VALKEYMODULE_OPTIONS_ALLOW_NESTED_KEYSPACE_NOTIFICATIONS))) {
             ValkeyModuleCtx ctx;
-            moduleCreateContext(&ctx, sub->module, VALKEYMODULE_CTX_TEMP_CLIENT);
-            selectDb(ctx.client, dbid);
+            if (c == NULL) {
```
QuChen88: Rather than passing in `client*` to this module API, can you simply check if `server.current_client` is NULL or not? 

Or... maybe check `server.executing_client` if you want to cover the case of a Lua script.

src/module.c
```diff
@@ -8091,14 +8112,27 @@ int moduleTryServeClientBlockedOnKey(client *c, robj *key) {
...
  *
- * Note: ValkeyModule_UnblockClient should be called for every blocked client,
- *       even if client was killed, timed-out or disconnected. Failing to do so
- *       will result in memory leaks.
+ * Notes:
+ * 1. ValkeyModule_UnblockClient should be called for every blocked client,
+ * even if client was killed, timed-out or disconnected. Failing to do so
+ * will result in memory leaks.
+ * 2. Blocking the client on keyspace event notification is supported from
+ * version 8.1
```
zuiderkwast: If we want to consider this a bugfix and release it in a patch version, then we should formulate it differently to comply with [semver](https://semver.org/). Something like this?

```suggestion
 * 2. Attempting to block the client on keyspace event notification in versions
 *    prior to 8.1.1 leads to a crash.
```

yairgott: Done

src/notify.c
yairgott: Done

src/module.c
```diff
@@ -8091,14 +8116,27 @@ int moduleTryServeClientBlockedOnKey(client *c, robj *key) {
...
+ * 3. If the client is a temporary module client.
+ * 4. If the client is already blocked.
+ *
+ * In cases 1 and 2, a call to ValkeyModule_BlockClient() will **not** block the
+ * client, but instead produce a specific error reply. Note that if the
+ * BlockClient call originated from within a keyspace notification, no error
+ * reply is generated and nullptr is returned.
+ *
+ * In case 3 and 4, a call to ValkeyModule_BlockClient() are no-op, returning
+ * nullptr and setting the errno to ENOTSUP.
```
zuiderkwast: In cases 1 and 2, what is errno set to?

If we set errno in some cases when returning NULL, then I think we should always set it when we return NULL.

If we need another errno value to distinguish the cases, maybe use `EWOULDBLOCK`?

yairgott: hmm, based on the linux [errno](https://man7.org/linux/man-pages/man3/errno.3.html) man, EWOULDBLOCK is the same as EAGAIN with a description: "Resource temporarily unavailable"

What do you think about EINVAL?

QuChen88:
Note: this PR is mainly fixing an engine crash and read after write consistency on the same connection related to the VSS module, so it is technically a bug fix in my mind, not new feature. 

Also with the latest revision, the module API change is kept to minimal.

src/module.c
zuiderkwast: Yeah WOULDBLOCK is probably wrong. I just thought it was cool that it mentions blocking but it's another kind of blocking. Blocking read/write because there is not enough data to read yet.

EINVAL is good for invalid parameters. We used that in `VM_GetCurrentUserName` when the context doesn't contain a real client. That includes the lua fake client I believe.

In the latest change, I see you wrote ENOTSUP for all four cases here. I'm fine with that too, if there is no need for modules to distinguish between the cases.

src/module.c
```diff
@@ -8300,6 +8338,10 @@ int moduleClientIsBlockedOnKeys(client *c) {
  * ValkeyModule_BlockClientOnKeys() is accessible from the timeout
  * callback via VM_GetBlockedClientPrivateData). */
 int VM_UnblockClient(ValkeyModuleBlockedClient *bc, void *privdata) {
+    if (!bc) {
+        /* No-op if the blocked client is null. */
+        return VALKEYMODULE_OK;
+    }
```
zuiderkwast: Is this where you want to set errno to EINVAL?

src/notify.c
```diff
@@ -96,7 +96,7 @@ sds keyspaceEventsFlagsToString(int flags) {
 
 /* The API provided to the rest of the serer core is a simple function:
  *
- * notifyKeyspaceEvent(int type, char *event, robj *key, int dbid);
+ * notifyKeyspaceEvent(client *c, int type, char *event, robj *key, int dbid);
```
zuiderkwast: We ended up not adding this argument to the function, so this line needs to be reverted, right?

zuiderkwast (commented):
I didn't review the test case yet. It's failing though in the latest CI run:

```
*** [err]: Event that fires twice in tests/unit/moduleapi/block_keyspace_notification.tcl
Rename event not propagated within 5 seconds
```

Please look into it. We've had many problems with flaky test.

src/module.c
zuiderkwast: Apparently this change is not pushed yet.

src/notify.c
yairgott: Done

src/module.c
yairgott: Done

src/module.c
yairgott: Changed it to set the errno to `EINVAL` for cases 1-3 and `ENOTSUP ` for #4.

src/module.c
yairgott: oops, fixed now.

QuChen88:
With the simplified module API code, I will take another look this week on the functional change in `module.c`. Will give my thumb-up if all looks good.

src/module.c
```diff
@@ -8300,6 +8338,10 @@ int moduleClientIsBlockedOnKeys(client *c) {
  * ValkeyModule_BlockClientOnKeys() is accessible from the timeout
  * callback via VM_GetBlockedClientPrivateData). */
 int VM_UnblockClient(ValkeyModuleBlockedClient *bc, void *privdata) {
+    if (!bc) {
+        errno = EINVAL;
+        return VALKEYMODULE_ERR;
```
zuiderkwast, line 8350: The return value (OK and ERR) are not documented and `errno` isn't documented.

There's another case in this function below these lines where ERR is returned and `errno` is not set in this case:

```C
    if (bc->blocked_on_keys) {
        /* In theory the user should always pass the timeout handler as an
         * argument, but better to be safe than sorry. */
        if (bc->timeout_callback == NULL) return VALKEYMODULE_ERR;
```

If a function sets `errno`, then it should set in all error cases. A missing `bc->timeout_callback` is an invalid argument too?

src/t_hash.c
```diff
@@ -869,10 +870,10 @@ void hincrbyCommand(client *c) {
     value += incr;
     new = sdsfromlonglong(value);
     hashTypeSet(o, c->argv[2]->ptr, new, HASH_SET_TAKE_VALUE);
-    addReplyLongLong(c, value);
     signalModifiedKey(c, c->db, c->argv[1]);
     notifyKeyspaceEvent(NOTIFY_HASH, "hincrby", c->argv[1], c->db->id);
     server.dirty++;
+    addReplyLongLong(c, value);
```
zuiderkwast, line 876: What's the behavior change here?

Is it possible in any way that this change it can break anything user code?

yairgott: This reordering allows the blocking of client to be called before adding the reply, ensuring that prior to client blocking reply buffers are allowed to be sent while later replies are blocked.  I don't see how it may break anything. Noting that today sometimes add reply is called before and sometimes after the call to `signalModifiedKey `/ `notifyKeyspaceEvent` , this PR makes the call to add reply consistently after `signalModifiedKey `/ `notifyKeyspaceEvent`.

zuiderkwast: Right, thanks, makes sense!

src/module.c
yairgott: This is very strange behavior for the following reasons:
1. There is no enforcement during blocking the client for a not null timeout callback when blocked keys are not NULL.  
2. it's not expected for this function is called when the client is blocked with blocked keys: https://github.com/valkey-io/valkey/blob/unstable/src/module.c#L8206.

Given this, I think that we should just maintain this weird behavior to avoid breaking compatibility.  I'll just add errno assignment of ENOTSUP and document it.

src/module.c
```diff
@@ -8328,6 +8328,12 @@ int moduleClientIsBlockedOnKeys(client *c) {
  * needs to be passed to the client, included but not limited some slow
  * to compute reply or some reply obtained via networking.
  *
+ * Returns VALKEYMODULE_OK on success. On failure, VALKEYMODULE_ERR is returned
+ * and `errno` is set as follows:
+ *
+ * - EINVAL if bc is NULL.
+ * - ENOTSUP if bc contains `blocked on keys` but it timeout callback is NULL.
```
zuiderkwast:
```suggestion
 * - ENOTSUP if bc contains `blocked on keys` but its timeout callback is NULL.
```

src/networking.c
```diff
@@ -1283,6 +1297,26 @@ void addReplySubcommandSyntaxError(client *c) {
     sdsfree(cmd);
 }
 
+/* Replies which were added after the client is blocked by a module are accumulated
+ * separately. We need to transmit those replies to the client. */
+void AddReplyModuleUnBlockedClient(client *c) {
```
zuiderkwast: Nits:
1. The convention is to start with a lowercase letter. (AddReplyFromClient is the odd one.)
2. "unblocked" is one word, so it should have a lowercase "b".

```suggestion
void addReplyModuleUnblockedClient(client *c) {
```

src/module.c
```diff
@@ -7807,8 +7807,28 @@ ValkeyModuleBlockedClient *moduleBlockClient(ValkeyModuleCtx *ctx,
...
+    int is_keyspace_notification = ctx->flags & (VALKEYMODULE_CTX_KEYSPACE_NOTIFICATION);
     int islua = scriptIsRunning();
     int ismulti = server.in_exec;
+    if ((islua || ismulti) && is_keyspace_notification) {
+        /* Avoid blocking within transactions when context initiated by
+         * keyspace notification. */
+        errno = EINVAL;
+        return NULL;
+    }
+    errno = 0;
```
zuiderkwast: Don't set it to zero. From the errno(3) man page:

*"The value of errno is never set to zero by any system call or library function."*

The caller checks errno only if the return value indicates error.

```suggestion
```

src/module.c
```diff
@@ -8300,14 +8344,22 @@ int moduleClientIsBlockedOnKeys(client *c) {
...
-        if (bc->timeout_callback == NULL) return VALKEYMODULE_ERR;
+        if (bc->timeout_callback == NULL) {
+            errno = ENOTSUP;
+            return VALKEYMODULE_ERR;
+        }
         if (bc->unblocked) return VALKEYMODULE_OK;
         if (bc->client) moduleBlockedClientTimedOut(bc->client, 1);
     }
     moduleUnblockClientByHandle(bc, privdata);
+    errno = 0;
```
zuiderkwast:
```suggestion
```

src/server.h
```diff
@@ -1220,11 +1220,13 @@ typedef struct client {
         uint64_t raw_flag;
         struct ClientFlags flag;
     };
-    uint16_t write_flags;            /* Client Write flags - used to communicate the client write state. */
-    volatile uint8_t io_read_state;  /* Indicate the IO read state of the client */
-    volatile uint8_t io_write_state; /* Indicate the IO write state of the client */
-    uint8_t resp;                    /* RESP protocol version. Can be 2 or 3. */
-    uint8_t cur_tid;                 /* ID of IO thread currently performing IO for this client */
+    list *blocked_client_pending_reply;                    /* List of reply objects to send to the client after the client is unblocked */
+    unsigned long long blocked_client_pending_reply_bytes; /* Total bytes of objects in the blocked client pending list.*/
```
zuiderkwast: These blocking-related fields should be added in the `blockingState` struct instead. The client's blocking state `c->bstate` is created lazily to reduce client memory usage.

This refactoring was done in #1405 and it saved a lot of memory per client.

tests/unit/moduleapi/block_keyspace_notification.tcl
```diff
@@ -0,0 +1,125 @@
+
+set testmodule [file normalize tests/modules/block_keyspace_notification.so]
+
+start_server {tags {"modules"}} {
+    set i [r info]
+    regexp {valkey_version:(.*?)\r\n} $i - version
+    if { $version eq "7.2.7" } {
+        puts "Skipping test: Valkey version is 7.2.7"
```
zuiderkwast: Exact match seems weird. Would the test work on 7.2.6?

Shouldn't it be a less-or-equal check instead?

yairgott: 7.2.6 has an issue loading modules, see: https://github.com/valkey-io/valkey/issues/1054

Valkey is also using 7.2.7 for testing modules: https://github.com/valkey-io/valkey/blob/add716b7ddce48d4e13ebffe65401c7d0e26b91a/.github/workflows/ci.yml#L25

zuiderkwast: > 7.2.6 has an issue loading modules

Ah, I didn't know. So we can use `>= 7.2.7` then?

> Valkey is also using 7.2.7 for testing modules

Yeah we use it compatibility testing, but we might change it to 7.2.8 or 8.0.2 or something else later. It's good to avoid the hard-coded value in multiple places.

zuiderkwast: Can we just skip it on all `7.*` versions?

```suggestion
    regexp {valkey_version:(.*?)\r\n} $i - version
    if {[string match "7.*" $version]} {
        puts "Skipping test: Valkey version is $version"
```

src/server.h
QuChen88: I am actually thinking of making this a bit more general by blocking certain responses on conditions other than `c->bstate` (i.e. blocking commands etc). For example, when you pipeline multiple commands over the client, we allow execution of subsequent commands but hold off on sending responses back to the client until some conditions becomes true later on. 

I would say coupling this with `c->bstate` won't allow us to extend the response blocking functionality later on to other use-cases. So prefer to put it outside of the `bstate` struct

tests/unit/moduleapi/block_keyspace_notification.tcl
yairgott: We should actually change it to either unstable or greater than 8.1.1. But the condition on the unstable can be added just after this PR lands. I suggest that we don't change it now and do a quick followup PR post landing. wdyt?

zuiderkwast: Unstable is reported as 255.255.255, so >= 8.1.1 would work then.

> But the condition on the unstable can be added just after this PR lands.

I don't undestand this. Why would we need to do this as a follow-up? This PR will not be on unstable until this PR is merged...

zuiderkwast: Shall we add a function to compare version numbers in tests/support/util.tcl? Something like this:

```tcl
proc version_greater_or_eq {version_a version_b} {
    regexp {^(\d+)\.(\d+)\.(\d+)$} $version_a -> a1 a2 a3
    regexp {^(\d+)\.(\d+)\.(\d+)$} $version_b -> b1 b2 b3
    return [expr {$a1 > $b1 ||
                  ($a1 == $b1 && $a2 > $b2) ||
                  ($a1 == $b1 && $a2 == $b1 && $a3 >= $b3)}]
}
```

src/module.c
yairgott: SG! Note that there are other places where errno is set to 0 when returning OK in module.c. See for instance: https://github.com/valkey-io/valkey/blob/unstable/src/module.c#L9518
I agree that it's better to follow the errno man page in general but to avoid breaking the old behavior, I'm not sure if we should actually fix the existing usage. wdyt?

src/networking.c
yairgott: Done

src/module.c
yairgott: Done

tests/unit/moduleapi/block_keyspace_notification.tcl
yairgott: I added the function `valkey_version_at_least` to `tests/support/util.tcl`. I chose a slightly different implementation.

src/networking.c
yairgott: Done

src/networking.c
```diff
@@ -1283,6 +1297,26 @@ void addReplySubcommandSyntaxError(client *c) {
...
+    if (c->blocked_client_pending_reply == NULL || listLength(c->blocked_client_pending_reply) == 0 || prepareClientToWrite(c) != C_OK) {
+        resetBlockedClientPendingReply(c);
+        return;
+    }
+    /* We're bypassing _addReplyProtoToList, so we need to add the pre/post
+     * checks in it. */
+    if (c->flag.close_after_reply) return;
+
+    listJoin(c->reply, c->blocked_client_pending_reply);
```
QuChen88: Not sure if this implementation is correct. 

The client output buffer consists of an initial 16KB buffer `c->buf`. Once this buffer is full, only then it will start creating the list of reply blocks `c->reply`. Technically we should not be populating `c->reply` if the `c->buf` is not filled up.

src/module.c
zuiderkwast: Yeah, let's not touch the existing usage. That's always risky.

src/server.h
zuiderkwast: What do you mean? `blocked_client_pending_reply` only exists when the client is blocked, and then `c->bstate` exists too. Am I missing anything?

zuiderkwast: "hold off on sending responses back to the client" = block client

This is what blocking the client means. If we want to use data tiering and load data from disk in the background, we would block the client too, I suppose.

session.pem
```diff
@@ -0,0 +1,55 @@
+-----BEGIN SSL SESSION PARAMETERS-----
```
zuiderkwast: Did you just add a lot of files that are not supposed to be added?

yairgott: Sorry, removing those.

madolson:
Just posting that the newly added test seems to be failing: https://github.com/valkey-io/valkey/actions/runs/14316235962/job/40122905972?pr=1819#step:6:1116.

src/module.c
```diff
@@ -7807,8 +7807,27 @@ ValkeyModuleBlockedClient *moduleBlockClient(ValkeyModuleCtx *ctx,
                                              void *privdata,
                                              int flags) {
     client *c = ctx->client;
+    if (c->flag.blocked || getClientType(c) != CLIENT_TYPE_NORMAL || c->flag.deny_blocking) {
+        /* Early return if duplicate block attempt or client is not normal or
+         * client is set to deny blocking. */
+        errno = ENOTSUP;
```
madolson, line 7815: This code is fine, but we should add the various error types in the comment block above so folks know how to check for them.

src/networking.c
yairgott: That is not accurate.
1. `c->reply` is added when either `c->buf` is full or `c->reply` is not empty, see: https://github.com/valkey-io/valkey/blob/unstable/src/networking.c#L392
2. The check wether is client has pending writes is based on either the reply list or the buffer, see: https://github.com/valkey-io/valkey/blob/unstable/src/networking.c#L1369

madolson: There are cases today we start filling up the reply blocks even when the 16KB buffer isn't filled, like with deferred replies. The deferred reply implementation is much less efficient, since we are doing extra memory allocations and losing locality, but it should be technically OK. EDIT: I made my comment around the same time as Yair, I think what Yair said is correct.

madolson (commented):
Functionally looks correct to me now, just a minor comment about docs and I need to clean up the excess files. Thanks for working with us to make this change correct.

src/server.h
yairgott: Done, moved pending_reply[bytes] to `blockedClient` struct.

yairgott:
> Just posting that the newly added test seems to be failing: https://github.com/valkey-io/valkey/actions/runs/14316235962/job/40122905972?pr=1819#step:6:1116.

I changed the test and I believe that now it should be more robust.

src/module.c
yairgott: Adding a comment block would duplicate the comment block in `VM_BlockClient`. I'll add a comment to refer to `VM_BlockClient` instead. wdyt?

madolson: Yeah, that's fine too.

madolson:
> should

https://github.com/valkey-io/valkey/actions/runs/14320388536/job/40135953887?pr=1819#step:6:1120

squash one bug, and another one pops up :)

src/networking.c
QuChen88: OK I see. This is fine functionally speaking but less efficient to have the initial 16KB buffer not filled. I can try to follow up with this after this PR goes in.

QuChen88:
PR looks good to me. Thanks!

tests/support/util.tcl
```diff
@@ -1225,3 +1225,21 @@ proc bp {{s {}}} {
...
+        set a [expr {[llength $actual] > $i ? [lindex $actual $i] : 0}]
+        set b [expr {[llength $required] > $i ? [lindex $required $i] : 0}]
+        if {$a > $b} {
+            return 1
+        } elseif {$a < $b} {
+            return 0
+        }
+    }
+    return 1 ;# exact match
+}
```
zuiderkwast: Why so complex? You didn't like my example code I posted before?

Instead of regexp + string split + loop, we can use only regexp and a condition.

I can adjust my suggestion to call the `INFO` command:

```suggestion
proc valkey_version_at_least {min_version} {
    set i [r info server]
    regexp {valkey_version:(\d+)\.(\d+)\.(\d+)} $i - major minor patch
    regexp {^(\d+)\.(\d+)\.(\d+)} $min_version - min_major min_minor min_patch
    return [expr {$major > $min_major ||
                  ($major == $min_major && $minor > $min_minor) ||
                  ($majro ==  $min_major && $minor == $min_minor && $patch >= $min_patch)}]
}
```

madolson (approved):
OK with the code as is, but also fine if @zuiderkwast wants to apply his recommendation.

tests/support/util.tcl
yairgott: Done

QuChen88 (approved):

tests/unit/moduleapi/block_keyspace_notification.tcl
zuiderkwast: Ack but no change?

Too much work? Ok.. If the test is found to be flaky later, we can do what Madelyn suggested.

tests/unit/moduleapi/block_keyspace_notification.tcl
```diff
@@ -0,0 +1,130 @@
+
+set testmodule [file normalize tests/modules/block_keyspace_notification.so]
+
+start_server {tags {"modules"}} {
+    if {[valkey_version_at_least "8.1.1"]} {
```
zuiderkwast: I don't fully understand why we need this version check. Aren't all tests designed to run against the current version of Valkey? The exception is cross-version tests but those explicitly start the older version and use it only for specific scenarios. We don't run arbitrary tests against the older version because they're not designed for that. Many of them would fail.

src/t_zset.c
```diff
@@ -1851,32 +1853,35 @@ static void zaddGenericCommand(client *c, int flags) {
...
         score = newscore;
     }
-    server.dirty += (added + updated);
+    if (!reply_err) {
+        server.dirty += (added + updated);
+    }
+    if (added || updated) {
+        signalModifiedKey(c, c->db, key);
+        notifyKeyspaceEvent(NOTIFY_ZSET, incr ? "zincr" : "zadd", key, c->db->id);
+    }
```
zuiderkwast, line 1870: Many commands are changed by reordering some lines in this way to make them work with blocking on keyspace notifications, but tests are not added for every changed command. Ideally we should cover all these changes. Otherwise, how do you know it will work?

How did you find out you need to change all these commands? Just by reading code?

tests/unit/moduleapi/block_keyspace_notification.tcl
QuChen88: I think the intent here is that we know valkey 8.1.1 will contain this bug fix... 

But I agree with you, this test should run against the current code and we can remove this check.

src/t_zset.c
madolson: Earlier on we considered adding a debug server assert to make sure we weren't adding data *before* the keyspace notification. That seemed like a more generic approach to cover this case, and we could also then run it for other modules.

tests/unit/moduleapi/block_keyspace_notification.tcl
yairgott: I've added this because we have compatibility testing which are are running against 7.2.7. Without this, the added tests are failing when the engine runs with 7.2.7. See https://github.com/valkey-io/valkey/blob/51387ddc99404dd4ecb426f43fbe4630659922ea/.github/workflows/ci.yml#L29

zuiderkwast: I think you mentioned this before, but I find it hard to understand.

Did you actually see the compatibility testing fail without this version check?

The "other server" (as defined by `--other-server-path`) is only supposed to be started when you call `start_server` with the option `start-other-server 1`. Otherwise, `start_server` starts the current version. The `start-other-server` option is used in only two places:

```
$ git grep -n 'start-other-server 1'
tests/integration/cross-version-replication.tcl:14:start_server {tags {"repl needs:other-server external:skip"} start-other-server 1 config "minimal.conf"} {
tests/unit/moduleapi/cross-version-cluster.tcl:16:    start_server {config "minimal-cluster.conf" start-other-server 1} {
```

Now to verify this, I actually checked out your branch, removed this check for `valkey_version_at_least` and ran the tests with `./runtest --verbose --tags -slow --dump-logs --other-server-path misc/valkey-server-7` (where `misc/valkey-server-7` is my local valkey 7 binary). All the tests passed. So I still think this check isn't needed... Am I missing anything?

src/t_zset.c
yairgott: > Ideally we should cover all these changes. Otherwise, how do you know it will work?

It's true that not every data-type mutation is covered but IMO we should strike a balance between the need to cover everything with testings and the value that it may produce. 

The assent of this part of the change is reordering the code to make sure that calls `notifyKeyspaceEvent` are done before `addReply*`.  In very high level, `addReply*` essentially calls `prepareClientToWrite` and later, before this PR, adds the buffer to be transmitted to the relevant buffer, either the `c->buf` or `c->reply`. By going over the code, it's clear that calling `prepareClientToWrite` consistently before `addReply*` wouldn't impact `addReply*`.

> How did you find out you need to change all these commands?

Searching for all `notifyKeyspaceEvent` calls.  In most of the cases, `notifyKeyspaceEvent` is followed by a `addReply*` call, in the same code block. There are some cases, like https://github.com/valkey-io/valkey/blob/51387ddc99404dd4ecb426f43fbe4630659922ea/src/evict.c#L678, in which `notifyKeyspaceEvent` is not relevant.

> Earlier on we considered adding a debug server assert to make sure we weren't adding data before the keyspace notification. That seemed like a more generic approach to cover this case, and we could also then run it for other modules.

I will add that but note that it requires an additional flag. The flag would be set on every public `addReply*` and later verified in `notifyKeyspaceEvent` that the flag is not set.

zuiderkwast: > It's true that not every data-type mutation is covered but IMO we should strike a balance between the need to cover everything with testings and the value that it may produce.

Sounds good. Yes, I agree it's too much to test all the commands.

The debug assert sounds like a good idea though. I think we can set the flag in `prepareClientToWrite()`, right? It's called by every `addReply*`.

yairgott: Yeah, that seems to be correct. Let me see if that works.

tests/unit/moduleapi/block_keyspace_notification.tcl
yairgott: I added this version check after the tests were failing due to the version. It might be that something has changed and now the module tests are running with the most recent version. Anyway, let me remove this check and submit the change to see if something breaks now.

src/t_zset.c
madolson: Yes, the original idea was to add it into prepareClientToWrite().

Alternatively, we could also add a proc that artificially calls every write command, which could be re-usuable between different tests.

zuiderkwast: The flag can have an extra benefit to make `prepareClientToWrite` faster with this short cut, for commands where `addReply*` is called multiple times for the same reply. Checking multiple flags together would be converted to a bitmask check by the compiler i.e. no extra instruction to check it.

```diff
 int prepareClientToWrite(client *c) {
     /* If it's the Lua client we always return ok without installing any
      * handler since there is no socket at all. */
-    if (c->flag.script || c->flag.module) return C_OK;
+    if (c->flag.script || c->flag.module || c->flag.prepared_to_write) return C_OK;

     /* If CLIENT_CLOSE_ASAP flag is set, we need not write anything. */
     if (c->flag.close_asap) return C_ERR;

     /* CLIENT REPLY OFF / SKIP handling: don't send replies.
      * CLIENT_PUSHING handling: disables the reply silencing flags. */
     if ((c->flag.reply_off || c->flag.reply_skip) && !c->flag.pushing) return C_ERR;

     /* Primaries don't receive replies, unless CLIENT_PRIMARY_FORCE_REPLY flag
      * is set. */
     if (c->flag.primary && !c->flag.primary_force_reply) return C_ERR;

     /* Skip the fake client, such as the fake client for AOF loading.
      * But CLIENT_ID_CACHED_RESPONSE is allowed since it is a fake client
      * but has a connection to cache the response. */
     if (c->flag.fake && c->id != CLIENT_ID_CACHED_RESPONSE) return C_ERR;
     serverAssert(c->conn);

     /* Schedule the client to write the output buffers to the socket, unless
      * it should already be setup to do so (it has already pending data). */
     if (!clientHasPendingReplies(c)) putClientInPendingWriteQueue(c);

     /* Authorize the caller to queue in the output buffer of this client. */
+    c->flag.prepared_to_write = 1;
     return C_OK;
 }
```

madolson: We still need to do some those other checks for correctness. For example the reply_off could have been triggered inside of a multi-exec where one of the commands is disabling client replying. We also don't want to be accumulating responses on a client with the close_asap flag that is set.

yairgott:
A few notes regarding the recent changes:
1. The assert uncovered a few edge cases, such as SET_GET and list pop command. In these cases,  `addReply` is called on an object before it's mutated, and the keyspace notification is triggered afterward.
2. To handle these cases correctly, I now use the deferred reply buffer to store replies regardless of whether the client is blocked. As a result, I moved the deferred reply buffer from the blockingState struct into the client structure.
3. The assert is currently enabled, but I believe it should be limited to debug mode, since our tests can’t ensure full coverage of all edge cases.
4. Noting that calling `addReply` before keyspace notification still maintains read-after-write consistency on the same connection, as the client is simply blocked on the subsequent command.

src/t_zset.c
yairgott: This might lead to a miss of a call to `putClientInPendingWriteQueue` . Also, this could be problematic with `primary_force_reply`, see: https://github.com/valkey-io/valkey/blob/89d45775c42241c537d366d6fb61d334386f887c/src/replication.c#L4260

src/blocked.c
```diff
@@ -98,7 +98,9 @@ void freeClientBlockingState(client *c) {
     dictRelease(c->bstate->keys);
     zfree(c->bstate);
     c->bstate = NULL;
-    if (c->deferred_reply) commitDeferredReplyBuffer(c);
+
+    serverAssert(!c->deferred_reply);
```
zuiderkwast: Use debugServerAssert makes it crash only during our tests and not in production. It is enabled by the undocumented config `enable-debug-assert` which is off by default.

```suggestion
    debugServerAssert(!c->deferred_reply);
```

src/networking.c
```diff
@@ -1299,6 +1342,7 @@ void AddReplyFromClient(client *dst, client *src) {
         return;
     }
 
+
```
```suggestion
```

src/t_list.c
```diff
@@ -744,7 +744,7 @@ void listElementsRemoved(client *c, robj *key, int where, robj *o, long count, i
         listTypeTryConversion(o, LIST_CONV_SHRINKING, NULL, NULL);
         if (deleted) *deleted = 0;
     }
-    if (signal) signalModifiedKey(c, c->db, key);
```
madolson, line 747: Why are we always signaling now?

src/t_list.c
```diff
@@ -791,9 +791,11 @@ void popGenericCommand(client *c, int where) {
         long rangeend = (where == LIST_HEAD) ? rangelen - 1 : -1;
         int reverse = (where == LIST_HEAD) ? 0 : 1;
 
+        initDeferredReplyBuffer(c);
```
madolson, line 794: This has performance implications, since we are now *always* adding the response to a deferred buffer and then moving it back.

src/blocked.c
```diff
@@ -98,6 +98,9 @@ void freeClientBlockingState(client *c) {
     dictRelease(c->bstate->keys);
     zfree(c->bstate);
     c->bstate = NULL;
+
+    serverAssert(!c->deferred_reply);
+    commitDeferredReplyBuffer(c, 0);
```
madolson: Why is there also a commit here? We already assert there is no deferred_reply.

madolson (commented):
The new changes impact code paths that are executed without the module, which give me more pause for releasing them in a patch. Maybe the module should register the fact that it needs deferred replies so steady state is not impacted?

src/blocked.c
yairgott: I switched to debug assert.

src/notify.c
```diff
@@ -107,12 +107,24 @@ void notifyKeyspaceEvent(int type, char *event, robj *key, int dbid) {
...
-
+    client *c = server.executing_client;
+    debugServerAssert((type & (NOTIFY_GENERIC | NOTIFY_STRING | NOTIFY_LIST | NOTIFY_SET | NOTIFY_HASH | NOTIFY_ZSET | NOTIFY_STREAM)) == 0 ||
+                 c == NULL ||
+                 c->cmd == NULL ||
+                 (c->cmd->flags & CMD_WRITE) == 0 ||
+                 c->flag.buffered_reply == 0 ||
+                 c->flag.keyspace_notified == 1 ||
+                 c->id == UINT64_MAX || // AOF client
+                 getClientType(c) != CLIENT_TYPE_NORMAL);
```
zuiderkwast: Clang-format complains about this.

```suggestion
    debugServerAssert((type & (NOTIFY_GENERIC | NOTIFY_STRING | NOTIFY_LIST | NOTIFY_SET | NOTIFY_HASH | NOTIFY_ZSET | NOTIFY_STREAM)) == 0 ||
                      c == NULL ||
                      c->cmd == NULL ||
                      (c->cmd->flags & CMD_WRITE) == 0 ||
                      c->flag.buffered_reply == 0 ||
                      c->flag.keyspace_notified == 1 ||
                      c->id == UINT64_MAX || // AOF client
                      getClientType(c) != CLIENT_TYPE_NORMAL);
```

src/t_list.c
yairgott: Right, but the performance implication is pretty negligible as there is no deep copy. Is this command super common that it's expected to actually make a difference?

src/blocked.c
zuiderkwast: So just in case the debug is disabled and there is a deferred reply, we commit it?

Can it happen in commands implemented in modules?

src/t_list.c
yairgott: I removed the `signal` parameter as it was always set to 1.

src/t_list.c
yairgott: The alternative is to switch to book-keeping the client reply/buf offsets but we'll loose the simplicity and robustness of this approach.

src/notify.c
yairgott: Done

src/networking.c
yairgott: Done

src/t_list.c
zuiderkwast: I guess it will not have a huge impact, but maybe it can be noticed in some use cases. Most users never use SPOP at all, but some clients call SPOP all the time, for some special use case... Should we run a quick benchmark on SPOP just to know?

I see that `commitDeferredReplyBuffer` calls `resetDeferredReplyBuffer` which frees the list, and then it's allocated again on the next `initDeferredReplyBuffer`. Maybe we should leave an empty list and not free it? Then the next SPOP in the same connection will be faster.

madolson: We have seen performance issues in the past by not directly adding it to the fixed buffer (for example, using a deferred reply), so I don't want to assume it's negligible. I think the SPOP command is not super common, but some of the others like GETSET or GETEX are more widely used commands.

I suppose this is why we need the automated performance benchmarks :/ (one day)

yairgott: I checked the performance using the below script on my local VM:
Avg with the deferred reply: 12.64 s
Avg without the deferred reply: 12.61 s

I applied the change to avoid releasing the list on a call to `commitDeferredReplyBuffer`. FWIW, the perf numbers are without this optimization. The overhead now is basically:
1. allocating/releasing a single sds.
2. move this sds from one list to the other.

Script:
```
import redis
import time

r = redis.Redis(host='localhost', port=6379, decode_responses=True)

# Step 1: Populate
pipe = r.pipeline()
for i in range(1_000_000):
    key = f"list_aaaaaaa:{i}"
    pipe.rpush(key, *range(10))
    if i % 1_000 == 0:
        pipe.execute()
    if i % 10_000 == 0:
        print(f"Populated {i} lists")
pipe.execute()

# Step 2: Time the popping phase
start = time.time()

for i in range(1_000_000):
    key = f"list_aaaaaaa:{i}"
    pipe.lpop(key, 5)
    if i % 1_000 == 0:
        pipe.execute()
    if i % 10_000 == 0:
        print(f"Popped from {i} lists")

pipe.execute()
end = time.time()
print(f"Total time to pop 5 elements from 1M lists: {end - start:.2f} seconds")
```

yairgott: If you're still concerned about the overhead, I suggest to avoid deferring the reply in those use-cases, there are 7 such cases. FWIW, those cases don't impact ValkeySearch as ValkeySearch indexes only the hash data-type and JSON, which its type is module. Additionally, read-after-write consistency is still preserved as the processing of the consecutive command is yield until the client is unblocked.

I understand that Qu already started looking at the book-keeping alternative and we can revisit deferring the reply for these cases once the book-keeping PR is available.

wdyt?

madolson: What was the server side CPU at the time of executing that? Adhoc python tests normally aren't saturating the server so the bottleneck might just have been on the python script. If it wasn't we probably need to run it again with benchmark.

zuiderkwast: Why not use valkey-benchmark with pipelining and threads?

Populate:

    valkey-benchmark -P 10 --threads 5 -n 1000000 RPUSH mylist 1 2 3 4 5 6 7 8 9 10

Benchmark:

    valkey-benchmark -P 10 --threads 5 -n 10000000 LPOP mylist

yairgott: hmm, would this populate  a single key many times?

zuiderkwast: It pushes to the same key many times yes. It would have 1M x 10 = 10M elements. Then LPOP 10M times.

zuiderkwast: If you want to test with many separate keys, i guess you can use the new `--sequential` option.

    valkey-benchmark -r 1000000 -n 1000000 --sequential rpush list:__rand_int__ 1 2 3 4 5 6 7 8 9 10

yairgott: I ran the script in parallel 10 times, reaching close to 100% CPU utilization on the Valkey instance. The performance results were:
* Avg with deferred reply: 14.79s
* Avg without deferred reply: 13.94s

To mitigate the performance impact, I applied the following changes:
1. The deferred reply buffer now has a minimum allocation size of 1KB.
2. `initDeferredReplyBuffer` is now a no-op if there are no keyspace notification subscribers. This ensures no performance regression when no module subscribed to the keyspace notifications.  To clarify, the above perf numbers show the regression only in case a module subscribed to keyspace notification and applicable just for a few commands.

zuiderkwast: > 2\. `initDeferredReplyBuffer` is now a no-op if there are no keyspace notification subscribers

Smart idea!

src/networking.c
```diff
@@ -1283,6 +1292,43 @@ void addReplySubcommandSyntaxError(client *c) {
...
+void initDeferredReplyBuffer(client *c) {
+    if (moduleNotifyKeyspaceSubscribersCnt() == 0) return;
+    if (c->deferred_reply == NULL) c->deferred_reply = listCreate();
+    if (!isDeferredReplyEnabled(c)) c->deferred_reply_bytes = 0;
+}
+
+static void resetDeferredReplyBuffer(client *c) {
+    listEmpty(c->deferred_reply);
+    c->deferred_reply_bytes = ULLONG_MAX;
+}
```
zuiderkwast: Making this a no-op if there are no modules subscribing to keyspace notifications is a smart idea. Almost too smart, so contributors will not understand why we do this. :smile:

Let's add some comment to document these functions. Especially `initDeferredReplyBuffer`, why some commands should use it and why module keyspace subscribers have anything to do with it.

Something like this? Correct it if it's wrong...

```C
/* Commands that produce replies before signaling keyspace notifications need
 * to use a deferred reply buffer to ensure read-after-write consistency.
 * This is because modules can be subscribing to keyspace notifications and
 * block the client from the keyspace notification callback. */
void initDeferredReplyBuffer(client *c) {
```

yairgott: I changed slightly the comment, mainly to clarify the motivation. FWIW, read-after-write is achieved without deferring the reply as a consecutive request would not be processed until the client is unblocked. The value is more around the ability to attribute the added latency to the right command.

src/blocked.c
madolson: I still have an open question here about why we have this defensive commit. If it's just defensive can we comment that we are being defensive? The one in resetClient also seems like the right place to be defensive instead of here.

src/server.c
```diff
@@ -3743,6 +3745,7 @@ void call(client *c, int flags) {
 
     exitExecutionUnit();
 
+    commitDeferredReplyBuffer(c, 1);
```
madolson: I assume this is another defensive call? I don't think there is any case we should be exiting the execution unit without having added the reply to output buffer. I'm fine leaving it but would like to have debugAssert() and a comment so it's clear why this is here.

yairgott: Removing this instance.

src/blocked.c
yairgott: hmm, isn't it self documented by following a `debugServerAssert` call?

yairgott: Reviewing the code again, a client could terminate the connection while a client is blocked which would lead to a call `freeClient` followed by a call to `freeClientBlockingState`. I removed `client::deferred_reply` related handling from `freeClientBlockingState`.

madolson (approved):
My approval is still technically lingering, but I'm okay with the new set of changes. I'm not happy with the size of changes for a "patch", but we'll do better next time.

zuiderkwast (approved):

zuiderkwast:
This test failed in Daily in the reply-schemas-validator job:

* https://github.com/valkey-io/valkey/actions/runs/14554113682/job/40828749380#step:7:627
* https://github.com/valkey-io/valkey/actions/runs/14543642514/job/40805916104#step:7:626

It's an assert `logreqres.c:263 'ret' is not true` and the stack trace is a module unblocking a client.

It happens in the test case **"Blocking keyspace notification one"**.

yairgott:
Could be. I suspect that the [following](https://github.com/valkey-io/valkey/blob/unstable/src/networking.c#L507) should be move to line 501. I'll give it a try.

yairgott:
> This test failed in Daily in the reply-schemas-validator job:
> 
> * https://github.com/valkey-io/valkey/actions/runs/14554113682/job/40828749380#step:7:627
> * https://github.com/valkey-io/valkey/actions/runs/14543642514/job/40805916104#step:7:626
> 
> It's an assert `logreqres.c:263 'ret' is not true` and the stack trace is a module unblocking a client.
> 
> It happens in the test case **"Blocking keyspace notification one"**.

Could be. I suspect that the [following](https://github.com/valkey-io/valkey/blob/unstable/src/networking.c#L507) should be moved to line 501. I'll give it a try.

yairgott:
> > This test failed in Daily in the reply-schemas-validator job:
> > 
> > * https://github.com/valkey-io/valkey/actions/runs/14554113682/job/40828749380#step:7:627
> > * https://github.com/valkey-io/valkey/actions/runs/14543642514/job/40805916104#step:7:626
> > 
> > It's an assert `logreqres.c:263 'ret' is not true` and the stack trace is a module unblocking a client.
> > It happens in the test case **"Blocking keyspace notification one"**.
> 
> Could be. I suspect that the [following](https://github.com/valkey-io/valkey/blob/unstable/src/networking.c#L507) should be moved to line 501. I'll give it a try.

Yeah, it fixed it. See https://github.com/valkey-io/valkey/pull/1989

src/module.c
```diff
@@ -7807,8 +7809,27 @@ ValkeyModuleBlockedClient *moduleBlockClient(ValkeyModuleCtx *ctx,
...
+    }
+    int is_keyspace_notification = ctx->flags & (VALKEYMODULE_CTX_KEYSPACE_NOTIFICATION);
     int islua = scriptIsRunning();
     int ismulti = server.in_exec;
+    if ((islua || ismulti) && is_keyspace_notification) {
+        /* Avoid blocking within transactions when context initiated by
+         * keyspace notification. */
+        errno = EINVAL;
+        return NULL;
+    }
```
enjoy-binbin, line 7832: Wouldn't this be a dead code, since we checked flag.deny_blocking earlier? The line 7871 also the become the same dead code?
```
    if (islua || ismulti) {
        c->bstate->module_blocked_handle = NULL;
        addReplyError(c, islua ? "Blocking module command called from Lua script"
                               : "Blocking module command called from transaction");
```

i can see in tests/modules/auth.c, line 7871 is indeed became a dead code, and it is somehow a breaking change... The reply used to from the core and now is from the module. I will cleanup that dead code.

src/module.c
```diff
@@ -8091,14 +8117,27 @@ int moduleTryServeClientBlockedOnKey(client *c, robj *key) {
...
  *
  * There are some cases where ValkeyModule_BlockClient() cannot be used:
  *
  * 1. If the client is a Lua script.
  * 2. If the client is executing a MULTI block.
+ * 3. If the client is a temporary module client.
+ * 4. If the client is already blocked.
+ *
+ * In cases 1 and 2, a call to ValkeyModule_BlockClient() will **not** block the
+ * client, but instead produce a specific error reply. Note that if the
```
enjoy-binbin, line 8135: We don't produce the error reply.
```
    client *c = ctx->client;
    if (c->flag.blocked || getClientType(c) != CLIENT_TYPE_NORMAL || c->flag.deny_blocking) {
        /* Early return if duplicate block attempt or client is not normal or
         * client is set to deny blocking. */
        errno = ENOTSUP;
        return NULL;
    }
```

src/module.c
enjoy-binbin: please also check #2215 to see i done it right.

Issue #1807, opened 2025-03-02
[NEW] Let user change his own password


gkorland:
**The problem/use-case that the feature addresses**

Currently the only way to change your own password is by using ACL command
And users without Admin right should not have access to this command

**Description of the feature**

Add command or some API to allow the user change his password

```
SETPASS <password>
```

- `>password`: Adds the specified clear text password as a hashed password in the list of the users passwords. Every user can have many active passwords, so that password rotation will be simpler. The specified password is not stored as clear text inside the server. Example: >mypassword.

- `#<hashedpassword>`: Adds the specified hashed password to the list of user passwords. A Redis hashed password is hashed with SHA256 and translated into a hexadecimal string. Example: #c3ab8ff13720e8ad9047dd39466b3c8974e592c2fa383d4a3960714caef0c4f2.

- `<password`: Like >password but removes the password instead of adding it.

- `!<hashedpassword>`: Like #<hashedpassword> but removes the password instead of adding it.

madolson:
I like this idea, I would be OK with implementing this.

The only alternative is that we talked about implementing flags on users, one of those could be that you can only update your own password.

hpatro:
Currently all the permissions (command access/keyspace access/channel access/modifying password) are all tied up under `ACL SETUSER`. The question I've is do users update their own passwords, Isn't it generally performed by administrators? Under which scenario do you see users updating their passwords on their own?

If we decide to build this. We could maybe make this a sub-command of ACL itself. Users can be granted specific permission for that via `+ACL|SETPASS`.

```
ACL SETPASS <user> <symbol><password>
```

PR #1821, opened 2025-03-05
Enhance ACL WHOAMI command to display active ACL rules for the current connection 

hwware:
Current ACL WHOAMI can only display the username for the current connection.
It need to call ACL LIST to get the active ACL rules, thus there are 2 disadvantages: 

1. CALL 2 commands to get ACL rules for the current connection
2. ACL LIST will display all users and their ACL rules, it is not necessary,

This PR allows the current connection to call once "ACL WHOAMI FULL" to get all information.

### DIFF
--- a/src/acl.c
+++ b/src/acl.c
@@ -2910,9 +2910,18 @@ void aclCommand(client *c) {
             }
         }
         raxStop(&ri);
-    } else if (!strcasecmp(sub, "whoami") && c->argc == 2) {
+    } else if (!strcasecmp(sub, "whoami") && (c->argc == 2 || c->argc == 3)) {
         if (c->user != NULL) {
-            addReplyBulkCBuffer(c, c->user->name, sdslen(c->user->name));
+            if (c->argc == 3) {
+                sds config = sdsnew(c->user->name);
+                config = sdscatlen(config, " ", 1);
+                robj *descr = ACLDescribeUser(c->user);
+                config = sdscatsds(config, descr->ptr);
+                decrRefCount(descr);
+                addReplyBulkSds(c, config);
+            } else {
+                addReplyBulkCBuffer(c, c->user->name, sdslen(c->user->name));
+            }
         } else {
             addReplyNull(c);
         }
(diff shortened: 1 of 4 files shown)


hpatro (commented):
What's the use case for this ?

src/acl.c
```diff
@@ -2910,9 +2910,18 @@ void aclCommand(client *c) {
             }
         }
         raxStop(&ri);
-    } else if (!strcasecmp(sub, "whoami") && c->argc == 2) {
+    } else if (!strcasecmp(sub, "whoami") && (c->argc == 2 || c->argc == 3)) {
         if (c->user != NULL) {
-            addReplyBulkCBuffer(c, c->user->name, sdslen(c->user->name));
+            if (c->argc == 3) {
```
madolson, line 2915: I generally don't like variadic arguments that change the type of response. So would also somewhat prefer another command.

Also, do we want this to be a variant of `ACL GETUSER` instead? Somewhat related to this issue, https://github.com/valkey-io/valkey/issues/1807.

hwware:
> What's the use case for this ?

As the top comment describe, I thought

src/acl.c
hwware: Another command? you mean a new command?  I am afraid the user need to remember one more. 
I guess this command is not same as ACL GETUSER, because this command only apply to current connection, even user do not know the username either.

madolson: > I guess this command is not same as ACL GETUSER, because this command only apply to current connection, even user do not know the username either.

A new command, like `ACL GETMYSELF`, is what I'm suggesting. I agree it's one more command.

hpatro:
> > What's the use case for this ?
> 
> As the top comment describe, I thought

I mean it's not an expensive operation and don't see what's the pain behind doing it one after another.

zuiderkwast:
WHOAMI comes from the UNIX command `whoami`. In UNIX it displays the username and nothing else.

* Adding a parameter that modifies the type of the response is extra complexity and logic for client libraries to handle.
* The alternative, adding a new subcommand, is also extra complexity and more to remember.

I think we don't need this extra complexity. It is OK to call 2 commands to get the information. We don't need to optimize this use case for performace. For ease of use, it can be improved by external admin tools.

ranshid:
since this is an API change (anyway we look at it) it probably needs a major decision flag

madolson:
I'm going to be opinioated and close it. At the very least @lucasyonge would probably want to re-open it since it's under his old alias.

PR #1823, opened 2025-03-06
Fix bug where invalidation messages were getting sent to closing clients

### LINKED ISSUE #1647: [test-failure] Test timeout for RESP3 client redirection for tracked key

madolson:
So I think we were seeing [these timeouts](https://github.com/valkey-io/valkey/actions/runs/13688155322/job/38276139543#step:6:839) because QUIT behaves differently between IO threading and non-IO Threading. In both cases, `QUIT` is a close after reply command. Once the client has written out the results, it gets added to the queue to that gets cleaned up at the end of the event loop. Normally this is fine, as before we circle around to the next event loop this client is definitely killed.

For IO threads, we need to process the pending IO commands to add the client to the kill queue. This may not happen immediately, which means we might go down and process that `SET` command *before* we free the client that is supposedly already quit. This is very sensitive to timing, so it's not very likely, but still possible. Once the `SET` has been executed, the invariants in the tests are off since it will get a correct invalidation.

The fix is to also mark a client as broken if it's being closed. 

The test was also hanging because of a test issue, because the conditional `lsearch` check was returning 1 or 0 strings, which are both valid exit criteria for the wait_for.

```
./runtest --io-threads --accurate --verbose --tags network --dump-logs --single unit/tracking --loops 500 --clients 25
```

Fixes #1647 (I believe this now!)

### DIFF
--- a/src/tracking.c
+++ b/src/tracking.c
@@ -280,7 +280,7 @@ void sendTrackingMessage(client *c, char *keyname, size_t keylen, int proto) {
     int using_redirection = 0;
     if (c->pubsub_data->client_tracking_redirection) {
         client *redir = lookupClientByID(c->pubsub_data->client_tracking_redirection);
-        if (!redir) {
+        if (!redir || redir->flag.close_after_reply || redir->flag.close_asap) {
             c->flag.tracking_broken_redir = 1;
             /* We need to signal to the original connection that we
              * are unable to send invalidation messages to the redirected
--- a/tests/unit/tracking.tcl
+++ b/tests/unit/tracking.tcl
@@ -243,8 +243,7 @@ start_server {tags {"tracking network logreqres:skip"}} {
         r CLIENT TRACKING on REDIRECT $redir_id
         $rd_sg SET key1 1
         r GET key1
-        $rd_redirection QUIT
-        assert_equal OK [$rd_redirection read]
+        r CLIENT KILL ID $redir_id
         $rd_redirection close
         $rd_sg SET key1 2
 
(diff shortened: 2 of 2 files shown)


rjd15372:
Fixes https://github.com/valkey-io/valkey/issues/1647

enjoy-binbin (approved):
Great, it is new to me...

ranshid (commented):
>I will say, that doesn't really explain to me why it was hanging on the final read.

I am not sure how we can thus say is "Fixes" https://github.com/valkey-io/valkey/issues/1647 ?

If I understand you claim that due to the fact that sync io is used, we might be able to identify the client in `sendTrackingMessage` so we will not send the `tracking-redir-broken`?

I agree this does not seem like the root cause for this specific error. 

I would also ask why we are only satisfied with checking the redirection client exists in `sendTrackingMessage` and not try to lookup his flags are not close_asap or close_after_reply? seems like a bug.

madolson:
> I am not sure how we can thus say is "Fixes" https://github.com/valkey-io/valkey/issues/1647 ?

I'm not confident there aren't more edge cases. However, it ran a little over 17 million times over night on my laptop and never hung (although some other tests failed I've never seen fail), whereas it consistently hung after ~100 iterations before the change, so, at the very least this is more stable.

> I would also ask why we are only satisfied with checking the redirection client exists in sendTrackingMessage and not try to lookup his flags are not close_asap or close_after_reply? seems like a bug.

I'm not exactly sure what you mean, we're not sending the tracking message to the client that is closing. EDIT: I understand now, let me try this.

madolson:
Oh, I figured out why it's hanging, the test is not actually checking for the invalidation.

enjoy-binbin (approved):

ranshid (approved):
LGTM

PR #1825, opened 2025-03-06
fix: add samples to stream object consumer trees

### LINKED ISSUE #1824: [BUG] object size sample did not take effect on stream
**Describe the bug**  
The native Valkey `MEMORY USAGE` calculation logic has incomplete sampling implementation and lacks sampling support for the **stream** data structure. Specifically, the memory calculation for the **consumer groups** part of streams does not utilize the sampling logic. This may result in extremely high computational overhead when using the `MEMORY USAGE sample 1` command to estimate memory usage on topics that have slow consumption rates.  

**To reproduce**  
1. Create a Valkey stream with multiple consumer groups and pending messages.
2. Run the command MEMORY USAGE <stream_key> sample 1.
3. Observe significant latency or high CPU usage during the command execution.

**Expected behavior**  
Memory estimation for stream consumer groups should leverage sampling logic to reduce computational overhead, especially in scenarios with slow consumption rates.  

**Additional information**  
This issue primarily affects streams with large numbers of pending messages or complex consumer group structures. A complete sampling implementation for all Valkey data types (including stream internals) is critical for accurate and efficient memory profiling.

arthurkiller:
Use sampling when calculating the memory usage of consumers as part of streams in MEMORY USAGE and MEMORY STATS.

Without this fix, all consumers are traversed, which is slow if there are very many consumers.

Fixes #1824

### DIFF
--- a/src/object.c
+++ b/src/object.c
@@ -1242,28 +1242,36 @@ size_t objectComputeSize(robj *key, robj *o, size_t sample_size, int dbid) {
         if (s->cgroups) {
             raxStart(&ri, s->cgroups);
             raxSeek(&ri, "^", NULL, 0);
-            while (raxNext(&ri)) {
+            samples = 0;
+            elesize = 0;
+            while (samples < sample_size && raxNext(&ri)) {
                 streamCG *cg = ri.data;
-                asize += sizeof(*cg);
-                asize += raxAllocSize(cg->pel);
-                asize += sizeof(streamNACK) * raxSize(cg->pel);
+                elesize += sizeof(*cg);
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
Makes sense, thanks!

Do we have any tests to cover a large number of consumer groups? It would be good to have but I will not require it.

arthurkiller:
> Makes sense, thanks!
> 
> Do we have any tests to cover a large number of consumer groups? It would be good to have but I will not require it.

Thanks for your review. I will try to add them later

PR #1826, opened 2025-03-06
Fix Lua VM crash after FUNCTION FLUSH ASYNC + FUNCTION LOAD

enjoy-binbin:
When backporting, please also pick #2750.

Backported to the 8.1 branch via #2900.

There will be two issues in this test:
```
test {FUNCTION - test function flush} {
    for {set i 0} {$i < 10000} {incr i} {
        r function load [get_function_code LUA test_$i test_$i {return 'hello'}]
    }
    set before_flush_memory [s used_memory_vm_functions]
    r function flush sync
    set after_flush_memory [s used_memory_vm_functions]
    puts "flush sync, before_flush_memory: $before_flush_memory, after_flush_memory: $after_flush_memory"

    for {set i 0} {$i < 10000} {incr i} {
        r function load [get_function_code LUA test_$i test_$i {return 'hello'}]
    }
    set before_flush_memory [s used_memory_vm_functions]
    r function flush async
    set after_flush_memory [s used_memory_vm_functions]
    puts "flush async, before_flush_memory: $before_flush_memory, after_flush_memory: $after_flush_memory"

    for {set i 0} {$i < 10000} {incr i} {
        r function load [get_function_code LUA test_$i test_$i {return 'hello'}]
    }
    puts "Test done"
}
```

The first one is the test output, we can see that after executing FUNCTION FLUSH,
used_memory_vm_functions has not changed at all:
```
flush sync, before_flush_memory: 2962432, after_flush_memory: 2962432
flush async, before_flush_memory: 4504576, after_flush_memory: 4504576
```

The second one is there is a crash when loading the functions during the async
flush:
```
 # valkey 255.255.255 crashed by signal: 11, si_code: 2
 # Accessing address: 0xe0429b7100000a3c
 # Crashed running the instruction at: 0x102e0b09c

------ STACK TRACE ------
EIP:
0   valkey-server                       0x0000000102e0b09c luaH_getstr + 52

Backtrace:
0   libsystem_platform.dylib            0x000000018b066584 _sigtramp + 56
1   valkey-server                       0x0000000102e01054 luaD_precall + 96
2   valkey-server                       0x0000000102e01b10 luaD_call + 104
3   valkey-server                       0x0000000102e00d1c luaD_rawrunprotected + 76
4   valkey-server                       0x0000000102e01e3c luaD_pcall + 60
5   valkey-server                       0x0000000102dfc630 lua_pcall + 300
6   valkey-server                       0x0000000102f77770 luaEngineCompileCode + 708
7   valkey-server                       0x0000000102f71f50 scriptingEngineCallCompileCode + 104
8   valkey-server                       0x0000000102f700b0 functionsCreateWithLibraryCtx + 2088
9   valkey-server                       0x0000000102f70898 functionLoadCommand + 312
10  valkey-server                       0x0000000102e3978c call + 416
11  valkey-server                       0x0000000102e3b5b8 processCommand + 3340
12  valkey-server                       0x0000000102e563cc processInputBuffer + 520
13  valkey-server                       0x0000000102e55808 readQueryFromClient + 92
14  valkey-server                       0x0000000102f696e0 connSocketEventHandler + 180
15  valkey-server                       0x0000000102e20480 aeProcessEvents + 372
16  valkey-server                       0x0000000102e4aad0 main + 26412
17  dyld                                0x000000018acab154 start + 2476

------ STACK TRACE DONE ------
```

The reason is that, in the old implementation (introduced in 7.0), FUNCTION FLUSH
use lua_unref to remove the script from lua VM. lua_unref does not trigger the gc,
it causes us to not be able to effectively reclaim memory after the FUNCTION FLUSH.

The other issue is that, since we don't re-create the lua VM in FUNCTION FLUSH,
loading the functions during a FUNCTION FLUSH ASYNC will result a crash because
lua engine state is not thread-safe.

The correct solution is to re-create a new Lua VM to use, just like SCRIPT FLUSH.


enjoy-binbin:
This problem has existed since 7.0, which means it affects our 7.2 and 8.0. So should we backport it? This code cannot be backported due to the refactoring related to the script module. If we want to backport it, i can re-write the code based on the old branchs.

rjd15372:
@enjoy-binbin thanks for fixing these bugs! I'll do a review later today.

src/lua/engine_lua.c
```diff
@@ -347,21 +350,21 @@ static void luaEngineFreeFunction(ValkeyModuleCtx *module_ctx,
...
-             */
-            lua_unref(lua_engine_ctx->eval_lua, script->function_ref);
-        }
-        zfree(script);
-    } else {
-        luaFunctionFreeFunction(lua_engine_ctx->function_lua, compiled_function->function);
+    lua_State *lua = type == VMSE_EVAL ? lua_engine_ctx->eval_lua : lua_engine_ctx->function_lua;
+    serverAssert(lua);
+
+    luaFunction *script = (luaFunction *)compiled_function->function;
```
rjd15372, line 349: There's a problem here. The `compiled_function->function` pointer points to a `struct luaFunctionCtx` when `VMSE_FUNCTION`, whereas in the `VMSE_EVAL` case, it points to a `struct luaFunction`.

We need to drop the `struct luaFunctionCtx` and use `struct luaFunction` for both subsystems. This will also fix the test failure.

rjd15372 (changes requested):
Apart from the problem of the struct used to represent the compiled function that I mentioned in the inline comment, everything else looks good.

src/lua/engine_lua.c
enjoy-binbin: ohh, i see, i didn't check failure last week, thanks for the review.

rjd15372 (approved):

src/functions.c
```diff
@@ -165,32 +165,62 @@ void functionsLibCtxClear(functionsLibCtx *lib_ctx, void(callback)(dict *)) {
...
+void functionReset(int async, void(callback)(dict *)) {
+    functionsLibCtxReleaseCurrent(async, callback);
+    functionsInit();
+}
+
 /* Free the given functions ctx */
-void functionsLibCtxFree(functionsLibCtx *functions_lib_ctx) {
-    functionsLibCtxClear(functions_lib_ctx, NULL);
+void functionsLibCtxFree(functionsLibCtx *functions_lib_ctx, void(callback)(dict *), list *engine_callbacks) {
+    functionsLibCtxClear(functions_lib_ctx, callback);
```
xbasel, line 205: Why is clear needed if the data structure (including its dictionaries) is already being released and freed?"

xbasel: Although this was here before the change, is it actually needed or just redundant?

src/functions.c
```diff
@@ -165,32 +165,62 @@ void functionsLibCtxClear(functionsLibCtx *lib_ctx, void(callback)(dict *)) {
...
+    }
+}
+
+void functionsLibCtxReleaseCurrent(int async, void(callback)(dict *)) {
     if (async) {
-        functionsLibCtx *old_l_ctx = curr_functions_lib_ctx;
-        curr_functions_lib_ctx = functionsLibCtxCreate();
-        freeFunctionsAsync(old_l_ctx);
+        list *engine_callbacks = listCreate();
+        scriptingEngineManagerForEachEngine(resetEngineFunctionEnvCallback, engine_callbacks);
```
xbasel: `resetEngineFunctionEnvCallback` is misleading, it doesn't reset or act as a callback. It collects reset callbacks. Suggest renaming to `collectEngineResetCallbacks` for clarity.

src/functions.c
```diff
@@ -165,32 +165,62 @@ void functionsLibCtxClear(functionsLibCtx *lib_ctx, void(callback)(dict *)) {
...
     dictRelease(functions_lib_ctx->engines_stats);
     zfree(functions_lib_ctx);
+
+    if (engine_callbacks) {
+        listIter *iter = listGetIterator(engine_callbacks, 0);
+        listNode *node = NULL;
+        while ((node = listNext(iter)) != NULL) {
+            callableLazyEnvReset *engine_callback = listNodeValue(node);
+            if (engine_callback != NULL) {
+                engine_callback->engineLazyEnvResetCallback(engine_callback->context);
```
xbasel, line 217: is this callback responsible for lua-gc?

src/functions.c
```diff
@@ -165,32 +165,62 @@ void functionsLibCtxClear(functionsLibCtx *lib_ctx, void(callback)(dict *)) {
...
+    if (async) {
+        list *callbacks = context;
+        listAddNodeTail(callbacks, callback);
+    }
+}
+
+void functionsLibCtxReleaseCurrent(int async, void(callback)(dict *)) {
     if (async) {
-        functionsLibCtx *old_l_ctx = curr_functions_lib_ctx;
-        curr_functions_lib_ctx = functionsLibCtxCreate();
```
xbasel, line 171: wondering why  the engine is crashing if new `curr_functions_lib_ctx` synchronously created ?

src/functions.c
enjoy-binbin: before, i guess it is a redundant when introduced it. And now we can pass the callback to it.

src/functions.c
enjoy-binbin: yes, it is a bit misleading, i copy the name from `resetEngineEvalEnvCallback`. @rjd15372 thoughts?

src/functions.c
enjoy-binbin: it actually reset the lua env
```
     * Callback function used for resetting the EVAL/FUNCTION context implemented by an
     * engine. This callback will be called by a background thread when it's
     * ready for resetting the context.
```

```
static void resetLuaContext(void *context) {
    lua_State *lua = context;
    lua_gc(lua, LUA_GCCOLLECT, 0);
    lua_close(lua);

#if !defined(USE_LIBC)
    /* The lua interpreter may hold a lot of memory internally, and lua is
     * using libc. libc may take a bit longer to return the memory to the OS,
     * so after lua_close, we call malloc_trim try to purge it earlier.
     *
     * We do that only when the server itself does not use libc. When Lua and the server
     * use different allocators, one won't use the fragmentation holes of the
     * other, and released memory can take a long time until it is returned to
     * the OS. */
    zlibc_trim();
#endif
}
```

src/functions.c
enjoy-binbin: Since we don't call lua_close before, both old ctx and the new ctx are using the same lua vm, one ctx is deleting the function in the bio thread and the other one ctx is loading the functions in the main thread, and look like lua vm is not thread-safe so it will crash.

src/functions.c
rjd15372: I agree that the name is not the best because the function behaves differently whether the reset is async or sync. In sync mode the `resetEngineEvalEnvCallback` will actually reset the lua environment, while in async mode it will just collect callbacks for reseting lua in the background.

So calling it just `collectEngineResetCallbacks` is also not accurate in sync mode.

Maybe call it `resetEngineOrCollectResetCallbacks`?

zuiderkwast:
@madolson You assigned yourself to this PR. Will you review it so we can include it in the patch releases?

hpatro:
@madolson / @rjd15372 Do we plan to get this into Valkey 9.0 ?

rjd15372:
Thanks @hpatro for noticing this PR, it gone under the radar for me. 

@enjoy-binbin we can only merge this bug fix after you fix conflicts with unstable, and finish addressing reviewers' comments. If we don't merge it before 9.0 release, we can always merge it after and then backport it.

src/functions.c
enjoy-binbin: rename to resetEngineOrCollectResetCallbacks

enjoy-binbin:
> we can only merge this bug fix after you fix conflicts with unstable, and finish addressing reviewers' comments. If we don't merge it before 9.0 release, we can always merge it after and then backport it.

it's been a while, please review it again (the new commit or the conflicting part), the comments should be ok now (i forgot a lot TBH...).

src/valkeymodule.h
```diff
@@ -1029,19 +1029,22 @@ typedef size_t (*ValkeyModuleScriptingEngineGetFunctionMemoryOverheadFunc)(
...
+ *
  * - `async`: if has value 1 then the reset is done asynchronously through
  * the callback structure returned by this function.
  */
-typedef ValkeyModuleScriptingEngineCallableLazyEvalReset *(*ValkeyModuleScriptingEngineResetEvalEnvFunc)(
+typedef ValkeyModuleScriptingEngineCallableLazyEnvReset *(*ValkeyModuleScriptingEngineResetEnvFunc)(
     ValkeyModuleCtx *module_ctx,
     ValkeyModuleScriptingEngineCtx *engine_ctx,
+    ValkeyModuleScriptingEngineSubsystemType type,
     int async);
```
rjd15372, line 1045: These API changes are a breaking change, and we need to version the `ValkeyModuleScriptingEngineResetEnvFunc` function pointer since its signature is changed.

src/valkeymodule.h
```diff
@@ -1081,8 +1084,8 @@ typedef struct ValkeyModuleScriptingEngineMethods {
     ValkeyModuleScriptingEngineGetFunctionMemoryOverheadFunc get_function_memory_overhead;
 
     /* The callback function used to reset the runtime environment used
-     * by the scripting engine for EVAL scripts. */
-    ValkeyModuleScriptingEngineResetEvalEnvFunc reset_eval_env;
```
rjd15372, line 1085: This should changed using the same method as the compile code function pointer above.

Like this:
```c
union {
    ValkeyModuleScriptingEngineResetEvalEnvFunc reset_eval_env;
    ValkeyModuleScriptingEngineResetEnvFunc reset_env;
}
```

This also requires to bump the `VALKEYMODULE_SCRIPTING_ENGINE_ABI_VERSION` version, in order for the code that uses this field to decide which branch of the union it should use.

rjd15372 (dismissed):
This PR introduces breaking changes to the scripting engine module API, and I'm afraid that it's to late to merge this at this stage.

Let's discuss this with the @valkey-io/core-team , since on the other hand this is fixing a bug.

madolson:
Core team discussion. Instead of changing the signature, add a new callback so it's an incremental change. This will allow it to be a "bug fix" on 9.0 and 8.1.

@enjoy-binbin Will you have time to update this before GA?

enjoy-binbin:
@rjd15372 i don't have much energy in these days, do you think you can take it over? You should be able to push the commit directly to the PR since you have the write access.

rjd15372:
@enjoy-binbin sure, I'll take over

rjd15372:
@zuiderkwast I've updated the code, and now the scripting engine API changes are backwards compatible. Can you do a quick review of the last two commits that I added?

src/valkeymodule.h
```diff
@@ -1058,7 +1076,12 @@ typedef ValkeyModuleScriptingEngineMemoryInfo (*ValkeyModuleScriptingEngineGetMe
     ValkeyModuleScriptingEngineSubsystemType type);
 
 /* Current ABI version for scripting engine modules. */
-#define VALKEYMODULE_SCRIPTING_ENGINE_ABI_VERSION 2UL
+/* Version Changelog:
+ *  - 1: Initial version.
+ *  - 2: Changed the `compile_code` callback to support binary data in the source code.
+ *  - 3: Added reset_env callback to reset both EVAL or FUNCTION scripts env.
```
zuiderkwast: The `reset_env` callback isn't just added. The old reset_eval_env is replaced with this one.

Also the numbered list is formatted in a funny way, with bullet and number and colon. I think a number and a dot is a more common style.

```suggestion
 *  1. Initial version.
 *  2. Changed the `compile_code` callback to support binary data in the source code.
 *  3. Renamed reset_eval_env callback to reset_env and added an type parameter to be
       able to reset both EVAL or FUNCTION scripts env.
```

src/valkeymodule.h
```diff
@@ -1039,11 +1039,29 @@ typedef size_t (*ValkeyModuleScriptingEngineGetFunctionMemoryOverheadFunc)(
  * - `async`: if has value 1 then the reset is done asynchronously through
  * the callback structure returned by this function.
  */
-typedef ValkeyModuleScriptingEngineCallableLazyEvalReset *(*ValkeyModuleScriptingEngineResetEvalEnvFunc)(
+typedef ValkeyModuleScriptingEngineCallableLazyEnvReset *(*ValkeyModuleScriptingEngineResetEvalFunc)(
     ValkeyModuleCtx *module_ctx,
     ValkeyModuleScriptingEngineCtx *engine_ctx,
     int async);
```
zuiderkwast: This is the old callback type that can't be used in the current version of the API, right? We should prevent module authors from using it, if possible. Can we rename it to something that contains "Old", "DontUse", "BackwardCompat", "V2" or something like that, so users don't get confused and try to use it?

madolson (approved):
I really only looked at the last two commits. The rest of the code looked okay at a high level

zuiderkwast (approved):
I looked only at the valkeymodule.h changes.

I believe the rest is already reviewed and should be fine.

zuiderkwast:
Note: For backporting to 8.0 and 7.2, we need separate PRs. Cherry-picking will not work for this one.

zuiderkwast:
I think we should include this one in the release notes but the title is very hard to understand for users.

Can we come up with a better title to use in release notes? How about *Fix Lua VM crash after FUNCTION FLUSH ASYNC + FUNCTION LOAD* ?

Issue #1775, opened 2025-02-25
[CRASH] valkey crashed after close same key twice via invoking ValkeyModule_CloseKey()


zh1029:
### Crash report

```shell
valkey 8.0.2

=== VALKEY BUG REPORT START: Cut & paste starting from here ===
26099:M 25 Feb 2025 02:13:46.791 # ------------------------------------------------
26099:M 25 Feb 2025 02:13:46.791 # !!! Software Failure. Press left mouse button to continue
26099:M 25 Feb 2025 02:13:46.791 # Guru Meditation: decrRefCount against refcount <= 0 #object.c:396

------ STACK TRACE ------

26102 bio_lazy_free
/lib64/libc.so.6(+0x93da9)[0x7f05aa8c8da9]
/lib64/libc.so.6(pthread_cond_wait+0x229)[0x7f05aa8cb7b9]
./src/valkey-server 127.0.0.1:6379(bioProcessBackgroundJobs+0x18e)[0x53225e]
/lib64/libc.so.6(+0x97697)[0x7f05aa8cc697]
/lib64/libc.so.6(+0x11b50c)[0x7f05aa95050c]

26101 bio_aof
/lib64/libc.so.6(+0x93da9)[0x7f05aa8c8da9]
/lib64/libc.so.6(pthread_cond_wait+0x229)[0x7f05aa8cb7b9]
./src/valkey-server 127.0.0.1:6379(bioProcessBackgroundJobs+0x18e)[0x53225e]
/lib64/libc.so.6(+0x97697)[0x7f05aa8cc697]
/lib64/libc.so.6(+0x11b50c)[0x7f05aa95050c]

26100 bio_close_file
/lib64/libc.so.6(+0x93da9)[0x7f05aa8c8da9]
/lib64/libc.so.6(pthread_cond_wait+0x229)[0x7f05aa8cb7b9]
./src/valkey-server 127.0.0.1:6379(bioProcessBackgroundJobs+0x18e)[0x53225e]
/lib64/libc.so.6(+0x97697)[0x7f05aa8cc697]
/lib64/libc.so.6(+0x11b50c)[0x7f05aa95050c]

26099 valkey-server *
./src/valkey-server 127.0.0.1:6379[0x48e49a]
./src/valkey-server 127.0.0.1:6379[0x49bdcc]
./src/valkey-server 127.0.0.1:6379(processInputBuffer+0x15b)[0x4935ab]
[...]

PR #1834, opened 2025-03-10
Update CloseKey module API documentation to avoid use-after-free behavior

sungming2:
### Issue: https://github.com/valkey-io/valkey/issues/1775

Without a clear comment, users might mistakenly attempt to access or close the same key object multiple times, leading to use-after-free errors, undefined behavior, or crashes due to double-free operations.

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -4199,7 +4199,7 @@ static void moduleCloseKey(ValkeyModuleKey *key) {
     decrRefCount(key->key);
 }
 
-/* Close a key handle. */
+/* Close a key handle. The key handle is freed and should not be accessed anymore */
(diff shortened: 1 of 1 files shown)


src/module.c
```diff
@@ -4199,7 +4199,7 @@ static void moduleCloseKey(ValkeyModuleKey *key) {
     decrRefCount(key->key);
 }
 
-/* Close a key handle. */
+/* Close a key handle. The key handle is freed and should not be accessed anymore */
```
hpatro:
```suggestion
/* Close a key handle. The key handle is freed and should not be accessed anymore. */
```

hpatro (approved):

PR #1838, opened 2025-03-11
Replication Flow Control – Prioritizing replication traffic in the replica

### LINKED ISSUE #1596: [NEW] Introduce Quality-of-Service for the replication stream to reduce full sync as a result of buffer overruns
When replica nodes are very busy processing customer traffic, the replication stream can get starved, causing disconnections due to exceeding buffer limits. This often triggers a full sync.

To reduce the likelihood of full syncs caused by client output buffer overruns, we can add a quality-of-service mechanism in the replica to prioritize replication traffic during high load.

This feature improves the availability of replicas and the stability of primaries by reducing the chances of full syncs.

The replica can detect replication traffic bursts by monitoring the buffer during primary socket reads. For example, if the application buffer has been fully filled for the last N reads, the replica can assume the kernel level TCP receive queue buffer isn’t empty and traffic is high. In this case, it can prioritize more socket reads for the primary file descriptor, helping to drain the shared replication buffer in the primary faster and reducing the chances of full syncs.

xbasel:
## Overview
This PR introduces Replication Flow Control (`repl-flow-control`), a dynamic mechanism that prioritizes replication traffic on the replica side. By detecting replication pressure and adjusting read frequency adaptively, it reduces the risk of primary buffer overflows and full syncs.

## Problem
In high-load scenarios, a replica might not consume replication data fast enough, leading to backpressure on the primary. When the primary’s buffer overflows, replication lag increases and it can drops the replica connection, triggering a full sync, a costly operation that impacts system performance.

Without this feature:

- Replication reads occur at a fixed rate, irrespective of data pressure.
- If the replica falls behind, the primary accumulates replication data leading to higher memory utilization.
- Once the primary buffer overflows, the connection drops, forcing a full sync.
- Full syncs cause high memory, CPU, and I/O spikes.

## Solution: Replication Flow Control

`repl-flow-control` enables the replica to dynamically increase its replication read rate if it detects that replication data is accumulating. The mechanism operates as follows:

**Detecting replication pressure**
Each read from the primary is checked against the max buffer byte limit. If the read hit the limit (filled the buffer), suggesting more data is likely available.

**Prioritizing replication reads**
If replication pressure is detected, the replica invokes multiple reads per I/O event instead of a single one. This allows the replica to catch up faster, reducing memory consumption and buffer overflows on the primary.

## Performance Impact ##

Test setup:
1. Bombard the replica with expensive commands, leading to high CPU utilization
2. Write to the main database to trigger replication traffic

image

<h4 data-start="2324" data-end="2363"><strong data-start="2329" data-end="2363">Latency and Throughput Changes</strong></h4>

Metric | Before (repl-flow-control Disabled) | After (repl-flow-control Enabled)
-- | -- | --
Throughput (requests/sec) | 941.71 | 760.98
Avg Latency (ms) | 52.865 | 65.534
p50 Latency (ms) | 59.743 | 68.543
p95 Latency (ms) | 79.231 | 106.687
p99 Latency (ms) | 90.303 | 126.527
Max Latency (ms) | 188.031 | 385.535

📌 Observations:

- Replication stability improves,no full syncs were observed after enabling flow control.
- Higher latency for normal clients due to increased resource allocation for replication.
- CPU and memory usage remain stable, with no major overhead.
- Replica throughput slightly decreases as replication takes priority.

**TODO**
- Consider limiting the maximum number of reads per event to a ratio of the total number of events returned by the epoll cycle. For example, if the ratio is 20% and EPOLL returns 100 events, the replica can read from the primary up to 20 times per primary I/O event. 
    
Closes #1596


valkey.conf
```diff
@@ -2535,3 +2535,22 @@ jemalloc-bg-thread yes
...
+################################## REPLICATION FLOW CONTROL ##################################
+
+# Prioritizes replication traffic to reduce primary buffer overflows,
+# reducing lag and the risk of full syncs. Allows the replica to
+# consume data faster under high load.
+#
+# If enabled, the replica invokes multiple reads per I/O event when it
+# detects replication pressure.
+#
+# Default: yes
```
hwware: I have reservations about the default value is yes due to the lower performance.

src/server.h
```diff
@@ -1181,6 +1181,7 @@ typedef struct client {
     /* Input buffer and command parsing fields */
     sds querybuf;        /* Buffer we use to accumulate client queries. */
     size_t qb_pos;       /* The position we have read in querybuf. */
+    int qb_full_read;    /* True if the last read returned the maximum allowed bytes */
```
hwware: It looks like a boolean variable (I check the codes, the variable is 1 or 0), how about change to is_qb_full_read or similar? It is easy to read for others.

src/networking.c
```diff
@@ -3264,19 +3266,66 @@ void readToQueryBuf(client *c) {
...
+ *
+ * @return          1 if another read should be attempted, 0 otherwise.
+ */
+int shouldRepeatRead(client *c, int iteration) {
+    // If the client is not a primary replica, is closing, or flow control is disabled, no more reads.
+    if (!(c->flag.primary) || c->flag.close_asap || !server.repl_flow_control_enabled) {
+        return 0;
+    }
+
+    bool is_last_iteration = iteration >= server.repl_cur_reads_per_io_event;
```
hwware: I do not find where this variable repl_cur_reads_per_io_event initialization, did you do in somewhere else? If not, it is dangerous, please initialize it.

src/networking.c
```diff
@@ -3264,19 +3266,66 @@ void readToQueryBuf(client *c) {
...
+
+    if (is_last_iteration) {
```
madolson: I suppose I don't really understand why this needs to be adaptive. As long as we got a full read, why can't we repeat up until the max io event size? There is a comment about "The read limit ramps up gradually if full reads continue, avoiding sudden spikes.", but we are just deferring the spikes until later. Most users don't like replication lag, they would rather have the most up to date data if possible.

xbasel: > I suppose I don't really understand why this needs to be adaptive. As long as we got a full read, why can't we repeat up until the max io event size? There is a comment about "The read limit ramps up gradually if full reads continue, avoiding sudden spikes.", but we are just deferring the spikes until later. Most users don't like replication lag, they would rather have the most up to date data if possible.

I have no strong opinion here, probably reading up to the max reads is enough (and simpler).

zuiderkwast (commented):
This looks to me like we're fixing a stability issue. Why would anyone want to disable it? Let's discuss if we actually need a config for this. I think we maybe don't need it and we can just keep this always enabled.

I think the benchmark numbers don't give a fair picture. Without this feature, there is a problem of replication lag and for the full sync, it means extra resources used by replica and primary and maybe even extra latency for commands sent to the primary.

Also, this affects only the latency of read-from-replicas. I think this is less important than read-from-primary. If the replica is not fast enough, the client can read from the primary or from another replica. If they do that, it's good because it will reduce the load on the overloaded replica.

artikell:
This seems to be the rate limiting for the control reading stage. There are two scenarios that also need to be considered:
- The time-consuming of executing commands is too high. like: `ZDIFFSTORE` some big key
- In the case of multi-threading, will reading not take up more time?

valkey.conf
xbasel: Under normal load, this shouldn't impact latency, as behavior remains the same as today.
Enabling this by default prioritizes replication stability and low replication lag, avoiding full syncs,  which IMO are far more disruptive than occasional latency impact.

So the real question is: what do we prefer?
 * With it off: better p99/p999 client latency, but with occasional high replication lag and disruptive full syncs.
 * With it on: slightly degraded client performance at times, but stable replication and low lag.

src/networking.c
xbasel: I forgot to initialize it

xbasel:
> This seems to be the rate limiting for the control reading stage. There are two scenarios that also need to be considered:
> 
>     * The time-consuming of executing commands is too high. like: `ZDIFFSTORE` some big key
> 
>     * In the case of multi-threading, will reading not take up more time?

Could you clarify what you meant about the control reading stage and those two scenarios?
Do you mean if clients run expensive commands in the replica, this mechanism won't be effective? Or that if we read more from the primary in one go and process expensive commands in the replication context, the replica might become even more unresponsive?

src/networking.c
xbasel: I've changed the impl to not ramp up, and read up to the threshold.

artikell:
> > This seems to be the rate limiting for the control reading stage. There are two scenarios that also need to be considered:
> > ```
> > * The time-consuming of executing commands is too high. like: `ZDIFFSTORE` some big key
> > 
> > * In the case of multi-threading, will reading not take up more time?
> > ```
> 
> Could you clarify what you meant about the control reading stage and those two scenarios? Do you mean if clients run expensive commands in the replica, this mechanism won't be effective? Or that if we read more from the primary in one go and process expensive commands in the replication context, the replica might become even more unresponsive?

As you understand, in many cases, expensive commands don't necessarily have a large throughput. But they can also lead to the disconnection between the master and the slave.

src/nohup.out
```diff
@@ -0,0 +1 @@
+stress: info: [1201400] dispatching hogs: 1 cpu, 0 io, 0 vm, 0 hdd
```
Fusl: should remove this file from the commit

valkey.conf
```diff
@@ -2535,3 +2535,22 @@ jemalloc-bg-thread yes
 # the empty string.
 #
 # availability-zone "zone-name"
+
+################################## REPLICATION FLOW CONTROL ##################################
```
madolson: I don't think we want a dedicated section for this, I would move it under the normal replication configs.

src/networking.c
```diff
@@ -3264,19 +3266,47 @@ void readToQueryBuf(client *c) {
     }
 }
 
+/**
+ * This function is designed to prioritize replication flow.
+ * Determines whether the replica should continue reading from the primary.
+ * It dynamically adjusts the read rate based on buffer utilization
+ *
+ * @return          1 if another read should be attempted, 0 otherwise.
```
madolson, line 3274:
```suggestion
 * Returns 
```

src/networking.c
```diff
@@ -3247,6 +3248,7 @@ void readToQueryBuf(client *c) {
     if (c->nread <= 0) {
         return;
     }
+    c->is_qb_full_read = (size_t)c->nread == readlen ? 1 : 0;
```
```suggestion
    c->is_qb_full_read = (size_t)c->nread == readlen;
```
This is a bool.

src/networking.c
```diff
@@ -3264,19 +3266,47 @@ void readToQueryBuf(client *c) {
...
 }
 
+/**
+ * This function is designed to prioritize replication flow.
+ * Determines whether the replica should continue reading from the primary.
+ * It dynamically adjusts the read rate based on buffer utilization
+ *
+ * @return          1 if another read should be attempted, 0 otherwise.
+ */
+int shouldRepeatRead(client *c, int iteration) {
```
```suggestion
static bool shouldRepeatReadFromPrimary(client *c, int iteration) {
```
Kind of a nit, I think generic function names are often confusing when it does something very specific.

src/networking.c
madolson: This comment is a bit stale.

valkey.conf
madolson: My general hot take. I would prefer have the default be on, but also only have a single config `repl-max-reads-per-io-event 25`. If someone wants to change it back to default, they can change that config to be 1. I don't really like having an "enabled" and "value" config when we can just have one.

madolson (commented):
I like the new design, mostly minor comments.

src/config.c
```diff
@@ -3295,6 +3295,7 @@ standardConfig static_configs[] = {
     createIntConfig("rdma-port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.rdma_ctx_config.port, 0, INTEGER_CONFIG, NULL, updateRdmaPort),
     createIntConfig("rdma-rx-size", NULL, IMMUTABLE_CONFIG, 64 * 1024, 16 * 1024 * 1024, server.rdma_ctx_config.rx_size, 1024 * 1024, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("rdma-completion-vector", NULL, IMMUTABLE_CONFIG, -1, 1024, server.rdma_ctx_config.completion_vector, -1, INTEGER_CONFIG, NULL, NULL),
+    createIntConfig("repl-max-reads-per-io-event", NULL, MODIFIABLE_CONFIG, 1, INT_MAX, server.repl_max_reads_per_io_event, 25, INTEGER_CONFIG, NULL, NULL),
```
madolson: Suggestion is to just start with a hardcoded value of 25, then we can evaluate making it better by making the behavior more dynamic.

If we need to, we can add a configuration later.

zuiderkwast: > Suggestion

More than a suggestion. Exactly this was actually decided in the core team meeting.

madolson: lol, maybe I'm too passive in my wording. I try to avoid acting like it's an order. We nicely asked for it :)

zuiderkwast: It needs to be clear that it was a decision. Otherwise we might come back and discuss it again and again. :)

madolson: Sure sure, I'll try to do that moving forward.

src/server.h
```diff
@@ -1187,16 +1188,17 @@ typedef struct client {
...
-    int argc;            /* Num of arguments of current command. */
-    int argv_len;        /* Size of argv array (may be more than argc) */
-    size_t argv_len_sum; /* Sum of lengths of objects in argv list. */
-    int reqtype;         /* Request protocol type: PROTO_REQ_* */
-    int multibulklen;    /* Number of multi bulk arguments left to read. */
-    long bulklen;        /* Length of bulk argument in multi bulk request. */
-    long long woff;      /* Last write global replication offset. */
+    sds querybuf;         /* Buffer we use to accumulate client queries. */
+    size_t qb_pos;        /* The position we have read in querybuf. */
+    bool is_qb_full_read; /* True if the last read returned the maximum allowed bytes */
```
zuiderkwast: This adds 4 bytes (or 8, if the other surrounding fields need 8 bytes alignment) to every client in primaries and replicas, but we only need it for one very specific client.

Some other ideas to save memory:

* Use a flag. I know we have very few unused flags left, but it's fine to use them. (When it's full, we can add another 32 or 64 bits for more flags.)
* Change the return type of `readToQueryBuf` from void to bool or an enum, to indicate if it was a full or a short read.

src/networking.c
```diff
@@ -3362,19 +3363,42 @@ void readToQueryBuf(client *c) {
...
+
 void readQueryFromClient(connection *conn) {
     client *c = connGetPrivateData(conn);
     /* Check if we can send the client to be handled by the IO-thread */
     if (postponeClientRead(c)) return;
 
     if (c->io_write_state != CLIENT_IDLE || c->io_read_state != CLIENT_IDLE) return;
 
-    readToQueryBuf(c);
+    bool shouldRepeat = false;
```
zuiderkwast: We usually use snake case for local variables.

src/networking.c
```diff
@@ -3362,19 +3363,42 @@ void readToQueryBuf(client *c) {
...
-        if (processInputBuffer(c) == C_ERR) return;
-    }
-    beforeNextClient(c);
+        if (handleReadResult(c) == C_OK) {
+            if (processInputBuffer(c) == C_ERR) return;
+        }
+        iter++;
+        shouldRepeat = shouldRepeatReadFromPrimary(c, iter);
+        beforeNextClient(c);
+    } while (shouldRepeat);
```
zuiderkwast: Does shouldRepeatReadFromPrimary need to be called before beforeNextClient? If not, we can do the check in the while condition:

```suggestion
        beforeNextClient(c);
    } while (shouldRepeatReadFromPrimary(c, iter));
```

zuiderkwast (commented):
This PR became very small and simple without the config. Nice. Just some minor comments.

src/networking.c
xbasel: Yes. `shouldRepeatReadFromPrimary` must run before `beforeNextClient`, since the latter may free the client.

src/server.h
xbasel: No variables, no flags, returning the result from `readToQueryBuf()` is simple  (no state to track).

src/networking.c
```diff
@@ -3280,13 +3281,17 @@ int processInputBuffer(client *c) {
 
 /* This function can be called from the main-thread or from the IO-thread.
  * The function allocates query-buf for the client if required and reads to it from the network.
- * It will set c->nread to the bytes read from the network. */
-void readToQueryBuf(client *c) {
+ * It will set c->nread to the bytes read from the network.
+ * Returns non-zero if the buffer was filled (more data may be available).
+ */
+
```
zuiderkwast: Does it return bool or int? Let's stick to one of them in wording and types.

```suggestion
 * Returns true if the buffer was filled (more data may be available). */
```

src/networking.c
```diff
@@ -3280,13 +3281,17 @@ int processInputBuffer(client *c) {
...
+static bool readToQueryBuf(client *c) {
     int big_arg = 0;
     size_t qblen, readlen;
+    int ret = 0;
```
zuiderkwast: Use bool for this variable if the function returns bool.

We could use a variable name that describes what the return value means, e.g. `full`, `full_read`, etc., instead of ret.

src/networking.c
```diff
@@ -3343,9 +3348,9 @@ void readToQueryBuf(client *c) {
 
     c->nread = connRead(c->conn, c->querybuf + qblen, readlen);
     if (c->nread <= 0) {
-        return;
+        return ret;
     }
-
+    ret = (size_t)c->nread == readlen;
```
zuiderkwast: (Nit) The error cases don't really need to return the variable. They can just return `false`. We can declare the variable where we actually need it. (Maybe it's just personal preference. Ignore it if you don't agree.)

```c
    if (c->nread <= 0) {
        return false;
    }

    bool full_read = (size_t)c->nread == readlen;
```

zuiderkwast (commented):
Just nits.

zuiderkwast (approved):

madolson (approved):

PR #1839, opened 2025-03-12
Add --sequential to valkey-benchmark (for populating entire keyspace)

rainsupreme:
I needed a way to systematically populate the keyspace before running `valkey-benchmark` tests like get, so I added a `--sequential` option that modifies the behavior of `-r` so that the replacements are sequential numbers in the keyspace instead of random numbers.

Example:
```
# populate keys by sequentially setting them
./valkey-benchmark -d 512 --sequential -r 3000 -n 3000 -c 650 -P 4 --threads 64 -t set -q

# benchmark get with 100% hit rate
./valkey-benchmark -r 3000 -n 250000 -c 650 -P 4 --threads 64 -t get
```


enjoy-binbin (approved):
LGTM, seems like a good idea.

madolson:
@valkey-io/core-team Explicitly pinging yall for votes, as this is a new API on one of our tools.

src/valkey-benchmark.c
```diff
@@ -393,18 +394,23 @@ static void resetClient(client c) {
...
+    static _Atomic size_t seq_key = 0;
+    for (size_t i = 0; i < c->randlen; i++) {
         char *p = c->randptr[i] + 11;
-        size_t r = 0;
-        if (config.randomkeys_keyspacelen != 0) r = random() % config.randomkeys_keyspacelen;
-        size_t j;
+        size_t key = 0;
+        if (config.keyspacelen != 0) {
+            if (config.sequential_replacement) {
+                key = atomic_fetch_add_explicit(&seq_key, 1, memory_order_relaxed);
```
zuiderkwast, line 404: To guarantee that multiple threads get unique values here, can we really use relaxed menory ordering? Can you explain why this is enough?

I would guess we need sequential ordering.

madolson:
Removing the major tag. As long as it's not a breaking change, benchmark changes.

src/valkey-benchmark.c
zuiderkwast: This is difficult stuff. I tried reading these two pages:

* https://en.cppreference.com/w/c/atomic/atomic_fetch_add
* https://en.cppreference.com/w/c/atomic/memory_order

Now I believe `memory_order_relaxed` is enough. The numbers may not be returned in right order according to the clock but I believe every number will be returned exactly once, because the function is atomic.

src/valkey-benchmark.c
```diff
@@ -1541,14 +1549,15 @@ int parseOptions(int argc, char **argv) {
         " -k <boolean>       1=keep alive 0=reconnect (default 1)\n"
         " -r <keyspacelen>   Use random keys for SET/GET/INCR, random values for SADD,\n"
         "                    random members and scores for ZADD.\n"
-        "                    Using this option the benchmark will expand the string\n"
-        "                    __rand_int__ inside an argument with a 12 digits number in\n"
+        "                    Using this option the benchmark will replace the string\n"
+        "                    __rand_int__ inside an argument with a random 12 digit number in\n"
         "                    the specified range from 0 to keyspacelen-1. The\n"
```
zuiderkwast: This is printed in a terminal which is 80 chars wide by default. It's not about the width of the source code here but the length of the string between "...\n". With this change the line overflows this length.

It's easy to compensate for it though.

```suggestion
        "                    Using this option the benchmark will replace the string\n"
        "                    __rand_int__ inside an argument with a random 12 digit\n"
        "                    number in the specified range from 0 to keyspacelen-1. The\n"
```

src/valkey-benchmark.c
```diff
@@ -1541,14 +1549,15 @@ int parseOptions(int argc, char **argv) {
...
         "                    substitution changes every time a command is executed.\n"
         "                    Default tests use this to hit random keys in the specified\n"
         "                    range.\n"
         "                    Note: If -r is omitted, all commands in a benchmark will\n"
         "                    use the same key.\n"
+        " --sequential       modifies -r argument to replace with sequential 12 digit numbers.\n"
```
```suggestion
        " --sequential       Modifies the -r argument to replace the string __rand_int__\n"
        "                    with 12 digit numbers sequentially instead of randomly.\n"
```

zuiderkwast (approved):

Issue #653, opened 2024-06-14
Revise config defaults for Valkey 8


zuiderkwast:
@madolson wrote in redis/redis#12191:

> The immediate ones that come to mind:
>
> * `replica-lazy-flush no` -> `yes`
> * `lazyfree-lazy-user-flush no` -> `yes`
>
> There are many problems with running flush synchronously. Even in single CPU environments, the thread managers should balance between the freeing and serving incoming requests. (handle in #913)
>
>We should probably also revisit all of the datastructure limits (set-max-listpack-entries for example), which were probably tuned in a world with CPUs from a decade ago.

The `set-max-listpack-entries` was discussed here, where it was first added: https://github.com/redis/redis/pull/11290#discussion_r993643256

Ideally we should tune this in some smart way, but I assume nobody will do this, I suggest the following (pseudo-random :rofl:) new defaults:

* hash-max-listpack-entries 512 *(Keep this default? An entry has key and value so the listpack's size is 1024 elements)*
* list-max-listpack-size -2 *(No suggestion. Any ideas?)*
* set-max-intset-entries 512 -> 1024
* set-max-listpack-entries 128 -> 1024
* zset-max-listpack-entries 128 -> 512 *(Each entry is element and score, so listpack size is 1024)*
* hll-sparse-max-bytes 3000 *(Keep this?)*
* stream-node-max-bytes 4096 *(Keep this?)*

@eduardobr wrote in redis/redis#12747:

> `repl-backlog-size 1mb` is too small in most cases, but at the same time I think the real fix would be to make it dynamic (rough example: set min and max instead, and let redis choose a value in this range over time)

@xuliang0317 wrote in redis/redis#12747:

> Now，repl-backlog-size is still 1mb,For most scenarios, this value is too small.Network transmission and bandwidth performance have improved rapidly in more than ten years，Memory is getting cheaper and cheaper.
> Although the impact is noted in the documentation, most Redis users will directly use the default configuration.In many scenarios, the traffic per second of a single node exceeds 1mb.Especially for businesses facing individual users.
> I recommend at least updating the configuration to 10mb to improve the rationality of the current repl-backlog-size.

So I suggest `repl-backlog-size 1m -> 10m`. Handle in #911

soloestoy:
some suggestions:

* maxmemory-clients `0 -> 100%`, limit client memory usage to avoid data eviction caused by network read/write buffering, or even machine system OOM.
[...]

PR #1840, opened 2025-03-12
Added comment to maxmemory-clients conf to mention the performance issue

enjoy-binbin:
It's worth mentioning that enabling `maxmemory-clients` will affect the
performance so people can perform its own benchmark before enabling it.

## This is an old top comment

The client eviction introduced in 2753429c99425e3d0216cba79e0e61192975f252,
at that time, the main idea of our introduction was to protect key data
from being evicted, which is the server's perspective.

In fact, the accumulated client output buffer can also violate the maxmemory
contract and causing the machine to OOM.


hwware:
I am not sure if 100% as default value is reasonable

hpatro (commented):
I think we shouldn't enable it by default. It cause throughput degradation (extra cpu cycle during command processing) due to constant tracking/metric update of client's input/output buffer size during read/write flow. 

https://github.com/redis/redis/pull/11348

zuiderkwast:
We discussed this briefly before we released 8.0. Here is the discussion: https://github.com/valkey-io/valkey/issues/653#issuecomment-2178153609.

My thoughts:

* Clients using 100% of the configured maxmemory is very unlikely. If it happens, then we have other serious problems: It means all keys are being evicted or are already evicted, or if key eviction is disabled, we are already far over maxmemory and clients receive the OOM error.
* I'm concerned about the performance degradation raised by @hpatro.
* A better value would be 50% or 10% but we can't use that as the default value because it would be a breaking change. I think users who want client eviction need to configure it themselves.

hwware:
> We discussed this briefly before we released 8.0. Here is the discussion: [#653 (comment)](https://github.com/valkey-io/valkey/issues/653#issuecomment-2178153609).
> 
> My thoughts:
> 
> * Clients using 100% of the configured maxmemory is very unlikely. If it happens, then we have other serious problems: It means all keys are being evicted or are already evicted, or if key eviction is disable, we are already far over maxmemory and clients receive the OOM error.
> * I'm concerned about the performance degradation raised by @hpatro.
> * A better value would be 50% or 10% but we can use that as the default value because it would be a breaking change. I think users who want client eviction need to configure it themselves.

I prefer to leave it and let client decide the default value.

madolson:
Consensus from our weekly meeting was that we aren't entirely clear what problem this is trying to solve. There were concerns about folks migrating from Redis seeing different performance. So the preference was to not change the default and let users decide. This seems aligned with what other folks have said in this thread.

enjoy-binbin:
@hpatro @zuiderkwast 

I revert the change and only add the comment to mention the performance issue, please check.

valkey.conf
```diff
@@ -2439,6 +2439,13 @@ client-output-buffer-limit pubsub 32mb 8mb 60
 # is based on a percentage of the maxmemory setting. For example to set client
 # eviction at 5% of maxmemory:
 # maxmemory-clients 5%
+#
+# By default client eviction is disabled since it is mostly not needed if you
+# don't have memory issues, and collecting data has a performance impact, can
+# be measured under big load. Client eviction can easily be enabled at runtime
+# using the command "CONFIG SET maxmemory-clients <value>" if needed.
+#
+# maxmemory-clients 0
```
zuiderkwast, line 2444: This file has a lot of text. We should try to keep it short.

It's better that we also edit the existing text, not only add more text.

There is already this text above which mentions the default. It's wrong though because it doesn't include the config name. :D We need to delete that.

```
# Client eviction is configured using the maxmemory-clients setting as follows:
# 0 - client eviction is disabled (default)
```

I think we can replace that text above with the new text, but I don't think we need to say it can be modified with CONFIG SET. That's true for most of the configs. I think this is enough:

```
# Client eviction is configured using the maxmemory-clients directive, which is
# 0 (disabled) by default, because it has a small performance impact under high
# load. It can be enabled at runtime when needed.
```

enjoy-binbin: SGTM. 

I actually copy that text from
```
# By default latency monitoring is disabled since it is mostly not needed
# if you don't have latency issues, and collecting data has a performance
# impact, that while very small, can be measured under big load. Latency
# monitoring can easily be enabled at runtime using the command
# "CONFIG SET latency-monitor-threshold <milliseconds>" if needed.
latency-monitor-threshold 0
```

zuiderkwast (approved):

valkey.conf
zuiderkwast: Ah, i didn't know

PR #1843, opened 2025-03-13
Update COPY and MOVE command description clearer in JSON file

hwware:
As I do the code review for PR https://github.com/valkey-io/valkey/pull/1671 (Add multi-database support to cluster mode), I find
the return value message of the command COPY and MOVE is not very clear to user. Thus I update them in this PR and valkey-doc repo as PR https://github.com/valkey-io/valkey-doc/pull/249

### DIFF
--- a/src/commands/copy.json
+++ b/src/commands/copy.json
@@ -82,7 +82,7 @@
                     "const": 1
                 },
                 {
-                    "description": "Source was not copied.",
+                    "description": "Source was not copied when the destination key already exists",
                     "const": 0
                 }
             ]
--- a/src/commands/move.json
+++ b/src/commands/move.json
@@ -52,7 +52,7 @@
                     "const": 1
                 },
                 {
-                    "description": "Key wasn't moved.",
+                    "description": "When key already exists in the destination database, or it does not exist in the source database",
                     "const": 0
                 }
             ]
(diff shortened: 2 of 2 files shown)


src/commands/copy.json
```diff
@@ -82,7 +82,7 @@
                     "const": 1
                 },
                 {
-                    "description": "Source was not copied.",
+                    "description": "Source was not copied when the destination key already exists",
```
madolson, line 85: Is this description actually used anywhere?

madolson (approved):
I'm okay with this wording.

src/commands/copy.json
hwware: The words come from our webpage https://valkey.io/commands/move/ and https://valkey.io/commands/copy/,  it makes user easier to understand

src/commands/move.json
```diff
@@ -52,7 +52,7 @@
                     "const": 1
                 },
                 {
-                    "description": "Key wasn't moved.",
+                    "description": "When key already exists in the destination database, or it does not exist in the source database",
```
enjoy-binbin: Lets also keep the old wording? The short one is a good wording too i think
```suggestion
                    "description": "Key wasn't moved. When key already exists in the destination database, or it does not exist in the source database",
```

src/commands/copy.json
zuiderkwast: The reply text on the website comes from the doc repo `resp2_replies.json` and `resp3_replies.json`.

Issue #788, opened 2024-07-15
[NEW] Can we add 'count' option to the 'RANDOMKEY' command?


AlanZhang1204:
Regarding to https://github.com/redis/redis/issues/11915   

Shall we add this feature to valkey?   

In https://github.com/redis/redis/pull/11987, @hwware add three options to randomkeycommand, that is count, duplicate and pattern. But under certain conditions, the time complexity will be O(n). Further more, the memory footprint will increase a lot under some extreme conditions. I commit a demo https://github.com/valkey-io/valkey/commit/2f4fe39079437c5cbe93f6ec91d47eadb6f7b3b6 to explain this. I think I can explain my idea better using this commit, though there remains some bug.   

*case1:* Duplicate is enabled, and no pattern, so the extraction method is just: "return N random elements" sampling the whole db every time.  This case return the element in random order.   

For the following case, I think we'd better extract the key that meets the requirements. But this may cause a huge memory footprint if lots of keys conform to the pattern and the keys(not key-value)are big. And, this step is a O(n) process like keyscommand, which may lead to a block of the valkey.   

*case2:* Duplicate is enabled, and some pattern required, we can do a ramdom select in the dict we created.   
*case3:* If duplicate disabled and the size of conformed keys is equal or smaller than the number of keys user required. In this case we can simply return all keys in the dict.   

If the code run to this case, the condition is: duplicate disabled and the conformed keys number is larger than required numbers. I refer this condition to "srandmemberWithCountCommand" case3&4.
*case4:* If the number of elements inside the dict is not greater than the number of requested elements. We do random deletions of the dict and return the remaining keys. The range need to be tested out.
*case5:* Opposite to case4, if the number of elements inside the dict is a lot greater than the number of requested elements. We just do a random select, then return the key and delete it from dict.

Above are my thoughts about the feature. @hwware Plz help me figure out whether we need this feature and if there is any better idea to implement the program. Thanks a lot.

madolson:
Can you add more information on why you think this feature is useful? I see the one use case which is random sampling, but it seems OK to just do `randomkey` a bunch in a multi-exec for that case. Is performance really that sensitive there?

AlanZhang1204:
> Can you add more information on why you think this feature is useful? I see the one use case which is random sampling, but it seems OK to just do `randomkey` a bunch in a multi-exec for that case. Is performance really that sensitive there?

[...]

PR #1847, opened 2025-03-13
Add 'count' option to the 'RANDOMKEY' command

hwware:
Reference the issue https://github.com/redis/redis/issues/11915 and https://github.com/valkey-io/valkey/issues/788
I would like to add one more option for RANDOMKEY command, allow user to get multiply random keys once.
It is similar to existing command ZRANDMEMBER key [count [WITHSCORES]] and SRANDMEMBER key [count]

The new command format as:  RANDOMKEY  [COUNT count]

If the provided count argument is positive, return an array of distinct elements. 
If called with a negative count, the behavior changes and the command is allowed to return the same elements multiple times.


madolson:
I'm still not inclined to accept this feature. I don't think it is really that useful.

hwware:
> I'm still not inclined to accept this feature. I don't think it is really that useful.

It is a very old requirement from Redis community, if recently there is no more people to care about it, I will close it later

hwware:
There is no requirement recently, I decide to close this PR.

PR #1850, opened 2025-03-14
Fix RANDOMKEY infinite loop during CLIENT PAUSE

### LINKED ISSUE #1848: [BUG] RANDOMKEY may cause infinite loop during client pause write.
**Describe the bug**

When the ``client pause write`` is set and all the keys in the server are expired keys, executing the ``randomkey`` command will lead to an infinite loop.

**To reproduce**

1. First, write a key with an expiration time. For example:

``` 
set key value ex 30
```

2. Second, execute the client pause command. For example:

```
client pause 60000 write
```

3. Third, wait for the key to expire, and then execute the randomkey command while the client pause is still in effect.

```
randomkey
```

**Expected behavior**

It is expected that the valkey-server will return an expired key. However, in reality, there is no response at all, and the CPU utilization of the process reaches 100%.

**Additional information**

This bug will directly render valkey-server unavailable.

li-benson:
When the `client pause write` is set and all the keys in the server
are expired keys, executing the `randomkey` command will lead to an
infinite loop.

The reason is that expired keys are not deleted in this case. Limit
the number of tries and return an expired key after the max number
tries in this case.

Closes #1848.

### DIFF
--- a/tests/unit/pause.tcl
+++ b/tests/unit/pause.tcl
@@ -417,11 +417,11 @@ start_server {tags {"pause network"}} {
         r flushall
 
         # then set a key with expire time
-        r set key value ex 3
+        r set key value px 3
 
         # set pause-write model and wait key expired
         r client pause 10000 write
-        after 5000
+        after 5
 
         wait_for_condition 50 100 {
             [r randomkey] == "key"
@@ -429,7 +429,7 @@ start_server {tags {"pause network"}} {
             fail "execute randomkey failed, caused by the infinite loop"
         }
 
-        after 6000
+        r client unpause
         assert_equal [r randomkey] {}
 
     }
--- a/tests/unit/pause.tcl
+++ b/tests/unit/pause.tcl
@@ -412,6 +412,17 @@ start_server {tags {"pause network"}} {
         } {bar2}
     }
 
+    test "Test the randomkey command will not cause the server to get into an infinite loop during the client pause write" {
+        r set key value ex 2
+        r client pause 5000 write
+        after 3000
+        wait_for_condition 30 100 {
+            [r randomkey] == {}
+        } else {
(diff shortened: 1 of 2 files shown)


enjoy-binbin (commented):
this make sense to me, it's a known issue with or without PAUSE_ACTION_EXPIRE, adding a max retry seems ok to me. @valkey-io/core-team thoughts?

tests/unit/pause.tcl
```diff
@@ -412,6 +412,28 @@ start_server {tags {"pause network"}} {
...
+    test "Test the randomkey command will not cause the server to get into an infinite loop during the client pause write" {
+        # first, clear the database to avoid interference from existing keys on the test results 
+        r flushall
+
+        # then set a key with expire time
+        r set key value ex 3
+
+        # set pause-write model and wait key expired
+        r client pause 10000 write
+        after 5000
```
zuiderkwast: Let's use shorter millisecond times to make the test faster.

```suggestion
        # then set a key with expire time
        r set key value px 3

        # set pause-write model and wait key expired
        r client pause 10000 write
        after 5
```

tests/unit/pause.tcl
```diff
@@ -412,6 +412,28 @@ start_server {tags {"pause network"}} {
...
+
+        wait_for_condition 50 100 {
+            [r randomkey] == "key"
+        } else {
+            fail "execute randomkey failed, caused by the infinite loop"
+        }
+
+        after 6000
```
zuiderkwast: We don't need to wait.

```suggestion
        r client unpause
```

zuiderkwast (commented):
Ah, because expire-if-needed just returns `KEY_EXPIRED` without deleting the key in this case. Good fix!

zuiderkwast (approved):

PR #1856, opened 2025-03-18
Rewrite the correct databases value when dbnum is clamped to 1 in cluster mode

### LINKED ISSUE #1853: [BUG] We should not rewrite configuration values when we clamp them.
Today in cluster mode, we clamp the value of the database parameter to a value of 1, regardless of what the user configured value is. When the user than calls `CONFIG REWRITE`, it will write out the value of 1 to the config file. Moving forward, we should still write out the configured value.

enjoy-binbin:
In cb1fff3cb6a944d1f9cc67f8aa8b1d92648bbade, we will override the databases
config at startup to 1 in cluster mode. When the user than calls CONFIG REWRITE,
it will write out the value of 1 to the config file. We should still write
out the configured value in this case.

This closes #1853.


tests/unit/cluster/introspection.tcl
```diff
@@ -0,0 +1,8 @@
+start_cluster 1 0 {tags {external:skip cluster}} {
+    test "CONFIG REWRITE databases" {
+        assert_equal 16 [lindex [r config get databases] 1]
```
enjoy-binbin, line 3: noted that with this change, config get will always get the config_dbnum

zuiderkwast (approved):

hpatro (approved):
Should this be treated as a breaking change?

hpatro:
With the upcoming PR of multi databases https://github.com/valkey-io/valkey/pull/1671/, `databases` config wouldn't be clamped down further for cluster mode as we're introducing `cluster-databases`. So, do we wait for 9.0 for this to get resolved with #1671 or do we take this change in right away ?

enjoy-binbin:
i am guessing @madolson brought it up with the intention of including it in 8.1? OR just a tmp fix that lives in unstable unit we get multdb support in 9.0

enjoy-binbin:
> Should this be treated as a breaking change?

not sure about it, i'd probably prefer it to be a bugfix.

zuiderkwast:
I think we can release it in 9.0 together with cluster-databases. They are related. No reason to confuse users about this in 8.1 release notes.

zuiderkwast:
If I understood correctly, we decided [in the meeting two days ago](https://github.com/valkey-io/valkey/wiki/Weekly-meeting#core-team-sync-2025-03-17--1400-utc-all-zones) that we will do this change. We can already mark it as *major-decision-approved*?

madolson (approved):
Cool.

madolson:
So, afaik this is now resolved with the introduced of multi cluster DB PR.

PR #1859, opened 2025-03-18
Hyperloglog ARM NEON SIMD optimization

xbasel:
Add ARM NEON optimization for HyperLogLog

- Implement two NEON optmized functions for converting between raw and
  dense representations in HyperLogLog:
  1. hllMergeDenseNEON
  2. hllDenseCompressNEON
  These functions process 16 registers in each iteration.

- Utilize existing SIMD test in hyperloglog.tcl (previously added for
  AVX2 optimization) to validate NEON implementation

Test:
``` valkey-benchmark -n 1000000 --dbnum  9  -p 21111 PFMERGE z hll1{t} hll2{t}```

```
+-------------------+-----------+-----------+---------------+
|      Metric       |  Before   |   After   | Improvement % |
+-------------------+-----------+-----------+---------------+
| Throughput (k rps)|    7.42   |   76.98   |    937.47%    |
+-------------------+-----------+-----------+---------------+
| Latency (msec)    |           |           |               |
|   avg             |   6.686   |   0.595   |     91.10%    |
|   min             |   0.520   |   0.152   |     70.77%    |
|   p50             |   7.799   |   0.599   |     92.32%    |
|   p95             |   8.039   |   0.767   |     90.46%    |
|   p99             |   8.111   |   0.807   |     90.05%    |
|   max             |   9.263   |   1.463   |     84.21%    |
+-------------------+-----------+-----------+---------------+
```

Hardware:
```
CPU: Graviton 3
Architecture:           aarch64
  CPU op-mode(s):       32-bit, 64-bit
  Byte Order:           Little Endian
CPU(s):                 64
  On-line CPU(s) list:  0-63
NUMA:
  NUMA node(s):         1
  NUMA node0 CPU(s):    0-63
Memory: 256 GB
```

Command stats:
Before:
```
cmdstat_pfmerge:calls=1000002,usec=126327984,**usec_per_call=126.33**,rejected_calls=0,failed_calls=0

```
After:
```
cmdstat_pfmerge:calls=1000002,usec=8588205,**usec_per_call=8.59**,rejected_calls=0,failed_calls=0
```
Improved by **~14.7x.**

Functional testing command:
```
./runtest --single unit/hyperloglog --only "PFMERGE results with simd"  --loops 10000  --fastfail
```
The SIMD test randomizes input and comapres scalar vs simd results.

### CODE REVIEW

File: src/hyperloglog.c [hunk 1]
```diff
@@ -1772,16 +1958,22 @@ void pfdebugCommand(client *c) {
         if (!strcasecmp(c->argv[2]->ptr, "on")) {
 #ifdef HAVE_AVX2
             simd_enabled = 1;
+#endif
+#ifdef __ARM_NEON
+            simd_enabled = 1;
 #endif
```
zuiderkwast: #if defined(HAVE_AVX) || defined(__ARM_NEON)

or we can consider some macro like `HLL_HAVE_SIMD` to avoid this repetition, also to avoid defining the variable twice.

zuiderkwast (commented):
10 times faster is a pretty good improvement. :)

I didn't read the NEON code carefully because I'm not familiar with it. Is the logic basically the same as the one for AVX2?

xbasel:
> 10 times faster is a pretty good improvement. :)
> 
> I didn't read the NEON code carefully because I'm not familiar with it. Is the logic basically the same as the one for AVX2?

It is similar. NEON vectors are 128 bit, AVX2 is 256 bit. The padding and lookup is a bit different in AVX2.
The execution time  of pfmerge ~**14.7x faster**. The end to end is ~10x faster.

src/hyperloglog.c
```diff
@@ -1193,6 +1204,95 @@ void hllMergeDenseAVX2(uint8_t *reg_raw, const uint8_t *reg_dense) {
...
+    uint8x16_t mask3 = vreinterpretq_u8_u32(vdupq_n_u32(0x0003f000)); // Bits 12-17
+    uint8x16_t mask4 = vreinterpretq_u8_u32(vdupq_n_u32(0x00fc0000)); // Bits 18-23
+
+    for (int i = 0; i < HLL_REGISTERS / 16 - 1; ++i) {
+        /* Load 16 bytes from dense registers but only the first 12 bytes are processed because they contain
+         * 16 registers, which is copied into 16 bytes raw registers.
+         * The last 4 bytes are ignored because (1) they do not form a complete number of registers, and do not fit
+         * in the 16 bytes. The unprocessed 4 bytes are processed in the next iteration.
+         */
+        uint8x16_t r = vld1q_u8(dense_ptr);
```
xbasel, line 1245: May crash on ARMv7-A if the address isn't 16-byte aligned. 

TODO

src/hyperloglog.c
```diff
@@ -227,6 +231,19 @@ static int simd_enabled = 1;
 #define HLL_USE_AVX2 0
 #endif
 
+#ifdef __ARM_NEON
+static int simd_enabled = 1;
+#define HLL_USE_NEON (simd_enabled)
+#else
+#define HLL_USE_NEON 0
+#endif
+
```
zhulipeng: Do we need add a runtime check here when server startup ?

xbasel: I've limited this to aarch64, which is guaranteed to have neon.

src/hyperloglog.c
zuiderkwast: ARMv7-A is 32-bit? It is solved now when you use `#if defined(__aarch64__)`?

xbasel: Yes. 
https://developer.arm.com/documentation/102376/0200/Alignment-and-endianness/Alignment

Arm7:
https://developer.arm.com/documentation/ddi0406/c/Application-Level-Architecture/Application-Level-Memory-Model/Alignment-support/Unaligned-data-access-restrictions-in-ARMv7-and-ARMv6

src/hyperloglog.c
```diff
@@ -227,6 +231,19 @@ static int simd_enabled = 1;
...
 #endif
 
+#ifdef __aarch64__
+static int simd_enabled = 1;
+#define HLL_USE_NEON (simd_enabled)
+#else
+#define HLL_USE_NEON 0
+#endif
+
+#if defined(HAVE_X86_SIMD) || defined(__aarch64__)
```
zuiderkwast: HAVE_X86_SIMD is always defined, either to 1 or to 0.

```suggestion
#if HAVE_X86_SIMD || defined(__aarch64__)
```

src/hyperloglog.c
zuiderkwast: So I'll resolve this comment.

src/hyperloglog.c
```diff
@@ -227,6 +231,19 @@ static int simd_enabled = 1;
...
 
+#ifdef __aarch64__
+static int simd_enabled = 1;
+#define HLL_USE_NEON (simd_enabled)
+#else
+#define HLL_USE_NEON 0
+#endif
+
+#if HAVE_X86_SIMD || defined(__aarch64__)
+#define SIMD_SUPPORTED 1
```
zuiderkwast: Let's put the declaration of `static int simd_enabled = 1;` here to cover all kinds of SIMD, instead of declaring the variable in two places (one for AVX2 and one for NEON).

src/hyperloglog.c
zhulipeng: It is guaranteed to have NEON support in the compile environment. Is it possible to run AArch64 binaries on a non-neon platform?

zuiderkwast: IIUC, all AArch64 include NEON. AArch64 was new in ARMv8. https://en.wikipedia.org/wiki/ARM_architecture_family#Armv8. (Even older ARM 32-bit have NEON but we don't care about those for SIMD.)

zuiderkwast (approved):
LGTM.

I didn't review the actual SIMD code very carefully, but we have a test to compare the results with/without SIMD so I think it's safe.

I'll wait for @lipzhu's approval before merge.

xbasel:
> LGTM.
> 
> I didn't review the actual SIMD code very carefully, but we have a test to compare the results with/without SIMD so I think it's safe.
> 
> I'll wait for @lipzhu's approval before merge.

The code is actually already being tested with  SIMD vs Scalar in `tests/unit/hyperloglog.tcl`. See `PFMERGE results with simd` test.
All you need to do is to run the test on ARM.

```
➜  valkey git:(hll_neon) ✗ ./runtest --single unit/hyperloglog --only "PFMERGE results with simd"
..
[ok]: PFMERGE results with simd (457 ms)
..
[1/1 done]: unit/hyperloglog (1 seconds)

                   The End

  1 seconds - unit/hyperloglog

\o/ All tests passed without errors!

➜  valkey git:(hll_neon) ✗ cat /proc/cpuinfo
processor	: 0
BogoMIPS	: 2100.00
Features	: fp asimd evtstrm aes pmull sha1 sha2 crc32 atomics fphp asimdhp cpuid asimdrdm jscvt fcma lrcpc dcpop sha3 sm3 sm4 asimddp sha512 sve asimdfhm dit uscat ilrcpc flagm ssbs paca pacg dcpodp svei8mm svebf16 i8mm bf16 dgh rng
CPU implementer	: 0x41
CPU architecture: 8
```

This is an ARMv8-A 64-bit CPU (and supports NEON).

zuiderkwast:
> The code is actually already being tested with SIMD vs Scalar in `tests/unit/hyperloglog.tcl`. See `PFMERGE results with simd` test.
> All you need to do is to run the test on ARM.

Yeah, I know. That's why I said "we have a test to compare the results with/without SIMD so I think it's safe".

zhulipeng (approved):

src/hyperloglog.c
zhulipeng: Yes, I'm wondering if we compiled on the AArch64 platform and released the ARM version binary, but the binary is running on an older ARM platform that doesn't support the NEON architecture (e.g., ARMv6). If this edge case isn't a concern, I'm fine with skipping the runtime checker.

xbasel: > Yes, I'm wondering if we compiled on the AArch64 platform and released the ARM version binary, but the binary is running on an older ARM platform that doesn't support the NEON architecture (e.g., ARMv6). If this edge case isn't a concern, I'm fine with skipping the runtime checker.

https://developer.arm.com/documentation/102474/0100/Fundamentals-of-Armv8-Neon-technology
```
AArch64 is the name used to describe the 64-bit Execution state of the Armv8-A architecture. In AArch64 state, the processor executes the A64 instruction set, which contains Neon instructions (also referred to as SIMD instructions). GNU and Linux documentation sometimes refers to AArch64 as ARM64.
```


Older architectures do not support AArch64 execution state, so the binary won't run.

PR #1869, opened 2025-03-20
[cluster] Add node id to log statement for closing link on first message as lightweight

hpatro:
I was playing around and was sending lightweight message before the cluster link had stabilized and the link was getting disconnected. I'm unsure how it happens in the current scenario. However, if it happens at all it will be good to capture the node name from which the message had originated.


enjoy-binbin (approved):

zuiderkwast (approved):

hwware:
Do we want to merge this PR before Valkey 8.1.0 GA?

zuiderkwast:
> Do we want to merge this PR before Valkey 8.1.0 GA?

Sounds good. No need to mention in release notes I think.

sarthakaggarwal97 (approved):

PR #1870, opened 2025-03-21
Add certificate rotation functionality to ValKey

ovaiskhansc:
This PR adds the capability of certificate rotation to ValKey by reloading the certificates from disk if it changes. The feature is disabled out of box and is only enabled if `tls-rotation` config is true.

### DIFF
--- a/runtest-rotation
+++ b/runtest-rotation
@@ -1,0 +1,31 @@
+#!/bin/sh
+TCL_VERSIONS="8.5 8.6"
+TCLSH=""
+
+export ASAN_OPTIONS=allocator_may_return_null=1  $ASAN_OPTIONS
+
+for VERSION in $TCL_VERSIONS; do
+	TCL=`which tclsh$VERSION 2>/dev/null` && TCLSH=$TCL
+done
+
+if [ -z $TCLSH ]
+then
+    echo "You need tcl 8.5 or newer in order to run the ValKey test"
+    exit 1
+fi
+if [ ! -r tests/tls ] || [ ! -r tests/tls_1 ] || [ ! -r tests/tls_2 ];
+then
+    echo "Generating neccessary certificates for TLS rotation testing."
+    rm -rf tests/tls tests/tls_1 tests/tls_2
+
+    utils/gen-test-certs.sh
+    mv tests/tls tests/tls_1
+    utils/gen-test-certs.sh
+    mv tests/tls tests/tls_2
+    utils/gen-test-certs.sh
+fi
+$TCLSH tests/test_helper.tcl \
+--single tls-rotation/tls-rotation \
+--tls \
+--config io-threads 3 \
+"${@}"
--- a/runtest-rotation
+++ b/runtest-rotation
@@ -15,7 +15,7 @@ then
 fi
 if [ ! -r tests/tls ] || [ ! -r tests/tls_1 ] || [ ! -r tests/tls_2 ];
 then
-    echo "Generating neccessary certificates for TLS rotation testing."
+    echo "Generating necessary certificates for TLS rotation testing."
     rm -rf tests/tls tests/tls_1 tests/tls_2
 
     utils/gen-test-certs.sh
--- a/src/server.c
+++ b/src/server.c
@@ -1670,10 +1670,10 @@ long long serverCron(struct aeEventLoop *eventLoop, long long id, void *clientDa
         migrateCloseTimedoutSockets();
     }
 
-    /* Reload the TLS cert if neccessary. This effectively rotates the 
+    /* Reload the TLS cert if necessary. This effectively rotates the
      * cert if a change has been made on disk, but the ValKey server hasn't
      * been notified. */
-    run_with_period(1000){
+    run_with_period(1000) {
         tlsReload();
     }
 
--- a/src/server.c
+++ b/src/server.c
@@ -1670,6 +1670,13 @@ long long serverCron(struct aeEventLoop *eventLoop, long long id, void *clientDa
         migrateCloseTimedoutSockets();
     }
 
+    /* Reload the TLS cert if neccessary. This effectively rotates the 
+     * cert if a change has been made on disk, but the ValKey server hasn't
+     * been notified. */
+    run_with_period(1000){
+        tlsReload();
+    }
+
     /* Resize tracking keys table if needed. This is also done at every
      * command execution, but we want to be sure that if the last command
      * executed changes the value via CONFIG SET, the server will perform
--- a/tests/test_helper.tcl
+++ b/tests/test_helper.tcl
@@ -249,6 +249,26 @@ proc valkey_client_by_addr {host port} {
     return $client
 }
 
+proc valkey_client_tls {args} {
+    set level 0
+    if {[llength $args] > 0 && [string is integer [lindex $args 0]]} {
+        set level [lindex $args 0]
+        set args [lrange $args 1 end]
+    }
+
+    set tlsoptions ""
+    if {[llength $args] > 0 && ![string is integer [lindex $args 0]]} {
+        set tlsoptions [lrange $args 0 end]
+    }
+
+    # create client that takes in custom tls options
+    set client [redis [srv $level "host"] [srv $level "port"] 0 $::tls $tlsoptions]
+
+    # # select the right db and read the response (OK)
+    $client select 9
+    return $client
+}
+
 # Provide easy access to INFO properties. Same semantic as "proc r".
 proc s {args} {
     set level 0
--- a/tests/tls-rotation/tls-rotation.tcl
+++ b/tests/tls-rotation/tls-rotation.tcl
@@ -1,0 +1,109 @@
+start_server {tags {"tls-rotation"} overrides {tls-rotation yes}} {
+    if {$::tls} {
+        package require tls
+
+        test {TLS: Create temporary location for certificate rotation} {
+            set rootdir [tmpdir tlscerts]
+
+            file copy "tests/tls" $rootdir
+            file copy "tests/tls_1" $rootdir
+            file copy "tests/tls_2" $rootdir
+
+            set serverdir [format "%s/$rootdir/tls" [pwd]]
+            set clientdir1 [format "%s/$rootdir/tls_1" [pwd]]
+            set clientdir2 [format "%s/$rootdir/tls_2" [pwd]]
+        }
+
+        test {TLS: Update server config to point to temporary location } {
+            r config set tls-key-file "$serverdir/server.key"
+            r config set tls-cert-file "$serverdir/server.crt"
+            r config set tls-ca-cert-file "$serverdir/ca.crt"
+        }
+
+        test {TLS: Connect client to server} {
+            set r2 [valkey_client_tls -keyfile "$serverdir/client.key" -certfile "$serverdir/client.crt" -require 1 -cafile "$serverdir/ca.crt"]
+            $r2 set x 50
+            assert_equal {50} [$r2 get x]
+        }
+
+        test {TLS: Rotate all TLS certificates} {
+            file delete -force -- $serverdir 
+            file copy $clientdir1 $serverdir
+            after 1000
+        }
+
+        test {TLS: Already connected clients do not lose connection post certificate rotation} {
+            $r2 incrby x 50
+            assert_equal {100} [$r2 get x]
+        }
+
+        test {TLS: Clients with outdated credentials cannot connect} {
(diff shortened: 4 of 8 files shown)


ovaiskhansc:
@madolson @zuiderkwast @uriyage - would be good to get your feedback on this pull request.

sungming2:
IMO, 
1. Relying solely on the modification time of certificate files to detect changes can be unreliable, as mtime can be changed without actual content change. I think another approach to detect change is to compare the fingerprints of the certificates which could minimize unnecessary reloads.
2. Before loading a new certificate, it might be necessary to validate the certificate to prevent potential security issues such as notBefore and notAfter fields to ensure that the certificate is currently valid and hasn't expired.
3. Is it possible to use monitoring filesystem events like inotify instead of reloading the certificate every second?

ovaiskhansc:
Thanks @sungming2 for the comments.

ovaiskhansc:
Thanks @sungming2 for the comments. Here are my initial thoughts:
1. We can surely do some extra validations to handle the cases where the last modified time changes without actual certificate changes, but it will also come with some overhead and how many times would the time just bump as opposed to the actual certificate changing as well
2. My understanding is that the `tlsConfigure` method is already doing some validations and is used currently to do the same on server startup as well as on cert rotation via `CONFIG SET` command. Why do we want to add more validations elsewhere?
3. I can look at the feasibility of doing that in a platform agnostic manner, any recommendations on the library that we can look at?

sungming2:
Thanks for the follow up. 
Just to clarify, when I mentioned validating the certificate, I wasn’t referring only to expiration checks. I think we also need to consider how this change impacts cluster mode.
If certificate rotation simply means replacing the old certificate with renewed one signed by the same CA, then that's fine. But if we think about other scenarios like when a new certificate is issued by a different CA, it can break the cluster. Valkey frequently needs to re-establish connections between nodes due to timeouts, network issue, or node config changes. If some nodes have reloaded the new certificate which is no longer trusted by others, TLS handshakes will be failing.

> My understanding is that the tlsConfigure method is already doing some validations

I don’t think `tlsConfigure` has explicit certificate checks. It can still load expired certificate without error (It probably needs to be fixed). It throws an error when the client/peer attempts to connect it since handshake fails

e.g.,
```
29132:M 25 Mar 2025 22:36:15.117 # Error accepting a client connection: error:14094415:SSL routines:ssl3_read_bytes:sslv3 alert certificate expired (addr= laddr=127.0.0.1:6379)
```

zuiderkwast:
Just some thoughts on @sungming2's comments:

> * Relying solely on the modification time of certificate files to detect changes can be unreliable, as mtime can be changed without actual content change. I think another approach to detect change is to compare the fingerprints of the certificates which could minimize unnecessary reloads.

I think there is no harm in reloading the certificate even if there is no change.

> * Before loading a new certificate, it might be necessary to validate the certificate to prevent potential security issues such as notBefore and notAfter fields to ensure that the certificate is currently valid and hasn't expired.

It is possible to configure an already-expired or not-yet-valid certificate using configuration file and CONFIG SET so I think the automatic reloading should behave the same.

Should we refuse to load expired certificates? This means CONFIG SET should fail if we try to configure an expired certificate. What about starting with a configuration file with an expired certificate, should the server refuse to start in this case? It would be a behavior change... Maybe we can do it as a separate PR?

> * Is it possible to use monitoring filesystem events like inotify instead of reloading the certificate every second?

We can do it as a future improvement. Inotify is not portable so we still need some fallback to periodic checks. Is every second too often to check, then maybe check less often? 10 sec, 60 sec?

> Just to clarify, when I mentioned validating the certificate, I wasn’t referring only to expiration checks. I think we also need to consider how this change impacts cluster mode.
If certificate rotation simply means replacing the old certificate with renewed one signed by the same CA, then that's fine. But if we think about other scenarios like when a new certificate is issued by a different CA, it can break the cluster. Valkey frequently needs to re-establish connections between nodes due to timeouts, network issue, or node config changes. If some nodes have reloaded the new certificate which is no longer trusted by others, TLS handshakes will be failing.

When changing the certificate chains and CA certificates, users need to have an overlap during which both old and new certificates are valid. It's possible to combine multiple CA certificates in one file if I'm not mistaken.

This too is no different to updating the certificates manually using CONFIG SET.

sungming2:
> Should we refuse to load expired certificates? This means CONFIG SET should fail if we try to configure an expired certificate. What about starting with a configuration file with an expired certificate, should the server refuse to start in this case? It would be a behavior change... Maybe we can do it as a separate PR?

Yes I think we can do. I don't see strong reason to load expired certificate

ovaiskhansc:
> I don't see strong reason to load expired certificate

I don't disagree, but just pointing out that that issue exists regardless of this change. We can however change the behavior of all ways you can configure the certificate to have that check and not club it with this change. In that change, we will have to answer questions around the behavior of server startup or config set commands as pointed out by @zuiderkwast 

> When changing the certificate chains and CA certificates, users need to have an overlap during which both old and new certificates are valid. It's possible to combine multiple CA certificates in one file if I'm not mistaken.

Yeah, there has to be significant amount of overlap in cert expiry. If it helps, we can put this information on the release notes of a wiki as well.

zuiderkwast:
So let's defer the cert expiry validation to another PR. It's somewhat orthogonal to this change.

> > When changing the certificate chains and CA certificates, users need to have an overlap during which both old and new certificates are valid. It's possible to combine multiple CA certificates in one file if I'm not mistaken.
> 
> Yeah, there has to be significant amount of overlap in cert expiry. If it helps, we can put this information on the release notes of a wiki as well.

We have documentation: https://valkey.io/topics/encryption/ . This page could need some improvements. It's markdown files in a git repo so it's easy to update it.

runtest-rotation
```diff
@@ -0,0 +1,31 @@
...
+if [ ! -r tests/tls ] || [ ! -r tests/tls_1 ] || [ ! -r tests/tls_2 ];
+then
+    echo "Generating necessary certificates for TLS rotation testing."
+    rm -rf tests/tls tests/tls_1 tests/tls_2
+
+    utils/gen-test-certs.sh
+    mv tests/tls tests/tls_1
+    utils/gen-test-certs.sh
+    mv tests/tls tests/tls_2
+    utils/gen-test-certs.sh
```
zuiderkwast, line 25: Generating the certificates is currently a separate step from running the tests. I think we can keep it that way.

We can modify the `utils/gen-test-certs.sh` script to generate all the certificates we need for the tests including these new tests.

I don't think we should add this new runtest script on the top level to run only this one test suite. We can put these rotation tests in `tests/unit/tls.tcl` instead.

tests/test_helper.tcl
```diff
@@ -249,6 +249,26 @@ proc valkey_client_by_addr {host port} {
...
+        set tlsoptions [lrange $args 0 end]
+    }
+
+    # create client that takes in custom tls options
+    set client [redis [srv $level "host"] [srv $level "port"] 0 $::tls $tlsoptions]
+
+    # # select the right db and read the response (OK)
+    $client select 9
+    return $client
+}
```
zuiderkwast, line 270: This function is a copy of `valkey_client` with the addition that it takes optional tls options and adds then to the arguments used for creating a client?

Can we avoid the duplicated code by some refactoring?
I think we can pass tlsoptions as a single optional argument to `valkey_client`.

Duplicated code is a source of errors. I can already see some mistakes in this copy of the function: This code tries to call a function called `redis`, which should be `valkey`. The call `$client select 9` should not be done if `$::singledb` is set.

tests/tls-rotation/tls-rotation.tcl
zuiderkwast, line 1: I think we can put these tests in `tests/unit/tls.tcl`, alternatively in a new file but it should be under `tests/unit/` or `tests/integration/`. I don't see a need for a new directory just for this test suite.

valkey.conf
```diff
@@ -300,6 +300,11 @@ tcp-keepalive 300
 #
 # tls-session-cache-timeout 60
 
+# Allow the server to monitor the filesystem and rotate out TLS certificates if 
+# they change on disk, defaults to no.
+# 
+# tls-rotation no
```
zuiderkwast, line 306: I would say reload instead of rotate. It can be used for rotation but it doesn't do the rotation for you. The server only cares about changed files and reloads them.

How about `tls-auto-reload-certs`?

tests/tls-rotation/tls-rotation.tcl
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        test {TLS: Check that old client is still connected post rotation} {
+            $r2 incrby x 50
+            assert_equal {300} [$r2 get x]
+        }
+
+        test {TLS: Clients with outdated credentials cannot connect} {
+            catch {set r2 [valkey_client_tls -keyfile "$clientdir1/client.key" -certfile "$clientdir1/client.crt" -require 1 -cafile "$clientdir1/ca.crt"]} e
+            assert_no_match {*::redis::redisHandle*} $e
```
zuiderkwast, line 79: redis? It needs to be valkey?

tests/tls-rotation/tls-rotation.tcl
```diff
@@ -0,0 +1,109 @@
...
+        }
+
+        test {TLS: Check that old client is still connected} {
+            $r2 incrby x 50
+            assert_equal {400} [$r2 get x]
+        }
+
+        test {TLS: Check that client with old credentials won't connect} {
+            catch {set r2 [valkey_client_tls -keyfile "$clientdir1/client.key" -certfile "$clientdir1/client.crt" -require 1 -cafile "$clientdir2/ca.crt"]} e
+            assert_no_match {*::redis::redisHandle*} $e
```
zuiderkwast, line 100: redis?

zuiderkwast (commented):
Good work. The reloading code looks good. I just have some comments on the tests structure and the config name.

src/server.c
```diff
@@ -1670,6 +1670,13 @@ long long serverCron(struct aeEventLoop *eventLoop, long long id, void *clientDa
         migrateCloseTimedoutSockets();
     }
 
+    /* Reload the TLS cert if necessary. This effectively rotates the
+     * cert if a change has been made on disk, but the ValKey server hasn't
+     * been notified. */
+    run_with_period(1000) {
+        tlsReload();
+    }
+
```
madolson, line 1679: We currently require users to explicitly trigger the certificate rotation, why do we think the right approach is to have a cron just that is running and checking periodically? Is this possible to do with a linux job or something?

valkey.conf
ovaiskhansc: Good call - `tls-auto-reload-certs` sounds as this more closely matches the functionality

src/server.c
zuiderkwast: Checking periodically is not perfect. A better solution would be to use [inotify](https://www.man7.org/linux/man-pages/man7/inotify.7.html) to watch for file system events, but it's not portable so we'll anyway need some fallback to periodic checking for other OSes.

I guess you could already call `valkey-cli CONFIG SET` from a system cron job to set one of the TLS configs such as `tls-port` to the value it already has. This triggers a reload of certs. But you need to know one of the configs, or you need to do CONFIG GET first. There is no command to only reload them.

Some programs reload their configs when you send a SIGHUP to it. Could that be something to consider?

madolson: > But you need to know one of the configs, or you need to do CONFIG GET first. There is no command to only reload them.

Yeah, seems like doing `CONFIG GET` first is a pretty simple solution then you can orchestrate it your own way. You also get explicit error handling since the response back will be synchronous.

murphyjacob4: > I guess you could already call valkey-cli CONFIG SET from a system cron job to set one of the TLS configs such as tls-port to the value it already has

Also note that refreshing TLS configs can be expensive. Not only is it file system IO on the main thread, but it drops the SSL context which means that the SSL session cache is dropped, meaning any clients using session resumption will feel the impact with more expensive connection establishment.

When writing a sidecar process to do automated rotation, I can imagine many might start with a simple design like:

1. Publish certificates to some place for downloading by the servers
2. Some cron on the server downloads the certificates every few minutes 
3. The cron also executes CONFIG SET after redownloading (since it should be idempotent)

Maybe we can verify if the checksum of the file has changed in tlsConfigure and short-circuit if not. I think it would help in either case (whether automatically triggered by certificate refresh in serverCron, or manually triggered by CONFIG SET) to avoid accidental drop of the session cache.

madolson: Fwiw we could do the TLS config refresh in a background thread, to limit the impact on the main thread.

madolson:
Replaced with https://github.com/valkey-io/valkey/pull/3020

PR #1885, opened 2025-03-25
Fix crash in VM_GetCurrentUserName when no valid user

jeon1226:
VM_GetCurrentUserName make the binary crash when it gets not fully structured ctx, when it doesn't have a valid ctx, client, user behind it. Rather than making the binary crash, safely return NULL

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -9763,7 +9763,8 @@ ValkeyModuleString *VM_GetModuleUserACLString(ValkeyModuleUser *user) {
  * The returned string must be released with ValkeyModule_FreeString() or by
  * enabling automatic memory management. */
 ValkeyModuleString *VM_GetCurrentUserName(ValkeyModuleCtx *ctx) {
-    if (ctx != NULL || ctx->client != NULL || ctx->client->user != NULL) {
+    if (ctx == NULL || ctx->client == NULL || ctx->client->user == NULL) {
+        errno = ENOTSUP;
         return NULL;
     }
     return VM_CreateString(ctx, ctx->client->user->name, sdslen(ctx->client->user->name));
--- a/src/module.c
+++ b/src/module.c
@@ -9770,7 +9770,6 @@ ValkeyModuleString *VM_GetCurrentUserName(ValkeyModuleCtx *ctx) {
         errno = EINVAL;
         return NULL;
     }
-    errno = 0;
     return VM_CreateString(ctx, ctx->client->user->name, sdslen(ctx->client->user->name));
 }
 
--- a/src/module.c
+++ b/src/module.c
@@ -9763,6 +9763,9 @@ ValkeyModuleString *VM_GetModuleUserACLString(ValkeyModuleUser *user) {
  * The returned string must be released with ValkeyModule_FreeString() or by
  * enabling automatic memory management. */
 ValkeyModuleString *VM_GetCurrentUserName(ValkeyModuleCtx *ctx) {
+    if (ctx != NULL || ctx->client != NULL || ctx->client->user != NULL) {
+        return NULL;
+    }
     return VM_CreateString(ctx, ctx->client->user->name, sdslen(ctx->client->user->name));
 }
 
--- a/src/module.c
+++ b/src/module.c
@@ -9763,8 +9763,8 @@ ValkeyModuleString *VM_GetModuleUserACLString(ValkeyModuleUser *user) {
  * The returned string must be released with ValkeyModule_FreeString() or by
  * enabling automatic memory management. */
 ValkeyModuleString *VM_GetCurrentUserName(ValkeyModuleCtx *ctx) {
-    if (ctx == NULL || ctx->client == NULL || ctx->client->user == NULL) {
-        errno = ENOTSUP;
+    if (ctx == NULL || ctx->client == NULL || ctx->client->user == NULL || ctx->client->user->name == NULL) {
+        errno = EINVAL;
         return NULL;
     }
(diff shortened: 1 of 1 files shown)


src/module.c
```diff
@@ -9763,6 +9763,9 @@ ValkeyModuleString *VM_GetModuleUserACLString(ValkeyModuleUser *user) {
  * The returned string must be released with ValkeyModule_FreeString() or by
  * enabling automatic memory management. */
 ValkeyModuleString *VM_GetCurrentUserName(ValkeyModuleCtx *ctx) {
+    if (ctx != NULL || ctx->client != NULL || ctx->client->user != NULL) {
+        return NULL;
```
yairgott, line 9771: set errno to ENOTSUP

zuiderkwast (approved):

src/module.c
xbasel: Not familiar with this code, but it seems to me that having an uninitialized context is essentially passing invalid parameters? If that's the case, it's not that the operation is unsupported, but rather that the input is invalid. Maybe EINVAL or ENODATA would be more appropriate than ENOTSUP.

src/module.c
```diff
@@ -9763,6 +9763,10 @@ ValkeyModuleString *VM_GetModuleUserACLString(ValkeyModuleUser *user) {
  * The returned string must be released with ValkeyModule_FreeString() or by
  * enabling automatic memory management. */
 ValkeyModuleString *VM_GetCurrentUserName(ValkeyModuleCtx *ctx) {
+    if (ctx == NULL || ctx->client == NULL || ctx->client->user == NULL) {
```
enjoy-binbin: let's also check the name in case.
```suggestion
    if (ctx == NULL || ctx->client == NULL || ctx->client->user == NULL || ctx->client->user->name == NULL) {
```

src/module.c
enjoy-binbin: If we need to set it, then we should also update the comment (the doc) to mention it.

zuiderkwast: EINVAL sounds good to me. A context without a user is an invalid parameter to this function.

ranshid: I am also in favor of using EINVAL. AFAIK ENODATA  in most known uses indicate that necessary output cannot be provided rather than "input" related.

ranshid:
@jeon1226 if we want this to be part of 8.1 release, we need to make sure this is completed ASAP. I think there are only some minor comments - can we followup on them and finalize this?

src/module.c
jeon1226: > If we need to set it, then we should also update the comment (the doc) to mention it.

Do you mean valkey-doc or the comment here?

src/module.c
```diff
@@ -9763,6 +9763,10 @@ ValkeyModuleString *VM_GetModuleUserACLString(ValkeyModuleUser *user) {
  * The returned string must be released with ValkeyModule_FreeString() or by
  * enabling automatic memory management. */
```
zuiderkwast: These comments are important because they are used for the generated API documentation.

```suggestion
 * enabling automatic memory management.
 *
 * If the context is not associated with a client connection, NULL is returned
 * and errno is set to EINVAL. */
```

src/module.c
```diff
@@ -9763,6 +9763,10 @@ ValkeyModuleString *VM_GetModuleUserACLString(ValkeyModuleUser *user) {
...
 ValkeyModuleString *VM_GetCurrentUserName(ValkeyModuleCtx *ctx) {
+    if (ctx == NULL || ctx->client == NULL || ctx->client->user == NULL || ctx->client->user->name == NULL) {
+        errno = EINVAL;
+        return NULL;
+    }
```
yairgott, line 9772: If no error, errno should be set to 0.

yairgott (changes requested):

jeon1226:
Hi the code is ready I think - how do I merge?

src/module.c
```diff
@@ -9761,8 +9761,16 @@ ValkeyModuleString *VM_GetModuleUserACLString(ValkeyModuleUser *user) {
...
+ * enabling automatic memory management.
+ *
+ * If the context is not associated with a client connection, NULL is returned
+ * and errno is set to EINVAL. */
 ValkeyModuleString *VM_GetCurrentUserName(ValkeyModuleCtx *ctx) {
+    if (ctx == NULL || ctx->client == NULL || ctx->client->user == NULL || ctx->client->user->name == NULL) {
+        errno = EINVAL;
+        return NULL;
+    }
+    errno = 0;
```
zuiderkwast: Don't set errno to 0. No functions ever do that.

The man page errno(3) mentions this for library functions:

> The value of errno is never set to zero by any system call or library function.

```suggestion
```

src/module.c
zuiderkwast: The errno(3) man page is very clear on this: "The  value of errno is never set to zero by any system call or library function."

```
errno(3)                   Library Functions Manual                   errno(3)

NAME
       errno - number of last error

LIBRARY
       Standard C library (libc, -lc)

SYNOPSIS
       #include <errno.h>

DESCRIPTION
       The  <errno.h> header file defines the integer variable errno, which is
       set by system calls and some library functions in the event of an error
       to indicate what went wrong.

   errno
       The value in errno is significant only when the  return  value  of  the
       call  indicated  an  error (i.e., -1 from most system calls; -1 or NULL
       from most library functions); a function that succeeds  is  allowed  to
       change  errno.   The  value of errno is never set to zero by any system
       call or library function.
```

PR #1888, opened 2025-03-26
Change the usec_per_call operation from float to long long division

bluayer:
## Affected command by this change
INFO commandstats (- only `usec_per_call` metric)

### Changed results example
`cmdstat_set:calls=2,usec=36,usec_per_call=18,rejected_calls=1,failed_calls=0
cmdstat_get:calls=3,usec=13,usec_per_call=4,rejected_calls=1,failed_calls=0
cmdstat_info:calls=6,usec=444,usec_per_call=74,rejected_calls=0,failed_calls=0
cmdstat_command|docs:calls=2,usec=3951,usec_per_call=1975,rejected_calls=0,failed_calls=0`

## Context

The `usec_per_call` is a simple AVG metric that divides the execution time by the number of calls. However, it was performing the calculation by converting long long to float. For two reasons, usec_per_call does not require the high accuracy of a float level:

1. There are not many reasons to know the average value in units smaller than microseconds.
2. It currently displays the value of only two digits after the decimal point for accuracy
3. The client(or monitoring tool) can also calculate because there are `calls` and `usec` metrics. 

Therefore, I propose changing the implementation to a simple long long division operation without the floating-point conversion and computation.

This change will benefit users who are using various commands in valkey and using high-resolution metrics in commandstats.

## Alternatives
The engine already has calls and usec metrics, so clients or users can calculate `usec_per_call` accurately enough. Also, engine needs to perform a type conversion and then divide to get this metric for every commands and subcommands. 

So I finally want to propose removing this metric, but that would be a breaking change and impact many clients. Therefore, I would first propose removing the unnecessary float conversions.

## Simple performance testing (Updated at Apr 15)

**Env**
- EC2 r6i.2xlarge
- random long long number division. 10000 iterations & 10 runs
- With type casting(float), Without type casting(long long - Below, it's written "int division" )
- Using rdtsc for getting cpu cycles

Screenshot 2025-04-14 at 1 04 37 PM

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -5490,10 +5490,10 @@ sds genValkeyInfoStringCommandStats(sds info, hashtable *commands) {
         char *tmpsafe;
         if (c->calls || c->failed_calls || c->rejected_calls) {
             info = sdscatprintf(info,
-                                "cmdstat_%s:calls=%lld,usec=%lld,usec_per_call=%.2f"
+                                "cmdstat_%s:calls=%lld,usec=%lld,usec_per_call=%lld"
                                 ",rejected_calls=%lld,failed_calls=%lld\r\n",
                                 getSafeInfoString(c->fullname, sdslen(c->fullname), &tmpsafe), c->calls,
-                                c->microseconds, (c->calls == 0) ? 0 : ((float)c->microseconds / c->calls),
+                                c->microseconds, (c->calls == 0) ? 0 : (c->microseconds / c->calls),
                                 c->rejected_calls, c->failed_calls);
             if (tmpsafe != NULL) zfree(tmpsafe);
         }


src/server.c
```diff
@@ -5490,10 +5490,10 @@ sds genValkeyInfoStringCommandStats(sds info, hashtable *commands) {
         char *tmpsafe;
         if (c->calls || c->failed_calls || c->rejected_calls) {
             info = sdscatprintf(info,
-                                "cmdstat_%s:calls=%lld,usec=%lld,usec_per_call=%.2f"
+                                "cmdstat_%s:calls=%lld,usec=%lld,usec_per_call=%lld"
```
zuiderkwast, line 5493: This will change the format slightly from `1.00` to `1`.

I think it's right to remove the sub-microsecond part. We only measure the execution time in microseconds anyway, so this is already not exact on sub-microsecond level.

We should probably label it as a potentially breaking change though, just in some case some tool is parsing this and expect the period to be there.

zuiderkwast (commented):
Please mention in the PR title or top comment which command is affected by this change.

A possible workaround: It's possible call INFO with a specific section (e.g. `INFO server`), to avoid computing this info section.

xbasel:
I have some concerns about this change.

It breaks existing behavior, commands that take <1µs will now just show 0, which looks off and we lose useful precision, especially when profiling fast commands.  

I still don’t quite understand why this change is needed, could you provide an example where the current float-based calculation causes issues? 


**UPDATE: Saw that the precision in µs, this means the reported usec_per_call is not accurate. So we either fix the accuracy or we remove these decimal digits (which is what this change is doing)...**

src/server.c
xbasel: I think sub-microsecond latency is useful - why not modify `getMonotonicUs` to return nanoseconds instead (getMonotonicNs)? both  __rdtsc/__cntvct  can provide sub-microsecond precision...

bluayer:
@xbasel I'm sorry to late response.

The given avg metric(`usec_per_call`) can be calculated precisely enough at the client using the existing `calls` and `usec` metrics that we already know.

Would you please explain your concerns in more detail? I'm curious to know if the nano-second level of precision at the avg metric is truly necessary.

Update : Add simple performance test results.

zuiderkwast:
Question regarding the benchmark: Is it 25 microseconds for 10000 divisions? We save 5 microseconds per 10000 divisions? That's not much per division.

bluayer:
Yes. But If users are using high-resolution metrics, the impact could be greater. 
And according to the previous comment(at this closed PR, #1620 ), I understand that we are continuously improving performance in INFO and it's recommended to calculate derivable metrics on the client.

However, it seems difficult to remove the metric, so I suggest a small improvement first.

xbasel:
I still think keeping sub-microsecond precision in `usec_per_call` is useful for usability and diagnostics, even if clients can recompute it. Removing it may degrade precision for fast commands. I get that current timings are at microsecond granularity, but that can be fixed.

Also unclear if float division here causes real-world perf issues; if it does, SIMD acceleration could be considered instead of dropping precision.

Just to clarify, this isn't about nanosecond precision - it's sub-microsecond. For example, a command taking 300ns would still register as 0µs if we drop the decimal. That level of detail can be important when analyzing fast commands like GET, SET, INCR. This can be useful when comparing pre- and post-upgrade scenarios.

PR #1889, opened 2025-03-26
Redact protocol error log when hide-user-data-from-log enabled

YueTang-Vanessa:
In this code logic: https://github.com/valkey-io/valkey/blob/unstable/src/networking.c#L2767-L2773, `c->querybuf + c->qb_pos` may also include user data.  
Update the log message when config `hide-user-data-from-log` is enabled.

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -2760,17 +2760,8 @@ void processInlineBuffer(client *c) {
 static void setProtocolError(const char *errstr, client *c) {
     if (server.verbosity <= LL_VERBOSE || c->flag.primary) {
         sds client = catClientInfoString(sdsempty(), c, server.hide_user_data_from_log);
-
-        /* Sample some protocol to given an idea about what was inside. */
         char buf[256];
         buf[0] = '\0';
-        if (c->querybuf && sdslen(c->querybuf) - c->qb_pos < PROTO_DUMP_LEN) {
-            snprintf(buf, sizeof(buf), "Query buffer during protocol error: '%s'", c->querybuf + c->qb_pos);
-        } else if (c->querybuf) {
-            snprintf(buf, sizeof(buf), "Query buffer during protocol error: '%.*s' (... more %zu bytes ...) '%.*s'",
-                     PROTO_DUMP_LEN / 2, c->querybuf + c->qb_pos, sdslen(c->querybuf) - c->qb_pos - PROTO_DUMP_LEN,
-                     PROTO_DUMP_LEN / 2, c->querybuf + sdslen(c->querybuf) - PROTO_DUMP_LEN / 2);
-        }
 
         /* Remove non printable chars. */
         char *p = buf;
@@ -2781,7 +2772,8 @@ static void setProtocolError(const char *errstr, client *c) {
 
         /* Log all the client and protocol info. */
         int loglevel = (c->flag.primary) ? LL_WARNING : LL_VERBOSE;
-        serverLog(loglevel, "Protocol error (%s) from client: %s. %s", errstr, client, buf);
+        serverLog(loglevel, "Protocol error (%s) from client: %s. Query buffer: %s", errstr, client,
+                  server.hide_user_data_from_log ? "*redacted*" : buf);
         sdsfree(client);
     }
     c->flag.close_after_reply = 1;
--- a/src/networking.c
+++ b/src/networking.c
@@ -2760,20 +2760,31 @@ void processInlineBuffer(client *c) {
 static void setProtocolError(const char *errstr, client *c) {
     if (server.verbosity <= LL_VERBOSE || c->flag.primary) {
         sds client = catClientInfoString(sdsempty(), c, server.hide_user_data_from_log);
+
+        /* Sample some protocol to given an idea about what was inside. */
         char buf[256];
         buf[0] = '\0';
+        if (server.hide_user_data_from_log) {
+            snprintf(buf, sizeof(buf), "*redacted*");
+        } else {
+            if (c->querybuf && sdslen(c->querybuf) - c->qb_pos < PROTO_DUMP_LEN) {
+                snprintf(buf, sizeof(buf), "'%s'", c->querybuf + c->qb_pos);
+            } else if (c->querybuf) {
+                snprintf(buf, sizeof(buf), "'%.*s' (... more %zu bytes ...) '%.*s'",
+                         PROTO_DUMP_LEN / 2, c->querybuf + c->qb_pos, sdslen(c->querybuf) - c->qb_pos - PROTO_DUMP_LEN,
+                         PROTO_DUMP_LEN / 2, c->querybuf + sdslen(c->querybuf) - PROTO_DUMP_LEN / 2);
+            }
 
-        /* Remove non printable chars. */
-        char *p = buf;
-        while (*p != '\0') {
-            if (!isprint(*p)) *p = '.';
-            p++;
+            /* Remove non printable chars. */
+            char *p = buf;
+            while (*p != '\0') {
+                if (!isprint(*p)) *p = '.';
+                p++;
+            }
         }
-
         /* Log all the client and protocol info. */
         int loglevel = (c->flag.primary) ? LL_WARNING : LL_VERBOSE;
-        serverLog(loglevel, "Protocol error (%s) from client: %s. Query buffer: %s", errstr, client,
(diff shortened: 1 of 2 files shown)


src/networking.c
```diff
@@ -2765,11 +2765,15 @@ static void setProtocolError(const char *errstr, client *c) {
         char buf[256];
         buf[0] = '\0';
         if (c->querybuf && sdslen(c->querybuf) - c->qb_pos < PROTO_DUMP_LEN) {
-            snprintf(buf, sizeof(buf), "Query buffer during protocol error: '%s'", c->querybuf + c->qb_pos);
+            snprintf(buf, sizeof(buf), "Query buffer during protocol error: '%s'",
+                     server.hide_user_data_from_log ? "[hide_user_data_from_log]Redacted" : (c->querybuf + c->qb_pos));
```
hpatro: I think we can just print `*redacted*` like we do in other places.

```suggestion
            snprintf(buf, sizeof(buf), "Query buffer during protocol error: '%s'",
                     server.hide_user_data_from_log ? "*redacted*" : (c->querybuf + c->qb_pos));
```

src/networking.c
```diff
@@ -2765,11 +2765,15 @@ static void setProtocolError(const char *errstr, client *c) {
...
         } else if (c->querybuf) {
-            snprintf(buf, sizeof(buf), "Query buffer during protocol error: '%.*s' (... more %zu bytes ...) '%.*s'",
-                     PROTO_DUMP_LEN / 2, c->querybuf + c->qb_pos, sdslen(c->querybuf) - c->qb_pos - PROTO_DUMP_LEN,
-                     PROTO_DUMP_LEN / 2, c->querybuf + sdslen(c->querybuf) - PROTO_DUMP_LEN / 2);
+            snprintf(buf, sizeof(buf), server.hide_user_data_from_log ?
+                     ("Query buffer during protocol error: [hide_user_data_from_log]Redacted %zu bytes",
+                      sdslen(c->querybuf) - c->qb_pos) :
+                     ("Query buffer during protocol error: '%.*s' (... more %zu bytes ...) '%.*s'",
+                      PROTO_DUMP_LEN / 2, c->querybuf + c->qb_pos, sdslen(c->querybuf) - c->qb_pos - PROTO_DUMP_LEN,
+                      PROTO_DUMP_LEN / 2, c->querybuf + sdslen(c->querybuf) - PROTO_DUMP_LEN / 2));
```
hpatro: This is still printing out the last few bytes I believe.

src/networking.c
```diff
@@ -2765,11 +2765,15 @@ static void setProtocolError(const char *errstr, client *c) {
...
         }
 
         /* Remove non printable chars. */
```
QuChen88: This part can also print out customer sensitive data.

src/networking.c
```diff
@@ -2765,11 +2765,15 @@ static void setProtocolError(const char *errstr, client *c) {
         char buf[256];
         buf[0] = '\0';
         if (c->querybuf && sdslen(c->querybuf) - c->qb_pos < PROTO_DUMP_LEN) {
```
QuChen88: Why not just do the check for `server.hide_user_data_from_log` before getting into all the other logic below? i.e. something like 

```
+     if (server.hide_user_data_from_log) {
+          snprintf(buf, sizeof(buf), "Query buffer during protocol error: <redacted>");
+         goto log_client_and_proto;
+     } 
     ....
     
+ log_client_and_proto:
    /* Log all the client and protocol info. */
    .... 
```

zuiderkwast:
When exactly is this triggered? Only when the client sends some incorrectly encoded RESP?

hpatro (commented):
Could you add a test case and assert `*redacted*` on it ?.

src/networking.c
```diff
@@ -2760,6 +2760,12 @@ void processInlineBuffer(client *c) {
 static void setProtocolError(const char *errstr, client *c) {
     if (server.verbosity <= LL_VERBOSE || c->flag.primary) {
         sds client = catClientInfoString(sdsempty(), c, server.hide_user_data_from_log);
+        int loglevel = (c->flag.primary) ? LL_WARNING : LL_VERBOSE;
+
+        if (server.hide_user_data_from_log) {
+            printf("Query buffer during protocol error: *redacted*");
```
zuiderkwast: Printf seems wrong. Did you mean snprintf?

```suggestion
            snprintf(buf, sizeof(buf), "Query buffer during protocol error: *redacted*");
```

YueTang-Vanessa: Since using goto after this line, there has no buf declared/initialed, I don't think we should use `snprintf`.

hpatro: Unresolving the conversation. `printf` isn't useful. Do you want to add a statement to logs? Try `serverLog` then.

YueTang-Vanessa: Got it. Will use the serverLog instead then

QuChen88:
Looks good to me

QuChen88 (approved):

QuChen88:
@zuiderkwast Basically any sort of protocol parsing error from the primary to the replica can lead to this condition to get triggered and the query buffer containing potential user data/commands would get logged. This can happen for various reasons including data corruption over the wire etc. i.e. we saw this in a couple of our redis clusters in the past. 

```
3524:S 26 Mar 2025 17:49:50.755 # WARNING: Receiving inline protocol from master, master stream corruption? Closing the master connection and discarding the cached master.
3524:S 26 Mar 2025 17:49:50.755 # Protocol error (Master using the inline protocol. Desync?) from client: ... Query buffer during protocol error: '293..$4..hdel..$31.. <key-name>...
```

The chances of this happening is pretty low, but still non zero.

src/networking.c
zuiderkwast: So now we're logging twice? First here, then goto and log again?

Let's log only once.

src/networking.c
```diff
@@ -2779,9 +2785,10 @@ static void setProtocolError(const char *errstr, client *c) {
             p++;
         }
 
+    log_client_and_proto_info:
         /* Log all the client and protocol info. */
-        int loglevel = (c->flag.primary) ? LL_WARNING : LL_VERBOSE;
-        serverLog(loglevel, "Protocol error (%s) from client: %s. %s", errstr, client, buf);
+        serverLog(loglevel, "Protocol error (%s) from client: %s. %s", errstr, client,
+                  server.hide_user_data_from_log ? "*redacted*" : buf);
```
zuiderkwast: When hide-user-data is disabled, the logged messages look like these, depending on buffer length:

* `"Protocol error (errstr) from client: (client). Query buffer during protocol error: (buf)"`

When hide-user-data is enabled, this code logs

* `"Protocol error (errstr) from client: (client). *redacted*"`
 
The message doesn't say what is redacted. It doesn't say that it's the query buffer.

Another thing that bothers me a little is that "protocol error" appears twice in the logged message. I mean in the old messages. This wasn't well formulated.

To align the two variants and also get rid of the duplicated wording, I think we should remove "Query buffer during protocol error" from the `snprintf` code above (where `buf` is constructed) and add "query buffer" here instead, i.e.

```suggestion
        serverLog(loglevel, "Protocol error (%s) from client: %s. Query buffer: %s", errstr, client,
                  server.hide_user_data_from_log ? "*redacted*" : buf);
```

src/networking.c
zuiderkwast: And if we don't log it here, there's no reason to move the line `int loglevel = (c->flag.primary) ? LL_WARNING : LL_VERBOSE;`. It can stay where it was.

YueTang-Vanessa: Removed snprint as suggested in the latest version

src/unit/test_networking.c
```diff
@@ -62,6 +62,29 @@ static fakeConnection *connCreateFake(void) {
     return conn;
 }
 
+static int find_pattern_in_log(char *str) {
```
madolson: I feel like this should be an integration test, in TCL, than in a unit test.

hpatro: Yeah, I had the same thought when I requested for it to address Viktor's question.

src/networking.c
YueTang-Vanessa: Removed the snprint in latest version as suggested

ranshid (approved):

zuiderkwast (approved):
LGTM

PR #1891, opened 2025-03-27
Remove TLSCONN_DEBUG dead code in tls.c

### LINKED ISSUE #1887: Undefined variable 'fd'
https://github.com/valkey-io/valkey/blob/d56a1f79b451d89904b421dcc94a519b271c16f4/src/tls.c#L700

WelongZuo:
This code is a dead code, there is also a undefined fd error in it.
Obviously this will not compile if this kind of debugging is enabled.

This seems to be broken ever since 5a477946065bcf05b335ededd6b794e82882ab73
which was done in 2019. It's dead code that we never test and never use.

Fixes #1887

### DIFF
--- a/src/tls.c
+++ b/src/tls.c
@@ -415,12 +415,6 @@ error:
     return C_ERR;
 }
 
-#ifdef TLS_DEBUGGING
-#define TLSCONN_DEBUG(fmt, ...) serverLog(LL_DEBUG, "TLSCONN: " fmt, __VA_ARGS__)
-#else
-#define TLSCONN_DEBUG(fmt, ...)
-#endif
-
 static ConnectionType CT_TLS;
 
 /* Normal socket connections have a simple events/handler correlation.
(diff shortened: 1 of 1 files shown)


enjoy-binbin:
Can you explain how this is triggered? Does it only happen in debug mode?

WelongZuo:
> Can you explain how this is triggered? Does it only happen in debug mode?

WelongZuo:
> Can you explain how this is triggered? Does it only happen in debug mode?

I haven't trigger this bug in production enviroment. Just find it when i learn the code.
From the define in tls.c maybe only when build with marco "TLS_DEBUGGING" can run it.
I think it's a very small problem that we can fix or ignore.

src/tls.c
```diff
@@ -697,7 +697,7 @@ static void TLSAccept(void *_conn) {
 static void tlsHandleEvent(tls_connection *conn, int mask) {
     int ret, conn_error;
 
-    TLSCONN_DEBUG("tlsEventHandler(): fd=%d, state=%d, mask=%d, r=%d, w=%d, flags=%d", fd, conn->c.state, mask,
+    TLSCONN_DEBUG("tlsEventHandler(): fd=%d, state=%d, mask=%d, r=%d, w=%d, flags=%d", conn->c.fd, conn->c.state, mask,
```
zuiderkwast: Interesting. Obviously this will not compile if this kind of debugging is enabled.

This seems to be broken ever since 5a47794606 which was done in 2019.

I think it's better that we simply delete the TLSCONN_DEBUG macro. It's dead code that we never test and never use, so we will not know if it's correct or broken.

WelongZuo: yeah, i agree with delete this line and the define of TLSCONN_DEBUG macro

zuiderkwast: OK, do you want to do this change in this PR?

WelongZuo: yeah, i have already done

enjoy-binbin (approved):

PR #1894, opened 2025-03-27
fix merge error introduced in #1186

JimB123:
Merge error introduced in #1186 .

Prior to the merge, the time value passed to modules was a `monotime`.  The merge reverts it to a wall clock time (on the sending side only).


zuiderkwast (approved):
Ack!

I obviousy reverted https://github.com/valkey-io/valkey/pull/1388 in the merge conflicts while working on #1186.

Issue #1884, opened 2025-03-25
[NEW] Enhance valkey-benchmark with Multi-Language Support and Advanced Testing Capabilities


asafpamzn:
**Current Limitations**

The existing valkey-benchmark tool, while useful for many scenarios, has several limitations that hinder its effectiveness in simulating real-world workloads:

1. Limited flexibility in test scenarios: Unable to test with varying value sizes, which is common in real-life workloads. Restricted to uniform key distribution, lacking support for more realistic distributions (e.g., hot keys).
Constrained cluster configuration:

2. Cannot handle topology updates or complex read-from-replica techniques.
3. Lack of end-to-end latency measurements: Does not provide actual client-side latency experience.

**Proposed Solution**

To address these limitations and provide a more comprehensive benchmarking solution,I propose Multi-language support: Add separate repositories for each supported programming language, while maintaining a consistent interface with the current valkey-benchmark.

**Enhanced testing capabilities:**

1. Support for variable value sizes
2. Customizable key distribution patterns
3. Improved cluster configuration options
4. More accurate end-to-end latency measurements

**Implementation**

I have created working code with most of the functionality of valkey-benchmark in multiple programming languages:

https://github.com/asafpamzn/valkey-benchmark-node
https://github.com/asafpamzn/valkey-benchmark-java
https://github.com/asafpamzn/valkey-benchmark-go
https://github.com/asafpamzn/valkey-benchmark-python

These implementations maintain the same interface as the original valkey-benchmark while addressing the aforementioned limitations. And also have some extra features, such as qps control, sequential pattern, test-duration and more.

The ask is to move these repos to the valkey repo and continue to develop them with the valkey community.

**Suggested Next Steps:**
1. Review one of the provided implementations (all are similar, so reviewing one is sufficient).
2. Discuss the integration of these new repositories into the main valkey project.
Collaborate with the valkey community to further develop and maintain these multi-language benchmarking tools.
Benefits

asafpamzn:
Some relevant issues:
1. [valkey-benchmark support for SET IFEQ](https://github.com/valkey-io/valkey/issues/1828)  - we support all custom commands.
2. [Benchmark CLI tool rewrite](https://github.com/valkey-io/valkey/issues/900)  -  Address some of the issues discussed in this issue, but the idea is not to replace the tool but more of add client experience for users, mainly for users that test latency.

EranAs:
Another capability suggestion:
Support stats per slot (or more practically - per shard).

[...]

Issue #368, opened 2024-04-24
A strategy for performance testing


madolson:
Opening this issue, as we no longer have a benchmarking framework. Performance is an essential part of Valkey, and we need to make it easier to evaluate if something is degrading (or improving) performance. The previous framework was not open-source, as it was maintained by filipe from Redis. The specifications are still here: https://github.com/redis/redis-benchmarks-specification, although they were never really reviewed so I'm not convinced we want to reference them, I think we should start fresh.

My vision, is that I would like us to implement some performance tests that run on test runners running on dedicated hardware (ideally bare-metal, but that might get expensive) that we can trigger to run when a specific tag get's added to a PR and during daily. The daily runs can be used for generating historical graphs.

I would like to at least see the following sets of tests:
1. Single instance running simple core tests for each data structure (SET, LPUSH/POP, SADD, HSET, XADD, ZADD).
2. A cluster mode instance running basic string operations. 

We should run them each for at least like a minute or so, ideally in parallel.

### Next steps
1. Get consensus about the priority and the strategy.
3. Followup with a working group to determine the specific tests we would like to automate
4. Get the hardware and setup the infrastructure. (@madolson and @stockholmux are on AWS hardware acquisition, but would appreciate others as well)
6. Profit!

### Future work

I would also like to extend to automatically generate perf + flamegraphs of the above results as well, and have them always available on the website. That gives folks a way to see where time is being spent and maybe investigate optimizations.

madolson:
@valkey-io/core-team Would like thoughts on the above proposal, and implicitly would appreciate a vote.

stockholmux:
re: hardware 

I guess one precursor question: What hardware/architectures is Valkey planning on targeting?

PingXie:
I wasn't aware that there were performance benchmarking tools but I love this idea so adding my vote explicitly here

madolson:
> I guess one precursor question: What hardware/architectures is Valkey planning on targeting?

Ideally one pair of arm hosts and one pair of x86 hosts. Something like an m7i and an m7g is probably "broadly sufficient". If GCP would like to donate some hardware, we could run it on their infra as well. :)

lipzhu:
Hi @madolson, really happy to see this issue, we care the performance:) I have some questions:
1. If the community decide to accept a patch, is the performance report a "nice to have" or "must"?
2. Will the performance runner integrate into pull request checks tool?
[...]

PR #1898, opened 2025-03-28
Support custom workflow benchmark

artikell:
This PR implements a custom benchmark method. The ability to generate workload through Lua. Using Lua to generate requests will enable more capabilities:
- Quickly support new commands
- Customizable key distribution patterns, such as https://github.com/valkey-io/valkey/issues/1884
- Support for variable value sizes
- ... ...

It can be considered as a continuation of [this issue](https://github.com/valkey-io/valkey/issues/368).

The first step is to write a workload script(workflow.lua)：
```
-- `request_id` represents the unique identifier of the request
return "LPUSH " .. tostring(request_id) .. " " .. tostring(request_id)
```

Execute through the following command:
```
./src/valkey-benchmark --workload ./src/workload.lua
```

Tip：
- workload mode cannot be combined with other commands

Todo：
- [x] Support generating requests based on return redisReply
- [ ] Support workload benchmark in cluster mode
- [ ] Support multi thread workload benchmark
- [ ] Add workloads for common scenarios

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -526,7 +526,7 @@ $(ENGINE_CLI_NAME): $(ENGINE_CLI_OBJ)
 
 # valkey-benchmark
 $(ENGINE_BENCHMARK_NAME): $(ENGINE_BENCHMARK_OBJ)
-	$(SERVER_LD) -o $@ $^ ../deps/hiredis/libhiredis.a ../deps/hdr_histogram/libhdrhistogram.a $(FINAL_LIBS) $(TLS_CLIENT_LIBS)
+	$(SERVER_LD) -o $@ $^ ../deps/hiredis/libhiredis.a ../deps/lua/src/liblua.a ../deps/hdr_histogram/libhdrhistogram.a $(FINAL_LIBS) $(TLS_CLIENT_LIBS)
 
 DEP = $(ENGINE_SERVER_OBJ:%.o=%.d) $(ENGINE_CLI_OBJ:%.o=%.d) $(ENGINE_BENCHMARK_OBJ:%.o=%.d)
 -include $(DEP)
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -149,6 +149,7 @@ static struct config {
     pthread_mutex_t is_updating_slots_mutex;
     int resp3; /* use RESP3 */
     char *workflow;
+    _Atomic int max_requests_id;
     _Atomic int requests_num;
     lua_State *lua;
 } config;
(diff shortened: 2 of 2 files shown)


zuiderkwast:
Cool idea. It's very customizable, but maybe overkill?

For key distribution patterns, it seems more practical to use a normal distribution and add a regular parameter to specify the desired variance or std deviation. If it's specified, then normal distribution is enabled, otherwise the legacy uniform distribution is used.

Key sizes distribution and cluster slot distribution can use a similar distributions and parameters.

We should think about performance too. Lua is quite slow. Valkey-benchmark already has a hard time generating traffic fast enough. It needs to use multiple threads to be able to keep valkey-server busy on a single thread.

artikell:
@zuiderkwast 
Based on the previous discussion, it requires some complex workflow bench (running multiple commands in parallel). From some issues, there are some customized requirements for benchmarks (customized value size, fcall benchmark).

I chose Lua because Valkey itself relies on Lua.

From a performance perspective, on the one hand, this is mainly due to the consumption of generating commands, which may not necessarily result in significant losses compared to network interactions (depending on the complexity of the workflow). On the other hand, it is not strongly dependent on Lua. In the future, you can refer to the module pattern or use `LD_PRELOAD` to replace `generateCommand` methods.

At least compared to the benchmarks of workflow implemented in different languages (with even worse performance), it's better for us to unify them.

arthurkiller:
Interesting idea! I have noticed that the benchmark tool currently lacks scalability, and using Lua is a great approach

asafpamzn:
@arthurkiller , cool idea.
I think that it is a complementary to https://github.com/valkey-io/valkey/issues/1884  which can also be used to create custom commands using code in the language to their choice.

https://github.com/asafpamzn/valkey-benchmark-node
https://github.com/asafpamzn/valkey-benchmark-java
https://github.com/asafpamzn/valkey-benchmark-go
https://github.com/asafpamzn/valkey-benchmark-python

I think that both should work, and some users will prefer the above and others will prefer to use 
```
./src/valkey-benchmark --workflow ./src/workflow.lua
```

zuiderkwast:
> Based on the previous discussion, it requires some complex workflow bench (running multiple commands in parallel). From some issues, there are some customized requirements for benchmarks (customized value size, fcall benchmark).

So is the main purpose that users can simulate their own application's workload to design their own benchmarks that are more realistic for their workload?

> I chose Lua because Valkey itself relies on Lua.
>
> From a performance perspective, on the one hand, this is mainly due to the consumption of generating commands, which may not necessarily result in significant losses compared to network interactions (depending on the complexity of the workflow).

Yeah, I get it and I think it's a good idea. Currently I need 3-4 threads in valkey-benchmark to keep one singe-threaded valkey-server instance busy (run near 100% CPU) locally with pipelining enabled. You can always use more threads though. How much more CPU does it take to to generate the same load with Lua? If it's not more than twice as much, then I think it's still fine. You can always run valkey-benchmark with more threads.

PR #1899, opened 2025-03-29
Update ACL SETUSER command help message

lucasyonge:
When I learn ACL SETUSER command from valkey.io, I find the helper message is out-of-date,
I think the 'attribute' should be updated to 'rule' to align with the link https://valkey.io/commands/acl-setuser/


enjoy-binbin (approved):

hpatro (changes requested):
I think it's apt as is. The specified attribute can be a rule / password.

zuiderkwast (approved):

zuiderkwast:
@hpatro The web page documentation says "rule" in the syntax and explanation, e.g. the heading [ACL rules](https://valkey.io/commands/acl-setuser/#acl-rules). Do you think the docs are wrong?

hpatro:
> @hpatro The web page documentation says "rule" in the syntax and explanation, e.g. the heading [ACL rules](https://valkey.io/commands/acl-setuser/#acl-rules). Do you think the docs are wrong?

I think passwords and certain other attributes like `reset`, `on`, `off` aren't rules. So, yeah, I think docs should be updated. WDYT @zuiderkwast ?

zuiderkwast:
@hpatro I think someone has thought about this in the past. There is an explanation in the docs:

> Valkey ACL rules are split into two categories: rules that define command permissions or *command rules*, and rules that define the user state or *user management rules*.

After this follows the two lists of all "rules".

This seems to be our official terminology. Maybe it's not perfect but if we change it, it can be even more confusing.

hpatro:
> @hpatro I think someone has thought about this in the past. There is an explanation in the docs:
> 
> > Valkey ACL rules are split into two categories: rules that define command permissions or _command rules_, and rules that define the user state or _user management rules_.
> 
> After this follows the two lists of all "rules".
> 
> This seems to be our official terminology. Maybe it's not perfect but if we change it, it can be even more confusing.

Ok, then let's stick with the same wording across this repo and website.

hpatro (approved):

PR #1901, opened 2025-03-31
Reuse RDB file as AOF preamble on replica after disk-based full sync

RayaCoo:
## Summary

When a replica finishes a **disk-based** full sync and both `appendonly yes` and
`aof-use-rdb-preamble yes` are enabled, this PR reuses the received `dump.rdb`
as the new AOF base file, instead of triggering `BGREWRITEAOF`.

This avoids generating an almost identical base snapshot and removes redundant CPU/IO work.

### Before
1. Replica receives RDB via disk-based full sync
2. Replica loads RDB into memory
3. Replica may delete synced RDB (`rdb-del-sync-files`)
4. `BGREWRITEAOF` runs and generates a new base snapshot
5. New snapshot becomes AOF base

### After
1. Replica receives RDB via disk-based full sync
2. Replica loads RDB into memory
3. Synced RDB is renamed to the new AOF base file
4. A new incremental AOF is created
5. No `BGREWRITEAOF` is needed

### Tests
- [x] Disk-based sync + preamble on: RDB reused as AOF base
- [x] New writes after sync: incr AOF works correctly
- [x] Replica restart: reused AOF loads correctly
- [x] Disk-based sync + preamble off: fallback to `BGREWRITEAOF`
- [x] Diskless sync + preamble on: fallback to `BGREWRITEAOF`
- [x] Diskless sync with stale local `dump.rdb`: stale file is not reused

### DIFF
--- a/src/aof.c
+++ b/src/aof.c
@@ -1097,7 +1097,7 @@ int restartAOFWithSyncRdb(void) {
 
     serverLog(LL_NOTICE, "Reused RDB file from primary sync as AOF base file: %s", new_base_filename);
     ret = C_OK;
-    goto done;
+    return ret;
 
 cleanup:
     if (rdbfile_renamed) {
@@ -1127,8 +1127,6 @@ cleanup:
     }
     if (temp_am) aofManifestFree(temp_am);
     if (new_base_filepath) sdsfree(new_base_filepath);
-
-done:
     return ret;
 }
 
--- a/src/aof.c
+++ b/src/aof.c
@@ -1005,6 +1005,133 @@ int startAppendOnly(void) {
     return C_OK;
 }
 
+/* Try to restart AOF after replica full sync by adopting `server.rdb_filename`
+ * as the new BASE file (RDB preamble mode), avoiding a redundant AOFRW.
+ * Returns C_OK on success; on C_ERR caller should fallback to
+ * restartAOFAfterSYNC(). */
+int restartAOFWithSyncRdb(void) {
+    serverAssert(server.aof_state == AOF_OFF);
+
+    int ret = C_ERR;
+    int newfd = -1, rdbfile_renamed = 0;
+    sds new_base_filename = NULL;
+    sds new_base_filepath = NULL;
+    sds new_incr_filename = NULL;
+    sds new_incr_filepath = NULL;
+    aofManifest *temp_am = NULL;
+
+    if (dirCreateIfMissing(server.aof_dirname) == -1) {
+        serverLog(LL_WARNING, "Can't open or create append-only dir %s: %s", server.aof_dirname, strerror(errno));
+        goto cleanup;
+    }
+
+    serverAssert(server.aof_manifest != NULL);
+    temp_am = aofManifestDup(server.aof_manifest);
+
+    new_base_filename = getNewBaseFileNameAndMarkPreAsHistory(temp_am, server.aof_use_rdb_preamble);
+    serverAssert(new_base_filename != NULL);
+    new_base_filepath = makePath(server.aof_dirname, new_base_filename);
+
+    if (rename(server.rdb_filename, new_base_filepath) == -1) {
+        serverLog(LL_WARNING, "Error trying to rename the RDB file %s into %s: %s", server.rdb_filename,
+                  new_base_filepath, strerror(errno));
+        goto cleanup;
+    }
+    rdbfile_renamed = 1;
+
+    /* Mark existing incr AOF files as history BEFORE creating the new one,
+     * so the new incr entry is not inadvertently moved to history. */
+    markRewrittenIncrAofAsHistory(temp_am);
+
+    new_incr_filename = getNewIncrAofName(temp_am);
+    new_incr_filepath = makePath(server.aof_dirname, new_incr_filename);
+    newfd = open(new_incr_filepath, O_WRONLY | O_TRUNC | O_CREAT, 0666);
+    if (newfd == -1) {
+        serverLog(LL_WARNING, "Can't open the append-only file %s: %s", new_incr_filename, strerror(errno));
+        goto cleanup;
+    }
+
+    if (persistAofManifest(temp_am) == C_ERR) {
+        goto cleanup;
+    }
+
+    aofManifestFreeAndUpdate(temp_am);
+    temp_am = NULL;
+
+    aofDelHistoryFiles();
+
+    sdsfree(new_base_filepath);
+    new_base_filepath = NULL;
+    sdsfree(new_incr_filepath);
+    new_incr_filepath = NULL;
+
+    /* Drain pending fsync jobs from the previous AOF before publishing the new
+     * fsynced offset to avoid races on fsynced_reploff_pending. */
+    bioDrainWorker(BIO_AOF_FSYNC);
+    atomic_store_explicit(&server.fsynced_reploff_pending, server.primary_repl_offset, memory_order_relaxed);
+    server.fsynced_reploff =
+        atomic_load_explicit(&server.fsynced_reploff_pending, memory_order_relaxed);
+
+    int aof_bio_fsync_status = atomic_load_explicit(&server.aof_bio_fsync_status, memory_order_relaxed);
+    if (aof_bio_fsync_status == C_ERR) {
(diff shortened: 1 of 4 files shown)


arthurkiller:
relying on slow failure-prone components like disks is not a good choice for a high-speed kv engine. For example, Redis 7.0 has been using diskless replication by default since then, which is exactly why they went with this approach

RayaCoo:
> relying on slow failure-prone components like disks is not a good choice for a high-speed kv engine. For example, Redis 7.0 has been using diskless replication by default since then, which is exactly why they went with this approach

I understand your point about diskless replication. However, this PR only optimizes a specific configuration where disk-based replication is used with aof-use-rdb-preamble enabled. In this scenario, we're currently deleting an RDB file only to recreate an identical one immediately after. This PR simply eliminates this redundancy without changing any default behaviors.

arthurkiller:
> > relying on slow failure-prone components like disks is not a good choice for a high-speed kv engine. For example, Redis 7.0 has been using diskless replication by default since then, which is exactly why they went with this approach
> 
> I understand your point about diskless replication. However, this PR only optimizes a specific configuration where disk-based replication is used with aof-use-rdb-preamble enabled. In this scenario, we're currently deleting an RDB file only to recreate an identical one immediately after. This PR simply eliminates this redundancy without changing any default behaviors.

thx for your kind reply. I have checked the code and get what this pr want

src/server.h
```diff
@@ -2984,6 +2984,7 @@ int rewriteAppendOnlyFileBackground(void);
 int loadAppendOnlyFiles(aofManifest *am);
 void stopAppendOnly(void);
 int startAppendOnly(void);
+int tryRestartAOFAfterSYNCWithRdb(void);
```
soloestoy: this function name is a bit ambiguous, need a more accurate one like `aofRestartWithSyncFile`

src/aof.c
```diff
@@ -1005,6 +1005,121 @@ int startAppendOnly(void) {
...
+                             "the primary. This replica has persistence "
+                             "disabled");
+        if (rdbfile_renamed) {
+            bg_unlink(new_base_filepath);
+        } else {
+            bg_unlink(server.rdb_filename);
+        }
+    }
+    if (temp_am) aofManifestFree(temp_am);
+    if (new_base_filename) sdsfree(new_base_filename);
```
soloestoy: the 2 file names are just pointers of manifest, don't need to be freed.

and free the 2 file paths after where they are used would be better.

soloestoy (commented):
generally LGTM, just need some minor changes.

soloestoy:
reuse the sync file as the base AOF file can save lots of resource usage, @valkey-io/core-team PTAL.

src/server.h
RayCao99: is `restartAOFWithSyncFile` ok here? similar with original funtion called `restartAOFAfterSYNC`.

src/aof.c
RayCao99: Yeah, The file names don't need to be freed, I will fix that. 
The `base_file_path` isn't freed immediately as it's needed during cleanup, so it's freed just before the safe steps.

src/aof.c
```diff
@@ -1005,6 +1005,120 @@ int startAppendOnly(void) {
...
+                  new_base_filepath, strerror(errno));
+        goto cleanup;
+    }
+    rdbfile_renamed = 1;
+
+    /* Create a new incr AOF file */
+    new_incr_filename = getNewIncrAofName(temp_am);
+    new_incr_filepath = makePath(server.aof_dirname, new_incr_filename);
+    newfd = open(new_incr_filepath, O_WRONLY | O_TRUNC | O_CREAT, 0644);
+    sdsfree(new_incr_filepath);
```
yangbodong22011: if newfd is `-1`, 1046L will call `sdsfree`, and then go to `cleanup`, will call `sdsfree` again, cause double free.
note: `sdsfree` will not make obj to NULL.

RayCao99: Thanks for the reviews, Will fix by move the `sdsfree(new_incr_filepath)` to success path.

rjd15372:
With this change the RDB file that is re-used as the base snapshot of the AOF file, will have it's `aof-base` info field set to `0`. I don't see any code logic in the core that depends on that value except for a logging message, and therefore it should be OK, but what if some external tool depends on the `aof-base` field value for its own logic? Is that a problem?

zuiderkwast:
This looks like a good idea, but unfortunately it's too late now to make it to 9.0 so I'm punting it to 9.1. 

@soloestoy do you plan to follow up on this review?

madolson:
@soloestoy @RayaCoo Any update on the progress of this?

RayaCoo:
> With this change the RDB file that is re-used as the base snapshot of the AOF file, will have it's `aof-base` info field set to `0`. I don't see any code logic in the core that depends on that value except for a logging message, and therefore it should be OK, but what if some external tool depends on the `aof-base` field value for its own logic? Is that a problem?

Good point.

In current core logic, `aof-base` is informational only (used for logging during RDB load),
and correctness decisions are driven by the AOF manifest (base/incr/history), not by this AUX field.

So the reuse path is safe for core behavior. If any external tool currently depends on `aof-base`,
I agree we should treat that as a compatibility concern and either:
1) update that tool to rely on manifest metadata, or
2) add a follow-up compatibility patch.

I'd prefer to keep it simple right now. What do you think?

RayaCoo:
> @soloestoy @RayaCoo Any update on the progress of this?

Sorry for the long delay! I've rebased the PR onto the latest unstable branch and reworked the implementation to adapt to the upstream refactoring.

@soloestoy @arthurkiller could you please take another look?

src/replication.c
```diff
@@ -2375,10 +2375,18 @@ void replicaAfterLoadPrimaryRDB(connection *conn, rdbSaveInfo *rsi) {
...
+     *
+     * When disk-based sync was used and aof-use-rdb-preamble is enabled,
+     * reuse the RDB file received from the primary as the AOF base file
+     * directly, avoiding a redundant bgrewriteaof. Otherwise (diskless
+     * sync or rdb-preamble disabled), fall back to bgrewriteaof. */
+    if (server.aof_enabled) {
+        if (disk_based_sync && server.aof_use_rdb_preamble && server.aof_state == AOF_OFF) {
+            restartAOFWithSyncRdb();
+        }
+        if (server.aof_state == AOF_OFF) restartAOFAfterSYNC();
```
soloestoy:
```suggestion
        if (disk_based_sync && server.aof_use_rdb_preamble) {
            if (restartAOFWithSyncRdb() == C_ERR) {
                restartAOFAfterSYNC();
            }
        } else {
            restartAOFAfterSYNC();
        }
```

soloestoy (commented):
mostly LGTM, do you like to take a look? @zuiderkwast

soloestoy (approved):

src/aof.c
```diff
@@ -1005,6 +1005,131 @@ int startAppendOnly(void) {
...
+ * Returns C_OK on success; on C_ERR caller should fallback to
+ * restartAOFAfterSYNC(). */
+int restartAOFWithSyncRdb(void) {
+    serverAssert(server.aof_state == AOF_OFF);
+
+    int ret = C_ERR;
+    int newfd = -1, rdbfile_renamed = 0;
+    sds new_base_filename = NULL;
+    sds new_base_filepath = NULL;
+    sds new_incr_filename = NULL;
```
soloestoy, line 1019: we can use `openNewIncrAofForAppend ` to simplify the incr file generation part

RayCao99: I checked openNewIncrAofForAppend(), and it doesn't seem directly resuable here. 

1. It assumes `AOF` is already enabled and returns immediately when `server.aof_state == AOF_OFF`.
2. It does more than create the incr file: it also persists the manifest update and switches `server.aof_fd` right away, while here we need to finish the base/incr/history update first and only publish the new state at the end.

zuiderkwast:
> mostly LGTM, do you like to take a look? @zuiderkwast

Not really. :) I have looked briefly but I don't know the AOF very well. I want to merge this so it can be included in 9.1.

PR #1908, opened 2025-04-01
Bump CMake to 3.10.0 to fix build error on Ubuntu-24

zuiderkwast:
This makes it build on Ubuntu-24, which otherwise fails with this error:

CMake Error at deps/hiredis/CMakeLists.txt:1 (CMAKE_MINIMUM_REQUIRED):
  Compatibility with CMake < 3.5 has been removed from CMake.


hpatro (commented):
Don't we need to bump the engine CMakeLists as well and shouldn't we have the same across the repository? For the valkey engine it's 3.1 currently.

https://github.com/valkey-io/valkey/blob/09f9630171501d1d62536045984220cad73588ba/CMakeLists.txt#L1

zuiderkwast:
> Don't we need to bump the engine CMakeLists as well and shouldn't we have the same across the repository? For the valkey engine it's 3.1 currently.

Oh, yeah probably. I just saw other builds fail on the above mentioned error and attempted to fix that. Let's see what the build jobs reveal on this PR.

zuiderkwast:
But hey, 3.10 > 3.5 in version order so it's still supported. :)

hpatro:
> But hey, 3.10 > 3.5 in version order so it's still supported. :)

Ha ha, I misread it. Let's make this 3.10 as well then.

zuiderkwast:
> Ha ha, I misread it. Let's make this 3.10 as well then.

The one I changed is in the hiredis part under deps that is vendored from another repo. It doesn't necessarily have to be the same version as the rest of the project.

Changing it directly here is a hack IMO, but I think it's acceptable as a temporary solution because we'll replace it with libvalkey soon, our hiredis fork.

hpatro (approved):

PR #1910, opened 2025-04-02
Add node pfail and fail count to cluster info metrics

hpatro:
New fields in CLUSTER INFO:

* `cluster_nodes_pfail`
* `cluster_nodes_fail`
* `cluster_voting_nodes_pfail`
* `cluster_voting_nodes_fail`

I'm running few tests and trying to capture partially failed and completely failed count. Slot partially failed / completely failed stats exists but is more difficult to assess the node failure count with that.

New output:

```
> CLUSTER INFO
cluster_state:fail
cluster_slots_assigned:0
cluster_slots_ok:0
cluster_slots_pfail:0
cluster_slots_fail:0
cluster_nodes_pfail:1
cluster_nodes_fail:0
cluster_voting_nodes_pfail:1
cluster_voting_nodes_fail:0
cluster_known_nodes:3
cluster_size:0
cluster_current_epoch:1
cluster_my_epoch:1
cluster_stats_messages_ping_sent:2104
cluster_stats_messages_pong_sent:1906
cluster_stats_messages_meet_sent:1
cluster_stats_messages_sent:4011
cluster_stats_messages_ping_received:1906
cluster_stats_messages_pong_received:1964
cluster_stats_messages_received:3870
total_cluster_links_buffer_limit_exceeded:0
```

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -6400,7 +6400,8 @@ sds genClusterInfoString(void) {
 
     if (nodes_pfail + nodes_fail > dictSize(server.cluster->nodes)) {
         serverLog(LL_WARNING, "Aggregated count of nodes marked as PFAIL and FAIL exceeds the total count of nodes."
-                  "PFAIL nodes: %u, FAIL nodes: %u", nodes_pfail, nodes_fail);
+                              "PFAIL nodes: %u, FAIL nodes: %u",
+                  nodes_pfail, nodes_fail);
     }
 
     info = sdscatprintf(info,
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -6384,18 +6384,39 @@ sds genClusterInfoString(void) {
         }
     }
 
+    dictIterator *di = dictGetIterator(server.cluster->nodes);
+    dictEntry *de;
+    u_int nodes_pfail = 0, nodes_fail = 0;
+    while ((de = dictNext(di)) != NULL) {
+        clusterNode *node = dictGetVal(de);
+        if (nodeTimedOut(node)) {
+            nodes_pfail++;
+        }
+        if (nodeFailed(node)) {
+            nodes_fail++;
+        }
+    }
+    dictReleaseIterator(di);
+
+    if (nodes_pfail + nodes_fail > dictSize(server.cluster->nodes)) {
+        serverLog(LL_WARNING, "Aggregated count of nodes marked as PFAIL and FAIL exceeds the total count of nodes."
+                  "PFAIL nodes: %u, FAIL nodes: %u", nodes_pfail, nodes_fail);
+    }
+
     info = sdscatprintf(info,
                         "cluster_state:%s\r\n"
                         "cluster_slots_assigned:%d\r\n"
                         "cluster_slots_ok:%d\r\n"
                         "cluster_slots_pfail:%d\r\n"
                         "cluster_slots_fail:%d\r\n"
+                        "cluster_nodes_pfail:%u\r\n"
+                        "cluster_nodes_fail:%u\r\n"
                         "cluster_known_nodes:%lu\r\n"
                         "cluster_size:%d\r\n"
                         "cluster_current_epoch:%llu\r\n"
                         "cluster_my_epoch:%llu\r\n",
                         statestr[server.cluster->state], slots_assigned, slots_ok, slots_pfail, slots_fail,
-                        dictSize(server.cluster->nodes), server.cluster->size,
+                        nodes_pfail, nodes_fail, dictSize(server.cluster->nodes), server.cluster->size,
                         (unsigned long long)server.cluster->currentEpoch, (unsigned long long)nodeEpoch(myself));
 
     /* Show stats about messages sent and received. */
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -6386,7 +6386,7 @@ sds genClusterInfoString(void) {
 
     dictIterator *di = dictGetIterator(server.cluster->nodes);
     dictEntry *de;
-    u_int nodes_pfail = 0, nodes_fail = 0;
+    unsigned nodes_pfail = 0, nodes_fail = 0;
     while ((de = dictNext(di)) != NULL) {
         clusterNode *node = dictGetVal(de);
         if (nodeTimedOut(node)) {
@@ -6398,10 +6398,11 @@ sds genClusterInfoString(void) {
     }
     dictReleaseIterator(di);
 
-    if (nodes_pfail + nodes_fail > dictSize(server.cluster->nodes)) {
+    if (nodes_pfail + nodes_fail > dictSize(server.cluster->nodes) && !server.crashed) {
         serverLog(LL_WARNING, "Aggregated count of nodes marked as PFAIL and FAIL exceeds the total count of nodes."
-                              "PFAIL nodes: %u, FAIL nodes: %u",
-                  nodes_pfail, nodes_fail);
+                              "PFAIL nodes: %u, FAIL nodes: %u, total nodes: %lu",
+                  nodes_pfail, nodes_fail, dictSize(server.cluster->nodes));
+        serverAssert(0);
     }
 
     info = sdscatprintf(info,
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -6404,13 +6404,6 @@ sds genClusterInfoString(void) {
     }
     dictReleaseIterator(di);
 
-    if (nodes_pfail + nodes_fail > dictSize(server.cluster->nodes) && !server.crashed) {
-        serverLog(LL_WARNING, "Aggregated count of nodes marked as PFAIL and FAIL exceeds the total count of nodes."
-                              "PFAIL nodes: %u, FAIL nodes: %u, total nodes: %lu",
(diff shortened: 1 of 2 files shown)


src/cluster_legacy.c
```diff
@@ -6384,18 +6384,40 @@ sds genClusterInfoString(void) {
         }
     }
 
+    dictIterator *di = dictGetIterator(server.cluster->nodes);
+    dictEntry *de;
+    u_int nodes_pfail = 0, nodes_fail = 0;
```
zuiderkwast: `u_int`? I haven't seen that before. A non-standard alias of unsigned int?

```suggestion
    unsigned nodes_pfail = 0, nodes_fail = 0;
```

src/cluster_legacy.c
```diff
@@ -6384,18 +6384,40 @@ sds genClusterInfoString(void) {
...
+            nodes_fail++;
+        }
+    }
+    dictReleaseIterator(di);
+
+    if (nodes_pfail + nodes_fail > dictSize(server.cluster->nodes)) {
+        serverLog(LL_WARNING, "Aggregated count of nodes marked as PFAIL and FAIL exceeds the total count of nodes."
+                              "PFAIL nodes: %u, FAIL nodes: %u",
+                  nodes_pfail, nodes_fail);
+    }
```
zuiderkwast: This is not supposed to be possible. Can we use an assert?

tests/unit/cluster/info.tcl
```diff
@@ -64,3 +64,32 @@ start_cluster 3 0 {tags {external:skip cluster} overrides {cluster-node-timeout
         wait_for_cluster_state ok
     }
 }
+
+start_cluster 3 0 {tags {external:skip cluster} overrides {cluster-node-timeout 1000}} {
+    test "node partially failure count" {
+        pause_process [srv 0 pid]
```
zuiderkwast: The test case below depends on this `pause_process`. If we run only the second test case using `runtest --single ... --only ...`, it will fail I think, but we can fix it like this?

```suggestion
    pause_process [srv 0 pid]

    test "node partially failure count" {
```

zuiderkwast (commented):
Makes sense.

So the use case is to be able to write tests more reliably, or is there a "real" use case?

hpatro:
> Makes sense.
> 
> 
> 
> So the use case is to be able to write tests more reliably, or is there a "real" use case?

I'm trying to observe first time to node failure detection and time to mark it as complete failure. Without this data, it seems difficult to modify the algorithm and observe the change in behavior.

zuiderkwast:
Observability of failure detection. It's a great concept! ;) Yeah it can be useful for users too, not only for us.

tests/unit/cluster/info.tcl
hpatro: Nice, didn't ever use this. Seems useful. Updated the tests for easier pattern match as well. Thanks.

src/cluster_legacy.c
hpatro: Added one. Realized this can get into recursive crash, so added an additional condition of `!server.crashed`.

src/cluster_legacy.c
hpatro: Possibly LLM suggested it. My bad :|

hpatro:
The test seems flaky. Looking at it 
https://github.com/valkey-io/valkey/actions/runs/14249542177/job/39938769646

src/cluster_legacy.c
zuiderkwast: https://www.linuxfoundation.org/legal/generative-ai

tests/unit/cluster/info.tcl
zuiderkwast: Never used it? Yeah it's useful for debugging. It only works in some test suites though. Many test suites aren't written with this in mind.

tests/unit/cluster/info.tcl
```diff
@@ -66,21 +66,23 @@ start_cluster 3 0 {tags {external:skip cluster} overrides {cluster-node-timeout
...
         } else {
-            puts [R 1 CLUSTER INFO]
             puts [R 2 CLUSTER INFO]
-            fail "Node 0 never timed out"
+            fail "Node 0/1 never timed out"
         }
     }
 
+    # Enable one more primary to reach quorum about node 0 failure
+    resume_process [srv -1 pid]
```
zuiderkwast, line 86: Without this, the node went too fast to FAIL so we couldn't observe PFAIL?

src/cluster_legacy.c
```diff
@@ -6384,18 +6384,41 @@ sds genClusterInfoString(void) {
...
+        }
+    }
+    dictReleaseIterator(di);
+
+    if (nodes_pfail + nodes_fail > dictSize(server.cluster->nodes) && !server.crashed) {
+        serverLog(LL_WARNING, "Aggregated count of nodes marked as PFAIL and FAIL exceeds the total count of nodes."
+                              "PFAIL nodes: %u, FAIL nodes: %u, total nodes: %lu",
+                  nodes_pfail, nodes_fail, dictSize(server.cluster->nodes));
+        serverAssert(0);
+    }
```
zuiderkwast: Recursive crash? So it means it does assert here sometimes, or only when you changed some logic to make it happen?

I was thinking an assert instead of log, if it's not supposed to happen. If it can happen, then maybe only log the warning. The situation might be harmless so we don't need to panic. Maybe.

```suggestion
    if (!server.crashed) {
        /* Prevent recursive crash */
        serverAssert(nodes_pfail + nodes_fail <= dictSize(server.cluster->nodes));
    }
```

zuiderkwast (approved):
LGTM.

@valkey-io/core-team Please ack ( :+1: ) two new fields in CLUSTER INFO.

src/cluster_legacy.c
hpatro: I thought of writing this way, would it print the actual numbers in that case? Or it would just crash saying assertion failed?

hpatro: Alright, I tested it with a small change. I don't get the variable output captured with the suggested approach.

```c
index 5affc8d8a..6303d9788 100644
--- a/src/server.c
+++ b/src/server.c
@@ -4698,6 +4698,8 @@ void pingCommand(client *c) {
         return;
     }

+    int x = 10, y = 15;
+    serverAssert((x+y) < 20);
```

```
38881:M 03 Apr 2025 21:56:40.416 # === ASSERTION FAILED ===
38881:M 03 Apr 2025 21:56:40.416 # ==> server.c:4702 '(x+y) < 20' is not true
```

tests/unit/cluster/info.tcl
hpatro: Yes.

src/cluster_legacy.c
zuiderkwast: Yeah. So you need the numbers for debugging. Let's do a warning then and no assert? It's not really necessary to abort I think, since it may be a harmless situation.

I.e. as in your original code, WDYT?

```suggestion
    if (nodes_pfail + nodes_fail > dictSize(server.cluster->nodes)) {
        serverLog(LL_WARNING, "Aggregated count of nodes marked as PFAIL and FAIL exceeds the total count of nodes."
                              "PFAIL nodes: %u, FAIL nodes: %u, total nodes: %lu",
                  nodes_pfail, nodes_fail, dictSize(server.cluster->nodes));
    }
```

hpatro: After you requested the assertion, I find it useful to have. So, that we will take notice of it (if it ever happens). No one looks at the log otherwise. WDYT ?

zuiderkwast: OK, as you wish. :)

enjoy-binbin: why we need this? Someone did not handle pfail and fail flags correctly? this like a too-much.

hpatro: @enjoy-binbin Felt good to capture if there is anything wrong on the server side. You think it's very defensive? Shall we clean it up completely?

enjoy-binbin: yes, i think it is very defensive, i would vote to remove it.

madolson (commented):
Do we think having this information is better than just asking end users to run cluster nodes/shards and count the number of failed/pfail nodes? I'm a little worried about end users alarming on this metric, even though it includes nodes that aren't part of quorum and aren't serving any traffic.

hpatro:
> Do we think having this information is better than just asking end users to run cluster nodes/shards and count the number of failed/pfail nodes? I'm a little worried about end users alarming on this metric, even though it includes nodes that aren't part of quorum and aren't serving any traffic. 

With large cluster I would prefer not pulling cluster nodes/shards output and compute this.
Thinking more on this, I think I would also need the same stats around voting members. Maybe users can alarm on those. 😉

src/cluster_legacy.c
hpatro: Alright I don't mind removing it. No strong opinion on this.

hpatro:
I've added voting nodes pfail/fail as well. If we decouple voting nodes from data serving node (primary) within the same architecture in the future, will have to add two additional metric (primary_fail / primary_pfail).

@madolson let me know your thoughts.

src/cluster_legacy.c
hpatro: Done.

hwware (approved):

madolson (approved):

PR #1919, opened 2025-04-05
Fix minor comment typo

lucasyonge:


src/quicklist.c
```diff
@@ -1175,7 +1175,7 @@ int quicklistDelRange(quicklist *quicklist, const long start, const long count)
     return 1;
 }
 
-/* compare between a two entries */
+/* Compares between two entries */
```
PingXie: Thanks! 

```suggestion
/* Compares two entries */
```

Btw, this looks like a pretty trivial function to me so I am actually more inclined to just remove the comment.

PingXie (approved):

PR #1920, opened 2025-04-06
Add support for automatic client authentication via TLS certificate fields

### LINKED ISSUE #1866: [NEW] Take username from client certificate

Avoid managing user passwords for ACL when we already use TLS with mutual authentication.

When password rotation is required and also certificate rotation is required, it becomes complex to rotate both of them.

The server can take the username from the client's certificate, the comman name (CN=username) field. It seems to be common practice.

When this field matches an existing user, the client gets authenticated as this user as if the AUTH command had been used.

Use AUTH or HELLO and manage passords and do password rotation for each user, in addition to TLS certificates rotations.

**Additional information**

-

omanges:
This PR implements support for automatic client authentication based on a field in the client's TLS certificate.

API Changes:

* New configuration directive `tls-auth-clients-user`, values `CN` | `off`, default `off`. CN means take username from the CommonName field in the client's certificate.
* New INFO field `acl_access_denied_tls_cert` under the `Stats` section, indicating the number of failed authentications using this feature, i.e. client certificates for which no matching username was found.
* New reason "tls-cert" in the ACL log, logged when a client certificate's CommonName fails to match any existing username.

🔒 Feature Overview:
* Introduces a new configuration directive: tls-auth-clients-user.

  * Allows specifying a certificate field (only CN for now) to map to a Valkey user.

  * If a matching user is found, the client is automatically authenticated as that user upon TLS handshake.

  * Falls back to the default user if no match is found.

⚙️ Implementation Details:
* Added getCertFieldByName() utility to extract fields from peer certificates.

* Added autoAuthenticateClientFromCert() to handle automatic login logic post-handshake.

* Integrated automatic authentication into the TLSAccept function after handshake completion.

* Updated test suite (tests/integration/tls.tcl) to validate the feature.

✅ Benefits:
* Enables smoother integration with mTLS-based infrastructures.

* Reduces the need for clients to manually send AUTH commands.

* Enhances security by tightly coupling certificate identity with Valkey access control.

🔬 Example:
```
tls-auth-clients-user CN
```
This will authenticate the client as the user matching the certificate's Common Name.

Closes #1866


src/config.c
```diff
@@ -3361,6 +3361,7 @@ standardConfig static_configs[] = {
     createBoolConfig("tls-prefer-server-ciphers", NULL, MODIFIABLE_CONFIG, server.tls_ctx_config.prefer_server_ciphers, 0, NULL, applyTlsCfg),
     createBoolConfig("tls-session-caching", NULL, MODIFIABLE_CONFIG, server.tls_ctx_config.session_caching, 1, NULL, applyTlsCfg),
     createStringConfig("tls-cert-file", NULL, VOLATILE_CONFIG | MODIFIABLE_CONFIG, EMPTY_STRING_IS_NULL, server.tls_ctx_config.cert_file, NULL, NULL, applyTlsCfg),
+    createStringConfig("tls-auth-clients-field", NULL, VOLATILE_CONFIG | MODIFIABLE_CONFIG, EMPTY_STRING_IS_NULL, server.tls_ctx_config.client_auth_field, NULL, NULL, applyTlsCfg),
```
zuiderkwast: The config should probably be an enum config, so we can validate that only values that we can handle are allowed.

src/tls.c
```diff
@@ -677,12 +678,69 @@ static void updateSSLState(connection *conn_) {
...
+static int getCertFieldByName(X509 *cert, const char *field, char *out, size_t outlen) {
+    if (!cert || !field || !out) return 0;
+
+    int nid = -1;
+
+    if (!strcasecmp(field, "CN"))
+        nid = NID_commonName;
+    else if (!strcasecmp(field, "O"))
+        nid = NID_organizationName;
+    // Add more mappings here as needed
```
zuiderkwast: Using "O" makes no sense to me, just like "C" (Country) makes no sense. Fields like that are not unique. If we start supporting it, I suspect we might regret it in the future. @madolson you mentioned "O" in a comment. Was it a serious suggestion or just an example?

Out of the top-level fields, only CN (common name) makes sense to me.

src/tls.c
```diff
@@ -677,12 +678,69 @@ static void updateSSLState(connection *conn_) {
...
+    if (!conn || !SSL_is_init_finished(conn->ssl)) return;
+
+    const char *field = server.tls_ctx_config.client_auth_field;
+    if (!field) return;
+
+    serverLog(LL_NOTICE, "TLS: Config Auto Auth Field = %s", field);
+
+    X509 *cert = SSL_get_peer_certificate(conn->ssl);
+    if (!cert) {
+        serverLog(LL_WARNING, "TLS: No peer certificate found after handshake");
```
zuiderkwast: WARNING is a serious log level. It means the server is in trouble and needs an administrator's help.

Missing cert is not an error. Client certificate can be configured to be optional. I think we shouldn't log anything here, not even a notice.

src/tls.c
```diff
@@ -677,12 +678,69 @@ static void updateSSLState(connection *conn_) {
...
+        return;
+    }
+
+    char field_value[256];
+    if (!getCertFieldByName(cert, field, field_value, sizeof(field_value))) {
+        serverLog(LL_WARNING, "TLS: Failed to extract field %s from certificate", field);
```
zuiderkwast: The should probably just be a NOTICE.

src/tls.c
```diff
@@ -677,12 +678,69 @@ static void updateSSLState(connection *conn_) {
...
+
+    client *c = connGetPrivateData(&conn->c);
+    if (!c) return;
+
+    user *u = ACLGetUserByName(conn->c.cert_user, sdslen(conn->c.cert_user));
+    serverLog(LL_NOTICE, "TLS: ACL lookup user: %s", u ? u->name : "NOT FOUND");
+    if (u) {
+        c->user = u;
+        c->flag.authenticated = true;
+        serverLog(LL_NOTICE, "TLS: Auto-authenticated client as %s", u->name);
```
zuiderkwast: Logging twice here is too verbose IMO. Let's log only the second one, when it has been found and authenticated.

zuiderkwast (commented):
Nice! I didn't do a complete review. For example I didn't look at the tests at all yet.

I'm not sure if the config should allow any field. Maybe only CN makes sense ever and then the config should maybe not be selecting a field. But let's discuss it and think about what we can do.

I found the corresponding features in ETCD and Kubernetes. The both use only common name as username:

* ETCD: https://etcd.io/docs/v3.4/op-guide/authentication/#using-tls-common-name

  > As of version v3.2 if an etcd server is launched with the option --client-cert-auth=true, the field of Common Name (CN) in the client’s TLS cert will be used as an etcd user.

* Kubernetes: https://kubernetes.io/docs/reference/access-authn-authz/authentication/#x509-client-certificates

  > If a client certificate is presented and verified, the common name of the subject is used as the user name for the request. As of Kubernetes 1.4, client certificates can also indicate a user's group memberships using the certificate's organization fields.

src/tls.c
madolson: Someone asked for O once inside AWS, but I don't know if they would have been okay with just CN as well. We can always just start with "CN" and see if anyone asks for more in the future?  The enum just allows us to add more in the future.

zuiderkwast: > The enum just allows us to add more in the future.

Yes, I agree enums in general are good for extensibility, but I doubt what we want to add in the future are top-level fields. It's more likely that we want to add

* subject alt name: DNS name
* an option to fail the handshake if the client cert doesn't match an existing user

But if you were asked once to use O for username, then I'm OK with it. It's an actual ask from a user.

madolson: I followed up with the folks, they did indeed ask for "O". They sign their certificates by team, and they want the auth to happen at the team level (imagine a bunch of apps with multiple services, they want all services to have the same permission). They said "CN" would be fine, they would just need to provision more users, but that would be "OK" for them.

src/connection.h
```diff
@@ -138,6 +138,7 @@ struct connection {
     ConnectionCallbackFunc conn_handler;
     ConnectionCallbackFunc write_handler;
     ConnectionCallbackFunc read_handler;
+    sds cert_user;
```
madolson: This should probably be on the underlying tlsConnection, not the higher level abstraction.

src/tls.c
zuiderkwast: Then, I think we shouldn't name the config `tls-auth-clients-field`. I prefer we just do like ETCD and use a boolean config. They have a config called `client-cert-auth` (true or false) for exactly the same feature: https://etcd.io/docs/v3.4/op-guide/authentication/#using-tls-common-name

zuiderkwast: How about boolean config `tls-auth-clients-common-name`?

src/tls.c
zuiderkwast: If many clients connect successfully, are we're flooding the logs too much if we log here?

Is it better to log only in the NOT FOUND case? It's the exception case that shouldn't happen often. @madolson WDYT?

madolson: We also mark u->name as sensitive in other places, we should add it behind the hide user data flag. I would drop the log level to verbose if we want to keep the second one.

src/connection.h
madolson: Actually, I'm not sure this needs to be on the struct, looks like it could just be stack allocated inside the function itself.

src/tls.c
zuiderkwast: Good point. If we redact this log message, there isn't much to log at all. Shall we just skip logging here?

If we want it, it would be something like this:

```C
        serverLog(LL_VERBOSE,
                  "TLS: Auto-authenticated client as %s",
                  server.hide_user_data_from_log ? "*redacted*" : u->name);
```

src/tls.c
madolson: I would prefer to focus on the fact we are authenticating the user of the client than on what field is being used. ```tls-auth-clients-user```

src/tls.c
```diff
@@ -677,12 +678,69 @@ static void updateSSLState(connection *conn_) {
...
+    }
+}
+
 static void TLSAccept(void *_conn) {
     tls_connection *conn = (tls_connection *)_conn;
     ERR_clear_error();
     int ret = SSL_accept(conn->ssl);
     if (ret > 0) {
         conn->flags |= TLS_CONN_FLAG_ACCEPT_SUCCESS;
+        autoAuthenticateClientFromCert(conn);
```
madolson: We can't do this here, it's executed from the IO threads and the ACL code is not thread safe, we probably need to grab the user and stuff it on the connection to evaluate later.

src/tls.c
zuiderkwast: `tls-auth-clients-user` SGTM

omanges: @zuiderkwast so we will still be now using the field name right like CN and O, your suggestion is to just change the name of the config to `tls-auth-clients-user` from `tls-auth-clients-field` right ?

src/tls.c
omanges: I am thinking of moving the authentication logic to `clientAcceptHandler` in networking.c and for that we would need to have `cert_user` variable set in `connection` object which can be used in `clientAcceptHandler`

madolson: I think it should be on the tls connection object, since it's not relevant for TLS disabled. Other than that sounds OK.

omanges: I think we can only add the `cert_user` field on the connection object level, rather than TLS because as you said we need to make the ACL call in main thread and we need to make sure the TLS handshake is completed and client is ready, so I think we can that in `clientAcceptHandler` safely in networking.c but the problem is we don't have access to `tls_connection`outside TLS logic hence we need it at connection level. Does that sound correct or am I understanding something incorrectly ?

omanges: @madolson does my earlier seem correct or there is something I am missing ?

src/tls.c
zuiderkwast: No, my suggestion is we use `tls-auth-clients-user yes/no` and it's always the CN. Pointing out the field name doesn't make much sense IMHO.

src/tls.c
zuiderkwast: You mean we shouldn't just add some hack can call a function from tls.c directly from networking.c?

```C
if (connIsTLS(conn)) {
    tlsAutoAuthenticateClientFromCert(conn);
}
```

Yeah, it's a hack that breaks the connection abstraction. I agree.

I don't think we should add the cert user to the connection struct though, because it makes the connection struct larger. Now, it fits within 64 bytes, which is one cache line.

A nicer abstraction would be to add a new callback in the `ConnectionType` struct. We can add a  function `connGetPeerUsername(conn)` that calls a `.get_peer_username` callback that each connection type defines or sets to NULL. If the callback is NULL, `connGetPeerUsername(conn)` just returns NULL (for all connection types except TLS).

The remaining ACL user logic can belong in networking.c instead, because the tls.c should be limited to the connection abstraction.

```C
void clientAcceptHandler(connection *conn) {

    ...

    sds username = connGetPeerUsername(conn);
    if (username != NULL) {
        user *u = ACLGetUserByName(username, sdslen(username));
        if (u) {
            c->user = u;
            c->flag.authenticated = true;
        } else {
            serverLog(LL_NOTICE,
                      "User (%s) from client certificate does not exist",
                      server.hide_user_data_from_log ? "*redacted*" : username);
        }
    }
}
```

omanges: Done updated code

src/connection.h
omanges: Done Updated Code

src/tls.c
omanges: Done update code

src/tls.c
omanges: Done updated

src/tls.c
omanges: Done updated

src/tls.c
omanges: Done updated code

src/config.c
omanges: Done updated

src/connection.c
```diff
@@ -198,3 +198,11 @@ sds getListensInfoString(sds info) {
 
     return info;
 }
+
+sds connGetPeerUsername(connection *conn) {
+    if (conn->type && conn->type->get_peer_username) {
+        return conn->type->get_peer_username(conn);
+    }
+    return NULL;
+}
```
zuiderkwast: We can make it `static inline` and put this together with `connGetPeerCert` in connection.h

Static inline or not doesn't matter much in practice because we compile with -O3 and link-time-optimizations so it can be inlined either way, but we can follow the existing style for similar functions.

src/connection.h
```diff
@@ -124,6 +124,9 @@ typedef struct ConnectionType {
 
     /* Miscellaneous */
     int (*connIntegrityChecked)(void); // return 1 if connection type has built-in integrity checks
+
+    /* Get peer username based on connection type */
+    sds (*get_peer_username)(connection *conn);
```
zuiderkwast, line 141: Let's put this next to `get_peer_cert` a few lines above

```c
    /* TLS specified methods */
    sds (*get_peer_cert)(struct connection *conn);
    sds (*get_peer_username)(connection *conn);
```

src/networking.c
```diff
@@ -1505,6 +1506,23 @@ void clientAcceptHandler(connection *conn) {
         }
     }
 
+    /* Auto-authenticate from cert_user field if set */
+    sds username = connGetPeerUsername(conn);
+    if (username != NULL) {
+        serverLog(LL_VERBOSE, "TLS: Attempting auto-auth using username = %s",
+                  server.hide_user_data_from_log ? "*redacted*" : username);
```
zuiderkwast: This log message is redundant. It will be logged below in both cases: found and not-found user.

src/tls.c
zuiderkwast: It is still an enum with "CN" and "O". If you insist on a config to select the field, I want a convincing motivation, because now I'm not convinced. I don't think it makes sense to use "O" as a username / unique identifier.

If we want to use subjectAltName in the future, then we probably still want CN at the same time, not selecting one of them.

src/networking.c
omanges: Done remove

src/connection.h
omanges: Done updated

src/connection.c
omanges: Done updated

src/tls.c
omanges: @zuiderkwast  I think I missed some comments in this conversation where you said we can make the config like a boolean like yes or no that's why I still made it as boolean, but I remember @madolson  in this conversation saying that there was need for `O` as well so I thought we would need an enum that's why ok NP, so @zuiderkwast  it is final right we will keep it as boolean like yes or no config ?

zuiderkwast: We will need to make this decision in the core team. I hope we will have time to talk about it in the next core team meeting.

No need to change the implementation back and forth before we have a decision.

madolson:
@valkey-io/core-team Trying to make an asynchronous decision. 

You can read the feature at the high level. We are discussing adding a new config flag, `tls-auth-clients-user (yes|no)` which automatically authenticates the client to an ACL user based on Common Name (CN) field of an x.509 certificate. There was a discussion about making it more generic, and allow the admin to configure which field to use, which might have a different config like `tls-auth-clients-user (CN|O|off)`.

Please 👍 if you are okay with the recommendation of just a simple yes/no bool. Otherwise, please comment.

zuiderkwast:
I haven't seen any system using "O" as the user name, but I'm OK with it if we also anticipate that we'll support other fields. Common Name (CN) seems to be the most common, but I've seen read about systems and/or recommendations for using these:

* subjectAltName UPN (User Principal Name) - non-standard field?
  * https://knowledgebase.paloaltonetworks.com/KCSArticleDetail?id=kA10g000000PO1GCAW
* subjectAltName DNS
  * For identifying using a machine's hostname, https://security.stackexchange.com/questions/192979/client-certificate-common-name-subject-alternative-name

What about combinations? If we use an enum, we'd need to define enum values like "CN or SAN DNS". Or shouldn't we accept combinations?

madolson:
Even if there is a remote chance of having some amount of customizations would lead me to recommend we should just use an enum with one option for now CN.

src/tls.c
madolson: Yeah, once we have a decision it should be a small change to finalize it.

src/networking.c
```diff
@@ -1505,6 +1506,20 @@ void clientAcceptHandler(connection *conn) {
...
+    if (username != NULL) {
+        user *u = ACLGetUserByName(username, sdslen(username));
+        if (u) {
+            c->user = u;
+            c->flag.authenticated = true;
+            serverLog(LL_VERBOSE, "TLS: Auto-authenticated client as %s",
+                      server.hide_user_data_from_log ? "*redacted*" : u->name);
+        } else {
+            serverLog(LL_NOTICE, "TLS: Auto-authentication failed, user '%s' not found", username);
+        }
```
madolson, line 1697:
```suggestion
        }
        sdsfree(username);
```

src/server.h
```diff
@@ -1485,6 +1490,7 @@ typedef struct serverTLSContextConfig {
     char *client_cert_file;     /* Certificate to use as a client; if none, use cert_file */
     char *client_key_file;      /* Private key filename for client_cert_file */
     char *client_key_file_pass; /* Optional password for client_key_file */
+    int client_auth_user;      /* Field to be used for Authomatic TLS Authenticaton */
```
```suggestion
    int client_auth_user;      /* Field to be used for automatic TLS authentication based on client TLS certificate */
```
Although, authomatic sounds like a great name for this automatic authentication :)

src/networking.c
```diff
@@ -1505,6 +1506,20 @@ void clientAcceptHandler(connection *conn) {
...
+    sds username = connGetPeerUsername(conn);
+    if (username != NULL) {
+        user *u = ACLGetUserByName(username, sdslen(username));
+        if (u) {
+            c->user = u;
+            c->flag.authenticated = true;
+            serverLog(LL_VERBOSE, "TLS: Auto-authenticated client as %s",
+                      server.hide_user_data_from_log ? "*redacted*" : u->name);
+        } else {
+            serverLog(LL_NOTICE, "TLS: Auto-authentication failed, user '%s' not found", username);
```
```suggestion
            serverLog(LL_NOTICE, "TLS: Auto-authentication failed, user '%s' not found", username);
```
I wonder if this should go in the ACL log instead? This is basically a failed authentication attempt. That removes the concern about dropping user data to the disk and also includes a lot of useful information like IP.

zuiderkwast: Sounds reasonable. I'm not familiar with the ACL log.

src/tls.c
omanges: @madolson @zuiderkwast 
So we would go ahead with enum right for with CN as the supported value correct ?
Just wanted to confirm

src/networking.c
omanges: So no change need right here ?

zuiderkwast: The suggestion is to log using the ACL log instead of serverLog. It's another way to log things, stored in memory instead of on disk. The username may be considered a little sensitive to store in logs. I guess it's this function: `addACLLogEntry`, https://github.com/valkey-io/valkey/blob/8.1.1/src/acl.c#L2604

It looks like it assumes the client is running a command, so it's accessing `c->argv` and/or `c->cmd`, but if the `object` argument (an sds string) is passed to the function, it uses this instead of taking information from the current command.

Not sure if we should invent some new `reason`. I guess we can. Maybe you can try, or do you @madolson have an idea?

src/tls.c
zuiderkwast: I'm OK with this, but it seems the rest of the @valkey-io/core-team members haven't acknowledged/voted for this, so we haven't been able to make a decision about the config yet.

src/networking.c
madolson: My top preference would be to add a new reason, like ACL_INVALID_TLS_CERT_AUTH. That seems very explicit on what happened.

PingXie:
> @valkey-io/core-team Trying to make an asynchronous decision.
> 
> You can read the feature at the high level. We are discussing adding a new config flag, `tls-auth-clients-user (yes|no)` which automatically authenticates the client to an ACL user based on Common Name (CN) field of an x.509 certificate. There was a discussion about making it more generic, and allow the admin to configure which field to use, which might have a different config like `tls-auth-clients-user (CN|O|off)`.
> 
> Please 👍 if you are okay with the recommendation of just a simple yes/no bool. Otherwise, please comment.

@madolson I thought you were advocating for using `O` but you voted for a fixed `CN` field? I would've been siding with a boolean flag if I didn't see your `O` comment. if we still see a need to support `O` in the future, is the plan to introduce a second config?

madolson:
> I thought you were advocating for using O but you voted for a fixed CN field? I would've been siding with a boolean flag if I didn't see your O comment. if we still see a need to support O in the future, is the plan to introduce a second config?

There is a team at amazon who wanted it off the O field, but they honestly said they didn't care. They wrote all this logic in a module so they probably won't adopt what we build regardless.

PingXie:
> > I thought you were advocating for using O but you voted for a fixed CN field? I would've been siding with a boolean flag if I didn't see your O comment. if we still see a need to support O in the future, is the plan to introduce a second config?
> 
> There is a team at amazon who wanted it off the O field, but they honestly said they didn't care. They wrote all this logic in a module so they probably won't adopt what we build regardless.

gotcha. I am less concerned about a specific team's implementation decision but if they see this being a legit use case maybe we should consider a special config instead, just to leave that door open? I feel that authentication on a team/org level isn't a uncommon use case.

src/config.c
```diff
@@ -117,6 +117,12 @@ configEnum tls_auth_clients_enum[] = {
     {"optional", TLS_CLIENT_AUTH_OPTIONAL},
     {NULL, 0}};
 
+configEnum tls_client_auth_user_enum[] = {
+    {"CN", TLS_CLIENT_FIELD_CN},
+    {"O", TLS_CLIENT_FIELD_O},
```
madolson: Latest from the weekly meeting is leave this as an enum but only support `CN` and `off` initially. We can add `O` or something else later as needed.

src/tls.c
madolson: Yes, the consensus from the meeting was do the enum for now and only support `CN` and `off` for the config so we can support future use cases. @omanges I think you should be unblocked to finish the updates and we can merge this shortly!

madolson:
@omanges Will you have time to update this? We would like to have in Valkey 9.0, and just trying to gauge the status of open PRs.

omanges:
@madolson  Yeah I will close it next 1-2 days is it fine ?

omanges:
@madolson found some time so updated the PR  :)

valkey.conf
```diff
@@ -310,6 +310,18 @@ tcp-keepalive 300
...
 
+# Automatically authenticate TLS clients as Valkey users based on their certificate.
+#
+# If set, the server will extract the specified certificate field (e.g., CN) and
+# attempt to find a Valkey user with the same name. If a matching user is found,
+# the client is automatically authenticated as that user upon TLS handshake.
+#
+# Falls back to the default user if no match is found.
+#
+# Supported values: CN
```
zuiderkwast: To disable it, you would need to set it to `""` but it's not documented here.

`""` is a little clunky to use. In the discussions we mentioned to use `off` to disable it. (Some other configs use `no` to disable, e.g. tls-auth-clients (yes/no/optional) but I think it was a boolean config originally (yes/no).)

In the maintainers meeting we decided to use `CN` | `off` (commented just after the meeting in https://github.com/valkey-io/valkey/pull/1920#discussion_r2125591646) so let's use this to avoid the need for another formal decision?

omanges: Yeah sry by mistakenly added "" fixed the issue

omanges:
@zuiderkwast can you help to understand like why are the test failing with `Configuration issue prevented Valkey startup` ?

zuiderkwast:
> @zuiderkwast can you help to understand like why are the test failing with `Configuration issue prevented Valkey startup` ?

Never seen this. Right now I don't see a link to the failed CI jobs. I started them now.

You don't need to force-push. You can just add more commits to the same PR. It makes it easier to review the changes since last review. We will squash-merge the PR into one commit in the end anyway.

src/networking.c
zuiderkwast: Now that ACL log was added, delete the logging using serverLog. We don't want sensitive information in the log and we don't want to flood the log.

```suggestion
```

src/networking.c
```diff
@@ -1682,6 +1683,22 @@ void clientAcceptHandler(connection *conn) {
...
+            addACLLogEntry(c, ACL_INVALID_TLS_CERT_AUTH, 0, 0, username, sdsnew("User not found in auto TLS auth"));
```
zuiderkwast: No need to store the string `"User not found in auto TLS auth"` in the ACL log. The new reason `ACL_INVALID_TLS_CERT_AUTH` already indicates this. This is a structured log, not a plain text verbose log.

The first 0 there (context) should be ACL_LOG_CTX_TOPLEVEL rather than the a magic number 0. The second 0 is an argument position so here 0 is fine. (Argpos is only used for some `reason` when the logging is triggered by some command.)

```suggestion
            addACLLogEntry(c, ACL_INVALID_TLS_CERT_AUTH, ACL_LOG_CTX_TOPLEVEL, 0, username, NULL);
```

Also, you need to handle this new reason when generating the reply of the `ACL LOG` command in `aclCommand()`, here: https://github.com/valkey-io/valkey/blob/8.1.2/src/acl.c#L3016. The output for this case can be something short like "tls-cert", which would match the other reasons.

You can view the log using the `ACL LOG` command. You should probably test this (at least manually in valkei-cli) to see how the reply looks after logging the new `ACL_INVALID_TLS_CERT_AUTH`.

zuiderkwast: Btw, I reopened the resolved comment at serverLog (one line above this comment). Let's delete the line serverLog now that we have ACL log instead.

omanges: Added suggested changes, also verified running locally :- 
```
127.0.0.1:6379> acl log
1)  1) "count"
    2) (integer) 1
    3) "reason"
    4) "tls-cert"
    5) "context"
    6) "toplevel"
    7) "object"
    8) ""
    9) "username"
   10) "client"
   11) "age-seconds"
   12) "48.788"
   13) "client-info"
   14) "id=2 addr=127.0.0.1:41126 laddr=127.0.0.1:6379 fd=10 name= age=0 idle=0 flags=N capa= db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=0 multi-mem=0 rbs=16384 rbp=0 obl=0 oll=0 omem=0 tot-mem=17024 events=r cmd=NULL user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=0 tot-net-out=0 tot-cmds=0"
   15) "entry-id"
   16) (integer) 0
   17) "timestamp-created"
   18) (integer) 1751444186268
   19) "timestamp-last-updated"
   20) (integer) 1751444186268
```

Logs for ACL Stats :-
```
127.0.0.1:6379> INFO all
# Stats
total_connections_received:3
total_commands_processed:8
instantaneous_ops_per_sec:0
total_net_input_bytes:325
total_net_output_bytes:657476
total_net_repl_input_bytes:0
total_net_repl_output_bytes:0
instantaneous_input_kbps:0.00
instantaneous_output_kbps:0.00
instantaneous_input_repl_kbps:0.00
instantaneous_output_repl_kbps:0.00
rejected_connections:0
sync_full:0
sync_partial_ok:0
sync_partial_err:0
expired_keys:0
expired_stale_perc:0.00
expired_time_cap_reached_count:0
expire_cycle_cpu_milliseconds:2
evicted_keys:0
evicted_clients:0
evicted_scripts:0
total_eviction_exceeded_time:0
current_eviction_exceeded_time:0
keyspace_hits:0
keyspace_misses:0
pubsub_channels:0
pubsub_patterns:0
pubsubshard_channels:0
latest_fork_usec:0
total_forks:0
migrate_cached_sockets:0
slave_expires_tracked_keys:0
active_defrag_hits:0
active_defrag_misses:0
active_defrag_key_hits:0
active_defrag_key_misses:0
total_active_defrag_time:0
current_active_defrag_time:0
tracking_total_keys:0
tracking_total_items:0
tracking_total_prefixes:0
unexpected_error_replies:0
total_error_replies:1
dump_payload_sanitizations:0
total_reads_processed:12
total_writes_processed:15
io_threaded_reads_processed:0
io_threaded_writes_processed:0
io_threaded_freed_objects:0
io_threaded_accept_processed:0
io_threaded_poll_processed:0
io_threaded_total_prefetch_batches:0
io_threaded_total_prefetch_entries:0
client_query_buffer_limit_disconnections:0
client_output_buffer_limit_disconnections:0
reply_buffer_shrinks:3
reply_buffer_expands:0
eventloop_cycles:1815
eventloop_duration_sum:615377
eventloop_duration_cmd_sum:5522
instantaneous_eventloop_cycles_per_sec:9
instantaneous_eventloop_duration_usec:284
acl_access_denied_auth:0
acl_access_denied_cmd:0
acl_access_denied_key:0
acl_access_denied_channel:0
invalid_tls_cert_auth_failures:1
```

src/networking.c
omanges: Done removed the `serverLog`

omanges:
@zuiderkwast @madolson  can you please review, I have fixed all the open items

src/tls.c
```diff
@@ -1221,6 +1273,9 @@ static ConnectionType CT_TLS = {
 
     /* Miscellaneous */
     .connIntegrityChecked = connTLSIsIntegrityChecked,
+
+    .get_peer_username = tlsGetPeerUsername,
+
```
zuiderkwast: Move this up a few lines and put it together with get_peer_cert?

```C
    /* TLS specified methods */
    .get_peer_cert = connTLSGetPeerCert,
    .get_peer_username = tlsGetPeerUsername,
```

tests/unit/tls.tcl
```diff
@@ -154,5 +154,23 @@ start_server {tags {"tls"}} {
...
+        test {TLS: Auto-authenticate using tls-auth-clients-user (CN)} {
+            # Create a user matching the CN in the client certificate (CN=Client-only)
+            r ACL SETUSER {Client-only} on >clientpass allcommands allkeys
+
+            # Enable the feature to auto-authenticate based on CN
+            r CONFIG SET tls-auth-clients-user CN
+
+            # With feature on, client should be auto-authenticated using CN=Client-only
+            set s [valkey [srv 0 host] [srv 0 port] 0 1]
+            ::tls::import [$s channel]
```
zuiderkwast, line 167: What is this doing? I'm not aware of `::tls::import`... (Nobody knows TCL for real...)

valkey.conf
```diff
@@ -310,6 +310,21 @@ tcp-keepalive 300
...
+# If set to a field like "CN", the server will extract the corresponding field from
+# the client's TLS certificate and attempt to find a Valkey user with the same name.
+# If a matching user is found, the client is automatically authenticated as that user
+# during the TLS handshake.
+#
+# If no matching user is found, the server falls back to the default user.
+#
+# Set to "off" to disable automatic user authentication via certificate fields.
+#
+# Supported values: CN, off
```
zuiderkwast: Mention that the default is off?

```suggestion
# Supported values: CN, off. Default: off.
```

tests/unit/tls.tcl
omanges: It enables TLS on the already-established connection represented by `$s`

src/server.c
```diff
@@ -5584,9 +5585,11 @@ sds genValkeyInfoStringACLStats(sds info) {
                         "acl_access_denied_auth:%lld\r\n"
                         "acl_access_denied_cmd:%lld\r\n"
                         "acl_access_denied_key:%lld\r\n"
-                        "acl_access_denied_channel:%lld\r\n",
+                        "acl_access_denied_channel:%lld\r\n"
+                        "invalid_tls_cert_auth_failures:%lld\r\n",
```
zuiderkwast: Oh, this is a new INFO field. This is also part of the user-facing API, so we should be careful.

This is not actually counting invalid certificates. Invalid certificate means that it is rejected, for example if it's expired or not trusted by the CA certs. In that case, it would be rejected in the TLS handshake. Here it can be valid but just not match an existing username.

Probably it should start with "acl" like the other fields here.

How about `"acl_access_denied_tls_cert"`? @madolson please chime in.

src/networking.c
```diff
@@ -43,6 +43,7 @@
 #include <math.h>
 #include <ctype.h>
 #include <stdatomic.h>
+#include "connection.h"
```
zuiderkwast: Nit: Usually we want to group the `#include "local-files.h"` together and the `#include <standard-libs.h>` together, so better but this one among the other `"..."` style includes.

zuiderkwast (commented):
Almost there. I noticed the new INFO field. We may need to think about the naming.

src/server.c
omanges: @zuiderkwast @madolson 
But this is not like any ACL rule which has caused access denied so like having `acl_access_denied` as prefix would be misleading, may be, `auto_tls_cert_auth_failures` ?

zuiderkwast: In a way it is just like a failed AUTH command. When the CommonName doesn't match an existing user, you get connected as the default user, which is the same as connecting without AUTH as well as after a failed AUTH command. If ACL is used at all, then typically the default user doesn't have access to anything other than trying to call AUTH again (and a few more commands like HELLO).

omanges: @zuiderkwast @madolson can we finalize something on this ?

omanges: @zuiderkwast  should we go ahead with `acl_access_denied_tls_cert` ?

madolson: acl_access_denied_tls_cert seems good to me!

zuiderkwast: Great, seems we can agree on this. Consider it decided.

src/server.h
```diff
@@ -1311,10 +1315,11 @@ typedef struct writePreparedClient writePreparedClient;
...
 typedef struct aclInfo {
-    long long user_auth_failures;       /* Auth failure counts on user level */
-    long long invalid_cmd_accesses;     /* Invalid command accesses that user doesn't have permission to */
-    long long invalid_key_accesses;     /* Invalid key accesses that user doesn't have permission to */
-    long long invalid_channel_accesses; /* Invalid channel accesses that user doesn't have permission to */
+    long long user_auth_failures;             /* Auth failure counts on user level */
+    long long invalid_cmd_accesses;           /* Invalid command accesses that user doesn't have permission to */
+    long long invalid_key_accesses;           /* Invalid key accesses that user doesn't have permission to */
+    long long invalid_channel_accesses;       /* Invalid channel accesses that user doesn't have permission to */
+    long long acl_access_denied_tls_cert; /* Invalid channel accesses that user doesn't have permission to */
```
zuiderkwast: The comment looks wrong. Copy-pasted from previous line?

Idea:

```suggestion
    long long acl_access_denied_tls_cert; /* TLS clients with cert not matching any exisiting user. */
```

Alternatively, TLS cert auto-authentication failure

zuiderkwast (approved):
Yeah, LGTM. Just one nit about a code comment.

src/server.h
omanges: Done updated

src/networking.c
omanges: Done updated

omanges:
The test were passing earlier like I just changed the text, that shouldn't fail the test :) 
Is it because of flakiness of the tests ?

zuiderkwast:
Flaky tests, yeah, don't worry about it. CI looks good for this change.

src/config.c
```diff
@@ -3372,6 +3378,7 @@ standardConfig static_configs[] = {
     createBoolConfig("tls-prefer-server-ciphers", NULL, MODIFIABLE_CONFIG, server.tls_ctx_config.prefer_server_ciphers, 0, NULL, applyTlsCfg),
     createBoolConfig("tls-session-caching", NULL, MODIFIABLE_CONFIG, server.tls_ctx_config.session_caching, 1, NULL, applyTlsCfg),
     createStringConfig("tls-cert-file", NULL, VOLATILE_CONFIG | MODIFIABLE_CONFIG, EMPTY_STRING_IS_NULL, server.tls_ctx_config.cert_file, NULL, NULL, applyTlsCfg),
+    createEnumConfig("tls-auth-clients-user", NULL, VOLATILE_CONFIG | MODIFIABLE_CONFIG, tls_client_auth_user_enum, server.tls_ctx_config.client_auth_user, TLS_CLIENT_FIELD_OFF, NULL, applyTlsCfg),
```
zuiderkwast: It doesn't need VOLATILE_CONFIG. That's only for configs that are files, e.g. cert file, which can be a new certificate even if the filename hasn't changed. Also no need to reconfigure OpenSSL I guess (which is what applyTlsCfg does).

```diff
-    createEnumConfig("tls-auth-clients-user", NULL, VOLATILE_CONFIG | MODIFIABLE_CONFIG, tls_client_auth_user_enum, server.tls_ctx_config.client_auth_user, TLS_CLIENT_FIELD_OFF, NULL, applyTlsCfg),
+    createEnumConfig("tls-auth-clients-user", NULL, MODIFIABLE_CONFIG, tls_client_auth_user_enum, server.tls_ctx_config.client_auth_user, TLS_CLIENT_FIELD_OFF, NULL, NULL),
```

And now that I think of it, inserting this config between cert file and key file is odd. Cert file and key file very related and used together. A better place for the new one is next to tls-auth-clients:

```suggestion
    createEnumConfig("tls-auth-clients-user", NULL, MODIFIABLE_CONFIG, tls_client_auth_user_enum, server.tls_ctx_config.client_auth_user, TLS_CLIENT_FIELD_OFF, NULL, NULL),
    createBoolConfig("tls-prefer-server-ciphers", NULL, MODIFIABLE_CONFIG, server.tls_ctx_config.prefer_server_ciphers, 0, NULL, applyTlsCfg),
    createBoolConfig("tls-session-caching", NULL, MODIFIABLE_CONFIG, server.tls_ctx_config.session_caching, 1, NULL, applyTlsCfg),
    createStringConfig("tls-cert-file", NULL, VOLATILE_CONFIG | MODIFIABLE_CONFIG, EMPTY_STRING_IS_NULL, server.tls_ctx_config.cert_file, NULL, NULL, applyTlsCfg),
```

zuiderkwast (approved):
Before merging I took a quick look. Noticed some config details to change. I will apply the change and merge. Thank you!

src/networking.c
```diff
@@ -1682,6 +1683,21 @@ void clientAcceptHandler(connection *conn) {
         }
     }
 
+    /* Auto-authenticate from cert_user field if set */
+    sds username = connGetPeerUsername(conn);
+    if (username != NULL) {
+        user *u = ACLGetUserByName(username, sdslen(username));
```
madolson, line 1689: We should probably check here if the user is enabled. We probably don't want to just authenticate straight into the user.

src/networking.c
```diff
@@ -1682,6 +1683,21 @@ void clientAcceptHandler(connection *conn) {
...
+        if (u) {
+            c->user = u;
+            c->flag.authenticated = true;
```
madolson, line 1692:
```
clientSetUser(c, u, 1);
```
I don't remember why we wrote this abstraction, but we should probably use it.

src/networking.c
```diff
@@ -1682,6 +1683,21 @@ void clientAcceptHandler(connection *conn) {
...
+            serverLog(LL_VERBOSE, "TLS: Auto-authenticated client as %s",
+                      server.hide_user_data_from_log ? "*redacted*" : u->name);
```
madolson, line 1694: We also need to trigger `moduleNotifyUserChanged(c);`

madolson (commented):
Sorry. I was late to review this. Would you be able to address some of these comments in a followup, otherwise I can address them.

omanges:
Will address these comments by raising another PR

omanges:
@madolson can you please review the PR :- https://github.com/valkey-io/valkey/pull/2364

I have addressed all the open comments

sandeepkunusoth:
Hi for automatic TLS cert based authentication, when user name with CN is having nopass in ACL how are we restricting this user to login without certificates when tls-auth-clients is optional. we have usecase where we are supporitng both password based and certfifcate based. in this case if user tries to login with CN as username without client certs and without password  they are able to login https://github.com/valkey-io/valkey/issues/2962

zuiderkwast:
We should document the new ACL LOG reason. Maybe we can do it together with some other docs, such as in https://github.com/valkey-io/valkey-doc/pull/439

PR #1926, opened 2025-04-07
Kill RDB child process on CLUSTER RESET command

### LINKED ISSUE #1925: Kill RDB child process on CLUSTER RESET
```
I think it is a coincidence that we don't use chained replication in the cluster mode but that is an orthogonal decision so it is still preferable imo to not take that dependency.

Yeah I think cluster reset should kill the rdb child process too.
```

_Originally posted by @PingXie in https://github.com/valkey-io/valkey/pull/1674#discussion_r2031909790_

hpatro:
Fixes: #1925


PingXie (approved):
Thanks!

PR #1927, opened 2025-04-07
Mark primary node as alive immediately if reachable and failover is not possible

hpatro:
Mark primary node as alive immediately if reachable and failover is not possible

Added test case for failover explicity disabled via `cluster-replica-no-failover`.  Currently, we wait for `cluster_node_timeout * 2` period to mark a failed primary as alive if we are able to communicate with it. For scenario, where a failover won't get triggered, we can mark it as immediately available for better availability.

Before
```
[ok]: no failover - verify replica is not promoted if failover has been disabled (6006 ms)
[ok]: no failover - primary is in failed state (123 ms)
[ok]: no failover - cluster is in healthy state (10138 ms)
```

After
```
[ok]: no failover - verify replica is not promoted if failover has been disabled (5863 ms)
[ok]: no failover - primary is in failed state (120 ms)
[ok]: no failover - cluster is in healthy state (1 ms)
```

The last test `no failover - cluster is in healthy state` showcases the cluster state reached to `ok` after 1ms (with this change) compared to 10ms (unstable) where cluster node timeout is set to 5 ms.


src/cluster_legacy.c
```diff
@@ -2146,12 +2146,21 @@ void clearNodeFailureIfNeeded(clusterNode *node) {
         clusterDoBeforeSleep(CLUSTER_TODO_UPDATE_STATE | CLUSTER_TODO_SAVE_CONFIG);
     }
 
+    /* If any of the replica of a given primary can't failover, then immediately mark it as alive. */
+    int cant_failover = 1;
```
madolson: Is this a common enough failure mode to have special handling for it? Most people only temporarily disable failover when doing upgrades and they probably shouldn't disable it on all replicas. It's also possible there is another replica we are unaware of that is eligible.

hpatro: I looked at the request behind this feature. Seems like user like to place replicas closer to the application for faster reads but don't want it to be part of the cluster failover process. The other use case mentioned is replica used only for backup reasons. Few others don't want automated failover to kick in and only perform the failover(s) manually. All of this documented here: https://github.com/redis/redis/issues/3021 

Also I believe with this change we make availability better for cluster with primary only setup. We are accelerating availability where it seems feasible and in the worst case I believe it was always possible to enter a partitioned state and have multiple primaries in a shard.

PingXie: I agree with this change. ultimately, the truth lies with the primary in question. whether or not a bystander node like this one marks the primary healthy immediately doesn't really change the fact that the primary will regain its primaryship after a flash restart. So the more important thing is the old primary staying down for 2 times cluster-node-timeout; it's less about the observer marking it alive immediately. this approach amplifies an existing problem a bit to trade consistency for availability, which I think is a reasonable decision, provided we are reasonably sure that there won't be replicas competing for the primaryship.

src/cluster_legacy.c
```diff
@@ -2146,12 +2146,21 @@ void clearNodeFailureIfNeeded(clusterNode *node) {
...
+        }
+    }
+
     /* If it is a primary and...
      * 1) The FAIL state is old enough.
      * 2) It is yet serving slots from our point of view (not failed over).
      * Apparently no one is going to fix these slots, clear the FAIL flag. */
     if (clusterNodeIsVotingPrimary(node) &&
-        (now - node->fail_time) > (server.cluster_node_timeout * CLUSTER_FAIL_UNDO_TIME_MULT)) {
+        ((now - node->fail_time) > (server.cluster_node_timeout * CLUSTER_FAIL_UNDO_TIME_MULT) || cant_failover)) {
```
PingXie: `cant_failover` is not a good name. maybe `dont_wait`? or another option would be setting `node->fail_time` to 0 when we know no known replicas would like to step up.

PS, it would be more readable I think if this function returns after handling the non primary case. There is no overlapping between the handling of non primary and primary cases.

tests/unit/cluster/failover2.tcl
```diff
@@ -101,6 +101,39 @@ start_cluster 7 3 {tags {external:skip cluster} overrides {cluster-ping-interval
...
+                fail "Cluster node $j cluster_state:[CI $j cluster_state]"
+            }
+        }
+    }
+
+    resume_process [srv 0 pid]
+
+    test "no failover - cluster is in healthy state" {
+        for {set j 0} {$j < [llength $::servers]} {incr j} {
+            wait_for_condition 1000 50 {
```
PingXie: the total wait is more than twice of the cloud-node-timeout (5s). shouldn't we expect the primary to be marked healthy within the node timeout?

```suggestion
            wait_for_condition 100 50 {
```

src/cluster_legacy.c
hpatro: You read my mind :). I was unhappy with the naming as well. `dont_wait` sounds better.

tests/unit/cluster/failover2.tcl
hpatro: Modifying the test to add primary only cluster scenario. will fix in it.

src/cluster_legacy.c
```diff
@@ -4735,6 +4745,10 @@ void clusterLogCantFailover(int reason) {
     case CLUSTER_CANT_FAILOVER_WAITING_DELAY: msg = "Waiting the delay before I can start a new failover."; break;
     case CLUSTER_CANT_FAILOVER_EXPIRED: msg = "Failover attempt expired."; break;
     case CLUSTER_CANT_FAILOVER_WAITING_VOTES: msg = "Waiting for votes, but majority still not reached."; break;
+    case CLUSTER_CANT_FAILOVER_DISABLED:
+        msg = "Failover has been disabled. "
+              "Please check the 'cluster-replica-no-failover' configuration option";
```
```suggestion
              "Please check the 'cluster-replica-no-failover' configuration option.";
```

src/cluster_legacy.c
```diff
@@ -2146,12 +2146,22 @@ void clearNodeFailureIfNeeded(clusterNode *node) {
...
+    int dont_wait = 1;
+    for (int j = 0; j < node->num_replicas; j++) {
+        if (!clusterNodeIsNoFailover(node->replicas[j])) {
+            dont_wait = 0;
+            break;
+        }
+    }
+
     /* If it is a primary and...
      * 1) The FAIL state is old enough.
```
enjoy-binbin, line 2160: let's also update this line to mention the dont_wait.

src/cluster_legacy.c
```diff
@@ -4827,14 +4841,19 @@ void clusterHandleReplicaFailover(void) {
      * 3) We don't have the no failover configuration set, and this is
```
enjoy-binbin, line 4841: this line need an update.

enjoy-binbin (commented):
Make sense to me.

src/cluster_legacy.c
enjoy-binbin: we have a zoneid in the internal fork (and we will gossip it), usually replicas in the same az as the primary node will have a better ranking. Of course, replication offset is the top priority. When the offsets are the same, the replicas of the same az have better rank and can initiate elections faster. We almost never use the no-failover configuration option.

hpatro: @madolson and I were talking about the node placement with zonal awareness thing and we can also improve the voting members placement and avoid placing all of them in same zone. Let me file an issue on this.

madolson:
I'll follow up with Hari in person. I think this just updates the cluster info field, and not sure if the improvement there is necessarily worth it.

PR #1931, opened 2025-04-08
Fix CLUSTER RESET to use lazyfree-lazy-user-flush to do the lazyfree

enjoy-binbin:
We use lazyfree-lazy-user-flush in #1190 but in #1926 it was changed to use replica-lazy-flush wrongly.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -1294,7 +1294,7 @@ void clusterReset(int hard) {
     if (nodeIsReplica(myself)) {
         clusterSetNodeAsPrimary(myself);
         replicationUnsetPrimary();
-        flushAllDataAndResetRDB(server.repl_replica_lazy_flush ? EMPTYDB_ASYNC : EMPTYDB_NO_FLAGS);
+        flushAllDataAndResetRDB(server.lazyfree_lazy_user_flush ? EMPTYDB_ASYNC : EMPTYDB_NO_FLAGS);
     }
 
     /* Close slots, reset manual failover state. */


enjoy-binbin:
@hpatro I don't see this discussed? If we want to change it, the description in valkey.conf needs to be changed as well.

PingXie (approved):
Good catch! Thanks @enjoy-binbin

hpatro (approved):
> @hpatro I don't see this discussed? If we want to change it, the description in valkey.conf needs to be changed as well.

Good catch. My bad. I did it as a part of the discussion from `CLUSTER REPLICATE NO ONE` PR. Should we have the same behavior for `CLUSTER REPLICATE NO ONE`?

https://github.com/valkey-io/valkey/blob/f7da0c1578cc48db8c7a0e7ed6a3619fa950ee58/src/cluster_legacy.c#L7069

enjoy-binbin:
i am not sure about it. replicate use `repl_replica_lazy_flush` also make sense to me.
@madolson Do you have a preference?

PingXie:
> replicate use repl_replica_lazy_flush also make sense to me.

hmm, that works too I think. that being said, there should've been just one lazy_flush config imo.

madolson:
> i am not sure about it. replicate use repl_replica_lazy_flush also make sense to me. madolson Do you have a preference?

I have a vague preference for user flush since this is a user triggered action. But, I think either are OK.

I honestly think we might consider merging user-flush and replication flush in the future.

hpatro:
> > i am not sure about it. replicate use repl_replica_lazy_flush also make sense to me. madolson Do you have a preference?
> 
> 
> 
> I have a vague preference for user flush since this is a user triggered action. But, I think either are OK.
> 
> 
> 
> I honestly think we might consider merging user-flush and replication flush in the future.

Let's do it in 9.0?

PR #1934, opened 2025-04-08
Update function clusterNodeSetSlotBit() return type to void

hwware:
No other functions use the return value of clusterNodeSetSlotBit(), thus this PR update the return type to void and
update the comments.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -5606,10 +5606,8 @@ int clusterPrimariesHaveReplicas(void) {
     return replicas != 0;
 }
 
-/* Set the slot bit */
 void clusterNodeSetSlotBit(clusterNode *n, int slot) {
-    int old = bitmapTestBit(n->slots, slot);
-    if (!old) {
+    if (!bitmapTestBit(n->slots, slot)) {
         bitmapSetBit(n->slots, slot);
         n->numslots++;
         /* When a primary gets its first slot, even if it has no replicas,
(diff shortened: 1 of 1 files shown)


src/cluster_legacy.c
```diff
@@ -5546,8 +5546,8 @@ int clusterPrimariesHaveReplicas(void) {
     return replicas != 0;
 }
 
-/* Set the slot bit and return the old value. */
-int clusterNodeSetSlotBit(clusterNode *n, int slot) {
+/* Set the slot bit */
+void clusterNodeSetSlotBit(clusterNode *n, int slot) {
     int old = bitmapTestBit(n->slots, slot);
     if (!old) {
```
xbasel:
```
if(!bitmapTestBit(n->slots, slot)) {
```

src/cluster_legacy.c
```diff
@@ -5546,8 +5546,8 @@ int clusterPrimariesHaveReplicas(void) {
     return replicas != 0;
 }
 
-/* Set the slot bit and return the old value. */
-int clusterNodeSetSlotBit(clusterNode *n, int slot) {
+/* Set the slot bit */
```
xbasel: Redundant. The function name `clusterNodeSetSlotBit` already says it.

src/cluster_legacy.c
hwware: Thanks for your comments. It is removed now.

zuiderkwast (approved):

PingXie (approved):

enjoy-binbin (approved):

PR #1943, opened 2025-04-11
Improve clarity of errors for GEO commands when member does not exist

chx9:
this pr aims to fix one of the issues from Redis Triage
https://github.com/orgs/redis/projects/6?pane=issue&itemId=69775417&issue=redis%7Credis%7C13350
it has been more than 1 year since this issue brought up

### DIFF
--- a/src/geo.c
+++ b/src/geo.c
@@ -112,11 +112,17 @@ int extractLongLatOrReply(client *c, robj **argv, double *xy) {
 /* Input Argument Helper */
 /* Decode lat/long from a zset member's score.
  * Returns C_OK on successful decoding, otherwise C_ERR is returned. */
-int longLatFromMember(robj *zobj, robj *member, double *xy) {
+int longLatFromMember(robj *zobj, robj *member, double *xy, sds* err) {
     double score = 0;
 
-    if (zsetScore(zobj, member->ptr, &score) == C_ERR) return C_ERR;
-    if (!decodeGeohash(score, xy)) return C_ERR;
+    if (zsetScore(zobj, member->ptr, &score) == C_ERR){
+        *err = sdsnew("member des not exist");
+        return C_ERR;
+    } 
+    if (!decodeGeohash(score, xy)) {
+        *err = sdsnew("failed to decode");
+        return C_ERR;
+    }
     return C_OK;
 }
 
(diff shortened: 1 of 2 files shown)


Fusl:
`longLatFromMember` can also return `C_ERR` if `geohashDecodeToLongLatType` returns a decoding error while the member did exist. Should we maybe differentiate those cases or just return `member %s does not exist or failed to decode`?

madolson:
> longLatFromMember can also return C_ERR if geohashDecodeToLongLatType returns a decoding error while the member did exist. Should we maybe differentiate those cases or just return member %s does not exist or failed to decode?

Ideally yes. It would be nice to differentiate the member be missing from some type of corruption.

chx9:
> > longLatFromMember can also return C_ERR if geohashDecodeToLongLatType returns a decoding error while the member did exist. Should we maybe differentiate those cases or just return member %s does not exist or failed to decode?
> 
> Ideally yes. It would be nice to differentiate the member be missing from some type of corruption.

agree, i added error infomation for each case.

src/geo.c
```diff
@@ -112,11 +112,17 @@ int extractLongLatOrReply(client *c, robj **argv, double *xy) {
 /* Input Argument Helper */
 /* Decode lat/long from a zset member's score.
  * Returns C_OK on successful decoding, otherwise C_ERR is returned. */
-int longLatFromMember(robj *zobj, robj *member, double *xy) {
+int longLatFromMember(robj *zobj, robj *member, double *xy, sds* err) {
```
madolson:
```suggestion
int longLatFromMemberOrReply(client *c, robj *zobj, robj *member, double *xy) {
```
A more common pattern we use is "validate" or reply. This makes it so we don't have to pass a buncch of stack allocated objects back and forth. If we pass in the client object, we can do all the logic and then also put the replies on the client object here. The replying code becomes centralized as well.

madolson (commented):
Functionality looks good, I have a high level comment for trying to make it a little bit simpler. Let me know if it makes sense.

src/geo.c
```diff
@@ -112,11 +112,17 @@ int extractLongLatOrReply(client *c, robj **argv, double *xy) {
...
     double score = 0;
 
-    if (zsetScore(zobj, member->ptr, &score) == C_ERR) return C_ERR;
-    if (!decodeGeohash(score, xy)) return C_ERR;
+    if (zsetScore(zobj, member->ptr, &score) == C_ERR){
+        *err = sdsnew("member des not exist");
```
zuiderkwast: Spelling: des -> does

src/geo.c
chx9: yeah, its cleaner, i fixed it

src/geo.c
chx9: thanks, fixed

tests/unit/geo.tcl
```diff
@@ -345,7 +351,12 @@ start_server {tags {"geo"}} {
         set ret [r GEORADIUSBYMEMBER k1 n1 5009431 m]
         assert_equal $ret {n1 n2}
     }
-
+    test {GEOSEARCH FROMMEMBER member does not exist} {
```
madolson, line 354: Code looks good. If you can add a test for the negative case for the invalid geohash, (you can just do ZADD on the Sicilty object with some invalid data) I think it'll be good to go.

chx9: i found out decodeGeohash could never return 0, maybe i should remove the second addReplyErrorFormat, what do you think？

madolson: You're right, invalid values get clamped into the "valid" range of values and returned, even when they should probably be returned as invalid. I think we can just leave it as a defensive check.

chx9: done, I left the invalid return val, but can't add tests for that.

chx9: done, I left the invalid return val, but can't add tests for that.

zuiderkwast (approved):
LGTM

madolson (approved):
Thanks!

PR #1944, opened 2025-04-11
Optimize string-to-integer performance using AVX512

zhulipeng:
### Description

This pull request aims to optimize the `string2ll` function using AVX512 instructions to enhance performance for string-to-integer conversions. The existing scalar version of the function was identified as a bottleneck in certain scenarios, and this optimization seeks to address that.

### Changes

- Introduced string2llAVX512 function in util.c to utilize AVX512 SIMD instructions.
- Updated config.h to define ATTRIBUTE_TARGET_AVX512 for AVX512 support.
- Replaced calls to `lpStringToInt64` with `string2ll` for consistency and performance improvement.
- Removed redundant code and optimized integer encoding logic.

### Performance Boost

The corresponding list commands can gain up to **~19%** performance improvement, the string length is 4 bytes. The longer the string, the greater the performance boost.

|Benchmark|Performance Boost|
|-|-|
|[memtier_benchmark-1key-list-10K-elements-lpos-integer](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-list-10K-elements-lpos-integer.yml)|**19%**|
|[memtier_benchmark-1key-list-10K-elements-linsert-lrem-integer](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-list-10K-elements-linsert-lrem-integer.yml)|**17%**|

#### Test Env
- OS: CentOS Stream 9
- Platform: Intel Xeon 6980P

```
port 9001
bind * -::*
daemonize no
protected-mode no
save
```

### Ref
1. http://0x80.pl/notesen/2018-04-19-simd-parsing-int-sequences.html#toc-entry-1
2. https://lemire.me/blog/2023/09/22/parsing-integers-quickly-with-avx-512/


zhulipeng:
Kindly ping @valkey-io/core-team and @xbasel, could you help to take a look at this?

src/unit/test_util.c
```diff
@@ -69,6 +69,9 @@ int test_string2ll(int argc, char **argv, int flags) {
     valkey_strlcpy(buf, "9223372036854775808", sizeof(buf)); /* overflow */
     TEST_ASSERT(string2ll(buf, strlen(buf), &v) == 0);
 
+    valkey_strlcpy(buf, "18446744073709551615", sizeof(buf)); /* overflow */
+    TEST_ASSERT(string2ll(buf, strlen(buf), &v) == 0);
+
```
xbasel, line 74: Anyway to add a test to make sure scalar and avx512 produce the same result for various inputs?

xbasel:
Invoked scalar vs avx512 impl in a loop, on
`11th Gen Intel(R) Core(TM) i5-1145G7 @ 2.60GHz`

Input:     `const char *input = "123456789012345";`


got:
```
Scalar: 11112155810 ns total
AVX512: 2084868190 ns total
AVX512 is 81.24% faster
```


However, for `input=1234`

```
Scalar: 725640551 ns total
AVX512: 1383065015 ns total
AVX512 is -90.60% faster
```

Worth checking various input lengths

src/util.c
```diff
@@ -493,6 +620,17 @@ int string2ll(const char *s, size_t slen, long long *value) {
     return 1;
 }
 
+int string2ll(const char *s, size_t slen, long long *value) {
+#ifdef HAVE_AVX512
+    /* TODO: Cache the cpu info when server startup */
+    if (__builtin_cpu_supports("avx512f") &&
+        __builtin_cpu_supports("avx512vl") &&
+        __builtin_cpu_supports("avx512bw"))
+        return string2llAVX512(s, slen, value);
```
xbasel, line 629: are we sure AVX512 outperforms  scalar for any length?

xbasel:
Quick length benchmark:
```
Len    Scalar(ns)   AVX512(ns)   Speedup
1      60030323     127997587    -113.22%
2      195774319    212610792     -8.60%
3      244857456    211755460     13.52%
4      321833764    212453782     33.99%
5      421814801    213579500     49.37%
6      545647849    214139725     60.75%
7      624711396    212615874     65.97%
8      662040600    210967195     68.13%
9      732496479    276282116     62.28%
10     845303767    384372263     54.53%
11     905955500    275522156     69.59%
12     992389416    283623473     71.42%
13     1059978028   275161163     74.04%
14     1161004405   275442902     76.28%
15     1235948240   278176613     77.49%
16     1328106710   276794788     79.16%
17     1403942807   351625731     74.95%
18     1503215390   352354691     76.56%
19     1625629382   352029760     78.35%
```

src/util.c
zhulipeng: I used to change the benchmark commands by using `memtier_benchmark -s 127.0.0.1 -p 9001 --command="LPOS intlist 5"  --hide-histogram --test-time 120`, didn't see an obvious regression for the 1-byte input.

src/unit/test_util.c
zhulipeng: Maybe the existing `test_string2ll` unit test is enough?

src/util.c
xbasel: Please note that the impact on short lengths, can be invisible on end to end testing, even if its different when measuring the runtime of these functions directly, in this case, we should always run avx512.

zhulipeng:
Ok, I just double confirmed on `Intel(R) Xeon(R) 6980P` for the short length, no performance regression observed.  
1. fill test data: `memtier_benchmark -s 127.0.0.1 -p 9001 --hide-histogram --command "RPUSH intlist __key__" --key-maximum 10000 --key-minimum 1 -n allkeys --key-prefix "" --command-key-pattern S -c 1 -t 1 --pipeline 10`
2. change to input 1-byte to benchmark: `memtier_benchmark -s 127.0.0.1 -p 9001 --command="LPOS intlist 5"  --hide-histogram --test-time 120`

src/unit/test_util.c
xbasel: yea reviewed the entire test now. this test should be enough.

xbasel (approved):

src/util.c
zuiderkwast: Branching based on length would be slower?

zhulipeng: > Branching based on length would be slower?

Yes. In my local environment, using the `slen > 2` check results in a 2% performance regression.

zuiderkwast:
Much of the integer parsing is done in the RESP parsing, like command arguments lengths and number of arguments. Most of these are single digit, so I think the single-digit case is important. Perhaps it's still very fast compared to the other parsing logic like conditional branching.

@xbasel How many loops did you do in the benchmarking? The total nanoseconds don't say much. The time for a single string2ll would be more interesting.

@lipzhu How come you and @xbasel got some different results?

Shall we add a micro benchmark like this test to `src/unit/test_util.c`?

```C
static long long usec(void) {
    struct timeval tv;
    gettimeofday(&tv, NULL);
    return (((long long)tv.tv_sec) * 1000000) + tv.tv_usec;
}

int test_string2llBenchmark(int argc, char **argv, int flags) {
    UNUSED(argc);
    UNUSED(argv);
    UNUSED(flags);
    const long loop = 1000000;
    const char *input = "123456789012345";
    int maxlen = strlen(input);
    for (int len = 1; len <= maxlen; len++) {
        long long v;
        long long start = usec();
        for (long i = 0; i < loop; i++) {
            TEST_ASSERT(string2ll(input, len, &v));
        }
        long long end = usec();
        printf("string2ll length %2d in %6.3lf ns\n", len, (end - start) * 1000.0 / loop);
    }
    return 0;
}
```

I tried this on my ARM laptop (without AVX512) and noticed that it takes less than one nanosecond per digit, so we're talking about quite small numbers here.

```
string2ll length  1 in  0.453 ns
string2ll length  2 in  2.275 ns
string2ll length  3 in  3.574 ns
string2ll length  4 in  3.908 ns
string2ll length  5 in  4.804 ns
string2ll length  6 in  5.045 ns
string2ll length  7 in  5.556 ns
string2ll length  8 in  6.031 ns
string2ll length  9 in  6.506 ns
string2ll length 10 in  7.019 ns
string2ll length 11 in  7.507 ns
string2ll length 12 in  8.311 ns
string2ll length 13 in  9.055 ns
string2ll length 14 in  9.842 ns
string2ll length 15 in 10.663 ns
```

xbasel:
@zuiderkwast 
I used 100000000 iterations (100 million)

I think @lipzhu performed end-to-end tests (by running Valkey commands). I extracted **string2llAVX512** and **string2llScalar** into a single C file and added a main function to test them. So the execution time reported is the pure execution time of these functions. Although my results show degradation on small lengths, it might not be noticeable when measured as part of a bigger transaction.

Code:
```

int main() {
    const int iters = 100000000;
    long long val;

    char input[20];  // 19 digits + null
    printf("%-6s %-12s %-12s %-10s\n", "Len", "Scalar(ns)", "AVX512(ns)", "Speedup");

    for (int len = 1; len <= 19; ++len) {
        // Generate input of length `len`: e.g. "111...1"
        for (int i = 0; i < len; ++i) input[i] = '1';
        input[len] = '\0';

        // Sanity check
        string2llScalar(input, len, &val);
        long long val2;
        string2llAVX512(input, len, &val2);
        if (val != val2) {
            printf("Mismatch at len %d: scalar=%lld, avx512=%lld\n", len, val, val2);
            continue;
        }

        // Scalar
        uint64_t t1 = now_ns();
        for (int i = 0; i < iters; ++i)
            string2llScalar(input, len, &val);
        uint64_t t2 = now_ns();
        uint64_t scalar_time = t2 - t1;

        // AVX512
        t1 = now_ns();
        for (int i = 0; i < iters; ++i)
            string2llAVX512(input, len, &val);
        t2 = now_ns();
        uint64_t avx_time = t2 - t1;

        double speedup = ((double)scalar_time - avx_time) * 100.0 / scalar_time;

        printf("%-6d %-12llu %-12llu %6.2f%%\n", len,
               (unsigned long long)scalar_time,
               (unsigned long long)avx_time,
               speedup);
    }

    return 0;
}

static inline uint64_t now_ns() {
    struct timespec ts;
    clock_gettime(CLOCK_MONOTONIC, &ts);
    return (uint64_t)ts.tv_sec * 1000000000 + ts.tv_nsec;
}
```

Normalized results:

| Len | Scalar per-op (ns) | AVX512 per-op (ns) |
|-----|--------------------|--------------------|
|  1  | 0.6003              | 1.2800              |
|  2  | 1.9577              | 2.1261              |
|  3  | 2.4486              | 2.1176              |
|  4  | 3.2183              | 2.1245              |
|  5  | 4.2181              | 2.1358              |
|  6  | 5.4565              | 2.1414              |
|  7  | 6.2471              | 2.1262              |
|  8  | 6.6204              | 2.1097              |
|  9  | 7.3249              | 2.7628              |
| 10  | 8.4530              | 3.8437              |
| 11  | 9.0596              | 2.7552              |
| 12  | 9.9239              | 2.8362              |
| 13  | 10.5998             | 2.7516              |
| 14  | 11.6100             | 2.7544              |
| 15  | 12.3595             | 2.7818              |
| 16  | 13.2811             | 2.7679              |
| 17  | 14.0394             | 3.5163              |
| 18  | 15.0322             | 3.5235              |
| 19  | 16.5626             | 3.5203              |

src/config.h
```diff
@@ -387,4 +387,16 @@ void setcpuaffinity(const char *cpulist);
 #define ATTRIBUTE_TARGET_AVX2
 #endif
 
+#if defined(__x86_64__) && ((defined(__GNUC__) && __GNUC__ >= 5) || (defined(__clang__) && __clang_major__ >= 4))
+#if defined(__has_attribute) && __has_attribute(target)
+#define HAVE_AVX512
+#endif
+#endif
```
zuiderkwast: This check is identical to the defition of `HAVE_AVX2`.

Let's rename `HAVE_AVX2` to something that actually explains what it does and avoid the duplicated definition.

We check that `__has_attribute(target)` exits but we also use this macro to guard the use of `__builtin_cpu_supports()`.

How about something like `HAVE_X86_SIMD`?

src/util.c
```diff
@@ -415,6 +415,133 @@ int ull2string(char *dst, size_t dstlen, unsigned long long value) {
...
+    int negative = 0;
+
+    /* Abort if length indicates this cannot possibly be an int */
+    if (slen == 0 || slen >= LONG_STR_SIZE) return 0;
+
+    /* Special case: first and only digit is 0. */
+    if (slen == 1 && p[0] == '0') {
+        if (value != NULL) *value = 0;
+        return 1;
+    }
```
zuiderkwast: Could we modify this case to include all single-digit numbers without any performance impact?

In the hot path in command parsing, `string2ll()` is used in `processMultibulkBuffer()`. Here, I guess it's often single digit when command names are parsed and key names are parsed.

```suggestion
    /* Special case: first and only digit is 0. Also handle single-digit 1-9 here. */
    if (slen == 1 && p[0] >= '0' && p[0] <= '9') {
        if (value != NULL) *value = p[0] - '0';
        return 1;
    }
```

zuiderkwast (commented):
LGTM in general. It's good that `lpStringToInt64()` could be deleted!

src/util.c
```diff
@@ -415,6 +415,133 @@ int ull2string(char *dst, size_t dstlen, unsigned long long value) {
...
+        negative = 1;
+        p++;
+        plen++;
+
+        /* Abort on only a negative sign. */
+        if (plen == slen) return 0;
+    }
+
+    /* If first digit is 0, the string should just be 0. */
+    if (p[0] == '0') {
```
zuiderkwast: We don't use the *unlikely* annotation much, but in highly optimized code like this, would it make sense to use it to optimize branch prediction for the success cases?

```suggestion
    if (unlikely(p[0] == '0')) {
```

src/util.c
zuiderkwast: Shall we do this in `string2llScalar()` too?

src/config.h
```diff
@@ -374,17 +374,19 @@ void setcpuaffinity(const char *cpulist);
...
+/* Check if we can compile SIMD code */
 #if defined(__x86_64__) && ((defined(__GNUC__) && __GNUC__ >= 5) || (defined(__clang__) && __clang_major__ >= 4))
 #if defined(__has_attribute) && __has_attribute(target)
-#define HAVE_AVX2
+#define HAVE_X86_SIMD 1
 #endif
 #endif
 
-#if defined(HAVE_AVX2)
+#if HAVE_X86_SIMD
```
zuiderkwast: When I try to compile with `-Wundef` (e.g. `make WARNINGS=-Wundef`) on ARM, I get a warnings about undefined macros. `HAVE_X86_SIMD` here is one of the undefined ones.

```
config.h:286:6: warning: ‘__i386’ is not defined, evaluates to ‘0’ [-Wundef]
  286 | #if (__i386 || __amd64 || __powerpc__) && __GNUC__
      |      ^~~~~~
config.h:286:16: warning: ‘__amd64’ is not defined, evaluates to ‘0’ [-Wundef]
  286 | #if (__i386 || __amd64 || __powerpc__) && __GNUC__
      |                ^~~~~~~
config.h:286:27: warning: ‘__powerpc__’ is not defined, evaluates to ‘0’ [-Wundef]
  286 | #if (__i386 || __amd64 || __powerpc__) && __GNUC__
      |                           ^~~~~~~~~~~
config.h:386:5: warning: ‘HAVE_X86_SIMD’ is not defined, evaluates to ‘0’ [-Wundef]
  386 | #if HAVE_X86_SIMD
      |     ^~~~~~~~~~~~~
```

If you want to change `#if defined` to just `#if`, the macros should always be defined, so add an `#else` clause and define it to 0 here. Ideally, we should fix the other three warnings too so we can enable `-Wundef` by default.

src/util.c
```diff
@@ -415,6 +415,134 @@ int ull2string(char *dst, size_t dstlen, unsigned long long value) {
     return 0;
 }
 
+#if HAVE_AVX512
```
zuiderkwast: This macro is undefined. It was renamed. (`-Wundef` helped me find it.)

src/util.c
```diff
@@ -493,6 +621,17 @@ int string2ll(const char *s, size_t slen, long long *value) {
     return 1;
 }
 
+int string2ll(const char *s, size_t slen, long long *value) {
+#if HAVE_AVX512
```
zuiderkwast: Renamed macro

src/util.c
```diff
@@ -493,6 +621,17 @@ int string2ll(const char *s, size_t slen, long long *value) {
...
+    /* TODO: Cache the cpu info when server startup */
+    if (__builtin_cpu_supports("avx512f") &&
+        __builtin_cpu_supports("avx512vl") &&
+        __builtin_cpu_supports("avx512bw"))
```
zuiderkwast: Should we fix the TODO? If it doesn't matter, then we can remove the TODO comment.

zhulipeng: I just remove the TODO comments; I will send another PR to address that, try to remove the runtime checker from the critical path, the impacted code is `hyperloglog.c` and `util.c`.

zuiderkwast: Ah, if you want to fix it in the future, it's OK to keep the TODO comment as a reminder. I just assumed we will never fix it and it's not important.

Can we use a simple static variable?

```c
    static int cpu_supports_avx512 = __builtin_cpu_supports("avx512f") && ...;
    if (cpu_supports_avx512) ...
```

src/config.h
zhulipeng: > Ideally, we should fix the other three warnings too so we can enable `-Wundef` by default.

@zuiderkwast, would you prefer to address the warning in this PR along with `HAVE_X86_SIMD`, or should we handle it in a separate one?

src/util.c
zhulipeng: > Can we use a simple static variable?
> 
> ```c
>     static int cpu_supports_avx512 = __builtin_cpu_supports("avx512f") && ...;
>     if (cpu_supports_avx512) ...
> ```

The `__builtin_cpu_supports()` function is meant to be executed at runtime to determine CPU capabilities. When used in a static initializer context like this, it evaluates at compile time based on the compilation target rather than the actual CPU where the code will run.

zhulipeng: I prefer to retrieve the CPUID information during server startup. We could introduce a new server data structure, `cpuid`, to check for support of SSE, SSE2, AVX2, AVX512, AVX10, and other SIMD instructions.

zuiderkwast: If it is, then why isn't also every `__builtin_cpu_supports()` evaluated at compile time even in code like ↓ ?

```C
    if (__builtin_cpu_supports("avx512f") &&
        __builtin_cpu_supports("avx512vl") &&
        __builtin_cpu_supports("avx512bw"))
```

zhulipeng: When I compile code like ```static int cpu_supports_avx512 = __builtin_cpu_supports("avx512f");```
error: initializer element is not constant. 

static initiation like that worked for c++, but failed on c.

src/config.h
zuiderkwast: It's not a blocker to merging this, but I think it would be nice to have and it doesn't matter if it's the same or a different PR.

It's easier to add it in the same PR?

src/util.c
zuiderkwast: Ah, ok. I will let you figure it out in a separate PR. :)

zuiderkwast (approved):
It looks good, thank you! ~~If you want to add the `-Wundef` flag in Makefile and CMake files, I can wait for that. Otherwise I'll merge this tomorrow.~~

I merged it.

src/config.h
zuiderkwast: Actually, let's do it in a separate PR. Maybe there will be some problems and I don't want to block this PR.

PR #1948, opened 2025-04-13
Fix panic in primary when blocking shutdown after previous block with timeout

murphyjacob4:
When previous timeout isn't zeroed out, shutdown will use the previously executed blocking commands timeout. Since shutdown is not expected to have timeouts, this causes a serverPanic.

Without this fix, this scenario leads to a server panic:

1. Call any command that uses blockClient with a timeout (this could be BLPOP or something that blocks implicitly like CLUSTER SETSLOT)
2. On the same client, call SHUTDOWN with some pending replication data, which will trigger shutdown to be blocked while we try to catch the replica up

### DIFF
--- a/src/blocked.c
+++ b/src/blocked.c
@@ -653,6 +653,8 @@ void blockPostponeClient(client *c) {
 
 /* Block client due to shutdown command */
 void blockClientShutdown(client *c) {
+    initClientBlockingState(c);
+    c->bstate->timeout = 0;
     blockClient(c, BLOCKED_SHUTDOWN);
 }
 
--- a/tests/integration/shutdown.tcl
+++ b/tests/integration/shutdown.tcl
@@ -244,3 +244,41 @@ test "Shutting down master waits for replica then aborted" {
         }
     }
 } {} {repl external:skip}
+
+test "Shutting down primary wait for replica after previous block" {
+    start_server {overrides {save "" repl-backlog-size 1MB}} {
+        start_server {overrides {save "" repl-backlog-size 1MB}} {
+            set primary [srv -1 client]
+            set primary_deferring [valkey_deferring_client -1]
+            set primary_host [srv -1 host]
+            set primary_port [srv -1 port]
+            set primary_pid [srv -1 pid]
+            set replica [srv 0 client]
+            set replica_pid [srv 0 pid]
+
+            $primary_deferring BLPOP key 0.1
+
+            # Config primary and replica.
+            $replica replicaof $primary_host $primary_port
+            wait_for_sync $replica
+
+            # Pause the replica and write a key on primary.
+            pause_process $replica_pid
+            $primary incr k
+
+            # Call shutdown.
+            $primary_deferring shutdown
+
+            set info_clients [$primary info clients]
+            assert_match "*connected_clients:2*" $info_clients
+            assert_match "*blocked_clients:1*" $info_clients
+
+            # Wake up replica, causing primary to continue shutting down.
+            resume_process $replica_pid
+            $primary_deferring close
+
+            # Check for crash.
+            assert_equal [count_log_message -1 "*BUG REPORT START*"] 0
+        }
+    }
+} {} {repl external:skip}


tests/unit/cluster/slot-migration.tcl
```diff
@@ -511,3 +511,34 @@ start_cluster 3 6 {tags {external:skip cluster} overrides {cluster-node-timeout
         wait_for_slot_state 2 ""
     }
 }
+
+
+proc log_file_matches {log pattern} {
```
madolson: Is there a reason you re-implemented this function? It seems to exist elsewhere?

src/blocked.c
```diff
@@ -251,6 +251,7 @@ void unblockClient(client *c, int queue_for_reprocessing) {
     c->bstate->btype = BLOCKED_NONE;
     c->bstate->unblock_on_nokey = 0;
     removeClientFromTimeoutTable(c);
+    c->bstate->timeout = 0;
```
madolson: This seems to be breaking some other tests.

tests/unit/cluster/slot-migration.tcl
murphyjacob4: I think we have a few copies of it, I can move it to util.tcl instead of copying it

murphyjacob4: I guess we have count_log_message already there - just going to use that

src/blocked.c
murphyjacob4: Looks like it needs to be reset on creation instead, since query reprocessing relies on the original timeout value. Should be fixed now

tests/integration/shutdown.tcl
```diff
@@ -244,3 +244,41 @@ test "Shutting down master waits for replica then aborted" {
         }
     }
 } {} {repl external:skip}
+
+test "Shutting down master wait for replica after previous block" {
+    start_server {overrides {save "" repl-backlog-size 1MB}} {
+        start_server {overrides {save "" repl-backlog-size 1MB}} {
+            set master [srv -1 client]
```
zuiderkwast: We should call it `primary` in new tests.

zuiderkwast (commented):
Great fix.

This problem isn't recently introduced, so we need to backport it to 7.2 and 8.0, right?

How do you reproduce the this panic? The PR title should mention what scenario this fixes, in the user's POV, more like in the title of the test case. This is what will be seen in the release notes, so it's better that it's comprehensive for users.

Not sure if this is correct but you get the idea: *Fix panic when shutting down primary waiting for replica when blocked client exists*

murphyjacob4:
> How do you reproduce the this panic? The PR title should mention what scenario this fixes, in the user's POV, more like in the title of the test case. This is what will be seen in the release notes, so it's better that it's comprehensive for users.

Updated the title. The repro is the following:

1. Call any command that uses blockClient with a timeout (this could be BLPOP or something that blocks implicitly like CLUSTER SETSLOT)
2. On the same client, call SHUTDOWN with some pending replication data, which will trigger shutdown to be blocked while we try to catch the replica up

Took a stab at rewording it, but let me know if it needs more details

zuiderkwast (approved):

tests/integration/shutdown.tcl
```diff
@@ -244,3 +244,41 @@ test "Shutting down master waits for replica then aborted" {
...
+            # Pause the replica and write a key on primary.
+            pause_process $replica_pid
+            $primary incr k
+
+            # Call shutdown.
+            $primary_deferring shutdown
+
+            set info_clients [$primary info clients]
+            assert_match "*connected_clients:2*" $info_clients
+            assert_match "*blocked_clients:1*" $info_clients
```
madolson, line 274: This failed after your merge, https://github.com/valkey-io/valkey/actions/runs/14476563889/job/40603532047#step:4:6837. 

This might need to be a wait since it's possible this client's command is executed before the block?

murphyjacob4: Thanks for surfacing, sent https://github.com/valkey-io/valkey/pull/1960 for the fix

PR #685, opened 2024-06-23
Align the entry read and lag rules of the consumer group

artikell:
In xinfo, `entry-read` is not calculated in real-time, but `lag` is calculated in real-time, resulting in many combinations of lag and entry read relationships.

After all data is read, the rule of `entries-read + lag = entries-added` will be followed.

Therefore, the current PR guarantees that the results of `xinfo` command always follow the `entries-read + lag = entries-added` rule.

PR #1952, opened 2025-04-14
Fix incorrect lag reported in XINFO GROUPS

### LINKED ISSUE #1951: [BUG] Stream consumer group lag incorrectly calculated when entries are deleted before being read
**Describe the bug**

When a tombstone is created after the last_id of a consumer group the lag is being incorrectly calculated after the next read, leaving always a lag that is not there.

**To reproduce**

```bash
127.0.0.1:6379> xadd mystream 1-0 data a
"1-0"
127.0.0.1:6379> xadd mystream 1-0 data a
"1-0"
127.0.0.1:6379> xadd mystream 2-0 data b
"2-0"
127.0.0.1:6379> xadd mystream 3-0 data c
"3-0"
127.0.0.1:6379> xadd mystream 4-0 data d
"4-0"
127.0.0.1:6379> XGROUP CREATE mystream g1 0
OK
127.0.0.1:6379> XREADGROUP GROUP g1 c1 COUNT 2 STREAMS mystream >
1) 1) "mystream"
   2) 1) 1) "1-0"
         2) 1) "data"
            2) "a"
      2) 1) "2-0"
         2) 1) "data"
            2) "b"
127.0.0.1:6379> XINFO GROUPS mystream
1)  1) "name"
    2) "g1"
    3) "consumers"
    4) (integer) 1
    5) "pending"
    6) (integer) 2
    7) "last-delivered-id"
    8) "2-0"
    9) "entries-read"
   10) (integer) 2
   11) "lag"
   12) (integer) 2
127.0.0.1:6379> xdel mystream 3-0
(integer) 1
127.0.0.1:6379> XINFO GROUPS mystream
1)  1) "name"
    2) "g1"
    3) "consumers"
    4) (integer) 1
    5) "pending"
    6) (integer) 2
    7) "last-delivered-id"
    8) "2-0"
    9) "entries-read"
   10) (integer) 2
   11) "lag"
   12) (nil)
127.0.0.1:6379> XREADGROUP GROUP g1 c1 COUNT 2 STREAMS mystream >
1) 1) "mystream"
   2) 1) 1) "4-0"
         2) 1) "data"
            2) "d"
127.0.0.1:6379> XINFO GROUPS mystream
1)  1) "name"
    2) "g1"
    3) "consumers"
    4) (integer) 1
    5) "pending"
    6) (integer) 3
    7) "last-delivered-id"
    8) "4-0"
    9) "entries-read"
   10) (integer) 3
   11) "lag"
   12) (integer) 1 # it should be 0 the message was deleted and there is not lag
```

**Expected behavior**

Lag should be 0 after the last read

**Additional Information**
The lag calculation is important for monitoring consumer group health and detecting falling behind consumers
and this bug makes it difficult to reliably monitor stream processing progress.

nesty92:
When a tombstone is created after the last_id of a consumer group.

The consumer group lag is reported in the reply of `XINFO GROUPS mystream`.

Close: #1951

### DIFF
--- a/src/t_stream.c
+++ b/src/t_stream.c
@@ -1394,11 +1394,6 @@ int streamRangeHasTombstones(stream *s, streamID *start, streamID *end) {
         return 0;
     }
 
-    if (streamCompareID(&s->first_id, &s->max_deleted_entry_id) > 0) {
-        /* The latest tombstone is before the first entry. */
-        return 0;
-    }
-
     if (start) {
         start_id = *start;
     } else {
@@ -1488,6 +1483,11 @@ long long streamEstimateDistanceFromFirstEverEntry(stream *s, streamID *id) {
      * it can set to the current added_entries value. */
     if (!s->length && streamCompareID(id, &s->last_id) < 1) {
         return s->entries_added;
+    }
+
+    if (!streamIDEqZero(id) && streamCompareID(id, &s->max_deleted_entry_id) < 0) {
+        /* The ID is before the last tombstone, so the counter is unknown. */
+        return SCG_INVALID_ENTRIES_READ;
     }
 
     int cmp_last = streamCompareID(id, &s->last_id);
@@ -1678,10 +1678,11 @@ size_t streamReplyWithRange(client *c,
     while (streamIteratorGetID(&si, &id, &numfields)) {
         /* Update the group last_id if needed. */
         if (group && streamCompareID(&id, &group->last_id) > 0) {
-            if (group->entries_read != SCG_INVALID_ENTRIES_READ && !streamRangeHasTombstones(s, &id, NULL)) {
-                /* A valid counter and no future tombstones mean we can
-                 * increment the read counter to keep tracking the group's
-                 * progress. */
+            if (group->entries_read != SCG_INVALID_ENTRIES_READ &&
+                streamCompareID(&group->last_id, &s->first_id) >= 0 &&
+                !streamRangeHasTombstones(s, &group->last_id, NULL)) {
+                /* A valid counter and no tombstones in the group's last-delivered-id and the stream's last-generated-id,
+                 * we can increment the read counter to keep tracking the group's progress. */
                 group->entries_read++;
             } else if (s->entries_added) {
                 /* The group's counter may be invalid, so we try to obtain it. */
--- a/tests/unit/type/stream-cgroups.tcl
+++ b/tests/unit/type/stream-cgroups.tcl
@@ -1110,6 +1110,25 @@ start_server {
         set group [lindex [dict get $reply groups] 0]
         assert_equal [dict get $group entries-read] 5
         assert_equal [dict get $group lag] 1
+
+        r XADD x 7-0 data g
+        r XADD x 8-0 data h
+        r XADD x 9-0 data i
+        r XADD x 10-0 data j
+        r XREADGROUP GROUP g1 c11 COUNT 3 STREAMS x >
+        r XDEL x 9-0
+        # Now there is a tombstone in the stream after the consumer group last_id 
+        # so the lag can't be calculated
+        set reply [r XINFO STREAM x FULL]
+        set group [lindex [dict get $reply groups] 0]
+        assert_equal [dict get $group entries-read] 8
+        assert_equal [dict get $group lag] {}
+
+        r XREADGROUP GROUP g1 c12 COUNT 1 STREAMS x >
+        set reply [r XINFO STREAM x FULL]
+        set group [lindex [dict get $reply groups] 0]
+        assert_equal [dict get $group entries-read] 10
+        assert_equal [dict get $group lag] 0
     }
 
     test {Consumer group lag with XDELs} {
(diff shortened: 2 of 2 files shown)


zuiderkwast (approved):
LGTM.

@hpatro Shall we backport it?

hpatro:
Haven't read the code. Is this related to https://github.com/valkey-io/valkey/pull/685 ?
 
> @hpatro Shall we backport it?

Do we backport or just make the change in latest version? If we do we would need to get it to 7.2.x, 8.0.x and 8.1.x I believe.

zuiderkwast:
> Is this related to #685 ?

:thinking: 

It looks very similar, yes. @artikell is it a duplicate? Can you take a look and decide which one we can merge?

nesty92:
@zuiderkwast @hpatro #685 try to fix others issues as well and it's more dangerous to handle all of them in a single fix

hpatro:
> @zuiderkwast @hpatro #685 try to fix others issues as well and it's more dangerous to handle all of them in a single fix

Right. I agree we shouldn't conflate the PR. I just want to ensure the correctness. @artikell would you mind reviewing this ?

hpatro (approved):
The overall change looks good to me. @nesty92 Could you share the before and after behavior (preferrable valkey-cli output) ?

artikell:
@zuiderkwast @hpatro Yes, this PR is a simple fix that also aligns redis. There is another fix on redis, and this pr is also dealt with together? @nesty92 

When using the XTRIM command to trim a stream, it does not update the maximal tombstone (max_deleted_entry_id). This leads to an issue where the lag calculation incorrectly assumes that there are no tombstones after the consumer group's last_id, resulting in an inaccurate lag.

nesty92:
> Could you share the before and after behavior (preferrable valkey-cli output) ?

@hpatro the full output is in #1951 description, here is just the last result

```bash
127.0.0.1:6379> XINFO GROUPS mystream
1)  1) "name"
    2) "g1"
    3) "consumers"
    4) (integer) 1
    5) "pending"
    6) (integer) 3
    7) "last-delivered-id"
    8) "4-0"
    9) "entries-read"
   10) (integer) 3
   11) "lag"
   12) (integer) 1 # it should be 0 the message was deleted and there is not lag
```
this is the new output with the fix

```bash
> ./src/valkey-cli
127.0.0.1:6379> xadd mystream 1-0 data a
"1-0"
127.0.0.1:6379> xadd mystream 2-0 data b
"2-0"
127.0.0.1:6379> xadd mystream 3-0 data c
"3-0"
127.0.0.1:6379> xadd mystream 4-0 data d
"4-0"
127.0.0.1:6379> XGROUP CREATE mystream g1 0
OK
127.0.0.1:6379> XREADGROUP GROUP g1 c1 COUNT 2 STREAMS mystream >
1) 1) "mystream"
   2) 1) 1) "1-0"
         2) 1) "data"
            2) "a"
      2) 1) "2-0"
         2) 1) "data"
            2) "b"
127.0.0.1:6379> XINFO GROUPS mystream
1)  1) "name"
    2) "g1"
    3) "consumers"
    4) (integer) 1
    5) "pending"
    6) (integer) 2
    7) "last-delivered-id"
    8) "2-0"
    9) "entries-read"
   10) (integer) 2
   11) "lag"
   12) (integer) 2
127.0.0.1:6379> xdel mystream 3-0
(integer) 1
127.0.0.1:6379> XINFO GROUPS mystream
1)  1) "name"
    2) "g1"
    3) "consumers"
    4) (integer) 1
    5) "pending"
    6) (integer) 2
    7) "last-delivered-id"
    8) "2-0"
    9) "entries-read"
   10) (integer) 2
   11) "lag"
   12) (nil)
127.0.0.1:6379> XREADGROUP GROUP g1 c1 COUNT 2 STREAMS mystream >
1) 1) "mystream"
   2) 1) 1) "4-0"
         2) 1) "data"
            2) "d"
127.0.0.1:6379> XINFO GROUPS mystream
1)  1) "name"
    2) "g1"
    3) "consumers"
    4) (integer) 1
    5) "pending"
    6) (integer) 3
    7) "last-delivered-id"
    8) "4-0"
    9) "entries-read"
   10) (integer) 4
   11) "lag"
   12) (integer) 0
127.0.0.1:6379> 
```

nesty92:
> When using the XTRIM command to trim a stream, it does not update the maximal tombstone (max_deleted_entry_id). This leads to an issue where the lag calculation incorrectly assumes that there are no tombstones after the consumer group's last_id, resulting in an inaccurate lag.

@artikell Oh, I didn't know about this one, but yes you are right I was able to replicate the issue with XTRIM but not sure if we should handle it together. What do you think @hpatro @zuiderkwast @artikell ?

It's my first time trying to contribute here so any feedback is more than welcome

zuiderkwast:
> It's my first time trying to contribute here so any feedback is more than welcome

Yes, thank you, it's great!

I am not very familiar with consumer groups logic so I need some help and advice. I don't care which PR we will merge, as long as it fixes the problems.

If the `max_deleted_entry_id` problem can still cause an incorrect lag, then we can't say this PR fixes the incorrect lag... It's better if we can fix this too. Is it easy?

#685 doesn't affect `max_deleted_entry_id` but it solves these problems by calculating the lag in a different way: by correctly calculating entries-read and then calculates lag from entries-read and entries-added. Is there a risk that this new calculation of lag is wrong or is the concern that PR also changes the output for entries-read for empty streams from NULL to a correct number?

nesty92:
@zuiderkwast @artikell I included a fix for the XTRIM, left the details also in the commit message

The issue arises when `XTRIM` or `XADD` removes entries from the beginning of the stream. If the trimming process removes the next entry after `max_deleted_entry_id` (and all entries before it), the stored ID becomes stale, it refers to a point *before* the stream's current start.

This fix updates the trimming logic to reset `max_deleted_entry_id` whenever the trim operation results in a new first entry ID that is later than the existing `max_deleted_entry_id`. This ensures the tombstone accurately reflects deletions *within* the stream's valid range.

In this way we are not changing the behavior of the lag all the existing test continue passing and a new one was added.

note: the problem with the XTRIM, could be also with XADD if trimming is set. That case is cover as well

src/t_stream.c
```diff
@@ -1678,10 +1684,9 @@ size_t streamReplyWithRange(client *c,
     while (streamIteratorGetID(&si, &id, &numfields)) {
         /* Update the group last_id if needed. */
         if (group && streamCompareID(&id, &group->last_id) > 0) {
-            if (group->entries_read != SCG_INVALID_ENTRIES_READ && !streamRangeHasTombstones(s, &id, NULL)) {
-                /* A valid counter and no future tombstones mean we can
-                 * increment the read counter to keep tracking the group's
-                 * progress. */
+            if (group->entries_read != SCG_INVALID_ENTRIES_READ && streamCompareID(&group->last_id, &s->first_id) >= 0 && !streamRangeHasTombstones(s, &group->last_id, NULL)) {
```
artikell: streamCompareID(&group->last_id, &s->first_id) >= 0

This judgment will result in most of them failing to perform, ultimately causing performance degradation.

Causes most requests to recalculate the distance.

nesty92: On a healthy workflow that should not impact the performance and group->last_id would indeed be >= s->first_id. It would only be false if a group falls significantly behind and aggressive trimming occurs. So, whether it's "most" requests is debatable and depends heavily on the specific workload.

artikell: Granted, that's not most requests. It seems acceptable.

src/t_stream.c
```diff
@@ -860,6 +860,12 @@ int64_t streamTrim(stream *s, streamAddTrimArgs *args) {
         streamGetEdgeID(s, 1, 1, &s->first_id);
     }
 
+    /* Reset the stream's maximal tombstone if needed. */
+    if (streamCompareID(&s->first_id, &s->max_deleted_entry_id) > 0) {
```
artikell: The xdel command also needs to do this. Because first_id will also be updated there.

artikell: as @zuiderkwast  said, causes the loss of max_deleted_entry_id and introduces incompatible changes.

You can modify the `streamRangeHasTombstones` and `streamEstimateDistanceFromFirstEverEntry` method instead of modifying this variable to maintain compatibility.

nesty92: That is a good point @artikell, actually we don't need to update the max_deleted_entry_id, what is actually important is to check if the last_id of the consumer is still in the range of the stream `streamCompareID(&group->last_id, &s->first_id) >= 0`, otherwise the stream was trimmed and the counter is not longer valid and need to be calculated from the start of the stream.

Like we mention before this will only be false if the consumer is falling behind the trim all the time forcing the calculation from the start.

nesty92: The reset of the max_deleted_entry_id, could be latter introduce as a breaking change for a next version @zuiderkwast

zuiderkwast: > Like we mention before this will only be false if the consumer is falling behind the trim all the time forcing the calculation from the start.

I don't understand exactly. Is the behavior correct now but it will be inefficient in some situations when it does the calculation from the start?

> The reset of the max_deleted_entry_id, could be latter introduce as a breaking change for a next version

Yes, we can merge this patch and backport it. @artikell I will wait for your approval.

Then, we can rebase and merge #685 and release it in 9.0, but is it really a breaking change? It is a behavior change but it corrects the behavior to follow the documentation. Therefore, I think it can also be considered a bugfix actually, even the changes of `entries-read`. The documentation says:

       • entries-read: the logical “read counter” of the last entry delivered to the group’s consumers

src/t_stream.c
```diff
@@ -1434,7 +1429,7 @@ void streamReplyWithCGLag(client *c, stream *s, streamCG *cg) {
         /* The lag of a newly-initialized stream is 0. */
         lag = 0;
         valid = 1;
-    } else if (cg->entries_read != SCG_INVALID_ENTRIES_READ && !streamRangeHasTombstones(s, &cg->last_id, NULL)) {
+    } else if (cg->entries_read != SCG_INVALID_ENTRIES_READ && streamCompareID(&cg->last_id, &s->first_id) >= 0 && !streamRangeHasTombstones(s, &cg->last_id, NULL)) {
```
zuiderkwast: It's a very long line. Let's break it.

```suggestion
    } else if (cg->entries_read != SCG_INVALID_ENTRIES_READ &&
               streamCompareID(&cg->last_id, &s->first_id) >= 0 &&
               !streamRangeHasTombstones(s, &cg->last_id, NULL)) {
```

tests/unit/type/stream-cgroups.tcl
```diff
@@ -1177,6 +1196,132 @@ start_server {
...
+        set reply [r XINFO STREAM x FULL]
+        set group [lindex [dict get $reply groups] 0]
+        assert_equal [dict get $group entries-read] 5
+        assert_equal [dict get $group lag] 2
+
+        r XTRIM x MINID = 7-0
+        set reply [r XINFO STREAM x FULL]
+        set group [lindex [dict get $reply groups] 0]
+        assert_equal [dict get $group entries-read] 5
+        assert_equal [dict get $group lag] 1
```
artikell: I think it is necessary to maintain the documentation description in xinfo：
```
Valkey reports the lag of a consumer group by keeping two counters: the number of all entries added to the stream and the number of logical reads made by the consumer group. The lag is the difference between these two.
```
It now appears that lag is the wrong value.

Since the entries-read returned by the xinfo method is not the same as the calculated one, it is necessary to make it the same.

nesty92: @artikell good point, let's keep XINFO out of this fix

src/t_stream.c
nesty92: > I don't understand exactly. Is the behavior correct now but it will be inefficient in some situations when it does the calculation from the start?

I was referring to [the previous performance concern from @artikell about this.](https://github.com/valkey-io/valkey/pull/1952#discussion_r2047016292)

 But it won't be a problem.

artikell (approved):
LGTM

zuiderkwast (approved):

nesty92:
Thanks for the help @zuiderkwast @hpatro @artikell

PR #1961, opened 2025-04-16
Introduce MPTCP for replica

pizhenwei:
Allow replicas to use MPTCP in the outgoing replication connection.

A new yes/no config is introduced `repl-mptcp`, default `no`.

For MPTCP to be used in replication, the primary needs to be configured with `mptcp yes` and the replica with `repl-mptcp yes`. Otherwise, the connection falls back to regular TCP.

Follow-up of #1811.

Cc Linux kernel MPTCP maintainer @matttbe @geliangtang @Dwyane-Yan
Cc @xbasel

### DIFF
--- a/src/anet.h
+++ b/src/anet.h
@@ -52,7 +52,7 @@
 #endif
 
 int anetTcpNonBlockConnect(char *err, const char *addr, int port);
-int anetTcpNonBlockBestEffortBindConnect(char *err, const char *addr, int port, const char *source_addr);
+int anetTcpNonBlockBestEffortBindConnect(char *err, const char *addr, int port, const char *source_addr, int multipath);
 int anetResolve(char *err, char *host, char *ipbuf, size_t ipbuf_len, int flags);
 int anetTcpServer(char *err, int port, char *bindaddr, int backlog, int mptcp);
 int anetTcp6Server(char *err, int port, char *bindaddr, int backlog, int mptcp);
--- a/src/anet.h
+++ b/src/anet.h
@@ -52,7 +52,7 @@
 #endif
 
 int anetTcpNonBlockConnect(char *err, const char *addr, int port);
-int anetTcpNonBlockBestEffortBindConnect(char *err, const char *addr, int port, const char *source_addr, int multipath);
+int anetTcpNonBlockBestEffortBindConnect(char *err, const char *addr, int port, const char *source_addr, int mptcp);
 int anetResolve(char *err, char *host, char *ipbuf, size_t ipbuf_len, int flags);
 int anetTcpServer(char *err, int port, char *bindaddr, int backlog, int mptcp);
 int anetTcp6Server(char *err, int port, char *bindaddr, int backlog, int mptcp);
--- a/src/config.c
+++ b/src/config.c
@@ -3167,6 +3167,7 @@ standardConfig static_configs[] = {
     createBoolConfig("lazyfree-lazy-user-del", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.lazyfree_lazy_user_del, 1, NULL, NULL),
     createBoolConfig("lazyfree-lazy-user-flush", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.lazyfree_lazy_user_flush, 1, NULL, NULL),
     createBoolConfig("repl-disable-tcp-nodelay", NULL, MODIFIABLE_CONFIG, server.repl_disable_tcp_nodelay, 0, NULL, NULL),
+    createBoolConfig("repl-mptcp", NULL, IMMUTABLE_CONFIG, server.repl_mptcp, 0, isValidMptcp, NULL),
     createBoolConfig("repl-diskless-sync", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.repl_diskless_sync, 1, NULL, NULL),
     createBoolConfig("dual-channel-replication-enabled", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.dual_channel_replication, 0, NULL, NULL),
     createBoolConfig("aof-rewrite-incremental-fsync", NULL, MODIFIABLE_CONFIG, server.aof_rewrite_incremental_fsync, 1, NULL, NULL),
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -1184,10 +1184,16 @@ static int connRdmaConnect(connection *conn,
                            const char *addr,
                            int port,
                            const char *src_addr,
+                           int multipath,
                            ConnectionCallbackFunc connect_handler) {
     rdma_connection *rdma_conn = (rdma_connection *)conn;
     struct rdma_cm_id *cm_id;
     RdmaContext *ctx;
+
+    if (multipath) {
+        serverLog(LL_NOTICE, "RDMA: multipath not supported");
+        return C_ERR;
+    }
 
     if (rdmaResolveAddr(rdma_conn, addr, port, src_addr) == C_ERR) {
         return C_ERR;
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -1190,9 +1190,8 @@ static int connRdmaConnect(connection *conn,
     struct rdma_cm_id *cm_id;
     RdmaContext *ctx;
 
-    if (multipath) {
-        serverLog(LL_NOTICE, "RDMA: multipath not supported, ignore");
-    }
+    /* RDMA does not support multipath, and there is no outgoing RDMA connection at the current stage */
+    assert(!multipath);
 
     if (rdmaResolveAddr(rdma_conn, addr, port, src_addr) == C_ERR) {
         return C_ERR;
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -1191,8 +1191,7 @@ static int connRdmaConnect(connection *conn,
     RdmaContext *ctx;
 
     if (multipath) {
-        serverLog(LL_NOTICE, "RDMA: multipath not supported");
-        return C_ERR;
+        serverLog(LL_NOTICE, "RDMA: multipath not supported, ignore");
     }
 
     if (rdmaResolveAddr(rdma_conn, addr, port, src_addr) == C_ERR) {
--- a/valkey.conf
+++ b/valkey.conf
@@ -830,21 +830,12 @@ dual-channel-replication-enabled no
 # be a good idea.
 repl-disable-tcp-nodelay no
 
-# Enables MPTCP for the replica's replication connection.
+# Enables MPTCP for the replica's connection to the primary.
 #
-# An MPTCP connection is established between the primary and the replica if and
-# only if the replica has set 'repl-mptcp yes' and the primary has 'mptcp yes'.
+# An MPTCP connection is established between the primary and the replica if
+# the replica has set 'repl-mptcp yes' and the primary has set 'mptcp yes'.
 # Otherwise, it will automatically and implicitly fall back to a regular TCP
 # connection.
-# When mptcp and repl-mptcp are used in combination, the protocol negotiation
-# will be carried out in accordance with the following table:
-# +----------------+-----------+----------+
-# |                | mptcp yes | mptcp no |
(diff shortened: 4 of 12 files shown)


valkey.conf
```diff
@@ -830,6 +830,17 @@ dual-channel-replication-enabled no
...
+# Multipath TCP (MPTCP) for the replication
+#
+# There is a TCP/MPTCP negotiation metric between server and the replication:
+# +-------------+----------+------------+
+# |             |server TCP|server MPTCP|
+# | replica TCP |   TCP    |    TCP     |
+# |replica MPTCP|   TCP    |   MPTCP    |
+# +-------------+----------+------------+
+#
+# repl-mptcp yes
```
xbasel: I'm not sure I understand the description above. Maybe this?

```
# Enables MPTCP for the replica's replication connection.
#
# Requires 'mptcp yes' to be set in the primary.
#
# Default: no
# repl-mptcp no
```

src/anet.h
```diff
@@ -52,7 +52,7 @@
 #endif
 
 int anetTcpNonBlockConnect(char *err, const char *addr, int port);
-int anetTcpNonBlockBestEffortBindConnect(char *err, const char *addr, int port, const char *source_addr);
+int anetTcpNonBlockBestEffortBindConnect(char *err, const char *addr, int port, const char *source_addr, int multipath);
```
xbasel: In some places we call the parameter `mptcp`, and in others `multipath`. Let's unify the naming.

src/config.c
```diff
@@ -3167,6 +3167,7 @@ standardConfig static_configs[] = {
     createBoolConfig("lazyfree-lazy-user-del", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.lazyfree_lazy_user_del, 1, NULL, NULL),
     createBoolConfig("lazyfree-lazy-user-flush", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.lazyfree_lazy_user_flush, 1, NULL, NULL),
     createBoolConfig("repl-disable-tcp-nodelay", NULL, MODIFIABLE_CONFIG, server.repl_disable_tcp_nodelay, 0, NULL, NULL),
+    createBoolConfig("repl-mptcp", NULL, IMMUTABLE_CONFIG, server.repl_mptcp, 0, isValidMptcp, NULL),
```
xbasel, line 3170: Does this config really need to be immutable?

src/rdma.c
```diff
@@ -1184,11 +1184,17 @@ static int connRdmaConnect(connection *conn,
...
                            ConnectionCallbackFunc connect_handler) {
     rdma_connection *rdma_conn = (rdma_connection *)conn;
     struct rdma_cm_id *cm_id;
     RdmaContext *ctx;
 
+    if (multipath) {
+        serverLog(LL_NOTICE, "RDMA: multipath not supported");
+        return C_ERR;
+    }
+
```
xbasel: Not familiar with RDMA.

But can  RDMA run alongside standard TCP ? If yes, maybe we don't want to throw an error and just ignore it.

src/anet.h
pizhenwei: Sure, `mptcp` is better for this function.
I suppose the generic rule: `mptcp` for TCP level, `multipath` for generic connection level.

src/rdma.c
pizhenwei: Yes, RDMA run alongside TCP. Currently there is no config to enable RDMA multipath, maybe assert here?

src/config.c
pizhenwei: I suppose it's immutable. there is no reason to change it dynamically.

valkey.conf
pizhenwei: Please take a look at the latest version.

src/rdma.c
zuiderkwast: Or ignore? If users want to enable RDMA and MPTCP.

pizhenwei: Done.

pizhenwei:
@zuiderkwast @xbasel What do you think about the latest version?

zuiderkwast:
I haven't looked yet. We are working on patch releases now. I will take a look in a few days.

valkey.conf
```diff
@@ -830,6 +830,25 @@ dual-channel-replication-enabled no
 # be a good idea.
 repl-disable-tcp-nodelay no
 
+# Enables MPTCP for the replica's replication connection.
+#
+# An MPTCP connection is established between the server and the replica if and
+# only if both the server and the replica use MPTCP simultaneously. Otherwise,
+# it will automatically fall back to a regular TCP connection. Note that the
+# negotiation between MPTCP and TCP is implicit, and there will be no explicit
+# connection errors.
```
zuiderkwast: Terminology: They are called *primary* and *replica*. Both of them are called *servers* in the sense that they can both accept client connections.

I think we can make it more clear that the replica needs to set 'repl-mptcp yes' and the primary needs to set 'mptcp yes'.

The note about no connection error can be shorter. If we say that this negotiation is automatic and with the implicit fallback, then I guess we don't need to mention that there will be no connection errors.

Idea:

```suggestion
# An MPTCP connection is established between the primary and the replica if and
# only if the replica has set 'repl-mptcp yes' and the primary has 'mptcp yes'.
# Otherwise, it will automatically and implicitly fall back to a regular TCP
# connection.
```

valkey.conf
```diff
@@ -830,6 +830,25 @@ dual-channel-replication-enabled no
...
+# will be carried out in accordance with the following table:
+# +----------------+-----------+----------+
+# |                | mptcp yes | mptcp no |
+# +----------------+-----------+----------+
+# | repl-mptcp yes |   TCP     |    TCP   |
+# +----------------+-----------+----------+
+# | repl-mptcp no  |   TCP     |   MPTCP  |
+# +----------------+-----------+----------+
+#
+# repl-mptcp yes
```
zuiderkwast: Perhaps we should say 'no' here. It is the default and also the recommendation for most users. Or shall we mention in the text that the default is 'no'? What do you think?

```suggestion
# repl-mptcp no
```

src/rdma.c
```diff
@@ -1184,11 +1184,16 @@ static int connRdmaConnect(connection *conn,
...
                            int port,
                            const char *src_addr,
+                           int multipath,
                            ConnectionCallbackFunc connect_handler) {
     rdma_connection *rdma_conn = (rdma_connection *)conn;
     struct rdma_cm_id *cm_id;
     RdmaContext *ctx;
 
+    if (multipath) {
+        serverLog(LL_NOTICE, "RDMA: multipath not supported, ignore");
```
zuiderkwast: I'm trying to understand what's best and the most logical choice here. We have three options:

1. Error, `assert(!multipath)`
2. Log notice
3. Silent

Do we use RDMA for outgoing connections at all?

I'm trying to think of some kind of future scenario where we use RDMA for some outgoing connections and MPTCP for some other connections and what should happen in this case.

For example, if we allow RDMA for replication, then we should reject the configuration `repl-mptcp yes` in the config validation. It means we still can't get to this case. If it can't happen, then maybe an assert is better?

zuiderkwast (commented):
I think this is mostly fine. I have only some minor comments.

src/rdma.c
pizhenwei: From the point of my view, assert is better at the current stage.

pizhenwei:
Apply @zuiderkwast 's suggestions, force push a commit over 3 commits to fix DCO error.

src/config.c
zuiderkwast: If we make it mutable and users change it when replication is already connected, it will not have any effect. Only if the replica needs to reconnect some time later it will be able to use MPTCP, but this may never happen. Maybe that can be confusing to users.

If there is a bad network and there are many reconnects, users can enable MTPCP to fix the problem?

I'm just guessing pros and cons of mutable vs immutable. I'm fine with keeping it immutable.

zuiderkwast (approved):

zuiderkwast:
@valkey-io/core-team This is the follow-up of MTPCP to allow replication to use MPTCP. I think we already decided about this when we discussed and decided about MPTCP in the meeting some weeks ago. Please comment if you disagree.

valkey.conf
```diff
@@ -830,6 +830,24 @@ dual-channel-replication-enabled no
...
+# Otherwise, it will automatically and implicitly fall back to a regular TCP
+# connection.
+# When mptcp and repl-mptcp are used in combination, the protocol negotiation
+# will be carried out in accordance with the following table:
+# +----------------+-----------+----------+
+# |                | mptcp yes | mptcp no |
+# +----------------+-----------+----------+
+# | repl-mptcp yes |   TCP     |    TCP   |
+# +----------------+-----------+----------+
+# | repl-mptcp no  |   TCP     |   MPTCP  |
```
xbasel: when both `repl-mptcp` and `mptcp` are set yo yes, TCP will be used? ,  also when both are "no" , mptcp will be used?

zuiderkwast:
```suggestion
# | repl-mptcp yes |  MPTCP    |    TCP   |
# +----------------+-----------+----------+
# | repl-mptcp no  |   TCP     |    TCP   |
```

pizhenwei: @xbasel @zuiderkwast Thank you!

Done.

valkey.conf
```diff
@@ -830,6 +830,24 @@ dual-channel-replication-enabled no
...
+# connection.
+# When mptcp and repl-mptcp are used in combination, the protocol negotiation
+# will be carried out in accordance with the following table:
+# +----------------+-----------+----------+
+# |                | mptcp yes | mptcp no |
+# +----------------+-----------+----------+
+# | repl-mptcp yes |   MPTCP   |    TCP   |
+# +----------------+-----------+----------+
+# | repl-mptcp no  |    TCP    |    TCP   |
+# +----------------+-----------+----------+
```
xbasel: I still think this is too verbose. Users enabling MPTCP are likely already familiar with MPTCP negotiation and fallback behavior. IMO no need to restate protocol-level detail here.

I'd suggest this:
```
# Enables MPTCP for the replica's connection to the primary.
# Requires 'mptcp yes' on the primary.
repl-mptcp no
```

xbasel: @zuiderkwast what do you think about @https://github.com/valkey-io/valkey/pull/1961#discussion_r2055915241 ?

zuiderkwast: @xbasel Yeah, it's a little verbose. But your suggestion is very short. We should make it clear that it's not dangerous to use this even if the primary might have 'no' configured, for example during rolling upgrade with manual failovers. If we say "Requires 'mptcp yes' on the primary", it sounds like it's forbidden and maybe dangerous to use 'no' on the primary and 'yes' on the replica.

I think some compromise can be good. For example, we can keep the first 4 lines and skip the table. How about that?

```
# Enables MPTCP for the replica's connection to the primary.
#
# An MPTCP connection is established between the primary and the replica if
# the replica has set 'repl-mptcp yes' and the primary has set 'mptcp yes'.
# Otherwise, it will automatically and implicitly fall back to a regular TCP
# connection.
#
# repl-mptcl no
```

pizhenwei: > @xbasel Yeah, it's a little verbose. But your suggestion is very short. We should make it clear that it's not dangerous to use this even if the primary might have 'no' configured, for example during rolling upgrade with manual failovers. If we say "Requires 'mptcp yes' on the primary", it sounds like it's forbidden and maybe dangerous to use 'no' on the primary and 'yes' on the replica.
> 
> I think some compromise can be good. For example, we can keep the first 4 lines and skip the table. How about that?
> 
> ```
> # Enables MPTCP for the replica's connection to the primary.
> #
> # An MPTCP connection is established between the primary and the replica if
> # the replica has set 'repl-mptcp yes' and the primary has set 'mptcp yes'.
> # Otherwise, it will automatically and implicitly fall back to a regular TCP
> # connection.
> #
> # repl-mptcl no
> ```

From the point of my view, this version is quite good, push a commit as this.

xbasel (approved):

Neustradamus:
MPTCP is really good!

Issue #1963, opened 2025-04-16
[NEW] A more safe and efficient way of Sentinel failover, avoiding full resynchronization and data loss.


li-benson:
**The problem/use-case that the feature addresses**

We use Sentinel to monitor the primary and replica nodes to ensure the high availability of the cluster. In some cases, we need to conduct an active failover through Sentinel. However, in practice, it has been found that performing the primary-replica switch in this current way will lead to some data loss and full resynchronization. We hope to minimize this impact as much as possible: avoiding full resynchronization and write loss.

The schematic diagram of the architecture is as follows:

```mermaid
---
config:
  theme: default
---
flowchart LR
 subgraph subGraph0["Sentinel Cluster"]
        S("Sentinel 1 - 3")
  end
 subgraph subGraph1["Valkey Primary-Replica Architecture"]
        M("Primary 1 - n")
        Sla("Replica 1 - n")
  end
    M -- "Primary-Replica Replication" --> Sla
    S -- Monitoring --> M
    C("Client") -- "Listening +switch-master" --> S
    C -- Connecting --> M
     S:::process
     M:::process
     Sla:::process
     C:::process
    classDef process fill:#E5F6FF,stroke:#73A6FF,stroke-width:2px

```

**Description of the feature**

We hope that when performing a manual primary-replica switch via the ``sentinel failover`` command, a less costly and safer way can be provided to complete the switch, so as to avoid write loss and full resynchronization (write failures are allowed).

**Alternatives you've considered**

After investigation, we found that Valkey provided the ``failover`` command in version 6.2.0, which can avoid the problem of full resynchronization. However, Sentinel does not use this command. Therefore, we first tried to actively send the failover command to Valkey primary, allowing Sentinel to passively sense the primary-replica switch. Through testing, it takes about 42 seconds (down_after_period + 2 * info cycle) from the moment when the user sends the ``failover`` command to the primary until Sentinel senses and sends the ``+switch-master`` message. This means that the user will lose 42 seconds of data. The entire process takes too long and is unacceptable.

In addition, we also tried to integrate the failover command into Sentinel, just as the modification of this feature. That is, when a user connects to Sentinel and manually executes the failover command, Sentinel will send the ``failover`` command to the primary instead of the previous ``slaveof no one``. This will solve the problem of full resynchronization after the switch. However, through testing, we found that users would still write data to the old primary, resulting in data inconsistency between the primary and replicas. Moreover, data would still be lost after the client switched to the new primary.
[...]

PR #1964, opened 2025-04-16
Add sentinel failover SAFE option

li-benson:
Implementation of the feature https://github.com/valkey-io/valkey/issues/1963

When performing a manual primary-replica switch via the sentinel failover command, a less costly and safer way can be provided to complete the switch, so as to avoid write loss and full resynchronization (write failures are allowed).


gmbnomis:
I think this is a similar approach to the one proposed by @enjoy-binbin in #1238, which was closed in favor of #1292 (which is authored by me, issue with more details at #1291). #1292 is in approved state since beginning of January, but has not moved forward since.

The short-comings with the approach proposed here are the same as mentioned in the discusion of #1238, see https://github.com/valkey-io/valkey/pull/1238#issuecomment-2460930795. (All of the issues mentioned in this thread have been addressed by #1292 by now.)

The main problem is that the approach violates the specified Sentinel client protocol, which mandates disconnecting clients after a role change.

Of course, this is not optimal for all clients. But by now, clients can indicate their read preferences using REDIRECT mode.  Thus. the proper solution IMHO is to modify Sentinel/the Sentinel protocol to not kill clients in REDIRECT mode (on top of #1292). @soloestoy, wdyt?

li-benson:
> I think this is a similar approach to the one proposed by @enjoy-binbin in #1238, which was closed in favor of #1292 (which is authored by me, issue with more details at #1291). #1292 is in approved state since beginning of January, but has not moved forward since.
> 
> The short-comings with the approach proposed here are the same as mentioned in the discusion of #1238, see [#1238 (comment)](https://github.com/valkey-io/valkey/pull/1238#issuecomment-2460930795). (All of the issues mentioned in this thread have been addressed by #1292 by now.)
> 
> The main problem is that the approach violates the specified Sentinel client protocol, which mandates disconnecting clients after a role change.
> 
> Of course, this is not optimal for all clients. But by now, clients can indicate their read preferences using REDIRECT mode. Thus. the proper solution IMHO is to modify Sentinel/the Sentinel protocol to not kill clients in REDIRECT mode (on top of #1292). @soloestoy, wdyt?



> #1292

Hello, thank you for your attention to this issue. I carefully looked at the previous sentinel code and it seems that there is no operation to actively disconnect the client connection of primary. Only when sending "replcaof no one" is the client connection of the replica closed. In my opinion, since the previous sentinel did not actively close the client connection of primary, we should not add this behavior in the new changes. It seems a bit too violent.

gmbnomis:
> ... I carefully looked at the previous sentinel code and it seems that there is no operation to actively disconnect the client connection of primary. Only when sending "replcaof no one" is the client connection of the replica closed.

"replica of no one" is just a special case in `sentinelSendReplicaOf` (`addr` is `NULL` is this case). *All* calls to `sentinelSendReplicaOf` will kill the client connections, there is no case distinction here. Thus, changing a primary into a replica will also kill the client connections.

This is also specified in the [Sentinel client spec](https://valkey.io/topics/sentinel-clients/#sentinel-failover-disconnection) (emphasis by me):

When Valkey Sentinel changes the configuration of an instance, for example promoting a replica to a primary, **demoting a primary to replicate to the new primary after a failover**, or simply changing the primary address of a stale replica instance, it sends a CLIENT KILL type normal command to the instance in order to make sure all the clients are disconnected from the reconfigured instance. This will force clients to resolve the primary address again.

> In my opinion, since the previous sentinel did not actively close the client connection of primary, we should not add this behavior in the new changes. It seems a bit too violent.

As said, this behavior is required by the specification and is currently implemented. The point of the Sentinel client protocol is that a client can be sure (after initially checking the role in a connection)  that the role will never change in this connection. Thus, one cannot assume that client libraries/client applications will be prepared to handle such a role switch in general.

hwware:
Overall, it looks good, but I have 2 concerns:

1. The command format:  SENTINEL FAILOVER <primary-name> [SAFE [timeout]] @valkey-io/client-maintainers The keyword "SAFE" is a little bit tricky, I do not think there are similar keywords in other command, do you support it or have better idea?

2.  I check the test case, client need to make the replica-read-only as yes. Thus, from my understanding, if client forget to set it and keep it as no, then safe failover is not safe. My suggestion is if with SRI_SAFE_FAILOVER flag, we can set this flag in the codes instead of depending on "config set replica-read-only yes"

gmbnomis:
@hwware I think that #1292 and this PR are mutually exclusive since they both intend to implement smooth failover for sentinel but they handle clients very differently. 

Accepting this PR requires a change of the sentinel client protocol specification. Given the potential impact, could we consider adding a "major-decision-pending" label to this PR? 

Just my 2 cents (if that is not already clear): From my perspective, this change does not have a clear justification and will introduce unnecessary incompatibility.

PR #1729, opened 2025-02-13
Use tcl8 on fedora rawhide

zuiderkwast:
Apparently TCL 8.x exists on Fedora Rawhide under the name tcl8.

@jonathanspw

PR #1965, opened 2025-04-16
Fix CI latest fedora: install awk and tcl8, build tcltls from source

vitahlin:
Fix two problems in fedora CI jobs:

1. Install awk where missing. It's required for building jemalloc.
2. Fix problems with TCL, required for running the tests. Fedora comes with
   TCL 9 by default, but the TLS package 'tcltls' isn't built for TCL 9.
   Install 'tcl8' and build 'tcltls' from source.


.github/actions/fedora-build-jemalloc/action.yml
```diff
@@ -0,0 +1,20 @@
...
+    - run: |
+        dnf -y install autoconf make gcc awk curl unzip
+        curl -L -o jemalloc-5.3.0.zip https://github.com/jemalloc/jemalloc/archive/refs/tags/5.3.0.zip
+        unzip jemalloc-5.3.0.zip
+        cd jemalloc-5.3.0
+        ./autogen.sh
+        ./configure --prefix=/opt/jemalloc --enable-jemalloc-prefix
+        make -j$(nproc)
+        make install
+      shell: bash
```
madolson: We include a vendored jemalloc so we don't need to download it. Can we not just alter the make file to pass these options during the building of our vendored one?

vitahlin: 🥲 Sorry for using a more complicated solution earlier. I’ve found that it’s actually sufficient to just install awk properly on Fedora.

vitahlin:
This is the latest CI result: https://github.com/vitahlin/valkey/actions/runs/14507811027

.github/workflows/daily.yml
```diff
@@ -876,6 +876,9 @@ jobs:
         run: dnf -y install epel-release
       - name: make
         run: |
+          if [[ "${{ matrix.name }}" == "test-fedoralatest-jemalloc" ]]; then
+            dnf -y install awk
+          fi
           dnf -y install gcc make procps-ng which /usr/bin/kill
```
zuiderkwast: Can we just add awk to the list of packages to install? In my fedora installation (not latest), the package is called `gawk` but it should also work to specify the path of the binary you want, like `/usr/bin/awk`.

```suggestion
          dnf -y install gcc make procps-ng which /usr/bin/kill /usr/bin/awk
```

If the above doesn't work, then can we can check if awk exists instead of matching the hard-coded name of the CI job, such as like below?

```
          which awk >/dev/null || dnf -y install awk
```

vitahlin: I tested on Fedora 41 and 42 using Docker. Fedora 41 has awk, but Fedora 42 has neither awk nor gawk. I think this is why the CI is failing now, as the latest Fedora has removed awk, and we need to manually install it.

Yeah, you’re absolutely right — the command `which awk >/dev/null || dnf -y install awk` is indeed more user-friendly.

zuiderkwast (approved):

vitahlin:
Latest CI result: https://github.com/vitahlin/valkey/actions/runs/14518932113

.github/workflows/daily.yml
zuiderkwast: I was talking about the package name. On Fedora 40 there is no package `awk` but the package name is `gawk` and it provides `/usr/bin/awk` and `/bin/awk`. I was thinking that we can install `/usr/bin/awk` it would work on any version, and if it is already installed, it is a no-op.

zuiderkwast:
I see in your Daily run, `which awk` displays an error. Maybe we should redirect stderr instead of stdout. And the package name seems to be gawk. The package awk is just an alias of gawk.

```
which: no awk in (/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin)
Updating and loading repositories:
Repositories loaded.
Total size of inbound packages is 1 MiB. Need to download 1 MiB.
After this operation, 3 MiB extra will be used (install 3 MiB, remove 0 B).
Package             Arch   Version      Repository      Size
Installing:
 gawk               x86_64 5.3.1-1.fc42 fedora       1.7 MiB
```

Maybe we can install the package `gawk` on all fedora versions? Then we don't need to check if it's already installed.

.github/workflows/daily.yml
```diff
@@ -877,6 +877,7 @@ jobs:
       - name: make
         run: |
           dnf -y install gcc make procps-ng which /usr/bin/kill
+          which awk >/dev/null || dnf -y install awk
```
zuiderkwast:
```suggestion
          dnf -y install gcc make procps-ng which /usr/bin/kill /usr/bin/awk
```

.github/workflows/daily.yml
```diff
@@ -943,6 +944,7 @@ jobs:
       - name: make
         run: |
           dnf -y install make gcc openssl-devel openssl procps-ng which /usr/bin/kill
+          which awk >/dev/null || dnf -y install awk
```
zuiderkwast:
```suggestion
          dnf -y install make gcc openssl-devel openssl procps-ng which /usr/bin/kill  /usr/bin/awk
```

.github/workflows/daily.yml
```diff
@@ -1015,6 +1017,7 @@ jobs:
...
```
zuiderkwast:
```suggestion
          dnf -y install make gcc openssl-devel openssl procps-ng which /usr/bin/kill  /usr/bin/awk
```

zuiderkwast (commented):
Can we just try installing awk in all Fedora systems and if it's already installed, it does nothing?

I added the `extra-tests` label which should run trigger these builds on every change.

zuiderkwast:
That seemed to work, but now it fails to start the tests. TCL 8.x is not available. Fedora 42 was just released days ago?

https://github.com/valkey-io/valkey/actions/runs/14519381245/job/40736301592?pr=1965

Is this what we observed earlier in the Fedora Rawhide builds? See

* #1769
* #1729

vitahlin:
Yeah, Fedora 42 was released on April 15.

vitahlin:
It seems to be the same issue. I will test Tcl in Fedora 42 later.

zuiderkwast:
I think Fedora changed to TCL 9, which is a new release of TCL. It works without TLS and I did the changes before here:

* https://github.com/valkey-io/valkey/pull/1673

However, for TLS, the `tcltls` package still requires TCL 8.x which doesn't exist... The packaging is broken, at least it was last time i checked.

zuiderkwast:
According to the Change log on this page, `tcltls` is fixed to work with TCL 9: https://packages.fedoraproject.org/pkgs/tcltls/tcltls/fedora-42.html

zuiderkwast:
I updated the [TLS 9.0 PR #1673](https://github.com/valkey-io/valkey/pull/1673) now. Let's see if it works in Fedora 42... No, it seems to still be broken: https://github.com/valkey-io/valkey/actions/runs/14520963361/job/40741560842?pr=1673#step:8:5

> version conflict for package "tcl": have 9.0.0, need 8.4

vitahlin:
With the release of Fedora 42, should we try manually installing Tcl 8 on Fedora 42 first to make sure the daily tests can pass? If it keeps failing, we might miss issues where some code changes behave unexpectedly on the Fedora system.

As for upgrading to Tcl 9, I’ve seen that it introduces breaking changes. Making everything compatible with Tcl 9 could be a much larger effort, so maybe we should keep each PR smaller and limit the scope of changes to make the transition more manageable.

vitahlin:
Added a new GitHub Action that installs TCL 8 from source on Fedora-based environments, avoiding the system’s default TCL 9 installation, latest CI result: https://github.com/vitahlin/valkey/actions/runs/14531492750

zuiderkwast:
In your branch, the tests don't seem to run but in the PR they do.

There is a TCL 8 package called `tcl8` in Fedora, so I hope we can use that instead of building TCL from source.

The problem seems to be the `tcltls` package. It seems to be built in a way that seems to not work with either TCL 9 or TCL 8. I created a small repo for testing this and here is the github action with TCL 8 and 9: https://github.com/zuiderkwast/tcl-tls-test/actions/runs/14523761705/job/40750566767#step:7:9

The problem with `tcltls` seems to be how the package is built. I found this README (https://core.tcl-lang.org/tcltls/file?name=README.txt&ci=trunk) with this information:

```
This package requires TCL 8.5 or later. It will also work with TCL 9, but it is
not binary compatible between major TCL versions. This means if this extension
is built with TCL 8.x it will not load into TCL 9 or vice versa. It is best
to compile both separately then install them with the compatible TCL versions.
```

Maybe we should install the `tcl8` package and then build `tcltls` from source in the TLS builds?

vitahlin:
> There is a TCL 8 package called `tcl8` in Fedora, so I hope we can use that instead of building TCL from source.

You’re right — I was able to successfully install tcl8 on Fedora 42 without building it from source. Thanks for the tip!
```shell
sudo dnf install -y tcl8
ln -s /usr/bin/tclsh8.6 /usr/bin/tclsh
```




> Maybe we should install the tcl8 package and then build tcltls from source in the TLS builds?

Thanks for the information you provided about tcltls. I’m trying to compile tcltls based on the instructions in the documentation, but I’m currently running into some issues with the installation and still working on resolving them.

vitahlin:
Latest CI result: https://github.com/vitahlin/valkey/actions/runs/14587957516
It works well in fedora 42.

Not sure why test-rdma is failing: https://github.com/valkey-io/valkey/actions/runs/14587947974/job/40916792116?pr=1965
image

vitahlin:
Daily CI `reply-schemas-validator` also failed on the unstable branch: https://github.com/valkey-io/valkey/actions/runs/14583803725/job/40905598989

I couldn’t find any record of the failure in job `test-ubuntu-32bit`, but the failures in the two CI jobs are probably unrelated to the changes in this PR

.github/actions/rpm-distros-tcl8/action.yml
```diff
@@ -0,0 +1,31 @@
...
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    # Fedora 42 comes with Tcl 9 by default, but tcltls is currently incompatible with Tcl 9.
+    # As a workaround, we install Tcl 8 and manually build tcltls 1.7.22 from source.
+    # Once tcltls adds support for Tcl 9, this logic can be removed and system packages used instead.
+    - run: |
+        if [[ "${{ inputs.matrix_name }}" =~ "fedora" ]]; then
```
zuiderkwast, line 16: This works because we the jobs we have are "fedoralatest" and "fedorarawhide". If we would have an older fedora job, it wouldn't work. But it's easy enough to change this condition later.

zuiderkwast (approved):
Awesome! Good job!

PR #1966, opened 2025-04-17
Disallowing Client Reply is On / Off / Skip when Client is Multi

### LINKED ISSUE #1268: [BUG] `CLIENT REPLY OFF` in transaction handled incorrectly
**Describe the bug**

Server returns malformed response.
```
127.0.0.1:7000> multi
OK
127.0.0.1:7000(TX)> ping before
QUEUED
127.0.0.1:7000(TX)> client reply off 
QUEUED
127.0.0.1:7000(TX)> ping 1
QUEUED
127.0.0.1:7000(TX)> ping 2
QUEUED
127.0.0.1:7000(TX)> client reply on
QUEUED
127.0.0.1:7000(TX)> ping after
QUEUED
127.0.0.1:7000(TX)> exec
```
Server replies with RESP array of declared size 6, but it contains only 2 elements: "before" and "after". That is correct to receive 2 elements, but size should be declared as 2.
Wireshark helps to track this.

In case if transaction is "reply off; <commands>; reply on; end", received RESP array is empty, but declared size reflects the number of commands.

See Wireshark screenshot:
{5716FB12-9246-4485-ABC3-465C6036E484}

**To reproduce**

See above

**Expected behavior**

Declared RESP array lenght should be valid (match number of elements in it, regardless of real transaction size)

**Additional information**

A low priority issue.

CLI tool also fails to process the response. Use Wireshark to see what happens on the wire.

sarthakaggarwal97:
Return an error when we try to perform `CLIENT REPLY` command under `MULTI/EXEC`.

Without this change, if `CLIENT REPLY` is used in a transaction, it causes `EXEC` to return broken RESP protocol, with a multi-bulk length mismatching the actual number of multi-bulk elements.

Resolves #1268

### DIFF
--- a/src/multi.c
+++ b/src/multi.c
@@ -128,12 +128,6 @@ void flagTransaction(client *c) {
 }
 
 void multiCommand(client *c) {
-    /* If the client has disabled replies, we disallow starting a transaction. */
-    if (c->flag.reply_off == 1 || c->flag.reply_skip == 1) {
-        addReplyError(c, "MULTI not allowed when CLIENT REPLY is off or skip.");
-        return;
-    }
-
     if (!c->mstate) initClientMultiState(c);
     c->flag.multi = 1;
     addReply(c, shared.ok);
--- a/src/multi.c
+++ b/src/multi.c
@@ -128,6 +128,12 @@ void flagTransaction(client *c) {
 }
 
 void multiCommand(client *c) {
+    /* If the client has disabled replies, we disallow starting a transaction. */
+    if (c->flag.reply_off == 1 || c->flag.reply_skip == 1) {
+        addReplyError(c, "MULTI not allowed when CLIENT REPLY is off or skip.");
+        return;
+    }
+
     if (!c->mstate) initClientMultiState(c);
     c->flag.multi = 1;
     addReply(c, shared.ok);
--- a/tests/unit/multi.tcl
+++ b/tests/unit/multi.tcl
@@ -901,6 +901,22 @@ start_server {tags {"multi"}} {
         r flushall
         r ping
      }
+
+    test {MULTI is rejected when CLIENT REPLY is ON/OFF/SKIP} {
+        r multi
+        assert_error "*ERR CLIENT REPLY not allowed during MULTI/EXEC transaction.*" {r client reply on}
+        r exec
+
+        r multi
+        assert_error "*ERR CLIENT REPLY not allowed during MULTI/EXEC transaction.*" {r client reply skip}
+        r exec
+
+        r multi
+        assert_error "*ERR CLIENT REPLY not allowed during MULTI/EXEC transaction.*" {r client reply off}
+        r exec
+
+        r client reply on
+    }
 }
 
 start_server {overrides {appendonly {yes} appendfilename {appendonly.aof} appendfsync always} tags {external:skip}} {
--- a/tests/unit/multi.tcl
+++ b/tests/unit/multi.tcl
@@ -937,8 +937,9 @@ start_server {tags {"multi"}} {
 
         $rd client reply skip
 
+        # Just this command was silenced
         $rd multi
-        # These replies were silenced.
+
         $rd ping pong2
         $rd ping pong3
         $rd exec
--- a/tests/unit/multi.tcl
+++ b/tests/unit/multi.tcl
@@ -925,7 +925,7 @@ start_server {tags {"multi"}} {
         $rd client reply off
 
         $rd multi
-        # These replies were silenced.
+        # These replies were skipped.
         $rd ping pong2
         $rd ping pong3
         $rd exec
@@ -937,7 +937,7 @@ start_server {tags {"multi"}} {
 
         $rd client reply skip
 
-        # Just this command was silenced
+        # Just this command was skipped
         $rd multi
 
         $rd ping pong2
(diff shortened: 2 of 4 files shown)


src/multi.c
```diff
@@ -128,6 +128,12 @@ void flagTransaction(client *c) {
 }
 
 void multiCommand(client *c) {
+    /* If the client has disabled replies, we disallow starting a transaction. */
+    if (c->flag.reply_off == 1 || c->flag.reply_skip == 1) {
+        addReplyError(c, "MULTI not allowed when CLIENT REPLY is off or skip.");
+        return;
+    }
+
```
madolson: I think we can allow this, this should work correctly? We maybe should have a test for this anyways.

sarthakaggarwal97: yeah I need to check how to test with client reply off, it seems like it is in a hang state or command is not finished until timeout.

madolson: You can use a deferring client and not read the response, send the multi, turn it back on and read the response.

sarthakaggarwal97: this makes sense, let me try it out. thank you!

sarthakaggarwal97: @madolson added a test, please take a look!

tests/unit/multi.tcl
```diff
@@ -901,6 +901,59 @@ start_server {tags {"multi"}} {
...
+
+        $rd client reply on
+        $rd ping
+        assert_equal {OK} [$rd read]
+        assert_equal {PONG} [$rd read]
+
+        $rd client reply skip
+
+        $rd multi
+        # These replies were silenced.
```
madolson:
```suggestion
        # Just this command was silenced
        $rd multi
```
I think this just the multi right?

sarthakaggarwal97: ah yes, this change crept in while copying lines, thank you. I have updated!

sarthakaggarwal97:
@madolson this should require a doc change as well right?

madolson (approved):

madolson:
Future note for those that are looking at this. This is a breaking change, but the client protocol is already broken by this change. @valkey-io/core-team just a ping to see if anything has changed since our previous conversations. I'll merge this if I don't hear back.

zuiderkwast:
High-level approved by me.

I didn't immediately understand the problem so I read the issue. Then I edited the top comment to explain what was broken. Please correct it if I got it wrong.

madolson:
@sarthakaggarwal97 Yeah, we should probably update the docs, good point.

tests/unit/multi.tcl
```diff
@@ -901,6 +901,60 @@ start_server {tags {"multi"}} {
...
+        $rd client reply on
+        $rd ping
+        assert_equal {QUEUED} [$rd read]
+        assert_equal {QUEUED} [$rd read]
+        assert_equal {pong2 pong3} [$rd read]
+        assert_equal {OK} [$rd read]
+        assert_equal {PONG} [$rd read]
+
+        $rd close
+    }
```
hpatro: This test is a bit confusing to me. Would prefer parsing response if possible next to the command being executed. And possibly rename comment from silenced to skipped.

```suggestion
    test "CLIENT REPLY OFF/SKIP: multi command" {
        set rd [valkey_deferring_client]

        # Turning the reply off
        $rd client reply off

        $rd multi
        # These replies were skipped.
        $rd ping pong2
        $rd ping pong3
        $rd exec

        $rd client reply on
        assert_equal {OK} [$rd read]
        $rd ping
        assert_equal {PONG} [$rd read]

        $rd client reply skip
        $rd multi
        
        $rd ping pong2
        $rd ping pong3
        $rd exec

        $rd client reply on
        
        # Replies for commands ran within multi/exec.
        assert_equal {QUEUED} [$rd read]
        assert_equal {QUEUED} [$rd read]
        assert_equal {pong2 pong3} [$rd read]

        # reply for `CLIENT REPLY ON`
        assert_equal {OK} [$rd read]

        $rd ping
        assert_equal {PONG} [$rd read]

        $rd close
    }
```
```

**Review by hpatro — APPROVED** [review]

File: tests/unit/multi.tcl (hunk 3 above)
Comment (sarthakaggarwal97) [reply tests/unit/multi.tcl]: reordered the assertions a bit.

File: valkey.conf [hunk 4]
```diff
@@ -1750,7 +1750,7 @@ aof-timestamp-enabled no
 # started as cluster nodes can. In order to start a server instance as a
 # cluster node enable the cluster support uncommenting the following:
 #
-# cluster-enabled yes
+cluster-enabled yes
```
zuiderkwast: Oops

tests/unit/multi.tcl
hpatro: Silenced is not a term we use, could you update it to skipped?

sarthakaggarwal97: there are tests where silenced is used 😅, I have updated it to skipped tho

valkey.conf
sarthakaggarwal97: lol, my bad. removed it! thank you

sarthakaggarwal97:
@madolson I have raised a doc PR as well. Let me know if everything looks good, thank you!

zuiderkwast:
The way CLIENT REPLY is disallowed here is different to other commands that just have the "NO_MULTI" flag in their json file. I created a follow-up to correct the behavior:

* #2152

PR #573, opened 2024-05-29
Avoid shard id update of replica if not matching with primary shard id

hpatro:
During cluster setup, the shard id gets established through extensions data propagation and if the engine crashes/restarts while the reconciliation of shard id is in progress, there is a possibility of corrupted config file and leads to failure of engine restart.

Scenario:

Let's say there are two nodes in a cluster i.e. Node A and Node B. All the admin operation is performed on Node B. Node A and Node B finish handshake and haven't shared the extensions information yet. Node B is made a replica of Node A. As part of Node B sharing the `slaveof` information, it also share(s) the temporary shard-id. During the regular packet processing in Node A, while handling the replication information, the shard id of Node A get(s) applied to Node B. And during the extensions processing in Node A, the shard id passed by Node B is applied which diverges from the shard id of Node A. A crash/restart followed by it leads to unrecoverable corrupted cluster configuration file state.

image

PR #1972, opened 2025-04-18
Fixes for Valkey 8.0.3

murphyjacob4:


zuiderkwast:
This test cases fails:

```
 *** [err]: Busy module name in tests/unit/moduleapi/basics.tcl
expected message not found in log file: *Module name is busy*
```

We should look into it. Same for the 7.2.9 PR.

hpatro:
Shall we get this in as well https://github.com/valkey-io/valkey/pull/573/ for 8.0.x and 8.1.x ?

murphyjacob4:
> expected message not found in log file: *Module name is busy*

Fixed this in both branches now

murphyjacob4:
> Shall we get this in as well https://github.com/valkey-io/valkey/pull/573 for 8.0.x and 8.1.x ?

Added in both - thanks!

00-RELEASENOTES
```diff
@@ -9,6 +9,37 @@ CRITICAL: There is a critical bug affecting MOST USERS. Upgrade ASAP.
...
+
+Upgrade urgency SECURITY: This release includes security fixes we recommend you
+apply as soon as possible.
+
+Bug fixes
+=========
+* Optimize RDB load performance and fix cluster mode resizing on replica side (#1199)
+* Fix memory leak in forgotten node ping ext code path" to 8.0 (#1574) 
+* Fix cluster info sent stats for message with light header (#1563)
+* Use fake client flag to replace not conn check (#1198)
```
zuiderkwast: This doesn't say much to the user. If it fixes an actual problem (does it?), we can mention that, or if it's just refactoring we don't need to mention it.

```Suggestion
```

00-RELEASENOTES
```diff
@@ -9,6 +9,37 @@ CRITICAL: There is a critical bug affecting MOST USERS. Upgrade ASAP.
...
+* Fix module LatencyAddSample still work when latency-monitor-threshold is 0 (#1541)
+* Fix raxRemove crash at memcpy() due to key size exceeds max Rax size (#1722)
```
zuiderkwast: It's difficult to understand what this means to the users.

We can leave it like this for now, because I don't know what to write instead. AFAIU it wasn't reported as a bug, just found during fuzzing.

It doesn't affect streams because streams don't use long key names. ACL? Huge key names are unlikely there too. Maybe the module API "Dict"? It's implemented as a rax. But then we should have reproduced it with a module API test case.

"Fix crash" sounds severe but I think it's very unlikely to happen except possibly in the module API. Maybe we can soften it saying "potential crash"?

In future PRs, we should try using better titles that can be understood by users, not only developers....

00-RELEASENOTES
```diff
@@ -9,6 +9,37 @@ CRITICAL: There is a critical bug affecting MOST USERS. Upgrade ASAP.
...
+* Fix cluster slot stats assertion during promotion of replica (#1950)
+* Fix panic in primary when blocking shutdown after previous block with timeout (#1948)
+* Ignore stale gossip packets that arrive out of order (#1777)
+* Fix flaky shutdown tests caused by timing issue (#1960)
+* Fix incorrect lag reported in XINFO GROUPS (#1952)
+* Avoid shard id update of replica if not matching with primary shard id (#573)
+
+Security fixes
+==============
+* to be added
```
zuiderkwast:
```suggestion
* CVE-2025-21605 Limit output buffer for unauthenticated clients (#1993)
```

zuiderkwast (approved):

PR #1975, opened 2025-04-18
Add DELIFEQ command

LinusU:
Please also pick #2120.

This feature introduces a new command: `DELIFEQ key value`.

The command deletes the given key _only if_ its current value is equal to the provided `value`. It returns:

- 1 if the key was removed (i.e., it existed and matched the value),
- 0 otherwise (key did not exist, or the value did not match).

This command complements `SET key value IFEQ match-value`, added in #1324.

A very common operation when synchronizing distributed locks is to remove a key, but only if the value matches a specific string. This pattern is frequently used in distributed locking algorithms like Redlock. The conditional delete prevents a client from inadvertently releasing a lock it doesn’t own—e.g., when a timeout or retry causes overlapping or stale attempts.

Today, this logic is typically implemented using a small Lua script like:

```lua
if redis.call("get",KEYS[1]) == ARGV[1] then
    return redis.call("del",KEYS[1])
else
    return 0
end
```

However, using scripts adds a layer of complexity and overhead. Having this logic as a built-in Redis command improves simplicity, reliability, and performance. It also complements `SET key IFEQ value` very nicely.

Example:

```redis
SET foo test
DELIFEQ foo test   # returns 1, key is deleted

SET foo nope
DELIFEQ foo test   # returns 0, key remains
```

### DIFF
--- a/src/commands/delifeq.json
+++ b/src/commands/delifeq.json
@@ -1,0 +1,60 @@
+{
+    "DELIFEQ": {
+        "summary": "Delete key if value matches string.",
+        "complexity": "O(1)",
+        "group": "string",
+        "since": "9.0.0",
+        "arity": 3,
+        "function": "delifeqCommand",
+        "command_flags": [
+            "WRITE"
+        ],
+        "acl_categories": [
+            "STRING"
+        ],
+        "key_specs": [
+            {
+                "flags": [
+                    "RM",
+                    "ACCESS",
+                    "DELETE"
+                ],
+                "begin_search": {
+                    "index": {
+                        "pos": 1
+                    }
+                },
+                "find_keys": {
+                    "range": {
+                        "lastkey": 0,
+                        "step": 1,
+                        "limit": 0
+                    }
+                }
+            }
+        ],
+        "reply_schema": {
+            "oneOf": [
+                {
+                    "description": "The key was not deleted.",
+                    "const": 0
+                },
+                {
+                    "description": "The key was deleted.",
+                    "const": 1
+                }
+            ]
+        },
+        "arguments": [
+          {
+              "name": "key",
+              "type": "key",
+              "key_spec_index": 0
+          },
+          {
+              "name": "value",
+              "type": "string"
+          }
+        ]
+    }
+}
--- a/src/t_string.c
+++ b/src/t_string.c
@@ -384,6 +384,32 @@ void psetexCommand(client *c) {
     setGenericCommand(c, OBJ_PX | OBJ_ARGV3, c->argv[1], c->argv[3], c->argv[2], UNIT_MILLISECONDS, NULL, NULL, NULL);
 }
 
+void delifeqCommand(client *c) {
+    robj *existing_value = lookupKeyWrite(c->db, c->argv[1]);
+
+    if (existing_value == NULL) {
+        addReplyLongLong(c, 0);
+        return;
+    }
+
+    if (checkType(c, existing_value, OBJ_STRING)) {
+        /* Error reply already sent */
+        return;
+    }
+
+    if (compareStringObjects(existing_value, c->argv[2]) != 0) {
+        addReplyLongLong(c, 0);
+        return;
+    }
+
+    if (!dbSyncDelete(c->db, c->argv[1])) {
+        addReplyLongLong(c, 0);
+        return;
+    }
+
+    addReplyLongLong(c, 1);
+}
+
 int getGenericCommand(client *c) {
     robj *o;
 
--- a/src/commands.def
+++ b/src/commands.def
@@ -10459,6 +10459,31 @@ struct COMMAND_ARG DECRBY_Args[] = {
 {MAKE_ARG("decrement",ARG_TYPE_INTEGER,-1,NULL,NULL,NULL,CMD_ARG_NONE,0,NULL)},
 };
 
+/********** DELIFEQ ********************/
+
+#ifndef SKIP_CMD_HISTORY_TABLE
+/* DELIFEQ history */
+#define DELIFEQ_History NULL
+#endif
+
+#ifndef SKIP_CMD_TIPS_TABLE
+/* DELIFEQ tips */
+#define DELIFEQ_Tips NULL
+#endif
+
+#ifndef SKIP_CMD_KEY_SPECS_TABLE
+/* DELIFEQ key specs */
+keySpec DELIFEQ_Keyspecs[1] = {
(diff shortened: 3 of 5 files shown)


xbasel:
IMO consider adding a more generic command eg.  `DELIF key EQ value`. This allows future extensions like NE, TYPE (eg. del if hash), etc.

Also worth considering a more general `CMPSET` command for compare-and-delete and compare-and-set.

sarthakaggarwal97:
I don't have a better alternative but I agree having `DEL IFEQ` where IFEQ is a subcommand would be confusing. Agree @xbasel's suggestion tho, if we are introducing a new command we should make it as generic as possible.

zuiderkwast:
I think DELIFEQ is a good idea, especially with the motivation that it's useful in the locking algorithm.

I agree we can't use DEL IFEQ. It's would be a breaking change. It already means Delete the key named "IFEQ", so we can't do that.

@xbasel I also thought about `DELIF key EQ value` that can be extended with GT, LT, NE in the future if we ever need that. I don't know if there is a use case for those extra conditions though. Maybe not? Maybe DELIFEQ is enough forever?

LinusU:
Thanks everyone for taking the time to look at this! 🙏 

I'm very open to making changes. Do you want me to update the code to `DELIF key EQ val` directly here? Or should we first reach consensus if we want that structure or not?

I would love some help with the test failures as well. The codecoverage feels wrong, since the reported untested line is the only line that returns `1`, and I have a test that asserts that the answer is `1`.

The other test failure, "External Server Tests", I'm a bit unsure of. Would love some pointers on where to start!

xbasel:
> I don't know if there is a use case for those extra conditions though. Maybe not? Maybe DELIFEQ is enough forever?

@zuiderkwast 
Probably most real-world cases (like locks) only need **IFEQ**, so **DELIFEQ** would work fine today.
But making it generic (`DELIF key OP value`) costs almost nothing now and avoids future regrets, for example, deleting expired sessions (eg last_access_time < now) might benefit from LT and atomicity...

src/commands/delifeq.json
```diff
@@ -0,0 +1,61 @@
...
+                    "DELETE"
+                ],
+                "begin_search": {
+                    "index": {
+                        "pos": 1
+                    }
+                },
+                "find_keys": {
+                    "range": {
+                        "lastkey": -1,
```
zuiderkwast: I believe this is why you get the CROSSSLOT error in the test. This says the rest of the arguments are keys. It's actually just one key.

Compare to set.json (The SET command) and see [README.md](https://github.com/valkey-io/valkey/blob/unstable/src/commands/README.md#key-specs) in the same directory.

```suggestion
                        "lastkey": 0,
```

src/commands/delifeq.json
```diff
@@ -0,0 +1,61 @@
+{
+    "DELIFEQ": {
+        "summary": "Delete key if value matches string.",
+        "complexity": "O(1)",
+        "group": "string",
+        "since": "8.2.0",
```
zuiderkwast: This should probably be 9.0.0. We don't plan to release 8.2.0.

src/t_string.c
```diff
@@ -380,6 +380,27 @@ void psetexCommand(client *c) {
     setGenericCommand(c, OBJ_PX | OBJ_ARGV3, c->argv[1], c->argv[3], c->argv[2], UNIT_MILLISECONDS, NULL, NULL, NULL);
 }
 
+void delifeqCommand(client *c) {
+    robj *existing_value = lookupKeyWrite(c->db, c->argv[1]);
+
```
zuiderkwast, line 389: We should fail with a WRONGTYPE error if someone tries to use this command on a key that's not a string.

Let's add a test case for this too.

```suggestion
    if (checkType(c, existing_value, OBJ_STRING)) return;

```

zuiderkwast (commented):
Don't change to DELIF yet. I guess DELIFEQ is more likely to be chosen. We will need to take a core team decision about the command and syntax. Until that's done, there is no hurry to update the details of the PR.

madolson:
@PingXie @enjoy-binbin @soloestoy We discussed it in the core meeting with the other 3 TSC folks. We were all aligned, will one of you 👍 so we mark this as an approved decision.

soloestoy:
I'm OK with adding a new command, the current command's name is not the best, but I cannot come up with a better name 😅 .

zuiderkwast:
@LinusU Will you address the comments so we can finish this up and get it merged?

LinusU:
@zuiderkwast thanks for all the pointers! 🙏 

I've:

- rebased on latest `unstable`
- fixed `lastkey` value
- changed `since` to `9.0.0`
- added `WRONGTYPE` error (with test case)
- fixed swapped `description` in reply schema

Please let me know if there is anything else you want me to do!

src/commands/delifeq.json
```diff
@@ -0,0 +1,61 @@
...
+        "group": "string",
+        "since": "9.0.0",
+        "arity": 3,
+        "function": "delifeqCommand",
+        "command_flags": [
+            "WRITE"
+        ],
+        "acl_categories": [
+            "KEYSPACE",
+            "STRING"
```
zuiderkwast: "KEYSPACE" seems to be used for commands that work on keys of all types, such as DEL, EXIST, EXPIRE, SCAN, etc.

I think it should be only "STRING" for this command.

```suggestion
            "STRING"
```

zuiderkwast (approved):
LGTM

sarthakaggarwal97 (approved):

LinusU:
✅ Removed `KEYSPACE`

-----

test failure looks unrelated?

```text

*** [err]: UNLINK can reclaim memory in background in tests/unit/lazyfree.tcl
Expected 17944320 > 18783952+1000000 (context: type eval line 11 cmd {assert {$peak_mem > $orig_mem+1000000}} proc ::test)
```

zuiderkwast:
Yeah the CI error is unrelated. Don't worry about it.

zuiderkwast:
Apparently you need to `make` and commit commands.def again after modifying the json file.

LinusU:
> Apparently you need to `make` and commit commands.def again after modifying the json file.

Sorry, must have missed that the last time 🙈 

I've rebased on latest `unstable` again, and fixed the outdated `commands.def` file ✅

zuiderkwast:
Thanks Linus, can you make a doc PR too? https://github.com/valkey-io/valkey-doc

The README file should provide most of what you need to know. If not, just reach out.

LinusU:
I've submitted docs now, sorry for the delay!

https://github.com/valkey-io/valkey-doc/pull/307

PR #1976, opened 2025-04-18
Clean up unnecessary use of strlen

artikell:
Clean up unnecessary use of strlen, and reduce string2ull useless length calculations

### DIFF
--- a/src/bitops.c
+++ b/src/bitops.c
@@ -556,6 +556,7 @@ int getBitOffsetFromArgument(client *c, robj *o, uint64_t *offset, int hash, int
  * On error C_ERR is returned and an error is sent to the client. */
 int getBitfieldTypeFromArgument(client *c, robj *o, int *sign, int *bits) {
     char *p = o->ptr;
+    size_t plen = sdslen(p);
     char *err = "Invalid bitfield type. Use something like i16 u8. Note that u64 is not supported but i64 is.";
     long long llbits;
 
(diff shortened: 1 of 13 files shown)


sarthakaggarwal97 (approved):

Nikhil-Manglore (approved):

src/bitops.c
```diff
@@ -556,6 +556,7 @@ int getBitOffsetFromArgument(client *c, robj *o, uint64_t *offset, int hash, int
  * On error C_ERR is returned and an error is sent to the client. */
 int getBitfieldTypeFromArgument(client *c, robj *o, int *sign, int *bits) {
     char *p = o->ptr;
+    size_t plen = sdslen(p);
```
zuiderkwast: Minor: `p` is not declared as `sds`. Use o->ptr directly instead, or declare `p` as sds?

```suggestion
    sds p = o->ptr;
    size_t plen = sdslen(p);
```

zuiderkwast (approved):
LGTM. Just a minor comment.

src/bitops.c
artikell: This is aligned with `getBitOffsetFromArgument` method implementation, so there is no special replacement.  
```
int getBitOffsetFromArgument(client *c, robj *o, uint64_t *offset, int hash, int bits) {
    long long loffset;
    char *err = "bit offset is not an integer or out of range";
    char *p = o->ptr;
    size_t plen = sdslen(p);
```
I tend to define `p` as sds.  So I modified the two logics together.

PR #1984, opened 2025-04-22
Allow scripts to support null characters again

### LINKED ISSUE #1942: [BUG] Valkey 8.1 does not accept null bytes in script source code
**Describe the bug**

Since commit b58088e31b19 "Adds support for running `EVAL` with different scripting engines (#1497)" Valkey
does not support null bytes in scripts anymore. (To be clear: this is not about handling strings with null bytes in scripts. It is about a script's source code containing null bytes)

**To reproduce**

In valkey-cli:

```
127.0.0.1:6379> eval "return '\\0';" 0
"\x00"
127.0.0.1:6379> eval "return '\x00';" 0
(error) ERR Error compiling script (new function): user_script:1: unfinished string near '<eof>'
```

**Expected behavior**

Prior to this change, a lua embedded zero and a real null byte both work: 

```
127.0.0.1:6379> eval "return '\\0';" 0
"\x00"
127.0.0.1:6379> eval "return '\x00';" 0
"\x00"
```

**Additional information**

It is unclear to me whether this is an intentional change (that was not described in the release nodes)

(For us, it is not a required use case, but it broke one of our tests. We intended to use a lua zero in a string literal and because of the library we use, this ended up becoming a null byte)

madolson:
We made a change during the refactoring of the engine changes that used a strlen on an SDS string. SDS are resizable, but they are binary safe and support NULL characters, which means we are truncating the string at the null character.

This is a breaking change for modules, so wanted to raise it for the upcoming patch.

An alternative is we could expose the script code as a server object, which would require some construction of the object in the function pathway and freeing. I opted to keep the code flow as similar as possible for the patch.

There is a test which only passes now.

For some reasons, functions don't support binary data. So as of right now this patch doesn't fix that case.

Resolves https://github.com/valkey-io/valkey/issues/1942

### DIFF
--- a/src/scripting_engine.c
+++ b/src/scripting_engine.c
@@ -217,6 +217,7 @@ static void engineTeardownModuleCtx(scriptingEngine *e) {
 compiledFunction **scriptingEngineCallCompileCode(scriptingEngine *engine,
                                                   subsystemType type,
                                                   const char *code,
+                                                  size_t code_len,
                                                   size_t timeout,
                                                   size_t *out_num_compiled_functions,
                                                   robj **err) {
@@ -229,6 +230,7 @@ compiledFunction **scriptingEngineCallCompileCode(scriptingEngine *engine,
         engine->impl.ctx,
         type,
         code,
+        code_len,
         timeout,
         out_num_compiled_functions,
         err);
--- a/src/scripting_engine.c
+++ b/src/scripting_engine.c
@@ -222,7 +222,7 @@ compiledFunction **scriptingEngineCallCompileCode(scriptingEngine *engine,
                                                   size_t *out_num_compiled_functions,
                                                   robj **err) {
     serverAssert(type == VMSE_EVAL || type == VMSE_FUNCTION);
-    compiledFunction **functions= NULL;
+    compiledFunction **functions = NULL;
     engineSetupModuleCtx(engine, NULL);
     if (engine->impl.methods.version == 1) {
         functions = engine->impl.methods.compile_code_v1(
--- a/src/scripting_engine.c
+++ b/src/scripting_engine.c
@@ -222,18 +222,30 @@ compiledFunction **scriptingEngineCallCompileCode(scriptingEngine *engine,
                                                   size_t *out_num_compiled_functions,
                                                   robj **err) {
     serverAssert(type == VMSE_EVAL || type == VMSE_FUNCTION);
-
+    compiledFunction **functions= NULL;
     engineSetupModuleCtx(engine, NULL);
+    if (engine->impl.methods.version == 1) {
+        functions = engine->impl.methods.compile_code_v1(
+            engine->module_ctx,
+            engine->impl.ctx,
+            type,
+            code,
+            timeout,
+            out_num_compiled_functions,
+            err);
+    } else {
+        /* Assume versions greater than 1 use updated interface. */
+        functions = engine->impl.methods.compile_code(
+            engine->module_ctx,
+            engine->impl.ctx,
+            type,
+            code,
+            code_len,
+            timeout,
+            out_num_compiled_functions,
+            err);
+    }
 
-    compiledFunction **functions = engine->impl.methods.compile_code(
-        engine->module_ctx,
-        engine->impl.ctx,
-        type,
-        code,
-        code_len,
-        timeout,
-        out_num_compiled_functions,
-        err);
 
     engineTeardownModuleCtx(engine);
 
--- a/src/valkeymodule.h
+++ b/src/valkeymodule.h
@@ -900,8 +900,8 @@ typedef ValkeyModuleScriptingEngineCompiledFunction **(*ValkeyModuleScriptingEng
     size_t *out_num_compiled_functions,
     ValkeyModuleString **err);
 
-/* Version one of code compilation, which may not safely handle
- * binary data. */
+/* Version one of source code compilation interface. This API does not allow the compiler to
+ * safely handle binary data. You should use a newer version of the API if possible.
 typedef ValkeyModuleScriptingEngineCompiledFunctionV1 **(*ValkeyModuleScriptingEngineCompileCodeFuncV1)(
     ValkeyModuleCtx *module_ctx,
     ValkeyModuleScriptingEngineCtx *engine_ctx,
--- a/src/valkeymodule.h
+++ b/src/valkeymodule.h
@@ -878,6 +878,8 @@ typedef struct ValkeyModuleScriptingEngineCallableLazyEvalReset {
  *
  * - `code`: string pointer to the source code.
  *
+ * - `code_len`: The length of the code string.
+ *
  * - `timeout`: timeout for the library creation (0 for no timeout).
  *
  * - `out_num_compiled_functions`: out param with the number of objects
@@ -894,6 +896,17 @@ typedef ValkeyModuleScriptingEngineCompiledFunction **(*ValkeyModuleScriptingEng
     ValkeyModuleScriptingEngineSubsystemType type,
     const char *code,
     size_t code_len,
+    size_t timeout,
+    size_t *out_num_compiled_functions,
+    ValkeyModuleString **err);
+
+/* Version one of code compilation, which may not safely handle
+ * binary data. */
+typedef ValkeyModuleScriptingEngineCompiledFunctionV1 **(*ValkeyModuleScriptingEngineCompileCodeFuncV1)(
+    ValkeyModuleCtx *module_ctx,
+    ValkeyModuleScriptingEngineCtx *engine_ctx,
+    ValkeyModuleScriptingEngineSubsystemType type,
+    const char *code,
     size_t timeout,
     size_t *out_num_compiled_functions,
     ValkeyModuleString **err);
@@ -986,7 +999,7 @@ typedef ValkeyModuleScriptingEngineMemoryInfo (*ValkeyModuleScriptingEngineGetMe
     ValkeyModuleScriptingEngineSubsystemType type);
 
 /* Current ABI version for scripting engine modules. */
-#define VALKEYMODULE_SCRIPTING_ENGINE_ABI_VERSION 1UL
+#define VALKEYMODULE_SCRIPTING_ENGINE_ABI_VERSION 2UL
 
 typedef struct ValkeyModuleScriptingEngineMethods {
     uint64_t version; /* Version of this structure for ABI compat. */
@@ -994,8 +1007,10 @@ typedef struct ValkeyModuleScriptingEngineMethods {
     /* Compile code function callback. When a new script is loaded, this
      * callback will be called with the script code, compiles it, and returns a
      * list of `ValkeyModuleScriptingEngineCompiledFunc` objects. */
-    ValkeyModuleScriptingEngineCompileCodeFunc compile_code;
-
+    union {
+        ValkeyModuleScriptingEngineCompileCodeFuncV1 compile_code_v1;
+        ValkeyModuleScriptingEngineCompileCodeFunc compile_code;
+    };
     /* Function callback to free the memory of a registered engine function. */
     ValkeyModuleScriptingEngineFreeFunctionFunc free_function;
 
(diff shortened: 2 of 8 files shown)


hpatro (approved):

src/valkeymodule.h
```diff
@@ -893,6 +893,7 @@ typedef ValkeyModuleScriptingEngineCompiledFunction **(*ValkeyModuleScriptingEng
     ValkeyModuleScriptingEngineCtx *engine_ctx,
     ValkeyModuleScriptingEngineSubsystemType type,
     const char *code,
+    size_t code_len,
```
zuiderkwast, line 898: This is breaking modules written for the 8.1.0 module API. We need to bump a struct version number, don't we?

zuiderkwast: Can we assume that no modules have started using this new struct yet? Fixing the issue is important enough to justify the breaking change?

zuiderkwast: Let's bump the version in `ValkeyModuleScriptingEngineMethods` and keep the old variant of the struct in module.c so we can cast and convert from it if version 1 is used?

rjd15372: Right, it's impossible to know if someone has already developed a module, although unlikely. 

The safest thing to do is to bump the value of `VALKEYMODULE_SCRIPTING_ENGINE_ABI_VERSION`, and version the new function in `valkeymodule.h`. And add logic in `scripting_engine.c` in `scriptingEngineCallCompileCode` function to deal with the different versions.

madolson: > Can we assume that no modules have started using this new struct yet? Fixing the issue is important enough to justify the breaking change?

The old API should also practically never be used, since it contains binary unsafe data. It's not that we added new functionality. I was trying to push this PR just because I wanted to make the revision on the API, if not this is less critical.

rjd15372 (approved):

sarthakaggarwal97 (approved):

src/valkeymodule.h
zuiderkwast: > The old API should also practically never be used, since it contains binary unsafe data.

Not all scripting languages allow nul characters in the source code, so it isn't necessarily a problem, assuming someone actually started using this API.

> I was trying to push this PR just because I wanted to make the revision on the API, if not this is less critical.

We can revise the API and still support the 8.1.0 API, so it's just a matter of internal complexity. We just need a little more internal handling (casting based on struct version), I'd guess 30 lines.

I can accept breaking it to avoid the added complexity though. I just want to understand if I understand the situation correctly.

rjd15372:
After talking with @zuiderkwast  I decided to check how the versioning code would look like to keep the compatibility of a module implemented against 8.1.0 and the changes are very small. These are the changes:

```diff
diff --git a/src/scripting_engine.c b/src/scripting_engine.c
index 818a2e21a..dc9e4e94a 100644
--- a/src/scripting_engine.c
+++ b/src/scripting_engine.c
@@ -225,15 +225,27 @@ compiledFunction **scriptingEngineCallCompileCode(scriptingEngine *engine,

     engineSetupModuleCtx(engine, NULL);

-    compiledFunction **functions = engine->impl.methods.compile_code(
-        engine->module_ctx,
-        engine->impl.ctx,
-        type,
-        code,
-        code_len,
-        timeout,
-        out_num_compiled_functions,
-        err);
+    compiledFunction **functions = NULL;
+
+    if (engine->impl.methods.version == VALKEYMODULE_SCRIPTING_ENGINE_ABI_V1) {
+        functions = engine->impl.methods.compile_code_v1(engine->module_ctx,
+                                                         engine->impl.ctx,
+                                                         type,
+                                                         code,
+                                                         timeout,
+                                                         out_num_compiled_functions,
+                                                         err);
+    } else {
+        functions = engine->impl.methods.compile_code(
+            engine->module_ctx,
+            engine->impl.ctx,
+            type,
+            code,
+            code_len,
+            timeout,
+            out_num_compiled_functions,
+            err);
+    }

     engineTeardownModuleCtx(engine);

diff --git a/src/valkeymodule.h b/src/valkeymodule.h
index 8e2b8f9ab..f95ecedb0 100644
--- a/src/valkeymodule.h
+++ b/src/valkeymodule.h
@@ -898,6 +898,15 @@ typedef ValkeyModuleScriptingEngineCompiledFunction **(*ValkeyModuleScriptingEng
     size_t *out_num_compiled_functions,
     ValkeyModuleString **err);

+typedef ValkeyModuleScriptingEngineCompiledFunction **(*ValkeyModuleScriptingEngineCompileCodeFuncV1)(
+    ValkeyModuleCtx *module_ctx,
+    ValkeyModuleScriptingEngineCtx *engine_ctx,
+    ValkeyModuleScriptingEngineSubsystemType type,
+    const char *code,
+    size_t timeout,
+    size_t *out_num_compiled_functions,
+    ValkeyModuleString **err);
+
 /* Free the given function.
  *
  * - `module_ctx`: the module runtime context.
@@ -986,7 +995,8 @@ typedef ValkeyModuleScriptingEngineMemoryInfo (*ValkeyModuleScriptingEngineGetMe
     ValkeyModuleScriptingEngineSubsystemType type);

 /* Current ABI version for scripting engine modules. */
-#define VALKEYMODULE_SCRIPTING_ENGINE_ABI_VERSION 1UL
+#define VALKEYMODULE_SCRIPTING_ENGINE_ABI_VERSION 2UL
+#define VALKEYMODULE_SCRIPTING_ENGINE_ABI_V1 1UL

 typedef struct ValkeyModuleScriptingEngineMethods {
     uint64_t version; /* Version of this structure for ABI compat. */
@@ -994,7 +1004,10 @@ typedef struct ValkeyModuleScriptingEngineMethods {
     /* Compile code function callback. When a new script is loaded, this
      * callback will be called with the script code, compiles it, and returns a
      * list of `ValkeyModuleScriptingEngineCompiledFunc` objects. */
-    ValkeyModuleScriptingEngineCompileCodeFunc compile_code;
+    union {
+        ValkeyModuleScriptingEngineCompileCodeFuncV1 compile_code_v1;
+        ValkeyModuleScriptingEngineCompileCodeFunc compile_code;
+    };

     /* Function callback to free the memory of a registered engine function. */
     ValkeyModuleScriptingEngineFreeFunctionFunc free_function;
```

src/valkeymodule.h
madolson: > Not all scripting languages allow nul characters in the source code, so it isn't necessarily a problem, assuming someone actually started using this API.

The one we support is LUA :)

zuiderkwast: No, this API is for integrating more scripting languages.

What I mean by it isn't necessarily a problem is it's not a problem if anyone used this API for integrating another scripting language.

Of course it's a problem that we broke Lua scripts containing nuls.

madolson:
Manually tested the ABI compatibility, we could add a test as well.

src/valkeymodule.h
madolson: Yeah, I understand, I was just using the only script we support today as an example of it being useful. Python also supports it. I'm not really sure about it more widely.

src/valkeymodule.h
```diff
@@ -889,6 +891,18 @@ typedef struct ValkeyModuleScriptingEngineCallableLazyEvalReset {
...
+    ValkeyModuleScriptingEngineCtx *engine_ctx,
+    ValkeyModuleScriptingEngineSubsystemType type,
+    const char *code,
+    size_t code_len,
+    size_t timeout,
+    size_t *out_num_compiled_functions,
+    ValkeyModuleString **err);
+
+/* Version one of code compilation, which may not safely handle
+ * binary data. */
```
madolson:
```suggestion
/* Version one of source code compilation interface. This API does not allow the compiler to
 * safely handle binary data. You should use a newer version of the API if possible.
```
This comment is a little terse.

src/valkeymodule.h
```diff
@@ -985,16 +999,18 @@ typedef ValkeyModuleScriptingEngineMemoryInfo (*ValkeyModuleScriptingEngineGetMe
...
     uint64_t version; /* Version of this structure for ABI compat. */
 
     /* Compile code function callback. When a new script is loaded, this
      * callback will be called with the script code, compiles it, and returns a
      * list of `ValkeyModuleScriptingEngineCompiledFunc` objects. */
-    ValkeyModuleScriptingEngineCompileCodeFunc compile_code;
-
+    union {
+        ValkeyModuleScriptingEngineCompileCodeFuncV1 compile_code_v1;
+        ValkeyModuleScriptingEngineCompileCodeFunc compile_code;
```
madolson, line 1012: This could have been flipped, so if someone was using the V1 API the compilation would continue to work. I do really want to gently nudge module developers to use v2 though.

src/valkeymodule.h
```diff
@@ -889,6 +891,18 @@ typedef struct ValkeyModuleScriptingEngineCallableLazyEvalReset {
...
+    ValkeyModuleScriptingEngineCtx *engine_ctx,
+    ValkeyModuleScriptingEngineSubsystemType type,
+    const char *code,
+    size_t code_len,
+    size_t timeout,
+    size_t *out_num_compiled_functions,
+    ValkeyModuleString **err);
+
+/* Version one of source code compilation interface. This API does not allow the compiler to
+ * safely handle binary data. You should use a newer version of the API if possible.
```
madolson:
```suggestion
 * safely handle binary data. You should use a newer version of the API if possible. */
```

src/valkeymodule.h
```diff
@@ -1012,9 +1028,9 @@ typedef struct ValkeyModuleScriptingEngineMethods {
     /* Function callback to get the used memory by the engine. */
     ValkeyModuleScriptingEngineGetMemoryInfoFunc get_memory_info;
 
-} ValkeyModuleScriptingEngineMethodsV1;
+} ValkeyModuleScriptingEngineMethodsV2;
```
madolson:
```suggestion
} ValkeyModuleScriptingEngineMethodsV1;
```

src/valkeymodule.h
```diff
@@ -1012,9 +1028,9 @@ typedef struct ValkeyModuleScriptingEngineMethods {
...
 
-#define ValkeyModuleScriptingEngineMethods ValkeyModuleScriptingEngineMethodsV1
+#define ValkeyModuleScriptingEngineMethods ValkeyModuleScriptingEngineMethodsV2
```
madolson:
```suggestion
#define ValkeyModuleScriptingEngineMethods ValkeyModuleScriptingEngineMethodsV1
```

madolson (commented):
Remove some unnecessary changes

src/scripting_engine.c
```diff
@@ -217,21 +217,35 @@ static void engineTeardownModuleCtx(scriptingEngine *e) {
...
                                                   const char *code,
+                                                  size_t code_len,
                                                   size_t timeout,
                                                   size_t *out_num_compiled_functions,
                                                   robj **err) {
     serverAssert(type == VMSE_EVAL || type == VMSE_FUNCTION);
-
+    compiledFunction **functions = NULL;
     engineSetupModuleCtx(engine, NULL);
+    if (engine->impl.methods.version == 1) {
```
zuiderkwast, line 227: Is this a hot code path? We would use an unlikely annotation if it is.

src/scripting_engine.c
```diff
@@ -217,21 +217,35 @@ static void engineTeardownModuleCtx(scriptingEngine *e) {
...
+        functions = engine->impl.methods.compile_code_v1(
+            engine->module_ctx,
+            engine->impl.ctx,
+            type,
+            code,
+            timeout,
+            out_num_compiled_functions,
+            err);
+    } else {
+        /* Assume versions greater than 1 use updated interface. */
```
zuiderkwast, line 237: Should we panic on unknown version?

src/valkeymodule.h
zuiderkwast: Agree. If someone uses the V1 API, it's because they've copied valkeymodule.h from Valkey 8.1.0 and use it when they're compiling the module. Otherwise, they shouldn't be using the V1 API.

zuiderkwast (approved):
Version compatibility adds some complexity. I wasn't sure it's worth it but I guess breaking the API is not a good thing to do in a patch release, so I like this.

src/scripting_engine.c
madolson: That would theoretically be another breaking change? If for some reason someone put some other version it would panic where it would have worked. I don't really have a strong opinion about how we should be validating the version though, I bounced around on a few implementations.

src/scripting_engine.c
madolson: Ok, ran some perf tests and it seems like it isn't. This is only used when doing code compilation, which takes a significant amount of time, which makes this call not detectable.

avifenesh:
@madolson Just a small note, this is relevant for many tools that have a strict check on the input of the users for no null, might be worth making a bit of noise about it, many users will spend a lot of checks on strings for no reason.

Issue #1882, opened 2025-03-25
[NEW] Possibility to use system jemalloc


zuiderkwast:
Packagers often want to use system jemalloc rather than the bundled one.

In the debian package for valkey 8.0.2, I found this patch that we could apply:

https://sources.debian.org/src/valkey/8.0.2+dfsg1-1/debian/patches/0004-Add-support-for-USE_SYSTEM_JEMALLOC-flag.patch/

@lamby Let's upstream it?

lamby:
You are very welcome to it! There are some similar patches you may be interested in, too.

zuiderkwast:
@lamby yeah I've looked at the other ones under https://sources.debian.org/src/valkey/8.0.2%252Bdfsg1-1/debian/patches/ too:

* [0001-Fix-FTBFS-on-kFreeBSD.patch](https://sources.debian.org/src/valkey/8.0.2%2Bdfsg1-1/debian/patches/0001-Fix-FTBFS-on-kFreeBSD.patch/): Yes, this one seems good to add, for the combination GNU/kFreeBSD kernel, which is a thing, obviously.

* [0002-Add-CPPFLAGS-to-upstream-makefiles.patch](https://sources.debian.org/src/valkey/8.0.2%2Bdfsg1-1/debian/patches/0002-Add-CPPFLAGS-to-upstream-makefiles.patch/): I'm not sure why you need this one. Some motivation would be useful.

* [0003-Use-get_current_dir_name-over-PATHMAX.patch](https://sources.debian.org/src/valkey/8.0.2%2Bdfsg1-1/debian/patches/0003-Use-get_current_dir_name-over-PATHMAX.patch/): This one avoids putting a 4K string on the stack by using `get_current_dir_name()`, but the latter uses `malloc()` (rather than `zmalloc()`) and we later free it using `zfree()`. This means it will mess up the memory usage tracking done in zmalloc and zfree. Therefore, we may not want to take this patch, at least not unmodified.

lamby:
> [0002-Add-CPPFLAGS-to-upstream-makefiles.patch](https://sources.debian.org/src/valkey/8.0.2%2Bdfsg1-1/debian/patches/0002-Add-CPPFLAGS-to-upstream-makefiles.patch/): I'm not sure why you need this one. Some motivation would be useful.

[...]

Issue #1585, opened 2025-01-18
[BUG] Memory corruption caught by hardened allocators


artemislena:
**Describe the bug**

When running with [Graphene's hardened malloc](/grapheneos/hardened_malloc) or [LLVM's scudo](https://llvm.org/docs/ScudoHardenedAllocator.html) (presumably other hardened malloc implementations too), Valkey will crash at startup with a segmentation fault. Even on debug log level, the crash happens very early, before any logs are being emitted.

**To reproduce**

On Linux, run Valkey with the `LD_PRELOAD` environment variable set to the hardened allocator's path, or use a global `/etc/ld.so.preload` file. [hardened_malloc has some instructions to help with that](/grapheneos/hardened_malloc?tab=readme-ov-file#traditional-linux-based-operating-systems). If you're on NixOS specifically, you can also use the `environment.memoryAllocator.provider` option. On other OSes, I'm not sure; it might be hard to do on macOS and no idea how preloading works on any BSD.

**Expected behavior**

Valkey shouldn't segfault regardless of allocator.

**Additional information**

Going off when we changed configs, bug was first observed sometime between October 27 and November 19 2024. Guessing it was introduced by 8.0.0 or 8.0.1 (nixpkgs updated from 7.2.7 to 8.0.1 on November 13, figure that's what caused it back then). The issue still persists in 8.0.2. This is **not** a bug in the allocator, rather, this is a bug in Valkey uncovered by more hardened allocators (which less hardened allocators like jemalloc or the glibc one wouldn't catch), as evidenced by the fact it happens with two different mallocs. This bug may result in a vulnerability, but I'm not reporting it as such because I don't know how to test for that sort of thing.

ranshid:
@artemislena thank you for reporting this. However I was unable to reproduce the issue.
I tried with both Valkey compiled with JEmalloc and LIBC malloc running linked to [Graphene's hardened malloc](https://github.com/grapheneos/hardened_malloc):

```
lsof -p `pidof valkey-server` | grep libhardened_malloc
valkey-se xxxxxx ubuntu  mem       REG   259,1    74520  800960 /home/ubuntu/hardened_malloc/out/libhardened_malloc.so
```

But no issues on startup and the server is behaving as expected.

I was running on AWS EC2 VM:

```
uname -a
[...]

PR #1985, opened 2025-04-22
Only enable defrag for vendored jemalloc

madolson:
Only enable defrag for vendored jemalloc. 

Resolves a point of discussion from https://github.com/valkey-io/valkey/issues/1585, although not the overall issue. This change should be reverted when we properly validate using system jemalloc as part of https://github.com/valkey-io/valkey/issues/1882.

### DIFF
--- a/deps/jemalloc/include/jemalloc/jemalloc.sh
+++ b/deps/jemalloc/include/jemalloc/jemalloc.sh
@@ -5,6 +5,9 @@ objroot=$1
 cat <<EOF
 #ifndef JEMALLOC_H_
 #define JEMALLOC_H_
+
+#define VALKEY_VENDORED_JEMALLOC yes
+
 #ifdef __cplusplus
 extern "C" {
 #endif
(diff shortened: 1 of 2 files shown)


zuiderkwast (approved):
Yes, let's backport this

deps/jemalloc/include/jemalloc/jemalloc.sh
```diff
@@ -5,6 +5,11 @@ objroot=$1
 cat <<EOF
 #ifndef JEMALLOC_H_
 #define JEMALLOC_H_
+
+/* A macro that is used to indicate that this the jemalloc vendored with the project
+ * and has been tested with active defragmentation. */
+#define VALKEY_VENDORED_JEMALLOC yes
```
zuiderkwast: Nit: If some code does `#if VALKEY_VENDORED_JEMALLOC`, it expands to invalid C code (or to a variable named yes). It's better that it expands to 1.

```suggestion
#define VALKEY_VENDORED_JEMALLOC 1
```

Issue #507, opened 2024-05-16
[Feature-Request]: Cross-Slot Command Execution in ValKey Cluster


bsmanit:
**Problem/Use-case:**

As we know that `single key command << pipelining <<< multi key` command  to write/read n key from Redis, and to leverage this multi key command on Redis to get the higher ops we can enable cross slot command execution only if all multi key are owned by executing shard (not involving multiple shards for single multi key command).

**Description:**
Introduce a new configuration option in redis.conf, enable-cross-slot, which defaults set to no but can be set to yes when running Redis in cluster mode. 
This feature enables cross-slot command execution if keys are owned by single shards.

**Design:**
Client: Already aware about slot range owned by each shard of Redis cluster and update with multiple topology refresh strategy.

To execute n key multi-key command, client groups this into K commands with hash ranges owned by each shards. (K is number of shards in cluster)
Then client async executes these multi key commands and join and return.

Since In above multiple nodes are involved to execute multi-key command, so their execution is not guaranteed to be atomic (so, they can actually break the atomic design of many Redis commands).

For that we can add a new config in `redis.conf` enable-cross-slot yes (by default no) and only considered when running Redis in cluster mode. {similar to Redis Proxy: https://github.com/RedisLabs/redis-cluster-proxy}

**Alternatives Considered:**
Leveraging Redis Proxy: Utilizing existing Redis Proxy solutions like https://github.com/RedisLabs/redis-cluster-proxy can mitigate the issue but adds another layer of complexity to the architecture.

Manual grouping of multi-key commands by the client based on hash slots: Given Redis's extensive 16K+ hash slots, users typically issue multi-key commands involving 10 or 100 keys. However, grouping these keys by hash slots often results in nearly single or double key multi-get commands. Unfortunately, this approach doesn't significantly enhance performance.

**Additional Information:**
With enable-cross-slot set to yes, Redis will ensure atomic execution of multi-key commands in individual shards, enhancing the reliability and consistency of operations.

Clients must be aware of the slot ranges owned by each shard in the Redis cluster and update their topology accordingly to leverage this feature effectively.

madolson:
I kind of like this idea, but my thought has been it should be scoped down to "very" specific commands. Specifically, I think it should be the commands with logical fan-in and fan out behavior, like `MGET`, `MSET`, `DEL`, etc. Esoteric commands like like `SUNIONSTORE` probably don't need to support multi-slot.

[...]

PR #1986, opened 2025-04-22
Add framework for cross-slot commands, enable for MGET

PingXie:
Introduces the `CMD_CROSS_SLOT` flag to allow specific commands to operate on keys residing in multiple slots, provided those slots are managed by the same shard.

Key changes:
- Added `CMD_CROSS_SLOT` command flag and corresponding `cross_slot` client flag.
- Modified `getNodeByQuery` in `cluster.c` to:
    - Check for the `CMD_CROSS_SLOT` flag.
    - Allow keys in different slots if they belong to the same node.
    - Introduce `CLUSTER_REDIR_CROSS_SHARD` error if keys span multiple nodes.
    - Prevent cross-slot operations if any involved slot is migrating/importing.
- Introduced `lazy_expire_disabled` client flag to prevent lazy expiration during cross-slot command execution.
- Updated `MGET` command definition to use the new `CMD_CROSS_SLOT` flag.
- Adjusted related logic (client reset, script error handling, key slot checks) to accommodate the new flags and behavior.

This lays the groundwork for easily enabling cross-slot semantics for other commands in the future while enabling it for MGET immediately.

### DIFF
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -2151,7 +2151,7 @@ size_t hashtableGetStatsMsg(char *buf, size_t bufsize, hashtableStats *stats, in
                       " top-level buckets: %lu\n"
                       " child buckets: %lu\n"
                       " max chain length: %lu\n"
-                      " avg chain length: %.02f\n"
+                      " avg chain length: %f\n"
                       " chain length distribution:\n",
                       stats->toplevel_buckets,
                       stats->child_buckets,


PingXie:
This PR is based on the following feedback provided by the community:

1. Cross-slots `MGET` has significant performance boost (validated);
1. Cross-slots `WRITE` is messy and not interesting;
1. Value of supporting cross-slots for commands other than `MGET` isn't clear;
1. Cross-slots `READ` shouldn't have the `WRITE` side effect because doing so complicates the replication;
1. We can start with `MGET` but it should be easy to extend it to other commands if needed without significantly increasing the command surface;

Note that this PR will reject an otherwise valid cross-slot `MGET` command if one of its slots is being migrated but when that is only for the legacy slot migration. It will work transparently with the planned atomic slot migration (#23).

PingXie:
@valkey-io/core-team PTAL

src/script.c
```diff
@@ -443,6 +443,10 @@ static int scriptVerifyClusterState(scriptRunCtx *run_ctx, client *c, client *or
             *err = sdscatfmt(sdsempty(),
                              "Command '%S' in script attempted to access keys that don't hash to the same slot",
                              c->cmd->fullname);
+        } else if (error_code == CLUSTER_REDIR_CROSS_SHARD) {
```
madolson: I actually think the else case error is more useful, because we are trying to access a key from a non-local node.

src/cluster.c
```diff
@@ -1105,45 +1155,29 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
...
-            /* Fall through and allow the command to be executed:
-             * this happens when server.cluster_allow_reads_when_down is
-             * true and the command is not a write command */
-        }
-    }
+    if (first_node == NULL) return myself;
 
     /* Return the hashslot by reference. */
-    if (hashslot) *hashslot = slot;
+    if (hashslot) *hashslot = first_slot;
```
madolson: Slot statistics are updated based off of this information, we will bias towards the first_slot.

src/server.h
```diff
@@ -2413,6 +2415,8 @@ typedef int serverGetKeysProc(struct serverCommand *cmd, robj **argv, int argc,
  * CMD_TOUCHES_ARBITRARY_KEYS: The command may touch (and cause lazy-expire)
  *                             arbitrary key (i.e not provided in argv)
  *
+ * CMD_CROSS_SLOT: The command may access keys in different slots.
```
madolson: We would need to expose this into the module API presumably, since there is a JSON command for multi-gets that can fan out. `JSON.MGET`.

src/db.c
```diff
@@ -258,7 +258,7 @@ int getKeySlot(sds key) {
      * so we must always recompute the slot for commands coming from the primary.
      */
     if (server.current_client && server.current_client->slot >= 0 && server.current_client->flag.executing_command &&
-        !server.current_client->flag.primary) {
+        !server.current_client->flag.cross_slot && !server.current_client->flag.primary) {
```
madolson: We've used `->slot = -1` to indicate it's cross slot in the past. We could also reserve `-2` for this purpose. This allows us to avoid taking another flag.

src/cluster.c
```diff
@@ -1202,7 +1234,9 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
  * node we want to mention in the redirection. Moreover hashslot should
  * be set to the hash slot that caused the redirection. */
 void clusterRedirectClient(client *c, clusterNode *n, int hashslot, int error_code) {
-    if (error_code == CLUSTER_REDIR_CROSS_SLOT) {
+    if (error_code == CLUSTER_REDIR_CROSS_SHARD) {
+        addReplyError(c, "-CROSSSHARD Keys in request don't hash to the same shard");
```
madolson: Will make the comment here, since I think this is the main point of discussion. So the client experience will be:
1. Clients will be updated so that they can decompose MGET into per shard MGETs and send them out. 
2. If the clients are updated and get a `CROSSSHARD` error, they will realize they need to discover topology again and then recompose the commands. If they aren't updated, they'll just throw the error up. So, technically a breaking change since we are changing the error code! (CLIENT CAPA mitigates this if we want)
3. The framework easily works on fan out commands. (As mentioned, DEL, UNLINK, EXISTS TOUCH, MSET, MGET) The commands that aren't fan out (SUNIONSTORE for example), we'll have issues with atomic slot migration in replicating the cross key dependencies since we can't replicate verbatim as the target won't have the other keys. Clients also don't have trivial strategies for merging, so maybe we'll never have a need for that.
4. Clients break atomicity when doing this fan out. I suppose our suggestion to clients will be to expose a way in each client to handle this special behavior. (cross_shard_mget() vs mget() or have an argument when creating a client to introduce the fanout option) If the client just exposes a raw API, it's up to the end user to orchestrate the fan out I suppose.

My other suggestion was that we should introduce a new caching focused command `MXGET`, that never throws the `CROSSSHARD` error. If a server receives that command, and it doesn't serve the slots associated with it, it'll just return empty values for it.

My only strong opinion is the slot migration problem in 3\, everything else in this PR seems OK to me.

src/cluster.c
```diff
@@ -1030,6 +1033,34 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
...
     int pubsubshard_included =
         (cmd_flags & CMD_PUBSUB) || (c->cmd->proc == execCommand && (c->mstate->cmd_flags & CMD_PUBSUB));
+    int is_write_command =
+        (cmd_flags & CMD_WRITE) || (c->cmd->proc == execCommand && (c->mstate->cmd_flags & CMD_WRITE));
+    int is_cross_slot_command =
+        (cmd_flags & CMD_CROSS_SLOT) || (c->cmd->proc == execCommand && (c->mstate->cmd_flags & CMD_CROSS_SLOT));
+
+    /* Cluster is globally down but we got keys? We only serve the request
+     * if it is a read command and when allow_reads_when_down is enabled. */
+    if (!isClusterHealthy()) {
```
madolson: This is a behavior change. Today we still try to redirect to another node, which could know of a steady quorum and serve the requests. Although not a significant issue, we've known of some issues in AWS where a node was not having issues but we didn't immediately replace it. Having this check at the bottom allows us to redirect away from it.

src/commands/mget.json
```diff
@@ -6,9 +6,16 @@
         "since": "1.0.0",
         "arity": -2,
         "function": "mgetCommand",
+        "history": [
+            [
+                "9.0.0",
+                "Supports cross-slot keys in cluster mode."
```
hwware: I prefer to indicate more detail:  Supports cross-slot keys in same shard in cluster mode.

src/server.h
```diff
@@ -246,6 +246,7 @@ extern int configOOMScoreAdjValuesDefaults[CONFIG_OOM_COUNT];
 #define CMD_ALLOW_BUSY ((1ULL << 26))
 #define CMD_MODULE_GETCHANNELS (1ULL << 27) /* Use the modules getchannels interface. */
 #define CMD_TOUCHES_ARBITRARY_KEYS (1ULL << 28)
+#define CMD_CROSS_SLOT (1ULL << 29)
```
hwware: We also need to add this flag in the function: addReplyFlagsForCommand() which is for COMMAND INFO  (https://valkey.io/commands/command-info/)

hwware:
I just take a short look this PR, my question is how client enables and disable this feature?  Or it is enabled by default?

src/cluster.c
hwware: Do we have a chance to add an option to allow client to enable or disable this feature?

madolson: `CLIENT CAPA` could be used maybe. I still think we should leave this call where it was.

hwware: > `CLIENT CAPA` could be used maybe. I still think we should leave this call where it was.

I think it works as we expect.

src/cluster.c
ebarskiy: I like adding new command like MXGET, we can make a more convenient interface this way. In particular, having per-slot or, even better, per-key errors would make usage easier for the cluster case.

zuiderkwast: Per key errors? So we can include an error in an array reply? That's interesting. I'm not sure any other command does that. Not even EXEC, I think.

ovaiskhansc: > If the clients are updated and get a CROSSSHARD error, they will realize they need to discover topology again and then recompose the commands. If they aren't updated, they'll just throw the error up. So, technically a breaking change since we are changing the error code! (CLIENT CAPA mitigates this if we want)

Yes, it's a breaking change in that respect and based on previous discussions (regarding admins may control the server settings and developers not being aware of those), this should be a thing that developer opts in via CLIENT CAPA or a new command. 

A sidenote on error handling, based on Snap's experience in running with this mode for over a year, this usually happens during scale-up or scale down when slot assignment is changing at a fairly fast rate so fast that just refreshing the topology isn't sufficient. The best way we have been able to mitigate this is by converting the re-try to fall back to pipelined gets or single slot mget. A new command that can return partial data, would be great.

ebarskiy: Correct, we don't have multiple errors right now. 
The typical use case in distributed world is when we ask multiple keys they actually randomly spread across multiple slots and machines. We don't care about atomicity there, something that MGET is designed for.
Here, we want to be able combine multiple requests into one to save a lot of compute.
But I also want to be able get partial results, so even if slot is migrated/ing I still data for all keys that is there. 
I don't want to call this cell again, do all the parsing etc.
In some applications I don't even care about getting a full response and can skip keys in moved slot altogether.

madolson: Should we return the data that is stable? I.E. if we have 4 of the 8 slots from the requests, we return the data from those 4 slots and then give some indication that the data was unavailable for the remaining ones?

ebarskiy: exactly

src/cluster.c
PingXie: I will revert this change

src/cluster.c
PingXie: thanks for the reminder on slot stats :)

the bias actually comes from  `ioThreadReadQueryFromClient` but now I think there is more work to make this change full compatible with slot-stats

src/cluster.c
PingXie: > Clients will be updated so that they can decompose MGET into per shard MGETs and send them out.

Yes regardless of the interface design/implementation, this needs client support though some clients are easier than others since they already split keys by slots today, including Valkey-go and Lettuce.

> If the clients are updated and get a CROSSSHARD error, they will realize they need to discover topology again and then recompose the commands.

I don't think this is the case. For clients that don't slice the keys (by slots), they could get CROSS_SLOT already, which will be handled by the application. With this new behavior of MGET (assuming we are going with this interface), CROSS_SHARD could be returned so there could be some confusion during development but that shouldn't happen in production. 

> The framework easily works on fan out commands. (As mentioned, DEL, UNLINK, EXISTS TOUCH, MSET, MGET) The commands that aren't fan out (SUNIONSTORE for example), 

Yes there will be more work for `WRITE` but I am convinced now that `WRITE` is not desired . We can revisit this framework if there is a legit use case and I think it just needs incremental work and we can keep the work in this PR.

> Clients break atomicity when doing this fan out. 

It is already the case for some clients like Valkey-go/Lettuce, which I checked.

> My other suggestion was that we should introduce a new caching focused command MXGET

Not a hard blocker for me. I am open to this suggestion.

> per-key errors would make usage easier for the cluster case.
> The best way we have been able to mitigate this is by converting the re-try to fall back to pipelined gets or single slot mget. A new command that can return partial data, would be great.

Interesting - would return empty string back in this case work? I think we would need a new data type (meaning modifying the RESP protocol) if we would like to introduce a per-key error. It is not clean to use a special string to indicate errors.

> Yes, it's a breaking change in that respect and based on previous discussions (regarding admins may control the server settings and developers not being aware of those), this should be a thing that developer opts in via CLIENT CAPA or a new command.

I thought about opting in via CLIENT CAPA earlier in the other thread but I feel that the value isn't much. Please see my comments above.

Also if we are in favor of a new command, this becomes a moot point.

PingXie:
### Explicit requirements
1. performance is the goal; atomicity is not required
2. `MGET` is the only command of interest for now; however, we should be able to introduce cross-slot to other "read" commands easily with this work
3. "write" is not desired/needed in the foreseeable future but the work in this PR should allow the support for cross-slot "write" commands to be introduced in a compatible experience for developers
4. need to support module
5. need to be compatible with existing engine features, including slot level metrics and memory prefetching

### High level decisions for discussion 
1. expanding `MGET` vs introducing a new command?
2. if new command, what should the new command be called? how to create an easily identifiable naming pattern to help reduce the learning curve for the developers?
3. should we give the client a say about whether it'd like to work with unstable slots or not?
4. if we support unstable slots, what should be returned if the requested key is not found in the unstable slot? an error or {}? if error, what data type should be used?
6. in the redirection case, should we introduce a new `-ASK/-MOVED` error that allows all the slots in question to be returned? otherwise, the client can update one slot's owning shard at a time upon receiving the existing `-ASK/-MOVED` error. it would take the client N retries to catch up with the cluster topology updates, where N is the number of slots in the original cross-slot request.
7. how do we divide up the resources, cpu/network/memory, consumed by a cross-slot mget?

@valkey-io/core-team

PingXie:
I did some more digging on pipelining. I think the core issue is that only the first command in a pipeline batch benefits from the memory prefetching. The remaining ones are all run with cold cache on the main thread (and get parsed on the main thread as well).

this is the flame chart that illustrates these two points, indirectly, 1) the presence of `processInputBuffer` on the main thread; 2) the much wider margin on the io threads, indicating that io threads are merely burning cycles without doing real work. 
image

Here is the mget flamechart for comparison.

image

I will see if I can get the more definitive (v)PMU counters next.

zuiderkwast:
Thanks Ping for the analysis!

The current IO threads design is indeed not optimized for pipelining. It parses one command from each client and hands it over to the main thread. The main thread executes the parsed command and then, if there is more in the input buffer, the main thread parses it and executes the rest of the pipeline. We can see this in [`processPendingCommandAndInputBuffer()`](https://github.com/valkey-io/valkey/blob/8.1.1/src/networking.c#L3162).

In [`processInputBuffer()`](https://github.com/valkey-io/valkey/blob/8.1.1/src/networking.c#L3219) there is a while-loop that reads, parses and executes all the available commands in the pipeline. I believe this explains what you see in the flame graph.

If IO-threads are disabled, that the main thread prioritizes pipelines and doesn't interleave it with commands from other clients, which is good for pipelines and transactions, but probably still not good enough to compete with cross-slot-MGET.

There is a follow-up idea for IO-threading to let the IO thread parse more than one command: https://github.com/valkey-io/valkey/issues/761, the bullet "Parsing offload to parse multiple commands". I created a specific issue for it now:

* https://github.com/valkey-io/valkey/issues/2044

Even with IO-threads disabled, we can do "look-ahead" of pipelined commands by peeking into the input buffer. If we detect that there is another GET or MGET command in the pipeline, we could parse it and prefetch the keys while we execute the current command.

There is a feature plan to allow executing readonly commands in parallel in different threads if the commands are operating on different slots: https://github.com/valkey-io/valkey/issues/2022. It takes advantage of the fact that each command operates on a single slot.

PingXie:
The key contributor here is actually the memory perfecting. In my test, with enabling memory prefetching just for the first slot in a cross-slot MGET command, which is what happens in this PR, its performance (in terms of keys  per second) is pretty much on par with that of pipelined GETs. That being said, you are right that being able to parse more than more commands is the prerequisite to the effective prefetching. 

BTW, I wonder if we really need to pay for the complexity of "look-ahead" - I feel that all we need is to parse the entire qbuf into a command queue maintained on the client struct and then ensure the memory prefetching logic pop one command off of that queue one at a time.

madolson:
> BTW, I wonder if we really need to pay for the complexity of "look-ahead" - I feel that all we need is to parse the entire qbuf into a command queue maintained on the client struct and then ensure the memory prefetching logic pop one command off of that queue one at a time.

I also wonder if we need to parse the whole query buffer, or just the *next* command. If we prefetch the data for the next command while we execute the current one, it might resolve to similar amounts of time.

zuiderkwast:
> > BTW, I wonder if we really need to pay for the complexity of "look-ahead" - I feel that all we need is to parse the entire qbuf into a command queue maintained on the client struct and then ensure the memory prefetching logic pop one command off of that queue one at a time.

Yeah, this is the better solution. It may require some more refactoring of the parsing code though.

> I also wonder if we need to parse the whole query buffer, or just the _next_ command. If we prefetch the data for the next command while we execute the current one, it might resolve to similar amounts of time.

To get maximum benefit, I believe we should parse a few commands ahead, so we can prefetch a batch of ~16 keys, whether it's a single MGET command or 16 GET/SET/HSET/SADD commands.

The prefetching is an iterative process of 3-4 steps. After doing the first step on each key lookup (e.g. prefetching the hashtable bucket), we need to wait for the memory to be loaded before we can do the next step (e.g. prefetching the key of a potential match before we check if it's a match). When we prefetch in batches, we pay the cost of the cold memory access once per batch instead of once per key. But we can also gain if we can always be a few steps ahead of the execution: When we execute a command, we do one step of prefetching on keys of the commands that are waiting in the queue. In this way, we can have everything prefetched if we just prefetch 3-4 commands ahead.

If you look at the recent optimization of the hashtable iterator, it uses this technique. It prefetches a few steps ahead of the element returned by hashtableNext(). See #1501.

madolson:
> BTW, I wonder if we really need to pay for the complexity of "look-ahead" - I feel that all we need is to parse the entire qbuf into a command queue maintained on the client struct and then ensure the memory prefetching logic pop one command off of that queue one at a time.

Maybe we should be thinking of the entire system as more like a pipeline of commands, just like the CPU does. The main command execution thread has a series of commands it needs to execute in various stages of prepared to execute:

```
Head: SET FOO BAR; FOO has hash bucket and serverObject prefetched. Execute!
Queued Command1: GET FOO2; FOO2 has hash bucket prefetched. Prefetch server Object.
Queued Command2: GET FOO3; FOO3 has nothing prefetched. Prefetch hash bucket.
Queued Command3: GET FOO4; Foo4 has nothing prefetched. Pipeline is full, don't start processing this command yet.
```

Complex operations like multi-exec might need more prefetching stages. The actual set of commands could still be maintained on the client.

dmytro-arkhypenko:
@PingXie Do I get it right - you are abandonning idea of cross-slot framework for MGET because you found the way to make single GET to work faster?

zuiderkwast:
@dmytro-arkhypenko yes, and/or use one MGET per slot, if they are pipelined. There are now two PRs for the same idea actually. Do you want to test this one? Preliminary results is it's makes pipelined commands twice as fast:

* #2092
* #2044 (the initial idea discussed)

dmytro-arkhypenko:
> @dmytro-arkhypenko yes, and/or use one MGET per slot, if they are pipelined. There are now two PRs for the same idea actually. Do you want to test this one?

Well might test this definitely, but isn't multislot command (not just MGET) potentially reduces the number of responses from 16k+(number of slots) to whatever number of shards you got and you can't probably beat it by x2 speedup. 

I was hopefull this PR would come as a solution to #507  and a bit said it is not.

PR #1992, opened 2025-04-23
Add output buffer limiting for unauthenticated clients

uriyage:
This commit introduces a mechanism to track client authentication state with a new `ever_authenticated` flag. 
It refactors client authentication handling by adding a `clientSetUser` function that properly sets both the authenticated and `ever_authenticated` flags.

The implementation limits output buffer size for clients that have never been authenticated.

Added tests to verify the output buffer limiting behavior for unauthenticated clients.

### DIFF
--- a/src/debug.c
+++ b/src/debug.c
@@ -499,8 +499,8 @@ void debugCommand(client *c) {
 "CLUSTERLINK KILL <to|from|all> <node-id>",
 "    Kills the link based on the direction to/from (both) with the provided node." ,
 "CLIENT-ENFORCE-REPLY-LIST <0|1>",
-            "When set to 1, it enforces the use of the client reply list directly",
-            "    and avoids using the client's static buffer.",
+"    When set to 1, it enforces the use of the client reply list directly",
+"    and avoids using the client's static buffer.",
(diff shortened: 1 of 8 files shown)


src/debug.c
```diff
@@ -498,6 +498,9 @@ void debugCommand(client *c) {
 "    Enable or disable the reply buffer resize cron job",
 "CLUSTERLINK KILL <to|from|all> <node-id>",
 "    Kills the link based on the direction to/from (both) with the provided node." ,
+"CLIENT-ENFORCE-REPLY-LIST <0|1>",
+            "When set to 1, it enforces the use of the client reply list directly",
+            "    and avoids using the client's static buffer.",
```
zuiderkwast: Align the output of the DEBUG HELP.

```suggestion
"CLIENT-ENFORCE-REPLY-LIST <0|1>",
"    When set to 1, it enforces the use of the client reply list directly",
"    and avoids using the client's static buffer.",
```

zuiderkwast (commented):
The new test case is flaky?

zuiderkwast (approved):

PR #1997, opened 2025-04-23
Release notes for 8.1.1

murphyjacob4:


00-RELEASENOTES
```diff
@@ -9,6 +9,32 @@ CRITICAL: There is a critical bug affecting MOST USERS. Upgrade ASAP.
...
+* Fix crash during TLS handshake with I/O threads (#1955)
+* Fix cluster slot stats assertion during promotion of replica (#1950)
+* Fix panic in primary when blocking shutdown after previous block with timeout (#1948)
+* Ignore stale gossip packets that arrive out of order (#1777)
+* Fix incorrect lag reported in XINFO GROUPS (#1952)
+* Fix engine crash on module client blocking during keyspace events (#1819)
+* Avoid shard id update of replica if not matching with primary shard id (#573)
+* Only enable defrag for vendored jemalloc (#1985)
+* Allow scripts to support null characters again (#1984)
+* Blocking client followup fix related to log_req_res (#1989)
```
zuiderkwast: This one was not a bug in 8.1.0 so maybe we don't need to mention it? It is only about a test, not a real bug.

zuiderkwast (approved):
LGTM. Maybe skip some bug fix that was not a bug, only a CI issue basically.

00-RELEASENOTES
zuiderkwast:
```suggestion
```

murphyjacob4: Removed

PR #1998, opened 2025-04-23
Release notes for 8.0.3

murphyjacob4:


00-RELEASENOTES
```diff
@@ -9,6 +9,36 @@ CRITICAL: There is a critical bug affecting MOST USERS. Upgrade ASAP.
...
+
+Upgrade urgency SECURITY: This release includes security fixes we recommend you
+apply as soon as possible.
+
+Bug fixes
+=========
+* Optimize RDB load performance and fix cluster mode resizing on replica side (#1199)
+* Fix memory leak in forgotten node ping ext code path (#1574) 
+* Fix cluster info sent stats for message with light header (#1563)
+* Use fake client flag to replace not conn check (#1198)
```
zuiderkwast: This line mentions internal details, not very useful to the users. If it fixes an actual problem (does it?), we can mention that, or if it's just refactoring we don't need to mention it IMHO.

```suggestion
```

00-RELEASENOTES
```diff
@@ -9,6 +9,36 @@ CRITICAL: There is a critical bug affecting MOST USERS. Upgrade ASAP.
...
+* Fix module LatencyAddSample still work when latency-monitor-threshold is 0 (#1541)
+* Fix raxRemove crash at memcpy() due to key size exceeds max Rax size (#1722)
```
zuiderkwast: This PR title wasn't very good as a release notes title IMHO.

"Fix crash" sounds severe, but this one is actually just a hypothetical crash. It was never reported as a bug. I think it was just discovered by reading code or by fuzzing.

* It doesn't affect streams because streams don't use long key names.
* ACL? Huge key names are unlikely there too.
* Client tracking with huge keys > 512MB? Maybe..
* The module API for "Dict"? It's implemented as a rax. But then we could have reproduced it with a module API test case. This fix just has a unit test of rax.

Maybe we can soften it saying "potential crash"?

```suggestion
* Fix potential crash in radix tree recompression of huge keys (#1722)
```

00-RELEASENOTES
```diff
@@ -9,6 +9,36 @@ CRITICAL: There is a critical bug affecting MOST USERS. Upgrade ASAP.
...
+* fix: add samples to stream object consumer trees (#1825)
+* Fix cluster slot stats assertion during promotion of replica (#1950)
+* Fix panic in primary when blocking shutdown after previous block with timeout (#1948)
+* Ignore stale gossip packets that arrive out of order (#1777)
+* Fix incorrect lag reported in XINFO GROUPS (#1952)
+* Avoid shard id update of replica if not matching with primary shard id (#573)
+
+Security fixes
+==============
+* CVE-2025-21605 Limit output buffer for unauthenticated clients (#1994)
```
zuiderkwast: This fix has one PR for each release actually. :) There is none targeting unstable yet.

```suggestion
* CVE-2025-21605 Limit output buffer for unauthenticated clients (#1993)
```

00-RELEASENOTES
murphyjacob4: Sounds good. Removed

00-RELEASENOTES
murphyjacob4: Sounds good to me. Applied

00-RELEASENOTES
murphyjacob4: Ah, thanks

zuiderkwast (approved):

PR #2000, opened 2025-04-24
CI build-macos version from macOS 14 to macOS 15

vitahlin:
macOS 15 has been officially released for some time, our current `build-macos` workflow is still running on macOS 14. 
This PR upgrades the macOS version used in our CI to macOS 15 to keep up with the latest OS version and ensure long-term compatibility.

The `build-macos` CI workflow has successfully run on macOS 15: https://github.com/vitahlin/valkey/actions/runs/14640672882

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1130,7 +1130,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-13, macos-14]
+        os: [macos-13, macos-15]
     runs-on: ${{ matrix.os }}
     if: |
       (github.event_name == 'workflow_dispatch' ||
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1130,7 +1130,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-13, macos-15]
+        os: [macos-13, macos-14,  macos-15]
     runs-on: ${{ matrix.os }}
     if: |
       (github.event_name == 'workflow_dispatch' ||
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1130,7 +1130,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-13, macos-14,  macos-15]
+        os: [macos-13, macos-14, macos-15]
     runs-on: ${{ matrix.os }}
     if: |
       (github.event_name == 'workflow_dispatch' ||


.github/workflows/daily.yml
```diff
@@ -1130,7 +1130,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-13, macos-14]
+        os: [macos-13, macos-15]
```
hwware:
```suggestion
        os: [macos-13, macos-14,  macos-15]
```

hwware: I would prefer to keep macOS 14

vitahlin: macOS 14 is an intermediate version—if our builds work correctly on both macOS 13 and 15, it’s reasonable to assume that 14 would also work as expected. Since this is in the daily CI, running one additional job is acceptable.

.github/workflows/daily.yml
```diff
@@ -1130,7 +1130,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [macos-13, macos-14]
+        os: [macos-13, macos-14, macos-15]
```
madolson, line 1133: I think the reason we have this is that the actual test-macos job builds on macos-15, so this is just covering the old versions for builds. We could add a comment documenting that though.

vitahlin: Got it, thanks for the clarification!

That makes sense — `build-macos` is mainly meant to ensure compatibility with older macOS versions, while `test-macos` already covers macOS 15. In that case, I’ll revert the addition of macOS-15 here and add a comment explaining the purpose of this job.

Also, would it make sense to rename this CI job to something like `build-old-version-macos` to make its role more clear at a glance? I'm happy to help update that if it sounds reasonable to you.

hwware: I have no strong opinion to rename the CI job name. Either way works for me.

hwware (approved):

.github/workflows/daily.yml
madolson: I'm good with either approach.

vitahlin: Ok, then let’s keep the original setup: let test-ci handle runtest, and build-ci handle make, just adding macOS 15 compilation support.

madolson:
@hwware @vitahlin I don't really understand the outcome and why we decided to merge. We are now running 1 extra build for macos 15 which is already covered elsewhere. This seems like unnecessary amounts of work. I thought we were discussing adding a comment vs changing the job name, I wasn't infavor of merging it as is.

vitahlin:
> @hwware @vitahlin I don't really understand the outcome and why we decided to merge. We are now running 1 extra build for macos 15 which is already covered elsewhere. This seems like unnecessary amounts of work. I thought we were discussing adding a comment vs changing the job name, I wasn't infavor of merging it as is.

@madolson  Totally understand your confusion — it’s possible I misunderstood the discussion and thought the conversation was resolved, which is why it was eventually merged.
I agree this could have been handled better.
At this point, would it make sense for me to open a new PR to update the CI job naming?
Just wanted to check if that would add extra review work on your side — or if it’s easier, would you be open to making the adjustment directly?
Happy to align on whatever approach works best — be it reverting or making improvements.

hwware:
> @hwware @vitahlin I don't really understand the outcome and why we decided to merge. We are now running 1 extra build for macos 15 which is already covered elsewhere. This seems like unnecessary amounts of work. I thought we were discussing adding a comment vs changing the job name, I wasn't infavor of merging it as is.

I misunderstand your thought totally.  @vitahlin How about create a PR to revert to previous CI? Sorry for that

vitahlin:
> > @hwware @vitahlin I don't really understand the outcome and why we decided to merge. We are now running 1 extra build for macos 15 which is already covered elsewhere. This seems like unnecessary amounts of work. I thought we were discussing adding a comment vs changing the job name, I wasn't infavor of merging it as is.
> 
> I misunderstand your thought totally. @vitahlin How about create a PR to revert to previous CI? Sorry for that

A new PR to fix that, sorry for adding more review work again

PR #2001, opened 2025-04-24
Minor change for sadd command

hwware:
Update the server.dirty only when it is necessary


zuiderkwast (approved):

src/t_set.c
```diff
@@ -614,8 +614,8 @@ void saddCommand(client *c) {
     if (added) {
         signalModifiedKey(c, c->db, c->argv[1]);
         notifyKeyspaceEvent(NOTIFY_SET, "sadd", c->argv[1], c->db->id);
+        server.dirty += added;
     }
-    server.dirty += added;
```
hpatro, line 618: If the `added` variable is zero, it will get optimized out by the compiler I believe. I don't think we necessarily need this change for perf. For code clarity we can.

PR #2004, opened 2025-04-24
Group client IO data in a sub-struct

ayush933:
Move client IO related fields to a new sub struct. 
One of the items from #761. 

Verified correctness by running all the unit tests locally.

### DIFF
--- a/src/acl.c
+++ b/src/acl.c
@@ -499,7 +499,7 @@ void ACLFreeUserAndKillClients(user *u) {
             /* We will write replies to this client later, so we can't
              * close it directly even if async. */
             if (c == server.current_client) {
-                c->flag.close_after_command = 1;
+                c->io_data->flag.close_after_command = 1;
             } else {
                 freeClientAsync(c);
             }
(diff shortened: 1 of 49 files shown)


ayush933:
Hi @uriyage, can you please check if I am in the right direction. 
Also are there any other fields we should move?
Thanks.

uriyage:
Hi @ayush933 , The motivation for this change was to set the `io_data` to NULL when sending the client to the IO threads, and send only the `io_data` sub-struct to the IO thread. 

This avoids: 1) the main thread accessing the io_data while the IO thread is changing it, and 2) the IO thread accessing other client data which the main thread may access. 

While this commit adds the `io_data` structure, it doesn't yet set it to NULL when sending it to IO threads, and doesn't change the functions that the IO thread may call receive only the `io_data` struct instead of the client struct. 

The io_data struct should include every field that the IO thread needs to access.

ayush933:
Not all changes are done yet but considering the size of the refactor, I want to do the `set the io_data to NULL when sending the client to the IO threads` part in a different PR so that this one mostly has the struct changes. 

Is that ok @uriyage ?

ayush933:
Had an offline discussion with @uriyage. 
Closing this at the moment since it makes sense to pick this up after some other planned IO threads changes are merged.

PR #2007, opened 2025-04-25
Convert pubsub dicts to hashtables

zuiderkwast:
Two dicts are converted to hashtables:

1. On each client, the set of channels/patterns/shard-channels the client is subscribed to
2. On each channel or pattern, the set of clients subscribed to it.

When this is done, we can remove a lot of complexity in dict.c including the pointer tricks for different kinds of dictEntry, because that will become dead code. We can delete the dead code in a separate PR.


hpatro (approved):

ranshid (approved):
LGTM

PR #633, opened 2024-06-12
Adjust the log level of some logs in the cluster

enjoy-binbin:
I think the log of pfail status changes will be very useful.
The other parts were scanned and found that it can be modified.

Changes:
1. Changing pfail status releated logs from VERBOSE to NOTICE.
2. Changing configEpoch collision log from VERBOSE(warning) to NOTICE.
3. Changing some logs from DEBUG to NOTICE.

PR #2010, opened 2025-04-26
Avoid log spam about cluster node failure detection by each primary

### LINKED ISSUE #2076: Logs consuming significant compute during cluster node failure detection

The issue is to discuss the server log emitted during cluster failover in large clusters. It looks like that this particular log takes about 3-4% of the total compute (which is already around 100%).

```
serverLog(LL_NOTICE, "Node %.40s (%s) reported node %.40s (%s) as not reachable.", sender->name,
     sender->human_nodename, node->name, node->human_nodename);
```

Sharing the profile over here:

Image

I would like to discuss if we can reduce the severity of this log.

hpatro:
Fixes: #2076

After node failure detection/recovery and gossip by each primary, we log about the failure detection/recovery at NOTICE level which can spam the server and the behavior is quite expensive on ec2 burstable instance types. I would prefer us rolling it back to VERBOSE level.

Change was introduced in #633

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -2409,13 +2409,13 @@ void clusterProcessGossipSection(clusterMsg *hdr, clusterLink *link) {
             if (sender) {
                 if (flags & (CLUSTER_NODE_FAIL | CLUSTER_NODE_PFAIL)) {
                     if (clusterNodeIsVotingPrimary(sender) && clusterNodeAddFailureReport(node, sender)) {
-                        serverLog(LL_NOTICE, "Node %.40s (%s) reported node %.40s (%s) as not reachable.", sender->name,
+                        serverLog(LL_VERBOSE, "Node %.40s (%s) reported node %.40s (%s) as not reachable.", sender->name,
                                   sender->human_nodename, node->name, node->human_nodename);
                     }
                     markNodeAsFailingIfNeeded(node);
                 } else {
                     if (clusterNodeDelFailureReport(node, sender)) {
-                        serverLog(LL_NOTICE, "Node %.40s (%s) reported node %.40s (%s) is back online.", sender->name,
+                        serverLog(LL_VERBOSE, "Node %.40s (%s) reported node %.40s (%s) is back online.", sender->name,
                                   sender->human_nodename, node->name, node->human_nodename);
                     }
                 }


src/cluster_legacy.c
```diff
@@ -2409,13 +2409,13 @@ void clusterProcessGossipSection(clusterMsg *hdr, clusterLink *link) {
             if (sender) {
                 if (flags & (CLUSTER_NODE_FAIL | CLUSTER_NODE_PFAIL)) {
                     if (clusterNodeIsVotingPrimary(sender) && clusterNodeAddFailureReport(node, sender)) {
-                        serverLog(LL_NOTICE, "Node %.40s (%s) reported node %.40s (%s) as not reachable.", sender->name,
+                        serverLog(LL_VERBOSE, "Node %.40s (%s) reported node %.40s (%s) as not reachable.", sender->name,
```
sarthakaggarwal97, line 2412: Should we make the level Warning? I am wondering if users depend on this log for any debugging.

Also, just out of curiosity, does changing log severity come under breaking change?

hpatro: I think it might be helpful in case of small clusters but unsure how valuable it is to log it for each primary in a large cluster setup. 

My suggestion is to log the state periodically every few seconds to debug better. #2011

hpatro: @sarthakaggarwal97 Btw warning would increase the severity of logging. I want to reduce it.

sarthakaggarwal97: ack! #2011 makes sense to me!

madolson:
I don't feel strongly either way, so would appreciate input @enjoy-binbin once he is back from vacation.

enjoy-binbin:
ok, in your environment, the nodes will frequently enter and exit the pfail/fail state? The main reason i made this change was to better track changes in node state.

hpatro:
> ok, in your environment, the nodes will frequently enter and exit the pfail/fail state? The main reason i made this change was to better track changes in node state.

My concern is with cluster of large size, this log statement becomes quite ineffective by logging for each primary's report and can cause spikes in CPU utilisation.

enjoy-binbin:
> My concern is with cluster of large size, this log statement becomes quite ineffective by logging for each primary's report and can cause spikes in CPU utilisation.

Have you already encountered this problem in a real way or is it just an issue you see in the test logs? If the node frequently pfail/fail, the log issue should be relaively less serious?

Or you are saying, a node is actually dying, it is too heavy for the cluster to print the log in each nodes (like if we have 256 shards)? I suppose that can be some cases.

Anyway, i don't have a strong opinion, lets seek other opinions. @PingXie @zuiderkwast Sorry to ping you with this small (old) change, please feel free to leave a comment.

The loglevel default is notice
```
# verbose (many rarely useful info, but not a mess like the debug level)
# notice (moderately verbose, what you want in production probably)
```

zuiderkwast:
It's hard to guess the scenarios users will encounter. To say yes to a PR like this, I would like a more thorough explanation of the possible scenarios and the effect of this.

When a node is down in a 256-shard cluster, each node will detect that it's down and then they will also receive 254 gossips that say that it's not reachable. It can be too verbose, I guess. So in what scenarios do you get this?

1. A node has actually crashed, is overloaded or has a network problem, etc.
2. The admin is taking down the node on purpose, for example to upgrade it.
3. Anything else?

And do users rely on it? I don't know. Probably it's enough to log as NOTICE when it's marked as FAIL.

sarthakaggarwal97:
I missed that this PR exists, but I also experienced a lot of compute just going into logging. Sharing the issue here: https://github.com/valkey-io/valkey/issues/2076

hpatro:
> Have you already encountered this problem in a real way or is it just an issue you see in the test logs? If the node frequently pfail/fail, the log issue should be relatively less serious?

Most of this is testing Valkey in large cluster setup and we want to avoid unnecessary resource utilization as much as possible. It's not about the node flip-flopping but in case of AZ failure we risk logging lot of information which is not really valuable.

> Or you are saying, a node is actually dying, it is too heavy for the cluster to print the log in each nodes (like if we have 256 shards)? I suppose that can be some cases.

Yes, a single node going down is still fine. an entire AZ going down, is very expensive based on our testing.

hpatro:
@enjoy-binbin / @zuiderkwast Shall we merge this and work on https://github.com/valkey-io/valkey/pull/2011 for better observability of cluster state?

enjoy-binbin (approved):
ok, i think i can take this

madolson (approved):
Didn't approve earlier, but I'm okay with it

sarthakaggarwal97 (approved):

PR #2011, opened 2025-04-26
Log failed cluster node(s) state periodically to capture transient state for debuggability

hpatro:
This PR logs CLUSTER INFO / CLUSTER NODES output every 5 seconds to the log file for verbose/debug loglevel mode.

Certain times few nodes are not in convergence with the entire cluster and there are no logs captured about the divergence. This logging could help us better analyze in test setup where we can aggressively log more cluster information.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -5533,6 +5533,24 @@ void clusterCron(void) {
     }
 
     if (update_state || server.cluster->state == CLUSTER_FAIL) clusterUpdateState();
+
+    /* Log cluster info and nodes information for better debugability of the transient cluster state */
+    if (server.verbosity <= LL_VERBOSE) {
+        run_with_period(5000) {
+            sds cluster_info = genClusterInfoString();
+            sds cluster_nodes = clusterGenNodesDescription(NULL, 0, 0);
+
+            sds infostring = sdscatprintf(sdsempty(), "\r\n# Cluster info\r\n");
+            infostring = sdscatsds(infostring, cluster_info);
+            infostring = sdscatprintf(infostring, "\n------ CLUSTER NODES OUTPUT ------\n");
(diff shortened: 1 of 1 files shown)


src/cluster_legacy.c
```diff
@@ -5533,6 +5533,24 @@ void clusterCron(void) {
...
 
     if (update_state || server.cluster->state == CLUSTER_FAIL) clusterUpdateState();
+
+    /* Log cluster info and nodes information for better debugability of the transient cluster state */
+    if (server.verbosity <= LL_VERBOSE) {
+        run_with_period(5000) {
+            sds cluster_info = genClusterInfoString();
+            sds cluster_nodes = clusterGenNodesDescription(NULL, 0, 0);
+
+            sds infostring = sdscatprintf(sdsempty(), "\r\n# Cluster info\r\n");
```
madolson, line 5543: New lines break log parsing. If someone turns on for some reason, it should still be "valid" log lines. I'm OK logging the state, but I don't think it should just be verbatim the info fields.

hpatro: Interesting. So, it's an exception for crash report to have new lines?

hpatro: @madolson and I discussed offline. We discussed we can have a single log line with information around failed nodes rather than all the nodes.

enjoy-binbin:
Is the main purpose for debugging? ie someone find the cluster is not normal and adjust the loglevel to verbose and catch it?

hpatro:
> Is the main purpose for debugging? ie someone find the cluster is not normal and adjust the loglevel to verbose and catch it?

Yes. Even to investigate incident which occurred in the past it's quite difficult for operators to figure out the issue with the current state of logging. I would like this to be active at NOTICE level with failed nodes information which is actually relevant https://github.com/valkey-io/valkey/pull/2011#discussion_r2067059493

PR #2012, opened 2025-04-26
Flip pfail flag while marking node as failed

hpatro:
This fail logic was added in #1191, we should also clear the
pfail flag in this case.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3534,6 +3534,7 @@ int clusterProcessPacket(clusterLink *link) {
                  * the clients, and the replica will never initiate a failover since the
                  * node is not actually in FAIL state. */
                 if (!nodeFailed(noaddr_node)) {
+                    noaddr_node->flags &= ~CLUSTER_NODE_PFAIL;
                     noaddr_node->flags |= CLUSTER_NODE_FAIL;
                     noaddr_node->fail_time = now;
                     clusterSendFail(noaddr_node->name);


zuiderkwast (commented):
Why?

Can this difference be observed by users at all?

hpatro:
> Why?
> 
> Can this difference be observed by users at all?

As far as I understand, the cluster node state transition from PFAIL to FAIL and both of these are mutually exclusive and shouldn't exist together. So, I think finding the below response under `CLUSTER NODES` is incorrect.

```bash
> ./valkey-cli cluster nodes | grep fail?
511676acd16696d5b8767aa6c4a8ba54c69b70a0 :0@0 master,fail?,fail,noaddr - 1745372213541 0 1993 disconnected
```

sarthakaggarwal97:
looks like we do a similar change here in [markNodeAsFailingIfNeeded](https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L2197). It is called in clusterCron and through cluster gossip.

zuiderkwast (approved):
Yeah, makes sense to me.

enjoy-binbin (approved):
LGTM. btw, how do you reproduce it?

sarthakaggarwal97 (approved):

zuiderkwast:
Backport this one? Is it a follow-up of #1191 which is included in 8.1.

PR #2013, opened 2025-04-27
Prefetch keys in MSET and MGET

zuiderkwast:
For multikey commands, prefetching the keys from the hash table to L1 cache can speed up the commands.

This change executes MGET and MSET in batches of 16 keys. For each batch, the keys are prefetched.

With the following benchmark, I got 13% higher throughput for MSET and 17% for MGET, when MSET and MGET are called with 10 keys each time.

    taskset -c 0 ./valkey-server --save ""
    taskset -c 1-6 ./valkey-benchmark --csv -r 1000000 -n 10000000 --threads 6 -t mset,mget

The benchmark test for MGET is added in #2015.

### DIFF
--- a/src/memory_prefetch.c
+++ b/src/memory_prefetch.c
@@ -282,13 +282,13 @@ void removeClientFromPendingCommandsBatch(client *c) {
 
 /* A simple way to prefetch the keys for a client command execution. This can be
  * used for optimizing multi-key commands like mget, mset, etc. */
-void prefetchKeys(client *c, int first, int step) {
+void prefetchKeys(client *c, int first, int step, int count) {
     /* Skip this for IO threads. Keys are already batch-prefetched. */
     if (c->io_parsed_cmd) return;
 
     int n = (c->argc - first) / step;
-    if (n < 2) return;  /* There's no benefit in prefetcing only one key. */
-    if (n > 32) n = 32; /* Prefetching too many keys doesn't help the cache. */
+    if (count < n) n = count;
+    if (n < 2) return; /* There's no benefit in prefetcing only one key. */
     hashtableIncrementalFindState states[n];
     int i, arg;
     for (i = 0, arg = first; i < n && arg < c->argc; i++, arg += step) {
--- a/src/memory_prefetch.c
+++ b/src/memory_prefetch.c
@@ -279,3 +279,30 @@ void removeClientFromPendingCommandsBatch(client *c) {
         }
     }
 }
+
+/* A simple way to prefetch the keys for a client command execution. This can be
+ * used for optimizing multi-key commands like mget, mset, etc. */
+void prefetchKeys(client *c, int first, int step) {
+    /* Skip this for IO threads. Keys are already batch-prefetched. */
+    if (c->io_parsed_cmd) return;
+
+    int n = (c->argc - first) / step;
+    if (n < 2) return;  /* There's no benefit in prefetcing only one key. */
+    if (n > 32) n = 32; /* Prefetching too many keys doesn't help the cache. */
+    hashtableIncrementalFindState states[n];
+    int i, arg;
+    for (i = 0, arg = first; i < n && arg < c->argc; i++, arg += step) {
+        sds key = c->argv[arg]->ptr;
+        int slot = server.cluster_enabled ? getKeySlot(key) : 0;
+        hashtable *ht = kvstoreGetHashtable(c->db->keys, slot);
+        if (unlikely(ht == NULL || hashtableSize(ht) == 0)) return; /* Nothing here. */
+        hashtableIncrementalFindInit(&states[i], ht, key);
+    }
+    int not_finished;
+    do {
+        not_finished = 0;
+        for (i = 0; i < n; i++) {
+            not_finished += hashtableIncrementalFindStep(&states[i]);
+        }
+    } while (not_finished != 0);
+}
--- a/src/t_string.c
+++ b/src/t_string.c
@@ -607,7 +607,7 @@ void getrangeCommand(client *c) {
 
 void mgetCommand(client *c) {
     int j;
-
+    prefetchKeys(c, 1, 1);
     addReplyArrayLen(c, c->argc - 1);
     for (j = 1; j < c->argc; j++) {
         robj *o = lookupKeyRead(c->db, c->argv[j]);
@@ -630,6 +630,9 @@ void msetGenericCommand(client *c, int nx) {
         addReplyErrorArity(c);
         return;
     }
+
+    /* Prefetch the keys from memory in parallel. */
+    prefetchKeys(c, 1, 2);
 
     /* Handle the NX flag. The MSETNX semantic is to return zero and don't
      * set anything if at least one key already exists. */
--- a/src/t_string.c
+++ b/src/t_string.c
@@ -606,10 +606,14 @@ void getrangeCommand(client *c) {
 }
 
 void mgetCommand(client *c) {
+    const int batch_size = 16;
     int j;
-    prefetchKeys(c, 1, 1);
     addReplyArrayLen(c, c->argc - 1);
     for (j = 1; j < c->argc; j++) {
+        if (j % batch_size == 1) {
+            /* Prefetch the next batch of keys */
+            prefetchKeys(c, j, 1, batch_size);
+        }
         robj *o = lookupKeyRead(c->db, c->argv[j]);
         if (o == NULL) {
             addReplyNull(c);
@@ -624,6 +628,7 @@ void mgetCommand(client *c) {
 }
 
 void msetGenericCommand(client *c, int nx) {
+    const int batch_size = 16;
     int j;
 
     if ((c->argc % 2) == 0) {
@@ -631,13 +636,14 @@ void msetGenericCommand(client *c, int nx) {
         return;
     }
 
-    /* Prefetch the keys from memory in parallel. */
-    prefetchKeys(c, 1, 2);
-
     /* Handle the NX flag. The MSETNX semantic is to return zero and don't
      * set anything if at least one key already exists. */
     if (nx) {
         for (j = 1; j < c->argc; j += 2) {
+            if (j % (batch_size * 2) == 1) {
+                /* Prefetch the next batch of keys */
+                prefetchKeys(c, j, 2, batch_size);
+            }
             if (lookupKeyWrite(c->db, c->argv[j]) != NULL) {
                 addReply(c, shared.czero);
                 return;
@@ -647,6 +653,10 @@ void msetGenericCommand(client *c, int nx) {
 
     int setkey_flags = nx ? SETKEY_DOESNT_EXIST : 0;
     for (j = 1; j < c->argc; j += 2) {
+        if (!nx && j % (batch_size * 2) == 1) {
+            /* Prefetch the next batch of keys */
+            prefetchKeys(c, j, 2, batch_size);
+        }
         robj *val = tryObjectEncoding(c->argv[j + 1]);
         setKey(c, c->db, c->argv[j], &val, setkey_flags);
         incrRefCount(val);
--- a/src/memory_prefetch.h
+++ b/src/memory_prefetch.h
@@ -8,4 +8,6 @@ void processClientsCommandsBatch(void);
 int addCommandToBatchAndProcessIfFull(struct client *c);
 void removeClientFromPendingCommandsBatch(struct client *c);
 
+void prefetchKeys(struct client *c, int first, int step);
+
 #endif /* MEMORY_PREFETCH_H */
--- a/src/memory_prefetch.h
+++ b/src/memory_prefetch.h
@@ -8,6 +8,6 @@ void processClientsCommandsBatch(void);
 int addCommandToBatchAndProcessIfFull(struct client *c);
 void removeClientFromPendingCommandsBatch(struct client *c);
 
-void prefetchKeys(struct client *c, int first, int step);
+void prefetchKeys(struct client *c, int first, int step, int count);
 
 #endif /* MEMORY_PREFETCH_H */


src/memory_prefetch.c
```diff
@@ -279,3 +279,30 @@ void removeClientFromPendingCommandsBatch(client *c) {
...
+    if (c->io_parsed_cmd) return;
+
+    int n = (c->argc - first) / step;
+    if (count < n) n = count;
+    if (n < 2) return; /* There's no benefit in prefetcing only one key. */
+    hashtableIncrementalFindState states[n];
+    int i, arg;
+    for (i = 0, arg = first; i < n && arg < c->argc; i++, arg += step) {
+        sds key = c->argv[arg]->ptr;
+        int slot = server.cluster_enabled ? getKeySlot(key) : 0;
```
ranshid, line 296: Isn't the slot when cluster mode is disabled be -1?

xbasel:
I performed a quick test,  `MGET` with 1k keys, and observed degradation. 
Perf  shows prefetch version shifts some work from findBucket into hashtableIncrementalFindStep, but total mgetCommand CPU cost increased slightly.

@zuiderkwast, can you please double-check the scenario below?

Hardware: ARM (Graviton 3)
Results:
```
Before:
  100000 requests completed in 16.41 seconds
  50 parallel clients
  23017 bytes payload
  keep alive: 1
  multi-thread: no
...

Summary:
  throughput summary: 6094.59 requests per second
  latency summary (msec):
        7.330     2.136     7.319     8.591     9.071    10.431



After:
  100000 requests completed in 19.68 seconds
  50 parallel clients
  23017 bytes payload
  keep alive: 1
  multi-thread: no
...
Summary:
  throughput summary: 5080.27 requests per second
  latency summary (msec):
        8.934     1.192     8.943     9.735    10.031    15.151
```


Perf without prefetch:
```
               main
               aeMain
               |
               |--95.12%--connSocketEventHandler
               |          |
               |           --95.09%--readQueryFromClient
               |                     |
               |                     |--89.20%--processInputBuffer
               |                     |          |
               |                     |          |--56.24%--processCommand
               |                     |          |          |
               |                     |          |           --56.12%--call
               |                     |          |                     |
               |                     |          |                      --55.82%--mgetCommand
               |                     |          |                                |
               |                     |          |                                |--21.55%--findBucket.lto_priv.0
               |                     |          |                                |          |
               |                     |          |                                |          |--5.67%--hashtableSdsKeyCompare
               |                     |          |                                |          |          |
               |                     |          |                                |          |           --2.99%--sdscmp
               |                     |          |                                |          |                     |
               |                     |          |                                |          |                      --1.32%--memcmp
               |                     |          |                                |          |
               |                     |          |                                |           --3.61%--hashtableObjectGetKey
               |                     |          |                                |
               |                     |          |                                |--14.65%--addReplyBulk
               |                     |          |                                |          |
               |                     |          |                                |          |--4.31%--addReply
               |                     |          |                                |          |          |
               |                     |          |                                |          |           --1.20%--prepareClientToWrite
               |                     |          |                                |          |
               |                     |          |                                |          |--3.95%--_addReplyToBufferOrList.part.0
               |                     |          |                                |          |          |
               |                     |          |                                |          |           --1.44%--__GI___memcpy_simd
               |                     |          |                                |          |
               |                     |          |                                |          |--2.23%--prepareClientToWrite
               |                     |          |                                |          |          |
               |                     |          |                                |          |           --0.51%--clientHasPendingReplies
               |                     |          |                                |          |
               |                     |          |                                |          |--1.85%--_addReplyLongLongWithPrefix.lto_priv.0
               |                     |          |                                |          |
               |                     |          |                                |          |--0.74%--stringObjectLen
               |                     |          |                                |          |
               |                     |          |                                |           --0.57%--_addReplyToBufferOrList
               |                     |          |                                |
               |                     |          |                                |--3.02%--_addReplyToBufferOrList.part.0
               |                     |          |                                |          |
               |                     |          |                                |           --1.07%--__GI___memcpy_simd
               |                     |          |                                |
               |                     |          |                                |--2.74%--siphash
               |                     |          |                                |
               |                     |          |                                |--1.38%--expireIfNeededWithDictIndex.lto_priv.0
               |                     |          |                                |
               |                     |          |                                 --0.57%--_addReplyToBufferOrList
```

MGET with prefetch:
```
               main
               aeMain
               |
               |--96.17%--connSocketEventHandler
               |          |
               |           --96.13%--readQueryFromClient
               |                     |
               |                     |--90.71%--processInputBuffer
               |                     |          |
               |                     |          |--63.56%--processCommand
               |                     |          |          |
               |                     |          |           --63.46%--call
               |                     |          |                     |
               |                     |          |                      --63.08%--mgetCommand
               |                     |          |                                |
               |                     |          |                                |--15.45%--findBucket.lto_priv.0
               |                     |          |                                |          |
               |                     |          |                                |          |--4.99%--hashtableSdsKeyCompare
               |                     |          |                                |          |          |
               |                     |          |                                |          |           --2.87%--sdscmp
               |                     |          |                                |          |                     |
               |                     |          |                                |          |                      --1.19%--memcmp
               |                     |          |                                |          |
               |                     |          |                                |           --1.13%--hashtableObjectGetKey
               |                     |          |                                |
               |                     |          |                                |--14.61%--mgetCommand
               |                     |          |                                |
               |                     |          |                                |--12.52%--hashtableIncrementalFindStep
               |                     |          |                                |          |
               |                     |          |                                |          |--3.96%--hashtableSdsKeyCompare
               |                     |          |                                |          |          |
               |                     |          |                                |          |           --1.94%--sdscmp
               |                     |          |                                |          |                     |
               |                     |          |                                |          |                      --0.66%--memcmp
               |                     |          |                                |          |
               |                     |          |                                |           --1.24%--hashtableObjectGetKey
               |                     |          |                                |
               |                     |          |                                |--11.34%--addReplyBulk
               |                     |          |                                |          |
               |                     |          |                                |          |--3.25%--addReply
               |                     |          |                                |          |          |
               |                     |          |                                |          |           --0.98%--prepareClientToWrite
               |                     |          |                                |          |
               |                     |          |                                |          |--3.16%--_addReplyToBufferOrList.part.0
               |                     |          |                                |          |          |
               |                     |          |                                |          |           --0.98%--__GI___memcpy_simd
               |                     |          |                                |          |
               |                     |          |                                |          |--1.85%--prepareClientToWrite
               |                     |          |                                |          |
               |                     |          |                                |          |--1.32%--_addReplyLongLongWithPrefix.lto_priv.0
               |                     |          |                                |          |
               |                     |          |                                |           --0.56%--stringObjectLen
               |                     |          |                                |
               |                     |          |                                |--3.99%--siphash
               |                     |          |                                |
               |                     |          |                                |--2.40%--_addReplyToBufferOrList.part.0
               |                     |          |                                |          |
               |                     |          |                                |           --0.63%--__GI___memcpy_simd
               |                     |          |                                |
               |                     |          |                                |--1.20%--expireIfNeededWithDictIndex.lto_priv.0
               |                     |          |                                |
               |                     |          |                                |--0.68%--hashtableSdsHash
               |                     |          |                                |
               |                     |          |                                 --0.57%--_addReplyToBufferOrList
```

Command used:
```
src/valkey-benchmark -h <ip> -n 1000000    $(cat mget_command.txt)
```
`mget_command.txt` contains `MGET 1k keys`.
Each value is 3 bytes.

src/memory_prefetch.c
xbasel: `slot` must be 0. -1 will can  `kvstoreGetHashtable` (accessing array[-1])

I think `getKVStoreIndexForKey` can be used here.

zuiderkwast: getKVStoreIndexForKey is static in db.c so either we make it non-static or I can just rename `slot` to `kv_index`?

zuiderkwast:
@xbasel That's surprising. I modified valkey-benchmark MSET and MGET tests (from #2015) to run with 1000 keys. I got much better throughput with prefetching, but I'm just running on a laptop...

unstable

```
$ taskset -c 1-6 ./valkey-benchmark --csv -r 1000000 -n 10000 --threads 6 -t mset,mget
"test","rps","avg_latency_ms","min_latency_ms","p50_latency_ms","p95_latency_ms","p99_latency_ms","max_latency_ms"
"MSET (1000 keys)","1592.36","28.423","3.520","27.231","32.079","85.567","159.231"
"MGET (1000 keys)","2210.43","19.773","5.416","20.015","21.903","22.863","24.671"
```

prefetch

```
$ taskset -c 1-6 ./valkey-benchmark --csv -r 1000000 -n 10000 --threads 6 -t mset,mget
"test","rps","avg_latency_ms","min_latency_ms","p50_latency_ms","p95_latency_ms","p99_latency_ms","max_latency_ms"
"MSET (1000 keys)","1987.68","22.103","4.304","21.023","24.831","68.415","150.655"
"MGET (1000 keys)","3599.71","11.182","3.512","11.175","13.871","16.063","19.807"
```

xbasel:
> @xbasel That's surprising. I modified valkey-benchmark MSET and MGET tests (from #2015) to run with 1000 keys. I got much better throughput with prefetching, but I'm just running on a laptop...
> 
> unstable
> 
> ```
> $ taskset -c 1-6 ./valkey-benchmark --csv -r 1000000 -n 10000 --threads 6 -t mset,mget
> "test","rps","avg_latency_ms","min_latency_ms","p50_latency_ms","p95_latency_ms","p99_latency_ms","max_latency_ms"
> "MSET (1000 keys)","1592.36","28.423","3.520","27.231","32.079","85.567","159.231"
> "MGET (1000 keys)","2210.43","19.773","5.416","20.015","21.903","22.863","24.671"
> ```
> 
> prefetch
> 
> ```
> $ taskset -c 1-6 ./valkey-benchmark --csv -r 1000000 -n 10000 --threads 6 -t mset,mget
> "test","rps","avg_latency_ms","min_latency_ms","p50_latency_ms","p95_latency_ms","p99_latency_ms","max_latency_ms"
> "MSET (1000 keys)","1987.68","22.103","4.304","21.023","24.831","68.415","150.655"
> "MGET (1000 keys)","3599.71","11.182","3.512","11.175","13.871","16.063","19.807"
> ```

Your input is random, I think this could be related. I'll test with random input.

xbasel:
> Your input is random, I think this could be related. I'll test with random input.

I tested with random input and saw improvement. When the data is already cached, prefetching is just overhead. The data set I used was too small.

src/memory_prefetch.c
xbasel: not worth changing db.c. I'd just rename slot.

src/t_string.c
```diff
@@ -606,10 +606,14 @@ void getrangeCommand(client *c) {
 }
 
 void mgetCommand(client *c) {
+    const int batch_size = 16;
```
sarthakaggarwal97, line 609: should we make this global to avoid multiple declarations?

zuiderkwast: I considered using the existing config `prefetch-batch-max-size` but according to documentation, it's specifically for IO threading.

I can change to a define, but I don't know if we should use the same number in all multi-key commands. I guess it depends how many other operations happen in between that can evict the prefetched keys from L1 cache.

src/memory_prefetch.c
```diff
@@ -279,3 +279,30 @@ void removeClientFromPendingCommandsBatch(client *c) {
         }
     }
 }
+
+/* A simple way to prefetch the keys for a client command execution. This can be
+ * used for optimizing multi-key commands like mget, mset, etc. */
+void prefetchKeys(client *c, int first, int step, int count) {
+    /* Skip this for IO threads. Keys are already batch-prefetched. */
```
xbasel, line 286: Not all keys are prefetched in the context of io-threads. Up to 16 keys.

zuiderkwast: That's right. What do you suggest? Move the check to the caller and skip prefetching the first 16 if io-thread are used?

zuiderkwast: I wanted to optimize for ~10 keys. MGET with thousands of keys isn't a good pattern since it starves other clients, so that's less important IMO.

xbasel:
This change effectively does this (at last for `MGET`, I only looked at `MGET`):

```
for (key in keys) {
    prefetch_16_keys_every_16_iterations();
    val = lookupKeyRead(key);
}
```
The prefetch phase already does the heavy lifting - walking the hashtable and locating entries. That work is then repeated by `lookupKeyRead`.  

Yes, some (likely most) data stays in L1, but there's still wasted effort and potential stalls for uncached or evicted entries, or just due to the code path.

I don’t see why we can’t fetch the actual values during prefetch and reuse them. The current "scan and discard" pattern makes sense in IO threads context (pre-proessing) where lookup happens later, but here it would be cleaner and more efficient to return the data directly.

I profiled the execution and found:
```
--33.62%--mgetCommand
          |
          |--16.80%--prefetchKeys
          |          |
          |          |--13.97%--hashtableIncrementalFindStep
          |          |          |
          |          |          |--6.21%--compareKeys (inlined)
          |          |          |          |
          |          |          |           --6.06%--hashtableSdsKeyCompare
          |          |          |                     |
          |          |          |                      --5.49%--sdscmp
          |          |          |                                |
          |          |          |                                 --4.79%--__memcmp_evex_movbe
          |          |          |
          |          |           --2.55%--entryGetKey (inlined)
          |          |                     |
          |          |                      --2.45%--hashtableObjectGetKey
          |          |                                |
          |          |                                 --2.44%--objectGetKey (inlined)
          |          |
          |           --1.75%--hashtableIncrementalFindInit (inlined)
          |                     hashtableIncrementalFindInit (inlined)
          |                     |
          |                      --1.72%--hashKey (inlined)
          |                                |
          |                                 --1.33%--siphash
          |
          |--8.90%--lookupKeyRead (inlined)
          |          lookupKeyReadWithFlags (inlined)
          |          |
          |           --8.73%--lookupKey
          |                     |
          |                      --7.31%--dbFindWithDictIndex (inlined)
          |                                |
          |                                 --7.18%--kvstoreHashtableFind (inlined)
          |                                           |
          |                                            --7.09%--hashtableFind
          |                                                      |
          |                                                      |--4.73%--findBucket (inlined)
          |                                                      |          |
          |                                                      |          |--1.84%--compareKeys (inlined)
          |                                                      |          |          |
          |                                                      |          |           --1.67%--hashtableSdsKeyCompare
          |                                                      |          |                     |
          |                                                      |          |                     |--0.96%--sdscmp
          |                                                      |          |                     |
          |                                                      |          |                      --0.63%--sdslen (inl

          |                                                      |          |
          |                                                      |           --1.22%--entryGetKey (inlined)
          |                                                      |
          |                                                       --1.87%--hashKey (inlined)
          |                                                                 |
          |                                                                  --1.48%--siphash
          |
```

It seems `hashtableFind` is still causing memory stalls.

I modified the code slightly to make `lookupKeyRead` redundant by changing:

```
void prefetchKeys(client *c, int first, int step, int count, void **out)
```

and using out to iterate over the entries directly.

I'm not 100% sure my code is correct, it was a quick and dirty experiment, but I achieved this:
```
Summary:
  throughput: 4432.23 requests/sec
  latency (ms):
       avg     min     p50     p95     p99     max
     10.203   3.312  10.103  11.991  13.807  25.695
```

Compared to the original version using prefetchKeys and lookupKeyRead:
```
Summary:
  throughput: 3497.24 requests/sec
  latency (ms):
       avg     min     p50     p95     p99     max
     12.934   5.824  12.887  14.815  16.271  21.535
```

Again, no guarantees my code is correct, but it’s a direction worth exploring. If you're interested, I can try to clean it up and submit it here or in a separate branch.

src/memory_prefetch.c
xbasel: Maybe MGET with large number of keys isn't a good pattern, however, clients still send large key batches in practice.

I suggest either skipping the first 16 keys if IO threads is enabled or avoiding lookupKey entirely, see my other comment.

src/t_string.c
xbasel: I don't think the batch size should be global - it should be per-command/unit, possibly hardcoded, since the optimal value likely varies across commands and code paths.

zuiderkwast:
> I modified the code slightly to make `lookupKeyRead` redundant by changing:

Yeah, I'm aware that batch-prefetching is just a hack, not a perfect solution.

`lookupKey()` has side effects like eviction and key hits/misses stats. We would need a version of lookupKey that only does the that part, without actually looking up the key itself.

I also considered a `lookupMultipleKeys()` that looks up multiple keys in parallel.

MSET doesn't use `lookupKey()` but it uses `setKey()`. For this, we could have a `setMultipleKeys()` variant that does it in parallel. It's more optimal but more complex too, to get it nice and simple and avoid duplicating code.

xbasel:
> > I modified the code slightly to make `lookupKeyRead` redundant by changing:
> 
> Yeah, I'm aware that batch-prefetching is just a hack, not a perfect solution.
> 
> `lookupKey()` has side effects like eviction and key hits/misses stats. We would need a version of lookupKey that only does the that part, without actually looking up the key itself.
> 
> I also considered a `lookupMultipleKeys()` that looks up multiple keys in parallel.
> 
> MSET doesn't use `lookupKey()` but it uses `setKey()`. For this, we could have a `setMultipleKeys()` variant that does it in parallel. It's more optimal but more complex too, to get it nice and simple and avoid duplicating code.


I might have slightly underestimated the complexity. I think it's fine to start with the current code and optimize later.

src/t_string.c
sarthakaggarwal97: true, it should be per command, but does mget and mset can only have same batch_size?

zuiderkwast:
I believe this is a better and more generic approach:

* #2092

zuiderkwast:
This optimization is mostly covered by #2092, at least for MGET/MSET with up to 16 keys, but we can still revive this PR to optimize MGET/MSET with more than 16 keys, if this scenario with many keys is relevant. This PR needs to be rebased in that case.

zuiderkwast:
Superseded by #3389.

PR #2015, opened 2025-04-27
valkey-benchmark: Add test for MGET

zuiderkwast:
There is already a test for MSET with 10 keys per batch. This change adds one for MGET.

### DIFF
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -2016,6 +2016,19 @@ int main(int argc, char **argv) {
             sdsfree(key_placeholder);
         }
 
+        if (test_is_selected("mget")) {
+            const char *cmd_argv[11];
+            cmd_argv[0] = "MGET";
+            sds key_placeholder = sdscatprintf(sdsnew(""), "key%s:__rand_int__", tag);
+            for (i = 1; i < 11; i++) {
+                cmd_argv[i] = key_placeholder;
+            }
+            len = redisFormatCommandArgv(&cmd, 11, cmd_argv, NULL);
+            benchmark("MGET (10 keys)", cmd, len);
+            free(cmd);
+            sdsfree(key_placeholder);
+        }
+
         if (test_is_selected("xadd")) {
             len = redisFormatCommand(&cmd, "XADD mystream%s * myfield %s", tag, data);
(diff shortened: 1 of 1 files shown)


Nikhil-Manglore (approved):

madolson (approved):
Sure.

zuiderkwast:
Thank. It's trivial so I merged it immediately. Now we can use this for benchmarking the upcoming MGET optimizations. :)

src/valkey-benchmark.c
```diff
@@ -2016,6 +2016,19 @@ int main(int argc, char **argv) {
...
+            sds key_placeholder = sdscatprintf(sdsnew(""), "key%s:__rand_int__", tag);
+            for (i = 1; i < 11; i++) {
+                cmd_argv[i] = key_placeholder;
+            }
+            len = redisFormatCommandArgv(&cmd, 11, cmd_argv, NULL);
+            benchmark("MGET (10 keys)", cmd, len);
+            free(cmd);
+            sdsfree(key_placeholder);
+        }
+
```
hpatro, line 2031: How does this work with cluster mode?

zuiderkwast: `%s` expands to a cluster tag, right?

src/valkey-benchmark.c
```diff
@@ -2016,6 +2016,19 @@ int main(int argc, char **argv) {
...
 
+        if (test_is_selected("mget")) {
+            const char *cmd_argv[11];
+            cmd_argv[0] = "MGET";
+            sds key_placeholder = sdscatprintf(sdsnew(""), "key%s:__rand_int__", tag);
+            for (i = 1; i < 11; i++) {
+                cmd_argv[i] = key_placeholder;
+            }
+            len = redisFormatCommandArgv(&cmd, 11, cmd_argv, NULL);
+            benchmark("MGET (10 keys)", cmd, len);
```
sarthakaggarwal97, line 2027: should we have added a test with more than 16 keys? not sure if it would matter tho

zuiderkwast: I just did 10 to match the existing benchmark for MSET.

In another PR, we can add a new argument such as `-m` to control how many keys to use in MSET and MGET?

PR #399, opened 2024-04-28
Remove lpAssertValidEntry from listpack functions to achieve up to 10% improvement on `HSET` command.

ashtul:
This PR come after a similar PR was declined on Redis repo redis/redis#11273 with additional info at redis/redis#11293.

The issue it comes to solve is excess checks for corrupted data whenever a listpack is traversing. There are flamecharts in the original PR show it can reach 10% improvement of common commands such a `HSET` 

The reasoning against the change is that the data can be corrupted if the `RESTORE` command was used and with flag `SANITIZE_DUMP_NO`.

IMO this isn't justified. To make all users pay a significant patently because a user can potentially load data without sanitizing it first, seems excessive. Anyone who does it must consider the server may crush.

This simple change can give ValKey a nice speed boost. 

image

NOTE: the current tests fail b/c there are tests which load corrupted listpacks without sanitizing them first.

PR #2027, opened 2025-04-29
Additional assertions for listpack

muelstefamzn:
Adds a few asserts to increase the chance of detecting corrupted listpack data.

It replaces a questionable code path that purposefully returns invalid data with a `panic()` and adds `assert()` statements to verify that `LP_EOF` is only detected at the end of the listpack. 

Listpack data is expected to be correctly encoded. These asserts should only trigger if the listpack data was corrupted in some way, in which case it is better to assert than to return invalid data to the client.

### DIFF
--- a/src/listpack.c
+++ b/src/listpack.c
@@ -43,6 +43,7 @@
 #include "listpack_malloc.h"
 #include "serverassert.h"
 #include "util.h"
+#include "config.h"
 
 #define LP_HDR_SIZE 6 /* 32 bit total len + 16 bit number of elements. */
 #define LP_HDR_NUMELE_UNKNOWN UINT16_MAX
@@ -395,8 +396,13 @@ unsigned char *lpSkip(unsigned char *p) {
 unsigned char *lpNext(unsigned char *lp, unsigned char *p) {
     assert(p);
     p = lpSkip(p);
-    if (p[0] == LP_EOF) return NULL;
-    lpAssertValidEntry(lp, lpBytes(lp), p);
+    size_t bytes = lpBytes(lp);
+    if (unlikely(p[0] == LP_EOF)) {
+        /* EOF must only appear at the end of a listpack. */
+        assert(p + 1 == lp + bytes);
+        return NULL;
+    }
+    lpAssertValidEntry(lp, bytes, p);
     return p;
 }
 
@@ -418,8 +424,13 @@ unsigned char *lpPrev(unsigned char *lp, unsigned char *p) {
  * listpack has no elements. */
 unsigned char *lpFirst(unsigned char *lp) {
     unsigned char *p = lp + LP_HDR_SIZE; /* Skip the header. */
-    if (p[0] == LP_EOF) return NULL;
-    lpAssertValidEntry(lp, lpBytes(lp), p);
+    size_t bytes = lpBytes(lp);
+    if (unlikely(p[0] == LP_EOF)) {
+        /* EOF must only appear at the end of a listpack. */
+        assert(p + 1 == lp + bytes);
+        return NULL;
+    }
+    lpAssertValidEntry(lp, bytes, p);
     return p;
 }
 
@@ -482,13 +493,8 @@ unsigned long lpLength(unsigned char *lp) {
  * bytes, the element data itself, and the backlen bytes.
  *
  * If the function is called against a badly encoded ziplist, so that there
- * is no valid way to parse it, the function returns like if there was an
- * integer encoded with value 12345678900000000 + <unrecognized byte>, this may
- * be an hint to understand that something is wrong. To crash in this case is
- * not sensible because of the different requirements of the application using
- * this lib.
- *
- * Similarly, there is no error returned since the listpack normally can be
+ * is no valid way to parse it, the function panics.
+ * There is no error returned since the listpack normally can be
  * assumed to be valid, so that would be a very high API cost. */
 static inline unsigned char *
 lpGetWithSize(unsigned char *p, int64_t *count, unsigned char *intbuf, uint64_t *entry_size) {
@@ -540,9 +546,7 @@ lpGetWithSize(unsigned char *p, int64_t *count, unsigned char *intbuf, uint64_t 
         if (entry_size) *entry_size = 5 + *count + lpEncodeBacklen(NULL, *count + 5);
         return p + 5;
     } else {
-        uval = 12345678900000000ULL + p[0];
-        negstart = UINT64_MAX;
-        negmax = 0;
+        panic("Invalid listpack encoding. Byte %02hhx is not a valid encoding.", p[0]);
     }
 
     /* We reach this code path only for integer encodings.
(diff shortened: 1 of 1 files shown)


src/listpack.c
```diff
@@ -540,9 +545,7 @@ lpGetWithSize(unsigned char *p, int64_t *count, unsigned char *intbuf, uint64_t
         if (entry_size) *entry_size = 5 + *count + lpEncodeBacklen(NULL, *count + 5);
         return p + 5;
     } else {
-        uval = 12345678900000000ULL + p[0];
-        negstart = UINT64_MAX;
-        negmax = 0;
+        panic("Invalid listpack encoding");
```
xbasel: This is a breaking change.

src/listpack.c
```diff
@@ -395,8 +395,13 @@ unsigned char *lpSkip(unsigned char *p) {
 unsigned char *lpNext(unsigned char *lp, unsigned char *p) {
     assert(p);
     p = lpSkip(p);
-    if (p[0] == LP_EOF) return NULL;
-    lpAssertValidEntry(lp, lpBytes(lp), p);
+    size_t bytes = lpBytes(lp);
+    if (p[0] == LP_EOF) {
```
xbasel: I'd suggest using `__builtin_expect()`  here to hint that this branch is unlikely, which may help the compiler optimize branch prediction and layout.
This applies to other if-statements introduced in this pr.

xbasel:
Few questions:

1. Is this addressing an issue that has occurred in practice (e.g. corruption caught in real-world usage)?
2. The code adds assertions in a hot path. Do we have any benchmarks or profiling to understand the performance penalty of adding these assertions?
3. How has this change been tested?

src/listpack.c
smurn: The `if` conditions already existed before. I only added the assert in the body of the condition. The `assert()` macro is using `__builtin_expect()`.

muelstefamzn: The `if` conditions all existed previously. I added the `assert()` macro invocation, which is already using `__builtin_expect()` internally.

src/listpack.c
muelstefamzn: This code path is only reached if the listpack memory is corrupted. 
If Valkey considers behaviour changes in a corrupted memory scenario breaking, then yes.

The previous behaviour causes invalid data to be returned from `lpGet()` and `lpFind()` would usually not find the searched element unless it happens to be `12345678900000000ULL+x`, in which case it will consider the invalid element a hit.
Find will also attempt to continue processing the rest of the corrupted listpack, likey resulting in undefined behaviour or a crash.

Let me know how you'd like to handle this.

muelstefamzn:
Thank you for reviewing it!

I see a few reasons why listpack data could be invalid, all have low probability:
 * Memory corruption due to a hardware defect.
 * Snapshot corruption during storage/transfer that is not caught by the checksum.
 * A bug that creates invalid data. I'm not aware of such a bug in Valkey.
 
If such a situation occurs, I prefer to panic/assert over returning invalid data or undefined behaviour. Of course, this change does not guarantee that it will be caught, just makes it more likely.

I've done a simple valkey-benchmark test:
 * Pre-filled the node with 100k hash keys with four fields each, containing integers.
 * Ran `valkey-benchmark -r 100000 -n 1000000 HINCRBY __rand_int__ field 1`.
   This exercises `lpFirst()`, `lpFind()`, `lpNext()`, `lpGetWithSize()`, and others.
 * Repeated the test 5x with and without the change.

I did not observe a significant change in performance.
* Without the change: 83157.334 RPS (standard deviation +-381.4)
* With the change: 83417.814 RPS  (standard deviation +-332.5)

This is hardly a comprehensive performance test, of course. At least for this setup, the difference
is less than the standard deviation (+-0.4%). By chance, averaged a little higher with the change.

Other than running the existing tests, I've manually spot checked some of the asserts by attaching
a debugger and modifying the listpack data in-situ to contain the `LP_EOF` byte at invalid locations.
The asserts trigger as expected, creating a crash report.

src/listpack.c
madolson: I think we normally use `unlikely` as opposed to __builtin_expect(). I agree though, I don't think we need to do it.

madolson: I suppose the unlikely might improve performance a bit, especially since we are effectively adding a second if in the assertion path.

src/listpack.c
madolson: Maybe too much to ask, but can we add a panicprintf("Invalid listpack encoding %b", p[0]); ? I would like to have some more information.

src/listpack.c
```diff
@@ -540,9 +547,8 @@ lpGetWithSize(unsigned char *p, int64_t *count, unsigned char *intbuf, uint64_t
         if (entry_size) *entry_size = 5 + *count + lpEncodeBacklen(NULL, *count + 5);
         return p + 5;
     } else {
-        uval = 12345678900000000ULL + p[0];
-        negstart = UINT64_MAX;
-        negmax = 0;
+        serverLog(LL_WARNING, "Invalid listpack encoding. Byte %02hhx is not a valid encoding.", p[0]);
```
ranshid: I would suggest avoid adding server dependency in this stage so we can keep the listpack as a standalone DS as much as possible. probably with @madolson suggestion (later) to use panicprintf we can have the same data printed.

src/listpack.c
```diff
@@ -43,6 +43,8 @@
 #include "listpack_malloc.h"
 #include "serverassert.h"
 #include "util.h"
+#include "config.h"
```
ranshid, line 46: why is config needed here?

muelstefamzn: for `unlikely()`.

src/listpack.c
muelstefamzn: There is no `panicprintf()` function, afaik. I had to add `server.h` for `serverLog()`.

madolson: Apparently the args of panic is `panic(const char * fmt, ...)`, I thought for some reason it actually didn't take a format. I think we can just do `panic("Invalid listpack encoding. Byte %02hhx is not a valid encoding.", p[0])`.

src/listpack.c
madolson: That should be OK.

madolson (approved):
LGTM

src/listpack.c
```diff
@@ -540,9 +546,7 @@ lpGetWithSize(unsigned char *p, int64_t *count, unsigned char *intbuf, uint64_t
         if (entry_size) *entry_size = 5 + *count + lpEncodeBacklen(NULL, *count + 5);
         return p + 5;
     } else {
-        uval = 12345678900000000ULL + p[0];
-        negstart = UINT64_MAX;
-        negmax = 0;
+        panic("Invalid listpack encoding. Byte %02hhx is not a valid encoding.", p[0]);
```
madolson, line 549: So I suppose we can't do this. While validating listpack entries, we need to fetch some of the data out to validate the integrity. However, if the listpack itself is itself malformed, we might try to extract the data out. The alternative would be to check every single point in `streamValidateListpackIntegrity` to verify we still have more data to read.

zuiderkwast:
The fact that we allow loading corrupted listpacks from RDB is the source of this problem and complexity. We have already decided that we want to validate the dump before we load it, so we will not need to allow correpted listpacks to exist in the database anymore. The decision is here: https://github.com/valkey-io/valkey/pull/399#issuecomment-2310401817 but the PR was never finished.

PR #2030, opened 2025-04-30
Optimize findBucket performance using SIMD

zhulipeng:
### Description

This PR introduces an optimization for the `findBucket` function utilizing SSE2 instructions. The `findBucket` function is widely be used in hashtable data structure, particularly in **SET** and **GET** related commands, but also for looking up a command by its name and many other things.

The core optimization happens with `_mm_cmpeq_epi8()`, which performs 16 bytes (with 7 bytes being valid) comparisons in parallel. The `_mm_movemask_epi8()` function then converts this vector of byte comparisons into a bit mask where each bit represents whether a comparison was successful.

This implementation demonstrates sophisticated use of SIMD instructions to accelerate hash table lookups by performing multiple hash comparisons in parallel.

### Performance Boost

The corresponding **GET** and **SET** commands can gain up to **~2% - ~6%** performance improvement especially with pipeline enabled.  Below test scenarios showed they are benefit for this optimization. 

|Benchmark|Performance Boost|
|-|-|
|[memtier_benchmark-10Mkeys-string-get-10B-pipeline-100-nokeyprefix](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-10Mkeys-string-get-10B-pipeline-100-nokeyprefix.yml) |**4%**|
|[memtier_benchmark-1Mkeys-string-get-10B-pipeline-50](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-string-get-10B-pipeline-50.yml) |**4%**|
|[memtier_benchmark-1Mkeys-string-get-10B-pipeline-100](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-string-get-10B-pipeline-100.yml) |**4%**|
|[memtier_benchmark-1Mkeys-string-get-10B-pipeline-100-nokeyprefix](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-string-get-10B-pipeline-100-nokeyprefix.yml) |**2%**|
|[memtier_benchmark-1Mkeys-string-get-10B-pipeline-500](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-string-get-10B-pipeline-500.yml) |**6%**|
|[memtier_benchmark-1Mkeys-load-string-with-10B-values-pipeline-500](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-load-string-with-10B-values-pipeline-500.yml) |**4%**|
|[memtier_benchmark-1Mkeys-load-string-with-10B-values-pipeline-100](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-load-string-with-10B-values-pipeline-100.yml) |**3%**|
|[memtier_benchmark-1Mkeys-load-string-with-10B-values-pipeline-100-nokeyprefix](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1Mkeys-load-string-with-10B-values-pipeline-100-nokeyprefix.yml) |**3%**|

#### Test Env
- OS: CentOS Stream 9
- Platform: Intel Xeon 6980P

```
port 9001
bind * -::*
daemonize no
protected-mode no
save
```

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -382,9 +382,11 @@ void setcpuaffinity(const char *cpulist);
 #endif
 
 #if HAVE_X86_SIMD
+#define ATTRIBUTE_TARGET_SSE2 __attribute__((target("sse2")))
 #define ATTRIBUTE_TARGET_AVX2 __attribute__((target("avx2")))
 #define ATTRIBUTE_TARGET_AVX512 __attribute__((target("avx512f,avx512bw,avx512vl")))
 #else
+#define ATTRIBUTE_TARGET_SSE2
 #define ATTRIBUTE_TARGET_AVX2
 #define ATTRIBUTE_TARGET_AVX512
 #endif
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -678,7 +678,7 @@ static int expand(hashtable *ht, size_t size, int *malloc_failed) {
  * entry at that position matches the provided key. If a match is found, it
  * updates the position and table index pointers and returns 1. Otherwise,
  * it returns 0. */
-static inline int checkCandidateInBucket(hashtable *ht, bucket *b, int pos, const void *key, int *pos_in_bucket, int *table_index, int table) {
+static inline int checkCandidateInBucket(hashtable *ht, bucket *b, int pos, const void *key, int table, int *pos_in_bucket, int *table_index) {
     /* It's a candidate. */
     void *entry = b->entries[pos];
     const void *elem_key = entryGetKey(ht, entry);
@@ -694,7 +694,7 @@ static inline int checkCandidateInBucket(hashtable *ht, bucket *b, int pos, cons
 
 #if HAVE_X86_SIMD
 ATTRIBUTE_TARGET_SSE2
-static int findKeyInBucketSSE2(hashtable *ht, bucket *b, uint8_t h2, const void *key, int *pos_in_bucket, int *table_index, int table) {
+static int findKeyInBucketSSE2(hashtable *ht, bucket *b, uint8_t h2, const void *key, int table, int *pos_in_bucket, int *table_index) {
     /* Get the bucket's presence mask - indicates which positions are filled. */
     BUCKET_BITS_TYPE presence_mask = b->presence & ((1 << ENTRIES_PER_BUCKET) - 1);
     __m128i hash_vector = _mm_loadu_si128((__m128i *)b->hashes);
@@ -709,7 +709,7 @@ static int findKeyInBucketSSE2(hashtable *ht, bucket *b, uint8_t h2, const void 
     newmask &= presence_mask;
     while (newmask > 0) {
         int pos = __builtin_ctz(newmask);
-        if (checkCandidateInBucket(ht, b, pos, key, pos_in_bucket, table_index, table)) return 1;
+        if (checkCandidateInBucket(ht, b, pos, key, table, pos_in_bucket, table_index)) return 1;
         /* Clear the processed bit and continue with next match. */
         newmask &= ~(1 << pos);
     }
@@ -743,12 +743,12 @@ static bucket *findBucket(hashtable *ht, uint64_t hash, const void *key, int *po
         do {
 #if HAVE_X86_SIMD
             /* All x86-64 CPUs have SSE2. */
-            if (findKeyInBucketSSE2(ht, b, h2, key, pos_in_bucket, table_index, table)) return b;
+            if (findKeyInBucketSSE2(ht, b, h2, key, table, pos_in_bucket, table_index)) return b;
 #else
             /* Find candidate entries with presence flag set and matching h2 hash. */
             for (int pos = 0; pos < numBucketPositions(b); pos++) {
                 if (isPositionFilled(b, pos) && b->hashes[pos] == h2 &&
-                    checkCandidateInBucket(ht, b, pos, key, pos_in_bucket, table_index, table)) return b;
+                    checkCandidateInBucket(ht, b, pos, key, table, pos_in_bucket, table_index)) return b;
             }
 #endif
             b = getChildBucket(b);
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -675,7 +675,7 @@ static int expand(hashtable *ht, size_t size, int *malloc_failed) {
  *
  * If 'table_index' is provided, it is set to the index of the table (0 or 1)
  * the returned bucket belongs to. */
-#ifdef HAVE_X86_SIMD
+#if HAVE_X86_SIMD
 #include <immintrin.h>
 ATTRIBUTE_TARGET_SSE2
 #endif
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -672,6 +672,12 @@ static int expand(hashtable *ht, size_t size, int *malloc_failed) {
     return resize(ht, size, malloc_failed);
 }
 
+/* Checks if a candidate entry in a bucket matches the given key.
+ *
+ * This function examines a specific position in a bucket to determine if the
+ * entry at that position matches the provided key. If a match is found, it
+ * updates the position and table index pointers and returns 1. Otherwise,
+ * it returns 0. */
 static inline int checkCandidateInBucket(hashtable *ht, bucket *b, int pos, const void *key, int *pos_in_bucket, int *table_index, int table) {
     /* It's a candidate. */
     void *entry = b->entries[pos];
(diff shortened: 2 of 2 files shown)


xbasel:
I'm not sure %3-4 justifies the change and the additional complexity...
Would be interesting to see the impact in other scenarios, eg. lua with heavy findBucket use 

@zuiderkwast  ?

xbasel:
fyi - I haven't reviewed the SIMD change yet. I'll do that once we decide if it's worth adopting.

zuiderkwast:
I think it's great! 4% is good considering most of the CPU time is spent on IO. It's not many lines of code and not very complex. Only 20 lines if we ignore comment and preprocessor stuff like `#if`. I just took a quick look. I'll do a proper review later.

Can you add a micro benchmark in src/unit/test_hashtable.c so we can measure only the lookup? For example we can measure and printe the elapsed time in `add_find_delete_test_helper()`.

xbasel:
I'll check the performance gain on ARM.

madolson:
> I think it's great!

20 lines of code for improving a code path that is hit by basically everything is a great win.

src/hashtable.c
```diff
@@ -675,6 +675,10 @@ static int expand(hashtable *ht, size_t size, int *malloc_failed) {
  *
  * If 'table_index' is provided, it is set to the index of the table (0 or 1)
  * the returned bucket belongs to. */
+#ifdef HAVE_X86_SIMD
```
madolson:
```suggestion
#if HAVE_X86_SIMD
```

src/config.h
```diff
@@ -382,9 +382,11 @@ void setcpuaffinity(const char *cpulist);
 #endif
 
 #if HAVE_X86_SIMD
```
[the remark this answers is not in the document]
zuiderkwast: Copilot's comment is not very useful comment IMHO. :)

A better comment for these target macros could be

```C
/* Attributes to compile a function for a specific CPU with certain
 * features. Such code needs to be guarded with runtime checks like
 *
 *     if(__builtin_cpu_supports("avx2")) ...
 */
```

src/hashtable.c
```diff
@@ -675,6 +675,10 @@ static int expand(hashtable *ht, size_t size, int *malloc_failed) {
  *
  * If 'table_index' is provided, it is set to the index of the table (0 or 1)
  * the returned bucket belongs to. */
+#if HAVE_X86_SIMD
+#include <immintrin.h>
+ATTRIBUTE_TARGET_SSE2
+#endif
 static bucket *findBucket(hashtable *ht, uint64_t hash, const void *key, int *pos_in_bucket, int *table_index) {
```
zuiderkwast: Let's do the `#include` near the top of the file with the other includes.

Regarding the attribute, if we add the SSE2 attribute here, the compiler is free to generate SSE2 instructions even for the code that is not explicitly using the `_mm` functions, right? It seems to me that all code marked with the attribute should be guarded with a runtime check, in general.

src/hashtable.c
```diff
@@ -693,6 +697,39 @@ static bucket *findBucket(hashtable *ht, uint64_t hash, const void *key, int *po
         }
         bucket *b = &ht->tables[table][bucket_idx];
         do {
+#if HAVE_X86_SIMD
+            if (__builtin_cpu_supports("sse2")) {
```
zuiderkwast: If we want to guard the code SSE2-specific code with a runtime check, then we need to check the CPU support before we enter code marked with the SSE2 attribute, don't we?

We can create a helper function where we use the SSE2 attribute and call it here. It can be a static inline function.

```c
#if HAVE_X86_SIMD
            if (__builtin_cpu_supports("sse2")) {
                if (findKeyInBucketSSE2(b, key, h2, pos_in_bucket, table_index)) return b;
                b = getChildBucket(b);
                continue;
            }
#endif
            /* Fallback non-SIMD code here */
```

The helper function can have the attribute.

```c
#if HAVE_X86_SIMD
ATTRIBUTE_TARGET_SSE2 static findInBucketSSE2(...) {
    ...
}
#endif
```

However, SSE2 appears to be mandatory in all x86-64 CPUs, according to https://en.wikipedia.org/wiki/X86-64, so maybe we don't need to check `__builtin_cpu_supports("sse2")` at all? Only the `HAVE_X86_SIMD` check is enough?

If yes, then we can use `#else` for the non-SIMD code.

```C
#if HAVE_X86_SIMD
            /* All x64-64 CPUs have SSE2. */
            if (findKeyInBucketSSE2(b, key, h2, pos_in_bucket, table_index)) return b;
#else
            /* Fallback non-SIMD code here */
            ...
#endif
            b = getChildBucket(b);
```

Do you know if SSE2 always exists on x86-64?

src/hashtable.c
```diff
@@ -693,6 +697,39 @@ static bucket *findBucket(hashtable *ht, uint64_t hash, const void *key, int *po
...
+                    /* It's a candidate. */
+                    void *entry = b->entries[pos];
+                    const void *elem_key = entryGetKey(ht, entry);
+                    if (compareKeys(ht, key, elem_key) == 0) {
+                        /* It's a match. */
+                        assert(pos_in_bucket != NULL);
+                        *pos_in_bucket = pos;
+                        if (table_index) *table_index = table;
+                        return b;
+                    }
```
zuiderkwast: This is duplicated code. We can let the compiler duplicate the code instead. Let's move it to a `static inline` function and call it in both places.

```c
                    if (checkCandidateInBucket(b, pos, key, pos_in_bucket, table_index) return b;
```

zuiderkwast (commented):
The actual SSE2 code is nice and simple. I just have some comments about the overall structure. I may be wrong though, so please reply to my comments if they are not good.

src/config.h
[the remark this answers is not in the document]
madolson: Yeah, I sometimes ask it to do a review to see if it can figure why the build was failing. It's so far not been very good.

src/hashtable.c
zhulipeng: > Do you know if SSE2 always exists on x86-64?

Thanks. The SSE2 should be baseline for x86-64 arch and cannot be "disabled" on an x86-64 system.

zhulipeng:
@zuiderkwast Thanks for your review, sorry for the late response due to the Labor Day holiday.

>Can you add a micro benchmark in src/unit/test_hashtable.c so we can measure only the lookup? For example we can measure and printe the elapsed time in add_find_delete_test_helper.

I noticed that we already have a time-elapsed based micro benchmark implemented here, https://github.com/valkey-io/valkey/blob/unstable/src/unit/test_hashtable.c#L393-L403. Do you think this is sufficient?

Below is the test results based on the micro bench:

**w/ SSE2**
[test_incremental_find - unit/test_hashtable.c:401] Lookup 2000000 elements one by one took 75862 microseconds.

**w/o SSE2**
[test_incremental_find - unit/test_hashtable.c:401] Lookup 2000000 elements one by one took 101799 microseconds.

src/hashtable.c
```diff
@@ -669,16 +672,50 @@ static int expand(hashtable *ht, size_t size, int *malloc_failed) {
     return resize(ht, size, malloc_failed);
 }
 
+static inline int checkCandidateInBucket(hashtable *ht, bucket *b, int pos, const void *key, int *pos_in_bucket, int *table_index, int table) {
```
zuiderkwast: Let's put the return-by-pointer variables `pos_in_bucket` and `table_index` last. It's an informal convention.

~~Here is a suggestion for a comment to explain how use the function.~~

```suggestion
static inline int checkCandidateInBucket(hashtable *ht, bucket *b, int pos, const void *key, int table, int *pos_in_bucket, int *table_index) {
```

src/hashtable.c
```diff
@@ -669,16 +672,50 @@ static int expand(hashtable *ht, size_t size, int *malloc_failed) {
...
+        assert(pos_in_bucket != NULL);
+        *pos_in_bucket = pos;
+        if (table_index) *table_index = table;
+        return 1;
+    }
+    return 0;
+}
+
+#if HAVE_X86_SIMD
+ATTRIBUTE_TARGET_SSE2 static int findKeyInBucketSSE2(hashtable *ht, bucket *b, uint8_t h2, const void *key, int *pos_in_bucket, int *table_index, int table) {
```
zuiderkwast: Let's put the return-by-pointer variables last.

```suggestion
ATTRIBUTE_TARGET_SSE2 static int findKeyInBucketSSE2(hashtable *ht, bucket *b, uint8_t h2, const void *key, int table, int *pos_in_bucket, int *table_index) {
```

zuiderkwast (commented):
Great, now just very minor comments.

> I noticed that we already have a time-elapsed based micro benchmark implemented here, https://github.com/valkey-io/valkey/blob/unstable/src/unit/test_hashtable.c#L393-L403. Do you think this is sufficient?

Yes, it's sufficient. Nice! I forgot about it.

> Below is the test results based on the micro bench:
>
> w/ SSE2
> [test_incremental_find - unit/test_hashtable.c:401] Lookup 2000000 elements one by one took 75862 microseconds.
>
> w/o SSE2
> [test_incremental_find - unit/test_hashtable.c:401] Lookup 2000000 elements one by one took 101799 microseconds.

It's 25% faster and here there buckets are not in L1 cache. With cached small hashtable such as the command lookup table, it is probably even more gain. Very good!

src/hashtable.c
zuiderkwast: Ah, you added comments. :) Ignore my suggested comment.

zuiderkwast (approved):

madolson (approved):
Nice! This PR helped me learn a bit about the SIMD instructions, which isn't an area I'm very familiar with.

xbasel (approved):
Looks good to me - simple and well-structured code!

PR #2033, opened 2025-04-30
Allow dynamic modification of io-threads num

ayush933:
Item from #761 

This PR has the following changes

1. Bug fix where calling `pthread_join()` from main thread for an IO thread would hang indefinitely. This is because `IOThreadMain()` doesn't have a cancellation point.So `pthread_cancel()` from main thread is not honored. 
Can be reproed by calling `shutdownIOThread()` from the main thread for any active thread with empty job queue.
 Fixed by adding cancellation point in `IOThreadMain()`.
2. Makes `io-threads` config runtime modifiable.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -28,6 +28,7 @@
  * POSSIBILITY OF SUCH DAMAGE.
  */
 
+#include "io_threads.h"
 #include "server.h"
 #include "cluster.h"
 #include "connection.h"
@@ -3258,8 +3259,8 @@ standardConfig static_configs[] = {
     /* Integer configs */
     createIntConfig("databases", NULL, IMMUTABLE_CONFIG, 1, INT_MAX, server.config_databases, 16, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("cluster-databases", NULL, IMMUTABLE_CONFIG, 1, INT_MAX, server.config_databases_cluster, 1, INTEGER_CONFIG, NULL, NULL),
-    createIntConfig("port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.port, 6379, INTEGER_CONFIG, NULL, updatePort),                                   /* TCP port. */
-    createIntConfig("io-threads", NULL, DEBUG_CONFIG | IMMUTABLE_CONFIG, 1, IO_THREADS_MAX_NUM, server.io_threads_num, 1, INTEGER_CONFIG, NULL, NULL), /* Single threaded by default */
+    createIntConfig("port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.port, 6379, INTEGER_CONFIG, NULL, updatePort),                                               /* TCP port. */
+    createIntConfig("io-threads", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, 1, IO_THREADS_MAX_NUM, server.io_threads_num, 1, INTEGER_CONFIG, NULL, updateIOThreads), /* Single threaded by default */
     createIntConfig("events-per-io-thread", NULL, MODIFIABLE_CONFIG | HIDDEN_CONFIG, 0, INT_MAX, server.events_per_io_thread, 2, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("prefetch-batch-max-size", NULL, MODIFIABLE_CONFIG, 0, 128, server.prefetch_batch_max_size, 16, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("auto-aof-rewrite-percentage", NULL, MODIFIABLE_CONFIG, 0, INT_MAX, server.aof_rewrite_perc, 100, INTEGER_CONFIG, NULL, NULL),
--- a/src/io_threads.c
+++ b/src/io_threads.c
@@ -215,18 +215,19 @@ static void *IOThreadMain(void *myid) {
     long id = (long)myid;
     char thdname[32];
 
-    serverAssert(server.io_threads_num > 0);
-    serverAssert(id > 0 && id < server.io_threads_num);
     snprintf(thdname, sizeof(thdname), "io_thd_%ld", id);
     valkey_set_thread_title(thdname);
     serverSetCpuAffinity(server.server_cpulist);
     makeThreadKillable();
     initSharedQueryBuf();
+    pthread_cleanup_push(freeSharedQueryBuf, NULL);
 
     thread_id = (int)id;
     size_t jobs_to_process = 0;
     IOJobQueue *jq = &io_jobs[id];
     while (1) {
+        /* Cancellation point so that pthread_cancel() from main thread is honored. */
+        pthread_testcancel();
         /* Wait for jobs */
         for (int j = 0; j < 1000000; j++) {
             jobs_to_process = IOJobQueue_availableJobs(jq);
@@ -255,12 +256,15 @@ static void *IOThreadMain(void *myid) {
          * As the main-thread main concern is to check if the queue is empty, it's enough to do it once at the end. */
         atomic_thread_fence(memory_order_release);
     }
-    freeSharedQueryBuf();
+    pthread_cleanup_pop(0);
     return NULL;
 }
 
 #define IO_JOB_QUEUE_SIZE 2048
 static void createIOThread(int id) {
+    serverAssert(server.io_threads_num > 0);
+    serverAssert(id > 0 && id < server.io_threads_num);
+
     pthread_t tid;
     pthread_mutex_init(&io_threads_mutex[id], NULL);
     IOJobQueue_init(&io_jobs[id], IO_JOB_QUEUE_SIZE);
@@ -287,7 +291,7 @@ static void shutdownIOThread(int id) {
     } else {
         serverLog(LL_NOTICE, "IO thread(tid:%lu) terminated", (unsigned long)tid);
     }
-
+    pthread_mutex_destroy(&io_threads_mutex[id]);
     IOJobQueue_cleanup(&io_jobs[id]);
 }
 
@@ -295,6 +299,44 @@ void killIOThreads(void) {
     for (int j = 1; j < server.io_threads_num; j++) { /* We don't kill thread 0, which is the main thread. */
         shutdownIOThread(j);
     }
+}
+
+int updateIOThreads(const char **err) {
+    serverAssert(inMainThread());
+    UNUSED(err);
+    int prev_threads_num = 1;
+    for (int i = IO_THREADS_MAX_NUM - 1; i > 0; i--) {
+        if (io_threads[i]) {
+            prev_threads_num = i + 1;
+            break;
+        }
+    }
+    if (prev_threads_num == server.io_threads_num) return 1;
+
+    serverLog(LL_NOTICE, "Changing number of IO threads from %d to %d.", prev_threads_num, server.io_threads_num);
+    drainIOThreadsQueue();
+    /* Set active threads to 1, will be adjusted based on workload later. */
+    for (int i = 1; i < server.active_io_threads_num; i++) {
+        pthread_mutex_lock(&io_threads_mutex[i]);
+    }
+    server.active_io_threads_num = 1;
+
+    // Create new threads.
+    if (server.io_threads_num > prev_threads_num) {
+        for (int i = prev_threads_num; i < server.io_threads_num; i++) {
+            createIOThread(i);
+        }
+    }
+    // Decrease the number of threads.
+    else {
+        for (int i = prev_threads_num - 1; i >= server.io_threads_num; i--) {
+            // Unblock inactive thread.
+            pthread_mutex_unlock(&io_threads_mutex[i]);
+            shutdownIOThread(i);
+            io_threads[i] = 0;
+        }
+    }
+    return 1;
 }
 
 /* Initialize the data structures needed for I/O threads. */
--- a/src/server.h
+++ b/src/server.h
@@ -2769,7 +2769,7 @@ void linkClient(client *c);
 void protectClient(client *c);
 void unprotectClient(client *c);
 void initSharedQueryBuf(void);
-void freeSharedQueryBuf(void);
+void freeSharedQueryBuf(void *dummy);
 client *lookupClientByID(uint64_t id);
 int authRequired(client *c);
 void clientSetUser(client *c, user *u, int authenticated);
--- a/tests/unit/other.tcl
+++ b/tests/unit/other.tcl
@@ -574,6 +574,25 @@ start_server {tags {"other external:skip"}} {
     }
 }
 
+start_server {tags {"other external:skip"}} {
+    test "test io-threads are runtime modifiable" {
+        # Test set
+        r config set io-threads 5
+        set thread_num [lindex [r config get io-threads] 1]
+        assert_equal 5 $thread_num
+
+        # Test decrease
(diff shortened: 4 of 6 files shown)


ayush933:
UTs pending atm.

src/io_threads.c
```diff
@@ -297,6 +303,40 @@ void killIOThreads(void) {
...
+    serverAssert(inMainThread());
+    UNUSED(err);
+    int prev_threads_num = 1;
+    for (int i = IO_THREADS_MAX_NUM - 1; i > 0; i--) {
+        if (io_threads[i]) {
+            prev_threads_num = i + 1;
+            break;
+        }
+    }
+    if (prev_threads_num == server.io_threads_num) goto done;
```
```suggestion
    if (prev_threads_num == server.io_threads_num) return 1;
```
There is no other logic in done, so we can just return here.

src/io_threads.c
```diff
@@ -227,6 +231,8 @@ static void *IOThreadMain(void *myid) {
     size_t jobs_to_process = 0;
     IOJobQueue *jq = &io_jobs[id];
     while (1) {
+        // Cancellation point so that pthread_cancel() from main thread is honored.
```
```suggestion
        /* Cancellation point so that pthread_cancel() from main thread is honored. */
```
Although we allow both, this is the more common style in the codebase.

madolson (commented):
Nice. Just minor things, looking forward to the tests.

src/io_threads.c
```diff
@@ -297,6 +303,40 @@ void killIOThreads(void) {
...
+    // Create new threads.
+    if (server.io_threads_num > prev_threads_num) {
+        for (int i = prev_threads_num; i < server.io_threads_num; i++) {
+            createIOThread(i);
+        }
+    }
+    // Decrease the number of threads.
+    else {
+        for (int i = prev_threads_num - 1; i >= server.io_threads_num; i--) {
+            drainIOThreadsQueueInternal(i);
```
uriyage: I think it would be safer to first drain all the queues, regardless of whether we increase or decrease the number of threads. Since we don't expect this to happen very often, performance is not an issue.

src/io_threads.c
```diff
@@ -297,6 +303,40 @@ void killIOThreads(void) {
...
+            // Unblock inactive thread.
+            if (i >= server.active_io_threads_num)
+                pthread_mutex_unlock(&io_threads_mutex[i]);
+            shutdownIOThread(i);
+            io_threads[i] = 0;
+        }
+    }
+    server.active_io_threads_num = min(server.active_io_threads_num, server.io_threads_num);
```
uriyage: It would be safer to just reset it to 1 (server.active_io_threads_num = 1), and later we can update it according to the workload.

src/config.c
```diff
@@ -3264,8 +3265,8 @@ standardConfig static_configs[] = {
 
     /* Integer configs */
     createIntConfig("databases", NULL, IMMUTABLE_CONFIG, 1, INT_MAX, server.dbnum, 16, INTEGER_CONFIG, NULL, NULL),
-    createIntConfig("port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.port, 6379, INTEGER_CONFIG, NULL, updatePort),                                   /* TCP port. */
-    createIntConfig("io-threads", NULL, DEBUG_CONFIG | IMMUTABLE_CONFIG, 1, IO_THREADS_MAX_NUM, server.io_threads_num, 1, INTEGER_CONFIG, NULL, NULL), /* Single threaded by default */
+    createIntConfig("port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.port, 6379, INTEGER_CONFIG, NULL, updatePort),                                               /* TCP port. */
+    createIntConfig("io-threads", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, 1, IO_THREADS_MAX_NUM, server.io_threads_num, 1, INTEGER_CONFIG, NULL, updateIOThreads), /* Single threaded by default */
```
uriyage, line 3263: Can we add tests to verify behavior when increasing or decreasing the number of threads

ayush933:
Hey @uriyage, just curious , can you please explain the safety concerns a bit more i.e. why we would want to drain all jobs and inactivate all threads instead of the min num required?

uriyage:
> Hey @uriyage, just curious , can you please explain the safety concerns a bit more i.e. why we would want to drain all jobs and inactivate all threads instead of the min num required?

@ayush933 I don't recall the exact workflow, but I remember attempting something similar, and it affected the logic of dynamically adjusting the threads or the client `cur_tid` logic. I might be mistaken, but I think it will be cleaner and safer.

tests/unit/other.tcl
```diff
@@ -574,6 +574,25 @@ start_server {tags {"other external:skip"}} {
     }
 }
 
+start_server {tags {"other external:skip"}} {
+    test "test io-threads are runtime modifiable" {
+        #Test set
+        r config set io-threads 5
```
uriyage, line 580: Consider adding a check for cases when we switch between single I/O thread (main thread only) and multiple I/O threads

ayush933: Is just changing the 3 to 1 in this test enough?

ayush933:
Hey @uriyage @madolson , anything else needed to move this PR forward?

tests/unit/other.tcl
```diff
@@ -574,6 +574,25 @@ start_server {tags {"other external:skip"}} {
...
+
+        #Test decrease
+        r config set io-threads 1
+        set thread_num [lindex [r config get io-threads] 1]
+        assert_equal 1 $thread_num
+
+        #Test increase
+        r config set io-threads 4
+        set thread_num [lindex [r config get io-threads] 1]
+        assert_equal 4 $thread_num
```
madolson, line 592:
```suggestion
        # Test set
        r config set io-threads 5
        set thread_num [lindex [r config get io-threads] 1]
        assert_equal 5 $thread_num

        # Test decrease
        r config set io-threads 1
        set thread_num [lindex [r config get io-threads] 1]
        assert_equal 1 $thread_num

        # Test increase
        r config set io-threads 4
        set thread_num [lindex [r config get io-threads] 1]
        assert_equal 4 $thread_num
```
Unwritten style comment.

src/server.h
```diff
@@ -2769,7 +2769,7 @@ void linkClient(client *c);
 void protectClient(client *c);
 void unprotectClient(client *c);
 void initSharedQueryBuf(void);
-void freeSharedQueryBuf(void);
+void freeSharedQueryBuf(void *);
```
```suggestion
void freeSharedQueryBuf(void *v);
```

tests/unit/other.tcl
```diff
@@ -574,6 +574,25 @@ start_server {tags {"other external:skip"}} {
     }
 }
 
+start_server {tags {"other external:skip"}} {
+    test "test io-threads are runtime modifiable" {
+        #Test set
```
madolson:
```suggestion
        # Test set
```

tests/unit/other.tcl
```diff
@@ -574,6 +574,25 @@ start_server {tags {"other external:skip"}} {
...
+        r config set io-threads 5
+        set thread_num [lindex [r config get io-threads] 1]
+        assert_equal 5 $thread_num
+
+        #Test decrease
```
madolson:
```suggestion
        # Test decrease
```

tests/unit/other.tcl
```diff
@@ -574,6 +574,25 @@ start_server {tags {"other external:skip"}} {
...
+        r config set io-threads 1
+        set thread_num [lindex [r config get io-threads] 1]
+        assert_equal 1 $thread_num
+
+        #Test increase
```
madolson:
```suggestion
        # Test increase
```

madolson (approved):
Just some nits, but still would like @uriyage to take sign off as well, he knows more about this code.

src/io_threads.c
```diff
@@ -297,6 +300,44 @@ void killIOThreads(void) {
...
+    int prev_threads_num = 1;
+    for (int i = IO_THREADS_MAX_NUM - 1; i > 0; i--) {
+        if (io_threads[i]) {
+            prev_threads_num = i + 1;
+            break;
+        }
+    }
+    if (prev_threads_num == server.io_threads_num) return 1;
+
+    serverLog(LL_DEBUG, "Changing number of IO threads from %d to %d.", prev_threads_num, server.io_threads_num);
```
uriyage: Consider changing it to `LL_NOTICE` since it is not expected to occur frequently and could be helpful when it appears in the logs

uriyage (approved):

uriyage:
> Just some nits, but still would like @uriyage to take sign off as well, he knows more about this code.

LGTM. I left one small comment about the log verbosity

tests/unit/other.tcl
```diff
@@ -574,6 +574,25 @@ start_server {tags {"other external:skip"}} {
     }
 }
 
+start_server {tags {"other external:skip"}} {
```
madolson, line 577: Looks like this test caught something:
https://github.com/valkey-io/valkey/actions/runs/14957133612/job/42014310397?pr=2033#step:5:1403

```
io_threads.c:219 'id > 0 && id < server.io_threads_num' is not true
```

It looks like we are probably updating this value on the main thread while the thread is getting initialized on another thread. Maybe we should move these assertions to `createIOThread`. @uriyage Any other ideas?

uriyage: Yes, we can move it to the main-thread.

ayush933: Done.

src/server.h
```diff
@@ -2769,7 +2769,7 @@ void linkClient(client *c);
 void protectClient(client *c);
 void unprotectClient(client *c);
 void initSharedQueryBuf(void);
-void freeSharedQueryBuf(void);
+void freeSharedQueryBuf(void *v);
```
zuiderkwast: It's difficult to understand what `v` is. My first thought is that `v` is the shared query buffer, but it's just an unused dummy variable. Call it `unused` or `dummy` or `junk`?

zuiderkwast (approved):
Conceptually approve. I didn't review the code (just looked briefly).

> Bug fix where calling `pthread_join()` from main thread for an IO thread would hang indefinitely

What is the severity of this bug? Which scenario triggers it, from a user's perspective? Do we need to backport the bugfix? (If we need that, may need to split out the bugfix to a separate commit that we can backport, since this PRs is a new feature + a bugfix. We should not backport the new feature.)

ayush933:
Before this feature, the only time we called `shutdownIoThread()` was on server shutdown. 
Considering this wasn't noticed until now, I don't think its too severe and needs a backport but lets wait for @uriyage's input on this.

uriyage:
> Before this feature, the only time we called `shutdownIoThread()` was on server shutdown. Considering this wasn't noticed until now, I don't think its too severe and needs a backport but lets wait for @uriyage's input on this.

Currently, it is only used by `doFastMemoryTest`, which is called from `printCrashReport` after a crash occurs,  in this case the threads are suspended by the `ThreadsManager_runOnThreads`  so not an issue.

src/config.c
```diff
@@ -3258,8 +3259,8 @@ standardConfig static_configs[] = {
     /* Integer configs */
     createIntConfig("databases", NULL, IMMUTABLE_CONFIG, 1, INT_MAX, server.config_databases, 16, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("cluster-databases", NULL, IMMUTABLE_CONFIG, 1, INT_MAX, server.config_databases_cluster, 1, INTEGER_CONFIG, NULL, NULL),
-    createIntConfig("port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.port, 6379, INTEGER_CONFIG, NULL, updatePort),                                   /* TCP port. */
-    createIntConfig("io-threads", NULL, DEBUG_CONFIG | IMMUTABLE_CONFIG, 1, IO_THREADS_MAX_NUM, server.io_threads_num, 1, INTEGER_CONFIG, NULL, NULL), /* Single threaded by default */
+    createIntConfig("port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.port, 6379, INTEGER_CONFIG, NULL, updatePort),                                               /* TCP port. */
+    createIntConfig("io-threads", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, 1, IO_THREADS_MAX_NUM, server.io_threads_num, 1, INTEGER_CONFIG, NULL, updateIOThreads), /* Single threaded by default */
```
zhulipeng, line 3263: I have a question about runtime thread allocation in Valkey; please correct me if I misunderstood.

For example, if a user provisions a system with 4 vCPUs and initially configures io-threads:4 at startup, everything operates efficiently with balanced resource utilization. However, what happens if they dynamically increase the I/O thread count beyond 4 while the server is running?

Would this runtime thread expansion impact resource availability for existing threads and cause significant performance degradation? Specifically, given Valkey's busy loop implementation, would this oversubscription scenario be particularly problematic?

zuiderkwast: Potentially yes.

Using more IO threads than available CPUs is a bad idea. There is a comment in valkey.conf about this.

https://github.com/valkey-io/valkey/blob/unstable/valkey.conf#L1435-L1457

The comments is quite outdated though and some improvement to this text would be appreciated. It says that you can double the thoughput with IO threads, which is an understatement. Also it doesn't mention that IO threads can offload the main thead with things like TLS handshake and various other things...

zhulipeng: While we can trust the admin to follow best practices during server startup, providing such a feature could pose a potential risk to the server. Alternatively, we could determine the CPU affinity and get a dynamic maximum number of I/O threads, regardless of whether it is scheduled by `taskset`, `numactl`, `server-cpulist` or by utilizing all available CPUs. At the very least, we should limit the maximum number of I/O threads based on the runtime environment. What do you think?

zuiderkwast: > At the very least, we should limit the maximum number of I/O threads based on the runtime environment. What do you think?

We can do this later? We don't need to make the maximum value dynamic because it makes the same config work on some machines and refuse to start on other machines. Instead, the server can just not start more IO threads than the number of CPUs, even if the configured number is higher. Maybe also log a warning in that case.

Can we merge this PR and do the runtime limit as a follow-up?

madolson: I think the limits would apply to the static configurations as well? If you set the max number of threads to 100 and the system only has 4 cores, we should probably gracefully limit the threads, but that is already a problem today. It's also worth mentioning that you could be in an over subscribed situation where cores can be dynamically re-allocated to your VM or container. We should be thoughtful about how we limit it.

I still think we should merge it as is and we can add better limits later.

madolson:
New crash https://github.com/valkey-io/valkey/actions/runs/15312340723/job/43079580795?pr=2149#step:4:1444

src/io_threads.c
```diff
@@ -297,6 +301,44 @@ void killIOThreads(void) {
     }
 }
 
+int updateIOThreads(const char **err) {
```
uriyage, line 304: When starting with a single thread, `initIOThreads` returns early without calling `prefetchCommandsBatchInit`, which we need to make sure we call it if we start with 1 then increase it

madolson: I'll grab that as well

zuiderkwast: This is related to the work to parse multiple pipelined commands into a queue. When we do that, we'll do prefetching regardless of the number of IO threads, even io threads disabled, because we can have multiple commands from a single client.

PR #2036, opened 2025-05-01
Properly escape double quotes and backslash in sdscatrepr

### LINKED ISSUE #2035: [BUG] `sdscatrepr` not escaping `"` and `\` characters anymore
When executing `monitor`, special characters such as `"`, `\` and others should be escaped properly. This used to be the case up to version `8.0.3`, but with `8.1.0`, commands such as `echo 'foo"bar'` and `echo "foo\\"` produce an incorrect monitor output of

```
+1746085153.283830 [0 127.0.0.1:55236] "echo" "foo"bar"
+1746085167.954618 [0 127.0.0.1:55236] "echo" "foo\"
```

whereas on `8.0.3`, the same commands produced a correct monitor output of

```
+1746085205.816900 [0 127.0.0.1:55272] "echo" "foo\"bar"
+1746085206.927962 [0 127.0.0.1:55272] "echo" "foo\\"
```

A `git bisect` shows that this bug was introduced in commit https://github.com/valkey-io/valkey/commit/cf1a1e0931bd2db77c23b0058d8660461873dc8f

The bug introduced in this commit is caused by the `isprint(*p)` returning a positive result for the `"` and `\` characters such that the switch block inside the else block is never reached for those.

Fusl:
Fixes https://github.com/valkey-io/valkey/issues/2035, a bug introduced in https://github.com/valkey-io/valkey/pull/1342

### DIFF
--- a/tests/unit/introspection.tcl
+++ b/tests/unit/introspection.tcl
@@ -678,6 +678,15 @@ start_server {tags {"introspection"}} {
         set _ $res
     } {*"set" "foo"*"get" "foo"*}
 
+	test {MONITOR properly escapes special characters through sdscatrepr} {
+		set rd [valkey_deferring_client]
+		$rd monitor
+		assert_match {*OK*} [$rd read]
+		r echo "backslash\\quotes\"newline\ncarriagereturn\rtab\talert\abackspace\bhexnormal\x7Ahexspecial\x7F"
+		assert_match {*"echo" "backslash\\\\quotes\\"newline\\ncarriagereturn\\rtab\\talert\\abackspace\\bhexnormalzhexspecial\\x7f"*} [$rd read]
+		$rd close
+	}
+
     test {MONITOR can log commands issued by the scripting engine} {
         set rd [valkey_deferring_client]
         $rd monitor
--- a/tests/unit/introspection.tcl
+++ b/tests/unit/introspection.tcl
@@ -678,14 +678,14 @@ start_server {tags {"introspection"}} {
         set _ $res
     } {*"set" "foo"*"get" "foo"*}
 
-	test {MONITOR properly escapes special characters through sdscatrepr} {
-		set rd [valkey_deferring_client]
-		$rd monitor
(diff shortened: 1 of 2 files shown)


zuiderkwast (approved):
LGTM, thanks!

@madolson FYI. Apparently you reviewed the PR where the bug was introduced.

tests/unit/introspection.tcl
```diff
@@ -678,6 +678,15 @@ start_server {tags {"introspection"}} {
...
 
+	test {MONITOR properly escapes special characters through sdscatrepr} {
+		set rd [valkey_deferring_client]
+		$rd monitor
+		assert_match {*OK*} [$rd read]
+		r echo "backslash\\quotes\"newline\ncarriagereturn\rtab\talert\abackspace\bhexnormal\x7Ahexspecial\x7F"
+		assert_match {*"echo" "backslash\\\\quotes\\"newline\\ncarriagereturn\\rtab\\talert\\abackspace\\bhexnormalzhexspecial\\x7f"*} [$rd read]
+		$rd close
+	}
+
```
madolson:
```suggestion
    test {MONITOR properly escapes special characters through sdscatrepr} {
        set rd [valkey_deferring_client]
        $rd monitor
        assert_match {*OK*} [$rd read]
        r echo "backslash\\quotes\"newline\ncarriagereturn\rtab\talert\abackspace\bhexnormal\x7Ahexspecial\x7F"
        assert_match {*"echo" "backslash\\\\quotes\\"newline\\ncarriagereturn\\rtab\\talert\\abackspace\\bhexnormalzhexspecial\\x7f"*} [$rd read]
        $rd close
    }

```

madolson (approved):

tests/unit/introspection.tcl
Fusl: My bad I have my editor configured for tab indent space align by default, thank you!

madolson:
We did have some unit tests for `sdscatrepr`, but apparently not enough. Thanks for catching this.

PR #2039, opened 2025-05-05
CI rename build-macos to build-old-version-macos

vitahlin:
This is a follow-up adjustment based on the previous PR: https://github.com/valkey-io/valkey/pull/2000
@hwware @madolson 
Apologies again for the extra review work this may have caused — I really appreciate your time and feedback.

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -1126,7 +1126,7 @@ jobs:
         if: true && !contains(github.event.inputs.skiptests, 'cluster')
         run: ./runtest-cluster ${{github.event.inputs.cluster_test_args}}
 
-  build-old-version-macos:
+  build-old-macos-versions:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


hwware (commented):
Do we need mark it as build-old-version-macos?  @madolson

.github/workflows/daily.yml
```diff
@@ -1126,11 +1126,11 @@ jobs:
         if: true && !contains(github.event.inputs.skiptests, 'cluster')
         run: ./runtest-cluster ${{github.event.inputs.cluster_test_args}}
 
-  build-macos:
+  build-old-version-macos:
```
madolson:
```suggestion
  build-old-macos-versions:
```
Is probably more straight forward.

madolson (commented):
We also need to update the reference to this job down in `notify-about-job-results`

madolson (approved):
Thanks!

PR #2043, opened 2025-05-05
Valkey-cli:  add missing -c to a test to fix resharding failures

xbasel:
A test was missing the `-c` flag for valkey-cli, which caused it to miscount the number of databases and led to resharding failures. This change adds the flag to fix the test. "-c" enables cluster mode in the cli.

Note: valkey-cli will be updated separately to not rely on `-c` to determine the number of databases.

### DIFF
--- a/src/t_hash.c
+++ b/src/t_hash.c
@@ -416,9 +416,6 @@ int hashTypeExists(robj *o, sds field) {
  * semantics of copying the values if needed.
  *
  */
-#define HASH_SET_TAKE_FIELD (1 << 0)
-#define HASH_SET_TAKE_VALUE (1 << 1)
-#define HASH_SET_COPY 0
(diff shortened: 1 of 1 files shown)


tests/cluster/tests/04-resharding.tcl
```diff
@@ -84,7 +84,7 @@ test "Cluster consistency during live resharding" {
             flush stdout
             set target [dict get [get_myself [randomInt 5]] id]
             set tribpid [lindex [exec \
-                ../../../src/valkey-cli --cluster reshard \
+                ../../../src/valkey-cli -c --cluster reshard \
```
zuiderkwast: This makes me think that if `--cluster` is provided, `valkey-cli` should be able to infer that it's in cluster mode.

zuiderkwast:
→ #2046

PR #2045, opened 2025-05-05
test

xbasel:


x
zuiderkwast, line 1: Don't add the files `x` and `y`.

(Never use `git add -a`.)

madolson:
I think this was related to the debugging of the cluster tests, so going to close.

PR #2046, opened 2025-05-05
valkey-cli: Count databases correctly in --cluster subcommands

xbasel:
Fixes a valkey-cli check that caused it to miscount the number of databases and led to resharding failures. This change makes the `--cluster` subcommands imply cluster mode as if the `-c` flag had been given.

### DIFF
--- a/tests/cluster/tests/04-resharding.tcl
+++ b/tests/cluster/tests/04-resharding.tcl
@@ -84,7 +84,7 @@ test "Cluster consistency during live resharding" {
             flush stdout
             set target [dict get [get_myself [randomInt 5]] id]
             set tribpid [lindex [exec \
-                ../../../src/valkey-cli -c --cluster reshard \
+                ../../../src/valkey-cli --cluster reshard \
                 127.0.0.1:[get_instance_attrib valkey 0 port] \
                 --cluster-from all \
                 --cluster-to $target \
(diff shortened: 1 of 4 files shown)


tests/cluster/tests/04-resharding.tcl
```diff
@@ -84,7 +84,7 @@ test "Cluster consistency during live resharding" {
             flush stdout
             set target [dict get [get_myself [randomInt 5]] id]
             set tribpid [lindex [exec \
-                ../../../src/valkey-cli --cluster reshard \
+                ../../../src/valkey-cli -c --cluster reshard \
```
madolson, line 87:
```suggestion
                ../../../src/valkey-cli -c --cluster reshard \
```

hwware:
I suggest you provide a link for this PR on your own repo daily CI test result.

tests/cluster/tests/04-resharding.tcl
zuiderkwast: Can't we make `--cluster` imply `-c`?

It looks redundant/silly to pass both these options.

xbasel: Maybe, but the priority now is to fix the build. I don't think this should be changed in this PR. We need to understand why '-c' was added in the first place.

zuiderkwast: > Maybe, but the priority now is to fix the build.

Yes, definitely.

> We need to understand why '-c' was added in the first place.

AFAIK:

* `-c` is for following redirects when calling simple commands like `valkey-cli -c get foo`.
* `--cluster` is for special cluster admin operations.

madolson: I agree with viktor, let's just make --cluster imply -c.

zuiderkwast (approved):

PR #2048, opened 2025-05-06
Return only the relevant bit in sdsGetAuxBit()

aradz44:
Previously, the function would return 1 if any auxiliary bit was set, which could cause problems when multiple auxiliary bits were present. This pull request modifies the function to return 1 or 0 based solely on the specific auxiliary bit being checked, rather than considering all auxiliary bits.


zuiderkwast (approved):
Thank you!

I have seen this bug. @ranshid saw it and we were talking about it but nobody fixed it until now.

Issue #2049, opened 2025-05-06
Test failure: Exception LOADING in valkey_deferring_client


zuiderkwast:
The `$client read` is just after sending a SELECT command. SELECT is allowed while loading so this is unexpected.

```
[exception]: Executing test client: LOADING Valkey is loading the dataset in memory.
LOADING Valkey is loading the dataset in memory
    while executing
"$client read"
    (procedure "valkey_deferring_client" line 18)
```

https://github.com/valkey-io/valkey/actions/runs/14860554436/job/41724086930#step:7:5321

This might have started to appear when #1671 was merged.

zuiderkwast:
@xbasel FYI

xbasel:
could be related to 
https://github.com/valkey-io/valkey/pull/1671/files#diff-0e26d3f37751154d568fba9aaa25e16b475454b6506d6ea8391de771c29cff04R452

xbasel:
This re produces the issue:
```
./runtest --single unit/cluster/cli --single integration/rdb  --fastfail  --clients 1
```


Output:
```
➜  valkey git:(fixFailuresAfterMultiDb2) ✗ ./runtest --single unit/cluster/cli --single integration/rdb  --fastfail  --clients 1
Cleanup: may take some time... OK
Starting test server at port 21079
[ready]: 100583
Testing unit/cluster/cli
[ok]: Create 1 node cluster (1944 ms)
[ok]: Create 2 node cluster (1040 ms)
zzzz singledb 1
[ok]: Create 3 node cluster (1941 ms)
[ok]: Run blocking command on cluster node3 (1 ms)
[ok]: Perform a Resharding (29 ms)
[ok]: Verify command got unblocked after resharding (65 ms)
[ok]: Wait for cluster to be stable (25 ms)
zzzz singledb 1
[ok]: use previous hostip in "cluster-preferred-endpoint-type unknown-endpoint" mode (6 ms)
[ok]: Sanity test push cmd after resharding (0 ms)
[ok]: Run blocking command again on cluster node1 (0 ms)
[ok]: Kill a cluster node and wait for fail state (1293 ms)
[ok]: Verify command got unblocked after cluster failure (51 ms)
[ok]: Functions are added to new node on valkey-cli cluster add-node (3175 ms)
[ok]: Migrate the last slot away from a node using valkey-cli (11293 ms)
[ok]: valkey-cli -4 --cluster create using 127.0.0.1 with cluster-port (1956 ms)
[...]

PR #2050, opened 2025-05-06
Use the correct port for migration

madolson:
One of the new tests that was added uses `CONFIG GET PORT`, which isn't right one for TLS.

Also removed some other use of the helper which aren't actually used.


zuiderkwast (approved):
It's from this one, yes. LGTM.

* https://github.com/valkey-io/valkey/pull/1671

Some of the other failures is this one:

* https://github.com/valkey-io/valkey/issues/2049

xbasel:
lgtm

PR #2056, opened 2025-05-06
Prevent tests setting singledb from breaking other tests 

zuiderkwast:
Some Tcl test suites set the global variable `::singledb`. If they don't restore the value afterwards, they can affect other test suites, which leads to problems that are hard to track down.

This change introduces a test tag 'singledb' that has the same effect, but it is local to the scope of the tag (a tags block, start_server or a single test case). The tests should use the tag instead of setting the global.

Additional change: Use ECHO instead of PING in `valkey_deferring_client` and `valkey_client` procs when singledb is set, to avoid failing when the server is loading. (SELECT and ECHO are allowed while loading, but not PING.)

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -103,6 +103,7 @@ The following compatibility and capability tags are currently used:
 | `needs:reset`             | Uses `RESET` to reset client connections. |
 | `needs:save`              | Uses `SAVE` or `BGSAVE` to create an RDB file. |
 | `needs:other-server`      | Requires `--other-server-path`. |
+| `singledb`                | Test runs with `--singledb`. |
 
 When using an external server (`--host` and `--port`), filtering using the
 `external:skip` tags is done automatically.
--- a/tests/README.md
+++ b/tests/README.md
@@ -103,7 +103,7 @@ The following compatibility and capability tags are currently used:
 | `needs:reset`             | Uses `RESET` to reset client connections. |
 | `needs:save`              | Uses `SAVE` or `BGSAVE` to create an RDB file. |
 | `needs:other-server`      | Requires `--other-server-path`. |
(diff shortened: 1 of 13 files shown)


tests/support/server.tcl
```diff
@@ -447,6 +447,7 @@ proc start_server {options {code undefined}} {
     set keep_persistence false
     set config_lines {}
     set start_other_server 0
+    set old_singledb $::singledb
```
xbasel, line 450: is this a backup of the initial value set on start up?

zuiderkwast: It's a backup of the previous value, the value it has when entering the `start_server` block. These things can be nested.

```
tags {singledb} {
    start_server {tags {singledb}} {
        test {bla bla bla} {
            ...
        } {} {singledb}
    }
}
```

These can be nested in any order and any number of times.

madolson (approved):
The global singledb stuff still annoys me, it would be nice if it wasn't global we had to toggle on/off, but instead set a random DB to connect to within the test. This seems like we're moving in the right direction though, since the tag is a better way to override test behavior, so LGTM.

tests/README.md
```diff
@@ -103,6 +103,7 @@ The following compatibility and capability tags are currently used:
 | `needs:reset`             | Uses `RESET` to reset client connections. |
 | `needs:save`              | Uses `SAVE` or `BGSAVE` to create an RDB file. |
 | `needs:other-server`      | Requires `--other-server-path`. |
+| `singledb`                | Test runs with `--singledb`. |
```
zuiderkwast:
```suggestion
| `singledb`                | Test runs as if `--singledb` was given. |
```

PR #2057, opened 2025-05-07
valkey-benchmark: Arbitrary command sequence

zuiderkwast:
The valkey-benchmark command line syntax is extended with the special token ";" to allow multiple commands in a sequence. The command sequence is sent together as a pipeline.

    valkey-benchmark set foo bar ';' get foo
    valkey-benchmark multi ';' set foo bar ';' get foo ';' exec

Each command can be prefixed with a number, indicating how many times to repeat the command. For example, to execute one set and four get commands, simulating 20/80 traffic:

    valkey-benchmark set foo bar ';' 4 get foo

An optional "--" argument is allowed to separate the options from the command sequence. This is allowed just for clarity.

    valkey-benchmark -n 1000 -- set foo bar ';' 4 get foo

A `__data__` placeholder is supported in the command arguments. This is expanded to a string of the data length given using the `-d` option. This example simulates SET and GET with a value of 512:

    valkey-benchmark -r 100000 -d 512 -- set key:__rand_int__ __data__ ';' get key:__rand_int__


hwware:
My only concern: is this improvement a break change? I notice the command line format updated from valkey-benchmark [OPTIONS] [COMMAND ARGS...] to valkey-benchmark [OPTIONS] [--] [COMMAND ARGS...].

hwware (approved):

zuiderkwast:
> My only concern: is this improvement a break change? I notice the command line format updated from valkey-benchmark [OPTIONS] [COMMAND ARGS...] to valkey-benchmark [OPTIONS] [--] [COMMAND ARGS...].

It's not a breaking change. What do you mean? Within square brackets means it's optional.

ranshid:
can you explain the interaction between the pipeline and the command sequence length? In case we have, for example, pipeline of 3 and a sequence of 5, do we expect the first 3 will always be pipelined and the next 2 commands will not?

zuiderkwast:
> can you explain the interaction between the pipeline and the command sequence length? In case we have, for example, pipeline of 3 and a sequence of 5, do we expect the first 3 will always be pipelined and the next 2 commands will not?

The sequence is always sent in a pipeline. It could be a transaction, etc. and I don't want to break it into pieces.

When combining with -P 3 for pipelining, the sequence is repeated 3 times, so 15 commands are sent in one pipeline in this case.

ranshid:
> > can you explain the interaction between the pipeline and the command sequence length? In case we have, for example, pipeline of 3 and a sequence of 5, do we expect the first 3 will always be pipelined and the next 2 commands will not?
> 
> The sequence is always sent in a pipeline. It could be a transaction, etc. and I don't want to break it into pieces.
> 
> When combining with -P 3 for pipelining, the sequence is repeated 3 times, so 15 commands are sent in one pipeline in this case.

I c. so maybe we need to take care of the -n accounting? 
```
 int requests_issued = atomic_fetch_add_explicit(&config.requests_issued, config.pipeline, memory_order_relaxed);)
```

zuiderkwast:
> I c. can so maybe we need to take care of the -n accounting?

@ranshid  Yeah, you're right that we should multiply by the sequence length. Updated.

zuiderkwast:
Fixed a test to match the updated request counting logic.

So now `-n` is the number of requests, but `-P` is the number *sequences* per pipeline. Is this intuitive or should I count pipelines in a different way? @ranshid

ranshid:
> Fixed a test to match the updated request counting logic.
> 
> So now `-n` is the number of requests, but `-P` is the number _sequences_ per pipeline. Is this intuitive or should I count pipelines in a different way? @ranshid

IMO the pipeline should remain the number of commands (not sequences) we send and wait for the aggregated results.
This would better match the way clients are working.

zuiderkwast:
> IMO the pipeline should remain the number of commands (not sequences) we send and wait for the aggregated results.
> This would better match the way clients are working.

For some use cases like generating benchmarks like 80/20 GET and SET, yes I agree.

For use cases like benchmarking a specific transaction, I don't agree, but it would be easy for the user to set -P to the same number of commands as in the sequence to achieve this.

More importantly, do you have an idea how to implement it? TBH, I don't have any clue.

ranshid:
> > IMO the pipeline should remain the number of commands (not sequences) we send and wait for the aggregated results.
> > This would better match the way clients are working.
> 
> For some use cases like generating benchmarks like 80/20 GET and SET, yes I agree.
> 
> For use cases like benchmarking a specific transaction, I don't agree, but it would be easy for the user to set -P to the same number of commands as in the sequence to achieve this.
> 
> More importantly, do you have an idea how to implement it? TBH, I don't have any clue.

I think we should be able to provide a way to mimic the same behavior clients do with pipelines. AFAIK they would usually use a transaction-like API to send the commands in a pipeline (so it would usually be -P a multiply of the sequence size), but it would be more flexible to provide the user with the ability to control it. either way, I agree this is not a blocker. we can modify it anytime if we choose to (it will be a breaking change, but not a major breaking change).

ranshid:
@zuiderkwast I spent very small time looking at the details. let me just make a second pass (hope by tomorrow)

zuiderkwast:
> I think we should be able to provide a way to mimic the same behavior clients do with pipelines. AFAIK they would usually use a transaction-like API to send the commands in a pipeline (so it would usually be -P a multiply of the sequence size), but it would be more flexible to provide the user with the ability to control it. either way, I agree this is not a blocker. we can modify it anytime if we choose to (it will be a breaking change, but not a major breaking change).

@ranshid Initially, I imagined two use cases, but I believe only one of them is a good fit for the command sequence feature: to benchmark a specific application scenario such as a transaction (loads a product, stores a session, etc.) so I believe I should change the example to something like `MULTI ; SET key:__rand_int__ __data__ ; INCR counter ; EXEC` and counting the sequence as 1 in terms of pipelining is fine IMHO. Without specifying -P, you get `-P 1` by default and I think one transaction per pipeline is a good default.

For benchmarking a ratio of reads/writes, the command sequence is not the right model. For that, we should instead have some file format where users can list commands and probabilities and the commands are sent randomly, such as this simple file format:

```
99% SADD k value:__rand_int__
1% SPOP k 100
```

For fancier traffic models, there is already an idea to add Lua scripting to valkey-benchmark:

* #1898 

I imagine the Lua scripting can be used for scenarios involving round-trips and some client logic, such as the example with WATCH written in pseudo-code in the docs https://valkey.io/topics/transactions/ 

```
WATCH mykey
val = GET mykey
val = val + 1
MULTI
SET mykey $val
EXEC
```

ranshid:
> > I think we should be able to provide a way to mimic the same behavior clients do with pipelines. AFAIK they would usually use a transaction-like API to send the commands in a pipeline (so it would usually be -P a multiply of the sequence size), but it would be more flexible to provide the user with the ability to control it. either way, I agree this is not a blocker. we can modify it anytime if we choose to (it will be a breaking change, but not a major breaking change).
> 
> @ranshid Initially, I imagined two use cases, but I believe only one of them is a good fit for the command sequence feature: to benchmark a specific application scenario such as a transaction (loads a product, stores a session, etc.) so I believe I should change the example to something like `MULTI ; SET key:__rand_int__ __data__ ; INCR counter ; EXEC` and counting the sequence as 1 in terms of pipelining is fine IMHO. Without specifying -P, you get `-P 1` by default and I think one transaction per pipeline is a good default.
> 
> For benchmarking a ratio of reads/writes, the command sequence is not the right model. For that, we should instead have some file format where users can list commands and probabilities and the commands are sent randomly, such as this simple file format:
> 
> ```
> 99% SADD k value:__rand_int__
> 1% SPOP k 100
> ```
> 
> For fancier traffic models, there is already an idea to add Lua scripting to valkey-benchmark:
> 
> * [Support custom workflow benchmark #1898](https://github.com/valkey-io/valkey/pull/1898)
> 
> I imagine the Lua scripting can be used for scenarios involving round-trips and some client logic, such as the example with WATCH written in pseudo-code in the docs https://valkey.io/topics/transactions/
> 
> ```
> WATCH mykey
> val = GET mykey
> val = val + 1
> MULTI
> SET mykey $val
> EXEC
> ```

@zuiderkwast I agree with what you wrote.

ranshid (approved):
LGTM

zuiderkwast:
Doc PR: https://github.com/valkey-io/valkey-doc/pull/276

PR #2059, opened 2025-05-07
Support RDMA for valkey-cli and benchmark

pizhenwei:
Add '--rdma' for both valkey-cli and valkey-benchmark to enable RDMA.

Valkey has already replaced dependency `hiredis` with `libvalkey` (#2032), and libvalkey also supports RDMA.

### DIFF
--- a/cmake/Modules/ValkeySetup.cmake
+++ b/cmake/Modules/ValkeySetup.cmake
@@ -91,6 +91,10 @@ macro (valkey_build_and_install_bin target sources ld_flags libs link_name)
 
     if (USE_RDMA)
         # Add required libraries needed for RDMA
+        # Bug in libvalkey 0.1.0: The order is important and we need to link
+        # valkey::valkey after valkey::valkey_rdma. This will be fixed upstream.
+        # TODO: Next time we lift libvalkey, remove valkey::valkey from
+        # the next line.
         target_link_libraries(${target} valkey::valkey_rdma valkey::valkey)
     endif ()
 
@@ -104,35 +108,6 @@ macro (valkey_build_and_install_bin target sources ld_flags libs link_name)
     # Install cli tool and create a redis symbolic link
     valkey_install_bin(${target})
     valkey_create_symlink(${target} ${link_name})
-endmacro ()
-
-# Helper function that defines, builds and installs `target` module.
-macro (valkey_build_and_install_module target sources ld_flags libs)
-    add_library(${target} SHARED ${sources})
-
-    if (USE_JEMALLOC)
-        # Using jemalloc
-        target_link_libraries(${target} jemalloc)
-    endif ()
-
-    # Place this line last to ensure that ${ld_flags} is placed last on the linker line
-    target_link_libraries(${target} ${libs} ${ld_flags})
-    if (USE_TLS)
-        # Add required libraries needed for TLS
-        target_link_libraries(${target} OpenSSL::SSL valkey::valkey_tls)
-    endif ()
-
-    if (USE_RDMA)
-        # Add required libraries needed for RDMA
-        target_link_libraries(${target} valkey::valkey_rdma valkey::valkey)
-    endif ()
-
-    if (IS_FREEBSD)
-        target_link_libraries(${target} execinfo)
-    endif ()
-
-    # Install cli tool and create a redis symbolic link
-    valkey_install_bin(${target})
 endmacro ()
 
 # Determine if we are building in Release or Debug mode
--- a/cmake/Modules/ValkeySetup.cmake
+++ b/cmake/Modules/ValkeySetup.cmake
@@ -89,6 +89,11 @@ macro (valkey_build_and_install_bin target sources ld_flags libs link_name)
         target_link_libraries(${target} OpenSSL::SSL valkey::valkey_tls)
     endif ()
 
+    if (USE_RDMA)
+        # Add required libraries needed for RDMA
+        target_link_libraries(${target} valkey::valkey_rdma valkey::valkey)
+    endif ()
+
     if (IS_FREEBSD)
         target_link_libraries(${target} execinfo)
     endif ()
@@ -115,6 +120,11 @@ macro (valkey_build_and_install_module target sources ld_flags libs)
     if (USE_TLS)
         # Add required libraries needed for TLS
         target_link_libraries(${target} OpenSSL::SSL valkey::valkey_tls)
+    endif ()
+
+    if (USE_RDMA)
+        # Add required libraries needed for RDMA
+        target_link_libraries(${target} valkey::valkey_rdma valkey::valkey)
     endif ()
 
     if (IS_FREEBSD)
--- a/deps/CMakeLists.txt
+++ b/deps/CMakeLists.txt
@@ -14,13 +14,13 @@ if (USE_TLS) # Module or no module
     message(STATUS "Building valkey_tls")
     set(ENABLE_TLS
         ON
-        CACHE BOOL "Should we test SSL connections")
+        CACHE BOOL "If TLS support should be compiled or not")
 endif ()
 if (USE_RDMA) # Module or no module
     message(STATUS "Building valkey_rdma")
     set(ENABLE_RDMA
         ON
-        CACHE BOOL "Should we test RDMA connections")
+        CACHE BOOL "If RDMA support should be compiled or not")
 endif ()
 # Let libvalkey use sds and dict provided by valkey.
 set(DICT_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/src)
--- a/deps/CMakeLists.txt
+++ b/deps/CMakeLists.txt
@@ -16,6 +16,12 @@ if (USE_TLS) # Module or no module
         ON
         CACHE BOOL "Should we test SSL connections")
 endif ()
+if (USE_RDMA) # Module or no module
+    message(STATUS "Building valkey_rdma")
+    set(ENABLE_RDMA
+        ON
+        CACHE BOOL "Should we test RDMA connections")
+endif ()
 # Let libvalkey use sds and dict provided by valkey.
 set(DICT_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/src)
 set(SDS_INCLUDE_DIR ${CMAKE_SOURCE_DIR}/src)
@@ -29,3 +35,4 @@ add_subdirectory(hdr_histogram)
 unset(BUILD_SHARED_LIBS CACHE)
 unset(DISABLE_TESTS CACHE)
 unset(ENABLE_TLS CACHE)
+unset(ENABLE_RDMA CACHE)
--- a/src/Makefile
+++ b/src/Makefile
@@ -341,6 +341,7 @@ endif
 ifeq ($(BUILD_RDMA),yes)
 	FINAL_CFLAGS+=-DUSE_RDMA=$(BUILD_YES) -DBUILD_RDMA_MODULE=$(BUILD_NO)
 	FINAL_LIBS += $(RDMA_LIBS)
+	FINAL_LIBS += ../deps/libvalkey/lib/libvalkey_rdma.a $(RDMA_LIBS)
 endif
 
 RDMA_MODULE=
@@ -348,6 +349,7 @@ RDMA_MODULE_NAME:=valkey-rdma$(PROG_SUFFIX).so
 RDMA_MODULE_CFLAGS:=$(FINAL_CFLAGS)
 ifeq ($(BUILD_RDMA),module)
 	FINAL_CFLAGS+=-DUSE_RDMA=$(BUILD_MODULE)
+	RDMA_CLIENT_LIBS = ../deps/libvalkey/lib/libvalkey_rdma.a $(RDMA_LIBS)
 	RDMA_MODULE=$(RDMA_MODULE_NAME)
 	RDMA_MODULE_CFLAGS+=-DUSE_RDMA=$(BUILD_MODULE) -DBUILD_RDMA_MODULE=$(BUILD_MODULE) $(RDMA_LIBS)
 endif
@@ -514,7 +516,7 @@ $(ENGINE_CHECK_AOF_NAME): $(SERVER_NAME)
 
 # valkey-tls.so
 $(TLS_MODULE_NAME): $(SERVER_NAME)
-	$(QUIET_CC)$(CC) -o $@ tls.c -shared -fPIC $(TLS_MODULE_CFLAGS) $(TLS_CLIENT_LIBS)
+	$(QUIET_CC)$(CC) -o $@ tls.c -shared -fPIC $(TLS_MODULE_CFLAGS) $(TLS_CLIENT_LIBS) $(RDMA_CLIENT_LIBS)
 
 # valkey-rdma.so
 $(RDMA_MODULE_NAME): $(SERVER_NAME)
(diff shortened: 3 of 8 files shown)


cmake/Modules/ValkeySetup.cmake
```diff
@@ -89,6 +89,11 @@ macro (valkey_build_and_install_bin target sources ld_flags libs link_name)
         target_link_libraries(${target} OpenSSL::SSL valkey::valkey_tls)
     endif ()
 
+    if (USE_RDMA)
+        # Add required libraries needed for RDMA
+        target_link_libraries(${target} valkey::valkey_rdma valkey::valkey)
```
zuiderkwast, line 98: `valkey::valkey` is already added a few lines above. No need to add it again, right?

```suggestion
        target_link_libraries(${target} valkey::valkey_rdma)
```

Btw, I don't understand why all the binaries use the same libraries? valkey-cli, valkey-benchmark and valkey-server don't necessarily need the same libraries. @eifrah-aws why is it like this? Please help.

cmake/Modules/ValkeySetup.cmake
```diff
@@ -117,6 +122,11 @@ macro (valkey_build_and_install_module target sources ld_flags libs)
...
```
zuiderkwast: Why is `valkey::valkey` (libvalkey) here? Dependency on libvalkey shouldn't depend on whether we build with RDMA or not.


This is the definition of a macro called valkey_build_and_install_module. I was curious what it's used for and it looks like it's not used for anything. Git grep just gives the line where it's defined.

```
$ git grep valkey_build_and_install_module
cmake/Modules/ValkeySetup.cmake:macro (valkey_build_and_install_module target sources ld_flags libs)
```

So let's delete this macro and check if anything breaks.

deps/CMakeLists.txt
```diff
@@ -16,6 +16,12 @@ if (USE_TLS) # Module or no module
         ON
         CACHE BOOL "Should we test SSL connections")
 endif ()
+if (USE_RDMA) # Module or no module
+    message(STATUS "Building valkey_rdma")
+    set(ENABLE_RDMA
+        ON
+        CACHE BOOL "Should we test RDMA connections")
```
zuiderkwast: Is this about testing?

deps/Makefile doesn't build any tests. Why does cmake?

src/Makefile
```diff
@@ -514,19 +516,19 @@ $(ENGINE_CHECK_AOF_NAME): $(SERVER_NAME)
 
 # valkey-tls.so
 $(TLS_MODULE_NAME): $(SERVER_NAME)
-	$(QUIET_CC)$(CC) -o $@ tls.c -shared -fPIC $(TLS_MODULE_CFLAGS) $(TLS_CLIENT_LIBS)
+	$(QUIET_CC)$(CC) -o $@ tls.c -shared -fPIC $(TLS_MODULE_CFLAGS) $(TLS_CLIENT_LIBS) $(RDMA_CLIENT_LIBS)
```
zuiderkwast: Why link RDMA stuff in valkey-tls.so? Is it a mistake?

src/cli_common.c
```diff
@@ -427,9 +430,15 @@ sds cliVersion(void) {
 }
 
 /* This is a wrapper to call valkeyConnect or valkeyConnectWithTimeout. */
-valkeyContext *valkeyConnectWrapper(const char *ip, int port, const struct timeval tv, int nonblock) {
+valkeyContext *valkeyConnectWrapper(const char *ip, int port, const struct timeval tv, int nonblock, int rdma) {
```
zuiderkwast: For Unix domain sockets, there is a special function to connect: valkeyConnectUnixWrapper. It's not nice structure that two kinds of connection use one function and unix sockets use another.

Let's do one of these alternatives:

1. Merge valkeyConnectUnixWrapper with valkeyConnectWrapper and let it take a `type` argument that is TCP, UNIX or RDMA. (We can use ip = path and port = 0 for unix sockets.)
2. Make all three separate functions: valkeyConnectWrapper, valkeyConnectUnixWrapper, valkeyConnectRdmaWrapper. The valkeyConnectRdmaWrapper should not exist if compiled without RDMA.

src/cli_common.c
```diff
@@ -427,9 +430,15 @@ sds cliVersion(void) {
...
     valkeyOptions options = {0};
-    VALKEY_OPTIONS_SET_TCP(&options, ip, port);
+    if (rdma) {
+#ifdef USE_RDMA
+        VALKEY_OPTIONS_SET_RDMA(&options, ip, port);
+#endif
```
zuiderkwast: If `rdma` is passed when this is compiled without USE_RDMA, we shouldn't raise an error instead of ignoring it.

Let's do some assert or just return NULL?

```suggestion
#else
        assert(!rdma);
#endif
```

src/valkey-benchmark.c
```diff
@@ -241,12 +245,12 @@ static dictType dtype = {
...
 };
 
-static valkeyContext *getValkeyContext(const char *ip, int port, const char *hostsocket) {
+static valkeyContext *getValkeyContext(const char *ip, int port, const char *hostsocket, int rdma) {
     valkeyContext *ctx = NULL;
     valkeyReply *reply = NULL;
     struct timeval tv = {0};
     if (hostsocket == NULL)
-        ctx = valkeyConnectWrapper(ip, port, tv, 0);
+        ctx = valkeyConnectWrapper(ip, port, tv, 0, rdma);
```
zuiderkwast: We can just use `config.rdma` here, just like `config.tls` is used some lines below. This function already depends on global config. No need to add the rdma parameter to the getValkeyContext function.

The reason this function takes ip and port is for cluster connections, so they can vary, but rdma and tls are always as in the config.

```suggestion
        ctx = valkeyConnectWrapper(ip, port, tv, 0, config.rdma);
```

src/valkey-benchmark.c
```diff
@@ -1511,8 +1534,15 @@ int parseOptions(int argc, char **argv) {
 #endif
         "";
 
+    rdma_usage =
+#ifdef USE_RDMA
+        " --rdma             Establish a RDMA connection.\n"
```
zuiderkwast, line 1533: Does it work with clusters?

Is guess it could work if RDMA port = TCP port, so CLUSTER SLOTS returns a port that can be used for RDMA for each node. Is this right?

cmake/Modules/ValkeySetup.cmake
pizhenwei: Build this on Ubuntu 24.04 LTS, cmake version 3.28.3, gcc (Ubuntu 13.3.0-6ubuntu2~24.04) 13.3.0:
If removing `valkey::valkey`, hitting a compiling error as following message. We can see `./lib/libvalkey.a ../lib/libvalkey_rdma.a` are linked, but `vk_usec_now` can't be linked. @eifrah-aws do you have any suggestion?

```
make valkey-cli VERBOSE=1

/usr/bin/cc   -DUSE_JEMALLOC -DUSE_RDMA=1 -DBUILD_RDMA_MODULE=0 -std=gnu11 -pedantic "CMakeFiles/valkey-cli.dir/anet.c.o" "CMakeFiles/valkey-cli.dir/adlist.c.o" "CMakeFiles/valkey-cli.dir/dict.c.o" "CMakeFiles/valkey-cli.dir/sds.c.o" "CMakeFiles/valkey-cli.dir/sha256.c.o" "CMakeFiles/valkey-cli.dir/util.c.o" "CMakeFiles/valkey-cli.dir/valkey-cli.c.o" "CMakeFiles/valkey-cli.dir/zmalloc.c.o" "CMakeFiles/valkey-cli.dir/release.c.o" "CMakeFiles/valkey-cli.dir/ae.c.o" "CMakeFiles/valkey-cli.dir/serverassert.c.o" "CMakeFiles/valkey-cli.dir/crcspeed.c.o" "CMakeFiles/valkey-cli.dir/crccombine.c.o" "CMakeFiles/valkey-cli.dir/crc64.c.o" "CMakeFiles/valkey-cli.dir/siphash.c.o" "CMakeFiles/valkey-cli.dir/crc16.c.o" "CMakeFiles/valkey-cli.dir/monotonic.c.o" "CMakeFiles/valkey-cli.dir/cli_common.c.o" "CMakeFiles/valkey-cli.dir/mt19937-64.c.o" "CMakeFiles/valkey-cli.dir/strl.c.o" "CMakeFiles/valkey-cli.dir/cli_commands.c.o" -o ../bin/valkey-cli  ../jemalloc-build/lib/libjemalloc.a ../lib/libfpconv.a ../lib/liblinenoise.a -rdynamic -pthread -ldl -lm ../lib/libvalkey.a ../lib/libvalkey_rdma.a /usr/lib/x86_64-linux-gnu/librdmacm.so /usr/lib/x86_64-linux-gnu/libibverbs.so 
/usr/bin/ld: ../lib/libvalkey_rdma.a(rdma.c.o): in function `vk_msec_now':
rdma.c:(.text+0x377): undefined reference to `vk_usec_now'
collect2: error: ld returned 1 exit status

```

cmake/Modules/ValkeySetup.cmake
pizhenwei: @eifrah-aws 
what about removing `valkey_build_and_install_module`?

deps/CMakeLists.txt
pizhenwei: `ENABLE_RDMA` is used by `libvalkey/CMakeLists.txt`, just copying this option from TLS `"Should we test SSL connections"`, the both two may be updated in the next version.

src/valkey-benchmark.c
pizhenwei: I did some test under cluster mode, it works fine.

src/cli_common.c
pizhenwei: Please take a look at the latest version.

cmake/Modules/ValkeySetup.cmake
bjosv: Since these dependencies are statically linked the order is important. If libvalkey_rdma.a uses code from libvalkey.a then libvalkey.a  need to be last in the link list.
CMake should be able to handle this ordering when calling the linker, but it seems that it doesn't know.

Looking at the CMakeLists.txt in libvalkey we could instead fix it by adding the dependency in deps/libvalkey/CMakeLists.txt:
`target_link_libraries(valkey_rdma LINK_PRIVATE valkey::valkey ${RDMACM_LIBRARIES} ${IBVERBS_LIBRARIES})`
We need to fix it upstream to, and probably change the usage of LINK_PRIVATE to make the transitive dependencies work when the libraries are installed (valkey includes the targets and uses the built libraries directly).

src/Makefile
```diff
@@ -341,13 +341,15 @@ endif
 ifeq ($(BUILD_RDMA),yes)
 	FINAL_CFLAGS+=-DUSE_RDMA=$(BUILD_YES) -DBUILD_RDMA_MODULE=$(BUILD_NO)
 	FINAL_LIBS += $(RDMA_LIBS)
```
bjosv: remove this line? No need to add RDMA_LIBS twice.

cmake/Modules/ValkeySetup.cmake
zuiderkwast: OK, good point. Let's add a code comment about it here.

```suggestion
        # Bug in libvalkey 0.1.0: The order is important and we need to link
        # valkey::valkey after valkey::valkey_rdma. This will be fixed upstream.
        # TODO: Next time we lift libvalkey, remove valkey::valkey from
        # the next line.
        target_link_libraries(${target} valkey::valkey_rdma valkey::valkey)
```

src/cli_common.c
zuiderkwast: Yeah, looks good!

src/valkey-benchmark.c
zuiderkwast: It works if you configure `port == rdma-port`, right? I guess it doesn't work if you configure `port != rdma-port`.

Commands like CLUSTER SLOTS and redirects like MOVED return the port of another node based on whether the connection is using TLS or not. It returns `tls-port` or `port` based on this. These two ports are propagated in the cluster. If we want to fully support RDMA in cluster mode with rdma-port != port, then we should probably also propagate the rdma-port in the cluster. But I'm OK with that we just say that we support RDMA in cluster only when you set RDMA port = TCP port.

zuiderkwast (commented):
Yes, this looks very good now.

I'm not sure if we have test coverage for valkey-benchmark and valkey-cli over unix domain socket. The refactoring of this code looks correct but it's always better to have test. Very small test cases where just check that we can send PING and receive PONG would be good to have.

There are a few open comments, just minor stuff.

I wish we can delete the dead code in the CMake file. If we delete it and nothing breaks, then it is really dead code. I don't want to maintain dead code. :)

cmake/Modules/ValkeySetup.cmake
bjosv: The issue at hand is addressed in https://github.com/valkey-io/libvalkey/pull/201

src/valkey-benchmark.c
pizhenwei: RDMA port must be as same as TCP port in the cluster, this is the limitation under current version. Additional fields may be needed, let's leave it as a TODO topic in the future.

zuiderkwast: Yes, maybe we don't even need to fix it ever. Just a note somewhere in the documentation would be good, or a comment in the valkey.conf.

pizhenwei:
> Yes, this looks very good now.
> 
> I'm not sure if we have test coverage for valkey-benchmark and valkey-cli over unix domain socket. The refactoring of this code looks correct but it's always better to have test. Very small test cases where just check that we can send PING and receive PONG would be good to have.
> 
> There are a few open comments, just minor stuff.
> 
> I wish we can delete the dead code in the CMake file. If we delete it and nothing breaks, then it is really dead code. I don't want to maintain dead code. :)

Please take a look at the latest version, except valkey-cli over unix socket test, TCL is difficult to me ...

zuiderkwast (approved):

zuiderkwast:
For the documentation, we should update https://valkey.io/topics/cli/ and https://valkey.io/topics/benchmark/ which are also used for building man pages.

src/cli_common.c
```diff
@@ -427,25 +430,29 @@ sds cliVersion(void) {
...
+        break;
 
-    return valkeyConnectWithOptions(&options);
-}
+    case VALKEY_CONN_RDMA:
+#ifdef USE_RDMA
+        VALKEY_OPTIONS_SET_RDMA(&options, ip_or_path, port);
+        break;
+#else
+        assert(0); /* requesting RDMA connection without RDMA support??? */
```
ranshid, line 450: nit: can we exit with error? for precompiled valkey-cli user might have problems understanding the root cause.

src/cli_common.c
```diff
@@ -427,25 +430,29 @@ sds cliVersion(void) {
...
+#endif
 
-/* This is a wrapper to call valkeyConnectUnix or valkeyConnectUnixWithTimeout. */
-valkeyContext *valkeyConnectUnixWrapper(const char *path, const struct timeval tv, int nonblock) {
-    valkeyOptions options = {0};
-    VALKEY_OPTIONS_SET_UNIX(&options, path);
+    default:
+        assert(0); /* this should not happen */
```
ranshid, line 454: Same comment - maybe exit with some explanatory error?

src/valkey-benchmark.c
```diff
@@ -1472,6 +1481,14 @@ int parseOptions(int argc, char **argv) {
...
             config.sslconfig.ciphersuites = strdup(argv[++i]);
 #endif
+#endif
+#ifdef USE_RDMA
+        } else if (!strcmp(argv[i], "--rdma")) {
+            if (valkeyInitiateRdma() != VALKEY_OK) {
+                fprintf(stderr, "Failed to initialize RDMA support from libvalkey\n");
+                exit(1);
+            }
+            config.ct = VALKEY_CONN_RDMA;
```
ranshid, line 1491: I would suggest to place the error here that rdma option was provided for executable not supporting rdma

ranshid (commented):
some minor level comments

src/cli_common.c
zuiderkwast: It can't happen. When compiling without RDMA, there is no --rdma option so the user can't get here. The assert is just for us and our CI to detect it and we want a stacktrace if it happens.

src/valkey-benchmark.c
zuiderkwast: Yeah, it could be better. This was modeled after the TLS support. When compilig without TLS, the TLS options simply don't exist.

If we change it, we should change the TLS options too, so they exist and return an error "compiled without TLS".

src/cli_common.c
ranshid: Oh. I missed that :) O.K so I guess the comment can better explain it, but not a blocker.

pizhenwei:
> For the documentation, we should update https://valkey.io/topics/cli/ and https://valkey.io/topics/benchmark/ which are also used for building man pages.

Please take a look at https://github.com/valkey-io/valkey-doc/pull/273

PR #2060, opened 2025-05-07
move statement in if condition for zremrangeGenericCommand function

hwware:
It is similar to PR https://github.com/valkey-io/valkey/pull/2001, this PR goal is to make logic clearer


Nikhil-Manglore (approved):

ranshid (approved):
LGTM. I understand this is just a visual fix.

sarthakaggarwal97 (approved):

enjoy-binbin (commented):
I generally don't like changes like this, it doesn't actually do anything (doesn't make the code easier to understand, doesn't improve anything). I hope we can avoid doing these in the future.

hwware:
> I generally don't like changes like this, it doesn't actually do anything (doesn't make the code easier to understand, doesn't improve anything). I hope we can avoid doing these in the future.

I have a different opinion for this: the changed line should not be outside the if condition, if there is no delete action, we do not need make server.dirty to update. We should not depend on compile optimization . It is about improving the code quality.
Otherwise, we can always write as if(1)  some_action.

PR #2063, opened 2025-05-08
Make sure that the same point reports for geo comparisons

### LINKED ISSUE #2051: [BUG] GEORADIUSBYMEMBER/GEOSEARCH returns inconsistent results between x86 and ARM architectures
**Describe the bug**

When using GEORADIUSBYMEMBER/GEOSEARCH with a zero radius on identical data sets, Valkey 7.2 and above versions returns different results depending on the architecture: the expected member name on x86 machines but `empty array` on ARM (aarch64) machines.

**To reproduce**

1. Set up Valkey server on both x86 and ARM (aarch64) machines.
2. Add the following geo data on both servers:

  ````
  GEOADD towers 139.810716 35.710033 "SkyTree"
  ````

3. Run the following command on both servers:
  ````
  GEORADIUSBYMEMBER towers SkyTree 0 km
  or 
  GEOSEARCH towers FROMMEMBER SkyTree BYRADIUS 0 km
  ````
- x86 machine: Returns "SkyTree" (expected)
- ARM (aarch64) machine: Returns `empty array` (unexpected)

**Expected behavior**

Valkey server returns `SkyTree` on both x86 and ARM machines.

**Additional information**

Any assistance in resolving this cross-architecture inconsistency would be greatly appreciated.

madolson:
Resolves https://github.com/valkey-io/valkey/issues/2051

We have an optimization where if the longitudes of two points are the same in a geospatial search, we can do a much simpler distance across the radius of the earth. This optimization was not being called for ARM with optimizations, since it was always accumulating a small amount of rounding error. This was also causing slightly divergent behavior between x86 and ARM.

The goal of this PR is to compare against a very small value instead of zero to avoid most of these rounding errors. There is a very simple test that worked on x86 but did not work on ARM.

Also deleted an extra constant.

### DIFF
--- a/src/geohash_helper.c
+++ b/src/geohash_helper.c
@@ -46,8 +46,6 @@
 #define ECCENT (sqrt(1.0 - (RATIO * RATIO)))
 #define COM (0.5 * ECCENT)
 
-/// @brief The usual PI/180 constant
-const double DEG_TO_RAD = 0.017453292519943295769236907684886;
 /// @brief Earth's quadratic mean radius for WGS-84
 const double EARTH_RADIUS_IN_METERS = 6372797.560856;
 
@@ -233,8 +231,10 @@ double geohashGetDistance(double lon1d, double lat1d, double lon2d, double lat2d
     lon1r = deg_rad(lon1d);
     lon2r = deg_rad(lon2d);
     v = sin((lon2r - lon1r) / 2);
-    /* if v == 0 we can avoid doing expensive math when lons are practically the same */
-    if (v == 0.0) return geohashGetLatDistance(lat1d, lat2d);
+    /* Reflects about 6nm on earth for comparing longitudes. */
+    const double GEO_EPSILON = 1e-15;
+    /* if v == 0, or practically 0, we can avoid doing expensive math when lons are practically the same */
+    if (fabs(v) <= GEO_EPSILON) return geohashGetLatDistance(lat1d, lat2d);
     lat1r = deg_rad(lat1d);
     lat2r = deg_rad(lat2d);
     u = sin((lat2r - lat1r) / 2);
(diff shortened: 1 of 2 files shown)


enjoy-binbin (approved):

zuiderkwast:
So this extends the radius by 6 nanometers?

src/geohash_helper.c
```diff
@@ -233,8 +231,10 @@ double geohashGetDistance(double lon1d, double lat1d, double lon2d, double lat2d
     lon1r = deg_rad(lon1d);
     lon2r = deg_rad(lon2d);
     v = sin((lon2r - lon1r) / 2);
-    /* if v == 0 we can avoid doing expensive math when lons are practically the same */
-    if (v == 0.0) return geohashGetLatDistance(lat1d, lat2d);
+    /* Reflects about 6nm on earth for comparing longitudes. */
+    const double GEO_EPSILON = 1e-15;
+    /* if v == 0, or practically 0, we can avoid doing expensive math when lons are practically the same */
```
zuiderkwast, line 236: This is not only to avoid expensive math. It is also to compensate for rounding errors. Could be mentioned in this comment.

zuiderkwast (approved):

src/geohash_helper.c
madolson: I originally had a comment, but comparing double to zero is already not a best practice because of rounding errors. I try to avoid comments I think are supposed to be obvious.

madolson:
> So this extends the radius by 6 nanometers?

Not the radius, if a two points are within 6 nanometers (east/west) we assume that we can just measure their (north/south) distance.

I didn't mention in the top comment, we don't have that much precision to begin with. We can only get to ~1m of precision at the equator, and a little less than 1mm at the poles (85 degrees north).

PR #2066, opened 2025-05-08
Minor Refactoring

sarthakaggarwal97:
I was going through the `cluster_legacy` code and found two conditions doing the same thing, so I am just merging those conditions as a refactor.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -1704,16 +1704,13 @@ void clusterNodeCleanupFailureReports(clusterNode *node) {
 
     listNode *ln;
     listIter li;
-    clusterNodeFailReport *fr;
     mstime_t maxtime = server.cluster_node_timeout * CLUSTER_FAIL_REPORT_VALIDITY_MULT;
-    mstime_t now = mstime();
+    const mstime_t now = mstime();
 
     listRewind(l, &li);
     while ((ln = listNext(&li)) != NULL) {
-        fr = ln->value;
-        if (now - fr->time > maxtime) {
-            listDelNode(l, ln);
-        } else if (!clusterNodeIsVotingPrimary(fr->node)) {
+        const clusterNodeFailReport *fr = ln->value;
+        if (now - fr->time > maxtime || !clusterNodeIsVotingPrimary(fr->node)) {
             listDelNode(l, ln);
         }
     }


ranshid (approved):
LGTM

Nikhil-Manglore (approved):

src/cluster_legacy.c
```diff
@@ -1704,16 +1704,13 @@ void clusterNodeCleanupFailureReports(clusterNode *node) {
 
     listNode *ln;
     listIter li;
-    clusterNodeFailReport *fr;
     mstime_t maxtime = server.cluster_node_timeout * CLUSTER_FAIL_REPORT_VALIDITY_MULT;
-    mstime_t now = mstime();
+    const mstime_t now = mstime();
```
hwware, line 1708: Is there any reason we need to set now as const variable?

src/cluster_legacy.c
```diff
@@ -1704,16 +1704,13 @@ void clusterNodeCleanupFailureReports(clusterNode *node) {
...
 
     listRewind(l, &li);
     while ((ln = listNext(&li)) != NULL) {
-        fr = ln->value;
-        if (now - fr->time > maxtime) {
-            listDelNode(l, ln);
-        } else if (!clusterNodeIsVotingPrimary(fr->node)) {
+        const clusterNodeFailReport *fr = ln->value;
```
hwware, line 1712: Is there any reason we need to declear *fr in the while loop and set it as const variable?

src/cluster_legacy.c
sarthakaggarwal97: minor nits suggested by the IDE, no other reason, I can switch it back if this doesn't look right

src/cluster_legacy.c
sarthakaggarwal97: no major reason, just that the variable was only locally used inside while loop.

src/cluster_legacy.c
hwware: @zuiderkwast any suggest?

zuiderkwast: TBH, I don't want to review changes like this. It's a waste of time. :smile: Also it affects the git history unnecessarily.

src/cluster_legacy.c
hwware: Then pls switch back to previous and only combine the 2 conditions together, then I will merge it.

enjoy-binbin (commented):
The old code is quite simple and easy to understand. I am not sure why i didn’t merge the condition at the time, but i generally like the old code (it makes the condition explicit). I generally don't like changes that don't make any real difference. We can merge the condition in the future if we add more condition, this is the right path we should follow i think, we should avoid doing things like these (unnecessary changes).

ranshid:
I have no problem with the decision taken. Just for future reference, IMHO we should not always think that hard about making cosmetic changes (eg #2060). the history can be easily tracked using `git log -L line_start,line_end:file`.

enjoy-binbin:
i dont see the value in some "cosmetic changes", so why do we need it? I know there are definitely some ways can help the track easier but why do we need to bother it? If the changes actually fixes something (typo, doc or anything), or improves something (make the code easy to read/understand), i am cool with that.

I usually see the latest history of code changes in the IDE first. When I am curious about what changes have been made to this piece of code, i will click in. I hate to see it has no value in that commit when i open it. When I find that it has done nothing, i need to select this piece of code additionally, and click "show history for selection"). I just hope we are (or we will) in the same page, every commit count. 

Another thing is to select commits. I don’t want to merge or pick these commits. (On the one hand, it requires you to review the commits again, and on the other hand, it may cause unnecessary conflicts)

zuiderkwast:
Good points Binbin. We should write this down in a guideline document such as CONTRIBUTING.md or another document. Many projects backport commits and need to do extra work for each commit.

sarthakaggarwal97:
Thanks everyone for the feedback.

@enjoy-binbin @zuiderkwast I believe that if this were a new PR, the recommendation / review might have been to merge the if statements, as that is generally considered best practice. I'm not sure about the historical context behind why the conditions were kept separate, but my intention was to improve the code quality. That said, I've closed the PR, as the change isn't significant enough to warrant further discussion.

PR #2070, opened 2025-05-11
[NEW] Introduce lttng based tracing

artikell:
## Introduce

In a production environment, it's quite challenging to figure out why a Valkey is under high load. Right now, tools like INFO or slowlog can offer some clues. But if the Valkey can't respond, we might not get any information at all. 
Usually, we have to rely on tools like `strace` or `perf` to find the root cause. If we set up trace points in advance during the project development, we can quickly pinpoint performance issues.

In this current PR, support has been added for all latency sampling points. Also, information reporting for command execution has been added. At the same time, it supports dynamically turning on or off the information reporting as required. The trace feature is implemented based on LTTng, and this capability is supported in projects like QEMU, Ceph. 

## How to use

Building Valkey with LTTng support:

```
USE_LTTNG=yes make
```

## How to trace

Enable lttng trace events dynamically:
```
~# lttng destroy valkey
~# lttng create valkey
~# lttng enable-event -u 'valkey_server:*'
~# lttng enable-event -u 'valkey_commands:*'
~# lttng track -u -p `pidof valkey-server`
~# lttng start
~# lttng stop
~# lttng view
```

Examples (a client run 'SET', another run 'keys'):
```
...
[15:30:19.334467738] (+0.000001222) xxx valkey_commands:command_call: { cpu_id = 15 }, { name = "set", duration = 0 }
[15:30:19.334469105] (+0.000001367) xxx valkey_commands:command_call: { cpu_id = 15 }, { name = "set", duration = 1 }
[15:30:19.334470327] (+0.000001222) xxx valkey_commands:command_call: { cpu_id = 15 }, { name = "set", duration = 0 }
[15:30:19.369348485] (+0.034878158) xxx valkey_commands:command_call: { cpu_id = 15 }, { name = "keys", duration = 34874 }
[15:30:19.369698322] (+0.000349837) xxx valkey_commands:command_call: { cpu_id = 15 }, { name = "set", duration = 4 }
[15:30:19.369702327] (+0.000004005) xxx valkey_commands:command_call: { cpu_id = 15 }, { name = "set", duration = 2 }
[15:30:19.369704098] (+0.000001771) xxx valkey_commands:command_call: { cpu_id = 15 }, { name = "set", duration = 1 }
[15:30:19.369705884] (+0.000001786) xxx valkey_commands:command_call: { cpu_id = 15 }, { name = "set", duration = 0 }
...
```

Then we can use another script to analyze topN slow commands and other system
level events.

About performance overhead (valkey-benchmark -t get -n 1000000 --threads 4):
1> no lttng builtin: 285632.69 requests per second
2> lttng builtin, no trace: 285551.09 requests per second (almost 0 overhead)
3> lttng builtin, trace commands: 266595.59 requests per second (about ~6.6 overhead)

Generally valkey-server would not run in full utilization, the overhead is acceptable.

## Problem analysis

Add prot and conn field into trace command

Run benchmark tool:
```
GET: rps=227428.0 (overall: 222756.2) avg_msec=0.114 (overall: 0.117)
GET: rps=225248.0 (overall: 223005.2) avg_msec=0.115 (overall: 0.117)
GET: rps=167474.1 (overall: 217942.2) avg_msec=0.193 (overall: 0.122) --> performance drop
GET: rps=220192.0 (overall: 218129.5) avg_msec=0.118 (overall: 0.122)
GET: rps=222868.0 (overall: 218493.7) avg_msec=0.117 (overall: 0.121)

```
Run another 'keys *' command in another connection, lead benchmark performance
drop.

At the same time, lttng traces events:
```
[21:16:30.420997167] (+0.000004064) zhenwei valkey:command_call: { cpu_id = 6 }, { prot = "tcp", conn = "127.0.0.1:6379-127.0.0.1:54668", name = "get", duration = 1 }
[21:16:30.421001262] (+0.000004095) zhenwei valkey:command_call: { cpu_id = 6 }, { prot = "tcp", conn = "127.0.0.1:6379-127.0.0.1:54782", name = "get", duration = 1 }
[21:16:30.485562459] (+0.064561197) zhenwei valkey:command_call: { cpu_id = 6 }, { prot = "tcp", conn = "127.0.0.1:6379-127.0.0.1:54386", name = "keys", duration = 64551 } --> root cause
[21:16:30.485583101] (+0.000020642) zhenwei valkey:command_call: { cpu_id = 6 }, { prot = "tcp", conn = "127.0.0.1:6379-127.0.0.1:54522", name = "get", duration = 1 }
[21:16:30.485763891] (+0.000180790) zhenwei valkey:command_call: { cpu_id = 6 }, { prot = "tcp", conn = "127.0.0.1:6379-127.0.0.1:54542", name = "get", duration = 1 }
[21:16:30.485766451] (+0.000002560) zhenwei valkey:command_call: { cpu_id = 6 }, { prot = "tcp", conn = "127.0.0.1:6379-127.0.0.1:54438", name = "get", duration = 1 }
```

From this change, we can see that connection 127.0.0.1:6379-127.0.0.1:54386
affects other connections.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,19 +104,6 @@ jobs:
       - name: show-kernel-log
         run: sudo dmesg -c
 
-  test-lttng:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
-      - name: prepare-development-libraries
-        run: sudo apt-get install lttng-tools lttng-modules-dkms liblttng-ust-dev
-      - name: make-lttng-builtin
-        run: |
-          make distclean
-          make -j4 USE_LTTNG=yes
-      - name: test
-        run: ./runtest --verbose --tags -slow --dump-logs
-
   build-debian-old:
     runs-on: ubuntu-latest
     container: debian:buster
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,6 +104,19 @@ jobs:
       - name: show-kernel-log
         run: sudo dmesg -c
 
+  test-lttng:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
+      - name: prepare-development-libraries
+        run: sudo apt-get install lttng-tools lttng-modules-dkms liblttng-ust-dev
+      - name: make-lttng-builtin
+        run: |
+          make distclean
+          make -j4 USE_LTTNG=yes
+      - name: test
+        run: ./runtest --verbose --tags -slow --dump-logs
+
   build-debian-old:
     runs-on: ubuntu-latest
     container: debian:buster
--- a/src/config.c
+++ b/src/config.c
@@ -3150,57 +3150,6 @@ static int applyClientMaxMemoryUsage(const char **err) {
     return 1;
 }
 
-static int setTraceEvents(standardConfig *config, sds *argv, int argc, const char **err) {
-    UNUSED(config);
-
-    struct valkeyTraceEvents events = {0};
-    for (int i = 0; i < argc; i++) {
-        if (!strcasecmp(argv[i], "aof")) {
-            events.aof = 1;
-        } else if (!strcasecmp(argv[i], "server")) {
-            events.server = 1;
-        } else if (!strcasecmp(argv[i], "cluster")) {
-            events.cluster = 1;
-        } else if (!strcasecmp(argv[i], "sys")) {
-            events.sys = 1;
-        } else if (!strcasecmp(argv[i], "db")) {
-            events.db = 1;
-        } else if (!strcasecmp(argv[i], "commands")) {
-            events.commands = 1;
-        } else {
-            *err = "trace events should between [server,aof,cluster,sys,db,commands]";
-            goto configerr;
-        }
-    }
-    trace_events = events;
-    sdsclear(server.trace_events);
-    for (int i = 0; i < argc; i++) {
-        server.trace_events = sdscatprintf(server.trace_events, "%s", argv[i]);
-        if (i != argc - 1) {
-            server.trace_events = sdscatlen(server.trace_events, " ", 1);
-        }
-    }
-    return 1;
-configerr:
-    return 0;
-}
-
-static sds getTraceEvents(standardConfig *config) {
-    UNUSED(config);
-    return sdsdup(server.trace_events);
-}
-
-void rewriteTraceEvents(standardConfig *config,
-                        const char *name,
-                        struct rewriteConfigState *state) {
-    UNUSED(config);
-    if (sdslen(server.trace_events) == 0) {
-        rewriteConfigMarkAsProcessed(state, name);
-        return;
-    }
-    rewriteConfigRewriteLine(state, name, sdsdup(server.trace_events), 1);
-}
-
 standardConfig static_configs[] = {
     /* Bool configs */
     createBoolConfig("rdbchecksum", NULL, IMMUTABLE_CONFIG, server.rdb_checksum, 1, NULL, NULL),
@@ -3254,6 +3203,7 @@ standardConfig static_configs[] = {
     createBoolConfig("hide-user-data-from-log", NULL, MODIFIABLE_CONFIG, server.hide_user_data_from_log, 1, NULL, NULL),
     createBoolConfig("import-mode", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.import_mode, 0, NULL, NULL),
     createBoolConfig("auto-failover-on-shutdown", NULL, MODIFIABLE_CONFIG, server.auto_failover_on_shutdown, 0, NULL, NULL),
+    createBoolConfig("trace-enabled", NULL, MODIFIABLE_CONFIG, trace_enabled, 0, NULL, NULL),
 
     /* String Configs */
     createStringConfig("aclfile", NULL, IMMUTABLE_CONFIG, ALLOW_EMPTY_STRING, server.acl_filename, "", NULL, NULL),
@@ -3445,7 +3395,6 @@ standardConfig static_configs[] = {
     createSpecialConfig("rdma-bind", NULL, MODIFIABLE_CONFIG | MULTI_ARG_CONFIG, setConfigRdmaBindOption, getConfigRdmaBindOption, rewriteConfigRdmaBindOption, applyRdmaBind),
     createSpecialConfig("replicaof", "slaveof", IMMUTABLE_CONFIG | MULTI_ARG_CONFIG, setConfigReplicaOfOption, getConfigReplicaOfOption, rewriteConfigReplicaOfOption, NULL),
     createSpecialConfig("latency-tracking-info-percentiles", NULL, MODIFIABLE_CONFIG | MULTI_ARG_CONFIG, setConfigLatencyTrackingInfoPercentilesOutputOption, getConfigLatencyTrackingInfoPercentilesOutputOption, rewriteConfigLatencyTrackingInfoPercentilesOutputOption, NULL),
-    createSpecialConfig("trace-events", NULL, MODIFIABLE_CONFIG | MULTI_ARG_CONFIG, setTraceEvents, getTraceEvents, rewriteTraceEvents, NULL),
 
     /* NULL Terminator, this is dropped when we convert to the runtime array. */
     {NULL},
--- a/src/config.c
+++ b/src/config.c
@@ -3203,7 +3203,6 @@ standardConfig static_configs[] = {
     createBoolConfig("hide-user-data-from-log", NULL, MODIFIABLE_CONFIG, server.hide_user_data_from_log, 1, NULL, NULL),
     createBoolConfig("import-mode", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, server.import_mode, 0, NULL, NULL),
     createBoolConfig("auto-failover-on-shutdown", NULL, MODIFIABLE_CONFIG, server.auto_failover_on_shutdown, 0, NULL, NULL),
-    createBoolConfig("trace-enabled", NULL, MODIFIABLE_CONFIG, trace_enabled, 0, NULL, NULL),
 
     /* String Configs */
     createStringConfig("aclfile", NULL, IMMUTABLE_CONFIG, ALLOW_EMPTY_STRING, server.acl_filename, "", NULL, NULL),
--- a/src/config.c
+++ b/src/config.c
@@ -3150,6 +3150,57 @@ static int applyClientMaxMemoryUsage(const char **err) {
     return 1;
 }
 
+static int setTraceEvents(standardConfig *config, sds *argv, int argc, const char **err) {
+    UNUSED(config);
+
+    struct valkeyTraceEvents events = {0};
+    for (int i = 0; i < argc; i++) {
+        if (!strcasecmp(argv[i], "aof")) {
+            events.aof = 1;
+        } else if (!strcasecmp(argv[i], "server")) {
+            events.server = 1;
+        } else if (!strcasecmp(argv[i], "cluster")) {
+            events.cluster = 1;
+        } else if (!strcasecmp(argv[i], "sys")) {
+            events.sys = 1;
+        } else if (!strcasecmp(argv[i], "db")) {
+            events.db = 1;
+        } else if (!strcasecmp(argv[i], "commands")) {
+            events.commands = 1;
+        } else {
+            *err = "trace events should between [server,aof,cluster,sys,db,commands]";
+            goto configerr;
+        }
+    }
+    trace_events = events;
+    sdsclear(server.trace_events);
+    for (int i = 0; i < argc; i++) {
+        server.trace_events = sdscatprintf(server.trace_events, "%s", argv[i]);
+        if (i != argc - 1) {
+            server.trace_events = sdscatlen(server.trace_events, " ", 1);
+        }
+    }
+    return 1;
+configerr:
+    return 0;
+}
+
+static sds getTraceEvents(standardConfig *config) {
+    UNUSED(config);
+    return sdsdup(server.trace_events);
+}
+
+void rewriteTraceEvents(standardConfig *config,
+                        const char *name,
+                        struct rewriteConfigState *state) {
+    UNUSED(config);
+    if (sdslen(server.trace_events) == 0) {
+        rewriteConfigMarkAsProcessed(state, name);
+        return;
+    }
+    rewriteConfigRewriteLine(state, name, sdsdup(server.trace_events), 1);
+}
+
 standardConfig static_configs[] = {
     /* Bool configs */
     createBoolConfig("rdbchecksum", NULL, IMMUTABLE_CONFIG, server.rdb_checksum, 1, NULL, NULL),
@@ -3394,6 +3445,7 @@ standardConfig static_configs[] = {
     createSpecialConfig("rdma-bind", NULL, MODIFIABLE_CONFIG | MULTI_ARG_CONFIG, setConfigRdmaBindOption, getConfigRdmaBindOption, rewriteConfigRdmaBindOption, applyRdmaBind),
     createSpecialConfig("replicaof", "slaveof", IMMUTABLE_CONFIG | MULTI_ARG_CONFIG, setConfigReplicaOfOption, getConfigReplicaOfOption, rewriteConfigReplicaOfOption, NULL),
     createSpecialConfig("latency-tracking-info-percentiles", NULL, MODIFIABLE_CONFIG | MULTI_ARG_CONFIG, setConfigLatencyTrackingInfoPercentilesOutputOption, getConfigLatencyTrackingInfoPercentilesOutputOption, rewriteConfigLatencyTrackingInfoPercentilesOutputOption, NULL),
+    createSpecialConfig("trace-events", NULL, MODIFIABLE_CONFIG | MULTI_ARG_CONFIG, setTraceEvents, getTraceEvents, rewriteTraceEvents, NULL),
 
     /* NULL Terminator, this is dropped when we convert to the runtime array. */
     {NULL},
--- a/src/connection.h
+++ b/src/connection.h
@@ -64,6 +64,14 @@ typedef enum {
 #define CONN_FLAG_WRITE_BARRIER (1 << 1)        /* Write barrier requested */
 #define CONN_FLAG_ALLOW_ACCEPT_OFFLOAD (1 << 2) /* Connection accept can be offloaded to IO threads. */
 
+typedef enum {
+    CONN_TYPE_ID_NONE = 0,
+    CONN_TYPE_ID_SOCKET,
+    CONN_TYPE_ID_UNIX,
+    CONN_TYPE_ID_TLS,
+    CONN_TYPE_ID_RDMA,
+} ConnectionTypeId;
+
 #define CONN_TYPE_SOCKET "tcp"
 #define CONN_TYPE_UNIX "unix"
 #define CONN_TYPE_TLS "tls"
@@ -74,6 +82,7 @@ typedef void (*ConnectionCallbackFunc)(struct connection *conn);
 
 typedef struct ConnectionType {
     /* connection type */
+    int (*get_type_id)(struct connection *conn);
     const char *(*get_type)(struct connection *conn);
 
     /* connection type initialize & finalize & configure */
@@ -307,6 +316,13 @@ static inline ssize_t connSyncReadLine(connection *conn, char *ptr, ssize_t size
 /* Return CONN_TYPE_* for the specified connection */
 static inline const char *connGetType(connection *conn) {
     return conn->type->get_type(conn);
+}
+
+static inline int connGetTypeId(connection *conn) {
+    if (!conn || conn->type->get_type_id == NULL) {
+        return CONN_TYPE_ID_NONE;
+    }
+    return conn->type->get_type_id(conn);
 }
 
 static inline int connLastErrorRetryable(connection *conn) {
--- a/src/connection.h
+++ b/src/connection.h
@@ -65,7 +65,7 @@ typedef enum {
 #define CONN_FLAG_ALLOW_ACCEPT_OFFLOAD (1 << 2) /* Connection accept can be offloaded to IO threads. */
 
 typedef enum {
-    CONN_TYPE_ID_NONE = 0,
+    CONN_TYPE_ID_INVALID = 0,
     CONN_TYPE_ID_SOCKET,
     CONN_TYPE_ID_UNIX,
     CONN_TYPE_ID_TLS,
@@ -153,8 +153,8 @@ struct connection {
     ConnectionCallbackFunc conn_handler;
     ConnectionCallbackFunc write_handler;
     ConnectionCallbackFunc read_handler;
-    char *saddr;
-    char *daddr;
+    char addr[NET_ADDR_STR_LEN];
+    char laddr[NET_ADDR_STR_LEN];
 };
 
 #define CONFIG_BINDADDR_MAX 16
@@ -283,14 +283,6 @@ static inline void connShutdown(connection *conn) {
 }
 
 static inline void connClose(connection *conn) {
-    if (conn->saddr) {
-        free(conn->saddr);
-        conn->saddr = NULL;
-    }
-    if (conn->daddr) {
-        free(conn->daddr);
-        conn->daddr = NULL;
-    }
     conn->type->close(conn);
 }
 
@@ -320,7 +312,7 @@ static inline const char *connGetType(connection *conn) {
 
 static inline int connGetTypeId(connection *conn) {
     if (!conn || conn->type->get_type_id == NULL) {
-        return CONN_TYPE_ID_NONE;
+        return CONN_TYPE_ID_INVALID;
     }
     return conn->type->get_type_id(conn);
 }
@@ -365,25 +357,10 @@ static inline int connAddrSockName(connection *conn, char *ip, size_t ip_len, in
     return connAddr(conn, ip, ip_len, port, 0);
 }
 
-/* Format the connection address and store the result in daddr/saddr, except Unix socket */
+/* Format the connection address and store the result in addr/laddr, except Unix socket */
 static inline int connFmtName(connection *conn) {
-    char addr[NET_ADDR_STR_LEN];
-    int ret;
-
-    ret = connFormatAddr(conn, addr, sizeof(addr), 0);
-    if (ret) {
-        conn->daddr = strdup(addr);
-    } else {
-        conn->daddr = NULL;
-    }
-
-    ret = connFormatAddr(conn, addr, sizeof(addr), 1);
-    if (ret) {
-        conn->saddr = strdup(addr);
-    } else {
-        conn->saddr = NULL;
-    }
-
+    connFormatAddr(conn, conn->laddr, sizeof(conn->laddr), 0);
+    connFormatAddr(conn, conn->addr, sizeof(conn->addr), 1);
     return 0;
 }
 
--- a/src/connection.h
+++ b/src/connection.h
@@ -153,8 +153,6 @@ struct connection {
     ConnectionCallbackFunc conn_handler;
     ConnectionCallbackFunc write_handler;
     ConnectionCallbackFunc read_handler;
-    char addr[NET_ADDR_STR_LEN];
-    char laddr[NET_ADDR_STR_LEN];
 };
 
 #define CONFIG_BINDADDR_MAX 16
@@ -355,13 +353,6 @@ static inline int connAddrPeerName(connection *conn, char *ip, size_t ip_len, in
 
 static inline int connAddrSockName(connection *conn, char *ip, size_t ip_len, int *port) {
     return connAddr(conn, ip, ip_len, port, 0);
-}
-
-/* Format the connection address and store the result in addr/laddr, except Unix socket */
-static inline int connFmtName(connection *conn) {
-    connFormatAddr(conn, conn->laddr, sizeof(conn->laddr), 0);
-    connFormatAddr(conn, conn->addr, sizeof(conn->addr), 1);
-    return 0;
 }
 
 /* Test a connection is local or loopback.
--- a/src/connection.h
+++ b/src/connection.h
@@ -144,7 +144,8 @@ struct connection {
     ConnectionCallbackFunc conn_handler;
     ConnectionCallbackFunc write_handler;
     ConnectionCallbackFunc read_handler;
-    char *fmtname;
+    char *saddr;
+    char *daddr;
 };
 
 #define CONFIG_BINDADDR_MAX 16
@@ -273,9 +274,13 @@ static inline void connShutdown(connection *conn) {
 }
 
 static inline void connClose(connection *conn) {
-    if (conn->fmtname) {
-        free(conn->fmtname);
-        conn->fmtname = NULL;
+    if (conn->saddr) {
+        free(conn->saddr);
+        conn->saddr = NULL;
+    }
+    if (conn->daddr) {
+        free(conn->daddr);
+        conn->daddr = NULL;
     }
     conn->type->close(conn);
 }
@@ -344,20 +349,25 @@ static inline int connAddrSockName(connection *conn, char *ip, size_t ip_len, in
     return connAddr(conn, ip, ip_len, port, 0);
 }
 
-/* build a generic name for a connection of schema "LADDR:LPORT-RADDR:RPORT", except Unix socket */
+/* Format the connection address and store the result in daddr/saddr, except Unix socket */
 static inline int connFmtName(connection *conn) {
-    char fmtname[NET_ADDR_STR_LEN + NET_ADDR_STR_LEN + 2] = {0};
-    char laddr[NET_ADDR_STR_LEN], raddr[NET_ADDR_STR_LEN];
+    char addr[NET_ADDR_STR_LEN];
     int ret;
 
-    ret = connFormatAddr(conn, laddr, sizeof(laddr), 0);
-    if (ret < 0) return ret;
+    ret = connFormatAddr(conn, addr, sizeof(addr), 0);
+    if (ret) {
+        conn->daddr = strdup(addr);
+    } else {
+        conn->daddr = NULL;
+    }
 
-    ret = connFormatAddr(conn, raddr, sizeof(raddr), 1);
-    if (ret < 0) return ret;
+    ret = connFormatAddr(conn, addr, sizeof(addr), 1);
+    if (ret) {
+        conn->saddr = strdup(addr);
+    } else {
+        conn->saddr = NULL;
+    }
 
-    snprintf(fmtname, sizeof(fmtname), "%s-%s", laddr, raddr);
-    conn->fmtname = strdup(fmtname);
     return 0;
 }
 
--- a/src/connection.h
+++ b/src/connection.h
@@ -34,12 +34,18 @@
 #include <errno.h>
 #include <stdio.h>
 #include <string.h>
+#include <stdlib.h>
 #include <sys/uio.h>
 
 #include "ae.h"
 
 #define CONN_INFO_LEN 32
 #define CONN_ADDR_STR_LEN 128 /* Similar to INET6_ADDRSTRLEN, hoping to handle other protocols. */
+
+#define NET_HOST_STR_LEN 256                          /* Longest valid hostname */
+#define NET_IP_STR_LEN 46                             /* INET6_ADDRSTRLEN is 46, but we need to be sure */
+#define NET_ADDR_STR_LEN (NET_IP_STR_LEN + 32)        /* Must be enough for ip:port */
+#define NET_HOST_PORT_STR_LEN (NET_HOST_STR_LEN + 32) /* Must be enough for hostname:port */
 
 struct aeEventLoop;
 typedef struct connection connection;
@@ -138,6 +144,7 @@ struct connection {
     ConnectionCallbackFunc conn_handler;
     ConnectionCallbackFunc write_handler;
     ConnectionCallbackFunc read_handler;
+    char *fmtname;
 };
 
 #define CONFIG_BINDADDR_MAX 16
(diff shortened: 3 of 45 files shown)


pizhenwei:
Cc lttng-ust maintainer @compudj

Hi @compudj,

We are going to introduce lttng-ust based tracing event into Valkey(one of the most popular KV database) to analyze performance issue. Please feel free to share any suggestion.

compudj:
> Cc lttng-ust maintainer @compudj
> 
> Hi @compudj,
> 
> We are going to introduce lttng-ust based tracing event into Valkey(one of the most popular KV database) to analyze performance issue. Please feel free to share any suggestion.

I've just started looking into your code base. The first thing that pops up is the fact that you have a "--daemonize" command line option which forks the process (daemonize()). Also restartServer() closes file descriptors indiscriminately of whether they are owned by the program or not.

For the daemonize behavior, you will need to run with LD_PRELOAD=liblttng-ust-fork.so.

For the closing of file descriptor behavior, you will need to run with LD_PRELOAD=liblttng-ust-fd.so.

Those are described in the lttng-ust(3) man page in the section "Using LTTng-UST with daemons".

I'll continue looking more deeply into the proposed integration, but those are the initial points that immediately come to mind.

src/trace/README.md
```diff
@@ -0,0 +1,92 @@
...
+
+```
+USE_LTTNG=yes make
+```
+
+Enable lttng trace events dynamically:
+```
+~# lttng destroy valkey
+~# lttng create valkey
+~# lttng enable-event -u "valkey*"
```
compudj: We typically use single quotes to make sure the globbing pattern is not expanded, e.g.

lttng enable-event -u 'valkey*'

src/trace/trace.h
```diff
@@ -0,0 +1,50 @@
...
+        unsigned commands : 1;
+    };
+    unsigned enabled;
+} valkeyTraceEvents;
+
+extern struct valkeyTraceEvents trace_events;
+
+#ifdef USE_LTTNG
+#define latencyTraceIfNeeded(type, event, var) \
+    if (trace_events.type) valkey_##type##_trace(valkey_##type, latency, (event), (var));
```
compudj: Why gate the tracepoint with another level of conditional check ?

Is the overhead of the conditional branch of the tracepoint too high when the instrumentation is disabled ?

compudj: The same effect of enabling a given "group" could be achieved with e.g.:

```
lttng enable-event -u 'valkey_aof:*'
```

or enabling many groups:
```
lttng enable-event -u 'valkey_aof:*'
lttng enable-event -u 'valkey_db:*'
```

pizhenwei: > Why gate the tracepoint with another level of conditional check ?
> 
> Is the overhead of the conditional branch of the tracepoint too high when the instrumentation is disabled ?

Valkey has a typical ~200K QPS over TCP and 450K QPS over RDMA by a single thread, the test result shows that tracing every command hurts ~6% performance in TCP scenario, so we would like to use a config for administrator. Depending on workload or different type of service, the administrator has a choice to config valkey instance tracing event types.
This is the reason we use two gates together for a event:
1, admin config
2, lttng dynamic trace event switch

pizhenwei:
> > Cc lttng-ust maintainer @compudj
> > Hi @compudj,
> > We are going to introduce lttng-ust based tracing event into Valkey(one of the most popular KV database) to analyze performance issue. Please feel free to share any suggestion.
> 
> I've just started looking into your code base. The first thing that pops up is the fact that you have a "--daemonize" command line option which forks the process (daemonize()). Also restartServer() closes file descriptors indiscriminately of whether they are owned by the program or not.
> 
> For the daemonize behavior, you will need to run with LD_PRELOAD=liblttng-ust-fork.so.
> 
> For the closing of file descriptor behavior, you will need to run with LD_PRELOAD=liblttng-ust-fd.so.
> 
> Those are described in the lttng-ust(3) man page in the section "Using LTTng-UST with daemons".
> 
> I'll continue looking more deeply into the proposed integration, but those are the initial points that immediately come to mind.

I guess the valkey process could check the env(liblttng-ust-fork.so&liblttng-ust-fd.so), daemon config during startup, this will protect itself basically. WDYT?

compudj:
> > > Cc lttng-ust maintainer @compudj
> > > Hi @compudj,
> > > We are going to introduce lttng-ust based tracing event into Valkey(one of the most popular KV database) to analyze performance issue. Please feel free to share any suggestion.
> > 
> > 
> > I've just started looking into your code base. The first thing that pops up is the fact that you have a "--daemonize" command line option which forks the process (daemonize()). Also restartServer() closes file descriptors indiscriminately of whether they are owned by the program or not.
> > For the daemonize behavior, you will need to run with LD_PRELOAD=liblttng-ust-fork.so.
> > For the closing of file descriptor behavior, you will need to run with LD_PRELOAD=liblttng-ust-fd.so.
> > Those are described in the lttng-ust(3) man page in the section "Using LTTng-UST with daemons".
> > I'll continue looking more deeply into the proposed integration, but those are the initial points that immediately come to mind.
> 
> I guess the valkey process could check the env(liblttng-ust-fork.so&liblttng-ust-fd.so), daemon config during startup, this will protect itself basically. WDYT?

The more I look at it, the more I think we can do better than LD_PRELOAD for an application which wants to integrate with lttng-ust.

We have the following symbols exposed by include/lttng/ust-fork.h:

void lttng_ust_before_fork(sigset_t *save_sigset);
void lttng_ust_after_fork_parent(sigset_t *restore_sigset);
void lttng_ust_after_fork_child(sigset_t *restore_sigset);

Calling those explicitly around fork() in daemonize() would remove the need to have end users preload the ust-fork helper. Is it something you would be ok with ? See src/lib/lttng-ust-fork/ustfork.c: fork() as an example.

We have something similar which could replace the need to preload liblttng-ust-fd.so: we have a public symbol which is not currently exposed in lttng-ust public headers (but we can publish it if needed). It currently sits in the following internal header: src/common/ust-fd.h

int lttng_ust_safe_close_fd(int fd, int (*close_cb)(int));

So rather than invoking "close(2)" directly from restartServer(), you could call lttng_ust_safe_close_fd(fd, close) instead.

This approach would remove the need for preloading all the lttng-ust helpers. Let me know if this would be OK.

Of course, you could wrap those calls under macros gated by a USE_LTTNG define in your project.

src/trace/trace.h
compudj: It makes sense. For the future, we'll keep a note to investigate how this could integrate with the upcoming "libside" instrumentation library, allowing the application to have an optional  "enable override" on each event, so we could integrate this application event gating concept within the instrumentation API. But for now, with what is available today from the LTTng-UST APIs, your approach with a separate flag check seems fine.

compudj: Just to confirm: in your scenario, are the server administrator and tracing user different roles or the same ? Is your objective to split those permission controls between different roles ?

pizhenwei: Valkey are deployed by both end-users(include companies and organizations) and cloud providers, there are several type of roles for this service ecosystem.

src/trace/trace_aof.h
```diff
@@ -0,0 +1,71 @@
...
+LTTNG_UST_TRACEPOINT_EVENT(
+	/* Tracepoint provider name */
+	valkey_aof,
+
+	/* Tracepoint name */
+	latency,
+
+	/* Input arguments */
+	LTTNG_UST_TP_ARGS(
+		const char *, event,
```
compudj: Wherever you have a const char *event argument with an "event" string field, I recommend to use instead LTTNG_UST_TRACEPOINT_ENUM(). See the section "LTTNG_UST_TRACEPOINT_ENUM() usage" within lttng-ust(3) for an example. A more complete example in present in the lttng-ust source tree under tests/compile/api1/ust-fields/ust_tests_ust_fields.h

The idea is to list all expected enumeration labels, and then only store the integer associated with the label into the trace. If you have small enumerations, then it can fit on a single byte. This is more efficient than copying a string into the ring buffer every time.

src/connection.h
```diff
@@ -333,6 +344,23 @@ static inline int connAddrSockName(connection *conn, char *ip, size_t ip_len, in
...
+    int ret;
+
+    ret = connFormatAddr(conn, laddr, sizeof(laddr), 0);
+    if (ret < 0) return ret;
+
+    ret = connFormatAddr(conn, raddr, sizeof(raddr), 1);
+    if (ret < 0) return ret;
+
+    snprintf(fmtname, sizeof(fmtname), "%s-%s", laddr, raddr);
+    conn->fmtname = strdup(fmtname);
```
compudj: Rather than saving the connection info as a string in the trace, you should describe each field using the lttng-ust tracepoint description facilities to keep the data structured and self-described. It will be easier for tools to interpret the data, and it will typically be more compact in the trace.

src/latency.h
```diff
@@ -69,23 +71,23 @@ void latencyAddSample(const char *event, mstime_t latency);
...
-        var = mstime() - var;               \
+#define latencyEndMonitor(var)                                      \
+    if (server.latency_monitor_threshold || trace_events.enabled) { \
+        var = ustime() - var;                                       \
     }
 
 /* Add the sample only if the elapsed time is >= to the configured threshold. */
 #define latencyAddSampleIfNeeded(event, var) \
-    if (server.latency_monitor_threshold && (var) >= server.latency_monitor_threshold) latencyAddSample((event), (var));
+    if (server.latency_monitor_threshold && ((var) / 1000) >= server.latency_monitor_threshold) latencyAddSample((event), ((var) / 1000));
```
compudj: If this is a fast-path, perhaps consider changing the comparison from a division to a multiplication, e.g.:

if (server.latency_monitor_threshold && (var) >= server.latency_monitor_threshold * 1000)

src/server.c
```diff
@@ -3737,6 +3745,9 @@ void call(client *c, int flags) {
     else
         duration = ustime() - call_timer;
 
+    if (trace_events.commands) {
+        valkey_commands_trace(valkey_commands, command_call, connGetType(c->conn), c->conn->fmtname, real_cmd->declared_name, duration);
```
compudj: every string here that can be turned into structured data or an enumeration would be a win for tooling that consume the traces (analyses are easier to create) and for compactness (and thus tracing performance: you can trace more with the same trace throughput).

src/trace/trace_aof.h
compudj: Actually, now that I understand more about the nature of your "events", I suspect you should just remove the "event" field, and have one tracepoint per event, e.g.:

provider: valkey_sys       event: fork
provider: valkey_server event: command-unblocking
provider: valkey_server event: while-blocked-cron
provider: valkey_server event: module-acquire-GIL
provider: valkey_db        event: expire-del

and so on. When you have multiple events that happen to have the same field types/layout within a provider, you can declare them as a LTTNG_UST_TRACEPOINT_EVENT_CLASS, and have one LTTNG_UST_TRACEPOINT_EVENT_INSTANCE for each of the event. This ensures that all of those event instances re-use mostly the same data and code.

PingXie:
I like lttgn. giving its low overhead, I think we should consider retiring (but not removing just yet) serverLog going forward. we should introduce lttng as a proper logging mechanism that supports level logging. I don't think we need to mimic the existing latency logging paradigm.

pizhenwei:
> I like lttgn. giving its low overhead, I think we should consider retiring (but not removing just yet) serverLog going forward. we should introduce lttng as a proper logging mechanism that supports level logging. I don't think we need to mimic the existing latency logging paradigm.

What about leaving LTTNG based log into next step? This PR will focus on latency related trace events ...

src/config.c
```diff
@@ -3150,6 +3150,57 @@ static int applyClientMaxMemoryUsage(const char **err) {
...
+            events.db = 1;
+        } else if (!strcasecmp(argv[i], "commands")) {
+            events.commands = 1;
+        } else {
+            *err = "trace events should between [server,aof,cluster,sys,db,commands]";
+            goto configerr;
+        }
+    }
+    trace_events = events;
+    sdsclear(server.trace_events);
```
PingXie: I don't see `server.trace_events` get initialized.

src/trace/README.md
```diff
@@ -0,0 +1,92 @@
...
+## LTTng QuickStart
+
+LTTng is an open source tracing framework for Linux that provides highly efficient and low-overhead tracing capabilities. It allows developers to trace both kernel and user-space applications.
+
+Building Valkey with LTTng support:
+
+```
+USE_LTTNG=yes make
+```
+
```
PingXie, line 47: let's mention that the server needs to be run with the `trace-events` config enabled and give a few examples too.

that being said, I do have a question on why an additional server config is needed. I added my question to [this](https://github.com/valkey-io/valkey/pull/2070/files#r2085377207) thread.

src/trace/README.md
```diff
@@ -0,0 +1,92 @@
...
+About performance overhead (valkey-benchmark -t get -n 1000000 --threads 4):
+1> no lttng builtin: 285632.69 requests per second
+2> lttng builtin, no trace: 285551.09 requests per second (almost 0 overhead)
+3> lttng builtin, trace commands: 266595.59 requests per second (about ~6.6 overhead)
+
+Generally valkey-server would not run in full utilization, the overhead is acceptable.
+
+## Supported Events
+
+| event                    | provider |
```
PingXie: missing "commands" provider?

src/trace/trace.c
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * This work is licensed under BSD 3-Clause, License 1 of the COPYING file in
+ * the top-level directory.
+ * ==========================================================================
+ */
+/*
+ * Copyright (c) Valkey Contributors
+ * All rights reserved.
+ * SPDX-License-Identifier: BSD-3-Clause
+ */
```
PingXie: please refer to https://github.com/valkey-io/valkey/blob/unstable/COPYING for consistent look-n-feel of licenses.

src/trace/trace.h
PingXie: > Valkey has a typical ~200K QPS over TCP and 450K QPS over RDMA by a single thread, the test result shows that tracing every command hurts ~6% performance in TCP scenario

@pizhenwei, are you saying even without `lttng enable-event -u "valkey*"` the overhead is already ~6%?

I have the same question on why we need a separate `trace-events` config. Doesn't lttng give us the per-event control already?

src/trace/README.md
```diff
@@ -0,0 +1,92 @@
...
+| cluster-config-close     | cluster  |
+| cluster-config-unlink    | cluster  |
+| aof-write-pending-fsync  | aof      |
+| aof-write-active-child   | aof      |
+| aof-write-alone          | aof      |
+| aof-write                | aof      |
+| aof-fsync-always         | aof      |
+| aof-fstat                | aof      |
+| aof-rename               | aof      |
+| aof-rename               | aof      |
```
PingXie:
```suggestion
| aof-rename               | aof      |
```

src/trace/trace_aof.h
PingXie: +1. I think another benefit is event discoverability because they are now visible via `lttng list -u`

src/trace/trace.h
pizhenwei: > > Valkey has a typical ~200K QPS over TCP and 450K QPS over RDMA by a single thread, the test result shows that tracing every command hurts ~6% performance in TCP scenario
> 
> @pizhenwei, are you saying even without `lttng enable-event -u "valkey*"` the overhead is already ~6%?
> 
> I have the same question on why we need a separate `trace-events` config. Doesn't lttng give us the per-event control already?

There is almost no overhead without `lttng enable-event -u "valkey*"`. When enabling valkey_commands trace point, it will be invoked on each command, this trace point will hurt ~6% performance.
LTTNG supports per-event control, it would be enough for general purpose. This is not strong opinion to use valkey config level control. Removing it is fine to me, too.

src/connection.h
artikell: get, I've separated them.

compudj: Are those strings representing ipv4, ipv6, or either types of addresses (+ port) ?

artikell: It is address, and the format is `[ip]:port`(ipv6) or `ip:port`(ipv4). Now, two fields, `saddr` and `daddr`, are added for description. Together with connType, a socket connection can be described.

compudj: This improved approach (splitting saddr and daddr) still relies on strings.

Then you won't be able to use powerful lttng features such as online filtering at their full extend. For instance, you could use the online filtering to only record events what match a specific sub network into your trace.

One approach you could use to keep your saddr and daddr in a structured binary format rather than convert them to text is as follows:

- Create two events:
  - valkey_commands:command_call_ipv4
  - valkey_commands:command_call_ipv6

For the ipv4 event, have the following fields:

lttng_ust_field_array(uint8_t, saddr, saddr_input, 4)
lttng_ust_field_integer(uint16_t, sport, sport_input)
lttng_ust_field_array(uint8_t, daddr, daddr_input, 4)
lttng_ust_field_integer(uint16_t, dport, dport_input)

For the ipv6 event, have the following fields:

lttng_ust_field_array(uint8_t, saddr, saddr_input, 16)
lttng_ust_field_integer(uint16_t, sport, sport_input)
lttng_ust_field_array(uint8_t, daddr, daddr_input, 16)
lttng_ust_field_integer(uint16_t, dport, dport_input)

Then you can do things like this:

lttng enable-event -u valkey_commands:command_call_ipv4 --filter 'saddr[0] == 127'

which will match all addresses within 127.0.0.0 /8.

If you convert to strings, you lose part of the metadata, and thus part of the features of LTTng and Babeltrace. Moreover, converting to strings adds overhead when tracing is active.

src/trace/trace.h
PingXie: Thanks @pizhenwei. that matches my understanding. let's remove the server config then.

src/trace/trace_aof.h
artikell: It's really nice. I removed the event field in this part.

src/trace/trace.h
artikell: I have retained a switch configuration: `trace_enabled` to ensure that there is also the ability to control it on the server.

src/server.c
artikell: Some information is designed as strings that are dynamically generated. For example, `connType` is not an enumerable piece of information.

compudj: AFAIU, if we look at connGetType, it ends up calling the "get_type()" member, which has the following implementations:

```
static const char *connTLSGetType(connection *conn_) {
    (void)conn_;

    return CONN_TYPE_TLS;
}
```

```
static const char *connSocketGetType(connection *conn) {
    (void)conn;

    return CONN_TYPE_SOCKET;
}
```

```
static const char *connRdmaGetType(connection *conn) {
    UNUSED(conn);

    return CONN_TYPE_RDMA;
}
```


```
static const char *connUnixGetType(connection *conn) {
    UNUSED(conn);

    return CONN_TYPE_UNIX;
}
```

So all that is needed is to introduce an enumeration, and add 4 new type getters that return an enumeration value rather than a string. I don't see what fundamentally requires strings to represent this information in the trace.

Am I missing something ?

artikell: I noticed that the `connType` can be registered, and I'm worried that there may be conflicts if this is done. In the `connTypeRegister`, there is a mechanism to intercept connTypes with the same name, you are right, this modification make sence. 

@pizhenwei What do you think of this part?

pizhenwei: There are some historical reasons here, please take a look at this [commit](https://github.com/valkey-io/valkey/commit/45617385e72eb814c6548064835cf19cdf466a5b).
Because TLS and RDMA connection could be built into the valkey or a module(shared library), the string style is flexible to load connection module dynamically and enable connection listener.

artikell:
> > > > Cc lttng-ust maintainer @compudj
> > > > Hi @compudj,
> > > > We are going to introduce lttng-ust based tracing event into Valkey(one of the most popular KV database) to analyze performance issue. Please feel free to share any suggestion.
> > > 
> > > 
> > > I've just started looking into your code base. The first thing that pops up is the fact that you have a "--daemonize" command line option which forks the process (daemonize()). Also restartServer() closes file descriptors indiscriminately of whether they are owned by the program or not.
> > > For the daemonize behavior, you will need to run with LD_PRELOAD=liblttng-ust-fork.so.
> > > For the closing of file descriptor behavior, you will need to run with LD_PRELOAD=liblttng-ust-fd.so.
> > > Those are described in the lttng-ust(3) man page in the section "Using LTTng-UST with daemons".
> > > I'll continue looking more deeply into the proposed integration, but those are the initial points that immediately come to mind.
> > 
> > 
> > I guess the valkey process could check the env(liblttng-ust-fork.so&liblttng-ust-fd.so), daemon config during startup, this will protect itself basically. WDYT?
> 
> The more I look at it, the more I think we can do better than LD_PRELOAD for an application which wants to integrate with lttng-ust.
> 
> We have the following symbols exposed by include/lttng/ust-fork.h:
> 
> void lttng_ust_before_fork(sigset_t *save_sigset); void lttng_ust_after_fork_parent(sigset_t *restore_sigset); void lttng_ust_after_fork_child(sigset_t *restore_sigset);
> 
> Calling those explicitly around fork() in daemonize() would remove the need to have end users preload the ust-fork helper. Is it something you would be ok with ? See src/lib/lttng-ust-fork/ustfork.c: fork() as an example.
> 
> We have something similar which could replace the need to preload liblttng-ust-fd.so: we have a public symbol which is not currently exposed in lttng-ust public headers (but we can publish it if needed). It currently sits in the following internal header: src/common/ust-fd.h
> 
> int lttng_ust_safe_close_fd(int fd, int (*close_cb)(int));
> 
> So rather than invoking "close(2)" directly from restartServer(), you could call lttng_ust_safe_close_fd(fd, close) instead.
> 
> This approach would remove the need for preloading all the lttng-ust helpers. Let me know if this would be OK.
> 
> Of course, you could wrap those calls under macros gated by a USE_LTTNG define in your project.

Regarding the behavior of the `daemonsize()`, it is an operation that will be used online. Providing the post-fork processing method is beneficial for integrating it into the service. Relying on an external `liblttng-ust-fork.so` file is troublesome for the operation and maintenance.

Regarding the `restartServer` method, it is often used in the debugging and testing processes, and this part will not be used online.

src/server.c
zuiderkwast: Is it only historical reasons? It seems to me that modules are useful for binary distribution.

If we or distros distribute binaries or containers built with statically linked rdma, they will not work on some systems. If we distribute rdma as a module, users can load it or skip it. No?

pizhenwei: > Is it only historical reasons? It seems to me that modules are useful for binary distribution.
> 
> If we or distros distribute binaries or containers built with statically linked rdma, they will not work on some systems. If we distribute rdma as a module, users can load it or skip it. No?

Right.

PingXie: tls/rdma is not a valkey module but a shared lib that is tightly integrated with the core engine. I am not concerned about namespace collision and I think we should switch to enum from strings.

src/trace/trace.c
PingXie: @pizhenwei @artikell any response to my comment above? at the minimum, I think we should keep the "valkey contributors" header at the very top but I would like to hear thoughts from @valkey-io/core-team as well

src/trace/README.md
```diff
@@ -0,0 +1,98 @@
+## Introduction
+
+This directory contains the implementation of tracing using [LTTng](https://lttng.org/) (Linux Trace Toolkit Next Generation).
+
+## LTTng Installation
+
+To install LTTng on your Linux system, follow the instructions provided in the [LTTng documentation](https://lttng.org/download/)
```
PingXie, line 7: maybe include the basic instructions here to save readers another click?

src/trace/trace.h
PingXie: >I have retained a switch configuration: trace_enabled to ensure that there is also the ability to control it on the server.

I think the intent is already expressed via the build time opt-in (USE_LTTNG=yes) and there is the runtime control via lttng itself. I don't think we want to add another server config.

PingXie:
> Regarding the behavior of the `daemonsize()`, it is an operation that will be used online. Providing the post-fork processing method is beneficial for integrating it into the service. Relying on an external `liblttng-ust-fork.so` file is troublesome for the operation and maintenance.

+1.

> Regarding the `restartServer` method, it is often used in the debugging and testing processes, and this part will not be used online.

`restartServer` is used in at least one of the tests. I agree it is exclusively for testing but lttng is also a great tool for debugging so I would like to see us integrate with `lttng_ust_safe_close_fd`.

PingXie:
@artikell @pizhenwei I think this PR is very close and I will approve if the following issues are addressed. We can have a discussion in the core team meeting too if it helps

1. server config `trace-enabled` - my vote is no
2. license header - my vote is to keep the Valkey license section at the top. I am good with keeping your license section but below the Valkey one.
3. incorporate the two recommendations from @compudj 
a. use enum for connection types
b. better support for daemonize/restartServer

This is a great feature. thanks!

src/trace/trace.c
artikell: @PingXie  I'm sorry, I didn't quite understand this part, so I temporarily dealt with other comments. Here, how about I directly refer to the licenses in other file headers and add the valkey contributors? It seems that this would be more reasonable.

compudj:
> @artikell @pizhenwei I think this PR is very close and I will approve if the following issues are addressed. We can have a discussion in the core team meeting too if it helps
> 
>     1. server config `trace-enabled` - my vote is no
> 
>     2. license header - my vote is to keep the Valkey license section at the top. I am good with keeping your license section but below the Valkey one.
> 
>     3. incorporate the two recommendations from @compudj
>        a. use enum for connection types
>        b. better support for daemonize/restartServer
> 
> 
> This is a great feature. thanks!

There is still an open question from me here about the formatting of source/destination addresses and ports: https://github.com/valkey-io/valkey/pull/2070#discussion_r2094506457

I'm fine if you end up deciding to keep formatting addresses to strings to save them into the trace, but this adds overhead, and will limit your ability to use the LTTng online filter.

PingXie:
> There is still an open question from me here about the formatting of source/destination addresses and ports: [#2070 (comment)](https://github.com/valkey-io/valkey/pull/2070#discussion_r2094506457)

+1. I missed that one. Thanks @compudj.

src/connection.h
artikell: During the development process, I found that the Valkey is not just a link based on IP. The link types for UNIX and RDMA cannot be described using IP. Therefore, I think using a string might be more suitable for the case.  
```
[16:34:12.299525464] (+0.014563783) n37-099-024 valkey_commands:command_call: { cpu_id = 7 }, { prot = "unix", saddr = "/unixsocket:0", daddr = "/unixsocket:0", name = "info", duration = 110 }
```
cc @compudj @PingXie

src/trace/trace.c
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * This work is licensed under BSD 3-Clause, License 1 of the COPYING file in
+ * the top-level directory.
+ * ==========================================================================
+ */
+
+#include "trace.h"
+
+#ifdef USE_LTTNG
+pid_t do_fork(void) {
```
artikell, line 21: @compudj I referred to the implementation of `ust-fork`. Could you please help review this logic? At the same time, I found that the method of closing the file descriptor (fd) is not exposed, so the restartServer cannot be modified？

src/connection.h
compudj: If we create events specifically per link type, then we can choose fields and field types specifically per link type.

I described the fields that could be used for ipv4 and for ipv6 earlier in this discussion thread. For UNIX sockets, a pathname string seems appropriate, but AFAIU there is no need to have a "source" and "destination" strings because those will have the same content. I am not too familiar with RDMA, what does the string look like currently, and what is the meaning of its sub-fields ?

src/trace/trace.c
```diff
@@ -1,19 +1,31 @@
...
- * All rights reserved.
- * SPDX-License-Identifier: BSD-3-Clause
- */
 
 #include "trace.h"
 
-int trace_enabled = 0;
+#ifdef USE_LTTNG
+pid_t do_fork(void) {
+    sigset_t sigset;
```
compudj, line 22: in addition, you may want to save errno after fork() returns, and restore it after calling the after_fork functions, just like the lttng-ust fork implementation does.

src/trace/trace.c
compudj: I have replied separately about ust-fork.

For the close fd, I've pushed a commit that promotes the ust-fd.h header to a public header of the LTTng-UST project into the master branch (commit 611ccd9f2f6 ("Move ust-fd.h to public headers")). While we await the next release of lttng-ust, you can simply copy this header file into Valkey and use it. The symbols are already exposed by liblttng-ust-common.so, so you really just need the header file to use them.

You should link against liblttng-ust-common.so as well.

src/trace/README.md
artikell: ok，I have supplemented the basic installation method.

src/connection.h
artikell: In RDMA (Remote Direct Memory Access), a specific node is represented by a 128-bit number named GID (Global Identifier).

src/server.c
artikell: Currently, I've added a type ID, which now serves merely as an enumeration value without any associated constraints. This seems rather odd. 
In the future, we'll need to use this ID as a unique identifier. Additionally, we'll have to maintain the relationship between the ID and the type in both connection.c and trace_commands.c.

src/connection.h
compudj: So for RDMA, those could be represented as an array of uint8_t, printed as hexadecimal (lttng_ust_field_array_hex(uint8_t, gid, gid, 16)).

src/connection.h
```diff
@@ -58,6 +64,14 @@ typedef enum {
 #define CONN_FLAG_WRITE_BARRIER (1 << 1)        /* Write barrier requested */
 #define CONN_FLAG_ALLOW_ACCEPT_OFFLOAD (1 << 2) /* Connection accept can be offloaded to IO threads. */
 
+typedef enum {
+    CONN_TYPE_ID_NONE = 0,
```
PingXie: nit. Invalid seems clearer.

```suggestion
    CONN_TYPE_ID_INVALID = 0,
```

src/latency.h
```diff
@@ -69,23 +75,23 @@ void latencyAddSample(const char *event, mstime_t latency);
...
-        var = mstime() - var;               \
+#define latencyEndMonitor(var)                                      \
+    if (server.latency_monitor_threshold || LATENCY_TRACE_SWITCH) { \
+        var = ustime() - var;                                       \
     }
 
 /* Add the sample only if the elapsed time is >= to the configured threshold. */
 #define latencyAddSampleIfNeeded(event, var) \
-    if (server.latency_monitor_threshold && (var) >= server.latency_monitor_threshold) latencyAddSample((event), (var));
+    if (server.latency_monitor_threshold && (var) >= server.latency_monitor_threshold * 1000) latencyAddSample((event), ((var) / 1000));
```
PingXie: `latencyAddSample` and `latencyAddSampleIfNeeded` using two different time units (`ms` vs `us`) is confusing. can we push the conversion into `latencyAddSample` so both speak `us` instead?

src/trace/README.md
PingXie: thanks @artikell! btw, I was looking for the basic installation steps on mainstream OSes (debian/ubuntu and perhaps mac OS). I think we can skip the more advanced topics like installing lttng on older systems and building-your-own-lttng from source. this helps achieve a balance between a quick startup guide and the risk of the information getting out of date quickly.

src/connection.h
```diff
@@ -138,6 +153,8 @@ struct connection {
     ConnectionCallbackFunc conn_handler;
     ConnectionCallbackFunc write_handler;
     ConnectionCallbackFunc read_handler;
+    char *saddr;
+    char *daddr;
```
PingXie: have you considered embedding these two fields directly in `connection`?  the downside is extra 30 bytes for ipv4 (actually less because there would be no pointers needed and rounding might work in our favor) but then we save two allocations/deallocations and 2 pointers.

src/connection.h
```diff
@@ -333,6 +365,28 @@ static inline int connAddrSockName(connection *conn, char *ip, size_t ip_len, in
...
+
+    ret = connFormatAddr(conn, addr, sizeof(addr), 1);
+    if (ret) {
+        conn->saddr = strdup(addr);
+    } else {
+        conn->saddr = NULL;
+    }
+
+    return 0;
+}
```
PingXie: I think we should populate these two fields as part of the accept handler and remove this one along with `connFormatAddr`. everyone should now use `conn->saddr` and `conn-daddr`.

src/connection.h
```diff
@@ -68,6 +82,7 @@ typedef void (*ConnectionCallbackFunc)(struct connection *conn);
 
 typedef struct ConnectionType {
     /* connection type */
+    int (*get_type_id)(struct connection *conn);
     const char *(*get_type)(struct connection *conn);
```
PingXie, line 86: I don't think `get_type` is needed anymore. we only need `get_type` today for logging so a simple helper function that maps a given type id to its string representation would suffice.

artikell: I have tried it. Considering that there are quite a lot of modifications, I plan to implement it in a separate pull request (PR).

src/connection.h
artikell: It's really nice. I'm inclined to this solution.

src/connection.h
PingXie: thanks @artikell. separate PR is good.

.github/workflows/ci.yml
```diff
@@ -104,6 +104,19 @@ jobs:
       - name: show-kernel-log
         run: sudo dmesg -c
 
+  test-lttng:
```
madolson: I would prefer this in the daily than in the CI, I don't think most PRs will need to address this, so I would prefer it in the daily. You can leave it here for now since it's useful to see everything work in the CI.

src/connection.h
```diff
@@ -333,6 +357,13 @@ static inline int connAddrSockName(connection *conn, char *ip, size_t ip_len, in
     return connAddr(conn, ip, ip_len, port, 0);
 }
 
+/* Format the connection address and store the result in addr/laddr, except Unix socket */
+static inline int connFmtName(connection *conn) {
```
madolson: We should probably be doing error handling on this path.

src/server.h
```diff
@@ -81,6 +81,13 @@ typedef long long ustime_t; /* microsecond time type. */
 #include "rax.h"        /* Radix tree */
 #include "connection.h" /* Connection abstraction */
 #include "memory_prefetch.h"
+#include "trace/trace.h"
+
+#ifdef USE_LTTNG
+#define zfork() do_fork()
+#else
+#define zfork() fork()
```
```suggestion
#define valkey_fork() do_fork()
#else
#define valkey_fork() fork()
```
`z<blah>` was historically reserved for memory stuff. Special implementation details normally get valkey_, valkey_strlcpy.

src/trace/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+Generally valkey-server would not run in full utilization, the overhead is acceptable.
+
+## Supported Events
+
+| event                      | provider          |
+| -------------------------- | ----------------- |
+| command_call               | valkey_commands   |
+| rdb_unlink_temp_file       | valkey_sys        |
+| fork                       | valkey_sys        |
```
madolson:
```suggestion
| rdb_unlink_temp_file       | valkey_bgsave        |
| fork                       | valkey_bgsave        |
```
For cluster and aof, we have it attached to the system that ran the system calls.

src/trace/README.md
```diff
@@ -0,0 +1,185 @@
...
+| command_unblocking         | valkey_server     |
```
madolson:
```suggestion
| command_unblocking         | valkey_commands   |
```
This is an extension of the command call code, it seems to be it should be part of the same provider.

src/networking.c
```diff
@@ -3379,7 +3373,11 @@ void genClientAddrString(client *client, char *addr, size_t addr_len, int remote
         snprintf(addr, addr_len, "%s:0", server.unixsocket);
     } else {
         /* TCP client. */
-        connFormatAddr(client->conn, addr, addr_len, remote);
+        if (remote) {
+            memcpy(addr, client->conn->addr, addr_len);
```
madolson: This memcpy is not really needed anymore, the downstream consumers of this function are really looking for an sds  with the ID in it, so this function can be evolved to return the SDS to avoid that unnecessary call.

madolson: This function also seems to no longer be correctly handling the addr for client kill. Which is why the test is failing. The code looks correct to me though.

madolson (commented):
The high level looks OK to me, just some minor comments about organizing the commands.

src/networking.c
artikell: This is because the `connConnect` method used during master-slave synchronization, which fails to generate the `addr` and `laddr`. I'll first attempt to restore the correctness of the CI and then figure out the appropriate place to generate them.

src/trace/README.md
artikell: done.

src/networking.c
artikell: I'm certain that we don't need to add `laddr` and `addr` to the `connection`. Instead, we can directly reuse `client->peerid` or `client->sockname`. Therefore, I've restored the original logic. cc @PingXie

src/trace/README.md
artikell: Since `command_call` might be invoked very frequently, it would be appropriate to dedicate a separate `provider` for it to prevent accidental operations.

src/connection.h
artikell: i remove it. It's not really necessary.

src/latency.h
artikell: I've standardized the parameters to use `us`. However, since `latencyTimeSeries` uses `ms` and it involves the display of `latency` commands, I didn't make substantial changes.

.github/workflows/ci.yml
artikell: Good idea. I have dealt with it.

src/trace/README.md
artikell: I have made some simplifications. I've retained the installation methods for Debian/Ubuntu. Since macOS is not supported, I won't mention it here.

src/trace/README.md
madolson: Most commands might flow through the unblocking flow as well. I guess I don't feel that strongly about this.

src/connection.h
PingXie: filed #2187 to track this cleanup work

PingXie (approved):
Great work. Thanks @artikell

PR #2074, opened 2025-05-12
Update commandlog parameters conf text to mention special value settings

enjoy-binbin:
Earlier we described the `slowlog-log-slower-than` configuration
option like this, we explicitly mentioned the special meaning of
negative numbers (actually -1) and 0.
```
The following time is expressed in microseconds, so 1000000 is equivalent
to one second. Note that a negative number disables the slow log, while
a value of zero forces the logging of every command.
slowlog-log-slower-than 10000
```

And after #1294 we lost this text, we need to mention these values,
we can mention the special number for all command log's configs,
also it seems we should also mention `slowlog-max-len`.

### DIFF
--- a/valkey.conf
+++ b/valkey.conf
@@ -2073,6 +2073,7 @@ aof-timestamp-enabled no
 # The following time is expressed in microseconds, so 1000000 is equivalent to 1 second.
 # Note that -1 disables the slow log, while 0 forces logging of every command.
 commandlog-execution-slower-than 10000
+# Record the number of commands.
 # There is no limit to this length. Just be aware that it will consume memory.
 # You can reclaim memory used by the slow log with SLOWLOG RESET or COMMANDLOG RESET SLOW.
 commandlog-slow-execution-max-len 128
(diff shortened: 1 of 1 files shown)


valkey.conf
```diff
@@ -2067,8 +2067,11 @@ aof-timestamp-enabled no
 #
 # The threshold is measured in microseconds.
 #
-# Backward Compatibility: The parameter `slowlog-log-slower-than` is still supported but 
-# deprecated in favor of `commandlog-slow-execution`.
+# Backward Compatibility: The parameters `slowlog-log-slower-than` and `slowlog-max-len`
+# are still supported but deprecated in favor of these commandlog parameters.
+#
+# The following time is expressed in microseconds, so 1000000 is equivalent to 1 second.
+# Note that -1 disables the slow log, while 0 forces logging of every command.
```
soloestoy, line 2074: we can mention the special number for all command log's configs

enjoy-binbin: right, updated.

enjoy-binbin:
@zuiderkwast This is quite small, do you want to review it?

zuiderkwast (approved):

PR #2075, opened 2025-05-12
fix: add release_header dependency to CLI and benchmark targets

poiuj:
This commit fixes a build issue that occurs when using the Ninja build system. The valkey-cli and valkey-benchmark targets both include release.c in their source files, which depends on release.h being generated by the mkreleasehdr.sh script. However, only the valkey-server target had a dependency on the release_header target.

By adding the missing dependencies, we ensure that release.h is generated before any source files that include it are compiled, preventing build failures in highly parallel build environments like Ninja.

The issue was not apparent when using Make due to its different dependency resolution approach, but became evident with Ninja's aggressive parallelism.

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -1398,7 +1398,17 @@ static int tryAvoidBulkStrCopyToReply(client *c, robj *obj) {
 
 /* Add an Object as a bulk reply */
 void addReplyBulk(client *c, robj *obj) {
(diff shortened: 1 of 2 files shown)


poiuj:
cc @eifrah-aws

eifrah-aws (approved):

poiuj:
Up. 
Any issues with this PR?

poiuj:
I see there is not interest in this PR – closing.

PR #2082, opened 2025-05-14
Make test cases independent of each other in the moduleapi/scan suite

vitahlin:
I try to running moduleapi test and this command may test failed:
```shell
./runtest-moduleapi --single "unit/moduleapi/scan" --only "Module scan hash dict"
```

I found that the test `Module scan hash dict` and `Module scan hash listpack` use the same key `hh`:
https://github.com/valkey-io/valkey/blob/41a93a2a77c1e5ce5c90d616f0fc14f406bfa87b/tests/unit/moduleapi/scan.tcl#L15-L19

https://github.com/valkey-io/valkey/blob/41a93a2a77c1e5ce5c90d616f0fc14f406bfa87b/tests/unit/moduleapi/scan.tcl#L27-L32

From what I’ve seen in other tests, when the same key is used, it’s typically deleted or reset to avoid side effects between tests.
Maybe we should adjust these tests to make them independent of each other?

### DIFF
--- a/tests/unit/moduleapi/scan.tcl
+++ b/tests/unit/moduleapi/scan.tcl
@@ -39,9 +39,9 @@ start_server {tags {"modules"}} {
 
     test {Module scan zset skiplist} {
         r config set zset-max-ziplist-entries 2
-        r zadd zz 3 f3
-        assert_encoding skiplist zz
-        lsort [r scan.scan_key zz]
+        r zadd zz1 1 f1 2 f2 3 f3
+        assert_encoding skiplist zz1
+        lsort [r scan.scan_key zz1]
     } {{f1 1} {f2 2} {f3 3}}
 
     test {Module scan set intset} {
@@ -52,9 +52,9 @@ start_server {tags {"modules"}} {
 
     test {Module scan set dict} {
         r config set set-max-intset-entries 2
-        r sadd ss 3
-        assert_encoding hashtable ss
-        lsort [r scan.scan_key ss]
+        r sadd ssa 1 2 3
+        assert_encoding hashtable ssa
+        lsort [r scan.scan_key ssa]
     } {{1 {}} {2 {}} {3 {}}}
 
     test {Module scan set listpack} {
--- a/tests/unit/moduleapi/scan.tcl
+++ b/tests/unit/moduleapi/scan.tcl
@@ -52,7 +52,8 @@ start_server {tags {"modules"}} {
 
     test {Module scan set dict} {
         r config set set-max-intset-entries 2
-        r sadd ssa 1 2 3
+        r sadd ssa 1 2 ; # Created as intset
+        r sadd ssa 3   ; # Converted to hashtable
         assert_encoding hashtable ssa
         lsort [r scan.scan_key ssa]
     } {{1 {}} {2 {}} {3 {}}}
--- a/tests/unit/moduleapi/scan.tcl
+++ b/tests/unit/moduleapi/scan.tcl
@@ -45,12 +45,14 @@ start_server {tags {"modules"}} {
     } {{f1 1} {f2 2} {f3 3}}
 
     test {Module scan set intset} {
+        r del ss
         r sadd ss 1 2
         assert_encoding intset ss
         lsort [r scan.scan_key ss]
     } {{1 {}} {2 {}}}
 
     test {Module scan set dict} {
+        r del ssa
         r config set set-max-intset-entries 2
         r sadd ssa 1 2 ; # Created as intset
         r sadd ssa 3   ; # Converted to hashtable
@@ -59,6 +61,7 @@ start_server {tags {"modules"}} {
     } {{1 {}} {2 {}} {3 {}}}
 
     test {Module scan set listpack} {
+        r del ss1
         r sadd ss1 a b c
         assert_encoding listpack ss1
         lsort [r scan.scan_key ss1]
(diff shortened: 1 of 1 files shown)


zuiderkwast (approved):
Yeah, in some test suites, it's possible to run the test cases one by one using `--only` but in some other suites, the test cases depend on each other very much and it's not possible to use `--only`.

With this change, are all test cases in this suite independent now, so `--only` works for each of them?

vitahlin:
> Yeah, in some test suites, it's possible to run the test cases one by one using `--only` but in some other suites, the test cases depend on each other very much and it's not possible to use `--only`.
> 
> With this change, are all test cases in this suite independent now, so `--only` works for each of them?

@zuiderkwast 
I refactored other tests in the `moduleapi/scan.tcl` to make them independent of each other, then I discovered a new issue.

The test `Module scan set dic` may failed after  the modifications.
```shell
    test {Module scan set dict} {
        r config set set-max-intset-entries 2
        r sadd ssa 1 2 3
        assert_encoding hashtable ssa
        lsort [r scan.scan_key ssa]
    } {{1 {}} {2 {}} {3 {}}}
```
> [err]: Module scan set dict in tests/unit/moduleapi/scan.tcl
Expected 'listpack' to match 'hashtable' (context: type source line 100 file /Users/vitah/Downloads/dev/kv/valkey/tests/support/test.tcl cmd {assert_match $enc $val} proc ::assert_encoding level 1) 

--- 

I check the source code:
https://github.com/valkey-io/valkey/blob/c4f9b4635be4c47d6914f33e4df0539aa9b29e46/src/t_set.c#L601-L609

https://github.com/valkey-io/valkey/blob/c4f9b4635be4c47d6914f33e4df0539aa9b29e46/src/t_set.c#L51-L61

Within the current test, `set` is `NULL`, so the code proceeds to evaluate the following condition:
`if (size_hint <= server.set_max_listpack_entries) return createSetListpackObject();` 

Could this be a bug?
Perhaps the logic should be something like:
```c
 if (isSdsRepresentableAsLongLong(value, NULL)) {
        if (size_hint <= server.set_max_intset_entries) {
            return createIntsetObject();
        }
    } else if (size_hint <= server.set_max_listpack_entries) {
        return createSetListpackObject();
    }
```

zuiderkwast:
> Could this be a bug?
> Perhaps the logic should be something like:

I think the logic is right. And your suggested logic looks the same, or am I missing something?

The difference when you refactored the tests is that now we do only one `sadd`. Before it was two. We never convert from intset to listpack. If the first sadd creates an intset, then the second sadd converts it to hashtable, but if it's done in one step, we get a listpack.

```tcl
    test {Module scan set dict} {
        r del ss
        r config set set-max-intset-entries 2
        r sadd ss 1 2 ; # Created as an intset
        r sadd ss 3   ; # Converted to hashtable
        assert_encoding hashtable ss
        lsort [r scan.scan_key ss]
    } {{1 {}} {2 {}} {3 {}}}
```

Btw, using new key names for each test case can also be error prone. We could do a `del` in the beginning of each test (or in the end of each test to clean up).

tests/unit/moduleapi/scan.tcl
```diff
@@ -52,9 +52,9 @@ start_server {tags {"modules"}} {
 
     test {Module scan set dict} {
         r config set set-max-intset-entries 2
-        r sadd ss 3
-        assert_encoding hashtable ss
-        lsort [r scan.scan_key ss]
+        r sadd ssa 1 2 3
```
zuiderkwast:
```suggestion
        r sadd ssa 1 2 ; # Created as intset
        r sadd ssa 3   ; # Converted to hashtable
```

vitahlin:
Since this issue only occurs with numeric types, everything I’m going to discuss next will be in the context of numeric types.


> I think the logic is right. And your suggested logic looks the same, or am I missing something?

When I first run `sadd ss`( the `ss` is a new key), and the **members count max than `set-max-intset-entries`, the encoding of `ss` is `listpack`, not `intset` and not `hashtable`.** This is my steps:
```shell
Last login: Thu May 15 15:36:02 on ttys004
$ ./src/valkey-cli
127.0.0.1:6379> flushall
OK
127.0.0.1:6379> config set set-max-intset-entries 2
OK
127.0.0.1:6379> sadd ss 1 2 3
(integer) 3
127.0.0.1:6379> object encoding ss
"listpack"
127.0.0.1:6379> sadd ss 4
(integer) 1
127.0.0.1:6379> object encoding ss
"listpack"
127.0.0.1:6379> 
```

In the original code, the two if statements in the `setTypeCreate` function were executed sequentially, so it ended up becoming a listpack.

zuiderkwast:
> In the original code, the two if statements in the setTypeCreate function were executed sequentially, so it ended up becoming a listpack.

Exactly. That's what I said earlier. :)

So, do it in two steps still now to make sure it is created as an intset in the first step.

vitahlin:
Got it. I originally thought that adding all the elements in the first step would directly create a hashtable. Thanks for your patience and explanation.

> > In the original code, the two if statements in the setTypeCreate function were executed sequentially, so it ended up becoming a listpack.
> 
> Exactly. That's what I said earlier. :)
> 
> So, do it in two steps still now to make sure it is created as an intset in the first step.

Got it. I originally thought that adding all the elements in the first step would directly create a hashtable. Thanks for your patience and explanation.

zuiderkwast (approved):

PR #2085, opened 2025-05-14
Fix random element in skewed sparse hash table

### LINKED ISSUE #2019: [BUG] valkey-server slows down with `spop <key>` and `srandmember <key>` spinning in hashtableScanDefrag after a few hours
After a few hours of uptime, one of our valkey-server processes (Docker `8.1.1-alpine` and `8.1.1` images) starts consuming 100% CPU usage and only sporadically responds to application requests.

It appears that one of our applications continuously running `SPOP <set> <n>` (while another application is continuously filling this same set with new data) is working just fine until a few hours later where CPU usage suddenly increases from the expected 25% to 100% and Valkey is responding to requests only very slowly and only after 30-60+ seconds.

I captured a couple gdb backtraces that show valkey is spending a lot of time in `hashtableScanDefrag` when this issue happens. All of the gdb backtraces I captured are the exact same except for the function argument pointer and cursor values varying:

```
#0  0x000059218e456671 in hashtableScanDefrag (defragfn=0x0, flags=0, privdata=<synthetic pointer>, fn=<optimized out>, cursor=1972326, ht=0x725cbcb93ab0) at /usr/src/valkey/src/hashtable.c:1722
#1  hashtableScanDefrag (flags=<optimized out>, defragfn=<optimized out>, privdata=<optimized out>, fn=<optimized out>, cursor=<optimized out>, ht=<optimized out>) at /usr/src/valkey/src/hashtable.c:1707
#2  hashtableScan (fn=<optimized out>, privdata=<synthetic pointer>, cursor=1972326, ht=0x725cbcb93ab0) at /usr/src/valkey/src/hashtable.c:1686
#3  hashtableScan (privdata=<synthetic pointer>, fn=<optimized out>, cursor=1972326, ht=0x725cbcb93ab0) at /usr/src/valkey/src/hashtable.c:1685
#4  hashtableSampleEntries (ht=<optimized out>, dst=dst@entry=0x7ffe67b59750, count=count@entry=280) at /usr/src/valkey/src/hashtable.c:2031
#5  0x000059218e4571c2 in hashtableFairRandomEntry (ht=<optimized out>, found=0x7ffe67b5a058) at /usr/src/valkey/src/hashtable.c:2007
#6  0x000059218e4cc7e5 in setTypeRandomElement (setobj=setobj@entry=0x725d06994530, str=str@entry=0x7ffe67b5a0c8, len=len@entry=0x7ffe67b5a0c0, llele=llele@entry=0x7ffe67b5a0b8) at /usr/src/valkey/src/t_set.c:424
#7  0x000059218e4cc97d in setTypePopRandom (set=set@entry=0x725d06994530) at /usr/src/valkey/src/t_set.c:463
#8  0x000059218e4ceb61 in spopWithCountCommand (c=0x726078f57700) at /usr/src/valkey/src/t_set.c:853
#9  0x000059218e4633de in call (c=c@entry=0x726078f57700, flags=flags@entry=3) at /usr/src/valkey/src/server.c:3742
#10 0x000059218e464b4c in processCommand (c=0x726078f57700) at /usr/src/valkey/src/server.c:4366
#11 0x000059218e48bdf1 in processCommandAndResetClient (c=0x726078f57700) at /usr/src/valkey/src/networking.c:3118
#12 processInputBuffer (c=c@entry=0x726078f57700) at /usr/src/valkey/src/networking.c:3246
#13 0x000059218e48d928 in readQueryFromClient (conn=<optimized out>) at /usr/src/valkey/src/networking.c:3351
#14 0x000059218e5ad375 in callHandler (handler=<optimized out>, conn=0x726049b26d80) at /usr/src/valkey/src/connhelpers.h:79
#15 connSocketEventHandler (el=<optimized out>, fd=<optimized out>, clientData=0x726049b26d80, mask=<optimized out>) at /usr/src/valkey/src/socket.c:300
#16 0x000059218e44e449 in aeProcessEvents (flags=27, eventLoop=0x72638b840780) at /usr/src/valkey/src/ae.c:486
#17 aeProcessEvents (flags=27, eventLoop=0x72638b840780) at /usr/src/valkey/src/ae.c:411
#18 aeMain (eventLoop=0x72638b840780) at /usr/src/valkey/src/ae.c:543
#19 0x000059218e442a4a in main (argc=<optimized out>, argv=<optimized out>) at /usr/src/valkey/src/server.c:7219
```

Some more information regarding the specific key this is happening to:
```
127.0.0.1:38542> debug object urls:unretrievable
Value at:0x725d06994530 refcount:1 encoding:hashtable serializedlength:1668169541 lru:1044411 lru_seconds_idle:0 lfu_freq:0 lfu_access_time_minutes:4079
(20.73s)
127.0.0.1:38542> debug htstats-key urls:unretrievable full
Hash table 0 stats (main hash table):
 table size: 14680064
 number of entries: 1850367
 top-level buckets: 2097152
 child buckets: 200170
 max chain length: 4
 avg chain length: 0.10
 chain length distribution:
   0: 1958960 (93.41%)
   1: 81106 (3.87%)
   2: 52290 (2.49%)
   3: 4700 (0.22%)
   4: 96 (0.00%)
Hash table 1 stats (rehashing target):
 table size: 29360128
 number of entries: 14611037
 top-level buckets: 4194304
 child buckets: 699571
 max chain length: 3
 avg chain length: 0.17
 chain length distribution:
   0: 3507259 (83.62%)
   1: 674541 (16.08%)
   2: 12482 (0.30%)
   3: 22 (0.00%)
```

config (note that we've also tried with setting `activedefrag no` and `activerehashing no` **during runtime** with no effect to combat this issue, but we're trying to re-run or tests with starting up valkey using `activedefrag no` now):
```
valkey-server
--activedefrag yes
--aof-timestamp-enabled yes
--appendfsync everysec
--appendonly yes
--bind ::
--bind 0.0.0.0
--client-output-buffer-limit normal 1gb 16mb 15
--client-output-buffer-limit pubsub 64mb 8mb 15
--client-output-buffer-limit replica 16gb 1gb 300
--client-query-buffer-limit 512mb
--databases 1
--dual-channel-replication-enabled yes
--enable-debug-command local
--enable-module-command local
--enable-protected-configs local
--hide-user-data-from-log no
--io-threads 8
--log-format logfmt
--loglevel verbose
--log-timestamp-format iso8601
--maxmemory 128gb
--maxmemory-clients 1%
--maxmemory-policy noeviction
--max-new-connections-per-cycle 100
--no-appendfsync-on-rewrite yes
--oom-score-adj relative
--port ${PORT}
--proto-max-bulk-len 512mb
--repl-backlog-size 64mb
--repl-diskless-load swapdb
--repl-ping-replica-period 1
--repl-timeout 300
--requirepass ${PASS}
--save ''
--set-proc-title yes
--proc-title-template "{title} ${PROJECT} {listen-addr} {server-mode}"
--shutdown-on-sigint nosave
--shutdown-on-sigterm nosave
--shutdown-timeout 300
--supervised no
--tcp-backlog 1024
--tcp-keepalive 30
```

zuiderkwast:
In cases when half of the elements in a hashtable have been deleted in scan order, i.e. using scan and deleting the returned elements, the random selection algorithm becomes very slow. Also, if a random element is deleted repeatedly, this can make the hash table to become skewed and make random selection very slow. Behavior without this fix:

* The fair random algorithm samples some elements by picking a random cursor and then scanning from that point until it finds a number of elements and then picking a random one of them.
* Now, assume there is an empty sequence of buckets (in scan order).
* When we pick a random cursor and it points into this empty sequence, we continue scanning until we find some elements and then pick one of these. Therefore, the buckets just after an empty sequence are more likely to be picked than other buckets.
* If we repeat SPOP many times, this area just after the empty section will be emptied too and the empty section grows, the problem gets worse and worse.

Bucket distribution (x = non-empty bucket, . = empty bucket)

    x..xx.x.xx.x.........x.xx..x.x..x.xx..x.x
                  ^      ^ ^^  ^ ^
                  |      | ||  | |
                random   elements
                cursor    sampled

This PR changes the sampling to pick a new random index to scan for each new bucket chain that has been sampled. This is more similar to how the fair random works in dict.c.

Additionally, this PR includes two small optimizations/tuning:

1. The change allows us to lower the sample sizes for fair random element selection. There are unit tests to verify the fairness of the algorithm.
2. Additionally, the sampling size is decreased even more in very sparse tables to prevent it from sampling empty buckets many times which makes it slow.

One existing unit test is changed from a stack-allocated to a heap-allocated array, because it was observed to use too much stack memory when running locally with the `--large-memory` option.

Fixes #2019.


src/unit/test_hashtable.c
```diff
@@ -879,6 +881,70 @@ int test_random_entry_with_long_chain(int argc, char **argv, int flags) {
...
+    }
+    uint64_t us0 = elapsedUs(timer);
+    printf("Fair random, filled hashtable, avg time: %.3lfµs\n", (double)us0 / num_rounds);
+
+    size_t cursor = random();
+
+    for (int n = 2; n <= 8; n *= 2) {
+        /* Scan and delete until only 1/n of the values remain. */
+        while (hashtableSize(ht) > count / n) {
+            cursor = hashtableScan(ht, cursor, deleteScanFn, ht);
```
ranshid, line 928: maybe we want to use a random cursor here as well?

src/hashtable.c
```diff
@@ -2069,9 +2071,8 @@ unsigned hashtableSampleEntries(hashtable *ht, void **dst, unsigned count) {
     samples.size = count;
     samples.seen = 0;
     samples.entries = dst;
-    size_t cursor = randomSizeT();
     while (samples.seen < count) {
-        cursor = hashtableScan(ht, cursor, sampleEntriesScanFn, &samples);
+        hashtableScan(ht, randomSizeT(), sampleEntriesScanFn, &samples);
```
ranshid, line 2075: Do you think this will have performance impact on random commands?

ranshid (commented):
Overall looks simple to me.
From what I get from [this](https://github.com/valkey-io/valkey/issues/2019#issuecomment-2881576262), this does not fix the issue though right?

Fusl:
> From what I get from [this](https://github.com/valkey-io/valkey/issues/2019#issuecomment-2881576262), this does not fix the issue though right?

I just added an update to that issue. It appears that this patch solves the issue we've been running into.

src/unit/test_hashtable.c
Fusl: Using a cursor here is correct since we want to purposely cause a sparse hash table that deletes the entire continuous range of half of the hash table to test for the behavior of sparse hash tables during spop operations.

src/hashtable.c
zuiderkwast: No, not really. Memory access is the slow operation here. PRNG is fast compared to that.

src/unit/test_hashtable.c
zuiderkwast: Exactly, this test reproduces the problem we get without this fix. The fair random selection gets 500 times slower after deleting half of the elements in scan order.

zuiderkwast:
> Overall looks simple to me. From what I get from [this](https://github.com/valkey-io/valkey/issues/2019#issuecomment-2881576262), this does not fix the issue though right?

In that comment, @Fusl was referring to another patch which was wrong and didn't fix the problem, this one: https://github.com/valkey-io/valkey/issues/2019#issuecomment-2879173275

enjoy-binbin (approved):
LGTM.

src/hashtable.c
soloestoy: IIUC, using a random cursor breaks the SCAN method’s guarantee of uniqueness, risking duplicate key processing. For functions like evict, this could degrade performance or cause logical errors.

soloestoy:
I have not yet carefully reviewed the `hashtableScanDefrag` function, but based on the implementation method of `dictGetSomeKeys`, I have two questions:

1. Does `hashtableScanDefrag` quickly skip the `rehashidx` when encountering an ongoing rehash during the scan?
2. For sparse hash tables, are there additional termination conditions (e.g., exiting the scan if a certain proportion of empty buckets are accessed), even if the target number of keys has not been reached?"

zuiderkwast:
> I have not yet carefully reviewed the `hashtableScanDefrag` function, but based on the implementation method of `dictGetSomeKeys`, I have two questions:
>
> 1. Does `hashtableScanDefrag` quickly skip the `rehashidx` when encountering an ongoing rehash during the scan?

Yes, it skips indexes in the source table (table 0) if they have already been rehashed.

> 2. For sparse hash tables, are there additional termination conditions (e.g., exiting the scan if a certain proportion of empty buckets are accessed), even if the target number of keys has not been reached?"

No, but it does fewer samples when the table is sparse, so it is potentially less fair. This makes it faster but it loops until it finds an entry. This is similar to how dictGetFairRandomKey has a fallback to dictGetRandomKey when dictGetSomeKeys didn't return any elements. dictGetRandomKey is less fair but it loops until it finds an entry.

src/hashtable.c
zuiderkwast: This is the code to get some random keys, hashtableSampleEntries, similar to dictGetSomeKeys. It uses the scan implementation at a random cursor because the scan implementation can already handle rehashing and skip indexes that have already been rehashed.

In SCAN we don't use a random cursor.

Eviction does use kvstoreHashtableSampleEntries so you're right, it can now return duplicate entries. It's the same in dictGetSomeKeys. This PR makes hashtable sampling more similar to dictGetSomeKeys.

By default maxmemory_samples is 5, so it basically only samples one or two buckets.

ranshid (approved):
The change LGTM.

src/hashtable.c
Fusl: > Eviction does use kvstoreHashtableSampleEntries so you're right, it can now return duplicate entries. It's the same in dictGetSomeKeys. This PR makes hashtable sampling more similar to dictGetSomeKeys.

It appears that's a more than acceptable trade-off considering we're basically improving the overall performance (at the very least for sparse hash tables) in here and a high keyspace churn rate (due repeat memory evictions) can evidently also quickly cause the keyspace hash table to become skewed, this is IMHO a net-positive tradeoff we're making.

Running a test with `--maxmemory 100mb`+`--maxmemory-policy allkeys-random` and filling the primary kv hashtable with random string entries:
Before this change, the primary kv hash table chain length can easily extend to over 40. After this change, the chain length stays under 10 for most of the time.

Red line is 8.1.1, blue line is 8.1.1 with the PR patch applied:

image

src/hashtable.c
```diff
@@ -2046,8 +2046,10 @@ int hashtableRandomEntry(hashtable *ht, void **found) {
 /* Points 'found' to a random entry in the hash table and returns 1. Returns 0
  * if the table is empty. This one is more fair than hashtableRandomEntry(). */
 int hashtableFairRandomEntry(hashtable *ht, void **found) {
-    void *samples[FAIR_RANDOM_SAMPLE_SIZE];
-    unsigned count = hashtableSampleEntries(ht, &samples[0], FAIR_RANDOM_SAMPLE_SIZE);
+    /* Sample less if it's very sparse. */
+    size_t num_samples = hashtableSize(ht) >= hashtableBuckets(ht) ? FAIR_RANDOM_SAMPLE_SIZE : WEAK_RANDOM_SAMPLE_SIZE;
```
Fusl, line 2050: This `hashtableSize(ht) >= hashtableBuckets(ht)` expression in my testing barely ever evaluates to false and when it does, we're typically already (or about to be) in the process of rehashing the hashtable to a smaller size. It doesn't hurt to reduce the sampling rate during this time, just figured it was worth a note in case this logic was intended to be different.

Fusl (approved):
Not an authoritative PR approval but this LGTM overall, just one note

src/hashtable.c
soloestoy: I suspect the primary reason for the improvement might be adjusting the `FAIR_RANDOM_SAMPLE_SIZE`. What would be the outcome if we only modified `FAIR_RANDOM_SAMPLE_SIZE` while keeping the previous `hashtableScan` implementation unchanged?

Fusl: I've already tried only adjusting `FAIR_RANDOM_SAMPLE_SIZE` downwards (20, 10, 5, and 1 times of `ENTRIES_PER_BUCKET`) in my very early investigation when I opened the issue and it only made matters significantly worse. I was easily able to reproduce the issue when adjusting that multiplier downwards in my testing (chain length would skyrocket very quickly) but didn't think it was worth mentioning since it wouldn't show the actual issue on an unmodified code base.

I think the only reason we're adjusting the `FAIR_RANDOM_SAMPLE_SIZE` multiplier down in here is that it's not necessary to do that many samples anymore after we're now randomizing the cursor during sampling rather than iterating on the same cursor and we therefore won't as easily run into the same randomness bias issue.

Fusl: In fact, I would even make the argument that with this entire patch applied it may be worth considering completely getting rid of `FAIR_RANDOM_SAMPLE_SIZE` and only using `WEAK_RANDOM_SAMPLE_SIZE` (and just rename it to `RANDOM_SAMPLE_SIZE`) since that also improved spop/srandmember performance by a good amount in my benchmarks without sacrificing too much of the randomness quality.

zuiderkwast: > I suspect the primary reason for the improvement might be adjusting the `FAIR_RANDOM_SAMPLE_SIZE`. What would be the outcome if we only modified `FAIR_RANDOM_SAMPLE_SIZE` while keeping the previous `hashtableScan` implementation unchanged?

@soloestoy Not at all. Only lowering the sample size makes the randomness worse without solving the issue.

I need to explain why this problem arises from the scan-order sampling:

* The fair random algorithm samples some elements by picking a random cursor and then scanning from that point until it finds a number of elements and then picking a random one of them.
* Now, assume there is an empty sequence (in scan order) of buckets.
* When we pick a random cursor and it points into this empty sequence, we continue scanning until we find some elements and then pick one of these. Therefore, the buckets just after an empty sequence are more likely to be picked than other buckets.
* If we repeat SPOP many times, this area just after the empty section will be emptied too and the empty section grows, the problem gets worse and worse.


```
    |x.xx.x.xx.x.........x.xx..x.x..x.xx..x.|
                  ^      ^ ^^  ^ ^
                  |      | ||  | |
                random   elements
                cursor    sampled
```

Making sampling *not* follow any scan order solves the issue. The bad algorithm is my mistake. I wrote it when I implemented the hashtable. The dictGetFairRandomKey does just what this PR does: it picks a random bucket every time. I should have followed the dict's design here but I thought it was smart to sample in scan order. But it wasn't.

We can keep the same sample size, but we have a unit test where we actually do some statistical measurement of the fairness. Without this change, that fairness test would fail with a smaller sample size. This change allows us to lower the sample size.

We could probably lower it more, but I didn't want to risk the fairness of it in cases like ongoing rehashing, very sparse tables, etc. Maybe I'm too conservative, but to lower it more, we should tune this more carefully. The unit test code is useful for tuning it.

src/hashtable.c
zuiderkwast: Yeah, I was conservative when I picked this threshold. I didn't want to sample too few elements to get unfair randomness in normal circumstances. This case will kick in at fill factor 14%. If resize policy is set to AVOID, the table can be very sparse: MIN_FILL_PERCENT_HARD is just 3%.

soloestoy (approved):

src/hashtable.c
soloestoy: it's better to add comments to explain the history.

zuiderkwast: I will improve the PR top comment.

PR #2087, opened 2025-05-14
Fix string embedding decision with keys

poiuj:
This commit fixes the size calculation for embedded strings with keys in createStringObjectWithKeyAndExpire(). The previous implementation used a fixed size of 3 bytes for the key's header, which was incorrect. The new implementation properly calculates the required size based on the key's length and SDS type.

Added a unit test (test_embedded_string_with_key) to verify that:
1. Short values with long keys are properly encoded as embedded strings
2. Values that would cause the total size to exceed 64 bytes are encoded as raw


zuiderkwast (approved):
Thanks, you're right!

So what's the impact of this bug? We sometimes use embedded strings when the total allocation size needed is 65 or 66?

poiuj:
That's right. And when we need 65 or 66 bytes, we actually get 80 bytes as this is the next jemalloc bucket.

src/object.c
```diff
@@ -229,7 +229,10 @@ robj *createStringObjectWithKeyAndExpire(const char *ptr, size_t len, const sds
      * heuristics, e.g. we can look at the jemalloc sizes (16-byte intervals up
      * to 128 bytes). */
     size_t size = sizeof(robj);
-    size += (key != NULL) * (sdslen(key) + 3); /* hdr size (1) + hdr (1) + nullterm (1) */
+    if (key) {
+        size_t key_len = sdslen(key);
+        size += sdsReqSize(key_len, sdsReqType(key_len)) + 1; /* 1 byte for prefixed sds hdr size */
+    }
     size += (expire != -1) * sizeof(long long);
     size += 4 + len; /* embstr header (3) + nullterm (1) */
```
ranshid: Although no problem here, I would love to see this part changed as well:
size += sdsReqSize(len, SDS_TYPE_8)
```suggestion
    size += sdsReqSize(len, SDS_TYPE_8); /* embstr header (3) + nullterm (1) */
```

ranshid (commented):
Thank you Vadym!

The change LGTM. added some minor addition which IMO makes the code more clear.

ranshid (approved):

src/object.c
zuiderkwast: Yes, and in that case, we don't need the comment on the right.

poiuj: Agree. Let met update the PR.

ranshid:
@zuiderkwast I think this should be backported right?

zuiderkwast:
> @zuiderkwast I think this should be backported right?

We can if you want, but I don't think it's very important. We save 8 bytes in very few specific cases. It's not a regression because 8.1 saves memory anyway compared too earlier versions.

Only when key >= 32, value < 7 and key + value < 48 (or 40 if there is an expire), we save 8 bytes here. In this case we get a 64 bytes allocation jemalloc bucket for the robj and 8 bytes for the value.

When the value is >= 7, we need a 16 bytes bucket for it, so the sum is 80 bytes again.

src/unit/test_object.c
```diff
@@ -48,3 +48,31 @@ int test_object_with_key(int argc, char **argv, int flags) {
...
+
+    /* value of length 8 should be raw, because the total size is over 64 */
+    const char *longer_value = "v:123456";
+    robj *longer_val_obj = createStringObject(longer_value, strlen(longer_value));
+
+    robj *embstr_obj = objectSetKeyAndExpire(short_val_obj, key, -1);
+    TEST_ASSERT(embstr_obj->encoding == OBJ_ENCODING_EMBSTR);
+
+    robj *raw_obj = objectSetKeyAndExpire(longer_val_obj, key, -1);
+    TEST_ASSERT(raw_obj->encoding == OBJ_ENCODING_RAW);
```
zuiderkwast, line 72: This test case fails in the daily runs for 32-bit:

* https://github.com/valkey-io/valkey/actions/runs/15125828365/job/42517576263#step:10:349
* https://github.com/valkey-io/valkey/actions/runs/15101473588/job/42442959926#step:10:344

@poiuj Can you look into it?

zuiderkwast: A pointer is 32 bits so the robj struct is smaller.

zuiderkwast:
Bump. This PR broke the daily build. ~~I'm willing to revert this PR if it isn't fixed.~~ Sorry that was rude. If you're not able to fix it, someone else will.

zuiderkwast:
I have a fix here:

* https://github.com/valkey-io/valkey/pull/2124

poiuj:
@zuiderkwast thanks for the fix!
Sorry, didn't get the notification.

PR #2088, opened 2025-05-14
Optimize saving cluster config file using sdscatfmt

sarthakaggarwal97:
Replace `sdscatprintf` with the faster variant `sdscatfmt` in the code for generating the cluster config file. This optimization also affects CLUSTER NODES which uses the same code.

When the cluster topology changes, such as during failovers, the cluster config file is written and at the same time the nodes may experience extra load due to failovers, redirects, full sync, etc.

In the profiles I am capturing for large scale clusters, I am observing decent compute cycles being utilized by `sdscatprintf`

Profile:
Screenshot 2025-05-14 at 2 57 51 PM

This profile was captured when the engine CPU utilization was more than 90% during multiple primary node failures.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -344,7 +344,7 @@ int auxHumanNodenameSetter(clusterNode *n, void *value, size_t length) {
 }
 
 sds auxHumanNodenameGetter(clusterNode *n, sds s) {
-    return sdscatfmt(s, "%s", n->human_nodename);
+    return sdscat(s, n->human_nodename);
 }
 
 int auxHumanNodenamePresent(clusterNode *n) {
@@ -370,7 +370,7 @@ int auxAnnounceClientIpV4Setter(clusterNode *n, void *value, size_t length) {
 }
 
 sds auxAnnounceClientIpV4Getter(clusterNode *n, sds s) {
-    return sdscatfmt(s, "%s", n->announce_client_ipv4);
+    return sdscat(s, n->announce_client_ipv4);
 }
 
 int auxAnnounceClientIpV4Present(clusterNode *n) {
@@ -396,7 +396,7 @@ int auxAnnounceClientIpV6Setter(clusterNode *n, void *value, size_t length) {
 }
 
 sds auxAnnounceClientIpV6Getter(clusterNode *n, sds s) {
-    return sdscatfmt(s, "%s", n->announce_client_ipv6);
+    return sdscat(s, n->announce_client_ipv6);
 }
 
 int auxAnnounceClientIpV6Present(clusterNode *n) {
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -327,7 +327,7 @@ int auxShardIdSetter(clusterNode *n, void *value, size_t length) {
 }
 
 sds auxShardIdGetter(clusterNode *n, sds s) {
-    return sdscatprintf(s, "%.40s", n->shard_id);
+    return sdscatlen(s, n->shard_id, CLUSTER_NAMELEN);
 }
 
 int auxShardIdPresent(clusterNode *n) {
@@ -344,7 +344,7 @@ int auxHumanNodenameSetter(clusterNode *n, void *value, size_t length) {
 }
 
 sds auxHumanNodenameGetter(clusterNode *n, sds s) {
-    return sdscatprintf(s, "%s", n->human_nodename);
+    return sdscatfmt(s, "%s", n->human_nodename);
 }
 
 int auxHumanNodenamePresent(clusterNode *n) {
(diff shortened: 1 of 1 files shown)


zuiderkwast:
The PR description could mention that this is about cluster nodes.

Is cluster nodes a hot path? Is it frequently called during failures? I thought most clients use cluster slots...

sarthakaggarwal97:
In the scenario where replica promotes itself to become the primary, it tries to save the latest cluster config. 

In the function `markNodeAsFailingIfNeeded`, if the weak agreement is reached between the primaries about a node failure, it tries to trigger an cluster configuration save by setting the [flag](https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L2207).

This is where it is saved I think: https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L5569

I don't think its related to cluster nodes API I guess. Please correct me if I am wrong.

zuiderkwast:
Ok, thanks for the explanation. It looked like the cluster nodes command in the diff.

Please update the PR comment to mention it is the save cluster config file.

Looking more carefully at the flame graph, I can see that it is clusterSaveConfig. But the same code is used for CLUSTER NODES. The output of that command is very similar to the config file.

zuiderkwast (approved):
LGTM, thanks.

I edited the PR title and top-comment.

IMO explaining what scenario is optimized is useful for understand the benefit. We use the title and link to the PR from the release notes, etc. Not sure if this is significant enough for release notes though, but in general, I think we should have user perspective in the titles and descriptions.

src/cluster_legacy.c
```diff
@@ -6045,7 +6045,7 @@ sds clusterGenNodeDescription(client *c, clusterNode *node, int tls_primary) {
                 continue;
             }
             if (auxFieldHandlers[i].isPresent(node)) {
-                ci = sdscatprintf(ci, ",%s=", auxFieldHandlers[i].field);
+                ci = sdscatfmt(ci, ",%s=", auxFieldHandlers[i].field);
                 ci = auxFieldHandlers[i].getter(node, ci);
```
zuiderkwast, line 6049: Looking at this code a bit more, it seems some of these aux field getters can also be optimized in the same way:

```c
sds auxShardIdGetter(clusterNode *n, sds s) {
    return sdscatprintf(s, "%.40s", n->shard_id);
}
```

```c
sds auxHumanNodenameGetter(clusterNode *n, sds s) {
    return sdscatprintf(s, "%s", n->human_nodename);
}
```

```c
sds auxAnnounceClientIpV4Getter(clusterNode *n, sds s) {
    return sdscatprintf(s, "%s", n->announce_client_ipv4);
}
```

```c
sds auxAnnounceClientIpV6Getter(clusterNode *n, sds s) {
    return sdscatprintf(s, "%s", n->announce_client_ipv6);
}
```

```c
sds auxTcpPortGetter(clusterNode *n, sds s) {
    return sdscatprintf(s, "%d", n->tcp_port);
}
```

```c
sds auxTlsPortGetter(clusterNode *n, sds s) {
    return sdscatprintf(s, "%d", n->tls_port);
}
```

src/cluster_legacy.c
```diff
@@ -6095,11 +6095,17 @@ sds clusterGenNodeDescription(client *c, clusterNode *node, int tls_primary) {
...
      * instances. */
     if (node->flags & CLUSTER_NODE_MYSELF) {
+        char nodename[CLUSTER_NAMELEN + 1];
+
         for (j = 0; j < CLUSTER_SLOTS; j++) {
             if (server.cluster->migrating_slots_to[j]) {
-                ci = sdscatprintf(ci, " [%d->-%.40s]", j, server.cluster->migrating_slots_to[j]->name);
+                memcpy(nodename, server.cluster->migrating_slots_to[j]->name, CLUSTER_NAMELEN);
+                nodename[CLUSTER_NAMELEN] = '\0';
+                ci = sdscatfmt(ci, " [%i->-%s]", j, nodename);
```
zuiderkwast: For these node names, I'm thinking we can maybe do something smarter to avoid copying the node name twice.

Disclaimer: untested :smile:

```suggestion
                ci = sdscatfmt(ci, " [%i->-", j);
                ci = sdscatlen(ci, server.cluster->migrating_slots_to[j]->name, CLUSTER_NAMELEN);
                ci = sdscat(ci, "]");
```

sarthakaggarwal97:
nice thanks @zuiderkwast! let me make those changes and test! thank you

sarthakaggarwal97:
@zuiderkwast addressed the comments, thank you!

src/cluster_legacy.c
```diff
@@ -344,7 +344,7 @@ int auxHumanNodenameSetter(clusterNode *n, void *value, size_t length) {
 }
 
 sds auxHumanNodenameGetter(clusterNode *n, sds s) {
-    return sdscatprintf(s, "%s", n->human_nodename);
+    return sdscatfmt(s, "%s", n->human_nodename);
```
zuiderkwast: Do you think simply `sdscat` is faster? I guess we could use it whenever the format is simply "%s".

sarthakaggarwal97: yeah, that's a good call out. I think sdscat should be faster based on the implementations. I can use this for the aux getters! addressing these.

zuiderkwast:
@xbasel It's waiting for your review. If you don't plan to review it, I think it's ready to merge.

xbasel:
> @xbasel It's waiting for your review. If you don't plan to review it, I think it's ready to merge.

Lgtm

PR #2091, opened 2025-05-15
Fix minor memory leak in valkey-cli when command table hint fails due to NOAUTH

### LINKED ISSUE #2064: [BUG] Memory leak in valkey-cli
**Describe the bug**

When the server requires authentication, valkey-cli leaks memory when connecting in interactive mode.

**To reproduce**

1. Compile with address santitizer.
2. Start valkey with authentication required, either ACL rules or the old `requirepass` option.
3. Start valkey-cli and exit again by pressing Ctrl+D.

```sh
cd src
make distclean
make -j10 SANITIZER=address
./valkey-server --requirepass 12345 &
./valkey-cli
```

**Expected behavior**

No leak.

**Additional information**

Valkey-cli sends some commands when it connects, e.g. COMMAND, COMMAND DOCS and INFO. These return a NOAUTH error when the client is unathenticated. I guess this code path leaks some memory.

We should add a test case that starts valkey with requirepass and runs valkey-cli in interactive mode (tests/integration/valkey-cli.tcl). Then our CI jobs would catch a memory leak here.

cxljs:
Valkey-cli sends some commands when it connects, e.g. COMMAND and
COMMAND DOCS to init the command table hint. These return a NOAUTH
error when the client is unathenticated and there is a memory leak.

Close #2064.

### DIFF
--- a/tests/integration/valkey-cli.tcl
+++ b/tests/integration/valkey-cli.tcl
@@ -879,4 +879,20 @@ if {!$::tls} { ;# fake_redis_node doesn't support TLS
             assert_equal {PONG} [exec {*}$cmdline PING]
         }
     }
+
+    test "check whether valkey-cli memory leak when the server require auth" {
+        # enable auth of server
+        set fd [open_cli]
+        assert_equal "OK" [run_command $fd "CONFIG SET requirepass 12345"]
+
+        # Using another client to send a command (non-AUTH command), then closing the client
+        # Valgrind should not detect memory leaks
+        set fd_2 [open_cli]
+        run_command $fd_2 "ping"
+        close_cli $fd_2
+
+        # disable auth of server
+        assert_equal "OK" [run_command $fd "CONFIG SET requirepass \"\""]
+        close_cli $fd
+    }
 }
--- a/tests/integration/valkey-cli.tcl
+++ b/tests/integration/valkey-cli.tcl
@@ -880,18 +880,17 @@ if {!$::tls} { ;# fake_redis_node doesn't support TLS
         }
     }
 
-    test "check whether valkey-cli memory leak when the server require auth" {
-        # enable auth of server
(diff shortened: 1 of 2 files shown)


hwware (commented):
Do you have a chance to add a test case to make sure we fix this memory leak? Here are some examples for valkey-cli \integration\valkey-cli.tcl

cxljs:
> Do you have a chance to add a test case to make sure we fix this memory leak? Here are some examples for valkey-cli \integration\valkey-cli.tcl

I've added the tests, but it seems GitHub CI is skipping the test-valgrind-* checks.

tests/integration/valkey-cli.tcl
```diff
@@ -879,4 +879,20 @@ if {!$::tls} { ;# fake_redis_node doesn't support TLS
...
+        # Using another client to send a command (non-AUTH command), then closing the client
+        # Valgrind should not detect memory leaks
+        set fd_2 [open_cli]
+        run_command $fd_2 "ping"
+        close_cli $fd_2
+
+        # disable auth of server
+        assert_equal "OK" [run_command $fd "CONFIG SET requirepass \"\""]
+        close_cli $fd
+    }
```
enjoy-binbin:
```suggestion
    test "valkey-cli command table hint will not leak memory when COMMAND fails due to auth" {
        # Enable auth of server.
        set fd [open_cli]
        assert_equal "OK" [run_command $fd "CONFIG SET requirepass 12345"]

        # Using another client to send a command (non-AUTH command), then closing the client.
        set fd2 [open_cli]
        assert_match "NOAUTH*" [run_command $fd2 "PING"]
        close_cli $fd2

        # Disable auth of server.
        assert_equal "OK" [run_command $fd "CONFIG SET requirepass \"\""]
        close_cli $fd
    }
```

enjoy-binbin (approved):

Issue #2077, opened 2025-05-13
Avoid recalculating cluster slot when the I/O thread has already done so


zuiderkwast:
Currently, the the IO thread just computes the cluster slot based on the first key, to be able to prefetch the keys.

When getNodeByQuery is executed (detects cross-slot commands, etc.) then it needs to check that all keys hash to the same slot. To make it possible to skip this step, the IO thread would need to check all keys and only set c->slot if all keys hash to the same slot. Then, getNodeByQuery can trust c->slot if it's already set and skip the expensive hashslot calculations.

zuiderkwast:
Related:

* #632

dvkashapov:
I guess I'm a little late to the train but still here are benchmarking results for #2165 vs 8.1:

- config (similar for 2nd shard):
```
cluster-enabled yes
cluster-config-file nodes-1.conf
cluster-node-timeout 5000
cluster-replica-validity-factor 0

databases 1
save ""
appendonly no
rdbcompression no
auto-aof-rewrite-percentage 0
auto-aof-rewrite-min-size 0
activedefrag no
maxclients 800
io-threads 9
protected-mode no
maxmemory 1gb
maxmemory-policy noeviction
hz 1
latency-monitor-threshold 0

port 7001
bind 0.0.0.0
tcp-backlog 65536
tcp-keepalive 300

no-appendfsync-on-rewrite yes
oom-score-adj no
disable-thp yes
```

- Cache population: `valkey-benchmark -h <ip> -p 7001 -c 96 --threads 96 -d 256 -t set -n 3000000 -r 3000000 --sequential --cluster`
- Benchmark itself: `valkey-benchmark -h <ip> -p 7001 -c 768 --threads 32 -d 256 -t get -n 30000000 -r 3000000 -P 7 --cluster`

Results #2165 
```
Summary:
  throughput summary: 1327610.00 requests per second
  latency summary (msec):
          avg       min       p50       p95       p99       max
        3.935     0.064     3.999     4.983     7.487    43.519

```

Results 8.1
```
Summary:
  throughput summary: 1108033.38 requests per second
  latency summary (msec):
          avg       min       p50       p95       p99       max
        4.768     0.096     4.839     5.191     5.575    35.071

```
Numbers are stable and compared to 8.1 on average difference in throughput is +16% and in latency -17%. Seems like a solid improvement!

Issue #632, opened 2024-06-12
Investigate performance improvements related to slot ownership in getNodeByQuery


madolson:
While playing around with profiling various workloads, I noticed that getNodeByQuery was quite slow, taking about 4% of the CPU of a given command. After deep dive, I realized that we basically are checking three 171kb (16384 * 8) [blocks of memory](https://github.com/valkey-io/valkey/blob/e65b2d235c300bb86cc7f960883ad919f75162e6/src/cluster_legacy.h#L326) to determine who owns the slot being accessed, whether that slot is migrating, and whether that is slot is being imported. The amount of data far exceeds L1 cache sizes and will also exceed most L2 caches, so we are likely experiencing a large number of cache misses because of these lookups. If you remove all three of those bottlenecks, you get about a 5% performance overall performance improvement.

My thought is that we can use the slot bits on the clusterNode object to determine if we own the slot. We assume that most of the time the client is going to send requests to the right node. We can also keep a smaller structure of all the nodes we are importing from and migrating to, so we don't have to check the map on every lookup.

Potential followup of https://github.com/valkey-io/valkey/pull/631.

adetunjii:
@madolson do you mind if we work on this together? I'm not exactly sure how some of these things work but we can definitely work through it together.

PingXie:
>I realized that we basically are checking three 171kb (16384 * 8) [blocks of memory](https://github.com/valkey-io/valkey/blob/e65b2d235c300bb86cc7f960883ad919f75162e6/src/cluster_legacy.h#L326) to determine who owns the slot being accessed

Aren't these point look-ups? The total size shouldn't matter but I see your point that will be cache misses given the way these 3 arrays are arranged. I wonder if we should group the 3 states (ownership/migrating/importing) by the slot, i.e., replacing them with a 2-d array like `slot_states[16384][3]`.

Also while checking out `getNodeByQuery`, I noticed that keys can be hashed more than once on the data serving path with per-slot dict.

madolson:
> Aren't these point look-ups? The total size shouldn't matter but I see your point that will be cache misses given the way these 3 arrays are arranged. 

They are! The problem is basically every lookup into these arrays is an L2 cache miss (and often an L3 + TLB miss), which are very expensive (even compared to hashing). 

> I wonder if we should group the 3 states (ownership/migrating/importing) by the slot, i.e., replacing them with a 2-d array like slot_states[16384][3].

[...]

PR #2092, opened 2025-05-16
Optimize pipelining by parsing and prefetching multiple commands

### LINKED ISSUE #2044: Parse multiple commands (look ahead)
Currently, only one command can be parsed per client.

If there is more data in the input buffer, we could look ahead in the input buffer to parse more commands, lookup the commands and incrementally prefetch the keys.

When IO-threads are used, the IO threads can do this work to offload the main thread. But even for single-threaded mode.

To store multiple commands in the client struct, we need some refactoring to the parsing code. The current command can still be `c->argv` but additional commands could be stored in a list belonging to the client.

The main benefit is to optimize pipelines.

zuiderkwast:
Instead of parsing only one command per client before executing it, parse multiple commands from the query buffer and batch-prefetch the keys accessed by the commands in the queue before executing them.

This is an optimization for pipelined commands, both with and without I/O threads. The optimization is currently disabled for the replication stream, due to failures (probably caused by how the replication offset is calculated based on the query buffer offset).

* When parsing commands from the input buffer, multiple commands are parsed and stored in a command queue per client.
* In single-threaded mode (I/O threads off) keys are batch-prefetched before the commands in the queue are executed. Multi-key commands like MGET, MSET and DEL benefit from this even if pipelining is not used.
* Prefetching when I/O threads are used does prefetching for multiple clients in parallel. This code takes client command queues into account, improving prefetching when pipelining is used. The batch size is controlled by the existing config `prefetch-batch-max-size` (default 16), which so far only was used together with I/O threads. The config is moved to a different section in `valkey.conf`.
* When I/O threads are used and the maximum number of keys are prefetched, a client's command is executed, then the next one in the queue, etc. If there are more commands in the queue for which the keys have not been prefetched (say the client sends 16 pipelined MGET with 16 keys in each) keys for the next few commands in the queue are prefetched before the commands is executed if prefetching has not been done for the next command. (This utilizes the code path used in single-threaded mode.)

Code improvements:

* Decoupling of command parser state and command execution state:
  * The variables reqtype, multibulklen and bulklen refer to the current position in the query buffer. These are no longer reset in resetClient (which runs after each command being executed). Instead, they are reset in the parser code after each completely parsed command.
  * The command parser code is partially decoupled from the client struct. The query buffer is still one per client, but the resulting argument vector is stored in caller-defined variables.

Benchmarking with and without I/O threads, see below.

Fixes #2044


src/memory_prefetch.c
```diff
@@ -258,6 +259,24 @@ int addCommandToBatchAndProcessIfFull(client *c) {
         getKeysFreeResult(&result);
     }
 
+    /* Prefetch keys in command queue. */
```
uriyage: Maybe we can refactor this code to avoid duplication

src/server.h
```diff
@@ -1197,6 +1209,11 @@ typedef struct client {
     int multibulklen;    /* Number of multi bulk arguments left to read. */
     long bulklen;        /* Length of bulk argument in multi bulk request. */
     long long woff;      /* Last write global replication offset. */
+    /* Parsed commands queue */
+    commandParserState *cmd_queue; /* Queue of parsed commands. */
+    short cmd_queue_len;           /* Number of elements in the queue. */
+    short cmd_queue_off;           /* Offset to the next element to execute. */
+    short cmd_queue_cap;           /* Allocation size (capacity) of the cmd_queue array. */
```
uriyage: Consider putting it all in one struct cmdQueue

zuiderkwast: I considered it. Also considered implementing some generic queue, but the code got simpler without too much abstraction.

Do you mean to put the queue struct inline in the client struct or to allocate it and store a pointer to it in the client struct?

dvkashapov (approved):
Did some benchmarks on AMD EPYC 7713 32 CPUs
- config:
```
databases 1
save ""
appendonly no
rdbcompression no
activedefrag no
maxclients 1000
io-threads *
io-threads-do-reads yes
protected-mode no
maxmemory 2500mb
maxmemory-policy noeviction
hz 1
latency-monitor-threshold 0
```
- cset shield
- cpu bind
- cache population
`valkey-benchmark -h <ip> -c 96 --threads 96 -d 256 -t set -n 3000000 -r 3000000 --sequential`
- benchmark itself
`valkey-benchmark -h <ip> -c 192 --threads 96 -d 256 -t get -n 10000000 -r 3000000 -P 10`

io-threads 1 unstable result:
```
Summary:
  throughput summary: 352795.91 requests per second
  latency summary (msec):
        5.396     0.048     5.279     5.727    10.455   124.287

```
io-threads 1 this pr result:
```
Summary:
  throughput summary: 604375.69 requests per second
  latency summary (msec):
        3.100     0.040     2.935     3.431     5.791   132.991

```
io-threads 9 unstable result:
```
Summary:
  throughput summary: 383435.59 requests per second
  latency summary (msec):
        4.981     0.056     4.959     5.255     5.503    79.039

```

io-threads 9 this pr result:
```
Summary:
  throughput summary: 738334.31 requests per second
  latency summary (msec):
        2.553     0.032     2.567     2.751     2.919    36.703

```

Seems like a solid 2x improvement in pipelining workload! Also did not find any memory leaks or instabilities during testing, LGTM!

src/server.h
```diff
@@ -1182,6 +1182,18 @@ typedef struct ClientModuleData {
...
 
+/* Parser state and parse result of a command from a client's input buffer. */
+typedef struct commandParserState {
+    int read_flags; /* complete, error or 0 (parsing not complete) */
+    int argc;
+    robj **argv;
+    int argv_len;
+    int slot;
+    size_t argv_len_sum;
+    unsigned long long input_bytes;
```
PingXie, line 1245: interesting that you need to save off `input_bytes` but I had to save off `duration` to make the slot-level metrics tests happy. not that a big issue at this moment though. 

`read_flags` though is more generic than my `valid_arity` but I think we should turn it into a bitfield struct; but again not a "now" problem

src/server.h
PingXie: I agree with @uriyage that a proper abstraction of `cmdQueue` helps with readability. but again this is quite a trivial change.

src/networking.c
```diff
@@ -2887,62 +2896,111 @@ static void setProtocolError(const char *errstr, client *c) {
...
  *
  * This function is called if processInputBuffer() detects that the next
  * command is in RESP format, so the first byte in the command is found
  * to be '*'. Otherwise for inline commands processInlineBuffer() is called. */
 void processMultibulkBuffer(client *c) {
+    int flag = parseMultibulk(c, &c->argc, &c->argv, &c->argv_len,
+                              &c->argv_len_sum, &c->net_input_bytes_curr_cmd);
+    c->read_flags |= flag;
+
+    if (c->read_flags & READ_FLAGS_AUTH_REQUIRED) {
```
PingXie, line 3402: this flag is important on the read IO path so we don't risk a DOS attach (via excessive memory allocation) but I don't think we need to worry about this flag on the command parsing path, which doesn't manage the query buffer size.

src/networking.c
```diff
@@ -2887,62 +2896,111 @@ static void setProtocolError(const char *errstr, client *c) {
...
+        /* Execute client's AUTH command before parsing more, because it affects
+         * parser limits for max allowed bulk and multibulk lengths. */
+        return;
+    }
+
+    /* Try parsing pipelined commands. */
```
PingXie, line 3415: this loop is duplicative imo - see `processInputBuffer`

src/networking.c
```diff
@@ -3238,17 +3299,152 @@ int canParseCommand(client *c) {
...
         c->read_flags |= authRequired(c) ? READ_FLAGS_AUTH_REQUIRED : 0;
 
-        parseCommand(c);
+        /* If commands are queued up, pop from the queue first */
+        if (!consumeCommandQueue(c)) {
+            parseCommand(c);
+        }
+
+        /* Prefetch keys for the next commands in queue, if not already done. */
+        prefetchCommandQueueKeys(c);
```
PingXie, line 3988: any particular reason not to integrate with `addCommandToBatchAndProcessIfFull`?

PingXie: also `processInputBuffer` is executed on both io threads and the main thread. I don't think we should prefetch keys on the io threads.

PingXie: one more thing. `processCommandAndResetClient` return `C_OK` as well when encountering blocking command. I think we should bail out the loop when the client is blocked.

PingXie (commented):
@zuiderkwast, I think our changes are indeed very close, especially around the command queue concept. that being said (and I know I am super biased here :)), I think my PR is closer to the finishing line. Do you want to hop onto my branch and collaborate on mine next?

https://github.com/PingXie/valkey/tree/pipeline

src/server.h
zuiderkwast: This springs from the fact that I went though each line in the parser to decouple the currently *parsed* command from the currently *executed* command.

`intput_bytes` is related to each command being *parsed*. It's computed by the parser and it describes the number of bytes of each command being parsed.

In `commandProcessed()`, after a command has been executed, the current command's input bytes are accumulated slot statistics network-bytes-in. This is per slot, and each command can be for a different slot.

I can't see how you can get this correct if you *don't* do this per command in the queue. Am I missing anything?

Regarding `read_flags`, these are per parsed command too. Parse errors and parsing-complete are read flags stored by the command parser.

`c->duration` on the other hand is related to command *execution*. Can you explain why you would need to store this per command in the queue?

src/networking.c
zuiderkwast: No, the parsing code has conditions on max multibulk length that depends on auth required:

```
        size_t multibulklen_slen = newline - (c->querybuf + 1 + c->qb_pos);
        ok = string2ll(c->querybuf + 1 + c->qb_pos, multibulklen_slen, &ll);
        if (!ok || ll > INT_MAX) {
            return READ_FLAGS_ERROR_INVALID_MULTIBULK_LEN;
        } else if (ll > 10 && auth_required) {
            return READ_FLAGS_ERROR_UNAUTHENTICATED_MULTIBULK_LEN;
        }
```

If we have a pipeline where the first command is AUTH (successfully authenticating a client when it's executed) and the second command is a some command with many arguments, then if we parse the full pipeline before the AUTH command is executed, the second command will be rejected by the parser.

I can write a test case for it if you want.

src/networking.c
zuiderkwast: I don't understand this comment. I guess I did something different that what you did. :)

This is the call chain:

    processInputBuffer -> parseCommand -> processMultibulkBuffer.

The loop here in processMultibulkBuffer just parses commands.

The loop in processInputBuffer is executing one command in each iteration. It calls processCommandAndResetClient in each iteration. It takes each command from the queue, or calls parseCommand if the queue is empty.

Perhaps we should rename one of these functions? Process = parse + execute?

src/networking.c
zuiderkwast: > any particular reason not to integrate with `addCommandToBatchAndProcessIfFull`?

I confess this is duplicated logic. :)

There are subtle differences though, but I suppose they can be solved and the code be unified.

`addCommandToBatchAndProcessIfFull` first prefetches the batch and then it executes each client's command. Here it executes *all* commands in the queue. But only the first 16 (by default) keys have been prefetched. If a single client has a long pipeline, let's say 20 MGET commands with 10 keys each, then most of the keys will not be prefetched.

What I want to do is to execute only the commands that have been prefetched, then prefetch and execute a second batch, and so on.

I added this code for single-threaded mode first, because the existing addCommandToBatchAndProcessIfFull seems quite coupled with io-threading code. With some work it can be used for single-theaded mode too though.



> also `processInputBuffer` is executed on both io threads and the main thread. I don't think we should prefetch keys on the io threads.

No, it isn't. The IO thread has this call chain:

    ioThreadReadQueryFromClient -> parseCommand -> processMultibulkBuffer

The IO thread doesn't run the prefetch code. When the main thread takes over from the IO threads, it is this code:

    processIOThreadsReadDone -> processPendingCommandAndInputBuffer -> processCommandAndResetClient

Then, processInputBuffer is called, which consumes the queue and potentially also parses more from the input buffer. Some commands in the queue may have been prefetched and some not, which is what the new `READ_FLAGS_PREFETCHED` is used for.

The fallback to parse more commands from the input buffer here covers the scenario where the first command in a pipeline is AUTH, i.e. the case where the auth required flag affects the command parsing. In most cases, there won't be anything more to parse though.

> one more thing. `processCommandAndResetClient` return `C_OK` as well when encountering blocking command. I think we should bail out the loop when the client is blocked.

If a command is executed and ends up blocked, then `canParseCommand(c)` returns false and this breaks the loops in `processInputBuffer` and then no more commands are processed for this client. (`canParseCommand` a bad name because what it checks isn't much about parsing, but whether the next command can be executed or not. Renaming it to `canProcessCommand(c)` makes it clearer maybe.)

No special handling is required for blocked commands with this approach AFAICT.

src/server.h
madolson: I'm not sure why duration was needed in Ping's PR. It should only be needed for partially executed commands, which we won't ever have here.

src/networking.c
```diff
@@ -5524,6 +5734,26 @@ void ioThreadReadQueryFromClient(void *data) {
...
+        commandParserState *st = &c->cmd_queue.cmds[i];
+        if (!(st->read_flags & READ_FLAGS_PARSING_COMPLETED) || c->argc == 0) continue;
+        struct serverCommand *cmd = lookupCommand(st->argv, st->argc);
+        if (!cmd || !commandCheckArity(c->io_parsed_cmd, c->argc, NULL)) continue;
+        st->cmd = cmd;
+        if (server.cluster_enabled) {
+            /* Offload slot calculation to I/O thread */
+            getKeysResult result;
+            initGetKeysResult(&result);
+            int numkeys = getKeysFromCommand(st->cmd, st->argv, st->argc, &result);
```
madolson: We should probably write a helper function for this, we don't need to fully populate the key structure, we can just get the first key and compute the slot on that.  Not a blocker here, but would offload the necessary work of finding all the keys when we just need one.

src/server.h
```diff
@@ -1181,6 +1181,26 @@ typedef struct ClientModuleData {
                                                 * unloaded for cleanup. Opaque for the Server Core.*/
 } ClientModuleData;
 
+/* Parser state and parse result of a command from a client's input buffer. */
+typedef struct commandParserState {
```
```suggestion
typedef struct parsedCommand {
```
Maybe this naming made sense earlier, but it seems to imply the orchestrator and not an individual parsed command.

src/networking.c
```diff
@@ -3238,17 +3307,158 @@ int canParseCommand(client *c) {
...
+        }
+    } while (not_complete_count != 0);
+
+    /* Prefetch value pointers. */
+    for (int i = 0; i < num_keys; i++) {
+        void *entry;
+        if (hashtableIncrementalFindGetResult(&key_incr_states[i], &entry)) {
+            robj *val = entry;
+            /* TODO? Prefetch all types and encodings except OBJ_ENCODING_EMBSTR
+             * and OBJ_ENCODING_INT. */
```
madolson, line 3962: This seems right? Everything else should be prefetchable for values.

madolson (commented):
The overall approach makes sense to me, I have some minor thoughts but I haven't thought to much how to think about this vs Ping's PR yet.

Looking back at the original IO threading, we batch N keys, maybe we should change it to be batch N commands instead, since that will make sure we are fully parsing a command. That way we can probably more tightly unify the code.

src/networking.c
zuiderkwast: This is an area where we could do a lot of improvements. There are many ideas but I'm reluctant to including it in this PR, to avoid scope creep. Doing it in separate PR would be better IMO.

The solution I have in mind here is to lookup all keys and offload the cross-slot detection to IO threads. Then getNodeByQuery can be much more lightweight and not do any crc16 at all in the execution path.

To prevent the IO-threaded and single-threaded code paths from diverging too much, I think we can detect cross-slot before command execution also in single-threaded mode. We just set the slot on the client (or in the command queue) and/or a flag that says it's a crossslot error. (We could also set a flag for wrong arity, although this is not a heavy operation to check.)

* #632 
* #2077

src/server.h
zuiderkwast: Sgtm. This is used also for partially parsed commands, but that's a side minor detail.

src/networking.c
zuiderkwast: Yes we could. Only some commands like EXISTS don't need the value, and commands like UNLINK access it only in the background threads, but most others do. Maybe we could look at command flags or key specs flags to determine this... We could move this idea to a separate issue/PR if it's non-trivial.

I'd like to unify this with prefetching code with the one in memory_prefetch.c and refactor that. I will do it.

zuiderkwast:
> The overall approach makes sense to me, I have some minor thoughts but I haven't thought to much how to think about this vs Ping's PR yet.

Thanks @madolson for taking the time to look at this.

> Looking back at the original IO threading, we batch N keys, maybe we should change it to be batch N commands instead, since that will make sure we are fully parsing a command. That way we can probably more tightly unify the code.

I think we shall prefetch full commands, but we can still configure it as the number of keys. If we go over the number N when adding a command to the batch, then we can allow prefetching a few more, let's say up to N*2. That way we prefetch by default 16 keys, but allow prefetching up to 32 keys to stick to full commands.

We should definitely unify the IO-threaded and single-threaded code paths for this code and more of the parse-lookup-prefetch-execute. I avoided refactoring to keep the diff small but I think it's necessary to at least rename some functions to keep this comprehensive.

src/server.h
PingXie: >c->duration on the other hand is related to command execution. Can you explain why you would need to store this per command in the queue?

I haven't looked deep but without it I got failure in one of the `cluster slot-stat` tests. I am not worried about it at this point. I think you have a good point that we should clean up the existing code a bit first.

PingXie:
> > Looking back at the original IO threading, we batch N keys, maybe we should change it to be batch N commands instead, since that will make sure we are fully parsing a command. That way we can probably more tightly unify the code.
> 
> I think we shall prefetch full commands, but we can still configure it as the number of keys. If we go over the number N when adding a command to the batch, then we can allow prefetching a few more, let's say up to N*2. That way we prefetch by default 16 keys, but allow prefetching up to 32 keys to stick to full commands.

my preference, for the sake of simplicity, is that we continue to count keys. as soon as we use up `max_pretech_size` we should just execute the command, irrespective of the number of the keys referenced by the command (e.g., `mget`) or the number of commands batached.

ultimately `max_prefetch_size` is determined by the L1 cache size. There is only so much we can prefetch. over-prefetching will lead to L1 cache thrashing, nullifying the effect.

the L1 data cache on Intel cpus is 32 KiB or 48 KiB, typically, with a cache line size of 64 bytes, which gives us 512 or 768 cache lines, respectively. I haven't counted carefully the number of memory accesses needed to prefetch a key-value pair of strings in the new hashtable but I guess it would be 5 (including both keys and hash table)? Assuming 5, prefetching 32 keys would consume 160 cache lines, which is not a trivial number compared to 512 already.

secwall:
Hello, it seems that this pull request has some conflicts with the unstable branch. Do you need any assistance working on it? Merging it into 9.0 would be beneficial for many users (a lot of clients use pipelines automatically).

zuiderkwast:
After discussion with @PingXie and others, I wanted to rewrite it a bit and unify the prefetching code paths for io-threaded and single-threaded modes. I probably can't make it to 9.0 though.

How would folks (@PingXie, @madolson) feel about merging this more or less as is for 9.0 and improving it later?

PingXie:
>How would folks (@PingXie, @madolson) feel about merging this more or less as is for 9.0 and improving it later?

works for me. Will review this pr next.

secwall:
Hello. I apologize for the inconvenience, but the 9.0 feature freeze is imminent. How can I assist in ensuring that this PR is included in the 9.0 version?

dvkashapov:
Re-did some benchmarks on currently supported versions. The most important conclusion: **9.0 without this PR will be 3rd consecutive version that is significantly slower than 7.2 with enabled pipelining**. I think it is important that we make sure this PR gets into 9.0 because valkey-go does pipelining by default. 
Config:
```
databases 1
save ""
appendonly no
rdbcompression no
activedefrag no
maxclients 1000
io-threads 9
protected-mode no
maxmemory 2500mb
maxmemory-policy noeviction
hz 1
latency-monitor-threshold 0
```

- cset shield
- cpu bind
- cache population: `valkey-benchmark -h <ip> -c 256 --threads 64 -d 256 -t set -n 3000000 -r 3000000`
- benchmark: `valkey-benchmark -h <ip> -c 196 --threads 48 -d 256 -t get -n 30000000 -r 3000000 -P (1 or 16)`

Results:
**7.2:
Pipe 1**
```
Summary:
  throughput summary: 204579.86 requests per second
  latency summary (msec):
        0.937     0.240     0.887     1.079     1.719  3000.319
```
**Pipe 16**
```
Summary:
  throughput summary: 949307.00 requests per second
  latency summary (msec):
        2.984     0.168     3.031     3.359     3.479     4.631
```

**8.0:
Pipe 1**
```
Summary:
  throughput summary: 590876.88 requests per second
  latency summary (msec):
        0.325     0.032     0.335     0.455     0.511     2.687
```
**Pipe 16**
```
Summary:
  throughput summary: 633058.31 requests per second
  latency summary (msec):
        4.933     0.064     4.911     5.151     5.375     9.327
```

**8.1:
Pipe 1**
```
Summary:
  throughput summary: 584487.69 requests per second
  latency summary (msec):
        0.327     0.040     0.335     0.463     0.519     2.623
```
**Pipe 16**
```
Summary:
  throughput summary: 653879.69 requests per second
  latency summary (msec):
        4.777     0.056     4.743     5.007     6.431     8.223
```

**unstable:
Pipe 1**
```
Summary:
  throughput summary: 567923.69 requests per second
  latency summary (msec):
        0.337     0.032     0.351     0.463     0.519     4.511
```
**Pipe 16**
```
Summary:
  throughput summary: 747924.50 requests per second
  latency summary (msec):
        4.170     0.096     4.167     4.287     4.431     8.743
```

Conclusions:
Without pipelining performance is significantly better in 8.0/8.1 compared to 7.2. 
With pipelining 16 performance is significantly better in 7.2 compared to 8.0/8.1.

I rebased this branch onto unstable in [JoBeR007/valkey@rebased](https://github.com/JoBeR007/valkey/tree/rebased) but I think i need to take a closer look at stability of this PR being applied to unstable, as some tests are now failing.

zuiderkwast:
@dvkashapov Thank you! I have rebased it locally and fixed some problems, but I still get failures locally. I can push what I have so we can debug it together. It would indeed be nice to get this in 9.0 (even after RC1 might be acceptable).

zuiderkwast:
@valkey-io/core-team This PR has been updated and I believe this is working well now. Can we review it and get it into 9.0-rc2?

As @JoBeR007 pointed out, 8.x is slower than 7.2 when pipelining is used, and even 9.0 would be unless we take this PR (or something like it).

This is also supposed to compensate for our reject of cross-slot MGET. This PR makes multiple pipelined small MGET or GET can be must faster than today.

src/networking.c
zuiderkwast: I wanted to refactor `addCommandToBatchAndProcessIfFull` (memory_prefetch.c) but it requires a more huge refactoring.

src/networking.c
zuiderkwast: I left this as is, to match the logic in `memory_prefetch.c`. I feel that it's a separate feature the value in more cases.

madolson: Can we make an issue instead of having this TODO? It'll be hard to find this specific context.

src/networking.c
```diff
@@ -3684,6 +3750,9 @@ int processPendingCommandAndInputBuffer(client *c) {
  *
  * Sets the client's read_flags to indicate the parsing outcome */
 void parseCommand(client *c) {
+    /* The command queue must be emptied before parsing. */
+    debugServerAssert(!consumeCommandQueue(c));
```
madolson: Not sure why this isn't a real assert. Won't the following code corrupt the command processing?

src/networking.c
```diff
@@ -3728,17 +3821,159 @@ int canParseCommand(client *c) {
     return 1;
 }
 
+/* Pops a commands from the command queue and sets it as the client's current
```
```suggestion
/* Pops a command from the command queue and sets it as the client's current
```
nit

src/networking.c
```diff
@@ -3702,6 +3771,30 @@ void parseCommand(client *c) {
...
+    if (queue->cmds != NULL) {
+        if (queue->len == 0) {
+            zfree(queue->cmds);
+            queue->cmds = NULL;
+            queue->cap = 0;
+        } else {
+            /* Try shrink to the next power of two >= len */
+            const int bits = CHAR_BIT * sizeof(unsigned int);
+            uint16_t cap = queue->len == 1 ? 1 : 1 << (bits - __builtin_clz(queue->len - 1));
+            if (cap == 0) return; /* overflow */
```
madolson: This should never overflow since the max size is 512 right? Would almost prefer this was an assert instead.

src/networking.c
madolson: Process = Parse + execute makes sense to me fwiw.

src/networking.c
```diff
@@ -3728,17 +3821,159 @@ int canParseCommand(client *c) {
     return 1;
 }
 
+/* Pops a commands from the command queue and sets it as the client's current
+ * command. Returns true on success and false if the queue was empty. */
+static bool consumeCommandQueue(client *c) {
+    cmdQueue *queue = &c->cmd_queue;
+    if (queue->off >= queue->len) return false;
+    parsedCommand *p = &queue->cmds[queue->off++];
+    c->read_flags |= p->read_flags;
```
madolson, line 3842:
```suggestion
    c->read_flags = p->read_flags;
```
Was this intentional? I think this looks safe, but it might subtly introduce bugs in the future if some flag is preserved .

src/server.c
```diff
@@ -4085,25 +4085,50 @@ uint64_t getCommandFlags(client *c) {
...
+    } else if (server.cluster_enabled) {
         debugServerAssert(c->slot == -1 &&
                           !(c->read_flags & READ_FLAGS_CROSSSLOT) &&
                           !(c->read_flags & READ_FLAGS_NO_KEYS));
         c->slot = clusterSlotByCommand(c->parsed_cmd, c->argv, c->argc, &c->read_flags);
     }
 }
 
+/* Prepare all parsed commands in the client's queue. See prepareCommand(). */
+void prepareCommandQueue(client *c) {
```
madolson, line 4118: It would be nice if this function called prepareCommand for both the main command as well as the ones in the command queue, instead of having all the code duplicated.

madolson (approved):
It would be nice to do some refactoring, there is a good bit of duplicate code, but I don't think it's strictly needed. I didn't see any blockers for me though.

I don't understand the test failure though, I tried to reproduce it locally. It looks like this one though: https://github.com/valkey-io/valkey/pull/2485/files

madolson:
I was able to reproduce the test locally. It's a bit flaky. It seems like the migration is getting stuck somehow.

src/networking.c
zuiderkwast: Yes, it's intentional. In the beginning of `processInputBuffer()`, just before we call `consumeCommandQueue()`, the client's read flags are initialized like this:

```C
        c->read_flags = isReplicatedClient(c) ? READ_FLAGS_REPLICATED : 0;
        c->read_flags |= authRequired(c) ? READ_FLAGS_AUTH_REQUIRED : 0;
```

We need to preserve those.

If this `|=` is bug prone, maybe we can change it. I don't know right now how we can do that. Do you have an idea?

In `processMultbulkBuffer()` (which is actually just parsing it, so rename to `parseMultibulkBuffer()`?) we also do `c->read_flags |= flag`.

src/networking.c
zuiderkwast: Yeah, I'll do a real assert. Actually, I wanted to check that it's empty, not try to pop a command. I'll change it to

```Suggestion
    serverAssert(client->cmdQueue.len == 0);
```

src/networking.c
zuiderkwast: I'm renaming existing function now:

        parseCommand -> parseInputBuffer
        processMultibulkBuffer -> parseMultibulkBuffer
        processInlineBuffer -> parseInlineBuffer

Should I also rename `processCommand -> executeCommand`? I'm not sure. I haven't done it yet.

src/server.c
zuiderkwast: Added a prepareCommandGeneric that works on pointers to argv, argc, read_flags, cmd and slot.

src/networking.c
madolson: Maybe just documenting it would be enough. It's confusing that some of these are command flags (READ_FLAGS_PARSING_COMPLETED ) and some of these are client state flags (AUTH_REQUIRED)

madolson:
We're getting de-sync errors in the test:
```
24786:S 19 Aug 2025 09:52:35.066 # Protocol error (Master using the inline protocol. Desync?) from client: id=32 addr=127.0.0.1:21116 laddr=127.0.0.1:61112 fd=25 name= age=11 idle=0 flags=M capa= db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=16010 qbuf-free=49520 argv-mem=0 multi-mem=0 rbs=1024 rbp=0 obl=0 oll=0 omem=0 tot-mem=67136 events=r cmd=set user=(superuser) redir=-1 resp=2 lib-name= lib-ver= tot-net-in=12800433 tot-net-out=451 tot-cmds=20579. Query buffer: '6ZJ}:226..*2..$6..UNLINK..$8..{6ZJ}:22..*2..$6..UNLINK..$8..{6ZJ' (... more 14880 bytes ...) '..*2..$6..UNLINK..$9..{6ZJ}:207..*2..$6..UNLINK..$9..{6ZJ}:280..'
```

Later in the day update:
The target eventually stops processing incoming requests, but it is not consistent. There doesn't seem to be anything non-deterministic about it that I can see outside of the speed of processing. @murphyjacob4 Since your back, if you have any thoughts as well for the test failures.

Just to comment since Viktor messaged me on slack. Apparently there is something getting incorrectly updated for replication offset. This will have impact for other deeply pipelined replication streams, but it looks like only the slot migration tests are triggering that.

src/networking.c
zuiderkwast: Test case added in auth.tcl

murphyjacob4:
> Apparently there is something getting incorrectly updated for replication offset. This will have impact for other deeply pipelined replication streams, but it looks like only the slot migration tests are triggering that.

Since slot migration uses chaining replication, if the replication offset is wrong, what is sent to the replicas of the target would be wrong:

https://github.com/valkey-io/valkey/blob/unstable/src/networking.c#L3677-L3689

So that would make sense why we see the desync errors

PR #831, opened 2024-07-26
Add maxmemory-reserved parameter for evicting key earlier to avoid OOM

hwware:
All cases are under caching, no 2nd storage disk. 
When user set **maxmemory** in config file,  OOM (out of memory) message could be returned to user once Valkey reaches the maxmemory threshold. 

The PR **Incremental eviction processing** https://github.com/redis/redis/pull/7653 solves some cases of OOM issues, but in some scenarios, OOM still happens. Through this PR, we could reduce the probability of OOM occurrence or delay the OOM, especially in heavy-write case.

**Example:**

We have one VM 100GB, and user set **maxmemory** as 10GB.  

If Valkey server is working on a heavy-write process with key-value pair is big (200MB), and we increase **maxmemory-eviction-tenacity** to adapt to this situation.  Before we apply this PR, the following actions will happen：
**step 0**:  used memory is 10GB
**step 1**:  a new write command processes successfully, used memory reaches 10.2GB (no eviction happens)
**step 2**:  a new write command comes, 150MB key is evicted, and used memory is 10.05GB,  OOM happens
**step 3**:  User retries the previous write command, another 150MB key is evicted. And used memory is just below 10GB, 
            the write command can be processed successfully

With this PR, we could set **maxmemory-reserved** as 9GB. Thus, once the Valkey comes to the 9GB memory, key eviction process will begin. 

**step 0**:  used memory is 9GB
**step 1**:  a new write command processes successfully, used memory reaches 9.2GB (no eviction happens)
**step 2**:  the second write command comes, 150MB key is evicted, and used memory is 9.05GB
                   the write command executes successfully, used memory update to 9.25 GB
**step 3**:   the third write command comes, 150MB key is evicted, and used memory is 9.1GB
                   the write command executes successfully, used memory update to 9.3 GB
...........
After more than 20 steps, OOM happens.

From the above example, we can see that although with a heavy-write system, the used memory is increasing, but we can significantly delay the OOM.

PR #2095, opened 2025-05-17
Allow shrinking hashtables in low memory situations

Fusl:
During memory eviction, we may end up with a primary kv hash table that's below the 13% `MIN_FILL_PERCENT_SOFT`, but inside `resize()` for the kv table we check with `ht->type->resizeAllowed` whether allocating a new table crosses the configured `maxmemory` limit without taking into account that shrinking, while temporarily increasing the memory usage, will ultimately reduce the memory usage in all situations.

Blocking hash tables from shrinking in situations where we would only temporarily cross the maxmemory limit has at least three negative side effects:
* The performance of valkey is negatively impacted if, for example, we're moving items from table A to table B since table A would eventually end up being a very sparse hash table that never shrinks.
* In extreme situations, the performance of valkey slows down significantly when trying to find more keys to evict since we're trying to find keys in a sparse table that only becomes more sparse over time.
* In more extreme situations, such as when reducing `maxmemory` by around 90% or writing large data to a single key, valkey can end up in a situation where, no matter how many keys are being evicted from the kv table, the table bucket struct overhead uses more memory than the data stores in the table, causing all keys to be evicted from the table and valkey becoming unusable before setting maxmemory to above the current memory usage:

```
127.0.0.1:6379> config set maxmemory 2gb
OK
127.0.0.1:6379> debug populate 20000000
OK
(8.94s)
127.0.0.1:6379> debug htstats 0
[Dictionary HT]
Hash table 0 stats (main hash table):
 table size: 29360128
 number of entries: 20000000
[Expires HT]
127.0.0.1:6379> config set maxmemory 200mb
OK
127.0.0.1:6379> dbsize
(integer) 18725794
[...]
127.0.0.1:6379> dbsize
(integer) 0
127.0.0.1:6379> keys *
(empty array)
127.0.0.1:6379> set foo bar
(error) OOM command not allowed when used memory > 'maxmemory'.
127.0.0.1:6379> config get maxmemory
1) "maxmemory"
2) "209715200"
127.0.0.1:6379> info memory
# Memory
used_memory:269890904
used_memory_human:257.39M
[...]
127.0.0.1:6379> config set maxmemory 300mb
OK
127.0.0.1:6379> info memory
# Memory
used_memory:1455512
used_memory_human:1.39M
[...]
127.0.0.1:6379> set foo bar
OK
```

This PR addresses these issues by allowing hash tables to allocate a new table for shrinking, even if allocating a new table causes the maxmemory limit to be crossed during resize.

Additionally, the check for whether we should fast-forward the resize operation has been moved to before checking whether we are allowed to allocate a new resize table. Because we don't start a resize in this section of the function yet, a fast-forward usually drops the memory usage enough to allow another resize operation to start immediately after the fast-forward resize completes.

As a side effect of this PR, we now also refuse to start a hashtable resize if the policy is set to `HASHTABLE_RESIZE_FORBID`.

### DIFF
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -614,14 +614,6 @@ static int resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
         return 0;
     }
 
-    if (ht->type->resizeAllowed) {
-        double fill_factor = (double)min_capacity / ((double)numBuckets(old_exp) * ENTRIES_PER_BUCKET);
-        if (fill_factor * 100 < MAX_FILL_PERCENT_HARD && !ht->type->resizeAllowed(alloc_size, fill_factor)) {
-            /* Resize callback says no. */
-            return 0;
-        }
-    }
-
     /* We can't resize if rehashing is already ongoing. Fast-forward ongoing
      * rehashing before we continue. This can happen only in exceptional
      * scenarios, such as when many insertions are made while rehashing is
@@ -630,6 +622,14 @@ static int resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
         if (hashtableIsRehashingPaused(ht)) return 0;
         while (hashtableIsRehashing(ht)) {
             rehashStep(ht);
+        }
+    }
+
+    if (ht->type->resizeAllowed) {
+        double fill_factor = (double)min_capacity / ((double)numBuckets(old_exp) * ENTRIES_PER_BUCKET);
+        if (fill_factor * 100 < MAX_FILL_PERCENT_HARD && !ht->type->resizeAllowed(exp > old_exp ? alloc_size : 0, fill_factor)) {
+            /* Resize callback says no. */
+            return 0;
         }
     }
 
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -634,7 +634,7 @@ static int resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
         return 0;
     }
     if (exp > old_exp && ht->type->resizeAllowed) {
-        /* If we're growing the table, let's check if the resize is allowed (for now just does a maxmemory checks). */
+        /* If we're growing the table, let's check if the resizeAllowed callback allows the resize. */
         double fill_factor = (double)min_capacity / ((double)numBuckets(old_exp) * ENTRIES_PER_BUCKET);
         if (fill_factor * 100 < MAX_FILL_PERCENT_HARD && !ht->type->resizeAllowed(alloc_size, fill_factor)) {
             /* Resize callback says no. */
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -464,10 +464,6 @@ static void rehashingCompleted(hashtable *ht) {
     ht->child_buckets[0] = ht->child_buckets[1];
     resetTable(ht, 1);
     ht->rehash_idx = -1;
-
-    /* Do a recursive call back to the resize function in case
-     * we need to do another resize run immediately after. */
-    hashtableRightsizeIfNeeded(ht);
 }
 
 /* Reverse bits, adapted to use bswap, from
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -464,6 +464,10 @@ static void rehashingCompleted(hashtable *ht) {
     ht->child_buckets[0] = ht->child_buckets[1];
     resetTable(ht, 1);
     ht->rehash_idx = -1;
+
+    /* Do a recursive call back to the resize function in case
+     * we need to do another resize run immediately after. */
+    hashtableRightsizeIfNeeded(ht);
 }
 
 /* Reverse bits, adapted to use bswap, from
@@ -625,9 +629,14 @@ static int resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
         }
     }
 
-    if (ht->type->resizeAllowed) {
+    if (resize_policy == HASHTABLE_RESIZE_FORBID && ht->tables[0]) {
+        /* Refuse to resize if resizing is forbidden and we already have a primary table. */
+        return 0;
+    }
+    if (exp > old_exp && ht->type->resizeAllowed) {
+        /* If we're growing the table, let's check if the resize is allowed (for now just does a maxmemory checks). */
         double fill_factor = (double)min_capacity / ((double)numBuckets(old_exp) * ENTRIES_PER_BUCKET);
-        if (fill_factor * 100 < MAX_FILL_PERCENT_HARD && !ht->type->resizeAllowed(exp > old_exp ? alloc_size : 0, fill_factor)) {
+        if (fill_factor * 100 < MAX_FILL_PERCENT_HARD && !ht->type->resizeAllowed(alloc_size, fill_factor)) {
             /* Resize callback says no. */
             return 0;
         }
@@ -1242,14 +1251,13 @@ int hashtableTryExpand(hashtable *ht, size_t size) {
 }
 
 /* Expanding is done automatically on insertion, but less eagerly if resize
- * policy is set to AVOID or FORBID. After restoring resize policy to ALLOW, you
- * may want to call hashtableExpandIfNeeded. Returns 1 if expanding, 0 if not
- * expanding. */
+ * policy is set to AVOID and not at all if set to FORBID.
+ * Returns 1 if expanding, 0 if not expanding. */
 int hashtableExpandIfNeeded(hashtable *ht) {
     size_t min_capacity = ht->used[0] + ht->used[1] + 1;
     size_t num_buckets = numBuckets(ht->bucket_exp[hashtableIsRehashing(ht) ? 1 : 0]);
     size_t current_capacity = num_buckets * ENTRIES_PER_BUCKET;
-    unsigned max_fill_percent = resize_policy == HASHTABLE_RESIZE_AVOID ? MAX_FILL_PERCENT_HARD : MAX_FILL_PERCENT_SOFT;
+    unsigned max_fill_percent = resize_policy == HASHTABLE_RESIZE_ALLOW ? MAX_FILL_PERCENT_SOFT : MAX_FILL_PERCENT_HARD;
     if (min_capacity * 100 <= current_capacity * max_fill_percent) {
         return 0;
     }
@@ -1257,19 +1265,32 @@ int hashtableExpandIfNeeded(hashtable *ht) {
 }
 
 /* Shrinking is done automatically on deletion, but less eagerly if resize
- * policy is set to AVOID and not at all if set to FORBID. After restoring
- * resize policy to ALLOW, you may want to call hashtableShrinkIfNeeded. */
+ * policy is set to AVOID and not at all if set to FORBID.
+ * Returns 1 if shrinking, 0 if not shrinking. */
 int hashtableShrinkIfNeeded(hashtable *ht) {
     /* Don't shrink if rehashing is already in progress. */
-    if (hashtableIsRehashing(ht) || resize_policy == HASHTABLE_RESIZE_FORBID || ht->pause_auto_shrink) {
+    if (hashtableIsRehashing(ht) || ht->pause_auto_shrink) {
         return 0;
     }
     size_t current_capacity = numBuckets(ht->bucket_exp[0]) * ENTRIES_PER_BUCKET;
-    unsigned min_fill_percent = resize_policy == HASHTABLE_RESIZE_AVOID ? MIN_FILL_PERCENT_HARD : MIN_FILL_PERCENT_SOFT;
+    unsigned min_fill_percent = resize_policy == HASHTABLE_RESIZE_ALLOW ? MIN_FILL_PERCENT_SOFT : MIN_FILL_PERCENT_HARD;
     if (ht->used[0] * 100 > current_capacity * min_fill_percent) {
         return 0;
     }
     return resize(ht, ht->used[0], NULL);
+}
+
+/* Resizes the hashtable to an optimal size, based on the current number of
+ * entries. This is a convenience function that first tries to shrink the table
+ * if needed, and then expands it if needed. After restoring resize policy
+ * to ALLOW, you may want to call hashtableShrinkIfNeeded.
+ * Returns 1 if resizing was performed, 0 otherwise. */
+int hashtableRightsizeIfNeeded(hashtable *ht) {
+    int ret = hashtableShrinkIfNeeded(ht);
+    if (!ret) {
+        ret = hashtableExpandIfNeeded(ht);
+    }
+    return ret;
 }
 
 /* Defragment the main allocations of the hashtable by reallocating them. The
--- a/src/kvstore.c
+++ b/src/kvstore.c
@@ -631,9 +631,7 @@ void kvstoreTryResizeHashtables(kvstore *kvs, int limit) {
     for (int i = 0; i < limit; i++) {
         int didx = kvs->resize_cursor;
         hashtable *ht = kvstoreGetHashtable(kvs, didx);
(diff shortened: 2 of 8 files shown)


src/hashtable.c
```diff
@@ -633,6 +625,14 @@ static int resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
         }
     }
 
+    if (ht->type->resizeAllowed) {
+        double fill_factor = (double)min_capacity / ((double)numBuckets(old_exp) * ENTRIES_PER_BUCKET);
+        if (fill_factor * 100 < MAX_FILL_PERCENT_HARD && !ht->type->resizeAllowed(exp > old_exp ? alloc_size : 0, fill_factor)) {
```
zuiderkwast: This effectively ignores the allocation size when shrinking. It means we allow memory usage to temporarily grow above maxmemory. This is risky in a system where maxmemory is set close to the available memory. In the worst case, the kernel's OOM killer will kill Valkey.

zuiderkwast (commented):
I don't think we should allow a large allocation to go above maxmemory in this case. It's risky, as mentioned in the comment below. Especially when eviction is disabled.

When the maxmemory config is changed, it's a special case maybe. Here we can maybe do something smart, such as remember how much memory we were just using and reuse some of the memory we just freed for evicting keys and use it for shrinking the hashtable...?

For the case when maxmemory is not changed, this can happen only in some unusual(?) cases like you mentioned: You delete most top-level keys while at the same time filling up single key (a hash, set, etc.).

We have had some discussions about hard and soft maxmemory limits, a separate maxmemory for keys and other ideas. See for example:

* #831

src/hashtable.c
Fusl: The `maxmemory` limit is and was never a guarantee that the memory usage of Valkey stays within those bounds, and is also explained in this comment inside `evict.c`:

https://github.com/valkey-io/valkey/blob/4092c9c8e091058e3103007b8dc05ae8ae00c1fc/src/evict.c#L498-L502

zuiderkwast: OK. This is internal comments though, not official documentation. But assuming this is the officially documented behavior, then I guess we can do this change.

zuiderkwast: After talking to @madolson about this PR, I think we can allow it.

Minor details: Just as we ignore and don't call the callback if we're over MAX_FILL_PERCENT_HARD, we can also ignore it if we're shrinking. I mean this:

```suggestion
        if (fill_factor * 100 < MAX_FILL_PERCENT_HARD &&
            exp > old_exp &&
            !ht->type->resizeAllowed(alloc_size, fill_factor)) {
```

Then we also don't need the change in evict.c

Fusl: I did it this way since `hashtableResizeAllowed` has an additional check `if (!server.dict_resizing) return 0;` (`/* For debug purposes, not allowed to be resized. */`) that I don't want skip.

zuiderkwast: What is it used for? I'm guessing it's only for testing. If it doesn't break any test cases, I think it's fine. Otherwise we can just adjust the tests that are failing.

Fusl: I moved the check up to the `if (ht->type->resizeAllowed && exp > old_exp) {` condition, this test is failing now:
```
[err]: expire scan should skip dictionaries with lot's of empty buckets in tests/unit/expire.tcl
scan didn't handle slot skipping logic.
[1/1 done]: unit/expire (17 seconds)

                   The End

  17 seconds - unit/expire

*** [err]: expire scan should skip dictionaries with lot's of empty buckets in tests/unit/expire.tcl
scan didn't handle slot skipping logic.
```
This test relies on the `dict_resizing` debug flag and in order to fix the test I'd have to move the check for `if (!server.dict_resizing) return 0;` into the `resize` function. That effectively makes the `resizeAllowed` callback function useless. Do we want to move the logic into the `resize` function and get rid of the `resizeAllowed` callback function?

zuiderkwast: There are several (too many) mechanisms to disable resizing. Another is to call `hashtableSetResizePolicy(HASHTABLE_RESIZE_FORBID)`. This is done in `updateDictResizePolicy()` in server.c (which is called e.g. after `fork()`).

I suggest that we change it like this:

```diff
 void updateDictResizePolicy(void) {
-    if (server.in_fork_child != CHILD_TYPE_NONE) {
+    if (server.in_fork_child != CHILD_TYPE_NONE || !server.dict_resizing) {
         dictSetResizeEnabled(DICT_RESIZE_FORBID);
         hashtableSetResizePolicy(HASHTABLE_RESIZE_FORBID);
     } else if (hasActiveChildProcess()) {
         dictSetResizeEnabled(DICT_RESIZE_AVOID);
         hashtableSetResizePolicy(HASHTABLE_RESIZE_AVOID);
     } else {
         dictSetResizeEnabled(DICT_RESIZE_ENABLE);
         hashtableSetResizePolicy(HASHTABLE_RESIZE_ALLOW);
     }
 }
```

and in debug.c let's call this after handling `DEBUG dict-resizing (0|1)`.

```diff
     } else if (!strcasecmp(c->argv[1]->ptr, "dict-resizing") && c->argc == 3) {
         server.dict_resizing = atoi(c->argv[2]->ptr);
+        updateDictResizePolicy();
         addReply(c, shared.ok);
```

With this change, we can delete the check for `server.dict_resizing` in `hashtableResizeAllowed()`.

I think this will be a cleaner implementation. WDYT?

Fusl: That would work but I just realized that we may not be able to get rid of the `hashtableResizeAllowed` callback anyway since I think we're using `hashtable` in non-server stuff as well and calling `overMaxmemoryAfterAlloc()` to check for the `maxmemory` constraint would mean having to pull in `evict.c` and `server.h`.
i do think that the way I initially implemented and PR'd the change was already the most reliable one we could do without introducing a bunch of scary side-effects and unknown gotchas.

zuiderkwast: > we may not be able to get rid of the `hashtableResizeAllowed` callback anyway

That's right. I didn't propose to get rid of `hashtableResizeAllowed` entirely. Only to skip calling it with 0 bytes when we're shrinking is odd and since we're not calling the `resizeAllowed` callback at all when we're over MAX_FILL_PERCENT_HARD, skipping the callback in the same way when shrinking seems more consistent with that.

> and calling `overMaxmemoryAfterAlloc()` to check for the `maxmemory` constraint would mean having to pull in `evict.c` and `server.h`.

No, why would we? The suggested changes in server.c and evict.c only affect server, not other usages such as valkey-benchmark, cli, etc.

> i do think that the way I initially implemented and PR'd the change was already the most reliable one we could do without introducing a bunch of scary side-effects and unknown gotchas.

Are you talking about usages of hashtable outside of this repo?

Fusl: Updated the PR with your proposed changes and included all the implementation changes to make it work. Please let me know if this looks good or if you'd like something changed up. Note that a lot of the changes you see here were necessary to properly merge the `dict-resizing` flag into the `resize_policy` enum and to satisfy the `expire scan should skip dictionaries with lot's of empty buckets` test.

src/hashtable.c
```diff
@@ -633,6 +629,19 @@ static int resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
         }
     }
 
+    if (resize_policy == HASHTABLE_RESIZE_FORBID && ht->tables[0]) {
+        /* Refuse to resize if resizing is forbidden and we already have a primary table. */
+        return 0;
+    }
+    if (exp > old_exp && ht->type->resizeAllowed) {
+        /* If we're growing the table, let's check if the resize is allowed (for now just does a maxmemory checks). */
```
zuiderkwast: We should not imply too much what the caller wants to use this for.

```suggestion
        /* If we're growing the table, let's check if the resize is allowed. */
```

src/hashtable.c
```diff
@@ -1242,36 +1251,48 @@ int hashtableTryExpand(hashtable *ht, size_t size) {
 }
 
 /* Expanding is done automatically on insertion, but less eagerly if resize
- * policy is set to AVOID or FORBID. After restoring resize policy to ALLOW, you
- * may want to call hashtableExpandIfNeeded. Returns 1 if expanding, 0 if not
- * expanding. */
+ * policy is set to AVOID and not at all if set to FORBID.
```
zuiderkwast, line 1250: Good fix. In an earlier phase, we wanted to use open addressing and then we had to resize even in the FORBID case, to make it possible to insert entries at all.

src/kvstore.c
```diff
@@ -631,9 +631,7 @@ void kvstoreTryResizeHashtables(kvstore *kvs, int limit) {
     for (int i = 0; i < limit; i++) {
         int didx = kvs->resize_cursor;
         hashtable *ht = kvstoreGetHashtable(kvs, didx);
-        if (ht && !hashtableShrinkIfNeeded(ht)) {
-            hashtableExpandIfNeeded(ht);
-        }
+        if (ht) hashtableRightsizeIfNeeded(ht);
```
zuiderkwast, line 634: This change isn't strictly needed, right?

I don't mind it though. We can keep it.

zuiderkwast (approved):
Yeah, this looks good to me. It's not that many changes.

An additional change worth mentioning is that FORBID is enforced now when expanding. It wasn't before. (The reason for this was that an earlier version of this hashtable was using open addressing which has an absolute limit: It could never be more than 100% full.)

There is a leak reported in the address-sanitizer job. Maybe it's fixed by #2288 that I merged yesterday. Can you merge the latest unstable and see if the leak is gone?

src/kvstore.c
Fusl: Not strictly needed but makes this look cleaner.

Fusl:
> There is a leak reported in the address-sanitizer job. Maybe it's fixed by #2288 that I merged yesterday. Can you merge the latest unstable and see if the leak is gone?

My latest force push where I rebased on the unstable branch already includes that commit, not sure what's going on. I'll take a closer look at it.

zuiderkwast:
The address-santitizer job https://github.com/valkey-io/valkey/actions/runs/16039519166/job/45258225409?pr=2095#step:7:437 complains about memory allocated by hashtableCreate. This is after a failed assertion here: https://github.com/valkey-io/valkey/actions/runs/16039519166/job/45258225409?pr=2095#step:7:109

When an asserion fails inside a test case, it is aborted, so we don't free the hash table. I think the leak is not a problem. When we fix the failed assert, this leak should go away.

The unit test framework complains about a leak after each test suite because it only checks that all memory allocated by zmalloc so far has been freed, so these are false positives. It'd be better if it would check it per test suite.

zuiderkwast:
I hope this PR can prevent a memory leak from being reporting multiple times:

* https://github.com/valkey-io/valkey/pull/2304

Fusl:
> This is after a failed assertion here: https://github.com/valkey-io/valkey/actions/runs/16039519166/job/45258225409?pr=2095#step:7:109

Last commit should fix that test case as well now.

src/unit/test_hashtable.c
```diff
@@ -803,6 +803,8 @@ int test_random_entry_with_long_chain(int argc, char **argv, int flags) {
     UNUSED(argv);
     UNUSED(flags);
 
+    hashtableSetResizePolicy(HASHTABLE_RESIZE_FORBID);
```
zuiderkwast: How will this test case work if it's not possible to resize the table? It's always possible to create the first table, which will be just one bucket, but then it can't resize for the rest of this test case, so all elements will be in one bucket chain...?

Sure, it will give you a long bucket chain, bit it's not exactly the idea of this test case. The idea is that you have a skewed hashtable where many entries just ended up in the same bucket chain and others are almost empty.

@SoftlyRaining You wrote this test test originally. Would you like to help out?

Fusl: This is to satisfy the `TEST_ASSERT(!hashtableIsRehashing(ht));` statement that follows right after setting the policy back to `HASHTABLE_RESIZE_ALLOW` since the `hashtableAdd` that happens all through this test otherwise immediately triggers a resize.

Fusl: Ah, disregard this. Looks like my [remove recursive hashtableRightsizeIfNeeded call](https://github.com/valkey-io/valkey/pull/2095/commits/8e6855f4004db75140a78cf5164502ba018e23b6) commit fixes this test case assert as well.

tests/unit/expire.tcl
```diff
@@ -1037,10 +1037,16 @@ start_cluster 1 0 {tags {"expire external:skip cluster"}} {
...
+        # put some data into slot 12182 and trigger the resize by deleting it to trigger shrink
+        # we put just barely enough data in to trigger the resize, not the full 1000 keys
+        # this triggers kicking off resizing and rehashing since we put the hash table
+        # into a bad state by disabling it with setting the dict-resizing flag and don't
+        # want to wait for databasesCron to kick off another resize.
+        # TODO: this is good enough for now but we should improve the logic of this test case
+        for {set j 1} {$j <= 100} {incr j} {
+            r psetex "{foo}0" 500 a
+            r del "{foo}0"
+        }
```
zuiderkwast: I investigated some last night and today. Without this loop, the test case passes but it takes 1-2 minutes instead of less than one second. I added some printouts in each iteration of this loop, e.g.

    puts [r debug htstats 0]
    puts "Num rehashing tables: [dict get [r memory stats] db.dict.rehashing.count]"

I can see that some hashtable is already in progress of growing when resizing is disabled using `debug dict-resizing 0`. When dict-resizing is enabled again, I would expect active rehashing to complete. For some reason it doesn't.

I tried modifying the numbers `threshold_us` in databasesCron, i.e. how long the active rehashing can run, but it doesn't seem to help. The rehashing is not completed even if I let it run for 100 times longer. For some reason, active rehashing doesn't run.

Then I tried changing `hz` from 10 to 100. With this change, the test case completes quickly even without this psetex+del loop.

What happens is that kvstoreTryResizeHashtables finally kicks in and shrinks the hashtable that was already in progress growing. Then, the rehashing to grow is fast-forwarded in hashtable's resize() function. After this, everything completes quickly.

That it takes so long for this to happen is that CRON_DICTS_PER_DB = 16, but we have 16384 hashtables to check, so with hz 10 it takes 1024 cron cycles to check all hash tables if they need resizing or not. With hz 10, this is 102 seconds.

Another change that makes the test case complete quickly is a change to hashtable's rehashStep() function. In the end of the function where we check if rehashing is completed, I added a fast path to end rehashing if `ht->used[0] == 0 && ht->child_buckets[0] == 0`. This change alone (and removing the psetex+del loop) also completes the test cases in < 1 second. (We should probably make this change, but in another PR. We have this condition when resizing starts, but the new check in rehashStep can end the rehashing if the table becomes empty during incremental rehashing.)

What still puzzles me is that incremental rehashing doesn't complete quickly on its own. When it actually runs, it's very fast. It might be another bug that we may want to solve, so I'm digging a little more now before I give up.

zuiderkwast: Solved. Rehashing was ongoing when we tried to trigger shrinking using psetex + del, but the ongoing rehashing prevented the shrinking. Soon after, the active rehashing completes. After this, nothing triggers shrinking except the databaseCron which takes a long time to check all 16384 per-slot hashtables.

A solution is to wait for rehashing to complete, before we use psetex + del to trigger shrinking. This kicks of another incremental rehashing which is fast. This is quite explicit and clear IMHO.

```suggestion
        # Wait for ongoing rehashing to complete, if any
        wait_for_condition 100 50 {
            [dict get [r memory stats] db.dict.rehashing.count] == 0
        } else {
            fail "Active rehashing didn't finish"
        }

        # put some data into slot 12182 and trigger the resize
        # by deleting it to trigger shrink
        r psetex "{foo}0" 500 a
        r del "{foo}0"
```

Fusl: Thank you for investigating and fixing this!

PR #2096, opened 2025-05-17
Further improve performance of sdssplitargs

Fusl:
This reduces the amount of `s_realloc` calls we do for multi-argument inline commands by pre-allocating enough for 8 arguments and doubling the capacity when growing instead of increasing the size linearly:

image

### DIFF
--- a/src/sds.c
+++ b/src/sds.c
@@ -1143,7 +1143,8 @@ static int sdsparsearg(const char *arg, unsigned int *len, char *dst) {
  */
 sds *sdssplitargs(const char *line, int *argc) {
     const char *p = line;
-    char **vector = NULL;
+    size_t cap = 0;
+    sds *vector = NULL;
 
     *argc = 0;
     while (*p) {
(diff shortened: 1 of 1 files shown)


src/sds.c
```diff
@@ -37,6 +37,7 @@
 #include "serverassert.h"
 #include "sds.h"
 #include "sdsalloc.h"
+#include "server.h"
```
zuiderkwast: Please don't. We don't want this low-level library to depend on the whole server.

(Note that sds is also used in valkey-cli which is not server. For example.)

Is there anything specific from server.h you need?

zuiderkwast (commented):
The optimization of sdssplitargs LGTM. I guess we always free the result array afterwards so the increased memory usage will not be an issue.

But why do you use inline commands instead of RESP? I thought it's kept only for historical reasons. Is it faster or anything?

Afaict, sdssplitargs is also used in valkey-cli and for parsing config files. It doesn't seem to be used in any other hot code path.

src/sds.c
Fusl: Only the `max()` function.

zuiderkwast: We could move max to `util.h`, or otherwise just use `? :`.

Fusl:
> why do you use inline commands instead of RESP?

I often times pass the `monitor` of multiple servers into a single other server without wanting to go through the process of having to parse inline protocol data and serializing it as RESP data to send to Valkey for processing.

src/sds.c
Fusl: Removed `server.h` and replaced it with a ternary operator instead.

zuiderkwast (approved):

PR #2098, opened 2025-05-18
Allow mixing quoted and unquoted inline args

Fusl:
Fix an inconsistency of `sdssplitargs` where `foo""` is allowed but `""foo` is rejected.

This patch loosens up the parsing to allow mixing quoted and unquoted text within the same argument, similar to how shell syntax works, allowing the use of `"foo"bar'baz'zzz` to produce a single argument of `"foobarbazzzz"`.

This is a backward compatible change. It affects the inline protocol, the arguments to valkey-cli and the parsing of config files.

### DIFF
--- a/tests/unit/protocol.tcl
+++ b/tests/unit/protocol.tcl
@@ -69,6 +69,22 @@ start_server {tags {"protocol network"}} {
         assert_equal "test-value" [r read]
     }
 
+    test "Unbalanced single quotes" {
+        reconnect
+        r write "set foo 'b'a'r\r\n"
+        r write "ping\r\n"
+        r flush
+        assert_error "*unbalanced*" {r read}
+    }
+
+    test "Unbalanced double quotes" {
+        reconnect
+        r write "set foo \"b\"a\"r\r\n"
+        r write "ping\r\n"
+        r flush
+        assert_error "*unbalanced*" {r read}
+    }
+
     set c 0
     foreach seq [list "\x00" "*\x00" "$\x00" "*1\r\n$\x00"] {
         incr c
--- a/tests/unit/protocol.tcl
+++ b/tests/unit/protocol.tcl
@@ -60,12 +60,13 @@ start_server {tags {"protocol network"}} {
         assert_error "*wrong*arguments*ping*" {r ping x y z}
     }
 
-    test "Unbalanced number of quotes" {
+    test "Mixing quoted and unquoted strings" {
         reconnect
-        r write "set \"\"\"test-key\"\"\" test-value\r\n"
-        r write "ping\r\n"
+        r write "set \"\"\"tes\"t-'k'e\"y\" \"test\"'-'value\r\n"
+        r write "get test'-'key\r\n"
         r flush
-        assert_error "*unbalanced*" {r read}
+        assert_equal "OK" [r read]
+        assert_equal "test-value" [r read]
     }
 
     set c 0
--- a/src/sds.c
+++ b/src/sds.c
@@ -1068,10 +1068,7 @@ static int sdsparsearg(const char *arg, unsigned int *len, char *dst) {
                 default: new_char = *p; break;
                 }
             } else if (*p == '"') {
-                /* closing quote must be followed by a space or
-                 * nothing at all. */
-                if (*(p + 1) && !isspace(*(p + 1))) return 0;
-                done = 1;
+                inq = 0;
             } else if (!*p) {
                 /* unterminated quotes */
                 return 0;
@@ -1083,10 +1080,7 @@ static int sdsparsearg(const char *arg, unsigned int *len, char *dst) {
                 p++;
                 new_char = *p;
             } else if (*p == '\'') {
-                /* closing quote must be followed by a space or
-                 * nothing at all. */
-                if (*(p + 1) && !isspace(*(p + 1))) return 0;
-                done = 1;
+                insq = 0;
             } else if (!*p) {
                 /* unterminated quotes */
                 return 0;
--- a/src/unit/test_sds.c
+++ b/src/unit/test_sds.c
@@ -380,10 +380,10 @@ int test_sdssplitargs(int argc, char **argv, int flags) {
     TEST_ASSERT(!strcmp("", sargv[2]));
     sdsfreesplitres(sargv, len);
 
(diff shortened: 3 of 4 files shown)


zuiderkwast:
Why are `foo""` and `"foo"'bar'` incorrect? They're accepted by shell syntax:

```sh
$ echo foo""
foo
$ echo "foo"'bar'
foobar
$ echo "'x'"='"y"'
'x'="y"
```

Maybe we break some user code if we forbid this. Does it have any performance impact?

Fusl:
> Why are `foo""` and `"foo"'bar'` incorrect? They're accepted by shell syntax:

These were never considerd valid in Valker or Redis:

```
"x"y
"x""y"
"x"'y'
x"y"z
```

`sdssplitargs` in Redis always had a check in place that ensures that a closing `"` and `'` are never followed by anything else but a whitespace character or nothing at all, therefore it is safe to assume that single- or double-quotes immediately following an unquoted text without whitespace char should also be considered invalid in this case.

> Maybe we break some user code if we forbid this. Does it have any performance impact?

No performance impact, this is mostly just to address the inconsistent behavior. Alternatively, we could also just turn it around and always allow single- or double-quotes mixed inside an unquoted argument such as my example cases above, that would not break any existing user code and we still address the inconsistent behavior.

madolson:
The tests do seem to be detecting a regression.

Fusl:
The test that's failing here is the
```
# invalid quotation
assert_equal "Invalid argument(s)" [run_command $fd "get \"\"key"]
assert_equal "Invalid argument(s)" [run_command $fd "get \"key\"x"]

# quotes after the argument are weird, but should be allowed
assert_equal "OK" [run_command $fd "set key\"\" bar"]  <----------
assert_equal "bar" [r get key]
```
where this patch will reject such command line strings in the future.

I can easily swap the test out to expect an `Invalid argument(s)` here as well instead but the question is whether we want to:
* reject such strings (what this PR does), or
* loosen up the parsing to allow mixing quoted and unquoted text within the same argument (similar to how shell syntax works, allowing the use of `"foo"bar'baz'zzz` to produce a single argument of `"foobarbazzzz"`), or
* just not change the existing behavior at all and ignore the inconsistency in parsing such command lines?

zuiderkwast:
> * reject such strings (what this PR does), or
> * loosen up the parsing to allow mixing quoted and unquoted text within the same argument (similar to how shell syntax works, allowing the use of `"foo"bar'baz'zzz` to produce a single argument of `"foobarbazzzz"`), or
> * just not change the existing behavior at all and ignore the inconsistency in parsing such command lines?

We have learnt to be very careful with behavior changes. They can cause trouble to users. If something works in Redis and not in Valkey, it's unfortunate for users who want to migrate.

Option 1 (reject) seems risky to me. Option 2 and 3 seem safe. But is it worth changing it at all? Did this cause you any real trouble or confusion in practice?

Fusl:
> Option 1 (reject) seems risky to me. Option 2 and 3 seem safe. But is it worth changing it at all? Did this cause you any real trouble or confusion in practice?

It didn't cause any trouble but the behavior is definitely very confusing when it works one way (quotes only at end) but not the other (quotes at start or in the middle).

To prevent a possible breaking change, I'll update my PR to instead loosen up the requirements a little to allow for quoted strings anywhere around unquoted strings.

zuiderkwast:
> It didn't cause any trouble but the behavior is definitely very confusing when it works one way (quotes only at end) but not the other (quotes at start or in the middle).

You found it by reading source code, didn't you? ;)

> To prevent a possible breaking change, I'll update my PR to instead loosen up the requirements a little to allow for quoted strings anywhere around unquoted strings.

I'm OK with that if it doesn't add any code complexity. If it removes complexity, better.

Fusl:
> You found it by reading source code, didn't you? ;)

I found it by [re-implementing the logic in Golang](https://github.com/Fusl/go-resp/blob/c4c34ae89024bd5248ccf1a3786fbba6a4c3f031/server.go#L200-L371) and repeatedly being confused about it. In a newer version I implemented a similar change to reject such arguments so I figured I might as well open a PR to fix it in Valkey. Loosening up the requirement works for me as well.

> I'm OK with that if it doesn't add any code complexity. If it removes complexity, better.

From what I can tell, it should remove complexity by getting rid of a handful of bounds checks.

Fusl:
@zuiderkwast This should do.

zuiderkwast (approved):
Yeah, I like the simplification.

It's a behavior change but not a breaking one.

zuiderkwast:
Btw, in future PRs, you can just push multiple commits. It makes it easier to review the changes after the previous review. We squash-merge PRs to one commit anyway.

zuiderkwast:
I updated the top comment to reflect the last change.

There are failing tests though. Rejected syntax that is now accepted. Please update those test cases.

```
*** [err]: Unbalanced number of quotes in tests/unit/protocol.tcl
Expected an error matching '*unbalanced*' but got 'OK' (context: type eval line 6 cmd {assert_error "*unbalanced*" {r read}} proc ::test)
*** [err]: Interactive CLI: Parsing quotes in tests/integration/valkey-cli.tcl
Expected 'Invalid argument(s)' to be equal to '"\tbar\t"' (context: type eval line 12 cmd {assert_equal "Invalid argument(s)" [run_command $fd "get \"\"key"]} proc ::test)
```

tests/unit/protocol.tcl
```diff
@@ -60,12 +60,13 @@ start_server {tags {"protocol network"}} {
         assert_error "*wrong*arguments*ping*" {r ping x y z}
     }
 
-    test "Unbalanced number of quotes" {
+    test "Mixing quoted and unquoted strings" {
```
zuiderkwast, line 63: We could keep a test that give the error for unbalanced quotes, like `"get 'foo"`, just to cover that error.

Fusl: Added some tests back that should cover unbalanced single and double quotes as well as escaped single quotes in the unit tests.

zuiderkwast:
You need `--signoff` in all commits. Click (...) -> Details on the failed DCO check.

Fusl:
> You need `--signoff` in all commits. Click (...) -> Details on the failed DCO check.

My bad, should be correct now and configured `format.signoff` for my locally cloned Valkey repo as well.

PR #2099, opened 2025-05-19
Optimize string2ll with load-time CPU feature check using IFUNC resolver

zhulipeng:
### Problem Statement

As we increasingly incorporate SIMD instructions to optimize performance-critical functions, our current runtime CPU feature detection pattern is introducing significant overhead. For example, in the **string2ll** function, the feature detection wrapper adds approximately 3.35% performance overhead (shown as Fig 1), which is very significant especially considering the AVX-optimized implementation itself is highly effective.

Current pattern:
```c
int string2ll(const char *s, size_t slen, long long *value) {
#if HAVE_X86_SIMD
    if (__builtin_cpu_supports("avx512f") &&
        __builtin_cpu_supports("avx512vl") &&
        __builtin_cpu_supports("avx512bw"))
        return string2llAVX512(s, slen, value);
#endif
    return string2llScalar(s, slen, value);
}
``` 
image

Fig 1. Profiling of [memtier_benchmark-1key-list-10K-elements-lpos-integer](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-list-10K-elements-lpos-integer.yml)

### Proposed Solution

This PR implements the **GNU IFUNC** (indirect function) mechanism, which resolves the optimal function implementation at program load time rather than checking CPU features during each function call. This approach eliminates repeated detection overhead in hot code paths.

#### Benefits

1. **Performance Improvement**: Eliminates feature detection overhead (~3.35% for string2ll)
2. **One-time Detection**: CPU capabilities are checked once at load time
3. **Zero Runtime Overhead**: No conditionals in the call path
4.  **Maintainability**: Cleaner code without conditional branches

### Performance Boost

The example test suite we can observe a ~3% performance improvement. 
|Benchmark|Performance Boost|
|-|-|
|[memtier_benchmark-1key-list-10K-elements-lpos-integer](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-list-10K-elements-lpos-integer.yml)|**3%**|

#### Test Env
- OS: CentOS Stream 9
- Platform: Intel(R) Xeon(R) Platinum 8380 CPU @ 2.30GHz

```
port 9001
bind * -::*
daemonize no
protected-mode no
save
```

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -391,4 +391,12 @@ void setcpuaffinity(const char *cpulist);
 #define ATTRIBUTE_TARGET_AVX512
 #endif
 
+#if defined(__linux__) &&                                                               \
+    ((defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 3))) || \
+     (defined(__clang__) && (__clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 4))))
+#define HAVE_IFUNC 1
+#else
+#define HAVE_IFUNC 0
+#endif
+
 #endif
--- a/src/config.h
+++ b/src/config.h
@@ -391,7 +391,7 @@ void setcpuaffinity(const char *cpulist);
 #define ATTRIBUTE_TARGET_AVX512
 #endif
 
-#if defined(__linux__) && defined(__GNUC__) && __GNUC__ > 4
+#if defined(__linux__) && (defined(__GNUC__) && (__GNUC__ > 4) || defined(__clang__) && (__clang_major__) > 5)
 #define HAVE_IFUNC 1
 #else
 #define HAVE_IFUNC 0
--- a/src/config.h
+++ b/src/config.h
@@ -391,9 +391,7 @@ void setcpuaffinity(const char *cpulist);
 #define ATTRIBUTE_TARGET_AVX512
 #endif
 
-#if defined(__linux__) &&                                                               \
-    ((defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 3))) || \
-     (defined(__clang__) && (__clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 4))))
+#if defined(__linux__) && defined(__GNUC__) && __GNUC__ > 4
 #define HAVE_IFUNC 1
 #else
 #define HAVE_IFUNC 0
--- a/src/util.c
+++ b/src/util.c
@@ -57,7 +57,6 @@
 
 #if HAVE_X86_SIMD
 #include <immintrin.h>
-#include <cpuid.h>
 #endif
 
 #define UNUSED(x) ((void)(x))
@@ -626,55 +625,18 @@ static int string2llScalar(const char *s, size_t slen, long long *value) {
 }
 
 #if HAVE_IFUNC && HAVE_X86_SIMD
-/* This function detects if the CPU supports the AVX-512 instruction set extensions
- * required.
- *
- * Execution timing:
- * - Called ONCE at program load time by the dynamic linker
- * - Executes before main() during the IFUNC resolution phase
- * - Not called during normal program execution, eliminating runtime overhead
- *
- * Note on implementation choice:
- * We use direct CPUID calls instead of __builtin_cpu_supports() because:
- * 1. IFUNC resolvers run before C runtime initialization
- * 2. __builtin_cpu_supports() relies on runtime initialization
- * 3. This approach ensures correct feature detection at load time
- * */
-static int cpu_has_avx512_features(void) {
-    unsigned int eax, ebx, ecx, edx;
-    unsigned int max_level;
-
-    /* Get the maximum supported CPUID level,
-     * checks if the CPU supports CPUID leaf 7,
-     * which is where AVX-512 features are reported. */
-    max_level = __get_cpuid_max(0, NULL);
-    if (max_level < 7) {
-        return 0; /* CPUID leaf 7 not supported, return 0. */
-    }
-
-    /* Call CPUID with leaf 7, subleaf 0 to get advanced feature flags */
-    if (!__get_cpuid_count(7, 0, &eax, &ebx, &ecx, &edx)) {
-        return 0; /* CPUID failed */
-    }
-
-    /* The AVX-512 feature bits are in the EBX register.*/
-    int has_avx512f = (ebx & bit_AVX512F) != 0;
-    int has_avx512vl = (ebx & bit_AVX512VL) != 0;
-    int has_avx512bw = (ebx & bit_AVX512BW) != 0;
-    return has_avx512f && has_avx512vl && has_avx512bw;
-}
-
-static int (*string2ll_ifunc(void))(const char *, size_t, long long *) {
-    if (cpu_has_avx512_features()) {
+static int (*string2ll_resolver(void))(const char *, size_t, long long *) {
+    __builtin_cpu_init();
+    if (__builtin_cpu_supports("avx512f") &&
+        __builtin_cpu_supports("avx512vl") &&
+        __builtin_cpu_supports("avx512bw"))
         return string2llAVX512;
-    }
     return string2llScalar;
 }
 
 int string2ll(const char *s, size_t slen, long long *value)
-    __attribute__((ifunc("string2ll_ifunc")));
+    __attribute__((ifunc("string2ll_resolver")));
 #else
-/* Fallback for platforms without IFUNC support */
 int string2ll(const char *s, size_t slen, long long *value) {
 #if HAVE_X86_SIMD
     if (__builtin_cpu_supports("avx512f") &&
@@ -684,7 +646,7 @@ int string2ll(const char *s, size_t slen, long long *value) {
 #endif
     return string2llScalar(s, slen, value);
 }
-#endif
+#endif /* HAVE_IFUNC && HAVE_X86_SIMD */
 
 /* Helper function to convert a string to an unsigned long long value.
  * The function attempts to use the faster string2ll() function inside
--- a/src/util.c
+++ b/src/util.c
@@ -625,7 +625,7 @@ static int string2llScalar(const char *s, size_t slen, long long *value) {
 }
 
 #if HAVE_IFUNC && HAVE_X86_SIMD
-static int (*string2ll_resolver(void))(const char *, size_t, long long *) {
+__attribute__((no_sanitize_address)) static int (*string2ll_resolver(void))(const char *, size_t, long long *) {
     __builtin_cpu_init();
     if (__builtin_cpu_supports("avx512f") &&
         __builtin_cpu_supports("avx512vl") &&
--- a/src/util.c
+++ b/src/util.c
@@ -626,6 +626,8 @@ static int string2llScalar(const char *s, size_t slen, long long *value) {
 
 #if HAVE_IFUNC && HAVE_X86_SIMD
 __attribute__((no_sanitize_address)) static int (*string2ll_resolver(void))(const char *, size_t, long long *) {
+    /* Ifunc resolvers run before ASan initialization and before CPU detection
+     * is initialized, so disable ASan and init CPU detection here. */
     __builtin_cpu_init();
     if (__builtin_cpu_supports("avx512f") &&
         __builtin_cpu_supports("avx512vl") &&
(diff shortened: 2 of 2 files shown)


src/util.c
```diff
@@ -621,6 +625,56 @@ static int string2llScalar(const char *s, size_t slen, long long *value) {
...
+
+static int (*string2ll_ifunc(void))(const char *, size_t, long long *) {
+    if (cpu_has_avx512_features()) {
+        return string2llAVX512;
+    }
+    return string2llScalar;
+}
+
+int string2ll(const char *s, size_t slen, long long *value)
+    __attribute__((ifunc("string2ll_ifunc")));
```
r-devulap: Have you considered using `__attribute__((constructor))` + ptr approach instead of `ifunc`? I find that usually more portable and easier to debug. It also enables you to continue using the `__builtin_cpu_supports` feature and avoid having to maintain low level cpuid code.

src/util.c
```diff
@@ -621,6 +625,56 @@ static int string2llScalar(const char *s, size_t slen, long long *value) {
...
+#else
+/* Fallback for platforms without IFUNC support */
```
r-devulap: Using a macro to conditionally define the attribute constructor at compile time might also help you get rid of a fallback.

src/util.c
zhulipeng: hi @r-devulap, thanks for your suggestion, the primary reason we opted for GNU IFUNC (Indirect Function) over the `__attribute__((constructor))` with function pointer approach is performance. Our benchmarks showed that calls through function pointers are slower than direct function calls enabled by IFUNC. The ifunc will update PLT to point directly to the chosen implementation, eliminating a level of indirection at runtime while the function pointer still needs to load function pointer from memory to do indirect call:
`callq  *0x2eb78c(%rip)       # 7757c8 <string2ll_impl.lto_priv.0>`

src/util.c
```diff
@@ -621,6 +625,56 @@ static int string2llScalar(const char *s, size_t slen, long long *value) {
...
+
+    /* The AVX-512 feature bits are in the EBX register.*/
+    int has_avx512f = (ebx & bit_AVX512F) != 0;
+    int has_avx512vl = (ebx & bit_AVX512VL) != 0;
+    int has_avx512bw = (ebx & bit_AVX512BW) != 0;
+    return has_avx512f && has_avx512vl && has_avx512bw;
+}
+
+static int (*string2ll_ifunc(void))(const char *, size_t, long long *) {
+    if (cpu_has_avx512_features()) {
```
r-devulap: Can't we call __builtin_cpu_init() within the resolver function to avoid writing the cpuid code? 
```
static int (*func_resolver(void))(...) {
    __builtin_cpu_init(); 
    if (__builtin_cpu_supports("avx512"))
        return func_avx512;
    else
        return func_generic;
}
```

src/util.c
r-devulap: Makes sense. indirect jump does add an overhead, but did it cause a 3% regression?

src/util.c
r-devulap: See https://gcc.gnu.org/onlinedocs/gcc/x86-Built-in-Functions.html

zhulipeng: I am open to using either `cpuid` or `__builtin_cpu_init` /`__builtin_cpu_supports` functions to implement `ifunc`.  :)

src/util.c
zhulipeng: The hotspots of cycles profiling list in top comment inspires that it is worth to introduce the **ifunc**.

r-devulap: Oh sorry, I should have been clear. I don't doubt the benefit of ifunc. I meant to ask does it show up when using `__attribute__((constructor))` too?

src/util.c
r-devulap: If valkey doesn't mind using builtin functions, which I assume it doesn't because it is used in the current code, then I would recommend using the builtin's so that you dont have to maintain additional low level code.

zuiderkwast: If we can use `__builtin_cpu_init` /`__builtin_cpu_supports` in the resolver, then I prefer it. Less code and easier to read and maintain.

Let's name the function `string2ll_resolver`, ok? It seems to be a convention to name them like this.

src/util.c
zhulipeng: Slightly improve performance but less than 2% for this case.

zuiderkwast (approved):
LGTM, thanks, also @r-devulap for review!

src/config.h
```diff
@@ -391,4 +391,12 @@ void setcpuaffinity(const char *cpulist);
 #define ATTRIBUTE_TARGET_AVX512
 #endif
 
+#if defined(__linux__) &&                                                               \
+    ((defined(__GNUC__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 3))) || \
+     (defined(__clang__) && (__clang_major__ > 3 || (__clang_major__ == 3 && __clang_minor__ >= 4))))
```
r-devulap: Do we expect valkey to be compiled with gcc4 and clang3 which are more than a decade old? If not, perhaps we can keep this simple: (clang also defines __GNUC__)

```
#if defined(__linux__) && (defined(__GNUC__) 
```

r-devulap (approved):
Minor comment, feel free to ignore. Otherwise LGTM.

src/config.h
zuiderkwast: Better safe. We can ignore 4.x to make the check simpler.

    #if defined(__linux__) && defined(__GNUC__) && __GNUC__ > 4

r-devulap: Clang defines `__GNUC__` as 4 regardless of the actual Clang version, so this check fails with clang.

zuiderkwast: So we need GNUC > 4 or clang_major > 3 then :)

Btw, if a macro isn't defined, for expample `__clang_major__`, don't you get a warning for `defined(__clang_major) && __clang_major__ > 3` when compiling with `-Wundef`?

r-devulap: > So we need GNUC > 4 or clang_major > 3 then :)

yeah. But they are so old though :) 

> Btw, if a macro isn't defined, for expample __clang_major__, don't you get a warning for defined(__clang_major) && __clang_major__ > 3 when compiling with -Wundef?

Looks like, not if proceeded by `#if defined`. This seems to be build with `gcc main.c -Wundef`: 

```
#include <stdio.h>
int main(void) {
#if defined(__clang_major) && __clang_major__ > 3
    printf("The value of __clang__major__ is: %d\n", __clang_major__);
#endif
    return 0;
}
```
This one throws a warning: 

```
#include <stdio.h>
int main(void) {
#if __clang_major__ > 3
    printf("The value of __clang__major__ is: %d\n", __clang_major__);
#endif
    return 0;
}
```

zuiderkwast: Interesting. Thanks, I learned something.

zhulipeng:
Seems the crash is caused by the ASan with these built-in cpu functions, particularly `__builtin_cpu_init`, might allocate data on the stack or call other functions that involve stack usage. When called before ASan's full initialization, these stack operations can trigger ASan errors.

A workaround described in https://github.com/google/sanitizers/issues/342 is to add the` __attribute__((no_sanitize_address)) ` to the resolver function.
Or back to the `cpuid` solution?

```c
#if HAVE_IFUNC && HAVE_X86_SIMD
__attribute__((no_sanitize_address))
static int (*string2ll_resolver(void))(const char *, size_t, long long *) {
    __builtin_cpu_init();
    if (__builtin_cpu_supports("avx512f") &&
        __builtin_cpu_supports("avx512vl") &&
        __builtin_cpu_supports("avx512bw"))
        return string2llAVX512;
    return string2llScalar;
}
```

Reproduce:

```bash
make -j SANITIZER=address # build server with ASan
```
``` bash
gdb --args ~/valkey/src/valkey-server ~/valkey.conf

Program received signal SIGSEGV, Segmentation fault.
0x0000000000528d5b in string2ll_resolver () at ~/valkey/src/util.c:630
630         if (__builtin_cpu_supports("avx512f") &&
(gdb) bt
#0  0x0000000000528d5b in string2ll_resolver () at ~/valkey/src/util.c:630
#1  0x00007ffff7fd5de0 in elf_machine_lazy_rel (skip_ifunc=0, reloc=0x450cf8, l_addr=<optimized out>, scope=0x7ffff7ffe580, map=0x7ffff7ffe210)
    at ../sysdeps/x86_64/dl-machine.h:591
#2  elf_dynamic_do_Rela (skip_ifunc=<optimized out>, lazy=<optimized out>, nrelative=<optimized out>, relsize=<optimized out>, reladdr=<optimized out>,
    scope=<optimized out>, map=0x7ffff7ffe210) at /usr/src/debug/glibc-2.34-180.el9.x86_64/elf/do-rel.h:79
#3  _dl_relocate_object (l=l@entry=0x7ffff7ffe210, scope=<optimized out>, reloc_mode=<optimized out>, consider_profiling=<optimized out>, consider_profiling@entry=0)
    at dl-reloc.c:288
#4  0x00007ffff7fe8089 in dl_main (phdr=<optimized out>, phnum=<optimized out>, user_entry=<optimized out>, auxv=<optimized out>) at rtld.c:2461
#5  0x00007ffff7fe3de3 in _dl_sysdep_start (start_argptr=start_argptr@entry=0x7fffffffde70, dl_main=dl_main@entry=0x7ffff7fe5c60 <dl_main>)
    at ../sysdeps/unix/sysv/linux/dl-sysdep.c:140
#6  0x00007ffff7fe5738 in _dl_start_final (arg=0x7fffffffde70) at rtld.c:502
#7  _dl_start (arg=0x7fffffffde70) at rtld.c:587
#8  0x00007ffff7fe4808 in _start () from /lib64/ld-linux-x86-64.so.2

```

src/config.h
zhulipeng: FYI, the `__builtin_cpu_init()` function was introduced in Clang with version 6.0, released in 2018. and in GCC with version 4.8, released in 2013.

src/util.c
```diff
@@ -621,6 +624,19 @@ static int string2llScalar(const char *s, size_t slen, long long *value) {
     return 1;
 }
 
+#if HAVE_IFUNC && HAVE_X86_SIMD
+__attribute__((no_sanitize_address)) static int (*string2ll_resolver(void))(const char *, size_t, long long *) {
+    __builtin_cpu_init();
```
zuiderkwast, line 631: Should we add a comment to explain why we need this? Just an idea.

```suggestion
    /* Ifunc resolvers run before ASan initialization and before CPU detection
     * is initialized, so disable ASan and init CPU detection here. */
    __builtin_cpu_init();
```

zuiderkwast (approved):
Good idea to disable ASan.

Good catch about clang 6 for cpu init. It's OK, it's already very old (2018).

Or do you actually like the cpuid solution more? You can try to convince me. :laughing: It's more code but less dependencies...

zhulipeng:
> Or do you actually like the cpuid solution more? You can try to convince me. 😆 It's more code but less dependencies

IMO, `cpuid` is simpler. Using the `cpuid` instruction directly provides more immediate access and full control over the detection logic compared to the `__builtin_cpu_init` and `__builtin_cpu_supports` functions.

While I understand your preference for the __builtin approach 😄, there's a clear trade-off here. The __builtin functions nicely abstract away low-level implementation details, resulting in cleaner, more portable code. However, this abstraction layer may occasionally introduce unexpected behavior due to internal optimizations or compiler-specific implementations, such as conflicts with ASan.

zuiderkwast:
> While I understand your preference for the __builtin approach 😄, there's a clear trade-off here. The __builtin functions nicely abstract away low-level implementation details, resulting in cleaner, more portable code. However, this abstraction layer may occasionally introduce unexpected behavior due to internal optimizations or compiler-specific implementations, such as conflicts with ASan.

Yes, it's a tradeoff.

But we use __builtin in the fallback code that runs without ifunc. It's good that these two checks for AVX use the same style (with and without ifunc). Do you want to use cpuid here too?

We have solved ASan and now. Do you worry about another problem, like another sanitizer or valgrind?

zhulipeng:
@zuiderkwast @r-devulap Thank you both for your efforts and suggestions on this matter.
@r-devulap Glad to see you in the Valkey community! 🙂

roshkhatri:
This commit would break alpine build for valkey as Alpine Linux uses `musl-libc` instead of GLIBC, and musl-libc does not implement the IFUNC mechanism. just saw the failure in a PR for valkey-conainter as mentioned by [alaviss](https://github.com/alaviss) here: https://github.com/valkey-io/valkey-container/pull/70

PR #2100, opened 2025-05-19
Parse all commands in I/O threads for effective memory prefetching

PingXie:
Previously, memory prefetching (introduced in Valkey 8.0) was primarily effective only for the first command in a client's query buffer. This limitation arose because I/O threads would parse and prepare only that initial complete command.

This commit refactors the command parsing and processing pipeline to extend the benefits of prefetching to all commands within a client's query buffer. This is achieved by enabling I/O threads to parse the entirety of the buffer and enqueue each command.

Key changes include:

- Introduction of a `parsedCommandQueue` within the client struct. Each `parsedCommand` in this queue stores its arguments, the looked-up server command, arity validation status, and pre-calculated slot information (for cluster mode). The queue uses capacity to track allocated size.
- New static functions `enqueueParsedCommand`, `dequeueParsedCommand`, and `resetParsedCommandQueue` are added to manage this queue.
- `enqueueParsedCommand` handles command lookup, arity checks, and (for the first key in cluster mode) slot calculation, offloading these tasks from the main thread to the I/O thread during parsing. Client's active argv/argc are cleared after enqueuing.
- `processInputBuffer` is modified: when run by an I/O thread, it now loops through the client's query buffer, parsing all available commands via `processInlineBuffer` or `processMultibulkBuffer`, which then call `enqueueParsedCommand`.
- `processCommandAndResetClient` (on the main thread) now dequeues and executes commands one by one from this queue until it's empty or the client is blocked.
- Blocked clients have their current command prepended back to the queue via `enqueueParsedCommand(c, false)` for later execution.
- Memory prefetching logic in memory_prefetch.c is updated to iterate over the `parsedCommandQueue`, enabling prefetching for arguments and keys of all queued commands.
- Cluster operations in `getNodeByQuery` now leverages the pre-calculated slot (c->slot, populated from parsed_cmd->slot during dequeue) for a command's first key, avoiding redundant hash slot calculations.
- Client lifecycle functions (`createClient`, `freeClientArgv`, `freeClient`) are updated to initialize and clean up the new `parsed_cmd_queue`.
- Supporting changes in valkey-benchmark.c were made to align with key generation and slot targeting for cluster mode testing.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -1070,7 +1070,12 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int 
 
         for (j = 0; j < numkeys; j++) {
             robj *thiskey = margv[keyindex[j].pos];
-            int thisslot = keyHashSlot((char *)thiskey->ptr, sdslen(thiskey->ptr));
+            /* Use the cached slot to avoid the expensive hash slot calcluation for the first key.
+             * Note that slot is not computed for the first key when the io-threads are inactive. */
+            int thisslot = (j == 0) ? c->slot : -1;
+
+            /* Compute the hash slot for this key if its slot is not computed yet. */
+            if (thisslot < 0) thisslot = (int) keyHashSlot((char *)thiskey->ptr, sdslen(thiskey->ptr));
 
             if (firstkey == NULL) {
                 /* This is the first key we see. Check what is the slot
--- a/src/networking.c
+++ b/src/networking.c
@@ -187,6 +187,177 @@ static inline int isReplicaReadyForReplData(client *replica) {
            !(replica->flag.close_asap);
 }
 
+/*
+ * Add a recently parsed command to the client's parsed command queue.
+ *
+ * This helps reduce the burden on the main thread by allowing the I/O thread to
+ * parse all the commands and store them in a queue. The new command can be added
+ * to the end of the queue or, alternatively, to the beginning.
+ *
+ * The queue is automatically enlarged if it is full and a command is being
+ * added to the end.
+ *
+ * Parameters:
+ *
+ * - c: The client structure whose `parsed_cmd_queue` will be modified.
+ *      After the command information (argc, argv, etc.) is copied,
+ *      these fields on `c` are cleared.
+ *
+ * - append_to_tail: If true, the command is directed to the end (tail) of the queue.
+ *                   If false, it is directed to the start (head). When adding
+ *                   to the head, it is required that `c->parsed_cmd_queue.head` be > 0.
+ *
+ * Notes:
+ * - The `c->argv` pointer (and its associated data) is moved into the queue.
+ * - If cluster mode is active, the first key's slot is computed within this function.
+ */
+static void enqueueParsedCommand(client *c, bool append_to_tail) {
+    /* If the tail is at 0, this indicate an empty queue.
+     * Initialization with a starting size is performed if this is the case. */
+    if (c->parsed_cmd_queue.tail == 0) {
+        serverAssert(c->parsed_cmd_queue.head == 0);
+        zfree(c->parsed_cmd_queue.parsed_cmds);
+        c->parsed_cmd_queue.capacity = 16;
+        c->parsed_cmd_queue.parsed_cmds = zmalloc(sizeof(parsedCommand) * c->parsed_cmd_queue.capacity);
+    }
+
+    /* If space at the tail is nearly exhausted, the queue is enlarged.
+     * This resizing logic is triggered only when adding to the tail. */
+    if (append_to_tail && c->parsed_cmd_queue.capacity == c->parsed_cmd_queue.tail) {
+        /* Double the size of the parsed command queue if it won't go beyond 256;
+         * otherwise, just add another 32 entries. */
+        if (c->parsed_cmd_queue.capacity > 256) {
+            c->parsed_cmd_queue.capacity += 32;
+        } else {
+            c->parsed_cmd_queue.capacity *= 2;
+        }
+        serverAssert(c->parsed_cmd_queue.capacity > 0);
+        c->parsed_cmd_queue.parsed_cmds =
+            zrealloc(c->parsed_cmd_queue.parsed_cmds,
+                     sizeof(parsedCommand) * c->parsed_cmd_queue.capacity);
+    }
+
+    serverAssert(append_to_tail || c->parsed_cmd_queue.head > 0);
+    parsedCommand *parsed_cmd;
+    if (append_to_tail) {
+        parsed_cmd = &c->parsed_cmd_queue.parsed_cmds[c->parsed_cmd_queue.tail++];
+    } else {
+        parsed_cmd = &c->parsed_cmd_queue.parsed_cmds[--c->parsed_cmd_queue.head];
+    }
+
+    /* Copy the command information to the parsed command structure. */
+    parsed_cmd->argc = c->argc;
+    parsed_cmd->argv = c->argv;
+    parsed_cmd->argv_len = c->argv_len;
+    parsed_cmd->argv_len_sum = c->argv_len_sum;
+
+    parsed_cmd->server_cmd = lookupCommand(c->argv, c->argc);
+    parsed_cmd->valid_arity = parsed_cmd->server_cmd && commandCheckArity(parsed_cmd->server_cmd, c->argc, NULL);
+    parsed_cmd->slot = -1;
+    parsed_cmd->duration = c->duration;
+
+    /* Offload slot calculations to the I/O thread to reduce main-thread load. */
+    if (parsed_cmd->valid_arity && server.cluster_enabled) {
+        getKeysResult result;
+        initGetKeysResult(&result);
+        int numkeys = getKeysFromCommand(parsed_cmd->server_cmd, parsed_cmd->argv, parsed_cmd->argc, &result);
+        if (numkeys) {
+            robj *first_key = parsed_cmd->argv[result.keys[0].pos];
+            parsed_cmd->slot = keyHashSlot(first_key->ptr, sdslen(first_key->ptr));
+        }
+        getKeysFreeResult(&result);
+    }
+
+    /* Reset the client's command information. */
+    c->argc = 0;
+    c->argv = NULL;
+    c->argv_len = 0;
+    c->argv_len_sum = 0;
+    c->slot = -1;
+    c->duration = 0;
+}
+
+/*
+ * Retrieve a command the front of the client's parsed command queue.
+ *
+ * When a command is available (meaning the queue's head is before its tail),
+ * it is taken from the head. The command's details, such as arguments and slot
+ * information, are then loaded into the client's active command fields.
+ * The queue's head pointer is advanced as part of this operation.
+ *
+ * Parameters:
+ * - c: The client structure from which a command is to be dequeued.
+ *      The client's active command fields (like `c->argc`, `c->argv`)
+ *      will be populated with data from the dequeued command.
+ *
+ * Returns:
+ * - `true` if a command was successfully taken from the queue.
+ * - `false` if the queue was found to be empty.
+ *
+ * Notes:
+ * - Ownership of the `argv` array and its contents is transferred from
+ *   the queue to the client's active fields.
+ * - To reflect this transfer, the `argv` pointer in the now-empty queue
+ *   slot is set to NULL.
+ */
+static bool dequeueParsedCommand(client *c) {
+    if (c->parsed_cmd_queue.head >= c->parsed_cmd_queue.tail) return false;
+
+    parsedCommand *parsed_cmd = &c->parsed_cmd_queue.parsed_cmds[c->parsed_cmd_queue.head++];
+
+    /* Command details are copied from the queue slot to the client's active fields. */
+    if (parsed_cmd->valid_arity) c->io_parsed_cmd = parsed_cmd->server_cmd;
+    c->argc = parsed_cmd->argc;
+    c->argv = parsed_cmd->argv;
+    c->argv_len = parsed_cmd->argv_len;
+    c->argv_len_sum = parsed_cmd->argv_len_sum;
+    c->slot = parsed_cmd->slot;
+    c->duration = parsed_cmd->duration;
+
+    /* Zero out the queued command's argv pointer and its length
+     * to indicate that the command has been moved out of the queue. */
+    parsed_cmd->argv = NULL;
+    parsed_cmd->argc = 0;
+
+    return true;
+}
+
+/*
+ * Reset the client's parsed command queue
+ *
+ * Commands residing in the queue, specifically those from index 0 up to
+ * the current `tail` position, are processed. For each such command, its
+ * associated arguments (`argv`) are freed. An attempt is made to offload
+ * this deallocation to I/O threads; if this is not successful, the
+ * deallocation occurs synchronously. This involves decrementing reference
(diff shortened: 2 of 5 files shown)


PingXie:
### Test setup

* Google Cloud: us-central1-a
* server: c4-highmem-8
* client: c4-standard-48 (client shouldn't be the bottlneck)

### Test scripts

* Cache population (no cache miss): 
`./valkey-benchmark -h <ip> -c 96 --threads 96 -d 2048 -t set -n 1000000 -r 1000000 --sequential`
* GET w/o pipelining: 
`./valkey-benchmark -h <ip> -c 48 --threads 48 -d 2048 -t get -n 100000000 -r 1000000 --csv -P 1`
* GET w/ pipeline of 10: 
`./valkey-benchmark -h <ip> -c 48 --threads 48 -d 2048 -t get -n 100000000 -r 1000000 --csv -P 10`
* MGET (10 keys):
`./valkey-benchmark -h <ip> -c 48 --threads 48 -d 2048 -t mget -n 10000000 -r 1000000 --csv -P 1`

### Test results

#### io-threads = 4
image

#### io-threads = 1
image

### Raw test data
setup | mode | io-threads | test | kps | rps | avg_latency_ms | min_latency_ms | p50_latency_ms | p95_latency_ms | p99_latency_ms | max_latency_ms
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
baseline | standalone | 4 | GET (no-pipeline) | 433744 | 433743.53 | 0.101 | 0.032 | 0.103 | 0.127 | 0.143 | 12.095
baseline | standalone | 4 | GET (pipeline 10) | 714566 | 714566.38 | 0.652 | 0.048 | 0.663 | 0.727 | 0.839 | 4.719
baseline | standalone | 4 | MGET (10 keys) | 1249001 | 124900.08 | 0.366 | 0.048 | 0.367 | 0.415 | 0.519 | 12.375
baseline | standalone | 1 | GET (no-pipeline) | 225768 | 225767.77 | 0.2 | 0.032 | 0.207 | 0.255 | 0.271 | 30.735
baseline | standalone | 1 | GET (pipeline 10) | 550737 | 550736.62 | 0.849 | 0.064 | 0.727 | 1.151 | 1.287 | 2.527
baseline | standalone | 1 | MGET (10 keys) | 629909 | 62990.93 | 0.74 | 0.056 | 0.647 | 0.991 | 1.135 | 3.007
optimized | standalone | 4 | GET (no-pipeline) | 429551 | 429551.44 | 0.101 | 0.032 | 0.103 | 0.127 | 0.143 | 12.079
optimized | standalone | 4 | GET (pipeline 10) | 1068479 | 1068478.75 | 0.429 | 0.056 | 0.431 | 0.503 | 0.615 | 11.999
optimized | standalone | 4 | MGET (10 keys) | 1241234 | 124123.38 | 0.367 | 0.048 | 0.367 | 0.423 | 0.527 | 1.487
optimized | standalone | 1 | GET (no-pipeline) | 224751 | 224750.91 | 0.201 | 0.04 | 0.207 | 0.255 | 0.271 | 12.231
optimized | standalone | 1 | GET (pipeline 10) | 558438 | 558437.69 | 0.837 | 0.056 | 0.719 | 1.127 | 1.255 | 2.007
optimized | standalone | 1 | MGET (10 keys) | 635930 | 63593 | 0.733 | 0.056 | 0.639 | 0.983 | 1.111 | 8.463

PingXie:
I have yet to get valkey-benchmark's mget to work without cache misses for a cluster setup so I wrote a small program for that. note that the numbers are not directly comparable to valkey-benchmark but hopefully shed some light on how pipelined GET compares to MGET

### Pipelined GET (this PR)
```
Command         : GET
Threads         : 48
Duration (sec)  : 120
Batch size      : 10
Value size      : 2048 bytes
Total keys      : 120838860
Total requests  : 120838860
Keys per sec    : 1006990
Requests per sec: 1006990
Latency avg     : 473.18 us
Latency p50     : 473 us
Latency p90     : 520 us
Latency p95     : 546 us
Latency p99     : 665 us
```

### Cross-slot MGET (#1986 w/ fixes)
```
Command         : MGET
Threads         : 48
Duration (sec)  : 120
Batch size      : 10
Value size      : 2048 bytes
Total keys      : 133664840
Total requests  : 13366484
Keys per sec    : 1113873
Requests per sec: 111387
Latency avg     : 427.65 us
Latency p50     : 428 us
Latency p90     : 462 us
Latency p95     : 481 us
Latency p99     : 593 us
```

zuiderkwast:
WHAT! You did my duplicate work! Look at mine:

* #2092

I assigned myself to this issue two weeks ago to make sure we don't do duplicate work:

* #2044

zuiderkwast:
> * Cluster operations in `getNodeByQuery` now leverages the pre-calculated slot (c->slot, populated from parsed_cmd->slot during dequeue) for a command's first key, avoiding redundant hash slot calculations.

This issue:

* #2077

We can let the IO thread hash all keys and set c->slot = -2 or something to indicate cross-slot. Then getNodeByQuery doesn't need to hash any key can trust c->slot if it's set earlier.

zuiderkwast:
> WHAT! You did my duplicate work! Look at mine:
> 
> * [Parse multiple commands and prefetch keys #2092](https://github.com/valkey-io/valkey/pull/2092)
> 
> I assigned myself to this issue two weeks ago to be sure we don't do duplicate work:
> 
> * [Parse multiple commands (look ahead) #2044](https://github.com/valkey-io/valkey/issues/2044)

src/networking.c
```diff
@@ -187,6 +187,177 @@ static inline int isReplicaReadyForReplData(client *replica) {
...
+        c->parsed_cmd_queue.parsed_cmds = zmalloc(sizeof(parsedCommand) * c->parsed_cmd_queue.capacity);
+    }
+
+    /* If space at the tail is nearly exhausted, the queue is enlarged.
+     * This resizing logic is triggered only when adding to the tail. */
+    if (append_to_tail && c->parsed_cmd_queue.capacity == c->parsed_cmd_queue.tail) {
+        /* Double the size of the parsed command queue if it won't go beyond 256;
+         * otherwise, just add another 32 entries. */
+        if (c->parsed_cmd_queue.capacity > 256) {
+            c->parsed_cmd_queue.capacity += 32;
```
zuiderkwast, line 230: Is unlimited queue size a problem?

An application that streams an endless stream of commands, can it make the queue grow arbitrarily large? Or is this effectively limited by the query buffer read, so we don't read more from the socket until we have executed the queue?

src/networking.c
```diff
@@ -187,6 +187,177 @@ static inline int isReplicaReadyForReplData(client *replica) {
...
+ * - The `c->argv` pointer (and its associated data) is moved into the queue.
+ * - If cluster mode is active, the first key's slot is computed within this function.
+ */
+static void enqueueParsedCommand(client *c, bool append_to_tail) {
+    /* If the tail is at 0, this indicate an empty queue.
+     * Initialization with a starting size is performed if this is the case. */
+    if (c->parsed_cmd_queue.tail == 0) {
+        serverAssert(c->parsed_cmd_queue.head == 0);
+        zfree(c->parsed_cmd_queue.parsed_cmds);
+        c->parsed_cmd_queue.capacity = 16;
```
zuiderkwast, line 220: In a setup with 10k clients and all of them use pipelines of length 2 or 3 (or a short transaction), then an initial size of 16 wastes some memory? It never shrinks again for the client?

zuiderkwast (commented):
I took a quick look.

Comparison of our PRs:

* Both PRs add a command queue in a very similar way. Almost identical.
* Your PR parses into `c->argc` and then moves the command to the command queue, while my PR parses directly into the command queue.
* Your PR is only for I/O threads (?) while mine is also for single-threaded operation.

Questions:

* How do you handle a pipeline that ends with an incomplete command? Parsing of such command needs to be resumed after the next read from the socket. Do you add the incomplete parse state to the command queue too? (That's what I did in my PR.)
* How do you handle read flags? They are per command, e.g. parse error.
* When debugging my work, I noticed `resetClient()` can't reset `c->reqtype`, `c->multibulklen`, `c->bulklen` anymore, because these fields describe the current location in the command buffer which may be a few commands ahead and be half-way parsing a command in a pipeline, while `resetClient()` runs after each command *execution*. When decoupling command parsing from execution, I needed to move the handling of these query buffer related fields to make the tests pass. I can't see this in your PR. Is that why you have some CI failures?

My PR probably has some other problems. Please take a look! We should merge our efforts. It's actually great that we both digged into this. It makes it much easier to understand and review in depth.

PingXie:
> WHAT! You did my duplicate work! Look at mine:
> 
> * [Parse multiple commands and prefetch keys #2092](https://github.com/valkey-io/valkey/pull/2092)
> 
> I assigned myself to this issue two weeks ago to make sure we don't do duplicate work:
> 
> * [Parse multiple commands (look ahead) #2044](https://github.com/valkey-io/valkey/issues/2044)

lol. looks like a (double) fun problem to solve.

dvkashapov:
I think there may be some memory leaks, I tried recreating similar benchmarking setup on AMD EPYC 7713 32 CPU 50gb RAM: 
cset shield
cpu pin for valkey-server
io-threads 9
same cache population
Results i get:
cache size before GET load: 2.44gb and right after starting GET load theres additional 150mb without new keys.
`./valkey-benchmark -h localhost -c 48 --threads 48 -d 2048 -t get -n 100000000 -r 1000000 -P 1`

Debug output:
```
69597:M 21 May 2025 14:04:35.445 - DB 0: 1000000 keys (0 volatile) in 1835008 slots HT.
69597:M 21 May 2025 14:04:35.445 . Total: 48 clients connected (0 replicas), 2627890496 (2.45G) bytes in use
69597:M 21 May 2025 14:04:40.572 - DB 0: 1000000 keys (0 volatile) in 1835008 slots HT.
69597:M 21 May 2025 14:04:40.573 . Total: 48 clients connected (0 replicas), 2778777576 (2.59G) bytes in use
```
haven't seen same behaviour in #2092:
```
71967:M 21 May 2025 14:16:25.148 - DB 0: 1000000 keys (0 volatile) in 1835008 slots HT.
71967:M 21 May 2025 14:16:25.148 . Total: 48 clients connected (0 replicas), 2619723920 (2.44G) bytes in use
71967:M 21 May 2025 14:16:30.179 - DB 0: 1000000 keys (0 volatile) in 1835008 slots HT.
71967:M 21 May 2025 14:16:30.180 . Total: 48 clients connected (0 replicas), 2619744424 (2.44G) bytes in use
71967:M 21 May 2025 14:16:35.211 - DB 0: 1000000 keys (0 volatile) in 1835008 slots HT.
71967:M 21 May 2025 14:16:35.211 . Total: 48 clients connected (0 replicas), 2619753520 (2.44G) bytes in use
```

src/networking.c
zuiderkwast: I got the answer. It's freed after consuming all the elements.

src/networking.c
zuiderkwast: I got answer. It's effectively limited by the size of the query buffer.

PingXie:
> How do you handle a pipeline that ends with an incomplete command? Parsing of such command needs to be resumed after the next read from the socket. Do you add the incomplete parse state to the command queue too? (That's what I did in my PR.)

incomplete commands are left in the query buffer just like what we are doing today.

> How do you handle read flags? They are per command, e.g. parse error.

I haven't handled read flags today. After thinking it some more, I don't see the need to handle auth_required. I will go over other errors next.

> When debugging my work, I noticed resetClient() can't reset c->reqtype, c->multibulklen, c->bulklen anymore, because these fields describe the current location in the command buffer which may be a few commands ahead and be half-way parsing a command in a pipeline, while resetClient() runs after each command execution. When decoupling command parsing from execution, I needed to move the handling of these query buffer related fields to make the tests pass. I can't see this in your PR. Is that why you have some CI failures?

my changes don't have a hard dependency on the query buffer parsing states, conceptually speaking, since only fully parsed commands are cached in the queue. I will look into the CI failures next.

PingXie:
> I think there may be some memory leaks, I tried recreating similar benchmarking setup on AMD EPYC 7713 32 CPU 50gb RAM: cset shield cpu pin for valkey-server io-threads 9 same cache population Results i get: cache size before GET load: 2.44gb and right after starting GET load theres additional 150mb without new keys. `./valkey-benchmark -h localhost -c 48 --threads 48 -d 2048 -t get -n 100000000 -r 1000000 -P 1`

Thanks @JoBeR007! I wonder if the leak is due to parsed commands getting freed on the new command arrivals and the reason for that is to avoid cross-threads deallocation, which will lead to memory fragmentation. Or there could still be excessive memory fragmentations in this PR? I will take a closer look next. 

Curious, memory leaks aside, what's your perf results look like? I noticed you reported [2x](https://github.com/valkey-io/valkey/pull/2092#pullrequestreview-2857688565) in #2092

src/networking.c
```diff
@@ -3266,7 +3453,10 @@ int processInputBuffer(client *c) {
             resetSharedQueryBuf(c);
         }
 
-        /* We are finally ready to execute the command. */
+        /* If we are in the IO thread, we continue to parse the next command in querybuf. */
+        if (!inMainThread()) continue;
```
zuiderkwast, line 3457: This function "processInputBuffer" isn't currently running in IO threads.

Having to check which thread we are in seems to hint that the code is not very well organized. This is nothing new. It's fairly hard to follow this code and IO threading have added some technical dept too.

I think we should try to separate the code for reading, parsing, prefetching and executing. Perhaps we should do it as a separate piece of work even before we implement this command queue stuff. If we use the term "process" for executing a command, then functions that just parse the input buffer shouldn't be named "process" and they shouldn't run in the IO thread IMHO.

dvkashapov:
> what's your perf results look like?

I wasn't really able to benchmark this PR properly, but for a few seconds after launching benchmark I observed something like 600k RPS with `io-threads 9` and haven't tested with `io-threads 1` as it seems like this PR is beneficial only for enabled `io-thread`.
On `unstable` result with same config had this results:
```
Summary:
  throughput summary: 383435.59 requests per second
  latency summary (msec):
        4.981     0.056     4.959     5.255     5.503    79.039

```

src/networking.c
PingXie: > Perhaps we should do it as a separate piece of work even before we implement this command queue stuff

+1

PingXie:
Thanks @JoBeR007. This is interesting. I was able to benchmark @zuiderkwast's PR and this one edged out a bit. That being said, @zuiderkwast and I discussed offline and we will collaborate on his PR next.

src/cluster.c
```diff
@@ -1070,7 +1070,12 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
 
         for (j = 0; j < numkeys; j++) {
             robj *thiskey = margv[keyindex[j].pos];
-            int thisslot = keyHashSlot((char *)thiskey->ptr, sdslen(thiskey->ptr));
+            /* Use the cached slot to avoid the expensive hash slot calcluation for the first key.
+             * Note that slot is not computed for the first key when the io-threads are inactive. */
+            int thisslot = (j == 0) ? c->slot : -1;
```
zuiderkwast, line 1075: I wanted to pick this change but I realized that it doesn't correctly detect cross-slot transactions like `MULTI ; GET foo ; GET bar ; EXEC`.

`j == 0` here is the first key in each command, but we don't store the slot for each command in the multi-state.

I have done a more thorough implementation of this here:

* #2165

PR #2102, opened 2025-05-19
[valkey-benchmark] Allow multiple random (or sequential) placeholders and fix zadd --sequential to create expected data

rainsupreme:
This extends the `__rand_int__` pattern to add `__rand_1st__` through `__rand_9th__`.

For the new placeholders, multiple occurrences within a command will have the same value.

With the `--sequential` option, each placeholder will use separate counters.

For `__rand_int__`, the behavior is unchanged - multiple occurrences of the pattern within a command will have different values.

Examples of using the `__rand_int__` with `--sequential`:

```
$ ./valkey-benchmark -r 300 -n 300 --sequential -q -- set key__rand_int__ val__rand_int__
set key__rand_int__ val__rand_int__: 60000.00 requests per second, p50=0.391 msec
$ ./valkey-cli info keyspace
# Keyspace
db0:keys=150,expires=0,avg_ttl=0
$ ./valkey-cli get key000000000050
"val000000000051"
```

For the new patterns, multiple occurrences within the command will have the same value.

```
$ ./valkey-benchmark -r 300 -n 300 --sequential -q -- set key__rand_1st__ val__rand_1st__
set key__rand_int__ val__rand_int1_: 60000.00 requests per second, p50=0.383 msec
$ ./valkey-cli info keyspace
# Keyspace
db0:keys=300,expires=0,avg_ttl=0
$ ./valkey-cli get key000000000050
"val000000000050"
```

I also fixed the zadd benchmark so it produces the expected number of keys when used with `--sequential`. (By using the same counter twice per command it was effectively counting by twos)

I made this for myself but raised a PR in case y'all like it. 🙂

### DIFF
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -76,7 +76,7 @@
 #define CLIENT_GET_EVENTLOOP(c) (c->thread_id >= 0 ? config.threads[c->thread_id]->el : config.el)
 
 #define PLACEHOLDER_COUNT 10
-static const size_t KEY_PLACEHOLDER_LEN = 12; // length of VALKEY_BENCHMARK_PLACEHOLDERS strings
+static const size_t KEY_PLACEHOLDER_LEN = 12; // length of BENCHMARK_PLACEHOLDERS strings
 static const char *BENCHMARK_PLACEHOLDERS[PLACEHOLDER_COUNT] = {
     "__rand_int__", "__rand_int1_", "__rand_int2_", "__rand_int3_", "__rand_int4_",
     "__rand_int5_", "__rand_int6_", "__rand_int7_", "__rand_int8_", "__rand_int9_"};
@@ -157,18 +157,18 @@ typedef struct _client {
     char **randptr[PLACEHOLDER_COUNT];  /* Pointers to __rand_int__ strings inside the command buf */
     size_t randlen[PLACEHOLDER_COUNT];  /* Number of pointers in client->randptr */
     size_t randfree[PLACEHOLDER_COUNT]; /* Number of unused pointers in client->randptr */
-    char **stagptr;     /* Pointers to slot hashtags (cluster mode only) */
-    size_t staglen;     /* Number of pointers in client->stagptr */
-    size_t stagfree;    /* Number of unused pointers in client->stagptr */
-    size_t written;     /* Bytes of 'obuf' already written */
-    long long start;    /* Start time of a request */
-    long long latency;  /* Request latency */
-    int seqlen;         /* Number of commands in the command sequence */
-    int pending;        /* Number of pending requests (replies to consume) */
-    int prefix_pending; /* If non-zero, number of pending prefix commands. Commands
-                           such as auth and select are prefixed to the pipeline of
-                           benchmark commands and discarded after the first send. */
-    int prefixlen;      /* Size in bytes of the pending prefix commands */
+    char **stagptr;                     /* Pointers to slot hashtags (cluster mode only) */
+    size_t staglen;                     /* Number of pointers in client->stagptr */
+    size_t stagfree;                    /* Number of unused pointers in client->stagptr */
+    size_t written;                     /* Bytes of 'obuf' already written */
+    long long start;                    /* Start time of a request */
+    long long latency;                  /* Request latency */
+    int seqlen;                         /* Number of commands in the command sequence */
+    int pending;                        /* Number of pending requests (replies to consume) */
+    int prefix_pending;                 /* If non-zero, number of pending prefix commands. Commands
+                                           such as auth and select are prefixed to the pipeline of
+                                           benchmark commands and discarded after the first send. */
+    int prefixlen;                      /* Size in bytes of the pending prefix commands */
     int thread_id;
     struct clusterNode *cluster_node;
     int slots_last_update;
@@ -943,7 +943,7 @@ static client createClient(char *cmd, int len, int seqlen, client from, int thre
 
             for (size_t placeholder = 0; placeholder < PLACEHOLDER_COUNT; placeholder++) {
                 c->randptr[placeholder] = zmalloc(sizeof(char *) * c->randlen[placeholder]);
-                /* copy the tags to be replaced. */
+                /* copy the locations to be replaced. */
                 for (size_t instance = 0; instance < c->randlen[placeholder]; instance++) {
                     ptrdiff_t offset = from->randptr[placeholder][instance] - from->obuf;
                     /* Adjust for the different select prefix length. */
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -75,6 +75,12 @@
 
 #define CLIENT_GET_EVENTLOOP(c) (c->thread_id >= 0 ? config.threads[c->thread_id]->el : config.el)
 
+#define PLACEHOLDER_COUNT 10
+static const size_t KEY_PLACEHOLDER_LEN = 12; // length of VALKEY_BENCHMARK_PLACEHOLDERS strings
+static const char *BENCHMARK_PLACEHOLDERS[PLACEHOLDER_COUNT] = {
+    "__rand_int__", "__rand_int1_", "__rand_int2_", "__rand_int3_", "__rand_int4_",
+    "__rand_int5_", "__rand_int6_", "__rand_int7_", "__rand_int8_", "__rand_int9_"};
+
 struct benchmarkThread;
 struct clusterNode;
 struct serverConfig;
@@ -148,9 +154,9 @@ static struct config {
 typedef struct _client {
     valkeyContext *context;
     sds obuf;
-    char **randptr;     /* Pointers to :rand: strings inside the command buf */
-    size_t randlen;     /* Number of pointers in client->randptr */
-    size_t randfree;    /* Number of unused pointers in client->randptr */
+    char **randptr[PLACEHOLDER_COUNT];  /* Pointers to __rand_int__ strings inside the command buf */
+    size_t randlen[PLACEHOLDER_COUNT];  /* Number of pointers in client->randptr */
+    size_t randfree[PLACEHOLDER_COUNT]; /* Number of unused pointers in client->randptr */
     char **stagptr;     /* Pointers to slot hashtags (cluster mode only) */
     size_t staglen;     /* Number of pointers in client->stagptr */
     size_t stagfree;    /* Number of unused pointers in client->stagptr */
@@ -395,7 +401,8 @@ static void freeClient(client c) {
     valkeyFree(c->context);
     if (c->paused) releasePausedClient(c);
     sdsfree(c->obuf);
-    zfree(c->randptr);
+    for (size_t placeholder = 0; placeholder < PLACEHOLDER_COUNT; placeholder++)
+        zfree(c->randptr[placeholder]);
     zfree(c->stagptr);
     zfree(c);
     if (config.num_threads) pthread_mutex_lock(&(config.liveclients_mutex));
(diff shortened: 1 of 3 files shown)


zuiderkwast:
Nice!

Have you considered `__rand_1st__`, `__rand_2nd__`, `__rand_3rd__`, `__rand_4th__`...? Looks nicer?

I want this for random values too!

It's in my mental backlog since I did #2057. I haveaslightly different behavior in mind though. In a transaction or pipeline sequence, I want use the same random number multiple times, for example:

    MULTI ; GET k__rand_1st__ ; PEXPIRETIME k__rand_1st__ ; EXEC

(This is a simulation of the suggested GETPXT command in #1455.)

Same placeholder, same value.

Would this semantic work for sequential too?

rainsupreme:
Hm, interesting. So you're suggesting that it only retrieve one value per placeholder in a command, whether that's random or sequential? I like that - with multiple placeholders you aren't stuck with duplicating the same value either.

How to implement it is not immediately obvious though. The way benchmark is written, for pipelined commands it just duplicates the test command multiple times into a string and forgets about it. It doesn't currently track which parts of the string correspond to prefix/auth stuff or which copy of the command. I'm pretty sure each client sends X pipelined commands and then waits for X replies before sending another batch of commands. (Assuming clients are kept alive)

zuiderkwast:
> How to implement it is not immediately obvious though. The way benchmark is written, for pipelined commands it just duplicates the test command multiple times into a string and forgets about it.

I have an idea: We can do one replacement when we write the pipeline to the buffer and another replacement when we execute it.

The first replacement inserts some special marker each time a placeholder's value (or counter) should be bumped, alternately when the placeholder's value should *not*  be bumped.

Example command sequence: `HSET __rand_1st__ __rand_2nd__ __rand_3rd__ ; EXPIRE __rand_1st__ 10`.

With `-P 3` we repeat this sequence 3 times into the buffer as RESP, and we modify the placeholders with some marker `#` for when the placeholder's counter shouldn't be bumped, i.e. all occurrences except the first of each placeholder. The buffer now looks like below (but in RESP):

```
HSET __rand_1st__ __rand_2nd__ __rand_3rd__ ; EXPIRE __rand#1st__ 10
HSET __rand_1st__ __rand_2nd__ __rand_3rd__ ; EXPIRE __rand#1st__ 10
HSET __rand_1st__ __rand_2nd__ __rand_3rd__ ; EXPIRE __rand#1st__ 10
```

When we execute it, we replace each placeholder with a random or sequential value and we don't bump it at the `#` placeholder.

Sounds doable?

rainsupreme:
Ok, I came up with a revision that is used like you suggested! If the same replacement is used with the same command string it will have the same value, and with --sequential they each use independent counters as before. I also made the placeholder location data global rather than duplicated across all clients. It's initialized before threads are created, and is not modified during a test.

I also revised the placeholder strings, though I'm reluctant to change the original `__rand_int__` to avoid breaking existing scripts/blogs/documentation/etc.

zuiderkwast:
> Ok, I came up with a revision that is used like you suggested! If the same replacement is used with the same command string it will have the same value, and with --sequential they each use independent counters as before. I also made the placeholder location data global rather than duplicated across all clients. It's initialized before threads are created, and is not modified during a test.

Awesome!

> I also revised the placeholder strings, though I'm reluctant to change the original `__rand_int__` to avoid breaking existing scripts/blogs/documentation/etc.

Yeah, we should probably not do this for `__rand_int__`. We can increment/bump that one at every occurrence to keep backward compatibility. We can use the new behavior only for the new placeholders. We'd just need to explain it in the --help text and then it will not be too confusing for users, or WDYT?

src/valkey-benchmark.c
```diff
@@ -2122,7 +2173,7 @@ int main(int argc, char **argv) {
         if (test_is_selected("mset")) {
             const char *cmd_argv[21];
             cmd_argv[0] = "MSET";
-            sds key_placeholder = sdscatprintf(sdsnew(""), "key%s:__rand_int__", tag);
+            sds key_placeholder = sdscatprintf(sdsnew(""), "key%s:%s", tag, BENCHMARK_PLACEHOLDERS[0]);
             for (i = 1; i < 21; i += 2) {
                 cmd_argv[i] = key_placeholder;
                 cmd_argv[i + 1] = data;
```
zuiderkwast: The 10 keys in MGET need to be different keys when running without --sequential and with --sequential too.

It seems safest we should keep the old behavior for `__rand_int__` and introduce the new behavior only for the new placeholders.

rainsupreme:
@zuiderkwast I've implemented your suggestions. What do you think? 😁

src/valkey-benchmark.c
```diff
@@ -1616,6 +1662,10 @@ int parseOptions(int argc, char **argv) {
         "                    use the same key.\n"
         " --sequential       Modifies the -r argument to replace the string __rand_int__\n"
         "                    with 12 digit numbers sequentially instead of randomly.\n"
+        "                    __rand_1st__ through __rand_9th__ are available with independent\n"
+        "                    counters. Used to create expected number of elements with multiple\n"
+        "                    replacements.\n"
+        "                    example: ZADD myzset __rand_int__ element:__rand_1st__\n"
```
zuiderkwast, line 1856: As mentioned, I think we need to keep the old behaviour just for `__rand_int__`, i.e. every occurrence is a different value. We can use the new behaviour for 1st to 9th.

An example is `MGET` with 10 keys. They should all be different, whether sequential or random values are used. With `__rand_int__` it's still possible to write this test.

For MSET, I guess a good command like could be

    MSET __rand_int__ __rand_1st__ __rand_int__ rand_2nd__ __rand_int__ __rand_3rd__ ...

This would make sure it sets different continous keys in sequential mode and different keys in random mode. When repeating the command, the values are different values every time so the command is not a no-op.

Does this make sense?

rainsupreme: That seems pretty reasonable to me. I'll go ahead and make that change.

src/valkey-benchmark.c
```diff
@@ -352,6 +361,108 @@ static void freeServerConfig(serverConfig *cfg) {
...
+
+    /* convert key to string at first location */
+    char *p = cmd + indices[0] + KEY_PLACEHOLDER_LEN - 1;
+    for (size_t j = 0; j < KEY_PLACEHOLDER_LEN; j++) {
+        *p = '0' + key % 10;
+        key /= 10;
+        p--;
+    }
+
+    /* copy the first instace to the other locations */
```
```suggestion
    /* copy the first instance to the other locations */
```

src/valkey-benchmark.c
```diff
@@ -140,24 +146,27 @@ static struct config {
     int resp3; /* use RESP3 */
 } config;
 
+static struct placeholders {
```
zuiderkwast, line 156: We have other placeholders too, e.g. `{tag}` so a comment can help understanding what this is.

```suggestion
/* Locations of the placeholders __rand_int__, __rand_1st__,
 * __rand_2nd, etc. within the RESP encoded command buffer. */
static struct placeholders {
```

src/valkey-benchmark.c
```diff
@@ -352,6 +361,108 @@ static void freeServerConfig(serverConfig *cfg) {
     zfree(cfg);
 }
 
+void resetPlaceholders(void) {
+    if (placeholders.indices[0])
+        zfree(placeholders.indices[0]); /* indices are a single contiguous allocation */
```
zuiderkwast: Oh, confusing!

Let's use a separate struct field for the indices continuous allocation and let each of them point into this one, i.e. even indices[0] can point to the beginning of the continuous allocation.

```suggestion
    if (placeholders.all_indices)
        zfree(placeholders.all_indices); /* indices are a single contiguous allocation */
```

src/valkey-benchmark.c
```diff
@@ -352,6 +361,108 @@ static void freeServerConfig(serverConfig *cfg) {
...
+            size_t index = p - cmd;
+            temp_indices[placeholder][*count] = index;
+            (*count)++;
+            total_count++;
+            p += KEY_PLACEHOLDER_LEN; // Move past the placeholder
+        }
+    }
+
+    /* consolidate temp data into contiguous allocation */
+    placeholders.indices[0] = zmalloc(sizeof(size_t) * total_count);
```
zuiderkwast: To avoid confusion, let's store this special allocation in a separate variable.

```suggestion
    placeholders.all_indices = zmalloc(sizeof(size_t) * total_count);
```

The code below makes sure even `placeholders.indices[0]` points to this one.

src/valkey-benchmark.c
```diff
@@ -75,6 +75,12 @@
 
 #define CLIENT_GET_EVENTLOOP(c) (c->thread_id >= 0 ? config.threads[c->thread_id]->el : config.el)
 
+#define PLACEHOLDER_COUNT 10
+static const size_t KEY_PLACEHOLDER_LEN = 12; // length of BENCHMARK_PLACEHOLDERS strings
+static const char *BENCHMARK_PLACEHOLDERS[PLACEHOLDER_COUNT] = {
+    "__rand_int__", "__rand_1st__", "__rand_2nd__", "__rand_3rd__", "__rand_4th__",
+    "__rand_5th__", "__rand_6th__", "__rand_7th__", "__rand_8th__", "__rand_9th__"};
```
zuiderkwast: Let's name these in a similar way, same prefix.

```suggestion
#define PLACEHOLDER_COUNT 10
static const size_t PLACEHOLDER_LEN = 12; // length of PLACEHOLDERS strings
static const char *PLACEHOLDERS[PLACEHOLDER_COUNT] = {
    "__rand_int__", "__rand_1st__", "__rand_2nd__", "__rand_3rd__", "__rand_4th__",
    "__rand_5th__", "__rand_6th__", "__rand_7th__", "__rand_8th__", "__rand_9th__"};
```

src/valkey-benchmark.c
```diff
@@ -2015,19 +2077,19 @@ int main(int argc, char **argv) {
         }
 
         if (test_is_selected("set")) {
-            len = valkeyFormatCommand(&cmd, "SET key%s:__rand_int__ %s", tag, data);
+            len = valkeyFormatCommand(&cmd, "SET key%s:%s %s", tag, BENCHMARK_PLACEHOLDERS[0], data);
```
zuiderkwast: Why not keep the existing code for these commands? The explicit `__rand_int__` is easier to read IMHO.

src/valkey-benchmark.c
```diff
@@ -352,6 +361,108 @@ static void freeServerConfig(serverConfig *cfg) {
...
+
+        /* for __rand_int__, multiple instances will have different values */
+        size_t *indices = placeholders.indices[0];
+        _Atomic uint64_t *key_counter = &seq_key[0];
+        for (size_t i = 0; i < placeholders.count[0]; i++) {
+            replacePlaceholder(indices + i, 1, cmd, key_counter);
+        }
+
+        /* For other placeholders, multiple occurences within the command will
+         * have the same value */
```
zuiderkwast: How does this work with pipelines? All occurrences within the pipeline don't get the same value?

Is it worth adding a test case with this with pipeline?

zuiderkwast (commented):
Great!

src/valkey-benchmark.c
rainsupreme: The outer for loop here (L448) handles this command-by-command, so it doesn't matter how many are duplicated into the buffer for pipelining. I did add it to a test though :)

tests/integration/valkey-benchmark.tcl
```diff
@@ -167,9 +167,12 @@ tags {"benchmark network external:skip logreqres:skip"} {
...
 
         test {benchmark: besides first placeholder, multiple placeholder occurences have same value} {
-            set cmd [valkeybenchmark $master_host $master_port "-r 100 -n 1000 -- set rain__rand_1st__ rain__rand_1st__"]
+            set cmd [valkeybenchmark $master_host $master_port "-r 100 -n 100 -P 5 --sequential -- set rain__rand_1st__ rain__rand_1st__"]
             common_bench_setup $cmd
-            assert_match  {*calls=1000,*} [cmdstat set]
+            assert_match  {*calls=100,*} [cmdstat set]
+            
+            # Each command is handled separately regardness of pipelining
+            assert_match  {100} [scan [regexp -inline {keys\=([\d]*)} [r info keyspace]] keys=%d]
```
zuiderkwast: Good check for pipelines, just a little difficult to read with scan + regexp. We can skip scan like this:

```suggestion
            assert_equal  {keys=100} [regexp -inline {keys=[\d]*} [r info keyspace]]
```

zuiderkwast (approved):
Great! Just a very minor nit. Clang-format and spellchecker have some comments.

zuiderkwast:
Need to add "nd" (for 2nd) to the wordlist I guess, `.config/typos.toml`. There is already an "nd" under `[type.c.extend-identifiers]` but we can move it to `[default.extend-words]` maybe.

zuiderkwast:
Can you update the title and top comment of the PR to what was finally implemented?

PR #2103, opened 2025-05-20
Remove unused calculation

vitahlin:
This PR cleans up some unused or redundant code in `object.c`.
1. Since `min_size` is a local variable that’s not used or returned, there’s no need to update it again.
2. Remove unreachable `break` after `serverPanic` in default case
3. Refactor `objectGetExpire` to avoid unnecessary pointer initialization

### DIFF
--- a/src/object.c
+++ b/src/object.c
@@ -406,7 +406,7 @@ robj *dupStringObject(const robj *o) {
         d->encoding = OBJ_ENCODING_INT;
         d->ptr = o->ptr;
         return d;
-    default: serverPanic("Wrong encoding."); break;
+    default: serverPanic("Wrong encoding.");
     }
 }
 
@@ -564,7 +564,7 @@ void decrRefCount(robj *o) {
             case OBJ_HASH: freeHashObject(o); break;
             case OBJ_MODULE: freeModuleObject(o); break;
             case OBJ_STREAM: freeStreamObject(o); break;
-            default: serverPanic("Unknown object type"); break;
+            default: serverPanic("Unknown object type");
             }
         }
         zfree(o);
--- a/src/object.c
+++ b/src/object.c
@@ -256,7 +256,7 @@ sds objectGetKey(const robj *val) {
 
 long long objectGetExpire(const robj *val) {
     if (val->hasexpire) {
-        unsigned char *data = (void *) (val + 1);
+        unsigned char *data = (void *)(val + 1);
         return *(long long *)data;
     } else {
         return -1;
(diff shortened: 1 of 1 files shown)


vitahlin:
The failed CI already exists in: https://github.com/valkey-io/valkey/actions/runs/15103516197/job/42448135060?pr=2100
This doesn’t seem to be related to the modifications made in this PR.

src/object.c
```diff
@@ -80,7 +80,6 @@ robj *createObjectWithKeyAndExpire(int type, void *ptr, const sds key, long long
      * don't need it now. Then we don't need to realloc if it's needed later. */
     if (has_embkey && !has_expire && bufsize >= min_size + sizeof(long long)) {
         has_expire = 1;
-        min_size += sizeof(long long);
```
zuiderkwast, line 83: When `has_expire` is set to 1, it affects the minimum size, so even though `min_size` is not used after this point, this code keeps the value correct. This is for the case we would change some code to use `min_size` in the future. If the code would be structured in a way so `min_size` goes out of scope, it would be even better, but since it doesn't, I think it's better to keep the value correct.

The compiler easily detects that the variable is unused after this point and removes the computation.

src/object.c
```diff
@@ -257,8 +255,8 @@ sds objectGetKey(const robj *val) {
 }
 
 long long objectGetExpire(const robj *val) {
-    unsigned char *data = (void *)(val + 1);
     if (val->hasexpire) {
+        unsigned char *data = (void *)(val + 1);
```
zuiderkwast, line 259: This is a completely unnecessary change. The compiler easily skips the initialization in the case when it's not used.

We avoid unnecessary changes for a few reasons. More changes to review, more commits to backport to downstream projects, affected git history, etc.

src/object.c
```diff
@@ -564,7 +562,7 @@ void decrRefCount(robj *o) {
             case OBJ_HASH: freeHashObject(o); break;
             case OBJ_MODULE: freeModuleObject(o); break;
             case OBJ_STREAM: freeStreamObject(o); break;
-            default: serverPanic("Unknown object type"); break;
+            default: serverPanic("Unknown object type");
```
zuiderkwast, line 565: This is the kind of changes that can introduce some compiler warnings in some unusual setups, like some old GCC version, etc. The change is unnecessary and not worth the risk IMHO.

zuiderkwast (changes requested):
Sorry, I don't agree. :)

vitahlin:
> Sorry, I don't agree. :)

That makes a lot of sense. Thanks for the detailed explanation — I’ll close this PR accordingly. 
Appreciate the feedback!

PR #2107, opened 2025-05-20
Optimize WATCH by equalStringObjects early length check 

vitahlin:
Avoid unnecessary `memcmp()` when both objects are sds and the lengths differ.

I found that this function is called in `watchCommand`, so I verified the performance improvement of this PR using the WATCH command:

```shell
$ memtier_benchmark \
  --command="WATCH a aa aaa aaaa aaaaa b bb bbb bbbb bbbbb" \
  --test-time 60 \
  -c 1 \
  -t 1 \
. --hide-histogram
```

Before:
```shell
ALL STATS

==================================================================================================
Type         Ops/sec    Avg. Latency     p50 Latency     p99 Latency   p99.9 Latency       KB/sec 
--------------------------------------------------------------------------------------------------
Watchs      43053.91         0.02273         0.02300         0.03900         0.06300      4666.98 
Totals      43053.91         0.02273         0.02300         0.03900         0.06300      9333.95 
```

After:
```shell
ALL STATS
==================================================================================================
Type         Ops/sec    Avg. Latency     p50 Latency     p99 Latency   p99.9 Latency       KB/sec 
--------------------------------------------------------------------------------------------------
Watchs      43414.55         0.02216         0.02300         0.03900         0.05500      4706.07 
Totals      43414.55         0.02216         0.02300         0.03900         0.05500      9412.14
```

### DIFF
--- a/src/object.c
+++ b/src/object.c
@@ -936,7 +936,7 @@ int collateStringObjects(const robj *a, const robj *b) {
 
 /* Equal string objects return 1 if the two objects are the same from the
  * point of view of a string comparison, otherwise 0 is returned. Note that
- * this function is faster then checking for (compareStringObject(a,b) == 0)
+ * this function is faster than checking for (compareStringObject(a,b) == 0)
  * because it can perform some more optimization. */
 int equalStringObjects(robj *a, robj *b) {
     if (a->encoding == OBJ_ENCODING_INT && b->encoding == OBJ_ENCODING_INT) {
--- a/src/object.c
+++ b/src/object.c
@@ -936,7 +936,9 @@ int equalStringObjects(robj *a, robj *b) {
          * long is the same. */
         return a->ptr == b->ptr;
     } else {
-        if (stringObjectLen(a) != stringObjectLen(b)) {
+        if (a->encoding != OBJ_ENCODING_INT &&
+            b->encoding != OBJ_ENCODING_INT &&
+            sdslen(a->ptr) != sdslen(b->ptr)) {
             return 0;
         }
         return compareStringObjects(a, b) == 0;
--- a/src/object.c
+++ b/src/object.c
@@ -85,11 +85,11 @@ robj *createObjectWithKeyAndExpire(int type, void *ptr, const sds key, long long
     o->hasexpire = has_expire;
 
     /* The memory after the struct where we embedded data. */
-    char *data = (void *) (o + 1);
+    char *data = (void *)(o + 1);
 
     /* Set the expire field. */
     if (o->hasexpire) {
-        *(long long *) data = expire;
+        *(long long *)data = expire;
         data += sizeof(long long);
     }
 
@@ -180,11 +180,11 @@ static robj *createEmbeddedStringObjectWithKeyAndExpire(const char *val_ptr,
     }
 
     /* The memory after the struct where we embedded data. */
-    char *data = (void *) (o + 1);
+    char *data = (void *)(o + 1);
 
     /* Set the expire field. */
     if (o->hasexpire) {
-        *(long long *) data = expire;
+        *(long long *)data = expire;
         data += sizeof(long long);
     }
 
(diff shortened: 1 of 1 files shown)


src/object.c
```diff
@@ -906,6 +909,11 @@ int compareStringObjectsWithFlags(const robj *a, const robj *b, int flags) {
     }
     if (flags & STRING_COMPARE_COLL) {
         return strcoll(astr, bstr);
+    } else if (flags & STRING_COMPARE_EQUAL) {
+        if (alen != blen) {
+            return 0;
+        }
+        return memcmp(astr, bstr, alen) == 0;
```
zhulipeng: It seems you have changed the behavior of `compareStringObjectsWithFlags`. Both `strcoll` and `memcpy` can return three types of values: **negative**, **zero**, and **positive**. If you want to add a quick path for string length comparison, is it possible to incorporate this into `equalStringObjects`?

vitahlin: I’ve thought about that as well. The current implementation is mostly similar to the original — the main difference is that it performs an early length check to avoid unnecessary comparisons, which can slightly improve performance in some cases. I’ve also added comments to make the logic clearer.

At the moment, this function is only used within this file — specifically in compareStringObjects, collateStringObjects, and equalStringObjects. So the impact is limited and contained.

That said, if we feel it’s more appropriate to keep the behavior strictly aligned with the original version, we could consider changing it to something like:
```c
if (a equals b) return 0;
else return 1;
```

That would align more closely with the old design. Let me know your preference — I’m open to adjusting it either way.

zuiderkwast: A flag that completely changes the meaning of the function and the return value is bad style.

Put this special case in equalStringObjects instead as Lipeng suggested. That the function you want to optimize and it doesn't need to compare them in alphabetic order.

It's just something like

```
    } else if (a->encoding != OBJ_ENCODING_INT &&
               b->encoding != OBJ_ENCODING_INT &&
               sdslen(a->ptr) != sdslen(b->ptr)) {
        return 0;
```

vitahlin: Sounds good! I’ll make the changes later. Marking the conversation as resolved. Thanks

vitahlin:
Test result of latest commit:

Unstable:
```
ALL STATS
==================================================================================================
Type         Ops/sec    Avg. Latency     p50 Latency     p99 Latency   p99.9 Latency       KB/sec 
--------------------------------------------------------------------------------------------------
Watchs      30603.56         0.03250         0.03100         0.06300         0.08700      3317.38 
Totals      30603.56         0.03250         0.03100         0.06300         0.08700      6634.76 
```

PR:
```
ALL STATS
==================================================================================================
Type         Ops/sec    Avg. Latency     p50 Latency     p99 Latency   p99.9 Latency       KB/sec 
--------------------------------------------------------------------------------------------------
Watchs      30813.01         0.03227         0.03100         0.06300         0.08700      3340.08 
Totals      30813.01         0.03227         0.03100         0.06300         0.08700      6680.16 
```

Fusl:
Shouldn't this length check be done inside `compareStringObjectsWithFlags` where we already grab `alen` and `blen`?

zuiderkwast:
@Fusl see earlier comments

src/object.c
```diff
@@ -928,14 +928,19 @@ int collateStringObjects(const robj *a, const robj *b) {
...
- * this function is faster then checking for (compareStringObject(a,b) == 0)
+ * this function is faster than checking for (compareStringObject(a,b) == 0)
  * because it can perform some more optimization. */
 int equalStringObjects(robj *a, robj *b) {
     if (a->encoding == OBJ_ENCODING_INT && b->encoding == OBJ_ENCODING_INT) {
         /* If both strings are integer encoded just check if the stored
          * long is the same. */
         return a->ptr == b->ptr;
     } else {
+        if (a->encoding != OBJ_ENCODING_INT &&
```
zuiderkwast: Nit: Reduce the nesting with elseif.

Fusl:
@zuiderkwast Ah I missed the earlier comments since they were tucked away as resolved, that makes more sense now, thanks.

zuiderkwast (approved):
:+1: 

Seems to be some problem with github actions today. I'll re-run them.

PR #2108, opened 2025-05-20
Fix a case of out of bound read when cluster node ID is provided with wrong length in CLUSTER FORGET

ranshid:
In case we issue a cluster forget with a node ID which has a smaller length than 40 bytes, we can read over the allocated string when checking the blacklist.

This will mainly impact memory inspection codes and should not introduce any real bug.

Example:

```
==30858==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x603000024e3a at pc 0x7f9b9fe7821a bp 0x7ffeb5712980 sp 0x7ffeb5712128
READ of size 40 at 0x603000024e3a thread T0
```

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -7131,7 +7131,10 @@ int clusterCommandSpecial(client *c) {
         /* CLUSTER FORGET <NODE ID> */
         clusterNode *n = clusterLookupNode(c->argv[2]->ptr, sdslen(c->argv[2]->ptr));
         if (!n) {
-            if (clusterBlacklistExists((char *)c->argv[2]->ptr))
+            if (sdslen(c->argv[2]->ptr) != CLUSTER_NAMELEN) {
+                /* Sanity check for the provided node ID length*/
+                addReplyErrorFormat(c, "Bad node name %s", (char *)c->argv[2]->ptr);
+            } else if (clusterBlacklistExists((char *)c->argv[2]->ptr))
                 /* Already forgotten. The deletion may have been gossipped by
                  * another node, so we pretend it succeeded. */
                 addReply(c, shared.ok);
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -2145,14 +2145,10 @@ void clusterBlacklistAddNode(clusterNode *node) {
  * You don't need to pass an sds string here, any pointer to 40 bytes
  * will work. */
 int clusterBlacklistExists(char *nodeid, size_t len) {
-    clusterBlacklistCleanup();
-
-    /* Sanity check. In case the provided node ID length is wrong we can bail early. */
-    if (len != CLUSTER_NAMELEN)
-        return 0;
-
-    sds id = sdsnewlen(nodeid, CLUSTER_NAMELEN);
+    sds id = sdsnewlen(nodeid, len);
     int retval;
+
+    clusterBlacklistCleanup();
 
     retval = dictFind(server.cluster->nodes_black_list, id) != NULL;
     sdsfree(id);
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -2144,11 +2144,16 @@ void clusterBlacklistAddNode(clusterNode *node) {
 /* Return non-zero if the specified node ID exists in the blacklist.
  * You don't need to pass an sds string here, any pointer to 40 bytes
  * will work. */
-int clusterBlacklistExists(char *nodeid) {
+int clusterBlacklistExists(char *nodeid, size_t len) {
+    clusterBlacklistCleanup();
+
+    /* Sanity check. In case the provided node ID length is wrong we can bail early. */
+    if (len != CLUSTER_NAMELEN)
+        return 0;
+
     sds id = sdsnewlen(nodeid, CLUSTER_NAMELEN);
     int retval;
 
-    clusterBlacklistCleanup();
     retval = dictFind(server.cluster->nodes_black_list, id) != NULL;
     sdsfree(id);
     return retval;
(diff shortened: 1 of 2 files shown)


src/cluster_legacy.c
```diff
@@ -7131,7 +7131,10 @@ int clusterCommandSpecial(client *c) {
         /* CLUSTER FORGET <NODE ID> */
         clusterNode *n = clusterLookupNode(c->argv[2]->ptr, sdslen(c->argv[2]->ptr));
         if (!n) {
-            if (clusterBlacklistExists((char *)c->argv[2]->ptr))
+            if (sdslen(c->argv[2]->ptr) != CLUSTER_NAMELEN) {
+                /* Sanity check for the provided node ID length*/
+                addReplyErrorFormat(c, "Bad node name %s", (char *)c->argv[2]->ptr);
```
ranshid: Maybe this is a poor error string?

madolson: Why not just use the same error? I guess I would expect we would have just done `if (verifyClusterNodeId(c->argv[2]->ptr, sdslen(c->argv[2]->ptr)) && clusterBlacklistExists((char *)c->argv[2]->ptr)) {`

madolson: Actually, maybe we do a length check in clusterBlacklistExists? Then we also cover the other location? Seems a bit more defensive.

ranshid: Well. I thought of that but was trying to avoid changing the API and provide a way to distinguish between the case the node was not found and the fact that the user made a mistake (like forgot to add the last character). However I am fine changing the   clusterBlacklistExists API to accept length as well. just +1 on this if you prefer it that way

ranshid: adding that it also makes it easier to add test :)

uriyage (approved):

src/cluster_legacy.c
enjoy-binbin: return a `Unknown node` seem better to me since we has other places like this, like we also don't do the length check in this case
```
    } else if (!strcasecmp(c->argv[3]->ptr, "node") && c->argc >= 5) {
        /* CLUSTER SETSLOT <SLOT> NODE <NODE ID> */
        n = clusterLookupNode(c->argv[4]->ptr, sdslen(c->argv[4]->ptr));
        if (!n) {
            addReplyErrorFormat(c, "Unknown node %s", (char *)c->argv[4]->ptr);
            return 0;
        }
```

madolson: Yeah. I agree with Binbin. We've discussed in the past having alternative node names as well, I don't want to couple the length to the format of the node name. It feels like we're writing a tightly coupled test to the implementation.

ranshid: @madolson @enjoy-binbin fixed per your comments.

enjoy-binbin (approved):

src/cluster_legacy.c
```diff
@@ -2144,11 +2144,16 @@ void clusterBlacklistAddNode(clusterNode *node) {
 /* Return non-zero if the specified node ID exists in the blacklist.
  * You don't need to pass an sds string here, any pointer to 40 bytes
  * will work. */
-int clusterBlacklistExists(char *nodeid) {
+int clusterBlacklistExists(char *nodeid, size_t len) {
+    clusterBlacklistCleanup();
+
+    /* Sanity check. In case the provided node ID length is wrong we can bail early. */
+    if (len != CLUSTER_NAMELEN)
```
PingXie: this check is a bit odd. rather, I think `nodeid`'s length being 40 bytes is part of the `clusterBlacklistExists` contract and it is the caller who is responsible for validating the `nodeid`'s length.

src/cluster_legacy.c
```diff
@@ -7131,7 +7136,7 @@ int clusterCommandSpecial(client *c) {
         /* CLUSTER FORGET <NODE ID> */
         clusterNode *n = clusterLookupNode(c->argv[2]->ptr, sdslen(c->argv[2]->ptr));
```
PingXie, line 7133: I would suggest blocking invalid nodeid here.

ranshid: So we discussed it in an earlier thread. initially I validated it here, but @madolson and @enjoy-binbin suggested it might be more logical to treat it the same as any unknown node.

src/cluster_legacy.c
ranshid: So I agree there is some "undocumented" contract here, which is why this is simply a sanity check which is meant to avoid allocating new string and searching the dictionary for no good reason. TBH we can drop this check and just search the dictionary (were we will not find it).
The reason for placing the check logic here was to make the code more defensive.

src/cluster_legacy.c
```diff
@@ -7131,7 +7136,7 @@ int clusterCommandSpecial(client *c) {
...
         if (!n) {
-            if (clusterBlacklistExists((char *)c->argv[2]->ptr))
+            if (clusterBlacklistExists((char *)c->argv[2]->ptr, sdslen(c->argv[2]->ptr)))
```
enjoy-binbin, line 7135: I guess Ping wants to return here, something like this (i also like this)
```suggestion
            if (sdslen(sdslen(c->argv[2]->ptr)) == CLUSTER_NAMELEN && clusterBlacklistExists((char *)c->argv[2]->ptr))
```

ranshid: I am fine with that as well, although making the check inside clusterBlacklistExists is more defensive. However I have no special argument against any of the suggestions.

ranshid: we can also just add the len argument to the clusterBlacklistExists and have it miss finding non existing nodes in the blacklist. this would make the code much more correct, but would keep the unneeded sds allocation and dictionary search as they are today.

ranshid: @enjoy-binbin I change as per my last comment. If you still think we should add another check for cluster forget input, I am fine with it (just to prevent the not-so-interesting case of user who provides a very long node-id and bursting memory usage for the sds copy).

PingXie (approved):

tests/unit/other.tcl
```diff
@@ -551,6 +551,11 @@ start_cluster 1 0 {tags {"other external:skip cluster slow"}} {
             fail "hash tables weren't resize."
         }
     } {} {needs:debug}
+
+    test "CLUSTER FORGET with invalid node ID" {
+         catch {r cluster forget 1} err
+         set _ $err
+    } {*ERR Unknown node*}
```
madolson, line 558: For future reference, you can do this in a one liner:
```
assert_error {*ERR Unknown node*} {r cluster forget 1}
```
The current pattern is useful primarily when the last command is already returning something.

PR #2109, opened 2025-05-21
Only mark the client reprocessing flag when unblocked on keys

ranshid:
When we refactored the blocking framework we introduced the client reprocessing infrastructure. In cases the client was blocked on keys, it will attempt to reprocess the command. One challenge was to keep track of the command timeout, since we are reprocessing and do not want to re-register the client with a fresh timeout each time. The solution was to consider the client reprocessing flag when the client is blockedOnKeys:

```
if (!c->flag.reprocessing_command) {
        /* If the client is re-processing the command, we do not set the timeout
         * because we need to retain the client's original timeout. */
        c->bstate->timeout = timeout;
    }
```

However, this introduced a new issue. There are cases where the client will consecutive blocking of different types for example:
```
CLIENT PAUSE 10000 ALL
BZPOPMAX zset 1
```
would have the client blocked on the zset endlessly if nothing will be written to it. 

**Credits to @uriyage for locating this with his fuzzer testing**

The suggested solution is to only flag the client when it is specifically unblocked on keys.

### DIFF
--- a/src/blocked.c
+++ b/src/blocked.c
@@ -403,7 +403,7 @@ void blockForKeys(client *c, int btype, robj **keys, int numkeys, mstime_t timeo
 
     initClientBlockingState(c);
 
-    if (!c->flag.reprocessing_command) {
+    if (!c->flag.reexecuting_command) {
         /* If the client is re-processing the command, we do not set the timeout
          * because we need to retain the client's original timeout. */
         c->bstate->timeout = timeout;
@@ -680,7 +680,7 @@ static void unblockClientOnKey(client *c, robj *key) {
      * we need to re process the command again */
     if (c->flag.pending_command) {
         c->flag.pending_command = 0;
-        c->flag.reprocessing_command = 1;
+        c->flag.reexecuting_command = 1;
         /* We want the command processing and the unblock handler (see RM_Call 'K' option)
          * to run atomically, this is why we must enter the execution unit here before
          * running the command, and exit the execution unit after calling the unblock handler (if exists).
@@ -700,7 +700,7 @@ static void unblockClientOnKey(client *c, robj *key) {
         exitExecutionUnit();
         afterCommand(c);
         /* Clear the reprocessing_command flag after the proc is executed. */
-        c->flag.reprocessing_command = 0;
+        c->flag.reexecuting_command = 0;
         server.current_client = old_client;
     }
 }
--- a/src/blocked.c
+++ b/src/blocked.c
@@ -699,6 +699,8 @@ static void unblockClientOnKey(client *c, robj *key) {
         }
         exitExecutionUnit();
         afterCommand(c);
+        /* Clear the reprocessing_command flag after the proc is executed. */
+        c->flag.reprocessing_command = 0;
         server.current_client = old_client;
     }
 }
--- a/src/blocked.c
+++ b/src/blocked.c
@@ -680,6 +680,7 @@ static void unblockClientOnKey(client *c, robj *key) {
      * we need to re process the command again */
     if (c->flag.pending_command) {
         c->flag.pending_command = 0;
+        c->flag.reprocessing_command = 1;
         /* We want the command processing and the unblock handler (see RM_Call 'K' option)
          * to run atomically, this is why we must enter the execution unit here before
          * running the command, and exit the execution unit after calling the unblock handler (if exists).
--- a/src/blocked.c
+++ b/src/blocked.c
@@ -699,7 +699,7 @@ static void unblockClientOnKey(client *c, robj *key) {
         }
         exitExecutionUnit();
         afterCommand(c);
-        /* Clear the reprocessing_command flag after the proc is executed. */
+        /* Clear the reexecuting_command flag after the proc is executed. */
         c->flag.reexecuting_command = 0;
         server.current_client = old_client;
     }
--- a/src/server.c
+++ b/src/server.c
@@ -4024,7 +4024,7 @@ int processCommand(client *c) {
 
     /* in case we are starting to ProcessCommand and we already have a command we assume
      * this is a reprocessing of this command, so we do not want to perform some of the actions again. */
-    int client_reprocessing_command = c->flag.reprocessing_command ? 1 : 0;
+    int client_reprocessing_command = c->cmd ? 1 : 0;
 
     /* only run command filter if not reprocessing command */
     if (!client_reprocessing_command) {
--- a/src/server.c
+++ b/src/server.c
@@ -4026,7 +4026,7 @@ int processCommand(client *c) {
 
     /* in case we are starting to ProcessCommand and we already have a command we assume
      * this is a reprocessing of this command, so we do not want to perform some of the actions again. */
-    int client_reprocessing_command = c->flag.reprocessing_command;
+    int client_reprocessing_command = c->flag.reprocessing_command ? 1 : 0;
     /* we should clear this flag, since it is possible the processing will bail out due to some error
      * and we do not want to keep this flag on for the next commands */
     c->flag.reprocessing_command = 0;
--- a/src/server.c
+++ b/src/server.c
@@ -3733,8 +3733,6 @@ void call(client *c, int flags) {
     /* In case client is blocked after trying to execute the command,
      * it means the execution is not yet completed and we MIGHT reprocess the command in the future. */
     if (!c->flag.blocked) c->flag.executing_command = 0;
-    /* Clear the reprocessing_command flag after the proc is executed. */
-    c->flag.reprocessing_command = 0;
 
     /* In order to avoid performance implication due to querying the clock using a system call 3 times,
      * we use a monotonic clock, when we are sure its cost is very low, and fall back to non-monotonic call otherwise. */
--- a/src/server.c
+++ b/src/server.c
@@ -3732,10 +3732,8 @@ void call(client *c, int flags) {
 
     /* In case client is blocked after trying to execute the command,
      * it means the execution is not yet completed and we MIGHT reprocess the command in the future. */
-    if (!c->flag.blocked) {
-        c->flag.executing_command = 0;
-    }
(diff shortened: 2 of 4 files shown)


src/blocked.c
```diff
@@ -680,6 +680,7 @@ static void unblockClientOnKey(client *c, robj *key) {
      * we need to re process the command again */
     if (c->flag.pending_command) {
         c->flag.pending_command = 0;
+        c->flag.reprocessing_command = 1;
```
enjoy-binbin: can we reset this flag after the `afterCommand` in line 701?

src/server.c
```diff
@@ -3722,24 +3722,21 @@ void call(client *c, int flags) {
...
     exitExecutionUnit();
 
     /* In case client is blocked after trying to execute the command,
      * it means the execution is not yet completed and we MIGHT reprocess the command in the future. */
-    if (!c->flag.blocked) c->flag.executing_command = 0;
+    if (!c->flag.blocked) {
+        c->flag.executing_command = 0;
+    }
+    /* Clear the CLIENT_REPROCESSING_COMMAND flag after the proc is executed. */
+    c->flag.reprocessing_command = 0;
```
enjoy-binbin:
```suggestion
    if (!c->flag.blocked) c->flag.executing_command = 0;
    /* Clear the reprocessing_command flag after the proc is executed. */
    c->flag.reprocessing_command = 0;
```

src/server.c
```diff
@@ -4031,7 +4028,10 @@ int processCommand(client *c) {
 
     /* in case we are starting to ProcessCommand and we already have a command we assume
      * this is a reprocessing of this command, so we do not want to perform some of the actions again. */
-    int client_reprocessing_command = c->cmd ? 1 : 0;
+    int client_reprocessing_command = c->flag.reprocessing_command;
```
enjoy-binbin:
```suggestion
    int client_reprocessing_command = c->flag.reprocessing_command ? 1 : 0;
```

src/server.c
```diff
@@ -4031,7 +4028,10 @@ int processCommand(client *c) {
...
+    /* we should clear this flag, since it is possible the processing will bail out due to some error
+     * and we do not want to keep this flag on for the next commands */
+    c->flag.reprocessing_command = 0;
```
enjoy-binbin: or can we do this in resetClient?

do we have some cases that a re-process will throw an error?

src/server.c
```diff
@@ -4355,7 +4355,9 @@ int processCommand(client *c) {
         addReply(c, shared.queued);
     } else {
         int flags = CMD_CALL_FULL;
-        if (client_reprocessing_command) flags |= CMD_CALL_REPROCESSING;
+        /* we should re-tag the client as reprocessing in case it was reprocessing the same command */
+        if (client_reprocessing_command)
+            c->flag.reprocessing_command = 1;
```
enjoy-binbin: we added this line because this logic? We can also remove CMD_CALL_REPROCESSING flag since it is uesless now. Or we can keep it if we keep this logic and use CMD_CALL_REPROCESSING flag and then use CMD_CALL_REPROCESSING flag in call to flag the client flag.
```
    /* we should clear this flag, since it is possible the processing will bail out due to some error
     * and we do not want to keep this flag on for the next commands */
    c->flag.reprocessing_command = 0;
```

so the all changes, that will something like this:
```diff
diff --git a/src/blocked.c b/src/blocked.c
index 42d1fa641..606f3fbd6 100644
--- a/src/blocked.c
+++ b/src/blocked.c
@@ -680,6 +680,7 @@ static void unblockClientOnKey(client *c, robj *key) {
      * we need to re process the command again */
     if (c->flag.pending_command) {
         c->flag.pending_command = 0;
+        c->flag.reprocessing_command = 1;
         /* We want the command processing and the unblock handler (see RM_Call 'K' option)
          * to run atomically, this is why we must enter the execution unit here before
          * running the command, and exit the execution unit after calling the unblock handler (if exists).
@@ -699,6 +700,7 @@ static void unblockClientOnKey(client *c, robj *key) {
         exitExecutionUnit();
         afterCommand(c);
         server.current_client = old_client;
+        c->flag.reprocessing_command = 0;
     }
 }
 
diff --git a/src/server.c b/src/server.c
index f7bc6e766..9622483b2 100644
--- a/src/server.c
+++ b/src/server.c
@@ -4031,7 +4031,10 @@ int processCommand(client *c) {
 
     /* in case we are starting to ProcessCommand and we already have a command we assume
      * this is a reprocessing of this command, so we do not want to perform some of the actions again. */
-    int client_reprocessing_command = c->cmd ? 1 : 0;
+    int client_reprocessing_command = c->flag.reprocessing_command ? 1 : 0;
+    /* we should clear this flag, since it is possible the processing will bail out due to some error
+     * and we do not want to keep this flag on for the next commands */
+    c->flag.reprocessing_command = 0;
 
     /* only run command filter if not reprocessing command */
     if (!client_reprocessing_command) {

```

enjoy-binbin (commented):
Good catch, the direction LGTM

src/blocked.c
ranshid: mmm I don;t think so since the client might still be blocked. I can do it in line 699, but I wanted to keep it close to the existing implementation.
I can try though

src/server.c
ranshid: sure. It was removed on my branch. not sure why it is still here

src/server.c
ranshid: yes. for example auth changes should fail the command which was previously authed

src/blocked.c
ranshid: NVM - my bad. yes it is possible

ranshid: done

src/server.c
enjoy-binbin: can you give an example? i remember we currently won't unset the auth flag during the connection.

also with b2bb238cb8ea59ecb5622045a32450c3b6c61341, do we still need this clear logic in here?

ranshid:
```
 test {blocked command gets rejected when reprocessed after permission change} {
        r auth default ""
        r config resetstat
        r ACL setuser psuser reset on nopass +@all allkeys
        $rd AUTH psuser pspass
        $rd read
        $rd BLPOP list1 0
        wait_for_blocked_client
        r ACL setuser psuser resetkeys
        r LPUSH list1 foo
        assert_error {*NOPERM No permissions to access a key*} {$rd read}
        $rd ping
        $rd close
        assert_match {*calls=0,usec=0,*,rejected_calls=1,failed_calls=0} [cmdrstat blpop r]
    }
```

enjoy-binbin: ok i see, it is a ACL auth.

ranshid: >also with https://github.com/valkey-io/valkey/commit/b2bb238cb8ea59ecb5622045a32450c3b6c61341, do we still need this clear logic in here?

mmm you are probably right

ranshid: I pushed a refactor. lets see how it goes

tests/unit/type/list.tcl
```diff
@@ -2430,4 +2430,30 @@ foreach {pop} {BLPOP BLMPOP_RIGHT} {
...
 
+    test "Blocking timeout following PAUSE should honor the timeout" {
+        # cleanup first
+        r del mylist
+        
+        # create a test client
+        set rd [valkey_deferring_client]
+        
+        # reset the server stats
+        r config resetstat
```
enjoy-binbin:
```suggestion
```

enjoy-binbin (approved):
LGTM.

src/server.c
```diff
@@ -4031,7 +4024,7 @@ int processCommand(client *c) {
 
     /* in case we are starting to ProcessCommand and we already have a command we assume
      * this is a reprocessing of this command, so we do not want to perform some of the actions again. */
-    int client_reprocessing_command = c->cmd ? 1 : 0;
+    int client_reprocessing_command = c->flag.reprocessing_command ? 1 : 0;
```
uriyage: Do we still need this variable? can't we just check the flag?

src/server.c
```diff
@@ -4031,7 +4024,7 @@ int processCommand(client *c) {
 
     /* in case we are starting to ProcessCommand and we already have a command we assume
      * this is a reprocessing of this command, so we do not want to perform some of the actions again. */
-    int client_reprocessing_command = c->cmd ? 1 : 0;
```
uriyage:
```c
    /* Now lookup the command and check ASAP about trivial error conditions
     * such as wrong arity, bad command name and so forth.
     * In case we are reprocessing a command after it was blocked,
```
Now we do a double look up even so `c->cmd` may be already set

ranshid: Yes. That is the downside

src/server.c
ranshid: I just aim at reducing the code change here

src/server.c
ranshid: We can check that explicitly

ranshid: Actually @uriyage I think we can limit the change as you point out. this was the original reason we handled it also in processCommand.
Let me offer a small refactor

ranshid: @enjoy-binbin @uriyage what do you guys think of the latest refactor?

src/blocked.c
```diff
@@ -698,6 +699,8 @@ static void unblockClientOnKey(client *c, robj *key) {
         }
         exitExecutionUnit();
         afterCommand(c);
+        /* Clear the reprocessing_command flag after the proc is executed. */
```
uriyage: need to update comment

uriyage (approved):

src/blocked.c
ranshid: fixed

ranshid:
some tests are failing that will require checking. I will try to get to it soon

ranshid:
> * Daily / test-alpine-jemalloc (pull_request)

Only tests failing are related to a known issue: https://github.com/valkey-io/valkey/pull/2133

PR #2116, opened 2025-05-21
 Optimize GEORADIUS command performance with  pre-allocated buffer

chzhoo:
**Description**

1. The GEORADIUS command stores searched map points in geoArray, which triggers at least two memory allocations: one for the `geoArray` creation and another for each `geoPoint` initialization.
2. We can reduce these two memory allocations: 
   - converting `geoArray` from a heap-allocated variable to a stack-allocated one
   - embedding `geoPoint` directly within geoArray
  
**Benchmark**
Run the benchmark command five times with the command below, and take the peak value as the final result.
> for((x=0;x<5;x+=1));
>do
>valkey-cli flushall;
>valkey-benchmark -r $POINTS_SIZE geoadd key 100 70 l__rand_int__;
>valkey-benchmark -P 2 -n 10000000 georadius key 100 70 300 km;
> done
> 

$POINTS_SIZE | QPS before optimization|QPS after optimization|Performance Boost
-- | -- | -- | -- |
1|430015.06|436585.88|1.5%|
4|381766.81|386279.34|1.2%|
16|264592.28|268391.53|1.4%|
64|120035.05|120277.60|0.2%|

CPU: AMD EPYC 9K65 192-Core Processor* 8
OS: 	Ubuntu Server 24.04 LTS 64bit
Memory: 32GB
VM: Tencent cloud  SA9 | SA9.2XLARGE32

### DIFF
--- a/src/geo.c
+++ b/src/geo.c
@@ -50,21 +50,23 @@ int zslValueLteMax(double value, zrangespec *spec);
  * geoArray implementation
  * ==================================================================== */
 
-/* Create a new array of geoPoints. */
-geoArray *geoArrayCreate(void) {
-    geoArray *ga = zmalloc(sizeof(*ga));
-    /* It gets allocated on first geoArrayAppend() call. */
-    ga->array = NULL;
-    ga->buckets = 0;
+/* Init a new array of geoPoints. */
+void geoArrayInit(geoArray *ga) {
+    ga->buckets = MAX_GEO_ARRAY_BUFFER;
     ga->used = 0;
-    return ga;
+    ga->array = ga->arraybuf;
 }
 
 /* Add and populate with data a new entry to the geoArray. */
 geoPoint *geoArrayAppend(geoArray *ga, double *xy, double dist, double score, char *member) {
     if (ga->used == ga->buckets) {
-        ga->buckets = (ga->buckets == 0) ? 8 : ga->buckets * 2;
-        ga->array = zrealloc(ga->array, sizeof(geoPoint) * ga->buckets);
+        ga->buckets = ga->buckets * 2;
+        if (ga->array == ga->arraybuf) {
+            ga->array = zmalloc(sizeof(geoPoint) * ga->buckets);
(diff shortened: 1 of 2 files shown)


src/geo.c
```diff
@@ -50,21 +50,23 @@ int zslValueLteMax(double value, zrangespec *spec);
...
 
 /* Add and populate with data a new entry to the geoArray. */
 geoPoint *geoArrayAppend(geoArray *ga, double *xy, double dist, double score, char *member) {
     if (ga->used == ga->buckets) {
-        ga->buckets = (ga->buckets == 0) ? 8 : ga->buckets * 2;
-        ga->array = zrealloc(ga->array, sizeof(geoPoint) * ga->buckets);
+        ga->buckets = ga->buckets * 2;
+        if (ga->array == ga->arraybuf) {
+            ga->array = zmalloc(sizeof(geoPoint) * ga->buckets);
+            memcpy(ga->array, ga->arraybuf, sizeof(geoPoint) * MAX_GEO_ARRAY_BUFFER);
```
zuiderkwast:
```suggestion
            memcpy(ga->array, ga->arraybuf, sizeof(*ga->arraybuf);
```

This is equivalent, if i'm not mistaken.

src/geo.c
```diff
@@ -80,8 +82,9 @@ geoPoint *geoArrayAppend(geoArray *ga, double *xy, double dist, double score, ch
 void geoArrayFree(geoArray *ga) {
```
zuiderkwast: This function doesn't free ga anymore, so rename it to reset or something like that. It is the opposite of init.

zuiderkwast (commented):
I like it. 😀

zuiderkwast:
@KarthikSubbarao you know the geo code. PTAL.

src/geo.c
chzhoo: Based on the pattern from IOJobQueue_init and IOJobQueue_cleanup functions, I have renamed the cleanup function to geoArrayCleanup.

zuiderkwast (approved):

PR #2117, opened 2025-05-22
CLIENT UNBLOCK should't be able to unpause paused clients

### LINKED ISSUE #2111: [CRASH] crash when unblocking a client on timeout
### Crash report

```shell
#0 0x00007ffff61fe2a9 in syscall () from /lib64/libc.so.6
(gdb) bt
#0 0x00007ffff61fe2a9 in syscall () from /lib64/libc.so.6
#1 0x00000000004523eb in ThreadsManager_runOnThreads (tids=0x7fffffffc0d0, tids_len=4, callback=0x5d7d48 <collect_stacktrace_data>) at threads_mngr.c:98
#2 0x00000000005d8130 in writeStacktraces (fd=1, uplevel=3) at debug.c:1729
#3 0x00000000005d8729 in logStackTrace (eip=0x0, uplevel=3, current_thread=0) at debug.c:1816
#4 0x00000000005d72b6 in _serverPanic (file=0x7f0080 "blocked.c", line=279, msg=0x7f0360 "Unknown btype in replyToBlockedClientTimedOut().") at debug.c:1158
#5 0x00000000006636d8 in replyToBlockedClientTimedOut (c=0x61600000e480) at blocked.c:279
#6 0x00000000006661e7 in unblockClientOnTimeout (c=0x61600000e480) at blocked.c:738
#7 0x00000000004e7856 in clientUnblockCommand (c=0x61600001da80) at networking.c:4334
#8 0x000000000048f1aa in call (c=0x61600001da80, flags=3) at server.c:3733
#9 0x00000000004931e6 in processCommand (c=0x61600001da80) at server.c:4359
#10 0x00000000004de07e in processCommandAndResetClient (c=0x61600001da80) at networking.c:3142
#11 0x00000000004de7fa in processInputBuffer (c=0x61600001da80) at networking.c:3270
#12 0x00000000004df248 in readQueryFromClient (conn=0x60600000a400) at networking.c:3375
#13 0x0000000000729304 in callHandler (conn=0x60600000a400, handler=0x4df15a <readQueryFromClient>) at connhelpers.h:79
#14 0x000000000072aaf8 in connSocketEventHandler (el=0x60c00000bf80, fd=89, clientData=0x60600000a400, mask=1) at socket.c:301
#15 0x0000000000462d14 in aeProcessEvents (eventLoop=0x60c00000bf80, flags=27) at ae.c:486
#16 0x000000000046349e in aeMain (eventLoop=0x60c00000bf80) at ae.c:543
#17 0x00000000004a487c in main (argc=1, argv=0x7fffffffe068) at server.c:7212
= VALKEY BUG REPORT START: Cut & paste starting from here ===
6160:M 16 May 2025 14:22:10.387 # ------------------------------------------------
6160:M 16 May 2025 14:22:10.387 # !!! Software Failure. Press left mouse button to continue
6160:M 16 May 2025 14:22:10.387 # Guru Meditation: Unknown btype in replyToBlockedClientTimedOut(). #blocked.c:279

------ STACK TRACE ------

Thread 2 "bio_close_file" received signal SIGUSR2, User defined signal 2.
[Switching to Thread 0x7fffec530700 (LWP 6167)]
0x00007ffff64ce377 in pthread_cond_wait@@GLIBC_2.3.2 () from /lib64/libpthread.so.0
(gdb) bt
#0 0x00007ffff64ce377 in pthread_cond_wait@@GLIBC_2.3.2 () from /lib64/libpthread.so.0
#1 0x00000000006290e9 in bioProcessBackgroundJobs (arg=0x0) at bio.c:233
#2 0x00007ffff64c844b in start_thread () from /lib64/libpthread.so.0
#3 0x00007ffff620352f in clone () from /lib64/libc.so.6
(gdb) thr 1
[Switching to thread 1 (Thread 0x7ffff7fe0c00 (LWP 6160))]
#0 0x00007ffff61fe2a9 in syscall () from /lib64/libc.so.6

...
.=================================================================
==6160==ERROR: AddressSanitizer: global-buffer-overflow on address 0x000000bc23d8 at pc 0x7ffff6ea121a bp 0x7fffffefc3b0 sp 0x7fffffefbb58
READ of size 221184 at 0x000000bc23d8 thread T0
#0 0x7ffff6ea1219 (/lib64/libasan.so.4+0x76219)
#1 0x62edba in memtest_preserving_test src/memtest.c:304
#2 0x5d9829 in memtest_test_linux_anonymous_maps src/debug.c:1983
#3 0x5d9ae2 in doFastMemoryTest src/debug.c:2026
#4 0x5dafae in printCrashReport src/debug.c:2230
#5 0x5d72c3 in _serverPanic src/debug.c:1162
#6 0x6636d7 in replyToBlockedClientTimedOut src/blocked.c:279
#7 0x6661e6 in unblockClientOnTimeout src/blocked.c:738
#8 0x4e7855 in clientUnblockCommand src/networking.c:4334
#9 0x48f1a9 in call src/server.c:3733
#10 0x4931e5 in processCommand src/server.c:4359
#11 0x4de07d in processCommandAndResetClient src/networking.c:3142
#12 0x4de7f9 in processInputBuffer src/networking.c:3270
#13 0x4df247 in readQueryFromClient src/networking.c:3375
#14 0x729303 in callHandler src/connhelpers.h:79
#15 0x72aaf7 in connSocketEventHandler src/socket.c:301
#16 0x462d13 in aeProcessEvents src/ae.c:486
#17 0x46349d in aeMain src/ae.c:543
#18 0x4a487b in main src/server.c:7212
#19 0x7ffff6135139 in __libc_start_main (/lib64/libc.so.6+0x21139)
#20 0x452029 in _start (/localsrc/valkey-server+0x452029)

0x000000bc23d8 is located 0 bytes to the right of global variable 'used_memory_thread_padded' defined in 'zmalloc.c:103:57' (0xbc1b80) of size 2136
0x000000bc23d8 is located 40 bytes to the left of global variable 'total_active_threads' defined in 'zmalloc.c:109:19' (0xbc2400) of size 4
SUMMARY: AddressSanitizer: global-buffer-overflow (/lib64/libasan.so.4+0x76219) 
Shadow bytes around the buggy address:
0x000080170420: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x000080170430: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x000080170440: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x000080170450: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x000080170460: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
=>0x000080170470: 00 00 00 00 00 00 00 00 00 00 00[f9]f9 f9 f9 f9
0x000080170480: 04 f9 f9 f9 f9 f9 f9 f9 00 f9 f9 f9 f9 f9 f9 f9
0x000080170490: 04 f9 f9 f9 f9 f9 f9 f9 00 00 00 00 f9 f9 f9 f9
0x0000801704a0: 00 f9 f9 f9 f9 f9 f9 f9 00 00 00 00 00 00 00 00
0x0000801704b0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
0x0000801704c0: 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00
Shadow byte legend (one shadow byte represents 8 application bytes):
Addressable: 00
Partially addressable: 01 02 03 04 05 06 07 
Heap left redzone: fa
Freed heap region: fd
Stack left redzone: f1
Stack mid redzone: f2
Stack right redzone: f3
Stack after return: f5
Stack use after scope: f8
Global redzone: f9
Global init order: f6
Poisoned by user: f7
Container overflow: fc
Array cookie: ac
Intra object redzone: bb
ASan internal: fe
Left alloca redzone: ca
Right alloca redzone: cb
```

### Additional information

Credits to @uriyage for locating this with his fuzzer testing
However we did not yet dive deep or found a reproduction

enjoy-binbin:
When a client is blocked by something like `CLIENT PAUSE`, we should not
allow `CLIENT UNBLOCK timeout` to unblock it, since some blocking types
does not has the timeout callback, it will trigger a panic in the core,
people should use `CLIENT UNPAUSE` to unblock it.

Also using `CLIENT UNBLOCK error` is not right, it will return a UNBLOCKED
error to the command, people don't expect a `SET` command to get an error.

So in this commit, in these cases, we will return 0 to `CLIENT UNBLOCK`
to indicate the unblock is fail. The reason is that we assume that if
a command doesn't expect to be timedout, it also doesn't expect to be
unblocked by `CLIENT UNBLOCK`.

The old behavior of the following command will trigger panic in timeout
and get UNBLOCKED error in error. Under the new behavior, client unblock
will get the result of 0.
```
client 1> client pause 100000 write
client 2> set x x

client 1> client unblock 2 timeout
or
client 1> client unblock 2 error
```

Potentially breaking change, previously allowed `CLIENT UNBLOCK error`.
Fixes #2111.

### DIFF
--- a/src/blocked.c
+++ b/src/blocked.c
@@ -254,6 +254,22 @@ void unblockClient(client *c, int queue_for_reprocessing) {
     if (queue_for_reprocessing) queueClientForReprocessing(c);
 }
 
+/* Check if the specified client can be safely timed out using
+ * unblockClientOnTimeout(). */
+int blockedClientMayTimeout(client *c) {
+    if (c->bstate->btype == BLOCKED_MODULE) {
+        return moduleBlockedClientMayTimeout(c);
+    }
+
+    if (c->bstate->btype == BLOCKED_LIST ||
+        c->bstate->btype == BLOCKED_ZSET ||
+        c->bstate->btype == BLOCKED_STREAM ||
+        c->bstate->btype == BLOCKED_WAIT) {
+        return 1;
+    }
+    return 0;
+}
+
 /* This function gets called when a blocked client timed out in order to
  * send it a reply of some kind. After this function is called,
  * unblockClient() will be called with the same client as argument. */
--- a/tests/unit/pause.tcl
+++ b/tests/unit/pause.tcl
@@ -434,7 +434,23 @@ start_server {tags {"pause network"}} {
 
         r client unpause
         assert_equal [r randomkey] {}
+    }
 
+    test "CLIENT UNBLOCK is not allow to unblock client blocked by CLIENT PAUSE" {
+        set rd [valkey_deferring_client]
+        $rd client id
+        set client_id [$rd read]
+
+        r client pause 100000 write
+        $rd set foo bar
+        wait_for_blocked_clients_count 1 50 100
+
+        # This used to will trigger a panic.
+        assert_equal 0 [r client unblock $client_id timeout]
+        # THis used to will return a UNBLOCKED error.
+        assert_equal 0 [r client unblock $client_id error]
+
+        r client unpause
     }
 
     # Make sure we unpause at the end
--- a/tests/unit/pause.tcl
+++ b/tests/unit/pause.tcl
@@ -437,20 +437,33 @@ start_server {tags {"pause network"}} {
     }
 
     test "CLIENT UNBLOCK is not allow to unblock client blocked by CLIENT PAUSE" {
-        set rd [valkey_deferring_client]
-        $rd client id
-        set client_id [$rd read]
+        set rd1 [valkey_deferring_client]
+        set rd2 [valkey_deferring_client]
+        $rd1 client id
+        $rd2 client id
+        set client_id1 [$rd1 read]
+        set client_id2 [$rd2 read]
 
+        r del mylist
         r client pause 100000 write
-        $rd set foo bar
-        wait_for_blocked_clients_count 1 50 100
+        $rd1 blpop mylist 0
+        $rd2 blpop mylist 0
+        wait_for_blocked_clients_count 2 50 100
 
-        # This used to will trigger a panic.
-        assert_equal 0 [r client unblock $client_id timeout]
-        # THis used to will return a UNBLOCKED error.
-        assert_equal 0 [r client unblock $client_id error]
+        # This used to trigger a panic.
+        assert_equal 0 [r client unblock $client_id1 timeout]
+        # THis used to return a UNBLOCKED error.
+        assert_equal 0 [r client unblock $client_id2 error]
 
+        # After the unpause, it must be able to unblock the client.
         r client unpause
+        assert_equal 1 [r client unblock $client_id1 timeout]
+        assert_equal 1 [r client unblock $client_id2 error]
+        assert_equal {} [$rd1 read]
+        assert_error "UNBLOCKED*" {$rd2 read}
+
+        $rd1 close
+        $rd2 close
     }
 
     # Make sure we unpause at the end
--- a/src/networking.c
+++ b/src/networking.c
@@ -4323,11 +4323,12 @@ void clientUnblockCommand(client *c) {
     }
     if (getLongLongFromObjectOrReply(c, c->argv[2], &id, NULL) != C_OK) return;
     struct client *target = lookupClientByID(id);
-    /* Note that we never try to unblock a client blocked on a module command, which
+    /* Note that we never try to unblock a client blocked on a module command,
+     * or a client blocked by CLIENT PAUSE or some other blocking type which
      * doesn't have a timeout callback (even in the case of UNBLOCK ERROR).
      * The reason is that we assume that if a command doesn't expect to be timedout,
      * it also doesn't expect to be unblocked by CLIENT UNBLOCK */
-    if (target && target->flag.blocked && moduleBlockedClientMayTimeout(target)) {
+    if (target && target->flag.blocked && blockedClientMayTimeout(target)) {
         if (unblock_error)
             unblockClientOnError(target, "-UNBLOCKED client unblocked via CLIENT UNBLOCK");
         else
--- a/src/server.h
+++ b/src/server.h
@@ -3552,6 +3552,7 @@ void unblockClient(client *c, int queue_for_reprocessing);
 void unblockClientOnTimeout(client *c);
 void unblockClientOnError(client *c, const char *err_str);
 void queueClientForReprocessing(client *c);
+int blockedClientMayTimeout(client *c);
 void replyToBlockedClientTimedOut(client *c);
 int getTimeoutFromObjectOrReply(client *c, robj *object, mstime_t *timeout, int unit);
 void disconnectAllBlockedClients(void);


tests/unit/pause.tcl
```diff
@@ -434,7 +434,23 @@ start_server {tags {"pause network"}} {
...
+        $rd set foo bar
+        wait_for_blocked_clients_count 1 50 100
+
+        # This used to will trigger a panic.
+        assert_equal 0 [r client unblock $client_id timeout]
+        # THis used to will return a UNBLOCKED error.
+        assert_equal 0 [r client unblock $client_id error]
+
+        r client unpause
     }
```
gmbnomis, line 467:
```suggestion
        assert_equal {OK} [$rd read]
        $rd close
    }
```

tests/unit/pause.tcl
```diff
@@ -434,7 +434,23 @@ start_server {tags {"pause network"}} {
...
```
gmbnomis, line 468: I am wondering if we should also include a test involving a blocking command. `CLIENT UNBLOCK` must not be able to unblock the command if the command is blocked for other reasons (i.e. `CLIENT PAUSE`). But once unpaused, it must be possible to unblock the command. Something like:

```tcl
    test "CLIENT UNBLOCK does not unblock blocking command blocked by CLIENT PAUSE" {
        $rd client id
        set client_id [$rd read]

        r client pause 100000 write
        $rd blpop list 0
        wait_for_blocked_clients_count 1 50 100

        assert_equal 0 [r client unblock $client_id timeout]
        assert_equal 0 [r client unblock $client_id error]

        r client unpause

        assert_equal 1 [r client unblock $client_id timeout]
        assert_equal {} [$rd read]

        $rd close
    }
```

gmbnomis (commented):
Although this could be regarded as a breaking change (of undocumented behavior), I think the behavior proposed by this PR is the most consistent option.

tests/unit/pause.tcl
enjoy-binbin: That is a good idea, updated, thanks

src/blocked.c
```diff
@@ -254,6 +254,22 @@ void unblockClient(client *c, int queue_for_reprocessing) {
...
+ * unblockClientOnTimeout(). */
+int blockedClientMayTimeout(client *c) {
+    if (c->bstate->btype == BLOCKED_MODULE) {
+        return moduleBlockedClientMayTimeout(c);
+    }
+
+    if (c->bstate->btype == BLOCKED_LIST ||
+        c->bstate->btype == BLOCKED_ZSET ||
+        c->bstate->btype == BLOCKED_STREAM ||
+        c->bstate->btype == BLOCKED_WAIT) {
```
ranshid, line 267: Not that important, but maybe we should check the client has a defined timeout? The downside is that it would make this a much more "breaking change" but the logic would be much more correct IMO since if a client application blocked a client without timeout it might also not expect to be timedout

ranshid (approved):
Overall LGTM. small comment which I myself am not sure of.

src/blocked.c
gmbnomis: The [documentation for [`CLIENT UNBLOCK`](https://valkey.io/commands/client-unblock/) already demonstrates usage without a timeout, which is a valid scenario IMO. In cases where a process manages blocking commands through a separate control connection, specifying a timeout isn’t always necessary—`CLIENT UNBLOCK` allows external interruption at any time, as shown in the example.

However, this use case may not be obvious to users reading the documentation for blocking commands like [`BLPOP`](https://valkey.io/commands/blpop/). We may want to make this more clear in the documentation.

ranshid: I agree that since the timeout is the default it would make this more breaking then current state is (Which is why I referred to it). not sure which is better, I mainly thought this might be confusing in the future as the function name might be misleading (blockedClientMayTimeout may imply that the client is blocked and might timeout).

zuiderkwast: It has always been possible to use CLIENT UNBLOCK TIMEOUT even for commands without a timeout, so why is that "more breaking"?

zuiderkwast: Ah, I get it. You want to forbid CLIENT UNBLOCK TIMEOUT when the blocking command didn't set a timeout. Yeah that would be more breaking. I don't think we should break this.

zuiderkwast (approved):
LGTM.

Conceptually, for users, blocked clients and paused clients are two separate concepts. They can't be mixed.

We just use the same mechanism internally.

zuiderkwast:
Let's backport it to all versions, ok?

Or only backport the `client unblock timeout` case where it could crash?

src/blocked.c
ranshid: Yeh. I guess it is fine going with the existing suggestion. Maybe the function name should be improved to better indicate the usage?
Like `blockedClientCanBeUnblockedOnTimeout`?

zuiderkwast: I like the suggested `blockedClientMayTimeout` which is matching the module.c function `moduleBlockedClientMayTimeout`.

ranshid:
> Let's backport it to all versions, ok?
> 
> Or only backport the `client unblock timeout` case where it could crash?

I think cherry picking the exact code would be easier and not a problem. I am not sure why the error is ever used (as I am also not sure when is this API ever used in realtime scenarios and not testing)

gmbnomis:
I wonder if the change is more severe than it appears at first glance. One could implement the "monitor a changing set of keys" idea from the documentation of `CLIENT UNBLOCK` as follows:


Loop forever:
1. `BLPOP <current set of observed keys> 0`
2. If not timed out: work on result

And in parallel:

Trigger when set of keys changes:
1. Update current set of observed keys
2. `CLIENT UNBLOCK <client id>`

Key property today: Client unblocking works even during CLIENT PAUSE, ensuring the key set gets updated on next iteration. (Note that there is no need to even look at the reply of `CLIENT UNBLOCK`)

The proposed change in this PR creates ambiguity in the `CLIENT UNBLOCK` response:

"0" reply could now mean either:
1.  Client wasn't blocked
2. Client couldn't be unblocked due to pause (new meaning)

This will break the code above if  the trigger occurs when `BLPOP` is blocking due to CLIENT PAUSE: The `BLPOP` will continue to wait using the outdated set of keys even if the clients will be unpaused later.

I still think that unblocking the client is wrong (as said by @zuiderkwast CLIENT PAUSE and blocking commands are two different concepts). But maybe `CLIENT UNBLOCK` should return an error if the command can't be unblocked due to a CLIENT PAUSE? (or should `CLIENT UNBLOCK` block?)

zuiderkwast:
@gmbnomis Good point. It seems that we should unblock the client if it's blocked and paused at the same time. I.e. it should be possible to unblock a BLPOP even during client pause write. Makes sense?

gmbnomis:
> It seems that we should unblock the client if it's blocked and paused at the same time. I.e. it should be possible to unblock a BLPOP even during client pause write.

That's an interesting perspective. Basically, because pausing clients and blocking commands are different concepts from a user's perspective, we could argue that the user does not need to be concerned with the underlying reason why a blocking command is blocked.

So yes, it makes sense that `CLIENT UNBLOCK` will just unblock it. And, as this neither reads or writes, this does no harm even if done during a read-write pause. (but I suppose it will be a pain to implement this behavior)

OTOH, my scenario from above turns out to be not fully correct: Today, the server panics when a `CLIENT UNBLOCK TIMEOUT` is issued for a client that is paused (even if it is a blocking command). Only a `CLIENT UNBLOCK ERROR` works. So, I wonder if anybody is actually using this pattern...

zuiderkwast:
> Today, the server panics when a `CLIENT UNBLOCK TIMEOUT` is issued for a client that is paused (even if it is a blocking command).

Yes, I can confirm this. This is an even worse bug IMO. It means the server can crash when you try to CLIENT UNBLOCK a client that is blocking on BLPOP. The user may not know that CLIENT PAUSE is in action, for example during a failover or something like that.

The fix prevents the crash, but unfortunately it prevents unblocking BLPOP while clients are paused, but this depends on which was first: block or pause.

* If a command was blocked before paused, it is possible to unblock it using CLIENT UNBLOCK.
* If clients were paused before a blocking command was called and blocked, it's not possible to unblock it using CLIENT UNBLOCK.

gmbnomis:
> The fix prevents the crash, but unfortunately it prevents unblocking BLPOP while clients are paused, but this depends on which was first: block or pause.
> 
>     * If a command was blocked before paused, it is possible to unblock it using CLIENT UNBLOCK.
> 
>     * If clients were paused before a blocking command was called and blocked, it's not possible to unblock it using CLIENT UNBLOCK.

@zuiderkwast Yes, this is true. I think that none of the alternatives to this behavior is optimal:

1. **Unblock clients regardless of pause/block state**
    - **Pros:** Consistent unblocking behavior.
    - **Cons:** When clients are paused early in command processing, we lack the necessary context (command type, expected reply) to generate a proper timeout response, especially for module commands. This would likely require a new module API to communicate this information.
2. **Return an error when paused**
    - **Pros:** Simple to implement.
    - **Cons:** Unexpected for clients, as they may not anticipate this error.
3. **Postpone `CLIENT UNBLOCK` of a paused command until the pause expires/is lifted**
    - **Pros:** From the client’s perspective, this is similar to option 1 but with some delay—in certain cases (like `WAIT`), this may be the most logical behavior.
    - **Cons:** This introduces a (new?) edge case where one command is deferred by another. We’d need to ensure that commands are unblocked and executed in the correct order, but since blocked commands are already managed in a linked list, this might be feasible.

Overall, option 3 makes most sense to me.

zuiderkwast:
All options have pros and cons. :thinking: They also require more work, and these cases are very unusual. Maybe there is no use case for these combinations?

IMO we can merge this PR. It is simple and it prevents the crash, which is the most important problem to fix. OK?

enjoy-binbin:
@gmbnomis Thank you for your summary comment
@valkey-io/core-team Please also take a look at the comments above and vote.

ranshid:
> > The fix prevents the crash, but unfortunately it prevents unblocking BLPOP while clients are paused, but this depends on which was first: block or pause.
> > ```
> > * If a command was blocked before paused, it is possible to unblock it using CLIENT UNBLOCK.
> > 
> > * If clients were paused before a blocking command was called and blocked, it's not possible to unblock it using CLIENT UNBLOCK.
> > ```
> 
> @zuiderkwast Yes, this is true. I think that none of the alternatives to this behavior is optimal:
> 
> 1. **Unblock clients regardless of pause/block state**
>    
>    * **Pros:** Consistent unblocking behavior.
>    * **Cons:** When clients are paused early in command processing, we lack the necessary context (command type, expected reply) to generate a proper timeout response, especially for module commands. This would likely require a new module API to communicate this information.
> 2. **Return an error when paused**
>    
>    * **Pros:** Simple to implement.
>    * **Cons:** Unexpected for clients, as they may not anticipate this error.
> 3. **Postpone `CLIENT UNBLOCK` of a paused command until the pause expires/is lifted**
>    
>    * **Pros:** From the client’s perspective, this is similar to option 1 but with some delay—in certain cases (like `WAIT`), this may be the most logical behavior.
>    * **Cons:** This introduces a (new?) edge case where one command is deferred by another. We’d need to ensure that commands are unblocked and executed in the correct order, but since blocked commands are already managed in a linked list, this might be feasible.
> 
> Overall, option 3 makes most sense to me.

I think that (as I think you also state) the ambiguity was always there. when we pause clients some commands with timeout are not fulfilling the timeout guarantee since it is just too hard to do so. For example we cannot timeout commands which are just sitting in the incoming socket stream as we did not yet parse them. 
I think this PR is fine in solving the immediate problem. I would also think that extending `CLIENT PAUSE` with an extra optional argument (client-id)  would be a better choice to extend the API to be more clear and allow unblocking a specific paused client

soloestoy (approved):

zuiderkwast:
Let's merge it so it can be included in

* #2173

madolson (approved):

PR #2120, opened 2025-05-22
Changes and fixes for the new DELIFEQ command

enjoy-binbin:
DELIFEQ was added in #1975 but had some issues.

Do some refactoring to reduce the number of lines of code,
write commands need to increment dirty and we can propagate
it as DEL command.

Changes:
1. Replicate as DEL is important for compatibility with replicas
   running an older Valkey version, also it can save the compare
   logic, like other commands like GETDEL.
2. Signal modified key is for WATCH and do invalidate client-side
   caching (client tracking)
3. Keyspace notifications.
4. Dirty++ indicates there are unsaved changes for RDB.
5. Using shared strings for the RESP zero and one replies for the
   code cleanup.
6. Add FAST command flag and remove the ACCESS key specs flag.


src/commands/delifeq.json
```diff
@@ -7,6 +7,7 @@
         "arity": 3,
         "function": "delifeqCommand",
         "command_flags": [
+            "FAST",
```
enjoy-binbin, line 10: i am not sure if we need the ACCESS flag, i dont know...
```
                "flags": [
                    "RM",
                    "ACCESS",
                    "DELETE"
                ],
```

zuiderkwast: I think RM, ACCESS and DELETE are keyspec flags. Not command flags. I checked other commands.

DEL: command flag WRITE, keyspec flags RM, DELETE.

src/t_string.c
```diff
@@ -384,30 +384,24 @@ void psetexCommand(client *c) {
...
-    }
+    serverAssert(dbSyncDelete(c->db, c->argv[1]));
 
-    addReplyLongLong(c, 1);
+    /* Propagate as DEL command */
+    rewriteClientCommandVector(c, 2, shared.del, c->argv[1]);
+    signalModifiedKey(c, c->db, c->argv[1]);
+    notifyKeyspaceEvent(NOTIFY_GENERIC, "del", c->argv[1], c->db->id);
+    server.dirty++;
+    addReply(c, shared.cone);
```
zuiderkwast, line 404: Five very good points. Thanks!

Just some explanations:

1. Replicate as DEL is important for compatibility with replicas running an older Valkey version.
2. Signal modified key is for WATCH and do invalidate client-side caching (client tracking)
3. Keyspace notifications
4. Dirty++ indicates there are unsaved changes
5. Using shared strings for the RESP zero and one replies avoids some allocation and string processing.

zuiderkwast (approved):

src/commands/delifeq.json
zuiderkwast: Even the `SET` command has `ACCESS`, so I think DELIFEQ should have it too.

enjoy-binbin: yes, i meant the keyspec flag

DEL is RM, DELETE.
GETDEL is RW, ACCESS, DELETE.

```
#define CMD_KEY_ACCESS (1ULL<<4) /* Returns, copies or uses the user data from
                                  * the value of the key. */
```

src/t_string.c
enjoy-binbin: about the No5, we actually have the fast paths:
```
void addReplyLongLong(client *c, long long ll) {
    if (ll == 0)
        addReply(c, shared.czero);
    else if (ll == 1)
        addReply(c, shared.cone);
    else {
        if (prepareClientToWrite(c) != C_OK) return;
        _addReplyLongLongWithPrefix(c, ll, ':');
    }
}
```

src/commands/delifeq.json
zuiderkwast: :thinking: DELIFEQ uses it for the condition to delete or not. Does it mean access?

I think not. Compare to SETNX. It uses the key as a condition too. It does not have the ACCESS flag.

enjoy-binbin: ok, removed.

hwware (approved):

PR #2123, opened 2025-05-22
Make Tclx optional in tests (fix Daily failure)

zuiderkwast:
Tclx is not available on all platforms. Skip tests that require it if it's not available.

This also fixes the Daily failures such as this: https://github.com/valkey-io/valkey/actions/runs/15175084203/job/42673707836#step:8:4677

The diff looks large, but it isn't. The test code is just wapped in a check and indented.

```
if {[catch {package require Tclx}]} {
    if {$::verbose} {
        puts "Skipping umask test. Package Tclx not available."
    }
} else {
    (old code here)
}
```

### DIFF
--- a/tests/integration/rdb.tcl
+++ b/tests/integration/rdb.tcl
@@ -114,29 +114,36 @@ start_server [list overrides [list "dir" $server_path] keep_persistence true] {
     r del stream
 }
 
-# Prepare custom umask test scenario
-package require Tclx
-set old_umask [umask]
-set old_perm [expr {666 - $old_umask}]
 set dump_path [file join $server_path dump.rdb]
-assert_equal [file attributes $dump_path -permissions] 00$old_perm
 
-if {$old_umask == 22} {
-    set new_umask 2
-} else {
-    set new_umask 22
-}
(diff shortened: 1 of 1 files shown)


tests/integration/rdb.tcl
```diff
@@ -114,29 +114,36 @@ start_server [list overrides [list "dir" $server_path] keep_persistence true] {
...
 set dump_path [file join $server_path dump.rdb]
-assert_equal [file attributes $dump_path -permissions] 00$old_perm
 
-if {$old_umask == 22} {
-    set new_umask 2
+# Prepare custom umask test scenario
+if {[catch {package require Tclx}]} {
+    if {$::verbose} {
+        puts "Skipping umask test. Package Tclx not available."
+    }
```
zuiderkwast: Just to trigger run-extra-tests

```suggestion
    if {$::verbose} {
        puts "Skipping umask test. Package Tclx not installed."
    }
```

Fusl (approved):
> \o/ All tests passed without errors!

madolson (approved):

PR #2126, opened 2025-05-22
Fix UBSAN run for hashtable unittest

zuiderkwast:
The unit test declared an array of size 1 and used pointers to elements outside the array. This is fine because the pointers are never dereferenced, but undefined-sanitizer complains.

Now, instead allocate a huge array to make sure all pointers into it are valid.

Example failure:

https://github.com/valkey-io/valkey/actions/runs/15175084203/job/42673713108#step:10:123


zuiderkwast:
*Daily / test-sanitizer-undefined* doesn't seem to run, but I've verified locally with the undefined santitizer.

madolson (approved):

PR #2127, opened 2025-05-23
Slot-level memory metrics POC

hwware:


madolson:
This implementation will have a high performance impact when the objects are large (millions of fields). We also want to see the POC for the sparse data structures keeping track of their underlying memory usage, so updates to the data structures happen in O(1) time.

PR #2129, opened 2025-05-23
Combine range element ranks calculation with range elements search to improve zcount performance

rainsupreme:
Currently in ZCOUNT we search the sorted list twice to locate the range endpoints and then we use an extra 2 searches to take each point rank.
This PR introduce 2 modifications:
1. Calculate the rank while searching each of the range points, thus avoid extra rank calculations
2. polish the code by removing the extra implementation for GetRank which was kept in order to avoid zcount performance degradation.

Following tests this PR seems to improve ZCOUNT performance by ~10%

### DIFF
--- a/src/t_zset.c
+++ b/src/t_zset.c
@@ -529,12 +529,11 @@ static unsigned long zslDeleteRangeByRank(zskiplist *zsl, unsigned int start, un
 /* Find the rank for a specific skiplist member node. Counts nodes after the one
  * specified and subtracts from list length. Note that rank is 1-based.  */
 static unsigned long zslGetRank(zskiplist *zsl, const zskiplistNode *node) {
-    int highest_node_span = zslGetNodeHeight(node) - 1;
-    unsigned long count_after_node = zslGetNodeSpanAtLevel(node, highest_node_span);
-    while (node->level[highest_node_span].forward) {
-        node = node->level[highest_node_span].forward;
-        highest_node_span = zslGetNodeHeight(node) - 1;
+    unsigned long count_after_node = 0;
+    while (node) { /* note this is never null the first time */
+        int highest_node_span = zslGetNodeHeight(node) - 1;
         count_after_node += zslGetNodeSpanAtLevel(node, highest_node_span);
+        node = node->level[highest_node_span].forward;
     }
 
     unsigned long rank = zsl->length - count_after_node;
--- a/src/t_zset.c
+++ b/src/t_zset.c
@@ -526,26 +526,19 @@ static unsigned long zslDeleteRangeByRank(zskiplist *zsl, unsigned int start, un
     return removed;
 }
 
-/* Find the rank for a specific skiplist node.
- * Returns 0 when the element cannot be found, rank otherwise.
- * Note that the rank is 1-based due to the span of zsl->header to the
- * first element. */
+/* Find the rank for a specific skiplist member node. Counts nodes after the one
+ * specified and subtracts from list length. Note that rank is 1-based.  */
 static unsigned long zslGetRank(zskiplist *zsl, const zskiplistNode *node) {
-    unsigned long rank = 0;
-
-    /* Count up nodes that come before */
-    zskiplistNode *x = zsl->header;
-    for (int i = zsl->level - 1; i >= 0; i--) {
-        while (zslCompareNodes(x->level[i].forward, node) <= 0) {
-            rank += zslGetNodeSpanAtLevel(x, i);
-            x = x->level[i].forward;
-        }
-
-        if (x == node) {
-            return rank;
-        }
+    int highest_node_span = zslGetNodeHeight(node) - 1;
+    unsigned long count_after_node = zslGetNodeSpanAtLevel(node, highest_node_span);
+    while (node->level[highest_node_span].forward) {
+        node = node->level[highest_node_span].forward;
+        highest_node_span = zslGetNodeHeight(node) - 1;
+        count_after_node += zslGetNodeSpanAtLevel(node, highest_node_span);
     }
-    return 0;
+
+    unsigned long rank = zsl->length - count_after_node;
+    return rank;
 }
 
 /* Finds an element by its rank from start node. The rank argument needs to be 1-based. */
--- a/src/t_zset.c
+++ b/src/t_zset.c
@@ -304,16 +304,13 @@ static void zslDelete(zskiplist *zsl, zskiplistNode *node) {
  * Otherwise the skiplist is modified by removing and re-adding a new
  * element, which is more costly. A pointer to the new node is returned. */
 static zskiplistNode *zslUpdateScore(zskiplist *zsl, zskiplistNode *node, double newscore) {
-    /* Update score in place if it doesn't invalidate correct ordering */
-    const double old_score = node->score;
-    node->score = newscore;
-    if (zslCompareNodes(node->backward, node) < 0 && zslCompareNodes(node->level[0].forward, node) > 0) {
-        /* list ordering is correct: we are able to update in place */
+    /* If the node, after the score update, would be still exactly
+     * at the same position, we can just update the score without
+     * actually removing and re-inserting the element in the skiplist. */
+    if ((node->backward == NULL || node->backward->score < newscore) &&
+        (node->level[0].forward == NULL || node->level[0].forward->score > newscore)) {
+        node->score = newscore;
         return NULL;
-    } else {
-        /* incorrect list ordering is an invalid state - we must restore previous
-         * state and use another method. */
-        node->score = old_score;
     }
 
     /* We need to remove the node from the skiplist and insert a new one */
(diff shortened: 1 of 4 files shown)


src/t_zset.c
```diff
@@ -304,15 +304,16 @@ static void zslDelete(zskiplist *zsl, zskiplistNode *node) {
...
-     * actually removing and re-inserting the element in the skiplist.
-     * (TODO: The check can be extended to check also equality of the
-     * score, but then we'll also need to compare the key order). */
-    if ((node->backward == NULL || node->backward->score < newscore) &&
-        (node->level[0].forward == NULL || node->level[0].forward->score > newscore)) {
-        node->score = newscore;
+    /* Update score in place if it doesn't invalidate correct ordering */
+    const double old_score = node->score;
+    node->score = newscore;
+    if (zslCompareNodes(node->backward, node) < 0 && zslCompareNodes(node->level[0].forward, node) > 0) {
```
ranshid: I am not sure what is the point of this change? is it to make the code cleaner? I am afraid of degradation in case the scores are identical in backward and forward.

ranshid (commented):
Thank you @SoftlyRaining 

1. What kind of tests did you check the performance of? for example does zcount yields better or worse? Initially I did not want to change zcount since I had a theoretical feeling it might be less cache friendly (since we already looked up each of the Nth elements...)
2. In case we do decide to go with this change. maybe a small refactor of the previously known `zslGetRankCountingNodesAfter` to do a "do... while" instead?

src/t_zset.c
rainsupreme: It improves a tiny corner case: In the situation where the score becomes the same as the previous/next item but it is still in the correct order after comparing the member string, it will now avoid removing and reinserting into the skip list. Tbh @ranshid I had thought this was your idea in my PR some time ago but I really don't remember any more 😅 My personal feeling is that it's fine both ways, I just want to satisfy the TODO promise I made back then.

ZADD and ZRANK performance were not affected by this change. Because the keyspace was full throughout my test, every ZADD call was a score update and used this code path, so I am confident there was no degradation. (Or improvement, honestly.)

rainsupreme:
Thanks for taking a look @ranshid ! 😊 

>     1. What kind of tests did you check the performance of? for example does zcount yields better or worse? Initially I did not want to change zcount since I had a theoretical feeling it might be less cache friendly (since we already looked up each of the Nth elements...)

I only tested ZADD and ZRANK, in the same way I benchmarked the 8.1 release. I tested throughput before and after this change, running it for 60 minutes after a 5 minute warmup period. More info: 650 clients, pipelining 4, 64 benchmark threads, 9 io-threads on server, preloaded with a full keyspace of 3 million items. (No score collisions in the preloaded state, but ZADD perf reached a stable state with collisions after the 5 minute warmup.)

> 2.  In case we do decide to go with this change. maybe a small refactor of the previously known zslGetRankCountingNodesAfter to do a "do... while" instead?

Good idea! I was able to tidy up the loop.

src/t_zset.c
ranshid: >It improves a tiny corner case: In the situation where the score becomes the same as the previous/next item but it is still in the correct order after comparing the member string,

I don't recall my old thoughts, but the existing code already handles the score comparison. regarding the string comparison, we will just evaluate string comparison when the string was unchanged for both the updated entry + back + forward, which seems like an overhead IMO. maybe I am/was missing something?

murphyjacob4:
> Initially I did not want to change zcount since I had a theoretical feeling it might be less cache friendly

Do we know if there is any reason we don't just sum up the rank in `zslNthInRange`? It seems we already have to compute it, it's just a matter of returning it to the `zcountCommand` function. This isn't a blocker for this PR, mostly just curious. Maybe I am missing something obvious

I do think we should get some perf data on the ZCOUNT command before switching the algorithm.

> zslGetRank was about 10% slower when used for ZRANK commands

Yeah it seems to me that the cost of counting the nodes after a given element would be roughly `O(log(N-r))` where `r` is the rank, and counting the nodes before a given element would be `log(r)`.

Assuming we would average to `r=N/2`, it seems somewhat arbitrary which one is picked.

I am curious more on the test setup. By full keyspace of 3 million items - you mean one skiplist of 3 million elements? I am assuming ZRANK is a random element in the skiplist. I think sharing the commands you ran and the results would be helpful.

rainsupreme:
> I am curious more on the test setup. By full keyspace of 3 million items - you mean one skiplist of 3 million elements? I am assuming ZRANK is a random element in the skiplist. I think sharing the commands you ran and the results would be helpful.

Yes, it was one skiplist of 3 million elements, and ZRANK was on random elements. There were no score collisions with the way I populated it.

I started Valkey like this:
`valkey-server --save --protected-mode no --daemonize yes --io-threads 9`

I fillled the keyspace sequentially, so there are scores from 0 to 2999999:
`valkey-benchmark -h <ip> -c 650 -P 4 --threads 50 -q --sequential -r 3000000 -n 3000000 -t zadd`
Note: I used my own version of valkey-benchmark #2102 to actually get 3 million elements with this command

I tested ZRANK with random scores:
`valkey-benchmark -h <ip> -r 3000000 -c 650 -P 4 --threads 64 -q -l -n 2000000000 -- ZRANK myzset element:__rand_int__`
(I didn't let it finish, rather I collected the RPS numbers it emits to stdout at 4Hz, ignored the first 5 minutes as warmup, and averaged the subsequent 60 minutes.)

src/t_zset.c
rainsupreme: I'm not sure I understand what you mean, but my change isn't faster and doesn't feel simpler to me, so I'll just revert this bit and remove the TODO. Is that ok with you? :)

murphyjacob4:
> Yes, it was one skiplist of 3 million elements, and ZRANK was on random elements

Thanks! Yeah this makes sense to me, appreciate you documenting the setup. Do we think it is also worth testing on a small skip list (~64 elements) to cover the other extreme?

> Assuming we would average to r=N/2, it seems somewhat arbitrary which one is picked.

Actually, thinking again, the "count before" solution is a bit different. We would on average need to access more nodes, since we "overshoot" the target node at each layer. Since most nodes are at the very bottom layer, this would result in having to do an extra O(logN) comparisons as opposed to the "ideal path" to the node. You could reimplement "count before" to start at the node and go backwards and upwards up the skiplist, and we _should_ see similar performance (although let's not do that, since it is meaningless which is picked)

Either way, great to see your performance testing paid off! Excited to see what we get for ZCOUNT

src/t_zset.c
```diff
@@ -402,22 +396,24 @@ zskiplistNode *zslNthInRange(zskiplist *zsl, zrangespec *range, long n) {
...
                 /* Count the rank of the last element in range. */
-                edge_rank += zslGetNodeSpanAtLevel(x, i);
+                end_rank += zslGetNodeSpanAtLevel(x, i);
                 x = x->level[i].forward;
             }
         }
         /* Check if the range is big enough. */
-        if (edge_rank < -n) return NULL;
+        if (end_rank < -n) return NULL;
+        if (rank) *rank = end_rank + n;
```
ranshid: Maybe, it is possible that after locating the node we would still get an item which is bellow the min score and we do NOT want to return the rank... suggest we move that post the line num 430.

src/t_zset.c
```diff
@@ -360,40 +358,36 @@ int zslIsInRange(zskiplist *zsl, zrangespec *range) {
...
                 /* Count the rank of the last element smaller than the range. */
-                edge_rank += zslGetNodeSpanAtLevel(x, i);
+                start_rank += zslGetNodeSpanAtLevel(x, i);
                 x = x->level[i].forward;
             }
         }
         /* Check if zsl is long enough. */
-        if ((unsigned long)(edge_rank + n) >= zsl->length) return NULL;
+        if ((unsigned long)(start_rank + n) >= zsl->length) return NULL;
+        if (rank) *rank = start_rank + n;
```
ranshid: it is possible we will eventually get an item with score that is more than the max score, so suggest to move that past line 403

ranshid (commented):
This change makes sense. some small comment

ranshid:
>Actually, thinking again, the "count before" solution is a bit different. We would on average need to access more nodes, since we "overshoot" the target node at each layer. Since most nodes are at the very bottom layer, this would result in having to do an extra O(logN) comparisons as opposed to the "ideal path" to the node. You could reimplement "count before" to start at the node and go backwards and upwards up the skiplist, and we should see similar performance (although let's not do that, since it is meaningless which is picked)

@murphyjacob4 not really sure I understand and what you mean by "backward" solution, as in this skiplist implementation the backward might payoff only in some cases where the node happen to be very close to the start of the list (like much closer than the log(n) list element) which I am not sure worth that much investing into as it might be hard to predict that.

murphyjacob4:
> which I am not sure worth that much investing into as it might be hard to predict that.

Sorry to confuse. I was just circling back on the "why is one implementation better than the other" question. Count before and count after should have the same average runtime supposing uniform random querying within the skiplist. I think that the previous implementation of the "count before" was less performant since it was scanning each layer until we were one node past the target node. As opposed to the "count after" which was just traversing from the target node to the list end. "Count before" could be implemented in a similar way by starting at the target node and traversing to list start, which I would guess would be similar performance as the existing "count after".

But as I said in my comment, I don't think we should change the implementation. Just trying to satisfy the "why".

ranshid:
> > which I am not sure worth that much investing into as it might be hard to predict that.
> 
> Sorry to confuse. I was just circling back on the "why is one implementation better than the other" question. Count before and count after should have the same average runtime supposing uniform random querying within the skiplist. I think that the previous implementation of the "count before" was less performant since it was scanning each layer until we were one node past the target node. As opposed to the "count after" which was just traversing from the target node to the list end. "Count before" could be implemented in a similar way by starting at the target node and traversing to list start, which I would guess would be similar performance as the existing "count after".
> 
> But as I said in my comment, I don't think we should change the implementation. Just trying to satisfy the "why".

If I understand, count before cannot do the same as it iterates only the lower layer which is linear. This is why we always use count after

murphyjacob4:
> If I understand, count before cannot do the same as it iterates only the lower layer which is linear. This is why we always use count after

Don't want to spend too much time discussing since it isn't impacting the PR. But I don't think that is how it is currently implemented. The current "count before" is doing a top to bottom layer traversal. So start at top layer, see how far we can go before we overshoot the target node, then go to the node just before that and repeat starting at that node on the next layer down, until we find the node.

https://github.com/valkey-io/valkey/blob/unstable/src/t_zset.c#L532-L548

So something like (if we are looking for D):

```
L3: H--------------------> C (12)                        T
                            |     (check E, but it is too far)
L2: H             B (8)    C (12)  -/->        E (21)    T
                            |                             
L1: H    A (3)    B (8)    C (12) -> D (18)    E (21)    T
```

So every layer, you are touching an extra node, since you need to overshoot the node before we can traverse to the lower layer.

Instead, a better "count before" (since we are doubly linked) is just to traverse the backwards pointers and go to the maximum height of each node, while summing the span:
```
L3: H <--------------------C (12)                        T
                            | 
L2: H             B (8)    C (12)              E (21)    T
                            |                             
L1: H    A (3)    B (8)    C (12) <- D (18)    E (21)    T
```

If I had to guess, this would be the same performance as "count after" for rank calculations. But I am not supposing we implement this, since it is the same :)

We could attempt to use "count before" if we are before some estimated midpoint, and "count after" otherwise, but I don't think it would be a worthwhile performance optimization.

ranshid:
> > If I understand, count before cannot do the same as it iterates only the lower layer which is linear. This is why we always use count after
> 
> Don't want to spend too much time discussing since it isn't impacting the PR. But I don't think that is how it is currently implemented. The current "count before" is doing a top to bottom layer traversal. So start at top layer, see how far we can go before we overshoot the target node, then go to the node just before that and repeat starting at that node on the next layer down, until we find the node.
> 
> https://github.com/valkey-io/valkey/blob/unstable/src/t_zset.c#L532-L548
> 
> So something like (if we are looking for D):
> 
> ```
> L3: H--------------------> C (12)                        T
>                             |     (check E, but it is too far)
> L2: H             B (8)    C (12)  -/->        E (21)    T
>                             |                             
> L1: H    A (3)    B (8)    C (12) -> D (18)    E (21)    T
> ```
> 
> So every layer, you are touching an extra node, since you need to overshoot the node before we can traverse to the lower layer.
> 
> Instead, a better "count before" (since we are doubly linked) is just to traverse the backwards pointers and go to the maximum height of each node, while summing the span:
> 
> ```
> L3: H <--------------------C (12)                        T
>                             | 
> L2: H             B (8)    C (12)              E (21)    T
>                             |                             
> L1: H    A (3)    B (8)    C (12) <- D (18)    E (21)    T
> ```
> 
> If I had to guess, this would be the same performance as "count after" for rank calculations. But I am not supposing we implement this, since it is the same :)
> 
> We could attempt to use "count before" if we are before some estimated midpoint, and "count after" otherwise, but I don't think it would be a worthwhile performance optimization.

@murphyjacob4 That is basically the way we do for GetRank today. I agree we would love to use it for count backward however we do not keep backward link in each height (only forward)

ranshid (approved):
@SoftlyRaining Mostly LGTM. maybe consider fixing the last 2 comments I placed about early returning form the zslNthInRange while we still set the rank?

murphyjacob4:
Sorry if I missed it - but did we ever measure the new ZCOUNT performance?

Perhaps it is more difficult to test with valkey-benchmark

ranshid:
> Sorry if I missed it - but did we ever measure the new ZCOUNT performance?
> 
> Perhaps it is more difficult to test with valkey-benchmark

Yeh. I am waiting on the zcount results. just provided a general approval as I do not expect it to degrade in performance

rainsupreme:
Sorry to leave folks hanging! I was on call while doing this and didn't have time to come back to it until now. I've benchmarked ZCOUNT now, although caveat - on average half the ranges I specify are null results because I can only use simple random indices with valkey-benchmark so half the time start > end, but it's still an apples-to-apples comparison. That aside, ZCOUNT was 10% faster after this change! 😁

unstable | 395991
-- | --
zrank-refactor | 434862

murphyjacob4 (approved):

ranshid (approved):

PR #2131, opened 2025-05-23
Correct VM_Yield timing and deflake blockedclient.tcl

PingXie:
Previously, the `next_yield_time` for a yielding module command was only correctly calculated for the initial yield, taking into account whether the server was loading an RDB. Subsequent yields incorrectly defaulted to using `server.hz` for the calculation, regardless of the loading state.

This bug had two consequences:
1. Incorrect Yield Timing: If not loading, the module command would yield based on `server.hz` instead of the configured `busy_reply_threshold`. This could lead to the main thread being blocked for significantly longer or shorter periods than intended between yields.
2. Test Flakiness: In the `blockedclient.tcl` test, this incorrect timing meant that new commands from other clients were delayed until the current, potentially overly long, yield cycle completed. With `busy_time_limit` set to 50s, this significantly increased test duration and risked timeouts, causing flakiness.

This commit refactors the `next_yield_time` calculation into a new static function `computeNextYieldTime` and ensures it's called both during the initial context creation and within `VM_Yield` after each yield. This guarantees that `next_yield_time` is always predicated on the current `server.loading` status.

The `busy_time_limit` in `blockedclient.tcl` has also been reduced from 50s to 5s. With the primary bug addressed, this shorter timeout is more appropriate and helps the test complete faster, reducing the risk of timeouts.

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -943,6 +943,20 @@ ValkeyModuleCtx *moduleAllocateContext(void) {
     return (ValkeyModuleCtx *)zcalloc(sizeof(ValkeyModuleCtx));
 }
 
+static void computeNextYieldTime(ValkeyModuleCtx *ctx) {
+    /* in loading we depend on the server hz, but in other cases we also wait
+     * for busy_reply_threshold.
+     * Note that in theory we could have started processing BUSY_MODULE_YIELD_EVENTS
+     * sooner, and only delay the processing for clients till the busy_reply_threshold,
+     * but this carries some overheads of frequently marking clients with BLOCKED_POSTPONE
+     * and releasing them, i.e. if modules only block for short periods. */
+    if (server.loading) {
+        ctx->next_yield_time = getMonotonicUs() + 1000000 / server.hz;
+    } else {
+        ctx->next_yield_time = getMonotonicUs() + server.busy_reply_threshold * 1000;
+    }
+}
+
 /* Create a module ctx and keep track of the nesting level.
  *
  * Note: When creating ctx for threads (VM_GetThreadSafeContext and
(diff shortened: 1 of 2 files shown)


src/module.c
```diff
@@ -943,6 +943,20 @@ ValkeyModuleCtx *moduleAllocateContext(void) {
     return (ValkeyModuleCtx *)zcalloc(sizeof(ValkeyModuleCtx));
 }
 
+static void computeNextYieldTime(ValkeyModuleCtx *ctx) {
+    /* in loading we depend on the server hz, but in other cases we also wait
```
madolson: Looks like you dropped a line from the comment, was that intentional? It actually maybe just looks weird because the I isn't capitalized so it looks like it was partially copied.

src/module.c
```diff
@@ -2544,7 +2549,7 @@ void VM_Yield(ValkeyModuleCtx *ctx, int flags, const char *busy_reply) {
         }
 
         /* decide when the next event should fire. */
-        ctx->next_yield_time = now + 1000000 / server.hz;
+        computeNextYieldTime(ctx);
```
```suggestion
        ctx->next_yield_time = computeNextYieldTime();
```
It seems odd that this function both computes and sets it on a context, especially since it's not dependent on the context. Not a strong opinion though.

madolson (approved):
Not a strong opinion about the API, but I think having the API just return the next yield time is a bit clearer.

src/module.c
PingXie: good point. I prefer "function" over "method" too.

src/module.c
PingXie: it is intentional. the line I dropped is about the caller.

```c
    /* Calculate the initial yield time for long blocked contexts. */
```

madolson (approved):

PingXie:
External server tests were broken between https://github.com/valkey-io/valkey/actions/runs/15080758945 and https://github.com/valkey-io/valkey/actions/runs/15059373963. Will merge this PR and investigate the external server tests failure separately.

PingXie:
External server tests flakiness should be fixed by #2138.

PR #2133, opened 2025-05-24
Fix the GNU IFUNC error in alpine (non-glibc) environment

zhulipeng:
Follow up of https://github.com/valkey-io/valkey/pull/2099

Fix https://github.com/valkey-io/valkey/actions/runs/15221314860/job/42817270745,
https://github.com/valkey-io/valkey/actions/runs/15221314860/job/42817271735

### DIFF
--- a/src/config.h
+++ b/src/config.h
@@ -391,10 +391,4 @@ void setcpuaffinity(const char *cpulist);
 #define ATTRIBUTE_TARGET_AVX512
 #endif
 
-#if defined(__linux__) && (defined(__GNUC__) && (__GNUC__ > 4) || defined(__clang__) && (__clang_major__) > 5)
-#define HAVE_IFUNC 1
-#else
-#define HAVE_IFUNC 0
-#endif
-
 #endif
--- a/src/config.h
+++ b/src/config.h
@@ -391,4 +391,10 @@ void setcpuaffinity(const char *cpulist);
 #define ATTRIBUTE_TARGET_AVX512
 #endif
 
+#if defined(__linux__) && defined(__GLIBC__) && (defined(__GNUC__) && (__GNUC__ > 4) || defined(__clang__) && (__clang_major__) > 5)
+#define HAVE_IFUNC 1
+#else
+#define HAVE_IFUNC 0
+#endif
+
 #endif
--- a/src/Makefile
+++ b/src/Makefile
@@ -150,6 +150,21 @@ ifneq (,$(findstring ppc,$(uname_M)))
 endif
 endif
 
+# Only enable ifunc on Linux x86_64 with glibc
+ifeq ($(uname_S),Linux)
+ifeq ($(uname_M),x86_64)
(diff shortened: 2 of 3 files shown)


Fusl (changes requested):
Should change the name of the temp file from `foo` to something more descriptive and exclude it from `.gitignore` or use an actual temporary file in `/tmp` using a random file suffix if we can do that logic in` Makefile`s.

src/config.h
```diff
@@ -391,10 +391,4 @@ void setcpuaffinity(const char *cpulist);
 #define ATTRIBUTE_TARGET_AVX512
 #endif
 
-#if defined(__linux__) && (defined(__GNUC__) && (__GNUC__ > 4) || defined(__clang__) && (__clang_major__) > 5)
```
zuiderkwast, line 394: Can't we just add `defined(__GLIBC__)` to this check? Like we did here:

* #1707

One more thing: Does ifunc require linux or just glibc? It's possible to run freebsd with glibc. There's a debian variant with this combo.

zhulipeng: The ifunc feature requires ELF. Even though BSD systems have glibc, I'm not sure if there are any gaps between the loader and runtime linker compared to Linux.

zuiderkwast (approved):
Lgtm. If we have linux and glibc, we should be safe. We have cmake and make parallel builds so it's better to do it in the c file only.

r-devulap (commented):
Might be an overkill, but if we want a more robust approach, we could add a check to Makefile, something along the lines: 

```
# In your Makefile

# Use a temporary file to test if 'ifunc' is supported
IFUNC_TEST := \
'#include <stdio.h>\n'\
'__attribute__((ifunc("resolve_func"))) int foo();\n'\
'int real_foo() { return 42; }\n'\
'int (*resolve_func(void))(void) { return real_foo; }\n'\
'int main() { return foo(); }\n'

check_ifunc_support:
	echo -e $(IFUNC_TEST) > test_ifunc.c && \
	$(CC) $(CFLAGS) test_ifunc.c -o test_ifunc 2>/dev/null && \
	echo "IFUNC_SUPPORTED = 1" > ifunc.mk || \
	echo "IFUNC_SUPPORTED = 0" > ifunc.mk
	rm -f test_ifunc.c test_ifunc

-include ifunc.mk

# Now you can use $(IFUNC_SUPPORTED) in your Makefile
ifeq ($(IFUNC_SUPPORTED),1)
  $(info IFUNC is supported)
else
  $(info IFUNC is NOT supported)
endif
```

Fusl:
The first version of this PR already had a very similar approach:

https://github.com/valkey-io/valkey/pull/2133/commits/72effba0c05ca92048aa1789fb82c1ff0ed34e33#diff-3e2513390df543315686d7c85bd901ca9256268970032298815d2f893a9f0685R153-R166

To simplify things we decided not to add a compiler check step but use one of the existing flags instead.

r-devulap:
Ah, apologies. I missed that commit :)

PR #2138, opened 2025-05-26
Fix fragile test in lazyfree in external test

PingXie:
Make sure the previous test is really done before sampling `used_memory`
in external server case.


enjoy-binbin (commented):
this begin with a new `start_server`, do we have the lazyfree issue in a new server?

PingXie:
I don't think so. This is an external server test only issue.

enjoy-binbin (approved):
ok, i see. LGTM

PR #2141, opened 2025-05-27
Update redis to server in internal variable names in benchmark source code

hwware:
Update Redis legacy word in valkey-benchmark tool: rename variable names redis_config to server_config.

### DIFF
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -126,7 +126,7 @@ static struct config {
     readFromReplica read_from_replica;
     int cluster_node_count;
     struct clusterNode **cluster_nodes;
-    struct serverConfig *redis_config;
+    struct serverConfig *server_config;
     struct hdr_histogram *latency_histogram;
     struct hdr_histogram *current_sec_latency_histogram;
     _Atomic int is_fetching_slots;
@@ -183,7 +183,7 @@ typedef struct clusterNode {
     int *updated_slots;      /* Used by updateClusterSlotsConfiguration */
     int updated_slots_count; /* Used by updateClusterSlotsConfiguration */
     int replicas_count;
-    struct serverConfig *redis_config;
+    struct serverConfig *server_config;
 } clusterNode;
 
 typedef struct serverConfig {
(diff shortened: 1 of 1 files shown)


zuiderkwast (dismissed):
Thanks. I saw some more in tls.c recently:

https://github.com/valkey-io/valkey/blob/unstable/src/tls.c#L53-L56

src/valkey-benchmark.c
```diff
@@ -126,7 +126,7 @@ static struct config {
     readFromReplica read_from_replica;
     int cluster_node_count;
     struct clusterNode **cluster_nodes;
-    struct serverConfig *redis_config;
+    struct serverConfig *server_config;
```
zuiderkwast, line 129: Hey, this is not the server! It's the benchmark tool. :laughing:

Maybe we should call it `benchmark_config`, or just `config`?

zuiderkwast: But the struct is called serverConfig. I guess we can also rename that..

hwware:
> Thanks. I saw some more in tls.c recently:
> 
> https://github.com/valkey-io/valkey/blob/unstable/src/tls.c#L53-L56

I just find them when I play valkey-benchmark tool, maybe there are more in our repo.

src/valkey-benchmark.c
hwware: > But the struct is called serverConfig. I guess we can also rename that..

Yes, Then I think the struct name is not proper, I need to update it as well. It should name as struct benchmarkConfig

src/valkey-benchmark.c
```diff
@@ -183,13 +183,13 @@ typedef struct clusterNode {
...
-    struct serverConfig *redis_config;
+    struct benchmarkConfig *benchmark_config;
 } clusterNode;
 
-typedef struct serverConfig {
+typedef struct benchmarkConfig {
     sds save;
     sds appendonly;
-} serverConfig;
+} benchmarkConfig;
```
zuiderkwast: Wait a minute. Actually `save` and `appendonly` sounds like valkey server configs. :thinking:

Is this actually valkey server configs, the benchmark tool gets this info from the valkey server?

There is another struct which is just called `static struct config { ... }` and it has all the benchmark configs like pipeline, datasize, etc. from the command line options.

Maybe the first commit was correct? :man_facepalming:

src/valkey-benchmark.c
zuiderkwast: Hmmmm, now, I'm not sure. :thinking: There is also another `struct config` in this file. It seems that one has the benchmark options. The `serverConfig` only has two fields: save and appendonly. I think we need to look at what this struct is actually used for. Maybe it is actually server config? :laughing:

src/valkey-benchmark.c
hwware: Let me take sometime to check which they represent.

murphyjacob4: Looks like these are from the server itself: https://github.com/valkey-io/valkey/blob/73696bf6e2cf754acc3ec24eaf9ca6b879bfc5d7/src/valkey-benchmark.c#L310-L311

`serverConfig` makes more sense to me with this context

zuiderkwast: Yes so let's revert the second commit. We can merge the first commit.

zuiderkwast (approved):

PR #2148, opened 2025-05-28
[draft] Replace void** with opaque hashtablePosition

rainsupreme:
This is still a draft - I'm currently running benchmarks to ensure there are no regressions.

This neatly avoids void** references to internal hashtable data, instead using hashtablePosition to replace elements when needed. The result is safer and more readable. (At present I've not changed the hashtableScanDefrag callback signature though.)

### DIFF
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -1318,17 +1318,25 @@ int hashtableFind(hashtable *ht, const void *key, void **found) {
     }
 }
 
-/* Returns a pointer to where an entry is stored within the hash table, or
- * NULL if not found. To get the entry, dereference the returned pointer. The
- * pointer can be used to replace the entry with an equivalent entry (same
- * key, same hash value), but note that the pointer may be invalidated by future
- * accesses to the hash table due to incermental rehashing, so use with care. */
-void **hashtableFindRef(hashtable *ht, const void *key) {
-    if (hashtableSize(ht) == 0) return NULL;
+/* Returns 1 when found, 0 when not found. If found, the position of the entry
+ * is returned via the position argument, which can be used to access or replace
+ * the entry with one with same key and hash. The position is for immediate use
+ * and may be invalidated by the next hashtable operation that triggers
+ * incremental rehashing or modifies the table. This method does not pause
+ * rehashing. */
+int hashtableFindPosition(hashtable *ht, const void *key, hashtablePosition *pos) {
+    position *p = positionFromOpaque(pos);
+    if (hashtableSize(ht) == 0) return 0;
     uint64_t hash = hashKey(ht, key);
-    int pos_in_bucket = 0;
-    bucket *b = findBucket(ht, hash, key, &pos_in_bucket, NULL);
-    return b ? &b->entries[pos_in_bucket] : NULL;
+    int pos_in_bucket, table_index;
+    bucket *b = findBucket(ht, hash, key, &pos_in_bucket, &table_index);
+    if (b != NULL) {
+        p->bucket = b;
+        p->pos_in_bucket = pos_in_bucket;
+        p->table_index = table_index;
+    }
+    
+    return b != NULL;
 }
 
 /* Adds an entry. Returns 1 on success. Returns 0 if there was already an entry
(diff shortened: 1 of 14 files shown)


src/hashtable.c
```diff
@@ -1318,17 +1318,25 @@ int hashtableFind(hashtable *ht, const void *key, void **found) {
...
-    return b ? &b->entries[pos_in_bucket] : NULL;
+    int pos_in_bucket, table_index;
+    bucket *b = findBucket(ht, hash, key, &pos_in_bucket, &table_index);
+    if (b != NULL) {
+        p->bucket = b;
+        p->pos_in_bucket = pos_in_bucket;
+        p->table_index = table_index;
+    }
+    
+    return b != NULL;
```
JimB123, line 1319: I know this works.  But it drives me nuts.  Regardless of implementation details, booleans should not be treated as integers.

Here you have a boolean expression... and are assigning it to be returned in an integer value.  Why not make the return value a bool?  (We allow that now.)

If you really want an integer, make the conversion explicit, like this:
```
return (b != NULL) ? 1 : 0;
```

src/hashtable.c
```diff
@@ -1318,17 +1318,25 @@ int hashtableFind(hashtable *ht, const void *key, void **found) {
...
-void **hashtableFindRef(hashtable *ht, const void *key) {
-    if (hashtableSize(ht) == 0) return NULL;
+/* Returns 1 when found, 0 when not found. If found, the position of the entry
+ * is returned via the position argument, which can be used to access or replace
+ * the entry with one with same key and hash. The position is for immediate use
+ * and may be invalidated by the next hashtable operation that triggers
+ * incremental rehashing or modifies the table. This method does not pause
+ * rehashing. */
+int hashtableFindPosition(hashtable *ht, const void *key, hashtablePosition *pos) {
+    position *p = positionFromOpaque(pos);
```
JimB123, line 1306: If you really want to ensure safety, you should:
* Add an "epoch" counter on the table.  Just an unsigned integer that increments on every modification.  It doesn't matter if it wraps around.
* Record the current epoch in the hashtablePosition when it is created.
* Validate that the epoch hasn't changed (assert) when the position is used.

src/hashtable.c
madolson: We do allow bools too, you could just make it a bool.

zuiderkwast: There are a bunch of hashtable functions that return 1 or 0. This was done before we started to accept bool. We can change all of them to bool.

I'm fine with just doing old-fashion C style too though. Apparently, Brian Kernighan and Dennis Ritchie didn't think bool was necessary. All the logic operators and conditional constructs are defined on integer values. For example, `!=` returns 0 or 1, `if` checks for non-zero or zero, etc. C is not really type safe anyway. You can crash in many ways.

zuiderkwast:
Making types more opaque can improve things, but we have many problems that are worse than this IMO. All the functions in networking.c and server.c and their dependency on global state is possibly more error-prone than this stuff. Refactoring some of that and making it possible to unit-test these parts of the system could possible add more value.

The gigantic server.h is a dependency monster. Whenever a unit includes it, we have a cross-dependency on the entire server. We could move prototypes and type definitions from server.h to new header files that would correspond to a C file, for example move prototypes related to db.c to a new header db.h, t_string.h for t_string.c, and so on.

zuiderkwast:
> Making types more opaque can improve things, but we have many problems that are worse than this IMO. All the functions in networking.c and server.c and their dependency on global state is possibly more error-prone than this stuff. Refactoring some of that and making it possible to unit-test these parts of the system could possible add more value.
> 
> The gigantic server.h is a dependency monster. Whenever a unit includes it, we have a cross-dependency on the entire server. We could move prototypes and type definitions from server.h to new header files that would correspond to a C file, for example move prototypes related to db.c to a new header db.h, t_string.h for t_string.c, and so on.

src/hashtable.c
rainsupreme: Bool makes it more immediately obvious how to interpret the result and makes for cleaner code, so I switched hashtable over to bool. (I was on call and needed a low-context easy-to-pause task to do between other work 😅)

rainsupreme:
Well, I did some benchmarks and it's not what I was hoping for. Is it from adding epoch math and checks? 🤔 Moreover, can it be fixed or does the change need to be dropped?

  | unstable | this change | % change
-- | -- | -- | --
set | 383230.5 | 367985.5719 | -4%
get | 702641.4 | 689760.9386 | -2%

src/hashtable.c
```diff
@@ -319,8 +320,10 @@ static_assert(sizeof(hashtableIterator) >= sizeof(iter),
 /* Position, used by some hashtable functions such as two-phase insert and delete. */
 typedef struct {
     bucket *bucket;
-    uint16_t pos_in_bucket;
-    uint16_t table_index;
+    uint32_t *current_epoch; /* Pointer to the epoch of the hashtable. */
```
JimB123, line 323: It's interesting that the `position` doesn't include a reference to the `hashtable`.  So there's no mechanism to validate that a given position is actually associated with a given hashtable.

So, I see what you're doing here, keeping a reference to the epoch so that you can compare.  But wouldn't it be more useful to keep a reference to the hashtable instead?

rainsupreme: Perhaps, but do I need access to the hashtable? I don't see how it would be possible to use a position with a different hashtable, given that it intrinsically points to a specific hashtable's data and is effectively read-only by being opaque 🤔

PR #2132, opened 2025-05-23
Free module context even if there was no content written in auxsave2

murphyjacob4:
When a module saves to the Aux metadata with aux_save2 callback, the module context is not freed if the module didn't save anything in the callback. 

https://github.com/valkey-io/valkey/blob/8.1.1/src/rdb.c#L1277-L1284. Note that we return 0, however we should be doing the cleanup done on https://github.com/valkey-io/valkey/blob/8.1.1/src/rdb.c#L1300-L1303 still.

Fixes #2125

PR #2150, opened 2025-05-29
Update tests to catch module context leaks if using aux save/load 

murphyjacob4:
See https://github.com/valkey-io/valkey/pull/2132 for the fix.

This just ensures that we always create context in all branches of aux_load and aux_save

### DIFF
--- a/tests/modules/testrdb.c
+++ b/tests/modules/testrdb.c
@@ -72,8 +72,11 @@ void *testrdb_type_load(ValkeyModuleIO *rdb, int encver) {
     ValkeyModuleString *str = ValkeyModule_LoadString(rdb);
     float f = ValkeyModule_LoadFloat(rdb);
     long double ld = ValkeyModule_LoadLongDouble(rdb);
+    /* Context creation is part of the test. Creating the context will force the
+     * core to allocate a context, which needs to be cleaned up when the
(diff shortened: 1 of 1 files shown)


enjoy-binbin (approved):

tests/modules/testrdb.c
```diff
@@ -98,6 +98,8 @@ void testrdb_aux_save(ValkeyModuleIO *rdb, int when) {
     if (!(conf_aux_count & CONF_AUX_OPTION_BEFORE_KEYSPACE)) assert(when == VALKEYMODULE_AUX_AFTER_RDB);
     if (!(conf_aux_count & CONF_AUX_OPTION_AFTER_KEYSPACE)) assert(when == VALKEYMODULE_AUX_BEFORE_RDB);
     assert(conf_aux_count!=CONF_AUX_OPTION_NO_AUX);
+    ValkeyModuleCtx *ctx = ValkeyModule_GetContextFromIO(rdb);
+    VALKEYMODULE_NOT_USED(ctx);
```
zuiderkwast, line 108: We could add a code comment explaining why we do this, that it has a side effect. Otherwise it looks like unnecessary code.

murphyjacob4: Makes sense. Done

PR #2152, opened 2025-05-29
Unify behavior of CLIENT REPLY in multi with other NO_MULTI commands

zuiderkwast:
CLIENT REPLY is forbidden inside transactions, but the error behavior is different to other commands that are forbidden in transactions (such as WATCH, SAVE or nested MULTI).

* The error message is different.
* The error within the transaction doesn't cause the EXEC to return error EXECABORT as it should.

This change unifies the behavior and implementation with the other commands by removing the special handling and instead just relying on the NO_MULTI command flag.

Follow-up of #1966.


enjoy-binbin (approved):

sarthakaggarwal97 (approved):
nice, I wasn't aware of the NO_MULTI when I made the change. thank you!

madolson (approved):
Sure

PR #2154, opened 2025-05-29
Limiting the new reconnections for failed nodes

### LINKED ISSUE #2122: [NEW] Excessive connection attempts to failed nodes in clusterCron() cause CPU overhead

In `clusterCron()`, the function `clusterNodeCronHandleReconnect()` attempts to reconnect to nodes whose `link == NULL` on every iteration (10 times per second by default). This leads to repeated connection attempts to nodes that are unreachable or in a failed (`PFAIL` or `FAIL`) state.

When there are multiple failing nodes, this retry logic results in:

- High CPU usage due to frequent `connConnect()` calls.
- Excessive memory allocation/free churn from repeatedly creating and destroying `clusterLink` objects.

*Profiles of the connected nodes:*

Image

---

*Engine CPU / Used Memory metrics:*

The initial spike is during the time when large numbers of primary nodes (n/2 - 1) were killed, and subsequent increase in the memory seems to be due to frequent reconnect attempts based on the above profile.

Image

---

Once the failed nodes are resumed, the compute is back to its steady state levels.

A backoff mechanism to avoid so many reconnects with the failed nodes.

sarthakaggarwal97:
In cluster mode, the node attempts to reconnect to nodes where `link == NULL` during each execution cycle, which occurs every 100 milliseconds by default. This behavior results in continuous reconnection attempts to nodes that are unreachable or in a failed state (PFAIL or FAIL)

This PR addresses an issue where the system aggressively attempts to reconnect to failed nodes, which can lead to resource exhaustion and potential instability. This change limits the number of attempts we make per failed node. 

The PR improves engine CPU of the P99 nodes (20-30 nodes in a cluster) in the by **35%,** P90 (200-300 nodes) and Avg (across all nodes) by **10%** when there are failed nodes in the cluster.

Screenshot 2025-06-11 at 2 20 11 PM

Screenshot 2025-06-12 at 7 16 48 PM

Resolves #2122

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -5397,7 +5397,8 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t now, int *
     }
 
     if (node->link == NULL) {
-        if (!node->inbound_link && (now - node->outbound_link_attempt_time < server.cluster_node_timeout / NODE_CONNECTION_RETRIES_PER_TIMEOUT && *cluster_conn_attempts > max_conn_attempts)) {
+        mstime_t reconnect_interval = server.cluster_node_timeout / 2;
+        if (!node->inbound_link && (now - node->outbound_link_attempt_time < reconnect_interval / NODE_CONNECTION_RETRIES_PER_TIMEOUT && *cluster_conn_attempts > max_conn_attempts)) {
             return 1;
         }
         node->outbound_link_attempt_time = now;
@@ -5453,15 +5454,15 @@ static void clusterNodeCronFreeLinkOnBufferLimitReached(clusterNode *node) {
  *
  * We want to guarantee that every node is contacted twice within cluster node timeout.
  */
-static long long maxConnectionAttemptsPerCron(const int nodes, const long long timeout_ms) {
-    if (nodes <= 0 || timeout_ms <= 0)
+static long long maxConnectionAttemptsPerCron(void) {
+    long long reconnect_interval = server.cluster_node_timeout / 2;
+    if (dictSize(server.cluster->nodes) <= 0 || reconnect_interval <= 0)
         return 0;
     /*
      * We run the cron loop every 100 ms.  To reach 100 % of the nodes
-     * within the timeout, we need: ceil(nodes * 100 / timeout_ms)
-     * We use (a + b − 1) / b to give us the integer-ceil without using floating-point.
+     * within the timeout, we need: ceil(nodes * 100 / reconnect_interval)
      */
-    const long long min_nodes_for_coverage = (nodes * 100 + timeout_ms - 1) / timeout_ms;
+    const long long min_nodes_for_coverage = (dictSize(server.cluster->nodes) * CLUSTER_CRON_PERIOD_MS + reconnect_interval - 1) / reconnect_interval;
     /*
      * Increase the coverage budget so each node can be probed 10 times
      * inside the timeout.
@@ -5490,7 +5491,7 @@ void clusterCron(void) {
     server.cluster->stats_pfail_nodes = 0;
     /* Run through some of the operations we want to do on each cluster node. */
     di = dictGetSafeIterator(server.cluster->nodes);
-    const long long max_conn_attempts = maxConnectionAttemptsPerCron(dictSize(server.cluster->nodes), server.cluster_node_timeout);
+    const long long max_conn_attempts = maxConnectionAttemptsPerCron();
     while ((de = dictNext(di)) != NULL) {
         clusterNode *node = dictGetVal(de);
         /* We free the inbound or outboud link to the node if the link has an
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -5398,7 +5398,7 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t now, long 
 
     if (node->link == NULL) {
         mstime_t reconnect_interval = server.cluster_node_timeout / 2;
-        /* Skip this outbound connection attempt when all three are true:
+        /* Skip this outbound connection attempt when all these conditions are true:
          *  1. No inbound link from the peer exists.
          *  2. The back‑off window since the last try is still active
          *  3. The cluster has already exceeded its retry budget for this cron cycle
@@ -5453,25 +5453,19 @@ static void clusterNodeCronFreeLinkOnBufferLimitReached(clusterNode *node) {
     freeClusterLinkOnBufferLimitReached(node->inbound_link);
 }
 
-/**
- * Compute the maximum number of connection attempts the cluster-cron
+/* Compute the maximum number of connection attempts the clusterCron
  * loop should schedule in a single cron.
  *
- * We want to guarantee that every node is contacted 10 times within node timeout.
- */
+ * We want to guarantee that every node is contacted 10 times within node timeout. */
 static long long maxConnectionAttemptsPerCron(void) {
     long long reconnect_interval = server.cluster_node_timeout / 2;
     if (reconnect_interval <= 0)
         return 0;
-    /*
-     * We run the cron loop every 100 ms.  To reach 100 % of the nodes
-     * within the timeout, we need: ceil(nodes * 100 / reconnect_interval)
-     */
+    /* We run the cron loop every 100 ms.  To reach 100 % of the nodes
+     * within the timeout, we need: ceil(nodes * 100 / reconnect_interval) */
     const long long min_nodes_for_coverage = dictSize(server.cluster->nodes) * CLUSTER_CRON_PERIOD_MS / reconnect_interval;
-    /*
-     * Increase the coverage budget so each node can be probed 10 times
-     * inside the timeout.
-     */
+    /* Increase the coverage budget so each node can be probed 10 times
+     * inside the timeout. */
     return min_nodes_for_coverage * NODE_CONNECTION_RETRIES_PER_TIMEOUT;
 }
 
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -5363,6 +5363,8 @@ static int nodeExceedsHandshakeTimeout(clusterNode *node, mstime_t now) {
     return now - node->ctime > getHandshakeTimeout() ? 1 : 0;
 }
 
+#define NODE_CONNECTION_RETRIES_PER_TIMEOUT 10
+
 /* Check if the node is disconnected and re-establish the connection.
  * Also update a few stats while we are here, that can be used to make
  * better decisions in other part of the code. */
@@ -5395,7 +5397,7 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t now, int *
     }
 
     if (node->link == NULL) {
-        if (!node->inbound_link && (now - node->outbound_link_attempt_time < server.cluster_node_timeout / 10 && *cluster_conn_attempts > max_conn_attempts)) {
+        if (!node->inbound_link && (now - node->outbound_link_attempt_time < server.cluster_node_timeout / NODE_CONNECTION_RETRIES_PER_TIMEOUT && *cluster_conn_attempts > max_conn_attempts)) {
             return 1;
         }
         node->outbound_link_attempt_time = now;
@@ -5461,10 +5463,10 @@ static long long maxConnectionAttemptsPerCron(const int nodes, const long long t
      */
     const long long min_nodes_for_coverage = (nodes * 100 + timeout_ms - 1) / timeout_ms;
     /*
-     * Double the coverage budget so each node can be probed twice
-     * inside the timeout, improving resilience to packet loss.
+     * Increase the coverage budget so each node can be probed 10 times
+     * inside the timeout.
      */
-    return min_nodes_for_coverage * 10;
+    return min_nodes_for_coverage * NODE_CONNECTION_RETRIES_PER_TIMEOUT;
 }
 
 /* This is executed 10 times every second */
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -5398,6 +5398,11 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t now, int *
 
     if (node->link == NULL) {
         mstime_t reconnect_interval = server.cluster_node_timeout / 2;
+        /* Skip this outbound connection attempt when all three are true:
+         *  1. No inbound link from the peer exists.
+         *  2. The back‑off window since the last try is still active
+         *  3. The cluster has already exceeded its retry budget for this cron cycle
+         */
         if (!node->inbound_link && (now - node->outbound_link_attempt_time < reconnect_interval / NODE_CONNECTION_RETRIES_PER_TIMEOUT && *cluster_conn_attempts > max_conn_attempts)) {
             return 1;
         }
@@ -5452,7 +5457,7 @@ static void clusterNodeCronFreeLinkOnBufferLimitReached(clusterNode *node) {
  * Compute the maximum number of connection attempts the cluster-cron
  * loop should schedule in a single cron.
  *
- * We want to guarantee that every node is contacted twice within cluster node timeout.
+ * We want to guarantee that every node is contacted 10 times within node timeout.
  */
 static long long maxConnectionAttemptsPerCron(void) {
     long long reconnect_interval = server.cluster_node_timeout / 2;
@@ -5462,7 +5467,7 @@ static long long maxConnectionAttemptsPerCron(void) {
      * We run the cron loop every 100 ms.  To reach 100 % of the nodes
      * within the timeout, we need: ceil(nodes * 100 / reconnect_interval)
      */
-    const long long min_nodes_for_coverage = (dictSize(server.cluster->nodes) * CLUSTER_CRON_PERIOD_MS + reconnect_interval - 1) / reconnect_interval;
+    const long long min_nodes_for_coverage = ceil(dictSize(server.cluster->nodes) * CLUSTER_CRON_PERIOD_MS / reconnect_interval);
     /*
      * Increase the coverage budget so each node can be probed 10 times
      * inside the timeout.
@@ -5502,7 +5507,6 @@ void clusterCron(void) {
          */
         if (!server.debug_cluster_disable_reconnection && clusterNodeCronHandleReconnect(node, now, &cluster_node_conn_attempts, max_conn_attempts)) continue;
     }
-    cluster_node_conn_attempts = 0;
     dictReleaseIterator(di);
 
     /* Ping some random node 1 time every 10 iterations, so that we usually ping
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -5368,7 +5368,7 @@ static int nodeExceedsHandshakeTimeout(clusterNode *node, mstime_t now) {
 /* Check if the node is disconnected and re-establish the connection.
  * Also update a few stats while we are here, that can be used to make
  * better decisions in other part of the code. */
-static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t now, int *cluster_conn_attempts, const long long max_conn_attempts) {
+static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t now, long long *cluster_conn_attempts) {
     /* Not interested in reconnecting the link with myself or nodes
      * for which we have no address. */
     if (node->flags & (CLUSTER_NODE_MYSELF | CLUSTER_NODE_NOADDR)) return 1;
@@ -5403,11 +5403,11 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t now, int *
          *  2. The back‑off window since the last try is still active
          *  3. The cluster has already exceeded its retry budget for this cron cycle
          */
-        if (!node->inbound_link && (now - node->outbound_link_attempt_time < reconnect_interval / NODE_CONNECTION_RETRIES_PER_TIMEOUT && *cluster_conn_attempts > max_conn_attempts)) {
+        if (!node->inbound_link && (now - node->outbound_link_attempt_time < reconnect_interval / NODE_CONNECTION_RETRIES_PER_TIMEOUT && *cluster_conn_attempts == 0)) {
             return 1;
         }
         node->outbound_link_attempt_time = now;
-        (*cluster_conn_attempts)++;
+        (*cluster_conn_attempts)--;
         clusterLink *link = createClusterLink(node);
         link->conn = connCreate(connTypeOfCluster());
         connSetPrivateData(link->conn, link);
@@ -5467,7 +5467,7 @@ static long long maxConnectionAttemptsPerCron(void) {
      * We run the cron loop every 100 ms.  To reach 100 % of the nodes
      * within the timeout, we need: ceil(nodes * 100 / reconnect_interval)
      */
-    const long long min_nodes_for_coverage = ceil(dictSize(server.cluster->nodes) * CLUSTER_CRON_PERIOD_MS / reconnect_interval);
+    const long long min_nodes_for_coverage = dictSize(server.cluster->nodes) * CLUSTER_CRON_PERIOD_MS / reconnect_interval;
     /*
      * Increase the coverage budget so each node can be probed 10 times
      * inside the timeout.
@@ -5486,8 +5486,6 @@ void clusterCron(void) {
     mstime_t min_pong = 0, now = mstime();
     clusterNode *min_pong_node = NULL;
     static unsigned long long iteration = 0;
-    int cluster_node_conn_attempts = 0;
-
     iteration++; /* Number of times this function was called so far. */
 
     clusterUpdateMyselfHostname();
@@ -5496,7 +5494,7 @@ void clusterCron(void) {
     server.cluster->stats_pfail_nodes = 0;
     /* Run through some of the operations we want to do on each cluster node. */
     di = dictGetSafeIterator(server.cluster->nodes);
-    const long long max_conn_attempts = maxConnectionAttemptsPerCron();
+    long long cluster_node_conn_attempts = maxConnectionAttemptsPerCron();
     while ((de = dictNext(di)) != NULL) {
         clusterNode *node = dictGetVal(de);
         /* We free the inbound or outboud link to the node if the link has an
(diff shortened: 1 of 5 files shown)


tests/unit/cluster/cluster-reliable-meet.tcl
```diff
@@ -160,6 +160,10 @@ start_cluster 2 0 {tags {external:skip cluster} overrides {cluster-node-timeout
                 after 100
                 # Since we are in handshake, we use a randomly generated ID we have to find
                 R 1 DEBUG CLUSTERLINK KILL ALL [cluster_get_first_node_in_handshake 1]
+
+                # Since we retry to connect for a specific node once in one second
+                after 1000
```
madolson: Ideally always use a wait_for instead of a hardcoded time. You can use the cluster links command to learn about the state.

src/cluster_legacy.c
```diff
@@ -1632,6 +1632,7 @@ clusterNode *createClusterNode(char *nodename, int flags) {
     node->replicaof = NULL;
     node->last_in_ping_gossip = 0;
     node->ping_sent = node->pong_received = 0;
+    node->last_reconnect_attempted = 0;
```
```suggestion
    node->last_connect_attempted = 0;
```

src/cluster_legacy.c
```diff
@@ -5387,8 +5396,9 @@ void clusterCron(void) {
         /* The protocol is that function(s) below return non-zero if the node was
          * terminated.
          */
-        if (clusterNodeCronHandleReconnect(node, now)) continue;
+        if (clusterNodeCronHandleReconnect(node, now, &cluster_node_conn_attempts)) continue;
```
madolson: I still think it would be a bit safer if we initialized at a random position inside the cluster->nodes (a random bucket is fine) and looped until we returned back to the initial node.

sarthakaggarwal97: yeah, that would bring in more randomness. The iterator goes in one direction so I can maybe iterate `N + K` times where `K` is the random position where we will circle back.

Move iterator through `K nodes -> start_position`, `start_position-> end_position` and finally `end_position -> start_position`

src/cluster_legacy.c
```diff
@@ -5378,18 +5399,33 @@ void clusterCron(void) {
...
+    int total = dictSize(server.cluster->nodes);
+    int skip = rand() % (total == 0 ? 1 : total);
+
+    /* First iterator: from skip point to end */
     di = dictGetSafeIterator(server.cluster->nodes);
+    int i = 0;
+
+    /* Skip first 'skip' entries */
+    while (i++ < skip && (de = dictNext(di)) != NULL) {
+    }
```
hpatro: I would prefer us taking this functionality to the dictionary level to avoid these random pointer access. It would be nice if we can introduce an API like `dictGetRandomIterator(d)` and avoid maintaining any information on the caller side.

src/cluster_legacy.h
```diff
@@ -352,6 +352,7 @@ struct _clusterNode {
     mstime_t fail_time;                     /* Unix time when FAIL flag was set */
     mstime_t repl_offset_time;              /* Unix time we received offset for this node */
     mstime_t orphaned_time;                 /* Starting time of orphaned primary condition */
+    mstime_t last_connect_attempted;        /* Unix time we last tried to establish a link */
```
hpatro: Better to term it as 

```suggestion
    mstime_t outbound_link_attempt_time; /* Unix time we last tried to establish an outgoing link */
```

src/cluster_legacy.c
```diff
@@ -5314,6 +5317,11 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t now) {
     }
 
     if (node->link == NULL) {
+        if (now - node->last_connect_attempted <= 1000 || *cluster_conn_attempts > MAX_CONNECTION_ATTEMPTS_PER_CRON) {
```
hpatro: Add or reuse a macro for wait time instead of hardcoding it as `1000`.

hpatro: Also, maybe we can avoid this condition here and let this method just perform reconnection in a guaranteed manner. We can move this logic to the caller. WDYT?

sarthakaggarwal97: I believe all the nodes still need to pass through `clusterNodeCronHandleReconnect` since we do again calculate number of pfail nodes, take some actions if the handshake times out etc.

I think if we throttle just the part where we introduce new connections, which is the bottleneck for us, it would keep overall impact limited. Let me know what you think about this.

src/cluster_legacy.c
madolson: Yeah, I think having a random iterator as a generic construct is very useful.

sarthakaggarwal97: thanks for the suggestion. I've now added a new API in `dict`

src/dict.c
```diff
@@ -857,6 +857,57 @@ dictIterator *dictGetSafeIterator(dict *d) {
     return i;
 }
 
+/* Initialize a randomized iterator that skips a random number of entries
+ * and then iterates over the entire dictionary, wrapping around to cover
+ * all entries.
+ * Use dictRandomIterNext to fetch the next element in the iterator.
+ */
+dictRandomIterator *dictGetRandomIterator(dict *d) {
```
hpatro: I don't see this API handling rehashing ?

hpatro (commented):
Could you also add unit tests for the new dictionary API(s)? 

Consider rehashing, entry addition/removal during the iteration.

src/cluster_legacy.c
```diff
@@ -5314,6 +5318,11 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t now) {
     }
 
     if (node->link == NULL) {
+        if (now - node->outbound_link_attempt_time <= OUTBOUND_CONN_RETRY_INTERVAL || *cluster_conn_attempts > MAX_CONNECTION_ATTEMPTS_PER_CRON) {
```
murphyjacob4: Would it make sense to also check node->inbound_link? If we have an inbound link, the node should be up, right?

```suggestion
        if (!node->inbound_link && (now - node->outbound_link_attempt_time <= OUTBOUND_CONN_RETRY_INTERVAL || *cluster_conn_attempts > MAX_CONNECTION_ATTEMPTS_PER_CRON)) {
```

src/dict.c
sarthakaggarwal97: added some tests, would you please check!

sarthakaggarwal97: I have added some tests which checks rehashing. Please check and see if they are correct and cover all the cases.

src/cluster_legacy.c
sarthakaggarwal97: so if we have an inbound connection, we needn't require outbound to gossip. I think that makes sense.

sungming2:
It might be minor improvement, but is it necessary to maintain two iterations in the cron?

https://github.com/valkey-io/valkey/blob/22872611fa0c999f05392df037075faf9e2e5edd/src/cluster_legacy.c#L5396-L5407

https://github.com/valkey-io/valkey/blob/22872611fa0c999f05392df037075faf9e2e5edd/src/cluster_legacy.c#L5443-L5444

src/cluster_legacy.c
```diff
@@ -5378,18 +5388,19 @@ void clusterCron(void) {
     /* Clear so clusterNodeCronHandleReconnect can count the number of nodes in PFAIL. */
     server.cluster->stats_pfail_nodes = 0;
     /* Run through some of the operations we want to do on each cluster node. */
-    di = dictGetSafeIterator(server.cluster->nodes);
-    while ((de = dictNext(di)) != NULL) {
+    dictRandomIterator *random_iter = dictGetRandomIterator(server.cluster->nodes);
```
sungming2: I like this idea but if I think about the worst case, some `unlucky nodes` may never get reconnected for a long time (many cron rounds). Please correct me if it's wrong

e.g., 
```
Nodes:  [ N1 N2 N3 N4 ... N2000 ]
Status: [ O O O O ... O X X ]

O = reconnect attempt issued
X = not yet attempted
```

src/dict.c
```diff
@@ -857,6 +857,57 @@ dictIterator *dictGetSafeIterator(dict *d) {
...
+    dictRandomIterator *it = zmalloc(sizeof(dictRandomIterator));
+    it->d = d;
+    const long size = dictSize(d);
+    it->skip = rand() % (size == 0 ? 1 : size);
```
hpatro: Should we maintain the skipped entries count? I think it will lead to incorrect iteration logic in the second half. What if there are more entries inserted above? We would most likely not perform a full iteration and skip some elements.

I think the logic would be more hardened if we maintain the random bucket index on which we start the iteration and then finish it when we hit that back again on wrap-around.

src/dict.h
```diff
@@ -108,6 +108,19 @@ typedef struct {
...
 } dictDefragFunctions;
 
+/* Iterator for randomized full-coverage dictionary traversal.
+ *
+ * Visits every entry in the dictionary exactly once, starting from a
+ * random offset and wrapping around. Useful for fairness in periodic tasks.
+ */
+typedef struct dictRandomIterator {
+    dict *d;
+    dictIterator *iter;  /* For phase 0: from skip to end */
```
hpatro: The comment is incorrect here. We reuse the same iter in the second phase as well.

```
    dictIterator *iter;  /* iterator to access the dict entries */
```

src/dict.h
```diff
@@ -108,6 +108,19 @@ typedef struct {
...
+    unsigned long skip;  /* Entries to skip initially */
+    unsigned long index; /* Number of entries returned in phase 1 */
+    int is_wrapped;      /* 0 = first half, 1 = wrap-around */
```
hpatro: What do you think about `iter_phase` ?

sarthakaggarwal97: yeah I initially had `iter_phase` but changed to wrapped. I am happy with `iter_phase`

src/dict.c
sarthakaggarwal97: This was a good case to point. I have added a test and made some changes to address this case. Please review.

src/cluster_legacy.c
sarthakaggarwal97: I think random first entry point is good way to guarantee coverage. We run this 10 times a second, which would give us good probability. I tried killing 475 primaries in the node, and it was able to recover without affecting convergence time.

src/dict.h
```diff
@@ -108,6 +108,22 @@ typedef struct {
...
+/* Iterator for randomized full-coverage dictionary traversal.
+ *
+ * Visits every entry in the dictionary exactly once, starting from a
+ * random offset and wrapping around. Useful for fairness in periodic tasks.
+ */
+typedef struct dictRandomIterator {
+    dict *d;                /* Underlying dictionary */
+    dictIterator *iter;     /* Iterator to access the dict entries */
+    unsigned long skip;     /* Entries to skip initially */
+    unsigned long index;    /* Number of entries returned in phase 1 */
```
sungming2: nit: Looks like this index field is never used

sarthakaggarwal97: I missed removing it in the latest commit! thanks

src/dict.c
```diff
@@ -857,6 +857,77 @@ dictIterator *dictGetSafeIterator(dict *d) {
...
+        dictReleaseIterator(it->iter);
+        it->iter = dictGetSafeIterator(it->d);
+    }
+
+    de = dictNext(it->iter);
+    /* Stop once we circle back to the first entry. */
+    if (de == NULL || de == it->first_entry) {
+        it->finished = 1;
+        return NULL;
+    }
```
sungming2: I wonder if it works when:
```
dict["A", "B", "C"]
randIter = getRandIter(dict) // Skip "A", iter will return "B" as firstentry
while(randIterNext(randIter) != NULL) {
  ... some logic..
  dictDelete("B")
   ... some logic..
}  
```

if it doens't work, is it possible to avoid pointer comparison?

hpatro:
Just thought about this @sarthakaggarwal97, let me know if you've already explored this and tested.

Could we make the diff smaller by removing the random iterator part and just introduce the delay in attempting reconnection. I think that should solve the issue. As during steady state iterating over all the entries doesn't lead to any regression and there is no syscall involved.

src/cluster_legacy.c
```diff
@@ -5297,10 +5298,13 @@ static int nodeExceedsHandshakeTimeout(clusterNode *node, mstime_t now) {
     return now - node->ctime > getHandshakeTimeout() ? 1 : 0;
 }
 
+#define MAX_CONNECTION_ATTEMPTS_PER_CRON 10
+#define OUTBOUND_CONN_RETRY_INTERVAL 1000
```
hpatro: Also, let's determine this value based on `cluster-node-timeout` and not a hardcoded value.

sarthakaggarwal97:
@hpatro and I had a discussion offline on this. We were both aligned that with the random dict iterator API, it was getting too complicated to limit the number of connections while ensuring that all the elements are traversed only once and none of the elements are traversed multiple times since deletions were possible. (specially with the introduction of dict API, where we need to maintain some contracts).

I have updated the PR with a much leaner change, where based on the number of nodes in the cluster and the cluster node timeout, we determine the max number of retryable nodes, such that, each of the nodes is atleast retried 10 times in a given node timeout window.

src/cluster_legacy.c
```diff
@@ -5419,6 +5427,28 @@ static void clusterNodeCronFreeLinkOnBufferLimitReached(clusterNode *node) {
...
+ */
+static long long maxConnectionAttemptsPerCron(const int nodes, const long long timeout_ms) {
+    if (nodes <= 0 || timeout_ms <= 0)
+        return 0;
+    /*
+     * We run the cron loop every 100 ms.  To reach 100 % of the nodes
+     * within the timeout, we need: ceil(nodes * 100 / timeout_ms)
+     * We use (a + b − 1) / b to give us the integer-ceil without using floating-point.
+     */
+    const long long min_nodes_for_coverage = (nodes * 100 + timeout_ms - 1) / timeout_ms;
```
hpatro: Let's add a macro here. I'm fine with ceil as well, do you see any risk here?

`#define CLUSTER_CRON_PERIOD_MS 100`

src/cluster_legacy.c
```diff
@@ -5439,6 +5470,7 @@ void clusterCron(void) {
     server.cluster->stats_pfail_nodes = 0;
     /* Run through some of the operations we want to do on each cluster node. */
     di = dictGetSafeIterator(server.cluster->nodes);
+    const long long max_conn_attempts = maxConnectionAttemptsPerCron(dictSize(server.cluster->nodes), server.cluster_node_timeout);
```
hpatro: Should this be server.cluster_node_timeout/2, we try to disconnect and reconnect at half the period of node timeout.

```suggestion
    const long long max_conn_attempts = maxConnectionAttemptsPerCron(dictSize(server.cluster->nodes), server.cluster_node_timeout/2);
```

hpatro (commented):
Mostly LGTM. I like this approach, simple and does the job. 

Could you share the latest numbers here for others, I believe the top comment is outdated?

src/cluster_legacy.c
```diff
@@ -5419,6 +5427,28 @@ static void clusterNodeCronFreeLinkOnBufferLimitReached(clusterNode *node) {
     freeClusterLinkOnBufferLimitReached(node->inbound_link);
 }
 
+/**
+ * Compute the maximum number of connection attempts the cluster-cron
+ * loop should schedule in a single cron.
+ *
+ * We want to guarantee that every node is contacted twice within cluster node timeout.
+ */
+static long long maxConnectionAttemptsPerCron(const int nodes, const long long timeout_ms) {
```
sungming2: Do we need to pass nodes and timeout_ms as arguments to this function? Will this function be reused elsewhere with different values?

sarthakaggarwal97:
Sharing the benchmark numbers with the latest change when I kill `450 out of 1000` primaries. The P99 is still improved by a 75% (best case) while avg and P90 still see 10% improvement.

Screenshot 2025-07-16 at 9 45 27 PM
Screenshot 2025-07-16 at 9 46 06 PM
Screenshot 2025-07-16 at 9 45 47 PM

src/cluster_legacy.c
```diff
@@ -5419,6 +5428,28 @@ static void clusterNodeCronFreeLinkOnBufferLimitReached(clusterNode *node) {
...
+ */
+static long long maxConnectionAttemptsPerCron(void) {
+    long long reconnect_interval = server.cluster_node_timeout / 2;
+    if (dictSize(server.cluster->nodes) <= 0 || reconnect_interval <= 0)
+        return 0;
+    /*
+     * We run the cron loop every 100 ms.  To reach 100 % of the nodes
+     * within the timeout, we need: ceil(nodes * 100 / reconnect_interval)
+     */
+    const long long min_nodes_for_coverage = (dictSize(server.cluster->nodes) * CLUSTER_CRON_PERIOD_MS + reconnect_interval - 1) / reconnect_interval;
```
hpatro: code and comment seem to suggest different. IIRC you mentioned you'll move the code to use ceil.

src/cluster_legacy.c
```diff
@@ -5374,6 +5377,12 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t now) {
     }
 
     if (node->link == NULL) {
+        mstime_t reconnect_interval = server.cluster_node_timeout / 2;
+        if (!node->inbound_link && (now - node->outbound_link_attempt_time < reconnect_interval / NODE_CONNECTION_RETRIES_PER_TIMEOUT && *cluster_conn_attempts > max_conn_attempts)) {
+            return 1;
+        }
```
hpatro: Please add a comment explaining this.

src/cluster_legacy.c
```diff
@@ -5447,8 +5480,9 @@ void clusterCron(void) {
         /* The protocol is that function(s) below return non-zero if the node was
          * terminated.
          */
-        if (!server.debug_cluster_disable_reconnection && clusterNodeCronHandleReconnect(node, now)) continue;
+        if (!server.debug_cluster_disable_reconnection && clusterNodeCronHandleReconnect(node, now, &cluster_node_conn_attempts, max_conn_attempts)) continue;
     }
+    cluster_node_conn_attempts = 0;
```
hpatro: Why do we need to set it to 0 here ?

src/cluster_legacy.c
```diff
@@ -5419,6 +5428,28 @@ static void clusterNodeCronFreeLinkOnBufferLimitReached(clusterNode *node) {
     freeClusterLinkOnBufferLimitReached(node->inbound_link);
 }
 
+/**
+ * Compute the maximum number of connection attempts the cluster-cron
+ * loop should schedule in a single cron.
+ *
+ * We want to guarantee that every node is contacted twice within cluster node timeout.
```
hpatro: > We want to guarantee that every node is contacted twice within cluster node timeout.

This seems incorrect.

hpatro (approved):
Thanks for simplifying it @sarthakaggarwal97, looks good to me. The logic is dynamic based on timeout value, each node will attempt reconnection after every 750 ms for the default config value i.e. attempt 10 connection retries within `pfail` state to `fail` (if all agree).

Some more explanation:

For the default cluster node timeout period (15 seconds), we try disconnecting and reconnecting at half of that interval which is at 7.5 seconds. With the current unstable, after we treat the node as `pfail`, we disconnect the link and then retry connecting at each cron cycle i.e. 75 times (7.5 seconds / 100 ms). With this change, we introduce a delay between the connection attempts.

This change avoids too many connection attempts being made in case of large no. of node failures in a cluster which caused high CPU utilization and spikes.

hpatro:
@enjoy-binbin / @madolson If either of you could take a look at this? that would be great.

src/cluster_legacy.c
```diff
@@ -5447,7 +5485,7 @@ void clusterCron(void) {
         /* The protocol is that function(s) below return non-zero if the node was
          * terminated.
          */
-        if (!server.debug_cluster_disable_reconnection && clusterNodeCronHandleReconnect(node, now)) continue;
+        if (!server.debug_cluster_disable_reconnection && clusterNodeCronHandleReconnect(node, now, &cluster_node_conn_attempts, max_conn_attempts)) continue;
```
```suggestion
        if (!server.debug_cluster_disable_reconnection && clusterNodeCronHandleReconnect(node, now, &cluster_node_conn_attempts)) continue;
```
Since max_conn_attempts is a constant, I think it would be simpler if you just set the number of attempts and decrement it within the function and check if it's greater than zero.

src/cluster_legacy.c
```diff
@@ -5419,6 +5433,28 @@ static void clusterNodeCronFreeLinkOnBufferLimitReached(clusterNode *node) {
...
 
+/**
+ * Compute the maximum number of connection attempts the cluster-cron
+ * loop should schedule in a single cron.
+ *
+ * We want to guarantee that every node is contacted 10 times within node timeout.
+ */
+static long long maxConnectionAttemptsPerCron(void) {
+    long long reconnect_interval = server.cluster_node_timeout / 2;
+    if (dictSize(server.cluster->nodes) <= 0 || reconnect_interval <= 0)
```
madolson: I think server.cluster->nodes always has itself, so it should always be greater than 0.

src/cluster_legacy.c
```diff
@@ -5419,6 +5433,28 @@ static void clusterNodeCronFreeLinkOnBufferLimitReached(clusterNode *node) {
...
+        return 0;
+    /*
+     * We run the cron loop every 100 ms.  To reach 100 % of the nodes
+     * within the timeout, we need: ceil(nodes * 100 / reconnect_interval)
+     */
+    const long long min_nodes_for_coverage = ceil(dictSize(server.cluster->nodes) * CLUSTER_CRON_PERIOD_MS / reconnect_interval);
```
madolson: All of these are integers, so ceil will not do anything since it will always be an integer
```suggestion
    const long long min_nodes_for_coverage = dictSize(server.cluster->nodes) * CLUSTER_CRON_PERIOD_MS / reconnect_interval;
```

tests/unit/cluster/cluster-reliable-meet.tcl
```diff
@@ -160,6 +172,13 @@ start_cluster 2 0 {tags {external:skip cluster} overrides {cluster-node-timeout
                 after 100
                 # Since we are in handshake, we use a randomly generated ID we have to find
                 R 1 DEBUG CLUSTERLINK KILL ALL [cluster_get_first_node_in_handshake 1]
+
+                wait_for_condition 5 400 {
+                    [is_node_in_handshake 1] && [is_node_in_handshake 0]
+                } else {
+                    fail "Node 0 or Node 1 never entered in handshake state"
+                }
```
madolson: What is this new assertion covering? It's not quite obvious.

madolson (commented):
The new strategy seems pretty smart, I like it!

tests/unit/cluster/cluster-reliable-meet.tcl
sarthakaggarwal97: I wanted to ensure that there is no regression in the way we establish a new connection, and a cluster link is present between the nodes within a certain time frame when the handshake times out.

src/cluster_legacy.c
sarthakaggarwal97: I kept this change because before we add the node to the dictionary (in `clusterInit()`), we have already initialized the cron through `initServer();` so there could be instances where this can fail? Code pointer over [here](https://github.com/valkey-io/valkey/blob/unstable/src/server.c#L7252).

hpatro: But the event loop starts much later. I think it's safe to remove it.

src/cluster_legacy.c
```diff
@@ -5439,6 +5453,28 @@ static void clusterNodeCronFreeLinkOnBufferLimitReached(clusterNode *node) {
...
+ * We want to guarantee that every node is contacted 10 times within node timeout.
+ */
+static long long maxConnectionAttemptsPerCron(void) {
+    long long reconnect_interval = server.cluster_node_timeout / 2;
+    if (reconnect_interval <= 0)
+        return 0;
+    /*
+     * We run the cron loop every 100 ms.  To reach 100 % of the nodes
+     * within the timeout, we need: ceil(nodes * 100 / reconnect_interval)
+     */
```
```suggestion
    /* We run the cron loop every 100 ms.  To reach 100 % of the nodes
     * within the timeout, we need: ceil(nodes * 100 / reconnect_interval) */
```

src/cluster_legacy.c
```diff
@@ -5439,6 +5453,28 @@ static void clusterNodeCronFreeLinkOnBufferLimitReached(clusterNode *node) {
...
+    const long long min_nodes_for_coverage = dictSize(server.cluster->nodes) * CLUSTER_CRON_PERIOD_MS / reconnect_interval;
+    /*
+     * Increase the coverage budget so each node can be probed 10 times
+     * inside the timeout.
+     */
```
```suggestion
    /* Increase the coverage budget so each node can be probed 10 times
     * inside the timeout. */
```

src/cluster_legacy.c
```diff
@@ -5439,6 +5453,28 @@ static void clusterNodeCronFreeLinkOnBufferLimitReached(clusterNode *node) {
     freeClusterLinkOnBufferLimitReached(node->inbound_link);
 }
 
+/**
+ * Compute the maximum number of connection attempts the cluster-cron
+ * loop should schedule in a single cron.
+ *
+ * We want to guarantee that every node is contacted 10 times within node timeout.
+ */
```
enjoy-binbin: Let's stick with the function name so we can search it easily. (avoid creating new name)
```suggestion
/* Compute the maximum number of connection attempts the clusterCron
 * loop should schedule in a single cron.
 *
 * We want to guarantee that every node is contacted 10 times within node timeout. */
```

src/cluster_legacy.c
```diff
@@ -5394,6 +5397,17 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t now) {
     }
 
     if (node->link == NULL) {
+        mstime_t reconnect_interval = server.cluster_node_timeout / 2;
+        /* Skip this outbound connection attempt when all three are true:
```
```suggestion
        /* Skip this outbound connection attempt when all these conditions are true:
```

enjoy-binbin (approved):
LGTM.

tests/unit/cluster/cluster-reliable-meet.tcl
hpatro: I'm skeptical as well. This check could easily turn this test to flaky failure.

src/cluster_legacy.c
```diff
@@ -5394,6 +5397,17 @@ static int clusterNodeCronHandleReconnect(clusterNode *node, mstime_t now) {
     }
 
     if (node->link == NULL) {
+        mstime_t reconnect_interval = server.cluster_node_timeout / 2;
+        /* Skip this outbound connection attempt when all these conditions are true:
+         *  1. No inbound link from the peer exists.
+         *  2. The back‑off window since the last try is still active
+         *  3. The cluster has already exceeded its retry budget for this cron cycle
```
hpatro: nit: The node has already exceeded its retry budget for this cron cycle

hpatro (approved):

hpatro:
Don't see any outstanding comments left. Will merge this in, after the test runs.

PR #2156, opened 2025-05-29
Have threads gracefully exit instead of killing them when checking thread count

madolson:
It appears that jemalloc *does not* like it when you cancel threads, and will sometimes crash. It does seem to be happy when it's allowed to gracefully exit though. 

Reference https://github.com/valkey-io/valkey/actions/runs/15312340723/job/43079580795?pr=2149#step:4:1444.

I also updated the test to randomly set the value a bunch, which makes the test fail a little more consistently. Issue seems no longer present.


src/io_threads.c
```diff
@@ -218,16 +219,12 @@ static void *IOThreadMain(void *myid) {
...
     initSharedQueryBuf();
-    pthread_cleanup_push(freeSharedQueryBuf, NULL);
 
     thread_id = (int)id;
     size_t jobs_to_process = 0;
     IOJobQueue *jq = &io_jobs[id];
-    while (1) {
-        /* Cancellation point so that pthread_cancel() from main thread is honored. */
-        pthread_testcancel();
+    while (io_thread_running[id]) {
```
murphyjacob4: Do we care about the memory fence?

```suggestion
    while (atomic_load_explicit(&io_thread_running[id], memory_order_relaxed)) {
```

madolson: We don't strictly need to based on the usage, we only transition once from true to false, so it doesn't really matter if it takes longer to recognize the load.

madolson:
I looked into this a bit more, it looks like it's actually because we set:
`pthread_setcanceltype(PTHREAD_CANCEL_ASYNCHRONOUS, NULL);` which immediately kills the thread, which has some unexpected side effects, which according to stackoverflow can result in crashing.

ranshid:
> I looked into this a bit more, it looks like it's actually because we set: `pthread_setcanceltype(PTHREAD_CANCEL_ASYNCHRONOUS, NULL);` which immediately kills the thread, which has some unexpected side effects, which according to stackoverflow can result in crashing.

I agree this might help cleaner solution for this case. I guess since we are draining the io-thread first the config change can still hang anyway and we already placed the cancellation point

src/io_threads.c
```diff
@@ -7,6 +7,7 @@
 #include "io_threads.h"
 
 static __thread int thread_id = 0; /* Thread local var */
+static atomic_bool io_thread_running[IO_THREADS_MAX_NUM] = {false};
```
uriyage: Why is atomic required?

madolson: It doesn't need to be, the code will still eventually exit regardless. It was more for clarity about thinking about it.

madolson:
> I agree this might help cleaner solution for this case. I guess since we are draining the io-thread first the config change can still hang anyway and we already placed the cancellation point

I didn't understand this comment.

ranshid:
> > I agree this might help cleaner solution for this case. I guess since we are draining the io-thread first the config change can still hang anyway and we already placed the cancellation point
> 
> I didn't understand this comment.

Your suggestion to have the io threads do deferred cancellation should also solve it right?

madolson:
> Your suggestion to have the io threads do deferred cancellation should also solve it right?

Yeah, the cancelling the thread seems less intuitive to me. We had an obvious bug that we didn't notice from testing, and I worry there might be others. Let me do some deeper testing.

src/io_threads.c
madolson: No longer relevant

src/io_threads.c
```diff
@@ -218,7 +218,7 @@ static void *IOThreadMain(void *myid) {
     snprintf(thdname, sizeof(thdname), "io_thd_%ld", id);
     valkey_set_thread_title(thdname);
     serverSetCpuAffinity(server.server_cpulist);
-    makeThreadKillable();
+    pthread_setcancelstate(PTHREAD_CANCEL_ENABLE, NULL);
```
ranshid: nit: suggest also explicitly set pthread_setcanceltype(PTHREAD_CANCEL_DEFERRED, NULL). although it is the default (PTHREAD_CANCEL_ENABLE is also the default) makews more sense to explicitly set both defaults or non.

ranshid (approved):
LGTM. small minor polish comment

src/io_threads.c
madolson: I'm just going to delete it then, I assumed it was necessary only because the other code path called it.

```suggestion
```

Issue #1783, opened 2025-02-26
[NEW] Kill unkillable scripts


zuiderkwast:
**The problem/use-case that the feature addresses**

If a Lua script has done any writes, it's not killable using SCRIPT KILL, because it would violate the atomicity guarantee of Lua script. Only shutting down the server can kill the script.

The atomicity of a script isn't always that important though. In some workloads, it would be totally fine to violate it. At least, it's not as bad as having to restart the server.

Often when you have a hanging script, it's because of some infinite loop or other bug in the script. If you have a buggy script, it may not write correct stuff to the database anyway.

**Description of the feature**

Allow SCRIPT KILL (and FUNCTION KILL) to kill a script even if it has done some writes.

We could add some extra FORCE argument, or a config, if we're worried about changing the default behavior.

**Alternatives you've considered**

Some way to undo the writes done by a script, such as a redo-list, a frozen copy of the key space similar, a copy-on-write implemented in user space.

**Additional information**

-

rjd15372:
I would vote for adding the `FORCE` extra argument. That way the user is acknowledging that understands the consequences, which will be described in the documentation of the command.

enjoy-binbin:
Great, a FORCE LGTM, i think we can do it. btw, lets also talk about https://github.com/valkey-io/valkey/pull/866#issuecomment-2290542519?

Currently to ensure script atomicity, we can bypass maxmemory (data memory or lua heap memory). If we allow kill unkillable scripts, i think we can fix it. (we allow killing them in internal fork, ie. return OOM early)
```
EVAL "for i=1,10000000 do redis.call('SET', 'key-kkkk:' .. i, 'value-tttt:' .. i) end" 0
EVAL "for i=1,10000000 do redis.pcall('SET', 'key-kkkk:' .. i, 'value-tttt:' .. i) end" 0
EVAL "local a={}; local i; for i=1, 30000000 do a[i]=i; end" 0
```

zuiderkwast:
To return OOM early is an automatic kill, not SCRIPT KILL, right?

If we let the script be killed automatically by the memory limit, it's a behavior change that maybe should have a config, or at least be done in a major release.

enjoy-binbin:
> To return OOM early is an automatic kill, not SCRIPT KILL, right?

yes, it is an automatic kill. There are always people messing around with scripts... We will give an extra 30MB of memory in case it is really lua tmp memory.

[...]

PR #2157, opened 2025-05-30
Kill the busy script during failover to avoid data inconsistency

enjoy-binbin:
When a primary node execute a busy write script and a failover
occurs druing the execution, we need to kill the busy script and
perform a full resync. Otherwise, after the failover is over, the
new replica will retain the data inserted during the script execution,
but there will be no such changes on the new primary node.

Killing the scrpt may break the atomicity of the script, but please
note in this case, the script will also get an error since the old
primary lost its primaryship, and this will make us retain some
data and break the atomicity badly.
```
Script attempted to access a non local key in a cluster node
```

This commit ensures the final consistency of the data between the
replica and the primary after the failover. We will not keep some
of the data, we will just like rollback to the beginning.

There is an quick simple example, current behavior on the unstable branch:
```
start_cluster 3 1 {tags {external:skip cluster} overrides {cluster-replica-no-failover yes}} {
    test "Test busy script during the failover" {
        r config set busy-reply-threshold 1000

        set rd1 [valkey_deferring_client 0]
        $rd1 eval "for i=1,1000000 do redis.pcall('SET', 'key-kkkk:' .. i, 'value-tttt:' .. i) end" 0

        R 3 cluster failover

        $rd1 read

        wait_for_condition 1000 50 {
            [s 0 role] == "slave" && [s -3 role] == "master"
        } else {
            fail "No failover detected"
        }

        puts [R 0 dbsize]  ;# output number > 0
        puts [R 3 dbsize]  ;# output 0
    }
}
```


src/script.c
```diff
@@ -295,6 +299,11 @@ int scriptIsEval(void) {
     return curr_run_ctx->flags & SCRIPT_EVAL_MODE;
 }
 
+void scriptMarkedKillByFailover(void) {
```
ranshid:
```suggestion
void scriptMarkKilledByFailover(void) {
```

src/cluster_legacy.c
```diff
@@ -5946,6 +5946,26 @@ static void clusterSetPrimary(clusterNode *n, int closeSlots, int full_sync_requ
...
         myself->flags |= CLUSTER_NODE_REPLICA;
+
+        if (scriptIsTimedout()) {
+            /* If a failover occurs during a busy script execution, we need to kill
+             * the script to avoid inconsistent data after becoming a new replica.
+             * Otherwise, after the final script execution ends, whether it ends
+             * normally or ends with a `CLUSTER_REDIR*` error, the new replica will
+             * retain the keys written during the execution, resulting in inconsistent
+             * data with the primary node and cannot be recovered. */
+            scriptMarkedKillByFailover();
```
ranshid:
```suggestion
            scriptMarkKilledByFailover();
```

src/script.h
```diff
@@ -107,11 +108,13 @@ int scriptSetResp(scriptRunCtx *r_ctx, int resp);
 int scriptSetRepl(scriptRunCtx *r_ctx, int repl);
 void scriptCall(scriptRunCtx *r_ctx, sds *err);
 int scriptInterrupt(scriptRunCtx *r_ctx);
+void scriptMarkedKillByFailover(void);
```
ranshid:
```suggestion
void scriptMarkKilledByFailover(void);
```

ranshid (approved):
LGTM.

Isn't it something we would like to backport?

rjd15372 (approved):
LGTM

src/cluster_legacy.c
```diff
@@ -5946,6 +5946,26 @@ static void clusterSetPrimary(clusterNode *n, int closeSlots, int full_sync_requ
     if (clusterNodeIsPrimary(myself)) {
         myself->flags &= ~(CLUSTER_NODE_PRIMARY | CLUSTER_NODE_MIGRATE_TO);
         myself->flags |= CLUSTER_NODE_REPLICA;
+
+        if (scriptIsTimedout()) {
+            /* If a failover occurs during a busy script execution, we need to kill
+             * the script to avoid inconsistent data after becoming a new replica.
```
madolson, line 6628: Killing the script after a write can also result in inconsistent data, since a script might be doing multiple modifications to a key and only some of them are applied. We can then serve that data from the replica while it's waiting on the full sync. I think we should fully fix it instead of having it fail in a new way.

The only idea that comes to mind is we shouldn't allow a primary to step down to a replica in the middle of executing a script. That way at least it finishes executing and/or the end user makes the conscious decision to kill it.

enjoy-binbin: so you prefer we just return when doing `clusterProcessPacket`? so that the primary won't process the packet and won't step down to a replica. In this case, the primary will be mark with FAIL since it won't do the PONG thing. That is ok to me, currently when the primary into a busy script, its replica wont do the auto failover thing since prmary can still process the cluster packet.

Also in this case, should we add a new force to SCRIPT KILL so that user can kill the busy write script?

PingXie: >Killing the script after a write can also result in inconsistent data, since a script might be doing multiple modifications to a key and only some of them are applied. 

I thought this is what this pr is intended to solve? trying to understand your statement better:

1. I assume this is not about killing a write script via `SCRIPT KILL`, which can't kill scripts that dirty the keyspace today;
2. so this should be the case that @enjoy-binbin is fixing. assuming so, my understanding is that this pr would, upon transitioning a primary stuck in such a write script to a replica: 
a) stop the long running write script; 
b) force a full re-sync from the newly elected primary
no read is served while the full sync is underway hence no intermediate writes would be exposed, imo.

> The only idea that comes to mind is we shouldn't allow a primary to step down to a replica in the middle of executing a script. That way at least it finishes executing and/or the end user makes the conscious decision to kill it.

this is trading off availability for consistency, which to me feels like a local decision that is not aligned with the global "AP" properties when it comes to user data.

yet another more generic option imo could be holding the non-committed writes in a shadow dictionary. we will need to merge the query results and apply the shadow dictionary to the main dictionary on successful scrip exits. this would allow a read-write script to be killed by `SCRIPT KILL` safely, either by the user or the engine itself (via honoring the script timeout policy). I have implemented this exact solution for a different database engine though it was disk based. I can give it a shot too but I would like to hear your thoughts first.

zuiderkwast: \+ @rjd15372

rjd15372: @PingXie 
> yet another more generic option imo could be holding the non-committed writes in a shadow dictionary. we will need to merge the query results and apply the shadow dictionary to the main dictionary on successful scrip exits

Considering that killing a script is a rare event, I would favor the opposite approach of storing the old state of the dictionary values that are written by the script, and apply the script writes directly in the dictionary. When the script finishes, we just discard the old state, which should be a much faster operation than merging the state of the shadow dictionary.

zuiderkwast: > Killing the script after a write can also result in inconsistent data, since a script might be doing multiple modifications to a key and only some of them are applied. We can then serve that data from the replica while it's waiting on the full sync. I think we should fully fix it instead of having it fail in a new way.

If there is a failover during the script execution, I think it makes sense to kill the script. The replica can still serve the stale data written by the script until the full sync completes. If the writes have already been replicated, they will come back after the full sync. The consistency between replica and primary is more important than the atomicity of the script in this case IMO.

> The only idea that comes to mind is we shouldn't allow a primary to step down to a replica in the middle of executing a script. That way at least it finishes executing and/or the end user makes the conscious decision to kill it.

Forbidding a failover sounds risky for the availability of the cluster.

zuiderkwast: Can't we achieve atomic script replication by replicating MULTI on the script's first write and EXEC on script completion?

It works as long as the script writes to a single slot. If it's a cross-slot script, I think it's less important to provide the atomicity guarantees.

madolson: > Can't we achieve atomic script replication by replicating MULTI on the script's first write and EXEC on script completion?

The script is replicated inside a multi-exec, so it's always correctly executed on the replica that is receiving the results of the script.  The problem this PR is trying to solve is on the primary that get's demoted during script execution, the script will basically be half executed as a primary and half executed as a replica, and behave incorrectly.

> this is trading off availability for consistency, which to me feels like a local decision that is not aligned with the global "AP" properties when it comes to user data.

This isn't about consistency between primary and the replica, it's about incorrectly executing the script on the primary. This part of the problem is unrelated to CAP. There is another discussion about whether or not we should immediately kill the node to restore availability, which is a C vs A tradeoff.

> no read is served while the full sync is underway hence no intermediate writes would be exposed, imo.

This is not how the implementation works today. There is a period of time where the old replica can still serve traffic, especially if you are using the "swapDB" approach to full sync. I suppose one other way to fully solve the corrupted state is we could just mark the full keyspace as dirty and stop serving any traffic. 

> so you prefer we just return when doing clusterProcessPacket?

I would prefer that approach yes. The only primary node should be correctly marked as `FAIL` and would get marked back as available when it exits the script (or someone kills it), after it recognizes it's a replica. 

## Shadow dictionary discussion

That would likely incur some amount of performance tradeoff, but will also be a pathway to [transactions with rollbacks](https://github.com/valkey-io/valkey/issues/1783). I think we should independently pursue being able to rollback a script, but we probably also need a tactical decision here.

murphyjacob4: > I suppose one other way to fully solve the corrupted state is we could just mark the full keyspace as dirty and stop serving any traffic.

Sounds like we just async flush the DB if `scriptIsWriteDirty` is true? We'll need a full sync anyways, the main difference is just we won't serve the "script half completed" state after demotion and before full sync to READONLY clients, which seems reasonable in this edge case

PingXie: > > this is trading off availability for consistency, which to me feels like a local decision that is not aligned with the global "AP" properties when it comes to user data.
> 
> This isn't about consistency between primary and the replica, it's about incorrectly executing the script on the primary. This part of the problem is unrelated to CAP. There is another discussion about whether or not we should immediately kill the node to restore availability, which is a C vs A tradeoff.
> 
> > no read is served while the full sync is underway hence no intermediate writes would be exposed, imo.
> 
> This is not how the implementation works today. There is a period of time where the old replica can still serve traffic, especially if you are using the "swapDB" approach to full sync. I suppose one other way to fully solve the corrupted state is we could just mark the full keyspace as dirty and stop serving any traffic.
> 
> > so you prefer we just return when doing clusterProcessPacket?
> 
> I would prefer that approach yes. The only primary node should be correctly marked as `FAIL` and would get marked back as available when it exits the script (or someone kills it), after it recognizes it's a replica.

thanks for the clarification. yeah shard-level availability is restored after the automatic failover. replica availability has to wait for the dirty script to either exit or get killed. between the two I would still prefer failing the script and stopping serving traffic (good catch on `swapdb`). letting the dirty script to finish on the old primary does keep the local data consistent but, shard wise, it is still inconsistent and when it full syncs with the new primary, the changes get rolled back. what if the user enables AOF `fsync always` on every node? I think it is important that the writer sees a failure.

>Considering that killing a script is a rare event, I would favor the opposite approach of storing the old state of the dictionary values that are written by the script

I think that is a good optimization/simplification.

> That would likely incur some amount of performance tradeoff, but will also be a pathway to #1783.

thanks for the pointer - I am yet again late for another party of @zuiderkwast's ... 😉

madolson: So, I'm seeing the following options, each with an emoji if you want to vote for your favorite:
🚀 : When the primary detects it's a replica, drop all it's data and then immediately start replicating from the new primary. (NOTE: Under some circumstances, this will never happen today, if you have swapdb on replication you will never fully lose a replica unless the hardware dies).
❤️ : Let the primary finish executing it's script or it's killed by some other process. The primary will only act on the fact it's been demoted *after* the script finishes. 
🎉 : (Current implementation) Immediately kill the script once it's been demoted. This will break atomicity of the lua script, but it will probably work better than letting it start sending lots of error.
👀 : Wait until we have transactional scripts, then we can simply roll the script back and not have to worry about it.

enjoy-binbin: > 🚀 : When the primary detects it's a replica, drop all it's data and then immediately start replicating from the new primary. (NOTE: Under some circumstances, this will never happen today, if you have swapdb on replication you will never fully lose a replica unless the hardware dies).

I dont get this. The primary is doing the busy script, how do we continue to drop data and start copying?

> ❤️ : Let the primary finish executing it's script or it's killed by some other process. The primary will only act on the fact it's been demoted after the script finishes.

I think i can accept it. Though it still has issue, like now we cannot kill a write script right now. Or even if it is a slow script, after the failover we will have the data inconsistency issue like i mentioned in the top comment (we need to issue a full sync after this case).

> 🎉 : (Current implementation) Immediately kill the script once it's been demoted. This will break atomicity of the lua script, but it will probably work better than letting it start sending lots of error.

This one compared with the previous one, it is to determine whether to automatically kill the script. One point is in these cases, after the failover, the old primary is indeed outdated (the inserted data is not available on the new primary). Check whether you need to wait for the old primary to finish exection before the full sync, or kill it in advance and do the full sync. Normally we can catch this error with server.call (because the slot ownership is transferred), so the script can terminate after the failover. But we can not terminate a pcall or a dead loop script, which is what we (or users?) want to terminate after a failover.

madolson: > Or even if it is a slow script, after the failover we will have the data inconsistency issue like i mentioned in the top comment (we need to issue a full sync after this case

Yes, it's the same consistency issue we can have today though, which is tail end data loss. Which is the replica can be promoted at any time and lose some of the writes from the primary.

madolson: > I dont get this. The primary is doing the busy script, how do we continue to drop data and start copying?

We would abort the script and flush the data, which would make the data consistent.

enjoy-binbin:
Do we want to include this in 9.0?

madolson:
> Do we want to include this in 9.0?

We still have time, should we pin this for a live discussion? I think it makes sense to fix it.

PR #2160, opened 2025-05-30
Optimize scan/sscan/hscan/zscan commands by replacing list with vector

chzhoo:
### Description
1. The scan/sscan/hscan/zscan commands store iterated keys and values in a `list`, where each list element triggers a separate memory allocation and deallocation. This causes performance degradation when the list contains a large number of elements.
2. We reduce memory allocation/deallocation by replacing `list` with `vector`, a new dynamic array implementation.
3. For specific performance gains, refer to the benchmark results below.

### Benchmark
Run the benchmark script five times, and take the peak value as the final result.

#### Scan command
**benchmark script**
```
valkey-benchmark -r 1000000000 -n 10000000 -P 32 set key-__rand_int__ value-__rand_int__
for count in `echo 4 16 64 512`;do
	for((x=0;x<5;x+=1));do
		valkey-benchmark --threads 2 -r 100000000 -n 2000000  scan __rand_int__ count $count;
	done
done
```
**benchmark result**
count | QPS after optimization | QPS before optimization | Performance Boost
-- | -- | -- | --
4 | 114148.73| 108090.58 | 5.6%
16 | 98745.93| 87892.77 | 12.3%
64 | 64432.99| 49971.27 | 28.9%
512 | 15725.87 | 10174.75| 54.5%

#### Hscan Command
**benchmark script**
```
valkey-benchmark -r 1024 -n 1000000 -P 32 hset hash field-__rand_int__ value-__rand_int__
for count in `echo 4 16 64 512`;do
	for((x=0;x<5;x+=1));do
		valkey-benchmark --threads 2 -r 100000000 -n 2000000 hscan hash __rand_int__ count $count;
	done
done
```
**benchmark result**
count | QPS after optimization | QPS before optimization | Performance Boost
-- | -- | -- | --
4 | 115921.87| 114266.12 | 1.4%
16 | 103879.91| 97546.70 | 6.5%
64 | 71405.62 | 62976.26 | 13.4%
512 | 23168.53 | 16064.64| 44.2%

#### Sscan command
**benchmark script**
```
valkey-benchmark -r 1024 -n 1000000 -P 32 sadd set element-__rand_int__
for count in `echo 4 16 64 512`;do
	for((x=0;x<5;x+=1));do
		valkey-benchmark --threads 2 -r 100000000 -n 2000000 sscan set __rand_int__ count $count;
	done
done
```
**benchmark result**
count | QPS after optimization | QPS before optimization | Performance Boost
-- | -- | -- | --
4 | 123054.20| 119388.73 | 3.1%
16 | 115928.59| 110975.48 | 4.5%
64 | 94099.94| 86945.18 | 8.2%
512 | 40320.95 | 32501.83| 24.1%

#### Zscan command
**benchmark script**
```
valkey-benchmark -r 1024 -n 1000000 -P 32 zadd zset __rand_int__ element-__rand_int__
for count in `echo 4 16 64 512`;do
	for((x=0;x<5;x+=1));do
		valkey-benchmark --threads 2 -r 100000000 -n 2000000 zscan zset __rand_int__ count $count;
	done
done
```
**benchmark result**
count | QPS after optimization | QPS before optimization | Performance Boost
-- | -- | -- | --
4 | 97437.40| 96362.33 | 1.1%
16 | 71405.62|70155.75 | 1.8%
64 | 35066.80| 33002.75 | 6.3%
512 | 7498.02 | 6654.89| 12.7%

CPU: AMD EPYC 9754 128-Core Processor * 8
OS: 	Ubuntu Server 22.04 LTS 64bit
Memory: 16GB
VM: Tencent cloud SA5 | SA5.2XLARGE16
Server startup command: valkey-server --save '' --appendonly no


src/db.c
```diff
@@ -912,15 +912,71 @@ void keysCommand(client *c) {
...
+#define MAX_SCAN_ARRAY_BUFFER 24
+
+/* Used to store the iteration results from the SCAN/HSCAN/SSCAN/ZSCAN commands. */
+typedef struct {
+    sds *array;  /* Points to buf or heap */
+    size_t used; /* Used size of the array */
+    size_t size; /* Allocated size of the array */
+    void (*free)(void *);
+    sds buf[MAX_SCAN_ARRAY_BUFFER]; /* Pre-allocated buffer reduces heap allocation */
+} scanArray;
```
zuiderkwast: This looks like a generic dynamic array to me.

To make it useful for other things in the future, like as replacing `list` in more places, can we make this a generic dynamic array instead? We can put it in a new file `array.c` and `array.h`. I imagine we could replace things like blocked keys that are currently lists with this kind of dynamic array instead.

To make it more slim, we could skip the pre-allocated buffer. It makes the struct fit within 32 bytes. Jemalloc is quite very fast and it will often return memory that is already in L1-L2-L3 cache.

src/db.c
```diff
@@ -912,15 +912,71 @@ void keysCommand(client *c) {
...
 /* Data used by the dict scan callback. */
 typedef struct {
-    list *keys;     /* elements that collect from dict */
-    robj *o;        /* o must be a hash/set/zset object, NULL means current db */
-    serverDb *db;   /* database currently being scanned */
-    long long type; /* the particular type when scan the db */
-    sds pattern;    /* pattern string, NULL means no pattern */
-    long sampled;   /* cumulative number of keys sampled */
-    int only_keys;  /* set to 1 means to return keys only */
+    scanArray *result; /* elements that collect from dict */
```
zuiderkwast: Should we put the struct inline in the scanData struct to save another allocation?

```suggestion
    scanArray result; /* elements that collect from dict */
```

zuiderkwast (commented):
This can be a good, but it's more code. The list implementation is generic but this array is not generic. Let's make it generic?

chzhoo:
> This can be a good, but it's more code. The list implementation is generic but this array is not generic. Let's make it generic?

Generic array is a more elegant implementation. I'll try to do this later.

src/db.c
chzhoo: `scanArray *result` is a pointer to a stack object and will not cause memory allocation, so it is better to use `scanArray *result`

src/db.c
chzhoo: 1. Agreed. In some cases, arrays are more suitable than lists, such as `watched_keys` and `clients_pending_write`.
2. Dropping the pre-alloc buffer may slightly impact performance – I'll track this in later tests.

src/db.c
zuiderkwast: True, but scanData is also stack-allocated. We could save 8 bytes to avoid storing a pointer but it doesn't matter. Simplicity and clarity is more important.

chzhoo:
What do you think of the newly submitted code? Detailed unit tests and benchmark tests will be provided later. @zuiderkwast

src/array.h
```diff
@@ -0,0 +1,86 @@
...
+ *                             `-> Pointer returned to the user.
+ */
+
+/* Metadata struct of array */
+typedef struct array {
+    unsigned long len;       /* Current length of the array */
+    unsigned long alloc;     /* Allocated size of the array */
+    void (*free)(void *ptr); /* Free method for each element */
+    char data[];             /* Pointer returned to the user */
+} array;
```
zuiderkwast: TBH, I don't think we should model it after sds. It is not type safe and not very intuitive.

I think it's better that we make the array struct fixed size, so it can be included in other structs, and the data a separate allocation. In this way, we still only need one allocation for the data when the array struct is on the stack or embedded in another struct.

```C
typedef struct {
    void *items;
    size_t len;
    size_t alloc;
    void (*free)(void *ptr);
} array;
```

The benefit is that when we need to reallocate the items array, the array is still the same, so the API will be simpler.

We can allow to init an array allocated by the caller, on the stack or inside another struct:

```C
void arrayInit(array *a, size_t initial_capacity, void (*free)(void *)) {
    a->items = zmalloc(initial_capacity * sizeof(void *));
    a->len = 0;
    a->alloc = initial_capacity;
    a->free = free;
}
```

For functions that modify the array, they don't need to return a new array. They can just return void.

```C
void arrayPush(array *a, void *item);
void arrayResize(array *a, size_t new_size);
...
```

src/array.h
```diff
@@ -0,0 +1,86 @@
...
+ *      we can access individual elements via dynamic_array[index].
+ */
+#define arrayNew(v, s, f)                                                          \
+    do {                                                                           \
+        array *__array_meta_ptr__ = zmalloc(ARRAY_META_SIZE + sizeof(*(v)) * (s)); \
+        __array_meta_ptr__->len = 0;                                               \
+        __array_meta_ptr__->alloc = (s);                                           \
+        __array_meta_ptr__->free = (f);                                            \
+        v = (void *)(__array_meta_ptr__ + 1);                                      \
+    } while (0)
```
zuiderkwast: Macro code like this is difficult to read and not type safe.

If we want to write this in the header file, it can be done as `static inline` functions. They are just like macros but easier to read. But there is not really a reason. We can put the implementation in array.c. We compile with link-time-optimization (-flto) by default so they can still be inlined in other C files.

src/array.h
```diff
@@ -0,0 +1,86 @@
...
+        v[ARRAY_HDR(v)->len - 1] = e;                    \
+    } while (0)
+
+#define arrayForeach(v, f)                                            \
+    do {                                                              \
+        if (!f) break;                                                \
+        for (unsigned long __x__ = 0; __x__ < arrayLen(v); __x__++) { \
+            (f)(v[__x__]);                                            \
+        }                                                             \
+    } while (0)
```
zuiderkwast: If we don't make the array opaque, we can just let the caller access the elements directly. It is less abstractions, but I don't think it's bad.

```C
for (i = 0; i < arr->len; i++) {
    doStuffWithItem(arr->items[i]);
}
```

What we need is functions that automatically resize it though.

zuiderkwast (commented):
It's interesting to use the SDS style to use only one allocation for header and elements. I have thought about it too in the past, but to be honest I don't think it's the best idea. It does not give a very nice and simple API.

I remember @JimB123 has an implementation of a FIFO queue that is more efficient than a linked list.

We can't always use a fifo queue to replace a linked list though. But I guess a dynamic array can't always replace a linked list either. Sometimes we delete a list item in the middle of the list. That's not trivial with an array, unless we can ignore the order and move the last element to fill the hole.

Still, I believe a simple dynamic array can replace the linked list in many situations, such as clients blocked on a key, clients with pending reads, and so on. Clients with pending reads is probably the most important list to look at because it's in the hot code path for every command.

@JimB123 it would be interesting to hear your opinion on this design.

chzhoo:
> It's interesting to use the SDS style to use only one allocation for header and elements. I have thought about it too in the past, but to be honest I don't think it's the best idea. It does not give a very nice and simple API.
> 
> I remember @JimB123 has an implementation of a FIFO queue that is more efficient than a linked list.
> 
> We can't always use a fifo queue to replace a linked list though. But I guess a dynamic array can't always replace a linked list either. Sometimes we delete a list item in the middle of the list. That's not trivial with an array, unless we can ignore the order and move the last element to fill the hole.
> 
> Still, I believe a simple dynamic array can replace the linked list in many situations, such as clients blocked on a key, clients with pending reads, and so on. Clients with pending reads is probably the most important list to look at because it's in the hot code path for every command.
> 
> @JimB123 it would be interesting to hear your opinion on this design.

Based on your feedback, there are two concerns with this implementation of the dynamic array:

1. Adjacent storage of metadata and data which complicates the API design.
2. Excessive use of macros.

The initial design was motivated by two features:

1. Adjacent storage of metadata and data to minimize memory jumps.
2. Support for variable-sized items rather than fixed-size items.

It appears these features are unnecessary for most scenarios.  I'll rewrite the implementation later based on your suggestions.

chzhoo:
After more consideration, I believe that the array still needs to support variable-sized elements. 

If the array only supports fixed-size elements (e.g. sizeof(void*)), storing structures larger than a pointer would necessitate additional memory allocations + memory jump. For example, if we refactor the ready_keys in struct valkeyServer using an array, this exact problem would occur.

```
typedef struct {
    void* items;
    size_t size;
    size_t alloc;
    size_t item_size;
    void (*free)(void *ptr);
} array;

void arrayNew(array* a, size_t capicity, size_t item_size, void (*free)(void *));
void arrayResize(array* a, size_t new_capicity);
void* arrayPush(array* a);
void arrayCleanup(array* a);
```
If the dynamic array design might be more suitable like this  ?  @zuiderkwast

zuiderkwast:
> After more consideration, I believe that the array still needs to support variable-sized elements.
> 
> If the array only supports fixed-size elements (e.g. sizeof(void*)), storing structures larger than a pointer would necessitate additional memory allocations + memory jump. For example, if we refactor the ready_keys in struct valkeyServer using an array, this exact problem would occur.

OK. I suppose you're right, with variable-size elements we can optimize some code better.

With this implementation, we can't allow users to just access element i as `arr->items[i]`. We need a function `arrayGet(i)` to access the elements. We should also rename `items` to `data` or something to indicate it's not a simple array of pointers.

> `void arrayNew(array* a, size_t capicity, size_t item_size, void (*free)(void *));`

Use `arrayInit` for this name. We use "new" when the function creates a new object.

Regarding the free callback: If the elements are pointers, we call it with the pointer, but how do we call it when the elements are structs? Do we only allow a free callback if the element size is `sizeof(void *)`?

chzhoo:
> With this implementation, we can't allow users to just access element i as `arr->items[i]`. We need a function `arrayGet(i)` to access the elements. We should also rename `items` to `data` or something to indicate it's not a simple array of pointers.

Agree.

chzhoo:
> Regarding the free callback: If the elements are pointers, we call it with the pointer, but how do we call it when the elements are structs? Do we only allow a free callback if the element size is `sizeof(void *)`?

We should allow a free callback if the element size is larger than sizeof(void *). Let me illustrate the use of free callback with the following two examples.

```
//When the array stores sds, the free call is as follows
void sdsfreecb(void* ptr) {
    sds ele = *((sds*)ptr);
    sdsfree(ele);
}

//When the array stores struct a, the free call is as follows
typedef struct {
    sds str;
    ...
} a;
void structfreecb(void* ptr) {
    a* ele = (a*)ptr;
    sdsfree(a->str);
    ...
}
```

zuiderkwast:
> ```
> //When the array stores sds, the free call is as follows
> void sdsfreecb(void* ptr) {
>     sds ele = *((sds*)ptr);
>     sdsfree(ele);
> }
> ```

So the idea is that the free callback is called with a pointer to what's inside the array. It means that if the array stores simple pointers, we can't just use `zfree` as the free callback. To call zfree, we would need to do this:

```c
void freecb(void *pointer_to_pointer) {
    void *alloc = *(void **)pointer_to_pointer;
    zfree(alloc);
}
```

This is a little inconvenient. If we only allow pointer-size elements, this could be simpler. Then we could just use zfree directly the free callback.

```c
    arrayInit(&array_of_pointers, 16, zfree);
```

chzhoo:
> This is a little inconvenient. If we only allow pointer-size elements, this could be simpler. Then we could just use zfree directly the free callback.

If the array stores pointers, the API is a little inconvenient. This problem also bothers me. In addition to the free callback, the same situation also exists when `arrayGet/arrayPush` function is used. For details, please see the following usage examples.

```
array sds_array;
arrayInit(&sds_array, 10, sdsfreecb);

...
sds* container = arrayPush(sds_array);
*container = sdsdup(source_string);
...
for (size_t x = 0; x < arrayLen(sds_array); x ++) {
    sds *sds_item = arrayGet(sds_array,  x);
    ...
}

...
arrayCleanup(sds_array);

```
It can be seen that both `arrayPush/arrayGet` return a pointer to a pointer, which is also a bit inconvenient to use.

chzhoo:
Some open-source projects I've learned about, such as Nginx and Twemproxy, implement dynamic arrays that support variable-sized elements. When storing pointers, their API is also a little inconvenient. I assume this design caters to a wider range of scenarios, albeit with a slight trade-off in API convenience.

1. Array implementation of nginx: https://github.com/nginx/nginx/blob/master/src/core/ngx_array.h
2. Array implementation of twemproxy: https://github.com/twitter/twemproxy/blob/master/src/nc_array.h

zuiderkwast:
> this design caters to a wider range of scenarios, albeit with a slight trade-off in API convenience

Yes. I believe variable size is OK. I noticed in nginx and twemproxy arrays:

* None of these have a free callback. It makes the API simpler. Maybe we can skip it too? We need to refactor the scan code a little but it seems to be easy.
* They use a 32-bit int for the number of elements. It can save some memory when we have many small arrays.

chzhoo:
> * None of these have a free callback. It makes the API simpler. Maybe we can skip it too? We need to refactor the scan code a little but it seems to be easy.
> * They use a 32-bit int for the number of elements. It can save some memory when we have many small arrays.

OK. I will do like this.

chzhoo:
- The new code + unit tests have been submitted, and the code is now in a review-ready state
-  I've completed the new benchmark runs, and most test results are consistent with previous ones. See detailed result below.

**scan command benchmark result**
count | QPS after optimization | QPS before optimization | Performance Boost
-- | -- | -- | --
4 | 112663.36| 106643.91 | 5.6%
16 | 96274.19| 86937.62 | 10.7%
64 | 61519.54| 49308.45 | 24.8%
512 | 15028.78 | 9679.65| 55.3%

**hscan command benchmark result**
count | QPS after optimization | QPS before optimization | Performance Boost
-- | -- | -- | --
4 | 114161.77| 112663.36 | 1.3%
16 | 98745.93| 96371.61 | 2.5%
64 | 67778.23 | 61523.32 | 10.2%
512 | 20925.97 | 15426.15| 35.7%

**sscan command benchmark result**
count | QPS after optimization | QPS before optimization | Performance Boost
-- | -- | -- | --
4 | 119260.58| 117633.21 | 1.4%
16 | 112555.58| 108096.41 | 4.1%
64 | 89867.45| 85008.71 | 5.7%
512 | 40948.36 | 31572.63| 29.7%

**zscan command benchmark result**
count | QPS after optimization | QPS before optimization | Performance Boost
-- | -- | -- | --
4 | 97546.70| 96371.61 | 1.2%
16 | 71410.72|68355.03 | 4.5%
64 | 35032.41| 31723.37 | 10.4%
512 | 7679.13| 6393.62| 20.1%

src/array.c
```diff
@@ -0,0 +1,43 @@
...
+uint32_t arrayLen(array *a) {
+    return a->len;
+}
+
+void *arrayGet(array *a, uint32_t idx) {
+    assert(idx < a->len);
+    return (uint8_t *)a->data + idx * a->item_size;
+}
+
+void *arrayPush(array *a) {
```
zuiderkwast: In valkey, the documentation comments are usually in the c file before each function, so this is where many developers look for the comments. Can you add a short comment for each function?

For example for this one: Adds a new uninitialized element to the array and returns a pointer to it.

src/db.c
```diff
@@ -912,9 +913,17 @@ void keysCommand(client *c) {
     setDeferredArrayLen(c, replylen, numkeys);
 }
 
+#define DEFAULT_SCAN_COMMAND_COUNT 10
+
+/* The SCAN command's default COUNT is 10.
+ * Since it may store keys + values, the
+ * buffer size is roughly 10 * 2 = 20.
+ * Adding a 20% buffer (20 * 1.2) gives 24. */
+#define MAX_SCAN_ARRAY_BUFFER 24
```
zuiderkwast: It is not the max size but the initial alloc size, right?

```suggestion
#define SCAN_ARRAY_INITIAL_ALLOC 24
```

src/db.c
```diff
@@ -1161,29 +1173,25 @@ void scanGenericCommand(client *c, robj *o, unsigned long long cursor) {
...
-
-    list *keys = listCreate();
-    /* Set a free callback for the contents of the collected keys list if they
-     * are deep copied temporary strings. We must not free them if they are just
-     * a shallow copy - a pointer to the actual data in the data structure */
-    if (!shallow_copied_list_items) {
-        listSetFreeMethod(keys, sdsfreeVoid);
+        free_callback = sdsfree;
     }
+    arrayInit(&result, MAX_SCAN_ARRAY_BUFFER, sizeof(sds));
```
zuiderkwast: If `use_pattern` is false, we can actually know the number of elements. It is `count` and sometimes `count * 2` if values are included. We can know if values are included from `type` and `only_keys`.

Is is maybe two lines of code to calculate it. Is it worth it to avoid realloc?

zuiderkwast (commented):
Sorry for the delay. The array looks very good now and the unit test too. I have just a few minor comments.

src/db.c
zuiderkwast: Maybe it's not enough... If count is extremely huge, this can cause out of memory.. Maybe it's more safe to just use one initial value like you did.

zhulipeng:
I'm not being pedantic, but in my opinion, this should be called something like a Java `ArrayList` or a C++ `std::vector` rather than just an `array` in valkey.

src/db.c
```diff
@@ -969,8 +978,8 @@ void keysScanCallback(void *privdata, void *entry) {
     }
 
     /* Keep this key. */
-    list *keys = data->keys;
-    listAddNodeTail(keys, key);
+    sds *item = arrayPush(data->result);
+    *item = key;
```
zhulipeng: The API design is a little strange here, why not design like `arrayPush(data->result, key);`

zhulipeng:
> 1. The scan/sscan/hscan/zscan commands store iterated keys and values in a `list`, where each list element triggers a separate memory allocation and deallocation. This causes performance degradation when the list contains a large number of elements.
> 2. We can reduce memory allocation/deallocation by replacing `list` with `array` that have pre-allocated buffers.
> 3. For specific performance gains, refer to the benchmark results below.

The QPS improvement is impressive. Do you have the memory usage comparison data from before and after?

src/array.c
chzhoo: OK.

src/db.c
chzhoo: Your advice fits better.

chzhoo:
> I'm not being pedantic, but in my opinion, this should be called something like a Java `ArrayList` or a C++ `std::vector` rather than just an `array` in valkey.

OK. Using 'vector' is more explicit. Switching to vector shouldn't cause any issues.

src/db.c
chzhoo: For the rationale behind this design, refer to this [comment](https://github.com/valkey-io/valkey/pull/2160#issuecomment-2959930338).

zuiderkwast: Because element size is dynamic. With only pointer-sized elements we could use a more nice API.

zuiderkwast:
`vector` is an OK name too, but "dynamic array" is a computer science name for this, so `array` is not wrong IMHO.

https://en.wikipedia.org/wiki/Dynamic_array

chzhoo:
> > 1. The scan/sscan/hscan/zscan commands store iterated keys and values in a `list`, where each list element triggers a separate memory allocation and deallocation. This causes performance degradation when the list contains a large number of elements.
> > 2. We can reduce memory allocation/deallocation by replacing `list` with `array` that have pre-allocated buffers.
> > 3. For specific performance gains, refer to the benchmark results below.
> 
> The QPS improvement is impressive. Do you have the memory usage comparison data from before and after?

I compared the memory usage with the following command.

```
# Step 1: Load test data. Memory usage after loading test data is approximately 15.67M
valkey-benchmark -r 100000 -n 10000000 -P 32 set key-__rand_int__ value-__rand_int__


# Step 2: Run the SCAN command with different COUNT parameters.
valkey-benchmark --threads 2 -r 100000000 -n 20000000  scan __rand_int__ count $count

# Step 3: Monitor memory usage for one minute using the command below, and select the peak value as the result.
for((x=0;x<60;x+=1));do valkey-cli info |grep used_memory_rss_human:;sleep 1;done
```


$count | before optimization | after optimization 
-- | -- | --  
512| 16.18M| 16.05M
4096| 27.75M | 22.25M
65536| 123.11M | 120.86M

src/db.c
chzhoo: > If `use_pattern` is false, we can actually know the number of elements. It is `count` and sometimes `count * 2` if values are included. We can know if values are included from `type` and `only_keys`.
> 
> Is is maybe two lines of code to calculate it. Is it worth it to avoid realloc?

I initially considered implementing size prediction for array pre-allocation, but concluded it couldn't be achieved with minimal code. It requires simultaneous analysis of multiple factors – including data types, usage patterns, key-only scenarios, element count, and object types – which would complicate the logic. I wasn't sure if it's worth to do it, so I dropped it.

zuiderkwast: Sounds reasonable. Keep it simple.

src/vector.c
```diff
@@ -0,0 +1,64 @@
...
+    if (a->len == a->alloc) {
+        size_t alloc = a->alloc ? 2 * a->alloc : 8;
+        a->data = zrealloc(a->data, alloc * a->item_size);
+        a->alloc = alloc;
+    }
+
+    void *item = (uint8_t *)a->data + a->len * a->item_size;
+    a->len++;
+    return item;
+}
```
lipzhu, line 54:
```suggestion
void vectorPush(vector *a, const void* ele) {
    if (a->len == a->alloc) {
        size_t alloc = a->alloc ? 2 * a->alloc : 8;
        a->data = zrealloc(a->data, alloc * a->item_size);
        a->alloc = alloc;
    }

    void *item = (uint8_t *)a->data + a->len * a->item_size;
    memcpy(item, ele, a->item_size);
    a->len++;
}
```
```

File: src/db.c (hunk 9 above)
Comment (zhulipeng) [reply src/db.c]: ~~But we already know the element size when initialize the array. Am I misunderstanding?~~ 
Ok, I get the point.

Comment (chzhoo) [reply src/db.c]: 1. For the current use case (storing sds), we could indeed have a simpler API like `vectorPush(vector *vec, void* ele)`.
2.  However, we plan to reuse this vector implementation in other case (e.g., `ready_keys` in `struct valkeyServer`), where the elements are not pointers but larger structs (exceeding the size of a pointer).

File: src/vector.c [hunk 11]
```diff
@@ -0,0 +1,64 @@
...
+/* Get pointer to element at specified index. */
+void *vectorGet(vector *a, uint32_t idx) {
+    assert(idx < a->len);
+    return (uint8_t *)a->data + idx * a->item_size;
+}
+
+/* Adds a new uninitialized element to the array and returns a pointer to it. */
+void *vectorPush(vector *a) {
+    if (a->len == a->alloc) {
+        size_t alloc = a->alloc ? 2 * a->alloc : 8;
```
zuiderkwast: a->alloc is uint32_t.

```suggestion
        uint32_t alloc = a->alloc ? 2 * a->alloc : 8;
```

zuiderkwast (approved):
Great, just one nit.

zhulipeng (approved):

PR #2161, opened 2025-05-31
Fix case sensitivity of escapes in `stringmatch`

thaliaarchi:
When matching with case-insensitive patterns, escaped characters are case-insensitive when outside a character class, but not when inside. This is undocumented and appears to be unintentional. Since case-insensitive patterns are only used for matching command and function names with `COMMAND LIST FILTERBY PATTERN <pattern>` and `FUNCTION LIST LIBRARYNAME <pattern>`, this behavior is not useful and is only confusing.


zuiderkwast (commented):
Good catch!

Just to check if I understand the problem: case-insensitive pattern `Hello [a-z]orld` matches "hello world" but not "Hello World"?

Can you add a unit test in `src/unit/test_util.c`?

(We should probably also move the function `stringmatchlen_fuzz_test` to the unit tests. It's used in the DEBUG command, which was probably some ad-hoc testing long ago. We can move it to unit test instead.)

thaliaarchi:
> Just to check if I understand the problem: case-insensitive pattern `Hello [a-z]orld` matches "hello world" but not "Hello World"?

The `[a-z]` in that is normalized to `[a-z]` and characters matched against it are lowercased. Consequently, `[a-Z]` (which would match nothing) and `[A-z]` (which would also match the symbols between upper and lower cases) also become `[a-z]` and match the same. I should fix that. It could be done by folding the string's character to lower and upper, not by folding the glob's range.

This PR is about the difference between `\A` and `[\A]`: The former matches `[aA]`, but the latter only matches `A`. This difference seems unintentional, as, in this glob matcher, backslash is only used for escaping metacharacters (`*`, `?`, `[`, `]`, `\`).

thaliaarchi:
I'm relying on CI to check whether moving the fuzz test worked, since I've been working in a [fork](https://github.com/thaliaarchi/antirez-stringmatch) with only the string matcher.

Edit: I figured out local tests.

src/util.c
```diff
@@ -120,7 +120,6 @@ static int stringmatchlen_impl(const char *pattern,
                 if (patternLen >= 2 && pattern[0] == '\\') {
                     pattern++;
                     patternLen--;
-                    if (pattern[0] == string[0]) match = 1;
```
zuiderkwast, line 123: Are there no escapes special characters in a character ranges? Here it looks like you just skip all backslashes. What about patterns like `[\]]`, `[\\]`, `[\^]` and `[a\-z]`?

To me it looks like this line is needed, but it should be case-insensitive if nocase is given.

src/unit/test_util.c
```diff
@@ -11,6 +11,27 @@
...
+        int strlen = rand() % sizeof(str);
+        int patlen = rand() % sizeof(pat);
+        for (int j = 0; j < strlen; j++) str[j] = rand() % 128;
+        for (int j = 0; j < patlen; j++) pat[j] = rand() % 128;
+        total_matches += stringmatchlen(pat, patlen, str, strlen, 0);
+    }
+    TEST_ASSERT(total_matches <= 100000);
+    return 0;
+}
+
```
zuiderkwast, line 157: > This PR is about the difference between \A and [\A]: The former matches [aA], but the latter only matches A.

The fuzz test basically only tests that it doesn't crash. Can you add a test to cover the code you've changed, for example the patterns you mentioned?

Idea:

```suggestion

int test_stringmatchlen(int argc, char **argv, int flags) {
    UNUSED(argc);
    UNUSED(argv);
    UNUSED(flags);

    TEST_ASSERT(stringmatch("\\A", "a", 1));
    TEST_ASSERT(stringmatch("[\\A]", "a", 1));
    TEST_ASSERT(stringmatch("[\]]", "]", 1));

    TEST_ASSERT(!stringmatch("\\A", "a", 0));
    TEST_ASSERT(!stringmatch("[\\A]", "a", 0));
    TEST_ASSERT(stringmatch("[\]]", "]", 0));
}

```

src/util.c
```diff
@@ -135,14 +135,14 @@ static int stringmatchlen_impl(const char *pattern,
...
                     }
-                    if (nocase) {
-                        start = tolower(start);
-                        end = tolower(end);
-                        c = tolower(c);
+                    if (c >= start && c <= end) {
+                        match = 1;
+                    } else if (nocase && isalpha(c)) {
+                        c = islower(c) ? toupper(c) : tolower(c);
+                        if (c >= start && c <= end) match = 1;
```
zuiderkwast, line 148: > When nocase is enabled, ranges are folded to lowercase, so the ranges
[a-Z] (which would match nothing) and [A-z] (which would also match the
symbols between upper and lower cases) are both folded to [a-z] and
match the characters [a-zA-Z]. Instead, fold only the string's
characters.

I believe this is a behavior change that could break some user's applications.

I don't think we need to worry too much about the strange behavior in mixed-case character ranges like `[a-Z]` or something like `[A-_]`. It's safer to leave it as it is.

thaliaarchi: The problem is that by naïvely lowercasing a range, it can change the set of non-alphabetic characters that it matches. Instead, it should only *add* the case folded alphabetic characters to the set. That's what regular expression engines do for case folding (I've compared against JavaScript `RegExp` and Rust `regex`), so it's reasonable to expect that behavior here.

The fix is easy: First test the character against the range without changing cases. Then, if that fails, try again with the character's case swapped. The range's case is not changed. That's what patch 2 does.

This does not affect the common case of ranges which do not cross case boundaries (symbol/letter, upper/lower), which is what I think you were worried about.

Here's some examples which show every combination of the current behavior involving the bug. I'll write regular expressions in slashes. For clarity, I'll split alphabetic and non-alphabetic ranges, even when they're contiguous, in the regular expressions.

Ranges with two alphabetic bounds in the same case fold as we'd expect:

- `[i-o]`: `/[i-o]/` (nocase=0); ✅ `/[I-Oi-o]/` (nocase=1)
- `[I-O]`: `/[I-O]/` (nocase=0); ✅ `/[I-Oi-o]/` (nocase=1)

Ranges with alphabetic bounds in different cases fold incorrectly:

- `[i-O]` and `[O-i]`: ``/[O-Z\[-`a-i]/`` (nocase=0); 🐞 `/[i-o]/`, 🆕 ``/[A-IO-Z\[-`a-io-z]/`` (nocase=1)
- `[I-o]` and `[o-I]`: ``/[I-Z\[-`a-o]/`` (nocase=0); 🐞 `/[i-o]/`, 🆕 ``/[A-Z\[-`a-z]/`` (nocase=1)

Ranges with one alphabetic bound and one non-alphabetic bound fold incorrectly:

- `[i-~]`: `/[i-z{-~]/` (nocase=0); ✅ `/[I-Zi-z{-~]/` (nocase=1)
- `[I-~]`: ``/[I-Z\[-`a-z{-~]/`` (nocase=0); 🐞 `/[I-Zi-z{-~]/`, 🆕 ``/[A-Z\[-`a-z{-~]/`` (nocase=1)
- `[!-o]`: ``/!-@A-Z\[-`a-o]/`` (nocase=0); 🐞 ``/!-@A-O\[-`a-o]/``, 🆕 ``/[!-@A-Z\[-`a-z]/`` (nocase=1)
- `[!-O]`: `/!-@A-O]/` (nocase=0); 🐞 ``/!-@A-Z\[-`a-o]/``, 🆕 `/[!-@A-Oa-o]/` (nocase=1)

So, on the contrary, I think it's safer to fix this. The current folding behavior is unintuitive, unless you've studied the code. Fixing it brings us in line with regular expression engines.

src/unit/test_util.c
thaliaarchi: I'm working on comprehensive tests.

src/util.c
thaliaarchi: I didn't change backslash handling in ranges. Though, your comment is making me consider handling backslash escapes in a range end bound.

What I think you're missing here is that I factored out the `else` branch to be shared with this case. So it strips the backslash, then executes the normal single-character matching. The new `continue` in the range case is needed to skip that.

thaliaarchi:
I'm working on comprehensive tests.

I don't like that `TEST_ASSERT` immediately bails, so I've created a `TEST_EXPECT` assertion that is non-fatal and reports every failed `TEST_EXPECT` result in a function. Once I've finished my tests, I'll drop that commit, revert to `TEST_ASSERT`, and cherry-pick it to another PR where I change it back, since I think it would be a generally useful pattern.

src/util.c
zuiderkwast: Ah, right.

zuiderkwast (commented):
Sorry for the delay. The implementation LGTM and these tests are very comprehensive. Thanks! Some intersting undocumented behavior revieled there, like unclosed ranges.

How about adding some checks for mixed-case ranges like `[A-b]` and `[a-B]`?

Backward ranges like `[z-a]` is another undocumented feature I guess.

Then, it's a good idea to move the test framework changes to a separate PR so we can discuss those details separately.

thaliaarchi:
I'm putting some time in today to finish this up. I'm still not satisfied with the comprehensiveness of the tests, but, yes, it's better.

I've opened https://github.com/valkey-io/valkey/pull/2223 for the non-fatal assertions feature. Since Valkey squashes PRs, I'll separate out the new tests into yet another PR, once I'm ready. Then, this PR will show the changes to the tests for the bug fixes, even when squashed.

thaliaarchi:
I've written tests for many more cases and some for two of the four bug fixes. So far, tests are mostly for syntactic regularity, not runtime properties. It also needs regression tests for the CVEs. I'll write those and regression tests for the other bug fixes in another sitting.

Once you asked for tests, this became a much bigger thing. Sorry I'm slow.

Also, I found and fixed two more bugs: a pattern of just `*` doesn't match the empty string, but should, and escapes starting a character class range are parsed as a single escaped character. I have atomic commits for each. Once I'm done, since they'll unfortunately be squashed, I'll summarize it all in the description instead of the commit messages.

src/unit/test_util.c
```diff
@@ -11,6 +11,150 @@
...
 #endif
 
+int test_stringmatch(int argc, char **argv, int flags) {
+    UNUSED(argc);
+    UNUSED(argv);
+    UNUSED(flags);
+
+    TEST_BEGIN();
+
+#define TEST_STRINGMATCH(p, s, expectCase, expectNocase)   \
```
zuiderkwast, line 21: This macro is defined inside this test case. I think we should `undef` it at the end of the function to leave no trace.

PR #2165, opened 2025-06-03
Offload slot calculation and cross-slot detection to I/O threads

### LINKED ISSUE #2077: Avoid recalculating cluster slot when the I/O thread has already done so

zuiderkwast:
Refactors `getNodeByQuery()` to be able to move the CRC16 slot calculations in I/O threads and skip many checks if slot migrations are not ongoing.

The slot calculation for a command is moved out of `getNodeByQuery()` into a new function `clusterSlotByCommand()` which is safe to call from I/O threads.

For MULTI-EXEC transactions, the slot is stored per command in the multi-state, to be able to detect cross-slot transactions on EXEC without computing the slots again.

Sorry for the large diff within `getNodeByQuery()`. This function is quite long and complex. I added a `goto` to skip parts of it, to avoid indenting a large code block. (We can do it, but it gives a larger diff to review.)

Additionally, cross-slot detection and arity check is offloaded to I/O threads. To unify the code paths for commands parsed by I/O threads and commands parsed by the main thread, the command lookup, arity check, slot lookup are moved out of `processCommand()` to a new `prepareCommand()` function that needs to be called before `processCommand()`.

The client's read flags are used for passing information about bad arity and cross-slot. These flags are already used to convey information per command between I/O thread and main thread.

Fixes #2077
Related to #632

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -962,7 +962,43 @@ void clusterCommand(client *c) {
     }
 }
 
+/* Compute cluster slot for the given command and arguments and detect cluster
+ * cross-slot errors. Returns the slot. If -1 is returned, one of the flags
+ * READ_FLAGS_NO_KEYS or READ_FLAGS_CROSSSLOT is set in `*read_flags` to
+ * indicate why the slot couldn't be determined. This function has no
+ * side-effects and can be called from I/O threads. */
+int clusterSlotByCommand(struct serverCommand *cmd, robj **argv, int argc, int *read_flags) {
+    getKeysResult result;
+    initGetKeysResult(&result);
+    int numkeys = getKeysFromCommand(cmd, argv, argc, &result);
+    int slot = -1;
+    if (numkeys == 0) *read_flags |= READ_FLAGS_NO_KEYS;
+    for (int i = 0; i < numkeys; i++) {
+        sds key = argv[result.keys[i].pos]->ptr;
+        int keyslot = keyHashSlot(key, sdslen(key));
+        if (slot == -1) {
+            slot = keyslot;
+        } else if (keyslot != slot) {
+            slot = -1;
+            *read_flags |= READ_FLAGS_CROSSSLOT;
+            break;
+        }
+    }
+    getKeysFreeResult(&result);
+    return slot;
+}
+
 /* Return the pointer to the cluster node that is able to serve the command.
+ *
+ * Note that this function doesn't compute the slot for each key. The client's
+ * slot needs to be set before calling this function. If it's set to -1, the
+ * client's read flags should indicate if it's a cross-slot command or if the
+ * command has no keys. This is computed by clusterSlotByCommand(). The commands
+ * can be called together like this:
+ *
+ *     c->slot = clusterSlotByCommand(c->cmd, c->argv, c->argc, &c->read_flags);
+ *     node = getNodeByQuery(c, &error_code);
+ *
  * For the function to succeed the command should only target either:
  *
  * 1) A single key (even multiple times like RPOPLPUSH mylist mylist).
@@ -994,15 +1030,17 @@ void clusterCommand(client *c) {
  *
  * CLUSTER_REDIR_DOWN_STATE and CLUSTER_REDIR_DOWN_RO_STATE if the cluster is
  * down but the user attempts to execute a command that addresses one or more keys. */
-clusterNode *
-getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int *hashslot, int *error_code) {
+clusterNode *getNodeByQuery(client *c, int *error_code) {
     clusterNode *myself = getMyClusterNode();
     clusterNode *n = NULL;
     robj *firstkey = NULL;
     int multiple_keys = 0;
     multiState *ms, _ms;
     multiCmd mc;
-    int i, slot = 0, migrating_slot = 0, importing_slot = 0, missing_keys = 0, existing_keys = 0;
+    int i, migrating_slot = 0, importing_slot = 0, missing_keys = 0, existing_keys = 0;
+
+    /* Slot must be calculated in advance, or cross-slot or no keys detected. */
+    serverAssert(c->slot >= 0 || c->read_flags & (READ_FLAGS_NO_KEYS | READ_FLAGS_CROSSSLOT));
 
     /* Allow any key to be set if a module disabled cluster redirections. */
     if (server.cluster_module_flags & CLUSTER_MODULE_FLAG_NO_REDIRECTION) return myself;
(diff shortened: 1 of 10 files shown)


src/cluster.c
```diff
@@ -1014,9 +1052,63 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
      * when writing a module that implements a completely different
      * distributed system. */
 
+    /* Determine transaction slot and return early on cross-slot. */
+    if (c->cmd->proc == execCommand && c->flag.multi) {
+        int slot = -1;
```
ranshid, line 1061: seems strange to the reader of the code IMO. the c->slot MUST be defined at this point so we can just match all transaction commands to that right?

src/networking.c
```diff
@@ -3240,6 +3240,23 @@ int canParseCommand(client *c) {
     return 1;
 }
 
+/* Prepare a parsed command for processing, including looking up the command,
+ * checking arity and calculating cluster slot. This can be done by I/O threads
+ * to offload the main-thread. */
+static void prepareCommand(client *c) {
+    c->io_parsed_cmd = lookupCommand(c->argv, c->argc);
+    if (c->io_parsed_cmd && commandCheckArity(c->io_parsed_cmd, c->argc, NULL) == 0) {
```
ranshid: If we do it here, shouldn't we also drop it in processCommand?

src/server.c
```diff
@@ -4045,7 +4045,7 @@ int processCommand(client *c) {
      * In case we are reprocessing a command after it was blocked,
      * we do not have to repeat the same checks */
     if (!client_reprocessing_command) {
-        struct serverCommand *cmd = c->io_parsed_cmd ? c->io_parsed_cmd : lookupCommand(c->argv, c->argc);
+        struct serverCommand *cmd = c->io_parsed_cmd;
```
ranshid: Not that important, but since we now operate the command lookup in a different place we have no way to distinguish a case were we missed prepareCommand and a case were the command does not exist (e.g error).
Maybe we can add a flag indicating the command was not found in prepareCommand and assert otherwise when the c->io_parsed_cmd does not exist?

This is mainly for making sure we called prepareCommand

ranshid (commented):
Looks good, but I will probably need a 2nd look. some small comments for now

src/cluster.c
zuiderkwast: Yes, `c->slot` needs to be defined in advance when executing the command, but it is for the current command. EXEC has no keys so `c->slot == -1` here.

I should probably clarify this in the comments.

src/networking.c
zuiderkwast: Yes we could, but commandCheckArity also produces an error message that we use in processCommand. The arity check is very cheap and trivial so I just just kept it for now.

It's only this `if` check:

```C
int commandCheckArity(struct serverCommand *cmd, int argc, sds *err) {
    if ((cmd->arity > 0 && cmd->arity != argc) || (argc < -cmd->arity)) {
```

src/server.c
zuiderkwast: Good idea.

I also considered renaming `io_parsed_cmd` to just `parsed_cmd` because it's not only for IO threads anymore.

src/networking.c
ranshid: In that case I think I missed the purpose of the READ_FLAGS_BAD_ARITY. why is it always turned on for cluster mode after command being pre-processed?

zuiderkwast: The answer is in the prefetching step. If it's bad arity, we can't extract the keys from argv and calculate the cluster slot.

Before this, the IO thread just set the command (c->io_parsed_cmd) to NULL to skip prefetching and then processCommand would look up the command again.

src/cluster.c
ranshid: Oh. I get it... sure. a comment would probably be helpful

src/networking.c
zuiderkwast: > why is it always turned on for cluster mode after command being processed?

What! That's a typo! Sorry... Why did the tests still pass? We just skipped prefetching.

I'll make sure we actually check this flag in processCommand(). It would fail the tests.

ranshid: ACK. so can't we just use this flag in order to skip the test in process Command? :) I understand it will probably gain 0 performance improvements, but it makes more sense to avoid running the test again IMO

ranshid: > I'll make sure we actually check this flag in processCommand()

Or maybe that is what you meant?

zuiderkwast: Yeah I meant let's check the flag in processCommand. Then, if it's bad arity, we can call the function again to get the error message. See the last commit: 298e53f4dceb187315d794007f52bd8d6764b955

src/cluster.c
```diff
@@ -994,15 +1034,17 @@ void clusterCommand(client *c) {
...
-getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int *hashslot, int *error_code) {
+clusterNode *getNodeByQuery(client *c, int *error_code) {
     clusterNode *myself = getMyClusterNode();
     clusterNode *n = NULL;
     robj *firstkey = NULL;
     int multiple_keys = 0;
     multiState *ms, _ms;
     multiCmd mc;
-    int i, slot = 0, migrating_slot = 0, importing_slot = 0, missing_keys = 0, existing_keys = 0;
+    int i, migrating_slot = 0, importing_slot = 0, missing_keys = 0, existing_keys = 0;
```
ranshid, line 1044:
```suggestion
    int i, migrating_slot = 0, importing_slot = 0, missing_keys = 0, existing_keys = 0;
    struct serverCommand *cmd = c->cmd;
    robj **argv = c->argv;
    int argc = c->argc;
```

src/cluster.c
```diff
@@ -1014,9 +1056,65 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
...
+
+    /* For migrating and importing slots, we need to go over all the keys to
+     * count existing keys and missing keys that we need for TRYAGAIN and ASK
+     * redirects. Skip this if we're not importing or migrating. */
+    if (!migrating_slot && !importing_slot) goto after_checking_each_key;
+
     /* We handle all the cases as if they were EXEC commands, so we have
      * a common code path for everything */
-    if (cmd->proc == execCommand) {
+    if (c->cmd->proc == execCommand) {
```
ranshid, line 1121:
```suggestion
    if (cmd->proc == execCommand) {
```

src/cluster.c
```diff
@@ -1028,21 +1126,14 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
         ms = &_ms;
         _ms.commands = &mc;
         _ms.count = 1;
-        mc.argv = argv;
-        mc.argc = argc;
-        mc.cmd = cmd;
+        mc.argv = c->argv;
+        mc.argc = c->argc;
+        mc.cmd = c->cmd;
```
ranshid, line 1135:
```suggestion
        mc.argv = argv;
        mc.argc = argc;
        mc.cmd = cmd;
```

ranshid (commented):
nit - just to reduce the diff little bit.

src/cluster.c
```diff
@@ -1014,9 +1056,65 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
...
+
+    uint64_t cmd_flags = getCommandFlags(c);
+
+    /* Only valid for sharded pubsub as regular pubsub can operate on any node and bypasses this layer. */
+    int pubsubshard_included =
+        (cmd_flags & CMD_PUBSUB) || (c->cmd->proc == execCommand && (c->mstate->cmd_flags & CMD_PUBSUB));
+
+    /* For migrating and importing slots, we need to go over all the keys to
+     * count existing keys and missing keys that we need for TRYAGAIN and ASK
+     * redirects. Skip this if we're not importing or migrating. */
```
ranshid: can we also state here that in case no slot migration we can also skip some multi-db commands which we do not permit during that phase?

src/cluster.c
zuiderkwast: I see what you mean.

I did this change locally, but I see argv and argc are only used in one place. c->cmd is used in a few places, but it seems easier to read and understand the code if we use `c->cmd` directly in code that also refer to other client fields. I mean I have a slight preference for green the the code below, because it also checks `c->flag.multi` and `c->mstate`:

```diff
 
     /* Determine transaction slot and return early on cross-slot. */
+    if (c->cmd->proc == execCommand && c->flag.multi) {
-    if (cmd->proc == execCommand && c->flag.multi) {
         int slot = -1;
         for (i = 0; i < c->mstate->count; i++) {
```

src/cluster.c
zuiderkwast: We skip checking for them, yeah I can add that.

src/cluster.c
ranshid: Valid. I agree it also makes sense

ranshid (approved):
LGTM

src/cluster.c
```diff
@@ -1014,9 +1056,69 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
...
+        }
+    }
+
+    uint64_t cmd_flags = getCommandFlags(c);
+
+    /* Only valid for sharded pubsub as regular pubsub can operate on any node and bypasses this layer. */
+    int pubsubshard_included =
+        (cmd_flags & CMD_PUBSUB) || (c->cmd->proc == execCommand && (c->mstate->cmd_flags & CMD_PUBSUB));
+
+    /* If we're importing or migration the slot, we need to do some more checks:
```
madolson:
```suggestion
    /* If we're importing or migrating the slot, we need to do some more checks:
```

madolson (approved):
No real concerns for me, looks good

ranshid (approved):

src/server.c
```diff
@@ -4045,22 +4074,28 @@ int processCommand(client *c) {
...
+        struct serverCommand *cmd = c->parsed_cmd;
         if (!cmd) {
             /* Handle possible security attacks. */
             if (!strcasecmp(c->argv[0]->ptr, "host:") || !strcasecmp(c->argv[0]->ptr, "post")) {
                 securityWarningCommand(c);
                 return C_ERR;
             }
+            /* Check that the command lookup should has been done before calling
+             * this function, by calling prepareCommand(). */
+            serverAssert(!(c->read_flags & READ_FLAGS_COMMAND_NOT_FOUND));
```
zuiderkwast, line 4086: This assert is flipped, but it turns out we never set this flag.

The correct behavior is to check that the flag is set here, and to actually set the flag when we lookup the command and it's not found.

I'll fix it in another PR.

dvkashapov:
Did some benchmarks, [results](https://github.com/valkey-io/valkey/issues/2077#issuecomment-2943193880) are in related issue

src/cluster.c
```diff
@@ -1014,9 +1056,69 @@ getNodeByQuery(client *c, struct serverCommand *cmd, robj **argv, int argc, int
      * when writing a module that implements a completely different
      * distributed system. */
 
+    /* Determine transaction slot and return early on cross-slot. */
+    if (c->cmd->proc == execCommand && c->flag.multi) {
+        int slot = -1;
```
PingXie, line 1061: maybe introduce a new `mc_slot` field in `client` and keep it updated after parsing every child command. this way we can avoid this loop, making the code easier to read and also less code to execute.

PingXie (commented):
sorry for being late, again :) looks great to me - just one minor comment and I think `getNodeByQuery` could be further simplified. thanks @zuiderkwast

PR #2166, opened 2025-06-03
Trigger that will be used to send version info over to Extensions repo

Nikhil-Manglore:
We want to automate the valkey-extensions repository so we don't have to manually update it whenever a new version of Valkey is released. Tasks like version updates and Dockerfile modifications in the valkey-extensions repository are handled manually, leading to potential inconsistencies. The goal is to automate these processes through a pipeline that can handle version management, dynamic Dockerfile generation, and streamlined build workflows.

This pull request is an additional trigger which will activate when a new version of Valkey has been released (RC or GA). Once activated, it will send the version over to a workflow in the valkey-extensions repository which will then carry out all the processes listed above.

We can use the same Personal Access Token (which has already been created) for both triggers.


ranshid:
maybe I am not an expert on github workflows, but why isn't it another step in trigger-build-release.yml ?

Nikhil-Manglore:
I had thought about that earlier but had kept them separate when I was still writing the automation in case I wanted to change something. Now I agree we can just combine the two, I will update the PR soon.

.github/workflows/trigger-build-release.yml
```diff
@@ -48,3 +48,15 @@ jobs:
               "version": "${{ steps.determine-vars.outputs.version }}",
               "environment": "${{ steps.determine-vars.outputs.environment }}"
             }
+
+      - name: Trigger Valkey Extensions Update
+        uses: peter-evans/repository-dispatch@ff45666b9427631e3450c54a1bcbee4d9ff4d7c0 # Version 3
```
sarthakaggarwal97, line 53: curious why we use this particular action, I guess the commit won't change but can this commit be deleted? I see it's been used before in the file as well.

sarthakaggarwal97 (commented):
LGTM!

roshkhatri:
I think this would have be a part of the release automation as valkey-bundle has a dependency on valkey-contianers releasing the latest version.

This is because valkey-bundle uses the latest valkey image as the base image.

Nikhil-Manglore:
Closing this PR, will move the bundle trigger to valkey-release-automation.

PR #1949, opened 2025-04-13
Introduce atomic slot migration

murphyjacob4:
Introduces a new family of commands for migrating slots via replication. The procedure is driven by the source node which pushes an AOF formatted snapshot of the slots to the target, followed by a replication stream of changes on that slot (a la manual failover).

This solution is an adaptation of the solution provided by @enjoy-binbin, combined with the solution I previously posted at #1591, modified to meet the designs we had outlined in #23. 

## New commands

* `CLUSTER MIGRATESLOTS SLOTSRANGE start end [start end]... NODE node-id`: Begin sending the slot via replication to the target. Multiple targets can be specified by repeating `SLOTSRANGE ... NODE ...`
*  `CLUSTER CANCELMIGRATION ALL`: Cancel all slot migrations
* `CLUSTER GETSLOTMIGRATIONS`: See a recent log of migrations

This PR only implements "one shot" semantics with an asynchronous model. Later, "two phase" (e.g. slot level replicate/failover commands) can be added with the same core.

## Slot migration jobs

Introduces the concept of a slot migration job. While active, a job tracks a connection created by the source to the target over which the contents of the slots are sent. This connection is used for control messages as well as replicated slot data. Each job is given a 40 character random name to help uniquely identify it.

All jobs, including those that finished recently, can be observed using the `CLUSTER GETSLOTMIGRATIONS` command.

## Replication

* Since the snapshot uses AOF, the snapshot can be replayed verbatim to any replicas of the target node.
* We use the same proxying mechanism used for chaining replication to copy the content sent by the source node directly to the replica nodes.

## `CLUSTER SYNCSLOTS`

To coordinate the state machine transitions across the two nodes, a new command is added, `CLUSTER SYNCSLOTS`, that performs this control flow.

Each end of the slot migration connection is expected to install a read handler in order to handle `CLUSTER SYNCSLOTS` commands:

* `ESTABLISH`: Begins a slot migration. Provides slot migration information to the target and authorizes the connection to write to unowned slots.
* `SNAPSHOT-EOF`: appended to the end of the snapshot to signal that the snapshot is done being written to the target.
* `PAUSE`: informs the source node to pause whenever it gets the opportunity
* `PAUSED`: added to the end of the client output buffer when the pause is performed. The pause is only performed after the buffer shrinks below a configurable size
* `REQUEST-FAILOVER`: request the source to either grant or deny a failover for the slot migration. The grant is only granted if the target is still paused. Once a failover is granted, the paused is refreshed for a short duration
[...]

PR #2167, opened 2025-06-03
Improve CLUSTER FLUSHSLOT routing and propagation

murphyjacob4:
Includes three changes:

1. **Route CLUSTER FLUSHSLOT based on the target slot**: Introduces a `USES_SLOT` flag to key_specs that allows for key_specs to denote target slots instead of individual keys. This allows CLUSTER FLUSHSLOT to return -MOVED for unowned slots, and will help with determining slot migrations to propagate to in #1949 
2. **Propagate as CLUSTER FLUSHSLOT**: When we execute delKeysInSlot after a cluster topology update that results in a primary losing some slots, this will now propagate as as single `CLUSTER FLUSHSLOT` rather than an `UNLINK` command for each key in the slot. This will be useful for #1949 when a migration is completed. 

I also attempted a change to delete the slot hashtable instead of iterating over each key in the hashtable and deleting one-by-one. However, it turns out there are events that are triggered by unlinking a specific key that would require iteration, accessing the key's value, and triggering. Overall this probably won't give much of an improvement unless we can figure out a better story for triggering keyspace notification, so not including in this PR, but noting for posterity.


murphyjacob4:
Hmm, looks like this breaks the module API since it deletes the keys from the dictionary after the keyspace event, will add a commit to fix this

murphyjacob4:
Removed the code from this PR that supported drop the slot hashtable from the DB instead of unlinking each key directly. The issue is that there are some special module events (`moduleNotifyKeyUnlink`) that we would still need to iterate and trigger for each key.

Ideally we can align FLUSHSLOT with FLUSHDB for module events - and have modules handle slot flush as a separate event rather than requiring each key to be broadcasted individually. But this would be a breaking change

madolson:
> Ideally we can align FLUSHSLOT with FLUSHDB for module events - and have modules handle slot flush as a separate event rather than requiring each key to be broadcasted individually. But this would be a breaking change

We could make a breaking change for atomic slot migration and modules in 9.0.

src/cluster_legacy.c
```diff
@@ -6278,13 +6278,26 @@ const char *clusterGetMessageTypeString(int type) {
...
-/* Get the slot from robj and return it. If the slot is not valid,
- * return -1 and send an error to the client. */
-int getSlotOrReply(client *c, robj *o) {
+int getSlotOrError(robj *o, sds *err_out) {
     long long slot;
 
     if (getLongLongFromObject(o, &slot) != C_OK || slot < 0 || slot >= CLUSTER_SLOTS) {
-        addReplyError(c, "Invalid or out of range slot");
+        if (err_out) {
+            *err_out = sdsnew("Invalid or out of range slot");
```
madolson: This seems to be a constant, this could probably just be a char *.

src/cluster_legacy.c
```diff
@@ -6543,7 +6556,30 @@ unsigned int delKeysInSlot(unsigned int hashslot, int lazy, bool propagate_del,
      * state so that we don't assert in propagateNow(). */
     server.server_del_keys_in_slot = 1;
     unsigned int j = 0;
+    robj *argv[4];
     int before_execution_nesting = server.execution_nesting;
+    enterExecutionUnit(1, 0);
+
+    /* Propagate as a single CLUSTER FLUSHSLOT <slot> ASYNC/SYNC. */
+    if (propagate_del) {
```
madolson: Do the hooks fire if the replica executes flushslot? Couldn't the module hooks also need to get executed on the replica? I would imagine something like search might break.

src/cluster_legacy.c
```diff
@@ -6543,7 +6556,30 @@ unsigned int delKeysInSlot(unsigned int hashslot, int lazy, bool propagate_del,
      * state so that we don't assert in propagateNow(). */
     server.server_del_keys_in_slot = 1;
     unsigned int j = 0;
+    robj *argv[4];
     int before_execution_nesting = server.execution_nesting;
+    enterExecutionUnit(1, 0);
```
madolson, line 6561: Do we need two execution units here? It seems like just one would be enough to cover both cases.

src/cluster_legacy.c
```diff
@@ -6543,7 +6556,30 @@ unsigned int delKeysInSlot(unsigned int hashslot, int lazy, bool propagate_del,
...
+        argv[0] = shared.cluster;
+        argv[1] = shared.flushslot;
+        argv[2] = createStringObjectFromLongLong(hashslot);
+        argv[3] = lazy ? shared.async : shared.sync;
+        incrRefCount(argv[0]);
+        incrRefCount(argv[1]);
+        incrRefCount(argv[3]);
+        /* DB number doesn't matter since FLUSHSLOT is a cross-DB operation,
+         * just hardcode zero. */
+        alsoPropagate(/*dbid=*/0, argv, 4, PROPAGATE_AOF | PROPAGATE_REPL);
```
madolson: Worth mentioning that if the replica is not on zero, it will emit an unnecessary select.

src/cluster_legacy.c
```diff
@@ -6543,7 +6556,30 @@ unsigned int delKeysInSlot(unsigned int hashslot, int lazy, bool propagate_del,
...
+    /* Propagate as a single CLUSTER FLUSHSLOT <slot> ASYNC/SYNC. */
+    if (propagate_del) {
+        enterExecutionUnit(1, 0);
+        argv[0] = shared.cluster;
+        argv[1] = shared.flushslot;
+        argv[2] = createStringObjectFromLongLong(hashslot);
+        argv[3] = lazy ? shared.async : shared.sync;
+        incrRefCount(argv[0]);
+        incrRefCount(argv[1]);
+        incrRefCount(argv[3]);
```
madolson: alsoPropagate shouldn't decrement the refcount, so these shouldn't be needed.

src/cluster_legacy.c
madolson: The replica might also not support `CLUSTER FLUSHSLOT`, we should check the version of the replica. Otherwise it's a backwards incompatible change.

madolson (commented):
It's a bit weird that only this command will do redirects, but it seems like a step in the right direction.

src/cluster_legacy.c
murphyjacob4: Makes sense

src/cluster_legacy.c
murphyjacob4: Whoops, I meant to only have the top-level one

src/cluster_legacy.c
murphyjacob4: Actually, I guess -1 is supported in this case

src/cluster_legacy.c
murphyjacob4: Hmm, yeah I lifted this from `propagateDeletion` which increments the refcount. Maybe this is just an artifact of previous behavior, since I see that it shouldn't be required.

src/cluster_legacy.c
murphyjacob4: > Do the hooks fire if the replica executes flushslot

Yeah - so CLUSTER FLUSHSLOT will trigger delKeysInSlot (this function) with `send_del_event == true`. `send_del_event` tells the function to propagate keyspace notifications to clients and modules, whereas if `false`, it would just propagate to modules.

Is CLUSTER FLUSHSLOT something we are going to let users do? I seem to remember conversation that this would be internal only. If we go the internal-only route - it makes sense to me that we would just fire the module events on CLUSTER FLUSHSLOT (`send_del_event == false`)

I guess there is a discrepancy where the primary would not fire keyspace events to clients, but the replica would. So we should align those behaviors.

Any objection to making CLUSTER FLUSHSLOT not send keyspace notifications to clients? The only downside is if a user triggers CLUSTER FLUSHSLOT - we should differentiate (probably through an additional argument) since it is a "true deletion" (no longer stored in cluster) vs a "local deletion" (due to migration, but still stored in the cluster). But if we aren't making this an end-user command, I think we can just assume this is from "local deletion"

murphyjacob4: > The replica might also not support CLUSTER FLUSHSLOT, we should check the version of the replica. Otherwise it's a backwards incompatible change.

Good point. I've addressed this now where we only send CLUSTER FLUSHSLOT to replicas if they are all on 9.0.0+

madolson: > I guess there is a discrepancy where the primary would not fire keyspace events to clients, but the replica would. So we should align those behaviors.

Yeah. I agree that consistency is more important.

hpatro:
Unrelated but feels like a good thread to check about improvement around FLUSHSLOT, Would it be beneficial to introduce FLUSHSLOTSRANGE as well ?

src/cluster_legacy.c
```diff
@@ -6543,7 +6556,34 @@ unsigned int delKeysInSlot(unsigned int hashslot, int lazy, bool propagate_del,
...
+    }
+
+    /* Propagate as a single CLUSTER FLUSHSLOT <slot> ASYNC/SYNC if replicas
+     * support it. */
+    if (propagate_del && !legacy_replica) {
+        argv[0] = shared.cluster;
+        argv[1] = shared.flushslot;
+        argv[2] = createStringObjectFromLongLong(hashslot);
+        argv[3] = lazy ? shared.async : shared.sync;
+        alsoPropagate(/*dbid=*/-1, argv, 4, PROPAGATE_AOF | PROPAGATE_REPL);
```
hpatro: nit:

```suggestion
        alsoPropagate(-1, argv, 4, PROPAGATE_AOF | PROPAGATE_REPL);
```

src/cluster_legacy.c
```diff
@@ -6278,13 +6278,26 @@ const char *clusterGetMessageTypeString(int type) {
...
 
     if (getLongLongFromObject(o, &slot) != C_OK || slot < 0 || slot >= CLUSTER_SLOTS) {
-        addReplyError(c, "Invalid or out of range slot");
+        if (err_out) {
+            *err_out = "Invalid or out of range slot";
+        }
+        return -1;
+    }
+
+    return (int)slot;
```
hpatro: Maybe we can introduce `getIntFromObject` helper.

src/cluster_legacy.c
madolson: https://github.com/valkey-io/valkey/pull/2175 Yeah, it looks like just dead code to me.

src/cluster_legacy.c
```diff
@@ -6278,13 +6278,26 @@ const char *clusterGetMessageTypeString(int type) {
...
+}
+
+/* Get the slot from robj and return it. If the slot is not valid,
+ * return -1 and send an error to the client. */
+int getSlotOrReply(client *c, robj *o) {
+    char *err = NULL;
+    int slot = getSlotOrError(o, &err);
+    if (err) {
+        addReplyErrorSds(c, sdsnew(err));
```
hpatro: Would this suffice?
```suggestion
        addReplyError(c, err);
```

src/cluster_legacy.c
```diff
@@ -6278,13 +6278,26 @@ const char *clusterGetMessageTypeString(int type) {
...
         return -1;
     }
     return (int)slot;
```
```suggestion
    return slot;
```

hpatro (commented):
mostly nit picks.

hwware:
I just begin going through the code changes, one question for the flag name: USES_SLOT.  Why it includes an 'S' instead of use_slot, or using_slot, I am a little bit confused the Singular and plural

murphyjacob4:
> one question for the flag name: USES_SLOT. Why it includes an 'S' instead of use_slot, or using_slot

Perhaps USING_SLOT is a little more readable. I was thinking grammatically along the lines of "this key_spec **_uses_** a slot instead of a key"

src/cluster_legacy.c
```diff
@@ -6278,13 +6278,26 @@ const char *clusterGetMessageTypeString(int type) {
     return "unknown";
 }
 
-/* Get the slot from robj and return it. If the slot is not valid,
- * return -1 and send an error to the client. */
-int getSlotOrReply(client *c, robj *o) {
+int getSlotOrError(robj *o, char **err_out) {
```
hwware, line 6281: I think the functions getSlotOrError and getSlotOrReply are overlapped in some logic, I prefre to combine them as getSlotOrReply(client *c, robj *o, char* message)

murphyjacob4: How would that work? We would return an error if message is provided, or a reply if client is provided?

To me, they don't overlap in logic, it is just that getSlotOrReply is wrapping getSlotOrError and sending that to the client. No logic is repeated, it is just composed of the other function. Not all the times we want to turn an robj to a slot do we want to require a client response to be involved.

hwware: It works like this way? 

int getSlotOrReply(client *c, robj *o, char **err_out) {
    long long slot;

    if (getLongLongFromObject(o, &slot) != C_OK || slot < 0 || slot >= CLUSTER_SLOTS) {
        *err_out = "Invalid or out of range slot";
         addReplyErrorSds(c, sdsnew(err));
         return -1;
    }

    return (int)slot;
}

murphyjacob4: In the case for this PR we want to parse the slot but not add a reply to the client (since we leave this to the actual command handler). This is why they are two functions (one does not reply to the client)

madolson:
> Perhaps USING_SLOT is a little more readable. I was thinking grammatically along the lines of "this key_spec uses a slot instead of a key"

I was expecting to be more like, `IS_SLOT`. It's less that the keyspec is using a slot, than that it's defining where the slot really is.

src/cluster_legacy.c
```diff
@@ -6543,7 +6556,34 @@ unsigned int delKeysInSlot(unsigned int hashslot, int lazy, bool propagate_del,
...
+    listRewind(server.replicas, &replicas_iter);
+    int legacy_replica = 0;
+    while ((replicas_list_node = listNext(&replicas_iter)) != NULL) {
+        client *replica = listNodeValue(replicas_list_node);
+        /* 0x90000 is 9.0.0, when CLUSTER FLUSHSLOT was added. */
+        if (replica->repl_data->replica_version < 0x90000) {
+            legacy_replica = 1;
+            break;
+        }
+    }
```
madolson: I expect we'll use this pattern a lot, we might want to make a macro or function to help here.

madolson (commented):
Mostly minor comments. I'm only like 80% convinced routing the `flushslot` command based on custom routing is really needed, but I also think it's a major decision since it will require client side changes.

src/cluster_legacy.c
murphyjacob4: Done!

src/cluster_legacy.c
murphyjacob4: Done

src/cluster_legacy.c
murphyjacob4: Added `primaryGetOldestReplicaVersion`

src/cluster_legacy.c
murphyjacob4: Sure, I guess this is just a Google style thing (https://google.github.io/styleguide/cppguide.html#Function_Argument_Comments). But I'll remove it

src/cluster_legacy.c
murphyjacob4: I guess this is the existing behavior, since delKeysInSlot will not generate keyspace events on the primary, but the replicated `UNLINK` will. Triggering `CLUSTER FLUSHSLOT` on the replica is the same effect.

src/cluster_legacy.c
murphyjacob4: Makes sense to me

murphyjacob4:
> I was expecting to be more like, IS_SLOT. It's less that the keyspec is using a slot, than that it's defining where the slot really is.

Yeah that naming sounds good to me

> Mostly minor comments. I'm only like 80% convinced routing the flushslot command based on custom routing is a good idea, but I also think it's a major decision since it will require client side changes.

Right... from an immediate usability perspective, I guess the other mechanism (any node will accept CLUSTER FLUSHSLOT, but it will be a no-op if it is not owned) could be easier to program against for clients, since the client could just send `CLUSTER FLUSHSLOT X` to every node. You could even do this today in `valkey-py` with something like `cluster.execute_command("CLUSTER FLUSHSLOT 16383", target=valkey.ALL)`.

But on the other hand - it does seem unintuitive - if I use `valkey-cli` on a node and call `CLUSTER FLUSHSLOT X` and get back `+OK` - I would kind of expect that the slot is flushed (even if it isn't owned locally).

Since this is the first slot-level write command that we are adding, I feel like we _should_ probably route it as we would a multi-key write command affecting all keys in the slot would be routed. A quick look at the code for valkey-py shows that they already need functionality like this for commands like SETSLOT: https://github.com/valkey-io/valkey-py/blob/main/valkey/asyncio/cluster.py#L628-L630

But yeah - let's discuss in a wider group

madolson:
> But yeah - let's discuss in a wider group

`CLUSTER SCAN` might also do the same thing, so it seems like there will be more future commands that need this routing behavior.

murphyjacob4:
Discussed with the wider group. We think that since it is a net new command, their is low likelihood of the new routing breaking client applications. If users want to use it and their client doesn't natively support it, they can send it as a raw command and their client should handle the MOVED redirect

murphyjacob4:
@valkey-io/core-team Please vote 👍/👎

PR #2176, opened 2025-06-04
Run external cluster tests with multiple databases

zuiderkwast:
Change the runtest flag `--cluster-mode` to no longer imply `--single-db`. This allows more tests to run against an external node running in cluster mode.

Also fix incorrectly tagged tests so that `./runtest --singledb` passes locally.

Without this change, this doesn't pass because several tests are incorrectly tagged as external:skip instead of singledb:skip or cluster:skip. In the CI, we apparently only run with singledb when implied by `--cluster-mode` (until now), which we only use in the external cluster mode test. That's why the incorrect tags were not detected.

* singledb:skip is used for tests that do select and assume that a client has selected db 9.
* cluster:skip is used when the test uses SWAPDB, cross-slot commands other things forbidden in cluster mode.

Additional fix: End tests with a linebreak when running with the `--fastfail` flag. A program that doesn't end its output with a linebreak annoys the prompt in the terminal.


tests/instances.tcl
```diff
@@ -499,8 +499,7 @@ while 1 {
             # letting the tests resume, so we'll eventually reach the cleanup and report crashes
 
             if {$::exit_on_failure} {
-                puts -nonewline "(Fast fail: test will exit now)"
-                flush stdout
+                puts "(Fast fail: test will exit now)"
```
madolson, line 503: Why was this change made?

zuiderkwast: I wrote in the PR top comment: "Additional fix: End tests with a linebreak when running with the --fastfail flag. A program that doesn't end its output with a linebreak annoys the prompt in the terminal."

When it ends without a linebreak in a shell, the prompt gets printed on the same line like this:

```
(Fast fail: test will exit now)viktor@fedora:~/repos/valkey$ 
```

It's independent of this PR, but I was annoyed by it when testing. Does it need a separate PR?

madolson: Oh, I didn't connect those two sentences, thanks!

madolson (approved):

enjoy-binbin (approved):

PR #2177, opened 2025-06-05
fix valkey-cli memory leak

charsyam:
fix memory leak with valkey-cli evalMode

and clusterManagerFixSlotsCoverage

### DIFF
--- a/src/valkey-cli.c
+++ b/src/valkey-cli.c
@@ -5907,8 +5907,7 @@ static int clusterManagerFixSlotsCoverage(char *all_slots) {
                 if (!clusterManagerCheckValkeyReply(n, reply, NULL)) {
                     fixed = -1;
                     if (reply) freeReplyObject(reply);
-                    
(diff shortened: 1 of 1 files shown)


src/valkey-cli.c
```diff
@@ -3545,6 +3545,12 @@ static int evalMode(int argc, char **argv) {
         /* Call it */
         int eval_ldb = config.eval_ldb; /* Save it, may be reverted. */
         retval = issueCommand(argc + 3 - got_comma, argv2);
+
+        for (j = 0; j < argc + 3 - got_comma; j++) {
+            sdsfree(argv[j]);
```
flowerysong:
```suggestion
            sdsfree(argv2[j]);
```

charsyam: @flowerysong You're right. Thanks. applied

hwware (approved):

Nikhil-Manglore (approved):

PR #2179, opened 2025-06-05
Update LPOP and RPOP command json file description

hwware:
When I work on Slot-level metrics and info keysize feature, I read lpop and rpop command source code, the description confused me. I think these 2 commands description on website (https://valkey.io/commands/lpop/ and https://valkey.io/commands/rpop/)  and json files were not updated after count argument was added. 
The return result could be one element only or up to count elements if the optional count argument is provided.

It is related to valkey-doc pr https://github.com/valkey-io/valkey-doc/pull/308


src/commands/lpop.json
```diff
@@ -1,6 +1,6 @@
 {
     "LPOP": {
-        "summary": "Returns the first elements in a list after removing it. Deletes the list if the last element was popped.",
+        "summary": "Returns the first elements in a list after removing it or up to count elements if the optional count argument is provided. Deletes the list if the last element was popped.",
```
zuiderkwast: What do you think about a simpler description like "one or more elements"?

```suggestion
        "summary": "Returns and removes one or more elements from the beginning of a list. Deletes the list if the last element was popped.",
```

hwware: Address comment, Thanks

zuiderkwast (approved):

PR #2180, opened 2025-06-06
Hash prefetching

xbasel:
Prefetches  hash type values. 

This is 1 hash with 30 million fields, 70 bytes each

```
No hash prefetching (hget)
Summary:
  throughput summary: 974468.94 requests per second
  latency summary (msec):
        0.295     0.112     0.295     0.343     0.391    33.151
With hash prefetching:
Summary:
  throughput summary: 1210800.38 requests per second
  latency summary (msec):
        0.240     0.104     0.239     0.311     0.407     7.271
~%20 improvement.
```


ranshid:
This is great!. I wonder if we went all the way though? in some cases we might also want to add another state to check if the hashtable has value pointer and prefetch it as well.

src/memory_prefetch.c
```diff
@@ -132,6 +152,33 @@ static void prefetchValue(KeyPrefetchInfo *info) {
...
+                    if (info->value_prefetch_info.state == HASHTABLE_PREFETCH_INIT) {
+                        hashtableIncrementalFindInit(&info->value_prefetch_info.data.hashtab_state,
+                                                     val->ptr, info->client->argv[2]->ptr
+                        );
+                        info->value_prefetch_info.state = HASHTABLE_PREFETCH_VALUE;
+                        return;
+                    }
+                    if (info->value_prefetch_info.state == HASHTABLE_PREFETCH_VALUE) {
+                        if (!hashtableIncrementalFindStep(&info->value_prefetch_info.data.hashtab_state)) {
+                            info->value_prefetch_info.state = HASHTABLE_PREFETCH_DONE;
```
naglera: The HASHTABLE_PREFETCH_DONE state appears unused and can be removed.

src/memory_prefetch.c
```diff
@@ -132,6 +152,33 @@ static void prefetchValue(KeyPrefetchInfo *info) {
         robj *val = entry;
         if (val->encoding == OBJ_ENCODING_RAW && val->type == OBJ_STRING) {
             valkey_prefetch(val->ptr);
+        } else if (val->encoding == OBJ_ENCODING_HASHTABLE && val->type == OBJ_HASH) {
+            if (info->client) {
```
naglera: Handle the case where info->client is NULL to prevent an infinite loop.

src/memory_prefetch.c
```diff
@@ -132,6 +152,33 @@ static void prefetchValue(KeyPrefetchInfo *info) {
...
+                if (info->client->io_parsed_cmd->proc == hsetCommand || info->client->io_parsed_cmd->proc ==
+                    hgetCommand) {
+                    if (info->value_prefetch_info.state == HASHTABLE_PREFETCH_ENTRY) {
```
naglera: Consider refactor this logic into a dedicated method for prefetching hashtable values, as similar functionality will be needed for set-based commands.

src/memory_prefetch.c
```diff
@@ -148,18 +195,25 @@ static void prefetchValue(KeyPrefetchInfo *info) {
  * prefetch_value - If true, we prefetch the value data for each key.
  * to bring the key's value data closer to the L1 cache as well.
  */
-static void hashtablePrefetch(hashtable **tables) {
+ static void hashtablePrefetch(hashtable **tables) {
```
naglera: nit: run formatter

src/memory_prefetch.c
```diff
@@ -148,18 +195,25 @@ static void prefetchValue(KeyPrefetchInfo *info) {
...
     while ((info = getNextPrefetchInfo())) {
         switch (info->state) {
-        case PREFETCH_ENTRY: prefetchEntry(info); break;
-        case PREFETCH_VALUE: prefetchValue(info); break;
-        default: serverPanic("Unknown prefetch state %d", info->state);
+            case PREFETCH_ENTRY: prefetchEntry(info); break;
+            case PREFETCH_VALUE: {
+                prefetchValue(info);
+                if (info->state != PREFETCH_DONE) {
+                    moveToNextKey();
```
naglera: Consider refactoring to handle index management internally within the prefetch methods rather than requiring explicit moveToNextKey() calls. This would simplify the control flow.

src/memory_prefetch.c
roshkhatri: should not need if we remove `HASHTABLE_PREFETCH_DONE` then it would go to `markKeyAsdone`

src/memory_prefetch.c
```diff
@@ -11,15 +11,33 @@
...
+    HASHTABLE_PREFETCH_INIT, /* Initial state, prefetch entries associated with the given key's hash */
+    HASHTABLE_PREFETCH_VALUE, /* prefetch the value object of the entry found in the previous step */
+    HASHTABLE_PREFETCH_DONE   /* Indicates that prefetching for this key is complete */
+} HashtablePrefetchState;
+
+typedef struct {
+    HashtablePrefetchState state; /* Current state of the prefetch operation */
+    union {
+        hashtableIncrementalFindState hashtab_state;
+        hashtableIncrementalFindState set_state;
```
roshkhatri: We aren't using `set_state` so I think we can remove it

xbasel: I will keep this impl for hash, will remove it.

src/memory_prefetch.c
xbasel: Please check the latest code

src/memory_prefetch.c
```diff
@@ -132,17 +151,48 @@ static void prefetchEntry(KeyPrefetchInfo *info) {
     }
 }
 
+static bool prefetchHashObj(KeyPrefetchInfo *info, robj *val) {
+    if (info->client) {
+        if (info->client->parsed_cmd->proc == hsetCommand || info->client->parsed_cmd->proc ==
+                                                                 hgetCommand) {
```
zuiderkwast: I have some concern for this kind of custom code per command. This is part of the HSET and HGET command, so the implementation/optimization of these commands becomes scattered. The more we add of this kind of code, the bigger this becomes, so if we want to do this for sets, sorted sets, etc., we should think about other ways to do this. For example, can we include add a prefetching callback in the command's JSON file?

PR #2193, opened 2025-06-10
valkey-check-rdb: Fix truncated long aux fields

kukey:
In some case, like in rdb has a long lua, the lua will be incomplete printed.
image

### DIFF
--- a/src/valkey-check-rdb.c
+++ b/src/valkey-check-rdb.c
@@ -135,14 +135,25 @@ void rdbCheckError(const char *fmt, ...) {
 
 /* Print information during RDB checking. */
 void rdbCheckInfo(const char *fmt, ...) {
-    char msg[1024];
+    char msg[1024], *msgbuf = msg;
     va_list ap;
+    int msglen;
 
     va_start(ap, fmt);
-    vsnprintf(msg, sizeof(msg), fmt, ap);
+    msglen = vsnprintf(msg, sizeof(msg), fmt, ap);
     va_end(ap);
 
-    printf("[offset %llu] %s\n", (unsigned long long)(rdbstate.rio ? rdbstate.rio->processed_bytes : 0), msg);
+    if (msglen > (int)(sizeof(msg) - 1)) {
+        msgbuf = sdsnewlen(SDS_NOINIT, sizeof(msg) * 2);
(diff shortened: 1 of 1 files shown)


enjoy-binbin (commented):
In https://github.com/valkey-io/valkey/pull/1448, i mentioned i dont want to print the lua when doing check-rdb. https://github.com/valkey-io/valkey/pull/1448/commits/940753d24782e99d175eddfe482d41225dfbba03

The reason is that on the one hand, i don't think it is real aux auxiliary data and it mees up the output, and on the other hand, after 7.0 we don't store lua script in rdb anymore.

But if we want to keep it, it does need to be fixed.

kukey:
> In #1448, i mentioned i dont want to print the lua when doing check-rdb. [940753d](https://github.com/valkey-io/valkey/commit/940753d24782e99d175eddfe482d41225dfbba03)
> 
> The reason is that on the one hand, i don't think it is real aux auxiliary data and it mees up the output, and on the other hand, after 7.0 we don't store lua script in rdb anymore.
> 
> But if we want to keep it, it does need to be fixed.

@enjoy-binbin thanks binbin to reply; you know and I know, there are many people to use old version redis; Maybe we can add a startup parameter to compatible with this？

zuiderkwast:
> Maybe we can add a startup parameter to compatible with this？

I'm OK with a parameter for this. I was not aware this would be useful, but you seem to be convinced. Earlier, I thought #1448 was not important. :)

See also #1743. It's another special flag to print some extra information in valkey-check-rdb.

zuiderkwast (approved):
The fix LGTM.

If we want to add a parameter to print Lua scripts or not, it's a separate PR.

enjoy-binbin:
> you know and I know, there are many people to use old version redis; 

Yes, i know it for sure. The point is this is a high version OSS, we should move forward.

> Maybe we can add a startup parameter to compatible with this？

i don't want to print it, but in #1448 we do make a agreement that we will keep it, so i dont want to bother to add the new parameter.

For now we are printing it, so the fix is needed. Also i guess we may a long aux field in the future, so that is another reason that we should fix it.

enjoy-binbin (approved):

PR #2195, opened 2025-06-10
Add SAFE option to SHUTDOWN to reject shutdown in unsafe situations

enjoy-binbin:
Add SAFE option to SHUTDOWN. If we passed SAFE, the SHUTDOWN
will refuse to shutdown if it is not safe to shutdown. Like if
myself is a voting primary, it will refuse to shutdown. This
avoids the situation where a replica suddenly becomes the primary
when we shutting down the replica, or shutting down a primary node
by mistake and then causing the cluster to down.

Add SAFE option to SHUTDOWN command.
Add safe option to shutdown-on-sigint and shutdown-on-sigterm.

Note that SAFE cannot prevent FORCE, in the case of FORCE, SAFE
will print the relevant logs and do the FORCE shutdown, we allow
this combination.

Doc PR: https://github.com/valkey-io/valkey-doc/pull/325

### DIFF
--- a/src/db.c
+++ b/src/db.c
@@ -1314,7 +1314,7 @@ void typeCommand(client *c) {
     addReplyStatus(c, getObjectTypeName(o));
 }
 
-/* SHUTDOWN [[NOSAVE | SAVE] [NOW] [FORCE] | ABORT] */
+/* SHUTDOWN [[NOSAVE | SAVE] [NOW] [FORCE] [SAFE] | ABORT] */
 void shutdownCommand(client *c) {
     int flags = SHUTDOWN_NOFLAGS;
     int abort = 0;
@@ -1329,6 +1329,8 @@ void shutdownCommand(client *c) {
             flags |= SHUTDOWN_FORCE;
         } else if (!strcasecmp(c->argv[i]->ptr, "abort")) {
             abort = 1;
+        } else if (!strcasecmp(c->argv[i]->ptr, "safe")) {
+            flags |= SHUTDOWN_SAFE;
         } else {
             addReplyErrorObject(c, shared.syntaxerr);
             return;
--- a/src/server.c
+++ b/src/server.c
@@ -4553,6 +4553,7 @@ int finishShutdown(void) {
     int save = server.shutdown_flags & SHUTDOWN_SAVE;
     int nosave = server.shutdown_flags & SHUTDOWN_NOSAVE;
     int force = server.shutdown_flags & SHUTDOWN_FORCE;
+    int safe = server.shutdown_flags & SHUTDOWN_SAFE;
 
     /* Log a warning for each replica that is lagging. */
     listIter replicas_iter;
@@ -4573,6 +4574,17 @@ int finishShutdown(void) {
     if (num_replicas > 0) {
         serverLog(LL_NOTICE, "%d of %d replicas are in sync when shutting down.", num_replicas - num_lagging_replicas,
                   num_replicas);
+    }
+
+    if (safe && server.cluster_enabled && clusterNodeIsVotingPrimary(getMyClusterNode())) {
+        if (force) {
+            serverLog(LL_WARNING, "I am a voting primary, shutting down may cause the cluster to down. Exit anyway.");
+        } else {
+            serverLog(LL_WARNING, "I am a voting primary, shutting down may cause the cluster to down, can't exit.");
+            if (server.supervised_mode == SUPERVISED_SYSTEMD)
+                serverCommunicateSystemd("I am a voting primary, shutting down may cause the cluster to down, can't exit.\n");
+            goto error;
+        }
     }
 
     /* Kill all the Lua debugger forked sessions. */
--- a/src/server.c
+++ b/src/server.c
@@ -4578,11 +4578,11 @@ int finishShutdown(void) {
 
     if (safe && server.cluster_enabled && clusterNodeIsVotingPrimary(getMyClusterNode())) {
         if (force) {
-            serverLog(LL_WARNING, "I am a voting primary, shutting down may cause the cluster to down. Exit anyway.");
+            serverLog(LL_WARNING, "This is a voting primary. Shutting down may cause the cluster to go down. Exit anyway.");
         } else {
-            serverLog(LL_WARNING, "I am a voting primary, shutting down may cause the cluster to down, can't exit.");
+            serverLog(LL_WARNING, "This is a voting primary. Shutting down may cause the cluster to go down. Can't exit.");
             if (server.supervised_mode == SUPERVISED_SYSTEMD)
-                serverCommunicateSystemd("I am a voting primary, shutting down may cause the cluster to down, can't exit.\n");
+                serverCommunicateSystemd("This is a voting primary. Shutting down may cause the cluster to go down. Can't exit.\n");
             goto error;
         }
     }
--- a/src/server.c
+++ b/src/server.c
@@ -4456,6 +4456,12 @@ void closeListeningSockets(int unlink_unix_socket) {
  * - SHUTDOWN_FORCE: Ignore errors writing AOF and RDB files on disk, which
  *   would normally prevent a shutdown.
  *
+ * - SHUTDOWN_SAFE: Shut down only when safe. Note that safe cannot prevent force,
+ *   in the case of force, safe will print the relevant logs. The definition of
+ *   safe may be different in different modes. Here are the definitions:
+ *   * In cluster mode, it is unsafe to shut down a primary with slots, and may
+ *     cause the cluster to go down.
+ *
  * Unless SHUTDOWN_NOW is set and if any replicas are lagging behind, C_ERR is
  * returned and server.shutdown_mstime is set to a timestamp to allow a grace
  * period for the replicas to catch up. This is checked and handled by
--- a/tests/unit/cluster/shutdown.tcl
+++ b/tests/unit/cluster/shutdown.tcl
@@ -1,0 +1,15 @@
+start_cluster 3 0 {tags {external:skip cluster shutdown}} {
+    test "Test shutdown safe is work" {
+        assert_error {ERR Errors trying to SHUTDOWN*} {R 0 shutdown safe}
+        assert_error {ERR Errors trying to SHUTDOWN*} {R 1 shutdown safe}
+        assert_error {ERR Errors trying to SHUTDOWN*} {R 2 shutdown safe}
+
+        catch {R 0 shutdown safe force}
+        wait_for_condition 1000 50 {
+            [CI 1 cluster_state] eq {fail} &&
+            [CI 2 cluster_state] eq {fail}
+        } else {
+            fail "Cluster doesn't fail"
+        }
+    }
+}
--- a/tests/unit/cluster/shutdown.tcl
+++ b/tests/unit/cluster/shutdown.tcl
@@ -1,5 +1,5 @@
 start_cluster 3 0 {tags {external:skip cluster shutdown}} {
-    test "Test shutdown safe is work" {
+    test "Test shutdown safe is working" {
         assert_error {ERR Errors trying to SHUTDOWN*} {R 0 shutdown safe}
         assert_error {ERR Errors trying to SHUTDOWN*} {R 1 shutdown safe}
         assert_error {ERR Errors trying to SHUTDOWN*} {R 2 shutdown safe}
--- a/valkey.conf
+++ b/valkey.conf
@@ -1705,6 +1705,11 @@ aof-timestamp-enabled no
 # nosave:   Prevents DB saving operation even if one or more save points are configured.
 # now:      Skips waiting for lagging replicas.
 # force:    Ignores any errors that would normally prevent the server from exiting.
+# safe:     Shutdown only when safe. Note that safe cannot prevent force, in the case of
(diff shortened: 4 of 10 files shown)


madolson:
Didn't really read the code, but directionally I'm OK with adding a safe option to shutdown.

madolson (approved):
I've now actually read the code and am OK with it.

src/server.c
```diff
@@ -4575,6 +4576,17 @@ int finishShutdown(void) {
                   num_replicas);
     }
 
+    if (safe && server.cluster_enabled && clusterNodeIsVotingPrimary(getMyClusterNode())) {
+        if (force) {
+            serverLog(LL_WARNING, "I am a voting primary, shutting down may cause the cluster to down. Exit anyway.");
```
jsoref:
```suggestion
            serverLog(LL_WARNING, "I am a voting primary, shutting down may cause the cluster to go down. Exit anyway.");
```

src/server.c
```diff
@@ -4575,6 +4576,17 @@ int finishShutdown(void) {
...
+        } else {
+            serverLog(LL_WARNING, "I am a voting primary, shutting down may cause the cluster to down, can't exit.");
```
jsoref:
```suggestion
            serverLog(LL_WARNING, "I am a voting primary, shutting down may cause the cluster to go down, can't exit.");
```

src/server.c
```diff
@@ -4575,6 +4576,17 @@ int finishShutdown(void) {
...
+            if (server.supervised_mode == SUPERVISED_SYSTEMD)
+                serverCommunicateSystemd("I am a voting primary, shutting down may cause the cluster to down, can't exit.\n");
```
jsoref:
```suggestion
                serverCommunicateSystemd("I am a voting primary, shutting down may cause the cluster to go down, can't exit.\n");
```

src/server.c
enjoy-binbin: @zuiderkwast Do you have any suggestions for texts? I usually feel like my English texts are very Chinese in sometimes.

src/db.c
```diff
@@ -1314,7 +1314,7 @@ void typeCommand(client *c) {
     addReplyStatus(c, getObjectTypeName(o));
 }
 
-/* SHUTDOWN [[NOSAVE | SAVE] [NOW] [FORCE] | ABORT] */
+/* SHUTDOWN [[NOSAVE | SAVE] [NOW] [FORCE] [SAFE] | ABORT] */
```
zuiderkwast, line 1317: Can we forbid the comination FORCE and SAFE?

```suggestion
/* SHUTDOWN [[NOSAVE | SAVE] [NOW] [FORCE | SAFE] | ABORT] */
```

Also in json file and in the check for illegal combo below.

enjoy-binbin: I somehow want to keep it, the safe thing is just like another error that we can think? I am ok with both anyway, forbid also seem reasonable
```c
            if (force) {
                serverLog(LL_WARNING, "Error trying to save the DB. Exit anyway.");
            } else {
                serverLog(LL_WARNING, "Error trying to save the DB, can't exit.");
```

zuiderkwast: I'm just thinking about the case someone uses FORCE and SAFE together by mistake. The server exits even if it's not safe?

What are the possible scenarios?

|           | not SAFE | SAFE |
|-----------|----------|------|
| not FORCE | ?        | ?    |
| FORCE     | ?        | ?    |

enjoy-binbin: I mentioned it in the conf
```
# safe:     Shutdown only when safe. Note that safe cannot prevent force, in the case of
#           force, safe will print the relevant logs. The definition of safe may be different
#           in different modes. Here are the definitions:
#           1. In cluster mode, it is unsafe to shutdown a primary with slots, and may cause
#              the cluster to go down.
```

|           | not SAFE | SAFE |
|-----------|----------|------|
| not FORCE | pass        | dont exit if not safe, exit only safe    |
| FORCE     | exit anyway        | exit anyway but print a log if unsafe    |

zuiderkwast: The "exit anyway but print a log if unsafe" seems risky. Maybe we can do "don't exit" in this case.. If they want to exit, they need FORCE without SAFE. Seems safer?

zuiderkwast: I guess FORCE + SAFE can still be useful. It can shut down a replica when save dump to disk failed.

enjoy-binbin: > I guess FORCE + SAFE can still be useful. It can shut down a replica when save dump to disk failed.

FORCE can anyway ignore a dump failure and exit right now (without SAFE)

enjoy-binbin: Do you want to forbid it? I think FORCE can anyway shutdown the node, but in some cases we may print some useful logs. I am ok with both, i actually forbid it the first time, but when i update isValidShutdownOnSigFlags, i thought again, why not allow them to be combined? haha

zuiderkwast: I'm OK with your idea.

src/server.c
jsoref: Fwiw, the `I am a voting primary` is kinda odd.

Normally programs don't use first person. I haven't checked valkey's style to see if it likes to personify itself. Typically I'd expect:
```suggestion
            serverLog(LL_WARNING, "This is a voting primary, shutting down may cause the cluster to go down. Exit anyway.");
```

But that's a bigger change...

tests/unit/cluster/shutdown.tcl
```diff
@@ -0,0 +1,15 @@
+start_cluster 3 0 {tags {external:skip cluster shutdown}} {
+    test "Test shutdown safe is work" {
```
jsoref:
```suggestion
    test "Test safe shutdown works" {
```

or `is working` (but not `is work`) -- I can't write a [check-spelling rule](https://github.com/check-spelling/spell-check-this/blob/prerelease/.github/actions/spelling/line_forbidden.patterns) against `is work` because it is legal to say "making breakfast is work" is correct -- it just means something different -- that the thing is in a category "work" as opposed to the action is successful.

src/server.c
zuiderkwast: @jsoref's suggestion is good. You can take it.

tests/unit/cluster/shutdown.tcl
```diff
@@ -0,0 +1,15 @@
...
+        assert_error {ERR Errors trying to SHUTDOWN*} {R 0 shutdown safe}
+        assert_error {ERR Errors trying to SHUTDOWN*} {R 1 shutdown safe}
+        assert_error {ERR Errors trying to SHUTDOWN*} {R 2 shutdown safe}
```
jsoref, line 5: I can't tell what these are doing, but if `SHUTDOWN` isn't some sort of command that a user gives as a verb (`shutdown` itself isn't a verb in English, see below) and `Errors trying to SHUTDOWN` is just going to be shown to a user, then it'd be more correct to show: `Errors trying to SHUT DOWN` because `to` is ["—used as a function word to indicate that the following verb is an infinitive" (1 of 3 "preposition"; 8; first case)](https://www.merriam-webster.com/dictionary/to) which means that the thing that follows needs to be an infinitive.

https://www.merriam-webster.com/dictionary/shutdown

> `shutdown` 1 of 2 [noun](https://www.merriam-webster.com/dictionary/noun)
> `shut down` 2 of 2 [verb](https://www.merriam-webster.com/dictionary/verb)

valkey.conf
```diff
@@ -1705,6 +1705,11 @@ aof-timestamp-enabled no
 # nosave:   Prevents DB saving operation even if one or more save points are configured.
 # now:      Skips waiting for lagging replicas.
 # force:    Ignores any errors that would normally prevent the server from exiting.
+# safe:     Shutdown only when safe. Note that safe cannot prevent force, in the case of
```
jsoref:
```suggestion
# safe:     Shut down only when safe. Note that safe cannot prevent force, in the case of
```

Again, this should be a verb as sentences need verbs and there aren't any others in that sentence...

valkey.conf
```diff
@@ -1705,6 +1705,11 @@ aof-timestamp-enabled no
...
+#           force, safe will print the relevant logs. The definition of safe may be different
+#           in different modes. Here are the definitions:
+#           1. In cluster mode, it is unsafe to shutdown a primary with slots, and may cause
+#              the cluster to go down.
```
jsoref:
```suggestion
#           * In cluster mode, it is unsafe to shutdown a primary with slots, and may cause
#             the cluster to go down.
```

Unless you're planning on referencing the mode numbers, which seems unlikely and error prone. As they're more of a bag of definitions as opposed to some sequential or otherwise ordered list of definitions, it makes more sense to use an unordered bulleted list then a numbered ordered list.

jsoref (commented):
This is a slightly more detailed reading of the text with an English lense.

tests/unit/cluster/shutdown.tcl
zuiderkwast: This test case is pattern matching on an error message. The full message is "Errors trying to SHUTDOWN. Check logs." Quite strange, I agree, but it's an existing error message and changing it can maybe affect other things, at least other test cases, but in the worst case also some software that relies on the exact error message. So I think it's best not to change it.

SHUTDOWN is a command, but here it's used as a verb, like in a double meaning.

src/server.c
zuiderkwast: Agree, "I am" is slightly funny style. "This is" can be better.

I searched the source code and found two error replies with "I'm":

* "I'm already the owner of hash slot %u" (where %u is a number)
* "I'm a replica but my master is unknown to me"

Btw, shouldn't the first comma be a period? Command seems like bad grammar to me.

```suggestion
            serverLog(LL_WARNING, "This is a voting primary. Shutting down may cause the cluster to down. Exit anyway.");
```

src/server.c
zuiderkwast: * I am → This is
* to down → to go down
* , → .


```suggestion
            serverLog(LL_WARNING, "This is a voting primary. Shutting down may cause the cluster to go down. Can't exit.");
```

src/server.c
zuiderkwast: @jsoref's suggestion + more

```suggestion
                serverCommunicateSystemd("This is a voting primary. Shutting down may cause the cluster to go down. Can't exit.\n");
```

tests/unit/cluster/shutdown.tcl
jsoref: I'd claim the error message is wrong, it should either be `Error trying to perform SHUTDOWN` or `Error trying to SHUT DOWN`.

Note that unless the program *knows* it has more than one `Error` it really shouldn't be saying `Errors`. And generally, if it's going to say `errors` it should probably give a count (`Failed to shut down because of 5 errors`).

zuiderkwast: I repeat:

> it's an existing error message and changing it can maybe affect other things, at least other test cases, but in the worst case also some software that relies on the exact error message

This overrides any grammar correctness reasoning.

We have changed error messages in the past and it has broken users' software. Backward compatibility is very important. Grammar is less important.

enjoy-binbin:
thanks for the review and suggestions, sorry about that, i have always had this problem, chinese english as a Chinese, i am very bad at grammar.

valkey.conf
enjoy-binbin: good point, thanks

tests/unit/cluster/shutdown.tcl
enjoy-binbin: ok, i see, thanks for the explanation

valkey.conf
```diff
@@ -1705,6 +1705,11 @@ aof-timestamp-enabled no
 # nosave:   Prevents DB saving operation even if one or more save points are configured.
 # now:      Skips waiting for lagging replicas.
 # force:    Ignores any errors that would normally prevent the server from exiting.
+# safe:     Shut down only when safe. Note that safe cannot prevent force, in the case of
+#           force, safe will print the relevant logs. The definition of safe may be different
+#           in different modes. Here are the definitions:
+#           * In cluster mode, it is unsafe to shutdown a primary with slots, and may cause
```
zuiderkwast:
```suggestion
#           * In cluster mode, it is unsafe to shut down a primary with slots, and may cause
```

tests/unit/cluster/shutdown.tcl
enjoy-binbin: I usually don't care much about these wording or grammar (maybe because I'm not good at it), and I believe we still have a lot of such mistakes, they are not fatal, and everyone may write them.

This will always happen anyway unless we have a strict robotic CI that tests our code every time it is merged. And we do have a spellcheck CI, but i guess we don't have a perfect tool that can catch all these?

jsoref: my check-spelling tool could help w/ quite a few of these, but not all (and for a bunch of them it tries to explain what the fix is which is hopefully a way for people to learn to avoid making it in the future).

I'll try to get improvements from my spelling PR merged and then the project can talk about whether they're interested in a stricter tool such as mine.

Both fully native English projects and international projects will have problems with English (native speakers make quite a few mistakes, although they tend to be different from the mistakes non native speakers make).

enjoy-binbin:
@valkey-io/core-team Do any of you want to review (or ack) it before the final merge?

zuiderkwast (approved):
ACK

PR #2178, opened 2025-06-05
Fix replica can't finish failover when config epoch is outdated

enjoy-binbin:
When the primary changes the config epoch and then down immediately,
the replica may not update the config epoch in time. Although we will
broadcast the change in cluster (see #1813), there may be a race in
the network or in the code. In this case, the replica will never finish
the failover since other primaries will refuse to vote because the
replica's slot config epoch is old.

We need a way to allow the replica can finish the failover in this case.

When the primary refuses to vote because the replica's config epoch is
less than the dead primary's config epoch, it can send an UPDATE packet
to the replica to inform the replica about the dead primary. The UPDATE
message contains information about the dead primary's config epoch and
owned slots. The failover will time out, but later the replica can try
again with the updated config epoch and it can succeed.

Fixes #2169.

PR #2196, opened 2025-06-11
Add packet-drop to fix the new flaky failover test

enjoy-binbin:
The new test was added in #2178, obviously there may be
pending reads in the connection, so there may be a race
in the DROP-CLUSTER-PACKET-FILTER part causing the test
to fail. Add CLOSE-CLUSTER-LINK-ON-PACKET-DROP to ensure
that the replica does not process the packet.

### DIFF
--- a/tests/unit/cluster/failover2.tcl
+++ b/tests/unit/cluster/failover2.tcl
@@ -147,6 +147,7 @@ proc test_replica_config_epoch_failover {type} {
             R 3 CONFIG SET cluster-replica-no-failover yes
         }
         R 3 DEBUG DROP-CLUSTER-PACKET-FILTER $CLUSTER_PACKET_TYPE_ALL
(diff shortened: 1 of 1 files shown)


enjoy-binbin:
https://github.com/valkey-io/valkey/actions/runs/15576081823
valgrind ci

enjoy-binbin:
I see both tests are passed, so i think we are good to go.
```
[ok]: Replica can update the config epoch when trigger the failover - automatic (16813 ms)
=== (external:skip cluster) Starting server on 127.0.0.1:21689 ok
=== (external:skip cluster) Starting server on 127.0.0.1:21690 ok
=== (external:skip cluster) Starting server on 127.0.0.1:21691 ok
=== (external:skip cluster) Starting server on 127.0.0.1:21692 ok
[ok]: Replica can update the config epoch when trigger the failover - manual (8716 ms)
```

zuiderkwast (approved):

PR #2199, opened 2025-06-11
Stabilize dual channel test: Adjust RDB loading progress calls

stav-bentov:
Modify test to increase frequency of `rdbLoadProgressCallback` calls by reducing `loading-process-events-interval-bytes` to 1024. 
This adjustment is needed because `rdb-key-save-delay` slows down bulk transfers, which may cause the replica to miss main connection messages. The change ensures that:
1. Messages on main channel are not missed.
2. Test properly handles graceful shutdown after user interrupt.

### DIFF
--- a/tests/integration/dual-channel-replication.tcl
+++ b/tests/integration/dual-channel-replication.tcl
@@ -1176,6 +1176,7 @@ start_server {tags {"dual-channel-replication external:skip"}} {
         $replica config set dual-channel-replication-enabled yes
         $replica config set loglevel debug
         $replica config set repl-timeout 10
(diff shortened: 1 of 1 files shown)


tests/integration/dual-channel-replication.tcl
```diff
@@ -1217,6 +1218,7 @@ start_server {tags {"dual-channel-replication external:skip"}} {
         } else {
             fail "Primary should abort sync"
         }
+        $replica config set loading-process-events-interval-bytes 1024;
```
naglera: You can remove this line. We don't need to set this config twice
```suggestion
```

PR #2200, opened 2025-06-11
Stabilize dual channel test: Split scenarios and set RDB loading interval

stav-bentov:
1. Split the dual-channel test into two independent test scopes:
 - One for killing the RDB connection.
 - One for killing the main connection.
2. Increase frequency of `rdbLoadProgressCallback` calls by reducing `loading-process-events-interval-bytes` to 1024. 
This adjustment is needed because `rdb-key-save-delay` slows down bulk transfers, which may cause the replica to miss main connection messages. The change ensures that:
 - Messages on main channel are not missed.
 - Test properly handles graceful shutdown after user interrupt.


ranshid:
Please fix DCO

naglera (approved):

naglera (approved):

ranshid (approved):

PR #2202, opened 2025-06-11
Remove `get_type` from `ConnectionType`

### LINKED ISSUE #2187: Remove `get_type` from `ConnctionType` after merging #2070
I don't think `get_type` is needed anymore. we only need `get_type` today for logging so a simple helper function that maps a given type id to its string representation would suffice.

_Originally posted by @PingXie in https://github.com/valkey-io/valkey/pull/2070#discussion_r2109691808_

cxljs:
close #2187

### DIFF
--- a/src/connection.c
+++ b/src/connection.c
@@ -31,15 +31,7 @@ static ConnectionType *connTypes[CONN_TYPE_MAX];
 
 int connTypeRegister(ConnectionType *ct) {
     int type_id = ct->get_type_id(NULL);
-    if (type_id <= 0 || type_id >= CONN_TYPE_MAX) {
-        serverLog(LL_WARNING, "Invalid connection type id %d", type_id);
-        return C_ERR;
-    }
-
-    if (connTypes[type_id]) {
-        serverLog(LL_WARNING, "Connection type %s already registered", getConnectionTypeName(type_id));
-        return C_ERR;
-    }
+    serverAssert(type_id > 0 && type_id < CONN_TYPE_MAX && !connTypes[type_id]);
 
     serverLog(LL_VERBOSE, "Connection type %s registering", getConnectionTypeName(type_id));
     connTypes[type_id] = ct;
@@ -68,10 +60,7 @@ int connTypeInitialize(void) {
 }
 
 ConnectionType *connectionByType(int type_id) {
-    if (type_id <= 0 || type_id >= CONN_TYPE_MAX) {
-        serverLog(LL_WARNING, "Invalid connection type id %d", type_id);
-        return NULL;
-    }
+    serverAssert(type_id > 0 && type_id < CONN_TYPE_MAX);
 
     ConnectionType *ct = connTypes[type_id];
 
@@ -117,11 +106,6 @@ ConnectionType *connectionTypeUnix(void) {
 
     ct_unix = connectionByType(CONN_TYPE_ID_UNIX);
     return ct_unix;
-}
-
-int connectionIndexByType(int type_id) {
-    if (!connTypes[type_id]) return -1;
-    return type_id;
 }
 
 void connTypeCleanupAll(void) {
--- a/src/connection.c
+++ b/src/connection.c
@@ -30,11 +30,11 @@
 static ConnectionType *connTypes[CONN_TYPE_MAX];
 
 int connTypeRegister(ConnectionType *ct) {
-    int type_id = ct->get_type_id(NULL);
-    serverAssert(type_id > 0 && type_id < CONN_TYPE_MAX && !connTypes[type_id]);
+    int type = ct->get_type();
+    serverAssert(type > 0 && type < CONN_TYPE_MAX && !connTypes[type]);
 
-    serverLog(LL_VERBOSE, "Connection type %s registering", getConnectionTypeName(type_id));
-    connTypes[type_id] = ct;
+    serverLog(LL_VERBOSE, "Connection type %s registering", getConnectionTypeName(type));
+    connTypes[type] = ct;
 
     if (ct->init) {
         ct->init();
@@ -59,13 +59,13 @@ int connTypeInitialize(void) {
     return C_OK;
 }
 
-ConnectionType *connectionByType(int type_id) {
-    serverAssert(type_id > 0 && type_id < CONN_TYPE_MAX);
+ConnectionType *connectionByType(int type) {
+    serverAssert(type > 0 && type < CONN_TYPE_MAX);
 
-    ConnectionType *ct = connTypes[type_id];
+    ConnectionType *ct = connTypes[type];
 
     if (!ct) {
-        serverLog(LL_WARNING, "Missing implement of connection type %s", getConnectionTypeName(type_id));
+        serverLog(LL_WARNING, "Missing implement of connection type %s", getConnectionTypeName(type));
         return NULL;
     }
     return ct;
@@ -77,7 +77,7 @@ ConnectionType *connectionTypeTcp(void) {
 
     if (ct_tcp != NULL) return ct_tcp;
 
-    ct_tcp = connectionByType(CONN_TYPE_ID_SOCKET);
+    ct_tcp = connectionByType(CONN_TYPE_SOCKET);
     serverAssert(ct_tcp != NULL);
 
     return ct_tcp;
@@ -92,7 +92,7 @@ ConnectionType *connectionTypeTls(void) {
      * So we need the cached pointer to handle NULL correctly too. */
     if (!cached) {
         cached = 1;
-        ct_tls = connectionByType(CONN_TYPE_ID_TLS);
+        ct_tls = connectionByType(CONN_TYPE_TLS);
     }
 
     return ct_tls;
@@ -104,7 +104,7 @@ ConnectionType *connectionTypeUnix(void) {
 
     if (ct_unix != NULL) return ct_unix;
 
-    ct_unix = connectionByType(CONN_TYPE_ID_UNIX);
+    ct_unix = connectionByType(CONN_TYPE_UNIX);
     return ct_unix;
 }
 
@@ -157,7 +157,7 @@ sds getListensInfoString(sds info) {
         connListener *listener = &server.listeners[j];
         if (listener->ct == NULL) continue;
 
-        info = sdscatfmt(info, "listener%i:name=%s", j, getConnectionTypeName(listener->ct->get_type_id(NULL)));
+        info = sdscatfmt(info, "listener%i:name=%s", j, getConnectionTypeName(listener->ct->get_type()));
         for (int i = 0; i < listener->count; i++) {
             info = sdscatfmt(info, ",bind=%s", listener->bindaddr[i]);
         }
--- a/src/connection.c
+++ b/src/connection.c
@@ -31,7 +31,7 @@ static ConnectionType *connTypes[CONN_TYPE_MAX];
 
 int connTypeRegister(ConnectionType *ct) {
     int type = ct->get_type();
-    serverAssert(type > 0 && type < CONN_TYPE_MAX && !connTypes[type]);
+    serverAssert(type >= 0 && type < CONN_TYPE_MAX && !connTypes[type]);
 
     serverLog(LL_VERBOSE, "Connection type %s registering", getConnectionTypeName(type));
     connTypes[type] = ct;
@@ -60,13 +60,12 @@ int connTypeInitialize(void) {
 }
 
 ConnectionType *connectionByType(int type) {
-    serverAssert(type > 0 && type < CONN_TYPE_MAX);
+    serverAssert(type >= 0 && type < CONN_TYPE_MAX);
 
     ConnectionType *ct = connTypes[type];
 
     if (!ct) {
         serverLog(LL_WARNING, "Missing implement of connection type %s", getConnectionTypeName(type));
-        return NULL;
     }
     return ct;
 }
@@ -112,8 +111,7 @@ void connTypeCleanupAll(void) {
     ConnectionType *ct;
     int type;
 
-    // type = 0 is an invalid conn type.
-    for (type = 1; type < CONN_TYPE_MAX; type++) {
+    for (type = 0; type < CONN_TYPE_MAX; type++) {
         ct = connTypes[type];
         if (!ct) continue;
 
@@ -127,8 +125,7 @@ int connTypeHasPendingData(void) {
     int type;
     int ret = 0;
 
-    // type = 0 is an invalid conn type.
-    for (type = 1; type < CONN_TYPE_MAX; type++) {
+    for (type = 0; type < CONN_TYPE_MAX; type++) {
         ct = connTypes[type];
         if (ct && ct->has_pending_data && (ret = ct->has_pending_data())) {
             return ret;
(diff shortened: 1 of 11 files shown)


src/connection.h
```diff
@@ -78,6 +78,21 @@ typedef enum {
...
+    case CONN_TYPE_ID_SOCKET:
+        return CONN_TYPE_SOCKET;
+    case CONN_TYPE_ID_UNIX:
+        return CONN_TYPE_UNIX;
+    case CONN_TYPE_ID_TLS:
+        return CONN_TYPE_TLS;
+    case CONN_TYPE_ID_RDMA:
+        return CONN_TYPE_RDMA;
+    default:
+        return "invalid type";
```
zuiderkwast, line 87: Can't modules add connection types?

cxljs: Hello, can valkey now add types other than TCP, Unix, TLS, and RDMA? If so, there are some areas I haven't considered, and I will correct them.

madolson: There is no module API for adding a new type of connection. In the past we had TLS as a module, which allowed someone to dynamically load TLS support, but the code had to already be present within the core. I think the same is basically true here. Modules can't add new fundamental types of connections, but they can provide alternative implementations. So this structure should be OK.

zuiderkwast: Right. I believe it's still possible to build TLS and RDMA as a module but I guess no module can add new types. So this is OK.

fukua95: Got it, thank you!

cxljs: > There is no module API for adding a new type of connection. In the past we had TLS as a module, which allowed someone to dynamically load TLS support, but the code had to already be present within the core. I think the same is basically true here. Modules can't add new fundamental types of connections, but they can provide alternative implementations. So this structure should be OK.

Got it, thanks!

cxljs: > Right. I believe it's still possible to build TLS and RDMA as a module but I guess no module can add new types. So this is OK.

Got it, thanks!

src/connection.c
```diff
@@ -39,14 +39,13 @@ int connTypeRegister(ConnectionType *ct) {
         tmpct = connTypes[type];
         if (!tmpct) break;
 
-        /* ignore case, we really don't care "tls"/"TLS" */
-        if (!strcasecmp(typename, tmpct->get_type(NULL))) {
-            serverLog(LL_WARNING, "Connection types %s already registered", typename);
+        if (tmpct->get_type_id(NULL) == type_id) {
```
artikell: The `type_id` is already unique and can be used as an index for querying. There's no need for a loop search.

artikell:
From a design perspective, each `connectionType` is independent of the others. Implementing a `getConnectionTypeName` method would reintroduce a dependency on centralized logic. 
I believe there is no significant issue with retaining the `get_type` method, as it maintains independence and allows for compatible changes. It would be reasonable to replace the unique identifier with a `type_id`.

src/connection.c
cxljs: Thank you.

src/connection.h
PingXie: >Right. I believe it's still possible to build TLS and RDMA as a module

That is a misnomer. TLS/RDMA are not Valkey **modules** but just shared libraries. I didn't mind it in the past but now I think it is important to officially clarify that in our docs. will find time to review the related docs.

PingXie: let's simplify the code a bit further (and remove the \#defines)
```suggestion
static inline const char *getConnectionTypeName(int type_id) {
    switch (type_id) {
    case CONN_TYPE_ID_SOCKET:
        return "tcp";
    case CONN_TYPE_ID_UNIX:
        return "unix";
    case CONN_TYPE_ID_TLS:
        return "tls";
    case CONN_TYPE_ID_RDMA:
        return "rdma";
    default:
        return "invalid type";
```

src/connection.c
```diff
@@ -143,7 +130,7 @@ void connTypeCleanupAll(void) {
 
     for (type = 0; type < CONN_TYPE_MAX; type++) {
         ct = connTypes[type];
-        if (!ct) break;
+        if (!ct) continue;
```
PingXie, line 116: good catch

src/connection.c
```diff
@@ -30,24 +30,19 @@
...
-
-    /* find an empty slot to store the new connection type */
-    for (type = 0; type < CONN_TYPE_MAX; type++) {
-        tmpct = connTypes[type];
-        if (!tmpct) break;
+    int type_id = ct->get_type_id(NULL);
+    if (type_id <= 0 || type_id >= CONN_TYPE_MAX) {
+        serverLog(LL_WARNING, "Invalid connection type id %d", type_id);
+        return C_ERR;
+    }
```
PingXie: a serverAssert seems better to me.

```suggestion
    serverAssert(type_id > 0 && type_id < CONN_TYPE_MAX);
```

src/connection.c
```diff
@@ -30,24 +30,19 @@
...
 
-        /* ignore case, we really don't care "tls"/"TLS" */
-        if (!strcasecmp(typename, tmpct->get_type(NULL))) {
-            serverLog(LL_WARNING, "Connection types %s already registered", typename);
-            return C_ERR;
-        }
+    if (connTypes[type_id]) {
+        serverLog(LL_WARNING, "Connection type %s already registered", getConnectionTypeName(type_id));
+        return C_ERR;
     }
```
PingXie: similarly, this sounds like a logic bug to me

```suggestion
    serverAssert(connTypes[type_id] == NULL);
```

src/connection.c
```diff
@@ -120,21 +115,13 @@ ConnectionType *connectionTypeUnix(void) {
...
-        if (!ct) break;
-
-        if (!strcasecmp(typename, ct->get_type(NULL))) return type;
-    }
-
-    return -1;
+int connectionIndexByType(int type_id) {
+    if (!connTypes[type_id]) return -1;
+    return type_id;
 }
```
PingXie: remove this function. `type_id` is `index`. this indirection is vestigial and confusing.

PingXie (commented):
I like this change. overall LGTM. thanks @cxljs!

cxljs:
@PingXie  Thank you very much for reviewing the PR. I have already made the changes.

src/rdma.c
```diff
@@ -17,6 +17,8 @@
 #include "server.h"
 #include "connection.h"
 
+#define CONN_TYPE_RDMA "rdma"
+
```
PingXie:
```suggestion
```

src/rdma.c
```diff
@@ -1821,7 +1815,6 @@ static void updateRdmaState(struct connection *conn) {
 static ConnectionType CT_RDMA = {
     /* connection type */
     .get_type_id = connRdmaGetTypeId,
```
PingXie: couple of comments:

1. I don't see anywhere a `get_type_id` actually needing a `conn` parameter. let's remove the `conn` parameter from the interface.
2. now that we have removed `get_type` I think we should just rename `get_type_id` to `get_type`
3. along the same line, we should drop the `ID` part from the type like `CONN_TYPE_ID_RDMA`, i.e., renaming it to `CONN_TYPE_RDMA`
4. there are still unused old string conn types in some c files. can you clean them up?

PingXie (commented):
thanks @cxljs. a few more minor things.

src/rdma.c
cxljs: @PingXie Thank you.

src/connection.h
```diff
@@ -64,25 +65,35 @@ typedef enum {
...
-    CONN_TYPE_ID_INVALID = 0,
-    CONN_TYPE_ID_SOCKET,
-    CONN_TYPE_ID_UNIX,
-    CONN_TYPE_ID_TLS,
-    CONN_TYPE_ID_RDMA,
+    CONN_TYPE_INVALID = 0,
+    CONN_TYPE_SOCKET,
+    CONN_TYPE_UNIX,
+    CONN_TYPE_TLS,
+    CONN_TYPE_RDMA,
```
PingXie, line 72:
```suggestion
    CONN_TYPE_RDMA,
    CONN_TYPE_MAX
```

src/connection.h
```diff
@@ -64,25 +65,35 @@ typedef enum {
...
 } ConnectionTypeId;
 
-#define CONN_TYPE_SOCKET "tcp"
-#define CONN_TYPE_UNIX "unix"
-#define CONN_TYPE_TLS "tls"
-#define CONN_TYPE_RDMA "rdma"
-#define CONN_TYPE_MAX 8 /* 8 is enough to be extendable */
+#define CONN_TYPE_MAX 5
```
PingXie:
```suggestion

```

src/connection.h
```diff
@@ -64,25 +65,35 @@ typedef enum {
 #define CONN_FLAG_ALLOW_ACCEPT_OFFLOAD (1 << 2) /* Connection accept can be offloaded to IO threads. */
 
 typedef enum {
-    CONN_TYPE_ID_INVALID = 0,
-    CONN_TYPE_ID_SOCKET,
-    CONN_TYPE_ID_UNIX,
-    CONN_TYPE_ID_TLS,
-    CONN_TYPE_ID_RDMA,
+    CONN_TYPE_INVALID = 0,
+    CONN_TYPE_SOCKET,
```
PingXie:
```suggestion
    CONN_TYPE_INVALID = -1,
    CONN_TYPE_SOCKET = 0,
```

src/connection.c
```diff
@@ -72,19 +59,16 @@ int connTypeInitialize(void) {
     return C_OK;
 }
 
-ConnectionType *connectionByType(const char *typename) {
-    ConnectionType *ct;
+ConnectionType *connectionByType(int type) {
+    serverAssert(type > 0 && type < CONN_TYPE_MAX);
```
PingXie:
```suggestion
    serverAssert(type >= 0 && type < CONN_TYPE_MAX);
```

src/connection.c
```diff
@@ -124,26 +108,14 @@ ConnectionType *connectionTypeUnix(void) {
...
-    return -1;
-}
-
 void connTypeCleanupAll(void) {
     ConnectionType *ct;
     int type;
 
-    for (type = 0; type < CONN_TYPE_MAX; type++) {
+    // type = 0 is an invalid conn type.
+    for (type = 1; type < CONN_TYPE_MAX; type++) {
```
PingXie:
```suggestion
    for (type = 0; type < CONN_TYPE_MAX; type++) {
```

src/connection.c
```diff
@@ -155,7 +127,8 @@ int connTypeHasPendingData(void) {
     int type;
     int ret = 0;
 
-    for (type = 0; type < CONN_TYPE_MAX; type++) {
+    // type = 0 is an invalid conn type.
+    for (type = 1; type < CONN_TYPE_MAX; type++) {
```
PingXie:
```suggestion
    for (type = 0; type < CONN_TYPE_MAX; type++) {
```

src/connection.c
```diff
@@ -171,7 +144,8 @@ int connTypeProcessPendingData(void) {
...
```
PingXie:
```suggestion
    for (type = 0; type < CONN_TYPE_MAX; type++) {
```

src/connection.c
```diff
@@ -182,11 +156,12 @@ int connTypeProcessPendingData(void) {
 }
 
 sds getListensInfoString(sds info) {
-    for (int j = 0; j < CONN_TYPE_MAX; j++) {
+    // server.listeners[0] is reserved.
+    for (int j = 1; j < CONN_TYPE_MAX; j++) {
```
PingXie:
```suggestion
    for (int j = 0; j < CONN_TYPE_MAX; j++) {
```

src/server.c
```diff
@@ -3042,49 +3043,49 @@ void initListeners(void) {
...
         listener->port = server.rdma_ctx_config.port;
-        listener->ct = connectionByType(CONN_TYPE_RDMA);
+        listener->ct = ct;
         listener->priv = &server.rdma_ctx_config;
     }
 
     /* create all the configured listener, and add handler to start to accept */
     int listen_fds = 0;
-    for (int j = 0; j < CONN_TYPE_MAX; j++) {
+    for (int j = 1; j < CONN_TYPE_MAX; j++) {
```
PingXie:
```suggestion
    for (int j = 0; j < CONN_TYPE_MAX; j++) {
```

src/server.c
```diff
@@ -4458,7 +4459,7 @@ void incrementErrorCount(const char *fullerr, size_t namelen) {
 void closeListeningSockets(int unlink_unix_socket) {
     int j;
 
-    for (int i = 0; i < CONN_TYPE_MAX; i++) {
+    for (int i = 1; i < CONN_TYPE_MAX; i++) {
```
PingXie:
```suggestion
    for (int i = 0; i < CONN_TYPE_MAX; i++) {
```

src/server.c
```diff
@@ -7281,11 +7279,11 @@ __attribute__((weak)) int main(int argc, char **argv) {
             serverAssert(verifyClusterConfigWithData() == C_OK);
         }
 
-        for (j = 0; j < CONN_TYPE_MAX; j++) {
+        for (j = 1; j < CONN_TYPE_MAX; j++) {
```
PingXie:
```suggestion
        for (j = 0; j < CONN_TYPE_MAX; j++) {
```

src/connection.c
```diff
@@ -30,23 +30,10 @@
...
-        if (!tmpct) break;
-
-        /* ignore case, we really don't care "tls"/"TLS" */
-        if (!strcasecmp(typename, tmpct->get_type(NULL))) {
-            serverLog(LL_WARNING, "Connection types %s already registered", typename);
-            return C_ERR;
-        }
-    }
+    int type = ct->get_type();
+    serverAssert(type > 0 && type < CONN_TYPE_MAX && !connTypes[type]);
```
PingXie:
```suggestion
    serverAssert(type >= 0 && type < CONN_TYPE_MAX && !connTypes[type]);
```

src/connection.c
```diff
@@ -72,19 +59,16 @@ int connTypeInitialize(void) {
...
 
-    for (int type = 0; type < CONN_TYPE_MAX; type++) {
-        ct = connTypes[type];
-        if (!ct) break;
+    ConnectionType *ct = connTypes[type];
 
-        if (!strcasecmp(typename, ct->get_type(NULL))) return ct;
+    if (!ct) {
+        serverLog(LL_WARNING, "Missing implement of connection type %s", getConnectionTypeName(type));
+        return NULL;
```
PingXie:
```suggestion
```

src/connection.h
```diff
@@ -64,25 +65,35 @@ typedef enum {
...
+    CONN_TYPE_UNIX,
+    CONN_TYPE_TLS,
+    CONN_TYPE_RDMA,
 } ConnectionTypeId;
```
PingXie, line 74:
```suggestion
} ConnectionType;
```

PingXie (commented):
I think we are there. made one last suggestion and gave the invalid type a negative numeric value. it adds cognitive load if we have to explain/remember why we need to start the index for connection types from 1.

src/connection.h
cxljs: Hello, I think we can keep `ConnectionTypeId` since there is already a struct `ConnectionType`. What do you think?

cxljs:
Hello @PingXie , Thank you for your suggestions.

src/connection.h
PingXie: You are right. Let's keep `ConnectionTypeId`. Thanks for catching this!

PingXie (approved):
LGTM. Thanks @cxljs!

Issue #1600, opened 2025-01-21
[NEW] Non-voting primaries, voting empty primaries, voting replicas


zuiderkwast:
**The problem/use-case that the feature addresses**

The cluster bus traffic overhead is huge for huge clusters because:

1. All nodes ping each other in a full mesh
2. Voting nodes = all primaries with slots

**Description of the feature**

Decouple right-to-vote from primary-serving-slots, so any node can be a voting or non-voting node.

This can allow non-voting primaries, voting replicas and voting empty nodes. Modifying these details may be easier than to rewrite the cluster protocol to Raft (#384).

By having only a few voting nodes (such as five) in a large cluster with hundreds of nodes, voting becomes faster.

Then we can also limit the ping-pong traffic between non-voting nodes. For example, non-voting nodes don't need to ping other nodes. They can just reply to pings from other nodes. This makes the voting nodes a team of leaders and the non-voters a team of followers.

The difficulty seems to be for all the nodes to agree about which nodes can vote and how to count the votes, and to do the switch in a rolling upgrade.

**How to introduce it in a rolling upgrade**

Idea: Once all the voting nodes in a cluster support the new rules, then automatically switch to the new voting rules. Add some flag bits to the cluster bus header:

1. Flag bit "I support the new voting rules" (yes/no)
2. Flag bit "I want to be in the quorum" (yes/no)
3. Flag bit "New voting rules are activated" (yes/no)

When a node wins a failover election, it checks if all nodes in the current quorum support the new voting rules. If they do, the new node activates the new voting rules and sets the 3rd bit "New voting rules are activated". This is spread to all nodes using broadcast and gossip.

**Alternatives you've considered**

Raft consensus protocol

**Additional information**

This is just an idea. We need to consider edge cases...

zuiderkwast:
@enjoy-binbin Are the voting empty primaries the same idea as your arbiter nodes? You say @madolson doesn't like the idea of arbiter nodes? @PingXie also invited to discuss. Do you see any obvious catches with this approach?

enjoy-binbin:
> Are the voting empty primaries the same idea as your arbiter nodes?

Yes, it is the same idea, i would like to open source it if we accept.

> You say @madolson doesn't like the idea of arbiter nodes?

madolson mention this thought in here https://github.com/valkey-io/valkey/pull/634#issuecomment-2164364730. (maybe i got she wrong)

[...]

PR #2204, opened 2025-06-11
Add cluster-replica-priority to allow better ranking in auto failover

### LINKED ISSUE #1937: [NEW] Propagate node availability-zone information across cluster and make failover smarter

Currently nodes are placed in AZs either randomly or operators carefully place them in specific AZs for providing low latency to their applications. There could be situation where all the primaries (voting members) are placed in a single AZ and a network partition/unavailability of that AZ could cause write outage for indefinite period of time until other processes are triggered or manual operator intervenes.

Sometimes Valkey nodes are also placed in the same AZ as the application for low latency and that could easily get disrupted with a failover. Currently failover doesn't account for the AZ and picks the replica with the best offset/lexicographic order.

Propagate the `availability-zone` data stored in configuration across the cluster via cluster bus extensions and provide users with option to pick a smarter failover logic.

Operator performs the above described feature manually.

**Additional information**

Related: https://github.com/valkey-io/valkey/pull/1927#discussion_r2034384718

enjoy-binbin:
Aux field nodes.conf change was added in #4454.
=====

Currently nodes are placed in AZs either randomly or operators carefully
place them in specific AZs for providing low latency to their applications.
There could be situation where all the primaries (voting members) are placed
in a single AZ and a network partition/unavailability of that AZ could cause
write outage for indefinite period of time until other processes are triggered
or manual operator intervenes.

Sometimes nodes are also placed in the same AZ as the application for low
latency and that could easily get disrupted with a failover. Currently failover
doesn't account for the AZ and picks the replica with the best offset/lexicographic
order.

In some cases, we may want to give cluster replicas some priority in auto failover.
We have the cluster-replica-no-failover configuration option, we can ensure that a
replica does not actively failover, for example, we want the remote backup replica
node to never be automatically promoted. We want replicas in the same AZ to have
better priority so that they can always initiate elections faster when automatic
failover occurs.

We introduced the cluster-replica-priority configuration item, and each replica
node can freely configure a numerical priority. The replica with a smaller priority
will have a better rank when automatic failover occurs, so that elections can be
initiated faster. 

Note that at all times, we will use the replication offset as the most prioritized
reference indicator. For example, if the replication offsets of two replicas are
the same, the replica with a smaller priority will initiate an election first,
and the other replica will initiate an election later according to its rank.

We don't re-use the replica-priority logic since it is used in sentinel,
and replica-priority is a higher priority metric than replication offset there,
we don't want to use it (or make people understand it again).

A new DEBUG CLUSTER-REPLICA-PRIORITY command for testing.

This can closes #1937.


tests/unit/cluster/replica-priority.tcl
```diff
@@ -0,0 +1,65 @@
...
+        for {set i 0} {$i < 1024} {incr i} {
+            R 0 incr key_991803
+        }
+        wait_for_ofs_sync [srv 0 client] [srv -3 client]
+        wait_for_ofs_sync [srv 0 client] [srv -6 client]
+
+        # Set two different priorities, R 3 will have a better priority.
+        R 3 config set cluster-replica-priority 0
+        R 6 config set cluster-replica-priority 10
+
```
enjoy-binbin, line 20: This test may currently flaky because we are not waiting for the priority to propagate.

zuiderkwast:
@valkey-io/valkey-committers This one is marked for 9.0. If we want to include it, we need to make a decision quickly and also review it within the next days.

madolson:
> we need to make a decision quickly and also review it within the next days.

I'm fine making a decision, but I'm against adding features to 9.0 at this point in time. I view rc1 as the main cutoff for new features unless we think it's strategically important. 

Specifically for this feature, I think I'm OK with it. I don't really understand why folks would have a mix of replicas in the same AZ vs not, but if it's a workload folks use I'm OK merging it.

zuiderkwast:
> I don't really understand why folks would have a mix of replicas in the same AZ vs not

@enjoy-binbin do you want to explain?

I'm not exactly sure, but I guess it can also be used for preserving anti-affinity on failover. The replica can be set to avoid being promoted, but in case of disaster recovery, it can step in.

enjoy-binbin:
> I don't really understand why folks would have a mix of replicas in the same AZ vs not

Is this a question that, why are the replicas placed in different AZs? We have a deployment mode that deploys replicas in non-main AZs to provide AZ-level high availability. Something like in the main AZ, there will be the primary and replica, and in a bak-AZ, there will be a replica. In this case, if a auto-failover happen, we want replicas in the main AZ to have the higher priority so that in the last, the primary is still in the main AZ after the failover. And the application APP is also in the same main AZ, so they have the good network conditions.

BTW, do we want to push #1600?

murphyjacob4:
This PR has been punted a few times. Do we want it for 9.2? It generally makes sense if you have a deployment where replicas are not equal (one is in a preferred AZ, one has more/better CPUs, one is on a newer version of Valkey, etc).

I will callout it only really matters for idle write-traffic clusters. For clusters with busy write-traffic, offset will still be the most important. If we add it, we should add documentation explaining this is only used when all else is equal.

enjoy-binbin:
Yes, let's do it. We do have this function internally (just implemented differently). I will refresh it later, we just lack an extra pair of eyes to review it, everyone seems to be short on bandwidth.

src/cluster_legacy.c
```diff
@@ -2997,6 +3008,17 @@ static uint32_t writePingExtensions(clusterMsg *hdr, int gossipcount) {
     totlen += getShardIdPingExtSize();
     extensions++;
 
+    /* Populate replica_priority */
+    if (cursor != NULL) {
+        clusterMsgPingExtReplicaPriority *ext = preparePingExt(cursor, CLUSTERMSG_EXT_TYPE_REPLICA_PRIORITY, getReplicaPriorityExtSize());
+        ext->replica_priority = myself->replica_priority;
```
murphyjacob4:
```suggestion
        ext->replica_priority = htonl(myself->replica_priority);
```

src/cluster_legacy.c
```diff
@@ -3052,6 +3074,9 @@ void clusterProcessPingExtensions(clusterMsg *hdr, clusterLink *link) {
         } else if (type == CLUSTERMSG_EXT_TYPE_SHARDID) {
             clusterMsgPingExtShardId *shardid_ext = (clusterMsgPingExtShardId *)&(ext->ext[0].shard_id);
             ext_shardid = shardid_ext->shard_id;
+        } else if (type == CLUSTERMSG_EXT_TYPE_REPLICA_PRIORITY) {
+            clusterMsgPingExtReplicaPriority *priority_ext = (clusterMsgPingExtReplicaPriority *)&(ext->ext[0].replica_priority);
+            sender->replica_priority = priority_ext->replica_priority;
```
murphyjacob4:
```suggestion
            sender->replica_priority = ntohl(priority_ext->replica_priority);
```

tests/unit/cluster/replica-priority.tcl
murphyjacob4: Are you planning to fix it? At least an after 3000 would help. Better yet, some kind of wait_for_condition loop

src/cluster_legacy.c
```diff
@@ -4740,12 +4768,22 @@ int clusterGetReplicaRank(void) {
...
+        /* Replication offset is the same.
+         * The smaller the replica-priority, the better the ranking, that is,
+         * the smaller the rank and the smaller the delay. */
+        if (primary->replicas[j]->replica_priority < myself->replica_priority) rank++;
+        if (primary->replicas[j]->replica_priority != myself->replica_priority) continue;
+
+        /* Replica priority is the same.
+         * If the replication offsets are the same, the one with the lexicographically
+         * smaller node id will have a lower rank to avoid simultaneous elections
+         * of replicas. */
```
murphyjacob4:
```suggestion
         * If the replication offsets and replica priorities are the same, the one with
         * the lexicographically smaller node id will have a lower rank to avoid
         * simultaneous elections of replicas. */

```

src/cluster_legacy.c
```diff
@@ -2997,6 +3008,17 @@ static uint32_t writePingExtensions(clusterMsg *hdr, int gossipcount) {
     totlen += getShardIdPingExtSize();
     extensions++;
 
+    /* Populate replica_priority */
+    if (cursor != NULL) {
```
murphyjacob4: Should we skip sending if it isn't configured?

valkey.conf
```diff
@@ -1888,6 +1888,9 @@ aof-timestamp-enabled no
 #
 # cluster-manual-failover-timeout 5000
 
+# xxx
```
murphyjacob4: Need to add a comment

src/cluster_legacy.c
enjoy-binbin: 0 is also a valid value. Of course, we have other ways to handle it if needed.

tests/unit/cluster/replica-priority.tcl
enjoy-binbin: It isn't included in commands like `cluster shards` or `cluster slots` (unlike AZ information), so we can't use `wait_for_condition`.

Let me think if there are any other ways.

enjoy-binbin: I added a new minor DEBUG CLUSTER-REPLICA-PRIORITY command

src/cluster_legacy.c
murphyjacob4: Gotcha, yeah I guess you need to broadcast when it is swapped from non-zero to zero.

murphyjacob4: Actually - do you think it is cleaner if we don't send it when it is set to zero? When we get a ping that doesn't have it, we can update the value to zero.

It will prevent the extra gossip for the 99%+ of clusters that don't need it. Also for nodes that are on older version, it will treat them correctly as priority zero. It would handle an edge case where a node is downgraded in place (e.g. 9.2 -> 9.1) without changing the node ID.

enjoy-binbin: Good point, handled in 141a929ef0afedd2ffb3940a7f77aaa9175d684c

murphyjacob4 (approved):
LGTM. Guess we need @valkey-io/core-team to finalize the decision. 👍 / 👎 ?

zuiderkwast:
We should persist the replica priorities of all nodes in nodes.conf in an aux field too (when it's non-zero), right? Let's do a follow-up?

enjoy-binbin:
I will handle it later this day.

PR: #4454.

PR #2206, opened 2025-06-11
Prevent getNodeByQuery from leaking DB changes into client

gmbnomis:
To support multiple databases in cluster mode (see #1671), `getNodeByQuery` temporarily switches databases when tracking `SELECT` statements during slot migration/import. The intended logic is to revert any database change after the operation. However, this approach is flawed: in some transactions the database change is not properly reverted, causing the client to remain on the wrong database.

For example, if a transaction includes `SELECT` statements, the current database may be changed even if the transaction is never executed (see added test).

Fix the issue by saving the original database once and restoring to it after a switch.

### DIFF
--- a/tests/unit/cluster/slot-migration.tcl
+++ b/tests/unit/cluster/slot-migration.tcl
@@ -535,6 +535,19 @@ start_cluster 3 3 {tags {external:skip cluster} } {
         set result [catch {R $primary_id_src exec} err]
         assert_match "ERR DB index is out of range" $err
 
+        # Multi/Exec on source before migration, mixed key existence
+        R $primary_id_src select 0
+        R $primary_id_src multi
+        R $primary_id_src exists "{3560}key1"
+        R $primary_id_src select 1
+        R $primary_id_src exists "{3560}key2"
+        R $primary_id_src select 2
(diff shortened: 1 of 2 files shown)


tests/unit/cluster/slot-migration.tcl
```diff
@@ -535,6 +535,19 @@ start_cluster 3 3 {tags {external:skip cluster} } {
...
+        # Multi/Exec on source before migration, mixed key existence
+        R $primary_id_src select 0
+        R $primary_id_src multi
+        R $primary_id_src exists "{3560}key1"
+        R $primary_id_src select 1
+        R $primary_id_src exists "{3560}key2"
+        R $primary_id_src select 2
+        R $primary_id_src exists "{3560}no_key"
+        set result [catch {R $primary_id_src exec} err]
+        assert_match "TRYAGAIN Multiple keys request during rehashing of slot" $err
```
zuiderkwast: Minor: We can use `assert_error`.

```suggestion
        assert_error "TRYAGAIN Multiple keys*" {R $primary_id_src exec}
```

zuiderkwast (approved):
Great, looks good!

Issue #2022, opened 2025-04-29
[High level design] Allow Read requests on I/O threads


touitou-dan:
**Overview**
We propose enhancing Valkey with multi-threaded command execution capabilities to significantly increase throughput from the current  0.92M  GET request per second (CME) to over 3.1M GET requests per second. Our prototype demonstrates that offloading read commands to the aync threads can yield up to 12.7x performance improvement for CPU-intensive operations.

**Background**
 Valkey 8.0 introduced async threads to handle input/output operations while the main thread focuses on command execution. Despite this advancement, a critical bottleneck remains: all command execution still occurs sequentially in the single main thread, which becomes increasingly problematic as systems scale to more cores and memory.  Furthermore, the current architecture's benefits are minimal for workloads dominated by CPU-intensive operations, since offloading only I/O tasks provides limited performance gains when processing is the primary constraint.

**Proposed Solution**
The new dict-per-slot architecture introduced in Valkey 8.0 provides an opportunity for parallelization. Since each slot has its own dictionary, commands operating on different slots could theoretically execute in parallel without contention.
We suggest leveraging this feature and propose an architectural update to how the main thread interacts with other threads. In this design, the main thread still maintains its role as a central scheduler, distributing tasks among threads to ensure safe access to clients and data structures. The main thread communicates with the worker threads through lock-free producer-consumer queues.
![Image](https://github.com/user-attachments/assets/33f5dbc0-7a02-4f00-ac00-155da5e1dafc)
The key enhancement is expanding the job types beyond I/O operations to include execute-command jobs, allowing worker threads to independently process commands. By maintaining this centralized coordination model, we eliminate the need for complex synchronization mechanisms such as locks or atomic operations. Importantly, this enhancement requires only minimal changes to the existing codebase while significantly improving resource utilization and performance scalability 

**Read commands execution**
The picture below depicts the execution of a GET command in the proposed architecture:
![Image](https://github.com/user-attachments/assets/f17e7cb1-0f57-415d-9353-c1dd00c0ce58)
[...]

PR #2208, opened 2025-06-11
Read commands offloading to IO threads

uriyage:
Issue: https://github.com/valkey-io/valkey/issues/2022
**Joined work with @touitou-dan, @akashkgit  @nadav-levanoni**

### Overview

Valkey 8.0 introduced worker threads for I/O operations, but a critical bottleneck remains: all command execution still occurs  sequentially in the main thread. This architecture limitation becomes increasingly problematic as systems scale to more cores and memory,  and provides minimal performance gains for CPU-intensive workloads where processing, not I/O, is the primary constraint.

This PR extends worker thread capabilities to execute read commands in parallel, removing the main thread bottleneck while maintaining data consistency. it uses the same centralized coordination model where the main-thread is acting as a scheduler as in valkey 8.0 eliminating the need for complex synchronization mechanisms such as locks or atomic operations.

### Key Benefits:

* 2.3x throughput improvement: 1.3M → 3M requests per second for read commands like GET
* Up to 17x acceleration for CPU-intensive operations like HGETALL
* Limited code changes required

#### Extended Job Types

Worker threads now handle both I/O operations and read command execution, expanding beyond the I/O-only limitation of version 8.0.

* Valkey 8.0: Main Thread (All Commands) + Worker Threads (I/O Only)
* This PR: Main Thread (Write Commands + Coordination) + Worker Threads (Read Commands + I/O)

### Unified Response Handling

All worker responses flow through a single multi-producer, single-consumer queue, eliminating the current need to constantly scan client lists

### Continuation Tasks

When workers encounter operations requiring global data access (like expired key deletion), they create continuation tasks for the main thread to execute, ensuring data consistency.

### Slot Access Control Mechanism

To prevent race conditions and ensure correct command execution order, the main thread uses deferred queues - that manage task scheduling and synchronization.

**Structure and Components**
Each deferred queue contains:

* Blocked clients list: Clients waiting for command execution
* Jobs list: Background tasks that need execution on specific slots or the entire database
* Reference counter: Tracks active operations to determine slot availability

The system maintains:

* Per-slot deferred queues: One queue for each of the 16,384 hash slots to synchronize read/write commands accessing the same slot.
* Global exclusive queue: A single queue (deferredCmdExclusive) for operations requiring exclusive database access

**How It Works**
Example 1: Exclusive Command Handling
When an EVAL command arrives:

1. If the exclusive queue is busy (refcount > 0), the client gets blocked and added to the pending clients list
2. All subsequent commands are also queued until the exclusive operation completes
3. Once the EVAL finishes, the main thread processes the blocked commands in order

Example 2: Slot-level Synchronization
Consider a GET and SET command targeting the same slot:

1. The GET command gets offloaded to a worker thread, incrementing the slot's refcount
2. The SET command (requiring exclusive slot access) gets blocked on the slot queue
3. When the worker completes the GET and responds, the refcount decreases
4. The SET command then executes once the slot becomes available

**Worker Thread Deferred Jobs**
Worker threads sometimes need to execute operations that access global data structures. Instead of doing this immediately, they:

1. Create deferred jobs and add them to their thread-local list (thread_delayed_jobs)
2. At command-execution completion, the thread post the entire list to the main thread's response queue
3. The main thread then executes these deferred operations safely

**Current deferred job types include:**

* Expired key processing: Deleting expired keys and propagating DEL commands to replicas
* Rehashing completion: Finalizing incremental rehashing by updating global dictionary structures
* Error statistics: Updating global error counters after command failures

**Special Case: ServerCron**

The ServerCron function requires exclusive database access. When other commands are running in parallel, ServerCron gets enqueued as a deferred job on the exclusive queue,  ensuring it runs only when no other operations are active.

### Commands Offloading

Commands that can be offloaded to I/O threads are commands that are mark with READONLY flag in their json files.

However, different types of commands have  different exclusivity requirements:
Database-exclusive commands require complete isolation and cannot run in parallel with any other offloaded command. These include:

* Write commands that don't target specific slots
* The EXEC command (which may contain commands affecting the entire database)
* Administrative commands (marked with CMD_ADMIN flag)
* Commands with the CMD_NO_MANDATORY_KEYS flag (like EVAL), where the affected keys cannot be determined in advance

Slot-exclusive commands have more limited restrictions - they can run in parallel with other read commands as long as those commands target different slots. They are only blocked by commands targeting the same slot.
This tiered approach to command exclusivity allows the system to maximize parallelism while maintaining data consistency and avoiding conflicts between concurrent operations.
A new configuration io-threads-do-command-offloading was added to disable command offloading.

### Event Processing (epoll)

Three key improvements were made to the event polling process:

a) Client Structure Prefetching: When epoll_wait returns a batch of file descriptors, the system now prefetches the associated client structures to improve cache performance
during event processing. This is implemented through a new prefetch callback function added to the aeEventLoop structure, which proactively loads client data into memory before it
's accessed.

b) Epoll Round-robin offloading to Worker Threads: Similar to version 8.0, epoll_wait operations can be offloaded to worker threads. The implementation uses round-robin scheduling across available worker threads, with epoll jobs receiving higher priority than regular I/O and command processing jobs within each worker thread.

c) Batch Size Optimization: The maxevents parameter for epoll_wait was changed from using eventLoop→setsize (which could be quite large) to a fixed value of 200. This change addresses a performance regression that occurred between Linux kernel versions 5.x and 6.x, where larger batch sizes negatively impacted epoll performance.

### Client Management 

**server.current_client converted to thread variable**
Converted global server.current_client and server.executing_client to thread-local variables (_current_client, _executing_client) with accessor macros. This enables thread-safe client context management for concurrent command execution on IO threads.

**New block type BLOCKED_SLOT for clients blocking on slots**
Added BLOCKED_SLOT blocking type for clients waiting on busy slots during command offloading. Includes new blockingState fields (slot_pending_list, pending_client_node) . Allows proper slot contention handling by queuing clients until slots become available.

**Async free client that are handled by IO threads**
Modified client freeing to eliminate busy waiting for IO thread operations. Previously, freeClient() would busy wait using waitForClientIO() when clients had ongoing IO operations, creating complexity in edge cases. Now defer freeing until IO operations complete, preventing blocking and simplifying the freeing logic.

### Limitations 

**Modules not supported by default**

* Command offloading to IO threads is disabled by default when modules are loaded to ensure compatibility and prevent potential conflicts. This can be overridden by setting io-threads-do-command-offloading-with-modules to yes, but should be used with caution as module behavior with offloaded commands is not guaranteed

**Keyspace miss notifications not supported**

* Commands cannot be offloaded to IO threads when keyspace miss notifications (NOTIFY_KEY_MISS) are enabled. This limitation exists because miss notifications require synchronous execution in the main thread.

**pipeline commands**

* Since pipeline commands may access different slots, which could result in sending each command to a different thread, the performance gain would be marginal or even negative. This is due to multiple write calls and numerous round trips between the main thread and I/O threads. Therefore, the main thread will process all commands except the last one, which is handled by a separate thread. This same thread also writes all command replies in a single write call.

### Performance Evaluation

 We implemented the proposed update into a prototype to evaluate the performance gain in read and read/write scenario . 

**Test Environment**

* Server: c7gn.16xlarge instance
* Clients: 3 or 4  c7g.16xlarge instances
* Configuration: All instances in same placement group (IAD region)
* Valkey 9.0 unstable as of 6/9/2025 : 8 io-threads  
* 3M: 19 threads (1 main thread, 18 worker threads) 
* Mode: All tests in cluster mode
* NIC IRQ  - pinned on cores 57--63

**Dataset**

* Strings: 3M keys with 512-byte values - GET and SET as read and write command respectively 
* Hashes: 1M hashes, each with ~ 50 fields (70 bytes/field)  - HGETALL and HSET as read and write command respectively
* Sorted Sets: 1M sorted sets, each with 50 members (70 bytes/member) - ZRANK and ZADD as read and write command respectively 
* Lists: 1M lists, each with ~50 elements (70 bytes/element) - LINDEX and LSET as read and write command respectively 

**Benchmark Scenarios**

1. 100% read operations
2. 80% read / 20% write clients 

**String Operations**

|     Workload        |     Valkey 9.0     |       3M               |     
|---------------------|--------------------|-------------------|
|    100% Read      |     1299K           |     3,003K          |    
| 80% R, 20% W  |      1242K           |     1,931K           |                 

**Hash Operations**

|     Workload        |     Valkey  9.0      |         3M             |    3M 40 threads       |   
|---------------------|--------------------|-------------------|----------------------------| 
|    100% Read      |       146K             |     1,415K           |         2,626K                 |      
| 80% R, 20% W   |       172K             |    1,584K           |         1,526K                  |    

**Sorted Set Operations**

|     Workload       |     Valkey 9.0       |            3M          |
|---------------------|---------------------|--------------------|
|   100% Read      |       365K             |       2,537K         |            
| 80% R, 20% W |      330K             |         780K          |

**List Operations**

|    Workload       |     Valkey 9.0      |            3M            |  
|--------------------|--------------------| --------------------|
|     100% Read   |         659K          |          2956K       | 
| 80% R, 20% W |         831K          |          1778K         |
---------------------------------------------------------------


madolson:
@uriyage Can you prioritize doing the merge so that we can properly run the tests?

src/ae.h
```diff
@@ -47,6 +47,7 @@
                          loop iteration. Useful when you want to persist \
                          things to disk before sending replies, and want \
                          to do that in a group fashion. */
+#define AE_PREFETCH 8 /* With PREFETCH, call prefetch callback for the events */
```
madolson: The naming is tightly coupled to the implementation, can we just call it like AE_PRE_READABLE_HOOK or something?

src/ae.h
```diff
@@ -113,8 +115,10 @@ typedef struct aeEventLoop {
     aeBeforeSleepProc *beforesleep;
     aeAfterSleepProc *aftersleep;
     aeCustomPollProc *custompoll;
+    aePrefetchProc *prefetch;
     pthread_mutex_t poll_mutex;
     int flags;
+    int epoll_batch_size; /* Optional batch size for epoll_wait */
```
madolson: It would be nice to explain what optional means.

src/server.c
```diff
@@ -2813,6 +2819,10 @@ void initServer(void) {
         serverLog(LL_WARNING, "Failed creating the event loop. Error message: '%s'", strerror(errno));
         exit(1);
     }
+    /* Set the epoll batch size for the server event loop */
+    server.el->epoll_batch_size = AE_EPOLL_EVENTS_BATCH_SIZE;
```
madolson: Ideally add a helper and don't directly set the variable.

src/commands/get.json
```diff
@@ -8,7 +8,8 @@
         "function": "getCommand",
         "command_flags": [
             "READONLY",
-            "FAST"
+            "FAST",
+	        "CAN_BE_OFFLOADED"
```
madolson: Lots of indentation errors in these files.

src/config.c
```diff
@@ -3262,6 +3262,8 @@ standardConfig static_configs[] = {
     createIntConfig("io-threads", NULL, DEBUG_CONFIG | IMMUTABLE_CONFIG, 1, IO_THREADS_MAX_NUM, server.io_threads_num, 1, INTEGER_CONFIG, NULL, NULL), /* Single threaded by default */
     createIntConfig("events-per-io-thread", NULL, MODIFIABLE_CONFIG | HIDDEN_CONFIG, 0, INT_MAX, server.events_per_io_thread, 2, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("prefetch-batch-max-size", NULL, MODIFIABLE_CONFIG, 0, 128, server.prefetch_batch_max_size, 16, INTEGER_CONFIG, NULL, NULL),
+    createBoolConfig("io-threads-do-commands-offloading", NULL, MODIFIABLE_CONFIG, server.io_threads_do_commands_offloading, 1, NULL, NULL),                           /* Command offloading enabled by default */
+    createBoolConfig("io-threads-do-commands-offloading-with-modules", NULL, MODIFIABLE_CONFIG, server.io_threads_do_commands_offloading_with_modules, 0, NULL, NULL), /* Module command offloading disabled by default */
```
madolson, line 3271: This shouldn't be server configuration. Either the module should be the one deciding if it can offload work "per command" or it should be a module wide configuration.

src/config.c
```diff
@@ -3262,6 +3262,8 @@ standardConfig static_configs[] = {
     createIntConfig("io-threads", NULL, DEBUG_CONFIG | IMMUTABLE_CONFIG, 1, IO_THREADS_MAX_NUM, server.io_threads_num, 1, INTEGER_CONFIG, NULL, NULL), /* Single threaded by default */
     createIntConfig("events-per-io-thread", NULL, MODIFIABLE_CONFIG | HIDDEN_CONFIG, 0, INT_MAX, server.events_per_io_thread, 2, INTEGER_CONFIG, NULL, NULL),
     createIntConfig("prefetch-batch-max-size", NULL, MODIFIABLE_CONFIG, 0, 128, server.prefetch_batch_max_size, 16, INTEGER_CONFIG, NULL, NULL),
+    createBoolConfig("io-threads-do-commands-offloading", NULL, MODIFIABLE_CONFIG, server.io_threads_do_commands_offloading, 1, NULL, NULL),                           /* Command offloading enabled by default */
```
madolson, line 3270: General question of why do we think this should be a config? In the fullness of time do we expect users to tune this, or should it dynamically optimize itself for performance. I'm OK with having a hidden config here primarily for testing and tuning.

src/kvstore.c
```diff
@@ -65,7 +66,7 @@ struct _kvstore {
     unsigned long long bucket_count;          /* Total number of buckets in this kvstore across hash tables. */
     unsigned long long *hashtable_size_index; /* Binary indexed tree (BIT) that describes cumulative key frequencies up until
                                                * given hashtable-index. */
-    size_t overhead_hashtable_lut;            /* Overhead of all hashtables in bytes. */
+    _Atomic size_t overhead_hashtable_lut;    /* Overhead of all hashtables in bytes, Atomic as it may be update by the IO threads */
```
madolson:
```suggestion
    _Atomic size_t overhead_hashtable_lut;    /* Overhead of all hashtables in bytes, Atomic as it may be updated by the IO threads for rehashing */
```

src/commands/get.json
madolson, line 1: I don't particularly understand the list of commands that are supported, other commands that seem to be supportable are:
Bitcount
bitfiled_ro
bitpos
all the geo read commands
getrange
getbit
mget
most the set readonly commands
most of the string readonly commands

tests/unit/maxmemory.tcl
```diff
@@ -51,6 +51,11 @@ start_server {tags {"maxmemory" "external:skip"}} {
     }
 
     foreach {client_eviction} {false true} {
+        # Skip client eviction test when IO threads are enabled because client freeing
+        # may be deferred with IO threads, making the eviction behavior unpredictable
+        if {$client_eviction && $::io_threads} {
```
madolson, line 60: I guess I don't understand why the behavior is unpredictable. If we generate 100 MB of output on a client, it should still eventually get evicted. We should still be able to emulate that. I think it's OK to punt it out of the main PR, but I don't think we should just skip it.

tests/unit/io-threads.tcl
```diff
@@ -0,0 +1,351 @@
...
+            r config set io-threads-do-commands-offloading yes
+        }
+
+        test {Key expiry is postponed when read from main thread} {
+            # This test verifies that when a key with expiry is read from the IO thread,
+            # its expiry deletion is postponed to the main-thread to prevent race conditions with the main-thread
+            set updated_info_stats [r INFO]    
+            set initial_postponed [getInfoProperty $updated_info_stats io_threaded_postponed_jobs_to_mainthread]
+
+            # Set a key with a short expiry time (2 seconds)
```
madolson: You can set the expiry to be like 1ms, then you don't have to wait for 2 seconds. You can also batch all of these operations together in a multi to make sure they're executed atomically and a slow test doesn't cause the `assert_equal "val" [r get key]` to fail.

tests/unit/io-threads.tcl
```diff
@@ -0,0 +1,351 @@
...
+            # Verify the command was not offloaded after disabling
+            assert_equal $mid_offloaded_commands $final_offloaded_commands
+
+            # Re-enable command offloading for other tests
+            r config set io-threads-do-commands-offloading yes
+        }
+
+        test {Key expiry is postponed when read from main thread} {
+            # This test verifies that when a key with expiry is read from the IO thread,
+            # its expiry deletion is postponed to the main-thread to prevent race conditions with the main-thread
```
madolson, line 326: There is a debug configuration to disable active expiration, that should probably also get set.

src/io_threads.c
```diff
@@ -5,13 +5,156 @@
...
+/******************************************************************************************************
+ * Multi-Producer Single-Consumer Queue Implementation
+ * This queue allows multiple producer threads to safely enqueue items
+ * that will be consumed by a single consumer thread. It's designed for
+ * passing messages from IO-threads to the Main-thread.
+ *******************************************************************************************************/
+
+/* Multi-Producer Single-Consumer Queue structure
+ * Cache line padding (64 bytes) is used to prevent false sharing between counter */
+typedef struct IoToMTQueue {
```
madolson, line 28: I would still strongly prefer that the MPSC queue be pulled out to it's own file. I would also it be called `MPSPCQueue` and be written in such a way that it's independent from the underlying IO threading implementation. I don't know what IOToMT means for example, and there is a lot of mixed explanation.

src/io_threads.c
```diff
@@ -126,30 +289,390 @@ static void IOJobQueue_peek(const IOJobQueue *jq, job_handler *handler, void **d
...
+        listLinkNodeTail(q->deferred_jobs, jobNode);
+    }
+}
+
+/* Returns whether the given command requires exclusive access to the whole database. */
+static int isDBExclusiveCmd(struct serverCommand *cmd, int slot) {
+    /* If no slot is specified but the client command changes the keyspace, we assume it is an exclusive command */
+    if (slot == -1 && (cmd->flags & CMD_WRITE)) return 1;
+    /* The exec command can contain commands that may affect the whole database */
+    if (cmd->proc == execCommand) return 1;
```
madolson, line 411: One concern I have is inconsistent performance for normal dataplane operations which will show up as P99+ latency spikes. Waiting for an exclusive lock for flushall or replicaof seems OK. I think waiting for exclusive locks on exec is not ideal, it's commonly used in a lot of real workloads and is our answer to many workloads. For Exec specifically, you should be able to introspect into the exec to see if it will execute any `isDBExclusiveCMD`.

src/io_threads.c
```diff
@@ -310,17 +869,110 @@ void initIOThreads(void) {
...
+
+    if (!(c->cmd->flags & CMD_CAN_BE_OFFLOADED)) {
+        return C_ERR;
+    }
+
+    if (!server.cluster_enabled) {
+        return C_ERR; /* Avoid offloading commands in non cluster mode. */
+    }
+
+    if (server.notify_keyspace_events & NOTIFY_KEY_MISS) {
```
madolson: We should have a test that this is gracefully handled if this config changes *after* the command has been offloaded.

hpatro (commented):
> Commands that can be offloaded to I/O threads are marked with the CAN_BE_OFFLOADED flag in their json files. However, different types of commands have different exclusivity requirements:
Database-exclusive commands require complete isolation and cannot run in parallel with any other offloaded command. 

Can't this be done implicitly instead of defining it via a flag explicitly? With that we would be aware all read commands will be offloaded and a dev doesn't need to check if it's enabled. In the future, we also don't risk of missing new read command being not offloaded.

src/io_threads.c
```diff
@@ -126,30 +289,390 @@ static void IOJobQueue_peek(const IOJobQueue *jq, job_handler *handler, void **d
...
+    } else {
+        if (q->deferred_jobs == NULL) {
+            q->deferred_jobs = listCreate();
+        }
+        listLinkNodeTail(q->deferred_jobs, jobNode);
+    }
+}
+
+/* Returns whether the given command requires exclusive access to the whole database. */
+static int isDBExclusiveCmd(struct serverCommand *cmd, int slot) {
```
```suggestion
static int isServerExclusiveCmd(struct serverCommand *cmd, int slot) {
```
These aren't really DB exclusive, since we don't allow those to be concurrently executed.

src/server.c
```diff
@@ -91,6 +91,9 @@
 
 struct sharedObjectsStruct shared;
 
+__thread client *_current_client;   /* The client that triggered the command execution (External or AOF). */
+__thread client *_executing_client; /* The client executing the current command (possibly script or module). */
```
madolson, line 97:
```suggestion
__thread client *current_client;   /* The client that triggered the command execution (External or AOF). */
__thread client *executing_client; /* The client executing the current command (possibly script or module). */
```

src/server.c
```diff
@@ -6001,10 +6016,17 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "io_threaded_reads_processed:%lld\r\n", server.stat_io_reads_processed,
                 "io_threaded_writes_processed:%lld\r\n", server.stat_io_writes_processed,
                 "io_threaded_freed_objects:%lld\r\n", server.stat_io_freed_objects,
+                "io_threaded_reads_pending:%lld\r\n", server.stat_io_reads_pending,
```
madolson, line 6147: I know this is late, but we should consider an IO thread section. There is a lot of information here now and it will be <empty> for most users.

src/networking.c
```diff
@@ -1751,19 +1773,19 @@ void clearClientConnectionState(client *c) {
...
+/* Free the client structure and all the data associated with it.
+ * Returns 0 if the client was not freed immediately, but scheduled for
+ * asynchronous freeing, and 1 if the client was freed immediately. */
+int freeClient(client *c) {
     listNode *ln;
 
     /* If a client is protected, yet we need to free it right now, make sure
      * to at least use asynchronous freeing. */
-    if (c->flag.protected || c->flag.protected_rdb_channel) {
+    if (c->flag.protected || c->flag.protected_rdb_channel || clientIOInProgress(c)) {
```
```suggestion
    if (c->flag.protected || c->flag.protected_rdb_channel || clientHandlingThreadedIO(c)) {
```
Something should indicate this is on the thread.

src/networking.c
```diff
@@ -5452,11 +5428,21 @@ void evictClients(void) {
...
+            }
             sds ci = catClientInfoString(sdsempty(), c, server.hide_user_data_from_log);
             serverLog(LL_NOTICE, "Evicting client: %s", ci);
-            freeClient(c);
             sdsfree(ci);
             server.stat_evictedclients++;
+            if (freeClient(c) == 0) {
+                /* The client is protected and will be closed later.
+                 * We don't want to continue evicting clients in this case
+                 * since it can cause multiple clients to be evicted unnecssarily */
```
```suggestion
                 * since it can cause multiple clients to be evicted unnecessarily */
```

src/io_threads.c
```diff
@@ -126,30 +289,390 @@ static void IOJobQueue_peek(const IOJobQueue *jq, job_handler *handler, void **d
...
+    /* If no slot is specified but the client command changes the keyspace, we assume it is an exclusive command */
+    if (slot == -1 && (cmd->flags & CMD_WRITE)) return 1;
+    /* The exec command can contain commands that may affect the whole database */
+    if (cmd->proc == execCommand) return 1;
+    /* If no mandatory keys are specified, we can't determine which slot will be accessed */
+    if (cmd->flags & CMD_NO_MANDATORY_KEYS) return 1;
+    /* Any Admin level command needs full exclusivity as it impacts system-wide behaviour */
+    if (cmd->flags & CMD_ADMIN) return 1;
```
madolson, line 415: This isn't really a great assumption. Some stuff like slowlog and acl log shouldn't really impact the wide system, we shouldn't need an exclusive lock to execute those. Many systems might have automation hitting this (like ours) which would cause random dips in performance.

src/io_threads.c
```diff
@@ -126,30 +289,390 @@ static void IOJobQueue_peek(const IOJobQueue *jq, job_handler *handler, void **d
...
+    int cur_tid;
+    list *deferred_jobs;
+    list *pending_clients;
+} deferredQueue;
+
+/* Global queues for deferred commands */
+deferredQueue deferredCmdExclusive = {0};
+deferredQueue slot_use_info[16384] = {0};
+
+typedef struct delayedJob {
```
```suggestion
typedef struct deferredJob {
```
Presumably?

src/memory_prefetch.c
```diff
@@ -235,6 +235,19 @@ void processClientsCommandsBatch(void) {
...
+static int isCommandBeingOffloaded(client *c) {
+    if (!server.io_threads_do_commands_offloading) {
+        return 0;
+    }
+
+    if (!server.cluster_enabled) {
+        return 0;
+    }
+
+    return (c->io_parsed_cmd->flags & CMD_CAN_BE_OFFLOADED) && (c->querybuf == NULL);
```
madolson: Why are the checks here different than the checks in trySendProcessCommandToIOThreads? Also worth commenting here why we are checking for c->querybuf.

src/networking.c
```diff
@@ -5543,7 +5542,50 @@ void ioThreadWriteToClient(void *data) {
     } else {
         _writeToClient(c);
     }
+    c->io_write_state = CLIENT_COMPLETED_IO;
+    threadRespond(c, R_WRITE);
+}
+
+void ioThreadProcessCommand(void *data) {
```
madolson: This feels like it's in the wrong file. It's more aligned with the process command work in server.c but has the naming convention of io threading. I think this is also the offloaded call variant, not process command.

src/io_threads.c
```diff
@@ -616,10 +1255,237 @@ int trySendAcceptToIOThreads(connection *conn) {
...
+void threadRespond(client *c, jobResponseType r) {
+    if (r == R_COMMAND) {
+        /* Make sure to send first the deferred jobs list */
+        threadRespondJobList();
+    }
+
+    IoToMTQueueProduce((uint64_t)c | (uint64_t)r, 0);
+}
+
+static void processClientIOCommandDone(client *c) {
```
madolson, line 1417: Can we deduplicate this code with the code in server.c, I don't like having two places we need to update. Ideally let's have a function in server.c that covers the overlap, and we can call that function from here.

madolson (dismissed):
Nice work! 

I'm primarily concerned about pipelining, multi-exec, and other operational commands that might give inconsistent performance. Commands queueing up periodically because we require the exclusive lock can cause P99 latency spikes. It would be nice to see other realistic workloads for testing performance. It would also be nice to understand how large the cluster can scale while still getting these improvements, can 500 primary clusters (~32 slots per primary) still achieve high enough concurrency?

madolson:
(Ooops, didn't mean to approve, but I can't dismiss it while there are merge conflicts apparently?

src/blocked.c
```diff
@@ -225,7 +227,7 @@ void unblockClient(client *c, int queue_for_reprocessing) {
         serverAssert(c->bstate->postponed_list_node);
         listDelNode(server.postponed_clients, c->bstate->postponed_list_node);
         c->bstate->postponed_list_node = NULL;
-    } else if (c->bstate->btype == BLOCKED_SHUTDOWN) {
+    } else if (c->bstate->btype == BLOCKED_SHUTDOWN || c->bstate->btype == BLOCKED_SLOT) {
         /* No special cleanup. */
```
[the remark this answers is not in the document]
uriyage: The cleanup is done in `ioThreadsOnUnlinkClient`

src/config.c
uriyage: Currently, module commands are not offloaded in any case. However, we also need to ensure that modules only access the keys declared  in their commands. If this is not guaranteed, we cannot offload commands for other slots. Once a single module fails to guarantee this behavior, we cannot offload any commands.

src/io_threads.c
uriyage: We expect admin commands to be infrequent, meaning not in the thousands per second. As long as 
this is the case, it won't affect performance, as we have observed. 
Exclusivity for admin commands simplifies the code and makes it more secure, as we don't need to worry about scenarios like server configurations being changed while a thread is executing a command, or client's
ACL permissions being modified during execution.

src/io_threads.c
uriyage: This will require some refactoring of the `call` function. Do you think we should include it in this PR, or should we create a separate follow-up task?

src/acl.c
```diff
@@ -2635,7 +2635,7 @@ void addACLLogEntry(client *c, int reason, int context, int argpos, sds username
     }
 
     /* if we have a real client from the network, use it (could be missing on module timers) */
-    client *realclient = server.current_client ? server.current_client : c;
+    client *realclient = getCurrentClient() ? getCurrentClient() : c;
```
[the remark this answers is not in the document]
uriyage: This is just a define

src/ae.h
uriyage: Done

src/ae.h
uriyage: done

src/commands/get.json
uriyage: fixed

src/server.c
uriyage: done

src/io_threads.c
uriyage: Done.

src/memory_prefetch.c
uriyage: Fixed, refactored to use one function in both cases.

src/networking.c
uriyage: Changed to `ioThreadCallCommand` moved to `server.c`

src/commands/get.json
uriyage: This was an initial list of the most common commands, we will expand it. Each additional command must be verified to ensure the code doesn't access any global variables that could be modified by other threads

src/io_threads.c
uriyage: IoToMT -> IO Thread to Main Thread.
This implementation is specifically tailored for the IO thread use case. Making it a more general solution would be complex and likely unnecessary. 
The naming and documentation can be improved

src/io_threads.c
uriyage: I will look into this. Additionally, we need to ensure that the multi-exec flow can be executed in parallel.

src/server.c
uriyage: Wouldn't it be a breaking change?

tests/unit/maxmemory.tcl
uriyage: The issue occurs when we evict a client while it's on the IO thread and that thread is already sending a reply so the client doesn't consume memory anymore. Currently, the tests use 5 MB. Do you suggest to increase the size to 100 MB to ensure the I/O threads cannot send the data fast enough?

src/server.c
uriyage: I wanted to ensure we don't accidentally shadow this global variable with a local variable.

tests/unit/io-threads.tcl
uriyage: Thanks, changed.

tests/unit/io-threads.tcl
uriyage: Done changed to 100 ms, I removed the `assert_equal "val" [r get key]`, and added a different key to test that case.

src/io_threads.c
uriyage: Done

src/kvstore.c
uriyage: Done

src/networking.c
uriyage: Done

src/networking.c
uriyage: Done

uriyage:
> Commands that can be offloaded to I/O threads are marked with the CAN_BE_OFFLOADED flag in their json files. However, different types of commands have different exclusivity requirements: > Database-exclusive commands require complete isolation and cannot run in parallel with any other offloaded command.
>> Can't this be done implicitly instead of defining it via a flag explicitly? With that we would be aware all read commands will be offloaded and a dev doesn't need to check if it's enabled. In the future, we also don't risk of missing new read command being not offloaded.

@hpatro
Each additional command must be verified to ensure the code doesn't access any global variables that could be modified by other threads. 
This is an initial list of the most common commands, we will expand it.

.github/workflows/ci.yml
```diff
@@ -80,6 +80,20 @@ jobs:
       - name: unit tests
         run: ./src/valkey-unit-tests
 
+  test-ubuntu-io-threads-sanitizer:
```
madolson, line 83: This should probably be in daily and not in the CI.

src/config.c
madolson: I don't really disagree with anything that you are saying, but I don't see how that conflicts with my comment that this shouldn't be a config. Administration shouldn't have to worry about if a module do work outside the context. The module should be declaring that it's safe to offload commands. If any module doesn't declare that, commands shouldn't be able to get offloaded.

src/commands/get.json
madolson: From an OSS perspective, we should probably do every command that's possible to offload. It's much harder to explain to end users when some commands are offloaded and some aren't. It's much easier to say X command (say MGET) for some reason doesn't work. Also isn't MGET quite popular? How did we decide common commands?

A more generic comment, how do we make sure these commands don't silently introduce multi-threaded issues later.

hpatro:
> > Commands that can be offloaded to I/O threads are marked with the CAN_BE_OFFLOADED flag in their json files. However, different types of commands have different exclusivity requirements: > Database-exclusive commands require complete isolation and cannot run in parallel with any other offloaded command.
> > > Can't this be done implicitly instead of defining it via a flag explicitly? With that we would be aware all read commands will be offloaded and a dev doesn't need to check if it's enabled. In the future, we also don't risk of missing new read command being not offloaded.
> 
> @hpatro Each additional command must be verified to ensure the code doesn't access any global variables that could be modified by other threads. This is an initial list of the most common commands, we will expand it.

Might become difficult to maintain in the future. I would like to keep the life of a developer implementing a new read command as easy as it is currently. Could we possibly introduce broad guard rails to avoid such mistakes? Thinking out loud here.

Could you share any command which is currently not isolated and what field do they access? For multi-commands, maybe we mark them with `MULTI_DB_ACCESS` and avoid this code flow.

uriyage:
> @uriyage Can you prioritize doing the merge so that we can properly run the tests?

@madolson, Done

uriyage:
> > Commands that can be offloaded to I/O threads are marked with the CAN_BE_OFFLOADED flag in their json files. However, different types of commands have different exclusivity requirements:
> > Database-exclusive commands require complete isolation and cannot run in parallel with any other offloaded command.
> 
> Can't this be done implicitly instead of defining it via a flag explicitly? With that we would be aware all read commands will be offloaded and a dev doesn't need to check if it's enabled. In the future, we also don't risk of missing new read command being not offloaded.

@hpatro, Thanks. I have reviewed all the missing read-only commands, and it can indeed be done with a few minor changes. 

I removed the flag, and now we offload all read-only commands with the following restrictions:

```c
#define CMD_CAN_BE_OFFLOADED(cmd) \
    ((cmd->flags & CMD_READONLY) && \
     !(cmd->flags & CMD_NO_MANDATORY_KEYS) && \
     !(cmd->flags & CMD_MAY_REPLICATE) && \
     !(cmd->flags & CMD_BLOCKING))
```
In addition, we won't offload commands if `c->slot` is not set (for example, in the KEYS command). This covers the case of multi-slot access.

src/server.c
madolson: By our definition I don't think so. Since the fields are still are present they just moved. We could discuss it though.

madolson:
> In addition, we won't offload commands if c->slot is not set (for example, in the KEYS command). This covers the case of multi-slot access.

It would still be good to understand how we can offload the SCAN command or at the very last have it not take a server level lock.

src/io_threads.c
madolson: > this is the case, it won't affect performance, as we have observed.

You haven't given any data on P99 performance, just throughput, so I don't think you have observed this. 

> Exclusivity for admin commands simplifies the code and makes it more secure, as we don't need to worry about scenarios like server configurations being changed while a thread is executing a command, or client's
ACL permissions being modified during execution.

I agree about changing ACL permissions, I was mentioning ACL log.

src/io_threads.c
madolson: > This implementation is specifically tailored for the IO thread use case. Making it a more general solution would be complex and likely unnecessary.

Make it generic then and split it out. If there are specific reasons it's highly tailored. 

> The naming and documentation can be improved

Yes, as of right now the naming is poor and inconsistent, please address.

src/io_threads.c
madolson: Yes, include it in this PR ideally.

src/server.c
madolson: then prepend it with global or put it in a struct, underscores aren't a convention.

src/io_threads.c
uriyage: Since config is an admin command, it will be executed with exclusivity

src/io_threads.c
```diff
@@ -353,17 +936,121 @@ void initIOThreads(void) {
...
+    if (server.active_io_threads_num == 1) {
+        return 0; /* No IO threads to offload to. */
+    }
+
+    if (!server.io_threads_do_commands_offloading) {
+        return 0; /* Command offloading is disabled. */
+    }
+
+    /* Check if modules are loaded and module offloading is disabled */
+    if (moduleCount() > 0 && !server.io_threads_do_commands_offloading_with_modules) {
```
jsoref, line 989: I'd imagine that a property access would be cheaper than a function call...
```suggestion
    if (!server.io_threads_do_commands_offloading_with_modules && moduleCount() > 0) {
```

src/kvstore.c
```diff
@@ -221,22 +222,38 @@ void kvstoreHashtableRehashingStarted(hashtable *ht) {
...
-    }
-
     size_t from, to;
     hashtableRehashingInfo(ht, &from, &to);
-    kvs->bucket_count -= from; /* Finished rehashing (Remove the old ht size) */
-    kvs->overhead_hashtable_rehashing -= from * HASHTABLE_BUCKET_SIZE;
+    rehashing_completion_ctx ctx = {.rehashing_node = metadata->rehashing_node, .kvs = metadata->kvs, .from = from};
+    metadata->rehashing_node = NULL;
+
+    /* If not in main-thread postpone the update of kvs rehashing info to be done later by the main-thread -*/
```
jsoref, line 251:
```suggestion
    /* If not in main-thread, postpone the update of kvs rehashing info to be done later by the main-thread */
```

src/networking.c
```diff
@@ -807,6 +796,51 @@ void afterErrorReply(client *c, const char *s, size_t len, int flags) {
...
+    if (c->flag.module) {
+        if (!c->deferred_reply_errors) {
+            c->deferred_reply_errors = listCreate();
+            listSetFreeMethod(c->deferred_reply_errors, sdsfreeVoid);
+        }
+        listAddNodeTail(c->deferred_reply_errors, sdsnewlen(s, len));
+        return;
+    }
+
+    /* Postpone error updates if its io-thread */
```
jsoref, line 836:
```suggestion
    /* Postpone error updates if it's io-thread */
```

src/server.h
```diff
@@ -918,9 +930,11 @@ typedef struct blockingState {
                                     which is opaque for the Redis core, only
                                     handled in module.c. */
 
-    void *async_rm_call_handle; /* ValkeyModuleAsyncRMCallPromise structure.
-                                   which is opaque for the Redis core, only
-                                   handled in module.c. */
+    void *async_rm_call_handle;   /* ValkeyModuleAsyncRMCallPromise structure.
+                                     which is opaque for the Redis core, only
+                                     handled in module.c. */
```
jsoref, line 935: I think style is:
```suggestion
    void *async_rm_call_handle;   /* ValkeyModuleAsyncRMCallPromise structure.
                                   * which is opaque for the Redis core, only
                                   * handled in module.c. */
```

src/server.h
```diff
@@ -1203,12 +1217,31 @@ typedef struct LastWrittenBuf {
 } LastWrittenBuf;
 
 typedef struct client {
+    volatile uint8_t io_read_state;               /* Indicate the IO read state of the client */
+    volatile uint8_t io_write_state;              /* Indicate the IO write state of the client */
+    volatile uint8_t io_command_state;            /* Indicate the IO command state of the client */
+    ustime_t duration;                            /* Current command duration. Used for measuring latency of blocking/non-blocking cmds */
+    robj **original_argv;                         /* Arguments of original command if arguments were rewritten. */
+    unsigned long long net_input_bytes_curr_cmd;  /* Total network input bytes read for the* execution of this client's current command. */
```
jsoref, line 1225:
```suggestion
    unsigned long long net_input_bytes_curr_cmd;  /* Total network input bytes read for the execution of this client's current command. */
```

src/server.h
```diff
@@ -1764,20 +1795,27 @@ struct valkeyServer {
...
-    long long stat_io_accept_offloaded;                /* Number of offloaded accepts */
-    long long stat_poll_processed_by_io_threads;       /* Total number of poll jobs processed by IO */
-    long long stat_total_reads_processed;              /* Total number of read events processed */
-    long long stat_total_writes_processed;             /* Total number of write events processed */
-    long long stat_client_qbuf_limit_disconnections;   /* Total number of clients reached query buf length limit */
-    long long stat_client_outbuf_limit_disconnections; /* Total number of clients reached output buf length limit */
-    long long stat_total_prefetch_entries;             /* Total number of prefetched dict entries */
-    long long stat_total_prefetch_batches;             /* Total number of prefetched batches */
+        stat_unexpected_error_replies;                  /* Number of unexpected (aof-loading, replica to primary, etc.) error replies */
+    long long stat_total_error_replies;                 /* Total number of issued error replies ( command + rejected errors ) */
```
jsoref, line 1799: Does this item now fit?
```suggestion
    long long stat_unexpected_error_replies;            /* Number of unexpected (aof-loading, replica to primary, etc.) error replies */
    long long stat_total_error_replies;                 /* Total number of issued error replies ( command + rejected errors ) */
```

yahorsi:
Guys it is SO big and SO game changing!
I beg you to release it in thr 9.1, please please please :)

yahorsi:
Just checking how is it going for the 10th release? )

Issue #2023, opened 2025-04-29
[NEW] Faster cluster failover


zuiderkwast:
In very fast networks, we don't need the hard-coded 500ms delay. Can we change these hard-coded numbers to be relative the configured node timeout?

This is to have less downtime during an automatic failover.

```C
        server.cluster->failover_auth_time = now +
                                             500 +           /* Fixed delay of 500 milliseconds, let FAIL msg propagate. */
                                             random() % 500; /* Random delay between 0 and 500 milliseconds. */
```

```C
        /* We add another delay that is proportional to the replica rank.
         * Specifically 1 second * rank. This way replicas that have a probably
         * less updated replication offset, are penalized. */
        server.cluster->failover_auth_time += server.cluster->failover_auth_rank * 1000;
```

@madolson @enjoy-binbin @hpatro Am I missing anything?

hpatro:
Slightly below the lines which you've referenced, we cleanup the failover_auth_time if it's a manual failover, so that the failover is performed immediately. 

https://github.com/valkey-io/valkey/blob/8df0a6b67587cb5505d628d76e5c83532f4a4c39/src/cluster_legacy.c#L4962-L4969

I guess we can flip the logic and not compute any delay if it's for manual failover to avoid the above confusion.

zuiderkwast:
Oh, my bad. :)

I got this from someone else and I think I got it wrong. Instead, for *automatic* failover, can we change these hard-coded numbers? In a very fast network, we don't need the fixed 500 ms.

Can we make these be calculated from the node timeout or something instead? Lower node timeout can result in lower failover delay?

hpatro:
> Can we make these be calculated from the node timeout or something instead? Lower node timeout can result in lower failover delay?

Yeah, better than hardcoded values.

enjoy-binbin:
Sorry i missed the thread. And yes, i think we indeed can adjust the number.

In #1018, i use the same 500 number to add the delay by default to avoid the conflict, i forgot if i ever mention that to see if we can set the 500 as a configuration item. Do you have a good way to do the calculate by node timeout? Or do you think we can make it become a configuration item (so that the admin can control this logic to avoid the conflict and make it more faster)?

zuiderkwast:
> Do you have a good way to do the calculate by node timeout?

Just an idea: node_timeout / 30. If gives us 500 for the default node timeout 15000.

For node timeout 3000, we get 100 ms. Seems ok?

[...]

PR #2209, opened 2025-06-12
Start failover without waiting for the next cluster cron cycle

enjoy-binbin:
When myself is a replica, if we receive a message that my primary is
FAIL, we can try to set CLUSTER_TODO_HANDLE_FAILOVER so that we can
try to failover as soon as possible in beforeSleep without waiting
for clusterCron to kick in.

Add a new markNodeAsFailing method, we move FAIL flag related code to
here so we can easily check whether the failing node is my primary.

## The following is all old content. It was deleted during the discussion, but the content is worth keeping. (Don't merge this text)
================================================================================

Remove the 500ms fixed delay for automatic failover for faster failover

An automatic failover has a fixed delay of 500ms, its main purpose is
to wait for the FAIL packet to propagate in the cluster so that other
primaries can respond to the vote request.
```
server.cluster->failover_auth_time =
now +
500 +           /* Fixed delay of 500 milliseconds, let FAIL msg propagate. */
random() % 500; /* Random delay between 0 and 500 milliseconds. */
```

A fixed delay of 500ms is a bit long in now days, we are looking into
removing it for a faster failover. If we can ensure it is safe to remove,
then we can remove it.

Currently a replica now can only learn of the death of its primary
node from two places:
1. Collecting pfail messages from other primary nodes, and then check
in markNodeAsFailingIfNeeded to see if we have the quorum.
2. Receiving FAIL messages from other primary nodes (or replica nodes).

Anyway both the failing state is triggered collecting failure reports
from primaries.

In markNodeAsFailingIfNeeded, if `nodeIsReplica(myself) && myself->replicaof == node`
is true, then we can try to trigger a failover in clusterBeforeSleep as soon as
possible without waiting for clusterCron. Moreover, we can remove 500ms, because
we broadcast a FAIL here, which means that myself's FAILOVER_AUTH will definitely
be after the FAIL, which means we don't have to wait 500ms for FAIL to propagate.
```
void markNodeAsFailingIfNeeded(clusterNode *node) {
    ...
    /* Broadcast the failing node name to everybody, forcing all the other
     * reachable nodes to flag the node as FAIL.
     * We do that even if this node is a replica and not a primary: anyway
     * the failing state is triggered collecting failure reports from primaries,
     * so here the replica is only helping propagating this status. */
    clusterSendFail(node->name);
```

And moreover, if we do the same when getting a FAIL, we check if failing is
myself's primary, and we call clusterSendFail to send the FAIL, this look like
we can also remove the 500ms delay. It is somehow like the first case in a very
badly way, like all the replicas got the pfail and then send the fail.
```
    } else if (type == CLUSTERMSG_TYPE_FAIL) {
        clusterNode *failing;

        if (sender) {
            failing = clusterLookupNode(hdr->data.fail.about.nodename, CLUSTER_NAMELEN);
            if (failing && !(failing->flags & (CLUSTER_NODE_FAIL | CLUSTER_NODE_MYSELF))) {
                serverLog(LL_NOTICE, "FAIL message received from %.40s (%s) about %.40s (%s)", hdr->sender,
                          sender->human_nodename, hdr->data.fail.about.nodename, failing->human_nodename);
                failing->flags |= CLUSTER_NODE_FAIL;
                failing->fail_time = now;
                failing->flags &= ~CLUSTER_NODE_PFAIL;
                clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_UPDATE_STATE);
            }
```

So in this commit, we try to remove the fixed delay of 500ms. For now, we
keep the random 500ms and use the ranking to avoid the vote confilct.

Related to #2023.


src/cluster_legacy.c
```diff
@@ -4984,6 +5004,11 @@ void clusterHandleReplicaFailover(void) {
             /* Reset auth_age since it is outdated now and we can bypass the auth_timeout
              * check in the next state and start the election ASAP. */
             auth_age = 0;
+        } else if (server.cluster->failover_auth_time == now) {
+            /* If we happen to initiate a failover immediately, reset auth_age since it is
+             * outdated now and we can bypass the auth_timeout check in the next state and
+             * start the election ASAP. */
+            auth_age = 0;
```
zuiderkwast: This can't happen (almost) as long as we keep the `random() % 500` delay.

zuiderkwast (commented):
In the time between scheduling a failover and starting the vote, the replica sends it's offset to other replicas in the same shard, to make sure the nodes agree on the rank before the failover starts.

```c
        /* Now that we have a scheduled election, broadcast our offset
         * to all the other replicas so that they'll updated their offsets
         * if our offset is better. */
        clusterBroadcastPong(CLUSTER_BROADCAST_LOCAL_REPLICAS);
```

Assuming the other replicas will also receive the FAIL, they will also broadcast their offset to other replicas in the same shard, so all replicas will have updated offset before any replica actually starts the vote.

This code updates the rank and the delay if some replica reports a better offset: 

```c
    /* It is possible that we received more updated offsets from other
     * replicas for the same primary since we computed our election delay.
     * Update the delay if our rank changed.
     *
     * It is also possible that we received the message that telling a
     * shard is up. Update the delay if our failed_primary_rank changed.
     *
     * Not performed if this is a manual failover. */
    if (server.cluster->failover_auth_sent == 0 && server.cluster->mf_end == 0) {
        int newrank = clusterGetReplicaRank();
        if (newrank != server.cluster->failover_auth_rank) {
            long long added_delay = (newrank - server.cluster->failover_auth_rank) * 1000;
            server.cluster->failover_auth_time += added_delay;
            server.cluster->failover_auth_rank = newrank;
            serverLog(LL_NOTICE, "Replica rank updated to #%d, added %lld milliseconds of delay.", newrank,
                      added_delay);
        }
```

With this PR, the delay can be near 0, so there is not enough time for the replicas to agree about the rank. In the worst case, two replicas both believe they are rank 0 and start the failover at the same time, and none of them will win the election, causing even more downtime before we have a successful failover.

Do we need a fixed delay for this case, or is there some other mechanism to avoid this race between replicas?

madolson:
> Do we need a fixed delay for this case, or is there some other mechanism to avoid this race between replicas?

We will always race without a consensus algorithm like raft or paxos with the data flowing through it. It's just what degree of raciness are we willing to tolerate.

We could wait for an affirmative ping from the other replicas that they believe their primary is failed (it will be in their gossip pfail) before starting to nominate ourselves if we have the highest offset or 500ms has elapsed.

hpatro (commented):
My assumption from https://github.com/valkey-io/valkey/issues/2023 we wanted to make this delay dynamic w.r.t. to `cluster-node-timeout`. I'm apprehensive to having a separate flow to handle faster failover.

src/cluster_legacy.c
enjoy-binbin: I did see it once, 0ms delay, on the daily CI test yesterday. So i made the change..
```
        server.cluster->failover_auth_time = now +
                                             random() % 500; /* Random delay between 0 and 500 milliseconds. */
```

enjoy-binbin:
> My assumption from https://github.com/valkey-io/valkey/issues/2023 we wanted to make this delay dynamic w.r.t. to cluster-node-timeout. I'm apprehensive to having a separate flow to handle faster failover.

yes, Viktor and i talked about this offline yesterday, i think we can do it with other fixed number, like the `random() % 500`, `server.cluster->failover_auth_rank * 1000`, `server.cluster->failover_failed_primary_rank * 500` something like these.

And with this fixed 500ms delay, i wonder if the current mechanism will be safe enough, and if we determine that it is indeed safe, then i would love to remove it.

src/cluster_legacy.c
enjoy-binbin: here is the link if you like to see https://github.com/enjoy-binbin/valkey/actions/runs/15610070529/job/43968663126#step:8:5087
```
20313:S 12 Jun 2025 12:31:35.055 * FAIL message received from affe2080a37ca98c0a0afd3d30038b66d9c829ca () about 1c04ae1561d47d097e75291c8aa06dde6aeeae4e ()
20313:S 12 Jun 2025 12:31:35.055 * Start of election delayed for 0 milliseconds (rank #0, primary rank #0, offset 2172).
20313:S 12 Jun 2025 12:31:35.055 * Currently unable to failover: Failover attempt expired.
20313:S 12 Jun 2025 12:31:35.055 * Needed quorum: 2. Number of votes received so far: 0
20313:S 12 Jun 2025 12:31:35.055 # Cluster state changed: fail
20313:S 12 Jun 2025 12:31:35.055 # Cluster is currently down: At least one hash slot is not served by any available node. Please check the 'cluster-require-full-coverage' configuration.
```

zuiderkwast: Sure, there is 1/500 chance that it will happen. :)

zuiderkwast:
> We will always race without a consensus algorithm like raft or paxos with the data flowing through it. It's just what degree of raciness are we willing to tolerate.

@madolson Sure, but we try to make the race less likely to happen. We have rank and now also primary rank that helps avoid a race if two primaries fail at the same time.

> We wait for an affirmative ping from the other replicas that they believe their primary is failed (it will be in their gossip pfail) before starting to nominate ourselves if we have the highest offset or 500ms has elapsed.

Do we do this? Whenever we get a pong from all replicas while waiting, we abort the 500ms wait and start the election immediately? I haven't seen this. Can you point to the code?

src/cluster_legacy.c
```diff
@@ -3777,6 +3789,15 @@ int clusterProcessPacket(clusterLink *link) {
...
                 clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_UPDATE_STATE);
+
+                if (nodeIsReplica(myself) && myself->replicaof == failing) {
+                    /* Immediately check if the node is our primary node, if so, we can try
+                     * to initiate a failover as soon as possible. We will also try to send
+                     * the FAIL packet in case we trigger the failover immediately but unable
+                     * to get the votes. */
+                    clusterSendFail(failing->name);
+                    clusterDoBeforeSleep(CLUSTER_TODO_HANDLE_FASTER_FAILOVER);
+                }
```
zuiderkwast: If we broadcast the FAIL such as here, we can be sure that all primaries will mark it as FAIL so they will not reject the vote because of this.

But there is no guarantee our replication offset is the best among replicas. Some other replica may have a better replication offset and we will find out only if we want some milliseconds before we start the vote.

src/cluster_legacy.c
```diff
@@ -4963,7 +4984,6 @@ void clusterHandleReplicaFailover(void) {
      * elapsed, we can setup a new one. */
     if (auth_age > auth_retry_time) {
         server.cluster->failover_auth_time = now +
-                                             500 +           /* Fixed delay of 500 milliseconds, let FAIL msg propagate. */
                                              random() % 500; /* Random delay between 0 and 500 milliseconds. */
```
zuiderkwast: So it seem that the main point of the fixed delay is not to let FAIL propagate, but to let the replicas converge on rank. The code comment is misleading.

The other (safer) idea that doesn't change any logic is to just make these 500 be relative to the node time, as discussed in #2023. The delay we need depends on speed, just like node timeout.

If we replace 500 with `node_timeout / 30`, we keep 500 for the default node timeout 15000 but use shorter delay for if node timeout is smaller.

| node timeout | node_timeout / 30 |
|--------------|-------------------|
| 15000        | 500               |
| 6000         | 200               |
| 3000         | 100               |
| 2000         | 66                |
| 1500         | 50                |

For node timeout 2000, we have observed the total time from killing a node until successful failover is usually 3.0-3.5 seconds. With this shorter delay, we can reduce it to 2.5-2.6 seconds and save almost one second for the highest random() value.

```c
        int delay = min(node_timeout / 30, 500);
        server.cluster->failover_auth_time = now +
                                             delay +           /* Fixed delay to converge on rank. */
                                             random() % delay; /* Random delay between 0 and delay. */
```

src/cluster_legacy.c
enjoy-binbin: i don't think that help much... If a primary is down, that is, we were passed the node-timeout, so in this window, these replica will also exchange ping-pong since we have this logic
```
        /* If we have currently no active ping in this instance, and the
         * received PONG is older than half the cluster timeout, send
         * a new ping now, to ensure all the nodes are pinged without
         * a too big delay. */
        mstime_t ping_interval =
            server.cluster_ping_interval ? server.cluster_ping_interval : server.cluster_node_timeout / 2;
```

i am not sure if i remember it correctly

src/cluster_legacy.c
enjoy-binbin: 35820549825791a8f188c2870a9069d36edd019d this commit also stand for it in some ways.

hwware:
My understanding for this requirement is to change current 500 ms delay (the fixed value) to a config parameter and let user decide the value (the default value keeps as 500 ms).  Added in Valkey weekly meeting.

madolson:
> Do we do this? Whenever we get a pong from all replicas while waiting, we abort the 500ms wait and start the election immediately? I haven't seen this. Can you point to the code?

No, I was proposing that was something we could do. I think I missed a word in my original comment.

We are waiting from the fail message from the other replicas to make sure we have their repl offset to compute the rank. If a given replica has received that information from all other replicas, it doesn't need to wait.

src/cluster_legacy.c
madolson: We updated pong_receive, which is what we check against ping_interval, with gossip, so we may have not recently received a message. We only send a message for sure when we mark the primary as failed, since we send out the fail message.

enjoy-binbin: I still think we can exchange one ping-pong during the window. Can you give an example? I somehow fail to see it (i am bad at this)

madolson:
We discussed this in the weekly meeting. The discussion was that we don't want to fully remove the 500ms logic, since we believe we do need to wait to get the messages from the other replicas to verify we have the best replication offset. We need some verification that we have the offsets from the other replicas before starting the election.

enjoy-binbin:
btw, if we only have one replica, that means, my rank is definitely is 0, and my offset is definitely the higher one. The waitting is just a waste in this case.

enjoy-binbin:
Assuming the comments are correct and we are waiting for FAIL to propagate. Or we consider it safe, then I would like to remove it completely.
```
        server.cluster->failover_auth_time = now +
                                             500 +           /* Fixed delay of 500 milliseconds, let FAIL msg propagate. */
                                             random() % 500; /* Random delay between 0 and 500 milliseconds. */
```

Or there are other options:
1. Make it become a configuration item, an experienced administrator can use any value he wants, zero or 500ms, one can set it based on network (if you have a small node timeout, you can change it) or cluster size (if you only have one replica, you can just go zero) etc. This work for everyone, except the disadvantage is that a new configuration item needs to be introduced, which is not easy for ordinary people to use, but to some extent many of our configuration items rely on users to understand some details.
2. zuiderkwast idea, doing some math like node_timeout / 30, i am fine with that, but I should mention that we don't modify this timeout value often, so it is unlikely to be used for us.

Anyway, bottom line, if we feel safe, then I want to remove it. If we are not sure, that's fine, after all, stability is also very important.

zuiderkwast:
@enjoy-binbin we can do Madelyn's idea. We wait but when we receive information from all replicas that they know the primary is failing and we get their offset, we start immediately. It can be very fast.

enjoy-binbin:
ok, that is a good idea and i think we are making a progress step by step, it can definitely solve the one replica case and help other cases, i will do it in a new PR.

new PR: #2227

enjoy-binbin:
@zuiderkwast I revert the 500ms change, and only keep the TODO FAILOVER logic. (I guess we can ignore the DCO for now, i can squash a signoff when i merge it)

src/cluster_legacy.c
```diff
@@ -5578,6 +5589,11 @@ void clusterBeforeSleep(void) {
      * called for flags set should be able to clear its flag). */
     server.cluster->todo_before_sleep = 0;
 
+    /* Update the cluster state. We handle this flag first so that if we happen
+     * to also has have failover flag, we can check the state first (and log the
```
```suggestion
     * to also have a failover flag, we can check the state first (and log the
```

zuiderkwast (approved):
Change LGTM. Now it's only to skip the wait for the next clusterCron cycle.

I think you can change the title from "as soon as possible" to for example something like "Start failover without waiting for the next cluster cron cycle".

hpatro:
@enjoy-binbin @zuiderkwast Could you explain the rational behind this change better? 

I don't understand the benefit of this very clearly. Let me share my reasoning on this. I believe on a single primary failure scenario with just one replica this change could potentially help with faster failover, however that is also not guaranteed if the healthy primaries are not yet aware about the failure of the primary for which vote is requested.

Further, if there are multiple replicas in a given shard, it's very likely they would conflict with each other and divide the votes amongst them, which isn't ideal. It's also possible that a replica with lower offset could be elected.

And in multiple primary failure scenario, we would again divide the votes for a given epoch amongst different shard causing wastage of election cycle and introduce delay.

I would also like the @valkey-io/core-team to share their opinion on this.

zuiderkwast:
@hpatro This PR is now very minor. It doesn't remove the 500 delay nor the random delay. It only removed the wait for next cluster cron cycle, which I believe is not an intentional delay. 

Maybe you want to discuss this one instead?

* #2227

PR #2213, opened 2025-06-13
Optimize bitcount by using AVX-512 intrinsic

shanwan1:
Description:
This PR introduces an optimized method for bitcount command in Valkey by using Intel AVX-512 intrinsic. The new implementation increases throughput by up to 27% and reduces latency up to 21% compared with the existing implementation by using AVX2.

Change:
1.Add a new function popcountAVX512 to perform the optimized method for bitcount by using Intel AVX-512 intrinsic.
2.Create one macro valkey_attribute_target to replace multiple macros ATTRIBUTE_TARGET_AVX*. Replace ATTRIBUTE_TARGET_AVX* with valkey_attribute_target wherever it's used.
3.Add runtime check for cpu support of AVX2.
4.Add test cases to validate the correctness of the implementation of popcountAVX512.

Performance boost:
The throughput of the benchmark can be increased up to 27% compared with AVX2, and the latency can be reduced up to 21%.
| Benchmark | Ops/sec | Latency |
|----------|----------|----------|
| [memtier_benchmark-1key-1Billion-bits-bitmap-bitcount](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-1Billion-bits-bitmap-bitcount.yml) | 27% | -21% |
| [memtier_benchmark-1key-100M-bits-bitmap-bitcount](https://github.com/redis/redis-benchmarks-specification/blob/main/redis_benchmarks_specification/test-suites/memtier_benchmark-1key-100M-bits-bitmap-bitcount.yml) | 7% | -6% |

Test env:
CPU: Intel(R) Xeon(R) Platinum 8380 CPU @ 2.30GHz, 2 sockets, 40 cores per socket, 1 thread per core
OS: Ubuntu 24.04 LTS
Memory: 256GB
Server config:
`taskset -c 0-39 valkey-server --port 6379 --dir logs --logfile server.log --save "" --daemonize yes`


src/util.c
```diff
@@ -459,8 +459,8 @@ int ull2string(char *dst, size_t dstlen, unsigned long long value) {
  * 6. Extract the final value.
  * Output: *value = 1234567890
  */
-ATTRIBUTE_TARGET_AVX512
-static int string2llAVX512(const char *s, size_t slen, long long *value) {
+valkey_attribute_target("avx512f,avx512bw,avx512vl")
+static int string2llAVX512(const char *s, unsigned long slen, long long *value) {
```
r-devulap: why change `size_t` to `unsigned long`?

src/unit/test_bitops.c
```diff
@@ -40,8 +41,14 @@ static int test_case(const char *msg, int size) {
         long long ret_scalar = popcountScalar(buf, size);
         TEST_ASSERT_MESSAGE(msg, expect == ret_scalar);
 #if HAVE_X86_SIMD
-        long long ret_avx2 = popcountAVX2(buf, size);
-        TEST_ASSERT_MESSAGE(msg, expect == ret_avx2);
+        __builtin_cpu_init();
+        if (__builtin_cpu_supports("avx512f") && __builtin_cpu_supports("avx512bw") && __builtin_cpu_supports("avx512vpopcntdq")) {
```
r-devulap, line 47: I'm wondering if we could simplify this approach a bit. CPUID checks might be better handled outside of test code. Perhaps we could leverage the higher level `popcount` function instead? This would help abstract away the CPUID complexity and keep the test simple.

src/unit/test_bitops.c
```diff
@@ -76,6 +83,10 @@ int test_popcount(int argc, char **argv, int flags) {
     TEST_CASE("Popcount(Part A + Part C)", 8 * 32 * 11 + 7);
     TEST_CASE("Popcount(Part A + Part B + Part C)", 8 * 32 * 3 + 3 * 32 + 5);
     TEST_CASE("Popcount(Corner case)", 0);
+
+    TEST_CASE("Popcountavx512(Part A)", 64 * 2);
```
r-devulap: Rename to `Popcount`? This is not avx512 specific, right? Also, perhaps add a test case when data size is an odd number?

r-devulap (commented):
Minor comments, but otherwise LGTM.

src/util.c
shanwan1: Might be typo, I change it to size_t.

src/unit/test_bitops.c
shanwan1: I add the test of the higher level function serverPopcount.

src/unit/test_bitops.c
shanwan1: I update the name and add the test case with an odd data size.

src/bitops.c
```diff
@@ -240,9 +264,12 @@ long long popcountNEON(void *s, long n) {
  * work with an input string length up to 512 MB or more (server.proto_max_bulk_len) */
 long long serverPopcount(void *s, long count) {
 #if HAVE_X86_SIMD
-    /* If length of s >= 256 bits and the CPU supports AVX2,
-     * we prefer to use the SIMD version */
-    if (count >= 32) {
+    __builtin_cpu_init();
```
zhulipeng: Consider removing `__builtin_cpu_init()` if we don't use the ifunc resolvers or constructors?

src/bitops.c
```diff
@@ -240,9 +264,12 @@ long long popcountNEON(void *s, long n) {
...
+    if (__builtin_cpu_supports("avx512f") && __builtin_cpu_supports("avx512bw") && __builtin_cpu_supports("avx512vpopcntdq")) {
+        return popcountAVX512(s, count);
+    }
+    
+    if (__builtin_cpu_supports("avx2")) {
```
zhulipeng: How about `else if (__builtin_cpu_supports("avx2")) {` ?

src/bitops.c
```diff
@@ -52,9 +52,33 @@ static const unsigned char bitsinbyte[256] = {
...
 
 #if HAVE_X86_SIMD
+valkey_attribute_target("avx512f,avx512bw,avx512vpopcntdq")
+long long popcountAVX512(void *s, long count) {
+    const size_t chunks = count / 64;
+    uint8_t *ptr = (uint8_t *)s;
+    const uint8_t *end = ptr + count;
+
+    __m512i accumulator = _mm512_setzero_si512();
+    for (size_t i = 0; i < chunks; i++, ptr += 64) {
```
zhulipeng: Is it possible to use a constant to define the magic number `64`, like `CHUNK_SIZE = 64`?

zhulipeng (commented):
The code LGTM, just a few minor comments.

src/bitops.c
r-devulap: +1

shanwan1: I remove the init function.

src/bitops.c
shanwan1: I change it to else if.

src/bitops.c
shanwan1: I define AVX512_CHUNK_SIZE 64.

zhulipeng (commented):
@shanwan1 Please resolve conflicts.

zhulipeng (approved):
@zuiderkwast @chzhoo PTAL.

src/bitops.c
zuiderkwast: Will you remove it? I still see it here.

src/bitops.c
```diff
@@ -240,9 +265,10 @@ long long popcountNEON(void *s, long n) {
...
 long long serverPopcount(void *s, long count) {
 #if HAVE_X86_SIMD
-    /* If length of s >= 256 bits and the CPU supports AVX2,
-     * we prefer to use the SIMD version */
-    if (count >= 32) {
+    __builtin_cpu_init();
+    if (__builtin_cpu_supports("avx512f") && __builtin_cpu_supports("avx512bw") && __builtin_cpu_supports("avx512vpopcntdq")) {
+        return popcountAVX512(s, count);
+    } else if (__builtin_cpu_supports("avx2")) {
         return popcountAVX2(s, count);
```
zuiderkwast, line 269: Was the check for `__builtin_cpu_supports("avx2")` missing here before this change? It means we try to run AVX2 code even if the CPU doesn't support it, which is only the oldest x86-64 processors like Pentium and Celeron according to https://en.wikipedia.org/wiki/Advanced_Vector_Extensions#CPUs_with_AVX2. It means this is a bugfix?

Please mention it in the PR top comment as an additional bugfix for x86-64 CPUs without AVX2.

(FYI: Usually we want bugfixes and new functionality in separate PRs, so we can backport bugfixes to patch versions of old releases but new features and optimizations are only released in the next release. However, in this case I guess we can keep it in the same PR. Those CPUs are very old. Maybe nobody is using them and nobody has reported the bug so far.)

src/unit/test_bitops.c
```diff
@@ -37,16 +31,8 @@ static int test_case(const char *msg, int size) {
...
-#if HAVE_X86_SIMD
-        long long ret_avx2 = popcountAVX2(buf, size);
-        TEST_ASSERT_MESSAGE(msg, expect == ret_avx2);
-#endif
-#if defined(__aarch64__)
-        long long ret_neon = popcountNEON(buf, size);
-        TEST_ASSERT_MESSAGE(msg, expect == ret_neon);
-#endif
+        long long ret = serverPopcount(buf, size);
+        TEST_ASSERT_MESSAGE(msg, expect == ret);
```
zuiderkwast: Let's keep the explicit test for each implementation.

It is more explicit and it adds coverage for the scalar and AVX2 implementations even when we run the test on on a machine that supports AVX512.

We should add a check `__builtin_cpu_supports` to run the avx2 and avx512 implementations though.

```c
        long long ret = serverPopcount(buf, size);
        TEST_ASSERT_MESSAGE(msg, expect == ret);

        /* Explicitly test special implementations. */
#if HAVE_X86_SIMD
        if (__builtin_cpu_supports("avx2")) {
            long long ret_avx2 = popcountAVX2(buf, size);
            TEST_ASSERT_MESSAGE(msg, expect == ret_avx2);
        }
        if (__builtin_cpu_supports("avx512f") && __builtin_cpu_supports("avx512bw") && __builtin_cpu_supports("avx512vpopcntdq")) {
            long long ret_avx512 = popcountAVX512(buf, size);
            TEST_ASSERT_MESSAGE(msg, expect == ret_avx512);
        }
#endif
#if defined(__aarch64__)
        /* All 64-bit ARM CPUs support NEON. */
        long long ret_neon = popcountNEON(buf, size);
        TEST_ASSERT_MESSAGE(msg, expect == ret_neon);
#endif

        /* Test the fallback implementation used on CPUs that can't use
         * any of the specialized implementations. */
        long long ret_scalar = popcountScalar(buf, size);
        TEST_ASSERT_MESSAGE(msg, expect == ret_scalar);

```

src/bitops.c
```diff
@@ -52,9 +52,34 @@ static const unsigned char bitsinbyte[256] = {
...
+    for (size_t i = 0; i < chunks; i++, ptr += AVX512_CHUNK_SIZE) {
+        const __m512i v = _mm512_loadu_si512((const __m512i *)ptr);
+        const __m512i p = _mm512_popcnt_epi64(v);
+        accumulator = _mm512_add_epi64(accumulator, p);
+    }
+
+    if (ptr < end) {
+        const size_t remaining = end - ptr;
+        __mmask64 mask = (1ULL << remaining) - 1;
+        const __m512i v = _mm512_maskz_loadu_epi8(mask, ptr);
```
zuiderkwast, line 71: This instruction doesn't access the memory where the mask is all zero? Just want to be sure it doesn't read memory outside the allocation.

zuiderkwast (commented):
Great! I have some minor comments and questions.

Just curious, is there a real use case for bitmaps with 1 billion bits?

zuiderkwast:
Clang Format has some comments too: https://github.com/valkey-io/valkey/actions/runs/15738618329/job/44391329891?pr=2213

src/bitops.c
chzhoo: Based on the information below, I think that this way of writing should be correct.

- The pg_popcount_avx512 function of postgresql is implemented in a similar way: https://doxygen.postgresql.org/pg__popcount__avx512_8c_source.html
- Description of _mm512_maskz_loadu_epi8:  https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm512_maskz_loadu_epi8&ig_expand=4084

zuiderkwast: Thanks. Yes, the pseudo-code seems to say that memory is not accessed for bytes where no bit is set.

```
FOR j := 0 to 63
	i := j*8
	IF k[j]
		dst[i+7:i] := MEM[mem_addr+i+7:mem_addr+i]
	ELSE
		dst[i+7:i] := 0
	FI
ENDFOR
dst[MAX:512] := 0
```

src/bitops.c
shanwan1: It should be removed now.

src/bitops.c
shanwan1: I add a comment of runtime check for cpu support of AVX2.

src/unit/test_bitops.c
shanwan1: I keep the explicit tests for each implementation of bitcount.

src/unit/test_bitops.c
```diff
@@ -6,13 +6,12 @@
...
 
+extern long long serverPopcount(void *s, long count);
 extern long long popcountScalar(void *s, long count);
 #if HAVE_X86_SIMD
+extern long long popcountAVX512(void *s, long count);
 extern long long popcountAVX2(void *s, long count);
 #endif
-#if defined(__aarch64__)
-extern long long popcountNEON(void *s, long count);
-#endif
```
zuiderkwast: There is a test for popcountNEON guarded by a similar ifdef. If we remove the prototype, does it compile on ARM?

I don't know why the CI tests didn't run on this PR.

zuiderkwast: I found the CI did run on your fork. The Mac build failed when this prototype was removed:

```
unit/test_bitops.c:55:30: error: call to undeclared function 'popcountNEON'; ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
        long long ret_neon = popcountNEON(buf, size);
                             ^
```

https://github.com/shanwan1/valkey/actions/runs/15786879582/job/44505112668#step:3:242

src/unit/test_bitops.c
```diff
@@ -6,13 +6,12 @@
...
```
zuiderkwast: Adding this back.

```suggestion
#if defined(__aarch64__)
extern long long popcountNEON(void *s, long count);
#endif

```

PR #2215, opened 2025-06-15
Remove deny_blocking check in moduleBlockClient, cleanup code and doc

enjoy-binbin:
This flag.deny_blocking check causes some code to becomde dead code.
Some of the code below checks islua or ismulti, but they are marked
with flag.deny_blocking and will return early.

In addition, cleanup some documents, some of them are inaccurate,
and restore the code of blocking_auth_cb in tests/modules/auth.c,
this reply should be returned from core. The reply used to from the
core and was changed to from the module in #1819, and now it is from
the core again.

Cleanup some dead code around #1819.


enjoy-binbin:
@zuiderkwast do you want to review this? You have reviewed the related PR and i have left some comment there. I seem to have forgotten why i was looking this issue, but i can recall there was something odd about it.

src/module.c
```diff
@@ -7876,7 +7875,7 @@ ValkeyModuleBlockedClient *moduleBlockClient(ValkeyModuleCtx *ctx,
 
     if (islua || ismulti) {
         c->bstate->module_blocked_handle = NULL;
-        addReplyError(c, islua ? "Blocking module command called from Lua script"
+        addReplyError(c, islua ? "Blocking module command called from lua script"
```
zuiderkwast: Why do you change Lua to lowercase? Lua is a name and it should normally be first letter uppercase. https://en.wikipedia.org/wiki/Lua

src/module.c
```diff
@@ -7815,9 +7815,8 @@ ValkeyModuleBlockedClient *moduleBlockClient(ValkeyModuleCtx *ctx,
                                              void *privdata,
                                              int flags) {
     client *c = ctx->client;
-    if (c->flag.blocked || getClientType(c) != CLIENT_TYPE_NORMAL || c->flag.deny_blocking) {
-        /* Early return if duplicate block attempt or client is not normal or
-         * client is set to deny blocking. */
+    if (c->flag.blocked || getClientType(c) != CLIENT_TYPE_NORMAL) {
```
zuiderkwast, line 7842: I don't understand why we can delete deny_blocking here. Is the client type always != NORMAL for deny_blocking clients?

src/module.c
```diff
@@ -8170,7 +8167,16 @@ ValkeyModuleBlockedClient *VM_BlockClient(ValkeyModuleCtx *ctx,
 /* Block the current client for module authentication in the background. If module auth is not in
  * progress on the client, the API returns NULL. Otherwise, the client is blocked and the VM_BlockedClient
  * is returned similar to the VM_BlockClient API.
- * Note: Only use this API from the context of a module auth callback. */
+ * Note: Only use this API from the context of a module auth callback.
+ *
+ * There are some cases where ValkeyModule_BlockClientOnAuth() cannot be used:
+ *
+ * 1. If the client is in the middle of module based authentication. This will not block the client
+ *    but instead produce a specific error reply.
```
rjd15372: Isn't the opposite? 

An error will be returned if the client is **not** in the middle of a module based authentication process.

src/module.c
rjd15372: I think @enjoy-binbin is assuming that whenever we get to this code with `deny_blocking == 1` it means we are either running in the context of a script or a MULTI command. And in that case, we only want to prevent blocking the client if `is_keyspace_notification` is true

enjoy-binbin: yes, if we keep this deny_blocking condition here, there will be some dead code like this one. If we are from lua or multi, we always set deny_blocking so this code become a dead code. In this case, i think we should return the correct error (or at least at that time i think, i did kind of forget something around it.)
```
    if (islua || ismulti) {
        c->bstate->module_blocked_handle = NULL;
        addReplyError(c, islua ? "Blocking module command called from Lua script"
```

src/module.c
enjoy-binbin: i don't remember... Probably just wanted to keep it lowercase like the transaction below. I'll change it back.

src/module.c
enjoy-binbin: yes, you are right, thanks.

src/module.c
rjd15372: Can we add an assert before this line to ensure the assumptions are correct?

```c
serverAssert(!c->flag.deny_blocking || (islua || ismulti));
```

rjd15372 (approved):
I'm approving the PR but please add the assert that I mentioned in the code before merging the PR.

src/module.c
enjoy-binbin: Thanks! Added

PR #2216, opened 2025-06-16
Fix unixsocket too long cause addr / laddr being truncated in CLIENT INFO / LIST and logging

enjoy-binbin:
When displaying client addr / laddr using catClientInfoString for example,
if the client is connected via unix domain socket and the unixsocket exceeds
NET_ADDR_STR_LEN, the output will be truncated because NET_ADDR_STR_LEN is
not long enough.

Currently NET_ADDR_STR_LEN is 46+32 that is 78, and the maximum length of a
UNIX domain socket path is typically 108 bytes on Linux and 104 bytes on macOS,
both number including null terminator. In this fix, we use CONN_ADDR_STR_LEN
instead which is 128, that should be long enough for most cases.

It affects the output of CLIENT INFO / CLIENT LIST commands, as well as the
printing of some logs.

Other cleanup:
1. In acceptCommonHandler, when calling connFormatAddr on a unix socket
client, the connFormatAddr function will always return /unixsocket since
in anetFdToString we hardcoded the string. We changed it to the same display.

2. We changed the return value of connSocketAddr from returning C_OK/C_ERR to
returning 0 and -1, which is consistent with the return values of other types.

One change worth mentioning is that in connSocketAddr, we used to call
anetFdToString which would return `/unixsocket` hardcoded in this case.
Now we will use server.unixsocket, the format is `path:0`.

### DIFF
--- a/src/connection.h
+++ b/src/connection.h
@@ -40,11 +40,10 @@
 #include "ae.h"
 
 #define CONN_INFO_LEN 32
-#define CONN_ADDR_STR_LEN 128 /* Similar to NET_ADDR_STR_LEN, hoping to handle other protocols. */
+#define CONN_ADDR_STR_LEN 128
 
 #define NET_HOST_STR_LEN 256                          /* Longest valid hostname */
 #define NET_IP_STR_LEN 46                             /* INET6_ADDRSTRLEN is 46, but we need to be sure */
-#define NET_ADDR_STR_LEN (NET_IP_STR_LEN + 32)        /* Must be enough for ip:port */
 #define NET_HOST_PORT_STR_LEN (NET_HOST_STR_LEN + 32) /* Must be enough for hostname:port */
 
 struct aeEventLoop;
--- a/src/connection.h
+++ b/src/connection.h
@@ -40,7 +40,7 @@
 #include "ae.h"
 
 #define CONN_INFO_LEN 32
-#define CONN_ADDR_STR_LEN 128 /* Similar to INET6_ADDRSTRLEN, hoping to handle other protocols. */
+#define CONN_ADDR_STR_LEN 128 /* Similar to NET_ADDR_STR_LEN, hoping to handle other protocols. */
 
 #define NET_HOST_STR_LEN 256                          /* Longest valid hostname */
 #define NET_IP_STR_LEN 46                             /* INET6_ADDRSTRLEN is 46, but we need to be sure */
--- a/src/networking.c
+++ b/src/networking.c
@@ -1692,13 +1692,8 @@ void acceptCommonHandler(connection *conn, struct ClientFlags flags, char *ip) {
 
     char addr[CONN_ADDR_STR_LEN] = {0};
     char laddr[CONN_ADDR_STR_LEN] = {0};
-    if (conn->type == connectionTypeUnix()) {
-        snprintf(addr, sizeof(addr), "%s:0", server.unixsocket);
-        snprintf(laddr, sizeof(laddr), "%s:0", server.unixsocket);
-    } else {
-        connFormatAddr(conn, addr, sizeof(addr), 1);
-        connFormatAddr(conn, laddr, sizeof(addr), 0);
-    }
+    connFormatAddr(conn, addr, sizeof(addr), 1);
+    connFormatAddr(conn, laddr, sizeof(addr), 0);
 
     if (connGetState(conn) != CONN_STATE_ACCEPTING) {
         serverLog(LL_VERBOSE, "Accepted client connection in error state: %s (addr=%s laddr=%s)",
--- a/src/networking.c
+++ b/src/networking.c
@@ -3815,13 +3815,7 @@ void readQueryFromClient(connection *conn) {
  * you want to relax error checking or need to display something anyway (see
  * anetFdToString implementation for more info). */
 void genClientAddrString(client *client, char *addr, size_t addr_len, int remote) {
-    if (client->flag.unix_socket) {
-        /* Unix socket client. */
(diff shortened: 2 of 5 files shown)


src/networking.c
```diff
@@ -1690,10 +1690,15 @@ void acceptCommonHandler(connection *conn, struct ClientFlags flags, char *ip) {
     client *c;
     UNUSED(ip);
 
-    char addr[NET_ADDR_STR_LEN] = {0};
-    char laddr[NET_ADDR_STR_LEN] = {0};
-    connFormatAddr(conn, addr, sizeof(addr), 1);
-    connFormatAddr(conn, laddr, sizeof(addr), 0);
```
enjoy-binbin, line 1696: this used to always return /unixsocket for unix socket client, we can keep it if it does not important.
```
    } else if (sa.ss_family == AF_UNIX) {
        if (ip) {
            int res = snprintf(ip, ip_len, "/unixsocket");
            if (res < 0 || (unsigned int)res >= ip_len) goto error;
        }
        if (port) *port = 0;
```

src/connection.h
```diff
@@ -40,7 +40,7 @@
 #include "ae.h"
 
 #define CONN_INFO_LEN 32
-#define CONN_ADDR_STR_LEN 128 /* Similar to INET6_ADDRSTRLEN, hoping to handle other protocols. */
+#define CONN_ADDR_STR_LEN 128 /* Similar to NET_ADDR_STR_LEN, hoping to handle other protocols. */
```
enjoy-binbin: now the line 47, NET_ADDR_STR_LEN is not used anymore, we can delete the define if needed. I'm keeping it for now in case we need it in the future.

src/networking.c
```diff
@@ -1690,10 +1690,15 @@ void acceptCommonHandler(connection *conn, struct ClientFlags flags, char *ip) {
...
+    char addr[CONN_ADDR_STR_LEN] = {0};
+    char laddr[CONN_ADDR_STR_LEN] = {0};
+    if (conn->type == connectionTypeUnix()) {
+        snprintf(addr, sizeof(addr), "%s:0", server.unixsocket);
+        snprintf(laddr, sizeof(laddr), "%s:0", server.unixsocket);
```
zuiderkwast: Connection-type specific logic shouldn't be needed if the connection abstraction is good enough.

Let's do this in unix.c? There is the callback for each connection type used by connFormatAddr.

The hard-coded "/unixsocket" is in `anetFdToString` which is called by connSocketAddr for TCP. But we don't need to use this in connUnixAddr.

```diff
 static int connUnixAddr(connection *conn, char *ip, size_t ip_len, int *port, int remote) {
-    return connectionTypeTcp()->addr(conn, ip, ip_len, port, remote);
+    UNUSED(remote);
+    snprintf(ip, ip_len, "%s", server.unixsocket);
+    *port = 0;
+    return 0;
 }
```

With this, we don't need the special case here in networking.c

zuiderkwast (commented):
The change to not use the hard-coded "/unixsocket" anymore is new in this PR? Mention in top comment.

enjoy-binbin:
> The change to not use the hard-coded "/unixsocket" anymore is new in this PR? Mention in top comment.

Yes, i already mention it in the top comment as a cleanup, it is now only affect the log. (wo only use it in the log right now in acceptCommonHandler)

src/networking.c
enjoy-binbin: yes, we can do it. I just didn't want to change connUnixAddr before.

enjoy-binbin: do you think we should format it like this? Like we use it in genClientAddrString?
```
For Unix sockets we use path:0, like in "/tmp/valkey:0".
```

src/connection.h
zuiderkwast: I think we should delete it. There are too many very similar defines and hard to know which one to use. If we keep it, we will forget it and keep it forever.

zuiderkwast:
> Yes, i already mention it in the top comment as a cleanup, it is now only affect the log. (wo only use it in the log right now in acceptCommonHandler)

OK, but the title says "too long display issue". If we use "/unixsocket" it's not too long. It's too long only when we use the correct path. :smile: So maybe using the correct path in the log is a better title?

enjoy-binbin:
> When displaying client addr / laddr using catClientInfoString for example,
if the client is connected via unix domain socket and the unixsocket exceeds
NET_ADDR_STR_LEN, the output will be truncated because NET_ADDR_STR_LEN is
not long enough.

ok, i see the problem, sorry about that. We will show `client addr / laddr` in CLIENT INFO / CLIENT LIST, they are using catClientInfoString. I should give a example, it is something like `id=7912 addr=/xxxxxx laddr=/xxxxxx`, see getClientPeerId function

And the other change `/unixsocket`, it is only used in here https://github.com/valkey-io/valkey/pull/2216#discussion_r2149077169

src/networking.c
zuiderkwast: Yes, "path:0" is OK.

> I just didn't want to change connUnixAddr before.

Because conn can maybe be used for more unix sockets, not only server.unixsocket? I guess it's a valid concern for the future. I can imagine future ideas like maybe we want another admin socket or something, or cluster bus for local cluster over unix socket...

We can postpone this problem. Maybe it will never happen. The TLS conn is hard-coded to server TLS configuration already, so it is a similar problem.

zuiderkwast (approved):

enjoy-binbin:
This isn't a common problem, and i just noticed it when looking at other issues, so we probably don't need release-notes or backports? @zuiderkwast please also check the title and top comment and see if i write it right

PR #2223, opened 2025-06-17
Add non-fatal unit test assertions

thaliaarchi:
Add a new kind of assertion, `TEST_EXPECT`, to the testing framework for non-fatal assertions, where all such failed assertions are reported at once. Existing tests still use fatal assertions.

Name it `TEST_EXPECT`, following the convention of [Google Test](https://google.github.io/googletest/reference/assertions.html), which uses “assert” and “expect” with the same fatal/non-fatal distinction.

### DIFF
--- a/src/unit/README.md
+++ b/src/unit/README.md
@@ -1,7 +1,7 @@
 ## Introduction
 Valkey uses a very simple C testing framework, built up over time but now based loosely off of [Unity](https://www.throwtheswitch.org/unity).
 
-All test files being test_ in the unit directory.
+All test files are located at `src/unit/test_*`.
 A single test file can have multiple individual tests, and they must be of the form `int test_<test_name>(int argc, char *argv[], int flags) {`, where test_name is the name of the test.
 The test name must be globally unique.
 A test will be marked as successful if returns 0, and will be marked failed in all other cases.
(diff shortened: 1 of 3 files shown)


thaliaarchi:
I developed this to assist in testing https://github.com/valkey-io/valkey/pull/2161.

zuiderkwast:
Is it possible to do this without the extra boilerplate TEST_BEGIN and TEST_END wrappers?

The TEST_EXPECT macros could increment a global and the framework that calls each test case can check it..?

thaliaarchi:
Are tests run in parallel? Or could they be in the future?

thaliaarchi:
If tests become executed in parallel, I don't think this would be the largest change that would need to be made, so I made the change.

zuiderkwast (approved):
Great that we could skip the BEGIN/END macros.

The tests don't currently run in parallel. There is more global state that prevent this, at least for some of the test suites, so it's not possible anyway, but we could probably fix it in the future if we want. The unit tests are already very fast though compared to the integration tests so it's not top prio.

thaliaarchi:
Do you know what's wrong with the test-ubuntu-latest-cmake in CI? Nothing from the log looks like a failure.

zuiderkwast:
> Do you know what's wrong with the test-ubuntu-latest-cmake in CI? Nothing from the log looks like a failure.

A flaky test timed out. When it happens, we can't see which one in the end of the output, but it's possible to search for it and find it. https://github.com/valkey-io/valkey/actions/runs/15701236939/job/44236309377?pr=2223#step:4:6554

    sock557d4eeb7b20 => (IN PROGRESS) Active Defrag big list: cluster

This one is not related to this PR so we can ignore it for now.

If you want fix more framework related problems, this is one. :) We've also talked about some way of flagging known flaky tests so they don't cause the CI to fail.

PR #2226, opened 2025-06-17
Reset failover failed primary rank when doing a manual failover

enjoy-binbin:
This rank was added in #1018, we should reset it to zero when doing
the manual failover, just like we reset failover auth rank.

It affects subsequent log printing. When there is a manual failover
and there are multiple primaries failure, `Start of election` will
print a confusing log.


zuiderkwast (approved):

sarthakaggarwal97 (approved):

zuiderkwast:
Backport to which versions?

enjoy-binbin:
it was added in 8.1, if we want to backport, that is the version

zuiderkwast:
Added to project 8.1.

PR #2227, opened 2025-06-17
Do the failover immediately if the replica is the best ranked replica

enjoy-binbin:
In #2023 (#2209, etc.), we are exploring ways to make failover faster, that is,
to minimize the delay.

When a node is marked as FAIL and before the failover starts, there is a delay of
500-1000ms. The original purpose of this delay:
1. Allow FAIL to propagate to at least a majority of the primaries. This makes sure
   they will vote when a replica sends failover auth request.
2. Allow replicas to exchange their offsets, so they will have a correct view of
   their own rank.

We want to minimize this delay while ensuring safety. It is useful for example in
these cases:
1. If there is only one replica, then we don't need any delay, or
2. If there are more replicas, with a fast network, the replicas can exchange the
   offsets very quickly and start the failover within a few milliseconds instead
   of 500-1000ms.

In this PR, when we can be sure that the replica is the best ranked replica, we let
it initiate a failover immediately and completely remove the delay.

### How to ensure safety?

1. To make sure this replica has the best rank, it only skips the delay if it is
sure that it have the best rank and that all replicas in the same shard agree that
the primary is failing. A new flag `CLUSTER_NODE_MY_PRIMARY_FAIL` is introduced to
indicate that each replica has marked its primary as FAIL. If all replicas say that
the primary is failing, we also know that the offset is not updated, because the
offset is not incrementing when the primary is failing. We can skip the delay only
if we have received a message from all replicas and they all have set this flag.

2. To make sure the primaries will vote even if they didn't receive the FAIL yet,
we use the `CLUSTERMSG_FLAG0_FORCEACK` to make sure they will vote. This is equivalent
to broadcasting a FAIL message to all primaries before we broadcast the failover
auth request (but cheaper).

The race between FAIL (broadcast by A) and AUTH REQUEST (broadcast by R) is illustrated
in the following sequence diagram:
```
A      R        B      C
|      |        |      |
| FAIL |        |      |
|----->| AUTH R.|      |
|      |------->|      |
| FAIL |        |      |
|-------------->|      |
|      | AUTH R.|      |
|      |-------------->|
| FAIL |        |      |
|--------------------->|
```

### Details

This is the how the failover is initiated, with new steps marked with **(new)**:

1. A majority of primaries have marked another primary as PFAIL.
2. Some nodes counts failure reports and marks the failing primary as FAIL. The node
   that detects FAIL broadcasts it to all nodes in the cluster.
3. When a replica receives FAIL (or detects FAIL itself by counting PFAIL reports) it
   schedules a failover:
   a. It sets a timeout (500ms + random 0-500ms).
   b. It broadcasts pong to the other replicas in the same shard.
   c. **(new)** The pong (actually the clusterMsg header) has a new flag `CLUSTER_NODE_MY_PRIMARY_FAIL`.
      When the replicas broadcast pong to each other here, this flag is set.
4. **(new)** When the following conditions are met, skip the remaining delay and start
   the failover using AUTH REQUEST with the FORCE ACK flag set, that is if
   a. a PONG is received from every other replica in the same shard (broadcast within
   the shard) and
   b. all replicas have marked that its primary is FAIL in their last message (the new
   `CLUSTER_NODE_MY_PRIMARY_FAIL` flag is set) and
   c. this is the best replica (rank = 0) and
   d. my replication offset != 0.
5. When the delay has passed and no other replica has initiated failover, then initiate failover.

Notes:

* With 3(c), we don't need to wait for FAIL to propagate to all voting primaries.
  At this point, a FAIL has already been broadcast by some node, but there is a race
  so our auth request may arrive to some node before the FAIL. Using the FORCE ACK flag
  ensures the primaries will vote for us. (It is equivalent to broacasting another FAIL
  just before broadcasting auth request.)
* 4(b) ensures that we have received the replication offset from all other replicas and
  that it's up to date. If a replica says that it's primary is failing, it also means that
  the replication from the primary to that replica has stopped.
* 4(c) is to avoid a special bad case. It can happen that not all replicas know about each
  other. In this case, two replicas can think they are both the best replica and start the
  failover at the same time. This can already happen without this PR. When it happens, it
  usually means that a new replica has just joined and it has no data (offset = 0) and if
  it wins the election, there is a problem of dataloss (discussed and partially mitigated
  for the replica migration case in #885). To avoid this case, skip this fast failover path
  if the replica has offset = 0.


src/cluster_legacy.c
```diff
@@ -2204,6 +2204,13 @@ void markNodeAsFailingIfNeeded(clusterNode *node) {
     node->flags |= CLUSTER_NODE_FAIL;
     node->fail_time = mstime();
 
+    if (nodeIsReplica(myself) && myself->replicaof == node) {
+        /* Immediately check if the node is our primary node, if so, we can try
+         * to initiate a failover as soon as possible. */
+        myself->flags |= CLUSTER_NODE_MY_PRIMARY_FAIL;
+        clusterDoBeforeSleep(CLUSTER_TODO_HANDLE_FASTER_FAILOVER);
+    }
```
zuiderkwast: This is inside the function markNodeAsFailingIfNeeded. The purpose of the function is to check failure report count and to mark the node as FAIL if enough PFAIL reports are found. Maybe trigger failover is not the purpose of this function... but it already has some side effect mentioned in the top comment of the function.

```
 * If a failure is detected we also inform the whole cluster about this
 * event trying to force every other node to set the FAIL flag for the node.
```

After this text, let's add the information that this function can trigger failover:

```
 *
 * If the node marked as FAIL is a primary and myself is a replica of this node,
 * we initiate a failover.
```

I'm not sure if it's unexpected that markNodeAsFailingIfNeeded also can trigger failover. I'm just thinking about separation of concern here. An alternative idea is to trigger the failover just after markNodeAsFailingIfNeeded in clusterProcessGossipSection. Maybe. But it can be OK to trigger the failover inside markNodeAsFailingIfNeeded if we describe it in the comment.

src/cluster_legacy.c
```diff
@@ -3319,7 +3328,7 @@ int clusterProcessPacket(clusterLink *link) {
         sender->flags |= CLUSTER_NODE_EXTENSIONS_SUPPORTED;
     }
 
-    /* Checks if the node supports light message hdr */
+    /* Checks if the node supports some flags. */
     if (sender) {
         if (flags & CLUSTER_NODE_LIGHT_HDR_PUBLISH_SUPPORTED) {
```
zuiderkwast: "node supports some flags" is not correct.

Some flags indicate that the node supports light messages. Other flags indicate other things.

The CLUSTER_NODE_MY_PRIMARY_FAIL flag doesn't indicate support for a feature. If it's set, it means that the sender supports this feature and it's primary is failing now. If it's not set, we don't know if the sender supports this feature or not... So "support" is not the right word for this new flag.

```suggestion
    /* Store some flags about the sender. */
    if (sender) {
        /* Check if the node supports light message hdr */
        if (flags & CLUSTER_NODE_LIGHT_HDR_PUBLISH_SUPPORTED) {
```

src/cluster_legacy.c
```diff
@@ -3777,6 +3798,16 @@ int clusterProcessPacket(clusterLink *link) {
...
+
+                if (nodeIsReplica(myself) && myself->replicaof == failing) {
+                    /* Immediately check if the node is our primary node, if so, we can try
+                     * to initiate a failover as soon as possible. We will also try to send
+                     * the FAIL packet in case we trigger the failover immediately but unable
+                     * to get the votes. */
+                    myself->flags |= CLUSTER_NODE_MY_PRIMARY_FAIL;
+                    clusterSendFail(failing->name);
+                    clusterDoBeforeSleep(CLUSTER_TODO_HANDLE_FASTER_FAILOVER);
+                }
```
zuiderkwast: This code is added inside clusterProcessPacket when we receive a FAIL packet.

It means another node has already broadcast the FAIL. If we broadcast it again to all nodes in the cluster, and every other replica of the failing node also broadcast it, it can be a storm of FAIL messages. In a cluster with N total nodes and M nodes per shard, it is total M * N FAIL messages in the cluster in a short time. Example: 100 nodes, 5 nodes per shard = 500 FAIL messages. Is it right?

Actually, I think we shouldn't broadcast FAIL if another node has already broadcast FAIL.

:bulb: There is another way to make sure all primaries will vote for us: We can use the flag CLUSTERMSG_FLAG0_FORCEACK, same as in manual failover. It has a very similar effect to broadcast FAIL + broadcast AUTH_REQ. I think we should only do this if clusterAllReplicasThinkPrimaryIsFail and my rank is 0.

zuiderkwast (commented):
Good job!

I have some ideas. See comments.

Some test case would be nice to see, to check that we can do a failover faster than the delay that we skip. For example with node timeout 2000, check that we can do failover in 2500 ms? I guess a test like this can be flaky but at least try it to see if the feature works, would be good.

I'm a little confused about `CLUSTER_TODO_HANDLE_FASTER_FAILOVER`. Why do we need it? Why is it different to `CLUSTER_TODO_HANDLE_FAILOVER`, when can we use it, when can we not use it, etc. It is not clear to me.

IIUC, it's not really needed for the the feature *skip delay if  clusterAllReplicasThinkPrimaryIsFail() returns true and my rank is 0*, is it? If it's a separate optimization, then I guess we skip the FASTER_FAILOVER idea in this PR, to keep it simpler. WDYT?

src/cluster_legacy.c
enjoy-binbin: > It means another node has already broadcast the FAIL. If we broadcast it again to all nodes in the cluster, and every other replica of the failing node also broadcast it, it can be a storm of FAIL messages. In a cluster with N total nodes and M nodes per shard, it is total M * N FAIL messages in the cluster in a short time. Example: 100 nodes, 5 nodes per shard = 500 FAIL messages. Is it right?

yes, it is indeed, in a worst case.

> There is another way to make sure all primaries will vote for us: We can use the flag CLUSTERMSG_FLAG0_FORCEACK, same as in manual failover. It has a very similar effect to broadcast FAIL + broadcast AUTH_REQ. I think we should only do this if clusterAllReplicasThinkPrimaryIsFail and my rank is 0.

I will think about it.

This broadcast actually has another meaning, it propagates CLUSTER_NODE_MY_PRIMARY_FAIL and offset, i forgot to mention it.

src/cluster_legacy.c
enjoy-binbin: we anyway do the failover in clusterDoBeforeSleep or in ClusterCron, and we are setting a beforesleep flag in here just like other flags cases, so i think it can be fine. And sure a comment will be helpful, i will add it later.

enjoy-binbin:
> Some test case would be nice to see, to check that we can do a failover faster than the delay that we skip. For example with node timeout 2000, check that we can do failover in 2500 ms? I guess a test like this can be flaky but at least try it to see if the feature works, would be good.

Yes, i will add the test later, i test it locally yesterday and it work.


> I'm a little confused about CLUSTER_TODO_HANDLE_FASTER_FAILOVER. Why do we need it? Why is it different to CLUSTER_TODO_HANDLE_FAILOVER, when can we use it, when can we not use it, etc. It is not clear to me.

Yes, it is indeed a bit confusing, sorry i did not mention it. I want to put CLUSTER_TODO_HANDLE_FASTER_FAILOVER after other flags, like after CLUSTER_TODO_UPDATE_STATE, so that we will first print CLUSTER IS DOWN and then try to trigger the failover. That is now we will trigger failover faster, in the past, we will update the cluster state and then try to failover in ClusterCron.

> IIUC, it's not really needed for the the feature skip delay if clusterAllReplicasThinkPrimaryIsFail() returns true and my rank is 0, is it? If it's a separate optimization, then I guess we skip the FASTER_FAILOVER idea in this PR, to keep it simpler. WDYT?

Sure, i can remove FASTER_FAILOVER and then only put it in #2209

src/cluster_legacy.c
zuiderkwast: Yes, but for that we only need to broadcast to local replicas in the same shard. We do `clusterBroadcastPong(CLUSTER_BROADCAST_LOCAL_REPLICAS)`.

enjoy-binbin: Or from another point of view, if myself replica think primary is down, that mean that at least a quorm in the cluster has been reached and mark primary to FAIL, so that we can also assume that vote will have a quorum.

zuiderkwast: Are you sure?

I imagine there is a race. We only know a majority think it is PFAIL, and one node knows it is FAIL.

When a majority think it is PFAIL and one node (A) finds out there is a majority, it will broadcast FAIL. When this replica (R) gets the FAIL and immediately sends AUTH REQ, there is change the AUTH REQ will reach the other primaries before the FAIL broadcast from A. No?

We should draw sequence diagrams to visualize the scenario.

zuiderkwast:
```
A      R        B      C
|      |        |      |
| FAIL |        |      |
|----->| AUTH R.|      |
|      |------->|      |
| FAIL |        |      |
|-------------->|      |
|      | AUTH R.|      |
|      |-------------->|
| FAIL |        |      |
|--------------------->|
```

enjoy-binbin: ohh, right, that is true, you are right, there will be a race anymore, silly me, i forgot it.

That is the same reason why we added this logic (even if the node is a replica). 
The main purpose of this is to handle a case that myself is a replica, and then replica receive a pfail message and got a quorum, and then mark my primary as FAIL, and then in the next clusterCron trigger a failover. 
```
    /* Broadcast the failing node name to everybody, forcing all the other
     * reachable nodes to flag the node as FAIL.
     * We do that even if this node is a replica and not a primary: anyway
     * the failing state is triggered collecting failure reports from primaries,
     * so here the replica is only helping propagating this status. */
    clusterSendFail(node->name);
```

If we want to save FAIL messages, we should add the `nodeIsReplica(myself) && myself->replicaof == node` condition same here. That is, become something like:
```
    if (nodeIsPrimary(myself) || (nodeIsReplica(myself) && myself->replicaof == node)) {
        clusterSendFail(node->name);
    }
```

Of course, if we consider its other function, which is to try to make all nodes reach a consensus on the FAIL node earlier, then it is also reasonable.

It seems that every automatic failover should succeed? We should remove this check?
```
        } else if (!nodeFailed(primary)) {
            serverLog(LL_WARNING, "Failover auth denied to %.40s (%s) for epoch %llu: its primary is up", node->name,
                      node->human_nodename, (unsigned long long)requestCurrentEpoch);
```

If in the above diagram, R's primary suddenly comes online in the middle, what should we do? There are actually many races here that may cause some writes to be lost, but i suppose it is ok since an automatic failover does not guaranteed it.

I might still want to do the sendFail logic instead of a CLUSTERMSG_FLAG0_FORCEACK flag. But it is indeed a lot of traffic. Although in the worst case we also send FAIL in markNodeAsFailingIfNeeded.

Maybe we need a CLUSTER_NODE_YOUR_PRIMARY_FAIL flag, if YOUR know my primary is FAIL, i would't send the FAIL message when broadcast the FAIL.

enjoy-binbin: i took a try (setting the mf force flag) and see how does it go.

zuiderkwast: Exactly. This is why I think the replica can use the force flag if it is sure it is the best ranked. We fallback to the old logic with delay and no force flag.

src/cluster_legacy.c
```diff
@@ -3539,12 +3572,10 @@ int clusterProcessPacket(clusterLink *link) {
                  * the clients, and the replica will never initiate a failover since the
                  * node is not actually in FAIL state. */
                 if (!nodeFailed(noaddr_node)) {
-                    noaddr_node->flags &= ~CLUSTER_NODE_PFAIL;
-                    noaddr_node->flags |= CLUSTER_NODE_FAIL;
-                    noaddr_node->fail_time = now;
+                    markNodeAsFailing(noaddr_node);
                     clusterSendFail(noaddr_node->name);
                 }
-                clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_UPDATE_STATE);
```
enjoy-binbin, line 4087: todo: should revert this change

src/cluster_legacy.c
```diff
@@ -127,13 +126,22 @@ sds clusterEncodeOpenSlotsAuxField(int rdbflags);
...
     return (n->flags & CLUSTER_NODE_PRIMARY) && n->numslots;
 }
 
+/* Returns if myself is the best ranked replica in an automatic failover process. */
+static inline int myselfIsBestRankedReplica(void) {
+    return (server.cluster->mf_end == 0 &&
+            server.cluster->failover_auth_rank == 0 &&
+            server.cluster->failover_failed_primary_rank == 0 &&
+            clusterAllReplicasThinkPrimaryIsFail());
+}
```
zuiderkwast: Let's also include a check that myself's replication offset != 0 here?

It makes sure that if myself is a new replica just connected that don't yet know about other replicas, we should not try to do this fast failover.

zuiderkwast: With #885, this would also ~~prevent~~ make it less likely that a "confused" replica will win the election.

zuiderkwast:
@hpatro I have discussed with @enjoy-binbin and updated the top comment with a more thorough reasoning about the possible scenarios. PTAL.

The special case that a new replica joins and doesn't know about the other replicas is a serious problem, but it's not new. It's not solved here but we can avoid making it worse. (I hope we can solve it in another PR, to make sure a replica is made aware of the other replicas in the shard ASAP, for example it starts replicating.)

enjoy-binbin:
> Let's also include a check that myself's replication offset != 0 here?

> It makes sure that if myself is a new replica just connected that don't yet know about other replicas, we should not try to do this fast failover.

> The special case that a new replica joins and doesn't know about the other replicas is a serious problem, but it's not new. It's not solved here but we can avoid making it worse. (I hope we can solve it in another PR, to make sure a replica is made aware of the other replicas in the shard ASAP, for example it starts replicating.)

Ok, i pushed the new commit, sorry for the late response, i somehow forgot it and thought we are waitting response from @hpatro

zuiderkwast:
@madolson This is based on your idea to keep track of other replicas' view of the primay's fail status. Great if you can take a look.

hpatro (commented):
Haven't taken a complete look. 

@enjoy-binbin Could you please share if we still continue the primary rank logic and not bypass that with fast failover logic?

enjoy-binbin:
> Could you please share if we still continue the primary rank logic and not bypass that with fast failover logic?

What do you means? sorry i don't quite get the question.

tests/unit/cluster/faster-failover.tcl
```diff
@@ -0,0 +1,164 @@
...
+        verify_no_log_message -6 "*Failover attempt expired*" 0
+        verify_no_log_message -11 "*Failover attempt expired*" 0
+    }
+}
+
+# This change and test is quite important, so we want to run it a few more times.
+
+start_cluster 5 7 {tags {external:skip cluster} overrides {cluster-ping-interval 1000 cluster-node-timeout 5000}} {
+    test_best_ranked_replica
+} my_slot_allocation cluster_allocate_replicas ;# start_cluster
```
zuiderkwast: Will you remove this before merging?

If you want to keep running it multiple times, wrap the code in for loop instead of duplicating the code.

zuiderkwast (approved):
I would like to have this in 9.0. I hope others will agree. Maybe we can merge it after RC1.

tests/unit/cluster/faster-failover.tcl
enjoy-binbin: ohh right, i change to use a loop, i indeed want to run it for a few times so that we are not messing this up. 5 times take 120 seconds in my PC, so i think that is acceptable (we can change it in the furture if we think they are slow).

madolson:
I would like to pause this for 9.0. We have seen a lot of cluster instability in AWS after merging 8.1 (much worse than in any past release), and I don't think we have the testing in place for all of these optimizations to make sure we aren't ending up in odd states.

Nikhil-Manglore:
@enjoy-binbin Hey BinBin we're currently running this PR through the Valkey-Fuzzer to check if it causes any failures. Do you think you can also rebase this PR when you get a chance so we can get more accurate results?

I'll update here later today if this PR caused any cluster bus issues after I go through the current fuzzer results

enjoy-binbin:
@Nikhil-Manglore Thanks! I updated the PR, please let me know if you find anything.

Nikhil-Manglore:
No real failures so far, i'll keep monitoring

zuiderkwast:
Let's merge this?

@madolson OK?

madolson:
Go for it!

zuiderkwast:
Test case failed in Daily:

```
*** [err]: The best replica can initiate an election immediately in an automatic failover in tests/unit/cluster/faster-failover.tcl
log message of '"*Successful partial resynchronization with primary*"' not found in ./tests/tmp/server.10822.103/stdout after line: 0 till line: 155
```

https://github.com/valkey-io/valkey/actions/runs/23774277197/job/69272507658#step:9:10598

PR #2237, opened 2025-06-20
New module API event for tracking authentication attempts

### LINKED ISSUE #2211: [NEW] Extend module-API with Authentication result

Modules currently can implement custom authentication logic(e.g. LDAP), but they don't have visibility into authentication events and outcomes. This prevents modules from implementing important security features such as:

Authentication audit logging and monitoring
- Failed login attempt tracking and brute force protection
- User activity analytics and reporting
- Integration with external security systems for alerting
- Rate limiting based on authentication patterns

While modules can handle the authentication process itself, they cannot observe authentication attempts made through other means (default user, built-in ACL, other modules).

Add a new module API callback that is triggered after every authentication attempt at the end of ACLAuthenticateUser with the authentication result so both success and failed attempts are available.

rjd15372:
In this commit we introduce a new module API event called `ValkeyModuleEvent_AuthenticationAttempt` to track successful/failed authentication attempts.

This event will fill a struct, called `ValkeyModuleAuthenticationInfo`, with the client ID of the connection, the username, the module name  that handle the authentication event, and the result of the authentication attempt.

Fixes: #2211


src/acl.c
```diff
@@ -1453,16 +1454,32 @@ void addAuthErrReply(client *c, robj *err) {
...
+    {
+        ValkeyModuleAuthenticationInfo info = VALKEYMODULE_AUTHENTICATIONINFO_INITIALIZER_V1;
+        info.client_id = c->id;
+        int port;
+        connAddrPeerName(c->conn, info.addr, sizeof(info.addr), &port);
+        info.port = port;
+        info.username = username->ptr;
+        info.module_name = NULL;
+        info.success = (result == AUTH_OK);
+        moduleFireServerEvent(VALKEYMODULE_EVENT_AUTHENTICATION_ATTEMPT, 0, &info);
```
hpatro: is `AUTH_OK` or `AUTH_ERR` enough for a module or should we have reason also passed down to the module?

src/valkeymodule.h
```diff
@@ -780,6 +782,21 @@ typedef struct ValkeyModuleKeyInfo {
 
 #define ValkeyModuleKeyInfo ValkeyModuleKeyInfoV1
 
+#define VALKEYMODULE_AUTHENTICATION_INFO_VERSION 1
+typedef struct ValkeyModuleAuthenticationInfo {
+    uint64_t version;        /* Version of this structure for ABI compat. */
+    uint64_t client_id;      /* Client ID. */
+    char addr[46];           /* IPv4 or IPv6 address. */
```
hpatro:
```
    char addr[NET_IP_STR_LEN];   /* IPv4 or IPv6 address. */
```

tests/modules/hooks.c
```diff
@@ -205,7 +205,7 @@ void persistenceCallback(ValkeyModuleCtx *ctx, ValkeyModuleEvent e, uint64_t sub
     /* modifying the keyspace from the fork child is not an option, using log instead */
     ValkeyModule_Log(ctx, "warning", "module-event-%s", keyname);
     if (sub == VALKEYMODULE_SUBEVENT_PERSISTENCE_SYNC_RDB_START ||
-        sub == VALKEYMODULE_SUBEVENT_PERSISTENCE_SYNC_AOF_START) 
+        sub == VALKEYMODULE_SUBEVENT_PERSISTENCE_SYNC_AOF_START)
```
hpatro: remove

tests/unit/moduleapi/hooks.tcl
```diff
@@ -26,7 +27,7 @@ tags "modules" {
             # assert server is still up
             assert_equal [r ping] PONG
             $rd close
-        } 
+        }
```
hpatro: remove

tests/unit/moduleapi/hooks.tcl
```diff
@@ -298,7 +339,7 @@ tags "modules" {
         }
 
         test {Test configchange hooks} {
-            r config set rdbcompression no 
+            r config set rdbcompression no
```
hpatro: remove

src/acl.c
rjd15372: I don't have a strong opinion, but since we also don't log a reason in the ACL log, I followed the same here.

src/valkeymodule.h
rjd15372: `NET_IP_STR_LEN` is defined in `connection.h`, which cannot be included in this file. So either we keep the `46`, or copy the macro definition to this file. What do you prefer?

src/valkeymodule.h
```diff
@@ -780,6 +782,21 @@ typedef struct ValkeyModuleKeyInfo {
...
+    uint16_t port;           /* TCP port. */
```
murphyjacob4: addr and port are accessible through ValkeyModule_GetClientInfoById. In my opinion - we should keep the struct minimal and make the module use this API to get more details. WDYT?

src/module.c
```diff
@@ -8067,17 +8067,41 @@ int checkModuleAuthentication(client *c, robj *username, robj *password, robj **
         serverAssert(result == VALKEYMODULE_AUTH_HANDLED);
         return AUTH_BLOCKED;
```
murphyjacob4, line 8068: Do we also need to trigger this in VM_AuthenticateClientWithACLUser? For asynchronous authentication?

src/valkeymodule.h
```diff
@@ -780,6 +782,21 @@ typedef struct ValkeyModuleKeyInfo {
...
+    const char *username;    /* Username used for authentication. */
+    const char *module_name; /* Name of the module that is handling the authentication. */
+    int success;             /* Result of the authentication, 1 for success, 0 for failure. */
```
murphyjacob4: Not a strong opinion - but does an enum make more sense for future proofing? If so, maybe naming it GRANTED/DENIED makes more sense, so we can differentiate a successful denial from an internal issue (timeout, etc).

hpatro: I like this suggestion. Kind of boils down to my request where we're able to differentiate about the failure.

src/valkeymodule.h
rjd15372: I agree with you.

src/module.c
rjd15372: No need. When the client unblocks it will go through this code path again and will call the callback below.

src/valkeymodule.h
rjd15372: Me too. I'll use an enum.

rjd15372: Unfortunately we can't distinguish between a password mismatch or an authentication failure because the module API to handle authentication does not provide that information.

src/module.c
```diff
@@ -11763,6 +11786,23 @@ static uint64_t moduleEventVersions[] = {
  *
  *         ValkeyModuleKey *key;    // Key name
  *
+ * * ValkeyModule_Authentication_Attempt
```
madolson: It looks like you modified the struct since you wrote the docs. Looking up the clients looks like the right solution to extract the ip.

src/module.c
```diff
@@ -8067,17 +8067,39 @@ int checkModuleAuthentication(client *c, robj *username, robj *password, robj **
         serverAssert(result == VALKEYMODULE_AUTH_HANDLED);
         return AUTH_BLOCKED;
     }
+
+    ValkeyModuleAuthCtx *auth_ctx = c->module_data ? c->module_data->module_auth_ctx : NULL;
```
madolson, line 8071: I assume we always want c->module_data to be set, do we know what cases it's null now?

rjd15372:
@madolson can we get this in for 9.0?

src/module.c
rjd15372: `c->module_data` will be null if no module has registered an authentication callback. This corresponds to the common code path when there are no auth modules loaded in the server.

enjoy-binbin (commented):
LGTM overall.

src/module.c
```diff
@@ -11763,6 +11786,20 @@ static uint64_t moduleEventVersions[] = {
  *
  *         ValkeyModuleKey *key;    // Key name
  *
+ * * ValkeyModule_Authentication_Attempt
```
zuiderkwast: This seems wrong but I'm not sure which struct or value is referenced here. (Maybe other event names in this comment are wrong too?)

```suggestion
 * * ValkeyModuleEvent_AuthenticationAttempt
```

madolson: I think it's supposed to be the define enum, `VALKEYMODULE_EVENT_AUTHENTICATION_ATTEMPT`. I'm not sure why these aren't in all CAPs.

zuiderkwast: We should fix it...

rjd15372: @zuiderkwast should I fix the whole comment, including the above event names?

rjd15372: I fixes the function comment. The names there match the list of `ValkeyModuleEvent` variables declared in valkeymodule.h

src/acl.c
```diff
@@ -1458,16 +1459,30 @@ void addAuthErrReply(client *c, robj *err) {
...
+    }
+
+    {
+        ValkeyModuleAuthenticationInfo info = VALKEYMODULE_AUTHENTICATIONINFO_INITIALIZER_V1;
+        info.client_id = c->id;
+        info.username = username->ptr;
+        info.module_name = NULL;
+        info.result = result == AUTH_OK ? VALKEYMODULE_AUTH_RESULT_GRANTED
+                                        : VALKEYMODULE_AUTH_RESULT_DENIED;
+        moduleFireServerEvent(VALKEYMODULE_EVENT_AUTHENTICATION_ATTEMPT, 0, &info);
```
zuiderkwast: This code is not large, but it's duplicated. If we change V1 to V2 in the future, we'll need to remember to change in both places.

Let's move this and the duplicated code in `checkModuleAuthentication` to a new function in `module.{c,h}`? e.g. `moduleFireAuthenticationEvent(client_id, username, module_name, is_granted)`.

Then, acl.c doesn't need to include valkeymodule.h. It's enough to include module.h.

rjd15372: Done. The code looks much better now. Thanks for the suggestion.

src/module.c
```diff
@@ -11763,6 +11791,20 @@ static uint64_t moduleEventVersions[] = {
...
+ *
+ *     The data pointer can be casted to a ValkeyModuleAuthenticationInfo
+ *     structure with the following fields:
+ *
+ *         uint64_t client_id;      // Client ID.
+ *         const char *username;    // Username used for authentication.
+ *         const char *module_name; // Name of the module that is handling the
+ *                                  // authentication. It is NULL if the
+ *                                  // authentication is handled by the core.
+ *         ValkeyModuleAuthenticationResult result;   // Result of the authentication.
```
zuiderkwast: This documentation is so complete that it can be used without looking at the structs in valkeymodule.h, except for the result values. Let's add them too.

```suggestion
 *         ValkeyModuleAuthenticationResult result;
 *                                  // Result of the authentication:
 *                                  // VALKEYMODULE_AUTH_RESULT_GRANTED or
 *                                  // VALKEYMODULE_AUTH_RESULT_DENIED.
```

zuiderkwast (approved):

zuiderkwast (approved):

PR #2183, opened 2025-06-06
Spelling

jsoref:
This PR corrects misspellings identified by the [check-spelling action](https://github.com/marketplace/actions/check-spelling)

The misspellings have been reported at https://github.com/jsoref/valkey/actions/runs/15496041308/attempts/1#summary-43632891663

The action reports that the changes in this PR would make it mostly happy: https://github.com/jsoref/valkey/actions/runs/15496042495/attempts/1#summary-43632894710

I'll fix up the commits based on feedback -- typically for PRs like this (which by my standards is a medium size), projects have preferences about things being dropped or split.

PR #2245, opened 2025-06-20
Spelling 9

jsoref:
- Subset of #2183

### DIFF
--- a/src/aof.c
+++ b/src/aof.c
@@ -854,13 +854,13 @@ cleanup:
  * AOFRW, which may be that we have reached the 'next_rewrite_time' or the number of INCR
  * AOFs has not reached the limit threshold.
  * */
-#define AOF_REWRITE_LIMITE_THRESHOLD 3
-#define AOF_REWRITE_LIMITE_MAX_MINUTES 60 /* 1 hour */
+#define AOF_REWRITE_LIMIT_THRESHOLD 3
+#define AOF_REWRITE_LIMIT_MAX_MINUTES 60 /* 1 hour */
 int aofRewriteLimited(void) {
     static int next_delay_minutes = 0;
     static time_t next_rewrite_time = 0;
 
-    if (server.stat_aofrw_consecutive_failures < AOF_REWRITE_LIMITE_THRESHOLD) {
+    if (server.stat_aofrw_consecutive_failures < AOF_REWRITE_LIMIT_THRESHOLD) {
         /* We may be recovering from limited state, so reset all states. */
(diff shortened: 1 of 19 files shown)


src/aof.c
```diff
@@ -854,13 +854,13 @@ int openNewIncrAofForAppend(void) {
  * AOFRW, which may be that we have reached the 'next_rewrite_time' or the number of INCR
  * AOFs has not reached the limit threshold.
  * */
-#define AOF_REWRITE_LIMITE_THRESHOLD 3
-#define AOF_REWRITE_LIMITE_MAX_MINUTES 60 /* 1 hour */
+#define AOF_REWRITE_LIMITED_THRESHOLD 3
+#define AOF_REWRITE_LIMITED_MAX_MINUTES 60 /* 1 hour */
```
Fusl: these should be `_LIMIT_` rather than `_LIMITED_`

src/valkey-cli.c
```diff
@@ -6150,7 +6150,7 @@ static int clusterManagerFixOpenSlot(int slot) {
             }
         }
 
-        /* If the node is neither migrating nor importing and it's not
+        /* If the node is neither migrating nor importing and it isn't
```
Fusl: `it's not` and `it isn't` are both correct

tests/unit/multi.tcl
```diff
@@ -657,7 +657,7 @@ start_server {tags {"multi"}} {
         # check that GET and PING are disallowed on stale replica, even if the replica becomes stale only after queuing.
         r multi
         r get xx
-        $r1 replicaof localhsot 0
+        $r1 replicaof localhost_ 0
```
Fusl, line 660: why the `_`?

Fusl (changes requested):

tests/unit/multi.tcl
jsoref: As with `file`, I have initial questions:
1. Is `localhsot` is a typo? I guess "yes"
2. Is it an _intentional_ or _unintentional_ typo? My guess is "yes (it's intentional)"

I need someone else to say one of:
* oh, no, it's accidental, we just want `localhost`
* we really don't want to use `localhost`, please use something to say _this isn't actually `localhost`_

Fusl: You're right, `localhsot` is an intentional typo of `localhost`. Unlike the `filx`->`file` change, I think we can keep this one as `localhost_` unless someone else has any objections.

hwware (approved):

PR #2247, opened 2025-06-20
Consolidating nonexistent spelling 

jsoref:
- Subset of #2183

### DIFF
--- a/src/acl.c
+++ b/src/acl.c
@@ -656,7 +656,7 @@ static void ACLChangeSelectorPerm(aclSelector *selector, struct serverCommand *c
 /* This is like ACLSetSelectorCommandBit(), but instead of setting the specified
  * ID, it will check all the commands in the category specified as argument,
  * and will set all the bits corresponding to such commands to the specified
- * value. Since the category passed by the user may be non existing, the
+ * value. Since the category passed by the user may be nonexistent, the
  * function returns C_ERR if the category was not found, or C_OK if it was
  * found and the operation was performed. */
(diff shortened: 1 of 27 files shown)


tests/sentinel/tests/03-runtime-reconf.tcl
```diff
@@ -197,7 +197,7 @@ test "Set parameters in normal case with bad format" {
 test "Sentinel Set with other error situations" {
 
    # non-existing script
-   assert_error "ERR Notification script seems non existing*" {S 0 SENTINEL SET mymaster notification-script test.txt}
+   assert_error "ERR Notification script seems nonexistent*" {S 0 SENTINEL SET mymaster notification-script test.txt}
```
Fusl: This changes the expected error message that Valkey returns without actually changing the error message in `sentinel.c` itself. I suggest reverting this for now since it's a behavior change.

Fusl (changes requested):

Nikhil-Manglore (approved):
LGTM, I think the test failure is a known flakey test.

zuiderkwast (approved):

rainsupreme (changes requested):
Bulk of this is good. Could you drop:

- src/commands/srem.json and src/commands/zrem.json — these are COMMAND DOCS reply_schema descriptions that may be consumed programmatically
- tests/unit/cluster/slot-stats.tcl — test fixture data (non-existent-arg is passed as an argument to the server)
The ~73 comment fixes are good though, once this has been rebased

sarthakaggarwal97 (approved):

sarthakaggarwal97 (approved):

PR #2255, opened 2025-06-21
Accept socket judge fd overflow

kukey:
when we use sentinel manage a large cluster, the fd may overflow; 
then the clients that cfd greater than the aeEvent size,  lead to 
these clients leak.

Because in createCleint(), we don`t check connSetReadHandler() return values.

### DIFF
--- a/src/socket.c
+++ b/src/socket.c
@@ -326,6 +326,11 @@ static void connSocketAcceptHandler(aeEventLoop *el, int fd, void *privdata, int
             if (errno != EWOULDBLOCK) serverLog(LL_WARNING, "Accepting client connection: %s", server.neterr);
             return;
         }
+        if (cfd >= (int)server.maxclients + CONFIG_FDSET_INCR) {
+            close(cfd);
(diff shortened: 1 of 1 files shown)


kukey:
image
like this sentinel client list, our sentinel set maxclients 10000, but there are some clients fd more than 10000.

arthurkiller:
good catch!

arthurkiller (commented):
perhaps we'd better handle error from `aeCreateFileEvent`

src/socket.c
arthurkiller:
```suggestion
    if (aeCreateFileEvent(server.el, conn->fd, AE_WRITABLE, conn->type->ae_handler, conn) == AE_ERR) {
        conn->state = CONN_STATE_ERROR;
        conn->last_errno = errno;
        return C_ERR
    }
```

arthurkiller (commented):
Check this error, also in `sentinel.c`

src/socket.c
kukey: LGTM!  it`s more better;

PR #2256, opened 2025-06-21
add a parameter to control print Lua scripts

luorong1999:
In https://github.com/valkey-io/valkey/pull/1448, @enjoy-binbin does not want to print the lua when doing check-rdb, we can add a startup parameter <--print_lua> to compatible with the old version?

### DIFF
--- a/src/valkey-check-rdb.c
+++ b/src/valkey-check-rdb.c
@@ -47,6 +47,7 @@ void computeDatasetProfile(int dbid, robj *keyobj, robj *o, long long expiretime
 int rdbCheckMode = 0;
 int rdbCheckStats = 0;
 int rdbCheckOutput = 0;
+int rdbCheckPrintLua = 0;
 long long now;
 
 #define LOW_TRACKE_VALUE 1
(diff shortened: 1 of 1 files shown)


enjoy-binbin:
I mentioned in here https://github.com/valkey-io/valkey/pull/2193#issuecomment-2958871493

> i don't want to print it, but in https://github.com/valkey-io/valkey/pull/1448 we do make a agreement that we will keep it, so i dont want to bother to add the new parameter.

First, it is already a history, i don't want we spent time to add new parameter around it. Either don't print it at all, or just keep it, and we already decided to keep it, the reason is in https://github.com/valkey-io/valkey/pull/1448#issuecomment-2551068281. Second, i don't think people will often use a new check-rdb to check an old replication RDB in the future. 

Are you experiencing an overprinting issue? Or is it simply just read https://github.com/valkey-io/valkey/pull/2193#issuecomment-2958871493? But anyway, please read the related issue for more details.

zuiderkwast:
@enjoy-binbin Shall we close this?

enjoy-binbin (commented):
Yes, as discussed above, we don't want to do anything extra for it at the moment, and it will rarely be used in new versions.

PR #2260, opened 2025-06-23
Do not report negative passing test count

jsoref:
This is quite nonsensical (https://github.com/jsoref/valkey/actions/runs/15787893006/job/44508156695#step:5:483):
> [END] - test_zmalloc.c: 3 tests, -1 passed, 4 failed

### DIFF
--- a/src/testhelp.h
+++ b/src/testhelp.h
@@ -57,13 +57,13 @@ extern int __test_num;
             __failed_tests++;                   \
         }                                       \
     } while (0)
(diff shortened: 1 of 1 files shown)


enjoy-binbin (approved):

zuiderkwast:
Thanks. Also failed 4 when the total number of tests is 3 also makes no sense. Something is wrong with how we count the tests?

Ah, the answer is in #2263

Issue #1978, opened 2025-04-18
[NEW] Tunnel Traffic After Failover On Non-Cluster Mode


yairgott:
### Scope
The purpose of this issue is to initiate a discussion and gather feedback on the proposed approach. While this is not intended to be a full design document, it outlines key ideas and considerations that could shape a potential implementation.

### Preface
Planned failovers, such as those needed when updating the Valkey server, are typically executed using a Create Before Destroy strategy. This ensures the new instance is fully synchronized and ready to serve traffic. However, in standalone setups, ensuring a smooth transition of active connections remains a reliability gap. 

To address this use-case, a new [capability](https://valkey.io/commands/client-capa/) for Standalone has been introduced in Valkey 8.0, enabling connection redirection from the old instance to the new one. This feature relies on client support and offers a long-term solution for graceful failovers. That said, it is expected to take time before this capability becomes broadly reachable. This functionality must first be implemented by client libraries and subsequently adopted by users, most of whom today still rely on Redis clients.

Addressing this at the server level enhances Valkey’s reliability posture, which is a key driver for broader adoption.

### Proposed Approach: Tunneling
The proposed approach is to tunnel the downstream traffic, from the old instance to the new one, once the new instance is ready to take over. This mechanism allows users, particularly those following best practices like periodic connection rotation, to experience a seamless transition, avoiding abrupt disconnections.

### Implementation Considerations
- **Prefer Redirect**: Tunneling is avoided for clients supporting the new Redirect capability.
- **Upstream Selection**: The upstream node's role should match the role of the node prior to failover. For example, if the original node was a replica, it should tunnel upstream to another active replica; likewise, a primary should connect to a new primary. The list of active replicas can be obtained by issuing the [`ROLE`](https://valkey.io/commands/role/) command, from which a suitable target can be randomly selected. If the role of the node prior to failover was a replica and there are no replicas, upstream to the primary. 
- **One-to-One Tunneling**: Once tunneling mode is activated, a dedicated upstream connection to the new instance is established upon receiving the first byte from the downstream connection. It’s important to note that the old instance does not inspect or interpret the data beyond TLS decryption. This behavior classifies the mechanism as tunneling rather than proxying.
- **Node State Triggering**: The node state conditions for entering tunnel mode align with those used for the Redirect mode.
[...]

PR #2265, opened 2025-06-24
Implement tunneling support for non-cluster post-failover scenarios

yairgott:
This is a followup to the following [issue](https://github.com/valkey-io/valkey/issues/1978).

Key Implementation Points:
1. The `FAILOVER` command has been extended to support a `TUNNEL` option, enabling traffic to be tunneled from the demoted primary to the new primary.
2. Tunneling is initiated by the demoted primary under specific conditions: Upon completion of a `FAILOVER` with the `TUNNEL` option, the demoted primary initiates tunneling for new connections.
3. Tunneled clients are handled exclusively by the main thread. The tunneling implementation involves maintaining two client objects, upstream and downstream, which must be handled by the same thread. Currently, since the system doesn't support assigning two clients to the same I/O thread, tunneling sessions are restricted to the main thread. This is not expected to cause performance issues, as tunneling is lightweight and unlikely to become a bottleneck.


src/networking.c
```diff
@@ -5062,6 +5064,31 @@ void clientCommand(client *c) {
     addReplySubcommandSyntaxError(c);
 }
 
+/* TUNNEL <user> <protocol-version>
+ * Internal command that marks the connection as a tunnel, used to proxy client commands
+ * on behalf of the specified user and using the given RESP protocol version.
+ */
+void tunnelCommand(client *c) {
+    robj *clientname = c->argv[1];
```
murphyjacob4: Is argv[1] just a clientname? Or is this an ACL user to authenticate to? I see the comment says `<user>`.

If it is the latter, a couple notes:

1. This is just setting the client name (not the user the client is authenticated to)
2. Allowing TUNNEL to bypass authentication isn't secure. It basically is a free bypass of the entire authentication system. Especially so since TUNNEL needs to run on an unauthenticated incoming connection.

I'm thinking we could support something like this if we could guarantee that the TUNNEL command is only coming from the node that just used to be the primary. But just validating IP wouldn't be secure.

Perhaps before failover completes, we could have the primary send the TUNNEL commands over the replication link as part of the failover sequence. When the replica gets these, it could validate them since it can just check if it is coming in over the replication connection (which is already fully trusted as it is the current primary). If so, it could establish the tunneling connections to the primary, and automatically authenticate them to the specified user.

So something like:

1. Primary gets FAILOVER TO <node> TUNNEL
2. Primary pauses and enqueues all TUNNEL commands for current connections, with format: TUNNEL CREATE <client_id> <user> <protocol>
3. Primary flushes replication backlog
4. Replica creates new connections to the primary for each tunnel, authenticates the connection, and sends command with format: TUNNEL BEGIN <client_id>
5. Primary looks up client by ID and attaches the tunneling connections. If client no longer exists, we could just close the connection

Another option could be having the primary send something like TUNNEL SETTOKEN <random_secure_token> at the end, and then the replica could keep this in memory and ensure future TUNNEL commands include the right token. Although I have a feeling getting this one to be secure might be more difficult

WDYT? Maybe there is a simpler solution you can come up with

src/replication.c
```diff
@@ -5108,10 +5109,11 @@ void abortFailover(const char *err) {
...
     clearFailoverState();
+    server.tunnel_failover = 0;
 }
 
 /*
- * FAILOVER [TO <HOST> <PORT> [FORCE]] [ABORT] [TIMEOUT <timeout>]
+ * FAILOVER [TO <HOST> <PORT> [FORCE] [TUNNEL]] [ABORT] [TIMEOUT <timeout>]
  *
  * This command will coordinate a failover between the primary and one
  * of its replicas. The happy path contains the following steps:
```
murphyjacob4, line 5147: Can we add TUNNEL description to the command comment?

e.g. TUNNEL flag indicates that after the failover, this node will tunnel traffic of the existing connections to the new primary

src/server.c
```diff
@@ -4255,14 +4265,18 @@ int processCommand(client *c) {
              * and then resume the execution. */
             blockPostponeClient(c);
         } else {
+            c->duration = 0;
+            c->cmd->rejected_calls++;
```
murphyjacob4: Should a tunneled command be considered rejected?

src/tunnel.c
```diff
@@ -0,0 +1,276 @@
...
+}
+
+static void initTunnelPipe(tunnelPipe *pipe, connection *read_conn, connection *write_conn) {
+    pipe->read_conn = read_conn;
+    pipe->write_conn = write_conn;
+    pipe->buffer_pos = 0;
+    pipe->buffer_len = 0;
+}
+
+static ConnectionType *connTypeOfTunneling(void) {
```
murphyjacob4: Is connTypeOfTunneling ever going to be different than connTypeOfReplication?

src/tunnel.c
```diff
@@ -0,0 +1,276 @@
...
+    if (isReplyError(buf, buf_len)) {
+        abortTunnel(session);
+        return 0;
+    }
+    if (!isEntireReplyReceived(buf_len, expected_str_len)) return 0;
+
+    if (!isExpectedReply(buf, expected_str, expected_str_len)) {
+        abortTunnel(session);
+        return 0;
+    }
```
murphyjacob4: Supposing it returns a response that isn't an error but is shorter than expected, would the connection just be left hanging?

Instead of checking error followed by checking if it is expected, I think this may be simpler/less error prone:


```suggestion
    size_t prefix_size = buf_len > expected_str_len ? expected_str_len : buf_len;
    /* Check if what is in buf partially or fully matches expected_str */ 
    if (strncmp(buf, expected_str, prefix_size)) {
        /* What we have so far does not match what we expected */
        abortTunnel(session);
        return 0;
    }
    if (!isEntireReplyReceived(buf_len, expected_str_len)) return 0;
```

src/tunnel.c
```diff
@@ -0,0 +1,276 @@
...
+    return (memcmp(buf, str, len) == 0);
+}
+
+static size_t expectedReplyLength(int multi_cnt) {
+    size_t ret = ok_str_len + multi_cnt * queued_str_len;
+    return multi_cnt ? ret + ok_str_len : ret;
+}
+
+static int verifyReadReply(tunnelSession *session, const char *buf, size_t buf_offset, size_t buf_len, const char *expected_str, size_t expected_str_len) {
+    if (buf_offset >= buf_len) return 0;
```
murphyjacob4: Should this be a serverAssert?

src/tunnel.c
```diff
@@ -0,0 +1,276 @@
...
+    return resp;
+}
+
+static tunnelSession *createTunnelSession(client *c, char *host, int port) {
+    serverAssert(!c->tunnel_session);
+    ++server.stat_tunnel_sessions;
+    ++server.stat_active_tunnel_sessions;
+    tunnelSession *session = zmalloc(sizeof(tunnelSession));
+    connection *conn = connCreate(connTypeOfTunneling());
+    session->upstream_client = createClient(conn);
```
murphyjacob4: When we create a client, upstream_client->buf will already be preallocated with some memory for buffering. It would be more memory efficient if we could reuse this memory instead of relying on the pipe memory.

On the same note - the opposite direction could be stored in upstream_client->querybuf

WDYT?

tests/integration/replica-tunnel.tcl
```diff
@@ -0,0 +1,79 @@
+start_server {tags {needs:repl external:skip "quit"}} {
```
murphyjacob4, line 1: A few more that would be good:

1. Node A FAILOVER TO node B TUNNEL followed by Node B FAILOVER TO node A TUNNEL (while keeping tunnels alive)
2. Tunneled connection is authenticated to a user

src/networking.c
murphyjacob4: Maybe I'm overthinking it - I see you added it in the admin group. If we just ensure that we first call AUTH with the `primaryauth` password before we use TUNNEL, that should be good enough, since admin also gives you access to stuff like ACL SETUSER, so this would be in a similar vein.

yairgott: Yeah, I'll extend `TUNNEL` to send  the primary auth.

src/tunnel.c
yairgott: buf_offset is the expected buf offset that the expected_str should reside.  buf_len lower than buf_offset means that partial of the reply is read.

src/tunnel.c
yairgott: `c->querybuf` is used in some logics for handling traditional requests. I think that it would be safer to isolate the tunneling as much as possible. FWIW, the `c->querybuf` is freed at the beginning of the tunneling flow.

src/server.c
```diff
@@ -4054,6 +4057,12 @@ uint64_t getCommandFlags(client *c) {
     return cmd_flags;
 }
 
+static int shouldTunnelOnFailure(client *c) {
+    return (!server.cluster_enabled && c->id != CLIENT_ID_AOF && !c->flag.blocked &&
```
omanges: Should we also add a check to tunnel only `CLIENT_TYPE_NORMAL` type of connection and other should be skipped from tunneling ?

src/server.c
```diff
@@ -4347,7 +4361,7 @@ int processCommand(client *c) {
 
     /* Don't accept write commands if this is a read only replica. But
      * accept write commands if this is our primary. */
```
omanges: Should we update the text specifying that in-case of tunneling we accept both read and write operation even when the node is replica ?

src/server.c
omanges: Or maybe we can add check to skip tunneling in case the `lastcmd` was any of the following :- 
* info 
* role
* replconf
* ping

So that unnecessary commands tunneling is prevented also likely genuine clients will been selected for tunneling ?

yairgott: Checking for CLIENT_TYPE_NORMAL makes sense. Note that the commands you mentioned don't initiate the tunnel flow as those are not read or write commands, see: https://github.com/valkey-io/valkey/pull/2265/files#diff-1abc5651133d108c0c420d9411925373c711133e7748d9e4f4c97d5fb543fdd9R4244

src/server.c
```diff
@@ -5824,7 +5840,9 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "total_blocking_keys_on_nokey:%lu\r\n", blocking_keys_on_nokey,
                 "paused_reason:%s\r\n", paused_reason,
                 "paused_actions:%s\r\n", paused_actions,
-                "paused_timeout_milliseconds:%lld\r\n", paused_timeout));
+                "paused_timeout_milliseconds:%lld\r\n", paused_timeout,
+                "total_tunnel_sessions:%lld\r\n", server.stat_tunnel_sessions,
+                "active_tunnel_sessions:%lld\r\n", server.stat_active_tunnel_sessions));
```
madolson, line 5836: wouldn't a more appropriate name be tunnel clients? 

I also generally don't really like the term tunnel, these feel more like proxies than tunnels, but I don't feel that strongly about the name.

src/networking.c
```diff
@@ -5062,6 +5064,68 @@ void clientCommand(client *c) {
...
 }
 
+/* TUNNEL <protocol-version> [CLIENTNAME <name> USER <user>]
+ * Internal command used to propagate downstream client information to the
+ * upstream tunnel client. The first argument specifies the RESP protocol version
+ * used by the downstream client, followed by two optional parameters:
+ * - CLIENTNAME <name>: The name of the downstream client.
+ * - USER <user>: The ACL-authenticated username.
+ */
+void tunnelCommand(client *c) {
```
madolson: I don't really like this approach of a net new command that get's sent in the middle of the failover. There is more things that can go wrong around the like the replication not user having access to this command, there is a rather major concern if this command is accidentally allowed since it just allows bypassing authentication. There are also a lot of weird esoteric options that aren't supported here like pubsub or client tracking which is silently getting dropped during the handover phase. 

I am more of a fan of just only supporting tunneling on net new connections, and starting the tunneling before or after the failover to start redirecting traffic. I believe that should have the same impact on availability. Afaik in this implementation, at some point you still need to terminate the old primary and force the reconnects. 

You can still do all the same logic of tunneling requests, but have it only apply on new connections. You proxy from the new primary to the old primary, and then once DNS or whatever is propagated, you can do the failover then slowly drain the proxied clients. You will have more disconnects, but it will transfer 100% of the state.

src/server.c
yairgott: > wouldn't a more appropriate name be tunnel clients?

Makes sense.

> I also generally don't really like the term tunnel, these feel more like proxies than tunnels, but I don't feel that strongly about the name.

IMO tunnel is the better term here because it simply forwards data without inspecting or parsing. Essentially, it reads from downstream and writes upstream.

src/networking.c
yairgott: > I am more of a fan of just only supporting tunneling on net new connections,

The main question is how to avoid tunneling new control plane connections which are designated to the demoted primary. I’m thinking we could extend the `FAILOVER` `TUNNEL` option to accept an IP range specifying which new connections are **excluded** to be tunneled.  WDYT?

sandeepk97:
@yairgott any update on when this change will be ready to use?

sandeepkunusoth:
Thanks for adding this feature. @yairgott any update on when this change will be ready to use?

yairgott:
> Thanks for adding this feature. @yairgott any update on when this change will be ready to use?

I want to apply some changes to make it more production ready. 

@madolson can you please advise if this can got into 9.0 version?

src/networking.c
madolson: > I’m thinking we could extend the FAILOVER TUNNEL option to accept an IP range specifying which new connections are excluded to be tunneled

Yeah. Internally we have the concept of "admin" connections, which are immune to our similar functionality. Admin is basically local/unix for us, but I don't see any reason why we can't extend that to an IP port.

madolson:
> @madolson can you please advise if this can got into 9.0 version?

I feel like you've been here long enough to learn we don't move that quickly on stuff :) Our code cutoff for new features for 9.0 was last week, you can find the future schedule: https://github.com/valkey-io/valkey/wiki/Upcoming-release-schedule.

PR #2269, opened 2025-06-25
Check module NO_FAILOVER flag before calling clusterHandleReplicaFailover

enjoy-binbin:
In other places where clusterHandleReplicaFailover is called, the
CLUSTER_MODULE_FLAG_NO_FAILOVER flag is checked, which is not done
here. It bothered me a lot sometimes, did a minor cleanup to make
them consistent, and update some outdated comments.

This should be no harm since the clusterHandleReplicaFailover call
in clusterCron is the main place that drives the failover forward,
and it had the NO_FAILOVER flag check.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -5270,7 +5270,7 @@ void manualFailoverCheckTimeout(void) {
     }
 }
 
-/* This function is called from the cluster cron function in order to go
+/* This function is called from clusterCron or clusterBeforeSleep in order to go
  * forward with a manual failover state machine. */
 void clusterHandleManualFailover(void) {
     /* Return ASAP if no manual failover is in progress. */
@@ -5585,9 +5585,13 @@ void clusterBeforeSleep(void) {
             if (!(server.cluster_module_flags & CLUSTER_MODULE_FLAG_NO_FAILOVER)) clusterHandleReplicaFailover();
         }
     } else if (flags & CLUSTER_TODO_HANDLE_FAILOVER) {
-        /* Handle failover, this is needed when it is likely that there is already
+        /* Handle failover as soon as possible so that won't have a 100ms
+         * as it was handled only in clusterCron. This is needed when it
+         * is likely that we can start the election or there is already
          * the quorum from primaries in order to react fast. */
-        clusterHandleReplicaFailover();
+        if (nodeIsReplica(myself)) {
+            if (!(server.cluster_module_flags & CLUSTER_MODULE_FLAG_NO_FAILOVER)) clusterHandleReplicaFailover();
+        }
     }
 
     /* Update the cluster state. */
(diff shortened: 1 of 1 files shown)


src/cluster_legacy.c
```diff
@@ -5585,9 +5585,13 @@ void clusterBeforeSleep(void) {
...
         }
     } else if (flags & CLUSTER_TODO_HANDLE_FAILOVER) {
-        /* Handle failover, this is needed when it is likely that there is already
+        /* Handle failover as soon as possible so that won't have a 100ms
+         * as it was handled only in clusterCron. This is needed when it
+         * is likely that we can start the election or there is already
          * the quorum from primaries in order to react fast. */
-        clusterHandleReplicaFailover();
+        if (nodeIsReplica(myself)) {
+            if (!(server.cluster_module_flags & CLUSTER_MODULE_FLAG_NO_FAILOVER)) clusterHandleReplicaFailover();
```
sarthakaggarwal97: unrelated to PR but just curiosity, but what is the general usecase of this flag? Like why would modules avoid failovers?

sarthakaggarwal97 (approved):

src/cluster_legacy.c
enjoy-binbin: Some module may use its own failover mechanism? Though i haven't seen those modules.

sarthakaggarwal97: yeah, haven't checked why it was introduced, but clustering felt like a valkey core thing, so was intrigued why did we expose this.

zuiderkwast: Why nested if? One if is enough.

```suggestion
        if (nodeIsReplica(myself) &&
            !(server.cluster_module_flags & CLUSTER_MODULE_FLAG_NO_FAILOVER)) {
            clusterHandleReplicaFailover();
```

enjoy-binbin: yean, that is right. I just copied the above content (you can expand it to see it).

I previously wrote a simpler IF block that combined HANDLE_MANUALFAILOVER and HANDLE_FAILOVER, but later changed it to the current one.

zuiderkwast (approved):
Is it possible to implement another failover mechanism in modules?

Is it possible to replace the cluster bus completely with another logic to create another cluster implementation in a module? I think the module API is not complete for this but it's an interesting idea.

src/cluster_legacy.c
zuiderkwast: Yes I see it can be simplified, but we don't have to change too much. As you want.

enjoy-binbin:
I see the comment (or doc) is
```
/* Set Cluster flags in order to change the normal behavior of
 * Cluster, especially with the goal of disabling certain functions.
 * This is useful for modules that use the Cluster API in order to create
 * a different distributed system, but still want to use the Cluster
 * message bus. Flags that can be set:
 *
 * * CLUSTER_MODULE_FLAG_NO_FAILOVER
 * * CLUSTER_MODULE_FLAG_NO_REDIRECTION
 *
 * With the following effects:
 *
 * * NO_FAILOVER: prevent Cluster replicas from failing over a dead primary.
 *                Also disables the replica migration feature.
 *
 * * NO_REDIRECTION: Every node will accept any key, without trying to perform
 *                   partitioning according to the Cluster algorithm.
 *                   Slots information will still be propagated across the
 *                   cluster, but without effect. */
```

They can all be options in the future i guess

PR #2275, opened 2025-06-26
Add keyspace-hits and keyspace-misses metrics under CLUSTER SLOT-STATS command

enjoy-binbin:
Adds two new metrics for per-slot statistics, it is the same meaning
as keyspace_hits and keyspace_misses in the INFO STATS command, but
for the single slot.

The new metrics are:
- keyspace-hits: Number of successful lookup of keys in the slot.
- keyspace-misses: Number of failed lookup of keys in the slot.


madolson:
Is there a specific use case for this? Seems like it could be useful, but it's a lot of extra memory to keep track of which might impact the CPU cache hit rates for commands since we're evicting memory.

enjoy-binbin:
It can track hot slots i guess? We haven't actually used it yet, but it seems like it would be useful to add it to the slot metrics.

madolson:
> It can track hot slots i guess? We haven't actually used it yet, but it seems like it would be useful to add it to the slot metrics.

In AWS we originally had planned to support scaling off of cache hit rate, but in reality the data was too noisy and it never went anywhere and we still don't do it. Maybe it works for some workloads?

enjoy-binbin:
I haven't really measured it, if it's a problem, maybe we can also put it under cluster-slot-stats-enabled if we want to add it somedays.

PR #2279, opened 2025-06-26
The smaller config epoch primary will become the replica when two primaries in the same shard

### LINKED ISSUE #2261: [BUG] Two primaries in same shard
**Describe the bug**

Bug where a shard ends up with 2 primaries and when `cluster-allow-replica-migration` is disabled.

**To reproduce**

Let’s say a shard has a primary(A) and a replica(B). Also let's assume that [cluster-allow-replica-migration](https://github.com/valkey-io/valkey/blob/8.1/src/cluster_legacy.c#L6873) is disabled.

1. Node A goes down, and Node B takes over primaryship.
2. Node A continues to be down while another Node C is added as a replica of B.
3. Node B goes down, and Node C takes over primaryship.
4. Node A and Node B come back up and start learning about the topology.
5. Node A comes up thinking it was the primary (but has an older config epoch compared to C).
6. Node A learns about Node C via gossip and assigns it a random `shard_id`.
7. Node A receives a direct ping from Node C.
    a. Node C advertises the same set of slots that Node A was earlier owning.
    b. Since Node A assigns a random shard ID to Node C, Node A thinks that it is still a primary and it lost all its slots to Node C, which is in another shard.
8. Node A then updates the actual `shard_id` of Node C while processing `shard_id` in ping extensions.
9. Node A and Node C end up being primaries in the same shard while Node C continues to own slots.

**Expected behavior**

I would have expected node A to become a replica of node C after learning that node C is in the same shard.

**Potential fix**

Currently, when a node A receives a ping from node C, it [first processes the slots config](https://github.com/valkey-io/valkey/blob/8.1/src/cluster_legacy.c#L3652) and [then processes the shard_id in the ping extension](https://github.com/valkey-io/valkey/blob/8.1/src/cluster_legacy.c#L3701). One way to fix this could be to process the shard_id or all the ping extensions and then later update slots configuration.

enjoy-binbin:
In #2261, there is a scenario that results in two primaries in a shard.

Let's say a shard has a primary(A) and a replica(B). Also let's assume that
cluster-allow-replica-migration is disabled.

1. Node A goes down, and Node B takes over primaryship.
2. Node A continues to be down while another Node C is added as a replica of B.
3. Node B goes down, and Node C takes over primaryship.
4. Node A and Node B come back up and start learning about the topology.
5. Node A comes up thinking it was the primary (but has an older config epoch compared to C).
6. Node A learns about Node C via gossip and assigns it a random `shard_id`.
7. Node A receives a direct ping from Node C.
    a. Node C advertises the same set of slots that Node A was earlier owning.
    b. Since Node A assigns a random shard ID to Node C, Node A thinks that it is still a
       primary and it lost all its slots to Node C, which is in another shard.
8. Node A then updates the actual `shard_id` of Node C while processing `shard_id` in ping extensions.
9. Node A and Node C end up being primaries in the same shard while Node C continues to own slots.

So in the step 8, Node A and Node C has the same shard_id and both report themselves
as a primary, but Node C has a bigger config epoch, i guess in this case, we can try
to make Node A become a replica of Node C.

Fixes #2261.

An ASCII seq diagram:
```
+------+          +------+          +------+
|Node A|          |Node B|          |Node C|
+------+          +------+          +------+
   |                 |                 |
   | <Primary>       | <Replica>       |
   |                 |                 |
   X (fails)         |                 |
                     |---------------->|
                     |  (promoted to primary)
                     |                 |
                     X (fails)         |
                                       |
                                       |---------->
                                       | (promoted to primary)
   | (rejoins)                         |
   | (no extensions shared)            |
   |---------------------------------->|
   | (assigns random shard_id)|
   |<----------------------------------|
   | (direct ping: slot info)          |
   | Updates C’s shard_id correctly    |
   | Detects slot conflict             |
   |                                  
   |                                  
+-----------------------------------------------+
| Outcome: Nodes A & C both see themselves as   |
| primary temporarily until shard_id is updated |
| and converged correctly based on epoch value  |
+-----------------------------------------------+
```


tests/unit/cluster/shardid-propagation.tcl
```diff
@@ -53,3 +53,84 @@ tags {tls:skip external:skip cluster singledb} {
...
+        R $A debug close-cluster-link-on-packet-drop 1
+        R $A debug drop-cluster-packet-filter $CLUSTER_PACKET_TYPE_ALL
+
+        # Wait for B to become a primary.
+        pause_process [srv -$A pid]
+        R $B cluster failover force
+        wait_for_condition 1000 50 {
+            [s -$B role] eq {master}
+        } else {
+            fail "Failover B does not happened"
```
deepakrn: nit - "Failover to B has not happened" or "Failed waiting for B to takeover primaryship" instead?

tests/unit/cluster/shardid-propagation.tcl
```diff
@@ -53,3 +53,84 @@ tags {tls:skip external:skip cluster singledb} {
...
+        }
+
+        # Make sure A and B become the replica.
+        R $A debug close-cluster-link-on-packet-drop 0
+        R $A debug drop-cluster-packet-filter $CLUSTER_PACKET_TYPE_NONE
+        wait_for_condition 1000 50 {
+            [s -$A role] eq {slave} &&
+            [s -$B role] eq {slave}
+        } else {
+            fail "Failover does not happened"
```
deepakrn: "Failed waiting for A and B to become replicas" instead? (Similarly on line 120, 111 etc) 

"Failover does not happened" doesn't sound grammatically correct.

src/cluster_legacy.c
```diff
@@ -3091,6 +3091,20 @@ void clusterProcessPingExtensions(clusterMsg *hdr, clusterLink *link) {
...
+     * Since the processing order is random, we may end up with multiple primary nodes
+     * in a shard, in this case, if the sender's config epoch is greater, we configure
+     * myself as its replica. */
+    if (nodeIsPrimary(myself) && nodeIsPrimary(sender) && areInSameShard(myself, sender) &&
+        nodeEpoch(sender) > nodeEpoch(myself)) {
+        serverLog(LL_NOTICE, "Two primaries in same shard, and the sender has a greater config epoch. "
+                             "Reconfiguring myself as a replica of %.40s (%s)",
+                  sender->name, sender->human_nodename);
+         clusterSetPrimary(sender, 1, 0);
+         clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_UPDATE_STATE | CLUSTER_TODO_FSYNC_CONFIG);
```
hpatro: Should we place this logic inside `clusterProcessPacket` method?

src/cluster_legacy.c
```diff
@@ -3091,6 +3091,20 @@ void clusterProcessPingExtensions(clusterMsg *hdr, clusterLink *link) {
     if (ext_shardid == NULL) ext_shardid = clusterNodeGetPrimary(sender)->shard_id;
 
     updateShardId(sender, ext_shardid);
+
+    /* In some cases, we may first learn about other nodes through gossip (assigning a
+     * random shard_id), and then learn the sender's shard_id through ping extensions.
```
```suggestion
    /* In some cases, we might first learn about other nodes through gossip and assign them a
     * random shard_id, and then further learn the actual shard_id via a direct ping/pong message through ping extensions.
```

src/cluster_legacy.c
hpatro: Would prefer creating an independent method which has the single responsibility of dealing with nodes in the same shard with conflicting epoch value and we deal with such conditions within that. WDYT @enjoy-binbin ?

hpatro (commented):
Overall, the fix looks good to me.

Liked the GPT generated (slightly modified) ASCII seq diagram for this, pasting it here for others as well.

```
+------+          +------+          +------+
|Node A|          |Node B|          |Node C|
+------+          +------+          +------+
   |                 |                 |
   | <Primary>       | <Replica>       |
   |                 |                 |
   X (fails)         |                 |
                     |---------------->|
                     |  (promoted to primary)
                     |                 |
                     X (fails)         |
                                       |
                                       |---------->
                                       | (promoted to primary)
   | (rejoins)                         |
   | (no extensions shared)            |
   |---------------------------------->|
   | (assigns random shard_id)|
   |<----------------------------------|
   | (direct ping: slot info)          |
   | Updates C’s shard_id correctly    |
   | Detects slot conflict             |
   |                                  
   |                                  
+-----------------------------------------------+
| Outcome: Nodes A & C both see themselves as   |
| primary temporarily until shard_id is updated |
| and converged correctly based on epoch value  |
+-----------------------------------------------+

```

src/cluster_legacy.c
enjoy-binbin: I did struggle a bit with the placement, I actually would have preferred it to be here.
```diff
        /* Get info from the gossip section */
        if (sender) {
            clusterProcessGossipSection(hdr, link);
            clusterProcessPingExtensions(hdr, link);
+            afterProcessxxx();
        }
```

Yes, we should place it inside the clusterProcessPacket. I see what you means, we are in the same page actually. I can do something like clusterHandleConfigEpochCollision

tests/unit/cluster/shardid-propagation.tcl
enjoy-binbin: right... sorry about that, i am suck with the gramma.

enjoy-binbin:
I see the test `New replica inherits multiple migrating slots` is failing:

```
47461:M 28 Jun 2025 12:18:24.199 * Two primaries in same shard, and the sender has a greater config epoch. Reconfiguring myself as a replica of 8af67730a16d4a47a947271cca28a03ef9fdf337 ()
cluster_legacy.c:5995 'myself->numslots == 0' is not true
```

myself cluster nodes:
````
d2081b715fb4b174127ec7a99de4e180de345fbd 127.0.0.1:21120@31120,,tls-port=0,shard-id=9d3ebac222cc295191331bd3dbba06e607b42019 master - 0 1751084304199 3 connected 10923-16383
ee0d3a2618bf64993ce1e71569f845aa6e33f46b 127.0.0.1:21121@31121,,tls-port=0,shard-id=d3f24d198dac6233c89454ed22f1a3a85b1e0e9f master - 0 1751084304199 2 connected 5462-10922
8af67730a16d4a47a947271cca28a03ef9fdf337 127.0.0.1:21119@31119,,tls-port=0,shard-id=d5ada9160b2d0a5196dcdfe0325f6accf187dec0 master - 0 1751084304199 4 connected
21fe04bcaaf3eb6dfb6217e27b3100dccb0993e4 127.0.0.1:21118@31118,,tls-port=0,shard-id=d3f24d198dac6233c89454ed22f1a3a85b1e0e9f slave ee0d3a2618bf64993ce1e71569f845aa6e33f46b 0 1751084304199 2 connected
1497ec4b6d1ecdc419741306f8d1ae44d9e8fc9d 127.0.0.1:21117@31117,,tls-port=0,shard-id=9d3ebac222cc295191331bd3dbba06e607b42019 slave d2081b715fb4b174127ec7a99de4e180de345fbd 0 1751084304199 3 connected
7c0a9f9b0610f3a3c4d895282ed41f51f9f97a5d 127.0.0.1:21122@31122,,tls-port=0,shard-id=d5ada9160b2d0a5196dcdfe0325f6accf187dec0 myself,master - 0 0 1 connected 0-5461 [7->-ee0d3a2618bf64993ce1e71569f845aa6e33f46b] [13->-ee0d3a2618bf64993ce1e71569f845aa6e33f46b] [17->-ee0d3a2618bf64993ce1e71569f845aa6e33f46b]
```

I will take a look later this day.

enjoy-binbin:
The problem seems to be, a primary used to had a larger config epoch, and then it became a replica, and then it update the shard_id, and then it doing the cluster reset soft and became a empty primary, but remain the larger config epoch and its shard_id, and causing this problem. IT is also the same issue, two primaries in the same shard, but the empty primary had a larger config epoch.

See #https://github.com/valkey-io/valkey/pull/2283. I think when a replica doing the CLUSTER RESET SOFT, we should generate a new shard_id for it.

tests/unit/cluster/shardid-propagation.tcl
```diff
@@ -53,3 +53,84 @@ tags {tls:skip external:skip cluster singledb} {
...
+# 4. Node A and Node B come back up and start learning about the topology.
+# 5. Node A comes up thinking it was the primary (but has an older config epoch compared to C).
+# 6. Node A learns about Node C via gossip and assigns it a random `shard_id`.
+# 7. Node A receives a direct ping from Node C.
+#    a. Node C advertises the same set of slots that Node A was earlier owning.
+#    b. Since Node A assigns a random shard ID to Node C, Node A thinks that it is still a primary and
+#       it lost all its slots to Node C, which is in another shard.
+# 8. Node A then updates the actual `shard_id` of Node C while processing `shard_id` in ping extensions.
+# 9. Node A and Node C end up being primaries in the same shard while Node C continues to own slots.
+start_cluster 3 2 {tags {external:skip cluster}} {
```
zuiderkwast, line 86: We can put some config we want for all nodes here.

```suggestion
start_cluster 3 2 {tags {external:skip cluster} overrides {allow-replica-migration no}} {
```

zuiderkwast:
> and then it doing the cluster reset soft and became a empty primary, but remain the larger config epoch and its shard_id, and causing this problem

This test case doesn's use CLUSTER RESET SOFT. ❓

zuiderkwast (commented):
LGTM in general.

enjoy-binbin:
> This test case doesn's use CLUSTER RESET SOFT. ❓

This is the other test case, i handle it in #2283
```
    test "New replica inherits multiple migrating slots" {
        # Reset R3 to turn it into an empty node
        assert_equal {OK} [R 3 CLUSTER RESET]
        # Add R3 back as a replica of R0
        assert_equal {OK} [R 3 CLUSTER MEET [srv 0 "host"] [srv 0 "port"]]
        wait_for_role 0 master
        assert_equal {OK} [R 3 CLUSTER REPLICATE $R0_id]
        wait_for_role 3 slave
        # Validate final states
        wait_for_slot_state 3 "\[7->-$R1_id\] \[13->-$R1_id\] \[17->-$R1_id\]"
    }
```

src/cluster_legacy.c
```diff
@@ -3797,6 +3809,15 @@ int clusterProcessPacket(clusterLink *link) {
...
             clusterProcessPingExtensions(hdr, link);
         }
+
+        /* If after processing everything, we find that myself and sender are on the
+         * same shard and are both primaries, if myself config epoch is smaller, make
+         * it a replica. */
+        if (sender && nodeIsPrimary(myself) && nodeIsPrimary(sender) && areInSameShard(myself, sender) &&
+            nodeEpoch(sender) > nodeEpoch(myself)) {
+            clusterHandlePrimariesSameShardCollision(sender);
+        }
```
enjoy-binbin, line 4422: The new test added in ##2283 now currently will fail in here:
```
86095:M 05 Jul 2025 12:57:44.202 # === ASSERTION FAILED ===
86095:M 05 Jul 2025 12:57:44.202 # ==> cluster_legacy.c:6020 'myself->numslots == 0' is not true

------ STACK TRACE ------

Backtrace:
0   valkey-server                       0x000000010227faac updateShardId + 0
1   valkey-server                       0x000000010227d12c clusterReadHandler + 7736
2   valkey-server                       0x00000001022f8040 connSocketEventHandler + 180
3   valkey-server                       0x00000001021bedd8 aeProcessEvents + 372
4   valkey-server                       0x00000001021e8f50 main + 26864
5   dyld                                0x0000000180b13154 start + 2476
```

That is because we drop the link->support_extension in here https://github.com/valkey-io/valkey/pull/2283#issuecomment-3031446204

The shard_id was not updated in time, which led to a misjudgment. So we either need to support link->support_extension or add node->numslots judgment here.

tests/unit/cluster/shardid-propagation.tcl
enjoy-binbin: yes, we can, here i am only explicitly setting for R A (i will apply this change later)

hpatro:
A bit too late to include this change to 9.0 GA. Let's target it for 9.1

src/cluster_legacy.c
```diff
@@ -3904,6 +3919,15 @@ int clusterProcessPacket(clusterLink *link) {
...
     }
 
+    /* We try to process extensions first, as we become more and more
+     * dependent on extensions. */
+    /* clusterProcessPingExtensions -> updateShardId may rely on node flag, but
+     * in here, we don't have the right node flag.
+    if (sender && (type == CLUSTERMSG_TYPE_PING || type == CLUSTERMSG_TYPE_PONG || type == CLUSTERMSG_TYPE_MEET)) {
+        clusterProcessPingExtensions(hdr, link);
+    }
+    */
```
zuiderkwast, line 3929: This is commented out. Why?

If we process the ping extensions first, we will update the sender's shard-id first and handle role and slot ownerships later. If we do that, myself will step down to replica?

@valkey-review-bot What do you think of this PR's solution and compared to this alternative approach of handling ping extensions first?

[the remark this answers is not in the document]
zuiderkwast: Would it be better to handle some things from the header first (such as role and most other things except slot ownership), then ping extensions, and last slot ownership?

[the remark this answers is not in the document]
zuiderkwast: @valkey-review-bot The follow-up question is for you. Alternatively, we could handle only the shard-id ping extension right after updating the role, and handle other ping extensions last. What do you think of that approach?

enjoy-binbin: This PR is a bit old, i've somewhat forgotten the details. We might need to address https://github.com/valkey-io/valkey/pull/2989#issuecomment-4222379039 first; there is more recent context there.

PR #2280, opened 2025-06-27
Use objectGetExpire instead of getExpire to reduce expire hashtable lookup

enjoy-binbin:
Since we have now embedded the expire TTL into robj, if we already
have the robj o, we can directly get the expire. This saves an expire
hashtable lookup in getExpire.

### DIFF
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -544,7 +544,7 @@ try_again:
     /* Create RESTORE payload and generate the protocol to call the command. */
     for (j = 0; j < num_keys; j++) {
         long long ttl = 0;
-        long long expireat = getExpire(c->db, kv[j]);
+        long long expireat = objectGetExpire(ov[j]);
 
         if (expireat != -1) {
             ttl = expireat - commandTimeSnapshot();
--- a/src/db.c
+++ b/src/db.c
@@ -1402,7 +1402,7 @@ void renameGenericCommand(client *c, int nx) {
     }
 
     incrRefCount(o);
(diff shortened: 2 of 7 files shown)


enjoy-binbin:
We may need to add a new module API in the future to get the expire time based on robj
```
mstime_t VM_GetExpire(ValkeyModuleKey *key) {
    mstime_t expire = getExpire(key->db, key->key);
    if (expire == -1 || key->value == NULL) return VALKEYMODULE_NO_EXPIRE;
    expire -= commandTimeSnapshot();
    return expire >= 0 ? expire : 0;
}
```

zuiderkwast (commented):
Yes, this is a good optimization. It's a follow-up from when we introduced storing expire inside robj in #1186.

I think we can also do it in module.c. The code you mentioned has a ValkeyModuleKey. It has a pointer to the value.

```diff
 mstime_t VM_GetExpire(ValkeyModuleKey *key) {
-    mstime_t expire = getExpire(key->db, key->key);
-    if (expire == -1 || key->value == NULL) return VALKEYMODULE_NO_EXPIRE;
+    mstime_t expire = key->value ? objectGetExpire(key->value) : -1;
+    if (expire == -1) return VALKEYMODULE_NO_EXPIRE;
     expire -= commandTimeSnapshot();
     return expire >= 0 ? expire : 0;
 }
```

zuiderkwast (approved):
The macos build reports a leak. Not from this PR, probably.

PR #2283, opened 2025-06-28
Generate a new shard_id when the replica executes CLUSTER RESET SOFT

enjoy-binbin:
When a replica doing the CLUSTER RESET SOFT, we should generate a
new shard_id for it. Since if the node was a replica, it inherits
the shard_id of its primary, and after the CLUSTER RESET SOFT, the
replica become a new empty primary, it should has its own shard_id.

Otherwise, when it joins back into the cluster or responds the PONG,
we will have two primaries in the same shard.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -1023,7 +1023,7 @@ void clusterUpdateMyselfFlags(void) {
     int nofailover = server.cluster_replica_no_failover ? CLUSTER_NODE_NOFAILOVER : 0;
     myself->flags &= ~CLUSTER_NODE_NOFAILOVER;
     myself->flags |= nofailover;
-    myself->flags |= CLUSTER_NODE_EXTENSIONS_SUPPORTED | CLUSTER_NODE_LIGHT_HDR_PUBLISH_SUPPORTED | CLUSTER_NODE_LIGHT_HDR_MODULE_SUPPORTED;
+    myself->flags |= CLUSTER_NODE_LIGHT_HDR_PUBLISH_SUPPORTED | CLUSTER_NODE_LIGHT_HDR_MODULE_SUPPORTED;
     if (myself->flags != oldflags) {
         clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_UPDATE_STATE);
     }
@@ -1493,7 +1493,6 @@ clusterLink *createClusterLink(clusterNode *node) {
     if (!link->inbound) {
         node->link = link;
     }
-    link->support_extension = 0;
     return link;
 }
 
@@ -3253,7 +3252,6 @@ int clusterIsValidPacket(clusterLink *link) {
                 explen += extlen;
                 ext = getNextPingExt(ext);
             }
-            link->support_extension = 1;
         }
     } else if (type == CLUSTERMSG_TYPE_FAIL) {
         explen = sizeof(clusterMsg) - sizeof(union clusterMsgData);
@@ -4271,7 +4269,7 @@ void clusterSendPing(clusterLink *link, int type) {
     estlen += (sizeof(clusterMsgDataGossip) * (wanted + pfail_wanted));
     /* If link supports it or the node indicates that it supports it, then we
      * pass the extensions. */
-    if (link->support_extension || (link->node && nodeSupportsExtensions(link->node))) {
+    if (link->node && nodeSupportsExtensions(link->node)) {
         estlen += writePingExtensions(NULL, 0);
     }
     /* Note: clusterBuildMessageHdr() expects the buffer to be always at least
@@ -4347,7 +4345,7 @@ void clusterSendPing(clusterLink *link, int type) {
     /* Compute the actual total length and send! */
     uint32_t totlen = 0;
 
-    if (link->support_extension || (link->node && nodeSupportsExtensions(link->node))) {
+    if (link->node && nodeSupportsExtensions(link->node)) {
         totlen += writePingExtensions(hdr, gossipcount);
     } else {
         serverLog(LL_DEBUG, "Unable to send extensions data, however setting ext data flag to true");
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3792,9 +3792,6 @@ int clusterProcessPacket(clusterLink *link) {
             clusterProcessGossipSection(hdr, link);
             clusterProcessPingExtensions(hdr, link);
         }
-
-        /* Make sure we don't have two primaries in the same shard. */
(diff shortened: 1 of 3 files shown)


src/cluster_legacy.c
```diff
@@ -3791,6 +3792,9 @@ int clusterProcessPacket(clusterLink *link) {
             clusterProcessGossipSection(hdr, link);
             clusterProcessPingExtensions(hdr, link);
         }
+
+        /* Make sure we don't have two primaries in the same shard. */
+        debugServerAssert(!(sender && nodeIsPrimary(sender) && nodeIsPrimary(myself) && areInSameShard(sender, myself)));
```
enjoy-binbin: Since there are a lot of failing tests, i am going to remove it for now.
```suggestion
```

src/cluster_legacy.h
```diff
@@ -38,6 +38,7 @@ typedef struct clusterLink {
     size_t rcvbuf_alloc;                   /* Allocated size of rcvbuf */
     clusterNode *node;                     /* Node related to this link. Initialized to NULL when unknown */
     int inbound;                           /* 1 if this link is an inbound link accepted from the related node */
+    int support_extension;                 /* 1 if this link is support extension. */
```
hpatro:
```suggestion
    int support_extension;                 /* 1 if this link supports passing extensions. */

File: src/cluster_legacy.c [hunk 3]
```diff
@@ -3461,9 +3465,6 @@ int clusterProcessPacket(clusterLink *link) {
                     memcpy(node->ip, ip, sizeof(ip));
                     getClientPortFromClusterMsg(hdr, &node->tls_port, &node->tcp_port);
                     node->cport = ntohs(hdr->cport);
-                    if (hdr->mflags[0] & CLUSTERMSG_FLAG0_EXT_DATA) {
-                        node->flags |= CLUSTER_NODE_EXTENSIONS_SUPPORTED;
-                    }
```
hpatro: This trick was used to propagate extension support as part of the node flags in the gossip section. By moving the abstraction to link level, we would need direct message between two nodes to be aware about the extension support.

So, I think we should leave it as is.

hpatro (commented):
Agree with the change around shard id generation. But would like the extension flag marking part left as is.

src/cluster_legacy.c
enjoy-binbin: ok, i see the point, i can keep it.

Do we need to keep link->support_extension? Without this link flag, if myself dont know the sender (like in this case, we do a cluster reset), the pong does not carry the extension, which results in the shard_id not being updated in time (but other flags like primary flag will be updated)
```
        /* Anyway reply with a PONG */
        clusterSendPing(link, CLUSTERMSG_TYPE_PONG);
```

enjoy-binbin: I pushed a new commit, do you like the current code?

src/cluster_legacy.c
```diff
@@ -1023,7 +1023,7 @@ void clusterUpdateMyselfFlags(void) {
     int nofailover = server.cluster_replica_no_failover ? CLUSTER_NODE_NOFAILOVER : 0;
     myself->flags &= ~CLUSTER_NODE_NOFAILOVER;
     myself->flags |= nofailover;
-    myself->flags |= CLUSTER_NODE_LIGHT_HDR_PUBLISH_SUPPORTED | CLUSTER_NODE_LIGHT_HDR_MODULE_SUPPORTED;
+    myself->flags |= CLUSTER_NODE_EXTENSIONS_SUPPORTED | CLUSTER_NODE_LIGHT_HDR_PUBLISH_SUPPORTED | CLUSTER_NODE_LIGHT_HDR_MODULE_SUPPORTED;
```
enjoy-binbin: This can be added or not, ours old versions does not rely on this flag, we used mflags[0].

src/cluster_legacy.c
enjoy-binbin: I am thinking if we should replace all this, or find a way to put this logic into getNodeFromLinkAndMsg or setClusterNodeToInboundClusterLink. I didn't end up doing this (we can clean it up in another PR in the future if needed)
```diff
-    if (sender && (hdr->mflags[0] & CLUSTERMSG_FLAG0_EXT_DATA)) {
+    if (sender && link->support_extension) {
        sender->flags |= CLUSTER_NODE_EXTENSIONS_SUPPORTED;
    }
```

src/cluster_legacy.c
hpatro: I think it's safe to add this. All new nodes should support extension by default.

src/cluster_legacy.c
hpatro: I think we should restrict this PR change to generating new shard and let the shard id be eventually propagated and introduce the `link->extension`/perform the refactoring in separate PR ? WDYT ?

enjoy-binbin: SGTM, i will make the change later

src/cluster_legacy.c
enjoy-binbin: i will move it to another PR.

enjoy-binbin:
@hpatro and I discussed and decided to temporarily remove link->support_extension for now. After this merge, we will discuss it in another new PR.

The new link->extension is aim for this case:
1. Node B cluster reset soft, and remove all the nodes in its cluster_nodes_dict
2. Node A sends the PING, and Node B will reply a PONG even it does not know A, and in here, since Node B does not know Node A, it won't send the extension. But the link has mf_flags that we all know Node A can handle extension. So in these packets, Node A will know B change the role but not able to learn the shard-id

That is something like that. So if we don't have link->extension, Node B needs to wait until the cluster meets  (I remember we have a compensation mechanism that will allow node B to eventually meet back into the cluster.) again before sending the extension

with link->support_extension time:
```
Testing unit/cluster/shardid-propagation
[ok]: The replica will have a new shard_id after cluster reset soft (65 ms)
[ok]: Check for memory leaks (pid 51984) (587 ms)
[ok]: Check for memory leaks (pid 51954) (312 ms)
[ok]: Check for memory leaks (pid 51924) (313 ms)
[ok]: Check for memory leaks (pid 51881) (293 ms)
[1/1 done]: unit/cluster/shardid-propagation (13 seconds)
```

without the link->support_extension
```
[ok]: The replica will have a new shard_id after cluster reset soft (3102 ms)
[ok]: Check for memory leaks (pid 54164) (301 ms)
[ok]: Check for memory leaks (pid 54127) (291 ms)
[ok]: Check for memory leaks (pid 54103) (294 ms)
[ok]: Check for memory leaks (pid 54050) (292 ms)
[1/1 done]: unit/cluster/shardid-propagation (16 seconds)
```

hpatro (approved):

PR #2286, opened 2025-06-30
Include command fullname in error message when returning NO_MULTI error

### LINKED ISSUE #1629: [BUG] Breaking Change to error handling in MULTI blocks
**Describe the bug**

During my tests of migrating from Redis (more precisely KeyDB) to Valkey, I found an undocumented change to the handling of errors within a `MULTI` block, which prevents Valkey to be used as a drop-in replacement. While there is a documentation about transactions in the Valkey Readme, I couldn't find any information about this change (and it took me days of debugging).

Behavior in Redis (and KeyDB): commands within a `MULTI` block are executed, even if there are errors in the chain
Behavior in Valkey: all commands are aborted, as soon as there is one error in the chain.

Please note, that personally, I am absolutely in favor of failing fast. So for me this behavior is much more desirable than before, because it pointed me to an error in our application logic. It's up to you to decide, whether you want this behavior only to apply with the strong compatibility mode.

But please at least mention this breaking change in the docs, as it would have saved me (and maybe others) days of searching and debugging.

**To reproduce**

* run a current Valkey image: `docker run -it --rm --detach 'valkey/valkey:8.0.2-alpine3.21'`
* note the ID of the started container
* connect to the started container: `docker exec -it <id-of-container> /bin/sh`
* connect via CLI: `valkey-cli`
* run these commands:
```
127.0.0.1:6379> MULTI
OK
127.0.0.1:6379(TX)> SET test-key test-value
QUEUED
127.0.0.1:6379(TX)> WATCH test-key
(error) ERR Command not allowed inside a transaction
127.0.0.1:6379(TX)> EXEC
(error) EXECABORT Transaction discarded because of previous errors.
127.0.0.1:6379> GET test-key
(nil)
```
=> the `test-key` was not stored

**Expected behavior**

In order to be used as a drop-in replacement, the valid commands should have been executed, like demonstrated here:

* run a Redis 7 image: `docker run -it --rm --detach 'valkey/valkey:8.0.2-alpine3.21'`
* note the ID of the started container
* connect to the started container: `docker exec -it <id-of-container> /bin/sh`
* connect via CLI: `redis-cli`
* run these commands:
```
127.0.0.1:6379> MULTI
OK
127.0.0.1:6379(TX)> SET test-key test-value
QUEUED
127.0.0.1:6379(TX)> WATCH test-key
(error) ERR WATCH inside MULTI is not allowed
127.0.0.1:6379(TX)> EXEC
1) OK
127.0.0.1:6379> GET test-key
"test-value"
```

**Additional information**

In our application, we don't make use of the CLI, but use a client library instead (for PHP). This client library didn't report the error back to the caller, effectively hiding the error from the developer. So as I said, failing hard in this case would be an acceptable behavior for me - maybe controllable by a flag.

The only issue I could find, that lead me into the right direction is this one (though it's of a different project): https://github.com/taskforcesh/bullmq/issues/2787.

Also worth mentioning is, that the error message in Valkey is not as meaningful as the one in Redis, which usually complicates finding and resolving errors.

enjoy-binbin:
Including command fullname in error messages can help users locate
problems more clearly. This is a potentially breaking change.

Previously we returned `ERR Command not allowed inside a transaction`,
now we return `ERR Command 'shutdown' not allowed inside a transaction`,
for subcommand it is `ERR Command 'client|reply' not allowed inside a transaction`.

Closes #1629.

### DIFF
--- a/tests/unit/multi.tcl
+++ b/tests/unit/multi.tcl
@@ -836,16 +836,11 @@ start_server {tags {"multi"}} {
             r del foo
             r multi
             r set foo bar
-            catch {r $cmd} e1
-            catch {r exec} e2
-            if {$cmd == "SAVE"} {
-                assert_match {*Command 'save' not allowed inside a transaction*} $e1
-            } elseif {$cmd == "SHUTDOWN"} {
-                assert_match {*Command 'shutdown' not allowed inside a transaction*} $e1
-            }
-            assert_match {EXECABORT*} $e2
-            r get foo
-        } {}
+            set cmd_lower [string tolower $cmd]
+            assert_error "ERR Command '$cmd_lower' not allowed inside a transaction*" {r $cmd}
+            assert_error "EXECABORT Transaction discarded because of previous errors*" {r exec}
+            assert_equal [r get foo] {}
+        }
     }
 
     test "MULTI with BGREWRITEAOF" {
--- a/tests/unit/multi.tcl
+++ b/tests/unit/multi.tcl
@@ -35,7 +35,7 @@ start_server {tags {"multi"}} {
 
     test {Nested MULTI are not allowed} {
         r multi
-        assert_error "ERR*" {r multi}
+        assert_error "ERR Command 'multi' not allowed inside a transaction*" {r multi}
         assert_error "EXECABORT*" {r exec}
     }
 
@@ -48,7 +48,7 @@ start_server {tags {"multi"}} {
 
     test {WATCH inside MULTI is not allowed} {
         r multi
-        assert_error "ERR*" {r watch}
+        assert_error "ERR Command 'watch' not allowed inside a transaction*" {r watch x}
         assert_error "EXECABORT*" {r exec}
     }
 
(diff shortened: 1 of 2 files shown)


tests/unit/multi.tcl
```diff
@@ -838,7 +838,11 @@ start_server {tags {"multi"}} {
             r set foo bar
             catch {r $cmd} e1
             catch {r exec} e2
-            assert_match {*Command not allowed inside a transaction*} $e1
+            if {$cmd == "SAVE"} {
+                assert_match {*Command 'save' not allowed inside a transaction*} $e1
+            } elseif {$cmd == "SHUTDOWN"} {
+                assert_match {*Command 'shutdown' not allowed inside a transaction*} $e1
+            }
```
zuiderkwast: Maybe enough like this:

```tcl
            assert_error {ERR Command '*' not allowed inside a transaction} {r $cmd}
```

Alternatively, if we convert the command to lowercase, we can do

```tcl
            assert_error "ERR Command '$cmd_lower' not allowed inside a transaction" {r $cmd}
```

zuiderkwast (approved):
In the issue, it seems that 4/6 of TSC members already voted for or expressed they want this change, so I guess we can already mark it as major-decision-approved. Let's mention it in the next meeting anyway.

tests/unit/multi.tcl
enjoy-binbin: yes, we can, updated.

madolson:
Brought it up again in core meeting, no concerns were discussed.

PR #2257, opened 2025-06-22
Fix use-after-free when active expiration triggers hashtable to shrink

gusakovy:
**Current state**
During `hashtableScanDefrag`, rehashing is paused to prevent entries from moving, but the scan callback can still delete entries which triggers `hashtableShrinkIfNeeded`. For example, the `expireScanCallback` can delete expired entries.

**Issue**
This can cause the table to be resized and the old memory to be freed while the scan is still accessing it, resulting in the following memory access violation:

```
[err]: Sanitizer error: =================================================================
==46774==ERROR: AddressSanitizer: heap-use-after-free on address 0x611000003100 at pc 0x0000004704d3 bp 0x7fffcb062000 sp 0x7fffcb061ff0
READ of size 1 at 0x611000003100 thread T0
    #0 0x4704d2 in isPositionFilled /home/gusakovy/Projects/valkey/src/hashtable.c:422
    #1 0x478b45 in hashtableScanDefrag /home/gusakovy/Projects/valkey/src/hashtable.c:1768
    #2 0x4789c2 in hashtableScan /home/gusakovy/Projects/valkey/src/hashtable.c:1729
    #3 0x47e3ca in kvstoreScan /home/gusakovy/Projects/valkey/src/kvstore.c:402
    #4 0x6d9040 in activeExpireCycle /home/gusakovy/Projects/valkey/src/expire.c:297
    #5 0x4859d2 in databasesCron /home/gusakovy/Projects/valkey/src/server.c:1269
    #6 0x486e92 in serverCron /home/gusakovy/Projects/valkey/src/server.c:1577
    #7 0x4637dd in processTimeEvents /home/gusakovy/Projects/valkey/src/ae.c:370
    #8 0x4643e3 in aeProcessEvents /home/gusakovy/Projects/valkey/src/ae.c:513
    #9 0x4647ea in aeMain /home/gusakovy/Projects/valkey/src/ae.c:543
    #10 0x4a61fc in main /home/gusakovy/Projects/valkey/src/server.c:7291
    #11 0x7f471957c139 in __libc_start_main (/lib64/libc.so.6+0x21139)
    #12 0x452e39 in _start (/local/home/gusakovy/Projects/valkey/src/valkey-server+0x452e39)

0x611000003100 is located 0 bytes inside of 256-byte region [0x611000003100,0x611000003200)
freed by thread T0 here:
    #0 0x7f471a34a1e5 in __interceptor_free (/lib64/libasan.so.4+0xd81e5)
[...]

PR #2288, opened 2025-06-30
Fix leak when shrinking a hashtable without entries

### LINKED ISSUE #2271: Leak when shrinking hash table
Follow-up of #2257

After some more testing, I found that in its current form the test causes another ASAN error. I ran it in a loop with the fix applied and it sometimes causes a different memory leak. Not sure what causes this--maybe some edge case when doing a lot of short expiries:
```
[err]: Sanitizer error: 
==38307==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 64 byte(s) in 1 object(s) allocated from:
    #0 0x7f864c1da753 in __interceptor_calloc (/lib64/libasan.so.4+0xd8753)
    #1 0x4ae4f2 in ztrycalloc_usable_internal /home/gusakovy/Projects/valkey/src/zmalloc.c:214
    #2 0x4ae7bd in valkey_calloc /home/gusakovy/Projects/valkey/src/zmalloc.c:257
    #3 0x4726b9 in bucketConvertToChained /home/gusakovy/Projects/valkey/src/hashtable.c:776
    #4 0x4733cf in findBucketForInsert /home/gusakovy/Projects/valkey/src/hashtable.c:889
    #5 0x470f02 in rehashBucket /home/gusakovy/Projects/valkey/src/hashtable.c:536
    #6 0x47125b in rehashStep /home/gusakovy/Projects/valkey/src/hashtable.c:551
    #7 0x475713 in hashtableRehashMicroseconds /home/gusakovy/Projects/valkey/src/hashtable.c:1225
    #8 0x47f4fd in kvstoreIncrementallyRehash /home/gusakovy/Projects/valkey/src/kvstore.c:658
    #9 0x485c75 in databasesCron /home/gusakovy/Projects/valkey/src/server.c:1308
    #10 0x486ef8 in serverCron /home/gusakovy/Projects/valkey/src/server.c:1577
    #11 0x4637dd in processTimeEvents /home/gusakovy/Projects/valkey/src/ae.c:370
    #12 0x4643e3 in aeProcessEvents /home/gusakovy/Projects/valkey/src/ae.c:513
    #13 0x4647ea in aeMain /home/gusakovy/Projects/valkey/src/ae.c:543
    #14 0x4a6262 in main /home/gusakovy/Projects/valkey/src/server.c:7291
    #15 0x7f864b40c139 in __libc_start_main (/lib64/libc.so.6+0x21139)

SUMMARY: AddressSanitizer: 64 byte(s) leaked in 1 allocation(s).
```
I think its best to just merge the fix and maybe open a separate issue for this.

_Originally posted by @gusakovy in https://github.com/valkey-io/valkey/issues/2257#issuecomment-2996749482_

yzc-yzc:
Fixes #2271

When we shrink a hash table and it is empty, we do it without iterating over it to rehash the entries. However, there may still be empty child buckets (`used[0]==0 && child_buckets[0]!=0`). These were leaked in this case.

This fix is to check for child buckets and don't skip the incremental rehashing if any child buckets exist. The incremental rehashing pass will free them.

An additional fix is to compact bucket chains in scan when the scan callback has deleted some entries. This was already implemented for the case when rehashing is ongoing but it was missing in the case rehashing is not ongoing.

Additionally, a test case for #2257 was added.

### DIFF
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -1761,7 +1761,9 @@ size_t hashtableScanDefrag(hashtable *ht, size_t cursor, hashtableScanFunction f
     if (!hashtableIsRehashing(ht)) {
         /* Emit entries at the cursor index. */
         size_t mask = expToMask(ht->bucket_exp[0]);
-        bucket *b = &ht->tables[0][cursor & mask];
+        size_t idx = cursor & mask;
+        size_t used_before = ht->used[0];
+        bucket *b = &ht->tables[0][idx];
         do {
             if (b->presence != 0) {
                 int pos;
@@ -1778,6 +1780,11 @@ size_t hashtableScanDefrag(hashtable *ht, size_t cursor, hashtableScanFunction f
             }
             b = next;
         } while (b != NULL);
+
+        /* If any entries were deleted, fill the holes. */
+        if (ht->used[0] < used_before) {
+            compactBucketChain(ht, idx, 0);
+        }
 
         /* Advance cursor. */
         cursor = nextCursor(cursor, mask);
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -652,7 +652,7 @@ static int resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
     if (ht->type->rehashingStarted) ht->type->rehashingStarted(ht);
 
     /* If the old table was empty, the rehashing is completed immediately. */
-    if (ht->tables[0] == NULL || ht->used[0] == 0) {
+    if (ht->tables[0] == NULL) {
         rehashingCompleted(ht);
     } else if (ht->type->instant_rehashing) {
         while (hashtableIsRehashing(ht)) {
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -652,7 +652,7 @@ static int resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
     if (ht->type->rehashingStarted) ht->type->rehashingStarted(ht);
 
     /* If the old table was empty, the rehashing is completed immediately. */
-    if (ht->tables[0] == NULL) {
+    if (ht->tables[0] == NULL || (ht->used[0] == 0 && ht->child_buckets[0] == 0)) {
         rehashingCompleted(ht);
     } else if (ht->type->instant_rehashing) {
         while (hashtableIsRehashing(ht)) {
--- a/tests/unit/expire.tcl
+++ b/tests/unit/expire.tcl
@@ -985,7 +985,7 @@ start_server {tags {expire} overrides {hz 100}} {
 
         # Verify server is still responsive
         assert_equal [r ping] {PONG}
-    }
+    } {} {needs:debug}
 }
 
 start_cluster 1 0 {tags {"expire external:skip cluster"}} {
--- a/tests/unit/expire.tcl
+++ b/tests/unit/expire.tcl
@@ -951,6 +951,43 @@ start_server {tags {"expire"}} {
     }
 }
 
+start_server {tags {expire} overrides {hz 100}} {
+    test {Active expiration triggers hashtable shrink} {
+        set persistent_keys 5
+        set volatile_keys 100
+        set total_keys [expr $persistent_keys + $volatile_keys]
+
+        for {set i 0} {$i < $persistent_keys} {incr i} {
+            r set "key_$i" "value_$i"
+        }
+        for {set i 0} {$i < $volatile_keys} {incr i} {
+            r psetex "expire_key_${i}" 100 "expire_value_${i}"
+        }
+        set table_size_before_expire [main_hash_table_size]
+
+        # Verify keys are set
+        assert_equal $total_keys [r dbsize]
+
+        # Wait for active expiration
+        wait_for_condition 100 50 {
+            [r dbsize] eq $persistent_keys
+        } else {
+            fail "Keys not expired"
+        }
+
+        # Wait for the table to shrink and active rehashing finish
+        wait_for_condition 100 50 {
+            [main_hash_table_size] < $table_size_before_expire
+        } else {
+            puts [r debug htstats 9]
+            fail "Table didn't shrink"
+        }
+
+        # Verify server is still responsive
+        assert_equal [r ping] {PONG}
+    }
+}
+
 start_cluster 1 0 {tags {"expire external:skip cluster"}} {
     test "expire scan should skip dictionaries with lot's of empty buckets" {
         r debug set-active-expire 0


enjoy-binbin (approved):
I see we do the same condition in the dict.c, so it should be fine. I did not dive into the details

src/hashtable.c
```diff
@@ -652,7 +652,7 @@ static int resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
     if (ht->type->rehashingStarted) ht->type->rehashingStarted(ht);
 
     /* If the old table was empty, the rehashing is completed immediately. */
-    if (ht->tables[0] == NULL || ht->used[0] == 0) {
+    if (ht->tables[0] == NULL) {
```
zuiderkwast: If we want to keep the fast path to avoid iterating over the large table, we can add a check for `child_buckets[0] == 0` here. Is it worth it or is it a very unlikely case?

```suggestion
    if (ht->tables[0] == NULL || (ht->used[0] == 0 && ht->child_buckets[0] == 0)) {
```

zuiderkwast (commented):
Great finding!

I have a minor suggestion. If I'm wrong, we can merge it as is.

zuiderkwast:
> Should we add the test case from #2257 ?

I tested it locally and noticed it is slow. It takes 15 seconds. I have modified it to remove the sleeps and some loops. Now it takes 160ms. You can add it if you want:

```tcl
start_server {tags {expire} overrides {hz 100}} {
    test {Active expiration triggers hashtable shrink} {
        set persistent_keys 5
        set volatile_keys 100
        set total_keys [expr $persistent_keys + $volatile_keys]

        for {set i 0} {$i < $persistent_keys} {incr i} {
            r set "key_$i" "value_$i"
        }
        for {set i 0} {$i < $volatile_keys} {incr i} {
            r psetex "expire_key_${i}" 100 "expire_value_${i}"
        }
        set table_size_before_expire [main_hash_table_size]

        # Verify keys are set
        assert_equal $total_keys [r dbsize]

        # Wait for active expiration
        wait_for_condition 100 50 {
            [r dbsize] eq $persistent_keys
        } else {
            fail "Keys not expired"
        }

        # Wait for the table to shrink and active rehashing finish
        wait_for_condition 100 50 {
            [main_hash_table_size] < $table_size_before_expire
        } else {
            puts [r debug htstats 9]
            fail "Table didn't shrink"
        }

        # Verify server is still responsive
        assert_equal [r ping] {PONG}
    }
}
```

src/hashtable.c
yzc-yzc: I think it depends. To trigger this scenario, you need to pause the shrink and remove entries from a hash table with at least 1 child bucket until there are 0 entries remaining. (e.g. expire dict delete remaining entries and pause the shrink in the meantime)  So if we reduce the shrink frequency accidentally, the probability of triggering may be higher. For now, I think this optimization can be added.

yzc-yzc:
1. According to the code comments shown below, pause_auto_shrink should not be used for safety issues (maybe just for performance), right? 

```
/* Pauses automatic shrinking. This can be called before deleting a lot of
 * entries, to prevent automatic shrinking from being triggered multiple times.
 * Call hashtableResumeAutoShrink afterwards to restore automatic shrinking. */
void hashtablePauseAutoShrink(hashtable *ht) {
    ht->pause_auto_shrink++;
}
```

2. During the scan function, we pause rehash, but why can we resize the hash table? Is it expected that pause_rehash doesn't work for resize?

@zuiderkwast The above two questions are bothering me, could you help me out? Thanks!

zuiderkwast:
> 1. According to the code comments shown below, pause_auto_shrink should not be used for safety issues (maybe just for performance), right?

@yzc-yzc That's right. It is used internally in hashtable.c for safety reasons but in the public API it should not be required for safety.

> 2. During the scan function, we pause rehash, but why can we resize the hash table? Is it expected that pause_rehash doesn't work for resize?

For safety of scan, I think it's OK to resize but not rehash during the scan. For example if the scan callback deletes entries. Resize only allocates a new table but no entries are moved, so it doesn't affect the scan algorithm.

However, if new entries are inserted while rehashing is paused (for example the scan callback inserts new entries, or in another situation where the rehashing is paused) they are inserted in the new table. If a lot of new entries are inserted, more than what can fit in the old table, I think it's good that resize is enabled even if rehashing is paused.

Makes sense?

gusakovy:
I also went through a similar debugging process that @yzc-yzc described, and I want to suggest an alternative way to handle this edge case which might be more efficient since you won't have to do any extra rehash steps.

The chained buckets are not released since `hashtablePop` does not compact the bucket chains when rehashing is paused:
```
int hashtablePop(hashtable *ht, const void *key, void **popped) {
  ...
        if (b->chained && !hashtableIsRehashingPaused(ht)) {
            /* Rehashing is paused while iterating and when a scan callback is
             * running. In those cases, we do the compaction in the scan and
             * iterator code instead. */
            fillBucketHole(ht, b, pos_in_bucket, table_index);
        }
   ...
}
```
According to the comment, in that case the compaction should be handles by the scan or iterator. For some reason we only call `compactBucketChain` in `hashtableScanDefrag` when the hashtable is rehashing, and I think we should add when we're not rehashing as well:
```
if (!hashtableIsRehashing(ht)) {
        /* Emit entries at the cursor index. */
-        bucket *b = &ht->tables[0][cursor & mask];
+        size_t idx = cursor & mask;
+        size_t used_before = ht->used[0];
+        bucket *b = &ht->tables[0][idx];
        do {
            if (b->presence != 0) {
                int pos;
                for (pos = 0; pos < ENTRIES_PER_BUCKET; pos++) {
                    if (isPositionFilled(b, pos)) {
                        void *emit = emit_ref ? &b->entries[pos] : b->entries[pos];
                        fn(privdata, emit);
                    }
                }
            }
            bucket *next = getChildBucket(b);
            if (next != NULL && defragfn != NULL) {
                next = bucketDefrag(b, next, defragfn);
            }
            b = next;
        } while (b != NULL);
+        /* If any entries were deleted, fill the holes. */
+        if (ht->used[0] < used_before) {
+            compactBucketChain(ht, idx, 0);
+        }

        /* Advance cursor. */
        cursor = nextCursor(cursor, mask);
    } else {
...
```

zuiderkwast:
> According to the comment, in that case the compaction should be handles by the scan or iterator. For some reason we only call `compactBucketChain` in `hashtableScanDefrag` when the hashtable is rehashing, and I think we should add when we're not rehashing as well

@gusakovy Why is this useful? If rehashing is not paused, the compaction happens immediately when an entry is deleted. Am I missing something?

There is another scenario that can lead to holes and empty buckets: If the scan callback deletes some other entry which is not in the same bucket that was just scanned. Also if rehashing was paused and entries were deleted (without scan or iterator) we can get empty buckets. Therefore, I think @yzc-yzc's fix is still needed.

gusakovy:
> Why is this useful? If rehashing is not paused, the compaction happens immediately when an entry is deleted. Am I missing something?

During scan we pause rehashing so `hashtableIsRehashingPaused(ht)` is always true and `hashtablePop` will never compact.

In `hashtableScanDefrag` we condition on `hashtableIsRehashing(ht)`, i.e whether the hashtable is **in the middle of rehashing or not** and for some reason currently call `compactBucketChain` only in the case when the hashtable is in the middle of rehashing: 

>There is another scenario that can lead to holes and empty buckets: If the scan callback deletes some other entry which is not in the same bucket that was just scanned.

If that is possible then yes definitely the proposed fix is still needed.

zuiderkwast:
@gusakovy Gotcha. Rehashing paused when rehashing is not ongoing still means that compaction doesn't happen automatically.

Your fix is good. It's not required for fixing the leak but it's needed to clean up empty buckets during scan. For example if many entries are expired, it can cause a lot of empty buckets.

@yzc-yzc Can you include @gusakovy's patch in https://github.com/valkey-io/valkey/pull/2288#issuecomment-3023907014 above?

yzc-yzc:
> Makes sense?

got it, thanks!




> @yzc-yzc Can you include @gusakovy's patch in [#2288 (comment)](https://github.com/valkey-io/valkey/pull/2288#issuecomment-3023907014) above?

sure

yzc-yzc:
> > Should we add the test case from #2257 ?
> 
> I tested it locally and noticed it is slow. It takes 15 seconds. I have modified it to remove the sleeps and some loops. Now it takes 160ms. You can add it if you want:
> 
> ```tcl
> start_server {tags {expire} overrides {hz 100}} {
>     test {Active expiration triggers hashtable shrink} {
>         set persistent_keys 5
>         set volatile_keys 100
>         set total_keys [expr $persistent_keys + $volatile_keys]
> 
>         for {set i 0} {$i < $persistent_keys} {incr i} {
>             r set "key_$i" "value_$i"
>         }
>         for {set i 0} {$i < $volatile_keys} {incr i} {
>             r psetex "expire_key_${i}" 100 "expire_value_${i}"
>         }
>         set table_size_before_expire [main_hash_table_size]
> 
>         # Verify keys are set
>         assert_equal $total_keys [r dbsize]
> 
>         # Wait for active expiration
>         wait_for_condition 100 50 {
>             [r dbsize] eq $persistent_keys
>         } else {
>             fail "Keys not expired"
>         }
> 
>         # Wait for the table to shrink and active rehashing finish
>         wait_for_condition 100 50 {
>             [main_hash_table_size] < $table_size_before_expire
>         } else {
>             puts [r debug htstats 9]
>             fail "Table didn't shrink"
>         }
> 
>         # Verify server is still responsive
>         assert_equal [r ping] {PONG}
>     }
> }
> ```

I ran this test on my PC and found that the probability of triggering leak is very low.(Triggered once after 257 times)
My build command is `make noopt SANITIZER=address valkey-server`. The code is the last commit before this pr. Am I missing something?

zuiderkwast:
> I ran this test on my PC and found that the probability of triggering leak is very low.(Triggered once after 257 times)
My build command is make noopt SANITIZER=address valkey-server. The code is the last commit before this pr. Am I missing something?

This is not for testing the leak. The main point is for testing that Active expiration triggers hashtable shrink. It's what was actually implemented in #2257.

The original test case was too slow IMO. I think we don't really need a special test case to trigger the leak. If you think we need one, we should try to write a unit test in `src/unit/test_hashtable.c` which can run faster.

zuiderkwast (approved):

PR #2290, opened 2025-07-01
Fix MEMORY USAGE to consider embedded keys

yulazariy:
Fix objectComputeSize function to use `zmalloc_size` instead of `sizeof` when calculating the size of the top level object.

That will take in account the embedded key within the object.


zuiderkwast (approved):
@yulazariy Thank you!

@ranshid FYI. You noticed this too, I guess. (You touched some of this in another PR.)

ranshid:
> @yulazariy Thank you!
> 
> @ranshid FYI. You noticed this too, I guess. (You touched some of this in another PR.)

@yulazariy found it. I only helped with the fix (indeed some of the example fix slipped in my PR :) )

PR #2292, opened 2025-07-01
Auto-failover on shutdown unified config

zuiderkwast:
In #1091, a new config `auto-failover-on-shutdown` was added. This PR changes the config to make it unified with other shutdown related options. This feature has not yet been released, so it's not a breaking change.

The auto-failover-on-shutdown config is replaced by

* A new "failover" option to the existing configs `shutdown-on-sigterm` and `shutdown-on-sigint`.
* A new FAILOVER option to the SHUTDOWN command.

Additionally, a history entry is added to the SHUTDOWN command which was missing in #2195.

Follow-up of #1091.


zuiderkwast:
@valkey-io/core-team Please ack ( :+1: ) the slightly modified configs and options mentioned in the top comment. (This idea was mentioned in https://github.com/valkey-io/valkey/pull/1091#issuecomment-2987983236.)

enjoy-binbin (approved):
LGTM! I originally planned to handle it after merging the SAFE.

src/db.c
```diff
@@ -1353,6 +1355,11 @@ void shutdownCommand(client *c) {
         return;
     }
 
+    if (flags & SHUTDOWN_FAILOVER && !server.cluster_enabled) {
```
enjoy-binbin, line 1358: i see we check this, so we may also need to check cluster in config.c

zuiderkwast: We didn't check it before... I just wanted to change the config, not implement more stuff. :) But I guess you are right. It's better to check it.

zuiderkwast: If we forbid it in config if cluster mode is disabled, it means Valkey refuses to start if this config is set. Maybe it's not good if some users want to just change cluster-enabled to enable/disable cluster mode. They would need to change multiple configs at the same time. This kind of dependency between configs is annoying, WDYT?

enjoy-binbin: sure, that is not a big deal, i am ok with this.

madolson (approved):

madolson:
@zuiderkwast We discussed in the weekly meeting to open a follow up for standalone.

My approval is for the direction of the feature, didn't look at the details.

src/db.c
PingXie: Do we have a plan (I thought we do?) to extend this feature to standalone? if yes, the current implementation makes sense.

PingXie: on a side note, it is a bit odd to see `shutdownCommand` implemented in `db.c`.

src/cluster_legacy.c
```diff
@@ -1348,9 +1348,9 @@ void clusterAutoFailoverOnShutdown(void) {
 }
 
 /* Called when a cluster node receives SHUTDOWN. */
-void clusterHandleServerShutdown(void) {
+void clusterHandleServerShutdown(bool auto_failover) {
     /* Check if we are able to do the auto failover on shutdown. */
-    clusterAutoFailoverOnShutdown();
+    if (auto_failover) clusterAutoFailoverOnShutdown();
```
PingXie, line 1356: nitpicks on names

1. `auto` isn't necessary imo - it is implied that the code is doing the work
2. the original function `clusterHandleServerShutdown` doesn't know where/why it is invoked. it just merely provides the failover functionality. 

```suggestion
    if (failover) clusterInitiateGracefulFailover();
```

PingXie (commented):
LGTM overall

src/db.c
zuiderkwast: Yes, I've opened an issue for standalone:

* #2349

db.c contains a lot of various stuff. We should maybe organize it better but not in this PR i hope.

src/cluster_legacy.c
zuiderkwast: True. This was added in #1091. I removed "auto" from the configs (public API) but I didn't change the internals... I'll see if I can get to it.

madolson:
@enjoy-binbin @zuiderkwast  Is this good to merge?

enjoy-binbin:
yes, it is good to merge, viktor will merge it when he come back.

src/cluster_legacy.c
zuiderkwast: I guess we can ignore this nitpick and merge, to get it off the table.

PR #2295, opened 2025-07-02
Build with fast_float by default

zuiderkwast:
When fast_float was added in 8.1 (#1260), we didn't want to build with it by default. It's a change to the build toolchain for the default build configuration and we didn't want to do this change in a minor version. We can do it now for the upcoming 9.0 though.

CI and Daily jobs are updated so that some jobs run without fast_float, which now needs to be specified explicitly. With these changes, most of the jobs run with fast_float and only a few jobs run without it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       - name: make
         # Fail build if there are warnings
         # build with TLS just for compilation coverage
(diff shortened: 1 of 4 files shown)


zuiderkwast:
We need to change various build dependencies on various platforms. It will probably cause trouble to users too. Maybe it's not worth it? Closing for now.

PR #2296, opened 2025-07-02
Make valkey-server --help/-h output to stdout and exit with 0

enjoy-binbin:
This is similar to 70dac4b4367485e0eddc1ae84b708cd14c18d626, happened
to and clean it up by the way.

"All" other CLI tools display the help information on the stdout.
Which is useful if you want to manipulate directly the data (grepping
for example, without needing to do redirections). The exit status code
is usually 0 too, which makes sense.

It's printed only if --help/-h is used, not on syntax error, any other
value will be treated as a configuration item value.

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -6472,24 +6472,24 @@ sds getVersion(void) {
 }
 
 void usage(void) {
-    fprintf(stderr, "Usage: ./valkey-server [/path/to/valkey.conf] [options] [-]\n");
-    fprintf(stderr, "       ./valkey-server - (read config from stdin)\n");
-    fprintf(stderr, "       ./valkey-server -v or --version\n");
-    fprintf(stderr, "       ./valkey-server -h or --help\n");
-    fprintf(stderr, "       ./valkey-server --test-memory <megabytes>\n");
-    fprintf(stderr, "       ./valkey-server --check-system\n");
-    fprintf(stderr, "\n");
-    fprintf(stderr, "Examples:\n");
-    fprintf(stderr, "       ./valkey-server (run the server with default conf)\n");
-    fprintf(stderr, "       echo 'maxmemory 128mb' | ./valkey-server -\n");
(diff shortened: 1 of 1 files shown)


src/server.c
```diff
@@ -6472,24 +6472,24 @@ sds getVersion(void) {
...
-    fprintf(stderr, "       echo 'maxmemory 128mb' | ./valkey-server -\n");
-    fprintf(stderr, "       ./valkey-server /etc/valkey/6379.conf\n");
-    fprintf(stderr, "       ./valkey-server --port 7777\n");
-    fprintf(stderr, "       ./valkey-server --port 7777 --replicaof 127.0.0.1 8888\n");
-    fprintf(stderr, "       ./valkey-server /etc/myvalkey.conf --loglevel verbose -\n");
-    fprintf(stderr, "       ./valkey-server /etc/myvalkey.conf --loglevel verbose\n\n");
-    fprintf(stderr, "Sentinel mode:\n");
-    fprintf(stderr, "       ./valkey-server /etc/sentinel.conf --sentinel\n");
-    exit(1);
+    fprintf(stdout, "Usage: ./valkey-server [/path/to/valkey.conf] [options] [-]\n");
```
zuiderkwast, line 6475: Simplification? If you want.

```suggestion
    printf("Usage: ./valkey-server [/path/to/valkey.conf] [options] [-]\n");
```

zuiderkwast (approved):
It's printed only if `--help/-h` is used, not on syntax error, right? If we print it on syntax error, we should exit with non-zero.

enjoy-binbin:
> It's printed only if --help/-h is used, not on syntax error, right? If we print it on syntax error, we should exit with non-zero.

yes, we don't have syntax error here. Any other value will be treated as a configuration item value.

src/server.c
enjoy-binbin: i am going to keep it since they are both quite easy. (or just because i am lazy..)

PR #2089, opened 2025-05-15
Introduce HASH items expiration

ranshid:
Closes https://github.com/valkey-io/valkey/issues/640

### Summary

This PR introduces support for **field-level expiration in Valkey hash types**, making it possible for individual fields inside a hash to expire independently — creating what we call **volatile fields**.
This is just the first out of 3 PRs. The content of this PR focus on enabling the basic ability to set and modify hash fields expiration as well as persistency (AOF+RDB) and defrag. 
[The second PR](https://github.com/ranshid/valkey/pull/5) introduces the new algorithm (volatile-set) to track volatile hash fields is in the last stages of review. The current implementation in this PR (in volatile-set.h/c) is just s tub implementation and will be replaced by [The second PR](https://github.com/ranshid/valkey/pull/5) 
[The third PR](https://github.com/ranshid/valkey/pull/4/) which introduces the active expiration and defragmentation jobs.

 
For more highlevel design details you can track the RFC PR: https://github.com/valkey-io/valkey-rfc/pull/22.

---

### Major decisions
Some highlevel major decisions which are taken as part of this work:
1. We decided to copy the existing Redis API in order to maintain compatibility with existing clients.

2. We decided to avoid introducing lazy-expiration at this point, in order to reduce complexity and rely only on active-expiration for memory reclamation. This will require us to continue to work on improving the active expiration job and potentially consider introduce lazy-expiration support later on. 

3. Although different commands which are adding expiration on hash fields are influencing the memory utilization (by allocating more memory for expiration time and metadata) we decided to avoid adding the DENYOOM for these commands (an exception is HSETEX) in order to be better aligned with highlevel keys commands like `expire`

4. Some hash type commands will produce unexpected results:
 - HLEN - will still reflect the number of fields which exists in the hash object (either actually expired or not). 
 - HRANDFIELD - in some cases we will not be able to randomly select a field which was not already expired. this case happen in 2 cases: 1/ when we are asked to provide a non-uniq fields (i.e negative count) 2/ when the size of the hash is much bigger than the count and we need to provide uniq results. In both cases it is possible that an empty response will be returned to the caller, even in case there are fields in the hash which are either persistent or not expired. 
 
5. For the case were a field is provided with a zero (0) expiration time or expiration time in the past, it is immediately deleted. We decided that, in order to be aligned with how high level keys are handled, we will emit hexpired keyspace event for that case (instead of hdel). For example:
for the case:
```
HSET myhash f1 v1
> 0
[...]

PR #2299, opened 2025-07-02
Adding support for sharing memory between the module and the engine

yairgott:
## Overview

Sharing memory between the module and engine reduces memory overhead by eliminating redundant copies of stored records in the module. This is particularly beneficial for search workloads that require indexing large volumes of documents.

### Vectors

Vector similarity search requires storing large volumes of high-cardinality vectors. For example, a single vector with 512 dimensions consumes 2048 bytes, and typical workloads often involve millions of vectors. Due to the lack of a memory-sharing mechanism between the module and the engine, valkey-search currently doubles memory consumption when indexing vectors, significantly increasing operational costs. This limitation introduces adoption friction and reduces valkey-search's competitiveness.

## Memory Allocation Strategy

At a fundamental level, there are two primary allocation strategies:
- [Chosen] Module-allocated memory shared with the engine.
- Engine-allocated memory shared with the module.

For valkey-search, it is crucial that vectors reside in cache-aligned memory to maximize SIMD optimizations. Allowing the module to allocate memory provides greater flexibility for different use cases, though it introduces slightly higher implementation complexity.

## Old Implementation

The old [implementation](https://github.com/valkey-io/valkey/pull/1804) was based on ref-counting and introduced a new SDS type. After further discussion, we [agreed](https://github.com/valkey-io/valkey/pull/1804#issuecomment-2905115712) to simplify the design by removing ref-counting and avoiding the introduction of a new SDS type.

## New Implementation - Key Points

1. The engine exposes a new interface, `VM_HashExternalize`, which externalizes a buffer as a hash field. The externalized buffer is owned by the module. The function accepts the hash key, hash field, and a buffer along with its length.
2. `ExternalizedValue` is a new data type that captures the externalized buffer and its length.
3. `FIELD_SDS_AUX_BIT_EXTERNALIZED` is a new `SDS` `AUX` flag used to distinguish between regular/embedded SDS values and ExternalizedValue.

### Insertion
1. Upon receiving a key space notification for a new hash or JSON key with an indexed vector attribute, valkey-search allocates cache-aligned memory and deep-copies the vector value.
2. valkey-search then calls VM_HashExternalize to avoid performing two copies of the vector.
### Deletion
When receiving a key space notification for a deleted hash key or hash field that was indexed as a vector, valkey-search deletes the corresponding entry from the index.

### Update
Handled similarly to insertion.


yairgott:
cc @zuiderkwast

madolson:
Viktor is currently out, @ranshid can you help with merging this?

yairgott:
> Viktor is currently out, @ranshid can you help with merging this?

@ranshid let me know if you want to have a bootstrap discussion.

ranshid:
> > Viktor is currently out, @ranshid can you help with merging this?
> 
> @ranshid let me know if you want to have a bootstrap discussion.

@yairgott I would be super happy and interested reviewing this. I will be pretty occupied this week (till the 20th) but I will try to look still. 

without taking a look, I suggest you also take a look at https://github.com/valkey-io/valkey/pull/2089 since there will probably need to be some gentle integration points

zvi-code:
I have some question regarding the simplified approach vs re-counting. The main points are: 
1. It creates strict dependency between the search-module and the engine. This is IMO anti-pattern for latency sensitive system, WDUT? 
2. It does not handle the sharing with more than 2 components. I can see easily that we have more than 1 vector index, even for rebuilding logic, pointing to the same vector.
3. We might also want to share the keys themselves, in case they are large, text and other looking forward. Not sure if this mechanism will be extensible to these use cases

yairgott:
> It creates strict dependency between the search-module and the engine. This is IMO anti-pattern for latency sensitive system, WDUT?

I'm not sure about this statement. Why this has to do with latency sensitive system? Fetching an externed hash field value is about as effective as fetching sds based value.

> I can see easily that we have more than 1 vector index, even for rebuilding logic, pointing to the same vector.

Ref counting is handled internally by valkey-search, we have a mechanism which is called [StringInterning](https://github.com/valkey-io/valkey-search/blob/main/src/utils/string_interning.h) to provide exactly this functionality.

> We might also want to share the keys themselves, in case they are large, text and other looking forward. Not sure if this mechanism will be extensible to these use cases

This is already implemented by using the StringInterning mechanism.

zvi-code:
> > It creates strict dependency between the search-module and the engine. This is IMO anti-pattern for latency sensitive system, WDUT?
> 
> I'm not sure about this statement. Why this has to do with latency sensitive system? Fetching an externed hash field value is about as effective as fetching sds based value.
> 
The main point was about Delete flows(I should have clarified). I emphasis `latency sensitive` because you want to have the flexibility to handle deletion async (without the need to coordinate between the indexes\kv\db). If you drop index, do you need to replace the vector in HASH?

yairgott:
> The main point was about Delete flows(I should have clarified). I emphasis latency sensitive because you want to have the flexibility to handle deletion async (without the need to coordinate between the indexes\kv\db). If you drop index, do you need to replace the vector in HASH?

This engine change doesn't dictate replacing the vector in the Hash. For simplicity, the way that I've implemented this is by not replacing the vector in the HASH, the StoreInterning object is not deleted. Alternatively, we can also issue replace but may consider with rate limiting.

Note that the vector externing in valkey-search is already [implemented](https://github.com/valkey-io/valkey-search/blob/main/src/vector_externalizer.h). It requires some modifications to be adjusted to this API. The main difference is that the existing implementation has the notion of normalized vectors. The new implementation keeps the vectors just in its raw form but requires some changes to the HNSWLib to support cosine distance calculation with a pre-computed magnitude. 

If you're interested to learn more, I think it would be easier to do a meeting.

zvi-code:
> > The main point was about Delete flows(I should have clarified). I emphasis latency sensitive because you want to have the flexibility to handle deletion async (without the need to coordinate between the indexes\kv\db). If you drop index, do you need to replace the vector in HASH?
> 
> This engine change doesn't dictate replacing the vector in the Hash. For simplicity, the way that I've implemented this is by not replacing the vector in the HASH, the StoreInterning object is not deleted. Alternatively, we can also issue replace but may consider with rate limiting.
> 
> Note that the vector externing in valkey-search is already [implemented](https://github.com/valkey-io/valkey-search/blob/main/src/vector_externalizer.h). It requires some modifications to be adjusted to this API. The main difference is that the existing implementation has the notion of normalized vectors. The new implementation keeps the vectors just in its raw form but requires some changes to the HNSWLib to support cosine distance calculation with a pre-computed magnitude.
> 
> If you're interested to learn more, I think it would be easier to do a meeting.

Thanks, it makes more sense now. I was not aware with the sharing at the search engine side. I agree that with this mechanism you can delay deletion.

yairgott:
> anshid

Hey Ran,

Have you managed to review this?

[ranshid](https://github.com/ranshid)

yairgott:
This PR was based on Ran's hash TTL branch. Closing this PR in favor of https://github.com/valkey-io/valkey/pull/2472.

Issue #2031, opened 2025-04-30
[NEW] Accelerate CRC16 with SIMD


xbasel:
Accelerate CRC16 with NEON and AVX

zuiderkwast:
I have found a Rust implementation that we can try to translate to C.

* https://github.com/TobiasBengtsson/crc-fast-rs/tree/master/crc16-xmodem-fast

It uses [carry-less multiplication instructions](https://en.m.wikipedia.org/wiki/CLMUL_instruction_set) and on Arm it uses aes instructions somehow.

It's lengthy so I doubt it's fast for very short input. Keys are usually short, ~20 bytes.

zuiderkwast:
Apparently, the Rust implementation falls back to the simple old table lookup implementation for strings shorter than 128 bytes. https://github.com/TobiasBengtsson/crc-fast-rs/blob/master/crc-fast-gen/src/lib.rs#L222

I doubt that it's possible to optimize with SIMD for short strings.

@xbasel Do you want to optimize for long keynames? I don't think we should, because it's not a common use case.

zuiderkwast:
These? https://gcc.gnu.org/onlinedocs/gcc/CRC-Builtins.html

Cuda-Chen:
Hi @xbasel  and @zuiderkwast ,

I have done the acceleration of NEON for CRC32-C on ARM (particularly Armv7-A and Armv8-A 32-bit) with both lookup table and carry-less multiplication (reference: https://github.com/DLTcollab/sse2neon/pull/627/files).
If this issue is still active, I would like to handle this (~I will implement ARM part first~ after reading the code, I think I should propose a full schedule).
For adding the assignee for me will be great, if possible.

zuiderkwast:
@Cuda-Chen Key names are typically only 20 bytes long. Is there any performance benefit for this length?

IIUC, these techniques would only optimeze payloads of length 128 or longer. Do you have another experience?

We should optimize this primarily for short keys, let's say up to 32 bytes. If we add branching that makes it slower for short keys, I don't think it's worth it, but i'd like to see some numbers and also see the code complexity to decide.

Cuda-Chen:
Hi @zuiderkwast ,

> IIUC, these techniques would only optimeze payloads of length 128 or longer. Do you have another experience?

[...]

PR #2300, opened 2025-07-03
Optimize CRC16 lookup table size

Cuda-Chen:
Apply half-byte lookup table for fitting the whole table into CPU's L1 cacheline (usually with size of 64 Bytes).

For an ordinary CRC16 lookup table, it needs 512 Bytes (2 Bytes for each polynomial *
256 entries for the bit permutation of one Byte).
Whereas this implementation merely needs 32 Bytes (2 Bytes for each polynomial *
16 entries for the bit permutation of one Byte).

Mention #2031.


src/crc16.c
```diff
@@ -29,60 +29,77 @@
...
- * Reflect Output CRC         : False
- * Xor constant to output CRC : 0000
- * Output for "123456789"     : 31C3
+ * Poly                       : 0x8408 (That is actually 0x1021, or x^16 + x^12 + x^5 + 1)
+ * Initialization             : 0x0000
+ * Reflect Input byte         : True
+ * Reflect Output CRC         : True
+ * Xor constant to output CRC : 0x0000
+ * Output for "123456789"     : 0x31C3
  */
```
zuiderkwast: Why change the description? It the old description not correct or is it equivalent to your new description, but yours is written in a reflected way...?

I think we can keep the old description, if it's not incorrect. It seems to be the most common way to describe this CRC-16 variant. For example, according to this page, the Xmodem variant has the parameters mentioned by Antirez:

* https://reveng.sourceforge.io/crc-catalogue/all.htm#crc.cat.crc-16-xmodem

If a reflected and non-reflected variant are equivalent with just different numbers in the table, is it possible to avoid the `reverse_8_bit()` and `reverse_16_bit()` calls, if we just use different numbers in the table?

zuiderkwast (commented):
Thank you! This is very interesting. Some first thoughts:

1. Have you done any benchmarks?

   My guess is that the large table is faster if all of the table is in L1 cache, so if the algorithms runs isolated in a benchmark loop, the old one will be faster. However, if the table is not in L1 cache, this new one will be faster.

   An idea is to just test using `valkey-benchmark` against a single-cluster. Another idea is to use perf and/or a flame graph and look at the part spent on crc16. If you are not familiar with benchmarking Valkey, there are others who can help with this (e.g. @xbasel, @SoftlyRaining, @roshkhatri).

2. Since the whole file is completely rewritten, we can also drop the copyright and licence header and replace it with our shorter variant

   ```
   /*
    * All rights reserved.
    */
   ```

3. Let's move the old implementation with the large table to a unit test and use it to compare the old and new implementation. Let's create a new unit test in `src/unit/test_crc16.c`. Follow the other files in that directory to see how it's done. It's a very minimal unit test framework.

src/crc16.c
Cuda-Chen: > but yours is written in a reflected way

For my opinion, writing in a reflected way has these advantages:

1. No need to 1) get input byte then shift, 2) "first get value then shift" when retrieving the entry from LUT.
    For example:
    
    ```c
    // a reflected way
    for(i = 0; i < len; i++) {
        uint_8 v = (uint8_t)buf[i];
        v = reverse_8_bit(v); // REQUIRED: reverse all bits
        crc ^= v;
        crc = (crc >> 4) ^ tbl_reduced[crc & 0x000f];
        crc = (crc >> 4) ^ tbl_reduced[crc & 0x000f];
    }
    return reverse_16_bit(crc); // REQUIRED: reverse all bits

    // a normal way
    for(i = 0; i < len; i++) {
        uint_8 v = (uint8_t)buf[i] << 8;
        crc ^= v;
        crc = (crc << 4) ^ tbl_forward_reduced[crc >> 12];
        crc = (crc << 4) ^ tbl_forward_reduced[crc >> 12];
    }

    return crc;
    ```

2. Looks more intuitive.
    Many implementations uses reflected way (including carry-less multiplication ones). Choosing this results in a great comparison for code readability. 

However, my current implementation needs additional two XOR operations in order to get the same result compared to a normal calculation.

> If a reflected and non-reflected variant are equivalent with just different numbers in the table, is it possible to avoid the reverse_8_bit() and reverse_16_bit() calls, if we just use different numbers in the table?

For my experiment, there exists a non-reflected variant of half-byte table solution as below:

```c
/* Adapted from: https://www.eevblog.com/forum/programming/crc32-half-bytenibble-lookup-table-algorithm/ */

// A forward LUT, just retrieve the first 16 entries from ordinary LUT.
static const uint16_t tbl_forward_reduced[16] = { 
    0x0000,0x1021,0x2042,0x3063,0x4084,0x50a5,0x60c6,0x70e7,
    0x8108,0x9129,0xa14a,0xb16b,0xc18c,0xd1ad,0xe1ce,0xf1ef,
};

// CRC-16 main subroutine
uint16_t crc16(const char *buf, int len) {
    int counter;
    uint16_t crc = 0;
    for(counter = 0; counter < len; counter++) {
        crc ^= (uint16_t)buf[counter] << 8;
        crc = (crc << 4) ^ tbl_forward_reduced[crc >> 12];
        crc = (crc << 4) ^ tbl_forward_reduced[crc >> 12];
    }  
    return crc;
}
```

For my current understanding, a normal calculation is equivalent to a bit-reversed calculation with reflecting bits of input bytes and output CRC, and vice versa (e.g., CRC32C denotes the normal calculation with reflecting bits of input bytes and output CRC, whereas many implementations choose bit-reversed calculation because there are no needs to reflect the bits of input bytes and output CRC).
You may see this post for further information: https://github.com/Michaelangel007/crc32?tab=readme-ov-file#tldr-crc32-summary

To sum up, I think we may try the normal calculation then benchmarking to choose which implementations may be the best.

Cuda-Chen:
Hi @zuiderkwast ,

> 1. Have you done any benchmarks?
> ...
> If you are not familiar with benchmarking Valkey, there are others who can help with this (e.g. @xbasel, @SoftlyRaining, @roshkhatri).

I haven't done benchmarks, yet.
Still, thanks for pointing some directions of benchmarking, including an initial entrypoint of how ValKey does benckmarking. I am going to adapt myself to the test framework within these days.

By the way, I would like to ask how should I inform these people you have mentioned for help (e.g., GitHub Issues)?

> 2. Since the whole file is completely rewritten, we can also drop the copyright and licence header and replace it with our shorter variant

I will replace the copyright and licence header in future commits.

> 3. Let's move the old implementation with the large table to a unit test and use it to compare the old and new implementation. Let's create a new unit test in `src/unit/test_crc16.c`. Follow the other files in that directory to see how it's done. It's a very minimal unit test framework.

Thanks for saving my day! I'm just thinking how ValKey executing unit tests.

Cuda-Chen:
Hi @zuiderkwast ,

I am going to update the process of benchmarking.
The code and the results will be written in later comments. 
For testing, here some factors:
- I create a code snippet for testing the implementations (original, forward, and reverse calculation). 
- The test program is compiled as follows: `gcc test.c -O2 -Wall -g -fno-omit-frame-pointer`
- Use `perf record -g -e cycles,instructions,L1-dcache-load-misses ./a.out` to save the results.
- Use `perf report -g graph,0.5,caller` for viewing the detailed results.
- Use `perf stat -g -e cycles,instructions,L1-dcache-load-misses ./a.out` to view the glance of results.

In short, I observe the following phenomena:

1. Both half-byte LUT implementations get longer execution time.
2. The forward calculation has fewer L1 cache miss counts compared to original one.
3. The reverse calculation has more L1 cache miss counts compared to original one.
    - The reason is that the reverse bit functions provide some L1 cache miss counts.

I also provide the L1 cache miss counts table as follows:

| original | forward | reverse |
| --- | --- | --- |
| 10727 | 9809 | 12288 |

Cuda-Chen:
## Test Environments

### CPU

```
$ lscpu
Architecture:             x86_64
  CPU op-mode(s):         32-bit, 64-bit
  Address sizes:          36 bits physical, 48 bits virtual
  Byte Order:             Little Endian
CPU(s):                   4
  On-line CPU(s) list:    0-3
Vendor ID:                GenuineIntel
  Model name:             Intel(R) Core(TM) i5-3230M CPU @ 2.60GHz
    CPU family:           6
    Model:                58
    Thread(s) per core:   2
    Core(s) per socket:   2
    Socket(s):            1
    Stepping:             9
    CPU max MHz:          3200.0000
    CPU min MHz:          1200.0000
    BogoMIPS:             5187.90
    Flags:                fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts a
                          cpi mmx fxsr sse sse2 ss ht tm pbe syscall nx rdtscp lm constant_tsc arch_perfmon peb
                          s bts rep_good nopl xtopology nonstop_tsc cpuid aperfmperf pni pclmulqdq dtes64 monit
                          or ds_cpl vmx est tm2 ssse3 cx16 xtpr pdcm pcid sse4_1 sse4_2 x2apic popcnt tsc_deadl
                          ine_timer aes xsave avx f16c rdrand lahf_lm cpuid_fault epb pti ssbd ibrs ibpb stibp 
                          tpr_shadow vnmi flexpriority ept vpid fsgsbase smep erms xsaveopt dtherm ida arat pln
                           pts md_clear flush_l1d
Virtualization features:  
  Virtualization:         VT-x
Caches (sum of all):      
  L1d:                    64 KiB (2 instances)
  L1i:                    64 KiB (2 instances)
  L2:                     512 KiB (2 instances)
  L3:                     3 MiB (1 instance)
NUMA:                     
  NUMA node(s):           1
  NUMA node0 CPU(s):      0-3
Vulnerabilities:          
  Gather data sampling:   Not affected
  Itlb multihit:          KVM: Mitigation: VMX disabled
  L1tf:                   Mitigation; PTE Inversion; VMX conditional cache flushes, SMT vulnerable
  Mds:                    Mitigation; Clear CPU buffers; SMT vulnerable
  Meltdown:               Mitigation; PTI
  Mmio stale data:        Unknown: No mitigations
  Reg file data sampling: Not affected
  Retbleed:               Not affected
  Spec rstack overflow:   Not affected
  Spec store bypass:      Mitigation; Speculative Store Bypass disabled via prctl and seccomp
  Spectre v1:             Mitigation; usercopy/swapgs barriers and __user pointer sanitization
  Spectre v2:             Mitigation; Retpolines; IBPB conditional; IBRS_FW; STIBP conditional; RSB filling; PB
                          RSB-eIBRS Not affected; BHI Not affected
  Srbds:                  Vulnerable: No microcode
  Tsx async abort:        Not affected
```

### Code for Testing

```c
#include <stdio.h>
#include <stdint.h>
#include <string.h>

#define NROUND 100000

static const uint16_t crc16tab[256]= {
    0x0000,0x1021,0x2042,0x3063,0x4084,0x50a5,0x60c6,0x70e7,
    0x8108,0x9129,0xa14a,0xb16b,0xc18c,0xd1ad,0xe1ce,0xf1ef,
    0x1231,0x0210,0x3273,0x2252,0x52b5,0x4294,0x72f7,0x62d6,
    0x9339,0x8318,0xb37b,0xa35a,0xd3bd,0xc39c,0xf3ff,0xe3de,
    0x2462,0x3443,0x0420,0x1401,0x64e6,0x74c7,0x44a4,0x5485,
    0xa56a,0xb54b,0x8528,0x9509,0xe5ee,0xf5cf,0xc5ac,0xd58d,
    0x3653,0x2672,0x1611,0x0630,0x76d7,0x66f6,0x5695,0x46b4,
    0xb75b,0xa77a,0x9719,0x8738,0xf7df,0xe7fe,0xd79d,0xc7bc,
    0x48c4,0x58e5,0x6886,0x78a7,0x0840,0x1861,0x2802,0x3823,
    0xc9cc,0xd9ed,0xe98e,0xf9af,0x8948,0x9969,0xa90a,0xb92b,
    0x5af5,0x4ad4,0x7ab7,0x6a96,0x1a71,0x0a50,0x3a33,0x2a12,
    0xdbfd,0xcbdc,0xfbbf,0xeb9e,0x9b79,0x8b58,0xbb3b,0xab1a,
    0x6ca6,0x7c87,0x4ce4,0x5cc5,0x2c22,0x3c03,0x0c60,0x1c41,
    0xedae,0xfd8f,0xcdec,0xddcd,0xad2a,0xbd0b,0x8d68,0x9d49,
    0x7e97,0x6eb6,0x5ed5,0x4ef4,0x3e13,0x2e32,0x1e51,0x0e70,
    0xff9f,0xefbe,0xdfdd,0xcffc,0xbf1b,0xaf3a,0x9f59,0x8f78,
    0x9188,0x81a9,0xb1ca,0xa1eb,0xd10c,0xc12d,0xf14e,0xe16f,
    0x1080,0x00a1,0x30c2,0x20e3,0x5004,0x4025,0x7046,0x6067,
    0x83b9,0x9398,0xa3fb,0xb3da,0xc33d,0xd31c,0xe37f,0xf35e,
    0x02b1,0x1290,0x22f3,0x32d2,0x4235,0x5214,0x6277,0x7256,
    0xb5ea,0xa5cb,0x95a8,0x8589,0xf56e,0xe54f,0xd52c,0xc50d,
    0x34e2,0x24c3,0x14a0,0x0481,0x7466,0x6447,0x5424,0x4405,
    0xa7db,0xb7fa,0x8799,0x97b8,0xe75f,0xf77e,0xc71d,0xd73c,
    0x26d3,0x36f2,0x0691,0x16b0,0x6657,0x7676,0x4615,0x5634,
    0xd94c,0xc96d,0xf90e,0xe92f,0x99c8,0x89e9,0xb98a,0xa9ab,
    0x5844,0x4865,0x7806,0x6827,0x18c0,0x08e1,0x3882,0x28a3,
    0xcb7d,0xdb5c,0xeb3f,0xfb1e,0x8bf9,0x9bd8,0xabbb,0xbb9a,
    0x4a75,0x5a54,0x6a37,0x7a16,0x0af1,0x1ad0,0x2ab3,0x3a92,
    0xfd2e,0xed0f,0xdd6c,0xcd4d,0xbdaa,0xad8b,0x9de8,0x8dc9,
    0x7c26,0x6c07,0x5c64,0x4c45,0x3ca2,0x2c83,0x1ce0,0x0cc1,
    0xef1f,0xff3e,0xcf5d,0xdf7c,0xaf9b,0xbfba,0x8fd9,0x9ff8,
    0x6e17,0x7e36,0x4e55,0x5e74,0x2e93,0x3eb2,0x0ed1,0x1ef0
};

static const uint16_t tbl_reduced[16] = {
    0x0000,0x1081,0x2102,0x3183,0x4204,0x5285,0x6306,0x7387,0x8408,0x9489,0xa50a,0xb58b,0xc60c,0xd68d,0xe70e,0xf78f,
};

static const uint16_t tbl_forward_reduced[16] = { 
    0x0000,0x1021,0x2042,0x3063,0x4084,0x50a5,0x60c6,0x70e7,
    0x8108,0x9129,0xa14a,0xb16b,0xc18c,0xd1ad,0xe1ce,0xf1ef,
};

#define REVERSE_8_bit(n) \
        do { \
                    n = ((n & 0xf0f0f0f0) >> 4) | ((n & 0x0f0f0f0f) << 4); \
                    n = ((n & 0xcccccccc) >> 2) | ((n & 0x33333333) << 2); \
                    n = ((n & 0xaaaaaaaa) >> 1) | ((n & 0x55555555) << 1); \
                } while(0)

static inline uint8_t reverse_bit_8(uint8_t n) {
        REVERSE_8_bit(n);
            return n;
}

static inline uint16_t reverse_bit_16(uint16_t n) {
        n = ((n & 0xff00ff00) >> 8) | ((n & 0x00ff00ff) << 8);
            REVERSE_8_bit(n);
                return n;
}

__attribute__((noinline)) uint16_t crc16_orig(const char *buf, int len) {
    int counter;
    uint16_t crc = 0;
    for (counter = 0; counter < len; counter++)
            crc = (crc<<8) ^ crc16tab[((crc>>8) ^ *buf++)&0x00FF];
    return crc;
}


static inline uint16_t crc16_reverse_base(uint16_t crc, uint8_t v) {
        crc ^= v;
        crc = (crc >> 4) ^ tbl_reduced[crc & 0x000f];
        crc = (crc >> 4) ^ tbl_reduced[crc & 0x000f];
        return crc;
}

__attribute__((noinline)) uint16_t crc16_reverse(const char *buf, int len) {
    int counter;
    uint16_t crc = 0;
    for(counter = 0; counter < len; counter++) {
        uint8_t tmp = reverse_bit_8((uint8_t)buf[counter]);
        crc = crc16_reverse_base(crc, tmp);
    }
    return reverse_bit_16(crc); 
}

static inline uint16_t crc16_forward_base(uint16_t crc, uint8_t v) {
        crc ^= v << 8;
        crc = (crc << 4) ^ tbl_forward_reduced[crc >> 12];
        crc = (crc << 4) ^ tbl_forward_reduced[crc >> 12];
        return crc;
}

__attribute__((noinline)) static uint16_t crc16_forward(const char *buf, int len) {
    int counter;
    uint16_t crc = 0;
    for(counter = 0; counter < len; counter++) {
        uint8_t tmp = (uint8_t)buf[counter];
        crc = crc16_forward_base(crc, tmp);
    }  

    return crc;
}

__attribute__((always_inline))
static inline void escape(void *p) {
    __asm__ volatile ("" : : "r"(p) : "memory");
}

int main() {
    const char li[] = "Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed "
                         "do eiusmod tempor incididunt ut labore et dolore magna "
                         "aliqua. Ut enim ad minim veniam, quis nostrud exercitation "
                         "ullamco laboris nisi ut aliquip ex ea commodo consequat. Duis "
                         "aute irure dolor in reprehenderit in voluptate velit esse "
                         "cillum dolore eu fugiat nulla pariatur. Excepteur sint "
                         "occaecat cupidatat non proident, sunt in culpa qui officia "
                         "deserunt mollit anim id est laborum.";
    uint16_t ret;
    for(int i = 0; i < NROUND; i++) {
        escape(li);
        ret = crc16_forward(li, strlen(li));
        escape(li);
    }
    printf("0x%x\n", ret);
    return 0;
}
```

Cuda-Chen:
## Statistics of Each Implementation

```
# original
 Performance counter stats for './a.out':

       403,048,189      cycles                                                      
       402,922,173      instructions              #    1.00  insn per cycle         
            24,123      L1-dcache-load-misses                                       

       0.133358807 seconds time elapsed

       0.133398000 seconds user
       0.000000000 seconds sys
       
# forward
 Performance counter stats for './a.out':

       798,062,363      cycles                                                      
       713,941,689      instructions              #    0.89  insn per cycle         
            31,513      L1-dcache-load-misses                                       

       0.260542423 seconds time elapsed

       0.260686000 seconds user
       0.000000000 seconds sys
  
# reverse
 Performance counter stats for './a.out':

       688,547,303      cycles                                                      
     1,250,619,386      instructions              #    1.82  insn per cycle         
            35,560      L1-dcache-load-misses                                       

       0.221963854 seconds time elapsed

       0.222047000 seconds user
       0.000000000 seconds sys       
```

Cuda-Chen:
## Detailed L1 Cache Miss Counts of Each Implementation

### original

```
Samples: 56  of event 'L1-dcache-load-misses', 4000 Hz, Event count (approx.): 41442
crc16_orig  /home/jio/c_code/a.out [Percent: local period]
Period      │    ↓ jle     40                                                                                 ▒
            │      movslq  %esi,%rsi                                                                          ▒
            │    uint16_t crc = 0;                                                                            ▒
            │      xor     %eax,%eax                                                                          ▒
            │      lea     (%rdi,%rsi,1),%rcx                                                                 ▒
            │      lea     crc16tab,%rsi                                                                      ▒
            │      nop                                                                                        ▒
            │    crc = (crc<<8) ^ crc16tab[((crc>>8) ^ *buf++)&0x00FF];                                       ▒
            │20:   mov     %eax,%edx                                                                          ▒
            │      add     $0x1,%rdi                                                                          ▒
            │      shl     $0x8,%eax                                                                          ▒
        434 │      shr     $0x8,%dx                                                                           ▒
            │      xor     -0x1(%rdi),%dl                                                                     ▒
        957 │      movzbl  %dl,%edx                                                                           ▒
        575 │      xor     (%rsi,%rdx,2),%ax                                                                  ▒
            │    for (counter = 0; counter < len; counter++)                                                  ▒
       8761 │      cmp     %rcx,%rdi                                                                          ▒
            │    ↑ jne     20                                                                                 ▒
            │    ← ret                                                                                        ▒
            │      nop                                                                                        ▒
            │    uint16_t crc = 0;                                                                            ▒
            │40:   xor     %eax,%eax                                                                          ▒
            │    return crc;                                                                                  ▒
            │    }                                                                                            ▒
            │    ← ret
```

### half-byte LUT forward calculation

```
Samples: 94  of event 'L1-dcache-load-misses', 4000 Hz, Event count (approx.): 55209
crc16_forward.constprop.0  /home/jio/c_code/a.out [Percent: local period]
Period      │    for(counter = 0; counter < len; counter++) {                                                 ▒
            │      add    $0x1,%rdi                                                                           ▒
            │    crc16_forward_base():                                                                        ▒
            │    crc ^= v << 8;                                                                               ▒
            │      shl    $0x8,%edx                                                                           ▒
        701 │      xor    %edx,%eax                                                                           ▒
            │    crc = (crc << 4) ^ tbl_forward_reduced[crc >> 12];                                           ▒
            │      mov    %eax,%edx                                                                           ▒
            │      shl    $0x4,%eax                                                                           ▒
            │      shr    $0xc,%dx                                                                            ▒
            │      and    $0xf,%edx                                                                           ◆
            │      xor    (%rcx,%rdx,2),%ax                                                                   ▒
            │    crc = (crc << 4) ^ tbl_forward_reduced[crc >> 12];                                           ▒
       1158 │      mov    %eax,%edx                                                                           ▒
            │      shl    $0x4,%eax                                                                           ▒
        637 │      shr    $0xc,%dx                                                                            ▒
         58 │      and    $0xf,%edx                                                                           ▒
            │      xor    (%rcx,%rdx,2),%ax                                                                   ▒
            │    crc16_forward():                                                                             ▒
            │    for(counter = 0; counter < len; counter++) {                                                 ▒
       7255 │      cmp    %rdi,%rsi                                                                           ▒
            │    ↑ jne    10                                                                                  ▒
            │    }                                                                                            ▒
            │                                                                                                 ▒
            │    return crc;
```

### half-byte LUT reverse calculation

```
Samples: 82  of event 'L1-dcache-load-misses', 4000 Hz, Event count (approx.): 49869
crc16_reverse  /home/jio/c_code/a.out [Percent: local period]
Period      │    for(counter = 0; counter < len; counter++) {                                                 ▒
            │      test    %esi,%esi                                                                          ▒
            │    ↓ jle     c0                                                                                 ▒
       3132 │      lea     -0x1(%rsi),%eax                                                                    ▒
            │    uint16_t crc = 0;                                                                            ▒
            │      xor     %r8d,%r8d                                                                          ▒
            │      lea     tbl_reduced,%rcx                                                                   ▒
            │      lea     0x1(%rdi,%rax,1),%rsi                                                              ▒
            │      xchg    %ax,%ax                                                                            ▒
            │    reverse_bit_8():                                                                             ▒
            │    REVERSE_8_bit(n);                                                                            ◆
            │20:   movzbl  (%rdi),%eax                                                                        ▒
            │    crc16_reverse():                                                                             ▒
            │    for(counter = 0; counter < len; counter++) {                                                 ▒
            │      add     $0x1,%rdi                                                                          ▒
            │    reverse_bit_8():                                                                             ▒
            │    REVERSE_8_bit(n);                                                                            ▒
            │      rol     $0x4,%al                                                                           ▒
       1692 │      mov     %eax,%edx                                                                          ▒
            │      shl     $0x2,%eax                                                                          ▒
            │      shr     $0x2,%dl                                                                           ▒
            │      and     $0xffffffcc,%eax                                                                   ▒
        966 │      and     $0x33,%edx                                                                         ▒
            │      or      %eax,%edx                                                                          ▒
            │      mov     %edx,%eax
            |      add     %edx,%edx                                                                          ▒
       1450 │      shr     %al                                                                                ▒
            │      and     $0xffffffaa,%edx                                                                   ▒
            │      and     $0x55,%eax                                                                         ▒
            │      or      %edx,%eax                                                                          ▒
            │    crc16_reverse_base():                                                                        ▒
            │    crc ^= v;                                                                                    ▒
       1643 │      movzbl  %al,%eax                                                                           ▒
            │      xor     %r8d,%eax                                                                          ▒
            │    crc = (crc >> 4) ^ tbl_reduced[crc & 0x000f];                                                ◆
            │      mov     %eax,%edx                                                                          ▒
            │      and     $0xf,%eax                                                                          ▒
       1180 │      shr     $0x4,%dx                                                                           ▒
            │      xor     (%rcx,%rax,2),%dx                                                                  ▒
        546 │      mov     %edx,%eax                                                                          ▒
            │    crc = (crc >> 4) ^ tbl_reduced[crc & 0x000f];                                                ▒
        281 │      mov     %edx,%r8d                                                                          ▒
            │      and     $0xf,%eax                                                                          ▒
        220 │      shr     $0x4,%r8w                                                                          ▒
            │      xor     (%rcx,%rax,2),%r8w                                                                 ▒
            │    crc16_reverse():                                                                             ▒
            │    for(counter = 0; counter < len; counter++) {                                                 ▒
       1178 │      cmp     %rdi,%rsi                                                                          ▒
            │    ↑ jne     20
```

zuiderkwast:
Sorry for the delay. I have time off now during the summer. I will get to it eventually.

I'm looking at your micro benchmark. The reason the large LUT is faster is because the LUT is in L1 cache, isn't it?  We should test this in a real scenario to see if the LUT will be in L1 cache or not. If it isn't in L1 cache, then I guess the small LUT will be faster, for a ~20 byte payload.

Another idea (regardless of big or small LUT) can be to prefetch the LUT into L1 cache some short time before we need it, i.e. we could provide some API to call `__builtin_prefetch(LUT)`, so later when we do the computation, it's already in L1.

I hope more people can find the time to look at this. @hpatro?

Cuda-Chen:
Hi @zuiderkwast ,

> Sorry for the delay. I have time off now during the summer. I will get to it eventually.

It's okey. I thought you're being caught by more than 10 issues for reviewing.

> The reason the large LUT is faster is because the LUT is in L1 cache, isn't it?

After thinking for half of day, for my personal guessing, here are what happens under the hood:
1. For the original implementation, we have a table with 256 entries occupying 512 Bytes of memory. An ordinary L1 cache is 64 Bytes, so we can store at most 32 entries (each entry costs 2 Bytes). Recall that CRC is basically division (XOR operation, precisely), if the permutation of input bytes just need to take these 32 entries already stored in L1 cache, then the original implementation will perform in less running time compared to current proposed implementations.
2. For the proposed implementations, there is a bottleneck at computing the CRC value: we need to compute twice, and these computations need to be done sequentially.

> Another idea (regardless of big or small LUT) can be to prefetch the LUT into L1 cache some short time before we need it, i.e. we could provide some API to call __builtin_prefetch(LUT), so later when we do the computation, it's already in L1.

I consider this might be a direction. What's more, I come up with an idea: we can choose computing method based on the "entropy" (such as the distance of each byte) of input bytes for prefetching the content of LUT.

Cuda-Chen:
I just have been caught by an interview these days. I will start other benchmarks after this week.

hpatro:
@Cuda-Chen Thanks for the update. Feel free to take your time and share the results at your ease.

Cuda-Chen:
Hi @zuiderkwast and @hpatro ,

I would like to make updates on benchmarking.
I will put the result and my findings in the following comments.

For running the benchmark I use `./runtest --single integration/valkey-benchmark`.

Cuda-Chen:
## Benchmark Results

### original

```
$ ./runtest --single integration/valkey-benchmark               
Starting test server at port 21079                
[ready]: 37876                                         
Testing integration/valkey-benchmark                                                                           
[ready]: 37877                                         
[ready]: 37878                                                                                                 
[ready]: 37879                                         
[ready]: 37880                                         
[ready]: 37881
[ready]: 37882            
[ready]: 37883
[ready]: 37884                                         
[ready]: 37885                                         
[ready]: 37886
[ready]: 37887                                         
[ready]: 37888
[ready]: 37889                                         
[ready]: 37890                                         
[ready]: 37891
[ok]: benchmark: set,get (7 ms)
[ok]: benchmark: connecting using URI set,get (7 ms)
[ok]: benchmark: connecting using URI with authentication set,get (7 ms)
[ok]: benchmark: full test suite (97 ms)
[ok]: benchmark: multi-thread set,get (512 ms)
[ok]: benchmark: pipelined full set,get (100 ms)
[ok]: benchmark: arbitrary command (7 ms)
[ok]: benchmark: arbitrary command sequence (7 ms)
value:VXKeHogKgJ=[5V9_X^b?48OKF2jGA<f:iR@50o7dS3
[ok]: benchmark: arbitrary command with data placeholder (7 ms)
[ok]: benchmark: keyspace length (35 ms)
[ok]: benchmark: clients idle mode should return error when reached maxclients limit (5 ms)
[ok]: benchmark: read last argument from stdin (14 ms)
[1/1 done]: integration/valkey-benchmark (1 seconds)

                   The End

  1 seconds - integration/valkey-benchmark

\o/ All tests passed without errors!

```

### half-byte LUT forward calculation

```
$ ./runtest --single integration/valkey-benchmark
Starting test server at port 21079
[ready]: 16624
Testing integration/valkey-benchmark
[ready]: 16625
[ready]: 16623
[ready]: 16626
[ready]: 16628
[ready]: 16629
[ready]: 16627
[ready]: 16630
[ready]: 16631
[ready]: 16633
[ready]: 16634
[ready]: 16635
[ready]: 16636
[ready]: 16637
[ready]: 16632
[ready]: 16638
[ok]: benchmark: set,get (6 ms)
[ok]: benchmark: connecting using URI set,get (5 ms)
[ok]: benchmark: connecting using URI with authentication set,get (7 ms)
[ok]: benchmark: full test suite (92 ms)
[ok]: benchmark: multi-thread set,get (520 ms)
[ok]: benchmark: pipelined full set,get (100 ms)
[ok]: benchmark: arbitrary command (6 ms)
[ok]: benchmark: arbitrary command sequence (10 ms)
value:VXKeHogKgJ=[5V9_X^b?48OKF2jGA<f:iR@50o7dS3
[ok]: benchmark: arbitrary command with data placeholder (8 ms)
[ok]: benchmark: keyspace length (37 ms)
[ok]: benchmark: clients idle mode should return error when reached maxclients limit (3 ms)
[ok]: benchmark: read last argument from stdin (12 ms)
[1/1 done]: integration/valkey-benchmark (1 seconds)

                   The End

  1 seconds - integration/valkey-benchmark

\o/ All tests passed without errors!

```

### half-byte LUT reverse calculation 

```
$ ./runtest --single integration/valkey-benchmark               
Starting test server at port 21079                
[ready]: 32023                                         
Testing integration/valkey-benchmark                                                   
[ready]: 32025                                         
[ready]: 32024                                                                                                 
[ready]: 32026                                         
[ready]: 32029                                         
[ready]: 32027             
[ready]: 32028            
[ready]: 32031             
[ready]: 32030                                         
[ready]: 32032                                         
[ready]: 32034             
[ready]: 32033                                         
[ready]: 32035             
[ready]: 32036                                         
[ready]: 32037                                         
[ready]: 32038                                         
[ok]: benchmark: set,get (8 ms)                        
[ok]: benchmark: connecting using URI set,get (6 ms)
[ok]: benchmark: connecting using URI with authentication set,get (7 ms)                                       
[ok]: benchmark: full test suite (99 ms)               
[ok]: benchmark: multi-thread set,get (512 ms)
[ok]: benchmark: pipelined full set,get (110 ms)
[ok]: benchmark: arbitrary command (9 ms)
[ok]: benchmark: arbitrary command sequence (7 ms)
value:VXKeHogKgJ=[5V9_X^b?48OKF2jGA<f:iR@50o7dS3
[ok]: benchmark: arbitrary command with data placeholder (8 ms)
[ok]: benchmark: keyspace length (33 ms)
[ok]: benchmark: clients idle mode should return error when reached maxclients limit (5 ms)
[ok]: benchmark: read last argument from stdin (13 ms)
[1/1 done]: integration/valkey-benchmark (1 seconds)

                   The End

  1 seconds - integration/valkey-benchmark

\o/ All tests passed without errors!

```

Cuda-Chen:
## Findings

- The running time between original, forward, and reverse calculation shows little difference.
  - As reverse needs additional bit-reversing operations, I would like to stick to forward calculation for coding logic simplicity.

## What I will Do Next

First will be prefetching LUT as mentioned in [the previous comment](https://github.com/valkey-io/valkey/pull/2300#issuecomment-3067310313).

What's more, I would like to conduct experiment on prefetching input data as well (as the half-byte LUT takes 32 Bytes, it may be beneficial to prefetch input data as we set the maximum size of input data to 32 Bytes, utilizing the whole 64 Bytes of L1 cache).

zuiderkwast:
> For running the benchmark I use `./runtest --single integration/valkey-benchmark`.

This just runs the integration test of the benchmark tool. It just tests that the benchmark tool is working correctly. It does not run any real benchmark.

To run benchmarks, use `src/valkey-benchmark` with a running server started separately using `src/valkey-server`. Use `src/valkey-benchmark --help` for help.

Additionally, the server needs to run in cluster mode. Otherwise, CRC-16 is not used at all.

zuiderkwast:
Actually it might be very difficult to see any difference at all in a benchmark.

I think is a very difficult thing to benchmark these differences without good knowledge about large parts of this code base. I guess we would need to use perf or flamegraph to see the part of execution that consists of the crc16 computation. The crc16 computation is already a very small part of the total execution time.

I wish I had some time to assist you with this, but I'm off for vacation. I guess others are off too during the summer. Maybe it's better to wait for later when people are back from vacation.

> ## What I will Do Next
> First will be prefetching LUT as mentioned in [the previous comment](https://github.com/valkey-io/valkey/pull/2300#issuecomment-3067310313).

To understand when to call this, we need to read code where the crc16 is called and read backwards to find a good point in time to do the prefetching. It should be done early enough to give enough time for the memory to be read into L1 cache in the background while some other code is executing. Then, when the crc16 code needs the LUT it will already be in L1 cache. We can't really do it inside crc16 itself just before we need to use it. That would be too late.

The main place crc16 is called is from keyHashSlot() which is called from clusterSlotByCommand() in cluster.c. My guess is that the beginning of clusterSlotByCommand() can be a good place to prefetch the LUT. It could be enough time before the crc16 will actually access the LUT.

> What's more, I would like to conduct experiment on prefetching input data as well (as the half-byte LUT takes 32 Bytes, it may be beneficial to prefetch input data as we set the maximum size of input data to 32 Bytes, utilizing the whole 64 Bytes of L1 cache).

64 bytes is not the total size of the L1 cache. It is the size of one cache line. Memory is fetched to cache in consecutive blocks of this size. The LUT and the input data will almost definitely be in different cache lines because they are not stored together in the same 64 byte range in memory.

The total L1 cache size is small, but not that small. At least it can store some hundred cache lines, or something like that.

Btw, the input has typically just been accessed by the command parser before the crc16 calculation happens, so I would assume it is already in the L1 cache, or at least in L2 or L3 cache, so my guess is that it is not useful to do prefetching of the input data.

Cuda-Chen:
Hi @zuiderkwast ,

> > For running the benchmark I use `./runtest --single integration/valkey-benchmark`.
> 
> This just runs the integration test of the benchmark tool. It just tests that the benchmark tool is working correctly. It does not run any real benchmark.
> 
> To run benchmarks, use `src/valkey-benchmark` with a running server started separately using `src/valkey-server`. Use `src/valkey-benchmark --help` for help.
> 
> Additionally, the server needs to run in cluster mode. Otherwise, CRC-16 is not used at all.

Thanks for pointing me out. I will adapt myself for the benchmarking programs.

> I wish I had some time to assist you with this, but I'm off for vacation.

Got it. Just follow your own paces.

> To understand when to call this, we need to read code where the crc16 is called and read backwards to find a good point in time to do the prefetching.

Thanks for the hint. I will mark the functions that calls crc16 and make some experiments.

Cuda-Chen:
Hi @zuiderkwast and @hpatro ,

I run the cluster mode for benchmarking using these two methods:

1. Create a tutorial cluster mentined in [ValKey cluster tutorial](https://valkey.io/topics/cluster-tutorial/)
2. Create a cluster using `./utils/create-cluster start` and `./utils/create-cluster create`.

Nevertheless, whenever I benchmark using `valkey-benchmark` I always receive the error of `Error from server: MOVED 13782 127.0.0.1:30003`:

```bash
$ ./src/valkey-benchmark -p 30001
====== PING_INLINE ======                                                   
  100000 requests completed in 2.19 seconds
  50 parallel clients
  3 bytes payload
  keep alive: 1
  host configuration "appendonly": yes
  multi-thread: no

0.000% <= 0.127 milliseconds (cumulative count 1)
50.000% <= 0.663 milliseconds (cumulative count 51553)
75.000% <= 0.759 milliseconds (cumulative count 76944)
87.500% <= 0.871 milliseconds (cumulative count 87946)
93.750% <= 0.983 milliseconds (cumulative count 93948)
96.875% <= 1.199 milliseconds (cumulative count 96917)
98.438% <= 1.527 milliseconds (cumulative count 98447)
99.219% <= 1.887 milliseconds (cumulative count 99233)
99.609% <= 2.407 milliseconds (cumulative count 99610)
99.805% <= 3.383 milliseconds (cumulative count 99806)
99.902% <= 4.999 milliseconds (cumulative count 99903)
99.951% <= 23.311 milliseconds (cumulative count 99952)
99.976% <= 24.159 milliseconds (cumulative count 99976)
99.988% <= 24.511 milliseconds (cumulative count 99988)
99.994% <= 24.719 milliseconds (cumulative count 99994)
99.997% <= 24.815 milliseconds (cumulative count 99997)
99.998% <= 24.895 milliseconds (cumulative count 99999)
99.999% <= 24.927 milliseconds (cumulative count 100000)
100.000% <= 24.927 milliseconds (cumulative count 100000)

0.007% <= 0.207 milliseconds (cumulative count 7)
0.087% <= 0.303 milliseconds (cumulative count 87)
1.614% <= 0.407 milliseconds (cumulative count 1614)
5.995% <= 0.503 milliseconds (cumulative count 5995)
35.850% <= 0.607 milliseconds (cumulative count 35850)
60.649% <= 0.703 milliseconds (cumulative count 60649)
83.229% <= 0.807 milliseconds (cumulative count 83229)
90.072% <= 0.903 milliseconds (cumulative count 90072)
94.560% <= 1.007 milliseconds (cumulative count 94560)
96.024% <= 1.103 milliseconds (cumulative count 96024)
96.979% <= 1.207 milliseconds (cumulative count 96979)
97.617% <= 1.303 milliseconds (cumulative count 97617)
98.051% <= 1.407 milliseconds (cumulative count 98051)
98.377% <= 1.503 milliseconds (cumulative count 98377)
98.666% <= 1.607 milliseconds (cumulative count 98666)
98.894% <= 1.703 milliseconds (cumulative count 98894)
99.106% <= 1.807 milliseconds (cumulative count 99106)
99.262% <= 1.903 milliseconds (cumulative count 99262)
99.371% <= 2.007 milliseconds (cumulative count 99371)
99.448% <= 2.103 milliseconds (cumulative count 99448)
99.762% <= 3.103 milliseconds (cumulative count 99762)
99.854% <= 4.103 milliseconds (cumulative count 99854)
99.917% <= 5.103 milliseconds (cumulative count 99917)
99.950% <= 6.103 milliseconds (cumulative count 99950)
99.975% <= 24.111 milliseconds (cumulative count 99975)
100.000% <= 25.103 milliseconds (cumulative count 100000)

Summary:
  throughput summary: 45745.65 requests per second
  latency summary (msec):
        0.717     0.120     0.663     1.031     1.751    24.927
====== PING_MBULK ======                                                   
  100000 requests completed in 2.26 seconds
  50 parallel clients
  3 bytes payload
  keep alive: 1
  host configuration "appendonly": yes
  multi-thread: no

0.000% <= 0.127 milliseconds (cumulative count 1)
50.000% <= 0.623 milliseconds (cumulative count 52206)
75.000% <= 0.719 milliseconds (cumulative count 75436)
87.500% <= 0.831 milliseconds (cumulative count 87900)
93.750% <= 0.951 milliseconds (cumulative count 93787)
96.875% <= 1.255 milliseconds (cumulative count 96887)
98.438% <= 1.703 milliseconds (cumulative count 98461)
99.219% <= 2.063 milliseconds (cumulative count 99229)
99.609% <= 2.399 milliseconds (cumulative count 99610)
99.805% <= 2.735 milliseconds (cumulative count 99807)
99.902% <= 3.039 milliseconds (cumulative count 99903)
99.951% <= 11.263 milliseconds (cumulative count 99952)
99.976% <= 11.807 milliseconds (cumulative count 99976)
99.988% <= 12.159 milliseconds (cumulative count 99988)
99.994% <= 12.319 milliseconds (cumulative count 99994)
99.997% <= 12.399 milliseconds (cumulative count 99997)
99.998% <= 12.455 milliseconds (cumulative count 99999)
99.999% <= 12.495 milliseconds (cumulative count 100000)
100.000% <= 12.495 milliseconds (cumulative count 100000)

0.031% <= 0.207 milliseconds (cumulative count 31)
0.220% <= 0.303 milliseconds (cumulative count 220)
3.032% <= 0.407 milliseconds (cumulative count 2032)
5.436% <= 0.503 milliseconds (cumulative count 5436)
40.825% <= 0.607 milliseconds (cumulative count 40825)
72.259% <= 0.703 milliseconds (cumulative count 72259)
86.140% <= 0.807 milliseconds (cumulative count 86140)
91.930% <= 0.903 milliseconds (cumulative count 91930)
94.869% <= 1.007 milliseconds (cumulative count 94869)
95.926% <= 1.103 milliseconds (cumulative count 95926)
96.580% <= 1.207 milliseconds (cumulative count 96580)
97.132% <= 1.303 milliseconds (cumulative count 97132)
97.603% <= 1.407 milliseconds (cumulative count 97603)
97.919% <= 1.503 milliseconds (cumulative count 97919)
98.215% <= 1.607 milliseconds (cumulative count 98215)
98.461% <= 1.703 milliseconds (cumulative count 98461)
98.739% <= 1.807 milliseconds (cumulative count 98739)
98.971% <= 1.903 milliseconds (cumulative count 98971)
99.143% <= 2.007 milliseconds (cumulative count 99143)
99.289% <= 2.103 milliseconds (cumulative count 99289)
99.911% <= 3.103 milliseconds (cumulative count 99911)
99.950% <= 4.103 milliseconds (cumulative count 99950)
99.986% <= 12.103 milliseconds (cumulative count 99986)
100.000% <= 13.103 milliseconds (cumulative count 100000)

Summary:
  throughput summary: 44247.79 requests per second
  latency summary (msec):
        0.694     0.120     0.623     1.023     1.919    12.495
Error from server: MOVED 13782 127.0.0.1:30003
```

I would like to realize whether there are more additional steps for benckmarking with cluster mode.

zuiderkwast:
You need the `--cluster` flag to valkey-benchmark to make it aware that it's a cluster.

```shell
$ ./src/valkey-benchmark --cluster -p 30001
```

Cuda-Chen:
Hi @zuiderkwast ,

Thanks for your help!

Now the `valkey-benchmark` will run for more tests but always fail with `Error from server 127.0.0.1:30001: MOVED 9189 127.0.0.1:30002` in `FUNCTION_LOAD` part:

```bash
$ ./src/valkey-benchmark --cluster -p 30001
<ommitted>
====== FUNCTION LOAD ======                                                   
  100000 requests completed in 10.31 seconds
  50 parallel clients
  3 bytes payload
  keep alive: 1
  cluster mode: yes (3 primary)
  node [0] configuration:
    save: 3600 1 300 100 60 10000
    appendonly: yes
  node [1] configuration:
    save: 3600 1 300 100 60 10000
    appendonly: yes
  node [2] configuration:
    save: 3600 1 300 100 60 10000
    appendonly: yes
  multi-thread: yes
  threads: 3

0.000% <= 0.279 milliseconds (cumulative count 3)
50.000% <= 3.879 milliseconds (cumulative count 50060)
75.000% <= 5.887 milliseconds (cumulative count 75032)
87.500% <= 7.999 milliseconds (cumulative count 87507)
93.750% <= 10.767 milliseconds (cumulative count 93754)
96.875% <= 13.679 milliseconds (cumulative count 96875)
98.438% <= 16.879 milliseconds (cumulative count 98441)
99.219% <= 19.887 milliseconds (cumulative count 99221)
99.609% <= 23.375 milliseconds (cumulative count 99610)
99.805% <= 27.391 milliseconds (cumulative count 99805)
99.902% <= 30.719 milliseconds (cumulative count 99907)
99.951% <= 33.759 milliseconds (cumulative count 99952)
99.976% <= 35.839 milliseconds (cumulative count 99977)
99.988% <= 36.095 milliseconds (cumulative count 99988)
99.994% <= 43.839 milliseconds (cumulative count 99994)
99.997% <= 44.095 milliseconds (cumulative count 99997)
99.998% <= 44.255 milliseconds (cumulative count 99999)
99.999% <= 44.351 milliseconds (cumulative count 100000)
100.000% <= 44.351 milliseconds (cumulative count 100000)

0.016% <= 0.303 milliseconds (cumulative count 16)
0.038% <= 0.407 milliseconds (cumulative count 38)
0.050% <= 0.503 milliseconds (cumulative count 50)
0.066% <= 0.607 milliseconds (cumulative count 66)
0.079% <= 0.703 milliseconds (cumulative count 79)
0.119% <= 0.807 milliseconds (cumulative count 119)
0.182% <= 0.903 milliseconds (cumulative count 182)
0.366% <= 1.007 milliseconds (cumulative count 366)
0.619% <= 1.103 milliseconds (cumulative count 619)
1.009% <= 1.207 milliseconds (cumulative count 1009)
1.692% <= 1.303 milliseconds (cumulative count 1692)
2.979% <= 1.407 milliseconds (cumulative count 2979)
4.190% <= 1.503 milliseconds (cumulative count 4190)
5.376% <= 1.607 milliseconds (cumulative count 5376)
6.183% <= 1.703 milliseconds (cumulative count 6183)
6.887% <= 1.807 milliseconds (cumulative count 6887)
7.487% <= 1.903 milliseconds (cumulative count 7487)
8.157% <= 2.007 milliseconds (cumulative count 8157)
8.785% <= 2.103 milliseconds (cumulative count 8785)
31.510% <= 3.103 milliseconds (cumulative count 31510)
53.721% <= 4.103 milliseconds (cumulative count 53721)
66.438% <= 5.103 milliseconds (cumulative count 66438)
76.821% <= 6.103 milliseconds (cumulative count 76821)
83.593% <= 7.103 milliseconds (cumulative count 83593)
87.899% <= 8.103 milliseconds (cumulative count 87899)
90.633% <= 9.103 milliseconds (cumulative count 90633)
92.716% <= 10.103 milliseconds (cumulative count 92716)
94.198% <= 11.103 milliseconds (cumulative count 94198)
95.462% <= 12.103 milliseconds (cumulative count 95462)
96.453% <= 13.103 milliseconds (cumulative count 96453)
97.133% <= 14.103 milliseconds (cumulative count 97133)
97.704% <= 15.103 milliseconds (cumulative count 97704)
98.173% <= 16.103 milliseconds (cumulative count 98173)
98.505% <= 17.103 milliseconds (cumulative count 98505)
98.795% <= 18.111 milliseconds (cumulative count 98795)
99.017% <= 19.103 milliseconds (cumulative count 99017)
99.241% <= 20.111 milliseconds (cumulative count 99241)
99.403% <= 21.103 milliseconds (cumulative count 99403)
99.531% <= 22.111 milliseconds (cumulative count 99531)
99.600% <= 23.103 milliseconds (cumulative count 99600)
99.659% <= 24.111 milliseconds (cumulative count 99659)
99.695% <= 25.103 milliseconds (cumulative count 99695)
99.764% <= 26.111 milliseconds (cumulative count 99764)
99.802% <= 27.103 milliseconds (cumulative count 99802)
99.833% <= 28.111 milliseconds (cumulative count 99833)
99.844% <= 29.103 milliseconds (cumulative count 99844)
99.880% <= 30.111 milliseconds (cumulative count 99880)
99.931% <= 31.103 milliseconds (cumulative count 99931)
99.932% <= 33.119 milliseconds (cumulative count 99932)
99.957% <= 34.111 milliseconds (cumulative count 99957)
99.970% <= 35.103 milliseconds (cumulative count 99970)
99.988% <= 36.127 milliseconds (cumulative count 99988)
99.993% <= 37.119 milliseconds (cumulative count 99993)
99.997% <= 44.127 milliseconds (cumulative count 99997)
100.000% <= 45.119 milliseconds (cumulative count 100000)

Summary:
  throughput summary: 9699.32 requests per second
  latency summary (msec):
        4.971     0.272     3.879    11.775    19.039    44.351
Error from server 127.0.0.1:30001: MOVED 9189 127.0.0.1:30002
```

I also write my benchmark processes:

1. Run `utils/create-cluster start`.
2. Run `utils/create-cluster create`.
  - This step fails after the first time of creating cluster, so I omit this step when benchmarking after creating cluster for the first time.
3. Run `src/valkey-benchmark --cluster -p 30001`.

hpatro:
@roshkhatri trying out the benchmark-on-demand functionality. 🤞

Cuda-Chen:
Hi @zuiderkwast and @hpatro ,

I would like to share some findings on prefetching.
In short, the half-byte table with prefetching reduces `L1-dcache-load-misses` at least 20% compared to the original implementation (1293734 / 1660821 = 0.778972568, should be 23% precisely).

The following table shows `L1-dcache-load-misses` (measured in period counts) in `crc16()`:

| original | original with prefetching | original with prefetching ASCII | half byte | half byte with prefetching |
| --- | --- | --- | --- | --- |
| 1660821 | 1942356 | 1816882 | 2174780 | 1293734 |

FYI, the "original with prefetching ASCII" means to prefetching the first 128 entries of LUT (I assume each byte in input message is in the range of ASCII).

## Test Procedures

1. Start a cluster using `utils/create-cluster start`.
2. Create a cluster using `utils/create-cluster create`.
3. Get the PID of one primary node using `ps aux | grep valkey-server`.
    - I choose the node opening port on 30001.
4. Run Perf using `sudo perf record -g -e cycles,instructions,L1-dcache-load-misses -p <PID in Step 3>`.
5. Benchmark using `src/valkey-benchmark --cluster -p <the port of one of primary node in Step 3>`.
6. Stop the cluster using `utils/create-cluster stop`.
    - The Perf program created in Step 4 will automatically close.
7. Get the Perf report using `sudo perf report -g graph,0.5,caller`
    - Type `t` after annotating the functions to show the metric from percent to period counts. 

## Period Counts in Each Scenario

### original

<details>

```
Period      │                                                                                     ▒
            │     crc16():                                                                                    ▒
       4341 │       cmp     $0x1,%ecx                                                                         ▒
            │     ↓ jle     f0                                                                                ▒
      14974 │       lea     -0x2(%rcx),%eax                                                                   ▒
            │       lea     crc16tab.lto_priv.0,%rsi                                                          ▒
     101534 │       lea     0x1(%rdx,%rax,1),%rdi                                                             ▒
      38324 │       xor     %eax,%eax                                                                         ▒
            │       cs      nopw 0x0(%rax,%rax,1)                                                             ▒
      10952 │ 90:   mov     %eax,%ecx                                                                         ▒
            │       add     $0x1,%rdx                                                                         ▒
      33605 │       shl     $0x8,%eax                                                                         ▒
      58766 │       shr     $0x8,%cx                                                                          ▒
      19480 │       xor     -0x1(%rdx),%cl                                                                    ▒
     140316 │       movzbl  %cl,%ecx                                                                          ◆
     110152 │       xor     (%rsi,%rcx,2),%ax                                                                 ▒
    1128377 │       cmp     %rdi,%rdx                                                                         ▒
            │     ↑ jne     90

```

</details>

### original with prefetching

<details>

```
Period      │     crc16():                                                                                    ▒
      14449 │       cmp     $0x1,%ecx                                                                         ▒
            │     ↓ jle     f0                                                                                ▒
            │       lea     -0x2(%rcx),%eax                                                                   ▒
            │       lea     crc16tab,%rsi                                                                     ▒
     101633 │       lea     0x1(%rdx,%rax,1),%rdi                                                             ▒
      27217 │       xor     %eax,%eax                                                                         ▒
            │       cs      nopw 0x0(%rax,%rax,1)                                                             ▒
      35103 │ 90:   mov     %eax,%ecx                                                                         ▒
      11208 │       add     $0x1,%rdx                                                                         ▒
      44787 │       shl     $0x8,%eax                                                                         ▒
      74518 │       shr     $0x8,%cx                                                                          ▒
      22719 │       xor     -0x1(%rdx),%cl                                                                    ▒
     200325 │       movzbl  %cl,%ecx                                                                          ▒
      92031 │       xor     (%rsi,%rcx,2),%ax                                                                 ▒
    1318366 │       cmp     %rdi,%rdx                                                                         ▒
            │     ↑ jne     90

```
</details>


### original with prefetching ASCII

<details>

```
Period      │     crc16():                                                                                    ▒
            │     };                                                                                          ▒
            │                                                                                                 ▒
            │     uint16_t crc16(const char *buf, int len) {                                                  ▒
            │     int counter;                                                                                ▒
            │     uint16_t crc = 0;                                                                           ▒
            │     for (counter = 0; counter < len; counter++)                                                 ▒
      21832 │       cmp     $0x1,%ecx                                                                         ▒
            │     ↓ jle     f0                                                                                ▒
       8239 │       lea     -0x2(%rcx),%eax                                                                   ▒
       5257 │       lea     crc16tab,%rsi                                                                     ▒
      18839 │       lea     0x1(%rdx,%rax,1),%rdi                                                             ▒
            │     uint16_t crc = 0;                                                                           ▒
      34404 │       xor     %eax,%eax                                                                         ▒
            │       cs      nopw 0x0(%rax,%rax,1)                                                             ◆
            │     crc = (crc<<8) ^ crc16tab[((crc>>8) ^ *buf++)&0x00FF];                                      ▒
      16072 │ 90:   mov     %eax,%ecx                                                                         ▒
      22349 │       add     $0x1,%rdx                                                                         ▒
      88006 │       shl     $0x8,%eax                                                                         ▒
      30951 │       shr     $0x8,%cx                                                                          ▒
      25537 │       xor     -0x1(%rdx),%cl                                                                    ▒
     209896 │       movzbl  %cl,%ecx                                                                          ▒
      56213 │       xor     (%rsi,%rcx,2),%ax                                                                 ▒
            │     for (counter = 0; counter < len; counter++)                                                 ▒
    1279287 │       cmp     %rdi,%rdx                                                                         ▒
            │     ↑ jne     90                                                                                ▒
```

</details>

### half byte

<details>

```
Period      │     crc16():                                                                                    ▒
            │     * Output for "123456789"     : 31C3                                                         ▒
            │     */                                                                                          ▒
            │     uint16_t crc16(const char *buf, int len) {                                                  ▒
            │     int counter;                                                                                ▒
            │     uint16_t crc = 0;                                                                           ▒
            │     for (counter = 0; counter < len; counter++) {                                               ▒
      52508 │       cmp     $0x1,%edx                                                                         ▒
            │     ↓ jle     128                                                                               ▒
      55700 │       lea     -0x2(%rdx),%eax                                                                   ▒
       4920 │       add     %rdi,%rcx                                                                         ▒
      51043 │       lea     0x2(%rdi,%r8,1),%rdi                                                              ◆
      47092 │       add     %rax,%rdi                                                                         ▒
      49250 │       lea     crc16_tbl.lto_priv.0,%rdx                                                         ▒
            │     uint16_t crc = 0;                                                                           ▒
            │       xor     %eax,%eax                                                                         ▒
       5380 │       xchg    %ax,%ax                                                                           ▒
            │     uint8_t tmp = (uint8_t)buf[counter];                                                        ▒
            │     crc = crc16_base(crc, tmp);                                                                 ▒
      35173 │ 90:   movzbl  (%rcx),%esi                                                                       ▒
            │     for (counter = 0; counter < len; counter++) {                                               ▒
      38302 │       add     $0x1,%rcx                                                                         ▒
            │     crc16_base():                                                                               ▒
            │     crc ^= v << 8;                                                                              ▒
      89547 │       shl     $0x8,%esi                                                                         ▒
     160026 │       xor     %esi,%eax                                                                         ▒
            │     crc = (crc << 4) ^ crc16_tbl[crc >> 12];                                                    ▒
      24457 │       mov     %eax,%esi                                                                         ▒
      14759 │       shl     $0x4,%eax                                                                         ▒
      77460 │       shr     $0xc,%si                                                                          ▒
      28913 │       and     $0xf,%esi                                                                         ▒
     105206 │       xor     (%rdx,%rsi,2),%ax                                                                 ▒
            │     crc = (crc << 4) ^ crc16_tbl[crc >> 12];                                                    ▒
     574675 │       mov     %eax,%esi                                                                         ▒
      28089 │       shl     $0x4,%eax                                                                         ◆
     114033 │       shr     $0xc,%si                                                                          ▒
      24910 │       and     $0xf,%esi                                                                         ▒
      72964 │       xor     (%rdx,%rsi,2),%ax                                                                 ▒
            │     crc16():                                                                                    ▒
            │     for (counter = 0; counter < len; counter++) {                                               ▒
     520373 │       cmp     %rcx,%rdi                                                                         ▒
```

</details>

### half byte with prefetching

<details>

```
Period      │     crc16():                                                                                    ▒
            │     */                                                                                          ▒
            │     uint16_t crc16(const char *buf, int len) {                                                  ▒
            │     int counter;                                                                                ▒
            │     uint16_t crc = 0;                                                                           ▒
            │     for (counter = 0; counter < len; counter++) {                                               ▒
            │     uint8_t tmp = (uint8_t)buf[counter];                                                        ▒
      41188 │       cmp     $0x1,%edx                                                                         ▒
            │     ↓ jle     128                                                                               ▒
      12012 │       lea     -0x2(%rdx),%eax                                                                   ▒
            │       add     %rdi,%rcx                                                                         ▒
      33207 │       lea     crc16_tbl,%rdx                                                                    ◆
      17390 │       lea     0x2(%rdi,%r8,1),%rdi                                                              ▒
       9692 │       add     %rax,%rdi                                                                         ▒
            │     for (counter = 0; counter < len; counter++) {                                               ▒
            │       xor     %eax,%eax                                                                         ▒
            │       xchg    %ax,%ax                                                                           ▒
            │     crc = crc16_base(crc, tmp);                                                                 ▒
            │     }                                                                                           ▒
      27587 │ 90:   movzbl  (%rcx),%esi                                                                       ▒
            │     uint8_t tmp = (uint8_t)buf[counter];                                                        ▒
     114663 │       add     $0x1,%rcx                                                                         ▒
            │     crc16_base():                                                                               ▒
            │     crc ^= v << 8;                                                                              ▒
      18385 │       shl     $0x8,%esi                                                                         ▒
      44236 │       xor     %esi,%eax                                                                         ▒
            │     crc = (crc << 4) ^ crc16_tbl[crc >> 12];                                                    ▒
      47317 │       mov     %eax,%esi                                                                         ▒
            │       shl     $0x4,%eax                                                                         ▒
      53990 │       shr     $0xc,%si                                                                          ▒
      20090 │       and     $0xf,%esi                                                                         ▒
      49666 │       xor     (%rdx,%rsi,2),%ax                                                                 ▒
            │     crc = (crc << 4) ^ crc16_tbl[crc >> 12];                                                    ▒
     389652 │       mov     %eax,%esi                                                                         ▒
            │       shl     $0x4,%eax                                                                         ▒
      97173 │       shr     $0xc,%si                                                                          ▒
      13615 │       and     $0xf,%esi                                                                         ▒
      51438 │       xor     (%rdx,%rsi,2),%ax                                                                 ◆
            │     crc16():                                                                                    ▒
            │     uint8_t tmp = (uint8_t)buf[counter];                                                        ▒
     252343 │       cmp     %rcx,%rdi                                                                         ▒
            │     ↑ jne     90                                                                                ▒
```

</details>

## Code Modifications

### original with prefetching

<details>

```diff
diff --git a/src/cluster.c b/src/cluster.c
index b831610..e142428 100644
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -972,6 +972,9 @@ int clusterSlotByCommand(struct serverCommand *cmd, robj **argv, int argc, int *
     initGetKeysResult(&result);
     int numkeys = getKeysFromCommand(cmd, argv, argc, &result);
     int slot = -1;
+
+    valkey_prefetch(crc16tab + (sizeof(crc16tab) / sizeof(crc16tab[0])));
+
     if (numkeys == 0) *read_flags |= READ_FLAGS_NO_KEYS;
     for (int i = 0; i < numkeys; i++) {
         sds key = argv[result.keys[i].pos]->ptr;
diff --git a/src/crc16.c b/src/crc16.c
index 2153177..799aeaf 100644
--- a/src/crc16.c
+++ b/src/crc16.c
@@ -44,7 +44,7 @@
  * Output for "123456789"     : 31C3
  */
 
-static const uint16_t crc16tab[256]= {
+uint16_t crc16tab[256]= {
     0x0000,0x1021,0x2042,0x3063,0x4084,0x50a5,0x60c6,0x70e7,
     0x8108,0x9129,0xa14a,0xb16b,0xc18c,0xd1ad,0xe1ce,0xf1ef,
     0x1231,0x0210,0x3273,0x2252,0x52b5,0x4294,0x72f7,0x62d6,
diff --git a/src/server.h b/src/server.h
index 6056101..1d0195a 100644
--- a/src/server.h
+++ b/src/server.h
@@ -3656,6 +3656,7 @@ int bzmpopGetKeys(struct serverCommand *cmd, robj **argv, int argc, getKeysResul
 int setGetKeys(struct serverCommand *cmd, robj **argv, int argc, getKeysResult *result);
 int bitfieldGetKeys(struct serverCommand *cmd, robj **argv, int argc, getKeysResult *result);
 
+extern uint16_t crc16tab[256];
 unsigned short crc16(const char *buf, int len);
 
 /* Sentinel */

```

</details>

### original with prefetching ASCII

<details>

```diff
diff --git a/src/cluster.c b/src/cluster.c
index b831610..a7ac7d3 100644
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -972,6 +972,9 @@ int clusterSlotByCommand(struct serverCommand *cmd, robj **argv, int argc, int *
     initGetKeysResult(&result);
     int numkeys = getKeysFromCommand(cmd, argv, argc, &result);
     int slot = -1;
+
+    valkey_prefetch(crc16tab + 128);
+
     if (numkeys == 0) *read_flags |= READ_FLAGS_NO_KEYS;
     for (int i = 0; i < numkeys; i++) {
         sds key = argv[result.keys[i].pos]->ptr;
diff --git a/src/crc16.c b/src/crc16.c
index 2153177..799aeaf 100644
--- a/src/crc16.c
+++ b/src/crc16.c
@@ -44,7 +44,7 @@
  * Output for "123456789"     : 31C3
  */
 
-static const uint16_t crc16tab[256]= {
+uint16_t crc16tab[256]= {
     0x0000,0x1021,0x2042,0x3063,0x4084,0x50a5,0x60c6,0x70e7,
     0x8108,0x9129,0xa14a,0xb16b,0xc18c,0xd1ad,0xe1ce,0xf1ef,
     0x1231,0x0210,0x3273,0x2252,0x52b5,0x4294,0x72f7,0x62d6,
diff --git a/src/server.h b/src/server.h
index 6056101..1d0195a 100644
--- a/src/server.h
+++ b/src/server.h
@@ -3656,6 +3656,7 @@ int bzmpopGetKeys(struct serverCommand *cmd, robj **argv, int argc, getKeysResul
 int setGetKeys(struct serverCommand *cmd, robj **argv, int argc, getKeysResult *result);
 int bitfieldGetKeys(struct serverCommand *cmd, robj **argv, int argc, getKeysResult *result);
 
+extern uint16_t crc16tab[256];
 unsigned short crc16(const char *buf, int len);
 
 /* Sentinel */

```

</details>

### half byte with prefetching

<details>

```diff
diff --cc src/crc16.c
index c5ff44e,799aeaf..0000000
--- a/src/crc16.c
+++ b/src/crc16.c
@@@ -1,46 -1,34 +1,46 @@@
  #include "server.h"
  
  /*
 - * Copyright 2001-2010 Georges Menie (www.menie.org)
 - * Copyright 2010-2012 Redis Ltd. (adapted to Redis coding style)
 + * Copyright Valkey Contributors.
   * All rights reserved.
 - *
 - * Redistribution and use in source and binary forms, with or without
 - * modification, are permitted provided that the following conditions are met:
 - *
 - *     * Redistributions of source code must retain the above copyright
 - *       notice, this list of conditions and the following disclaimer.
 - *     * Redistributions in binary form must reproduce the above copyright
 - *       notice, this list of conditions and the following disclaimer in the
 - *       documentation and/or other materials provided with the distribution.
 - *     * Neither the name of the University of California, Berkeley nor the
 - *       names of its contributors may be used to endorse or promote products
 - *       derived from this software without specific prior written permission.
 - *
 - * THIS SOFTWARE IS PROVIDED BY THE REGENTS AND CONTRIBUTORS ``AS IS'' AND ANY
 - * EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
 - * WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
 - * DISCLAIMED. IN NO EVENT SHALL THE REGENTS AND CONTRIBUTORS BE LIABLE FOR ANY
 - * DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES
 - * (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES;
 - * LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND
 - * ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
 - * (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS
 - * SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 + * SPDX-License-Identifier: BSD 3-Clause
   */
  
 +
 +/* Adapted from: https://www.eevblog.com/forum/programming/crc32-half-bytenibble-lookup-table-algorithm/
 + * The lookup table just needs to store the entries from 1st to 16th
 + * of the standard lookup table.
 + */
- static const uint16_t crc16_tbl[16] = {
++uint16_t crc16_tbl[16] = {
 +    0x0000,
 +    0x1021,
 +    0x2042,
 +    0x3063,
 +    0x4084,
 +    0x50a5,
 +    0x60c6,
 +    0x70e7,
 +    0x8108,
 +    0x9129,
 +    0xa14a,
 +    0xb16b,
 +    0xc18c,
 +    0xd1ad,
 +    0xe1ce,
 +    0xf1ef,
 +};
 +
 +/* Adapted from: https://create.stephan-brumme.com/crc32/
 + * Apply half-byte lookup table algorithm for the balance
 + * between performance and the size of lookup table.
 + */
 +static inline uint16_t crc16_base(uint16_t crc, uint8_t v) {
 +    crc ^= v << 8;
 +    crc = (crc << 4) ^ crc16_tbl[crc >> 12];
 +    crc = (crc << 4) ^ crc16_tbl[crc >> 12];
 +    return crc;
 +}
 +
  /* CRC16 implementation according to CCITT standards.
   *
   * Note by @antirez: this is actually the XMODEM CRC 16 algorithm, using the
@@@ -55,6 -43,42 +55,7 @@@
   * Xor constant to output CRC : 0000
   * Output for "123456789"     : 31C3
   */
+ 
 -uint16_t crc16tab[256]= {
 -    0x0000,0x1021,0x2042,0x3063,0x4084,0x50a5,0x60c6,0x70e7,
 -    0x8108,0x9129,0xa14a,0xb16b,0xc18c,0xd1ad,0xe1ce,0xf1ef,
 -    0x1231,0x0210,0x3273,0x2252,0x52b5,0x4294,0x72f7,0x62d6,
 -    0x9339,0x8318,0xb37b,0xa35a,0xd3bd,0xc39c,0xf3ff,0xe3de,
 -    0x2462,0x3443,0x0420,0x1401,0x64e6,0x74c7,0x44a4,0x5485,
 -    0xa56a,0xb54b,0x8528,0x9509,0xe5ee,0xf5cf,0xc5ac,0xd58d,
 -    0x3653,0x2672,0x1611,0x0630,0x76d7,0x66f6,0x5695,0x46b4,
 -    0xb75b,0xa77a,0x9719,0x8738,0xf7df,0xe7fe,0xd79d,0xc7bc,
 -    0x48c4,0x58e5,0x6886,0x78a7,0x0840,0x1861,0x2802,0x3823,
 -    0xc9cc,0xd9ed,0xe98e,0xf9af,0x8948,0x9969,0xa90a,0xb92b,
 -    0x5af5,0x4ad4,0x7ab7,0x6a96,0x1a71,0x0a50,0x3a33,0x2a12,
 -    0xdbfd,0xcbdc,0xfbbf,0xeb9e,0x9b79,0x8b58,0xbb3b,0xab1a,
 -    0x6ca6,0x7c87,0x4ce4,0x5cc5,0x2c22,0x3c03,0x0c60,0x1c41,
 -    0xedae,0xfd8f,0xcdec,0xddcd,0xad2a,0xbd0b,0x8d68,0x9d49,
 -    0x7e97,0x6eb6,0x5ed5,0x4ef4,0x3e13,0x2e32,0x1e51,0x0e70,
 -    0xff9f,0xefbe,0xdfdd,0xcffc,0xbf1b,0xaf3a,0x9f59,0x8f78,
 -    0x9188,0x81a9,0xb1ca,0xa1eb,0xd10c,0xc12d,0xf14e,0xe16f,
 -    0x1080,0x00a1,0x30c2,0x20e3,0x5004,0x4025,0x7046,0x6067,
 -    0x83b9,0x9398,0xa3fb,0xb3da,0xc33d,0xd31c,0xe37f,0xf35e,
 -    0x02b1,0x1290,0x22f3,0x32d2,0x4235,0x5214,0x6277,0x7256,
 -    0xb5ea,0xa5cb,0x95a8,0x8589,0xf56e,0xe54f,0xd52c,0xc50d,
 -    0x34e2,0x24c3,0x14a0,0x0481,0x7466,0x6447,0x5424,0x4405,
 -    0xa7db,0xb7fa,0x8799,0x97b8,0xe75f,0xf77e,0xc71d,0xd73c,
 -    0x26d3,0x36f2,0x0691,0x16b0,0x6657,0x7676,0x4615,0x5634,
 -    0xd94c,0xc96d,0xf90e,0xe92f,0x99c8,0x89e9,0xb98a,0xa9ab,
 -    0x5844,0x4865,0x7806,0x6827,0x18c0,0x08e1,0x3882,0x28a3,
 -    0xcb7d,0xdb5c,0xeb3f,0xfb1e,0x8bf9,0x9bd8,0xabbb,0xbb9a,
 -    0x4a75,0x5a54,0x6a37,0x7a16,0x0af1,0x1ad0,0x2ab3,0x3a92,
 -    0xfd2e,0xed0f,0xdd6c,0xcd4d,0xbdaa,0xad8b,0x9de8,0x8dc9,
 -    0x7c26,0x6c07,0x5c64,0x4c45,0x3ca2,0x2c83,0x1ce0,0x0cc1,
 -    0xef1f,0xff3e,0xcf5d,0xdf7c,0xaf9b,0xbfba,0x8fd9,0x9ff8,
 -    0x6e17,0x7e36,0x4e55,0x5e74,0x2e93,0x3eb2,0x0ed1,0x1ef0
 -};
 -
  uint16_t crc16(const char *buf, int len) {
      int counter;
      uint16_t crc = 0;
diff --git a/src/cluster.c b/src/cluster.c
index e142428..9a83bd5 100644
--- a/src/cluster.c
+++ b/src/cluster.c
@@ -973,7 +973,7 @@ int clusterSlotByCommand(struct serverCommand *cmd, robj **argv, int argc, int *
     int numkeys = getKeysFromCommand(cmd, argv, argc, &result);
     int slot = -1;
 
-    valkey_prefetch(crc16tab + (sizeof(crc16tab) / sizeof(crc16tab[0])));
+    valkey_prefetch(crc16_tbl + (sizeof(crc16_tbl) / sizeof(crc16_tbl[0])));
 
     if (numkeys == 0) *read_flags |= READ_FLAGS_NO_KEYS;
     for (int i = 0; i < numkeys; i++) {
diff --git a/src/server.h b/src/server.h
index 1d0195a..a8361a2 100644
--- a/src/server.h
+++ b/src/server.h
@@ -3656,7 +3656,7 @@ int bzmpopGetKeys(struct serverCommand *cmd, robj **argv, int argc, getKeysResul
 int setGetKeys(struct serverCommand *cmd, robj **argv, int argc, getKeysResult *result);
 int bitfieldGetKeys(struct serverCommand *cmd, robj **argv, int argc, getKeysResult *result);
 
-extern uint16_t crc16tab[256];
+extern uint16_t crc16_tbl[16];
 unsigned short crc16(const char *buf, int len);
 
 /* Sentinel */
```

</details>

Cuda-Chen:
Hi @roshkhatri ,

I would like to request your help of triggering the `run-benchmark` CI, thanks!

roshkhatri:
Hey, I am waiting for this PR to be merged: https://github.com/valkey-io/valkey/pull/2625

Cuda-Chen:
Hi @zuiderkwast ,

After the testing, the half-byte implementation produces inferior benchmark (about 5% loss on both RPS and latency).

Should I just close this PR, or re-use this PR for other tasks mentioned in #2031 ?

zuiderkwast:
Hi @Cuda-Chen! Yes, unfortunately it seems slower. Close and try another idea in a new PR seems better to me.

Do you think another idea will be faster or is the lut already the best solution?

Cuda-Chen:
Hi @zuiderkwast ,

> Close and try another idea in a new PR seems better to me.

Thanks for the hint! I will close this PR later.

> Do you think another idea will be faster or is the lut already the best solution?

I recall that you mention the key names are typically only 20 bytes long, so we can set the length to 32 bytes (as it is a power of 2).

Currently, I have come up with these ideas:

- For another approaches:
  - We can try carry-less multiplication or breaking input message into segments then merge (also needs carry-less multiplication). Or even we can apply both methods.
  - For carry-less multiplication, it will be applied on calculating the CRC16 on each byte of input message. What's more, this method does not need to store LUT.
  - For breaking input message into segments then merge:
    - The breaking input message into segments part will utilize ILP (Instruction Level Parallelism) for calculating the CRC16 of each segment.
    - The merging part will merge the CRC16 of each segment to produce the result. However, the merging requires carry-less multiplication.
  - For carry-less multiplication, the hardware support on both X86 and Arm "claim" it will operate in O(1) time.
- For LUT solution:
  - We can see that the original implementation performs better compared to the proposed half-byte LUT on RPS and latency metrics.
  - As the original implementation operates on every single byte, there might be a chance that operating on multiple bytes (for example, 8 bytes each time) yields better performance, and I think for many user scenarios, using more memory space for performance gain will be acceptable.

PR #754, opened 2024-07-07
Ensure only primary sender drives slot ownership updates

PingXie:
Fixes a regression introduced in PR #445, which allowed a message from a replica
to update the slot ownership of its primary. The regression results in a
`replicaof` cycle, causing server crashes due to the cycle detection assert. The
fix restores the previous behavior where only primary senders can trigger
`clusterUpdateSlotsConfigWith`.

Additional changes:

* Handling of primaries without slots is obsoleted by new handling of when a
  sender that was a replica announces that it is now a primary.
* Replication loop detection code is unchanged but shifted downwards.
* Some variables are renamed for better readability and some are introduced to
  avoid repeated memcmp() calls.

Fixes #753.

Issue #753, opened 2024-07-07
Regression from PR #445 Incorrectly Allows Slot Ownership Updates via Replica


PingXie:
I took a look too and realized it’s a regression introduced by my slot migration PR #445. This change started allowing a replica to report its primary’s slot states and trigger clusterUdpateSlotsConfigWith.

![image](https://github.com/valkey-io/valkey/assets/11568491/a66e4c44-16c6-473c-adf0-648782254ce9)

[PR #445 - Slot Migration Changes](https://github.com/valkey-io/valkey/pull/445/files#diff-2515500619600c5a1e7a8d9aaa8761a6071314cd226c3b7ee4d0e054691f0a8eL3094-R3302).

Here's what I think happens in these test failures involving a 3-node shard:

[T1] - Node `A`, `B`, and `C` are in the same shard with `A` as the primary.
[T2] - Node `A` loses its primaryship to `B` via a graceful/manual failover.
[T3] - After winning the election, `B` broadcasts the news to every node in the cluster, including `C`.
[T4] - `C` receives `B`'s latest `PING` message and correctly registers `B` as its new primary.
[T5] - `C` then sends a new `PING` message to `A`, claiming `B` is its primary with all the slots.
[T6] - `A` still hasn't received `B`'s broadcast message from [T3], and `C`'s `PING` message from [T4] arrives at `A`.
And this is where things go wrong—a `replicaof` cycle is created.

At this point, `A` still thinks it’s the primary of the shard, and `B -> replicaof == A`. Since `C` is still a replica (as before), the [role change handling logic](https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L3051) doesn’t apply. So, `A` enters `clusterUdpateSlotsConfigWith` using `C`’s slot information (which is up to date with `B`’s). More importantly, `B` is the sender, and `A` assumes `B -> replicaof == A`. The slot ownership update logic correctly gives the ownership of the slots to `B`. Because `A` loses all its slots to another node in the same shard with a higher config epoch, this demotes `A` to a replica of the winner, `B`. And we set `A -> replicaof = B`, completing the `replicaof` cycle.

_Originally posted by @PingXie in https://github.com/valkey-io/valkey/issues/573#issuecomment-2212224376_

hpatro:
Thanks for the nice explanation @PingXie! 

[...]

PR #445, opened 2024-05-07
Slot migration improvement

PingXie:
See #245 for the review discussion.

## Overview
This PR significantly enhances the reliability and automation of
the Valkey cluster re-sharding process, specifically during slot
migrations in the face of primary failures. These updates address
critical failure issues that previously required extensive manual
intervention and could lead to data loss or inconsistent cluster states.

### Automatic Failover Support in Empty Shards 
The cluster now supports automatic failover in shards that do not
own any slots, which is common during scaling operations. This
improvement ensures high availability and resilience from the
outset of shard expansion.

### Replication of Slot Migration States
All `CLUSTER SETSLOT` commands are now initially executed on replica
nodes before the primary. This ensures that the slot migration state
is consistent within the shard, preventing state loss in the event of
primary failure. A new timeout parameter has been introduced, allowing
users to specify the duration in milliseconds to wait for replication
to complete, with a default set at 2 seconds.
```
CLUSTER SETSLOT slot { IMPORTING node-id | MIGRATING node-id | NODE node-id | STABLE } [ TIMEOUT timeout ]
```

### Recovery of Logical Migration Links
The update automatically repairs the logical links between source
and target nodes during failovers. This ensures that requests are
correctly redirected to the new primary in the target shard after
a primary failure, maintaining cluster integrity.

### Enhanced Support for New Replicas
New replicas added to shards involved in slot migrations will now
automatically inherit the slot's migration state as part of their
initialization. This ensures that new replicas are immediately
consistent with the rest of the shard.

### Improved Logging for Slot Migrations
Additional logging has been implemented to provide operators with
clearer insights into the slot migration processes and automatic
recovery actions, aiding in monitoring and troubleshooting.

### cluster-allow-replica-migration

When `cluster-allow-replica-migration` is disabled, primary nodes that
lose their last slot to another shard will no longer automatically become
replicas of the receiving shard. Instead, they will remain in their own
shards, which will now be empty, having no slots assigned to them.

Fixes #21.

PR #2301, opened 2025-07-03
Fix replica (the old primary) claims to sitll have slots after manual failover

enjoy-binbin:
## update
Please also pick #2370 and #2431 and #2441 and #2477 you want to pick this one.

================

After a failover occurs, when the PING-PONG of the replica, that is,
the old primary, reaches other shard nodes faster, the corresponding
code path will be reached. We will adjust the sender to be a replica
and sender_claimed_primary to be the primary node, but in myself view,
slots still belong to the sender, which is a replica.

This actually restores part of the code in 28976a9003c6dd5cdd7225c5bc90743b4fcde13c.
It was lost in the changes to #445 and #754.

These 3 changes are about avoiding the replicaof cycle but ended up creating
a "cycle" of assumptions themselves. when #445 removed this logic, it was
assumed that a replica could also drive clusterUpdateSlotsConfigWith
but this was actually a regression that resulted in a replicaof cycle (#753).
#754 fixed the replicaof cycle by disallowing a replica to drive
clusterUpdateSlotsConfigWith but missed restoring the original logic.

Added a new `DEBUG DISABLE-CLUSTER-RECONNECTION <0|1>` this will prevent
cluster nodes from reconnecting so that the issue can be reproduce in test.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -71,6 +71,7 @@ list *clusterGetNodesInMyShard(clusterNode *node);
 int clusterNodeAddReplica(clusterNode *primary, clusterNode *replica);
 int clusterAddSlot(clusterNode *n, int slot);
 int clusterDelSlot(int slot);
+void clusterMoveNodeSlots(clusterNode *from_node, clusterNode *to_node);
 int clusterDelNodeSlots(clusterNode *node);
 void clusterNodeSetSlotBit(clusterNode *n, int slot);
 static void clusterSetPrimary(clusterNode *n, int closeSlots, int full_sync_required);
@@ -3638,6 +3639,10 @@ int clusterProcessPacket(clusterLink *link) {
                              * this may cause a primary-replica chain issue. */
                             return 1;
                         } else if (nodeIsReplica(sender_claimed_primary)) {
+                            /* A failover occurred in the shard where `sender` belongs to and `sender` is
+                             * no longer a primary. Update slot assignment to `sender_claimed_config_epoch`,
+                             * which is the new primary in the shard */
+                            clusterMoveNodeSlots(sender, sender_claimed_primary);
                             /* `primary` is still a `replica` in this observer node's view;
                              * update its role and configEpoch */
                             clusterSetNodeAsPrimary(sender_claimed_primary);
@@ -5742,6 +5747,18 @@ int clusterDelSlot(int slot) {
     bitmapClearBit(server.cluster->owner_not_claiming_slot, slot);
     clusterSlotStatReset(slot);
     return C_OK;
+}
+
+/* Transfer slots from `from_node` to `to_node`.
+ * Iterates over all cluster slots, transferring each slot covered
+ * by `from_node` to `to_node`. */
+void clusterMoveNodeSlots(clusterNode *from_node, clusterNode *to_node) {
+    for (int j = 0; j < CLUSTER_SLOTS; j++) {
+        if (clusterNodeCoversSlot(from_node, j)) {
+            clusterDelSlot(j);
+            clusterAddSlot(to_node, j);
+        }
+    }
 }
 
 /* Delete all the slots associated with the specified node.
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3639,6 +3639,7 @@ int clusterProcessPacket(clusterLink *link) {
                              * this may cause a primary-replica chain issue. */
                             return 1;
                         } else if (nodeIsReplica(sender_claimed_primary)) {
+                            serverAssert(sender_claimed_primary->replicaof == sender);
                             /* A failover occurred in the shard where `sender` belongs to and `sender` is
                              * no longer a primary. Update slot assignment to `sender_claimed_config_epoch`,
                              * which is the new primary in the shard */
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -686,7 +686,6 @@ int clusterLoadConfig(char *filename) {
             } else if (!strcasecmp(s, "master") || !strcasecmp(s, "primary")) {
                 n->flags |= CLUSTER_NODE_PRIMARY;
             } else if (!strcasecmp(s, "slave") || !strcasecmp(s, "replica")) {
-                serverAssert(n->numslots == 0);
                 n->flags |= CLUSTER_NODE_REPLICA;
             } else if (!strcasecmp(s, "fail?")) {
                 n->flags |= CLUSTER_NODE_PFAIL;
@@ -3655,6 +3654,7 @@ int clusterProcessPacket(clusterLink *link) {
                                       sender->shard_id, sender->name, sender->human_nodename, slots,
                                       sender_claimed_primary->name, sender_claimed_primary->human_nodename,
                                       (unsigned long long)sender_claimed_primary->configEpoch);
+                            serverAssert(sender->numslots == 0);
                         }
                     } else {
                         /* `sender` was moved to another shard and has become a replica, remove its slot assignment */
@@ -3667,9 +3667,9 @@ int clusterProcessPacket(clusterLink *link) {
                             serverLog(LL_NOTICE, "Node %.40s (%s) is now part of shard %.40s", sender->name,
                                       sender->human_nodename, sender_claimed_primary->shard_id);
                         }
+                        serverAssert(sender->numslots == 0);
                     }
 
-                    serverAssert(sender->numslots == 0);
                     sender->flags &= ~(CLUSTER_NODE_PRIMARY | CLUSTER_NODE_MIGRATE_TO);
                     sender->flags |= CLUSTER_NODE_REPLICA;
 
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -71,8 +71,8 @@ list *clusterGetNodesInMyShard(clusterNode *node);
 int clusterNodeAddReplica(clusterNode *primary, clusterNode *replica);
 int clusterAddSlot(clusterNode *n, int slot);
 int clusterDelSlot(int slot);
-void clusterMoveNodeSlots(clusterNode *from_node, clusterNode *to_node);
 int clusterDelNodeSlots(clusterNode *node);
+int clusterMoveNodeSlots(clusterNode *from_node, clusterNode *to_node);
 void clusterNodeSetSlotBit(clusterNode *n, int slot);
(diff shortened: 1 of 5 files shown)


enjoy-binbin:
In the test, there will be a short time window that the replica claims to still have slots.
```
start_cluster 3 1 {tags {external:skip cluster} overrides {cluster-replica-validity-factor 0}} {
    set R0_nodeid [R 0 cluster myid]
    set R1_nodeid [R 1 cluster myid]
    set R2_nodeid [R 2 cluster myid]
    set R3_nodeid [R 3 cluster myid]


    R 3 multi
    R 3 debug clusterlink kill all $R1_nodeid
    R 3 debug clusterlink kill all $R2_nodeid
    R 3 cluster failover takeover
    R 3 exec

    puts [R 1 cluster nodes]
    after 10
    puts [R 1 cluster nodes]
    after 10
    puts [R 1 cluster nodes]
    after 10
    puts [R 1 cluster nodes]
    after 10
    puts [R 1 cluster nodes]
    ...
}
```

We can see 84088c54fcae0666b552229b806e4f96eaba2a0b become a replica but own the slots
```
84088c54fcae0666b552229b806e4f96eaba2a0b 127.0.0.1:21114@31114 master - 0 1752039102266 1 connected 0-5461
f58610cd23c75be51a27c717e9f4997830bcfadf 127.0.0.1:21111@31111 slave 84088c54fcae0666b552229b806e4f96eaba2a0b 0 1752039102266 1 disconnected
bc07472c09d9e93fc7feb98a846acda76f97e67f 127.0.0.1:21112@31112 master - 0 1752039102266 3 connected 10923-16383
0d74d311b9432a64eeb1dfaefae5a25acf8a2300 127.0.0.1:21113@31113 myself,master - 0 0 2 connected 5462-10922

84088c54fcae0666b552229b806e4f96eaba2a0b 127.0.0.1:21114@31114 slave f58610cd23c75be51a27c717e9f4997830bcfadf 0 1752039102266 4 connected 0-5461
f58610cd23c75be51a27c717e9f4997830bcfadf 127.0.0.1:21111@31111 master - 0 1752039102266 4 disconnected
bc07472c09d9e93fc7feb98a846acda76f97e67f 127.0.0.1:21112@31112 master - 0 1752039102266 3 connected 10923-16383
0d74d311b9432a64eeb1dfaefae5a25acf8a2300 127.0.0.1:21113@31113 myself,master - 0 0 2 connected 5462-10922

{0 5461 {127.0.0.1 21114 84088c54fcae0666b552229b806e4f96eaba2a0b {}}} {5462 10922 {127.0.0.1 21113 0d74d311b9432a64eeb1dfaefae5a25acf8a2300 {}}} {10923 16383 {127.0.0.1 21112 bc07472c09d9e93fc7feb98a846acda76f97e67f {}}}
```


I will try to think of a way to add the test, there are currently no stable steps to reproduce it.

zuiderkwast (approved):
LGTM

Is this the failures we saw in Daily a few days ago?

enjoy-binbin:
> Is this the failures we saw in Daily a few days ago?

I guess not. I found this from another code review.

enjoy-binbin:
btw, there is a saying that, we should not update sender->replicaof info based on the sender infomartion, like in here, we updated the sender_claimed_primary flag, so i guess it’s OK if we update the slots info.

But I also want to hear your thoughts, like, should sender->replicaof information be updated based on sender information?

zuiderkwast:
> btw, there is a saying that, we should not update sender->replicaof info based on the sender infomartion

I don'tknow this saying. Is it written anywhere?

I think information about sender directly from the sender can be trusted. It is direct information, so it should be correct. Why not?

enjoy-binbin:
> I think information about sender directly from the sender can be trusted

ohh, maybe i gave the wrond understanding, i mean, update the other node (sender->replicaof) infomation based on the sender infomartion.

> I don'tknow this saying. Is it written anywhere?

I guess not, i did not see this, a person from my team said this, he mentioned something that we usually don't update the info of another node based on the info of one node. Like in here, we update the info of sender->replicaof (flags / slots info) based on the info of sender.

PingXie:
> In the test, there will be a short time window that the replica claims to still have slots.

I think this should be a transitional state and eventually the replica drops these slots. that being said, this pr makes sense to me and is the right thing to do, since the role change and the slot ownership change should've been an atomic operation.

> This actually restores part of the code in https://github.com/valkey-io/valkey/commit/28976a9003c6dd5cdd7225c5bc90743b4fcde13c.
It was lost in the changes to https://github.com/valkey-io/valkey/pull/445 and https://github.com/valkey-io/valkey/pull/754.

yeah these 3 changes are about avoiding the `replicaof` cycle but ended up creating a "cycle" of assumptions themselves. when #445 removed this logic, it was assumed that a replica could also drive `clusterUpdateSlotsConfigWith` but this was actually a regression that resulted in a `replicaof` cycle (#753). #754 fixed the `replicaof` cycle by disallowing a replica to drive `clusterUpdateSlotsConfigWith` but missed restoring the original logic.

PingXie (approved):
LGTM - thanks @enjoy-binbin. great investigation.

enjoy-binbin:
@PingXie @zuiderkwast Please check the new changes, i added some assert, a new debug command, and the test (Tests can fail reliably before the fix).

src/cluster_legacy.c
```diff
@@ -686,6 +686,7 @@ int clusterLoadConfig(char *filename) {
             } else if (!strcasecmp(s, "master") || !strcasecmp(s, "primary")) {
                 n->flags |= CLUSTER_NODE_PRIMARY;
             } else if (!strcasecmp(s, "slave") || !strcasecmp(s, "replica")) {
+                serverAssert(n->numslots == 0);
```
PingXie: a crash seems to be too harsh. maybe consider printing out a warning and auto-correct `numslots` to 0?

tests/unit/cluster/manual-failover.tcl
```diff
@@ -395,3 +395,82 @@ start_cluster 3 1 {tags {external:skip cluster} overrides {cluster-ping-interval
...
+        set R0_shardid [R 0 cluster myshardid]
+        set R3_shardid [R 3 cluster myshardid]
+        assert_equal $R0_shardid $R3_shardid
+
+        # Ensure that related nodes do not reconnect.
+        R 1 debug disable-cluster-reconnection 1
+        R 2 debug disable-cluster-reconnection 1
+        R 3 debug disable-cluster-reconnection 1
+
+        # After killing the cluster link, ensure that R1 and R3 do not receive
```
PingXie:
```suggestion
        # After killing the cluster link, ensure that R1 and R2 do not receive
```

PingXie (approved):
great test!

src/cluster_legacy.c
enjoy-binbin: or we can goto fmterr? I treat this to be something like a corrupted nodes.conf file so maybe we can just goto fmterr?

Or we can leave it at it is, some old version may generate this nodes.conf, and we can correct it when the cluster become stable.

enjoy-binbin: i decided to remove it for now.

zuiderkwast:
Merged, very nice. Backport or not?

PingXie:
> Merged, very nice. Backport or not?

This is a safe change so I vote for backport

enjoy-binbin:
Added to 8.0/8.1 project for backport.

src/cluster_legacy.c
```diff
@@ -3638,16 +3639,22 @@ int clusterProcessPacket(clusterLink *link) {
                              * this may cause a primary-replica chain issue. */
                             return 1;
                         } else if (nodeIsReplica(sender_claimed_primary)) {
+                            serverAssert(sender_claimed_primary->replicaof == sender);
```
madolson, line 3642: I don't see why this assert must be true at this point in time in the code. It also is crashing here: https://github.com/valkey-io/valkey/actions/runs/16720515261/job/47323708440?pr=2421#step:5:6544.

madolson:
> This is a safe change so I vote for backport

@PingXie, unless we are extremely confident, I wouldn't consider backporting novel severAsserts to be safe. This can introduce new crashes that have caused CVEs in the past.

src/cluster_legacy.c
enjoy-binbin: ok, i am taking a look

enjoy-binbin: The CI did not print the server logs, so i'm still guessing what the scenario might be.

i add this assert because, `sender_claimed_primary` and `sender` are in the same shard, and `sender` is the old primary, and `sender_claimed_primary` is the old replica, so `sender` should be the primary of `sender_claimed_primary`

enjoy-binbin: i guess the scenario could be:
- sender is an empty primary
- sender_claimed_primary doing a failover in its shard
- sender replicate to sender_claimed_primary

enjoy-binbin: Do you have any other possible scenarios in mind? Or should we just remove this assert?

enjoy-binbin: i think the problem could be #2279, like due to some reasons, there are two primaries in the same shard

enjoy-binbin: ok, i combined the tests in #2279 (two primaries in the same shard) and #2301 (this one) and successfully reproduced this assert.

so there will be a case that can reproduce this assert:
1. pre condition: two primaries in the same shard (from #2279)
2. replica doing the failover
3. the empty primary doing a replicate
4. they are in the same shard and meet every conditions in the code and trigger the assert

enjoy-binbin: i will submit the PR tomorrow since the test is quite ugly (quite temporary) right now and i need to refine it.

zuiderkwast: Great! Maybe it's safer to just remove the assert, or change it to an `if` condition?

```c
                        } else if (nodeIsReplica(sender_claimed_primary) &&
                                   sender_claimed_primary->replicaof == sender)) {
```

zuiderkwast: > Great! Maybe it's safer to just remove the assert, or change it to an `if` condition?
> 
> ```c
>                         } else if (nodeIsReplica(sender_claimed_primary) &&
>                                    sender_claimed_primary->replicaof == sender)) {
> ```

madolson:
I don't think we should backport these. I removed from the 8.0 and 8.1 projects. Let me know if anyone disagrees.

PR #2306, opened 2025-07-04
Unexpected variable overriding from .make-settings file

yzc-yzc:
```
# Override default settings if possible
-include .make-settings
```
The above code will overwrite the original variables, such as `STD`, etc. This results in `FINAL_CFLAGS`, `FINAL_LDFLAGS` not being updated correctly.

```
# Clean everything, persist settings and build dependencies if anything changed
ifneq ($(strip $(PREV_FINAL_CFLAGS)), $(strip $(FINAL_CFLAGS)))
.make-prerequisites: persist-settings
endif

ifneq ($(strip $(PREV_FINAL_LDFLAGS)), $(strip $(FINAL_LDFLAGS)))
.make-prerequisites: persist-settings
endif
```
So the above lines of Makefile don't work properly.

The bug was introduced from https://github.com/valkey-io/valkey/commit/7e7b69fee1c87b4802394a8637bb1dc0b474614c. I'm a little shocked that it's been so long. Am I missing something?

### DIFF
--- a/src/Makefile
+++ b/src/Makefile
@@ -466,17 +466,6 @@ all-with-unit-tests: all $(ENGINE_UNIT_TESTS)
 .PHONY: all
 
 persist-settings: distclean
-	echo STD=$(STD) >> .make-settings
-	echo WARN=$(WARN) >> .make-settings
-	echo OPT=$(OPT) >> .make-settings
-	echo MALLOC=$(MALLOC) >> .make-settings
-	echo BUILD_TLS=$(BUILD_TLS) >> .make-settings
-	echo BUILD_RDMA=$(BUILD_RDMA) >> .make-settings
-	echo USE_SYSTEMD=$(USE_SYSTEMD) >> .make-settings
-	echo CFLAGS=$(CFLAGS) >> .make-settings
-	echo LDFLAGS=$(LDFLAGS) >> .make-settings
-	echo SERVER_CFLAGS=$(SERVER_CFLAGS) >> .make-settings
-	echo SERVER_LDFLAGS=$(SERVER_LDFLAGS) >> .make-settings
 	echo PREV_FINAL_CFLAGS=$(FINAL_CFLAGS) >> .make-settings
 	echo PREV_FINAL_LDFLAGS=$(FINAL_LDFLAGS) >> .make-settings
 	-(cd ../deps && $(MAKE) $(DEPENDENCY_TARGETS))
--- a/src/Makefile
+++ b/src/Makefile
@@ -12,6 +12,33 @@
 # Dependencies are stored in the Makefile.dep file. To rebuild this file
 # Just use 'make dep', but this is only needed by developers.
 
+# Override default settings if undefined
+-include .make-settings
+
+define save_var_if_defined
+$(if $($(1)), \
+    echo "ifndef $(1)" >> .tmp.make-settings; \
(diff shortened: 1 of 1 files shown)


zuiderkwast:
The idea of persisting variables to a file is that if you build with some variables like `make SANITIZER=address BUILD_TLS=yes` or something, the next time you just type `make`, it should use the previous variables. This appears to break with your changes.

Maybe we can reorder lines to solve the problem and still keep the intended behavior.

It's not exactly clear to me what behavior is wanted here. We should probably define the expected behavior clearly first.

yzc-yzc:
> The idea of persisting variables to a file is that if you build with some variables like `make SANITIZER=address BUILD_TLS=yes` or something, the next time you just type `make`, it should use the previous variables. This appears to break with your changes.
> 
> Maybe we can reorder lines to solve the problem and still keep the intended behavior.
> 
> It's not exactly clear to me what behavior is wanted here. We should probably define the expected behavior clearly first.

I used to mistakenly think that **the next time** I type `make`, the default settings will be executed again. A further question is what kind of behavior should there be if **the next time** I type `make SANITIZER=address`. Should `BUILD_TLS=yes` be maintained? 

On the other hand I searched the repo for instructions on how to use makefiles and only found this one.
https://github.com/valkey-io/valkey/blob/unstable/README.md?plain=1#L92-L95
I wonder under the above guidance, if **the next time** I type `make SANITIZER=undefined BUILD_TLS=yes` should work, or do I have to type `make distclean`?

How should I know what the community thinks? I'm worried about disturbing the community.

zuiderkwast:
The main thing to keep is probably a scenario like `make SOMEOPTION=value... ; make test ; sudo make install`. We don't want `make test` and `make install` to forget the variables from the first make. This, and just developers changing some file and typing `make` again without variables is the main thing to keep. Other than this, I think it's OK to change the behavior.

I have an idea:

1. Read cached variables from file.
2. Read user explicit variables and let them override cached variables.
3. Write to cache file.

It will give quite nice behavior. What do you think? 

It will let users add to the cached variables. If first `make SANITIZER= address`,  later `make BUILD_TLS=yes` will result in sanitizer + tls.

However, we might need to rebuild everything if any variable is changed, i.e. implicit distclean.

yzc-yzc:
> The main thing to keep is probably a scenario like `make SOMEOPTION=value... ; make test ; sudo make install`. We don't want `make test` and `make install` to forget the variables from the first make. This, and just developers changing some file and typing `make` again without variables is the main thing to keep. Other than this, I think it's OK to change the behavior.
> 
> I have an idea:
> 
> 1. Read cached variables from file.
> 2. Read user explicit variables and let them override cached variables.
> 3. Write to cache file.
> 
> It will give quite nice behavior. What do you think?
> 
> It will let users add to the cached variables. If first `make SANITIZER= address`, later `make BUILD_TLS=yes` will result in sanitizer + tls.
> 
> However, we might need to rebuild everything if any variable is changed, i.e. implicit distclean.

SGTM.

So the expected behavior will be as follows.
First step: `make VAR1=VALUE1 VAR2=VALUE2`(without `.make-settings` file)
Second step: `make VAR2=VALUE3`(with .make-settings file) 
The above is equivalent to `make VAR1=VALUE1 VAR2=VALUE3`(without `.make-settings` file)

Issue #2282, opened 2025-06-28
Make sentinel manager large cluster easy


kukey:
We aims use sentinel to manager large cluster easy, and not depend any other components.
Make the sentinel a stable, reliable, faster and simple HA component.

This issue like #2281 , but we are centralized management architecture.

## Expectations:

* proactive failover 2k node (or more) time cost from 10min -> 15s

  ~~~ 
  in our cluster test, 1725 master node and 3450 replica node（1 master/ 2 replica node) proactive failvoer 1725 all master nodes, the faster time cost is 5s.
  ~~~

* optimize the strategy of updating the configuration file of sentinel to flush it to the disk

* send switch-master quickly

* support the psync failover capability support

* optimize the observability

  * the sentinel self info message, there are many invalid info
  * sentinel collect server info, many not used, this may cost more network bandwith and cpu cost

  * sentinel should have a complete observable indicators, like cron time cost, current_epoch, flush config time cost, total_titl and so on
  * optimize lttng support for sentinel

* manage cluster slot info, offload this to sentinel; make sentinel like a shadow node

* optimize sentinel command, now miss some mange command, like add/del/reset node and so on

## High Level Areas:

My friends and colleagues and I have done some work, and we will keep completing this.

### core && performance
- [ ] https://github.com/valkey-io/valkey/pull/1511
- [ ] https://github.com/valkey-io/valkey/pull/2307
- [x] https://github.com/valkey-io/valkey/pull/2308
### bugfix
- [ ] https://github.com/valkey-io/valkey/pull/2255
### new feature
- [ ] https://github.com/valkey-io/valkey/pull/1964
### metric
- [x] https://github.com/valkey-io/valkey/pull/1904

like to hear your voice and suggestions!

zuiderkwast:
I don't know much about Sentinel. We don't use it.

@hwware and @dmitrypol are you interested in this topic?

dmitrypol:
Yes, I am interested

hwware:
This is a cluster case, not a sentinel case. In a cluster,  failover based on master nodes and replica nodes.

PR #2307, opened 2025-07-04
[optimization] Optimization of Sentinel Configuration File Update Strategy

youngmore1024:
Optimize the strategy of updating the configuration file of sentinel to flush it to the disk, ref issue [#2282](https://github.com/valkey-io/valkey/issues/2282).

### DIFF
--- a/deps/libvalkey/src/net.c
+++ b/deps/libvalkey/src/net.c
@@ -395,7 +395,7 @@ int valkeyHasMptcp(void) {
     return 0;
 }
 
-static int valkeyTcpGetProtocol(int is_mptcp_enabled) {
+static int valkeyTcpGetProtocol(VALKEY_UNUSED int is_mptcp_enabled) {
     assert(!is_mptcp_enabled);
     return IPPROTO_TCP;
 }
--- a/src/sentinel.c
+++ b/src/sentinel.c
@@ -96,6 +96,8 @@ static mstime_t sentinel_election_timeout = 10000;
 static mstime_t sentinel_script_max_runtime = 60000; /* 60 seconds max exec time. */
 static mstime_t sentinel_script_retry_delay = 30000; /* 30 seconds between retries. */
 static mstime_t sentinel_default_failover_timeout = 60 * 3 * 1000;
+static int config_dirty = 0;
+static mstime_t flush_config_timestamp = 0;
 
 #define SENTINEL_HELLO_CHANNEL "__sentinel__:hello"
 #define SENTINEL_DEFAULT_REPLICA_PRIORITY 100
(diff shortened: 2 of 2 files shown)


gmbnomis:
I could not find a description that motivates the changes in this PR (#2282 does not describe them). So, could you please elaborate on:
* In which cases does postponing to write the config (by a second) help?
* Some of the config flushes are forced (as before), some are not. Why do you think that the latter actually do not require flushing to disk immediately?

Other comments:
* If the sentinel instance shuts down gracefully with `config_dirty > 0` , I think we need to flush the config. I don't think that is part of the standard shut down procedure, but I may be wrong.
* If we can't write a dirty config, `config_dirty` is not reset (which is good). But doesn't that mean that sentinel will try to save the config again on every cron run until the save succeeds? Do we need to backoff here or to limit the number of times we try?

youngmore1024:
> I could not find a description that motivates the changes in this PR (#2282 does not describe them). So, could you please elaborate on:
> 
> * In which cases does postponing to write the config (by a second) help?
> * Some of the config flushes are forced (as before), some are not. Why do you think that the latter actually do not require flushing to disk immediately?
> 
> Other comments:
> 
> * If the sentinel instance shuts down gracefully with `config_dirty > 0` , I think we need to flush the config. I don't think that is part of the standard shut down procedure, but I may be wrong.
> * If we can't write a dirty config, `config_dirty` is not reset (which is good). But doesn't that mean that sentinel will try to save the config again on every cron run until the save succeeds? Do we need to backoff here or to limit the number of times we try?

@gmbnomis Sorry for the delayed reply to your comment raised half a year ago.
This PR mainly addresses the first requirement expected in #2282: reducing the proactive failover time for 2K+ nodes from approximately 10 minutes to 15 seconds.

**1. In which cases does postponing to write the config (by a second) help?**
Previously, every single failover triggered a full rewrite of the entire configuration. With a very large number of failover instances, this caused the failover time for 2K nodes to reach around 10 minutes, which was unacceptable.
By changing the logic to flush the config only once per second, we aggregate all failover results within that 1-second window. This drastically reduces the number of config writes and significantly improves failover efficiency — this is exactly where the 1-second batching helps.

**2. Some config flushes are forced, some are not. Why don’t the latter need immediate disk flush?**
This PR only optimizes config updates for large-scale node failover scenarios:
- +slave events
- Proactively execute failover
- Config updates upon receiving +switch-master messages
All other config update flows keep the original immediate/forced flush logic.

For the batched, non-forced flushes:
The 1-second delay is tiny and does not harm persistence timeliness.
The top priority is fixing the catastrophic 10-minute delay for large clusters, and this batching does not break configuration consistency or reliability.

**Responses to your other comments**
**Comment 1: Graceful shutdown with config_dirty > 0**
You’re right. To stay consistent with the original logic, we ensure that if config_dirty > 0 when Sentinel shuts down gracefully, one final config flush is performed. This guarantees all pending changes are persisted before exit.

**Comment 2: Config save failures and retry behavior**
Your concern is valid. We can improve this in two ways:
Add an observability metric to count how many times config saving fails.
Add a retry backoff mechanism: as consecutive failures increase, we extend the config save interval gradually, avoiding excessive frequent retries while still ensuring the dirty config will eventually be written once the issue is resolved.

Thank you again for your careful review and valuable feedback. If you have any further questions or concerns, please feel free to let me know.
Looking forward to your reply.

PR #2310, opened 2025-07-05
Add extensions supported to cluster link layer to propagate extension faster during handshake

enjoy-binbin:
Add support_extension to cluster link, so that when myself receives a message from sender,
even if myself does not know the sender, for example during handshake the sender node is
still treated as random node, in thi case, the packet we reply to sender can also carry the
extension, since the sender link explicitly indicates that it supports the extension, by doing
this, we can speed up the spread of the extension.

See https://github.com/valkey-io/valkey/pull/2283#issuecomment-3031446204 and
https://github.com/valkey-io/valkey/pull/2279#discussion_r2186673007 for more details.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -4282,6 +4282,8 @@ void clusterSendPing(clusterLink *link, int type) {
      * to put inside the packet. */
     estlen = sizeof(clusterMsg) - sizeof(union clusterMsgData);
     estlen += (sizeof(clusterMsgDataGossip) * (wanted + pfail_wanted));
+    /* If the link or the node indicates that it supports extensions, then we
+     * pass the extensions. */
     if (linkSupportsExtension(link) || (link->node && nodeSupportsExtensions(link->node))) {
         estlen += writePingExtensions(NULL, 0);
     }
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3353,7 +3353,7 @@ int clusterProcessPacket(clusterLink *link) {
     int sender_last_reported_as_replica = sender && nodeIsReplica(sender);
     int sender_last_reported_as_primary = sender && nodeIsPrimary(sender);
 
-    if (sender && (hdr->mflags[0] & CLUSTERMSG_FLAG0_EXT_DATA)) {
+    if (sender && link->support_extension) {
         sender->flags |= CLUSTER_NODE_EXTENSIONS_SUPPORTED;
     }
 
@@ -3471,7 +3471,7 @@ int clusterProcessPacket(clusterLink *link) {
                     memcpy(node->ip, ip, sizeof(ip));
                     getClientPortFromClusterMsg(hdr, &node->tls_port, &node->tcp_port);
                     node->cport = ntohs(hdr->cport);
-                    if (hdr->mflags[0] & CLUSTERMSG_FLAG0_EXT_DATA) {
+                    if (link->support_extension) {
                         node->flags |= CLUSTER_NODE_EXTENSIONS_SUPPORTED;
                     }
                     setClusterNodeToInboundClusterLink(node, link);
@@ -4271,9 +4271,7 @@ void clusterSendPing(clusterLink *link, int type) {
      * to put inside the packet. */
     estlen = sizeof(clusterMsg) - sizeof(union clusterMsgData);
     estlen += (sizeof(clusterMsgDataGossip) * (wanted + pfail_wanted));
-    /* If the link or the node indicates that it supports extensions, then we
-     * pass the extensions. */
-    if (link->support_extension || (link->node && nodeSupportsExtensions(link->node))) {
+    if (link->support_extension) {
         estlen += writePingExtensions(NULL, 0);
     }
     /* Note: clusterBuildMessageHdr() expects the buffer to be always at least
@@ -4349,7 +4347,7 @@ void clusterSendPing(clusterLink *link, int type) {
     /* Compute the actual total length and send! */
     uint32_t totlen = 0;
 
-    if (link->support_extension || (link->node && nodeSupportsExtensions(link->node))) {
+    if (link->support_extension) {
         totlen += writePingExtensions(hdr, gossipcount);
     } else {
         serverLog(LL_DEBUG, "Unable to send extensions data, however setting ext data flag to true");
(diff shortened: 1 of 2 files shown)


src/cluster_legacy.c
```diff
@@ -4267,7 +4271,9 @@ void clusterSendPing(clusterLink *link, int type) {
      * to put inside the packet. */
     estlen = sizeof(clusterMsg) - sizeof(union clusterMsgData);
     estlen += (sizeof(clusterMsgDataGossip) * (wanted + pfail_wanted));
-    if (link->node && nodeSupportsExtensions(link->node)) {
+    /* If the link or the node indicates that it supports extensions, then we
+     * pass the extensions. */
+    if (link->support_extension || (link->node && nodeSupportsExtensions(link->node))) {
```
```suggestion
    if (link->support_extension) {
```

src/cluster_legacy.c
```diff
@@ -4343,7 +4349,7 @@ void clusterSendPing(clusterLink *link, int type) {
     /* Compute the actual total length and send! */
     uint32_t totlen = 0;
 
-    if (link->node && nodeSupportsExtensions(link->node)) {
+    if (link->support_extension || (link->node && nodeSupportsExtensions(link->node))) {
```
```suggestion
    if (link->support_extension) {
```

hpatro (commented):
I think the problem is during handshake the sender node is still treated as random node and is still returned as `NULL` from the below method.

https://github.com/valkey-io/valkey/blob/dcf32220c914588123e6d7001a72e7643d428130/src/cluster_legacy.c#L3119-L3137

I think we can store these information at two places but use the `link->extension` explicitly to determine if we can send extensions or not and `node->flags` to gossip about extension support to others.

WDYT @enjoy-binbin ?

enjoy-binbin:
> I think the problem is during handshake the sender node is still treated as random node and is still returned as NULL from the below method.

1. Node B cluster reset soft, and remove all the nodes in its cluster_nodes_dict
2. Node A sends the PING, and Node B will reply a PONG even it does not know A, and in here, since Node B does not know Node A, it won't send the extension. But the link has mf_flags that we all know Node A can handle extension. So in these packets, Node A will know B change the role but not able to learn the shard-id

Let me confirm, we are saying the same thing? Node B does not know Node A (during handshake, random node), so just two different ways of saying it (yours is a little more specific)?

> I think we can store these information at two places but use the link->extension explicitly to determine if we can send extensions or not and node->flags to gossip about extension support to others.

SGTM, i will make the change

hpatro (approved):
LGTM

hpatro:
@madolson Could you also give a look at this?

hpatro:
> 2\. Node A sends the PING, and Node B will reply a PONG even it does not know A, and in here, since Node B does not know Node A, it won't send the extension. But the link has mf_flags that we all know Node A can handle extension. So in these packets, Node A will know B change the role but not able to learn the shard-id

Didn't follow the `mf_flags` part.

I was explaining from the code point of view where until we haven't finished the handshake and settled on the actual nodename, we don't retrieve the sender `node` object, which leads us to not send extensions message. Now with this change, the information moves to the link object and the `support_extension` field gets set on the first `PING`/`MEET` message itself and we are able to send extensions in the first `PONG` response itself.

enjoy-binbin:
It suddenly occurred to me that we might need to keep this node check? Like when myself add a unknown node from the gossip part, in this case, aren't we need this node condition to send the packet in the very first time (before myself recevie the packet)
```
if (link->support_extension || (link->node && nodeSupportsExtensions(link->node))) {
```

enjoy-binbin:
> Didn't follow the mf_flags part.

Sorry i gave the wrong info, i mean `if (hdr->mflags[0] & CLUSTERMSG_FLAG0_EXT_DATA) {`

> I was explaining from the code point of view where until we haven't finished the handshake and settled on the actual nodename, we don't retrieve the sender node object, which leads us to not send extensions message. Now with this change, the information moves to the link object and the support_extension field gets set on the first PING/MEET message itself and we are able to send extensions in the first PONG response itself.

great, good to know, i just want to be sure we are in the same page, that is a good explanation.

hpatro:
> It suddenly occurred to me that we might need to keep this node check? Like when myself add a unknown node from the gossip part, in this case, aren't we need this node condition to send the packet in the very first time (before myself recevie the packet)
> 
> ```
> if (link->support_extension || (link->node && nodeSupportsExtensions(link->node))) {
> ```

That's right 👍 .

src/cluster_legacy.c
```diff
@@ -3252,6 +3255,7 @@ int clusterIsValidPacket(clusterLink *link) {
                 explen += extlen;
                 ext = getNextPingExt(ext);
             }
+            link->support_extension = 1;
```
hpatro: Nit:

Btw if we move the extensions processing ahead of the core message processing. I would prefer this check to be moved to `clusterProcessPacket`, this feels like a bit of bypassing.

enjoy-binbin: ok, i made a new commit, please see if you like it.

hpatro: Tbh, I don't like the latest commit, most of the flags are unrelated to link and we are starting off by the 10th bit. If you want to save memory, we could start with another macro for links `LINK_EXTENSIONS_SUPPORTED (1<<0)` or use the `ClientFlags` approach. Will leave it upto you.

enjoy-binbin: Sorry if this does not suit you well. I just thought maybe we can re-use the node flags, but i do agree it is kind of overdoing.

src/cluster_legacy.c
```diff
@@ -3461,13 +3475,14 @@ int clusterProcessPacket(clusterLink *link) {
...
-                     * to reduce the amount of time needed to stabilize the shard ID. */
+                     * we want to propagate extensions support as part of the node flags
+                     * in the gossip section, so that we can send extension right away
+                     * in the future packet. */
                     clusterNode *node = createClusterNode(NULL, CLUSTER_NODE_HANDSHAKE);
                     memcpy(node->ip, ip, sizeof(ip));
                     getClientPortFromClusterMsg(hdr, &node->tls_port, &node->tcp_port);
                     node->cport = ntohs(hdr->cport);
-                    if (hdr->mflags[0] & CLUSTERMSG_FLAG0_EXT_DATA) {
+                    if ((flags & ~CLUSTER_NODE_EXTENSIONS_SUPPORTED) && linkSupportsExtension(link)) {
```
hpatro: Unsure about this line:

```
(flags & ~CLUSTER_NODE_EXTENSIONS_SUPPORTED)
```


Why check for other bits ?

enjoy-binbin: Just an idea to check flags explicitly during coding, it is indeed possible to overdo

hpatro (approved):

PR #2318, opened 2025-07-07
change check order for xautoclaim

charsyam:
change check order for xautocommit parameters

I think it is more readable and eailer.

because we already checked object is existed by lookupKeyRead at the beginning of code.

and it is a little bit faster when key is not exist.

### DIFF
--- a/src/t_stream.c
+++ b/src/t_stream.c
@@ -3317,6 +3317,13 @@ void xautoclaimCommand(client *c) {
     int startex;
     int justid = 0;
 
+    if (o == NULL) {
+        addReplyErrorFormat(c, "-NOGROUP No such key '%s'", (char *)c->argv[1]->ptr);
+        return;
+    }
+
+    if (checkType(c, o, OBJ_STREAM)) return; /* Type error. */
+
     /* Parse idle/start/end/count arguments ASAP if needed, in order to report
      * syntax errors before any other error. */
     if (getLongLongFromObjectOrReply(c, c->argv[4], &minidle, "Invalid min-idle-time argument for XAUTOCLAIM") != C_OK)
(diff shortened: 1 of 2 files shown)


src/t_stream.c
```diff
@@ -3317,6 +3317,13 @@ void xautoclaimCommand(client *c) {
     int startex;
     int justid = 0;
 
+    if (o == NULL) {
+        addReplyErrorFormat(c, "-NOGROUP No such key '%s'", (char *)c->argv[1]->ptr);
+        return;
+    }
```
sungming2, line 3323: I'm not sure this is the correct error message. If I understand correctly, `-NOGROUP` typically means the consumer group doesn't exist for the key, not that the key itself is missing. In this case, the error seems more related to the key not existing at all, so maybe a different error message would be more appropriate.

src/t_stream.c
```diff
@@ -3347,16 +3354,12 @@ void xautoclaimCommand(client *c) {
...
     }
 
-    if (o) {
-        if (checkType(c, o, OBJ_STREAM)) return; /* Type error. */
-        group = streamLookupCG(o->ptr, c->argv[2]->ptr);
-    }
+    group = streamLookupCG(o->ptr, c->argv[2]->ptr);
 
     /* No key or group? Send an error given that the group creation
      * is mandatory. */
```
sungming2, line 3360: We can change comment if it is necessary

src/t_stream.c
```diff
@@ -3347,16 +3354,12 @@ void xautoclaimCommand(client *c) {
...
-    if (o == NULL || group == NULL) {
-        addReplyErrorFormat(c, "-NOGROUP No such key '%s' or consumer group '%s'", (char *)c->argv[1]->ptr,
-                            (char *)c->argv[2]->ptr);
+    if (group == NULL) {
+        addReplyErrorFormat(c, "-NOGROUP No consumer group '%s'", (char *)c->argv[2]->ptr);
```
sungming2, line 3362: Can I know why the group null check is still needed here, now that the key existence check has been moved earlier? I'm also wondering if there is a specific reason to keep the group check separate.

daemyung-lablup: @sungming2 Good Question. before that line. we built group by streamLookupCG. so, at that time we can check group is exist.

```
group = streamLookupCG(o->ptr, c->argv[2]->ptr);
```

src/t_stream.c
daemyung-lablup: @sungming2 Good Catch. Thanks.

PR #2319, opened 2025-07-07
fix test Flushall while watching several keys by one client

ranshid:
There is a need to unwatch keys at the end of the test inm order to cleanup the dirty multi state so that it is not carried to other tests.


soloestoy (approved):

hpatro:
Just seeing this PR @ranshid, my bad. You should have pointed me to your PR 😕

This is fixed by #2323

PR #2323, opened 2025-07-07
Update test to unwatch all keys before auth within multi

### LINKED ISSUE #2320: [BUG] Failing test for release 8.1.3: `AUTH errored inside MULTI`
**Describe the bug**

```

*** [err]: AUTH errored inside MULTI will add the reply in tests/unit/multi.tcl
Expected an error matching 'WRONGPASS invalid username-password pair or user is disabled.' but got '' (context: type eval line 5 cmd {assert_error {WRONGPASS invalid username-password pair or user is disabled.} {r exec}} proc ::test)
make[1]: *** [Makefile:580: test] Error 1
make[1]: Leaving directory '/build/valkey/src/valkey-8.1.3/src'
make: *** [Makefile:6: test] Error 2
make: Leaving directory '/build/valkey/src/valkey-8.1.3'
```

8.1.2 tests are good.

**To reproduce**

Build ArchLinux PKGBUILD with `pkgctl build` (and bumped version).

**Expected behavior**

Tests pass.

**Additional information**

-

hpatro:
Needs backporting to 8.1 to fix https://github.com/valkey-io/valkey/issues/2320

### DIFF
--- a/tests/unit/multi.tcl
+++ b/tests/unit/multi.tcl
@@ -957,6 +957,7 @@ start_server {tags {"multi"}} {
 
     test "AUTH errored inside MULTI will add the reply" {
         r config set requirepass ""
+        r unwatch
         r multi
         r auth no-user foobar
         assert_error {WRONGPASS invalid username-password pair or user is disabled.} {r exec}
--- a/tests/unit/multi.tcl
+++ b/tests/unit/multi.tcl
@@ -900,6 +900,7 @@ start_server {tags {"multi"}} {
         r watch b{t} a{t}
         r flushall
         r ping
+        r unwatch
      }
 
     test {MULTI is rejected when CLIENT REPLY is ON/OFF/SKIP} {
@@ -957,7 +958,6 @@ start_server {tags {"multi"}} {
 
     test "AUTH errored inside MULTI will add the reply" {
         r config set requirepass ""
-        r unwatch
         r multi
         r auth no-user foobar
         assert_error {WRONGPASS invalid username-password pair or user is disabled.} {r exec}


hpatro:
Verified on top of 8.1.3, the test passes successfully with this change.

hpatro:
@ranshid Maybe we should trigger another 8.1 patch release. 8.1.3 build fails consistently because of this test failure. WDYT ?

ranshid:
> @ranshid Maybe we should trigger another 8.1 patch release. 8.1.3 build fails consistently because of this test failure. WDYT ?

I do not think we need to release just for that. we can just port a fix to 8.1 and it will be picked by the next release

tests/unit/multi.tcl
```diff
@@ -957,6 +957,7 @@ start_server {tags {"multi"}} {
 
     test "AUTH errored inside MULTI will add the reply" {
         r config set requirepass ""
+        r unwatch
```
ranshid: why not fix the test which actually "leaked" this? ("Flushall while watching several keys by one client")

hpatro: Thought about it and then saw we were cleaning up the requirepass here. So, added it here. I agree, let me clean it up at the source.

hpatro:
> > @ranshid Maybe we should trigger another 8.1 patch release. 8.1.3 build fails consistently because of this test failure. WDYT ?
> 
> I do not think we need to release just for that. we can just port a fix to 8.1 and it will be picked by the next release

Folks build from source and seem to be affected by this https://github.com/valkey-io/valkey/issues/2320

madolson:
I agree I don't think we should do a release for this.

madolson (approved):

PR #2328, opened 2025-07-08
handle prefetch-batch-max-size config update from zero to positive value

naglera:
Previously, onMaxBatchSizeChange was only called during processClientsCommandsBatch, making it impossible to change prefetch-batch-max-size from 0 to a positive number since the batch processing would never occur with size 0.

Now onMaxBatchSizeChange is called when the configuration changes, ensuring the new batch size takes effect. Note that the change may not be immediate and will be applied during the next batch processing cycle.


hpatro (approved):
Good catch.

PR #848, opened 2024-07-30
Standalone FAILOVER: Fix disconnect of blocked clients in standalone failover and support REDIRECT response

gmbnomis:
This PR consists of two commits: If a client waits for the response to a blocking command when a failover is initiated, it will be unblocked and disconnected during the failover.  While this needs to happen in general, it happens too early currently. The first commit lets the disconnect happen only _after_ a successful failover.  It can be considered as a bugfix and could be a candidate for a 7.2.x release.

Now that the disconnect happens at the correct point in time, we can redirect clients that support the "redirect" capability instead of disconnecting them. This is implemented in the second commit.

These changes are related to the bug described in issue #821.

More details are covered in the commit messages of the two commits:

**Fix early unblock/disconnect of clients during failover**

When entering the `FAILOVER_IN_PROGRESS` state during a failover,
`disconnectAllBlockedClients` is called (as part of
`replicationSetPrimary`) to disconnect all blocked clients (except
those that are postponed due to a client pause).

While this makes sense in case of a role change in general, this is
happening too early in case of failover:
1. The role switch hasn't happened in `FAILOVER_IN_PROGESS` yet,
   thus, there is no new primary to connect to.
2. The failover may still fail. In this case, there is no need to
   unblock/disconnect clients at all.

Therefore, move the `disconnectAllBlockedClients` to when we unblock
the paused clients after the failover completes. As said above, this
call must only happen if the failover succeeds.

Extend two tests to verify the changed behavior:

1. "failover to a replica with force works": Issue a blocking command
   before the failover and verify that it is unblocked after the
   failover finished.

2. "failover aborts if target rejects sync request": Both a blocking
  command issued before a failover as well as one issued during the
  failover won't be interrupted when the failover eventually
  aborts. (Note that the location of the previous `SET FOO BAR`
  command in the test was wrong, as it did not block and thus,
  wouldn't be "interrupted" by the failover in any case.)

**Role change: Redirect suitable clients instead of disconnecting them**

In case of an instance role switch in standalone mode, clients with
blocking commands get disconnected sending an UNBLOCKED error to the
client.

Now that the REDIRECT error is available as an alternative (#325),
issue a REDIRECT (redirecting the client to the primary) for clients
that are supporting it. There is no need to disconnect the client in
this case. This is in-line with the semantics of the MOVED error in
cluster mode.

PR #2329, opened 2025-07-08
Redirect blocked clients after failover

cjx-zar:
In standalone mode, after a switch over, the command that was originally blocking on primary returns -REDIRECT instead of -UNBLOCKED when the client has the redirect capability.

Similarly, in cluster mode, after a switch over, the blocked commands receive a -MOVED redirect instead of -UNBLOCKED.

After this fix, the handling of blocked connections during a switch over between standalone and cluster is nearly identical. This can be summarized as follows:

Standalone: 

1. Client that has the redirect capability blocked on the key on the primary node will receive a -REDIRECT after the switch over completes instead of -UNBLOCKED. 
2. Readonly clients blocked on the primary or replica node will remain blocked throughout the switch over.

Cluster:

1. Client blocked on the key served by the primary node will receive a -MOVED instead of a probabilistic -UNBLOCKED error. 
2. Readonly clients blocked on the key served by primary or replica node will remain blocked throughout the switch over.


src/blocked.c
```diff
@@ -335,9 +335,14 @@ void disconnectAllBlockedClients(void) {
              * which the command is already in progress in a way. */
             if (c->bstate->btype == BLOCKED_POSTPONE) continue;
 
-            unblockClientOnError(c, "-UNBLOCKED force unblock from blocking operation, "
-                                    "instance state changed (master -> replica?)");
-            c->flag.close_after_reply = 1;
+            if (c->capa & CLIENT_CAPA_REDIRECT) {
+                addReplyErrorSds(c, sdscatprintf(sdsempty(), "-REDIRECT %s:%d", server.primary_host, server.primary_port));
+                unblockClientOnError(c, NULL);
+            } else {
```
hpatro: Need to update the top level comment to reflect the same.

hpatro: I believe we would need similar checks, if primary_host is defined or not.

https://github.com/valkey-io/valkey/blob/dcf32220c914588123e6d7001a72e7643d428130/src/server.c#L4232

hpatro (commented):
Makes sense to me. ~~I believe this is a breaking change. @valkey-io/core-team Could you please vote 👍 / 👎  ?~~ 

Discussed below with @zuiderkwast about the minor behavior change and shouldn't be considered as a breaking change.

zuiderkwast:
@hpatro Why is it a breaking change?

There is no new error reply here. Cluster clients already handle MOVED and clients with the capa should already handle REDIRECT.

hpatro:
> @hpatro Why is it a breaking change?
> 
> There is no new error reply here. Cluster clients already handle MOVED and clients with the capa should already handle REDIRECT.

The client with capa `r` enabled is currently receiving `-UNBLOCKED` and will receive `-REDIRECT` going forward. Wouldn't that be considered a breaking change ?

zuiderkwast:
> > @hpatro Why is it a breaking change?
> > There is no new error reply here. Cluster clients already handle MOVED and clients with the capa should already handle REDIRECT.
> 
> The client with capa `r` enabled is currently receiving `-UNBLOCKED` and will receive `-REDIRECT` going forward. Wouldn't that be considered a breaking change ?

It's a behavior change, but any client that handles this also handles REDIRECT. If the failover happens just before the command is executed, the server already now returns REDIRECT. It just depends on timing. With this change, REDIRECT is return in both cases. There is no way for a client to tell the difference between these two cases.

Therefore, I'd argue it can't break anything, so it's backward compatible behavior... But please prove  me wrong!

Is this UNBLOCKED error after failover documented?

hpatro:
> > > @hpatro Why is it a breaking change?
> > > There is no new error reply here. Cluster clients already handle MOVED and clients with the capa should already handle REDIRECT.
> > 
> > 
> > The client with capa `r` enabled is currently receiving `-UNBLOCKED` and will receive `-REDIRECT` going forward. Wouldn't that be considered a breaking change ?
> 
> It's a behavior change, but any client that handles this also handles REDIRECT. If the failover happens just before the command is executed, the server already now returns REDIRECT. It just depends on timing. With this change, REDIRECT is return in both cases. There is no way for a client to tell the difference between these two cases.

Yeah, I was trying to highlight that. I think I agree with you, it's pretty subtle and shouldn't bother clients which already handle the `-REDIRECT`. Good to remove the vote/breaking change label then?

hpatro:
@cjx-zar Seems like the test is flaky. Could you take a look?

https://github.com/valkey-io/valkey/actions/runs/16141887799/job/45595086463?pr=2329
```
[err]: client with REDIRECT capability receives REDIRECT error during failover in tests/integration/replica-redirect.tcl
Failover did not start
```

src/blocked.c
cjx-zar: OK! Comments updated. 
server.primary_host has just been set here, and the obey_client check should not be necessary IMHO?
```
@@ -4078,7 +4078,6 @@ void replicationSetPrimary(char *ip, int port, int full_sync_required) {
         server.primary->flag.dont_cache_primary = full_sync_required;
         freeClient(server.primary);
     }
-    disconnectAllBlockedClients(); /* Clients blocked in primary, now replica. */

     /* Setting primary_host only after the call to freeClient since it calls
      * replicationHandlePrimaryDisconnection which can trigger a re-connect
@@ -4086,6 +4085,10 @@ void replicationSetPrimary(char *ip, int port, int full_sync_required) {
     server.primary_host = sdsnew(ip);
     server.primary_port = port;

+    if (!server.cluster_enabled) {
+        disconnectAllBlockedClients(); /* Clients blocked in primary, now replica. */
+    }
```

cjx-zar:
> @cjx-zar Seems like the test is flaky. Could you take a look?
> 
> https://github.com/valkey-io/valkey/actions/runs/16141887799/job/45595086463?pr=2329
> 
> ```
> [err]: client with REDIRECT capability receives REDIRECT error during failover in tests/integration/replica-redirect.tcl
> Failover did not start
> ```

Sure, the case may be a little unstable, I have fixed it.

src/blocked.c
soloestoy: yes, cluster enabled or disabled needs different reply.

hpatro: @cjx-zar yeah looks right to me, missed it. Thanks.

cjx-zar:
Sorry for doing the force push because I forgot to sign the previous commit...

src/replication.c
```diff
@@ -4078,14 +4078,17 @@ void replicationSetPrimary(char *ip, int port, int full_sync_required) {
...
     }
-    disconnectAllBlockedClients(); /* Clients blocked in primary, now replica. */
 
     /* Setting primary_host only after the call to freeClient since it calls
      * replicationHandlePrimaryDisconnection which can trigger a re-connect
      * directly from within that call. */
     server.primary_host = sdsnew(ip);
     server.primary_port = port;
 
+    if (!server.cluster_enabled) {
```
ranshid: minor: can we add some comment above stating why this HAS to come after we set the new primary host (since unblocking clients might require redirecting to the new primary)

BTW - I think in the current implementation clients blocked on xread are also disco. we might consider changing this in a separate PR

ranshid (commented):
left a minor comment.

Worth mentioning that the client cron might NOT handle all clients in one cycle.
Is it possible that in such a case there might be blocked write clients which are left hanging and in case they will be unblocked (eg by replication stream writing to a list) they might attempt to do the write and return an error?
I am not sure this is horrifying, but it might be counted differently by applications suddenly getting unexpected errors. maybe we could have call the 
```
if (server.cluster_enabled) {
            if (clusterRedirectBlockedClientIfNeeded(c)) unblockClientOnError(c, NULL);
        }
```

from the disconnectAllBlockedClients?

hpatro:
Thanks @ranshid for taking a look. Didn't quite catch the flow. Could you help @cjx-zar create a test case for that?

At least the changes you suggest feels complete as we handle unblock condition for both standalone and cluster in a single code flow.

cjx-zar:
> left a minor comment.
> 
> Worth mentioning that the client cron might NOT handle all clients in one cycle. Is it possible that in such a case there might be blocked write clients which are left hanging and in case they will be unblocked (eg by replication stream writing to a list) they might attempt to do the write and return an error? I am not sure this is horrifying, but it might be counted differently by applications suddenly getting unexpected errors. maybe we could have call the
> 
> ```
> if (server.cluster_enabled) {
>             if (clusterRedirectBlockedClientIfNeeded(c)) unblockClientOnError(c, NULL);
>         }
> ```
> 
> from the disconnectAllBlockedClients?

Makes sense. Should we remove the same code in `clientsCronHandleTimeout` to make `clusterRedirectBlockedClientIfNeeded` only called in `disconnectAllBlockedClients`?

soloestoy:
I tend to handle both cluster-enabled (-MOVED) and cluster-disabled (-REDIRECT) uniformly in `disconnectAllBlockedClients`.

cjx-zar:
> I tend to handle both cluster-enabled (-MOVED) and cluster-disabled (-REDIRECT) uniformly in `disconnectAllBlockedClients`.

Done :)

src/blocked.c
```diff
@@ -318,7 +318,9 @@ void replyToClientsBlockedOnShutdown(void) {
  * in an instance which turns from primary to replica is unsafe, so this function
  * is called when a primary turns into a replica.
  *
- * The semantics is to send an -UNBLOCKED error to the client, disconnecting
+ * The semantics is try to send a -MOVED to the client when cluster_enabled,
+ * try to send a -REDIRECT when the client has redirect capability when cluster-disabled,
+ * and send an -UNBLOCKED error to the client, disconnecting
```
hpatro:
```suggestion
 * The semantics are as follows:  
 * - Send a -MOVED to the client in cluster-enabled mode.  
 * - Send a -REDIRECT when the client has redirect capability in standalone mode.  
 * - Otherwise, send a -UNBLOCKED error to the client while disconnecting it at the same time.
```

src/replication.c
```diff
@@ -4078,14 +4078,18 @@ void replicationSetPrimary(char *ip, int port, int full_sync_required) {
...
     /* Setting primary_host only after the call to freeClient since it calls
      * replicationHandlePrimaryDisconnection which can trigger a re-connect
      * directly from within that call. */
     server.primary_host = sdsnew(ip);
     server.primary_port = port;
 
+    /* Clients blocked in primary, now replica.
+     * Note that we need to set the new primary first
+     * since unblocking clients require redirecting to the new primary */
+    disconnectAllBlockedClients();
```
gmbnomis: I proposed the same functionality for REDIRECT a while ago, see https://github.com/valkey-io/valkey/pull/848/. Redirecting the client here is too early in the failover case:

1. The role switch hasn't happened in `FAILOVER_IN_PROGESS` yet,
    thus, there is no new primary to connect to. The designated primary is
   still a replica and wil promptly redirect the client back to this node.
2. The failover may still fail. In this case, there is no need to
    unblock/disconnect clients at all.

Therefore, https://github.com/valkey-io/valkey/pull/848/ moved the `disconnectAllBlockedClients` to after the failover completes successfully and the new primary is ready to accept clients.

src/timeout.c
```diff
@@ -63,12 +63,6 @@ int clientsCronHandleTimeout(client *c, mstime_t now_ms) {
         serverLog(LL_VERBOSE, "Closing idle client");
         freeClient(c);
         return 1;
-    } else if (c->flag.blocked) {
-        /* Cluster: handle unblock & redirect of clients blocked
-         * into keys no longer served by this server. */
-        if (server.cluster_enabled) {
-            if (clusterRedirectBlockedClientIfNeeded(c)) unblockClientOnError(c, NULL);
-        }
```
gmbnomis, line 71: `clusterRedirectBlockedClientIfNeeded()` here handles a lot more cases than `disconnectAllBlockedClients`. We can't simply stop to call it.

src/replication.c
ranshid: @gmbnomis I must say I am sorry I completely forgot about https://github.com/valkey-io/valkey/pull/848 (why was it closed BTW?).
I am trying to recall the reasoning behind keeping the disconnectAllBlockedClients in cluster mode. I think that before we introduce blocked client reprocessing it was problematic to keep blocked clients till failover is fully completed (as some blocked clients could have attempted to change the data). However now that we DO reprocess the commands it might be O.K to let them follow the regular path and allow them to be redirected by the clients timeout cron. I think it is indeed a problem that this cron is active while we pause clients though... maybe there are pros and cons for that, but would preventing this logic while clients are paused would solve the issue of failed failover?

gmbnomis: > @gmbnomis I must say I am sorry I completely forgot about #848 (why was it closed BTW?). 

I realized only after posting that PR that cluster mode simply disconnects as well. I thought that I will have a hard time justifying such a change and closed it after the discussion petered out. It did not come to my mind that we could try to address both cluster and redirect mode.

> However now that we DO reprocess the commands it might be O.K to let them follow the regular path and allow them to be redirected by the clients timeout cron. 

I reckon we don't have the time to get this picked up by a cron run in this case:  When we switch a primary to a replica and keep e.g. a `BLPOP` blocked, it might happen that we get a replicated `RPUSH` before cron runs for that client. So, I suppose we need to do something with these clients before the role change (like it is currently done) or, if we paused the clients, directly when the pause ends.

(I just found your comment above that mentions exactly this scenario, so you obviously thought of it...

IIUC, in cluster node, we could do nothing in `replicationSetPrimary` w.r.t. to blocked clients. Then, there are three triggers to unblock the command:

1. An event that unblocks the client (replicated write), but when re-executing the command, we will find out that  we can't execute the command and we will redirect or return an error. (Or, in specific cases (READONLY client blocking on XREAD), we will actually be able to execute the command)
2. Cron timeout picks it up and issues a redirect/returns an error
3. Block's timeout expires. 

This would delay the redirection a little, but simplify the logic. And possibly, we could do something similar for REDIRECT mode.)

> I think it is indeed a problem that this cron is active while we pause clients though... maybe there are pros and cons for that, but would preventing this logic while clients are paused would solve the issue of failed failover?

I am not familiar with the specifics of CLUSTER FAILOVER. However, I understand your point: this issue could interfere with failover, whether we use explicit logic to redirect after a failover or rely on cron to manage all redirects.

zuiderkwast: Best would be to redirect directly when client pause ends. +1 on that.

cjx-zar: Not sure if I understand it clearly, so the conclusion is that we should handle both -MOVED and -REDIRECT logic in `clientsCron` and remove `disconnectAllBlockedClients`?

zuiderkwast: I didn't review the code, but i thought directly here means don't wait for the next cron cycle to run.

gmbnomis: Here is my conclusion from the discussion:

- `disconnectAllBlockedClients` is basically the function proposed by this PR (I think we need to be more specific when redirecting REDIRECT clients, but that's another discussion. And we might want to rename the function)
- `disconnectAllBlockedClients` is _not_ called in case of a failover in `replicationSetPrimary` (#848 proposed to introduce a flag for this. This way the caller can decide whether it wants to take care of clients itself)
- In standalone failover, `disconnectAllBlockedClients` is called when we unpause clients after a successful failover (see #848)
- I don't know much about cluster failover, but, analogously, `disconnectAllBlockedClients` is called when we unpause clients. Here, it probably does not matter whether the failover was successful or not, since only those clients that need to be redirected will be unblocked.
- In `clientsCronHandleTimeout` we might need to prevent clients from being redirected too early during a cluster failover (@ranshid proposed to not run this logic during the client pause)
- `clientsCronHandleTimeout` does not need to handle standalone redirect mode

@ranshid,  @zuiderkwast Did I get this correctly?

soloestoy:
After reread #848 , I realize that it is a very complex case. We need to handle clients `blocked on keys` during failover while also considering the possibility that the instance may be in a `failover-in-progress` state. I think there are 3 cases we should handle carefully:
1. Simplest case: No clients are `blocked on keys` → No action is required.
2. Clients `blocked on keys`, but the instance is NOT in a `failover-in-progress` state:
    * Simply return `-REDIRECT` or `-UNBLOCKED` based on the capability.
3. Clients `blocked on keys` AND the instance is in a `failover-in-progress state`, special handling is required:
    * Temporarily ignore the blocked clients until the failover completes. At that point, it will fall back to case 2 for resolution.
    * Temporarily ignore the blocked clients. If the failover fails and is aborted, these clients can be left unhandled, as they have their own timeout mechanisms.
    * ~~Temporarily ignore the blocked clients. However, if these clients timeout and retry during the failover period, they will return to `processCommand`. At this stage, they might either be postponed or receive a `-READONLY` error.~~

ranshid:
> After reread #848 , I realize that it is a very complex case. We need to handle clients `blocked on keys` during failover while also considering the possibility that the instance may be in a `failover-in-progress` state. I think there are 3 cases we should handle carefully:
> 
> 1. Simplest case: No clients are `blocked on keys` → No action is required.
> 2. Clients `blocked on keys`, but the instance is NOT in a `failover-in-progress` state:
>    
>    * Simply return `-REDIRECT` or `-UNBLOCKED` based on the capability.
> 3. Clients `blocked on keys` AND the instance is in a `failover-in-progress state`, special handling is required:
>    
>    * Temporarily ignore the blocked clients until the failover completes. At that point, it will fall back to case 2 for resolution.
>    * Temporarily ignore the blocked clients. If the failover fails and is aborted, these clients can be left unhandled, as they have their own timeout mechanisms.
>    * Temporarily ignore the blocked clients. However, if these clients timeout and retry during the failover period, they will return to `processCommand`. At this stage, they might either be postponed or receive a `-READONLY` error.

>* Temporarily ignore the blocked clients. However, if these clients timeout and retry during the failover period, they will return to `processCommand`. At this stage, they might either be postponed or receive a `-READONLY` error.

@soloestoy what do you mean by `timeout and retry`? blocked client does not reattempt on timeout (maybe a module?) 

I think that a blocked client getting a -READONLY error is bad and I prefer to get -UNBLOCKED in this case, however I am not sure when exactly we can reach that point?

soloestoy:
> @soloestoy what do you mean by timeout and retry? blocked client does not reattempt on timeout (maybe a module?)
>
>I think that a blocked client getting a -READONLY error is bad and I prefer to get -UNBLOCKED in this case, however I am not sure when exactly we can reach that point?

~~the block command may be executed before failover, for example:~~
~~1. 00:00 blpop null 3~~
~~2. 00:01 failover to xxx 6379 timeout 5000, server paused~~
~~3. 00:02 server steps into `failover-in-progress` state, change role to replica (in `failover-in-progress` state, we don't disconnect blocked clients)~~
~~4. 00:03 blpop timeout -- reprocess command and server is replica, get -READONLY error~~
~~5. 00:04 failover succeed or abort~~

ranshid:
> 4\. blpop timeout -- reprocess command and server is replica, get -READONLY error

that part. AFAIK we do not do that. we handle timeout in handleBlockedClientsTimeout and we reply with error to the client that it was timedout

cjx-zar:
Hey guys, sorry for the delay to reply. The new code and test refer to part of https://github.com/valkey-io/valkey/pull/848

src/blocked.c
```diff
@@ -335,9 +337,17 @@ void disconnectAllBlockedClients(void) {
              * which the command is already in progress in a way. */
             if (c->bstate->btype == BLOCKED_POSTPONE) continue;
 
-            unblockClientOnError(c, "-UNBLOCKED force unblock from blocking operation, "
-                                    "instance state changed (master -> replica?)");
-            c->flag.close_after_reply = 1;
+            if (canRedirectClient(c) && (c->bstate->btype == BLOCKED_LIST || c->bstate->btype == BLOCKED_ZSET ||
```
soloestoy: it's better to distinguish and handle cluster and standalone modes separately:

```c
        if (server.cluster_enabled) {
            if (clusterRedirectBlockedClientIfNeeded(c))
                unblockClient(c);
        } else {
            if (canRedirectClient(c) && (c->bstate->btype == BLOCKED_LIST || c->bstate->btype == BLOCKED_ZSET ||
                                         c->bstate->btype == BLOCKED_STREAM || c->bstate->btype == BLOCKED_MODULE)) {
                if (c->bstate->btype == BLOCKED_MODULE && !moduleClientIsBlockedOnKeys(c)) continue;
                /* Client has redirect capability and blocked on keys */
                addReplyErrorSds(c, sdscatprintf(sdsempty(), "-REDIRECT %s:%d", server.primary_host, server.primary_port));
                unblockClientOnError(c, NULL);
            } else {
                unblockClientOnError(c, "-UNBLOCKED force unblock from blocking operation, "
                                        "instance state changed (master -> replica?)");
                c->flag.close_after_reply = 1;
            }
        }
```

soloestoy (approved):

soloestoy:
LGTM, @hpatro @zuiderkwast @ranshid @gmbnomis PTAL

src/replication.c
```diff
@@ -5254,7 +5260,7 @@ void updateFailoverStatus(void) {
                       server.target_replica_port);
             server.failover_state = FAILOVER_IN_PROGRESS;
             /* If timeout has expired force a failover if requested. */
-            replicationSetPrimary(server.target_replica_host, server.target_replica_port, 0);
+            replicationSetPrimary(server.target_replica_host, server.target_replica_port, 0, 0);
```
zuiderkwast: This function has a comment that says that blocked clients are disconnected if failover fails. Maybe it should be updated?

```c
/* Failover cron function, checks coordinated failover state.
 *
 * Implementation note: The current implementation calls replicationSetPrimary()
 * to start the failover request, this has some unintended side effects if the
 * failover doesn't work like blocked clients will be unblocked and replicas will
 * be disconnected. This could be optimized further.
 */
void updateFailoverStatus(void) {
```

cjx-zar: Done ：）

src/replication.c
```diff
@@ -4064,7 +4064,7 @@ int cancelReplicationHandshake(int reconnect) {
 }
 
 /* Set replication to the specified primary address and port. */
-void replicationSetPrimary(char *ip, int port, int full_sync_required) {
+void replicationSetPrimary(char *ip, int port, int full_sync_required, int disconnect_blocked) {
```
ranshid: We are trying to get rid of boolean integers. can we make it a bool?

src/replication.c
```diff
@@ -5084,13 +5089,14 @@ const char *getFailoverStateString(void) {
 /* Resets the internal failover configuration, this needs
  * to be called after a failover either succeeds or fails
  * as it includes the client unpause. */
-void clearFailoverState(void) {
+void clearFailoverState(int success) {
```
ranshid: same as last comment. van we use bool instead of integer?

src/cluster_legacy.c
```diff
@@ -2513,7 +2513,7 @@ void clusterProcessGossipSection(clusterMsg *hdr, clusterLink *link) {
                  * updated correctly, and we will not have the opportunity to call
                  * replicationSetPrimary and update the primary host. */
                 if (nodeIsReplica(myself) && myself->replicaof == node)
-                    replicationSetPrimary(node->ip, getNodeDefaultReplicationPort(node), 0);
+                    replicationSetPrimary(node->ip, getNodeDefaultReplicationPort(node), 0, 1);
```
ranshid: IIRC this flow will be called when the primary's address change. do we really need to disco blocked clients here?

src/cluster_legacy.c
```diff
@@ -2607,7 +2607,7 @@ int nodeUpdateAddressIfNeeded(clusterNode *node, clusterLink *link, clusterMsg *
     /* Check if this is our primary and we have to change the
      * replication target as well. */
     if (nodeIsReplica(myself) && myself->replicaof == node)
-        replicationSetPrimary(node->ip, getNodeDefaultReplicationPort(node), 0);
+        replicationSetPrimary(node->ip, getNodeDefaultReplicationPort(node), 0, 1);
```
ranshid: same as before. in this flow I am not sure we should disco blocked clients. we were replica before and we will not perform re-sync.

src/cluster_legacy.c
```diff
@@ -5562,7 +5562,7 @@ void clusterCron(void) {
      * enable it if we know the address of our primary and it appears to
      * be up. */
     if (nodeIsReplica(myself) && server.primary_host == NULL && myself->replicaof && nodeHasAddr(myself->replicaof)) {
-        replicationSetPrimary(myself->replicaof->ip, getNodeDefaultReplicationPort(myself->replicaof), 0);
+        replicationSetPrimary(myself->replicaof->ip, getNodeDefaultReplicationPort(myself->replicaof), 0, 1);
```
ranshid: same as before. in this flow I am not sure we should disco blocked clients. we were replica before and we will not perform re-sync.

src/server.c
```diff
@@ -3462,6 +3462,10 @@ int mustObeyClient(client *c) {
     return c->id == CLIENT_ID_AOF || c->flag.primary;
 }
 
+int canRedirectClient(client *c) {
```
ranshid: can we return bool here?

tests/integration/failover.tcl
```diff
@@ -140,6 +140,16 @@ start_server {overrides {save {}}} {
...
+        set rd_blocking [valkey_deferring_client -2]
+
+        $rd_blocking BRPOP FOO_LIST 0
+
+        wait_for_condition 50 100 {
+            [s -2 blocked_clients] == 1
+        } else {
+            fail "rd_blocking client should be blocked"
+        }
+
```
ranshid, line 152: Not critical, but I prefer to have tests for a specific scenario separated. It is tru that this way we can reduce the amount of test runtime but it also makes the tests coverage more complex to control. I would suggest separate this to another test

tests/integration/failover.tcl
```diff
@@ -165,6 +181,10 @@ start_server {overrides {save {}}} {
         } else {
             fail "Failover from node 2 to node 0 did not finish"
         }
+
+        assert_error "UNBLOCKED force unblock from blocking operation, instance state changed (master -> replica?)" {$rd_blocking read}
+        $rd_blocking close
+
```
ranshid, line 187: should we also check for blocked_clients to be 0?

tests/integration/failover.tcl
```diff
@@ -255,19 +275,38 @@ start_server {overrides {save {}}} {
...
+            [s 0 blocked_clients] == 1
         } else {
-            fail "Client creation failed"
+            fail "rd_brpop_before should be blocked"
         }
-        $rd SET FOO BAR
+
         $node_0 failover to $node_1_host $node_1_port
-        resume_process [srv -1 pid]
+        after 1000
```
ranshid: can we instead wait till node_0 is in state "failover-in-progress"?

tests/integration/replica-redirect.tcl
```diff
@@ -88,12 +93,154 @@ start_server {tags {needs:repl external:skip}} {
...
+
+            assert_match *slave* [r role]
+            assert_match *master* [r -1 role]
             assert_error "REDIRECT $replica_host:$replica_port" {$rd read}
             $rd close
         }
     }
 }
+
+start_cluster 1 1 {tags {external:skip cluster}} {
```
ranshid: is it possible to merge this with another batch of tests which already has cluster setup, instead of spawning new cluster?

src/server.c
```diff
@@ -3462,6 +3462,10 @@ int mustObeyClient(client *c) {
...
+    return !server.cluster_enabled && server.primary_host && c->capa & CLIENT_CAPA_REDIRECT;
```
zuiderkwast, line 3492: The name of this function doesn't hint that it's only for standalone mode. (It always returns false in cluster mode.)

Can we make sure this function is only used in standalone mode and do an assert that cluster is disabled? I think it removes some confusion about cluster redirects.
 
```suggestion
    serverAssert(!server.cluster_enabled);
    return server.primary_host && c->capa & CLIENT_CAPA_REDIRECT;
```

Where we call it, we could check cluster mode before calling it, like

```
    if (!server.cluster_enabled && canRedirectClient(c)) { ... }
```

src/replication.c
cjx-zar: Sure, I just didn't notice that Valkey starts using the boolean type before.

src/cluster_legacy.c
cjx-zar: IMHO, it might be safer to try to MOVED to a new master, prevent the connection from being blocked on an outdated replica for a long time.

src/server.c
cjx-zar: Since `REDIRECT` is an error code only in standalone mode, I included the cluster_enabled check in `canRedirectClient`. I'm OK with your suggestion too.

src/replication.c
zuiderkwast: It was decided here:

* #1278

We allow bool now but we don't require to change old code to bool. I guess it's also OK to keep the old style when changing old code (such as here). I don't have a strong opinion though.

src/server.c
zuiderkwast: Ah, so Redirect in the name canRedirectClient refers only to the REDIRECT reply? We often refer to MOVED and ASK as redirects.

tests/integration/failover.tcl
gmbnomis: Actually, this does not seem to reach `failover-in-progress`, but is in `waiting-for-sync` state if the test works.

And, surprisingly, it does not work in `failover-in-progress` state (the reason is unrelated to this PR):

Because the node already looks like a replica during `failover-in-progress`, this condition in `processCommand()` is true:

https://github.com/valkey-io/valkey/blob/33a43f2a8a66d751b88bce83f5f6bc9921b9dcb7/src/server.c#L4356

and the condition is checked before the pause state later in the same function:

https://github.com/valkey-io/valkey/blob/33a43f2a8a66d751b88bce83f5f6bc9921b9dcb7/src/server.c#L4426

So, during `failover-in-progress`, the write pause is completely ineffective for regular clients (REDIRECT clients are blocked separately) and instead, clients get a `READONLY You can't write against a read only replica` error instead of being blocked.

I think this is a bug.

cjx-zar: What I find a bit odd is that if the tests depend on the original master staying in the `waiting-for-sync`, then adding the `after 1000` should make it worse?

gmbnomis: Yes, this is odd indeed. Basically, waiting here shouldn't make a difference.

For me, the test begins to fail when I add a  `after 2000` before the `pause_process $node_1_pid` above. This seems to give the replica enough time to fully sync.

src/cluster_legacy.c
ranshid: IMO this will just cause stream readers to be shifted to the primary without any good reason. I understand that on role changes, it is important to keep, but on cases were there was no "logical" change I thin we should favor keeping the  existing propose of the client.

cjx-zar: Sounds reasonable, I've modified it.

zuiderkwast:
We talked about this in the core team meeting today but we did't reach a conclusion.

@ranshid Will you follow up?

madolson:
@ranshid Follow up!

ranshid:
Sorry - slipped under my radar. will followup tomorrow

src/server.c
ranshid: I agree with @zuiderkwast that we can consider having a more explicit name e.g. `clientSupportStandAloneRedirect`or something

src/replication.c
```diff
@@ -5460,9 +5466,8 @@ void failoverCommand(client *c) {
 /* Failover cron function, checks coordinated failover state.
  *
  * Implementation note: The current implementation calls replicationSetPrimary()
- * to start the failover request, this has some unintended side effects if the
- * failover doesn't work like blocked clients will be unblocked and replicas will
- * be disconnected. This could be optimized further.
+ * to start the failover request, but blocked clients are no longer processed during
+ * failover. Instead, they are deferred and handled later in clearFailoverState().
```
ranshid:
```suggestion
 * failover. Instead, their handling is deferred and performed later in clearFailoverState().
```

ranshid:
I think this PR is getting close. I wonder if separate disconnectOrRedirectAllBlockedClients from replicationSetPrimary would help clarify and simplify the code as we will have to be more explicit on when we trigger it (all the replicationSetPrimary with the boolean flag confuse me) but the current approach does have an advantage to guarantee we have coverage (So I would keep it for now)

I think. this change is risky though.
1. Since we now only redirect/disconnect clients in some of the code paths which set the target primary, we have a risk of missing some flows (which we did not consider) which might be unpleasant. 
2. Redirect is somewhat different than disconnect. I am not 100% sure how clients will react to redirect when they wanted to target replica. When we disconnect a blocked client the client library will repeat the choosing logic and will target the node according to that logic. However, when we redirect we are taking the ownership on that logic and we might miss correlating with the library logic.

@soloestoy / @gmbnomis  as you also reviewed this, do you have any concerns with the current imp[lementation?

ranshid (approved):
I will provide a tentative approval.
I still think there are some open comments, but they are minor or cosmetic.

soloestoy:
Mostly LGTM, please modify the `clientSupportStandAloneRedirect` part @cjx-zar , then we can merge it

cjx-zar:
Ok, I have changed the function name :)

zuiderkwast (approved):
`disconnectOrRedirectAllBlockedClients()` is only called from `replicationSetPrimary()`, so this code only runs when a primary turns into a replica, right? It never runs when a replica turns into a primary or some other scenario?

If the above is true, then I think this is very safe.

Just curious, what do we do with clients blocked on a replica (using `XREAD`, one of the few blocking readonly commands) when the replica turns into a primary? We don't disconnect them or redirect them, but we just continue blocking?

ranshid:
@cjx-zar can you please check the CI failure? it is new in your changes

zuiderkwast:
Is the title and top-comment correct? I can't see that this PR changed anything for slot migration. Please update the top-comment if needed, so we can correctly use these as the commit message.

cjx-zar:
> @cjx-zar can you please check the CI failure? it is new in your changes

Sure, but I can't reproduce this error on my own machine. I checked the test logic and found no problem. I have added debug logs where the test error occurs. I hope it can help.

zuiderkwast:
I edited the top comment and deleted the text that seemed incorrect to me about the old function name that doesn't exist anymore and about slot migration which doesn't look like it's affected but I'm might be wrong. Please correct it if so.

gmbnomis:
@zuiderkwast:

> Just curious, what do we do with clients blocked on a replica (using `XREAD`, one of the few blocking readonly commands) when the replica turns into a primary? We don't disconnect them or redirect them, but we just continue blocking?

In the current state of this PR, the behavior seems inconsistent:
- In standalone redirect mode, a READONLY client gets a "REDIRECT" response for a blocking `XREAD`. This is in contrast to my original proposal, see https://github.com/valkey-io/valkey/pull/848/files#diff-fdef28e02ac8048541ebebbb829193ffd25622dc438bd8d921b167e51564bdd0R152
- In cluster mode, a READONLY client seems to be blocked still (the already existing behavior of `clusterRedirectBlockedClientIfNeeded`)

Has this change been discussed, or is this an oversight?

tests/integration/replica-redirect.tcl
```diff
@@ -93,12 +98,154 @@ start_server {tags {needs:repl external:skip}} {
...
+
+            set rd_wait [valkey_deferring_client -1]
+            $rd_wait client capa redirect
+            $rd_wait read ; # Consume the OK reply
+            $rd_wait set foo bar
+            $rd_wait read
+            $rd_wait wait 1 0 ; # Blocks as we can't sync
+
+            # XREAD is a reading command and thus, should
+            # not be redirected if the client is read only.
```
gmbnomis, line 131: This comment does not match the assertion below, which tests for a redirect response.

(See my other comment, I think this comment is right and we shouldn't get a REDIRECT response here)

tests/integration/replica-redirect.tcl
```diff
@@ -93,12 +98,154 @@ start_server {tags {needs:repl external:skip}} {
...
+            [s -1 master_link_status] eq {up}
+        } else {
+            fail "Replica not online."
+        }
+
+        test {client with REDIRECT capability receives REDIRECT error during failover} {
+            set rd [valkey_deferring_client]
+            $rd client capa redirect
+            assert_match "OK" [$rd read]
+
```
gmbnomis, line 216: Perhaps we should verify the `XREAD` case for a ` READONLY` client as well here?

soloestoy:
interesting, [Allow blocked XREAD on a cluster replica #7881](https://github.com/redis/redis/pull/7881) is a new feature in 6.0, we never notice it.

for consistency, we can allow blocked XREAD on a standalone replica.

ranshid:
> @zuiderkwast:
> 
> > Just curious, what do we do with clients blocked on a replica (using `XREAD`, one of the few blocking readonly commands) when the replica turns into a primary? We don't disconnect them or redirect them, but we just continue blocking?
> 
> In the current state of this PR, the behavior seems inconsistent:
> 
> * In standalone redirect mode, a READONLY client gets a "REDIRECT" response for a blocking `XREAD`. This is in contrast to my original proposal, see https://github.com/valkey-io/valkey/pull/848/files#diff-fdef28e02ac8048541ebebbb829193ffd25622dc438bd8d921b167e51564bdd0R152
> * In cluster mode, a READONLY client seems to be blocked still (the already existing behavior of `clusterRedirectBlockedClientIfNeeded`)
> 
> Has this change been discussed, or is this an oversight?

I fully agree. I also gave this comment earlier. I think that this is one of the cases were we take the ownership about re-placing the blocked client instead of having the client operate this logic. IMO we should only redirect when the node role has changed (replica <--> primary). In case the client was blocked on a primary node we should have it redirected to the new primary and in the case the client was blocked on a replica it should stay on that replica. I am not sure the READONLY is a good enough reason to keep the blocked client on a failed over primary (which is turning into a replica) and IIRC the current cluster redirect logic in clusterRedirectBlockedClientIfNeeded will not redirect READONLY clients only if they were on a replica node (which is fine IMO)

zuiderkwast:
@gmbnomis Yes it seems to be an oversight as @soloestoy and @ranshid also agree.

In the standalone mode scenario *a replica turns into a primary and a client is blocked on XREAD, continue blocking*, let's do it in a separate PR to avoid scope creep? Then we can merge this PR as is, because it has been reviewed and approved in its current form.

ranshid:
> @gmbnomis Yes it seems to be an oversight as @soloestoy and @ranshid also agree.
> 
> In the standalone mode scenario _a replica turns into a primary and a client is blocked on XREAD, continue blocking_, let's do it in a separate PR to avoid scope creep? Then we can merge this PR as is, because it has been reviewed and approved in its current form.

@zuiderkwast I think only if this is O.K to release this PR as part of 9.0 and risking the followup "fix" in the next patch/minor release? This sounds like it might cause some degradation if we will start redirecting all the blocked clients to the primary node and than cause some CPU load when the stream is written?

tests/integration/replica-redirect.tcl
zuiderkwast: So far, this suite has no cluster tests. Should these be in a file under tests/unit/cluster/ instead?

cjx-zar:
> In the current state of this PR, the behavior seems inconsistent:
> 
> * In standalone redirect mode, a READONLY client gets a "REDIRECT" response for a blocking `XREAD`. This is in contrast to my original proposal, see https://github.com/valkey-io/valkey/pull/848/files#diff-fdef28e02ac8048541ebebbb829193ffd25622dc438bd8d921b167e51564bdd0R152
> * In cluster mode, a READONLY client seems to be blocked still (the already existing behavior of `clusterRedirectBlockedClientIfNeeded`)

I modified the code and test so that standalone and cluster now behave the same.

tests/integration/replica-redirect.tcl
```diff
@@ -93,12 +98,162 @@ start_server {tags {needs:repl external:skip}} {
...
+        } else {
+            fail "The failover does not happen"
+        }
+
+        wait_for_condition 1000 50 {
+            [catch {$rd read} err] &&
+            [string match "MOVED*" $err]
+        } else {
+            fail "Client didn't receive MOVED error"
+        }
```
zuiderkwast: `$rd read` will block until the command returns a reply, so there is no point to loop this in `wait_for_condition`. If it loops, it will only hang forever the second time.

```suggestion
        assert_error "MOVED *" {$rd read}
```

zuiderkwast (changes requested):
I revoke the approval for now, because the PR is still changing. :smile: I will approve soon, when the behavior is clear.

I think the current behavior looks good, but some test cases are missing and all the behavior changes are not described in the top comment. We need tests for these 6 cases:

1. Client blocked on key on old primary, cluster
2. Client with READONLY blocked on key on old primary, cluster
3. Client with READONLY blocked on key on old replica, cluster	
4. Client blocked on key on old primary, standalone
5. Client with READONLY blocked on key on old primary, standalone	
6. Client with READONLY blocked on key on old replica, standalone	

Also the top comment of the PR needs to clearly describe the behavior change for all these cases, especially the cases that are changed in this PR.

Please move the cluster tests to a new file `tests/unit/cluster/replica-redirect.tcl`.

There seems to be some confusion about the current behavior, so I created this draft PR to investigate the current behavior of these 6 cases: https://github.com/valkey-io/valkey/pull/2653

You can use it for some inspiration for the tests, to cover all 6 cases.

cjx-zar:
> I think the current behavior looks good, but some test cases are missing and all the behavior changes are not described in the top comment. We need tests for these 6 cases:
> 
> 1. Client blocked on key on old primary, cluster
> 2. Client with READONLY blocked on key on old primary, cluster
> 3. Client with READONLY blocked on key on old replica, cluster
> 4. Client blocked on key on old primary, standalone
> 5. Client with READONLY blocked on key on old primary, standalone
> 6. Client with READONLY blocked on key on old replica, standalone

Thanks for your draft. I have added the test for these 6 cases :)

ranshid:
I am fine with the code, However would like to clarify:
AFAIK, `READONLY` is rarely (if any) used by client libraries when standalone cluster is used. I think we will probably have no impact on real use cases. 

I think we might think to keep the unblocked error in case the command was xread, but we can wait to see if this is a real issue.

zuiderkwast:
@cjx-zar Thanks for adding all 6 test cases.

Now, please update the top comment of the PR to cover all the behavior changes. It doesn't mention the change for readonly clients on primaries.

gmbnomis (approved):

cjx-zar:
> @cjx-zar Thanks for adding all 6 test cases.
> 
> Now, please update the top comment of the PR to cover all the behavior changes. It doesn't mention the change for readonly clients on primaries.

Done.

tests/integration/failover.tcl
```diff
@@ -165,6 +181,17 @@ start_server {overrides {save {}}} {
...
             fail "Failover from node 2 to node 0 did not finish"
         }
+
+        assert_error "UNBLOCKED force unblock from blocking operation, instance state changed (master -> replica?)" {$rd_blocking read}
+
+        wait_for_condition 50 100 {
+            [s -2 blocked_clients] == 0
+        } else {
+            fail "rd_blocking client should still be unblocked"
+        }
```
zuiderkwast: Small simplification: We already checked above that the client returned UNBLOCKED, so it must be already unblocked on the server. No need to wait for it.

```suggestion
        assert_equal [s -2 blocked_clients] 0
```

tests/integration/failover.tcl
```diff
@@ -157,6 +167,12 @@ start_server {overrides {save {}}} {
         assert_match *slave* [$node_1 role]
         assert_match *slave* [$node_2 role]
 
+        wait_for_condition 50 100 {
+            [s -2 blocked_clients] == 1
+        } else {
+            fail "rd_blocking client should still be blocked in failover-in-progress"
+        }
```
zuiderkwast: Still blocked means we don't have to wait, right? Small simplification.

```suggestion
        # rd_blocking client should still be blocked in failover-in-progress
        assert_equal [s -2 blocked_clients] 1
```

zuiderkwast (approved):
Yes, very good. Thanks for your patience!

@ranshid I'm not worried about readonly clients blocking on XREAD on a primary continue to be blocked. Think about this scenario: The application has 100 readonly clients blocking on XREAD evenly distributed accross all nodes. After manual failover, all clients are still blocked. Still evenly distributed.

ranshid:
> Yes, very good. Thanks for your patience!
> 
> @ranshid I'm not worried about readonly clients blocking on XREAD on a primary continue to be blocked. Think about this scenario: The application has 100 readonly clients blocking on XREAD evenly distributed accross all nodes. After manual failover, all clients are still blocked. Still evenly distributed.

@zuiderkwast I am not worried about that as well. I mainly pointed out that the added code is ALSO protecting against READONLY client which are NOT supporting the REDIRECT capability. I think that in this case the READONLY command will simply never be used by client libraries and thus the fix will probably not mean anything for these cases (which I am fine with)

ranshid:
I will add this PR to the maintainers meeting adgenda today for final approval before we merge it.

zuiderkwast:
@ranshid We have discussed it enough already. :laughing:

PR #2330, opened 2025-07-08
Disable link-time optimization in 'noopt' target

xbasel:
lto slows down builds during development.


xbasel:
need to run few checks

ranshid (commented):
LGTM

zuiderkwast (commented):
We only use LTO when O3 is used. See line 23-33

```c
ifeq ($(OPTIMIZATION),-O3)
	ifeq (clang,$(CLANG))
		OPTIMIZATION+=-flto
	else
		OPTIMIZATION+=-flto=auto -ffat-lto-objects
	endif
endif
ifneq ($(OPTIMIZATION),-O0)
	OPTIMIZATION+=-fno-omit-frame-pointer
endif
```

PR #2331, opened 2025-07-08
Adding support for DumpSerializedValue API

cdorantes05:
The purpose of this API is for Modules to be able to access a key and retrieve its associated value in order to serialize that value. This allows modules to extract and store data in a format that can later be reconstructed. 

A use case is when building a CDC module, we need to generate the serialized version of any given key (of any data type). This allows for any specific key and its value(s) to be logged in the CDC message.

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -248,6 +248,7 @@ struct ValkeyModuleBlockedClient;
 typedef int (*ValkeyModuleCmdFunc)(ValkeyModuleCtx *ctx, void **argv, int argc);
 typedef int (*ValkeyModuleAuthCallback)(ValkeyModuleCtx *ctx, void *username, void *password, ValkeyModuleString **err);
 typedef void (*ValkeyModuleDisconnectFunc)(ValkeyModuleCtx *ctx, struct ValkeyModuleBlockedClient *bc);
+void createDumpPayload(rio *payload, robj *o, robj *key, int dbid);
 
 /* This struct holds the information about a command registered by a module.*/
 struct ValkeyModuleCommand {
@@ -7532,6 +7533,32 @@ ValkeyModuleString *VM_SaveDataTypeToString(ValkeyModuleCtx *ctx, void *data, co
     }
 }
 
+/**
+ * Serialize the value of a given key from the database of the module context's client.
+ * If the serialization fails for any reason (e.g. key doesn't exist, or rdbSaveObject() fails), return NULL
+ * Otherwise return ValkeyModuleString which contains the serialized value for the key.
+ */
+ValkeyModuleString *VM_DumpSerializedValue(ValkeyModuleCtx *ctx, ValkeyModuleString *key) {
+    if (!ctx || !key) {
+        return NULL;
+    }
+
+    int flags = LOOKUP_NOEFFECTS;
+    robj *value = lookupKeyReadWithFlags(ctx->client->db, key, flags);
+    if (!value) {
+        return NULL;
+    }
+    rio payload;
+    rioInitWithBuffer(&payload, sdsempty());
+    createDumpPayload(&payload, value, key, ctx->client->db->id);
+
+    // Create a ValkeyModuleString (robj*) with the result SDS and return it.
+    sds result = payload.io.buffer.ptr;
+    ValkeyModuleString *o = createObject(OBJ_STRING, result);
+    autoMemoryAdd(ctx, VALKEYMODULE_AM_STRING, o);
+    return o;
+}
+
 /* Returns the name of the key currently being processed. */
 const ValkeyModuleString *VM_GetKeyNameFromDigest(ValkeyModuleDigest *dig) {
     return dig->key;
@@ -14012,6 +14039,7 @@ void moduleRegisterCoreAPI(void) {
     REGISTER_API(SaveLongDouble);
     REGISTER_API(LoadLongDouble);
     REGISTER_API(SaveDataTypeToString);
+    REGISTER_API(DumpSerializedValue);
     REGISTER_API(LoadDataTypeFromString);
     REGISTER_API(LoadDataTypeFromStringEncver);
     REGISTER_API(EmitAOF);
--- a/tests/unit/moduleapi/datatype.tcl
+++ b/tests/unit/moduleapi/datatype.tcl
@@ -131,4 +131,24 @@ start_server {tags {"modules"}} {
 
         assert_equal 1 [llength $keys]
     }
+
+    test {DataType: Dump Serialized Value correctly serializes strings} {
+        r set key value
+        set serialized [r datatype.save_object key]
+        r restore newkey 0 $serialized
+        r get newkey
+    } {value}
+
+    test {DataType: Dump Serialized Value correctly serializes hashes} {
+        r hset myhash field1 myvalue
+        set serialized [r datatype.save_object myhash]
+        r restore newhash 0 $serialized
+        r hgetall newhash
+    } {field1 myvalue}
+
+    test {DataType: Dump Serialized Value handles nonexistent keys} {
+        r del nonexistent_key
+        catch {r datatype.save_object nonexistent_key} result
+        set result
+    } {Failed to serialize}
 }
--- a/src/valkeymodule.h
+++ b/src/valkeymodule.h
@@ -1473,6 +1473,8 @@ VALKEYMODULE_API void *(*ValkeyModule_LoadDataTypeFromStringEncver)(const Valkey
 VALKEYMODULE_API ValkeyModuleString *(*ValkeyModule_SaveDataTypeToString)(ValkeyModuleCtx *ctx,
                                                                           void *data,
                                                                           const ValkeyModuleType *mt)VALKEYMODULE_ATTR;
+VALKEYMODULE_API ValkeyModuleString *(*ValkeyModule_DumpSerializedValue) (ValkeyModuleCtx *ctx, 
+                                             ValkeyModuleString *key) VALKEYMODULE_ATTR;
 VALKEYMODULE_API void (*ValkeyModule_Log)(ValkeyModuleCtx *ctx, const char *level, const char *fmt, ...)
     VALKEYMODULE_ATTR VALKEYMODULE_ATTR_PRINTF(3, 4);
 VALKEYMODULE_API void (*ValkeyModule_LogIOError)(ValkeyModuleIO *io, const char *levelstr, const char *fmt, ...)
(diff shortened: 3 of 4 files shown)


src/module.c
```diff
@@ -7532,6 +7533,32 @@ ValkeyModuleString *VM_SaveDataTypeToString(ValkeyModuleCtx *ctx, void *data, co
     }
 }
 
+/**
+ * Serialize the value of a given key from the database of the module context's client.
+ * If the serialization fails for any reason (e.g. key doesn't exist, or rdbSaveObject() fails), return NULL
```
hpatro, line 7538: Wouldn't the module benefit from knowing the actual reason of failure?

hpatro: When does this operation `rdbSaveObject` fail ?

src/module.c
```diff
@@ -7532,6 +7533,32 @@ ValkeyModuleString *VM_SaveDataTypeToString(ValkeyModuleCtx *ctx, void *data, co
...
+        return NULL;
+    }
+    rio rdb;
+    rioInitWithBuffer(&rdb, sdsempty());
+    createDumpPayload(&rdb, value, key, ctx->client->db->id);
+
+    // Create a ValkeyModuleString (robj*) with the result SDS and return it.
+    sds result = rdb.io.buffer.ptr;
+    ValkeyModuleString *o = createObject(OBJ_STRING, result);
+    if (ctx != NULL) autoMemoryAdd(ctx, VALKEYMODULE_AM_STRING, o);
```
hpatro: on the entry of the method we check if ctx is NULL, we perform an early exit so we could avoid the NULL check here.

```suggestion
    autoMemoryAdd(ctx, VALKEYMODULE_AM_STRING, o);
```

src/module.c
```diff
@@ -7532,6 +7533,32 @@ ValkeyModuleString *VM_SaveDataTypeToString(ValkeyModuleCtx *ctx, void *data, co
...
+ValkeyModuleString *VM_SaveObjectToString(ValkeyModuleCtx *ctx, ValkeyModuleString *key) {
+    if (!ctx || !key) {
+        return NULL;
+    }
+
+    int flags = LOOKUP_NOTOUCH | LOOKUP_NONOTIFY | LOOKUP_NOEXPIRE | LOOKUP_NOEFFECTS;
+    robj *value = lookupKeyReadWithFlags(ctx->client->db, key, flags);
+    if (value == NULL) {
+        return NULL;
+    }
```
hpatro: nit: consistent with your code style

```suggestion
    if (!value) {
        return NULL;
    }
```

src/redismodule.h
```diff
@@ -362,6 +362,7 @@
 #define RedisModule_SelectDb ValkeyModule_SelectDb
 #define RedisModule_KeyExists ValkeyModule_KeyExists
 #define RedisModule_OpenKey ValkeyModule_OpenKey
+#define RedisModule_SaveObjectToString ValkeyModule_SaveObjectToString
```
hpatro: Do we keep adding this for new API(s) as well? @PingXie

hpatro (commented):
LGTM overall. 

This is a new module API, so I believe this requires voting @valkey-io/core-team. Please vote 👍 / 👎

src/redismodule.h
zuiderkwast: No, these are only for redis module API compat. We can add it if redis has it imo, but otherwise not.

zuiderkwast:
I'm fine with this API but I would like to know more about the use case(s). What module do you want to use it for and why?

src/module.c
```diff
@@ -7532,6 +7533,32 @@ ValkeyModuleString *VM_SaveDataTypeToString(ValkeyModuleCtx *ctx, void *data, co
     }
 }
 
+/**
+ * Serialize the value of a given key from the database of the module context's client.
+ * If the serialization fails for any reason (e.g. key doesn't exist, or rdbSaveObject() fails), return NULL
+ * Otherwise return ValkeyModuleString which contains the serialized value for the key.
+ */
+ValkeyModuleString *VM_SaveObjectToString(ValkeyModuleCtx *ctx, ValkeyModuleString *key) {
```
sungming2: Could we have a better name since it basically retruns serialized value?
e.g., VM_DumpSerializedValue

src/module.c
```diff
@@ -7532,6 +7533,32 @@ ValkeyModuleString *VM_SaveDataTypeToString(ValkeyModuleCtx *ctx, void *data, co
...
+    if (!ctx || !key) {
+        return NULL;
+    }
+
+    int flags = LOOKUP_NOTOUCH | LOOKUP_NONOTIFY | LOOKUP_NOEXPIRE | LOOKUP_NOEFFECTS;
+    robj *value = lookupKeyReadWithFlags(ctx->client->db, key, flags);
+    if (value == NULL) {
+        return NULL;
+    }
+    rio rdb;
```
sungming2: nit:

```suggestion
    rio payload;
```

tests/unit/moduleapi/datatype.tcl
```diff
@@ -131,4 +131,18 @@ start_server {tags {"modules"}} {
...
+        r restore newkey 0 $serialized
+        r get newkey
+    } {value}
+
+    test {DataType: Save Object to String correctly serializes hashes} {
+        r hset myhash field1 myvalue
+        set serialized [r datatype.save_object myhash]
+        r restore newhash 0 $serialized
+        r hgetall newhash
+    } {field1 myvalue}
```
sungming2: Could we also add a test for nonexistent key case

hpatro:
> I'm fine with this API but I would like to know more about the use case(s). What module do you want to use it for and why?

The team is exploring change data capture use case around Valkey. On a key space notification to the module, they would like to create a serialized dump of the value and pass it forward.

zuiderkwast:
Change data capture can be used for implementing custom replication and AOF-like backups, etc.?

If the new API returns a dump of a key in RDB format, it's very similar to the DUMP command? like VM_Call(DUMP)

hpatro:
> If the new API returns a dump of a key in RDB format, it's very similar to the DUMP command? like VM_Call(DUMP)

Yeah, basically the same. Talked to @QuChen88 about this yesterday. Do we plan on introducing more information to the module API or does `VM_Call` suffice?

src/module.c
```diff
@@ -7532,6 +7533,32 @@ ValkeyModuleString *VM_SaveDataTypeToString(ValkeyModuleCtx *ctx, void *data, co
...
+ * Serialize the value of a given key from the database of the module context's client.
+ * If the serialization fails for any reason (e.g. key doesn't exist, or rdbSaveObject() fails), return NULL
+ * Otherwise return ValkeyModuleString which contains the serialized value for the key.
+ */
+ValkeyModuleString *VM_DumpSerializedValue(ValkeyModuleCtx *ctx, ValkeyModuleString *key) {
+    if (!ctx || !key) {
+        return NULL;
+    }
+
+    int flags = LOOKUP_NOTOUCH | LOOKUP_NONOTIFY | LOOKUP_NOEXPIRE | LOOKUP_NOEFFECTS;
```
```suggestion
    int flags = LOOKUP_NOEFFECTS;
```
LOOKUP_NOEFFECTS implies the other three :)

src/module.c
madolson: rdbSaveObject can fail when writing out to network, but it never fails when pushing an object in-memory.

src/module.c
```diff
@@ -7532,6 +7533,32 @@ ValkeyModuleString *VM_SaveDataTypeToString(ValkeyModuleCtx *ctx, void *data, co
     }
 }
 
+/**
+ * Serialize the value of a given key from the database of the module context's client.
+ * If the serialization fails for any reason (e.g. key doesn't exist, or rdbSaveObject() fails), return NULL
+ * Otherwise return ValkeyModuleString which contains the serialized value for the key.
+ */
+ValkeyModuleString *VM_DumpSerializedValue(ValkeyModuleCtx *ctx, ValkeyModuleString *key) {
```
madolson, line 7541: I think the more correct API would be to operate on an already opened Key. The handling of the DB parameters here isn't quite right (it will always be DB zero). something like:
```
ValkeyModuleString *VM_DumpSerializedValue(ValkeyModuleKey *key);
```

madolson (commented):
I'm OK with the high level proposal, but think we should flow through the normal mechanism of opening a key instead of a new way to access keys. Otherwise I agree with others that the right way would be to do VM_Call + DUMP.

QuChen88:
Yes this module API is for change data capture where we can capture keyspace notifications and generate a message containing the information about key name / type of update / serialized value post modification, and making the changes available elsewhere in a different system. 

Doing it via `VM_Call()` with `DUMP` command requires constructing a fake client with the command and passing it through the full command processing flow. It carries a fair amount of overhead compared to the current route of opening the key in the module and getting the dump of its value directly.

src/module.c
hpatro: Thought so. We can simplify the comment to if NULL, the key doesn't exists.

zuiderkwast:
@QuChen88 yes VM_Call has a lot of overhead. It's just good to point out the equivalents and alternatives.

Maybe since this is equivalent to the DUMP command, should we call it VM_Dump?

What about VM_Restore, do we need it too? Dump/restore are used together and providing both maybe be more complete?

QuChen88:
@zuiderkwast For the API name, I think either `VM_DumpKey()` or `VM_DumpValue()` is fine. 

Even though `VM_Restore` is not required for CDC, I agree that we can create another module API `VM_RestoreKey()` as well. Maybe something like: `int VM_RestoreKey(ValkeyModuleCtx *ctx, ValkeyModuleString *key, ValkeyModuleString *serializedValue)`. The new key name and the serialized value are passed in as arguments. The method returns `VALKEYMODULE_OK` if success, and `VALKEYMODULE_ERR` if failure. 

A couple of questions here about the implement for `VM_RestoreKey()` due to the complexity of the regular RESTORE command:

- Can we return ERR if the key already exists? 
- Do we also need to pass in a TTL to create the key with like for regular RESTORE command?
- Do we also need to do `signalModifiedKey()` and `notifyKeyspaceEvent()` here?

zuiderkwast:
I have no strong opinion about the name. DumpSerializedValue is OK. It should take a ValkeyModuleKey though.

We can postpone Restore to a separate PR if it's non-trivial. (It too should operate on ValkeyNoduleKey probably.)

JimB123:
> Doing it via `VM_Call()` with `DUMP` command requires constructing a fake client with the command and passing it through the full command processing flow. It carries a fair amount of overhead compared to the current route of opening the key in the module and getting the dump of its value directly.

I am strongly against creating new module APIs as an alternative for calling existing commands.  `VM_Call` is designed expressly for modules to execute commands.  If the module API is inherently inefficient, we should address that inefficiency (if it exists) rather than creating new APIs, willy-nilly, for each command where we suspect that `VM_Call` is adding "too much" overhead.

In this case, in particular, as far as I can tell, there has been no measurements to show that `VM_Call` is introducing any significant overhead.  A change at this point would be (at best) pre-mature optimization.  At worst, this sets a precedent for introducing, documenting, and maintaining hundreds of new module API functions.

We need to address API changes with caution.

QuChen88:
I looked into the implementation of `VM_Call()`, it pretty much copies most of the validation logic from `processCommand()` before it reaches the actual command execution logic via `call()`. See  https://github.com/valkey-io/valkey/blob/unstable/src/module.c#L6336. I am not sure if that was by design or not. Maybe we can optimize out with a new API option to skip validation and directly get to `call()`?

I agree that in the meantime we can proceed with `VM_Call()` in the module to achieve the equivalent functionality and do some perf measurements later to understand how much overhead this is.

PR #2333, opened 2025-07-09
Fix DEBUG CLUSTERLINK KILL args check to avoid crash

enjoy-binbin:
Currently typing `debug clusterlink` will crash the server since
in here we are accessing `c->argv[2]->ptr`.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -6689,7 +6689,7 @@ int clusterNodeIsPrimary(clusterNode *n) {
 }
 
 int handleDebugClusterCommand(client *c) {
-    if (strcasecmp(c->argv[1]->ptr, "CLUSTERLINK") || strcasecmp(c->argv[2]->ptr, "KILL") || c->argc != 5) {
+    if (c->argc != 5 || strcasecmp(c->argv[1]->ptr, "CLUSTERLINK") || strcasecmp(c->argv[2]->ptr, "KILL")) {
         return 0;
     }
 


sarthakaggarwal97 (approved):

madolson (approved):

src/cluster_legacy.c
```diff
@@ -6689,7 +6689,7 @@ int clusterNodeIsPrimary(clusterNode *n) {
 }
 
 int handleDebugClusterCommand(client *c) {
-    if (strcasecmp(c->argv[1]->ptr, "CLUSTERLINK") || strcasecmp(c->argv[2]->ptr, "KILL") || c->argc != 5) {
+    if (c->argc != 5 || strcasecmp(c->argv[1]->ptr, "CLUSTERLINK") || strcasecmp(c->argv[2]->ptr, "KILL")) {
```
hwware, line 6692: I think this condition should be moved after Line 6696.  First check if cluster is enabled, then check the number of parameters.

hpatro (approved):

src/cluster_legacy.c
enjoy-binbin: no, we can not move this. Otherwise any debug nonexist-subcommand will throw the error.
```
    } else if (!strcasecmp(c->argv[1]->ptr, "client-enforce-reply-list") && c->argc == 3) {
        server.debug_client_enforce_reply_list = atoi(c->argv[2]->ptr);
        addReply(c, shared.ok);
    } else if (!handleDebugClusterCommand(c)) {
        return;
    }
```

hwware: Which error it will throw? If the subcommand is not in cluster mode, handleDebugClusterCommand will return 1.

enjoy-binbin: If we are doing this, in non-cluster mode, any debug nonexist-subcommand will throw `Debug option only available for cluster mode enabled setup` error
```
int handleDebugClusterCommand(client *c) {
    if (!server.cluster_enabled) {
        addReplyError(c, "Debug option only available for cluster mode enabled setup!");
        return 1;
    }
    
    if (strcasecmp(c->argv[1]->ptr, "CLUSTERLINK") || strcasecmp(c->argv[2]->ptr, "KILL") || c->argc != 5) {
        return 0;
    }
```

PR #2335, opened 2025-07-09
Exit early when all the covered slots are deleted

sarthakaggarwal97:
We do not need to iterate through all the slots while deleting the slots. We can short circuit the loop as soon as all the covered slots are deleted.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -5754,12 +5754,14 @@ int clusterDelSlot(int slot) {
 /* Delete all the slots associated with the specified node.
  * The number of deleted slots is returned. */
 int clusterDelNodeSlots(clusterNode *node) {
-    int deleted = 0, j;
+    int deleted = 0;
+    int remaining = node->numslots;
 
-    for (j = 0; j < CLUSTER_SLOTS; j++) {
+    for (int j = 0; j < CLUSTER_SLOTS && remaining > 0; j++) {
         if (clusterNodeCoversSlot(node, j)) {
             clusterDelSlot(j);
             deleted++;
+            remaining--;
         }
     }
     return deleted;
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -4883,7 +4883,6 @@ void clusterLogCantFailover(int reason) {
  * Note that it's up to the caller to be sure that the node got a new
  * configuration epoch already. */
 void clusterFailoverReplaceYourPrimary(void) {
-    int j;
     clusterNode *old_primary = myself->replicaof;
 
     if (clusterNodeIsPrimary(myself) || old_primary == NULL) return;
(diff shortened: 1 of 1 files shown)


zuiderkwast (commented):
It's simplicity vs performance. If this is important for performance, the we should do something that lets us skip looping in more cases. Otherwise, maybe keeping it simple might be more important. Which scenario is this?

An idea for a more optimal optimization: use the slot bitmap and `__builtin_clz` or `__builtin_ctz` to directly find the node's next slot. Then clear this bit and repeat.

hpatro (approved):
Sure. Do you want to update `clusterFailoverReplaceYourPrimary` as well?

hpatro:
> It's simplicity vs performance. If this is important for performance, the we should do something that lets us skip looping in more cases. Otherwise, maybe keeping it simple might be more important. Which scenario is this?
> 
> An idea for a more optimal optimization: use the slot bitmap and `__builtin_clz` or `__builtin_ctz` to directly find the node's next slot. Then clear this bit and repeat.

The change is not necessarily complex but I agree with your sentiment to not make changes which doesn't bring in a lot of benefit.

sarthakaggarwal97:
> It's simplicity vs performance.

I thought this is a simple change. If you feel that we can follow a complex approach to solve this, like the one you mentioned, I can take a stab at it. I am ok we feel we needn't touch this. I generally when I see some code, I tend to find to look for optimizations, that's where the change came from.

zuiderkwast:
Ok, i'm fine with merging it as is.

(I'm not so sure my idea is that complex though.)

sarthakaggarwal97:
@zuiderkwast you are right, will update the PR!

zuiderkwast:
Ok if you want. :) some inspiration: https://github.com/valkey-io/valkey/blob/unstable/src/hashtable.c#L715-L720

sarthakaggarwal97:
@zuiderkwast made some change. Let me know what you think.

src/cluster_legacy.c
```diff
@@ -5754,12 +5757,17 @@ int clusterDelSlot(int slot) {
...
-    int deleted = 0, j;
-
-    for (j = 0; j < CLUSTER_SLOTS; j++) {
-        if (clusterNodeCoversSlot(node, j)) {
-            clusterDelSlot(j);
-            deleted++;
+    int deleted = 0;
+    if (node->numslots == 0) return 0;
+
+    for (unsigned long i = 0; i < sizeof(node->slots); ++i) {
```
hpatro: Might as well name `i` as `byte`.

hpatro (commented):
Why did we remove the remaining logic altogether ?

src/cluster_legacy.c
sarthakaggarwal97: yep I was planning to. I thought there might be some other comments so I can push the change at once. Will do it.

hpatro (approved):

zuiderkwast (commented):
Ah, I didn't remember it's an array of char, so we have to loop over the 2048 bytes and only use ctz on each byte. Not a huge win then... But I'm ok with either solution.

hpatro:
> Ah, I didn't remember it's an array of char, so we have to loop over the 2048 bytes and only use ctz on each byte. Not a huge win then... But I'm ok with either solution.

Yeah, @sarthakaggarwal97 and I were discussing it yesterday. It's still 1/8 the no. of operation.

hpatro (approved):
LGTM

sarthakaggarwal97:
thanks @hpatro and @zuiderkwast for helping out on this one!

PR #2336, opened 2025-07-09
Add cross engine (Redis OSS) compatible test

hpatro:
Add cross engine test to ensure we don't introduce regression with older engine version compatibility. 

Engine version added: Redis OSS 6.2 / 7.0

### DIFF
--- a/.github/workflows/daily.yml
+++ b/.github/workflows/daily.yml
@@ -78,6 +78,24 @@ jobs:
       - name: unittest
         if: true && !contains(github.event.inputs.skiptests, 'unittest')
         run: ./src/valkey-unit-tests --accurate
+      - name: install Redis OSS 6.2 server for compatibility testing
+        run: |
+          cd tests/tmp
+          wget https://download.redis.io/releases/redis-6.2.14.tar.gz
+          tar -xvf redis-6.2.14.tar.gz
+      - name: backward compatibility tests with OSS Redis 6.2
+        run: |
+          sudo apt-get install tcl8.6 tclx
+          ./runtest --verbose --tags compatible-redis --dump-logs --other-server-path tests/tmp/redis-6.2.14/src/redis-server
+      - name: install Redis OSS 7.0 server for compatibility testing
+        run: |
+          cd tests/tmp
+          wget https://download.redis.io/releases/redis-7.0.15.tar.gz
+          tar -xvf redis-7.0.15.tar.gz
+      - name: backward compatibility tests with OSS Redis 7.0
+        run: |
+          sudo apt-get install tcl8.6 tclx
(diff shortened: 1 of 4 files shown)


hpatro:
Note: The existing test relies on `CLUSTER SHARDS` API for verifying the cluster state across nodes, so thought of adding this minimal test which uses `CLUSTER NODES` API which were introduced since the cluster feature launch.

hpatro:
I think it would be nice to keep testing against older Redis OSS versions so that we don't break compatibility accidentally. Redis still floats 6.2/7.0/~~7.2~~ OSS version binary. Should we add a CI for that? What are your thoughts about it?

zuiderkwast (approved):
Good. Yeah I like the idea of running against multiple versions.

To run these test against very old versions, we need to avoid recent commands like cluster shards, as you mentioned. To make sure we don't break this, we should run against the oldest version (6.2?) in CI (or Daily).

hpatro:
> Good. Yeah I like the idea of running against multiple versions.
> 
> To run these test against very old versions, we need to avoid recent commands like cluster shards, as you mentioned. To make sure we don't break this, we should run against the oldest version (6.2?) in CI (or Daily).

The other issue was with the minimal config file.`enable-debug-command` / `enable-module-command` configs are pretty recent and doesn't work. Yeah, daily should be reasonable.

enjoy-binbin (approved):
LGTM.

hpatro:
TIL: Redis OSS 7.2 doesn't exist

sarthakaggarwal97 (approved):

hpatro:
Successful run against OSS Redis 6.2/7.0 - https://github.com/hpatro/valkey/actions/runs/16204582629/job/45751625762

enjoy-binbin (approved):

.github/workflows/daily.yml
```diff
@@ -78,6 +78,32 @@ jobs:
...
+          wget https://download.redis.io/releases/redis-7.0.15.tar.gz
+          tar -xvf redis-7.0.15.tar.gz
+      - name: make Redis OSS 7.0 server
+        run: |
+          cd tests/tmp/redis-7.0.15
+          make
+      - name: backward compatibility tests with OSS Redis 7.0
+        run: |
+          sudo apt-get install tcl8.6 tclx
+          ./runtest --verbose --tags compatible-redis --dump-logs --other-server-path tests/tmp/redis-7.0.15/src/redis-server
```
zuiderkwast, line 106: Running all tests takes some time and only some test suites do cross-version testing. Should we run only those suites instead of all?

zuiderkwast (approved):

.github/workflows/daily.yml
hpatro: Yeah, the tag restricts to only run the selective tests.

zuiderkwast: Ah, i missed that. Perfect!

Issue #2112, opened 2025-05-21
[BUG] Use after free with lua-debug


ranshid:
```
=================================================================
==9998==ERROR: AddressSanitizer: heap-use-after-free on address 0x606000009f90 at pc 0x00000072b7a3 bp 0x7fff21143be0 sp 0x7fff21143bd0
READ of size 4 at 0x606000009f90 thread T0
#0 0x72b7a2 in connBlock src/socket.c:459
#1 0x74f7f2 in ldbStartSession lua/debug_lua.c:178
#2 0x628078 in evalGenericCommandWithDebugging src/eval.c:678
#3 0x62650d in evalCommand src/eval.c:531
#4 0x626528 in evalRoCommand src/eval.c:535
#5 0x48f1a9 in call src/server.c:3733
#6 0x4931e5 in processCommand src/server.c:4359
#7 0x4de07d in processCommandAndResetClient src/networking.c:3142
#8 0x4de7f9 in processInputBuffer src/networking.c:3270
#9 0x4df247 in readQueryFromClient src/networking.c:3375
#10 0x729303 in callHandler src/connhelpers.h:79
#11 0x72aaf7 in connSocketEventHandler src/socket.c:301
#12 0x462d13 in aeProcessEvents src/ae.c:486
#13 0x46349d in aeMain src/ae.c:543
#14 0x4a487b in main src/server.c:7212
#15 0x7f4e65b2c139 in __libc_start_main (/lib64/libc.so.6+0x21139)
#16 0x452029 in _start (src/valkey-server+0x452029)

0x606000009f90 is located 16 bytes inside of 64-byte region [0x606000009f80,0x606000009fc0)
freed by thread T0 here:
#0 0x7f4e668fa1e5 in __interceptor_free (/lib64/libasan.so.4+0xd81e5)
#1 0x4ad826 in zfree_internal src/zmalloc.c:400
#2 0x4ad85f in valkey_free src/zmalloc.c:415
#3 0x72994e in connSocketClose src/socket.c:155
#4 0x729185 in connClose src/connection.h:271
#5 0x729376 in callHandler src/connhelpers.h:82
#6 0x72aaf7 in connSocketEventHandler src/socket.c:301
#7 0x462d13 in aeProcessEvents src/ae.c:486
#8 0x46349d in aeMain src/ae.c:543
#9 0x4a487b in main src/server.c:7212
[...]

PR #2340, opened 2025-07-10
Valkey Fuzzer

uriyage:
### Overview
This PR adds a fuzzing capability to Valkey, allowing developers and users to stress test their Valkey deployments with randomly generated commands. The fuzzer is integrated with the existing valkey-benchmark tool, making it easy to use without requiring additional dependencies.

### Key Features
• **Command Generator**: Automatically generates Valkey commands by retrieving command information directly from the server
• **Two Fuzzing Modes**:
  - normal: Generates only valid commands, doesn't modify server configurations
  - aggressive: Includes malformed commands and allows CONFIG SET operations

• **Multi-threaded Testing**: Each client runs in a dedicated thread to maximize interaction between clients and enable testing of complicated scenarios

• **Integration with valkey-benchmark**: Uses the existing CLI interface

### Implementation Details
• Added new files:
  - `fuzzer_command_generator.h/c`: Dynamically generates valkey commands.
  - `fuzzer_client.c`: Orchestrate all the client threads, report test progress, and handle errors.

• Modified existing files:
  - valkey-benchmark.c: Added fuzzing mode options and integration

### Command Generation Approach
The fuzzer dynamically retrieves command information from the server, allowing it to adapt to different Valkey versions and custom modules. Since the command information generated from JSON files is sometimes limited, not all generated commands will be valid, but approximately 95% valid command generation is achieved.

It is important to generate valid commands to cover as much code path as possible and not just the invalid command/args path. The fuzzer prioritizes generating syntactically and semantically correct commands to ensure thorough testing of the server's core functionality, while still including a small percentage of invalid commands in `aggressive` mode to test error handling paths

#### Config modification
For CONFIG SET command, the situation is more complex as the server currently provides limited information through CONFIG GET *. Some hardcoded logic is implemented that will need to be modified in the future. Ideally, the server should provide self-inspection commands to retrieve config keys-values with their properties (enum values, modifiability status, etc.).

### Issue Detection
The fuzzer is designed to identify several types of issues:
• Server crashes
• Server memory corruptions / memory leaks(when compiled with ASAN)
• Server unresponsiveness
• Server malformed replies

For unresponsiveness detection, command timeout limits are implemented to ensure no command blocks for excessive periods. If a server doesn't respond within 30 seconds, the fuzzer signals that something is wrong.

### Proven Effectiveness
When running against the latest unstable version, the fuzzer has already identified several issues, demonstrating its effectiveness:
* https://github.com/valkey-io/valkey/issues/2111
* https://github.com/valkey-io/valkey/issues/2112
* https://github.com/valkey-io/valkey/pull/2109
* https://github.com/valkey-io/valkey/pull/2113
* https://github.com/valkey-io/valkey/pull/2108
* https://github.com/valkey-io/valkey/pull/2137
* https://github.com/valkey-io/valkey/issues/2106
* https://github.com/valkey-io/valkey/pull/2347
* https://github.com/valkey-io/valkey/pull/2973
* https://github.com/valkey-io/valkey/pull/2974
### How to Use
Run the fuzzer using the valkey-benchmark tool with the --fuzz flag:

```bash
# Basic usage (10000 commands 1000 commands per client, 10 clients)
./src/valkey-benchmark --fuzz -h 127.0.0.1 -p 6379 -n 10000 -c 10

# With aggressive fuzzing mode
./src/valkey-benchmark --fuzz --fuzz-level aggressive -h 127.0.0.1 -p 6379 -n 10000 -c 10

# With detailed logging
./src/valkey-benchmark --fuzz --fuzz-log-level debug -h 127.0.0.1 -p 6379 -n 10000 -c 10
```

The fuzzer supports existing valkey-benchmark options, including TLS and cluster mode configuration.


src/fuzzer_client.c
```diff
@@ -0,0 +1,814 @@
...
+        logMessage(LOG_DEBUG, "Command -> Reply: %s -> (array of %zu elements)", command, reply->elements);
+        break;
+    case VALKEY_REPLY_INTEGER:
+        logMessage(LOG_DEBUG, "Command -> Reply: %s -> %lld", command, reply->integer);
+        break;
+    case VALKEY_REPLY_NIL:
+        logMessage(LOG_DEBUG, "Command -> Reply: %s -> (nil)", command);
+        break;
+    default:
+        logMessage(LOG_DEBUG, "Command -> Reply: %s -> (unknown reply type)", command);
```
hpatro: nit: `Command -> Reply` feels redundant, we could remove it.

src/fuzzer_command_generator.h
```diff
@@ -0,0 +1,28 @@
...
+#define FUZZER_COMMAND_GENERATOR_H
+
+#include <valkey/valkey.h>
+#include "sds.h"
+
+typedef struct Command {
+    sds *argv;
+    int argc;
+    int size;
+} Command;
```
hpatro: Shall we prefix the various struct around these with Fuzzer* ?

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2015 @@
...
+
+#define MAX_DEFAULT_NUMKEYS 5
+#define MAX_NUM_PER_LUA 5 /* Maximum number of commands per Lua script */
+#define DEFAULT_INTEGER_MAX 50
+#define DEFAULT_DOUBLE_MIN 1.0
+#define DEFAULT_DOUBLE_MAX 50.0
+#define RANDOM_TIME_VARIANCE 60
+#define OPTIONAL_ARG_PROBABILITY 2 /* 1/2 probability */
+#define LUA_SCRIPT_PROBABILITY 100 /* 1/100 (1%) probability of generating a Lua script */
+#define MULTI_EXEC_PROBABILITY 50  /* 1/50 (2%) probability of generating a MULTI/EXEC block */
```
hpatro: neat!

src/fuzzer_client.c
```diff
@@ -0,0 +1,814 @@
...
+    sdsfree(cmd_str);
+}
+
+static void handleCommandTimeout(Command *cmd, time_t start_time) {
+    /* Timeout error */
+    time_t current_time = time(NULL);
+    double elapsed_time = difftime(current_time, start_time);
+
+    char cmd_buffer[256];
+    formatCommandToBuffer(cmd, cmd_buffer, sizeof(cmd_buffer));
```
hpatro: Might want to use sds. Should we truncate or just log the entire command ?

hpatro (commented):
Nice work @uriyage! Gave it a run locally, works pretty well. 

I had a question about the final report, do you manually analyze all the errors and spot the issues, what is your mechanism around it?

ranshid (commented):
@uriyage shell we also add a daily job to run the fuzzer?

hpatro:
> @uriyage shell we also add a daily job to run the fuzzer?

I also thought about this but unsure about the triage part?

src/fuzzer_client.c
uriyage: Thanks, Done.

src/fuzzer_command_generator.h
uriyage: Since they're only used within the fuzzer-specific files, I don't think it is required.

src/fuzzer_client.c
uriyage: Thanks, I changed it to use the printCommand function. I limit the size to 1K since some args can be very big.

uriyage:
> Nice work @uriyage! Gave it a run locally, works pretty well.
> 
> I had a question about the final report, do you manually analyze all the errors and spot the issues, what is your mechanism around it?

The errors can be divided into two groups: server-side-observable and client-side-observable.

For server-side issues, we can identify crashes, memory corruptions, or memory leaks. With client-side issues, 
we can observe timeouts (where the server is unexpectedly unresponsive) or malformed replies from the server 
that indicate problems with reply generation.

For the first two server-side issues, which are the most common, we can simply validate that after the fuzzer run the server didn't crash and validate that no memory issues were reported (when compiled with ASAN).

Client-side issues are more difficult to root cause and require manual work to understand what went wrong. We 
could potentially add in the future more server-side debug capabilities, such as having the server inspect its own output and
crash whenever it sends a malformed reply. We could also add some thread monitoring capability to identify when the server 
becomes unresponsive.

uriyage:
> @uriyage shell we also add a daily job to run the fuzzer?

Thanks, I added it to the TCL tests so it will run with all the variations we currently use for TCL testing. Currently, the test will be considered a failure only if the server crashes or becomes unresponsive after the fuzzer run

src/fuzzer_command_generator.h
madolson: If they are in the header, they should have a prefix.

uriyage: Done

madolson:
This isn't required for 9.0, but I would like us to try to get it merged after the 9.0 rc-1 goes out.

src/valkey-benchmark.c
```diff
@@ -1881,6 +1900,13 @@ int parseOptions(int argc, char **argv) {
         tls_usage,
         rdma_usage,
         " --mptcp            Enable an MPTCP connection.\n"
+        " --fuzz             Enable fuzzy mode to generate random commands.\n"
+        " --fuzz-level <mode> Set fuzzing mode: normal (default) or aggressive.\n"
+        "                    normal: Generates valid commands only, doesn't change server configs.\n"
+        "                    aggressive: Generates also malformed commands and allows CONFIG SET commands.\n"
```
rainsupreme: Might we want to add more modes in the future? If so I think it would make sense to call this `--fuzz-mode` and give these modes more descriptive names that are more closely related to what they do - perhaps `args` and `config-malformed` or something. It's hard to change this later because automation tends to be build around these features.

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+    CMD_NO_AUTH = (1ULL << 15),
+    CMD_NO_MANDATORY_KEYS = (1ULL << 19),
+    CMD_MODULE_GETKEYS = (1ULL << 21),
+    CMD_MODULE_NO_CLUSTER = (1ULL << 22),
+    CMD_NO_ASYNC_LOADING = (1ULL << 23),
+    CMD_NO_MULTI = (1ULL << 24),
+    CMD_MOVABLE_KEYS = (1ULL << 25),
+    CMD_ALLOW_BUSY = (1ULL << 26),
+    CMD_MODULE_GETCHANNELS = (1ULL << 27),
+} CommandFlag;
```
rainsupreme: Could there be some kind of automated way to identify new commands that are not yet supported by the fuzzer? Something like a enum length check?

src/valkey-benchmark.c
```diff
@@ -1881,6 +1900,13 @@ int parseOptions(int argc, char **argv) {
         tls_usage,
         rdma_usage,
         " --mptcp            Enable an MPTCP connection.\n"
+        " --fuzz             Enable fuzzy mode to generate random commands.\n"
```
rainsupreme: Should we warn people that this could cause undefined behavior and should never be run with production hardware/data/etc? It's obvious enough to us, yes, but maybe not to every human or AI that finds this

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+    free(client_ctx);
+    client_ctx = NULL;
+}
+
+/* Generates a random command or a Lua script with commands */
+FuzzerCommand *generateCmd(void) {
+    FuzzerCommand *cmd = allocCommand();
+
+    /* In aggressive mode, generate malformed commands 5% of the time */
+    if (client_ctx->fuzz_level == AGGRESSIVE && (rand() % 20 == 0)) {
```
rainsupreme: Define this probability at the top with the others?

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+        generateMalformedCommand(cmd);
+        return cmd;
+    }
+
+    if (client_ctx->in_multiexec) {
+        if (rand() % 5 == 0) {
```
rainsupreme: another hardcoded probability

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+    cmd->argc--;
+}
+
+static void addRandomArgument(FuzzerCommand *cmd) {
+    if (cmd->argc >= cmd->size) return;
+
+    sds new_arg = NULL;
+    switch (rand() % 3) {
+    case 0: /* Random string */
+        new_arg = sdsnew("random_value");
```
rainsupreme, line 1875: nit: not random

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+        return sdsnew(invalid_nums[rand() % 5]);
+    }
+    }
+}
+
+/* Corrupt an argument's content */
+static void corruptArgument(FuzzerCommand *cmd) {
+    int idx = (rand() % cmd->argc);
+    /* 70% chance to preserve command name */
+    if (idx == 0 && (rand() % 10 < 7)) return;
```
rainsupreme, line 1849: Would it be worth putting this probability at the top with the rest?

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+}
+
+/* Generate a random ACL rule */
+static void generateAclRule(FuzzerCommand *cmd, int r) {
+    /* Generate valid ACL rules dynamically using fetched categories */
+    int rule_type = r % 10;
+
+    switch (rule_type) {
+    case 0: /* User state rules */
+        appendArg(cmd, sdsnew(r % 2 ? "on" : "off"));
```
rainsupreme: arriving here, it's a given that r ends in 0 when written in decimal (r%10 == 0). It will always be even - thus it is impossible for r % 2 will never be 1.

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+        break;
+    case 1: /* Password rules */
+        if (r % 3 == 0) {
+            appendArg(cmd, sdsnew("nopass"));
+        } else {
+            appendArg(cmd, sdscatprintf(sdsempty(), ">pass%d", r % 100));
+        }
+        break;
+    case 2: /* Key pattern rules */
+        if (r % 4 == 0) {
```
rainsupreme: Similarly, the number written in decimal at this point ends in a 2 (r%10 == 2). r%4 cannot ever be 1 or 3 here, so those values are unreachable.

I don't know how many categories there are offhand for some of the cases below, but there are likely similar issues. I'd say use either `rand()` or `r/=10;` after getting rule_type, I think

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+            appendArg(cmd, sdsnew("-@dangerous"));
+        }
+        break;
+    }
+}
+
+/* Generate a random network address (IP or hostname) with port */
+static sds generateRandomAddress(int r) {
+    /* Generate a valid IP:port format */
+    int port = 10000 + (r % 55535); /* Port range 10000-65535 */
```
rainsupreme: Actually 10000-10999, because the caller initializes r with `int r = rand() % 1000;`

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+        int bits_idx = (r / 2) % (sizeof(bits) / sizeof(bits[0]));
+        appendArg(cmd, sdscatprintf(sdsempty(), "%s%d", signs[sign_idx], bits[bits_idx]));
+    } else if (strcmp(argName, "old-format") == 0) {
+        appendArg(cmd, sdscatprintf(sdsempty(), "192.168.%d.%d:%d", r % 256, (r / 256) % 256, 6379 + (r % 1000)));
+    } else if (strcmp(argName, "runid") == 0 || strcmp(argName, "replicationid") == 0) {
+        appendArg(cmd, sdscatprintf(sdsempty(), "%x%x%x%x%x%x%x%x%x%x%x%x%x%x%x%x",
+                                    r % 16, (r / 16) % 16, (r / 256) % 16, (r / 4096) % 16,
+                                    (r * 17) % 16, (r * 19) % 16, (r * 23) % 16, (r * 29) % 16,
+                                    (r * 31) % 16, (r * 37) % 16, (r * 41) % 16, (r * 43) % 16,
+                                    (r * 47) % 16, (r * 53) % 16, (r * 59) % 16, (r * 61) % 16));
```
rainsupreme: In this and similar places nearby, r is already in the range 0-999: `int r = rand() % 1000;`
- `(r / 256) % 16` will be 0-3, not 0-15
- `(r / 4096) % 16` will always be 0

This causes poor entropy in the other lines as well - the string will be less random than intended

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+    if (choice == 0) {
+        /* Negative infinity */
+        appendArg(cmd, sdsnew("-"));
+    } else if (choice == 1) {
+        /* Positive infinity */
+        appendArg(cmd, sdsnew("+"));
+    } else if (choice < 6) {
+        /* Inclusive range with a string value */
+        appendArg(cmd, sdscatprintf(sdsempty(), "[%s%d",
+                                    strcmp(argName, "min") == 0 ? "a" : "z", r % 100));
```
rainsupreme: there are 40 possible values here, not 100. It will be in range 0-99 AND last digit is in the range 2-5. (Similar below of course, 40 possible values)

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+    valkeyReply *commandInfo = NULL;
+    valkeyReply *configOutput = NULL;
+    valkeyReply *aclCatOutput = NULL;
+
+    /* Execute COMMAND DOCS to get command documentation */
+    commandDocs = valkeyCommand(ctx, "COMMAND DOCS");
+    if (!commandDocs || commandDocs->type == VALKEY_REPLY_ERROR) {
+        printf("Error: Failed to execute COMMAND DOCS. %s\n",
+               commandDocs ? commandDocs->str : "No reply received");
+        goto cleanup;
```
rainsupreme, line 1071: 🤨

j/k this is good - you used goto to keep things simple and clean 😌

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+    free(command->argv);
+    command->argv = NULL;
+}
+
+static void removeInvalidCommands(FuzzMode fuzz_level) {
+    /* Iterate over the registry and remove if args require and have subcommands */
+    for (size_t i = 0; i < fuzz_ctx->commandRegistrySize;) {
+        CommandEntry *cmd = &fuzz_ctx->commandRegistry[i];
+
+        if (shouldFilterCommand(cmd->fullname, fuzz_level) || (cmd->has_subcommands && cmd->arity == -2)) {
```
rainsupreme: This is the only place `shouldFilterCommand` is called - how about moving it closer?

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+        /* Convert to uppercase */
+        for (sds p = command->argv[1]; *p; p++) {
+            *p = toupper(*p);
+        }
+    }
+
+    /* Create fullname */
+    int fullnameLength = strlen(command->argv[0]) + 1;
+    if (subcommandName) {
+        fullnameLength += strlen(command->argv[1]) + 1;
```
rainsupreme: nit: a comment would be nice to note what the +1s are for

src/fuzzer_command_generator.c
```diff
@@ -0,0 +1,2032 @@
...
+    return strdup(key);
+}
+
+int configDictKeyCompare(const void *key1, const void *key2) {
+    return strcmp(key1, key2) == 0;
+}
+
+void configDictKeyDestructor(void *key) {
+    free(key);
+}
```
rainsupreme: delete - these are unused

rainsupreme (changes requested):
This is great! This seems quite thoroughly designed, and I like it a lot 😁

A few questions/suggestions:
1. Can we add unit tests? Or make AI do it? (actual UTs, in src/unit and written in C++)
2. How difficult is it to keep the fuzzer up to date if more commands or arguments are added in the future?
3. Could it potentially support fuzzing module commands in the future?

uriyage:
> This is great! This seems quite thoroughly designed, and I like it a lot 😁
> 
> A few questions/suggestions:
> 
> 1. Can we add unit tests? Or make AI do it? (actual UTs, in src/unit and written in C++)
> 2. How difficult is it to keep the fuzzer up to date if more commands or arguments are added in the future?
> 3. Could it potentially support fuzzing module commands in the future?

@rainsupreme 
1. Yes, I will try to do it as a follow up item.
2. The fuzzer dynamically retrieves commands and arguments from the server at runtime, so new commands are automatically tested as long as they're properly defined in the JSON specification files. Manual updates are only needed for edge cases like filtering dangerous commands or handling special argument patterns.
3. Yes, it already supports module commands. The fuzzer dynamically retrieves all commands via COMMAND DOCS, including module commands, so they're automatically tested as long as the module supplies proper documentation in the JSON

src/valkey-benchmark.c
uriyage: Good point, I changed it to allow multiple values per your suggestion

src/fuzzer_command_generator.c
uriyage: Added static assert and changed to include server.h flags

src/valkey-benchmark.c
uriyage: Added warning.

src/fuzzer_command_generator.c
uriyage: Done

src/fuzzer_command_generator.c
uriyage: Done

src/fuzzer_command_generator.c
uriyage: Good catch, fixed.

src/fuzzer_command_generator.c
uriyage: Done

src/fuzzer_command_generator.c
uriyage: Done

src/fuzzer_command_generator.c
uriyage: Done

src/fuzzer_command_generator.c
rainsupreme: Why not `int r = rand()`? I'm not sure why we're constraining it to the 0-999 range here, and I'm also not sure why we're passing this r value to other functions (where I commented about the misleadingly low entropy) instead of calling rand() again. 😕

rainsupreme (changes requested):
These are all good improvements! In fuzzer_command_generator.c around line 1382 I'd still like to see fixes for the reduced randomness/entropy caused by `int r = rand() % 1000;` and where it's passed into other functions instead of calling rand() again. (why?)

Perhaps this is for a future improvement, but do you think it'd be possible for a sequence of commands to be needed to reproduce a failure? If we want to consider that possibility then we'd want to be able to reproduce the sequence of commands that led to the failure. It seems you've been running the fuzzer for a while - have you encountered any failures you haven't been able to reproduce?

src/fuzzer_command_generator.c
uriyage: Thanks @rainsupreme ! I’ve updated the code to avoid reusing rand() results , as suggested.

Regarding reproducibility, simply knowing the sequence of commands is usually not enough to reproduce a failure. Since most bugs result from interaction between different clients commands which run on separate threads, the specific order of execution on the server is critical. But the server execution order depends on non-deterministic factors.
In my experience, given enough time, the specific sequence and timing that led to a crash/memory corruption tend to recur and easy to reproduce. A more effective aid for reproduction would maybe be a server-side logging option similar to AOF but for all commands that will capture the exact execution order of all the commands.

rainsupreme (approved):
LGTM!

tests/unit/fuzzer.tcl
```diff
@@ -0,0 +1,18 @@
+tags {"slow"} {
+    run_solo {fuzzer} {
```
ranshid, line 2: I feel this is the wrong way to run the fuzzer test.
I would prefer if we had a dedicated test workflow for the fuzzer tests so we can track issues with correlation to the fuzzer and not as part of some other kind of suite. For example, [this](https://github.com/valkey-io/valkey/actions/runs/20007259414/job/57371396141?pr=2340) failed on a probably correct issue, but failing as part of a code coverage workflow is probably not something we would like to have (+it will make the test coverage non-persistent).

ranshid:
@uriyage 
We have these fixes pending:
https://github.com/valkey-io/valkey/pull/2974
and
https://github.com/valkey-io/valkey/pull/2973

after which I hope the fuzzer is unblocked and we can merge it.
Please also open a followup issues about better filtering option (we discussed only run commands of specific groups...)

ranshid (approved):

ranshid:
I think it all looks good now. Merging.
Thank you @uriyage for this important contribution!

tests/unit/fuzzer.tcl
madolson: So what's the plan with this then?

ranshid: We discussed it. I was thinking at first that having a dedicated fuzzer run would be better, but then I thought that there are so many fixtures permutations it might be relevant to test the fuzzer with so I suggested we start like this and monitor the results. we can always move it to a dedicated suite with some supported fixtures

PR #2342, opened 2025-07-11
Add compatibility test with Valkey 7.2/8.0

hpatro:
* Add cross version compatibility test to run with Valkey 7.2 and 8.0
* Add mechanism in TCL test to skip tests dynamically - #2711


enjoy-binbin (approved):

hpatro:
Note: 

`Send cluster message via module from other server to latest` fails consistently with 8.1 -> latest. Need to debug further.

.github/workflows/ci.yml
```diff
@@ -21,14 +21,37 @@ jobs:
       - name: install old server for compatibility testing
         run: |
           cd tests/tmp
-          wget https://download.valkey.io/releases/valkey-7.2.7-noble-x86_64.tar.gz
-          tar -xvf valkey-7.2.7-noble-x86_64.tar.gz
-      - name: test
+          wget https://download.valkey.io/releases/valkey-7.2.10-noble-x86_64.tar.gz
```
madolson: Can we make this a matrix instead of having each individual test listed out? Would make it easier for adding more testing in the future.

zuiderkwast (approved):
The jobs look good. Maybe a matrix is good for de-duplication of yaml code.

When we bump the RDB version (#2422), there will be some incompatibilities, for example ~~downgrade~~ replication to an older replica will not work, unless the replica is 8.1 with the relaxed RDB version check enabled. Not sure if we need to care about this in this PR.

Perhaps we should instead add some conditional code in each test to skip tests that are not possible and/or not relevant depending on which other server version we're running against.

Edit: This test will only work if the other server >= 8.1, because it needs to have the relaxed RDB version check:

* https://github.com/valkey-io/valkey/blob/unstable/tests/unit/cluster/cross-version-cluster.tcl#L19

zuiderkwast:
@hpatro Will you merge this?

hpatro:
> @hpatro Will you merge this?

Will take a look at the comments next week and wrap it up.

madolson:
> Will take a look at the comments next week and wrap it up.

https://www.youtube.com/watch?v=NU9JoFKlaZ0

hpatro:
> > Will take a look at the comments next week and wrap it up.
> 
> 
> 
> https://www.youtube.com/watch?v=NU9JoFKlaZ0

Raft-land has sucked me in. This PR will be ready before the standup, I swear 😅

hpatro:
I'm seeing flaky test failure with 7.2/8.0 as the old replica node never gets established and there is no legacy replica as per the primary. Hence, the auto failover goes through.

```
19553:M 06 Oct 2025 21:32:33.606 - Client closed connection id=79 addr=127.0.0.1:38482 laddr=127.0.0.1:21114 fd=22 name= age=0 idle=0 flags=S capa= db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=20474 argv-mem=0 multi-mem=0 rbs=16384 rbp=0 obl=0 oll=0 omem=0 tot-mem=37504 events=r cmd=psync user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=153 tot-net-out=17 tot-cmds=4
19553:M 06 Oct 2025 21:32:33.606 * Connection with replica 127.0.0.1:21115 lost.
19553:M 06 Oct 2025 21:32:33.606 * Diskless rdb transfer, last replica dropped, killing fork child.
19553:M 06 Oct 2025 21:32:33.606 - Accepted 127.0.0.1:38490
19553:M 06 Oct 2025 21:32:33.606 * Replica 127.0.0.1:21115 asks for synchronization
19553:M 06 Oct 2025 21:32:33.606 * Full resync requested by replica 127.0.0.1:21115
19553:M 06 Oct 2025 21:32:33.606 * Current BGSAVE has socket target. Waiting for next BGSAVE for SYNC
19553:M 06 Oct 2025 21:32:33.702 * Background RDB transfer terminated with success
19553:M 06 Oct 2025 21:32:33.702 * Starting BGSAVE for SYNC with target: replicas sockets using: normal sync
19553:M 06 Oct 2025 21:32:33.702 * Background RDB transfer started by pid 19813 to pipe through parent process
19813:C 06 Oct 2025 21:32:33.702 * Fork CoW for RDB: current 0 MB, peak 0 MB, average 0 MB
19553:M 06 Oct 2025 21:32:33.702 * Diskless rdb transfer, done reading from pipe, 1 replicas still up.
19553:M 06 Oct 2025 21:32:33.707 - Client closed connection id=80 addr=127.0.0.1:38490 laddr=127.0.0.1:21114 fd=22 name= age=0 idle=0 flags=S capa= db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=20474 argv-mem=0 multi-mem=0 rbs=16384 rbp=0 obl=0 oll=0 omem=0 tot-mem=37504 events=r cmd=psync user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=153 tot-net-out=17 tot-cmds=4
19553:M 06 Oct 2025 21:32:33.707 * Connection with replica 127.0.0.1:21115 lost.
19553:M 06 Oct 2025 21:32:33.707 * Diskless rdb transfer, last replica dropped, killing fork child.
19553:M 06 Oct 2025 21:32:33.707 - Accepted 127.0.0.1:38496
19553:M 06 Oct 2025 21:32:33.707 - Reading from client: Connection reset by peer
19553:M 06 Oct 2025 21:32:33.803 * 1 of 1 replicas are in sync when shutting down.
19553:M 06 Oct 2025 21:32:33.803 # There is a child saving an .rdb. Killing it!
19553:M 06 Oct 2025 21:32:33.803 * Removing the pid file.
19553:M 06 Oct 2025 21:32:33.803 * Replica version: 16777215
19553:M 06 Oct 2025 21:32:33.803 * Perform auto failover to replica c5ba08789adb372b778788d255d9101a75925028 on shutdown.
19553:M 06 Oct 2025 21:32:33.803 * Saving the cluster configuration file before exiting.
19553:M 06 Oct 2025 21:32:33.808 * Removing the unix socket file.
19553:M 06 Oct 2025 21:32:33.808 # Valkey is now ready to exit, bye bye...
```

hpatro:
With the change introduced in https://github.com/valkey-io/valkey/pull/2422, we have updated the magic to be `VALKEY*` and this magic is not handled by [Valkey 7.2.x](https://github.com/valkey-io/valkey/blob/97b6663c775193cce6e41c752e7683f000f80ad5/src/rdb.c#L3037-L3048)  and [Valkey 8.0.x](https://github.com/valkey-io/valkey/blob/1cac48f92dc938f292480ea229877a18b001bb46/src/rdb.c#L3025-L3035). However, with [Valkey 8.1.x](https://github.com/valkey-io/valkey/blob/8eb91b321d04862247c006b68f1f42917cc34c22/src/rdb.c#L3041-L3050)  VALKEY magic is handled appropriately.

This leads to constant full sync attempt and replica running on 7.2.x/8.0.x fails to sync with a new primary. I guess the assumption we made is never to sync with a primary running on higher version as it's not backward compatible. So, the cross-version-cluster test needs to be updated accordingly.

hpatro:
We don't have a mechanism currently to run cross version test against a specific version. Need to find a way to skip `Cross version cluster - failover` test against 7.2/8.0.

hpatro:
After thinking a bit more on the test `Cross version cluster - failover` in `cross-version-cluster.tcl`. I believe it's incorrect to assume that a primary node running on latest engine version would function correctly with a replica node running on older version. Hence, I feel we should clean up this test. 

@zuiderkwast / @enjoy-binbin any thoughts on this ?

zuiderkwast:
Replica older than primary only works if both have the same RDB version or if the old version is 8.1 with the relaxed RDB version check.

Downgrade from 9.0 to 8.0 and below doesn't work and we need to skip this test for this combination.

... as mentioned earlier in this comment above: https://github.com/valkey-io/valkey/pull/2342#pullrequestreview-3107419828

hpatro:
> Replica older than primary only works if both have the same RDB version or if the old version is 8.1 with the relaxed RDB version check.
> 
> Downgrade from 9.0 to 8.0 and below doesn't work and we need to skip this test for this combination.

Yes, we need to add that mechanism to our cross version testing. currently, the server boot up would fail with the unsupported config. If there are no issue with config, the sync would still fail.

tests/unit/cluster/cross-version-cluster.tcl
```diff
@@ -17,15 +17,23 @@ tags {external:skip needs:other-server cluster singledb} {
...
-                wait_for_cluster_state "ok"
-
-                # Make sure the primary won't do the auto-failover.
-                catch {$primary shutdown nosave failover}
-                verify_log_message -1 "*Unable to perform auto failover on shutdown since there are legacy replicas*" 0
+                set res [dict get [r hello] version]
+                if {[regexp {([0-9]+)\.([0-9]+)\.[0-9]+} $res -> major minor]} {
+                    if {($major < 8) || ($major == 8 && $minor < 1)} {
+                        puts "Skipping test: requires Valkey 8.1 or above"
+                    }
```
zuiderkwast: Do you think we should use tags or something similar for skipping tests based on other-server version? The benefit is we can count the skipped tests in `::num_skipped`, similar to how we do for tests that are skipped based on tags.

A slightly similar test framework feature is the ongoing PR for skipping IPv6 tests if the machine doesn't support IPv6: #2674

zuiderkwast: I suppose the current logic for skipping would be OK for now though, if we want to include this in 9.0.

How does this printout look when it's running? This line gets printed among other printouts? When multiple tests are running in parallel, I guess it's hard to see which test this refers to.

zuiderkwast: I found it: https://github.com/valkey-io/valkey/actions/runs/18358356672/job/52296064651?pr=2342#step:5:3796

It gets printed among other printouts:

```
[ok]: HSETEX EX single field expires leaving other fields intact (1113 ms)
[ok]: List quicklist -> listpack encoding conversion (229 ms)
=== () Starting server on 127.0.0.1:21118 ok
[ok]: List encoding conversion when RDB loading (14 ms)
Skipping test: requires Valkey 8.1 or above
[ok]: MIGRATE timeout actually works (774 ms)
[ok]: List invalid list-max-listpack-size config (58 ms)
[ok]: List of various encodings (102 ms)
[ok]: List of various encodings - sanitize dump (7 ms)
```

It'd be ideal to print it like we do for ignored/skipped tests, like

    [ignore]: EVAL - JSON string encoding a string larger than 2GB: large memory flag not provided

.github/workflows/ci.yml
```diff
@@ -20,34 +20,62 @@ permissions:
...
+      - name: Run compatibility tests against ${{ matrix.server.version }}
         run: |
-          sudo apt-get install tcl8.6 tclx
-          ./runtest --verbose --tags -slow --dump-logs --other-server-path tests/tmp/valkey-8.1.3-noble-x86_64/bin/valkey-server
-      - name: module api test
-        run: CFLAGS='-Werror' ./runtest-moduleapi --verbose --dump-logs --other-server-path tests/tmp/valkey-8.1.3-noble-x86_64/bin/valkey-server
-      - name: validate commands.def up to date
+          sudo apt-get install -y tcl8.6 tclx
+          ./runtest --verbose --tags -slow --dump-logs \
+            --other-server-path tests/tmp/valkey-${{ matrix.server.version }}-noble-x86_64/bin/valkey-server
```
zuiderkwast: It looks like we're running a lot of tests that don't depend on the other-server, so we get a lot of duplicate test runs.

Should we run only some specific cross-version test suites using the `--single` parameter?

hpatro: This will be a lot of pain to maintain tbh. Things will get merged in without being added here.

tests/unit/cluster/cross-version-cluster.tcl
hpatro: I did think about it but couldn't find an easy way to skip it (actually). Currently it just succeeds.

.github/workflows/ci.yml
zuiderkwast: Yes, I get the concern. But we're also trying to limiting the number of full test runs in the CI scope though, to make it finish faster and not choke github too much i guess. If we want to run the full tests 3 times with this matrix, then maybe we should move it to Daily instead? We'll also be adding more versions to this matrix over time.

hpatro: I think we can just add the tag other-server here and we're good. Let me try that. Most of the other test should just skip then.

hpatro: With the tag it takes around ~2m of resource time compared to ~20m earlier. Thanks.

.github/workflows/ci.yml
```diff
@@ -20,34 +20,62 @@ permissions:
   contents: read
 
 jobs:
-  test-ubuntu-latest:
+  test-ubuntu-latest-compatibility:
```
zuiderkwast, line 47: Skipping everything that doesn't have the tag `needs:other-server` makes the run very fast indeed.

Now, we need to keep one runtest that covers all the tests, the old `test-ubuntu-latest`.

hpatro: Ah, we had it all together. Thanks for catching it.

tests/unit/cluster/cross-version-cluster.tcl
zuiderkwast: OK, let's keep a simple `if` for now. I see you opened an issue for it.

I see now that this if-else is wrongly nested. The `else` below where the test code actually runs is the else matching `if {[regexp {([0-9]+)\.([0-9]+)\.[0-9]+} $res -> major minor]} {`, i.e. it runs if the version doesn't match the regexp, i.e. never...? The actual version check `if {($major < 8) || ($major == 8 && $minor < 1)}` has no else clause.

zuiderkwast: We could do like this instead?

```suggestion
                set res [dict get [r hello] version]
                assert [regexp {([0-9]+)\.([0-9]+)\.[0-9]+} $res -> major minor]
                if {($major < 8) || ($major == 8 && $minor < 1)} {
                    puts "Skipping test: requires Valkey 8.1 or above"
                }
```

tests/unit/cluster/cross-version-cluster.tcl
```diff
@@ -16,16 +16,23 @@ tags {external:skip needs:other-server cluster singledb} {
...
+                    # Add a replica of the old version to the cluster
+                    r cluster meet $primary_host $primary_port
+                    wait_for_cluster_propagation
+                    r cluster replicate $primary_id
+                    wait_for_cluster_state "ok"
+
+                    # Make sure the primary won't do the auto-failover.
+                    catch {$primary shutdown nosave failover}
+                    verify_log_message -1 "*Unable to perform auto failover on shutdown since there are legacy replicas*" 0
+                }
```
zuiderkwast: Technically, `skip` raises an exception, so we don't need to wrap this in an `else` block and the extra indentation.

```suggestion
                }
                r config set rdb-version-check relaxed
                # Add a replica of the old version to the cluster
                r cluster meet $primary_host $primary_port
                wait_for_cluster_propagation
                r cluster replicate $primary_id
                wait_for_cluster_state "ok"

                # Make sure the primary won't do the auto-failover.
                catch {$primary shutdown nosave failover}
                verify_log_message -1 "*Unable to perform auto failover on shutdown since there are legacy replicas*" 0
```

zuiderkwast (approved):

hpatro:
Thanks @zuiderkwast for the suggestions!

zuiderkwast:
The `skip` causes dump logs, when running with `--dump-logs`.

I tried `./runtest --other-server-path misc/valkey-server-7 --single tests/unit/cluster/cross-version-cluster.tcl --dump-logs`

I get `[skip]: Cross version cluster - failover in tests/unit/cluster/cross-version-cluster.tcl: Requires Valkey 8.1 or above` but also the logs of both servers are dumped on stdout.

The reason is the "skipped" exception is caught in the start_server proc in `tests/support/server.tcl`. This is where the `$::dump_logs` variable is checked.

PR #2345, opened 2025-07-13
update build-debian-old to use Bullseye instead of EOL buster

ranshid:


madolson (approved):
I wonder if we should consider doing just the `debian:latest`. Anyways, this makes sense.

PR #2347, opened 2025-07-13
Prevent bad memory access when NOTOUCH client gets unblocked

uriyage:
Fix missing check for executing client in `lookupKey` function

### Issue
The `lookupKey` function in db.c accesses `server.executing_client->cmd->proc` without first verifying that `server.executing_client` is not NULL. This was introduced in #1499 where the check for executing client was added without verifying it could be null.

The server crashes with a null pointer dereference when the current_client's flag.no_touch is set.
```
27719 valkey-server *
/lib64/libpthread.so.0(+0x118e0)[0x7f34cb96a8e0]
src/valkey-server 127.0.0.1:21113(lookupKey+0xf5)[0x4a14b7]
src/valkey-server 127.0.0.1:21113(lookupKeyReadWithFlags+0x50)[0x4a15fc]
src/valkey-server 127.0.0.1:21113[0x52b8f1]
src/valkey-server 127.0.0.1:21113(handleClientsBlockedOnKeys+0xa5)[0x52b16f]
src/valkey-server 127.0.0.1:21113(processCommand+0xf1e)[0x4712c9]
src/valkey-server 127.0.0.1:21113(processCommandAndResetClient+0x35)[0x490fd5]
src/valkey-server 127.0.0.1:21113(processInputBuffer+0xe1)[0x4912e5]
src/valkey-server 127.0.0.1:21113(readQueryFromClient+0x8c)[0x49177b]
src/valkey-server 127.0.0.1:21113[0x57daa6]
src/valkey-server 127.0.0.1:21113[0x57e280]
src/valkey-server 127.0.0.1:21113(aeProcessEvents+0x261)[0x45b259]
src/valkey-server 127.0.0.1:21113(aeMain+0x2a)[0x45b450]
src/valkey-server 127.0.0.1:21113(main+0xd43)[0x479bf6]
/lib64/libc.so.6(__libc_start_main+0xea)[0x7f34cb5cd13a]
src/valkey-server 127.0.0.1:21113(_start+0x2a)[0x454e3a]
```

### Fix
Added a null check for `server.executing_client` before attempting to dereference it:

### Tests
Added a regression test in tests/unit/type/list.tcl.


src/db.c
```diff
@@ -126,7 +126,7 @@ robj *lookupKey(serverDb *db, robj *key, int flags) {
          * Don't do it if we have a saving child, as this will trigger
          * a copy on write madness. */
         if (server.current_client && server.current_client->flag.no_touch &&
-            server.executing_client->cmd->proc != touchCommand)
+            server.executing_client && server.executing_client->cmd->proc != touchCommand)
             flags |= LOOKUP_NOTOUCH;
```
PingXie: nit - check if `flags` already carries `LOOKUP_NOTOUCH`

```suggestion
        if ((flags & LOOKUP_NOTOUCH) == 0 &&
            server.current_client && server.current_client->flag.no_touch &&
            server.executing_client && server.executing_client->cmd->proc != touchCommand) {
            flags |= LOOKUP_NOTOUCH;
        }
```

PingXie (approved):
Thanks @uriyage

src/db.c
```diff
@@ -126,7 +126,7 @@ robj *lookupKey(serverDb *db, robj *key, int flags) {
          * Don't do it if we have a saving child, as this will trigger
          * a copy on write madness. */
         if (server.current_client && server.current_client->flag.no_touch &&
-            server.executing_client->cmd->proc != touchCommand)
+            server.executing_client && server.executing_client->cmd->proc != touchCommand)
```
gmbnomis, line 120: In #1499, I originally proposed to fall back to the `current_client` if the `executing_client` was NULL. I wonder if that's better as the `current_client` was used in this check before #1499.

That being said, I think it is odd that the `executing_client` is not set when we are re-processing an unblocked blocking command. @ranshid Do you know whether this is intentional?

gmbnomis: I had a look and this crash happens outside of the `call` command execution in `handleClientsBlockedOnKeys`. So, not having an `executing_client` is fine in this case and I think we should not fallback to the `current_client`.

Sorry for the noise.

PingXie: 'executing_client' is valid inside of 'call' only. The unblocking of a blocked client occurs after the running client finishes its command (that triggers the unblocking). This particular lookup is actually done with "NOTOUCH".

gmbnomis (approved):

src/db.c
ranshid: Yes. the executing_client was an evolution of the different client types we once had like script_caller. before that we had many cases were we had to change the current_client many times during a flow of operations. BTW, It is valid even outside of "call' flows in cases were we execute commands outside of processing context (eg AOF load)

ranshid (approved):

enjoy-binbin (approved):

zuiderkwast:
Backport to 8.1 since #1499 was released in 8.1, right?

zuiderkwast:
For the release notes, the title should mention when it would happen, e.g. no-touch + blocking commands.

ranshid:
Oooops - seems like a fix for that was also inserted as part of https://github.com/valkey-io/valkey/pull/2089 - I recall we had this issue which was blocking us during fuzzer testing and we just fixed it in order to progress and probably missed.
I still think this PR adds more tests and should be used. @uriyage do you want to drive it to the last mile?

src/db.c
```diff
@@ -115,8 +115,8 @@ robj *lookupKey(serverDb *db, robj *key, int flags) {
         /* Update the access time for the ageing algorithm.
          * Don't do it if we have a saving child, as this will trigger
          * a copy on write madness. */
-        if (server.current_client && server.current_client->flag.no_touch && server.executing_client &&
-            server.executing_client->cmd->proc != touchCommand)
+        if (server.current_client && server.current_client->flag.no_touch &&
```
ranshid: reapply @PingXie suggestion
```suggestion
        if ((flags & LOOKUP_NOTOUCH) == 0 &&
            server.current_client && server.current_client->flag.no_touch &&
```

zuiderkwast:
If we want to backport it, we need to re-apply the actual fix too, i.e. the first commit in this PR.

ranshid:
> If we want to backport it, we need to re-apply the actual fix too, i.e. the first commit in this PR.

It still has the fix as part of the commit. I think we can cherry-pick it as is

PR #2355, opened 2025-07-14
[TEST] Fix CI to run only other server tests for compatibility

hpatro:


hpatro:
@zuiderkwast Your comment https://github.com/valkey-io/valkey/pull/2336#discussion_r2203129643 is relevant for the CI flow for compatibility test which runs bunch of unnecessary tests.

tests/unit/cluster/failover2.tcl
```diff
@@ -102,7 +102,7 @@ start_cluster 7 3 {tags {external:skip cluster} overrides {cluster-ping-interval
 } ;# start_cluster
 
 run_solo {cluster} {
-    start_cluster 32 15 {tags {external:skip cluster} overrides {cluster-ping-interval 1000 cluster-node-timeout 15000}} {
```
hpatro, line 105: This test takes longer than 30 seconds.

.github/workflows/ci.yml
```diff
@@ -26,9 +26,9 @@ jobs:
       - name: test
         run: |
           sudo apt-get install tcl8.6 tclx
-          ./runtest --verbose --tags -slow --dump-logs --other-server-path tests/tmp/valkey-7.2.7-noble-x86_64/bin/valkey-server
+          ./runtest --verbose --tags "-slow needs:other-server" --dump-logs --other-server-path tests/tmp/valkey-7.2.7-noble-x86_64/bin/valkey-server
```
zuiderkwast, line 29: This is `test-ubuntu-latest`. I guess it should run not only the cross-compat tests...

hpatro: Ah my bad. I think I misread it to be following similar approach to daily.yml i.e. test cross server compatibility separately.

PR #2309, opened 2025-07-04
Database-level access control

dvkashapov:
## API changes and user behavior:

- [x] Default behavior for database access.

Default is `alldbs` permissions.

### Database Permissions (`db=`)
- [x] Accessing particular database

```
> ACL SETUSER test1 on +@all ~* resetdbs db=0,1 nopass
"user test1 on nopass sanitize-payload ~* resetchannels db=0,1 +@all"
```

- [x] (Same behavior without usage of `resetdbs`)
```
> ACL SETUSER test1 on +@all ~* db=0,1 nopass
"user test1 on nopass sanitize-payload ~* resetchannels db=0,1 +@all"
```

- [x] Multiple selector can be provided
```
> ACL SETUSER test1 on nopass (db=0,1 +@write +select ~*) (db=2,3 +@read +select ~*)
"user test1 on nopass sanitize-payload resetchannels alldbs -@all (~* resetchannels db=0,1 -@all +@write +select) (~* resetchannels db=2,3 -@all +@read +select)"
```

- [x] Restricting special commands which access databases as part of the command. 

The user needs to have access to both the commands and db(s) part of the command to run these commands.

1. SWAPDB
2. SELECT
3. MOVE - (Select command would have went through for the source database). Have access for the target database.
4. COPY

- [x] Restricting special commands which doesn't specify database number, however, accesses multiple databases.

The user needs to have access to both the commands and all databases (`alldbs`) to run these commands.

1. FLUSHALL - Access all databases
2. CLUSTER commands that access all databases:
    - CANCELSLOTMIGRATIONS
    - MIGRATESLOTS

- [x] New connection establishment behavior
New client connection gets established to DB 0 by default. Authentication and authorisation are decoupled and the user can connect/authenticate and further perform `SELECT` or other operation that do not access keyspace.

(Do we want to extend HELLO?) Alternative suggestion by @madolson: Extend `HELLO` command to pass the dbid to which the user should get connected after authentication if they have right set of permission. I think it will become a long poll for adoption.

- [x] Observability
Extend `ACL LOG` to log user which received denied permission error while accessing a database.

- [x] Module API
* Introduce module API `int VM_ACLCheckPermissions(ValkeyModuleUser *user, ValkeyModuleString **argv, int argc, int dbid, ValkeyModuleACLLogEntryReason *denial_reason);` 
* Stop support of `VM_ACLCheckCommandPermissions()`.

Resolves: #1336

PR #2357, opened 2025-07-15
valkey-cli: Fix dbid reset after error in MULTI

dvkashapov:
I was testing #2309 and found that if you do `MULTI, SELECT incorrect dbid and EXEC`  combination of commands, then database ID right next to valkey-server address won't change, but it should. Turns out, for this combination of commands reply->type is `VALKEY_REPLY_ARRAY`, which was added to if statement for EXEC handling.
```
BEFORE:
127.0.0.1:6379> multi
OK
127.0.0.1:6379(TX)> select 438764
QUEUED
127.0.0.1:6379[438764](TX)> exec
1) (error) ERR DB index is out of range
127.0.0.1:6379[438764]> multi
OK
127.0.0.1:6379[438764](TX)> select 327648273645872634
QUEUED
127.0.0.1:6379[1911465466](TX)> exec
1) (error) ERR value is out of range, value must between -2147483648 and 2147483647

AFTER:
127.0.0.1:6379> multi
OK
127.0.0.1:6379(TX)> select 7463746
QUEUED
127.0.0.1:6379[7463746](TX)> exec
1) (error) ERR DB index is out of range
127.0.0.1:6379>
             /\
No incorrect DB ID
```

### DIFF
--- a/src/valkey-cli.c
+++ b/src/valkey-cli.c
@@ -2479,7 +2479,8 @@ static int cliSendCommand(int argc, char **argv, long repeat) {
                 cliRefreshPrompt();
             } else if (!strcasecmp(command, "exec") && argc == 1 && config.in_multi) {
                 config.in_multi = 0;
-                if (config.last_cmd_type == VALKEY_REPLY_ERROR || config.last_cmd_type == VALKEY_REPLY_NIL) {
+                if (config.last_cmd_type == VALKEY_REPLY_ERROR || config.last_cmd_type == VALKEY_REPLY_NIL ||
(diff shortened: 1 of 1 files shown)


roshkhatri (approved):
LGTM

dvkashapov:
@roshkhatri I've thought a bit, and this clears dbid indicator even for valid `MULTI, SELECT id, EXEC`, when it should not, we need to investigate why reply->type for it is `VALKEY_REPLY_ARRAY`. This is not a correct fix, closing.

PR #2358, opened 2025-07-15
Fix module context object re-usage in scripting engines

rjd15372:
This commit refactors the scripting engine to support multiple cached module contexts per engine, rather than relying on a single cached `ValkeyModuleCtx` object.

Previously, having only one cached context object caused data races over the state stored in the context object, because it's possible that a script that is running for a long time to yield and the server event loop may call the `scriptingEngineCallGetMemoryInfo` function to get the scripting engine memory information, which re-uses the same cached context object. Another possible data-race is caused by the asynchronous scripts flush, which calls the `scriptingEngineCallFreeFunction` function in an background thread, and also re-uses the cached context object.

To address this, a cache array of module contexts was introduced in the scripting engine structure, with each slot dedicated to a specific use case—such as script execution, memory info queries, or function freeing.

### DIFF
--- a/src/scripting_engine.c
+++ b/src/scripting_engine.c
@@ -8,6 +8,25 @@
 #include "dict.h"
 #include "functions.h"
 #include "module.h"
+#include "server.h"
+#include "valkeymodule.h"
+
+/* Module context object cache size is set to 3 because at each moment there can
+ * be at most 3 module contexts in use by the scripting engine.
+ *
+ * 1. The module context used by the scripting engine to run the script;
+ * 2. The module context used by `scriptingEngineCallGetMemoryInfo` that is
+ *    periodically called by the server cron; and
+ * 3. The module context used by `scriptingEngineCallFreeFunction` that is
+ *    called when the server needs to reset the evaluation environment in the
+ *    asynchronous mode.
+ */
+enum {
+    COMMON_MODULE_CTX_INDEX = 0,        /* Common module context used by the scripting engine. */
+    GET_MEMORY_MODULE_CTX_INDEX = 1,    /* Module context used by `scriptingEngineCallGetMemoryInfo`. */
+    FREE_FUNCTION_MODULE_CTX_INDEX = 2, /* Module context used by `scriptingEngineCallFreeFunction`. */
+    _MODULE_CTX_CACHE_SIZE = 3          /* Total number of module contexts in the cache. */
+};
 
 typedef struct scriptingEngineImpl {
     /* Engine specific context */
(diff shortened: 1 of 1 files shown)


madolson:
This theoretically should be backported, but we probably won't since the impact doesn't seem that high. This is not required for RC1 but should be merged before 9.0.

src/scripting_engine.c
```diff
@@ -8,6 +8,25 @@
...
+ *    periodically called by the server cron; and
+ * 3. The module context used by `scriptingEngineCallFreeFunction` that is
+ *    called when the server needs to reset the evaluation environment in the
+ *    asynchronous mode.
+ */
+enum {
+    COMMON_MODULE_CTX_INDEX = 0,        /* Common module context used by the scripting engine. */
+    GET_MEMORY_MODULE_CTX_INDEX = 1,    /* Module context used by `scriptingEngineCallGetMemoryInfo`. */
+    FREE_FUNCTION_MODULE_CTX_INDEX = 2, /* Module context used by `scriptingEngineCallFreeFunction`. */
+    _MODULE_CTX_CACHE_SIZE = 3          /* Total number of module contexts in the cache. */
```
enjoy-binbin: We don't seem to use _ very often, can we drop the _ here?

src/scripting_engine.c
```diff
@@ -200,17 +223,24 @@ void scriptingEngineManagerForEachEngine(engineIterCallback callback,
...
-    if (e->module != NULL) {
-        serverAssert(e->module_ctx != NULL);
-        moduleScriptingEngineInitContext(e->module_ctx, e->module, c);
+static ValkeyModuleCtx *engineSetupModuleCtx(int module_ctx_cache_index,
+                                             scriptingEngine *e,
+                                             client *c) {
+    debugServerAssert(e != NULL);
+    if (e->module == NULL) {
+        return NULL;
     }
```
enjoy-binbin:
```suggestion
    if (e->module == NULL) return NULL;
```

enjoy-binbin (approved):
I may not be familiar with the details, but overall LGTM.

rjd15372:
@enjoy-binbin thanks for the review!

src/scripting_engine.c
```diff
@@ -8,6 +8,25 @@
...
+ * be at most 3 module contexts in use by the scripting engine.
+ *
+ * 1. The module context used by the scripting engine to run the script;
+ * 2. The module context used by `scriptingEngineCallGetMemoryInfo` that is
+ *    periodically called by the server cron; and
+ * 3. The module context used by `scriptingEngineCallFreeFunction` that is
+ *    called when the server needs to reset the evaluation environment in the
+ *    asynchronous mode.
+ */
+enum {
```
madolson: This seems OK to have as a named enum, is there a reason it's anonymous?

src/scripting_engine.c
```diff
@@ -200,17 +223,22 @@ void scriptingEngineManagerForEachEngine(engineIterCallback callback,
...
 
-static void engineSetupModuleCtx(scriptingEngine *e, client *c) {
-    if (e->module != NULL) {
-        serverAssert(e->module_ctx != NULL);
-        moduleScriptingEngineInitContext(e->module_ctx, e->module, c);
-    }
+static ValkeyModuleCtx *engineSetupModuleCtx(int module_ctx_cache_index,
+                                             scriptingEngine *e,
+                                             client *c) {
+    debugServerAssert(e != NULL);
```
madolson: This can just be a serverAssert. If it's null, it's crashing on the next line anyways. The point of debug is to avoid crashing for cases we can recover from.

madolson (commented):
I didn't originally approve it because I was confused about the code call path. It's odd to me that we call `engineSetupModuleCtx(engine, NULL);` even on the non-module call path. I sort of understand the fix, but it sort of feels like there might be a more structural issue.

src/scripting_engine.c
rjd15372: no reason, I just forgot to write the name.

rjd15372:
> I didn't originally approve it because I was confused about the code call path. It's odd to me that we call `engineSetupModuleCtx(engine, NULL);` even on the non-module call path. I sort of understand the fix, but it sort of feels like there might be a more structural issue.

We setup a module context for all callback functions as a conservative approach because the callback implementation might use the context object for some unforeseen reason.

PR #2360, opened 2025-07-16
Fix client tracking memory overhead calculation

enjoy-binbin:
This should be + instread of *, otherwise it does not make any sense.
Otherwise we would have to calculate 20 more bytes for each prefix rax
node in 64 bits build.

### DIFF
--- a/tests/unit/tracking.tcl
+++ b/tests/unit/tracking.tcl
@@ -38,6 +38,34 @@ start_server {tags {"tracking network logreqres:skip"}} {
         }
     }
 
+    test {Client tracking prefixes memory overhead} {
+        r CLIENT TRACKING off
+        set tot_mem_before [get_field_in_client_info [r client info] "tot-mem"]
+
+        # We add multiple $i to prefix to avoid prefix conflicts, so in this
+        # args we will have about 20000 rax nodes.
+        set args {}
+        for {set i 0} {$i < 10240} {incr i} {
+            lappend args PREFIX
+            lappend args PREFIX-$i-$i-$i-$i-$i-$i-$i
+        }
+        r CLIENT TRACKING on BCAST {*}$args
+
+        set arch_bits [s arch_bits]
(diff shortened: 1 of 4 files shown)


enjoy-binbin:
I'll try to add a test when i am free.

tests/unit/tracking.tcl
```diff
@@ -38,6 +38,34 @@ start_server {tags {"tracking network logreqres:skip"}} {
...
+        }
+        r CLIENT TRACKING on BCAST {*}$args
+
+        set arch_bits [s arch_bits]
+        set tot_mem_after [get_field_in_client_info [r client info] "tot-mem"]
+        set diff [expr $tot_mem_after - $tot_mem_before]
+
+        # In 64 bits, before we would consume about 20000 * (4 * 8), that is 640000.
+        # And now we are 20000 * (4 + 8), that is 240000.
+        if {$arch_bits == "64"} {
```
zuiderkwast: Number with quotes looks strange, but I know in TCL everything is strings so it is the same...

```suggestion
        if {$arch_bits == 64} {
```

zuiderkwast (approved):
Good catch.

We have many flakt tests now? The CI has many failures but not from this PR i think.

tests/unit/tracking.tcl
```diff
@@ -38,6 +38,34 @@ start_server {tags {"tracking network logreqres:skip"}} {
...
+            assert_lessthan $diff 300000
+        } elseif {$arch_bits == "32"} {
```
enjoy-binbin:
```suggestion
        } elseif {$arch_bits == 32} {
```

enjoy-binbin:
> We have many flakt tests now? The CI has many failures but not from this PR i think.

yes, there is one failure realted to slot migration, and i push the new PR to fix it (just waitting for Ping to review it), i haven't watched the others yet.

PR #2362, opened 2025-07-16
Fix memory leak with CLIENT LIST/KILL duplicate filters

hpatro:
With #1401, we introduced additional filters to CLIENT LIST/KILL subcommand. The intended behavior was to pick the last  value of the filter. However, we introduced memory leak for all the preceding filters. 

Before this change:
```
> CLIENT LIST IP 127.0.0.1 IP 127.0.0.1
id=4 addr=127.0.0.1:37866 laddr=127.0.0.1:6379 fd=10 name= age=0 idle=0 flags=N capa= db=0 sub=0 psub=0 ssub=0 multi=-1 watch=0 qbuf=0 qbuf-free=0 argv-mem=21 multi-mem=0 rbs=16384 rbp=16384 obl=0 oll=0 omem=0 tot-mem=16989 events=r cmd=client|list user=default redir=-1 resp=2 lib-name= lib-ver= tot-net-in=49 tot-net-out=0 tot-cmds=0
```
Leak:
```
Direct leak of 11 byte(s) in 1 object(s) allocated from:
    #0 0x7f2901aa557d in malloc (/lib64/libasan.so.4+0xd857d)
    #1 0x76db76 in ztrymalloc_usable_internal /workplace/harkrisp/valkey/src/zmalloc.c:156
    #2 0x76db76 in zmalloc_usable /workplace/harkrisp/valkey/src/zmalloc.c:200
    #3 0x4c4121 in _sdsnewlen.constprop.230 /workplace/harkrisp/valkey/src/sds.c:113
    #4 0x4dc456 in parseClientFiltersOrReply.constprop.63 /workplace/harkrisp/valkey/src/networking.c:4264
    #5 0x4bb9f7 in clientListCommand /workplace/harkrisp/valkey/src/networking.c:4600
    #6 0x641159 in call /workplace/harkrisp/valkey/src/server.c:3772
    #7 0x6431a6 in processCommand /workplace/harkrisp/valkey/src/server.c:4434
    #8 0x4bfa9b in processCommandAndResetClient /workplace/harkrisp/valkey/src/networking.c:3571
    #9 0x4bfa9b in processInputBuffer /workplace/harkrisp/valkey/src/networking.c:3702
    #10 0x4bffa3 in readQueryFromClient /workplace/harkrisp/valkey/src/networking.c:3812
    #11 0x481015 in callHandler /workplace/harkrisp/valkey/src/connhelpers.h:79
    #12 0x481015 in connSocketEventHandler.lto_priv.394 /workplace/harkrisp/valkey/src/socket.c:301
    #13 0x7d3fb3 in aeProcessEvents /workplace/harkrisp/valkey/src/ae.c:486
    #14 0x7d4d44 in aeMain /workplace/harkrisp/valkey/src/ae.c:543
    #15 0x453925 in main /workplace/harkrisp/valkey/src/server.c:7319
    #16 0x7f2900cd7139 in __libc_start_main (/lib64/libc.so.6+0x21139)
```

Note: For filter ID / NOT-ID we group all the option and perform filtering whereas for remaining filters we only pick the last filter option.

### DIFF
--- a/src/networking.c
+++ b/src/networking.c
@@ -4457,12 +4457,10 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
         int moreargs = c->argc > index + 1;
 
         if (!strcasecmp(c->argv[index]->ptr, "id")) {
-            if (filter->ids) {
-                zfree(filter->ids);
-            }
+            if (filter->ids == NULL) {
             /* Initialize the intset for IDs */
             filter->ids = intsetNew();
-
+            }
             index++; /* Move to the first ID after "ID" */
 
             /* Process all IDs until a non-numeric argument or end of args */
@@ -4481,12 +4479,10 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
                 index++; /* Move to the next argument */
             }
         } else if (!strcasecmp(c->argv[index]->ptr, "not-id")) {
-            if (filter->not_ids) {
-                zfree(filter->not_ids);
-            }
+            if (filter->not_ids == NULL) {
             /* Initialize the intset for NOT-IDs */
             filter->not_ids = intsetNew();
-
+            }
             index++; /* Move to the first ID after "NOT-ID" */
 
             /* Process all NOT-IDs until a non-numeric argument or end of args */
--- a/src/networking.c
+++ b/src/networking.c
@@ -4460,6 +4460,9 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
             if (filter->ids == NULL) {
                 /* Initialize the intset for IDs */
                 filter->ids = intsetNew();
+            } else {
+                zfree(filter->ids);
+                filter->ids = intsetNew();
             }
             index++; /* Move to the first ID after "ID" */
 
@@ -4481,6 +4484,9 @@ static int parseClientFiltersOrReply(client *c, int index, clientFilter *filter)
         } else if (!strcasecmp(c->argv[index]->ptr, "not-id")) {
             if (filter->not_ids == NULL) {
                 /* Initialize the intset for NOT-IDs */
+                filter->not_ids = intsetNew();
+            } else {
+                zfree(filter->not_ids);
                 filter->not_ids = intsetNew();
             }
             index++; /* Move to the first ID after "NOT-ID" */
--- a/src/networking.c
+++ b/src/networking.c
@@ -71,8 +71,6 @@ typedef struct {
     int type;
     int not_type;
     /* Boolean flag to determine if the current client (`me`) should be filtered. 1 means "skip me", 0 means otherwise. */
-    /* Flag to determine if the current client (`me`) should be filtered. 1 means "skip me", 0 means otherwise.
-     * If set to -1, the filter is not applied. */
     int skipme;
     /* Client name to filter. If NULL, no name filtering is applied. */
     char *name;
@@ -4760,7 +4758,7 @@ static int clientMatchesFilter(client *client, clientFilter *client_filter) {
     if (client_filter->type != -1 && getClientType(client) != client_filter->type) return 0;
     if (client_filter->ids && !intsetFind(client_filter->ids, client->id)) return 0;
     if (client_filter->user && client->user != client_filter->user) return 0;
-    if (client_filter->skipme == 1 && client == server.current_client) return 0;
+    if (client_filter->skipme && client == server.current_client) return 0;
     if (client_filter->max_age != 0 && (long long)(commandTimeSnapshot() / 1000 - client->ctime) < client_filter->max_age) return 0;
     if (client_filter->idle != 0 && (long long)(commandTimeSnapshot() / 1000 - client->last_interaction) < client_filter->idle) return 0;
     if (client_filter->flags && clientMatchesFlagFilter(client, client_filter->flags) == 0) return 0;
@@ -5058,7 +5056,6 @@ void clientListCommand(client *c) {
         filter.not_type = -1;
         filter.db_number = -1;
         filter.not_db_number = -1;
-        filter.skipme = -1;
 
         int i = 2;
 
@@ -5137,11 +5134,6 @@ void clientKillCommand(client *c) {
         if (parseClientFiltersOrReply(c, i, &client_filter) != C_OK) {
             /* Free the intset on error */
             goto client_kill_done;
-        } else {
-            /* No skipme option provided, default behavior with the new syntax is to skip */
-            if (client_filter.skipme == -1) {
-                client_filter.skipme = 1;
-            }
         }
     } else {
         addReplyErrorObject(c, shared.syntaxerr);
--- a/src/networking.c
+++ b/src/networking.c
@@ -71,6 +71,8 @@ typedef struct {
     int type;
     int not_type;
     /* Boolean flag to determine if the current client (`me`) should be filtered. 1 means "skip me", 0 means otherwise. */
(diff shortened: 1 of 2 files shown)


hpatro:
Initially thought of doing preprocessing first for duplicate filters. Felt it was lot of redundancy, so proceeded with single pass.

sarthakaggarwal97 (approved):
LGTM. Do we need to mention in the CLI help or documentation about supporting one filter type at a time?

hpatro:
> LGTM. Do we need to mention in the CLI help or documentation about supporting one filter type at a time?

valkey-cli / website command definition is correct as the keyspec was defined as per our expectation. So, I think it doesn't need any change.

From https://valkey.io/commands/client-kill/ :

```
CLIENT KILL 〈 old-format | 〈 [ ID client-id ] | [ [ ID client-id ] | ... ] [ TYPE normal | master | primary | slave | replica | pubsub ] | [ USER username ] | [ ADDR addr ] | [ LADDR laddr ] | [ SKIPME YES | NO ] | [ MAXAGE maxage ] | [ NAME name ] | [ IDLE idle ] | [ FLAGS flags ] | [ LIB-NAME lib-name ] | [ LIB-VER lib-ver ] | [ DB db ] | [ CAPA capa ] | [ IP ip ] 〉 [ 〈 [ ID client-id ] | [ [ ID client-id ] | ... ] [ TYPE normal | master | primary | slave | replica | pubsub ] | [ USER username ] | [ ADDR addr ] | [ LADDR laddr ] | [ SKIPME YES | NO ] | [ MAXAGE maxage ] | [ NAME name ] | [ IDLE idle ] | [ FLAGS flags ] | [ LIB-NAME lib-name ] | [ LIB-VER lib-ver ] | [ DB db ] | [ CAPA capa ] | [ IP ip ] 〉 ... ] 〉
```

soloestoy:
i'm afraid this may introduce breaking change, since in old version the `CLIENT KILL` command allows duplicated filters and the last win.

sarthakaggarwal97:
@soloestoy I think we can break it in 9.0 right? I don't think it was very intuitive to pick the last filter as well.

hpatro:
#2378 will have some conflict with this. 

@soloestoy I think we should go for the breaking change and have uniformity across all the filters. Your PR also adds quite a few new negative filters and I would prefer enforcing a filter to be applied once rather than multiple times with the last writer win approach. 

@valkey-io/core-team Could you please take a decision on this? 
👍 (one filter allowed once - breaking change) / 👎 (one filter allowed multiple times - last filter wins)

madolson:
I would rather not include the breaking changes. So let's document that the last filter wins and apply it accordingly.

soloestoy:
@hpatro I think we can fix just the memory leak first, WDYT?

hpatro:
> @hpatro I think we can fix just the memory leak first, WDYT?

Sure. Will update it.

sarthakaggarwal97 (approved):
@hpatro would you please fix the DCO as well. Thank you!

sarthakaggarwal97:
Removing the label as it is not a breaking change anymore.

enjoy-binbin (approved):

zuiderkwast:
> Removing the label as it is not a breaking change anymore.

I'm confused. How is "Allow each filter only once for CLIENT LIST/KILL" no longer a breaking change. Is the title wrong?

hpatro:
> > Removing the label as it is not a breaking change anymore.
> 
> I'm confused. How is "Allow each filter only once for CLIENT LIST/KILL" no longer a breaking change. Is the title wrong?

Submitted a change yesterday to avoid the change in behavior and fix the memory leak via last writer win approach for duplicate filters. Updated the title now 🙂.

hpatro:
@madolson / @soloestoy do you folks agree with the new change?

Will add some tests following that.

madolson:
@hpatro Yeah, I'm fine with the latest code change. Let's add tests for sure to make sure we don't regress. We should also update the documentation as well.

soloestoy (approved):

src/networking.c
```diff
@@ -4787,6 +4831,11 @@ void clientKillCommand(client *c) {
         if (parseClientFiltersOrReply(c, i, &client_filter) != C_OK) {
             /* Free the intset on error */
             goto client_kill_done;
+        } else {
+            /* No skipme option provided, default behavior with the new syntax is to skip */
+            if (client_filter.skipme == -1) {
+                client_filter.skipme = 1;
+            }
```
zuiderkwast: This is not necessary. It's initialized to `1` 15 lines above this.

src/networking.c
```diff
@@ -70,6 +70,8 @@ typedef struct {
     int type;
     int not_type;
     /* Boolean flag to determine if the current client (`me`) should be filtered. 1 means "skip me", 0 means otherwise. */
+    /* Flag to determine if the current client (`me`) should be filtered. 1 means "skip me", 0 means otherwise.
+     * If set to -1, the filter is not applied. */
     int skipme;
```
zuiderkwast: I don't see why it's a good idea to add the new value -1 to this. Does it make the logic simpler in any way? It just pushes the decision about the default value to to a different place in the code. I wasn't needed before.

Until now, it was done like this: In the CLIENT KILL implementation, we just initialize it to 1 and in CLIENT LIST we initialize it to 0, before we parse the filters. I think this is a very simple approach and I see no reason to change it, but I may be missing something.

(We could technically change it to a `bool` if we want to.)

zuiderkwast (commented):
LGTM in general. Just skeptical to some unrelated refactoring about SKIPME that seems unnecessary to me.

src/networking.c
hpatro: This is unnecessary after the last writer win approach. I've cleaned it up.

zuiderkwast (approved):
The behavior was already last-writer-wins, but with a memory leak, and this PR is actually only fixing the memory leak now?

If that's right, then maybe the title should be something like "Fix memory leak for CLIENT LIST/KILL duplicate filters" or similar?

sarthakaggarwal97 (approved):

hpatro:
> The behavior was already last-writer-wins, but with a memory leak, and this PR is actually only fixing the memory leak now?

Partially true. For ID/NOT-ID we were keeping all the ids from multiple filter(s) now I'm discarding and keeping the last one.

zuiderkwast:
> For ID/NOT-ID we were keeping all the ids from multiple filter(s) now I'm discarding and keeping the last one.

NOT-ID is new but ID is not new. Is this a behavior change then? I mean was `CLIENT LIST ID 1 ID 2 ID 3` was the same as `CLIENT LIST ID 1 2 3` since 6.2 but not anymore...?

hpatro:
> > > For ID/NOT-ID we were keeping all the ids from multiple filter(s) now I'm discarding and keeping the last one.
> > 
> > 
> > NOT-ID is new but ID is not new. Is this a behavior change then? I mean was `CLIENT LIST ID 1 ID 2 ID 3` was the same as `CLIENT LIST ID 1 2 3` since 6.2 but not anymore...?

You're right. Somewhere I thought multiple id filters is a new thing. So, let's stick to ID with previous behavior. NOT-ID aligned with ID's behavior and rest with LWW. @zuiderkwast ? I would like symmetry but I see the TSC is apprehensive to breaking changes.

zuiderkwast:
Yeah, I like symmetry too but avoiding breaking changes is even more important.

ranshid (approved):

ranshid:
@hpatro LGTM, but we should fix the failing tests before we merge this in.

PR #2365, opened 2025-07-17
Fix module function within setkey call lookupkey twice

kukey:
In VM_StringSet()/VM_StringTruncate()/VM_ModuleTypeSetValue() we could assert the key doesnt exists, so we could set SETKEY_DOESNT_EXIST flags to reduce one call to lookupkey.


rjd15372 (approved):
LGTM

PR #2366, opened 2025-07-17
Allow partial sync after loading AOF with preamble

### LINKED ISSUE #2677: [NEW] Support partial sync of a replica after a restart when aof is enabled

Currently, valkey supports partial sync for a replica after a restart only when rdb persistence is enabled. Partial sync is not possible after restart when aof is enabled.
From the valkey [docs](https://valkey.io/topics/replication/) - It is not possible to partially sync a replica that restarted via the AOF file. However the instance may be turned to RDB persistence before shutting down it, than can be restarted, and finally AOF can be enabled again.

Support partial sync of a replica that is restarted via aof file.

Considered the suggestion of turning to rdb persistence before shutting down and enabling aof after restart. But it doesnt work reliably when we always want aof to be enabled especially on masters.

**Additional information**

arthurkiller:
Noted this was reverted in #2925.

=====================

The AOF preamble mechanism replaces the traditional AOF base file with an RDB snapshot during rewrite operations, which reduces I/O overhead and improves loading performance. 
However, when valkey loads the RDB-formatted preamble from the base AOF file, it does not process the replication ID (replid) information within the RDB AUX fields. This omission has two limitations:

* On a primary, it prevents the primary from accepting PSYNC continue requests after restarting with a preamble-enabled AOF file.
* On a replica, it prevents the replica from successfully performing partial sync requests (avoiding full sync) after restarting with a preamble-enabled AOF file.

To resolve this, this commit aligns the AOF preamble handling with the logic used for standalone RDB files, by storing the replication ID and replication offset in the AOF preamble and restoring them when loading the AOF file.

Resolves #2677

### DIFF
--- a/src/aof.c
+++ b/src/aof.c
@@ -1470,7 +1470,7 @@ int loadSingleAppendOnlyFile(char *filename) {
             goto cleanup;
         } else {
             /* Restore the replication ID / offset from the RDB file. */
-            rsi_is_valid = replicationRestoreOffsetFromSaveInfo(&rsi, true);
+            rsi_is_valid = rdbRestoreOffsetFromSaveInfo(&rsi, true);
             loadingAbsProgress(ftello(fp));
             last_progress_report_size = ftello(fp);
             if (old_style) serverLog(LL_NOTICE, "Reading the remaining AOF tail...");
--- a/src/aof.c
+++ b/src/aof.c
@@ -1481,6 +1481,7 @@ int loadSingleAppendOnlyFile(char *filename) {
                     /* If this is a primary, we can save the replication info
                      * as secondary ID and offset, in order to allow replicas
                      * to partial resynchronizations with primaries. */
+                    if (server.repl_backlog == NULL) createReplicationBacklog();
                     memcpy(server.replid2, rsi.repl_id, sizeof(server.replid));
                     server.second_replid_offset = rsi.repl_offset + 1;
                     /* Rebase primary_repl_offset from rsi.repl_offset. */
--- a/src/aof.c
+++ b/src/aof.c
@@ -1453,7 +1453,13 @@ int loadSingleAppendOnlyFile(char *filename) {
         if (fseek(fp, 0, SEEK_SET) == -1) goto readerr;
         rioInitWithFile(&rdb, fp);
         rdbSaveInfo rsi = RDB_SAVE_INFO_INIT;
-        if (rdbLoadRio(&rdb, RDBFLAGS_AOF_PREAMBLE, &rsi) != C_OK) {
+        int rdb_flags = RDBFLAGS_AOF_PREAMBLE;
+        int rsi_is_valid = 0;
+        if (iAmPrimary()) {
+            if (server.repl_backlog == NULL) createReplicationBacklog();
+            rdb_flags |= RDBFLAGS_FEED_REPL;
+        }
+        if (rdbLoadRio(&rdb, rdb_flags, &rsi) != C_OK) {
             if (old_style)
                 serverLog(LL_WARNING, "Error reading the RDB preamble of the AOF file %s, AOF loading aborted",
                           filename);
@@ -1464,24 +1470,21 @@ int loadSingleAppendOnlyFile(char *filename) {
             goto cleanup;
         } else {
             /* Restore the replication ID / offset from the RDB file. */
-            if (rsi.repl_id_is_set && rsi.repl_offset != -1 &&
-                /* Note that older implementations may save a repl_stream_db
-                 * of -1 inside the RDB file in a wrong way, see more
-                 * information in function rdbPopulateSaveInfo. */
-                rsi.repl_stream_db != -1) {
+            if (rsi.repl_id_is_set && rsi.repl_offset != -1 && rsi.repl_stream_db != -1) {
+                rsi_is_valid = 1;
                 if (!iAmPrimary()) {
                     memcpy(server.replid, rsi.repl_id, sizeof(server.replid));
                     server.primary_repl_offset = rsi.repl_offset;
-                    /* If this is a replica, create a cached primary from this
-                     * information, in order to allow partial resynchronizations
-                     * with primaries. */
-                    replicationCachePrimaryUsingMyself();
-                    selectDb(server.cached_primary, rsi.repl_stream_db);
+                    if (!server.cached_primary) {
+                        replicationCachePrimaryUsingMyself();
+                        selectDb(server.cached_primary, rsi.repl_stream_db);
+                    }
+                    serverLog(LL_NOTICE, "preamble rdb changed replication info, replid: %s, primary_repl_offset: %lld", 
+                        server.replid, server.primary_repl_offset);
                 } else {
                     /* If this is a primary, we can save the replication info
                      * as secondary ID and offset, in order to allow replicas
                      * to partial resynchronizations with primaries. */
-                    if (server.repl_backlog == NULL) createReplicationBacklog();
                     memcpy(server.replid2, rsi.repl_id, sizeof(server.replid));
                     server.second_replid_offset = rsi.repl_offset + 1;
                     /* Rebase primary_repl_offset from rsi.repl_offset. */
@@ -1490,8 +1493,13 @@ int loadSingleAppendOnlyFile(char *filename) {
                     server.repl_backlog->offset = server.primary_repl_offset - server.repl_backlog->histlen + 1;
                     rebaseReplicationBuffer(rsi.repl_offset);
                     server.repl_no_replicas_since = time(NULL);
+                    serverLog(LL_NOTICE, "preamble rdb changed replication info, replid2: %s, secondary_repl_offset: " 
+                            "%lld, primary_repl_offset: %lld", 
+                            server.replid2, server.second_replid_offset, server.primary_repl_offset);
                 }
             }
+            if (!rsi_is_valid && server.repl_backlog) freeReplicationBacklog();
+
             loadingAbsProgress(ftello(fp));
             last_progress_report_size = ftello(fp);
             if (old_style) serverLog(LL_NOTICE, "Reading the remaining AOF tail...");
--- a/src/aof.c
+++ b/src/aof.c
@@ -1475,6 +1475,8 @@ int loadSingleAppendOnlyFile(char *filename) {
             last_progress_report_size = ftello(fp);
             if (old_style) serverLog(LL_NOTICE, "Reading the remaining AOF tail...");
         }
+        /* If the AOF didn't contain replication info, it's not possible to
+         * support partial resync, so we can free the backlog to save memory. */
         if (!rsi_is_valid && server.repl_backlog && listLength(server.replicas) == 0) freeReplicationBacklog();
     }
 
--- a/src/aof.c
+++ b/src/aof.c
@@ -1476,19 +1476,15 @@ int loadSingleAppendOnlyFile(char *filename) {
                     memcpy(server.replid, rsi.repl_id, sizeof(server.replid));
                     server.primary_repl_offset = rsi.repl_offset;
                     if (!server.primary && !server.cached_primary) {
-                        /* we will only cache primary if replica did not synced to its primary node yet */
+                        /* only cache myself primary if replica did not synced to its primary node yet */
                         replicationCachePrimaryUsingMyself();
                         selectDb(server.cached_primary, rsi.repl_stream_db);
                     }
                     serverLog(LL_NOTICE, "Loading preamble rdb changed replication info, replid: %s, primary_repl_offset: %lld",
                               server.replid, server.primary_repl_offset);
                 } else {
-                    /* If this is a primary, we can save the replication info
-                     * as secondary ID and offset, in order to allow replicas
-                     * to partial resynchronizations with primaries. */
                     memcpy(server.replid2, rsi.repl_id, sizeof(server.replid));
                     server.second_replid_offset = rsi.repl_offset + 1;
-                    /* Rebase primary_repl_offset from rsi.repl_offset. */
                     server.primary_repl_offset += rsi.repl_offset;
                     serverAssert(server.repl_backlog);
                     server.repl_backlog->offset = server.primary_repl_offset - server.repl_backlog->histlen + 1;
--- a/src/aof.c
+++ b/src/aof.c
@@ -1470,32 +1470,7 @@ int loadSingleAppendOnlyFile(char *filename) {
             goto cleanup;
         } else {
             /* Restore the replication ID / offset from the RDB file. */
-            if (rsi.repl_id_is_set && rsi.repl_offset != -1 && rsi.repl_stream_db != -1) {
-                rsi_is_valid = 1;
-                if (!iAmPrimary()) {
-                    memcpy(server.replid, rsi.repl_id, sizeof(server.replid));
-                    server.primary_repl_offset = rsi.repl_offset;
-                    if (!server.primary && !server.cached_primary) {
-                        /* only cache myself primary if replica did not synced to its primary node yet */
-                        replicationCachePrimaryUsingMyself();
-                        selectDb(server.cached_primary, rsi.repl_stream_db);
-                    }
-                    serverLog(LL_NOTICE, "Loading preamble rdb changed replication info, replid: %s, primary_repl_offset: %lld",
-                              server.replid, server.primary_repl_offset);
-                } else {
-                    memcpy(server.replid2, rsi.repl_id, sizeof(server.replid));
-                    server.second_replid_offset = rsi.repl_offset + 1;
-                    server.primary_repl_offset += rsi.repl_offset;
-                    serverAssert(server.repl_backlog);
-                    server.repl_backlog->offset = server.primary_repl_offset - server.repl_backlog->histlen + 1;
-                    rebaseReplicationBuffer(rsi.repl_offset);
-                    server.repl_no_replicas_since = time(NULL);
-                    serverLog(LL_NOTICE, "Loading preamble rdb changed replication info, replid2: %s, secondary_repl_offset: "
-                                         "%lld, primary_repl_offset: %lld",
-                              server.replid2, server.second_replid_offset, server.primary_repl_offset);
-                }
-            }
-
+            rsi_is_valid = replicationRestoreOffsetFromSaveInfo(&rsi, true);
             loadingAbsProgress(ftello(fp));
             last_progress_report_size = ftello(fp);
             if (old_style) serverLog(LL_NOTICE, "Reading the remaining AOF tail...");
--- a/src/aof.c
+++ b/src/aof.c
@@ -1475,12 +1475,13 @@ int loadSingleAppendOnlyFile(char *filename) {
                 if (!iAmPrimary()) {
                     memcpy(server.replid, rsi.repl_id, sizeof(server.replid));
                     server.primary_repl_offset = rsi.repl_offset;
-                    if (!server.cached_primary) {
+                    if (!server.primary && !server.cached_primary) {
+                        /* we will only cache primary if replica did not synced to its primary node yet */
                         replicationCachePrimaryUsingMyself();
                         selectDb(server.cached_primary, rsi.repl_stream_db);
                     }
-                    serverLog(LL_NOTICE, "preamble rdb changed replication info, replid: %s, primary_repl_offset: %lld", 
-                        server.replid, server.primary_repl_offset);
+                    serverLog(LL_NOTICE, "Loading preamble rdb changed replication info, replid: %s, primary_repl_offset: %lld",
+                              server.replid, server.primary_repl_offset);
                 } else {
                     /* If this is a primary, we can save the replication info
                      * as secondary ID and offset, in order to allow replicas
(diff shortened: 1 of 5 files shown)


secwall:
`./runtest-cluster` fails on my machine on new assert with backlog (`serverAssert(server.repl_backlog)`).
Could we also write some tests on this? For example for positive scenario?

arthurkiller:
> `./runtest-cluster` fails on my machine on new assert with backlog (`serverAssert(server.repl_backlog)`). Could we also write some tests on this? For example for positive scenario?

thx for your interest. The main purpose of this PR is to clarify why we missed this in preamble-aof, and make sure adding this will not break the consistency of data and replication. 

I will try add more test cases.

arthurkiller:
> `./runtest-cluster` fails on my machine on new assert with backlog (`serverAssert(server.repl_backlog)`). Could we also write some tests on this? For example for positive scenario?

@secwall I still can not get an error after a weekend of testing. Can you give some hints on reproducing?

secwall:
Here is how I get an error:
```
git clone https://github.com/arthurkiller/valkey.git
cd valkey
git checkout fix/load-replid-from-preamble-base-aof
make -j
./runtest-cluster
```
It fails on failover loop test:
```
*** Crash report found in valkey_6/log.txt ***
1555848:M 28 Jul 2025 11:02:33.690 # === ASSERTION FAILED ===
1555848:M 28 Jul 2025 11:02:33.690 # ==> aof.c:1488 'server.repl_backlog' is not true

------ STACK TRACE ------

1555848 valkey-server *
../../../src/valkey-server *:30006 [cluster](loadSingleAppendOnlyFile+0xf39)[0x5d6ce14d5af9]
../../../src/valkey-server *:30006 [cluster](loadAppendOnlyFiles+0x112)[0x5d6ce14d5cd2]
../../../src/valkey-server *:30006 [cluster](loadDataFromDisk+0x114)[0x5d6ce1443434]
../../../src/valkey-server *:30006 [cluster](main+0x378)[0x5d6ce1415778]
/lib/x86_64-linux-gnu/libc.so.6(+0x2a1ca)[0x73b7bf42a1ca]
/lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x8b)[0x73b7bf42a28b]
../../../src/valkey-server *:30006 [cluster](_start+0x25)[0x5d6ce14171a5]

1555849 bio_close_file
/lib/x86_64-linux-gnu/libc.so.6(+0x98d71)[0x73b7bf498d71]
/lib/x86_64-linux-gnu/libc.so.6(pthread_cond_wait+0x20d)[0x73b7bf49b7ed]
../../../src/valkey-server *:30006 [cluster](bioProcessBackgroundJobs+0x1af)[0x5d6ce1516a9f]
/lib/x86_64-linux-gnu/libc.so.6(+0x9caa4)[0x73b7bf49caa4]
/lib/x86_64-linux-gnu/libc.so.6(+0x129c3c)[0x73b7bf529c3c]

1555850 bio_aof
/lib/x86_64-linux-gnu/libc.so.6(+0x98d71)[0x73b7bf498d71]
/lib/x86_64-linux-gnu/libc.so.6(pthread_cond_wait+0x20d)[0x73b7bf49b7ed]
../../../src/valkey-server *:30006 [cluster](bioProcessBackgroundJobs+0x1af)[0x5d6ce1516a9f]
/lib/x86_64-linux-gnu/libc.so.6(+0x9caa4)[0x73b7bf49caa4]
/lib/x86_64-linux-gnu/libc.so.6(+0x129c3c)[0x73b7bf529c3c]

1555851 bio_lazy_free
/lib/x86_64-linux-gnu/libc.so.6(+0x98d71)[0x73b7bf498d71]
/lib/x86_64-linux-gnu/libc.so.6(pthread_cond_wait+0x20d)[0x73b7bf49b7ed]
../../../src/valkey-server *:30006 [cluster](bioProcessBackgroundJobs+0x1af)[0x5d6ce1516a9f]
/lib/x86_64-linux-gnu/libc.so.6(+0x9caa4)[0x73b7bf49caa4]
/lib/x86_64-linux-gnu/libc.so.6(+0x129c3c)[0x73b7bf529c3c]

4/4 expected stacktraces.

------ STACK TRACE DONE ------
```

ranshid:
@arthurkiller can you also signoff the commits?

arthurkiller:
> @arthurkiller can you also signoff the commits?

@ranshid done

@secwall I've fixed this. Cause repl_buffer is created when we get a replica attached. We'd create this on data load. I'll add a tcl case for this later.

Not so sure if we use this replid from the previous RDB, will it make any side effects

arthurkiller:
> Here is how I get an error:
> 
> ```
> git clone https://github.com/arthurkiller/valkey.git
> cd valkey
> git checkout fix/load-replid-from-preamble-base-aof
> make -j
> ./runtest-cluster
> ```
> 
> It fails on failover loop test:
> 
> ```
> *** Crash report found in valkey_6/log.txt ***
> === VALKEY BUG REPORT START: Cut & paste starting from here ===
> 1555848:M 28 Jul 2025 11:02:33.690 # === ASSERTION FAILED ===
> 1555848:M 28 Jul 2025 11:02:33.690 # ==> aof.c:1488 'server.repl_backlog' is not true
> 
> ------ STACK TRACE ------
> 
> 1555848 valkey-server *
> ../../../src/valkey-server *:30006 [cluster](loadSingleAppendOnlyFile+0xf39)[0x5d6ce14d5af9]
> ../../../src/valkey-server *:30006 [cluster](loadAppendOnlyFiles+0x112)[0x5d6ce14d5cd2]
> ../../../src/valkey-server *:30006 [cluster](loadDataFromDisk+0x114)[0x5d6ce1443434]
> ../../../src/valkey-server *:30006 [cluster](main+0x378)[0x5d6ce1415778]
> /lib/x86_64-linux-gnu/libc.so.6(+0x2a1ca)[0x73b7bf42a1ca]
> /lib/x86_64-linux-gnu/libc.so.6(__libc_start_main+0x8b)[0x73b7bf42a28b]
> ../../../src/valkey-server *:30006 [cluster](_start+0x25)[0x5d6ce14171a5]
> 
> 1555849 bio_close_file
> /lib/x86_64-linux-gnu/libc.so.6(+0x98d71)[0x73b7bf498d71]
> /lib/x86_64-linux-gnu/libc.so.6(pthread_cond_wait+0x20d)[0x73b7bf49b7ed]
> ../../../src/valkey-server *:30006 [cluster](bioProcessBackgroundJobs+0x1af)[0x5d6ce1516a9f]
> /lib/x86_64-linux-gnu/libc.so.6(+0x9caa4)[0x73b7bf49caa4]
> /lib/x86_64-linux-gnu/libc.so.6(+0x129c3c)[0x73b7bf529c3c]
> 
> 1555850 bio_aof
> /lib/x86_64-linux-gnu/libc.so.6(+0x98d71)[0x73b7bf498d71]
> /lib/x86_64-linux-gnu/libc.so.6(pthread_cond_wait+0x20d)[0x73b7bf49b7ed]
> ../../../src/valkey-server *:30006 [cluster](bioProcessBackgroundJobs+0x1af)[0x5d6ce1516a9f]
> /lib/x86_64-linux-gnu/libc.so.6(+0x9caa4)[0x73b7bf49caa4]
> /lib/x86_64-linux-gnu/libc.so.6(+0x129c3c)[0x73b7bf529c3c]
> 
> 1555851 bio_lazy_free
> /lib/x86_64-linux-gnu/libc.so.6(+0x98d71)[0x73b7bf498d71]
> /lib/x86_64-linux-gnu/libc.so.6(pthread_cond_wait+0x20d)[0x73b7bf49b7ed]
> ../../../src/valkey-server *:30006 [cluster](bioProcessBackgroundJobs+0x1af)[0x5d6ce1516a9f]
> /lib/x86_64-linux-gnu/libc.so.6(+0x9caa4)[0x73b7bf49caa4]
> /lib/x86_64-linux-gnu/libc.so.6(+0x129c3c)[0x73b7bf529c3c]
> 
> 4/4 expected stacktraces.
> 
> ------ STACK TRACE DONE ------
> ```

Currently, our cluster cases can not cover this situation

I still failed get this by run cluster test
```
21:10:51> Wait cluster to be stable: OK
21:10:51> Master #0 still should have its replicas: OK
21:10:51> Each master should have at least two replicas attached: OK
Testing unit: 28-cluster-shards.tcl
21:10:51> (init) Restart killed instances: OK
21:10:51> Cluster nodes are reachable: OK
21:10:51> Cluster nodes hard reset: OK
21:10:51> Cluster Join and auto-discovery test: OK
21:10:55> Before slots allocation, all nodes report cluster failure: OK
21:10:55> Create a 8 nodes cluster with 4 shards: OK
21:10:55> Cluster should start ok: OK
21:10:59> Set cluster hostnames and verify they are propagated: OK
21:11:01> Verify information about the shards: OK
21:11:01> Verify no slot shard: OK
21:11:01> Kill a node and tell the replica to immediately takeover: OK
21:11:02> Verify health as fail for killed node: OK
21:11:06> Restarting primary node: OK
21:11:06> Instance #0 gets converted into a replica: OK
21:11:06> Test the replica reports a loading state while it's loading: OK
21:11:09> Regression test for a crash when calling SHARDS during handshake: OK
21:11:09> Cluster is up: OK
21:11:12> Shard ids are unique: OK
21:11:12> CLUSTER MYSHARDID reports same id for both primary and replica: OK
21:11:12> New replica receives primary's shard id: OK
21:11:12> CLUSTER MYSHARDID reports same shard id after shard restart: OK
21:11:14> CLUSTER MYSHARDID reports same shard id after cluster restart: OK
Cleaning up...
killing stale instance 2625352
killing stale instance 2625364
killing stale instance 2625395
killing stale instance 2625405
killing stale instance 2625416
killing stale instance 2625432
killing stale instance 2625443
killing stale instance 2625452
killing stale instance 2625462
killing stale instance 2625473
killing stale instance 2626003
killing stale instance 2626009
killing stale instance 2628745
killing stale instance 2628751
killing stale instance 2628757
killing stale instance 2628763
killing stale instance 2628769
killing stale instance 2628775
killing stale instance 2628782
killing stale instance 2628789
GOOD! No errors.
```

zuiderkwast:
Great finding @arthurkiller! One of the main points of AOF is to be able to avoid full sync after a restart, so it's pretty sad that it doesn't work and isn't tested. Maybe this is why our documentation about [prsistence](https://valkey.io/topics/persistence/#ok-so-what-should-i-use) says this:

> There are many users using AOF alone, but we discourage it since to have an RDB snapshot from time to time is a great idea for doing database backups, for faster restarts, and in the event of bugs in the AOF engine.

It's too late to get this in 9.0 but I hope we can have it ready for 9.1. Will you add a test of the scenario (AOF rewrite, restart, psync)?

arthurkiller:
> Great finding @arthurkiller! One of the main points of AOF is to be able to avoid full sync after a restart, so it's pretty sad that it doesn't work and isn't tested. Maybe this is why our documentation about [prsistence](https://valkey.io/topics/persistence/#ok-so-what-should-i-use) says this:
> 
> > There are many users using AOF alone, but we discourage it since to have an RDB snapshot from time to time is a great idea for doing database backups, for faster restarts, and in the event of bugs in the AOF engine.
> 
> It's too late to get this in 9.0 but I hope we can have it ready for 9.1. Will you add a test of the scenario (AOF rewrite, restart, psync)?

@zuiderkwast Thank you for your attention. I almost forgot about this Pull Request :) .

I'll try to add more comprehensive tests as soon as possible. Hectic these days.

zuiderkwast:
When you have time, will you add these two in some file under `tests/unit/cluster/`:

1. Replica restart with AOF load, then PSYNC.
2. Primary loads AOF, then replica doing partial sync from the primary.

We should use an AOF file with preamble and multiple AOF commands after the preamble to check that the offset is calculated correctly.

tests/integration/replication-4.tcl
```diff
@@ -293,3 +293,51 @@ start_server {tags {"repl external:skip"}} {
         }
     }
 }
+
+# test aof persistence replid info and load it when reboot server
+start_server {tags {"repl"} overrides {appendonly yes}} {
```
zuiderkwast: Needs to add these tags:

* `external:skip`, because I don't think the external tests can handle nested start-server. It runs against an externally running server which is only one. See this message: https://github.com/valkey-io/valkey/actions/runs/18647217370/job/53156999565?pr=2366#step:5:3349
* `cluster:skip`, because `SLAVEOF` is not allowed in cluster mode. See https://github.com/valkey-io/valkey/actions/runs/18647217370/job/53156999555?pr=2366#step:6:3351

tests/integration/replication-4.tcl
```diff
@@ -293,3 +293,51 @@ start_server {tags {"repl external:skip"}} {
...
+    start_server {overrides {appendonly yes}} {
+        set master_id -1
+        set replica_id 0
+        set master [srv $master_id client]
```
zuiderkwast: Let's use the word `primary` instead of `master`.

tests/integration/replication-4.tcl
```diff
@@ -293,3 +293,51 @@ start_server {tags {"repl external:skip"}} {
...
+        set master_host [srv $master_id host]
+        set master_port [srv $master_id port]
+        set replica [srv $replica_id client]
+
+        $replica slaveof $master_host $master_port
```
zuiderkwast: Let's use REPLICAOF instead of SLAVEOF. (Same command, different name.)

zuiderkwast (commented):
Some hints for the failed CI jobs.

We can ignore this one because it's a known flaky test: https://github.com/valkey-io/valkey/actions/runs/18647217373/job/53157000675?pr=2366#step:4:8645

* #2699

arthurkiller:
> Some hints for the failed CI jobs.
> 
> We can ignore this one because it's a known flaky test: https://github.com/valkey-io/valkey/actions/runs/18647217373/job/53157000675?pr=2366#step:4:8645
> 
> * #2699

I have fixed all the test cases, and now it can run stably.
But the implementation of the tcl cases is not perfect yet. I will try to find a way to remove the `after` in the case later.

arthurkiller:
`wait_for_ofs_sync` and `wait_for_sync` did not work stably in these cases, and `wait_for_log_messages` is hard to use when I need to restart the server.

zuiderkwast:
> `wait_for_log_messages` is hard to use when I need to restart the server.

Not sure what the problem is, but maybe this helps.

Before you restart, you can get the number of lines in the log file using `wc -l`, and after restart, start searching from that line:

```tcl
set logfile [srv 0 stdout]
set num_lines [lindex [exec wc -l $logfile] 0]
# ...
restart_server
# ...
wait_for_log_messages 0 [list {*Pattern*}] $num_lines 100 100
```

Another way is to count the matching lines before and after restart using `count_message_lines $logfile $pattern`.

arthurkiller:
@zuiderkwast I think we are almost done, `wc -l` really helps. 
The real problem here is restart server should be called with `now`. This will avoid additional writes, which will do harm to the replication offset in cases

src/aof.c
```diff
@@ -1462,10 +1469,38 @@ int loadSingleAppendOnlyFile(char *filename) {
             ret = AOF_FAILED;
             goto cleanup;
         } else {
+            /* Restore the replication ID / offset from the RDB file. */
+            if (rsi.repl_id_is_set && rsi.repl_offset != -1 && rsi.repl_stream_db != -1) {
```
zuiderkwast: This code is duplicated from `loadDataFromDisk()` where an RDB file is loaded. Let's avoid duplicated code by moving this common code to a function?

src/aof.c
```diff
@@ -1462,10 +1469,38 @@ int loadSingleAppendOnlyFile(char *filename) {
...
+                rsi_is_valid = 1;
+                if (!iAmPrimary()) {
+                    memcpy(server.replid, rsi.repl_id, sizeof(server.replid));
+                    server.primary_repl_offset = rsi.repl_offset;
+                    if (!server.primary && !server.cached_primary) {
```
arthurkiller: I did not merge the code cause we have a little difference here.

This condition is a workaround for `debug loadaof`, if we call this after the instance becomes a replica, then we should not call `replicationCachePrimaryUsingMyself` here.

arthurkiller: @zuiderkwast

zuiderkwast: We can't do this for RDB too?

arthurkiller: > We can't do this for RDB too?

@zuiderkwast we do not need to care about this when loading the RDB

The data loading in the debug command makes the implementation here a bit complex. The implementation of debug reload directly loads RDB data and ignores updates related to replication info, while debug loadaof executes into loadSingleAppendOnlyFile — where the issue of replication info during AOF file loading is handled based on the type (this is also the logic that this MR intends to modify). 
From the perspective of not breaking existing functionality, I tend to keep the implementation of the `debug loadaof` command with the semantics of only loading data. Therefore, it is necessary to introduce an additional parameter in the function signature to indicate whether this loading only loads data (because during startup loading, both data and replication info should be loaded).

1. Add another args to distinguish from loading aof data on startup or by debug command
2. or, discard the current replication info and disconnect all replicas before calling `debug loadaof`, and have the same behavior as what was done in rdb loading

WDYT

arthurkiller: In addition, the reason for this additional check is that directly updating the cached primary will lead to the unexpected release of `server.primary`, causing the replication logic to fail. Therefore, in Plan 2, we need to ensure that before executing `debug loadaof`, the primary and replica have been disconnected — this way, we can safely update the cached primary.

zuiderkwast: OK, I agree with the logic here, but I think we could still de-duplicate the code without affecting the logical difference between AOF and RDB loading. Sometimes, logic differences can make common code be very messy, but here it is just one small if condition that differs. Future changes to this logic should most likely affect both RDB and AOF-preamble, so I imagine there is some benefit to avoid code duplication.

I diffed this part of the code with the comment `/* Restore the replication ID / offset from the RDB file. */` for RDB and AOF and notices these differences:

* This `if` condition for AOF preamble.
* The RDB code has more code comments.
* The AOF code (in this PR) has a few extra log notice. (Do we need them?)

Here is an attempt to unify this code in a function that we could place in replication.c, if you agree.

```c
/* Restore the replication ID / offset from the RDB file. */
void replicationRestoreOffsetFromSaveInfo(rdbSaveInfo *rsi, bool is_aof_preamble) {
    if (rsi->repl_id_is_set && rsi->repl_offset != -1 && rsi->repl_stream_db != -1) {
        /* Note that older implementations may save a repl_stream_db
         * of -1 inside the RDB file in a wrong way, see more
         * information in function rdbPopulateSaveInfo. */
        rsi_is_valid = 1;
        if (!iAmPrimary()) {
            memcpy(server.replid, rsi->repl_id, sizeof(server.replid));
            server.primary_repl_offset = rsi->repl_offset;
            if (!is_aof_preamble || (!server.primary && !server.cached_primary)) {
                /* If this is a replica, create a cached primary from this
                 * information, in order to allow partial resynchronizations
                 * with primaries. For AOF, only cache the primary if replica
                 * has not synced to its primary node yet. */
                replicationCachePrimaryUsingMyself();
                selectDb(server.cached_primary, rsi->repl_stream_db);
            }
        } else {
            /* If this is a primary, we can save the replication info
             * as secondary ID and offset, in order to allow replicas
             * to partial resynchronizations with primaries. */
            memcpy(server.replid2, rsi->repl_id, sizeof(server.replid));
            server.second_replid_offset = rsi->repl_offset + 1;
            /* Rebase primary_repl_offset from rsi.repl_offset. */
            server.primary_repl_offset += rsi->repl_offset;
            serverAssert(server.repl_backlog);
            server.repl_backlog->offset = server.primary_repl_offset - server.repl_backlog->histlen + 1;
            rebaseReplicationBuffer(rsi->repl_offset);
            server.repl_no_replicas_since = time(NULL);
        }
    }
}
```

arthurkiller: LGTM, thx for the comment

src/rdb.c
```diff
@@ -3952,3 +3952,41 @@ rdbSaveInfo *rdbPopulateSaveInfo(rdbSaveInfo *rsi) {
     }
     return NULL;
 }
+
+/* Restore the replication ID / offset from the RDB file
+ * return 1 if rdbSaveInfo is valid */
+int replicationRestoreOffsetFromSaveInfo(rdbSaveInfo *rsi, bool is_aof_preamble) {
```
zuiderkwast: With the prefix "replication" I was expecting it to be in `replication.c`.

If you think it better belongs in `rdb.c`, then it should use "rdb" prefix instead.

I'm not sure what's better. My thinking was that `rdb.c` is lower level library and it shouldn't access globals like the `server` object, while `replication.c` is more tightly coupled with the server state.

src/rdb.c
```diff
@@ -3952,3 +3952,41 @@ rdbSaveInfo *rdbPopulateSaveInfo(rdbSaveInfo *rsi) {
...
+    int rsi_is_valid = 0;
+    if (rsi == NULL) return rsi_is_valid;
```
zuiderkwast: We always pass a valid `rsi` pointer, don't we? No need to allow passing NULL. We can do an assert instead.

```suggestion
    serverAssert(rsi != NULL);
```

arthurkiller: This is a defensive code, but I think `assert` makes sense

src/rdb.c
arthurkiller: I considered it from the perspective of handling `rdb save info`. The relevant processing logics are all here, so I put them together.

arthurkiller: I prefer changing the prefix to `rdb`

src/rdb.c
```diff
@@ -3952,3 +3952,41 @@ rdbSaveInfo *rdbPopulateSaveInfo(rdbSaveInfo *rsi) {
     }
     return NULL;
 }
+
+/* Restore the replication ID / offset from the RDB file
+ * return 1 if rdbSaveInfo is valid */
```
zuiderkwast: "valid" here means that the save info contained replication ID and offset and could be successfully restored, right? Let's make the comment more explicit?

```suggestion
 * return 1 if replication ID and offset were restored from the rdbSaveInfo */
```

src/aof.c
```diff
@@ -1462,10 +1469,13 @@ int loadSingleAppendOnlyFile(char *filename) {
             ret = AOF_FAILED;
             goto cleanup;
         } else {
+            /* Restore the replication ID / offset from the RDB file. */
+            rsi_is_valid = rdbRestoreOffsetFromSaveInfo(&rsi, true);
             loadingAbsProgress(ftello(fp));
             last_progress_report_size = ftello(fp);
             if (old_style) serverLog(LL_NOTICE, "Reading the remaining AOF tail...");
         }
+        if (!rsi_is_valid && server.repl_backlog && listLength(server.replicas) == 0) freeReplicationBacklog();
```
zuiderkwast, line 1480: I was curious why we do this and I found the answer in a comment in the corresponding code for RDB load. Is it worth adding a similar comment here? A shorter one perhaps like this?

```suggestion
        /* If the AOF didn't contain replication info, it's not possible to
         * support partial resync, so we can free the backlog to save memory. */
        if (!rsi_is_valid && server.repl_backlog && listLength(server.replicas) == 0) freeReplicationBacklog();
```

zuiderkwast (approved):
Very good, thank you for your patience!

I have only very minor comments about comments. (I can apply them before merging, if you don't.)

arthurkiller:
Thanks a lot :)

arthurkiller:
@zuiderkwast I think this patch did not work correctly for the replica.
We loaded replication info from rdb and set it into `repl_offset`. But the incr aof loading did not increase this offset.

If the master has some non-idempotent command such as `incrby` replicated to replica and write into the Incr aof after the Base Aof, this can cause inconsistency between the replica and the master.

arthurkiller:
One possible approach is replica should try to connect to the master while loading the data, only load the rdb data and try to do psync-continue with master if the master is alive. Otherwise, we should load the whole data.

ranshid:
removing the release-notes label for now to avoid accidentally adding this to the release note as this was currently reverted in https://github.com/valkey-io/valkey/pull/2925

PR #2368, opened 2025-07-17
add parameter for initiating bgrewriteaof on exceeding threshold AOF size

kronwerk:
there are some scenarios of filling up the whole disk besides any chosen AOF rewrite strategy - the simplest is this:
- we have auto-aof-rewrite-percentage 100,
- and last automatic rewrite compacted AOF to 51% of disk,
-> we'll never try again.

we can lower auto-percentage, but in some cases it's not what we want to. I suggest that we add soft upper limit in this PR:
- if no AOF rewrite is running currently,
- and limit is set in config,
- and AOF exceed the limit,
- and last rewrite compacted AOF to somewhat smaller than limit (otherwise we might end up with useless attempts of uninterrupted rewrites)
-> bgrewriteaof is initiated.

this doesn't completely solves the problem, but gives some additional possibilities for some users to improve the situation initially described in #540 (noticed that solution from the issue seems to be inappropriate for some cases)

### DIFF
--- a/src/server.c
+++ b/src/server.c
@@ -1576,15 +1576,15 @@ long long serverCron(struct aeEventLoop *eventLoop, long long id, void *clientDa
     databasesCron();
 
     /* Start a scheduled AOF rewrite if this was requested by the user while
-     * a BGSAVE was in progress. Also handle rewrites triggered by aof-max-size. */
+     * a BGSAVE was in progress. Also handle rewrites triggered by auto-aof-rewrite-max-size. */
     if (!hasActiveChildProcess() && !aofRewriteLimited()) {
         if (server.aof_rewrite_scheduled ||
-            server.aof_current_size >= server.aof_rewrite_min_size &&
-            server.aof_current_size >= server.aof_max_size &&
-            server.aof_rewrite_base_size < server.aof_max_size) {
+            (server.aof_current_size >= server.aof_rewrite_min_size &&
+             server.aof_current_size >= server.aof_rewrite_max_size &&
+             server.aof_rewrite_base_size < server.aof_rewrite_max_size)) {
             /* user-initiated OR
                (AOF is large enough AND
-               last rewrite success should be less than aof-max-size - or we might end up with eternal rewrites)
+               last rewrite success should be less than auto-aof-rewrite-max-size - or we might end up with eternal rewrites)
             */
             rewriteAppendOnlyFileBackground();
         }
@@ -6164,17 +6164,10 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "slot_migration_fork_in_progress:%d\r\n", server.child_type == CHILD_TYPE_SLOT_MIGRATION));
 
         if (server.aof_enabled) {
-            char aof_current_size_hdsk[64];
-            char aof_max_size_hdsk[64];
-            bytesToHuman(aof_current_size_hdsk, sizeof(aof_current_size_hdsk), (unsigned long long)server.aof_current_size);
-            bytesToHuman(aof_max_size_hdsk, sizeof(aof_max_size_hdsk), (unsigned long long)server.aof_max_size);
             info = sdscatprintf(
                 info,
                 FMTARGS(
                     "aof_current_size:%lld\r\n", (long long)server.aof_current_size,
-                    "aof_current_size_human:%s\r\n", aof_current_size_hdsk,
-                    "aof_max_size:%lld\r\n", (long long)server.aof_max_size,
-                    "aof_max_size_human:%s\r\n", aof_max_size_hdsk,
                     "aof_base_size:%lld\r\n", (long long)server.aof_rewrite_base_size,
                     "aof_pending_rewrite:%d\r\n", server.aof_rewrite_scheduled,
                     "aof_buffer_length:%zu\r\n", sdslen(server.aof_buf),
@@ -7548,12 +7541,12 @@ __attribute__((weak)) int main(int argc, char **argv) {
                   server.maxmemory);
     }
 
-    /* Warning the user about suspicious aof-max-size setting. */
-    if (server.aof_max_size > 0 && server.aof_max_size < 1024 * 1024) {
+    /* Warning the user about suspicious auto-aof-rewrite-max-size setting. */
+    if (server.aof_rewrite_max_size > 0 && server.aof_rewrite_max_size < 1024 * 1024) {
         serverLog(LL_WARNING,
-                  "WARNING: You specified a aof-max-size value that is less than 1MB (current value is %lld bytes). Are "
+                  "WARNING: You specified a auto-aof-rewrite-max-size value that is less than 1MB (current value is %lld bytes). Are "
                   "you sure this is what you really want?",
-                  (long long)server.aof_max_size);
+                  (long long)server.aof_rewrite_max_size);
     }
 
     serverSetCpuAffinity(server.server_cpulist);
--- a/src/server.c
+++ b/src/server.c
@@ -1576,9 +1576,18 @@ long long serverCron(struct aeEventLoop *eventLoop, long long id, void *clientDa
     databasesCron();
 
     /* Start a scheduled AOF rewrite if this was requested by the user while
-     * a BGSAVE was in progress. */
-    if (!hasActiveChildProcess() && server.aof_rewrite_scheduled && !aofRewriteLimited()) {
-        rewriteAppendOnlyFileBackground();
+     * a BGSAVE was in progress. Also handle rewrites triggered by aof-max-size. */
+    if (!hasActiveChildProcess() && !aofRewriteLimited()) {
+        if (server.aof_rewrite_scheduled ||
+            server.aof_current_size >= server.aof_rewrite_min_size &&
+            server.aof_current_size >= server.aof_max_size &&
+            server.aof_rewrite_base_size < server.aof_max_size) {
+            /* user-initiated OR
+               (AOF is large enough AND
+               last rewrite success should be less than aof-max-size - or we might end up with eternal rewrites)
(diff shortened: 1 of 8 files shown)


kronwerk:
@valkey-io/core-team PTAL

kronwerk:
@valkey-io/core-team anybody?

src/server.c
```diff
@@ -5977,10 +5981,10 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "rdb_last_load_keys_expired:%lld\r\n", server.rdb_last_load_keys_expired,
                 "rdb_last_load_keys_loaded:%lld\r\n", server.rdb_last_load_keys_loaded,
                 "aof_enabled:%d\r\n", server.aof_state != AOF_OFF,
-                "aof_rewrite_in_progress:%d\r\n", server.child_type == CHILD_TYPE_AOF,
+                "aof_rewrite_in_progress:%d\r\n", isAofRewriteInProgress(),
                 "aof_rewrite_scheduled:%d\r\n", server.aof_rewrite_scheduled,
                 "aof_last_rewrite_time_sec:%jd\r\n", (intmax_t)server.aof_rewrite_time_last,
-                "aof_current_rewrite_time_sec:%jd\r\n", (intmax_t)((server.child_type != CHILD_TYPE_AOF) ? -1 : time(NULL) - server.aof_rewrite_time_start),
+                "aof_current_rewrite_time_sec:%jd\r\n", (intmax_t)((!isAofRewriteInProgress()) ? -1 : time(NULL) - server.aof_rewrite_time_start),
```
dvkashapov, line 6156: Offtopic: @kronwerk do we use `aof_current_rewrite_time_sec` anywhere internally?

dvkashapov (commented):
@hpatro can you PTAL? Do we have any concerns against this feature being added? We're using this in our Valkey deployments and seems to be very useful for AOF.

src/server.c
kronwerk: yeah, in a few places

dvkashapov:
@zuiderkwast or @enjoy-binbin Can you PTAL? I think this PR was open at an unfortunate moment close to 9.0 release and was overseen because of that.

valkey.conf
```diff
@@ -1720,6 +1720,25 @@ aof-use-rdb-preamble yes
 # the AOF format in a way that may not be compatible with existing AOF parsers.
 aof-timestamp-enabled no
 
+# Maximum target size for AOF files on disk in bytes. This is a soft limit
+# that triggers automatic background rewrites when exceeded, not a hard limit
+# that prevents writes.
+#
+# When the AOF file size reaches this threshold, a background rewrite is
+# automatically triggered (similar to auto-aof-rewrite-percentage but based
+# on absolute size rather than percentage growth).
```
zuiderkwast: Put this together with auto-aof-rewrite-percentage and auto-aof-rewrite-min-size above and combine the documentation with that text. They control the same thing: auto-aof-rewrite. The comment there starts with `# Automatic rewrite of the append only file`. I think the new config name should also use the prefix auto-aof-rewrite.

How about `auto-aof-rewrite-max-size`?

The PR description mentions:

> and last rewrite compacted AOF to somewhat smaller than limit (otherwise we might end up with useless attempts of uninterrupted rewrites)

How do we control this threshold? Do we need a config for that too?

src/server.c
```diff
@@ -6151,10 +6164,17 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
...
+            char aof_max_size_hdsk[64];
+            bytesToHuman(aof_current_size_hdsk, sizeof(aof_current_size_hdsk), (unsigned long long)server.aof_current_size);
+            bytesToHuman(aof_max_size_hdsk, sizeof(aof_max_size_hdsk), (unsigned long long)server.aof_max_size);
             info = sdscatprintf(
                 info,
                 FMTARGS(
                     "aof_current_size:%lld\r\n", (long long)server.aof_current_size,
+                    "aof_current_size_human:%s\r\n", aof_current_size_hdsk,
+                    "aof_max_size:%lld\r\n", (long long)server.aof_max_size,
+                    "aof_max_size_human:%s\r\n", aof_max_size_hdsk,
```
zuiderkwast: I don't think we should add these new INFO fields. We want to avoid increasing the size and latency of the INFO command.

One is just the config value. You can get that using CONFIG GET.

The other two are human representation of existing fields. The user can calculate this themselves.

.codeassistant/rules/rules.md
zuiderkwast, line 1: Don't add this file.

zuiderkwast (commented):
I took a quick look at the config and INFO fields.

I'm not the right maintainer to decide about if it's a good idea or not or if there is another way to solve this kind of problems, because don't have enough experience with AOF. Someone else from @valkey-io/core-team needs to be involved.

valkey.conf
kronwerk: 1. gonna move it, yeah
2. naming seems ok, let's change it
3. this threshold is just the same config option:
- when AOF current exceeds auto-aof-rewrite-max-size - we initiate rewrite;
- but additionally to this when AOF base exceeds auto-aof-rewrite-max-size - we don't, cause that lead to possible infinite (and more likely not necessary) rewrite attempts

we can add 1 more limit to make it 2 different thresholds (e.g. 40% of disk for current to attempt rewrite, 50% of disk for base to prevent it) - but that seems a bit more confusing really, and I'm not sure if that's gonna help somebody (debatable)

zuiderkwast: OK, sounds good. Less configs is better.

kronwerk:
@hpatro maybe you?

enjoy-binbin (commented):
I'm curious about this: the message is "Your working dataset is larger than the limit," if the dataset is indeed that large, what should we do? Or, if the dataset grows over time, do we need to continuously adjust this configuration setting dynamically?
```
# Option 2 (auto-aof-rewrite-max-size).
# AOF base target maximum in bytes. This is a soft limit
# that triggers automatic background rewrites when exceeded, not a hard limit
# that prevents writes.
#
# When the AOF file size reaches this threshold, a background rewrite is
# automatically triggered (similar to auto-aof-rewrite-percentage but based
# on absolute size rather than percentage growth).
#
# Important notes:
# - Set to 0 to disable size-based triggering (default)
# - The actual AOF size may exceed this limit if:
#   * Your working dataset is larger than the limit
#   * A rewrite is already in progress
#   * The rewritten AOF cannot compress below this size
# - This works alongside auto-aof-rewrite-percentage. Whichever condition
#   is met first will trigger the rewrite.
# - Recommended: Set this to at least 2x your typical working dataset size
```

kronwerk:
> Or, if the dataset grows over time, do we need to continuously adjust this configuration setting dynamically?

@enjoy-binbin the idea was that you don't - because we have a real limit of our current disk size, and if it's not increasing dynamically in real-time it doesn't make too much sense

so most static disk disk sizes are 2-3x of of memory (or maxmemory, it depends for different users, as I've seen) - and after some limit chances that AOF rewrite succeeds are very low. i.e. 90% disk filled means that your rewrite is likely to be failed - but if you're sure that for your situation (a lot of TTL and/or deletion) this limit is ok, you just set it for the first time or reset it later manually

to change that automatically with the considerations above we need to track recent AOF rewrite speed and/or efficiency cause it gets influenced by the recent write RPS - not sure if we want to implement that, hmmm

murphyjacob4:
I think generally this makes sense. Disks aren't infinite and this is a more natural way to express that limit to Valkey.

>  if the dataset is indeed that large, what should we do?

@kronwerk we definitely should find some way to handle this so we don't constantly sit in a compaction loop. The percentage based limit did not hit this problem since the bigger your base file grew, the less frequent compaction happens. With this approach, the bigger your base file, the more frequent the compaction. As you approach the limit, compaction will happen basically every write.

Perhaps we can detect when the base file hits, say 75% of the specified limit, and turn the soft limit off in such cases?

kronwerk:
@murphyjacob4 

> With this approach, the bigger your base file, the more frequent the compaction.

yeah, after exceeding this limit it becomes something best-effort attempt to save our disk from filling
`server.aof_current_size >= server.aof_rewrite_max_size`

> As you approach the limit, compaction will happen basically every write.

not really like that all the time - but at some point it could be from time to time, yeah

> Perhaps we can detect when the base file hits, say 75% of the specified limit, and turn the soft limit off in such cases?

I guess we do something more or less like you proposed by this check:
`server.aof_rewrite_base_size < server.aof_rewrite_max_size`

i.e. we intensify our attempts in this window:
BASE < LIMIT < AOF

and when it breaks out successfully down to this window:
BASE < AOF < LIMIT - we stop trying, just living with the usual percentage, if it's turned on

and when it goes out unhappily up to this window:
LIMIT < BASE < AOF - we stop trying, knowing at this time that our disk gonna be almost always filled (but in some cases percentage rewrite gonna save us, of course, if a lot of data would be deleted/expired, but that's just a lucky coincidence for us here)

the only real further improvement here I might see really is like I've mentioned earlier - to track writing RPS and AOF rewrite speed and make some rewrite decisions based on that. that would be some expert system guess style - not sure if this complex approach is viable for this kind of task at all really

murphyjacob4:
> server.aof_rewrite_base_size < server.aof_rewrite_max_size

Ah I see. I guess I would mostly be concerned about the period where we run up towards server.aof_rewrite_max_size. It would be good to leave a minimum gap between `server.aof_rewrite_base_size` and `server.aof_rewrite_max_size` to prevent over-frequent compaction.

So maybe, only trigger `server.aof_rewrite_max_size` based compaction if `server.aof_rewrite_max_size-server.aof_rewrite_base_size>server.aof_rewrite_min_size` (or some other minimum incremental file space config).

PR #2375, opened 2025-07-23
config option to stop master on AOF write error

kronwerk:
when we have a primary disk filled with AOF we might finally have it stalled forever in that state - in this PR the config option is added to kill master having enough good replicas on such occasion (with recently added failover attempt for cluster-version reused)

### DIFF
--- a/src/aof.c
+++ b/src/aof.c
@@ -1201,6 +1201,7 @@ void flushAppendOnlyFile(int force) {
                           "Can't recover from AOF write error when replicas-to-fail-on-aof-short-write is '%d' "
                           "and there are enough actual replicas. Exiting...",
                           server.replicas_to_fail_on_aof_short_write);
+                if (server.cluster_enabled) clusterAutoFailoverOnShutdown();
                 exit(1);
             }
 
--- a/src/aof.c
+++ b/src/aof.c
@@ -1196,12 +1196,16 @@ void flushAppendOnlyFile(int force) {
                       "Can't recover from AOF write error when the AOF fsync policy is 'always'. Exiting...");
             exit(1);
         } else {
-            if (getGoodReplicasCount(0) >= server.replicas_to_fail_on_aof_short_write) {
+            if (server.replicas_to_fail_on_aof_write_error >= 0 &&
+                getGoodReplicasCount(0) >= server.replicas_to_fail_on_aof_write_error) {
                 serverLog(LL_WARNING,
-                          "Can't recover from AOF write error when replicas-to-fail-on-aof-short-write is '%d' "
+                          "Can't recover from AOF write error when replicas-to-fail-on-aof-write-error is '%d' "
                           "and there are enough actual replicas. Exiting...",
-                          server.replicas_to_fail_on_aof_short_write);
-                if (server.cluster_enabled) clusterAutoFailoverOnShutdown();
+                          server.replicas_to_fail_on_aof_write_error);
+                if (server.cluster_enabled) {
+                    usleep(server.cluster_node_timeout * 1000);
+                    clusterAutoFailoverOnShutdown();
+                }
                 exit(1);
             }
 
--- a/src/aof.c
+++ b/src/aof.c
@@ -1196,6 +1196,14 @@ void flushAppendOnlyFile(int force) {
                       "Can't recover from AOF write error when the AOF fsync policy is 'always'. Exiting...");
             exit(1);
         } else {
+            if (getGoodReplicasCount(0) >= server.replicas_to_fail_on_aof_short_write) {
+                serverLog(LL_WARNING,
+                          "Can't recover from AOF write error when replicas-to-fail-on-aof-short-write is '%d' "
+                          "and there are enough actual replicas. Exiting...",
+                          server.replicas_to_fail_on_aof_short_write);
+                exit(1);
+            }
+
             /* Recover from failed write leaving data into the buffer. However
              * set an error to stop accepting writes as long as the error
              * condition is not cleared. */
--- a/src/aof.c
+++ b/src/aof.c
@@ -1196,17 +1196,14 @@ void flushAppendOnlyFile(int force) {
                       "Can't recover from AOF write error when the AOF fsync policy is 'always'. Exiting...");
             exit(1);
         } else {
-            if (server.replicas_to_fail_on_aof_write_error >= 0 &&
-                getGoodReplicasCount(0) >= server.replicas_to_fail_on_aof_write_error) {
-                serverLog(LL_WARNING,
-                          "Can't recover from AOF write error when replicas-to-fail-on-aof-write-error is '%d' "
-                          "and there are enough actual replicas. Exiting...",
-                          server.replicas_to_fail_on_aof_write_error);
-                if (server.cluster_enabled) {
-                    usleep(server.cluster_node_timeout * 1000);
-                    clusterAutoFailoverOnShutdown();
+            if (server.shutdown_on_aof_error) {
+                int flags = SHUTDOWN_RO | SHUTDOWN_FORCE | SHUTDOWN_FAILOVER;
+                if (prepareReplicasForShutdown(flags) == C_OK) {
+                    serverLog(LL_WARNING, "Can't recover from AOF write error with shutdown-on-aof-error=yes. Exiting...");
+                    exit(1);
+                } else {
+                    return; /* Shutdown gonna be processed in serverCron. */
                 }
-                exit(1);
             }
 
             /* Recover from failed write leaving data into the buffer. However
--- a/src/replication.c
+++ b/src/replication.c
@@ -4842,24 +4842,28 @@ void replicationResurrectProvisionalPrimary(void) {
 
 /* ------------------------- MIN-REPLICAS-TO-WRITE  --------------------------- */
 
-/* This function counts the number of replicas with lag <= min-replicas-max-lag.
- * If the option is active, the server will prevent writes if there are not
- * enough connected replicas with the specified lag (or less). */
-void refreshGoodReplicasCount(void) {
+int getGoodReplicasCount(int max_lag) {
     listIter li;
     listNode *ln;
     int good = 0;
-
-    if (!server.repl_min_replicas_to_write || !server.repl_min_replicas_max_lag) return;
 
     listRewind(server.replicas, &li);
     while ((ln = listNext(&li))) {
         client *replica = ln->value;
         time_t lag = server.unixtime - replica->repl_data->repl_ack_time;
 
-        if (replica->repl_data->repl_state == REPLICA_STATE_ONLINE && lag <= server.repl_min_replicas_max_lag) good++;
+        if (replica->repl_data->repl_state == REPLICA_STATE_ONLINE && lag <= max_lag) good++;
     }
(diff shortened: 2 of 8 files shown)


valkey.conf
```diff
@@ -1696,6 +1696,10 @@ aof-use-rdb-preamble yes
 # the AOF format in a way that may not be compatible with existing AOF parsers.
 aof-timestamp-enabled no
 
+# Fail primary on AOF short write error if there are this replicas count with 0 lag.
+# -1 (default) prevents failing.
```
murphyjacob4: It looks like we allow 0 as a value. What does that mean? Always exit on short write, regardless of the number of good replicas?

Maybe we should use 0 as the disabled value, unless the 0 case has some use case I am missing.

valkey.conf
```diff
@@ -1696,6 +1696,10 @@ aof-use-rdb-preamble yes
...
+replicas-to-fail-on-aof-short-write -1
```
murphyjacob4: Unless I'm missing something - it looks like this triggers for all AOF write failures - not just short write - right?

src/aof.c
```diff
@@ -1196,6 +1196,15 @@ void flushAppendOnlyFile(int force) {
                       "Can't recover from AOF write error when the AOF fsync policy is 'always'. Exiting...");
             exit(1);
         } else {
+            if (getGoodReplicasCount(0) >= server.replicas_to_fail_on_aof_short_write) {
```
murphyjacob4: I'm thinking the use case here is mostly as an escape from disk full or some non-recoverable disk issue that leaves the primary not accepting writes indefinitely.

Instead of just immediately failing over when we encounter a disk error or short write - would it make sense to wait for some amount of time before doing the failover?

I am worried about us being too trigger happy with failover. We do already have cluster-node-timeout which we could use

src/replication.c
```diff
@@ -4516,24 +4516,28 @@ void replicationResurrectProvisionalPrimary(void) {
...
     listIter li;
     listNode *ln;
     int good = 0;
 
-    if (!server.repl_min_replicas_to_write || !server.repl_min_replicas_max_lag) return;
-
     listRewind(server.replicas, &li);
     while ((ln = listNext(&li))) {
         client *replica = ln->value;
         time_t lag = server.unixtime - replica->repl_data->repl_ack_time;
```
murphyjacob4, line 4853: Lag here means "how long (in seconds) since our last REPLCONF ACK" response

If we are testing lag==0, that doesn't necessarily guarantee that the replica is fully caught up. It just means that we have gotten a REPLCONF ACK on the current second (this could have been up to 999ms ago in the worst case).

Should we compare the replication offset instead?

murphyjacob4 (commented):
Thanks for the PR! I think there is definitely a gap here - thanks for identifying it and putting together a solution. I have a few comments about the approach, please take a look!

valkey.conf
kronwerk: yeah - I've thought about 0 replicas case. in this case 0 means "stop", -1 means "stall" - we might have different infrastructure installations which make reasonable both cases in different situations

valkey.conf
kronwerk: yeah, I think you're right - naming is a bit misleading, let's rename it

src/aof.c
kronwerk: seems reasonable - added sleep before failover for cluster_enabled case

src/replication.c
kronwerk: @murphyjacob4 hmmm, this was introduced 12 years ago - maybe could be improved.
image

not sure though about the right way to do it - in my opinion repl_min_replicas_max_lag is a way to allow considering a replica "good" even if there is a set-by-user lag.

if we want to switch to offsets - what should we do? ignore repl_min_replicas_max_lag, deprecate it, introduce repl_min_replicas_max_offset? would it be convenient and clear for a user which offset is enough to use for the same use case? or repl_min_replicas_max_offset should be always 0 (equal offsets for primary and replica) - but in this case we lose this functionality (allowing to consider lagged replica "good"), is it ok for us?

src/aof.c
murphyjacob4: By "wait for some amount of time" - I was thinking:

1. Record the time when we first encounter AOF write failure
2. If after X amount of time, attempts to write to AOF are still failing, do the failover

I don't think `usleep` will really help since we won't actually retry the write failure

src/replication.c
murphyjacob4: I think the context is a little different:

* repl_min_replicas_max_lag/repl_min_replicas_to_write seem more for "maximum tolerable write loss". Basically, when an unavoidable issue occurs and a failover happens, how many seconds of data are you okay losing
* Here - I think we are making a decision instead that is more like - when the disk gets full, or disk becomes unavailable - what conditions should be met before we demote ourselves (kind of like a manual failover)

My thinking was that since we are in control, why not just wait for repl offset to hit 0, and guarantee that all our data is synced with the replica? We could even send a REPLCONF GETACK to speed this up.

kronwerk: @murphyjacob4 ah. I see. I've rewritten this option according to your comment - reusing currently existing code to shutdown with GETACK and FAILOVER for cluster mode within shutdown-timeout now. PTAL

kronwerk:
@valkey-io/core-team PTAL

src/replication.c
kronwerk: @murphyjacob4 hi, PTAL

kronwerk:
@murphyjacob4 or may be @ranshid could you PTAL? thx

kronwerk:
@enjoy-binbin could you PTAL? I've seen your activity in some AOF review recently

enjoy-binbin (commented):
I don't quite understand the use case.

> when we have a primary disk filled with AOF we might finally have it stalled forever in that state

BTW, what do you mean this `stalled forever in that state` part? You mean the server is still running but not able to write the AOF, and we should exit? You want something like `stop-writes-on-aof-error`?
```
int writeCommandsDeniedByDiskError(void) {
    if (server.stop_writes_on_bgsave_err && server.saveparamslen > 0 && server.lastbgsave_status == C_ERR) {
        return DISK_ERROR_TYPE_RDB;
    } else if (server.aof_state != AOF_OFF) {
        if (server.aof_last_write_status == C_ERR) {
            return DISK_ERROR_TYPE_AOF;
        }
        /* AOF fsync error. */
        int aof_bio_fsync_status = atomic_load_explicit(&server.aof_bio_fsync_status, memory_order_acquire);
        if (aof_bio_fsync_status == C_ERR) {
            server.aof_last_write_errno = atomic_load_explicit(&server.aof_bio_fsync_errno, memory_order_relaxed);
            return DISK_ERROR_TYPE_AOF;
        }
    }

    return DISK_ERROR_TYPE_NONE;
}
```

kronwerk:
> I don't quite understand the use case.
> 
> > when we have a primary disk filled with AOF we might finally have it stalled forever in that state

yeah, there're cases when Valkey is unable to get out of this state ever - even if we add some space and try to start BGREWRITEAOF, it fails and continues to throw errors on any interaction attempts. not sure if I can thoroughly design such a test case, but I've seen it in production

that's why I offer a setting to stop master here - it seems better than this behaviour

> You want something like `stop-writes-on-aof-error`?

I'm not sure if that's gonna fix the situation described - but sounds like something that could help, yeah, what do you think?
@enjoy-binbin

enjoy-binbin:
Currently, if AOF is enabled and the AOF write fails and cannot be recovered, we will stop client writes.
```
sds writeCommandsGetDiskErrorMessage(int error_code) {
    sds ret = NULL;
    if (error_code == DISK_ERROR_TYPE_RDB) {
        ret = sdsdup(objectGetVal(shared.bgsaveerr));
    } else {
        ret = sdscatfmt(sdsempty(), "-MISCONF Errors writing to the AOF file: %s\r\n",
                        strerror(server.aof_last_write_errno));
    }
    return ret;
}

127.0.0.1:6379> set k v
(error) MISCONF Errors writing to the AOF file: xxx
```

In this case, even if we don't have `stop-writes-on-aof-error`, it should still be sufficient. The behavior is similar to `stop-writes-on-bgsave-error`.
```
# By default the server will stop accepting writes if RDB snapshots are enabled
# (at least one save point) and the latest background save failed.
# This will make the user aware (in a hard way) that data is not persisting
# on disk properly, otherwise chances are that no one will notice and some
# disaster will happen.
#
# If the background saving process will start working again, the server will
# automatically allow writes again.
#
# However if you have setup your proper monitoring of the server
# and persistence, you may want to disable this feature so that the server will
# continue to work as usual even if there are problems with disk,
# permissions, and so forth.
stop-writes-on-bgsave-error yes
```

> yeah, there're cases when Valkey is unable to get out of this state ever - even if we add some space and try to start BGREWRITEAOF, it fails and continues to throw errors on any interaction attempts

Do you know the fail reason? From the server logs? We will print the log if write failed

kronwerk:
> Do you know the fail reason? From the server logs? We will print the log if write failed

@enjoy-binbin I've tried to find any logs from this incident we've been seen someday - but sadly it's long gone, of course.

but to make the idea of this PR a bit more clear, I hope: this is not a try-to-fix behaviour, this seems to be a killswitch. because increasing disk sometimes doesn't help (no proofs found) - but sometimes it's impossible, this case seems to be fair without reproduce attempts. so if you can somewhat safely restart master to clean the disk as a last resort and you're sure about that enough to set this option on (off by default), you can do it

of course parallel RDB write for preambule and/or make (re)writing AOF faster is better than this - but it's unavailable yet, so fixing stalled master is needed from time to time. and it can be done with some coordinator scipt outside (but we should be sure that the situation is ok) or with some kind of inner (this PR) + outer script to just start stopped master back without dangerous guess about server state from outside the server

murphyjacob4:
To resurrect this: I think the case here is:

1) AOF writes are failing/stalling continously
2) We have a replica that can takeover
3) Let's do a graceful failover to the replica

This happens in some cases when you have a network attached disk like AWS's EBS or GCP's PD. An issue in the dependency service can leave you in a non-recoverable state where the disk is down for a very long time.

> Currently, if AOF is enabled and the AOF write fails and cannot be recovered, we will stop client writes.

IMO - just stopping client writes isn't enough if we aren't going to relinquish primaryship.

I do wonder - if we don't have a replica - is it worth shutting down still?

src/aof.c
```diff
@@ -1196,6 +1196,16 @@ void flushAppendOnlyFile(int force) {
                       "Can't recover from AOF write error when the AOF fsync policy is 'always'. Exiting...");
             exit(1);
         } else {
+            if (server.shutdown_on_aof_error) {
+                int flags = SHUTDOWN_RO | SHUTDOWN_FORCE | SHUTDOWN_FAILOVER;
+                if (prepareReplicasForShutdown(flags) == C_OK) {
+                    serverLog(LL_WARNING, "Can't recover from AOF write error with shutdown-on-aof-error=yes. Exiting...");
+                    exit(1);
```
murphyjacob4, line 1203: Will this actually do the failover? I think you need to call `finishShutdown` in that case. 

Probably we can swap `prepareReplicasForShutdown` for `prepareForShutdown` which also calls `finishShutdown`.

PR #2378, opened 2025-07-24
support negative filtering for client command filters

### LINKED ISSUE #1936: [NEW] Negative filtering for client command filters
does this PR support negative filtering?

_Originally posted by @soloestoy in https://github.com/valkey-io/valkey/issues/1466#issuecomment-2647089232_

soloestoy:
introduce negative filters for `CLIENT LIST` and `CLIENT KILL` commands:

1. `NOT-ID`: Excludes clients in the IDs set
2. `NOT-TYPE`: Excludes clients of the specified type
3. `NOT-ADDR`: Excludes clients of the specified address and port
4. `NOT-LADDR`: Excludes clients connected to the specified local address and port
5. `NOT-USER`: Excludes clients of the specified user
6. `NOT-FLAGS`: Excludes clients with the specified flag string
7. `NOT-NAME`: Excludes clients with the specified name
8. `NOT-LIB-NAME`: Excludes clients using the specified library name
9. `NOT-LIB-VER`: Excludes clients with the specified library version
10. `NOT-DB`: Excludes clients with the specified database ID
11. `NOT-CAPA`: Excludes clients with the specified capabilities
12. `NOT-IP`: Excludes clients with the specified IP address

close #1936

and fix the matching algorithm for flag 'N'.

### DIFF
--- a/src/commands.def
+++ b/src/commands.def
@@ -1334,7 +1334,7 @@ commandHistory CLIENT_KILL_History[] = {
 {"8.0.0","`MAXAGE` option."},
 {"8.0.0","Replaced `master` `TYPE` with `primary`. `master` still supported for backward compatibility."},
 {"8.1.0","`ID` option accepts multiple IDs."},
-{"9.0.0","Added filters NAME, IDLE, FLAGS, LIB-NAME, LIB-VER, DB, CAPA, and IP. And negative filters NOT-ID, NOT-MAXAGE, NOT-TYPE, NOT-ADDR, NOT-LADDR, NOT-USER, NOT-IDLE, NOT-FLAGS, NOT-NAME, NOT-LIB-NAME, NOT-LIB-VER, NOT-DB, NOT-CAPA, NOT-IP."},
+{"9.0.0","Added filters NAME, IDLE, FLAGS, LIB-NAME, LIB-VER, DB, CAPA, and IP. And negative filters NOT-ID, NOT-TYPE, NOT-ADDR, NOT-LADDR, NOT-USER, NOT-FLAGS, NOT-NAME, NOT-LIB-NAME, NOT-LIB-VER, NOT-DB, NOT-CAPA, NOT-IP."},
 };
 #endif
 
@@ -1396,9 +1396,7 @@ struct COMMAND_ARG CLIENT_KILL_filter_new_format_Subargs[] = {
 {MAKE_ARG("not-username",ARG_TYPE_STRING,-1,"NOT-USER",NULL,"9.0.0",CMD_ARG_OPTIONAL,0,NULL)},
 {MAKE_ARG("not-addr",ARG_TYPE_STRING,-1,"NOT-ADDR",NULL,"9.0.0",CMD_ARG_OPTIONAL,0,NULL),.display_text="ip:port"},
 {MAKE_ARG("not-laddr",ARG_TYPE_STRING,-1,"NOT-LADDR",NULL,"9.0.0",CMD_ARG_OPTIONAL,0,NULL),.display_text="ip:port"},
-{MAKE_ARG("not-maxage",ARG_TYPE_INTEGER,-1,"NOT-MAXAGE",NULL,"9.0.0",CMD_ARG_OPTIONAL,0,NULL)},
 {MAKE_ARG("not-name",ARG_TYPE_STRING,-1,"NOT-NAME",NULL,"9.0.0",CMD_ARG_OPTIONAL,0,NULL)},
-{MAKE_ARG("not-idle",ARG_TYPE_INTEGER,-1,"NOT-IDLE",NULL,"9.0.0",CMD_ARG_OPTIONAL,0,NULL)},
 {MAKE_ARG("not-flags",ARG_TYPE_STRING,-1,"NOT-FLAGS",NULL,"9.0.0",CMD_ARG_OPTIONAL,0,NULL)},
 {MAKE_ARG("not-lib-name",ARG_TYPE_STRING,-1,"NOT-LIB-NAME",NULL,"9.0.0",CMD_ARG_OPTIONAL,0,NULL)},
 {MAKE_ARG("not-lib-ver",ARG_TYPE_STRING,-1,"NOT-LIB-VER",NULL,"9.0.0",CMD_ARG_OPTIONAL,0,NULL)},
@@ -1410,7 +1408,7 @@ struct COMMAND_ARG CLIENT_KILL_filter_new_format_Subargs[] = {
 /* CLIENT KILL filter argument table */
 struct COMMAND_ARG CLIENT_KILL_filter_Subargs[] = {
 {MAKE_ARG("old-format",ARG_TYPE_STRING,-1,NULL,NULL,NULL,CMD_ARG_NONE,0,"2.8.12"),.display_text="ip:port"},
-{MAKE_ARG("new-format",ARG_TYPE_ONEOF,-1,NULL,NULL,NULL,CMD_ARG_MULTIPLE,29,NULL),.subargs=CLIENT_KILL_filter_new_format_Subargs},
+{MAKE_ARG("new-format",ARG_TYPE_ONEOF,-1,NULL,NULL,NULL,CMD_ARG_MULTIPLE,27,NULL),.subargs=CLIENT_KILL_filter_new_format_Subargs},
 };
 
 /* CLIENT KILL argument table */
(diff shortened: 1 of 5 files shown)


src/commands.def
```diff
@@ -1334,7 +1334,7 @@ commandHistory CLIENT_KILL_History[] = {
 {"8.0.0","`MAXAGE` option."},
 {"8.0.0","Replaced `master` `TYPE` with `primary`. `master` still supported for backward compatibility."},
 {"8.1.0","`ID` option accepts multiple IDs."},
-{"9.0.0","Added filters NAME, IDLE, FLAGS, LIB-NAME, LIB-VER, DB, CAPA, and IP."},
+{"9.0.0","Added filters NAME, IDLE, FLAGS, LIB-NAME, LIB-VER, DB, CAPA, and IP. And negative filters NOT-ID, NOT-MAXAGE, NOT-TYPE, NOT-ADDR, NOT-LADDR, NOT-USER, NOT-IDLE, NOT-FLAGS, NOT-NAME, NOT-LIB-NAME, NOT-LIB-VER, NOT-DB, NOT-CAPA, NOT-IP."},
```
madolson:
```suggestion
{"9.0.0","Added filters NAME, IDLE, FLAGS, LIB-NAME, LIB-VER, DB, CAPA, and IP. And negative filters NOT-ID, NOT-MAXAGE, NOT-TYPE, NOT-ADDR, NOT-LADDR, NOT-USER, NOT-IDLE, NOT-FLAGS, NOT-NAME, NOT-LIB-NAME, NOT-LIB-VER, NOT-DB, NOT-CAPA, NOT-IP."},
```
NOT-MAXAGE and NOT-IDLE are confusing with the NOT prefix, and I also don't see the use case. Why would you only want to kill clients that are active and/or only kill clients that are the youngest. We made some bad name choices in the past, so unless there is a strong reason to include them I would prefer to skip them.

tests/unit/introspection.tcl
```diff
@@ -1362,6 +1634,58 @@ start_server {tags {"introspection"}} {
...
+        r client list not-lib-ver 1.2.3
+    } {}
+
+    test {CLIENT LIST can filter by NOT-DB number} {
+        r select 2
+        r client list not-db 2
+    } {} {external:skip}
+
+    test {CLIENT KILL can filter by NOT-DB} {
+        set c1 [valkey_client]
```
madolson, line 1649: c1 isn't cleaned up and the end of this function or in the next two.

tests/unit/introspection.tcl
```diff
@@ -380,6 +382,260 @@ start_server {tags {"introspection"}} {
...
+
+    test {CLIENT KILL with multiple negative filters including idle time} {
+        # Create two clients
+        set c1 [valkey_client]
+        set c2 [valkey_client]
+        $c1 client setname client1
+        $c2 client setname client2
+
+        # Wait 1 second to ensure idle time
+        after 1000 ;# Wait 1 second
```
madolson:
```suggestion
        # Wait 1 second to ensure idle time
        after 1000
```

tests/unit/introspection.tcl
```diff
@@ -380,6 +382,260 @@ start_server {tags {"introspection"}} {
...
+        set c1 [valkey_client]
+        set c2 [valkey_client]
+        set c3 [valkey_client]
+        $c1 client setname client1
+        $c2 client setname client1
+        $c3 client setname client2
+        $c3 multi
+
+        # Wait 1 second to ensure idle time
+        after 1000  ;
```
madolson:
```suggestion
        after 1000
```

tests/unit/introspection.tcl
```diff
@@ -380,6 +382,260 @@ start_server {tags {"introspection"}} {
...
+
+        # Kill client by NOT-IP
+        r client kill not-ip "1.2.3.4" skipme yes
+
+        assert_error "*I/O error*" {$c1 ping}
+        catch {$c1 close}
+
+        set cl [r client list]
+        assert_match "*name=client-normal*" $cl
+        assert_no_match "*name=killme-not-ip*" $cl
```
madolson:
```suggestion
        set list_response [r client list]
        assert_match "*name=client-normal*" $list_response
        assert_no_match "*name=killme-not-ip*" $list_response
```
cl and c1 look similar, just suggesting any name that looks more different. the `cl` name is used a few more places later in this same file.

src/networking.c
```diff
@@ -4444,15 +4585,15 @@ static int clientMatchesFlagFilter(client *c, sds flag_filter) {
...
-            if (!c->flag.replica && !c->flag.primary && !c->flag.pubsub &&
-                !c->flag.multi && !c->flag.blocked && !c->flag.tracking &&
-                !c->flag.tracking_broken_redir && !c->flag.tracking_bcast &&
-                !c->flag.dirty_cas && !c->flag.close_after_reply &&
-                !c->flag.unblocked && !c->flag.close_asap &&
-                !c->flag.unix_socket && !c->flag.readonly &&
-                !c->flag.no_evict && !c->flag.no_touch &&
-                !c->flag.import_source) {
-                return 1; /* Matches 'N' */
+            if (c->flag.replica || c->flag.primary || c->flag.pubsub ||
```
madolson, line 4558: Is there a reason you changed this

madolson (commented):
One high level comment, but the code looked mostly good. I liked some of the other improvements you made as well.

src/networking.c
sarthakaggarwal97: yes, I'd also like to know this.

src/commands.def
soloestoy: These two are only for symmetry and completeness of the filters. NOT-MAXAGE and NOT-IDLE are not ideal names, maybe MINAGE and ACTIVE would be better. We can revisit this discussion when there is actual user request in the future.

src/networking.c
soloestoy: it is a bugfix, witch the bug can lead to the test case fails:

```tcl
    # Test CLIENT LIST with NOT-FLAGS filter
    test {CLIENT LIST with NOT-FLAGS filter} {
        set c1 [valkey_client]
        $c1 readonly
        set cl [r client list not-flags N]
        assert_match "*flags=r*" $cl
        assert_no_match "*flags=N*" $cl
        catch {$c1 close}
    }
```

with the old code, the READONLY client cannot match the condition:
```c
        case 'N': /* Check for no flags */
            if (!c->flag.replica && !c->flag.primary && !c->flag.pubsub &&
                !c->flag.multi && !c->flag.blocked && !c->flag.tracking &&
                !c->flag.tracking_broken_redir && !c->flag.tracking_bcast &&
                !c->flag.dirty_cas && !c->flag.close_after_reply &&
                !c->flag.unblocked && !c->flag.close_asap &&
                !c->flag.unix_socket && !c->flag.readonly &&
                !c->flag.no_evict && !c->flag.no_touch &&
                !c->flag.import_source) {
                return 1; /* Matches 'N' */
            }
            break;
        default:
            /* Invalid flag, return false */
            return 0;
        }
    }
    /* If the loop completes, the client matches the flag filter */
    return 1;
}
```
but return the wrong "1" after the loop completes.

yangbodong22011 (approved):

src/networking.c
```diff
@@ -4444,15 +4585,15 @@ static int clientMatchesFlagFilter(client *c, sds flag_filter) {
...
+                c->flag.multi || c->flag.blocked || c->flag.tracking ||
+                c->flag.tracking_broken_redir || c->flag.tracking_bcast ||
+                c->flag.dirty_cas || c->flag.close_after_reply ||
+                c->flag.unblocked || c->flag.close_asap ||
+                c->flag.unix_socket || c->flag.readonly ||
+                c->flag.no_evict || c->flag.no_touch ||
+                c->flag.import_source) {
+                return 0;
             }
```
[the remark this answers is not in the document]
madolson: This is wrong right? Bad bot.

src/networking.c
madolson: Oh nice!

madolson (approved):
I'm good, thanks!

src/networking.c
sarthakaggarwal97: so basically we are now matching for no flags as well. Makes sense, helps with negative filters.

sarthakaggarwal97 (approved):

sarthakaggarwal97:
@soloestoy it will need similar changes like  #2362 right?

soloestoy:
> @soloestoy it will need similar changes like #2362 right?

it would introduce breaking change and needs further discussion, we can decide later

PR #2379, opened 2025-07-26
Defrag should only skip non-existent dbs, not empty dbs

enjoy-binbin:
In #1609, we now doing on-demand database allocation instead
of preallocation. And in beginDefragCycle, we should not skip
empty databases if they exist. The defrag still defrags the
internal allocations of the hashtables structs if they exist.

Call chain: defragStageDbKeys -> defragStageKvstoreHelper ->
kvstoreHashtableDefragTables -> hashtableDefragTables.


zuiderkwast (approved):

xbasel:
lgtm

PR #2385, opened 2025-07-28
Allow dynamic modification of databases num if the db is not been used

enjoy-binbin:
Once we no longer pre-allocate server.db (see #1609), we may be able
to provide the ability to dynamically update databases. Being able to
dynamically grow it seems very attractive to me, but i am not sure
dynamically shrinking it.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -2490,6 +2490,27 @@ static int updatePort(const char **err) {
     return 1;
 }
 
+static int updateDatabases(const char **err) {
+    int old_dbnum = server.dbnum;
+    int new_dbnum = server.cluster_enabled ? server.config_databases_cluster : server.config_databases;
+    if (old_dbnum == new_dbnum) return 1;
+
+    /* If it is adjusted smaller, then we need to check whether the
+     * database has been used. */
+    if (new_dbnum < old_dbnum) {
+        for (int dbid = new_dbnum; dbid < old_dbnum; dbid++) {
+            if (server.db[dbid] != NULL) {
+                *err = "The database has been used and cannot be adjusted "
+                       "to a smaller number.";
+                return 0;
+            }
+        }
+    }
+    server.dbnum = new_dbnum;
+    server.db = zrealloc(server.db, sizeof(serverDb *) * server.dbnum);
+    return 1;
+}
+
 static int updateDefragConfiguration(const char **err) {
     UNUSED(err);
     server.active_defrag_configuration_changed = 1;
@@ -3265,8 +3286,8 @@ standardConfig static_configs[] = {
     createEnumConfig("rdb-version-check", NULL, MODIFIABLE_CONFIG, rdb_version_check_enum, server.rdb_version_check, RDB_VERSION_CHECK_STRICT, NULL, NULL),
 
     /* Integer configs */
-    createIntConfig("databases", NULL, IMMUTABLE_CONFIG, 1, INT_MAX, server.config_databases, 16, INTEGER_CONFIG, NULL, NULL),
-    createIntConfig("cluster-databases", NULL, IMMUTABLE_CONFIG, 1, INT_MAX, server.config_databases_cluster, 1, INTEGER_CONFIG, NULL, NULL),
+    createIntConfig("databases", NULL, MODIFIABLE_CONFIG, 1, INT_MAX, server.config_databases, 16, INTEGER_CONFIG, NULL, updateDatabases),
+    createIntConfig("cluster-databases", NULL, MODIFIABLE_CONFIG, 1, INT_MAX, server.config_databases_cluster, 1, INTEGER_CONFIG, NULL, updateDatabases),
     createIntConfig("port", NULL, MODIFIABLE_CONFIG, 0, 65535, server.port, 6379, INTEGER_CONFIG, NULL, updatePort),                                               /* TCP port. */
     createIntConfig("io-threads", NULL, DEBUG_CONFIG | MODIFIABLE_CONFIG, 1, IO_THREADS_MAX_NUM, server.io_threads_num, 1, INTEGER_CONFIG, NULL, updateIOThreads), /* Single threaded by default */
     createIntConfig("events-per-io-thread", NULL, MODIFIABLE_CONFIG | HIDDEN_CONFIG, 0, INT_MAX, server.events_per_io_thread, 2, INTEGER_CONFIG, NULL, NULL),


enjoy-binbin:
This is just a draft that popped up, i will improve it or add more tests later once we are on the same page.

src/config.c
```diff
@@ -2490,6 +2490,27 @@ static int updatePort(const char **err) {
...
+        for (int dbid = new_dbnum; dbid < old_dbnum; dbid++) {
+            if (server.db[dbid] != NULL) {
+                *err = "The database has been used and cannot be adjusted "
+                       "to a smaller number.";
+                return 0;
+            }
+        }
+    }
+    server.dbnum = new_dbnum;
+    server.db = zrealloc(server.db, sizeof(serverDb *) * server.dbnum);
```
hwware, line 2510: I have 2 concerns for this PR：
1. With which case, we need to update this number?
2.  If new_dbnum < old_dbnum, why we need to allocate new memory to server.db?

hpatro (commented):
In which scenario do you think this would be beneficial? The users can always start a replica with more no. of databases and failover to achieve this.

I'm very skeptical about this change. We're trying to introduce ACL rules at DB level and the default permission for such db needs to be considered well with dynamic resize.

enjoy-binbin:
> In which scenario do you think this would be beneficial? The users can always start a replica with more no. of databases and failover to achieve this.

that is true, if we can simply support it in core, why no? Adding a replica will trigger a full sync, doing a failover will trigger some redirect, some user don't like these process very much.

> I'm very skeptical about this change. We're trying to introduce ACL rules at DB level and the default permission for such db needs to be considered well with dynamic resize.

I get the point. We can only allow it to work in scenarios where ACL-db is not set, until we find some easy ways to do it. (we don't use ACL much)

src/config.c
enjoy-binbin: 1. we want to update the number to use more DBs without other process
2. both are ok, doing realloc look clearer to me

artikell:
If the database (db) between the master and slave is inconsistent, it will trigger a crash. Refer to `RDB_OPCODE_SELECTDB`. Therefore, this operation requires ensuring synchronization between the master and slave.

ranshid:
>This is just a draft that popped up

@enjoy-binbin I suppose in this case maybe open this as a draft and not a ready pull-request?

My thoughts are that although I find this motivation correct, I think that maybe the issue we are trying to solve is more extensive. 

1. Although we introduced lazy allocation of DBs in https://github.com/valkey-io/valkey/pull/1609, we did not take in all the way. For example, we do not free the DB after it was emptied. 
IMO the configuration for dbnum should be deprecated completely and we could support dynamically allocation of databases as they will be freed once not used. 
2. Specifically in case we implement (1) and even without it, We have no good way to distinguish a case were the DB is used (but empty) to when a DB is not used. I think it makes it harder to support changing this parameter when a specific database is in real "use". If (for example) we could have known that the application is using a specific range or number of DBs and we could verify the configuration value when shrinking.
3. We do have plans to support logical names for databases. I think many requests for that came up. I think that following this potential implementation would make it harder to support dynamically changing of this config.

I suggest to look at the next gen databases design as part of a more "holistic" effort to design our "multi-tenancy" support and plan it for Valkey 10 release.

enjoy-binbin:
yes, that is true, i agree with everything.

we long ago support multi db in cluster, and set the databases default to 256, that is a bad move actually. I guess the idea was that if users wanted more dbs they would be able to use them without any adjustments. BUT most user doesn't need this 256 db (people only used db0 at most time since OSS version does not support multi db), and the default 256 dbs cause some toubles, like 1bd0b54957ea71ac69d30d91806936161ac2b65f and memory overhead (#1609 help a lot in this case)

So in my opinion, if modifying it does not increase a lot of complexity, I hope it can be done. (Of course, dynamic named database is also a direction in the future)

> We have no good way to distinguish a case were the DB is used (but empty) to when a DB is not used

that is true, it's easy to increase the number, we don't have this problem. But shrinking it is indeed a problem that we don't know if a db is actually used. So my idea in here is that once the user has used select (that is, db is not NULL), it is assumed that it will always be used and it cannot be shrunk. Like in the current code, if we select all DBs but not actually use it, the memory overhead will still be here.

> I suppose in this case maybe open this as a draft and not a ready pull-request?

yes. It was just an idea at the time and i played with it a bit. I expect it will have these challenges.

ranshid:
> yes, that is true, i agree with everything.
> 
> we long ago support multi db in cluster, and set the databases default to 256, that is a bad move actually. I guess the idea was that if users wanted more dbs they would be able to use them without any adjustments. BUT most user doesn't need this 256 db (people only used db0 at most time since OSS version does not support multi db), and the default 256 dbs cause some toubles, like [1bd0b54](https://github.com/valkey-io/valkey/commit/1bd0b54957ea71ac69d30d91806936161ac2b65f) and memory overhead (#1609 help a lot in this case)
> 
> So in my opinion, if modifying it does not increase a lot of complexity, I hope it can be done. (Of course, dynamic named database is also a direction in the future)
> 
> > We have no good way to distinguish a case were the DB is used (but empty) to when a DB is not used
> 
> that is true, it's easy to increase the number, we don't have this problem. But shrinking it is indeed a problem that we don't know if a db is actually used. So my idea in here is that once the user has used select (that is, db is not NULL), it is assumed that it will always be used and it cannot be shrunk. Like in the current code, if we select all DBs but not actually use it, the memory overhead will still be here.
> 
> > I suppose in this case maybe open this as a draft and not a ready pull-request?
> 
> yes. It was just an idea at the time and i played with it a bit. I expect it will have these challenges.

Wanted to add another thing:
Currently we always treat DBs as an existing array. As I mentioned we have no way to identify which DB is in use, but we also have no efficient way to skip empty DBs and we spend time iterating over them (which I never tested how much can be saved by DO skip them). I guess for a small number of DBs there is no issue, but I think have a BIT support could help. given said that, maybe having a DB container (like the kvstore we could create a dbstore) which will enable such features and we could manage the DB usage and allocation correctly.

I also suggest to run this as a "major-decision"

madolson:
Chatted about this briefly in the core meeting. Folks were in agreement that we should have a better strategy for databases, and we weren't fully convinced that changing this was all that useful. The use case that binbin suggested seems like it would be better mitigated by lazy allocation of memory.

enjoy-binbin:
Closing since we don't need it.

PR #2386, opened 2025-07-28
Fix dual-channel-replication test due to typo error and stabilize it

enjoy-binbin:
It should not be a negative value, otherwise we lose the
meaning of delay and the rdb generation may finish too fast
in a fast machine and cause the test to fail.

Also, the max_tries number is increased and the
loading-process-events-interval-bytes is adjusted so that the
replica has more opportunities to process the events and can
be aware of the disconnection.

### DIFF
--- a/tests/integration/dual-channel-replication.tcl
+++ b/tests/integration/dual-channel-replication.tcl
@@ -1310,7 +1310,7 @@ start_server {tags {"dual-channel-replication external:skip"}} {
         } else {
             $primary config set repl-timeout 10
             $replica config set repl-timeout 10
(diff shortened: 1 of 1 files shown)


hpatro (approved):

hpatro:
First test in the file failed 

```
Replica notice main-connection killed during rdb load callback in tests/integration/dual-channel-replication.tcl
```

enjoy-binbin:
ok, weird, it did passed in my local, and look like CI hang in somewhere (replica did not catch the disconnection during the loading).

naglera (approved):

PR #2390, opened 2025-07-29
Add config for maximum random field count for HRANDFIELD/ZRANDMEMBER/SRANDMEMBER

### LINKED ISSUE #844: [BUG] A large negative number(e.g. -LONG_MAX) in HRANDFIELD/ZRANDMEMBER can easily cause Valkey to crash.
**Describe the bug**

When using the **HRANDFIELD** or **ZRANDMEMBER** commands, if an extremely negative value(e.g. -LONG_MAX) is provided, the memory will continue to rise and eventually cause Redis to crash.

**To reproduce**

HRANDFIELD hash -9223372036854775807

**Expected behavior**

IMHO, it is more reasonable to set a limit on **count** according to the memory limit or the actual size of the hash table to prevent Redis from crashing. This can prevent crashes caused by careless and incorrect use. If a user really needs to sample a lot of elements that allow duplicates, he can achieve the same purpose by calling the command multiple times.

parimoh:
This PR is meant to resolve #844, I got inspiration from https://github.com/Snapchat/KeyDB/pull/639

**Description**

This PR adds a new configuration parameter max-rand-count to limit the maximum number of random elements that can be retrieved using the following commands:

- HRANDFIELD
- SRANDMEMBER
- ZRANDMEMBER

Changes include:

- Added new config parameter max-rand-count with default value of LONG_MAX/2
- Added validation checks in hash, set and zset random field commands
- Added corresponding test cases to verify the new configuration
- Error message "value is out of range" will be returned if count exceeds the configured limit

The default `max_rand_count` of `LONG_MAX/2` is because when using the `withscores` or `withvalues` arguments for `zrandmember` and `hrandfield`, the number of outputs will double. Meaning if a count of less than `-LONG_MAX/2` is entered, it causes the server to crash since it overflows. Either way though, from the issue it sounded like it's reasonable to expect the user to be able to call the command multiple times if they really need more values.

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -3344,6 +3344,7 @@ standardConfig static_configs[] = {
     createLongLongConfig("stream-node-max-entries", NULL, MODIFIABLE_CONFIG, 0, LLONG_MAX, server.stream_node_max_entries, 100, INTEGER_CONFIG, NULL, NULL),
     createLongLongConfig("repl-backlog-size", NULL, MODIFIABLE_CONFIG, 1, LLONG_MAX, server.repl_backlog_size, 10 * 1024 * 1024, MEMORY_CONFIG, NULL, updateReplBacklogSize), /* Default: 10mb */
     createLongLongConfig("cluster-manual-failover-timeout", NULL, MODIFIABLE_CONFIG, 1, INT_MAX, server.cluster_mf_timeout, 5000, INTEGER_CONFIG, NULL, NULL),
+    createLongLongConfig("max-rand-count", NULL, MODIFIABLE_CONFIG, 0, LONG_MAX/2, server.max_rand_count, LONG_MAX/2, INTEGER_CONFIG, NULL, NULL),
 
     /* Unsigned Long Long configs */
     createULongLongConfig("maxmemory", NULL, MODIFIABLE_CONFIG, 0, ULLONG_MAX, server.maxmemory, 0, MEMORY_CONFIG, NULL, updateMaxmemory),
--- a/src/config.c
+++ b/src/config.c
@@ -3344,7 +3344,7 @@ standardConfig static_configs[] = {
     createLongLongConfig("stream-node-max-entries", NULL, MODIFIABLE_CONFIG, 0, LLONG_MAX, server.stream_node_max_entries, 100, INTEGER_CONFIG, NULL, NULL),
     createLongLongConfig("repl-backlog-size", NULL, MODIFIABLE_CONFIG, 1, LLONG_MAX, server.repl_backlog_size, 10 * 1024 * 1024, MEMORY_CONFIG, NULL, updateReplBacklogSize), /* Default: 10mb */
     createLongLongConfig("cluster-manual-failover-timeout", NULL, MODIFIABLE_CONFIG, 1, INT_MAX, server.cluster_mf_timeout, 5000, INTEGER_CONFIG, NULL, NULL),
-    createLongLongConfig("max-rand-count", NULL, MODIFIABLE_CONFIG, 0, LONG_MAX/2, server.max_rand_count, LONG_MAX/2, INTEGER_CONFIG, NULL, NULL),
+    createLongLongConfig("max-rand-count", NULL, MODIFIABLE_CONFIG, 0, LONG_MAX / 2, server.max_rand_count, LONG_MAX / 2, INTEGER_CONFIG, NULL, NULL),
 
     /* Unsigned Long Long configs */
     createULongLongConfig("maxmemory", NULL, MODIFIABLE_CONFIG, 0, ULLONG_MAX, server.maxmemory, 0, MEMORY_CONFIG, NULL, updateMaxmemory),
--- a/src/t_hash.c
+++ b/src/t_hash.c
@@ -1316,16 +1316,13 @@ void hrandfieldCommand(client *c) {
     listpackEntry ele;
 
     if (c->argc >= 3) {
-        if (getRangeLongFromObjectOrReply(c, c->argv[2], -LONG_MAX, LONG_MAX, &l, NULL) != C_OK) return;
-        if (l < -server.max_rand_count || l > server.max_rand_count) {
-            addReplyError(c, "value is out of range");
-            return;
-        }
+        if (getRangeLongFromObjectOrReply(c, c->argv[2], -server.max_rand_count, server.max_rand_count, &l, NULL) != C_OK) return;
         if (c->argc > 4 || (c->argc == 4 && strcasecmp(c->argv[3]->ptr, "withvalues"))) {
             addReplyErrorObject(c, shared.syntaxerr);
             return;
-        } else if (c->argc == 4)
+        } else if (c->argc == 4) {
             withvalues = 1;
+        }
         hrandfieldWithCountCommand(c, l, withvalues);
         return;
     }
--- a/src/t_hash.c
+++ b/src/t_hash.c
@@ -1317,16 +1317,14 @@ void hrandfieldCommand(client *c) {
 
     if (c->argc >= 3) {
         if (getRangeLongFromObjectOrReply(c, c->argv[2], -LONG_MAX, LONG_MAX, &l, NULL) != C_OK) return;
+        if (l < -server.max_rand_count || l > server.max_rand_count) {
+            addReplyError(c, "value is out of range");
+            return;
+        }
         if (c->argc > 4 || (c->argc == 4 && strcasecmp(c->argv[3]->ptr, "withvalues"))) {
             addReplyErrorObject(c, shared.syntaxerr);
             return;
-        } else if (c->argc == 4) {
-            withvalues = 1;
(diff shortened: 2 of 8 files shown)


src/t_hash.c
```diff
@@ -1317,16 +1317,14 @@ void hrandfieldCommand(client *c) {
...
+        }
         if (c->argc > 4 || (c->argc == 4 && strcasecmp(c->argv[3]->ptr, "withvalues"))) {
             addReplyErrorObject(c, shared.syntaxerr);
             return;
-        } else if (c->argc == 4) {
-            withvalues = 1;
-            if (l < -LONG_MAX / 2 || l > LONG_MAX / 2) {
-                addReplyError(c, "value is out of range");
-                return;
-            }
```
parimoh, line 1328: This `else if` condition is only true when the user includes `withvalues`. Therefore I moved the range validation to outside of this because we want it to be checked whenever the number of arguments is `>= 3` (aka whenever the user includes a count).

src/t_zset.c
```diff
@@ -4246,16 +4246,14 @@ void zrandmemberCommand(client *c) {
...
         if (c->argc > 4 || (c->argc == 4 && strcasecmp(c->argv[3]->ptr, "withscores"))) {
             addReplyErrorObject(c, shared.syntaxerr);
             return;
-        } else if (c->argc == 4) {
-            withscores = 1;
-            if (l < -LONG_MAX / 2 || l > LONG_MAX / 2) {
-                addReplyError(c, "value is out of range");
-                return;
-            }
-        }
```
parimoh: The same logic as the above comment applies here.

src/t_set.c
```diff
@@ -1009,6 +1009,10 @@ void srandmemberWithCountCommand(client *c) {
     int64_t llele;
 
     if (getRangeLongFromObjectOrReply(c, c->argv[2], -LONG_MAX, LONG_MAX, &l, NULL) != C_OK) return;
+    if (l < -server.max_rand_count || l > server.max_rand_count) {
+        addReplyError(c, "value is out of range");
+        return;
+    }
```
parimoh: Question: Should I make a common function for this logic or is it too small for it to matter? Saw in other similar PRs that others haven't so I did the same as them for now.

Nikhil-Manglore: It's fine as it currently is

src/t_hash.c
```diff
@@ -1317,16 +1317,15 @@ void hrandfieldCommand(client *c) {
 
     if (c->argc >= 3) {
         if (getRangeLongFromObjectOrReply(c, c->argv[2], -LONG_MAX, LONG_MAX, &l, NULL) != C_OK) return;
+        if (l < -server.max_rand_count || l > server.max_rand_count) {
```
sarthakaggarwal97: I think this changes the current compatibility by adding an additional restriction, so we might have discuss to this before proceeding further. We will have to document this change in behaviour as well.

src/t_set.c
sarthakaggarwal97: yes it is fine.

src/config.c
```diff
@@ -3344,6 +3344,7 @@ standardConfig static_configs[] = {
     createLongLongConfig("stream-node-max-entries", NULL, MODIFIABLE_CONFIG, 0, LLONG_MAX, server.stream_node_max_entries, 100, INTEGER_CONFIG, NULL, NULL),
     createLongLongConfig("repl-backlog-size", NULL, MODIFIABLE_CONFIG, 1, LLONG_MAX, server.repl_backlog_size, 10 * 1024 * 1024, MEMORY_CONFIG, NULL, updateReplBacklogSize), /* Default: 10mb */
     createLongLongConfig("cluster-manual-failover-timeout", NULL, MODIFIABLE_CONFIG, 1, INT_MAX, server.cluster_mf_timeout, 5000, INTEGER_CONFIG, NULL, NULL),
+    createLongLongConfig("max-rand-count", NULL, MODIFIABLE_CONFIG, 0, LONG_MAX / 2, server.max_rand_count, LONG_MAX / 2, INTEGER_CONFIG, NULL, NULL),
```
sarthakaggarwal97, line 3347: why did we choose LONG_MAX / 2?

src/t_set.c
sarthakaggarwal97: yes this is fine.

src/config.c
parimoh: The default of `LONG_MAX/2` is because when using the `withscores` or `withvalues` arguments for `zrandmember` and `hrandfield`, the number of outputs will double. Meaning if a count of less than `-LONG_MAX/2` is entered, it causes the server to crash since it overflows. Either way though, from the issue it sounded like it's reasonable to expect the user to be able to call the command multiple times if they really need more values.

src/t_set.c
ranshid: why not just use the range limits of getRangeLongFromObjectOrReply?

```suggestion
    if (getRangeLongFromObjectOrReply(c, c->argv[2], -server.max_rand_count, server.max_rand_count, &l, NULL) != C_OK) return;
```

src/t_zset.c
```diff
@@ -4246,16 +4246,15 @@ void zrandmemberCommand(client *c) {
 
     if (c->argc >= 3) {
         if (getRangeLongFromObjectOrReply(c, c->argv[2], -LONG_MAX, LONG_MAX, &l, NULL) != C_OK) return;
+        if (l < -server.max_rand_count || l > server.max_rand_count) {
+            addReplyError(c, "value is out of range");
+            return;
+        }
```
ranshid:
```suggestion
        if (getRangeLongFromObjectOrReply(c, c->argv[2], -server.max_rand_count, server.max_rand_count, &l, NULL) != C_OK) return;
```

src/t_zset.c
```diff
@@ -4246,16 +4246,15 @@ void zrandmemberCommand(client *c) {
...
         if (c->argc > 4 || (c->argc == 4 && strcasecmp(c->argv[3]->ptr, "withscores"))) {
             addReplyErrorObject(c, shared.syntaxerr);
             return;
-        } else if (c->argc == 4) {
+        } else if (c->argc == 4)
             withscores = 1;
```
ranshid: suggest it is clearer with the curly brackets
```suggestion
        } else if (c->argc == 4) {
            withscores = 1;
        }
```

src/t_hash.c
```diff
@@ -1317,16 +1317,15 @@ void hrandfieldCommand(client *c) {
...
+            addReplyError(c, "value is out of range");
+            return;
+        }
```
ranshid:
```suggestion
        if (getRangeLongFromObjectOrReply(c, c->argv[2], -server.max_rand_count, server.max_rand_count, &l, NULL) != C_OK) return;
```

ranshid (commented):
Left some minor comments.

I am not sure I get the reason we need a config? It the issue only related to the growth of COB? why the same cannot happen in hgetall or any other unbounded command?

Also if this is needed, why a config is better than just ALWAYS limits to LONG_MAX/2?

parimoh:
> Left some minor comments.
> 
> I am not sure I get the reason we need a config? It the issue only related to the growth of COB? why the same cannot happen in hgetall or any other unbounded command?
> 
> Also if this is needed, why a config is better than just ALWAYS limits to LONG_MAX/2?

So I think the idea (and please correct me if I'm wrong, I'm very new to this :)) is that these commands are easy to abuse since they can cause Valkey to hang. You're right that the same can happen with a command like HGETALL, LRANGE, ZRANGE, etc. But these RAND member commands in particularly are easy to abuse because the caller only needs to create a hash with a single field to hang Valkey. For something like HGETALL it would require either creating a huge hash or trying to find an existing one. But maybe this too is considered an easy way to abuse Valkey and is something that should be implemented here as well.

We went with a config for this because the user may want to limit the impact of something like this while still setting it to a reasonable value for their own use case. With the default of LONG_MAX/2 Valkey could still hang very easily, so the user may want to lower this. But maybe if we don't foresee the abuse of this command being a big issue, we could remove the config and just keep it as a bounds check.

ranshid:
I think that users have a more general protection mechanism by limiting clients cob sizes. We usually try to avoid adding configs which users normally do not use and we have the responsibility to set a default according to the best practices. This is why I tend to not like this change ( even though there is no problem in the code)

parimoh:
Yeah that makes sense, thanks for the feedback. Do you have any suggestions for how to address the issue or if the issue should even be addressed in the first place? Because as is actually the bounds I believe are being checked properly, it's just the hanging "issue" that needs to be addressed.

ranshid:
> #844

Personally I do not think we need to address the issue. Valkey users have enough ways they can shoot themselves in the foot and this will not be a general solution for the problem. I guess the only thing we could do better is to check the client CLOSE_ASAP flag during the random generation so we can bail out early. 

Let me write something about it in the issue (maybe others will disagree)

madolson:
As we discussed, the preference is to not use a config and instead just exit out of the loop once the client exceeds the client output buffer limits for normal clients.

PR #2393, opened 2025-07-29
Add path filters to skip CI on documentation changes

### LINKED ISSUE #626: Investigate using path filters to reduce the number of tests we are running
If someone updates the README file, we run all of the tests. We should investigate using path filters, https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore, so we only have to run tests when it actually makes sense to do so. My only initial suggestion is we shouldn't run the functional tests if *.md is modified, but maybe folks have other ideas.

hanxizh9910:
Fixes #626 - Skip CI tests when only documentation files are modified


.github/workflows/spell-check.yml
```diff
@@ -7,7 +7,11 @@ name: Spellcheck
 
 on:
   push:
+    paths-ignore:
```
sarthakaggarwal97: I think let's keep the spell-check on the md files. That might be still relevant.

sarthakaggarwal97 (commented):
Overall looks fine to me except for one comment. It currently covers just .md files, do we want to include any more file patterns, if you can take a look?

.github/workflows/spell-check.yml
hanxizh9910: Thank's for the feedback! I'll remove the path filters from spell-check.yml and keep it for the other workflows.

hanxizh9910:
> Overall looks fine to me except for one comment. It currently covers just .md files, do we want to include any more file patterns, if you can take a look?

We can definitely add more documentation-related files. In addition to .md files, I suggest we include License files like COPYING and some release notes

hanxizh9910:
I tested the path filters by making minor modifications to both documentation files and source code files to verify that CI workflows are properly skipped for documentation-only changes while still running for code changes.

Screenshot 2025-07-31 at 6 20 45 PM

sungming2:
LGTM, I think you don't need to push this commit even though it doesn't change: [c867213](https://github.com/valkey-io/valkey/pull/2393/commits/c86721384266e6dede0df25d62090ce95d60702c)

sarthakaggarwal97:
@sungming2 I think @hanxizh9910 pushed a dummy commit to rerun flaky CI workflows, but I would prefer to revert it so that there is no extra diff.

@hanxizh9910 - you can either use `commit --amend`, and force push, use empty commit or maybe request one of the maintainers to rerun the workflows.

sungming2 (approved):

sarthakaggarwal97 (approved):

madolson (approved):

.github/workflows/clang-format.yml
```diff
@@ -2,9 +2,17 @@ name: Clang Format Check
...
+      - '**/*.md'
+      - '**/00-RELEASENOTES'
+      - '**/COPYING'
   pull_request:
     paths:
       - 'src/**'
+    paths-ignore:
+      - '**/*.md'
+      - '**/00-RELEASENOTES'
+      - '**/COPYING'
```
madolson, line 15: Apparently you can't do both of these.

hanxizh9910: Thanks for the feedback! I made a follow-up PR to fix this: #2455

PR #2394, opened 2025-07-30
Fix prefetchNextBucketEntries so it fetches the correct next 2 buckets.

Nicky-2000:
This PR fixes a bug in prefetchNextBucketEntries which is used in the hashtable iterator. The current version of prefetchNextBucketEntries does not correctly prefetch the next two buckets that will be iterated over. This is due next_index being incremented twice (once in prefetchNextBucketEntries and again in getNextBucket).

## Fix Explanation: 
1. Add check in prefetchNextBucketEntries to ensure we pass the correct 'next_index' to getNextBucket.
2. The extra increment was removed from getNextBucket.

## Performance
The fix was tested in a similar manner as #1501 where these changes were introduced.

All the data below was taken on a 
The data below was collected by executing the KEYS * command and Save command on a 64-logical core (32 physical cores, 2 threads per core) AMD EPYC 7B13 processor. The dataset had 50 million keys with a value size of 100 bytes each.

### KEYS * Command

| Fix Applied | Mean Time (seconds) Over 8 trials | Std Dev (s) Over 8 Trials |
|-------------|-----------------------------------|---------------------------|
| No          | 6.374759                          | 0.10                      |
| Yes         | 6.184731                          | 0.06                      |

Improvement:

- Modest 2.98% reduction in command duration, 
- 40% improvement in standard deviation. 
- This shows that the fix slightly speeds up the command and makes its performance more consistent.

#### Setup:

-  64-logical core (32 physical cores, 2 threads per core) AMD EPYC 7B13
-  50 mil keys in size of 100 bytes each.
-  Running valkey server over standard file system with 402 M/s throughput.
-  crc checksum ON and comperssion ON .

#### Results

| Fix Applied | Mean Time (s) Over 8 trials | Std Dev (s) Over 8 Trials |
|-------------|-----------------------------|---------------------------|
| No          | 49.87                       | 0.38                      |
| Yes         | 49.85                       | 0.15                      |

Improvement:

- No noticeable improvement or regression in average Save performance
- Performance is more consistent with the fix (standard deviation 62% smaller over the 8 test trials).

### DIFF
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -992,16 +992,16 @@ static void prefetchBucketEntries(bucket *b) {
     }
 }
 
-/* Returns the child bucket if chained, otherwise the next bucket in the table. returns NULL if neither exists. */
-static bucket *getNextBucket(bucket *current_bucket, size_t bucket_index, hashtable *ht, int table_index) {
+/* Returns the child bucket if 'current_bucket' is chained. Otherwise, returns the bucket
+ * at 'next_top_level_index' in the table. Returns NULL if neither exists. */
+static bucket *getNextBucket(bucket *current_bucket, size_t next_top_level_index, hashtable *ht, int table_index) {
     bucket *next_bucket = NULL;
     if (current_bucket->chained) {
         next_bucket = getChildBucket(current_bucket);
     } else {
         size_t table_size = numBuckets(ht->bucket_exp[table_index]);
-        size_t next_index = bucket_index + 1;
-        if (next_index < table_size) {
-            next_bucket = &ht->tables[table_index][next_index];
+        if (next_top_level_index < table_size) {
+            next_bucket = &ht->tables[table_index][next_top_level_index];
         }
     }
     return next_bucket;
@@ -1012,7 +1012,7 @@ static bucket *getNextBucket(bucket *current_bucket, size_t bucket_index, hashta
  * - The next of the next bucket
  * It attempts to bring this data closer to the L1 cache to reduce future memory access latency.
  *
- * Cache state before this function is called(due to last call for this function):
+ * Cache state before this function is called (due to last call for this function):
  * 1. The current bucket and its entries are likely already in cache.
  * 2. The next bucket is in cache.
  */
@@ -1021,7 +1021,9 @@ static void prefetchNextBucketEntries(iter *iter, bucket *current_bucket) {
     bucket *next_bucket = getNextBucket(current_bucket, next_index, iter->hashtable, iter->table);
     if (next_bucket) {
         prefetchBucketEntries(next_bucket);
-        bucket *next_next_bucket = getNextBucket(next_bucket, next_index + 1, iter->hashtable, iter->table);
+        /* Calculate the target top-level index for the next-next bucket. */
+        if (!current_bucket->chained) next_index++;
+        bucket *next_next_bucket = getNextBucket(next_bucket, next_index, iter->hashtable, iter->table);
         if (next_next_bucket) {
             valkey_prefetch(next_next_bucket);
         }


zuiderkwast:
@NadavGigi PTAL

src/hashtable.c
```diff
@@ -1021,7 +1021,9 @@ static void prefetchNextBucketEntries(iter *iter, bucket *current_bucket) {
     bucket *next_bucket = getNextBucket(current_bucket, next_index, iter->hashtable, iter->table);
     if (next_bucket) {
         prefetchBucketEntries(next_bucket);
-        bucket *next_next_bucket = getNextBucket(next_bucket, next_index + 1, iter->hashtable, iter->table);
+        /* Calculate the target top-level index for the next-next bucket. */
+        if (!current_bucket->chained) next_index++;
+        bucket *next_next_bucket = getNextBucket(next_bucket, next_index, iter->hashtable, iter->table);
```
zuiderkwast, line 1026: This PR looks correct, but I think the code might be slightly cleaner if we increment the index in `getNextBucket` only, instead of here.

I mean we can keep `getNextBucket` unchanged and instead skip increment it here. Passing current bucket and current index appears to be a cleaner API of the getNextBucket function, to me.

How about this?

```c
    size_t current_index = iter->index;
    bucket *next_bucket = getNextBucket(current_bucket, current_index, iter->hashtable, iter->table);
    if (next_bucket) {
        prefetchBucketEntries(next_bucket);
        /* Calculate the target top-level index for the next bucket. */
        size_t next_index = current_bucket->chained ? current_index : current_index + 1;
        bucket *next_next_bucket = getNextBucket(next_bucket, next_index, iter->hashtable, iter->table);
```

Nicky-2000: Hi!! Thanks for the review. 

The reason I have it this way is so that I can add "striding" to this function which is useful in my work for multi-thread the RDB Save operations (Save, BGSave, and Full Sync). (WIP PR is located [here](https://github.com/Nicky-2000/valkey/pull/2).

The change you suggested would require passing the "stride" argument into the getNextBucket. Which is totally fine and will work. Just wanted to make you aware of why I decided to do it this way.

Here is the stride implementation for these functions that I am going to propose in a later PR. 
```C
/* Returns the child bucket if 'current_bucket' is chained. Otherwise, returns the bucket
 * at 'next_top_level_index' in the table. Returns NULL if neither exists. */
static bucket *getNextBucket(bucket *current_bucket, size_t next_top_level_index, hashtable *ht, int table_index) {
    bucket *next_bucket = NULL;
    if (current_bucket->chained) {
        next_bucket = getChildBucket(current_bucket);
    } else {
        size_t table_size = numBuckets(ht->bucket_exp[table_index]);
        if (next_top_level_index < table_size) {
            next_bucket = &ht->tables[table_index][next_top_level_index];
        }
    }
    return next_bucket;
}

/* This function prefetches data that will be needed in subsequent iterations:
 * - The entries of the next bucket
 * - The next of the next bucket
 * It attempts to bring this data closer to the L1 cache to reduce future memory access latency.
 *
 * Cache state before this function is called (due to last call for this function):
 * 1. The current bucket and its entries are likely already in cache.
 * 2. The next bucket is in cache.
 * 
 * The stride variable determines how many top level buckets to jump forward. If we complete
 * a bucket chain, we "stride forward" to the next top-level bucket index.
 */
static void prefetchNextBucketEntries(iter *iter, bucket *current_bucket, size_t stride) {
    size_t next_index = iter->index + stride;
    bucket *next_bucket = getNextBucket(current_bucket, next_index, iter->hashtable, iter->table);
    if (next_bucket) {
        prefetchBucketEntries(next_bucket);
        /* Calculate the target top-level index for the next-next bucket. */
        if (!current_bucket->chained) next_index += stride;
        bucket *next_next_bucket = getNextBucket(next_bucket, next_index, iter->hashtable, iter->table);
        if (next_next_bucket) {
            valkey_prefetch(next_next_bucket);
        }
    }
}
```

If you think it is easier to understand by passing through the "stride" then I will go ahead and use the version you suggested!

The stride prefetch implementation will be used [here](https://github.com/Nicky-2000/valkey/blob/39ea6d52d4b0428a6fdc0239eaa0a12c48b7a129/src/hashtable.c#L2231) in the hashtableStrideNext Iterator.

NadavGigi: Great work with finding this bug!
I support @zuiderkwast approach with keeping `getNextBucket` unchanged.

zuiderkwast: Thanks for the context! getNextBucket is only used in prefetchNextBucketEntries anyway so we might as well just pass forward the stride to getNextBucket. However, I see that we still need to increment the index in prefetchNextBucketEntries for the next-next bucket, so we can't fully delegate the incrementing to getNextBucket anyway...

I'm fine with your approach of doing all index increments in prefetchNextBucketEntries.

zuiderkwast: @NadavGigi does this seem OK to you?

zuiderkwast (approved):

src/hashtable.c
NadavGigi: LGTM

PR #2395, opened 2025-07-30
Cleanup and rename isExpiryTableValidForSamplingCb to expiryTableShouldSkipForSamplingCb

enjoy-binbin:
We used to return C_ERR (-1) or C_OK (0), it was not that bool.
usage: int skip = !ht || (skip_cb && skip_cb(ht));

Also isExpiryTableValidForSamplingCb is actually a skip cp, rename
it to expireShouldSkipTableForSamplingCb for a better reading.

### DIFF
--- a/src/expire.c
+++ b/src/expire.c
@@ -156,9 +156,9 @@ static inline int isExpiryTableValidForSamplingCb(hashtable *ht) {
      * space is expensive, so stop here waiting for better times...
      * The dictionary will be resized asap. */
     if (buckets > 0 && (numkeys * 100 / buckets < 1)) {
-        return C_ERR;
+        return 1;
     }
-    return C_OK;
+    return 0;
 }
 
 void activeExpireCycle(int type) {
--- a/src/expire.c
+++ b/src/expire.c
@@ -149,7 +149,7 @@ void expireScanCallback(void *privdata, void *entry) {
     data->sampled++;
 }
 
-static inline int isExpiryTableValidForSamplingCb(hashtable *ht) {
+static inline int expiryTableShouldSkipForSamplingCb(hashtable *ht) {
     long long numkeys = hashtableSize(ht);
     unsigned long buckets = hashtableBuckets(ht);
     /* When there are less than 1% filled buckets, sampling the key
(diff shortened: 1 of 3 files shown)


src/expire.c
```diff
@@ -156,9 +156,9 @@ static inline int isExpiryTableValidForSamplingCb(hashtable *ht) {
      * space is expensive, so stop here waiting for better times...
      * The dictionary will be resized asap. */
     if (buckets > 0 && (numkeys * 100 / buckets < 1)) {
-        return C_ERR;
+        return 1;
     }
-    return C_OK;
+    return 0;
```
zuiderkwast, line 161: The "is valid" function returns 0 if valid? Seems like inverted boolean, quite confusing. 😵‍💫

Can we change to `bool isValidXxxxx` instead and true means "yes, it's valid"?

src/kvstore.h
```diff
@@ -8,7 +8,9 @@ typedef struct _kvstore kvstore;
 typedef struct _kvstoreIterator kvstoreIterator;
 typedef struct _kvstoreHashtableIterator kvstoreHashtableIterator;
 
+/* Return 1 if we should skip the hashtable in scan. */
 typedef int(kvstoreScanShouldSkipHashtable)(hashtable *d);
+/* Return 1 if we should skip the hashtable in expand. */
 typedef int(kvstoreExpandShouldSkipHashtableIndex)(int didx);
```
zuiderkwast, line 14: So the isValid function is actually a shouldSkip callback? We should rename all to isValid IMO. (Or rename all to shouldSkip.)

Something like this?

```suggestion
typedef bool(kvstoreExpandHashtableIndexIsValid)(int didx);
```

zuiderkwast (commented):
Good catch. To clean it up better, we can fix the names and make it bool.

src/kvstore.h
enjoy-binbin: yes, it is actually a skip_cp, i updated it to use skip (seem better and less diff)

src/expire.c
```diff
@@ -149,16 +149,16 @@ void expireScanCallback(void *privdata, void *entry) {
     data->sampled++;
 }
 
-static inline int isExpiryTableValidForSamplingCb(hashtable *ht) {
+static inline int expiryTableShouldSkipForSamplingCb(hashtable *ht) {
```
zuiderkwast: Good, but maybe reorder the words for better grammar? "Should skip table for sampling" is better than "Table should skip for sampling" I think.

```suggestion
static inline int expiryShouldSkipTableForSamplingCb(hashtable *ht) {
```

Oh, is "expiry" the prefix for this file, or is "expiryTable" one concept? Maybe it should be one of these?

1. expireShouldSkipTableForSamplingCb (prefix "expire" like the filename)
2. shouldSkipExpiryTableForSamplingCb (if expiryTable is one concept)

enjoy-binbin: expireShouldSkipTableForSamplingCb seems better to me.
I don't see we have or use expiryTable this concept a lot

zuiderkwast (approved):

PR #2402, opened 2025-07-31
Update check times for AOF loading in memefficiency.tcl

asagege:
Update check times from 100k to 80k for AOF loading. 

Reason: During AOF loading, we only hit 95k items rather than 100k in one defrag test

Defrag test: https://github.com/valkey-io/valkey/actions/runs/15452801881/job/43498734253?pr=2078#step:6:7047 (line 6157)

### DIFF
--- a/tests/unit/memefficiency.tcl
+++ b/tests/unit/memefficiency.tcl
@@ -308,7 +308,7 @@ run_solo {defrag} {
                 # Make sure we had defrag hits during AOF loading.  Note that we don't worry about
                 # the actual fragmentation ratio here.  It will vary based on when defrag stopped
                 # mid-cycle.  Just check that we are defragging by the number of hits.
-                assert {[s active_defrag_hits] > 100000}
+                assert {[s active_defrag_hits] > 80000}
             }
             } ;# Active defrag - AOF loading
         }


sarthakaggarwal97:
@asagege why did we choose 80K?

JimB123:
> @asagege why did we choose 80K?

100k is just a historical value.  The actual value will fluctuate based on architecture, OS, and random factors.  The key point of the test is that defragging is occurring during the AOF loading.  When I re-worked these tests in the fall, 100k seemed ok.  But observations at 95k show that this isn't a reliable value.

IMO 80k is a reasonable value to ensure test reliability, while maintaining the integrity of the test.

JimB123 (approved):
This is a reasonable change to improve test robustness while maintaining test integrity.

tests/unit/memefficiency.tcl
```diff
@@ -308,7 +308,7 @@ run_solo {defrag} {
                 # Make sure we had defrag hits during AOF loading.  Note that we don't worry about
                 # the actual fragmentation ratio here.  It will vary based on when defrag stopped
                 # mid-cycle.  Just check that we are defragging by the number of hits.
-                assert {[s active_defrag_hits] > 100000}
+                assert {[s active_defrag_hits] > 80000}
```
hpatro, line 311: If the test doesn't get stabilized with this change, we would be again left in the dark about what was the defrag hit value. On failure if we print out some more information, it will help us debug better.

Could you please update the PR with a conditional statement to print out the result (maybe more defrag information) and then fail?

JimB123: There's already information that's printed.  Example: https://github.com/valkey-io/valkey/actions/runs/15452801881/job/43498734253?pr=2078#step:6:6154-6167   (line 6157)

hpatro: Nice.

```
Frag info: after AOF loading
    used:  30.26 MB
    frag:   8.68 MB
    frag ratio: 1.29
    hits: 95442
    misses: 998152
```

hpatro (approved):
@asagege Thanks for the PR.

sarthakaggarwal97 (approved):

Issue #1929, opened 2025-04-08
[NEW] Measure the amount of network in/out over cluster bus


hpatro:
**The problem/use-case that the feature addresses**

Measure the amount of network in/out over cluster bus.

**Description of the feature**

Get network in/out metrics for control and data (pubsub) traffic passed via cluster bus. `CLUSTER INFO` currently provide messages_sent and messages_received however it doesn't provide the network in/out metrics.

**Alternatives you've considered**

Run single Valkey node on a server and capture network stats of the host or use CloudWatch metrics while no other command is running

sarthakaggarwal97:
@hwware are you working on this? If not, I can take it up. @valkey-io/core-team it would be good to know if there are any concerns over implementing this feature.

zuiderkwast:
I have no concerns. The feature sounds good to me.

enjoy-binbin:
This looks like a good feature, I previously estimated it using cluster_stats_messages_sent / cluster_stats_messages_received.

hwware:
> [@hwware](https://github.com/hwware) are you working on this? If not, I can take it up. [@valkey-io/core-team](https://github.com/orgs/valkey-io/teams/core-team) it would be good to know if there are any concerns over implementing this feature.

Please take it freely, recently I am working on other features.  You could change the assignee to you.  
Thanks

hpatro:
> This looks like a good feature, I previously estimated it using cluster_stats_messages_sent / cluster_stats_messages_received.

Yeah, that is a good approximation. With user workload, it can get tricky with pubsub traffic as the message size is variable.

hpatro:
It would be also good to hear perspective of people who run it on-prem or self deploy in different environments.

sarthakaggarwal97:
In order to have better visibility into how much traffic is caused by internal cluster operations (like PING, FAIL, and CONFIG messages) versus application-level pub/sub traffic (PUBLISH), I am proposing separate counters for admin and pub/sub traffic, we can provide more meaningful insight with very little overhead.

### Cluster Admin Traffic
- `cluster_bus_admin_bytes_sent`
- `cluster_bus_admin_bytes_received`

### Pub/Sub Traffic
- `cluster_bus_pubsub_bytes_sent`
- `cluster_bus_pubsub_bytes_received`

We can probably update these counters in `clusterWriteHandler()` and `clusterReadHandler()`. Please let me know if this approach looks good.

hpatro:
[...]

PR #2404, opened 2025-07-31
track bytes of traffic sent and received

parimoh:
This PR is meant to address issue #1929 by tracking cluster bus traffic separately for different sources of traffic. I don't believe this has received TSC approval yet, so I just left this as a draft PR.

**Features Added:**

- Track administrative traffic (PING, PONG, MEET, FAIL, etc.)
- Track pub/sub traffic separately from other cluster communications
- Add module-specific traffic tracking with per-module statistics
- Implement new counters in the server structure for bytes sent and received
- Add new fields to cluster info command output

**Implementation Details:**

- Added new counters:
- cluster_bus_admin_bytes_sent/received
- cluster_bus_pubsub_bytes_sent/received
- Per-module traffic tracking using hashtables

**Testing:**

- Added comprehensive test suite (bus-traffic-stats.tcl) that verifies:
  - Initial counter states
  - Administrative traffic tracking
  - Pub/sub traffic tracking
  - Module-specific message tracking

This enhancement will help operators better understand and monitor cluster communication patterns and identify potential bottlenecks in their Valkey cluster deployments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       - name: make
         # Fail build if there are warnings
         # build with TLS just for compilation coverage
-        run: make -j4 all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes USE_FAST_FLOAT=yes
+        run: make -j4 all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes USE_FAST_FLOAT=yes BUILD_EXAMPLE_MODULES=yes
       - name: install old server for compatibility testing
         run: |
           cd tests/tmp
@@ -48,7 +48,7 @@ jobs:
           sudo apt-get install -y cmake libssl-dev
           mkdir -p build-release
           cd build-release
-          cmake -DCMAKE_BUILD_TYPE=Release .. -DBUILD_TLS=yes -DBUILD_UNIT_TESTS=yes
+          cmake -DCMAKE_BUILD_TYPE=Release .. -DBUILD_TLS=yes -DBUILD_UNIT_TESTS=yes -DBUILD_EXAMPLE_MODULES=yes
           make -j$(nproc)
       - name: test
         run: |
@@ -70,7 +70,7 @@ jobs:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - name: make
         # build with TLS module just for compilation coverage
-        run: make -j4 all-with-unit-tests SANITIZER=address SERVER_CFLAGS='-Werror' BUILD_TLS=module
+        run: make -j4 all-with-unit-tests SANITIZER=address SERVER_CFLAGS='-Werror' BUILD_TLS=module BUILD_EXAMPLE_MODULES=yes
       - name: testprep
         run: sudo apt-get install tcl8.6 tclx -y
       - name: test
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       - name: make
         # Fail build if there are warnings
         # build with TLS just for compilation coverage
-        run: make -j4 all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes USE_FAST_FLOAT=yes BUILD_EXAMPLE_MODULES=yes
+        run: make -j4 all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes USE_FAST_FLOAT=yes
       - name: install old server for compatibility testing
         run: |
           cd tests/tmp
@@ -48,7 +48,7 @@ jobs:
           sudo apt-get install -y cmake libssl-dev
           mkdir -p build-release
           cd build-release
-          cmake -DCMAKE_BUILD_TYPE=Release .. -DBUILD_TLS=yes -DBUILD_UNIT_TESTS=yes -DBUILD_EXAMPLE_MODULES=yes
+          cmake -DCMAKE_BUILD_TYPE=Release .. -DBUILD_TLS=yes -DBUILD_UNIT_TESTS=yes
           make -j$(nproc)
       - name: test
         run: |
@@ -70,7 +70,7 @@ jobs:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - name: make
         # build with TLS module just for compilation coverage
-        run: make -j4 all-with-unit-tests SANITIZER=address SERVER_CFLAGS='-Werror' BUILD_TLS=module BUILD_EXAMPLE_MODULES=yes
+        run: make -j4 all-with-unit-tests SANITIZER=address SERVER_CFLAGS='-Werror' BUILD_TLS=module
       - name: testprep
         run: sudo apt-get install tcl8.6 tclx -y
       - name: test
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -3990,7 +3990,7 @@ static void updateModuleTraffic(hashtable *ht, uint64_t module_id, long long byt
         entry = zmalloc(sizeof(moduleTrafficEntry));
         entry->module_id = module_id;
         entry->sent_bytes = isSendTraffic ? bytes : 0;
-        entry->received_bytes= !isSendTraffic ? bytes : 0;
+        entry->received_bytes = !isSendTraffic ? bytes : 0;
         hashtableAdd(ht, entry);
     }
 }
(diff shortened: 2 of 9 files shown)


src/cluster_legacy.c
```diff
@@ -6724,11 +6780,17 @@ sds genClusterInfoString(void) {
                      "cluster_known_nodes:%U\r\n"
                      "cluster_size:%i\r\n"
                      "cluster_current_epoch:%U\r\n"
-                     "cluster_my_epoch:%U\r\n",
+                     "cluster_my_epoch:%U\r\n"
+                     "cluster_bus_admin_bytes_sent:%U\r\n"
+                     "cluster_bus_admin_bytes_received:%U\r\n"
+                     "cluster_bus_pubsub_bytes_sent:%U\r\n"
+                     "cluster_bus_pubsub_bytes_received:%U\r\n",
```
hpatro: I think we can append it to the end.

src/cluster_legacy.c
```diff
@@ -6675,6 +6715,22 @@ void clusterCommandShards(client *c) {
...
+    if (!ht) return info;
+
+    hashtableIterator iter;
+    hashtableInitIterator(&iter, ht, 0);
+    moduleTrafficEntry *entry;
+
+    while (hashtableNext(&iter, (void **)&entry)) {
+        info = sdscatfmt(info, "cluster_bus_module_%s_bytes_%U:%I\r\n", 
+                        direction, entry->module_id, entry->traffic_bytes);
+    }
```
hpatro: unsure about using module id to print out as part of cluster info, not readable.

Might as well fetch the modulename and form the key.

src/cluster_legacy.c
```diff
@@ -3977,6 +3977,19 @@ void handleLinkIOError(clusterLink *link) {
...
+/* Helper function to update module traffic in hashtable */
+static void updateModuleTraffic(hashtable *ht, uint64_t module_id, long long bytes) {
+    moduleTrafficEntry *entry;
+    if (hashtableFind(ht, &module_id, (void **)&entry)) {
+        entry->traffic_bytes += bytes;
+    } else {
+        entry = zmalloc(sizeof(moduleTrafficEntry));
+        entry->module_id = module_id;
+        entry->traffic_bytes = bytes;
+        hashtableAdd(ht, entry);
```
hpatro, line 3994: On module unload, maybe we should cleanup and stop maintaining the entry.

src/cluster_legacy.c
hpatro: We should maintain both send/receive traffic at module level as well in one entry. Will reduce the number of pointer access ideally by colocating.

src/server.c
```diff
@@ -2256,6 +2283,12 @@ void initServerConfig(void) {
     server.shutdown_flags = 0;
     server.shutdown_mstime = 0;
     server.cluster_module_flags = CLUSTER_MODULE_FLAG_NONE;
+    server.cluster_bus_admin_bytes_received = 0;
+    server.cluster_bus_admin_bytes_sent = 0;
+    server.cluster_bus_pubsub_bytes_received = 0;
+    server.cluster_bus_pubsub_bytes_sent = 0;
+    server.cluster_bus_module_id_sent = hashtableCreate(&moduleTrafficHashtableType);
+    server.cluster_bus_module_id_received = hashtableCreate(&moduleTrafficHashtableType);
```
hpatro: Shall we just have one hashtable with sent/received as part of the entry?

src/server.c
```diff
@@ -685,6 +685,33 @@ hashtableType sdsReplyHashtableType = {
...
+    const moduleTrafficEntry *traffic_entry = entry;
+    return &traffic_entry->module_id;
+}
+
+void moduleTrafficDestructor(void *entry) {
+    zfree(entry);
+}
+
+uint64_t moduleTrafficHash(const void *key) {
+    return hashtableGenHashFunction((const char *)key, sizeof(uint64_t));
```
hpatro: Might as well send the module_id directly as the hash and not compute the hash. :D

hpatro (commented):
Thanks for the PR. Have left some comments. 

I don't think it's critical so we don't need to push it in 9.0.RC1. Will bring it up with the core team after that. I think we can still get it in 9.0 if everyone agrees.

parimoh:
Thanks for the feedback! I think I addressed all of it, I'm currently looking into the CI testing failures. Seems like it's something to do with the module load. Also I have a couple comments/questions about some parts of the code below, thanks

src/cluster_legacy.c
```diff
@@ -6675,6 +6729,37 @@ void clusterCommandShards(client *c) {
...
+        moduleType *mt = moduleTypeLookupModuleByID(entry->module_id);
+        if (mt) {
+            module_name = moduleTypeModuleName(mt);
+        }
+
+        /* Fallback to type name if module name not available */
+        if (!module_name) {
+            moduleTypeNameByID(type_name, entry->module_id);
+            module_name = type_name;
+        }
```
parimoh, line 6788: Does this logic for retrieving the module name seem okay? I was kinda just basing this off of the function names/comments but not totally sure if it's correct, gonna look into this more as well

src/server.h
```diff
@@ -2137,6 +2137,11 @@ struct valkeyServer {
                                                             * the cluster after it is forgotten with CLUSTER FORGET. */
     int cluster_slot_stats_enabled;                        /* Cluster slot usage statistics tracking enabled. */
     mstime_t cluster_mf_timeout;                           /* Milliseconds to do a manual failover. */
+    unsigned long long cluster_bus_admin_bytes_sent;       /* Cluster bus admin traffic sent. */
+    unsigned long long cluster_bus_admin_bytes_received;   /* Cluster bus admin traffic received. */
+    unsigned long long cluster_bus_pubsub_bytes_sent;      /* Cluster bus pubsub traffic sent. */
+    unsigned long long cluster_bus_pubsub_bytes_received;  /* Cluster bus pubsub traffic received. */
```
parimoh, line 2143: Is it worth making a `trafficData` type that stores `sent_bytes` and `received_bytes` so that the `cluster_bus_admin` and `cluster_bus_pubsub` traffic data can each just be one variable?

.github/workflows/ci.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: make
         # Fail build if there are warnings
         # build with TLS just for compilation coverage
-        run: make -j4 all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes USE_FAST_FLOAT=yes
+        run: make -j4 all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes USE_FAST_FLOAT=yes BUILD_EXAMPLE_MODULES=yes
```
sarthakaggarwal97: why did we make these changes?

src/cluster_legacy.c
```diff
@@ -3977,6 +3977,33 @@ void handleLinkIOError(clusterLink *link) {
     freeClusterLink(link);
 }
 
+/* Helper function to update module traffic in hashtable */
+static void updateModuleTraffic(hashtable *ht, uint64_t module_id, long long bytes, bool isSendTraffic) {
```
sarthakaggarwal97: we should not use camel case `isSendTraffic`

src/cluster_legacy.c
sarthakaggarwal97: +1

sarthakaggarwal97 (commented):
Took an initial look. I have added the issue to the weekly agenda for core team, let's see if we get time to discuss this.

.github/workflows/ci.yml
parimoh: Sorry yeah just reverted it, thought it might fix the CI issues that were happening but it didn't

tests/unit/cluster/bus-traffic-stats.tcl
```diff
@@ -0,0 +1,94 @@
...
+            assert {$bytes >= 0}
+        }
+    }
+
+    # Verify that both sent and received module traffic were tracked
+    assert {$sent_found == 1}
+    assert {$recv_found == 1}
+}
+
+} ;# start_cluster
```
sarthakaggarwal97:
```suggestion
} 
```

src/cluster_legacy.h
```diff
@@ -430,4 +430,10 @@ struct clusterState {
     slotStat slot_stats[CLUSTER_SLOTS];
 };
 
+/* Struct used for storing module traffic statistics. */
+typedef struct {
+    uint64_t module_id;
+    long long traffic_bytes;
+} moduleTrafficEntry;
```
sarthakaggarwal97: I would probably name it as `moduleClusterTrafficEntry`, unless we see more usecase of this structure? Then in that case it should not be in this class

src/cluster_legacy.c
```diff
@@ -3997,6 +4010,19 @@ void clusterWriteHandler(connection *conn) {
             handleLinkIOError(link);
             return;
         }
+        // Get message type to differentiate traffic
```
sarthakaggarwal97: can we keep the same comment style as surrounding comments using `/* */`

parimoh:
Thanks for the feedback, just updated the code. Gonna take a look at also adding testing for the `moduleUnload` functionality I added.

parimoh:
I think I finally fixed the testing issues, there were 4 main changes I had to make:

1. I was double freeing the hashtable entry upon delete
2. I wasn't considering light cluster messages in my code which was causing a read to an invalid address when trying to get the module ID
3. I had to add retries for certain commands in the testing for the `sanitizer` tests since they run a lot slower
4. I had to build the module if it wasn't built already in the tests

sarthakaggarwal97:
Closing this as #1929 is done now.

PR #2409, opened 2025-08-01
Dictionary Implementation for Migrating and Importing slots

### LINKED ISSUE #2184: Consider using dictionary for `migrating_slots_to`/`importing_slots_from`
Both `migrating_slots_to` and `importing_slots_from` are typically very sparse and should be empty in steady state. These take 128 KiB each and the cost to figure out if any slot is being migrated is not cache efficient (because we need to scan 128 KiB each).

It seems like a dictionary would probably be a better data structure, and we could quickly use the size to know if any slot is importing or migrating.

See https://github.com/valkey-io/valkey/pull/1949#discussion_r2124848616 for initial discussion

sarthakaggarwal97:
Resolves #2184


src/cluster_legacy.c
```diff
@@ -262,6 +281,47 @@ static void clusterNodeIterReset(ClusterNodeIterator *iter) {
...
+static void setMigratingSlot(int slot, clusterNode *node) {
+    dictEntry *de = dictFind(server.cluster->migrating_slots_to, &slot);
+    if (node == NULL) {
+        if (de) dictDelete(server.cluster->migrating_slots_to, &slot);
+        return;
+    }
+    if (de) {
+        dictSetVal(server.cluster->migrating_slots_to, de, node);
+    } else {
+        int *k = zmalloc(sizeof(int));
```
murphyjacob4: Instead of storing int* as the key and dynamically allocating it- could we just cast `int` to `void *` and store it directly?

I had a need for something similar here: https://github.com/valkey-io/valkey/pull/1949/files#diff-56dde90ed71d382a4bb92d3d07e2be5df85a40e91f2b4ad56e0b75bc87abc95eR104-R117

src/cluster_legacy.c
```diff
@@ -2764,36 +2826,38 @@ void clusterUpdateSlotsConfigWith(clusterNode *sender, uint64_t senderConfigEpoc
...
-                if (server.cluster->migrating_slots_to[j] != NULL) {
+                clusterNode *mn = getMigratingSlot(j);
+                if (mn != NULL) {
                     if (!are_in_same_shard) {
                         serverLog(LL_NOTICE, "Slot %d is no longer being migrated to node %.40s (%s) in shard %.40s.",
-                                  j, server.cluster->migrating_slots_to[j]->name,
-                                  server.cluster->migrating_slots_to[j]->human_nodename,
-                                  server.cluster->migrating_slots_to[j]->shard_id);
-                        server.cluster->migrating_slots_to[j] = NULL;
+                                  j, mn->name,
```
murphyjacob4: nit: the formatting is odd here and throughout the file after renaming to `mn`

src/cluster_legacy.c
```diff
@@ -7670,11 +7738,11 @@ clusterNode *clusterNodeGetReplica(clusterNode *node, int replica_idx) {
...
 
 clusterNode *getMigratingSlotDest(int slot) {
-    return server.cluster->migrating_slots_to[slot];
+    return getMigratingSlot(slot);
 }
 
 clusterNode *getImportingSlotSource(int slot) {
-    return server.cluster->importing_slots_from[slot];
+    return getImportingSlot(slot);
 }
```
murphyjacob4: Are these functions still needed?

src/cluster_legacy.c
```diff
@@ -6327,13 +6393,15 @@ sds clusterGenNodeDescription(client *c, clusterNode *node, int tls_primary) {
...
+                ci = sdscatlen(ci, mn->name, CLUSTER_NAMELEN);
                 ci = sdscat(ci, "]");
-            } else if (server.cluster->importing_slots_from[j]) {
+            } else if (in) {
                 ci = sdscatfmt(ci, " [%i-<-", j);
-                ci = sdscatlen(ci, server.cluster->importing_slots_from[j]->name, CLUSTER_NAMELEN);
+                ci = sdscatlen(ci, in->name, CLUSTER_NAMELEN);
                 ci = sdscat(ci, "]");
             }
         }
```
murphyjacob4: We can probably use dictIter to generate this now, right? Or is the concern around them not being sorted?

murphyjacob4 (commented):
Thanks! Overall change LGTM, just a few comments

src/cluster_legacy.c
sarthakaggarwal97: It looks pretty much similar as before. clang-format also passes. Did you notice any particular concern? (I might have missed it!)

src/cluster_legacy.c
sarthakaggarwal97: good call out, removing it.

src/cluster_legacy.c
sarthakaggarwal97: yeah, I was considering not playing with the order. If that's not a major concern, we can switch to dictiter. Let me know what you think.

src/cluster.c
```diff
@@ -1260,7 +1260,7 @@ clusterNode *getNodeByQuery(client *c, int *error_code) {
             return NULL;
         } else {
             if (error_code) *error_code = CLUSTER_REDIR_ASK;
-            return getMigratingSlotDest(c->slot);
+            return getMigratingSlot(c->slot);
```
zuiderkwast: The name `getMigratingSlot` looks like it returns a slot (a number 0-16383). The old names `getMigratingSlotDest` and `getImportingSlotSource` more clearly hint that they return a node. Can we keep the old names?

Keeping the old names (but the new implementation) would make the diff of this PR smaller.

For the setter functions, we could use names to match the getters, i.e.

    setMigratingSlotDest
    setImportingSlotSource

zuiderkwast (commented):
Good work! I just looked quickly and put a small comment. Jacob: It's great that you're reviewing this.

src/cluster_legacy.c
zuiderkwast: It seems safer to keep them ordered, if this code is not a bottleneck. Some control planes, scripts, etc. might rely on them being ordered.

src/cluster_legacy.c
zuiderkwast: The code becomes sorter so I guess the idea is that we could fit more arguments on the same line. Personally I don't care too much about it though.

src/cluster.c
sarthakaggarwal97: thanks! The old names makes more sense, I agree.

src/cluster_legacy.c
zuiderkwast: So Jacob is already adding this in the Atomic Slot Migration PR.

If we just compare the pointers as is, then we don't need to care that it's an integer. We could call the functions dictPtrCompare, etc.

```c
static uint64_t dictPtrHash(const void *key) {
    /* We hash the pointer value itself. */
    return dictGenHashFunction(&key, sizeof(key));
}

static int dictPtrCompare(const void *key1, const void *key2) {
    return key1 == key2;
}
```

sarthakaggarwal97: Using these as hash and compare functions (something like [this](https://github.com/valkey-io/valkey/commit/aeabe3f2543c25b906f6a3058f5c19b1e8b12b0e)) are causing some test failures. I think I am maybe missing something.

zuiderkwast: You're referring to these? https://github.com/sarthakaggarwal97/valkey/actions/runs/16729917885/job/47355244190#step:5:48755

@murphyjacob4 Do you have a clue?

sarthakaggarwal97: yes these are failing.

murphyjacob4: I think you are missing this update [https://github.com/valkey-io/valkey/compare/unstable...sarthakaggarwal97:valkey:dict-mig-imp-slots-1#diff-2515500619600c5a1[…]ee4d0e054691f0a8eR7800](https://github.com/valkey-io/valkey/compare/unstable...sarthakaggarwal97:valkey:dict-mig-imp-slots-1#diff-2515500619600c5a1e7a8d9aaa8761a6071314cd226c3b7ee4d0e054691f0a8eR7800)

sarthakaggarwal97: ah I think you are right. good callout, let me try! thank you.

sarthakaggarwal97: The tests are passing now. Thanks @murphyjacob4 @zuiderkwast! Please take a look when you guys have time, I think I have addressed the comments.

src/cluster_legacy.c
```diff
@@ -282,43 +283,39 @@ static void clusterNodeIterReset(ClusterNodeIterator *iter) {
 }
 
 /* Helpers to access the migrating/importing slot dictionaries. */
-clusterNode *getMigratingSlot(int slot) {
-    dictEntry *de = dictFind(server.cluster->migrating_slots_to, &slot);
+clusterNode *getMigratingSlotDest(int slot) {
+    dictEntry *de = dictFind(server.cluster->migrating_slots_to, (void *)(long)slot);
```
zuiderkwast: Nit: For maximum portability, i guess we should ideally use `intptr_t` instead of `long` in these casts. There is no guarantee in the C standard that long is pointer-sized but `intptr_t` always is.

zuiderkwast (approved):
Lgtm but not a full review. I'll merge when @murphyjacob4 has approved.

src/cluster_legacy.c
sarthakaggarwal97: TIL! I have made the change. thank you

sarthakaggarwal97: Just fyi, I referenced it from [here](https://github.com/valkey-io/valkey/pull/1949/files#diff-56dde90ed71d382a4bb92d3d07e2be5df85a40e91f2b4ad56e0b75bc87abc95eR961), I am not sure if this nit would fit there too!

src/cluster_legacy.c
murphyjacob4: Right, on my screen it looks like we are line wrapping more than needed (thus more lines of code than needed). Our clang format settings are very loose so it isn't always the best measure of what is most readable. But I come from Google where we are very strict about formatting :)

src/cluster_legacy.c
murphyjacob4: If we want sorted, we could always go with an intset or something. But I agree it probably won't be a performance bottleneck (unless someone is calling CLUSTER NODES a lot)

murphyjacob4 (approved):
Looks good!

src/cluster_legacy.c
murphyjacob4: I am going to update that before someone can leave a comment :)

zuiderkwast (approved):

PR #2414, opened 2025-08-03
Use unsigned long for `maxiterations`, fixing undefined behavior in scan with extremely large count

Fusl:
This one was found by [afl++](https://github.com/AFLplusplus/AFLplusplus). Executing `scan 0 count n` with a count that is within 10% of `LONG_MAX`, `count * 10` would cause `maxiterations` to overflow. This is technically not a real problem since the way `maxiterations` is used would eventually cause it to underflow back to `LONG_MAX` again and continue counting down from there but I figured we may want to fix this regardless for expected behavior correctness?


enjoy-binbin (approved):

ranshid (approved):
LGTM

PR #2419, opened 2025-08-04
Add STALE command flag to SCRIPT-EXISTS, SCRIPT-SHOW and SCRIPT-FLUSH

enjoy-binbin:
We marked SCRIPT-LOAD/EVAL* with STALE in 7eadc5ee7062c6de00323c1b8a9598d1b5684217,
it is odd that we can load but won't be able to exists or show it.
Also it is technically ok since these commands doesn't relate directly
to the server's dataset.

Also since now we don't have the script replication, flush also seems
safe to add the flag.


enjoy-binbin:
I'm struggling with script-flush, it seem we can do the same thing (or can't)

madolson (approved):
Yeah, I don't see any reason why not to show the script contents while the server is stale.

zuiderkwast (approved):
Seems OK. Not sure why it's useful though. According to this comment, only a few commands are accepted during STALE:

```
 * CMD_STALE:       Allow the command while a replica has stale data but is not
 *                  allowed to serve this data. Normally no command is accepted
 *                  in this condition but just a few.
```

enjoy-binbin:
The comment is a bit outdated i guess, searing "STALE" in json files we can see we allow 140+ commands to run in a stale server. I guess it originally meant to say a few "keyspace or data" commands

enjoy-binbin:
what about script-flush? i am going to add STATE to it as well, i don't see the problem here since now we don't have the script replication, so even the user flush it all, it is his choice?

zuiderkwast (approved):

enjoy-binbin:
I see we already have three votes with the new change. This isn't particularly matter and it is harmless, so i don't think it will be a major decision and i am going to merge it.

Issue #2205, opened 2025-06-11
[test-failure] TIMEOUT: Active defrag big list: cluster


hpatro:
[TIMEOUT]: clients state report follows.
sock55a174240320 => (IN PROGRESS) Active Defrag big list: cluster
Killing still running Valkey server 32608
Killing still running Valkey server 32626
Killing still running Valkey server 38472
Killing still running Valkey server 47726

Observed on 8.1.2 release patch #2173:
https://github.com/valkey-io/valkey/actions/runs/15591253939/job/43910656808?pr=2173
https://github.com/valkey-io/valkey/actions/runs/15591253939/job/43910656725?pr=2173

zuiderkwast:
In https://github.com/valkey-io/valkey/issues/2398#issuecomment-3149373325, @sungming2 mentioned:

> I believe $rd read does blocking socket read() that stalls while engine is busy such as long defrag, and if no reply arrives within the socket timeout, Tcl raises an error and the harness logs “[TIMEOUT]”.

The test case client sends a massive pipeline of 200000 requests and then it does a blocking read. It just reads and discards all the replies. In #2483, I wait for the server after every 1000 requests.

Issue #2398, opened 2025-07-30
[Test Failure] Fix Flaky Tests in Workflows


sungming2:
We've observed intermittent failures in our CI workflows that appear to be caused by flaky tests. These failures do not consistently reproduce and often pass when re-run, making it difficult to identify true regressions that need to be investigated and fixed.

List of falky tests (keep it updated):

* https://github.com/valkey-io/valkey/issues/2423

* https://github.com/valkey-io/valkey/actions/runs/16585350083/job/46909359496
```
[err]: Active Defrag stream: cluster in tests/unit/memefficiency.tcl
defrag didn't stop.
```

* https://github.com/valkey-io/valkey/issues/2205
```
*** [TIMEOUT]: Active Defrag big list: cluster in tests/unit/memefficiency.tcl
```

* https://github.com/valkey-io/valkey/actions/runs/16540895816/job/46781944780
```
*** [TIMEOUT]: pid:38921 - tests/integration/aof.tcl in tests/integration/aof.tcl
```

* ~~https://github.com/valkey-io/valkey/actions/runs/16440028719/job/46458486359~~
  * Fixed https://github.com/valkey-io/valkey/pull/2370
```
!!! WARNING The following tests failed:

*** Migration source is auto-updated after failover in source shard in tests/unit/cluster/slot-migration.tcl
*** Replica redirects key access in migrating slots in tests/unit/cluster/slot-migration.tcl
*** Replica of migrating node returns ASK redirect after READONLY in tests/unit/cluster/slot-migration.tcl
*** Replica of migrating node returns TRYAGAIN after READONLY in tests/unit/cluster/slot-migration.tcl
*** Replica of importing node returns TRYAGAIN after READONLY and ASKING in tests/unit/cluster/slot-migration.tcl
*** New replica inherits importing slot in tests/unit/cluster/slot-migration.tcl
```

* ~~https://github.com/valkey-io/valkey/actions/runs/16256552218/job/45937858917#logs~~
  * Fixed https://github.com/valkey-io/valkey/pull/2345
```
/__w/_temp/cd6e3fe0-daaa-448e-b09f-76d9689007a0.sh: 2: /__w/_temp/cd6e3fe0-daaa-448e-b09f-76d9689007a0.sh: make: not found
```

sarthakaggarwal97:
https://github.com/valkey-io/valkey/actions/runs/16440028719/job/46458486359

[...]

PR #2421, opened 2025-08-04
Increase latency for big list test in defrag

sungming2:
Across multiple runs of the big list test in defrag, the latency check is tripping because the maximum observed defrag cycle latency occasionally spikes above our 5 ms limit. While most cycles complete in just a few milliseconds, rare slowdowns push some cycles into the double digit millisecond range, so a 5 ms hard cap is too aggressive for stable testing.

```
/runtest --verbose --tls --single unit/memefficiency --only '/big list' --accurate --loop --fastfail
```

```
[err]: Active Defrag big list: standalone in tests/unit/memefficiency.tcl
Expected 12 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
(Fast fail: test will exit now)

[err]: Active Defrag big list: standalone in tests/unit/memefficiency.tcl
Expected 21 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
(Fast fail: test will exit now)

[err]: Active Defrag big list: standalone in tests/unit/memefficiency.tcl
Expected 25 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
(Fast fail: test will exit now)

[err]: Active Defrag big list: standalone in tests/unit/memefficiency.tcl
Expected 17 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
(Fast fail: test will exit now)

```

### DIFF
--- a/tests/unit/memefficiency.tcl
+++ b/tests/unit/memefficiency.tcl
@@ -172,6 +172,8 @@ run_solo {defrag} {
     #    while_defragging {code} - optional, code executed after defrag has started
     #    latency <ms> - optional, verifies the latency to a ms target (default 5)
     proc perform_defrag_test {name args} {
+        # This value should be 5 ms, but it was flaky in github runners. The issue 
+        # https://github.com/valkey-io/valkey/issues/2444 is tracking if we can raise the limit again.
         set opts(latency) 40
         set opts(while_defragging) {}
         array set opts $args
--- a/tests/unit/memefficiency.tcl
+++ b/tests/unit/memefficiency.tcl
@@ -172,7 +172,7 @@ run_solo {defrag} {
     #    while_defragging {code} - optional, code executed after defrag has started
     #    latency <ms> - optional, verifies the latency to a ms target (default 5)
     proc perform_defrag_test {name args} {
-        set opts(latency) 5
+        set opts(latency) 40
         set opts(while_defragging) {}
         array set opts $args
         assert {[info exists opts(populate)]}
@@ -378,7 +378,7 @@ run_solo {defrag} {
             # number of total fields.  lists are progressively increasing sizes.
             set n 200000
 
-            perform_defrag_test $title latency 30 populate {
+            perform_defrag_test $title populate {
                 set rd [valkey_deferring_client]
                 set val [string repeat A 350]
                 set k 0
--- a/tests/unit/memefficiency.tcl
+++ b/tests/unit/memefficiency.tcl
@@ -378,7 +378,7 @@ run_solo {defrag} {
             # number of total fields.  lists are progressively increasing sizes.
             set n 200000
 
-            perform_defrag_test $title populate {
+            perform_defrag_test $title latency 30 populate {
                 set rd [valkey_deferring_client]
                 set val [string repeat A 350]
                 set k 0


zuiderkwast (approved):
LGTM

hpatro (approved):
Tagging @JimB123 to check if he has anything to add to this.

madolson:
Do we understand what the rare slow downs are? If it's related to the system (perhaps valkey-server is getting descheduled) then I would expect every test to need this latency increase. If it's specific to just the lists, we probably should investigate that, since it's probably some type of bug.

sungming2:
Let me check other tests too

zuiderkwast:
Fixes #2205?

zuiderkwast:
> Do we understand what the rare slow downs are?

According to https://github.com/valkey-io/valkey/issues/2398#issuecomment-3138684467 (and a linked Redis issue) it can be related to kernel versions. A kernel mitigation for some CVE. @arkamar might be able to explain more...(?)

sungming2:
> According to https://github.com/valkey-io/valkey/issues/2398#issuecomment-3138684467 (and a linked Redis issue) it can be related to kernel versions. A kernel mitigation for some CVE. @arkamar might be able to explain more...(?)

I am not sure it's specific to the OS/kernel version because I was able to get above latency on both Ubuntu 20.04 and 24.04

sungming2:
* big hash
```
[err]: Active Defrag big hash: standalone in tests/unit/memefficiency.tcl
Expected 12 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
(Fast fail: test will exit now)

[err]: Active Defrag big hash: standalone in tests/unit/memefficiency.tcl
Expected 18 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
(Fast fail: test will exit now)
```

* big set
```
[err]: Active Defrag big set: standalone in tests/unit/memefficiency.tcl
Expected 16 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
(Fast fail: test will exit now)

Test error (last server port:23623, log:./tests/tmp/server.152762.13/stdout), test will exit now
[err]: Active Defrag big set: standalone in tests/unit/memefficiency.tcl
Expected 12 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
```

* big zset
```

[err]: Active Defrag big zset: standalone in tests/unit/memefficiency.tcl
Expected 24 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
(Fast fail: test will exit now)

[err]: Active Defrag big zset: standalone in tests/unit/memefficiency.tcl
Expected 16 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
(Fast fail: test will exit now)
```

Can we consider increasing the default limit latency to ~~30~~ 40?

https://github.com/valkey-io/valkey/blob/34480ac35eaab8e08afecf68a53238a553098b93/tests/unit/memefficiency.tcl#L173-L175

zuiderkwast:
> Can we consider increasing the default limit latency to 30?

Yes, good idea.

madolson:
> According to https://github.com/valkey-io/valkey/issues/2398#issuecomment-3138684467 (and a linked Redis issue) it can be related to kernel versions. A kernel mitigation for some CVE. @arkamar might be able to explain more...(?)

This doesn't seem related. The original issue that was identified by arkamar is for timeouts, where as this is just taking slightly longer times.

madolson:
> Yes, good idea.

Setting it to 30ms almost defeats the purpose of the latency testing. The goal was to verify that active defragmentation isn't introducing latency spikes, but it seems like it is. Maybe we should allow 1 data point above the threshold?

sungming2:
> Maybe we should allow 1 data point above the threshold?

I think that is also good idea but IMO, allowing 1 spike might not generalize across machines, environments, or workloads. Also, future contributors may misunderstand or misuse the logic (e.g., tolerating 1 or 2..), and if a single spike exceeds a drastically high value such as 1 second, it could still be allowed under this rule and go undetected

madolson:
Maybe instead we could add a debug option to use ```clock_gettime(CLOCK_PROCESS_CPUTIME_ID)``` for testing? This should not include time when the process isn't being scheduled, which should hopefully remove some of the edge case times.

sungming2:
We are using the latency data reported by the server via the LATENCY command, which internally measures duration. Based on the implementation ([defrag.c#L1165-L1190](https://github.com/valkey-io/valkey/blob/unstable/src/defrag.c#L1165-L1190)), I believe this provides sufficiently accurate duration. I also double checked the source of the latency, and it is coming from `active-defrag-cycle`, not `while-blocked-cron`

https://github.com/valkey-io/valkey/blob/a50690f02c08e4e67b8b472f81587b05c405b6f7/tests/unit/memefficiency.tcl#L118-L137

zuiderkwast:
There may be other causes of latency spikes, such as synchronous IO, that also affect these tests? Such as

* #2424

madolson:
> There may be other causes of latency spikes, such as synchronous IO, that also affect these tests? Such as

I have a suspicion that it's related to the process being interrupted by the either the guest or host OS. The time will keep ticking up in those cases. We shouldn't be doing any synchronous IO in these tests.

zuiderkwast:
> I have a suspicion that it's related to the process being interrupted by the either the guest or host OS.

If this is true, we can't really do much about it, can we? Let's merge this PR for now to unflake the tests?

madolson:
> If this is true, we can't really do much about it, can we? Let's merge this PR for now to unflake the tests?

I suggested we might be able to use the processor clock to actually track time used by the process. If that doesn't work, I think we should remove the latency check and create an issue for later.

sungming2:
I am checking the cpu clock, will get the result soon

sungming2:
I tested with CLOCK_PROCESS_CPUTIME_ID and other options like CLOCK_MONOTONIC, but the latency spikes still appear:

```
[err]: Active Defrag big hash: standalone in tests/unit/memefficiency.tcl
Expected 21 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
(Fast fail: test will exit now)
[err]: Active Defrag big list: standalone in tests/unit/memefficiency.tcl
Expected 8 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
(Fast fail: test will exit now)
[err]: Active Defrag big set: standalone in tests/unit/memefficiency.tcl
Expected 10 <= 5 (context: type proc line 18 cmd {assert {$max_latency <= $limit_ms}} proc ::validate_latency level 1)
(Fast fail: test will exit now)
```

To investigate further, I wrote a simple loop to compare the timing behavior across different clocks,  CLOCK_PROCESS_CPUTIME_ID sometimes takes longer than others:

```
gettimeofday             : 1067362 us
CLOCK_REALTIME           : 1052658 us
CLOCK_MONOTONIC          : 1062182 us
CLOCK_PROCESS_CPUTIME_ID : 1073927 us
CLOCK_THREAD_CPUTIME_ID  : 1057760 us
```

I found similar issues from online indicating that CLOCK_PROCESS_CPUTIME_ID incurs more overhead, as it queries the scheduler subsystem (via syscall) for fine grained/per thread/process CPU usage. This makes it sometimes more expensive than wall time clocks.

madolson:
> I found similar issues from online indicating that CLOCK_PROCESS_CPUTIME_ID incurs more overhead, as it queries the scheduler subsystem (via syscall) for fine grained/per thread/process CPU usage. This makes it sometimes more expensive than wall time clocks.

What you showed is all less than 1% differences, so I don't think that is causing the spike (the spike is on the order of 10000us) So, it seems like we still aren't sure what is causing the issue :/

I'm OK closing this. I'll make an issue to track it, and then we can merge this. EDIT: new issue https://github.com/valkey-io/valkey/issues/2444

tests/unit/memefficiency.tcl
```diff
@@ -172,7 +172,7 @@ run_solo {defrag} {
     #    while_defragging {code} - optional, code executed after defrag has started
     #    latency <ms> - optional, verifies the latency to a ms target (default 5)
     proc perform_defrag_test {name args} {
-        set opts(latency) 5
+        set opts(latency) 40
```
madolson, line 177:
```suggestion
        # This value should be 5 ms, but it was flaky in github runners. The issue 
        # https://github.com/valkey-io/valkey/issues/2444 is tracking if we can raise the limit again.
        set opts(latency) 40
```

madolson (approved):

PR #2422, opened 2025-08-04
bump the RDB version

ranshid:
This is needed due to changes presented in https://github.com/valkey-io/valkey/pull/2089

### DIFF
--- a/src/aof.c
+++ b/src/aof.c
@@ -1437,9 +1437,10 @@ int loadSingleAppendOnlyFile(char *filename) {
     /* Check if the AOF file is in RDB format (it may be RDB encoded base AOF
      * or old style RDB-preamble AOF). In that case we need to load the RDB file
      * and later continue loading the AOF tail if it is an old style RDB-preamble AOF. */
-    char sig[5]; /* "REDIS" */
-    if (fread(sig, 1, 5, fp) != 5 || memcmp(sig, "REDIS", 5) != 0) {
+    char sig[6]; /* "REDIS" or "VALKEY" */
+    if (fread(sig, 1, 6, fp) != 6 || (memcmp(sig, "REDIS", 5) != 0 && memcmp(sig, "VALKEY", 6) != 0)) {
         /* Not in RDB format, seek back at 0 offset. */
+        serverLog(LL_WARNING, "Error reading the RDB preamble signature of the AOF file, AOF loading aborted");
         if (fseek(fp, 0, SEEK_SET) == -1) goto readerr;
     } else {
         /* RDB format. Pass loading the RDB functions. */
--- a/src/aof.c
+++ b/src/aof.c
@@ -1440,7 +1440,6 @@ int loadSingleAppendOnlyFile(char *filename) {
     char sig[6]; /* "REDIS" or "VALKEY" */
     if (fread(sig, 1, 6, fp) != 6 || (memcmp(sig, "REDIS", 5) != 0 && memcmp(sig, "VALKEY", 6) != 0)) {
         /* Not in RDB format, seek back at 0 offset. */
-        serverLog(LL_WARNING, "Error reading the RDB preamble signature of the AOF file, AOF loading aborted");
         if (fseek(fp, 0, SEEK_SET) == -1) goto readerr;
     } else {
         /* RDB format. Pass loading the RDB functions. */
--- a/src/aof.c
+++ b/src/aof.c
@@ -1438,7 +1438,7 @@ int loadSingleAppendOnlyFile(char *filename) {
      * or old style RDB-preamble AOF). In that case we need to load the RDB file
      * and later continue loading the AOF tail if it is an old style RDB-preamble AOF. */
     char sig[6]; /* "REDIS" or "VALKEY" */
-    if (fread(sig, 1, 6, fp) != 6 || (memcmp(sig, "REDIS", 5) != 0 && memcmp(sig, "VALKEY", 6) != 0)) {
+    if (fread(sig, 1, 6, fp) != 6 || (memcmp(sig, "REDIS0", 6) != 0 && memcmp(sig, "VALKEY", 6) != 0)) {
         /* Not in RDB format, seek back at 0 offset. */
         if (fseek(fp, 0, SEEK_SET) == -1) goto readerr;
     } else {
--- a/src/rdb.c
+++ b/src/rdb.c
@@ -3036,14 +3036,14 @@ int rdbLoadRioWithLoadingCtx(rio *rdb, int rdbflags, rdbSaveInfo *rsi, rdbLoadin
     char buf[1024];
     int error;
     long long empty_keys_skipped = 0;
-    bool is_valkey_magic;
+    bool is_valkey_magic = false, is_redis_magic = false;
 
     rdb->update_cksum = rdbLoadProgressCallback;
     rdb->max_processing_chunk = server.loading_process_events_interval_bytes;
     if (rioRead(rdb, buf, 9) == 0) goto eoferr;
     buf[9] = '\0';
     if (memcmp(buf, "REDIS0", 6) == 0) {
-        is_valkey_magic = false;
+        is_redis_magic = true;
     } else if (memcmp(buf, "VALKEY", 6) == 0) {
         is_valkey_magic = true;
     } else {
@@ -3051,7 +3051,7 @@ int rdbLoadRioWithLoadingCtx(rio *rdb, int rdbflags, rdbSaveInfo *rsi, rdbLoadin
         return C_ERR;
     }
     rdbver = atoi(buf + 6);
-    if (rdbver < 1 ||
+    if (rdbver < 1 || (rdbver < RDB_FOREIGN_VERSION_MIN && !is_redis_magic) ||
         (rdbver >= RDB_FOREIGN_VERSION_MIN && rdbver <= RDB_FOREIGN_VERSION_MAX) ||
         (rdbver > RDB_FOREIGN_VERSION_MAX && !is_valkey_magic) ||
         (rdbver > RDB_VERSION && server.rdb_version_check == RDB_VERSION_CHECK_STRICT)) {
--- a/src/rdb.c
+++ b/src/rdb.c
@@ -1429,8 +1429,7 @@ int rdbSaveRio(int req, rio *rdb, int *error, int rdbflags, rdbSaveInfo *rsi) {
     int j;
 
     if (server.rdb_checksum) rdb->update_cksum = rioGenericUpdateChecksum;
-    /* TODO: Change this to "VALKEY%03d" next time we bump the RDB version. */
-    snprintf(magic, sizeof(magic), "REDIS%04d", RDB_VERSION);
+    snprintf(magic, sizeof(magic), "VALKEY%03d", RDB_VERSION);
     if (rdbWriteRaw(rdb, magic, 9) == -1) goto werr;
     if (rdbSaveInfoAuxFields(rdb, rdbflags, rsi) == -1) goto werr;
     if (!(req & REPLICA_REQ_RDB_EXCLUDE_DATA) && rdbSaveModulesAux(rdb, VALKEYMODULE_AUX_BEFORE_RDB) == -1) goto werr;
--- a/src/rdb.c
+++ b/src/rdb.c
@@ -3052,7 +3052,7 @@ int rdbLoadRioWithLoadingCtx(rio *rdb, int rdbflags, rdbSaveInfo *rsi, rdbLoadin
     }
     rdbver = atoi(buf + 6);
     if (rdbver < 1 ||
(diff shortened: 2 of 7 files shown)


src/rdb.h
```diff
@@ -42,15 +42,15 @@
  * RDB 11 is the last open-source Redis RDB version, used by Valkey 7.x and 8.x.
  *
  * RDB 12+ are non-open-source Redis formats.
+ * VALKEY 80+ are Valkey open-source versions.
```
zuiderkwast: Mention which Valkey versions use this? We already mentioned above that Valkey 7.x and 8.x use RDB 11.

```suggestion
 * VALKEY 80+ are Valkey >= 9.0.
```

No need to mention open source. All Valkeys are open source. :)

Btw now that Redis added AGPL, maybe we should change the comment about Redis to say restrictive-licensed or non-BSD licenced?

src/aof.c
```diff
@@ -1437,9 +1437,10 @@ int loadSingleAppendOnlyFile(char *filename) {
     /* Check if the AOF file is in RDB format (it may be RDB encoded base AOF
      * or old style RDB-preamble AOF). In that case we need to load the RDB file
      * and later continue loading the AOF tail if it is an old style RDB-preamble AOF. */
-    char sig[5]; /* "REDIS" */
-    if (fread(sig, 1, 5, fp) != 5 || memcmp(sig, "REDIS", 5) != 0) {
+    char sig[6]; /* "REDIS" or "VALKEY" */
+    if (fread(sig, 1, 6, fp) != 6 || (memcmp(sig, "REDIS", 5) != 0 && memcmp(sig, "VALKEY", 6) != 0)) {
         /* Not in RDB format, seek back at 0 offset. */
+        serverLog(LL_WARNING, "Error reading the RDB preamble signature of the AOF file, AOF loading aborted");
```
zuiderkwast: This is not necessarily an error? Can it be an AOF without preamble?

It's what the comment says if I understand it correctly. If it's not an RDB preamble, we fseek to offset 0 and start reading AOF commands directly.

So I guess we should not add this logging.

ranshid: Yes. Sorry left it there when I attempted debugging the failures :)

src/valkey-check-rdb.c
```diff
@@ -612,11 +613,16 @@ int redis_check_rdb(char *rdbfilename, FILE *fp) {
...
+        rdb_version_offset = 5;
+    } else if (memcmp(buf, "VALKEY", 6) == 0) {
+        rdb_version_offset = 6;
+    } else {
         rdbCheckError("Wrong signature trying to load DB from file");
         goto err;
     }
-    rdbver = atoi(buf + 5);
+
+    rdbver = atoi(buf + rdb_version_offset);
```
zuiderkwast: Just an idea: We can treat the magic as 6 characters in both cases.

```c
    if (memcmp(buf, "REDIS0", 6) != 0 && memcmp(buf, "VALKEY", 6) != 0) {
        rdbCheckError("Wrong signature trying to load DB from file");
        goto err;
    }
    rdbver = atoi(buf + 6);
```

src/valkey-check-rdb.c
```diff
@@ -612,11 +613,16 @@ int redis_check_rdb(char *rdbfilename, FILE *fp) {
...
     if (rdbver < 1 || rdbver > RDB_VERSION) {
         rdbCheckError("Can't handle RDB format version %d", rdbver);
```
zuiderkwast: Should we error also on rdbver 12-79 here?

zuiderkwast (approved):

src/valkey-check-rdb.c
ranshid: Yeh - I thought about it, but do we know for sure it will always have a leading 0? if so probably it will be O.K to atio with 6...

ranshid: @zuiderkwast done. TU!

src/valkey-check-rdb.c
ranshid: that's a good point. we do not want to support checking redis RDB I guess...

sarthakaggarwal97 (approved):
nice!

src/valkey-check-rdb.c
ranshid: should probably be handled in rdbLoadRioWithLoadingCtx as well

ranshid: @zuiderkwast done.

ranshid: wait... not done (I am just tiered)

ranshid: @zuiderkwast maybe check the last one now?

madolson (approved):

src/valkey-check-rdb.c
zuiderkwast: Yeah since we only support REDIS up to number 11.

zuiderkwast (approved):

src/rdb.c
```diff
@@ -3053,7 +3052,8 @@ int rdbLoadRioWithLoadingCtx(rio *rdb, int rdbflags, rdbSaveInfo *rsi, rdbLoadin
     }
     rdbver = atoi(buf + 6);
     if (rdbver < 1 ||
-        (rdbver >= RDB_FOREIGN_VERSION_MIN && !is_valkey_magic) ||
+        (rdbver >= RDB_FOREIGN_VERSION_MIN && rdbver <= RDB_FOREIGN_VERSION_MAX) ||
+        (rdbver > RDB_FOREIGN_VERSION_MAX && !is_valkey_magic) ||
         (rdbver > RDB_VERSION && server.rdb_version_check == RDB_VERSION_CHECK_STRICT)) {
         serverLog(LL_WARNING, "Can't handle RDB format version %d", rdbver);
```
madolson, line 3058: We should probably extend this to include whether or not it's the Redis magic as well?

ranshid: @madolson to which one of the check conditions do you refer to? The [12-79] or just validate that below 12 we have the redis magic?

ranshid: @madolson done. want to take a look?

ranshid:
@valkey-io/core-team just to highlight the implications of this: this basically breaks the 7.2 and 8.0 compatibility. 
IIRC we only added the strict version check in 8.1 which will now become the only downgradable version.
Should we include some wording about it in the documentation? I expect it is going to be somewhat confusing.
Also, I expect some of the  Cross version cluster tests will now fail. I think it is expected, but would be happy to hear if you feel otherwise.

ranshid:
I think that the only issue is with the [failover test](https://github.com/valkey-io/valkey/blob/unstable/tests/unit/cluster/cross-version-cluster.tcl#L9-L32).
This test will attempt to first sync-up an old version which is now unsupported. It will probably not be an issue for the daily compat tests as they are being run with `compatible-redis` tag. 

I think that for the CI tests we can run only against 8.1 and enable the relaxed rdb check in this specific test.

@zuiderkwast WDYT?

zuiderkwast:
> IIRC we only added the strict version check in 8.1 which will now become the only downgradable version.

Ran, yes, because of this, we consider 8.1 a *stepping stone*. All our customers must upgrade to (a version of our products that contain) 8.1 before they can upgrade further to 9 and later.

> I think that for the CI tests we can run only against 8.1 and enable the relaxed rdb check in this specific test.

Yes.

We have discussed a way to let Valkey produce older RDB versions for even better downgrade possibilities. It would involve a config to set the RDB version and it would be up to the user to only use features supported by the chosen RDB version. I still think we should do it. Technically, it's not very difficult... #1108.

ranshid:
> > IIRC we only added the strict version check in 8.1 which will now become the only downgradable version.
> 
> Ran, yes, because of this, we consider 8.1 a _stepping stone_. All our customers must upgrade to (a version of our products that contain) 8.1 before they can upgrade further to 9 and later.
> 
> > I think that for the CI tests we can run only against 8.1 and enable the relaxed rdb check in this specific test.
> 
> Yes.
> 
> We have discussed a way to let Valkey produce older RDB versions for even better downgrade possibilities. It would involve a config to set the RDB version and it would be up to the user to only use features supported by the chosen RDB version. I still think we should do it. Technically, it's not very difficult... #1108.

Thank you @zuiderkwast . I am mainly referring to the plan to change the CI as part of THIS pr so that we will start running the cross version compat against 8.1 + configure relaxed RDB check in the [failover test](https://github.com/valkey-io/valkey/blob/unstable/tests/unit/cluster/cross-version-cluster.tcl#L9-L32)

ranshid:
hhh There is a failure in test `Send cluster message via module from other server to latest`, but it is because this test should have fail IMO. It is based on the fact that `CONFIG resetstat` will reset the cluster msg stats (which is not true). I actually think that there might be a bug in the use of light headers in this cross version case? will check

@hpatro FYI

ranshid:
> hhh There is a failure in test `Send cluster message via module from other server to latest`, but it is because this test should have fail IMO. It is based on the fact that `CONFIG resetstat` will reset the cluster msg stats (which is not true). I actually think that there might be a bug in the use of light headers in this cross version case? will check
> 
> @hpatro FYI

O.K after some fighting, I found the issue is known (https://github.com/valkey-io/valkey/pull/1708) and that is the node-id is passed NULL terminated only from 8.1 to clusterMassageRecievers. because of that the 7.2 module is unable to send the DONG message back to the originator and this is why the test fails for 7.2. So there are some AIs:
1. fix the [failover test](https://github.com/valkey-io/valkey/blob/unstable/tests/unit/cluster/cross-version-cluster.tcl#L9-L32) as it is now not testing correctly the message transfer.
2. @zuiderkwast / @hpatro seems like you guys have context on the issue. did we decide to backport this? if not what do we expect?
I guess after enabling this test to work with 8.1 there will be no failure though

zuiderkwast (approved):

src/rdb.c
madolson: Apparently not :D

madolson: I also meant we should print out whether or not the version is incompatible or it's the Redis Magic.

zuiderkwast: Since we don't reuse the same numbers as with the different magics, it will never be ambiguous (except for hand-hacked RDB files with wrong magic/number).

tests/unit/moduleapi/cross-version-cluster.tcl
```diff
@@ -33,24 +42,18 @@ tags {external:skip needs:other-server cluster modules singledb} {
...
-                assert_match "*cluster_stats_messages_module_sent:1*" [r -1 CLUSTER INFO]
-                wait_for_condition 50 100 {
-                    [string match {*cluster_stats_messages_module_received:1*} [r CLUSTER INFO]]
-                } else {
-                    fail "Node didn't receive the message"
-                }
+                wait_for_cluster_stat 1 messages_module_sent 2
+                wait_for_cluster_stat 0 messages_module_received 2
+                wait_for_cluster_stat 0 messages_module_sent 2
+                wait_for_cluster_stat 1 messages_module_received 2
```
hpatro, line 56: I was going to look into this further, I observed it while updating the cross version test to work with other versions and latest as well. https://github.com/valkey-io/valkey/pull/2342. Thanks for deep diving on this.

src/rdb.h
```diff
@@ -41,16 +41,15 @@
  *
  * RDB 11 is the last open-source Redis RDB version, used by Valkey 7.x and 8.x.
  *
- * RDB 12+ are non-open-source Redis formats.
+ * RDB 12-79 are reserved for Redis non-compatible RDB formats
```
hpatro, line 44: Any particular reason to pick 79? 

With Valkey 9.0 we could have picked 90 as our new RDB version :).

ranshid: This was discussed in the last core-meeting and we decided on 80. Personally I think the 80's (although the bad hair thing) were so cool

tests/unit/moduleapi/cross-version-cluster.tcl
ranshid: I think however we should have 2 followup items:
1. make config resetstats reset the cluster message stats
2. fix the test module so it will always provide null terminated node-id.
Will try to remember open an issue/PR tomorrow

src/rdb.c
madolson: For us yes, not for end users. Ideally we should print out that this is a Redis RDB and not a valkey RDB.

src/rdb.h
zuiderkwast: 80+ for Valkey magic was picked in #1604. It was arbitrary-ish.

Yeah we could have used 90 for Valkey 9.0 but with this scheme we would run out of RDB versions after Valkey 99.9. If we just bumb it when necessary, we don't run out of RDB numbers as fast.

tests/unit/moduleapi/cross-version-cluster.tcl
zuiderkwast: * 2 is fixed in #2484

PR #2425, opened 2025-08-05
Skip sorting empty lists in sortCommandGeneric to prevent undefined behavior when calling pqsort

Fusl:
This one was found by [afl++](https://github.com/AFLplusplus/AFLplusplus). Executing `sort key by *` with a non-existing key, `vectorlen` gets initially set to `0` and then set to `-1` at `vectorlen = end - start + 1;` which then gets passed to `pqsort` as a `size_t` causing an underflow. Since we're operating on an empty list here, we can just skip the entire sorting step altogether and either return the expected zero-length array or unlink the store key in the database.

### DIFF
--- a/src/sort.c
+++ b/src/sort.c
@@ -506,7 +506,7 @@ void sortCommandGeneric(client *c, int readonly) {
         server.sort_alpha = alpha;
         server.sort_bypattern = sortby ? 1 : 0;
         server.sort_store = storekey ? 1 : 0;
-        /* If the source keys are empty, we have will have a vector length of zero,
+        /* If the source keys are empty, we will have a vector length of zero,
          * so no need to sort. */
         if (vectorlen != 0) {   
             if (sortby && (start != 0 || end != vectorlen - 1))
--- a/src/sort.c
+++ b/src/sort.c
@@ -508,7 +508,7 @@ void sortCommandGeneric(client *c, int readonly) {
         server.sort_store = storekey ? 1 : 0;
         /* If the source keys are empty, we will have a vector length of zero,
          * so no need to sort. */
-        if (vectorlen != 0) {   
+        if (vectorlen != 0) {
             if (sortby && (start != 0 || end != vectorlen - 1))
                 pqsort(vector, vectorlen, sizeof(serverSortObject), sortCompare, start, end);
             else
--- a/src/sort.c
+++ b/src/sort.c
@@ -302,13 +302,27 @@ void sortCommandGeneric(client *c, int readonly) {
         return;
     }
 
+    /* If we have nothing to sort (the specified key doesn't exist), we return
+     * an empty array, or, if STORE is used, delete the target key and return 0. */
+    if (!sortval) {
+        listRelease(operations);
+        if (storekey == NULL) {
+            addReplyArrayLen(c, 0);
+        } else {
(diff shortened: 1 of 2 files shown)


tests/unit/sort.tcl
```diff
@@ -204,6 +204,11 @@ foreach command {SORT SORT_RO} {
         assert_equal [lsort -real $floats] [r sort mylist]
     }
 
+    test "SORT returns an empty array if the list is empty" {
+        r flushdb
+        r sort mylist
```
madolson, line 209: Running this with gdb, vectorlen for both `sort mylist` and `sort mylist by *` both showed vectorlen with a value of zero. I didn't spend too much time validating, but I couldn't quite trace the path that caused this to be -1.

Fusl: My bad, it's not vectorlen that underflows, it's start/end.

madolson: I see! That makes more sense.

Instead of adding a new special return flow, I think it might be a bit cleaner to avoid calling pqsort when start or end is negative, which seems to only happen on empty strings. We could also clamp the negative values to zero.

Fusl: I specifically made this an early return since we can avoid multiple allocations and other semi-expensive calls this way if the key doesn't exist at all.

madolson: I agree it's more expensive, but now there are two more exit paths (SORT with/without store) that we need to make sure are aligned with the code path at the end of the function. SORT ultimately is not that much of a performance critical function. The alternative might be a goto or a helper method so the cleanup code is common.

src/sort.c
```diff
@@ -506,10 +506,14 @@ void sortCommandGeneric(client *c, int readonly) {
         server.sort_alpha = alpha;
         server.sort_bypattern = sortby ? 1 : 0;
         server.sort_store = storekey ? 1 : 0;
-        if (sortby && (start != 0 || end != vectorlen - 1))
-            pqsort(vector, vectorlen, sizeof(serverSortObject), sortCompare, start, end);
-        else
-            qsort(vector, vectorlen, sizeof(serverSortObject), sortCompare);
+        /* If the source keys are empty, we have will have a vector length of zero,
```
Fusl:
```suggestion
        /* If the source keys are empty, we will have a vector length of zero,
```

src/sort.c
```diff
@@ -506,10 +506,14 @@ void sortCommandGeneric(client *c, int readonly) {
...
+         * so no need to sort. */
+        if (vectorlen != 0) {
```
madolson:
```suggestion
        if (vectorlen != 0) {
```

madolson:
We finally saw this prop in a daily failure, https://github.com/valkey-io/valkey/actions/runs/20629592512, so felt it was better to merge this one regardless. If you still feel strongly about your original proposal, I'm open to it. Both changes seem to resolve https://github.com/valkey-io/valkey/actions/runs/20670751250.

madolson (approved):

Fusl:
This lgtm, thank you for taking the time to address this :heart:

madolson:
> This lgtm, thank you for taking the time to address this ❤️

Thank You!

PR #2427, opened 2025-08-05
Make ./runtest --dump-logs dump logs on valgrind / sanitizer errors

enjoy-binbin:
Sometimes we want to print server logs even if valgrind or sanitizer
errors occur, which might help troubleshoot the problem.


enjoy-binbin:
For example, this cluster test does not print the server log and cannot analyze what happened.
https://github.com/valkey-io/valkey/actions/runs/16720515261/job/47323708440?pr=2421#step:5:4018

zuiderkwast (approved):

Issue #2297, opened 2025-07-02
Migrate tests under tests/cluster to tests/unit/cluster


hpatro:
Migrate tests under tests/cluster to tests/unit/cluster to use the same framework all the cluster tests.

### Test file migration:

- [ ] 03-failover-loop
- [ ] 04-resharding
- [ ] 07-replica-migration
- [ ] 12-replica-migration-2
- [ ] 12.1-replica-migration-3
- [ ] 28-cluster-shards

### Usage Cleanup:

- [ ] runtest-cluster and it's access across build files and continuous integration (yml)

enjoy-binbin:
I remember the reason I didn't move them at the time was because there were some dependent functions that we didn't implement in tests/unit/, and these dependent functions didn't seem to be easy to move over. (But it should be much better now around these day)

After we clean it up, we can make runtest-cluster == runtest --cluster

hpatro:
> After we clean it up, we can make runtest-cluster == runtest --cluster

I have this old PR which introduces filtering by file/directory which will be quite handy I believe. Updated today. Please have a look at #630 @enjoy-binbin

hpatro:
> I remember the reason I didn't move them at the time was because there were some dependent functions that we didn't implement in tests/unit/, and these dependent functions didn't seem to be easy to move over. (But it should be much better now around these day)

Do you mean with the help of AI agents? 😅

enjoy-binbin:
> Do you mean with the help of AI agents?

nope, i guess this `But it should be much better now around these day` sentence gave the wrong assumption.

I mean, we've all embraced `tests/unit/cluster` a lot now, so maybe these moves will be easier than they were back then.

enjoy-binbin:
We actually have this...
```
    } elseif {$opt eq {--cluster}} {
        set ::all_tests $::cluster_all_tests
    } elseif {$opt eq {--moduleapi}} {
        set ::all_tests $::module_api_all_tests
```

ozzyozbourne:
I would like to work on this issue

ozzyozbourne:
hello @enjoy-binbin , @hpatro Iam having difficulties in porting "get_instance_attrib" function from the instances.tcl file, 
is there any reference I can look at ?

enjoy-binbin:
maybe see proc srv can do the help

ozzyozbourne:
@enjoy-binbin @hpatro Hello, do the old tests needs to have any tags and overrrides present when migrating to the unit/cluster 
[...]

PR #2428, opened 2025-08-05
Print the complete log instead of just the crash log in log_crashes

enjoy-binbin:
We used to, for example in runtest-cluster, we will only print the crash
log since in here, when we match the pattern, we start setting found and
then print the log starting from the current line. We can print the complete
log in this case and it might help troubleshoot the crash.

### DIFF
--- a/tests/instances.tcl
+++ b/tests/instances.tcl
@@ -187,17 +187,15 @@ proc log_crashes {} {
     set logs [glob */log.txt]
     foreach log $logs {
         set fd [open $log]
-        set found 0
         while {[gets $fd line] >= 0} {
             if {[string match $start_pattern $line]} {
                 puts "\n*** Crash report found in $log ***"
-                set found 1
-            }
-            if {$found} {
-                puts $line
+                puts [exec cat $log]
                 incr ::failed
+                break
             }
         }
+        close $fd
     }
 
     set logs [glob */err.txt]


enjoy-binbin:
The old output:
https://github.com/valkey-io/valkey/actions/runs/16721768957/job/47327608338?pr=2227#step:7:372

tests/instances.tcl
```diff
@@ -187,17 +187,15 @@ proc log_crashes {} {
...
         set fd [open $log]
-        set found 0
         while {[gets $fd line] >= 0} {
             if {[string match $start_pattern $line]} {
                 puts "\n*** Crash report found in $log ***"
-                set found 1
-            }
-            if {$found} {
-                puts $line
+                puts [exec cat $log]
```
zuiderkwast, line 193: There must be a better way to get the file contents than to exec the separate program `cat`. How about `seek` + `read`?

```suggestion
                seek $log 0
                puts [read $log]
```

zuiderkwast (commented):
There are other ways to print the complete log, isn't it? Run with `--dump-logs` prints the complete log?

tests/instances.tcl
enjoy-binbin: ./runtest-cluster (or ./runtest-sentinel) does not have the dump-logs option, it is very old and we don't maintain it a lot. Most ot the cluster test are migrate to normal test suite, some are left in #2297, we may completely remove it in the future (maybe not possible, because there is still sentinel test..)

we doing this exec cat thing a lot, like:
```
proc dump_server_log {srv} {
    set pid [dict get $srv "pid"]
    puts "\n===== Start of server log (pid $pid) =====\n"
    puts [exec cat [dict get $srv "stdout"]]
    puts "===== End of server log (pid $pid) =====\n"

    puts "\n===== Start of server stderr log (pid $pid) =====\n"
    puts [exec cat [dict get $srv "stderr"]]
    puts "===== End of server stderr log (pid $pid) =====\n"
}
```

seek + read would work too, which one do you prefer now?

zuiderkwast: So `log_crashes` is not used in the normal `runtest`?

> seek + read would work too, which one do you prefer now?

Ah, I don't care which one. :laughing:

zuiderkwast (approved):

tests/instances.tcl
enjoy-binbin: > So log_crashes is not used in the normal runtest?

yes, i think so. Normal runtest has it own cleanup proc there.

PR #963, opened 2024-08-29
Exclude '.' and ':' from `isValidAuxChar`'s banned charset

PingXie:
Fix a bug in isValidAuxChar where valid characters '.' and ':' were incorrectly included in the banned charset. This issue affected the validation of auxiliary fields in the nodes.conf file used by Valkey in cluster mode, particularly when handling IPv4 and IPv6 addresses. The code now correctly allows '.' and ':' as valid characters, ensuring proper handling of these fields. Comments were added to clarify the use of the banned charset.
 
Related to #736

PR #736, opened 2024-07-02
Dual stack and client-specific IPs in cluster

zuiderkwast:
New configs:

* `cluster-announce-client-ipv4`
* `cluster-announce-client-ipv6`

New module API function:

* `ValkeyModule_GetClusterNodeInfoForClient`, takes a client id and is otherwise just like its non-ForClient cousin.

If configured, one of these IP addresses are reported to each client in CLUSTER SLOTS, CLUSTER SHARDS, CLUSTER NODES and redirects, replacing the IP (`custer-announce-ip` or the auto-detected IP) of each node. Which one is reported to the client depends on whether the client is connected over IPv4 or IPv6.

Benefits:

* This allows clients using IPv4 to get the IPv4 addresses of all cluster nodes and IPv6 clients to get the IPv6 addresses.
* This allows the IPs visible to clients to be different to the IPs used between the cluster nodes due to NAT'ing.

The information is propagated in the cluster bus using new Ping extensions. (Old nodes without this feature ignore unknown Ping extensions.)

This adds another dimension to CLUSTER SLOTS reply. It now depends on the client's use of TLS, the IP address family and RESP version. Refactoring: The cached connection type definition is moved from connection.h (it actually has nothing to do with the connection abstraction) to server.h and is changed to a bitmap, with one bit for each of TLS, IPv6 and RESP3.

Fixes #337

PR #2429, opened 2025-08-05
Add cluster-announce-client-(port|tls-port) configs

### LINKED ISSUE #2377: [NEW] Announce different ports to clients in cluster

We are running/planning to run Valkey in cluster mode on Kubernetes with external traffic being ingested via `NodePort` services. Thus, the announced ports for a Valkey cluster node should be its assigned node port. Meanwhile, the cluster internally—both for the cluster bus and replication—should ideally use direct pod-to-pod communication with whatever port the `NodePort` service(s) map to. This should apply to mixed plaintext/TLS ports as well as TLS-only operation.

Currently, the `cluster-announce-port` and `cluster-announce-tls-port` options exist to change which ports are exposed "externally" for a node, but there is no distinction between "end user" client connections and intra-cluster ones. So if the announced port is set to the exposed Kubernetes node port, replication will fail as replicas try to connect to their primary via the pod IP but a port it doesn't listen to.

Since clients generally don't need to access the cluster bus, `cluster-announce-bus-port` doesn't necessarily need the same attention.

Akin to #337 and its implementation to #736, add two (or three) new options to complement `cluster-announce-client-ipv4`/`cluster-announce-client-ipv6`:
 * `cluster-announce-client-port`: client-visible version of `cluster-announce-port`
 * `cluster-announce-client-tls-port`: client-visible version of `cluster-announce-tls-port`
 * Maybe `cluster-announce-client-bus-port`: client-visible version of `cluster-announce-bus-port`

These will be used in place of the default announced ports in response to `CLUSTER NODES` commands, `MOVED` responses etc., while the default ones continue to be used for cluster-internal operations.

One alternative we considered was not using TLS for replication, so the `cluster-announce-port` can be internal-only while the TLS port is used for external communication. We'd like to enable TLS for everything though to also make use of usernames derived from client certificates as implemented in #1920 and completely disable password-based authentication.

Another alternative would be to have each cluster node pod listen on its assigned Kubernetes node port so they match up "coincidentally" with the announced ports. This can work as a stopgap for us, but might not be viable in all setups in general.

marvin-roesch:
New config options:
 * `cluster-announce-client-port`
 * `cluster-announce-client-tls-port`

If enabled, clients will always get to see the configured port for a node instead of the internally announced port(s), the same way that `cluster-announce-client-ipv4` and `cluster-announce-client-ipv6` work. Cluster-internal communication uses the non-client variant of these options.

The configuration is propagated throughout the cluster using new ping extensions.

Closes #2377


madolson:
Viktor from the issue thread:

> I like these new configs, but I think we should skip the cluster bus port. The bus is not for clients.
>
> We already have cluster-announce-bus-port that is used together with cluster-announce-ip for the cluster bus.

marvin-roesch:
I went ahead and added an easily reversible commit that removes the bus port announcement change again. The main rationale behind including it in the first place was that the bus port *is* visible to clients as part of the `CLUSTER NODES` response.

I'm not sure why the external tests were failing. They ran fine for me locally and I could not see anything related to my changes in the failed test logs.

src/cluster_legacy.c
```diff
@@ -3023,6 +3107,19 @@ writeSdsPingExtIfNonempty(uint32_t *totlen_ptr, clusterMsgPingExt **cursor_ptr,
     return 1;
 }
 
+static uint32_t
+writePortPingExtIfNonzero(uint32_t *totlen_ptr, clusterMsgPingExt **cursor_ptr, clusterMsgPingtypes type, uint16_t value) {
+    if (value == 0) return 0;
+    size_t size = getAlignedPingExtSize(sizeof(clusterMsgPingExtClientPort));
+    if (*cursor_ptr != NULL) {
+        void *ext = preparePingExt(*cursor_ptr, type, size);
+        memcpy(ext, &value, sizeof(value));
```
zuiderkwast, line 3119: We need to handle endianess. All numbers in the cluster bus are stored in network byte order (big endian). Let's convert value before memcpy?

```suggestion
        value = htons(value);
        memcpy(ext, &value, sizeof(value));
```

src/cluster_legacy.c
```diff
@@ -3118,6 +3221,14 @@ void clusterProcessPingExtensions(clusterMsg *hdr, clusterLink *link) {
             clusterMsgPingExtClientIpV6 *clientipv6_ext =
                 (clusterMsgPingExtClientIpV6 *)&(ext->ext[0].announce_client_ipv6);
             ext_clientipv6 = clientipv6_ext->announce_client_ipv6;
+        } else if (type == CLUSTERMSG_EXT_TYPE_CLIENT_PORT) {
+            clusterMsgPingExtClientPort *clientport_ext =
+                (clusterMsgPingExtClientPort *)&(ext->ext[0].announce_client_port);
+            ext_clientport = clientport_ext->announce_client_port;
```
zuiderkwast: And here we need to convert endianess from network to host byte order, right?

```suggestion
            ext_clientport = ntohs(clientport_ext->announce_client_port);
```

src/cluster_legacy.c
```diff
@@ -3118,6 +3221,14 @@ void clusterProcessPingExtensions(clusterMsg *hdr, clusterLink *link) {
...
+        } else if (type == CLUSTERMSG_EXT_TYPE_CLIENT_TLS_PORT) {
+            clusterMsgPingExtClientTlsPort *clienttlsport_ext =
+                (clusterMsgPingExtClientTlsPort *)&(ext->ext[0].announce_client_tls_port);
+            ext_clienttlsport = clienttlsport_ext->announce_client_tls_port;
```
```suggestion
            ext_clienttlsport = ntohs(clienttlsport_ext->announce_client_tls_port);
```

tests/unit/cluster/announce-client-ports.tcl
```diff
@@ -0,0 +1,93 @@
...
+        assert_match "MOVED * *:*" $e
+        set dest_port [lindex [split $e ":"] end]
+        assert {[lsearch -exact $original_ports $dest_port] >= 0}
+
+        # Now that everything is propagated, assert everyone agrees
+        wait_for_cluster_propagation
+    }
+}
+
+start_cluster 1 0 {tags {external:skip cluster} overrides {cluster-replica-no-failover yes bind {127.0.0.1 ::1}}} {
```
zuiderkwast: If we bind to `::1` we should also include `ipv6` in tags, so the test can be skipped wherever IPv6 is not available. Or maybe we don't need IPv6 in this test? We only care about port, not ip, right?

Also no need to prevent failover in a cluster with only one node.

```suggestion
start_cluster 1 0 {tags {external:skip cluster}} {
```

tests/unit/cluster/announce-client-ports.tcl
```diff
@@ -0,0 +1,93 @@
...
+    test "Load cluster announced client port config on server start" {
+        R 0 config set cluster-announce-client-port 6380
+        restart_server 0 true false
+    }
```
zuiderkwast: What is this test actually checking? It sets a config and restarts the server, but it doesn't check anything after restarting. Shouldn't it check what `config get` returns or check in some other way?

tests/unit/cluster/announce-client-ports.tcl
```diff
@@ -0,0 +1,93 @@
...
+}
+
+start_cluster 1 0 {tags {external:skip cluster} overrides {cluster-replica-no-failover yes bind {127.0.0.1 ::1}}} {
+    test "Load cluster announced client TLS port config on server start" {
+        R 0 config set cluster-announce-client-tls-port 6380
+        restart_server 0 true false
+    }
```
zuiderkwast: See the two comments above.

zuiderkwast (commented):
It's a very well-written PR. Thank you. I have only a few comments.

src/cluster_legacy.c
marvin-roesch: Good point, thanks! I've applied all your suggestions in a single commit, as GitHub wouldn't let me add them to a batch for some reason.

zuiderkwast: Great. In the diff view, you can add them to a batch, but not in the "Conversation" view.

tests/unit/cluster/announce-client-ports.tcl
marvin-roesch: Admittedly I was copying this one a little blindly from the "announce-client-ip" tests. I'll see about making this actually check for things.

I'll do the same on the announced IP one.

tests/unit/cluster/announce-client-ports.tcl
marvin-roesch: True, I've removed this. Also merged the two tests for config loading to run on the same instance as we just check config options.

tests/unit/cluster/announce-client-ports.tcl
zuiderkwast: https://github.com/valkey-io/valkey/pull/736? I did that one. I can't see a similar test in that PR. I'm missing something?

zuiderkwast: Ah, these tests were added in #963.

zuiderkwast: I think the idea is to test that these can be stored in nodes.conf as aux fields. The fix in #963 was to make sure it's possible to store the dots and colons of IP addresses in the aux fields in nodes.conf. Without that fix, I guess `restart_server` would fail.

I'm not exactly sure what triggers the write of nodes.conf. A more explicit way would be to check nodes.conf contents on disk. Not sure if it's worth it though.

The ports don't contain any special characters. It'd be fine skipping these tests for the ports.

marvin-roesch: Ahhh, I'll revert my changes to the "announce-client-ip" tests then and drop these for the ports.

marvin-roesch:
I'm not super familiar with Tcl, but from what I could find, the macOS tests might fail due to an old-ish version being installed where `lset` will not expand a list for the set index to fit. Since the tests don't rely on the indices of the list items, I went for `lappend` instead.

zuiderkwast:
> I'm not super familiar with Tcl, but from what I could find, the macOS tests might fail due to an old-ish version being installed where `lset` will not expand a list for the set index to fit. Since the tests don't rely on the indices of the list items, I went for `lappend` instead.

Sounds reasonable! None of us are TCL experts, but I know the macOS worker runs 8.5 while other workers run 8.6.

src/server.h
```diff
@@ -2142,6 +2142,8 @@ struct valkeyServer {
     int cluster_announce_port;                             /* base port to announce on cluster bus. */
     int cluster_announce_tls_port;                         /* TLS port to announce on cluster bus. */
     int cluster_announce_bus_port;                         /* bus port to announce on cluster bus. */
+    int cluster_announce_client_port;                      /* base port for clients to announce on cluster bus. */
```
```suggestion
    int cluster_announce_client_port;                      /* TCP port for clients to announce on cluster bus. */
```
Not sure what base means here, but you don't use it for the TLS port.

valkey.conf
```diff
@@ -2027,6 +2029,9 @@ aof-timestamp-enabled no
 # If cluster-announce-client-ipv4 and cluster-announce-client-ipv6 are omitted,
 # then cluster-announce-ip is exposed to clients.
 #
+# If cluster-announce-client-port or cluster-announce-client-tls-port are omitted,
+# then cluster-announce-port/tls-port is exposed to clients.
```
```suggestion
# If the port that clients will use to connect to Valkey is different than
# the one other valkey nodes in the cluster will connect to it on, either
# through special networking rules or because Valkey is behind a load balancer,
# you can configure the port that clients will see by setting 
# cluster-announce-port or cluster-announce-tls-port.
```
I see that you are copying the line above it, but I don't think it gives enough context to explain why this config exists or what it does.

tests/support/cluster_util.tcl
```diff
@@ -403,6 +403,27 @@ proc are_cluster_announced_ips_propagated {match_string {clients {}}} {
     return 1
 }
 
+# Check if cluster's announced ports are consistent and come from a predefined list
+# Optionally, a list of clients can be supplied.
```
madolson: You never use this optional functionality. Is it strictly needed? This adds more extra complexity instead of just passing it through.

madolson (commented):
Didn't do a deep review, just wanted to check some things we have missed in the past, but it all looked pretty good to me.

valkey.conf
zuiderkwast: Isn't this the background for all these announce configs? IPs and ports.

I like the explanation but maybe put it above the configs as an introduction?

zuiderkwast: We could also revise the heading "CLUSTER DOCKER/NAT support". Docker isn't the only use case, as it may have been a decade ago.

src/cluster_legacy.c
```diff
@@ -1000,7 +1061,7 @@ int clusterLockConfig(char *filename) {
 }
 
 /* Derives our ports to be announced in the cluster bus. */
-void deriveAnnouncedPorts(int *announced_tcp_port, int *announced_tls_port, int *announced_cport) {
+void deriveAnnouncedPorts(int *announced_tcp_port, int *announced_tls_port, int *announced_cport, int *announced_client_tcp_port, int *announced_client_tls_port) {
```
zuiderkwast: Let's linebreak it. It's very long.

```suggestion
void deriveAnnouncedPorts(int *announced_tcp_port, int *announced_tls_port, int *announced_cport,
                          int *announced_client_tcp_port, int *announced_client_tls_port) {
```

src/cluster_legacy.c
```diff
@@ -1038,7 +1102,7 @@ void clusterUpdateMyselfFlags(void) {
  * The option can be set at runtime via CONFIG SET. */
 void clusterUpdateMyselfAnnouncedPorts(void) {
     if (!myself) return;
-    deriveAnnouncedPorts(&myself->tcp_port, &myself->tls_port, &myself->cport);
+    deriveAnnouncedPorts(&myself->tcp_port, &myself->tls_port, &myself->cport, &myself->announce_client_tcp_port, &myself->announce_client_tls_port);
```
```suggestion
    deriveAnnouncedPorts(&myself->tcp_port, &myself->tls_port, &myself->cport,
                         &myself->announce_client_tcp_port, &myself->announce_client_tls_port);
```

src/cluster_legacy.c
```diff
@@ -1251,7 +1333,7 @@ void clusterInit(void) {
 
     /* Set myself->port/cport/pport to my listening ports, we'll just need to
      * discover the IP address via MEET messages. */
-    deriveAnnouncedPorts(&myself->tcp_port, &myself->tls_port, &myself->cport);
+    deriveAnnouncedPorts(&myself->tcp_port, &myself->tls_port, &myself->cport, &myself->announce_client_tcp_port, &myself->announce_client_tls_port);
```
```suggestion
    deriveAnnouncedPorts(&myself->tcp_port, &myself->tls_port, &myself->cport,
                         &myself->announce_client_tcp_port, &myself->announce_client_tls_port);
```

tests/support/cluster_util.tcl
zuiderkwast: It appears to be copied from `are_cluster_announced_ips_propagated` above.

Very little differs between `are_cluster_announced_ips_propagated` and `are_cluster_announced_ports_propagated`. Is it possible to combine them into one function and supply some argument to say what is to be compared (ip or port)?

valkey.conf
madolson: > Isn't this the background for all these announce configs? IPs and ports.

Kind of? This specific config also includes load balancers and weird networking, which isn't necessarily the same as the normal announce port which is really for docker/nat. We could maybe have a more holistic explanation instead of just a dump of fields.

tests/support/cluster_util.tcl
madolson: I'm also okay with that suggestion. I saw a lot of duplication in the tests, but didn't feel too strongly about merging them.

valkey.conf
zuiderkwast: At least kind of same as cluster-announce-client-ipv4, which is also "weird networking"?

src/server.h
marvin-roesch: I mirrored the description for the `cluster_announce_port` value. Should I adjust that comment as well to say `TCP port`?

zuiderkwast: No, cluster_announce_port is actually either TCP or TLS depending on various logic, for historical reasons

tests/support/cluster_util.tcl
marvin-roesch: I'll replace the two functions with a single one that can check either announced IPs or ports.

I was also thinking about reducing duplication between the tests in general, but don't feel comfortable enough with Tcl to get the degree of abstraction right.

zuiderkwast: Yeah, I know what you mean. If you give up, that's fine. But you seem to be doing great.

valkey.conf
```diff
@@ -2027,6 +2029,12 @@ aof-timestamp-enabled no
 # If cluster-announce-client-ipv4 and cluster-announce-client-ipv6 are omitted,
 # then cluster-announce-ip is exposed to clients.
 #
+# If the port that clients will use to connect to Valkey is different than
+# the one other valkey nodes in the cluster will connect to it on, either
+# through special networking rules or because Valkey is behind a load balancer,
+# you can configure the port that clients will see by setting 
+# cluster-announce-port or cluster-announce-tls-port.
```
zuiderkwast: This text is about the new configs, right?

```suggestion
# cluster-announce-client-port or cluster-announce-client-tls-port.
```

src/cluster_legacy.c
```diff
@@ -143,8 +150,20 @@ static inline int getNodeDefaultReplicationPort(clusterNode *n) {
...
+    if (c != NULL) {
+        if (n->announce_client_tcp_port) {
+            port = n->announce_client_tcp_port;
+        }
+        if (n->announce_client_tls_port) {
+            tlsport = n->announce_client_tls_port;
+        }
+    }
+
+    return use_tls ? tlsport : port;
```
zuiderkwast: Nit. Here,_ `port` and `tlsport` are both assigned but only one is returned in the end. It looks suboptimal, but maybe it's optimized by the compiler.

Just thinking, can we refactor it in some way to make it more strait-forward. Is this equivalent?

```suggestion
int clusterNodeClientPort(clusterNode *n, int use_tls, client *c) {
    if (use_tls && c != NULL && n->announce_client_tls_port) {
        return n->announce_client_tls_port;
    } else if (use_tls) {
        return n->tls_port;
    } else if (c != NULL && n->announce_client_tcp_port) {
        return n->announce_client_tcp_port;
    } else {
        return n->tcp_port;
    }
```

zuiderkwast (commented):
The Tcl refactoring looks great.

src/cluster_legacy.c
```diff
@@ -1000,7 +1061,8 @@ int clusterLockConfig(char *filename) {
 }
 
 /* Derives our ports to be announced in the cluster bus. */
-void deriveAnnouncedPorts(int *announced_tcp_port, int *announced_tls_port, int *announced_cport) {
+void deriveAnnouncedPorts(int *announced_tcp_port, int *announced_tls_port, int *announced_cport,
+                          int *announced_client_tcp_port, int *announced_client_tls_port) {
```
zuiderkwast: Clang-format didn't like this linebreak. :man_facepalming: Sorry for suggesting it.

So, I guess we have two options. Either all on one line or one parameter per line. Feel free to chose either way to silence clang-format.

valkey.conf
marvin-roesch: Ah, yep. Missed that while reading the suggestion.

src/cluster_legacy.c
marvin-roesch: No problemo, should've run clang-format again locally. I think one parameter per line is preferable. clang-format appears to be fine with the call sites, so I'll leave them with your suggestion.

zuiderkwast (approved):

rjd15372:
@marvin-roesch I'm about to merge this PR but if you intend to still apply the suggestion in https://github.com/valkey-io/valkey/pull/2429/files/0c0d599099e0f69e3f060cb8c6bb5a2d5b34dd20#diff-2515500619600c5a1e7a8d9aaa8761a6071314cd226c3b7ee4d0e054691f0a8e please let me know soon.

src/cluster_legacy.c
```diff
@@ -1000,7 +1061,12 @@ int clusterLockConfig(char *filename) {
 }
 
 /* Derives our ports to be announced in the cluster bus. */
-void deriveAnnouncedPorts(int *announced_tcp_port, int *announced_tls_port, int *announced_cport) {
+void deriveAnnouncedPorts(
+    int *announced_tcp_port,
+    int *announced_tls_port,
+    int *announced_cport,
+    int *announced_client_tcp_port,
+    int *announced_client_tls_port) {
```
zuiderkwast: Clang-format accepts this too, I think:

```suggestion
void deriveAnnouncedPorts(int *announced_tcp_port,
                          int *announced_tls_port,
                          int *announced_cport,
                          int *announced_client_tcp_port,
                          int *announced_client_tls_port) {
```

I prefer it...

marvin-roesch:
Thanks for merging! Sorry about not commenting on that suggestion before. My intention was mostly to avoid repeating conditions, but comparing generated assembly, the compiler at least produced *shorter* output for the suggested (and now applied) changes.

zuiderkwast:
> My intention was mostly to avoid repeating conditions, but comparing generated assembly, the compiler at least produced _shorter_ output for the suggested (and now applied) changes.

OK, good to know. I didn't check. :) This is not a very hot code path, so we can prefer readability over micro optimizations here.

PR #2430, opened 2025-08-05
Fix duplicate Acks for RDMA events and fix extremely large max latency for RDMA benchmark.

ruihong123:
(1) The old logic may result in the RDMA event being acknowledged unexpectly in the following two scenarios. 
* ibv_get_cq_event get an EAGAIN error. 
* ibv_get_cq_event get one event but it may ack multiple times in the pollcq loop.

(2) In the benchmark result of valkey over RDMA, the tail latency is as high as 177 milliseconds(almost 80x of TCP). This results from incorrect benchmark client setup which includes the connection setup time into the benchmark latency recording. This patch fixes this crazy tail latency issue by modifying the valkey-benchmark.c. This change only affects benchmark over RDMA as updates are regulated under Macro USE_RDMA.

There are following updates on valkey RDMA but I am willing to create separated pull requests.

### DIFF
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -1047,9 +1047,7 @@ static client createClient(char *cmd, int len, int seqlen, client from, int thre
         el = thread->el;
     }
     if (config.idlemode == 0) {
-        if (config.ct == VALKEY_CONN_RDMA) {
-            writeHandler(el, c->context->fd, c, 0);
-        } else {
+        if (config.ct != VALKEY_CONN_RDMA) {
             aeCreateFileEvent(el, c->context->fd, AE_WRITABLE, writeHandler, c);
         }
     } else
@@ -1193,6 +1191,20 @@ static void startBenchmarkThreads(void) {
     for (i = 0; i < config.num_threads; i++) pthread_join(config.threads[i]->thread, NULL);
 }
 
+#ifdef USE_RDMA
+static void issueFirstRequestForClients(aeEventLoop *el, int this_thread, int nt) {
+    listNode *ln = config.clients->head;
+    int count = 0;
+    while (ln) {
+        if (count++ % nt == this_thread) {
+            client c = ln->value;
+            writeHandler(el, c->context->fd, c, 0);
+        }
+        ln = ln->next;
+    }
+}
+#endif
+
 /* Benchmark a sequence of commands. The cmd is RESP encoded of length len and
  * seqlen is the number of commands included in cmd. */
 static void benchmarkSequence(const char *title, char *cmd, int len, int seqlen) {
@@ -1226,9 +1238,14 @@ static void benchmarkSequence(const char *title, char *cmd, int len, int seqlen)
     createMissingClients(c);
 
     config.start = mstime();
-    if (!config.num_threads)
+    if (!config.num_threads) {
+#ifdef USE_RDMA
+        if (config.idlemode == 0 && config.ct == VALKEY_CONN_RDMA) {
+            issueFirstRequestForClients(config.el, 0, 1);
+        }
+#endif
         aeMain(config.el);
-    else
+    } else
         startBenchmarkThreads();
     config.totlatency = mstime() - config.start;
 
@@ -1274,6 +1291,11 @@ static void freeBenchmarkThreads(void) {
 
 static void *execBenchmarkThread(void *ptr) {
     benchmarkThread *thread = (benchmarkThread *)ptr;
+#ifdef USE_RDMA
+    if (config.idlemode == 0 && config.ct == VALKEY_CONN_RDMA) {
+        issueFirstRequestForClients(thread->el, thread->index, config.num_threads);
+    }
+#endif
     aeMain(thread->el);
     return NULL;
 }
--- a/src/rdma.c
+++ b/src/rdma.c
@@ -603,7 +603,12 @@ static int connRdmaHandleCq(rdma_connection *rdma_conn) {
             serverLog(LL_WARNING, "RDMA: get CQ event error");
             return C_ERR;
         }
-    } else if (ibv_req_notify_cq(ev_cq, 0)) {
+
+        return C_OK;
+    }
+
+    ibv_ack_cq_events(ctx->cq, 1);
+    if (ibv_req_notify_cq(ev_cq, 0)) {
         serverLog(LL_WARNING, "RDMA: notify CQ error");
         return C_ERR;
     }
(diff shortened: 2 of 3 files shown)


deps/libvalkey/src/rdma.c
```diff
@@ -404,7 +409,10 @@ static int connRdmaHandleCq(valkeyContext *c) {
         return VALKEY_OK;
     }
 
-    ibv_ack_cq_events(ctx->cq, 1);
+    if (need_ack) {
+        ibv_ack_cq_events(ctx->cq, 1);
+        need_ack = false;
```
ranshid: nit: can we document that we could loop back here, which is why we "reset" the need_ack? 
I found it confusing for a sec there...

src/valkey-benchmark.c
```diff
@@ -1226,8 +1238,14 @@ static void benchmarkSequence(const char *title, char *cmd, int len, int seqlen)
     createMissingClients(c);
 
     config.start = mstime();
-    if (!config.num_threads)
+    if (!config.num_threads) {
+#ifdef USE_RDMA
+        if (config.idlemode == 0) {
+            issueFirstRequestForClients(config.el, 0, 1);
+        }
+#endif
```
ranshid: why don't we do it once as part of the aeMain?

ranshid (commented):
some minor comments overall LGTM

deps/libvalkey/src/rdma.c
ruihong123: Yes, I updated a comment in the latest git commit --amend.

src/valkey-benchmark.c
ruihong123: This's a good question. As RDMA do not have ready-to-write signal, valkey benchmark over RDMA has to trigger the first request manually. In the code before this patch, the first request is triggered in createClient (line 1049-1055 in valkey-benchmark.c), which makes the latency of the first request including all the following connection setup. I just moved the first manual trigger from createClient to a place right before the start of eventloop. 

Note: This patch only takes effect when compiling the code base with BUILD_RDMA=yes and set --rdma for valkey benchmark running.

ranshid: I understand. But aeMain is also called once per each thread. I think instead of placing this code in 2 places. Once inside aeMain should probably do the same?

ruihong123: If I understand correctly, you are saying that I can write issueFirstRequestForClients inside the function of void aeMain(aeEventLoop *eventLoop). Please correct me if I understand it wrong. If this is the case then I did not do it because of three reasons:  

(1) There are some arguments which may only be accessed by valkey-benchmark.c. If I put this logic inside aeMain, the argument list of aeMain will be changed. 
(2) valkey server also uses this aeMain and it does not need this logic. 
(3) I would like to minimize the files that this commit touches, as RDMA is an experimental feature.

ranshid: ah. you are right. I meant that we could potentially wrap the aeMain call in a funciton which will be called from the different places, but that is not critical. LGTM

ranshid (approved):

pizhenwei:
Hi @ruihong123 
I think the changes in `src/valkey-benchmark.c` seems correct, but the other changes do not make sense to me. You may need to read some codes from `rdma-core/libibverbs/verbs.c` to understand the full logic.

jakemoroni:
Hi @pizhenwei 

My understanding is that the `ibv_ack_cq_events` function should only be called if an event is actually received from `ibv_get_cq_event`.

The original code had a race where `ibv_get_cq_event` may return -1 and `errno == EAGAIN` (because the FD is nonblocking), but a completion could arrive just immediately before the `ibv_poll_cq` call, which would then lead to calling `ibv_ack_cq_events` even though no event was received.

A similar scenario could occur if a new completion arrives just after the one that was polled from `ibv_poll_cq` (but before exiting the loop).

Looking at the change, it may be possible to simplify it further by just doing the `ibv_ack_cq_events` in the same `else` block where `ibv_req_notify_cq` is called and eliminating the `need_ack` variable. Either way should be functionally correct though.

pizhenwei:
> Hi @pizhenwei
> 
> My understanding is that the `ibv_ack_cq_events` function should only be called if an event is actually received from `ibv_get_cq_event`.
> 
> The original code had a race where `ibv_get_cq_event` may return -1 and `errno == EAGAIN` (because the FD is nonblocking), but a completion could arrive just immediately before the `ibv_poll_cq` call, which would then lead to calling `ibv_ack_cq_events` even though no event was received.
> 
> A similar scenario could occur if a new completion arrives just after the one that was polled from `ibv_poll_cq` (but before exiting the loop).
> 
> Looking at the change, it may be possible to simplify it further by just doing the `ibv_ack_cq_events` in the same `else` block where `ibv_req_notify_cq` is called and eliminating the `need_ack` variable. Either way should be functionally correct though.

Thanks, you are right. There is no need to use additional `need_ack`.
Hi @ruihong123  I prefer the following style rather than the `need_ack` style:
```
    if (ibv_get_cq_event(ctx->comp_channel, &ev_cq, &ev_ctx) < 0) {
        if (errno != EAGAIN) {
            serverLog(LL_WARNING, "RDMA: get CQ event error");
            return C_ERR;
        }

        return C_OK;
    }

    ibv_ack_cq_events(ctx->cq, 1);
    if (ibv_req_notify_cq(ev_cq, 0)) {
        serverLog(LL_WARNING, "RDMA: notify CQ error");
        return C_ERR;
    }
```

By the way, you should keep valkey/deps/libvalkey, and modify the client side [libvalkey](https://github.com/valkey-io/libvalkey), maintainers will bump these changes into valkey later.

ruihong123:
Hi @pizhenwei , 

I have updated the commits according to your suggestions. I will also submit a commit on the ack on libvalkey right away.  

By the way, there will be following pull requests on valkey over RDMA from our side and they may require your reviews as well.

Thanks,

Ruihong

pizhenwei:
> Hi @pizhenwei ,
> 
> I have updated the commits according to your suggestions. I will also submit a commit on the ack on libvalkey right away.
> 
> By the way, there will be following pull requests on valkey over RDMA from our side and they may require your reviews as well.
> 
> Thanks,
> 
> Ruihong

Sure, please AT me in that PR, also add this page link into that one.

Sorry, I forgot `valkey/tests/rdma/rdma-test.c`, please also fix this file in this PR together.

This version LGTM, thanks.

ruihong123:
> > Hi @pizhenwei ,
> > I have updated the commits according to your suggestions. I will also submit a commit on the ack on libvalkey right away.
> > By the way, there will be following pull requests on valkey over RDMA from our side and they may require your reviews as well.
> > Thanks,
> > Ruihong
> 
> Sure, please AT me in that PR, also add this page link into that one.
> 
> Sorry, I forgot `valkey/tests/rdma/rdma-test.c`, please also fix this file in this PR together.
> 
> This version LGTM, thanks.

Hi, I amended the modification of rdma-test.c in last commit "Fix wrong max latency for benchmark over RDMA". Now this PR should be ready to go.

pizhenwei (approved):
LGTM

PR #2432, opened 2025-08-05
Use bool for return types in hashtable functions

### LINKED ISSUE #2352: Use bool for return types in hashtable functions
Many hashtable functions return 1 or 0 as an int. This was done before we started allowing bool. We should probably change them all to bool. (A separate PR at some point.)

_Originally posted by @zuiderkwast in https://github.com/valkey-io/valkey/pull/2089#discussion_r2203523634_

cxljs:
closes #2352


ranshid (approved):
Nice! I looked at the changes and they all seem correct to me.

src/defrag.c
```diff
@@ -691,15 +691,13 @@ static void defragKey(defragKeysCtx *ctx, robj **elemref) {
             /* Replace the pointer in the expire table without accessing the old
              * pointer. */
             hashtable *expires_ht = kvstoreGetHashtable(db->expires, slot);
-            int replaced = hashtableReplaceReallocatedEntry(expires_ht, ob, newob);
-            serverAssert(replaced);
+            serverAssert(hashtableReplaceReallocatedEntry(expires_ht, ob, newob));
```
zuiderkwast: Nit.

I wanted to avoid doing stuff with side-effects inside asserts, because if the code is ever compiled without asserts, this action will not happen. Now, we don't do that with serverAssert but we do for debugServerAssert and the generic assert in C. It's more of a good practice to not do important stuff within asserts.

ranshid: @zuiderkwast big +1, but I suspect as part of such effort we will have to go through ALL the assert cases anyway.
regardless. lets indeed change it to match as you suggest (it will reduce the code change anyhow)

cxljs: Thank you, your suggestion is excellent. I hadn't considered that before.

madolson: > debugServerAssert and the generic assert in C

Just to clarify, we don't use generic C asserts anywhere in Valkey today. All asserts should use serverAssert.

zuiderkwast: True, thanks for the clarification. Still it's good to follow the same practice and not rely on that code in asserts will run.

madolson: Apparently, rdma.c does though https://github.com/valkey-io/valkey/blob/2f1b25067e8428955c938ea6c1c34242d33d9de0/src/rdma.c#L25. Probably just a miss because that's not obvious.

cxljs: Thank you, I'll submit a PR to remove the generic C assert. 

And I have a question: Valkey defines both serverAssert and assert. Do we want to unify them into one, or keep the current setup?
@madolson @zuiderkwast @ranshid  

https://github.com/valkey-io/valkey/blob/2f1b25067e8428955c938ea6c1c34242d33d9de0/src/server.h#L687
https://github.com/valkey-io/valkey/blob/2f1b25067e8428955c938ea6c1c34242d33d9de0/src/serverassert.h#L60

ranshid: SO the current rule is that you include server.h in order to get access to the general Valkey API including serverAssert.
However in some cases we would like to write or include a generic module or data structure and we do not want to use "Valkey" specific stuff or because the code is a fork or a different OSS project. For the second case we can include "serverassert.h" and use the "generic assert. I think we should just be careful about including <assert.h>

ranshid: specifically in rdma.c I am not sure what is the intention in including server.h, I think it was for serverLog. (NOTE: when including some headers for a specific purpose I think we should comment aside why we included it)
In case server.h was included you should use serverAssert.

cxljs: Thank you very much @ranshid , I understand now.

PR #2434, opened 2025-08-06
Skip failure reports for already failed nodes

sungming2:
This change avoids additional failure report creation if the node is already marked as failed. The failure report count has never been used after a node has been marked as failed. So, there is no value addition in maintaining it further. This reduces operation of both add and delete failure report. Hence, the performance benefit.

We can observe an avg. of 10% reduce in p99 CPU utilization (in a 2000 nodes cluster (1000 primary/ 1000 replica) with 330 nodes in failed state with this change.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -1785,7 +1785,7 @@ static void decodeFailureReportKey(unsigned char *buf, mstime_t *report_time, cl
  * 'sender' node.
  *
  * Returns 0:
- *   - Early exit when the node is already in FAIL state
+ *   - The node is already in FAIL state
  *   - The same sender has already reported within the same time bucket
  *   - An existing report from 'sender' was refreshed (timestamp updated)
  * Returns 1 if a brand new failure report entry is created. */
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -1784,14 +1784,20 @@ static void decodeFailureReportKey(unsigned char *buf, mstime_t *report_time, cl
  * 'failing' is the node that is in failure state according to the
  * 'sender' node.
  *
- * The function returns 0 if it early‐exits (same sender & time bucket)
- * or updates a timestamp of an existing failure report from the same sender.
- * 1 is returned if a new failure report is created. */
+ * Returns 0:
+ *   - Early exit when the node is already in FAIL state
+ *   - The same sender has already reported within the same time bucket
+ *   - An existing report from 'sender' was refreshed (timestamp updated)
+ * Returns 1 if a brand new failure report entry is created. */
 int clusterNodeAddFailureReport(clusterNode *failing, clusterNode *sender) {
     unsigned char buf[FAILURE_REPORT_KEYLEN];
     mstime_t now = mstime();
     const mstime_t bucketed_time = (now / SEC_IN_MS) * SEC_IN_MS + SEC_IN_MS;
     int is_new = 1;
+
+    /* This avoids unnecessary iteration and memory ops, improving performance
+     * when handling repeated reports for already failed nodes. */
+    if (nodeFailed(failing)) return 0;
 
     /* Look for any existing entry from this sender and remove it */
     raxIterator ri;
--- a/src/commands/cluster-count-failure-reports.json
+++ b/src/commands/cluster-count-failure-reports.json
@@ -7,6 +7,12 @@
         "arity": 3,
         "container": "CLUSTER",
         "function": "clusterCommand",
+        "history": [
+            [
+                "9.0.0",
+                "Avoid accepting failure reports for nodes already marked as failed."
+            ]
+        ],
         "command_flags": [
             "ADMIN",
             "STALE"
--- a/src/commands/cluster-count-failure-reports.json
+++ b/src/commands/cluster-count-failure-reports.json
@@ -1,6 +1,6 @@
 {
     "COUNT-FAILURE-REPORTS": {
-        "summary": "Returns the number of active failure reports active for a node.",
+        "summary": "Returns the number of active failure reports for a node. No new reports are created once the node is marked as failed.",
         "complexity": "O(N) where N is the number of failure reports",
         "group": "cluster",
         "since": "3.0.0",
@@ -10,7 +10,7 @@
         "history": [
             [
                 "9.0.0",
-                "Avoid accepting failure reports for nodes already marked as failed."
+                "Ignore additional failure reports for a node which has been marked as failed."
             ]
         ],
         "command_flags": [
--- a/tests/unit/cluster/failure-marking.tcl
+++ b/tests/unit/cluster/failure-marking.tcl
@@ -131,10 +131,10 @@ start_cluster 3 1 {tags {external:skip cluster}} {
         set replica0_pid [srv -3 pid]
         set replica0_id [dict get [cluster_get_myself 3] id]
 
-        assert {[lindex [$primary0 role] 0] eq {master}}
-        assert {[lindex [$primary1 role] 0] eq {master}}
-        assert {[lindex [$primary2 role] 0] eq {master}}
-        assert {[lindex [$replica0 role] 0] eq {slave}}
+        assert_equal [lindex [$primary0 role] 0] {master}
+        assert_equal [lindex [$primary1 role] 0] {master}
+        assert_equal [lindex [$primary2 role] 0] {master}
+        assert_equal [lindex [$replica0 role] 0] {slave}
 
         # Ensure replica is synced before simulating failure
         wait_for_sync $replica0
--- a/tests/unit/cluster/failure-marking.tcl
+++ b/tests/unit/cluster/failure-marking.tcl
@@ -159,13 +159,12 @@ start_cluster 3 1 {tags {external:skip cluster}} {
         # Other nodes should not add a new failure report from this primary1
         resume_process $primary1_pid
 
-        # Ensure no primary has more than one reports
+        # Ensure that primary0 and primary2 do not have more than one report
         wait_for_condition 1000 50 {
             [R 0 CLUSTER COUNT-FAILURE-REPORTS $replica0_id] < 2 &&
-            [R 1 CLUSTER COUNT-FAILURE-REPORTS $replica0_id] < 2 &&
             [R 2 CLUSTER COUNT-FAILURE-REPORTS $replica0_id] < 2
         } else {
-            fail "No primary should exceed two failure reports"
+            fail "Ensure primary0 and primary2 do not exceed one failure report"
         }
 
         # Bring the replica back online and verify cleanup
--- a/tests/unit/cluster/failure-marking.tcl
+++ b/tests/unit/cluster/failure-marking.tcl
@@ -58,69 +58,56 @@ start_cluster 2 1 {tags {external:skip cluster}} {
 
 tags {external:skip tls:skip cluster singledb} {
     set base_conf [list cluster-enabled yes cluster-ping-interval 100 cluster-node-timeout 3000 save ""]
-    start_multiple_servers 7 [list overrides $base_conf] {
+    start_multiple_servers 5 [list overrides $base_conf] {
         test "Only primary with slots has the right to mark a node as failed" {
             set primary_host [srv 0 host]
             set primary_port [srv 0 port]
             set primary_pid [srv 0 pid]
             set primary_id [R 0 CLUSTER MYID]
-
-            set primary2_host [srv -1 host]
-            set primary2_port [srv -1 port]
-            set primary2_pid [srv -1 pid]
-            set primary2_id [R 1 CLUSTER MYID]
-
-            set replica_id [R 2 CLUSTER MYID]
-            set replica_pid [srv -2 pid]
+            set replica_id [R 1 CLUSTER MYID]
+            set replica_pid [srv -1 pid]
 
             # Meet others nodes.
             R 1 CLUSTER MEET $primary_host $primary_port
             R 2 CLUSTER MEET $primary_host $primary_port
             R 3 CLUSTER MEET $primary_host $primary_port
             R 4 CLUSTER MEET $primary_host $primary_port
-            R 5 CLUSTER MEET $primary_host $primary_port
-            R 6 CLUSTER MEET $primary_host $primary_port
 
             # Build a single primary cluster.
-            cluster_allocate_slots 2 2
+            cluster_allocate_slots 1 1
             wait_for_cluster_propagation
-            R 2 CLUSTER REPLICATE $primary_id
-            R 3 CLUSTER REPLICATE $primary2_id
+            R 1 CLUSTER REPLICATE $primary_id
             wait_for_cluster_propagation
             wait_for_cluster_state "ok"
 
             # Pause the primary, marking the primary as pfail.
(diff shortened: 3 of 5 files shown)


src/cluster_legacy.c
```diff
@@ -1720,15 +1720,21 @@ static void decodeFailureReportKey(unsigned char *buf, mstime_t *report_time, cl
  * 'failing' is the node that is in failure state according to the
  * 'sender' node.
  *
- * The function returns 0 if it early‐exits (same sender & time bucket)
- * or updates a timestamp of an existing failure report from the same sender.
- * 1 is returned if a new failure report is created. */
+ * Returns 0:
```
sarthakaggarwal97, line 1787: nit: not sure if we need to update the top comment as it doesn't add any new information.

src/cluster_legacy.c
```diff
@@ -1720,15 +1720,21 @@ static void decodeFailureReportKey(unsigned char *buf, mstime_t *report_time, cl
...
+ * Returns 1 if a brand new failure report entry is created. */
 int clusterNodeAddFailureReport(clusterNode *failing, clusterNode *sender) {
     unsigned char buf[FAILURE_REPORT_KEYLEN];
     mstime_t now = mstime();
     const mstime_t bucketed_time = (now / SEC_IN_MS) * SEC_IN_MS + SEC_IN_MS;
     int is_new = 1;
 
+    /* This avoids unnecessary iteration and memory ops, improving performance
+     * when handling repeated reports for already failed nodes. */
+    if (nodeFailed(failing)) return 0;
```
sarthakaggarwal97, line 1800: isn't it better to branch out from [clusterProcessGossipSection](https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L2560)? `markNodeAsFailingIfNeeded` also has a similar condition where we just return 0 if the node is failed.

tests/unit/cluster/failure-marking.tcl
```diff
@@ -58,54 +58,57 @@ start_cluster 2 1 {tags {external:skip cluster}} {
...
+            wait_node_marked_pfail 6 $primary_id
 
             # Pause the replica, marking the replica as pfail.
             pause_process $replica_pid
-            wait_node_marked_pfail 2 $replica_id
             wait_node_marked_pfail 3 $replica_id
             wait_node_marked_pfail 4 $replica_id
-
-            # Resume the primary, marking the replica as fail.
-            resume_process $primary_pid
```
sarthakaggarwal97: why did we remove this part of the test?

src/cluster_legacy.c
sungming2: I thought about that first but I think we need to keep this behavior regardless of where or how this function is called in the future.

tests/unit/cluster/failure-marking.tcl
sungming2: It doesn't change any result in this test, but let me add it back

zuiderkwast:
For reference, the list was replaced by rax in #2277.

hpatro:
> For reference, the list was replaced by rax in #2277.

Btw It's unrelated to this change. This is a minor optimization to avoid maintaining failure report for nodes which are already in failed state.

zuiderkwast:
> > For reference, the list was replaced by rax in #2277.
> 
> Btw It's unrelated to this change. This is a minor optimization to avoid maintaining failure report for nodes which are already in failed state.

Sure. I was just curious. Both PRs are optimizions in the same area of failures in huge clusters.

src/cluster_legacy.c
sarthakaggarwal97: As discussed offline, going into the if condition is a no-op after we put the check for failed node over here. I would prefer to split the branch to have better clarity on the operations we are performing for PFAIL and FAIL nodes.

tests/unit/cluster/failure-marking.tcl
```diff
@@ -124,3 +114,80 @@ tags {external:skip tls:skip cluster singledb} {
         }
     }
 }
+
+# Test that no new failure-report is added once the node is already marked as FAIL
+start_cluster 4 1 {tags {external:skip cluster}} {
```
hpatro: 4 primaries seems weird. Why not use 3 ?

src/cluster_legacy.c
hpatro: I'm fine with either approach. This method already does early exit and this seems another case.

tests/unit/cluster/human-announced-nodename.tcl
```diff
@@ -1,5 +1,5 @@
 # Check if cluster's view of human announced nodename is reported in logs
-start_cluster 3 0 {tags {external:skip cluster}} {
+start_cluster 4 0 {tags {external:skip cluster}} {
```
hpatro, line 2: Why do we need to update this test ? There should be no behavior change, right ?

tests/unit/cluster/failure-marking.tcl
```diff
@@ -100,16 +100,6 @@ tags {external:skip tls:skip cluster singledb} {
...
-            # Check if we got the right failure reports.
-            wait_for_condition 1000 50 {
-                [R 0 CLUSTER COUNT-FAILURE-REPORTS $replica_id] == 0 &&
-                [R 2 CLUSTER COUNT-FAILURE-REPORTS $replica_id] == 1 &&
-                [R 3 CLUSTER COUNT-FAILURE-REPORTS $replica_id] == 1 &&
-                [R 4 CLUSTER COUNT-FAILURE-REPORTS $replica_id] == 1
-            } else {
-                fail "Cluster COUNT-FAILURE-REPORTS is not right."
-            }
-
```
hpatro, line 112: Why this needs to be removed ?

tests/unit/cluster/failure-marking.tcl
```diff
@@ -124,3 +114,80 @@ tags {external:skip tls:skip cluster singledb} {
         }
     }
 }
+
+# Test that no new failure-report is added once the node is already marked as FAIL
```
hpatro, line 118: I don't agree with this test comment. This test doesn't verify no new failure report being added. We just verify we have unique failure report for each node and clean it up once the node is reachable. 

Some other tests already cover this, do we need a new test for this?

hpatro (commented):
The code change looks good to me. Unsure about the changes in tests.

tests/unit/cluster/failure-marking.tcl
sungming2: When a primary with slots comes back up, it can immediately mark its replica as failed once the timeout expires. Since it is the only node voting for that primary, it sends out a fail broadcast. Other nodes receive that broadcast and also mark the replica as failed, which blocks additional failure reports.
I wanted to remove this behavior because it no longer guarantees a consistent number of failure reports across nodes. It now depends on the timing of when each node receives the fail broadcast.

tests/unit/cluster/human-announced-nodename.tcl
sungming2: We only return 1 if there is a new report, so in order to get failure reports, the node shouldn't be marked as failed. Do you think  we should keep this beahvior?

```
 * Returns 0:
 *   - Early exit when the node is already in FAIL state
 *   - The same sender has already reported within the same time bucket
 *   - An existing report from 'sender' was refreshed (timestamp updated)
 * Returns 1 if a brand new failure report entry is created. */
 * int clusterNodeAddFailureReport(clusterNode *failing, clusterNode *sender) {
```

https://github.com/valkey-io/valkey/blob/2ca0dd878196033811a69a9a2a3381030e5264dd/src/cluster_legacy.c#L2565-L2569

tests/unit/cluster/failure-marking.tcl
```diff
@@ -124,3 +114,68 @@ tags {external:skip tls:skip cluster singledb} {
...
+        resume_process $primary1_pid
+
+        # Ensure no primary has more than one reports
+        wait_for_condition 1000 50 {
+            [R 0 CLUSTER COUNT-FAILURE-REPORTS $replica0_id] < 2 &&
+            [R 1 CLUSTER COUNT-FAILURE-REPORTS $replica0_id] < 2 &&
+            [R 2 CLUSTER COUNT-FAILURE-REPORTS $replica0_id] < 2
+        } else {
+            fail "No primary should exceed two failure reports"
+        }
```
hpatro: This could be zero as well , right? We don't verify here if the primary has more than 1 report.

sungming2: Could you clarify a bit more? Are you asking whether one, some, or all of them could be zero?
If I understand the question correctly, my answer is they can be either 0 or 1, since we don’t know which one will reach quorum first, so we can say “one of them got quorum” with this very small cluster. But in a larger cluster, multiple nodes could achieve quorum at the same time.

e.g.,
P0 = 1, P1 = 1, P2 = 0
P0 = 0, P1 = 1, P2 = 1

hpatro: Since all the primaries are back up now. Wouldn't one node reach to 2 failure report count ? I'm worried this will turn into a flaky failure.

tests/unit/cluster/human-announced-nodename.tcl
hpatro: This test is quite weirdly written. We are doing unit testing via integration tests.

tests/unit/cluster/failure-marking.tcl
sungming2: I commented this at line 158
```
        # Resume the final paused primary (primary1)
        # Other nodes should not add a new failure report from this primary1
        resume_process $primary1_pid
```

I should have removed primary1 check from this
```
        wait_for_condition 1000 50 {
            [R 0 CLUSTER COUNT-FAILURE-REPORTS $replica0_id] < 2 &&
            [R 1 CLUSTER COUNT-FAILURE-REPORTS $replica0_id] < 2 && <<< Remove
            [R 2 CLUSTER COUNT-FAILURE-REPORTS $replica0_id] < 2
        } else {
            fail "No primary should exceed two failure reports"
        }
```

so that it always ensures primary0 and primary2 have 0 or 1 report. Does it make sense to you

image

hpatro: But we don't clear the failure report immediately at this point while marking a node as failed. So, it can still show 2 failure reports on any of the node for `node timeout * 2` period.

hpatro: total failure report count is `other reported count + 1 (myself if primary)`. That's where the confusion was. Thanks for the clarification.

hpatro (approved):
LGTM.

This change avoids additional failure report creation if the node is already marked as failed. The failure report count has never been used after a node has been marked as failed. So, there is no value addition in maintaining it further. This reduces operation of both add and delete failure report. Hence, the performance benefit.

sarthakaggarwal97 (approved):
LGTM!

hpatro:
There is a slight behavior change `CLUSTER COUNT-FAILURE-REPORTS <node-id>` will start reporting `0` once the node is marked as `FAIL` and the old failure reports have expired. Earlier it used to show non negative failure report count (other primaries view) even after the node is marked as `FAIL`. 

@valkey-io/core-team I don't know if you folks want to vote on this but have a look.

tests/unit/cluster/failure-marking.tcl
```diff
@@ -124,3 +114,67 @@ tags {external:skip tls:skip cluster singledb} {
...
+
+        # Replica node
+        set replica0 [srv -3 client];
+        set replica0_pid [srv -3 pid]
+        set replica0_id [dict get [cluster_get_myself 3] id]
+
+        assert {[lindex [$primary0 role] 0] eq {master}}
+        assert {[lindex [$primary1 role] 0] eq {master}}
+        assert {[lindex [$primary2 role] 0] eq {master}}
+        assert {[lindex [$replica0 role] 0] eq {slave}}
```
madolson: You should basically never use assert, since it won't show the other state on failure.
```suggestion
        assert_equal [lindex [$primary0 role] 0] {master}
        assert_equal [lindex [$primary1 role] 0] {master}
        assert_equal [lindex [$primary2 role] 0] {master}
        assert_equal [lindex [$replica0 role] 0] {slave}
```

src/cluster_legacy.c
```diff
@@ -1720,15 +1720,21 @@ static void decodeFailureReportKey(unsigned char *buf, mstime_t *report_time, cl
  * 'failing' is the node that is in failure state according to the
  * 'sender' node.
  *
- * The function returns 0 if it early‐exits (same sender & time bucket)
- * or updates a timestamp of an existing failure report from the same sender.
- * 1 is returned if a new failure report is created. */
+ * Returns 0:
+ *   - Early exit when the node is already in FAIL state
```
```suggestion
 *   - The node is already in FAIL state
```

madolson:
> I don't know if you folks want to vote on this but have a look.

I think it's fine. We can update the documentation to make it clear that we might not report on errors if we believe a node is failed.

hpatro:
@sungming2 Could we add a history section to this https://valkey.io/commands/cluster-count-failure-reports/ ?

tests/unit/cluster/failure-marking.tcl
sungming2: I just copied that from another test, I can udpate it

https://github.com/valkey-io/valkey/blob/a97d584cc20013c6797843a30633f0f268874d28/tests/unit/cluster/failure-marking.tcl#L37-L39

src/commands/cluster-count-failure-reports.json
```diff
@@ -7,6 +7,12 @@
         "arity": 3,
         "container": "CLUSTER",
         "function": "clusterCommand",
+        "history": [
+            [
+                "9.0.0",
+                "Avoid accepting failure reports for nodes already marked as failed."
```
hpatro: WDYT ?
nit:
```suggestion
                "Stop maintaining failure report for a node which has been marked as failed."
```

sungming2: "Stop maintaining" sounds like permanent to me

sungming2: Maybe "ignore additional failure reports for a node which has been marked as failed" sounds better?

hpatro: It's permanent only, right ?

hpatro (approved):

src/commands/cluster-count-failure-reports.json
zuiderkwast: I may be wrong, but I believe in the past we only used "history" for syntax changes such as command parameters, not for behavior changes. I'm OK with it though.

Regarding the formulation "stop maintaining", "avoid accepting", etc. I think the formulation should be according to this command which is about counting failure reports, so I suggest we write something about counting, for example "Failure reports received when a node is already marked as failing are not counted".

sungming2: > I suggest we write something about counting, for example "Failure reports received when a node is already marked as failing are not counted".

It's mentioned in this description: https://github.com/valkey-io/valkey/pull/2434/commits/e9c22408cf26dd88f00a0f0d6d615c074d7a2717
and it will be on the doc as well

hpatro: Yes @zuiderkwast, we also had a similar discussion to add information about the command behavior and updated the description accordingly.

PR #2276, opened 2025-06-26
Log test details at the end when the test times out.

sarthakaggarwal97:
Resolves #2267

Verified that the name of the test now prints in the bottom during test time out.

```
[TIMEOUT]: clients state report follows.
sock15b01d810 => (IN PROGRESS) WAIT should not acknowledge 2 additional copies of the data
sock15b01dd10 => (IN PROGRESS) WAIT should not acknowledge 2 additional copies of the data
sock15b01d110 => (IN PROGRESS) WAIT should not acknowledge 2 additional copies of the data
sock15b01e490 => (IN PROGRESS) WAIT should not acknowledge 2 additional copies of the data
sock15b01e710 => (IN PROGRESS) WAIT should not acknowledge 2 additional copies of the data
Killing still running Valkey server 26294
Killing still running Valkey server 26299
Killing still running Valkey server 26302
Killing still running Valkey server 26304
Killing still running Valkey server 26306
Killing still running Valkey server 26342
Killing still running Valkey server 26349
Killing still running Valkey server 26357
Killing still running Valkey server 26358
Killing still running Valkey server 26359

                   The End

Execution time of different units:

!!! WARNING The following tests failed:

*** [TIMEOUT]: WAIT should not acknowledge 2 additional copies of the data in tests/unit/wait.tcl
*** [TIMEOUT]: WAIT should not acknowledge 2 additional copies of the data in tests/unit/wait.tcl
*** [TIMEOUT]: WAIT should not acknowledge 2 additional copies of the data in tests/unit/wait.tcl
*** [TIMEOUT]: WAIT should not acknowledge 2 additional copies of the data in tests/unit/wait.tcl
*** [TIMEOUT]: WAIT should not acknowledge 2 additional copies of the data in tests/unit/wait.tcl
Cleanup: may take some time... OK

------- 

[TIMEOUT]: clients state report follows.
sock146070790 => (IN PROGRESS) Check for memory leaks (pid 26494)
Killing still running Valkey server 26480
Killing still running Valkey server 26494

                   The End

Execution time of different units:

!!! WARNING The following tests failed:

*** [TIMEOUT]: Check for memory leaks in tests/unit/introspection.tcl
Cleanup: may take some time... OK
```

Issue #2267, opened 2025-06-24
[BUG] `The following tests failed:`/`clients state report follows.` precede no test list and no state report


jsoref:
**Describe the bug**

A short description of the bug.

**To reproduce**

- #2205

Visit https://github.com/valkey-io/valkey/actions/runs/15836683365/job/44641622150 (or many other runs ...) and see:

```
[TIMEOUT]: clients state report follows.
sock55d6f62ed480 => (IN PROGRESS) Active Defrag big list: cluster
Killing still running Valkey server 54846
```
...
```

!!! WARNING The following tests failed:

*** [TIMEOUT]: clients state report follows.
Cleanup: may take some time... OK
```

**Expected behavior**

1. A message that says a state report follows should *either* be followed by a state report or not be present.
2. A message that says "the following tests failed" should be followed by a list of tests

**Additional information**

This code generates the final output:
https://github.com/valkey-io/valkey/blob/2aea99b26f6520ab0ee5068a0a16e66c78c3fc47/tests/test_helper.tcl#L530-L532

This code is what contributes the incompatible text:
https://github.com/valkey-io/valkey/blob/2aea99b26f6520ab0ee5068a0a16e66c78c3fc47/tests/test_helper.tcl#L336-L338

jsoref:
There are other variations, e.g. https://github.com/valkey-io/valkey/actions/runs/15850765175/job/44683384808

```
!!! WARNING The following tests failed:

*** [err]: Active Defrag big zset: standalone in tests/unit/memefficiency.tcl
defrag didn't stop.
Cleanup: may take some time... OK
```

`defrag didn't stop.`
https://github.com/valkey-io/valkey/blob/2aea99b26f6520ab0ee5068a0a16e66c78c3fc47/tests/unit/memefficiency.tcl#L105-L108
and
`Cleanup: may take some time... OK`
https://github.com/valkey-io/valkey/blob/2aea99b26f6520ab0ee5068a0a16e66c78c3fc47/tests/test_helper.tcl#L281
are not tests that failed and thus don't belong in this summary.

Nor does the `*** [err]: ` qualifier. The test is _probably_ `Active Defrag big zset: standalone in tests/unit/memefficiency.tcl`
https://github.com/valkey-io/valkey/blob/2aea99b26f6520ab0ee5068a0a16e66c78c3fc47/tests/unit/memefficiency.tcl#L434-L435

PR #2437, opened 2025-08-06
Print error context when test fails

enjoy-binbin:
We used to did print the context but after #2276, we lost the context.

unstable:
```
*** Extract version and sha1 details from info command and print in tests/unit/info-command.tcl
```

now:
```
*** [err]: Extract version and sha1 details from info command and print in tests/unit/info-command.tcl
Expected '0' to be equal to '1' (context: type source line 7 file /xxx/info-command.tcl cmd {assert_equal 0 1} proc ::test)
```

We can see the different, we have provided enough context when asserting fail. Otherwise we need to scroll back (which is usually a lot of server logs) to see the context.

### DIFF
--- a/tests/test_helper.tcl
+++ b/tests/test_helper.tcl
@@ -403,7 +403,7 @@ proc test_server_cron {} {
                     if {[info exist ::active_clients_file($fd)]} {
                         set file $::active_clients_file($fd)
                     }
-                    lappend ::failed_tests "\[TIMEOUT]: $test_name in $file"
+                    lappend ::failed_tests "\[[colorstr red TIMEOUT]\]: $test_name in $file"
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


tests/test_helper.tcl
```diff
@@ -474,8 +474,7 @@ proc read_from_test_client fd {
     } elseif {$status eq {err}} {
         set err "\[[colorstr red $status]\]: $data"
         puts $err
-        set test_name [lindex [split $data "\n"] 0]
-        lappend ::failed_tests $test_name
```
enjoy-binbin, line 478: we can print the whole err actually, it have more info

ranshid: Oh so you are referring to the print at the end of the test? we do have a print when the test is reporting the error during the run right?

ranshid (approved):
Basically LGTM. There will be error print duplication, but I guess no harm in that.

tests/test_helper.tcl
enjoy-binbin: yes, the line 476 print the log, it happen during the run.

and here, the `lappend ::failed_tests`, we will print it at the end of the test.

sarthakaggarwal97: ok LGTM, if err has the test file name and the context, then we can append it directly. 
But I don't think we had context before as mentioned in #2267.

sarthakaggarwal97 (approved):

tests/test_helper.tcl
enjoy-binbin: only timeout test don't have the context and you made the change, the normal assert_xxx both have the context. Some other like may not have the context but we can fix it when we need it.

PR #2441, opened 2025-08-06
Fix primary replica chain cycle, always update sender_claimed_primary in advance

enjoy-binbin:
When writing test case for https://github.com/valkey-io/valkey/pull/2431, i found that the corresponding test
case occasionally triggered this debug assert:
```
Node ef54ff291855537b7f4f61c815955310e55bc9e2 () is no longer primary of shard ab1e16d7e11cbb0d81acf0bab4191d84e19df092; removed all 5462 slot(s) it used to own
Node ef54ff291855537b7f4f61c815955310e55bc9e2 () is now part of shard aad3ac4bf806c2bd5441ed5c421893be20f02797
Node ef54ff291855537b7f4f61c815955310e55bc9e2 () is now a replica of node a1d09af28a10d58deb43ccc8bddd8c8d095a5af5 () in shard aad3ac4bf806c2bd5441ed5c421893be20f02797

=== ASSERTION FAILED ===
==> cluster_legacy.c:7004 'primary->replicaof == ((void*)0)' is not true

------ STACK TRACE ------

Backtrace:
0   valkey-server                       0x000000010507ac70 clusterUpdateSlotsConfigWith + 0
1   valkey-server                       0x0000000105077564 clusterReadHandler + 6180
2   valkey-server                       0x00000001050f3b0c connSocketEventHandler + 180
3   valkey-server                       0x0000000104fb7f9c aeProcessEvents + 372
4   valkey-server                       0x0000000104fe2d30 main + 27676
5   dyld                                0x0000000199177154 start + 2476
```

And we can see from the cluster nodes output, there are indeed a replication loop:
```
sender: ef54ff291855537b7f4f61c815955310e55bc9e2
sender_claimed_primary: a1d09af28a10d58deb43ccc8bddd8c8d095a5af5

ef54ff291855537b7f4f61c815955310e55bc9e2 shard-id=ab1e16d7e11cbb0d81acf0bab4191d84e19df092 slave a1d09af28a10d58deb43ccc8bddd8c8d095a5af5
a1d09af28a10d58deb43ccc8bddd8c8d095a5af5 shard-id=aad3ac4bf806c2bd5441ed5c421893be20f02797 slave ef54ff291855537b7f4f61c815955310e55bc9e2
```

The sender and sender_claimed_primary are not in the same shard, and
sender_claimed_primary is a replica in myself view, if we don't update
sender_claimed_primary role, in the next `sender->replicaof = sender_claimed_primary`,
we will create the replication loop.


src/cluster_legacy.c
```diff
@@ -3824,6 +3826,14 @@ int clusterProcessPacket(clusterLink *link) {
...
                 }
 
+                if (sender_claimed_primary && nodeIsReplica(sender_claimed_primary)) {
+                    /* `primary` is still a `replica` in this observer node's view;
+                     * update its role and configEpoch */
+                    clusterSetNodeAsPrimary(sender_claimed_primary);
+                    sender_claimed_primary->configEpoch = sender_claimed_config_epoch;
+                    clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_FSYNC_CONFIG | CLUSTER_TODO_UPDATE_STATE);
+                }
+
```
madolson: I'm worried that we don't have the ability to reason about if this is right anymore. This code seems like it should be hit in the earlier case when the primary was previously a replica and is no longer a replica. I guess we had other special code for matching the shard, so maybe it's covering migration cases as well.

enjoy-binbin: We opened the gate at 28976a9003c6dd5cdd7225c5bc90743b4fcde13c, it did fix a tricky issue at the time (which we encountered a few times as well). We may need @PingXie to take a closer look, i am a little worried. Now the relationship between shard_id and processPacket is getting closer and closer, but our shard_id is not updated so in time. (I suspect this debug assertion is because the shard_id hasn't been updated correctly during the pause or the packet drop, causing it to pass the same shard condition, if we want to avoid the crash, we also need to set sender_claimed_primary to primary in the else condition, or at least fix it when we see a replication cycle.)

enjoy-binbin:
A test fail example: https://github.com/valkey-io/valkey/actions/runs/16754537649/job/47433388245?pr=2431#step:5:6004

src/cluster_legacy.c
zuiderkwast: If the sender says it's a primary and it's epoch is greater than the one we think is the primary, we should trust it, right?

zuiderkwast (approved):
Seems safe to me.

@enjoy-binbin We have some PRs that seem to affect each other. Shall we create a branch with this + #2431 + (some more PR?) to run the tests and see if solves the problems?

enjoy-binbin:
i have the memory this will solve the problem in my local, i have a fresh day today, i will work on it (merge it one by one and test it)

src/cluster_legacy.c
enjoy-binbin: this change also exposing a problem, that is, we modified the epoch of a node here.
https://github.com/valkey-io/valkey/actions/runs/16897847310/job/47870978571?pr=2441#step:7:7237

```
22304:S 12 Aug 2025 03:11:23.390 * Node 7db2919fe69b33a5bf8a70f1dec1fff19a77c8ff () is now a replica of node a0ad02b553db9024f194b8e9d2a2e8031e4428c5 () in shard 0b5882b4b1ae8066956a3875479b42f0db44e75c
22304:S 12 Aug 2025 03:11:23.397 * Replica rank updated to #0, added -1000 milliseconds of delay.
22304:S 12 Aug 2025 03:11:23.397 * Starting a failover election for epoch 11, node config epoch is 9
```

like in the log, myself update sender_claimed_primary config epoch, and sender_claimed_primary happen is myself's primary. And in the next failover cron, myself start the election with the right node epoch, and it win the election. 
```
    /* Ask for votes if needed. */
    if (server.cluster->failover_auth_sent == 0) {
        server.cluster->currentEpoch++;
        server.cluster->failover_auth_epoch = server.cluster->currentEpoch;
        serverLog(LL_NOTICE, "Starting a failover election for epoch %llu, node config epoch is %llu",
                  (unsigned long long)server.cluster->currentEpoch, (unsigned long long)nodeEpoch(myself));
        clusterRequestFailoverAuth();
```

Previously, in this situation, its nodeEpoch would be an old value, and it would not be able to win the election.

enjoy-binbin: So there are two ways to handle this i guess:
1. We check that myself->replicaof == node and cancel the election in here.
2. We record the current epoch in the failover cron. If the nodeEpoch changes during the delay, abort the failover.

enjoy-binbin: I pushed a new commit for the option 2, see if you guys will like it. We are patching the code again and again...

src/cluster_legacy.c
```diff
@@ -5273,6 +5276,18 @@ void clusterHandleReplicaFailover(void) {
 
     /* Ask for votes if needed. */
     if (server.cluster->failover_auth_sent == 0) {
+        /* Because we may update the node config epoch in processPacket function,
+         * in order to avoid us starting an old election with a new node config epoch,
+         * we will cancel it here. */
+        if (nodeEpoch(myself) > server.cluster->failover_config_epoch) {
+            server.cluster->failover_auth_time = 0;
+            serverLog(LL_WARNING, "Cancel the election that has not yet started because the node config epoch "
```
zuiderkwast: NOTICE is enough. It's not error. Usually it will resolve itself, right?

src/cluster_legacy.h
```diff
@@ -405,6 +405,7 @@ struct clusterState {
     int failover_auth_rank;           /* This replica rank for current auth request. */
     int failover_failed_primary_rank; /* The rank of this instance in the context of all failed primary list. */
     uint64_t failover_auth_epoch;     /* Epoch of the current election. */
+    uint64_t failover_config_epoch;   /* Epoch of the current primary config epoch. */
```
zuiderkwast: Repeated word epoch. Maybe like this?

```suggestion
    uint64_t failover_config_epoch;   /* Config epoch of the current primary. */
```

src/cluster_legacy.c
enjoy-binbin: ....

the reason is that after `clusterSetNodeAsPrimary(sender_claimed_primary);`, myself replica rank become No.0 since sender_claimed_primary is not a replica anymore and then myself trigger the election immediately

enjoy-binbin: pushed another new commit for the option 1 and see if i can make it right (make the CI happy)

src/cluster_legacy.c
```diff
@@ -3855,6 +3857,24 @@ int clusterProcessPacket(clusterLink *link) {
                     clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_UPDATE_STATE | CLUSTER_TODO_FSYNC_CONFIG);
                 }
 
+                if (sender_claimed_primary && nodeIsReplica(sender_claimed_primary)) {
+                    if (nodeIsReplica(myself) && areInSameShard(myself, sender_claimed_primary) &&
```
madolson, line 3861: Is it possible that the sender_claimed_primary actually has a different shard_id, since we are just learning about this through gossip? In this specific instance the sender_claimed_primary object has our our local view of its shard_id, but it may in reality have a different shard_id.

Also below, `clusterSetNodeAsPrimary` doesn't configure us to do replication, so we'll mark it as our primary but we won't start replicating from it.

enjoy-binbin:
After merging #2477, unstable now has these issue:
1. replication loop
```
clusterNode *clusterNodeGetPrimary(clusterNode *node) {
    clusterNode *primary = node;
    while (primary->replicaof != NULL) {
        primary = primary->replicaof;
        if (primary == node) break;
    }
    /* Assert that a node's replicaof/primary chain does not form a cycle. */
    debugServerAssert(primary->replicaof == NULL);
    return primary;
}
```
The #2477 test case will create a loop which trigger the assert. (This replicaof will eventually recover after some packets).

TO fix it, maybe we should always update sender_claimed_primary's flag even though we are not in the same shard.

2. And if we fix 1, alwasy update sender_claimed_primary's flag, we now meet other issue.
Do we need to consider the timeliness of epochs?
```
if (nodeEpoch(sender_claimed_primary) > sender_claimed_config_epoch) {
       serverLog(LL_NOTICE,
                 "Ignore stale message from %.40s (%s) in shard %.40s;"
                 " gossip config epoch: %llu, current config epoch: %llu",
                 sender->name, sender->human_nodename, sender->shard_id,
                 (unsigned long long)sender_claimed_config_epoch,
                 (unsigned long long)nodeEpoch(sender_claimed_primary));
```
we have this code, and in the old slot migration case, we will incr epoch when we import a slot, the new epoch may not reach all nodes in time, so some nodes will know the new epoch and some other won't. So some node will drop the packet since it think the packet is stale.

3. Since now we will update sender_claimed_primary's flag by calling clusterSetNodeAsPrimary, it will also remove the replica from the primary.
So there are a test case, that replica1 and replica2 doing the failover, replica1 won the election, and replica2 are still doing the delay. At this moment, replica2 learn the message and call this function, and then it's replica_rank become 0 and trigger the new election. The new election will win again since it have the newly epoch during the packet. This is somehow like two replicas winning the election in the same epoch.
```
void clusterSetNodeAsPrimary(clusterNode *n) {
    if (clusterNodeIsPrimary(n)) return;

    if (n->replicaof) {
        clusterNodeRemoveReplica(n->replicaof, n);
        if (n != myself) n->flags |= CLUSTER_NODE_MIGRATE_TO;
    }
    n->flags &= ~CLUSTER_NODE_REPLICA;
    n->flags |= CLUSTER_NODE_PRIMARY;
    n->replicaof = NULL;

    /* Update config and state. */
    clusterDoBeforeSleep(CLUSTER_TODO_SAVE_CONFIG | CLUSTER_TODO_UPDATE_STATE);
}
```

madolson:
> TO fix it, maybe we should always update sender_claimed_primary's flag even though we are not in the same shard.

Can we consider not marking a node to be a primary based off of another replica node claiming it as a primary? These problems have been endemic since we introduced the logic, since there are a bunch of odd cases.

PR #2442, opened 2025-08-06
Adds benchmark on demand workflow

roshkhatri:
# Adds On-demand Benchmark Workflow

This PR introduces a new GitHub Actions workflow that enables on-demand performance benchmarking for pull requests through label-based triggers. This uses the new framework [valkey-perf-benchmark](https://github.com/valkey-io/valkey-perf-benchmark) developed for standard benchmarks and PR benchmarking.

## What is being added by this PR

- **Workflow File**: `.github/workflows/benchmark_on_label.yml`
- **Trigger**: Activated when specific labels are added to PRs
- **Supported Labels**:
  - `run-benchmark` - Runs standard benchmarks
  
### On-Demand Execution
- Benchmarks run only when explicitly requested via PR labels
- No automatic execution to avoid unnecessary resource usage

### Performance Comparison
- Compares PR performance against the `unstable` baseline
- Generates detailed comparison reports
- Posts results directly as PR comments for easy review

### Flexible Configuration
- Currently uses github runners by will use dedicated performance runners (`ec2-perf-ubuntu-24`)
- Configurable benchmark suites via JSON config files

### Artifact Management
- Uploads benchmark results as workflow artifacts
- Preserves both baseline and PR metrics for analysis
- Includes comparison markdown for offline review

### Automatic Cleanup
- Removes trigger labels after execution (success or failure)
- Prevents accidental re-runs from stale labels

## Usage

To run benchmarks on a PR:

1. Add the `run-benchmark` label for standalone and cluster, non-tls mode
2. Wait for the workflow to complete
3. Review results in the automated PR comment

This workflow enhances our CI/CD pipeline by providing easy access to performance testing without impacting regular development workflows.

Github Bot comments the results something like below:
Screenshot 2025-08-06 at 3 44 23 PM


roshkhatri:
The reason this fails to comment on the PR is due to the `github-token: ${{ secrets.GITHUB_TOKEN }}`
We would need to create one for this like `${{ secrets.CODECOV_TOKEN }}` for letting the bot comment on the PR @madolson 

This would create something like this: https://github.com/roshkhatri/valkey/pull/2#issuecomment-3162217043

roshkhatri:
I have update the workflow to run on `pull_request_target` trigger so it runs with the base commit context with access to the secrets.

As we are running this on self hosted, we are checking out `ref: refs/pull/${{ github.event.pull_request.number }}/merge` which is not the authors commit for security reasons, but this ref is github generated merge on the base that is `unstable commit. 

As this workflow is not in the base and it runs on the base commits context, it would not run on this PR.

I tested this with on my personal repo with this workflow in unstable and it works: https://github.com/roshkhatri/valkey/pull/3

Screenshot 2025-08-11 at 1 38 26 PM

The double print will be removed from the benchmark tool itself, but this workflow is ready to be merged.

hpatro (commented):
Why tls with cluster mode and non tls for standalone ?

I think we can just stick with non TLS mode for both type.

.github/workflows/benchmark-on-label.yml
```diff
@@ -0,0 +1,128 @@
+name: On-demand Benchmark
+
+on:
+  pull_request_target:
+    types: [labeled]
+
+concurrency:
+  group: valkey-pr-benchmark
+  cancel-in-progress: false
```
hpatro, line 9: Is this cancel-in-progress for the same PR? If yes, we should stop the running ones on new commit.

roshkhatri: cancel-in-progress is for all instances of this workflow as we are using only one self hosted runner. We can parallelize it further in some iterations

The benchmark will only run if you add run-benchmark label, the label will be removed once the workflow is complete.

roshkhatri:
>Why tls with cluster mode and non tls for standalone ?
I think we can just stick with non TLS mode for both type.

Yeah I can do that

hpatro (commented):
Could we just create one label overall for a person to tag and execute different benchmarking behind the scenes ?

.github/benchmark_configs/pr_benchmark_cluster.json
```diff
@@ -0,0 +1,25 @@
...
+      "MSET",
+      "INCR",
+      "LPUSH",
+      "RPUSH",
+      "LPOP",
+      "RPOP",
+      "SADD",
+      "HSET",
+      "ZADD",
+      "XADD"
```
hpatro:
```suggestion
      "SET",
      "GET"
```

.github/benchmark_configs/pr_benchmark_standalone.json
```diff
@@ -0,0 +1,25 @@
...
+      "GET",
+      "RPUSH",
+      "LPUSH",
+      "LPOP",
+      "SADD",
+      "SPOP",
+      "HSET",
+      "MGET",
+      "LRANGE",
+      "ZPOPMIN"
```
hpatro:
```suggestion
      "SET",
      "GET",
      "MGET",
      "MSET"
```

hpatro (commented):
I believe we need to merge these two json objects into one file.

.github/workflows/benchmark-on-label.yml
sarthakaggarwal97: sorry, but maybe I didn't understand, what happens if there is a new commit? I am expecting that the benchmark workflow will cancel and label be removed automatically.

.github/workflows/benchmark-on-label.yml
```diff
@@ -0,0 +1,122 @@
...
+
+      - name: Install dependencies
+        working-directory: valkey-perf-benchmark
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y build-essential
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt
+
+      - name: Run benchmarks
```
sarthakaggarwal97, line 55: is there an option to add timeouts here? Maybe there are cases where the run is taking quite long to complete?

.github/workflows/benchmark-on-label.yml
```diff
@@ -0,0 +1,122 @@
...
+        with:
+          repository: valkey-io/valkey-perf-benchmark
+          path: valkey-perf-benchmark
+          fetch-depth: 1
+          persist-credentials: false
+
+      - name: Set up Python
+        uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5
+        with:
+          python-version: "3.10"
```
sarthakaggarwal97, line 45: should we add an option to cache? https://github.blog/changelog/2021-11-23-github-actions-setup-python-now-supports-dependency-caching/

sarthakaggarwal97 (commented):
Thanks @roshkhatri for the PR. Some minor comments from my end!

sarthakaggarwal97:
@roshkhatri how much time does it typically take for the benchmark to run? Just checking if it is too long, so we can change the parameters a bit.

.github/workflows/benchmark-on-label.yml
roshkhatri: The current run benchmark will complete and remove the label

.github/workflows/benchmark-on-label.yml
roshkhatri: with the current config file it takes around 20 mins. on the self hosted might take a little less.

The time would totally depend on the commits performance. yeah we can add a time out for an hour, worst case we can always increase it

.github/workflows/benchmark-on-label.yml
roshkhatri: Good idea, will add it

.github/workflows/benchmark-on-label.yml
sarthakaggarwal97: is it possible to make it cancellable?

hpatro: @roshkhatri Could we print out the commit-id in the results on which the benchmark ran ?

hpatro (approved):
A good starting point. Thanks for adding this feature @roshkhatri !

.github/workflows/benchmark-on-label.yml
sarthakaggarwal97: +1 a link would even be better.

sarthakaggarwal97 (approved):
LGTM! Thanks @roshkhatri for taking up this effort!

madolson (approved):

.github/workflows/benchmark-on-label.yml
```diff
@@ -0,0 +1,123 @@
...
+            const body = fs.readFileSync('comparison.md', 'utf8');
+            await github.rest.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body
+            });
+
+      - name: Remove ${{ github.event.label.name }} label
+        if: always() && (github.event.label.name == 'run-benchmark' || github.event.label.name == 'run-cluster-benchmark')
```
hpatro: We don't need the `run-cluster-benchmark` label further. What does `always()` do ?

```suggestion
        if: always() && github.event.label.name == 'run-benchmark'
```

PR #2448, opened 2025-08-07
Compile with fast_float by default

zuiderkwast:
Changes the Makefile to compile with USE_FAST_FLOAT=yes by default.

When fast_float was added in a minor version, we didn't want to compile with it by default, but we can change that in the upcoming major version 9.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       - name: make
         # Fail build if there are warnings
         # build with TLS just for compilation coverage
-        run: make -j4 all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes USE_FAST_FLOAT=yes
+        run: make -j4 all-with-unit-tests SERVER_CFLAGS='-Werror' BUILD_TLS=yes
       - name: install old server for compatibility testing
         run: |
           cd tests/tmp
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -112,15 +112,15 @@ jobs:
...
+          make -j4 SERVER_CFLAGS='-Werror' USE_FAST_FLOAT=no
 
   build-macos-latest:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4.1.1
       - name: make
         # Build with additional upcoming features
-        run: make -j3 all-with-unit-tests SERVER_CFLAGS='-Werror' USE_FAST_FLOAT=yes
+        run: make -j3 all-with-unit-tests SERVER_CFLAGS='-Werror'
```
zuiderkwast: Dummy change to trigger run-extra-tests.

```suggestion
        run: make -j3 all-with-unit-tests SERVER_CFLAGS='-Werror' USE_FAST_FLOAT=yes
```

zuiderkwast:
:sleepy: I forgot that I already tried this before in #2295 and it needs some changes to some environments.....

madolson:
@roshkhatri Do we compile this by default for the container?

roshkhatri:
Yes we do compile containers with fast float by default.

madolson:
We did this already with https://github.com/valkey-io/valkey/pull/3329/changes#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5

PR #2449, opened 2025-08-07
Make cluster failover delay relative to node timeout

zuiderkwast:
In clusters with a very short node timeout such as 2-3 seconds, the extra failover delay of 500-1000 milliseconds (500 + random value 0-500; total 750 on average) before initiating a failover is a significant extra downtime to the cluster. This PR makes this delay relative to node timeout, using a shorter failover delay for a smaller configured node timeout. The formula is `fixed_delay = min(500, node_timeout / 30)`.

| Node timeout  | Fixed failover delay |
|---------------|----------------------|
| 15000 or more | 500 (same as before) |
| 7500          | 250                  |
| 3000          | 100                  |
| 1500          | 50                   |

Additional change: Add an extra 500ms delay to new replicas that may not yet know about the other replicas. This avoids the scenario where a new replica with no data wins the failover. This change turned out to be needed to for the stability of some test cases.

The purposes of the failover delay are

1. Allow FAIL to propagate to the voting primaries in the cluster
2. Allow replicas to exchange their offsets, so they will have a correct view of their own rank.

A third (undocumented) purpose of this delay is to allow newly added replicas to discover other replicas in the cluster via gossip and to compute their rank, to realize it's are not the best replica. This case is mitigated by adding another 500ms delay to new replicas, i.e. if it has replication offset 0.

A low node timeout only makes sense in fast networks, so we can assume that the above needs less time than in a cluster with a higher node timeout.

These delays don't affect the correctness of the algorithm. They are just there to increase the probability that a failover will succeed by making sure that the FAIL message has enough time to propagate in the cluster and to the random part is to reduce the probability that two replicas initiates the failover at the same time.

The typical use case is when data consistency matters and writes can't be skipped. For example, in some application, we buffer writes in the application during node failures to be able to apply them when the failover is completed. The application can't buffer them for a very long time, so we need the cluster to be up again within e.g. 5 seconds from the time a node starts to fail.

I hope this PR can be considered safer than #2227, although the two changes are orthogonal.

Part of issue #2023.


hpatro (commented):
This looks safe to me.

Do you think we need the minimum set to 500 milliseconds or just keep everything relative to node timeout?

zuiderkwast:
> Do you think we need the minimum set to 500 milliseconds or just keep everything relative to node timeout?

I did use 500 as a maximum, not a minimum. This is for any cluster using node timeout higher than 15 seconds, to not make failover slower than now for them. I don't know if there are any such clustes though and whether it matters.

hpatro:
> > Do you think we need the minimum set to 500 milliseconds or just keep everything relative to node timeout?
> 
> 
> 
> I did use 500 as a maximum, not a minimum. This is for any cluster using node timeout higher than 15 seconds, to not make failover slower than now for them. I don't know if there are any such clustes though and whether it matters.

Yes, I wrote it as verbatim as the code is. We use 15 seconds as the timeout so this change doesn't have any impact on our setup. 

I think keeping it relative keeps the logic simpler and the additional delay shouldn't bother them as the time it takes to fail a node is already set to a high value.

enjoy-binbin (approved):
We've discussed this a lot before. I am ok with this change.

PR #2453, opened 2025-08-07
Fix expectations for manual failover tests

amanosme:
Test `Instance #5 is still a slave after some time (no failover)` is supposed to verify that command `CLUSTER FAILOVER` will not promote a replica without quorum from the primary; later in the file (`Instance 5 is a master after some time`), we verify that `CLUSTER FAILOVER FORCE` does promote a replica under the same conditions.

There's a couple issues with the tests:

1. `Instance #5 is still a slave after some time (no failover)` should verify that instance 5 is a replica (i.e. that there's no failover), but we call `assert {[s -5 role] eq {master}}`. 
2. The reason why the above assert works is that we previously send `DEBUG SLEEP 10` to the primary, which pauses the primary for longer than the  [configured 3 seconds for `cluster-node-timeout`](https://github.com/valkey-io/valkey/blob/unstable/tests/support/cluster_util.tcl#L266). The primary is marked as failed from the perspective of the rest of the cluster, so quorum can be established and instance 5 is promoted as primary.

This commit fixes the two by shortening the sleep to less than 3 seconds, and then asserting the role is still replica. Test `Instance #5 is a master after some time` is updated to sleep for a shorter duration to ensure that `FAILOVER FORCE` succeeds under the exact same conditions.

### Testing
`./runtest --single unit/cluster/manual-failover --loop --fastfail`


sarthakaggarwal97 (approved):

enjoy-binbin (approved):
This seem corret, did you encounter this failure somewhere? Or did you just doing a code review and find this?

amanosme:
@enjoy-binbin I caught this while debugging the test under valgrind. I think valgrind slows the tests down enough to the point where the primary would be back up when the `CLUSTER FAILOVER` command was issued (and the check for master role failed)

PR #2454, opened 2025-08-08
Fix AOF rewrite behavior for hashes with expirations

murphyjacob4:
The bug caused many invalid HMSET entries to be added to the AOF during rewriting.

This now properly skips emitting HMSET until we find an entry with no expiry.

### DIFF
--- a/tests/unit/hashexpire.tcl
+++ b/tests/unit/hashexpire.tcl
@@ -2693,119 +2693,166 @@ start_cluster 3 0 {tags {"cluster mytest external:skip"} overrides {cluster-node
 }
 
 #### AOF Test #####
+proc validate_aof_content {aof_file pxat_count hdel_count} {
+    wait_for_condition 100 100 {
+        [file exists $aof_file] eq 1
+    } else {
+        fail "hash value was not expired after timeout"
+    }
+
+    set aof_content [exec cat $aof_file]
+
+    # Verify amount of PXAT and HDEL
+    # Count PXAT commands
+    set got_pxat_count [regexp -all {PXAT} $aof_content]
+    assert_equal $got_pxat_count $pxat_count
+    # Count HDEL commands
+    set got_hdel_count [regexp -all {HDEL} $aof_content]
+    assert_equal $got_hdel_count $hdel_count
+}
 tags {"aof external:skip"} {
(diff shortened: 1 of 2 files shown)


tests/unit/hashexpire.tcl
```diff
@@ -2803,11 +2806,55 @@ tags {"aof external:skip"} {
...
+                # Get the last base AOF file path and validate its content
+                # Count PXAT commands (should be 10: just the long expiry fields)
+                # Count HDEL commands (should be rewritten out)
+                validate_aof_content [get_base_aof_path r] 10 0
+            }
+
+            # Restart the server and load the AOF
+            restart_server 0 true false
+            r debug loadaof
+            r DEBUG SET-ACTIVE-EXPIRE 0
```
ranshid: I think it is possible that active expiration will be able to remove some elements in the time between these operations, suggest to switch
```suggestion
            r DEBUG SET-ACTIVE-EXPIRE 0
            r debug loadaof
```

tests/unit/hashexpire.tcl
```diff
@@ -2693,12 +2693,30 @@ start_cluster 3 0 {tags {"cluster mytest external:skip"} overrides {cluster-node
 }
 
 #### AOF Test #####
+proc validate_aof_content {aof_file pxat_count hdel_count} {
```
ranshid, line 2696: nice. I would like if we also validate the number of keys and the number of fields, but not critical

ranshid (approved):
LGTM

Some small test comments
I think onlt one is important for making sure the test will not be flaky

tests/unit/hashexpire.tcl
murphyjacob4: Makes sense, also updated it above

tests/unit/hashexpire.tcl
murphyjacob4: It makes sense to me that we would validate the raw contents here (how many of each command do we see) and validate the semantics (# of keys, # of fields) by checking the state after loading the AOF.

We could count the raw number of lines as well, but I am worried about that being a bit brittle (e.g. we'd have to include SELECT and any other boilerplate commands).

ranshid:
Before we continue, lets understand the test failure as it might be related and is not something I see on dailies

ranshid:
I am not sure why the memory leak in this case, will followup to check if it appears in dailies.

PR #2455, opened 2025-08-08
Fix and remove conflicting paths from clang-format workflow

hanxizh9910:
Fixes the GitHub Actions error where both `paths` and `paths-ignore` were defined for the same event, which is not allowed.

Removed the conflicting `paths` section from the `pull_request` trigger, keeping only `paths-ignore` to skip documentation changes while allowing the workflow to run on all other changes.

Tested the path filters by creating a PR to my own fork and verified that the Documentation-only changes correctly skip the unnecessary tests.

Screenshot 2025-08-07 at 10 37 22 PM

Resolves the error: "you may only define one of `paths` and `paths-ignore` for a single event"

This is a follow-up fix to address the issue identified in the previous PR.

### DIFF
--- a/.github/workflows/clang-format.yml
+++ b/.github/workflows/clang-format.yml
@@ -7,8 +7,6 @@ on:
       - '**/00-RELEASENOTES'
       - '**/COPYING'
   pull_request:
-    paths:
-      - 'src/**'
     paths-ignore:
       - '**/*.md'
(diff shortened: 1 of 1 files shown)


enjoy-binbin (approved):

sarthakaggarwal97 (commented):
LGTM

Issue #2065, opened 2025-05-08
[New] IO threads utilization INFO field


zuiderkwast:
When IO threads are running, they use busy-wait loop to wait for jobs. Therefore, they run at 100% all the time and CPU utilization is not a useful metric to understand the load.

When I/O threads are used, the main thread does non-blocking `epoll()` so it too does busy-looping and runs at 100% CPU. Therefore we need a CPU utilization metric for the main thread too.

The problem is explained in this issue:

* #2055

### Proposed INFO fields

Measure the "active" running time = wall clock minus the time spent on busy-waiting and blocking epoll.

```diff
 # CPU
 used_cpu_sys:68.476389
 used_cpu_user:219.408941
 used_cpu_sys_children:0.003157
 used_cpu_user_children:0.008264
 used_cpu_sys_main_thread:68.478353
 used_cpu_user_main_thread:219.404508
+used_active_time_main_thread:123.456789
+used_active_time_io_thread_1:97.013234
+used_active_time_io_thread_2:45.678901
```

Implementation in two PRs:

* #2463 (for I/O threads)
* #2931 (for the main thread)

### Alternatives considered

**Naming**

* "used": The word "used" is already used for CPU time actually consumed by user space or kernel space in the INFO fields `used_cpu_sys_main_thread` and `used_cpu_user_main_thread`.
* "useful": This word is a little similar to "used" so they can be confusing.
* "utilization": This word is a little long and also similar to "useful" and "used".
* "busy": This work is often used in words like busy-wait looping, which is the non-useful activity. It may be confusing to use this word for the useful time metric.
* "active": This is my current preference. We can talk about CPU utilization as "active CPU time" = "used CPU time" minus "busy-waiting".

**Why not more fields for I/O threads?**

Since I/O threads always run at 100% CPU, there's not much point in presenting the actual used CPU time.

**Why measure active time using monotonic clock?**

An alternative to measuring the duration of active work using a monotonic clock is to use CPU-time clock, `clock_gettime()`. However, a low CPU usage is no guarantee that the thread has spare capacity to handle more work. In addition to waiting for work, it may be preempted by the kernel or wait for blocking IO or other syscalls.

OTOH, measuring active time using a monotonic clock and it revelase that 90% of the time consists of active work, it implies that the remaining 10% of the time, the thread has been waiting for more work. This is inline with the purpose of these metrics: to monitor the spare capacity of the threads.

**Why not present a percentage?**

[...]

PR #2463, opened 2025-08-10
Cumulative metrics for active I/O threads usage

deepakrn:
I/O threads when active do busy polling for work. This puts the CPU time near 100% per thread even when they aren't fully utilized. Tracking the time an IO thread spends doing useful work vs waiting for work helps understand the utilization of individual threads.

INFO fields:

    used_active_time_io_thread_%d

... where %d is the IO thread ID number (between 1 and the number of I/O threads).

To get a percentage over a period of time, compare the delta of this metric against a delta of the server's uptime or the INFO field `server_time_usec`.


deepakrn:
This change aims to address https://github.com/valkey-io/valkey/issues/2065 except for tracking the main_thread utilization.

zuiderkwast:
@deepakrn Good, I hope we can also find a way to track main thread usefulness.

Can you link the PR to the issue where this was discussed?

src/io_threads.c
```diff
@@ -255,11 +264,22 @@ static void *IOThreadMain(void *myid) {
...
 }
 
+long long getIOThreadUsefulTimeMicroseconds(int id) {
+    return atomic_load_explicit(&io_threads_useful_time_us[id], memory_order_acquire);
+}
+
+long long getIOThreadUptimeMicroseconds(int id) {
+    return atomic_load_explicit(&io_threads_uptime_us[id], memory_order_acquire);
+}
+
```
ranshid: can you explain why is it so important to use acquire load in this case?

src/io_threads.h
```diff
@@ -3,6 +3,10 @@
 
 #include "server.h"
 
+#define TIMESPEC_DIFF_US(start_time, end_time)                                     \
+    (((long long)(end_time).tv_sec - (long long)(start_time).tv_sec) * 1000000LL + \
+     ((long long)(end_time).tv_nsec - (long long)(start_time).tv_nsec) / 1000L)
```
zuiderkwast: `static inline` function is almost always better than macro. We get type checking and better readability. It's still defined in the header file like a macro.

```suggestion
static inline long long timespec_diff_us(struct timespec start_time, struct timespec end_time) {
    return ((long long)end_time.tv_sec - (long long)start_time.tv_sec) * 1000000LL +
           ((long long)end_time.tv_nsec - (long long)start_time.tv_nsec) / 1000LL;
}
```

src/io_threads.h
```diff
@@ -3,6 +3,11 @@
 
 #include "server.h"
 
+static inline long long timespec_diff_us(struct timespec start_time, struct timespec end_time) {
```
deepakrn: @zuiderkwast  - sorry, I was trying to create a single commit for my changes and it looked like I lost your comment. I have replaced the macro with the inline function as per your suggestion.

zuiderkwast: OK, no worries, but you don't need a single commit. We will squash-merge anyway. If you make individual commits, it's easier to follow up when reviewing the changes since the last review.

src/io_threads.c
zuiderkwast: It's only used for printing the INFO fields, right? Yeah, why is not relaxed ordering enough?

deepakrn: Thanks for the callout @ranshid and @zuiderkwast . It makes sense to use relaxed, agreed. Updated.

dvkashapov:
Do we have any concerns against this approach? The only concern for me is that we would print times for all io-threads, should we cap number of printed thread times to 16 for example?

deepakrn:
Thanks for the question @dvkashapov. 

See https://github.com/valkey-io/valkey/issues/2065#issuecomment-2875398377 for relevant discussion. Since setups with thread count > 16 are rare, we thought we could start without that optimization.

zuiderkwast:
@deepakrn Finally, we have made a decision about these INFO fields. Sorry for the delay. What we decided is written in the issue https://github.com/valkey-io/valkey/issues/2065:

* Use field names `used_cpu_active_io_thread_1`, `_2`, etc.
* Don't introduce the thread uptime field. (Users can just compare the active CPU time between two different INFO calls.)
* Make sure that we don't reset the number to zero if the number of IO threads decrease and then increase again when the configured number of I/O threads is changed during runtime. (We should add a test case for this scenario.)

Do you want to update this PR to match the above?

Now, we also have the implementation for the main thread CPU utilization #2931 by Daniil, so I'm excited to get both of these PRs released in 9.1.

deepakrn:
Hi @zuiderkwast , Thanks for the update and the idea makes sense to me. I'll work on making these changes.

deepakrn:
@zuiderkwast - couple of follow ups.

> Don't introduce the thread uptime field. (Users can just compare the active CPU time between two different INFO calls.)

If we are getting rid of thread uptime field, there wouldn't be an easy way for the user to know if the thread was running but was busy looping waiting for work? Wouldn't that be a gap in observability? The user could still track total time elapsed and look at delta of used_active_cpu_time but that wouldn't tell for how long the IO thread was enabled in that duration?

> Make sure that we don't reset the number to zero if the number of IO threads decrease and then increase again when the configured number of I/O threads is changed during runtime. (We should add a test case for this scenario.)

What shouldn't be reset to zero? Are you referring to the active cpu time? If yes, isn't that already the case? Also, I am assuming that INFO should display the active CPU time of all IO threads with value > 0?

dvkashapov:
> The user could still track total time elapsed and look at delta of used_active_cpu_time but that wouldn't tell for how long the IO thread was enabled in that duration?

From relevant issue:

> Even without it, the caller can compare the the delta between two points in time and use this delta to compare the I/O thread's active time delta between two INFO calls. To calculate the exact time between two calls to INFO, the delta of the INFO field for the server's uptime can be used.

> What shouldn't be reset to zero? Are you referring to the active cpu time? If yes, isn't that already the case?

Looks like you're right about not resetting cpu time currently. The case we want to make sure works is not resetting metric when io thread becomes active->not active->active again.

zuiderkwast:
> If we are getting rid of thread uptime field, there wouldn't be an easy way for the user to know if the thread was running but was busy looping waiting for work? Wouldn't that be a gap in observability? The user could still track total time elapsed and look at delta of used_active_cpu_time but that wouldn't tell for how long the IO thread was enabled in that duration?

I don't think it's important to know whether the thread was disabled or just waiting during some time.

To get a percentage between two points in time, you can for example take the delta of the active CPU time for an IO thread and divide by the delta of the `server_time_usec` INFO field.

> What shouldn't be reset to zero? Are you referring to the active cpu time? If yes, isn't that already the case? Also, I am assuming that INFO should display the active CPU time of all IO threads with value > 0?

I believe that's already the case yes, but I'm not completely sure. I haven't verified it. If it's not too much work to write a test case for this, it would be nice to have.

INFO should display the active CPU time of the IO threads that are currently configured in the config `io-threads` config, even if some of them report zero. If we have had more I/O threads configured in the past, these threads will not run and will not change their CPU time. We don't need to report them in INFO every time. A server can run for years.

src/io_threads.c
```diff
@@ -238,9 +242,14 @@ static void *IOThreadMain(void *myid) {
         if (jobs_to_process == 0) {
             pthread_mutex_lock(&io_threads_mutex[id]);
             pthread_mutex_unlock(&io_threads_mutex[id]);
+            clock_gettime(CLOCK_THREAD_CPUTIME_ID, &thread_uptime_end);
+            atomic_fetch_add_explicit(&io_threads_uptime_us[id], timespec_diff_us(thread_uptime_start, thread_uptime_end), memory_order_relaxed);
             continue;
         }
 
+        struct timespec work_start_time, work_end_time;
+        clock_gettime(CLOCK_THREAD_CPUTIME_ID, &work_start_time);
```
zuiderkwast: Here, we call clock_gettime before and after the active work in each cycle.

I believe it's possible to do it only once per cycle, if there is work to do and no busy-loop spinning happens. Idea using pseudo-C code (using the CPU clock time as an integer):

```C
static void *IOThreadMain(void *myid) {
    // ...

    work_start_time = 0;
    while (1) {
        /* Cancellation point so that pthread_cancel() from main thread is honored. */
        pthread_testcancel();

        /* Active CPU time, excluding busy-wait loop */
        prev_start_time = work_start_time;
        clock_gettime(CLOCK_THREAD_CPUTIME_ID, &work_start_time);
        if (prev_start_time != 0) {
            atomic_fetch_add_explicit(&io_threads_active_us[id],
                                      work_start_time - prev_start_time,
                                      memory_order_relaxed);
        }

        jobs_to_process = IOJobQueue_availableJobs(jq);
        if (jobs_to_process == 0) {
            /* Wait for jobs */
            for (int j = 0; j < 1000000; j++) {
                jobs_to_process = IOJobQueue_availableJobs(jq);
                If (jobs_to_process) break;
            }
            clock_gettime(CLOCK_THREAD_CPUTIME_ID, &work_start_time);
        }        

        // ...        
    }
}
```

zuiderkwast:
Update.

I have been discussing back and forth with various people weather these metrics should measure the CPU-time during active work or just monotonic time of active work.

Currently, the preference is to use the monotonic clock, not the CPU-time clock. The reason is that a low CPU usage doesn't necessarily mean that the thread has spare capacity to handle more work.

If we instead measure monotonic time and see that 90% of the time, the thread is doing work, it means that the remaining 10%, the thread is waiting for more work. This is better information, because the purpose of these metrics is to be able to tell how much more work the I/O threads can handle.

For a longer expanation, see this https://github.com/valkey-io/valkey/pull/2931#issuecomment-3753992616.

@deepakrn do you still think CPU-time is better? If there are any more aspects to consider, I'm interested so we can be sure we make the right decision.

If we decide to use the monotonic clock, we shouldn't include "cpu" in the field names, because it can be misleading. We can rename them to `used_active_time_io_thread_1`.

deepakrn:
@zuiderkwast - Clarifying my thought process below:

If we do monotonic clock, we won't have visibility into whether the thread was context switched or was actually doing work in the Valkey process.

If we do CPU clock, it tells us the amount of time that was spent running the process excluding blocking syscalls, preemptions etc.

CPU clock time < monotonic clock time < total time

Understanding the metric with an example:

If (CPU_clock_time_delta/total_time) for a io_thread is 80%, it means 80% of the time was spent doing real work in Valkey and 20% comprises - context_swtich, blocking syscall, busy polling.

if (monotic_time_delta/total_time) for a io_thread is 80%, it means 80% comprises of (context_swith + blocking syscall + active work) and 20% comprises primarily of (busy polling + some more context switch potentially).

So, (100 - monotonic_time_delta/total_time) could be closer to the time spent busy polling (and hence potential free capacity) than (100 - cpu_time_delta/total_time). 

If my understanding is correct, it seems like monotonic_clock would better tell us about free capacity than cpu_time. WDYT?

zuiderkwast:
Yes, exactly, good explanation. We came to the same conclusion.

* Active time = time trying to work = (cpu-time + context switch + blocking syscalls)
* Not active time = time waiting for work

So let's change to getMonotonicUs and the name `used_active_time_io_thread_%d`. When you have time.

src/io_threads.c
deepakrn: Nice! That is a pretty clever optimization. Thank you! I have updated the code.

deepakrn: I am yet to make the below changes:
1. Adding tests to ensure that the io_threads_active_us doesn't get reset to 0 when the thread becomes inactive.
2. Always display the io_threads_active_us values for all the active threads even when they are 0.

I'll get to them soon :)

src/io_threads.c
```diff
@@ -224,14 +225,26 @@ static void *IOThreadMain(void *myid) {
...
     while (1) {
         /* Cancellation point so that pthread_cancel() from main thread is honored. */
         pthread_testcancel();
 
-        /* Wait for jobs */
-        for (int j = 0; j < 1000000; j++) {
-            jobs_to_process = IOJobQueue_availableJobs(jq);
-            if (jobs_to_process) break;
+        prev_work_start_time = work_start_time;
+        clock_gettime(CLOCK_MONOTONIC, &work_start_time);
```
zuiderkwast: Looks good, but let's use our internal `getMonotonicUs` (defined in monotonic.c), like we did in the main thread metrics PR #2931. It returns a 64-bit integer already so we don't need to do the struct timespec calculations. Also there is potential that it can be faster than clock_gettime with #2597.

deepakrn: Thank you! Addressed this comment.

src/io_threads.c
deepakrn: Updated the PR.

src/io_threads.c
```diff
@@ -260,6 +273,10 @@ static void *IOThreadMain(void *myid) {
     return NULL;
 }
 
+long long getIOThreadUsefulTimeMicroseconds(int id) {
```
zuiderkwast: Let's rename this to match the new name of the INFO field.

```suggestion
long long getIOThreadActiveTimeMicroseconds(int id) {
```

src/io_threads.h
```diff
@@ -3,6 +3,11 @@
...
+    return ((long long)end_time.tv_sec - (long long)start_time.tv_sec) * 1000000LL +
+           ((long long)end_time.tv_nsec - (long long)start_time.tv_nsec) / 1000LL;
+}
+
```
zuiderkwast: This is unused now (right?) so we can delete it.

src/server.c
```diff
@@ -6414,6 +6414,10 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                             (long)m_ru.ru_stime.tv_sec, (long)m_ru.ru_stime.tv_usec, (long)m_ru.ru_utime.tv_sec,
                             (long)m_ru.ru_utime.tv_usec);
 #endif /* RUSAGE_THREAD */
+        for (int i = 1; i < server.io_threads_num; i++) {
+            long long used_active_time_io_thread = getIOThreadUsefulTimeMicroseconds(i);
+            info = sdscatprintf(info, "used_active_time_io_thread_%d:%lld.%06lld\r\n", i, used_active_time_io_thread / 1000000, used_active_time_io_thread % 1000000);
+        }
         long long active_seconds = server.stat_active_time / 1000000;
         long long active_microseconds = server.stat_active_time % 1000000;
         info = sdscatprintf(info,
```
zuiderkwast: Let's output `used_active_time_main_thread` first and the `used_active_time_io_thread_%d` fields below it.

tests/unit/io-threads.tcl
```diff
@@ -0,0 +1,48 @@
+
+start_server {config "minimal.conf" tags {"external:skip"} overrides {enable-debug-command {yes} io-threads 5}} {
+    set server_pid [s process_id]
+    # Skip if non io-threads mode - as it is relevant only for io-threads mode
+    if {[r config get io-threads] ne "io-threads 1"} {
```
zuiderkwast: When we start server with `io-threads 5` in overrides, I don't think it can ever be 1 here. I think we can delete this `if` check.

If you are concerned that it can be 1 here, we can add an assert that it is actually 5, like this:

```suggestion
    assert_equal [r config get io-threads] "io-threads 5"
```

tests/unit/io-threads.tcl
```diff
@@ -0,0 +1,48 @@
...
+            }
+            # Resume the server
+            resume_process $server_pid
+            
+            # Wait until active io-threads count is 1
+            wait_for_condition 1000 50 {
+                [getInfoProperty [r info server] io_threads_active] eq 1
+            } else {
+                fail "Failed to wait for active io_threads to come down to 1"
+            }
```
zuiderkwast: This test case seems to fail, but I think it doesn't tests exactly what we want to test. What we discussed in the core team meeting (mentioned here: https://github.com/valkey-io/valkey/issues/2065#issuecomment-3739555507) is that we want to test changing the config `io-threads` in runtime.

High level idea:

1. Start with `io-threads 5`.
2. Check that INFO reports active time > 0 for all threads.
3. `r config set io-threads 1`
4. Check that INFO doesn't returns active for the IO threads.
5. `r config set io-threads 5`
6. Check that INFO returns active time for the threads again and didn't reset the number. Check new_value >= old_value for each threads.

src/server.c
```diff
@@ -6414,6 +6414,10 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                             (long)m_ru.ru_stime.tv_sec, (long)m_ru.ru_stime.tv_usec, (long)m_ru.ru_utime.tv_sec,
                             (long)m_ru.ru_utime.tv_usec);
 #endif /* RUSAGE_THREAD */
+        for (int i = 1; i < server.io_threads_num; i++) {
+            long long used_active_time_io_thread = getIOThreadUsefulTimeMicroseconds(i);
+            info = sdscatprintf(info, "used_active_time_io_thread_%d:%lld.%06lld\r\n", i, used_active_time_io_thread / 1000000, used_active_time_io_thread % 1000000);
```
zuiderkwast: It's a very long line. Let's split it.

```suggestion
            info = sdscatprintf(info,
                                "used_active_time_io_thread_%d:%lld.%06lld\r\n",
                                i,
                                used_active_time_io_thread / 1000000,
                                used_active_time_io_thread % 1000000);
```

src/server.c
deepakrn: Done

src/server.c
deepakrn: Done

src/io_threads.c
deepakrn: Done.

tests/unit/io-threads.tcl
deepakrn: Thanks! updated.

tests/unit/io-threads.tcl
deepakrn: Ah, I didn't realize that we wanted to resize the io-threads configuration as well. SG. I have updated the tests.

src/io_threads.c
```diff
@@ -224,14 +225,26 @@ static void *IOThreadMain(void *myid) {
...
 
-        /* Wait for jobs */
-        for (int j = 0; j < 1000000; j++) {
-            jobs_to_process = IOJobQueue_availableJobs(jq);
-            if (jobs_to_process) break;
+        prev_work_start_time = work_start_time;
+        work_start_time = getMonotonicUs();
+        if (prev_work_start_time != 0) {
+            atomic_fetch_add_explicit(&used_active_time_io_thread[id], (work_start_time - prev_work_start_time), memory_order_relaxed);
+        }
```
zuiderkwast: (Updated) suggestion about how to exclude time the thread is inactivated and hangs waiting for mutex.

* No jobs to process in the last cycle means no active work was done.
* prev_work_start_time can be a local variable inside the loop. It's only used very temporarily.
* Line length.

```suggestion
        monotime prev_work_start_time = work_start_time;
        work_start_time = getMonotonicUs();
        if (jobs_to_process != 0) {
            atomic_fetch_add_explicit(&used_active_time_io_thread[id],
                                      work_start_time - prev_work_start_time,
                                      memory_order_relaxed);
        }
```

tests/unit/io-threads.tcl
```diff
@@ -0,0 +1,66 @@
+
+start_server {config "minimal.conf" tags {"external:skip"} overrides {enable-debug-command {yes} io-threads 5}} {
+    set server_pid [s process_id]
+    # Skip if non io-threads mode - as it is relevant only for io-threads mode
+    assert_equal {io-threads 5} [r config get io-threads]
+    test {Force prefetching via IO threads and assert active io_thread usage} {
+        # Create 16 (prefetch batch size) + 1 clients
```
zuiderkwast: Minor naming.

IO threads are mainly for IO. Memory prefetching is another optimization that's running on the main thread. That's not what this test case needs to test.

```suggestion
    test {Force the use of IO threads and assert active IO thread usage} {
        # Create 16 clients
```

tests/unit/io-threads.tcl
```diff
@@ -0,0 +1,66 @@
...
+        } else {
+            fail "Failed to wait until no io_threads are active"
+        }
+        set info [r info]
+        set io_threads_count [dict get [r config get io-threads] io-threads]
+        array set initial_active_times {}
+        for {set i 1} {$i <= $io_threads_count} {incr i} {
+            set used_active_time [getInfoProperty $info used_active_time_io_thread_$i]
+            if {$i < $io_threads_count} {
+                assert {$used_active_time > 0}
```
zuiderkwast:
```suggestion
                assert_morethan $used_active_time 0
```

tests/unit/io-threads.tcl
```diff
@@ -0,0 +1,66 @@
...
+                set initial_active_times($i) $used_active_time
+            } else {
+                assert {$used_active_time eq ""}
```
zuiderkwast:
```suggestion
                assert_equal $used_active_time ""
```

tests/unit/io-threads.tcl
```diff
@@ -0,0 +1,66 @@
...
+        # Adjust io-threads to a lower value and assert that active io_threads fields are >= values found initially
+        assert_equal {OK} [r config set io-threads 1]
+        set info [r info]
+        wait_for_condition 1000 50 {
+            [getInfoProperty [r info server] io_threads_active] eq 0
+        } else {
+            fail "Failed to wait until no io_threads are active"
+        }
+        set used_active_time_1 [getInfoProperty $info used_active_time_io_thread_1]
+        assert {$used_active_time_1 eq ""}
```
zuiderkwast:
```suggestion
        assert_equal $used_active_time_1 ""
```

tests/unit/io-threads.tcl
```diff
@@ -0,0 +1,66 @@
...
+
+        # Re-adjust io-threads to a higher value and assert that active io_threads field values aren't reset to 0.
+        assert_equal {OK} [r config set io-threads 5]
+        set info [r info]
+        for {set i 1} {$i <= $io_threads_count} {incr i} {
+            set used_active_time [getInfoProperty $info used_active_time_io_thread_$i]
+            if {$i < $io_threads_count} {
+                assert {$used_active_time >= $initial_active_times($i)}
```
zuiderkwast: The active_time should not increase when the IO threads are not running. We should make sure that it does't and also check it here in the test case using `assert_equal`.

```suggestion
                assert_equal $used_active_time $initial_active_times($i)
```

zuiderkwast (commented):
I was going to merge this and just put some minor comments, but I realize now that we don't stop the counter when an IO thread is inactivated. We need to fix that.

While the thread is inactivated, it hangs in `pthread_mutex_lock`, waiting for the mutex. We need to exclude that from the active time. I don't have an exact idea about how we can do it.

(after thinking)

1. ~~Maybe it's enough to just set `work_start_time = getMonotonicUs();` after `pthread_mutex_unlock`.~~ Idea 2 is better.

2. Another idea, if we want to make sure the clock doesn't tick at all when the thread has no work, we can add a condition on the number of jobs where we increment the active time, like

```C
        if (jobs_to_process != 0) {
            atomic_fetch_add_explicit(&used_active_time_io_thread[id],
                                      work_start_time - prev_work_start_time,
                                      memory_order_relaxed);
        }
```

Intuitively, I think this seems better and Ithen we can also simplify some other code. We don't need to reset the work_start_time in multiple places, for example.

src/io_threads.c
```diff
@@ -224,14 +225,26 @@ static void *IOThreadMain(void *myid) {
     thread_id = (int)id;
     size_t jobs_to_process = 0;
     IOJobQueue *jq = &io_jobs[id];
+    monotime work_start_time = 0;
+    monotime prev_work_start_time = 0;
```
zuiderkwast: This one can be local inside the while loop.

```suggestion
```

deepakrn: Makes sense.

tests/unit/io-threads.tcl
deepakrn: Great catch @zuiderkwast!

I realized that the active_time gets incremented only if the thread gets reactivated again. In order to catch this bug, I made some changes to the test:
1. activate io_threads
2. assert > 0 thread usage
3. Resize and reset thread counts and assert active usage metric isn't reset to 0
4. Sleep for a time duration >> than expected active usage time
5. Reactivate iothreads
6. Assert total active usage and delta are both less than the sleep time.

src/io_threads.c
deepakrn: Great catch! Updated the code as per your suggestion.

tests/unit/io-threads.tcl
```diff
@@ -1,42 +1,56 @@
+proc activate_io_threads_and_wait {server_pid} {
+    # Create 16 (prefetch batch size) + 1 clients
+    for {set i 0} {$i < 16} {incr i} {
+        set rd$i [valkey_deferring_client]
+    }
+    # set a key that will be later be prefetch
```
zuiderkwast: Minor suggestions:

* Using variables inside other variable names is a bit too much meta programming. Let's use `array set` instead for these clients.

* It's not possible to pass another server pid to get this function to operate on another server. Therefore, I think `server_pid` shouldn't be a parameter. We can just fetch it inside the function using `set server_pid [s process_id]` instead.

* The code comments here needs some changes too. This code creates 16 clients, not 16 + 1, and prefetching has nothing to do with it. We just want to keep the IO threads busy with IO.

zuiderkwast (approved):
Yeah, the behavior seems very good now. It was a long way to get here. Thanks for your patience!

I have only a few minor style suggestions for the test code.

tests/unit/io-threads.tcl
```diff
@@ -0,0 +1,102 @@
...
+                # Assert active thread usage isn't reset to 0.
+                assert_morethan $used_active_time 0
+            } else {
+                assert_equal $used_active_time {}
+            }
+        }
+
+        # Sleep for a time duration that is significantly longer than how much
+        # time each of the io_threads would be active again when reactivated.
+        set sleep_time_ms 4000
```
zuiderkwast: Do we need 4 seconds here or can we reduce it to 1 second? The tests already take a long time to run and we don't want to increase it too much.

```suggestion
        set sleep_time_ms 1000
```

PR #2465, opened 2025-08-11
Don't allow resize hashtable if rehashing is ongoing

enjoy-binbin:
Similar to dicts, we disallow resizing while the hashtable is
rehashing. In the previous code, if a resize was triggered during
rehashing, like if the rehashing wasn't fast enough, we would do
a while loop until the rehashing was complete, which could be a
potential issue when doing resize.

### DIFF
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -1250,11 +1250,11 @@ bool hashtableExpand(hashtable *ht, size_t size) {
     return expand(ht, size, NULL);
 }
 
-/* Returns false if expand failed due to memory allocation failure. */
+/* Returns true if expand was performed or if expand is not needed. Returns false if
+ * expand failed due to memory allocation failure. */
 bool hashtableTryExpand(hashtable *ht, size_t size) {
     int malloc_failed = 0;
-    expand(ht, size, &malloc_failed);
-    return malloc_failed ? false : true;
+    return expand(ht, size, &malloc_failed) || !malloc_failed;
 }
 
 /* Expanding is done automatically on insertion, but less eagerly if resize
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -617,7 +617,7 @@ static bool resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
         return false;
     }
 
-    signed char old_exp = ht->bucket_exp[hashtableIsRehashing(ht) ? 1 : 0];
+    signed char old_exp = ht->bucket_exp[0];
     size_t alloc_size = num_buckets * sizeof(bucket);
     if (exp == old_exp) {
         /* Can't resize to same size. */
--- a/src/hashtable.c
+++ b/src/hashtable.c
@@ -601,6 +601,9 @@ static inline void rehashStepOnWriteIfNeeded(hashtable *ht) {
 static bool resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
     if (malloc_failed) *malloc_failed = 0;
 
+    /* We can't resize twice if rehashing is ongoing. */
+    assert(!hashtableIsRehashing(ht));
+
     /* Adjust minimum size. We don't resize to zero currently. */
     if (min_capacity == 0) min_capacity = 1;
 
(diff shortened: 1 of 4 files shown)


src/hashtable.c
```diff
@@ -1259,17 +1250,22 @@ bool hashtableExpand(hashtable *ht, size_t size) {
...
 
-/* Returns true if expand was performed or if expand is not needed. Returns false if
- * expand failed due to memory allocation failure. */
+/* Returns false if expand failed due to memory allocation failure. */
 bool hashtableTryExpand(hashtable *ht, size_t size) {
     int malloc_failed = 0;
-    return expand(ht, size, &malloc_failed) || !malloc_failed;
+    expand(ht, size, &malloc_failed);
+    return malloc_failed ? false : true;
 }
```
zuiderkwast: Why delete the documentation about what return true means? I think we should keep it.

Why change the logic? It is equivalent to the existing code, but is it simpler? I'm not sure. You ignore the return value of `expand`, which I think is less clear. I think we can return true if expand returns true and only look at `malloc_failed` if expand returned false.

If you really want to change this, then maybe like this?

```c
    if (expand(ht, size, &malloc_failed)) return true;
    return !malloc_failed;
```

src/hashtable.c
```diff
@@ -601,6 +601,9 @@ static inline void rehashStepOnWriteIfNeeded(hashtable *ht) {
 static bool resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
     if (malloc_failed) *malloc_failed = 0;
 
+    /* We can't resize twice if rehashing is ongoing. */
+    assert(!hashtableIsRehashing(ht));
```
zuiderkwast, line 606: We can mention in the comment above the function something like *Must not be called if rehashing is already ongoing*

zuiderkwast (commented):
Good, just some nits.

> if the rehashing wasn't fast enough, we would do
a while loop until the rehashing was complete, which could be a
potential issue when doing resize

The potential issue is a latency spike?

zuiderkwast:
@Fusl I know you have been in this code. Any concerns about this change?

Fusl:
> which could be a
> potential issue when doing resize.

Has this been an issue? The tight while loop to process the entire hash table at once only triggers in extremely rare cases. Typically the resize operation completes _well_ before another resize operation is kicked off. Can you provide a normal use-case example where this is problematic?

enjoy-binbin:
> Has this been an issue? The tight while loop to process the entire hash table at once only triggers in extremely rare cases. Typically the resize operation completes well before another resize operation is kicked off. Can you provide a normal use-case example where this is problematic?

We have a mechanism internally that may pause the lazy rehashing (and only do it in active rehashing), so this might be a problem. In other words, even if we don't actually encounter any problems at the moment, I think we are safer this way inside hashtable.c.

src/hashtable.c
enjoy-binbin: ok you are right, i must be drunk (this return value somehow make me weak..)

zuiderkwast: :beers:

zuiderkwast (approved):

src/hashtable.c
```diff
@@ -620,17 +624,6 @@ static bool resize(hashtable *ht, size_t min_capacity, int *malloc_failed) {
         return false;
     }
 
-    /* We can't resize if rehashing is already ongoing. Fast-forward ongoing
-     * rehashing before we continue. This can happen only in exceptional
-     * scenarios, such as when many insertions are made while rehashing is
-     * paused. */
-    if (hashtableIsRehashing(ht)) {
```
Fusl, line 627: IMO This should be replaced with
```c
if (hashtableIsRehashing(ht)) {
    return false;
}
```
and all other changes in this file should be reverted. We shouldn't be checking for `hashtableIsRehashing()` in all the other places and that makes `resize()` a reliable no-op in case of mis-use rather than triggering the `assert(!hashtableIsRehashing(ht));` assert.

Fusl (changes requested):

src/hashtable.c
enjoy-binbin: that is a good point, i did this at the very first begining, and then i changed it to this way just like we did in dict.c. I actually dont have the preference, doing this inside or outside (the caller) both ok to me as long as we have a clear API. @zuiderkwast which one do you perfer?

enjoy-binbin: todo: line 616 (`signed char old_exp = ht->bucket_exp[hashtableIsRehashing(ht) ? 1 : 0]` ) also need to be updated if we accept this change

zuiderkwast: Fusl's idea is a smaller code change and smaller code in total.

Binbin's idea does an early return in the hot code path. Maybe it is faster? We can avoid calculating the fill factor when we insert while rehashing is ongoing.

Anyway, I think the difference is very small. I don't care which way we do this.

enjoy-binbin: ok, i am going to keep the current one since it is similar to the old dict.c

enjoy-binbin: about the mis-use usage, people should be able to read the comment and add it's own check.

zuiderkwast: resize() is a static function, right? If it's misused, it's only from within the same file, so I think it's safe.

zuiderkwast: @Fusl thanks for your opinion. You don't get in your way this time though. :)

PR #2466, opened 2025-08-11
Fix pre-size hashtables per slot when reading RDB files

enjoy-binbin:
When reading RDB files with information about the number of keys per cluster slot, we need to create the hashtables if they don't exist.

Currently, when processing RDB slot-info, our expand has no effect because the hashtable does not exist (we initialize it only when we need it).

```
if (server.cluster_enabled) {
    /* In cluster mode we resize individual slot specific dictionaries based on the number of keys that
     * slot holds. */
    kvstoreHashtableExpand(db->keys, slot_id, slot_size);
    kvstoreHashtableExpand(db->expires, slot_id, expires_slot_size);
    should_expand_db = 0;
}
```

We also update kvstoreExpand to use the kvstoreHashtableExpand to make
sure there is only one code path. Also see #1199 for more details.


enjoy-binbin:
This seem broken at the very beginning.

src/kvstore.c
```diff
@@ -729,11 +733,15 @@ unsigned int kvstoreHashtableSampleEntries(kvstore *kvs, int didx, void **dst, u
 }
 
 bool kvstoreHashtableExpand(kvstore *kvs, int didx, unsigned long size) {
-    hashtable *ht = kvstoreGetHashtable(kvs, didx);
-    if (!ht) return false;
+    hashtable *ht = createHashtableIfNeeded(kvs, didx);
```
zuiderkwast, line 817: Shall we skip it if `size == 0`?

```suggestion
    if (size == 0) return false;
    hashtable *ht = createHashtableIfNeeded(kvs, didx);
```

src/kvstore.c
```diff
@@ -729,11 +733,15 @@ unsigned int kvstoreHashtableSampleEntries(kvstore *kvs, int didx, void **dst, u
...
     return hashtableExpand(ht, size);
 }
 
+bool kvstoreHashtableTryExpand(kvstore *kvs, int didx, unsigned long size) {
+    hashtable *ht = createHashtableIfNeeded(kvs, didx);
```
zuiderkwast, line 823: Skip if `size == 0`?

```suggestion
    if (size == 0) return false;
    hashtable *ht = createHashtableIfNeeded(kvs, didx);
```

zuiderkwast (approved):
Oops, #1199 was broken from the beginning. :) @naglera @hpatro PTAL.

src/kvstore.c
enjoy-binbin: Do you think we need to judge internally or externally? I am worried about its return value, such as the following code, we return true when newsize == 0, if kvstoreHashtableTryExpand return false when newsize is 0, they're both a bit strange to me. i think i will add they both (internally and externally)
```
bool kvstoreExpand(kvstore *kvs, uint64_t newsize, int try_expand, kvstoreExpandShouldSkipHashtableIndex *skip_cb) {
    if (newsize == 0) return true;
    for (int i = 0; i < kvs->num_hashtables; i++) {
        if (skip_cb && skip_cb(i)) continue;
        if (try_expand) {
            if (!kvstoreHashtableTryExpand(kvs, i, newsize)) return false;
        } else {
            kvstoreHashtableExpand(kvs, i, newsize);
        }
    }

    return true;
}
```

zuiderkwast: Good point. Only externally is OK too.

hpatro: Is this incorrect API usage altogether? Should we crash if the expand request is for `0` ?

src/kvstore.c
hpatro: Is this incorrect API usage altogether? Should we crash if the expand request is for `0` ?

zuiderkwast: We can decide what's correct usage.

I think we can let this function accept 0 and return false. The API is simpler this way. You can call it with any size without exceptions.

zuiderkwast:
Backporting this to 8.0 is non-trivial, because of the dict/hashtable change. I suggest we make a separate PR targeting the 8.0 branch.

enjoy-binbin:
I pushed the commit to the target branch already.

PR #2406, opened 2025-08-01
Attempt to fix flaky testRdb

sarthakaggarwal97:
In the daily runs, I saw macos-build failed for `testrdb.tcl`

```
 [err]: Diskless load swapdb RedisModuleEvent_ReplAsyncLoad handling: during loading, can keep module variable same as before in tests/unit/moduleapi/testrdb.tcl
Module didn't receive or react to REDISMODULE_SUBEVENT_REPL_ASYNC_LOAD_STARTED
[err]: Diskless load swapdb RedisModuleEvent_ReplAsyncLoad handling: when loading aborted, can keep module variable same as before in tests/unit/moduleapi/testrdb.tcl
Expected '1010' to be equal to '200' (context: type source line 272 file /Users/runner/work/valkey/valkey/tests/unit/moduleapi/testrdb.tcl cmd {assert_equal [$replica dbsize] 200} proc ::test)
```

Link: https://github.com/valkey-io/valkey/actions/runs/16610259318/job/46991731591

I could pinpoint that the `wait_for_condition` would have timed out for the error. I was however unable to reproduce in the local mac, or [github macos runners](https://github.com/sarthakaggarwal97/valkey/actions/runs/16659569003/attempts/1) in my repo. I do feel we can increase the number of retries here and observe if it stabilizes the daily test. 

Please do let me know if this doesn't look like a valid fix or if there are any other ideas.

PR #2468, opened 2025-08-11
Adding testType to testRdb

sarthakaggarwal97:
> @ranshid the same test runs with different test cases, and it failed for one of them.

Ah... we should probably add the testType in the test name to make it more clear then

_Originally posted by @ranshid in https://github.com/valkey-io/valkey/issues/2406#issuecomment-3172489117_

            

```
[ok]: Diskless load swapdb RedisModuleEvent_ReplAsyncLoad handling (Successful): after db loaded, can set module variable with new value (1 ms)
[ok]: Check for memory leaks (pid 81673) (1426 ms)
[ok]: Check for memory leaks (pid 81662) (1327 ms)
=== (external:skip) Starting server on 127.0.0.1:21135 ok
=== () Starting server on 127.0.0.1:21136 ok
[ok]: Diskless load swapdb RedisModuleEvent_ReplAsyncLoad handling (Aborted): during loading, can keep module variable same as before (939 ms)
[ok]: Diskless load swapdb RedisModuleEvent_ReplAsyncLoad handling (Aborted): when loading aborted, can keep module variable same as before (1 ms)
```

The testType is visible now.

### DIFF
--- a/tests/unit/moduleapi/testrdb.tcl
+++ b/tests/unit/moduleapi/testrdb.tcl
@@ -242,7 +242,7 @@ tags "modules" {
                             # 10ms per key, with 1000 keys is 10 seconds
                             $master config set rdb-key-save-delay 10000
 
-                            test {Diskless load swapdb RedisModuleEvent_ReplAsyncLoad handling: during loading, can keep module variable same as before} {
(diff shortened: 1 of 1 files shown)


enjoy-binbin (commented):
they are both under the switch condition (different conditions), so we can know it testType actually?

sarthakaggarwal97:
@enjoy-binbin good point. I think I missed this detail. We don't need this change, closing it out. Thank you.

Issue #1483, opened 2024-12-24
Multithreaded BGSave


madolson:
One of the bottlenecks that we often observe in AWS is the time it takes to do a full sync either when replacing a node or bootstrapping a new nodes into a cluster. We can partially solve this bottleneck by supporting the usage of additional threads for a multi-threaded save and restore. The save can be trivially parallelized by either having each thread own a subset of the slots (for cluster mode) or a subset of the hash table buckets (for standalone). The restore is a bit less trivial, but by properly setting the initial state of the dictionary, we should be able to safely load data into it from multiple threads. 

Each of these threads can use its own TCP connection for transmitting the full-sync data. The primary and replica will need to negotiate the number of RDB full sync connections to use.

ranshid:
>Each of these threads can use its own TCP connection for transmitting the full-sync data. The primary and replica will need to negotiate the number of RDB full sync connections to use.

Does your suggestion here implies that the feature will only work for rd-channel enabled primary/replica?

I think we can also look at some ways we can use a single multiplexing connection to pass all traffic on a single stream. For example QUIC can help logically separate the data sent by each thread OR we can also logically multiplex the slot/bucket data using a dedicated headers which will proceed each data chunk (like crc is).

madolson:
> I think we can also look at some ways we can use a single multiplexing connection to pass all traffic on a single stream. For example QUIC can help logically separate the data sent by each thread OR we can also logically multiplex the slot/bucket data using a dedicated headers which will proceed each data chunk (like crc is).

I suppose this is related, https://github.com/valkey-io/valkey/issues/1300.

PingXie:
I like this idea!

By the way, what you described in this issue is the diskless full sync. I think there's also value in speeding up the RDB snapshot (to disk). We'll likely need to use a "scatter-gather" solution that writes different parts of the keyspace in parallel to separate files and then concatenates them to form the final RDB file.

xbasel:
Currently, BGSAVE is not the bottleneck. The limiting factor is on the replica side during RDB processing. There's ongoing work to improve replica-side RDB handling (I'll open an issue and update here). Until then, there's little benefit in optimizing BGSAVE further. (there's still benefit when saving to file though)

madolson:
[...]

Issue #2376, opened 2025-07-23
Consolidate IO Threads and BIO Threads


nadav-levanoni:
Today we have two threading systems that offload work from the main thread - IO threads and BIO threads. IO Threads allow for generic scheduling, and are actively being maintained and improved by contributors. BIO threads have a poor hard-coded design that requires a thread per job, and have not been directly improved in years. There is an opportunity to consolidate BIO threads into IO threads. The high level benefits include: Simplifying the code base, unified/simpler threading configurations, improved maintainability & design, and better resource utilization.

Extending IO threads to handle lazyfree has been a part of IO Threads backlog for about a year now: https://github.com/valkey-io/valkey/issues/761

The design issues with BIO threads have been previously addressed by this unmerged PR: https://github.com/valkey-io/valkey/pull/356
BJM allows for generic, thread-safe, background job scheduling. This is still a viable path forward because it addresses the major design flaws of BIO threads, however, it doesn’t address the issue of maintaining two threading systems.

---

Looking forward, major considerations are:

* IO job priority/types. The jobs originally completed by BIO threads should be handled differently than current IO work. Main differentiating factors are, BIO jobs:
    * Cannot be rejected by the IO threads during enqueuing. A second job queue may be required to support this.
    * Lazy jobs have lower priority/urgency to be completed. They should not be prioritized over more important jobs.
    * Must be time bound to avoid starvation of other jobs on the thread. This may require adding logic that allows us to delete large data structures over the span of several cycles.
* Consolidating configs - This change will deprecate/update BIO configs and info metrics.
* Large code diff. The change would involve completely removing bio.c/h and refactoring parts of lazyfree.c (I would consider rebranding to something like lazyjobs.c to make it more generic).

---

This is a high-level overview. If anyone has any ideas topics that I missed please add them.

PR #356, opened 2024-04-23
Background Job Manager (BJM) - replacement for BIO

JimB123:
The BIO module is a source of minor annoyance. Background Job Manager (BJM) is designed as a modular replacement for BIO. Assuming there is positive response for this PR, the PR will be extended to convert existing BIO usage to BJM and remove BIO.

Motivations:
* BIO (Background I/O) is a misnomer. I'd guess that this was originally a mechanism to perform an async flush on the AOF file, but it has since morphed into a mechanism for performing other kinds of background activities - like lazy flush - which have nothing to do with I/O
* BIO isn't well layered. BIO should be a low-level utility for handling async background tasks. However the design of this utility requires that the "low-level" code have specific knowledge of the "high-level" application.
* BIO isn't very modular. For each new type of background task, the BIO code must be altered. A new thread is created for each type of task. New definitions and logic are added for each new task.
* Improve performance. BIO creates multiple threads, but doesn't use them effectively. There is 1 unique thread for each type of task. During a burst of lazy evictions, only the (single) eviction thread will be active while other BIO threads sit idle. During a lazy flush operation, 2 dictionaries (main and expire) are processed sequentially rather than in parallel.
* Reduced memory usage. BIO uses a list to maintain the queue of jobs. The new Fifo uses much less memory (and is more performant) if there's a large burst of jobs.

PR Contents:
* Fifo - presents a simple FIFO queue, it is over 50% more space efficient and over 50% faster than using Redis list. It's delivered as an independent, modular, & reusable utility data structure.
* BJM provides a simple fire-and-forget interface to have something done by a background thread. It maintains a fixed/configurable quantity of threads (rather than one per "job type"). This prevents having too many active background threads and also allows faster processing for a large number of jobs of the same job type.

Originally proposed to Redis and stalled here: https://github.com/redis/redis/pull/12029

Note especially that in the original proposal, there was some concern about WAITAOF which currently relies on BIO to serialize 2 independent background jobs.  There were 3 possible solutions mentioned:
* Retain BIO for this case
* Modify the WAITAOF code so that it enqueues a single sequenced job rather than 2 jobs
* Enhance the proposed BJM mechanism with "futures" to provide a mechanism for synchronization

PR #2470, opened 2025-08-11
Multi threaded RDB Save

Nicky-2000:
# Multi-threaded RDB Save

### Contents
1. Introduction 
2. Performance Improvement using RDB Threads
3. Profiling Current, Single-Threaded, Implementation
4. Detailed Design
5. Testing
6. RDB Threads Configuration
7. Future Extension

---

### Introduction:
This PR introduces a multi-threaded RDB Save, which aims to improve Valkey's persistence performance. There will be a future PR that introduces multithread RDB Load.

**Motivation for PR:**
This PR addresses Issue: https://github.com/valkey-io/valkey/issues/1483

Currently, RDB operations, including snapshots and full synchronizations, are single-threaded and often take a very long time (especially with large datasets). This poor performance can lead to Out-of-Memory (OOM) errors due to Copy-On-Write (COW) overhead, long recovery times, reduced availability and failures during cluster scaling events. Host machines running Valkey often have underutilized CPU cores, and since the data encoding and compression work during RDB saves is highly repetitive, it's an ideal situation for to use threading to improve performance.

TODO: 
1. Remove min and max value from rdb thread flush buffer threshold as per Jun's suggestion. 
---

# Performance Improvements
### Test Environment:
Performance tests were run on a Google Cloud n2-highmem-48 VM with the following hardware configuration:
- **Machine Type:** n2-highmem-48
- **CPU:** 48 vCPUs (24 physical cores) on an Intel Cascade Lake platform.
- **Memory:** 384 GB of RAM.
- **Storage:** A persistent 1 TB SSD was attached for RDB file saves.
- **Write Bandwidth (Important)**: ~ 760 MB/s

 
### Workloads Tested
It is important to note that the test data (keys and values) used for these benchmarks was generated with valkey-benchmark and is entirely non-compressible.
| Number of Keys | Value Size | Notes |
|:---:|:---:|:---|
| 2 Billion | 50 B | Small Value Size Workload. |
| 1 Billion | 300 B | Small Value Size Workload. |
| 500 Million | 500 B | Small Value Size Workload. |
| 50 Million | 5 KB | Medium Value Size Workload. |
| 10 Million | 20 KB | Medium Value Size Workload. |
| 5 Million | 64 KB | Slightly smaller than 64KiB buffer write threshold. |
| 5 Million | 66 KB | Slightly larger than 64KiB buffer write threshold. |
| 1 Million | 100 KB | Large Value Size Workload. |
| 0.5 Million | 262 KB | Slightly smaller than 256KiB max buffer capacity threshold.|
| 0.5 Million | 272 KB | Slightly larger than 256KiB max buffer capacity threshold.|
| 0.1 Million | 1000 KB | Large Value Size Workload. |

All tests were run using the `run_full_save_benchmark.sh` test script located in the [valkey-rdb-benchmarking]( https://github.com/Nicky-2000/valkey-rdb-benchmarking). This bash script is a wrapper of [save_benchmark.py](https://github.com/Nicky-2000/valkey-rdb-benchmarking/blob/main/scripts/save_benchmark.py) python script with different arguments passed in (different workload configurations).

The code for the graphs and the raw Data can be found here: [Visualization Code and Raw Data Notebook]( https://github.com/Nicky-2000/valkey-rdb-benchmarking/blob/main/notebooks/large_tests_64k_buffer/performance_improvement_vs_value_size.ipynb)

### Key Metrics:
Speed up (num_threads) = (Single Thread Save Time) / (Save Time with 'num_threads')
Efficiency (num_threads) = (Speedup with 'num_threads') / ('num_threads')
   
### Results:
Multithreading significantly speeds up the runtime of RDB Save, especially for workloads with "small" keys/values or if compression is on. For large keys, and compression off, there are limited improvements. For the 66KB Value Size Workload there was a 7% Regression when the number of threads is larger than 3.

##### Compression OFF
With compression off, we observed a substantial speedup for small value size workloads. A 50 B value size workload showed a speedup of over 3x with 4 threads. Workloads with value sizes of 300 B and 500 B achieved a speedup of 3x to 3.2x.

Screenshot 2025-08-08 1 58 23 PM

##### Compression ON
With compression on, the performance gains are even larger. For a very small value size (50 B), we achieved up to a 10x speedup with 10 threads. For value sizes in the 300-500 B range, a 4.5x speedup was observed.
Screenshot 2025-08-08 2 01 30 PM

##### Compression OFF
For medium-sized keys with compression off, the improvement was minimal, ranging from 1x to 1.2x speedup. This is because the keys are large enough to cause the operating system to buffer them effectively, leading to disk I/O throughput becoming the primary bottleneck.
Screenshot 2025-08-08 1 59 50 PM

##### Compression ON
Enabling compression for medium value size workloads yielded a substantial speedup of 3.5x to 4x.
Screenshot 2025-08-08 2 02 02 PM

##### Compression OFF
Similar to medium value size workloads, we observed limited improvement of 1x to 1.2x speedup for large value size workloads with compression off. The system remains bottlenecked by disk I/O.
Screenshot 2025-08-08 2 00 30 PM

##### Compression ON
With compression on, large value size workloads demonstrated a significant speedup of approximately 4x.
Screenshot 2025-08-08 2 02 44 PM

### Raw Runtime comparison with Compression On and Off
It is important to note that the values used in our tests were generated with valkey-benchmark and were **_entirely non-compressible_**. This resulted in the RDB files for both compression ON and OFF being identical in size.

Therefore, the multi-threaded speedup shown in these results is due solely to the effect of parallelizing the CPU intensive operations , without any additional benefit from a smaller file size. If the data had been compressible, we would expect the compression ON runs to be even faster due to less bytes needing to be written to persistent storage. This provides a strong motivation to evaluate enabling RDB compression in environments using multi-threaded RDB save. Furthermore, because we can distribute the compression workload across multiple threads, there could be significant value in using a more powerful compression algorithm (with a higher compression ratio). This would not only makes the saved file sizes smaller but could also reduce the overall save runtime, since fewer bytes will need to be written to storage. See the "Testing on More Complex, Real World Workloads" and "Performance and Compression Ratio Relationship" sections for more details on this.
  
#### Improvement on Small  Value Size Workloads
Screenshot 2025-08-08 2 03 59 PM

#### Improvement on Medium Key Size Workloads
Screenshot 2025-08-08 2 04 41 PM

#### Improvement on Large Value Size Workloads
Screenshot 2025-08-08 2 05 12 PM

### Testing on More Complex, Real World Workloads: 
The performance tests described above used uncompressible string values of different lengths. Therefore, there was no change in the generated RDB file size if compression was on.

Valkey, however, supports more complex data types such as lists, sets, and JSON. Performance was also evaluated using these types, and a general trend emerged: the more complex the data type, the greater the potential performance improvement.

For example, the figure below shows the performance gains when saving a database filled with large, complex JSON objects. While the specific speedup depends on the nature of the data, the testing shows it's possible to achieve a 20x+ speedup when compression is enabled, and a 4x to 10x speedup with compression off. Depending on the workload the speedup could be even greater.

The JSON Generation function used to create values for this workload [can be found here](https://github.com/Nicky-2000/valkey-rdb-benchmarking/blob/d92deab821bb86240aa77b7e19ddfbb5920f58df/utilities/key_value_generation_utilities.py#L141).

Screenshot 2025-08-28 3 54 20 PM

### Performance and Compression Ratio Relationship
The performance test on the complex JSON (above) showed that the ratio between the fastest save time with compression off and the fastest save time with compression on is approximately equal to the compression ratio (the original file size divided by the compressed file size)!

This strongly motivates the value of introducing a better-performing compression algorithm (with a  higher compression ratio) into Valkey as it could lead to faster overall RDB Save times (and Load).

### Important Note on Regression:
The multi-threaded implementation showed a performance regression when processing key-value pairs with a 66KB value size. This was the only workload where the multi-threaded approach was slower than the single-threaded version.

The cause of this regression is the interaction between the multi-threaded write buffer and the data size. The implementation uses a write buffer with a 64KB threshold. Since the 66KB value size causes the buffer to exceed this limit, a write operation is immediately triggered for each key/value pair. This means we are writing to the disk after buffering exactly one key-value pair. Therefore we lose much of the performance benefits of multi-threading. Instead, for this workload, adding additional threads create a performance penalty due to coordination overhead. To help alleviate this issue you can change the buffer write threshold. See the "RDB File Writing Mechanism" section for more details on this buffer. 

The regression becomes noticeable at 4 threads and worsens as more threads are added, reaching a peak of a 7% slower runtime than the single-threaded baseline when running with 8 threads..
Screenshot 2025-08-08 2 06 46 PM

---

# Profiling Current, Single-Threaded, Implementation
To inform the design of RDB Save multi-thread, performance profiling was conducted to identify the bottlenecks under different workloads.  

### Profiling Environment
Note: This profiling was done on a different machine than the "Performance Improvement Tests"

- **CPU:** 64 vCPUs (32 physical cores) on an AMD EPYC 7B13 platform.
- **Memory:** 117 GiB of RAM.
- **Storage:** A 500 GiB SSD was attached and mounted for RDB file saves and benchmarking.
- **Write Bandwidth (Important):** ~ 402 MB/s

### Results
Valkey Data Throughput by Value Size and Compression

The graph above shows that for compression OFF we are I/O bound when keys/values are greater than ~5,000 - 10,000 Bytes (for this system). When keys/values are  < ~5000 Bytes then we are CPU Bound . When compression is ON we are always CPU bound. 

This shows that there are significant performance gains to be made by parallelizing the CPU intensive work of encoding (and compressing) the data.  

---

# Detailed Design
### High-Level Architecture
The proposed architecture uses a multi-threaded encode, serial-write model. This design separates the Valkey RDB Save process into two phases:

1. Parallel Encoding: Multiple worker threads concurrently encode Valkey objects and compress the data (if enabled). Each thread has a local buffer to store its encoded data, which avoids contention during this phase. The local buffer is bounded in size to reduce the probability of OOM errors.

2. Serial Writing: Upon filling up its local buffer with data, each thread acquires a shared lock before writing data to a single RDB file. Therefore, only a single thread writes to the file at any point in time. This phase is limited by the disk's I/O throughput and is the new bottleneck after the encoding is parallelized.

### Work Allocation
Work is partitioned at the **Bucket level**, which provides an ideal granularity for division and is compatible with both Cluster and Standalone modes. Note: We partition and process each hashtable. So in cluster mode, when there are multiple slots (I.e hashtables), we will partition and process each of them sequentially.

To handle tables in the middle of a `rehash` operation, threads are assigned a `start_index` and a `stride` to iterate through a flattened view of "Live-Buckets." This design ensures all data is processed correctly while the rehashing is paused.

The diagram below shows how threads will be allocated certain buckets in the the "flattened" (logical) view of the tables[0] and tables[1] bucket arrays. Note: this is a view of the "top level buckets". All child buckets will be processed by the thread assigned to the top level bucket that is the ancestor of the said child bucket. 

RDB Save Multi-Thread Diagrams (31)

The next diagram shows the full flow of how the main thread allocates work for the worker threads for a single hashtable. Note: In the code the "Main Thread" is also responsible for processing a portion of the hashtable. The diagram does not include this for simplicity. 

RDB Save Multi-Thread Diagrams (32)

### RDB File Writing Mechanism

Each worker thread follows a coordinated process:
1.  **Buffer Data**: Threads encode key-value pairs into a local buffer.
2.  **Acquire Write Lock**: When a buffer is full (exceeds a `WORKER_BUFFER_DEFAULT_SIZE` currently set to 64KB), the thread acquires a shared write lock.
3.  **Write**: The thread writes its buffer to the RDB file.
4.  **Release Lock**: The lock is released, and the thread continues processing more keys.

The diagram below shows how each thread processes its allocated buckets, by striding through the hashtable, stores encoded data in a thread local buffer, and dumps data to the rdb file / target Rio. 
RDB Save Multi-Thread Diagrams (33)

### Handling Large Keys/Values
To prevent Out-of-Memory (OOM) errors, a **dynamic buffering strategy with a direct-to-file fallback** is used. If a key-value pair would exceed a hard buffer limit (`WORKER_BUFFER_CAPACITY_LIMIT` currently set to 256KB), the thread flushes its existing buffer and then streams the large key's data directly to the file, bypassing the buffer while holding the write lock. This ensures that each thread is allocated a maximum of 256KB of memory. 

### Threading Model
We chose to abstract the `JobQueue` implementation into `thread_common.c` and create a new set of dedicated functions for RDB thread initialization and management located in `rdb_threads.c`. This approach avoids tight coupling with the existing I/O and BIO threads, allowing for independent control over the number of threads, thread local storage, thread CPU affinity, and job queue sizes.

**Other Options Considered:**
* **I/O Threads**: Rejected due to tight coupling, potential for CPU contention during BGSAVE, and inflexible job queues.
* **BIO Threads**: Rejected because the model's fixed number of threads and "fire-and-forget" mentality are incompatible with the active coordination and variable thread count required for RDB operations.

---

# Testing 
### Unit Tests 
New unit tests were added for the `hashtableStrideNext` function in `test_hashtable.c`. These tests ensure the correct, complete, traversal of the hashtable occurs in the multi-threaded context.

### TCL Integration Tests
A new TCL test configuration flag (`rdb-threads`) was added. This flag configures the Valkey servers to run with rdb-threads active.

The full Valkey TCL test suites were run with the `rdb-threads` flag enabled. All existing tests passed, confirming that this implementation is backward-compatible and introduces no regressions or unintended changes to the RDB format or replication protocol.

Testing command used: ` ./runtest --large-memory --rdb-threads`

I also modified integration/rdb.tcl, integration/replication-2.tcl and integration/replication-3.tcl to run in a loop with different number of rdb-threads enabled.

I am open to feedback on this testing strategy. We could add a "daily" job (like there is for IO threads) that will run the test suite with the --rdb-threads flag. 

---

# RDB Threads Configuration
This PR introduces a new configuration directive, `rdb-threads`, to control the number of threads used for RDB saving.

This directive specifies the number of parallel threads to use for encoding and compressing data during an RDB save operation. The default value is `1`, which reverts to the existing single-threaded behavior.

`rdb-threads <num_threads>`
-  `num_threads`: An integer value in the range 1 to 256
- Behavior: If `num_threads` is set to `N`, the main thread will function as one of the save threads. The system will then spawn `N-1` additional worker threads. For example, if `rdb-threads` is set to `5`, a total of five threads—the main thread and four worker threads—will be used. This is the same behavior as IO threads.

Using a higher number of threads can significantly improve RDB save performance, especially on multi-core systems. The optimal number of threads often depends on the workload and the number of available CPU cores. Analyzing the **_Performance Improvements_** section can help inform your decision about how many threads to use.

---

# Potential Future Extension: Sharded RDB Files
With this threading model in place, we can transition, with not too much effort, to "sharded" rdb files. I.e. we can pass a different target Rio to each thread (or groups of threads). This would enable us to write to multiple RDB Files or multiple sockets on a replica.

**Benefits to sharded RDB files:**
1. This would enable easier multi-threading of RDB Load. Each thread could be assigned one of the sharded RDB files/streams. 
     - Note: The current RDB version (non-sharded) does enable some parallelization of RDB Load by reading "chunks" of the RDB file and then passing them to threads for decoding and insertion into the database. This version of RDB Load (which is a non-breaking backward compatible change) will be done in a future PR.

2. Sharded RDB files would enable concurrent writing to multiple persistent disks. This would help us overcome the I/O bottleneck by writing sharded RDB files to separate disks simultaneously.

**Issues with sharded RDB files:**
1. This would be a significant breaking change to the RDB file format. And would break backwards compatibility.
2. Increased complexity in handling the sharded RDB files.


src/hashtable.c
```diff
@@ -323,6 +323,13 @@ typedef struct {
     uint16_t table_index;
 } position;
 
+/* Represents a specific bucket's location. Used by hashtableStrideNext. */
+typedef struct BucketLocation {
+    uint8_t table_index;
+    size_t bucket_index;
+    uint8_t valid;
```
luoos: Can we have a comment explaining what `valid` means?

src/hashtable.c
```diff
@@ -2075,6 +2085,188 @@ bool hashtableNext(hashtableIterator *iterator, void **elemptr) {
...
+ *
+ * Iteration Flow for one thread (Each call returns ONE element):
+ *
+ * 1. Initial calls:
+ * - Successive calls return elements from [B0] until exhausted.
+ * - If [B0] is chained, subsequent calls return elements from (C0) until (C0) is exhausted.
+ * (At this point, B0's entire chain is processed.)
+ *
+ * 2. The NEXT call after B0's chain is exhausted:
+ * - STRIIDE! Jumps 2 logical buckets forward.
```
luoos: typo?

src/hashtable.c
```diff
@@ -2075,6 +2085,188 @@ bool hashtableNext(hashtableIterator *iterator, void **elemptr) {
...
+ * - Starts returning elements from [B2].
+ * - If [B2] is chained, subsequent calls return elements from (C2).
+ * (At this point, B2's entire chain is processed.)
+ *
+ * 3. The NEXT call after B2's chain is exhausted:
+ * - STRIIDE! Jumps 2 logical buckets forward.
```
luoos: typo?

src/rdb_threads.c
```diff
@@ -0,0 +1,377 @@
...
+#include "rdb.h"
+
+static pthread_t rdb_threads[RDB_THREADS_MAX_NUM] = {0};
+static pthread_mutex_t rdb_threads_mutex[RDB_THREADS_MAX_NUM];
+JobQueue rdb_jobs[RDB_THREADS_MAX_NUM] = {0}; // Job queues for each RDB worker thread.
+
+
+/* --------- RDB Worker Threads Core Logic --------- */
+
+static void *RDBThreadMain(void *myid) {
```
luoos, line 18: As discussed before, the word `Main` is overloaded. Maybe pick a different word?

src/rdb_threads.h
```diff
@@ -0,0 +1,48 @@
+#ifndef __RDB_THREADS_H__
+#define __RDB_THREADS_H__
+
+#include "server.h"
+#include "thread_common.h"
+
+/* Threshold for flushing a worker's buffer to the main RDB file (64 KB). */
```
luoos: it doesn't have to be file

src/rdb_threads.h
```diff
@@ -0,0 +1,48 @@
...
+#define WORKER_BUFFER_DEFAULT_SIZE 64 * (1024)
+/* Maximum capacity for a worker's buffer. Keys causing this limit to be exceeded are streamed directly to RDB file (256 KB). */
```
luoos: Maybe explain more about the big key scenario? Why do we need two thresholds, and how do they solve the big key case.

src/rio.c
```diff
@@ -57,6 +57,102 @@
...
+        pthread_mutex_lock(r->io.buf_to_target.target_rio_mutex);
+
+        /* Dump existing buffered data to underlying RIO. */
+        if (r->io.buf_to_target.pos > 0) {
+            if (rdbWriteRaw(r->io.buf_to_target.target_rio, r->io.buf_to_target.ptr, r->io.buf_to_target.pos) < 0) {
+                pthread_mutex_unlock(r->io.buf_to_target.target_rio_mutex); /* Release lock on error. */
+                return 0;
+            }
+        }
+        // The caller (rdbEncodeHashtableRange) is responsible for clearing r->io.buf_to_underlying.ptr/pos.
```
luoos, line 92: nit: inconsistent comment style

luoos (commented):
Overall LGTM. Some minor comments. Can we have some dedicated tcl for this feature?

Nicky-2000:
> Overall LGTM. Some minor comments. Can we have some dedicated tcl for this feature?

Yep! I'll add a few tests that don't require the `--rdb-threads` flag in order to test the new functionality.

dvkashapov:
> BIO Threads: Rejected because the model's fixed number of threads and "fire-and-forget" mentality are incompatible with the active coordination and variable thread count required for RDB operations.

I get the fixed num of threads part but can you explain a little what type of coordination is needed for RDB? I've looked a little at the code and do you mean the part where `drainRDBThreadsQueue` waits for all jobs to be finished in `rdbSaveDbMultiThreaded`?

> This approach avoids tight coupling with the existing I/O and BIO threads, allowing for independent control over the number of threads, thread local storage, thread CPU affinity, and job queue sizes.

To avoid introducing third thread type we could make BJM (#356) come to life first or maybe unify thread types (#2376). Once this is done, multithreaded RDB save could be added as a job type. This way we won't have to create 3rd type of threads that needs to be maintained. Because BJM is still in progress, we can modify it a little for the needs of coordination with the main thread. What do you think?

Nicky-2000:
Thanks for the feedback! Here are my thoughts. 

> I get the fixed num of threads part but can you explain a little what type of coordination is needed for RDB? I've looked a little at the code and do you mean the part where `drainRDBThreadsQueue` waits for all jobs to be finished in `rdbSaveDbMultiThreaded`?

You're right about drainRDBThreadsQueue. That's one of the key differences. The RDB saving process can't use a simple "fire-and-forget" model because it requires active coordination.

We use a fork-join pattern: the main thread allocates (forks) work to the workers (like saving slot-1 in cluster mode), and then we "join" them by waiting for them all to finish. We need this coordination to properly propagate errors (one thread fails means everything should fail). It's also necessary so the main thread knows when it's safe to resume hashtable rehashing.

> To avoid introducing third thread type we could make BJM (#356) come to life first or maybe unify thread types (#2376). Once this is done, multithreaded RDB save could be added as a job type. This way we won't have to create 3rd type of threads that needs to be maintained. Because BJM is still in progress, we can modify it a little for the needs of coordination with the main thread. What do you think?

I agree that a unified threading model is a great long-term goal.

However, there are two key differences that make a dedicated approach more pragmatic for now. First, as discussed, RDB requires a coordinated "fork-join" pattern, which is fundamentally different from BIO's "fire-and-forget" model. Second, RDB threads are ephemeral (they only exist for a specific save/load operation), unlike the permanent daemon threads used for BIO.

Designing a generic manager to handle both patterns and lifecycles is a significant architectural undertaking. This PR takes a direct approach and uses the familiar I/O threads pattern, making it maintainable and allowing us to deliver this performance improvement now. Additionally, it doesn't prevent a future migration to a unified system (once it is developed).

src/rdb_threads.c
Nicky-2000: I think its best to leave it like this since IOThreads use this convention with IOThreadMain. However, if you have a strong preference for something like "RDBThreadEntryPoint" I'll change it.

Nicky-2000:
> > Overall LGTM. Some minor comments. Can we have some dedicated tcl for this feature?
> 
> Yep! I'll add a few tests that don't require the `--rdb-threads` flag in order to test the new functionality.

I modified integration/rdb.tcl, integration/replication-2.tcl and integration/replication-3.tcl to run in a loop with different number of rdb-threads enabled.

murphyjacob4:
Hey @Nicky-2000! Hello from the future. So obviously this has been languishing for a while now. There are going to be a lot of merge conflicts. Are you at all interested in fixing it up and getting it reviewed? If not, I can close it out and we can try to reassign to someone else. I do think most of the design work is solid

rainsupreme:
Hi @Nicky-2000 — I'm closing this PR to help keep the queue tidy, since it's been a while since Jacob's message. It can always be reopened or used as reference for a new PR!

I wanted to leave some relatively surface-level notes since this is solid design work and I'd like to capture them regardless of what happens with this specific PR. These would be the things I'd want addressed for merge, and they also serve as a checklist for anyone picking this up in the future:

**What's strong here:**
- The profiling-first approach (identifying the CPU-bound vs I/O-bound crossover at ~5KB value size) is exactly how performance work should be motivated
- Bucket-stride partitioning over the flattened hash tables is elegant and appears to handle mid-rehash correctly
- The bounded thread-local buffer design (256KB cap, 64KB flush threshold) balances memory safety with batching efficiency
- Exploiting idle cores in the forked child process is a clean win with no contention against the parent's IO threads

**What I'd want to see for merge:**
- **Rebase onto current unstable** — the hashtable layer has evolved significantly since Aug 2025 (dict removal, etc.)
- **Regression documentation** — the 66KB-value regression needs to live in `valkey.conf` comments or docs, not just the PR description. Operators need guidance: "value size range → recommended thread count"
- **Dedicated test file** (e.g. `tests/unit/rdb-threads.tcl`) covering:
  - Save during active rehash (stride correctness)
  - Large key exceeding buffer capacity (direct-write fallback path)
  - Error propagation (one thread fails → all abort, no leaked threads)
  - Bit-identical output regardless of thread count (the strongest correctness invariant)
- **Testing seam** — the multi-threaded encode path should be exercisable with a mock rio, independently of full BGSAVE, for unit-level verification
- **Thread unification discussion** (raised by @dvkashapov) — doesn't need to be solved here, but the code should be structured so migration to a unified thread pool (BJM / #2376) isn't painful later

**Broader design questions for whoever carries this forward:**
- **COW quantification** — the biggest operational win may be *reduced peak COW memory* rather than raw speed. A benchmark measuring parent RSS / COW pages under fixed write rate with 1 vs N threads would make the case compelling for operators
- **Interaction with forkless sync** — when forkless snapshot/sync lands, the fork-based COW problem disappears. But the encoding pipeline (parallel encode + serial write) could potentially be reused in that context too. Worth considering.
- **Could the IO thread pool be reused?** — the child process doesn't inherit the parent's threads post-fork, so it creates fresh ones. That's fine, but the question of whether to maintain a separate thread pool type long-term is worth revisiting

Thanks for the thorough design doc and benchmarking — this is the kind of contribution that moves the project forward regardless of the PR's immediate outcome. 🙏

Issue #2219, opened 2025-06-16
[NEW] Benchmarking: Detect if valkey-benchmark or valkey-server is the bottleneck


zuiderkwast:
When running valkey-benchmark, it can happen that valkey-server is not running at full capacity but valkey-benchmark can't send commands faster. The solution to this is usually to run valkey-benchmark with more threads or on a more powerful machine.

### Description of the feature

The user can't easily see which one is the bottleneck of the two. If we can detect it, we could print a note after the benchmark, such as

    Can't push the server to its limits. Consider running valkey-benchmark
    with more threads or on a faster machine.

### How to detect it?

If valkey-benchmark is running at full speed without blocking (writes and reads don't return EAGAIN, epoll doesn't block) then it can mean that valkey-benchmark is the bottleneck.

*Original idea from https://github.com/valkey-io/valkey/pull/1761#issuecomment-2974707119*

hanxizh9910:
I am working on this issue

zuiderkwast:
@hanxizh9910 That's great.

TBH, I don't know if this is easy or even possible to do. It's just an idea. I hope it works though.

hanxizh9910:
@zuiderkwast  I've implemented bottleneck detection using EAGAIN tracking as suggested and created a draft PR https://github.com/valkey-io/valkey/pull/2471. Would appreciate your feedback!

PR #2471, opened 2025-08-11
Show RPS histogram in valkey-benchmark

hanxizh9910:
When benchmarking with the a target thoughput using the `--rps` flag, display an RPS histogram in the benchmark report. This can help identify if there is a bottleneck.

Related to #2219.

### DIFF
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -1171,7 +1171,7 @@ static void showLatencyReport(void) {
             printf("\n  RPS Analysis:\n");
             printf("    Target RPS: %.0f\n", target_rps);
             printf("    Actual RPS: %.2f (%.1f%% of target)\n", reqpersec, achievement_pct);
-            
+
             if (reqpersec < threshold_90) {
                 printf("    Status: SEVERE BOTTLENECK DETECTED\n");
                 printf("    Performance is significantly below target (< 90%%).\n");
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -1140,7 +1140,8 @@ static void showLatencyReport(void) {
         }
         printf("  multi-thread: %s\n", (config.num_threads ? "yes" : "no"));
         if (config.num_threads) printf("  threads: %d\n", config.num_threads);
-
+        /* Show the RPS Report */
+        showRPSReport();
         printf("\n");
         printf("Latency by percentile distribution:\n");
         struct hdr_iter iter;
@@ -1276,9 +1277,7 @@ static void benchmarkSequence(const char *title, char *cmd, int len, int seqlen)
     } else
         startBenchmarkThreads();
     config.totlatency = mstime() - config.start;
-
     showLatencyReport();
-    showRPSReport();
     freeAllClients();
     if (config.threads) freeBenchmarkThreads();
     if (config.current_sec_latency_histogram) hdr_close(config.current_sec_latency_histogram);
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -1188,15 +1188,10 @@ static void showLatencyReport(void) {
                        rps_p50, (rps_p50 / target_rps) * 100);
 
                 /* Bottleneck analysis using P1 (99% of time achieved this or better) */ 
-                if (rps_p1 >= 0.99 * target_rps) {
-                    printf("    Status: GOOD PERFORMANCE\n");
-                    printf("    99%% of time achieved 99%% of target RPS.\n");
-                } else if (rps_p1 >= 0.90 * target_rps) {
-                    printf("    Status: BOTTLENECK DETECTED\n");
-                    printf("    99%% of time achieved 90%% of target RPS.\n");
-                } else {
+                if (rps_p1 < 0.90 * target_rps) {
                     printf("    Status: SEVERE BOTTLENECK DETECTED\n");
-                    printf("    Performance significantly below target.\n");
+                } else if (rps_p1 < 0.99 * target_rps) {
+                    printf("    Status: BOTTLENECK DETECTED\n");
                 }
             } else {
                 /* Fallback to original average-based analysis */
@@ -1204,8 +1199,6 @@ static void showLatencyReport(void) {
                     printf("    Status: SEVERE BOTTLENECK DETECTED\n");
                 } else if (reqpersec < target_rps * 0.99f) {
                     printf("    Status: BOTTLENECK DETECTED\n");
-                } else {
-                    printf("    Status: GOOD PERFORMANCE\n");
                 }
             }
         }
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -1088,19 +1088,12 @@ static void showRPSReport(void) {
 
         const float target_rps = (float)config.rps;
 
-        // Bottleneck detection: if 95% of the RPS < 90% of target RPS
-        bool bottleneck_detected = (p95 < 0.9f * target_rps);
-
-        if (bottleneck_detected) {
-            printf("\n");
-            printf("RPS Summary (Bottleneck Detected):\n");
-            printf("  target RPS: %.2f\n", target_rps);
-            printf("  RPS distribution (reqs/sec):\n");
-            printf("    %9s %9s %9s %9s %9s %9s\n", "avg", "min", "p50", "p95", "p99", "max");
-            printf("    %9.3f %9.3f %9.3f %9.3f %9.3f %9.3f\n", avg_rps, p0, p50, p95, p99, p100);
-            printf("    Bottleneck rule: 95%% of RPS < 90%% of target RPS (%.2f)\n", 0.9f * target_rps);
-            printf("  status: BOTTLENECK DETECTED\n");
-        }
+        printf("\n");
+        printf("RPS Summary:\n");
+        printf("  target RPS: %.2f\n", target_rps);
+        printf("  RPS distribution (reqs/sec):\n");
+        printf("    %9s %9s %9s %9s %9s %9s\n", "avg", "min", "p50", "p95", "p99", "max");
+        printf("    %9.3f %9.3f %9.3f %9.3f %9.3f %9.3f\n", avg_rps, p0, p50, p95, p99, p100);
     }
 }
 
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -1160,12 +1160,12 @@ static void showLatencyReport(void) {
         printf("  latency summary (msec):\n");
         printf("    %9s %9s %9s %9s %9s %9s\n", "avg", "min", "p50", "p95", "p99", "max");
         printf("    %9.3f %9.3f %9.3f %9.3f %9.3f %9.3f\n", avg, p0, p50, p95, p99, p100);
-        
+
         /* RPS bottleneck detection */
         if (config.rps > 0) {
             float target_rps = (float)config.rps;
-            float threshold = target_rps * 0.9f;  /* 90% threshold */
-            
+            float threshold = target_rps * 0.9f; /* 90% threshold */
+
             if (reqpersec < threshold) {
                 printf("\n  RPS Analysis:\n");
                 printf("    Target RPS: %.0f\n", target_rps);
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -1080,10 +1080,10 @@ static void createMissingClients(client c) {
 static void showRPSReport(void) {
     if (config.rps_histogram && config.rps_histogram->total_count > 0) {
         const float avg_rps = hdr_mean(config.rps_histogram);
-        const float p0   = (float)hdr_min(config.rps_histogram);
-        const float p50  = (float)hdr_value_at_percentile(config.rps_histogram, 50.0);
-        const float p95  = (float)hdr_value_at_percentile(config.rps_histogram, 95.0);
-        const float p99  = (float)hdr_value_at_percentile(config.rps_histogram, 99.0);
+        const float p0 = (float)hdr_min(config.rps_histogram);
+        const float p50 = (float)hdr_value_at_percentile(config.rps_histogram, 50.0);
+        const float p95 = (float)hdr_value_at_percentile(config.rps_histogram, 95.0);
+        const float p99 = (float)hdr_value_at_percentile(config.rps_histogram, 99.0);
         const float p100 = (float)hdr_max(config.rps_histogram);
 
         const float target_rps = (float)config.rps;
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -1161,18 +1161,28 @@ static void showLatencyReport(void) {
         printf("    %9s %9s %9s %9s %9s %9s\n", "avg", "min", "p50", "p95", "p99", "max");
         printf("    %9.3f %9.3f %9.3f %9.3f %9.3f %9.3f\n", avg, p0, p50, p95, p99, p100);
 
-        /* RPS bottleneck detection */
+        /* RPS analysis - always show when --rps is used */
         if (config.rps > 0) {
             float target_rps = (float)config.rps;
-            float threshold = target_rps * 0.9f; /* 90% threshold */
+            float threshold_99 = target_rps * 0.99f; /* 99% threshold */
+            float threshold_90 = target_rps * 0.90f; /* 90% threshold */
+            float achievement_pct = (reqpersec / target_rps) * 100.0f;
 
-            if (reqpersec < threshold) {
-                printf("\n  RPS Analysis:\n");
-                printf("    Target RPS: %.0f\n", target_rps);
-                printf("    Actual RPS: %.2f\n", reqpersec);
-                printf("    Status: BOTTLENECK DETECTED\n");
-                printf("    The benchmark could not achieve the target RPS.\n");
+            printf("\n  RPS Analysis:\n");
+            printf("    Target RPS: %.0f\n", target_rps);
+            printf("    Actual RPS: %.2f (%.1f%% of target)\n", reqpersec, achievement_pct);
+            
+            if (reqpersec < threshold_90) {
+                printf("    Status: SEVERE BOTTLENECK DETECTED\n");
+                printf("    Performance is significantly below target (< 90%%).\n");
                 printf("    This indicates either client or server limitations.\n");
+            } else if (reqpersec < threshold_99) {
+                printf("    Status: BOTTLENECK DETECTED\n");
+                printf("    Performance is below optimal target (< 99%%).\n");
+                printf("    This indicates either client or server limitations.\n");
+            } else {
+                printf("    Status: GOOD PERFORMANCE\n");
+                printf("    Successfully achieved target RPS.\n");
             }
         }
     } else if (config.csv) {
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -1160,6 +1160,21 @@ static void showLatencyReport(void) {
         printf("  latency summary (msec):\n");
         printf("    %9s %9s %9s %9s %9s %9s\n", "avg", "min", "p50", "p95", "p99", "max");
         printf("    %9.3f %9.3f %9.3f %9.3f %9.3f %9.3f\n", avg, p0, p50, p95, p99, p100);
+        
+        /* RPS bottleneck detection */
+        if (config.rps > 0) {
+            float target_rps = (float)config.rps;
+            float threshold = target_rps * 0.9f;  /* 90% threshold */
+            
+            if (reqpersec < threshold) {
+                printf("\n  RPS Analysis:\n");
+                printf("    Target RPS: %.0f\n", target_rps);
+                printf("    Actual RPS: %.2f\n", reqpersec);
+                printf("    Status: BOTTLENECK DETECTED\n");
+                printf("    The benchmark could not achieve the target RPS.\n");
+                printf("    This indicates either client or server limitations.\n");
+            }
+        }
     } else if (config.csv) {
         printf("\"%s\",\"%.2f\",\"%.3f\",\"%.3f\",\"%.3f\",\"%.3f\",\"%.3f\",\"%.3f\"\n", config.title, reqpersec, avg,
                p0, p50, p95, p99, p100);
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -1997,7 +1997,6 @@ long long showThroughput(struct aeEventLoop *eventLoop, long long id, void *clie
                hdr_mean(config.current_sec_latency_histogram) / 1000.0f, hdr_mean(config.latency_histogram) / 1000.0f);
     config.last_printed_bytes = printed_bytes;
     hdr_reset(config.current_sec_latency_histogram);
-    sampleRPS();
     fflush(stdout);
     return SHOW_THROUGHPUT_INTERVAL;
 }
--- a/src/valkey-benchmark.c
+++ b/src/valkey-benchmark.c
@@ -1268,10 +1268,10 @@ static void benchmarkSequence(const char *title, char *cmd, int len, int seqlen)
              &config.current_sec_latency_histogram);     // Pointer to initialise
 
     if (config.rps > 0) {
-        hdr_init(1,                 // Min RPS: 1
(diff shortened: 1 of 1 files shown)


zuiderkwast:
@artikell

zuiderkwast:
How does this combine with #1761? When the `--rps` option is used, we don't want to push the server to maximum.

With `--rps`, the we want to be able to send the specified number of requests per second. If we can't, then it's a bottleneck of the benchmark script or of the server, but if we can, then neither of them is a bottleneck.

artikell:
It seems there is a misunderstanding. For this issue, we hope to give a reminder only if the pressure does not meet expectations under the condition of limited RPS.

My idea is to use a histogram to statistically analyze the Requests Per Second (RPS) and provide a specific distribution, such as p50, p90, and p99 values. Then, if the water level does not meet expectations, a prompt will be given.

@hanxizh9910

sarthakaggarwal97:
> My idea is to use a histogram to statistically analyze the Requests Per Second (RPS) and provide a specific distribution, such as p50, p90, and p99 values. Then, if the water level does not meet expectations, a prompt will be given.

@artikell I didn't completely get this. Would you please elaborate

hanxizh9910:
> > My idea is to use a histogram to statistically analyze the Requests Per Second (RPS) and provide a specific distribution, such as p50, p90, and p99 values. Then, if the water level does not meet expectations, a prompt will be given.
> 
> @artikell I didn't completely get this. Would you please elaborate

I don't know if my understanding is correct or not: my understanding is that we should create a statistical distribution of RPS measurements and calculate percentiles (p50, p90, p99). We'll establish a water level threshold (e.g., 90% of the user's target RPS). For each percentile, we check if it meets or exceeds this threshold. If any percentile falls below the water level, we then identify whether the bottleneck is on the client side (benchmark) or server side. Could you please you provide more details @artikell ?

artikell:
> > > My idea is to use a histogram to statistically analyze the Requests Per Second (RPS) and provide a specific distribution, such as p50, p90, and p99 values. Then, if the water level does not meet expectations, a prompt will be given.
> > 
> > 
> > @artikell I didn't completely get this. Would you please elaborate
> 
> I don't know if my understanding is correct or not: my understanding is that we should create a statistical distribution of RPS measurements and calculate percentiles (p50, p90, p99). We'll establish a water level threshold (e.g., 90% of the user's target RPS). For each percentile, we check if it meets or exceeds this threshold. If any percentile falls below the water level, we then identify whether the bottleneck is on the client side (benchmark) or server side. Could you please you provide more details @artikell ?

Here's the thing. We want users to know that the benchmark has reached a bottleneck. Locating the specific bottleneck is quite difficult, and I believe it's not within the scope of the benchmark's responsibility.

hanxizh9910:
> > > > My idea is to use a histogram to statistically analyze the Requests Per Second (RPS) and provide a specific distribution, such as p50, p90, and p99 values. Then, if the water level does not meet expectations, a prompt will be given.
> > > 
> > > 
> > > @artikell I didn't completely get this. Would you please elaborate
> > 
> > 
> > I don't know if my understanding is correct or not: my understanding is that we should create a statistical distribution of RPS measurements and calculate percentiles (p50, p90, p99). We'll establish a water level threshold (e.g., 90% of the user's target RPS). For each percentile, we check if it meets or exceeds this threshold. If any percentile falls below the water level, we then identify whether the bottleneck is on the client side (benchmark) or server side. Could you please you provide more details @artikell ?
> 
> Here's the thing. We want users to know that the benchmark has reached a bottleneck. Locating the specific bottleneck is quite difficult, and I believe it's not within the scope of the benchmark's responsibility.

Hi, I added bottleneck detection logic in `showLatencyReport()` that:
- Only activates when `--rps` parameter is specified
- Compares actual achieved RPS against target RPS
- Shows warning when actual RPS falls below 90% of target RPS
Would appreciate your feedback! @artikell 
Screenshot 2025-08-18 at 1 54 04 PM
Screenshot 2025-08-18 at 2 02 50 PM

artikell:
There is a difference between p99 and 90% here. If 10% of the requests do not meet expectations, it is also possible that the RPS has reached 90% @hanxizh9910

hanxizh9910:
> There is a difference between p99 and 90% here. If 10% of the requests do not meet expectations, it is also possible that the RPS has reached 90% @hanxizh9910

Thank you for the feedback! 

I've updated the implementation:

**New approach:**
- **≥99%**: "GOOD PERFORMANCE" - Successfully achieved target RPS
- **90-99%**: "BOTTLENECK DETECTED" - Performance below optimal target  
- **<90%**: "SEVERE BOTTLENECK DETECTED" - Performance significantly below target
Screenshot 2025-08-19 at 11 03 38 AM
Screenshot 2025-08-19 at 11 03 56 AM
Screenshot 2025-08-19 at 11 04 49 AM

Would you mind taking a look at this updated approach? @artikell

artikell:
@hanxizh9910 I didn't give a good example before, but here we need to use percentiles instead of averages. To avoid misjudgment.

hanxizh9910:
> @hanxizh9910 I didn't give a good example before, but here we need to use percentiles instead of averages. To avoid misjudgment.

@artikell Thank you for your clarification. I have a question about implementing RPS percentiles for short benchmark runs.

The issue: When benchmarks complete quickly, we get very few RPS samples:

Examples:

valkey-benchmark -n 1000 --rps 1000 → completes in ~1 second → only 1 RPS sample

valkey-benchmark -n 3000 --rps 1000 → completes in ~3 seconds → only 3 RPS samples

### Potential solution - Dynamic sampling intervals:
Instead of fixed 1-second intervals, could we use adaptive intervals to always get a consistent number of samples?

Approach: Always divide benchmark duration into 100 samples

- 5-second benchmark: Sample every 50ms (5s ÷ 100) → 100 samples

  - Expected requests per 50ms interval: 1000 RPS × 0.05s = 50 requests
  
- 30-second benchmark: Sample every 300ms (30s ÷ 100) → 100 samples

  - Expected requests per 300ms interval: 1000 RPS × 0.3s = 300 requests

- 2-minute benchmark: Sample every 1.2s (120s ÷ 100) → 100 samples

  - Expected requests per 1.2s interval: 1000 RPS × 1.2s = 1200 requests

This would ensure we always get exactly 100 RPS samples regardless of benchmark duration, making percentile calculations consistent and meaningful.

### Questions:

Do you think 100 samples is a good target number for reliable percentile analysis?

Or would you prefer a different approach for handling variable benchmark durations?

artikell:
@hanxizh9910 
The meaning of RPS is the meaning of every second, so we can use second level intervals. And we guarantee at least 1 second of data. Just make sure there is one sample.

For data that does not meet the requirement of 1 second, such as 200ms data with 1.2 seconds, this type of data can be skipped without affecting statistics.

Most users are more concerned about whether there is insufficient pressure during stress testing.

hanxizh9910:
> @hanxizh9910 The meaning of RPS is the meaning of every second, so we can use second level intervals. And we guarantee at least 1 second of data. Just make sure there is one sample.
> 
> For data that does not meet the requirement of 1 second, such as 200ms data with 1.2 seconds, this type of data can be skipped without affecting statistics.
> 
> Most users are more concerned about whether there is insufficient pressure during stress testing.

Thanks for the great suggestion! I've implemented RPS percentile analysis based on your feedback. Here's what I added:

New functionality:

- 1-second RPS sampling : Measures actual RPS every 1000ms during benchmark execution
- HDR histogram tracking : Uses same histogram library as latency analysis for consistent percentile calculation
- Performance guarantees display : Shows user-friendly "X% of time: RPS ≥ Y"

Example output: 

Screenshot 2025-08-28 at 11 22 46 AM
Screenshot 2025-08-28 at 11 20 21 AM
Screenshot 2025-08-28 at 11 16 08 AM

For tests shorter than 1 second, it falls back to the original average-based analysis to avoid misleading percentile data:

Screenshot 2025-08-28 at 2 01 05 PM


Would appreciate your feedback! @artikell

src/valkey-benchmark.c
```diff
@@ -1160,6 +1163,55 @@ static void showLatencyReport(void) {
...
+        /* RPS analysis - always show when --rps is used */
+        if (config.rps > 0) {
+            float target_rps = (float)config.rps;
+            float achievement_pct = (reqpersec / target_rps) * 100.0f;
+
+            printf("\n  RPS Analysis:\n");
+            printf("    Target RPS: %.0f\n", target_rps);
+            printf("    Actual RPS: %.2f (%.1f%% of target)\n", reqpersec, achievement_pct);
+
+            // ADD RPS PERCENTILE ANALYSIS
```
sarthakaggarwal97:
```suggestion
            /* ADD RPS PERCENTILE ANALYSIS */
```

src/valkey-benchmark.c
```diff
@@ -1160,6 +1163,55 @@ static void showLatencyReport(void) {
...
+            if (config.rps_histogram && config.rps_histogram->total_count > 0) {
+                const float rps_p1 = hdr_value_at_percentile(config.rps_histogram, 1.0);
+                const float rps_p10 = hdr_value_at_percentile(config.rps_histogram, 10.0);
+                const float rps_p50 = hdr_value_at_percentile(config.rps_histogram, 50.0);
+                // const float rps_p90 = hdr_value_at_percentile(config.rps_histogram, 90.0);
```
sarthakaggarwal97: Can we remove these comments if they are unused?

src/valkey-benchmark.c
```diff
@@ -1160,6 +1163,55 @@ static void showLatencyReport(void) {
...
+                    printf("    99%% of time achieved 99%% of target RPS.\n");
+                } else if (rps_p1 >= 0.90 * target_rps) {
+                    printf("    Status: BOTTLENECK DETECTED\n");
+                    printf("    99%% of time achieved 90%% of target RPS.\n");
+                } else {
+                    printf("    Status: SEVERE BOTTLENECK DETECTED\n");
+                    printf("    Performance significantly below target.\n");
+                }
+            } else {
+                // Fallback to original average-based analysis
```
sarthakaggarwal97: let's fix formatting style for all the comments.

src/valkey-benchmark.c
```diff
@@ -1224,6 +1276,15 @@ static void benchmarkSequence(const char *title, char *cmd, int len, int seqlen)
              config.precision,                           // Number of significant figures
              &config.current_sec_latency_histogram);     // Pointer to initialise
 
+    if (config.rps > 0) {
+        hdr_init(1,       // Min RPS: 1
+                 1000000, // Max RPS: 1M
```
sarthakaggarwal97: Is there a limit to what valkey benchmark can support?

sarthakaggarwal97 (approved):

sarthakaggarwal97 (changes requested):
Took an initial look. Will post more feedback.

src/valkey-benchmark.c
zuiderkwast: We can do more than 1M with threading + pipelining + fast machine.

src/valkey-benchmark.c
```diff
@@ -1160,6 +1163,55 @@ static void showLatencyReport(void) {
...
+                printf("      99%% of time: RPS ≥ %.0f (%.1f%% of target)\n",
+                       rps_p1, (rps_p1 / target_rps) * 100);
+                printf("      90%% of time: RPS ≥ %.0f (%.1f%% of target)\n",
+                       rps_p10, (rps_p10 / target_rps) * 100);
+                printf("      50%% of time: RPS ≥ %.0f (%.1f%% of target)\n",
+                       rps_p50, (rps_p50 / target_rps) * 100);
+
+                // Bottleneck analysis using P1 (99% of time achieved this or better)
+                if (rps_p1 >= 0.99 * target_rps) {
+                    printf("    Status: GOOD PERFORMANCE\n");
```
zuiderkwast: I'm not sure if no bottleneck always means good performance. If we tested with a very low RPS, we don't know much.

sarthakaggarwal97: yeah, it was in my pending comments too. I don't think we need to print this, we can just detect if the bottleneck is observed.

sarthakaggarwal97 (changes requested):

src/valkey-benchmark.c
```diff
@@ -1924,6 +1986,30 @@ int parseOptions(int argc, char **argv) {
     exit(exit_status);
 }
 
+static void sampleRPS(void) {
```
sarthakaggarwal97: I am not sure about the use of this function. Can we use similar mechanics as latency to build the histogram, they already do that to generate percentile values. I am guessing we also just need to build the right histogram for RPS and then let the user know if we were able to reach that RPS or not.

src/valkey-benchmark.c
```diff
@@ -1955,6 +1989,11 @@ long long showThroughput(struct aeEventLoop *eventLoop, long long id, void *clie
     const float rps = (float)requests_finished / dt;
     const float instantaneous_dt = (float)(current_tick - config.previous_tick) / 1000.0;
     const float instantaneous_rps = (float)(requests_finished - previous_requests_finished) / instantaneous_dt;
+
+    if (config.rps_histogram) {
```
sarthakaggarwal97, line 1988: did you try running the benchmarks with threads option? Can we check if it's running as expected with multiple threads and clients?

hanxizh9910: yes, i did. It's running as expected
Screenshot 2025-10-03 at 12 49 06 PM

zuiderkwast: Should we print the RPS percentiles below the latency percentiles, above the summary section?

Is the RPS percentiles useful for anything other than detecting a bottleneck? I'm still thinking we may want to hide it if no bottleneck is detected, to avoid spamming/bloating the output, if/when it is not useful.

hanxizh9910: Hi, i just updated the code based on your suggestion, now the RPS section will only be shown when there is a bottleneck. It will also appear above the summary section:

Screenshot 2025-10-06 at 12 20 28 PM
Screenshot 2025-10-06 at 12 20 39 PM

zuiderkwast:
@hanxizh9910 This behavior seems good to me now.

@artikell Is this working the way you think it should work?

@sarthakaggarwal97 When you have finished the code review, I will take a quick look and then think it will be ready to merge.

src/valkey-benchmark.c
```diff
@@ -1076,6 +1077,35 @@ static void createMissingClients(client c) {
...
+
+        float p50 = (float)hdr_value_at_percentile(config.rps_histogram, 50.0);
+        float p95 = (float)hdr_value_at_percentile(config.rps_histogram, 95.0);
+        float p99 = (float)hdr_value_at_percentile(config.rps_histogram, 99.0);
+
+        // Bottleneck detection using target RPS
+        float target_rps = (float)config.rps;
+        const char *rps_status = NULL;
+
+        if (p99 < 0.9 * target_rps) {
```
sarthakaggarwal97: This logic is a little confusing one, but I think it means that if 99% of the requests are less than 90% of the throughput, then we are severely bottlenecked. If 99% of the requests are less than RPS, then it is just bottlenecked.

I would suggest to make it P90 instead of P99 for both cases (as it is too lenient). If 90% of the requests are not able to reach 90% of the set RPS, it is a bottleneck. I also feel these thresholds 90% of the throughput, and 100% of the throughput, to detect bottleneck, could be a configuration that a user can set. Not everyone may agree with our thresholds.

@zuiderkwast what do you think?

src/valkey-benchmark.c
```diff
@@ -1076,6 +1077,35 @@ static void createMissingClients(client c) {
...
+            rps_status = "SEVERE BOTTLENECK DETECTED";
+        } else if (p99 < target_rps) {
+            rps_status = "BOTTLENECK DETECTED";
+        }
+
+        if (rps_status) {
+            printf("\nRPS Percentiles:\n");
```
sarthakaggarwal97: can we use the similar printing format as latency?

src/valkey-benchmark.c
sarthakaggarwal97: @hanxizh9910 when the logic is finalized, it would be nice to add a comment to explain this as well.

sarthakaggarwal97 (changes requested):
The PR looks better, left some comments to discuss.

src/valkey-benchmark.c
zuiderkwast: I'm don't really know. I think @artikell is better to answer this question.

artikell: Yes, it is appropriate to set this up as a configuration here. Indeed, not everyone accepts 90%.

zuiderkwast: More configuration? I'm skeptical to more configuration. It's hard to understand for users.

If we are uncertain, why not just print the info in that case so the user can judge from the output?

hanxizh9910: Hi @zuiderkwast, I added a print statement to explain how the bottleneck is detected. Does it look good to you?
Screenshot 2025-10-09 at 12 53 45 PM

sarthakaggarwal97: > More configuration? I'm skeptical to more configuration. It's hard to understand for users.
If we are uncertain, why not just print the info in that case so the user can judge from the output?

I like that @zuiderkwast always tries to avoid new configurations 😄. I think it's a much better idea that should print say the %age of requests that didn't meet the RPS configurations and let the user judge it. I'd prefer if we don't make the decisions on behalf of the user. @artikell do you have any objections to this?

artikell: LTGM.

hanxizh9910: > > More configuration? I'm skeptical to more configuration. It's hard to understand for users.
> > If we are uncertain, why not just print the info in that case so the user can judge from the output?
> 
> I like that @zuiderkwast always tries to avoid new configurations 😄. I think it's a much better idea that should print say the %age of requests that didn't meet the RPS configurations and let the user judge it. I'd prefer if we don't make the decisions on behalf of the user. @artikell do you have any objections to this?

Hi Sarthak @sarthakaggarwal97 , thank you for your feedback. I modified the code. Now, the output looks like this: 
Screenshot 2025-10-13 at 11 26 51 AM
If the user specifies the rps parameter, the RPS summary will be displayed; otherwise, only the latency will be shown:
Screenshot 2025-10-13 at 11 27 09 AM

zuiderkwast (commented):
After the last changes, it looks like the code is now ignoring the `-q` (quiet) and `--csv` options. Did you test with these? For example like this:

    ./valkey-benchmark -q --rps 1000 -t set
    ./valkey-benchmark --csv --rps 1000 -t set

Also I suggested earlier that we print the RPS percentiles above the Summary section and only when there is a bottleneck and you changed according to these suggestions. Now, it's changed again to be printed below the summary and always, without an explanation. I can accept changes if there is an explanation. :)

I still think the Summary should be printed last. The Summary includes the throughput (RPS). It's not only about latencies.

        printf("Summary:\n");
        printf("  throughput summary: %.2f requests per second\n", reqpersec);
        printf("  latency summary (msec):\n");
        printf("    %9s %9s %9s %9s %9s %9s\n", "avg", "min", "p50", "p95", "p99", "max");
        printf("    %9.3f %9.3f %9.3f %9.3f %9.3f %9.3f\n", avg, p0, p50, p95, p99, p100);

hanxizh9910:
Hi @zuiderkwast, thank you for your feedback. I had an offline discussion with @sarthakaggarwal97.
We’ve moved the RPS section to appear before the latency distribution. Is that what you had in mind?
Also, the RPS section will now be printed only when 10% or more of the samples fall below 90% of the target RPS.

## Show the rps:
Screenshot 2025-10-14 at 11 13 35 AM

## Not show the rps:
Screenshot 2025-10-14 at 11 14 28 AM

Also, now the showRPSReport function is called within showLatencyReport, the -q (quiet) and --csv options are handled by showLatencyReport. Below are the current outputs.

## -q
Screenshot 2025-10-14 at 11 11 48 AM

## --csv
Screenshot 2025-10-14 at 11 12 45 AM

zuiderkwast:
> Hi @zuiderkwast, thank you for your feedback. I had an offline discussion with @sarthakaggarwal97.

Ah, good to know.

> We’ve moved the RPS section to appear before the latency distribution. Is that what you had in mind?

That's OK. I don't mind which one is first (RPS distribution or latency distribution) as long as both of these are before the Summary. Also good that you put this back within the `if` so -q and --csv are handled correctly.

If you think the function name `showLatencyReport` is incorrect because it shows throughput *and* latency, feel free to rename it. Maybe just showReport or something like that?

> Also, the RPS section will now be printed only when 10% or more of the samples fall below 90% of the target RPS.

OK. If you want to print it always, I guess it's fine too. I just want to understand why. :smile: I didn't see any comment mention that it's better to show it always. (To me it seems like it will not be useful when there is no bottleneck, but I don't know if you and others agree.)

hanxizh9910:
> > Hi @zuiderkwast, thank you for your feedback. I had an offline discussion with @sarthakaggarwal97.
> 
> Ah, good to know.
> 
> > We’ve moved the RPS section to appear before the latency distribution. Is that what you had in mind?
> 
> That's OK. I don't mind which one is first (RPS distribution or latency distribution) as long as both of these are before the Summary. Also good that you put this back within the `if` so -q and --csv are handled correctly.
> 
> If you think the function name `showLatencyReport` is incorrect because it shows throughput _and_ latency, feel free to rename it. Maybe just showReport or something like that?
> 
> > Also, the RPS section will now be printed only when 10% or more of the samples fall below 90% of the target RPS.
> 
> OK. If you want to print it always, I guess it's fine too. I just want to understand why. 😄 I didn't see any comment mention that it's better to show it always. (To me it seems like it will not be useful when there is no bottleneck, but I don't know if you and others agree.)

@zuiderkwast Hi Victor, thank you for your quick reply! I’ve changed the function from showLatencyReport to showReport.

We were thinking that if we cannot decide a default threshold for what counts as a bottleneck, we could simply print the RPS section whenever the user uses the --rps option.

zuiderkwast:
> We were thinking that if we cannot decide a default threshold for what counts as a bottleneck, we could simply print the RPS section whenever the user uses the --rps option.

I understand. OK, I'm fine with that, if you prefer it.

hanxizh9910:
@zuiderkwast thank you for your feedback! This is the latest output:
Screenshot 2025-10-14 at 3 41 05 PM
It will always print the RPS summary if the user uses the --rps option

hanxizh9910:
Hi @zuiderkwast, just wanted to follow up on my last update regarding the PR. Have you had a chance to review the latest change? Appreciate your time! Thanks!

zuiderkwast:
Hi @hanxizh9910, yeah, as I mentioned in https://github.com/valkey-io/valkey/pull/2471#issuecomment-3373937957 above, I was waiting for @sarthakaggarwal97 to approve. I'm trying to delegate a little here.

> @sarthakaggarwal97 When you have finished the code review, I will take a quick look and then think it will be ready to merge.

I still see this though:

image

zuiderkwast (approved):

sarthakaggarwal97 (approved):
LGTM! I had discussed offline with @hanxizh9910 but forgot to approve it :D
Thanks @hanxizh9910 for taking up the change.

zuiderkwast:
OK, I'll merge this then. We use the title and top-comment as the commit message, but they are outdated now. We don't use EAGAIN and I'm not sure we if can call it bottleneck-detection anymore. Maybe "Show RPS histogram in valkey-benchmark" is a better title? I updated it now.

My hope was that it would be possible to understand if the bottleneck is valkey-benchmark (then you can run valkey-benchmark from multiple machines in parallel to generate more load) or valkey-server or the network or something else, but apparently this is too hard to determine? I'm still not quite clear about why the EAGAIN approach isn't useful for that though.... In my mind, it could at least give some indication in some cases... We can continue the discussion if you want. Anyway, I don't think this implementation really solves #2219 so I changed "closes" to "related to".

hanxizh9910:
@zuiderkwast Yes, i totally agree with you. this PR does not solve #2219, it just use a histogram to statistically show the RPS summary. I was following these suggestions: https://github.com/valkey-io/valkey/pull/2471#issuecomment-3188848782,  https://github.com/valkey-io/valkey/pull/2471#issuecomment-3193452398.

sarthakaggarwal97:
yeah @zuiderkwast thanks for updating the description, that is important.

zuiderkwast:
@hanxizh9910 Yeah. I think this can be useful anyway. Thank you!

PR #2475, opened 2025-08-12
HSETEX 'hset' notification should only be generated if not expired

ranshid:
Currently HSETEX always generate `hset` notification. In order to align with generic `set` command, it should only generate `hset` if the provided time-to-live is a valid future time.


ranshid:
A test for it should be taken and modified from https://github.com/valkey-io/valkey/pull/2474

enjoy-binbin (approved):

PR #2477, opened 2025-08-13
Remove if condition and disable the new failover test

enjoy-binbin:
In #2431 we changed the assert to a if condition, and the test cause some trouble, now we just remove the assert (if condition) and disable the test for now due to #2441.

### DIFF
--- a/tests/unit/cluster/manual-failover.tcl
+++ b/tests/unit/cluster/manual-failover.tcl
@@ -584,6 +584,8 @@ start_cluster 3 2 {tags {external:skip cluster}} {
     }
 }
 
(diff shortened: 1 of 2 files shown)


PingXie (approved):

tests/unit/cluster/manual-failover.tcl
```diff
@@ -584,6 +584,8 @@ start_cluster 3 2 {tags {external:skip cluster}} {
     }
 }
 
+# Disable this case due to #2441.
```
enjoy-binbin:
```suggestion
# Disable this test case due to #2441.
```

PR #2486, opened 2025-08-14
Ensures presence of slots on the node before test is run

### LINKED ISSUE #2480: [test failure] MOVED 16382 127.0.0.1:25618 in `proc populate`
Probably introduced in #1949.

test "Import slot range with multiple slots", tests/unit/cluster/cluster-migrateslots.tcl

https://github.com/valkey-io/valkey/actions/runs/16941711521/job/48012293693?pr=2092#step:5:6617

sarthakaggarwal97:
The change will ensure that the slot is present on the node before the slot is populated. This will avoid the errors during populating the slot.

Resolves #2480

### DIFF
--- a/tests/unit/cluster/cluster-migrateslots.tcl
+++ b/tests/unit/cluster/cluster-migrateslots.tcl
@@ -78,13 +78,13 @@ proc wait_for_migration {node_idx slot} {
 # Ensure that a given hash slot is served by the specified node.
 # If the slot is currently owned by a different node, move it to the target
 # and wait for the migration to complete.
-proc ensure_slot_on_node {slot target_idx} {
-    if {[is_slot_migrated $target_idx $slot]} {return}
-    set target_id [R $target_idx CLUSTER MYID]
+proc ensure_slot_on_node {node_idx slot} {
+    if {[is_slot_migrated $node_idx $slot]} {return}
+    set target_id [R $node_idx CLUSTER MYID]
     for {set i 0} {$i < 3} {incr i} {
         if {[is_slot_migrated $i $slot]} {
             assert_match "OK" [R $i CLUSTER MIGRATESLOTS SLOTSRANGE $slot $slot NODE $target_id]
-            wait_for_migration $target_idx $slot
+            wait_for_migration $node_idx $slot
             return
         }
     }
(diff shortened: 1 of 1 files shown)


tests/unit/cluster/cluster-migrateslots.tcl
```diff
@@ -75,6 +75,22 @@ proc wait_for_migration {node_idx slot} {
     wait_for_cluster_propagation
 }
 
+# Ensure that a given hash slot is served by the specified node.
+# If the slot is currently owned by a different node, move it to the target
+# and wait for the migration to complete.
+proc ensure_slot_on_node {slot target_idx} {
```
enjoy-binbin: we usually do this? I dont mind srv_id or target_id (or node_idx, i see in this tcl file, we call it node_idx in some procs), lets put xxx_id in the first parameter
```suggestion
proc ensure_slot_on_node {node_idx slot} {
```

sarthakaggarwal97: I have addressed this, thank you!

enjoy-binbin (approved):

tests/unit/cluster/cluster-migrateslots.tcl
```diff
@@ -75,6 +75,22 @@ proc wait_for_migration {node_idx slot} {
     wait_for_cluster_propagation
 }
 
+# Ensure that a given hash slot is served by the specified node.
+# If the slot is currently owned by a different node, move it to the target
+# and wait for the migration to complete.
+proc ensure_slot_on_node {node_idx slot} {
+    if {[is_slot_migrated $node_idx $slot]} {return}
```
zuiderkwast, line 82: Nice function. I believe it can be useful also for other cluster test suites. Many tests use a hard-coded key and slot to write some key.

We could move some of these functions to a generic file used for cluster tests, but we can do it later when we need them in other suites.

zuiderkwast (approved):
Great!

tests/unit/cluster/cluster-migrateslots.tcl
zuiderkwast: `is_slot_migrated` should probably be called `is_slot_owned_by_node` instead. That's what it checks.

PR #2210, opened 2025-06-13
Fixes for Valkey 8.0.5

vitarb:
This backports all remaining PRs from https://github.com/orgs/valkey-io/projects/2/views/1

* :heavy_check_mark:  PR #2109
* :heavy_check_mark:  PR #2137
* :heavy_check_mark:  PR #2132
* :heavy_check_mark:  PR #2117 
* :heavy_check_mark:  PR #2140 
* :heavy_check_mark:  PR #2144 
* :heavy_check_mark:  PR #2146 
* :heavy_check_mark:  PR #2178 
* :heavy_check_mark:  PR #2186 
* :heavy_check_mark:  PR #2196 (is a test dependency of #2178)
* :heavy_check_mark:  PR #1252 (is a test dependency of #2178)
* :heavy_check_mark:  PR #1341 (is a test dependency of #2178)
* :heavy_check_mark:  PR #1067
* :heavy_check_mark:  PR #1274 (is a test dependency of #2178, also we can backport it)
* :heavy_check_mark:  #1077
  * Fixes `[exception]: Executing test client: CLUSTERDOWN The cluster is down.`
* :heavy_check_mark:  #1549
  * Fixes reply-schema-validator job
* :heavy_check_mark:  #2399
* :heavy_check_mark:  Disable clang-format in 8.0 to simplify backporting
* :heavy_check_mark:  #2345
* :heavy_check_mark:  Fix defrag timeouts in 8.0
* :heavy_check_mark:  #2229
* :heavy_check_mark:  #2360
* :heavy_check_mark:  #2174
* :heavy_check_mark:  #2526
* :heavy_check_mark:  #2466

PR #2488, opened 2025-08-15
Disable clang-format in 8.0 to ease backporting

zuiderkwast:
Our clang-format rules were introduced in 8.0 but they were changed later, so we get formatting errors when we backport PRs to 8.0. It's better to just disable clang-format for backporting to 8.0.


zuiderkwast:
I'm doing it directly in #2210 instead.

PR #2493, opened 2025-08-16
CONFIG GET command return sorted output

### LINKED ISSUE #2042: [NEW] CONFIG GET should preserve input order or return sorted results

When using CONFIG GET with multiple parameters, the returned key-value pairs come back in an unspecified order. This makes it harder to match results to inputs in scripts or tools, complicates automated validation, and makes config diffs less predictable.

CONFIG GET should return results in the same order as the requested parameters. Alternatively, it could return the results sorted alphabetically by parameter name for predictability.

- Post-processing the results in client code to re-order them. This is error-prone and unnecessary overhead.
- Using separate CONFIG GET calls per parameter, which adds round-trip latency.

**Additional information**
Other commands already preserve order (eg.  CLUSTER SLOTS,)

yzc-yzc:
Previously, the config names and values were stored in a temporary dict. This ensures that no duplicates are returned, but it also makes the order random.

In this PR, the config names and values still stored in the temporary dict, but then they are copied to an array, which is sorted, before the reply is sent.

Resolves #2042

### DIFF
--- a/src/config.c
+++ b/src/config.c
@@ -997,29 +997,30 @@ void configGetCommand(client *c) {
     }
 
     di = dictGetIterator(matches);
-    unsigned long matchesSize = dictSize(matches);
-    addReplyMapLen(c, matchesSize);
+    int n = dictSize(matches);
+    addReplyMapLen(c, n);
 
     struct {
         const char *key;
         sds value;
-    } *sorted = zmalloc(sizeof(*sorted) * matchesSize);
-    unsigned long sortedIndex = 0;
+    } *sorted = zmalloc(sizeof(*sorted) * n);
+
+    i = 0;
     while ((de = dictNext(di)) != NULL) {
         standardConfig *config = (standardConfig *)dictGetVal(de);
-        sorted[sortedIndex].key = dictGetKey(de);
-        sorted[sortedIndex].value = config->interface.get(config);
-        sortedIndex++;
+        sorted[i].key = dictGetKey(de);
+        sorted[i].value = config->interface.get(config);
+        i++;
     }
-    qsort(sorted, matchesSize, sizeof(*sorted), configKeyCompare);
-
-    for (sortedIndex = 0; sortedIndex < matchesSize; sortedIndex++) {
-        addReplyBulkCString(c, sorted[sortedIndex].key);
-        addReplyBulkSds(c, sorted[sortedIndex].value);
-    }
-    zfree(sorted);
     dictReleaseIterator(di);
     dictRelease(matches);
+
+    qsort(sorted, n, sizeof(*sorted), configKeyCompare);
+    for (i = 0; i < n; i++) {
+        addReplyBulkCString(c, sorted[i].key);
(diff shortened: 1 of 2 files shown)


src/config.c
```diff
@@ -990,12 +997,27 @@ void configGetCommand(client *c) {
...
     di = dictGetIterator(matches);
-    addReplyMapLen(c, dictSize(matches));
+    unsigned long matchesSize = dictSize(matches);
+    addReplyMapLen(c, matchesSize);
+
+    struct {
+        const char *key;
+        sds value;
+    } *sorted = zmalloc(sizeof(*sorted) * matchesSize);
+    unsigned long sortedIndex = 0;
```
zuiderkwast: For local variables, we normally use snake case like `sorted_index` or just short names like `i`. Since the scope is pretty small here, I don't mind that we reuse the `i` variable already declared in this function.

Also matchesSize could be named `matches_size`, `num_matches` or just `n`.

If we use `i` and `n`, then the for loop below will be the typical short C style `for (i = 0; i < n; i++)`. I kind of like it. (Of course this brief style only works in simple functions without risk of confusion, so here it works well IMHO.)

zuiderkwast (commented):
Great! Just some minor comment.

I updated the PR top comment to explain the change, so people can understand it without looking at the code.

src/config.c
```diff
@@ -990,12 +997,27 @@ void configGetCommand(client *c) {
...
+    }
+    qsort(sorted, matchesSize, sizeof(*sorted), configKeyCompare);
+
+    for (sortedIndex = 0; sortedIndex < matchesSize; sortedIndex++) {
+        addReplyBulkCString(c, sorted[sortedIndex].key);
+        addReplyBulkSds(c, sorted[sortedIndex].value);
     }
+    zfree(sorted);
     dictReleaseIterator(di);
     dictRelease(matches);
```
zuiderkwast: I don't know if it matters, but I guess we can release the dict and dict iterator as early as possible, before qsort, to save some memory.

src/config.c
yzc-yzc: Thanks for the detailed description of the code naming style. Fixed

tests/unit/introspection.tcl
```diff
@@ -1143,6 +1143,15 @@ start_server {tags {"introspection"}} {
         r client info
     } {*lib-name= *}
 
+    test {CONFIG get should return sorted output} {
```
zuiderkwast:
```suggestion
    test {CONFIG GET should return sorted output} {
```

zuiderkwast (approved):

enjoy-binbin (approved):

zuiderkwast:
@yzc-yzc Thank you! Can you also open a PR in valkey-doc? It's the file commands/config-get.md, to mention that the reply is sorted by name/field/key (not sure which terminology we use) and to fix the example to be sorted.

madolson:
@enjoy-binbin This is merged in unstable, doesn't that mean it should go out in 9.0?

PR #2494, opened 2025-08-17
Split SLOT_EXPORT_AUTHENTICATING into SEND and READ to avoid synchronous reading of auth response

enjoy-binbin:
The old SLOT_EXPORT_AUTHENTICATING added in #1949, when processed by the source node,
we will send the AUTH command and then reads the response. If the target node is blocked
during this process, the source node will also be blocked. We should use a read handler
to handle this. We split SLOT_EXPORT_AUTHENTICATING into SLOT_EXPORT_SEND_AUTH and
SLOT_EXPORT_READ_AUTH_RESPONSE to avoid this issue.

### DIFF
--- a/src/cluster_migrateslots.c
+++ b/src/cluster_migrateslots.c
@@ -2012,6 +2012,7 @@ bool canSlotMigrationJobSendAck(slotMigrationJob *job) {
      *    job. */
     return job->state != SLOT_EXPORT_SNAPSHOTTING &&
            job->state != SLOT_IMPORT_WAIT_ACK &&
+           job->state != SLOT_EXPORT_CONNECTING &&
            job->state != SLOT_EXPORT_SEND_AUTH &&
            job->state != SLOT_EXPORT_READ_AUTH_RESPONSE &&
            job->state != SLOT_EXPORT_SEND_ESTABLISH &&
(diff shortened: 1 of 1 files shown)


src/cluster_migrateslots.c
```diff
@@ -1012,41 +1017,48 @@ int proceedWithSlotExportJobConnecting(slotMigrationJob *job, bool *completed) {
...
-        return C_ERR;
-    }
-    err = receiveSynchronousResponse(job->conn);
+/* Read a response to the AUTH command, moving to the next stage of the migration
+ * if the response is a success. If there is an error, fail the migration with the
+ * error message. */
+void slotMigrationJobReadAuthResponse(connection *conn) {
+    slotMigrationJob *job = (slotMigrationJob *)connGetPrivateData(conn);
+
+    sds err = receiveSynchronousResponse(job->conn);
```
murphyjacob4, line 1026: Ideally we wouldn't use `receiveSynchronousResponse` (e.g. in case the response isn't fully received) but in practice I think this should work good enough.

We can iterate on it if needed

murphyjacob4 (approved):
LGTM

src/cluster_migrateslots.c
enjoy-binbin: yes, it should work good enough since the reply is quite small and we can mostly read it all at once. We using receiveSynchronousResponse a lot in ours code right now.

murphyjacob4: Make sense 👍

Issue #2343, opened 2025-07-11
[BUG] HyperLogLog broken with embedded values


ChiliPaneer:
**Describe the bug**

We do not account for HyperLogLogs being embedded. While we never create a HyperLogLog as an embedded value (we use `createObject(OBJ_STRING, s);` bypassing the embedding logic), when loaded from an rdb, there are no checks in place to prevent an HLL from being embedded.

We encounter undefined behaviour in at least two instances within the HyperLogLog implementation if the HyperLogLog is embedded. Since embedded values are a part of a larger single allocation the following occur:

- `hllSparseAdd()` calls `sdsResize()`, which calls `sdsFree()`, which would attempt to free in the middle of an allocation.
- `hllSparseToDense()` calls `sdsfree()` directly

**To reproduce**
I'm still working on a reproduction, but it should look something like this:
1. Create a HyperLogLog item that is short
2. Save a snapshot
3. Load the snapshot into 8.1
4. The HLL will load as embedded
5. Keep adding values till the size increases or trigger the conversion to the dense representation.
6. This will cause the free in the middle of the allocation.

**Expected behavior**
With some additional usage, if it doesn't segfault immediately, we should eventually see some sort of unexpected behavior.

JimB123:
My guess is that the best way to address this would be WITHIN the HLL code.  Basically, when a HLL function is called which may modify the HLL, it should call some type of object function which will un-embed the value, if embedded.

Ideally, the entire solution is within the HLL code.  We shouldn't need code within tiered storage or RDB loading to be aware of this.

Or when HLL modifies the value, it needs to call the proper hash table functions to do it correctly - so that it remains embedded if it will fit within the allocated space or reallocates or unembeds.  (This is probably more involved than simply unembedding if HLL ever encounters an embedded value.)

zuiderkwast:
If encoding != RAW, we can call `dbUnshareStringValue`, similar to how we do in VM_StringDMA: https://github.com/valkey-io/valkey/blob/unstable/src/module.c#L4444

zuiderkwast:
To test it, I guess we could also use GET and then SET to make it an embedded string.

yzc-yzc:
Taking a quick look at the HLL-related code, I found that there has long been `dbUnshareStringValue` on the critical path to ensure correctness.
e.g.
[...]

PR #2496, opened 2025-08-17
Add static specifier to the internal functions of HLL

yzc-yzc:
Related: #2343


rainsupreme (approved):
lightweight and worthwhile 👍

PR #2497, opened 2025-08-17
Don't allow slot migration to myself node

enjoy-binbin:
This may result in meaningless slot migration job, we should
return an error to user in advance to avoid operation error.
Also `by myself` is not correct English grammar and `myself`
is a internal code terminology, changed to `by this node`.

Was introduced in #1949.

### DIFF
--- a/src/cluster_migrateslots.c
+++ b/src/cluster_migrateslots.c
@@ -893,7 +893,7 @@ void clusterCommandMigrateSlots(client *c) {
             goto cleanup;
         }
         if (target_node == server.cluster->myself) {
-            addReplyError(c, "Slots are served by this node.");
+            addReplyError(c, "Target node can not be this node.");
             goto cleanup;
         }
         curr_index++;
--- a/src/cluster_migrateslots.c
+++ b/src/cluster_migrateslots.c
@@ -847,7 +847,7 @@ void clusterCommandMigrateSlots(client *c) {
             goto cleanup;
         }
         if (source_node != server.cluster->myself) {
-            addReplyError(c, "Slots are not served by myself.");
+            addReplyError(c, "Slots are not served by this node.");
             goto cleanup;
         }
 
@@ -893,7 +893,7 @@ void clusterCommandMigrateSlots(client *c) {
             goto cleanup;
         }
         if (target_node == server.cluster->myself) {
-            addReplyError(c, "Slots are served by myself.");
+            addReplyError(c, "Slots are served by this node.");
             goto cleanup;
         }
         curr_index++;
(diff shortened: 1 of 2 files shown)


src/cluster_migrateslots.c
```diff
@@ -892,6 +892,10 @@ void clusterCommandMigrateSlots(client *c) {
                                 (sds)c->argv[curr_index]->ptr);
             goto cleanup;
         }
+        if (target_node == server.cluster->myself) {
+            addReplyError(c, "Slots are served by myself.");
```
enjoy-binbin: Maybe we can return other errors, such as the target node is myself, `Slots are served by myself` seems a bit strange in this condition.

zuiderkwast: "by myself" is not correct English grammar. It should be "by me", right? Or maybe "by this node" sounds better?

`myself` is internal code terminology.

src/cluster_migrateslots.c
```diff
@@ -847,7 +847,7 @@ void clusterCommandMigrateSlots(client *c) {
             goto cleanup;
         }
         if (source_node != server.cluster->myself) {
-            addReplyErrorFormat(c, "Slots are not served by myself.");
+            addReplyError(c, "Slots are not served by myself.");
```
zuiderkwast: "by this node"?

src/cluster_migrateslots.c
```diff
@@ -892,6 +892,10 @@ void clusterCommandMigrateSlots(client *c) {
                                 (sds)c->argv[curr_index]->ptr);
             goto cleanup;
         }
+        if (target_node == server.cluster->myself) {
+            addReplyError(c, "Slots are served by this node.");
+            goto cleanup;
```
zuiderkwast, line 897: Another possibility is to return OK, success immediately, without starting any migration job. I'm not sure if OK or error is better. I will let @murphyjacob4 decide.

madolson: I think throwing an error is a better idempotent behavior. Otherwise systems might have logic looking for the started slot migration. I don't think we should return OK if we didn't start a migration.

src/cluster_migrateslots.c
```diff
@@ -892,6 +892,10 @@ void clusterCommandMigrateSlots(client *c) {
                                 (sds)c->argv[curr_index]->ptr);
             goto cleanup;
         }
+        if (target_node == server.cluster->myself) {
+            addReplyError(c, "Slots are served by this node.");
```
madolson:
```suggestion
            addReplyError(c, "Target node can not be this node.");
```

zuiderkwast: can not => cannot or can't?

madolson: Most style guides avoid contractions. "can not" and "cannot" mean the same thing, so either works for me.

hpatro (approved):

madolson (approved):

PR #2498, opened 2025-08-17
Remove the outupdated unknown key/value pairs comment in CLUSTER SYNCSLOTS ESTABLISH

enjoy-binbin:
We now have #2688 SYNCSLOTS CAPA, remove the outupdated comment.


src/cluster_migrateslots.c
```diff
@@ -474,8 +474,6 @@ void clusterCommandSyncSlotsEstablish(client *c) {
 
             continue;
         }
-        addReplyErrorObject(c, shared.syntaxerr);
-        goto cleanup;
```
enjoy-binbin, line 612: We have this comment at line 422
```
    /* Order agnostic. We skip unknown key/value pairs forwards
     * compatibility. */
    int i = 3;
    while (i < c->argc) {
        if (!strcasecmp(c->argv[i]->ptr, "source")) {
            if (source_node || i + 1 >= c->argc ||
                sdslen(c->argv[i + 1]->ptr) != CLUSTER_NAMELEN) {
                goto cleanup;
            }
```

murphyjacob4: I think the comment may be out of sync.

In the review - we talked about it and went a different direction: https://github.com/valkey-io/valkey/pull/1949/files/a4e787a7a7dda25891d48c1e3f88233929d01dc7

Let me know if you think we should ignore unknown fields

enjoy-binbin: Do you have the discussion link? I can't find the discussion.

Since this is a internal command, i tend to ignore the unknown k/v, like replconf capa.

BUT both are ok as long as we don't miss it.

zuiderkwast:
We have SYNCSLOTS CAPA now: #2688. We don't use it now, but for the future it provides a way to send information where unknown information will be ignored by older nodes.

Do we still need this PR or can we close it?

murphyjacob4:
Yeah I agree - let's use SYNCSLOTS CAPA - if anything unknown arrives outside of that - it would be unexpected. @enjoy-binbin wdyt?

enjoy-binbin:
i normally like the ignore way in internal command, but i guess capa can work well as well, let's just remove the outupdated comment.

murphyjacob4 (approved):

PR #2502, opened 2025-08-18
Fix memory leak in rdbLoadObject when loading a wrong HFE

enjoy-binbin:
There are memory leaks when we return NULL.

PR #2503, opened 2025-08-18
Fix memory leak in saveSnapshotToConnectionSockets

enjoy-binbin:
We now pass in rdbSnapshotOptions options in this function, and options.conns
is now malloc'ed in the caller side, so we need to zfree it when returning early
due to an error. Previously, conns was malloc'ed after the error handling, so we
don't have this.

Introduced in #1949.

### DIFF
--- a/src/rdb.c
+++ b/src/rdb.c
@@ -3586,8 +3586,9 @@ void killRDBChild(void) {
 /* Save snapshot to the provided connections, spawning a child process and
  * running the provided function.
  *
- * Connections array provided will be freed after the save is completed, and
- * should not be freed by the caller. */
+ * The connections array (the conns field in the rdbSnapshotOptions) is a
+ * heap-allocated array that will be freed by this function and shall not be
+ * freed by the caller. */
 int saveSnapshotToConnectionSockets(rdbSnapshotOptions options) {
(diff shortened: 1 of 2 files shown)


zuiderkwast (approved):
The top comment of this function has this comment:

```c
 * Connections array provided will be freed after the save is completed, and
 * should not be freed by the caller. */
```

We should update it to say that this array is freed by this function (success or failure doesn't matter). Maybe also clarify that it's the `conns` field in the `rdbSnapshotOptions` struct. It's also required that this array is heap allocated. Otherwise it can't be freed. Idea:

```c
 * The connections array (the conns field in the rdbSnapshotOptions) is a
 * heap-allocated array that will be freed by this function and shall not be
 * freed by the caller. */
```

It's a little odd that this struct is passed by value, but it contains a heap-allocated array. Maybe we could use a different design, but the simple fix is just like this PR and to update the comment.

zuiderkwast:
Is it this leak seen in Daily today? https://github.com/valkey-io/valkey/actions/runs/17027650597/job/48265126023#step:6:7328

enjoy-binbin:
that one was fixed in #2502

PR #2505, opened 2025-08-18
fix hash ignore ttl management during active expiry

ranshid:
several fixes:
1. fix not using bool input type for hashTypeIgnoreTTL - probably lost during the 3 HFE PR merges
2. remove vset change hashtable encoding to single - The current code is a bug. The entry is probably expired (or about to be expired soon) so we can leave it as a hashtable till it does.
3. enable incremental rehashing for volatile item keys kvstore - This is the center issue of this PR. without it the activeexpiry might not scan the kvstore which is very fragmented with lots of empty buckets.


src/t_hash.c
```diff
@@ -2138,21 +2136,21 @@ static int hashTypeExpireEntry(void *entry, void *c) {
...
     expiryContext ctx = {.key = o, .fields = out_entries, .n_fields = 0};
     size_t expired = vsetRemoveExpired(vset, entryGetExpiry, hashTypeExpireEntry, now, max_fields, &ctx);
     serverAssert(ctx.n_fields <= max_fields);
-    hashTypeIgnoreTTL(o, 0);
-    if (!hashTypeHasVolatileFields(o)) {
+    if (vsetIsEmpty(vset)) {
         hashTypeFreeVolatileSet(o);
+    } else {
+        hashTypeIgnoreTTL(o, false);
     }
```
zuiderkwast, line 2154: The logic looks correct. The bug is that we didn't free the empty volatile set after removing all the expired fields, right?

I can see in the source code of `hashTypeFreeVolatileSet()` that it implies `hashTypeIgnoreTTL(o, true)`, but the naming and documentation of these functions don't indicate that it's OK to skip restoring `hashTypeIgnoreTTL` afterwards if the volatile set is freed. It looks unbalanced to me, as if leaving it as `true` here sets a flag that keeps TTL ignored forever even after adding more volatile fields. Ideally this should be obvious from the function names alone.

src/vset.c
```diff
@@ -1456,14 +1456,11 @@ static inline size_t vsetBucketRemoveExpired_HASHTABLE(vsetBucket **bucket, vset
...
 
-    /* in case we completed scanning the hashtable or a single element is left, we can convert the hashtable. */
+    /* in case we completed scanning the hashtable which is now empty */
     size_t ht_size = hashtableSize(ht);
     if (ht_size == 0) {
         hashtableRelease(ht);
         *bucket = vsetBucketFromNone();
-    } else if (ht_size == 1) {
-        assert(entry);
-        *bucket = vsetBucketFromSingle(entry);
```
zuiderkwast, line 1466: You skip this because if it goes from 2 to 1 to 2 entries, we need to recreated the hash table again and again? Or why is it bad to convert to single here?

zuiderkwast (approved):

src/vset.c
zuiderkwast: It was a leak? We didn't call hashtableRelease here?

zuiderkwast: Ah, we're not even sure if `entry` was deleted or not. It may have been deleted in the loop above.

ranshid: It was just an optimization, and the code is a bug since it kept using the old entry. I think that given the entries in the bucket are already expired it makes more sense to just avoid changing the encoding and keep it as a hashtable

src/t_hash.c
ranshid: The bug was that we did not provide bool as input to the hashTypeIgnoreTTL. I can still do hashTypeIgnoreTTL in any case but it is redundant in case we free the vset in hashTypeFreeVolatileSet

zuiderkwast: Just using 0 and 1 instead of false and true is not a bug... It's the same, just a little less clear, I think.

zuiderkwast: I can see that it's redundant to call hashTypeIgnoreTTL, but it's not immediately clear from this code that it's redundant and that it's OK to skip it. Maybe a different naming of these functions would make it more clear.

Either way, I'm fine with this change. Let's merge. :)

src/vset.c
zuiderkwast: It was probably also a leak, since we did `*bucket = vsetBucketFromSingle(entry);` without calling `hashtableRelease(ht);`

src/t_hash.c
ranshid: agree on both. the main issue in this case was the vset bug + the fact that active expiration was not registered to incremental rehashing, leading to that the active expiration stopped before clearing all the fields.

src/vset.c
ranshid: right

PR #2508, opened 2025-08-19
Optimize zset memory usage by embedding element in skiplist

chzhoo:
By default, when the number of elements in a zset exceeds 128, the underlying data structure adopts a skiplist. We can reduce memory usage by embedding elements into the skiplist nodes. Change the `zskiplistNode` memory layout as follows:

```

Before
                 +-------------+
         +-----> | element-sds |
         |       +-------------+
         |
 +------------------+-------+------------------+---------+-----+---------+
 | element--pointer | score | backward-pointer | level-0 | ... | level-N |
 +------------------+-------+------------------+---------+-----+---------+



After
 +-------+------------------+---------+-----+---------+-------------+
 | score | backward-pointer | level-0 | ... | level-N | element-sds |
 +-------+------------------+---------+-----+---------+-------------+
```

Before the embedded SDS representation, we include one byte representing the size of the SDS header, i.e. the offset into the SDS representation where that actual string starts.

The memory saving is therefore one pointer minus one byte = 7 bytes per element, regardless of other factors such as element size or number of elements.

### Benchmark step

I generated the test data using the following lua script && cli command. And check memory usage using the `info` command.

**lua script**
```
local start_idx = tonumber(ARGV[1])
local end_idx = tonumber(ARGV[2])
local elem_count = tonumber(ARGV[3])

for i = start_idx, end_idx do
    local key = "zset:" .. string.format("%012d", i)
    local members = {}

    for j = 0, elem_count - 1 do
        table.insert(members, j)
        table.insert(members, "member:" .. j)
    end

    redis.call("ZADD", key, unpack(members))
end

return "OK: Created " .. (end_idx - start_idx + 1) .. " zsets"
```

**valkey-cli command**
`valkey-cli EVAL "$(cat create_zsets.lua)" 0 0 100000 ${ZSET_ELEMENT_NUM}`

### Benchmark result
|number of elements in a zset | memory usage before optimization | memory usage after optimization | change |
|-------|-------|-------|-------|
| 129 | 1047MB | 943MB | -9.9% |
| 256 |  2010MB|  1803MB| -10.3%|
| 512 |  3904MB|3483MB| -10.8%|


ranshid:
I like the direction of this. I would like, however, to share some of my initial thought on that:
While developing the new HFE, I also looked for a generic way to structure a mapped items in Valkey. For this reason I took the existing hashTypeEntry introduced in 8.1 and made it a generic "entry" so it can potentially be used for sets and sorted sets (once we change it to be able to work without a value). In the initial implementation of the new "entry" type I also included the ability to assign metadata to it in addition to the optional value pointer and expiration time. something "in the lines of" https://github.com/valkey-io/valkey/issues/640#issuecomment-2498799606 was supposed to allow linking an entry in skiplists.

Now I like your solution as it is optimal in memory saving. I would suggest to consider extending the entry as I initially did.
The downside of my proposal is that it might cost slightly more memory. For example: the field will have to use a minimal sds8 header for the entry bytes (so potentially 2 bytes overhead) + (lets say) 1 byte for the metadata size = 3 bytes overhead from your suggestion. However this way we might be able to get a more generic and reused code, and potentially also be easier to implement sorted-set field expiration (which TBH, I am not even sure how much is being requested).   

@zuiderkwast  / @madolson what are you thoughts?

zuiderkwast:
@ranshid Is the idea to use `entry.c` to embed the complete skiplist node? It seems like a larger refactoring. If we assume that we will not need sorted-set element expiration any time soon, then I think the simple approach done in this PR is better.

src/object.c
```diff
@@ -657,21 +657,11 @@ void dismissSetObject(robj *o, size_t size_hint) {
...
-        if (size_hint / zsl->length >= server.page_size) {
-            zskiplistNode *zn = zsl->tail;
-            while (zn != NULL) {
-                dismissSds(zn->ele);
-                zn = zn->backward;
-            }
-        }
-
+        /* In rare cases, the key of the skiplist may exceed the page size(typically 4KB),
+         * so we skip releasing the skiplist and only release the hashtable. */
```
zuiderkwast: This comment explains why we skip releasing the skiplist. The first line was confusing to me because it looks like it focuses on the rare cases where the keys are large. It would clearer to mention the normal case where the keys are small. We can say that it is usually smaller than a page size.

```suggestion
        /* The skiplist nodes are usually smaller than the page size (typically 4KB),
         * so we skip releasing the skiplist and only release the hashtable. */
```

ranshid:
> @ranshid Is the idea to use `entry.c` to embed the complete skiplist node? It seems like a larger refactoring. If we assume that we will not need sorted-set element expiration any time soon, then I think the simple approach done in this PR is better.

Yes. that was my point as well, but I wanted to point out this option so that we can take a correct decision. I am fine with going with the suggested solution. Will find the time to review the details.

hpatro:
@vitarb You might be interested in reviewing this.

chzhoo:
- Implementing a generic skiplist could be a major update. It might be better to postpone this until we have more well-defined use cases for it.
- I tend to favor keeping this PR focused and minimal, as it simplifies code reviews and reduces the risk of introducing new bugs. In fact, during the development of this PR, I also noted two additional potential optimizations:
     1. Changing the function signature of zslInsert from:`zskiplistNode *zslInsert(zskiplist *zsl, double score, sds ele)` to: `zskiplistNode *zslInsert(zskiplist *zsl, double score, const void* ele, size_t ele_len)`. This can reduce memory allocation and deallocation in certain scenarios.
     2. Further improving the cache locality of the optimized `zskiplistNode` layout. The optimized memory layout of `zskiplistNode` may not be optimal for cache locality, as frequently accessed fields such as `score`, `zslNodeHeight`, and `element-sds` might not reside within the same cache line.

   These optimizations were intentionally excluded from this PR to reduce both review complexity and the risk of introducing new bugs. If the current changes are acceptable, I would be happy to explore these enhancements in follow-up PRs over the coming weeks.

- Thank you for sharing your insights and your subsequent review feedback. @ranshid

zuiderkwast:
> * I tend to favor keeping this PR focused and minimal

@chzhoo I agree with this approach. :+1:

> 2\. Further improving the cache locality of the optimized `zskiplistNode` layout. The optimized memory layout of `zskiplistNode` may not be optimal for cache locality, as frequently accessed fields such as `score`, `zslNodeHeight`, and `element-sds` might not reside within the same cache line.

Regarding this (future optimization idea) I wouldn't worry too much about it, unless you have a smart idea that I haven't though about yet(?). The height of the node is usually low, like 3/4 of the nodes have only one level, 3/4 of the remaining nodes have only two levels, and so forth.

For nodes with only one level, we the score, backward-pointer, forward-pointer and span make up 32 bytes. We have 32 bytes left for the element in this cache line. It means we can fit an element of length 29 (excluding the sds header and nul-terminator). For nodes with two levels, we can fit an element of length 13 bytes. I think it seems acceptable, unless you have another smart idea. :)

chzhoo:
> Regarding this (future optimization idea) I wouldn't worry too much about it, unless you have a smart idea that I haven't though about yet(?). The height of the node is usually low, like 3/4 of the nodes have only one level, 3/4 of the remaining nodes have only two levels, and so forth.
> 
> For nodes with only one level, we the score, backward-pointer, forward-pointer and span make up 32 bytes. We have 32 bytes left for the element in this cache line. It means we can fit an element of length 29 (excluding the sds header and nul-terminator). For nodes with two levels, we can fit an element of length 13 bytes. I think it seems acceptable, unless you have another smart idea. :)

 The memory layout of `skiplistNode` in this PR:
```
 +-------+------------------+---------+-----+---------+-------------+
 | score | backward-pointer | level-0 | ... | level-N | element-sds |
 +-------+------------------+---------+-----+---------+-------------+
 |
 | 
 `-> zskiplistNode-pointer returned to the user
```

A memory layout with potential performance improvements，referencing [RocksDB's inline-skiplist](https://github.com/facebook/rocksdb/blob/b9957c991cae44959f96888369caf1b145398132/memtable/inlineskiplist.h#L1):
```
 +---------+-----+---------+-------+------------------+-------------+
 | level-n | ... | level-0 | score | backward-pointer | element-sds |
 +---------+-----+---------+-------+----------+-------+-------------+
                 |
                 | 
                 `-> zskiplistNode-pointer returned to the user
```

The reason I believe the second layout holds potential for optimization lies in the process of accessing the `element-sds`:

- In the second layout, the position of `element-sds` can be directly calculated using the `zskiplistNode-pointer` alone. 

- In contrast, the first layout requires first accessing the `zslNodeHeight` to determine the position of `element-sds`. Since `zslNodeHeight` and `element-sds` may not reside in the same cache line (though the probability is relatively low), the first layout carries a higher risk of cache misses.

- This approach of directly calculating the `element-sds` position via pointer arithmetic in the second layout reduces dependency on potentially uncached data, thereby improving memory access efficiency.

The above is merely my hypothesis. I will further investigate and validate this aspect, and look forward to continuing this discussion in a separate PR. @zuiderkwast

zuiderkwast (commented):
I have skimmed though this change. I didn't do a very careful review of all the details, but I believe this PR looks great and can soon be ready to be merged.

> If this change looks acceptable, I will look into implementing some tests in the next round.

Yes, please go ahead. What tests do you have in mind?

I do believe we have existing tests for sorted sets that will cover this code.

chzhoo:
> Yes, please go ahead. What tests do you have in mind?
> 
> I do believe we have existing tests for sorted sets that will cover this code.

I've looked into the test coverage, and you're correct that existing tests for sorted sets cover the most of the updated code. The one exception is the updated code in `module.c`, which currently isn't covered by tests. I will add the relevant unit tests for this part.

chzhoo:
> I've looked into the test coverage, and you're correct that existing tests for sorted sets cover the most of the updated code. The one exception is the updated code in `module.c`, which currently isn't covered by tests. I will add the relevant unit tests for this part.

The unit tests have been submitted, and the code is now in a review-ready state.

src/t_zset.c
```diff
@@ -115,15 +115,40 @@ static inline void zslSetNodeHeight(zskiplistNode *x, int height) {
...
+    size_t ele_sds_size = ele ? sdsReqSize(ele_sds_len, ele_sds_type) : 0;
+    /* Allocate enough space for the node, levels, and the element SDS string.
+     * The +1 is for the header size of the SDS string. */
+    zskiplistNode *zn = zmalloc(sizeof(*zn) + height * sizeof(struct zskiplistLevel) + 1 + ele_sds_size);
     zn->score = score;
-    zn->ele = ele;
     zslSetNodeHeight(zn, height);
+    if (ele) {
+        char *data = ((char *)(zn + 1)) + height * sizeof(struct zskiplistLevel);
+        *data++ = sdsHdrSize(ele_sds_type);
```
ranshid: Q: why do we need to maintain sds header size? can we just keep all pointers to the sds iteself and from there we can always reach the skiplist data... I think the most reasonable layout is?:
```
+------------------+--------+------+----------+------+---------------+
+ backward-pointer | level-N | ... | level-0 | score  |   element-sds |
+------------------+--------+------+----------+------+---------------+
                                                      |
                                                      |
                                                      Node ptr
```

zuiderkwast: @ranshid I agree this is probably the more optimal representation, but it requires larger code changes. We'll need to change all the direct accesses to the `zskiplistNode` structs such as `node->level[i]`. There are many such accesses in the code. Thus, this can be the refinement to the "future optimization ideas" outlined in https://github.com/valkey-io/valkey/pull/2508#issuecomment-3214944455.

For this simple PR, I think one byte for the header size is acceptable. We still get a lot of optimization without a large code change. I'm not even sure we will need the "future optimization" at all. It might not be worth the effort.

src/t_zset.c
```diff
@@ -142,11 +167,9 @@ zskiplist *zslCreate(void) {
     return zsl;
 }
 
-/* Free the specified skiplist node. The referenced SDS string representation
- * of the element is freed too, unless node->ele is set to NULL before calling
- * this function. */
+/* Free the specified skiplist node. No need to explicitly free the SDS string 'ele',
+ * as it is embedded within the structure. */
```
zuiderkwast: (regarding zslFreeNode) The comment above the function is API documentation for the caller of the function. No need to explain how it works internally.

That the element is embedded in the same allocation is implementation details. If you think it's important to mention it somewhere, we can add a comment about it inside the body of the function instead.

```suggestion
/* Free the specified skiplist node. */
```

src/t_zset.c
```diff
@@ -247,8 +269,7 @@ static zskiplistNode *zslInsertNode(zskiplist *zsl, zskiplistNode *node) {
 }
 
 /* Insert a new node in the skiplist. Assumes the element does not already
- * exist (up to the caller to enforce that). The skiplist takes ownership
- * of the passed SDS string 'ele'. */
+ * exist (up to the caller to enforce that). */
```
zuiderkwast: (regarding zslInsert) Mention that the element is copied?

```suggestion
 * exist (up to the caller to enforce that). The string 'ele' is copied. */
```

src/t_zset.c
```diff
@@ -115,15 +115,40 @@ static inline void zslSetNodeHeight(zskiplistNode *x, int height) {
...
+ * By embedding elements and levels into the skiplist nodes,
+ * we achieve better cache-friendliness and a more compact memory structure.
+ *
+ * The memory layout is as follows:
+ *
+ *     +-------+------------------+---------+-----+---------+-------------+
+ *     | score | backward-pointer | level-0 | ... | level-N | element-sds |
+ *     +-------+------------------+---------+-----+---------+-------------+
+ */
 static zskiplistNode *zslCreateNode(int height, double score, sds ele) {
```
zuiderkwast: For functions that copy the sds, it's possible to use the `const_sds` type.

```suggestion
static zskiplistNode *zslCreateNode(int height, double score, const_sds ele) {
```

zuiderkwast (commented):
Posted a few comments.

It's a bit late to get this into 9.0, but we can get it into 9.1 instead. Added it to the plan.

Regarding the benchmarks, I believe the memory saving is exactly 7 bytes per element, regardless of other factors, right? (remove one pointer and add one byte for sds hdr size). The relative savings then very much depends on the size of the elements then, not very much the number of elements in a sorted set. I can see in your Lua script you used 12 bytes elements in all tests.

chzhoo:
> Regarding the benchmarks, I believe the memory saving is exactly 7 bytes per element, regardless of other factors, right?

Yes, you are right.

src/t_zset.c
ranshid: > @ranshid I agree this is probably the more optimal representation, but it requires larger code changes. We'll need to change all the direct accesses to the `zskiplistNode` structs such as `node->level[i]`. There are many such accesses in the code. Thus, this can be the refinement to the "future optimization ideas" outlined in [#2508 (comment)](https://github.com/valkey-io/valkey/pull/2508#issuecomment-3214944455).
> 
> For this simple PR, I think one byte for the header size is acceptable. We still get a lot of optimization without a large code change. I'm not even sure we will need the "future optimization" at all. It might not be worth the effort.

@zuiderkwast I agree this will require more changes. I also think what I suggested is slightly different than the suggestion in [#2508 (comment)](https://github.com/valkey-io/valkey/pull/2508#issuecomment-3214944455) as this suggestion also save the extra byte used to calculate the sds header size (which I agree is minor). but if we ARE targeting this for 9.1 I am not sure why we cannot put the extra effort to build this as we think would be best.
Given said that, we can do this in 2 stages even for 9.1. We can first introduce the naïve change and also followup with a second PR to optimize it. I would still think we need to have a performance benchmark results so we can evaluate the potential impact.

ranshid (commented):
.

src/defrag.c
```diff
@@ -283,7 +279,7 @@ static void activeDefragZsetNode(void *privdata, void *entry_ref) {
         zskiplistNode *next = x->level[i].forward;
         while (next &&
                (next->score < score ||
-                (next->score == score && sdscmp(next->ele, ele) < 0))) {
+                (next->score == score && sdscmp(zslGetNodeElement(next), ele) < 0))) {
```
ranshid: Not related to your PR, but I think this code could be optimized.
There is no real need to prepare the update array in case the node is not really relocated. we already know the old node pointer so there is no need to do string comparisons. I think we can just compare the next to the old allocation.

src/t_zset.c
```diff
@@ -327,10 +347,8 @@ static zskiplistNode *zslUpdateScore(zskiplist *zsl, zskiplistNode *node, double
 
     zslDeleteNode(zsl, node, update);
     /* update pointer inside hashtable with new node */
-    zskiplistNode *new_node = zslInsert(zsl, newscore, node->ele);
-    /* We reused the old node->ele SDS string, free the node now
-     * since zslInsert created a new node */
-    node->ele = NULL;
+    sds ele = zslGetNodeElement(node);
+    zskiplistNode *new_node = zslInsert(zsl, newscore, ele);
```
ranshid: question: can't we just use zslInsertNode with the already existing node?
maybe now that we will have to relocate potentially large strings it will be proven more useful?

src/t_zset.c
zuiderkwast: > I also think what I suggested is slightly different than the suggestion in [#2508](https://github.com/valkey-io/valkey/pull/2508#issuecomment-3214944455)

Yes that's right.

> if we ARE targeting this for 9.1 I am not sure why we cannot put the extra effort to build this as we think would be best

We can, but it's just a large effort for only one byte. I'm not sure if it's worth it. If you want to review it, fine. I just think first PR is small and achieves most of the benefit.

> I would still think we need to have a performance benchmark results so we can evaluate the potential impact.

Sure. I can guess the outcome: No performance impact, except for insertion of very large elements. In this case we need to copy large strings.

I don't know if we care much about very large strings in sorted sets. Do we? For key names and hash fields, we don't. We already embed them. Using large keys names is an anti-pattern.

src/t_zset.c
```diff
@@ -456,8 +474,9 @@ static unsigned long zslDeleteRangeByScore(zskiplist *zsl, zrangespec *range, ha
     while (x && zslValueLteMax(x->score, range)) {
         zskiplistNode *next = x->level[0].forward;
         zslDeleteNode(zsl, x, update);
-        hashtableDelete(ht, x->ele);
-        zslFreeNode(x); /* Here is where x->ele is actually released. */
+        sds ele = zslGetNodeElement(x);
+        hashtableDelete(ht, ele);
```
ranshid: nit: although this type has no entry destructor, I would be happy if we could explicitly use hashtablepop instead just to be clear that we are NOT planning to delete the element.

ranshid (commented):
some small comments

src/t_zset.c
ranshid: > We can, but it's just a large effort for only one byte. I'm not sure if it's worth it. If you want to review it, fine. I just think first PR is small and achieves most of the benefit.
> 
> > I would still think we need to have a performance benchmark results so we can evaluate the potential impact.
> 
> Sure. I can guess the outcome: No performance impact, except for insertion of very large elements. In this case we need to copy large strings.
> 
> I don't know if we care much about very large strings in sorted sets. Do we? For key names and hash fields, we don't. We already embed them. Using large keys names is an anti-pattern.

I agree there is no expected impact, but some things (like the fact that the score is drifted away from the sds pointer) might have some relevancy although it might even up with the embedded sds improvement. I just think that we need to take care about presenting potential regressions during development since later it makes the bisecting job much more difficult.   

Also in this case there is no "keyname". The zset elements are practically values of something and might actually be large. For example very long URLs... In AWS I know sometimes it is used to store ARNs which can reach 2K long...

ranshid:
> **Benchmark ran on this commit:** [`0d1cc2b`](https://github.com/valkey-io/valkey/commit/0d1cc2bb6a9a09ce86ee8dc4c7139ad1c469ccfa)
> 
> # Benchmark Comparison by Command
> ## GET | cluster disabled | tls disabled
> Metric	Baseline	PR	Diff	% Change
> rps	968054.19	982318.25	14264.06	+1.47%
> latency_avg_ms	0.46	0.45	-0.01	-1.53%
> latency_p50_ms	0.44	0.43	-0.01	-1.82%
> latency_p95_ms	0.61	0.61	-0.01	-1.30%
> latency_p99_ms	0.73	0.72	-0.01	-1.10%
> ## GET | cluster enabled | tls disabled
> Metric	Baseline	PR	Diff	% Change
> rps	836120.44	848896.44	12776.00	+1.53%
> latency_avg_ms	0.54	0.53	-0.01	-1.85%
> latency_p50_ms	0.54	0.53	-0.01	-1.50%
> latency_p95_ms	0.70	0.69	-0.01	-1.14%
> latency_p99_ms	0.76	0.75	-0.01	-1.05%
> ## MGET | cluster disabled | tls disabled
> Metric	Baseline	PR	Diff	% Change
> rps	313479.62	318877.56	5397.94	+1.72%
> latency_avg_ms	1.42	1.39	-0.03	-2.04%
> latency_p50_ms	1.42	1.38	-0.03	-2.26%
> latency_p95_ms	1.77	1.75	-0.02	-1.35%
> latency_p99_ms	1.85	1.82	-0.02	-1.30%
> ## MSET | cluster disabled | tls disabled
> Metric	Baseline	PR	Diff	% Change
> rps	216262.97	216450.22	187.25	+0.09%
> latency_avg_ms	2.22	2.22	-0.01	-0.27%
> latency_p50_ms	2.33	2.32	-0.01	-0.34%
> latency_p95_ms	2.56	2.55	-0.01	-0.31%
> latency_p99_ms	2.63	2.62	-0.01	-0.30%
> ## SET | cluster disabled | tls disabled
> Metric	Baseline	PR	Diff	% Change
> rps	858369.12	827129.88	-31239.24	-3.64%
> latency_avg_ms	0.53	0.55	0.02	+4.00%
> latency_p50_ms	0.52	0.54	0.02	+4.62%
> latency_p95_ms	0.69	0.71	0.02	+3.49%
> latency_p99_ms	0.73	0.75	0.02	+2.18%
> ## SET | cluster enabled | tls disabled
> Metric	Baseline	PR	Diff	% Change
> rps	659195.81	667111.44	7915.63	+1.20%
> latency_avg_ms	0.70	0.69	-0.01	-1.28%
> latency_p50_ms	0.71	0.69	-0.02	-2.25%
> latency_p95_ms	0.89	0.88	-0.01	-0.90%
> latency_p99_ms	1.14	1.11	-0.02	-2.11%

Oh... no zset benchmark yet :(

src/t_zset.c
chzhoo: > Given said that, we can do this in 2 stages even for 9.1. We can first introduce the naïve change and also followup with a second PR to optimize it. I would still think we need to have a performance benchmark results so we can evaluate the potential impact.

- Are you concerned about a significant performance drop in `ZADD` command? I can run benchmark tests later to check.
- I'm not very familiar with the PR merging process yet. If this PR is targeting version 9.1, will it take a relatively long time to get merged?

madolson:
@ranshid its easier to add. It's only a few lines and we can re-run the test.

ranshid:
> @ranshid its easier to add. It's only a few lines and we can re-run the test.

@madolson depends on how you refer to "few lines"...
Since the current benchmarking framework is built on is built on existing valkey-benchmark testing scenarios it is missing some of the required commands like ZRANGE and ZSCORE. In fact the current benchmark define SPOP and ZPOP as read commands [here](https://github.com/valkey-io/valkey-perf-benchmark/blob/main/valkey_benchmark.py#L22) (which is not true and might impact the test as the original fillup will be cleaned)   

So I feel some things needs to be done:
1. Add ZSCORE and ZRANGE test scenarios to valkey-benchmark. Alternatively we can maybe modify the performance framework to accept a specific command instead of running a test scenario.
2. Fix the performance-benchmark to not use SPOP and ZPOPMIN as READ commands.
3. Add the new command scenarios to performance-benchmark
4. Modify the workflow config to use more data size values + support more commands

I will try to evaluate if I can simply so that or I will open up issues to followup on that

ranshid:
> > @ranshid its easier to add. It's only a few lines and we can re-run the test.
> 
> @madolson depends on how you refer to "few lines"... Since the current benchmarking framework is built on is built on existing valkey-benchmark testing scenarios it is missing some of the required commands like ZRANGE and ZSCORE. In fact the current benchmark define SPOP and ZPOP as read commands [here](https://github.com/valkey-io/valkey-perf-benchmark/blob/main/valkey_benchmark.py#L22) (which is not true and might impact the test as the original fillup will be cleaned)
> 
> So I feel some things needs to be done:
> 
> 1. Add ZSCORE and ZRANGE test scenarios to valkey-benchmark. Alternatively we can maybe modify the performance framework to accept a specific command instead of running a test scenario.
> 2. Fix the performance-benchmark to not use SPOP and ZPOPMIN as READ commands.
> 3. Add the new command scenarios to performance-benchmark
> 4. Modify the workflow config to use more data size values + support more commands
> 
> I will try to evaluate if I can simply so that or I will open up issues to followup on that

@madolson indeed few line changes. take a look at:
https://github.com/valkey-io/valkey/pull/2575 and https://github.com/valkey-io/valkey-perf-benchmark/pull/7

src/t_zset.c
zuiderkwast: > Are you concerned about a significant performance drop in `ZADD` command? I can run benchmark tests later to check.

It's one of the concerns yes, for large element sizes such as 2KB. Let's run a benchmark with this setup. It's possible to specify commands explicitly on the valkey-benchmark command line.

Additionally, running some benchmark for various z-commands like zscore and zrange to verify that they don't get slower. @ranshid can you line out what specific benchmarks you would want to see? I think we can do it manually rather than waiting for the automated benchmarking to be in place. I don't like PRs hanging indefinitely.

> I'm not very familiar with the PR merging process yet. If this PR is targeting version 9.1, will it take a relatively long time to get merged?

We can merge this PR whenever we feel that it's ready. We don't have a feature freeze for 9.0, because we have already created the 9.0 branch.

ranshid:
I might not be able to review till Sunday - so apologies about that (have to take a short vacation)

src/object.c
```diff
@@ -657,21 +657,11 @@ void dismissSetObject(robj *o, size_t size_hint) {
...
-            zskiplistNode *zn = zsl->tail;
-            while (zn != NULL) {
-                dismissSds(zn->ele);
-                zn = zn->backward;
-            }
-        }
-
+        /* The skiplist nodes are usually smaller than the page size (typically 4KB),
+         * so we skip releasing the skiplist and only release the hashtable. */
         dismissHashtable(zs->ht);
```
zuiderkwast, line 676: There is an O(1) check here to estimate if it makes sense to dismiss the individual nodes. I agree in 99% of the times this is false, but for the few users that might actually store 4KB+ strings in a sorted sets, removing this logic can cause a regression.

I prefer that we avoid dropping this dismiss logic and instead adjust it to embedded strings, which means dismiss the skiplist node allocations. We can pass 0 as the size hint instead of trying to calculate it.

```C
                dismissMemory(zn, 0);
```

src/server.h
```diff
@@ -1402,7 +1402,6 @@ struct sharedObjectsStruct {
 
 /* ZSETs use a specialized version of Skiplists */
 typedef struct zskiplistNode {
-    sds ele;
```
zuiderkwast, line 1426: Add a comment somewhere in this struct definition, for example after the `level[]` array where the embedded element is located in memory, explaining that there is an embedded sds header length (1 byte) and an embedded sds string here.

src/t_zset.c
```diff
@@ -115,15 +115,40 @@ static inline void zslSetNodeHeight(zskiplistNode *x, int height) {
 }
 
 /* Create a skiplist node with the specified number of levels.
- * The SDS string 'ele' is referenced by the node after the call. */
-static zskiplistNode *zslCreateNode(int height, double score, sds ele) {
-    zskiplistNode *zn = zmalloc(sizeof(*zn) + height * sizeof(struct zskiplistLevel));
+ * By embedding elements and levels into the skiplist nodes,
+ * we achieve better cache-friendliness and a more compact memory structure.
```
zuiderkwast: Better and more compact than what? Once this is merged, the old representation is history.

```suggestion
 * we achieve good cache-friendliness and a compact memory structure.
```

src/t_zset.c
```diff
@@ -115,15 +115,40 @@ static inline void zslSetNodeHeight(zskiplistNode *x, int height) {
...
+ *     +-------+------------------+---------+-----+---------+-------------+
+ *     | score | backward-pointer | level-0 | ... | level-N | element-sds |
+ *     +-------+------------------+---------+-----+---------+-------------+
+ */
+static zskiplistNode *zslCreateNode(int height, double score, const_sds ele) {
+    size_t ele_sds_len = ele ? sdslen(ele) : 0;
+    char ele_sds_type = ele ? sdsReqType(ele_sds_len) : 0;
+    size_t ele_sds_size = ele ? sdsReqSize(ele_sds_len, ele_sds_type) : 0;
+    /* Allocate enough space for the node, levels, and the element SDS string.
+     * The +1 is for the header size of the SDS string. */
```
zuiderkwast: This extra byte deserves some more explanation. We've used it in other places where we embedded sds but the reader of this code might not be aware of that. Something like this?

```suggestion
    /* Allocate enough space for the node, levels, and the element SDS string.
     * Before the embedded sds, we include one extra byte representing the sds
     * header size, which is the offset into the embedded sds data where the
     * string content starts. */
```

src/t_zset.c
```diff
@@ -115,15 +115,40 @@ static inline void zslSetNodeHeight(zskiplistNode *x, int height) {
...
     zslSetNodeHeight(zn, height);
+    if (ele) {
+        char *data = ((char *)(zn + 1)) + height * sizeof(struct zskiplistLevel);
+        *data++ = sdsHdrSize(ele_sds_type);
+        sdswrite(data, ele_sds_size, ele_sds_type, ele, ele_sds_len);
+    }
     return zn;
 }
 
+sds zslGetNodeElement(const zskiplistNode *x) {
```
zuiderkwast, line 154: Add a line or two documenting the function, like we have for the existing functions. "Returns the embedded element" or something better.

Bts, it takes a `const` node but it returns a non-const sds. Shouldn't it return a `const_sds` instead? (We added the `const_sds` type some year ago.) Ideally. But if changing these constness types causes a lot of trouble, we can leave it as is.

src/t_zset.c
```diff
@@ -188,12 +210,11 @@ static int zslCompareNodes(const zskiplistNode *a, const zskiplistNode *b) {
...
 
-    return sdscmp(a->ele, b->ele);
+    return sdscmp(zslGetNodeElement(a), zslGetNodeElement(b));
 }
 
 /* Insert a node in the skiplist. Assumes the element does not already exist in
- * the skiplist (up to the caller to enforce that). The skiplist takes ownership
- * of the passed node. */
+ * the skiplist (up to the caller to enforce that). */
 static zskiplistNode *zslInsertNode(zskiplist *zsl, zskiplistNode *node) {
```
zuiderkwast, line 233: Why change this comment? The skiplist does take ownership of the passed node.

src/t_zset.c
```diff
@@ -472,18 +489,19 @@ static unsigned long zslDeleteRangeByLex(zskiplist *zsl, zlexrangespec *range, h
 
     x = zsl->header;
     for (i = zsl->level - 1; i >= 0; i--) {
-        while (x->level[i].forward && !zslLexValueGteMin(x->level[i].forward->ele, range)) x = x->level[i].forward;
+        while (x->level[i].forward &&
+               !zslLexValueGteMin(zslGetNodeElement(x->level[i].forward), range)) x = x->level[i].forward;
```
zuiderkwast: A two-lines while loop without braces is slightly hard to read. I suggest expanding to this:

```suggestion
        while (x->level[i].forward &&
               !zslLexValueGteMin(zslGetNodeElement(x->level[i].forward), range)) {
            x = x->level[i].forward;
        }
```

zuiderkwast (commented):
This looks great to me. I only have some minor comments. Sorry for the delay.

src/object.c
chzhoo: The new code has retained the dismiss logic.

src/server.h
chzhoo: Comments have been added following the definition of the `level[]` array.

src/t_zset.c
chzhoo: > But if changing these constness types causes a lot of trouble, we can leave it as is.

This is the main obstacle preventing the change from `sds` to `const_sds`.

src/t_zset.c
chzhoo: This was my mistake. I have reverted this comments back to their original state.

src/t_zset.c
chzhoo: I've fixed.

src/defrag.c
ranshid: any reason why we need to do sdscmp here?

src/t_zset.c
ranshid: yeh the downsides of C. I am fine with doing this

src/geo.c
```diff
@@ -825,7 +826,6 @@ void georadiusGeneric(client *c, int srcKeyIndex, int flags) {
             totelelen += elelen;
             znode = zslInsert(zs->zsl, score, gp->member);
             serverAssert(hashtableAdd(zs->ht, znode));
-            gp->member = NULL;
```
ranshid, line 828: question: can't we just sdsfree the old gp->member instead of saving it for the cleanup?

src/t_zset.c
```diff
@@ -251,8 +276,7 @@ static zskiplistNode *zslInsertNode(zskiplist *zsl, zskiplistNode *node) {
 }
 
 /* Insert a new node in the skiplist. Assumes the element does not already
- * exist (up to the caller to enforce that). The skiplist takes ownership
- * of the passed SDS string 'ele'. */
+ * exist (up to the caller to enforce that). The string 'ele' is copied. */
```
ranshid, line 286: so maybe we can use a const_sds to imply that?

tests/modules/zset.c
```diff
@@ -2,6 +2,8 @@
 #include <math.h>
 #include <errno.h>
 
+#define UNUSED(V) ((void)V)
```
ranshid: why not include util.h?

ranshid (approved):
I think the code is mostly good.
Left some minor comments I could observe during a quick look.
I give an approval to avoid extra delay.

ranshid:
@chzhoo please also see my small comment of past discussion: https://github.com/valkey-io/valkey/pull/2508#discussion_r2518346728

src/geo.c
chzhoo: I understand that both approaches are acceptable and functionally equivalent. Just to clarify, am I missing something that prompted your question?

chzhoo:
> @chzhoo please also see my small comment of past discussion: [#2508 (comment)](https://github.com/valkey-io/valkey/pull/2508#discussion_r2518346728)

@ranshid I've fixed. Please take a look.

tests/modules/zset.c
chzhoo: Previously, we referenced the style of files in `./test/modules/,` but now we have switched to using `#include "util.h"`.

src/t_zset.c
chzhoo: I have fixed.

src/geo.c
ranshid: I am not sure you are missing anything. I just thought that free it now and nullify the member would better be aligned with what we did before

src/defrag.c
```diff
@@ -283,7 +282,7 @@ static void activeDefragZsetNode(void *privdata, void *entry_ref) {
         zskiplistNode *next = x->level[i].forward;
         while (next &&
                (next->score < score ||
-                (next->score == score && sdscmp(zslGetNodeElement(next), ele) < 0))) {
+                (next->score == score && next != node))) {
```
zuiderkwast, line 285: If two elements have the same score, they  are sorted lexicographically. That's what the sdscmp < 0 is for. ~~I don't think we can remove it.~~

Maybe we can? This is the defrag code...

ranshid: but in the scope of defrag we are not changing the order right?

ranshid: @zuiderkwast read a cached version of your comment. yes. I think in this context just matching pointers is enough

zuiderkwast: Yeah, I was wrong.

tests/modules/zset.c
zuiderkwast: > why not include util.h?

Because the modules are supposed to be independent of valkey's server code. They're not supposed to access the internals directly. They're only supposed to include `valkeymodule.h` and access valkey through the module API.

`util.h` maybe doesn't access anything in the server internals, but not including any valkey headers at all is cleaner and a better example for users who copy-paste these modules when they're creating their own modules. I think it's better to avoid including util.h in these modules.

chzhoo: > `util.h` maybe doesn't access anything in the server internals, but not including any valkey headers at all is cleaner and a better example for users who copy-paste these modules when they're creating their own modules.

Thank you for pointing that out. I have reverted the code to no longer include `util.h`.

src/geo.c
chzhoo: I have fixed.

src/sort.c
```diff
@@ -449,9 +449,11 @@ void sortCommandGeneric(client *c, int readonly) {
         hashtableIterator iter;
         hashtableInitIterator(&iter, ht, 0);
         void *next;
+        sds sdsele;
```
rainsupreme: nit: `sdsele` could be declared inside the while loop. It won't be slower on modern compilers. :)

zuiderkwast (approved):

src/t_zset.c
```diff
@@ -330,13 +354,9 @@ static zskiplistNode *zslUpdateScore(zskiplist *zsl, zskiplistNode *node, double
...
     zslDeleteNode(zsl, node, update);
-    /* update pointer inside hashtable with new node */
-    zskiplistNode *new_node = zslInsert(zsl, newscore, node->ele);
-    /* We reused the old node->ele SDS string, free the node now
-     * since zslInsert created a new node */
-    node->ele = NULL;
-    zslFreeNode(node);
-    return new_node;
+    node->score = newscore; /* reuse existing node to avoid memory allocation */
+    zslInsertNode(zsl, node);
```
rainsupreme, line 365: with this change the skiplist height can't change any more, but either way it has the same random height and follows the same distribution, so I suppose this is fine! 🤔

src/t_zset.c
```diff
@@ -115,15 +115,43 @@ static inline void zslSetNodeHeight(zskiplistNode *x, int height) {
...
- * The SDS string 'ele' is referenced by the node after the call. */
-static zskiplistNode *zslCreateNode(int height, double score, sds ele) {
-    zskiplistNode *zn = zmalloc(sizeof(*zn) + height * sizeof(struct zskiplistLevel));
+ * By embedding elements and levels into the skiplist nodes,
+ * we achieve good cache-friendliness and a compact memory structure.
+ *
+ * The memory layout is as follows:
+ *
+ *     +-------+------------------+---------+-----+---------+-------------+
+ *     | score | backward-pointer | level-0 | ... | level-N | element-sds |
```
rainsupreme: perhaps show the 1-byte sds header size field in the diagram?

src/t_zset.c
```diff
@@ -115,15 +115,43 @@ static inline void zslSetNodeHeight(zskiplistNode *x, int height) {
...
+ */
+static zskiplistNode *zslCreateNode(int height, double score, const_sds ele) {
+    size_t ele_sds_len = ele ? sdslen(ele) : 0;
+    char ele_sds_type = ele ? sdsReqType(ele_sds_len) : 0;
+    size_t ele_sds_size = ele ? sdsReqSize(ele_sds_len, ele_sds_type) : 0;
+    /* Allocate enough space for the node, levels, and the element sds.
+     * Before the embedded sds, we include one extra byte representing the sds
+     * header size, which is the offset into the embedded sds data where the
+     * string content starts. */
+    zskiplistNode *zn = zmalloc(sizeof(*zn) + height * sizeof(struct zskiplistLevel) + 1 + ele_sds_size);
```
rainsupreme, line 137: When there is no `ele` this still adds 1 byte for the sds header size

rainsupreme (approved):
Just a few nits. Overall LGTM!

src/sort.c
chzhoo: I have fixed.

src/t_zset.c
chzhoo: Your suggestion is better. I have fixed.

src/t_zset.c
chzhoo: ~~`ele` always exists. I'm wondering if I missed something that prompted your question.~~

zuiderkwast: The lines earlier in this functions hint that `ele` can be NULL.

```C
    size_t ele_sds_len = ele ? sdslen(ele) : 0;
    char ele_sds_type = ele ? sdsReqType(ele_sds_len) : 0;
    size_t ele_sds_size = ele ? sdsReqSize(ele_sds_len, ele_sds_type) : 0;
```

If this is defensive code and ele is never NULL, then we should change to an `serverAssert(ele != NULL)`.

chzhoo: I missed that when creating the head node of the skiplist, the `ele` is NULL.

zuiderkwast: Ah, makes sense. The head node has no element. It's OK that we store one byte sds_header_size = 0, right? It indicates that there is no embedded element.

chzhoo: > Ah, makes sense. The head node has no element. It's OK that we store one byte sds_header_size = 0, right? It indicates that there is no embedded element.

As the header node, keeping or removing the `sds_header_size` is acceptable since the header node will never be accessed for its element value.  I prefer to remove the `sds_header_size` to save memory.

src/t_zset.c
```diff
@@ -123,16 +123,19 @@ static inline void zslSetNodeHeight(zskiplistNode *x, int height) {
  *   +-------+------------------+---------+-----+---------+-----------------+-------------+
  *   | score | backward-pointer | level-0 | ... | level-N | sds-header-size | element-sds |
  *   +-------+------------------+---------+-----+---------+-----------------+-------------+
+ *   
+ *   sds-header-size and element-sds are only valid for non-header nodes.
```
zuiderkwast, line 127: There is now some difference between header and non-header nodes. Maybe we should create a separate function for the header nodes? Then we can require that `ele` is always set here and remove the conditional code that checks if ele is NULL or not.

zslCreateHeaderNode is very simple:

```C
static zskiplistNode *zslCreateHeaderNode(void) {
    /* Allocate enough space for the node and levels. */
    zskiplistNode *zn = zmalloc(sizeof(*zn) + ZSKIPLIST_MAXLEVEL * sizeof(struct zskiplistLevel));
    zslSetNodeHeight(zn, ZSKIPLIST_MAXLEVEL);
    return zn;
}
```

chzhoo: I have fixed.

zuiderkwast (approved):
Thanks!

PR #2509, opened 2025-08-19
fix hsetex handling of wrong number of fields

ranshid:
currently hsetex is verifying the number of fields matches the provided number of fields by using div. instead it can match to the multiplication in order to prevent rounding the verified value down.

### DIFF
--- a/tests/unit/hashexpire.tcl
+++ b/tests/unit/hashexpire.tcl
@@ -624,9 +624,8 @@ start_server {tags {"hashexpire"}} {
     } {ERR *}
 
     test {HSETEX PX - mismatched field/value count} {
-         catch {r HSETEX myhash PX 100 FIELDS 1 field1 val1 extra} e
-         set e
-    } {ERR numfields should be greater than 0 and match the provided number of fields}
+         assert_error {ERR numfields should be greater than 0 and match the provided number of fields} {r HSETEX myhash PX 100 FIELDS 1 field1 val1 extra}
+    } 
 
 
     ## FNX/FXX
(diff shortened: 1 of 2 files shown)


tests/unit/hashexpire.tcl
```diff
@@ -623,10 +623,10 @@ start_server {tags {"hashexpire"}} {
...
     } {ERR *}
 
-    # test {HSETEX PX - mismatched field/value count} {
-    #     catch {r HSETEX myhash PX 100 FIELDS 2 field1 val1} e
-    #     set e
-    # } {ERR wrong number of arguments for 'hsetex' command}
+    test {HSETEX PX - mismatched field/value count} {
+         catch {r HSETEX myhash PX 100 FIELDS 1 field1 val1 extra} e
+         set e
+    } {ERR numfields should be greater than 0 and match the provided number of fields}
```
madolson:
```suggestion
         assert_error {ERR numfields should be greater than 0 and match the provided number of fields} {r HSETEX myhash PX 100 FIELDS 1 field1 val1 extra}
    } 
```
Btw, this will throw a more useful error, since it will say command X didn't produce the desired output.

madolson (approved):

tests/unit/hashexpire.tcl
zuiderkwast: We should align this message with other commands that return similar errors.

ranshid: >We should align this message with other commands that return similar errors.

Well, eval, for example returns `Number of keys can't be greater than number of args` and ZUNION flavors (as like BITOPS), for another example, just return syntax error. Originally I just did the syntax error return (as I was probably lazy) and then I placed this which is more informative. We can improve that (on a separate PR IMO)

zuiderkwast (approved):

enjoy-binbin (approved):

PR #2510, opened 2025-08-19
Atomically update cluster and replication layer while marking self node as primary

### LINKED ISSUE #2460: [BUG] Divergence in cluster and replication layer with node promotion in empty shard
**Describe the bug**

Setup:

* Three nodes in a new shard (Let's say Node 1, Node 2, Node 3)
* No slots assigned
* Replica migration disabled
* Node 2 is picked as the primary in the shard
* Node 1 has higher epoch than Node 2
* Engine version - 8.1

Image

**Step 6** is the interesting part in the above sequence diagram which is the source of the failover. It occurs within `clusterUpdateSlotsConfigWith`.  The primary receives a message which satisfies the condition that there is another node in this empty shard with an higher epoch value and leads to the demotion of the primary.

`clusterUpdateSlotsConfigWith` can originate from two sources i.e. while processing `PING` / `PONG` / `MEET` (https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L3897) or `UPDATE` (https://github.com/valkey-io/valkey/blob/unstable/src/cluster_legacy.c#L4016) message. The logs on node 1 during the same period indicates it was a replica and haven't yet established the source of the message to Node 2).

The below condition satisfies 

https://github.com/valkey-io/valkey/blob/cc465383c419a0cbfdd60a17256b7a0fcffe21c8/src/cluster_legacy.c#L2942-L2949

and we find logs from few lines below in our logs stating that `Node 2` has demoted itself and it broadcasts this information to all other nodes.

https://github.com/valkey-io/valkey/blob/cc465383c419a0cbfdd60a17256b7a0fcffe21c8/src/cluster_legacy.c#L2970-L2981

Further, `node 1` receives this broadcast message i.e. `node 1` itself is the primary of the shard and does process it and changes the cluster topology information. However, the replication layer is not updated and `node 1` still continues replicating from `node 2` which leads to divergence.

https://github.com/valkey-io/valkey/blob/02d7ee08489fe34f853ffccce9057dea6f03d957/src/cluster_legacy.c#L3551-L3562

**Expected behavior**

Replication layer and cluster layer to be in sync

hpatro:
Fixes: #2460

With this change we avoid divergence in cluster and replication layer view. I've observed node can be marked as primary in cluster while it can be marked as replica in replication layer view and have a active replication link. Without this change, we used to end up in a invalid replica chain in replication layer.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -1453,7 +1453,6 @@ void clusterReset(int hard) {
     if (nodeIsReplica(myself)) {
         was_replica = 1;
         clusterSetNodeAsPrimary(myself);
-        replicationUnsetPrimary();
         flushAllDataAndResetRDB(server.lazyfree_lazy_user_flush ? EMPTYDB_ASYNC : EMPTYDB_NO_FLAGS);
     }
 
@@ -5085,7 +5084,6 @@ void clusterFailoverReplaceYourPrimary(void) {
 
     /* 1) Turn this node into a primary. */
     clusterSetNodeAsPrimary(myself);
-    replicationUnsetPrimary();
 
     int remaining = old_primary->numslots;
     /* 2) Claim all the slots assigned to our primary. */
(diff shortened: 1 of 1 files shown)


hpatro:
This makes the code correct around updating state for cluster and replication layer together. I was not able to come up with a test case to produce this scenario.

zuiderkwast (approved):
Looks reasonable to me.

Until now, how and when does a node stop replicating after becoming a primary e.g. after a failover?

sarthakaggarwal97 (approved):

src/cluster_legacy.c
```diff
@@ -2731,6 +2731,8 @@ int nodeUpdateAddressIfNeeded(clusterNode *node, clusterLink *link, clusterMsg *
 void clusterSetNodeAsPrimary(clusterNode *n) {
```
madolson, line 2730: There are other places we replicationUnsetPrimary after clusterSetNodeAsPrimary. I suppose if the expectation is that we just need it here, I think we can remove the other places we call it.

hpatro:
> Until now, how and when does a node stop replicating after becoming a primary e.g. after a failover?

`replicationUnsetPrimary` was invoked independently . It's an idempotent method. One example:

https://github.com/valkey-io/valkey/blob/eae23babe63e4e32498da3d16e4cb368354c044b/src/cluster_legacy.c#L5096-L5106

madolson (approved):

PR #2512, opened 2025-08-19
Update pong_received time via gossip only if the node is healthy from our view

hpatro:
This code change aligns with the comment mentioned above.

We should let direct communication between node happen when they are not in steady state. We can use the piggybacked information from the gossip section otherwise.

Note: `pong_received` field is used to communicate with the other node within node timeout once.

### DIFF
--- a/src/cluster_legacy.c
+++ b/src/cluster_legacy.c
@@ -2580,7 +2580,9 @@ void clusterProcessGossipSection(clusterMsg *hdr, clusterLink *link) {
              * we have no pending ping for the node, nor we have failure
              * reports for this node, update the last pong time with the
              * one we see from the other nodes. */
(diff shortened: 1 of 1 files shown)


madolson (approved):

sarthakaggarwal97 (approved):

src/cluster_legacy.c
```diff
@@ -2580,7 +2580,9 @@ void clusterProcessGossipSection(clusterMsg *hdr, clusterLink *link) {
              * we have no pending ping for the node, nor we have failure
              * reports for this node, update the last pong time with the
              * one we see from the other nodes. */
```
sungming2, line 2582: I think the original comment was incorrect. Do we need to update the comment as well to reflect both our POV and the other node’s POV

hpatro: I would say it's incomplete now.

PR #2513, opened 2025-08-19
Crash on divergence in view at cluster layer and replication layer for self

hpatro:


hpatro:
Just trying to understand the scenarios when the cluster and replication layer can diverge. Ignore this PR.

hpatro:
Based on my understanding it shouldn't ever diverge. However, from the test failures we can observe it crashes plenty of times in multiple scenario. Need to deep dive to understand the situation better.

madolson:
@hpatro Can we close this one? Doing some stale PR cleanup and this looks like it was exploratory. Happy to leave it open if you're still investigating.

hpatro:
@madolson will close it out, the divergence problem during steady state was addressed via https://github.com/valkey-io/valkey/pull/2510

PR #2514, opened 2025-08-19
Refactor scanLaterList to fix latency issue

asagege:
This fix only addresses the big list test.

Refactor `scanLaterList` to fix latency issue:

1. Remove the `iterations `stuff and check the time every loop - The iteration optimization should be removed because the time check (19ns) is negligible compared to the actual defragmentation work (even a single 8-byte reallocation takes 43ns and the allocatorShouldDefrag function call takes 49ns per block), so the time check cost at most 30% and likely much less in practice.
2. Remove `bookmark_failed`  - Simplify control flow by handling bookmark failure inline.
3. Move` (*cursor)++` inside time check - We only increment cursor when we actually need to resume.
4. Refactor bookmark failure handling: when bookmark creation failed, break out of the loop, and return with 0 - Skip this one and continue with other quicklists.

Introduced in #2444 and #2205

### DIFF
--- a/src/defrag.c
+++ b/src/defrag.c
@@ -377,43 +377,41 @@ static void defragLater(robj *obj) {
 static long scanLaterList(robj *ob, unsigned long *cursor, monotime endtime) {
     quicklist *ql = ob->ptr;
     quicklistNode *node;
-    long iterations = 0;
-    int bookmark_failed = 0;
     serverAssert(ob->type == OBJ_LIST && ob->encoding == OBJ_ENCODING_QUICKLIST);
 
+    /* Find starting node */
     if (*cursor == 0) {
-        /* if cursor is 0, we start new iteration */
         node = ql->head;
     } else {
         node = quicklistBookmarkFind(ql, "_AD");
         if (!node) {
-            /* if the bookmark was deleted, it means we reached the end. */
             *cursor = 0;
-            return 0;
+            return 0; /* either bookmark failed to create or deleted, skip this one and continue with other quicklists*/
         }
         node = node->next;
     }
 
-    (*cursor)++;
+    /* Process nodes until time expires or list ends */
     while (node) {
         activeDefragQuickListNode(ql, &node);
         server.stat_active_defrag_scanned++;
-        if (++iterations > 128 && !bookmark_failed) {
-            if (getMonotonicUs() > endtime) {
-                if (!quicklistBookmarkCreate(&ql, "_AD", node)) {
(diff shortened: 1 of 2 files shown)


madolson:
Can you also revert the change from https://github.com/valkey-io/valkey/pull/2421/files, then we can more easily run all the tests repeatedly and verify it worked as intended.

JimB123 (approved):
This fixes the existing logic error around bookmark creation (though it should never actually fail).  The code is more readable.

I concur with removing the loop counter.  This simplifies the code, ensures that time constraints are more closely adhered to, and won't substantially impact defrag cycle time.

JimB123:
> Can you also revert the change from https://github.com/valkey-io/valkey/pull/2421/files, then we can more easily run all the tests repeatedly and verify it worked as intended.

Note that Alina has addressed the issue with LIST, but hasn't completed researching other types.  We may see occasional failures.  (But I agree that this is a good thing - we will be able to identify remaining issues.)

madolson:
> Note that Alina has addressed the issue with LIST, but hasn't completed researching other types. We may see occasional failures. (But I agree that this is a good thing - we will be able to identify remaining issues.)

Yeah, we can override it just for lists then. We can run the test in the loop to verify it resolved the issue.

asagege:
> > Note that Alina has addressed the issue with LIST, but hasn't completed researching other types. We may see occasional failures. (But I agree that this is a good thing - we will be able to identify remaining issues.)
> 
> Yeah, we can override it just for lists then. We can run the test in the loop to verify it resolved the issue.

Hi Madelyn, could you be more specific on what tests need to be added? Revert the changes from PR #2421 and override the behavior specifically for lists? Thanks!

madolson:
@asagege You should be able to update the reference in https://github.com/valkey-io/valkey/blob/ab3ee2ca9530ba527959fa36b1c3b0222a2675a6/tests/unit/memefficiency.tcl#L389 to include latency 5. That should allow us to validate that the big list tests aren't taking long periods of time.

asagege:
> @asagege You should be able to update the reference in
> 
> https://github.com/valkey-io/valkey/blob/ab3ee2ca9530ba527959fa36b1c3b0222a2675a6/tests/unit/memefficiency.tcl#L389
> 
> to include latency 5. That should allow us to validate that the big list tests aren't taking long periods of time.

Thanks, updated the PR.

madolson (approved):
The double full test run makes this looks good.

PR #2516, opened 2025-08-20
Remove internal server object pointer overhead in small strings

rainsupreme:
Achieved 20-30% reduction in overhead.

This was accomplished by reusing the 8B robj->ptr memory to store embedded data. For 16B key and 16B value, this reduces the per-item memory overhead by ~20%. Overall performance was not measurably changed. Raising the threshold for embedding strings produces ~30% reduction in per-item overhead for affected value sizes.

Here's what I did:
1. Modify all `robj->ptr` access to use get/set methods. Most of this was done with a script (code listed below), with a few manual touches. Manual and programmatic changes are in separate commits to make review easier.
2. Next I changed `objectGetVal()` to calculate the location instead of using `robj->ptr`, and modified the embedding code to start writing embedded data 8B earlier, overwriting the ptr location. I changed `objectSetVal()` to assert that no value is embedded - all of the code that assigned `o->ptr` would have been incorrect for embedded strings anyway. (Except for one instance in module.c which I made a separate method for.)

Memory efficiency comparison:
image

Performance is not significantly changed (0-2% faster):
image
For each test I collected 65 minutes of data and discarded the first 5 minutes as warmup. I did not use cluster mode, iothreads=1, pipelining=100, and key size was 16B. Note that for 64B values valkey:unstable did not embed and my branch did embed because I raised the size threshold.

Appendix 1 - refactoring script
```python
# parse robj->ptr references and replace them with objectGetVal(robj) and objectSetVal(robj)

from dataclasses import dataclass
from pathlib import Path
import re

@dataclass
class Ref:
    path: Path
    line: int
    char: int
    note: str

def load(base_path: Path, ref_listing_path: Path):
    with open(ref_listing_path, "r") as f:
        data = f.readlines()

    file = ""
    reflist: list[Ref] = []
    for line in data:
        if line.strip() == "":
            continue
        if not line.startswith(' '):
            file = line.strip()
            continue

        # parse lines like "  2443, 39: sdscatsds(user, descr->ptr);\n"
        # Extract line number, char position, and note
        try:
            parts = line.strip().split(":", 1)
            loc, note = parts[0], parts[1].strip()
            line_num, char_num = [int(x.strip()) for x in loc.split(",")]
            ref = Ref(base_path / file, line_num, char_num, note)
            reflist.append(ref)
        except Exception as e:
            print(f"Failed to parse line: {line.strip()} ({e})")

    return reflist

def get_obj_var(line: str):
    # extract the variable name from a line like "sdsdup(c->argv[argpos]->ptr"
    assert line.endswith("->ptr"), f"Line does not end with '->ptr': {line}"
    line = line[:-5].strip()  # exclude "->ptr"

    pairs = {
        ')': '(',
        ']': '[',
        '}': '{',
        }
    pairends = "([{"

    i = len(line) - 1
    unclosed = ""
    if line[i] == ')': # only start new parenthesis when no other unclosed pairs if it's an outer parenthesis wrapping the whole variable
        unclosed += '('
        i -= 1

    for i in range(i, -1, -1):
        char = line[i]
        if unclosed:
            if char in pairends:
                assert char == unclosed[-1]
                unclosed = unclosed[:-1]
        else:
            # no unclosed pairs, check stop conditions
            if char in pairends or char.isspace() or char in ')!':
                i += 1 # this char was invalid so don't include it
                break

        if char in pairs:
            unclosed += pairs[char]

    var = line[i:]
    stripped_var = var.strip()
    while stripped_var.startswith('(') and stripped_var.endswith(')'):
        stripped_var = stripped_var[1:-1].strip()
    return var, stripped_var

assert get_obj_var("sdsdup(c->argv[argpos]->ptr")[0] == "c->argv[argpos]"
assert get_obj_var("sdsdup(user->descr->ptr")[0] == "user->descr"
assert get_obj_var('addReplyErrorFormat(c, "Unknown node %s", (char *)c->argv[2]->ptr')[0] == "c->argv[2]"
assert get_obj_var("        sds key->ptr = argv[result.keys[i].pos]->ptr")[0] == "argv[result.keys[i].pos]"
assert get_obj_var("     sds err = getAclErrorMessage(result, u, cmd, c->argv[idx + 3]->ptr")[0] == "c->argv[idx + 3]"
assert get_obj_var("batch->keys[i] = ((robj *)batch->keys[i])->ptr")[0] == "((robj *)batch->keys[i])"
assert get_obj_var("if (!client->name || !client->name->ptr")[0] == "client->name"
assert get_obj_var("    set->ptr")[0] == "set"
assert get_obj_var("set->ptr")[0] == "set"

def get_assignment(line) -> str:
    assert line.startswith("ptr = "), f"Line does not start with 'ptr = ': {line}"
    first_semicolon = line.find(';')
    if first_semicolon == -1:
        raise ValueError(f"Line does not contain a semicolon")
    assignment = line[5:first_semicolon].strip()  # exclude "ptr = "
    return assignment

def replace_read_ref(ref: Ref, line) -> str:
    var, stripped_var = get_obj_var(line[:ref.char + 2])  # include "->ptr"


    new_line = line[:ref.char - 1 - (len(var) + 2)] + f"objectGetVal({stripped_var})" + line[ref.char - 1 + 3:]
    return new_line

def replace_write_ref(ref: Ref, line_num: int, lines: list[str]) -> str:
    line = lines[ref.line - 1]

    var, stripped_var = get_obj_var(line[:ref.char + 2])  # include "->ptr"
    assignment = get_assignment(line[ref.char-1:]) # include "ptr = "
    new_line = line[:ref.char - 1 - (len(var) + 2)] + f"objectSetVal({stripped_var}, {assignment})" + line[ref.char + 5 + len(assignment):]
    if " = " in assignment:
        print(f"Follow up on {ref.path}:{ref.line}:{ref.char} - assignment contains '=': {assignment}")
        print("  old:", line[:-1])
        print("  new:", new_line[:-1])
    return new_line

def replace_all(refs: list[Ref], dry_run=False):
    to_replace = refs.copy()
    skipped: list[Ref] = []

    i = 0
    file = None
    data: list[str] = []
    while to_replace:
        ref = to_replace.pop() # backwards in case there are multiple refs in the same line
        if file != ref.path:
            if file and not dry_run:
                assert data, f"No data loaded for {file}"
                # write previous file
                with open(file, "w") as f:
                    f.writelines(data)
            file = ref.path
            with open(file, 'r') as f:
                data = f.readlines()

        assert ref.line - 1 < len(data), f"Line {ref.line} out of range for {file}"
        line = data[ref.line - 1]
        assert ref.char + 2 <= len(line), f"Char {ref.char} out of range for {file} at line {ref.line}"

        if line[ref.char - 3:ref.char - 1] != "->":
            print(f"Skipping {ref.path}:{ref.line}:{ref.char} - not pointer deref")
            print(f"  line: {line[:-1]}")
            skipped.append(ref)
            continue
        if line[ref.char - 1:ref.char + 5] == "ptr = ":
            try:
                data[ref.line - 1] = replace_write_ref(ref, ref.line, data)
            except ValueError as e:
                print(f"Skipping {ref.path}:{ref.line}:{ref.char} - {e}")
                print(f"  line: {line[:-1]}")
                skipped.append(ref)
                continue
        else:
            # read ref
            data[ref.line - 1] = replace_read_ref(ref, line)

        i += 1

    if file and not dry_run:
        assert data, f"No data loaded for {file}"
        # write previous file
        with open(file, "w") as f:
            f.writelines(data)

    print(f"Replaced {i} references, skipped {len(skipped)} references")


reflist = load(Path("/home/ubuntu/rainsupreme"), Path("robj-ptr-refs.txt"))
replace_all(reflist, dry_run=False)
```

Script output (I addressed all of these manually in the following manual commit):
```
~/rainsupreme$ python ptr-refactor-script.py 
Skipping /home/ubuntu/rainsupreme/src/t_list.c:171:22 - Line does not contain a semicolon
  line:             subject->ptr = (where == LIST_HEAD) ? lpPrepend(objectGetVal(subject), objectGetVal(value), sdslen(objectGetVal(value)))
Skipping /home/ubuntu/rainsupreme/src/t_list.c:168:22 - Line does not contain a semicolon
  line:             subject->ptr = (where == LIST_HEAD) ? lpPrependInteger(objectGetVal(subject), (long)objectGetVal(value))
Skipping /home/ubuntu/rainsupreme/src/server.h:816:11 - not pointer deref
  line:     void *ptr;
Follow up on /home/ubuntu/rainsupreme/src/module.c:14278:16 - assignment contains '=': mv = newmv
  old:         value->ptr = mv = newmv;
  new:         objectSetVal(value, mv = newmv);
Follow up on /home/ubuntu/rainsupreme/src/defrag.c:646:44 - assignment contains '=': s = news
  old:     if ((news = activeDefragAlloc(s))) ob->ptr = s = news;
  new:     if ((news = activeDefragAlloc(s))) objectSetVal(ob, s = news);
Follow up on /home/ubuntu/rainsupreme/src/defrag.c:460:46 - assignment contains '=': zs = newzs
  old:     if ((newzs = activeDefragAlloc(zs))) ob->ptr = zs = newzs;
  new:     if ((newzs = activeDefragAlloc(zs))) objectSetVal(ob, zs = newzs);
Follow up on /home/ubuntu/rainsupreme/src/defrag.c:446:46 - assignment contains '=': ql = newql
  old:     if ((newql = activeDefragAlloc(ql))) ob->ptr = ql = newql;
  new:     if ((newql = activeDefragAlloc(ql))) objectSetVal(ob, ql = newql);
Replaced 1773 references, skipped 3 references
```

Appendix 2 - micro benchmarking results
```
ubuntu@ip-10-0-145-171:~/robj-microbench$ ./robj-bench
2025-08-18T23:54:11+00:00
Running ./robj-bench
Run on (4 X 2300.18 MHz CPU s)
CPU Caches:
  L1 Data 32 KiB (x4)
  L1 Instruction 32 KiB (x4)
  L2 Unified 256 KiB (x4)
  L3 Unified 46080 KiB (x1)
Load Average: 0.35, 0.18, 0.06
-----------------------------------------------------------------------
Benchmark                             Time             CPU   Iterations
-----------------------------------------------------------------------
BM_objectGetExpire/1               1.88 ns         1.88 ns    372201373
BM_objectGetExpire/2               1.89 ns         1.89 ns    371970617
BM_objectGetExpire/4               1.89 ns         1.89 ns    369780061
BM_objectGetExpire/8               1.90 ns         1.90 ns    366829968
BM_objectGetExpire/16              1.90 ns         1.90 ns    367467478
BM_objectGetExpire/32              1.88 ns         1.88 ns    369002953
BM_objectGetExpire/64              1.91 ns         1.91 ns    367496136
BM_objectGetExpire/128             1.90 ns         1.90 ns    372766529
BM_objectGetExpire/256             1.89 ns         1.88 ns    371629344
BM_objectGetExpire/512             1.90 ns         1.90 ns    369819056
BM_objectGetExpire/1024            1.90 ns         1.90 ns    372576954
BM_objectGetExpire/2048            1.89 ns         1.89 ns    367325743
BM_objectGetExpire/4096            1.88 ns         1.88 ns    371473964
BM_objectGetExpire_BigO            1.89 (1)        1.89 (1)
BM_objectGetExpire_RMS                0 %             0 %
BM_objectGetKey/1                  2.88 ns         2.88 ns    243820017
BM_objectGetKey/2                  2.91 ns         2.91 ns    242765864
BM_objectGetKey/4                  2.91 ns         2.91 ns    241573647
BM_objectGetKey/8                  2.89 ns         2.89 ns    242279472
BM_objectGetKey/16                 2.90 ns         2.90 ns    240966931
BM_objectGetKey/32                 2.92 ns         2.91 ns    241347563
BM_objectGetKey/64                 2.92 ns         2.91 ns    239829973
BM_objectGetKey/128                2.92 ns         2.92 ns    240860846
BM_objectGetKey/256                2.89 ns         2.89 ns    239168435
BM_objectGetKey/512                2.91 ns         2.90 ns    240518452
BM_objectGetKey/1024               2.91 ns         2.91 ns    242862430
BM_objectGetKey/2048               2.91 ns         2.91 ns    240383966
BM_objectGetKey/4096               2.91 ns         2.91 ns    240724659
BM_objectGetKey_BigO               2.90 (1)        2.90 (1)
BM_objectGetKey_RMS                   0 %             0 %
BM_objectGetVal/1                  3.75 ns         3.75 ns    186373286
BM_objectGetVal/2                  3.75 ns         3.75 ns    186349886
BM_objectGetVal/4                  3.76 ns         3.76 ns    186069434
BM_objectGetVal/8                  3.77 ns         3.77 ns    185854886
BM_objectGetVal/16                 3.76 ns         3.76 ns    185938601
BM_objectGetVal/32                 3.77 ns         3.77 ns    185350054
BM_objectGetVal/64                 3.76 ns         3.76 ns    185015799
BM_objectGetVal/128                1.88 ns         1.88 ns    372679151
BM_objectGetVal/256                1.88 ns         1.88 ns    372541933
BM_objectGetVal/512                1.88 ns         1.88 ns    372547356
BM_objectGetVal/1024               1.88 ns         1.88 ns    372967463
BM_objectGetVal/2048               1.88 ns         1.88 ns    371578466
BM_objectGetVal/4096               1.88 ns         1.88 ns    372867744
BM_objectGetVal_BigO               2.89 (1)        2.89 (1)
BM_objectGetVal_RMS                  32 %            32 %
BM_objectGetVal_Using_ptr/1        1.50 ns         1.50 ns    466016074
BM_objectGetVal_Using_ptr/2        1.51 ns         1.51 ns    465033531
BM_objectGetVal_Using_ptr/4        1.51 ns         1.51 ns    465690307
BM_objectGetVal_Using_ptr/8        1.50 ns         1.50 ns    465292160
BM_objectGetVal_Using_ptr/16       1.50 ns         1.50 ns    466284090
BM_objectGetVal_Using_ptr/32       1.50 ns         1.50 ns    465706801
BM_objectGetVal_Using_ptr/64       1.50 ns         1.50 ns    466260051
BM_objectGetVal_Using_ptr/128      1.50 ns         1.50 ns    465387679
BM_objectGetVal_Using_ptr/256      1.50 ns         1.50 ns    464241882
BM_objectGetVal_Using_ptr/512      1.50 ns         1.50 ns    466336830
BM_objectGetVal_Using_ptr/1024     1.51 ns         1.50 ns    465435503
BM_objectGetVal_Using_ptr/2048     1.50 ns         1.50 ns    466323005
BM_objectGetVal_Using_ptr/4096     1.50 ns         1.50 ns    465025662
BM_objectGetVal_Using_ptr_BigO     1.50 (1)        1.50 (1)
BM_objectGetVal_Using_ptr_RMS         0 %             0 %
```

Appendix 3 - rebasing
I cherry picked and merged each commit in sequence. For the programmatic commit I regenerated it instead of cherry picking. (So it contains no manual edits still.)


madolson:
We probably also want a performance test for IO threading as well either that or a deep pipeline. It will tease out a performance regression more clearly.

src/object.c
```diff
@@ -153,7 +163,11 @@ static robj *createEmbeddedStringObjectWithKeyAndExpire(const char *val_ptr,
     char key_sds_type = has_embkey ? sdsReqType(key_sds_len) : 0;
     size_t key_sds_size = has_embkey ? sdsReqSize(key_sds_len, key_sds_type) : 0;
     size_t val_sds_size = sdsReqSize(val_len, SDS_TYPE_8);
-    size_t min_size = sizeof(robj) + val_sds_size;
+    if (val_sds_size < sizeof(void *)) {
+        val_sds_size = sizeof(void *); /* Ensure it's possible to "unembed" value later */
+    }
+
+    ptrdiff_t min_size = sizeof(robj) - sizeof(void *) + val_sds_size; /* reusing 'ptr' memory */
```
zuiderkwast: Shouldn't this be a `size_t`? We're summing sizes, not pointers.

src/object.c
```diff
@@ -197,57 +212,82 @@ static robj *createEmbeddedStringObjectWithKeyAndExpire(const char *val_ptr,
...
-    size_t remaining_size = bufsize - (data - (char *)(void *)o);
-    objectSetVal(o, sdswrite(data, remaining_size, SDS_TYPE_8, val_ptr, val_len));
+    ptrdiff_t remaining_size = bufsize - (data - (char *)(void *)o);
+
+    /* max bufsize for SDS_TYPE_8 is 255 */
+    assert(val_len <= 255);
+    if (remaining_size > 255) {
+        remaining_size = 255;
+    }
+    sdswrite(data, remaining_size, SDS_TYPE_8, val_ptr, val_len);
```
zuiderkwast: It's not perfect to clamp remaining size to 255. Then the sds alloc field is not correct anymore. Some code uses the sds alloc field for accurate memory usage tracking.

We only embed values when key+value+everything fit within 128 bytes, so bufsize <= 128 which implices that remaining_size < 128. So I believe it can never be > 255. Correct? If yes, then we can change this `if` to an assert instead.

Btw, I think `remaining_size` should be a `size_t` ... because it's a size.

src/object.c
```diff
@@ -288,7 +328,41 @@ robj *objectSetExpire(robj *o, long long expire) {
...
+        if (sdslen(new_sds) > sdsalloc(old_sds)) {
+            /* no space for new value - need to realloc or use separate allocation */
+            assert(false); // TODO implement something?
+        } else {
+            /* Reuse the existing sds memory. */
+            sdssetlen(old_sds, sdslen(new_sds));
+            memcpy(old_sds, new_sds, sdslen(new_sds));
+            sdsfree(new_sds); // TODO should we free it?
+        }
+    } else {
```
zuiderkwast: This logic wasn't needed until now. In practice, I think we treated the embedded strings as immutable and just created a new one instead of updating an existing one. Should we continue doing the same? We could forbid this function from being used on embedded strings and add an `assert(!o->hashembval)`.

zuiderkwast (commented):
Great! I'm aligned with this change. We have been chatting about it before.

I posted a few random comments.

src/object.c
rainsupreme: Yes, I agree. All current code that set robj->ptr already ensured that it wasn't using OBJ_ENCODING_EMBSTR, or else it was wrong. (Except for one occurrence in module.c which I wrote an un-embed method for.)

rainsupreme:
I updated the test results using a pipeline depth of 100, and I've also rebased the branch, appended some minor fixes, and addressed PR feedback. :)

src/unit/test_object.c
```diff
@@ -58,30 +58,30 @@ int test_embedded_string_with_key(int argc, char **argv, int flags) {
     sds key = sdsnew("k:123456789012345678901234567890");
     TEST_ASSERT(sdslen(key) == 32);
 
-    /* 15B key should be embedded within 64B. Contents:
+    /* 15B key should be embedded within 128B. Contents:
```
zuiderkwast: ~~15B key~~

32B key + 79B value?

src/unit/test_object.c
```diff
@@ -99,29 +99,30 @@ int test_embedded_string_with_key_and_expire(int argc, char **argv, int flags) {
     sds key = sdsnew("k:123456789012345678901234567890");
     TEST_ASSERT(sdslen(key) == 32);
 
-    /* 7B key should be embedded within 64B. Contents:
+    /* 7B key should be embedded within 128B. Contents:
```
zuiderkwast: ~~7B key~~

32B key + 71B value?

zuiderkwast (commented):
Not a full review; just follow-up on previous comments.

Also take a look at the failing test on 32-bit x86:

```
*** [err]: Memory usage of embedded string value in tests/unit/type/string.tcl
Expected '6' to be equal to '10' (context: type eval line 24 cmd {assert_equal 6 $sds_overhead} proc ::test)
```

src/server.h
```diff
@@ -800,9 +800,11 @@ struct serverObject {
                               * and most significant 16 bits access time). */
     unsigned hasexpire : 1;
     unsigned hasembkey : 1;
+    unsigned hasembval : 1;
     unsigned refcount : OBJ_REFCOUNT_BITS;
-    void *ptr;
+    void *val_ptr;
```
zuiderkwast: To make sure future contributors use objectGetVal instead of accessing this field directly, let's add a comment about it.

Idea:

```suggestion
    void *val_ptr; /* Not always present. Use objectGetVal(obj) and
                    * objectSetVal(obj, val) instead. */
```

src/robj-ptr-refs.txt
zuiderkwast, line 1: File added by mistake?

You can add notes under `misc/` that's in .gitignore so it won't be added by mistake.

src/server.h
```diff
@@ -800,9 +800,11 @@ struct serverObject {
...
 };
+static_assert(sizeof(struct serverObject) <= 16, "unexpected size - verify struct is packed correctly");
```
zuiderkwast: To make this assert correct on 64-bit and 32-bit, how about computing it as `8 + sizeof(void *)`?

```suggestion
static_assert(sizeof(struct serverObject) <= 8 + size(void *), "unexpected size - verify struct is packed correctly");
```

src/object.c
```diff
@@ -231,31 +260,44 @@ robj *createStringObjectFromSds(const sds s) {
...
+            /* Skip embedded key */
+            uint8_t hdr_size = *(uint8_t *)data;
+            data += 1 + hdr_size;                /* +1 for header size byte */
+            data += sdslen((const_sds)data) + 1; /* +1 for null terminator */
+        }
+        if (o->encoding == OBJ_ENCODING_EMBSTR) {
+            return data + sdsHdrSize(SDS_TYPE_8);
+        } else {
+            return data;
+        }
```
zuiderkwast: Embedded value is only used for OBJ_ENCODING_EMBSTR, right? The else clause is dead code. Better use an assert, or do you have a plan to use embedded values of other encodings?

```suggestion
        assert(o->encoding == OBJ_ENCODING_EMBSTR);
        return data + sdsHdrSize(SDS_TYPE_8);
```

tests/unit/type/string.tcl
```diff
@@ -816,22 +816,30 @@ if {[string match {*jemalloc*} [s mem_allocator]]} {
...
-        # +--------------+--------------+---------------+----------------+
-        # | serverObject |              | sds5 key      | sds8 value     |
-        # | header       | key-hdr-size | hdr "quux" \0 | hdr "xyzzy" \0 |
-        # | 16 bytes     | 1            | 1  + 4    + 1 | 3  + 5     + 1 |
-        # +--------------+--------------+---------------+----------------+
+        # +----------+---------------+----------+---------------+----------------+
+        # | robj     | embval reused | key sds  | sds5 key      | sds8 value     |
+        # | header   | void *ptr     | hdr size | hdr "quux" \0 | hdr "xyzzy" \0 |
+        # | 16 bytes | -8            | 1        | 1  + 4    + 1 | 3  + 5     + 1 |
+        # +----------+---------------+----------+---------------+----------------+
```
zuiderkwast: A block with negative size is a bit abstract. I suggest you just draw one box with "robj header excluding val_ptr, 8 bytes" instead of the first two boxes here.

tests/unit/type/string.tcl
```diff
@@ -816,22 +816,30 @@ if {[string match {*jemalloc*} [s mem_allocator]]} {
...
         #
+        # In 32 bit builds the robj header is 12 bytes and the overwritten pointer is 4 bytes
+        # The test must work when compiled for 32 bit but run on a 64 bit system.
```
zuiderkwast: I suppose this is applicable also to when TCL runs on one platform and valkey-server on another. It's possible to run the tests against an external server with certain flags to the runtest script.

```suggestion
        # The test must work when the server is compiled for 32 bit and TCL is
        # compiled and run on a 64 bit system or vice versa. The size of a
        # pointer in the server is what matters.
```

src/debug.c
```diff
@@ -612,38 +612,48 @@ void debugCommand(client *c) {
...
+        if (!strcasecmp(objectGetVal(c->argv[2]), "prevent-pause")) {
+            server.debug_slot_migration_prevent_pause = atoi(objectGetVal(c->argv[3]));
+        } else if (!strcasecmp(objectGetVal(c->argv[2]), "prevent-failover")) {
+            server.debug_slot_migration_prevent_failover = atoi(objectGetVal(c->argv[3]));
+        } else {
+            addReplySubcommandSyntaxError(c);
+            return;
+        }
+        addReply(c, shared.ok);
+    } else if (!strcasecmp(objectGetVal(c->argv[1]), "object") && (c->argc == 3 || c->argc == 4)) {
```
zuiderkwast, line 640: What happened here? "slotmigration" is covered in the `else if` above. Merge problem?

src/replace_obj_refs.py
zuiderkwast, line 1: Don't add this file.

src/t_hash.c
```diff
@@ -1008,7 +1008,10 @@ void hmgetCommand(client *c) {
...
     addReplyArrayLen(c, c->argc - 2);
     for (i = 2; i < c->argc; i++) {
-        addHashFieldToReply(c, o, c->argv[i]->ptr);
+        addHashFieldToReply(c, o, objectGetVal(c->argv[i]));
+    }
+    if (o && hashTypeLength(o) == 0) {
+        dbDelete(c->db, c->argv[1]);
     }
     if (o && hashTypeLength(o) == 0) {
         dbDelete(c->db, c->argv[1]);
```
zuiderkwast: Duplicated code.

src/debug.c
SoftlyRaining: Yes, seemingly. I told VS Code to resolve all conflicts with current version, but maybe it wasn't so smart here 🤦🏻‍♀️

rainsupreme: Yes, seemingly. I told VS Code to resolve all conflicts with current version, but maybe it wasn't so smart here 🤦🏻‍♀️

rainsupreme: I reviewed it all and found a couple other instances of duplicated code. Should be all fixed now, but this change is kind of comically vulnerable to merge conflicts 😅 

edit: I fixed this issue by re-generating the mechanical changes instead of trying to rebase that commit.

src/object.c
```diff
@@ -200,62 +215,86 @@ static robj *createEmbeddedStringObjectWithKeyAndExpire(const char *val_ptr,
...
+static robj *createStringObjectWithKeyAndExpire(const char *ptr, size_t len, const_sds key, long long expire) {
+    if (shouldEmbedStringObject(len, key, expire)) {
         return createEmbeddedStringObjectWithKeyAndExpire(ptr, len, key, expire);
     } else {
-        return createObjectWithKeyAndExpire(OBJ_STRING, sdsnewlen(ptr, len), key, expire);
+        return createUnembeddedObjectWithKeyAndExpire(OBJ_STRING, sdsnewlen(ptr, len), key, expire);
+    }
+}
+
+void *objectGetVal(const robj *o) {
```
zuiderkwast, line 263: We compile with `-O3 -flto` by default, but I know some users compile with `-O2` because it's safer (less risk of weird stuff like strict aliasing issues, etc). I believe we do this internally at Ericsson.

I have a small concern that this function can make things slower if it's not inlined. Should we test this to be sure?

If it's true, then maybe we could explicitly inline a part of it.

<details>
<summary>Micro-optimization ideas</summary>

Static inline some part of it like below, or even `static inline` all of it (and move it to server.h).

```c
static inline objectGetValue(robj *o) {
    return o->hasembval ? objectGetEmbeddedValue(o) : o->val_ptr;
}
```

... or a perhaps hotter code path, which is hit for `c->argv[i]` i.e. parsed incoming commands:

```c
    if (o->hasembval && !o->hasexpire && !hasembkey) {
        unsigned char *data = (char *)o + offsetof(robj, val_ptr);
        return data + sdsHdrSize(SDS_TYPE_8);
    }
```

It's hard to know without proper benchmarking (with -O2).
</details>

zuiderkwast:
"Reduce small string memory overhead by ~20-30%"

This PR title isn't perfect. It describes the benefit, which is important, but a PR title should instead primarily summarize the change itself.

I mean, reducing memory overhead can be done in other ways too, so it's not immediately clear from the title what this change is.

Naming isn't easy. We also want to avoid going into code details, because these titles will end up in release notes and be read by users. How about something like "Remove internal pointer overhead in small strings"?

ranshid:
Small note: we should definitely take that, but it would probably impact our ability to simply backport fixed to older versions. Maybe we could just backport the function objectGetVal as a macro or something which will use the ptr in older versions?

(These are the cases I miss c++ operator overloading :))

src/object.c
JimB123: I generally dislike explicitly marking things as inline.  Largely because I see "inline" misused and misunderstood.  IMO the compiler is generally better at deciding if/when to inline.

In this case - this is an API function used massively across the code space.  Without LTO, most usage can't be inlined anyway.  With LTO, there's no reason for inline to be explicitly specified.

JimB123: A bigger concern is to understand what the function is returning.  It's returning a `void*`.

EDIT: on strings, the void is sds.  This removes my concern about length.

For non-string types, a void return makes sense, but returning a value, without the associated type seems awkward, requiring the type to retrieved through other means before it's possible to interpret the value.  The same concern arises for string INT encoding.

JimB123: For sds strings being returned, the caller needs to be aware that these can't be modified.

For other types, modification may technically be possible, but if there is a refcount on the object, modification would be invalid.

zuiderkwast: > I generally dislike explicitly marking things as inline. Largely because I see "inline" misused and misunderstood. IMO the compiler is generally better at deciding if/when to inline.

@JimB123 Sure, but cross-compilation-unit inlining isn't done automatically except when compiling with `-O3 -flto`.

> In this case - this is an API function used massively across the code space. Without LTO, most usage can't be inlined anyway. With LTO, there's no reason for inline to be explicitly specified.

Not true. I agree that marking a function as `inline` is usually useless, but `static inline` is very different. A `static inline` function is just like a macro. It's defined in a header file and always inlined everywhere, regardless of compiler flags.

> A bigger concern is to understand what the function is returning. It's returning a void*.

You're right, it's not type safe, but a function returning `void *` isn't worse than the current `void *ptr` struct member. Nothing new to this PR.

When it's an embedded string, it can't be modified and this PR doesn't change that. Same for the refcounter. If you want to change this `void *` to a union of some sort, it can be done in a different PR. It's orthogonal to the changes here AFAICT.

rainsupreme:
It seems that the two failures are both flaky tests failing in unstable prior to this change.

1. https://github.com/valkey-io/valkey/issues/2893
```
*** [TIMEOUT]: benchmark: warmup and duration are cumulative in tests/integration/valkey-benchmark.tcl
```

2. https://github.com/valkey-io/valkey/issues/2908
```
*** [err]: TTL, TYPE and EXISTS do not alter the last access time of a key in tests/unit/introspection-2.tcl
Expected [r object idletime foo] >= 2 (context: type eval line 7 cmd {assert {[r object idletime foo] >= 2}} proc ::test)
```

src/object.c
```diff
@@ -136,19 +139,30 @@ robj *createRawStringObject(const char *ptr, size_t len) {
...
     size_t key_sds_len = has_embkey ? sdslen(key) : 0;
     char key_sds_type = has_embkey ? sdsReqType(key_sds_len) : 0;
     size_t key_sds_size = has_embkey ? sdsReqSize(key_sds_len, key_sds_type) : 0;
     size_t val_sds_size = sdsReqSize(val_len, SDS_TYPE_8);
-    size_t min_size = sizeof(robj) + val_sds_size;
+    if (val_sds_size < sizeof(void *)) {
+        val_sds_size = sizeof(void *); /* Ensure it's possible to "unembed" value later */
+    }
+
+    size_t min_size = sizeof(robj) - sizeof(void *) + val_sds_size; /* reusing 'ptr' memory */
```
ranshid:
```suggestion
    size_t min_size = sizeof(robj) - sizeof(void *) + val_sds_size; /* override 'ptr' memory */
```

src/object.c
```diff
@@ -193,62 +208,86 @@ static robj *createEmbeddedStringObjectWithKeyAndExpire(const char *val_ptr,
     /* Copy embedded value (EMBSTR) always as SDS TYPE 8. Account for unused
      * memory in the SDS alloc field. */
     size_t remaining_size = bufsize - (data - (char *)(void *)o);
-    o->ptr = sdswrite(data, remaining_size, SDS_TYPE_8, val_ptr, val_len);
+
+    /* max bufsize for SDS_TYPE_8 is 255 */
+    assert(val_len <= 255);
+    assert(remaining_size <= 255);
```
ranshid: can we use sdsTypeMaxSize instead of a hardcoded 255?

src/object.c
```diff
@@ -193,62 +208,86 @@ static robj *createEmbeddedStringObjectWithKeyAndExpire(const char *val_ptr,
...
- * used.
- *
- * The current limit of 44 is chosen so that the biggest string object
- * we allocate as EMBSTR will still fit into the 64 byte arena of jemalloc. */
-#define OBJ_ENCODING_EMBSTR_SIZE_LIMIT 44
+static bool shouldEmbedStringObject(size_t val_len, const_sds key, long long expire) {
+    /* When to embed? Embed when the sum is up to 64 bytes. There may be better
+     * heuristics, e.g. we can look at the jemalloc sizes (16-byte intervals up
+     * to 128 bytes). */
+    if (val_len > 255) return false; /* SDS_TYPE_8 max size is 255 */
```
ranshid: woudl be possible to use sdsTypeMaxSize?

src/object.c
```diff
@@ -193,62 +208,86 @@ static robj *createEmbeddedStringObjectWithKeyAndExpire(const char *val_ptr,
...
 
-/* Create a string object with EMBSTR encoding if it is smaller than
- * OBJ_ENCODING_EMBSTR_SIZE_LIMIT, otherwise the RAW encoding is
- * used.
- *
- * The current limit of 44 is chosen so that the biggest string object
- * we allocate as EMBSTR will still fit into the 64 byte arena of jemalloc. */
-#define OBJ_ENCODING_EMBSTR_SIZE_LIMIT 44
+static bool shouldEmbedStringObject(size_t val_len, const_sds key, long long expire) {
+    /* When to embed? Embed when the sum is up to 64 bytes. There may be better
```
ranshid: ?
```suggestion
    /* When to embed? Embed when the sum is up to 128 bytes. There may be better
```

src/object.c
```diff
@@ -193,62 +208,86 @@ static robj *createEmbeddedStringObjectWithKeyAndExpire(const char *val_ptr,
...
+     * heuristics, e.g. we can look at the jemalloc sizes (16-byte intervals up
+     * to 128 bytes). */
+    if (val_len > 255) return false; /* SDS_TYPE_8 max size is 255 */
+
+    size_t size = sizeof(robj) - sizeof(void *); /* reusing 'ptr' memory when embedding */
+    if (key) {
+        size_t key_len = sdslen(key);
+        size += sdsReqSize(key_len, sdsReqType(key_len)) + 1; /* 1 byte for prefixed sds hdr size */
+    }
+    size += (expire != -1) * sizeof(long long);
```
ranshid:
```suggestion
    size += (expire != EXPIRY_NONE) * sizeof(long long);
```

src/object.c
```diff
@@ -193,62 +208,86 @@ static robj *createEmbeddedStringObjectWithKeyAndExpire(const char *val_ptr,
...
+    size += sdsReqSize(val_len, SDS_TYPE_8);
+    return size <= 128;
+}
+
+/* Create a string object with EMBSTR encoding if it is small, otherwise RAW encoding */
 robj *createStringObject(const char *ptr, size_t len) {
-    if (len <= OBJ_ENCODING_EMBSTR_SIZE_LIMIT)
+    if (shouldEmbedStringObject(len, NULL, -1))
```
ranshid:
```suggestion
    if (shouldEmbedStringObject(len, NULL, EXPIRY_NONE))
```

src/object.c
```diff
@@ -166,6 +180,7 @@ static robj *createEmbeddedStringObjectWithKeyAndExpire(const char *val_ptr,
     o->lru = 0;
     o->hasexpire = (expire != -1);
```
ranshid:
```suggestion
    o->hasexpire = (expire != EXPIRY_NONE);
```

src/object.c
```diff
@@ -269,53 +308,78 @@ long long objectGetExpire(const robj *val) {
...
+ * returned object instead of 'o' after calling this function. */
+robj *objectSetExpire(robj *o, long long expire) {
+    if (o->hasexpire) {
         /* Update existing expire field. */
-        unsigned char *data = (void *)(val + 1);
+        unsigned char *data = objectEmbeddedData(o);
         *(long long *)data = expire;
-        return val;
+        return o;
     } else if (expire == -1) {
```
ranshid: If we are changing stuff...
```suggestion
    } else if (expire == EXPIRY_NONE) {
```

src/object.c
```diff
@@ -269,53 +308,78 @@ long long objectGetExpire(const robj *val) {
...
+/* Caller is responsible for ensuring that robj does not have an embedded value */
+void objectSetVal(robj *o, void *val) {
+    assert(!o->hasembval);
+    o->val_ptr = val;
+}
+
+/* Sometimes it's necessary to grow an object's value without reallocating it.
+ * The old embedded value memory becomes wasted for the remaining lifetime of this
+ * object. Consider using dbUnshareStringValue() or similar if at all possible */
+void objectUnembedVal(robj *o, void *new_val) {
```
ranshid: Suggestion: can we assert in case the provided new_val is equal to the objectGetVal? might help identify corruptions

src/object.c
```diff
@@ -269,53 +308,78 @@ long long objectGetExpire(const robj *val) {
...
-        return val;
+        return o;
     } else {
-        return objectSetKeyAndExpire(val, objectGetKey(val), expire);
+        return objectSetKeyAndExpire(o, objectGetKey(o), expire);
     }
 }
 
+/* Caller is responsible for ensuring that robj does not have an embedded value */
+void objectSetVal(robj *o, void *val) {
```
ranshid, line 325: nit: The get and set are pretty asymmetric in the sense that they cannot be used on the same objects. for example the get can be used on an embedded value object but set cannot. I suggest to name this into 
 objectSetValPtr maybe?

src/object.c
```diff
@@ -328,7 +392,7 @@ robj *tryCreateRawStringObject(const char *ptr, size_t len) {
 
 /* Same as createStringObject, can return NULL if allocation fails */
 robj *tryCreateStringObject(const char *ptr, size_t len) {
-    if (len <= OBJ_ENCODING_EMBSTR_SIZE_LIMIT)
+    if (shouldEmbedStringObject(len, NULL, -1))
```
ranshid:
```suggestion
    if (shouldEmbedStringObject(len, NULL, EXPIRY_NONE))
```

src/object.c
```diff
@@ -786,19 +850,19 @@ void trimStringObjectIfNeeded(robj *o, int trim_small_values) {
      * 1. When an arg len is greater than PROTO_MBULK_BIG_ARG the query buffer may be used directly as the SDS string.
      * 2. When utilizing the argument caching mechanism in Lua.
      * 3. When calling from RM_TrimStringAllocation (trim_small_values is true). */
-    size_t len = sdslen(o->ptr);
+    size_t len = sdslen(o->val_ptr);
     if (len >= PROTO_MBULK_BIG_ARG || trim_small_values ||
         (server.executing_client && server.executing_client->flag.script && len < LUA_CMD_OBJCACHE_MAX_LEN)) {
-        if (sdsavail(o->ptr) > len / 10) {
-            o->ptr = sdsRemoveFreeSpace(o->ptr, 0);
+        if (sdsavail(o->val_ptr) > len / 10) {
```
ranshid, line 858: do we even need to get here? we could just return in case the value is embedded without need to access the sds right?

src/object.c
```diff
@@ -1133,13 +1195,13 @@ size_t objectComputeSize(robj *key, robj *o, size_t sample_size, int dbid) {
 
     if (o->type == OBJ_STRING) {
         if (o->encoding == OBJ_ENCODING_RAW) {
-            asize += sdsAllocSize(o->ptr);
+            asize += sdsAllocSize(objectGetVal(o));
```
ranshid, line 1200: So when the value is embedded, do we count the value twice (zmalloc_size((void *)o) + dsAllocSize(objectGetVal(o)))?

ranshid (commented):
I gave some comments. mainly around object.c

src/object.c
zuiderkwast: sdsTypeMaxSize is `static inline` and defined in sds.c, but we can move it to sds.h and then we can use it.

Alternatively, we can define these macros in sds.h:

```C
#define SDS_TYPE_5_MAX_SIZE ((1 << 5) - 1)
#define SDS_TYPE_8_MAX_SIZE ((1 << 8) - 1)
#define SDS_TYPE_16_MAX_SIZE ((1 << 16) - 1)
#define SDS_TYPE_32_MAX_SIZE ((1ll << 32) - 1)
```

src/object.c
zuiderkwast: Let's revise the next two lines of this comment as well. With this change, we actually do what the comment suggests that we could do.

src/object.c
zuiderkwast: FWIW, this concern still stands. Can we do a benchmark run without link-time-optimization?

src/object.c
zuiderkwast: I'm leaning towards not agreeing with this comment. :smile:

An object's value is a pointer. `objectGetVal` also returns a pointer. When we use Ptr in a name like this, it sometimes mean another level of pointers as in `void **`. (`val_ptr` is internal and should not accessed directly.)

src/object.c
zuiderkwast: No, when it's embedded, we don't get to this case (OBJ_ENCODING_EMBSTR) and the object allocation itself counted above (`zmalloc_size((void *)o)`) covers it all.

src/server.c
```diff
@@ -407,12 +407,12 @@ void *dictSdsDup(const void *key) {
...
 int dictObjKeyCompare(const void *key1, const void *key2) {
     const robj *o1 = key1, *o2 = key2;
-    return dictSdsKeyCompare(o1->ptr, o2->ptr);
+    return dictSdsKeyCompare(objectGetVal(o1), objectGetVal(o2));
 }
 
 uint64_t dictObjHash(const void *key) {
     const robj *o = key;
-    return dictGenHashFunction(o->ptr, sdslen((sds)o->ptr));
+    return dictGenHashFunction(objectGetVal(o), sdslen((sds)objectGetVal(o)));
```
zuiderkwast, line 415: If link-time-optimization is disabled, code like this is a concern. If it's not inlined, we should probably call objectGetVal once instead of twice in hot code paths like this.

src/server.h
```diff
@@ -795,9 +795,12 @@ struct serverObject {
     unsigned lru : LRULFU_BITS;
     unsigned hasexpire : 1;
     unsigned hasembkey : 1;
+    unsigned hasembval : 1;
```
zuiderkwast, line 852: What's the reasoning for adding this bit instead of relying on `o->encoding == OBJ_ENCODING_EMBSTR` only?

src/module.c
```diff
@@ -3052,11 +3052,10 @@ ValkeyModuleString *moduleAssertUnsharedString(ValkeyModuleString *str) {
     if (str->encoding == OBJ_ENCODING_EMBSTR) {
         /* Note: here we "leak" the additional allocation that was
          * used in order to store the embedded string in the object. */
-        str->ptr = sdsnewlen(str->ptr, sdslen(str->ptr));
-        str->encoding = OBJ_ENCODING_RAW;
+        objectUnembedVal(str, sdsnewlen(objectGetVal(str), sdslen(objectGetVal(str))));
```
zuiderkwast: This is the only place `objectUnembedVal` is called AFAICT.

It'd be cleaner to just pass one argument and let `objectUnembedVal` do the rest.

```suggestion
        objectUnembedVal(str);
```

See also the comment where this function is defined: https://github.com/valkey-io/valkey/pull/2516/files#r2622257872

zuiderkwast (commented):
Posted a few minor comments. Also agreeing with Ran's comments. Please address them so we can merge before we get merge conflicts.

My main concern is still this comment though: https://github.com/valkey-io/valkey/pull/2516/files#r2299611959. It was marked as resolved, by mistake?

Since we don't run any benchmarks or tests without LTO or with other optimization flags like `-O2` (but which I know we use in Ericsson in production and maybe others do too) I find it safest to force inlining by moving `objectGetVal` to `server.h` and marking it as `static inline`, which has the same effect as if it were a macro. In the hot code paths like in `GET foo`, we access that `argv[1]` argument many times using objectGetVal.

If we move it to server.h, to make sure `objectGetVal` doesn't depend on any internal functions in object.c, we can replace the call

```diff
-    const unsigned char *data = objectEmbeddedData((robj *)o);
+    const unsigned char *data = (char *)o + offsetof(robj, val_ptr);
```

... because here we know that `o->hasembval` is set.

src/object.c
ranshid: I was mainly referring to why do we not handle the embedded value case? because it might require relocating the object in order to save memory (and will complicate many cases were the refcount is > 1) right?

IMO it makes it clearer explicitly stating that in the function name (maybe personal taste?) and we could also do `objectSetValueRef`.

owever TBH this is not a blocker

src/object.c
rainsupreme: "overwriting" maybe?

src/object.c
```diff
@@ -136,19 +139,30 @@ robj *createRawStringObject(const char *ptr, size_t len) {
     return createObject(OBJ_STRING, sdsnewlen(ptr, len));
 }
 
-/* Creates a new embedded string object and copies the content of key, val and
- * expire to the new object. LRU is set to 0. */
+/* get beginning of embedded data. embedded data flags must be accurate when called. */
+static unsigned char *objectEmbeddedData(const robj *o) {
```
JimB123, line 143: Can we change the name to `objectEmbeddedValue`?
For forkless, we will be proposing an embedded metadata for DB entries.  I think the name will get confused if this is just "data".

src/object.c
JimB123: >In this case - this is an API function used massively across the code space. Without LTO, most usage can't be inlined anyway. With LTO, there's no reason for inline to be explicitly specified.
>>Not true. I agree that marking a function as inline is usually useless, but static inline is very different. A static inline function is just like a macro. It's defined in a header file and always inlined everywhere [...]

I absolutely agree.  `static inline` is legitimate in header files.  It's far superior to MACRO definition due to the ability to type check.  (Unless the macro is playing with text substitution.)

But THIS case, it's in the .c file.  I agree that if we move it to the header file, `static inline` is an option.  My only disagreement was with explicitly marking it as `inline` in the `.c` file.

Also, FYI, I'm concerned that we might start to overuse inlines in the headers.  We've measured that LTO can slow down our code if we aren't careful.  It can increase code size, pushing things out of L1 cache.  When we explicitly create `static inline` functions in the `.h` files, it gives the compiler/linker no choice - the function MUST be inlined (or at least replicated in every compilation unit).  This can impact our ability to smartly tune via the LTO options.  For this reason, we need to ensure that only the most trivial of functions are implemented this way - and our default should still be to put the function in the `.c` file and rely on LTO to make the right decisions.

src/object.c
JimB123: I think the whole object "API" is a little questionable, and is out of scope for this update.  We need to spend some time really understanding the use cases and API requirements for `robj` (as a ref counted object) vs `dbEntry` (when we are using this in the DB as a key/value).

I've spent a little time thinking about this, and one of the ideas I had was that maybe `robj` (as a concept - and not as a DB entry) - might match the pattern of an "immutable".  But this needs some more thought.

As is though, I'd lean away from your suggested name change - as I believe that in the fullness of time, the caller shouldn't need to worry about the internal representation.

src/object.c
rainsupreme: Hmm. This isn't just value though, this refers to all the embedded data - with this PR that means expire, key, and/or value. I can't think of a better name that addresses your point while still being accurate, so for now I'll improve the comment. Lmk if you think of a good accurate name for this though ofc

src/server.h
rainsupreme: Mainly I thought this would be more uniform, easier to understand, and better organized. It might make it easier to embed other encodings in the future (small listpacks, or new datatypes?), but I haven't investigated that possibility. Do you object?

src/object.c
rainsupreme: yes, the first line of this function means we only reach this for raw (sds) encoding: `if (o->encoding != OBJ_ENCODING_RAW) return;` (i.e. not OBJ_ENCODING_EMBSTR)

src/object.c
rainsupreme: I'll try the first suggestion :)

src/object.c
rainsupreme: Would it suffice to create the sds copy within this function and trust that?

src/object.c
rainsupreme: I agree with Jim. If we investigate we'll probably find that our current robj use cases would be best served by separate structs, and I expect to find opportunities to save more memory. For now let's just bookmark this issue. :)

src/module.c
rainsupreme: You're right, this is the only place it's called! I was trying to be too type-agnostic in object.c perhaps - I'll implement your suggestion.

src/server.h
zuiderkwast: I don't object. :smile: I think these are good points. It's good to have them spelled out.

src/object.c
zuiderkwast: Jim, if you want to add more embedded data, that should try to build on the same mechanism for embedding data, if possible.

We can also refactor and rename things later. Let's not prepare too much for future work because YAGNI.

src/object.c
zuiderkwast: I have spent much time with the object API when I did the work to embed key in the object. Ideally we should split it into two types: a dbEntry and a string object, but robj is everywhere in the code base and the two have many overlapping uses so I think breaking them apart is not feasable.

It is not entirely immutable. Mutation is done by APPEND, INCR, bit operation commands,  ValkeyModule_StringDMA and others. When a object's value is a listpack, hashtable or similar, those values are not immutable either.

src/object.c
zuiderkwast: > Would it suffice to create the sds copy within this function and trust that?

You mean `void objectUnembedVal(robj *o)` and let it do the string copying, right? Yes, if we do that, it guarantees that the string value is the same before and after.

src/object.c
```diff
@@ -331,14 +330,17 @@ void objectSetVal(robj *o, void *val) {
 /* Sometimes it's necessary to grow an object's value without reallocating it.
  * The old embedded value memory becomes wasted for the remaining lifetime of this
  * object. Consider using dbUnshareStringValue() or similar if at all possible */
-void objectUnembedVal(robj *o, void *new_val) {
+void objectUnembedVal(robj *o) {
     assert(o->hasembval);
+    assert(o->encoding == OBJ_ENCODING_EMBSTR);
+
+    const_sds embedded_sds = objectGetVal(o);
+    assert(sdsAllocSize(embedded_sds) >= sizeof(void *));
```
zuiderkwast, line 338: I don't think this assert is meaingful. For very small embstr, we have an embedded SDS type 5 here. For these, sdsAllocSize uses malloc_usable_size which returns the size of the whole robj allocation.

rainsupreme:
Seems like the additional code coverage failure was probably from bad memory...
```
------ FAST MEMORY TEST ------
59636:M 16 Dec 2025 21:04:43.813 # Bio worker thread #0 terminated
59636:M 16 Dec 2025 21:04:43.813 # Bio worker thread #1 terminated
59636:M 16 Dec 2025 21:04:43.814 # Bio worker thread #2 terminated
59636:M 16 Dec 2025 21:04:43.814 # Bio worker thread #3 terminated
*** Preparing to test memory region 560549015000 (2859008 bytes)
*** Preparing to test memory region 5605587c9000 (135168 bytes)
*** Preparing to test memory region 7fb1fc000000 (135168 bytes)
*** Preparing to test memory region 7fb202d7c000 (16777216 bytes)
*** Preparing to test memory region 7fb203d7d000 (16777216 bytes)
*** Preparing to test memory region 7fb204d7e000 (16777216 bytes)
*** Preparing to test memory region 7fb207d80000 (2621440 bytes)
*** Preparing to test memory region 7fb208400000 (8388608 bytes)
*** Preparing to test memory region 7fb208e1c000 (53248 bytes)
*** Preparing to test memory region 7fb208ec1000 (12288 bytes)
*** Preparing to test memory region 7fb208fb7000 (8192 bytes)
.E.E.E.E.E.E.E.E.E.E.E
!!! MEMORY ERROR DETECTED! Check your memory ASAP !!!
```

src/object.c
rainsupreme: Actually, the EMBSTR encoding always uses SDS_TYPE_8. This also means that the size of the sds header is constant so we don't need to spend a byte storing it like we do with embedded keys. (Since embedded keys can use any sds header type and be any size)

zuiderkwast: Ah, you're right!

src/object.c
zuiderkwast: > I was mainly referring to why do we not handle the embedded value case? because it might require relocating the object in order to save memory (and will complicate many cases were the refcount is > 1) right?

Yeah, it would need to return a new value. In a reference counted system, you could use copy-on-write style so when you modify a shared object, you get two copies. I believe it'd be a large change. Currently, our code explicitly creates new objects when needed.
 
> IMO it makes it clearer explicitly stating that in the function name (maybe personal taste?) and we could also do `objectSetValueRef`.

I agree about using clear and exact function names, but I don't like the suggested name. :) The symmetry with the getter for name and type is more more important IMO. A name with Ref or Ptr would hint that it sets a pointer to the value rather than setting the value itself, which I think would be counter-intuitive.

We have a comment and an assert to prevent the forbidden case, so it's not a risk in practice.

src/object.c
rainsupreme: Actually, the EMBSTR encoding always uses SDS_TYPE_8. This also means that the size of the sds header is constant so we don't need to spend a byte storing it like we do with embedded keys. (Since embedded keys can use any sds header type and be any size)

This was meant as a very cautious assert in case our embedding code ever got changed and failed to ensure there is 8B available (assuming 64 bit) for a pointer after removing the value data. For an sds string here that would mean 3 characters or less. (3B SDS_TYPE_8 header, <= 3B data, 1B null terminator). Given that this is only called in one place for modules, I don't think this is a common code path.

Alternatively, I could redo the assert to query for the overall robj allocation size and do it that way. Do you think that would be better?

zuiderkwast: We rely on the `sds_hdr8->alloc` field is populated correctly, but I believe it is.

> Alternatively, I could redo the assert to query for the overall robj allocation size and do it that way. Do you think that would be better?

No, it's fine like this. Maybe just add a code comment if you want, but not necessary.

src/object.c
rainsupreme: Jim and I talked in person and didn't think of a better name. I did improve the function comment though. :)

zuiderkwast (approved):
If this turns out to be a performance regression when compiling without LTO, then we can tackle that later.

Let's get this merged!

src/server.h
ranshid: I actually liked @zuiderkwast comment. We better save these bits for whatever future reason and only use them in case we have to.

ranshid (approved):
giving a tentative approve, as I think the only comment left on my side is the fact that we better avoid using the extra bit in object

src/object.c
```diff
@@ -140,7 +140,7 @@ robj *createRawStringObject(const char *ptr, size_t len) {
 }
 
 /* Get beginning of embedded data, which may contain expire, key, and/or value. Embedded data flags must be accurate when called. */
-static unsigned char *objectEmbeddedData(const robj *o) {
+__attribute__((pure)) static unsigned char *objectEmbeddedData(const robj *o) {
```
JimB123: The pure/const changes look correct.  Is there a reason that the implementations include the attribute on the left, and the forward declarations include the attribute on the right?

rainsupreme: - For the implementation it can be skipped but is important for documentation, and syntax parsing requires it to be on the left.
- For headers it can be either side but right is conventional in public libraries and such from what I've been able to find. 🤷‍♀️  Our codebase puts it on both sides, but within server.h it is consistently on the right.

JimB123 (approved):

src/server.h
rainsupreme: This change only impacts code within object.c - no external code should access these flags. That means that it would be quite easy to replace this bit flag with an encoding check method later. Let's not worry about it until we actually need the bit.

rainsupreme:
For anyone wondering what I'm waiting for: In my latest tests I'm seeing a 4-5% regression with and without LTO for SET with 512B values on ARM architecture. So I'm investigating further to see if this came from adding the pure attribute and how smaller data sizes were affected.

tests/unit/type/string.tcl
```diff
@@ -816,22 +816,31 @@ if {[string match {*jemalloc*} [s mem_allocator]]} {
...
         # Check that the SDS overhead of the embedded key and value is 6 bytes
         # (sds5 + sds8). This is the memory layout:
         #
-        # +--------------+--------------+---------------+----------------+
-        # | serverObject |              | sds5 key      | sds8 value     |
-        # | header       | key-hdr-size | hdr "quux" \0 | hdr "xyzzy" \0 |
-        # | 16 bytes     | 1            | 1  + 4    + 1 | 3  + 5     + 1 |
-        # +--------------+--------------+---------------+----------------+
+        # +-------------------+----------+---------------+----------------+
+        # | robj header       | key sds  | sds5 key      | sds8 value     |
```
madolson, line 828:
```suggestion
        # | serverObject       | key sds  | sds5 key      | sds8 value     |
```
Is there a reason we reverted the name back to robj?

tests/unit/type/string.tcl
```diff
@@ -816,22 +816,31 @@ if {[string match {*jemalloc*} [s mem_allocator]]} {
...
+        # | excluding val_ptr | hdr size | hdr "quux" \0 | hdr "xyzzy" \0 |
+        # | 8 bytes           | 1        | 1  + 4    + 1 | 3  + 5     + 1 |
+        # +-------------------+----------+---------------+----------------+
         #
+        # The test must work when the server is compiled for 32 bit and TCL is
+        # compiled and run on a 64 bit system or vice versa. The size of a
+        # pointer in the server is what matters.
+        set info_server [r info server]
+        regexp {arch_bits:(\d+)} $info_server _ arch_bits
+        set pointer_size [expr {$arch_bits / 8}]
```
madolson, line 838:
```suggestion
        set pointer_size [expr {[s arch_bits] / 8}]
```
Not that important.

src/server.h
madolson: It's a good reminder that we should probably make valkey object opaque. Do we have an issue for that?

madolson: Honestly, I also kind of agree with Ran and Viktors original comment, but moreso because now we have two ways to refer for the same thing, one is embedded value and the other is embedded string. I also associate integers has having an embedded value.

madolson (approved):
Curious about the performance delta, but besides that I think all the code is fine.

tests/unit/type/string.tcl
zuiderkwast: Reverted? In the past it was typedef struct redisObject robj. We de-branded redisObject to serverObject. Robj has bever been changed.

Robj stands for reference-counted object. 😆

madolson: I'm talking specifically about the comment. It went from serverObject -> robj. I don't know, it's not a very important comment :)

zuiderkwast: Ah... Idk

rainsupreme:
So here's the result of the tests I ran for various commits that are part of this change. Many mysteries here that I'm not sure what to think of. Strange that O2 is faster for get and slower for set, and strange that -O2 SET was unaffected by switching the 64B size from raw encoding to embstr encoding. I guess adding the pure attributes didn't help though, so I should revert that bit.

Error bars indicate 1 standard deviation:
image

This version is percent change relative to unstable as a baseline (error bars = 1 std dev):
image

rainsupreme:
I've rebased again. This time I've omitted the `pure` markers because it did not improve performance.

My conclusions from the performance tests:
- The added complexity in determining the location of the embedded value data caused a ~4% performance decrease for SET performance, and no clear change to GET performance. I'd argue losing ~4% SET perf to eliminate ~25% of memory overhead cost is worth it. (Note: after this change it would be pretty easy to maintain a fork with different behavior or change it here. It's easy to undo this tradeoff. 😉)
- There are different performance patterns for embedded vs non-embedded values. The changes to the 64B value size reflect the change from non-embedded to embedded caused by raising the size threshold.
- For some reason when compiled with -O2, SET commands perform up to ~70% faster. I am surprised, curious, excited, and I plan to investigate this further in the future. Feel free to ping me if you want to chat about that finding 😀

src/server.c
madolson: Rain said that she validated this, so resolving.

rainsupreme:
After discussion with @madolson I've reverted the increase in size threshold for embedding values for now, since for the sizes affected we see 25% worse SET performance, which might break some customer use cases. (Though we do see 20% better GET.) I'll investigate the difference in SET between O2 and O3/LTO compiler options and perhaps we can revisit the threshold change in the future. :)

madolson:
Merging this so that it stops accumulating merge conflicts. There was a comment about doing this in 2 commits, but I think 1 is OK.

enjoy-binbin (commented):
Another memory optimization!
Just leave the comment, good job!

zuiderkwast:
Great, finally!

> After discussion with @madolson I've reverted the increase in size threshold for embedding values for now

So the memory usage is the green line in the "Memory efficiency comparison" diagram. Makes sense.

For a 16B key, it means we only embed values up to 33 bytes, instead of up to 97 bytes.

    +--------+-----------------+------------+------------+
    | header | key header size | key sds5   | val sds8   |
    | 8      | 1               | 1 + 16 + 1 | 3 + 33 + 1 |  = 64 bytes
    +--------+-----------------+------------+------------+

    +--------+-----------------+------------+------------+
    | header | key header size | key sds5   | val sds8   |
    | 8      | 1               | 1 + 16 + 1 | 3 + 97 + 1 |  = 128 bytes
    +--------+-----------------+------------+------------+

Many string values fall in the range 34-97 bytes, so it a little sad that we can't apply this memory saving to these values. Hopefully we can solve the performence issue in the future and increase the threshold later.

PR #2518, opened 2025-08-20
Implement new Module API to test client readonly flag

allenss-amazon:
Search module wants to dispatch cross-shard query operations in accordance with requesting connection readonly flag. This module API exposes the state of the readonly flag for the current client.


allenss-amazon:
Abandoned

PR #2522, opened 2025-08-20
Module API: Add READONLY flag to ClientInfo.flags output structure

### LINKED ISSUE #2487: [NEW] Create Module Interface to detect READONLY flag for current client.

The search module dispatches cross-shard query operations. Proper handling of the READONLY flag for the requesting connection would be to only query the primaries unless the READONLY flag was set. However, currently there is no interface to provide this information.

Have a module interface to report whether the current client connection has the READONLY flag set.

Current code distributes query operations randomly, without regard to READONLY/READWRITE.

**Additional information**

allenss-amazon:
* Add a flag `VALKEYMODULE_CLIENTINFO_FLAG_READONLY` to `ValkeyModuleClientInfo.flags` and set it in `ValkeyModule_GetClientInfoById()`.
* Add an optimization for accessing the current client by ID, to avoid looking it up in a radix tree.

Closes #2487

### DIFF
--- a/src/module.c
+++ b/src/module.c
@@ -3789,7 +3789,7 @@ int modulePopulateClientInfoStructure(void *ci, client *client, int structver) {
     if (client->flag.tracking) ci1->flags |= VALKEYMODULE_CLIENTINFO_FLAG_TRACKING;
     if (client->flag.blocked) ci1->flags |= VALKEYMODULE_CLIENTINFO_FLAG_BLOCKED;
     if (client->conn->type == connectionTypeTls()) ci1->flags |= VALKEYMODULE_CLIENTINFO_FLAG_SSL;
-    if (client->flag.readonly) ci1->flags |= VALKEYMODULE_CLIENTINFO_FLAG_READONLY;    
+    if (client->flag.readonly) ci1->flags |= VALKEYMODULE_CLIENTINFO_FLAG_READONLY;
 
     int port;
     connAddrPeerName(client->conn, ci1->addr, sizeof(ci1->addr), &port);
@@ -3863,9 +3863,9 @@ int modulePopulateReplicationInfoStructure(void *ri, int structver) {
  *      }
  */
 int VM_GetClientInfoById(void *ci, uint64_t id) {
-    client *client = 
+    client *client =
         (server.executing_client && server.executing_client->id == id)
-            ? server.executing_client 
+            ? server.executing_client
             : lookupClientByID(id);
     if (client == NULL) return VALKEYMODULE_ERR;
     if (ci == NULL) return VALKEYMODULE_OK;
--- a/src/module.c
+++ b/src/module.c
@@ -3849,6 +3849,7 @@ int modulePopulateReplicationInfoStructure(void *ri, int structver) {
  *     VALKEYMODULE_CLIENTINFO_FLAG_TRACKING     Client with keys tracking on.
  *     VALKEYMODULE_CLIENTINFO_FLAG_UNIXSOCKET   Client using unix domain socket.
  *     VALKEYMODULE_CLIENTINFO_FLAG_MULTI        Client in MULTI state.
+ *     VALKEYMODULE_CLIENTINFO_FLAG_READONLY     Client in ReadOnly state.
  *
  * However passing NULL is a way to just check if the client exists in case
  * we are not interested in any additional information.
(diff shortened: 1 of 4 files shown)


src/module.c
```diff
@@ -3871,6 +3872,17 @@ int VM_GetClientInfoById(void *ci, uint64_t id) {
...
 
+/*
+Same as ValkeyModule_GetClientInfoById, except it's always for the current client.
+*/
+int VM_GetClientInfo(ValkeyModuleCtx *ctx, void *ci) {
+    if (ci == NULL) return VALKEYMODULE_OK;
+
+    /* Fill the info structure if passed. */
+    uint64_t structver = ((uint64_t *)ci)[0];
+    return modulePopulateClientInfoStructure(ci, ctx->client, structver);
```
zuiderkwast: `VM_GetClientId` also handles the case `ctx->client == NULL`. We should do it here too.

utils/module-api-since.rb
```diff
@@ -80,6 +80,7 @@
 "ReplicateVerbatim" => "4.0.0",
 "GetClientId" => "4.0.0",
 "GetClientUserNameById" => "6.2.1",
+"GetClientInfo" => "9.0.0",
```
zuiderkwast: This is not needed. It's taken automatically from the git tag, except for old redis versions for which we didn't keep the git tag. See the comment at the top of this file:

```
# This file maps module API functions to the Redis OSS versions that added them.
# This is needed only for old API functions from the pre-forking time.
```

(Off topic: Maybe all this information about versions < 7.2 is useless? IIRC we added this info just after forking, when we deleted the redis version tags, but later we changed the prefix from RedisModule to ValkeyModule. ValkeyModule implies it's Valkey, which started at 7.2. So I guess we can delete this info in another PR.)

zuiderkwast (commented):
Adding this looks fine, but it opens up the door to also adding current client variants of all the functions that currently work with a client id:

* VM_GetClientUserNameById
* VM_GetClientNameById
* VM_SetClientNameById
* VM_GetClusterNodeInfoForClient
* VM_DeauthenticateAndCloseClient
* VM_GetClientCertificate

The API might explode if we keep adding all combinations, so maybe instead of opening up this door, we can just keep the convention of referring to clients by ID and this optimization?

```diff
 int VM_GetClientInfoById(void *ci, uint64_t id) {
-    client *client = lookupClientByID(id);
+    client *client = server.current_client != NULL &&
+                     server.current_client->id == id ? server.current_client->id
+                                                     : lookupClientByID(id);
```

utils/module-api-since.rb
madolson: Yeah, seems like we can delete the older tags.

madolson:
> The API might explode if we keep adding all combinations, so maybe instead of opening up this door, we can just keep the convention of referring to clients by ID and this optimization?

I like this!

allenss-amazon:
Removed the extra API, now has only the READONLY Flag plus the optimization in the GetClientInfoByID code.

src/module.c
```diff
@@ -3862,7 +3863,10 @@ int modulePopulateReplicationInfoStructure(void *ri, int structver) {
  *      }
  */
 int VM_GetClientInfoById(void *ci, uint64_t id) {
```
zuiderkwast, line 3866: The new flag needs to be added in the documentation. The module API docs are generated from the comments in this file, so just add this flag in the comment above this function.

zuiderkwast (commented):
LGTM. Just need to add one line of documentation.

A somewhat different topic: I was wondering about `current_client` vs `executing_client`. Executing client is what you have in `ctx->client` in the APIs that get a ctx, including VM_GetClientId, so it seems to be right. It can be a fake-client though, if a module command is called from Lua or from another module, so I wonder if there is a way to get the "real" client in the module API, but that's a somewhat different topic. It's not related to this PR. I'm just curios.

```c
    client *current_client;                /* The client that triggered the command execution (External or AOF). */
    client *executing_client;              /* The client executing the current command (possibly script or module). */
```

src/module.c
allenss-amazon: Ok, pushed a fix to the documentation.

The musing on current_client vs executing_client is interesting. The subtlety between the two of them is not apparent in the documentation. I had to guess which was right one, and I can only claim to be lucky in my choice. It would be nice if the doc was improved to clarify the difference between the two.

zuiderkwast: Yeah, maybe not only the docs need updates here. It seems that it would always be more useful to get the `current_client` in the module API. Not sure if that would be a breaking change though.

Anyway, if current_client is readonly, the executing_client should be marked as readonly too, so for this PR it probably doesn't make any difference. For things like client name, it would probably matter though.

zuiderkwast (approved):

madolson (approved):

zuiderkwast:
@valkey-io/core-team This is a very small addition to the module API. I'm going to merge it without a vote.

PR #2526, opened 2025-08-20
Fix assumptions that pthread functions set errno

### LINKED ISSUE #2525: [BUG] `pthread` functions are expected to set `errno` in some places
When logging the reason for the `pthread_*` failures in these places, `strerror(errno)` will print whatever `errno` was set to instead of the real reason, which is _returned_ by the `pthread_*` functions.

TedLyngmo:
pthread functions return the error instead of setting errno.

Fixes #2525


zuiderkwast (approved):
You're right! Thanks.

I knew this, but I hadn't noticed this programming error until now.

PR #2527, opened 2025-08-20
Fix slot range lists overlap to rewind the nested list again 

sarthakaggarwal97:
In the current implementation, the second list was never rewound again once it was iterated. So for the first element of `ranges1`, `ranges2` was iterated fully. But when the second element of `ranges1` was processed, the `ranges2` was not rewound again.

With this change, for every element of `ranges1`, we start from the beginning for `ranges2`


ranshid (approved):
LGTM. Thank you!

hpatro (approved):

PR #2530, opened 2025-08-20
bio.c: Organize all worker data in a struct

### LINKED ISSUE #2529: Reorganize `bio.c` to keep worker data in a `struct`
Instead of having a number of arrays with data for each worker, create _one_ array of `bio_worker_data` and send in a pointer to the right `bio_worker_data` in `pthread_create`. That'd be cleaner and safer than letting a `void*` carry the value of the worker number.

TedLyngmo:
This gets rid of the need to use a void* as a carrier for the worker number. Instead a pointer to the relevant worker data is passed to the started thread.

Fixes #2529

### DIFF
--- a/src/bio.c
+++ b/src/bio.c
@@ -87,7 +87,7 @@ static pthread_mutex_t bio_mutex[BIO_WORKER_NUM];
 static pthread_cond_t bio_newjob_cond[BIO_WORKER_NUM];
 static list *bio_jobs[BIO_WORKER_NUM];
 static unsigned long bio_jobs_counter[BIO_NUM_OPS] = {0};
-static __thread unsigned long bio_thread_id = 0;
+static _Thread_local unsigned long bio_thread_id = 0;
 
 /* This structure represents a background Job. It is only used locally to this
  * file as the API does not expose the internals at all. */
--- a/src/bio.c
+++ b/src/bio.c
@@ -65,15 +65,6 @@
 #include "bio.h"
 #include <stdatomic.h>
 
-static char *bio_worker_title[] = {
-    "bio_close_file",
-    "bio_aof",
-    "bio_lazy_free",
-    "bio_rdb_save",
-};
-
-#define BIO_WORKER_NUM (sizeof(bio_worker_title) / sizeof(*bio_worker_title))
-
 static unsigned int bio_job_to_worker[] = {
     [BIO_CLOSE_FILE] = 0,
     [BIO_AOF_FSYNC] = 1,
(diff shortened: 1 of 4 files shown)


src/bio.c
```diff
@@ -82,12 +73,24 @@ static unsigned int bio_job_to_worker[] = {
...
+    {"bio_close_file"},
+    {"bio_aof"},
+    {"bio_lazy_free"},
+    {"bio_rdb_save"}};
+static const bio_worker_data *const bio_worker_end = bio_workers + (sizeof bio_workers / sizeof *bio_workers);
+#define BIO_WORKER_NUM(X) ((X) - bio_workers)
+
 static unsigned long bio_jobs_counter[BIO_NUM_OPS] = {0};
-static __thread unsigned long bio_thread_id = 0;
+static _Thread_local size_t bio_worker_num = 0;
```
ranshid, line 100: I like the use of c11 thread local. maybe we should use this PR to also replace all the existing __thread uses (not many) as we are c11 compatible and although we do only support gcc/clang it would be nice to remove this dependency

src/bio.c
```diff
@@ -82,12 +73,24 @@ static unsigned int bio_job_to_worker[] = {
...
+    list *bio_jobs;
+} bio_worker_data;
+
+static bio_worker_data bio_workers[] = {
+    {"bio_close_file"},
+    {"bio_aof"},
+    {"bio_lazy_free"},
+    {"bio_rdb_save"}};
+static const bio_worker_data *const bio_worker_end = bio_workers + (sizeof bio_workers / sizeof *bio_workers);
+#define BIO_WORKER_NUM(X) ((X) - bio_workers)
```
ranshid: In general I do not find it bad to wrap this as a macro. However I would suggest to make this a static function.
why?
1. macro's expansion is int (IIRC) and in some cases you might want to get an unsigned value out of it and it will force the user to case. a function can return unsigned
2. it might cause compiler warnings if the passes worker is const... probably not something that will happen (unless we pass it the bio_worker_end ) but a function can convert the input pointer correctly IMO.
3. a function is a good place to place asserts and/or return some invalid value in cases the user provides some bad pointer which is not a bio_workers

ranshid (commented):
Mostly LGTM. Left 2 minor comments.

Thank you for making this improvement!

src/bio.c
TedLyngmo: Sure thing, I'll get at it when I get back from work!

src/bio.c
TedLyngmo: I agree 100% - will fix that too!

TedLyngmo:
Hmm, I can't see any of the failures being related to my changes. Can the tests be retriggered or do I need to force-push a dummy-change?

src/bio.c
```diff
@@ -82,12 +73,28 @@ static unsigned int bio_job_to_worker[] = {
...
+    {"bio_close_file"},
+    {"bio_aof"},
+    {"bio_lazy_free"},
+    {"bio_rdb_save"},
+};
+static const bio_worker_data *const bio_worker_end = bio_workers + (sizeof bio_workers / sizeof *bio_workers);
+
+static size_t bioWorkerNum(const bio_worker_data *const bwd) {
+    return bwd - bio_workers;
+}
```
ranshid: small suggestion:

```suggestion
static size_t bioWorkerNum(const bio_worker_data *const bwd) {
    serverAssert(bwd >= bio_workers && bwd < bio_worker_end); // ensure pointer is valid
    return (size_t)(bwd - bio_workers);
}
```

ranshid (approved):
almost done. just a small comment about assertion.
Placing a pending approval

src/bio.c
TedLyngmo: I thought about that too (and It'll most probably work on all modern archs) but if the pointer is outside the valid range that test has UB in itself, so I didn't add it. I don't think there is _any_ way to do this check _with_ defined behavior if the pointer isn't already guaranteed to be in the range - but the pragmatic approach could be to add the test anyway. 😄 

Another way could be to make the functions that uses `bioWorkerNum` `static` so they can't be called from the outside. That way it's easy to reason about it and there will never be a case when `bwd` is outside the range - but making them `static` will make at least one `.tcl` test case, `tests/integration/logging.tcl`, fail 😲

TedLyngmo: Hmm, what about this uglyness?
```
// ensure the pointer is valid - casting to uintptr_t to not make the comparison itself UB in case
// the pointer is outside the valid range
serverAssert((uintptr_t)bwd >= (uintptr_t)bio_workers && (uintptr_t)bwd < (uintptr_t)bio_worker_end);
```

Edit: I added that...

src/bio.c
```diff
@@ -82,12 +73,31 @@ static unsigned int bio_job_to_worker[] = {
...
+    {"bio_close_file"},
+    {"bio_aof"},
+    {"bio_lazy_free"},
+    {"bio_rdb_save"},
+};
+static const bio_worker_data *const bio_worker_end = bio_workers + (sizeof bio_workers / sizeof *bio_workers);
+
+static size_t bioWorkerNum(const bio_worker_data *const bwd) {
+    // Ensure the pointer is valid - casting to uintptr_t to not make the comparison itself UB in case
+    // the pointer is outside the valid range. It's a best effort thing anyway.
```
enjoy-binbin: we usually use this
```suggestion
    /* Ensure the pointer is valid - casting to uintptr_t to not make the comparison itself UB in case
     * the pointer is outside the valid range. It's a best effort thing anyway. */
```

TedLyngmo: > we usually use this

Oki! I changed that everywhere to make it consistent.

TedLyngmo: @enjoy-binbin I hope I understood your review correctly. I fixed the issue and corrected all other violations, but I can revert that change and clean up the commenting inconsistencies w.r.t. `//` and `/* ... */`  in a later patch if you prefer?

TedLyngmo: @enjoy-binbin I've now extracted the use of `_Thread_local` and the fix for the commenting inconsistencies and put that into a separate patch.

Issue #2504, opened 2025-08-18
[NEW] Use atomic slot migration in valkey-cli


zuiderkwast:
`valkey-cli` is the official cluster management tool. If we consider atomic slot migration to be fully supported, the cli should use atomic slot migration instead of the old SETSLOT method, in these commands:

```
valkey-cli --cluster reshard
valkey-cli --cluster rebalance
```

**Alternatives you've considered**

If we think that atomic slot migration is experimental, then we can wait and do this change in 9.1 or 10.0 instead.

**Additional information**

Even if not everyone uses valkey-cli as a cluster admin tool, it is used as a reference implementation when implementing another control planes such as kubernetes operators.

Follow-up of #23, #1949

enjoy-binbin:
Since the call is now quite simple, users only need to call MIGRATESLOTS and then check the cluster nodes or use GETSLOTMIGRATIONS to check the result.

I'm wondering what needs to be done in valkey-cli? We can change reshard and rebalance to use the new MIGRATESLOTS command, and then what? Since now the new migration is a async way, do we need to call GETSLOTMIGRATIONS in valkey-cli to block and check until the job succeeds or fails? Or should we just send the MIGRATESLOTS command and then quit (tell user to check MIGRATESLOTS itself some other times)?

zuiderkwast:
> I'm wondering what needs to be done in valkey-cli? We can change reshard and rebalance to use the new MIGRATESLOTS command, and then what? Since now the new migration is a async way, do we need to call GETSLOTMIGRATIONS in valkey-cli to block and check until the job succeeds or fails?

I think we should try to keep the existing behavior of valkey-cli, to block until reharding/rebalancing is complete. Idea: check, sleep 1 second, loop.

I guess some control plane tools use valkey-cli for this stuff in non-interactive mode. We can try to not break them.

murphyjacob4:
I have something locally in the valkey-cli. I can polish it up and send the PR.

PR #2533, opened 2025-08-21
Separate RDB snapshotting from atomic slot migration

enjoy-binbin:
When we adding atomic slot migration in #1949, we reused a lot of rdb save code,
it was an easier way to implement ASM in the first time, but it comes with some
side effect. Like we are using CHILD_TYPE_RDB to do the fork, we use rdb.c/rdb.h
function to save the snapshot, these mess up the logs (we will print some logs
saying we are doing RDB stuff) and mess up the info fields (we will say we are
rdb_bgsave_in_progress but actually we are doing slot migration).

In addition, it makes the code difficult to maintain. The rdb_save method uses
a lot of rdb_* variables, but we are actually doing slot migration. If we want
to support one fork with multiple target nodes, we need to rewrite these code
for a better cleanup.

Note that the changes to rdb.c/rdb.h are reverting previous changes from when
we was reusing this code for slot migration. The slot migration snapshot logic
is similar to the previous diskless replication. We use pipe to transfer the
snapshot data from the child process to the parent process.

Interface changes:
- New slot_migration_fork_in_progress info field.
- New cow_size field in CLUSTER GETSLOTMIGRATIONS command.
- Also add slot migration fork to the cluster class trace latency.


src/cluster_migrateslots.c
```diff
@@ -1275,6 +1371,7 @@ void clusterHandleSlotExportBackgroundSaveDone(int bgsaveerr) {
             continue;
         }
         if (bgsaveerr == C_OK) {
+            connNonBlock(job->client->conn);
```
murphyjacob4: This shouldn't be needed with the pipe based solution

```suggestion
```

src/replication.c
```diff
@@ -1820,6 +1813,94 @@ void rdbPipeReadHandler(struct aeEventLoop *eventLoop, int fd, void *clientData,
...
+    /* Remove the write handler and setup the pipe read handler. */
+    connSetWriteHandler(conn, NULL);
+    target->repl_data->repl_last_partial_write = 0;
+    if (aeCreateFileEvent(server.el, server.slot_migration_pipe_read, AE_READABLE, slotMigrationPipeReadHandler, NULL) == AE_ERR) {
+        serverPanic("Unrecoverable error creating server.slot_migration_pipe_read file event.");
+    }
+}
+
+/* Called in slot migration source, when there's data to read from the child's slot migration pipe. */
+void slotMigrationPipeReadHandler(struct aeEventLoop *eventLoop, int fd, void *clientData, int mask) {
```
murphyjacob4, line 1846: Given slot migration pipe read/write handlers are effectively the same as rdb pipe read/write handlers (just with one connection instead of potentially many connections) - should we just have a single "pipeReadHandler" and "pipeWriteHandler" and then add something like "server.pipePurpose" which could be "PIPE_PURPOSE_RDB" or "PIPE_PURPOSE_SLOT_MIGRATION" (for logging)?

Although, if we are going to move to a "dual channel" style transfer for slot migration, then perhaps it isn't worth the effort.

src/cluster_migrateslots.c
```diff
@@ -1231,34 +1233,128 @@ int childSnapshotForSyncSlot(int req, rio *rdb, void *privdata) {
...
-    rioWriteBulkString(rdb, "SYNCSLOTS", 9);
-    rioWriteBulkString(rdb, "SNAPSHOT-EOF", 12);
+    rioWrite(aof, "*3\r\n", 4);
+    rioWriteBulkString(aof, "CLUSTER", 7);
+    rioWriteBulkString(aof, "SYNCSLOTS", 9);
+    rioWriteBulkString(aof, "SNAPSHOT-EOF", 12);
     return C_OK;
 }
 
+void killSlotMigrationChild(void) {
```
murphyjacob4, line 1296: I think we need to substitute:

```
        if (job->state == SLOT_EXPORT_SNAPSHOTTING) killRDBChild();
```

With

```
        if (job->state == SLOT_EXPORT_SNAPSHOTTING) killSlotMigrationChild();
```

In finishSlotMigrationJob

enjoy-binbin: right, i forgot to replace it. I also added some changes, please take a check.

src/replication.c
enjoy-binbin: yes, i think we will get into dual channel at some point in the future. Such diffs are mostly newly added diffs and may be easier to handle in the future.

murphyjacob4: Sounds good to me

src/cluster_migrateslots.c
murphyjacob4: Looks good 👍

src/cluster_migrateslots.c
```diff
@@ -1231,34 +1233,128 @@ int childSnapshotForSyncSlot(int req, rio *rdb, void *privdata) {
...
+    int statloc;
+    /* No slot migration child? return. */
+    if (server.child_type != CHILD_TYPE_SLOT_MIGRATION) return;
+    /* Kill slot migration child, wait for child exit. */
+    serverLog(LL_NOTICE, "Killing running slot migration child: %ld", (long)server.child_pid);
+    if (kill(server.child_pid, SIGUSR1) != -1) {
+        while (waitpid(-1, &statloc, 0) != server.child_pid);
+    }
+    resetChildState();
+    serverLog(LL_NOTICE, "Slot migration child %ld killed", (long)server.child_pid);
```
murphyjacob4:
```suggestion
    serverLog(LL_NOTICE, "Killing running slot migration child: %ld", (long)server.child_pid);
    if (kill(server.child_pid, SIGUSR1) != -1) {
        while (waitpid(-1, &statloc, 0) != server.child_pid);
    }
    serverLog(LL_NOTICE, "Slot migration child %ld killed", (long)server.child_pid);
    resetChildState();
```
This should fix the log:

```
20601:M 23 Aug 2025 08:23:48.111 * Slot migration child -1 killed
```

src/cluster_migrateslots.c
```diff
@@ -1231,34 +1233,128 @@ int childSnapshotForSyncSlot(int req, rio *rdb, void *privdata) {
...
+         * get a write error and exit. */
+        close(server.rdb_pipe_read);
+
+        serverSetProcTitle("valkey-slot-migration-to-target");
+        serverSetCpuAffinity(server.slot_migration_cpulist);
+
+        int retval = childSnapshotForSyncSlot(&aof, job);
+        if (retval == C_OK && rioFlush(&aof) == 0) retval = C_ERR;
+        if (retval == C_OK) {
+            sendChildCowInfo(CHILD_INFO_TYPE_SLOT_MIGRATION_COW_SIZE, "slot migration");
```
murphyjacob4: In rewriteSlotToAppendOnlyFileRio - we use "AOF rewrite" which leads to a bit of a confusing log:

```
27259:C 23 Aug 2025 08:23:48.036 - Fork CoW for AOF rewrite: current 0 MB, peak 0 MB, average 0 MB
27259:C 23 Aug 2025 08:23:48.042 * Fork CoW for slot migration: current 0 MB, peak 0 MB, average 0 MB
```

Probably we can pass in the name to `rewriteSlotToAppendOnlyFileRio`?

src/cluster_migrateslots.c
```diff
@@ -1231,34 +1233,128 @@ int childSnapshotForSyncSlot(int req, rio *rdb, void *privdata) {
...
+    /* No slot migration child? return. */
+    if (server.child_type != CHILD_TYPE_SLOT_MIGRATION) return;
+    /* Kill slot migration child, wait for child exit. */
+    serverLog(LL_NOTICE, "Killing running slot migration child: %ld", (long)server.child_pid);
+    if (kill(server.child_pid, SIGUSR1) != -1) {
+        while (waitpid(-1, &statloc, 0) != server.child_pid);
+    }
+    resetChildState();
+    serverLog(LL_NOTICE, "Slot migration child %ld killed", (long)server.child_pid);
+    clusterHandleSlotExportBackgroundSaveDone(C_ERR);
```
murphyjacob4: 1. I think we need to call backgroundSlotMigrationDoneHandler instead to ensure pipes are closed and reset
2. Should we merge backgroundSlotMigrationDoneHandler and clusterHandleSlotExportBackgroundSaveDone? I don't really see a need for both, WDYT?

I think 1 should solve the panic in the tests:

```
20601:M 23 Aug 2025 08:23:48.236 # ==> cluster_migrateslots.c:1271 'server.slot_migration_pipe_read == -1 && server.slot_migration_child_exit_pipe == -1' is not true
```

src/cluster_migrateslots.c
enjoy-binbin: good catch, i was thinking if we can add a flag to rio.flags to indicate that it is a slot migration? So that we don't need to change function

src/cluster_migrateslots.c
enjoy-binbin: yes, we can merge it, there is no need to keep it now after we do the change

src/cluster_migrateslots.c
murphyjacob4: Looks good to me

src/cluster_migrateslots.c
murphyjacob4: Looks good! thanks

murphyjacob4 (approved):
@madolson - I think you mentioned you had some opinions on the INFO fields during this weeks meeting?

src/server.c
```diff
@@ -6029,7 +6040,9 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "aof_last_write_status:%s\r\n", (server.aof_last_write_status == C_OK && aof_bio_fsync_status == C_OK) ? "ok" : "err",
                 "aof_last_cow_size:%zu\r\n", server.stat_aof_cow_bytes,
                 "module_fork_in_progress:%d\r\n", server.child_type == CHILD_TYPE_MODULE,
-                "module_fork_last_cow_size:%zu\r\n", server.stat_module_cow_bytes));
+                "module_fork_last_cow_size:%zu\r\n", server.stat_module_cow_bytes,
+                "slot_migration_fork_in_progress:%d\r\n", server.child_type == CHILD_TYPE_SLOT_MIGRATION,
```
madolson: I think boolean flags can almost always include more information. In this case, we can have how long the current slot migration is in progress. Although, we have a bunch of similar fields for in_progress already, so I think it's OK to add one more.

src/server.c
```diff
@@ -6029,7 +6040,9 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
                 "aof_last_write_status:%s\r\n", (server.aof_last_write_status == C_OK && aof_bio_fsync_status == C_OK) ? "ok" : "err",
                 "aof_last_cow_size:%zu\r\n", server.stat_aof_cow_bytes,
                 "module_fork_in_progress:%d\r\n", server.child_type == CHILD_TYPE_MODULE,
-                "module_fork_last_cow_size:%zu\r\n", server.stat_module_cow_bytes));
+                "module_fork_last_cow_size:%zu\r\n", server.stat_module_cow_bytes,
```
madolson, line 6043: I also think we should get rid of the global COW size, since the amount of time spent for each slot migration will change, I'm not sure having how much COW memory is useful. I think we should add it to the `GETSLOTMIGRATIONS` command instead.

src/server.c
```diff
@@ -6029,7 +6040,9 @@ sds genValkeyInfoString(dict *section_dict, int all_sections, int everything) {
...
+                "slot_migration_fork_in_progress:%d\r\n", server.child_type == CHILD_TYPE_SLOT_MIGRATION,
+                "slot_migration_fork_last_cow_size:%zu\r\n", server.stat_slot_migration_cow_bytes));
```
```suggestion
                "slot_migration_fork_in_progress:%d\r\n", server.child_type == CHILD_TYPE_SLOT_MIGRATION));
```
I also think we should get rid of the global COW size, since the amount of time spent for each slot migration will change, I'm not sure having how much COW memory is useful. I think we should add it to the `GETSLOTMIGRATIONS` command instead.

valkey.conf
```diff
@@ -2602,8 +2602,11 @@ jemalloc-bg-thread yes
 # Set aof rewrite child process to cpu affinity 8,9,10,11:
 # aof-rewrite-cpulist 8-11
 #
-# Set bgsave child process to cpu affinity 1,10,11
+# Set bgsave child process to cpu affinity 1,10,11:
 # bgsave-cpulist 1,10-11
+#
+# Set slot migration child process to cpu affinity 1,8,9,10:
+# slot-migration-cpulist 1,8-10
```
madolson: Since slot migration and full-sync bgsave are functionally the same operation, I would have expected slot-migration-cpu list to use the same cpu-list as bgsave. I would rather not add another configuration for this.

To get this PR merged, I would like to suggest we don't add the new config right now.

src/cluster_migrateslots.c
```diff
@@ -2043,6 +2045,8 @@ void clusterCommandGetSlotMigrations(client *c) {
         addReplyBulkCString(c, slotMigrationJobStateToString(job->state));
         addReplyBulkCString(c, "message");
         addReplyBulkCString(c, job->status_msg ? job->status_msg : "");
+        addReplyBulkCString(c, "last_cow_size");
```
zuiderkwast: last? There is only one fork per job. I guess we can skip "last" and just say "cow_size"?

src/cluster_migrateslots.c
```diff
@@ -1975,6 +2064,8 @@ void clusterCommandGetSlotMigrations(client *c) {
         addReplyBulkCString(c, slotMigrationJobStateToString(job->state));
         addReplyBulkCString(c, "message");
         addReplyBulkCString(c, job->status_msg ? job->status_msg : "");
+        addReplyBulkCString(c, "cow_size");
+        addReplyLongLong(c, (long long)job->stat_cow_bytes);
```
enjoy-binbin, line 2112: let's also expose output buffer size in here? We may be missing this information right now to allow people to monitor its progress. (Or we could do this in another PR, by exposing both the import slot client and the export slot client in the client info, i.e. adding a client flag. But i can't think of a good flag char, since import source already taken 'I' char)

enjoy-binbin: we can probably go with 'i' and 'e', stand for import or export. But no_evict already taken 'e'.

So we can go with 'i' and 'm', stand for importing or migrating as the old word.s

zuiderkwast: To monitor it, use CLIENT LIST? :thinking: I guess it's possible, yes, but maybe it's easier to use if we put some progress information in CLUSTER SLOTMIGRATIONS.

Maybe we can do it at the same time as #2504, if we want valkey-cli to print some progress indicator in interactive mode (if stdout is a TTY).

src/aof.c
```diff
@@ -2302,7 +2302,11 @@ int rewriteSlotToAppendOnlyFileRio(rio *aof, int db_num, int hashslot, size_t *k
         if (key_count && ((*key_count)++ & 1023) == 0) {
             long long now = mstime();
             if (now - updated_time >= 1000) {
-                sendChildInfo(CHILD_INFO_TYPE_CURRENT_INFO, *key_count, "AOF rewrite");
+                if (aof->flags & RIO_FLAG_SLOT_MIGRATION_AOF) {
+                    sendChildInfo(CHILD_INFO_TYPE_CURRENT_INFO, *key_count, "Slot migration");
+                } else {
+                    sendChildInfo(CHILD_INFO_TYPE_CURRENT_INFO, *key_count, "AOF rewrite");
+                }
```
PingXie: this changes introduce unnecessary coupling and adds cognitive load on the future developer. I don't think `aof.c` should understand what "slot migration" is. the `pname` should be opaque to aof.c and aof.c should just blindly passes along.
```suggestion
                sendChildInfo(CHILD_INFO_TYPE_CURRENT_INFO, *key_count, aof->pname);
```

src/cluster_migrateslots.c
```diff
@@ -1247,40 +1250,128 @@ int childSnapshotForSyncSlot(int req, rio *rdb, void *privdata) {
...
+    rioWriteBulkString(aof, "CLUSTER", 7);
+    rioWriteBulkString(aof, "SYNCSLOTS", 9);
+    rioWriteBulkString(aof, "SNAPSHOT-EOF", 12);
     return C_OK;
 }
 
+/* Kill the slot migration child using SIGUSR1 (so that the parent will know
+ * the child did not exit for an error, but because we wanted), and performs
+ * the cleanup needed. */
+void killSlotMigrationChild(void) {
```
PingXie, line 1296: can we generalize this function and dedup `killRDBChild` instead?

```c
void killChild(int type) {
    /* Not the right child? return. */
    if (server.child_type != type) return;
    serverLog(LL_NOTICE, "Killing running %s: %ld", ..., (long)server.child_pid);

    /* Because we are not using here waitpid (like we have in killAppendOnlyChild
     * and TerminateModuleForkChild), all the cleanup operations is done by
     * checkChildrenDone, that later will find that the process killed. */
    kill(server.child_pid, SIGUSR1);
}
```

src/aof.c
enjoy-binbin: The rdbSaveDb in rdb.c also use this trick (or similar flags trick). aof.c will anyway need to understand the "slot migration" in the furture as long as we use this format (like module event something like that). We may need new aofflags in that time. 
```
ssize_t rdbSaveDb(rio *rdb, int dbid, int rdbflags, long *key_counter) {
    ssize_t written = 0;
    ssize_t res;
    kvstoreIterator *kvs_it = NULL;
    static long long info_updated_time = 0;
    char *pname = (rdbflags & RDBFLAGS_AOF_PREAMBLE) ? "AOF rewrite" : "RDB";
```

src/cluster_migrateslots.c
enjoy-binbin: I've thought about this before, but I don't want to touch other code back then. This function will only be called in places where it's explicitly checked (for example, after checking child_type we kill the child). So i guess some duplicate code isn't that bad?

src/aof.c
enjoy-binbin: Do you want to add the pname to rio? I think we should either use rio->flags (although it is a bit less elegant) or similar to rdbflags we use a new aofflags

madolson: @enjoy-binbin We talked about this in monday meeting. We can punt this, but we want to merge this PR for rc2.

enjoy-binbin: I realize that rewriteSlotToAppendOnlyFileRio is currently only used in the slot migration, so i am gonna remove the flags first. We can change it again in the future if needed. @murphyjacob4 please ack again so that we can merge it.

murphyjacob4: Ack, seems reasonable, IMO anyone reusing the function can add the parameterization then

PR #2537, opened 2025-08-22
deflake test: relax time requirement in hash ttl test

rainsupreme:
tiny change. It failed once for me (a little time passed and it returned 2 seconds instead of 3), so I figured it's probably a little flaky for others too


ranshid:
I think this test should just use HEXPIREAT and HEXPIRETIME to verify

ranshid (approved):
LGTM. thank you!

PR #2538, opened 2025-08-23
Consistently use static_assert across code

madolson:
In C99, we had to use `#define static_assert(expr, lit) extern char __static_assert_failure[(expr) ? 1 : -1]` for static assertions. However, we now have native support for static_assert with _Static_assert. Previously one of the correct #defines was getting called first, setting it to _Static_assert, however after https://github.com/valkey-io/valkey/commit/65215e5378db7b47c3460567da134ab7f0bfeb58 the first import defining the symbol was "serverassert.h", which included the old style.

This change removes all unnecessary imports and always defines static_assert as _Static_assert.

### DIFF
--- a/src/dict.c
+++ b/src/dict.c
@@ -54,10 +54,6 @@
 #include "monotonic.h"
 #include "config.h"
 
-#ifndef static_assert
-#define static_assert(expr, lit) _Static_assert(expr, lit)
-#endif
-
 #define UNUSED(V) ((void)V)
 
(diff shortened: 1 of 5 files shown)


enjoy-binbin (approved):

ranshid (commented):
LGTM

I like following c11!

zuiderkwast (approved):

hpatro (approved):

PR #2539, opened 2025-08-24
Use _Thread_local instead of __thread

TedLyngmo:
Extracted from https://github.com/valkey-io/valkey/pull/2530


enjoy-binbin (changes requested):
thanks, i am afraid i have to ask you remove the changes about he multiline comments. 

If it already exists, we try to avoid modifying it (this will reduce diffs and reduce git blame logs). If we happen to modify it, we can change it to multilines in the same one diff.

TedLyngmo:
> thanks, i am afraid i have to ask you remove the changes about he multiline comments.

No problem, removed.

TedLyngmo:
Merged via 9d682bad0bb731de8e21e15e4a1795c73aee9d0d

PR #2541, opened 2025-08-24
Update reply schema for LMOVE and BLMOVE

### LINKED ISSUE #2532: [BUG] Reply_schema for LMOVE does not include the fact it can return nil
**Describe the bug**

reply_schema for LMOVE in json file does not include the fact it can return `nil`

**To reproduce**

https://github.com/valkey-io/valkey/blob/84fd6264190094c045de488308b905853683d135/src/commands/lmove.json#L55-L58

**Expected behavior**

Should the reply_schema be this instead
```json
    "oneOf": [
        {
            "description": "The element being popped and pushed.",
            "type": "string"
        },
        {
            "description": "Source does not exist.",
            "type": "null"
        }
    ]
```
**Additional information**

Any additional information that is relevant to the problem.
This was found here https://github.com/valkey-io/valkey-swift/issues/184

adam-fowler:
Both LMOVE and BLMOVE can return null values because the source key is empty.
PR changes include 
- change the LMOVE reply_schema to include the possibility of a nil return value
- Add comment to BLMOVE reply_schema to indicate it can return nil because the source does not exist

This fixes #2532

### DIFF
--- a/src/commands/blmove.json
+++ b/src/commands/blmove.json
@@ -60,7 +60,7 @@
                     "type": "string"
                 },
                 {
-                    "description": "Operation timed-out",
+                    "description": "Operation timed-out or the source does not exist.",
                     "type": "null"
                 }
             ]
--- a/src/commands/blmove.json
+++ b/src/commands/blmove.json
@@ -60,7 +60,7 @@
                     "type": "string"
                 },
                 {
-                    "description": "Operation timed-out or the source does not exist.",
+                    "description": "Operation timed-out or the command is issued from a transaction or a script and the source does not exist.",
                     "type": "null"
                 }
             ]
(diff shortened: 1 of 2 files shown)


enjoy-binbin (approved):

src/commands/blmove.json
```diff
@@ -60,7 +60,7 @@
                     "type": "string"
                 },
                 {
-                    "description": "Operation timed-out",
+                    "description": "Operation timed-out or the source does not exist.",
```
ranshid: the source does not exist is not the exact case.
```suggestion
                    "description": "Operation timed-out or the command is issued from a transaction or a script and the source does not exist.",
```

ranshid (approved):
LGTM. a small comment suggestion

PR #665, opened 2024-06-18
Support RDB compatability with Redis 7.2.4 RDB format

ranshid:
This PR makes our current RDB format compatible with the Redis 7.2.4 RDB format. there are 2 changes introduced in this PR:
1. Move back the RDB version to 11
2. Make slot info section persist as AUX data instead of dedicated section.

We have introduced slot-info as part of the work to replace cluster metadata with slot specific dictionaries. This caused us to bump the RDB version and thus we prevent downgrade (which is conceptualy O.K but better be prevented). We do not require the slot-info section to exist, so making it an AUX section will help suppport version downgrade from Valkey 8.

fixes: [#645](https://github.com/valkey-io/valkey/issues/645)

NOTE: tested manually by:
1. connecting Redis 7.2.4 replica to a Valkey 8(RC) 
2. upgrade/downgrade Redis 7.2.4 cluster and Valkey 8(RC) cluster

PR #2543, opened 2025-08-25
Relaxed RDB check for foreign RDB formats

zuiderkwast:
In #1604, we attempt to read future Valkey RDB formats, but we rejected foreign RDB formats, because of the risk that the opcodes and types added by other projects collide with the new types and opcodes added in recent Valkey versions.

This change accepts foreign RDB versions but limits the types and opcodes to the ones that we can understand, to prevent misinterpretation of types/opcodes which could lead to undefined behavior. If unsupported RDB types or opcodes are seen, we error out.

Additional changes:

* Improve error reporting when encountering unknown RDB types in relaxed version check mode.
* Tests for loading various RDB files.
* Improvement to valkey-check-rdb to accept future and foreign RDB versions, including tests.


src/rdb.c
```diff
@@ -3103,6 +3103,13 @@ int rdbLoadRioWithLoadingCtx(rio *rdb, int rdbflags, rdbSaveInfo *rsi, rdbLoadin
         /* Read type. */
         if ((type = rdbLoadType(rdb)) == -1) goto eoferr;
 
+        /* Safeguard for unknown foreign opcode interpretations. */
+        if (is_redis_magic && type >= RDB_FOREIGN_TYPE_MIN && type <= RDB_FOREIGN_TYPE_MAX) {
+            serverLog(LL_WARNING, "Can't handle foreign type or opcode %d in RDB %d",
```
ranshid:
```suggestion
            serverLog(LL_WARNING, "Can't handle foreign type or opcode %d in RDB with version %d",
```

ranshid (approved):
LGTM - small comment about message

src/rdb.h
```diff
@@ -131,7 +136,9 @@ typedef struct rdbSnapshotOptions {
 /* Test if a type is an object type. */
 #define rdbIsObjectType(t) (((t) >= 0 && (t) <= 7) || ((t) >= 9 && (t) < RDB_TYPE_LAST))
 
-/* Special RDB opcodes (saved/loaded with rdbSaveType/rdbLoadType). */
+/* Special RDB opcodes (saved/loaded with rdbSaveType/rdbLoadType).
+ * These are special RDB types, but they start from 255 and grow down. */
+#define RDB_OPCODE_SLOT_INFO 244       /* Foreign slot info, safe to ignore. */
```
enjoy-binbin, line 145: we don't support RDB_OPCODE_SLOT_INFO? Didn't it exist before the fork, like we added it when we support per dict per slot.

zuiderkwast: We had it in unstable but not in any released Valkey version, so officially we don't really support it. We need to be able to at least read and ignore this to be able to read any redis rdb for cluster mode.

enjoy-binbin: ok, i see, we now use the slot-info aux field. Anyway, since now we can read it, do we want to use this RDB_OPCODE_SLOT_INFO code to resize the hashtable? Like if we can read (or understand it), why should we ignore it? I believe Redis has released this op code.

zuiderkwast: We can, but this is not covered by our tests. It will be a few more lines of untested code. But I guess it's OK because it's similar to the aux field logic.

It will be like this, as we reverted it in https://github.com/valkey-io/valkey/pull/665/files :

```c
            uint64_t slot_id, slot_size, expires_slot_size;
            if ((slot_id = rdbLoadLen(rdb, NULL)) == RDB_LENERR) goto eoferr;
            if ((slot_size = rdbLoadLen(rdb, NULL)) == RDB_LENERR) goto eoferr;
            if ((expires_slot_size = rdbLoadLen(rdb, NULL)) == RDB_LENERR) goto eoferr;
            if (server.cluster_enabled) {
                if (slot_size) kvstoreHashtableExpand(db->keys, slot_id, slot_size);
                if (expires_slot_size) kvstoreHashtableExpand(db->expires, slot_id, expires_slot_size);
                should_expand_db = 0;
            }
```

zuiderkwast: Do we care about malicious RDB files? I can imagine an RDB file with slot > 16383 or with very huge slot sizes can cause valkey to crash.

src/cluster.c
```diff
@@ -166,8 +166,9 @@ int verifyDumpPayload(unsigned char *p, size_t len, uint16_t *rdbver_ptr) {
     if (rdbver_ptr) {
         *rdbver_ptr = rdbver;
     }
-    if ((rdbver >= RDB_FOREIGN_VERSION_MIN && rdbver <= RDB_FOREIGN_VERSION_MAX) ||
-        (rdbver > RDB_VERSION && server.rdb_version_check == RDB_VERSION_CHECK_STRICT)) {
+    if (server.rdb_version_check == RDB_VERSION_CHECK_STRICT &&
+        (rdbver > RDB_VERSION ||
+         (rdbver >= RDB_FOREIGN_VERSION_MIN && rdbver <= RDB_FOREIGN_VERSION_MAX))) {
```
hpatro: Would prefer moving this to a macro/method.

```
#define RDB_VERSION_NOT_VALKEY_SUPPORTED        (rdbver > RDB_VERSION ||
         (rdbver >= RDB_FOREIGN_VERSION_MIN && rdbver <= RDB_FOREIGN_VERSION_MAX))) {
```

src/cluster.c
```diff
@@ -251,14 +253,28 @@ void restoreCommand(client *c) {
...
+    if (type == -1) {
+        addReplyError(c, "Bad data format");
+        return;
+    }
+
+    /* If it's a foreign RDB format. Only accept old data types that we know the
+     * meaning of. */
+    if (rdbver >= RDB_FOREIGN_VERSION_MIN && rdbver <= RDB_FOREIGN_VERSION_MAX &&
+        type >= RDB_FOREIGN_TYPE_MIN) {
+        addReplyError(c, "Unsupported foreign data format");
```
hpatro: Would it be more helpful for administrator to figure which data type crept in ?
```suggestion
        addReplyErrorFormat(c, "Unsupported foreign data type: %d", type);
```

hpatro (commented):
Mostly LGTM. Will be good to have coverage via adding foreign RDB version/type not supported. Might need to start a Redis server and take some snapshots and copy it our test assets.

src/cluster.c
zuiderkwast: Sure.

src/cluster.c
zuiderkwast: Sure, I'll move the logic to `rdb.{c,h}` and de-duplicate the logic with the checks used for reading RDB files in `rdb.c`.

src/rdb.h
enjoy-binbin: yes, i guess we should, we should add a check around it

zuiderkwast:
> Mostly LGTM. Will be good to have coverage via adding foreign RDB version/type not supported. Might need to start a Redis server and take some snapshots and copy it our test assets.

@hpatro I have added various RDB files and tests. I created an RDB 12 with the commit just before #665. We actually had RDB 12 in valkey unstable before that PR was merged.

The other files RDB 75 and RDB 987 with and without unknonw types were created by editing the binary files. The unknown type is just a string where I changed the type byte.

src/rdb.c
```diff
@@ -3352,6 +3389,10 @@ int rdbLoadRioWithLoadingCtx(rio *rdb, int rdbflags, rdbSaveInfo *rsi, rdbLoadin
             if (error == RDB_LOAD_ERR_EMPTY_KEY) {
                 if (empty_keys_skipped++ < 10) serverLog(LL_NOTICE, "rdbLoadObject skipping empty key: %s", key);
                 sdsfree(key);
+            } else if (error == RDB_LOAD_ERR_UNKNOWN_TYPE) {
+                sdsfree(key);
+                serverLog(LL_WARNING, "Unknown type or opcode when loading DB. Unrecoverable error, aborting now.");
```
enjoy-binbin, line 3394: do we have opcode in rdbLoadObject? I guess it only have type and encoding info when we in here?
```suggestion
                serverLog(LL_WARNING, "Unknown object type when loading DB. Unrecoverable error, aborting now.");
```

enjoy-binbin (approved):

src/rdb.c
```diff
@@ -3215,7 +3252,7 @@ int rdbLoadRioWithLoadingCtx(rio *rdb, int rdbflags, rdbSaveInfo *rsi, rdbLoadin
                     goto eoferr;
                 }
 
-                if (server.cluster_enabled) {
+                if (server.cluster_enabled && slot_id >= 0 && slot_id < CLUSTER_SLOTS) {
```
ranshid, line 3255: So in case the AUX is injecting illegal data for slot ID we will just avoid expending and that's it? 
I mean it was like this before, but maybe we should at least issue some warning about it?

src/rdb.c
zuiderkwast: Types are 1-22, opcodes are types in the range 245-255. If it's between 22 and 245, we don't know if it's an unknown type or an unknown opcode.

```
/* Special RDB opcodes (saved/loaded with rdbSaveType/rdbLoadType). */
#define RDB_OPCODE_FUNCTION2 245       /* function library data */
#define RDB_OPCODE_FUNCTION_PRE_GA 246 /* old function library data for 7.0 rc1 and rc2 */
#define RDB_OPCODE_MODULE_AUX 247      /* Module auxiliary data. */
#define RDB_OPCODE_IDLE 248            /* LRU idle time. */
#define RDB_OPCODE_FREQ 249            /* LFU frequency. */
#define RDB_OPCODE_AUX 250             /* RDB aux field. */
#define RDB_OPCODE_RESIZEDB 251        /* Hash table resize hint. */
#define RDB_OPCODE_EXPIRETIME_MS 252   /* Expire time in milliseconds. */
#define RDB_OPCODE_EXPIRETIME 253      /* Old expire time in seconds. */
#define RDB_OPCODE_SELECTDB 254        /* DB number of the following keys. */
#define RDB_OPCODE_EOF 255             /* End of the RDB file. */
```

tests/integration/rdb.tcl
```diff
@@ -69,6 +72,42 @@ start_server [list overrides [list "dir" $server_path \
     } $csv_dump
 }
 
+start_server_and_kill_it [list dir $server_path \
+                              dbfilename "encodings-rdb987-unknown-types.rdb" \
+                              rdb-version-check relaxed] {
+    test "RDB future version loading with unknown types, relaxed version check" {
+        wait_for_condition 50 100 {
+            [string match {*Fatal error loading*} \
```
ranshid: Maybe we should explicitly verify the point of the failing?
```suggestion
            [string match {*Unknown type or opcode when loading DB. Unrecoverable error, aborting now.*} \
```

tests/integration/rdb.tcl
```diff
@@ -69,6 +72,42 @@ start_server [list overrides [list "dir" $server_path \
...
+        assert_equal bar [r get foo]
+    }
+}
+
+start_server_and_kill_it [list dir $server_path \
+                              dbfilename "encodings-rdb75-unknown-types.rdb" \
+                              rdb-version-check relaxed] {
+    test "RDB foreign version loading with unknown types, relaxed version check" {
+        wait_for_condition 50 100 {
+            [string match {*Fatal error loading*} \
```
ranshid: Like before, maybe consider checking the specific failing reason is the foreign version + unknown type?
```suggestion
            [string match {*Can't handle foreign type or opcode*} \
```

ranshid (approved):
Overall LGTM.
left some minor comments to consider

src/rdb.c
zuiderkwast: If we consider RDB files untrusted, there are many ways to cause trouble. A very large size can cause a very large allocation, for example.

I added it because of previous comments (regarding the equivalent RDB 12 opcode, but I added it here too), but I don't want to escalate this further without a threat story.

zuiderkwast: https://github.com/valkey-io/valkey/pull/2543#discussion_r2308966969

ranshid: Sure. just to circle back (after merge) - I only meant to provide some warning, not to escalate any farther.

PR #2547, opened 2025-08-26
Do not migrate function in new atomic slot migration

enjoy-binbin:
If all cluster nodes have functions, slot migration will fail
since the target will return the function already exists error
when doing the FUNCTION LOAD.

And in addition, the target's replica could panic when it executes
the FUNCTION LOAD propagated from the primary (see propagation-error-behavior).

Introduced in #1949.

### DIFF
--- a/tests/unit/cluster/cluster-migrateslots.tcl
+++ b/tests/unit/cluster/cluster-migrateslots.tcl
@@ -403,6 +403,35 @@ start_cluster 3 3 {tags {logreqres:skip external:skip cluster} overrides {cluste
     set 16382_slot_tag "{4oi}"
     set 16383_slot_tag "{6ZJ}"
 
+    test "Slot migration won't migrate the functions" {
+        assert_does_not_resync {
+            # R 2 load a function then trigger a slot migration to R 0
+            populate 1 "$16383_slot_tag:" 1000 -2
+            R 2 function load {#!lua name=test
+                server.register_function('test', function() return 'hello1' end)
+            }
+
+            assert_match "OK" [R 2 CLUSTER MIGRATESLOTS SLOTSRANGE 16383 16383 NODE $node0_id]
+            wait_for_migration 0 16383
+
+            # Make sure R 0 does not have any function.
+            assert_match {} [R 0 function list]
+
+            # R 0 load a function then migrate the slot back to R 2
+            R 0 function load {#!lua name=test
+                server.register_function('test', function() return 'hello1' end)
+            }
+            assert_match "OK" [R 0 CLUSTER MIGRATESLOTS SLOTSRANGE 16383 16383 NODE $node2_id]
+            wait_for_migration 2 16383
+
+            # Cleanup for next test
+            assert_match "OK" [R 0 FLUSHDB SYNC]
+            assert_match "OK" [R 0 FUNCTION FLUSH SYNC]
+            assert_match "OK" [R 2 FLUSHDB SYNC]
+            assert_match "OK" [R 2 FUNCTION FLUSH SYNC]
+        }
+    }
+
     test "Single source import - one shot" {
         assert_does_not_resync {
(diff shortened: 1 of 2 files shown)


tests/unit/cluster/cluster-migrateslots.tcl
```diff
@@ -403,6 +403,34 @@ start_cluster 3 3 {tags {logreqres:skip external:skip cluster} overrides {cluste
     set 16382_slot_tag "{4oi}"
     set 16383_slot_tag "{6ZJ}"
 
+    test "Slot migration won't migrate the functions" {
+        assert_does_not_resync {
+            # R 2 load a function then trigger a slot migration to R 0
+            populate 1 "$16383_slot_tag:" 1000 -2
+            R 2 function load {#!lua name=test
+                server.register_function('test', function() return 'hello1' end)
+            }
```
murphyjacob4, line 413: Do we also need to test loading a function during the incremental stream?

It should be handled (since the command has no slot) but probably a good test case to check in this test

murphyjacob4 (approved):

zuiderkwast (approved):
Makes sense to me. (I didn't review the test.)

tests/unit/cluster/cluster-migrateslots.tcl
enjoy-binbin: good idea, i added it.

murphyjacob4 (approved):

tests/unit/cluster/cluster-migrateslots.tcl
murphyjacob4: Thanks, LGTM!

zuiderkwast:
I don't think we need to mention this in the release notes. We just modified (fixed) behavior that's not yet released.

enjoy-binbin:
> I don't think we need to mention this in the release notes. We just modified (fixed) behavior that's not yet released.

We release this in RC1, didn't we? So when we release RC2 we need to mention the fix?

zuiderkwast:
> > I don't think we need to mention this in the release notes. We just modified (fixed) behavior that's not yet released.
> 
> We release this in RC1, didn't we? So when we release RC2 we need to mention the fix?

Good point. But this was not documented behaviour. Should this PR be listed as a bugfix in the RC2 release notes?

madolson:
> Good point. But this was not documented behaviour. Should this PR be listed as a bugfix in the RC2 release notes?

I thought this was intended behavior. How are users supposed to migrate functions? Through the CLI?

zuiderkwast:
@madolson Right. Migrating functions was not intended but it happened and it could cause trouble, so it's a bug in RC 1 and this is bugfix to be released in RC 2. Correct?

Or don't we need to mention it in the release notes of rc2?

madolson:
We could mention it as a bug fix. I do wonder if we shouldn't just migrate the functions though? Instead of crashing we could send them with replace.

enjoy-binbin:
In our fork, back to 7.0, when we adapted the function, i do converted it to a replace so that the target node will have the function (Noted that we are adding an empty target node to the cluster when doing the migration).

And in the OSS version, at the time (7.0), i had the impression that functions were something that cluster administrators needed to load. For example, when adding a new node, the function needed to be loaded on all nodes (including the new one) (this is why we added function support to the CLI). In another scenario, in our OSS, the target node already exists in the cluster (in most cases), so it should already have the function.

I guess we should identify whether a function needs migration or not. Note that in the old setslot migration, we did not migrate the function either.

zuiderkwast:
@enjoy-binbin Agree, adding functions is more related to adding a new node to the cluster.

We could make `valkey-cli --cluster add-node` fetch the functions from another in the cluster, such as from the node it sends CLUSTER MEET to. The cli is more or less as a reference implementation for control planes. It's even used by some iiuc.

PR #2549, opened 2025-08-26
Roll backward downgrade compatibility from Redis 7.0/7.2 and Valkey 7.2/8.0 to Redis 6.0

satheeshaGowda:
===========NOT TO BE MERGED======

This patch extends Redis 6.0 to support  roll forward  upgrade and roll backward downgrade compatibility with 

- **RDB v9**: Redis 6.2 
- **RDB v10**: Redis 7.0
- **RDB v11**: Redis 7.2
- **RDB v11**: Valkey 7.2
- **RDB v11**: Valkey 8.0

### Listpack to Ziplist Conversion Process
1. Parse listpack header and validate structure
2. Walk through each listpack element
3. Extract element data (string or integer)
4. Add each element to new ziplist structure
5. Return converted ziplist for normal processing


src/cluster.c
```diff
@@ -4123,11 +4146,15 @@ sds clusterGenNodeDescription(clusterNode *node) {
     sds ci;
 
     /* Node coordinates */
-    ci = sdscatprintf(sdsempty(),"%.40s %s:%d@%d ",
+    ci = sdscatprintf(sdsempty(),"%.40s %s:%d@%d",
```
hpatro, line 4149: if we need to add hostname support, we would need to bring in extensions mechanism to older version as well.

src/cluster.c
```diff
@@ -4123,11 +4146,15 @@ sds clusterGenNodeDescription(clusterNode *node) {
...
         node->name,
         node->ip,
         node->port,
         node->cport);
+    if (node->hostname && sdslen(node->hostname) > 0) {
```
hpatro, line 4154: Would be good to add a nodes.conf file with hostname in it and verify it gets loaded fine.

hpatro (commented):
Thanks for submitting the change along with #2550 #2551.

Just a high level comment:

Could we make these change as individual commits to be able to stack on top of each other. Will be easier for us to create for 8.1/9.0 in the future.

OR

Individual commits based on the feature support like `listpack to ziplist`, `hostname support`, etc. This would enable folks to easily pick it for their needs and support older version if required in their case.

zuiderkwast:
See comment https://github.com/valkey-io/valkey/pull/2551#issuecomment-3228019282

hpatro:
Continuing the conversation here as it's the largest of all changes.

@zuiderkwast IIUC the intention here is not to merge into different branch of this project (already signalled by @satheeshaGowda by "NOT TO BE MERGED"). Rather with this PR we are able to visaulize the diff and provide feedback. We could maintain these patches independently under the Valkey organisation in a separate repo and folks can use it for their usecases where they might need to rollback after usage of new features in the latest version.

madolson:
Closing as part of stale PR cleanup — this is marked "NOT TO BE MERGED" and hasn't been updated since August 2025. Feel free to reopen if you'd like to formulate this into mergeable patches.

PR #2551, opened 2025-08-26
Roll backward downgrade compatibility from Redis 7.2 and Valkey 7.2/8.0 to Redis 7.0

satheeshaGowda:
===========NOT TO BE MERGED======

This patch extends Redis 7.0 to support  roll forward  upgrade and roll backward downgrade compatibility with 

- **RDB v11**: Redis 7.2
- **RDB v11**: Valkey 7.2
- **RDB v11**: Valkey 8.0

### Listpack to Ziplist Conversion Process
1. Parse listpack header and validate structure
2. Walk through each listpack element
3. Extract element data (string or integer)
4. Add each element to new ziplist structure
5. Return converted ziplist for normal processing


zuiderkwast:
We're not releasing Redis/Valkey 7.0.x and older. The first Valkey is 7.2.x.

Even if we would, these PRs need to be applied in the old nodes to be able to downgrade. It means it's not exactly a rollback to the software the user had before.

Another point is that we only backport bugfixes, small patches, to the old versions. We don't backport new features.

To be able to downgrade to an old version like this, I think it's better to add the ability in the latest version of Valkey to be able to produce old RDB versions. In this way, we don't need any changes in the old software.

To do it, we don't need to generate ziplist. We just need to avoid generqting listpacks. We can generate the rdb type like RDB_TYPE_HASH, etc.

We also know that old nodes don't send REPLCONF VERSION. If they don't, we can serve them old rdb data.

madolson:
> We're not releasing Redis/Valkey 7.0.x and older. The first Valkey is 7.2.x.

I think the point was never to merge these, but to have them around for review.

rainsupreme:
Closing this to take it out of the PR review queue. It can still be read and referred to, and can always be re-opened later.

PR #2552, opened 2025-08-26
Multi Threaded RDB Load

Nicky-2000:
# Multi-threaded RDB Load

### Note: 
This PR builds upon the changes in the Multi-threaded RDB Save PR (#2470). Therefore, this PR should be reviewed and merged after the Save PR. The diff is large because it includes all the changes from the save PR. 

This PR addresses Issue: https://github.com/valkey-io/valkey/issues/1483

### Contents
1. Introduction 
2. Performance Improvement using RDB Threads
3. Profiling Current, Single-Threaded, Implementation
4. Detailed Design
5. Testing
6. Future Extension

---
TODO: 
1. Add check to see if we have offloaded a chunk before we lock on DB insertion in the main thread. This way we can avoid excessive locking when loading rdb files that were not generated with multi-thread save and therefore do not have any chunks in them.

### Introduction:
This PR introduces a multi-threaded RDB Load function to improve the performance of restoring a Valkey dataset from an RDB file. This optimization in conjunction with the multi-threaded RDB Save PR also improve the performance of the Full Sync operation.

**Motivation for PR:**
Similar to the RDB Save, the current RDB Load is a single-threaded process that is a significant bottleneck for large datasets. This can lead to prolonged recovery, increased scaling times, and even impact instance or cluster availability. Since the load operation involves many CPU-intensive tasks like data parsing, decompression, and object creation, it is a good candidate for parallelization. This work directly addresses the need for faster recovery and more efficient cluster management.

---

# RDB Load and Full Sync Performance Improvements
### Test Environment:
Performance tests were run on a Google Cloud n2-highmem-48 VM with the following hardware configuration:
- **Machine Type:** n2-highmem-48
- **CPU:** 48 vCPUs (24 physical cores) on an Intel Cascade Lake platform.
- **Memory:** 384 GB of RAM.
- **Storage:** A persistent 1 TB SSD was attached for RDB file saves.
- **Write Bandwidth (Important)**: ~ 760 MB/s

For the RDB Load performance tests a single VM was used. For the Full Sync performance tests two identical VMs in the same region (with the same hardware as detailed above) were used.
 
### Workloads Tested
| Number of Keys | Value Size | Notes |
|:---:|:---:|:---|
| 50 Million | 50 B | Small Value Size Workload. |
| 50 Million | 300 B | Small Value Size Workload. |
| 50 Million | 500 B | Small Value Size Workload. |
| 3 Million | 5 KB | Medium Value Size Workload. |
| 3 Million | 20 KB | Medium Value Size Workload. |
| 1 Million | 100 KB | Large Value Size Workload. |
| 0.1 Million | 500 KB | Large Value Size Workload. |
| 0.1 Million | 1000 KB | Large Value Size Workload. |

All tests were run using the `run_full_load_benchmark.sh` test script located in the [valkey-rdb-benchmarking]( https://github.com/Nicky-2000/valkey-rdb-benchmarking). This bash script is a wrapper of [load_benchmark.py](https://github.com/Nicky-2000/valkey-rdb-benchmarking/blob/main/scripts/load_benchmark.py) python script with different arguments passed in (different workload configurations).

### Key Metrics:
Speed up (num_threads) = (Single Thread Save Time) / (Save Time with 'num_threads')
Efficiency (num_threads) = (Speedup with 'num_threads') / ('num_threads')
   
### RDB Load Performance Improvements
The multi-threaded RDB Load implementation allows us to double the speed of of loading in an RDB file in many cases. We are now bottlenecked by the DB Insertion code path.
#### Speed Up and Efficiency For Small Value Size Workloads (50 - 500 Bytes)
Compression ON
With compression on, the performance gains on small keys is significant. We achieve around a ~2x performance increase.
Screenshot 2025-08-29 1 23 42 PM
Compression OFF
For small keys with compression off, we observe a similar 2x performance increase with the Multi-Thread RDB Load implementation. Running with 3 threads is optimal. 
Screenshot 2025-08-29 1 24 41 PM

#### Speed Up and Efficiency For Medium Value Size Workloads (5 KB - 20 KB)
Compression ON
Enabling compression for medium value size workloads yielded a speedup of~2.2x. Running with 3 threads is optimal. 
Screenshot 2025-08-29 1 25 52 PM

Compression OFF
For medium-sized keys with compression off, the improvement was less (around 1.6x).

Screenshot 2025-08-29 1 26 13 PM

#### Speed Up and Efficiency For Large Value Size Workloads (100 KB - 1 MB)
Compression ON
With compression on, large value size workloads saw the biggest speedup of up to 2.6x. This is because there is more CPU work to do in order to decompress the large keys.
Screenshot 2025-08-29 1 27 06 PM

Compression OFF
With compression off, we don’t see much of a speedup. This is because larger keys already result in fewer, larger read calls which nullifies the benefits we get from our “chunked reader” strategy. Workloads with key sizes that exceed the save buffer flush threshold will see no speedup at all.
Screenshot 2025-08-29 1 27 28 PM

### Full Sync Performance Improvements
In general we see a solid speedup of 2-3x on the Diskless Full Sync operation for small and medium keys. We are now bottlenecked by the DB insertion on the load path. The save operation is happening faster so if we can optimize the load path (and make it thread safe), we could be able to get another 3x improvement on top of this. For very small keys the performance improvement is less due to many keys needing to be inserted.

#### Speed Up and Efficiency For Small Value Size Workloads (50 - 500 Bytes)

Compression ON
With compression on, we see a limited performance improvement for very small keys (50B) and great improvement of ~3x for slightly larger small keys (500B). This is because with many very small keys we are inserting into Valkey’s hashtable more frequently causing a bottleneck.

Screenshot 2025-08-29 2 43 26 PM

#### Speed Up and Efficiency For Medium Value Size Workloads (5 KB - 20 KB)
Compression ON
We see a 2x performance improvement for medium size keys. This is because we are now bottlenecked on the DB insertion. The primary is forced to slow down the rate at which it is sending data because the replica cannot insert fast enough.
Screenshot 2025-08-29 2 44 53 PM

#### Speed Up and Efficiency For Large Value Size Workloads (100 KB)
Compression ON
For large keys we see a ~1.6x speedup in Full Sync performance.
Screenshot 2025-08-29 2 45 13 PM
---

# Profiling Current, Single-Threaded, RDB Load Implementation
To inform the design of RDB Load multi-thread, detailed performance profiling was conducted to identify the bottlenecks under different workloads.  

### Profiling Environment
Note: This profiling was done on a different machine than the "Performance Improvement Tests"
- **CPU:** 64 vCPUs (32 physical cores) on an AMD EPYC 7B13 platform.
- **Memory:** 117 GiB of RAM.
- **Storage:** A 500 GiB SSD was attached and mounted for RDB file saves and benchmarking.
- **Write Bandwidth (Important):** ~ 402 MB/s

### Profiling Results and Insights

The current RDB Load operation is limited by its single-threaded design, which creates a sequential "read, process, insert" pipeline.

When loading small, compressed keys, the CPU utilization is high. The dominant functions (dbAddRDBLoad, rdbGenericLoadStringObject and rdbLoadObject as seen in the Flamegraph below) are focused on computationally expensive tasks like decompression and object creation.

When loading large uncompressed keys the Throughput graph shows that the system's throughput is limited to 254 MB/s, well below the host machine's 400 MB/s disk I/O capacity. The CPU Utilization graph shows a significant drop because the single thread is forced to wait for large data chunks to be fully read before it can begin processing, causing the disk to sit idle.

The core issue is the sequential pipeline. This profiling motivates the need to decouple the disk reads from the data CPU tasks (data processing and insertion). 

The Flamegraph, Throughput, and CPU Utilization graphs used for profiling are shown below.

<div align="center">
  Flamegraph of RDB Load
</div>
<div align="center">
  Throughput (MB/s) of RDB Load vs Value size
</div>
<div align="center">
  CPU Utilization of RDB Load vs Value size
</div>

---

# Detailed Design
### RDB File Format Update: New Aux Field
To enable a multi-threaded RDB Load operation a new "Auxiliary Field" was added. This Aux Field is called `rdb-data-segment`. This metadata field informs the main reading thread of the exact byte length of the next data chunk to be read. This allows the main thread to simply read the specified data into a buffer and pass it directly to a worker thread for processing. This effectively decouples the reading and processing tasks. Additionally, since we are reading larger chunks of data from the file we should get more efficient IO. The figure below shows how this new field will appear in the RDB file.
<div align="center">
  
</div>

This new Aux Field will be generated by the Multi-Thread RDB Save process. This design is a perfect fit, because each save thread already buffers data before flushing it to disk. Therefore, right before flushing the buffer, the save thread will add the Aux Field describing its size. However, this does mean that the multi-thread RDB Load operation is dependent on the RDB files being created by the multi-threaded implementation of RDB Save.

**Notes on Backwards Compatibility:** 
Adding a new Aux Field to the RDB file is 100% backwards compatible. Valkey servers running earlier versions will parse this Aux Field and simply discard it. Therefore, aside from reading a small, negligible, amount of extra data there is no downside to adding this Aux Field.

### Work Allocation

In the multi-threaded RDB Load the main thread will act as a dedicated reader. It will perform bulk reads of raw bytes from the dump.rdb file. It will be able to determine the exact size of the next data chunk to read using the Auxiliary Field (described above). Once a chunk is read, the main thread creates a task and inserts it into a queue for the worker threads to process. Each thread will have its own JobQueue and the main thread will round robin allocate work. As the sole reader of the file, the main thread also has the responsibility of validating the file's checksum. 

The worker threads will pick up tasks from their queue and perform the CPU-intensive work of decompression and deserialization. The figures below show the process of the main thread reading and distributing chunks of data to the worker threads.

<div align="center">
  
</div>
<div align="center">
  
</div>

### Data Processing and DB Insertion
Once a worker thread receives a data chunk, it decodes the raw bytes into the appropriate Valkey data types and temporarily store a batch of keys and values in a local buffer. When the buffer is full, the thread acquires the shared "Insert Lock" and adds its batch of keys to the database. This single lock ensures that only one thread can insert at a time, making the process safe and free of race conditions.

RDB Save Multi-Thread Diagrams (59)

### Note on Parallel DB Insertions
Lots of effort was put into parallelizing the insert path, but this approach was ultimately abandoned in favor of the single insert lock. The Valkey insert path is highly complex with many side effects, global state variable modifications, and incremental rehashing. This complexity makes it very hard to make the insert path thread-safe.

A prototype using a "striped locking pattern" was explored, where threads acquired locks for specific hash table buckets. Though this method showed promising performance in some tests (4x improvement in runtime), it was impossible to handle all edge cases. The high risk of data corruption and unpredictable behavior made this approach unfeasible. The single insert lock was chosen to maintain data integrity and align with Valkey's existing threading model.

---

# Testing 

### TCL Integration Tests
A new TCL test configuration flag (rdb-threads) was added. This flag configures the Valkey servers to run with rdb-threads active.

The full Valkey TCL test suites were run with the rdb-threads flag enabled. All existing tests passed, confirming that this implementation is backward-compatible and introduces no regressions or unintended changes to the RDB format or replication protocol.

Testing command used:  ./runtest --large-memory --rdb-threads

I also modified integration/rdb.tcl, integration/replication-2.tcl and integration/replication-3.tcl to run in a loop with different number of rdb-threads enabled.

I am open to feedback on this testing strategy. We could add a "daily" job (like there is for IO threads) that will run the test suite with the --rdb-threads flag.

---

# Future Extension: Multi-Socket Full Sync

### Current Multi-Thread Full Sync Operation (Using Multi-Thread RDB Save and Load)
At a high level, the Full Sync operation is a two-step process: a save on the primary server followed by a load on the replica. This is shown in Figure 28. Therefore, our multi-threaded RDB Save and Load implementation directly parallelized this entire operation, as the same underlying APIs are used. The figure below shows how the primary server now generates the RDB file with multiple threads, and the replica concurrently loads it with multiple threads (assuming both servers have the rdb-threads parameter enabled).
Screenshot 2025-08-28 3 07 16 PM

### Future Full Sync Dedicated Improvements
The main primary limitation of the current parallelization is that the data transfer still occurs over a single TCP connection, which can become a bottleneck. To address this, we need to use multiple sockets to stream the data from the primary to the replica. The figure below shows the high level idea of what we should move towards.
Screenshot 2025-08-28 3 06 02 PM


sarthakaggarwal97:
This has a lot of great ideas but unfortunately not being able to move. I also see that RDB Save (on top of which this PR is built is closed). I would close this too. Thank you @Nicky-2000 for all the efforts. If you ever feel that you would want to resume these efforts, please reopen both of the PRs.

